PR #183, opened 2015-01-04
Hostdb

DavidVorick:
I'm not 100% sure that this is complete, but I moved all of the hostdb stuff into its own package. I also switched from using an array that you went through in linear time every request to using a tree that only requires log time every request.

It's a weighted binary tree, but has no sorting. The tree helps with the weight, and there's a map to do lookups because we don't really care what order stuff is in, just that it's weighted. There are a few tests to make sure that the tree works alright.

### DIFF
--- a/sia/hostdb/hostannouncement.go
+++ b/sia/hostdb/hostannouncement.go
@@ -2,6 +2,7 @@ package hostdb
 
 import (
 	"github.com/NebulousLabs/Sia/consensus"
+	"github.com/NebulousLabs/Sia/encoding"
 	"github.com/NebulousLabs/Sia/network"
 )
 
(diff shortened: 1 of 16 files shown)


sia/hostdb/hostannouncement.go
```diff
@@ -0,0 +1,83 @@
...
+
+func findHostAnnouncements(height consensus.BlockHeight, b consensus.Block) (entries []HostEntry, err error) {
+	for _, t := range b.Transactions {
+		// Check the arbitrary data of the transaction to fill out the host database.
+		if len(t.ArbitraryData) == 0 {
+			continue
+		}
+		if len(t.ArbitraryData[0]) < 8 {
+			continue
+		}
```
lukechampine: redundant check?

lukechampine: code like this makes me question the decision to make ArbitraryData a `[]string`. What's the payoff again?

sia/hostdb/weightedlist.go
```diff
@@ -0,0 +1,106 @@
...
+		return hn.right.entryAtWeight(weight)
+	}
+
+	// Sanity check
+	if hn.hostEntry == nil {
+		panic("should not be returning a nil entry")
+	}
+
+	// Return the root entry.
+	entry = *hn.hostEntry
```
lukechampine: it'll panic anyway if it tries to dereference `nil`

lukechampine:
Is a tree necessary? I suppose it depends on how many entries there are, and how frequently the HostDB is queried. I think it's fair to assume that there will be a lot of entries, but queries won't be terribly frequent. Surely the performance boost is negligible unless you're querying more than once every few seconds. Trees require more effort to maintain; there are more pitfalls in their implementation, and I can't imagine it'll be fun to encode/decode them.

DavidVorick:
Easy to encode and decode, just pull out the entries. It's not consensus or anything, everyone can have trees that look different and it won't cause problems. Especially since they'll all be functionally equivalent anyway.

I just wasn't happy with doing a linear search every time we wanted to pull a random host. Theoretically you're going to be pulling dozens to hundreds of hosts all at once, out of a pool of hopefully thousands. Linear will make that hundreds of thousands of array accesses, which isn't terrible but it's not great either.

Perhaps a bit of over optimization but at this point I've written a decent number of tests and I don't think tree failure is a risk.

DavidVorick:
I think that I also did it because over break I felt like working with trees. This was perhaps also a mistake, because it might introduce code burden in the future.

It's quite modular though, the new interface model keeps things compartmentalized well. Changing the tree won't affect anything outside of the hostdb package.

lukechampine:
True. Interfaces definitely work to our favor here. And linear search is definitely not pretty if it will truly require 100k accesses.

I guess the real issue with encoding/decoding is that you may not want to load the entire HostDB into memory (is this realistic?). Traversing and manipulating a tree on-disk == not fun.

sia/hostdb/hostannouncement.go
DavidVorick: Makes soft forks much much easier. Also, it adds flexibility for stuff like the code above, I just don't take advantage of the flexibility yet.

DavidVorick:
You'd want the whole thing in memory if you're selecting a bunch of items at random. If it becomes cumbersome, we could build the tree out of a random subset of hosts and just ignore the rest.

PR #184, opened 2015-01-08
RPC errors

lukechampine:
RPCs now send and receive errors. Registered functions like `SendBlocks` that return an error will write that error to the connection. If it's a `net.Conn` function like `RetrieveFile`, you have to write the error manually. All existing RPCs have been updated to make use of errors. However, there still isn't much logic for processing specific RPC errors, except in `CatchUp`, which now looks for a specific "more blocks are available" error.

### DIFF
--- a/encoding/prefix.go
+++ b/encoding/prefix.go
@@ -19,21 +19,9 @@ func ReadPrefix(r io.Reader, maxLen uint64) ([]byte, error) {
 		return nil, fmt.Errorf("length %d exceeds maxLen of %d", dataLen, maxLen)
 	}
 	// read dataLen bytes
-	var data []byte
-	buf := make([]byte, 1024)
-	var total uint64
-	for total = 0; total < dataLen; {
-		n, err := r.Read(buf)
-		if err != nil {
-			return nil, err
-		}
-		data = append(data, buf[:n]...)
-		total += uint64(n)
-	}
-	if total != dataLen {
-		return nil, errors.New("length mismatch")
-	}
-	return data, nil
+	data := make([]byte, dataLen)
+	_, err := io.ReadFull(r, data)
+	return data, err
 }
 
 // ReadObject reads and decodes a length-prefixed and marshalled object.
--- a/sia/synchronize.go
+++ b/sia/synchronize.go
@@ -11,6 +11,8 @@ const (
 	MaxCatchUpBlocks = 100
 )
 
+var moreBlocksErr = errors.New("more blocks are available")
+
 // SendBlocks takes a list of block ids as input, and sends all blocks from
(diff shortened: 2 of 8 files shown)


sia/synchronize.go
```diff
@@ -85,20 +92,20 @@ func (c *Core) CatchUp(peer network.Address) {
...
+		c.AcceptBlock(block)
 	}
 
 	// TODO: There is probably a better approach than to call CatchUp
 	// recursively. Furthermore, if there is a reorg that's greater than 100
 	// blocks, CatchUp is going to fail outright.
-	if prevHeight != c.Height() {
-		go c.CatchUp(c.RandomPeer())
+	if err == moreBlocksErr {
+		go c.CatchUp(peer)
```
lukechampine: I'm reusing the old peer here, but that might not be best practice. On one hand, if a peer tells you it has more blocks, you probably want to ask that specific peer again. On the other hand, this allows a malicious host to trap you in an infinite CatchUp loop by always returning `moreBlocksErr`.

sia/synchronize.go
```diff
@@ -85,20 +92,20 @@ func (c *Core) CatchUp(peer network.Address) {
...
 	}
```
lukechampine: I'm reusing the old peer here, but that might not be best practice. On one hand, if a peer tells you it has more blocks, you probably want to ask that specific peer again. On the other hand, this allows a malicious host to trap you in an infinite CatchUp loop by always returning `moreBlocksErr`.

encoding/prefix.go
```diff
@@ -19,21 +19,9 @@ func ReadPrefix(r io.Reader, maxLen uint64) ([]byte, error) {
...
-		}
-		data = append(data, buf[:n]...)
-		total += uint64(n)
-	}
-	if total != dataLen {
-		return nil, errors.New("length mismatch")
-	}
-	return data, nil
+	data := make([]byte, dataLen)
+	_, err := io.ReadFull(r, data)
```
DavidVorick: That's a pleasant change.

sia/synchronize.go
DavidVorick: this is avoidable by checking yourself to see that the host has actually supplied more blocks.

DavidVorick:
Can you add more documentation? Specifically a file somewhere that explains the general philosophies and approaches behind the network package?

DavidVorick:
I'll probably make an issue about it.

encoding/prefix.go
lukechampine: yes indeed. The standard library has a lot of useful helper functions.

lukechampine:
sure. I'm thinking the package should be renamed anyway (sianet?), so I'll lump those two changes together.

PR #185, opened 2015-01-08
Refactor the Sia package

DavidVorick:
Somehow this turned from me bringing back the host functions to me refactoring everything. And the idea of keeping PR's small completely fell apart, probably partially because I've been using my laptop which doesn't have X working right now - just a bunch of ttys.

There are some pretty substantial changes here, and some more on the way but I'm still figuring out the best order to do everything in. I've come up with a convention for working with mutexes which I think will significantly reduce mutex based developer errors. I've already caught a handful of subtle race conditions (which really would probably not ever cause problems but good to fix them anyway) and I find things a lot easier to reason about now. I've increased the documentation for the conventions I've been using, and this can be found in ProgramFlow.info

One of the biggest changes is the introduction of the components package, which moves all of the interfaces of the core into a separate package. This prevents circular dependencies and enables everything to be grabbed from the components package. It allows components to use each other, and to use the state, and it allows the sia package to import each of the component implementations (which are all their own packages).

There's still a decent amount of distance to go unfortunately. Most of the interfaces (except for host and renter, which are still incomplete) are good, but the Core doesn't seem to have a mutex, and it really needs one. And I'm not happy with the way that the Core, State, and components all communicate changes/updates, it's loose and prone to mistakes/errors.

I'm going to slow down the refactoring a bit though so that I can get the renter and hosts working again, and add pruning to the hostdb. But I do feel strongly about changing the way that the core+components+state all stay on the same page, and that might take some time to figure out and implement.


ProgramFlow.info
```diff
@@ -1,17 +1,52 @@
...
+the state as the result of information being introduced to the state. For
+example 'ApplyBlock' and 'applyStorageProof'.
 
-Within the consensus package, it is convention to use the term 'valid' any time
-that the validity of something is being checked. It is convention not to change
-the state in any capacity when inside of a validation function. For example
-'validInput' and 'validContract'.
+It is convention to use the term 'invert' any time that a function is changing
+the state as the result of information being removed from the blockchain. For
+example 'invertBlock'.
```
lukechampine: not "revert?"

sia/components/renter.go
```diff
@@ -0,0 +1,14 @@
...
+)
+
+type RenterUpdate struct {
+	HostDB HostDB
+}
+
+type Renter interface {
+	// UpdateRenter changes the settings used by the host.
+	UpdateRenter(RenterUpdate) error
+}
```
lukechampine: This design is a bit confusing to me. I would expect the `Renter` to have a `Rent` method. Also, instead of a single `UpdateRenter` method, I would expect a separate method for each setting, like `UpdateHostDB`.

sia/core.go
```diff
@@ -1,27 +1,46 @@
...
-	"github.com/NebulousLabs/Sia/sia/hostdb"
-	"github.com/NebulousLabs/Sia/sia/miner"
-	"github.com/NebulousLabs/Sia/sia/wallet"
+	"github.com/NebulousLabs/Sia/sia/components"
 )
 
+// CoreInput is just to prevent the inputs to CreateCore() from being
+// excessively long. Potentially it could be used as a struct within Core
+// instead of manually copying over all of the inputs.
+type Config struct {
```
lukechampine: fix docstring

lukechampine:
I'm a fan of the components package. It keeps all the high-level information about the important types/interfaces in one place.

sia/core.go
```diff
@@ -36,31 +55,70 @@ type Core struct {
...
+		err = errors.New("cannot have nil renter")
+		return
+	}
+	if config.Wallet == nil {
+		err = errors.New("cannot have nil wallet")
+		return
+	}
+
+	// Set the number of procs equal to the number of cpus.
+	runtime.GOMAXPROCS(runtime.NumCPU())
```
lukechampine: I feel like this belongs in an `init` function rather than `CreateCore`.

sia/core.go
```diff
@@ -36,31 +55,70 @@ type Core struct {
...
+	}
+	err = c.hostDB.Update(0, nil, []consensus.Block{genesisBlock})
+	if err != nil {
+		return
+	}
+	err = c.UpdateMiner(c.miner.Threads())
+	if err != nil {
+		return
+	}
+	err = c.UpdateRenter()
```
lukechampine: these `Update` functions are quite opaque. `hostDB.Update` could have a more descriptive name, and I have no idea what `UpdateMiner` or `UpdateRenter` are doing.

sia/host/contracts.go
```diff
@@ -0,0 +1,214 @@
...
+// transaction, we store only the id.
+type ContractEntry struct {
+	ID       consensus.ContractID
+	Contract consensus.FileContract
+}
+
+func (h *Host) getFilename() string {
+	h.fileCounter++
+	return strconv.Itoa(h.fileCounter)
+}
```
lukechampine: To me, a function named `get` implies that no state is being changed. Maybe it should be `nextFilename`?

sia/host/contracts.go
```diff
@@ -0,0 +1,214 @@
...
+	penalty := h.announcement.Burn * consensus.Currency(fileSize) * consensus.Currency(contractDuration)
+	id, err := h.wallet.RegisterTransaction(t)
+	if err != nil {
+		return
+	}
+	err = h.wallet.FundTransaction(id, penalty)
+	if err != nil {
+		// TODO: This leaks that the host is out of money.
+		return
+	}
```
lukechampine: hmm. This is tricky. I assume there are good reasons why we wouldn't want to leak that info. We could return a generic error, like `"contract refused"` or `"internal error"`, but even those would be a giveaway for anyone who's read the source code.

sia/host/contracts.go
```diff
@@ -0,0 +1,214 @@
...
+	if err != nil {
+		return
+	}
+	defer file.Close()
+
+	// If there's an error upon return, delete the file that's been created.
+	defer func() {
+		if err != nil {
+			os.Remove(filename)
+		}
```
lukechampine: could also `h.fileCounter--` here, though I don't suppose it really matters.

sia/host/contracts.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+
+	// record filename for later retrieval
+	h.files[t.FileContracts[0].FileMerkleRoot] = filename
+
+	// Submit the transaction.
+	h.transactionChan <- t
+	if err != nil {
+		return
+	}
```
lukechampine: what err??

sia/host/host.go
```diff
@@ -0,0 +1,105 @@
...
+	transactionChan chan consensus.Transaction // Can send channels to the state.
+
+	hostDir     string
+	fileCounter int
+	files       map[hash.Hash]string
+
+	forwardContracts  map[consensus.BlockHeight][]ContractEntry
+	backwardContracts map[consensus.BlockHeight][]ContractEntry
+
+	rwLock sync.RWMutex
```
lukechampine: any reason why the mutex is not embedded here?

sia/host/lock.go
```diff
@@ -0,0 +1,21 @@
...
+
+// rLock read-locks the host.
+func (h *Host) rLock() {
+	h.rwLock.RLock()
+}
+
+// rUnlock read-unlocks the host.
+func (h *Host) rUnlock() {
+	h.rwLock.RUnlock()
+}
```
lukechampine: uh...

sia/host/proofs.go
```diff
@@ -0,0 +1,139 @@
...
+//
+// TODO: Make sure that hosts don't need to submit a storage proof for the last
+// window.
+func (h *Host) storageProofMaintenance(initialStateHeight consensus.BlockHeight, rewoundBlocks []consensus.Block, appliedBlocks []consensus.Block) {
+	// Resubmit any proofs that changed as a result of the rewinding.
+	height := initialStateHeight
+	var proofs []consensus.StorageProof
+	for _ = range rewoundBlocks {
+		needActionContracts := h.backwardContracts[height]
+		for _, contractEntry := range needActionContracts {
```
lukechampine: just do `for _, contractEntry := range h.backwardContracts[height] {`

sia/hostdb/hostdb.go
```diff
@@ -84,14 +114,16 @@ func (hdb *HostDatabase) Update(initialStateHeight consensus.BlockHeight, rewoun
...
+		var entries []components.HostEntry
 		entries, err = findHostAnnouncements(initialStateHeight, b)
 		if err != nil {
 			return
 		}
 
 		for _, entry := range entries {
+			hdb.unlock()
 			err = hdb.Insert(entry)
+			hdb.lock()
```
lukechampine: mutex convention makes this kinda ugly. You could have an `insert` function that was `Insert` without the locks, and then define `Insert` as:
```
func (hdb *HostDB) Insert(entry) error {
    hdb.lock()
    defer hdb.unlock()
    return hdb.insert(entry)
}
```

sia/hostdb/lock.go
```diff
@@ -0,0 +1,21 @@
...
+
+// rLock read-locks the hostDB.
+func (hdb *HostDB) rLock() {
+	hdb.rwLock.RLock()
+}
+
+// rUnlock read-unlocks the hostDB.
+func (hdb *HostDB) rUnlock() {
+	hdb.rwLock.RUnlock()
+}
```
lukechampine: just embed the mutex

sia/hostdb/weightedlist.go
```diff
@@ -17,27 +23,31 @@ type hostNode struct {
 	left  *hostNode
 	right *hostNode
 
-	hostEntry *HostEntry
+	taken     bool // Used because components.HostEntry can't be compared to nil.
+	hostEntry components.HostEntry
```
lukechampine: why switch from a pointer?

ProgramFlow.info
DavidVorick: invert was actually confusing me because it looks a lot like insert.

I'll probably switch to revert when I rewrite the consensus package, which will be further down the road.

sia/miner/lock.go
```diff
@@ -0,0 +1,21 @@
...
+
+// rLock read-locks the miner.
+func (m *Miner) rLock() {
+	m.rwLock.RLock()
+}
+
+// rUnlock read-unlocks the miner.
+func (m *Miner) rUnlock() {
+	m.rwLock.RUnlock()
+}
```
lukechampine: I'm starting to wonder if there's a good reason to do this, but I can't think of one.

sia/components/renter.go
DavidVorick: Renter is very far from complete. I've also been waffling on the monolithic style of calling a single update function. It'll definitely have a Rent() function. Tolerance and whatnot will probably be decided on a per-file basis.

Actually, I was thinking file could have high level tags (like 'Chicago', 'High Redundancy', 'Low Latency'; or 'Global', 'High Throughput') and then these would translate into more specific details that the renter can modify itself or can be adjusted through inputs. But that's also all down the line for now we'll just have one renter-controlled setting.

ProgramFlow.info
lukechampine: btw, I believe files like this are usually named `DEVELOPERS.md`.

sia/core.go
DavidVorick: I was really unsure where to put it. Main? Daemon? But it seemed to me like Core is the package that's calling the shots, so it should be manipulating the runtime environment to suit itself.

sia/core.go
DavidVorick: They are really doing initialization. I'm not happy with the current update model but I'm also not sure it would be good to fracture it a bunch.

sia/host/host.go
lukechampine: OH I get it. It's so that the locks aren't exported.

I guess it's safer this way. But having a whole `lock.go` file in all these packages really bothers me.

sia/host/contracts.go
DavidVorick: Privacy is always better than not privacy. If some attack involves draining a host's funds, this is a good way for the attacker to tell if they've executed it properly. There are good reasons why you turn off debugging information in php in production environments.

I don't know a good way to disguise this problem though. For the beta it's fine. We could just return 'storage full', because that's effectively true given that the host can't follow through on their promised burn.

sia/host/contracts.go
DavidVorick: That would cause a race condition, because the lock is only in place the first time we get the name. If we were to defer a `h.fileCounter--` we'd need to keep the lock in place for the full duration of the function.

Since we're breaking mutex convention, I like making as few reads and writes as possible, because mutex mistakes are really easy to make.

sia/host/contracts.go
DavidVorick: lol

sia/host/host.go
DavidVorick: can't embed a mutex unless you export it, at least as far as I understand. I don't want external packages calling lock or unlock.

DavidVorick: =/ yeah it's not the cleanest thing ever. But I didn't want to be writing host.lock.RLock() everywhere.

sia/host/proofs.go
DavidVorick: this whole file is getting rewritten

sia/hostdb/hostdb.go
DavidVorick: I actually did that in multiple other places, this one must have slipped thorugh

sia/host/host.go
lukechampine: I guess it's slightly safer this way. However, you'll only be able to call the locking functions if you're working with the `Host` *type*, not the `components.Host` interface. My understanding is that we are almost always using the latter, so it's safe to embed to the lock.

sia/hostdb/weightedlist.go
DavidVorick: Pointers can cause problems. I didn't see any need for it to be a pointer, so I removed the potential pitfall.

DavidVorick:
I think that I submitted two commits since you started reviewing, might want to glance over the diffs on each. (address comments and the one before it).

Tests pass, etc.

lukechampine:
Still not terribly happy with the locks, but I guess I'll come back to it later.

PR #186, opened 2015-01-13
Host, Renter, State subscriptions

DavidVorick:
I reimplemented host and rent. Host now works by subscription, and should work a decent amount better than before (meaning less bugs when submitting storage proofs). There are still an handful of problems, but these are detailed in a series of TODOs.

I implemented renter so that you upload to multiple places and then download from multiple places. So there's naive redundancy happening. At the moment there is no file repair whatsoever. That should be alright for now.

I had to add a subscription based model of passing out updates to the state. Overall, I think that this is very much the right thing to do. Unfortunately, I didn't have time to completely fix how the state works to accommodate this. So the subscription updates actually give incomplete and misleading information, but in such a way that the host (who is currently the only one using the subscription stuff) is completely unaffected (... at least... I think).

The code was basically just shoved into place. I will go back and make it nice, and on the way probably remove a ton of complicated code from the state that won't need to be there anymore (like I said, this change makes sense for a lot of reasons). This will take me a lot of time, and won't be done by release which is why I chose to just shove things together for the time being. I'll clean it up afterwards.


DavidVorick:
This PR has no testing for the hosting or renting, that'll be next.

consensus/notifications.go
```diff
@@ -0,0 +1,84 @@
+package consensus
+
+// A block is composed of many transactions. Blocks that have transactions that
+// depend on other transactions, but the transactions must all be applied in a
+// deterministic order. Transactions cannot have inter-dependencies, meaning
+// that an output cannot be created and then spent in the same transaction. As
+// far as diffs are concenred, this means that the elements of a transaction
+// diff should be able to be applied in any order and the outcome should be the
+// same. The elements of a block diff however must be applied in a specific
+// order, as transactions may depend on each other.
```
lukechampine: I'm confused as to why this comment is up here. It seems like it should be split up and moved to the `TransactionDiff` and `BlockDiff` docstrings.

consensus/notifications.go
```diff
@@ -0,0 +1,84 @@
...
+//
+// TODO: What happens if a subscriber stops pulling info from their channel. If
+// they don't close the channel but stop pulling out elements, will the system
+// lock up? If something stops responding suddenly, there needs to be a way to
+// keep going, the state just deletes, closes, or ignores the channel or
+// something. Perhaps the state will close the channel if the buffer fills up,
+// assuming that the component has shut down unexpectedly. If the component was
+// just being slow, it can do some catching up and re-subscribe. If we do end
+// up closing subscription channels then we should switch from a slice to a
+// map for s.consensusSubscriptions.
```
lukechampine: This shouldn't be an issue unless the subscription model is very complex. I don't expect to see subscribers appearing and disappearing regularly; more likely, each component will subscribe during initialization, and remain open until termination.

The only worry, then, is that the send operation can block, delaying propagation to other subscribers. This could be fixed by using buffered channels, or by spawning a new goroutine for each send (not recommended).

Note that attempting to send down a closed channel causes a panic.

sia/components/hostdb.go
```diff
@@ -20,18 +20,17 @@ const (
 // interfaces.
 
 type HostDB interface {
+	// FlagHost alerts the HostDB that a host is not behaving as expected. The
+	// HostDB may decide to remove the host, or just reduce the weight, or it
+	// may decide to ignore the flagging. If the flagging is ignored, an error
+	// will be returned explaining why.
+	FlagHost(id string) error
```
lukechampine: An error probably isn't terribly useful here, unless you expect the flagger to change their behavior depending on whether or not their flag was accepted. I imagine down the road we'll want more sophisticated flagging logic, but I don't think errors are of much use there either.

sia/components/renter.go
```diff
@@ -1,14 +1,6 @@
...
-
-type RenterUpdate struct {
-	HostDB HostDB
-}
-
 type Renter interface {
-	// UpdateRenter changes the settings used by the host.
-	UpdateRenter(RenterUpdate) error
+	RenameFile(currentName, newName string) error
+	RentFile(filename, nickname string, totalPieces, requiredPieces, optimalRecoveryPieces int) error
```
lukechampine: a struct might be nicer here...it feels kinda like overkill, but on the other hand it lets the caller write:
```go
r.RentFile(components.File{
    Name: "foo",
    Nickname: "bar",
    TotalPieces: 1,
    RequiredPieces: 2,
    OptimalRecPieces: 3,
})
```
instead of:
```go
r.RentFile("foo", "bar", 1, 2, 3) // what do these numbers mean??
```

sia/host/contractcreation.go
```diff
@@ -190,21 +194,21 @@ func (h *Host) NegotiateContract(conn net.Conn) (err error) {
...
 
 	// Put the contract in a list where the host will be performing proofs of
 	// storage.
-	firstProof := t.FileContracts[0].Start + StorageProofReorgDepth
-	h.forwardContracts[firstProof] = append(h.forwardContracts[firstProof], ContractEntry{ID: t.FileContractID(0), Contract: t.FileContracts[0]})
+	co := contractObligation{
+		inConsensus: false,
+		filename:    filename,
+	}
+	h.contracts[t.FileContractID(0)] = co
```
lukechampine: you can just use a literal here. It's not always the right choice, but I feel like if your variable name is `co` it should probably just be a literal.

sia/renter/renter.go
```diff
@@ -143,30 +146,90 @@ func (e *Core) ClientProposeContract(filename, nickname string) (err error) {
...
+	// We just need to get one piece, we'll keep contacting hosts until one
+	// doesn't return an error.
+	for _, piece := range entry.Pieces {
+		err = r.downloadPiece(piece, filename)
+		if err != nil {
+			return
+		} else {
+			fmt.Println(err)
+			r.hostDB.FlagHost(piece.Host.ID)
+		}
```
lukechampine: pretty sure you have these branches reversed.

consensus/notifications.go
DavidVorick: I believe  there is a way to wrap the channel to see if it's got elements in it.

I was just thinking about this for robustness. If some part of the program goes into an error state, I don't want it to freeze the whole program, especially if it's a non-critical process like the cpu miner vs. something like a host or renter.

sia/components/hostdb.go
DavidVorick: Component functions pretty much always return an error. But perhaps what I've stated in the docstring is a bad approach.

sia/components/renter.go
DavidVorick: sounds good to me

consensus/notifications.go
lukechampine: True. In that case, though, I wonder if this subscription model is really the best choice. It shifts certain responsibilities from the components to the state. The mutex approach wasn't fantastic either, but at least one failing component couldn't bring the whole system down.

Actually, I take that back, since components could just `Lock()` and never `Unlock()`. So I guess the goal is a design that prevents components from sabotaging each other.

consensus/notifications.go
DavidVorick: I think I was just explaining things overall here. I'll take a closer look at the documentation when I revamp the consensus code.

DavidVorick:
Made my fix commits.

lukechampine:
"I'll clean it up afterwards" never fails to make me uneasy.

DavidVorick:
New rule: stuff that involves the sia package needs to be code reviewed by Seve in addition to Luke b/c he's working on stuff that depends on the sia package, and needs to see all of the changes.

also, I've thus far followed through on pretty much every promise of "will clean it up later".

Issue #163, opened 2014-12-15
Updating


lukechampine:
We need a way for users to automatically update. This would be especially helpful during closed/open beta: instead of publishing each new release and asking people to update, we could just push out an update and be confident that everyone saw it.

There's a service I saw a while ago called [Equinox](https://equinox.io), which might be our best bet, though I don't fully know how it works. The alternative would be to roll our own updater.

EDIT: Equinox is a paid service. It basically just provides servers for users to download from. The actually update library is called [go-update](https://github.com/inconshreveable/go-update). Incidentally, it looks like the binary patching algorithm used by go-update was designed by Colin Percival.

I wonder how Bitcoin handles updates. Having people download updates from (our) centralized servers seems reasonable for the beta, but after that we'll probably want something more decentralized. I can imagine nodes sharing updates (signed, of course) with their peers.

lukechampine:
There are quite a few more features I'd like to implement before considering this "closed." I can just make new issues for them if that makes more sense though.

DavidVorick:
I think that I would prefer that.

Having these large scope issues makes me uneasy, not because they're bad but just because it makes it seem like no progress is being made.

PR #188, opened 2015-01-14
Updates

lukechampine:
Basic update support (#163), and some tweaks to siad.

Updates work like this: each version is stored in a folder on the Linode. The most recent version is stored in `current/`. The folder contains the files changed by the update, as well as a `MANIFEST` file that contains the version number and a file listing. To check for an update, siad just reads the version number from `current/MANIFEST`. If the version is newer, it downloads and applies the files listed in the update.

There's a lot of room for improvement here. And I haven't actually tried updating from the server yet.

### DIFF
--- a/sia/core.go
+++ b/sia/core.go
@@ -139,15 +139,13 @@ func CreateCore(config Config) (c *Core, err error) {
 	*/
 
 	// Bootstrap to the network (may take a few seconds).
-	go func() {
-		err = c.initializeNetwork(config.ServerAddr, config.Nobootstrap)
-		if err == network.ErrNoPeers {
-			// log.Println("Warning: no peers responded to bootstrap request. Add peers manually to enable bootstrapping.")
-		} else if err != nil {
-			return
-		}
-		// c.host.Settings.IPAddress = c.server.Address()
-	}()
+	err = c.initializeNetwork(config.ServerAddr, config.Nobootstrap)
+	if err == network.ErrNoPeers {
+		// log.Println("Warning: no peers responded to bootstrap request. Add peers manually to enable bootstrapping.")
+	} else if err != nil {
+		return
+	}
+	// c.host.Settings.IPAddress = c.server.Address()
 
 	return
 }
--- a/sia/core.go
+++ b/sia/core.go
@@ -138,14 +138,16 @@ func CreateCore(config Config) (c *Core, err error) {
 	}
 	*/
 
-	// Bootstrap to the network.
-	err = c.initializeNetwork(config.ServerAddr, config.Nobootstrap)
-	if err == network.ErrNoPeers {
-		// log.Println("Warning: no peers responded to bootstrap request. Add peers manually to enable bootstrapping.")
-	} else if err != nil {
-		return
-	}
-	// c.host.Settings.IPAddress = c.server.Address()
+	// Bootstrap to the network (may take a few seconds).
+	go func() {
+		err = c.initializeNetwork(config.ServerAddr, config.Nobootstrap)
+		if err == network.ErrNoPeers {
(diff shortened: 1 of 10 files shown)


sia/core.go
```diff
@@ -138,14 +138,16 @@ func CreateCore(config Config) (c *Core, err error) {
...
+	// Bootstrap to the network (may take a few seconds).
+	go func() {
+		err = c.initializeNetwork(config.ServerAddr, config.Nobootstrap)
+		if err == network.ErrNoPeers {
+			// log.Println("Warning: no peers responded to bootstrap request. Add peers manually to enable bootstrapping.")
+		} else if err != nil {
+			return
+		}
+		// c.host.Settings.IPAddress = c.server.Address()
+	}()
```
lukechampine: looks like this goroutine is causing a data race, but I'm not totally sure why.

siad/api.go
```diff
@@ -191,6 +192,30 @@ func (d *daemon) downloadHandler(w http.ResponseWriter, req *http.Request) {
...
+			Available bool
+			Error     string
+		}{available, err.Error()})
+
+	case "apply":
+		applied, err := d.applyUpdate()
+		enc.Encode(struct {
+			Applied bool
+			Error   string
+		}{applied, err.Error()})
```
lukechampine: going to convert the rest of the API to this style (JSON only). I thought it'd be a pain at first, but the use of `NewEncoder` and struct literals makes things fairly clean.

sia/core.go
DavidVorick: We talked about it, but just to reiterate:

the go func() should probably happen inside of the initializeNetwork(), because some things potentially need to be initialized before you continue processing stuff. You don't need to find peers right away.

siad/api.go
DavidVorick: I think I would prefer these structs declared, so they can have a docstring. We'll get seve's opinion

siad/main.go
```diff
@@ -29,56 +33,65 @@ type Config struct {
...
-// report an error for that. It'll only report an error if it can't find
-// template.html.
-func findSiaDir() (home, siaDir string, err error) {
-	// Check the current directory for the template file.
-	var found bool
-	if _, err = os.Stat("style/template.html"); err == nil {
-		found = true
+// expand all ~ characters in Config values
+func (c *Config) expand() (err error) {
+	ex, err := homedir.Expand(c.Siacore.HostDirectory)
```
DavidVorick: ex is a bad variable name

siad/update.go
```diff
@@ -0,0 +1,77 @@
...
+	"runtime"
+	"strconv"
+	"strings"
+
+	"github.com/inconshreveable/go-update"
+)
+
+const VERSION = "0.1.0"
+
+var updateURL = "http://23.239.14.98/releases/" + runtime.GOOS + "_" + runtime.GOARCH + "/current/"
```
DavidVorick: Document where this ip address comes from.

DavidVorick:
Need a docstring in the update file that gives an overview of how updating works. Something close to what you explained at the top of the package would be sufficient.

The reasoning is that someone who's new to the code will want to be able to get an overview without needing to read and piece together the docstrings of each individual function. The overview of course doesn't need to be terribly detailed.

siad/main.go
```diff
@@ -29,56 +31,77 @@ type Config struct {
...
+// Helper function for determining existence of a file. Technically, err != nil
+// does not necessarily mean that the file does not exist, but it does mean
+// that it cannot be read, and for our purposes these are equivalent.
+func exists(filename string) bool {
+	ex, err := homedir.Expand(filename)
 	if err != nil {
-		fmt.Println("Error starting client:", err)
+		return false
+	}
+	_, err = os.Stat(ex)
```
DavidVorick: exists (ex)

PR #189, opened 2015-01-14
Webinterface

seveibar:
Completely replaced old web interface. Several changes to Sia HTTP API, fixes and restructuring.

The initial commits produce a mockup application, later commits add a scripts directory that contains data bindings to update the application with actual data.

The interface has two global modules, ui and controller. The controller controls polling the go http server, while the ui takes the data and translates it to the interface. The ui also has a method to add listeners for events (e.g. "stop-mining") that the controller can bind to to send requests to the server.


sia/core_test.go
```diff
@@ -26,14 +26,27 @@ func establishTestingEnvironment(t *testing.T) (c *Core) {
...
+	hdb, err := hostdb.New()
+	if err != nil {
+		t.Fatal(err)
+	}
+	mrHost, err := host.New(state, w)
+	if err != nil {
+		return
+	}
+	mrRenter, err := renter.New(state, hdb, w)
+	if err != nil {
```
lukechampine: not a fan of this naming convention. Normally I would use one-letter names here, but that's a problem because you have both `hostdb` and `host`. Maybe use `HostDB, `Host`, and `Renter`.

Also, some of these call `t.Fatal` while others just `return`.

sia/host/host.go
```diff
@@ -43,16 +43,24 @@ type Host struct {
...
 
 // New returns an initialized Host.
-func New(s *consensus.State) (h *Host) {
+func New(state *consensus.State, wallet components.Wallet) (h *Host, err error) {
+	if wallet == nil {
+		err = errors.New("host.New: cannot have nil wallet")
+	}
+	if state == nil {
+		err = errors.New("host.New: cannot have nil state")
+	}
```
lukechampine: need `return` statements here

sia/wallet.go
```diff
@@ -52,6 +53,6 @@ func (c *Core) CoinAddress() (consensus.CoinAddress, error) {
 }
 
 // Returns a []byte that's supposed to be json of some struct.
-func (c *Core) WalletInfo() ([]byte, error) {
+func (c *Core) WalletInfo() (components.WalletStatus, error) {
```
lukechampine: either `WalletInfo()` should return a `WalletInfo`, or `WalletStatus()` should return a `WalletStatus`.

siad/daemon.go
```diff
@@ -63,18 +64,26 @@ func createDaemon(config Config) (d *daemon, err error) {
 		downloadDir: expandedDownloadDir,
 	}
 
-	// Create the web interface template.
-	d.template, err = template.ParseFiles(expandedStyleDir + "template.html")
+	// mr is used to resolve conflicts between packages and variable names
+
+	state, _ := consensus.CreateGenesisState() // the `_` is not of type error.
+	mrWallet, err := wallet.New(expandedWalletFile)
```
lukechampine: `Wallet` or `w`

PR #195, opened 2015-01-15
Pure JSON API

lukechampine:
The siad API now uses JSON exclusively. The API is documented in API.md.

siac is still fairly broken. To make things cleaner, I'm thinking of adding some helper functions/structs to siad/api.go and importing siad in siac. This should make it easier to change the API without having to fix things in 3 different places.


DavidVorick:
I'm not sure what's causing the build fail, but given that it passed on an earlier commit I think it just got unlucky with the transactions.

API.md
```diff
@@ -0,0 +1,67 @@
...
+maxfile 
+mintolerance 
+minduration 
+maxduration 
+minwin 
+maxwin 
+freezeduration 
+price 
+penalty 
+freezevolume
```
DavidVorick: no braces here but there are braces on the other one.

sia/host/announce.go
```diff
@@ -9,8 +9,8 @@ import (
 // HostAnnounceSelf creates a host announcement transaction, adding
 // information to the arbitrary data and then signing the transaction.
 func (h *Host) AnnounceHost(freezeVolume consensus.Currency, freezeUnlockHeight consensus.BlockHeight) (t consensus.Transaction, err error) {
-	h.lock()
-	defer h.unlock()
+	h.mu.Lock()
+	defer h.mu.Unlock()
```
DavidVorick: ah damn

DavidVorick: renaming should be pretty easy, just do a codebase scan for \.mu\.

siac/cmds.go
```diff
@@ -12,25 +12,26 @@ import (
 var hostname = "http://localhost:9980"
 
 // helper function for reading http GET responses
-func getResponse(handler string, vals *url.Values) string {
+func getResponse(handler string, vals *url.Values) (map[string]interface{}, error) {
 	// create query string, if supplied
-	qs := "?"
```
DavidVorick: can this be renamed to `query`? `qs` is pretty opaque, even if the comment says query string. Just trying to minimize the amount of thinking required to read the code.

siac/cmds.go
```diff
@@ -12,25 +12,26 @@ import (
...
-		return err.Error()
-	}
-	return string(data)
+
+	// decode response
+	obj := make(map[string]interface{})
+	// don't capture this error; we only care about HTTP errors
+	json.NewDecoder(resp.Body).Decode(&obj)
+	return obj, nil
 }
```
DavidVorick: I don't understand what this function does. Can you expand the docstring?

siac/cmds.go
```diff
@@ -47,20 +48,64 @@ func updatecmd(cmd *cobra.Command, args []string) {
...
+	case "check":
+		if m, err := getResponse("/update/check", nil); err == nil {
+			if m["Available"].(bool) {
+				fmt.Printf("Update %s is available!\nRun 'siac update %s' to install it.\n", m["Version"], m["Version"])
+			} else {
+				fmt.Println("Up to date!")
+			}
+		} else {
+			fmt.Println("Update check failed:", err)
+		}
```
DavidVorick: We haven't standardized on this, but I prefer the form:
```go
m, err = stuff
if err != nil {
    // control statement
}
// more stuff
```

The benefit being consistency. I like having `err != nil` be what always happens because then you get used to looking for it and looking at it, and because by doing `if err != nil { control }` you get to skip out on a bunch of chained `if`s for when there's lots of error checking.

Forest would definitely disagree here, he likes his flexibility, but I feel like being strict and consistent is important for having highly scalable and highly maintainable codebases.

siac/cmds.go
```diff
@@ -47,20 +48,64 @@ func updatecmd(cmd *cobra.Command, args []string) {
 		cmd.Usage()
 		return
 	}
-	fmt.Println(getResponse("/update", &url.Values{
-		"action": {args[0]},
-	}))
+	switch args[0] {
+	case "check":
+		if m, err := getResponse("/update/check", nil); err == nil {
+			if m["Available"].(bool) {
```
DavidVorick: what is m["Available"].(bool) ??

siac/cmds.go
DavidVorick: trying to open a discussion on this so that we can make a standard. I'm not certain that using control statements instead of `else` is best, but I like it this way because you usually type `return` after detecting an error, so I want to standardize around always having a control statement after detecting errors, at least where possible. And then adding documentation when things don't flow like that.

The other thing is that having a bunch of strict conventions makes it easier to document stuff. If you find a good reason to break convention, you have to document why which makes everyone's life easier, because it's something that would normally otherwise go undocumented.

It's analogous to documenting everything, because the convention serves as documentation that doesn't need to be written.

siac/cmds.go
```diff
@@ -47,20 +48,64 @@ func updatecmd(cmd *cobra.Command, args []string) {
...
+			}
+		} else {
+			fmt.Println("Update check failed:", err)
+		}
+
+	default:
+		_, err := getResponse("/update/apply", &url.Values{
+			"version": {args[0]},
+		})
+		if err == nil {
```
DavidVorick: similar deal, I think we should standardize around always writing `if err != nil {`, because occasionally you get confused (like when I switched up the responses in a recent PR).

It makes the developer less likely to make errors, and it makes the reviewer less likely to miss errors.

We were arguing a bit about trusting developers, but humans are really, really error prone machines. And with code, a single `>` vs. `>=` or `!=` vs. `==` in the wrong place can be as severe as giving root access to any application that runs certain code (I believe this happened to the linux kernel at one point). Managing code with the expectation that the developer is going to make mistakes in random places is good practice, because it's the reality of programming.

siac/cmds.go
```diff
@@ -47,20 +48,64 @@ func updatecmd(cmd *cobra.Command, args []string) {
...
+			fmt.Println("Could not stop miner:", err)
+		} else {
+			fmt.Println("Stopped mining.")
+		}
+
+	case "status":
+		m, err := getResponse("/miner/status", nil)
+		if err != nil {
+			fmt.Println("Could not get miner status:", err)
+		} else {
```
DavidVorick: just going back to the first one, I think it would be better to:

```go
if err != nil {
    fmt.Println()
    continue
}
fmt.Println()
```
as opposed to what you have because most of the time when you error check you throw in a control statement (ie `return`), so it's good to just make that a standard.

siad/api.go
```diff
@@ -37,21 +27,33 @@ func (d *daemon) handle(addr string) {
...
-	http.HandleFunc("/json/status", d.jsonStatusHandler)
+	// File API Calls
+	http.HandleFunc("/host", d.hostHandler)
+	http.HandleFunc("/rent", d.rentHandler)
+	http.HandleFunc("/download", d.downloadHandler)
+
+	// Misc. API Calls
+	http.HandleFunc("/sync", d.syncHandler)
+	http.HandleFunc("/peer/add", d.peerAddHandler)
+	http.HandleFunc("/peer/remove", d.peerRemoveHandler)
```
DavidVorick: peering will be it's own component, doesn't need to be in misc

siad/api.go
```diff
@@ -62,56 +64,31 @@ func (d *daemon) stopHandler(w http.ResponseWriter, req *http.Request) {
...
-		_, err := fmt.Sscanf(caString, "%x", &coinAddressBytes)
-		if err != nil {
-			http.Error(w, "Malformed coin address", 400)
-			return
-		}
-		copy(coinAddress[:], coinAddressBytes)
-	}
-
-	// other vars require no special parsing
 	qsVars := map[string]interface{}{
```
DavidVorick: qs again is opaque, and in this case there's nothing indicating what it might stand for.

siad/apiminer.go
```diff
@@ -17,13 +17,13 @@ func (d *daemon) minerStartHandler(w http.ResponseWriter, req *http.Request) {
...
 	d.core.StartMining()
-	fmt.Fprintf(w, "Now mining on %v threads.", threads)
+	w.WriteHeader(200)
 }
 
 // Calls StopMining() on the core.
 func (d *daemon) minerStopHandler(w http.ResponseWriter, req *http.Request) {
 	d.core.StopMining()
-	fmt.Fprintf(w, "Turning off mining.")
+	w.WriteHeader(200)
```
DavidVorick: I thought these were implicit?

siad/apiwallet.go
```diff
@@ -44,7 +43,7 @@ func (d *daemon) walletSendHandler(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
-	fmt.Fprintf(w, "Sent %v coins to %x", amount, dest)
+	w.WriteHeader(200)
```
DavidVorick: also implicit?

siad/update.go
```diff
@@ -31,50 +31,50 @@ func newerVersion(version string) bool {
 		// inputs are trusted, so no need to check the error
 		ni, _ := strconv.Atoi(nums[i])
 		Ni, _ := strconv.Atoi(NUMS[i])
-		if ni > Ni {
-			return true
+		if ni != Ni {
+			return ni > Ni
 		}
```
DavidVorick: ni?

Please try to avoid using two letter names. I have no idea what this variable means.

DavidVorick: uneasy about the `ni` names. took a while to figure out what they meant.

siad/update.go
```diff
@@ -31,50 +31,50 @@ func newerVersion(version string) bool {
 		// inputs are trusted, so no need to check the error
```
DavidVorick: wrt TODO: make this more robust

We could always pick a convention like updates always follow 3 nums: x.y.z, and then we explicitly ignore everything after the third bullet. If there's something important enough to push an update through, it's worth increasing the 'z'

DavidVorick:
I decided to merge this prematurely so Seve can use the changes.

Please review and submit new changes when you get the chance

lukechampine:
Seve could have just worked off my branch. I guess it would be a bit uglier if I made additional changes though.

siad/apiwallet.go
lukechampine: yeah, missed these two.

siad/update.go
lukechampine: I don't like them either. Probably the worst variable names I've ever used.

I think my preference stems from the fact that I tend to write shorter functions, where it's easier to infer a variable's meaning from context. This is especially true in helper functions where the arguments can be generic. e.g.:
```
func greater(first, second int) bool { return first > second }
// is this really better than:
func greater(a, b int) bool { return a > b }
```
My functions also just tend to be less variable-heavy (possible due to my preference for literals), so there are fewer things to keep track of.

siac/cmds.go
lukechampine: https://golang.org/doc/effective_go.html#interface_conversions

siac/cmds.go
lukechampine: You say these things like they're known facts, but until we actually *have* more developers it's all just speculation. Not to mention premature optimization.

Consistency is nice, and I don't think your logic is totally wrong here. But ultimately I don't think it's a terrible burden on other developers. IMHO, it's much more important that code is:
- modular (i.e. non-monolithic; split into smaller sub-functions with clear inputs+outputs)
- well-documented
- well-tested
...than whether or not it follows some particular convention.

Issue #94, opened 2014-12-02
Large transactions can fail


DavidVorick:
sometimes we'll send like 25,000,000 coins and it'll produce an invalid signature, but the wallet won't realize it and won't try again.

I think it's a problem with either one of the inputs in particular, or just a problem when you have 30+ inputs to your transaction. Needs more investigation.

DavidVorick:
Now that we have unit tests, it's pretty clear that it's unrelated to large transactions, it just happens sometimes and it seems to happen more often when your transactions are large.

lukechampine:
any theories as to why? How does the codepath for a large transaction differ from that of a small transaction?

DavidVorick:
The biggest difference is that there are multiple signatures. Three ideas:

1. There's a bug or race condition in the wallet somewhere that only triggers once in a while.
2. There's a bug in the encoding library somewhere that only triggers once in a while.
3. Either we're misusing the crypto library or that has a bug that only triggers once in a while.

I have a good way to probe this further, I'll get to it soon. Should be pretty easy to figure out where the mistake is by making thousands of small transactions.

PR #197, opened 2015-01-15
Pulling together hosting and renting

DavidVorick:
Added a test that rents and hosts a file, and verify that renting and hosting actually work.
I don't verify that the host properly submits storage proofs though, that's on the way.

Unfortunately, there are now 2-3 problems with our test suite that only come up occasionally.

1. Sometimes transaction signing fails - this is what I'll be doing next.
2. Sometimes getting a block accepted fails - This is a problem with test synchronization. State subscription should actually fix this, though there are deeper issues regarding test synchronization
3. Sometimes the wallet will spend a timelocked output. This isn't a problem until you become a host and start freezing coins. This should be an easy fix.


DavidVorick:
I can try to fit these fixes into this PR

sia/hostdir1
```diff
@@ -0,0 +1,11 @@
...
+// interface. The list:
+//    Host
+//    HostDB
+//    Miner
+//    Network
+//    (Peer Manager)
+//    Renter
+//    Wallet
+//
+package sia
```
DavidVorick: derp, how to fix?

DavidVorick: oh wait I think I know:

git rebase -i
edit

DavidVorick:
At this point, most of the major problems have been fixed. Things that pop up occasionally have tests that encourage them to pop up almost every time, and those things have been bugfixed.

hooray

DavidVorick:
A new dependency has been added, this is for a cryptography library that seems to fit our needs better. It also uses the algorithm that I wanted to use (ed25519).

DavidVorick:
Fixes #94

consensus/transactions.go
```diff
@@ -10,6 +10,7 @@ import (
 // Exported errors
 var (
 	ConflictingTransactionErr = errors.New("conflicting transaction exists in transaction pool")
+	InvalidSignatureErr       = errors.New("invalid signature in transaction")
```
lukechampine: IIRC the stdlib convention is to use `Err` as the prefix, not suffix. But I think it's more readable this way.

sia/peers.go
```diff
@@ -14,11 +14,26 @@ func (c *Core) initializeNetwork(addr string, nobootstrap bool) (err error) {
...
+		return
+	}
+	c.server.RegisterRPC("NegotiateContract", c.host.NegotiateContract)
+	if err != nil {
+		return
+	}
+	c.server.RegisterRPC("RetrieveFile", c.host.RetrieveFile)
+	if err != nil {
+		return
+	}
```
lukechampine: I still don't like this. A non-nil error here is unrecoverable, and an error doesn't help the user any more than a panic would. As I've said multiple times, the only way for a user to see such an error is if the `RegisterRPC` call has *never been run even a single time* before being shipped. 
Plus the last two errors aren't even checked.

sia/renter/rent.go
```diff
@@ -0,0 +1,159 @@
...
+		})
+		if err == nil {
+			break
+		}
+
+		fmt.Println("Problem from NegotiateContract:", err)
+		err = r.hostDB.FlagHost(host.ID)
+		if err != nil {
+			return
+		}
```
lukechampine: the new error conventions would make this a bit awkward:
```go
if err != nil {
    fmt.Println("Failed to negotiate contract:", err)
    // regardless of why negotiation failed, we should flag this host
    err = r.hostDB.FlagHost(host.ID)
    if err != nil {
        return
    }
}
```

sia/upload_test.go
```diff
@@ -0,0 +1,42 @@
...
+	err := c.renter.RentFile(components.RentFileParameters{
+		Filepath:    "sia.go",
+		Nickname:    "one",
+		TotalPieces: 1,
+	})
+	if err != nil {
+		t.Error(err)
+	}
+
+	time.Sleep(10 * time.Second)
```
lukechampine: probably overkill for 171 byte file

sia/wallet/outputs.go
```diff
@@ -112,5 +112,36 @@ func (w *Wallet) Update(diffs []consensus.OutputDiff) error {
...
+		//
+		// The for loop iterates over (height, prevHeight]
+		for i := height + 1; i <= w.prevHeight; i++ {
+			// timelockedSpendableAddresses contains a list of
+			// spendableAddresses that get unlocked at the given height. Delete
+			// each from the map of spendable addresses.
+			for _, spendableAddy := range w.timelockedSpendableAddresses[i] {
+				coinAddress := spendableAddy.spendConditions.CoinAddress()
+				delete(w.spendableAddresses, coinAddress)
+			}
```
lukechampine: I usually use `Addr` lol

sia/wallet/transactions.go
```diff
@@ -186,12 +183,12 @@ func (w *Wallet) SignTransaction(id string, wholeTransaction bool) (transaction
...
 
 	// Fetch the transaction.
-	ot, exists := w.transactions[id]
+	openTransaction, exists := w.transactions[id]
 	if !exists {
 		err = errors.New("no transaction found for given id")
 		return
 	}
-	transaction = *ot.transaction
+	transaction = *openTransaction.transaction
```
lukechampine: any reason why this is dereferenced? Just an optimization?

signatures/signatures.go
```diff
@@ -1,55 +1,45 @@
...
-	"math/big"
+
+	"github.com/agl/ed25519"
 )
 
+// One thing that worries me about this file is that the library returns a
+// bunch of pointers to data, but the types copy the data into other memory and
+// pass that around instead. This may result in side channel attacks becoming
+// possible.
+
```
lukechampine: could we switch to using pointers instead? I imagine it'd be pretty easy.

consensus/transactions.go
DavidVorick: I remember writing a bunch of ErrXYZ, I'm not sure when I started doing it this way. It is more readable but it might impact organization and alphabetization.

Except, between grep and chunking things, I'm not really convinced it needs to be a prefix.

sia/renter/rent.go
DavidVorick: *proposed error conventions, I'm not certain they're all good ideas.

But problems I spot with this code:

1. `err == nil` just worries me because that's confusing. I do like always having `err != nil`. I think there's a good chance of mistakes.

2.
```go 
err := // long stuff
        // long stuff
        /// long stuff
if err //
```
This makes me think there's a risk the error gets missed or misinterpreted or something.

So let's try:

```go
err = host.IPAddress.Call("NegotiateContract", handleNegotiateContract)
if err != nil {
    fmt.Println("problem")
    r.handleFailedNegotiation(host, err)
}
```

sia/upload_test.go
DavidVorick: I'm not sure why it takes 10 seconds but 5 wasn't working 100% of the time.

This is the wrong approach anyway and a hack. 10 seconds means there's decent motivation to actually fix it.

sia/wallet/outputs.go
DavidVorick: probably better. I think adamantine when I see addy anyway lol. Thanks Runescape

sia/wallet/transactions.go
DavidVorick: You don't want to pass around pointers to transactions, because if someone modifies the transaction you'll be in trouble.

Also this just adheres to the convention that you should only use pointers if it's aboslutely necessary, and never as an optimization.

signatures/signatures.go
DavidVorick: I almost did it last night. I think it would be pretty easy.

sia/wallet/transactions.go
lukechampine: fair enough. A comment explaining this might be a good idea then.

Issue #201, opened 2015-01-16
Miner causing all API calls to fail


seveibar:
multiple get requests to /miner/start?threads=1 and then changing threads causing all API requests to fail

seveibar:
It looks like api calls related to the miner don't work

seveibar:
Based on the output from htop, it looks like the mining is NOT happening when it holds on the http request

PR #202, opened 2015-01-16
Fix major and miner minor race condition

seveibar:


sia/miner/info.go
```diff
@@ -29,7 +29,7 @@ type MinerStatus struct {
 // Address is the current address that is receiving block payouts.
 func (m *Miner) Info() ([]byte, error) {
 	m.mu.RLock()
-	defer m.mu.RLock()
+	defer m.mu.RUnlock()
```
seveibar: we spent an hour and then we discovered this was the cause of all the pain.

\>.< we should have found this so much faster.

seveibar:
fixes #201

Issue #193, opened 2015-01-14
Web interface edit host configuration and active hosting


seveibar:
The web interface host configuration should be editable and the user should be able to active hosting.

seveibar:
I don't have the API calls to populate the host config yet

DavidVorick:
should be ready in a few hours

seveibar:
The latest API change for host allows it to be set, but I still can't retrieve config values.

Maybe instead of `/host` (which is for setting the config), make `/host/config` and `/host/setconfig` where `/host/setconfig` is equivalent to the current `/host`.

DavidVorick:
still getting there, =/ sorry it's taking a while.

Getting sidetracked by the signing bug.

seveibar:
Ideally `/host/config` would return

```javascript
{
  "HostSettings": {
    "IPAddress": ":9988",
    "TotalStorage": 0,
    "MinFilesize": 0,
    "MaxFilesize": 0,
    "MinDuration": 0,
    "MaxDuration": 0,
    "MinChallengeWindow": 0,
    "MaxChallengeWindow": 0,
    "MinTolerance": 0,
    "Price": 0,
    "Burn": 0,
    "CoinAddress": [
      0,
      0,
      "...",
      0,
      0
    ],
    "SpendConditions": {
      "TimeLock": 0,
      "NumSignatures": 0,
      "PublicKeys": null
    },
    "FreezeIndex": 0
  }
}```

DavidVorick:
fixed with #210

PR #203, opened 2015-01-16
Web Interface Sending/Receiving siacoin, host configuration bindings, general UI fixes

### LINKED ISSUE #191: Web interface send/receive siacoin
There is currently no way to send or receive siacoin from the web interface, this should probably be done through a new view (transfer-funds).

To do this, the /wallet/status request should return an array of addresses and their balances (so that people have addresses to send to).

The "Create Address" button needs to be able to create addresses for people to send to eachother and this functionality is built into the API so that portion shouldn't take much time.

seveibar:
This should add enough functionality to the web interface to send and receive SC. The big addition is the "Transfer Funds" view which allows a user to transfer SC from their account to an address. 

Changes to how views interact with incoming data, layout tweaks etc. The host configuration view now has a warning and accurate field names.

fixes #191 
helps progress on #193


DavidVorick:
you grabbed .miner.swp again

style/index.html
```diff
@@ -504,8 +459,74 @@
...
+                from
+                <div class='from'>Savings Account</div>
+                to
+                <div class='to'>0cc175bd71c7...</div>
+              </div>
+              <div class='button send-money'>
+                <i class='fa fa-send'></i>
+                Send Funds
+              </div>
+            </div>
```
DavidVorick: I would code review this if I had any idea how to read it.

Do you write all this by hand or is it generated?

seveibar: All the html and css is generated, here's the relevant line in haml...

```haml
.control
        .summary
            Transferring
            .amt 2.142 SC
            from
            .from Savings Account
            to
            .to 0cc175bd71c7...
```

In haml, everything is by default a div. In this code, the display type of the divs are changed so that they are displayed inline or side by side (this may have been a point of confusion). This makes the haml really readable and hopefully understandable. When changing the project to pure html we'd want to convert these divs to spans or another inline element because there would be no haml to "look clean".

seveibar: Oh the values in there are just mockup values- those can be removed whenever.

seveibar: the .swp problem has been resolved with a rebase/force push

sia/miner/mine.go
```diff
@@ -86,7 +86,9 @@ func (m *Miner) SolveBlock() (b consensus.Block, solved bool, err error) {
 	// to find a winnning solution.
 	for maxNonce := b.Nonce + iterations; b.Nonce != maxNonce; b.Nonce++ {
 		if b.CheckTarget(target) {
+			m.mu.RLock()
 			m.blockChan <- b
+			m.mu.RUnlock()
```
lukechampine: what is the purpose of this?

sia/miner_test.go
```diff
@@ -19,3 +19,12 @@ func mineSingleBlock(t *testing.T, c *Core) {
...
 }
+
+func testMinerDeadlocking(t *testing.T, c *Core) {
+	threads := 1
+	c.MinerInfo()
+	c.UpdateMiner(threads)
+	c.StartMining()
+	c.MinerInfo()
+	c.StopMining()
+}
```
lukechampine: this test seems superfluous if the locking issue has been fixed. It looks more like debugging code than testing code.

sia/miner/mine.go
DavidVorick: We added some tests and discovered a race condition.

When you call UpdateMiner() you can change the channel. Here we send a block down a channel without having a lock on the state, which means you could potentially send a block down a channel at the same time that the channel is being updated.

DavidVorick: Next iteration won't have channels at all for blocks, they'll just go directly to the state.

sia/miner_test.go
DavidVorick: It makes sure that all of the calls are actually doing something, and does protect/detect future locking errors. And it actually did catch the race condition above which I hadn't noticed before.

I actually want to implement something like this for every component, just slam all of the functions in a semi-random order and try to trigger a deadlock. It's a good way to add certainty to your mutex code, especially because the race library can only detect a race condition if you trigger it during runtime.

PR #206, opened 2015-01-16
Add size requirements to blocks

DavidVorick:
I also threw in a quick commit to stop the core from zeroing the host's defaults.


consensus/blocks.go
```diff
@@ -408,6 +409,13 @@ func (s *State) AcceptBlock(b Block) (rewoundBlocks []Block, appliedBlocks []Blo
 		return
 	}
 
+	// Check that the block is the correct size.
+	encodedBlock := encoding.Marshal(b)
+	if len(encodedBlock) > BlockSizeLimit {
+		err = errors.New("Block is too large, will not be accepted.")
+		return
+	}
+
```
lukechampine: I think I prefer to write `len(encoding.Marshal(b))` in code like this because it makes it clear that you aren't going to use `encodedBlock` later.

DavidVorick: that's a good point

PR #208, opened 2015-01-16
Update tweaks

lukechampine:
Using these changes, I was able to successfully update Sia from the Linode. However, there's a bug where all updated files have their executable mode bit set. I'll probably want to fix this before release.

### DIFF
--- a/siad/api.go
+++ b/siad/api.go
@@ -129,7 +129,7 @@ func (d *daemon) downloadHandler(w http.ResponseWriter, req *http.Request) {
 }
 
 func (d *daemon) updateCheckHandler(w http.ResponseWriter, req *http.Request) {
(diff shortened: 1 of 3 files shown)


siad/api.go
```diff
@@ -129,7 +129,7 @@ func (d *daemon) downloadHandler(w http.ResponseWriter, req *http.Request) {
 }
 
 func (d *daemon) updateCheckHandler(w http.ResponseWriter, req *http.Request) {
-	available, version, err := d.checkForUpdate()
+	available, version, err := checkForUpdate()
 	if err != nil {
 		http.Error(w, err.Error(), 500)
 	} else {
```
DavidVorick:
```go
if err != nil {
    http.Error
    return
}
// don't need else block anymore
```

PR #209, opened 2015-01-17
Renter improvements

DavidVorick:
The api should be usable by Seve now. There is not enough testing, instead there is finger crossing.

### DIFF
--- a/sia/renter.go
+++ b/sia/renter.go
@@ -1,0 +1,25 @@
+package sia
+
+import (
+	"github.com/NebulousLabs/Sia/sia/components"
+)
+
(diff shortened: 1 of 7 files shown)


sia/renter.go
```diff
@@ -0,0 +1,25 @@
...
+	return c.renter.RenameFile(currentName, newName)
+}
+
+func (c *Core) RentFile(params components.RentFileParameters) error {
+	return c.renter.RentFile(params)
+}
+
+func (c *Core) RentSmallFile(params components.RentSmallFileParameters) error {
+	return c.renter.RentSmallFile(params)
+}
```
lukechampine: might consider sticking all these passthrough functions in one file.

Issue #207, opened 2015-01-16
File API


seveibar:
We need a file API for non-desktop use cases (i.e. no access to file path).

Here's a proposal for the API
`/file/upload`: Uploads file to network (file sent via POST)
`/file/download`: Downloads a file, takes an address
`/file/uploaded`: Returns list of files with nicknames for downloading

lukechampine:
`/file/uploaded` will probably be useful for other purposes too.

Issue #190, opened 2015-01-14
Web interface: file upload


seveibar:
The web interface needs a file upload feature which will give a file path to the Sia client.

Because web security doesn't let sites look at the file path of dragged or selected files, we're going to have to use another method to get the file path. Some options:
- Package web interface as atom-shell or node-webkit application to get permissions
- Spawn a GUI file picker from go (no drag and drop)
- Build an in-browser file picker app (no drag and drop)

We could also just do a regular upload for small files or chunk the upload.

PR #211, opened 2015-01-17
File uploading

seveibar:
MVP-esque file uploading


seveibar:
fixes #190 
fixes #207

PR #215, opened 2015-01-17
test and fix renter download

DavidVorick:
Downloads weren't working because the renter needed the contract id.

I changed the specification to make it easier for the renter to know this without needing to know what the full transaction looked like.

There is now a test for download and a fix for the things that weren't working.

### DIFF
--- a/sia/upload_test.go
+++ b/sia/upload_test.go
@@ -1,6 +1,7 @@
 package sia
 
 import (
+	"crypto/rand"
 	"testing"
 	"time"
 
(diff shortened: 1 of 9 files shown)


sia/upload_test.go
```diff
@@ -39,4 +56,10 @@ func testUploadFile(t *testing.T, c *Core) {
 	if c.hostDB.Size() < 1 {
 		t.Fatal("Hostdb got pruned while trying to make a contract?")
 	}
+
+	// Check that the file can be downloaded.
+	err = c.renter.Download("one", "renterDownload")
+	if err != nil {
+		t.Error(err)
+	}
```
lukechampine: `os.Remove()`? I don't like having stuff like this cluttering the gitignore. If you really need to inspect it, commenting out the `Remove` isn't a big deal.

DavidVorick: Do we leave it in the make clean though?

Should we just consolidate to one directory? I feel like if you're going to inspect it at all, it should be in both the .gitignore and in the `make clean` command, even if it's normally removed automatically.

lukechampine: I don't want another `filesCreatedDuringTesting`. I guess it doesn't hurt anything to keep it in the gitignore, but I'd still like to see it removed.

PR #219, opened 2015-01-17
Host Announcement Debugging

DavidVorick:
Host announcing wasn't working in the api. There were actually a few issues that were preventing things from running smoothly.

This PR fixes all of those things, although in a hacky way that definitely needs to be revisited. The problem is really that announcing yourself as a host should be it's own call, but it's too late tonight to make that actually happen. Additionally, the way that HostUpdate currently works is completely wrong and should be like that at all. Additionally, the way the API leaves a bunch of values blank in the HostUpdate struct is also incorrect, it needs to fill in the old values instead of just leaving some blank.

But that was going to be difficult, so I moved the code elsewhere.

### DIFF
--- a/sia/host.go
+++ b/sia/host.go
@@ -1,6 +1,7 @@
 package sia
 
 import (
+	"github.com/NebulousLabs/Sia/consensus"
 	"github.com/NebulousLabs/Sia/sia/components"
 )
 
@@ -10,8 +11,22 @@ func (c *Core) HostInfo() (components.HostInfo, error) {
 
 // TODO: Make a better UpdateHost thing.
 func (c *Core) UpdateHost(announcement components.HostAnnouncement) error {
+	info, err := c.host.HostInfo()
+	if err != nil {
+		return err
+	}
+	// TODO: This stuff should not happen here, it should be managed by
+	// hostSetConfigHandler.
+	announcementUpdate := info.Announcement
+	announcementUpdate.IPAddress = c.server.Address()
+	announcementUpdate.TotalStorage = announcement.TotalStorage
+	announcementUpdate.MaxFilesize = announcement.MaxFilesize
+	announcementUpdate.MinTolerance = announcement.MinTolerance
+	announcementUpdate.Price = announcement.Price
+	announcementUpdate.Burn = announcement.Burn
+
 	update := components.HostUpdate{
-		Announcement:    announcement,
+		Announcement:    announcementUpdate,
 		Height:          c.Height(),
 		HostDir:         c.hostDir,
 		TransactionChan: c.transactionChan,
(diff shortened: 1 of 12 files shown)


sia/host.go
```diff
@@ -10,8 +11,22 @@ func (c *Core) HostInfo() (components.HostInfo, error) {
...
+	}
+	// TODO: This stuff should not happen here, it should be managed by
+	// hostSetConfigHandler.
+	announcementUpdate := info.Announcement
+	announcementUpdate.IPAddress = c.server.Address()
+	announcementUpdate.TotalStorage = announcement.TotalStorage
+	announcementUpdate.MaxFilesize = announcement.MaxFilesize
+	announcementUpdate.MinTolerance = announcement.MinTolerance
+	announcementUpdate.Price = announcement.Price
+	announcementUpdate.Burn = announcement.Burn
```
lukechampine: If I understand this correctly: in order to update the host, you need to pass in a `HostAnnouncement`. If you only want to change certain fields, you need to first obtain the old `HostAnnouncement`, modify it, then pass it back in. `UpdateHost` then *also* grabs the old `HostAnnouncement`, but it only modifies certain fields (the ones listed above). Then it sticks this into a different `HostUpdate` object, only one field of which is modified, and finally passes this object to a different `UpdateHost` function, which happens to share the same name despite taking a completely different argument.

sia/host/proofs.go
```diff
@@ -127,7 +127,7 @@ func (h *Host) createStorageProof(entry ContractEntry, heightForProof consensus.
 		err = errors.New("no record of that file")
 		return
 	}
-	fullname := h.hostDir + contractObligation.filename
+	fullname := filepath.Join(h.hostDir, contractObligation.filename)
```
lukechampine: we should standardize on using `filepath.Join` instead of `path.Join`. They are almost identical, except that `filepath.Join` uses an OS-specific separator, while `path.Join` always uses `"/"`

sia/host.go
DavidVorick: \>.< yeah

sia/host/proofs.go
DavidVorick: do we use path.Join anywhere?

DavidVorick: oh nvm

sia/host.go
DavidVorick: the problem is that for right now we're hiding certain updates from the users.

sia/host/proofs.go
lukechampine: I used it a bunch in siad, so those will need to be replaced.

sia/host.go
DavidVorick: Might be able to fix this right now, really exhausted though not sure if a great idea.

PR #223, opened 2015-01-17
fixes for file uploading

DavidVorick:
This should prevent files from not appearing if you don't have enough outputs.

This is a dirty fix for something that's actually the fault of the wallet and the transaction pool (the transaction pool needs a way to recognize transaction dependencies before the wallet can chain transactions). Again, a pretty easy fix in theory but it's going to take a few hours to get right.

Plus we're at the point where we'll be writing a great deal more tests, which will slow down new features even more.

### DIFF
--- a/sia/wallet/outputs.go
+++ b/sia/wallet/outputs.go
@@ -2,9 +2,9 @@ package wallet
 
 import (
 	"errors"
-	"fmt"
 
 	"github.com/NebulousLabs/Sia/consensus"
+	"github.com/NebulousLabs/Sia/sia/components"
 	"github.com/NebulousLabs/Sia/signatures"
 )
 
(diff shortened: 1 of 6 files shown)


DavidVorick:
Also added some code to help with mutex debugging. It doesn't help the api at this point, just provides some println insights on why things are frozen.

lukechampine:
That test looks like it will return immediately. Should probably use a `sync.WaitGroup` to prevent that.

sia/wallet/outputs.go
```diff
@@ -63,7 +63,12 @@ func (w *Wallet) findOutputs(amount consensus.Currency) (spendableOutputs []*spe
 
 	// This code will only be reached if total < amount, meaning insufficient
 	// funds.
-	err = fmt.Errorf("insufficient funds, requested %v but only have %v", amount, total)
+	//
+	// TODO: Figure out some way to return a named error that also contains
+	// custom information.
+	// err = fmt.Errorf("insufficient funds, requested %v but only have %v", amount, total)
+	err = components.LowBalanceErr
+
```
lukechampine: stdlib would do this with a new error type and a type assertion. See `os.IsExist` for an equivalent.
```go
// in components package(?)
type balanceErr struct {
    error
}

func IsBalanceErr(err error) bool {
    _, ok := err.(*balanceErr)
    return ok
}

// in findOutputs:
err = balanceErr(amount, total)

// in proposeSmallContract:
if err != nil && components.IsBalanceErr(err) {
    ... 
}
```

DavidVorick:
it's not a test, just a series of printlns

PR #240, opened 2015-01-18
Add encryption for byte slices

DavidVorick:
Feel free to go hard on the code review. This is code that I consider to be 'up to par'. Perhaps not exceptional, but it's the type of code that I would like to see everywhere.

If there's anything you don't like you should bring it up so I stop doing it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ install: fmt
 
 clean:
 	rm -rf hostdir release whitepaper.aux whitepaper.log whitepaper.pdf         \
-	sia.wallet sia/test.wallet sia/hostdir* sia/renterDownload
+		sia.wallet sia/test.wallet sia/hostdir* sia/renterDownload
 
 test: clean install
 	go test -short ./...
@@ -34,7 +34,7 @@ dependencies:
 xc:
 	go get -u github.com/laher/goxc
 	goxc -arch="amd64" -bc="linux windows darwin" -d=release -pv=0.1.0          \
-	-br=developer -pr=beta -include=style/,example-config,LICENSE*,README*      \
-	-tasks-=deb,deb-dev,deb-source -build-tags=release
+		-br=developer -pr=beta -include=style/,example-config,LICENSE*,README*  \
+		-tasks-=deb,deb-dev,deb-source -build-tags=release
 
 .PHONY: all fmt install test test-long test-long-race whitepaper dependencies distribution clean xc
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,7 @@ dependencies:
 	go get -u github.com/spf13/cobra
 	go get -u github.com/inconshreveable/go-update
 	go get -u github.com/agl/ed25519
+	go get -u golang.org/x/crypto/twofish
 
 # Cross Compile - makes binaries for windows, linux, and mac, 32 and 64 bit.
 xc:
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,8 @@ install: fmt
 	go install ./...
 
 clean:
-	rm -rf hostdir release whitepaper.aux whitepaper.log whitepaper.pdf sia.wallet sia/test.wallet sia/hostdir* sia/renterDownload
+	rm -rf hostdir release whitepaper.aux whitepaper.log whitepaper.pdf         \
+	sia.wallet sia/test.wallet sia/hostdir* sia/renterDownload
 
 test: clean install
 	go test -short ./...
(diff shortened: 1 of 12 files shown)


DavidVorick:
I did just enforce an 80 character line limit in the Makefile. It's pretty easy to remove if there are objections.

Makefile
```diff
@@ -7,7 +7,8 @@ install: fmt
 	go install ./...
 
 clean:
-	rm -rf hostdir release whitepaper.aux whitepaper.log whitepaper.pdf sia.wallet sia/test.wallet sia/hostdir* sia/renterDownload
+	rm -rf hostdir release whitepaper.aux whitepaper.log whitepaper.pdf         \
+	sia.wallet sia/test.wallet sia/hostdir* sia/renterDownload
```
lukechampine: subsequent lines after `\` should be indented (unless that breaks the Makefile, which isn't out of the question)

crypto/encrypt.go
```diff
@@ -0,0 +1,101 @@
...
+	// Create the padded + unencrypted ciphertext.
+	ciphertext = make([]byte, len(plaintext)+padding)
+	copy(ciphertext, plaintext)
+
+	// Create the iv.
+	iv = make([]byte, BlockSize)
+	_, err = rand.Read(iv)
+	if err != nil {
+		return
+	}
```
lukechampine: what is `iv`?

crypto/encrypt.go
```diff
@@ -0,0 +1,101 @@
...
+
+// EncryptBytes encrypts a []byte using a key. The padded ciphertext, iv, and
+// amount of padding used are returned. `plaintext` is not overwritten.
+func EncryptBytes(key EncryptionKey, plaintext []byte) (ciphertext []byte, iv []byte, padding int, err error) {
+	// Determine the length needed for padding. The ciphertext must be padded
+	// to a multiple of BlockSize.
+	padding = BlockSize - (len(plaintext) % BlockSize)
+	if padding == BlockSize {
+		padding = 0
+	}
```
lukechampine: this took a second for me to parse. I might have written it:
```go
padding := 0
if rem := len(plaintext) % BlockSize; rem != 0 {
    padding = BlockSize - rem
}
```
but neither way is clearly better than the other.

crypto/encrypt.go
```diff
@@ -0,0 +1,101 @@
...
+		return
+	}
+
+	// Encrypt the ciphertext.
+	blockCipher, err := twofish.NewCipher(key[:])
+	if err != nil {
+		return
+	}
+	encrypter := cipher.NewCBCEncrypter(blockCipher, iv)
+	encrypter.CryptBlocks(ciphertext, ciphertext)
```
lukechampine: I would chain these together, since `encrypter` isn't used elsewhere, but I acknowledge it isn't your style.

crypto/encrypt.go
```diff
@@ -0,0 +1,101 @@
...
+		return
+	}
+	if len(ciphertext)%BlockSize != 0 {
+		err = errors.New("ciphertext is not correct size")
+		return
+	}
+	if padding > len(ciphertext) {
+		err = errors.New("stated padding is longer than the ciphertext")
+		return
+	}
```
lukechampine: probably want to add a check for negative padding too.
```go
if padding < 0 || padding > len(ciphertext) {
    err = errors.New("invalid padding")
    return
}
```

crypto/encrypt.go
```diff
@@ -0,0 +1,101 @@
...
+	// Decrypt the ciphertext.
+	blockCipher, err := twofish.NewCipher(key[:])
+	if err != nil {
+		return
+	}
+	decrypter := cipher.NewCBCDecrypter(blockCipher, iv)
+	decrypter.CryptBlocks(plaintext, ciphertext)
+
+	// Remove the padding.
+	plaintext = plaintext[:len(ciphertext)-padding]
```
lukechampine: should this be `len(ciphertext)`?

crypto/encrypt.go
DavidVorick: Initialzation Vector. Very ubiquitous term when working with ciphers.

crypto/encrypt.go
DavidVorick: The amount of padding is returned, so at some point you need to calculate the amount of padding. In you example, you'd need to add another line `padding = cipherBlocks * BlockSize - len(plaintext)`

crypto/encrypt.go
DavidVorick: It weirds me out that CryptBlocks has no return values.

`cipher.NewCBCEncrypter(blockCipher, iv).CryptBlocks(ciphertext, ciphertext)`

It just doesn't seem clean enough. I did that a few times in the consensus code, but I don't like any of those either. Plus this way it's pretty far over 80 chars, which is also messy.

The example code in the cipher documentation wrote it out like this.

crypto/encrypt.go
DavidVorick: I don't like the `!( X && Y)`, because you have to worry about both the `!` and a double conditional. I actually didn't see the `!` the first time around.

Makefile
DavidVorick: I believe that after you start a line, all whitespace will be ignored. So I'll fix this.

crypto/encrypt.go
DavidVorick: I do wish there was a way to do it without any `if` statements. That would make me happy.

lukechampine: could you just not return the padding? See comment below.

crypto/encrypt.go
DavidVorick: hmmm. We could pass around a length instead of a padding and that might clean a few things up. It's not generally how I reason about slices though, I typically think in terms of:

[thing one][thing two][thing three]. So to me having a padding makes sense because it becomes [thing three]. Otherwise you think about [thing one] and [thing two], but the slice is longer than each and you implicitly figure out that there's padding at the end which composes the extra stuff. 

I think it's less clear that there's padding if you're passing around a length that doesn't match the size of the returned data.

PR #248, opened 2015-01-19
API fixes + working siac

lukechampine:
siac has been totally overhauled and now implements the full API. Previous it was broken in many small but crucial ways, rendering it largely unusable. I feel like we shouldn't release a broken siac to beta users, even if few of them wind up using it.


lukechampine:
Build failure is caused by non-deterministic data race. I was able to get the build to pass on my machine

siac/hostcmd.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+	fmt.Sprintf(`Host settings:
+Storage:      %v bytes (%v remaining)
+Price:        %v coins
+Max Filesize: %v
+Max Duration: %v
+Tolerance:    %v
+Burn:         %v
+`, config.Announcement.TotalStorage, config.StorageRemaining, config.Announcement.Price, config.Announcement.MaxFilesize,
+		config.Announcement.MaxDuration, config.Announcement.MinTolerance, config.Announcement.Burn)
```
DavidVorick: blegh. Is there a way to space this better? Maybe have it match the outputs?

```go
`Host settings:
Storage: %v bytes (%v remaining)
Price: %v coins
Max Filesize: %v
Max Duration: %v
Tolerance: %v
Burn: %v
`,
totalstorage, storageremaining,
price,
max,
...
```

siac/hostcmd.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+// TODO: settings should be updated individually, then submitted together in a
+// separate API call.
+func hostsetconfigcmd(totalstorage, maxfilesize, mintolerance, maxduration, price, burn string) {
+	err := callAPI(fmt.Sprintf("/host/setconfig?totalstorage=%s&maxfilesize=%s&mintolerance=%s"+
```
DavidVorick: also a bit blegh

siac/main.go
```diff
@@ -1,101 +1,117 @@
...
+// as arguments.
+func wrap(fn interface{}) func(*cobra.Command, []string) {
+	fnVal, fnType := reflect.ValueOf(fn), reflect.TypeOf(fn)
+	if fnType.Kind() != reflect.Func {
+		panic("wrapped function has wrong type signature")
+	}
+	for i := 0; i < fnType.NumIn(); i++ {
+		if fnType.In(i).Kind() != reflect.String {
+			panic("wrapped function has wrong type signature")
+		}
```
DavidVorick: still uneasy about the panics. A panic isn't terribly user friendly. Could we print something and then log a stack trace?

DavidVorick: fine for now just want to find something friendlier than panic for production code.

siac/main.go
```diff
@@ -1,101 +1,117 @@
...
 
 func main() {
 	root := &cobra.Command{
 		Use:   os.Args[0],
-		Short: "Sia Client v0.1.0",
-		Long:  "Sia Client v0.1.0",
-		Run:   versioncmd,
+		Short: "Sia Client v" + VERSION,
+		Long:  "Sia Client v" + VERSION,
+		Run:   wrap(func() { println("Sia Client v" + VERSION) }),
```
DavidVorick: I'm not sure that I view this abstraction as a good thing. It's pretty intense in terms of complexity.

siac/main.go
```diff
@@ -1,101 +1,117 @@
...
+	peerCmd.AddCommand(peerAddCmd)
+	peerCmd.AddCommand(peerRemoveCmd)
+	peerCmd.AddCommand(peerStatusCmd)
+
+	root.AddCommand(updateCmd)
+	updateCmd.AddCommand(updateCheckCmd)
+	updateCmd.AddCommand(updateApplyCmd)
+	root.AddCommand(statusCmd)
+	root.AddCommand(stopCmd)
+	root.AddCommand(syncCmd)
```
DavidVorick: hmm. this is pretty.

siac/main.go
DavidVorick: the docstring is pretty good though

siac/misccmd.go
```diff
@@ -0,0 +1,132 @@
...
+
+	syncCmd = &cobra.Command{
+		Use:   "sync",
+		Short: "Synchronize with the network",
+		Long:  "Attempt to synchronize with a randomly selected peer.",
+		Run:   wrap(synccmd),
+	}
+)
+
+// TODO: this should be defined outside of siac
```
DavidVorick: it's not already? I'm curious that you would need it for siac and not for siad

siac/main.go
lukechampine: it doesn't need to be user friendly because the user will never see it.

siac/main.go
lukechampine: this used to be:
```go
func versioncmd(cmd *cobra.Command, args []string) {
    println("Sia Client v" + VERSION)
}
//...
root := &cobra.Command{
    //...
    Run: versioncmd
}
```
Neither approach is all that much better than the other. Now that I review it, the one above seems more readable though. I'll switch it back.

siac/misccmd.go
lukechampine: siad doesn't define it, instead just using a struct literal. Later on I plan to consolidate+export some of the API stuff to make API code cleaner in both siad and siac.

siad/apimisc.go
```diff
@@ -0,0 +1,50 @@
...
+	// TODO: don't spawn multiple CatchUps
+	if len(d.core.AddressBook()) == 0 {
+		http.Error(w, "No peers available for syncing", 500)
+		return
+	}
+
+	go d.core.CatchUp(d.core.RandomPeer())
+
+	writeSuccess(w)
+}
```
DavidVorick: I'm still of the opinion that this belongs in server/peers. Syncing is about having the same information as your peers. Just like transaction broadcasting and block broadcasting is in the server/peers stuff, syncing should also be there. The job of the server/peers is to make sure everyone is on the same page, and is getting to the same page efficiently.

In my head anyway.

siac/filecmd.go
```diff
@@ -0,0 +1,70 @@
...
+	fileStatusCmd = &cobra.Command{
+		Use:   "file status",
+		Short: "View a list of uploaded files",
+		Long:  "View a list of files that have been uploaded to the network.",
+		Run:   wrap(filestatuscmd),
+	}
+)
+
+func fileuploadcmd(file, nickname, pieces string) {
+	err := callAPI(fmt.Sprintf("/file/upload?file=%s&nickname=%s&pieces=%s", file, nickname, pieces))
```
DavidVorick: I think this much nesting is pretty much the maximum amount of nesting that should happen.

siad/apipeer.go
```diff
@@ -0,0 +1,31 @@
+package main
+
+import (
+	"net/http"
+
+	"github.com/NebulousLabs/Sia/network"
+)
+
+func (d *daemon) peerAddHandler(w http.ResponseWriter, req *http.Request) {
+	err := d.core.AddPeer(network.Address(req.FormValue("addr")))
```
DavidVorick: too much.

Daemon.Core.AddPeer(network.Address(req.FormValue("addr")))

That's two dereferences into a typecast into a function call.

```go
addr := network.Address(req.FormValue("addr"))
err := d.core.AddPeer(addr)
```

siac/main.go
DavidVorick: =/

PR #250, opened 2015-01-19
google analytics for website

seveibar:


seveibar:
Haha indeed I have, it looks like the google analytics snippet is throwing an error on every page

PR #261, opened 2015-01-19
Consensus Tweaks

DavidVorick:
Just a few changes, but felt like making a PR because it's the type of thing I'm going to keep doing.

Mostly, I'm just trying to keep the line count on my PRs small.

### DIFF
--- a/Consensus.md
+++ b/Consensus.md
@@ -1,0 +1,9 @@
+This file documents the rules of consensus in the Sia protocol.
+
+---------------------
+--Block Timestamps --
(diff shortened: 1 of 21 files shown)


DavidVorick:
*david tried to make a small PR*
It's not very effective...

DavidVorick:
Mutex stuff should be fixed now. No more race conditions, no more deadlock.

... I think... but I'm also pretty sure the testing is comprehensive enough right now to catch most of the problems.

PR #270, opened 2015-01-20
Graceful shutdown

lukechampine:
siad no longer terminates via `os.Exit`, but simply by returning. It accomplishes this by waiting on a channel. It can also be killed via interrupt, and will attempt to shutdown cleanly. I'd like it to save the wallet file before closing; however, the Wallet interface no longer has a `Save` method, so right now there's no way to do it cleanly. (It can be done "uncleanly," via type assertion, but I opted not to do so.)

### DIFF
--- a/sia/core.go
+++ b/sia/core.go
@@ -205,6 +205,7 @@ func (c *Core) ScanMutexes() {
 
 // Close does any finishing maintenence before the environment can be garbage
 // collected. Right now that just means closing the server.
+// TODO: save wallet
 func (c *Core) Close() {
 	c.server.Close()
 }
--- a/siad/apimisc.go
+++ b/siad/apimisc.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"net/http"
-	"os"
 )
 
 func (d *daemon) updateCheckHandler(w http.ResponseWriter, req *http.Request) {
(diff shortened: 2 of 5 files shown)


seveibar:
Awesome!

sia/core.go
```diff
@@ -205,6 +205,7 @@ func (c *Core) ScanMutexes() {
 
 // Close does any finishing maintenence before the environment can be garbage
 // collected. Right now that just means closing the server.
+// TODO: save wallet
```
DavidVorick: shouldn't be needed

siad/apimisc.go
```diff
@@ -33,9 +32,10 @@ func (d *daemon) statusHandler(w http.ResponseWriter, req *http.Request) {
 }
 
 func (d *daemon) stopHandler(w http.ResponseWriter, req *http.Request) {
-	// TODO: more graceful shutdown?
 	d.core.Close()
-	os.Exit(0)
+	writeSuccess(w)
+	// send stop signal
+	d.stop <- struct{}{}
```
DavidVorick: this only goes to handle() right?

sia/core.go
lukechampine: ah. In that case I think that the "Write & Copy" approach is good enough. Unless handled carefully, triggering a save right before exit might even lead to more frequent corruption, not less.

PR #271, opened 2015-01-20
Miner

DavidVorick:
Added a convention for threaded functions.

Added a purely notification based subscription to the state, I'm going to deprecate the other kind. Except that my plan for doing atomic reads on the state has failed, I'm not sure what the problem is. I'm really confused by the deadlock that's happening and would like you to review it and see if you have any ideas.

sia/miner.go got removed, the daemon doesn't need passthrough functions anymore because it has the miner directly. The core now only talks to the miner so that it can pass block updates on to other modules. This will also become unneeded after I finish my series of changes, and we will probably be able to drop the core altogether.

Right now there's a decent amount of legacy code. It'll all get removed soon enough.


consensus/notifications.go
```diff
@@ -71,11 +74,22 @@ func (s *State) notifySubscribers(cc ConsensusChange) {
...
 		sub <- cc
 	}
+
+	for _, sub := range s.subscriptions {
+		select {
+		case sub <- struct{}{}:
+			continue
+		default:
+			continue
+		}
```
DavidVorick: My understanding is that if the channel is full, there won't be any blocking.

consensus/state.go
```diff
@@ -361,6 +366,16 @@ func (s *State) stateHash() hash.Hash {
...
 
+// RLock is an exported function of the state, allowing modules to lock the
+// state to make multiple atmoic reads. Lock() is not exported, and is not
+// meant to be used by modules.
+func (s *State) RLock() {
+	s.mu.RLock()
+}
+func (s *State) RUnlock() {
+	s.mu.RUnlock()
+}
```
DavidVorick: These are the functions that seem to be causing the deadlock.

sia/miner/mine.go
```diff
@@ -54,7 +54,14 @@ func (m *Miner) mine() {
...
 		if desiredThreads >= myThread {
-			m.SolveBlock()
+			// Grab the necessary variables for mining, and then attempt to
+			// mine a block.
+			m.mu.RLock()
+			bfw := m.blockForWork()
+			target := m.target
+			iterations := m.iterationsPerAttempt
+			m.mu.RUnlock()
+			m.solveBlock(bfw, target, iterations)
```
DavidVorick: This is kind of gross, but it follows conventions to do it this way. Otherwise, solveBlock needs to grab mutexes, which violates convention.

There really wasn't a pretty way to do this. Channels might clean things up, but I'm not really worried about it.

sia/miner/mine.go
```diff
@@ -69,27 +76,40 @@ func (m *Miner) mine() {
...
+			// Unfortunately, that throws off all of the other modules right
+			// now.
+			/*
+				// TODO: If debug, check the error value of AcceptBlock and panic
+				// for err != nil.
+				m.state.AcceptBlock(b)
+			*/
 			m.mu.RLock()
 			m.blockChan <- b
 			m.mu.RUnlock()
```
DavidVorick: This code is deprecated. It'll just be one line `m.state.AcceptBlock(b)` once other modules no longer depend on the core for updates.

sia/miner/miner.go
```diff
@@ -8,54 +8,124 @@ import (
...
+// Until the miner has an atomic way to grab the 4 values, this is a race
+// condition, but not one that the race library will be able to detect.
+func (m *Miner) update() {
+	// m.state.RLock()
+	m.parent = m.state.CurrentBlock().ID()
+	m.transactions = m.state.TransactionPoolDump()
+	m.target = m.state.CurrentTarget()
+	m.earliestTimestamp = m.state.EarliestTimestamp()
+	// m.state.RUnlock()
+}
```
DavidVorick: This is where I'm getting the really strange deadlock.

DavidVorick:
I updated the daemon api to talk to the miner directly, instead of going through the core.

DavidVorick:
the build fail is nondeterministic, and is a result of the components grabbing updates from different places.

=/ the pain will end soon.

consensus/notifications.go
lukechampine: correct. The `continue`s are also unnecessary. I would replace them with comments:
```go
select {
case sub <- struct{}{}:
    // all good
default:
    // channel is full; skip send.
    // This is not a problem, because the channel is merely used to
    // notify the receiver of updates. If the channel is full, it means the
    // receiver has already been informed that the state has changed.
}

File: sia/miner/miner.go [hunk 6]
```diff
@@ -8,54 +8,129 @@ import (
...
+// then call update() upon receiving one.
+func (m *Miner) threadedListen() {
+	for {
+		select {
+		case _ = <-m.stateSubscription:
+			m.mu.Lock()
+			m.update()
+			m.mu.Unlock()
+		}
+	}
```
lukechampine: `for _ = range m.stateSubscription {`

PR #272, opened 2015-01-20
HTTP API Test Suite

seveibar:
This is the start of the test suite for the HTTP API.

At the moment, it doesn't seem to be possible to run multiple tests because I can't start/stop the daemon multiple times.

I think to expand on this test suite we're going to want to enable the daemon to be started/stopped and enable multiple daemons to test communication, uploading and downloading etc.

### DIFF
--- a/siad/api_test.go
+++ b/siad/api_test.go
@@ -132,10 +132,17 @@ func reqHostSetConfig(t *testing.T, hostInfo components.HostInfo) SuccessRespons
 	return fResponse
 }
 
+func reqMinerStart(t *testing.T) SuccessResponse {
+	return reqSuccess(t, "/miner/start")
+}
+
+func reqMinerStop(t *testing.T) SuccessResponse {
+	return reqSuccess(t, "/miner/stop")
+}
+
 // /update/check
 
 // /host/setconfig
-// /miner/start
 // /miner/stop
 // /wallet/send
 // /file/upload
--- a/siad/api_test.go
+++ b/siad/api_test.go
@@ -70,7 +70,7 @@ func reqSuccess(t *testing.T, url string) SuccessResponse {
 
 func reqWalletStatus(t *testing.T) components.WalletInfo {
 	var r components.WalletInfo
-	reqJSON(t, "/wallet/status", r)
+	reqJSON(t, "/wallet/status", &r)
 	return r
 }
 
--- a/siad/api_test.go
+++ b/siad/api_test.go
@@ -2,9 +2,11 @@ package main
 
 import (
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"math/big"
 	"net/http"
+	"net/url"
 	"path/filepath"
 	"strconv"
 	"testing"
@@ -20,6 +22,8 @@ type SuccessResponse struct {
 	Success bool
 }
 
+// httpReq will request a byte stream from the provided url and log and return
+// and errors
 func httpReq(t *testing.T, url string) ([]byte, error) {
 	resp, err := http.Get("http://127.0.0.1:9980" + url)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


siad/api_test.go
```diff
@@ -0,0 +1,340 @@
...
+
+func setupDaemon(t *testing.T) {
+
+	// Settings to speed up operations
+	consensus.BlockFrequency = consensus.Timestamp(1)
+	consensus.TargetWindow = consensus.BlockHeight(1000)
+	network.BootstrapPeers = []network.Address{"localhost:9988"}
+	consensus.RootTarget[0] = 255
+	consensus.MaxAdjustmentUp = big.NewRat(1005, 1000)
+	consensus.MaxAdjustmentDown = big.NewRat(995, 1000)
```
DavidVorick: Write a TODO here, this should all happen automatically when you use the makefile. It's a TODO for me, but it should be here nonetheless.

siad/api_test.go
```diff
@@ -0,0 +1,340 @@
...
+
+	go func() {
+		err = startDaemon(config)
+		if err != nil {
+			t.Fatal("Couldn't start daemon: " + err.Error())
+		}
+	}()
+
+	// Give the daemon time to initialize
+	time.Sleep(10 * time.Millisecond)
```
DavidVorick: write a TODO here as well, to change it to signal-based instead of sleep based.

siad/api_test.go
```diff
@@ -0,0 +1,340 @@
...
+	mInfo := reqMinerStatus(t)
+	if mInfo.State != "Off" {
+		t.Fatal("Miner initial state was not off: (" + strconv.Itoa(mInfo.Threads) + ")")
+	}
+	if mInfo.RunningThreads != 0 {
+		t.Fatal("Miner initial RunningThreads was not 0: (" + strconv.Itoa(mInfo.Threads) + ")")
+	}
+	if mInfo.Threads != 1 {
+		t.Fatal("Miner initial Threads was not 1 (" + strconv.Itoa(mInfo.Threads) + ")")
+	}
```
DavidVorick: I'm not sure how necessary these tests are, because it's not necessarily behavior that we want to enforce. Starting in the "Off" state is good, and running threads is probably okay too. So I guess it's just the last one.

siad/api_test.go
```diff
@@ -0,0 +1,340 @@
...
+	if hConfig.Announcement.TotalStorage > 0 {
+		t.Fatal("Host started with more than 0 TotalStorage!")
+	}
+
+	// TODO: we need some way of guranteeing that the daemon will stop so that
+	// other tests can be run
+}
+
+// TODO: this test can't be run because we have no way of shutting off the other
+// daemon ATM
```
DavidVorick: Make one daemon at the beginning and pass it around a bunch.

DavidVorick: see sia/core_test.go for an example

DavidVorick:
I was thinking, the API tests in golang are kind of misplaced. They help in the sense  that they enforce a specific set of behaviors, which is good if you've settled on a set of behaviors to support. But while we're still changing things, what you really care about is compatibility with the tools that will be using the api.

So really, there should probably be a layer between the frontend and the api that manages all of the calls, and then the tests should act on that layer. That way, if something in the api changes, the tests will fail exactly until the frontend layer is changed to restore compatibility, and you don't ever need to modify the tests.

Right now, you'd need to modify both the tests and the frontend to restore compatibility after making a change to the api.

All of that may be too much for now. It's just something I was thinking about.

seveibar:
That approach definitely has a lot of advantages e.g. developers could quickly make modifications to the API then worry about adjusting it to the interface later. But I question the usefulness of that, because IMO the API is one of the first things to get static (if you know what features you're going to provide, then the API is relatively straightforward to nail down).

We would also want to be careful as to how this "layer" is created. I think the high-level API tests are the only realistic way to simulate real-world scenarios and don't just serve to test the API. So the layer would need a way to start multiple daemons, adjust ports etc. The only point to me saying this is we can't do it with JS haha- it should probably be in the go code (although maybe not in the test suite).

I'm also not sure it's a bad thing if when you make changes to the API the tests fail- it could serve as a reminder to update documentation and tests.

siad/api_test.go
DavidVorick: ah, TODO no more :). It's done

DavidVorick:
I'm probably going to pull these commits and fix this PR up so it matches how things have changed, b/c I think we're ready for API tests. I should be able to re-use most of what is here. You'll get to keep your commits I believe.

seveibar:
:DD

DavidVorick:
I suppose this was inevitable but we've replaced your api testing with some of our own. It just wasn't meant to be.

PR #275, opened 2015-01-20
graceful http shutdown

lukechampine:
Apologies to Seve (again).

Only bummer here is that our ability to handle "/stop" calls versus interrupts is a bit limited. But I wasn't using that for anything anyway, except to print "Caught deadly signal" in the event of an interrupt.

### DIFF
--- a/siad/api.go
+++ b/siad/api.go
@@ -55,14 +55,14 @@ func (d *daemon) handle(addr string) {
 	mux.HandleFunc("/mutextest", d.mutexTestHandler)
 
 	// create graceful HTTP server
-	d.srv = &graceful.Server{
+	d.apiServer = &graceful.Server{
 		Timeout: apiTimeout,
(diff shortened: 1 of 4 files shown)


siad/api.go
```diff
@@ -3,53 +3,66 @@ package main
...
+	mux.HandleFunc("/update/apply", d.updateApplyHandler)
+	mux.HandleFunc("/status", d.statusHandler)
+	mux.HandleFunc("/stop", d.stopHandler)
+	mux.HandleFunc("/sync", d.syncHandler)
 	// For debugging purposes only
-	http.HandleFunc("/mutextest", d.mutexTestHandler)
+	mux.HandleFunc("/mutextest", d.mutexTestHandler)
+
+	// create graceful HTTP server
+	d.srv = &graceful.Server{
```
DavidVorick: pls different name

PR #276, opened 2015-01-20
File upload view improvements and "No updates available" notification

seveibar:


Makefile
```diff
@@ -1,4 +1,4 @@
-all: install
+all: install
```
lukechampine: ಠ_ಠ

style/index.html
```diff
@@ -476,10 +476,28 @@
...
+                </div>
+              </div>
+              <div class='step2 step'>
+                <div class='step-text'>Give it a name</div>
+                <input class='description-field' placeholder='File Nickname'>
+              </div>
+              <div class='step3 step'>
+                <div class='step-text'>Upload to Sia</div>
+                <div class='button upload'>Upload</div>
+              </div>
```
DavidVorick: It will be nice once the generated html is gone. I'm also increasingly in support of breaking the frontend into its own repo, though we'll need a good way to keep the projects integrated.

Makefile
seveibar: it's better this way ;)

style/index.html
lukechampine: I'm now of the opinion that it should be a submodule, tracking a release branch which contains the generated HTML.

Makefile
DavidVorick: pls fix

seveibar:
Don't bother merging this- I'll move it to the new branch

Issue #283, opened 2015-01-22
CatchUp fails


lukechampine:
CatchUp seems to have some base failure rate, so the more times you call it, the greater the chance of failure. It also seems that once you stop catching up, no further attempts to catch up will succeed. This becomes a serious problem when the blockchain grows larger than a few thousand blocks.

lukechampine:
`CatchUp` calls itself in a loop, but it recurses before the state has finished processing all the blocks from the previous `CatchUp`. This causes `knownBlocks` to reflect an outdated view of the state, and consequently, `SendBlocks` will send blocks that the receiver has already received, but hasn't yet processed.

This explains why I was seeing so many `KnownBlockErr`s, but I'm not sure that it explains why `CatchUp` is failing.

lukechampine:
okay, I think I've traced it. The code is freezing during `AcceptBlock`. This would indicate a deadlock somewhere in the state. A deadlock would explain why subsequent calls to `CatchUp` failed as well.

DavidVorick:
I imagine it will. I think the problem originated from ConsensusSubscribe()

PR #285, opened 2015-01-22
Catch Up fixes

lukechampine:
`CatchUp` should be more reliable now. It still deadlocks sometimes (see #283) but at least you don't get 50% duplicate blocks anymore.

### DIFF
--- a/sia/synchronize.go
+++ b/sia/synchronize.go
@@ -2,6 +2,7 @@ package sia
 
 import (
 	"errors"
+	"time"
 
 	"github.com/NebulousLabs/Sia/consensus"
 	"github.com/NebulousLabs/Sia/network"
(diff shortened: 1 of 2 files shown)


sia/synchronize.go
```diff
@@ -100,6 +108,9 @@ func (c *Core) CatchUp(peer network.Address) {
 	// recursively. Furthermore, if there is a reorg that's greater than 100
 	// blocks, CatchUp is going to fail outright.
 	if err != nil && err.Error() == moreBlocksErr.Error() {
+		// sleep long enough for state to accept all blocks
+		// TODO: this needs to be replaced by a more deterministic wait.
+		time.Sleep(time.Second)
```
lukechampine: This should become unnecessary once `AcceptBlock` becomes a blocking call.

PR #286, opened 2015-01-22
Massive Refactor

DavidVorick:
Right now there's only minimal testing in place, but I wanted to give you a chance to start looking through the refactored code. There's a good deal left to do, like refactor renter and host, but I did my best to make it so at least the wallet and the miner were working.

The state code has been cleaned up quite a bit, a lot of data structures removed or refactored. The core has been made essentially useless though I'm going to need some of the tests so I've kept some of the files around temporarily.

More to come.


consensus/blocks.go
```diff
@@ -126,36 +123,9 @@ func (s *State) validHeader(b Block) (err error) {
...
-	s.blockMap[b.ID()] = newNode
-	parentNode.Children = append(parentNode.Children, newNode)
-
-	return
-}
-
 // State.AcceptBlock() will add blocks to the state, forking the blockchain if
 // they are on a fork that is heavier than the current fork.
-func (s *State) AcceptBlock(b Block) (rewoundBlocks []Block, appliedBlocks []Block, outputDiffs []OutputDiff, err error) {
+func (s *State) AcceptBlock(b Block) (err error) {
```
lukechampine: now that is satisfying

consensus/blocks.go
```diff
@@ -163,6 +133,8 @@ func (s *State) AcceptBlock(b Block) (rewoundBlocks []Block, appliedBlocks []Blo
 	_, exists := s.badBlocks[b.ID()]
 	if exists {
 		err = errors.New("block is known to be invalid")
+		errStr := fmt.Sprintf("%v", b.ID())
+		panic(errStr)
```
lukechampine: not much point to the `err` if you're going to panic...I'm guessing the panic won't stick around forever.

consensus/contracts.go
```diff
@@ -36,274 +36,152 @@ func (s *State) storageProofSegmentIndex(contractID ContractID, windowIndex Bloc
...
+// is invalid.
+func (s *State) validContract(fc FileContract) (err error) {
+	if fc.Start <= s.height() {
+		err = errors.New("contract must start in the future.")
+		return
+	}
+	if fc.End <= fc.Start {
+		err = errors.New("contract duration must be at least one block.")
+		return
+	}
```
lukechampine: this convention is a bit grating.

consensus/contracts.go
```diff
@@ -36,274 +36,152 @@ func (s *State) storageProofSegmentIndex(contractID ContractID, windowIndex Bloc
...
+		Value:     contract.Payout,
+		SpendHash: contract.ValidProofAddress,
 	}
-	output, err := s.output(outputID)
+	outputID, err := contract.StorageProofOutputID(sp.ContractID, s.height(), true)
 	if err != nil {
-		panic(err)
+		if DEBUG {
+			panic(err) // hard to avoid
+		}
```
lukechampine: what sort of conditions would trigger this panic?

consensus/blocks.go
DavidVorick: was there for debugging, meant to remove it

consensus/contracts.go
DavidVorick: we can do `return errors.New`, it's just not what first comes to mind.

consensus/contracts.go
DavidVorick: apparently none. StorageProofOutputID used to be a more complex function that had error conditions.

Now it seems that the error is unneeded, thanks to some simplifications.

DavidVorick:
77 files changed is definitely my record.  4000 line diff probably is too.

consensus/contracts.go
lukechampine: I was thinking more along the lines of:
```go
switch {
case fc.Start <= s.height():
    return errors.New("...")
case fc.End <= fc.Start:
    return errors.New("...")
// ... more cases
// ...
default:
    return nil
}
```
This way is fine though.

consensus/contracts.go
```diff
@@ -36,274 +36,144 @@ func (s *State) storageProofSegmentIndex(contractID ContractID, windowIndex Bloc
...
-	for _, missedProof := range s.currentBlockNode().MissedStorageProofs {
-		cid, oid := missedProof.ContractID, missedProof.OutputID
-		s.openContracts[cid].FundsRemaining += s.unspentOutputs[oid].Value
-		s.openContracts[cid].Failures -= 1
-		diff := OutputDiff{New: false, ID: oid, Output: s.unspentOutputs[oid]}
-		delete(s.unspentOutputs, oid)
-		diffs = append(diffs, diff)
+	// Delete all of the contracts that terminated.
+	for _, id := range expiredContracts {
+		contract := s.openContracts[id]
```
lukechampine: if you're just going to lookup the contract again, why not make `expiredContracts` a `[]*OpenContract` instead of a `ContractID`?

DavidVorick: s.applyMissedProof deletes a contract from the map, but we are iterating through the map. So it's gotta be 2 loops

consensus/diffs.go
```diff
@@ -0,0 +1,88 @@
...
+		if DEBUG {
+			_, exists := s.unspentOutputs[od.ID]
+			if !exists {
+				panic("rogue non-new output in applyOutputDiff")
+			}
+		}
+
+		delete(s.unspentOutputs, od.ID)
+	}
+}
```
lukechampine: Muh brevity!
```go
add := od.New == forward
// sanity check
if DEBUG {
    _, exists := s.unspentOutputs[od.ID]
    if exists == add {
        panic("rogue output")
    }
}
if add {
    s.unspentOutputs[od.ID] = od.Output
} else {
    delete(s.unspentOutputs[od.ID])
}
```
definitely less readable though

consensus/fork.go
```diff
@@ -16,95 +16,108 @@ func (s *State) heavierFork(newNode *BlockNode) bool {
...
-		panic(err)
+// backtrackToBlockchain returns a list of nodes that go from the current node
+// to the first parent that is in the current blockchain.
+func (s *State) backtrackToBlockchain(bn *BlockNode) (nodes []*BlockNode) {
+	nodes = append(nodes, bn)
+	currentChainID := s.currentPath[bn.Height]
+	for currentChainID != bn.Block.ID() {
+		bn = bn.Parent
+		currentChainID = s.currentPath[bn.Height]
+		nodes = append(nodes, bn)
```
lukechampine:
```go
for ; s.currentPath[bn.Height] != bn.Block.ID(); bn = bn.Parent {
    nodes = append(nodes, bn)
}
```
SO CONCISE

consensus/fork.go
```diff
@@ -16,95 +16,108 @@ func (s *State) heavierFork(newNode *BlockNode) bool {
...
-	diffSet := s.applyContractMaintenance(&bd.BlockChanges)
-	diffs = append(diffs, diffSet...)
-
-	// Update the current block and current path variables of the longest fork.
-	height := s.blockMap[b.ID()].Height
-	s.currentBlockID = b.ID()
-	s.currentPath[height] = b.ID()
+	outputDiffs, contractDiffs := s.applyContractMaintenance()
+	bn.OutputDiffs = append(bn.OutputDiffs, outputDiffs...)
+	bn.ContractDiffs = append(bn.ContractDiffs, contractDiffs...)
```
lukechampine: this pattern pops up in a couple places. Would be nice if you could just write:
```go
bn.Diffs = append(bn.Diffs, s.applyContractMaintenance())
```
this might complicate the code in other places though.

consensus/notifications.go
```diff
@@ -85,27 +30,38 @@ func (s *State) notifySubscribers(cc ConsensusChange) {
...
+	reversedNodes := s.backtrackToBlockchain(node)
+	// Eliminate the last node, which is the pivot node, whose diffs are already
+	// known.
+	reversedNodes = reversedNodes[:len(reversedNodes)-1]
+	for _, reversedNode := range reversedNodes {
+		removedBlocks = append(removedBlocks, reversedNode.Block.ID())
+	}
+
+	// Get all the ids going forward from the pivot node.
+	height := reversedNodes[len(reversedNodes)-1].Height
```
lukechampine: just clarifying: you call this the "pivot node", but in the previous comment you said you removed the pivot node. Is "pivot node" just a name for the last node?

consensus/notifications.go
```diff
@@ -85,27 +30,38 @@ func (s *State) notifySubscribers(cc ConsensusChange) {
...
+	_, exists = s.currentPath[height]
+	for exists {
+		node := s.blockMap[s.currentPath[height]]
+		addedBlocks = append(addedBlocks, node.Block.ID())
+
+		height++
+		_, exists = s.currentPath[height]
+	}
```
lukechampine:
```go
for node, exists := s.currentPath[height]; exists; height++ {
    addedBlocks = append(addedBlocks, node.Block.ID())
    node, exists = s.currentPath[height]
}
```
don't mind me

consensus/transactionpool.go
```diff
@@ -164,3 +164,30 @@ func (s *State) TransactionPoolDump() (transactions []Transaction) {
...
+	err = s.validTransaction(t)
+	if err != nil {
+		return
+	}
+
+	// Add the transaction to the pool.
+	s.addTransactionToPool(&t)
+
+	return
+}
```
lukechampine: this...this is a good function. Only thing that sticks out is `s.validTransaction` taking `t` instead of `&t`.

modules/host/contractcreation.go
```diff
@@ -54,68 +66,60 @@ func (h *Host) considerContract(t consensus.Transaction) (updatedTransaction con
...
 		return
 	}
 	updatedTransaction, err = h.wallet.SignTransaction(id, true)
+	if err != nil {
+		err = HostCapacityErr // hide the fact that the host is having wallet issues.
+		return
+	}
+
+	// Update the amount of space the host has for sale.
+	h.spaceRemaining -= int64(fileSize)
```
lukechampine: why is `spaceRemaining` an `int64`? Is it because `FileInfo.Size` is an `int64`?

modules/host/contractcreation.go
```diff
@@ -129,25 +133,60 @@ func (h *Host) considerContract(t consensus.Transaction) (updatedTransaction con
...
+	if err != nil {
 		return
 	}
+	contract := txn.FileContracts[0]
 
 	// Check that the contained FileContract fits host criteria for taking
 	// files, replying with the error if there's a problem.
 	h.mu.Lock()
-	t, err = h.considerContract(t)
+	txn, err = h.considerContract(txn, startBlock)
```
lukechampine: can we make `considerContract` just take a contract instead of a transaction?

modules/host/proofs.go
```diff
@@ -0,0 +1,153 @@
...
+
+	"github.com/NebulousLabs/Sia/consensus"
+	"github.com/NebulousLabs/Sia/hash"
+)
+
+// TODO: Hold off on both storage proofs and deleting files for a few blocks
+// after the first possible opportunity to reduce risk of loss due to
+// blockchain reorganization.
+func (h *Host) threadedConsensusListen(updateChan chan struct{}) {
+	for _ = range updateChan {
```
lukechampine: I think as of Go 1.4 you can just write `for range updateChan {`

consensus/fork.go
DavidVorick: I was doing that for a bit actually, but I found it to be opaque to call a function inside of append. When I went back to the code later I got confused "where does it apply the maintenace?".

So I split it out.

consensus/notifications.go
DavidVorick: The pivot node is the node around which the blockchain is changing. So you go 10a -> 9a -> 8 -> 9b -> 10b -> 11b. 8 is the pivot node. reversedNodes is going to be [10a, 9a, 8]. We pop off 8 b/c it doesn't need to be reversed: [10a, 9a]. Then we add diffs for 10a then 9a, and then we add diffs going forward from 8, diffs for [9b, 10b, 11b]

modules/host/contractcreation.go
DavidVorick: yes. But also because it can go negative.

If you're storing 10GB of files, and then you set the max to 8GB, the host is just going to happily sit at -2GB until stuff expires.

modules/host/contractcreation.go
DavidVorick: no :(

At some point, `considerContract` is actually going to take a set of contracts and consider them as 1. Contracts got changed so that there's only 1 output. But hosts would prefer to get paid every N blocks right, so you'll have a transaction with like 12 contracts, each for longer and longer durations (one finalizes at block 100, one at 200, 300, etc.) but all paying the same price. But since it's the same hash, the host knows that it's fair.

It's going to get more complicated, need to think of the simplest way to achieve all goals.

modules/wallet/wallet.go
```diff
@@ -0,0 +1,108 @@
...
+		return
+	}
+	t, err = w.SignTransaction(id, true)
+	if err != nil {
+		return
+	}
+	err = w.state.AcceptTransaction(t)
+	if err != nil {
+		return
+	}
```
lukechampine: code like this really makes me miss monads :(
It still looks very clean though.

siad/main.go
```diff
@@ -87,10 +82,28 @@ func init() {
...
+	err := config.expand()
+	if err != nil {
 		fmt.Println("Bad config value:", err)
-	} else if err := startDaemon(config); err != nil {
+	}
+	_, err = newDaemon(daemonConfig)
+	if err != nil {
 		fmt.Println("Failed to start daemon:", err)
+	} else {
+		select {}
```
lukechampine: I thought the error convention was:
```go
_, err = newDaemon(daemonConfig)
if err != nil {
    fmt.Println("Failed to start daemon:", err)
    return
}

// block indefinitely
select {}
```

I don't like `select {}` though. How will the program exit? I'd prefer something like:

```go
d, err = newDaemon(daemonConfig) // doesn't spawn handler process
if err != nil {
    fmt.Println("Failed to start daemon:", err)
    return
}

// handle requests until quit signal is received
d.listen()
```

siad/network.go
```diff
@@ -92,16 +151,19 @@ func (c *Core) SendBlocks(knownBlocks [32]consensus.BlockID) (blocks []consensus
...
 		// log error
 		// TODO: try a different peer?
 		return
 	}
 	for _, block := range newBlocks {
-		c.AcceptBlock(block)
+		err = d.state.AcceptBlock(block)
+		if err != nil {
+			// TODO: something
+		}
```
lukechampine: :+1: 
need to remove the sleep at the bottom of the function though.

lukechampine:
What is `siad/walletDir`? If it's a dir, why does git think it's a binary file?

modules/host/contractcreation.go
lukechampine: I'm just saying you could split the functionality somewhat, so that `considerContract` only operates on the contract, and the rest is handled elsewhere. Alternatively, rename it to `considerContractTxn`. But right now there's an discrepancy between the function name and its arguments.

DavidVorick:
I was thinking wallet could be a dir, with one file for each set of addresses. But I think that's premature so I'll rename it back to walletFile

siad/network.go
DavidVorick: if an error is returned, it means that there's a problem with the block, and that that block will always return an error.

(actually, not strictly true, the block might be a FutureBlock, which means you need to handle it differently).

But mostly it means you should stop catching up from the person you're currently asking.

modules/host/contractcreation.go
lukechampine: I mean, you could add checks for that. You need a negative check anyway. Not a big deal though, it just seemed unusual for you to use a signed int.

siad/main.go
DavidVorick: oh right. still getting used to it.

The select{} was mostly a temporary thing. I know there's some sort of signal handling in the code, I figured we'd add it later.

consensus/statehash.go
```diff
@@ -0,0 +1,96 @@
...
+		if err != nil {
+			panic(err)
+		}
+		sortedOutputs = append(sortedOutputs, output)
+	}
+	return
+}
+
+// StateHash returns the markle root of the current state of consensus.
+func (s *State) stateHash() hash.Hash {
```
lukechampine: this stutters a bit -- is just `Hash` a bad name?

consensus/transactionpool.go
DavidVorick: actually they should all take `t` and not `&t`, but I haven't gotten there yet.

modules/host/contractcreation.go
DavidVorick: what are we supposed to do if someone sets a value too low though? I think the most would be to add a notifiaction. It would be bad to delete stuff.

modules/host/contractcreation.go
DavidVorick: hm. I will keep that in mind when I return to modifying the host.

I'm not very certain about how hosting is going to work yet.

modules/host/proofs.go
DavidVorick: didn't know that

still though would rather wait, since most of our machines are currently on go 1.3

modules/wallet/wallet.go
DavidVorick: interesting.

modules/host/contractcreation.go
lukechampine: yeah, I would just return an error. Maybe add a "-f" option if you want to set it anyway.

siad/main.go
```diff
@@ -87,11 +82,32 @@ func init() {
...
+	}
+	_, err = newDaemon(daemonConfig)
+	if err != nil {
 		fmt.Println("Failed to start daemon:", err)
+		return
 	}
+
+	// TODO: Instead of blocking forever, call a function that can receive a
+	// close signal.
+	select {}
```
lukechampine: you could just move the call to `d.handle()` here...

consensus/statehash.go
DavidVorick: is it a problem that state.stateHash returns a hash.Hash?

For now I would like to keep the name, but will keep thinking about it. Mostly, there are a lot of things that could be hashed. It should really probably be called consensusHash

siad/main.go
DavidVorick: :}

consensus/buildtest.go
```diff
@@ -0,0 +1,32 @@
...
+	MinimumCoinbase = Currency(30000)
+
+	GenesisTimestamp = Timestamp(1417070299) // Approx. 1:47pm EST Nov. 13th, 2014
+)
+
+var (
+	RootTarget = Target{64}
+	RootDepth  = Target{255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255}
+
+	MaxAdjustmentUp   = big.NewRat(1001, 1001)
```
lukechampine: should be `1001, 1000`?

PR #288, opened 2015-01-24
Minor fixes

DavidVorick:
The miner now pulls updates as intended.

Also `make release` and `make xc` should work as intended too, which was not previously the case.

I'm not sure how build flags work in goxc, setting -build-tags=release wasn't doing the trick. So I also modified the build tags in buildrelease.go as a workaround. It's probably for the better anyway, now someone naively running `go install` won't be greeted with an excessive amount of errors, but instead will be greeted with a binary that has release constants.

### DIFF
--- a/consensus/state.go
+++ b/consensus/state.go
@@ -202,13 +202,10 @@ func (s *State) EarliestTimestamp() Timestamp {
 	return s.currentBlockNode().earliestChildTimestamp()
 }
 
-// Cheater function.
-func (s *State) MinerVars() (parent BlockID, txns []Transaction, target Target, earliestTimestamp Timestamp) {
+func (s *State) RLock() {
 	s.mu.RLock()
-	defer s.mu.RUnlock()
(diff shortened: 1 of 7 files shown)


consensus/state.go
```diff
@@ -202,13 +202,10 @@ func (s *State) EarliestTimestamp() Timestamp {
...
-	defer s.mu.RUnlock()
-	parent = s.currentBlockNode().Block.ID()
-	txns = s.transactionPoolDump()
-	target = s.currentBlockNode().Target
-	earliestTimestamp = s.currentBlockNode().earliestChildTimestamp()
-	return
+}
+
+func (s *State) RUnlock() {
+	s.mu.RUnlock()
```
lukechampine: not much prettier than having a single cheater function. Oh well.

DavidVorick: This is nicer because the other modules will need to be doing similar things.

We'd end up with a different cheater function for every module, less pretty.

PR #289, opened 2015-01-26
add orphan handling

DavidVorick:
this probably fixes a few synchronization issues you've noticed, but it's not sufficiently tested.

Even if it doesn't fix the one's you've noticed, it fixes ones that would exist if someone tried to make them happen.

### DIFF
--- a/consensus/blocknode.go
+++ b/consensus/blocknode.go
@@ -61,9 +61,14 @@ func (node *BlockNode) setTarget() {
 
 // State.addBlockToTree() takes a block and a parent node, and adds a child
 // node to the parent containing the block. No validation is done.
-func (s *State) addBlockToTree(b Block) (newNode *BlockNode) {
+func (s *State) addBlockToTree(b Block) (err error) {
+	err = s.validHeader(b)
+	if err != nil {
+		return
+	}
+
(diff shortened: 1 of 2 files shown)


consensus/blocknode.go
```diff
@@ -72,11 +77,35 @@ func (s *State) addBlockToTree(b Block) (newNode *BlockNode) {
...
+			// allows them to cluster the processing. Utimately I think
+			// bandwidth will be the bigger issue, we'll reject bad orphans
+			// before we verify signatures.
+			_ = s.addBlockToTree(child)
+			// There's nothing we can really do about this error, because it
+			// doesn't reflect a problem with the current block. It means that
+			// someone managed to give us orphans with errors.
+		}
+		delete(s.missingParents, b.ID())
+	}
```
lukechampine: `exists` is technically unnecessary here, because a `range` over a nil map is a no-op, and a `delete` on a nil map is too.
```go
for _, child := range s.missingParents[b.ID()] {
    s.addBlockToTree(child)
}
delete(s.missingParents, b.ID())
```

DavidVorick: fixed

PR #291, opened 2015-01-26
make RandomPeer return an error

### LINKED ISSUE #140: RandomPeer should not panic
Really, there should always be a peer to select, even if it's yourself. But this is probably best practice anyway.

lukechampine:
This actually fixes #140.

### DIFF
--- a/network/network_test.go
+++ b/network/network_test.go
@@ -81,7 +81,7 @@ func TestPeerSharing(t *testing.T) {
 	peer := Address("foo:9001")
 	tcps.AddPeer(peer)
 	// tcps only has one peer, so RandomPeer() should return peer
-	if tcps.RandomPeer() != peer {
+	if p, err := tcps.RandomPeer(); err != nil || p != peer {
 		t.Fatal("server has bad peer list:", tcps.AddressBook())
 	}
 
--- a/siad/network.go
+++ b/siad/network.go
@@ -54,19 +54,22 @@ func (d *daemon) initializeNetwork(addr string, nobootstrap bool) (err error) {
 		return
 	}
 
-	// Bootstrapping may take a while, use a gothread.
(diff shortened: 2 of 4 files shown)


network/network_test.go
```diff
@@ -81,7 +81,7 @@ func TestPeerSharing(t *testing.T) {
 	peer := Address("foo:9001")
 	tcps.AddPeer(peer)
 	// tcps only has one peer, so RandomPeer() should return peer
-	if tcps.RandomPeer() != peer {
+	if p, err := tcps.RandomPeer(); err != nil || p != peer {
```
DavidVorick: IF thing1, thing2 := function1; statement1 || statement 2 {

pretty busy line

siad/network.go
```diff
@@ -65,7 +65,10 @@ func (d *daemon) initializeNetwork(addr string, nobootstrap bool) (err error) {
 		// Every 2 minutes, call CatchUp() on a random peer. This helps with
 		// synchronization.
 		for {
-			go d.CatchUp(d.network.RandomPeer())
+			peer, err := d.network.RandomPeer()
+			if err == nil {
```
DavidVorick: need to add a comment if you're leaving it `err == nil`

something like 
```go
if err == nil { // only call CatchUp if there was no error
}
```

would be fine

DavidVorick: just something to acknowledge that convention is being broken

lukechampine: how about this?
```go
for {
    time.Sleep(time.Minute * 2)
    peer, err := d.network.RandomPeer()
    if err != nil {
        continue
    }
    go d.CatchUp(peer)
}
```

DavidVorick: it's weird that you can't just call `for time.Sleep(time.Minute * 2) {`, but yeah this looks fine to me.

PR #293, opened 2015-01-28
Transaction Pool (first bit)

DavidVorick:
I implemented a transaction pool, or at least the beginnings of one.

It doesn't get updates from the state, which means it'll never realize that it's using transactions which have been accepted to the state.

It doesn't add storage proof transactions to the transaction set. This is also a problem.

In this PR, there's a decent amount of documentation that got added. There is also a decent amount of moving things around, and a decent amount of new functionality.

I expect to finish up with the wallet and transaction pool functionality tomorrow, including some more tests and making the API functional again for the wallet and miner stuff.


Developers.md
```diff
@@ -59,3 +63,41 @@ Functions prefixed 'threaded' (example 'threadedMine') are meant to be called
...
+transactions. Where possible, these explicit assumptions should be validated.
+
+Example:
+
+if consensus.DEBUG {
+	_, exists := tp.usedOutputs[input.OutputID]
+	if exists {
+		panic("incorrect use of addTransaction")
+	}
+}
```
lukechampine: if you want this to be a .md file, you should use markdown conventions: use \``` for code blocks, # for headers, etc. Otherwise, make it a .txt.

Developers.md
```diff
@@ -59,3 +63,41 @@ Functions prefixed 'threaded' (example 'threadedMine') are meant to be called
...
+
+In the example, a panic is called for incorrect use of the function, but only
+in panic mode. This failure will be invisible in production code, but the code
+will have higher performace because the code should never fail anyway.
```
lukechampine: "only in debug mode"

Makefile
```diff
@@ -4,16 +4,18 @@ fmt:
...
+	touch consensus/blocknode.go
 	go install -tags=dev ./...
 
 clean:
 	rm -rf hostdir release whitepaper.aux whitepaper.log whitepaper.pdf         \
 		sia.wallet sia/test.wallet sia/hostdir* sia/renterDownload
 
-test: clean
+test: clean fmt
+	touch consensus/blocknode.go
```
lukechampine: replace these with `.FORCE`

Standard.md
```diff
@@ -0,0 +1,66 @@
...
+forks, and for protocol relevant information. Any arbitrary data is allowed by
+consensus, but only certain arbitrary data is considered standard.
+
+Arbitrary data that is prefixed by the string 'NonSia' is always allowed. This
+indicates that the remaining data has no relevance to Sia protocol rules, and
+never will.
+
+Arbitrary data that is prefixed by the string 'HostAnnouncement' is allowed,
+but only if the data within accurately decodes to the HostAnnouncement struct
+found in modules/hostdb.go, and contains no extra information.
```
lukechampine: These prefixes should probably be constant length for easier decoding. I mean, I guess you could use `strings.Prefix`, but...bleh. I use 8 bytes for RPCs.

The name of this file is a bit vague. It should probably just go in the wiki. So should the API docs, now that I think of it.

consensus/fork.go
```diff
@@ -177,13 +177,27 @@ func (s *State) forkBlockchain(newNode *BlockNode) (err error) {
...
+
+		// If the diffs have not been generated, call generateAndApply.
 		err = s.generateAndApplyDiff(backtrackNodes[i])
 		if err != nil {
 			// Invalidate and delete all of the nodes after the bad block.
 			s.invalidateNode(backtrackNodes[i])
 
 			// Rewind the validated blocks
-			for j := 0; j < i; j++ {
+			for j := 0; j < len(appliedNodes)-1; j++ {
```
lukechampine: I feel like there's a better way to write this... `for range appliedNodes {` would *almost* work, but you want one iteration fewer than that. Maybe `for range appliedNodes[1:] {`.

Alternatively, use a function called `s.invertBlock(Block)`, or at least rename this one to `s.invertLastBlock()`.

consensus/statehash.go
```diff
@@ -18,9 +18,9 @@ func (s *State) sortedUtxoSet() (sortedOutputs []Output) {
 	for _, utxoString := range unspentOutputStrings {
 		var outputID OutputID
 		copy(outputID[:], utxoString)
-		output, err := s.output(outputID)
-		if err != nil {
-			panic(err)
+		output, exists := s.output(outputID)
+		if !exists {
+			panic("output doesn't exist?")
```
lukechampine: had to double check this one, but it looks like `sortedUtxoSet` is only ever called when `DEBUG` is set.

consensus/target.go
```diff
@@ -0,0 +1,44 @@
...
+// IntToTarget converts a big.Int to a Target.
+func IntToTarget(i *big.Int) (t Target) {
+	// i may overflow the maximum target.
+	// In the event of overflow, return the maximum.
+	if i.BitLen() > 256 {
+		return RootDepth
+	}
+	b := i.Bytes()
+	// need to preserve big-endianness
+	offset := hash.HashSize - len(b)
```
lukechampine: We could get rid of the `hash` import by writing `len(t[:])` here instead.

modules/transactionpool/accept.go
```diff
@@ -0,0 +1,182 @@
...
+// AcceptTransaction takes a new transaction from the network and puts it in
+// the transaction pool after checking it for legality and consistency.
+func (tp *TransactionPool) AcceptTransaction(t consensus.Transaction) (err error) {
+	tp.mu.Lock()
+	defer tp.mu.Unlock()
+	tp.state.RLock()
+	defer tp.state.RUnlock()
+
+	// Check that the transaction follows 'Standard.md' guidelines.
+	err = standard(t)
```
lukechampine: solid function name ಠ_ಠ

modules/transactionpool/standard.go
```diff
@@ -0,0 +1,43 @@
...
+	TransactionSizeLimit = 64 * 1024
+)
+
+// standard implements the rules outlined in Standard.md, and will return an
+// error if any of the rules are violated.
+func standard(t consensus.Transaction) (err error) {
+	// Check that the size of the transaction does not exceed the standard
+	// established in Standard.md
+	encodedTxn := encoding.Marshal(t)
+	if len(encodedTxn) > TransactionSizeLimit {
```
lukechampine: can you combine these lines?

modules/transactionpool/storageproofs.go
```diff
@@ -0,0 +1,87 @@
...
+	heightMap[hash.HashObject(t)] = t
+	return
+}
+
+func (tp *TransactionPool) storageProofTransactionSet(remainingSize int) (transactions []consensus.Transaction, sizeUsed int) {
+	contractsSatisfied := make(map[consensus.ContractID]struct{})
+
+	// Get storage proofs for all heights from 12 earlier to the current
+	// height.
+	for height := tp.state.Height() - 12; height != tp.state.Height(); height++ {
```
lukechampine: beware of magic numbers

modules/transactionpool/standard.go
```diff
@@ -0,0 +1,43 @@
...
+
+	// Check that transactions with storage proofs follow the storage proof
+	// rules.
+	if len(t.StorageProofs) != 0 {
+		if len(t.Inputs) > 1 ||
+			len(t.MinerFees) > 1 ||
+			len(t.Outputs) > 1 ||
+			len(t.FileContracts) != 0 ||
+			len(t.ArbitraryData) != 0 {
+			err = errors.New("transaction has storage proofs but does not follow the storage proof rules")
```
lukechampine: if you're going to return a generic error, you may as well make `standard` return a `bool`.

modules/transactionpool/storageproofs.go
```diff
@@ -0,0 +1,87 @@
...
+			if err != nil {
+				continue // don't remove the transaction because it might be valid on another fork. (this action is only taken for storage proofs)
+			}
+
+			for _, proof := range txn.StorageProofs {
+				_, exists := contractsSatisfied[proof.ContractID]
+				if exists {
+					continue // this storage proof was already made in a different transaction.
+				}
+			}
```
lukechampine: what is the purpose of this loop?

Developers.md
DavidVorick: hm. The problem is that you can't read markdown in vim :(

Makefile
DavidVorick: how?

Standard.md
DavidVorick: I find wiki's to be confusing, because there isn't a list of pages, or there are too many pages. I'd rather have a bunch of files in the top directory or in a doc/ folder. That forces you to keep the volume of information manageable.

consensus/fork.go
DavidVorick: oh I think it was confusing me too. `j < i` was probably sufficient.

DavidVorick: need to write some tests for this so I can be sure

consensus/statehash.go
DavidVorick: go up two function calls?

It's only ever called during testing and when hashing the state. Right now, the state is only ever hashed in debug mode. We might fix this by adding a new rule to consensus about including the state hash in blocks.

modules/transactionpool/accept.go
DavidVorick: hmmm. Bitcoin has an equivalent function named IsStandard(). would that be a better name?

It basically enforces a set of rules that protects miners against potential future soft forks without requiring the miners to upgrade when the soft forks happen. It's a good thing to have around.

modules/transactionpool/storageproofs.go
DavidVorick: should I make that a constant?

The only risk here is that there's a reorg that's greater than 12 blocks, in which case a bunch of storage proofs might get dumped.

modules/transactionpool/standard.go
DavidVorick: it's only partially generic, you get several bits of information:

- if the transaction is too large
- if the transaction violates storage proof rules
- if the transaction violates arbitrary data rules

modules/transactionpool/storageproofs.go
DavidVorick: oh. A label is probably needed.

Developers.md
lukechampine: The whole point of markdown is that it's readable whether formatted or not.

Makefile
lukechampine: look at the [mkvsynth Makefile](https://github.com/NebulousLabs/mkvsynth/blob/master/Makefile#L48). You add a blank `.FORCE` target, and then add it as a dependency wherever you have `touch` now.

Standard.md
lukechampine: We don't need to go as far as bitcoin, but most altcoins have wiki-based documentation. The point is to separate documentation from code. ("Why would you want to do that?" -- because things like Standard.md and the API docs aren't directly tied to the codebase. They're implementation-independent.)

consensus/statehash.go
lukechampine: right, first I had to look at everyone who was calling `sortedUtxoSet`, and so I grepped and saw that it was only called from `hashState`, but the calls weren't wrapped in `DEBUG`. So then I had to grep for all the calls to `hashState` to verify that *those* calls were wrapped in `DEBUG`s.

modules/transactionpool/accept.go
lukechampine: It's definitely a good function, it's just the name that's lacking. `IsStandard` would be better, or maybe even `IsStandardTransaction`.

modules/transactionpool/storageproofs.go
lukechampine: not a big deal, but I noticed some magic numbers in a couple places. Basically, whenever you find yourself writing a literal, you should consider if it's worth defining a `const` for it.

Developers.md
DavidVorick: Alright I'll switch it all over in a different pr

Standard.md
DavidVorick: That's fine, except that I view wikis to be absolutely awful forms of documentation. I think most of the core devs agree with us, which is why you get this: https://bitcoin.org/en/developer-guide

I'm okay with breaking the documentation out of the codebase but not okay with using a wiki as the replacement.. Wikis are good for when you already know what you are looking for, less good if you don't know what you are looking for.

Putting it on a webpage might work well enough.

consensus/fork.go
DavidVorick: `len(appliedNodes)-1` will always be equal to `i` at this part of the code.

consensus/statehash.go
DavidVorick: yeah. Doing my best to remove all the sharp edges. Just needs more love is all.

Sharp edges are good in developer mode though they let you know stuff is not right.

Issue #296, opened 2015-01-29
Distilled Goals for Release


DavidVorick:
I tried to think about what we really need right now, and I narrowed it down to 4 things.

- [x] Synchronization must be strong. All nodes should always be within one block of each other. All blocks should propagate through the network within a minute. All transactions should propagate the network within a minute, and all transactions should be confirmed within one or two blocks.
    CatchUp should always work on the first shot. This may mean that we need it to talk to several nodes, or prioritize the linode as the first source of information (at least for the headers and the highest block). The user should be able to see some sort of progress bar, indicating what block the user has synchronized to and what block is currently understood to be the highest block. There should never be any need to manually tell the node to synchronize.
- [x] Sending an receiving money should feel immediate. When you send money, it immediately disappears from your wallet, and quickly appears in the other person's wallet. The transaction should appear in the other person's wallet long before a block is mined.
- [x] Uploading a file should always work, or should instantly return a notification that there is insufficient balance. If the user has enough raw coins to upload a file, then the client+backend should immediately acknowledge that an upload is in progress, and give some estimate for progress. The estimate can be wrong and the uploading can take a while, but the user needs to feel certain that the upload *is* happening and needs to be able to see the file that they put on the network in their file list. The drop in balance should be immediate, and should be enough to cover all hosts. We don't actually know how much it's going to cost, but we should find something creative to do with the remaining money. I don't think it makes sense to give it back, because then the user is receiving money that they don't understand. (For now, we'll just keep it in a 'repair fund' or something, to use to repair files, and to use when buying more files).
- [x] Downloading a file should also be immediate and responsive. There needs to be a progress bar where the user can view all of their in-progress downloads. If there are not enough peers online to facilitate the download, some sort of 'finding peers' icon should be displayed/animated. (hopefully this never happens because redundancy is high enough, but in the early days of the network it might be a problem).

That's it. 4 things that have to work well. 2 is almost done, but that's not going to work until we have 1 finished as well.

--------------

I'm adding a fifth thing.

- [x] Users do not lose data when they turn off their machine. This means that the renter's files need to be saved to disk, the wallet's keys need to be saved to disk, and the host's obligations need to be saved to disk. Stuff that can be re-downloaded (like the blockchain) does not need to be saved.

DavidVorick:
[...]

PR #297, opened 2015-01-29
Wallet Update

DavidVorick:
The wallet now updates when you ask for the balance. The update should be thread-safe and accurate, and should include unconfirmed transactions (excluding those with storage proofs, but our software will never put money sends in the same transaction as a storage proof).

Also not tested.

My future plans are to bring back the api, and then add a bunch of testing around the wallet and tpool and consensus code to make sure things are running smoothly.

Then we can test things live and see if we can check off points one and two in issue #296. My guess is that the synchronization code needs work.

### DIFF
--- a/modules/wallet/update.go
+++ b/modules/wallet/update.go
@@ -1,62 +1,96 @@
 package wallet
 
 import (
-// "github.com/NebulousLabs/Sia/consensus"
+	"github.com/NebulousLabs/Sia/consensus"
 )
 
-func (w *Wallet) update() error {
-	/*
-		for _, diff := range diffs {
-			if diff.New {
-				if spendableAddress, exists := w.spendableAddresses[diff.Output.SpendHash]; exists {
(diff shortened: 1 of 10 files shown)


modules/wallet/update.go
```diff
@@ -0,0 +1,96 @@
...
+
+// applyDiff will take the output and either add or delete it from the set of
+// outputs known to the wallet. If adding is true, then new outputs will be
+// added and expired outputs will be deleted. If adding is false, then new
+// outputs will be deleted and expired outputs will be added.
+func (w *Wallet) applyDiff(diff consensus.OutputDiff, adding bool) {
+	isNew := diff.New
+	if !adding {
+		isNew = !isNew
+	}
```
lukechampine: just a thought, but maybe we should replace the `New`/`adding` boolean with an enum:
```go
const (
    DiffForward = iota
    DiffBackward
)
```

PR #299, opened 2015-01-30
Testing Round One

DavidVorick:
Added some stuff, including:

1. `make cover`, which produces testing coverage stats for all of our packages that have test files. Eventually all packages should have test files.

2. documentation to the makefile

3. Some organizational stuff

4. A whole bunch of tests for the consensus package, bringing coverage up to 44%. Right now pretty much only empty blocks are tested, which is why the number is so low. No contracts or financial stuff is tested, and the exported functions don't get much love either. That'll be in round two.


DavidVorick:
Okay added a bunch more stuff.

Some more tests, some changes to how block ids are calculated, and some extra organizational things.

consensus/blocks_test.go
```diff
@@ -0,0 +1,459 @@
...
+	// Compare the metrics and report an error if something has changed.
+	if bbLen != len(s.badBlocks) ||
+		bmLen != len(s.blockMap) ||
+		mpLen != len(s.missingParents) ||
+		cpLen != len(s.currentPath) ||
+		uoLen != len(s.unspentOutputs) ||
+		ocLen != len(s.openContracts) ||
+		stateHash != s.StateHash() {
+		t.Error("state changed after getting a repeat block.")
+	}
```
lukechampine: Isn't it sufficient just to recalculate and compare the state hash?

consensus/blocks_test.go
```diff
@@ -0,0 +1,459 @@
...
+
+// TestBlockTimestamps creates a new state and uses it to call
+// testBlockTimestamps.
+func TestBlockTimestamps(t *testing.T) {
+	s := CreateGenesisState()
+	testBlockTimestamps(t, s)
+}
+
+// TestConstants makes sure that the testing constants are being used instead
+// of the developer constants or the release constants.
```
lukechampine: this is getting a little ridiculous

consensus/types.go
```diff
@@ -135,12 +132,14 @@ func (b Block) CheckTarget(target Target) bool {
...
 // ExpectedTransactionMerkleRoot() returns the expected transaction
 // merkle root of the block.
-func (b Block) TransactionMerkleRoot() hash.Hash {
-	var transactionHashes []hash.Hash
-	for _, transaction := range b.Transactions {
-		transactionHashes = append(transactionHashes, hash.HashBytes(encoding.Marshal(transaction)))
+func (b Block) MerkleRoot() hash.Hash {
+	var hashes []hash.Hash
+	hashes = append(hashes, hash.HashObject(b.Timestamp))
+	hashes = append(hashes, hash.HashObject(b.MinerAddress))
```
lukechampine:
```go
hashes := []hash.Hash{
    hash.HashObject(b.Timestamp)),
    hash.HashObject(b.MinerAddress)),
}
```

DavidVorick: Not worried about performance right now, just something that's clean and easy to read.

I'd imagine that consecutive appends get optimized by the compiler anyway.

consensus/blocks_test.go
DavidVorick: what is?

Issue #290, opened 2015-01-26
Procedure for Creating a Contract


DavidVorick:
1. Renter contacts host with proposed terms. 

- FileSize
- Window size (how many blocks a host has to submit each proof)
- Window start points (which block each window will start at) (each window should start an equal number of blocks apart)
- Client contribution towards payout each window
- Host contribution towards payout each window
- Valid Proof Address
- Missed Proof Address

2. If the host agrees with the contract, the host will allocate space for the file and send an agreement message. Host should agree with the client contribution and the host contribution, should agree that the window size is large enough, that the start points are acceptable, and that the distance between window start points is acceptable, the total duration of the contract is acceptable (total duration starts from the current height, not from the starting point of the first window).

3. The renter will upload the file

4. The renter will upload a transaction containing all of the contracts, and inputs, with a signature. 'whole transaction' should be set to false  when the renter signs the transaction.

5. The host will check that everything matches the prior agreement. There should be one contract for each window, the payout for each contract should be (client payout per window + host payout per window). Each contract should have a start point lined up with the `window start points` and and end point that is `window size` after the start point.

Each contract should now have a merkle root, and that merkle root should match  the merkle root of the file uploaded to the host.

6. the host adds ('host contribution each window' * number of windows) funds to the transaction and signs 'whole transaction', and submits it to the blockchain.

7. Renter watches blockchain for submitted transaction + contracts. If they don't appear in time, the renter knows to try again with a different host. ('in time' means that the transaction appears in the blockchain before the first window starts. Otherwise the signatures will all be invalidated).

lukechampine:
implemented in #306

DavidVorick:
ah sorry

PR #306, opened 2015-02-01
Bring back host and renter

lukechampine:
The host and renter (and their associated API calls) are back online. File contracts now conform to the spec outlined in #290 (I hope), so be sure to close that issue in the merge commit. Another nice thing is that Merkle roots are now calculated as the file is being transferred over the network, instead of having to process the entire file twice. Finally, I tweaked the Makefile a bit to prevent slow rebuilds. (tl;dr: back to the `touch` method, but a bit more sophisticated.)

As it often goes, delving into the codebase shed light on a lot of "gotchas", as well as areas for future improvement. I fixed most of the ones I came across, and tried to write the rest down.

Things that should be done soon:
- Need to be making one file contract per window, instead of one big contract
- Storage proof stuff is still commented out; we need a good way for the host to interact with the state
- When checking for host announcements, call `IsStandard` to catch duplicate `FreezeIndex`s
- Tests. Lots of tests.

Thoughts for the future:
- Timeouts need to be more sophisticated. As it stands, when we do file transfers we just use a really, really long timeout. The network package should be in charge of monitoring connections and closing them when certain timeout conditions have been met.
- It's probably a good idea to have a protocol package, or at least one file with all the protocol information in it -- structs, strings, errors, everything that gets sent over the wire. A good time to do this would be when we're ready to start locking down the protocol.
- I don't like having network-level code in the host/renter modules. The specifics of who sends what, and in what order, seems like a protocol-level issue. With careful design, I think it could be moved elsewhere, perhaps to the daemon -- but I'm not yet confident that it would be worth the effort.

Also: net negative. awwwwright.


DavidVorick:
build failed

Makefile
```diff
@@ -6,9 +6,15 @@ all: install
 fmt:
 	go fmt ./...
 
+# REBUILD touches all of the build-dependent source files, forcing them to be
+# rebuilt. This is necessary because the go tool is not smart enough to trigger
+# a rebuild when build tags have been changed.
+REBUILD:
+	@touch consensus/build*.go
```
DavidVorick: positioning seems weird, but I guess it makes sense. The name is weird too because you aren't actually rebuilding anything you're just touching a few files.

Makefile
```diff
@@ -76,6 +81,5 @@ xc: dependencies test-long
 	goxc -arch="amd64" -bc="linux windows darwin" -d=release -pv=0.2.0          \
 		-br=release -pr=beta -include=example-config,LICENSE*,README*           \
 		-tasks-=deb,deb-dev,deb-source,go-test
-	# Need some command here to make sure that the release constants got used.
 
 .PHONY: all fmt install clean test test-long cover whitepaper dependencies release xc
```
DavidVorick: need to add REBUILD to .PHONY

Makefile
```diff
@@ -68,7 +73,7 @@ dependencies:
 
 # release builds and installs release binaries.
 release: dependencies test-long
-	go install -a ./...
```
DavidVorick: need to add REBUILD to both `make release` and `make xc`

consensus/builddev.go
```diff
@@ -21,6 +21,8 @@ const (
 	GenesisTimestamp = Timestamp(1417070299) // Approx. 1:47pm EST Nov. 13th, 2014
 )
 
+// Though these are variables, they should never be changed during runtime.
+// They get altered during testing.
```
DavidVorick: I don't think these comments are needed? Also they are wrong, nothing gets altered during testing.

consensus/buildrelease.go
```diff
@@ -21,6 +21,8 @@ const (
 	GenesisTimestamp = Timestamp(1417070299) // Approx. 1:47pm EST Nov. 13th, 2014
 )
 
+// Though these are variables, they should never be changed during runtime.
+// They get altered during testing.
```
DavidVorick: .

consensus/buildtest.go
```diff
@@ -21,6 +21,8 @@ const (
 	GenesisTimestamp = Timestamp(1417070299) // Approx. 1:47pm EST Nov. 13th, 2014
 )
 
+// Though these are variables, they should never be changed during runtime.
+// They get altered during testing.
```
DavidVorick: .

consensus/types.go
```diff
@@ -21,6 +21,8 @@ type (
 	Target      hash.Hash
 )
 
+var ZeroAddress = CoinAddress{0}
```
DavidVorick: hm. I guess this is good to have.

modules/host.go
```diff
@@ -4,29 +4,54 @@ import (
...
+	NumContracts     int
+}
+
 type Host interface {
-	// Announce puts an annoucement out so that clients can find the host.
-	AnnounceHost(freezeVolume consensus.Currency, freezeUnlockHeight consensus.BlockHeight) (consensus.Transaction, error)
+	// Announce announces the host on the blockchain. A host announcement
+	// requires two things: the host's address, and a volume of "frozen"
+	// (time-locked) coins, used to mitigate Sybil attacks.
+	Announce(network.Address, consensus.Currency, consensus.BlockHeight) error
```
DavidVorick: names might help document stuff in this case. It's not really clear that the BlockHeight corresponds to when you want the coins unfrozen, or that the act of calling Announce is actually going to freeze your coins.

modules/host/contractcreation.go
```diff
@@ -2,258 +2,250 @@ package host
...
-			return
-		}
-		// Output for failed proofs needs to be the 0 address.
-		emptyAddress := consensus.CoinAddress{}
-		if contract.MissedProofAddress != emptyAddress {
-			err = errors.New("burn payout needs to go to the empty address")
-			return
-		}
+	switch {
+	// TODO: check for minheight too?
```
DavidVorick: minheight is meant to have a host reject contracts that don't have a duration of at least X. Checking for it becomes  a bit trickier when you have multiple windows

modules/host/contractcreation.go
```diff
@@ -2,258 +2,250 @@ package host
...
+
+	case terms.ValidProofAddress != h.CoinAddress:
+		return errors.New("coins are not paying out to correct address")
+
+	case terms.MissedProofAddress != consensus.ZeroAddress:
+		return errors.New("burn payout needs to go to the zero address")
+	// TODO: should this be <= ?
+	case terms.Price != h.Price:
+		return errors.New("price does not match host settings")
+	// TODO: should this be >= ?
```
DavidVorick: there's no reason for a host to reject contracts that favor them. In consensus it's nice to be strict because it prevents dumb mistakes and potential future headaches, but with hosts everyone is going to behave differently so you might as well take advantage.

modules/host/contractcreation.go
```diff
@@ -2,258 +2,250 @@ package host
...
+	case terms.Collateral != h.Collateral:
+		return errors.New("collateral does not match host settings")
+	}
+
+	return nil
+}
+
+// verifyContract verifies that the values in the FileContract match the
+// ContractTerms agreed upon.
+// TODO: could this just return a bool (i.e. "contract does not match terms"?)
```
DavidVorick: larger amounts of debugging information are usually nice. Since the code is always written, it's probably better to leave it.

modules/host/contractcreation.go
```diff
@@ -2,258 +2,250 @@ package host
...
+	}()
+
+	// signal that we are ready to download file
+	_, err = encoding.WriteObject(conn, modules.AcceptContractResponse)
+	if err != nil {
+		return
+	}
+
+	// file transfer is going to take a while, so extend the timeout.
+	// This assumes a minimum transfer rate of ~1 Mbps
```
DavidVorick: 1Mbps is pretty optimistic for some connections, because one of the parties is doing uploading.

modules/host/contractcreation.go
```diff
@@ -2,258 +2,250 @@ package host
...
+	conn.SetDeadline(time.Now().Add(time.Duration(terms.FileSize) * 8 * time.Microsecond))
+
+	// simultaneously download file and calculate its Merkle root.
+	tee := io.TeeReader(
+		// use a LimitedReader to ensure we don't read indefinitely
+		io.LimitReader(conn, int64(terms.FileSize)),
```
DavidVorick: :D

modules/renter.go
```diff
@@ -1,19 +1,21 @@
...
+type UploadParams struct {
+	Data     io.ReadSeeker
+	Duration consensus.BlockHeight
+	Delay    consensus.BlockHeight
+	FileSize uint64
+
+	// these fields are not seen by the host
+	Nickname string
+	Pieces   int
 }
```
DavidVorick: Can you add a docstring to this struct?

Makefile
lukechampine: REBUILD is already a dependency of both of them. I suppose it could be more explicit though.

consensus/builddev.go
lukechampine: gotcha, I just noticed they were in the wrong place

modules/host.go
lukechampine: yeah, I was on the fence about that one.

modules/host/contractcreation.go
lukechampine: it's so beautiful. It's like interface satori

Makefile
DavidVorick: oh right. Maybe a comment?

DavidVorick: wait WRONG!

REBUILD is a dependency, but then the binaries are built with test constants. REBUILD needs to be called again because the constants change between testing and releasing.

lukechampine: Is there a reason that `release` and `xc` need to be running `test-long` anyway?

Okay, I actually know the reason, but I don't think it's terribly good one. And you'd need to add `REBUILD` regardless.

modules/host/contractcreation.go
lukechampine: generally, although I don't know if this error is actually printed anywhere. Whatevs

modules/host/contractcreation.go
lukechampine: This is a temporary fix anyway. I imagine 1Mbps will be quite generous during in-house testing.

Makefile
lukechampine: true, but I don't think there's really a better name for it, except `TouchBuildFilesToAlertGoToolThatObjectsNeedToBeRebuild`

PR #309, opened 2015-02-02
Reference Implementation pt 1

### LINKED ISSUE #307: Orphan Handling
I realized that, like with the transaction pool and the future block problem, orphan handling should not be done by the consensus package. It adds a substantial amount of attack surface area that can be avoided by separating orphan handling from the consensus package.

DavidVorick:
This PR contains a bunch of tweaks that get us closer to a full reference implementation. We're pretty close right now.

Stuff that's in here:
- expand Consensus.md to define everything except for encoding
- switch hashing to the blake2 algorithm
- remove orphan handling from the consensus package (fixes #307)
- increased test coverage
- CoveredFields rules are now enforced
- storage proof transaction rules are now enforced
- type system now includes siafunds (though nothing is done with them)
- CreateGenesisState now takes a timestamp, which will make certain test types easier.

For a reference implementation, what remains is:
- 128bit uints that don't overflow
- siafund implementation
- 100 block cooldown on the certain output types
- testing

I think most of it will be done in the next 3 days.


consensus/blocks_test.go
```diff
@@ -153,195 +153,189 @@ func testLargeBlock(t *testing.T, s *State) {
...
 	}
-	if s.currentBlockID != orphan.ID() {
-		t.Error("the states current block is not the reconnected orphan")
+	for b.CheckTarget(s.CurrentTarget()) && b.Nonce < 1000*1000 {
+		b.Nonce++
 	}
-	if beforeHeight != s.Height()-2 {
-		t.Error("height should now be reporting 2 new blocks.")
+	if b.CheckTarget(s.CurrentTarget()) {
+		panic("unable to mine a block with a failing target (lol)")
```
lukechampine: lol

consensus/builddev.go
```diff
@@ -14,6 +14,7 @@ const (
 	TargetWindow          = BlockHeight(80)        // Number of blocks to use when calculating the target.
 	MedianTimestampWindow = 11                     // Number of blocks that get considered when determining if a timestamp is valid. Should be an odd number.
 	FutureThreshold       = Timestamp(3 * 60 * 60) // Seconds into the future block timestamps are valid.
+	SiafundCount          = Currency(10 * 1000)    // The total (static) number of siafunds.
```
lukechampine: I tend to prefer "NumFoos` to `FooCount`. Maybe `TotalSiafunds` would be better here.

consensus/signatures.go
```diff
@@ -0,0 +1,198 @@
...
+			return
+		}
+		err = sortedUnique(cf.ArbitraryData, len(cf.ArbitraryData)-1)
+		if err != nil {
+			return
+		}
+		err = sortedUnique(cf.Signatures, len(cf.Signatures)-1)
+		if err != nil {
+			return
+		}
```
lukechampine: could be simplified further like so:
```go
fields := [][]uint64{cf.Inputs, cf.MinerFees, cf.Outputs...}
for _, field := range fields {
    err = checkSortedUnique(field, len(field)-1)
    if err != nil {
        return
    }
}
```

consensus/types.go
```diff
@@ -14,15 +24,24 @@ type (
...
 	Target      hash.Hash
 )
 
 var ZeroAddress = CoinAddress{0}
 
+var FileContractIdentifier = Identifier{'f', 'i', 'l', 'e', ' ', 'c', 'o', 'n', 't', 'r', 'a', 'c', 't'}
+var SiacoinOutputIdentifier = Identifier{'s', 'i', 'a', 'c', 'o', 'i', 'n', ' ', 'o', 'u', 't', 'p', 'u', 't'}
+var SiafundOutputIdentifier = Identifier{'s', 'i', 'a', 'f', 'u', 'n', 'd', ' ', 'o', 'u', 't', 'p', 'u', 't'}
+
+var ED25519Identifier = Identifier{'e', 'd', '2', '5', '5', '1', '9'}
```
lukechampine: muh verbosity

I'm starting to shift towards the standard library style of "prefixes rather than suffixes" (e.g. ErrFoo). I'd prefer to see these named `IdFileContract`, `Id25519`, etc.

consensus/types.go
```diff
@@ -197,29 +222,78 @@ func (b Block) MinerPayoutID(i int) OutputID {
...
-		"contract",
+		FileContractIdentifier,
+		t.Inputs,
+		t.MinerFees,
+		t.Outputs,
+		t.FileContracts,
+		t.StorageProofs,
+		t.SiafundInputs,
+		t.SiafundOutputs,
+		t.ArbitraryData,
```
lukechampine: now we can safely use `hash.HashAll`... except it doesn't exist.

This should work though; it avoids the double-interface problem:
```go
func HashAll(objs ...interface{}) Hash {
    var b[]byte
    for _, obj := range objs {
        b = append(b, encoding.Marshal(obj))
    }
    return HashBytes(b)
}
```

consensus/types.go
```diff
@@ -197,29 +222,78 @@ func (b Block) MinerPayoutID(i int) OutputID {
...
 	)))
 }
 
+// SiaClaimOutputID returns the id of the siacoin output that is created when
+// the siafund output gets spent.
+func (id OutputID) SiaClaimOutputID(i int) OutputID {
+	return OutputID(hash.HashBytes(encoding.MarshalAll(
+		id,
+	)))
+}
```
lukechampine: ??? What is `i` for?

lukechampine:
Looks like github.com/codahale/blake2 uses cgo. Just sayin'

modules/wallet/addresses.go
```diff
@@ -16,8 +16,13 @@ func (w *Wallet) timelockedCoinAddress(unlockHeight consensus.BlockHeight) (coin
...
 		TimeLock:      unlockHeight,
 		NumSignatures: 1,
-		PublicKeys:    []crypto.PublicKey{pk},
+		PublicKeys: []consensus.SiaPublicKey{
+			consensus.SiaPublicKey{
+				Algorithm: consensus.ED25519Identifier,
+			},
+		},
 	}
+	copy(spendConditions.PublicKeys[0].Key, pk[:])
```
lukechampine: why can't you write:
```go
consensus.SiaPublicKey{
    Algorithm: consensus.ED25519Identifier,
    Key:       pk[:],
}
```
?

consensus/builddev.go
DavidVorick: it's for readability.

I could do 1e4, could I also do 300e5? I guess that I could.

consensus/types.go
DavidVorick: hmm.

so many things about go are backwards

consensus/types.go
DavidVorick: the... uhh.... it's for...

wait no it's not.

DavidVorick:
is cgo a problem?

I thought it wasn't a problem as of 1.2 or something. When working with crypto you absolutely want to use the library that's been seen by the most eyes, as opposed to a port. Our ed25519 library is a port, but I pretty much consider that to be a bad thing.

consensus/signatures.go
DavidVorick: why wouldn't it work for signatures?

modules/wallet/addresses.go
DavidVorick: That would be the next step.  It doesn't really have much sense to have 'SignatureKeys' function if you're planning on supporting multiple signature types. Instead you'd have a GenerateEd25519Keys(). But even that couldn't return a SiaPublicKey, because you'd have an import cycle between consensus and crypto.

So you'd need a consensus.GenerateSignatureKeys(consensus.IdentifierEd25519)

DavidVorick:
ok, I've switch to a pure go implementation of blake2b

modules/wallet/addresses.go
DavidVorick: We can do that in another pr

consensus/types.go
DavidVorick: this is addressed though the diff doesn't show that

consensus/builddev.go
lukechampine: or 3e7...

consensus/signatures.go
lukechampine: just that the `if` statement doesn't check signatures, so you couldn't do the `for` loop thing. (Unless you resliced it, but ugh. Just leave it.)

consensus/types.go
lukechampine: hahaha

modules/wallet/addresses.go
lukechampine: `crypto` wouldn't return a `SiaPublicKey`, the `wallet` would take care of that. Anyway, as it stands it's definitely broken, but the fix is easy.

consensus/types.go
```diff
@@ -194,30 +219,79 @@ func (b Block) MerkleRoot() hash.Hash {
...
-	)))
+	))
+}
+
+// SiaClaimOutputID returns the id of the siacoin output that is created when
+// the siafund output gets spent.
+func (id OutputID) SiaClaimOutputID() OutputID {
+	return OutputID(hash.HashAll(
+		id,
+	))
```
lukechampine: are more fields going to be added to this? If not, you can just use `hash.HashObject`

consensus/builddev.go
DavidVorick: I think I meant 3e5 lol

modules/wallet/addresses.go
DavidVorick: also fixed it, diff isn't updating

DavidVorick: oh wait no just didn't push

consensus/types.go
DavidVorick: it's nice to keep the format. People don't suddenly get surprised with a new function.

Of all the files, types.go is probably what'll get read the most

Issue #157, opened 2014-12-12
Do not include signatures in the transaction id


DavidVorick:
This makes it hard to know what output ids are going to be when doing mutliparty transactions.

PR #313, opened 2015-02-03
Further testing

DavidVorick:
I change Standard.md to be an actual markdown file.

I renamed Transaction.Output and Transaction.Input to SiacoinOutput and SiacoinInput respectively, because there are siafunds as well.

I added a forking test to make sure the state rewinds and reapplies correctly (it didn't, but now it does).

I added contract testing, transaction testing, and expanded the crypto package a little bit to check for nil stuff.

I changed to using `encoding.Marshal` and `encoding.Unmarshal` instead of doing it manually. This may have introduced regressions, but only in packages that have minimal to no testing. The bugs should pop up immediately once we add testing.

The consensus testing suite is a bit of a mess right now. The functions could be more oragnized, and it could use better tooling in general. I've decided to let it be messy for now though, because the testing is now pretty thorough (74%!) and we don't need to test it to death before we return to working on the user experience. I think we can come back in a few weeks and do organizational cleanup, and improve the testing tooling.

Pretty much all that remains at this point is the 100 block limit and the siafunds, and testing for each. (We also need to test the exported functions).

### DIFF
--- a/consensus/contracts_test.go
+++ b/consensus/contracts_test.go
@@ -82,7 +82,7 @@ func contractTxn(t *testing.T, s *State, delay BlockHeight, duration BlockHeight
 	}
 	txn.Signatures = append(txn.Signatures, sig)
 	sigHash := txn.SigHash(0)
-	rawSig, err := crypto.SignBytes(sigHash, sk)
+	rawSig, err := crypto.SignHash(sigHash, sk)
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -168,7 +168,7 @@ func storageProofTxn(t *testing.T, s *State) (txn Transaction, cid ContractID) {
 	}
 	txn.Signatures = append(txn.Signatures, sig)
 	sigHash := txn.SigHash(0)
-	rawSig, err := crypto.SignBytes(sigHash, sk)
+	rawSig, err := crypto.SignHash(sigHash, sk)
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/consensus/contracts_test.go
+++ b/consensus/contracts_test.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"crypto/rand"
 	"testing"
-	"time"
 
 	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/encoding"
@@ -36,7 +35,7 @@ func contractTxn(t *testing.T, s *State, delay BlockHeight, duration BlockHeight
 	}
 
 	// Mine the block that creates the output.
-	b, err := mineTestingBlock(s.CurrentBlock().ID(), Timestamp(time.Now().Unix()), minerPayouts, nil, s.CurrentTarget())
+	b, err := mineTestingBlock(s.CurrentBlock().ID(), currentTime(), minerPayouts, nil, s.CurrentTarget())
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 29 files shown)


DavidVorick:
fixes #157

consensus/contracts_test.go
```diff
@@ -0,0 +1,331 @@
...
+	txn, cid := storageProofTxn(t, s)
+
+	// Get the contract for a later check. (Contract disappears after the block
+	// is accepted).
+	contract, exists := s.openContracts[cid]
+	if !exists {
+		t.Fatal("file contract doesn't exist in state")
+	}
+
+	b, err := mineTestingBlock(s.CurrentBlock().ID(), Timestamp(time.Now().Unix()), nullMinerPayouts(s.Height()+1), []Transaction{txn}, s.CurrentTarget())
```
lukechampine: this appears in so many places (mostly tests) that I wonder if it's worth defining a convenience function:
```go
func currentTimestamp() Timestamp {
    return Timestamp(time.Now().Unix())
}
```

crypto/signatures.go
```diff
@@ -30,11 +31,30 @@ func GenerateSignatureKeys() (sk SecretKey, pk PublicKey, err error) {
 
 // SignBytes signs a message using a secret key.
 func SignBytes(data []byte, sk SecretKey) (sig Signature, err error) {
+	if len(data) != 32 {
+		err = errors.New("will only sign 32 byte hashes!")
+		return
+	}
```
lukechampine: If it only signs hashes, why isn't it `SignBytes(h hash.Hash, sk SecretKey)`? Is there a circular dependency? Even if there is, you could make it take a `[32]byte` instead of a slice.

crypto/signatures.go
```diff
@@ -30,11 +31,30 @@ func GenerateSignatureKeys() (sk SecretKey, pk PublicKey, err error) {
...
 	return
 }
 
 // VerifyBytes uses a public key and input data to verify a signature.
+//
+// TODO: Switch VerifyBytes to also returning an error.
 func VerifyBytes(data []byte, pk PublicKey, sig Signature) bool {
+	if len(data) != 32 {
+		return false
+	}
```
lukechampine: ditto here (I think)

consensus/contracts_test.go
```diff
@@ -0,0 +1,331 @@
...
+		StorageProofs: []StorageProof{sp},
+	}
+
+	return
+}
+
+// testContractCreation adds a block with a file contract to the state and
+// checks that the contract is accepted.
+func testContractCreation(t *testing.T, s *State) {
+	txn := contractTxn(t, s, 2, 25*1000)
```
lukechampine: 25e3?

crypto/signatures.go
DavidVorick: hmm. or just a hash.Hash.

I don't think there's a circular dependency

consensus/contracts_test.go
DavidVorick: lots of this was created before we discussed that.

DavidVorick:
unless there is something urgent about one of these, I would like to implement them in my next pr. I think there would be merge conflicts otherwise.

DavidVorick:
next pr will hopefully be ready in a few hours.

DavidVorick:
nvm doin it now

crypto/signatures.go
DavidVorick: we should probably move the hash files to crypto, and the merkle files to merkle

DavidVorick:
comments addressed

lukechampine:
if I'm being nitpicky...`SignBytes`/`VerifyBytes` should be `SignHash`/`VerifyHash` now.

DavidVorick:
I'll move it over.

Eventually we're going to need to distinguish between the types of crypto that we're using, because we are planning on supporting multiple types. `SignEd25519Blake2b` or something. heh. Maybe we can come up with a better scheme than that.

DavidVorick:
Maybe signing and verifying could be properies of the key types.

Ed25519GenerateKeys()
secretKey.SignBlake2b(hash Blake2b)
publicKey.VerifyBlake2b(hash Blake2b, pk Ed25519PublicKEy, sig Ed25519Signature)

PR #314, opened 2015-02-04
128-bit Unsigned Overflow-safe Currency

lukechampine:
I'm going to need some help with most of the overflow checks here. It isn't always clear how the error should be handled.

It occurred to me that it'd be most convenient to just embed a `*big.Int` in the `Currency` type. Then you wouldn't need pass-through functions. The problem, of course, is that that the pass-through functions are what catch the overflow errors. So now we're back to manual overflow checking.
...which I still think is a reasonable approach. I mean, there are tons of integers in the codebase that could potentially overflow, and they aren't being checked at all. Even manual checking is a lot better than nothing, and it greatly increases readability. I think it would make some of the logic easier to handle too.


modules/host/contractcreation.go
```diff
@@ -66,10 +66,10 @@ func (h *Host) considerTerms(terms modules.ContractTerms) error {
 	case terms.MissedProofAddress != consensus.ZeroAddress:
 		return errors.New("burn payout needs to go to the zero address")
 
-	case terms.Price < h.Price:
+	case terms.Price.Cmp(h.Price) < 0:
 		return errors.New("price does not match host settings")
 
-	case terms.Collateral > h.Collateral:
+	case terms.Collateral.Cmp(h.Collateral) > 0:
```
lukechampine: I didn't appreciate the value of `Cmp` and `Sign` until I had to convert all of these. The return value of `Cmp` and `Sign` makes it easy because you can just move the comparison operator to the right and compare to 0.

DavidVorick:
YES!

also build failed

modules/hostdb/weightedlist.go
```diff
@@ -75,11 +75,11 @@ func (hn *hostNode) insert(entry modules.HostEntry) (nodesAdded int, newNode *ho
 // remove takes a node and removes it from the tree by climbing through the
 // list of parents. Remove does not delete nodes.
 func (hn *hostNode) remove() {
-	hn.weight -= entryWeight(hn.hostEntry)
+	hn.weight.Sub(entryWeight(hn.hostEntry))
 	hn.taken = false
 	current := hn.parent
 	for current != nil {
-		current.weight -= entryWeight(hn.hostEntry)
+		current.weight.Sub(entryWeight(hn.hostEntry))
```
lukechampine: these `Sub`s should be safe since it's always guaranteed that `currentWeight >= entryWeight(hn.hostEntry)`, right?

lukechampine:
oh right. I didn't fix the tests at all lol

DavidVorick:
the other integers are less important, because overflowing doesn't result in coins coming from nowhere.

modules/hostdb/weightedlist.go
DavidVorick: yeah. any time you have that belief though, you should write an if DEBUG { panic } statement. Just in case the logic breaks somewhere unexpected, you'll get a good warning.

consensus/blocks.go
```diff
@@ -69,18 +69,24 @@ func (s *State) checkMinerPayouts(b Block) (err error) {
...
+		if err != nil {
+			return
+		}
 	}
 
 	// Return an error if the subsidy isn't equal to the payouts.
-	if subsidy != payoutSum {
+	if subsidy.Cmp(payoutSum) != 0 {
 		err = MinerPayoutErr
 		return
```
DavidVorick: you missed this one :P

consensus/builddev.go
```diff
@@ -9,15 +9,15 @@ import (
...
-	MinimumCoinbase = Currency(30 * 1000)
+	BlockSizeLimit        = 1024 * 1024 // Blocks cannot be more than 1MB.
+	BlockFrequency        = 10          // In seconds.
+	TargetWindow          = 80          // Number of blocks to use when calculating the target.
+	MedianTimestampWindow = 11          // Number of blocks that get considered when determining if a timestamp is valid. Should be an odd number.
+	FutureThreshold       = 3 * 60 * 60 // Seconds into the future block timestamps are valid.
+	SiafundCount          = 1e4         // The total (static) number of siafunds.
+
+	InitialCoinbase = 3e5
+	MinimumCoinbase = 3e4
```
DavidVorick: can we have everything be e3/6/9/etc?

10e3
300e3
30e3

While we're at it, BlockSizeLimit can just be
1e6

consensus/buildrelease.go
```diff
@@ -9,15 +9,15 @@ import (
...
-	TargetWindow          = BlockHeight(2000)      // Number of blocks to use when calculating the target.
-	MedianTimestampWindow = 11                     // Number of blocks that get considered when determining if a timestamp is valid - should be an odd number.
-	FutureThreshold       = Timestamp(3 * 60 * 60) // Seconds into the future block timestamps are valid.
-	SiafundCount          = Currency(10 * 1000)    // The total (static) number of siafunds.
-
-	InitialCoinbase = Currency(300 * 1000)
-	MinimumCoinbase = Currency(30 * 1000)
+	BlockSizeLimit        = 1024 * 1024 // Blocks cannot be more than 1MB.
+	BlockFrequency        = 600         // In seconds.
+	TargetWindow          = 2000        // Number of blocks to use when calculating the target.
```
DavidVorick: not 2e3?

consensus/currency.go
```diff
@@ -0,0 +1,114 @@
...
+
+func BigToCurrency(b *big.Int) (c Currency, err error) {
+	if b.BitLen() > 128 || b.Sign() < 0 {
+		c.of = true
+		err = ErrOverflow
+		return
+	}
+	c.i = *b
+	return
+}
```
DavidVorick: do we still use this?

consensus/currency.go
```diff
@@ -0,0 +1,114 @@
...
+type Currency struct {
+	i  big.Int
+	of bool // has an overflow ever occurred?
+}
+
+func NewCurrency(x uint64) Currency {
+	// no possibility of error
+	c, _ := BigToCurrency(new(big.Int).SetUint64(x))
+	return c
+}
```
DavidVorick: This might not be sufficient. uint64max is a lot, lot lower than the coin count per block (300,000 * 2^80). It works for testing though, which might be most of the purpose for this function.

consensus/currency.go
```diff
@@ -0,0 +1,114 @@
...
+	c.i = *b
+	return
+}
+
+func (c *Currency) SetBig(b *big.Int) (err error) {
+	oldOF := c.of
+	*c, err = BigToCurrency(b)
+	c.of = c.of || oldOF // preserve overflow flag
+	return
+}
```
DavidVorick: It's an interesting choice that you'd preserve the overflow when calling SetBig. Can you explain that choice more?

DavidVorick: oh I see. You use it internally a lot.

Might be best not to export this function, because it's a bit of a strange behavior.

consensus/currency.go
```diff
@@ -0,0 +1,114 @@
...
+	f, _ := new(big.Rat).SetInt(&c.i).Float64()
+	rat := new(big.Rat).SetFloat64(f)
+	s, _ := BigToCurrency(new(big.Int).Div(rat.Num(), rat.Denom()))
+	s.of = c.of // preserve overflow
+	return s
+}
+
+func (c *Currency) Sign() int {
+	return c.i.Sign()
+}
```
DavidVorick: This is interesting because a currency is supposed to be unsigned.

DavidVorick: having a negative currency is equivalent to an underflow, which should probably return an error.

Then again, I don't think we ever do subtraction in the consensus package. We only start from zero and add things up.

consensus/fork.go
```diff
@@ -98,6 +98,7 @@ func (s *State) generateAndApplyDiff(bn *blockNode) (err error) {
 	s.currentBlockID = bn.block.ID()
 	s.currentPath[bn.height] = bn.block.ID()
 
+	// TODO: minerSubsidy isn't used anywhere?
```
DavidVorick: woah what?

I have testing for this, the miner fees do appear in the block subsidy. Unless I screwed that up.

Oh wait you might not have those tests in this branch.

consensus/state.go
```diff
@@ -75,10 +75,10 @@ func CreateGenesisState(genesisTime Timestamp) (s *State) {
 		Value:     CalculateCoinbase(0),
 		SpendHash: ZeroAddress,
 	}
-	s.unspentSiafundOutputs[OutputID{}] = SiafundOutput{
-		Value:            10 * 1000,
-		SpendHash:        ZeroAddress, // TODO: change to Nebulous Genesis Address
-		ClaimDestination: ZeroAddress, // TODO: change to Nebulous Genesis Address
+	s.unspentSiafundOutputs[OutputID{0}] = SiafundOutput{
+		Value:            NewCurrency(10 * 1000), // should this be SiafundCount?
```
DavidVorick: yes.

consensus/transactions.go
```diff
@@ -164,19 +165,27 @@ func (s *State) applyTransaction(t Transaction) (outputDiffs []OutputDiff, contr
...
 
 	// Add the outputs
 	for _, output := range t.Outputs {
-		sum += output.Value
+		sum.Add(output.Value)
+	}
+
+	// Check for overflow
+	if sum.Overflow() {
+		// TODO: ???
```
DavidVorick: return an error, which means you'll need to grep for OutputSum and do error checking everywhere that it's used.

DavidVorick: what is the behavior of Currency.Cmp if an overflow has occurred? Should it ignore the fact that it's only supposed to be 128 bit?

consensus/types.go
```diff
@@ -183,11 +182,11 @@ type CoveredFields struct {
...
 func CalculateCoinbase(height BlockHeight) Currency {
-	if Currency(height) >= InitialCoinbase-MinimumCoinbase {
-		return MinimumCoinbase * 100000
-	} else {
-		return (InitialCoinbase - Currency(height)) * 100000
+	base := InitialCoinbase - uint64(height)
+	if base < MinimumCoinbase {
+		base = MinimumCoinbase
 	}
+	return NewCurrency(base * 1e5)
```
DavidVorick: We need a new constant. Idk what to call it but it should be 2^80, `1 << 80`.

`return NewCurrency(base * PrecisionAugmenter)

modules/hostdb/entries.go
```diff
@@ -11,39 +9,43 @@ import (
...
+	if entry.Freeze.Sign() == 0 {
+		entry.Freeze = consensus.NewCurrency(1)
 	}
 
-	adjustedCollateral := float64(entry.Collateral)
-	adjustedFreeze := float64(entry.Freeze)
-	adjustedPrice := math.Sqrt(float64(entry.Price))
+	// weight := entry.Freeze * entry.Collateral / sqrt(entry.Price)
+	weight.Add(entry.Freeze)
+	// mathematically, overflow can only occur here
```
DavidVorick: add if DEBUG { panic } just to be sure. I just realized this is somewhat equivalent to making the program self aware.

modules/hostdb/entries.go
```diff
@@ -11,39 +9,43 @@ import (
...
+	err := weight.Mul(entry.Collateral)
+	if err != nil {
+		// TODO: ???
```
DavidVorick: Maybe set the value to consensus.CurrencyMax?

or maybe this math has to be done using a normal big.Int. We don't really care about this guy overflowing. It's not unreasonable to have a freeze around 2^90 or even as high as 2^100, and it's also not unreasonable that it would be paired with a collateral of 2^60 or 2^70.

Price will probably also be 2^60ish, but doesn't necessarily have to be. Someone I guess could offer free storage.

modules/miner/mine.go
```diff
@@ -31,7 +31,10 @@ func (m *Miner) blockForWork() (b consensus.Block) {
 	subsidy := consensus.CalculateCoinbase(height + 1)
 	for _, txn := range m.transactions {
 		for _, fee := range txn.MinerFees {
-			subsidy += fee
+			err := subsidy.Add(fee)
+			if err != nil {
+				// TODO: ???
```
DavidVorick:
```go
if consensus.DEBUG {
    if err != nil {
        panic
    }
```

modules/renter/negotiate.go
```diff
@@ -13,12 +13,14 @@ import (
...
 const (
-	// TODO: ask wallet
-	minerFee = 10
-
 	defaultWindowSize = 100
 )
 
+var (
+	// TODO: ask wallet
+	minerFee = consensus.NewCurrency(10)
```
DavidVorick: it'll be better just to prune all of these and go without fees for a while.

modules/wallet/outputs.go
```diff
@@ -39,7 +39,7 @@ type key struct {
 func (w *Wallet) findOutputs(amount consensus.Currency) (knownOutputs []*knownOutput, total consensus.Currency, err error) {
 	w.update()
 
-	if amount == consensus.Currency(0) {
+	if amount.Sign() == 0 {
 		err = errors.New("cannot fund 0 coins") // should this be an error or nil?
```
DavidVorick: wrt the comment, it should probably stay an error. We can probably delete this comment.

modules/wallet/outputs.go
```diff
@@ -90,7 +93,8 @@ func (w *Wallet) Balance(full bool) (total consensus.Currency) {
 			if !full && knownOutput.age > w.age-AgeDelay {
 				continue
 			}
-			total += knownOutput.output.Value
+			// TODO: ???
+			total.Add(knownOutput.output.Value)
```
DavidVorick: what is the `???` about?

consensus/fork.go
DavidVorick: no, what happened is there's a new mechanism for dealing with the miner fees and payouts. The validity of the payouts is checked earlier in the code (when the block is first accepted, before calling fork), thus making this code unnecessary.

I cleaned it out on my branch don't worry about it here.

DavidVorick: actually, b/c of all the currency changes...

you can delete the `minerSubsidy :=` declaration. and the ` for _, fee` loop.

consensus/builddev.go
lukechampine: seems reasonable

consensus/currency.go
lukechampine: it's used in exactly one place, I think: https://github.com/NebulousLabs/Sia/blob/128bit/modules/hostdb/hostdb.go#L176

consensus/currency.go
lukechampine: hmm. Right now vast the majority of `Currency` objects are created from `uint64`s. I suppose you could just internally multiply by some huge factor though.

consensus/currency.go
lukechampine: good point. I mean, the docstring of `Currency` says that there's no way to clear the overflow flag, and this function respects that. But there's no practical reason for it to do so.
Unexporting it is probably the way to go. I don't think it's used anywhere outside this file.

consensus/currency.go
lukechampine: true. Operations that produce a negative result will indeed produce an error. I only ever use `Sign` for checking if a `Currency` is zero, so it could probably be replaced by an `IsZero` method.

consensus/currency.go
lukechampine: On a related note, I thought about just unexporting the `Currency` type entirely. It used to use `*big.Int` instead of a `big.Int`, and it caused nil dereference panics because I was writing `var sum Currency` instead of `sum := NewCurrency(0)`.
It should be safe to write it either way now, but unexporting would give us total control over how `Currency` objects are created and manipulated.

consensus/fork.go
lukechampine: woo, one less check!

consensus/transactions.go
lukechampine: In a formal spec I would say its behavior is undefined, but as it's written now it will behave as you said, as though the overflow did not occur. That is, `Currency(2^128 + 9001).Cmp(2^128) == 1`.

consensus/types.go
lukechampine: should probably be a unit name, like Bitcoin's satoshi and Finney. `1 << 80` is going to be the standard "siacoin," right?

modules/hostdb/entries.go
lukechampine: true. So the maximum possible weight is 2^64 * 2^128 * 2^128 / 1 = 2^320, which is 40 bytes. It wouldn't be so bad to just use `*big.Int`s throughout the weighted list. Solves a lot of the overflow problems, really.

modules/wallet/outputs.go
lukechampine: just another case where I was uncertain how to handle the error.

consensus/currency.go
DavidVorick: me be better to have hostdb just be a weight, instead of a currency

consensus/currency.go
DavidVorick: I think we'll just chalk it up as a testing convenience. Most apps will be expected to use BigToCurrency

consensus/currency.go
DavidVorick: hmm.

I don't like the idea of not exporting a type that's critical to specification.

DavidVorick: oh, especially because it's part of a bunch of types that are exported. So we have to also export it.

lukechampine: I agree, it feels like all spec types should be exported.
Having `currency` fields in other exported types wouldn't be an issue though.

consensus/types.go
DavidVorick: https://github.com/NebulousLabs/Sia-UI/issues/9

I was sort of thinking that `1e12` would be one sia. `1 << 80` would be 1.2 Tsia. The first block would have 300Psia. `1 << 0` would be 1 psia.

Most people would probably talk in Tsia under this method. But it's stil confusing and could probably use some work.

consensus/currency.go
lukechampine: So maybe rename `BigToCurrency` -> `NewCurrency` and `NewCurrency` -> `NewCurrency64`?

consensus/types.go
DavidVorick: We could also make the sia be `1e24`, and 1 block would be 300 ksia. or k-sia.

modules/hostdb/entries.go
DavidVorick: said this above too, but I don't see the justification for using a Currency in hostdb, because a big.Int should be sufficient

consensus/currency.go
DavidVorick: sounds good to me

consensus/currency.go
DavidVorick: except that outside packages need to set it lol

DavidVorick: oh I guess you could just have a 'Set' and 'Get' function that takes and returns bigs

consensus/currency.go
lukechampine: or, you know, we could have a generic `NewCurrency` that takes an interface and performs a type switch...tempting :P

modules/wallet/outputs.go
DavidVorick: oh. yeah it probably is, because the wallet is getting its inputs from the user. So the wallet should check for that and return an error if it overflows.

The wallet has a few protections but I still think you could sneak an overflow through somewhere.

consensus/currency.go
lukechampine: I'm still confused. If you want to set the value, you just write something like `entry.Price = NewCurrency`. You don't have to work with the underlying `currency` type directly.

consensus/currency.go
```diff
@@ -0,0 +1,114 @@
...
+
+func (c *Currency) UnmarshalSia(b []byte) int {
+	var err error
+	*c, err = BigToCurrency(new(big.Int).SetBytes(b[:16]))
+	if err != nil {
+		return -1
+	}
+	return 16
+}
```
lukechampine: I'm not sure if it's correct to return an unmarshalling error here. You're decoding a maximum of 16 bytes, so the worst that can happen is you get a negative currency (whose overflow flag will be set). I feel like that's an error that should be dealt with later; otherwise if you receive a block with a negative currency value you won't even be able to decode it.

consensus/currency.go
DavidVorick: no

consensus/currency.go
DavidVorick: how do negative big ints get encoded?

Will potential Target values encode as negative? We should write some tests to see if the corner case encodings of our targets (namely, anything with a 1 in the most significant bit) comes out negative and screws up the math.

lukechampine: A negative `big.Int` is encoded just like any other `big.Int`. It'll decode to a negative `big.Int`, which will cause an error.
I guess `Currency` is still technically signed, which means it's only a 127-bit unsigned integer. I think I could fix this by extending it to 129 bits and dropping the leading bit though.

ACTUALLY:
When I investigated further, I found that there's actually no way to create a negative `big.Int` just using `SetBytes`. You to set the negative flag separately. See for yourself: http://play.golang.org/p/kvi76glQ1g

This is great, actually, because it means it's actually impossible to create a negative `Currency`. It's also impossible to create a "negative" `Target`, because `Target.Int` also uses `SetBytes`. So we should be safe.

modules/wallet/outputs.go
lukechampine: hmm. Is it possible to move the overflow check to wherever we're adding outputs to the wallet? I feel like if adding an output causes an overflow, it should just be rejected. Having `Balance` return an error just seems backwards.

consensus/currency.go
DavidVorick: w00t scared me

Issue #315, opened 2015-02-05
Cancellation Address


DavidVorick:
I'm thinking about adding a single field to a file contract, the cancellation field. 
This would come with a new field for transactions, the file contract cancellation field. The contract cancellation would contain a file contract id and a slice of siacoin outputs.

The cancellation field would actually be an address  that works like a normal wallet address. Presumably, the client and the host would use a multi-sig address. If the spend conditions of the address get filled, then the contract becomes void and payout is sent to the siacoin outputs in the slice. The payout must equal the the sum of the values in the slice.

This would allow people to edit the hashes by cancelling existing contracts and replacing them with new contracts. Because you can sign whole transactions, you can refuse to sign a file contract cancellation unless it's in a transaction that contains a replacement contract.

Or you can just be nice and return the customer's money, and also get out of your own obligation.

This is the type of feature we'd want to look at adding after we've got people using the system and asking for the feature. This feature would need to be a hardfork, because it involves removing things from the state. Or we could soft-fork in a whole new type of contract, a cancellable contract.

Regardless, this isn't on my immediate todo list. I'd like to see a user-friendly beta first.

DavidVorick:
Actually, this might be a really good way to test soft forking once we're launched.

Issue #308, opened 2015-02-02
Encoding/Marshalling Specification


DavidVorick:
An encoding/marshalling specification is all that remains  to finish the rough draft specification, to a level of completeness that someone else could write their own implementation and follow all the rules with no ambiguity. Everything else is ready to go.

which might seem daunting but I think most of it will be done in the next 3 days.

lukechampine:
The specification more or less exists already in the docstring of `Marshal`. I can pull it out into a separate file though.

DavidVorick:
Can you put it in Consensus.md, below the section 'Cryptographic Algorithms"?

Issue #298, opened 2015-01-30
Add Siafunds


DavidVorick:
The time has come to start thinking about how to add siafunds to consensus.

I have some ideas but a lot of them are inefficient.

DavidVorick:
The one I like the most right now is to have siafunds that mirror normal siacoins. Instead of being finitely divisible though they're just finite at 10,000. There's a global tracker that counts how many siacoins per siafund have been given out period.

Each siafund will have a number indicating how many coins that siafund had most recently received. When someone wants to do a cash out, they just issue a spend, and they'll get an output worth (global number of coins per siafund - most recent cashout value) * number of siafunds. Then the number gets updated so that it's equal to the global number of coins.

The coin output which gets created can't be spent for X blocks, where is is the number of blocks that miner subsidies and storage proofs aren't allowed to be spent (likely 108).

We should be able to fit this model into the standard way of doing diffs and state hashing that we currently have. Either the siafund thing exists or it doesn't, and when you cash out you consume the existing fund and it gets replaced with a new fund. Sending funds will move around who owns the funds.

Issue #49, opened 2014-11-12
Block subsidy spending limit


DavidVorick:
Bitcoin enforces that the block subsidy should not be spent for 100 blocks. I think that this is a good precedent to follow, but it's not currently implemented.

DavidVorick:
What makes subsidies special is that they can be reversed easily. A miner doesn't need to intentionally double spend, if there's a reorg than someone is going to be out their subsidy.

This same problem applies to storage proofs - you can invalidate storage proofs simply by reorganizing the blockchain (this will change the random seed that is used for determining which segment the host needs to prove possession of).

If we implement a spending limit for miners, it also needs to be in place for storage proofs.

DavidVorick:
And finally, sianote security also needs to work this way, because the implementation I'm thinking of will also have variable outputs depending on what fork you are on.

DavidVorick:
It might be worth making it a consensus rule that storage proofs are exclusively allowed to have miner fees and storage proofs. Otherwise the spending limit safety stuff gets pretty complicated.

PR #318, opened 2015-02-06
Siafunds + delays for output types

DavidVorick:
I'm going to go ahead and make this PR, but the build fails in a pretty bad way. The testing needs to be re-written unfortunately. Hopefully the new testing convention means that the tests will break less often.

I added siafunds, and everything now should be implemented. All that's left (heh) is testing and adding a way to nullify contracts. But I pretty much want to add contract nullifcation as a softfork just to see how it would work, and if it would work well.

I've also been increasing the amount of documentation. I feel like we're approaching a solid production codebase. Still a ways to go but it's feeling like the strongest of every project I've ever done, even of the smaller ones.


DavidVorick:
The build now passes, but at the cost of commenting out most of the testing.

Throughout the day tomorrow I'll be cleaning everything up and bringing all of the tests back. They aren't incredibly broken, it's just that the behavior they were testing for modified (miner payouts aren't available for X blocks, contract payouts lose 3.9% to siafunds, storage proof payouts aren't available for X blocks).

DavidVorick:
At this point I'm feeling good about the completeness of things. Still needed:

- encoding documentation in Consensus.md
- rename some of the variables and types
- further testing
- editable contracts

Really close to calling this 'completev0.3' though. It pretty much fills all of the original design goals with a lot of improvements on top.

DavidVorick:
Supposedly, the consensus package is complete.

All of the modules still have variable names in their data structures that don't align with the changed names to the consensus package. Consensus.md isn't complete either, it's missing FileContractTerminations. But types.go should be finished, and everything should be implemented as intended.

I guess it's an exciting moment.

DavidVorick:
Fixes #49 
Fixes #157
Fixes #298 
Fixes #308 
Fixes #315

Consensus.md
```diff
@@ -69,15 +79,15 @@ Currency
...
-asset limited to 10,000 indivisible units. Each time a contract is created,
-3.9% of the payout is put into the siafund pool. The number of siacoins in the
-siafund pool must always be divisible by 10,000; the number of coins taken from
-the payout is rounded down to the nearest 10,000.
+asset limited to 10,000 indivisible units. Each time a file contract payout is
+made, 3.9% of the payout is put into the siafund pool. The number of siacoins
+in the siafund pool must always be divisible by 10,000; the number of coins
+taken from the payout is rounded down to the nearest 10,000. The siafund is
+also represented by a 128 bit unsigned integer, even though only 16 bits of
+resolution are required.
```
lukechampine: why is this?

Consensus.md
```diff
@@ -88,6 +98,22 @@ There are future plans to enable sidechain compatibility with Sia. This would
...
+consistent algorithm for marshalling types into a set of bytes that can be
+hashed. The following rules are used for hashing:
+
+ - Integers are little-endian, and are always encoded as 8 bytes.
+ - Bools are encoded as one byte, where zero is false and one is true.
+ - Variable length types such as strings are prefaced by 8 bytes containing
+   their length.
+ - Arrays and structs are encoded as their individual elements concatenated
+   together. The ordering of the struct is determined by the struct definition.
+   There is only one way to encode each type.
```
lukechampine: hmm. I'm wondering now if this isn't the best approach. It might be better to manually implement `MarshalSia` and `UnmarshalSia` for all of the struct types. Not sure yet.

Consensus.md
DavidVorick: well, it *was* so we could just use the same Currency type we already had, but...

my comment on #316, I was thinking we should probably just use a big.Int everywhere and have no fixed amount of upwards precision.... \*ducks\*

lukechampine: I mean, why is it necessary to use more bits than needed? Is there a possibility of overflow?

README.md
```diff
@@ -88,3 +88,16 @@ Version History
...
+
+If you wish to know more about the protocol and the consensus rules, you should
+read the documentation in the following order:
+
+- Consensus.md
+- consensus/types.go
+- consensus/state.go
+
+From there, you can start reading through the logic of the consensus package,
+starting at the function AcceptBlock in consensus/blocks.go
```
lukechampine: does this information belong in the README? Right now the README is what gets distributed with the release.

consensus/blocks.go
```diff
@@ -107,13 +100,16 @@ func (s *State) validHeader(b Block) (err error) {
 		return EarlyTimestampErr
 	}
 
-	// Check that the block is not too far in the future.
-	skew := b.Timestamp - Timestamp(time.Now().Unix())
+	// Check that the block is not too far in the future. An external process
+	// will need to be responsible for resubmitting the block once it is no
+	// longer in the future.
+	skew := int(b.Timestamp) - int(Timestamp(time.Now().Unix()))
```
lukechampine: redundant cast here; could just be `int(time.Now().Unix())`. Although I don't know why you're casting to `int` anyway; `FutureThreshold` is untyped, so you could just as well use an `int64` (or a `Timestamp`, for that matter).

consensus/buildtest.go
```diff
@@ -28,4 +30,7 @@ var (
 
 	MaxAdjustmentUp   = big.NewRat(1001, 1000)
 	MaxAdjustmentDown = big.NewRat(999, 1000)
+
+	// The CoinbaseAugment should be a big.Int equal to 1 << 80.
+	CoinbaseAugment = new(big.Int).Mul(big.NewInt(1<<40), big.NewInt(1<<40))
```
lukechampine: try `new(big.Int).Lsh(big.NewInt(1), 80)`

consensus/consensus_test.go
```diff
@@ -0,0 +1,159 @@
...
+// together transactions. It's designed to be simple, and it's not very smart
+// or efficient.
+type assistant struct {
+	state  *State
+	tester *testing.T
+
+	unlockConditions UnlockConditions
+	unlockHash       UnlockHash
+	secretKey        crypto.SecretKey
+}
```
lukechampine: I would probably embed the `testing.T`

consensus/contracts.go
```diff
@@ -76,78 +71,202 @@ func (s *State) validProof(sp StorageProof) error {
...
+		// Delete the contract.
+		fc, exists := s.fileContracts[fct.ParentID]
+		// Sanity check - termination should be terminating an existing
+		// contract.
+		if !exists {
+			if DEBUG {
+				panic("file contract termination terminates a nonexisting contract")
+			} else {
+				return
+			}
```
lukechampine: `else` is unnecessary here. I think in general `DEBUG` checks shouldn't have `else` blocks.

README.md
DavidVorick: foo.

Where do you think it should go? Consensus.md?

consensus/blocks.go
DavidVorick: overflow problems.

I changed both sides at the same time. I'm still not used to the untyped constants.

consensus/consensus_test.go
DavidVorick: what would happen if I also embedded the State? Do I get all of the fields too or just the functions?

consensus/info.go
```diff
@@ -66,57 +74,108 @@ func (s *State) sortedUtxoSet() (sortedOutputs []SiacoinOutput) {
...
+	// 4.	current target
+	// 5.	current depth
+	// 6.	earliest allowed timestamp of next block
+	// 7.	current path, ordered by height.
+	// 8.	unspent siacoin outputs, sorted by id.
+	// 9.	open file contracts, sorted by id.
+	// 10.	unspent siafund outputs, sorted by id.
+	// 11.	delayed siacoin outputs, sorted by height, then sorted by id.
+
+	// Create a slice of cryptoes representing all items of interest.
```
lukechampine: lol @ `s/hash/crypto/g`

consensus/contracts.go
DavidVorick: hmm. sounds good.

consensus/info.go
DavidVorick: cryptoes is a type of cereal

consensus/signatures.go
```diff
@@ -123,44 +119,77 @@ func (s *State) validSignatures(t Transaction) (err error) {
 	}
 
 	// Create the InputSignatures object for each input.
-	sigMap := make(map[OutputID]*InputSignatures)
+	sigMap := make(map[string]*InputSignatures)
```
lukechampine: why the switch?

consensus/state.go
```diff
@@ -71,23 +92,31 @@ func CreateGenesisState(genesisTime Timestamp) (s *State) {
...
+		ClaimUnlockHash: ZeroAddress, // TODO: change to Nebulous Genesis ClaimDestination Address
 	}
 
 	return
 }
 
+// RLock will readlock the state.
+//
+// TODO: Add a safety timer which will auto-unlock if the readlock is held for
+// more than a second. (panic in debug mode)
```
lukechampine: how will this work? Seems like:
1. function calls `RLock`
2. 1 second passes; `RUnlock` is automatically called
3. function calls `RUnlock` and panics

consensus/transactions.go
```diff
@@ -53,39 +58,152 @@ func (s *State) validInput(input SiacoinInput) (err error) {
...
+
+	// Check that all siafund outputs are valid and that the siafund output sum
+	// is equal to the siafund input sum.
+	var siafundOutputSum Currency
+	for _, sfo := range t.SiafundOutputs {
+		// Check that the claimStart is set to 0. Type safety should enforce
+		// this, but check anyway.
+		if sfo.ClaimStart.Cmp(ZeroCurrency) != 0 {
+			return errors.New("invalid siafund output presented")
+		}
```
lukechampine: I believe there's an `IsZero` method for this purpose. What do you mean about type safety?

consensus/transactions.go
```diff
@@ -113,68 +236,157 @@ func (s *State) validTransaction(t Transaction) (err error) {
...
+
+	// Apply each component of the transaction. Miner fees are handled as a
+	// separate process.
+	s.applySiacoinInputs(bn, t)
+	s.applySiacoinOutputs(bn, t)
+	s.applyFileContracts(bn, t)
+	s.applyFileContractTerminations(bn, t)
+	s.applyStorageProofs(bn, t)
+	s.applySiafundInputs(bn, t)
+	s.applySiafundOutputs(bn, t)
```
lukechampine: nice

consensus/types.go
```diff
@@ -1,61 +1,52 @@
...
+
+// Specifiers that get used when determining the IDs of various elements in
+// the consensus state.
+var (
+	SpecifierSiacoinOutput                 = Specifier{'s', 'i', 'a', 'c', 'o', 'i', 'n', ' ', 'o', 'u', 't', 'p', 'u', 't'}
+	SpecifierFileContract                  = Specifier{'f', 'i', 'l', 'e', ' ', 'c', 'o', 'n', 't', 'r', 'a', 'c', 't'}
+	SpecifierFileContractTerminationPayout = Specifier{'f', 'i', 'l', 'e', ' ', 'c', 'o', 'n', 't', 'r', 'a', 'c', 't', ' ', 't'}
+	SpecifierStorageProofOutput            = Specifier{'s', 't', 'o', 'r', 'a', 'g', 'e', ' ', 'p', 'r', 'o', 'o', 'f'}
+	SpecifierSiafundOutput                 = Specifier{'s', 'i', 'a', 'f', 'u', 'n', 'd', ' ', 'o', 'u', 't', 'p', 'u', 't'}
+)
```
lukechampine: I think the documentation still refers to these as "Identifiers"

consensus/transactions.go
DavidVorick: outdated comment

consensus/state.go
DavidVorick: 1. function calls RLock, and RLock writes to a semaphore or something that tracks all of the locks in place
2. 1 second passes, RUnlock is automatically called and the semaphore is updated to show that the lock is gone.
3. function calls RUnlock, which checks the semaphore and realizes that someone screwed up. It doesn't call unlock and avoids the panic

README.md
lukechampine: Yeah, that seems best.

consensus/blocks.go
lukechampine: what does "overflow problems" mean? How does casting to a *smaller* type help?

consensus/consensus_test.go
lukechampine: I believe you would get the fields as well, since you're in the same package.

consensus/state.go
lukechampine: hmm. Doesn't seem like a perfect solution, but I suppose the debug output will be helpful.

consensus/blocks.go
DavidVorick: subtraction.

timestamp - (slightly bigger timestamp) = super big timestamp, because timestamps are unsigned. but casting it to an int means that you get a negative timestamp, which is acceptable.

except (derp), this comparison still has problems because you can submit a super huge timestamp and it'll evaluate to being really far in the past. So the timestamp checking code needs some reworking.

Consensus.md
DavidVorick: yeah someone can try to spend 10 quindrillion siafunds or whatever, and it'll overflow and evaluate to spending 0

consensus/signatures.go
DavidVorick: 3 different types of objects now have UnlockConditions, and each of these objects points to a parent with a different type of id. The signature package avoids this messyness by casting the ids to a string, resulting in a more generic validation of signatures.

consensus/blocks.go
lukechampine: honestly I would just keep them as `Timestamp`s and add a check for `b.Timestamp > Now`

consensus/signatures.go
lukechampine: yeah that was my guess. Could use a `hash.Hash` instead for clarity but whatevs

consensus/consensus_test.go
DavidVorick: too confusing to embed multiple fields?

DavidVorick: I'm not planning on fixing all of the testing in this PR. the diffs are plenty large enough :p

DavidVorick:
fixes #321 

okay. this is done now. No more commits. I promise.

Everything is implemented, organized, documented. All features are in place, and the codebase is clean. I like the names for everything. This excludes the test files, info.go, and currency.go

Currency stuff needs to be adjusted, and then all of the error checking needs to be adjusted. This will clean things up a bit. Preferably, all of the documentation and functions would fit neatly into types.go instead of needing to have a separate currency.go. There is no marshalling file but as we discussed it might be good to force the encoder to follow explicit rules for each type.

I am now proud of the consensus package. We've now got 1 week to fix everything else so that when we hackathon with Seve we can plug everything into the frontend and expect a reasonably well working product.

I don't expect to have time to write proper tests until after our hackathon with Seve, but hopefully we can get things to a state where we can make tweaks and recompile and have everything update and work together smoothly.

We've got something like 50 days until our goal of launching the currency outright.

lukechampine:
Okay, I'll merge this and then commence with a full review of the consensus package. The currency changes will probably be included in that.

PR #322, opened 2015-02-10
Unbounded Currency

lukechampine:
The `Currency` type can no longer overflow. However, its `MarshalSia` method restricts the length prefix to one byte, giving a maximum value of 2^(255 * 8) =~ 10^614. Larger values can be used internally, but will fail to encode correctly.

Also, `Currency` arithmetic will always return a new value. This means that you often have to write `sum = sum.Add(value)` instead of just `sum.Add(value)`, but it isn't a big deal. Plus, you can chain operations now, and they're 10x more readable.
This change might have introduced some sneaky bugs. When I changed the return value from an `error` to a `Currency`, all the places where we had `err = x.Add(y)` threw compiler errors. But the places where we just had `x.Add(y)` were silent, so I had to hunt them all down manually. Hopefully I got them all.

Large-scale review of the consensus package is next.

### DIFF
--- a/modules/host/contractcreation.go
+++ b/modules/host/contractcreation.go
@@ -79,12 +79,6 @@ func (h *Host) considerTerms(terms modules.ContractTerms) error {
 // verifyContract verifies that the values in the FileContract match the
 // ContractTerms agreed upon.
 func verifyContract(contract consensus.FileContract, terms modules.ContractTerms, merkleRoot crypto.Hash) error {
-	payout := terms.Price
-	err := payout.Add(terms.Collateral)
-	if err != nil {
-		return err
-	}
-
 	switch {
 	case contract.FileSize != terms.FileSize:
 		return errors.New("bad FileSize")
@@ -95,7 +89,7 @@ func verifyContract(contract consensus.FileContract, terms modules.ContractTerms
 	case contract.Expiration != terms.StartHeight+(terms.WindowSize*consensus.BlockHeight(terms.NumWindows)):
 		return errors.New("bad End")
 
-	case contract.Payout.Cmp(payout) != 0:
+	case terms.Price.Add(terms.Collateral).Cmp(contract.Payout) != 0:
 		return errors.New("bad Payout")
 
 	// TODO: reconstruct how the terms work.
@@ -119,14 +113,7 @@ func verifyContract(contract consensus.FileContract, terms modules.ContractTerms
 func (h *Host) acceptContract(txn consensus.Transaction) error {
 	contract := txn.FileContracts[0]
 	duration := uint64(contract.Expiration - contract.Start)
-
-	penalty := h.Collateral
-	penalty.Mul(consensus.NewCurrency64(contract.FileSize))
-	err := penalty.Mul(consensus.NewCurrency64(duration))
-	// TODO: move this check to a different function?
-	if err != nil {
-		return err
-	}
+	penalty := h.Collateral.Mul(consensus.NewCurrency64(contract.FileSize)).Mul(consensus.NewCurrency64(duration))
 
 	id, err := h.wallet.RegisterTransaction(txn)
 	if err != nil {
--- a/modules/host/contractcreation.go
+++ b/modules/host/contractcreation.go
@@ -113,7 +113,9 @@ func verifyContract(contract consensus.FileContract, terms modules.ContractTerms
 func (h *Host) acceptContract(txn consensus.Transaction) error {
 	contract := txn.FileContracts[0]
 	duration := uint64(contract.Expiration - contract.Start)
-	penalty := h.Collateral.Mul(consensus.NewCurrency64(contract.FileSize)).Mul(consensus.NewCurrency64(duration))
+	filesizeCost := consensus.NewCurrency64(contract.FileSize)
+	durationCost := consensus.NewCurrency64(duration)
+	penalty := h.Collateral.Mul(filesizeCost).Mul(durationCost)
 
 	id, err := h.wallet.RegisterTransaction(txn)
 	if err != nil {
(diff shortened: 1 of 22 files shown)


DavidVorick:
it's impossible to encode negative numbers right?

DavidVorick:
Also, as we start to throw up more thorough testing, anywhere we wrote x.Add(y) should get caught

DavidVorick:
Can you add docstrings to all of the functions in currency.go?

modules/host/contractcreation.go
```diff
@@ -119,14 +113,7 @@ func verifyContract(contract consensus.FileContract, terms modules.ContractTerms
...
 	duration := uint64(contract.Expiration - contract.Start)
-
-	penalty := h.Collateral
-	penalty.Mul(consensus.NewCurrency64(contract.FileSize))
-	err := penalty.Mul(consensus.NewCurrency64(duration))
-	// TODO: move this check to a different function?
-	if err != nil {
-		return err
-	}
+	penalty := h.Collateral.Mul(consensus.NewCurrency64(contract.FileSize)).Mul(consensus.NewCurrency64(duration))
```
DavidVorick: this line is too long, needs to be broken down more.

modules/hostdb/announcements.go
```diff
@@ -32,9 +32,8 @@ func findHostAnnouncements(height consensus.BlockHeight, b consensus.Block) (ann
 			}
 
 			// calculate freeze and check for sane value
-			freeze := consensus.NewCurrency64(uint64(ha.SpendConditions.Timelock - height))
-			err = freeze.Mul(t.SiacoinOutputs[ha.FreezeIndex].Value)
-			if err != nil || freeze.IsZero() {
+			freeze := consensus.NewCurrency64(uint64(ha.SpendConditions.Timelock - height)).Mul(t.SiacoinOutputs[ha.FreezeIndex].Value)
```
DavidVorick: this line is also too long

modules/renter/negotiate.go
```diff
@@ -25,30 +25,16 @@ var (
...
 	duration := terms.WindowSize * consensus.BlockHeight(terms.NumWindows)
-	fund := host.Price
-	fund.Mul(consensus.NewCurrency64(uint64(duration)))
-	err = fund.Mul(consensus.NewCurrency64(terms.FileSize))
-	if err != nil {
-		return
-	}
+
+	// Determine our portion of the payout.
+	fund := host.Price.Mul(consensus.NewCurrency64(uint64(duration))).Mul(consensus.NewCurrency64(terms.FileSize))
```
DavidVorick: too long

modules/renter/negotiate.go
```diff
@@ -25,30 +25,16 @@ var (
...
 
 	// Determine the host portion of the payout.
-	collateral := host.Collateral
-	collateral.Mul(consensus.NewCurrency64(uint64(duration)))
-	err = collateral.Mul(consensus.NewCurrency64(terms.FileSize))
-	if err != nil {
-		return
-	}
+	collateral := host.Collateral.Mul(consensus.NewCurrency64(uint64(duration))).Mul(consensus.NewCurrency64(terms.FileSize))
```
DavidVorick: too long

lukechampine:
Correct, negative Currency values cannot be encoded. They can, however, be created internally, since `NewCurrency` does not strip the sign information from the `big.Int` it is passed.

modules/host/contractcreation.go
lukechampine: hmm. I don't see a great way to shorten this. One way would be to do the Currency conversions beforehand:
```go
filesizeCost, durationCost := consensus.NewCurrency64(contract.FileSize), (consensus.NewCurrency64(duration)
penalty := h.Collateral.Mul(filesizeCost).Mul(durationCost)
```
The other way would be to split the `Mul`s apart:
```go
penalty := h.Collateral
penalty = penalty.Mul(consensus.NewCurrency64(contract.FileSize))
penalty = penalty.Mul(consensus.NewCurrency64(duration))
```
I definitely prefer the former to the latter, but `filesizeCost` and `durationCost` are awkward names.

DavidVorick:
hm. That makes we wonder if there's anywhere we do subtraction that's vulnerable.

lukechampine:
`Sub` is currently used in four places (and some commented-out tests):
- In `SplitContractPayout`, to determine the output portion. The result is guaranteed to be non-negative.
- In `FundTransaction`, to calculate the refund. After patching the bug in , the refund is also guaranteed to be non-negative.
- In the weighted list code; I haven't rigorously verified that these are all okay, but they look fine.
- In `applySiafundInputs`, to calculate the claim output. I don't know enough to verify this one; you should look it over.

DavidVorick:
I'm not as worried about the module code.

Also problematic: mul, div, and mulfloat could all take negative numbers. We also need to check that those values can't be manipulated to be negative inside of the consensus code.

DavidVorick:
for `applySiafundInputs`, claimStart is guaranteed to be less than or equal to s.siafundPool. This should actually be an `if DEBUG {` assert.

lukechampine:
From a quick grepping, `Mul`, `Div`, and `MulFloat` are all safe; their arguments are always non-negative. But we'll definitely have to be careful about using them.

modules/host/contractcreation.go
DavidVorick: I too prefer the former.

DurationSize could be added to info.go, but it's not ever used inside of the consensus package so that's the only place it would belong. I also think it's probably not worth it.

DavidVorick:
For the too long lines, it seems best to call `NewCurrency` in different lines, and then chain the muls. Just trying to keep things readable.

PR #337, opened 2015-02-16
Transaction Pool

DavidVorick:
I added/updated the transaction pool module to be consistent with the new set of consensus rules. I implemented a lot of things, perhaps too many. Then I wrote tests for all of the siacoin stuff since that's what we're most likely to run into in the wild for now.

I made a few changes to the consensus test suite and the consensus package code, but I tried to keep things as minimal as possible. Most of the changes should be to the transaction pool module.


modules/transactionpool/valid.go
```diff
@@ -0,0 +1,190 @@
...
+		inputSum = inputSum.Add(sfo.Value)
+	}
+
+	// Go through the outputs, making sure each is valid and summing the number
+	// of funds allocated.
+	var outputSum consensus.Currency
+	for _, sfo := range t.SiafundOutputs {
+		if sfo.ClaimStart.Cmp(consensus.ZeroCurrency) != 0 {
+			return errors.New("invalid siafund output presented")
+		}
```
lukechampine: Down the line I don't think we should even include the `ClaimStart` in the transmitted data. Not an immediate priority though.

DavidVorick: Ideally we wouldn't, especially because there are multiple ways to implement something like ClaimStart

PR #338, opened 2015-02-17
Consensus Tests

DavidVorick:
I brought back some of the consensus tests, and added a few more that weren't there before. Right now, only siacoin stuff is tested, none of the contract stuff or siafund stuff is tested. I did bring back some of the forking testing though, and I'm currently working on adding a good amount more of forking testing.


DavidVorick:
At this point, all of blocknode.go, blocks.go, and fork.go are tested to more or less completion. Not a lot of energy is spent actively attacking the system or probing corner cases, but for these file we've at least hit 100% coverage (minus the intentionally unreachable code)

consensus/buildtest.go
```diff
@@ -20,19 +20,18 @@ const (
 	SiafundPortion        = 0.039       // Percent of all contract payouts that go to the siafund pool.
 
 	InitialCoinbase = 300e3
-	MinimumCoinbase = 30e3
+	MinimumCoinbase = 299990
```
lukechampine: need a comment explaining this value

PR #339, opened 2015-02-17
Consensus Review

lukechampine:
whew. This was fun. I was especially nitpicky here so I'm willing to tone it down in places where you think my changes are unnecessary.

Remaining issues:
- SigHash behaves a bit differently now; instead of manually marshaling and appending the fields, it instead removes all the irrelevant information from the transaction and then marshals the transaction as a whole. I think this should be fine as long as the behavior is well-defined and nothing unexpected winds up being included. Worst case, you marshal a few empty length prefixes.
- StorageProofOutputID hashes a bool. Bools are currently marshaled as a single byte, either `1` or `0`. Again, this is fine as long as the behavior is well defined. That is, people need to know to include a `1` or `0` in the hash, not something like `"true"` or `"false"`.
- It took me a while to understand the SurpassThreshold equation. I think I understand it now, but the math is a bit awkward. There might be a better way to do things.
- There are a few places where code could be organized better. Specifically, I'm thinking of the Target/Big conversion functions, and info.go which is currently a huge mix of exported and unexported functions. It also makes me uneasy having all of the methods of a type defined across 4 separate files. What we have now isn't bad though.
- The flow of the code was pretty opaque at times. Like, what's the difference between `ValidTransaction` and `ValidTransactionComponents`? Who calls which one, and why? Do they overlap? etc.
- Diffs. Are they really necessary? I think they're helpful going backwards, but how frequently is a diff going to be cached when we're going forward? Seems like something that would only happen if you're switching back and forth between two competing chains. Even still, diffs feel more like an optimization than crucial functionality. I think diff/forking code could be a lot cleaner if we only used diffs going backwards. I might be missing something here though.

### DIFF
--- a/consensus/accept.go
+++ b/consensus/accept.go
@@ -93,8 +93,9 @@ func (s *State) validHeader(b Block) (err error) {
 	return
 }
 
-// addBlockToTree takes a block and a parent node, and adds a child node to the
-// parent containing the block. No validation is done.
+// addBlockToTree inserts a block into the blockNode tree by adding it to its
+// parent's list of children. If the new blockNode is heavier than the current
+// node, the blockchain is forked.
 func (s *State) addBlockToTree(b Block) (err error) {
 	parentNode := s.blockMap[b.ParentID]
 	newNode := parentNode.newChild(b)
--- a/consensus/accept.go
+++ b/consensus/accept.go
@@ -2,41 +2,20 @@ package consensus
 
 import (
 	"errors"
-	"sort"
 
 	"github.com/NebulousLabs/Sia/encoding"
 )
 
-// Exported Errors
 var (
-	BadBlockErr       = errors.New("block is known to be invalid.")
-	BlockKnownErr     = errors.New("block exists in block map.")
-	EarlyTimestampErr = errors.New("block timestamp is too early, block is illegal.")
-	FutureBlockErr    = errors.New("block timestamp too far in future")
-	OrphanErr         = errors.New("block has no known parent")
-	LargeBlockErr     = errors.New("block is too large to be accepted")
-	MinerPayoutErr    = errors.New("miner payout sum does not equal block subsidy")
-	MissedTargetErr   = errors.New("block does not meet target")
+	BadBlockErr        = errors.New("block is known to be invalid")
+	BlockKnownErr      = errors.New("block exists in block map")
+	EarlyTimestampErr  = errors.New("block timestamp is too early")
+	FutureTimestampErr = errors.New("block timestamp too far in future")
+	OrphanErr          = errors.New("block has no known parent")
+	LargeBlockErr      = errors.New("block is too large to be accepted")
+	MinerPayoutErr     = errors.New("miner payout sum does not equal block subsidy")
+	MissedTargetErr    = errors.New("block does not meet target")
 )
-
-// earliestChildTimestamp returns the earliest timestamp that a child node
-// can have while still being valid. See section 'Timestamp Rules' in
-// Consensus.md.
-func (bn *blockNode) earliestChildTimestamp() Timestamp {
-	// Get the previous `MedianTimestampWindow` timestamps.
-	var intTimestamps []int
-	referenceNode := bn
-	for i := 0; i < MedianTimestampWindow; i++ {
-		intTimestamps = append(intTimestamps, int(referenceNode.block.Timestamp))
-		if referenceNode.parent != nil {
-			referenceNode = referenceNode.parent
-		}
-	}
-	sort.Ints(intTimestamps)
-
-	// Return the median of the sorted timestamps.
-	return Timestamp(intTimestamps[MedianTimestampWindow/2])
-}
 
 // checkMinerPayouts verifies that the sum of all the miner payouts is equal to
 // the block subsidy (which is the coinbase + miner fees).
@@ -80,11 +59,16 @@ func (s *State) validHeader(b Block) (err error) {
 		return OrphanErr
 	}
 
-	// Check the id meets the target. This is one of the earliest checks to
+	// Check the ID meets the target. This is one of the earliest checks to
 	// enforce that blocks need to have committed to a large amount of work
 	// before being verified - a DoS protection.
 	if !b.CheckTarget(parent.target) {
 		return MissedTargetErr
+	}
+
+	// Check that the block is the correct size.
+	if len(encoding.Marshal(b)) > BlockSizeLimit {
+		return LargeBlockErr
 	}
 
 	// If timestamp is too far in the past, reject and put in bad blocks.
@@ -96,7 +80,7 @@ func (s *State) validHeader(b Block) (err error) {
 	// will need to be responsible for resubmitting the block once it is no
 	// longer in the future.
 	if b.Timestamp > CurrentTimestamp()+FutureThreshold {
-		return FutureBlockErr
+		return FutureTimestampErr
 	}
 
 	// Verify that the miner payouts sum to the total amount of fees allowed to
@@ -106,9 +90,23 @@ func (s *State) validHeader(b Block) (err error) {
 		return
 	}
 
-	// Check that the block is the correct size.
-	if len(encoding.Marshal(b)) > BlockSizeLimit {
-		return LargeBlockErr
+	return
+}
+
+// addBlockToTree takes a block and a parent node, and adds a child node to the
+// parent containing the block. No validation is done.
+func (s *State) addBlockToTree(b Block) (err error) {
+	parentNode := s.blockMap[b.ParentID]
+	newNode := parentNode.newChild(b)
+
+	// Add the node to the block map
+	s.blockMap[b.ID()] = newNode
+
+	if newNode.heavierThan(s.currentBlockNode()) {
+		err = s.forkBlockchain(newNode)
+		if err != nil {
+			return
+		}
 	}
 
 	return
--- a/consensus/accept.go
+++ b/consensus/accept.go
@@ -7,14 +7,14 @@ import (
 )
 
 var (
-	BadBlockErr        = errors.New("block is known to be invalid")
-	BlockKnownErr      = errors.New("block exists in block map")
-	EarlyTimestampErr  = errors.New("block timestamp is too early")
-	FutureTimestampErr = errors.New("block timestamp too far in future")
-	OrphanErr          = errors.New("block has no known parent")
-	LargeBlockErr      = errors.New("block is too large to be accepted")
-	MinerPayoutErr     = errors.New("miner payout sum does not equal block subsidy")
-	MissedTargetErr    = errors.New("block does not meet target")
+	ErrBadBlock        = errors.New("block is known to be invalid")
+	ErrBlockKnown      = errors.New("block exists in block map")
+	ErrEarlyTimestamp  = errors.New("block timestamp is too early")
+	ErrFutureTimestamp = errors.New("block timestamp too far in future")
+	ErrOrphan          = errors.New("block has no known parent")
+	ErrLargeBlock      = errors.New("block is too large to be accepted")
+	ErrMinerPayout     = errors.New("miner payout sum does not equal block subsidy")
+	ErrMissedTarget    = errors.New("block does not meet target")
 )
 
 // checkMinerPayouts verifies that the sum of all the miner payouts is equal to
@@ -26,7 +26,7 @@ func (s *State) checkMinerPayouts(b Block) (err error) {
 		if DEBUG {
 			panic("misuse of checkMinerPayouts - block has no known parent")
 		}
-		return OrphanErr
+		return ErrOrphan
 	}
 
 	// Find the total subsidy for the miners: coinbase + fees.
@@ -45,7 +45,7 @@ func (s *State) checkMinerPayouts(b Block) (err error) {
 
 	// Return an error if the subsidy isn't equal to the payouts.
 	if subsidy.Cmp(payoutSum) != 0 {
-		return MinerPayoutErr
+		return ErrMinerPayout
 	}
 
 	return
@@ -56,31 +56,31 @@ func (s *State) validHeader(b Block) (err error) {
 	// Grab the parent of the block.
 	parent, exists := s.blockMap[b.ParentID]
 	if !exists {
-		return OrphanErr
+		return ErrOrphan
 	}
 
 	// Check the ID meets the target. This is one of the earliest checks to
 	// enforce that blocks need to have committed to a large amount of work
 	// before being verified - a DoS protection.
 	if !b.CheckTarget(parent.target) {
-		return MissedTargetErr
+		return ErrMissedTarget
 	}
 
 	// Check that the block is the correct size.
 	if len(encoding.Marshal(b)) > BlockSizeLimit {
-		return LargeBlockErr
+		return ErrLargeBlock
 	}
 
 	// If timestamp is too far in the past, reject and put in bad blocks.
 	if parent.earliestChildTimestamp() > b.Timestamp {
-		return EarlyTimestampErr
+		return ErrEarlyTimestamp
 	}
 
 	// Check that the block is not too far in the future. An external process
 	// will need to be responsible for resubmitting the block once it is no
 	// longer in the future.
 	if b.Timestamp > CurrentTimestamp()+FutureThreshold {
-		return FutureTimestampErr
+		return ErrFutureTimestamp
 	}
 
 	// Verify that the miner payouts sum to the total amount of fees allowed to
@@ -121,11 +121,11 @@ func (s *State) AcceptBlock(b Block) (err error) {
 	// Check maps for information about the block.
 	_, exists := s.badBlocks[b.ID()]
 	if exists {
-		return BadBlockErr
+		return ErrBadBlock
 	}
 	_, exists = s.blockMap[b.ID()]
 	if exists {
-		return BlockKnownErr
+		return ErrBlockKnown
 	}
 
 	err = s.validHeader(b)
--- a/consensus/applytransaction.go
+++ b/consensus/applytransaction.go
@@ -14,7 +14,7 @@ func (s *State) applySiacoinInputs(bn *blockNode, t Transaction) {
 		}
 
 		scod := SiacoinOutputDiff{
-			New:           false,
+			Direction:     DiffRevert,
 			ID:            sci.ParentID,
 			SiacoinOutput: s.siacoinOutputs[sci.ParentID],
 		}
@@ -38,7 +38,7 @@ func (s *State) applySiacoinOutputs(bn *blockNode, t Transaction) {
 		}
 
 		scod := SiacoinOutputDiff{
-			New:           true,
+			Direction:     DiffApply,
 			ID:            scoid,
 			SiacoinOutput: sco,
 		}
@@ -62,7 +62,7 @@ func (s *State) applyFileContracts(bn *blockNode, t Transaction) {
 		}
 
 		fcd := FileContractDiff{
-			New:          true,
+			Direction:    DiffApply,
 			ID:           fcid,
 			FileContract: fc,
 		}
@@ -88,7 +88,7 @@ func (s *State) applyFileContractTerminations(bn *blockNode, t Transaction) {
 
 		// Add the diff for the deletion to the block node.
 		fcd := FileContractDiff{
-			New:          false,
+			Direction:    DiffRevert,
 			ID:           fct.ParentID,
 			FileContract: fc,
 		}
@@ -138,7 +138,7 @@ func (s *State) applyStorageProofs(bn *blockNode, t Transaction) {
 		}
 
 		fcd := FileContractDiff{
-			New:          false,
+			Direction:    DiffRevert,
 			ID:           sp.ParentID,
 			FileContract: fc,
 		}
@@ -177,7 +177,7 @@ func (s *State) applySiafundInputs(bn *blockNode, t Transaction) {
 		// Create the siafund output diff and remove the output from the
 		// consensus set.
 		sfod := SiafundOutputDiff{
-			New:           false,
+			Direction:     DiffRevert,
 			ID:            sfi.ParentID,
 			SiafundOutput: s.siafundOutputs[sfi.ParentID],
 		}
@@ -204,7 +204,7 @@ func (s *State) applySiafundOutputs(bn *blockNode, t Transaction) {
 
 		// Create and apply the diff.
 		sfod := SiafundOutputDiff{
-			New:           true,
+			Direction:     DiffApply,
 			ID:            sfoid,
 			SiafundOutput: sfo,
 		}
--- a/consensus/applytransaction.go
+++ b/consensus/applytransaction.go
@@ -120,8 +120,7 @@ func (s *State) applyStorageProofs(bn *blockNode, t Transaction) {
 
 		// Get the portion of the contract that goes into the siafund pool and
 		// add it to the siafund pool.
-		poolPortion, _ := SplitContractPayout(fc.Payout)
-		s.siafundPool = s.siafundPool.Add(poolPortion)
+		s.siafundPool = s.siafundPool.Add(fc.Payout.ContractTax())
 
 		// Add all of the outputs in the ValidProofOutputs of the contract.
 		for i, output := range fc.ValidProofOutputs {
--- a/consensus/applytransaction.go
+++ b/consensus/applytransaction.go
@@ -13,13 +13,12 @@ func (s *State) applySiacoinInputs(bn *blockNode, t Transaction) {
 			}
 		}
 
-		scod := SiacoinOutputDiff{
+		bn.siacoinOutputDiffs = append(bn.siacoinOutputDiffs, SiacoinOutputDiff{
 			Direction:     DiffRevert,
 			ID:            sci.ParentID,
 			SiacoinOutput: s.siacoinOutputs[sci.ParentID],
-		}
+		})
 		delete(s.siacoinOutputs, sci.ParentID)
-		bn.siacoinOutputDiffs = append(bn.siacoinOutputDiffs, scod)
 	}
 }
 
@@ -37,13 +36,12 @@ func (s *State) applySiacoinOutputs(bn *blockNode, t Transaction) {
 			}
 		}
 
-		scod := SiacoinOutputDiff{
+		bn.siacoinOutputDiffs = append(bn.siacoinOutputDiffs, SiacoinOutputDiff{
 			Direction:     DiffApply,
 			ID:            scoid,
 			SiacoinOutput: sco,
-		}
+		})
 		s.siacoinOutputs[scoid] = sco
-		bn.siacoinOutputDiffs = append(bn.siacoinOutputDiffs, scod)
 	}
 }
 
@@ -61,13 +59,12 @@ func (s *State) applyFileContracts(bn *blockNode, t Transaction) {
 			}
 		}
 
-		fcd := FileContractDiff{
+		bn.fileContractDiffs = append(bn.fileContractDiffs, FileContractDiff{
 			Direction:    DiffApply,
 			ID:           fcid,
 			FileContract: fc,
-		}
+		})
 		s.fileContracts[fcid] = fc
-		bn.fileContractDiffs = append(bn.fileContractDiffs, fcd)
 	}
 	return
 }
@@ -87,13 +84,12 @@ func (s *State) applyFileContractTerminations(bn *blockNode, t Transaction) {
 		}
 
 		// Add the diff for the deletion to the block node.
-		fcd := FileContractDiff{
+		bn.fileContractDiffs = append(bn.fileContractDiffs, FileContractDiff{
 			Direction:    DiffRevert,
 			ID:           fct.ParentID,
 			FileContract: fc,
-		}
+		})
 		delete(s.fileContracts, fct.ParentID)
-		bn.fileContractDiffs = append(bn.fileContractDiffs, fcd)
 
 		// Add all of the payouts to the consensus set and block node diffs.
 		for i, payout := range fct.Payouts {
@@ -137,13 +133,12 @@ func (s *State) applyStorageProofs(bn *blockNode, t Transaction) {
 			bn.delayedSiacoinOutputs[id] = output
 		}
 
-		fcd := FileContractDiff{
+		bn.fileContractDiffs = append(bn.fileContractDiffs, FileContractDiff{
 			Direction:    DiffRevert,
 			ID:           sp.ParentID,
 			FileContract: fc,
-		}
+		})
 		delete(s.fileContracts, sp.ParentID)
-		bn.fileContractDiffs = append(bn.fileContractDiffs, fcd)
 	}
 	return
 }
@@ -176,13 +171,12 @@ func (s *State) applySiafundInputs(bn *blockNode, t Transaction) {
 
 		// Create the siafund output diff and remove the output from the
 		// consensus set.
-		sfod := SiafundOutputDiff{
+		bn.siafundOutputDiffs = append(bn.siafundOutputDiffs, SiafundOutputDiff{
 			Direction:     DiffRevert,
 			ID:            sfi.ParentID,
 			SiafundOutput: s.siafundOutputs[sfi.ParentID],
-		}
+		})
 		delete(s.siafundOutputs, sfi.ParentID)
-		bn.siafundOutputDiffs = append(bn.siafundOutputDiffs, sfod)
 	}
 }
 
@@ -203,22 +197,18 @@ func (s *State) applySiafundOutputs(bn *blockNode, t Transaction) {
 		sfo.ClaimStart = s.siafundPool
 
 		// Create and apply the diff.
-		sfod := SiafundOutputDiff{
+		bn.siafundOutputDiffs = append(bn.siafundOutputDiffs, SiafundOutputDiff{
 			Direction:     DiffApply,
 			ID:            sfoid,
 			SiafundOutput: sfo,
-		}
+		})
 		s.siafundOutputs[sfoid] = sfo
-		bn.siafundOutputDiffs = append(bn.siafundOutputDiffs, sfod)
 	}
 }
 
-// applyTransaction takes a transaction and uses the contents to update the
-// state of consensus according to the contents of the transaction. The
-// transaction is assumed to be valid. A set of diffs are returned that
-// represent how the state of consensus has changed. The changes to the
-// siafundPool and the delayedSiacoinOutputs are not recorded, as they are
-// handled externally.
+// applyTransaction applies the contents of a transaction to the State. This
+// produces a set of diffs, which are stored in the blockNode containing the
+// transaction.
 func (s *State) applyTransaction(bn *blockNode, t Transaction) {
 	// Sanity check - the input transaction should be valid.
 	if DEBUG {
--- a/consensus/applytransaction.go
+++ b/consensus/applytransaction.go
@@ -116,7 +116,7 @@ func (s *State) applyStorageProofs(bn *blockNode, t Transaction) {
 
 		// Get the portion of the contract that goes into the siafund pool and
 		// add it to the siafund pool.
-		s.siafundPool = s.siafundPool.Add(fc.Payout.ContractTax())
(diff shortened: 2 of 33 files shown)


DavidVorick:
As of this comment, I haven't read anything. But to address what you've said:

SigHash: not a fan of the changes. Maybe that'll change after I actually look at them, but I don't like the idea of building a partial transaction and then hashing that instead of creating a hash by appending several parts of a transaction. It doesn't strike me as a simplification.

SurpassThreshold: only switch from old fork to new fork if new fork is heavier than the old fork by 25% of the weight of the most recent block in the old fork. Kind of a complicated concept to express, but once you get it it's pretty simple. I'm not sure there's a simpler way to apply the same safety rule, but maybe there's a simpler way to code it up or document it.

ValidTransaction checks that a transaction is valid in the current consensus set. ValidTransactionComponents checks that a transaction is valid except for the parts that depend on the consensus set. So, ValidTransactionComponents can check that all the signatures are cryptographically valid, but it can't check that all of the inputs spend existing outputs (because it might spend unconfirmed outputs). Basically, ValidTransaction checks the whole package, and ValidTransactionComponents ignores everything that might be valid if there are other unconfirmed transactions it doesn't know about.

Diffs: absolutely necessary. Or at least, the easiest way to do things. The code simplified a lot after I added diffs. We don't necessarily need to call 'applyDiffs' going forward, but we do want them for going backward - way simpler. Also very fast since you don't have to validate any signatures or do any computation. Also there are fewer map lookups.

consensus/accept.go
```diff
@@ -108,9 +90,23 @@ func (s *State) validHeader(b Block) (err error) {
...
 
-	// Check that the block is the correct size.
-	if len(encoding.Marshal(b)) > BlockSizeLimit {
-		return LargeBlockErr
+	return
+}
+
+// addBlockToTree takes a block and a parent node, and adds a child node to the
+// parent containing the block. No validation is done.
+func (s *State) addBlockToTree(b Block) (err error) {
```
DavidVorick: What is the rationale for moving this to accept.go instead of keeping it in blocknode.go?

I guess it makes more sense in accept.go than in block.go, but renaming from block.go to accept.go weakens the connection between the things like validHeader and ErrLargeBlock.

consensus/accept.go
```diff
@@ -2,43 +2,21 @@ package consensus
...
+	ErrOrphan          = errors.New("block has no known parent")
+	ErrLargeBlock      = errors.New("block is too large to be accepted")
+	ErrMinerPayout     = errors.New("miner payout sum does not equal block subsidy")
+	ErrMissedTarget    = errors.New("block does not meet target")
 )
 
-// earliestChildTimestamp returns the earliest timestamp that a child node
-// can have while still being valid. See section 'Timestamp Rules' in
-// Consensus.md.
-func (bn *blockNode) earliestChildTimestamp() Timestamp {
```
DavidVorick: I understand why you moved this. I guess I agree.

consensus/applytransaction.go
```diff
@@ -13,13 +13,12 @@ func (s *State) applySiacoinInputs(bn *blockNode, t Transaction) {
 			}
 		}
 
-		scod := SiacoinOutputDiff{
-			New:           false,
+		bn.siacoinOutputDiffs = append(bn.siacoinOutputDiffs, SiacoinOutputDiff{
```
DavidVorick: idk about this man, declaring and then appending seems like a good way to do it.

consensus/applytransaction.go
```diff
@@ -37,13 +36,12 @@ func (s *State) applySiacoinOutputs(bn *blockNode, t Transaction) {
 			}
 		}
 
-		scod := SiacoinOutputDiff{
-			New:           true,
+		bn.siacoinOutputDiffs = append(bn.siacoinOutputDiffs, SiacoinOutputDiff{
+			Direction:     DiffApply,
```
DavidVorick: ++ Direction: DiffApply is way more intuitive than the bools. Definite upgrade.

consensus/blocknode.go
```diff
@@ -44,12 +25,14 @@ type blockNode struct {
...
-	// Below are the set of diffs for the block, and a bool indicating if the
-	// diffs have been generated yet or not. Diffs are computationally
-	// expensive to generate, and so will only be generated if at some point
-	// the block is a part of the longest fork.
+	// Diffs are computationally expensive to generate, so a lazy approach is
+	// taken wherein the diffs are only generated when needed. A boolean
+	// prevents duplicate work from being performed.
+	//
+	// Note that diffsGenerated == true iff the node is in the State's
+	// currentPath; this is because diffs must be generated to apply the node.
```
DavidVorick: note quite right. diffsGenerated == true iff the node has been in the currentPath at some point.

If it's currently in the current path, diffsGenerated == true, but it could also be true if it used to be, but isn't anymore due to a fork.

consensus/blocknode.go
```diff
@@ -58,143 +41,103 @@ type blockNode struct {
 	delayedSiacoinOutputs map[SiacoinOutputID]SiacoinOutput
 }
 
-// CheckTarget returns true if the block id is lower than the target.
-func (b Block) CheckTarget(target Target) bool {
-	blockHash := b.ID()
-	return bytes.Compare(target[:], blockHash[:]) >= 0
```
DavidVorick: this is more of a meta-comment, but I wish github made it easier to see where a function was moved to. Pretty inconvenient to have to search through 2k of diffs to learn that it moved to types.go

types.go seems like a good place to put it.

consensus/blocknode.go
```diff
@@ -58,143 +41,103 @@ type blockNode struct {
...
 }
 
-// Int returns a Target as a big.Int.
-func (t Target) Int() *big.Int {
-	return new(big.Int).SetBytes(t[:])
+// heavierThan returns true if the blockNode is sufficiently heavier than
+// 'cmp', where "sufficient" is defined as:
+//
+//     (1/bn.depth) - (1/cmp.depth) > (1/cmp.target) * SurpassThreshold
+func (bn *blockNode) heavierThan(cmp *blockNode) bool {
```
DavidVorick: I'm not sure I understand the change here. Why are we now comparing two blocknodes instead of comparing the blocknode to the state? I get that it's more general, but when you are commencing a fork you definitely need to be comparing against the state.

It seems like a potential place to make mistakes. But I guess there could be some value in knowing if block 500 on fork A is heavier than block 500 on fork B, though I don't think that in practice we'll ever see forks longer than 3 blocks, and probably not something you'd explore from within the core application.

consensus/blocknode.go
```diff
@@ -58,143 +41,103 @@ type blockNode struct {
...
-// Rat returns a Target as a big.Rat.
-func (t Target) Rat() *big.Rat {
-	return new(big.Rat).SetInt(t.Int())
-}
+// timestampSlice is used for sorting.
+type timestampSlice []Timestamp
+
+func (ts timestampSlice) Len() int           { return len(ts) }
+func (ts timestampSlice) Less(i, j int) bool { return ts[i] < ts[j] }
+func (ts timestampSlice) Swap(i, j int)      { ts[i], ts[j] = ts[j], ts[i] }
```
DavidVorick: I am not a fan of the mini functions.

If that surprises you :p

consensus/blocknode.go
```diff
@@ -58,143 +41,103 @@ type blockNode struct {
...
-		if err != nil {
-			return
-		}
-	}
+	// add child to parent
+	bn.children = append(bn.children, child)
 
-	return
+	return child
 }
```
DavidVorick: You should look at this diff. GH went through so much effort to preserve the newlines in the diff that the function is pretty much entirely unreadable. You'll get 2 lines of newChild and then a big red diff where all of heavierNode was removed. Then 2 more lines, and a big red diff where all of addBlockToTree was removed.

But hey, at least you found the minimum diff, preserved those newlines that were in both the old and new.

consensus/blocknode.go
DavidVorick: Also, I think I'd rather have the timestamp type and its func defined closer to the top of the file. It's in a fairly awkward space at the moment.

Or maybe even in its own smaller file, which is a bit weird.

consensus/accept.go
```diff
@@ -108,9 +90,23 @@ func (s *State) validHeader(b Block) (err error) {
...
+
+	// Add the node to the block map
+	s.blockMap[b.ID()] = newNode
+
+	if newNode.heavierThan(s.currentBlockNode()) {
+		err = s.forkBlockchain(newNode)
+		if err != nil {
+			return
+		}
 	}
```
DavidVorick: hmmm. This function changed a decent amount.

Need to think about it.

consensus/buildtest.go
```diff
@@ -32,7 +31,7 @@ var (
 
 	CoinbaseAugment = new(big.Int).Lsh(big.NewInt(1), 80)
```
DavidVorick: oh. no more!

instead of 2^80, I want to use 10^24 (makes it easier to talk in SI notation).

consensus/complete_test.go
```diff
@@ -35,12 +35,12 @@ func currentPathCheck(t *testing.T, s *State) {
 // Then the state moves forwards to the initial starting place and verifies
 // that the state hash is the same.
 func rewindApplyCheck(t *testing.T, s *State) {
-	stateHash := s.stateHash()
+	stateHash := s.Hash()
```
DavidVorick: ah. I was going to rename this to `ConsensusSetHash`, because you're not hashing the whole state you're only hashing the consensus set.

consensus/currency.go
```diff
@@ -94,6 +95,12 @@ func (c Currency) RoundDown(n uint64) (y Currency) {
 	return
 }
 
+// ContractTax returns the amount of Currency that would be taxed from a
+// contract payout with value c.
+func (c Currency) ContractTax() Currency {
+	return c.MulFloat(SiafundPortion).RoundDown(SiafundCount)
+}
```
DavidVorick: I'd rather have ContractTax be a function of a FileContract. Also, it should be named FileContractTax, because in the scheme of cryptocurrencies contracts have a lot of meanings.

DavidVorick: If it was a function of a FileContract, it could probably just be called `Tax()`

consensus/diffs.go
```diff
@@ -27,133 +28,85 @@ type FileContractDiff struct {
...
 	SiafundOutput SiafundOutput
 }
 
-// A SiafundPool diff contains the value of the siafundPool before the block
+// A SiafundPoolDiff contains the value of the siafundPool before the block
 // was applied, and after the block was applied. When applying the diff, set
-// siafundPool to `Adjusted`. When reversing the diff, set siafundPool to
-// `Previous`.
+// siafundPool to 'Adjusted'. When reverting the diff, set siafundPool to
+// 'Previous'.
```
DavidVorick: `aw`

consensus/diffs.go
```diff
@@ -27,133 +28,85 @@ type FileContractDiff struct {
...
-			if exists {
-				panic("rogue new output in applyOutputDiff")
-			}
+// commitSiacoinOutputDiff applies or reverts a SiacoinOutputDiff.
+func (s *State) commitSiacoinOutputDiff(scod SiacoinOutputDiff, dir DiffDirection) {
+	// Sanity check - should not be adding an output twice, or deleting an
+	// output that does not exist.
+	if DEBUG {
+		_, exists := s.siacoinOutputs[scod.ID]
+		if exists == (scod.Direction == dir) {
```
DavidVorick: eek lol. I guess within a DEBUG statement it's okay to have more abstract code. It does simplify things a bit.

consensus/diffs.go
```diff
@@ -27,133 +28,85 @@ type FileContractDiff struct {
...
+			panic("rogue output in commitSiacoinOutputDiff")
```
DavidVorick: rogue siacoin output

while we're at it. Not terribly important because the location of the panic gives you enough context even without any words.

consensus/diffs.go
```diff
@@ -27,133 +28,85 @@ type FileContractDiff struct {
...
-				panic("rogue new contract in applyContractDiff")
-			}
+// commitFileContractDiff applies or reverts a FileContractDiff.
+func (s *State) commitFileContractDiff(fcd FileContractDiff, dir DiffDirection) {
+	// Sanity check - should not be adding a contract twice, or deleting a
+	// contract that does not exist.
+	if DEBUG {
+		_, exists := s.fileContracts[fcd.ID]
+		if exists == (fcd.Direction == dir) {
+			panic("rogue contract in commitFileContractDiff")
```
DavidVorick: rogue file contract. At some point I decided that merely saying 'contract' wasn't good enough, because although file contracts are the only types of contracts we support, a contract is a much much more general term. I'd rather try and be religious about saying 'file contract' everywhere.

consensus/diffs.go
```diff
@@ -27,133 +28,85 @@ type FileContractDiff struct {
...
-				panic("rogue new output in applyOutputDiff")
-			}
+// commitSiafundOutputDiff applies or reverts a SiafundOutputDiff.
+func (s *State) commitSiafundOutputDiff(sfod SiafundOutputDiff, dir DiffDirection) {
+	// Sanity check - should not be adding an output twice, or deleting an
+	// output that does not exist.
+	if DEBUG {
+		_, exists := s.siafundOutputs[sfod.ID]
+		if exists == (sfod.Direction == dir) {
+			panic("rogue output in commitSiafundOutputDiff")
```
DavidVorick: siafund output. This one I also try and be religious because 'output' all by itself typically refers to a siacoin output. Not too terrible to see 'output' all by itself if it's referring to a siacoin output but worse if it's referring to a siafund output.

consensus/fork.go
```diff
@@ -4,119 +4,115 @@ import (
...
 )
 
-// backtrackToBlockchain returns a list of nodes that go from the current node
-// to the first common parent. The common parent will be the final node in the
-// slice.
-func (s *State) backtrackToBlockchain(bn *blockNode) (nodes []*blockNode) {
-	nodes = append(nodes, bn)
+// invalidateNode recursively moves the children of a blockNode to the bad
+// blocks list.
+func (s *State) invalidateNode(node *blockNode) {
```
DavidVorick: Maybe this makes more sense somewhere else. I believe it's only ever called from diffs.go, might be worth putting in blocknode.go

Also, if it's not hard to implement, badBlocks should probably only contain the first parent, instead of recursively containing the parent and all the children.  All of the nodes need to be deleted, but only the first node needs to be added to badBlocks.

It's not a huge tragedy to put all of the nodes in badBlocks though.

consensus/fork.go
```diff
@@ -4,119 +4,115 @@ import (
...
 		if bn == nil {
 			if DEBUG {
 				panic("backtrack hit a nil node?")
 			}
-			return
+			break
 		}
 	}
-	return
+	return path
```
DavidVorick: path confused me as a name, I got confused with CurrentPath.

Maybe that's b/c it's 7am and I haven't gone to bed yet.

consensus/fork.go
```diff
@@ -4,119 +4,115 @@ import (
...
+// backtrackToBlockchain traces backwards from 'bn' until it reaches a node in
+// the State's current path (the "common parent"). It returns the (inclusive)
+// set of nodes between the common parent and 'bn', starting from the former.
+func (s *State) backtrackToBlockchain(bn *blockNode) []*blockNode {
+	path := []*blockNode{bn}
 	for s.currentPath[bn.height] != bn.block.ID() {
 		bn = bn.parent
-		nodes = append(nodes, bn)
+		// prepend, not append
+		path = append([]*blockNode{bn}, path...)
```
DavidVorick: hmm.

Do me a favor and closely double check everywhere (grep style) that calls backtrackToBlockchain to make sure that you adjusted everything since you reversed the order that the blocks get returned.

DavidVorick: I'm not confident that our current test suite would catch this error. Some parts are pretty thorough, some parts are pretty sparse. No parts are exceptionally thorough.

consensus/fork.go
```diff
@@ -4,119 +4,115 @@ import (
...
-// the list of nodes that got rewound.
-func (s *State) rewindToNode(bn *blockNode) (rewoundNodes []*blockNode) {
+// rewindToNode will rewind blocks from the State's current path until 'bn' is
+// the current block.
+func (s *State) rewindToNode(bn *blockNode) {
 	// Sanity check - make sure that bn is in the currentPath.
 	if DEBUG {
-		if bn.block.ID() != s.currentPath[bn.height] {
-			panic("bad use of rewindToNode")
+		if s.currentPath[bn.height] != bn.block.ID() {
```
DavidVorick: interesting that you would switch up the order.

consensus/fork.go
```diff
@@ -4,119 +4,115 @@ import (
...
+			break
 		}
 	}
-	return
+	return path
 }
 
-// rewindToNode will rewind blocks until `bn` is the highest block, returning
-// the list of nodes that got rewound.
-func (s *State) rewindToNode(bn *blockNode) (rewoundNodes []*blockNode) {
```
DavidVorick: we don't return rewoundNodes anymore?

0.o

consensus/fork.go
```diff
@@ -4,119 +4,115 @@ import (
...
 }
 
-// invalidateNode recursively deletes all the generational children of a block
-// and puts them all on the bad blocks list.
-func (s *State) invalidateNode(node *blockNode) {
-	for i := range node.children {
-		s.invalidateNode(node.children[i])
+// applyUntilNode will successively apply the blocks between the state's
+// currentPath and 'bn'.
+func (s *State) applyUntilNode(bn *blockNode) (err error) {
```
DavidVorick: black magic

consensus/diffs.go
```diff
@@ -174,39 +127,27 @@ func (s *State) applyDiffSet(bn *blockNode, applied bool) {
...
-		for _, scod := range bn.siacoinOutputDiffs {
-			s.commitSiacoinOutputDiff(scod, applied)
-		}
-		for _, fcd := range bn.fileContractDiffs {
-			s.commitFileContractDiff(fcd, applied)
-		}
-		for _, sfod := range bn.siafundOutputDiffs {
-			s.commitSiafundOutputDiff(sfod, applied)
-		}
+	// apply each of the diffs
```
DavidVorick: comment style mismatch. All of my comments are approx. complete sentences.

// Like this.

consensus/fork.go
```diff
@@ -4,119 +4,115 @@ import (
...
 
-		// If the diffs for this node have already been generated, apply diffs
-		// directly instead of generating them. This is much faster.
-		if bn.diffsGenerated {
-			direction := true // the blockNode is being applied, direction is set to true.
-			s.applyDiffSet(bn, direction)
-			continue
-		}
+	// rewind to the common parent
+	s.rewindToNode(commonParent)
```
DavidVorick: you can probably squish these 3 statements into one code block:

```go
// Something covering everything.
oldHead := s.currentBlockNode()
commonParent := s.backtrackToCurrentPath(newNode) // See what I did there?
s.rewindToNode(commonParent)
```

DavidVorick: I do think that backtrackToBlockchain should get renamed to backtrackToCurrentPath.

consensus/fork.go
```diff
@@ -4,119 +4,115 @@ import (
...
+	err = s.applyUntilNode(newNode)
+	if err != nil {
+		// restore old path
+		s.rewindToNode(commonParent)
+		errReapply := s.applyUntilNode(oldHead)
+		if DEBUG {
+			if errReapply != nil {
+				panic("couldn't reapply previously applied diffs!")
+			} else if s.Hash() != oldHash {
+				panic("state hash changed after an unsuccessful fork attempt")
```
DavidVorick: this if DEBUG really uglies things up. Maybe there's a better approach.

idk though don't have any great ideas besides suggesting that we just not check errReapply. I would really like to keep the s.Hash() check

consensus/maintenance.go
```diff
@@ -24,9 +24,9 @@ func (s *State) applyMinerSubsidy(bn *blockNode) {
 	return
 }
 
-// applyDelayedSiacoinOutputMaintenance goes through all of the outputs that
+// applyMaturedSiacoinOutputs goes through all of the outputs that
 // have matured and adds them to the list of siacoinOutputs.
-func (s *State) applyDelayedSiacoinOutputMaintenance(bn *blockNode) {
+func (s *State) applyMaturedSiacoinOutputs(bn *blockNode) {
```
DavidVorick: approve

consensus/maintenance.go
```diff
@@ -96,14 +95,13 @@ func (s *State) applyContractMaintenance(bn *blockNode) {
 	// Expiring a contract deletes it from the map we are iterating through, so
 	// we need to store it and deleted once we're done iterating through the
 	// map.
+	currentHeight := s.height()
 	var expiredContracts []FileContractID
 	for id, contract := range s.fileContracts {
-		if s.height() == contract.Expiration {
+		if contract.Expiration == currentHeight {
```
DavidVorick: please rename from `contract` to `fc`.

I know it was like this when you found it just asking for a rename.

consensus/maintenance.go
```diff
@@ -96,14 +95,13 @@ func (s *State) applyContractMaintenance(bn *blockNode) {
 	// Expiring a contract deletes it from the map we are iterating through, so
 	// we need to store it and deleted once we're done iterating through the
 	// map.
+	currentHeight := s.height()
 	var expiredContracts []FileContractID
```
DavidVorick: also `expiredFileContracts`

consensus/signatures.go
```diff
@@ -12,220 +12,181 @@ var (
...
 		if elem <= biggest {
-			err = errors.New("covered fields sorting violation")
-			return
+			return false
 		}
 	}
-	if int(biggest) > biggestAllowed {
-		err = errors.New("covered fields indexing violation")
-		return
+	if biggest >= uint64(len(elems)) {
```
DavidVorick: not going to work. elems could potentially be a lot smaller in length than biggestAllowed.

For example, there could be 12 inputs, but only 4 of them are signed. len(elems) == 4, but biggestAllowed = 11.

consensus/signatures.go
```diff
@@ -12,220 +12,181 @@ var (
...
-			return
-		}
-		err = sortedUnique(cf.Signatures, len(cf.Signatures)-1)
-		if err != nil {
-			return
+		for _, field := range fields {
+			if !sortedUnique(field) {
+				return errors.New("field does not satisfy 'sorted and unique' requirement")
+			}
 		}
```
DavidVorick: you need to call sortedUnique on the siagnatures too, right now you only call it on the other items.

signatures should not be checked in the WholeTransaction loop, but it should be checked in the sortedUnique loop. Maybe append it after the first loop?

consensus/state.go
```diff
@@ -4,53 +4,45 @@ import (
...
+	// valid blocks, including those not on the main blockchain. badBlocks
+	// is a "blacklist" of blocks known to be invalid.
 	blockMap  map[BlockID]*blockNode
+	badBlocks map[BlockID]struct{}
 
 	// currentPath and currentBlockID track which blocks are currently accepted
 	// as the longest known blockchain.
-	currentBlockID BlockID
 	currentPath    map[BlockHeight]BlockID
+	currentBlockID BlockID
```
DavidVorick: currentPath and currentBlockID are really parts of the consensus set.

Though, on a super pedantic level some people might disagree with me.

consensus/types.go
```diff
@@ -56,10 +86,15 @@ type Block struct {
...
-// be self-dependent.
+// A Transaction is an atomic component of a block. Transactions can contain
+// inputs and outputs, file contracts, storage proofs, and even arbitrary
+// data. They can also contain signatures to prove that a given party has
+// approved the transaction, or at least a particular subset of it.
+//
+// Transactions can depend on other previous transactions in the same block,
+// but transactions cannot spend outputs that they create or otherwise be
+// self-dependent. For a more in-depth description of transactions, refer to
+// the Sia whitepaper.
```
DavidVorick: sia whitepaper might be inaccurate in that regard

consensus/types.go
```diff
@@ -77,53 +112,43 @@ type Transaction struct {
...
-// 'Start' and 'Expiration', then the outputs for 'ValidProofOutputs' get created. If
-// the party does not submit a storage proof by the contract expiration, then
-// the outputs for 'MissedProofOutputs' get created. The value of the valid
-// proof outputs must sum to the payout after the fee has been applied. The
-// value of the missed proof outputs must sum to the payout (no fee is applied
-// for missed storage proofs).
+// A FileContract is a public record of a storage agreement between a "host"
+// and a "renter." It mandates that a host must submit a storage proof to the
+// network, proving that they still possess the file they have agreed to
+// store. See the Sia whitepaper for more details on storage proofs.
```
DavidVorick: I'm still nervous about referring people to a whitepaper.

consensus/types.go
```diff
@@ -210,71 +221,68 @@ type SiafundOutput struct {
...
+// 'ParentID', which specifies the UnlockConditions, and then
+// 'PublicKeyIndex', which indicates the key in the UnlockConditions. There
+// are three types that use UnlockConditions: SiacoinInputs, SiafundInputs,
+// and FileContractTerminations. Each of these types also references a
+// ParentID, and this is the hash that 'ParentID' must match. The 'Timelock'
+// prevents the signature from being used until a certain height.
+// 'CoveredFields' indicates which parts of the transaction are being signed;
+// see CoveredFields.
+//
+// TODO: use an enum here?
```
DavidVorick: probably not use an enum, since this is a type that's going over the wire and needs to be clearly defined.

consensus/types.go
```diff
@@ -450,73 +498,57 @@ func (t Transaction) SiafundOutputID(i int) SiafundOutputID {
...
-		}
-		for _, minerFee := range cf.MinerFees {
-			signedData = append(signedData, encoding.Marshal(t.MinerFees[minerFee])...)
-		}
-		for _, arbData := range cf.ArbitraryData {
-			signedData = append(signedData, encoding.Marshal(t.ArbitraryData[arbData])...)
-		}
+	// if WholeTransaction flag is set, hash everything but the signatures.
+	if t.Signatures[i].CoveredFields.WholeTransaction {
+		t.Signatures = []TransactionSignature{}
```
DavidVorick: We want to be able to sign signatures, it shouldn't be required to be empty if WholeTransaction is set.

consensus/types.go
```diff
@@ -450,73 +498,57 @@ func (t Transaction) SiafundOutputID(i int) SiafundOutputID {
...
+	}
+	for _, i := range cf.ArbitraryData {
+		maskedTxn.ArbitraryData = append(maskedTxn.ArbitraryData, t.ArbitraryData[i])
+	}
+	for _, i := range cf.Signatures {
+		maskedTxn.Signatures = append(maskedTxn.Signatures, t.Signatures[i])
 	}
 
-	return crypto.HashBytes(signedData)
+	return crypto.HashObject(maskedTxn)
```
DavidVorick: I don't like the changes made to SigHash.

DavidVorick:
I sort of wore out near the end because it's almost 8am, but I think I was pretty thorough. I'll probably look it over entirely again once you've addressed all the comments because consensus code is serious business and deserves multiple lookovers.

consensus/accept.go
lukechampine: yeah, I think it would make sense in either place, but I moved it here because it was only called by `AcceptBlock`. That way, all the functions called by `AcceptBlock` are defined right next to it.

consensus/diffs.go
lukechampine: lol. I'm not sure which I prefer, actually. \`\` makes some sense because it's what we use on github to refer to code, but '' was more prevalent in the codebase.
I think in the Go stdlib they don't use either, they just write the variable name like any other word.

consensus/applytransaction.go
lukechampine: woo. Only downside was that it made some of the boolean logic more verbose (e.g. the `commitDiff` functions which now use `if scod.Direction == dir`; not terribly clear, but it wasn't in the first place).

consensus/blocknode.go
lukechampine: gotcha. So just `if` then, not `iff`.

consensus/blocknode.go
lukechampine: doesn't seem feasible since the diffing algorithm is pretty dumb.
It'd be cool to have language-specific diffs someday. I'm sure you could make them way more readable if you were actually parsing the syntax.

consensus/blocknode.go
lukechampine: hmm. I guess it's a product of my FP instincts. Methods of the State can have side effects on the State. (In this case, the State is only being read, not modified -- but there's no way to know that without reading the function body.) In the new version, it's clear that only blockNodes are involved. Really, this function could take objects, not pointers, to make it even clearer that the nodes are not being modified.

consensus/blocknode.go
lukechampine: yeah, I never got around to finding a good place for this. This file is the only place it's used, so top of the file is probably best.

I will note that the stdlib uses this style internally: http://golang.org/src/sort/sort.go?s=5371:5390#L223

consensus/blocknode.go
lukechampine: wrt to "It seems like a potential place to make mistakes," I wish there was a way to indicate that function `foo` is *only* intended to be a helper function of function `bar`, so that `foo` could not be called outside of `bar`.
This can already be done by defining `foo` inside of `bar`, but the result is much less clean (and probably less efficient as well, since IIRC the compiler has to assume that `foo` may close around variables in `bar`).

consensus/blocknode.go
lukechampine: lol

consensus/accept.go
lukechampine: I'm reasonably confident that the forking code is correct, given that it is passing the new tests. It actually failed after I merged your tests, and I had to go back and fix a bug. Good thing they were merged before these changes.

consensus/buildtest.go
lukechampine: okay, we have two options here then:
```go
CoinbaseAugment = new(big.Int).Exp(big.NewInt(10), big.NewInt(24), nil)
```
or:
```go
CoinbaseAugment, _ = new(big.Int).SetString("1000000000000000000000000", 10) // 10^24
```
They're both pretty ugly. I don't like the extra `nil` on the end of `Exp`, but it's probably better than using a string.

consensus/complete_test.go
lukechampine: ah, that's an important distinction. I think it's mostly implied, but I'd be fine renaming this to `ConsensusHash`.

consensus/currency.go
lukechampine: yeah, that makes more sense. This is an improvement everywhere except for renter/negotiate.go, because it needs to calculate the tax before creating the contract. So that function got slightly more awkward.

consensus/diffs.go
lukechampine: also the panic already says it's from `commitSiacoinOutputDiff`...

consensus/diffs.go
lukechampine: seems reasonable. There are a few docstrings where I'm more loose with it; I'll say "File Contract" the first time, and then just "contract" afterwards. Ditto with "SiacoinOutput"/"SiafundOutput" and just "output."

consensus/fork.go
lukechampine: hmm. `applyTransaction` and `applyMaintenance` are also only called in diffs.go, but I don't think they belong there. Maybe move `invalidateNode` to maintenance.go?

consensus/blocknode.go
DavidVorick: *not quite right

consensus/accept.go
DavidVorick: sweet I love when testing complains about actual bugs

consensus/buildtest.go
DavidVorick: it seems though that it  was meant to be used this way. I  think nil is a pretty common part of using Int.Exp

I'd much rather have exp(10, 24, nil) than Set("10000000howmany?000")

consensus/complete_test.go
DavidVorick: but it's the consensus set that's being hashed :P

consensus/currency.go
DavidVorick: I would think it could it could just add the outputs after it makes the file contract, that's how transactions work. You need to fill out the signature, call CoveredFields, and then sign the transaction

consensus/fork.go
DavidVorick: I was thinking accept.go, since that's where most of the block related stuff happens. This is a hard one.

consensus/fork.go
lukechampine: yeah, I knew this one might cause some problems. It looks like only `BlocksSince` uses it, so I updated it as necessary. I'm not sure if I correctly preserved the ordering of blocks that `BlocksSince` returns though; they may be "backwards."

consensus/fork.go
lukechampine: I tend to write these comparisons as I would say them out loud. "If the State doesn't match the blockNode" as opposed to "If the blockNode doesn't match the State". This particular instance is pretty ambiguous though; I don't think it was worth reordering.

consensus/fork.go
lukechampine: yeah, it's kind of unclear how the State "knows" which blocks to apply. :/

consensus/fork.go
lukechampine: lol @ `s.currentBlockNode()`, clearly I wasn't thinking straight.

consensus/fork.go
lukechampine: ugh I know, it's pretty gross.
`errReapply` is actually what was causing my forking code to fail tests. I was reusing the `err` variable, so it would get cleared to `nil` before being returned, even if there actually was an error!

consensus/signatures.go
lukechampine: If that's the case, then why is `sortedUnique` always called with `biggestAllowed = len(elems)`? This code is functionally equivalent. Are you planning to call `sortedUnique` with a different value for `biggestAllowed`?

consensus/state.go
lukechampine: Right, because isn't the currentPath determined by the value of `SurpassThreshold`?

consensus/types.go
lukechampine: I can just remove these for now.

consensus/fork.go
DavidVorick: I'd be infavor of

```go
if err == nil {
    return
}
// more
```

consensus/applytransaction.go
lukechampine: my habit is to not declare new variables if it can be avoided (except in multi-stage computations). Makes it more clear that you aren't going to use `scod` later. Just like:
```go
encBlock := encoding.Marshal(b)
if len(encBlock) > MaxSize {
```
vs.
```go
if len(encoding.Marshal(b)) > MaxSize {
```
I will admit it's a bit uglier in this case though.

lukechampine:
Oh right, diffs.
I agree that diffs going backwards are a good idea. Mostly I'm wondering if the code would be significantly cleaner if diffs could *only* go backwards. We might be able to remove a lot of the confusing backwards/forwards/apply/revert code.

consensus/signatures.go
DavidVorick: I think I just implemented it wrong the first time around. This needs to be changed so that biggestAllowed == len (t.Item), and len(elems) == len(t.Sig.CF.Item)

DavidVorick:
Hmm. We might be able to clean it up a bit, but not that much because diffs still have a direction. You just get to drop the lines where you check if you invert the direction or not, which ultimately I think is only 30-50loc anyway.

You would get to drop the complexity where applying a block doesn't always follow the same codepath, which simplifies some of the testing, but overall (including testing) I think we're only looking at about 300loc to keep it around, which I think is worthwhile for the performance tradeoff.

but maybe not b/c I'm not sure how often you'd actually be reapplying diffs you generated.

DavidVorick:
This looks good except for the biggestAllowed stuff in the signatures file.

PR #341, opened 2015-02-17
Add make cover to travis

DavidVorick:
When looking at travis builds, we'll get a good idea of our test coverage. It just puts that statistic somewhere accessible.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,4 +7,4 @@ install:
   - make dependencies
   - make
 
-script: make test-long
+script: make test-long && make cover


lukechampine:
I recommend we do something like this instead: http://mnapoli.fr/code-coverage-reports-with-github-travis-and/

DavidVorick:
lol it'll probably be around 30%. Not sure if I'm ready to add that button.

DavidVorick:
https://github.com/mattn/goveralls

DavidVorick:
I don't think it's currently worth taking the time to set up since we've already got something in place.

PR #344, opened 2015-02-18
Get wallet working

DavidVorick:
This PR adds some tests and fixes a small handful of bugs. The tests are not very thorough, but enough to catch the major errors that we were having.

Over the next day or two I plan on expanding out the wallet tests a bit more.


consensus/testconsistency.go
```diff
@@ -0,0 +1,177 @@
+package consensus
+
+import (
+	"sort"
+
+	"github.com/NebulousLabs/Sia/crypto"
+)
+
+// TODO: rename stateHash and StateHash to consensusSetHash and
+// ConsensusSetHash.
```
lukechampine: looks like that's TODONE.

PR #345, opened 2015-02-19
File Contracts

DavidVorick:
this PR adds some test coverage for the file contract stuff (submitting file contracts, submitting storage proofs, missing storage proofs).


DavidVorick:
I changed how wallet transactions work a bit, now it's safe to make file contract transactions, and they won't be dependent on each other for refunds.

This works by making a financial transaction first that creates an output of the exact desired size. The resulting transaction has a dependency on the transaction that creates the exact output, but at least the transaction being built won't need to be finished/spent to give the wallet access to the refund.

Also there's a fix for some things that were happening in the transaction pool.

modules/transactionpool/update_test.go
```diff
@@ -41,6 +43,38 @@ func (tpt *tpoolTester) testUpdateTransactionRemoval() {
...
+	}
+	if len(tpt.transactionPool.fileContractTerminations) != 0 {
+		tpt.assistant.Tester.Error("a field wasn't properly emptied out")
+	}
+	if len(tpt.transactionPool.storageProofs) != 0 {
+		tpt.assistant.Tester.Error("a field wasn't properly emptied out")
+	}
+	if len(tpt.transactionPool.usedSiafundOutputs) != 0 {
+		tpt.assistant.Tester.Error("a field wasn't properly emptied out")
+	}
```
lukechampine: blegh. should probably be one `if` statement if you're going to use a generic error message.

modules/wallet/transactions_test.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	wt.wallet.update()
+
+	// Check that the length of the created transaction is 1 siacoin, and that
+	// the unconfirmed balance of the wallet is 1.
+	if len(t.SiacoinOutputs) != 0 {
+		wt.assistant.Tester.Error("more than zero siacoin outputs created in custom transaction")
+	}
+	if wt.wallet.Balance(true).Cmp(consensus.NewCurrency64(1)) != 0 {
+		wt.assistant.Tester.Error(wt.wallet.Balance(true).MarshalJSON())
```
lukechampine: Marshaling may not be necessary. I think `Error` should be able to print the struct directly.

PR #348, opened 2015-02-20
Gateway

lukechampine:
This PR implements the gateway module. The gateway manages the interaction between the local node and the rest of the network. Specifically, it relays blocks and transactions, and maintains a list of active peers.

Bootstrapping and synchronizing are now the responsibility of the gateway. The Miner will send its blocks to the Gateway (in addition to the State), and the Wallet will send transactions to the Gateway (in addition to the TransactionPool).

One potential design problem occurred to me: eventually we'll want to blacklist peers if they abuse things like the network timeout. However, the network package is in charge of monitoring those abuses. So the network package needs a way of notifying the Gateway that a peer is behaving badly. Maybe it could return a special error?


modules/gateway.go
```diff
@@ -0,0 +1,54 @@
...
+	// RelayBlock announces a block to all of the Gateway's known peers.
+	RelayBlock(consensus.Block) error
+
+	// RelayTransaction announces a transaction to all of the Gateway's
+	// known peers.
+	RelayTransaction(consensus.Transaction) error
+
+	// AddMe is the RPC version of AddPeer. It is assumed that the supplied
+	// peer is the peer making the RPC.
+	AddMe(network.Address) error
```
DavidVorick: hmmm.

Sybil attacks? How do we use this?

modules/gateway.go
```diff
@@ -0,0 +1,54 @@
...
+
+	// SendBlocks is an RPC that returns a set of sequential blocks following
+	// the most recent known block ID in of the 32 IDs provided. The number of
+	// blocks returned is unspecified.
+	SendBlocks([32]consensus.BlockID) ([]consensus.Block, error)
+
+	// SharePeers is an RPC that returns a set of the Gateway's peers. The
+	// number of peers returned is unspecified.
+	SharePeers() ([]network.Address, error)
```
DavidVorick: Also sybil attack fears.

I'm not sure the best way to handle these, but trying to get a peer in every /16 seems to be Bitcoin's idea.

Not worth worrying about too much.

modules/gateway/gateway.go
```diff
@@ -0,0 +1,133 @@
...
+var (
+	ErrNoPeers     = errors.New("no peers")
+	ErrUnreachable = errors.New("peer did not respond to ping")
+)
+
+// Gateway implements the modules.Gateway interface.
+type Gateway struct {
+	tcps  *network.TCPServer
+	state *consensus.State
+	tpool modules.TransactionPool
```
DavidVorick: do you think that we should embed these?

modules/gateway.go
lukechampine: The old version took a `net.Conn` to confirm that the supplied hostname matched the connecting IP. Easy enough to restore that code, I just switched because it was ugly and I couldn't think of a potential attack off the top of my head. What's the incentive to call `AddMe` with an address other than yourself?

modules/gateway.go
lukechampine: What's the attack here?

modules/gateway/gateway.go
lukechampine: I don't see any reason to. The same reasoning would apply to the other modules.

siad/network.go
```diff
@@ -1,150 +1,35 @@
...
-		}
-		// log error
+	// TODO: add more bootstrap peers.
+	err := d.gateway.Bootstrap(BootstrapPeers[0])
+	if err != nil {
+		println("Warning: no peers responded to bootstrap request.")
+		println("Add peers manually to enable bootstrapping.")
+		// TODO: wait for new peers?
 		return
 	}
```
DavidVorick: pls no println if it's not debugging.

fmt.Println

modules/gateway.go
DavidVorick: Millions of samedude calls AddMe from different IP addresses, and you'll pretty much always select him b/c he's 99% of your addr book.

DavidVorick: not that millions of ip addresses are easy to get, but IPv6 might change that.

lukechampine: hmm. I tried to look into how Bitcoin handles this, but it wasn't clear if they protect against this attack either. The main difference is that there is no explicit "AddMe" call; instead, nodes you connect to choose whether or not they want to add you. But it's not much different -- instead of calling "AddMe", you call "ver".
Bitcoin also periodically removes addresses that it hasn't heard from in a while, but the spec says this only occurs if it has more than 1000 addresses. So I dunno.

Issue #312, opened 2015-02-03
Signature marshalling corner case


DavidVorick:
Since the signature type is actually a pointer to 32 byte array, it encodes as 33 bytes.

I'm not sure if this is a problem.

DavidVorick:
It means that `encoding.Marshal(pk)` is not equivalent to `pk[:]`, which is a property that might be desirable.

Same goes for signatures.

lukechampine:
It doesn't seem like a technical problem, but I don't think it should be sent that way over the wire, unless there's a conceivable reason why a signature should be `nil`.
That reminds me...even though our Marshalling algorithm is well-defined, we're still going to have to write a specification that spells out exactly how each type is encoded. This is especially important if, e.g., we switch to a new signature algorithm, because everyone needs to agree on how it will be encoded.

For now I would just dereference the pointer before encoding it. But a more radical alternative would be to simply ban `nil` pointers from being encoded. Then pointers just encode to what they point to. I don't know if there are any `nil` pointers in consensus types, but I imagine there will be when encoding other local types (stuff that gets saved to disk). So this may not work. Or we may just decide not to use the same marshalling algorithm for local stuff.

also...pls spellcheck your issues, since you can't fix them silently afterwards :P

DavidVorick:
huh. We should tell github about the corener case of misspelled issue titles.

DavidVorick:
Right now, it seems sufficient to ban `nil` from being sent over the wire. Except for the strings of arbitrary data, I can't see what would cause problems.

We do all the signature stuff with pointers because that's how the crypto library has it. I'm not sure if there's a sidechannel based reason to do it that way, but I imagine so. (Not that it matters for public keys and signatures, but still)

lukechampine:
Arbitrary data should be an issue either, since it doesn't involve pointers. (`""` is encoded as a length prefix of 0).

Are there potentially `nil` pointers being used in other types that are encoded? Like the state?

DavidVorick:
block nodes don't get encoded. All the types should be pointer free, at least as far as sending consensus objects over the wire goes.

That might mean we need to adopt a different strategy for saving to disk though.

lukechampine:
that's what my second paragraph said lol.
I'll wait until we have a DB to decide though.

DavidVorick:
I'm just going to remove the pointers from our types in package crypto.

Issue #199, opened 2015-01-15
Proposed error conventions


DavidVorick:
I propose the following changes:

I like the idea of creating an error convention:

```go
err = someFunc()
if err != nil {
    errorHandlingFunc()
    // control statement, IE break, continue, or return
}
// else code
```

If for some reason there is no control statement (this would happen if you are merely logging errors), then it needs to be documented inside of the if statement. If there is no `err != nil`, that needs to be documented directly above where error is created.

This is not okay:
```go
err = someFunc()
return
```
because errors should always be checked. If code gets moved around, the person moving the code will need to remember to add error checking, and that's a chance for making mistakes, especially if the person moving the code is not the person who originally wrote the code.

This is also not okay:
```go
err = someFunc()
if err != nil {
    // stuff1
    // stuff2
    /// control statement
}
```

Either stuff1 and stuff2 need to be combined into a single function, or documentation needs to be added explaining why the convention was broken and what the error handling should look like.

lukechampine:
This seems reasonable enough that I'll try to follow it in the future, except for maybe the last point, which feels just a bit too rigid. I don't think it occurs much in practice anyway though.

DavidVorick:
I mean, if you need 2 or 3 simple statements, then throw a comment in explaining it. But if you need like 12 lines of code inside of the `err != nil`, it really should be it's own function because you don't want readers to forget that they're doing error handling.

DavidVorick:
Sometimes we get an error like "Failed to start daemon: could not unmarshal type []wallet.AddressKey" and we're not sure where the error is being called from. Is there some convention we can add that will allow us to more easily trace an error chain?

PR #352, opened 2015-02-22
General Changes

DavidVorick:
I removed the frozen coin stuff. I also added some documentation, and addressed issues #199 and #312.

### DIFF
--- a/modules/host.go
+++ b/modules/host.go
@@ -35,7 +35,7 @@ type Host interface {
 	// Announce announces the host on the blockchain. A host announcement
 	// requires two things: the host's address, and a volume of "frozen"
 	// (time-locked) coins, used to mitigate Sybil attacks.
-	Announce(addr network.Address, freezeValue consensus.Currency, freezeDuration consensus.BlockHeight) error
+	Announce(addr network.Address) error
 
 	// NegotiateContract is a strict function that enables a client to
 	// communicate with the host to propose a contract.
--- a/modules/host.go
+++ b/modules/host.go
@@ -32,9 +32,7 @@ type HostInfo struct {
 }
 
 type Host interface {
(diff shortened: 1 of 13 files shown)


lukechampine:
pls squash the last three commits

modules/host.go
```diff
@@ -35,7 +35,7 @@ type Host interface {
 	// Announce announces the host on the blockchain. A host announcement
 	// requires two things: the host's address, and a volume of "frozen"
 	// (time-locked) coins, used to mitigate Sybil attacks.
-	Announce(addr network.Address, freezeValue consensus.Currency, freezeDuration consensus.BlockHeight) error
+	Announce(addr network.Address) error
```
lukechampine: nice. But docstring needs to be updated

modules/hostdb.go
```diff
@@ -7,7 +7,7 @@ import (
 
 const (
 	// Denotes a host announcement in the Arbitrary Data section.
-	HostAnnouncementPrefix = "HostAnnouncement"
+	PrefixHostAnnouncement = "HostAnnouncement\n"
```
lukechampine: ?

DavidVorick: Makes it easier to read in plaintext. Not sure how useful that will end up being but I figured if the prefix is going to be human readable it might as well be formatted.

It's conceivable that someone would be viewing raw data in a text editor.

lukechampine: The remainder of the announcement is binary-encoded. No one is going to be reading or writing it by hand.

DavidVorick: hmm. You don't think it might be JSON encoded?

This stuff is easy enough to change later, I guess I can remove it for now.

DavidVorick:
fixed.

Issue #350, opened 2015-02-21
Removing Freeze Time from Host Picking


DavidVorick:
I don't think it's something we want to integrate yet, that's the type of thing you'd add after there's an entire ecosystem and an easy way to bootstrap. For now it's just a barrier to entry and a headache to implement. So I'm scrapping it entirely.

Instead, you'll create a list of hosts following this naively chosen set of rules, which we will update and refine with time:

1. Find all hosts and put them into a large database. This database will mostly remain on disk and not in memory.

2. Select hosts randomly from this database. Only 1 host is accepted per IP address, and generally speaking only 1 host is accepted per /24. Hosts within a /24 or on the same IP address will be able to sybil attack each other, but they shouldn't be able to sybil attack the wider network. To get unfair global representation, people would need a wide range of IP addresses, and they'd have to handle all uploads and downloads from those IP addresses (clients will refuse to talk to a host at a different address than the one advertised on the blockchain) which means their bandwidth will go up unless their datacenter is distributed.

This will cause a problem for multiple hosts on the same /24. We might be able to make a whitelist for universities or something (particularly RPI since that's likely to be a hotspot initially). Eventually we might implement freezing or something specifically to let multiple computers be picked from the same /24 with higher weight. I'm sure there will be evolving strategies and I'm okay with this.

3. Each host that was randomly selected (perhaps up to 500 live hosts or something) will be pinged infrequently to get a sense of uptime, throughput, and capacity. For now, we'll just trust a host that talks about capacity and treat everyone the same beyond 1TB of capacity. (easy enough to hit, and not that big a deal if someone lies about having a whole 1 TB).

This will at some level discourage hosts from offering more than 1TB, or disadvantage them a bit but that'll help with decentralization in the early network. This is a cap that'll probably be raised and again these rules will be evolving. It would be pretty simple to implement proof-of-capacity tests.

More briefly:

1. all hosts announced on the blockchain will be put into one large collection
2. this collection will be organized into buckets of hosts from the same /24 and buckets will be selected at random (up to 500). (perhaps favoring geographically close buckets)
3. One host is selected from each bucket (also at random) and pinged. If it's offline a new host will be picked from that bucket. (DoS attack vectors exist here, not sure we can do much about that easily)

You then have ~500 geographically diverse hosts to pick from. In the ultra early network we might relax this from /24 buckets to just one bucket per ip address. Actually, for now that's how I'm planning on implementing it.

DavidVorick:
[...]

PR #353, opened 2015-02-23
1 Host per IP

lukechampine:
Pretty straightforward. We can check off item 2 of #350.

### DIFF
--- a/modules/hostdb/hostdb.go
+++ b/modules/hostdb/hostdb.go
@@ -10,14 +10,15 @@ import (
 	"github.com/NebulousLabs/Sia/network"
 )
 
-// The HostDB is a set of hosts that get weighted and inserted into a tree
+// The HostDB is a database of potential hosts. It assigns a weight to each
+// host based on their hosting parameters.
 type HostDB struct {
 	state       *consensus.State
 	recentBlock consensus.BlockID
 
 	hostTree      *hostNode
-	activeHosts   map[network.Address]*hostNode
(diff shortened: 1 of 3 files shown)


modules/hostdb/hostdb.go
```diff
@@ -31,39 +32,44 @@ func New(state *consensus.State) (hdb *HostDB, err error) {
...
 
 // insert will add a host entry to the state.
 func (hdb *HostDB) insert(entry modules.HostEntry) error {
-	_, exists := hdb.activeHosts[entry.IPAddress]
+	// Entries are stored by address, sans port number. This limits each IP to
+	// advertising 1 host.
+	hostname := entry.IPAddress.Host()
+	_, exists := hdb.activeHosts[hostname]
 	if exists {
 		return errors.New("entry of given id already exists in host db")
```
DavidVorick: This error should probably be updated

DavidVorick:
Perhaps this is more complexity than we need now, (and perhaps item 1 of #350 is also more complexity than we need now) but I was thinking that all entries should go into the inactive database, even if there are already hosts with the same ip address. Then we can pull them out of the inactive database later.

For now, we might not need to keep an inactive database. I don't think we actually use it all currently, and maybe it's best to just delete it for the time being, to keep things simple.

That only leaves adding some method of pruning hosts if they go offline for an extended period of time. 'Extended' might be difficult to define, but pruning will probably be important because I'd expect a lot of churn in the early days of the network.

DavidVorick:
many thanks

Issue #351, opened 2015-02-21
Add developer documentation


cyrozap:
The `make dependencies && make` that Travis does doesn't seem to be enough to get a dev environment set up on Arch--some documentation would be helpful to help developers get started on the project.

DavidVorick:
Hmmm. We've tried our best to make it sufficient to run `make dependencies && make`, however we might have missed a few things.

One thing that is required is a working go environment, and I think you need 1.3 or later. This should probably be added to Developers.md.

Can you post the errors you are getting?

cyrozap:
I'm running Go 1.4.2, so the version shouldn't be a problem. Where I'm having issues, though, is getting the dependencies installed and, after messing around with the Makefile to get that to work, running `make` on the project. It appears as though the `go install -tags=dev ./...` step fails when it can't find any of the subpackages in `$GOROOT` or `$GOPATH`. That's the only step I'm stuck on. I fixed the `make dependencies` issue by commenting out `go install -race std` and `go get -u golang.org/x/tools/cmd/cover` since they seemed to conflict with Arch's packaged distribution of Go.

Here's a sample of the errors I get running `make`:

```
consensus/fork.go:4:2: cannot find package "github.com/NebulousLabs/Sia/crypto" in any of:
	/usr/lib/go/src/github.com/NebulousLabs/Sia/crypto (from $GOROOT)
	/home/cyrozap/.go/src/github.com/NebulousLabs/Sia/crypto (from $GOPATH)
```

DavidVorick:
oh. This happens if you're using a sim link. I'm not sure how to fix it, so I just set up bash aliases to cd to the Sia folder. I'll add this to the documentation if I can't find a better solution.

cyrozap:
I don't have any symlinks in either of those paths, so I don't know why I'm getting that error. Once I do, though, I'll write a small how-to on avoiding this issue.

DavidVorick:
It's difficult for me to guess without seeing your machine, but did you make sure to call `git clone` in `$GOPATH/github.com/NebulousLabs`? Simply cloning Sia into the gopath isn't going to be sufficient, that's also not documented anywhere.

cyrozap:
Ah, I see, I needed to clone Sia inside that path and work on it in there.

PR #356, opened 2015-02-23
Module Cleanup

### LINKED ISSUE #324: Add website to readme
right now, people who find us through the github page (I mentioned Sia in a reddit comment earlier today) don't have any way to find the website and download the binaries.

### LINKED ISSUE #354: Gateway behavior tweak
The gateway should only broadcast blocks and transactions that are valid. To help make this an inherent property, I'm tweaking the behavior a bit.

The gateway will only broadcast blocks as they are added to the consensus set. This essentially means listening on a state subscription and doing a broadcast whenever new blocks appear in the BlocksSince() function. When the gateway receives blocks, it just uses them to call AcceptState(), and doesn't do anything more (error or not). If it's a block that should be broadcasted, it'll appear in a BlocksSince call a few seconds later.

For transactions, the transaction pool will import the gateway, and call RelayTransaction() whenever a transaction is accepted. I know this is backwards from how the state does things, but you can't really subscribe to the transaction pool because the structure is much less formal.

If invalid blocks or transactions start getting broadcasted, DoS vectors open up.

I'm working on this now, just wanted to explain where the changes are coming from and why.

DavidVorick:
Gateway now gets updated by the transaction pool, instead of getting updated by things submitting transactions. Gateway also now gets blocks from the consensus set, instead of getting blocks from rpcs. Rpcs directly submit to the transaction pool and the state. Fixes #354 

The website has been added to the readme. Fixes #324 

Timeouts have been added to `make test`.

Some thread safety was added to the consensus tester functions. I'm pretty sure they're still not fully thread safe, because I didn't do a good job of following threading conventions. More thorough lookover of these functions to come later.

I added notifications to the state. Now, the gateway and the host get notified every time the state updates. Sometimes, the gateway seems to get too many notifications. Idk what this is about, but it might cause sporadic testing failures until I fix it. It's up there in priority.

The gateway now checks for nil input. This does cause some port problems when testing in bulk. There are multiple ways to handle this but I chose to introduce some global state, which is kind of gross. I do think it's much more acceptable than just passing in a nil tcp server, and I don't think that would end up being a good solution anyway because a lot of the tests do use tcp functions at some point, so the tests would fail anyway and we'd still need some sort of way to prevent port conflicts. I'm sure there's a better way to do this, but I think we should mostly devote our energy towards the next iteration being finished.

Host/renter negotiation has been cleaned up a bit to match the new file contracts. Unfortunately, download is broken (just remembered now) and the renter requires the host to be non-malicious. Both of these can be fixed with finesse, but it adds 2.5 more steps to the negotiation. (current steps: 1. send terms, 2. send file, 3. send signed contract. New steps: 1. send terms, 2. send file, 3. send unsigned contract 4. receive unsigned contract but with host inputs added, 5. send signed whole transaction. The host then needs to check again that the signed contract matches the unsigned contract before signing it).

This whole PR comes without tests. Those to be added soon.


Developers.md
```diff
@@ -5,10 +5,16 @@ Sia is mostly written in go. To build and test Sia, you are going to need a
...
 
 If you plan on cross compiling Sia, you may need to install go from source. You
-can find information on that [here](golang.org/doc/install/source).
+can find information on that [here](http://golang.org/doc/install/source).
+
+When you clone the Sia repository, make sure that you call `git clone` from the
+folder `$GOPATH/github.com/NebulousLabs/`. It will not be sufficient to clone
+Sia from the gopath. Alternatively, you should be able to call `go get -u
+github.com/NebulousLabs/Sia/... && go get -u
+github.com/NebulousLabs/Sia/modules/...`.
```
lukechampine: AFAIK the second command is unnecessary.
We should probably make this the preferred way of installing; it's easier than using git anyway.

consensus/state.go
```diff
@@ -49,6 +49,13 @@ type State struct {
 	siafundOutputs        map[SiafundOutputID]SiafundOutput
 	delayedSiacoinOutputs map[BlockHeight]map[SiacoinOutputID]SiacoinOutput
 
+	// subscriptions is a map containing a bunch of channels that are being
+	// listend on by modules. An empty struct is thrown down the channel any
+	// time that the consensus set of the state changes. subscriptionCounter
+	// only ever increments, and prevents collisions in the map.
+	subscriptions       map[int]chan struct{}
+	subscriptionCounter int
```
lukechampine: why use a map + int instead of just a slice?

consensus/notifications.go
```diff
@@ -0,0 +1,24 @@
...
+// SubscribeToConsensusChanges returns a channel that will be sent an empty
+// struct every time the consensus set changes.
+func (s *State) SubscribeToConsensusChanges() (c chan struct{}) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	c = make(chan struct{}, 1)
+	s.subscriptions[s.subscriptionCounter] = c
+	s.subscriptionCounter++
+	return
+}
```
lukechampine: `chan struct{}` is bidirectional, which means subscribers will be able to send down it. Not a big deal, but casting it to a receive-only channel is best practice. This is as easy as changing the return type to `<- chan struct{}`

modules/gateway/blocks.go
```diff
@@ -0,0 +1,42 @@
...
+
+		g.RelayBlock(block)
+	}
+}
+
+func (g *Gateway) threadedConsensusListen(consensusChan chan struct{}) {
+	for _ = range consensusChan {
+		g.update()
+	}
+}
```
lukechampine: Is there a reason `update` is its own function, instead of merging with `threadedConsensusListen`?

modules/host/contractcreation.go
```diff
@@ -244,9 +254,18 @@ func (h *Host) NegotiateContract(conn net.Conn) (err error) {
...
-		path: path,
+	h.obligationsByHeight[proofHeight] = append(h.obligationsByHeight[proofHeight], contractObligation{
+		id:           id,
+		fileContract: fc,
+		path:         path,
+	})
+	h.obligationsByID[id] = contractObligation{
+		id:           id,
+		fileContract: fc,
+		path:         path,
```
lukechampine: Literals are nice, but not if you have to repeat yourself.
```go
co := contractObligation{
    id: id,
    fileContract: fc,
    path: path,
}

h.obligationsByHeight[proofHeight] = append(h.obligationsByHeight[proofHeight], co)
h.obligationsByID[id] = co
```

lukechampine:
I'm starting to wonder if the integration tests should be moved to the siad package. It doesn't seem right for all the module tests to importing all the other modules.
On the other hand, the majority of our tests (as of now) are integration tests, so the result would be a gorillion _test.go files in siad. Maybe the real problem is that our modules aren't conducive enough to unit testing.

modules/host/host_test.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+	g, err := gateway.New(tcps, ct.State)
+	if err != nil {
+		t.Fatal(err)
+	}
+	tp, err := transactionpool.New(ct.State, g)
+	if err != nil {
+		t.Fatal(err)
+	}
+	w, err := wallet.New(ct.State, tp, "")
```
lukechampine: isn't this more or less equivalent to initializing with `nil`?

modules/host/upload.go
```diff
@@ -0,0 +1,54 @@
...
+
+	// Open the file.
+	file, err := os.Open(contractObligation.path)
+	if err != nil {
+		return
+	}
+	defer file.Close()
+	info, _ := file.Stat()
+
+	conn.SetDeadline(time.Now().Add(time.Duration(info.Size()) * 8 * time.Microsecond))
```
lukechampine: this timeout should match the other one (change 8 to 128)

lukechampine:
This probably should have been split into multiple PRs. There are three separate things going on here: updates, contracts, and the gateway shenanigans. It's harder to review them when they're all intertwined. Reviewing them separately might have made it easier to figure out what was causing the subscription bug you described.

DavidVorick:
Fixes #351

DavidVorick:
I would have liked to do them all separately, but they were unfortunately all intertwined.

Actually, I think I could have done the gateway stuff separately. Most of the rest of it though all belonged in the same PR, because it wouldn't build until I had touched all of it.

Developers.md
DavidVorick: the only disadvantage is that origin gets set to what is usually called upstream, instead of origin getting set to your fork of the project.

consensus/state.go
DavidVorick: there were at some points plans to let subscriptions die/expire/timeout. If your host is restarting or something individually you'll want a way to release a subscription.

Not a big deal though. For the time being, it's identical to using a slice and calling append.

modules/gateway/blocks.go
DavidVorick: oh. I was trying to figure out where the deadlock was happening. This was just to make life easier  with some `defer` calls I was making.

I just forgot to move it back.

modules/host/contractcreation.go
DavidVorick: another one where I was just finding the path of least resistance. Bad me

DavidVorick:
> It doesn't seem right for all the module tests to importing all the other modules.

The problem is that you've got no way to test that SubmitStorageProof is working without some sort of transaction pool to submit it to. I could have made some sort of fake transaction pool, and perhaps that's the direction we should move in. Potentially, the consensus test suite could be forked into it's own package containing a bunch of barley-functioning modules that everyone can import and use.

The problem is that these parts weren't meant to exist individually. How am I supposed to test that the transaction pool properly handles conflicts appearing in the blockchain if I've got no way to mine blocks?

lukechampine:
Yeah, we're always going to need integration tests, and I think they'll be the most important part of the testing suite. But unit tests are important too, and our current design makes them difficult to write. They get easier to write the more pure your functions are; the problem is that most of our module functions are impure in one way or another.

modules/host/host_test.go
DavidVorick: yes, and it's actually going to cause problems once we start saving and loading files from the wallet again. In the near future it's going to be `filesCreatedDuringTesting/wallet`.

Developers.md
lukechampine: Hmm. So I guess contributers would want to first fork the repo, and then run `go get` on their fork?

DavidVorick:
I'm all for having more pure modules if you've got any ideas. I just don't think there's a way to do it without rewriting a bunch of functionality.

DavidVorick:
Fixes #324, #351, #354 

I still think we should have the gateway not accepting blocks, and instead being a subscriber. That was causing errors though. Or at least, not accepting blocks on the RelayBlock function call. There's definitely a good reason for the gateway to get the blocks before passing them to the state though: DoS management.

Notably, I left the names of the RPCs at 'AcceptBlock' and 'AcceptTransaction', though it's the gateway's 'RelayBlock' and the tpool's 'AcceptTransaction' that do the handling. I do think that overall the model needs improvement. I don't like how it currently is, but I don't really like anything else that's been suggested either.

modules/gateway/gateway.go
```diff
@@ -56,62 +56,33 @@ func (g *Gateway) Bootstrap(bootstrapPeer network.Address) (err error) {
...
 
-// SharePeers returns up to 10 randomly selected peers.
-func (g *Gateway) SharePeers() (peers []network.Address, err error) {
-	g.mu.RLock()
-	defer g.mu.RUnlock()
-	for peer := range g.peers {
-		if len(peers) > 10 {
-			return
+	// Check if b is in the current path.
+	height, exists := g.state.HeightOfBlock(b.ID())
```
lukechampine: maybe a `IsInCurrentPath` helper is warranted...would that be useful anywhere else?

DavidVorick: maybe. I'm actually not 100% certain where all the helpers are used, or if there's a more efficient way to use them in some places. There's probably room for cleanup and polishing here.

Issue #274, opened 2015-01-20
Make Travis Build Fail if code is not go fmt'd


lukechampine:

DavidVorick:
I don't know how we'd approach making this happen

lukechampine:
Well, `go fmt` outputs a list of files it changed, so we can run `go fmt` during the `install` key and check for output. I think this should do it:
```bash
install:
    $(exit $(go fmt ./... | wc -l))
    make dependencies
    make # is this necessary?
```

PR #357, opened 2015-02-24
travis fails if go fmt makes changes

DavidVorick:
I have made a small change to accept.go to demonstrate that travis will fail.

`make` will not fail if the code is not formatted, only `make fmt-strict`, which means workflow should not be disrupted. Travis runs `make fmt-strict` before running make to catch the problem.

See the first commit for a demonstration of the build failing. Second commit fixes this so we don't merge failed code. And additional useful change was made so it's not a wasted commit.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,9 @@ all: install
 fmt:
 	go fmt ./...
 
+fmt-strict:
+	@exit $(shell go fmt ./... | wc -l)
+
 # REBUILD touches all of the build-dependent source files, forcing them to be
 # rebuilt. This is necessary because the go tool is not smart enough to trigger
 # a rebuild when build tags have been changed.
--- a/Makefile
+++ b/Makefile
@@ -6,9 +6,6 @@ all: install
 fmt:
 	go fmt ./...
 
-fmt-strict:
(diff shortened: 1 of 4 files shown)


DavidVorick:
I switched it to go 1.4.1 because travis has go 1.4.1 preinstalled, but not go 1.4 preinstalled. It just makes life easier on travis.

DavidVorick:
Fixes #274

Makefile
```diff
@@ -6,6 +6,9 @@ all: install
 fmt:
 	go fmt ./...
 
+fmt-strict:
+	@exit $(shell go fmt ./... | wc -l)
+
```
lukechampine: can this be moved to the .travis file? I don't feel like it belongs in the Makefile.

lukechampine:
I think that commit broke Travis lol. You need to run the `exit` in a subshell, otherwise it will terminate the Travis process. Try `$(exit $(go fmt ./... | wc -l))`

DavidVorick:
urg. Looks like `go get` is failing b/c go.googlesource.com is down? :(

lukechampine:
it's working for me locally now. Try again (commit --amend with no changes)

PR #358, opened 2015-02-24
Add HostDB Tests

DavidVorick:
As usually happens, writing tests for the hostdb helped me to reason about a lot of things that it was doing, so there are a bunch of changes.

I also moved some things around in the network package to increase the organization. I renamed network.go to tcpserver.go, and I moved the address functions to their own address.go file. None of the functions or types were changed, just which files they are in.

Entry weight no longer weights according to price. This is of course a pretty big problem. I couldn't think of a good way to write the equation. I think I'd want the weight to increase linearly as the collateral increases, and I'd want the weight to decrease with the square of the price.

The problem is that we're stuck with integer weights, so you'll only have positive weights in that scheme if the collateral is equal to the square of the price, which is unreasonable. We should probably also enforce a max and min on the collateral. Max of 2x price, min of 1/2 price. The price would then be the dominating factor.


modules/hostdb/hostdb.go
```diff
@@ -10,165 +9,45 @@ import (
...
-		// If the node is in the set of inactive hosts, delete from that set,
-		// otherwise return a not found error.
-		_, exists := hdb.inactiveHosts[hostname]
-		if exists {
-			delete(hdb.inactiveHosts, hostname)
-			return nil
-		} else {
-			return errors.New("address not found in host database")
+		if consensus.DEBUG {
+			panic(ErrMissingGenesisBlock.Error())
```
lukechampine: `panic(ErrMissingGenesisBlock)` is sufficient

lukechampine:
You can just add a coefficient to the collateral or price term, e.g. `weight := 4*collateral - 0.5*price^2`. The trick would be choosing the right values.

DavidVorick:
It multiply and divide, not add and subtract.

My only idea so far has been to increase the precision. We could also use a rat but we'll be adding a bunch of these together so they'll probably get pretty out of hand quickly in terms of memory and computation time.

If the price gets set at at 2^80, we'd want 160 bits for precision after squaring it. Currency has unlimited precision, so we can jump it as much as we want. On the other hand, if we limit the collateral (when weighting) to be at least 1/2 of the price, that's guaranteed to kill all but 1 bit of the doubling. So a price of 2^80 will have a collateral of 2^79, and the squaring factor will leave us with a weight of 1/2^80.

I think I'm going to set the baseline at 2^100, then multiply by the adjusted collateral, then divide by the price squared. For any reasonable price, the result should be positive with 20-60 bits to spare. At a price of 100 bits, you'd literally need a full block of coins *per byte* being spent, so I think it's quite safe. Anyone pricing higher than that will just have a weight of zero.

modules/hostdb/entries.go
```diff
@@ -1,42 +1,153 @@
...
+	maxCollateral := entry.Price.Mul(consensus.NewCurrency64(2))
+	minCollateral := entry.Price.Div(consensus.NewCurrency64(2))
+	var collateral consensus.Currency
+	if maxCollateral.Cmp(entry.Collateral) < 0 {
+		collateral = maxCollateral
+	} else if minCollateral.Cmp(entry.Collateral) > 0 {
+		collateral = minCollateral
+	} else {
+		collateral = entry.Collateral
+	}
```
lukechampine: docstring is incomplete.
I would write this as:
```go
collateral := entry.Collateral
// clamp collateral to between 0.5x and 2x the price
if collateral.Cmp(entry.Price.MulFloat(2)) > 0 {
    collateral = entry.Price.MulFloat(2)
} else if collateral.Cmp(entry.Price.MulFloat(0.5)) < 0 {
    collateral = entry.Price.MulFloat(0.5)
}
```
(Might be cleaner to write `if maxCollateral := entry.Price.MulFloat(2); collateral.Cmp(maxCollateral) > 0 {` to avoid recalculation, but this function doesn't need to be terribly efficient anyway.)

modules/hostdb/entries.go
```diff
@@ -1,42 +1,153 @@
...
+
+	// Create a baseline weight of 10^29, which adds precision to the equation
+	// and makes sure that all reasonable prices end with a weight that's
+	// greater than zero.
+	weight = consensus.NewCurrency(new(big.Int).Exp(big.NewInt(10), big.NewInt(29), nil))
+	weight = weight.Mul(collateral).Div(entry.Price).Div(entry.Price)
```
lukechampine: 10^29 is a lot smaller than 2^100.
Dividing twice is pretty ugly too, I'd rather just add an `Exp` method to `Currency`.

DavidVorick: Currency doesn't have Exp implemented, and I don't think it's necessary just for this one use case. These equations are probably going to change substantially.

I meant 10^30, which is 2^99

modules/hostdb/entries.go
DavidVorick: that does look cleaner

PR #359, opened 2015-02-25
Host DB Testing

DavidVorick:
There's not a whole lot happening in this PR. I wrote a few tests to check that inactive hosts were working correctly. Checking that active hosts are working correctly is going to require making fake RPCs... we're just going to have to leave this to the integration testing for this week.

I completely refactored insert, now it's got a lot less side effects and more standard/expected behavior. insert puts stuff into the full set, then determines if it belongs in the active set. threadedActiveInsert gets the host settings and gives up if the host is offline. If the host is online, it grabs the settings and calls insertCompleteHostEntry, which removes any conflicts and then inserts the host.

Each of these functions is independent, and can be called without problems or side effects individually.

The docstrings might not be perfect. I tried to go through and fix them all.

### DIFF
--- a/modules/hostdb/entries.go
+++ b/modules/hostdb/entries.go
@@ -18,6 +18,8 @@ var (
 
 	// Convenience variables for doing currency math. Originally we were just
 	// using MulFloat but this was causing precision problems during testing.
+	// The actual functionality of the program isn't affected by loss of
+	// precision, it just makes testing simpler.
 	currencyZero     = consensus.NewCurrency64(0)
 	currencyOne      = consensus.NewCurrency64(1)
 	currencyTwo      = consensus.NewCurrency64(2)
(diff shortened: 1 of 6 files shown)


modules/hostdb/entries.go
```diff
@@ -10,39 +10,62 @@ import (
...
-	return
+	// Prevent a divide by zero error by making sure the price is at least one.
+	price := entry.Price
+	if price.Cmp(currencyZero) <= 0 {
+		price = currencyOne
+	}
+
+	// Take the base weight, multiply it by the clapmed collateral, then divide
+	// it by the square of the price.
+	return baseWeight.Mul(collateral).Div(price).Div(price)
```
lukechampine: I think `.Div(price.Mul(price))` is slightly clearer but whatevs

PR #360, opened 2015-02-26
Save/Load Wallet

lukechampine:
Pretty straightforward. The only thing a wallet saves is that `{UnlockConditions, SecretKey}` pair of each entry in its `w.keys` map. This means that saves are relatively efficient. Loads, on the other hand, require rebuilding both the `w.keys` map and the `w.timelockedKeys` map. More importantly, they require reconstructing all of the outputs associated with a key, which means rescanning the blockchain from the beginning. Fortunately, loads are much less frequent than saves, so this isn't a huge deal. In the future we can worry about making them more efficient (probably by saving the outputs and then adjusting them as needed to account for consensus changes).

### DIFF
--- a/modules/host/host_test.go
+++ b/modules/host/host_test.go
@@ -12,7 +12,8 @@ import (
 )
 
 var (
-	tcpsPort int = 10500
+	tcpsPort  int = 10500
+	walletNum int = 0
 )
 
 // A HostTester contains a consensus tester and a host, and provides a set of
@@ -40,7 +41,7 @@ func CreateHostTester(t *testing.T) (ht *HostTester) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	w, err := wallet.New(ct.State, tp, "")
+	w, err := wallet.New(ct.State, tp, "../../host_test"+strconv.Itoa(walletNum)+".wallet")
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/modules/miner/miner_test.go
+++ b/modules/miner/miner_test.go
@@ -27,7 +27,7 @@ func TestMiner(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	w, err := wallet.New(s, tpool, "")
+	w, err := wallet.New(s, tpool, "../../miner_test.wallet")
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/modules/wallet/files.go
+++ b/modules/wallet/files.go
@@ -30,7 +30,6 @@ func (w *Wallet) save() (err error) {
 
 	// Write the data to file.
 	// TODO: Instead of using WriteFile, write to a temp file and then do a move.
-	// TODO: append instead of overwriting everything.
 	err = ioutil.WriteFile(w.filename, encoding.Marshal(keySlice), 0666)
 	if err != nil {
 		return
--- a/modules/wallet/files.go
+++ b/modules/wallet/files.go
@@ -1,112 +1,80 @@
 package wallet
 
 import (
-	// "io/ioutil"
-	// "os"
+	"errors"
+	"io/ioutil"
+	"os"
 
 	"github.com/NebulousLabs/Sia/consensus"
 	"github.com/NebulousLabs/Sia/crypto"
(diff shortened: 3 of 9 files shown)


modules/host/host_test.go
```diff
@@ -40,7 +40,7 @@ func CreateHostTester(t *testing.T) (ht *HostTester) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	w, err := wallet.New(ct.State, tp, "")
+	w, err := wallet.New(ct.State, tp, "../../host_test.wallet")
```
DavidVorick: do these still need to go in a top level.

But a bigger question, `CreateHostTester` is called many times. You probably need a number here tacked onto the name. `host_test.num.wallet` perhaps?

modules/miner/miner_test.go
```diff
@@ -27,7 +27,7 @@ func TestMiner(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	w, err := wallet.New(s, tpool, "")
+	w, err := wallet.New(s, tpool, "../../miner_test.wallet")
```
DavidVorick: same issue as the host test wallet

modules/wallet/files.go
```diff
@@ -1,112 +1,80 @@
...
-				SpendConditions: address.spendConditions,
-				SecretKey:       address.secretKey,
-			}
-			timelockedKeys[i] = key
-			i++
-		}
+	// convert key map to a slice
+	keySlice := make([]savedKey, len(w.keys))
+	for _, key := range w.keys {
+		keySlice = append(keySlice, savedKey{key.secretKey, key.unlockConditions})
```
DavidVorick: I think you meant `keySlice[i] = savedKey{...}`

Calling append will give you a doubly long slice where the first half is just blank.

modules/wallet/files.go
```diff
@@ -1,112 +1,80 @@
...
-		// TODO: Instead of using WriteFile, write to a temp file and then do a move,
-		// this is an action that should probably appear in a library somewhere.
-		err = ioutil.WriteFile(w.saveFilename, fileData, 0666)
-		if err != nil {
-			return
-		}
-	*/
+	// Write the data to file.
+	// TODO: Instead of using WriteFile, write to a temp file and then do a move.
+	// TODO: append instead of overwriting everything.
```
DavidVorick: Append might be tricky because you have to have some sort of awareness of which keys are already in the file. What we'd probably do instead is call a saveSingleKey() function to write to the temp file, and then a mergeTemp() function that takes the temp file and the full file and puts them together.

Or if you get appending working, saveSingleKey() can just do the append.

I don't consider it necessary to make these a TODO, since there's already an issue for it.

modules/wallet/files_test.go
```diff
@@ -0,0 +1,36 @@
...
+	for mapKey := range wt.keys {
+		if _, exists := newWallet.keys[mapKey]; !exists {
+			wt.Fatal("Loaded wallet is missing a key")
+		}
+	}
+	for mapKey := range wt.timelockedKeys {
+		if _, exists := newWallet.timelockedKeys[mapKey]; !exists {
+			wt.Fatal("Loaded wallet is missing a time-locked key")
+		}
+	}
```
DavidVorick: Can you add some code to make sure the balances match? I doubt the test will fail but I still think the test should be in place.

modules/host/host_test.go
lukechampine: `rm -rf *.wallet` doesn't seem to catch wallets in nested directories.

good point. Looks like it didn't cause a problem here because there's only one test.

modules/wallet/files.go
lukechampine: ach, you're right. The correct way is to use `make([]savedKey, 0, len(w.keys))` to allocate all the memory while keeping the length 0.

modules/miner/miner_test.go
lukechampine: The miner doesn't have a testing type yet. We can implement this once it does.

modules/wallet/files_test.go
lukechampine: done

PR #361, opened 2015-02-26
Save/Load Host, Renter, and Gateway

lukechampine:
~~Plus a minor change to `threadedUploadPiece`; I figure passing a slice is more idiomatic Go compared to the more C-style approach of passing a pointer. (I mean, they're both pointers under the hood, but you get the point.)~~


DavidVorick:
Is it idomatic go to pass around a slice of len 1? If it's always going to be len 1 shouldn't you just use a pointer?

lukechampine:
Boy. Second time today I've been wrong about slices.

PR #363, opened 2015-02-27
Host testing

DavidVorick:
for all of the commits, there's actually only a little bit of real host testing happening.

I renamed some stuff, refactored the negotiation algorithm to work in a byzantine setting (previously it did not), and tested that the host is correctly producing proofs of storage, which it does seem to be doing.

This PR is actually not terribly complete, there's a decent amount of cleanup left to do but I started working a bit on the crypto package and decided to PR while things were moderately disorgainzed, instead of having 1 PR that did 40 things. Now it only does 25, but I guess you take what you can get.

More testing to come.


.gitignore
```diff
@@ -1,6 +1,8 @@
 cover
 hostdir
+hostdata_test
 release
+renterdata_test
```
lukechampine: how about *_test

modules/File Contract Negotiation.md
```diff
@@ -0,0 +1,70 @@
+File Contract Negotitation
```
lukechampine: Negotiation

modules/host/negotiate_test.go
```diff
@@ -0,0 +1,88 @@
...
+	file, path, err := ht.allocate(4e3)
+	if err != nil {
+		ht.Fatal(err)
+	}
+	file.Close()
+
+	// Check that the file has a real name and that it exists on disk.
+	if path == "" {
+		ht.Fatal("path of allocated file is empty!")
+	}
```
lukechampine: I don't think this scenario is possible.

modules/host/negotiate_test.go
```diff
@@ -0,0 +1,88 @@
...
+	fullpath := filepath.Join(ht.Host.hostDir, path)
+	_, err = os.Stat(fullpath)
+	if os.IsNotExist(err) {
+		ht.Fatal("file does not exist on disk")
+	}
+
+	// Check that spaceRemaining has decreased appropriately.
+	if currentSpaceRemaining != ht.spaceRemaining+4e3 {
+		ht.Error("space remaining did not decrease appropriately after allocating a file")
+	}
```
lukechampine: this condition confuses me. How about:
```go
const filesize = 4e3 // use an untyped const, otherwise Go will think you want a float

initialSpace := ht.spaceRemaining
//...
if ht.spaceReamaining != initialSpace - filesize {
```

modules/host/negotiate_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+	// Check that the file has a real name and that it exists on disk.
+	if path == "" {
+		ht.Fatal("path of allocated file is empty!")
+	}
+	fullpath := filepath.Join(ht.Host.hostDir, path)
+	_, err = os.Stat(fullpath)
+	if os.IsNotExist(err) {
+		ht.Fatal("file does not exist on disk")
+	}
```
lukechampine: I wrote a helper for this in wallet/files.go. Seems like it's needed in more places though.

modules/host/update.go
```diff
@@ -0,0 +1,117 @@
...
+
+// threadedConsensusListen listens to a channel that's subscribed to the state
+// and updates every time the consensus set changes. When the consensus set
+// changes, the host checks if there are any storage proofs that need to be
+// submitted and submits them.
+func (h *Host) threadedConsensusListen(consensusChan <-chan struct{}) {
+	for _ = range consensusChan {
+		h.update()
+	}
+}
```
lukechampine: you know, if every module had an update function...
and if the State was a module...
we could make an `Updater` interface...
and the State could takes a slice of `Updaters`...
and call `Update` on them itself instead of sending down a bunch of channels.

modules/File Contract Negotiation.md
DavidVorick: hmm. Doc folder, or not even in the Sia repo?

modules/host/update.go
DavidVorick: that actually sounds pretty good. Modules would still subscribe, but they'd subscribe by adding themselves. Then the state could just iterate and send the diffs, and a lot of problems would probably float away.

modules/host/negotiate_test.go
DavidVorick: maybe you could put it directly in package modules.

modules/host/update.go
DavidVorick: We could do it without making the state a module too. Just write the updater interface in the consensus package.

PR #364, opened 2015-02-27
Merkle package

DavidVorick:
I left all of the other merkle stuff because it's working well enough. I put the new merkle stuff into it's own package.

I sat down and worked through everything, and I feel like I have a pretty pristine understanding of how and why all this code works. I haven't documented it well, but I plan on it. The proof builder needs a few more tests, marked with TODO. Unfortunately, they have 15 nodes each, so it'll be a decent amount of code to build everything by hand.

I would like more thorough coverage of the smaller numbers too, but practically speaking we can't manually build all of the possible trees. I don't think that the existing merkle code would pass the tests I have written, we never did any manual testing for the proof building and verification.

Verification is not implemented, though it's probably going to be quite a bit simpler than everything else. Testing it should also be pretty simple, just build stuff with the already-tested builder and see that the verifier passes it. (then change the index and see that the verifier fails it for all possible other indices). No need to make anything by hand.


merkletree/root.go
```diff
@@ -0,0 +1,78 @@
...
+
+// A Tree builds a Merkle tree. Add data one slice at a time, and the tree will
+// hash the data and build out the Merkle tree using log(n) memory, where n is
+// the number of times 'Push' is called.
+type Tree struct {
+	head *node
+	hash hash.Hash
+}
+
+// A node is a memeber of the Tree.
```
lukechampine: nice meme

nice meme

merkletree/root.go
```diff
@@ -0,0 +1,78 @@
...
+		height++
+		t.head = t.head.next
+	}
+
+	t.head = &node{
+		next:   t.head,
+		height: height,
+		value:  value,
+	}
+}
```
lukechampine: recursive implementation:
```go
func (t *Tree) Push(data []byte) {
    t.head = &node{
        next: t.head,
        height: 1,
        value: sum(t.hash, data)
    }
    t.collapse()
}

func (t *Tree) collapse() {
    next := t.head.next
    if t.head.height != next.height {
        return
    }

    next.value = sum(t.hash, append(t.head.value, next.value))
    next.height++
    t.head = next
    t.collapse()
}

File: merkletree/root.go [hunk 3]
```diff
@@ -0,0 +1,78 @@
...
+	value  []byte
+}
+
+// sum returns the sha256 hash of the input data.
+func sum(h hash.Hash, data []byte) (result []byte) {
+	h.Write(data)
+	result = h.Sum(nil)
+	h.Reset()
+	return
+}
```
lukechampine: docstring is inaccurate.

merkletree/root_test.go
```diff
@@ -0,0 +1,164 @@
...
+
+	// Try building the trees for sizes 0 through 8 and see if it matches the
+	// manually obtained Merkle roots.
+	for i := 0; i < 8; i++ {
+		tree.Push(data[i])
+	}
+	tree8 := tree.Root()
+	if bytes.Compare(root8, tree8) != 0 {
+		t.Error("tree8 doesn't match root8")
+	}
```
lukechampine: make a slice of roots and you can condense all of these into one loop

merkletree/root.go
```diff
@@ -0,0 +1,78 @@
...
+	head *node
+	hash hash.Hash
+}
+
+// A node is a memeber of the Tree.
+type node struct {
+	next   *node
+	height int
+	value  []byte
+}
```
lukechampine: couldn't resist...here's an implementation with slices:
```go
type stackElem struct {
    height int
    sum    []byte
}

type Tree struct {
    hash hash.Hash
    stack []stackElem
}

func (t *Tree) Push(data []byte) {
    // grow front of stack
    newStack := make([]stackElem, len(t.stack)+1)
    copy(newStack[1:], t.stack)
 
    t.stack[0] = stackElem{1, sum(t.hash, data)}
    t.collapse(true)
}

func (t *Tree) Root() []byte {
    t.collapse(false)
    root := t.stack[0].sum
    t.stack = nil
    return root
}

func (t *Tree) collapse(sameheight bool) {
    if len(t.stack) < 2 {
        return
    } else if sameheight && t.stack[0].height != t.stack[1].height {
        return
    }
    t.stack[1].sum = sum(t.hash, append(t.stack[0].sum, t.stack[1].sum...))
    t.stack[1].height++
    // shift stack down without reducing capacity
    n := copy(t.stack, t.stack[1:])
    t.stack = t.stack[:n]
    // collapse next pair
    t.collapse()
}
```

merkletree/prove.go
```diff
@@ -0,0 +1,89 @@
...
+		provingForIndex: index,
+	}
+}
+
+func (t *TreeProve) SetIndex(index int) {
+	t.head = nil
+	t.provingForIndex = index
+	t.currentIndex = 0
+	t.proveSet = nil
+}
```
lukechampine: this method seems unnecessary

merkletree/root.go
```diff
@@ -0,0 +1,78 @@
...
+// the docstring of a function, that should only explain 'what'. Why and how
+// belong elsewhere.
+
+// A Tree builds a Merkle tree. Add data one slice at a time, and the tree will
+// hash the data and build out the Merkle tree using log(n) memory, where n is
+// the number of times 'Push' is called.
+type Tree struct {
+	head *node
+	hash hash.Hash
+}
```
lukechampine: Does this need to be exported?

merkletree/root.go
```diff
@@ -0,0 +1,78 @@
+package merkletree
+
+import (
+	"hash"
+)
+
+// TODO: Explain how and why this works here. Why + how should not be part of
+// the docstring of a function, that should only explain 'what'. Why and how
+// belong elsewhere.
```
lukechampine: not necessarily true; see the docstring of `json.Marshal`, for example. I think an explanatory docstring would be fine. Preferably it would be on `Tree`.

lukechampine:
General comments: the proof stuff needs better variable names and a lot more documentation. I'm really struggling to understand how it works.

We should also have a helper function for `io.Reader`s. Should be as simple as:
```go
func BuildReaderProof(r io.Reader, h hash.Hash, segmentSize int, index int) (root []byte, proofSet [][]byte, err error) {
    proofTree := NewProofTree(h, index)
    for {
        segment := make([]byte, segmentSize)
        if _, err = io.ReadFull(r, segment); err != nil {
            return
        }
        proofTree.Push(segment)
    }
    root, proofSet = proofTree.Prove()
    return
}
```

merkletree/root.go
DavidVorick: wha?

DavidVorick:
I will get to the rest of your comments later today, don't have much time right now.

I'm pushing the changes I made (prior to having internet access) because it addresses some of the things you have mentioned. The test suite is cleaner now, though not as clean as some of your suggestions would make it.

I'll have to look into the collapse() function.

I merged ProveTree and Tree into one structure, originally they were separate so I could understand them better.

I don't want ReaderMerkleRoot and, ReaderProof, and FileMerkleRoot implemented in package merkletree, which I now fully intend to make its own library (it's generalized enough at this point). Instead, we can put them in package crypto, which is a bunch of helper functions that use crypto. That will also allow us to automatically use the hash functions that we want, instead of needing to pass one in.

I'm not sure that the documentation is much better, but the variable names should be better at this point. I'll get to it soon.

lukechampine:
I don't see the disadvantage of including helper functions. FileMerkleRoot is less important, but the Reader functions are less straightforward than people might assume due to the necessity of `ReadFull`. If we don't include them, we force every user to implement them manually, and potentially make a mistake.

DavidVorick:
Sorry it took so long to push some of this. You'll now find that BuildReader is properly commented, better organized, and hopefully fairly intuitive, though it might take a bit to sink in.

After documenting Verify and adding the last set of tests, I'll add the ReaderBuild and ReaderRoot functions. And probably some benchmarks.

Is there a way to get these commits over to the other repo? I don't want to lose them lol.

lukechampine:
Yep, you can use subtrees to move it. Once I merge this, I'll move it over. Then we need a commit to remove Sia's merkletree folder and import the other repo; do you want to handle that, or should I?

merkletree/merkletree.go
```diff
@@ -0,0 +1,288 @@
...
+import (
+	"bytes"
+	"hash"
+)
+
+// A Tree takes data as leaves and returns the merkle root. Each call to 'Push'
+// adds one leaf to the merkle tree. Calling 'Root' returns the Merkle root.
+// The Tree also constructs proof that a single leaf is a part of the tree. The
+// leaf can be chosen with 'SetIndex'.
+type Tree struct {
```
lukechampine: I guess the other issue with naming it merkletree is that `tree := merkletree.Tree` kinda stutters. But I don't expect people to be working with the `Tree` type directly anyway.

lukechampine:
hmm. My initial impression is that combining the "root" tree and the "proof" tree complicates both of them.

And please. We don't need a whole testing framework here. Stick with simple unit tests.

merkletree/merkletree.go
```diff
@@ -0,0 +1,288 @@
...
+	t.proveIndex = 0
+	t.proveSet = nil
+}
+
+// SetIndex resets the tree, and then sets the index for which a proof that the
+// element is in the Tree will be built.
+func (t *Tree) SetIndex(i int) {
+	t.Reset()
+	t.proveIndex = i
+}
```
lukechampine: Reseting the tree is surprising behavior. If I were a user, my instinct would be to create the tree first, then set the index, then create the proof.

I think I preferred it when the proof tree was separate. Then at least you could require the index at the beginning, when you created the tree.

merkletree/merkletree.go
```diff
@@ -0,0 +1,288 @@
...
+		root = join(t.hash, current.next.sum, root)
+		current = current.next
+	}
+	return root
+}
+
+// Prove returns a proof that the data at index 'proveIndex' is an element in
+// the current Tree. The proof will be invalid if any more elements are added
+// to the tree after calling Prove. The tree is left unaltered.
+func (t *Tree) Prove() (h hash.Hash, merkleRoot []byte, proveSet [][]byte, proveIndex int, numLeaves int) {
```
lukechampine: why does this return the hash? Because you need to pass the hash to `Verify`?

merkletree/merkletree.go
```diff
@@ -0,0 +1,288 @@
...
+	// index will have a height equal to the current length of the prove set.
+
+	// Iterate through all of the smaller subTrees and combine them.
+	current := t.head
+	sum := current.sum
+	for current.next != nil && current.next.height < len(proveSet) {
+		// Combine this subTree with the next subTree.
+		sum = join(t.hash, current.next.sum, sum)
+		current = current.next
+	}
```
lukechampine: This is just `Root`, yes?

merkletree/merkletree.go
```diff
@@ -0,0 +1,288 @@
...
+	for current != nil {
+		proveSet = append(proveSet, current.sum)
+		current = current.next
+	}
+	return t.hash, t.Root(), proveSet, t.proveIndex, t.currentIndex
+}
+
+// VerifyProof takes a merkle, a proveSet, and a proveIndex and returns true if
+// the first element of the prove set is a leaf of data in the merkle root.
+func VerifyProof(h hash.Hash, merkleRoot []byte, proveSet [][]byte, proveIndex int, numLeaves int) bool {
```
lukechampine: takes a what now?

lukechampine:
Maybe it's just because it's late, but the comments didn't help me much. Maybe if the buildproof process was explained all in one block comment, at a high level, I'd get it.

merkletree/merkletree.go
DavidVorick: The problem is that if you SetIndex to an index that's already passed, you can't do anything.

We could have it return an error if the tree hasn't been Reset(), that might be better.

merkletree/merkletree.go
DavidVorick: Yeah. It's not strictly necessary, but I figured it should give you all of the inputs to verify.

DavidVorick:
I'm doing my best but it's not really easy to explain. When we're in person I can draw diagrams and you can redo the docs. It'll be a lot easier if I'm talking to you.

As for 'whole test framework', there's not much left to do, I've spent most of my time adding comments. Just having trouble articulating what's happening and why it works.

Benchmarks will be like 15 loc, and the testing is really important because we want to be sure there are no mistakes.

merkletree/merkletree.go
DavidVorick: lol I did it the other way. s/proof/prove

heh. I'll let you do whatever naming changes and other finishing touches. I think the test suite is nearly comprehensive, so as long as that passes you shouldn't unintentionally introduce any errors.

DavidVorick:
About merging the trees: there's 100% code overlap between the proof tree and the standard root tree. All of the code that appears in the root tree also appears in the build tree.

But you can't have use the root tree functions in a proof tree because you need to interject logic mid-function. The proof tree would look almost identical to how it currently does if we pulled the root tree out. But you would have a cleaner root tree to stare at.

merkletree/merkletree.go
lukechampine: I mean, the whole point is that the person creating the proof is not the person verifying the proof. You wouldn't ever call `Prove()` and `Verify` in the same function, except during testing.

DavidVorick: okay

DavidVorick:
If you migrate Sia to the using the new merkle package, don't change any of the types in types.go. The only thing that should be affect is StorageProof, you'll have to manually move the first element in and out of StorageProof.Segment.

merkletree/merkletree_test.go
```diff
@@ -0,0 +1,403 @@
...
+	}
+}
+
+// BenchmarkTree64Sha256 runs a benchmark on creating a Merkle tree out of 4MB
+// using 64 bytes at a time, using sha256 as the hashing algorithm.
+func BenchmarkTree64Sha256(b *testing.B) {
+	data := make([]byte, 4*1024*1024)
+	rand.Read(data)
+	segmentSize := 64
+	tree := New(sha256.New())
```
lukechampine: these benchmarks should not include the time required to generate the data. Use `b.ResetTimer()`

DavidVorick:
I almost didn't test the reader functions, but I'm glad I did b/c there are 2 mistakes.

ReadFull can return EOF and ErrUnexpectedEOF, both of which are okay for what we're doing. I guess that justifies not wanting users to do it themselves.

I was also thinking, when the time comes, we should add files, because you can parallelize that. Plus there might be even more pitfalls that we're currently unaware of. I'm leaving it out for now though, b/c generally speaking we won't be using files anyway. (it gets encrypted in memory first)

DavidVorick:
Ok. DONE. All tests written, desired benchmarks written, comments written to my personal satisfaction.

If you've got any more things you want addressed, I can tackle those.

lukechampine:
good catch on the unexpected EOF.
I realized that since Hash is a Writer, you can use `io.CopyN` to copy bytes directly from the reader to the hash. Unfortunately, it seems to use a much larger buffer size than we do, so it winds up being less efficient.
One thing I was wondering about was making the Reader functions methods of the tree. The instead of calling `ReaderMerkleRoot(h, r, leafSize)`, you would call `tree := New(h).ReadFull(r, leafSize); tree.Root()`. The reasoning is that this more closely matches the "standard" (non-reader) workflow; just substitute `Push` for `ReadFull`. On the other hand, calculating the Merkle root of a reader is probably the operation people are going to be doing the most, so maybe we should make it as simple as possible.

DavidVorick:
I like the functions as they stand, I don't think there are many cases where you'll prefer to call New().Read() as opposed to ReaderMerkleRoot()

But, is the name ReaderMerkleRoot stuttering? merkletree.ReaderMerkleRoot()? Should we just leave it?

lukechampine:
okay, this looks alright. I understand buildproof at a high level, though the exact logic of choosing which of the two hashes to append eludes me.
One minor change would be to add a `Proof` type, to bundle up all those values.
`merkletree.ReaderMerkleRoot()` stutters a bit; maybe just `ReaderRoot` would be better.

Long term, I do think we're going to want to split the trees up, if only for efficiency reasons. It's true that there is some overlap between the two, but I think further optimization would lessen it. We'd also implement a "real" tree that kept all the hashes in memory, for people who want to do batch proofs. But this is sufficient for now. I have to keep reminding myself that Merkle trees shouldn't be our main focus right now; we're just ensuring that the implementation is correct.

DavidVorick:
I'm against having a proof type. For instance, the Hash interface returns a []byte, as opposed to a hash. Internal to Sia, the types are nice because it's protective, but for libraries it could be too restrictive.

DavidVorick:
Maybe we can set a flag or something for splitting the trees. I'd rather only have 1 exported type. But that's for a later time.

PR #365, opened 2015-03-02
Import merkletree package

lukechampine:
wow that +/- ratio.

The merkletree library isn't a perfect fit for our purposes; I had to modify it to use `uint64` for `proofIndex` and `numSegments`. I'm also not a fan of how the base segment is lumped in with the hash set; that caused me a bit of confusion.

I wound up implementing (in the crypto package) a Sia-specific `tree` type so that you can do things like `PushObject` (and `Root` returns a `crypto.Hash` instead of a `[]byte`). I'm pleased with how this cleaned up the code. I could take it a step further by making `PushObject` variadic; then you could just call it once with a bunch of objects. In practice this is only useful in one function though (`consensusSetHash`).


Makefile
```diff
@@ -66,6 +66,7 @@ dependencies:
 	go get -u github.com/stretchr/graceful
 	go get -u golang.org/x/crypto/twofish
 	go get -u golang.org/x/tools/cmd/cover
+	go get -u github.com/NebulousLabs/merkletree
```
DavidVorick: :)

DavidVorick:
looks good

Issue #362, opened 2015-02-26
Siad Testing


DavidVorick:
Using only siad, and where it makes sense, using only API calls, you should test the following things:

- [x] Can you start the miner? Does it mine blocks? Does the wallet balance change as it mines blocks? (with the test constants, the outputs are not given to you for 3 blocks, so mine at least 4 before checking that the balance has increased)
- [x] Can you send coins to another address? Does the first balance decrease? Upon receiving coins, does the balance increase?
- [x] Can the host announce itself to the blockchain? When it does, does the hostdb update and add the host to the hostdb?
- [x] Can the renter upload a file? Once it has uploaded, can it be downloaded? Can you check the status on the download in progress? Does the renter's balance decrease after uploading the file?
- [x] Does the host report having new contracts after a renter has uploaded a file? Once the file contract ends, does the host's balance increase appropriately?
- [x] Do unconfirmed transactions appear on everyone's machine? Does the transaction broadcasting work right?
- [x] Do blocks appear on everyone's machine? Does block broadcasting happen correctly?
- [x] Does a bootstrapping party catch up immediately? What if there are 2500 800kb blocks to download?
- [x] Do peers propagate well enough? If someone joins the network, will everyone get the new peer? Will people stop collecting peers before they are broadcasting to an unreasonable number of people?

I think that this will cover everything well enough. But I'm not sure how you're going to manage testing the stuff that happens asynchronously. We might need to do some of these tests outside of go.

DavidVorick:
Synchronize is broken. Mostly, it seems like it's not being called, but if you look at the synchronize function there seems to be TODOs that are incorrect.

DavidVorick:
So, using the channels was good for simple situations, but it started giving me problems, I think I was using them incorrectly or something (making calls but not checking the channels).

I've replaced it all with the spinning method of synchronizing for now (just checking and sleeping in a for loop). The disadvantage to that approach is that, while it'll make sure stuff works if everything is working, the output (a several hundred thread panic) isn't terribly useful in the event that things aren't correct. You don't get helpful information and the timer might explode somewhere that's not looping forever.

I think a better solution (if we continue using spinsync) is to have the spinsync call panic itself after a second or so, and then just not include any spinsync tests in the short tests. Then we at least see where the panic was called, which is going to be nearly as useful as a t.Error or t.Fatal

lukechampine:
[...]

PR #367, opened 2015-03-02
siad tests

lukechampine:
See #362 for a list of completed tests.

Preliminary testing revealed that a lot of siad internals were completely broken. Host, Renter, and Wallet did not satisfy their respective interfaces. Miner doesn't have an interface at all. The Gateway was calling `"AcceptBlock"` even though that RPC is registered to `"RelayBlock"`. `"SharePeers"` wasn't registered at all (it used to be registered by the TCPServer). I guess the moral is that RPCs are tricky business, although I don't know of a way to make them safer. There isn't really a way to check their correctness at compile time.

~~The mining tests are pretty hacky; they just start the miner and wait for a second. We need a much more reliable mechanism for mining, similar to the `ConsensusTester.MineCurrentBlock` method.~~
I added a naive function for accomplishing this. Maybe when you take it over you can make better use of the ConsensusTester.

~~There is one serious problem that I left unaddressed, which is that transactions cannot be relayed between synchronized peers. It seems that this causes an infinite loop: peer1 accepts a transaction and relays it via the Gateway to peer2. Then peer2 accepts the transaction and relays it back to peer1. The chain should stop there, but instead it seems peer1 will relay back to peer2 again, and the transaction will bounce between them forever. Perhaps the Transaction Pool is not rejecting duplicate transactions?~~
This was actually caused by synchronous broadcasts, which are fixed now. However, the fix was kind of hacky. It requires you to keep track of all the RPCs going on behind the scenes. Hopefully we can work out a better method soon.

### DIFF
--- a/consensus/currency.go
+++ b/consensus/currency.go
@@ -1,6 +1,7 @@
 package consensus
 
 import (
+	"fmt"
 	"math"
 	"math/big"
 )
@@ -127,4 +128,10 @@ func (c Currency) MarshalJSON() ([]byte, error) {
 // UnmarshalJSON implements the json.Unmarshaler interface.
 func (c *Currency) UnmarshalJSON(b []byte) error {
 	return c.i.UnmarshalJSON(b)
+}
+
+// Scan implements the fmt.Scanner interface, allowing Currency values to be
+// scanned from text.
+func (c *Currency) Scan(s fmt.ScanState, ch rune) error {
+	return c.i.Scan(s, ch)
 }
--- a/modules/gateway/gateway.go
+++ b/modules/gateway/gateway.go
@@ -87,6 +87,8 @@ func (g *Gateway) RelayTransaction(t consensus.Transaction) (err error) {
 
 // Info returns metadata about the Gateway.
 func (g *Gateway) Info() (info modules.GatewayInfo) {
+	g.mu.RLock()
+	defer g.mu.RUnlock()
 	for peer := range g.peers {
 		info.Peers = append(info.Peers, peer)
 	}
--- a/modules/gateway/gateway.go
+++ b/modules/gateway/gateway.go
@@ -81,6 +81,7 @@ func (g *Gateway) RelayBlock(b consensus.Block) (err error) {
 
 // RelayTransaction relays a transaction, both locally and to the network.
 func (g *Gateway) RelayTransaction(t consensus.Transaction) (err error) {
+	// no locking necessary
 	go g.threadedBroadcast("AcceptTransaction", t, nil)
 	return
 }
--- a/modules/gateway/gateway.go
+++ b/modules/gateway/gateway.go
@@ -50,7 +50,7 @@ func (g *Gateway) Bootstrap(bootstrapPeer network.Address) (err error) {
 	}
 
 	// announce ourselves to new peers
-	g.broadcast("AddMe", g.tcps.Address(), nil)
+	go g.threadedBroadcast("AddMe", g.tcps.Address(), nil)
 
 	return
 }
@@ -75,13 +75,13 @@ func (g *Gateway) RelayBlock(b consensus.Block) (err error) {
 		return errors.New("block added, but it does not extend the state height.")
 	}
 
-	g.broadcast("RelayBlock", b, nil)
+	go g.threadedBroadcast("RelayBlock", b, nil)
 	return
 }
 
 // RelayTransaction relays a transaction, both locally and to the network.
 func (g *Gateway) RelayTransaction(t consensus.Transaction) (err error) {
-	g.broadcast("AcceptTransaction", t, nil)
+	go g.threadedBroadcast("AcceptTransaction", t, nil)
 	return
 }
 
--- a/modules/gateway/helpers.go
+++ b/modules/gateway/helpers.go
@@ -40,19 +40,17 @@ func (g *Gateway) randomPeer() (network.Address, error) {
 // threadedBroadcast broadcasts an RPC to all of the Gateway's peers. The
 // calls are run in parallel.
 func (g *Gateway) threadedBroadcast(name string, arg, resp interface{}) {
-	// get peer list
 	g.mu.RLock()
-	peers := g.peers
-	g.mu.RUnlock()
-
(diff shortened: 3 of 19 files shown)


consensus/currency.go
```diff
@@ -128,3 +129,9 @@ func (c Currency) MarshalJSON() ([]byte, error) {
 func (c *Currency) UnmarshalJSON(b []byte) error {
 	return c.i.UnmarshalJSON(b)
 }
+
+// Scan implements the fmt.Scanner interface, allowing Currency values to be
+// scanned from text.
+func (c *Currency) Scan(s fmt.ScanState, ch rune) error {
```
DavidVorick: Does this work? Can an entire currency fit into a single rune?

I guess I just don't know.

modules/gateway/gateway.go
```diff
@@ -75,18 +75,20 @@ func (g *Gateway) RelayBlock(b consensus.Block) (err error) {
...
 	}
 
-	g.broadcast("RelayBlock", b, nil)
+	go g.threadedBroadcast("RelayBlock", b, nil)
 	return
 }
 
 // RelayTransaction relays a transaction, both locally and to the network.
 func (g *Gateway) RelayTransaction(t consensus.Transaction) (err error) {
-	g.broadcast("AcceptTransaction", t, nil)
```
DavidVorick: Convention is that all exported functions start with a

```
lock 
defer unlock
```,

 can you add a single comment:
```
// no locking necessary
```

File: modules/gateway/helpers.go [hunk 3]
```diff
@@ -36,10 +37,23 @@ func (g *Gateway) randomPeer() (network.Address, error) {
...
 
-func (g *Gateway) broadcast(name string, arg, resp interface{}) {
-	for peer := range g.peers {
-		peer.RPC(name, arg, resp)
+// threadedBroadcast broadcasts an RPC to all of the Gateway's peers. The
+// calls are run in parallel.
+func (g *Gateway) threadedBroadcast(name string, arg, resp interface{}) {
+	// get peer list
+	g.mu.RLock()
+	peers := g.peers
```
DavidVorick: This is not safe:
https://play.golang.org/p/aeXh_poP1i

siad/apiwallet.go
```diff
@@ -15,8 +15,9 @@ func (d *daemon) walletAddressHandler(w http.ResponseWriter, req *http.Request)
...
 	}
 
-	// TODO: Explain what's happening here, and why it doesn't look like all of
-	// the other calls.
+	// Since coinAddress is not a struct, we define one here so that writeJSON
+	// writes an object instead of a bare value. In addition, we transmit the
+	// coinAddress as a hex-encoded string rather than a byte array.
 	writeJSON(w, struct {
 		Address string
 	}{fmt.Sprintf("%x", coinAddress)})
```
DavidVorick: hmm. Interesting that we would write it as hex through the api. Half of me thinks that the front-end should be what worries about format, but the other half says that defaulting to hex for really simple calls is good.

I guess it's good to leave it for now

consensus/currency.go
lukechampine: haha no, it gets read from the ScanState, which is an `io.RuneScanner`. [Relevant source code](http://golang.org/src/math/big/int.go?s=12416:12466#L452)

siad/apiwallet.go
lukechampine: hmm. I never really thought about it, but you have a point. It looks like json will marshal a byte array as a json array, so it wouldn't be too difficult to work with. I don't know how easy it is to convert a json array to a hex-encoded string though.

modules/gateway/helpers.go
lukechampine: derp.

siad/daemon_test.go
```diff
@@ -1,24 +1,144 @@
...
 
 		HostDir: "hostDir",
 
-		Threads: 2,
+		Threads: 1,
 
 		DownloadDir: "downloadDir",
 
-		WalletDir: "walletDir",
+		WalletDir: "walletDir" + strconv.Itoa(APIPort),
```
lukechampine: this is pretty awful, it makes a bunch of walletDir directories. But it was necessary to prevent daemonTesters from reading from the same wallet file.

DavidVorick: yeah it's a huge eyesore. I was thinking we could have an in-directory folder, or we could just document somewhere that everything is going to /tmp and then put everything in /tmp.

modules/gateway/helpers.go
```diff
@@ -36,10 +37,21 @@ func (g *Gateway) randomPeer() (network.Address, error) {
...
+	wg.Add(len(g.peers))
 	for peer := range g.peers {
-		peer.RPC(name, arg, resp)
+		go func(peer network.Address) {
+			peer.RPC(name, arg, resp)
+			wg.Done()
+		}(peer)
 	}
+	// release lock while we wait for RPCs to complete
+	g.mu.RUnlock()
```
DavidVorick: This is still not good because you'll be holding the lock for at least as long as the timeout.

What you might need to do is add all of the peers into a slice.

```go
g.mu.RLock()
var peers []network.Address
for peer, _ := range g.peers {
    peers = append(peers, peer)
}
g.mu.RUnlock()
for _, peer := range peers {
    go func() {
        //
    }
}()
```

siad/daemon_test.go
```diff
@@ -1,24 +1,144 @@
...
+
+func (dt *daemonTester) acceptTransaction(t consensus.Transaction) error {
+	err := dt.tpool.AcceptTransaction(t)
+	dt.rpcChan <- struct{}{}
+	return err
+}
 
-	go d.listen(dc.APIAddr)
+// mineMoney mines 5 blocks, enough for the coinbase to be accepted by the
+// wallet. This may take a while.
```
DavidVorick: testing difficulty is set to ~4 hashes per block. Mining 5 blocks is still really fast. The bottleneck might actually be calling AcceptBlock()

modules/gateway/helpers.go
DavidVorick: oh nope this is fine

PR #371, opened 2015-03-04
Siad Testing

DavidVorick:
This PR hits all the goals for what we wanted to test.

Once we merge this, we can can start adding the final touches and getting ready to move things into the release branch. I think I would like to have 'master' look like idealized version of release by Friday.

Stuff is getting messy with the testing folders. At some point, we'll move it all to /tmp but for now we should focus on getting everything prepared for release.


modules/gateway/gateway.go
```diff
@@ -26,15 +26,21 @@ type Gateway struct {
 }
 
 // Bootstrap joins the Sia network and establishes an initial peer list.
+//
+// Bootstrap handles mutexes manually to avoid having a lock during network
+// communication.
+//
+// TODO: Peers are pinged sequentially!
```
lukechampine: this could be improved by using `broadcast`. I'll add it to my RPC PR.

modules/host/negotiate.go
```diff
@@ -143,6 +143,9 @@ func (h *Host) addCollateral(txn consensus.Transaction, terms modules.ContractTe
 	if err != nil {
 		return
 	}
+	if collateral.Cmp(consensus.NewCurrency64(0)) == 0 {
```
lukechampine: can use `consensus.ZeroCurrency` for this. I really ought to bring back `IsZero` though.

PR #372, opened 2015-03-05
API organization

DavidVorick:
I wanted to do a trial PR to see what you think. You can view the markdown on my branch to see what it looks like. I want to start finalizing the API for version 1.

When doing this I did get a few questions about updating:

1. is it hard to add signatures? B/c they are really important. Otherwise anyone can provide bogus updates and steal coins.

2. Will the updating work if you are more than 1 update out of date? From what I remember, you a manifest which contains diffs? But if you don't get all of the diffs since you last updated, you will run into problems. (or does the code already handle that well?)

3. I did change it so that the code is safe if version numbers have different volumes of dots.

### DIFF
--- a/API.md
+++ b/API.md
@@ -1,8 +1,52 @@
 Siad API
 ========
 
-Unless otherwise specified, API calls return the JSON object { "Success": true }.
-Errors are sent as plaintext, accompanied by an appropriate status code.
+All API calls return JSON objects. If there is an error, the error is returned
+in plaintext. The standard response is { "Success": true}. In this document,
+the API responses are defined as go structs. The go structs will get encoded to
+JSON before being sent. Go structs are used here to provide type information.
+
+Daemon
+------
+
+Queries:
+
+* /daemon/stop
+* /daemon/update/check
+* /daemon/update/apply
+
+#### /daemon/stop
+
+Function: Cleanly shuts down the daemon. May take a while.
+
+Params: none
+
+Response: standard
+
+#### /daemon/update/check:
+
+Function: Checks for an update, returning a bool indicating whether
+there is an update and a version indicating the version of the update.
+
+Params: none
+
+Response:
+```
+struct {
+	Available bool
+	Version string
+}
+```
+
+#### /daemon/update/apply:
+
+Function: Applies any updates that are available.
+
+Params: none
+
+Response: standard
+
+-----
 
 | Path              | Params                           | Response                     |
 |:------------------|:---------------------------------|:-----------------------------|
@@ -29,7 +73,7 @@ Errors are sent as plaintext, accompanied by an appropriate status code.
 
 HostInfo is a JSON object containing the following values:
 ```
-    TotalStorage"
+    "TotalStorage"
     "Minfilesize"
     "Maxfilesize"
     "Minwindow"
--- a/siad/daemonapi.go
+++ b/siad/daemonapi.go
@@ -13,6 +13,11 @@ import (
 
 const VERSION = "0.2.0"
 
+// TODO: Updates need to be signed!
+// TODO: Updating on Windows may not work correctly.
+// TODO: Will this code properly handle the case where multiple versions in a
+// row have been missed?
+
 // Updates work like this: each version is stored in a folder on a Linode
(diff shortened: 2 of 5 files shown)


API.md
```diff
@@ -1,8 +1,52 @@
...
+}
+```
+
+#### /daemon/update/apply:
+
+Params: none
+
+Response: standard
+
+Function: applies any updates that are available
```
lukechampine: these items are out of order

API.md
```diff
@@ -1,8 +1,52 @@
...
+
+Queries:
+
+* /daemon/stop
+* /daemon/update/check
+* /daemon/update/apply
+
+#### /daemon/stop
+
+Function: cleanly shuts down the daemon. May take a while
```
lukechampine: All Function descriptions should end in a period.

siad/daemonapi.go
```diff
@@ -22,26 +27,28 @@ const VERSION = "0.2.0"
...
+	nums := strings.Split(version, ".")
+	for i := range NUMS {
 		Ni, _ := strconv.Atoi(NUMS[i])
-		if ni != Ni {
-			return ni > Ni
+		ni, _ := strconv.Atoi(nums[i])
+		if Ni != ni {
+			return Ni < ni
+		}
+		if len(ni)-1 == i {
```
lukechampine: how can you say you "fixed" this code if you didn't even compile it? :P

lukechampine: wonder what this line was supposed to do...

siad/daemonapi.go
```diff
@@ -22,26 +27,28 @@ const VERSION = "0.2.0"
...
+			return false
 		}
 	}
-	// versions are equal
+	if len(NUMS) < len(nums) {
+		return true
+	}
```
lukechampine: just return the condition, `len(NUMS) < len(nums)`

API.md
DavidVorick: what if they are only one sentence? Should they all be full sentences (capital letter at the start?)

lukechampine: hmm. I'm gonna say yes on this one. Otherwise it ends up looking inconsistent. Imagine if we made "Function", "Response", etc. into headers. Then it'd be obvious that the description should be capitalized.

siad/daemonapi.go
DavidVorick: oh  yea, len(nums).

I can also change it to local and remote

siad/daemonapi.go
DavidVorick: magic

lukechampine: although I think if you do the truncation I mentioned above, you don't need this check anymore

DavidVorick: *writes tests*

lukechampine:
can you add a unit test for `newerVersion`? Should be really easy, I just want to make sure it's working as we expect it to.

DavidVorick:
lol

I did, look again :P

DavidVorick:
Closing until it's actually ready, I just wanted to PR a sneak peek so you could give your thoughts before I did the whole thing in that format.

siad/daemonapi_test.go
```diff
@@ -0,0 +1,39 @@
...
+	}
+	if !newerVersion("0.2.1") {
+		t.Error("Comparing to 0.2.1 should return true")
+	}
+	if !newerVersion("0.2.0.0") {
+		t.Error("Comparing to 0.2.0.0 should return true")
+	}
+	if !newerVersion("0.2.0.1") {
+		t.Error("Comparing to 0.2.0.1 should return true")
+	}
```
lukechampine: use a map lol

DavidVorick: what would that look like?

lukechampine:
```go
testcases := map[string]bool{
    VERSION: false,
    "0.1":   false,
    "0.1.1": false,
    "1":     true,
    //...
}

for version, expected := range testcases {
   if result := newerVersion(version); result != expected {
        t.Errorf("Wrong result for version %s: got %v, expected %v", version, result, expected)
   }
}
```

PR #373, opened 2015-03-06
Merge network into gateway (plus new RPCs)

lukechampine:
All network package functionality is now handled by the gateway.

RPCs must now have the function signature `func(modules.NetConn) error` (meaning: no more reflection!). Internally, the gateway uses a new monitored connection type that satisfies the `modules.NetConn` interface. This allows the gateway to terminate malicious-seeming connections and blacklist them. (Oh right, peers can be blacklisted now -- well, sort of. After a number of "strikes" they get removed from the peer list, but there's nothing preventing them from being added again). The logic that identifies a connection as "malicious" has not yet been implemented.

Gateway-specific RPCs like `SharePeers` and `AddMe` have been moved out of the Gateway interface and into the concrete type, along with the RPCs that the network used to own, like `Ping`.

The gateway package now stands at 78% test coverage, woo!

Notes:
- The `relayBlock` and `acceptTransaction` functions in daemon.go are temporary, and should be moved to their respective modules when the time comes.
- I tried creating a `GatewayTester` so that other modules wouldn't have to manage their own RPC ports, but I ran into problems. During testing, each module loads the gateway separately, so the port counter gets reset. So the only way to avoid port conflicts is to enforce that all tests must `Close` the gateway when they complete. We can still do this, but for now I felt it wasn't worth the effort.


DavidVorick:
78%! Huge upgrade from 14%, really glad to see that.

.travis.yml
```diff
@@ -5,7 +5,7 @@ go:
 
 install:
   - make dependencies
-  - $(exit $(go fmt ./... | wc -l))
+  - test -z "$(go fmt ./...)"
```
DavidVorick: Can you submit another PR where gofmt hasn't been run, to demonstrate that this command fails properly?

modules/gateway.go
```diff
@@ -1,12 +1,48 @@
...
+}
+
+// RPCFunc is the type signature of functions that handle incoming RPCs.
+type RPCFunc func(NetConn) error
+
 type GatewayInfo struct {
-	Peers []network.Address
+	Address NetAddress
+	Peers   []NetAddress
 }
```
DavidVorick: I think my comment is out of scope for this PR, but I do believe it would be cleaner to just throw a `Peers` function and an `Address` function into the interface. Info structs just seem dirty.

modules/gateway.go
```diff
@@ -37,19 +82,22 @@ type Gateway interface {
...
+	}
+}
+
+// WriterRPC returns a closure that can be passed to Gateway.RPC to write a
+// single value.
+func WriterRPC(obj interface{}) RPCFunc {
+	return func(conn NetConn) error {
+		return conn.WriteObject(obj)
+	}
+}
```
DavidVorick: Are these two functions used extensively? Because this file is full of stuff that we're going to want to support through the entirety of every major version.

I think this is more okay because it's entirely dependent on other code in this file. It'll work as long all the other parts of this file stay the same, and the whole worry is related to maintaining things outside of this file. It shouldn't be any more work to maintain the project whether this code is here or not.

modules/gateway/gateway.go
```diff
@@ -2,63 +2,90 @@ package gateway
...
+
+const (
+	// maxStrikes is the number of "strikes" that can be incurred by a peer
+	// before it will be removed.
+	// TODO: need a way to whitelist peers (e.g. hosts)
+	maxStrikes = 5
 )
 
 var (
+	// TODO: unexport these, or move them to modules
```
DavidVorick: I'm not sure which to do.

If we move them to modules, we'll be obligated to maintain them, which means we have little flexibility for adjusting the errors. At the same time, if we don't export them, then external forces have no way to check the error.

I think we should move the errors to modules if we're expecting other programs to have different behavior depending on which error gets returned. Otherwise I'd just unexport them.

modules/gateway.go
lukechampine: yeah, I figured we could split it up along with the others when we get rid of all the Info structs

modules/gateway.go
lukechampine: They're used extensively inside the gateway module, but little elsewhere. So it's probably fine to move them inside the gateway.

modules/gateway/gateway.go
```diff
@@ -105,27 +133,34 @@ func (g *Gateway) Info() (info modules.GatewayInfo) {
 }
 
 // New returns an initialized Gateway.
-func New(tcps *network.TCPServer, s *consensus.State) (g *Gateway, err error) {
-	if tcps == nil {
-		err = errors.New("gateway cannot use nil tcp server")
-		return
-	}
+func New(addr string, s *consensus.State) (g *Gateway, err error) {
```
DavidVorick: shouldn't the gateway determine it's own address? It needs a port number but not a whole address.

modules/gateway/gateway.go
```diff
@@ -105,27 +133,34 @@ func (g *Gateway) Info() (info modules.GatewayInfo) {
...
-	g = &Gateway{
-		tcps:  tcps,
-		state: s,
-		peers: make(map[network.Address]struct{}),
-	}
-	block, exists := g.state.BlockAtHeight(0)
-	if !exists {
-		err = errors.New("gateway state is missing the genesis block")
+	// create listener
+	l, err := net.Listen("tcp", addr)
```
DavidVorick: Can we rename `Listen` to `Listener`? When I see `Listen`, I think that it's a separate process that's doing listening, and that we don't get an object back. `NewListener` if the object is already called `Listener`.

modules/gateway/gateway.go
lukechampine: I can see them being used by clients, so that you can show the user a special "Your peer list is empty!" dialogue instead of a generic "no peers" error. But those clients generally won't have access to the modules package anyway, so it's a moot point. They don't seem terribly useful for other modules either. I think we should unexport them until we have a compelling reason to do otherwise.

modules/gateway/gateway.go
DavidVorick: ah darn. lol you didn't write the `net` package.

lukechampine: hahaha

DavidVorick: Can we rename `l` to `listener` then? I'm just confused, because we call `net.Listen`, and then we call `go g.listen()` later on, and the whole process is a bit opaque.

modules/gateway/gateway.go
DavidVorick: ok. Can you add that to this PR?

modules/gateway/gateway_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+import (
+	"strconv"
+	"testing"
+
+	"github.com/NebulousLabs/Sia/consensus"
+	"github.com/NebulousLabs/Sia/modules"
+)
+
+var port = 9001
```
DavidVorick: I very badly want to create some higher level testing package that'll allow us to get rid of this forever.

modules/gateway/gateway_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+func newTestGateway(t *testing.T) *Gateway {
+	g, err := New(":"+strconv.Itoa(port), consensus.CreateGenesisState())
+	if err != nil {
+		t.Fatal(err)
+	}
+	port++
+	return g
+}
```
DavidVorick: This seems like a good way to do things.

modules/gateway/gateway.go
lukechampine: hmm, I can see why that would be confusing. I could also move the `net.Listen` call into `g.listen`, though it makes the code a bit uglier (need to spawn a separate go func inside `g.listen`)

modules/gateway/gateway_test.go
lukechampine: haha me too, our port numbers are all over the place. I tried to fix it; see top comment.

modules/gateway/gateway.go
lukechampine: I suppose. I feel like there was a good reason why we weren't doing that though...

modules/gateway/helpers.go
```diff
@@ -1,57 +1,60 @@
...
-func (g *Gateway) addPeer(peer network.Address) error {
-	if _, exists := g.peers[peer]; exists {
-		return errors.New("peer already added")
-	}
-	g.peers[peer] = struct{}{}
-	return nil
+var pong = [4]byte{'p', 'o', 'n', 'g'}
+
+// Ping returns whether an Address is reachable and responds correctly to the
+// ping request -- in other words, whether it is a potential peer.
```
DavidVorick: Are there protections against other programs that might return 'pong'? If you're just calling "Ping" on a random ip address and port, it's not terribly far fetched that another program might return "pong". Perhaps we should add magic bytes here. ("siapong"?).

.travis.yml
DavidVorick: https://travis-ci.org/NebulousLabs/Sia/builds/53296291

thanks, looks good

modules/gateway/gateway.go
DavidVorick: hmm. Now that you mention it...

I can't think of what the reason was though.

modules/gateway/gateway_test.go
DavidVorick: close() might not be good enough anyway. iirc the test framework will run different package tests in parallel.

modules/gateway/peer_test.go
```diff
@@ -2,59 +2,143 @@ package gateway
...
-	err = myAddr.RPC("SharePeers", nil, &resp)
+// TestBadPeer tests that "bad" peers are correctly identified and removed.
+func TestBadPeer(t *testing.T) {
+	g := newTestGateway(t)
+	defer g.Close()
+
+	// create bad peer
+	badpeer := newTestGateway(t)
+	// overwrite badpeer's Ping RPC with an incorrect one
+	badpeer.RegisterRPC("Ping", modules.WriterRPC("lol"))
```
DavidVorick: badpeer is a troll.

"lol"
"lol"
"lol"

modules/gateway/peers.go
```diff
@@ -1,42 +1,129 @@
...
+		return ErrUnreachable
+	}
+	g.AddPeer(peer)
+	return nil
+}
+
+// sharePeers is an RPC that returns up to 10 randomly selected peers.
+func (g *Gateway) sharePeers(conn modules.NetConn) error {
+	g.mu.RLock()
+	defer g.mu.RUnlock()
```
DavidVorick: mutex conventions enforce that a non-exported function cannot lock or unlock the object mutex unless they are threaded functions.

modules/gateway/peers.go
```diff
@@ -1,42 +1,129 @@
...
+	return conn.WriteObject(peers)
+}
+
+// requestPeers calls the SharePeers RPC to learn about new peers. Each
+// returned peer is pinged to ensure connectivity, and then added to the peer
+// list. Each ping is performed in its own goroutine, which manages its own
+// mutexes.
+//
+// TODO: maybe iterate until we have enough new peers?
+func (g *Gateway) requestPeers(addr modules.NetAddress) error {
```
DavidVorick: mutex conventions enforce that a non-exported function should assume that there's a lock in place whenever it is called. requestPeers blocks while network communication is happening, which is an issue if there's a lock in place when it gets called.

modules/gateway/gateway.go
```diff
@@ -2,63 +2,90 @@ package gateway
...
+// Address returns the NetAddress of the Gateway.
+func (g *Gateway) Address() modules.NetAddress {
+	g.mu.RLock()
+	defer g.mu.RUnlock()
+	return g.myAddr
+}
+
+// Close stops the Gateway's listener process.
+func (g *Gateway) Close() error {
+	return g.listener.Close()
```
DavidVorick: do we need a lock here?

modules/gateway/rpc.go
```diff
@@ -0,0 +1,104 @@
...
+	defer conn.Close()
+	var id rpcID
+	if err := conn.ReadObject(&id, 8); err != nil {
+		// TODO: log error
+		return
+	}
+	// call registered handler for this ID
+	g.mu.RLock()
+	fn, ok := g.handlerMap[id]
+	g.mu.RUnlock()
```
DavidVorick: can't grab locks in nonexported functions.

looks like it just needs to be renamed 'threadedHandleConn".

Annoying to write, but helpful for preventing deadlocks.

modules/gateway/synchronize.go
```diff
@@ -24,7 +21,7 @@ func (g *Gateway) Synchronize() (err error) {
 	if err != nil {
 		return
 	}
-	go g.synchronize(peer)
+	g.synchronize(peer)
```
DavidVorick: We want Synchronize() to be a blocking function?

I thought we wanted people to be able to use the interface while startup was happening.

modules/gateway/helpers.go
lukechampine: it would be pretty easy to insert magic bytes into all network communication if need be. Like <10 line diff.

modules/gateway/gateway.go
lukechampine: nope; the `g.listener` is never modified, only created.

modules/gateway/synchronize.go
```diff
@@ -34,26 +31,93 @@ func (g *Gateway) Synchronize() (err error) {
...
-		acceptErr := g.state.AcceptBlock(block)
-		if acceptErr != nil {
-			// TODO: If the error is a FutureTimestampErr, need to wait before trying the
-			// block again.
+
+	// Find the most recent block from knownBlocks that is in our current path.
+	found := false
+	var start consensus.BlockHeight
+	for _, id := range knownBlocks {
+		if height, exists := g.state.HeightOfBlock(id); exists {
```
DavidVorick: need to check that the block is also in the current path, you can use `BlockAtHeight`

modules/gateway/synchronize.go
```diff
@@ -34,26 +31,93 @@ func (g *Gateway) Synchronize() (err error) {
...
+		// altogether.
+		return errors.New("no matching block found")
+	}
 
-	if err != nil && err.Error() == moreBlocksErr.Error() {
-		//fmt.Println("getting more blocks")
-		go g.synchronize(peer)
+	// Send blocks, starting with the child of the most recent known block.
+	//
+	// TODO: use BlocksSince instead?
```
DavidVorick: what do you mean by this TODO? I'm not seeing what would change.

modules/gateway/synchronize.go
```diff
@@ -34,26 +31,93 @@ func (g *Gateway) Synchronize() (err error) {
...
+	if err != nil {
 		return
 	}
+
+	// Indicate whether more blocks are available.
+	_, more := g.state.BlockAtHeight(start + MaxCatchUpBlocks)
+	return conn.WriteObject(more)
+}
+
+func (g *Gateway) requestBlocks(peer modules.NetAddress) (newBlocks []consensus.Block, moreAvailable bool, err error) {
```
DavidVorick: a lot of these functions are missing docstrings

modules/host/host.go
```diff
@@ -113,8 +113,10 @@ func (h *Host) SetSettings(settings modules.HostSettings) {
 }
 
 // Settings is an RPC used to request the settings of a host.
-func (h *Host) Settings() (modules.HostSettings, error) {
-	return h.HostSettings, nil
+func (h *Host) Settings(conn modules.NetConn) error {
+	h.mu.RLock()
+	defer h.mu.RUnlock()
+	return conn.WriteObject(h.HostSettings)
```
DavidVorick: don't hold a lock during network communication, need to make a copy of the data.

modules/gateway/synchronize.go
lukechampine: hmm. good point. I changed it to make testing easier, but I might have to revert it.

modules/renter/transfer.go
```diff
@@ -32,6 +30,7 @@ func downloadPiece(piece FilePiece, path string) error {
 		defer file.Close()
 
 		// Write the host's response into the file.
+		// TODO: use a tee reader here
```
DavidVorick: In the effort of minimizing TODOs, can you add this to the PR?

siad/gateway_test.go
```diff
@@ -28,8 +28,11 @@ func (dt *daemonTester) addPeer() *daemonTester {
 	for len(dt.gateway.Info().Peers) != len(newPeer.gateway.Info().Peers) {
 		time.Sleep(time.Millisecond)
 	}
+	// force synchronization to dt, in case newPeer tried to synchronize to
+	// one of dt's peers.
 	for dt.state.Height() != newPeer.state.Height() {
-		time.Sleep(time.Millisecond)
+		newPeer.gateway.Synchronize()
+		time.Sleep(10 * time.Millisecond)
```
DavidVorick: 10 milliseconds? lol

DavidVorick:
okay, I think I'm done with the comments. Last half wasn't so bad because it was mostly red.

modules/gateway/synchronize.go
lukechampine: I think I meant it would be nice to have a function similar to `BlockSince` that would return a slice of blocks. So you could call `s.BlocksInRange(start, start+MaxCatchUpBlocks)` and get back a `[]Block`, instead of doing this silly loop + map lookup business

modules/host/host.go
lukechampine: whoops, good catch

siad/gateway_test.go
lukechampine: this code bothers me haha. I'm considering making `Synchronize` take a peer, and moving it out of `Bootstrap`. It makes testing easier and it shouldn't impact usability much.

modules/gateway/gateway_test.go
lukechampine: well, you could throw a lock on the port. I actually tried that when I was trying to make the GatewayTester work, and it seemed to help a bit, but it didn't fix the core problem.

modules/gateway/gateway.go
lukechampine: I suppose one reason is that you could have the Gateway listening on a custom hostname. But the hostname will be overwritten during bootstrapping anyway, so...

modules/gateway/peers.go
lukechampine: RPCs are a special case, because they are, in fact, "exported" in the sense that they can be called externally (by another peer). So the locks are definitely needed. It definitely needs a docstring explaining why, though.

modules/gateway/peers.go
lukechampine: hmm. How about instead, `requestPeers` returns a slice of addresses? Then it doesn't need to hold a mutex.

modules/renter/transfer.go
lukechampine: done, but worth noting that this code isn't tested yet.

modules/gateway/helpers.go
DavidVorick: might be best

modules/gateway/peers.go
DavidVorick: as long as it's safe to be returning a slice to memory you don't want to be changed.

I don't know the structure well enough to know if it's a bad idea, I just know it might be a bad idea.

siad/gateway_test.go
DavidVorick: that sounds pretty reasonable. Selecting a random peer I don't think is the best strategy anyway, you probably want some priority setup or something.

modules/gateway/peers.go
lukechampine: `requestPeers` now runs in the same thread as `Bootstrap`, so there shouldn't be a locking issue. Having it return a slice might make it more clear where the locking happens, but I don't know if it's worth it now.

modules/gateway/rpc.go
lukechampine: done

modules/gateway/peers.go
DavidVorick: we could add a convention for RPCs then. rpcSharePeers, for example

modules/gateway/synchronize.go
lukechampine: fixed in 1a9a276368c667d4b87d79df05868bf8975d6305

modules/gateway/synchronize.go
lukechampine: fixed in 1a9a276368c667d4b87d79df05868bf8975d6305

PR #377, opened 2015-03-08
Atomic wallet saving

lukechampine:
This is untested, but only because the bug is difficult to reproduce (and hopefully impossible once this is merged).

The `String` method allows for proper printing when a currency is passed to things like `fmt.Println` or used with the `"%v"` format specifier.

### DIFF
--- a/modules/wallet/files.go
+++ b/modules/wallet/files.go
@@ -22,16 +22,21 @@ type savedKey struct {
 
 // save writes the contents of a wallet to a file.
 func (w *Wallet) save() (err error) {
-	// convert key map to a slice
+	// Convert the key map to a slice.
 	keySlice := make([]savedKey, 0, len(w.keys))
 	for _, key := range w.keys {
 		keySlice = append(keySlice, savedKey{key.secretKey, key.unlockConditions})
 	}
 
-	// Write the data to file.
-	// TODO: Instead of using WriteFile, write to a temp file and then do a move.
-	err = ioutil.WriteFile(w.filename, encoding.Marshal(keySlice), 0666)
+	// Write the data to a temp file
+	err = ioutil.WriteFile(w.filename+"_temp", encoding.Marshal(keySlice), 0666)
 	if err != nil {
+		return
+	}
+	// Atomically overwrite the old wallet file with the new wallet file.
+	err = os.Rename(w.filename+"_temp", w.filename)
+	if err != nil {
+		// TODO: instruct user to recover wallet from w.filename+"_temp"
 		return
 	}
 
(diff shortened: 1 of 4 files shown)


lukechampine:
btw, seems like `TestTransactionRelay` fails nondeterministically.

DavidVorick:
Can you post the errors that were seen?

modules/wallet/files.go
```diff
@@ -22,18 +22,23 @@ type savedKey struct {
...
-	// Write the data to file.
-	// TODO: Instead of using WriteFile, write to a temp file and then do a move.
-	err = ioutil.WriteFile(w.filename, encoding.Marshal(keySlice), 0666)
+	// Write the data to a temp file
+	err = ioutil.WriteFile(w.filename+"_temp", encoding.Marshal(keySlice), 0666)
 	if err != nil {
 		return
 	}
+	// Atomically overwrite the old wallet file with the new wallet file.
+	err = os.Rename(w.filename+"_temp", w.filename)
```
DavidVorick: rather than use rename, should we use copy?

modules/wallet/files.go
```diff
@@ -22,18 +22,23 @@ type savedKey struct {
...
+	if err != nil {
+		// TODO: instruct user to recover wallet from w.filename+"_temp"
```
DavidVorick: should the wallet automatically attempt the tmp file, and then just post a notice that corruption was detected and fixed? (Also copying the corrupted file somewhere safe?)

lukechampine:
```
--- FAIL: TestTransactionRelay (0.13 seconds)
	gateway_test.go:120: 0
	gateway_test.go:121: 0
	gateway_test.go:122: balances are incorrect for 0-conf transaction
```

modules/wallet/files.go
lukechampine: I suppose the advantage to `Copy` is that you always have two files, at least one of which is guaranteed to be uncorrupted. The downside is that it's much slower and non-atomic. I'm fairly certain that `Rename` will never leave the wallet file in a corrupted state though. If people want to make extra backups, they're free to do so.

modules/wallet/files.go
lukechampine: The `err` here never indicates corruption, only that the move failed, which I believe only happens if 1) one or both paths do not exist, or 2) there's a permissions problem. Since we just wrote `w.filename`, 1) will only occur if it somehow gets deleted in the interim, in which case `w.filename` will still exist. Likewise, 2) should only occur if the permissions of `w.filename` have been changed since the last `save`, in which case it's up to the user to fix them.

Maybe for extra security we should have `load` try to load `w.filename+"_temp"` if loading `w.filename` fails?

DavidVorick:
oh. I think I know how to fix it.

We just need to add another `for` loop.

modules/wallet/files.go
DavidVorick: yeah we should try loading the temp too before complaining.

DavidVorick: actually that might confuse the user, I'm not sure what the best behavior is.

PR #378, opened 2015-03-09
Make currentPath a slice

lukechampine:
This turned out to be simpler than expected, mostly because the syntax for map access and slice access is the same. The only trouble I had was with `s.height()`; originally I had it return `len(s.currentPath)`, which is 1 too big.

Just comparing test times, the speed-up seems pretty minimal, though I'd be interested in running a real benchmark. This is less of an efficiency thing and more of a simplicity thing anyway, though.

This made it easy to write `BlockRange`, which is now used in `Synchronize`. The semantics are a little weird because of the +1 thing. Maybe `height` should return `len(s.currentPath)` after all? Anyway, this is nice because `Synchronize` is now guaranteed to draw from `currentPath`.

### DIFF
--- a/consensus/fork.go
+++ b/consensus/fork.go
@@ -26,10 +26,14 @@ func (s *State) invalidateNode(node *blockNode) {
 // set of nodes between the common parent and 'bn', starting from the former.
 func (s *State) backtrackToCurrentPath(bn *blockNode) []*blockNode {
 	path := []*blockNode{bn}
-	for s.currentPath[bn.height] != bn.block.ID() {
+	for {
+		// stop when we reach the common parent
+		if bn.height <= s.height() && s.currentPath[bn.height] == bn.block.ID() {
+			break
+		}
+
 		bn = bn.parent
-		// prepend, not append
-		path = append([]*blockNode{bn}, path...)
+		path = append([]*blockNode{bn}, path...) // prepend, not append
 
 		// Sanity check - all block nodes should have a parent except the
 		// genesis block, and this loop should break before reaching the
--- a/consensus/fork.go
+++ b/consensus/fork.go
@@ -59,7 +59,7 @@ func (s *State) rewindToNode(bn *blockNode) {
 	}
 
 	// Rewind blocks until we reach 'bn'.
-	for s.currentBlockID != bn.block.ID() {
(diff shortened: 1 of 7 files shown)


consensus/info.go
```diff
@@ -151,15 +179,7 @@ func (s *State) BlocksSince(id BlockID) (removedBlocks, addedBlocks []BlockID, e
...
-	for height := path[0].height + 1; ; height++ {
-		if _, exists := s.currentPath[height]; !exists {
-			break
-		}
-
-		node := s.blockMap[s.currentPath[height]]
-		addedBlocks = append(addedBlocks, node.block.ID())
-	}
-
+	addedBlocks = s.currentPath[path[0].height+1:]
```
lukechampine: this code was pretty funny anyway... using the `BlockID` to lookup the `*blockNode` to get the `Block` to calculate the... `BlockID`.

consensus/fork.go
```diff
@@ -26,10 +26,14 @@ func (s *State) invalidateNode(node *blockNode) {
 // set of nodes between the common parent and 'bn', starting from the former.
 func (s *State) backtrackToCurrentPath(bn *blockNode) []*blockNode {
 	path := []*blockNode{bn}
-	for s.currentPath[bn.height] != bn.block.ID() {
+	for {
+		// stop when we reach the common parent
+		if bn.height <= s.height() && s.currentPath[bn.height] == bn.block.ID() {
+			break
+		}
```
lukechampine: the old `for` loop check was actually masking a `nil` comparison; while `bn.height > s.height()`, the map lookup will return `nil`. Can't pull that trick with slices.

consensus/info.go
```diff
@@ -18,17 +18,21 @@ type StateInfo struct {
 
 // blockAtHeight returns the block on the current path with the given height.
 func (s *State) blockAtHeight(height BlockHeight) (b Block, exists bool) {
-	bn, exists := s.blockMap[s.currentPath[height]]
-	if !exists {
-		return
+	exists = height <= s.height()
+	if exists {
```
DavidVorick: `if height <= s.height() {`

lukechampine: need to set the return value of `exists` :P

PR #381, opened 2015-03-10
API+

DavidVorick:
This contains changes from each of us. Most of the changes are me documenting the api and making it adhere to some conventions. I haven't documented the conventions yet.

This will break the UI, so it shouldn't be merged until I can update the UI and PR that too.

This PR contains all of the livetest changes as well.

My biggest concern at this point is that the renter doesn't have any way to communicate information about files. There's no indication of how expensive an upload will be. There's no indication of how much time a  file has remaining. There's no indication of how many hosts are online who have the file (the 'health'). There's no way to repair a file. There's no safety mechanism to prevent the same piece from ending up on one host multiple times. Finally, the hostdb doesn't update automatically (and it probably should).

After we fix the deadlocking problems, this I think is the highest priority.


modules/gateway.go
```diff
@@ -40,11 +25,33 @@ type NetConn interface {
...
 	Address NetAddress
 	Peers   []NetAddress
 }
 
+// TODO: Move this and it's functionality into the gateway package.
+var (
+	BootstrapPeers = []NetAddress{
+		"23.239.14.98:9988",
+	}
+)
```
lukechampine: uh...why is this here and not in siad

modules/transactionpool/standard.go
```diff
@@ -65,8 +65,8 @@ func (tp *TransactionPool) IsStandardTransaction(t consensus.Transaction) (err e
 	// Check that any file contracts do not start for at least
 	// FileContractConfirmWindow blocks.
 	for _, fc := range t.FileContracts {
-		if fc.Start > tp.state.Height()-FileContractConfirmWindow {
-			return errors.New("file contract cannot start so close to the current height.")
+		if fc.Start < tp.state.Height()+FileContractConfirmWindow {
+			return errors.New("file contract cannot start so close to the current height...")
```
lukechampine: errors shouldn't end in punctuation

siad/API.md
```diff
@@ -0,0 +1,387 @@
+Siad API
+========
+
+All API calls return JSON objects. If there is an error, the error is returned
+in plaintext. The standard response is { "Success": true}. In this document,
+the API responses are defined as go structs. The go structs will get encoded to
+JSON before being sent. Go structs are used here to provide type information.
```
lukechampine: s/go/Go

siad/API.md
```diff
@@ -0,0 +1,387 @@
...
+
+Function: Will add a peer to the gateway.
+
+Parameters:
+```
+struct {
+	Address string
+}
+```
+Address should be an IP address of the form: "a.b.c.d:port"
```
lukechampine: technically there are no such restrictions; all that matters is that the host is pingable. Perhaps more accurate would be:
```
Address should be a reachable IP address or hostname, along with a port number. Examples:
    123.123.123.123:9001
    [2001:0DB8:AC10:FE01::]:9001
    domain.example:9001
```

same goes for `remove`

modules/gateway.go
DavidVorick: my rationale is that it should be up to the gateway to decide how & where the bootstrapping happens. It's whole job is to find other peers and stay synchronized, that should include the bootstrap peers. We don't have siad tell consensus what the genesis block is.

modules/transactionpool/standard.go
DavidVorick: leftover from debugging, forgot to clean it up

siad/API.md
DavidVorick: does siad work over IPv6?

that's pretty cool. golang is great.

siad/API.md
```diff
@@ -0,0 +1,387 @@
...
+```
+struct {
+	Files []string
+}
+```
+Files is a list of all the files by their nickname.
+
+#### /renter/upload
+
+Function: Upload a file using a datastream.
```
lukechampine: This section is special. The `Source` field is kind of misleading, because it isn't sent along with the `Nickname`, but rather as a "multipart file."

Hopefully this call will be removed soon anyway though.

siad/API.md
```diff
@@ -0,0 +1,387 @@
...
+
+Function: Get the status of the wallet.
+
+Parameters: none
+
+Response:
+```
+struct {
+	Balance      int
+}
```
lukechampine: fix spacing

siad/api.go
```diff
@@ -87,11 +89,17 @@ func (d *daemon) listen() error {
...
+	log.Printf("%d HTTP ERROR: %s", err, msg)
+	http.Error(w, msg, err)
+}
+
 // writeJSON writes the object to the ResponseWriter. If the encoding fails, an
 // error is written instead.
 func writeJSON(w http.ResponseWriter, obj interface{}) {
 	if json.NewEncoder(w).Encode(obj) != nil {
-		http.Error(w, "Failed to encode response", 500)
+		http.Error(w, "Failed to encode response", http.StatusInternalServerError)
```
lukechampine: :+1:

siad/API.md
DavidVorick: yeah this call is rough. I think I'm just going to remove the documentation for it.

siad/daemonapi.go
```diff
@@ -98,3 +96,36 @@ func applyUpdate(version string) (err error) {
 
 	return
 }
+
+// daemonStopHandler handles the api call to stop the daemon cleanly.
+func (d *daemon) daemonStopHandler(w http.ResponseWriter, req *http.Request) {
+	writeSuccess(w)
+
+	// send stop signal
+	d.apiServer.Stop(1e9) // TODO: What's with the constant?
```
lukechampine: yeah, should be `time.Second`, I just didn't feel like importing `time` :P

DavidVorick: \>.<

DavidVorick: I fixed it.

siad/gateway_test.go
```diff
@@ -43,7 +43,7 @@ func TestPeering(t *testing.T) {
 	// Create to peers and add the first to the second.
 	peer1 := newDaemonTester(t)
 	peer2 := newDaemonTester(t)
-	peer1.callAPI("/gateway/peer/add?addr=" + string(peer2.netAddress()))
+	peer1.callAPI("/gateway/peer/add?Address=" + string(peer2.netAddress()))
```
lukechampine: not a fan of capitalization. It seems uncommon in other APIs, and we don't want to be surprising.

One thing we could do is normalize the query strings to lower case before processing them. I'm not totally sure how that would work though.

siad/main.go
```diff
@@ -111,8 +113,10 @@ func startEnvironment(*cobra.Command, []string) {
 	}
 
 	// join the network
+	//
+	// TODO: This should proabably happen in newDaemon.
```
lukechampine: perhaps. You'd have to add a `NoBootstrap` field to `daemonConfig`. And idk, I like that `newDaemon` just creates the daemon and nothing else. We're already calling `d.listen` in main anyway.

siad/main.go
```diff
@@ -111,8 +113,10 @@ func startEnvironment(*cobra.Command, []string) {
...
 	if !config.Siacore.NoBootstrap {
-		go d.bootstrap()
+		go d.gateway.Bootstrap(modules.BootstrapPeers[0])
```
lukechampine: perhaps. You'd have to add a `NoBootstrap` field to `daemonConfig`. And idk, I like that `newDaemon` just creates the daemon and nothing else. We're already calling `d.listen` in main anyway.

siad/miner.go
```diff
@@ -5,40 +5,38 @@ import (
...
 	if err != nil {
-		writeError(w, "Failed to encode status object", 500)
+		writeError(w, "Failed to encode status object", http.StatusInternalServerError)
 		return
 	}
 	writeJSON(w, mInfo)
 }
 
-// minerStopHandler stops the miner.
+// minerStopHandler handles the api call to stop the miner.
```
lukechampine: s/api/API everywhere

modules/gateway.go
lukechampine: true...but we could :P
Alternatively, we could move the hard-coded addresses out of the code entirely, and load them from the config file instead.

siad/API.md
lukechampine: it should, though I haven't formally tested it yet.

siad/main.go
DavidVorick: d.listen isn't called in New?

I definitely think that New should return something that's fully initialized and running. I also think that `NoBootstrap` belongs in `daemonConfig`, in the gateway config section.

siad/miner.go
DavidVorick: hmph

siad/gateway_test.go
DavidVorick: we are stuck between two standards. One is the way that the web typically works and the other is the way that Go typically works.

I guess this is slightly different because it's not in json, it's in a url. For json, I would definitely defer to go but here we're doing the conversion anyway.

I guess I'll make it all lowercase.

siad/miner.go
DavidVorick: I'm not sure I agree with this one.

lukechampine: haha wat, it's an initialism. Why shouldn't it be capitalized?

DavidVorick: I've been using lowercase everywhere. api, json, and go. I fixed the go to Go because go means multiple things, but it's pretty clear what api and json mean, and I find that have a giant string of capital words in the middle of a sentence can disrupt the flow. You're talking about the functionality of the listener and how it uses JSON to yadda yadda etc.

idk.

siad/main.go
lukechampine: I guess that's fine too. I mean, when `daemon.New` initializes all the modules, their `New` functions are often doing additional internal setup anyway.

siad/miner.go
lukechampine: I file it under "comments should be full, grammatically correct sentences." If our comments are going to start with a capital and end with a period, they should obey other standard grammar conventions as well.

DavidVorick: that's true. I also just don't want fix it atm, because you can't find/replace since it's used in variable names. It's also in multiple packages, and in many files of those packages.

Maybe it can be filed under "we'll fix it as we see it, but until we do a full code review we won't worry about it too much"?

lukechampine: >you can't find/replace since it's used in variable names

just use `s/api /API /g`.

>It's also in multiple packages, and in many files of those packages

nope, only in siad.

PR #384, opened 2015-03-11
Fix wallet saving

lukechampine:
`os.Rename` fails on Windows if the destination already exists. Now we simply create a backup and then overwrite the old file.

### DIFF
--- a/modules/wallet/files.go
+++ b/modules/wallet/files.go
@@ -27,16 +27,17 @@ func (w *Wallet) save() (err error) {
 	for _, key := range w.keys {
 		keySlice = append(keySlice, savedKey{key.secretKey, key.unlockConditions})
 	}
+	walletData := encoding.Marshal(keySlice)
 
-	// Write the data to a temp file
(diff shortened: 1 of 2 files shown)


modules/wallet/files.go
```diff
@@ -27,16 +27,18 @@ func (w *Wallet) save() (err error) {
 	for _, key := range w.keys {
 		keySlice = append(keySlice, savedKey{key.secretKey, key.unlockConditions})
 	}
+	walletData := encoding.Marshal(keySlice)
 
-	// Write the data to a temp file
-	err = ioutil.WriteFile(w.filename+"_temp", encoding.Marshal(keySlice), 0666)
+	// Write the wallet data to a backup file, in case something goes wrong
+
+	err = ioutil.WriteFile(w.filename+".backup", walletData, 0666)
```
DavidVorick: looks like the comment here has an extra newline, or was that intentional?

lukechampine: whoops, not sure how that happened...

PR #385, opened 2015-03-11
/consensus/status returns more sane types

DavidVorick:
meant to get this in last night, just forgot.


DavidVorick:
hmm. Looks like send is failing nondeterministically. Not sure what to do about that. I think the deadlock takes higher priority for now.

I'm planning on working on all of this soon.

PR #386, opened 2015-03-11
renter can now provide more thorough file status

DavidVorick:
This ended up requiring a fair amount of changes. I also realized that the renter was never calling save or load.

One thing that I left unfinished was that the renter's file is never set. This may result in errors (perhaps silent) while the renter is running.

### DIFF
--- a/modules/renter/files.go
+++ b/modules/renter/files.go
@@ -1,43 +1,76 @@
 package renter
 
 import (
-	"io/ioutil"
-
-	"github.com/NebulousLabs/Sia/encoding"
+	"github.com/NebulousLabs/Sia/consensus"
+	"github.com/NebulousLabs/Sia/modules"
 )
 
-type savedFilePieces struct {
-	FilePieces []FilePiece
-	Nickname   string
+// A file is a single file that has been uploaded to the network.
+type File struct {
+	nickname    string
+	pieces      []FilePiece
+	startHeight consensus.BlockHeight
+
+	renter *Renter
 }
 
-func (r *Renter) save(filename string) (err error) {
-	// create slice of savedFilePieces
-	savedPieces := make([]savedFilePieces, 0, len(r.files))
-	for nickname, pieces := range r.files {
-		savedPieces = append(savedPieces, savedFilePieces{pieces, nickname})
-	}
-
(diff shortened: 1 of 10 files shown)


DavidVorick:
why didn't the build run? How do we get the build to run?

modules/renter/files.go
```diff
@@ -1,43 +1,68 @@
...
+func (f *File) Available() bool {
+	f.renter.mu.RLock()
+	defer f.renter.mu.RUnlock()
+
+	for _, piece := range f.pieces {
+		if piece.Active {
+			return true
+		}
 	}
+	return false
```
lukechampine: I think you meant:
```go
for _, piece := range f.pieces {
    if !piece.Active {

File: siad/renter.go [hunk 2]
```diff
@@ -25,6 +32,22 @@ func (d *daemon) renterDownloadHandler(w http.ResponseWriter, req *http.Request)
...
+// renterFilesHandler handles the API call to list all of the files.
+func (d *daemon) renterFilesHandler(w http.ResponseWriter, req *http.Request) {
+	files := d.renter.FileList()
+	fileSet := make([]FileInfo, 0, len(files))
+	for _, file := range files {
+		fileSet = append(fileSet, FileInfo{
+			Available:     file.Available(),
+			Nickname:      file.Nickname(),
+			Repairing:     file.Repairing(),
+			TimeRemaining: int(file.TimeRemaining()),
```
lukechampine: why cast to an `int`?
The name is a bit confusing; I thought for a second it might mean a `time.Duration`. Maybe "BlocksRemaining" would be better.

modules/renter/files.go
DavidVorick: It's supposed to return true if even a single piece is active

siad/renter.go
DavidVorick: hmmm. okay.

Right now it also gets initialized improperly.

Do you want me to remove the conversion to an int?

modules/renter/files.go
lukechampine: oh, right. I forgot that the pieces were fully redundant

siad/renter.go
lukechampine: yeah

DavidVorick:
As long as the build passes, the problems I noticed should be fixed.

PR #389, opened 2015-03-12
Safelocks

DavidVorick:
This PR depends on the other one I have open. (#388)

I created a lock package, it records who locked it and it complains if there's a deadlock, and it also automatically fixes the deadlock. It also prints a call stack, so you can see who caused the deadlock.

The current implementation results in thousands of sleeping goroutines during standard Sia use. I think that we can implement a garbage collection style thread, so that instead of thousands of sleeping goroutines, you just have 1 that runs every 2 seconds or so and checks the timestamps on the lock calls to see if any of them have held a lock for too long. Then you'll only have 1 thread per unique lock, which should really only be one thread per module. This is much more manageable.

As best I can tell, using the lock package is impacting performance a noticeable amount, but not a significant amount. For now, I have no real desire to implement the garbage collection style manager. Probably best to leave it.

This PR also fixes a few deadlocks, which were discovered thanks to the new mutex type. Debugging was _substantially_ easier once you knew which call wasn't releasing in time. As best I can tell, there's still probably a deadlock somewhere in the transaction pool but I haven't been able to trigger it through testing.

I think that, until performance becomes absolutely necessary, we should leave all production binaries with the protected locking. In the event that a deadlock is triggered, it'll automatically fix itself instead of locking up the program and affecting user experience. I think that we should probably just keep it indefinitely, but then also provide 'performance' binaries that have a lot of the safeties removed.

Right now `I call fmt.Printf` to talk about any deadlocks that occur. It should probably instead go into a logging file somewhere. I'm not sure the best way to handle that. Also not worried about it for the time being. I'm more concerned about getting the final UX stuff ready. (files/directories, indications of failed uploads/downloads, auto-repair, and some indication for the host of how much money total has been made from offering storage).

### DIFF
--- a/sync/lock.go
+++ b/sync/lock.go
@@ -112,7 +112,7 @@ func (rwm *RWMutex) safeUnlock(read bool, counter int) {
 		} else {
 			lockType = ""
 		}
-		fmt.Printf("A %v lock was held until deadlock, subsequent call to %v unlock failed. id '%v'. Call stack:\n", lockType, lockType, counter)
+		fmt.Printf("A%v lock was held until deadlock, subsequent call to%v unlock failed. id '%v'. Call stack:\n", lockType, lockType, counter)
 		for i := 0; i <= rwm.callDepth; i++ {
 			fmt.Printf("\tFile '%v', Line '%v'\n", callingFiles[i], callingLines[i])
 		}
--- a/sync/lock.go
+++ b/sync/lock.go
@@ -10,7 +10,7 @@ import (
 // RWMutex provides locking functions, and an ability to detect and remove
 // deadlocks.
 type RWMutex struct {
-	openLocks        map[int]struct{}
+	openLocks        map[int]lockInfo
 	openLocksCounter int
 	openLocksMutex   sync.Mutex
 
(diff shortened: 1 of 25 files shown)


DavidVorick:
I am led to believe that the locking is actually slow enough to be causing problems. Apparently having 7000 goroutines all sleeping is hard on the scheduler or something.

I think I'm going to implement the single goroutine method and see if there are noticeable improvements. Then I might try profiling it.

After the safelock stuff was applied to the gateway, the short tests starting taking 3s+ for siad. This is pretty sigificant, since it was under 1s before I added them to the gateway. Idk if the gateway is being particularly taxing or what, but maybe you have some ideas for why safelocking the gateway suddenly slowed everything down.

DavidVorick:
oh wait it was running that slow before I added the gateway.

lukechampine:
You must have rebased or something because the old hostdb commits have different hashes.

I'm going to go ahead and revert the hostdb PR. When this gets merged it'll have those changes anyway.

DavidVorick:
this is fine I'll just re-rebase

sync/lock.go
```diff
@@ -0,0 +1,157 @@
...
+	// Check if a deadlock has been detected and fixed manually.
+	_, exists := rwm.openLocks[counter]
+	if !exists {
+		var lockType string
+		if read {
+			lockType = "read "
+		} else {
+			lockType = ""
+		}
+		fmt.Printf("A%v lock was held until deadlock, subsequent call to%v unlock failed. id '%v'. Call stack:\n", lockType, lockType, counter)
```
lukechampine: missing a space here

sync/lock.go
```diff
@@ -0,0 +1,157 @@
...
+
+	rwm.openLocksMutex.Lock()
+	defer rwm.openLocksMutex.Unlock()
+
+	// Check if a deadlock has been detected and fixed manually.
+	_, exists := rwm.openLocks[counter]
+	if !exists {
+		var lockType string
+		if read {
+			lockType = "read "
```
lukechampine: should this be `" read"`?

sync/lock.go
```diff
@@ -0,0 +1,157 @@
...
+		} else {
+			lockType = ""
+		}
+		fmt.Printf("A%v lock was held until deadlock, subsequent call to%v unlock failed. id '%v'. Call stack:\n", lockType, lockType, counter)
+		for i := 0; i <= rwm.callDepth; i++ {
+			fmt.Printf("\tFile '%v', Line '%v'\n", callingFiles[i], callingLines[i])
```
lukechampine: typical debug convention is just `file:line`.

Really you should be using `runtime.FuncForPC` to get the name of the calling function. Then you could report something like:
`(*State).Height (info.go:141)`

sync/lock.go
DavidVorick: I thought I got rid of that code. I guess I left it in the unlock part.

I think you're right.

sync/lock.go
DavidVorick: hmm.

I like the way it is now because it closely resembles the way a panic looks. I will change it to `file:line` but I'm not sure I want to report like the above alternative.

Issue #387, opened 2015-03-11
Sia Directory


DavidVorick:
Right now the way all of the files + testing works is haphazard and not fully implemented. On launch, a sia folder should be checked for and then created. The daemon should pick this folder based on some configuration, and then it should tell each module a file to use based on that configuration. For the time being, we don't need to support modules that go in different directories, we should just have it all under one folder, and let the daemon manage giving filepaths to each of the modules. The daemon should pass in 'headfolder/modulename'. If the module wants to add a file extention (like .wallet or .dat), the module will worry about that.

Each module should manage its own folder or file. It'll be passed something (like 'renter') and then it'll decide if it wants an entire folder or just a file. My vote is that by convention they make their own folder, even if there's only one file inside.

The module then checks that location for previous data and loads it. Then, every X often (right now every time anything changes at all), the module will save to it's folder or file. This should follow the theme of each module being self-contained.

PR #392, opened 2015-03-13
Persistance

lukechampine:
Persistence has been implemented for the gateway, renter, and host.

The daemon now makes use of a general `siaDir`, where (as per #387) each module stores its configuration data in a subfolder.

This works fairly well, except for testing. As it stands, the tests are using `""` as their module subfolder, which results in a lot of .dat files littering the repo. I didn't bother to add all of these to the `.gitignore` and `make clean`; I think we should instead focus on cleaning up our testing framework so that all of these files have a sane place to live.


siad/main.go
```diff
@@ -147,16 +123,10 @@ func main() {
...
-	defaultHostDir := filepath.Join(siaDir, "hostdir")
-	defaultDownloadDir := "~/Downloads"
-	defaultWalletFile := filepath.Join(siaDir, "sia.wallet")
 	root.PersistentFlags().StringVarP(&config.Siad.APIaddr, "api-addr", "a", "localhost:9980", "which host:port is used to communicate with the user")
 	root.PersistentFlags().StringVarP(&config.Siacore.RPCaddr, "rpc-addr", "r", ":9988", "which port is used when talking to other nodes on the network")
 	root.PersistentFlags().BoolVarP(&config.Siacore.NoBootstrap, "no-bootstrap", "n", false, "disable bootstrapping on this run")
 	root.PersistentFlags().StringVarP(&config.Siad.ConfigFilename, "config-file", "c", defaultConfigFile, "location of the siad config file")
-	root.PersistentFlags().StringVarP(&config.Siacore.HostDirectory, "host-dir", "H", defaultHostDir, "location of hosted files")
-	root.PersistentFlags().StringVarP(&config.Siad.DownloadDirectory, "download-dir", "d", defaultDownloadDir, "location of downloaded files")
-	root.PersistentFlags().StringVarP(&config.Siad.WalletFile, "wallet-file", "w", defaultWalletFile, "location of the wallet file")
```
DavidVorick: It's nice to see the main get cleaned up a lot.

DavidVorick:
This looks good.

PR #394, opened 2015-03-13
Renter upgrades

DavidVorick:
The download call is no longer blocking.

The /renter/status call is no longer documented (it's deprecated)

There's a new API call /renter/downloadqueue which returns the list of downloads that the renter has made, including information about whether or not they've completed.

Finally, downloading does the same try-wait strategy as uploading now. If a download fails, it'll wait some arbitrary amount of time before trying again.

The download queue is not saved to disk. There's no way to remove items from the download queue except by restarting the renter.

The renter still does not remove files that have expired. For the time being, I consider this sufficient.

### DIFF
--- a/modules/renter/download.go
+++ b/modules/renter/download.go
@@ -1,12 +1,18 @@
 package renter
 
 import (
+	"crypto/rand"
 	"errors"
 	"io"
 	"os"
+	"time"
 
 	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/modules"
+)
+
+var (
+	downloadAttempts = 5
 )
 
 // downloadPiece attempts to retrieve a file from a host.
@@ -65,15 +71,19 @@ func (r *Renter) Download(nickname, filename string) error {
 
 	// We only need one piece, so iterate through the hosts until a download
 	// succeeds.
-	for _, piece := range pieces {
-		downloadErr := r.downloadPiece(piece, filename)
-		if downloadErr == nil {
-			return nil
-		} else {
-			// log error
+	go func() {
+		for i := 0; i < downloadAttempts; i++ {
+			for _, piece := range pieces {
+				downloadErr := r.downloadPiece(piece, filename)
+				if downloadErr == nil {
+					return
+				}
+			}
+			randSource := make([]byte, 1)
+			rand.Read(randSource)
+			time.Sleep(time.Second * time.Duration(i) * time.Duration(i) * time.Duration(randSource[0]))
 		}
-		// r.hostDB.FlagHost(piece.Host.IPAddress)
-	}
+	}()
 
-	return errors.New("Too many hosts returned errors - could not recover the file")
+	return nil
 }
--- a/modules/renter/download.go
+++ b/modules/renter/download.go
@@ -1,0 +1,79 @@
+package renter
+
+import (
+	"errors"
+	"io"
+	"os"
+
+	"github.com/NebulousLabs/Sia/crypto"
+	"github.com/NebulousLabs/Sia/modules"
+)
+
+// downloadPiece attempts to retrieve a file from a host.
+func (r *Renter) downloadPiece(piece FilePiece, path string) error {
+	return r.gateway.RPC(piece.HostIP, "RetrieveFile", func(conn modules.NetConn) (err error) {
+		// Send the id of the contract for the file piece we're requesting. The
+		// response will be the file piece contents.
+		if err = conn.WriteObject(piece.ContractID); err != nil {
+			return
+		}
+
+		// Create the file on disk.
+		file, err := os.Create(path)
+		if err != nil {
+			return
+		}
+		defer file.Close()
+
+		// Simultaneously download file and calculate its Merkle root.
+		tee := io.TeeReader(
+			// use a LimitedReader to ensure we don't read indefinitely
+			io.LimitReader(conn, int64(piece.Contract.FileSize)),
+			// each byte we read from tee will also be written to file
+			file,
+		)
+		merkleRoot, err := crypto.ReaderMerkleRoot(tee)
+		if err != nil {
+			return
+		}
+
+		if merkleRoot != piece.Contract.FileMerkleRoot {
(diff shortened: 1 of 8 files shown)


modules/renter/download.go
```diff
@@ -0,0 +1,150 @@
...
+
+// DownloadQueue returns the list of downloads in the queue.
+func (r *Renter) DownloadQueue() []modules.DownloadInfo {
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+
+	downloads := make([]modules.DownloadInfo, len(r.downloadQueue))
+	for i := 0; i < len(r.downloadQueue); i++ {
+		downloads[i] = &r.downloadQueue[i]
+	}
```
lukechampine: `for i := range r.downloadQueue`

modules/renter/download.go
```diff
@@ -0,0 +1,150 @@
...
+	return downloads
+}
+
+// Getter for the completed status of the download.
+func (d *Download) Completed() bool {
+	d.renter.mu.RLock()
+	defer d.renter.mu.RUnlock()
+	return d.completed
+}
```
lukechampine: it's a bit weird that these have to lock the renter, though I understand why.

modules/renter/download.go
```diff
@@ -0,0 +1,150 @@
...
+)
+
+// A Download is a download that has been queued by the renter.
+type Download struct {
+	completed   bool
+	destination string
+	nickname    string
+
+	renter *Renter
+}
```
lukechampine: does this type need to be exported? It's just satisfying an interface, right?

modules/renter/download.go
```diff
@@ -0,0 +1,150 @@
...
+
+var (
+	downloadAttempts = 5
+)
```
lukechampine: any reason why this isn't a `const`?

modules/renter/download.go
DavidVorick: Alternatively I could give them their own lock, but that seemed excessive especially if using safelocks.

modules/renter/download.go
```diff
@@ -0,0 +1,150 @@
...
+					r.mu.Unlock()
+					return
+				}
+			}
+
+			// This iteration failed, no hosts returned the piece. Try again
+			// after waiting a random amount of time.
+			randSource := make([]byte, 1)
+			rand.Read(randSource)
+			time.Sleep(time.Second * time.Duration(i) * time.Duration(i) * time.Duration(randSource[0]))
```
lukechampine: `time.Duration(i*i)` also works

modules/renter/upload.go
```diff
@@ -68,15 +36,23 @@ func (r *Renter) threadedUploadPiece(up modules.UploadParams, piece *FilePiece)
 		// file will be uploaded and we'll be done.
 		contract, contractID, err := r.negotiateContract(host, up)
 		if err != nil {
+			// The previous attempt didn't work. We will try again after
+			// sleeping for a randomized amount of time to increase our chances
+			// of success. This will help spread things out if there are
+			// problems with network congestion or other randomized issues.
+			randSource := make([]byte, 1)
+			rand.Read(randSource)
+			time.Sleep(time.Duration(attempts) * time.Duration(attempts) * 250 * time.Millisecond * time.Duration(randSource[0]))
```
lukechampine: `time.Duration(attempts * attempts)`

siad/renter.go
```diff
@@ -33,6 +41,22 @@ func (d *daemon) renterDownloadHandler(w http.ResponseWriter, req *http.Request)
...
+// queue.
+func (d *daemon) renterDownloadqueueHandler(w http.ResponseWriter, req *http.Request) {
+	downloads := d.renter.DownloadQueue()
+	downloadSet := make([]DownloadInfo, 0, len(downloads))
+	for _, dl := range downloads {
+		downloadSet = append(downloadSet, DownloadInfo{
+			Completed:   dl.Completed(),
+			Destination: dl.Destination(),
+			Nickname:    dl.Nickname(),
+		})
```
lukechampine: I'm not really seeing the value of an interface if all we do is immediately call all of its getters...
It'd be another thing if we kept the `downloadSet` around and called its getters multiple times.

modules/renter/download.go
DavidVorick: doesn't it? It satisfies a modules interface. Can you return unexported types as an interface?

siad/renter.go
DavidVorick: A few reasons. First, that's how it works for file info. Second, I like having interfaces as the modules package objects instead of having structs. Third, it gives the module flexibility to change the getters and not adhere to the struct.

For example, destination could go to a download folder, and the path seen by the renter could be a relative path. Then, when Destination() is called, it can append the relative path to the download path. This gives you flexibility to change the download folder without needing to modify all of your active downloads.

It's not a perfect example. Other more likely examples include adding a field that gives some indication of progress for the download.

modules/renter/download.go
lukechampine: yes. It's actually a pretty common pattern; for example, [Sha256.New](http://golang.org/src/crypto/sha256/sha256.go?s=1643:1663#L72) returns an unexported type which satisfies `hash.Hash` In fact, all of the module types (renter, host, etc.) could be unexported if their `New` function returned an interface instead of the concrete type.
...that might be a good idea, actually.

DavidVorick: okay. I think I'll put that on the list of 'to do eventually' things.

One thing though about not exporting them is that then external functions can't take advantage of any extra features they might offer, which could be useful for testing.

I guess 'extra features' kinda defeats the purpose of an interface.

modules/renter/download.go
DavidVorick: not right now, but you might have a theoretic future where renter's decide to adjust that number based on a myriad of factors.

siad/renter.go
lukechampine: Like I said, I think they'll be more useful once we want to do more sophisticated things with them. Right now they're just needlessly redundant.
I guess there's no harm in implementing them now though.

PR #396, opened 2015-03-14
Orphan synchronization

lukechampine:
The gateway will now synchronize to peers who send it orphan blocks.
This required changing the function signatures of `Synchronize` to take a peer.

I feel that the current relayblock/transaction design is pretty ugly. It would be much cleaner if the State had a Gateway and just announced blocks as they were accepted. It's not really possible to accomplish this with subscription because there's no way to synchronize to a peer who sent an orphan block.

### DIFF
--- a/siad/daemon.go
+++ b/siad/daemon.go
@@ -118,6 +118,7 @@ func (d *daemon) acceptBlock(conn modules.NetConn) error {
 	err = d.state.AcceptBlock(b)
 	if err == consensus.ErrOrphan {
 		go d.gateway.Synchronize(conn.Addr())
+		return err
 	} else if err != nil {
 		return err
 	}
--- a/siad/daemon.go
+++ b/siad/daemon.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"errors"
 	"os"
 	"path/filepath"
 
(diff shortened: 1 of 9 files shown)


DavidVorick:
We'll have to keep thinking about it. I hope there's a better way.

siad/daemon.go
```diff
@@ -107,13 +108,36 @@ func createSubdirs(rootDir string) error {
...
 	var b consensus.Block
 	err := conn.ReadObject(&b, consensus.BlockSizeLimit)
 	if err != nil {
 		return err
 	}
-	return d.gateway.RelayBlock(b)
+
+	err = d.state.AcceptBlock(b)
+	if err == consensus.ErrOrphan {
+		go d.gateway.Synchronize(conn.Addr())
```
DavidVorick: If calling Synchronize always returns an orphan (because that peer is being malicious), will this cause problems? Or will the sillyness stop after 1 call?

siad/daemon.go
```diff
@@ -107,13 +108,36 @@ func createSubdirs(rootDir string) error {
...
+	height, exists := d.state.HeightOfBlock(b.ID())
+	if !exists {
+		if consensus.DEBUG {
+			panic("could not get the height of a block that did not return an error when being accepted into the state")
+		}
+		return errors.New("state malfunction")
+	}
+	currentPathBlock, exists := d.state.BlockAtHeight(height)
+	if !exists || b.ID() != currentPathBlock.ID() {
+		return errors.New("block added, but it does not extend the state height")
```
DavidVorick: This looks like a race condition, because multiple threads could call AcceptBlock by the time you get to here.

Maybe we should have the state return ErrNotHeaviestChain or something. We use locks a bunch and I think it's causing a lot of problems; poor design choice I guess.

DavidVorick:
Merging this early b/c I need the changes.

PR #397, opened 2015-03-14
Release prep

DavidVorick:
This makes some small tweaks that basically prepare everything to work in production.

I don't think the host is correctly submitting storage proofs, although I was running on an unrealistically fast mini-network, and that could have introduced problems which might not exist in real time.

With this PR, I feel moderately confident that things will work reasonably well on a full scale network with 5-10 node, except that I don't think hosts will ever actually make money lol. But that's okay because the files are uploaded for 2 weeks and the host doesn't get paid until the end of those 2 weeks. We might be able to push an update out in time to fix that problem :P

### DIFF
--- a/modules/transactionpool/accept.go
+++ b/modules/transactionpool/accept.go
@@ -1,8 +1,14 @@
 package transactionpool
 
 import (
+	"errors"
+
 	"github.com/NebulousLabs/Sia/consensus"
 	"github.com/NebulousLabs/Sia/crypto"
+)
+
+var (
+	ErrDuplicate = errors.New("transaction is a duplicate")
 )
 
 // applySiacoinInputs adds every siacoin input to the transaction pool by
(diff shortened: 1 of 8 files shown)


modules/transactionpool/accept.go
```diff
@@ -192,21 +198,35 @@ func (tp *TransactionPool) addTransactionToPool(t consensus.Transaction, directi
...
+	id := tp.mu.Lock()
+	txnHash := crypto.HashObject(t)
+	_, exists := tp.seenTransactions[txnHash]
+	if exists {
+		tp.mu.Unlock(id)
+		return ErrDuplicate
+	}
+	if len(tp.seenTransactions) > 1200 {
+		tp.seenTransactions = make(map[crypto.Hash]struct{})
+	}
```
lukechampine: hmm. Seems less than ideal to just wipe the map entirely. Limiting the set to the 1200 most recent txns would be better, but more of a pain to implement. I guess we'll leave it for now.

PR #402, opened 2015-03-14
Last minute API fixes

lukechampine:
API.md had some errors. siac was implementing a couple calls wrong, and downloadqueue wasn't implemented at all. There was also a `Synchronize` error where synchronizing to a peer with the same height would cause an error.


modules/gateway/synchronize.go
```diff
@@ -83,14 +82,21 @@ func (g *Gateway) sendBlocks(conn modules.NetConn) (err error) {
...
 	}
-	if !found {
-		// The genesis block should be included in knownBlocks - if no matching
-		// blocks are found, the caller is probably on a different blockchain
-		// altogether.
-		return errors.New("no matching block found")
+	// If we didn't find any matching blocks, or if we're already
+	// synchronized, don't send any blocks. The genesis block should be
+	// included in knownBlocks, so if no matching blocks are found, the caller
+	// is probably on a different blockchain altogether.
```
DavidVorick: hmm. Shouldn't knownBlocks[0] be a match if the two are already synchronized?

siad/API.md
```diff
@@ -170,28 +173,31 @@ windowSize   int
...
+`minFilesize is the minimum allowed file size.
 
-maxFilesize is the maximum allowed file size.
+`maxFilesize is the maximum allowed file size.
 
-minDuration is the minimum amount of time a contract is allowed to last.
+`minDuration is the minimum amount of time a contract is allowed to last.
 
-maxDuration is the maximum amount of time a contract is allowed to last.
+`maxDuration is the maximum amount of time a contract is allowed to last.
```
DavidVorick: you didn't close the hooks on some of these

siad/API.md
```diff
@@ -388,7 +411,13 @@ Parameters: none
...
+	FullBalance  int
+	NumAddresses int
 }
 ```
-Balance is the spendable balance of the wallet.
+`Balance` is the spendable balance of the wallet.
+
+`FullBalance` is the balance of the wallet, including unconfirmed coins.
+
+`NumAddresses is the number of addresses controlled by the wallet.
```
DavidVorick: didn't close the hook here either

DavidVorick:
done reviewing

modules/gateway/synchronize.go
lukechampine: yes, so `found` will be true, but `start` will be greater than `g.state.Height()` (actually it should be exactly `g.state.Height() + 1`, but no harm in being safe).

PR #403, opened 2015-03-15
Fix contract bug

lukechampine:
Found the bug!

Seems like it was actually caused by too *many* pieces, not too few. The problem was that each piece was being uploaded in parallel, but the actual data -- the `io.ReadSeeker` -- was shared between them. So their seeks, reads, merkle roots, etc. were being interleaved.

I changed the `UploadParams` back to just using a filename instead of a generic interface. I don't see a real need to support more than that. Maybe one day we will, but for now this is a lot easier. Each upload opens a separate file handle, so the reads don't conflict. We could potentially optimize this by writing to multiple connections at once, instead of having to read the entire file n times, but it'd be wasted effort since the pieces are eventually going to be erasure-coded anyway.

I also added a naive connection timeout; any time there's a read or write, the timeout is reset to 10 seconds. This should be more than enough time, even for very slow connections. (In case you're worried about large files: they call `Read` many times on a small buffer, so they shouldn't be in danger of timing out.)

### DIFF
--- a/modules/gateway/gateway_test.go
+++ b/modules/gateway/gateway_test.go
@@ -74,3 +74,28 @@ func TestRPC(t *testing.T) {
 		t.Fatal("Foo gave wrong response:", foo)
 	}
 }
+
+// TestTimeout tests that connections time out properly.
+func TestTimeout(t *testing.T) {
+	if testing.Short() {
+		t.Skip()
+	}
+
+	g := newTestGateway(t)
+	defer g.Close()
+
+	// create unresponsive peer
+	badpeer := newTestGateway(t)
+	// overwrite badpeer's Ping RPC with an incorrect one
+	// since g is expecting 4 bytes, it will time out.
+	badpeer.RegisterRPC("Ping", func(conn modules.NetConn) error {
+		// write a length prefix, but no actual data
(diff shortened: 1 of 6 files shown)


DavidVorick:
They were all using the same ReadSeeker? whoops

Thanks for finding this.

modules/gateway/gateway_test.go
```diff
@@ -74,3 +74,28 @@ func TestRPC(t *testing.T) {
...
+		// write a length prefix, but no actual data
+		conn.Write([]byte{1, 0, 0, 0, 0, 0, 0, 0})
+		select {}
+	})
+
+	err := g.RPC(badpeer.Address(), "Ping", readerRPC([1]byte{}, 1))
+	if err != ErrTimeout {
+		t.Fatalf("Got wrong error: expected %v, got %v", ErrTimeout, err)
+	}
+}
```
DavidVorick: Does this also test that the connection is properly reset to + 10 seconds after each read or write?

lukechampine: nope. I'll add more sophisticated tests once the connection monitoring is more sophisticated.

PR #404, opened 2015-03-16
Update interface

DavidVorick:
I'm making this PR now even though the build is failing. I wanted to explain what all the changes are before I continue improving the transaction pool.

The original intent of this PR was to make the transaction pool follow the update interface model of interacting with the state. The idea behind the update interface model is that the modules will spend significantly less time making locking calls to the state. Fewer locking calls also means less room for programmer error with regards to race conditions and deadlocks, of which there currently seem to be a few of each.

This expanded a bit as I decided to try and do the whole transaction pool package "correctly". This meant changing the way that the directories and the ports were chosen. I discovered that if you use ":0", the operating system will find you an open port and give you that one. It didn't work for the gateway tests, but it seems to have worked for all of the other tests.

This also meant changing the way file directories were chosen. Files are now saved to /tmp/SiaTesting or something like that. Each consensus set gets it's own directory. Each module, upon start, is responsible for creating its own directory using MkdirAll. I believe this is in conflict with how it currently works, where the daemon is responsible for creating the directories. I think the daemon should be responsible for choosing the directories, but the modules responsible for creating and populating them.

In some of the modules files (I think all of them with persistence), there's a variable like `WalletDir`, which just indicates the standard name of the sub-directory that the module uses (this is in the modules package, not any of the subpackages).

Because the transaction pool synchronization mechanism changed, some of the tests broke. Luckily, not many of them broke and a few of them had simple fixes. It'll probably take me a good amount of time to get to the ones that did break, because that's probably going to mean completely updating how that module operates.

I believe that the new method for synchronization is superior. We shouldn't need spin locks while testing any more, or sleeping. We should just be able to wait on a channel until all of the expected updates have been posted.

The transaction pool is under active construction. The structural changes should be more or less complete but I want to review the code and try to reduce the amount of locking that happens, as well as increase the test coverage.


lukechampine:
>if you use ":0", the operating system will find you an open port and give you that one
whoa, that's super useful. I bet that's how programs like torrent clients pick "random" port numbers.

Makefile
```diff
@@ -21,7 +21,8 @@ install: fmt REBUILD
 clean:
 	rm -rf hostdir release whitepaper.aux whitepaper.log whitepaper.pdf         \
 		*.wallet* *_test testdir* */*_test hostdir* siad/walletDir*             \
-		siad/hostDir* info.log modules/*/*.dat modules/*/*.backup
+		siad/hostDir* info.log modules/*/*.dat modules/*/*.backup               \
+		/tmp/SiaTesting
```
lukechampine: could we reduce this to just `/tmp/SiaTesting`?

consensus/state.go
```diff
@@ -53,6 +53,8 @@ type State struct {
 	// listend on by modules. An empty struct is thrown down the channel any
 	// time that the consensus set of the state changes. subscriptionCounter
 	// only ever increments, and prevents collisions in the map.
+	revertUpdates [][]*blockNode
+	applyUpdates  [][]*blockNode
```
lukechampine: outdated comment?

consensus/subscribers.go
```diff
@@ -0,0 +1,86 @@
+package consensus
+
+// A Subscriber is an object that receives updates to the consensus set every
+// time there is a change in consensus.
+type Subscriber interface {
+	// ReceiveConsensusUpdate notifies subscribers that there has been a change in the
+	// consensus set, and provides an enumeration of the changes.
+	ReceiveConsensusUpdate(revertedBlocks []Block, appliedBlocks []Block)
+}
```
lukechampine: docstring is misleading; the method isn't the one doing the notification, it's the one processing the changes. `ProcessConsensusUpdate` might be a better name.

Also, doesn't `BlockSince` just return the `BlockID`s? Seems like that could be more efficient, esp. if the Subscriber doesn't need the contents of every block.

consensus/subscribers.go
```diff
@@ -0,0 +1,86 @@
...
+// with consensus; updates cannot make blocking calls from any thread that is
+// holding a lock on consensus. The result is a construction where all updates
+// are added to a list of updates in the consensus set while the consensus set
+// is locked. Then, a separate thread for each subscriber will be notified (via
+// the update chan) that there are new updates. The thread will lock the
+// consensus set for long enough to get the updates, and then will unlock the
+// consensus set while it makes a blocking call to the subscriber. If the
+// subscriber deadlocks or has problems, the thread will stall indefinitely,
+// but the rest of consensus will not be disrupted.
+func (s *State) threadedSendUpdates(update chan struct{}, subscriber Subscriber) {
```
lukechampine: what if instead of calling `go threadedSendUpdates`, you just called `go subscriber.ReceiveConsensusUpdate`? would that work?

modules/miner.go
```diff
@@ -0,0 +1,18 @@
+package modules
+
+import (
+	"github.com/NebulousLabs/Sia/consensus"
+)
+
+// The Miner interface provides access to mining features.
+type Miner interface {
+	// FindBlock will have the miner make 1 attempt to find a solved block. It
+	// will give up after a few seconds, returning a block, a bool indiciating
```
lukechampine: something about "indic" really makes David want to type another i...

modules/miner/miner_test.go
```diff
@@ -12,9 +15,13 @@ import (
 // TestMiner creates a miner, mines a few blocks, and checks that the wallet
 // balance is updating as the blocks get mined.
 func TestMiner(t *testing.T) {
+	directory := "TestMiner"
+
 	// Create the miner and all of it's dependencies.
 	s := consensus.CreateGenesisState()
-	g, err := gateway.New(":8900", s, "")
+
+	gDir := filepath.Join(tester.TempDir(directory), modules.GatewayDir)
```
lukechampine: this pattern seems common enough that you might want to add a function in `tester` that does the `Join` for you. Then you wouldn't have to import `filepath` all over the place.

modules/tester/tester.go
```diff
@@ -0,0 +1,18 @@
...
+const (
+	SiaTestingDir = "SiaTesting"
+)
+
+// TempDir returns the tempified version of the input directory, after creating
+// the tempified directory.
+func TempDir(dir string) string {
+	temp := os.TempDir()
+	tempified := filepath.Join(temp, SiaTestingDir, dir)
+	return tempified
```
lukechampine: okay come on, you know this could be one line :P

Makefile
DavidVorick: We can once everything has been moved over. The daemon is still creating a bunch of files though. We'll be able to clean out the .gitignore as well.

consensus/subscribers.go
DavidVorick: blocks are just slices, which take less memory than block ids.

lol

modules/miner.go
DavidVorick: inidiciating

consensus/subscribers.go
DavidVorick: no, because they need to get the updates in the correct order. If you have two threads waiting to call AcceptBlock, it's possible that the second AcceptBlock will send the ConsensusUpdate before the first AcceptBlock sends it. That's why you need a separate function listening on a channel to push everything out in the right order.

modules/miner/miner_test.go
DavidVorick: It's probably worth making TempDir variadic.

modules/miner.go
DavidVorick: FindBlock: finds a block on top of the current consensus set, and submits that block to the consensus package.

SolveBlock: given a block, which doesn't need to be valid, and a target, which doesn't need to be the current target of the consensus set, solve the block. Don't send the block to the consensus package.

SolveBlock is useful for testing, and not much more. FindBlock is ultimately also useful for testing and not much more, but it's a simpler function call and it does more things.

modules/tester/tester.go
DavidVorick: This should probably be one line lol.

My guess is that os.TempDir also is just a global variable. I don't think there's a huge reason to not make the function call every time.

consensus/subscribers.go
lukechampine: hmm. Didn't we used to use a channel to serialize `AcceptBlock`? i.e. to enforce that only one thread was calling `AcceptBlock`? Maybe returning to that model would simply this.

modules/tester/tester.go
lukechampine: yeah, just define `var siaTempDir = filepath.Join(os.TempDir(), SiaTestingDir)` and then `TempDir(dir)` just becomes `return filepath.Join(siaTempDir, dir)`

consensus/subscribers.go
DavidVorick: That's not the problem. The issue is that either you need AcceptBlock to block until all of the modules have received the update, or you need to have some other process that keeps things organized. Because we don't want to block while modules are receiving updates, I chose to have some other process (one gothread per module) keep things organized.

DavidVorick:
I will consider this PR finished.

modules/gateway/gateway_test.go
```diff
@@ -6,21 +6,27 @@ import (
...
-var port = 9001
+var (
+	rpcPort = 10000
+)
 
-func newTestGateway(t *testing.T) *Gateway {
-	g, err := New(":"+strconv.Itoa(port), consensus.CreateGenesisState(), "")
+// newTestingGateway returns a gateway read to use in a testing environment.
+func newTestingGateway(directory string, t *testing.T) *Gateway {
+	gDir := tester.TempDir(directory, modules.GatewayDir)
```
lukechampine: I suppose you could take it a step further and use `runtime` to automatically get the name of the Test* function and use that for the folder name. Then you wouldn't need to pass it in manually. Probably overkill though.

modules/tester/tester.go
```diff
@@ -0,0 +1,18 @@
...
+const (
+	SiaTestingDir = "SiaTesting"
+)
+
+// TempDir takes a set of directory names and joins them to the sia temp
+// folder.
+func TempDir(dirs ...string) string {
+	tmp := append([]string{os.TempDir()}, SiaTestingDir)
+	full := append(tmp, dirs...)
+	return filepath.Join(full...) // filepath.Join(tmp, testing, dirs...) returns 'too many arguments' error.
```
lukechampine: yeah, isn't that a pain in the ass?

modules/gateway/gateway_test.go
DavidVorick: also gives you a little less control. Some functions need multiple. (Like the daemon testing peer stuff uses multiple)

modules/tester/tester.go
DavidVorick: I can at least do that

PR #405, opened 2015-03-16
Download progress

lukechampine:
The `DownloadInfo` interface now exposes `Filesize() uint64` and `Received() uint64` methods. The value returned by `Received()` changes in real-time as the file downloads. These are included in the `/renter/downloadqueue` response, so the client can implement a "download progress bar" by calculating the received/total ratio.
This functionality has been added to siac; to watch a file download in real-time, start a download with `siac renter download [nickname] [destination]` and then run `watch -n0.1 siac renter queue`.

### DIFF
--- a/modules/renter/download.go
+++ b/modules/renter/download.go
@@ -11,114 +11,161 @@ import (
 	"github.com/NebulousLabs/Sia/modules"
 )
 
-// A Download is a download that has been queued by the renter.
-type Download struct {
-	completed   bool
-	destination string
-	nickname    string
-
-	renter *Renter
-}
-
 var (
 	downloadAttempts = 5
 )
 
-// downloadPiece attempts to retrieve a file from a host.
-func (r *Renter) downloadPiece(piece FilePiece, path string) error {
-	return r.gateway.RPC(piece.HostIP, "RetrieveFile", func(conn modules.NetConn) (err error) {
-		// Send the id of the contract for the file piece we're requesting. The
-		// response will be the file piece contents.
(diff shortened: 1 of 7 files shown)


modules/renter/download.go
```diff
@@ -11,114 +11,169 @@ import (
...
+	complete    bool
+	filesize    uint64
+	received    uint64
 	destination string
 	nickname    string
 
-	renter *Renter
+	pieces  []FilePiece
+	file    *os.File
+	gateway modules.Gateway
```
DavidVorick: should this have a mutex?

DavidVorick: standard procedure here I think would be to give it a renter and lock the renter any time you change the value of the download.

DavidVorick:
A quick glance makes me think this is all okay, except I'm pretty sure there are race conditions on the Download object.

modules/renter/download.go
lukechampine: a normal mutex should be fine, no need to lock the renter.

actually though, I was thinking this would be a good place to use `sync/atomic`.

DavidVorick: no need to get fancy right now.

The reason I would want to lock the renter is that I'm not sure where else the renter is looking at the download object. I'm not really worried about it too much though right now.

lukechampine: I actually quite prefer the atomic implementation. The need for a mutex is eliminated because only one field is ever updated (`received`), and the update happens atomically.

DavidVorick: ok

PR #415, opened 2015-03-20
Tpool polishing

DavidVorick:
There were a handful of things about the tpool which were incorrect. I have fixed them. I have also combined much of the code that was redundant between the consensus package and the tpool package. A minor amount of testing was added, but most of this code is not tested. It'll make more sense to test it once the host and wallet packages have been built out a bit more.

A lot of the locking complexity has been minimized. The lock is still pretty complex, but it's less complex.

### DIFF
--- a/consensus/validtransaction.go
+++ b/consensus/validtransaction.go
@@ -88,6 +88,7 @@ func (s *State) validUnlockConditions(uc UnlockConditions, uh UnlockHash) (err e
 // value of the inputs and checking that the inputs are legal.
 func (s *State) validSiacoins(t Transaction) (err error) {
 	var inputSum Currency
+	spentIDs := make(map[SiacoinOutputID]struct{})
 	for _, sci := range t.SiacoinInputs {
 		// Check that the input spends an existing output, and that the
 		// UnlockConditions are legal (signatures checked elsewhere).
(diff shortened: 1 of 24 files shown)


consensus/validtransaction.go
```diff
@@ -143,38 +81,198 @@ func (s *State) validFileContracts(t Transaction) (err error) {
...
+func (t Transaction) followsMinimumValues() error {
+	for _, sco := range t.SiacoinOutputs {
+		if sco.Value.Cmp(ZeroCurrency) < 1 {
+			return errors.New("empty siacoin output not allowed")
+		}
+	}
+	for _, fc := range t.FileContracts {
+		if fc.Payout.Cmp(ZeroCurrency) < 1 {
+			return errors.New("file contract must have non-zero payout")
+		}
```
lukechampine: I prefer `<= 0` but I suppose it doesn't make much difference

consensus/validtransaction.go
```diff
@@ -317,11 +433,13 @@ func (s *State) validTransaction(t Transaction) (err error) {
...
-		return
-	}
+	return
+}
 
+// ValidStorageProofs is the exported version of validStorageProofs.
+func (s *State) ValidStorageProofs(t Transaction) (err error) {
+	id := s.mu.RLock()
+	defer s.mu.RUnlock(id)
+	err = s.validStorageProofs(t)
```
lukechampine: why not just `return s.validStorageProofs(t)`?

DavidVorick: oh.

I'll switch the documentation too.

PR #418, opened 2015-03-31
Transaction Pool Polishing

DavidVorick:
Order of doing:

1. Review this PR, make comments, I'll fix stuff
2. I'll make another PR into this PR containing code that will fix the build. Idk how much coding that will take
3. Review that PR, make comments, I'll fix stuff
4. Merge that PR into this PR
5. Merge this PR into master.

it's a little complicated but I didn't want this to turn into a monster PR. Despite the branch name of `wallet-updates`, the majority of the work here is on the transaction pool. I feel like the transaction pool is very solid/beautiful at this point. Still rather light on testing, but overall it's well documented, pretty simple in construction, and has been designed a lot more carefully than ever in the past.

The amount of times that the transaction pool grabs a lock on the state has been substantially reduced, I think it only grabs a lock in 2 places right now. Also, the test suite now follows much stricter synchronization. The miner also synchronizes much more strictly to the transaction pool.

There is now an api call to see which transactions are in the transaction pool. We probably won't use it much, and so it's just a raw JSON dump but even that should be useful to us.

A good deal of code has been removed. The diffs indicate that only 80 lines have been removed, but actually the amount of documentation has gone up substantially. Many of the `+`'s are comments.

This PR broke the wallet and siad tests. All of the transaction pool tests however are still around, and work well. My second PR (parts 3-4) will fix the testing for the wallet and siad.


Makefile
```diff
@@ -35,6 +35,9 @@ test: clean fmt REBUILD
 test-long: clean fmt REBUILD
 	go test -v -race -tags=test -timeout=180s ./...
 
+test-tpool: clean fmt REBUILD
+	go test -v -race -tags=test -timeout=8s github.com/NebulousLabs/Sia/modules/transactionpool
+
```
lukechampine: do we need a separate target for this? `go test -v -race -tags=test ./transactionpool` seems succinct enough to just keep in history.

DavidVorick: I didn't realize you could type `./transactionpool`. But I still think I'd rather have a make target for each of the modules (I added one for the wallet in the next PR). I also reorganized the makefile a bit.

It's not super clean but anyone who takes 10 or 20 minutes to figure out everything that's happening in the Makefile should be able to catch up. I find it a lot nicer to have `make test-tpool` than to need to remember what all the flags are, to set a timeout, to set the tags.

The timeout came in to play many times while I was fixing things up.

Makefile
```diff
@@ -35,6 +62,12 @@ test: clean fmt REBUILD
 test-long: clean fmt REBUILD
 	go test -v -race -tags=test -timeout=180s ./...
 
+# Testing for each package individually. Packages are added to this set as needed.
+test-tpool: clean fmt REBUILD
+	go test -v -race -tags=test -timeout=8s github.com/NebulousLabs/Sia/modules/transactionpool
+test-wallet: clean fmt REBUILD
+	go test -v -race -tags=test -timeout=8s github.com/NebulousLabs/Sia/modules/wallet
```
lukechampine: `./modules/transactionpool` is sufficient

modules/transactionpool/update.go
```diff
@@ -1,241 +1,375 @@
...
-		fc, exists := tp.referenceFileContracts[fct.ParentID]
-		if consensus.DEBUG {
-			if !exists {
-				panic("could not locate file contract")
-			}
+		scods = tmpScods
+
+		var tmpFcds []consensus.FileContractDiff
+		for i := len(fcds) - 1; i >= 0; i-- {
+			tmpFcds = append(tmpFcds, fcds[i])
```
lukechampine: why not reverse these in-place?

PR #419, opened 2015-03-31
Test fix

DavidVorick:
To the best of my knowledge, this PR brings all of the tests back online. A few of the wallet tests have been deleted, but that's because they were utterly trivial and didn't really test what they originally claimed. (This was a result of changing the testing style).

Some very minor organizational stuff happened. For example, the miner is now a proper module.

I do feel like the synchronization problems have not really been addressed at all, but maybe they were never addressed in the first place, which would explain why everything appears to work again.


modules/wallet/transactions_test.go
```diff
@@ -9,75 +9,57 @@ import (
...
 	// the unconfirmed balance of the wallet is 1.
 	if len(t.SiacoinOutputs) != 0 {
-		wt.Error("more than zero siacoin outputs created in custom transaction")
+		wt.t.Error("more than zero siacoin outputs created in custom transaction")
 	}
-	if wt.Balance(true).Cmp(consensus.NewCurrency64(1)) != 0 {
-		wt.Error(wt.Balance(true).MarshalJSON())
-		wt.Error("wallet balance not reporting at one?")
+	if wt.wallet.Balance(true).Cmp(consensus.NewCurrency64(1)) != 0 {
+		wt.t.Error(wt.wallet.Balance(true).MarshalJSON())
```
lukechampine: Does `Currency` still not have a `String` method? If it does, `MarshalJSON` isn't necessary here

PR #420, opened 2015-04-02
API package

lukechampine:
siad is now just two files. The siad testing has moved to api. Functionality should be unaffected.

I consider the api package to be incomplete as of now. I'd like to make it more useful to siac, but I haven't yet hit upon the best way of accomplishing that.


api/api.go
```diff
@@ -0,0 +1,112 @@
...
+func (srv *Server) initAPI(addr string) {
+	mux := http.NewServeMux()
+
+	// Consensus API Calls
+	handleHTTPRequest(mux, "/consensus/status", srv.consensusStatusHandler)
+
+	// Server API Calls
+	handleHTTPRequest(mux, "/daemon/stop", srv.daemonStopHandler)
+	handleHTTPRequest(mux, "/daemon/update/apply", srv.daemonUpdateApplyHandler)
+	handleHTTPRequest(mux, "/daemon/update/check", srv.daemonUpdateCheckHandler)
```
DavidVorick: these should all be renamed to /server, and the function names should be srv.serverXXXHandler

api/server.go
```diff
@@ -0,0 +1,106 @@
...
+	return srv
+}
+
+func (srv *Server) updateWait() {
+	<-srv.minerUpdates
+	<-srv.walletUpdates
+}
+
+// TODO: move this to the state module?
+func (srv *Server) acceptBlock(conn modules.NetConn) error {
```
DavidVorick: do these functions even still get used?

api/server.go
```diff
@@ -0,0 +1,106 @@
...
+		return errors.New("block added, but it does not extend the state height")
+	}
+
+	srv.gateway.RelayBlock(b)
+	return nil
+}
+
+// TODO: move this to the tpool module?
+func (srv *Server) acceptTransaction(conn modules.NetConn) error {
+	var t consensus.Transaction
```
DavidVorick: same here

modules/host/host.go
```diff
@@ -71,11 +71,11 @@ func New(state *consensus.State, tpool modules.TransactionPool, wallet modules.W
...
-			TotalStorage: 2e9,                                    // 2 GB
-			MaxFilesize:  300e6,                                  // 300 MB
-			MaxDuration:  5e3,                                    // Just over a month.
-			WindowSize:   288,                                    // 48 hours.
-			Price:        consensus.NewCurrency64(1000000000000), // 10^12
+			TotalStorage: 2e9,                           // 2 GB
+			MaxFilesize:  300e6,                         // 300 MB
+			MaxDuration:  5e3,                           // Just over a month.
+			WindowSize:   288,                           // 48 hours.
+			Price:        consensus.NewCurrency64(1e12), // 10^12
```
DavidVorick: actually, before it is forgotten, can we change this to 100e6?

DavidVorick: or maybe 1e9, both should be alright.

siad/daemon.go
```diff
@@ -28,137 +25,49 @@ type DaemonConfig struct {
...
-	if err == consensus.ErrOrphan {
-		go d.gateway.Synchronize(conn.Addr())
-		return err
-	} else if err != nil {
-		return err
+	// bootstrap to the network
+	// TODO: probably a better way of doing this.
+	if !config.Siacore.NoBootstrap {
+		go gateway.Bootstrap(modules.BootstrapPeers[0])
 	}
```
DavidVorick: should the bootstrapping be handled by the gateway? I feel like it's not really the daemon's job, it's more of the gateway's job.

The daemon is really just a tiny shell that strings everything together.

siad/main.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/mitchellh/go-homedir"
 	"github.com/spf13/cobra"
 
-	"github.com/NebulousLabs/Sia/modules"
+	"github.com/NebulousLabs/Sia/api"
```
DavidVorick: I thought the api was going to be in the modules package?

DavidVorick:
Some smaller comments.

I also thought the api was going to be in the modules folder.

api/server.go
lukechampine: yes, they are registered to the gateway above.

siad/daemon.go
lukechampine: yeah. What should probably happen is that gateway.New should take a bootstrap flag. We also need a way to override the bootstrap peer. (For consistency, probably both a cmdline flag and a config value.)

lukechampine:
I could see arguments for both sides. My first thought was, "well, none of the other modules import the API server." But that's also true of the host and hostdb. So it isn't 100% clear.

It seems like the modules folder is eating all the other packages, so I think down the line we may wind up moving all the modules stuff up a directory.

DavidVorick:
Even if we move all the modules up a layer, we still need an explicit modules package.

lukechampine:
Are you sure? Why is it required?

DavidVorick:
The modules need to import eachother's interfaces without actually importing eachother. This prevents import cycles when doing testing. Otherwise the tpool would be importing a miner (which has a wallet) to help with testing, but the miner imports the tpool.

Using the modules package + interfaces, we can have testing scenarios where the tpool uses a miner but avoids import loops.

lukechampine:
Sure, so all you do is move the interface definitions to a top-level file, or maybe to the types package.

PR #423, opened 2015-04-03
Consensus module

DavidVorick:
There's some organizational stuff left but I'l leave it for another PR.

I gave the currency type a decent amount of attention, and I have high confidence in the current code for it.


types/currency.go
```diff
@@ -1,24 +1,44 @@
...
+
+	ErrNegativeCurrency = errors.New("negative currency not allowed")
+)
+
+// currency.go defines the internal currency object. One design goal of the
+// currency type is immutability: the currency type should be safe to pass
+// directly to other objects and packages. The currency object should never
+// have a negative value. The currency should never overflow. There is a
+// maximum size value that can be encoded (around 10^600), however exceeding
+// this value will not result in overflow.
```
lukechampine: with the marshalling changes, the maximum value is now approx. 10^10^20

types/currency.go
```diff
@@ -28,34 +48,27 @@ func NewCurrency64(x uint64) (c Currency) {
...
-	y.i.Add(&c.i, &x.i)
-	return
-}
-
-// Sub returns a new Currency value y = c - x.
-func (c Currency) Sub(x Currency) (y Currency) {
-	y.i.Sub(&c.i, &x.i)
+// Div returns a new Currency value y = c / x.
+func (c Currency) Div(x Currency) (y Currency) {
+	y.i.Div(&c.i, &x.i)
```
lukechampine: alphabetizing? don't really see the gain here but w/e.

types/currency.go
```diff
@@ -1,24 +1,44 @@
...
+// NewCurrency creates a Currency value from a big.Int. Undefined behavior
+// occurs if a negative input is used.
 func NewCurrency(b *big.Int) (c Currency) {
 	c.i = *b
+	if c.Cmp(ZeroCurrency) < 0 {
+		c = Currency{}
+		if DEBUG {
+			panic("cannot have a negative currency")
+		}
+	}
```
lukechampine: you can use the same style you used for MulFloat:
```go
if b.Sign() < 0 {
    if DEBUG {
        panic("cannot have a negative currency")
    }
} else {
    c.i = *b
}
return
```

types/currency.go
```diff
@@ -89,54 +111,66 @@ func (c Currency) Sqrt() (y Currency) {
...
-		b...,
-	)
+// UnmarshalJSON implements the json.Unmarshaler interface. An error is
+// returned if a negative number is provided.
+func (c *Currency) UnmarshalJSON(b []byte) error {
+	err := c.i.UnmarshalJSON(b)
+	if err != nil {
+		return err
+	}
+	if c.Cmp(ZeroCurrency) < 0 {
```
lukechampine: I would use `c.i.Sign() < 0` for clarity

types/currency.go
```diff
@@ -89,54 +111,66 @@ func (c Currency) Sqrt() (y Currency) {
...
 
 // Scan implements the fmt.Scanner interface, allowing Currency values to be
 // scanned from text.
 func (c *Currency) Scan(s fmt.ScanState, ch rune) error {
-	return c.i.Scan(s, ch)
+	err := c.i.Scan(s, ch)
+	if err != nil {
+		return err
+	}
+	if c.Cmp(ZeroCurrency) < 0 {
```
lukechampine: ditto here

types/currency_test.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+}
+
+// TestCurrencyString probes the String function of the currency type.
+func TestCurrencyString(t *testing.T) {
+	b := big.NewInt(7135)
+	c := NewCurrency64(7135)
+	if b.String() != c.String() {
+		t.Error("string function not behaving as expected")
+	}
```
lukechampine: seems kinda silly to test the pass-through functions, but I guess the coverage % is nice.
here you can just use `NewCurrency64(7135).String() != "7135"`

types/currency.go
DavidVorick: It's how the go standard library does it. It makes finding functions easier when looking at a file.

types/currency_test.go
```diff
@@ -0,0 +1,205 @@
...
+// TestNegativeCurrencySub checks that negative numbers are prevented when
+// using subtraction on the currency type.
+func TestNegativeCurrencySub(t *testing.T) {
+	// In debug mode, attempting to get a negative currency results in a panic.
+	defer func() {
+		r := recover()
+		if r == nil {
+			t.Error("no panic occured when trying to create a negative currency")
+		}
+	}()
```
lukechampine: you can abstract away the recovery by using a closure:
```go
func catchErr(fn func()) (err error) {
    defer func() {
        if r := recover(); r != nil {
            err = fmt.Errorf("%v", r)
        }
    }()
    fn()
}

func TestNegativeCurrencySub(t *testing.T) {
    err := catchErr(func() {
        NewCurrency64(1).Sub(NewCurrency64(2))
    })
    if err == nil {
        t.Fatal("no panic occurred when using improper arguments to Sub")
    }
}
```

types/currency.go
```diff
@@ -65,23 +78,32 @@ func (c Currency) Mul(x Currency) (y Currency) {
...
 
-// Div returns a new Currency value y = c / x.
-func (c Currency) Div(x Currency) (y Currency) {
-	y.i.Div(&c.i, &x.i)
+// MulFloat returns a new Currency value y = c * x, where x is a float64.
+// Behavior is undefined when x is negative.
+func (c Currency) MulFloat(x float64) (y Currency) {
+	if x < 0 {
+		if DEBUG {
+			panic("cannot multiple currency by a negative number")
```
lukechampine: multiply

types/currency_test.go
DavidVorick: I actually caught an error in one of the pass-through functions, so it's not completely silly.

types/currency.go
DavidVorick: There is no Sign() function anymore since it can't ever be negative.

I guess I should add it back though. Cmp(ZeroCurrency) is a pretty awful construction, especially when you have to import the types package just to do that.

types/currency.go
lukechampine: Sometimes they do, sometimes they don't. The math.Big arithmetic is grouped together, and the formatting functions are grouped together as well: http://golang.org/src/math/big/int.go?s=2391:2424#L100

I can see the advantage for alphabetizing large files, but in this case it's very obvious that Add, Sub, Mul, and Div are all the same "type" of function and belong together. Besides, you didn't separate Marshal and Unmarshal :P

types/currency.go
lukechampine: Not `Currency.Sign()`, `Int.Sign()`

DavidVorick: ohhhhhhh. Got it.

types/currency.go
DavidVorick: life is hard.

I left Marshal and Unmarshal together because it seemed really natural. New is at the top.

DavidVorick:
Sorry for being disorganized, but I fixed all the errors in the 'types-cleanup' branch. The reason is that since I split consensus into two packages, all the imports and stuff had changed and I wanted to avoid merge conflicts as much as possible.

lukechampine:
You shouldn't be scared of a simple rebase like this. It's worth the effort to keep things organized.

DavidVorick:
\>.< just tried but somehow lost the commit entirely.

=/

lukechampine:
lol. I cherry-picked it into this branch locally, I can push it if you want

DavidVorick:
oh thanks. That would be helpful

Issue #421, opened 2015-04-02
Gateway failure modes


DavidVorick:
The gateway is occasionally kicking all peers, which desynchronizes the network the isolates peers. The gateway needs some sort of failure mode that it regularly checks (perhaps every time a peer is potentially removed? Or perhaps a gothread that just runs a check every 5 minutes?) if there are peers available.

If there are less than 3 peers, it re-adds all the peers that it knows, even the abusive ones, giving priority to the hardcoded peers and the user-input peers.

The network is still a pronounced weakpoint of the software.

lukechampine:
The gateway no longer kicks peers; if you lose a peer, it's because your connection to them was lost. When we bring back the strike/banning code, we'll be careful not to remove peers when we fall below the "well-connected" threshold.

PR #425, opened 2015-04-03
Peering

lukechampine:
This PR address #421. I'm not certain was the best approach is. Here, the gateway will simply stop automatically removing peers when it drops below 3 (they can still be manually removed, though perhaps this should cause a warning). Another way would be to not throw any peers away, but just keep them in a "bad peers" map. Then if we fall below 3 peers, we add the best of the bad peers. Finally, if we fall below 3 peers, we could query the remaining ones for their peers. But this won't help if those peers aren't well-connected themselves.

btw, we should be using SkipNow instead of Skip. Skip produces a log entry, while SkipNow does not.

### DIFF
--- a/modules/gateway/peer_test.go
+++ b/modules/gateway/peer_test.go
@@ -106,7 +106,7 @@ func TestBadPeer(t *testing.T) {
 // TestBootstrap tests the bootstrapping process, including synchronization.
 func TestBootstrap(t *testing.T) {
 	if testing.Short() {
-		t.Skip()
+		t.SkipNow()
 	}
 
 	// create bootstrap peer
--- a/modules/gateway/peer_test.go
+++ b/modules/gateway/peer_test.go
@@ -1,6 +1,7 @@
 package gateway
 
 import (
+	"strconv"
 	"testing"
 
 	"github.com/NebulousLabs/Sia/consensus"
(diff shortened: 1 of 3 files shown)


DavidVorick:
I like seeing the message that a test was skipped when running '-v'

modules/gateway/peer_test.go
```diff
@@ -84,16 +85,28 @@ func TestBadPeer(t *testing.T) {
 		g.Ping(badpeer.Address())
 	}
 
-	// badpeer should no longer be in our peer list
-	if len(g.peers) != 0 {
-		t.Fatal("gateway did not remove bad peer:", g.Info().Peers)
+	// since we are poorly-connected, badpeer should still be in our peer list
+	if len(g.peers) != 1 {
```
DavidVorick: can you change '1' to a configurable variable?

DavidVorick: oh sorry looking at the wrong thing

modules/gateway/peers.go
```diff
@@ -63,6 +75,7 @@ func (g *Gateway) AddPeer(peer modules.NetAddress) error {
 }
 
 // RemovePeer removes a peer from the Gateway's peer list.
+// TODO: warn if less than minPeers?
```
DavidVorick: probably

lukechampine:
It still shows up as being skipped:
```
=== RUN TestTableTennis
--- PASS: TestTableTennis (0.00 seconds)
=== RUN TestRPC
--- PASS: TestRPC (0.00 seconds)
=== RUN TestTimeout
--- SKIP: TestTimeout (0.00 seconds)
=== RUN TestPeerSharing
--- PASS: TestPeerSharing (0.00 seconds)
=== RUN TestBadPeer
--- PASS: TestBadPeer (0.00 seconds)
=== RUN TestBootstrap
--- SKIP: TestBootstrap (0.00 seconds)
PASS
ok  	github.com/NebulousLabs/Sia/modules/gateway	0.005s
```

vs Skip:
```
=== RUN TestTableTennis
--- PASS: TestTableTennis (0.00 seconds)
=== RUN TestRPC
--- PASS: TestRPC (0.00 seconds)
=== RUN TestTimeout
--- SKIP: TestTimeout (0.00 seconds)
	gateway_test.go:87: 
=== RUN TestPeerSharing
--- PASS: TestPeerSharing (0.00 seconds)
=== RUN TestBadPeer
--- PASS: TestBadPeer (0.00 seconds)
=== RUN TestBootstrap
--- SKIP: TestBootstrap (0.00 seconds)
	peer_test.go:107: 
PASS
ok  	github.com/NebulousLabs/Sia/modules/gateway	0.004s
```

DavidVorick:
oh. Looks good then.

PR #426, opened 2015-04-03
Build separation

DavidVorick:
This separates the build constants and files into their own package. I took a slightly different approach than before. The types package still defines all of  the things like `BlockWindow`, but it defines them in `init()` depending on the value of  the constants placed in the build package.


lukechampine:
The `init()` approach is interesting, but ugly. It also makes it possible to modify the values, and (I'm assuming) prevents certain optimizations.

DavidVorick:
At least with the init() method, everything is in one file.

Less safe isn't a big deal, because we're already stuck with the more complex objects being 'var' anyway. Finally, we're not worried about optimizations right now.

I definitely think it's a lot cleaner than having 3 files with an identical set of variables with different comments for each.

types/constants.go
```diff
@@ -0,0 +1,141 @@
...
+import (
+	"math/big"
+	"time"
+
+	"github.com/NebulousLabs/Sia/build"
+)
+
+// constants.go contains the Sia constants. Depending on which build tags are
+// used, the constants will be initialized to different values.  and therefore
+// cannot be declared as constants.
```
lukechampine: typo

lukechampine:
hmm. I guess both approaches have their tradeoffs.
There are a few values that don't vary, like `GenesisUnlockHash`. Can you just initialize those in the declaration, and remove them from `init()`?

DavidVorick:
They don't vary currently, but as we start introducing them the dev keys will probably be stored somewhere in the code so anyone can test it. The release keys will only be accessible to us.

DavidVorick:
sorry, another sloppy rebase. idk how to undo it. I've been rather distracted today and it's causing problems.

DavidVorick:
I think it's all better now.

PR #427, opened 2015-04-05
Types cleanup + testing

DavidVorick:
I added a bunch of testing. We're effectively at 100%, although it's only reported at 89% by the cover tool. Lots of small bugs were discovered and squashed.

The one thing that's missing is checking that the encoding package can marshal all of the types. Perhaps some other time... I'm pretty burned out on the types package at this point.


types/currency.go
```diff
@@ -9,28 +16,23 @@ import (
...
+type (
+	// A Currency represents a number of siacoins or siafunds. Internally, a
+	// Currency value is unbounded; however, Currency values sent over the wire
+	// protocol are subject to a maximum size of 255 bytes (approximately 10^614).
+	// Unlike the math/big library, whose methods modify their receiver, all
+	// arithmetic Currency methods return a new value. Currency cannot be negative.
+	Currency struct {
+		i big.Int
+	}
+)
```
lukechampine: What's with the `type ( )`?

types/signatures_test.go
```diff
@@ -0,0 +1,371 @@
...
+
+// TestSortedUnique probes the sortedUnique function.
+func TestSortedUnique(t *testing.T) {
+	su := []uint64{
+		3,
+		5,
+		6,
+		8,
+		12,
+	}
```
lukechampine: all on one line pls

types/currency.go
DavidVorick: yeah

DavidVorick:
fixed

PR #428, opened 2015-04-07
experimental logging

lukechampine:
Exploring how we might do logging. The gateway now has fairly comprehensive logs. Eventually we may want to write a custom logger, but using the default one is okay for now. Some thoughts I had:

- Does it make sense to establish logging conventions (in the same style as our mutex conventions, that is)? e.g. "functions that return an error should not also log that error" or similar.
- If each module is writing to a separate log file, what's the best way to merge them so that we can see everything in chronological order? Or should all logs be written to the same file, and simply tagged by module? Should we write a "log viewer" tool to assist debugging?
- How can logs preserve context? That is, if there are two concurrent calls to the same function, their logs are going to be intertwined, so how do you tell how they're grouped?
- What's the best way to integrate logging with our `if DEBUG { panic() }` pattern? Maybe the logger could handle it implicitly, like:
```go
func (l *logger) Fatal(v ...interface{}) {
    msg := fmt.Sprint(v...)
    l.output(msg)
    if DEBUG {
        panic(msg)
    }
}
```

I also noticed that the tester package is apparently a module?? Is there some plan to expand its functionality? Because right now it's just the `TempDir` function. If it isn't clear how tester will be used, we could just move `TempDir` to the build package.
Speaking of which, `SiaTestingDir` is now organized by package, rather than throwing all the test folders into one directory. And all the tests use a random RPC port instead of incrementing a number. (This could be done with the API port too, but it'd be kind of a hassle.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,16 +1,6 @@
 cover
-hostDir
-hostdir
 release
-renterDownload
 whitepaper.aux
 whitepaper.log
 whitepaper.pdf
-modules/*/*.dat
-modules/*/*.backup
-*_test
-testdir*
-*.wallet*
-walletDir*
 *.swp
-info.log
--- a/api/miner_test.go
+++ b/api/miner_test.go
@@ -7,11 +7,14 @@ import (
 	"github.com/NebulousLabs/Sia/modules"
 )
 
-func (st *serverTester) testMining() {
+// TestMining starts the miner, mines a few blocks, and checks that the wallet
+// balance increased.
+func TestMining(t *testing.T) {
 	if testing.Short() {
-		st.Skip()
+		t.Skip()
 	}
 
+	st := newServerTester("TestMining", t)
 	// start miner
 	st.callAPI("/miner/start?threads=1")
 	// check that miner has started
@@ -28,11 +31,4 @@ func (st *serverTester) testMining() {
 	if walletstatus.FullBalance.IsZero() {
 		st.Fatalf("Mining did not increase wallet balance: %v", walletstatus.FullBalance.Big())
 	}
-}
-
-// TestMining starts the miner, mines a few blocks, and checks that the wallet
-// balance increased.
-func TestMining(t *testing.T) {
-	st := newServerTester(t)
-	st.testMining()
 }
--- a/api/renter_test.go
+++ b/api/renter_test.go
@@ -41,7 +41,8 @@ func TestUploadAndDownload(t *testing.T) {
 	}
 
 	// Try to download the file.
-	st.callAPI("/renter/download?destination=renterTestDL_test&nickname=first")
+	downloadName := "renterTestDL_test"
+	st.callAPI("/renter/download?nickname=first&destination=" + downloadName)
 	time.Sleep(time.Second * 2)
 
 	// Check that the downloaded file is equal to the uploaded file.
@@ -50,7 +51,7 @@ func TestUploadAndDownload(t *testing.T) {
 		t.Fatal(err)
 	}
 	defer upFile.Close()
-	downFile, err := os.Open("renterTestDL_test")
+	downFile, err := os.Open(downloadName)
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -66,4 +67,5 @@ func TestUploadAndDownload(t *testing.T) {
 	if upRoot != downRoot {
 		t.Error("uploaded and downloaded file have a hash mismatch")
 	}
+	os.Remove(downloadName)
 }
--- a/api/renter_test.go
+++ b/api/renter_test.go
@@ -17,7 +17,7 @@ func TestUploadAndDownload(t *testing.T) {
 	}
 
 	// Create a server and add a host to the network.
-	st := newServerTester(t)
+	st := newServerTester("TestUploadAndDownload", t)
 	st.announceHost()
 
 	for st.hostdb.NumHosts() == 0 {
--- a/modules/gateway/gateway.go
+++ b/modules/gateway/gateway.go
@@ -2,6 +2,7 @@ package gateway
 
 import (
 	"errors"
+	"log"
 	"net"
 	"os"
 	"time"
@@ -43,6 +44,8 @@ type Gateway struct {
 	// saveDir is the path used to save/load peers.
 	saveDir string
 
+	log *log.Logger
+
 	mu *sync.RWMutex
 }
 
@@ -63,6 +66,8 @@ func (g *Gateway) Close() error {
 // Bootstrap handles mutexes manually to avoid having a lock during network
 // communication.
 func (g *Gateway) Bootstrap(bootstrapPeer modules.NetAddress) (err error) {
+	g.log.Println("INFO: initiated bootstrapping to", bootstrapPeer)
+
 	// contact the bootstrap peer
 	if !g.Ping(bootstrapPeer) {
 		return errUnreachable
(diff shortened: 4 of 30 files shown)


DavidVorick:
Logging conventions make a lot of sense. I'm not sure what would be best. Having logs regarding things we expect to fail is a good starting place. I have no idea what sort of conventions might help logging.

I'm hoping that the modular design means we don't need to see things in chronological order between modules. Each module should be independent enough that it's own failure is either directly because of itself or entirely because of another module failing, and that the other module failed in an obvious way. I think we should start with 1 module per file and not worry about combining them until it becomes apparent that it's absolutely necessary.

Hopefully we can lock the log, or hold the logger under the same mutex that the rest of the module is under. Or we can use some sort of buffer that's guaranteed to atomically print to the log. If two functions are running concurrently, their output should not overlap in the log, though it would be nice to know that they were running at/around the same time. I guess timestamps will help with that. If logging writes can be made atomic, perhaps a function should log into a single string and then dump it into the log after it's done? I'm not sure. I want to keep logging as simple as possible.

I like integrating logging with `if DEBUG` stuff. Even if debugging is off, a message should still make the logs that something is wrong.

Originally the tester package had 4 or 5 functions that all did different things. Then I discovered tricks like ":0" that made those functions unnecessary. It's not clear to me how the tester package might evolve, so I would like to default to leaving it as a module. If more time passes and it remains 1 lonely function, we can find a better place for it.

.gitignore
```diff
@@ -1,16 +1,6 @@
...
 whitepaper.log
 whitepaper.pdf
-modules/*/*.dat
-modules/*/*.backup
-*_test
-testdir*
-*.wallet*
-walletDir*
 *.swp
-info.log
```
DavidVorick: lol my current working branch has these same changes

api/miner_test.go
```diff
@@ -7,11 +7,14 @@ import (
 	"github.com/NebulousLabs/Sia/modules"
 )
 
-func (st *serverTester) testMining() {
+// TestMining starts the miner, mines a few blocks, and checks that the wallet
+// balance increased.
+func TestMining(t *testing.T) {
 	if testing.Short() {
-		st.Skip()
+		t.Skip()
```
DavidVorick: SkipNow

api/renter_test.go
```diff
@@ -66,4 +67,5 @@ func TestUploadAndDownload(t *testing.T) {
 	if upRoot != downRoot {
 		t.Error("uploaded and downloaded file have a hash mismatch")
 	}
+	os.Remove(downloadName)
```
DavidVorick: I'd rather leave the file. If the renter has been moved to the new folder paradigm then it'll go in a temp folder instead of cluttering up the code folder.

modules/gateway/gateway.go
```diff
@@ -63,6 +66,8 @@ func (g *Gateway) Close() error {
 // Bootstrap handles mutexes manually to avoid having a lock during network
 // communication.
 func (g *Gateway) Bootstrap(bootstrapPeer modules.NetAddress) (err error) {
+	g.log.Println("INFO: initiated bootstrapping to", bootstrapPeer)
```
DavidVorick: INFO, WARN, ect. should all be constants. Probably established in the build package.

modules/gateway/peers.go
```diff
@@ -61,6 +64,7 @@ func (g *Gateway) addStrike(peer modules.NetAddress) {
 		return
 	}
 	g.peers[peer]++
+	g.log.Println("INFO: added a strike to peer", peer)
```
DavidVorick: can we get a reason for the strike?

modules/gateway/peers.go
```diff
@@ -39,6 +41,7 @@ func (g *Gateway) removePeer(peer modules.NetAddress) error {
 	}
 	delete(g.peers, peer)
 	g.save()
+	g.log.Println("INFO: removed peer", peer)
```
DavidVorick: Is there a way to get a reason for the peer's removal? I know there are strikes and there are Remove calls, perhaps this log should be moved to the caller instead of inside `removePeer`

I feel like logging should be as top level as possible, because it gives you more context.

modules/gateway/persist.go
```diff
@@ -31,3 +33,13 @@ func (g *Gateway) load(filename string) (err error) {
 	}
 	return
 }
+
+// create logger
+// TODO: when is the logFile closed? Does it need to be closed?
```
DavidVorick: Only needs to be closed when the daemon is closed. I guess that means each module should get a Close() function in addition to the logger.

lukechampine:
>Hopefully we can lock the log, or hold the logger under the same mutex that the rest of the module is under. Or we can use some sort of buffer that's guaranteed to atomically print to the log.

I considered this; it wouldn't be too hard to write methods for grouping logs together. The problem with making logs atomic is that you often want to see functions interacting with each other, rather than a full dump of function 1 followed by a full dump of function 2. So if you want to inspect the back-and-forth, you have to re-sort the log file by time.
I think the opposite approach is probably better, where logs remain in chronological order but can be "tagged" by group.

>It's not clear to me how the tester package might evolve, so I would like to default to leaving it as a module. If more time passes and it remains 1 lonely function, we can find a better place for it.

As I'm sure you know, my style leans more towards "if you can't immediately justify it, throw it out." But for now I'll hold off. I think there are ways to clean up the module testing that would fit well in a tester package (like moving all the `newTestingX` functions there). I guess we'll see.

modules/tester/tester.go
```diff
@@ -5,14 +5,14 @@ import (
...
-// TempDir takes a set of directory names and joins them to the sia temp
-// folder.
+// TempDir joins the provided directories and prefixes them with the Sia
+// testing directory.
 func TempDir(dirs ...string) string {
-	tmp := append([]string{os.TempDir()}, SiaTestingDir)
-	full := append(tmp, dirs...)
-	return filepath.Join(full...) // filepath.Join(tmp, testing, dirs...) returns 'too many arguments' error.
+	path := filepath.Join(SiaTestingDir, filepath.Join(dirs...))
+	os.RemoveAll(path) // remove old test data
```
DavidVorick: Is this the right place for the RemoveAll? I prefer it to be in the makefile.

DavidVorick: maybe instead we should check if the path already exists and complain if it does.

DavidVorick:
I'm not sure if this is true of what's currently on master, but the release binaries don't synchronize on launch. You either have to wait a while (seems like more than 2 minutes), or you have to manually call /gateway/synchronize

maybe it's a problem with threadedSynchronize.

modules/gateway/persist.go
lukechampine: I think in this case it's actually fine to not close the file. It's standard practice to close resources *when you're done with them*, but we're never really "done with" logging until the program exits.

modules/gateway/gateway.go
lukechampine: yeah I was assuming they would eventually become constants (or the logger would just have `Info`, `Warn`, etc. methods).

api/renter_test.go
lukechampine: good catch, for some reason it didn't occur to me to move the downloaded file to the tempdir.

modules/gateway/peers.go
lukechampine: there will be a log of every time an RPC returned an error, which counts as a strike. So if you want to see why a peer was booted you can grep the log for its IP. Not perfect, but certainly better than passing the error messages around or storing them somewhere.

I think context is probably the hardest problem to solve in logging. The downside of top-level logging is that you have to do a lot more of it; that is, instead of one log in `removePeer`, you need a separate log for every single function that calls `removePeer`. My intuition is that you can get most of the context you need via `runtime.Caller`, but that introduces its own problems, like "when should a log report its caller, and when shouldn't it?"

modules/tester/tester.go
lukechampine: The problem with doing it in the Makefile is that it's not cross-platform. See the comment on c8b2e441f9cb6aa52909caf7aa1fcaa17a6cb061

lukechampine:
Hmm. Maybe logging can tell us.

PR #429, opened 2015-04-07
Stream encoding

lukechampine:
The encoding package now works with readers and writers instead of byte slices. This is how json and gob do things, and I can see why.

Also some minor fixes that didn't make it into the last PR.


lukechampine:
Also, the build is randomly timing out. No idea why, but we should investigate for sure.

encoding/marshal.go
```diff
@@ -1,91 +1,110 @@
...
-// An Unmarshaler can be decoded from a byte slice. If a decoding error occurs,
-// UnmarshalSia should panic.
-type SiaUnmarshaler interface {
-	UnmarshalSia([]byte)
-}
-
-// Marshal encodes a value as a byte slice. The encoding rules are as follows:
+// Package encoding converts arbitrary objects into byte slices, and vis
+// versa. It also contains helper functions for reading and writing length-
+// prefixed data. The encoding rules are as follows:
```
DavidVorick: package comments should go in packagename.go, file comments should refer the specific file. (eg 'marshal.go')

encoding/marshal.go
```diff
@@ -1,91 +1,110 @@
...
 // Integers are little-endian, and are always encoded as 8 bytes, i.e. their
 // int64 or uint64 equivalent.
 //
-// Booleans are encoded as one byte, either zero (false) or non-zero (true).
+// Booleans are encoded as one byte, either zero (false) or one (true). No
+// other values may be used.
 //
-// Nil pointers are represented by a zero.
+// Nil pointers are represented by a zero. Valid pointers are represented by a
+// non-zero, followed by the encoding of the dereferenced value.
```
DavidVorick: Just like with the bools, can we have the zero/non-zero switch to zero/one?

encoding/marshal.go
```diff
@@ -94,111 +113,164 @@ func marshal(val reflect.Value) (b []byte) {
...
+	_, err := io.ReadFull(d.r, b)
+	if err != nil {
+		panic(err)
+	}
+	return b
+}
+
+func (d *Decoder) readPrefix() []byte {
+	// TODO: what should maxlen be?
+	b, err := ReadPrefix(d.r, 1<<32)
```
DavidVorick: Maybe it should be defined at the top of the file?

Golang has a maximum amount of memory that it's allowed to use, I believe it's around 1<<30.

encoding/marshal.go
```diff
@@ -94,111 +113,164 @@ func marshal(val reflect.Value) (b []byte) {
...
 	case reflect.Ptr:
+		b := d.readN(1)
 		// nil pointer, nothing to decode
 		if b[0] == 0 {
-			return 1
+			return
 		}
 		// make sure we aren't decoding into nil
 		if val.IsNil() {
 			val.Set(reflect.New(val.Type().Elem()))
```
DavidVorick: Can you do this? Won't the result just be the same as returning early? (because the caller is still looking at a nil value)

Should there be a panic here?

DavidVorick: wait, decode should *never* panic, because you could be decoding anything. An error should be returned.

encoding/marshal.go
lukechampine: are you referring to a common practice, or a Sia-specific one? Because there's no encoding.go here, and I think it'd be silly to make a new file just for the docstring. It makes sense in marshal.go.

encoding/marshal.go
lukechampine: since `len()` returns an int I figured 32 bits made sense. In practice I don't think will be an issue because we enforce a smaller maxlen when doing network communication anyway.

encoding/marshal.go
lukechampine: It's plausible that you might want to encode a nil pointer, e.g. in the last node of a linked list. I don't think we make use of it anywhere, but it's certainly not an error.

btw, decode *does* panic, but the panic will be caught by Decode.

encoding/marshal.go
lukechampine: good catch

encoding/marshal.go
DavidVorick: oh. I was just trying to refer to the standard practice.

encoding/marshal.go
DavidVorick: oh. That's good enough I guess. Is it the same way with marshal?

DavidVorick:
Is this ready to go?

I too have noticed the build randomly timing out. I'm 98% sure it started when I flipped the subscription model of the wallet. I think there's an api call that tries to wait for an update which has already happened. It hasn't shown up on travis ci yet though so I've mostly ignored it.

It's not currently super high on my todo list, but it'll get fixed when I move the api package over to the new testing paradigms. I don't think it's a simple fix but I also don't think it's a difficult fix. It'll just take time.

lukechampine:
Yep, good to go.
If you want to hunt down the bug you can use git bisect. Might be tedious waiting for all those tests to time out though.

encoding/marshal.go
lukechampine: nope, Marshal will panic if you try to marshal a bad type. Encode, however, can return an error, because it outputs to an `io.Writer` and the `Write` call can fail.

PR #432, opened 2015-04-09
Hostdb Rewrite

DavidVorick:
It's not a complete rewrite because a good amount of the old code still remains. But everything got moved around and almost the entire test suite got scrapped.

The new testing framework is in place, and is going to be much nicer to use and read and reason about. I've also done some cleanup stuff along the way, as I notice things which could be improved I've tried to make a habit of improving them on the spot, so they don't get lost in limbo and then never completed.

I removed the sybil attack defenses from the hostdb because I think they were hurting more than helping. We're basically resigned to acknowledging that we can't stop sybil attacks from hosts until we start demanding that they burn coins. I'd rather be vulnerable to sybil attacks than force hosts to burn coins until we've got a bigger network.

I didn't implement any of the other host improvements that I was planning on adding (continuously checking if hosts are active or not, and changing the RandomHost call to avoid returning hosts that the caller already has). These will be put in soon (after testing is brought back) and hopefully that will make the hostdb a lot better. I also haven't added an api call to query the hosts, though I changed the interface so making such an api call would be easy.


modules/Modules.md
```diff
@@ -4,9 +4,20 @@ Modules Conventions
 The modules are a work in progress, not all modules have implemented all
 conventions.
 
-All modules should have an update.go which contains the functions for
-integrating changes in the environment into the module (for example, changes in
-the consensus set).
+All modules have a ModuleNotify method, which returns a channel that receives a
+struct{} every time that the is an update to the module. The ModuleNotify class
```
lukechampine: that there is

modules/hostdb/hostdb.go
```diff
@@ -2,59 +2,100 @@ package hostdb
...
+		allHosts map[modules.NetAddress]*modules.HostEntry
+
+		mu *sync.RWMutex
+	}
+)
 
 // New returns an empty HostDatabase.
-func New(s *consensus.State, g modules.Gateway) (hdb *HostDB, err error) {
-	if s == nil {
+func New(cs *consensus.State, g modules.Gateway) (hdb *HostDB, err error) {
```
lukechampine: not `modules.ConsensusSet`?

modules/hostdb/hostentry.go
```diff
@@ -0,0 +1,92 @@
...
+// from the set of active hosts.
+func (hdb *HostDB) threadedProbeHost(entry *modules.HostEntry) {
+	// If the host is already in the set of active hosts, remove it. It will be
+	// re-added if it responds to the probing, and left out if not. The db
+	// needs to be locked for this operation, but then unlocked while the
+	// network communication is happening.
+	id := hdb.mu.Lock()
+	{
+		node, exists := hdb.activeHosts[entry.IPAddress]
+		if exists {
```
lukechampine: I assume the bare block is a style thing

DavidVorick: I felt like I needed to highlight the Lock and Unlock, but that separating them with newlines didn't make it clear enough. So I put the locked code a bare block.

PR #437, opened 2015-04-13
Hostdb tests

DavidVorick:
This PR ended up going a lot different direction that I was expecting. I wrote a few tests, decided that the way we were doing things was bad, then decided that I wasn't sure of a better way to do things. So after a lot of contemplation, I decided that it's premature to be trying to make a great hostdb.

I did wrote a scanning function which will move hosts from inactive to active, something that thus far didn't exist. I also added 2 api calls, one which will tell you the list of active hosts, and one which will tell you the list of inactive hosts.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,7 @@ xc: dependencies test test-long REBUILD
 # clean removes all directories that get automatically created during
 # development.
 clean:
-	rm -rf /tmp/SiaTesting release whitepaper.aux whitepaper.log whitepaper.pdf
+	rm -rf release whitepaper.aux whitepaper.log whitepaper.pdf
 
 # test runs the short tests for Sia, and aims to always take less than 2
 # seconds.
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,7 @@ xc: dependencies test test-long REBUILD
 # clean removes all directories that get automatically created during
 # development.
 clean:
-	rm -rf hostdir release whitepaper.aux whitepaper.log whitepaper.pdf
+	rm -rf /tmp/SiaTesting release whitepaper.aux whitepaper.log whitepaper.pdf
 
 # test runs the short tests for Sia, and aims to always take less than 2
 # seconds.
--- a/api/hostdb.go
+++ b/api/hostdb.go
@@ -1,0 +1,23 @@
+package api
+
+import (
+	"net/http"
+
+	"github.com/NebulousLabs/Sia/modules"
+)
+
+// ActiveHosts is the struct that pads the response to the hostdb module call
+// "ActiveHosts". The padding is used so that the return value can have an
+// explicit name, which makes adding or removing fields easier in the future.
+type ActiveHosts struct {
+	Entries []modules.HostEntry
+}
+
+// hostdbHostsActiveHandler handes the API call asking for the list of active
+// hosts.
+func (srv *Server) hostdbHostsActiveHandler(w http.ResponseWriter, req *http.Request) {
+	ah := ActiveHosts{
+		Entries: srv.hostdb.ActiveHosts(),
+	}
+	writeJSON(w, ah)
+}
--- a/api/hostdb.go
+++ b/api/hostdb.go
@@ -21,3 +21,11 @@ func (srv *Server) hostdbHostsActiveHandler(w http.ResponseWriter, req *http.Req
 	}
 	writeJSON(w, ah)
 }
+
+// hostdbHostsAllHandler handes the API call asking for the list of all hosts.
+func (srv *Server) hostdbHostsAllHandler(w http.ResponseWriter, req *http.Request) {
+	ah := ActiveHosts{
+		Entries: srv.hostdb.AllHosts(),
+	}
+	writeJSON(w, ah)
+}
--- a/modules/hostdb.go
+++ b/modules/hostdb.go
@@ -40,8 +40,9 @@ type HostSettings struct {
 // coins frozen in the host's announcement transaction.
 type HostEntry struct {
 	HostSettings
-	IPAddress NetAddress
-	Weight    types.Currency
+	IPAddress   NetAddress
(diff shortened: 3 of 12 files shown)


Makefile
```diff
@@ -47,7 +47,7 @@ xc: dependencies test test-long REBUILD
 # clean removes all directories that get automatically created during
 # development.
 clean:
-	rm -rf hostdir release whitepaper.aux whitepaper.log whitepaper.pdf
+	rm -rf /tmp/SiaTesting release whitepaper.aux whitepaper.log whitepaper.pdf
```
lukechampine: os-dependent :/

api/hostdb.go
```diff
@@ -0,0 +1,31 @@
...
+
+	"github.com/NebulousLabs/Sia/modules"
+)
+
+// ActiveHosts is the struct that pads the response to the hostdb module call
+// "ActiveHosts". The padding is used so that the return value can have an
+// explicit name, which makes adding or removing fields easier in the future.
+type ActiveHosts struct {
+	Entries []modules.HostEntry
+}
```
lukechampine: what would an `ActiveHosts` object contain other than a list of hosts?

modules/hostdb.go
```diff
@@ -40,8 +40,9 @@ type HostSettings struct {
 // coins frozen in the host's announcement transaction.
 type HostEntry struct {
 	HostSettings
-	IPAddress NetAddress
-	Weight    types.Currency
+	IPAddress   NetAddress
+	Weight      types.Currency
+	Reliability types.Currency
```
lukechampine: why is Reliability a Currency??

modules/hostdb/scan.go
```diff
@@ -0,0 +1,102 @@
...
+		// Beacuse there was an error, decrement the reliability.
+		entry.Reliability = entry.Reliability.Sub(types.NewCurrency64(UnreachablePenalty))
+
+		// If the reliability has fallen below InactiveReliability, remove the host from the list
+		// of active hosts.
+		node, exists := hdb.activeHosts[entry.IPAddress]
+		if exists && entry.Reliability.Cmp(types.NewCurrency64(InactiveReliability)) < 0 {
+			delete(hdb.activeHosts, entry.IPAddress)
+			node.removeNode()
+		}
```
lukechampine: maybe factor this out into an `hdb.removeNode(entry.IPAddress)`, which would be a no-op if the entry didn't exist.

modules/hostdb/scan.go
```diff
@@ -0,0 +1,102 @@
...
+		//
+		// TODO: enforce some limit on the number of hosts that will be
+		// queried.
+		id := hdb.mu.Lock()
+		{
+			for _, host := range hdb.allHosts {
+				go hdb.threadedProbeHost(host)
+			}
+		}
+		hdb.mu.Unlock(id)
```
lukechampine: more weird block style...

Makefile
DavidVorick: That's true. I thought it'd be okay in the Makefile but I guess not.

api/hostdb.go
DavidVorick: The API should always return objects wrapped in a struct. It makes it easier for us to change things later, and it makes things more generic and predictable.

The big advantage here is you can request the field 'Entries' specifically, so if we change the output, we can change the name and preserve compatibility.

modules/hostdb.go
DavidVorick: Gives us more flexibility and access to unbounded unsigned integers. Reliability right now could just be a uint8 but I'm not sure how that's going to change in the future.

Weight on the other hand needs a lot of bits. It could probably just be a Big but things would break if there were negative numbers, I'd prefer to have the unsigned safety of the currency type.

I do need to take a second look at the equations, maybe there's a better way to do things. Maybe I should be using big.Rat everywhere.

modules/hostdb/scan.go
DavidVorick: the block style makes more sense on the new branch, where the block is surrounding a significantly larger portion of code. Also when I originally wrote this there was a lot more code.

It's important that if you're not doing `lock && defer unlock` that you don't lose track of where the unlock happens.

modules/hostdb/scan.go
DavidVorick: hmm yea. `decrementReliability` probably makes sense.

api/hostdb.go
lukechampine: I guess I see the value of it, it just strikes me as needlessly verbose right now. Probably because I can't think of what else you would want to include in that struct. But I suppose if you're going to do it here you should do it for the renter as well.

modules/hostdb.go
lukechampine: I'm averse to using the wrong type for the job. Currency is *almost* the right type for Weight, so I guess keeping that wouldn't be so bad. But Reliability definitely seems like it should be a fractional type, either a float or a big.Rat.

modules/hostdb/scan.go
lukechampine: yeah I figured you would flesh it out a bit. For functions that manage their own locks it seems like a good idea.

modules/hostdb/scan.go
```diff
@@ -0,0 +1,112 @@
...
+
+// decrementReliability reduces the reliability of a node, moving it out of the
+// set of active hosts or deleting it entirely if necessary.
+func (hdb *HostDB) decrementReliability(addr modules.NetAddress, penalty types.Currency) {
+	// Look up the entry and decrement the reliability.
+	entry, exists := hdb.allHosts[addr]
+	if !exists {
+		return
+	}
+	entry.Reliability = entry.Reliability.Sub(UnreachablePenalty)
```
lukechampine: I believe you want to pass in `penalty` here

DavidVorick: drat

Issue #435, opened 2015-04-13
Enhanced Synchronize Rules


DavidVorick:
I was poking around with a bunch of daemons on my computer and noticing that synchronization didn't always happen right away, but once things were synchronized they seem to maintain it. So I wanted to suggest a few rules and see what you think.

One thing that was apparent is that if you don't have peers (but peers have you), you will fall behind. You call resynchronize but it doesn't help you because you don't have the peer who sent you the block. Basically, relationships aren't reciprocal. If you get something unknown, you should ask the person who sent it to you for more information, even if you don't have them in your peer list. (I think. I'm not sure how DoS plays into this). Because I'm not sure about DoS stuff, this didn't get a number.

1. Synchronize should be called upon getting a new peer, on the new peer. I don't think this is currently the case, even after adding peers I had to manually call 'synchronize'.

2. Synchronize should go both directions. Or rather, when someone calls synchronize on you, there should be something that checks if they have blocks you don't recognize. If they do, you should call synchronize on them. Maybe this is overzealous and unneeded.

3. Peer adding should be reciprocal. I know we had worries about this in the past, but I was thinking you should be aggressive about finding peers when you have less than n. (n = 12?) If you m peers and m > n, you should add the peer with probability 1/(m-n), and otherwise put them in some container so they can't just call add on you 10000 times to guarantee they'll make it. Once you've rejected them the rejection is permanent.  Anyway, up to some number of peers, you should add anyone who adds you.

4. When requesting peers, you should ask everyone you know for a list, and then from that aggregate you should randomly select peers until you have N total (12?)

lukechampine:
Sounds reasonable. Reciprocal peer adding is something I've considered for a while, I just wasn't sure about the possible downsides. Bitcoin seems to take a similar approach, e.g. it always asks new peers for *their* peers, which we don't do (yet). Bitcoin nodes also keep track of what they think their peers' peer lists look like, and rebroadcast to them every 24 hours.

DavidVorick:
That's probably useful to have as well.

PR #442, opened 2015-04-15
More robust peering

lukechampine:
A few peering tweaks, most notably to initial peer discovery and the resynchronization loop.

This partially addresses #435.

### DIFF
--- a/api/gateway_test.go
+++ b/api/gateway_test.go
@@ -22,7 +22,6 @@ func (st *serverTester) addPeer(name string) *serverTester {
 	if err != nil {
 		st.Fatal("bootstrap failed:", err)
 	}
-
 	// Wait for bootstrapping to finish, then check that each has the same
 	// number of peers and blocks.
 	for len(st.gateway.Info().Peers) != len(newPeer.gateway.Info().Peers) {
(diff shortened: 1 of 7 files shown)


api/gateway_test.go
```diff
@@ -51,18 +50,22 @@ func TestPeering(t *testing.T) {
...
 	}
 
-	// peer2 should have received peer3 via peer1. Note that it does not have
-	// peer1 though, because /gateway/add does not contact the added peer.
+	// When peer3 bootstraps to peer1, it should learn about peer2 and
+	// announce to it.
 	peer2.getAPI("/gateway/status", &info)
-	if len(info.Peers) != 1 || info.Peers[0] != peer3.netAddress() {
+	for len(info.Peers) != 1 {
+		time.Sleep(time.Millisecond)
```
DavidVorick: Why is there a millisecond sleep?

api/gateway_test.go
```diff
@@ -51,18 +50,22 @@ func TestPeering(t *testing.T) {
...
+		peer2.getAPI("/gateway/status", &info)
+	}
```
DavidVorick: We should probably add a gateway subscription channel here instead of busywaiting.

modules/gateway/gateway.go
```diff
@@ -89,13 +89,15 @@ func (g *Gateway) Bootstrap(bootstrapPeer modules.NetAddress) (err error) {
...
-	g.requestPeers(bootstrapPeer)
-	// save new peers
-	g.save()
+	// ask the bootstrapPeer to add us back
+	go g.RPC(bootstrapPeer, "AddMe", writerRPC(g.Address()))
 
-	// announce ourselves to the new peers
-	go g.threadedBroadcast("AddMe", writerRPC(g.myAddr))
+	// initial peer discovery
+	go func() {
```
DavidVorick: add some more comments here explaining what's happening and why it's happening that way instead of some other way.

api/gateway_test.go
DavidVorick: I guess we'll get there when we ravamp the api test suite.

PR #444, opened 2015-04-16
Hostdb polish

DavidVorick:
I moved the HostEntry type to the hostdb package. I renamed a few functions, and changed the way a few variables work. I added some testing. I struggled a lot with rebaseing and merging, as you can see form the commit history.

Nothing really fundamental changed. I want to test more but will be holding off until we decide how the RPC stuff should work, and until we have a better idea of how all the modules are supposed to fit together.

I don't think that the module situation is very dire. I think what we have is very close to good and there are some corner cases (particularly around the gateway) that need to be reconsidered. Beyond that, I do really like what we have and don't think any more major changes to the fundamental design will be necessary.

### DIFF
--- a/modules/hostdb/hostentry_test.go
+++ b/modules/hostdb/hostentry_test.go
@@ -1,0 +1,49 @@
+package hostdb
+
+import (
+	"testing"
+
+	"github.com/NebulousLabs/Sia/modules"
+	"github.com/NebulousLabs/Sia/types"
+)
+
+// TestHostWeight probes the hostWeight function.
+func TestHostWeight(t *testing.T) {
+	hdbt := newHDBTester("TestHostWeight", t)
+
+	// Create two identical entries, except that one has a price that is 2x the
+	// other. The weight returned by hostWeight should be 1/8 for the more
+	// expensive host.
+	entry1 := hostEntry{
+		HostSettings: modules.HostSettings{
+			Price: types.NewCurrency64(3),
+		},
+	}
+	entry2 := hostEntry{
+		HostSettings: modules.HostSettings{
+			Price: types.NewCurrency64(6),
+		},
+	}
+
+	weight1 := hdbt.hostdb.hostWeight(entry1)
+	weight2 := hdbt.hostdb.hostWeight(entry2)
+	expectedWeight := weight1.Div(types.NewCurrency64(8))
+	if weight2.Cmp(expectedWeight) != 0 {
+		t.Error("Weight of expensive host is not the correct value.")
+	}
+
+	// Try a 0 price.
+	entry3 := hostEntry{
+		HostSettings: modules.HostSettings{
+			Price: types.NewCurrency64(0),
+		},
+	}
+	weight3 := hdbt.hostdb.hostWeight(entry3)
+	if weight3.Cmp(weight1) <= 0 {
+		t.Error("Free host not weighing fairly")
+	}
+}
+
+// TestInsertHost probes the insertHost and InsertHost functions.
+func TestInsertHost(t *testing.T) {
+}
--- a/modules/hostdb/hostentry_test.go
+++ b/modules/hostdb/hostentry_test.go
@@ -67,13 +67,8 @@ func TestInsertHost(t *testing.T) {
 	if len(hdbt.hostdb.allHosts) != 2 {
 		t.Error("host was not inserted")
 	}
-
-	// TODO: Bring this stuff back. Can't do it until the API module RPC etc.
-	// stuff is sorted out.
-	/*
 	hdbt.hdbUpdateWait()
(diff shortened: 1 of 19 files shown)


modules/hostdb/hostentry_test.go
```diff
@@ -0,0 +1,74 @@
...
+	entry1 := hostEntry{
+		HostSettings: modules.HostSettings{
+			Price: types.NewCurrency64(3),
+		},
+	}
+	entry2 := hostEntry{
+		HostSettings: modules.HostSettings{
+			Price: types.NewCurrency64(6),
+		},
+	}
```
lukechampine: since `HostSettings` is embedded, you could alternatively write:
```go
var entry1, entry2 hostEntry
entry1.Price = types.NewCurrency64(3)
entry2.Price = types.NewCurrency64(6)
```

modules/hostdb/scan.go
```diff
@@ -79,10 +95,60 @@ func (hdb *HostDB) threadedProbeHost(entry *modules.HostEntry) {
...
 	for {
+		// Determine who to scan. At most 'MaxActiveHosts' will be scanned,
+		// starting with the active hosts followed by a random selection of the
+		// inactive hosts.
+		id := hdb.mu.Lock()
+		{
+			// Check all of the active hosts.
+			for _, host := range hdb.activeHosts {
+				go hdb.threadedProbeHost(host.hostEntry)
+			}
```
lukechampine: `threadedProbeHost` needs a lock, since it won't update anything until this block ends, right?

modules/hostdb/scan.go
```diff
@@ -79,10 +95,60 @@ func (hdb *HostDB) threadedProbeHost(entry *modules.HostEntry) {
...
+
+			// Assemble all of the inactive hosts into a single array and
+			// shuffle it.
+			i := 0
+			random := make([]*hostEntry, len(hdb.allHosts))
+			for _, entry := range hdb.allHosts {
+				random[i] = &entry
+				i++
+			}
```
lukechampine: Doesn't `allHosts` contain active hosts as well?

modules/hostdb/scan.go
```diff
@@ -79,10 +95,60 @@ func (hdb *HostDB) threadedProbeHost(entry *modules.HostEntry) {
...
+				}
+
+				n := int(N.Int64()) + i
+				tmp := random[i]
+				random[i] = random[n]
+				random[n] = tmp
+			}
+
+			// Select the first InactiveHostCheckupQuantity hosts from the
+			// shuffled list.
```
lukechampine: seems like overkill, but I may be wrong. The important factors are: `InactiveHostCheckupQuantity`, `len(allHosts)`, and how frequently `Scan` is run. If probing inactive hosts is low-priority, then it would probably suffice to just select N random hosts (without the guarantee of uniqueness).

modules/hostdb/scan.go
DavidVorick: it's `threaded`, it'll manage its own locks

DavidVorick: I'm not sure I understand your question, but none of the threadedProbeHost calls will start until all of them have been created.

I'm not sure what sort of memory impact this will have but I'm guess it'll be O(1mb)

modules/hostdb/scan.go
DavidVorick: oh yeah :(

modules/hostdb/scan.go
DavidVorick: I did realize that some other process might remove the host from active hosts between the locks, so then threadedProbeHost might be acting on a missing host.

\>.< foo

modules/hostdb/scan.go
DavidVorick: The sad part is trying to avoid the active hosts.

modules/hostdb/scan.go
DavidVorick: oh nvm it's not a problem anyway. There's already an implicit assumption that a host may or may not exist when it's being probed.

modules/hostdb/scan.go
lukechampine: naive solution: iterate through `allHosts` at random, and check that the host isn't in `activeHosts` before you probe it.

modules/hostdb/scan.go
lukechampine: could you just move the active host check to the end of the block?

modules/hostdb/scan.go
lukechampine: btw: `random[i], random[n] = random[n], random[i]` ;)

DavidVorick: oh

PR #445, opened 2015-04-19
Siafund keygen

DavidVorick:
This adds a cli tool for generating multisig keys and addresses. While there is no way to spend the results, support for that will be added before launch. At least this way people can get their keys.

The files created are a simple struct:
```go
type KeyPair struct {
    Header string
    Version string
    Index int
    SecretKey crypto.SecretKey
    UnlockConditions types.UnlockConditions
}
```

The header is for identifying the file without knowing the extension. The version is in case we need to change anything. Then there's an index, which allows you to find the public key associated with your secret key. Finally there's the unlock conditions, which is required for spending an address.

The keys generated can be used for siacoins or siafunds, but are mostly intended for helping people spend siafunds.

### DIFF
--- a/siakg/keys.go
+++ b/siakg/keys.go
@@ -32,6 +32,13 @@ type KeyPair struct {
 
 // generateKeys will generate a set of keys and save the keyfiles to disk.
 func generateKeys(*cobra.Command, []string) {
+	// Check that the total number of keys is at least as large as the required
+	// number of keys.
+	if config.Siakg.TotalKeys < config.Siakg.RequiredKeys {
+		fmt.Printf("Total Keys (%v) must be greater than or equal to Required Keys (%v)\n", config.Siakg.TotalKeys, config.Siakg.RequiredKeys)
+		return
+	}
+
 	fmt.Printf("Creating key '%s' with %v total keys and %v required keys.\n", config.Siakg.KeyName, config.Siakg.TotalKeys, config.Siakg.RequiredKeys)
 
 	// Generate 'TotalKeys' keyparis and fill out the metadata.
--- a/siakg/keys.go
+++ b/siakg/keys.go
@@ -95,6 +95,66 @@ func generateKeys(*cobra.Command, []string) {
 		}
 	}
 
+	// Load the keys from disk back into memory, then verifiy that the keys on
+	// disk are able to sign outputs in transactions.
+	loadedKeys := make([]KeyPair, config.Siakg.TotalKeys)
+	for i := 0; i < len(loadedKeys); i++ {
+		err := encoding.ReadFile(filepath.Join(config.Siakg.Folder, config.Siakg.KeyName)+"_Key"+strconv.Itoa(i)+FileExtension, &loadedKeys[i])
+		if err != nil {
+			fmt.Println(err)
+			return
+		}
+	}
+
+	// Check that the keys can be used to spend transactions. Load them back into memory.
+	for _, loadedKey := range loadedKeys {
+		if loadedKey.UnlockConditions.UnlockHash() != uc.UnlockHash() {
+			fmt.Println("corruption occured while saving the keys to disk")
+		}
+	}
+	// Create a transaction for the keys to sign.
+	txn := types.Transaction{
+		SiafundInputs: []types.SiafundInput{
+			types.SiafundInput{
+				UnlockConditions: loadedKeys[0].UnlockConditions,
+			},
+		},
+	}
+	// Loop through and sign the transaction multiple times. All keys will be
+	// used at least once by the time the loop terminates.
+	var i int
+	for i != config.Siakg.TotalKeys {
+		// i tracks which key is next to be used. If i + RequiredKeys results
+		// in going out-of-bounds, reduce i so that the last key will be used
+		// for the final signature.
+		if i+config.Siakg.RequiredKeys > config.Siakg.TotalKeys {
+			i = config.Siakg.TotalKeys - config.Siakg.RequiredKeys
+		}
+		var j int
+		for j < config.Siakg.RequiredKeys {
+			txn.Signatures = append(txn.Signatures, types.TransactionSignature{
+				PublicKeyIndex: uint64(i),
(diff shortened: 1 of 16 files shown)


types/signatures.go
```diff
@@ -57,7 +57,7 @@ type (
 		SiafundOutputs           []uint64
 		MinerFees                []uint64
 		ArbitraryData            []uint64
-		Signatures               []uint64
+		TransactionSignatures    []uint64
```
lukechampine: what exactly is unambiguous about `txn.Signatures`? "transaction dot transaction signatures" stutters.

siakg/keys.go
```diff
@@ -0,0 +1,172 @@
...
+
+const (
+	// The header for all siakg files. Do not change.
+	FileHeader = "siakg"
+)
+
+// A KeyPair is the object that gets saved to disk for a signature key. All the
+// information necessary to sign a transaction is in the struct, and the struct
+// can be directly written to disk.
+type KeyPair struct {
```
lukechampine: fyi you don't need to export a type to marshal it, just the fields. I guess it doesn't make much difference since this is a `main` anyway.

siakg/keys.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+	for i, key := range keys {
+		err := encoding.WriteFile(filepath.Join(config.Siakg.Folder, config.Siakg.KeyName)+"_Key"+strconv.Itoa(i)+FileExtension, key)
+		if err != nil {
+			fmt.Println(err)
+			return
+		}
+	}
+
+	// Load the keys from disk back into memory, then verifiy that the keys on
```
lukechampine: verify

siakg/main.go
```diff
@@ -0,0 +1,68 @@
...
+		Folder       string
+		KeyName      string
+		RequiredKeys int
+		TotalKeys    int
+	}
+
+	Address struct {
+		Filename string
+	}
+}
```
lukechampine: a config file seems like overkill for such a simple utility.

siakg/main_test.go
```diff
@@ -0,0 +1,84 @@
...
+	"path/filepath"
+	"testing"
+
+	"github.com/NebulousLabs/Sia/modules/tester"
+)
+
+// The tests in this file need to be checked manually, even if errors are
+// returned it will not be apparent to the test suite.
+//
+// CONTRIBUTE: Make these tests automatic.
```
lukechampine: this would be more easily testable if the key-generation code was separated into a pure function.

types/signatures.go
```diff
@@ -96,17 +96,17 @@ type (
...
+	// If 'RequiredSignatures' == 0, the UnlockConditions are effectively "anyone can
+	// unlock." If 'RequiredSignatures' > len('PublicKeys'), then the UnlockConditions
 	// cannot be fulfilled under any circumstances.
 	UnlockConditions struct {
-		Timelock      BlockHeight
-		PublicKeys    []SiaPublicKey
-		NumSignatures uint64
+		Timelock           BlockHeight
+		PublicKeys         []SiaPublicKey
+		RequiredSignatures uint64
```
lukechampine: ditto here, `NumSignatures` seems pretty clear to me

types/signatures.go
DavidVorick: The problem is that there's a TransactionSignature type and a Signature type, and it's weird to call Trasaction.Signatures and not have it reference the Signature type.

I've been waffling over this one for a bit but it has confused me a few times.

siakg/main.go
DavidVorick: oh, it should say 'keyinfo'

siakg/main_test.go
DavidVorick: I thought about that. I'll probably do it. I might also add a global error variable, though that's pretty ugly.

types/signatures.go
DavidVorick: NumSignatures is confusing because multisig has 2 "num signatures". There's the number of signatures required, and there's the number of different keys that you can use.

hmm. I guess that's not so confusing. Maybe it was okay the other way.

DavidVorick: `NumRequiredSignatures`

siakg/main_test.go
lukechampine: why would you need a global error?

types/signatures.go
lukechampine: yeah I knew that was coming haha

siakg/main.go
lukechampine: I think in the case of `KeyInfo` it makes more sense to make the filename a required argument, rather than a flag. Probably true of a few other flags as well, like `RequiredKeys` and `TotalKeys`.

DavidVorick: the `folder` flag must stay. I can change the default to `""` though. I'm not sure how to do required arguments? I only know how to make new flags and whole new commands.

siakg/main_test.go
DavidVorick: you set the error during main, then the testing package can see that it got set to the expected value. This is only necessary b/c main can't return an error.

I'll start with the separate functions, and see how effective that is.

types/signatures.go
DavidVorick: Actually, what if we did `SignaturesRequired`? To me that firmly implies it's a numerical value.

siakg/main.go
lukechampine: a cobra function takes a `args []slice` as its second parameter. So just like with `os.Args`, you can make your function return an error if `len(args) < 1`, etc.

siakg/main_test.go
lukechampine: if you separate out the function, that function can return an error.

types/signatures.go
lukechampine: fine by me

DavidVorick:
of all the reasons for a built to fail... I didn't go fmt :(

lukechampine:
the point of refactoring was so that you could write tests that didn't just call `main()` with different flags...

PR #447, opened 2015-04-20
Rename Sia Keygen to Sia Address Generator

DavidVorick:
Also, the defaults for multisig have been changed to 1-of-1.

Also, a readme has been added explaining how to use siag.


siag/keys.go
```diff
@@ -196,8 +196,12 @@ func printKeyInfo(filename string) error {
 
 // keyInfo receives the cobra call 'keyInfo', and is essentially a wrapper for
 // printKeyInfo. This structure makes testing easier.
-func keyInfo(*cobra.Command, []string) {
-	err := printKeyInfo(config.KeyInfo.Filename)
+func keyInfo(c *cobra.Command, args []string) {
+	if len(args) != 1 {
+		fmt.Println("Usage: siag keyinfo [filename]")
```
lukechampine: generally you should call `c.Usage()` but this is fine

PR #449, opened 2015-04-20
don't overwrite existing keys

DavidVorick:
also a quick fix to the makefile

### DIFF
--- a/siag/keys.go
+++ b/siag/keys.go
@@ -24,6 +24,7 @@ const (
 var (
 	ErrCorruptedKey       = errors.New("A corrupted key has been presented")
 	ErrInsecureAddress    = errors.New("An address needs at least one required key to be secure")
+	ErrOverwrite          = errors.New("Keys already exist with that keyname.")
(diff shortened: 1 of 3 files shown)


siag/keys.go
```diff
@@ -88,7 +89,15 @@ func generateKeys(requiredKeys int, totalKeys int, folder string, keyname string
...
 	for i, key := range keys {
-		err := encoding.WriteFile(filepath.Join(folder, keyname+"_Key"+strconv.Itoa(i)+FileExtension), key)
+		keyFilename := filepath.Join(folder, keyname+"_Key"+strconv.Itoa(i)+FileExtension)
+		_, err := os.Stat(keyFilename)
+		if !os.IsNotExist(err) {
+			if err != nil {
+				return types.UnlockConditions{}, err
+			}
+			return types.UnlockConditions{}, ErrOverwrite
+		}
```
lukechampine: this should be sufficient:
```go
// return an error if the file already exists
if _, err := os.Stat(keyFilename); err == nil {
    return types.UnlockConditions{}, ErrOverwrite
}
```

PR #451, opened 2015-04-21
switch from cbc to gcm

DavidVorick:
I also changed the way nonces and padding works. Padding is taken care of automatically by the AEAD interface from what I can tell. The nonce now gets prepended to the ciphertext. To create type safety, I made a new type 'Ciphertext'. (Since a ciphertext is not just a byte slice, it's a byte slice with a specific format, IE a nonce prepended at the beginning).

I was going to include functions for readers and writers but I actually wanted to work more on the renter first before doing that. One concern is that if we're encrypting a 1TB file, we have to find someway to encrypt it and then upload it. I'm not sure what the best design there is but I want to make sure we're not using the disk too much.

### DIFF
--- a/crypto/encrypt.go
+++ b/crypto/encrypt.go
@@ -1,5 +1,8 @@
 package crypto
 
+// encrypt.go contains functions for encrypting and decrypting data byte slices
+// and readers.
+
 import (
 	"crypto/cipher"
 	"crypto/rand"
@@ -8,16 +11,13 @@ import (
 	"golang.org/x/crypto/twofish"
 )
 
+var (
+	ErrInsufficientLen = errors.New("supplied ciphertext is not long enough to contain a nonce")
+)
+
 type (
 	TwofishKey [32]byte
 )
-
-// Encryption and decryption of []bytes is supported, but requires keeping some
-// overhead information. Also, nothing is done in-place, which means the
-// functions are more memory intensive, but that's not generally the bottleneck
-// when doing encryption. Developers are less likely to make mistakes when
-// inputs and outputs are different memory, which is why this method has been
-// chosen. Additionally, the overhead of using encryption is more transparent.
 
 // GenerateEncryptionKey produces a key that can be used for encrypting and
 // decrypting files.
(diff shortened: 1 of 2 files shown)


lukechampine:
I'm not too familiar with the cipher package, but it looks like we want a [Stream](https://golang.org/pkg/crypto/cipher/#Stream). Manually implementing a reader/writer using these primitives would be rough.

crypto/encrypt.go
```diff
@@ -26,67 +27,52 @@ func GenerateTwofishKey() (key TwofishKey, err error) {
...
-		return
+		return nil, err
 	}
-	encrypter := cipher.NewCBCEncrypter(blockCipher, iv)
-	encrypter.CryptBlocks(ciphertext, ciphertext)
 
-	return
+	// Encrypt the data. No authenticated data is provided, as EncryptBytes is
+	// meant for file encryption.
+	ct = append(nonce, aead.Seal(nil, nonce, plaintext, nil)...)
```
lukechampine: I'm not 100% sure, but I think it's okay to write `aead.Seal(nonce, nonce, plaintext, nil)`. You should test it to make sure.

DavidVorick: you are correct

DavidVorick:
Overhead is going to be less than 1% even if your segment size is 4kb. At 64kb it's under 0.1%, which is definitely small enough for me.

I just don't know how to approach the task of breaking the file apart into erasure coded pieces, encrypting each piece, and then uploading the pieces while doing it fast *and* not consuming a lot of memory + diskspace. It does seem like a solvable challenge though.

I'm just not sure when is the best time to approach such a challenge. Maybe for the time being we should do everything in memory and limit filesizes to 100MB or something. Then, once we have 20+ hosts we can introduce erasure coding and takle the problem then.

lukechampine:
I guess it would depend on the semantics of the erasure coding function. The ideal, for our purposes, would be a function that takes a Reader and returns a slice of Writers, each containing one segment. Then we could easily pass each Writer through an encryption pipe, and directly to the host from there. I don't think we're going to be so lucky though.

PR #452, opened 2015-04-22
add a compatibility package with tests for siag

DavidVorick:
A bunch of binary files are included. In total they're about 10kb. They will never be altered, they are binary files generated by siag 1.0. The approach here is to literally copy-paste the code from siag (just the verification code) and check it against the siag binaries.

In hindsight I guess a direct copy-paste wasn't necessary but the verification code is quite good, so I just reused it. It *is* necessary to have the binary files though.


compatibility/siag_1.0_test.go
```diff
@@ -0,0 +1,133 @@
...
+	"strconv"
+	"testing"
+
+	"github.com/NebulousLabs/Sia/crypto"
+	"github.com/NebulousLabs/Sia/encoding"
+	"github.com/NebulousLabs/Sia/types"
+)
+
+// KeyPairSiagOnePointZero matches the KeyPair struct of the siag 1.0 code.
+type KeyPairSiagOnePointZero struct {
```
lukechampine: I would suggest `KeyPairSiag_1_0` instead

DavidVorick: ok

lukechampine:
Copy-paste sort of defeats the purpose of the package, which is to ensure that the current code validates old binary data.

Maybe instead of a separate package, we should just require a `compatibility_test.go`+binary data in relevant packages.

PR #453, opened 2015-04-23
Contract revisions

DavidVorick:
This pull request exchanges contract terminations for contract revisions. After careful analysis of the situation, I realized that contract terminations weren't sufficient to make fully functional microdiff channels. I have replaced them with contract revisions, which are extremely close to being a superset of functionality. At the very least, you can make microdiff channels now, which are a superset of micropayment channels. Yay.

I also changed the 'Start' and 'Expiration' variable names in the FileContract type to 'WindowStart' and 'WindowEnd', because I felt that these names more accurately reflected their function.

### DIFF
--- a/types/filecontracts.go
+++ b/types/filecontracts.go
@@ -1,0 +1,107 @@
+package types
+
+// filecontracts.go contains the basic structs and helper functions for file
+// contracts.
+
+import (
+	"github.com/NebulousLabs/Sia/crypto"
+)
+
+type (
+	// A FileContract is a public record of a storage agreement between a "host"
+	// and a "renter." It mandates that a host must submit a storage proof to the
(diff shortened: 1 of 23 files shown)


types/filecontracts.go
```diff
@@ -0,0 +1,119 @@
...
+)
+
+type (
+	// A FileContract is a public record of a storage agreement between a "host"
+	// and a "renter." It mandates that a host must submit a storage proof to the
+	// network, proving that they still possess the file they have agreed to
+	// store.
+	//
+	// The party must submit the storage proof in a block that is between 'Start'
+	// and 'Expiration'. Upon submitting the proof, the outputs for
```
lukechampine: Start/Expiration -> WindowStart/WindowEnd?

PR #456, opened 2015-04-27
Siag Header + Version checking

DavidVorick:
while modifying the persistence stuff for the renter I realized that siag doesn't check the header or version number on the files that it opens. I added checks for that with test coverage.


siag/keys_test.go
```diff
@@ -149,3 +182,45 @@ func TestVerifyKeys(t *testing.T) {
 		}
 	}
 }
+
+// TestPrinkKeyInfo probes the printKeyInfo function.
```
lukechampine: typo

PR #457, opened 2015-04-28
Renter filehandling

DavidVorick:
This PR changes the way that the renter creates and stores files, and adds test coverage for files.go and persist.go

Functionally, not a whole lot has changed.

### DIFF
--- a/modules/renter/persist.go
+++ b/modules/renter/persist.go
@@ -1,41 +1,73 @@
 package renter
 
 import (
+	"encoding/json"
+	"errors"
+	"io/ioutil"
 	"path/filepath"
-
-	"github.com/NebulousLabs/Sia/encoding"
 )
 
-// savedFiles contains the list of all the files that have been saved by the
-// renter.
-type savedFiles struct {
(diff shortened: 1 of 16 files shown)


modules/renter/persist.go
```diff
@@ -1,44 +1,74 @@
...
+	if err != nil {
+		return err
+	}
+
+	if rp.Header != PersistHeader {
+		return ErrUnrecognizedHeader
+	}
+	if rp.Version != PersistVersion {
+		return ErrUnrecognizedVersion
+	}
```
lukechampine: fyi, this approach wouldn't work so well with the standard encoder. You would have to separately decode the header and version first before continuing. And everything will break anyway if we change encoding schemes. So maybe sticking with JSON is a good idea.

DavidVorick: Why won't this work so well with the standard encoder? Because the struct might look different?

DavidVorick: Compatibility is going to be a pain regardless, but json definitely gives much better flexibility.

Issue #446, opened 2015-04-19
Notes from local test (gateway)


DavidVorick:
Steps:

**1:** Start mining on :9971

**unwanted result**: :9971 added the linode as a peer, even though :9971 is running dev constants and the linode is running release constants. This is true for all other nodes I ran too.

**2:** Add :9981 as a peer to 9971

**unwanted result**: 9981 is not catching up, despite being sent blocks

:9981 log file has a lot of entries like this:
```
2015/04/19 13:31:45.274430 rpc.go:123: INFO: handling RPC "AcceptBl" from 127.0.0.1:44647
2015/04/19 13:31:45.274487 rpc.go:125: WARN: incoming RPC "AcceptBl" failed: block has no known parent
2015/04/19 13:31:45.274497 synchronize.go:42: INFO: synchronizing to 127.0.0.1:44647
2015/04/19 13:31:45.274597 synchronize.go:47: ERR: synchronization to 127.0.0.1:44647 failed: dial tcp 127.0.0.1:44647: connection refused
```

**3:** Add :9971 as a peer to :9991 

**unwanted result**: :9991 is not synchronizing. I had to call 'synchronize' manually. After manually calling synchronize, things seem to work.

**unwanted result**: :9971 does not have :9991 as a peer. Perhaps being on localhost is corrupting things, my ports aren't forwarded so any attempt to contact the actual IP address won't work. (it might be nice to have :9971, :9981, and :9991 all forwarded to 10.0.0.11)

**4:** Add :9971 as a peer to :9981. Set :9971 to :9981's only peer. The 2 minute re-sychronize loop should synchronize them eventually.

**unwanted result**: synchronization does not happen, even after waiting 10 minutes. Resynchronization loop is broken

**5:** Call synchronzie manually for everyone. Everything catches up. As :9971 has :9981 as a peer, :9981 remains caught up. :9991 falls behind immediately. Does not catch up.

The resynchronize loop is broken.

**6:** Close and restart :9981. Reload. :9981 has :9971 as a peer. :9971 also has :9981 as a peer. :9981's previous blocks have all loaded.

**unwanted result**: upon loading, the gateway does not call 'synchronize'.  :9981 is stuck at a lower block, despite being sent new blocks.

=========

Other notes:

calling 'synchronize' manually didn't always work the first time when the linode was also a peer. Sometimes I had to call it a few times to get the synchronization to happen.

[...]

Issue #441, opened 2015-04-15
Peer connections


lukechampine:
Bitcoin only allows 8 outgoing connections and 125 incoming connections (though the latter is configurable). Pieter Wuille says:

>**Please don't change this, as there is no need.** Connectable peers on the network are a scarce resource, and essential to the decentralization. If people go try connect to all of them like some sites do, we'll very quickly run out.

It also appears (though I'm not 100% sure) that connections are kept alive indefinitely. That is, instead of redialing for each connection, you just maintain a pool of 8 open connections and send/receive on them as needed. This is not really feasible if you're connecting to lots of peers, like we are, but it's okay for just 8.
I think I understand the real advantage of this, which is that it (partially) *solves the NAT problem*. If you're behind NAT, peer's can't dial you. But if you dial a peer, and you keep that connection alive, they can continue to use the connection to talk to you.

My opinion is that we should transition to this model, preferably sooner rather than later. I have a feeling it will improve our connection reliability as well. ~~Note that this will require us to implement "keepalive" messages, which are sent every 30 minutes or so to prevent the connection from being automatically closed.~~ (muxado handles this internally)

lukechampine:
David pointed out that hosts still cannot be behind NAT, because clients won't necessarily maintain active connections to them. So we either need a way for clients to tell a host that they want to connect (e.g. by passing the message through one of the host's peers), implement UPnP, or simply require that hosts not be behind NAT. I think the latter is preferable in the short term. Also, note that while the network remains very small, there is a good chance that you *will* in fact have your host as a peer.

lukechampine:
Useful links:

Bitcoin network spec: https://en.bitcoin.it/wiki/Network

How Bitcoin discovers peers: https://en.bitcoin.it/wiki/Satoshi_Client_Node_Discovery

How Bitcoin chooses peers: https://en.bitcoin.it/wiki/Satoshi_Client_Node_Connectivity

How Bitcoin exchanges blocks: https://en.bitcoin.it/wiki/Satoshi_Client_Block_Exchange

PR #458, opened 2015-04-28
Persistent peer connections

lukechampine:
Implements the scheme described in #441, and hopefully resolves many of the issues raised in #446. As a consequence, peer communication (peer discovery, synchronization) has been completely separated from hosting communication (negotiating, uploading+downloading). Note that this means an additional "host port" is now required.

This turned out to be more involved than I thought. Sorry about the large size. The good news is that connections should be much more stable now, and our NAT-traversal ability has improved significantly. It may not even be necessary for everyone on the testnet to forward their ports, though I'd still recommend it.

There's one "blemish" on this PR, which is that the network communication employed by the host and renter is not very "consistent;" it is redefined in multiple places, and is largely cut-and-paste from the gateway. It would be much cleaner to just reuse the gateway code, but ultimately I don't think this is the best approach. The semantics of host/renter network activity is very different from the activity between persistent connected peers, and so we'll inevitably need host/renter-specific netcode to account for that difference. We may even separate them further, e.g. when the renter is downloading it should tolerate longer timeouts, etc.
Still, it would be straightforward enough to consolidate the host/renter network protocol in one place rather than two. I can include that in this PR if you think that makes sense.


DavidVorick:
Build failed.

No need to worry about how the renter and host will do things right now. They need a lot of changes to have correctness.

For the next week, the big concern is that everything works when the user tries it out. I will be hacking encryption into the renter but other than that we're only trying to pass high-level testing on renting and hosting.

Makefile
```diff
@@ -18,6 +18,7 @@ dependencies:
 	go get -u golang.org/x/tools/cmd/cover
 	go get -u github.com/NebulousLabs/merkletree
 	go get -u github.com/boltdb/bolt
+	go get -u github.com/inconshreveable/muxado
```
DavidVorick: what is muxado and why do we need it?

api/api.go
```diff
@@ -41,7 +42,6 @@ func (srv *Server) initAPI(addr string) {
 
 	// Gateway API Calls
 	handleHTTPRequest(mux, "/gateway/status", srv.gatewayStatusHandler)
-	handleHTTPRequest(mux, "/gateway/synchronize", srv.gatewaySynchronizeHandler)
```
DavidVorick: You need to update API.md to reflect the changes you made to the api.

api/consensus.go
```diff
@@ -6,17 +6,30 @@ import (
...
-		Target       types.Target
-	}{
+	writeJSON(w, ConsensusInfo{
 		srv.cs.Height(),
-		currentBlock,
-		target,
+		srv.cs.CurrentBlock().ID(),
+		srv.cs.CurrentTarget(),
 	})
 }
```
DavidVorick: You'd rather not define it inline? It doesn't get used outside of this function.

api/gateway_test.go
```diff
@@ -134,3 +142,4 @@ func TestBlockBootstrap(t *testing.T) {
 		time.Sleep(time.Millisecond)
 	}
 }
+*/
```
DavidVorick: Is this code commented out because something is incomplete, or is it no longer necessary?

modules/consensus/state.go
```diff
@@ -128,6 +139,10 @@ func New(saveDir string) (*State, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	// spawn resynchronize loop
+	go s.threadedResynchronize()
```
DavidVorick: Are there tests which make sure that the resynchronize loop is working correctly? I know it wasn't before this PR.

Makefile
lukechampine: muxado is a stream multiplexing library. It allows us to layer multiple RPCs over the same persistent connection. It provides nice abstractions that make this feel very similar to the old one-conn-per-RPC scheme.

api/consensus.go
lukechampine: it wasn't before, but now it's used by [siac/consensuscmd.go](https://github.com/NebulousLabs/Sia/blob/keepalive/siac/consensuscmd.go#L37)

api/gateway_test.go
lukechampine: hmm. I think I could actually uncomment it now, although it's really more of a consensus test than a gateway test.

modules/consensus/state.go
lukechampine: I don't think so. We should definitely write one.

api/api.go
lukechampine: good point

modules/consensus/synchronize.go
```diff
@@ -45,121 +42,121 @@ func (g *Gateway) threadedResynchronize() {
...
 //
 // TODO: don't run two Synchronize threads at the same time
-func (g *Gateway) Synchronize(peer modules.NetAddress) error {
-	g.log.Println("INFO: synchronizing to", peer)
-	for {
+func (s *State) Synchronize(peer modules.NetAddress) error {
+	// loop until there are no more blocks available
+	moreAvailable := true
+
+	for moreAvailable {
```
DavidVorick: no need for whitespace here (sorry for being pedantic)

modules/consensus/synchronize.go
```diff
@@ -45,121 +42,121 @@ func (g *Gateway) threadedResynchronize() {
...
 		if err != nil {
-			g.log.Printf("ERR: synchronization to %v failed: %v\n", peer, err)
 			return err
 		}
-		g.log.Printf("INFO: %v sent us %v blocks\n", peer, len(newBlocks))
+
+		// integrate received blocks
 		for _, block := range newBlocks {
-			acceptErr := g.state.AcceptBlock(block)
+			// TODO: don't Broadcast these blocks
```
DavidVorick: should we not?

I think instead we should just consider making broadcast more powerful. For example, what if the consensus package knew the weight for all of its peers? Then it would only send them a block if it knew it had a higher weight than they did.

Or sending a block could be multi-step. "do you have this hash?" -> then send block.

This might be the wrong place to add complexity.

modules/consensus/synchronize.go
```diff
@@ -45,121 +42,121 @@ func (g *Gateway) threadedResynchronize() {
...
 	return nil
 }
 
-// sendBlocks returns a sequential set of blocks based on the 32 input block
+// SendBlocks returns a sequential set of blocks based on the 32 input block
 // IDs. The most recent known ID is used as the starting point, and up to
 // 'MaxCatchUpBlocks' from that BlockHeight onwards are returned. It also
 // sends a boolean indicating whether more blocks are available.
-func (g *Gateway) sendBlocks(conn modules.NetConn) (err error) {
+func (s *State) SendBlocks(conn modules.PeerConn) error {
```
DavidVorick: why is this exported? I thought modules registered their own rpcs?

modules/consensus/synchronize.go
lukechampine: we may eventually switch to the bitcoin model, which is to just advertise hashes and leave it up to your peers whether they want to download the full block from you (ditto for txns).

In general though, I think it's a good idea for consensus to know more about the states of its peers.

modules/consensus/synchronize.go
lukechampine: hmm. I guess the state could register its RPCs in `New`. It's convenient to have them all registered in the same place, but the gateway breaks that convention anyway...

modules/consensus/synchronize.go
lukechampine: how about `for moreAvailable := true; moreAvailable; {`?

modules/consensus/synchronize.go
lukechampine: I think this would also solve another problem I encountered.

We would like to synchronize to new peers as soon as we connect to them. But since the gateway isn't in charge of synchronizing any more, there isn't a good way to do it.

But if you had the consensus regularly updating metadata about the peer set, it could keep track of whether we had synchronized to a peer or not, and act accordingly.

modules/consensus/synchronize.go
DavidVorick: not my preferred style

2 statement for loops are legal? I didn't even know that

modules/consensus/synchronize.go
DavidVorick: hmm. Perhaps the synchronize loop could request the latest hashes from everybody instead of just picking one person to randomly synchronize to. That seems more robust anyway.

modules/consensus/synchronize.go
DavidVorick: My vote is that each of the modules worries about their own rpcs. Most of the time a module's rpcs are only going to be called by other instances of that module on the network (except for the host-renter relationship)

modules/consensus/synchronize.go
lukechampine: yeah I don't like the look of it either.
it's actually the normal 3-statement form, with the last statement empty.

modules/gateway/conn.go
```diff
@@ -1,96 +1,18 @@
...
-
-// dial wraps the connection returned by net.Dial in a conn.
-func dial(addr modules.NetAddress) (*conn, error) {
-	nc, err := net.DialTimeout("tcp", string(addr), dialTimeout)
-	if err != nil {
-		return nil, err
-	}
-	return newConn(nc), nil
+// peerConn is a simple type that implements the modules.PeerConn interface.
+// Eventually it may monitor the connection to thwart malicious behavior.
```
DavidVorick: I'm not sure that the second sentence belongs in the docstring.

modules/gateway/conn.go
```diff
@@ -1,96 +1,18 @@
...
 }
 
-// accept wraps the connection return by net.Accept in a conn.
-func accept(l net.Listener) (*conn, error) {
-	nc, err := l.Accept()
-	if err != nil {
-		return nil, err
-	}
-	return newConn(nc), nil
+func (pc *peerConn) CallbackAddr() modules.NetAddress {
```
DavidVorick: needs a docstring

modules/gateway/gateway.go
```diff
@@ -51,94 +53,53 @@ type Gateway struct {
 
 // Address returns the NetAddress of the Gateway.
 func (g *Gateway) Address() modules.NetAddress {
-	counter := g.mu.RLock()
-	defer g.mu.RUnlock(counter)
+	id := g.mu.RLock()
+	defer g.mu.RUnlock(id)
```
DavidVorick: at some point I decided id was too vague. It's not a big deal but I've started switching to lockID instead. I'll probably go through the whole repo and change all the names to lockID eventually.

modules/gateway/gateway_test.go
```diff
@@ -1,106 +1,142 @@
...
-	badpeer.RegisterRPC("Ping", func(conn modules.NetConn) error {
-		// write a length prefix, but no actual data
-		conn.Write([]byte{1, 0, 0, 0, 0, 0, 0, 0})
-		select {}
-	})
-
-	err := g.RPC(badpeer.Address(), "Ping", readerRPC([1]byte{}, 1))
-	if err != ErrTimeout {
-		t.Fatalf("Got wrong error: expected %v, got %v", ErrTimeout, err)
+func TestLoad(t *testing.T) {
```
DavidVorick: This should probably be moved to persist_test.go

DavidVorick:
Every file should probably have an accompanying '_test' file that does unit testing on each function.

modules/gateway/nodes.go
```diff
@@ -0,0 +1,70 @@
...
+
+// addNode adds an address to the set of nodes on the network.
+func (g *Gateway) addNode(addr modules.NetAddress) error {
+	if _, exists := g.nodes[addr]; exists {
+		return errors.New("node already added")
+	}
+	g.nodes[addr] = struct{}{}
+	g.log.Println("INFO: added node", addr)
+	return nil
+}
```
DavidVorick: should addNode also request the peerlist from the new node?

modules/host/net.go
```diff
@@ -0,0 +1,49 @@
+package host
+
+import (
+	"net"
+
+	"github.com/NebulousLabs/Sia/encoding"
+)
+
+type rpcID [8]byte
```
DavidVorick: does it make sense to use the types package 'Identifier' type? Maybe it doesn't, just a thought.

DavidVorick: It is interesting that the host would have a net.go. hmm

modules/gateway/gateway.go
lukechampine: yeah, this caused a naming conflict for me once, where I was already using `id` as an `rpcID`.

modules/host/net.go
lukechampine: It will become more necessary as the renter/host network semantics become more complex (e.g. monitored connections)

modules/gateway/nodes.go
lukechampine: perhaps, though it would make more sense to do it in `Connect` rather than `addNode`.
currently the only time we request nodes is in `Bootstrap`. Eventually we should add ongoing advertisements of our own address, e.g. every 24 hours.

PR #461, opened 2015-04-29
Renter api

DavidVorick:
Add some calls to the renter api: delete, rename, and share (load and save).

delete and rename are self explanatory

share/save creates a file that you can share with others, then they can download stuff that you uploaded

share/load loads a '.sia' that someone shared with you, now you can download stuff they uploaded.


modules/renter/persist.go
```diff
@@ -72,3 +88,72 @@ func (r *Renter) load() error {
...
+	defer r.mu.RUnlock(lockID)
+
+	if len(nicknames) == 0 {
+		return ErrNoNicknames
+	}
+	// Suffix enforcement is not really necessary, but I want explicit
+	// enforcement of the suffix until people are used to seeing '.sia' files.
+	if !strings.HasSuffix(sharedest, ShareExtension) {
+		return ErrNonShareSuffix
+	}
```
lukechampine: `filepath.Ext` might be clearer here, and you're already including `filepath` anyway

Issue #459, opened 2015-04-29
encoding.Read/WriteFile do not close the file


lukechampine:

lukechampine:
fixed in 366cb3cff41fea0b8d092

PR #462, opened 2015-04-29
Random Stuff

DavidVorick:
I overhauled the siad package and added some testing.

### DIFF
--- a/siad/main_test.go
+++ b/siad/main_test.go
@@ -3,7 +3,6 @@ package main
 import (
 	"net/http"
 	"os"
-	"path/filepath"
 	"testing"
 
 	"github.com/NebulousLabs/Sia/modules/tester"
@@ -28,9 +27,12 @@ func TestMain(t *testing.T) {
 		"-H",
 		"localhost:45152",
 		"-d",
-		filepath.Join(testDir, "Naive Run"),
+		testDir,
 	}
 	go main()
+
+	// Wait until the daemon has started and then send a kill signal to the
+	// daemon.
 	<-started
 	resp, err := http.Get("http://localhost:45150/daemon/stop")
 	if err != nil {
--- a/siad/main_test.go
+++ b/siad/main_test.go
@@ -1,0 +1,50 @@
+package main
+
+import (
+	"net/http"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/NebulousLabs/Sia/modules/tester"
+)
+
+// TestMain tries running the main executable using a few different commands.
+func TestMain(t *testing.T) {
+	if testing.Short() {
+		t.SkipNow()
+	}
+
+	testDir := tester.TempDir("siad", "TestMain")
+
+	// Try running and closing an instance of siad.
+	os.Args = []string{
+		"siad",
+		"-n",
+		"-a",
+		"localhost:45150",
+		"-r",
+		"localhost:45151",
+		"-H",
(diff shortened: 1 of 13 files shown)


DavidVorick:
Fixes #459

siad/main_test.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	testDir := tester.TempDir("siad", "TestMain")
+
+	// Try running and closing an instance of siad.
+	os.Args = []string{
+		"siad",
+		"-n",
+		"-a",
+		"localhost:45150",
```
lukechampine: just use the port.

lukechampine:
I don't agree with making `startDaemonCmd` a cobra command. It makes it harder to test, since it no longer returns an error that can be checked. Also, the use of a channel strikes me as ugly (and to be extra nit-picky, you send down the channel before you actually start serving requests).

My suggestion would be to have a cobra command that calls `createDaemon` and then starts the API server separately.

DavidVorick:
It would just be a passthrough function, nothing more.

Sending down the channel has to happed before the server starts listening, because Serve() is a blocking function. I guess I could do `go startChan <- struct{}{}`

DavidVorick:
My workflow got messed up and 2 extra commits squeezed through.

Some changes to the transaction pool to prevent transactions from being added to every block, and then a brief comment change + makefile change.

lukechampine:
no, putting it in a go func doesn't help anything.
here's what I suggested:
```go
func startDaemonCmd(*cobra.Command, []string) {
    d, err := createDaemon()
    if err != nil {
        fmt.Println(err)
        return
    }
    startChan <- struct{}{}
    err := d.Serve()
    if err != nil {
        fmt.Println(err)
    }
}
```

really, `createDaemon` should take arguments instead of using global variables. Then you could test it without needing to pass flags to `main()`.

siad/main_test.go
DavidVorick: ?

DavidVorick:
=/

hmph.

main() is my least favorite type of code

lukechampine:
mine too :/

DavidVorick:
putting it in a gofunc helps because the gofunc isn't going to trigger until later, so the code will at least get one step farther?

I guess Serve() is a block box.

lukechampine:
the "right" way would be:
```go
go Serve()
startChan <- struct{}{}
select {} // or better, wait on a channel signalled by Serve()
```

DavidVorick:
that's worse than sending down the channel in the go func, because you're almost guaranteed to have the `startChan <- struct{}{}` code run before the `Serve()` code.

The best way would be to push something down the channel inside of `Serve()`, but that's too much effort. The way it's happening now is working well enough.

DavidVorick:
sorry, can't pull Serve() out of the `createDaemon` function. It breaks the flow and doesn't make sense to me.

lukechampine:
what does that mean

DavidVorick:
uhh, you'll see. I did my best but it's still not great.

DavidVorick:
I have found the quest to give this more functional purity greatly frustrating and damaging to the code flow. I don't think that cobra was designed to operate in a pure functional environment (obviously, as it relies on global variables) and neither was gcfg (which relies on nested structs).

Trying to make things work this way has resulted in a lot of ugly modifications to the code and I don't want to keep the changes.

PR #463, opened 2015-05-01
Debugging

DavidVorick:
1. Mutexes all now lock for the same amount of time, as specified by a module constant. In testing mode, they wait 3 seconds. In dev mode, they wait 5 seconds. In release mode, they wait 8 seconds. This should stop the mutexes from complaining when there are performance issues.
Performance issues will probably be addressed when I rehash the consensus testing.

2. If the host runs into problems during a negotiation, it'll complain to the renter so that the renter won't report a successful file.

NOTE: A host and a renter must be in the same peer graph for renting to work. (The renter broadcasts a transaction that must be seen by the host or the negotiation will fail. The host won't see it if there's no chain of broadcasts that lead from the renter to the host. This is the type of thing we will fix when we redo the negotiation)

### DIFF
--- a/modules/consensus/state.go
+++ b/modules/consensus/state.go
@@ -4,7 +4,6 @@ import (
 	"errors"
 	"os"
 	"testing"
-	"time"
 
 	"github.com/NebulousLabs/Sia/blockdb"
 	"github.com/NebulousLabs/Sia/build"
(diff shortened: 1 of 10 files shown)


lukechampine:
if you're always going to call `sync.New` with the same argument, couldn't you remove the argument?

DavidVorick:
Yeah probably.

Originally I hadn't expected to do it that way, and it's fine to have some modules slower than others. (modules that use other modules need longer wait times, that's the only restriction). There are a few changes I think we could make to the sync package but I don't want to do anything big right now.

PR #464, opened 2015-05-01
More node sharing

lukechampine:
Nodes are now requested whenever you connect to a peer, and that peer will also broadcast your address to its peers when you connect to it.
This should mean that at any given point, your node list will be the union of the node lists of all of your peers. Ideally, this implies that every node knows about every other node, and that when a node joins the network, its address is quickly relayed to the entire network. Whether or not these ideals will hold in reality remains to be seen.

### DIFF
--- a/modules/gateway/rpc.go
+++ b/modules/gateway/rpc.go
@@ -72,13 +72,13 @@ func (g *Gateway) listenPeer(p *peer) {
 	for {
 		conn, err := p.accept()
 		if err != nil {
+			g.log.Println("WARN: lost connection to peer", p.addr)
 			break
 		}
 
 		// it is the handler's responsibility to close the connection
 		go g.threadedHandleConn(conn)
 	}
-	g.log.Println("WARN: lost connection to peer", p.addr)
 	g.Disconnect(p.addr)
 }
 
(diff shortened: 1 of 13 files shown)


DavidVorick:
This should be fine for the beta.

Perhaps a node can be limited to sharing 200 peers or something, but we can worry about that later. I don't see it being a problem in the coming weeks.

modules/gateway/rpc.go
```diff
@@ -50,6 +52,7 @@ func (g *Gateway) RPC(addr modules.NetAddress, name string, fn modules.RPCFunc)
 	// call fn
 	err = fn(conn)
 	if err != nil {
+		// TODO: log error here?
 		// give peer a strike
 		atomic.AddUint32(&peer.strikes, 1)
 	}
```
DavidVorick: Wait I thought there were no strikes?

DavidVorick:
Looks good to me. I guess I have one final feature request.

I would really like broadcast to try and send information to each node multiple times if the send fails. This is just one of those things that I think is going to save a few times. I know the transaction pool relies on it currently, but I also see it being useful in other places, such as when blocks are found abnormally close together.

PR #465, opened 2015-05-01
Sort of Magnet Links

DavidVorick:
You can get ascii '.sia' files now, they're about 2000 characters per file. Eventually we can probably optimize this down to a few hundred or even just a single hash, but for the next few months we're stuck with longer strings.

I also cleaned up a lot of the api calls, but left the old ones (marked DEPRECATED) so that we don't break the UI.


modules/renter/persist.go
```diff
@@ -112,15 +111,49 @@ func (r *Renter) ShareFiles(nicknames []string, sharedest string) error {
...
+	if err != nil {
+		return nil, err
+	}
+	_, err = zip.Write(shareBytes)
+	if err != nil {
+		return nil, err
+	}
+	zip.Close()
+
+	return zipBuffer.Bytes(), nil
```
lukechampine: since you wind up writing the zip data to a file, you should just pass the file to `shareBytes` as an `io.Writer` and create the `gzip.NewWriterLevel` with that.

lukechampine: nvm I did it

API.md
```diff
@@ -16,14 +16,47 @@ There may be functional API calls which are not documented. These are not
...
 Daemon
 ------
 
 Queries:
 
 * /daemon/stop
-* /daemon/update/apply
-* /daemon/update/check
+* /daemon/updates/apply
+* /daemon/updates/check
```
lukechampine: breaking the API for no reason???

API.md
```diff
@@ -59,47 +92,14 @@ struct {
...
 Gateway
 -------
 
 Queries:
 
 * /gateway/status
-* /gateway/peer/add
-* /gateway/peer/remove
+* /gateway/peers/add
+* /gateway/peers/remove
```
lukechampine: ???

API.md
```diff
@@ -378,20 +393,32 @@ typically best not to shut down siad until files are no longer being repaired.
...
 
-#### /renter/file/rename
+Function: Load a '.sia' into the renter.
+
+Parameters:
+```
+file string
+```
+`file` is the ascii representation of the '.sia' file being loaded into the
+renter.
```
lukechampine: URL length limit restrictions may be an issue here. We may need to use a POST instead of a GET.

lukechampine:
I don't agree with all of the API changes. Seems like completely needless renaming. Why is "updates" better than "update"? Why is "peers" better than "peer"?

modules/renter/persist.go
DavidVorick: sweet thanks

DavidVorick:
There was inconsistency. I naturally tend to write about things in the plural context, 'peers' seemed a lot more natural. There were plenty of times manually typing api calls that I wrote '/renter/files/' instead of 'renter/file', and 'gateway/peers' instead of 'gateway/peer'.

Most importantly, I just wanted everything to be consistent.

There are more places in the codebase that things are inconsistent, I'm slowly trying to clean all of that up.

lukechampine:
"updates" doesn't make sense.

it's nice that *you* write things that way, but *I* think they're more natural as-is. "apply the update." "add the peer." etc.
personal preference isn't worth deprecating API calls over.

DavidVorick:
but you're not applying an update. You are performing one of a class of actions that relates to updates. 

"from the daemon, relating to updates, apply an update"

"from the renter, relating to files, share a set of files"

lukechampine:
it doesn't make sense to talk about updates in plural. there are no API calls that operate on a plurality of updates. `/peers/` would make sense if `/gateway/status` were `/peers/list`, but it isn't.
there is no benefit to making `update` plural.

lukechampine:
yeah, it makes sense to make things plural...when you're talking about things that are plural. Payments are obviously plural. But [vault](https://developer.paypal.com/docs/api/#vault) is obviously not.

API.md
```diff
@@ -297,29 +297,16 @@ Renter
...
-Response: standard
+* /renter/files/delete
+* /renter/files/download
+* /renter/files/list
+* /renter/files/load
+* /renter/files/loadascii
+* /renter/files/rename
+* /renter/files/share
+* /renter/files/shareascii
+* /renter/files/upload
```
lukechampine: it doesn't make sense to group actions on all files (`list`) with actions on single files.

DavidVorick:
it should be `/gateway/peers/list` instead of `/gateway/status`. For the renter, the proper call is `/renter/files/list`, I thought I changed it for the gateway too.

Updates are plural, there could be multiple that you missed and there are definitely multiple updates throughout the lifetime of the program. It's also possible that we'd get calls like `/daemon/updates/changelog?backtrack=3` => provide the changelog for the previous 3 updates.

Stuff that makes sense to be singular is the renter, the gateway, etc.

lukechampine:
>It's also possible that

I think you are too focused on trying to solve the problems of tomorrow. Deprecating an API call has a real cost today. Making the API more "consistent" has an intangible payoff tomorrow.

I'm going to merge this because I don't feel like arguing about such trivialities any longer. It's clear to me that neither of us knows shit about good API design, and we're probably going to revise it a few more times anyway.

PR #466, opened 2015-05-02
Renter encryption

DavidVorick:
This builds on the magnet links pull request.

I added encryption to the renter. I also found the bug that was preventing the host from getting paid for storing files.

As far as my limited testing can tell, everything is good to go. This is the best Sia has ever worked to the extent of my knowledge, and it should not be confusing to use any more.


modules/host/update.go
```diff
@@ -60,17 +60,16 @@ func (h *Host) ReceiveConsensusSetUpdate(revertedBlocks []types.Block, appliedBl
 	lockID := h.mu.Lock()
 	defer h.mu.Unlock(lockID)
 
+	for _ = range revertedBlocks {
+		h.blockHeight--
+	}
```
lukechampine: maybe `h.blockHeight - len(revertedBlocks)`?

modules/renter/download.go
```diff
@@ -112,9 +113,18 @@ func (d *Download) start() {
 		for _, piece := range d.pieces {
 			downloadErr := d.downloadPiece(piece)
 			if downloadErr == nil {
-				d.complete = true
-				d.file.Close()
-				return
+				// Decrypt the file.
+				d.file.Seek(0, 0)
+				cryptBytes, err := ioutil.ReadAll(d.file)
```
lukechampine: storing the entire file in memory?

modules/renter/persist.go
```diff
@@ -184,6 +184,16 @@ func (r *Renter) loadSharedFile(reader io.Reader) error {
...
 	for i := range rsf.Files {
+		for {
+			_, exists := r.files[rsf.Files[i].Name]
+			if !exists {
+				break
+			}
+			if len(rsf.Files[i].Name) > 50 {
+				break
+			}
+			rsf.Files[i].Name += "_"
```
lukechampine: how about
```go
for i := range rsf.Files
    if _, exists := r.files[rsf.Files[i].Name]; !exists {
        continue
    }
    // if a file with this name already exists, add a duplicate number
    dupCount := 1
    for {
        name := rsf.Files[i].Name + strconv.Itoa(dupCount)
        _, exists := r.files[name]
        if !exists {
            rsf.Files[i].Name = name
            break
        }
        dupCount++
    }
}
```

modules/host/update.go
DavidVorick: I don't think there's anything else to do in the loop

modules/renter/download.go
DavidVorick: this is a short term solution. The longer term solution involves loading about 10MB at a time and leveraging micro-diffs to keep all of the memory requirements sane.

modules/renter/persist.go
DavidVorick: ok

modules/renter/download.go
lukechampine: yeah I figured, just checking

lukechampine:
Remember to update the .gitignore and Makefile

DavidVorick:
There aren't any tests for the changes, but I tested it manually and the file renaming works as expected (loading the same file repeatedly goes: file2 -> file2_1 -> file2_2 -> file2_3)

PR #468, opened 2015-05-04
Api testing

DavidVorick:
This PR is definitely half-baked. Since we're changing a lot of the requests to POST requests, I felt that it didn't make sense to continue working. The changes here are useful/organizational nonetheless.

### DIFF
--- a/api/server_test.go
+++ b/api/server_test.go
@@ -22,10 +22,25 @@ import (
 )
 
 var (
-	APIPort int = 9020
+	// CONTRIBUTE: The naive version of ":0" doesn't work because then the rest
+	// of the api tests attempt to call ":0" when making api requests. It would
+	// be prefereable to not need to use a port counter and rely on N ports
+	// being free starting from 25500 for testing to work.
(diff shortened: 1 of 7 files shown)


api/server_test.go
```diff
@@ -22,10 +22,25 @@ import (
 )
 
 var (
-	APIPort int = 9020
+	// CONTRIBUTE: The naive version of ":0" doesn't work because then the rest
+	// of the api tests attempt to call ":0" when making api requests. It would
+	// be prefereable to not need to use a port counter and rely on N ports
+	// being free starting from 25500 for testing to work.
+	APIPort int = 25500
```
lukechampine: yeah, there isn't a super clean way of doing this. The best way I can think of would be to manually create the HTTP listener instead of using `ListenAndServe` -- but then you need to add the listener to the Server struct so that you can call `Addr()` on it.

I think I've said before that I'd like to refactor our HTTP server code (maybe removing graceful), so I'll make the above change then.

PR #477, opened 2015-05-06
Release tuneup

DavidVorick:
This is a very hacky and low quality PR. I was aiming to change as little as possible.

+ Miner no longer hashes all of the transaction every time it changes the block nonces. That operation was slowing things down by a massive amount of time when there were any  transactions at all. Now, transactions more or less don't affect  the speed of mining.
+ If there is a mining error, the error is printed and all transactions in the transaction pool are discarded. Ideally, this will never happen, but it's good to have failsafes in place.
+  'TimeRemaining' returned for a file was reported incorrect values. Rather, it was only considering the first contract instead of all of them. Now it considers all contracts and returns the contract that has the most time left.
+ an error is returned if you try to share an inactive file.
+ when you load a shared file, println(name of file) occurs. This isn't very complete, it should be returned in the JSON. I hesistated b/c we're changing all of the GET and POST requests soon anyway, but I may return to this and make it a proper part of the api call.

I had to disable a test that no longer works. It doesn't work anymore because the test doesn't wait until a file is active before attempting to upload it. Now that this action is  no longer allowed, the test was failing. I will fix it later, I have tried manually and everything appears functional.

### DIFF
--- a/modules/miner/mine.go
+++ b/modules/miner/mine.go
@@ -106,6 +106,7 @@ func (m *Miner) solveBlock(blockForWork types.Block, target types.Target, iterat
 				m.tpool.PurgeTransactionPool()
 			}
 			solved = true
+			println("Found a block. If the block is not orphaned, you will receive the reward after 50 more blocks have been mined. Blocks are only orphaned when two miners find a block at the same time, as only 1 block can be accepted.")
 
 			// Grab a new address for the miner.
 			m.mu.Lock()
--- a/modules/miner/mine.go
+++ b/modules/miner/mine.go
@@ -1,9 +1,11 @@
 package miner
 
 import (
+	"bytes"
 	"math/rand" // We should probably switch to crypto/rand, but we should use benchmarks first.
 
 	"github.com/NebulousLabs/Sia/build"
+	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/types"
 )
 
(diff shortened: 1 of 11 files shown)


modules/miner/mine.go
```diff
@@ -94,18 +105,22 @@ func (m *Miner) solveBlock(blockForWork types.Block, target types.Target, iterat
...
-		if b.CheckTarget(target) {
+		if fastCheckTarget(target, b, bRoot) {
 			err = m.state.AcceptBlock(b)
-			if build.DEBUG {
-				if err != nil {
-					println(err.Error())
-				}
+			if err != nil {
+				print("mined a block, but with an error: ")
+				println(err.Error())
```
lukechampine: `println("Mined a bad block: " + err.Error())`

modules/miner/mine.go
```diff
@@ -94,18 +105,22 @@ func (m *Miner) solveBlock(blockForWork types.Block, target types.Target, iterat
...
+				m.tpool.PurgeTransactionPool()
 			}
 			solved = true
+			if build.Release != "testing" {
+				println("Found a block! If the block is not orphaned, you will receive the reward after 50 more blocks have been mined. Blocks are only orphaned when two miners find a block at the same time, as only 1 block can be accepted.")
+			}
```
lukechampine: how about just `println("Found a block! You will receive the reward in 50 blocks")`

PR #481, opened 2015-05-06
Remove logging from the api package

DavidVorick:
and add a new make target

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -3,7 +3,6 @@ package api
 import (
 	"encoding/json"
 	"fmt"
-	"log"
 	"net/http"
 	"strings"
 	"time"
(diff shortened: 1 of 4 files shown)


api/api.go
```diff
@@ -15,11 +14,10 @@ const (
 	apiTimeout = 5 * time.Second
 )
 
-// handleHTTPRequest is a wrapper function that logs and then handles all
-// incoming calls to the API.
+// handleHTTPRequest is a wrapper function that handles all incoming calls to
+// the API.
```
lukechampine: this wrapper is essentially useless now; it's a no-op

api/api.go
```diff
@@ -121,9 +119,8 @@ func (srv *Server) Serve() error {
 	return err
 }
 
-// writeError logs an writes an error to the API caller.
+// writeError an error to the API caller.
```
lukechampine: ditto

DavidVorick:
Josh and I tested this, it's ready

PR #482, opened 2015-05-07
Async load

lukechampine:
This should bring startup time down under a second.

### DIFF
--- a/modules/gateway/nodes_test.go
+++ b/modules/gateway/nodes_test.go
@@ -77,7 +77,7 @@ func TestRandomNode(t *testing.T) {
 	}
 	for node, count := range nodes {
 		if count == 0 { // 1-in-200000 chance of occuring naturally
-			t.Errorf("node %v was never selected", node, count)
+			t.Errorf("node %v was never selected", node)
 		}
 	}
 }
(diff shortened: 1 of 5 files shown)


modules/gateway/nodes_test.go
```diff
@@ -77,7 +77,7 @@ func TestRandomNode(t *testing.T) {
 	}
 	for node, count := range nodes {
 		if count == 0 { // 1-in-200000 chance of occuring naturally
```
DavidVorick: hmmm. Might want an extra 2-3 zeroes.

nbd I guess, but if you have 10 of these then it's down to 1-in-20k, and we run 'test' quite a few times. More than once per commit I imagine.

modules/host/announce.go
```diff
@@ -1,44 +1,69 @@
...
 	if err != nil {
-		return
+		return err
 	}
 
 	// create and encode the announcement and add it to the arbitrary data of
 	// the transaction.
+	lockID := h.mu.RLock()
 	announcement := encoding.Marshal(modules.HostAnnouncement{
 		IPAddress: h.myAddr,
```
DavidVorick: is there a check to make sure you actually got the correct external address?

lukechampine: oh right, derp

PR #488, opened 2015-05-16
Organization

DavidVorick:
I merged the tester package into the build package.

I reorganized the Makefile, it should look better now.

I renamed blockdb to persist. I intend to add a bunch of persistence helper functions to the persist package.

Bharath was quite convinced that we should use protocol buffers instead of JSON.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,16 +35,14 @@ install: fmt REBUILD
 	go install -tags='dev debug' ./...
 
 # release builds and installs release binaries.
-release: dependencies test-long REBUILD
-	go install -a ./...
-release-fast: REBUILD
+release: REBUILD
 	go install -a ./...
 
 # xc builds and packages release binaries for all systems by using goxc.
 # Cross Compile - makes binaries for windows, linux, and mac, 32 and 64 bit.
 xc: dependencies test test-long REBUILD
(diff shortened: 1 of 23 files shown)


Makefile
```diff
@@ -52,47 +50,23 @@ xc: dependencies test test-long REBUILD
...
+# 3 commands and a variable are available for testing Sia packages. 'pkgs'
+# indicates which packages should be tested, and defaults to all the packages
+# with test files. Using './...' as default breaks compatibility with the cover
+# command. 'test' runs short tests that should last no more than a few seconds,
+# 'test-long' runs more thorough tests which should not last more than a few
+# minutes.
+pkgs = ./api ./compatibility ./crypto ./encoding ./modules/consensus            \
+	./modules/gateway ./modules/host ./modules/hostdb ./modules/miner           \
+	./modules/renter ./modules/transactionpool ./modules/wallet ./siad ./siag   \
+	./types
```
lukechampine: you could use `find` to list these: `find . -type d | grep -v 'git\|release\|doc'`

unfortunately if you have a testing directory or other untracked directory in your Sia folder, it's gonna cause problems. So I guess we're stuck manually specifying them.

PR #490, opened 2015-05-17
100% encoding test coverage

lukechampine:
Feels good man. I also performed some manual mutation testing, and was unable to sneak any mutations by the tests. (Okay, there was one instance where I increased the capacity of a slice without changing its length -- but that's pretty inconsequential.)


encoding/marshal.go
```diff
@@ -144,6 +148,7 @@ func NewEncoder(w io.Writer) *Encoder {
 
 // Marshal returns the encoding of v. For encoding details, see the package
 // docstring.
+// TODO: merge with MarshalAll?
```
DavidVorick: Probably best not to, I think we can remove this.

DavidVorick:
Please add docstrings to all functions, even if it's mostly pointless.

DavidVorick:
Other than those two things, this looks good.

PR #492, opened 2015-05-18
jobs page

DavidVorick:
A bit of housekeeping, and a job openings page.

The two openings currently listed are UI/UX Engineer, and Decentralization Engineer.

I'm also thinking about adding an opening for a Marketing Noun. We could definitely use some help in the marketing department.


lukechampine:
What about nebulouslabs.com?

jobs.html
```diff
@@ -0,0 +1,99 @@
...
+							Nebulous is the parent company of Sia, and funds the majority of Sia's development.
+							Nebulous is based in Boston, and requires all employees to live within commuting distance from the office.
+							We are looking primarily for full time positions, however we are also accepting interships.
+							<br><br>
+							You can apply by sending an email to david@nebulouslabs.com
+							<br><br>
+							<b>UI/UX Engineer - High Priority</b>
+							<br>
+							We are currently looking for a UI/UX engineer and designer.
+							Competence with html, css, and javascript is required.
```
lukechampine: case

DavidVorick: you want it to be

HTML, CSS, and Javacript?

DavidVorick:
Put the jobs page on that site?

We could probably put one in each location.

PR #498, opened 2015-05-19
Node tweaks

lukechampine:


DavidVorick:
build failed

modules/gateway/persist.go
```diff
@@ -14,12 +14,25 @@ func (g *Gateway) save() error {
...
+	if err != nil {
+		return err
+	}
+	file, err := os.Create(filepath.Join(g.saveDir, "nodes.json"))
+	if err != nil {
+		return err
+	}
+	_, err = file.Write(data)
+	return err
 }
```
DavidVorick: This change is compatibility breaking. We should have some way to document all the places we break compatibility. I think a list could be useful when debugging user problems.

Maybe a changelog?

lukechampine: this particular change shouldn't break anything since I changed the filename.
But probably we should just rely on version numbers to catch breaking changes.

lukechampine:
Hmm. There's a tricky problem here.

We want to reject loopback node addresses, like 127.0.0.1 and [::1]. But these addresses are used during testing to avoid calling `getMyExternalIP`. Not sure how to tackle this.

lukechampine:
I fixed the problem by manually inserting nodes (bypassing `addNode`) in tests. There are still a few unfortunate consequences, but it's not as bad as I feared.

PR #506, opened 2015-05-20
stream encryption

lukechampine:
Files are no longer held in memory at any point during upload or download.

Currently, an OFB stream cipher is used to perform file encryption. This should be secure (if used properly), but I'm not totally confident in it. This is one of those areas where it'd be nice to have a professional cryptographer looking over our shoulder.

### DIFF
--- a/crypto/encrypt.go
+++ b/crypto/encrypt.go
@@ -28,14 +28,19 @@ func GenerateTwofishKey() (key TwofishKey, err error) {
 	return
 }
 
+// NewCipher creates a new Twofish cipher from the key.
+func (key TwofishKey) NewCipher() cipher.Block {
+	// NOTE: NewCipher only returns an error if len(key) != 16, 24, or 32.
+	cipher, _ := twofish.NewCipher(key[:])
+	return cipher
+}
+
 // EncryptBytes encrypts a []byte using the key. EncryptBytes uses GCM and
 // prepends the nonce (12 bytes) to the ciphertext.
 func (key TwofishKey) EncryptBytes(plaintext []byte) (Ciphertext, error) {
 	// Create the cipher.
-	// NOTE: NewCipher only returns an error if len(key) != 16, 24, or 32.
 	// NOTE: NewGCM only returns an error if twofishCipher.BlockSize != 16.
-	twofishCipher, _ := twofish.NewCipher(key[:])
-	aead, _ := cipher.NewGCM(twofishCipher)
+	aead, _ := cipher.NewGCM(key.NewCipher())
 
 	// Create the nonce.
 	nonce := make([]byte, aead.NonceSize())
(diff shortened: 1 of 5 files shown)


crypto/encrypt.go
```diff
@@ -29,40 +30,31 @@ func GenerateTwofishKey() (key TwofishKey, err error) {
...
-	aead, err := cipher.NewGCM(twofishCipher)
-	if err != nil {
-		return nil, err
-	}
+func (key TwofishKey) EncryptBytes(plaintext []byte) (Ciphertext, error) {
+	// Create the cipher.
+	// NOTE: NewCipher only returns an error if len(key) != 16, 24, or 32.
+	// NOTE: NewGCM only returns an error if twofishCipher.BlockSize != 16.
+	twofishCipher, _ := twofish.NewCipher(key[:])
+	aead, _ := cipher.NewGCM(twofishCipher)
```
DavidVorick: hmmm.

crypto/encrypt.go
```diff
@@ -29,40 +30,31 @@ func GenerateTwofishKey() (key TwofishKey, err error) {
...
-	twofishCipher, err := twofish.NewCipher(key[:])
-	if err != nil {
-		return nil, err
-	}
-	aead, err := cipher.NewGCM(twofishCipher)
-	if err != nil {
-		return nil, err
-	}
+	twofishCipher, _ := twofish.NewCipher(key[:])
+	aead, _ := cipher.NewGCM(twofishCipher)
```
DavidVorick: need to copy the comments from above into this function too, b/c it's a different function and people don't always read a file in order.

crypto/encrypt.go
```diff
@@ -70,9 +62,27 @@ func (key TwofishKey) DecryptBytes(ct Ciphertext) (plaintext []byte, err error)
...
+	block, _ := twofish.NewCipher(key[:])
+
+	// OK to use a zero IV if the key is unique for each ciphertext.
+	iv := make([]byte, twofish.BlockSize)
+	stream := cipher.NewOFB(block, iv) // TODO: is this the right mode to use?
+
+	return &cipher.StreamWriter{S: stream, W: w}
+}
+
+// NewReader returns a reader that decrypts or decrypts its input stream.
```
DavidVorick: decrypts or decrypts

crypto/encrypt.go
```diff
@@ -70,9 +62,27 @@ func (key TwofishKey) DecryptBytes(ct Ciphertext) (plaintext []byte, err error)
...
+func (key TwofishKey) NewReader(r io.Reader) io.Reader {
+	block, _ := twofish.NewCipher(key[:])
```
DavidVorick: also need comment here about not checking the err, same for NewWriter

crypto/encrypt.go
```diff
@@ -70,9 +62,27 @@ func (key TwofishKey) DecryptBytes(ct Ciphertext) (plaintext []byte, err error)
...
+
+	// OK to use a zero IV if the key is unique for each ciphertext.
+	iv := make([]byte, twofish.BlockSize)
+	stream := cipher.NewOFB(block, iv)
+
+	return &cipher.StreamReader{S: stream, R: r}
```
DavidVorick: These algorithms look correctly implemented to me.

DavidVorick:
looks like the code is much cleaner after this PR

Issue #494, opened 2015-05-18
Add getwork/submitblock API for external miners


cyrozap:
This was touched upon in #325, but I thought it might be good to make this a separate issue since it'll be easier to get an API for this working first than it'll be to implement an OpenCL miner.

DavidVorick:
I was thinking about this more, and I think these api calls make sense:

'/miner/blockforwork'
```
Block types.Block
MerkleRoot crypto.Hash
Target types.Target
```

'/miner/submitblock'
```
Block types.Block
```

In blockforwork, the parentid and the nonce can be found inside of the block. (Block.ParentID and Block.Nonce). Though the merkle root can be derived from the block, we're not sure if the miner has the right libraries to calculate the merkle root correctly, so we just give it as a freebie.

When a block is found, you just provide the winning block. All that's changed from the original block is the nonce.

---

One more requirement - every block returned by /miner/blockforwork must be unique. This can be easily achieved by adding a single transaction to the begginning that is empty except for the 'Arbitrary Data' field, which has a single random 16 byte string. Doing this means that the miner can request as many blocks as it wants and always start from nonce 0 without risking any overlapping computation.

DavidVorick:
In light of Forest and I getting confused at little endian, I've added an optional argument, 'Nonce', to submitblock. The nonce will be ignored if zeroed out.

lukechampine:
we should probably just change the nonce type in the block to a `[8]byte`. There's no real reason for it to be a `uint64`, other than that it makes it easy to increment.

Performing a little-endian conversion is not difficult though. In C it's annoying because I don't think there's a builtin function to do it, but it's just a google search away.

DavidVorick:
it's even more annoying when writing an fpga or an asic - you're basically manipulating logic gates

cyrozap:
FPGAs can actually do bit-order manipulation fairly easily, so I wouldn't worry about it. Either way, there will be a program in-between siad and the FPGA anyways, so as long as the FPGA gets something it can change, it doesn't matter whether it's BE or LE. The Python program I'm using to talk to my FPGA can handle the data as it's in this patch just fine. I don't think the `nonce` parameter in the `submitblock` API will be necessary since as we found out, my issues are with the FPGA generating non-deterministic hashes, not the nonce.

PR #507, opened 2015-05-20
Add initial getwork API

cyrozap:
This addresses the "getting work" half of issue #494.

### DIFF
--- a/modules/miner.go
+++ b/modules/miner.go
@@ -1,6 +1,7 @@
 package modules
 
 import (
+	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/types"
 )
 
@@ -12,6 +13,15 @@ type MinerInfo struct {
 	Threads        int
 	RunningThreads int
 	Address        types.UnlockHash
+}
+
+// MinerWork is all the information a miner needs to do work
+type MinerWork struct {
+	Block      types.Block
+	ParentID   types.BlockID
+	Nonce      uint64
+	MerkleRoot crypto.Hash
+	Target     types.Target
 }
 
 // The Miner interface provides access to mining features.
@@ -37,6 +47,10 @@ type Miner interface {
 	// It will give up after a few seconds, returning the block, a bool
 	// indicating whether it has been solved, and an error.
 	SolveBlock(types.Block, types.Target) (types.Block, bool)
+
+	// GetWork() returns a MinerWork struct which can be converted to JSON to be
+	// parsed by external miners
+	GetWork() MinerWork
 
 	// StartMining turns on the miner, which will endlessly work for new
 	// blocks.
--- a/modules/miner/getwork.go
+++ b/modules/miner/getwork.go
@@ -1,0 +1,24 @@
+package miner
+
+import (
+	"github.com/NebulousLabs/Sia/modules"
+)
+
+// GetWork() returns a MinerWork struct which can be converted to JSON to be
+// parsed by external miners
+func (m *Miner) GetWork() modules.MinerWork {
+	m.mu.Lock()
+	b := m.blockForWork()
+	target := m.target
+	m.mu.Unlock()
+
+	work := modules.MinerWork{
+		Block:      b,
+		ParentID:   b.ParentID,
+		Nonce:      b.Nonce,
+		MerkleRoot: b.MerkleRoot(),
+		Target:     target,
+	}
+
+	return work
+}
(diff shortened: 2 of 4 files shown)


modules/miner.go
```diff
@@ -14,6 +15,15 @@ type MinerInfo struct {
...
 }
 
+// MinerWork is all the information a miner needs to do work
+type MinerWork struct {
+	Block      types.Block
+	ParentID   types.BlockID
+	Nonce      uint64
+	MerkleRoot crypto.Hash
+	Target     types.Target
+}
```
DavidVorick: Block already has a ParentID and Nonce field, you could probably just fill those out in the block.

But I think really you should be sending a block header, which woud look like this:

```
    ParentID
    Nonce
    TransactionMerkleRoot
```

The ID is derived by hashing the three of those together, so you don't need any more information.

Then you'd return stuff to a `/postwork' api call that just has a nonce.

DavidVorick: There is a catch, the nonce is only 64 bits. This isn't really a problem while FPGA mining but when ASICs start using the api they will be close to using all 64 bits.

The fix to that is to have 'getwork' always return a unique Merkle root, which it achieves by putting a single arbitrary data transaction in the block and manipulating the string for the arbitrary data. Relatively easy to implement.

Then the asics just call 'getwork' again after using up all 64 bits and are guaranteed to have no overlap on their blocks.

modules/miner.go
```diff
@@ -38,6 +48,10 @@ type Miner interface {
 	// indicating whether it has been solved, and an error.
 	SolveBlock(types.Block, types.Target) (types.Block, bool)
 
+	// GetWork() returns a MinerWork struct which can be converted to JSON to be
+	// parsed by external miners
+	GetWork() MinerWork
```
DavidVorick: Hmm. I'm not sure this is the best name, but I can't really think of a better one.

This should be returning a block header that the miner can grind on.

modules/miner.go
cyrozap: Oh, right, sending the whole block was the hack I was using until I could figure out exactly what information needs to be sent back to the network. I'm aware that I only need those three parameters, but the host program I'm using to communicate with the FPGA needs to know the rest of the block information in order to submit it. As for the target, that needs to be sent to the FPGA so it can do the comparison internally--this saves it from having to send each hash over the wire to the host (very slow).

DavidVorick: oh yeah, you need the target.

but you don't need to assemble the block from the fpga, you can just write a 'submitheader' api call. I guess that could get complicated though if new transactions are coming in at a rapid rate.

modules/miner/getwork.go
```diff
@@ -0,0 +1,24 @@
...
+	"github.com/NebulousLabs/Sia/modules"
+)
+
+// GetWork() returns a MinerWork struct which can be converted to JSON to be
+// parsed by external miners
+func (m *Miner) GetWork() modules.MinerWork {
+	m.mu.Lock()
+	b := m.blockForWork()
+	target := m.target
+	m.mu.Unlock()
```
DavidVorick: unless there's a strongly compelling reason, you should have:

```
m.mu.Lock()
defer m.mu.Unlock()
```

at the top of the function. This makes it easier to catch mistakes with locks. For example, if m.blockForWork could return an error, you might forget to unlock before handling the error.  But with the defer, you don't have to remember stuff like that.

It's not a problem in this case but we stick to the conventions anyway.

Generally the only reason that you avoid `defer` on the unlock is if later on in the function you do some sort of network communication or intensive disk io. Everything else tends to be fast enough that it's not worth breaking convention.

DavidVorick:
I was thinking about this more, and I think these api calls make sense:

`/miner/blockforwork`
```
Block types.Block
MerkleRoot crypto.Hash
Target types.Target
```

`/miner/submitblock`
```
Block types.Block
```

In blockforwork, the parentid and the nonce can be found inside of the block. (Block.ParentID and Block.Nonce). Though the merkle root can be derived from the block, we're not sure if the miner has the right libraries to calculate the merkle root correctly, so we just give it as a freebie.

When a block is found, you just provide the winning block. All that's changed from the original input block is the nonce.

DavidVorick:
I have implemented a full solution in the api + miner. There are now two new calls: /miner/blockforwork, and /miner/submitblock.

Both are documented, and currently require being able to encode and decode json (a block). There is some debate about whether this is the best strategy (as we could just give out a string of bytes and document which bytes are the nonce), but for the time being it's what we're using.

PR #512, opened 2015-05-21
Consensus Changes

DavidVorick:
This was originally meant to add testing, but there's not much testing in this PR. I figured the number of changes that I have made are already significant enough to justify a PR.

The api package now uses the consensus set as a module instead of importing the state directly.

RelayBlock is not called when loading blocks from disk or during synchronization.

The synchronization loop is a lot smoother, and is no longer random. Instead, it iterates through all peers and calls 'synchronize' on each of them, taking 20s breaks in between each peer. If a synchronize call fails, the 20s break is skipped.

Then another 20s break occurs before refreshing the peer list and starting over.

I also added a function to test a batch of transactions. I'm not sure if this function works as intended, ti doesn't have testing.


DavidVorick:
hmm that's not an error I can reproduce. There's a pretty simple fix but I wonder why there's a disparity between my machines and travis.

lukechampine:
weird, me neither.

DavidVorick:
oh, try `go install -a`

It's because we already have the package on our machines, go isn't trying to rebuild it for us.

modules/consensus.go
```diff
@@ -49,3 +60,37 @@ type SiafundPoolDiff struct {
...
+	// will be sent to the module via the 'ReceiveConsensusSetUpdate' function.
+	// This is a thread-safe way of managing updates.
+	ConsensusSetSubscribe(ConsensusSetSubscriber)
+
+	// Synchronize will try to synchronize to a specific peer. During general
+	// use, this call should never be necessary.
+	//
+	// QUESTION: Can you only supply an address that's a peer you are connected
+	// to, or can you supply any address?
+	Synchronize(NetAddress) error
```
lukechampine: ANSWER: only a peer you are connected to. So it might not make sense for this to take an argument.

modules/consensus/synchronize.go
```diff
@@ -11,47 +11,35 @@ import (
...
+			// to all peers.
+			err := s.Synchronize(peer)
+			if err != nil {
+				continue
 			}
-		}()
-		time.Sleep(time.Minute * 2)
+			time.Sleep(ResynchronizeTimeout)
+		}
+		time.Sleep(ResynchronizeTimeout)
```
lukechampine: constantly resynchronizing every 20s seems like overkill. Maybe extend the outer sleep to the original 2m delay

modules/consensus/synchronize.go
```diff
@@ -94,24 +113,24 @@ func (s *State) sendBlocks(conn modules.PeerConn) error {
...
-			break
+	lockID := s.mu.RLock()
+	{
+		for _, id := range knownBlocks {
+			bn, exists := s.blockMap[id]
+			if exists && bn.height <= s.height() && id == s.currentPath[bn.height] {
+				found = true
+				start = bn.height + 1 // start at child
+				break
+			}
```
lukechampine: what happened here? Just enclosing everything in a new block? `for` is already a block so I don't think the mutex use is confusing.

modules/consensus/validtransaction.go
```diff
@@ -174,42 +175,70 @@ func (s *State) validSiafunds(t types.Transaction) (err error) {
...
-// of the consensus set.
-func (s *State) ValidStorageProofs(t types.Transaction) (err error) {
-	id := s.mu.RLock()
-	defer s.mu.RUnlock(id)
-	return s.validStorageProofs(t)
+// TryTransactions applies the input transactions to the consensus set to
+// determine if they are valid. An error is returned IFF they are not a valid
+// set in the current consensus set. The size of the transactions and the set
+// is not checked.
+func (s *State) TryTransactions(txns []types.Transaction) error {
```
lukechampine: this function isn't used anywhere. Is it meant for a later PR?

lukechampine:
It's nice that the Consensus interface is so small, but subscribing the API server to it is kinda gross. None of the other modules have that requirement.

Also, any exported functions that aren't in the interface should be unexported.

modules/consensus/synchronize.go
DavidVorick: or maybe make it 60s?

You only contact 1 peer each time, is that an expensive operation?

modules/consensus/synchronize.go
DavidVorick: hmm. Maybe. It just felt like a lot of code was separating the two, I didn't see the for block when looking at it.

I also moved the variables some, perhaps that helped. Maybe the extra scoping isn't needed.

modules/consensus/validtransaction.go
DavidVorick: It's for the tpool to use, I thought it made sense to design the function here and put it to use when I tackle the tpool.

DavidVorick:
The consensus interface isn't complete, only the server uses it as an interface. All of the other modules still import it directly and they use a lot of the exported functions.

DavidVorick:
Most of the modules subscribe to consensus? Or are you saying that the api doesn't subscribe to anyone else? That will probably change once we start to use web sockets to keep the UI up to date.

lukechampine:
The latter. I think the API should have as little internal state as possible. It's supposed to be a very thin layer over the modules. It should introduce as little additional complexity as possible. That's why I don't like giving it module-specific fields and its own mutex.

>That will probably change once we start to use web sockets to keep the UI up to date.

Maybe, maybe not. That doesn't seem particularly immanent though.

DavidVorick:
How is the server supposed to communicate that there have been updates if it isn't subscribed to the updates? The only other way I see of doing that is having the UI subscribe to the modules directly, which means that each of the modules needs websockets. Or to move the polling from the UI to the server, but we can remove polling entirely

lukechampine:
The server doesn't push anything to the UI right now. Everything is done via polling. If that changes, we'll have to rethink the API model.

It's not like polling is universally bad either. It has its place.

DavidVorick:
One of the big advantages to using the update model is that it substantially lowers the locking complexity in the daemon, and I currently believe that locking complexity is one of our biggest impacts to performance, and it certainly adds a lot of developer overhead when things go wrong - simpler locking structures are much easier to debug.

If the server is switched to receiving updates from everybody then it won't have to make locking calls to the modules any more.

DavidVorick:
I do believe that having statefullness is bad, but I believe that locking complexity is substantially worse.

modules/consensus/synchronize.go
lukechampine: my understanding of the loop is that it will synchronize to up to 8 peers, sleeping for 20s in between each synchronize, and then sleep for an additional 20s. I'm suggesting that the latter sleep be a few minutes instead of 20s.

DavidVorick: ok, I'll do that

lukechampine:
That is true. I just can't shake the feeling that there's a better way.

modules/consensus/synchronize.go
DavidVorick: The one thing that gives me reservations is that you want the resynchronization loop to be faster than the median block. The average block is 10m but the median is closer to 7 minutes.

I know presumably it doesn't matter but as a failsafe it's more effective that way. Is it okay if I keep it at 2 minutes?

lukechampine: Yeah, I guess that's fine. But why do you want it to be faster than the block time?

PR #513, opened 2015-05-22
Ux tweaks

DavidVorick:
Renter returns an error if you upload something more than 500mb.
Host returns more reasonable error if you try to announce multiple times.
Changing your host settings works now (it didn't before).
File nicknames must have an extension matching the original extension.
Host defaults changed to 1gb max filesize and 5gb starting storage.

### DIFF
--- a/api/host.go
+++ b/api/host.go
@@ -20,15 +20,17 @@ func (srv *Server) hostAnnounceHandler(w http.ResponseWriter, req *http.Request)
 func (srv *Server) hostConfigureHandler(w http.ResponseWriter, req *http.Request) {
 	// load current settings
 	config := srv.host.Info().HostSettings
+	fmt.Println(config)
+	fmt.Println(req.URL)
 
 	// map each query string to a field in the host announcement object
 	qsVars := map[string]interface{}{
-		"totalStorage": &config.TotalStorage,
-		"minFilesize":  &config.MinFilesize,
-		"maxFilesize":  &config.MaxFilesize,
-		"minDuration":  &config.MinDuration,
-		"maxDuration":  &config.MaxDuration,
-		"windowSize":   &config.WindowSize,
+		"totalstorage": &config.TotalStorage,
+		"minfilesize":  &config.MinFilesize,
+		"maxfilesize":  &config.MaxFilesize,
+		"minduration":  &config.MinDuration,
+		"maxduration":  &config.MaxDuration,
+		"windowsize":   &config.WindowSize,
 		"price":        &config.Price,
 		"collateral":   &config.Collateral,
 	}
@@ -50,6 +52,7 @@ func (srv *Server) hostConfigureHandler(w http.ResponseWriter, req *http.Request
 		return
 	}
 
+	fmt.Println(config)
 	srv.host.SetSettings(config)
 	writeSuccess(w)
 }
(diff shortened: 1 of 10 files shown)


api/host.go
```diff
@@ -20,15 +20,17 @@ func (srv *Server) hostAnnounceHandler(w http.ResponseWriter, req *http.Request)
 func (srv *Server) hostConfigureHandler(w http.ResponseWriter, req *http.Request) {
 	// load current settings
 	config := srv.host.Info().HostSettings
+	fmt.Println(config)
+	fmt.Println(req.URL)
```
lukechampine: debugging?

api/host.go
```diff
@@ -50,12 +52,8 @@ func (srv *Server) hostConfigureHandler(w http.ResponseWriter, req *http.Request
 		return
 	}
 
+	fmt.Println(config)
```
lukechampine: ditto

modules/host/announce.go
```diff
@@ -66,6 +66,9 @@ func (h *Host) Announce() error {
 
 	// Add the transaction to the transaction pool.
 	err = h.tpool.AcceptTransaction(t)
+	if err == modules.ErrTransactionPoolDuplicate {
+		return errors.New("You have already announced yourself.")
+	}
```
lukechampine: errors shouldn't be capitalized or end in punctuation

modules/renter/upload.go
```diff
@@ -102,6 +103,10 @@ func (r *Renter) Upload(up modules.FileUploadParams) error {
 	lockID := r.mu.Lock()
 	defer r.mu.Unlock(lockID)
 
+	if filepath.Ext(up.Filename) != filepath.Ext(up.Nickname) {
+		return errors.New("nickname and file name must have the same extension")
+	}
```
lukechampine: this feels overly restrictive. How about instead we just autofill the Nickname field in the UI with the original filename? I think that's the behavior most people will want anyway.

api/host.go
DavidVorick: oh yeah whoops

modules/renter/upload.go
DavidVorick: that is a complicated fix sadly.

It's important for sharing that the file extension be reasonable, at least at the moment. I think this is an okay solution.

Auto-loading the name is going to be a good feature to have, but I'm putting it off for the time being.

lukechampine: https://github.com/NebulousLabs/Sia-UI/pull/118 ?

DavidVorick: ok

I still want to keep the restriction

lukechampine: I dunno. Definitely feels like a front-end thing to me. Why do you want to restrict it?

DavidVorick: I guess it is a frontend thing. If you add it to the front I'll remove it from the back.

It's one of those sharp-corner things. It makes it easier to shoot yourself in the foot and share a file where nobody will know the file extension they are supposed to use when downloading.

lukechampine: Yeah, I guess that would be an unfortunate scenario, even though I think it's unlikely. My preference would be for the UI to warn you about that possibility if you change the extension. I just don't want to restrict people who know what they're doing.

DavidVorick: let's leave it for now, I'll put a TODO here.

PR #514, opened 2015-05-22
Signed updates

lukechampine:
For reference, the command used to a sign a binary is:
```
openssl dgst -sha256 -sign [keyfile] -out siad.sig siad;
```


DavidVorick:
This looks good. Can I have some sort of testing confirmation that naive attempts to break it are unsuccessful (i.e., use an old siganture to validate a new binary, use no signature and see if it works, use an invalid signature and see if it works, and of course use a valid signature and see if it works)

lukechampine:
ok, these tests aren't super comprehensive, but they at least check that an invalid signature will be caught. During manual testing it took me a while just to figure out how to create a valid signature lol. So I'm pretty confident in this code.

It's possible to test a valid update, but it'd be a pain.

PR #515, opened 2015-05-22
Miner hashrate

DavidVorick:
This should be merged after the ux pr.

I reorganized the miner a bit. Now it uses the consensus package as a module instead of importing the state directly. Other smaller organizations were made, mostly that reduce dependency on locking.

Though it probably wasn't the best use of my time, I ran the miner through a profiler and optimized it. As it happens, the call `crypto.HashAll(b.ID, b.Nonce, blockRoot)` was taking the vast majority of the time, as one might expect, except that the hashing portion was only active 20% of the time. Garbage collection was taking up something like 70% of the time??? So I rewrote it, bypassing the HashAll function and the encoding package and instead encoding it manually. The result was a 3x speedup to our miner. My desktop is running this code and it's working very much to my favor.

Then I added some code to give the miner awareness of how fast it's hashing. The old code would get all the way up to about 1.1 Mhash/s on 7+ threads. To my surprise, 20 threads were just as efficient as 7, there didn't seem to be overhead with having too many threads. The new code runs at about 3.1 Mhash/s on my desktop, which is comparable to Forest's single 100mhz fpga core.

Finally, I added a tool to convert hashrate to blocks per month. The math took me a while but that's probably a calculation we're going to want to keep. It would probably be nice to have some api call that allows external miners to report their hashrate.

I plan on integrating these 2 new statistics with the UI. The only thing that's missing now is some indicator of how many blocks total you've found, (this indicator should probably also have an orphan count), and additional information about how long it will be until you get your payments.

### DIFF
--- a/modules/consensus.go
+++ b/modules/consensus.go
@@ -87,6 +87,16 @@ type ConsensusSet interface {
 	// This is a thread-safe way of managing updates.
 	ConsensusSetSubscribe(ConsensusSetSubscriber)
 
+	// CurrentBlock returns the most recent block on the heaviest fork known to
+	// the consensus set.
+	CurrentBlock() types.Block
+
+	// EarliestChildTimestamp returns the earliest timestamp that is acceptable
+	// on the current longest fork according to the consensus set. This is a
+	// required piece of information for the miner, who could otherwise be at
+	// risk of mining invalid blocks.
+	EarliestChildTimestamp(types.BlockID) (types.Timestamp, bool)
+
 	// Synchronize will try to synchronize to a specific peer. During general
 	// use, this call should never be necessary.
 	Synchronize(NetAddress) error
--- a/modules/miner/mine.go
+++ b/modules/miner/mine.go
@@ -20,14 +20,7 @@ func (m *Miner) blockForWork() (b types.Block) {
 	}
 
 	// Calculate the subsidy and create the miner payout.
-	height, exists := m.state.HeightOfBlock(m.parent)
-	if !exists {
-		if build.DEBUG {
-			panic("parent is not in state?")
-		}
-		return
-	}
-	subsidy := types.CalculateCoinbase(height + 1)
+	subsidy := types.CalculateCoinbase(m.height + 1)
 	for _, txn := range m.transactions {
 		for _, fee := range txn.MinerFees {
 			subsidy = subsidy.Add(fee)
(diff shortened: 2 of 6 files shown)


modules/miner/miner.go
```diff
@@ -28,12 +34,13 @@ type Miner struct {
 
 	subscribers []chan struct{}
 
-	mu sync.RWMutex
+	mu sync.Mutex
```
lukechampine: porque?

modules/miner/mine.go
```diff
@@ -45,6 +41,64 @@ func (m *Miner) blockForWork() (b types.Block) {
...
+// made 100 attempts to find a block, it calculates a hashrate based on how
+// much time has passed. The number of attempts in progress is set to 0
+// whenever mining starts or stops, which prevents weird low values from
+// cropping up.
+func (m *Miner) increaseAttempts() {
+	m.attempts++
+	if m.attempts >= 100 {
+		m.hashRate = int64((m.attempts * m.iterationsPerAttempt * 1e9)) / (time.Now().UnixNano() - m.startTime)
+		m.startTime = time.Now().UnixNano()
+		m.attempts = 0
```
lukechampine: I think you can avoid calling `.UnixNano()` all over the place. Just store it as a `time.Time` and convert where necessary.

modules/miner/mine.go
```diff
@@ -45,6 +41,64 @@ func (m *Miner) blockForWork() (b types.Block) {
...
+
+	// Iterate through a bunch of nonces (from a random starting point) and try
+	// to find a winnning solution.
+	nonce := (*uint64)(unsafe.Pointer(&hashbytes[32]))
+	*nonce = b.Nonce
+	for i := 0; i < int(iterations); i++ {
+		*nonce++
+		id := crypto.HashBytes(hashbytes)
+		if bytes.Compare(target[:], id[:]) >= 0 {
+			b.Nonce = binary.LittleEndian.Uint64(hashbytes[32:])
```
lukechampine: b.Nonce is still a uint64 eh?

Should probably specify `hashbytes[32:40]` just for clarity's sake.

modules/consensus.go
```diff
@@ -87,6 +87,16 @@ type ConsensusSet interface {
 	// This is a thread-safe way of managing updates.
 	ConsensusSetSubscribe(ConsensusSetSubscriber)
 
+	// CurrentBlock returns the most recent block on the heaviest fork known to
+	// the consensus set.
+	CurrentBlock() types.Block
```
lukechampine: Wasn't this why you needed to subscribe the API server?
I guess you need the height as well.

modules/miner/mine.go
DavidVorick: This PR is still compatible with the testnet, I want another testnet release today/tomorrow with all of our upgrades to get another round of feedback.

modules/consensus.go
DavidVorick: This is actually only called at miner startup =/

There's probably a way to avoid this call in the miner.

modules/miner/miner.go
DavidVorick: I thought it might be faster. Benchmarking seemed to indicate that it was a tiny bit faster, so I just left it. I think there's overhead when you start using readlocks instead of plainlocks.

modules/miner/mine.go
DavidVorick: That's going to include a few UI changes I have in mind.

modules/miner/miner.go
lukechampine: Interesting. I thought it would be the other way around. I guess it depends how much readlocking you're doing.

DavidVorick:
lol hadn't finished addressing your comments.

I'll pick it up on the next PR, I've decided to go ahead and address #494

PR #516, opened 2015-05-22
GET -> POST

lukechampine:
This only affects siac for now.

https://github.com/NebulousLabs/Sia-UI/pull/119 should be merged first, to prevent breaking the API calls. We can switch the UI over to POST requests later.


DavidVorick:
Looks like there are still deprecated api calls used by the testsuite.

PR #517, opened 2015-05-23
Miner polish

DavidVorick:
I finished polishing up the miner, including adding api calls for external miners, and adding 'hashrate' and 'blocks per month' to the miner info struct.

All that I feel remains is removing the multi-threading, which I refuse to do until cpu mining is utterly worthless. As long as a desktop cpu can find blocks occasionally on p2pool, it's worthwhile to me to keep the multithreading around.

If you read the extended message of the first commit, you see that I've started triggering the race detector. I do believe that this problem has been there the whole time, I can't really figure out why it would start complaining now. Also, it seems like it's a false positive, I'm not sure why it's complaining or how to make it stop.

I'll give it a closer look later on, but for the time being I'm not too worried.

### DIFF
--- a/api/miner.go
+++ b/api/miner.go
@@ -1,9 +1,29 @@
 package api
 
 import (
+	"encoding/json"
 	"fmt"
 	"net/http"
+
+	"github.com/NebulousLabs/Sia/crypto"
+	"github.com/NebulousLabs/Sia/types"
 )
+
+// MinerBlockforworkResp is the response struct returned by the API call
+// '/miner/blockforwork'.
+type MinerBlockforworkResp struct {
+	Block      types.Block
+	MerkleRoot crypto.Hash
+	Target     types.Target
+}
+
+// minerBlockforworkHandler handles the API call that retrieves a block for
+// work.
+func (srv *Server) minerBlockforworkHandler(w http.ResponseWriter, req *http.Request) {
+	var bfw MinerBlockforworkResp
+	bfw.Block, bfw.MerkleRoot, bfw.Target = srv.miner.BlockForWork()
+	writeJSON(w, bfw)
+}
 
 // minerStartHandler handles the API call that starts the miner.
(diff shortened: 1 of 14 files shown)


api/miner.go
```diff
@@ -1,10 +1,30 @@
...
+	Block      types.Block
+	MerkleRoot crypto.Hash
+	Target     types.Target
+}
+
+// minerBlockforworkHandler handles the API call that retrieves a block for
+// work.
+func (srv *Server) minerBlockforworkHandler(w http.ResponseWriter, req *http.Request) {
+	var bfw MinerBlockforworkResp
+	bfw.Block, bfw.MerkleRoot, bfw.Target = srv.miner.BlockForWork()
```
lukechampine: why do you need more than just the block? You can calculate the MerkleRoot by calling `.MerkleRoot` on the block, and you can get the target by calling `srv.cs.ChildTarget`. Unless the server and the miner are out of sync, I guess.

modules/miner/mine.go
```diff
@@ -38,48 +50,64 @@ func (m *Miner) blockForWork() (b types.Block) {
...
+		childtarget, _ := m.cs.ChildTarget(b.ID())
+		fmt.Println(childtarget)
+		fmt.Println(b.Nonce)
+		m.tpool.PurgeTransactionPool()
+		m.mu.Unlock()
+		return err
+	}
+	if build.Release != "testing" {
+		fmt.Println("Found a block! Reward will be received in 50 blocks.")
+	}
```
lukechampine: now that the miner keeps track of the blocks it's mined, we can move the "found block" messages to the UI. (Not in this PR, but eventually.)

api/miner.go
DavidVorick: the only time that bfw is called, the merkle root and target are also called immediately after. It seemed to make sense to lump them together in a function since they are inherently linked.

PR #518, opened 2015-05-24
Visible addresses

DavidVorick:
3 things (commits sum it up nicely):

1. I added visible addresses, which are addresses that get displayed to the user. Right now, that's only addresses that the user created manually and not addresses created by the miner, wallet, or host. Visible addresses are saved and reloaded.

2. UnlockHash type is now encoded as hex by JSON. There's no decoder at the moment.

3. Switch from 'blocks per month' to 'blocks per week', and reduce 2^(log(x) - log(y)) to x/y...

### DIFF
--- a/modules/miner/info.go
+++ b/modules/miner/info.go
@@ -25,10 +25,12 @@ func (m *Miner) MinerInfo() modules.MinerInfo {
 
 	// Using the hashrate and target, determine the number of blocks per month
 	// that could be mined.
-	hashesRequired, _ := big.NewRat(0, 1).SetFrac(types.RootDepth.Int(), m.target.Int()).Float64()
-	hashesPerWeek := big.NewInt(0).Mul(big.NewInt(60*60*24*7), big.NewInt(m.hashRate))
-	floatHPW, _ := big.NewRat(0, 1).SetInt(hashesPerWeek).Float64()
-	blocksPerWeek := floatHPW / hashesRequired
+	blocksPerWeek := float64(0)
+	if m.target.Int().Sign() > 0 {
+		hashesPerWeek := new(big.Int).Mul(big.NewInt(60*60*24*7), big.NewInt(m.hashRate))
+		hashesPerBlock := new(big.Int).Div(types.RootDepth.Int(), m.target.Int())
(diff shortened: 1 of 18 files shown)


modules/miner/info.go
```diff
@@ -26,19 +25,17 @@ func (m *Miner) MinerInfo() modules.MinerInfo {
 
 	// Using the hashrate and target, determine the number of blocks per month
 	// that could be mined.
-	floatMaxTarget, _ := big.NewRat(0, 1).SetInt(types.RootDepth.Int()).Float64()
-	floatCurTarget, _ := big.NewRat(0, 1).SetInt(m.target.Int()).Float64()
-	hashesRequired := math.Exp2(math.Log2(floatMaxTarget) - math.Log2(floatCurTarget))
-	hashesPerMonth := big.NewInt(0).Mul(big.NewInt(60*60*24*30), big.NewInt(m.hashRate))
-	floatHPM, _ := big.NewRat(0, 1).SetInt(hashesPerMonth).Float64()
-	blocksPerMonth := floatHPM / hashesRequired
+	hashesRequired, _ := big.NewRat(0, 1).SetFrac(types.RootDepth.Int(), m.target.Int()).Float64()
```
lukechampine: `new(big.Rat)`

modules/miner/info.go
```diff
@@ -26,19 +25,17 @@ func (m *Miner) MinerInfo() modules.MinerInfo {
...
+	hashesPerWeek := big.NewInt(0).Mul(big.NewInt(60*60*24*7), big.NewInt(m.hashRate))
+	floatHPW, _ := big.NewRat(0, 1).SetInt(hashesPerWeek).Float64()
+	blocksPerWeek := floatHPW / hashesRequired
```
lukechampine: I would do:
```go
hashesRequired := new(big.Int).Div(types.RootDepth.Int(), m.target.Int())
hashesPerWeek := new(big.Int).Mul(big.NewInt(60*60*24*7), big.NewInt(m.hashRate))
floatHPW, _ := new(big.Rat).SetFrac(hashesPerWeek, hashesRequired).Float64()
```

DavidVorick:
Should be better now.

PR #520, opened 2015-05-26
Peer protections

lukechampine:
This PR addresses the problem of bootstrap nodes becoming "full" (128 connections) and not accepting any new peers. If a node is full and a new peer tries to connect, it will disconnect one of its current peers to "make room" for the new peer.

This exposes us to a simple Sybil attack, where the attacker simple connects to the target repeatedly until all of its "honest" peers have been disconnected. Two protections against this attack are implemented here. One is to disallow multiple connections from the same IP address; this forces the attacker to acquire 128 unique IP addresses. The second (and stronger) defense is to only disconnect *inbound* peers (that is, peers who initiated the connection). This means that the node will always remain connected to the peers it has specifically chosen to connect to, such as the bootstrap nodes.

The attack is still possible, but it is much more difficult. The attacker has little control over who the target forms outbound connections to, so they need to force them to disconnect from all of their honest peers. This can be accomplished by performing the same Sybil attack on each of *those* peers, causing them to disconnect from the target. Note, however, that this requires that the attacker discover which peers the target has formed outbound connections to! As far as I can tell, this would not be an easy thing to do.

### DIFF
--- a/modules/gateway/peers.go
+++ b/modules/gateway/peers.go
@@ -28,6 +28,7 @@ type peer struct {
 	strikes uint32
 	addr    modules.NetAddress
 	sess    muxado.Session
+	inbound bool
 }
 
 func (p *peer) open() (modules.PeerConn, error) {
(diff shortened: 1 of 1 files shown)


modules/gateway/peers.go
```diff
@@ -2,10 +2,12 @@ package gateway
 
 import (
 	"errors"
+	"math/rand"
```
DavidVorick: don't use math/rand in a cryptosystem

it should be fine abusing math/rand still requires a lot of knowledge but it's not as good as crypto/rand.

DavidVorick:
This is good enough for now, though I sense that there are still problems with our peering protocol.

A lot of people believe that the peer network is Bitcoin's weakest link. It does seem to be the least well understood as far as theoretical strength goes.

PR #522, opened 2015-05-26
Siag xc

DavidVorick:
I added a command to build 'siag' for all platforms, and added a new name to 'siag'.

This change doesn't need to make master by 0.3.2.


DavidVorick:
bump

lukechampine:
oh yeah

PR #530, opened 2015-05-29
Consensus Testing

DavidVorick:
This PR started with me trying to find the source of a deadlock, and then discovering a panic that happens when trying to build the storage proof.

After about 40 minutes of searching I was unable to locate the bug, but it's pretty serious (consensus is returning a block that doesn't actually exist?). Instead of spending more time on that, I decided to just go ahead with 100% testing the consensus package. So this PR contains the first bit of that.

There shouldn't be any regressions in this PR, the problems that I failed to fix already exist.

Consensus testing is only partially complete. This does fix the 'future block' issue that's been high priority for an unfortunate number of months. I reorganized a handful of things to hopefully locking complexity.

I'm noticing that we still create a massive number of goroutines. It's not high priority, but we should begin to think about ways to reduce this.


modules/consensus/accept.go
```diff
@@ -82,51 +47,65 @@ func (s *State) validHeader(b types.Block) (err error) {
...
 
-	if newNode.heavierThan(s.currentBlockNode()) {
-		err = s.forkBlockchain(newNode)
+	if newNode.heavierThan(cs.currentBlockNode()) {
+		revertedNodes, appliedNodes, err = cs.forkBlockchain(newNode)
 		if err != nil {
-			return
+			return nil, nil, err
 		}
+		return revertedNodes, appliedNodes, nil
```
lukechampine: I'd prefer `return cs.forkBlockchain(newNode)`, or even
```go
revertedNodes, appliedNodes, err = cs.forkBlockchain(newNode)
return
```
I suppose this conforms to error conventions though

modules/consensus/accept.go
```diff
@@ -135,16 +114,27 @@ func (s *State) acceptBlock(b types.Block) error {
...
+		cs.updateSubscribers(revertedNodes, appliedNodes)
+	}
+
+	// Sanity check, if applied nodes is len 0, revertedNodes should also be
+	// len 0.
+	if build.DEBUG {
+		if len(appliedNodes) == 0 && len(revertedNodes) != 0 {
+			panic("appliedNodes and revertedNodes are mismatched!")
+		}
+	}
```
lukechampine: we should add an `assert` function to the build package.
```go
func Assert(assertion bool, err interface{}) {
    if build.DEBUG && !assertion {
        panic(err)
    }
}
```

modules/wallet/update.go
```diff
@@ -55,47 +55,33 @@ func (w *Wallet) applyDiff(scod modules.SiacoinOutputDiff, dir modules.DiffDirec
...
 	for _, diff := range w.unconfirmedDiffs {
 		w.applyDiff(diff, modules.DiffApply)
 	}
 
+	// Update the wallet age and consensus height. Though they update together,
+	// the wallet age can be altered/reset, but the consensus height cannot.
+	w.age -= len(cc.RevertedBlocks)
+	w.consensusHeight -= types.BlockHeight(len(cc.RevertedBlocks))
+	w.age += len(cc.AppliedBlocks)
+	w.consensusHeight += types.BlockHeight(len(cc.RevertedBlocks))
```
lukechampine: should be `AppliedBlocks`?

types/block_test.go
```diff
@@ -65,6 +65,136 @@ func TestBlockID(t *testing.T) {
...
+
+	// Try a block with 2 payouts that are too large.
+	b = Block{
+		MinerPayouts: []SiacoinOutput{
+			SiacoinOutput{
+				Value: coinbase,
+			},
+			SiacoinOutput{
+				Value: coinbase,
+			},
```
lukechampine: fyi, in compound struct literals you can omit the type:
```go
MinerPayouts: []SiacoinOutput{
    {Value: coinbase},
    {Value: coinbase},
}
```

modules/consensus/accept.go
DavidVorick: I've become a big fan of explicitly declaring your intentions in your return statements. If you say `return nil` instead of `return err`, then you are explicitly saying "if we reach this code, err is supposed to be nil, and even if I forgot to reset it or something, I'm going to return nil".

I would imagine that it also helps reviewers understand what's supposed to be happening.

modules/consensus/accept.go
DavidVorick: that could clean things up a bit.

assertion := if ... && ... || ...
build.Assert(assertion, ErrAssertionFailed)

modules/consensus/accept.go
lukechampine: agreed. I just think it's needlessly verbose in this case. The meaning of `return cs.forkBlockchain(newNode)` is pretty clear.

modules/wallet/update.go
DavidVorick: yes. whoops

types/block_test.go
DavidVorick: oh. There's probably a few tests in the `types` package that could also be cleaned up.

modules/consensus/accept.go
DavidVorick: one downside to this is that the 'if' statement will always be evaluated

types/block_test.go
DavidVorick: I'll go through at somepoint and find everywhere that this happens. Probably makes more sense in a later PR though.

DavidVorick: I committed the changes to my next pr.

PR #532, opened 2015-05-29
Synchronous uploads + downloads

lukechampine:
`Upload` and `Download` now block until finished instead of returning immediately. For `Download`, this means waiting for the file to be completely transferred, decrypted, and (eventually) rebuilt. For `Upload`, this just means waiting until the file is available, i.e. ready to be downloaded. Since we are currently using full replication, this means `Upload` returns as soon as one `filePiece` is uploaded.

The advantage to making these functions synchronous is that the error handling is simplified. Previously, upload and download errors were completely invisible to the user. Now, they will be reported like any other error.
Error handling is tricky for `Upload`, because it may fail for a combination of different reasons, i.e. hosts rejecting file contracts for different reasons. For now we simply return a generic "no pieces could be uploaded" error. It would be fairly easy to return the most common error though (e.g. "file is too large/small"), which would be much more helpful to the user.

Next step is to implement upload progress, probably modeled after download progress (i.e. adding methods to the `File` interface and polling it from the UI). Again, though, `Upload` is tricky because you are uploading to multiple sources. So how do you report a percentage? For now we can get away with using the largest percentage among the file pieces, but eventually we'll need something more sophisticated.
Once upload progress is implemented on the backend, it should be added to the frontend. Hopefully we can adopt the existing download bar code to this purpose.

### DIFF
--- a/modules/renter/download.go
+++ b/modules/renter/download.go
@@ -26,6 +26,7 @@ type Download struct {
 	// correctly on ARM and x86-32.
 	received uint64
 
+	startTime   time.Time
 	complete    bool
 	filesize    uint64
 	destination string
@@ -33,6 +34,11 @@ type Download struct {
 
 	pieces []filePiece
 	file   *os.File
+}
+
+// StartTime returns when the download was initiated.
+func (d *Download) StartTime() time.Time {
+	return d.startTime
 }
 
 // Complete returns whether the file is ready to be used.
@@ -126,7 +132,8 @@ func newDownload(file *file, destination string) (*Download, error) {
 	}
 
 	return &Download{
-		complete: false,
+		startTime: time.Now(),
+		complete:  false,
 		// for now, all the pieces are equivalent
 		filesize:    file.Pieces[0].Contract.FileSize,
 		received:    0,
--- a/modules/renter/download.go
+++ b/modules/renter/download.go
@@ -142,8 +142,6 @@ func newDownload(file *file, destination string) (*Download, error) {
 // specified.
 func (r *Renter) Download(nickname, destination string) error {
 	lockID := r.mu.Lock()
-	defer r.mu.Unlock(lockID)
-
 	// Lookup the file associated with the nickname.
 	file, exists := r.files[nickname]
 	if !exists {
@@ -158,6 +156,7 @@ func (r *Renter) Download(nickname, destination string) error {
 
 	// Add the download to the download queue.
 	r.downloadQueue = append(r.downloadQueue, d)
+	r.mu.Unlock(lockID)
 
 	// Download the file. We only need one piece, so iterate through the hosts
 	// until a download succeeds.
--- a/modules/renter/download.go
+++ b/modules/renter/download.go
@@ -18,7 +18,8 @@ var (
 	downloadAttempts = 5
 )
 
-// A Download is a file download that has been queued by the renter.
+// A Download is a file download that has been queued by the renter. It
+// implements the modules.DownloadInfo interface.
 type Download struct {
 	// Implementation note: received is declared first to ensure that it is
 	// 64-bit aligned. This is necessary to ensure that atomic operations work
@@ -64,6 +65,7 @@ func (d *Download) Nickname() string {
 func (d *Download) Write(b []byte) (int, error) {
 	n, err := d.file.Write(b)
 	// atomically update d.received
+	// TODO: atomic operations may not be necessary
 	atomic.AddUint64(&d.received, uint64(n))
 	return n, err
 }
@@ -104,36 +106,6 @@ func (d *Download) downloadPiece(piece filePiece) error {
 	return nil
 }
 
-// start initiates the download of a File.
-func (d *Download) start() {
-	// We only need one piece, so iterate through the hosts until a download
-	// succeeds.
-	for i := 0; i < downloadAttempts; i++ {
-		for _, piece := range d.pieces {
-			downloadErr := d.downloadPiece(piece)
-			if downloadErr == nil {
-				d.complete = true
-				d.file.Close()
-				return
-			}
-			// Reset seek, since the file may have been partially written. The
-			// next attempt will overwrite these bytes.
-			d.file.Seek(0, 0)
-			atomic.SwapUint64(&d.received, 0)
-		}
-
-		// This iteration failed, no hosts returned the piece. Try again
-		// after waiting a random amount of time.
-		randSource := make([]byte, 1)
-		rand.Read(randSource)
-		time.Sleep(time.Second * time.Duration(i*i) * time.Duration(randSource[0]))
-	}
-
-	// File could not be downloaded; delete the copy on disk.
-	d.file.Close()
-	os.Remove(d.destination)
-}
-
 // newDownload initializes a new Download object.
 func newDownload(file *file, destination string) (*Download, error) {
 	// Create the download destination file.
@@ -172,7 +144,7 @@ func (r *Renter) Download(nickname, destination string) error {
 	lockID := r.mu.Lock()
 	defer r.mu.Unlock(lockID)
 
-	// Lookup the File associated with the nickname.
+	// Lookup the file associated with the nickname.
 	file, exists := r.files[nickname]
 	if !exists {
 		return errors.New("no file of that nickname")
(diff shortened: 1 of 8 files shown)


DavidVorick:
upload should probably not return as complete until there are 3 available copies.

Adds some protection against downtime and such. Maybe 1 is enough.

lukechampine:
The [current default for PiecesRequired](https://github.com/NebulousLabs/Sia/blob/master/modules/renter/upload.go#L150) is 1. Should I change the success condition to use that field? And maybe change that value to 3? In other words:
```go
successPieces := 0
for i := 0; i < up.Pieces; i++ {
    if <-errChan == nil {
        successPieces++
        if successPieces >= file.RequiredPieces {
            return nil
        }
    }
}
```

DavidVorick:
file.RequiredPieces is so the downloader knows when the download is complete and more pieces don't need to be fetched. 

If you look at the file.Available() code, I believe that it waits until 3 copies have been uploaded. This is just to add extra safety margin before telling a user that a file is available. I'm not sure it's necessary but it makes me feel better inside.

lukechampine:
Nope, `Available` also uses `file.RequiredPieces`.
I'm going to use that as well for now. At the very least, it makes sense that they would use the same variable. Maybe we'll introduce a new field in the `file` type, distinct from `RequiredPieces`.

modules/renter/download.go
```diff
@@ -183,11 +160,40 @@ func (r *Renter) Download(nickname, destination string) error {
...
+			if downloadErr == nil {
+				// done
+				d.complete = true
+				d.file.Close()
+				return nil
+			}
+			// Reset seek, since the file may have been partially written. The
+			// next attempt will overwrite these bytes.
+			d.file.Seek(0, 0)
+			atomic.SwapUint64(&d.received, 0)
```
DavidVorick: ouch lol. 35% -> 0%

modules/renter/renter.go
```diff
@@ -68,7 +68,9 @@ func New(cs *consensus.State, hdb modules.HostDB, wallet modules.Wallet, saveDir
 	// need to make sure that scanAllFiles() didn't get called until the entire
 	// balance had loaded, which would require loading the entire blockchain.
 	// This also won't be a problem once we're also saving the addresses.
-	r.scanAllFiles()
+	//
+	// TODO: bring back this functionality when we have resumable uploads.
+	//r.scanAllFiles()
```
DavidVorick: I don't think that this ever worked well in the first place, but it's definitely something we need. If you turned off the renter at 7 out of 12 completed uploads, then when you turn it back on it should try those 5 remaining ones again.

I guess this is sufficient for now though. I'm looking forward to the overhaul.

modules/renter/upload.go
```diff
@@ -62,47 +62,46 @@ func (r *Renter) threadedUploadPiece(up modules.FileUploadParams, piece *filePie
 		// Select a host. An error here is unrecoverable.
 		host, err := r.hostDB.RandomHost()
 		if err != nil {
-			return
+			return err
```
DavidVorick: imagine that

modules/renter/download.go
lukechampine: I suppose for the user's sake you could just leave it at 35% until the next attempt surpassed that. This was just the easiest way.

modules/renter/upload.go
```diff
@@ -62,47 +62,46 @@ func (r *Renter) threadedUploadPiece(up modules.FileUploadParams, piece *filePie
...
-		if err != nil {
-			// The previous attempt didn't work. We will try again after
-			// sleeping for a randomized amount of time to increase our chances
-			// of success. This will help spread things out if there are
-			// problems with network congestion or other randomized issues.
-			randSource := make([]byte, 1)
-			rand.Read(randSource)
-			time.Sleep(time.Duration(attempts) * time.Duration(attempts) * 250 * time.Millisecond * time.Duration(randSource[0]))
-			continue
+		if err == nil {
```
DavidVorick: why did you switch these?

modules/renter/upload.go
```diff
@@ -62,47 +62,46 @@ func (r *Renter) threadedUploadPiece(up modules.FileUploadParams, piece *filePie
...
+		// sleeping for a randomized amount of time to increase our chances
+		// of success. This will help spread things out if there are
+		// problems with network congestion or other randomized issues.
+		randSource := make([]byte, 1)
+		rand.Read(randSource)
+		time.Sleep(time.Duration(attempts) * time.Duration(attempts) * 250 * time.Millisecond * time.Duration(randSource[0]))
 	}
+
+	// All attempts failed.
+	return errors.New("failed to upload filePiece")
```
DavidVorick: good enough I suppose. It would be nice to display the dominant error if there was one (like 80% of tries or something).

modules/renter/upload.go
```diff
@@ -115,23 +114,25 @@ func (r *Renter) Upload(up modules.FileUploadParams) error {
 	}
 
 	// Check for a nickname conflict.
+	lockID := r.mu.RLock()
 	_, exists := r.files[up.Nickname]
 	if exists {
 		return errors.New("file with that nickname already exists")
 	}
+	r.mu.RUnlock(lockID)
```
DavidVorick: bad!

Need to unlock before you return

DavidVorick:
```
lock()
_, exists = stuff
unlock()
if {}
```

modules/renter/upload.go
```diff
@@ -115,23 +114,25 @@ func (r *Renter) Upload(up modules.FileUploadParams) error {
...
 
-	// Check that the file exists and is less than 500mb.
+	// Check that the file exists and is less than 500 MiB.
```
DavidVorick: I read that as '500 Men in Black'

modules/renter/upload.go
```diff
@@ -140,10 +141,11 @@ func (r *Renter) Upload(up modules.FileUploadParams) error {
 	// number of pieces plus some buffer before we decide that an upload is
 	// okay.
 	if len(r.hostDB.ActiveHosts()) < 1 {
-		return errors.New("not enough hosts on the network to upload a file :( - maybe you need to upgrade your software")
+		return errors.New("not enough hosts on the network to upload a file")
```
DavidVorick: If there aren't enough hosts it's probably because they need to update their client.

I feel like they should be told that.

DavidVorick: oh or they are having network problems or something. I guess there are more troubleshooting things than just the one statement.

modules/renter/upload.go
```diff
@@ -152,13 +154,40 @@ func (r *Renter) Upload(up modules.FileUploadParams) error {
...
-		go r.threadedUploadPiece(up, &r.files[up.Nickname].Pieces[i])
+		// threadedUploadPiece will change the memory that the piece points
+		// to, which is useful because it means the file can be renamed
+		// without disrupting the upload process.
+		go func(piece *filePiece) {
+			errChan <- r.threadedUploadPiece(up, piece)
+		}(&r.files[up.Nickname].Pieces[i])
+	}
+	reqPieces := r.files[up.Nickname].PiecesRequired
+	r.mu.Unlock(lockID)
```
DavidVorick: This unlock is far enough away from the lock that it probably deserves separate scoping.

modules/renter/download.go
DavidVorick: what we would really want is to start the next download from the same place the failure left off. I think we will get that after the revamp

modules/renter/upload.go
lukechampine: ooh, good catch

modules/renter/upload.go
lukechampine: can't do separate scoping because it would capture local variables. (i.e. `reqPieces` and `errChan` would be undefined)

modules/renter/upload.go
lukechampine: Normally I wouldn't have touched it but I thought I was going to modify the function more. This control flow made more sense for the modifications I had in mind.
I still think it's slightly better this way, since the sleep is associated with the loop instead of the error condition. idk, it weirds me out to have the success condition in the main body of the loop.

modules/renter/upload.go
lukechampine: yeah that shouldn't be too hard since you have the whole set of errors available. stick them in a `map[error]int` or something to pick the most frequent one.

PR #535, opened 2015-05-30
Hardfork

DavidVorick:
This should be merged after consensus-testing.

This package does expand further on the testing, but the primary purpose is to introduce a few hardforking changes. Primarily:

+ a vulnerability in the fee structure has been closed
+ block nonce is now [8]byte
+ timestamp is in the block header

I've updated the miner api calls and documentation for /miner/blockforwork and /miner/submitblock. Josh was testing them with the gpu miner and they were not working as expected. It did seem as though the first block was getting through, but then everything just froze?

Maybe the program deadlocked. I mentioned a few other ideas in the slack.

Thanks to the thorough testing, I've been finding ways to clean up the code and have squashed a small number of bugs. There's a lot more ground to cover. I've yet to find any of the problems that have been appearing at random, but I haven't gotten to the suspected problem points either.

### DIFF
--- a/modules/miner/mine.go
+++ b/modules/miner/mine.go
@@ -54,7 +54,7 @@ func (m *Miner) blockForWork() (types.Block, crypto.Hash, types.Target) {
 
 // submitBlock takes a solved block and submits it to the blockchain.
 // submitBlock should not be called with a lock.
-func (m *Miner) submitBlock(b types.Block) error {
+func (m *Miner) SubmitBlock(b types.Block) error {
 	// Give the block to the consensus set.
 	err := m.cs.AcceptBlock(b)
 	if err != nil {
@@ -97,7 +97,7 @@ func (m *Miner) solveBlock(blockForWork types.Block, blockMerkleRoot crypto.Hash
 			id := crypto.HashBytes(hashbytes)
 			if bytes.Compare(target[:], id[:]) >= 0 {
 				copy(b.Nonce[:], nonce)
-				err = m.submitBlock(b)
+				err = m.SubmitBlock(b)
 				if err != nil {
 					return
 				}
@@ -197,12 +197,4 @@ func (m *Miner) SolveBlock(blockForWork types.Block, target types.Target) (b typ
 		}
 	}
 	return b, false
-}
-
-// SubmitBlock accepts a block with a valid target and presents it to the
-// consensus set.
-func (m *Miner) SubmitBlock(b types.Block) error {
-	m.mu.Lock()
-	defer m.mu.Unlock()
-	return m.submitBlock(b)
 }
--- a/modules/miner/mine.go
+++ b/modules/miner/mine.go
@@ -3,6 +3,7 @@ package miner
 import (
 	"bytes"
 	"crypto/rand"
+	"encoding/binary"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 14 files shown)


modules/miner/mine.go
```diff
@@ -85,24 +84,27 @@ func (m *Miner) submitBlock(b types.Block) error {
...
-	nonce := (*uint64)(unsafe.Pointer(&hashbytes[32]))
-	*nonce = b.Nonce
-	for i := 0; i < iterationsPerAttempt; i++ {
-		id := crypto.HashBytes(hashbytes)
-		if bytes.Compare(target[:], id[:]) >= 0 {
-			b.Nonce = binary.LittleEndian.Uint64(hashbytes[32:40])
-			err = m.submitBlock(b)
-			if err != nil {
+	binary.LittleEndian.PutUint64(hashbytes[40:48], uint64(b.Timestamp))
+	copy(hashbytes[48:], blockMerkleRoot[:])
```
lukechampine: why drop the unsafe stuff?

DavidVorick: I thought it would be simpler / more clear to do it this way. I have no idea which is faster, but my loose experimentation suggests that it really doesn't make a difference.

modules/miner/mine.go
```diff
@@ -188,13 +190,13 @@ func (m *Miner) FindBlock() (types.Block, bool, error) {
 // and is very slow.
 func (m *Miner) SolveBlock(blockForWork types.Block, target types.Target) (b types.Block, solved bool) {
 	b = blockForWork
-	for b.Nonce = 0; b.Nonce < iterationsPerAttempt; b.Nonce++ {
+	for i := 0; i < 255; i++ {
+		b.Nonce[0] = byte(i)
 		if b.CheckTarget(target) {
-			solved = true
-			return
+			return b, true
```
lukechampine: you only modify the first byte of the nonce here. Is that intentional?

modules/miner/mine.go
lukechampine: my concern is that this makes it harder to increment the nonce properly.

modules/miner/mine.go
DavidVorick: it's only purpose is for testing. Maybe we should keep it a little more useful than it currently is though.

modules/miner/mine.go
DavidVorick: that's a good point. perhaps I should go back to using unsafe.

DavidVorick:
This PR comes after the other PR I have open. And I've got another one on the way. I found out why there were panics during storage proof submitting - it has to do with the way 'append' works, and the fact that slices share memory underneath instead of being atomic.

I realized though that `copy` can be used with any slice, perhaps this will be our saving grace.

In the meantime, I've fixed the immediate issues on my upcoming pr.

DavidVorick:
uh.

I merged this branch with my 'further-testing' branch, and complete disaster ensued. I think there was a rebase that happened in this branch that wasn't properly applied to the child branch, and now I can't reconcile them =/

DavidVorick:
other than that this PR is ready to go, but please don't merge it until we've successfully got the history of the other one working. (child branch is called 'further-testing')

lukechampine:
I would prefer if you simply didn't have two dependent PRs open at the same time. There's little reason to do so.

PR #536, opened 2015-05-30
Upload progress

### LINKED ISSUE #527: Upload progress
There should be a `/renter/uploadqueue` call to provide information about uploads. This would help debug some uploading bugs as well.

Also, the renter gives no indication when a upload fails, or why.

lukechampine:
Upload progress was a bit more of a pain to implement than download progress. Hopefully when we turn our full attention to the renter, we can refactor things into a more sane design.

You can now watch your files upload in real-time, which is pretty cool. I'm going to explore how difficult this would be to add to Sia-UI. If it seems easy I'll do it, otherwise I'll switch my focus to the persist package.

Fixes #527

### DIFF
--- a/api/renter.go
+++ b/api/renter.go
@@ -25,10 +25,11 @@ type DownloadInfo struct {
 
 // FileInfo is a helper struct for the files API call.
 type FileInfo struct {
-	Available     bool
-	Nickname      string
-	Repairing     bool
-	TimeRemaining types.BlockHeight
+	Available      bool
+	UploadProgress float32
+	Nickname       string
+	Repairing      bool
+	TimeRemaining  types.BlockHeight
 }
 
 // LoadedFiles lists files that were loaded into the renter.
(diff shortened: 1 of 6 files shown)


api/renter.go
```diff
@@ -25,10 +25,11 @@ type DownloadInfo struct {
 
 // FileInfo is a helper struct for the files API call.
 type FileInfo struct {
-	Available     bool
-	Nickname      string
-	Repairing     bool
-	TimeRemaining types.BlockHeight
+	Available      bool
+	UploadProgress float32
```
DavidVorick: float32?

modules/renter/files.go
```diff
@@ -70,6 +73,22 @@ func (f *file) Available() bool {
 	return false
 }
 
+// UploadProgress indicates how close the file is to being available.
+func (f *file) UploadProgress() float32 {
+	lockID := f.renter.mu.RLock()
+	defer f.renter.mu.RUnlock(lockID)
+
+	// full replication means we just use the progress of most-uploaded piece.
```
DavidVorick: seems like a good approach.

DavidVorick:
Looks good, curious about the float32/64 choice though.

There may be some merge conflicts with the 'hardfork' branch, since that touches negotiate.go in both the renter and the host. But I don't think that it touches the same functions that you did, so maybe we're okay.

lukechampine:
64 bits of precision seemed unnecessary when you only care about 4 significant digits (e.g. 32.25%).

PR #538, opened 2015-06-01
Delayed Outputs Refactor

DavidVorick:
In the process of testing the consensus package, I realized that the way delayed siacoin outputs had been implemented was a little ad-hoc, and probably had some bugs stashed away somewhere.

I refactored the situation so that delayed siacoin outputs are handled in a strict fashion and so that there are many DEBUG checks which make sure nothing unusual has happened to the structure of the consensus set while forking.

Some additional refactoring in other places means that it's less likely that a line of code has been forgotten, and more likely to be detected if anything has been implemented incorrectly.

This branch should be merged after 'hardfork'.


DavidVorick:
Once the build completes, this PR is good to go.

modules/consensus/consensusset.go
```diff
@@ -14,14 +14,32 @@ import (
...
+	// skipping such a step noticably increases speed when loading many blocks
+	// at once. Usually, partialVerification is used when loading blocks from
+	// disk.
+	partialVerification verificationRigor = 1
+)
+
+// verificationRigor is a type indicating the intensity of verification that
+// should be using while accepting a block. For blocks that come from trusted
+// sources, the computationally expensive steps can be skipped.
+type verificationRigor byte
```
lukechampine: minor quibble, but I would use a `uint8` here. The types are equivalent, but one implies an integer and one implies a character. (ditto for `rune` and `int32`)

PR #547, opened 2015-06-03
Block node is fully tested now

DavidVorick:
Also I added all of the siafunds to the genesis block. Coverage for the consensus package is now at 2/3.

A lot of the untested code is auxiliary to the consensus process itself. I've only got major files left to test, and they should go faster than the others I've been working on.


DavidVorick:
Closing because I'm merging it with my next pr

PR #549, opened 2015-06-03
Persist

lukechampine:
Added persistence convenience functions (`Save`, `Load`, and `SaveFile`/`LoadFile`).
The wallet is next since it's more complicated. It will use a database rather than a flat JSON file.


api/daemon_test.go
```diff
@@ -27,7 +27,7 @@ func TestNewerVersion(t *testing.T) {
 
 	for version, expected := range versionMap {
 		if newerVersion(version) != expected {
-			t.Error("Comparing %v to %v should return %v", version, VERSION, expected)
+			t.Errorf("Comparing %v to %v should return %v", version, VERSION, expected)
```
DavidVorick: hwoops

persist/files.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+	if err := dec.Decode(&version); err != nil {
+		return err
+	}
+	if version != meta.Version {
+		return ErrBadVersion
+	}
+	if err := dec.Decode(data); err != nil {
+		return err
+	}
```
DavidVorick: Does this properly handle the MarshalIndent encoded json?

DavidVorick: testing would indicate that it does

modules/gateway/persist.go
```diff
@@ -1,38 +1,30 @@
...
 }
 
 func (g *Gateway) load() error {
-	file, err := os.Open(filepath.Join(g.saveDir, "nodes.json"))
-	if err != nil {
-		return err
-	}
 	var nodes []modules.NetAddress
-	err = json.NewDecoder(file).Decode(&nodes)
+	err := persist.LoadFile(persistMetadata, &nodes, filepath.Join(g.saveDir, "nodes.json"))
```
DavidVorick: That gives me an idea. Instead of putting everything in one file, we can put each major type of thing into a different file. That makes it relatively easy for users to swap out and modify certain configurations without needing to worry about preserving completely unrelated stuff.

future thoughts, of course.

modules/host/persist.go
```diff
@@ -3,39 +3,42 @@ package host
...
 	"path/filepath"
 
-	"github.com/NebulousLabs/Sia/encoding"
 	"github.com/NebulousLabs/Sia/modules"
+	"github.com/NebulousLabs/Sia/persist"
 )
 
+var persistMetadata = persist.Metadata{
+	Header:  "Sia Host",
+	Version: "0.3.3",
```
DavidVorick: One thing to note about these versions is that they don't need to keep parity with the Sia version.

Though, it probably does make sense to have the version be in-line with a Sia release. That makes it easier to go back and figure out where things changed later.

modules/host/persist.go
```diff
@@ -43,6 +46,8 @@ func (h *Host) load() error {
 		height := obligation.FileContract.WindowStart + StorageProofReorgDepth
 		h.obligationsByHeight[height] = append(h.obligationsByHeight[height], obligation)
 		h.obligationsByID[obligation.ID] = obligation
+		// update spaceRemaining
+		h.spaceRemaining -= int64(obligation.FileContract.FileSize)
```
DavidVorick: I'm glad to see this change.

modules/renter/persist.go
```diff
@@ -4,94 +4,54 @@ import (
...
 	"io"
 	"os"
 	"path/filepath"
 	"strconv"
+
+	"github.com/NebulousLabs/Sia/persist"
 )
 
 const (
 	PersistFilename = "renter.dat"
```
DavidVorick: looks like some of the filenames are *.json, and others are *.dat

We should be consistent

DavidVorick:
I don't know that 'files.go' is the best name. On Unix, everything is a file. Perhaps 'humanreadable.go' or something in that spirit. Or even 'json.go'.

modules/host/persist.go
lukechampine: yeah, that was my reasoning. We could do better in that respect though. To my knowledge, there are still two places where we don't use the same version as Sia: in the gateway connection protocol (version exchange) and in the shared file header. We should consider updating one or both of those.

lukechampine:
agreed, I couldn't really think of a name to capture what those functions are really doing, which is encoding data and prepending it with a standard header.

PR #550, opened 2015-06-03
DB handling

lukechampine:
Using recursion to solve this problem is a little unconventional, I know, but the alternatives were either "copy and paste code" or "use a GOTO", so...

### DIFF
--- a/modules/consensus/persist.go
+++ b/modules/consensus/persist.go
@@ -1,6 +1,7 @@
 package consensus
 
 import (
+	"os"
 	"path/filepath"
 
 	"github.com/NebulousLabs/Sia/persist"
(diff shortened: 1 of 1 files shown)


DavidVorick:
That runs the risk of an infinite loop, I think it's a lot safer to just copy code, or make the first bit a separate function. You don't have to copy much:

```
        db, err := persist.OpenDB(filepath.Join(saveDir, "chain.db"))
 	if err != nil {
 		return err
 	}
 	height, err := db.Height()
 	if err != nil {
 		return err
 	}
	// add genesis block
	cs.db = db
	return db.AddBlock(cs.blockRoot.block)
}
```

PR #552, opened 2015-06-04
Testing

DavidVorick:
This PR adds a lot of testing. Some unused functions get removed, some things get shifted and reorganized.

The difficulty adjustment algorithms was found to be non-converging, something I had feared but never verified. I wrote a bunch of simulations and replaced it with something that converges to the current difficulty within about 6000 blocks, it gets very close after 1000 though. It's closer to Bitcoin's difficulty adjustment algorithm, and sadly the difficulty only changes every 500 blocks now instead of every block.

In this PR all of the siafunds were added to the genesis block. The crypto package got more testing.


types/constants.go
```diff
@@ -103,21 +105,259 @@ func init() {
...
 		MinimumCoinbase = 30e3
 
 		RenterZeroConfDelay = 60 * time.Second
 	}
+
+	GenesisSiafundAllocation = []SiafundOutput{
+		{
+			Value:           NewCurrency64(2),
+			UnlockHash:      UnlockHash{4, 57, 229, 188, 127, 20, 204, 245, 211, 167, 232, 130, 208, 64, 146, 62, 69, 98, 81, 102, 221, 7, 123, 100, 70, 107, 199, 113, 121, 26, 198, 252},
+			ClaimUnlockHash: UnlockHash{4, 57, 229, 188, 127, 20, 204, 245, 211, 167, 232, 130, 208, 64, 146, 62, 69, 98, 81, 102, 221, 7, 123, 100, 70, 107, 199, 113, 121, 26, 198, 252},
```
lukechampine: would be nice if we had a function that would convert a hex-encoded string to a Hash...

PR #557, opened 2015-06-05
Hosting profit and potential profit fields

Mingling94:
This is untested as of 6.5, 6:42 pm. Though I've finished what I think needs to be done, it needs to be tested on livenet and verified through the corresponding Sia-UI PR


lukechampine:
moved to #558

PR #559, opened 2015-06-06
Renter sanity

DavidVorick:
The renter no longer chooses repeat hosts.

3 parallel uploads happen at a time. If a host is maliciously timing out or something, there are two backups. If all 3 are malicious, I guess that's a successful DoS vector right now.

If a host is just missing, the thread will be hogged for maybe ~100 seconds while it still does the exponential backoff thing, no more than that. If a host still isn't responsive, the next one is chosen.

### DIFF
--- a/modules/renter/upload.go
+++ b/modules/renter/upload.go
@@ -8,11 +8,18 @@ import (
 	"time"
 
 	"github.com/NebulousLabs/Sia/modules"
+	"github.com/NebulousLabs/Sia/sync"
 	"github.com/NebulousLabs/Sia/types"
 )
 
 const (
-	maxUploadAttempts = 8
+	maxUploadAttempts = 3
+)
+
+var (
+	errUploadFailed = errors.New("failed to upload to the desired host")
+
+	redundancy = 12
 )
 
 // checkWalletBalance looks at an upload and determines if there is enough
@@ -26,15 +33,11 @@ func (r *Renter) checkWalletBalance(up modules.FileUploadParams) error {
 	}
 	curSize := types.NewCurrency64(uint64(fileInfo.Size()))
 
-	// TODO: Change average to median so that outliers are ignored.
-	sampleSize := 12
 	var averagePrice types.Currency
-	for i := 0; i < sampleSize; i++ {
-		potentialHost, err := r.hostDB.RandomHost()
-		if err != nil {
-			return err
-		}
-		averagePrice = averagePrice.Add(potentialHost.Price)
+	sampleSize := redundancy * 2
+	hosts := r.hostDB.RandomHosts(sampleSize)
+	for _, host := range hosts {
+		averagePrice = averagePrice.Add(host.Price)
 	}
 	averagePrice = averagePrice.Div(types.NewCurrency64(uint64(sampleSize)))
 	estimatedCost := averagePrice.Mul(types.NewCurrency64(uint64(up.Duration))).Mul(curSize)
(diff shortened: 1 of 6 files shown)


modules/renter/upload.go
```diff
@@ -151,14 +151,52 @@ func (r *Renter) Upload(up modules.FileUploadParams) error {
 	r.mu.Unlock(lockID)
 
 	// Upload a piece to every host on the network.
```
lukechampine: every host on the network?? lol

DavidVorick: oh jeez

modules/renter/upload.go
```diff
@@ -151,14 +151,52 @@ func (r *Renter) Upload(up modules.FileUploadParams) error {
...
+	poolMutex := sync.New(250*time.Millisecond, 1)
 	errChan := make(chan error, len(f.Pieces))
-	for i := range f.Pieces {
-		// threadedUploadPiece will change the memory that the piece points
-		// to, which is useful because it means the file can be renamed
-		// without disrupting the upload process.
-		go func(piece *filePiece) {
-			errChan <- r.threadedUploadPiece(up, piece)
-		}(&f.Pieces[i])
+	for i := 0; i < 3; i++ {
```
lukechampine: 3 should be a `const`

lukechampine:
The new upload code is pretty complex. I can't see any obvious bugs, but I'd feel a lot better if we had even one upload/download test.

DavidVorick:
I did manual upload/download testing on this code. The api has an integration test which does uploading and downloading as well.

Therefore, this code does get covered automatically, even if not shown by the cover tool. Additionally, I made sure to cover it manually. In this PR at the very least, it works.

lukechampine:
also, wasn't the purpose of batching uploads to speed up (perceived) upload time? This PR doesn't seem to affect that, since you still wait for all of the pieces to complete.

idk, I guess I'm skeptical of the benefit, especially considering that we're going to overhaul the renter soon anyway.

modules/renter/upload.go
```diff
@@ -150,15 +151,54 @@ func (r *Renter) Upload(up modules.FileUploadParams) error {
...
+				if pieceFinished {
+					piece = &piecePool[0]
+					piecePool = piecePool[1:]
+					pieceFinished = false
+				}
+				host := hostPool[0]
+				hostPool = hostPool[1:]
+				poolMutex.Unlock(lockID)
+
+				err := r.threadedUploadPiece(host, up, piece)
```
lukechampine: this seems to require 1 host per piece. What happens if `len(hostPool)` < `len(piecePool)`?

modules/renter/upload.go
```diff
@@ -150,15 +151,54 @@ func (r *Renter) Upload(up modules.FileUploadParams) error {
...
-		// without disrupting the upload process.
-		go func(piece *filePiece) {
-			errChan <- r.threadedUploadPiece(up, piece)
-		}(&f.Pieces[i])
+	for i := 0; i < parallelUploads; i++ {
+		go func() {
+			lockID := poolMutex.Lock()
+			pieceFinished := true
+			var piece *filePiece
+			poolMutex.Unlock(lockID)
```
lukechampine: why do you need the lock here?

DavidVorick: hnnnng.

should be renter lock, not pool lock.

DavidVorick: oh wait no, was looking in the wrong place

lukechampine:
boom, no locks
```go
	// Upload to hosts in parallel. To facilitate this, we create channels of
	// hosts and file pieces, and spawn goroutines that attempt to match each
	// piece to a host.
	hostPool := make(chan modules.HostSettings, 3*redundancy)
	for _, host := range r.hostDB.RandomHosts(3 * redundancy) {
		hostPool <- host
	}
	piecePool := make(chan *filePiece, len(f.Pieces))
	for i := range f.Pieces {
		piecePool <- &f.Pieces[i]
	}

	errChan := make(chan error, len(f.Pieces))
	for i := 0; i < parallelUploads; i++ {
		go func() {
			for piece := range piecePool {
				var err error
				for host := range hostPool {
					err = r.threadedUploadPiece(host, up, piece)
					if err == nil {
						break
					}
				}
				errChan <- err
			}
		}()
	}
```

DavidVorick:
ok

PR #562, opened 2015-06-06
Synchronize tests

lukechampine:


modules/consensus/synchronize_test.go
```diff
@@ -0,0 +1,212 @@
...
+
+	// Mine on cst2 until it is more than 'MaxCatchUpBlocks' ahead of cst2.
+	// NOTE: we have to disconnect prior to this, otherwise cst2 will relay
+	// blocks to cst1.
+	err = cst1.gateway.Disconnect(cst2.gateway.Address())
+	if err != nil {
+		t.Fatal(err)
+	}
+	// TODO: more than 30 causes a race condition!
+	for cst2.cs.Height() < cst1.cs.Height()+20 {
```
lukechampine: this seems to be non-deterministic.

modules/consensus/synchronize_test.go
```diff
@@ -0,0 +1,212 @@
...
+	cst2.cs.mu.Unlock(lockID)
+
+	err = cst1.cs.Synchronize(cst2.gateway.Address())
+	if err != ErrBlockKnown {
+		t.Fatal("expected ErrBlockKnown, got", err)
+	}
+}
+
+func TestResynchronize(t *testing.T) {
+	t.Skip("takes way too long")
```
lukechampine: "too long" meaning 3 minutes.

DavidVorick:
deadlock in negotiate?

hmm

DavidVorick:
I looked at the code causing the deadlock and.... ???

I'm not sure why that deadlocked. There doesn't seem to be any locking inside of that function anywhere.

PR #563, opened 2015-06-07
Tweaks

DavidVorick:
A few last-minute tweaks to the api.

Stuff that's still not done:

- renter price estimator API call (with siac support)
- siafund API calls (with siac support)

Other than those two things, which are features as opposed to bugs, we are good to go.

### DIFF
--- a/modules/host/update.go
+++ b/modules/host/update.go
@@ -10,48 +10,13 @@ import (
 	"github.com/NebulousLabs/Sia/types"
 )
 
-// threadedCreateStorageProof creates a storage proof for a file contract
-// obligation and submits it to the blockchain.
-func (h *Host) threadedCreateStorageProof(obligation contractObligation, heightForProof types.BlockHeight) {
-	fullpath := filepath.Join(h.saveDir, obligation.Path)
-	file, err := os.Open(fullpath)
-	if err != nil {
-		fmt.Println(err)
-	}
-	defer file.Close()
-
-	segmentIndex, err := h.cs.StorageProofSegment(obligation.ID)
-	if err != nil {
-		fmt.Println(err)
-	}
-	base, hashSet, err := crypto.BuildReaderProof(file, segmentIndex)
-	if err != nil {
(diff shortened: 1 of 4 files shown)


modules/host/update.go
```diff
@@ -35,24 +34,35 @@ func (h *Host) createStorageProof(obligation contractObligation, heightForProof
...
 	}
 	_, _, err = h.wallet.AddStorageProof(id, sp)
 	if err != nil {
-		return
+		fmt.Println(err)
 	}
 	t, err := h.wallet.SignTransaction(id, true)
 	if err != nil {
-		return
+		fmt.Println(err)
```
lukechampine: err...don't you still want to return in this case? Continuing execution could be dangerous. Were these debugging statements?

DavidVorick: oh right. There needs to be a defer statement that deletes the obligation. I forgot to implement that.

modules/host/update.go
```diff
@@ -65,37 +89,15 @@ func (h *Host) ReceiveConsensusSetUpdate(cc modules.ConsensusChange) {
...
-			fullpath := filepath.Join(h.saveDir, obligation.Path)
-			stat, err := os.Stat(fullpath)
-			if err != nil {
-				fmt.Println(err)
-				return
-			}
-			h.deallocate(uint64(stat.Size()), obligation.Path) // TODO: file might actually be the wrong size.
-			shouldSave = true
-			delete(h.obligationsByID, obligation.ID)
+			go h.threadedCreateStorageProof(obligation, h.blockHeight)
```
lukechampine: since `threadedDeleteObligation` is a separate function now, you could write:
```go
go func() {
    err := h.threadedCreateStorageProof(obligation, h.blockHeight)
    if err != nil {
        fmt.Println(err)
        return
    }
    h.threadedDeleteObligation(obligation)
}()
```
then `threadedCreateStorageProof` could just return an error instead of printing all over the place

PR #565, opened 2015-06-07
price estimator api call

DavidVorick:


modules/renter/renter.go
```diff
@@ -83,8 +83,22 @@ func (r *Renter) Info() (ri modules.RentInfo) {
...
+
+	// Calculate the average cost of a file.
+	var averagePrice types.Currency
+	sampleSize := redundancy * 2
+	hosts := r.hostDB.RandomHosts(sampleSize)
+	for _, host := range hosts {
+		averagePrice = averagePrice.Add(host.Price)
+	}
+	averagePrice = averagePrice.Div(types.NewCurrency64(uint64(len(hosts))))
+	estimatedCost := averagePrice.Mul(types.NewCurrency64(6000)).Mul(types.NewCurrency64(1024 * 1024 * 1024))
```
lukechampine: a comment here would be nice

DavidVorick: saying what?

PR #567, opened 2015-06-08
friendlier price estimations

DavidVorick:
Redundancy reduced to 8

Estimations use 12 hosts instead of 30.

Since there are only ~20 hosts on the network, the most expensive hosts will usually get excluded (but not always).

### DIFF
--- a/api/server_test.go
+++ b/api/server_test.go
@@ -100,13 +100,13 @@ func newServerTester(name string, t *testing.T) *serverTester {
 	if err != nil {
 		t.Fatal("Failed to create miner:", err)
 	}
-	h, err := host.New(cs, tp, w, ":0", filepath.Join(testdir, "host"))
-	if err != nil {
-		t.Fatal("Failed to create host:", err)
-	}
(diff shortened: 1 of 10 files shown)


lukechampine:
why does the renter need to report `KnownHosts` if it's already available via `/hostdb/hosts/active`?

DavidVorick:
It was the easiest way I could think of. It's pretty ugly and I'm pretty unhappy with it overall.

I'll do my best to clean it all up in a bit.

lukechampine:
It seems very weird to me that you would add the hostdb to the host just to make a purely hostdb-related function available via the host API. If the host used the value of `Competition` internally, it would make more sense.
`host` and `hostdb` fulfill different roles. I don't understand why you would prefer to ask the host for hostdb-specific information.

DavidVorick:
hmmm

DavidVorick:
at the very least, the host is going to need hostdb information in the future anyway. I'm planning on adding some adaptive settings

PR #569, opened 2015-06-09
Blockchain explorer module base

triazo:


modules/blockexplorer/blockexplorer.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/NebulousLabs/Sia/sync"
+	"github.com/NebulousLabs/Sia/types"
+)
+
+// The blockexplorer module provides a glimpse into what the blockchain
+// currently looks like.
+
+// Basic structure to store the blockchain. Metadata may also be
+// stored here in the future
+type ExplorerState struct {
```
DavidVorick: Don't use the word 'State' in your structs. I think 'BlockExplorer' is a better name.

The consensus package has a 'State', but that's only because we hadn't made the convention yet. I'm in the process of renaming it to 'ConsensusSet', and that change is going to be completed once I've finished thoroughly testing the package.

modules/blockexplorer/blockexplorer.go
```diff
@@ -0,0 +1,55 @@
...
+
+	mu *sync.RWMutex
+}
+
+// New creates the internal data structures, and subscribes to
+// consensus for changes to the blockchain
+func New(cs modules.ConsensusSet) (bc *ExplorerState, err error) {
+
+	// Check that input modules are non-nil
+	if cs == nil {
```
DavidVorick: don't put a space between the function declaration and the first line of code.

modules/blockexplorer/blockexplorer.go
```diff
@@ -0,0 +1,55 @@
...
+		err = errors.New("Blockchain explorer cannot use a nil ConsensusSet")
+		return
+	}
+
+	// Initilize the module state
+	bc = &ExplorerState{
+		currentBlock:     cs.CurrentBlock(),
+		blockchainHeight: 0,
```
DavidVorick: I am in the process of deprecating the CurrentBlock() call, use the subscription model.

Also, it's very likely that 'CurrentBlock' will not return the genesis block.

DavidVorick:
There are just a few things to fix.

modules/blockexplorer/blockexplorer.go
DavidVorick: It is okay to call GenesisBlock()

lukechampine:
btw, the convention for commit messages is to write them like commands. So yours would be:
```
Add new module...
Keep track of the blockchain diff...
Add Api for blockexplorer...
Delete unused Struct items...
```

modules/blockexplorer/blockexplorer.go
triazo: I don't think either implementation accounts for this, but what would happen if consensus finds blocks and publishes them to subscribers before this module is initialized and has a chance to subscribe to consensus?

DavidVorick:
This looks good, but can you write unit tests?

The unit tests might be a bit strange. A good place to look for an example is modules/host_test.go. The `updateWait()` functions are important for keeping all parts of your program synchronized, since the update model is asynchronous by design. Any time that the consensus set updates, you're going to want to call csUpdateWait().

The block explorer doesn't look at the transaction pool at all, so you don't need to worry about the tpUpdateWait() stuff. I can help you if you get confused.

triazo:
Added some tests.  As this version of the module doesn't do much, the tests are a bit sparse, just checking that the package is updating properly with consensus.

DavidVorick:
rebase this on top of master, hopefully the build will start working again.

PR #571, opened 2015-06-09
gateway fixes

lukechampine:
debugging why Travis is failing. Don't merge yet pls

### DIFF
--- a/modules/consensus/synchronize.go
+++ b/modules/consensus/synchronize.go
@@ -64,7 +64,11 @@ func (s *State) receiveBlocks(conn modules.PeerConn) error {
 			s.verificationRigor = fullVerification
 			acceptErr := s.acceptBlock(block)
 			s.mu.Unlock(lockID)
-			if acceptErr != nil && acceptErr != modules.ErrNonExtendingBlock {
+			// these errors are benign
+			if acceptErr == modules.ErrNonExtendingBlock || acceptErr == ErrBlockKnown {
+				acceptErr = nil
+			}
+			if acceptErr != nil {
 				return acceptErr
 			}
 		}
--- a/modules/gateway/nodes_test.go
+++ b/modules/gateway/nodes_test.go
@@ -49,39 +49,45 @@ func TestRemoveNode(t *testing.T) {
 func TestRandomNode(t *testing.T) {
 	g := newTestingGateway("TestRandomNode", t)
 	defer g.Close()
-	id := g.mu.RLock()
 
+	id := g.mu.RLock()
 	if addr, err := g.randomNode(); err != nil {
 		t.Fatal("randomNode failed:", err)
 	} else if addr != g.Address() {
 		t.Fatal("randomNode returned wrong address:", addr)
 	}
 	g.mu.RUnlock(id)
-	g.removeNode(g.Address())
+
 	id = g.mu.Lock()
+	g.removeNode(g.myAddr)
+	g.mu.Unlock(id)
+
+	id = g.mu.RLock()
 	if _, err := g.randomNode(); err != errNoPeers {
 		t.Fatalf("randomNode returned wrong error: expected %v, got %v", errNoPeers, err)
 	}
+	g.mu.RUnlock(id)
 
 	nodes := map[modules.NetAddress]int{
 		"111.111.111.111:1111": 0,
 		"111.111.111.111:2222": 0,
 		"111.111.111.111:3333": 0,
 	}
+	id = g.mu.Lock()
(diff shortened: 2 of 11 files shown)


lukechampine:
ok, gateway stuff is fixed now. Other stuff is still failing though

DavidVorick:
Tests are only failing remotely?

DavidVorick:
TestHostAnnouncement is timing out, that's not such a big deal.

And then there's the deadlock, which looks like my fault.

lukechampine:
Last bug seems to be in wallet/transaction_test.go. Maybe you can figure out why

DavidVorick:
\>.< it spilled 400 goroutines on panic. I wish there was a setting to make it not do that.

TestUploadAndDownload still seems to be having occasional trouble. I think I can look into that one too, but mostly I probably just want to 'Skip()' it, because we're going to be completely rehashing it all anyway, and it's a synchronization thing.

DavidVorick:
The errors being spit out during TestMining are because it's mining faster than the consensus package can verify blocks, so it's making a bunch of orphans. We may just want to disable that test.

It doesn't throw errors though, just prints them.

DavidVorick:
what is the error in wallet/transaction_test.go? That's not what's showing up on the travis build.

DavidVorick:
ok, here's what's happening with the wallet:

on the fast computers, the synchronization is actually wrong and it checks the balance before the block is mined. This is bad, it should wait until after the block is mined. After 'FindBlock()', call 'csUpdateWait()'.

Thing is, that's going to consistently cause an error, because in the check I forgot to take into account the balance earned from mining the block (a previous block matured and gave the balance to the wallet). So you need to update the number in the check to reflect a new block coming into maturity (with coinbase 299,998e24)

That will fix the test.

lukechampine:
do you want to handle it or should I? Sounds like a simple fix either way.

DavidVorick:
Probably best if you do it

lukechampine:
ok, fingers crossed...

lukechampine:
thank god

DavidVorick:
Muxado totally panicked

DavidVorick:
I'm merging this anyway b/c we need to get the fixes to the other branches, but it's not done yet.

modules/consensus/synchronize.go
```diff
@@ -64,7 +64,11 @@ func (s *State) receiveBlocks(conn modules.PeerConn) error {
 			s.verificationRigor = fullVerification
 			acceptErr := s.acceptBlock(block)
 			s.mu.Unlock(lockID)
-			if acceptErr != nil && acceptErr != modules.ErrNonExtendingBlock {
+			// these errors are benign
+			if acceptErr == modules.ErrNonExtendingBlock || acceptErr == ErrBlockKnown {
+				acceptErr = nil
+			}
+			if acceptErr != nil {
```
DavidVorick: Is this the right approach? Don't we want to indicate that we didn't need the blocks?

I guess not.

modules/gateway/nodes_test.go
```diff
@@ -96,7 +102,9 @@ func TestShareNodes(t *testing.T) {
 	defer g2.Close()
 
 	// add a node to g2
+	id := g2.mu.Lock()
 	g2.addNode(dummyNode)
+	g2.mu.Unlock(id)
```
DavidVorick: yikes this is a lot of locking. You sure there isn't a better approach?

modules/transactionpool/accept.go
```diff
@@ -235,5 +235,10 @@ func (tp *TransactionPool) RelayTransaction(conn modules.PeerConn) error {
 	if err != nil {
 		return err
 	}
-	return tp.AcceptTransaction(t)
+	err = tp.AcceptTransaction(t)
+	// ErrTransactionPoolDuplicate is benign
+	if err == modules.ErrTransactionPoolDuplicate {
+		err = nil
```
DavidVorick: no error returned on duplicates?

lukechampine:
weird, it panicked during `make cover` but not `make test-long`

modules/gateway/nodes_test.go
lukechampine: I could combine them in some places, but it's not a big deal.

modules/transactionpool/accept.go
lukechampine: only in RelayTransaction. The main reason I took it out is because it was clogging gateway.log with "errors." Receiving a duplicate is expected behavior so I decided to not count it as an error. Ditto for `ErrKnownBlock` above.

PR #573, opened 2015-06-09
Added additional fields from `/miner/status/` to siac.

Nutomic:
Thought this might be useful ;)

### DIFF
--- a/siac/minercmd.go
+++ b/siac/minercmd.go
@@ -57,8 +57,13 @@ func minerstatuscmd() {
 	fmt.Printf(`Miner status:
 State:   %s
 Threads: %d (%d active)
+HashRate: %d
+BlocksPerWeek: %f
+BlocksMined: %d
+OrphansMined: %d
 Address: %x
-`, status.State, status.Threads, status.RunningThreads, status.Address)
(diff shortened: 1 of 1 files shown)


siac/minercmd.go
```diff
@@ -57,8 +57,13 @@ func minerstatuscmd() {
 	fmt.Printf(`Miner status:
 State:   %s
 Threads: %d (%d active)
+HashRate: %d
+BlocksPerWeek: %f
+BlocksMined: %d
+OrphansMined: %d
```
lukechampine: how about this format:
```
State:   %s
Threads: %d (%d active)
Address: %x

Hash Rate:    %d (est. %0.2f blocks per week)
Blocks Mined: %d (%d orphans)
```

lukechampine:
Sorry about this, but I'm going to have to close this in the interest of expedience. We are trying to get a release out tonight, and we'd like more detailed miner stats to be in that release. I've implemented those changes in 88c7aa6ed077894e868e7c6903d3d44039a8fbaf.
We appreciate this PR though; without it, it probably would not have occurred to me to update siac. And of course, we'd love to see more pull requests from you (and the rest of the Sia community) in the future.

PR #577, opened 2015-06-10
break apart locking in long test

DavidVorick:
the deadlock being reported was happening because the function was taking too long - a false positive on the deadlock detector.


DavidVorick:
Additionally, I believe that there is a fix in master which isn't in your branch which should solve some of the race conditions related to the transaction pool.

DavidVorick:
on this build, the only thing that failed was Test-RPC2

PR #585, opened 2015-06-12
Better Versioning

lukechampine:
This change consolidates the various version numbers into one spot: `build.Version`. Well, mostly; the persist stuff is a work in progress. It also adds a `/daemon/version` API call.

The gateway is the test case for this new version management scheme. The idea is that in order to check that a version is compatible, you pass it to `build.VersionCmp`, along with a "minimum compatible version." As an example, when a connecting peer sends its version to the gateway, the gateway checks "is this version greater than v0.3.3?" by calling `build.VersionCmp(remoteVersion, "0.3.3") > 0`. If the check fails, the gateway refuses to connect.

This can be extended to the persist files in a similar fashion: they simply indicate the oldest version they are compatible with (i.e. the last version that changed the save format). This has not been implemented yet.

The downside of this is that we have to keep track of what is compatible with what. But that was sort of inevitable anyway. Basically, any time you alter a protocol -- consensus, peer negotiation, persist data -- you just have to remember to bump the "minimum acceptable version" to the current version.
(Of course, now that we're launched, we should also maintain compatibility tools that can read the old formats and update them to the new format.)

I also found a hackish way to ensure that we don't build releases with the wrong version number. You can pass a flag to the Go linker that sets the value of an unset string variable. So we could add a special test that compares such a string to `build.Version` and fails if they are not equal. Then, in the Makefile, we alter the `make xc` target to run a special test that sets the string to the same version passed to `goxc`.
This struck me as fairly ugly though, so I left it out of this PR.

### DIFF
--- a/api/daemon_test.go
+++ b/api/daemon_test.go
@@ -4,33 +4,9 @@ import (
 	"fmt"
 	"net/http"
 	"testing"
+
+	"github.com/NebulousLabs/Sia/build"
 )
-
-// TestNewerVersion checks that in all cases, newerVesion returns the correct
-// result.
-func TestNewerVersion(t *testing.T) {
-	// If the VERSION is changed, these tests might no longer be valid.
-	if VERSION != "0.3.3" {
-		t.Fatal("Need to update version tests")
-	}
-
-	versionMap := map[string]bool{
-		VERSION:   false,
-		"0.1":     false,
-		"0.1.1":   false,
-		"1":       true,
-		"0.9":     true,
-		"0.3.2.9": false,
-		"0.3.3.0": true,
-		"0.3.3.1": true,
-	}
-
-	for version, expected := range versionMap {
-		if newerVersion(version) != expected {
-			t.Errorf("Comparing %v to %v should return %v", version, VERSION, expected)
-		}
-	}
-}
 
 type updateHandler struct {
 	version string
@@ -60,7 +36,7 @@ func TestSignedUpdate(t *testing.T) {
 	updateURL = "http://localhost:8080"
 
 	// same version
-	uh.version = VERSION
+	uh.version = build.Version
 	var info UpdateInfo
 	st.getAPI("/daemon/updates/check", &info)
 	if info.Available {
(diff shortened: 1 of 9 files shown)


api/daemon_test.go
```diff
@@ -68,7 +44,7 @@ func TestSignedUpdate(t *testing.T) {
 	}
 
 	// newer version
-	uh.version = "0.4"
+	uh.version = "100.0"
```
DavidVorick: 100.0?

build/version.go
```diff
@@ -0,0 +1,59 @@
...
+
+// VersionCmp returns an int indicating the difference between a and b. It
+// follows the convention of bytes.Compare and big.Cmp:
+//
+//   -1 if a <  b
+//    0 if a == b
+//   +1 if a >  b
+//
+// One important quirk is that "1.1.0" is considered newer than "1.1", despite
+// being numerically equal.
```
DavidVorick: it's good that the quirk is documented, but we should never have a release that goes 1.1 -> 1.1.0 nor 1.1.0 -> 1.1

build/version_test.go
```diff
@@ -0,0 +1,53 @@
...
+		a, b string
+		exp  int
+	}{
+		{"0.1", "0.0.9", 1},
+		{"0.1", "0.1", 0},
+		{"0.1", "0.1.1", -1},
+		{"0.1", "0.1.0", -1},
+		{"0.1", "1.1", -1},
+		{"0.1.1.0", "0.1.1", 1},
+	}
```
DavidVorick: :)

glad we don't have to update the test every time anymore

modules/gateway/peers.go
```diff
@@ -177,7 +179,7 @@ func (g *Gateway) Connect(addr modules.NetAddress) error {
 		return errors.New("peer rejected connection")
 	}
 	// decide whether to accept this version
-	if remoteVersion != version {
+	if build.VersionCmp(remoteVersion, "0.3.3") < 0 {
```
DavidVorick: I like this new way of doing things

build/version.go
lukechampine: yeah, that would be pretty weird.

PR #590, opened 2015-06-14
Siafunds

DavidVorick:
This PR feels like a hack job. The last hack job for a while, thankfully, but it builds on a lot of incomplete concepts. We need to get the siafunds spender out to our crowdfund holders asap, which is why I was okay building on top of deprecated designs.

One seemingly small change had major cascades. The genesis block outputs hadn't originally been included in the subscription list, which is a problem because that's where all the siafund outputs exist. So I added the genesis block to the notification list.

Problem is, this happens immediately, on startup. So, before the test suite ever mines a block, there's already a notification, and there needs to be a check that the notification spread to all of the other modules. This resulted in some headache, and ultimately a conclusion that the notification system could be revised.

Support for siag files has been added, but in a bit of a weird way. You don't load them into your wallet, because siag files are actually multisig files. The idea of multisig files is that you don't have everything in one place, yet you need to put everything in one place to spend the siafunds. Usage is complex and counterintuitive. Could use revising at some point, but I just accepted the limitations for now.

Siag stuff eventually needs to be fully integrated into the wallet itself.

There's no way to rescan the diff set when you add a new siafund key to the wallet. So, after adding the siafund key you just have to restart siad. That's pretty lame, but changing it would require some fairly major architectural changes. This just points back to the limitations of our current subscription architecture. I'll be looking at it more closely later this week as I return to polishing up the consensus package.

The api calls still aren't implemented for spending siafunds, because you might need to load multiple keys, instead of just one key. I figured that api call is something you can add when you implement the other calls in siac.

Ideally, the siafund stuff (including siac) will be merged into master sometime before tomorrow ends. I don't want to keep people waiting.


modules/wallet/siag.go
```diff
@@ -0,0 +1,210 @@
...
+	Header           string
+	Version          string
+	Index            int // should be uint64 - too late now
+	SecretKey        crypto.SecretKey
+	UnlockConditions types.UnlockConditions
+}
+
+// SpendSiagSiafunds sends siafunds to another address. The siacoins stored in
+// the siafunds are sent to an address in the wallet.
+func (w *Wallet) SpendSiagSiafunds(amount types.Currency, dest types.UnlockHash, keyfiles []string) (txn types.Transaction, err error) {
```
lukechampine: these named return values are never used

modules/wallet/siag.go
```diff
@@ -0,0 +1,210 @@
...
+		if err != nil {
+			return types.Transaction{}, err
+		}
+
+		if skps[i].Header != SiagFileHeader {
+			return types.Transaction{}, ErrUnknownHeader
+		}
+		if skps[i].Version != SiagFileVersion {
+			return types.Transaction{}, ErrUnknownVersion
+		}
```
lukechampine: the `persist.LoadFile` function is useful here. See `gateway/persist.go` for an example.

modules/wallet/siag.go
```diff
@@ -0,0 +1,210 @@
...
+			break
+		}
+	}
+	w.mu.RUnlock(lockID)
+	if availableSiafunds.Cmp(amount) < 0 {
+		return types.Transaction{}, ErrInsufficientSiafunds
+	}
+
+	// Truncate the keys to exactly the number needed.
+	skps = skps[:skps[0].UnlockConditions.SignaturesRequired]
```
lukechampine: out-of-bounds potential?

modules/wallet/siag.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+	if skp.Header != SiagFileHeader {
+		return ErrUnknownHeader
+	}
+	if skp.Version != SiagFileVersion {
+		return ErrUnknownVersion
+	}
+	w.siafundAddresses[skp.UnlockConditions.UnlockHash()] = struct{}{}
+	// Janky println... but it's important to get this message to the user.
+	println("Loaded a siafund address. Please note that the private key was not loaded, you must KEEP the original keyfile. You must restart said before your balance can be displayed.")
```
lukechampine: siad

api/debug.go
```diff
@@ -125,3 +125,26 @@ func (srv *Server) mutexTestHandler(w http.ResponseWriter, req *http.Request) {
...
+// sending works.
+func (srv *Server) debugSiafundsendtestHandler2(w http.ResponseWriter, req *http.Request) {
+	dest1o1 := types.UnlockHash{214, 166, 197, 164, 29, 201, 53, 236, 106, 239, 10, 158, 127, 131, 20, 138, 63, 221, 230, 16, 98, 247, 32, 77, 210, 68, 116, 12, 241, 89, 27, 223}
+	_, err := srv.wallet.SpendSiagSiafunds(types.NewCurrency64(30), dest1o1, []string{"siag0of2of3.siakey", "siag1of2of3.siakey"})
+	if err != nil {
+		writeError(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+	writeSuccess(w)
+}
```
lukechampine: pls remove these

lukechampine:
you're right, this is super janky. I don't like any of it. But I don't really understand it either.
Could the .siakey files be moved inside the test files? It doesn't look like you're using all of them anyway.

modules/wallet/siag.go
DavidVorick: too late for the time being

modules/wallet/siag.go
DavidVorick: hmm. At this point I think we've both verified that there's at least 1 key and that there's also at least SignaturesRequired keys, but I'll look again.

DavidVorick:
As far as 'it works without being buggy' goes, it should be pretty solid. It's just not a very strong design, nor is it much in the way of maintainability.

modules/wallet/siag.go
DavidVorick: yeah, all good here, no change needed

DavidVorick:
ok, hopefully addressed all of your concerns

PR #591, opened 2015-06-14
Implements additional API calls used in the Sia-Block-Explorer

triazo:
This pull request is on top of my previous one.

Can you comment on the data structures names in modules/blockexplorer.go? They seem very verbose to me but I didn't want to create potential confusion/name collisions with other modules


lukechampine:
ok, rebase this onto master now

lukechampine:
test failing is my fault -- need to delete line 123 of rpc_test.go. You can do it in this PR if you want.

lukechampine:
David and I think the API calls should be organized like so:
- `/blockexplorer/status`: contains everything in `/blockchain`, `/current`, `/siacoins`, and `/filecontracts`
- `/blockexplorer/block`: rename of `/blockdata`. Supply height or ID in the query string.

modules/blockexplorer.go
```diff
@@ -4,12 +4,31 @@ import (
...
+// Used for the BlockInfo call
+type ExplorerBlockData struct {
+	Timestamp types.Timestamp // The timestamp on the block
+	Target    types.Target    // The target the block was mined for
+	Size      uint64          // The size in bytes of the marshalled block
+}
+
+type ExplorerStatus struct {
+	Height            types.BlockHeight
+	Block             types.Block
```
lukechampine: make this `CurrentBlock`

DavidVorick:
please don't name a file 'data.go'. In a computer, everything is data.

modules/blockexplorer.go
```diff
@@ -4,12 +4,31 @@ import (
...
+	Target            types.Target
+	CurrencySent      types.Currency
```
DavidVorick: I don't think that CurrencySent is a useful metric

modules/blockexplorer/blockexplorer.go
```diff
@@ -16,11 +17,30 @@ import (
...
+	// currencySent keeps track of how much currency has been
+	// i.e. sending siacoin to somebody else
+	currencySent types.Currency
+
+	// fileContracts holds the current number of file contracts
+	fileContracts uint64
+
+	// fileContractCost hold the amout of currency tied up in file
+	// contracts
+	fileContractCost types.Currency
```
DavidVorick: Does this include expired contracts?

api/blockexplorer.go
```diff
@@ -1,10 +1,40 @@
...
+	"fmt"
 	"net/http"
+
+	"github.com/NebulousLabs/Sia/types"
 )
 
-// Handles the api call to get the current block from the block explorer
-func (srv *Server) blockexplorerCurrentBlockHandler(w http.ResponseWriter, req *http.Request) {
-	writeJSON(w, srv.blocke.CurrentBlock())
+// Handles the call to get high level information about the blockchain
```
DavidVorick: 'high level information' is a really vague term. What does it mean?

api/blockexplorer.go
```diff
@@ -1,10 +1,40 @@
...
+
+	// If a range end is not given, assume the range end to be one
+	// greater than the range start, returning one block
+	_, err = fmt.Sscan(req.FormValue("finish"), &finish)
+	if err != nil {
+		finish = start + 1
+	}
+
+	// Bounds checking is done inside BlockInfo
+	blocks, err := srv.blocke.BlockInfo(start, finish)
```
DavidVorick: naming the variable 'blocks' makes me think that you are returning blocks, but you are returning a different object. The name should reflect what's being returned.

modules/blockexplorer/blockexplorer.go
```diff
@@ -38,21 +58,28 @@ func New(cs modules.ConsensusSet) (bc *BlockExplorer, err error) {
 	}
 
 	// Initilize the module state
-	bc = &BlockExplorer{
+	be = &BlockExplorer{
 		currentBlock:     cs.GenesisBlock(),
 		blockchainHeight: 0,
```
DavidVorick: values in go default to the '0' value, you don't need to declare it

modules/blockexplorer/blockexplorer.go
```diff
@@ -38,21 +58,28 @@ func New(cs modules.ConsensusSet) (bc *BlockExplorer, err error) {
...
+		currencySent:     types.NewCurrency64(0),
+		fileContracts:    0,
+		fileContractCost: types.NewCurrency64(0),
+		blocks:           make([]modules.ExplorerBlockData, 0),
```
DavidVorick: this also applies to slices, the default is nil, but append() can accept nil values

modules/blockexplorer/data.go
```diff
@@ -0,0 +1,66 @@
...
+	totalCoins := uint64(0)
+	coinbase := types.InitialCoinbase
+	minCoinbase := types.MinimumCoinbase
+	for i := types.BlockHeight(0); i < be.blockchainHeight; i++ {
+		totalCoins += coinbase
+		if coinbase > minCoinbase {
+			coinbase--
+		}
+	}
+	return types.NewCurrency64(totalCoins).Mul(types.SiacoinPrecision)
```
DavidVorick: I find it interesting that you would multiply this value by SiacoinPrecision when that precision is never shown to the user nor used in the calculation

modules/blockexplorer/data.go
```diff
@@ -0,0 +1,66 @@
...
+	"errors"
+
+	"github.com/NebulousLabs/Sia/build"
+	"github.com/NebulousLabs/Sia/modules"
+	"github.com/NebulousLabs/Sia/types"
+)
+
+// Calculates the total number of coins that have ever been created
+// from the bockheight
+func (be *BlockExplorer) totalCurrency() types.Currency {
```
DavidVorick: It's probably worth having this function accept a blockheight as input, and then not being of type 'BlockExplorer'

DavidVorick: `func totalCurrency(height types.BlockHeight) types.Currency {`

modules/blockexplorer/data.go
```diff
@@ -0,0 +1,66 @@
...
+
+// Returns a partial slice of our stored data on the blockchain. Data
+// obtained from consensus updates
+func (be *BlockExplorer) BlockInfo(start types.BlockHeight, finish types.BlockHeight) ([]modules.ExplorerBlockData, error) {
+	lockID := be.mu.RLock()
+	defer be.mu.RUnlock(lockID)
+
+	// Error checking on the given range
+	if start > finish {
+		return nil, errors.New("The start block must be higher than the end block")
```
DavidVorick: errors should have minimal punctuation, start with lower case letters and don't end with periods.

`the .. end block`

modules/blockexplorer/data.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+	if finish > be.blockchainHeight+1 {
+		return nil, errors.New("Cannot get info on a block higher than the blockchain")
```
DavidVorick: ditto

modules/blockexplorer/update.go
```diff
@@ -1,23 +1,61 @@
...
+// well as the current block height
 func (be *BlockExplorer) ReceiveConsensusSetUpdate(cc modules.ConsensusChange) {
-	// Do lock stuff here
 	lockID := be.mu.Lock()
 	defer be.mu.Unlock(lockID)
 
-	// Update the current block, and block height
+	// Modify the currency value
+	for _, diff := range cc.SiacoinOutputDiffs {
+		if diff.Direction == true {
```
DavidVorick: `if diff.Direction == modules.DiffApply`, 'true' doesn't tell us anything about what it means

modules/blockexplorer/update.go
```diff
@@ -1,23 +1,61 @@
...
+			be.currencySent = be.currencySent.Add(diff.SiacoinOutput.Value)
+		}
+	}
+
+	// Modify the number of file contracts and their values
+	for _, diff := range cc.FileContractDiffs {
+		if diff.Direction == true {
```
DavidVorick: ditto

modules/blockexplorer/update.go
```diff
@@ -1,23 +1,61 @@
...
+			be.fileContracts += 1
+			be.fileContractCost = be.fileContractCost.Add(diff.FileContract.Payout)
+		} else {
+			be.fileContracts -= 1
+			be.fileContractCost = be.fileContractCost.Sub(diff.FileContract.Payout)
+		}
```
DavidVorick: it would be interesting to have both a 'currently open' and an 'ever open' value.

DavidVorick:
merge conflicts again.

Sorry that this PR has been through so much, we're nearly done with the crazy pace of changing things for the near future. Had to address a lot of the immediate problems people were having.

doc/API.md
```diff
@@ -677,3 +677,76 @@ struct {
...
+Block Explorer
+--------------
+
+Queries:
+
+* /blockexplorer/blockchain
+* /blockexplorer/current
+* /blockexplorer/siacoins
+* /blockexplorer/filecontracts
+* /blockexplorer/blockdata
```
DavidVorick: need to update this

DavidVorick:
how is the progress on this pr?

triazo:
Don't know why the build is failing, my best guess is some networking thing specific to travis. Aside from that we should be good to go unless you see any other feature requests.

DavidVorick:
TestHostAnnounce just occasionally fails for some reason. We should just disable it. (we'll do that in another pr)

api/blockexplorer.go
DavidVorick: bump

doc/API.md
```diff
@@ -677,3 +677,79 @@ struct {
...
+struct {
+       Height int
+       Block  Types.Block
+       Target []byte
+       CurrencySent  int
+       TotalCurrency int
+       ActiveContractCount int
+       ActiveContractcosts int
+	   TotalContractCount int
+       TotalContractcosts int
```
DavidVorick: tabbing

modules/blockexplorer/update.go
```diff
@@ -1,23 +1,65 @@
...
 	lockID := be.mu.Lock()
 	defer be.mu.Unlock(lockID)
 
-	// Update the current block, and block height
+	// Modify the number of file contracts and how much they costed
+	for _, diff := range cc.FileContractDiffs {
+		// Diff direction is a bool representing if the file
+		// contract is new or getting removed. True signifies
+		// that it is a new file contract
+		if diff.Direction == true {
```
DavidVorick: this needs to be changed to `modules.DiffApply` (instead of `true`)

DavidVorick:
there's just one final nitpick

PR #592, opened 2015-06-14
Add Filesize to FileInfo

### LINKED ISSUE #589: FileInfo should include filesize
Right now there's no way to see the size of a file before downloading it. This is especially a problem when loading a shared file.

lukechampine:
Fixes #589

I'll add this to the UI file view soon.

### DIFF
--- a/modules/renter/files.go
+++ b/modules/renter/files.go
@@ -100,11 +100,13 @@ func (f *file) Nickname() string {
 func (f *file) Filesize() uint64 {
 	lockID := f.renter.mu.RLock()
 	defer f.renter.mu.RUnlock(lockID)
-	if len(f.Pieces) == 0 {
-		return 0
-	}
 	// TODO: this will break when we switch to erasure coding.
-	return f.Pieces[0].Contract.FileSize
+	for i := range f.Pieces {
+		if f.Pieces[i].Contract.FileSize != 0 {
+			return f.Pieces[i].Contract.FileSize
+		}
+	}
+	return 0
 }
 
 // Repairing returns whether or not the file is actively being repaired.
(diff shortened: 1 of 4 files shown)


siac/rentercmd.go
```diff
@@ -134,6 +135,16 @@ func renterfilesdownloadcmd(nickname, destination string) {
...
 
+// filesize returns a string that displays a filesize in human-readable units.
+func filesizeUnits(size uint64) string {
+	if size == 0 {
+		return "0 B"
+	}
+	sizes := []string{"B", "KB", "MB", "GB", "TB", "PB", "EB", "ZB", "YB"}
+	i := int(math.Log10(float64(size)) / 3)
+	return fmt.Sprintf("%.*f %s", i, float64(size)/math.Pow10(3*i), sizes[i])
+}
```
lukechampine: example output:
```
      31.5 KB
      2.83 MB
     42.66 MB
     3.152 GB
      3.51 MB
   283.320 GB
   351.193 GB
    4.2664 TB
```

DavidVorick: Not sure that we ever need more than 2 digits of precision. Having 4 follow the TB seems a bit silly.

It's okay for now though

modules/renter/files.go
```diff
@@ -96,6 +96,19 @@ func (f *file) Nickname() string {
 	return f.Name
 }
 
+// Filesize returns the size of the file.
+func (f *file) Filesize() uint64 {
+	lockID := f.renter.mu.RLock()
+	defer f.renter.mu.RUnlock(lockID)
+	// TODO: this will break when we switch to erasure coding.
```
DavidVorick: it'll break when we start adding multiple files to a single file contract as well.

PR #597, opened 2015-06-15
Support for siafund API calls in siac

lukechampine:
Plus the addition of the `SpendSiafunds` API call.

The calls should be correct but I haven't verified that they work as expected. I imagine it would be easy to test them with the .siakey files using testing constants, yes?

btw, if you assign PRs they'll show up when you click the "Pull requests" link the header. Maybe we should start doing that.


DavidVorick:
I like the idea of assigning pull requests.

That's the reason the types package has keys now, very easy for you to test the siac commands manually by loading those keys into the wallet and then spending from them.

lukechampine:
Points of confusion:
- Siafunds can't be sent to siacoin addresses, and there's no way for siad to check whether you're sending to a siacoin address by mistake. Need to educate users on what siafund addresses are and how to get them.
- No zero-conf, and other weirdness -- need to mine a block to clear up certain problems.

DavidVorick:
how badly would it break things if I added a checksum and address type right now?

api/api.go
```diff
@@ -87,6 +87,7 @@ func (srv *Server) initAPI(addr string) {
 	handleHTTPRequest(mux, "/wallet/send", srv.walletSendHandler)
 	handleHTTPRequest(mux, "/wallet/status", srv.walletStatusHandler)
 	handleHTTPRequest(mux, "/wallet/siafunds/balance", srv.walletSiafundsBalanceHandler)
+	handleHTTPRequest(mux, "/wallet/siafunds/spend", srv.walletSiafundsSpendHandler)
```
DavidVorick: bad that it's `wallet/send` and `wallet/siafunds/spend`

DavidVorick: I think I typically call the functions `spendCoins` in the interfaces, maybe it's best to keep parity with that.

At the same time, you really aren't spending them you are sending them. There's not any sort of contract recognized by the blockchain that you'll get something in return for doing so.

api/wallet.go
```diff
@@ -15,6 +16,37 @@ type WalletSiafundsBalance struct {
...
+	var addrBytes []byte
+	_, err = fmt.Sscanf(addrStr, "%x", &addrBytes)
+	if err != nil {
+		return
+	}
+	if copy(addr[:], addrBytes) != len(addr) {
+		err = errors.New("Malformed coin address")
+	}
+	return
+}
```
DavidVorick: maybe a type should be added to the return values. 'this is a siacoin addr' 'this is a siafund addr', etc.

DavidVorick: the return value type would of course be recognized by a prefix that we append

DavidVorick: then again, maybe not. I think it's premature to do that.

lukechampine:
it wouldn't be terrible, but I think it would make more sense to merge this first

api/api.go
lukechampine: I was surprised it wasn't `SendSiafunds`. That seemed more natural.

api/wallet.go
lukechampine: In this example it's actually nice that they're the same type, because I can reuse the scan function for both siacoins and siafunds.

api/api.go
DavidVorick: should we change it?

PR #599, opened 2015-06-16
add checksums to addresses

DavidVorick:
wasn't too bad. I'm pretty sure I got everything too, no human should ever look at or input an address without a checksum.

### DIFF
--- a/siag/keys.go
+++ b/siag/keys.go
@@ -195,7 +195,7 @@ func siag(*cobra.Command, []string) {
 		return
 	}
 
-	fmt.Printf("Keys created for address: %s\n", unlockConditions.UnlockHash().String())
+	fmt.Printf("Keys created for address: %v\n", unlockConditions.UnlockHash())
 	fmt.Printf("%v file(s) created. KEEP THESE FILES. To spend money from this address, you will need at least %v of the files.\n", config.Siag.TotalKeys, config.Siag.RequiredKeys)
 }
 
@@ -215,7 +215,7 @@ func printKeyInfo(filename string) error {
 	}
 
 	fmt.Printf("Found a key for a %v of %v address.\n", kp.UnlockConditions.SignaturesRequired, len(kp.UnlockConditions.PublicKeys))
-	fmt.Printf("The address is: %s\n", kp.UnlockConditions.UnlockHash().String())
+	fmt.Printf("The address is: %v\n", kp.UnlockConditions.UnlockHash())
 	return nil
 }
 
--- a/siag/keys.go
+++ b/siag/keys.go
@@ -195,7 +195,7 @@ func siag(*cobra.Command, []string) {
 		return
 	}
 
-	fmt.Printf("Keys created for address: %x\n", unlockConditions.UnlockHash())
+	fmt.Printf("Keys created for address: %s\n", unlockConditions.UnlockHash().String())
 	fmt.Printf("%v file(s) created. KEEP THESE FILES. To spend money from this address, you will need at least %v of the files.\n", config.Siag.TotalKeys, config.Siag.RequiredKeys)
 }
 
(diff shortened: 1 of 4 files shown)


siag/keys.go
```diff
@@ -215,7 +215,7 @@ func printKeyInfo(filename string) error {
 	}
 
 	fmt.Printf("Found a key for a %v of %v address.\n", kp.UnlockConditions.SignaturesRequired, len(kp.UnlockConditions.PublicKeys))
-	fmt.Printf("The address is: %x\n", kp.UnlockConditions.UnlockHash())
+	fmt.Printf("The address is: %s\n", kp.UnlockConditions.UnlockHash().String())
```
lukechampine: the `.String()` call is implicit if you use `%v`

types/signatures.go
```diff
@@ -407,17 +417,72 @@ func (uh UnlockHash) MarshalJSON() ([]byte, error) {
...
+	var checksum []byte
+	_, err := fmt.Sscanf(string(b[1:1+crypto.HashSize*2]), "%x", &byteUnlockHash)
+	if err != nil {
+		return err
+	}
+	// Decode the checksum.
+	_, err = fmt.Sscanf(string(b[1+crypto.HashSize*2:1+crypto.HashSize*2+UnlockHashChecksumSize*2]), "%x", &checksum)
+	if err != nil {
+		return err
+	}
```
lukechampine: probably cleaner to split them up beforehand:
```go
// remove ""
b = b[1:len(b)-2]
// split unlock hash and checksum
uhBytes := b[:crypto.HashSize*2]
chkBytes := b[crypto.HashSize*2:]
// Decode...
```
It's safe to use a trailing `:` because the length has already been checked at the top of the function.

siag/keys.go
DavidVorick: I thought it might be, but didn't test it.

types/signatures.go
```diff
@@ -407,17 +417,72 @@ func (uh UnlockHash) MarshalJSON() ([]byte, error) {
...
+
+	// Verify the checksum - leave uh as-is unless the checksum is valid.
+	var unlockHash UnlockHash
+	copy(unlockHash[:], byteUnlockHash)
+	expectedChecksum := crypto.HashObject(unlockHash)
```
lukechampine: you can just use `crypto.HashBytes` on the raw `byteUnlockHash`

types/signatures.go
```diff
@@ -407,17 +417,72 @@ func (uh UnlockHash) MarshalJSON() ([]byte, error) {
...
+	// Decode the unlock hash.
 	var byteUnlockHash []byte
-	_, err := fmt.Sscanf(string(b[1:crypto.HashSize*2+1]), "%x", &byteUnlockHash)
+	var checksum []byte
+	_, err := fmt.Sscanf(strUH[:crypto.HashSize*2], "%x", &byteUnlockHash)
+	if err != nil {
+		return err
+	}
+	// Decode the checksum.
+	_, err = fmt.Sscanf(strUH[crypto.HashSize*2:crypto.HashSize*2+UnlockHashChecksumSize*2], "%x", &checksum)
```
lukechampine: lots of duplicated code here. Either `UnmarshalJSON` should call `LoadString`, or the other way around.

Really we should implement a `Scan` method so that we can use `fmt.Scan` directly, but it's more trouble than it's worth right now.

DavidVorick:
all better

lukechampine:
what happens to the addresses generated by siag?

DavidVorick:
they have checksums too now

DavidVorick:
the ones that were already generated are now invalid, but only their human-readable forms.

You can get new addresses though by running `siag keyinfo $(keyfile)`

PR #600, opened 2015-06-16
Changed instances of "spendCoins" to "sendCoins"

### LINKED ISSUE #598: Spend Coins vs. Send Coins
in some places, we 'spend coins', and in others, we 'send coins', but they mean the same thing.

In all places, we should use the same vocabulary. My vote is to 'send coins'

Mingling94:
Fix #598 through simple find & replace


DavidVorick:
lol why is this a merge conflict?

I think you forgot to pull the recent changes to master.

PR #603, opened 2015-06-16
keep a loadable backup wallet in the home folder

### LINKED ISSUE #594: Safer Wallets
Wallets should be saved in both homedir and the current place they are saved. When loading, both need to be checked, and if they are not the same an error should be displayed and the program should close.

This prevents wallet files from accidentally getting deleted or overwritten.

DavidVorick:
wallet gets put in the homedir as backup.

If no wallet is found in the regular place, the homedir is checked.

fixes #594

### DIFF
--- a/modules/wallet/persist.go
+++ b/modules/wallet/persist.go
@@ -3,6 +3,7 @@ package wallet
 import (
 	"path/filepath"
 
+	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/encoding"
 	"github.com/NebulousLabs/Sia/modules"
@@ -95,11 +96,12 @@ func (w *Wallet) loadKeys(savedKeys []savedKey) error {
 // load reads the contents of a wallet from a file.
 func (w *Wallet) load() error {
 	var savedKeys []savedKey
-	err := encoding.ReadFile(filepath.Join(w.saveDir, "wallet.backup"), &savedKeys)
-	if err != nil {
+	err := encoding.ReadFile(filepath.Join(persist.HomeFolder, modules.WalletDir, "wallet.dat"), &savedKeys)
+	// never load from the home folder during testing
+	if build.Release == "testing" || err != nil {
 		// try loading the backup
 		// TODO: display/log a warning?
-		err = encoding.ReadFile(filepath.Join(persist.HomeFolder, modules.WalletDir, "wallet.dat"), &savedKeys)
+		err = encoding.ReadFile(filepath.Join(w.saveDir, "wallet.backup"), &savedKeys)
 		if err != nil {
 			return err
 		}
@@ -112,7 +114,8 @@ func (w *Wallet) load() error {
 	// Load the siafunds file, which is intentionally called 'outputs.dat'.
 	var siafundAddresses []types.UnlockHash
 	err = encoding.ReadFile(filepath.Join(w.saveDir, "outputs.backup"), &siafundAddresses)
-	if err != nil {
+	// never load from the home folder during testing
+	if build.Release == "testing" || err != nil {
 		// try loading the backup
 		// TODO: display/log a warning?
 		err = encoding.ReadFile(filepath.Join(persist.HomeFolder, modules.WalletDir, "outputs.dat"), &siafundAddresses)
--- a/modules/wallet/persist.go
+++ b/modules/wallet/persist.go
@@ -16,14 +16,6 @@ type savedKey struct {
 	Visible          bool
 }
 
-// legacySavedKey preserves compatibility with the other beta wallets. After
-// the full currency is launched, this struct and the related code can be
-// discarded.
-type legacySavedKey struct {
-	SecretKey        crypto.SecretKey
-	UnlockConditions types.UnlockConditions
-}
-
 // save writes the contents of a wallet to a file.
 func (w *Wallet) save() error {
 	// Convert the key map to a slice.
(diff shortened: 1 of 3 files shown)


lukechampine:
`make test` definitely shouldn't overwrite your existing home folder wallet.
However, I found that running `make test-long` twice would always produce an error. Loading the "backup" wallet first seemed to solve the problem, but I feel there's a better way.

lukechampine:
hmm. Still not working. Maybe we just have to not save to the homedir at all during testing.

DavidVorick:
No, use some `if build.Release == "testing" { HomeFolder = /tmp/SiaHome...`
`if build.Release == "dev" { HomeFolder = /home/.config/SiaDev/... `

lukechampine:
The core problem is that tests can't share the same wallet file. It causes a whole host of errors.

DavidVorick:
I guess I would move the `if build.Release == "testing"` inside of the load and save functions within the wallet then. Definitely add the change for dev, that should be a different folder for sure.

It's kind of gross, but I don't see another way around it. Maybe you have a better idea?

DavidVorick:
we should also avoid using the home folder during 'dev'

lukechampine:
I thought you wanted a `/home/.config/SiaDev/` as stated above...

modules/wallet/persist.go
```diff
@@ -14,14 +17,6 @@ type savedKey struct {
 	Visible          bool
 }
 
-// legacySavedKey preserves compatibility with the other beta wallets. After
-// the full currency is launched, this struct and the related code can be
-// discarded.
-type legacySavedKey struct {
-	SecretKey        crypto.SecretKey
-	UnlockConditions types.UnlockConditions
-}
```
DavidVorick: legacy keys are removed?

I guess it should be okay, but it does seem a bit weird to only support them for 2 weeks.

modules/wallet/persist.go
```diff
@@ -35,8 +30,8 @@ func (w *Wallet) save() error {
 	if err != nil {
 		return err
 	}
-	// Overwrite the wallet file.
-	err = encoding.WriteFile(filepath.Join(w.saveDir, "wallet.dat"), keySlice)
+	// Save to home.
+	err = encoding.WriteFile(filepath.Join(persist.HomeFolder, modules.WalletDir, "wallet.dat"), keySlice)
```
DavidVorick: saving to home needs to be wrapped in a check

modules/wallet/persist.go
```diff
@@ -56,7 +51,8 @@ func (w *Wallet) save() error {
 	if err != nil {
 		return err
 	}
-	err = encoding.WriteFile(filepath.Join(w.saveDir, "outputs.dat"), siafundSlice)
+	// Save to home.
+	err = encoding.WriteFile(filepath.Join(persist.HomeFolder, modules.WalletDir, "outputs.dat"), siafundSlice)
```
DavidVorick: ditto

modules/wallet/persist.go
DavidVorick: I don't think I support this remove.

I know it's a bit gross but we need to maintain legacy support for at least a few versions.

persist/persist.go
```diff
@@ -0,0 +1,26 @@
...
+	}
+
+	switch build.Release {
+	case "testing":
+		return filepath.Join(build.SiaTestingDir, "home")
+	case "dev":
+		return filepath.Join(home, ".config", "SiaDev")
+	default:
+		return filepath.Join(home, ".config", "Sia")
+	}
```
DavidVorick: oh. I like this way of doing it. Do we need the other 'if' statements in this case?

DavidVorick: sorry for being all over the place on this pr

modules/wallet/persist.go
```diff
@@ -108,12 +90,39 @@ func (w *Wallet) load() error {
...
-	// Load the siafunds file, which is intentionally called 'outputs.dat'.
-	var siafundAddresses []types.UnlockHash
-	err = encoding.ReadFile(filepath.Join(w.saveDir, "outputs.dat"), &siafundAddresses)
+	return nil
+}
+
+// load reads the contents of a wallet from a file.
+func (w *Wallet) load() error {
+	var savedKeys []savedKey
+	err := encoding.ReadFile(filepath.Join(persist.HomeFolder, modules.WalletDir, "wallet.dat"), &savedKeys)
```
DavidVorick: Does it make more sense to read from the home folder as primary or as backup?

What can we assume about people running Sia from multiple locations?

I guess home folder as primary makes sense.

lukechampine: it's a toss up. You don't want to load the backup first because it would confuse users who replace their home folder's wallet.dat and expect it to be loaded. On the other hand, if you load the home folder first, then when you plug your Sia USB stick into someone else's computer, you'll get their wallet instead of the one on the stick.

modules/wallet/persist.go
lukechampine: I thought the legacy wallets were from before the fullnet?

modules/wallet/persist.go
DavidVorick: I think the next iteration of the wallet (which will introduce 'accounts' -> one file per account) will solve a lot of these problems.

modules/wallet/persist.go
DavidVorick: oh you are correct, I just looked at our history.

nvm then it's okay

DavidVorick:
I'll call this good enough. I'm guessing Josh will be overhauling the wallet.

PR #605, opened 2015-06-17
disable faulty host annoucne test

DavidVorick:


lukechampine:
I would lol if that build failed

DavidVorick:
I was actually pretty close to disabling the wrong test lol

PR #606, opened 2015-06-17
Seperate the blockexplore into a seperate siae binary

triazo:
Branch is on top of my other pull request.

### DIFF
--- a/siae/daemon_test.go
+++ b/siae/daemon_test.go
@@ -1,0 +1,38 @@
+package main
+
+import (
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/NebulousLabs/Sia/build"
+)
+
+// TestStartDaemon probes the startDaemon function.
+func TestStartDaemon(t *testing.T) {
+	testDir := build.TempDir("siad", "TestStartDaemon")
+	config.Siad.NoBootstrap = false
+	config.Siad.APIaddr = "localhost:45170"
+	config.Siad.RPCaddr = ":45171"
+	config.Siad.HostAddr = ":45172"
+	config.Siad.SiaDir = testDir
+	go func() {
+		err := startDaemon()
+		if err != nil {
+			t.Error(err)
+		}
+	}()
+
+	// Wait until the server has started, and then send a kill command to the
+	// daemon.
+	<-started
+	time.Sleep(250 * time.Millisecond)
+	resp, err := http.Get("http://localhost:45170/daemon/stop")
+	if err != nil {
+		t.Fatal(err)
+	}
+	if resp.StatusCode != http.StatusOK {
+		t.Fatal(resp.StatusCode)
+	}
+	resp.Body.Close()
+}
--- a/siae/daemon_test.go
+++ b/siae/daemon_test.go
@@ -12,9 +12,8 @@ import (
 func TestStartDaemon(t *testing.T) {
 	testDir := build.TempDir("siad", "TestStartDaemon")
 	config.Siad.NoBootstrap = false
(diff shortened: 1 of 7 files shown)


DavidVorick:
need to resolve the merge conflicts

siae/daemon_test.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+// TestStartDaemon probes the startDaemon function.
+func TestStartDaemon(t *testing.T) {
+	testDir := build.TempDir("siad", "TestStartDaemon")
+	config.Siad.NoBootstrap = false
+	config.Siad.APIaddr = "localhost:45170"
+	config.Siad.RPCaddr = ":45171"
+	config.Siad.HostAddr = ":45172"
+	config.Siad.SiaDir = testDir
```
DavidVorick: @lukechampine: how do we deal with port testing again?

I think you can just use ":0" for all of them

siae/main.go
```diff
@@ -0,0 +1,72 @@
...
+		Short: "Print version information",
+		Long:  "Print version information about the Sia Daemon",
+		Run:   versionCmd,
+	})
+
+	// Set default values, which have the lowest priority.
+	root.PersistentFlags().BoolVarP(&config.Siad.NoBootstrap, "no-bootstrap", "n", false, "disable bootstrapping on this run")
+	root.PersistentFlags().StringVarP(&config.Siad.APIaddr, "api-addr", "a", "localhost:9980", "which host:port the API server listens on")
+	root.PersistentFlags().StringVarP(&config.Siad.RPCaddr, "rpc-addr", "r", ":9981", "which port the gateway listens on")
+	root.PersistentFlags().StringVarP(&config.Siad.SiaDir, "sia-directory", "d", "", "location of the sia directory")
```
DavidVorick: Seems odd to me to duplicate so much code between siad and siae. I wonder if they should be the same binary.

siae/main_test.go
```diff
@@ -0,0 +1,54 @@
...
+	// Try running and closing an instance of siad.
+	os.Args = []string{
+		"siad",
+		"-n",
+		"-a",
+		"localhost:45150",
+		"-r",
+		"localhost:45151",
+		"-H",
+		"localhost:45152",
```
DavidVorick: ":0" is probably okay.

triazo:
It does seem a bit odd to have two top level binaries that are so similar. Due to the import tree, you don't even save in size of the binary by splitting the blockexplorer out. Alternatively we could have a command line flag to put it into blockexplorer mode, or generalize that to flags that toggle every module from the default. Also, we could make a hardlink to the binary and switch based on arg0. The amount of things we need to switch is small and I think there are numerous ways to prevent code replication.

siae/daemon_test.go
lukechampine: you should be able to use ":0", but you'll need to call various methods to discover which port the OS has assigned you. See other testing code for examples

DavidVorick:
I would like to merge this. We can postpone the discussion about whether or not 'siae' should exist until we have an alternative (probably a stack of flags in 'siad' that enables and disables all of the modules).

triazo:
Updated and rebased. Should merge now.

siae/main_test.go
```diff
@@ -0,0 +1,52 @@
...
+	testDir := build.TempDir("siad", "TestMain")
+
+	// Try running and closing an instance of siad.
+	os.Args = []string{
+		"siad",
+		"-n",
+		"-a",
+		"localhost:45150",
+		"-r",
+		"localhost:45151",
```
DavidVorick: these should be :0 I believe

PR #608, opened 2015-06-17
Miner cleanup

DavidVorick:
CPU miner is now single threaded.

The interface got broken into multiple interfaces.

FindBlock no longer also submits the block to the consensus set, which just seems like a more intuitive way of doing things.

There's a logger for the miner now.

Miner got safelocks.

Lots of code de-duplication.


modules/miner.go
```diff
@@ -5,68 +5,69 @@ import (
...
-}
+const (
+	MinerDir = "miner"
+)
 
-// The Miner interface provides access to mining features.
-type Miner interface {
+// A BlockBuilder contains functions that can interface with external miners,
+// providing and receiving blocks that have experienced nonce grinding.
+type BlockManager interface {
```
lukechampine: wrong docstring name

modules/miner.go
```diff
@@ -5,68 +5,69 @@ import (
...
+	BlockManager
+	CPUMiner
+
+	// BlocksMined returns the number of blocks and stale blocks that have been
+	// mined using this miner.
+	BlocksMined() (goodBlocks, staleBlocks int)
+
+	// MinerNotify is a channel to inform subscribers of when the miner has
+	// updated. This is primarily used for synchronization during testing.
+	MinerNotify() <-chan struct{}
```
lukechampine: I am a little wary of this interface separation, given that no type implements just `BlockManager` or just `CPUMiner`. It may make more sense to relegate the CPU miner to a testing package or subfolder, given that it's only used for testing.

DavidVorick: it's also used on devnets and will probably be used on the testnet. It's definitely important to have for more than just test-package reasons.

I only broke the interfaces up because it made them a lot more readable.

DavidVorick: We could put these two outer functions in each, and then the tester's would only need the CPUMiner and not the BlockManager, but I don't think it would make things that much neater.

PR #611, opened 2015-06-18
Encoding safety

lukechampine:
Patch vulnerabilities in the decoding functions. Variable length types will no longer be allocated if they exceed 4MB. In addition, if the decoder will not decode more than 10MB while decoding a single type.

### DIFF
--- a/encoding/marshal.go
+++ b/encoding/marshal.go
@@ -220,7 +220,7 @@ func (d *Decoder) readN(n int) []byte {
 // readPrefix reads a length-prefixed byte slice and panics if the read fails.
 func (d *Decoder) readPrefix() []byte {
 	// TODO: what should maxlen be?
-	b, err := ReadPrefix(d.r, 1<<32)
+	b, err := ReadPrefix(d.r, maxSliceLen)
 	if err != nil {
 		panic(err)
 	}
--- a/encoding/marshal.go
+++ b/encoding/marshal.go
@@ -61,7 +61,8 @@ type Encoder struct {
 }
 
 const (
-	maxSliceLen = 4 * 1024 * 1024 // 4 MB
+	maxDecodeLen = 10 * 1024 * 1024 // 10 MB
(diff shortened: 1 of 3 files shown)


encoding/marshal.go
```diff
@@ -179,6 +184,16 @@ func WriteFile(filename string, v interface{}) error {
 // A Decoder reads and decodes values from an input stream.
 type Decoder struct {
 	r io.Reader
+	n int
+}
+
+func (d *Decoder) Read(p []byte) (int, error) {
```
DavidVorick: need docstring

encoding/marshal_test.go
```diff
@@ -133,8 +134,27 @@ func TestDecode(t *testing.T) {
 
 	// unknown type
 	err = Unmarshal([]byte{1, 2, 3}, new(map[int]int))
-	if err == nil {
-		t.Error("expected error, got nil")
+	if err == nil || err.Error() != "could not decode type map[int]int: unknown type" {
```
DavidVorick: I wish there was a better way to verify error equivalence.

encoding/marshal_test.go
```diff
@@ -133,8 +134,27 @@ func TestDecode(t *testing.T) {
...
+	if err == nil || err.Error() != "could not decode type []uint8: slice is too large" {
+		t.Error("expected large slice error, got", err)
+	}
+
+	// many small slices (total larger than maxDecodeLen)
+	bigSlice := strings.Split(strings.Repeat("0123456789abcdefghijklmnopqrstuvwxyz", (maxSliceLen/16)-1), "0")
+	err = Unmarshal(Marshal(bigSlice), new([]string))
+	if err == nil || err.Error() != "could not decode type []string: encoded type exceeds size limit" {
+		t.Error("expected large slice error, got", err)
 	}
```
DavidVorick: Is there a test to check the global limit?

PR #613, opened 2015-06-20
Siac units

### LINKED ISSUE #593: Add Units to Siac
I believe all units in siac right now are base units, we should add support for S, KS, MS, GS to be recognized.

lukechampine:
Also deprecate the use of exponential format on the backend.

fixes #593

### DIFF
--- a/api/wallet.go
+++ b/api/wallet.go
@@ -1,7 +1,6 @@
 package api
 
 import (
-	"fmt"
 	"math/big"
 	"net/http"
 	"strings"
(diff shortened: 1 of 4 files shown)


api/wallet.go
```diff
@@ -16,21 +15,14 @@ type WalletSiafundsBalance struct {
...
-	// exponential format
-	if strings.ContainsAny(amountStr, "Ee") {
-		amountRat := new(big.Rat)
-		_, err = fmt.Sscan(amountStr, amountRat)
-		if err != nil {
-			return
-		}
-		amount = types.NewCurrency(new(big.Int).Div(amountRat.Num(), amountRat.Denom()))
-		return
+func scanAmount(amount string) (types.Currency, bool) {
```
DavidVorick: why a bool?

siac/walletcmd.go
```diff
@@ -28,8 +54,19 @@ var (
...
+	pS (pico,  10^-12 SC)
+	nS (nano,  10^-9 SC)
+	uS (micro, 10^-6 SC)
+	mS (milli, 10^-3 SC)
+	SC
+	KS (kilo, 10^3 SC)
+	MS (mega, 10^6 SC)
+	GS (giga, 10^9 SC)
+	TS (tera, 10^12 SC)
+If no unit is supplied, hastings (smallest possible unit, 10^-27 SC) will be assumed.`,
```
DavidVorick: a hasting is 10^-24 S

PR #621, opened 2015-06-22
Host Scanning

DavidVorick:
Host scanning is now done by a thread pool, instead of all being done at once. The timeout for host response has been reduced from 10 seconds to 5 seconds, to facilitate speed during startup.

There are 25 threads that pull hosts out of channels. At startup, the hosts are added to the channels. Then, over time, hosts are selected randomly (though to the tune of 250-750 at a time) and placed into the channels. This happens every 3.5 hours on average.

If a host fails to respond 25 times in a row, the host is kicked from the database. This is different from before where the host would just get kicked immediately. It does mean that there is some pollution for bad hostnames (which do exist in the database) and that those names will take a few days to get cleared out.

After responding positively, the host can fail to respond up to 50 times in a row before getting kicked.

After failing to respond once, the host is removed from the set of active hosts. This is different from previously, where a host was allowed to not respond up to 10 times before being pulled from the set of active hosts.

Overall, I think this represents significant improvements to our host finding algorithm.

### DIFF
--- a/modules/hostdb/scan.go
+++ b/modules/hostdb/scan.go
@@ -17,20 +17,26 @@ import (
 )
 
 const (
-	DefaultScanSleep = 4 * time.Hour
-	MaxScanSleep     = 8 * time.Hour
+	DefaultScanSleep = 2*time.Hour + 18*time.Minute
+	MaxScanSleep     = 6 * time.Hour
 	MinScanSleep     = 1 * time.Hour
 
-	MaxActiveHosts              = 200
-	InactiveHostCheckupQuantity = 100
+	MaxActiveHosts              = 500
+	InactiveHostCheckupQuantity = 250
 
 	maxSettingsLen = 1024
+
+	hostRequestTimeout = 15 * time.Second
+
+	// scanningThreads is the number of threads that will be probing hosts for
+	// their settings and checking for reliability.
+	scanningThreads = 4
 )
 
 var (
-	ActiveReliability   = types.NewCurrency64(20)
-	InactiveReliability = types.NewCurrency64(10)
-	UnreachablePenalty  = types.NewCurrency64(1)
+	MaxReliability     = types.NewCurrency64(50) // Given the scanning defaults, about 1 week of survival.
+	DefaultReliability = types.NewCurrency64(20) // Given the scanning defaults, about 3 days of survival.
+	UnreachablePenalty = types.NewCurrency64(1)
 )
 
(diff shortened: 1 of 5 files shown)


DavidVorick:
build is fixed

modules/hostdb/scan.go
```diff
@@ -62,46 +68,50 @@ func (hdb *HostDB) decrementReliability(addr modules.NetAddress, penalty types.C
...
-		return encoding.ReadObject(conn, &settings, maxSettingsLen)
-	}()
-
-	// Now that network communication is done, lock the hostdb to modify the
-	// host entry.
-	id := hdb.mu.Lock()
-	defer hdb.mu.Unlock(id)
+func (hdb *HostDB) threadedProbeHosts() {
+	for hostEntry := range hdb.scanPool {
+		// Request settings from the queue'd host entry.
```
lukechampine: queued is a word lol

modules/renter/negotiate.go
```diff
@@ -110,7 +110,7 @@ func (r *Renter) negotiateContract(host modules.HostSettings, up modules.FileUpl
 	terms := modules.ContractTerms{
 		FileSize:      filesize,
 		Duration:      up.Duration,
-		DurationStart: height - 1,
+		DurationStart: height - 3,
```
lukechampine: what's this about?

modules/hostdb/scan.go
```diff
@@ -154,32 +165,35 @@ func (hdb *HostDB) threadedScan() {
...
+			// shuffled list and scan them.
 			n := InactiveHostCheckupQuantity
 			if len(random) < InactiveHostCheckupQuantity {
 				n = len(random)
 			}
 			for i := 0; i < n; i++ {
-				go hdb.threadedProbeHost(random[i])
+				go func() {
+					hdb.scanPool <- random[i]
+				}()
```
lukechampine: this appears often enough that it might be cleaner to separate it into a `hdb.addToScanPool` function. Then you wouldn't need to explain about blocking channels and such.

modules/renter/negotiate.go
DavidVorick: There's an off-by-one on the network, and there were a significant number of errors being returned about 'DurationStart is in the future', which means that renters and hosts weren't always perfectly synchronized.

This increases the tolerance for de-synchronization slightly.

modules/hostdb/scan.go
DavidVorick: ok

PR #622, opened 2015-06-22
v0.3.3.3 gateway tweaks

lukechampine:
The gateway now advertises itself as v0.3.3. Otherwise, v0.3.3 nodes will refuse to connect to it. This is a temporary measure to alleviate network splitting.

Nodes are pruned by `makeOutboundConnections` if they aren't reachable. If the connection fails for a different reason, the node is not pruned.

The `ExternalIP` function now times out after 3 seconds. This is the easiest way to prevent clients without Internet access from hanging. I'll implement a more robust solution when I tackle UPnP, which gives us an alternative way of discovering the external IP.

### DIFF
--- a/modules/gateway/peers.go
+++ b/modules/gateway/peers.go
@@ -239,8 +239,9 @@ func (g *Gateway) makeOutboundConnections() {
 			if err != nil || numPeers >= wellConnectedThreshold {
 				break
 			}
-			if g.Connect(addr) != nil {
-				// aggressively remove unresponsive nodes
+			err = g.Connect(addr)
+			// aggressively remove unresponsive nodes
+			if netErr, ok := err.(net.Error); ok && netErr.Timeout() {
 				id = g.mu.Lock()
 				g.removeNode(addr)
 				g.mu.Unlock(id)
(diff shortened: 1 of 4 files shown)


DavidVorick:
"only prune on timeout errors" -> seems good enough

modules/gateway/peers.go
```diff
@@ -168,7 +168,7 @@ func (g *Gateway) Connect(addr modules.NetAddress) error {
 		return err
 	}
 	// send our version
-	if err := encoding.WriteObject(conn, build.Version); err != nil {
+	if err := encoding.WriteObject(conn, "0.3.3"); err != nil {
```
DavidVorick: Based on how Bitcoin works, I think we can drop support for this after 95% of nodes have build.Version >= 0.3.3.2

lukechampine:
I feel pretty good about the pruning. Once people upgrade, which should slowly see unresponsive nodes being phased out of everyone's node list. And it should be easy to rejoin the network even if you were previously pruned.

modules/gateway/peers.go
lukechampine: I was actually just thinking that it would be cool to gather some version metrics, i.e. what percentage is running each version. I think I could write a test program to determine this pretty easily.

PR #623, opened 2015-06-22
Force announce

DavidVorick:
It probably didn't have to (sorry), but this PR builds on the other one I've got open.

This adds a force-announce api call, which is a response to the frustrated people whose ports are actually open but for some reason are failing our 'are your ports open?' internal test. Force-announce is only available directly in the api and in 'siac'.

I do believe that I implemented the siac command correctly.

### DIFF
--- a/modules/host/announce.go
+++ b/modules/host/announce.go
@@ -10,10 +10,14 @@ import (
 	"github.com/NebulousLabs/Sia/types"
 )
 
+const (
+	pingTimeout = 10 * time.Second
+)
+
 // ping establishes a connection to addr and then immediately closes it. It is
 // used to verify that an address is connectible.
 func ping(addr modules.NetAddress) bool {
-	conn, err := net.DialTimeout("tcp", string(addr), time.Second)
+	conn, err := net.DialTimeout("tcp", string(addr), pingTimeout)
 	if err != nil {
 		return false
 	}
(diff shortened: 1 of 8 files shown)


lukechampine:
a simply boolean flag would probably be better than two functions.

DavidVorick:
I disagree, I don't want normal announce to have a flag as input.

Siac can use a flag though

lukechampine:
how about we just always force? And if you aren't connectible, print a warning.
```
WARNING: your host port does not appear to be reachable. Please ensure you have forwarded port XXXX.
```

DavidVorick:
It pollutes the blockchain, and might give the user a false sense of 'oh well I probably did it right'.

modules/host/announce.go
```diff
@@ -67,3 +59,28 @@ func (h *Host) Announce() error {
...
+// Announce creates a host announcement transaction, adding information to the
+// arbitrary data, signing the transaction, and submitting it to the
+// transaction pool.
+func (h *Host) Announce() error {
+	// check that our address is reachable
+	lockID := h.mu.RLock()
+	addr := h.myAddr
+	h.mu.RUnlock(lockID)
+	if addr.Host() == "" {
+		return errors.New("can't announce without knowing external IP")
```
lukechampine: shouldn't this check be in `ForceAnnounce` as well? Or rather, moved to `announce`?

DavidVorick: I am not sure.

On one hand, --force means "do it I know what I'm doing". On the other, it's pretty hard to imagine that you are doing the right thing if your addr is `""`

lukechampine:
`Announce` only appears in the codebase twice (once in the API, once in a test). I think a boolean flag is fine as long as those two instances are properly commented.
```go
func (srv *Server) hostAnnounceHandler(w http.ResponseWriter, req *http.Request) {
        // Announce checks that the host is connectible before proceeding. The user
        // can override this check with the 'force' flag.
        force := req.FormValue("force") == "true"
	err := srv.host.Announce(force)
	if err != nil {
		writeError(w, err.Error(), http.StatusBadRequest)
		return
	}
	writeSuccess(w)
}
```

DavidVorick:
:(

modules/host/announce.go
DavidVorick: I think I want to defer to --force means "do it anyway", simply because you might be on a testnet or you might be curious or etc. etc. etc. So we should leave any sort of check in Announce instead of announce.

lukechampine:
haha I know, I know.
Also consider that if we add a new API call, we have to deprecate it later. If we're just adding a query value, it can be safely ignored by future versions.

DavidVorick:
api call will need to be deprecated anyway when we switch over to the api that aaron designed.

but still, I'm okay even with having a flag in the api call, just averse to having a function with a bool for input. I don't think that this is the last that we will see of 'Announce'

DavidVorick:
fine

lukechampine:
boolean args are definitely bad if you supply them as literals. `Announce(true)` would definitely be confusing. But if you use an explicit variable name I think it's okay.

DavidVorick:
build failed!

lukechampine:
fixed

PR #627, opened 2015-06-23
add a wallet merge tool

DavidVorick:
There is now a command to help you merge wallets together, which means you're not at risk of losing coins when deleting or moving around other wallet files. This is also useful for people mining on many computers at the same time.

Thanks to this feature, I was able to recover 4m of the 6m I lost. I now am not sure whether I actually lost 6m or just 4m. Regardless, a significant chunk is back.

### DIFF
--- a/modules/wallet/persist.go
+++ b/modules/wallet/persist.go
@@ -3,40 +3,33 @@ package wallet
 import (
 	"path/filepath"
 
-	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/encoding"
 	"github.com/NebulousLabs/Sia/types"
 )
 
+// a savedKey contains a single-signature key and all the tools needed to spend
+// outputs at its address.
 type savedKey struct {
 	SecretKey        crypto.SecretKey
 	UnlockConditions types.UnlockConditions
 	Visible          bool
 }
 
-// save writes the contents of a wallet to a file.
-func (w *Wallet) save() error {
-	// Convert the key map to a slice.
+// saveKeys saves the current set of keys known to the wallet to a file.
+func (w *Wallet) saveKeys(filepath string) error {
+	// Convert the key map to a slice and write to disk.
 	keySlice := make([]savedKey, 0, len(w.keys))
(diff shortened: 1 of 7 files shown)


modules/wallet/persist.go
```diff
@@ -129,3 +137,19 @@ func (w *Wallet) load() error {
...
+	err := encoding.ReadFile(filepath, &savedKeys)
+	if err != nil {
+		return err
+	}
+	err = w.loadKeys(savedKeys)
+	if err != nil {
+		return err
+	}
+	return nil
+}
```
lukechampine: seems to me like "Merge" is really just "load"

DavidVorick: initially it wasn't, but I think I changed load to work like Merge.

`load` also grabs outputs, but I can deduplicate a lot of this.

lukechampine:
how does this relate to eventual loading of multiple wallets?
would wallet merging be strictly superseded by multiple wallet support?

DavidVorick:
I believe it would

DavidVorick:
rebased to master, cleaned files up, and then fixed a siafund reporting bug.

modules/wallet/persist.go
```diff
@@ -45,12 +38,29 @@ func (w *Wallet) save() error {
 	// 'siafunds.dat' or something similar, people might think it's okay to
 	// delete their siafund keys, which is NOT okay. Instead of potentially
 	// having this confusion, I chose a less suggestive name.
-	err = encoding.WriteFile(filepath.Join(w.saveDir, "outputs.backup"), siafundSlice)
+	return encoding.WriteFile(filepath, siafundSlice)
```
lukechampine: need to move the comment

DavidVorick:
done

lukechampine:
I don't like implementing API calls that we know will soon be deprecated. Do we even plan on supporting this in UI? If not, why not just confine it so siac?

DavidVorick:
not really, but this is very important for the people who have lost funds in the same way that I did.

Also, it really eases some of the upgrading tension.

I feel strongly about this PR. Right now there are lots of people (especially multi-machine people) who would like some way to combine their wallets.

PR #628, opened 2015-06-24
wallet loads siafunds correctly

DavidVorick:
A small change was also made to the encoding package, when files get
written any tail data is now truncated.


DavidVorick:
muh printlns

whoops.

PR #630, opened 2015-06-24
fix a few uploading race conditions

DavidVorick:
There is still a race condition remaining when we save file pieces, because there we actually copy the whole file when passing it to save, and this happens even mid-upload.

The solution is either to add a lock to the uploading updater, or to exclude in-progress uploads from being saved. I'm in favor of the latter given our current design (uploads can't be resumed anyway), but neither can be easily implemented.

I'm just going to leave it for now, since the renter is going to be getting a pretty big facelift anyway.

### DIFF
--- a/modules/renter/files.go
+++ b/modules/renter/files.go
@@ -2,6 +2,7 @@ package renter
 
 import (
 	"errors"
+	"sync/atomic"
 
 	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/modules"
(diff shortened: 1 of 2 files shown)


modules/renter/files.go
```diff
@@ -61,9 +66,13 @@ func (f *file) Available() bool {
...
 	defer f.renter.mu.RUnlock(lockID)
 
+	// The loop uses an index instead of a range because range copies the piece
+	// to fresh data. Atomic operations are being concurrently performed on the
+	// piece, and the copy results in a race condition against the atomic
+	// operations. By removing the copying, the race condition is eliminated.
 	var active int
-	for _, piece := range f.Pieces {
-		if piece.Active {
+	for i := 0; i < len(f.Pieces); i++ {
```
lukechampine: `i := range f.Pieces`???

DavidVorick: you can do that? That's a thing?

lukechampine: it's a highly experimental feature of Go 12.0

PR #635, opened 2015-06-25
Save node list on close

lukechampine:
The node list wasn't being saved after removing nodes, so the gateway now calls `g.save` in `g.Close`. Saving inside `g.removeNode` is another option, but I wanted to avoid saving repeatedly when removing many nodes.

TODO: the gateway should ask for more nodes if its node list is getting too sparse.

### DIFF
--- a/modules/gateway/peers_test.go
+++ b/modules/gateway/peers_test.go
@@ -28,18 +28,38 @@ func TestListen(t *testing.T) {
 	g := newTestingGateway("TestListen", t)
 	defer g.Close()
 
-	// "compliant" connect
+	// compliant connect with old version
 	conn, err := net.Dial("tcp", string(g.Address()))
 	if err != nil {
 		t.Fatal("dial failed:", err)
 	}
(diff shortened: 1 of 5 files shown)


modules/gateway/peers_test.go
```diff
@@ -28,20 +28,40 @@ func TestListen(t *testing.T) {
...
-	// "compliant" connect
+	// compliant connect with old version
 	conn, err := net.Dial("tcp", string(g.Address()))
 	if err != nil {
 		t.Fatal("dial failed:", err)
 	}
 	addr := modules.NetAddress(conn.LocalAddr().String())
 	// send version
-	if err := encoding.WriteObject(conn, build.Version); err != nil {
+	if err := encoding.WriteObject(conn, "0.1"); err != nil {
```
DavidVorick: 0.1?

lukechampine: it just needs be a version old enough to be rejected.

DavidVorick:
build failed

DavidVorick:
jk

PR #648, opened 2015-06-28
No strings

DavidVorick:
Compatibility unit testing is not complete, but I wanted to PR this so Adam could have it.

I removed all strings from the types package, which means blocks, etc. should now be compatible with JSON.

This change is neither a hardfork or a softfork

### DIFF
--- a/modules/hostdb/update.go
+++ b/modules/hostdb/update.go
@@ -23,16 +23,19 @@ import (
 // announcement is actually a valid ip address.
 func findHostAnnouncements(b types.Block) (announcements []modules.HostSettings) {
 	for _, t := range b.Transactions {
-		for _, data := range t.ArbitraryData {
-			// the HostAnnouncement must be prefaced by the standard host announcement string
-			if !strings.HasPrefix(data, modules.PrefixHostAnnouncement) {
+		// the HostAnnouncement must be prefaced by the standard host
+		// announcement string
+		var prefix types.Specifier
+		for _, arb := range t.ArbitraryData {
+			copy(prefix[:], arb)
+			if !strings.HasPrefix(string(arb), modules.PrefixStrHostAnnouncement) && // preserved for compatibility with 0.3.3.3
+				prefix != modules.PrefixHostAnnouncement {
 				continue
 			}
 
 			// decode the HostAnnouncement
 			var ha modules.HostAnnouncement
-			encAnnouncement := []byte(strings.TrimPrefix(data, modules.PrefixHostAnnouncement))
-			err := encoding.Unmarshal(encAnnouncement, &ha)
+			err := encoding.Unmarshal(arb[types.SpecifierLen:], &ha)
 			if err != nil {
 				continue
 			}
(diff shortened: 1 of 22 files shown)


modules/hostdb/update.go
```diff
@@ -23,16 +23,19 @@ import (
...
-		for _, data := range t.ArbitraryData {
-			// the HostAnnouncement must be prefaced by the standard host announcement string
-			if !strings.HasPrefix(data, modules.PrefixHostAnnouncement) {
+		// the HostAnnouncement must be prefaced by the standard host
+		// announcement string
+		var prefix types.Specifier
+		for _, arb := range t.ArbitraryData {
+			copy(prefix[:], arb)
+			if !strings.HasPrefix(string(arb), modules.PrefixStrHostAnnouncement) && // preserved for compatibility with 0.3.3.3
+				prefix != modules.PrefixHostAnnouncement {
```
lukechampine: may be unnecessary, since strings and []bytes are encoded equivalently

modules/transactionpool.go
```diff
@@ -6,10 +6,16 @@ import (
...
+const (
+	TransactionSizeLimit = 16 * 1024
+)
+
 var (
 	// ErrTransactionPoolDuplicate is returned when a duplicate transaction is
 	// submitted to the transaction pool.
 	ErrTransactionPoolDuplicate = errors.New("transaction is a duplicate")
+	PrefixNonSia                = types.Specifier{'N', 'o', 'n', 'S', 'i', 'a'}
+	PrefixStrNonSia             = "NonSia" // preserved for compatibility with 0.3.3.3
```
lukechampine: ditto

modules/transactionpool/standard.go
```diff
@@ -87,9 +82,14 @@ func (tp *TransactionPool) IsStandardTransaction(t types.Transaction) (err error
...
-		if !strings.HasPrefix(data, modules.PrefixHostAnnouncement) &&
-			!strings.HasPrefix(data, PrefixNonSia) {
+	var prefix types.Specifier
+	for _, arb := range t.ArbitraryData {
+		copy(prefix[:], arb)
+		strData := string(arb) // preserved for compatibility with 0.3.3.3
+		if !strings.HasPrefix(strData, modules.PrefixStrHostAnnouncement) &&
+			!strings.HasPrefix(strData, modules.PrefixStrNonSia) &&
+			prefix != modules.PrefixHostAnnouncement &&
+			prefix != modules.PrefixNonSia {
```
lukechampine: ditto

DavidVorick:
Because `Specifier` is a byte-array of len 16, compatibility stuff is still needed. `"NonSia"` is different from `[16]byte{'N', 'o', 'n', 'S', 'i', 'a'}`

lukechampine:
We probably should have been using a specifier instead of a string to begin with. Oh well. Maybe someday we can drop compatibility.

DavidVorick:
0.5.0 is probably safe to drop compatibility

DavidVorick:
depends on how many hosts still rely on the original announcement, but it's probably going to be not-many.

DavidVorick:
gah this was not ready

DavidVorick:
sorry about that

PR #649, opened 2015-06-28
Remove alias from types.Signature to []byte

triazo:


compatibility/siag_1.0_test.go
```diff
@@ -63,7 +63,7 @@ func verifyKeysSiag_1_0(uc types.UnlockConditions, folder string, keyname string
 			if err != nil {
 				return err
 			}
-			txn.TransactionSignatures[j].Signature = types.Signature(sig[:])
+			txn.TransactionSignatures[j].Signature = sig[:]
```
DavidVorick: should be sufficient to use `sig`, instead of `sig[:]`

modules/wallet/siag.go
```diff
@@ -161,7 +161,7 @@ func (w *Wallet) SendSiagSiafunds(amount types.Currency, dest types.UnlockHash,
 			if err != nil {
 				return types.Transaction{}, err
 			}
-			txn.TransactionSignatures[sigIndex].Signature = types.Signature(encodedSig[:])
+			txn.TransactionSignatures[sigIndex].Signature = encodedSig[:]
```
DavidVorick: same

types/signatures_test.go
```diff
@@ -249,13 +249,13 @@ func TestTransactionValidSignatures(t *testing.T) {
...
 	sig0[0]++
-	txn.TransactionSignatures[0].Signature = Signature(sig0[:])
+	txn.TransactionSignatures[0].Signature = sig0[:]
 	err = txn.validSignatures(10)
 	if err == nil {
 		t.Error("Corrupted a signature but the txn was still accepted as valid!")
 	}
 	sig0[0]--
-	txn.TransactionSignatures[0].Signature = Signature(sig0[:])
+	txn.TransactionSignatures[0].Signature = sig0[:]
```
DavidVorick: and same with all the other diffs :P

lukechampine:
I ran some checks in the Go playground and it does indeed seem to be an alias problem, and specifically a []byte alias problem (string aliases are fine). Bummer.

triazo:
Removed the slice copying.

triazo:
Turns out that they were necessary for compilation, so added back in.

DavidVorick:
oh

PR #652, opened 2015-06-29
Add compatibility tags to conventions

DavidVorick:
and host data.

Also, remove the type alias for Signature

### DIFF
--- a/modules/transactionpool/standard.go
+++ b/modules/transactionpool/standard.go
@@ -85,9 +85,9 @@ func (tp *TransactionPool) IsStandardTransaction(t types.Transaction) (err error
 	var prefix types.Specifier
 	for _, arb := range t.ArbitraryData {
 		copy(prefix[:], arb)
-		strData := string(arb) // preserved for compatibility with 0.3.3.3
-		if !strings.HasPrefix(strData, modules.PrefixStrHostAnnouncement) &&
-			!strings.HasPrefix(strData, modules.PrefixStrNonSia) &&
+		strData := string(arb)                                               // COMPATv0.3.3.3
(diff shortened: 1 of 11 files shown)


modules/transactionpool/standard.go
```diff
@@ -85,9 +85,9 @@ func (tp *TransactionPool) IsStandardTransaction(t types.Transaction) (err error
 	var prefix types.Specifier
 	for _, arb := range t.ArbitraryData {
 		copy(prefix[:], arb)
-		strData := string(arb) // preserved for compatibility with 0.3.3.3
-		if !strings.HasPrefix(strData, modules.PrefixStrHostAnnouncement) &&
-			!strings.HasPrefix(strData, modules.PrefixStrNonSia) &&
+		strData := string(arb)                                               // COMPATv0.3.3.3
+		if !strings.HasPrefix(strData, modules.PrefixStrHostAnnouncement) && // COMPATv0.3.3.3
+			!strings.HasPrefix(strData, modules.PrefixStrNonSia) && // COMPATv0.3.3.3
```
lukechampine: I would prefer to see this split into a separate block. It makes it more clear what the compatibility marker applies to:
```go
if prefix == modules.PrefixHostAnnouncement || prefix == modules.PrefixNonSia {
    continue
}
// COMPATv0.3.3.3
strData := string(arb)
if strings.HasPrefix(strData, modules.PrefixStrHostAnnouncement) ||
    strings.HasPrefix(strData, modules.PrefixStrNonSia) {
    continue
}
return errors.New("arbitrary data contains unrecognized prefix")
```

DavidVorick:
ok. I did abuse the fact that "HostAnnouncement" is 16 characters to remove that compatibility code. I figured, if it's not necessary then we shouldn't bend backwards to support it.

lukechampine:
:+1:

Issue #653, opened 2015-06-29
Remove strike system


lukechampine:
They aren't being used; there isn't really a need for high-granularity punishments. If they do become necessary in the future, it's easy to add them back (they do not break compatibility).

A timestamp approach would probably work just fine. Every time you talk to a peer, you reset their "last heard from" timestamp. If the timestamp is too far in the past, you disconnect. Timestamps could also be used to select which nodes to connect to on startup.

DavidVorick:
progress?

lukechampine:
Strikes are not used anymore, but I left this open due to the discussion of timestamps. Perhaps a new issue is more appropriate.

lukechampine:
Closing this for now. We will reconsider strikes/bans at a later time. The gateway's threat model needs thorough examination, as peer management tends to be one of the trickiest parts of p2p systems.

PR #655, opened 2015-06-30
Remove TODOs from gateway

### LINKED ISSUE #610: Remove all TODOs from the gateway
gateway should be close enough to stability that all TODOs should be resolved.

### LINKED ISSUE #636: Gateway asks for more nodes if node list drops below 100 nodes

lukechampine:
Fixes #610, fixes #636, and partially addresses #653

### DIFF
--- a/modules/gateway/peers.go
+++ b/modules/gateway/peers.go
@@ -20,6 +20,8 @@ const (
 	wellConnectedThreshold = 8
 	// the gateway will not accept inbound connections above this threshold
 	fullyConnectedThreshold = 128
+	// the gateway will ask for more addresses below this threshold
+	minNodeListLen = 100
 )
 
 type peer struct {
@@ -51,8 +53,23 @@ func (g *Gateway) addPeer(p *peer) {
 	go g.listenPeer(p)
 }
 
+// randomPeer returns a random peer from the gateway's peer list.
+func (g *Gateway) randomPeer() (modules.NetAddress, error) {
+	if len(g.peers) > 0 {
+		r := rand.Intn(len(g.peers))
+		for addr := range g.peers {
+			if r == 0 {
+				return addr, nil
+			}
+			r--
+		}
+	}
+
+	return "", errNoPeers
+}
+
 // randomInboundPeer returns a random peer that initiated its connection.
-func (g *Gateway) randomInboundPeer() modules.NetAddress {
+func (g *Gateway) randomInboundPeer() (modules.NetAddress, error) {
 	if len(g.peers) > 0 {
 		r := rand.Intn(len(g.peers))
 		for addr, peer := range g.peers {
@@ -61,13 +78,13 @@ func (g *Gateway) randomInboundPeer() modules.NetAddress {
 				continue
 			}
 			if r == 0 {
-				return addr
+				return addr, nil
 			}
 			r--
 		}
 	}
 
-	return ""
+	return "", errNoPeers
 }
 
 // listen handles incoming connection requests. If the connection is accepted,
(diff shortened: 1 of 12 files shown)


modules/gateway/peers.go
```diff
@@ -52,8 +53,23 @@ func (g *Gateway) addPeer(p *peer) {
 	go g.listenPeer(p)
 }
 
+// randomPeer returns a random peer from the gateway's peer list.
+func (g *Gateway) randomPeer() (modules.NetAddress, error) {
+	if len(g.peers) > 0 {
+		r := rand.Intn(len(g.peers))
```
DavidVorick: I still have a preference for always using crypto/rand

modules/gateway/peers.go
```diff
@@ -52,8 +53,23 @@ func (g *Gateway) addPeer(p *peer) {
...
+	}
+
+	return "", errNoPeers
+}
+
 // randomInboundPeer returns a random peer that initiated its connection.
-func (g *Gateway) randomInboundPeer() modules.NetAddress {
+func (g *Gateway) randomInboundPeer() (modules.NetAddress, error) {
 	if len(g.peers) > 0 {
 		r := rand.Intn(len(g.peers))
```
DavidVorick: ditto

modules/gateway/peers.go
DavidVorick: the problem is that peer selection does play into things like isolation attacks. Though it seems harmless here, there's always a chance that someone figures out how to abuse non-cryptographic random numbers to manipulate the network for their purposes (i.e. target specific peers for DOS, or something to that effect. If they can guess what your selection will be, it's possible that they can use it against you.)

DavidVorick:
I think it also makes sense to slip the peer culling loop into this PR

modules/gateway/peers.go
lukechampine: well, map iteration is pseudo-random as well, so that helps. More importantly, the attacker has no way of knowing the current ordering of your peer list in the first place.

lukechampine:
I held back on that because I'm not sure how it should be implemented. I think it would be a bad idea for everyone on the network to be constantly opening and immediately closing connections to each other. The current scheme is nice because the connections form a functional purpose, which is to connect to a new peer.
Most peers will be filtering out dead addresses, so as a whole the network should not contain too many dead ones. Dead addresses are only a problem when they vastly outnumber live addresses, and this is less of a problem now that the gateway requests more addresses under 100. Adding timestamps would further improve this, as `shareNodes` could prioritize sending nodes with a recent timestamp.

DavidVorick:
It's really a problem for the bootstrap node though, which will have approx. every peer that's ever connected, and is the go-to place for getting your first round of peers. That list never gets cleaned.

Is opening a connection that expensive? Especially if each peer is only doing it once every 30 minutes?

modules/gateway/peers.go
DavidVorick: You can ask for 30 random peers, do that enough times and you'll start to get an idea through the randomness. People have done really impressive things by manipulating less than perfectly-random numbers, and I don't want them doing equally impressive things to us.

Additionally, maps are not psuedo-random. They are very low entropy, and usually quite predictable.

Is it that bad to use crypto/rand instead of math/rand?

lukechampine:
The bootstrap semantics are going to change anyway, which will make it easier for the bootstrap to send fewer unreachable addresses.

modules/gateway/peers.go
lukechampine: it's annoying to use big.Ints.

DavidVorick: granted, but it's important that a cryptosystem covers all its bases.

lukechampine: our crypto package should probably have its own `Rand` function(s). It's confusing for `crypto` to refer to different packages throughout the codebase.

DavidVorick: I thought it did at one point.

Regardless, I'm in favor of that. We use random numbers a decent amount.

lukechampine:
updated

PR #657, opened 2015-06-30
Explorer Database

triazo:
Open to suggestions on how to clean up addblock.go

### DIFF
--- a/api/blockexplorer.go
+++ b/api/blockexplorer.go
@@ -40,3 +40,24 @@ func (srv *Server) blockexplorerBlockDataHandler(w http.ResponseWriter, req *htt
 
 	writeJSON(w, blockSummaries)
 }
+
+// Handles the api call to retrieve data about a specific hash
+func (srv *Server) blockexplorerGetHashHandler(w http.ResponseWriter, req *http.Request) {
+	// Extract the hash from the request
+	var hash []byte
+	_, err := fmt.Sscanf(req.FormValue("hash"), "%x", &hash)
+	fmt.Printf("%x\n", hash)
+	if err != nil {
+		writeError(w, "Malformed or no hash provided: " + err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	// returnData will be a generic interface. The json encoder
+	// should still work though
+	returnData, err := srv.blocke.GetHashInfo(hash)
+	if err != nil {
+		writeError(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	writeJSON(w, returnData)
+}
--- a/api/blockexplorer.go
+++ b/api/blockexplorer.go
@@ -56,7 +56,7 @@ func (srv *Server) blockexplorerGetHashHandler(w http.ResponseWriter, req *http.
 
 	// returnData will be a generic interface. The json encoder
 	// should still work though
-	returnData, err := srv.blocke.GetHashInfo(hash)
+	returnData, err := srv.blocke.GetHashInfo(data)
 	if err != nil {
 		writeError(w, err.Error(), http.StatusInternalServerError)
 		return
--- a/api/blockexplorer.go
+++ b/api/blockexplorer.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"net/http"
 
-	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/types"
 )
 
@@ -45,14 +44,12 @@ func (srv *Server) blockexplorerBlockDataHandler(w http.ResponseWriter, req *htt
 // Handles the api call to retrieve data about a specific hash
 func (srv *Server) blockexplorerGetHashHandler(w http.ResponseWriter, req *http.Request) {
 	// Extract the hash from the request
-	var hash crypto.Hash
 	var data []byte
 	_, err := fmt.Sscanf(req.FormValue("hash"), "%x", &data)
 	if err != nil {
 		writeError(w, "Malformed or no hash provided: "+err.Error(), http.StatusBadRequest)
 		return
 	}
-	copy(hash[:], data[:crypto.HashSize])
 
 	// returnData will be a generic interface. The json encoder
 	// should still work though
--- a/api/blockexplorer.go
+++ b/api/blockexplorer.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net/http"
 
+	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/types"
 )
 
@@ -44,13 +45,14 @@ func (srv *Server) blockexplorerBlockDataHandler(w http.ResponseWriter, req *htt
 // Handles the api call to retrieve data about a specific hash
 func (srv *Server) blockexplorerGetHashHandler(w http.ResponseWriter, req *http.Request) {
 	// Extract the hash from the request
-	var hash []byte
-	_, err := fmt.Sscanf(req.FormValue("hash"), "%x", &hash)
-	fmt.Printf("%x\n", hash)
+	var hash crypto.Hash
+	var data []byte
+	_, err := fmt.Sscanf(req.FormValue("hash"), "%x", &data)
 	if err != nil {
-		writeError(w, "Malformed or no hash provided: " + err.Error(), http.StatusBadRequest)
+		writeError(w, "Malformed or no hash provided: "+err.Error(), http.StatusBadRequest)
 		return
 	}
+	copy(hash[:], data[:crypto.HashSize])
 
 	// returnData will be a generic interface. The json encoder
 	// should still work though
--- a/modules/blockexplorer/addblock.go
+++ b/modules/blockexplorer/addblock.go
@@ -1,0 +1,297 @@
+package blockexplorer
+
+import (
+	"github.com/NebulousLabs/Sia/build"
+	"github.com/NebulousLabs/Sia/crypto"
+	"github.com/NebulousLabs/Sia/encoding"
+	"github.com/NebulousLabs/Sia/persist"
+	"github.com/NebulousLabs/Sia/types"
+)
+
+// func appendModification(modifications []persist.BoltModification, bucketName string, modID crypto.Hash, txid crypto.Hash, t reflect.Type, modFunc func(input t) t ) {
+// 	modIDBytes = encoding.Marshal(modID)
+// 	mapFunc := func(modStructBytes []byte) (presist.BoltItem, error) {
+// 		if modStructBytes == nil {
+// 			return errors.New(fmt.Sprintf("requested item %x does not exist in bucket %s", modIDBytes, bucketName))
+// 		}
+// 		var modStruct t
+// 		encoding.Unmarshal(modStructBytes, modStructType)
+
+// 	}
+// }
+
+func appendSiacoinInput(modifications []persist.BoltModification, outputID types.SiacoinOutputID, txid crypto.Hash) []persist.BoltModification {
+	oID := encoding.Marshal(outputID)
+	mapFunc := func(outputBytes []byte) (persist.BoltItem, error) {
+		var output outputTransactions
+		err := encoding.Unmarshal(outputBytes, &output)
+		if err != nil {
+			return *new(persist.BoltItem), err
+		}
+
+		output.InputTx = txid
+		return persist.BoltItem{
+			BucketName: "SiacoinOutputs",
+			Key:        oID,
+			Value:      encoding.Marshal(output),
+		}, nil
+	}
+	return append(modifications, persist.BoltModification{
+		BucketName: "SiacoinOutputs",
+		Key:        oID,
+		Map:        mapFunc,
+	})
+}
+
+func appendSiafundInput(modifications []persist.BoltModification, outputID types.SiafundOutputID, txid crypto.Hash) []persist.BoltModification {
+	oID := encoding.Marshal(outputID)
+	mapFunc := func(outputBytes []byte) (persist.BoltItem, error) {
+		var output outputTransactions
+		err := encoding.Unmarshal(outputBytes, &output)
+		if err != nil {
+			return *new(persist.BoltItem), err
+		}
+
+		output.InputTx = txid
+		return persist.BoltItem{
+			BucketName: "SiafundOutputs",
+			Key:        oID,
+			Value:      encoding.Marshal(output),
+		}, nil
+	}
+	return append(modifications, persist.BoltModification{
+		BucketName: "SiafundOutputs",
+		Key:        oID,
+		Map:        mapFunc,
+	})
+}
+
+func appendFcRevision(modifications []persist.BoltModification, fcid types.FileContractID, txid crypto.Hash) []persist.BoltModification {
+	fcidBytes := encoding.Marshal(fcid)
+	mapFunc := func(fcBytes []byte) (persist.BoltItem, error) {
+		var fcInfoStruct fcInfo
+		err := encoding.Unmarshal(fcBytes, &fcInfoStruct)
+		if err != nil {
+			return *new(persist.BoltItem), err
+		}
+
+		fcInfoStruct.Revisions = append(fcInfoStruct.Revisions, txid)
+		return persist.BoltItem{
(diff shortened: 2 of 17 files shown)


api/blockexplorer.go
```diff
@@ -40,3 +41,25 @@ func (srv *Server) blockexplorerBlockDataHandler(w http.ResponseWriter, req *htt
...
+
+	// returnData will be a generic interface. The json encoder
+	// should still work though
+	returnData, err := srv.blocke.GetHashInfo(data)
+	if err != nil {
+		writeError(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	writeJSON(w, returnData)
+}
```
DavidVorick: This function should probably also support addresses

triazo: It does. The different lengths between a raw address with checksum and a crypto.Hash is the reason data is a []byte.

api/blockexplorer.go
```diff
@@ -40,3 +41,25 @@ func (srv *Server) blockexplorerBlockDataHandler(w http.ResponseWriter, req *htt
...
+func (srv *Server) blockexplorerGetHashHandler(w http.ResponseWriter, req *http.Request) {
+	// Extract the hash from the request
+	var hash crypto.Hash
+	var data []byte
+	_, err := fmt.Sscanf(req.FormValue("hash"), "%x", &data)
+	if err != nil {
+		writeError(w, "Malformed or no hash provided: "+err.Error(), http.StatusBadRequest)
+		return
+	}
+	copy(hash[:], data[:crypto.HashSize])
```
lukechampine: this will panic without a proper bounds check.

modules/blockexplorer/addblock.go
```diff
@@ -0,0 +1,390 @@
...
+		Value:      encoding.Marshal(txInfo{b.ID(), -1}),
+	})
+	// TODO add a dummy txid with block id and txNum of -1, so
+	// that txid lookups on a block id are still valid
+
+	// Insert the miner payouts as new outputs
+	for i := range b.MinerPayouts {
+		modifications = appendAddress(modifications, b.MinerPayouts[i].UnlockHash, crypto.Hash(b.ID()))
+		additions = appendHashType(additions, crypto.Hash(b.MinerPayouts[i].UnlockHash), hashUnlockHash)
+		additions = appendNewOutput(additions, b.MinerPayoutID(i), crypto.Hash(b.ID()))
```
lukechampine: instead of performing the append inside the functions, do it here:
```go
modifications = append(modifications, newAddressDiff(b.MinerPayouts[i].UnlockHash, crypto.Hash(b.ID()))
```
this makes functions easier to test, and is generally more "functional" than passing around a pointer to be modified.

modules/blockexplorer/addblock.go
```diff
@@ -0,0 +1,390 @@
...
+}
+
+func (be *BlockExplorer) addTransaction(tx types.Transaction, bID types.BlockID, txNum int) error {
+	// Store this for quick lookup
+	txid := tx.ID()
+
+	// A list of things to be added to the database. They will all
+	// be performed at the end of this function
+	modifications := make([]persist.BoltModification, 0)
+	changes := make([]persist.BoltItem, 0)
```
lukechampine: don't use `make(0)`. Instead:
```go
var modifications []persist.BoltModification
var changes []persist.BoltItem
```

modules/blockexplorer/addblock.go
```diff
@@ -0,0 +1,390 @@
...
+	}
+
+	err := be.db.BulkUpdate(modifications, additions)
+	if err != nil {
+		return err
+	}
+
+	// Insert each transaction
+	for i, tx := range b.Transactions {
+		err = be.addTransaction(tx, b.ID(), i)
```
lukechampine: why not have `addTransaction` return modifications and additions? Then you only need one `BulkUpdate`.

modules/blockexplorer/addblock.go
```diff
@@ -0,0 +1,390 @@
...
+		changes = appendNewOutput(changes, tx.SiacoinOutputID(i), txid)
+	}
+
+	// Handle each file contract individually
+	for i, contract := range tx.FileContracts {
+		fcid := tx.FileContractID(i)
+		changes = append(changes, persist.BoltItem{
+			BucketName: "FileContracts",
+			Key:        encoding.Marshal(fcid),
+			Value: encoding.Marshal(fcInfo{
```
lukechampine: didn't go `go fmt`'d?

modules/blockexplorer/database.go
```diff
@@ -0,0 +1,157 @@
...
+
+Heights
+  types.BlockHeight ------> types.BlockID
+Hashes
+  crypto.Hash ------------> hash type (int)
+
+*/
+
+var meta persist.Metadata = persist.Metadata{
+	Version: ".1",
```
lukechampine: need a leading 0

modules/blockexplorer/database.go
```diff
@@ -0,0 +1,157 @@
...
+// Returns an array of block summaries. Bounds checking should be done elsewhere
+func (db *explorerDB) dbBlockSummaries(start types.BlockHeight, finish types.BlockHeight) ([]modules.ExplorerBlockData, error) {
+	summaries := make([]modules.ExplorerBlockData, int(finish-start))
+	err := db.View(func(tx *bolt.Tx) error {
+		heights := tx.Bucket([]byte("Heights"))
+
+		// Iterate over each block height, constructing a
+		// summary data structure for each block
+		for i := start; i < finish; i++ {
+			bSummaryBytes := heights.Get(encoding.Marshal(types.BlockHeight(i)))
```
lukechampine: `i` should already be a `types.BlockHeight`

modules/blockexplorer/database.go
```diff
@@ -0,0 +1,157 @@
...
+
+// Returns an array of block summaries. Bounds checking should be done elsewhere
+func (db *explorerDB) dbBlockSummaries(start types.BlockHeight, finish types.BlockHeight) ([]modules.ExplorerBlockData, error) {
+	summaries := make([]modules.ExplorerBlockData, int(finish-start))
+	err := db.View(func(tx *bolt.Tx) error {
+		heights := tx.Bucket([]byte("Heights"))
+
+		// Iterate over each block height, constructing a
+		// summary data structure for each block
+		for i := start; i < finish; i++ {
```
lukechampine: if the heights are stored in sorted order, you can iterate more idiomatically using a `Cursor`:
```go
b := tx.Bucket([]byte("Heights"))
c := b.Cursor()
k, v := c.Seek(encoding.Marshal(start))
for i := 0; i < finish-start; i++ {
    // process k,v
    k, v = c.Next() 
}
```
need to test whether they are, in fact, sorted though.

modules/blockexplorer/hashes.go
```diff
@@ -0,0 +1,259 @@
...
+	typeBytes, err := be.db.GetFromBucket("Hashes", hash[:crypto.HashSize])
+	if err != nil {
+		return nil, err
+	}
+	if typeBytes == nil {
+		return nil, errors.New("requested hash not found in database")
+	}
+
+	var hashType int
+	err = encoding.Unmarshal(typeBytes, &hashType)
```
lukechampine: err is not handled

modules/blockexplorer/hashes.go
```diff
@@ -0,0 +1,259 @@
...
+	defer be.mu.RUnlock(lockID)
+
+	// Perform a lookup to tell which type of hash it is
+	typeBytes, err := be.db.GetFromBucket("Hashes", hash[:crypto.HashSize])
+	if err != nil {
+		return nil, err
+	}
+	if typeBytes == nil {
+		return nil, errors.New("requested hash not found in database")
+	}
```
lukechampine: shouldn't this condition be checked for (and returned as an error) in the `GetFromBucket` function? Or are there cases where we want to check for nil explicitly (i.e. to branch on it)?

modules/blockexplorer/hashes.go
```diff
@@ -0,0 +1,259 @@
...
+		var id types.SiacoinOutputID
+		copy(id[:], hash[:crypto.HashSize])
+		return be.db.getSiacoinOutput(id)
+	case hashFundOutputID:
+		var id types.SiafundOutputID
+		copy(id[:], hash[:crypto.HashSize])
+		return be.db.getSiafundOutput(id)
+	case hashUnlockHash:
+		var id types.UnlockHash
+		copy(id[:], hash[:crypto.HashSize])
```
lukechampine: don't index `hash` (here and elsewhere).

modules/blockexplorer/hashes.go
```diff
@@ -0,0 +1,259 @@
...
+	var bd blockData
+	err = encoding.Unmarshal(b, &bd)
+	if err != nil {
+		return block, err
+	}
+	block.Block = bd.Block
+	block.Height = bd.Height
+	block.ResponseType = responseBlock
+	return block, nil
+}
```
lukechampine: if you're using named return values, just use `return`

persist/boltdb.go
```diff
@@ -0,0 +1,189 @@
...
+	var bytes []byte
+	err := db.View(func(tx *bolt.Tx) error {
+		bucket := tx.Bucket([]byte(bucketName))
+		if bucket == nil {
+			return errors.New("requested bucket does not exist: " + bucketName)
+		}
+
+		// Note that bytes could be nil. This is OK, but needs
+		// to be checked in calling functions
+		bytes = bucket.Get(key)
```
lukechampine: >Please note that values returned from Get() are only valid while the transaction is open. If you need to use a value outside of the transaction then you must use copy() to copy it to another byte slice.

modules/blockexplorer/addblock.go
```diff
@@ -0,0 +1,390 @@
...
+				BucketName: "Addresses",
+				Key:        addr,
+				Value:      encoding.Marshal([]crypto.Hash{txid}),
+			}, nil
+		}
+
+		var txids []crypto.Hash
+		err := encoding.Unmarshal(addrBytes, &txids)
+		if err != nil {
+			return *new(persist.BoltItem), err
```
lukechampine: use `persist.BoltItem{}`

lukechampine:
I am wary of the `BoltItem` and `BoltModification` approach, though I do like how you used first-class functions to implement `BoltModification`. You actually only need one type, which is:
```go
type boltModification func(tx *bolt.Tx) error
```
`BulkUpdate` might also be unnecessary, since you can managed transactions manually using `db.Begin(true)` (see the Bolt docs). To summarize, `addBlock` might look something like:
```go
func insertObject(tx *bolt.Tx, bucket string, key, val interface{}) error {
    b := tx.Bucket([]byte(bucket))
    if b == nil {
        return errors.New("nil bucket")
    }
    return b.Put(key, encoding.Marshal(val))
}

func (be *BlockExplorer) addBlockDB(b types.Block) error {
    // ... more initialization here ...

    tx, err := be.db.Begin(true)
    if err != nil {
        return err
    }
    defer tx.Rollback()

    err := insertObject(tx, "Blocks", b.ID(), blockstruct)
    err := insertObject(tx, "Heights", be.blockchainHeight, bsum)
    // etc.

    for i := range b.MinerPayouts {
        addressFn(b.MinerPayouts[i].UnlockHash, crypto.Hash(b.ID()))(tx)
        // etc.
    }

    for i, txn := range b.Transactions {
        err = transactionFn(txn, b.ID(), i)(tx)
    }

    return tx.Commit()
}
where ``addressFn`, `transactionFn`, etc. are functions with return type `func(*bolt.Tx) error`.

File: modules/blockexplorer/database.go (hunk 7 above)
Comment (DavidVorick) [inline modules/blockexplorer/database.go]: or

`meta := persist.Metadata{`

Comment (lukechampine) [inline modules/blockexplorer/database.go]: that syntax isn't allowed at top level

File: modules/blockexplorer/addblock.go (hunk 6 above)
Comment (triazo) [inline modules/blockexplorer/addblock.go]: `go fmt` doesn't align nested structs

Comment (DavidVorick) [inline modules/blockexplorer/addblock.go]: I believe it does

Comment (triazo) [inline modules/blockexplorer/addblock.go]: Just made a test file:

    $ go fmt /tmp/test.go
    $ cat /tmp/test.go   
    package main
    
    var x = s{
    	long: 1,
    	m: s{
    		m: 1,
    	},
    }

Comment (DavidVorick) [inline modules/blockexplorer/addblock.go]: oh, you are right

**Comment by triazo** [comment]
with `err = payoutFn(payout, b.ID(), i)(tx)` what is the purpose of returning a function when you are calling it at that moment. Would it not just be better to pass tx in as an argument in this case?

**Comment by DavidVorick** [comment]
your build is still failing

**Comment by triazo** [comment]
Here's what appendAddress looks like when given access to the bolt transaction, similar to how you suggested (functionalizing it would just require a `return func (tx *bolt.Tx) error {` before line 2, so the two cases are very similar)

```

func addAddress(tx *bolt.Tx, addr types.UnlockHash, txid crypto.Hash) error {
	b = tx.Bucket([]byte("Addresses"))
	if b == nil {
		return errors.New("Addresses bucket does not exist")
	}

	txBytes := b.Get(encoding.Marshal(addr))
	if addrs == nil {
		err := b.Put(encoding.Marshal(addr), encoding.Marshal([]crypto.Hash{txid}))
		if err != nil {
			return err
		}
		return nil
	}

	var txns []crypto.Hash
	err = encoding.Unmarshal(txBytes, &txns)
	if err != nil {
		return err
	}

	txns = append(txns, txid)

	return b.Put(encoding.Marshall(addr), encoding.Marshal(txns))
}
```

It does seem much simpler this way.

**Comment by lukechampine** [comment]
that looks good to me. The only advantage of returning a `func(*bolt.Tx) error` is that you can use those to batch your modifications. But I think that's largely unnecessary anyway, so perhaps passing in the tx is the way to go.

**Comment by triazo** [comment]
Changed DB update model. Even though its longer now in line count, its more direct in how it does things, and clearer.

File: modules/blockexplorer/database.go (hunk 9 above)
Comment (triazo) [inline modules/blockexplorer/database.go]: They are sorted by the wrong endianness. Better to keep sorted as they are then change the key type.

**Comment by lukechampine** [comment]
awesome, much better.
One pattern I see a lot is fetching bytes from a bucket and then unmarshalling the bytes into a type. Perhaps you could define a helper function:
```go
func getObject(b *bolt.Bucket, key []byte, obj interface{}) error {
    objBytes := b.Get(key)
    if objBytes == nil {
        return errNilEntry
    }
    return encoding.Unmarshal(objBytes, obj)
}
```

modules/blockexplorer/hashes.go
```diff
@@ -0,0 +1,224 @@
...
+	err = encoding.Unmarshal(typeBytes, &hashType)
+	if err != nil {
+		return nil, err
+	}
+
+	switch hashType {
+	case hashBlock:
+		var id types.BlockID
+		copy(id[:], hash[:])
+		return be.db.getBlock(types.BlockID(id))
```
lukechampine: if `len(id) == len(hash`), isn't a cast sufficient? i.e. `return be.db.getBlock(types.BlockID(hash)`?

triazo: `hash` might contain an address, which is longer than a `crypto.Hash` due to the checksum

lukechampine:
I'm not totally onboard with the generic `hash` approach. Would it be sufficient to just return the hash type? That would require the caller to make a separate request to retrieve the actual data, but that doesn't seem so bad to me.

triazo:
Currently, when you request a Siacoin output, it returns two ID's. As the output may be either a transaction or a block (due to miner payouts), the api user would not know which type of lookup to do, unless additional information was sent. This is also the case with address lookups. Having a single request simplified this.

To change to having separate requests, we can either do a lookup every time we request a transaction, and branch if its a block, or provide some additional information in along with the output data for the case where the output is a miner payout.

The address, ouput, and file contract lookups should always return roughly the same thing so changing them woudl be easier.

triazo:
I managed to combine a few of the insertion functions, and reduced two more to only two lines. They could have been put into the main code, but for clarity's sake I left them seperate. I still can't figure out how to combine the modification functions, as the modification done is slightly different every time.

PR #659, opened 2015-07-01
Consensus testing

DavidVorick:
Decided that this is a good stopping place.

I added a good amount of testing to the consensus package, but I'm probably only about 40% of the way to completion. I made more changes to the actual consensus code than I was expecting, but they should all be non-forking.

maintenance.go needs a little more work, verifytransaction.go needs a lot more work, and then I need to write a bunch of tests that use full blocks.

After that the persist stuff needs to be changed and updates need to be written for that.

Then the notifications.

Finally, there are some logic changes to the synchronize package that will eventually be necessary for DoS purposes (you don't accept blocks that don't extend the longest chain unless you first get a set of headers that demonstrate a longer chain).


DavidVorick:
I think the build is failing because, with all the debug checks that the consensus now does, it's actually really computationally expensive to mine 500 blocks that build on eachother. I did upgrade the sanity checks and consistency checks in this PR, and now the entire consensus set gets scanned every single block.

My next commit pulls the miner back from 500 blocks to 100 blocks, that will probably stop the deadlock detector from triggering as well as keep the test below 3 minutes.

I do think we should keep the consistency checks. Although expensive, they check important things like the total number of siacoins and siafunds. The one consensus bug that slipped through (so far) would have been caught by the consistency check that counts siacoins.

modules/consensus/consistency.go
```diff
@@ -2,12 +2,46 @@ package consensus
...
 
+var (
+	errSiacoinMiscount = errors.New("consensus set has the wrong number of siacoins given the height")
+)
+
+// checkCurrentPath looks at the blocks in the current path and verifies that
+// they are all ordered correctly and in the block map.
+func (cs *State) checkCurrentPath() error {
+	currentNode := cs.currentBlockNode()
+	for i := cs.height(); i != 0; i-- {
```
lukechampine: this won't check the genesis block, correct?

modules/consensus/diffs_test.go
```diff
@@ -0,0 +1,736 @@
...
+	defer func() {
+		r := recover()
+		if r != errWrongRevertDiffSet {
+			t.Error("expected errWrongRevertDiffSet, got", r)
+		}
+
+		// trigger a panic about applying the wrong block.
+		bn.block.ParentID[0]++
+		cst.cs.commitDiffSetSanity(bn, modules.DiffApply)
+	}()
```
lukechampine: I don't think you can defer multiple `recover`s. Only the first panic will be caught.

modules/consensus/fork.go
```diff
@@ -1,82 +1,84 @@
...
 // revertToNode will revert blocks from the State's current path until 'bn' is
 // the current block. Blocks are returned in the order that they were reverted.
-func (s *State) revertToNode(bn *blockNode) (revertedNodes []*blockNode) {
+// 'bn' is not reverted.
+func (cs *State) revertToNode(bn *blockNode) (revertedNodes []*blockNode) {
 	// Sanity check - make sure that bn is in the currentPath.
 	if build.DEBUG {
-		if s.currentPath[bn.height] != bn.block.ID() {
-			panic("can't revert to node not in current path")
+		if types.BlockHeight(len(cs.currentPath)) <= bn.height || cs.currentPath[bn.height] != bn.block.ID() {
```
lukechampine: use `cs.height` instead?

types/currency.go
```diff
@@ -82,26 +82,22 @@ func (x Currency) Mul(y Currency) (c Currency) {
...
 			panic(ErrNegativeCurrency)
 		}
 	} else {
-		cRat := new(big.Rat).Mul(
-			new(big.Rat).SetInt(&x.i),
-			new(big.Rat).SetFloat64(y),
-		)
-		c.i.Div(cRat.Num(), cRat.Denom())
+		tmp := new(big.Rat).Mul(new(big.Rat).SetInt(&x.i), y)
+		c.i.Div(tmp.Num(), tmp.Denom())
```
lukechampine: alternatively:
```go
c.i.Mul(&x.i, y.Num())
c.i.Div(&c.i, y.Denom())
```

DavidVorick:
apparently the siad tests are now failing, but I can't reproduce it locally.

modules/consensus/consistency.go
DavidVorick: Yes, I believe it skips the genesis block

DavidVorick:
Because I broke compatibility somewhere, don't merge this yet. I am working on fixing it.

DavidVorick:
There's a race condition on the consensus variable I set up, still need to fix that.

Except for that fix, everything is good to go.

DavidVorick:
At long last, this guy is finished :)

types/constants.go
```diff
@@ -8,31 +8,43 @@ package types
...
+// COMPATv0.4.0 - until the first 10e3 blocks have been archived, CurrentHeight
+// is needed to keep code consistent while verifying earlier blocks.
+//
+// CurrentHeight is a global variable to help with an upcoming hardfork. When
+// the height is less than 12,000, one set of code will be run. When the height
+// is greater than 12,000, a different set of code will be run. After the
+// hardfork has passed and everyone has upgraded, a checkpoint will be created
+// and blocks below 12,000 will never be viewed again.
+var CurrentHeight BlockHeight = 0
+var CurrentHeightLock sync.Mutex
```
lukechampine: is CurrentHeight ever modified by more than one goroutine? Seems like it only gets modified when a new block is added/removed, which should be strictly serial.

is there a reason why the compatibility code has to live in the types package? Wouldn't it make more sense in consensus?

DavidVorick: almost.

The amount of code that I would have had to introduce I believe is a lot higher. Maybe I didn't look hard enough for a better solution, but the only function that gets affected is the FileContract.Tax() function. It made sense to just put the limits there, instead of having a larger architecture try to choose between two different Tax funtions.

DavidVorick: I can see the argument for doing it the other way.

PR #661, opened 2015-07-02
add host flag to siac

lukechampine:
I considered making this work the same as siad (i.e. ":9980" instead of "9980") but then you have to check for a nil hostname and change it to "localhost"...
I'm open to sticking with the siad convention if you think it's preferable though.

### DIFF
--- a/siac/main.go
+++ b/siac/main.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"io/ioutil"
+	"net"
 	"net/http"
 	"net/url"
 	"os"
@@ -16,7 +17,7 @@ import (
 )
 
 var (
-	port  string
+	addr  string
 	force bool
 )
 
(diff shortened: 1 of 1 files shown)


DavidVorick:
If I understand correctly

`siac -p siacoin.com:9980` is illegal but
`siac -p :9980 -h siacoin.com` is legal?

Definitely we need to do things the same way on both binaries.

I'm not sure which one I prefer. It seems weird to me to have 2 flags though, but `--port` is a bad name if it's also accepting hostnames.

lukechampine:
yeah, the issue was the existing `port` flag wasn't amenable to having a hostname as well.

I think it's reasonable to replace it with a `-a`/`--api-addr` flag, as in siad.

DavidVorick:
ok, let's replace -p and -H with -a

lukechampine:
fixed

Issue #664, opened 2015-07-06
Block Explorer Tests


DavidVorick:
I'm not sure why, but the block explorer tests timed out.

The time limit is currently set to 3 minutes. For the time being, I would like to keep it there. Are there some tests that you can reduce or shorten without reducing code coverage?

PR #665, opened 2015-07-07
Explorer optimize startup and testing deadlock

triazo:
Addresses #664 and partially https://github.com/NebulousLabs/Sia-Explorer/issues/11

### DIFF
--- a/modules/blockexplorer/addblock.go
+++ b/modules/blockexplorer/addblock.go
@@ -13,10 +13,6 @@ import (
 	"github.com/boltdb/bolt"
 )
 
-var (
-	ErrNilEntry = errors.New("entry does not exist")
-)
-
 // A boltTx is a bolt transaction. It implements monadic error handling, such that
 // any operation that occurs after an error becomes a no-op.
 type boltTx struct {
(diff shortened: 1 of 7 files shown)


modules/blockexplorer/addblock.go
```diff
@@ -172,7 +173,17 @@ func (be *BlockExplorer) addBlockDB(b types.Block) error {
 	}
 	defer tx.Rollback()
 
-	// Construct the struct that will be inside the database
+	// Check if the block is already there before adding it
+	blkBytes, err := be.db.GetFromBucket("Blocks", encoding.Marshal(b.ID()))
+	if err != nil && err != persist.ErrNilEntry {
+		return err
+	}
```
lukechampine: whoops, I didn't realize you already had a `GetFromBucket` function. It kinda overlaps with `getObject`.
fine for now, but maybe later we could add an `exists` helper function that checks for an entry.

persist/boltdb.go
```diff
@@ -26,6 +26,10 @@ type BoltItem struct {
 	Value      []byte
 }
 
+var (
+	ErrNilEntry = errors.New("entry does not exist")
+)
```
lukechampine: I believe an identical error is defined in addblock.go as well, so you can replace those errors with this one.

lukechampine:
build failed; possibly a deadlock timeout?

triazo:
Travis build is now passing

PR #667, opened 2015-07-08
Organization

DavidVorick:
Was rather exhausted when I wrote some of this code. I did write tests for everything. Each commit is standalone.

For the last commit, which should maybe have a longer commit message, I fixed a bug where certain types of storage proofs were actually impossible. When creating merkle roots and verifying merkle proofs, the last segment of the file does not get padded out. Prior to this PR, when creating merkle proofs, the last segment of the file DID get padded out, which would result in invalid proofs being created. Invalid proofs would only be created if the segment that needed to be provided was the last segment (a rare occurrence) AND that segment was not the full 64 bytes.

This bug has been fixed. BuildMerkleProof now works correctly, and VerifyMerkleProof verifies the proofs correctly. This ends up being a softfork.

### DIFF
--- a/crypto/merkle.go
+++ b/crypto/merkle.go
@@ -71,13 +71,13 @@ func ReaderMerkleRoot(r io.Reader) (h Hash, err error) {
 }
 
 // BuildReaderProof will build a storage proof when given a reader.
-func BuildReaderProof(r io.Reader, proofIndex uint64) (base [SegmentSize]byte, hashSet []Hash, err error) {
+func BuildReaderProof(r io.Reader, proofIndex uint64) (base []byte, hashSet []Hash, err error) {
 	_, proofSet, _, err := merkletree.BuildReaderProof(r, NewHash(), SegmentSize, proofIndex)
 	if err != nil {
 		return
 	}
 	// convert proofSet to base and hashSet
-	copy(base[:], proofSet[0])
+	base = proofSet[0]
 	hashSet = make([]Hash, len(proofSet)-1)
 	for i, proof := range proofSet[1:] {
 		copy(hashSet[i][:], proof)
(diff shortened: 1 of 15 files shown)


crypto/merkle.go
```diff
@@ -71,26 +71,26 @@ func ReaderMerkleRoot(r io.Reader) (h Hash, err error) {
...
 		return
 	}
 	// convert proofSet to base and hashSet
-	copy(base[:], proofSet[0])
+	base = proofSet[0]
 	hashSet = make([]Hash, len(proofSet)-1)
-	for i, proof := range proofSet[1:] {
-		copy(hashSet[i][:], proof)
+	for i := range proofSet[1:] {
+		copy(hashSet[i][:], proofSet[i+1])
```
lukechampine: why drop the `proof`? just a general dislike for value ranges? In this case they're just pointers so the efficiency is good.

modules/consensus/validtransaction.go
```diff
@@ -84,14 +85,31 @@ func (cs *State) validStorageProofs(t types.Transaction) error {
...
+		if segmentIndex == leaves-1 {
+			segmentLen = fc.FileSize % crypto.SegmentSize
+		}
+
+		// COMPATv0.4.0
+		//
+		// Fixing the padding situation resulted in a hardfork. The below code
+		// will stop the hardfork from triggering before block 12,000.
+		types.CurrentHeightLock.Lock()
+		if (build.Release == "standard" && types.CurrentHeight < 15e3) || (build.Release == "testing" && types.CurrentHeight < 10) {
```
lukechampine: that doesn't look like a 12 to me

DavidVorick: I changed it to 15. I think we'll have 0.4.0 out somewhere between 10k and 12k blocks, and then we need another few weeks for people to upgrade.

crypto/merkle.go
DavidVorick: yeah. Just feels like better practice not to make a copy if you don't really need one. Even if it's just a slice header.

DavidVorick:
good to go?

modules/consensus/validtransaction.go
lukechampine: the comment is wrong. And what about the `Tax()` hardfork; that one still triggers at 12k, right?

DavidVorick: no they've both been adjusted to 15k. I'll fix the comment

DavidVorick:
fixed

crypto/merkle.go
lukechampine: I disagree. I feel like it's better practice to write for readability first and optimize where necessary. Using an array index is especially confusing because you're ranging over `proofSet[1:]` instead of just `proofSet`.

DavidVorick: ok, fixed.

I don't like making extra memory copies when you don't have to.

But these might actually be efficient because they just use the same memory repeatedly. I guess it's something we could benchmark.

PR #669, opened 2015-07-10
More organization

DavidVorick:
I added the 'debug' flag to the `make release`, I'm surprised it wasn't already enabled. I had to disable a few of the debug stuff, as well as optimize others, simply because it was very slow otherwise. Right now, with both 'debug' and '-race' enabled, verifying the whole blockchain takes about 2 minutes on my machine. The majority of that time is garbage collection and the race library, I forget exactly but I think it's something like 95% between the two of them, even with the signatures, hashing, and encoding being so slow.

I'm guessing the majority of that is because of the sheer volume of threads that we use, and because we create a lot of new memory when hashing things.

I also created a release-std, which doesn't have the race libraries or the debug flag enabled, which is useful for profiling what the user is going to experience.

The ConsensusSet finally got renamed from the State.

Some obsolete documentation got updated, but it'd still be really nice to revisit the whitepaper.

### DIFF
--- a/doc/Developers.md
+++ b/doc/Developers.md
@@ -1,7 +1,7 @@
 Developer Environment
 =====================
 
-Sia is mostly written in go. To build and test Sia, you are going to need a
+Sia is written in golang. To build and test Sia, you are going to need a
 working go environment, including having both $GOROOT/bin and $GOPATH/bin in
 your $PATH. For most Linux distributions, go will be in the package manager.
 Then it should be sufficient to run `make dependencies && make`. For more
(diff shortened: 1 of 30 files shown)


doc/Developers.md
```diff
@@ -151,7 +147,7 @@ Error Handling
 All errors need to be checked as soon as they are received, even if they are
 known to not cause problems. The statement that checks the error needs to be
 `if err != nil`, and if there is a good reason to use an alternative statement
-(such ass `err == nil`), it must be documented. The body of the if statement
+(such as `err == nil`), it must be documented. The body of the if statement
```
lukechampine: kek

PR #670, opened 2015-07-10
General improvements

DavidVorick:
There's still a ways to go as far as performance is concerned. BUT:

1. file contract maintenance is now O(n) in the number of file contracts expiring, not O(n) in the total number of file contracts. Because of the use of maps, performance has actually decreased right now, because there are only a few thousand file contracts. I'm not sure where the break-even point is, but ideally we'll eventually have somewhere between 100k and 10m file contracts on the blockchain, this change was necessary.

2. The subscription model no longer saves an entire extra copy of the diff set, instead calculating it on the fly when needed.

3. Some profiling code got added to the daemon. You can run the profiler with '-p'. It's not very sophisticated at the moment, and I actually think that the memory profiler is broken in some ways. I still don't understand how to use the memory profiler.

4. We now use travis-CI's new build system. http://docs.travis-ci.com/user/migrating-from-legacy/

This concludes this round of me changing the consensus package. There's still a handful of tests that need to be written, but they are less critical. I plan on getting back to those before we release, probably the week before.

My next focus is going to be the transaction pool.

### DIFF
--- a/modules/consensus.go
+++ b/modules/consensus.go
@@ -136,6 +136,11 @@ type ConsensusSet interface {
 	// run any required closing routines.
 	Close() error
 
+	// ConsensusChange returns the consensus change that occured at the input
+	// index. An error is returned if there is not a change at the provided
+	// index.
+	ConsensusChange(int) (ConsensusChange, error)
+
 	// ConsensusSetSubscribe will subscribe another module to the consensus
 	// set. Every time that there is a change to the consensus set, an update
 	// will be sent to the module via the 'ReceiveConsensusSetUpdate' function.
--- a/modules/consensus.go
+++ b/modules/consensus.go
@@ -136,10 +136,11 @@ type ConsensusSet interface {
 	// run any required closing routines.
 	Close() error
 
-	// ConsensusChange returns the consensus change that occured at the input
-	// index. An error is returned if there is not a change at the provided
-	// index.
-	ConsensusChange(int) (ConsensusChange, error)
+	// ConsensusChange returns the ith consensus change that was broadcast to
+	// subscribers by the consensus set. An error is returned if i consensus
+	// changes have not been broadcast. The primary purpose of this funciton is
+	// to rescan the blockchain.
+	ConsensusChange(i int) (ConsensusChange, error)
 
 	// ConsensusSetSubscribe will subscribe another module to the consensus
 	// set. Every time that there is a change to the consensus set, an update
--- a/modules/consensus/subscribers.go
+++ b/modules/consensus/subscribers.go
@@ -1,6 +1,8 @@
 package consensus
 
 import (
+	"errors"
+
 	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/types"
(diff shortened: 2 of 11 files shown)


modules/consensus.go
```diff
@@ -136,6 +136,11 @@ type ConsensusSet interface {
 	// run any required closing routines.
 	Close() error
 
+	// ConsensusChange returns the consensus change that occured at the input
+	// index. An error is returned if there is not a change at the provided
+	// index.
+	ConsensusChange(int) (ConsensusChange, error)
```
lukechampine: What does "index" mean in this context? How is the caller supposed to know what index to use?

modules/consensus/subscribers.go
```diff
@@ -1,10 +1,95 @@
...
+		}
+		for _, sfod := range appliedNode.siafundOutputDiffs {
+			cc.SiafundOutputDiffs = append(cc.SiafundOutputDiffs, sfod)
+		}
+		for _, dscod := range appliedNode.delayedSiacoinOutputDiffs {
+			cc.DelayedSiacoinOutputDiffs = append(cc.DelayedSiacoinOutputDiffs, dscod)
+		}
+		for _, sfpd := range appliedNode.siafundPoolDiffs {
+			cc.SiafundPoolDiffs = append(cc.SiafundPoolDiffs, sfpd)
+		}
```
lukechampine: can't you just use `copy` for these?

siad/daemon.go
```diff
@@ -24,6 +28,20 @@ func startDaemon() error {
 	// Establish multithreading.
 	runtime.GOMAXPROCS(runtime.NumCPU())
 
+	// Print a startup message.
+	fmt.Println("siad is loading")
```
lukechampine: I don't think startup messages are necessary. Maybe leave the `"loaded in %d seconds"`, but inside a `if config.Siad.Profile` block.

lukechampine:
In general I'm uneasy about including profiling code in the master branch. When you profile, you're targeting specific areas of the codebase; it's the sort of thing you do individually when working on an optimization branch. e.g. this code is specific to optimizing the initial blockchain load, and tracking goroutine leakage. When I do profiling, I'll probably want to focus on something else.

If I were doing profiling, my workflow would be:
- install https://github.com/pkg/profile and add it to `main`
- write any other custom profiling code
- run siad and reproduce the behavior I want to profile
- optimize
- remove profiling code and commit

siad/daemon.go
DavidVorick: It's going to stay there until it's almost immediate. It's really painful to start siad an have no idea when it's done loading.

DavidVorick:
I'll have to look into the http/pprof package, it could potentially replace the logging file I set up.

Profiling code is something that I want to distribute to users. If they are complaining about it being slow or having other problems, we can have them run siad with profiling enabled, then we can figure out what went wrong.

modules/consensus/subscribers.go
DavidVorick: the code would get more complicated because there can be multiple reverted and applied blocks in a single consensus change. You'd have to make sure you're copying to the right index, and you'd have to do a loop in advance to figure out how long to make the slice in the first place.

The code will simplify if we switch to 1 consensus change == 1 block, which I think will be a good change to make but I also think it'll throw off the block explorer.

PR #671, opened 2015-07-11
Encoding coverage/documentation

lukechampine:


DavidVorick:
You broke something, it won't marshal []types.SiacoinInput anymore, probably not other types either, but that's the one that caused the build to fail for most of the packages.

PR #675, opened 2015-07-13
Add basic benchmark for consensus adding blocks

triazo:
My worries:

 - Adding a block does not actually seem to be constant time. For now the go benchmark doesn't seem to notice since this doesn't show until you have many more blocks.

 - The testing benchmark records memory usage, but doesn't seem to have any way of outputting it easily, so I manully ran the benchmark and made it a test function. (running with `go test ./modules/consensus/ -run=Bench -v -tags='debug testing'`)


modules/consensus/accept_bench_test.go
```diff
@@ -0,0 +1,46 @@
...
+func TestBenchEmptyBlocks(t *testing.T) {
+	br := testing.Benchmark(func(b *testing.B) {
+		err := benchmarkEmptyBlocks(b)
+		if err != nil {
+			b.Fatal(err.Error())
+		}
+	})
+	fmt.Printf("Operations:    %d\n", br.N)
+	fmt.Printf("Time per op:   %d ns\n", br.NsPerOp())
+	fmt.Printf("Allocs per op: %d allocs, %d bytes\n", br.AllocsPerOp(), br.AllocedBytesPerOp())
```
lukechampine: use `t.Log` instead

modules/consensus/accept_bench_test.go
```diff
@@ -0,0 +1,46 @@
...
+        bt.StopTimer()
+
+        // The test dir will be reset each time the benchmark
+        // is done.
+        cst, err := createConsensusSetTester("BenchmarkEmptyBlocks")
+        if err != nil {
+                return errors.New("Error creating tester: "+err.Error())
+        }
+
+        for j:= 0; j < bt.N; j++ {
```
lukechampine: go fmt

modules/consensus/accept_bench_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+        return nil
+}
+
+// TestBenchEmptyBlocks
+func TestBenchEmptyBlocks(t *testing.T) {
+	br := testing.Benchmark(func(b *testing.B) {
+		err := benchmarkEmptyBlocks(b)
+		if err != nil {
+			b.Fatal(err.Error())
```
lukechampine: the `.Error` call is implicit

modules/consensus/accept_bench_test.go
```diff
@@ -0,0 +1,46 @@
+package consensus
+
+import (
+        "errors"
+        "fmt"
+        "testing"
+)
+
+// benchmarkEmptyBlocks ta
```
lukechampine: ?

triazo:
This also measures the memory allocations for the initial startup, which is probably a significant portion of the memory usage. We should be able to make use of the information regardless though.

DavidVorick:
if you have debugging enabled, adding a block is going to be linear time in the size of the consensus set, I believe. This happens because we hash the whole consensus set every block.

I guess we need a better way to toggle the debugging intensity, this isn't the only place where it's bad.

modules/consensus/accept_bench_test.go
```diff
@@ -0,0 +1,44 @@
...
+	for j := 0; j < bt.N; j++ {
+		b, _ := cst.miner.FindBlock()
+		bt.StartTimer()
+		err := cst.cs.AcceptBlock(b)
+		if err != nil {
+			errstr := fmt.Sprintf("Error accepting %d: %s", j, err.Error())
+			return errors.New(errstr)
+		}
+		cst.csUpdateWait()
+		bt.StopTimer()
```
DavidVorick: I don't think it makes sense to include the csUpdateWait() inside of the timing, because that's going to stall until all modules have received the update and processed it. You're timing the modules as well as the consensus package.

DavidVorick: Actually, because of changes I'm planning to make to the subscription model, it might be better not to use a full consensus tester, which I believe also contains a wallet and a transaction pool.

I had some ideas for other ways to approach this benchmark, but our debugging tools and DoS protections actually make benchmarking the consensus package difficult. It'd be easier to start with the lower level functions like StandaloneValid, applyMaturedSiacoinOutputs, and such.

One way around the debugging tools would be to have a `make bench` which runs with debugging disabled.

modules/blockexplorer/addblock_test.go
```diff
@@ -19,7 +20,7 @@ func genHashNum(x byte) crypto.Hash {
 
 // Add a couple blocks to the database, then perform lookups to see if
 // they were added and crossed referenced correctly
-func (et *explorerTester) testAddBlock(t *testing.T) {
+func (et *explorerTester) testAddBlock(t *testing.T) error {
```
DavidVorick: this function shouldn't need to take `t` as an argument anymore

modules/blockexplorer/blockexplorer_test.go
```diff
@@ -38,18 +39,12 @@ type explorerTester struct {
 func (et *explorerTester) csUpdateWait() {
 	<-et.csUpdateChan
 	<-et.beUpdateChan
-	et.tpUpdateWait()
-}
-
-// tpUpdateWait blocks until a transaction pool update has propagated to all
-// modules.
-func (et *explorerTester) tpUpdateWait() {
```
DavidVorick: getting rid of tpUpdateWait is a little weird because all testers have it. You're also probably going to want to have it if you start building custom blocks at some point.

modules/consensus/accept_bench_test.go
DavidVorick: I thought about this test a bit more, and I think it's just in the wrong package. AcceptBlock, because of the subscription stuff, necessarily impacts all of the other modules, and therefore isn't very useful for benchmarking the consensus package alone.

`siad` might be a better place for this benchmark.

modules/consensus/accept_bench_test.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	return nil
+}
+
+// BenchmarkEmptyBlocks is a wrapper for benchmarkEmptyBlocks, which
+// handles error catching
+func BenchmarkEmptyBlocks(b *testing.B) {
```
DavidVorick: The current codebase is pretty inconsistent with this stuff, but the only reason you should have a lowercase function is if you plan on using it in future benchmarks.

But, because of the nature of benchmarking, I don't think that really makes sense. Just having everything be in one function is probably best.

triazo:
So I'll leave this PR out there for a bit, but as it is a PR into an old branch it needs to be remade to pull into master. I do think there is some value in having such a benchmark specifically for consensus that does not wait for all the other modules to update, but a similar more comprehensive benchmark will also be useful.

DavidVorick:
If you want a benchmark that targets *just* the consensus package, you're going to need to use a consensus module that has no subscribers (no miner, wallet, etc.), which means you are going to need to manually make all of the blocks.

The miner has a function `SolveBlock` that will take any block and solve it for a given target, it's completely possible to make that benchmark (just have the miner subscribed to a different consensus set), but it's a lot of manual coding.

DavidVorick:
Is this PR into the right branch? You probably want to merge it with master.

PR #676, opened 2015-07-14
UPnP

### LINKED ISSUE #609: UPnP support
Enough users have had trouble announcing that I think this should be a high priority for v0.4.

Reference implementations: 
- https://github.com/jackpal/Taipei-Torrent/blob/master/torrent/upnp.go
- https://github.com/prestonTao/upnp
- https://github.com/huin/goupnp

lukechampine:
UPnP support is here! In local tests it works like a charm. However, the architecture feels a bit hacky to me, and there's certainly room for improvement.

The biggest issue here is that UPnP discovery is performed during initialization, the same way `ExternalIP` used to be calculated. This has the advantage of centralizing our UPnP code, which is used by both the gateway and the host. But it prevents a common piece of functionality seen in other P2P apps, which is the ability to toggle UPnP support during runtime.

It wouldn't be too tough to add `ToggleUPnP` functions to the gateway and host (and a `/daemon/toggleupnp` API call), but it wouldn't be pretty. I think we should wait until the use case is more apparent.

fixes #609


DavidVorick:
Most p2p apps support toggling? What is the purpose?

modules/host/host.go
```diff
@@ -105,8 +106,19 @@ func New(cs *consensus.ConsensusSet, hdb modules.HostDB, tpool modules.Transacti
 	if err != nil {
 		return nil, err
 	}
+	host, err := modules.IGD.ExternalIP()
+	if err != nil {
+		// log error? should be fatal
```
DavidVorick: I don't think it should be fatal. We want people to include the host module by default, even if they are having trouble finding their external IP address (we don't want siad to exit just because the host can't be used).

Logging an error is probably okay.

Does IGD.ExternalIP work if upnp is not supported?

modules/host/host.go
```diff
@@ -105,8 +106,19 @@ func New(cs *consensus.ConsensusSet, hdb modules.HostDB, tpool modules.Transacti
...
+		host = "::1"
+	}
 	_, port, _ := net.SplitHostPort(h.listener.Addr().String())
-	h.myAddr = modules.NetAddress(net.JoinHostPort(modules.ExternalIP, port))
+	h.myAddr = modules.NetAddress(net.JoinHostPort(host, port))
+
+	// Forward port, if possible
+	portInt, _ := strconv.Atoi(port)
+	if portErr := modules.IGD.Forward(uint16(portInt), "Sia Host"); portErr != nil {
+		// log error? should be fatal if err != modules.ErrNoUPnP
```
DavidVorick: logging an error here should also be okay.

modules/upnp.go
```diff
@@ -0,0 +1,65 @@
...
+var ErrNoUPnP = errors.New("UPnP not available")
+
+// IGD is an interface to the local Internet Gateway Device. If no such device
+// is available (e.g. if the user's router does not support UPnP), a dummy IGD
+// is returned instead.
+var IGD = func() upnp.IGD {
+	// always use the loopback address during testing
+	if build.Release == "testing" {
+		return &dummyIGD{"::1"}
+	}
```
DavidVorick: yuck

modules/host/host.go
lukechampine: if you're expecting your port to be forwarded and it isn't, that's definitely a problem.
Then you run into another issue though, which is when users forwarded their ports manually. Then they shouldn't care if UPnP fails, but if it does it will return an error. So then we need a `--no-upnp` flag to disable the check or something. Or we just fail silently/log it in all cases. But then we need a way to inform the user whether or not they're reachable, which so far has had its own bugs...

DavidVorick: let upnp fail silently (with a logging error), and add a logger to the host to record the error.

any function that requires a forwarded port (such as announcing, or even setting the host to accept contracts) should return an error if the 'have I forwarded my ports' test hasn't been passed.

lukechampine:
ok, better now (I think).
One thing that could still be improved is device discovery time. Right now we're calling `Discover` every time we want to do something UPnP related, which in my tests takes ~2.5 seconds. If we cached the location of the device, we could cut that way down (~0.05 seconds). But it's not a huge issue because you only make three calls total: two at startup and one at shutdown.

PR #677, opened 2015-07-15
Consensus api

DavidVorick:
The api paradigm was switched up, and I moved over the consensus api calls. The one thing we haven't figured out is how to return an error string when the api call fails. I'm not sure what is idiomatic, but I would want to be as idiomatic and RESTful as possible.

I added a profiling package that should make profiling a lot easier. Siad, when passed a profiling flag, will make a log with basic profiling information.

There are additional calls that make it easy to start and stop memory and cpu profiling.

http/pprof is not used anywhere currently. Profiling is not done by default. The log sizes are pretty small even if you do turn on profiling.

### DIFF
--- a/api/consensus.go
+++ b/api/consensus.go
@@ -4,15 +4,85 @@ import (
 	"net/http"
 
 	"github.com/NebulousLabs/Sia/build"
+	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/types"
 )
 
+// The ConsensusSetStatus struct contains general information about the
+// consensus set, with tags to have idiomatic json encodings.
+type ConsensusSetStatus struct {
+	Height       types.BlockHeight `json:"height"`
+	CurrentBlock types.BlockID     `json:"currentblock"`
+	Target       types.Target      `json:"target"`
+}
+
+// DEPRECATED
+//
+// The ConsensusInfo struct contains general information about the consensus
+// set.
 type ConsensusInfo struct {
 	Height       types.BlockHeight
 	CurrentBlock types.BlockID
 	Target       types.Target
 }
 
+// consensusHandlerGET handles a GET request to /consensus.
+func (srv *Server) consensusHandlerGET(w http.ResponseWriter, req *http.Request) {
+	currentTarget, exists := srv.cs.ChildTarget(srv.currentBlock.ID())
+	if build.DEBUG {
+		if !exists {
+			panic("server has nonexistent current block")
+		}
+	}
+
+	writeJSON(w, ConsensusSetStatus{
+		types.BlockHeight(srv.blockchainHeight),
+		srv.currentBlock.ID(),
+		currentTarget,
+	})
+}
+
+// consensusHandler handles the API calls to /consensus.
+func (srv *Server) consensusHandler(w http.ResponseWriter, req *http.Request) {
+	lockID := srv.mu.RLock()
+	defer srv.mu.RUnlock(lockID)
+
+	if req.Method == "" || req.Method == "GET" {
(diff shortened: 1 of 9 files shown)


profile/profile.go
```diff
@@ -0,0 +1,117 @@
...
+		logFile, err := os.OpenFile(filepath.Join(profileDir, "continuousProfiling.log"), os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0660)
+		if err != nil {
+			fmt.Println("Goroutine logging failed:", err)
+			return
+		}
+		log := log.New(logFile, "", log.Ldate|log.Ltime|log.Lmicroseconds|log.Lshortfile)
+		log.Println("Continuous profiling started.")
+
+		// Inifinite loop to print out the goroutine count.
+		sleepTime := time.Second * 10
```
lukechampine: if sleep time is exponential you can probably start much smaller than 10 secs.

siad/daemon.go
```diff
@@ -91,17 +82,16 @@ func startDaemon() error {
 	// that daemon startup has completed. A gofunc is used with the hope that
 	// srv.Serve() will start running before the value is sent down the
 	// channel.
+	//
+	// TODO: There are better ways to approach this.
```
lukechampine: most projects don't test main, which I don't take too much issue with. We don't really test main either.

profile/profile.go
DavidVorick: inifinite

inidicies

inidefinite

siad/daemon.go
DavidVorick: I like having the code coverage.

The main tests can't be very detailed but I like having them. I have some new ideas for how to approach them.

profile/profile.go
DavidVorick: There's no need to start smaller than 10 seconds, because things haven't even started up at that point.

DavidVorick:
ok fixed

api/consensus.go
```diff
@@ -25,25 +35,69 @@ func (srv *Server) consensusStatusHandler(w http.ResponseWriter, req *http.Reque
...
-func (srv *Server) consensusSynchronizeHandler(w http.ResponseWriter, req *http.Request) {
+// consensusHandler handles the API calls to /consensus.
+func (srv *Server) consensusHandler(w http.ResponseWriter, req *http.Request) {
+	lockID := srv.mu.RLock()
+	defer srv.mu.RUnlock(lockID)
+
+	if req.Method == "" || req.Method == "GET" {
+		srv.consensusHandlerGET(w, req)
+		return
+	}
```
lukechampine: if you're planning on doing this for the rest of the API functions, you should write a wrapper function.
```go
func methodWrapper(fn http.HandleFunc, method string) http.HandleFunc {
    return func(w http.ResponseWriter, req *http.Request) {
        if req.Method != method {
            writeError(w, "bad method; use " + method)
            return
        }
        fn(w, req)
    }
}
```

DavidVorick: Not that simple unfortunately. Some methods do different things depending on the method.

if req.Method == "" || req.Method == "GET" {
    doGETStuff()
} else if req.Method == "PUT" {
    doPUTStuff()
}
writeError(w, "unrecognized method used when calling /fancy/call")

DavidVorick: as an example, `/gateway/peers` has different defined behavior for all of GET, POST, and DELETE

lukechampine: ah, true. This is where using a real framework like [Martini](https://github.com/go-martini/martini) or [Gorilla](http://www.gorillatoolkit.org/pkg/mux) would be useful.

PR #678, opened 2015-07-15
Keep track of the seen times of the most recent 144 blocks

triazo:
As the update model doesn't differentiate between loaded blocks and recieved ones, I used the time that the module was created as a switch for to use the seen timestamp or the block timestamp.

### DIFF
--- a/modules/explorer/addblock_test.go
+++ b/modules/explorer/addblock_test.go
@@ -1,6 +1,7 @@
 package explorer
 
 import (
+	"errors"
 	"testing"
 
 	"github.com/NebulousLabs/Sia/crypto"
@@ -19,7 +20,7 @@ func genHashNum(x byte) crypto.Hash {
 
 // Add a couple blocks to the database, then perform lookups to see if
 // they were added and crossed referenced correctly
-func (et *explorerTester) testAddBlock(t *testing.T) {
+func (et *explorerTester) testAddBlock(t *testing.T) error {
 	// This block will *NOT* be valid, but should contain
 	// addresses that can cross reference each other.
 	b1 := types.Block{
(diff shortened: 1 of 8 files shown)


DavidVorick:
you forgot to `go fmt`

lukechampine:
add a "go fmt on save" binding to your vim, it's the best way

triazo:
I moved the explorer test fix to this PR as it made more sense to keep the scope of the request more limited.

modules/explorer/addblock_test.go
```diff
@@ -19,7 +20,7 @@ func genHashNum(x byte) crypto.Hash {
 
 // Add a couple blocks to the database, then perform lookups to see if
 // they were added and crossed referenced correctly
-func (et *explorerTester) testAddBlock(t *testing.T) {
+func (et *explorerTester) testAddBlock(t *testing.T) error {
```
DavidVorick: still don't need a `t *testing.T` in the function definition

modules/explorer/update.go
```diff
@@ -35,6 +36,14 @@ func (e *Explorer) ReceiveConsensusSetUpdate(cc modules.ConsensusChange) {
 	// Handle incoming blocks
 	for _, block := range cc.AppliedBlocks {
 		e.blockchainHeight += 1
+
+		// Add the current time to seenTimes
+		if time.Unix(int64(block.Timestamp), 0).Before(e.startTime) {
+			e.seenTimes[e.blockchainHeight%types.BlockHeight(len(e.seenTimes))] = time.Unix(int64(block.Timestamp), 0)
+		} else {
+			e.seenTimes[e.blockchainHeight%types.BlockHeight(len(e.seenTimes))] = time.Now()
+		}
```
DavidVorick: this won't always work because block timestamps can be in the past (only a bit though), but I don't see a better way to do it.

After the explorer has been running for 36+ hours it shouldn't be a problem at all.

PR #679, opened 2015-07-19
Tpool batching

DavidVorick:
The transaction pool now takes much less memory, which was the original goal of this PR. In the process of doing that, I threw out the existing subscription model and replaced it with something much easier to understand and harder to screw up.

The most important change is that the transaction pool now accepts sets of transactions instead of single transactions. A 'set' must have all the required parents, and none of the transactions in the set can already be in the transaction pool. Transaction sets cannot depend on eachother, but instead should be combined into a larger set.

There are plans to eventually add a way to extend existing sets with new transactions, but that will not be implemented until there's a clear need for it.

There are some new pitfalls, it's not very well documented at the moment. (perhaps I should do that this PR)

Then the transaction builder in the wallet got a complete facelift to match the 'transaction set' way of doing things. I also think that overall it's just cleaner and structured better now.

I believe those are the only changes, they just touched massive parts of the codebase.


api/server_test.go
```diff
@@ -45,34 +45,9 @@ type serverTester struct {
...
 	server *Server
 
-	csUpdateChan     <-chan struct{}
-	hostUpdateChan   <-chan struct{}
-	hostdbUpdateChan <-chan struct{}
-	minerUpdateChan  <-chan struct{}
-	renterUpdateChan <-chan struct{}
-	tpoolUpdateChan  <-chan struct{}
-	walletUpdateChan <-chan struct{}
-
```
lukechampine: praise the lord

modules/consensus/accept_test.go
```diff
@@ -650,17 +639,15 @@ func (cst *consensusSetTester) testFileContractsBlocks() error {
 		}},
 	}
 	copy(txn.StorageProofs[0].Segment[:], segment)
-	err = cst.tpool.AcceptTransaction(txn)
+	err = cst.tpool.AcceptTransactionSet([]types.Transaction{txn})
```
lukechampine: you know, if you made `AcceptTransactionSet` variadic, then the n=1 case would be much cleaner...guess it depends on how frequent it is.

modules/consensus/subscribe.go
```diff
@@ -149,41 +104,36 @@ func (cs *ConsensusSet) updateSubscribers(revertedNodes []*blockNode, appliedNod
...
-		select {
-		case subscriber <- struct{}{}:
-		default:
-		}
+	cc, err := cs.computeConsensusChange(len(cs.changeLog) - 1)
+	if err != nil && build.DEBUG {
+		panic(err)
+	}
+	for _, subscriber := range cs.subscribers {
+		subscriber.ReceiveConsensusSetUpdate(cc)
```
lukechampine: perhaps in light of the new subscription model this should be called `ProcessConsensusChange`?

modules/consensus/subscribe.go
```diff
@@ -149,41 +104,36 @@ func (cs *ConsensusSet) updateSubscribers(revertedNodes []*blockNode, appliedNod
...
-
 // ConsensusSetSubscribe accepts a new subscriber who will receive a call to
 // ReceiveConsensusSetUpdate every time there is a change in the consensus set.
 func (cs *ConsensusSet) ConsensusSetSubscribe(subscriber modules.ConsensusSetSubscriber) {
-	c := make(chan struct{}, modules.NotifyBuffer)
-	c <- struct{}{} // Notify subscriber about the genesis block.
 	id := cs.mu.Lock()
-	cs.subscriptions = append(cs.subscriptions, c)
+	cs.subscribers = append(cs.subscribers, subscriber)
+	for i := 0; i < len(cs.changeLog); i++ {
```
lukechampine: could use `i := range cs.changeLog` here

modules/miner/update.go
```diff
@@ -60,3 +34,26 @@ func (m *Miner) ReceiveTransactionPoolUpdate(cc modules.ConsensusChange, unconfi
...
+			break
+		}
+		remainingSize -= len(encoding.Marshal(unconfirmedTransactions[0]))
+		if remainingSize < 0 {
+			break
+		}
+
+		m.transactions = append(m.transactions, unconfirmedTransactions[0])
+		unconfirmedTransactions = unconfirmedTransactions[1:]
+	}
```
lukechampine: is there a reason for not using a traditional range loop here? It's not clear to me.

modules/modules.go
```diff
@@ -18,7 +14,7 @@ func init() {
 	if build.Release == "dev" {
 		SafeMutexDelay = 3 * time.Second
 	} else if build.Release == "standard" {
-		SafeMutexDelay = 8 * time.Second
+		SafeMutexDelay = 60 * time.Second
```
lukechampine: wowza

modules/consensus/accept_test.go
DavidVorick: probably frequent enough

I worry about clarity though. If you make it a variadic function, I feel like the connotation is that the objects are not necessarily fundamentally related. But here, the objects are part of a set, and must have a specific order.

modules/modules.go
DavidVorick: The lock will be held until all modules have processed the consensus change, because ReceiveConsensusSetUpdate is now a blocking function. This will slow things down a lot.

I also don't want false positives whatsoever in our production code, but don't feel like the codebase is ready to be shipped without the safelocks.

modules/transactionpool/accept.go
```diff
@@ -10,273 +10,183 @@ import (
...
 
 var (
-	ErrLargeTransactionPool = errors.New("transaction size limit reached within pool")
-	ErrLowMinerFees         = errors.New("transaction miner fees too low to be accepted")
-	TransactionMinFee       = types.NewCurrency64(3).Mul(types.SiacoinPrecision)
+	ErrObjectConflict      = errors.New("transaction set conflicts with an existing transaction set")
+	ErrFullTransactionPool = errors.New("transaction pool cannot accept more transactions")
+	ErrLowMinerFees        = errors.New("transaction set needs more miner fees to be accepted")
+
+	TransactionMinFee = types.NewCurrency64(2).Mul(types.SiacoinPrecision)
```
lukechampine: I feel like `SiacoinPrecision` isn't a very useful name. Maybe `UnitSiacoin` or something.

modules/transactionpool/transactionpool.go
```diff
@@ -9,86 +9,53 @@ import (
...
-	subscribers             []chan struct{}
-
-	mu *sync.RWMutex
-}
+type (
+	// ObjectIDs are the IDs of objects such as siacoin outputs and file
+	// contracts, and are used to see if there are conflicts or overlaps within
+	// the transaction pool. A TransactionSetID is the hash of a transaction
+	// set.
+	ObjectID         crypto.Hash
```
lukechampine: yeesh, `ObjectID` is dangerously generic. It's only slightly better than just using `crypto.Hash`.

modules/transactionpool/accept.go
```diff
@@ -10,273 +10,183 @@ import (
...
 	}
 
-	// Add the transaction to the pool, notify all subscribers, and broadcast
-	// the transaction.
-	tp.addTransactionToPool(t)
-	tp.updateSubscribers(modules.ConsensusChange{}, tp.transactionList, tp.unconfirmedSiacoinOutputDiffs())
-	go tp.gateway.Broadcast("RelayTransaction", t)
-	return
+	// Notify subscribers and broadcast the transaction set.
+	go tp.gateway.Broadcast("RelayTransactionSet", ts)
```
lukechampine: this is a pretty breaking change. To smooth the transition, we will probably want to retain support for `"RelayTransaction"` for at least one version. Even then, new nodes won't be able to rely their transaction sets to old nodes.

modules/transactionpool/accept.go
```diff
@@ -10,273 +10,183 @@ import (
...
+
+// DEPRECATED
+func (tp *TransactionPool) AcceptTransaction(t types.Transaction) error {
+	return tp.AcceptTransactionSet([]types.Transaction{t})
+}
+
+// DEPRECATED
+func (tp *TransactionPool) RelayTransaction(conn modules.PeerConn) error {
+	return tp.RelayTransactionSet(conn)
+}
```
lukechampine: this will fail because the object will not decode properly.

modules/transactionpool/standard.go
```diff
@@ -12,22 +12,43 @@ import (
...
-//   Purpose: Leave room for more involved soft-forks in the future.
+//		The arbitrary data field can be used to orchestrate soft-forks to Sia
+//		that add features. Legacy miners are at risk of creating invalid blocks
+//		if they include arbitrary data which has meanings that the legacy miner
+//		doesn't understand.
+//
+// Rule: The transaction set size is limited.
+//		A group of dependent transactions
+//		cannot exceed 100kb to limit how quickly the transaction pool can be filled
+//		with new transactions.
```
lukechampine: formatting

modules/transactionpool/standard.go
```diff
@@ -12,22 +12,43 @@ import (
...
+	ErrLargeTransaction    = errors.New("transaction is too large")
+	ErrLargeTransactionSet = errors.New("transaction set is too large")
 )
 
 // checkUnlockConditions looks at the UnlockConditions and verifies that all
 // public keys are recognized. Unrecognized public keys are automatically
-// accpeted as valid by the state, but rejected by the transaction pool. This
-// allows new types of keys to be added via a softfork without alienating all
-// of the older nodes.
+// accpeted as valid by the consnensus set, but rejected by the transaction
```
lukechampine: typo

modules/transactionpool/subscribe.go
```diff
@@ -0,0 +1,46 @@
...
+		subscriber.ReceiveUpdatedUnconfirmedTransactions(txns, cc)
+	}
+}
+
+// updateSubscribersConsensus sends a new consensus change to all subscribers.
+func (tp *TransactionPool) updateSubscribersConsensus(cc modules.ConsensusChange) {
+	for _, subscriber := range tp.subscribers {
+		subscriber.ReceiveConsensusSetUpdate(cc)
+	}
+}
```
lukechampine: so the wallet subscribes to the tpool but not the consensusset, and receives consensus changes through the tpool? I suppose that makes sense; it keeps the two updates synchronized.

DavidVorick: yes

modules/transactionpool/transactionpool.go
```diff
@@ -9,86 +9,53 @@ import (
...
+		// another transaction set spends. This is done to minimize the
+		// computation and memory load on the transaction pool. Dependent
+		// transactions should be lumped into a single transaction set.
+		//
+		// transactionSetDiffs map form a transaction set id to the set of
+		// diffs that resulted from the transaction set.
+		knownObjects        map[ObjectID]struct{}
+		transactionSets     map[TransactionSetID][]types.Transaction
+		transactionSetDiffs map[TransactionSetID]modules.ConsensusChange
+		databaseSize        int
```
lukechampine: oh man, so few maps...and they aren't even maps of maps :P

why call it a database? why not `poolSize`? Or even just `size`?

modules/transactionpool/transactionpool.go
```diff
@@ -107,24 +74,33 @@ func New(cs modules.ConsensusSet, g modules.Gateway) (tp *TransactionPool, err e
...
+		// change will then have an index of '0'.
+		consensusChangeIndex: -1,
 
-		mu: sync.New(modules.SafeMutexDelay, 1),
+		mu: sync.New(modules.SafeMutexDelay, 5),
 	}
 
 	// Register RPCs
-	g.RegisterRPC("RelayTransaction", tp.RelayTransaction)
+	g.RegisterRPC("RelayTransactionSet", tp.RelayTransactionSet)
```
lukechampine: to retain support we'll want to register both

modules/wallet.go
```diff
@@ -22,68 +24,102 @@ type WalletInfo struct {
...
-	// AddSiafundOutput adds an output to a transaction. It returns the transaction
-	// with index of the output that got added.
-	AddSiafundOutput(id string, output types.SiafundOutput) (types.Transaction, uint64, error)
+	// RegisterTransaction takes a transaction and its parents returns an id
+	// that can be used to modify the transaction. The most typical call is
+	// 'RegisterTransaction(types.Transaction{}, nil)', which registers a new
+	// transaction that doesn't have any parents. The id that gets returned is
+	// not a types.TransactionID, it is an int and is only useful within the
+	// transaction builder.
+	RegisterTransaction(t types.Transaction, parents []types.Transaction) (id int, err error)
```
lukechampine: this model is still a bit weird to me -- getting back an object and calling methods on it feels cleaner to me. i.e.:
```go
tx := wallet.NewTransaction()
tx.AddFileContract(fc)
tx.AddMinerFee(types.NewCurrency64(10))
// etc.
tx.Submit() // or maybe wallet.Submit(tx)
```

modules/transactionpool/transactionpool.go
DavidVorick: unfortunately the internals of the transaction pool changed dramatically enough that retaining support for the old call would be harmful.

The problem is that on the old system, you would submit a parent and then submit the child. On the new system, once you submit a parent you /cannot/ submit a child. It'll be rejected. This could cause weird problems and lost coins.

There might be a way to have the new tpool automatically detect if an incoming transaction is a child to an existing known transaction set, and then extend the transaction set. But there are lots of corner cases involved with doing that, and I think the code complexity doesn't really justify the use-case.

modules/wallet.go
```diff
@@ -126,7 +158,7 @@ type Wallet interface {
 
 	// SendSiagSiafunds sends siafunds to another address. The siacoins stored
 	// in the siafunds are sent to an address in the wallet.
-	SendSiagSiafunds(amount types.Currency, dest types.UnlockHash, keyfiles []string) (types.Transaction, error)
+	SendSiagSiafunds(amount types.Currency, dest types.UnlockHash, keyfiles []string) ([]types.Transaction, error)
```
lukechampine: I feel like the wallet shouldn't need to know about `siag`. Not a big deal though.

modules/wallet/update.go
```diff
@@ -90,13 +73,35 @@ func (w *Wallet) ReceiveTransactionPoolUpdate(cc modules.ConsensusChange, _ []ty
...
-			w.siafundPool = cc.SiafundPoolDiffs[len(cc.SiafundPoolDiffs)-1].Adjusted
-		} else {
-			w.siafundPool = cc.SiafundPoolDiffs[len(cc.SiafundPoolDiffs)-1].Previous
+}
+
+// ReceiveTransactionPoolUpdate gets all of the changes in the confirmed and
+// unconfirmed set and uses them to update the balance and transaction history
+// of the wallet.
+func (w *Wallet) ReceiveUpdatedUnconfirmedTransactions(_ []types.Transaction, unconfirmedCC modules.ConsensusChange) {
+	// TODO: Restructure whole wallet.
```
lukechampine: oof

modules/consensus/accept_test.go
lukechampine: I think it would be pretty rare to see anything other than `Accept(tx)` or `Accept(txns...)`. But perhaps it's not worth the loss in clarity.

lukechampine:
overall I'm quite happy with this. My primary concerns are the breaking RPC change and the whole `FundTransaction` model. If we decide not to support old txn RPCs, we should think about how best to manage that transition. The `FundTransaction` stuff is wallet-specific so it doesn't need to be addressed here, but if you're working on the wallet next, it'd be nice to implement it then. (If we decide it makes sense.)

modules/miner/update.go
DavidVorick: just wasn't being creative enough.

modules/transactionpool/transactionpool.go
DavidVorick: it is generic though. There's probably a better name but I haven't yet thought of it.

modules/transactionpool/accept.go
DavidVorick: I was thinking we could have one of the bootstrap nodes playing middleman or something.

New nodes can send their transaction sets to old nodes just fine by sending them out 1 minute apart. The problem is that old nodes can't get their transactions to the new nodes. We'll need at least a handful of miners to not upgrade until most of the standard users have upgraded.

In the interim I might think of a better solution.

modules/transactionpool/standard.go
DavidVorick: two tabs kept it from being super close to the '/' for 4 spaces.

modules/transactionpool/transactionpool.go
DavidVorick: oh. I waffled on the name for a while before settling on transactionList

so the right name I think would be listSize or transactionListSize

modules/wallet.go
DavidVorick: NewTransaction is a bad name because it's not a transaction, it's a transaction + a bunch of additional data.

I'm not sure why I didn't think about returning a transaction object, the register model is kind of clunky, AND you have to worry about leaks. I think it would be possible to adjust the builder to return an object instead of an id, but that's a bigger change than I want to tackle in this PR.

modules/wallet.go
DavidVorick: the two are going to be merged at some point.

modules/wallet/update.go
DavidVorick: :)

it needs a lot of love

DavidVorick:
There are two major concerns I have about the current transaction pool.

1. The transition is going to be difficult, because the old way of doing things is broken in a fairly severe way, and I'm not sure how to bridge the gap. I think though, we could have the bootstrap nodes run some double-code or something that can take a set of transactions every 60s or so and connect them into one giant transaction set.

But then I'd also need to implement safe-replacement (a transaction set can be replaced if it is being replaced by a superset)

2. When there is a reorg, there's not an equivalently clean way to pull out reverted transactions which still belong in a set. I know of a conceptually simple model that will run in n^2 in the number of transactions per block, but ideally there'd be a linear solution.

It seems like there's room to improve the transaction builder too, returning an in-progress object instead of an id.

DavidVorick:
`peers_test.go:195: disconnect failed: not connected to that node`

We need to do something about that test

lukechampine:
try deleting line 184 of peers_test.go

DavidVorick:
Maybe worked. I guess it will take a few builds to be certain.

PR #680, opened 2015-07-19
Benchmark for adding empty blocks

triazo:
A copy of #675 into the correct branch this time.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,6 +77,8 @@ test-v: clean fmt REBUILD
 	go test -race -v -short -tags='debug testing' -timeout=15s $(pkgs)
 test-long: clean fmt REBUILD
 	go test -v -race -tags='testing debug' -timeout=180s $(pkgs)
+bench: clean fmt REBUILD
+	go test -tags='testing' -timeout=600s -run=XXX -bench=. $(pkgs)
 cover: clean REBUILD
 	@mkdir -p cover/modules
 	@for package in $(pkgs); do                                                                                     \
--- a/Makefile
+++ b/Makefile
@@ -78,7 +78,7 @@ test-v: clean fmt REBUILD
 test-long: clean fmt REBUILD
 	go test -v -race -tags='testing debug' -timeout=180s $(pkgs)
 bench: clean fmt REBUILD
-	go test -tags='testing' -timeout=600s -run=XXX -bench=. $(pkgs)
+	go test -tags='testing' -timeout=120s -run=XXX -bench=. $(pkgs)
 cover: clean REBUILD
 	@mkdir -p cover/modules
 	@for package in $(pkgs); do                                                                                     \
--- a/modules/consensus/accept_bench_test.go
+++ b/modules/consensus/accept_bench_test.go
@@ -1,0 +1,73 @@
+package consensus
+
+import (
+	"errors"
+	"fmt"
+	"path/filepath"
+	"testing"
+
+	"github.com/NebulousLabs/Sia/build"
+	"github.com/NebulousLabs/Sia/modules"
+	"github.com/NebulousLabs/Sia/modules/gateway"
+)
+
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -76,6 +76,8 @@ test-v: clean fmt REBUILD
 	go test -race -v -short -tags='debug testing' -timeout=15s $(pkgs)
 test-long: clean fmt REBUILD
 	go test -v -race -tags='testing debug' -timeout=180s $(pkgs)
+bench: clean fmt REBUILD
+	go test -tags='testing' -timeout=600s -run=XXX -bench=. $(pkgs)
```
DavidVorick: 10 minute timeout is too much, for now it should probably be between 60s and 120s

modules/consensus/accept_bench_test.go
```diff
@@ -0,0 +1,73 @@
...
+		}
+		bt.StartTimer()
+		err = cs.AcceptBlock(b)
+		if err != nil {
+			errstr := fmt.Sprintf("Error accepting %d for timing: %s", j, err.Error())
+			return errors.New(errstr)
+		}
+		bt.StopTimer()
+		cst.csUpdateWait()
+	}
```
DavidVorick: Do you think you can adjust this test so that you're only calling `AcceptBlock` on a consensus set with no subscribers? In the new subscription model, updating subscribers is a blocking call.

triazo: It calls accept block twice, the first time for the miner.  Only the second call is timed though, and that consensus set doesn't have any subscribers. If there's an easier way to ensure correct blocks without a second consensus set let me know.

DavidVorick: oh sorry I read that too fast. This looks fine.

I don't think there's an easy way to do it otherwise without adding a second consensus set.

modules/consensus/accept_bench_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+	"github.com/NebulousLabs/Sia/build"
+	"github.com/NebulousLabs/Sia/modules"
+	"github.com/NebulousLabs/Sia/modules/gateway"
+)
+
+// benchmarkEmptyBlocks is a benchmark that mines many blocks, and
+// measures how long it takes to add them to the consensusset
+func benchmarkEmptyBlocks(bt *testing.B) error {
+	bt.StopTimer()
```
DavidVorick: not needed, instead use bt.ResetTimer() before the main loop starts.

modules/consensus/accept_bench_test.go
```diff
@@ -0,0 +1,73 @@
...
+		return errors.New("Error creating tester: " + err.Error())
+	}
+	for _, bID := range cst.cs.currentPath[1:] {
+		err = cs.AcceptBlock(cst.cs.blockMap[bID].block)
+		if err != nil {
+			return err
+		}
+	}
+
+	for j := 0; j < bt.N; j++ {
```
DavidVorick: throw a `bt.ResetTimer()` right here

Also, idiomatic go typically has the benchmark variable called `b`, it should not be replaced by the block variable. You should call the `testing.B` `b`, and then just call blocks `block`. It's a bit more verbose but will be less confusing to golang programmers unfamiliar with the project.

DavidVorick:
bump

modules/consensus/accept_bench_test.go
```diff
@@ -0,0 +1,71 @@
...
+			return errors.New(errstr)
+		}
+		b.StartTimer()
+		err = cs.AcceptBlock(block)
+		if err != nil {
+			errstr := fmt.Sprintf("Error accepting %d for timing: %s", j, err.Error())
+			return errors.New(errstr)
+		}
+	}
+
```
DavidVorick: probably no need for this newline. I'm inconsistent on using it.

PR #681, opened 2015-07-19
Block manager remembers more blocks and uses less memory

JoshVorick:
In order to reduce the miner's memory usage, the map of headers to blocks is now a map of header to block references and a map of headers to random transactions.

This way, the miner can create multiple headers for work for each block it stores in memory. Because blocks are can be over a megabyte, this greatly reduces the amount of memory used by the miner. 

Every `headersPerBlockMemory` calls to `getHeaderForWork()`, a new block is generated. Then for the next `headersPerBlockMemory` blocks, the same block is used, but with a different random transaction. This gives the block a unique header for mining. In total, the miner now stores `headerForWorkMemory / headersPerBlockMemory` blocks.


modules/miner/blockmanager.go
```diff
@@ -69,21 +69,53 @@ func (m *Miner) HeaderForWork() (types.BlockHeader, types.Target) {
...
-	block, target := m.blockForWork()
+	var header types.BlockHeader
+	var randTxn types.Transaction
+	var block *types.Block
 
-	// Save a mapping between the block and its header, replacing the block
-	// that was stored 'headerForWorkMemory' requests ago.
+	if m.memProgress%headersPerBlockMemory == 0 {
+		// Grab a new block
+		block = new(types.Block)
```
lukechampine: why are you allocated memory here?

modules/miner/miner.go
```diff
@@ -16,9 +16,15 @@ const (
 
 	// headerForWorkMemory is the number of previous calls to 'headerForWork'
 	// that are remembered.
-	headerForWorkMemory = 50
+	headerForWorkMemory = 20
+
+	// headersPerBlockMemory is the number of headers created by changing the
+	// random transaction of a block before updating the block to contain any
+	// new transactions
+	headersPerBlockMemory = 5
```
DavidVorick: is this number too low? I was thinking X00

modules/miner/miner.go
```diff
@@ -16,9 +16,15 @@ const (
...
 )
 
+// TODO: docstring
```
DavidVorick: need to add a docstring in this PR :P

modules/miner/miner.go
```diff
@@ -33,14 +39,15 @@ type Miner struct {
...
 
-	// A list of blocks that have been through SubmitBlock.
+	// A list of blocks that have been submitted through SubmitBlock.
 	blocksFound []types.BlockID
 
 	// BlockManager variables. The BlockManager passes out and receives unique
 	// block headers on each call, these variables help to map the received
 	// block header to the original block. The headers are passed instead of
 	// the block because a full block is 2mb and is a lot to send over http.
-	blockMem    map[types.BlockHeader]types.Block
```
DavidVorick: need to update the comment above these variables to better explain what they do.

modules/miner/miner.go
```diff
@@ -68,8 +75,7 @@ type Miner struct {
 }
 
 // New returns a ready-to-go miner that is not mining.
-func New(cs modules.ConsensusSet, tpool modules.TransactionPool, w modules.Wallet, persistDir string) (*Miner, error) {
-	// Create the miner and its dependencies.
+func New(cs modules.ConsensusSet, tpool modules.TransactionPool, w modules.Wallet, persistDir string) (*Miner, error) { // Create the miner and its dependencies.
```
DavidVorick: This comment moved to a weird place.

modules/miner/blockmanager.go
```diff
@@ -69,21 +69,53 @@ func (m *Miner) HeaderForWork() (types.BlockHeader, types.Target) {
 	lockID := m.mu.Lock()
 	defer m.mu.Unlock(lockID)
 
-	// Grab a block for work.
-	block, target := m.blockForWork()
+	var header types.BlockHeader
+	var randTxn types.Transaction
+	var block *types.Block
 
-	// Save a mapping between the block and its header, replacing the block
```
DavidVorick: typically (within this project - idk about other projects), the variables are put directly above the codeblock that first uses them.

```
// comment explaining vars + codeblock
var x y
var z a
if {
} else {
}
```

modules/miner/blockmanager.go
```diff
@@ -69,21 +69,53 @@ func (m *Miner) HeaderForWork() (types.BlockHeader, types.Target) {
...
-	// that was stored 'headerForWorkMemory' requests ago.
+	if m.memProgress%headersPerBlockMemory == 0 {
+		// Grab a new block
+		block = new(types.Block)
+		*block, _ = m.blockForWork()
```
DavidVorick: I don't think you need the `block = new(types.Block)` line

modules/miner/blockmanager.go
```diff
@@ -69,21 +69,53 @@ func (m *Miner) HeaderForWork() (types.BlockHeader, types.Target) {
...
+		header = block.Header()
+		randTxn = block.Transactions[0]
+	} else {
+		// Set block to previous block and create a randTxn
+		block = m.blockMem[m.headerMem[m.memProgress-1]]
```
DavidVorick: there's probably a way to structure the code so that the `-1` is not needed

modules/miner/blockmanager.go
```diff
@@ -69,21 +69,53 @@ func (m *Miner) HeaderForWork() (types.BlockHeader, types.Target) {
...
+
+		randBytes, _ := crypto.RandBytes(types.SpecifierLen)
+		randTxn = types.Transaction{
+			ArbitraryData: [][]byte{append(modules.PrefixNonSia[:], randBytes...)},
+		}
+
+		// Overwrite the old block's random transaction
+		blockTransactions := append([]types.Transaction{randTxn}, block.Transactions[1:]...)
```
DavidVorick: There's an easier way to do this:

`block.Transactions[0].ArbitraryData[0] = randBytes`

Then you don't have to make a transaction or do the (somewhat expensive) append call.

modules/miner/blockmanager.go
```diff
@@ -69,21 +69,53 @@ func (m *Miner) HeaderForWork() (types.BlockHeader, types.Target) {
...
+
+		// Assemble the block
+		newBlock := types.Block{
+			ParentID:     block.ParentID,
+			Timestamp:    block.Timestamp,
+			MinerPayouts: block.MinerPayouts,
+			Transactions: blockTransactions,
+		}
+		header = newBlock.Header()
```
DavidVorick: additionally, is there an explicit reason that you need to make a new block? Isn't it sufficient to just call `block.Header()`?

modules/miner/miner.go
```diff
@@ -16,9 +16,15 @@ const (
 
 	// headerForWorkMemory is the number of previous calls to 'headerForWork'
 	// that are remembered.
-	headerForWorkMemory = 50
+	headerForWorkMemory = 20
```
DavidVorick: I believe that this is too low. I was thinking X000.

Ideally, the total amount of memeory in use would be less than 50mb. I guess that gives you room for 20 blocks, and then if you do 100 headers per block you get to 2000 headers total in memory.

I think that's good enough...

modules/miner/miner.go
JoshVorick: I made it low for testing. I didn't mean to commit such a low number. What should I change it too?
Maybe 500 and 20? so it will store up to 25 blocks?

modules/miner/blockmanager.go
```diff
@@ -69,21 +69,53 @@ func (m *Miner) HeaderForWork() (types.BlockHeader, types.Target) {
...
+
+	// Save a mapping between the block and its header as well as the
+	// random transaction and its header, replacing the block that was
+	// stored 'headerForWorkMemory' requests ago.
 	delete(m.blockMem, m.headerMem[m.memProgress])
-	m.blockMem[block.Header()] = block
-	m.headerMem[m.memProgress] = block.Header()
+	delete(m.randTxnMem, m.headerMem[m.memProgress])
+	m.blockMem[header] = block
+	m.randTxnMem[header] = randTxn
```
DavidVorick: you probably don't need the whole transaction, just the arb data would be fine

modules/miner/miner.go
```diff
@@ -108,8 +114,9 @@ func New(cs modules.ConsensusSet, tpool modules.TransactionPool, w modules.Walle
 		earliestTimestamp: earliestTimestamp,
 		address:           addr,
 
-		blockMem:  make(map[types.BlockHeader]types.Block),
-		headerMem: make([]types.BlockHeader, headerForWorkMemory),
+		blockMem:   make(map[types.BlockHeader]*types.Block),
+		randTxnMem: make(map[types.BlockHeader]types.Transaction),
```
DavidVorick: I put this somewhere else, but you probably don't need the whole transaction, just the arb data will be fine.

modules/miner/blockmanager.go
```diff
@@ -117,13 +149,28 @@ func (m *Miner) SubmitHeader(bh types.BlockHeader) error {
 	lookupBH := bh
 	lookupBH.Nonce = zeroNonce
 	lockID := m.mu.Lock()
-	b, exists := m.blockMem[lookupBH]
-	m.mu.Unlock(lockID)
-	if !exists {
+	b, bExists := m.blockMem[lookupBH]
+	randTxn, txnExists := m.randTxnMem[lookupBH]
+
+	if !bExists || !txnExists {
```
DavidVorick: newline at line 154 is not needed

modules/miner/blockmanager.go
```diff
@@ -117,13 +149,28 @@ func (m *Miner) SubmitHeader(bh types.BlockHeader) error {
...
+		m.mu.Unlock(lockID)
 		err := errors.New("block header returned late - block was cleared from memory")
 		m.log.Println("ERROR:", err)
 		return err
 	}
-	b.Nonce = bh.Nonce
-	return m.SubmitBlock(b)
+	// Reset block memory
+	m.memProgress = 0
```
DavidVorick: why is mem progress set to 0 here?

DavidVorick:
ok, done with first round of comments. I'm happy to see the unit tests, but I'm going to hold off on looking at them until you've finished with my current comments, since some of it might change.

modules/miner/blockmanager.go
lukechampine: is `ArbitraryData` guaranteed to be non-nil?

DavidVorick: it is, except for the first time you make the block, but that runs a different codepath anyway.

modules/miner/blockmanager.go
JoshVorick: If I don't allocate new memory, block won't be pointing to any memory, so it will seg fualt when I try to copy the new block into *block. 

Another way to do it would be:
`tempBlock, _ := m.blockForWork()`
`block = &tempBlock`

modules/miner/blockmanager.go
JoshVorick: I think I ended up creating a new block because trying to change the random transaction inside the old block caused a race condition. If I use `block.Transactions[0].ArbitraryData[0]` instead, I should be able to just call `block.Header()`

modules/miner/miner.go
JoshVorick: I think I'm going to change the constants to `headerForWorkMemory` and `blockForWorkMemory` (max number of blocks to be stored) because I think it will make code more understandable.

modules/miner/blockmanager.go
JoshVorick: I guess i was thinking that since a block was submitted, the miner should start over since no old blocks will be valid any more. But along that line of thinking, I should have cleared `blockMem`, `randTxnMem` and `headers`

Should I worry about deleting old block memory?

modules/miner/blockmanager.go
lukechampine: I suggest not using a pointer.
```go
var block types.Block
if m.memProgress%headersPerBlockMemory == 0 {
    block, _ = m.blockForWork()
} else {
    block = *m.blockMem[m.headerMem[m.memProgress-1]]
}
```
perhaps you could even make blockMem a `[]types.Block` instead of a `[]*types.Block`

DavidVorick:
```go
bfw, _ := m.blockForWork()
block = &bfw
```

modules/miner/blockmanager.go
DavidVorick: ok. For future reference, you should throw a comment about the race condition that you are trying to avoid.

modules/miner/blockmanager.go
DavidVorick: Had to think about this a bit.

I think that you should not delete the old memory. If a miner calls submitHeader for a block that is old (which means that an orphan was mined), a special 'stale err' should be returned so that the miner knows it's happening. I also don't think you need to set the memProgress to 0, you can just keep going from where you currently are. You just have to make sure that you get a new blockForWork at the current memory location, which is not necessarily divisible by anything.

Setting it to '0' is a convenient way to do that, but it might be harmful if it means miners are going to have less warning that they are occasionally mining stale blocks.

modules/miner/blockmanager.go
DavidVorick: @lukechampine 

We are trying to keep 100s of different headers in memory. If each block is up to 2mb (depending on the amount of transaction overlap), that's X00mb to make the list. Headers on the other hand are only 16 bytes each.

I think that using a []types.Block is going to be very expensive once we start producing larger blocks.

modules/miner/miner.go
DavidVorick: max headers + max blocks?

makes sense to me.

DavidVorick:
don't forget to drop a comment after you push code - that way I get notifications

modules/miner/miner.go
```diff
@@ -15,10 +15,19 @@ const (
...
-	headerForWorkMemory = 50
+	// that are remembered. Additionally, 'headerForWork' will only poll for a
+	// new block every 'headerForWorkMemory / blockForWorkMemory' times it is
+	// called. This reduces the amount of memory used, but comes at the cost of
+	// not always having the most recent transactions
+	headerForWorkMemory = 1000
+
+	// blockForWorkMemory is the maximum number of blocks the miner will store
+	// Blocks take up to 2 megabytes of memory, so it is important to keep a cap
+	blockForWorkMemory = 50
```
DavidVorick: 50 is too many. Blocks are big and I would like to keep Sia as lightweight as possible. I think 20 would be better.

modules/miner/blockmanager_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+	"github.com/NebulousLabs/Sia/types"
+)
+
+// reconstructBlock reconstructs a block given the miner and the header
+func reconstructBlock(m *Miner, header types.BlockHeader) (block *types.Block, err error) {
+	block, exists := m.blockMem[header]
+	if !exists {
+		err = errors.New("Header is either invalid or too old")
+		return
```
DavidVorick: `return nil, errors.New(...)`

modules/miner/blockmanager_test.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	arbData := m.arbDataMem[header]
+
+	block.Transactions[0].ArbitraryData[0] = arbData
+	return
```
DavidVorick: `return block, nil`

modules/miner/blockmanager_test.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+// mineHeader takes a header, and nonce grinds it. It returns
+// the header with a nonce that solves the corresponding block
+func mineHeader(m *Miner, header types.BlockHeader) (bh types.BlockHeader, err error) {
+	b, err := reconstructBlock(m, header)
+	if err != nil {
+		return
```
DavidVorick: `return types.BlockHeader{}, nil`

modules/miner/blockmanager_test.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	*b, _ = m.SolveBlock(*b, m.target)
+	bh = b.Header()
+	return
```
DavidVorick: `return bh, nil`

modules/miner/blockmanager_test.go
```diff
@@ -0,0 +1,105 @@
...
+	stored := make(map[*types.Block]bool)
+	for _, value := range mt.miner.blockMem {
+		if !stored[value] {
+			stored[value] = true
+			numUniqueBlocks++
+		}
+	}
+	if numUniqueBlocks != blockForWorkMemory {
+		t.Error("Miner is storing an incorrect number of blocks ", numUniqueBlocks)
+	}
```
DavidVorick: this strikes me as an add way to do this, but it does look like it works as intended.

modules/miner/blockmanager_test.go
```diff
@@ -0,0 +1,105 @@
...
+// TestBlockManager creates a miner, then polls the Miner for block
+// headers to mine. It ensures that the blockmanager properly hands
+// out headers, then reconstructs the blocks
+func TestBlockManager(t *testing.T) {
+	mt, err := createMinerTester("TestBlockManager")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	headers := make([]types.BlockHeader, 2*headerForWorkMemory)
```
DavidVorick: you should probably write a comment indicating the reason it's 2*headerForWorkMemory

DavidVorick:
The changes look good, just a few things left.

DavidVorick:
Actually, I have 1 further concern.

What happens if a miner is being super conservative with requesting headers, and is only pulling 1 header every 5 minutes or something? The transaction list for the blocks they are mining is going to be out of date most of the time - the miner will only get a new block every 25 minutes (or when there's a change to the blockchain)

Perhaps there should be 3 conditions under which a new block will be formed: 
1. you've requested (headerForWorkMemory / blockForWorkMemory) blocks
2. new block from consensus
3. more than 30s has passed since the previous block was requested.

modules/miner/blockmanager_test.go
JoshVorick: Does Sia not use Go's named return values? (e.g. func foo() (a, b int) { ... } )

DavidVorick: for the most part, no. It's easier to make mistakes when you have an empty `return` line, for example you might not realize that you forgot to clear an error that was checked but not returned (this has happened a few times). If you explicitly return `nil`, it's clear that you intended to return nil.

This also removes guesswork from reviewing. A reviewer may look at such a mistake and think "he meant to do that". Even though the review has noticed the mistake, it is not fixed.

I think there are a few places where it makes sense. Mostly, really small functions where not using named returns would require adding multiple lines of additional code.

DavidVorick: Coming back to this, you'll find examples of Sia using named returns all over the codebase. This is because we stopped using named return values within the past few weeks, much of the codebase hasn't been ported to the new strategy.

JoshVorick:
Okay, I think I addressed everything. The testing might benefit from another test or two, but for now it should be fine.

PR #682, opened 2015-07-20
Api test tuning

DavidVorick:
I changed the api functions slightly, and then expanding the synchronize test for the consensus package.

I'm led to believe, based on the difficulty I'm having constructing a situation where two peers are connected but aren't synchronized, that the synchronize command is actually not needed. It was definitely needed in the past, but doesn't seem to be needed any more.

### DIFF
--- a/api/consensus.go
+++ b/api/consensus.go
@@ -67,7 +67,10 @@ func (srv *Server) consensusSynchronizeHandlerGET(w http.ResponseWriter, req *ht
 	if err != nil {
 		writeError(w, "System error", http.StatusInternalServerError)
 	}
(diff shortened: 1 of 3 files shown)


lukechampine:
Yes, we can probably remove `synchronize`, as well as the `threadedResynchronize` loop in consensus/synchronize.go

lukechampine:
You should make an issue for the `TestIntegration`/`TestUnit` plan

DavidVorick:
I want to keep the synchronize call for now, simply because it's a nice failsafe and it's not hurting anything to keep it.

I'll go ahead and strip out the `threadedResynchronize` loop though. It'll be nice to drop another goroutine.

PR #683, opened 2015-07-21
Tpool testing

DavidVorick:
Some testing.

Most notable: changes to server_test.go, Makefile

everything else is pretty standard and not-fancy. tpool test coverage is nearly at 80% already, personally I think it's good enough for the present.

I'm now going to be working on the wallet.

### DIFF
--- a/modules/transactionpool/transactionpool.go
+++ b/modules/transactionpool/transactionpool.go
@@ -9,6 +9,11 @@ import (
 	"github.com/NebulousLabs/Sia/types"
 )
 
+var (
+	errNilCS      = errors.New("transaction pool cannot initialize with a nil consensus set")
+	errNilGateway = errors.New("transaction pool cannot initialize with a nil gateway")
+)
+
 type (
 	// ObjectIDs are the IDs of objects such as siacoin outputs and file
 	// contracts, and are used to see if there are conflicts or overlaps within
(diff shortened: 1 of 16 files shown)


modules/transactionpool/transactionpool.go
```diff
@@ -58,19 +63,17 @@ type (
...
 	// Check that the input modules are non-nil.
 	if cs == nil {
-		err = errors.New("transaction pool cannot use a nil state")
-		return
+		return nil, errNilCS
 	}
 	if g == nil {
-		err = errors.New("transaction pool cannot use a nil gateway")
-		return
+		return nil, errNilGateway
```
lukechampine: these checks still feel unnecessary to me.

lukechampine:
looks good, I think splitting up the unit/integration tests was a good idea and we came up with a clean way of implementing it.

modules/transactionpool/transactionpool.go
DavidVorick: too bad :)

I think it does pay off to be meticulous. Just have to be meticulous in the right places. I do feel like 100% code coverage is one of those places.

lukechampine: that's what I'm saying -- this is a wrong place. It's a waste of effort. Outside of testing, the `New` functions are called in exactly one place, where it is easy to verify that their inputs are non-nil. These checks are superfluous and force us to write useless tests for the sake of coverage.
If we do somehow pass `nil` to `New`, the result will almost always be an immediate panic. The risk of such a bug making it into a release is minute, and would be discovered quickly by users anyway.
I would give some examples of where nil arguments are far more likely, but I'm afraid you would want to add checks for all of those too.

PR #684, opened 2015-07-21
Replace BlockDB with new SetDB and load processed blocks from disk

triazo:


modules/consensus/consensusset.go
```diff
@@ -50,6 +48,11 @@ type ConsensusSet struct {
 	// up loading blocks from memory.
 	verificationRigor verificationRigor
```
DavidVorick: Are we able to delete the verificationRigor variable?

modules/consensus/consensusset.go
```diff
@@ -50,6 +48,11 @@ type ConsensusSet struct {
...
 
+	// updatePath is a flag that determines if a block node will
+	// be added to the path. It should be false when loading the
+	// current path from disk and true otherwise
+	updatePath bool
```
DavidVorick: Is this a temporary variable? The current path will not need to be loaded into memory once everything has been moved to a database, right?

modules/consensus/consensusset.go
```diff
@@ -187,15 +191,10 @@ func New(gateway modules.Gateway, saveDir string) (*ConsensusSet, error) {
...
-	if build.Release == "testing" && testing.Short() {
-		cs.db = persist.NilDB
-	} else {
-		// Otherwise, try to load an existing database from disk.
-		err = cs.load(saveDir)
-		if err != nil {
-			return nil, err
-		}
+	// Try to load an existing database from disk.
+	err = cs.load(saveDir)
```
DavidVorick: I'm a little surprised that the saveDir is not already in the consensus struct. I guess there's not an explicit need for it, but as we do more things like logging I think it would make sense. (even though, I think, it would still only be used at startup.

No need to do anything about it this PR, just an observation.

modules/consensus/consensusset.go
triazo: yes, updatePath is there only because blocknodes still need to be run through commitDiffSet, which alters the path. Once all the diffs are saved to disk this won't be necessary.

modules/consensus/persist.go
```diff
@@ -4,40 +4,41 @@ import (
...
 	if height == 0 {
 		// add genesis block
-		cs.db = db
-		return db.AddBlock(cs.blockRoot.block)
+		return db.addPath(cs.blockRoot.block)
 	}
 
 	// Check that the db's genesis block matches our genesis block.
-	b, err := db.Block(0)
+	bID, err := db.path(0)
```
DavidVorick: more typically I just use `bid` since `bID` looks kinda silly.

modules/consensus/consensusset.go
triazo: We should be able to. I for now just to keep the volume of changes small, but as it currently stands it is never set as anything other than full, so removing it shouldn't pose a problem.

modules/consensus/setdb.go
```diff
@@ -0,0 +1,226 @@
...
+	"github.com/NebulousLabs/Sia/crypto"
+	"github.com/NebulousLabs/Sia/encoding"
+	"github.com/NebulousLabs/Sia/modules"
+	"github.com/NebulousLabs/Sia/persist"
+	"github.com/NebulousLabs/Sia/types"
+	"github.com/boltdb/bolt"
+)
+
+var meta = persist.Metadata{
+	Version: "0.1",
```
DavidVorick: Put the version at the current release. It helps in the future when we're trying to figure out where things changed and when compatibility was updated or broken.

"0.4.0"

modules/consensus/setdb.go
```diff
@@ -0,0 +1,226 @@
...
+	Header:  "Consensus Set Backing Database",
```
DavidVorick: "Consensus Set Database" is probably sufficient.

modules/consensus/setdb.go
```diff
@@ -0,0 +1,226 @@
...
+	errBadSetInsert = errors.New("attempting to add an already existing item to the consensus set")
+	errNilBucket    = errors.New("using a bucket that does not exist")
+	errNilItem      = errors.New("Requested item does not exist")
+)
+
+// setDB is a wrapper around the persist bolt db which backs the
+// consensus set
+type setDB struct {
+	*persist.BoltDatabase
+}
```
DavidVorick: why is a wrapper used?

modules/consensus/setdb.go
```diff
@@ -0,0 +1,226 @@
...
+
+	DiffsGenerated            bool
+	SiacoinOutputDiffs        []modules.SiacoinOutputDiff
+	FileContractDiffs         []modules.FileContractDiff
+	SiafundOutputDiffs        []modules.SiafundOutputDiff
+	DelayedSiacoinOutputDiffs []modules.DelayedSiacoinOutputDiff
+	SiafundPoolDiffs          []modules.SiafundPoolDiff
+
+	ConsensusSetHash crypto.Hash
+}
```
DavidVorick: I know we are still in the transition period, but eventually this struct should fully replace `blockNode`

modules/consensus/setdb.go
```diff
@@ -0,0 +1,226 @@
...
+
+func bnToPb(bn blockNode) processedBlock {
```
DavidVorick: need docstring

DavidVorick: you should also add a comment indicating that this code is DEPRECATED

modules/consensus/setdb.go
```diff
@@ -0,0 +1,226 @@
...
+		siafundOutputDiffs:        pb.SiafundOutputDiffs,
+		delayedSiacoinOutputDiffs: pb.DelayedSiacoinOutputDiffs,
+		siafundPoolDiffs:          pb.SiafundPoolDiffs,
+
+		consensusSetHash: pb.ConsensusSetHash,
+	}
+	return bn
+}
+
+func openDB(filename string) (*setDB, error) {
```
DavidVorick: need docstring

modules/consensus/setdb.go
```diff
@@ -0,0 +1,226 @@
...
+	db, err := persist.OpenDatabase(meta, filename)
+	if err != nil {
+		return nil, err
+	}
+
+	var buckets []string = []string{
+		"Path", "BlockMap",
```
DavidVorick: putting these on their own line probably makes more sense for readability.

DavidVorick:
```
[]string{
    "Path",
    "BlockMap",
}
```

modules/consensus/setdb.go
```diff
@@ -0,0 +1,226 @@
...
+// addItem and getItem are part of consensus due to stricter error conditions.
+//
+// addItem should only be called from this file, and adds a new item
+// to the database
+func (db *setDB) addItem(bucket string, key, value interface{}) error {
+	v := encoding.Marshal(value)
+	k := encoding.Marshal(key)
+	return db.Update(func(tx *bolt.Tx) error {
+		b := tx.Bucket([]byte(bucket))
+		if build.DEBUG {
```
DavidVorick: need to write a docstring for the sanity check

modules/consensus/setdb.go
```diff
@@ -0,0 +1,226 @@
...
+			if err != nil {
+				return err
+			}
+		}
+		return nil
+	})
+	return &setDB{db}, nil
+}
+
+// addItem and getItem are part of consensus due to stricter error conditions.
```
DavidVorick: This part of the docstring I believe should come second

modules/consensus/setdb.go
```diff
@@ -0,0 +1,226 @@
...
+		}
+		return b.Put(k, v)
+	})
+}
+
+func (db *setDB) getItem(bucket string, key interface{}) (item []byte, err error) {
+	k := encoding.Marshal(key)
+	err = db.View(func(tx *bolt.Tx) error {
+		b := tx.Bucket([]byte(bucket))
+		if build.DEBUG {
```
DavidVorick: need docstring on the sanity check

modules/consensus/setdb.go
```diff
@@ -0,0 +1,226 @@
...
+			i := b.Get(k)
+			if i != nil {
+				panic(errBadSetInsert)
+			}
+		}
+		return b.Put(k, v)
+	})
+}
+
+func (db *setDB) getItem(bucket string, key interface{}) (item []byte, err error) {
```
DavidVorick: need docstring

modules/consensus/setdb.go
```diff
@@ -0,0 +1,226 @@
...
+			}
+		}
+		return nil
+	})
+	return
+}
+
+// AddBlock inserts a block into the database at the "end" of the chain, i.e.
+// the current height + 1.
+func (db *setDB) addPath(block types.Block) error {
```
DavidVorick: docstring is out of date

DavidVorick: `addToPath` might be a better name

modules/consensus/setdb.go
```diff
@@ -0,0 +1,226 @@
...
+	return db.Update(func(tx *bolt.Tx) error {
+		b := tx.Bucket([]byte("Path"))
+		key := encoding.EncUint64(uint64(b.Stats().KeyN))
+		return b.Put(key, value)
+	})
+}
+
+// RemoveBlock removes a block from the "end" of the chain, i.e. the block
+// with the largest height.
+func (db *setDB) rmPath() error {
```
DavidVorick: docstring is out of date

DavidVorick: removeFromPath might be a better name.

modules/consensus/setdb.go
```diff
@@ -0,0 +1,226 @@
...
+	return db.Update(func(tx *bolt.Tx) error {
+		b := tx.Bucket([]byte("Path"))
+		key := encoding.EncUint64(uint64(b.Stats().KeyN - 1))
+		return b.Delete(key)
+	})
+}
+
+// path retreives the block id of a block at a given hegiht from the path
+func (db *setDB) path(h types.BlockHeight) (id types.BlockID, err error) {
```
DavidVorick: I don't think that this function name is appropriate.

perhaps blockAtHeight, which is also a bit awkward but at least it's more descriptive.

modules/consensus/setdb.go
triazo: The wrapper lets us have consensus specific member functions. Also there will eventually be additional items in the struct like error flags.

modules/consensus/setdb.go
```diff
@@ -0,0 +1,226 @@
...
+}
+
+// pathHeight returns the size of the current path
+func (db *setDB) pathHeight() (types.BlockHeight, error) {
+	h, err := db.BucketSize("Path")
+	return types.BlockHeight(h), err
+}
+
+// addBlockMap adds a block node to the block map
+func (db *setDB) addBlockMap(bn blockNode) error {
```
DavidVorick: wouldn't `addBlockNode` make more sense?

Or even, `addProcessedBlock`?

modules/consensus/setdb.go
triazo: Since the path is essentially a stack, maybe pushPath and popPath?

modules/consensus/setdb.go
```diff
@@ -0,0 +1,226 @@
...
+	return db.addItem("BlockMap", bn.block.ID(), bnToPb(bn))
+}
+
+// Function needs to have access to the blockMap inside consensusSet
+// because bnFromPb requires the parent ID. This could be fixed at
+// some point, as this function really should be part of setDB
+func (db *setDB) getBlockMap(id types.BlockID) (*processedBlock, error) {
```
DavidVorick: docstring is incorrect. `Function` should be the function name.

function name should probably be `getProcessedBlock`

modules/consensus/setdb.go
triazo: my thought with the naming was to have, for each bucket, add{bucketname}, rm{bucketname}, and get{bucketname} functions. I worry that leaving the bucket name out may leave some ambiguity as to exactly which data structure you are modifying. For example, `addProcessedBlock` may imply that it will add to all diffs due to it being processed.

modules/consensus/setdb.go
triazo: maybe getPath to keep with the convention below?

modules/consensus/consensusset.go
DavidVorick: ok, you should tag it with a `// DEPRECATED` comment

modules/consensus/setdb.go
DavidVorick: that's probably better

modules/consensus/setdb.go
DavidVorick: hmm.

I don't think that it's ambigious, because you are calling db.AddProcessedBlock. The consensus package distinguishes between `add` and `commit` as well.

PR #685, opened 2015-07-22
further refine the txn builder

DavidVorick:
I ditched the registration system in favor of returning an interface, as you suggested. I consider it to be a pretty big improvement.

### DIFF
--- a/modules/host/announce.go
+++ b/modules/host/announce.go
@@ -27,35 +27,23 @@ func ping(addr modules.NetAddress) bool {
 
 // announce creates an announcement transaction and submits it to the network.
 func (h *Host) announce(addr modules.NetAddress) error {
-	// create the transaction that will hold the announcement
-	id, err := h.wallet.RegisterTransaction(types.Transaction{}, nil)
-	if err != nil {
-		return err
-	}
-
-	// create and encode the announcement and add it to the arbitrary data of
-	// the transaction.
+	// Create a transaction with a host announcement.
+	txnBuilder := h.wallet.RegisterTransaction(types.Transaction{}, nil)
 	announcement := encoding.Marshal(modules.HostAnnouncement{
 		IPAddress: addr,
 	})
-	_, err = h.wallet.AddArbitraryData(id, append(modules.PrefixHostAnnouncement[:], announcement...))
-	if err != nil {
-		return err
-	}
-	txnSet, err := h.wallet.SignTransaction(id, true)
-	if err != nil {
-		return err
-	}
+	_ = txnBuilder.AddArbitraryData(append(modules.PrefixHostAnnouncement[:], announcement...))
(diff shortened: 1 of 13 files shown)


lukechampine:
`RegisterTransaction(types.Transaction{}, nil)` is so common that I would strongly consider defining a `StartTransaction` method for convenience. As far as I can tell, `RegisterTransaction` is only called with a preexisting transaction in once place, and is never called with parents at all.

modules/host/announce.go
```diff
@@ -27,35 +27,23 @@ func ping(addr modules.NetAddress) bool {
...
-	if err != nil {
-		return err
-	}
-	txnSet, err := h.wallet.SignTransaction(id, true)
-	if err != nil {
-		return err
-	}
+	_ = txnBuilder.AddArbitraryData(append(modules.PrefixHostAnnouncement[:], announcement...))
+	txn, parents := txnBuilder.View()
+	txnSet := append(parents, txn)
```
lukechampine: isn't `parents` guaranteed to be nil here?

modules/host/negotiate.go
```diff
@@ -234,7 +228,7 @@ func (h *Host) rpcContract(conn net.Conn) (err error) {
 	}
 
 	// Add the collateral to the transaction, but do not sign the transaction.
-	collateralTxn, txnID, err := h.addCollateral(unsignedTxn, terms)
+	collateralTxn, txnBuilder, err := h.addCollateral(unsignedTxn, terms)
```
lukechampine: would probably be cleaner if you created the TransactionBuilder before, and passed it `addCollateral` for modification.

modules/wallet.go
```diff
@@ -24,115 +24,97 @@ type WalletInfo struct {
...
+	//
+	// If the whole transaction flag  is set to true, then the whole
+	// transaction flag will be set in the covered fields object. If the whole
+	// transaction flag is set to false, then the covered fields object will
+	// cover all fields that have already been added to the transaction, but
+	// will also leave room for more fields to be added.
+	Sign(wholeTransaction bool) ([]types.Transaction, error)
+
+	// View returns the incomplete transaction along with all of its parents.
+	View() (txn types.Transaction, parents []types.Transaction)
```
lukechampine: why does `View` need to return the parents? I'm not totally clear on this, but it doesn't seem like the TransactionBuilder ever modifies the set of parents.

modules/host/announce.go
DavidVorick: I think it makes sense to use the builder idiomatically.

modules/host/negotiate.go
DavidVorick: I agree, but the host and renter files are going to be changing dramatically enough that I'm not really worried about maintaining them.

DavidVorick:
updated

PR #686, opened 2015-07-23
Path db

triazo:
In short, moves the currentPath structure in consensus to disk, and fixes associated bugs.

Proposal for error handling: panic on read, and error catch on writes. We do a fair amount of verification when modifying the structures, so it makes sense to use this existing structure. Ideally, database inconsistencies would be caught on write where they can be handled more elegantly (e.g. rejecting a call to acceptBlock).

Also worth noting that this is on top of the Set db pull request.

### DIFF
--- a/modules/consensus/consensusset.go
+++ b/modules/consensus/consensusset.go
@@ -159,7 +159,6 @@ func New(gateway modules.Gateway, saveDir string) (*ConsensusSet, error) {
 	cs.blockMap[genesisBlock.ID()] = cs.blockRoot
 
 	// Fill out the consensus information for the genesis block.
-	cs.currentPath[0] = genesisBlock.ID()
 	cs.siacoinOutputs[genesisBlock.MinerPayoutID(0)] = types.SiacoinOutput{
 		Value:      types.CalculateCoinbase(0),
 		UnlockHash: types.ZeroUnlockHash,
@@ -177,6 +176,7 @@ func New(gateway modules.Gateway, saveDir string) (*ConsensusSet, error) {
 		cs.commitSiafundOutputDiff(sfod, modules.DiffApply)
 		cs.blockRoot.siafundOutputDiffs = append(cs.blockRoot.siafundOutputDiffs, sfod)
 	}
+	cs.currentPath[0] = genesisBlock.ID()
 
 	// Send out genesis block update.
 	cs.updateSubscribers(nil, []*blockNode{cs.blockRoot})
--- a/modules/consensus/consensusset.go
+++ b/modules/consensus/consensusset.go
@@ -10,7 +10,6 @@ import (
 	"errors"
 	"os"
 
-	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/sync"
 	"github.com/NebulousLabs/Sia/types"
@@ -177,9 +176,6 @@ func New(gateway modules.Gateway, saveDir string) (*ConsensusSet, error) {
 		}
 		cs.commitSiafundOutputDiff(sfod, modules.DiffApply)
 		cs.blockRoot.siafundOutputDiffs = append(cs.blockRoot.siafundOutputDiffs, sfod)
-	}
-	if build.DEBUG {
-		cs.blockRoot.consensusSetHash = cs.consensusSetHash()
 	}
 
 	// Send out genesis block update.
--- a/modules/consensus/consensusset.go
+++ b/modules/consensus/consensusset.go
@@ -52,6 +52,12 @@ type ConsensusSet struct {
 	// current path from disk and true otherwise
 	updatePath bool // DEPRECATED
 
+	// blocksLoaded is the number of blocks that have been loaded
+	// from memory. This variable only exists while some
+	// structures are still in memory, and should always equal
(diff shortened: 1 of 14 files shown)


lukechampine:
The `BulkUpdate` and associated types were not removed from a previous PR. Could you remove them here?

triazo:
Good idea. Also removed InsertIntoDatabase, as that is not used either.

DavidVorick:
Looks like the rebase got screwed up somewhere.

When this happens to me, I find it's easiest to just make a new branch from master, and then cherry-pick all the commits that I want to keep. Luke I think frowns upon this method, but it gets the job done.

modules/consensus/consensusset.go
```diff
@@ -53,6 +52,12 @@ type ConsensusSet struct {
 	// current path from disk and true otherwise
 	updatePath bool // DEPRECATED
 
+	// blocksLoaded is the number of blocks that have been loaded
+	// from memory. This variable only exists while some
+	// structures are still in memory, and should always equal
+	// db.pathHeight after loading from disk
+	blocksLoaded types.BlockHeight
```
DavidVorick: also needs a DEPRECATED tag, that makes it easy to grep

modules/consensus/diffs.go
```diff
@@ -288,15 +288,20 @@ func (cs *ConsensusSet) deleteObsoleteDelayedOutputMaps(bn *blockNode, dir modul
...
+				panic(err)
+			}
 		}
+		cs.blocksLoaded += 1
 	} else {
-		cs.currentPath = cs.currentPath[:len(cs.currentPath)-1]
-		if cs.updatePath {
-			cs.db.popPath()
+		err := cs.db.popPath()
+		if err != nil {
```
DavidVorick: `if err != nil && build.DEBUG {`

modules/consensus/diffs.go
```diff
@@ -288,15 +288,20 @@ func (cs *ConsensusSet) deleteObsoleteDelayedOutputMaps(bn *blockNode, dir modul
 func (cs *ConsensusSet) updateCurrentPath(bn *blockNode, dir modules.DiffDirection) {
 	// Update the current path.
 	if dir == modules.DiffApply {
-		cs.currentPath = append(cs.currentPath, bn.block.ID())
 		if cs.updatePath {
-			cs.db.pushPath(bn.block)
+			err := cs.db.pushPath(bn.block.ID())
+
+			if err != nil {
```
DavidVorick: `if err != nil && build.DEBUG {`

triazo: One of the reasons for checking build.DEBUG is to remove the code when the flag isn't true. If you have it as part of a multipart conditional will this still be the case?

modules/consensus/persist.go
```diff
@@ -47,14 +59,20 @@ func (cs *ConsensusSet) load(saveDir string) error {
 		return cs.load(saveDir)
 	}
 
+	// The state cannot be easily reverted to a point where the
+	// consensusSetHash can be re-made. Load from disk instead
+	pb, err := cs.db.getBlockMap(bid)
```
DavidVorick: `pb` is a name that breaks conventions because there's no immediate context to tell you what it stands for. This will not be a problem after renaming `getBlockMap` to `getProcessedBlock`.

So I guess it's okay for this pr

modules/consensus/persist.go
```diff
@@ -62,6 +80,10 @@ func (cs *ConsensusSet) load(saveDir string) error {
 
 		// Blocks loaded from disk are trusted, don't bother with verification.
 		lockID := cs.mu.Lock()
+		// This gaurd is for when the program is stopped. It is temporary.
```
DavidVorick: guard

DavidVorick: tag all temporary code with // DEPRECATED

modules/consensus/synchronize.go
```diff
@@ -43,6 +44,9 @@ func (s *ConsensusSet) receiveBlocks(conn modules.PeerConn) error {
 			// Blocks received during synchronize aren't trusted; activate full
 			// verification.
 			lockID := s.mu.Lock()
+			if !s.db.open {
+				return errors.New("database not open")
```
DavidVorick: is this a problem that might be common?

DavidVorick:
Is there a test that creates a new consensus set, saves a bunch of blocks, then closes the consensus set and opens it again (which means that everything will load from the database) and checks that the consensus set hash is still correct?

modules/consensus/synchronize.go
triazo: It happens when you close the database and there are still threads running. As the closing happens after the API shuts down, I believe there are only two spots where this guard needs to happen: here and in the synchronize loop.

I'm not accounting for random occurrences like the disk being disconnected; that will just panic.

DavidVorick:
A test like that should be added to persist_test.go

Ideally, the test would include every type of transaction (mining blocks, sending coins, creating + revising file contrcts, storage proofs, and siafunds). Block builders for all these types already exist in accept_test.go, but I'm not sure how easy it would be to adapt them to the persist_test.go stuff.

modules/consensus/synchronize.go
DavidVorick: ok, makes sense.

PR #688, opened 2015-07-28
Wallet refactor

DavidVorick:
I was debating, but I think this refactoring is ready for its first PR. More to come :)

This primarily focuses on the persistence, using encrypted files, creating files if they don't exist, initializing the wallet in a first stage (before the encryption key is known), initializing the wallet in a second stage (after the encryption key is known).

The code is only lightly tested, and depends on the changes to the Nebulous fork of the 'ed25519' package. That package is also going to see more changes.


modules/wallet/persist.go
```diff
@@ -1,175 +1,212 @@
...
+
+	PrimarySeed     string // Name of the primary seed for the wallet.
+	AddressProgress uint64 // Number of addresses used in the primary seed.
+}
+
+// saveSettings writes the wallet's settings to the wallet's settings file,
+// replacing the existing file.
+func (w *Wallet) saveSettings() error {
+	return persist.SaveFile(settingsMetadata, &w.settings, filepath.Join(w.persistDir, settingsFile))
+}
```
lukechampine: I think this should just be `w.settings`.

modules/wallet/seed.go
```diff
@@ -0,0 +1,176 @@
...
+	return nil
+}
+
+// integrateSeed takes an address seed as input and from that generates
+// 'publicKeysPerSeed' addresses that the wallet is able to spend.
+func (w *Wallet) integrateSeed(masterKey crypto.TwofishKey, s seed, seedFilename string) error {
+	for i := uint64(0); i < publicKeysPerSeed; i++ {
+		w.generateAndTrackKey(masterKey, s, seedFilename, i)
+	}
+	return nil
```
lukechampine: under what conditions would this return an error?

modules/wallet/seed.go
```diff
@@ -0,0 +1,176 @@
...
+	cryptSeed, err := sek.EncryptBytes(s[:])
+	if err != nil {
+		return err
+	}
+	w.settings.PrimarySeed = filename
+	w.settings.AddressProgress = 0
+	return persist.SaveFile(seedMetadata, &SeedFile{encryptionVerification, cryptSeed}, filename)
+}
+
+// initAuxillarySeeds scans the wallet folder for wallet seeds. Auxiallry seeds
```
lukechampine: typo

persist/persist.go
```diff
@@ -8,19 +11,39 @@ import (
...
+
+		switch build.Release {
+		case "testing":
+			return filepath.Join(build.SiaTestingDir, "home")
+		case "dev":
+			return filepath.Join(home, ".config", "SiaDev")
+		default:
+			return filepath.Join(home, ".config", "Sia")
+		}
+	}()
```
lukechampine: we no longer use the home folder, so this function can be deleted.

persist/persist_test.go
```diff
@@ -0,0 +1,25 @@
...
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/NebulousLabs/Sia/build"
+)
+
+// TestIntegrationRandomSuffix checks that the random suffix creator creates
+// valid files.
+func TestIntegrationRandomSuffix(t *testing.T) {
```
lukechampine: seems like a unit test to me...

DavidVorick: RandomSuffix is an exported function with no state. The definition I've been using is that 'Integration' means no internal state was modified manually and no unexported functions are variables were used.

Perhaps 'Integration' is the wrong word or the definition should be changed, but under that definition this is an integration test.

lukechampine: Yeah that definition is a little odd. To me, a pure (i.e. stateless) function is exactly the sort that you'd want to run unit tests on. My view is:
- Unit test: single function being tested.
- Integration test: multiple functions interacting with each other.

Both types of test may requiring poking around in the internal state. The important thing is that unit tests are testing a single "unit" of code, either an entire function or a well-defined and easily separable piece of code.

Obviously, unit tests are easier than integration tests, which is why it's advantageous to use as many pure functions as possible.

DavidVorick: Hmm. I feel like the useful distinction is not between unit vs. integration, but rather exported-only vs. non-exported.

DavidVorick:
all set

PR #689, opened 2015-07-28
Use database for some of block map

triazo:
Unfortunately, I can't completely get rid of cs.blockMap until I replace blockNode with processedBlock. 

As they are mostly different things, I'm attempting to make them separate pull requests.

This PR also changes the load sequence slightly, by opening the database, then returning to do the genesis block announcement, then loading all the diffs from the database.


DavidVorick:
Ok, thanks for keeping the PRs small :)

PR #691, opened 2015-07-29
Block Node -> Processed Block

triazo:
To my knowledge, the only outstanding problem with this is slow startup causing some issues with the update system when running with debug turned on. (a couple of lock held too long warnings). There's probably more.


modules/consensus/accept.go
```diff
@@ -42,7 +43,7 @@ func (cs *ConsensusSet) validHeader(b types.Block) error {
 
 	// Check that the timestamp is not in 'the past', where the past is defined
 	// by earliestChildTimestamp.
-	if parent.earliestChildTimestamp() > b.Timestamp {
+	if parent.earliestChildTimestamp(cs.db) > b.Timestamp {
```
DavidVorick: probably makes more sense to call `cs.earliestChildTimestamp(parent)`

modules/consensus/accept_bench_test.go
```diff
@@ -35,7 +35,7 @@ func benchmarkEmptyBlocks(b *testing.B) error {
 	}
 	h := cst.cs.db.pathHeight()
 	for i := types.BlockHeight(1); i < h; i++ {
-		err = cs.AcceptBlock(cst.cs.blockMap[cst.cs.db.getPath(i)].block)
+		err = cs.AcceptBlock(cst.cs.db.getBlockMap(cst.cs.db.getPath(i)).Block)
```
DavidVorick: is there an easier way to do this? The nested call seems unnecessary.

modules/consensus/applytransaction_test.go
```diff
@@ -427,7 +427,7 @@ func TestMisuseApplyFileContractRevisions(t *testing.T) {
 	txn := types.Transaction{
 		FileContractRevisions: []types.FileContractRevision{{}},
 	}
-	cst.cs.applyFileContractRevisions(bn, txn)
+	cst.cs.applyFileContractRevisions(pb, txn)
```
DavidVorick: such a huge fraction of this diff set is just 'bn' -> 'pb'.

I was afraid of this PR but it's actually not so bad.

modules/consensus/processed_block.go
```diff
@@ -18,53 +17,42 @@ import (
...
+	current := pb
 	for i := uint64(0); i < types.MedianTimestampWindow; i++ {
-		windowTimes[i] = current.block.Timestamp
+		windowTimes[i] = current.Block.Timestamp
 
 		// If we are at the genesis block, keep using the genesis block for the
 		// remaining times.
-		if current.parent != nil {
-			current = current.parent
+		if current.Parent != types.ZeroID {
```
DavidVorick: :+1:

modules/consensus/processed_block.go
```diff
@@ -115,11 +103,20 @@ func (bn *blockNode) targetAdjustmentBase() *big.Rat {
...
 	// target.
-	timePassed := bn.block.Timestamp - windowStart.block.Timestamp
+	timePassed := pb.Block.Timestamp - windowStart.Block.Timestamp
 	expectedTimePassed := types.BlockFrequency * windowSize
 	return big.NewRat(int64(timePassed), int64(expectedTimePassed))
 }
 
+// setChildTarget computes the target of a blockNode's child. All children of a node
+// have the same target.
+func (pb *processedBlock) setChildTarget(db *setDB) {
```
DavidVorick: probably more intuitive to have ConsensusSet.setChildTarget(pb)

modules/consensus/processed_block.go
```diff
@@ -134,29 +131,23 @@ func clampTargetAdjustment(base *big.Rat) *big.Rat {
...
-	adjustment := clampTargetAdjustment(bn.targetAdjustmentBase())
-	adjustedRatTarget := new(big.Rat).Mul(bn.parent.childTarget.Rat(), adjustment)
-	bn.childTarget = types.RatToTarget(adjustedRatTarget)
-}
-
 // newChild creates a blockNode from a block and adds it to the parent's set of
-// children. The new node is also returned.
-func (bn *blockNode) newChild(b types.Block) *blockNode {
+// children. The new node is also returned. It necessairly modifies the database
+func (pb *processedBlock) newChild(b types.Block, db *setDB) *processedBlock {
```
DavidVorick: same here

DavidVorick:
Ok, I think I've been through the whole thing. Overall looks pretty good, not a whole lot of substance changed, mostly just busywork.

Let me know if there are any code spots you think should get more attention.

modules/consensus/processed_block.go
triazo: I agree, and passing the database here does seem unintuitive here. Unfortunately targetAdjustmentBase currently uses the database to find the timestamp of the block 512 earlier and this seemed like the simplest way to give it that.

triazo:
My biggest worry with these changes (including the last PR) is the loading and handling of the genesis block. This got changed a bit, and while it hasn't broken yet its very possible there is some edge case I haven't thought of. I think most of the changes happened in the previous PR though.

Other than that not much changed here aside from simple bn->pb translations, which I'm not that worried about. I suspect the biggest source of bugs comes from the fact that modifications to a processed block no longer persist unless you manually update it, but this would be difficult to find without a full reading through of the package.

PR #694, opened 2015-08-03
Consistency gaurd

triazo:
Also removes the verification rigor flag, which is currently not being used. (it was always set to full verification)

### DIFF
--- a/modules/consensus/consensusset.go
+++ b/modules/consensus/consensusset.go
@@ -15,38 +15,15 @@ import (
 	"github.com/NebulousLabs/Sia/types"
 )
 
-const (
-	// fullVerification indicates that a block should be fully verified when
-	// being loaded from disk.
-	fullVerification verificationRigor = 0
-
-	// partialVerification indicates that transaction verification can be
-	// skipped. Transaction verification is computationally intensive, and
-	// skipping such a step noticably increases speed when loading many blocks
-	// at once. Usually, partialVerification is used when loading blocks from
-	// disk.
-	partialVerification verificationRigor = 1
-)
-
 var (
 	ErrNilGateway = errors.New("cannot have a nil gateway as input")
 )
-
-// verificationRigor is a type indicating the intensity of verification that
-// should be using while accepting a block. For blocks that come from trusted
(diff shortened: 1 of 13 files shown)


triazo:
Will admit that I laughed at the spelling error. Also changed some functions to be part of the consensusset.

modules/consensus/consensusset.go
```diff
@@ -15,38 +15,15 @@ import (
...
-	// being loaded from disk.
-	fullVerification verificationRigor = 0
-
-	// partialVerification indicates that transaction verification can be
-	// skipped. Transaction verification is computationally intensive, and
-	// skipping such a step noticably increases speed when loading many blocks
-	// at once. Usually, partialVerification is used when loading blocks from
-	// disk.
-	partialVerification verificationRigor = 1
-)
```
DavidVorick: hooray

modules/consensus/persist.go
```diff
@@ -84,7 +97,7 @@ func (cs *ConsensusSet) loadDiffs() {
 		// Blocks loaded from disk are trusted, don't bother with verification.
 		lockID := cs.mu.Lock()
 		// This guard is for when the program is stopped. It is temporary.
-		// DEPRICATED
+		// DEPRECATED
 		if !cs.db.open {
 			break
 		}
```
DavidVorick: what makes this code temporary?

modules/consensus/validtransaction.go
```diff
@@ -257,7 +255,9 @@ func (cs *ConsensusSet) TryTransactionSet(txns []types.Transaction) (modules.Con
 	diffHolder.Height = cs.height()
 	defer func() {
```
DavidVorick: typically you put the defer after the initalize. IE:

```
startGuard()
defer stopGuard()
```

modules/consensus/persist.go
triazo: When every structure is loaded from disk immediately, there will be a single write lock on consensus when loading and the database can't be closed while a write lock is held. We'll probably need this guard in gateway though.

PR #695, opened 2015-08-03
/wallet/status VisibleAddresses gives empty array

Mingling94:
Before it would give nil, as I found from JS errors on the front end

There are at least 10 other instances of `[]` using `ctrl+f` across `API.md`. I'm not sure I'm the best person though to make all the changes for the API endpoints to return empty slices instead of nil.

### DIFF
--- a/modules/wallet/info.go
+++ b/modules/wallet/info.go
@@ -11,8 +11,9 @@ import (
 // Info fills out and returns a WalletInfo struct.
 func (w *Wallet) Info() modules.WalletInfo {
 	wi := modules.WalletInfo{
-		Balance:     w.Balance(false),
-		FullBalance: w.Balance(true),
+		Balance:          w.Balance(false),
+		FullBalance:      w.Balance(true),
+		VisibleAddresses: []types.UnlockHash{},
 	}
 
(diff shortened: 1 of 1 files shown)


lukechampine:
yes, this is an ugly quirk of Go's JSON encoding. The advantage is that it allows you to distinguish missing slices from those that are merely empty. We should be careful that we always send empty slices rather than nil. I'm sure there are other places where it's causing problem too.

modules/wallet/info.go
```diff
@@ -11,8 +11,9 @@ import (
 // Info fills out and returns a WalletInfo struct.
 func (w *Wallet) Info() modules.WalletInfo {
 	wi := modules.WalletInfo{
-		Balance:     w.Balance(false),
-		FullBalance: w.Balance(true),
+		Balance:          w.Balance(false),
+		FullBalance:      w.Balance(true),
+		VisibleAddresses: []types.UnlockHash{},
```
lukechampine: perhaps better would be `make([]types.UnlockHash, w.visibleAddresses)` (and then assign instead of append below), but this is fine for now.

PR #696, opened 2015-08-05
Benchmarks

triazo:
Add benchmarks for:

Adding blocks to Database
Retrieving blocks from database
Verifying a large transaction 
Accepting a block with a large transaction

### DIFF
--- a/modules/consensus/accept_bench_test.go
+++ b/modules/consensus/accept_bench_test.go
@@ -14,7 +14,7 @@ import (
 
 // benchmarkEmptyBlocks is a benchmark that mines many blocks, and
 // measures how long it takes to add them to the consensusset
-func benchmarkEmptyBlocks(b *testing.B) error {
+func benchmarkAcceptEmptyBlocks(b *testing.B) error {
 	// Create an alternate testing consensus set, which does not
 	// have any subscribers
 	testdir := build.TempDir(modules.ConsensusDir, "BenchmarkEmptyBlocksB")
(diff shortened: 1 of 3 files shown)


types/validtransaction_bench_test.go
```diff
@@ -0,0 +1,59 @@
...
+	"testing"
+
+	"github.com/NebulousLabs/Sia/crypto"
+	"github.com/NebulousLabs/Sia/encoding"
+)
+
+// BenchmarkStandaloneValid times how long it takes to verify a single
+// large transaction, with 100 signatures
+func BenchmarkStandaloneValid(b *testing.B) {
+	// make a transaction 10 with valid inputs with valid signatures
```
DavidVorick: docstrings should probably not have exact numbers in them.

types/validtransaction_bench_test.go
```diff
@@ -0,0 +1,59 @@
...
+	b.ReportAllocs()
+	txn := Transaction{}
+	numSigs := 100
```
DavidVorick: I don't think that 100 signatures is a super interesting number. The majority of transactions are going to have somewhere between 1 and 10 signatures, so the benchmark should be in that range.

The other interesting number is the really large transactions, because people can create those to be malicious. But I don't think there's a standard use case for 100 signature transactions - it'll neither apply to an attack case (b/c more signatures is more powerful) nor a standard case.

types/validtransaction_bench_test.go
```diff
@@ -0,0 +1,59 @@
...
+	sk := make([]crypto.SecretKey, numSigs)
+	pk := make([]crypto.PublicKey, numSigs)
```
DavidVorick: nifty trick

pk = sk[32:]

types/validtransaction_bench_test.go
triazo: 100 sigs is already more than what's allowed by transaction pool. But I agree that having 5 or so signatures would be more interesting to see normal usage.

types/validtransaction_bench_test.go
```diff
@@ -0,0 +1,59 @@
...
+		copy(txn.TransactionSignatures[i].ParentID[:], encoding.Marshal(i))
+	}
+	for i := 0; i < numSigs; i++ {
+		sigHash := txn.SigHash(i)
+		sig0, err := crypto.SignHash(sigHash, sk[i])
+		if err != nil {
+			b.Fatal(err)
+		}
+		txn.TransactionSignatures[i].Signature = sig0[:]
+	}
```
DavidVorick: It looks like all 3 of these loops can be combined into a single loop, and then you can ditch the external variables.

triazo: You could probably combine the first two loops, but the third one signs the transaction, which requires all of the inputs to be in the transaction already.

DavidVorick: ok

DavidVorick: you should still merge the top two loops

modules/consensus/accept_bench_test.go
```diff
@@ -64,10 +64,109 @@ func benchmarkEmptyBlocks(b *testing.B) error {
...
+	}
+	h := cst.cs.db.pathHeight()
+	for i := types.BlockHeight(1); i < h; i++ {
+		err = cs.AcceptBlock(cst.cs.db.getBlockMap(cst.cs.db.getPath(i)).Block)
+		if err != nil {
+			b.Fatal(err)
+		}
+	}
+
+	// construct a transaction using 100 utxo's, and signed 100 times
```
DavidVorick: comment should not have explicit numbers

PR #697, opened 2015-08-06
Move Siafund outputs to disk

triazo:
Mostly translation, some new code in setdb.go for handling bucket size and iterating over an entire bucket.

### DIFF
--- a/modules/consensus/accept_test.go
+++ b/modules/consensus/accept_test.go
@@ -321,13 +321,13 @@ func TestInconsistentCheck(t *testing.T) {
 	// Corrupt the consensus set.
 	var sfod types.SiafundOutputID
 	var sfo types.SiafundOutput
-	for id, output := range cst.cs.siafundOutputs {
+	cst.cs.db.forEachSiafundOutputs(func(id types.SiafundOutputID, output types.SiafundOutput) {
 		sfod = id
 		sfo = output
-		break
-	}
+	})
 	sfo.Value = sfo.Value.Add(types.NewCurrency64(1))
-	cst.cs.siafundOutputs[sfod] = sfo
+	cst.cs.db.rmSiafundOutputs(sfod)
+	cst.cs.db.addSiafundOutputs(sfod, sfo)
 
 	// Mine and submit a block, triggering the inconsistency check.
 	defer func() {
@@ -730,13 +730,12 @@ func (cst *consensusSetTester) testSpendSiafundsBlock() error {
 	var srcID types.SiafundOutputID
 	var srcValue types.Currency
 	anyoneSpends := types.UnlockConditions{}.UnlockHash()
-	for id, sfo := range cst.cs.siafundOutputs {
+	cst.cs.db.forEachSiafundOutputs(func(id types.SiafundOutputID, sfo types.SiafundOutput) {
 		if sfo.UnlockHash == anyoneSpends {
 			srcID = id
 			srcValue = sfo.Value
-			break
 		}
-	}
+	})
 
 	// Create a transaction that spends siafunds.
 	txn := types.Transaction{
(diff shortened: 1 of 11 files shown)


modules/consensus/accept_test.go
```diff
@@ -321,13 +321,13 @@ func TestInconsistentCheck(t *testing.T) {
 	// Corrupt the consensus set.
 	var sfod types.SiafundOutputID
 	var sfo types.SiafundOutput
-	for id, output := range cst.cs.siafundOutputs {
+	cst.cs.db.forEachSiafundOutputs(func(id types.SiafundOutputID, output types.SiafundOutput) {
 		sfod = id
 		sfo = output
-		break
-	}
+	})
```
DavidVorick: does this forEach work?

You are passing a function to the database that sets sfod to id and sfo to output, but then it looks like the values are only changes once?

I guess I don't understand the construction here.

modules/consensus/consistency.go
```diff
@@ -112,9 +112,9 @@ func (cs *ConsensusSet) checkSiacoins() error {
 // checkSiafunds counts the siafund outputs and checks that there are 10,000.
 func (cs *ConsensusSet) checkSiafunds() error {
 	totalSiafunds := types.ZeroCurrency
-	for _, sfo := range cs.siafundOutputs {
+	cs.db.forEachSiafundOutputs(func(sfoid types.SiafundOutputID, sfo types.SiafundOutput) {
 		totalSiafunds = totalSiafunds.Add(sfo.Value)
-	}
```
DavidVorick: this forEach on the other hand makes more sense. You are adding to totalSiafunds repeatedly.

modules/consensus/accept_test.go
triazo: It does behave differently. Basically instead of setting sfod and sfo to the first element, it sets them to the last after iterating over the entire set. As this construction only occurs in tests and my current setdb implementation has no mechanism for escaping the loop part way through it seemed better not to dwell on it too much.

modules/consensus/persist.go
```diff
@@ -25,7 +25,15 @@ func (cs *ConsensusSet) initSetDB() error {
...
 		return err
 	}
-	// Explicit initilization preferred to implicit
+
+	// Update the siafundoutput diffs map for the genesis block
+	// on disk
+	cs.updateDatabase = true
+	for _, sfod := range cs.blockRoot.SiafundOutputDiffs {
+		cs.commitSiafundOutputDiff(sfod, modules.DiffApply)
+	}
```
DavidVorick: I am confused about what is happening here. Why are you committing to the diff set in `persist.go`?

modules/consensus/setdb.go
```diff
@@ -180,6 +182,9 @@ func (db *setDB) rmItem(bucket string, key interface{}) error {
 	if build.DEBUG && !db.checkConsistencyGuard() && build.Release != "testing" {
 		panic(errNotGuarded)
 	}
+	if build.DEBUG && key == nil {
+		panic(errNilInterface)
+	}
```
DavidVorick: I don't think that this check is necessary. Won't it panic anyway if you pass in a nil interface?

modules/consensus/setdb.go
```diff
@@ -200,6 +205,9 @@ func (db *setDB) rmItem(bucket string, key interface{}) error {
 
 // inBucket checks if an item with the given key is in the bucket
 func (db *setDB) inBucket(bucket string, key interface{}) bool {
+	if key == nil {
+		return false
+	}
```
DavidVorick: Along similar lines, is this code ever supposed to be reached? If not, it should probably just be a `if DEBUG { panic() }`

modules/consensus/setdb.go
```diff
@@ -207,6 +215,30 @@ func (db *setDB) inBucket(bucket string, key interface{}) bool {
...
+	s, err := db.BucketSize(bucket)
+	if build.DEBUG && err != nil {
+		panic(err)
+	}
+	return s
+}
+
+// forEachInBucket runs a given function on every element in a given
+// bucket name, and will panic on any error
+func (db *setDB) forEachItem(bucket string, fn func(k, v []byte) error) {
```
DavidVorick: function name and docstring don't match

modules/consensus/persist.go
triazo: This is init code, that needs to happen after the genesis block is added to the consensus set and before the consensus set hash for it is computed. The function that this is in is supposed to be a first time database initialization. This loop was in consensusset.go before.

modules/consensus/validtransaction_test.go
```diff
@@ -468,10 +468,10 @@ func TestValidSiafunds(t *testing.T) {
 
 	// Create a transaction with invalid unlock conditions.
 	var sfoid types.SiafundOutputID
-	for mapSfoid, _ := range cst.cs.siafundOutputs {
+	cst.cs.db.forEachSiafundOutputs(func(mapSfoid types.SiafundOutputID, sfo types.SiafundOutput) {
 		sfoid = mapSfoid
-		break
-	}
+		// pointless to do this but I can't think of a better way.
+	})
```
DavidVorick: lol you just need a way to grab a single element from the bucket.

modules/consensus/accept_test.go
DavidVorick: All you need to do is pull a single element out of the loop lol.

Perhaps there's a better way to do that than forEach?

triazo: I'm considering some sort of cursor based thing. What I'm apprehensive about are the couple of spots where you need an arbitrary element that fits certain conditions. An implementation of something like firstSiacoinOutputs(func conditions(k, v) bool) would be just as complicated as the for each loop. Getting a single arbitrary value won't require passing in a function and should be much simpler. But even this only has a couple use cases inside of tests.

DavidVorick: I don't think it's used very much overall

modules/consensus/setdb.go
```diff
@@ -207,6 +208,30 @@ func (db *setDB) inBucket(bucket string, key interface{}) bool {
...
+
+// forEachItem runs a given function on every element in a given
+// bucket name, and will panic on any error
+func (db *setDB) forEachItem(bucket string, fn func(k, v []byte) error) {
+	err := db.View(func(tx *bolt.Tx) error {
+		b := tx.Bucket([]byte(bucket))
+		if build.DEBUG && b == nil {
+			panic(errNilBucket)
+		}
+		return b.ForEach(fn)
```
lukechampine: won't you panic anyway if you try to call `ForEach` on `nil`?

PR #698, opened 2015-08-07
Wallet Refactor

DavidVorick:
The wallet interface should now be compatible with exchanges. There has been a good amount of testing, especially considering that the TestComplexForking from the consensus set actually also makes sure that the forking passes all of the sanity checks in the wallet as well.

One unfortunate change is that creating a wallet now takes up to a full second (depending on the machine). I'm not sure where the bottleneck is. PublicKeysPerSeed has been set to 100 for now. I'm guess that 100 is not enough, and that 250+ will be necessary. I would like to be able to go much higher than that, but that might start to get resource intensive. The biggest concern is for renters and hosts, who will be making frequent new addresses. It's possible that they will be making their own seeds.

Compatibility with the old wallet files has not been added yet. Compatibility with siag files has not been added yet. Documentation is pretty sparse.

I promise I did not do all of this on August 7th. Github is confused.


api/server_test.go
```diff
@@ -34,15 +35,16 @@ var (
...
-	cs      *consensus.ConsensusSet
-	gateway modules.Gateway
-	host    modules.Host
-	hostdb  modules.HostDB
-	miner   modules.Miner
-	renter  modules.Renter
-	tpool   modules.TransactionPool
-	exp     modules.Explorer
-	wallet  modules.Wallet
+	cs        *consensus.ConsensusSet
```
lukechampine: can this be a `modules.ConsensusSet` now?

modules/explorer/addblock.go
```diff
@@ -129,14 +129,14 @@ func (tx *boltTx) addNewHash(bucketName string, t int, hash crypto.Hash, value i
...
+func (tx *boltTx) addNewOutput(outputID types.SiacoinOutputID, txid types.TransactionID) {
+	otx := outputTransactions{txid, types.TransactionID{}}
 	tx.addNewHash("SiacoinOutputs", hashCoinOutputID, crypto.Hash(outputID), otx)
 }
 
 // addNewSFOutput does the same thing as addNewOutput does, except for siafunds
-func (tx *boltTx) addNewSFOutput(outputID types.SiafundOutputID, txid crypto.Hash) {
-	otx := outputTransactions{txid, crypto.Hash{}}
+func (tx *boltTx) addNewSFOutput(outputID types.SiafundOutputID, txid types.TransactionID) {
+	otx := outputTransactions{txid, types.TransactionID{}}
```
lukechampine: :+1: woo type information

modules/explorer/explorer_test.go
```diff
@@ -17,11 +18,12 @@ import (
 // Explorer tester struct is the helper object for explorer
 // testing. It holds the helper modules for its testing
 type explorerTester struct {
-	cs      *consensus.ConsensusSet
-	gateway modules.Gateway
-	miner   modules.Miner
-	tpool   modules.TransactionPool
-	wallet  modules.Wallet
+	cs        *consensus.ConsensusSet
```
lukechampine: ditto

modules/host/host_test.go
```diff
@@ -18,11 +19,12 @@ import (
 // A hostTester is the helper object for host testing, including helper modules
 // and methods for controlling synchronization.
 type hostTester struct {
-	cs      *consensus.ConsensusSet
-	gateway modules.Gateway
-	miner   modules.Miner
-	tpool   modules.TransactionPool
-	wallet  modules.Wallet
+	cs        *consensus.ConsensusSet
```
lukechampine: ditto

modules/host/negotiate.go
```diff
@@ -75,7 +75,7 @@ func (h *Host) considerTerms(terms modules.ContractTerms) error {
 	case len(terms.MissedProofOutputs) != 1:
 		return errors.New("refund len does not match host settings")
 
-	case terms.MissedProofOutputs[0].UnlockHash != types.ZeroUnlockHash:
+	case terms.MissedProofOutputs[0].UnlockHash != types.UnlockHash{}:
```
lukechampine: why the change? Explicitly naming ZeroUnlockHash makes the code more readable (and prevents the rare, but annoying syntax of `if x != (types.UnlockHash{})` (note parens)

modules/renter/renter_test.go
```diff
@@ -17,11 +18,12 @@ import (
 
 // renterTester contains all of the modules that are used while testing the renter.
 type renterTester struct {
-	cs     *consensus.ConsensusSet
-	hostdb modules.HostDB
-	miner  modules.Miner
-	tpool  modules.TransactionPool
-	wallet modules.Wallet
+	cs        *consensus.ConsensusSet
```
lukechampine: ditto

api/server_test.go
DavidVorick: ah yeah. I think there are still a few of those floating around, but it's not necessary anymore outside of the consensus package.

modules/transactionpool/transactionpool_test.go
```diff
@@ -16,11 +18,12 @@ import (
 // A tpoolTester is used during testing to initialize a transaction pool and
 // useful helper modules.
 type tpoolTester struct {
-	cs      *consensus.ConsensusSet
-	gateway modules.Gateway
-	tpool   *TransactionPool
-	miner   modules.Miner
-	wallet  modules.Wallet
+	cs        *consensus.ConsensusSet
```
lukechampine: ditto

modules/wallet.go
```diff
@@ -3,149 +3,257 @@ package modules
...
-type WalletInfo struct {
-	Balance          types.Currency
-	FullBalance      types.Currency
-	VisibleAddresses []types.UnlockHash
-	NumAddresses     int
-}
+type (
+	// AddressSeed is cryptographic entropy that is used to derive spendable
+	// wallet addresses.
+	Seed [crypto.EntropySize]byte
```
lukechampine: wrong name in the docstring

modules/wallet.go
```diff
@@ -3,149 +3,257 @@ package modules
...
-	// that were added by calling 'FundSiacoins' or 'FundSiafunds'.
-	AddTransactionSignature(types.TransactionSignature) uint64
-
-	// Sign will sign any inputs added by 'FundSiacoins' or 'FundSiafunds' and
-	// return a transaction set that contains all parents prepended to the
-	// transaction. If more fields need to be added, a new transaction builder
-	// will need to be created.
+		FundType       types.Specifier
+		OutputID       types.OutputID
+		RelatedAddress types.UnlockHash
```
lukechampine: what is meant by "Related?" I already know that the address is related if it's bundled in the same struct as the other info.

modules/wallet/history.go
```diff
@@ -0,0 +1,85 @@
...
+		i++
+	}
+	j := i
+	for j < len(w.walletTransactions) {
+		if w.walletTransactions[j].ConfirmationHeight > endBlock {
+			break
+		}
+		j++
+	}
+	return w.walletTransactions[i:j], nil
```
lukechampine: perhaps clearer:
```go
var start, end int
for i, txn := range w.walletTransactions {
    if txn.ConfirmationHeight >= startBlock {
        start = i
    }
    if txn.ConfirmationHeight > endBlock {
        end = i
        break
    }
}
return w.walletTransactions[start:end], nil
```
there may be another problem here: if `start == end`, then `[start:end] == nil`. I'm not sure if this is the desired/expected behavior.

modules/wallet/money.go
```diff
@@ -0,0 +1,46 @@
...
+
+// ConfirmedBalance returns the balance of the wallet according to all of the
+// confirmed transactions.
+func (w *Wallet) ConfirmedBalance() (siacoinBalance types.Currency, siafundBalance types.Currency, siafundClaimBalance types.Currency) {
+	lockID := w.mu.Lock()
+	defer w.mu.Unlock(lockID)
+
+	for _, sco := range w.siacoinOutputs {
+		siacoinBalance = siacoinBalance.Add(sco.Value)
+	}
```
lukechampine: it occurs to me now that we could actually write a sum function. It would look like:
```go
total := types.CurrencySum(w.siacoinOutputs, "Value")
```
You would need to use reflection though, so it'd be error-prone and slow. Fun to think about though.

modules/wallet/persist.go
```diff
@@ -37,8 +38,33 @@ type WalletSettings struct {
...
+	// to be generated.
+	PrimarySeedFile     SeedFile
+	PrimarySeedProgress uint64
+	PrimarySeedFilename string
+}
+
+// unlockKey creates a wallet unlocking key from the input master key.
+func unlockKey(masterKey crypto.TwofishKey) crypto.TwofishKey {
+	keyBase := append(masterKey[:], unlockModifier[:]...)
+	return crypto.TwofishKey(crypto.HashObject(keyBase))
```
lukechampine: `crypto.HashAll(masterKey, unlockModifier)` may be preferable

modules/wallet.go
DavidVorick: Hmm. I see what you are saying but just saying 'Address' really doesn't seem sufficient to me.

Maybe there's another name that could be used?

modules/wallet/history.go
DavidVorick: every time that txn.Confirmation height >= startBlock, it's going to be set again. You'd need something like `if txn.ConfirmationHeight >= startBlock && start == -1`, and then you'd need to set `end = i+1`

modules/wallet/persist.go
```diff
@@ -37,8 +38,33 @@ type WalletSettings struct {
...
+
+// checkMasterKey verifies that the master key is correct.
+func (w *Wallet) checkMasterKey(masterKey crypto.TwofishKey) error {
+	unlockKey := unlockKey(masterKey)
+	verification, err := unlockKey.DecryptBytes(w.settings.EncryptionVerification)
+	if err != nil {
+		return err
+	}
+	expected := make([]byte, encryptionVerificationLen)
+	if !bytes.Equal(expected, verification) {
```
lukechampine: Comparing to zero confused me for a bit. I thought it might be an error. Perhaps a comment would be helpful here. Better would be a `bytes.IsEmpty` function, but it doesn't exist.

modules/wallet/seed.go
```diff
@@ -39,121 +37,158 @@ type (
...
+	}
+	copy(seed[:], plainSeed)
+	return seed, nil
+}
+
+// generateAndTrackKey tracks a key of a given index from a given seed.
+func (w *Wallet) generateAndTrackKey(seed modules.Seed, index uint64) {
+	// Generate the key and check it is new to the wallet.
+	spendableKey := generateSpendableKey(seed, index)
+	w.keys[spendableKey.unlockConditions.UnlockHash()] = spendableKey
```
lukechampine: I would probably inline this function unless you think it's going to be called in places other than `integrateSeed`. As it stands, the meaning of "index" is a big confusing.

modules/wallet/seed.go
```diff
@@ -39,121 +37,158 @@ type (
...
+		return err
+	}
+	w.integrateSeed(seed)
+	return nil
+}
+
+// recoverSeed integrates a recovery seed into the wallet.
+func (w *Wallet) recoverSeed(masterKey crypto.TwofishKey, seed modules.Seed) error {
+	// Integrate the seed with the wallet.
+	seedFilename := filepath.Join(w.persistDir, seedFilePrefix+persist.RandomSuffix()+seedFileSuffix)
```
lukechampine: why is this variable defined so far away from where it is used?

modules/wallet/update.go
```diff
@@ -1,107 +1,243 @@
...
+		// Sanity check - the last element of the wallet transaction array
+		// should be the item we are deleteing.
+		lastIndex := len(w.walletTransactions) - 1
+		if build.DEBUG && wtid != modules.CalculateWalletTransactionID(w.walletTransactions[lastIndex].TransactionID, w.walletTransactions[lastIndex].OutputID) {
+			panic("wallet transactions are being deleted in the wrong order")
+		}
+		w.walletTransactions = w.walletTransactions[:lastIndex]
 		return
 	}
+	return
```
lukechampine: the `return` inside the block is redundant

modules/wallet/update.go
```diff
@@ -1,107 +1,243 @@
...
+			RelatedAddress: uh,
+			Value:          value,
 		}
-		key.outputs[scod.ID].spendable = false
+		w.walletTransactions = append(w.walletTransactions, wt)
+		w.walletTransactionMap[wtid] = &w.walletTransactions[len(w.walletTransactions)-1]
+		w.historicOutputs[oid] = value
+		return
 	}
+	return
```
lukechampine: ditto

modules/wallet/update.go
```diff
@@ -1,107 +1,243 @@
...
-		w.applyDiff(scod, modules.DiffApply)
+	// There are two different situations under which a subscribee calls
+	// ProcessConsensusChange. The first is when w.subscribed is set to false
+	// AND the mutex is already locked. The other situation is that subscribed
+	// is set to true and is not going to be changed. Therefore there is no
+	// race condition here. If w.subscribed is set to false, trying to grab the
+	// lock would cause a deadlock.
+	if w.subscribed {
+		lockID := w.mu.Lock()
+		defer w.mu.Unlock(lockID)
```
lukechampine: yikes, that's a bit janky. But I trust that you wouldn't have done it unless you had to.

modules/wallet/wallet.go
```diff
@@ -4,44 +4,51 @@ import (
...
+	siafundPool        types.Currency
+
+	keys            map[types.UnlockHash]spendableKey
+	siacoinOutputs  map[types.SiacoinOutputID]types.SiacoinOutput
+	siafundOutputs  map[types.SiafundOutputID]types.SiafundOutput
+	historicOutputs map[types.OutputID]types.Currency
+	spentOutputs    map[types.OutputID]types.BlockHeight
+
+	walletTransactions            []modules.WalletTransaction
+	walletTransactionMap          map[modules.WalletTransactionID]*modules.WalletTransaction
```
lukechampine: why have both?

modules/wallet/wallet.go
```diff
@@ -122,3 +143,34 @@ func (w *Wallet) SendCoins(amount types.Currency, dest types.UnlockHash) ([]type
...
+	if err != nil {
+		return nil, err
+	}
+	err = txnBuilder.FundSiafunds(amount)
+	if err != nil {
+		return nil, err
+	}
+	txnBuilder.AddMinerFee(tpoolFee)
+	txnBuilder.AddSiafundOutput(output)
+	txnSet, err := txnBuilder.Sign(true)
```
lukechampine: you know...you could give the txn builder a sticky error, like the block explorer:
```go
txnBuilder := w.StartTransaction()
txnBuilder.FundSiacoins(tpoolFee)
txnBuilder.FundSiafunds(amount)
txnBuilder.AddMinerFee(tpoolFee)
txnBuilder.AddSiafundOutput(output)
txnSet, err := txnBuilder.Sign(true) // returns the error of the first operation that failed
```
put it on the "maybe later" list :P

siad/main.go
```diff
@@ -35,13 +31,6 @@ type Config struct {
 	}
 }
 
-// init creates the channel that's used to communicate with the testing
-// framework.
-func init() {
-	// Initialize the started channel, only used for testing.
-	started = make(chan struct{}, 1)
-}
```
lukechampine: bleh, good riddance

modules/wallet.go
lukechampine: what does it actually mean?

modules/wallet/history.go
lukechampine: oh, good catch. So I guess I'd keep it mostly the same, but move more code into the `for` expression:
```go
var start, end int
for start = 0; start < len(w.walletTransactions); start++ {
    if w.walletTransactions[start] >= startBlock {
        break
    }
}
for end = start; end < len(w.walletTransactions); end++ {
    if w.walletTransactions[end] > endBlock {
        break
    }
}
```

lukechampine:
okay, this mostly looks good. My comments are very minor. It seems like this PR leaves things in a semi-broken state, but they're semi-broken already and I assume you will address the brokenness in short order.

modules/wallet/seed.go
DavidVorick: probably an artifact of me rewriting this function half a dozen times

modules/wallet/update.go
DavidVorick: Yeah there really wasn't anything else I could think of. The wallet needed a lock during the call to 'Unlock', but the call to 'Unlock' triggers a consensus subscription, which calls 'ProcessConsensusChange', which also needs a lock.

This wouldn't have been necessary with the old, threaded model of subscriptions.

modules/wallet/wallet.go
DavidVorick: yeah. Other than a linear lookup, there's no way to get from an ID to a specific wallet transaction.

Linear lookup may be okay for the average user, but power users, such as hosts, will probably want constant time lookups.

siad/main.go
DavidVorick: it might come back

DavidVorick:
I intended to profile the wallet pretty heavily this week, I imagine I can cut down startup time to 20% or 30% of what it currently is. Wanted to make this PR before everything was finished because... it's already huge. And plenty more to come, but hopefully in much smaller pieces.

modules/host/negotiate.go
DavidVorick: honestly it felt clunky to me, especially since you actually lose characters. I also do not feel that readability was lost, it seems more clear to me that UnlockHash{} is an empty structure than `ZeroUnlockHash` (not that either is particularly opaque).

Mostly it just felt like an unnecessary alias, and I think you should try to eliminate everything in your codebase that doesn't directly contribute a lot of value.

modules/wallet.go
DavidVorick: It's the address that's relevant to the motion of the funds.

So if it's a 'SiacoinInput', then it's the address that the original output was sent to (mostly a useless thing to know). If it's a 'SiacoinOutput', then it's the address that the siacoin output was sent to.

modules/wallet/seed.go
DavidVorick: When I originally wrote these two functions, they had a lot more code in them.

modules/wallet/wallet.go
DavidVorick: A few problems:

1. harder to tell where the error occured
2. multiple functions (Sign and View) need to express the error.

But it would make the above code a lot cleaner.

DavidVorick:
```
=== RUN TestRandomInboundPeer-2
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x30 pc=0x693f32]
goroutine 180 [running]:
github.com/inconshreveable/muxado/proto/frame.(*BasicTransport).Write(0xc208086700, 0xc2081c3c08, 0x10, 0x10, 0xc240000000, 0x0, 0x0)
	<autogenerated>:141 +0xa2
github.com/inconshreveable/muxado/proto/frame.(*WStreamSyn).writeTo(0xc2081c3bf0, 0x2b0e47326690, 0xc208086700, 0x0, 0x0)
	/home/travis/gopath/src/github.com/inconshreveable/muxado/proto/frame/syn.go:75 +0xd4
github.com/inconshreveable/muxado/proto/frame.(*BasicTransport).WriteFrame(0xc208086700, 0x2b0e47326668, 0xc2081c3bf0, 0x0, 0x0)
	/home/travis/gopath/src/github.com/inconshreveable/muxado/proto/frame/transport.go:27 +0x7f
github.com/inconshreveable/muxado/proto.(*Session).OpenStream(0xc2080668c0, 0xfffffeff7fffffff, 0xc208076500, 0x0, 0x0, 0x0, 0x0)
	/home/travis/gopath/src/github.com/inconshreveable/muxado/proto/session.go:147 +0x410
github.com/inconshreveable/muxado/proto/ext.(*Heartbeat).request(0xc2081c3b90)
	/home/travis/gopath/src/github.com/inconshreveable/muxado/proto/ext/heartbeat.go:99 +0xf4
created by github.com/inconshreveable/muxado/proto/ext.(*Heartbeat).Start
	/home/travis/gopath/src/github.com/inconshreveable/muxado/proto/ext/heartbeat.go:47 +0xcc
goroutine 1 [runnable]:
testing.RunTests(0xa129b8, 0xbfa580, 0x14, 0x14, 0xc14601)
	/usr/local/go/src/testing/testing.go:556 +0xdc3
testing.(*M).Run(0xc2080600f0, 0xc145c0)
	/usr/local/go/src/testing/testing.go:485 +0xe1
main.main()
	github.com/NebulousLabs/Sia/modules/gateway/_test/_testmain.go:90 +0x28d
goroutine 36 [sleep]:
github.com/NebulousLabs/Sia/sync.(*RWMutex).threadedDeadlockFinder(0xc2080d2a40)
	/home/travis/gopath/src/github.com/NebulousLabs/Sia/sync/lock.go:78 +0x6a3
created by github.com/NebulousLabs/Sia/sync.New
	/home/travis/gopath/src/github.com/NebulousLabs/Sia/sync/lock.go:47 +0x10d
```

this needs a fix

modules/wallet/wallet.go
lukechampine: why keep the slice then? Could you get by with just the map?

lukechampine:
I just wish I could find a way to reproduce it. Perhaps we'll be able to trigger it via fuzzing, but I don't know...it depends on how the fuzzing is performed. If we're writing to the bare connection, then the fuzzer will need to implement at least part of the muxado protocol in order to find any meaningful bugs. But if the fuzzer is writing to the layer above (as we do now), it may not trigger anything.

modules/wallet/wallet.go
DavidVorick: The slice is in chronological order.

Almost did a linked list instead of a slice.

PR #700, opened 2015-08-09
write a fuzzer for storage proof verification

DavidVorick:
Ran the fuzzer for about 8 hours. 45 corpuses, 0 crashes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,6 @@ all: install
 # dependencies installs all of the dependencies that are required for building
 # Sia.
 dependencies:
-	go install -race std
 	go get -u github.com/NebulousLabs/ed25519
 	go get -u github.com/NebulousLabs/go-upnp
 	go get -u github.com/NebulousLabs/merkletree
@@ -14,10 +13,14 @@ dependencies:
 	go get -u github.com/inconshreveable/go-update
 	go get -u github.com/inconshreveable/muxado
 	go get -u github.com/kardianos/osext
-	go get -u github.com/laher/goxc
 	go get -u github.com/spf13/cobra
 	go get -u github.com/stretchr/graceful
 	go get -u golang.org/x/crypto/twofish
+dev-dependencies: dependencies
+	go install -race std
+	go get -u github.com/dvyukov/go-fuzz/go-fuzz
+	go get -u github.com/dvyukov/go-fuzz/go-fuzz-build
+	go get -u github.com/laher/goxc
 	go get -u golang.org/x/tools/cmd/cover
 
 # fmt calls go fmt on all packages.
@@ -56,7 +59,7 @@ xc-siag: dependencies test test-long REBUILD
 # clean removes all directories that get automatically created during
 # development.
 clean:
-	rm -rf release doc/whitepaper.aux doc/whitepaper.log doc/whitepaper.pdf
+	rm -rf release doc/whitepaper.aux doc/whitepaper.log doc/whitepaper.pdf *Fuzz*
 
 # 3 commands and a variable are available for testing Sia packages. 'pkgs'
 # indicates which packages should be tested, and defaults to all the packages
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -18,6 +17,10 @@ dependencies:
 	go get -u github.com/spf13/cobra
 	go get -u github.com/stretchr/graceful
 	go get -u golang.org/x/crypto/twofish
+dev-dependencies: dependencies
+	go install -race std
+	go get -u github.com/dvyukov/go-fuzz/go-fuzz
+	go get -u github.com/dvyukov/go-fuzz/go-fuzz-build
```
lukechampine: goxc is also a dev dependency.

lukechampine:
It's a bit silly to fuzz `BuildReaderProof`, since it does very little. The function is short enough that you can manually verify its correctness: a panic is only possible when `len(proofSet) == 0` (which should never happen unless the merkletree package is seriously broken; if you're still paranoid, you can add a zero-check that returns an error). All the real work is being done in the merkletree package, which you already fuzzed.

As always, I am irked by the addition of yet more dependencies and yet more Makefile targets. The behavior of `fuzz` is not even consistent with `cover` (using a loop). If you look at all the packages listed in the [go-fuzz trophy list](https://github.com/dvyukov/go-fuzz/#trophies), how many of those have added go-fuzz as a dependency and permanently cemented it in their codebase?

The large majority of Sia is not amenable to fuzzing. The best candidates that come to my mind are the crypto and encoding packages. And the nice thing about those is that they are very stable. Once they have been fuzzed, we know that we won't need to fuzz them again for a long time, because they change so infrequently.

DavidVorick:
Did you rebase master recently? I thought I put this branch directly on top of master?

I felt that it was necessary to fuzz BuildReaderProof because I got another 'invalid storage proof' error during testing. It was not clear to me that the padding was being done correctly, and I want to apply some fuzzing to make sure.

It seems weird that you would be opposed to fuzzing added as a standard tool of the library. The fuzzing commands are very long and I'm expecting to add quite a few fuzzing targets. Here is a list of things I would like to fuzz:

+ signatures
+ hostdb selection algorithm
+ encoding package
+ accept block
+ accept transaction (especially making sure that all possible transaction pool configurations will lead to valid blocks)
+ gateway rpcs
+ adding peers
+ the host using 'buildReaderProof' after a contract is ready for proving

Basically any place that takes foreign input, and any place where I'm not certain about off-by-one errors, etc.

To me, since fuzzing is a part of the development process, it belongs in the makefile. It's also good to keep the fuzzing files around (with the 'gofuzz' build tag)  so that when reviewing later you know what the fuzz function looked like. Especially something as sensitive as crypto, you want to be confident about the type of fuzzing that happened, or at least able to reference fuzzing you did in the past.

And I put them in the makefile for convenience, but also because we use build tags when we fuzz. It's a lot of typing:

`'go-fuzz-build -tags='debug gofuzz' -func=FuzzX github.com/NebulousLabs/Sia/crypto -o crypto-FuzzX && go-fuzz -bin=crypto-FuzzX.zip -workdir=crypto`

In short, I feel like it makes sense to keep it in our makefile.

DavidVorick:
I haven't done this yet, but sometimes you want to make your own corpuses for the fuzzer. It's possible that those would be good to keep in the repo, depending on their nature.

Makefile
DavidVorick: github didn't detect this as an outdated diff, but if you check, you'll see that I fixed the issue.

lukechampine:
You make a good point about keeping `Fuzz` functions around for later inspection. I guess my main beef is with the Makefile. Using variables for `fuzzfunc` and `fuzzpkg` is ugly and confusing. I don't think the fuzz commands belong there.

lukechampine:
is there a way to run `go-fuzz-build` with additional build flags? We probably want to enable `debug` if possible (and maybe `testing`?)

DavidVorick:
`'go-fuzz-build -tags='debug gofuzz'`

Or did you mean something else?

lukechampine:
`go-fuzz-build` doesn't take a `tags` argument on my machine.

DavidVorick:
:'(

PR #701, opened 2015-08-09
Move siacoin outputs map to disk

triazo:
Mostly straightforward changes, had no hiccups when implementing. This is on top of moving file contracts to disk so look at that first.

### DIFF
--- a/modules/consensus/accept_test.go
+++ b/modules/consensus/accept_test.go
@@ -439,16 +439,19 @@ func (cst *consensusSetTester) testSpendSiacoinsBlock() error {
 
 	// Find the destAddr among the outputs.
 	var found bool
-	for id, output := range cst.cs.siacoinOutputs {
+	cst.cs.db.forEachSiacoinOutputs(func(id types.SiacoinOutputID, output types.SiacoinOutput) {
 		if id == outputID {
 			if found {
-				return errors.New("output found twice")
+				err = errors.New("output found twice")
 			}
 			if output.Value.Cmp(txnValue) != 0 {
-				return errors.New("output has wrong value")
+				err = errors.New("output has wrong value")
 			}
 			found = true
 		}
+	})
+	if err != nil {
+		return err
 	}
 	if !found {
 		return errors.New("could not find created siacoin output")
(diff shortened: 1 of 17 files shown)


DavidVorick:
looks like this needs to be rebased on top of the other one.

DavidVorick:
test timeout

triazo:
Seems to be better now. It seems that Travis was just being really slow, as the timeout occurred on creating the consensus set tester object, which every other test case does.

DavidVorick:
I am cautious to think that Travis is at fault. It's weird that it would take 3 minutes to make a consensus set tester object, and difficult to blame that on Travis.

DavidVorick:
there are now merge conflicts due to my recent PR being merged.

modules/consensus/accept_test.go
```diff
@@ -439,16 +450,19 @@ func (cst *consensusSetTester) testSpendSiacoinsBlock() error {
...
-	for id, output := range cst.cs.siacoinOutputs {
+	cst.cs.db.forEachSiacoinOutputs(func(id types.SiacoinOutputID, output types.SiacoinOutput) {
 		if id == outputID {
 			if found {
-				return errors.New("output found twice")
+				err = errors.New("output found twice")
 			}
 			if output.Value.Cmp(txnValue) != 0 {
-				return errors.New("output has wrong value")
+				err = errors.New("output has wrong value")
```
DavidVorick: You want to let it finish iterating even though you are going to return an error at the end anyway?

modules/consensus/accept_test.go
```diff
@@ -1495,14 +1513,17 @@ func TestComplexForking(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	defer cst1.cs.Close()
```
DavidVorick: did you mean to call closeCst()?

modules/consensus/accept_test.go
```diff
@@ -1495,14 +1513,17 @@ func TestComplexForking(t *testing.T) {
...
 	cst2, err := createConsensusSetTester("TestComplexForking - 2")
 	if err != nil {
 		t.Fatal(err)
 	}
+	defer cst2.cs.Close()
 	cst3, err := createConsensusSetTester("TestComplexForking - 3")
 	if err != nil {
 		t.Fatal(err)
 	}
+	defer cst3.cs.Close()
```
DavidVorick: ditto for cst2 and cst3

modules/consensus/consensusset_test.go
```diff
@@ -87,6 +87,11 @@ func createConsensusSetTester(name string) (*consensusSetTester, error) {
 	return cst, nil
 }
 
+// closeCst safely closes the consensus set tester. 'close' is a builtin
+func (cst *consensusSetTester) closeCst() error {
+	return cst.gateway.Close()
+}
```
DavidVorick: @lukechampine 

Should instead this be `cst.cs.Close()` and then the consensus set calls `gateway.Close()`? Or do you think that modules should not close their dependencies.

I guess modules should not close their dependencies.

modules/consensus/accept_test.go
triazo: It looks like all the places where this function is called have error handling, so I can break out by returning an error.

PR #702, opened 2015-08-09
Reflected Competition & Potential Profit in API.md

Mingling94:


doc/API.md
```diff
@@ -274,8 +274,18 @@ struct {
...
+	Potential Profit int
+	Competition      int
 }
 ```
+`StorageRemaining` is how much storage (in bytes) the host has available for
+renting out.
+
+`NumContracts` is the number of contacts the Host is currently handling
+
+`Potential Profit` is how much the host's current contracts are worth
```
lukechampine: The `PotentialProfit` field does not have a space.

Mingling94:
@lukechampine bloop

PR #703, opened 2015-08-10
Move Delayed Siacoin Outputs to disk

triazo:
As this was a map inside of a map, I implemented this as one bucket pointing to the id's of a bunch of other buckets in the same root level. It feels silly, but this allowed me to use all of the functions already in the set database for modifying top level buckets, making the code a bit cleaner.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,27 +75,27 @@ test: clean fmt REBUILD
 test-v: clean fmt REBUILD
 	go test -race -v -short -tags='debug testing' -timeout=20s $(pkgs)
 test-long: clean fmt REBUILD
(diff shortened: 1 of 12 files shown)


DavidVorick:
More timeouts, TestBlockHistory, and 1200 gothreads in the panic?

Why are there so many threads now?

lukechampine:
#704

DavidVorick:
merge conflicts

triazo:
Fixed conflicts. Its still possible it could fail due to timeout.

triazo:
Doubled the timeouts.

PR #705, opened 2015-08-11
Wallet api

DavidVorick:
The wallet api is now complete.

I am worried a bit about inconsistency, some crypto.Hash alias are encoded into json as hex, and others are encoded into json as byte arrays, and addresses are encoded as byte arrays with checksums.

I think the line count on this PR is scarier than the actual PR.


lukechampine:
Perhaps we should refrain from doing any formatting on the API-side, and leave it up to the frontend to format values as they see fit.

api/api.go
```diff
@@ -93,31 +92,32 @@ func (srv *Server) initAPI(addr string) {
...
+	// BlockExplorer API Calls - Unfinished
 	if srv.exp != nil {
 		handleHTTPRequest(mux, "/explorer/status", srv.explorerStatusHandler)
 		handleHTTPRequest(mux, "/explorer/blockdata", srv.explorerBlockDataHandler)
 		handleHTTPRequest(mux, "/explorer/gethash", srv.explorerGetHashHandler)
-
 	}
 
-	// create graceful HTTP server
+	// create graceful HTTP server - Unfinished
```
lukechampine: unfinished, eh?

I don't see why the wallet is supposedly "finished" but no other modules are.

api/consensus.go
```diff
@@ -4,28 +4,17 @@ import (
...
 
-// The ConsensusSetStatus struct contains general information about the
-// consensus set, with tags to have idiomatic json encodings.
-type ConsensusSetStatus struct {
+// The ConsensusGET struct contains general information about the consensus
+// set, with tags to support idiomatic json encodings.
+type ConsensusGET struct {
 	Height       types.BlockHeight `json:"height"`
 	CurrentBlock types.BlockID     `json:"currentblock"`
 	Target       types.Target      `json:"target"`
```
lukechampine: off-topic, but isn't it confusing to have some types lowercase and others uppercase?

api/api.go
DavidVorick: The wallet has been moved over to the new RESTful paradigm that we layed out around 6 weeks ago.

The other modules haven't been moved over yet.

api/daemon.go
```diff
@@ -57,6 +54,39 @@ bwIDAQAB
 // manifest.
 var updateURL = "http://23.239.14.98/releases/" + runtime.GOOS + "_" + runtime.GOARCH
 
+// SiaConstants is a struct listing all of the constants in use.
+type SiaConstants struct {
```
lukechampine: maybe this should be defined in `types`, eh?

api/consensus.go
DavidVorick: are you refering to `currentblock` or are you referring to `ConsensusGET` vs. `types.Transaction`?

`currentblock` is inconsistent with everything else, and should either be `currentBlock` or everything else should be switched to all lowercase.

api/wallet.go
```diff
@@ -25,125 +74,343 @@ func scanAmount(amount string) (types.Currency, bool) {
...
 		return
 	}
 
-	// Since coinAddress is not a struct, we define one here so that writeJSON
-	// writes an object instead of a bare value. In addition, we transmit the
-	// coinAddress as a hex-encoded string rather than a byte array.
-	writeJSON(w, struct{ Address types.UnlockHash }{unlockConditions.UnlockHash()})
+	confirmedHistory, err := srv.wallet.History(types.BlockHeight(start), types.BlockHeight(end))
+	if err != nil {
+		writeError(w, "/walet/history [GET] Error:"+err.Error(), http.StatusBadRequest)
```
lukechampine: typo, and this message is inconsistent with others.

lukechampine:
definitely consider writing an API method wrapper, so you can write:
```go
handleHTTPRequest(mux, "/wallet/seed", restrictMethods(srv.walletSeedHandler, "GET", "PUT", "POST"))
```
and save the trouble of writing 50 if-else blocks and redirection functions.

api/consensus.go
lukechampine: the latter; if we want to be consistent, we need to add json tags to every field of every struct that gets sent over the API. That's not so bad as long as most of those structs are defined in the api package, though. Perhaps we should enforce that all structs sent over the API **must** be defined in the api package, to avoid that sort of thing (at the expense of some duplicate types).

api/wallet.go
```diff
@@ -25,125 +74,343 @@ func scanAmount(amount string) (types.Currency, bool) {
...
-func (srv *Server) walletMergeHandler(w http.ResponseWriter, req *http.Request) {
-	// Scan the wallet file.
-	err := srv.wallet.MergeWallet(req.FormValue("walletfile"))
+// walletHistoryHandlerGETaddr handles a GET request to
+// /wallet/history/$(addr).
+func (srv *Server) walletHistoryHandlerGETaddr(w http.ResponseWriter, req *http.Request, addr types.UnlockHash) {
+	addrHistory, err := srv.wallet.AddressHistory(addr)
 	if err != nil {
-		writeError(w, err.Error(), http.StatusBadRequest)
+		writeError(w, "error after call to /wallet/history/$(addr): "+err.Error(), http.StatusBadRequest)
```
lukechampine: also inconsistent. Use `req.Url` to display the call, at least.

api/wallet.go
```diff
@@ -25,125 +74,343 @@ func scanAmount(amount string) (types.Currency, bool) {
...
+
+	// The only remaining possibility is a GET call to /wallet/history/$(addr);
+	// check that the method is correct.
+	if req.Method != "GET" && req.Method != "" {
+		writeError(w, "unrecognized method in call to /wallet/history", http.StatusBadRequest)
+		return
+	}
+
+	// Parse the address from the url and call the GETaddr Handler.
+	jsonAddr := "\"" + req.URL.Path[len("/wallet/history/"):] + "\""
```
lukechampine: use `strings.TrimPrefix`

api/wallet.go
```diff
@@ -25,125 +74,343 @@ func scanAmount(amount string) (types.Currency, bool) {
...
+		allSeedsStrs = append(allSeedsStrs, str)
+	}
+	writeJSON(w, WalletSeedGET{
+		PrimarySeed:        primarySeedStr,
+		AddressesRemaining: int(modules.PublicKeysPerSeed - progress),
+		AllSeeds:           allSeedsStrs,
+	})
+}
+
+// walletSeedHandlerPUT handles a PUT request to /wallet/seed.
```
lukechampine: these docstrings could be a bit more helpful. What is the difference between PUT and POST? I'd have to read the function body to find out.

api/wallet.go
```diff
@@ -25,125 +74,343 @@ func scanAmount(amount string) (types.Currency, bool) {
...
+func (srv *Server) walletSeedHandler(w http.ResponseWriter, req *http.Request) {
+	if req.Method == "" || req.Method == "GET" {
+		srv.walletSeedHandlerGET(w, req)
+	} else if req.Method == "PUT" {
+		srv.walletSeedHandlerPUT(w, req)
+	} else if req.Method == "POST" {
+		srv.walletSeedHandlerPOST(w, req)
+	} else {
+		writeError(w, "unrecognized method when calling /wallet/seed", http.StatusBadRequest)
+	}
```
lukechampine: use a switch:
```go
switch req.Method {
case "", "GET":
// etc.
```

api/wallet.go
```diff
@@ -25,125 +74,343 @@ func scanAmount(amount string) (types.Currency, bool) {
...
 
-		// Scan the keyfile list.
-		keyfiles := strings.Split(req.FormValue("keyfiles"), ",")
-		if len(keyfiles) == 0 {
-			writeError(w, "Missing keyfiles", http.StatusBadRequest)
-			return
-		}
+	// Parse the id from the url.
+	var id types.TransactionID
+	jsonID := "\"" + req.URL.Path[len("/wallet/transaction/"):] + "\""
```
lukechampine: ditto

api/consensus.go
DavidVorick: It seems acceptable to me to add json tags to all of the structs in the types package

doc/API.md
```diff
@@ -4,13 +4,17 @@ Siad API
 The siad API is currently under construction. The deprecated way of doing
 things is documented at the end of the (incomplete) new documentation.
 
+All responses from the API contain a 'success' field (a bool) and an 'error'
+field (a string). If the query returns an error, the 'success' field will be
+set to false and there will be an error message. If no error occurs, 'success'
+will be set to true and the error field will be empty.
```
lukechampine: this doesn't appear to be the case. Also, you can just leave off the error field entirely from successful responses.

doc/API.md
```diff
@@ -28,17 +32,379 @@ struct {
...
+
+'confirmedSiacoinBalance' is the number of siacoins available to the wallet as
+of the most recent block in the blockchain.
+
+'unconfirmedOutgoingSiacoins' is the number of siacoins that are leaving the
+wallet according to the set of unconfirmed transactions. Often this number
+appears inflated, because outputs are frequently larger than the number of
+coins being sent, and there is a refund. These coins are counted as outgoing,
+and the refund is counted as incoming. The difference in balance can be
+claculated using 'unconfirmedIncomingSiacoins' - 'unconfirmedOutgoingSiacoins'
```
lukechampine: typo

doc/API.md
```diff
@@ -28,17 +32,379 @@ struct {
...
+}
+```
+'confirmedTransactions' lists all of the confirmed wallet transactions
+appearing between height 'start' and height 'end' (inclusive).
+
+'unconfirmedTransactions' lists all of the unconfirmed transactions.
+
+Wallet transactions are transactions that have been processed by the wallet and
+given more information, such as a confirmation height and a timestamp. Each
+wallet transaction contins information about only a single input or output. One
```
lukechampine: typo

doc/API.md
```diff
@@ -28,17 +32,379 @@ struct {
...
+	RelatedAddress string
+	Value          int
+}
+```
+'TransactionID' is the id of the transaction from which the wallet transaction
+was derived. The full transaction can be obtaied by calling
+'/wallet/transaction/$(id)'
+
+'ConfirmationHeight' is the height at which the transaction was confirmed. The
+height will be set to 'uint64MAX' if the transaction has not been confirmed.
```
lukechampine: well that's gonna cause problems, since in the struct its type is `int`. How about -1?

doc/API.md
```diff
@@ -28,17 +32,379 @@ struct {
...
+
+Parameters: none
+
+Response:
+```
+struct {
+	address string
+}
+```
+'address' is the address that cna
```
lukechampine: ...

doc/API.md
```diff
@@ -28,17 +32,379 @@ struct {
...
+	newSeed string
+}
+```
+'newSeed' is the new seed that will be used as the seed for generating new
+addresses.
+
+#### /wallet/siacoins [PUT]
+
+Function: Send siacoins to an address. The outputs are arbitrarily selected
+from addresses in the wallet.
```
lukechampine: PUT is usually used to update the resource referenced by the URL path. So `PUT /wallet/siacoins/${addr}` would make a bit more sense. Even then, PUT (as I understand it) usually means you are overwriting the existing value, whereas here we are adding to it.

doc/API.md
```diff
@@ -28,17 +32,379 @@ struct {
...
+can be used to derive secret keys, and must be kept safe. This call is
+unavailable when the wallet is locked or closed.
+
+Parameters:
+```
+struct {
+	dictionary string
+}
+```
+'dictionary' is the name of the dicitionary that should be used when encoding
```
lukechampine: typo

doc/API.md
```diff
@@ -28,17 +32,379 @@ struct {
...
+copying. The library
+[entropy-mnemonics](https://github.com/NebulousLabs/entropy-mnemonics) is used
+when encoding.
+
+#### /wallet/seed [PUT]
+
+Function: Get a new address from the wallet generated by the primary seed. An
+error will be returned if the seed has been exhausted (there is a limit on the
+number of addresses that can be generated from one seed), or if the wallet is
+locked/closed.
```
lukechampine: PUT takes no params and returns a new resource? Seems off to me. I would use POST here.

doc/API.md
```diff
@@ -28,17 +32,379 @@ struct {
...
+	confirmedTransactions   []types.Transaction
+	unconfirmedTransactions []types.Transaction
+}
+```
+'confirmedTransactions' lists all of the confirmed transactions appearing
+between height 'start' and height 'end' (inclusive).
+
+'unconfirmedTransactions' lists all of the unconfirmed transactions.
+
+#### /wallet/unlock [PUT]
```
lukechampine: I would also use POST here.

modules/wallet/history.go
```diff
@@ -80,3 +80,46 @@ func (w *Wallet) AddressUnconfirmedTransactions(uh types.UnlockHash) (wts []modu
...
+	defer w.mu.Unlock(lockID)
+
+	if startHeight > w.consensusSetHeight || startHeight > endHeight {
+		return nil, errOutOfBounds
+	}
+	if len(w.walletTransactions) == 0 {
+		return nil, nil
+	}
+
+	var prevTxid types.TransactionID
```
lukechampine: what is this used for? Some sort of duplicate check, it looks like?

api/api.go
lukechampine: At any rate, I think this one was a typo.

api/consensus.go
lukechampine: yeah I don't like having to conform to JavaScript conventions. I would rather just leave them all uppercase, but I'm not 100% confident that's the right thing to do.

lukechampine:
ok, done with first round of comments

doc/API.md
DavidVorick: don't even worry about it

api/daemon.go
DavidVorick: we'd have `types.Constants.MaturityDelay`?

I don't like it

api/wallet.go
DavidVorick: what's wrong with this one?

api/wallet.go
DavidVorick: you're supposed to reference the api documentation.

Maybe it's the wrong approach, but the idea was that we could make the api documentation more consistent with the code behavior if you did all the coding directly from the api docs.

api/wallet.go
DavidVorick: switch everywhere, or just here?

doc/API.md
DavidVorick: This is currently not implemented. It was more of a proposal. I was wondering if we could wrap it into the writeError and writeJSON functions

doc/API.md
DavidVorick: hmmm.

Maybe it should be a types.BlockHeight. I'm not sure why I suddenly made everything an int in the json structs. But we should probably stick to the original types I guess.

doc/API.md
DavidVorick: What would you recommend?

I don't like the idea of calling /wallet/siacoins/$(addr), it doesn't seem very clear to me what is happening. Is it coming from that addr or going to that addr?

Ming's gut reaction was that it should be POST

doc/API.md
DavidVorick: /wallet/seed POST does something completely different.

/wallet/seed PUT grabs the next address from the existing seed, essentially advancing the seed a step.

/wallet/seed POST replaces the current seed with a completely new seed.

modules/wallet/history.go
DavidVorick: oh yeah I can add a comment. It's a duplicate check.

doc/API.md
DavidVorick: And also for /wallet/close?

/wallet/unlock and /wallet/close are inverses of eachother. Also, you can cycle between them repeatedly.

api/consensus.go
DavidVorick: I guess we'll drop it for now then.

api/wallet.go
DavidVorick: I will look into this more down the line

doc/API.md
DavidVorick: tipo

DavidVorick:
> definitely consider writing an API method wrapper, so you can write:
```go
handleHTTPRequest(mux, "/wallet/seed",
    restrictMethods(srv.walletSeedHandler, "GET", "PUT", "POST"))
```
> and save the trouble of writing 50 if-else blocks and redirection functions.
> At this point we ought to consider using a framework though, such as [gorilla/mux](https://github.com/gorilla/mux).

hmmm. I don't think having a wrapper would help too much, you still have to write all the code that handles the methods differently. In most cases they run completely different logic depending on what method you are using.

Haven't looked very much into the mux framework, I'm naturally opposed to dependencies, but for something like an API it's a lot less critical than for consensus code. Mux also seems reasonably well maintained.

api/daemon.go
lukechampine: well, it does have the advantage of making it clear that MaturityDelay is a constant and not a type...
I'll drop it for now

api/wallet.go
lukechampine: just that you should use the actual call instead of `(`$addr)`

DavidVorick: In the documentation, and for the other stuff, I treat anything `/$(xxx)` as a separate call. Same goes for `/wallet/transaction/$(id)`.

doc/API.md
lukechampine: `POST 10KS /wallet/siacoins/send` or `/wallet/send/siacoins` would be pretty clear. Just changing it to POST seems okay though.

doc/API.md
lukechampine: yeah, I would either switch the POST and PUT, or consider adding a new route like `/wallet/seed/new/` or `/wallet/seed/nextaddress/`

doc/API.md
DavidVorick: I will switch it to post

doc/API.md
DavidVorick: I'll do `wallet/seed/new`, it should probably be explicit anyway.

DavidVorick:
I kept the way that /wallet/seed works. I'm not 100% sure switching is the best approach, perhaps we should consult @agundy

PR #706, opened 2015-08-12
Mining pool - rough draft

JoshVorick:
This pull request contains the code for a mining pool that technically works (I'm pretty sure, haven't done much testing beyond one miner connecting to the pool viz localhost). 

I wanted to make this pull request for two reasons:
* Avoid making a 1500+ line PR later on
* Get the structure of the pool reviewed before I work out all the finer details.

This code is littered with TODOs. If there's anything that the pool code should include that isn't in a TODO comment, let me know and I'll throw it in. I think the one major thing that the pool needs that isn't written in the code is some sort of persistence layer so that the pool can save/load payment channels in case siad gets restarted.

The basic structure of the pool:

1.  API call to siad 'miner/connecttopool' this connects to localhost:9983 and retrieves the mining pool's settings. After getting the pool's settings, the miner makes an RPC to the pool to negotiate a new payment channel.
2.  The miner makes calls to miner/poolheaderforwork/, which returns headers of blocks with the proper pool payouts and adjusted target.
3.  The miner calls miner/submitpoolheader when it finds a partial block. This submits the block to the pool, and retrieves a payout transaction from the pool. For now, the miner immediately broadcasts said transaction, effectively closing the payment channel. The miner then makes an RPC to the pool to negotiate a new payment channel.

Right now the pool is insecure because the miner and the pool both assume each other is doing the 'right' thing (e.g. submitting valid blocks, signing transactions, etc).


api/api.go
```diff
@@ -76,6 +76,14 @@ func (srv *Server) initAPI(addr string) {
 		handleHTTPRequest(mux, "/miner/submitblock", srv.minerSubmitblockHandler)
 		handleHTTPRequest(mux, "/miner/headerforwork", srv.minerHeaderforworkHandler)
 		handleHTTPRequest(mux, "/miner/submitheader", srv.minerSubmitheaderHandler)
+		handleHTTPRequest(mux, "/miner/connecttopool", srv.minerConnectToPoolHandler)
+		handleHTTPRequest(mux, "/miner/poolheaderforwork", srv.minerPoolHeaderForWorkHandler)
+		handleHTTPRequest(mux, "/miner/submitpoolheader", srv.minerSubmitPoolHeaderHandler)
```
lukechampine: to match the existing API calls, I would suggest `/miner/pool/connect`, `/miner/pool/headerforwork`, and `/miner/pool/submitheader`

modules/miner.go
```diff
@@ -31,6 +31,25 @@ type BlockManager interface {
...
+
+	// HeaderForPool returns the header of a block that is ready for pool
+	// mining.  The block contains all the correct pool payouts. The header is
+	// meant to be grinded by a miner and, shuold the target be beat,
+	// resubmitted through SubmitHeaderToPool
+	PoolHeaderForWork() (types.BlockHeader, types.Target)
+
+	// SubmitPoolHeader takes a header that has been solved and submits it to
+	// the pool
+	SubmitPoolHeader(types.BlockHeader) error
```
lukechampine: `PoolSubmitHeader` is probably better

modules/miner/poolmanager.go
```diff
@@ -0,0 +1,292 @@
+package miner
+
+import (
+	"errors"
+	"fmt"
```
lukechampine: but why

modules/miner/poolmanager.go
```diff
@@ -0,0 +1,292 @@
...
+func (m *Miner) ConnectToPool(ip string) error {
+	fmt.Println("connect to pool: ", ip)
+	conn, err := net.DialTimeout("tcp", ip, 10e9)
+	if err != nil {
+		return err
+	}
+	defer conn.Close()
+	m.poolIP = ip
+
+	err = encoding.WriteObject(conn, [8]byte{'S', 'e', 't', 't', 'i', 'n', 'g', 's'})
```
lukechampine: should probably be defined as a variable in modules/miner.go.
I recommend defining it as a `types.Specifier`.

modules/miningpool/blockmanager.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/NebulousLabs/Sia/encoding"
+	"github.com/NebulousLabs/Sia/types"
+)
+
+// SubmitBlockShare does TODO
+func (mp *MiningPool) rpcSubmit(conn net.Conn) error {
+	fmt.Println("Block attempt submitted") // testing statement
+	// Read the Block in from conn
+	var b types.Block
+	err := encoding.ReadObject(conn, &b, 1e6) // TODO: figure out how big the read number should be
```
lukechampine: I believe there's a MaxBlockSize defined somewhere...

types/constants.go
```diff
@@ -61,7 +61,7 @@ func init() {
 		BlockFrequency = 6                       // 6 seconds: slow enough for developers to see ~each block, fast enough that blocks don't waste time.
 		MaturityDelay = 10                       // 60 seconds before a delayed output matures.
 		GenesisTimestamp = Timestamp(1424139000) // Change as necessary.
-		RootTarget = Target{0, 0, 4}             // Standard developer CPUs will be able to mine blocks.
+		RootTarget = Target{4, 0, 4}             // Standard developer CPUs will be able to mine blocks.
```
lukechampine: hmmmm

lukechampine:
I don't like all the `fmt.Println`s.

siad/daemon.go
```diff
@@ -66,7 +71,7 @@ func startDaemon() error {
 	if err != nil {
 		return err
 	}
-	srv, err := api.NewServer(config.Siad.APIaddr, state, gateway, host, hostdb, miner, renter, tpool, wallet, nil)
+	srv, err := api.NewServer(config.Siad.APIaddr, state, gateway, host, hostdb, miner, miningpool, renter, tpool, wallet, nil)
```
lukechampine: so, everyone is going to run a miningpool now?

types/constants.go
JoshVorick: Oops, shouldn't have let this slip in.

modules/miner/poolmanager.go
JoshVorick: I left in all my testing statements because it's too much to remove them every time I make a commit and the pool is definitely still in testing.

api/api.go
DavidVorick: Under the new paradigm, I think we'd see the following calls:

/miner GET -> return the miner status
/miner/start POST -> start the miner
/miner/stop POST -> stop the miner

/miner/block GET -> get a block for work
/miner/block POST -> submit a block that was worked

/miner/header GET -> get a header for work
/miner/header POST -> submit a header for work

/miner/pool  PUT -> connect to a pool (or maybe /miner/pool/connect PUT)
/miner/pool GET -> get a pool header for work
/miner/pool POST -> submit a solved pool header

DavidVorick: /miner/pool/connect PUT seems like a better choice to me than /miner/pool PUT

modules/miningpool/blockmanager.go
DavidVorick: yeah types/constants.go

types/constants.go
DavidVorick: This would suggest that the developer is getting a few dozen hashes per second.

This is really slow, you are probably hashing an entire block somewhere, or calling ID() a bunch of times or something.

siad/daemon.go
DavidVorick: siap

api/api.go
lukechampine: POST! not PUT. PUT is for updating resources.

siad/daemon.go
lukechampine: if we're gonna stick `nil` checks everywhere anyway, we might as well keep it all in one binary.

api/api.go
DavidVorick: ok

DavidVorick: is the connect call idempotent? :P

api/miner.go
```diff
@@ -98,3 +98,55 @@ func (srv *Server) minerSubmitheaderHandler(w http.ResponseWriter, req *http.Req
...
+		return
+	}
+	writeSuccess(w)
+}
+
+// minerPoolHeaderForWorkHandler handles the API call to request a header
+// from the miner for pool ming
+func (srv *Server) minerPoolHeaderForWorkHandler(w http.ResponseWriter, req *http.Request) {
+	bhfw, target := srv.miner.PoolHeaderForWork()
+	w.Write(encoding.MarshalAll(target, bhfw))
```
DavidVorick: I forgot that the header doesn't submit json.

api/miner.go
```diff
@@ -98,3 +98,55 @@ func (srv *Server) minerSubmitheaderHandler(w http.ResponseWriter, req *http.Req
 	}
 	writeSuccess(w)
 }
+
+// minerConnectToPoolHandler handles the API call to connect to a mining pool
+func (srv *Server) minerConnectToPoolHandler(w http.ResponseWriter, req *http.Request) {
+	var ip string
+	// TODO: actually read ip from req.Body
+	/*encodedIP, err := ioutil.ReadAll(req.Body)
```
DavidVorick: I believe there are some helper functions for this already.

api/miner.go
```diff
@@ -98,3 +98,55 @@ func (srv *Server) minerSubmitheaderHandler(w http.ResponseWriter, req *http.Req
...
+// minerPoolHeaderForWorkHandler handles the API call to request a header
+// from the miner for pool ming
+func (srv *Server) minerPoolHeaderForWorkHandler(w http.ResponseWriter, req *http.Request) {
+	bhfw, target := srv.miner.PoolHeaderForWork()
+	w.Write(encoding.MarshalAll(target, bhfw))
+}
+
+// minerSubmitPoolHeaderHandler handles the API call to submit a header to
+// the miner
+func (srv *Server) minerSubmitPoolHeaderHandler(w http.ResponseWriter, req *http.Request) {
```
DavidVorick: naming convention is to put caps following the '/'s.

minerSubmitpoolheaderHandlerPOST

modules/miner/blockmanager.go
```diff
@@ -9,6 +9,21 @@ import (
...
+		return nil, errors.New("Header is either invalid or too old")
+	}
+	arbData, exists := m.arbDataMem[header]
+	if !exists {
+		return nil, errors.New("Header is either invalid or too old")
+	}
+
+	block.Transactions[0].ArbitraryData[0] = arbData
+	return block, nil
+}
```
DavidVorick: for some reason I thought you already PR'd all of this stuff?

modules/miner/poolmanager.go
DavidVorick: We can't merge anything with testing statements.

modules/miner/poolmanager.go
```diff
@@ -0,0 +1,292 @@
...
+
+	// Connect to the pool
+	conn, err := net.DialTimeout("tcp", m.poolIP, 10e9)
+	if err != nil {
+		fmt.Println(1)
+		return err
+	}
+	defer conn.Close()
+
+	err = encoding.WriteObject(conn, [8]byte{'C', 'h', 'a', 'n', 'n', 'e', 'l'})
```
DavidVorick: I think this is Luke's problem, but that should probably be a 'types.Specifier' instead of a [8]byte

modules/miner/poolmanager.go
```diff
@@ -0,0 +1,292 @@
...
+
+	// TODO: Check that the pool's transactions are correct (not trying to cheat us)
+
+	// Sign the refund transaction, but with a timelock. This way the pool will
+	// know it can get its money back if the miner disappears or fails to mine
+	// any blocks.
+	refundTxn.TransactionSignatures = append(refundTxn.TransactionSignatures, types.TransactionSignature{
+		ParentID:       crypto.Hash(refundTxn.SiacoinInputs[0].ParentID),
+		PublicKeyIndex: 1,
+		Timelock:       m.height + 30, // TODO: Make this a less arbitrary number
```
DavidVorick: 1 week is good. 1008

modules/miner/poolmanager.go
```diff
@@ -0,0 +1,292 @@
...
+	if err != nil {
+		return err
+	}
+	m.poolTransaction.TransactionSignatures[1].Signature = cryptoSig[:]
+	err = m.tpool.AcceptTransactionSet([]types.Transaction{m.poolTransaction})
+	if err != nil {
+		return err
+	}
+
+	// TODO: Tell pool that we closed the channel
```
DavidVorick: pool should be watching blockchain, hopefully no need to tell pool.

Pool also has nothing to lose lol.

modules/miningpool/miningpool.go
```diff
@@ -0,0 +1,127 @@
...
+)
+
+type MiningPool struct {
+	// Module dependencies
+	cs     modules.ConsensusSet
+	tpool  modules.TransactionPool
+	wallet modules.Wallet
+
+	// List of headers whose miner has already been paid (prevent double submits)
+	spentHeaders []types.Block // Should this be a map?
```
DavidVorick: yes it should be a map

siad/daemon.go
DavidVorick: just a joke lol

DavidVorick:
I can't say that I looked at it too closely, looks like it's still a pretty big work in progress.

But overall it seems like the right direction.

modules/miner/miner.go
```diff
@@ -67,6 +68,23 @@ type Miner struct {
...
+	// with the pool are stored as well as constants like the percentage of a
+	// block's subsidy that goes to the miner.
+	// The poolTransaction is the transaction in the payment channel whose
+	// output is the miner's wallet. The miner may broadcast this transaction
+	// to the network at any time in order to receive payment from the pool.
+	// poolHeaderMap is a map from headers of blocks that payout to the pool to
+	// headers of blocks with normal payouts. This allows the poolmanager to
+	// use the blockmanager and only have to worry about changing payouts.
+	poolIP            string
+	minerPercentCut   uint8
```
lukechampine: use a `big.Rat`

modules/miner/blockmanager.go
JoshVorick: I did. But I moved those to its own function.

api/api.go
lukechampine: idempotency is only part of the picture. I think in general it's confusing to use PUT when you aren't actually PUTting anything.

modules/miner/poolmanager.go
JoshVorick: The pool shouldn't keep spent payment channels in memory.

But it should be watching the blockchain, not relying on miners to tell the pool when they spend the output.

JoshVorick:
I think I addressed everything you guys mentioned in this PR in some form or another.

Some of it, I'll address more thoroughly in the next pull request.

DavidVorick:
did you deal with the precision? Luke and I both feel that, barring some strong technical reason, you should be using a big.Rat to do the percentage stuff.

JoshVorick:
Currently, they're just uint8s, so no precision loss. I will change them to big.Rats eventually, but encoding big.Rats isn't trivial, so I feel like it is more important to continue adding functionality before spending time on converting them to big.Rats.

lukechampine:
uint8s are fine until you want something fine-grained. What if you want 0.8%, or 5.6%?

What we've discovered from working on Sia is that it's very dangerous to put off protocol changes until later. You really want to get your protocols as perfect and stable as possible before releasing them, because it's a big pain to revise a protocol later on. Backwards compatibility is a major headache for us.

If you don't think a big.Rat is justified, you could roll your own fractions using a pair of uint64s. That would probably provide more than enough precision for this use case. (Maybe @DavidVorick disagrees.) But a uint8 is definitely not gonna cut it. Also, in general you want to multiply by the numerator and *then* divide by the denominator, not the other way around; this minimizes rounding errors. I think your current code gets it backwards.

JoshVorick:
I defnitely plan to change them to big.Rats before release, but I don't think its in the scope of the rough draft.

DavidVorick:
so close... but so far...

chy373:
@DavidVorick hello, is mining pool supported now? I want to set up a sia mining pool

PR #707, opened 2015-08-12
Finish moving structures to disk

triazo:
This last pr covers file contract expirations and the siafund pool. The siafund pool is still kept in memory as it is just one value which is used a fair amount, but mirrored on disk on every change for loading.

With these two variables moved to disk, the consensus set can be restored without reapplying every block from disk, greatly speeding up startup. They still need to be loaded onto disk to broadcast updates, but I expect that if this is time consuming it would be a relatively simple fix.


DavidVorick:
needs rebasing

PR #708, opened 2015-08-12
Miner startup

DavidVorick:
The miner and the host each needed some adjustments so that they could start up without the wallet being unlocked, without impacting their ability to receive coins when mining/hosting.

The adjustments have been made.

The miner interface needed to change a bit. As the miner is so integral to testing, that resulting in some wider changes too.


lukechampine:
can you rebase? Looks like this PR includes the longer timeouts introduced by adam.

DavidVorick:
np

PR #711, opened 2015-08-13
Wallet API Testing

DavidVorick:
I cleaned up the documentation and server tester.

I switched from verifying batches of 50 blocks to verifying batches of 10 blocks because the database is slow and it causes problems to do it in batches of 50.

/wallet/seed PUT was changed to /wallet/address POST

Some of the api documentation was cleaned up.

The server tester object got functions for doing POST requests.


api/wallet_test.go
```diff
@@ -0,0 +1,170 @@
...
+// wallet has never been encrypted.
+func TestIntegrationWalletGETEncrypted(t *testing.T) {
+	if testing.Short() {
+		t.SkipNow()
+	}
+
+	// Check a wallet that has never been encrypted.
+	testdir := build.TempDir("api", "TestIntegrationWalletGETEncrypted")
+	APIAddr := ":" + strconv.Itoa(APIPort)
+	APIPort++
```
lukechampine: we're still using a global API port variable? I thought I got rid of that.

api/wallet_test.go
```diff
@@ -0,0 +1,170 @@
...
+		t.Error("there should not be unconfirmed incoming siacoins")
+	}
+
+	// Send coins to a wallet address through the api.
+	var wap WalletAddressPOST
+	err = st.postAPI("/wallet/address", url.Values{}, &wap)
+	if err != nil {
+		t.Fatal(err)
+	}
+	sendSiacoinsValues := url.Values{}
```
lukechampine: generally prefer `var sendSiacoinsValues url.Values`

modules/consensus/synchronize_test.go
```diff
@@ -171,7 +171,7 @@ func TestBlockHistory(t *testing.T) {
 	defer cst.closeCst()
 
 	// mine until we have enough blocks to test blockHistory
-	for cst.cs.Height() < MaxCatchUpBlocks {
+	for cst.cs.Height() < 50 {
```
lukechampine: why did changing `MaxCatchUpBlocks` break this?

modules/consensus/synchronize.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 const (
-	MaxCatchUpBlocks          = 50
+	MaxCatchUpBlocks          = 10
 	MaxSynchronizeAttempts    = 8
 	ResynchronizePeerTimeout  = time.Second * 30
 	ResynchronizeBatchTimeout = time.Minute * 3
```
lukechampine: these vars aren't used anymore

modules/consensus/synchronize.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 const (
-	MaxCatchUpBlocks          = 50
+	MaxCatchUpBlocks          = 10
```
lukechampine: what's the rationale here?

api/wallet_test.go
DavidVorick: a ghost?

I just copy-pasted whatever had existed in the gateway test I think

modules/consensus/synchronize_test.go
DavidVorick: If you read the rest of the test, the data is input manually to represent what would happen in a request of 50.

And it has nothing to do with MaxCatchUpBlocks lol, I'm not sure why the constant was used in the first place.

api/wallet_test.go
DavidVorick: it's a map, doing `var ssv url.Values` causes nil map errors when you try to use it.

lukechampine: oh yeah. Confusing. I've never liked `url.Values`...usually it's easier to just write the encoded url by hand.

PR #712, opened 2015-08-13
Capitalized json tags in consensus call

Mingling94:


lukechampine:
Capitals are the default. The tags should be removed entirely

PR #714, opened 2015-08-14
Performance and Death

DavidVorick:
So...

This PR is all over the place. It's basically a disaster, but it does provide some strong performance gains. In some places, more than 2x. There are some very obvious places to improve performance further, and I fully plan on doing so in the near future.

There's also a huge downside though. Because what was required was essentially an overhaul/rewrite, I only got partway through with it. There's lots of duplicated code, and the test coverage isn't as reliable as it used to be because some of the tests follow different codepaths than the production code. All of the integration tests are still passing though, and those are pretty comprehensive. I'm fairly confident that nothing crazy-important broke.

The error handling is very inconsistent, the sanity checking isn't as strong as it used to be, and the overall codebase structure has deteriorated substantially.

I also believe that the net line count will be negative after all performance stuff is finished. It's only grossly positive right now because there's so much code duplication between the optimized vs. non-optimzed codepaths. There are also a lot of places where the disk is accessed repeatedly without necessity. For example, getting height requires going to disk right now.

This git history is also very terrible. Commits are not atomic, are littered with profiling code, and would be very temperamental to explore or revise.

Basically this is the worst pull request ever. I can continue to clean up the code if you like, but it's going to be a lot of time and effort.

These changes are here to stay. After everything is reorganized, I believe that the consensus package will be a lot stronger than it was previously. The sanity checking will be more targeted, the error checking will be more consistent, and the code structure will return to approx. its previous integrity. Basically this PR represents a massive work-in-progress and I feel confident about where it's going.


Makefile
```diff
@@ -78,6 +78,7 @@ test-long: clean fmt REBUILD
 	go test -v -race -tags='testing debug' -timeout=360s $(pkgs)
 bench: clean fmt REBUILD
 	go test -tags='testing' -timeout=120s -run=XXX -bench=. $(pkgs)
+	go test -race -tags='testing debug' -timeout=120s -run=XXX -bench=. $(pkgs)
```
lukechampine: wouldn't it make sense to exclude debug checks from benchmarking?

modules/consensus/accept.go
```diff
@@ -25,16 +27,48 @@ var (
...
+	if exists {
+		return ErrBlockKnown
+	}
+
+	// Do read-only verification operations inside of a single read-only boltdb
+	// tx.
+	var parent processedBlock
+	err := cs.db.View(func(tx *bolt.Tx) error {
+		// Check that the parent exists.
+		blockMap := tx.Bucket(BlockMap)
```
lukechampine: will cause a panic if BlockMap does not exist

modules/consensus/consistency.go
```diff
@@ -132,7 +132,7 @@ func (cs *ConsensusSet) consensusSetHash() crypto.Hash {
 	// 4.	current target
 	// 5.	current depth
 	// 6.	current path + diffs
-	// 7.	earliest allowed timestamp of next block
+	// !7	earliest allowed timestamp of next block
```
lukechampine: maybe add a comment in parentheses instead

Makefile
DavidVorick: It does both now.

I can't remember if I intended to commit this or not.

Basically, the tests were running _really_ slow for a bit, and I was trying to figure out what was causing the extreme slowdown.

Turned out that both the wallet and the boltdb changes had each slowed the project down by a few hundred milliseconds per tester object created. I can remove it now that the tests are running reasonably quickly again.

modules/consensus/accept.go
DavidVorick: BlockMap exists :)

PR #715, opened 2015-08-15
Performance and Death

DavidVorick:
This PR has slightly less death, and slightly more performance.

This PR concludes my work on the consensus package for the time being.


DavidVorick:
On an hdd, IBD took about 90 minutes. I consider this acceptable. On an SSD I imagine the speed would be 3x faster or so. I haven't tested.

modules/consensus/info.go
```diff
@@ -55,14 +58,29 @@ func (s *ConsensusSet) ChildTarget(bid types.BlockID) (target types.Target, exis
 // TODO: Right now, this function is only thread safe when called inside of
 // 'ReceiveConsensusSetUpdate', but that should change once boltdb replaces the
 // block mape
-func (s *ConsensusSet) EarliestChildTimestamp(bid types.BlockID) (timestamp types.Timestamp, exists bool) {
-	exists = s.db.inBlockMap(bid)
-	if !exists {
-		return
+func (cs *ConsensusSet) EarliestChildTimestamp(bid types.BlockID) (timestamp types.Timestamp, exists bool) {
+	err := cs.db.View(func(tx *bolt.Tx) error {
```
DavidVorick: needs a lock

DavidVorick:
I just rebased the pull request to pull out a commit related to the wallet seeds.

This commit is pretty important for performance, but I decided that it should be in the wallet PR that I'm currently working on, instead of the consensus PR where it's really out of place.

modules/consensus/consensusset_bench_test.go
```diff
@@ -0,0 +1,19 @@
...
+	"testing"
+)
+
+// BenchmarkCreateServerTester benchmarks creating a server tester from
+// scratch. The consensus package creates over 60 server testers (and
+// counting), and optimizations to the server tester creation process are
+// likely to generalize to the project as a whole.
+func BenchmarkCreateServerTester(b *testing.B) {
+	for i := 0; i < b.N; i++ {
+		_, err := createConsensusSetTester("BenchmarkCreateServerTester - " + strconv.Itoa(i))
```
lukechampine: yikes, how many folders do you wind up with? And shouldn't these be closed after being created?

modules/consensus/processed_block.go
```diff
@@ -42,18 +48,30 @@ type processedBlock struct {
...
 		}
+
+		// Get the next parent's bytes. Because the ordering is specific, the
+		// parent does not need to be decoded entirely to get the desired
+		// information. This provides a performance boost. The id of the next
+		// parent lies at the first 32 bytes, and the timestamp of the block
+		// lies at bytes 40-48.
+		parentBytes := blockMap.Get(parent[:])
+		copy(parent[:], parentBytes[:32])
+		windowTimes[i] = types.Timestamp(binary.LittleEndian.Uint64(parentBytes[40:48]))
```
lukechampine: you're using `binary` here but `encoding` below (in `targetAdjustmentBase`). Should probably be consistent. I would lean towards `encoding`.

DavidVorick: I will use encoding. It's not that slow to copy 8 bytes.

modules/consensus/processed_block.go
```diff
@@ -131,23 +138,56 @@ func clampTargetAdjustment(base *big.Rat) *big.Rat {
...
+	// Fetch the parent block.
+	var parent processedBlock
+	parentBytes := blockMap.Get(pb.Block.ParentID[:])
+	err := encoding.Unmarshal(parentBytes, &parent)
+	if build.DEBUG && err != nil {
+		panic(err)
+	}
+
+	if pb.Height%(types.TargetWindow/2) != 0 {
+		pb.ChildTarget = parent.ChildTarget
```
lukechampine: what's this about?

profile/timing.go
```diff
@@ -0,0 +1,48 @@
...
+	return s
+}
+
+// ToggleTimer actives a timer known by a given string. If the timer does not
+// yet exist, it is created.
+func ToggleTimer(s string) {
+	if times == nil {
+		times = make(map[string]int64)
+		activeTimers = make(map[string]int64)
+	}
```
lukechampine: you can initialize these directly above, in the `var` statements.

lukechampine:
I didn't comb through this too carefully, but I trust that it doesn't break anything.
Long term, we should make two changes:
- Define a wrapper around `bolt.Tx` and write useful methods for it
- Pass the wrapped txn around to every function that is involved in a database update

This will grant us perfect atomicity, removing the need for consistency guards and probably speeding up our benchmarks by a fair margin. However, there are some big structural difficulties in doing this, so it can't be an immediate priority.

modules/consensus/consensusset_bench_test.go
DavidVorick: 2-5 usually, it's a pretty slow test.

I should be closing them though.

modules/consensus/processed_block.go
DavidVorick: target is only changed every 500 blocks. Otherwise the target is the same as the parent target.

When the target does change, you have to reach back through 500 parents to find your parent (can't use the BlockPath because it might be that 500th grandparent isn't actually in the BlockPath - rare case but still an issue)

profile/timing.go
DavidVorick: good to know

DavidVorick:
InTestSuite I trust.

But really, it's a pretty good test suite. That combined with the fact that it has no issues validating the entire existing blockchain is pretty reassuring. That combined with the fact that we're not important enough yet for someone to actively try to break our system is doubly-reassuring.

modules/consensus/info.go
DavidVorick: just kidding

DavidVorick:
fixed

PR #717, opened 2015-08-16
new go-update API

lukechampine:
The [go-update](http://github.com/inconshreveable/go-update) API has changed. You will need to reinstall the go-update package.

### DIFF
--- a/api/daemon.go
+++ b/api/daemon.go
@@ -129,42 +129,48 @@ func checkForUpdate() (bool, string, error) {
 }
 
 // applyUpdate downloads and applies an update.
-func applyUpdate(version string) (err error) {
+func applyUpdate(version string) error {
 	manifest, err := fetchManifest(version)
 	if err != nil {
-		return
+		return err
 	}
 
 	// Get the executable directory.
 	binDir, err := osext.ExecutableFolder()
 	if err != nil {
-		return
+		return err
 	}
 
-	// Initialize the updater object.
-	up, err := update.New().VerifySignatureWithPEM([]byte(developerKey))
-	if err != nil {
-		// should never happen
-		return
-	}
+	// Configure the update.
+	opts := new(update.Options)
+	opts.SetPublicKeyPEM([]byte(developerKey))
 
 	// Perform updates as indicated by the manifest.
 	for _, file := range manifest[1:] {
+		// set update path
+		opts.TargetPath = filepath.Join(binDir, file)
+
 		// fetch the signature
-		var sig []byte
-		sig, err = getHTTP(version, file+".sig")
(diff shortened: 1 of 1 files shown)


api/daemon.go
```diff
@@ -129,42 +129,48 @@ func checkForUpdate() (bool, string, error) {
...
 
 	// Perform updates as indicated by the manifest.
 	for _, file := range manifest[1:] {
+		// set update path
+		opts.TargetPath = filepath.Join(binDir, file)
+
 		// fetch the signature
-		var sig []byte
-		sig, err = getHTTP(version, file+".sig")
+		opts.Signature, err = getHTTP(version, file+".sig")
```
DavidVorick: an error is returned if the signature fails?

It's not clear from this code.

Do we have a test that tries to apply an update with a bad signature? If not, we should make that a high priority.

lukechampine: `getHTTP` returns a byte slice (response) and an error if the GET failed. Here, the GET is fetching the signature of the file to be downloaded, and storing it in the `opts` struct. This struct is later passed to `update.Apply`, which performs the actual update (including verifying the signature).

There is such a test in daemon_test.go. However, there isn't a valid signature test, since that's more of a pain.

DavidVorick: would it be easier to test if you created a helper function which did the actual verifying and updating (and had no http stuff in it)? Then the important stuff is at least covered.

A second question, do you switch keys every update? That's probably good practice, though I believe we're okay without it for the time being.

lukechampine: no, I'm already spoofing the update server so it's about as easy as it can get. I would just need to create a valid update and store it on the fake server. Updates have worked in the wild, so we know that the signature verification works. It would still be good to have coverage for it, but I just haven't gotten around to it.

The same key is reused. Using a new key for each update is more secure, but it has some practical implications. For example, it prevents you from skipping an update, and it makes it harder for someone to verify that the key is a valid developer key (we would have to list all the keys somewhere).

DavidVorick: > Updates have worked in the wild, so we know that the signature verification works.

What we don't know is that invalid signatures fail. That one is equally important.

> For example, it prevents you from skipping an update

If the updates are diffs, doesn't that stop you from skipping updates anyway?

lukechampine: We *do* know that invalid signatures fail. At least, we know that clearly wrong inputs (see daemon_test.go) fail verification.

Updates are not diffs, they are full binaries.

DavidVorick: oh sorry.

I'm not sure why, but I got the impression that you weren't testing that from our conversation. That's the biggest thing I was worried about.

PR #719, opened 2015-08-16
Wallet polish

DavidVorick:
This took more effort than I thought it would. Nonetheless, this PR is more bark than bite - lots of tests, and some moving things around.

The wallet has switched to an 'unlimited addresses' model, though, if you exceed 2500 addresses for a single seed, you're not going to be able to do seed recovery. I don't think most people will get to that point.

The unlocking process has changed a lot. Now, you have to explicitly make a call to encrypt the wallet, instead of encryption happening on your first call to unlock.

The backup seed files are no longer used by the wallet. Instead, the wallet keeps everything in the settings file. The backup seed files are still created, and given a random suffix so that they don't obliterate existing seed files.

There's now an api call to backup the wallet, but no api call to load backups.

I tried my best, but I'm not super confident that the api spec matches the implementation. There are probably a few small things that slipped through.

I did change all PUT calls to POST calls, simply because I didn't want to deal with making PUT calls in go. I also changed /wallet/address from POST to GET, because you really are getting an address. It just is a lot more intuitive that way.

We're only doing halfway REST at this point. I think that's a good place to be. Our use cases is very different from the traditional REST model and we should break the rules a bit where it makes a  ton of sense.

@Mingling94 this api implementation should fix some of the problems you were having with the UI. If you call /wallet/encrypt without supplying an encryption password (or supplying a blank one), then the seed that gets returned is not only the primary seed but also the encryption password. If you do supply a password during the encrypt process, then that password is what gets used as the encryption password.

Wallet has some key features missing still: 

1. when recovering a seed or loading a file, there is no blockchain rescanning. Instead you'll need to restart the wallet. Crashes might actually happen if you don't. I'm not implementing this quite yet because I don't actually expect anyone to use the recover function in the near future.

2. No support yet for the old wallet files. I almost want to keep it this way too, as the old files are likely to add a lot of bloat. Maybe there can be some form of pruning for addresses that never got a  transaction. That shouldn't be too hard to add on top of rescanning.

3. No way to load backup files that you've saved through the wallet.

4. No way to load keys that you made using siag, which means no siafund support as of this PR. Siafunds are supported for normal addresses, just not the siag multisig addresses.

5. Multisig spending is supported (though loading is not... lol), but it's not tested. The multisig spending needs to be tested.


api/wallet.go
```diff
@@ -106,35 +130,89 @@ func (srv *Server) walletHandler(w http.ResponseWriter, req *http.Request) {
...
+		srv.walletEncryptHandlerPOST(w, req)
+		return
+	}
+	writeError(w, "unrecognized method when calling /wallet/encrypt", http.StatusBadRequest)
 }
 
 // walletHistoryHandlerGET handles a GET request to /wallet/history.
 func (srv *Server) walletHistoryHandlerGET(w http.ResponseWriter, req *http.Request) {
-	start, err := strconv.Atoi(req.FormValue("startHeight"))
+	start, err := strconv.Atoi(req.FormValue("StartHeight"))
```
lukechampine: It looks like a lot of parameters were capitalized. This definitely isn't standard practice. I would go as far as to suggest automatically lower-casing all query string params as they come in.

modules/wallet/encrypt_test.go
```diff
@@ -0,0 +1,272 @@
...
+	"github.com/NebulousLabs/Sia/modules/consensus"
+	"github.com/NebulousLabs/Sia/modules/gateway"
+	"github.com/NebulousLabs/Sia/modules/miner"
+	"github.com/NebulousLabs/Sia/modules/transactionpool"
+	"github.com/NebulousLabs/Sia/types"
+)
+
+// For some of the encryption testing, a wallet that has not been encrypted,
+// unlocked, or been used to mine coins is necessary.
+func createBlankWallet(name string) (string, modules.ConsensusSet, modules.TransactionPool, modules.Miner, *Wallet, error) {
```
lukechampine: The caller is going to have to defer `Close` calls on each of these. This is especially dangerous in tests where you don't assign all the variables (e.g. `_, _, _, _, w, err := createBlankWallet()`). I suggest defining a `blankWallet` object with a `Close` method. This would also spare you the ugliness of 6 return values, lol

modules/wallet/seed.go
```diff
@@ -98,27 +102,14 @@ func (w *Wallet) integrateSeed(seed modules.Seed) {
...
-	return nil
-}
-
 // recoverSeed integrates a recovery seed into the wallet.
 func (w *Wallet) recoverSeed(masterKey crypto.TwofishKey, seed modules.Seed) error {
-	// Integrate the seed with the wallet.
-	w.integrateSeed(seed)
+	// Check that the seed is not already known.
+	for _, wSeed := range w.seeds {
+		if bytes.Equal(seed[:], wSeed[:]) {
```
lukechampine: you should be able to compare these directly, since they are arrays.

modules/wallet/seed_test.go
```diff
@@ -0,0 +1,187 @@
...
+// TestPrimarySeed checks that the correct seed is returned when calling
+// PrimarySeed.
+func TestPrimarySeed(t *testing.T) {
+	if testing.Short() {
+		t.SkipNow()
+	}
+
+	// Create a wallet and fetch the seed at startup.
+	dir := build.TempDir(modules.WalletDir, "TestPrimarySeed")
+	g, err := gateway.New(":0", filepath.Join(dir, modules.GatewayDir))
```
lukechampine: need to close these resources

modules/wallet/wallet_test.go
```diff
@@ -15,6 +16,16 @@ import (
...
 )
 
+var (
+	// Sometimes during testing, you get a statement
+	// `if err != expectedErr { return err }`. But if err == nil, then the
+	// calling function will think that the test has passed, even though the
+	// test failed. [This is a developer error, instead return
+	// errors.New("expecting..")] The new type of error catches when the
+	// developer has made a mistake and failed to catch a test suite failure.
+	errTestCompleted = errors.New("test complete")
```
lukechampine: gonna veto this. Returning nil is a developer error, but so is forgetting to return `errTestCompleted` at the end of a test. In fact, I would say devs would be more prone to make the latter mistake, since it's unintuitive to return an "error" in the event of success.

api/wallet.go
DavidVorick: seems like what we want to do is implement something in writeJSON that will automatically make all parameter names lowercase.

This has to be careful not to touch other items though. {"Name":"David"} should become {"name":"David"}

modules/wallet/encrypt_test.go
DavidVorick: ok

modules/wallet/seed.go
DavidVorick: oh yeah.

whoops

api/wallet.go
lukechampine: even more dangerous would be to lower-case a base-64 encoded string!

modules/wallet/wallet_test.go
DavidVorick: thought you might.

need a better solution then. There needs to be assurance that the test is making it all the way to the end, instead of failing silently.

If the dev uses errTestCompleted at all, it will be apparent that they forgot to use it correctly if they don't use it in both cases. You have to `return errTestCompleted` as well as run `if err != errTestCompleted` - if you do one but not the other:

1. The test fails because an unexpected error was returned, or because `nil` was returned.
2. The test fails silently because it would have failed silently anyway - `nil` was returned accidentally, and the dev forgot to  check `if err != errTestCompleted`.

If they forget to use it at all, then it's equivalent to having just not used it.

Therefore, using `errTestCompleted` is strictly less error-prone than not using `errTestCompleted`.

I recognize that it's unconventional. If you have an alternative proposal for making sure that a test doesn't accidentally return nil when it has actually failed, I'm happy to consider that. Passing the function a `testing.T` object is an unattractive alternative. Having the function call `panic` is also an unattractive alternative. Perhaps there's some construction using defer that will make sure the test completed - 

```go
defer func() {
    if !testFinished && err == nil {
        panic("test is failing silently!")
    }
}
```

If you are strongly opposed to `errTestFinished`, then I think I would be okay putting panics everywhere. I don't like the `testing.T` method of failing because it doesn't help you figure out what conditions caused the test to fail. `panic` on the other hand gives you  a full stack trace, which has the benefit of telling you both what line failed in the helper function and what line failed in the caller function.

Maybe `panic` is the best option.

lukechampine: https://en.wikipedia.org/wiki/Principle_of_least_astonishment

The implicit understanding is: if this function fails, it will return a non-nil error. So we need to be careful that our functions obey this expected behavior. Calling test functions within test functions is dangerous for this reason. If you're expecting a specific error, you must write:
```go
if err != ErrFoo {
    return fmt.Errorf("expected %v, got %v", ErrFoo, err)
}
```
It's a little annoying, but it's also a fact of language. Trying to hide it with fancy wrappers and special variables is an anti-pattern.

As always, I recommend taking a step back and considering: "have I seen anyone else write code like this?" If you haven't, that should give you pause.

DavidVorick: >  If you're expecting a specific error, you must write...

and if you space out and forget to do that, you might put the entire test suite at risk by silently failing to run half of a function. It's doubly bad because 

```go
if err != errBenignErr {
    return err
} 
```
is such a common pattern.

A silently failing test suite is tragic, especially because it's such an easy mistake to make. You get strong, reliable code when you write in ways that make it difficult make mistakes.

I'll switch it to panics. I feel strongly about this.

DavidVorick:
screwed up the rebase and slammed two commits together =/

api/wallet.go
```diff
@@ -106,35 +130,89 @@ func (srv *Server) walletHandler(w http.ResponseWriter, req *http.Request) {
...
+	if req.FormValue("EncryptionPassword") != "" {
+		encryptionKey = crypto.TwofishKey(crypto.HashObject(req.FormValue("EncryptionPassword")))
+	}
+	seed, err := srv.wallet.Encrypt(encryptionKey)
+	if err != nil {
+		writeError(w, "error when calling /wallet/encrypt: "+err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	dictID := mnemonics.DictionaryID(req.FormValue("Dictonary"))
```
triazo: Should be 'Dictionary'

PR #721, opened 2015-08-17
Siac wallet

triazo:
Updates siac to use the new wallet api.

### DIFF
--- a/siac/walletcmd.go
+++ b/siac/walletcmd.go
@@ -8,7 +8,7 @@ import (
 
 	"github.com/spf13/cobra"
 
-	// "github.com/NebulousLabs/Sia/api"
+	"github.com/NebulousLabs/Sia/api"
 	// "github.com/NebulousLabs/Sia/modules"
 )
 
@@ -68,13 +68,6 @@ The smallest unit of siacoins is the hasting. One siacoin is 10^24 hastings. Oth
 		Run:   wrap(walletaddresscmd),
 	}
 
-	walletMergeCmd = &cobra.Command{
-		Use:   "merge [walletfile]",
-		Short: "Merge wallet",
-		Long:  "Merge another wallet with the existing wallet. The existing wallet will inherit all keys and addresses.",
-		Run:   wrap(walletmergecmd),
-	}
-
 	walletSendCmd = &cobra.Command{
 		Use:   "send [amount] [dest]",
 		Short: "Send coins to another wallet",
@@ -116,6 +109,28 @@ Run 'wallet send --help' to see a list of available units.`,
 		Long:  "View wallet status, including the current balance and number of addresses.",
 		Run:   wrap(walletstatuscmd),
 	}
+
+	walletUnlockCmd = &cobra.Command{
+		Use:   "unlock \"[password]\"",
+		Short: "Unlock the wallet",
+		Long:  "Decrypt and load the wallet into memory",
+		Run:   wrap(walletunlockcmd),
+	}
+
+	walletEncryptCmd = &cobra.Command{
+		Use:   "encrypt",
+		Short: "Initilize and encrypt the wallet",
+		Long: `Generate primary seed used for the wallet and encrypt using that
+The encryption password and primary seed will be output`,
+		Run: wrap(walletencryptcmd),
+	}
+
+	walletLockCmd = &cobra.Command{
+		Use:   "lock",
+		Short: "Lock the wallet",
+		Long:  "Lock the wallet, preventing further use",
+		Run:   wrap(walletlockcmd),
+	}
 )
 
 // TODO: this should be defined outside of siac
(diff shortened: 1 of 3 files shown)


siac/walletcmd.go
```diff
@@ -116,6 +109,42 @@ Run 'wallet send --help' to see a list of available units.`,
 		Long:  "View wallet status, including the current balance and number of addresses.",
 		Run:   wrap(walletstatuscmd),
 	}
+
+	walletUnlockCmd = &cobra.Command{
+		Use:   "unlock \"[password]\"",
```
lukechampine: prefer \` when you need to use `"` in a string

siac/walletcmd.go
```diff
@@ -116,6 +109,42 @@ Run 'wallet send --help' to see a list of available units.`,
...
+		Short: "Unlock the wallet",
+		Long:  "Decrypt and load the wallet into memory",
+		Run:   wrap(walletunlockcmd),
+	}
+
+	walletInitCmd = &cobra.Command{
+		Use:   "init",
+		Short: "Initilize and encrypt a new wallet",
+		Long: `Generate primary seed used for the wallet and encrypt using that
```
lukechampine: wording is a bit awkward. I would write:
```
Generate a new wallet from a seed string, and encrypt it.
The seed string and encryption password will be returned.
```

also, the above should be "Initialize"

siac/walletcmd.go
```diff
@@ -206,22 +238,86 @@ func walletsiafundssendcmd(cmd *cobra.Command, args []string) {
...
+	err := post("/wallet/unlock", qs)
+	if err != nil {
+		fmt.Println("Could not unlock wallet:", err)
+		return
+	}
+	fmt.Println("Wallet unlocked")
+}
+
+// walletencryptcmd encrypts the wallet with the given password
+func walletinitcmd() {
```
lukechampine: wrong docstring

siac/walletcmd.go
triazo: Sounds pretty good. Might this imply that you are giving it the seed though?

siac/walletcmd.go
```diff
@@ -206,22 +238,86 @@ func walletsiafundssendcmd(cmd *cobra.Command, args []string) {
...
+	var er api.WalletEncryptPOST
+	qs := fmt.Sprintf("dictionary=%s", "english")
+	err := postResp("/wallet/encrypt", qs, &er)
+	if err != nil {
+		fmt.Println("Error when encrypting wallet:", err)
+		return
+	}
+	fmt.Printf("Encrypted wallet with primary seed: %s\n", er.PrimarySeed)
```
lukechampine: help dialog says the encryption password is returned too?

triazo: help dialog has confusing wording - they're the same thing.

siac/walletcmd.go
lukechampine: meh, maybe a little bit. The fact that it returns the seed, and that the command takes no arguments, should enough of a hint.

siac/walletcmd.go
DavidVorick: They are only the same thing if you don't supply a separate password when encrypting the wallet.

The output should specify both.

"Wallet keys will be generated with a seed. You can use the seed to recover the pregenerated addresses if you lose your wallet or encryption password. Seed is:\n %s\n\nWallet encrypted with password: %s"

lukechampine: should init take an argument then? I think it would be easier to just always use the seed as the password. Although that means your password is pretty long...

siac/walletcmd.go
```diff
@@ -206,22 +238,94 @@ func walletsiafundssendcmd(cmd *cobra.Command, args []string) {
...
+}
+
+// walletinitcmd encrypts the wallet with the given password
+func walletinitcmd() {
+	var er api.WalletEncryptPOST
+	qs := fmt.Sprintf("dictionary=%s", "english")
+	if initPassword != "" {
+		qs += fmt.Sprintf("&encryptpassword=%s", initPassword)
+	}
+	defer func() { initPassword = "" }()
```
lukechampine: not necessary; siac commands are one-off, so nothing persists across commands anyway.

DavidVorick: that doesn't mean that the data doesn't stick around.

But, in go, strings are immutable which means you can't wipe them anyway. If we wanted to be properly careful about wiping any data that got passed to the program, we'd need to not use strings.

siac/walletcmd.go
DavidVorick: when protecting money it makes sense to have a cryptographically secure password. The 15 word password that we use is accepted as the state-of-the-art in wallet encryption among the altcoin community. Idk how the Bitcoin devs feel about it, they haven't implemented it. They are opposed to brain wallets usually, but our passwords are guaranteed to be cryptographically secure.

siac/walletcmd.go
triazo: Also, the terminal will record the command in its history. Somebody taking a .histfile is far more likely to happen than a malicious heap allocation.

PR #722, opened 2015-08-18
Transaction Pool Merging

DavidVorick:
Transaction sets now get merged by the transaction pool where appropriate.

Some of the code is wonky, but it does work. I plan on cleaning it up.

There are also some api things that I discovered while testing the UI, added those fixes here as well.

### DIFF
--- a/modules/transactionpool/accept.go
+++ b/modules/transactionpool/accept.go
@@ -2,6 +2,7 @@ package transactionpool
 
 import (
 	"errors"
+	"time"
 
 	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/encoding"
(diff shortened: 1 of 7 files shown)


modules/transactionpool/accept.go
```diff
@@ -29,10 +30,45 @@ var (
 	errObjectConflict      = errors.New("transaction set conflicts with an existing transaction set")
 	errFullTransactionPool = errors.New("transaction pool cannot accept more transactions")
 	errLowMinerFees        = errors.New("transaction set needs more miner fees to be accepted")
+	errEmptySet            = errors.New("transaction set is empty")
 
 	TransactionMinFee = types.NewCurrency64(2).Mul(types.SiacoinPrecision)
 )
 
+// relatedObjectIDs determins all of the object ids related to a transction.
```
lukechampine: two typos

PR #724, opened 2015-08-18
Prompt user for passwords in siac

triazo:
Also adds unconfirmed balance to status

### DIFF
--- a/siac/main.go
+++ b/siac/main.go
@@ -19,7 +19,7 @@ import (
 var (
 	addr         string
 	force        bool
-	initPassword string
+	initPassword bool
 )
 
 // apiGet wraps a GET request with a status code check, such that if the GET does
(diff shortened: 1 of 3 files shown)


lukechampine:
looks good, although there's no escaping the fact that we're sending the password in plaintext over HTTP. We could switch to HTTPS, but that would only be delaying the inevitable.

siac/main.go
```diff
@@ -179,7 +179,7 @@ func main() {
 
 	root.AddCommand(walletCmd)
 	walletCmd.AddCommand(walletAddressCmd, walletSendCmd, walletSiafundsCmd, walletStatusCmd, walletUnlockCmd, walletLockCmd, walletInitCmd, walletAddseedCmd, walletSeedsCmd)
-	walletInitCmd.Flags().StringVarP(&initPassword, "password", "p", "", "Encryption password used for wallet")
+	walletInitCmd.Flags().BoolVarP(&initPassword, "password", "p", false, "Encryption password used for wallet")
```
lukechampine: need to change help text

triazo:
I'll be updating documentation and including a readme in a seperate pull request on top of this.

siac/main.go
DavidVorick: you only need to supply a password in 3 commands (encrypt, unlock, recover). Encrypt I think by default should just use a blank password, and the others are required to use a password.

So the flag is probably not needed at all.

DavidVorick:
I've merged this PR because I'm still uncertain about how the password situation should be handled. It might be more clear after some use.

Issue #723, opened 2015-08-18
Host + Renter API


lukechampine:
Here's my proposal:

Host
----

Queries:

* /host                 [GET]
* /host/announce        [POST]
* /host/settings        [GET]
* /host/settings        [POST]

#### /host [GET]

Function: Returns basic information about the host, including the number of
file contracts, the amount of storage remaining, and the host's profit to date.

Parameters: none

Response:
```
struct {
	storageremaining int64
	numcontracts     int
	profit           int
	potentialprofit  int
	competition      int
	unlockhash       string
	ipaddress        string
}
```

#### /host/announce [POST]

Function: Announces the host to the network as a source of storage. Generally
only needs to be called once.

Parameters: none

Response: standard

#### /host/settings [GET]

Function: Returns the current host configuration.

Parameters: none

Response:
```
struct {
	totalstorage int
	minfilesize  int
	maxfilesize  int
	minduration  int
	maxduration  int
	windowsize   int
	price        int
	collateral   int
}
```

`totalstorage` is how much storage (in bytes) the host will rent to the
network.

`minfilesize` is the minimum allowed file size.

`maxfilesize` is the maximum allowed file size.

`minduration` is the minimum amount of time a contract is allowed to last.

`maxduration` is the maximum amount of time a contract is allowed to last.

`windowsize` is the number of blocks a host has to prove they are holding the
file.

`price` is the cost (in hastings per byte) of data stored.

`collateral` is the amount of collateral the host will offer (in hastings per
byte per block) for losing files on the network.

#### /host/settings [POST]

Function: Sets the host configuration.

Parameters:
```
totalstorage int
minfilesize  int
maxfilesize  int
minduration  int
maxduration  int
windowsize   int
price        int
collateral   int
```

See the corresponding GET call for a description of each parameter.

Response: standard

----
----

The biggest change here is that `settings` is now symmetrical. Also, the old `configure` call would only modify supplied parameters; in the new scheme, all parameters are mandatory. This means that if you want to modify one field, you have to GET the old values and then POST the full updated version.

I also suggest two new endpoints, to be added some time in the future:

* /host/contracts       [GET]
* /host/contracts/$(id) [GET]

`contracts` would return a list of ids, which could be used to fetch a specific contract. (Alternatively, all contracts could be returned in one call.) There isn't much use for this right now, but I think eventually we'll want access to this information.

DavidVorick:
[...]

PR #726, opened 2015-08-19
Host + Renter overhaul

### LINKED ISSUE #159: Erasure Coding
Need to be able to split files up and store them in multiple places, and perhaps have some sort of file health indicator or whatever. Files need to go up and be healthy.

I don't currently know if the regenerating codes are finished/ready, so we should probably just use the longhair library again (if we can manage cross-compilation) or some slower library that has been written in go.

It's not a big deal to require the user to have the file to do repairs, especially since it's only a temporary requirement thanks to regenerating codes.

### LINKED ISSUE #642: 4GB+ file limit using file contract revisions
This is related to #616 

When uploading, files should be uploaded 20mb at a time to a select set of hosts.  The whole process should be done using file contract revisions.

The crypto here is fairly finicky, we'll have to work through the design together and be careful about it.

lukechampine:
At long last, we have erasure coding, parallel uploads and downloads, and file contract revisions. But I'm not terribly proud of some of this code; a lot of it is a bit shaky, and needs thorough review (especially contract negotiation + revision -- I'm pretty confident I'm not doing signatures correctly, because I can't figure out where to supply the secret keys).

Definitely missing from this PR:
- Host + renter API updates, as described in #723. This includes adding json tags.
- Compatibility is broken for .sia files, as well as host+renter persistence files and host RPCs. At the very least, hosts should be able to provide old files and submit old proofs, and renters should be able to retrieve old files. This may be tricky and/or ugly. We should find out what users are willing to put up with.
- ~~Revisions are never submitted to the blockchain. Both the host and renter need to submit the revision before WindowStart, and or immediately in the event of negotiation failure.~~ (works for now, but will need modification when resumability is introduced)
- ~~The host and renter probably need to save some information (to disk) that they currently aren't. I'm fairly sure the renter needs to save the unlock conditions for each file contract, but there may be others.~~ (Will be implemented later, to enable resumable uploads)
- Revisions need miner fees. (not immediately)
- Revisions do not account for changes in host settings
- Lots of testing. Current coverage is renter: 43.9%, host: 17.2% (ouch!)

Finally, the upload and download algorithms need more work. The upload algorithm is (I believe) close to optimal in terms of speed, but may impact redundancy. The download algorithm is still terrible; I haven't come up with anything satisfactory yet. To make things easier, I've included commented-out statistics in `TestErasureUpload` and `TestErasureDownload`, so you can get immediate feedback on any modifications you make to the upload and download algorithms. Also, the `testHost` object should be modified to disconnect (either randomly or at a specific time) so that we can ensure such disconnects won't break uploads or downloads.

I think that's it. We're almost there, fellas. Just needs some polishing and testing, and then the real fun begins...

fixes #159 and fixes #642.

### DIFF
--- a/api/renter.go
+++ b/api/renter.go
@@ -5,12 +5,15 @@ import (
 	"time"
 
 	"github.com/NebulousLabs/Sia/modules"
+	"github.com/NebulousLabs/Sia/modules/renter"
 	"github.com/NebulousLabs/Sia/types"
 )
 
 const (
-	duration   = 6000 // Duration that hosts will hold onto the file.
-	redundancy = 15   // Redundancy of files uploaded to the network.
+	duration     = 6000    // Duration that hosts will hold onto the file.
+	dataPieces   = 2       // Default data pieces per erasure-coded chunk
+	parityPieces = 10      // Default parity pieces per erasure-coded chunk
+	pieceSize    = 1 << 22 // Default size of one piece (4 MiB)
 )
 
 // DownloadInfo is a helper struct for the downloadqueue API call.
@@ -119,7 +122,7 @@ func (srv *Server) renterFilesLoadHandler(w http.ResponseWriter, req *http.Reque
 }
 
 // renterFilesLoadAsciiHandler handles the API call to load a '.sia' file
-// in ascii form.
+// in ASCII form.
 func (srv *Server) renterFilesLoadAsciiHandler(w http.ResponseWriter, req *http.Request) {
 	files, err := srv.renter.LoadSharedFilesAscii(req.FormValue("file"))
 	if err != nil {
@@ -161,11 +164,13 @@ func (srv *Server) renterStatusHandler(w http.ResponseWriter, req *http.Request)
 
 // renterFilesUploadHandler handles the API call to upload a file.
 func (srv *Server) renterFilesUploadHandler(w http.ResponseWriter, req *http.Request) {
+	ecc, _ := renter.NewRSCode(dataPieces, parityPieces)
 	err := srv.renter.Upload(modules.FileUploadParams{
-		Filename: req.FormValue("source"),
-		Duration: duration,
-		Nickname: req.FormValue("nickname"),
-		Pieces:   redundancy,
+		Filename:  req.FormValue("source"),
+		Duration:  duration,
+		Nickname:  req.FormValue("nickname"),
+		ECC:       ecc,
+		PieceSize: pieceSize,
 	})
 	if err != nil {
 		writeError(w, "Upload failed: "+err.Error(), http.StatusInternalServerError)
--- a/api/renter.go
+++ b/api/renter.go
@@ -75,7 +75,7 @@ func (srv *Server) renterFilesListHandler(w http.ResponseWriter, req *http.Reque
 			UploadProgress: file.UploadProgress(),
 			Nickname:       file.Nickname(),
 			Filesize:       file.Filesize(),
-			TimeRemaining:  file.TimeRemaining(),
+			TimeRemaining:  file.Expiration() - types.BlockHeight(srv.blockchainHeight),
 		})
 	}
 
--- a/api/renter.go
+++ b/api/renter.go
@@ -28,7 +28,6 @@ type FileInfo struct {
 	UploadProgress float32
 	Nickname       string
 	Filesize       uint64
-	Repairing      bool
 	TimeRemaining  types.BlockHeight
 }
 
@@ -76,7 +75,6 @@ func (srv *Server) renterFilesListHandler(w http.ResponseWriter, req *http.Reque
 			UploadProgress: file.UploadProgress(),
 			Nickname:       file.Nickname(),
 			Filesize:       file.Filesize(),
-			Repairing:      file.Repairing(),
 			TimeRemaining:  file.TimeRemaining(),
 		})
 	}
--- a/api/renter.go
+++ b/api/renter.go
@@ -164,13 +164,13 @@ func (srv *Server) renterStatusHandler(w http.ResponseWriter, req *http.Request)
 
 // renterFilesUploadHandler handles the API call to upload a file.
 func (srv *Server) renterFilesUploadHandler(w http.ResponseWriter, req *http.Request) {
-	ecc, _ := renter.NewRSCode(dataPieces, parityPieces)
+	rsc, _ := renter.NewRSCode(dataPieces, parityPieces)
 	err := srv.renter.Upload(modules.FileUploadParams{
-		Filename:  req.FormValue("source"),
-		Duration:  duration,
-		Nickname:  req.FormValue("nickname"),
-		ECC:       ecc,
-		PieceSize: pieceSize,
+		Filename:    req.FormValue("source"),
+		Duration:    duration,
+		Nickname:    req.FormValue("nickname"),
+		ErasureCode: rsc,
+		PieceSize:   pieceSize,
 	})
 	if err != nil {
 		writeError(w, "Upload failed: "+err.Error(), http.StatusInternalServerError)
--- a/api/renter.go
+++ b/api/renter.go
@@ -100,13 +100,17 @@ func (srv *Server) renterFilesDeleteHandler(w http.ResponseWriter, req *http.Req
 // renterFilesRenameHandler handles the API call to rename a file entry in the
 // renter.
 func (srv *Server) renterFilesRenameHandler(w http.ResponseWriter, req *http.Request) {
-	err := srv.renter.RenameFile(req.FormValue("nickname"), req.FormValue("newname"))
-	if err != nil {
-		writeError(w, err.Error(), http.StatusBadRequest)
-		return
-	}
+	writeError(w, "renaming temporarily disabled", http.StatusBadRequest)
 
-	writeSuccess(w)
+	/*
+		err := srv.renter.RenameFile(req.FormValue("nickname"), req.FormValue("newname"))
+		if err != nil {
+			writeError(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+
+		writeSuccess(w)
+	*/
 }
 
 // renterFilesLoadHandler handles the API call to load a '.sia' file.
--- a/api/renter.go
+++ b/api/renter.go
@@ -134,6 +134,7 @@ func (srv *Server) renterFilesLoadAsciiHandler(w http.ResponseWriter, req *http.
 
 // renterFilesShareHandler handles the API call to create a '.sia' file that
 // shares a file.
+// TODO: allow sharing of multiple files.
 func (srv *Server) renterFilesShareHandler(w http.ResponseWriter, req *http.Request) {
 	err := srv.renter.ShareFiles([]string{req.FormValue("nickname")}, req.FormValue("filepath"))
 	if err != nil {
--- a/api/renter.go
+++ b/api/renter.go
@@ -16,7 +16,6 @@ const (
 // DownloadInfo is a helper struct for the downloadqueue API call.
 type DownloadInfo struct {
 	StartTime   time.Time
-	Complete    bool
 	Filesize    uint64
 	Received    uint64
 	Destination string
@@ -57,7 +56,6 @@ func (srv *Server) renterDownloadqueueHandler(w http.ResponseWriter, req *http.R
 	for _, dl := range downloads {
 		downloadSet = append(downloadSet, DownloadInfo{
 			StartTime:   dl.StartTime(),
-			Complete:    dl.Complete(),
 			Filesize:    dl.Filesize(),
 			Received:    dl.Received(),
 			Destination: dl.Destination(),
--- a/api/renter.go
+++ b/api/renter.go
@@ -109,10 +109,9 @@ func (srv *Server) renterFilesRenameHandler(w http.ResponseWriter, req *http.Req
 	writeSuccess(w)
 }
 
-// renterFilesLoadHandler handles the API call to load a '.sia' that
-// contains filesharing information.
+// renterFilesLoadHandler handles the API call to load a '.sia' file.
 func (srv *Server) renterFilesLoadHandler(w http.ResponseWriter, req *http.Request) {
-	files, err := srv.renter.LoadSharedFile(req.FormValue("filename"))
+	files, err := srv.renter.LoadSharedFiles(req.FormValue("filename"))
 	if err != nil {
 		writeError(w, err.Error(), http.StatusBadRequest)
 		return
--- a/encoding/marshal.go
+++ b/encoding/marshal.go
@@ -17,10 +17,9 @@ type SiaMarshaler interface {
 	MarshalSia() []byte
 }
 
-// A SiaUnmarshaler can decode itself from a byte slice. If a decoding error
-// occurs, UnmarshalSia should panic.
+// A SiaUnmarshaler can decode itself from a byte slice.
 type SiaUnmarshaler interface {
-	UnmarshalSia([]byte)
+	UnmarshalSia([]byte) error
 }
 
 // An Encoder writes objects to an output stream.
@@ -222,7 +221,10 @@ func (d *Decoder) decode(val reflect.Value) {
 	// check for UnmarshalSia interface first
 	if val.CanAddr() {
 		if u, ok := val.Addr().Interface().(SiaUnmarshaler); ok {
-			u.UnmarshalSia(d.readPrefix())
+			err := u.UnmarshalSia(d.readPrefix())
+			if err != nil {
+				panic(err)
+			}
 			return
 		}
 	}
--- a/modules/host.go
+++ b/modules/host.go
@@ -5,17 +5,18 @@ import (
 )
 
 const (
-	AcceptTermsResponse = "accept"
-	HostDir             = "host"
+	AcceptResponse = "accept"
+	HostDir        = "host"
 )
 
 // RPC identifiers
 var (
-	RPCSettings = [8]byte{'S', 'e', 't', 't', 'i', 'n', 'g', 's'}
-	RPCContract = [8]byte{'C', 'o', 'n', 't', 'r', 'a', 'c', 't'}
-	RPCDownload = [8]byte{'D', 'o', 'w', 'n', 'l', 'o', 'a', 'd'}
-	// deprecated
-	RPCRetrieve = [8]byte{'R', 'e', 't', 'r', 'i', 'e', 'v', 'e'}
+	// Each identifier has a version number at the end, which will be
+	// incremented whenever the protocol changes.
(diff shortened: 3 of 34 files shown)


Makefile
```diff
@@ -15,6 +15,7 @@ dependencies:
 	go get -u github.com/inconshreveable/go-update
 	go get -u github.com/inconshreveable/muxado
 	go get -u github.com/kardianos/osext
+	go get -u github.com/klauspost/reedsolomon
```
DavidVorick: dependencies are a sin

DavidVorick: jk just wanted to start out strong

api/renter.go
```diff
@@ -5,18 +5,20 @@ import (
...
 	"github.com/NebulousLabs/Sia/types"
 )
 
 const (
-	duration   = 6000 // Duration that hosts will hold onto the file.
-	redundancy = 15   // Redundancy of files uploaded to the network.
+	duration     = 6000    // Duration that hosts will hold onto the file.
+	dataPieces   = 2       // Default data pieces per erasure-coded chunk
+	parityPieces = 10      // Default parity pieces per erasure-coded chunk
+	pieceSize    = 1 << 22 // Default size of one piece (4 MiB)
```
DavidVorick: I'm not sure the api is the right place for these.

api/renter.go
```diff
@@ -29,7 +31,6 @@ type FileInfo struct {
 	UploadProgress float32
 	Nickname       string
 	Filesize       uint64
-	Repairing      bool
```
DavidVorick: what is the reason for removing the repairing variable?

api/renter.go
```diff
@@ -29,7 +31,6 @@ type FileInfo struct {
 	UploadProgress float32
```
DavidVorick: why a float32?

modules/host.go
```diff
@@ -5,21 +5,25 @@ import (
...
-	WindowSize         types.BlockHeight     // How long the host has to submit a proof of storage.
-	Price              types.Currency        // Client contribution towards payout each window
-	Collateral         types.Currency        // Host contribution towards payout each window
-	ValidProofOutputs  []types.SiacoinOutput // Where money goes if the storage proof is successful.
-	MissedProofOutputs []types.SiacoinOutput // Where the money goes if the storage proof fails.
+// RPC identifiers
+var (
+	// Each identifier has a version number at the end, which will be
+	// incremented whenever the protocol changes.
+	RPCSettings = types.Specifier{'S', 'e', 't', 't', 'i', 'n', 'g', 's', 0}
```
DavidVorick: huh. mind explaining this?

modules/host/host.go
```diff
@@ -95,10 +102,21 @@ func New(cs *consensus.ConsensusSet, hdb modules.HostDB, tpool modules.Transacti
...
 		obligationsByHeight: make(map[types.BlockHeight][]contractObligation),
 
-		mu: sync.New(modules.SafeMutexDelay, 1),
+		mu: safesync.New(modules.SafeMutexDelay, 1),
 	}
 	h.spaceRemaining = h.TotalStorage
 
+	// Generate signing key, for revising contracts.
+	// TODO: what do we do with the secret key??
+	_, pk, err := crypto.GenerateSignatureKeys()
```
DavidVorick: uhhhhh.... you keep it?

DavidVorick: keep it safe until it's time to sign the transaction

modules/host/negotiate.go
```diff
@@ -15,280 +15,356 @@ import (
...
+		if err := signedTxn.StandaloneValid(height); err != nil {
+			return err
+		}
+
+		// sign and return transaction
+		txnBuilder := h.wallet.RegisterTransaction(signedTxn, nil)
+		for _, sig := range signedTxn.TransactionSignatures {
+			txnBuilder.AddTransactionSignature(sig)
+		}
+		signedTxnSet, err := txnBuilder.Sign(true)
```
DavidVorick: this won't work.

The txn builder can only sign inputs it funded itself. Since you didn't use FundTransaction, this is basically a no-op.

To sign the transcation you're going to need the secret key you made earlier. The process is a bit convoluted because you sign everything except the signature itself, which means you have to craft the signature items. As a reference, checkout transactionbuilder.Sign.

Here's a somewhat condensed version of the code you'll need:

```
    coveredFields := types.CoveredFields{WholeTransaction: true}

    // TransactionSignatures don't get covered by the 'WholeTransaction' flag,
    // and must be covered manually.
    for i := range txn.TransactionSignatures {
        coveredFields.TransactionSignatures = append(coveredFields.TransactionSignatures, uint64(i))
    }

         // Create the unsigned transaction signature.
        sig := types.TransactionSignature{
            ParentID:       parentID,
            CoveredFields:  coveredFields,
            PublicKeyIndex: 1, // depends on if the host key is 1st or second, you'll have to figure that out
        }
        txn.TransactionSignatures = append(txn.TransactionSignatures, sig) // txn not signed yet, just prepared

        // Get the signature.
        sigIndex := len(txn.TransactionSignatures) - 1
        sigHash := txn.SigHash(sigIndex) // this will figure out what hash is supposed to be signed using coveredFields and such
        encodedSig, err := crypto.SignHash(sigHash,  secretKey[i]) // actual sig
        if err != nil {
            return err
        }
        txn.TransactionSignatures[sigIndex].Signature = encodedSig[:]
    }


```

DavidVorick: There's probably a good example somewhere in the types test suite as well

modules/renter.go
```diff
@@ -10,13 +11,37 @@ var (
 	RenterDir = "renter"
 )
 
+// An ECC is an error-correcting code.
+// TODO: rename? (can be confused with Elliptic Curve Crypto)
```
DavidVorick: rename would be preferred

api/renter.go
lukechampine: Well, I think we want them to be settable at some point, which means they need to be passed as parameters to Upload.

I just realized that `duration` is redefined in `renter/negotiate.go`. Whoops. It should be passed in to `newHostUploader` as an argument.

api/renter.go
lukechampine: Well, we're not using it now and we never did. We can add it back when we start doing repairing, but that may not happen for a while.

api/renter.go
lukechampine: as opposed to...? Precision isn't crucial here, but making it a float64 wouldn't hurt.

modules/host.go
lukechampine: Otherwise, when you change the protocol, you have to pick a new identifier. You can't (easily) reuse the old one, because you won't know which protocol your peer wants to speak.

lukechampine: If we want to remain compatible with v0.3.3.3, for the time being we'll have to check for the old IDs as well. This would look something like:
```go
var rpcID [8]byte
encoding.ReadObject(conn, &rpcID, 8)
if fn, ok := oldRPCmap[rpcID]; ok {
    fn(conn)
    return
}
// otherwise, read another 8 bytes and convert to a types.Specifier
var rpcSpec types.Specifier
copy(rpcSpec[:], rpcID)
io.ReadFull(conn, rpcSpec[8:])
if fn, ok := newRPCmap[rpcSpec]; ok {
   fn(conn)
   return
}
log("unknown RPC")
```

modules/host/negotiate.go
lukechampine: yeesh. Could we add a method to the txnBuilder that signs using a supplied key? I don't understand the process well enough to judge whether that idea makes sense.

modules/renter/negotiate.go
```diff
@@ -2,233 +2,328 @@ package renter
...
+	key := deriveKey(hu.masterKey, p.chunkIndex, p.pieceIndex)
+	encPiece, err := key.EncryptBytes(p.data)
 	if err != nil {
 		return err
 	}
-	for i := len(unsignedTxn.SiacoinInputs); i < len(collateralTxn.SiacoinInputs); i++ {
-		txnBuilder.AddSiacoinInput(collateralTxn.SiacoinInputs[i])
+
+	// calculate new merkle root
+	hu.tree.Push(encPiece) // TODO: WRONG!
```
lukechampine: whoops, missed this one. We actually need to push in 64-byte chunks.

modules/renter/negotiate.go
```diff
@@ -2,233 +2,328 @@ package renter
...
 
-const (
-	defaultWindowSize = 288 // 48 Hours
-)
+// A hostUploader uploads pieces to a host. It implements the uploader interface.
+type hostUploader struct {
+	// constants
+	settings         modules.HostSettings
+	masterKey        crypto.TwofishKey
+	unlockConditions types.UnlockConditions // renter needs to save this!
```
lukechampine: is this right? If so, the UnlockConditions include both public keys, so that's convenient. At the moment, there's no support for revising an "old" contract (you only revise during upload; no resuming). Even though we don't need to save it yet, we may want to in order to avoid breaking compatibility later. (Although that would also mean saving all the fields necessary for resumable uploads/downloads, which may be a bit trickier...)

api/renter.go
DavidVorick: I guess I'm just surprised it's not an int. Like bytes or segments vs. full size.

api/renter.go
DavidVorick: ok, that makes sense I guess.

api/renter.go
DavidVorick: I'm pretty sure it was in use for 0.3.3.3. I definitely wrote and rewrote some repairing code at some point. I also remember it never worked very well.

modules/host.go
DavidVorick: To prevent splitting the network, it's absolutely crucial that things like blocks and transactions be able to move from the 0.3.3.3 nodes to the 0.4.0 nodes and vice-versa.

I'm not sure which calls are affected or if you've already preserved compatibility, but I do consider it a requirement as far as transactions and blocks are concerned. Other calls are less important.

modules/host/negotiate.go
DavidVorick: the transaction builder would also need some way to figure out which key it's supposed to be signing.

Probably best if I write the signing code. The process may look daunting, but I think it's ultimately pretty simple:

1. use the CoveredFields object to indicate what parts of the transaction should get covered by the signature.
2. create the TransactionSignature object to indicate which object + public key the signature is signing (and also supply the covered fields from 1).
3. call txn.SigHash to get the object you are supposed to sign.
4. call crypto.SignHash to create the cryptographic signature
5. add the cryptographic signature to the transaction.

While it is a lot of steps, the process is pretty straigtforward I think.

DavidVorick:
"fix test errors"

:P

modules/renter/negotiate.go
DavidVorick: This looks right to me. You need to be certain which public key is which, but if you formally define that in the protocol spec (IE renter is always [0] and host is always [1]), then problem solved.

api/renter.go
lukechampine: The important difference is that `UploadProgress` indicates the *total* progress, including redundancy and encryption overhead. So a `NumBytesUploaded` field wouldn't be very helpful. Even if `NumBytesUploaded > Filesize`, the file may not be available yet.

modules/host.go
lukechampine: Blocks and transactions are both handled by the gateway, which hasn't been changed since v0.3.3.3. So they should continue to propagate normally, as long as the consensus set and transaction pool are compatible as well.

modules/host/negotiate.go
lukechampine: yeah, I was thinking you might want to jump in and push a commit to this branch. You're much more likely to get it right.

lukechampine:
really weird -- Travis says the renter "build failed," but there's no output indicating why.

lukechampine:
swer 2 god Travis I will wrek u m8

DavidVorick:
83 commits

modules/host.go
DavidVorick: I still don't understand the trailing 0

modules/host/host.go
```diff
@@ -4,10 +4,12 @@ import (
 	"errors"
 	"net"
 	"os"
+	"sync"
 
+	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/modules/consensus"
-	"github.com/NebulousLabs/Sia/sync"
+	safesync "github.com/NebulousLabs/Sia/sync"
```
DavidVorick: should this be moved to its own grouping since it's ruining the party?

modules/host/upload.go
```diff
@@ -50,3 +51,51 @@ func (h *Host) rpcRetrieve(conn net.Conn) error {
...
+	// Open the file.
+	file, err := os.Open(co.Path)
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+
+	// Process requests until 'stop' signal is received.
+	var request modules.DownloadRequest
+	for {
```
DavidVorick: I generally think it's better practice to avoid this type of loop

modules/renter.go
```diff
@@ -10,13 +11,37 @@ var (
...
+type ECC interface {
```
DavidVorick: rename pls

DavidVorick:
Definitely missing from this PR:
- Host + renter API updates, as described in #723. This includes adding json tags.
- Compatibility is broken for .sia files, as well as host+renter persistence files and host RPCs. At the very least, hosts should be able to provide old files and submit old proofs, and renters should be able to retrieve old files. This may be tricky and/or ugly. We should find out what users are willing to put up with.
- Revisions are never submitted to the blockchain. Both the host and renter need to submit the revision before WindowStart, and or immediately in the event of negotiation failure (maybe?).
- The host and renter probably need to save some information (to disk) that they currently aren't. I'm fairly sure the renter needs to save the unlock conditions for each file contract, but there may be others.
- Revisions need miner fees. (not immediately)
- Revisions do not account for changes in host settings
- Lots of testing. Current coverage is renter: 43.9%, host: 17.2% (ouch!)
---

What of the above is still true?

```
=== RUN TestIntegrationWalletGETSiacoins-2

A lock was held for too long, id '13'. Call stack:

	File: '/home/travis/gopath/src/github.com/NebulousLabs/Sia/modules/consensus/accept.go:198'

	File: '/home/travis/gopath/src/github.com/NebulousLabs/Sia/modules/miner/cpuminer.go:71'

A lock was held until deadlock, subsequent call to unlock failed. id '13'. Call stack:

	File: '/home/travis/gopath/src/github.com/NebulousLabs/Sia/modules/consensus/accept.go:210'

	File: '/home/travis/gopath/src/github.com/NebulousLabs/Sia/modules/miner/cpuminer.go:71'
```

You may have introduced a deadlock

DavidVorick:
+ Revisions are never submitted to the blockchain. Both the host and renter need to submit the revision before WindowStart, and or immediately in the event of negotiation failure (maybe?).

+ The host and renter probably need to save some information (to disk) that they currently aren't. I'm fairly sure the renter needs to save the unlock conditions for each file contract, but there may be others

These two are the big ones. The last revision needs to be submitted to the blockchain by both parties. Relatively simple because you are doing it in just one file. If you get this far, you don't actually need to save anything to disk. Otherwise, you need some way to resume the upload, or you need to just accept that if the renter goes offline in the middle of an upload, that file and the associated balance are going to be lost (not such a bad thing - it's bad but minimally so. Files are at most 5GB right now). You also need to make sure that if negotiation fails anywhere (IE the host runs off halfway through uploading, or some other failure occurs), the most recent completed file contract revision needs to be submitted to the blockchain.

crypto/encrypt_test.go
```diff
@@ -44,7 +44,7 @@ func TestTwofishEncryption(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if bytes.Compare(plaintext, decryptedPlaintext) != 0 {
+	if !bytes.Equal(plaintext, decryptedPlaintext) {
 		t.Fatal("Encrypted and decrypted zero plaintext do not match")
```
DavidVorick: the wallet has similar code. I'm wondering if this shouldn't just be moved to the crypto package.

encoding/marshal.go
```diff
@@ -17,10 +17,9 @@ type SiaMarshaler interface {
 	MarshalSia() []byte
 }
 
-// A SiaUnmarshaler can decode itself from a byte slice. If a decoding error
-// occurs, UnmarshalSia should panic.
+// A SiaUnmarshaler can decode itself from a byte slice.
 type SiaUnmarshaler interface {
-	UnmarshalSia([]byte)
+	UnmarshalSia([]byte) error
```
DavidVorick: also we need an int I think, so that the number of bytes used can be returned. UnmarshalSia and MarshalSia can't add a prefix that wouldn't be there otherwise.

modules/host.go
lukechampine: It'll get bumped to `1` when the protocol changes

modules/host/host.go
lukechampine: hmm, I don't think so. In my mind, the grouping indicates "distance" from your project.

modules/host/upload.go
lukechampine: loops without a condition, you mean?

encoding/marshal.go
lukechampine: that would imply reading the entire byte slice into memory before passing it to `Unmarshal`, which means you can't do any sort of streaming. I would prefer to pass a reader. Then `Currency` unmarshalling, for example, would look like:
```go
// UnmarshalSia implements the encoding.SiaUnmarshaler interface.
func (c *Currency) UnmarshalSia(r io.Reader) error {
	b, err := encoding.ReadPrefix(r, 256) // max currency length
	if err != nil {
		return err
	}
	c.i.SetBytes(b)
	return nil
}
```

modules/host.go
DavidVorick: ah. Should an ascii '0' be used instead then? Or maybe a {' ', '-', ' ', 'v', '0'}, or something?

I guess whatever is easiest is what makes sense.

modules/host/upload.go
DavidVorick: yeah.

encoding/marshal.go
DavidVorick: ok, reader is fine.

lukechampine:
ok, revisions are now submitted. You should check that commit to make sure I'm doing it right, though. For now I'm not saving anything extra to disk; since we're using JSON, it will be easy to add new fields as needed without breaking compatibility.

I also discovered a bug related to renaming, although I think we talked about it before. Since we now load siafiles from disk, any renaming you do is lost when you close siad. I see two possible fixes: either change the name inside the siafile, or maintain a name->file mapping inside the renter persistence. I'm leaning towards the latter here; modifying the siafile directly would change its checksum, and the I/O is expensive for large rewrites.

PR #727, opened 2015-08-19
Processed transactions

DavidVorick:
I am worried about the level of testing. Some of the history tests have been commented out because I completely changed the api. I will write them back in.

The full test suite passes, but I'm guessing there are a handful of latent bugs in this PR.

I changed the api according to have more descriptive wallet transactions. Also less confusing. Calling /wallet/siacoins now returns a list of transaction ids - the ids of the transactions that were created to send the siacoins.

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -45,20 +45,19 @@ Wallet
 
 Queries:
 
-* /wallet                   [GET]
-* /wallet/address           [GET]
-* /wallet/backup            [POST]
-* /wallet/encrypt           [POST]
-* /wallet/history           [GET]
-* /wallet/history/$(addr)   [GET]
-* /wallet/lock              [POST]
-* /wallet/seeds             [GET]
-* /wallet/seeds             [POST]
-* /wallet/siacoins          [POST]
-* /wallet/siafunds          [POST]
-* /wallet/transaction/$(id) [GET]
-* /wallet/transactions      [GET]
-* /wallet/unlock            [POST]
+* /wallet                      [GET]
+* /wallet/address              [GET]
+* /wallet/backup               [POST]
+* /wallet/encrypt              [POST]
+* /wallet/lock                 [POST]
+* /wallet/seeds                [GET]
+* /wallet/seeds                [POST]
+* /wallet/siacoins             [POST]
+* /wallet/siafunds             [POST]
+* /wallet/transaction/$(id)    [GET]
+* /wallet/transactions         [GET]
+* /wallet/transactions/$(addr) [GET]
+* /wallet/unlock               [POST]
 
 The first time that the wallet is ever created, the wallet will be unencrypted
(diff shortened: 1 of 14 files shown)


doc/API.md
```diff
@@ -399,21 +319,106 @@ Parameters: none
...
+the related address is usually not important because the wallet arbitrarily
+selects which addresses will fund a transaction. For outputs (incoming money),
+the related address field can be used to determine who has sent money to the
+wallet.
+
+'value' indicates how much money has been moved in the input or output.
+
+A modules.ProcessedOutput takes the following form:
+```
+struct modules.ProcessedTransaction {
```
lukechampine: wrong type name

doc/API.md
```diff
@@ -399,21 +319,106 @@ Parameters: none
...
+
+'processedinputs' is an array of processed inputs detailing the inputs to the
+transaction. More information below.
+
+'processedoutputs' is an array of processed outputs detailing the outputs of
+the transaction. Outputs related to file contracts are excluded.
+
+A modules.ProcessedInput takes the following form:
+```
+struct modules.ProcessedTransaction {
```
lukechampine: wrong type name

lukechampine:
I'm uneasy about packing multiple meanings into one field (as in `ProcessedInput.RelatedAddress` and `ProcessedTransaction.TransactionID`). But I trust that you arrived at this API with the help of poloniex, so if it satisfies him, I can't complain.

DavidVorick:
> But I trust that you arrived at this API with the help of poloniex, so if it satisfies him, I can't complain.

This design is mostly my own, adjusted according to what seemed to be giving him the most trouble. There are a few things in here that were directly requested, but stacking multiple meanings into one type was my idea. The thought was that the structs would become overly complicated if you kept them separate (they are already complicated), and that what was most important was whether money was moving in or out of the wallet.

'FundType' and 'Value' are the fields with stacked meanings?

ProcessedTransaction.RelatedAddress always contains an unlock hash, I'm not sure why you think that's a stacked meaning.

As for TransactionID, I didn't see how else to get a transaction id for the miner payouts. They would have to be separated entirely. I don't have a test for it, but I'm pretty sure that if you put a block id into /wallet/transaction/$(id) it'll spit out the miner payout ProcessedTransaction. It only contains a stacked meaning if you try to use it outside of the wallet api.

I tried to keep things as clean as possible, but the design could potentially use more iterations.

lukechampine:
`FundType` doesn't have stacked meanings; it's a type. But the value of `FundType` dictates the meaning of the `Value` field. It's sort of like a C `union`. Likewise, `RelatedAddress` may always be an unlock hash, but the meaning of "Related" changes depending on context. So the danger is that you lose some type-safety; you have to determine type information at runtime.

My guess is that JS devs are used to stacked meanings, since the language is dynamically typed.

PR #730, opened 2015-08-21
Performance and testing 2

DavidVorick:
Just some brief stuff. Meant for this to be larger, but poloniex asked for another feature - indicate the miner fees in the outputs of a transaction. That has been added.

There's also a typo fix, and a benchmark for unmarshalling empty blocks. I had started to optimize that a bit, but ran into problems with MarshalSia and UnmarshalSia that we discussed.

Because we are upgrading a lot of the persistence stuff, we should probably address the MarshalSia and UnmarshalSia things sooner rather than later.


types/block_bench_test.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"testing"
+
+	"github.com/NebulousLabs/Sia/encoding"
+)
+
+// BenchmarkDecodeEmptyBlock benchmarks decoding an empty block.
+//
+// David's Computer, 08-20-2015: 38 MB/s
```
lukechampine: processor type would probably be more informative

DavidVorick: I waffled on that one. but afaik the processor means more than any of the other hardware. Windows vs. Linux vs. Mac might mean something.

DavidVorick:
Another commit has been added. This one corrects behavior where the wallet was erroneously marking outputs as 'spent' when it hadn't actually spent them.

DavidVorick:
pushed another commit to fix a DoS attack the wallet is currently vulnerable to.

PR #732, opened 2015-08-24
Wallet Polishing

DavidVorick:
Some of the file constructions in the wallet were moved around. Some things were simplified. Compatibility with previous editions of the wallet 0.4.0 have been broken, this should be okay though because it was never officially released.

As of this PR, wallet is release-ready as far as I know. Could use a few more tests, but his should be merged asap.


modules/wallet/encrypt.go
```diff
@@ -120,14 +130,30 @@ func (w *Wallet) unlock(masterKey crypto.TwofishKey) error {
 	return nil
 }
 
+// wipeSecrets erases all of the seeds and secret keys in the wallet.
+func (w *Wallet) wipeSecrets() {
+	// 'for i := range' must be used to prevent copies of secret data from
+	// being made.
```
lukechampine: since you're only accessing the `SecretKeys` field, which is a pointer, a copy should okay. nbd

DavidVorick: you never actually drop keys from the wallet. When you unlock the wallet, all the keys that existed previously will be initialized again. So there's no reason to clean out the map entirely, you just need to wipe the secret keys.

PR #733, opened 2015-08-25
v0.4 tweaks

lukechampine:
Fix a few bugs and implement some minor missing bits of functionality


crypto/rand.go
```diff
@@ -18,3 +18,15 @@ func RandIntn(n int) (int, error) {
 	r, err := rand.Int(rand.Reader, big.NewInt(int64(n)))
 	return int(r.Int64()), err
 }
+
+// Perm returns, as a slice of n ints, a pseudo-random permutation of the
```
DavidVorick: it's not pseudo-random, it's cryptographically random

modules/gateway.go
```diff
@@ -13,6 +13,7 @@ var (
 	BootstrapPeers = []NetAddress{
 		"23.239.14.98:9981",
 		"87.98.216.46:9981",
+		"45.79.132.35:9981",
```
DavidVorick: ping these and check that they are port forwarded + online before you add them:

 68.55.10.144 
109.206.33.225
79.51.183.132
83.211.176.23

they are all volunteers and are aware that boostrap nodes see more traffic than other nodes

lukechampine:
wow, build failing due to github service outage

DavidVorick:
I will do it

DavidVorick:
The siac unconfirmed balance is confusing, it should probably be relative to the confirmed balance. (which means it can be positive or negative)

DavidVorick:
it appears that, if there is an error during the storage proof part of hosting, the host does not properly deallocate and delete the file

DavidVorick:
I've been having trouble keeping my nodes synchronized... maybe we do need to bring back the resync loop

lukechampine:
The only desync I've noticed is that slow nodes can lag behind when blocks are being mined quickly. What sort of problems are you seeing?

DavidVorick:
My nodes would be 100 or so blocks behind, and wouldn't catch up even after disconnecting and reconnecting them to their peers. Restarting the nodes fixed it, but that's a bit dramatic.

Would hate to see stuff like this in the wild. I think it's less likely outside of the testing environment (which is pretty hostile), but it still makes me nervous

PR #737, opened 2015-08-26
Final v0.4 fixes

lukechampine:
Testing confirms that hosts can submit storage proofs and receive siacoins/funds.
After this, we should be ready for release.

### DIFF
--- a/crypto/encrypt.go
+++ b/crypto/encrypt.go
@@ -13,6 +13,10 @@ import (
 	"golang.org/x/crypto/twofish"
 )
 
+const (
+	TwofishOverhead = 28 // number of bytes added by EncryptBytes
+)
+
 var (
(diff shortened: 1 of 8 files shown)


crypto/encrypt.go
```diff
@@ -13,6 +13,10 @@ import (
 	"golang.org/x/crypto/twofish"
 )
 
+const (
+	TwofishOverhead = 28 // number of bytes added by EncryptBytes
```
DavidVorick: should probably be `EncryptOverhead`, it's the GCM that's causing overhead, not Twofish.

DavidVorick:
ok, sounds good. Better/cleaner downloading can be done in 0.4.1, along with better, cleaner uploading.

PR #738, opened 2015-08-26
Consensus Probing

DavidVorick:
The biggest change here is that the consensus package no longer tracks the children of processed blocks. It was only necessary to delete them, and that's a form of pruning that can't even be used for most attacks.

Removing them gives moderate speed boosts to `acceptBlock` (proportional to the size of the block), and removes a fair amount of complex code from the codebase, as well as removing some code from the test suite.

I also fixed an issue in TryTransactionSet, but I don't think that solves the problem where the database occasionally goes inconsistent. I'm guessing the database is at risk of going inconsistent when there are a lot of blocks and reorging. Shouldn't be as much of a problem on the live network where the block times are more sane.  On dev net, my guess is some blocks with file contracts were taking 2-3x as long to verify as they took to mine, which was causing the problems.

There's also some code here which continues to move the consensus set in the direction of being a single database transaction. It's got a fairly long way to go before reaching that, but every step helps.


modules/consensus/accept_bench_test.go
```diff
@@ -1,99 +1,78 @@
...
 	defer cs.Close()
 
-	// The test dir will be reset each time the benchmark
-	// is done.
-	cst, err := createConsensusSetTester("BenchmarkEmptyBlocks")
-	if err != nil {
-		return errors.New("Error creating tester: " + err.Error())
-	}
-	defer cst.closeCst()
+	// Synchronisze the cst and the subscirberless consensus set.
```
lukechampine: typo

PR #739, opened 2015-08-26
host grabs address from Announce now

DavidVorick:


modules/host/announce.go
```diff
@@ -54,11 +55,26 @@ func (h *Host) Announce() error {
...
+		uc, err := h.wallet.NextAddress()
+		if err != nil {
+			return err
+		}
+		h.UnlockHash = uc.UnlockHash()
+		err = h.save()
+		if err != nil {
+			return err
+		}
+	}
```
lukechampine: reorder

modules/host/negotiate.go
```diff
@@ -153,6 +153,11 @@ func (h *Host) considerRevision(txn types.Transaction, obligation contractObliga
 // rpcUpload is an RPC that negotiates a file contract. Under the new scheme,
 // file contracts should not initially hold any data.
 func (h *Host) rpcUpload(conn net.Conn) error {
+	// Check that the host has grabbed an address from the wallet.
+	if h.UnlockHash == (types.UnlockHash{}) {
+		return errors.New("host has not yet been properly initialized - call 'announce' to finish initialization")
```
lukechampine: users won't see this error; maybe it should be printed?

modules/host/announce.go
DavidVorick: I did, though it kept your comment for some reason

modules/host/negotiate.go
DavidVorick: hmm, I guess so. I'll add it.

DavidVorick:
ah crap missed the println lol

PR #743, opened 2015-08-27
eliminate locking bugs

DavidVorick:
there are a few locking problems in the consensus set that were causing database inconsistencies. They should be fixed now.

### DIFF
--- a/modules/consensus/applytransaction.go
+++ b/modules/consensus/applytransaction.go
@@ -5,7 +5,6 @@ import (
 
 	"github.com/boltdb/bolt"
 
-	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/types"
 )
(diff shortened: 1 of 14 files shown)


DavidVorick:
oh I didn't realize this PR was open.

I want a chance to review the diff and make sure it's all sane.

modules/consensus/applytransaction.go
```diff
@@ -26,13 +26,22 @@ var (
...
 	for _, sci := range t.SiacoinInputs {
+		scoBytes := scoBucket.Get(sci.ParentID[:])
+		if scoBytes == nil {
+			panic("nil sco bytes")
+		}
+		var sco types.SiacoinOutput
+		err := encoding.Unmarshal(scoBytes, &sco)
+		if err != nil {
+			return err
+		}
```
DavidVorick: I am worried that this is out of place. Perhaps this should be moved in to a `getSiacoinOutput` function (which doesn't exist at the moment).

modules/consensus/validtransaction.go
```diff
@@ -25,8 +25,41 @@ var (
 	ErrUnfinishedFileContract     = errors.New("file contract window has not yet openend")
 	ErrUnrecognizedFileContractID = errors.New("cannot fetch storage proof segment for unknown file contract")
 	ErrWrongUnlockConditions      = errors.New("transaction contains incorrect unlock conditions")
+
+	// dont judge me
+	errSuccess = errors.New("please")
```
DavidVorick: ah whoops

modules/consensus/applytransaction.go
lukechampine: are you going to leave the bare `panic` in there?

modules/consensus/validtransaction.go
lukechampine: kek

modules/consensus/validtransaction.go
```diff
@@ -271,29 +495,32 @@ func (cs *ConsensusSet) validTransaction(t types.Transaction) error {
...
+				return err
+			}
 		}
+		return errSuccess
+	})
+	// Revert the siafund pool before checking the error.
+	cs.siafundPool = sfpOld
+	if err != errSuccess {
+		return modules.ConsensusChange{}, err
 	}
```
DavidVorick: this is the code block that intentionally never commits the bolt tx.

modules/consensus/applytransaction.go
DavidVorick: oh, good catch.

DavidVorick: I did fix this though github doesn't seem to think so for this comment window

PR #751, opened 2015-08-31
add txn purging to the renter

DavidVorick:
If negotiation failed at some point, the transaction was left in limbo - outputs had been consumed according to the wallet, but actually had not been consumed since the transaction had failed. The transaction builder now includes a purge function that allows for outputs to be 'dropped'
from a transaction builder in the event of a failed transaction.
And the renter will now purge any in-progress transaction if negotiation fails, so that any outputs can be restored.

### DIFF
--- a/modules/host/host.go
+++ b/modules/host/host.go
@@ -91,7 +91,7 @@ func New(cs *consensus.ConsensusSet, hdb modules.HostDB, tpool modules.Transacti
 		// default host settings
 		HostSettings: modules.HostSettings{
 			TotalStorage: 10e9,                        // 10 GB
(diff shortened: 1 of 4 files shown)


lukechampine:
there's probably a cleaner way to do this (e.g. just defer the `Drop` and make it a no-op after `View` or `Sign` is called, a la Bolt) but it's fine for nowt

DavidVorick:
I wanted to keep the ability to call `Drop` after signing the transaction. It's possible that you discard the signed transaction for some other reason.

But maybe not likely that such a situation would happen. I'm not sure yet.

PR #756, opened 2015-08-31
Loading 0.3.3.x Wallets

DavidVorick:
With api and siac support.

### DIFF
--- a/modules/wallet/unseeded.go
+++ b/modules/wallet/unseeded.go
@@ -38,6 +38,14 @@ type SiagKeyPair struct {
 	UnlockConditions types.UnlockConditions
 }
 
+// SavedKey033x is the persist structure that was used to save and load private
+// keys in versions v0.3.3.x for siad.
+type SavedKey033x struct {
+	SecretKey        crypto.SecretKey
+	UnlockConditions types.UnlockConditions
+	Visible          bool
+}
+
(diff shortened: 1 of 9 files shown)


modules/wallet/unseeded.go
```diff
@@ -104,36 +158,69 @@ func (w *Wallet) loadSiagKeys(masterKey crypto.TwofishKey, keyfiles []string) er
...
+		}
+	}
+	err = w.saveSettings()
+	if err != nil {
+		return err
+	}
+	if seedsLoaded != 0 {
+		return w.createBackup(filepath.Join(w.persistDir, "Sia Wallet Encrypted Backup - "+persist.RandomSuffix()+settingsFileSuffix))
+	}
+	return errAllDuplicates
```
lukechampine: should probably flip these blocks

DavidVorick: You think? I am not certain either way. Seems weird to make a backup if the changes are going to fail in your primary.

lukechampine: no, I mean it should be:
```go
if seedsLoaded == 0 {
    return errAllDuplicates
}
```
since that's the failure condition.

DavidVorick: oh yeah. I'll do it in my next PR then

PR #757, opened 2015-08-31
report TimeRemaining properly

lukechampine:
Prevents an underflow


DavidVorick:
odd use of a function. I probably just would have put it as its own function.

PR #758, opened 2015-09-01
Refactoring host weighting calculation to separate file

mtlynch:
Refactoring the host weighting functionality to its own file, expanding its
unit tests.


DavidVorick:
This looks good to me, I'm going to give @lukechampine a chance to share his thoughts

modules/hostdb/hostweight.go
```diff
@@ -0,0 +1,27 @@
...
+var (
+	// Because most weights would otherwise be fractional, we set the base
+	// weight to 10^80 to give ourselves lots of precision when determing the
+	// weight of a host
+	baseWeight = types.NewCurrency(new(big.Int).Exp(big.NewInt(10), big.NewInt(120), nil))
+)
+
+// calculateHostWeight returns the weight of a host according to the settings of
+// the host database entry. Currently, only the price is considered.
+func calculateHostWeight(entry hostEntry) (weight types.Currency) {
```
DavidVorick: guess we didn't need a `HostDB` object in this function.

modules/hostdb/hostweight.go
```diff
@@ -0,0 +1,27 @@
...
+	// If the price is <= 0, just return the base weight.
+	price := entry.Price
+	if price.Cmp(types.NewCurrency64(0)) <= 0 {
+		return baseWeight
+	}
```
lukechampine: AFAIK it is not possible for a negative currency to exist, except if a programming error is made (e.g. subtracting a large currency from a small currency) *and* the code is run without the `build.DEBUG` flag. It should be safe to assume that `price >= 0`, in which case this function simplifies to just the last line.

modules/hostdb/hostweight_test.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/NebulousLabs/Sia/modules"
+	"github.com/NebulousLabs/Sia/types"
+)
+
+func calculateWeightFromUInt64Price(price uint64) (weight types.Currency) {
+	entry := hostEntry{
+		HostSettings: modules.HostSettings{
+			Price: types.NewCurrency64(price),
+		},
+	}
```
lukechampine: since `HostSettings` is embedded, you can access it a little more compactly with:
```go
var entry hostEntry
entry.Price = types.NewCurrency64(price)
```

modules/hostdb/hostweight_test.go
```diff
@@ -0,0 +1,54 @@
...
+		t.Error("Weight of identically priced hosts should be equal.")
+	}
+}
+
+func TestHostWeightWithOnePricedZero(t *testing.T) {
+	weight1 := calculateWeightFromUInt64Price(5)
+	weight2 := calculateWeightFromUInt64Price(0)
+	if weight1.Cmp(weight2) >= 0 {
+		t.Error("Zero-priced host should have higher weight than nonzero-priced " +
+			"host.")
```
lukechampine: we don't wrap error messages, only comments.

modules/hostdb/hostweight.go
mtlynch: If we assume `price >= 0`, don't we still need to make sure `price > 0` (or at least `price != 0`) so we don't divide by zero?

lukechampine: right, right. I thought `price` was the dividend for some reason. In that case, use:
```go
if price.IsZero() {
    return baseWeight
}
```

DavidVorick: baseWeight is something different. You would want to return 1

DavidVorick: edit: got confused sorry

mtlynch: Sounds good, fixed.

modules/hostdb/hostweight_test.go
mtlynch: Fixed

modules/hostdb/hostweight_test.go
mtlynch: Good point, fixed.

lukechampine:
Awesome, thanks!

PR #759, opened 2015-09-01
Ignore IntelliJ IDEA files

bitspill:


DavidVorick:
It seems odd to me that we would put Java IDE files in our .gitignore, but I don't see the harm in it.

We should probably add comments for the other ignored files.

Vim is '*.swp'

pdflatex is 'doc/' stuff.

cover and release are both folders that are created during different parts of the build process (cover measures test coverage, release contains cross-platform binaries)

DavidVorick:
@lukechampine what are your thoughts?

bitspill:
IntelliJ started out as a Java IDE but now they have plugins for "everything".

https://github.com/go-lang-plugin-org/go-lang-idea-plugin

lukechampine:
I don't see any harm in merging this, esp. with comments.

PR #763, opened 2015-09-01
Wallet polish

DavidVorick:
A number of 'polishing' things have been implemented here:

+ there's now an api call + siac command to view all of the addresses known to the wallet in sorted order
+ /wallet/seeds [POST] -> /wallet/load/seed [POST]
+ 25 keys are now preloaded into the wallet. This should reduce some amount of confusion for miners who mine using a single wallet file. Also helps a bit if someone is loading an old backup incorrectly. 25 is probably too low, I think it should be closer to 250, but since that delays startup time by a few seconds I've elected to keep it low until most of the startup routines are happening in the background.

Still on my todo list for 0.4.1:

+ API + siac support for loading siag keys (it's already implemented in the wallet)
+ siac command to view a list of transactions related to the wallet + the net balance for each transaction
+ more unit tests

Wallet is really starting to come together as a finished product. At least, until we decide to start using a database to house the transactions and such =/. But that's not going to be soon.


modules/wallet/wallet.go
```diff
@@ -133,3 +134,19 @@ func New(cs modules.ConsensusSet, tpool modules.TransactionPool, persistDir stri
...
+func (w *Wallet) AllAddresses() []types.UnlockHash {
+	lockID := w.mu.RLock()
+	defer w.mu.RUnlock(lockID)
+
+	i := 0
+	addrs := types.UnlockHashSlice(make([]types.UnlockHash, len(w.keys)))
+	for addr := range w.keys {
+		addrs[i] = addr
+		i++
+	}
```
lukechampine: prefer
```go
addrs := make(types.UnlockHashSlice, 0, len(w.keys))
for addr := range w.keys {
    addrs = append(addrs, addr)
}
```

api/api.go
```diff
@@ -101,12 +101,14 @@ func (srv *Server) initAPI(addr string) {
 	if srv.wallet != nil {
 		handleHTTPRequest(mux, "/wallet", srv.walletHandler)                           // GET
 		handleHTTPRequest(mux, "/wallet/address", srv.walletAddressHandler)            // GET
+		handleHTTPRequest(mux, "/wallet/addresses", srv.walletAddressesHandler)        // GET
 		handleHTTPRequest(mux, "/wallet/backup", srv.walletBackupHandler)              // POST
 		handleHTTPRequest(mux, "/wallet/encrypt", srv.walletEncryptHandler)            // POST - COMPATv0.4.0
 		handleHTTPRequest(mux, "/wallet/init", srv.walletInitHandler)                  // POST
 		handleHTTPRequest(mux, "/wallet/load/033x", srv.walletLoad033xHandler)         // POST
+		handleHTTPRequest(mux, "/wallet/load/seed", srv.walletLoadSeedHandler)         // POST
```
lukechampine: does this change need to be reflected in `siac`?

DavidVorick: Oh I didn't realize that siac had ever implemented it

DavidVorick:
ok fixed up

PR #765, opened 2015-09-02
Refactors TransactionBuilder so that it does not depend on Wallet type

mtlynch:
Adds a new interface walletSpender, which wraps a subset of the Wallet type's
functionality. This allows us to unit test TransactionBuilder without a
dependency on the Wallet type. It will be possible to test TransactionBuilder
using a mock that implements walletSpender.

### DIFF
--- a/modules/wallet/transactionbuilder.go
+++ b/modules/wallet/transactionbuilder.go
@@ -9,6 +9,90 @@ import (
 	"github.com/NebulousLabs/Sia/types"
 )
 
+// walletSpender allows clients to retrieve and modify keys and outputs in a
+// wallet in order to spend siacoins.
+type walletSpender interface {
+	// acquireLock acquires a mutex on walletSpender. Clients must call
+	// acquireLock before calling any other function in walletSpender and must
+	// call releaseLock before exiting scope.
+	acquireLock()
+	releaseLock()
+
+	siacoinOutputs() map[types.SiacoinOutputID]types.SiacoinOutput
+	siafundOutputs() map[types.SiafundOutputID]types.SiafundOutput
+	unconfirmedProcessedTransactions() []modules.ProcessedTransaction
+
+	// spendHeight returns the block height at which a given output was spent.
+	spendHeight(types.OutputID) types.BlockHeight
+
+	// setSpendHeight marks the output with the given ID as spent in the given
+	// block.
+	setSpendHeight(types.OutputID, types.BlockHeight)
+
+	// key returns the key associated with the given hash.
+	key(types.UnlockHash) spendableKey
+
+	// hasKey indicates whether a key with the given hash exists in the wallet.
+	hasKey(types.UnlockHash) bool
+
+	// consensusSetHeight returns the current block height of the wallet.
+	consensusSetHeight() types.BlockHeight
+
+	// nextPrimarySeedAddress fetches the next address from the primary seed.
+	nextPrimarySeedAddress() (types.UnlockConditions, error)
+}
+
+// spendableWallet is a wrapper over Wallet that implements walletSpender.
+type spendableWallet struct {
+	wallet *Wallet
+	lockID int
+}
+
+func (w *spendableWallet) acquireLock() {
+	w.lockID = w.wallet.mu.Lock()
+}
+
+func (w *spendableWallet) releaseLock() {
+	w.wallet.mu.Unlock(w.lockID)
+}
+
+func (w *spendableWallet) siacoinOutputs() map[types.SiacoinOutputID]types.SiacoinOutput {
+	return w.wallet.siacoinOutputs
+}
+
+func (w *spendableWallet) siafundOutputs() map[types.SiafundOutputID]types.SiafundOutput {
+	return w.wallet.siafundOutputs
+}
+
+func (w *spendableWallet) spendHeight(oid types.OutputID) types.BlockHeight {
+	return w.wallet.spentOutputs[oid]
+}
+
+func (w *spendableWallet) setSpendHeight(oid types.OutputID, height types.BlockHeight) {
+	w.wallet.spentOutputs[oid] = height
+}
+
+func (w *spendableWallet) unconfirmedProcessedTransactions() []modules.ProcessedTransaction {
+	return w.wallet.unconfirmedProcessedTransactions
+}
+
+func (w *spendableWallet) key(hash types.UnlockHash) spendableKey {
+	return w.wallet.keys[hash]
+}
+
+func (w *spendableWallet) hasKey(hash types.UnlockHash) bool {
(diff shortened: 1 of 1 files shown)


modules/wallet/transactionbuilder.go
```diff
@@ -9,6 +9,90 @@ import (
...
+
+	// hasKey indicates whether a key with the given hash exists in the wallet.
+	hasKey(types.UnlockHash) bool
+
+	// consensusSetHeight returns the current block height of the wallet.
+	consensusSetHeight() types.BlockHeight
+
+	// nextPrimarySeedAddress fetches the next address from the primary seed.
+	nextPrimarySeedAddress() (types.UnlockConditions, error)
+}
```
DavidVorick: This doesn't really succeed at the goal of decoupling the transaction builder from the wallet. All you really did was wrap the used fields with setters and getters to the wallet, and not in a way that promotes type safety.

As much as possible, you should avoid getters and setters, and instead focus on the bigger picture. What you really need is some way to select a set of outputs to fund a transaction, to get signatures for those outputs, and then to have some control over indicating whether they are 'reserved' or not (that's all the 'spendHeight' is for - it tells the wallet not to use that output again for a while).

I think you can reduce this interface to about 4 functions:

`findOutputs`, `signTransaction` (implementation details on this one might be tricky), `reserveOutputs`, `releaseOutputs`.

But you definitely don't want to be exposing the inner state of the wallet. It defeats the point of decoupling and adds a lot of overhead when trying to figure out what's safe to modify and whatnot.

modules/wallet/transactionbuilder.go
```diff
@@ -9,6 +9,90 @@ import (
...
 )
 
+// walletSpender allows clients to retrieve and modify keys and outputs in a
+// wallet in order to spend siacoins.
+type walletSpender interface {
+	// acquireLock acquires a mutex on walletSpender. Clients must call
+	// acquireLock before calling any other function in walletSpender and must
+	// call releaseLock before exiting scope.
+	acquireLock()
+	releaseLock()
```
DavidVorick: These don't belong here. Check out the mutex section:

https://github.com/NebulousLabs/Sia/blob/master/doc/Developers.md

Sia is a pretty complex project, and the locking is already really complex in its simplified form. As soon as you start letting modules grab and release locks manually, you open a whole set of complexity and places for things like deadlocks and race conditions to hide. It gets out of control quickly. The rule that we use to keep things as simple as possible is that modules need to handle their own locking internally. If they expose a function through an interface, then that function needs to be thread-safe.

There's another rule that's only just being introduced. You haven't violated it, but I thought I'd mention it anyway since it's not documented anywhere yet. A module that is holding a write-lock on itself is not allowed to make any external calls. This prevents a lot of potential deadlocks and eliminates the need to understand how the locking structure of foreign interfaces operates.

modules/wallet/transactionbuilder.go
```diff
@@ -9,6 +9,90 @@ import (
...
+func (w *spendableWallet) acquireLock() {
+	w.lockID = w.wallet.mu.Lock()
+}
+
+func (w *spendableWallet) releaseLock() {
+	w.wallet.mu.Unlock(w.lockID)
+}
+
+func (w *spendableWallet) siacoinOutputs() map[types.SiacoinOutputID]types.SiacoinOutput {
+	return w.wallet.siacoinOutputs
```
DavidVorick: you definitely can't be retuning a map that's part of the wallet's internal state. Sia is heavily concurrent, and maps are not thread safe. When you read from a map, you need to have certainty that it is not being concurrently modified somewhere else. When you pass the map outside the wallet, you open yourself to race conditions that are almost guaranteed to trigger.

modules/wallet/transactionbuilder.go
```diff
@@ -9,6 +9,90 @@ import (
...
+
+	// nextPrimarySeedAddress fetches the next address from the primary seed.
+	nextPrimarySeedAddress() (types.UnlockConditions, error)
+}
+
+// spendableWallet is a wrapper over Wallet that implements walletSpender.
+type spendableWallet struct {
+	wallet *Wallet
+	lockID int
+}
```
DavidVorick: Any interface that you create to use with the transaction builder should be satisfied by the wallet directly, instead of using a secondary struct that has a wallet inside of it.

DavidVorick:
You're probably going to have to chalk this up as a good first attempt, but I don't think much of the code in this PR can be salvaged. What really kills it is the thread-unsafety, and I don't really see any easy way to make this code thread-safe in a maintainable way - it just needs to be rewritten.

To keep developer overhead minimal, it's important to enforce that locking is never exposed on a Sia interface and that all Sia interfaces are fully thread-safe.

mtlynch:
Hmm, I think I probably wasn't as clear as I should be about my strategy in this PR. Here's what I had in mind:

1. Break up the direct dependence between `TransactionBuilder` and `Wallet` with an interface over `Wallet` that gives `TransactionBuilder` the same access to `Wallet` that it had before, just through an explicit interface. This interface will *not* be thread-safe because `TransactionBuilder`'s previous access to `Wallet` was not thread-safe (assumed that the caller locked and unlocked properly). The goal is to minimize the amount of code churn needed to add unit tests to the existing behavior.
1. Add unit tests for `TransactionBuilder` to give us confidence that any changes we make do not break current functionality.
1. Fix #765, add unit tests
1. Clean up the `walletSpender` interface to something clean (I think your four method suggestion above is a good one). Locks will not be part of the interface.

In other words, this PR introduces a lot of sin (or rather makes existing sin more explicit), but `TransactionBuilder` is delicate code and we should minimize changes until the code is under test. Going directly to step 4 requires a lot of changes to the implementation of `TransactionBuilder` which is risky if we have no unit tests to verify we are not breaking existing functionality. The things you correctly pointed out as thread-unsafe will not be part of the final `walletSpender` interface, but exist as temporary scaffolding to help us transition to a cleaner interface.

Given this, does this change your perception of the PR? Does it make more sense for me to pull into a feature branch?

DavidVorick:
This definitely can't go into master :P

If you want to keep working from here, I guess that's fine. I can make a feature branch, or we can just keep this PR open and I'll review commits as you add them. I don't think a feature branch is necessary, just having this PR open should be sufficient.

mtlynch:
>If you want to keep working from here, I guess that's fine. I can make a feature branch, or we can just keep this PR open and I'll review commits as you add them. I don't think a feature branch is necessary, just having this PR open should be sufficient.

Cool, can you create a feature branch? It's much easier process-wise if I send you pull requests than if you have to look at each individual commit (e.g. there might be commits when I'm just renaming something or fixing a typo)

DavidVorick:
Ok, created a 'tbuilder' branch for you

PR #766, opened 2015-09-02
Migrate siac to Sia-Agent

bitspill:
Go 1.5 changed the default net.http user-agent causing siac to trigger "Browser access disabled due to security vulnerability. Use Sia-UI or siac." 

This pull is intended to migrate siac to using "Sia-Agent" and thus resolve that issue.

### DIFF
--- a/siac/main.go
+++ b/siac/main.go
@@ -29,7 +29,7 @@ func apiGet(call string) (*http.Response, error) {
 		addr = net.JoinHostPort("localhost", port)
 	}
 	client := &http.Client{}
-	req, err := http.NewRequest("GET", "http://"+addr+call, nil)
+	req, _ := http.NewRequest("GET", "http://"+addr+call, nil)
 	req.Header.Add("User-Agent", "Sia-Agent")
 	resp, err := client.Do(req)
 	if err != nil {
@@ -80,7 +80,7 @@ func apiPost(call, vals string) (*http.Response, error) {
 	}
 
 	client := &http.Client{}
-	req, err := http.NewRequest("POST", "http://"+addr+call, strings.NewReader(vals))
+	req, _ := http.NewRequest("POST", "http://"+addr+call, strings.NewReader(vals))
 	req.Header.Add("User-Agent", "Sia-Agent")
 	req.Header.Add("Content-Type", "application/x-www-form-urlencoded")
 	resp, err := client.Do(req)
--- a/siac/main.go
+++ b/siac/main.go
@@ -29,8 +29,13 @@ func apiGet(call string) (*http.Response, error) {
 	if host, port, _ := net.SplitHostPort(addr); host == "" {
 		addr = net.JoinHostPort("localhost", port)
 	}
-
-	resp, err := http.Get("http://" + addr + call)
+	client := &http.Client{}
+	req, err := http.NewRequest("GET", "http://"+addr+call, nil)
+	if err != nil {
+		return nil, errors.New("no response from daemon")
+	}
+	req.Header.Add("User-Agent", "Sia-Agent")
+	resp, err := client.Do(req)
 	if err != nil {
 		return nil, errors.New("no response from daemon")
 	}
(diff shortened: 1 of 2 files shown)


bitspill:
Hmm... Guess it's about time I set up local testing and stop abusing travis...

lukechampine:
hmm. That error is nondeterministic. I need to look into it.

lukechampine:
I wanted to do this anyway, so thank you!

You should also remove the `"Go 1.1 package http"` user agent check from api/api.go

siac/main.go
```diff
@@ -29,8 +28,13 @@ func apiGet(call string) (*http.Response, error) {
 	if host, port, _ := net.SplitHostPort(addr); host == "" {
 		addr = net.JoinHostPort("localhost", port)
 	}
-
-	resp, err := http.Get("http://" + addr + call)
+	client := &http.Client{}
+	req, err := http.NewRequest("GET", "http://"+addr+call, nil)
+	if err != nil {
+		return nil, errors.New("no response from daemon")
```
lukechampine: `NewRequest` only returns an error if the url is malformed, and we know in advance that that will never be the case. So this error (and its parallel in `apiPost`) doesn't need to be checked.

bitspill:
Made the two suggested changes.

I was confused as to how the renter test was failing, maybe it will work this time.

lukechampine:
hmm... removing the check from `api/api.go` broke some existing tests. They will have to be switched over as well. :/

Not sure if it's best to leave the check, or fix everything now. Probably the latter, but it depends on how much work that requires.

bitspill:
I'm looking into those tests now.

siac/main.go
```diff
@@ -29,8 +28,10 @@ func apiGet(call string) (*http.Response, error) {
 	if host, port, _ := net.SplitHostPort(addr); host == "" {
 		addr = net.JoinHostPort("localhost", port)
 	}
-
-	resp, err := http.Get("http://" + addr + call)
+	client := &http.Client{}
+	req, err := http.NewRequest("GET", "http://"+addr+call, nil)
```
lukechampine: use `req, _ :=` to show your intent; otherwise it looks like the error is just being overwritten by the subsequent call to `client.Do`.

lukechampine:
you know, maybe the `get` and post` functions should just be defined in the api package (and exported). Then both siac and the tests could use the same functions. Just a thought.

lukechampine:
ok, this looks good. I'll migrate the remaining tests later.

bitspill:
I think I just found the issue with the tests

lukechampine:
Rather than fix the test versions of `get` and `post`, I would prefer to do what I suggested above and just define those functions inside the api package. If you want to tackle that yourself, that would be great.

bitspill:
Ah, well I just fixed the other tests. I'll try and see about moving the functions as you suggested now.

siac/main.go
```diff
@@ -29,8 +28,10 @@ func apiGet(call string) (*http.Response, error) {
 	if host, port, _ := net.SplitHostPort(addr); host == "" {
 		addr = net.JoinHostPort("localhost", port)
 	}
-
-	resp, err := http.Get("http://" + addr + call)
+	client := &http.Client{}
+	req, _ := http.NewRequest("GET", "http://"+addr+call, nil)
```
DavidVorick: should this error be getting checked?

siac/main.go
```diff
@@ -78,11 +79,11 @@ func apiPost(call, vals string) (*http.Response, error) {
 		addr = net.JoinHostPort("localhost", port)
 	}
 
-	data, err := url.ParseQuery(vals)
-	if err != nil {
-		return nil, errors.New("bad query string")
-	}
-	resp, err := http.PostForm("http://"+addr+call, data)
+	client := &http.Client{}
+	req, _ := http.NewRequest("POST", "http://"+addr+call, strings.NewReader(vals))
```
DavidVorick: ditto with this error

siac/main.go
bitspill: That change was made at the request of luke, 
"NewRequest only returns an error if the url is malformed, and we know in advance that that will never be the case. So this error (and its parallel in apiPost) doesn't need to be checked."

siac/main.go
bitspill: ditto with this error

siac/main.go
DavidVorick: Hmm, there's still a chance of a developer mistake down the line that results in malformed urls. I'm not sure I agree that the error should not be checked.

siac/main.go
DavidVorick: Then at least it needs a comment explaining why the error is not checked.

DavidVorick:
I thought I commented this already but I don't see it anywhere so I'll comment again. Sorry if it's redundant.

Looks like the rebase failed, PR has 6 commits when I think it only needed 1 or 2. The old commits are still there.

PR #767, opened 2015-09-02
Refactor api get/post requests

bitspill:
Just after the previous pull request was merged I think I found the issue.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -14,12 +14,33 @@ const (
 	apiTimeout = 5 * time.Second
 )
 
+// HttpGET is a utility function for making http get requests to sia with a whitelisted user-agent
+func HttpGET(url string) (resp *http.Response, err error) {
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Add("User-Agent", "Sia-Agent")
+	return new(http.Client).Do(req)
+}
+
+// HttpPOST is a utility function for making post requests to sia with a whitelisted user-agent
+func HttpPOST(url string, data string) (resp *http.Response, err error) {
+	req, err := http.NewRequest("POST", url, strings.NewReader(data))
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Add("User-Agent", "Sia-Agent")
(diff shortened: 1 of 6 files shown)


bitspill:
Currently looking into refactoring the get/post api functions so they are not duplicated in the way they currently are.

bitspill:
I'm new to Go so let me know if I did something completely wrong here.

bitspill:
I believe the refactor is complete. I'm probably going to sleep now so responses may be slow.

api/api.go
```diff
@@ -14,12 +14,29 @@ const (
 	apiTimeout = 5 * time.Second
 )
 
+// HttpGET is a utility function for making http get requests with Sia-Agent set
```
DavidVorick: the wording is a little awkward., same for the HttpPOST comment

DavidVorick:
Looks good to me

api/api.go
```diff
@@ -14,12 +14,29 @@ const (
...
+	client := &http.Client{}
+	req, _ := http.NewRequest("GET", url, nil)
+	req.Header.Add("User-Agent", "Sia-Agent")
+	return client.Do(req)
+}
+
+// HttpPOST is a utility function for making post requests with Sia-Agent set
+func HttpPOST(url string, data string) (resp *http.Response, err error) {
+	client := &http.Client{}
+	req, _ := http.NewRequest("POST", url, strings.NewReader(data))
```
DavidVorick: I still think this error should either be checked or have a comment. Waiting for @lukechampine to give his opinion

lukechampine: sure, no harm in it.
Also, since we're not configuring the `client` at all, you can just use `return new(http.Client).Do(req)`.

api/api.go
lukechampine: I would even prefer just plain `Get` and `Post`, since the meaning is fairly clear. (Actually, it's less clear inside the package; `api.Get` is much more obvious than `Get`)

bitspill: I considered `api.Get/Post` but to me those were quite similar to `apiPost` and `apiGet` within siac and a few other locations so I chose `api.HttpGET/POST` to ensure the distinction was clear

PR #769, opened 2015-09-02
Siag support in api + siac

DavidVorick:
Now you can load siag keys into the standard wallet, which means the wallet will be able to see and send siafunds. Siafunds are still strictly a `siac` thing, but at least it's a lot easier than it used to be.

siad still needs to be restarted after loading keys to find balances - live rescanning not yet supported.


modules/wallet.go
```diff
@@ -187,10 +187,9 @@ type (
 		Drop()
 	}
 
-	// Wallet stores and manages siacoins and siafunds. The wallet file is
-	// encrypted using a user-specified password. Common addresses are all
-	// dervied from a single address seed.
-	Wallet interface {
+	// EncryptionManager can encrypt, lock, unlock, and indicate the current
+	// status of the EncryptionManager.
+	EncryptionManager interface {
```
lukechampine: I'm not sure what the value is in splitting the Wallet interface up like this. Is an `EncryptionManager` or a `KeyManager` useful on its own?

siac/walletcmd.go
```diff
@@ -97,6 +90,20 @@ The seed string, which is also the encryption password, will be returned.`,
...
 	}
 
+	walletLoadseedCmd = &cobra.Command{
+		Use:   `loadseed`,
+		Short: "Add a seed to the wallet",
+		Long:  "Uses the given password to create a new wallet with that as the primary seed",
+		Run:   wrap(walletloadseedcmd),
+	}
+
+	walletLoadsiagCmd = &cobra.Command{
```
lukechampine: I thought the format was "capitalize after each /"

modules/wallet.go
DavidVorick: The value is the same as splitting up a large function into smaller functions: it breaks the code into more edible - and potentially more easily tested - chunks.

Having a 25 function interface didn't seem great to me. Breaking it into 3 chunks is nicer.

siac/walletcmd.go
DavidVorick: Oh. I had a different interpretation for siac: Capitalize with each new command.

IE:

siac dog cat manbearpig => dogCatManbearpigCmd

I do agree that `load` should become its own command, but I didn't know how to do it so I left it for you :P

modules/wallet.go
lukechampine: the interface still has 25 functions, you just hid some of them.

the disadvantage is that it's confusing to define a separate interface when there isn't a clear use case for it. None of the other modules work with an `EncryptionManager` or a `KeyManager`; they all need the full `Wallet`. Maybe there's a stronger need for such interfaces inside the wallet package, but in that case, those interfaces should be defined there, instead of at the module level.

siac/walletcmd.go
lukechampine: ok, I can push directly to this branch if you want

DavidVorick: yeah that is fine

PR #770, opened 2015-09-02
Refactors TransactionBuilder so that it does not depend on Wallet type

mtlynch:
Adds a new interface walletSpender, which wraps a subset of the Wallet type's
functionality. This allows us to unit test TransactionBuilder without a
dependency on the Wallet type. It will be possible to test TransactionBuilder
using a mock that implements walletSpender.

### DIFF
--- a/modules/wallet/transactionbuilder.go
+++ b/modules/wallet/transactionbuilder.go
@@ -9,6 +9,90 @@ import (
 	"github.com/NebulousLabs/Sia/types"
 )
 
+// walletSpender allows clients to retrieve and modify keys and outputs in a
+// wallet in order to spend siacoins.
+type walletSpender interface {
+	// acquireLock acquires a mutex on walletSpender. Clients must call
+	// acquireLock before calling any other function in walletSpender and must
+	// call releaseLock before exiting scope.
+	acquireLock()
+	releaseLock()
+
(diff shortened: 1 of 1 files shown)


modules/wallet/transactionbuilder.go
```diff
@@ -17,7 +108,7 @@ type transactionBuilder struct {
 	siacoinInputs []int
 	siafundInputs []int
 
-	wallet *Wallet
+	wallet walletSpender
```
DavidVorick: not that it's a huge deal, but the name probably shouldn't be 'wallet' since it's actually a walletSpender.

DavidVorick:
ok, just so we're clear on some things that have to happen at some point, even if not this PR:

+ interface should contain only exported functions
+ transaction builder should be split into a separate package
+ maps and other internal state cannot be exposed by the interface, but especially not maps
+ locks cannot be exposed by the interface, interface must be thread-safe without exposed locks

DavidVorick:
other than the name 'wallet', I think this is ready to be merged into tbuilder

mtlynch:
Cool, fixed thanks for the review.

DavidVorick:
Luke and I discussed, if the transaction builder gets moved into its own package, it should be a subpackage of the wallet

PR #771, opened 2015-09-02
Put SiacoinPool inside of database

DavidVorick:
There are some optimizations that could be made here, but the important part - always use siafund pool from within the database - is complete.

This is necessary to provide safety with the TryTransactions function.

### DIFF
--- a/modules/consensus/accept_test.go
+++ b/modules/consensus/accept_test.go
@@ -8,6 +8,8 @@ import (
 	"testing"
 	"time"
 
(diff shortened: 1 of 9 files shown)


DavidVorick:
muxado =/

lukechampine:
yeah, really annoying. Is it always the `TestListen` function? I'm trying to reproduce it locally.

DavidVorick:
build is passed

PR #772, opened 2015-09-03
User-specified announcements

lukechampine:
The user can now supply a specific address to be announced. This allows you to announce a domain instead of an IP. It also replaces the old "force" flag; now we assume that if you supply an address, you want to force the announcement. I like this because it means if you want to force an announcement, you have to lookup your IP first.

### DIFF
--- a/modules/host/announce.go
+++ b/modules/host/announce.go
@@ -51,10 +51,6 @@ func (h *Host) announce(addr modules.NetAddress) error {
 // arbitrary data, signing the transaction, and submitting it to the
 // transaction pool.
 func (h *Host) Announce() error {
-	lockID := h.mu.RLock()
-	addr := h.myAddr
-	h.mu.RUnlock(lockID)
-
 	// Generate an unlock hash, if necessary
 	if h.UnlockHash == (types.UnlockHash{}) {
 		uc, err := h.wallet.NextAddress()
@@ -67,6 +63,12 @@ func (h *Host) Announce() error {
 			return err
 		}
 	}
+
+	// Get the external IP again; it may have changed.
+	h.learnHostname()
+	lockID := h.mu.RLock()
+	addr := h.myAddr
+	h.mu.RUnlock(lockID)
 
 	// Check that the host's ip address is both known and reachable.
 	if addr.Host() == "::1" {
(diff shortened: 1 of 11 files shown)


DavidVorick:
`modules/host/announce_test.go:17: not enough arguments in call to ht.host.ForceAnnounce`

modules/host/announce.go
```diff
@@ -27,6 +27,19 @@ func ping(addr modules.NetAddress) bool {
...
+		uc, err := h.wallet.NextAddress()
+		if err != nil {
+			return err
+		}
+		h.UnlockHash = uc.UnlockHash()
+		err = h.save()
+		if err != nil {
+			return err
+		}
+	}
```
DavidVorick: :|

I guess the definition of --force announce has changed

modules/renter/download.go
```diff
@@ -266,11 +253,23 @@ func (r *Renter) Download(nickname, destination string) error {
...
+	perm := os.FileMode(file.mode)
+	if perm == 0 {
+		// sane default
+		perm = 0666
+	}
+	f, err := os.OpenFile(destination, os.O_CREATE|os.O_RDWR|os.O_TRUNC, perm)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
```
DavidVorick: I'm not sure how SUID and SGID stuff works, but that might be a concern. I think the system is safer than that though - I don't think you'd be allowed to download a SUID program that lets anyone run as root if you are in userspace.

But I'm not sure if it would cause errors during the download, we should probably test it somehow.

DavidVorick: but after 0.4.1

modules/renter/persist_test.go
```diff
@@ -126,11 +126,19 @@ func TestRenterSaveLoad(t *testing.T) {
...
+	f1 = newTestingFile()
+	f2 = newTestingFile()
+	f3 = newTestingFile()
+	// names must not conflict
+	for f2.name == f1.name || f2.name == f3.name {
+		f2 = newTestingFile()
+	}
+	for f3.name == f1.name || f3.name == f2.name {
+		f3 = newTestingFile()
+	}
```
DavidVorick: lol

PR #773, opened 2015-09-03
add transactions command to siac

DavidVorick:


siac/walletcmd.go
```diff
@@ -373,6 +381,61 @@ Siafund Claims:      %v SC
...
+// providing a net flow of siacoins and siafunds for each.
+func wallettransactionscmd() {
+	wtg := new(api.WalletTransactionsGET)
+	err := getAPI("/wallet/transactions?startheight=0&endheight=10000000", wtg)
+	if err != nil {
+		fmt.Println("Could not fetch transaction history:", err)
+		return
+	}
+
+	fmt.Println("[height]\t[transaction id]\t\t\t\t\t\t\t[net siacoin transfer]\t[net siafund transfer]")
```
lukechampine: tabs aren't constant length; need to use spaces

siac/walletcmd.go
```diff
@@ -373,6 +381,61 @@ Siafund Claims:      %v SC
...
+				incomingSiafunds = incomingSiafunds.Add(output.Value)
+			}
+		}
+
+		// Convert the siacoins to a float.
+		incomingSiacoinsFloat, _ := new(big.Rat).SetFrac(incomingSiacoins.Big(), types.SiacoinPrecision.Big()).Float64()
+		outgoingSiacoinsFloat, _ := new(big.Rat).SetFrac(outgoingSiacoins.Big(), types.SiacoinPrecision.Big()).Float64()
+
+		// Print the results.
+		fmt.Printf("%v\t\t%v\t %.2f SC\t\t", txn.ConfirmationHeight, txn.TransactionID, incomingSiacoinsFloat-outgoingSiacoinsFloat)
```
lukechampine: see if you can use a width specifier to do the alignment, something like `"%10v %v %10.2f SC"`

siac/walletcmd.go
DavidVorick: maybe I should just not worry about formatting the first line until we can get a more consistent unit

siac/walletcmd.go
DavidVorick: actually, it looks like you can do that... nice!

DavidVorick:
fixed up:  use spaces not tabs, and have everything align correctly always

PR #774, opened 2015-09-03
blacklist localhost

DavidVorick:
this is an incomplete solution, but it eliminates all the problems that have been seen in the wild. In the future, a more powerful solution should be used.

### DIFF
--- a/modules/hostdb/hostentry.go
+++ b/modules/hostdb/hostentry.go
@@ -1,6 +1,8 @@
 package hostdb
 
 import (
+	"net"
+
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/types"
 )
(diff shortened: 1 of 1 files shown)


modules/hostdb/hostentry.go
```diff
@@ -16,6 +16,18 @@ type hostEntry struct {
...
+	case "localhost":
+		return
+	case "127.0.0.1":
+		return
+	case "::1":
+		return
+	default:
+		// do nothing
+	}
+
```
lukechampine: probably better:
```go
// Blacklist localhost.
if host.IPAddress.Host() == "localhost" {
    return
} else if ip := net.ParseIP(string(host.IPAddress)); ip != nil && ip.IsLoopback() {
    return
}
```

DavidVorick: will IsLoopback() catch "localhost" as well?

PR #775, opened 2015-09-03
remove siag

### LINKED ISSUE #587: Merge siag and siad
I'm not really sure why they were ever separate in the first place, time crunch perhaps?

Anyway I think it makes sense to combine them.

DavidVorick:
fix #587 

Siag is not needed anymore. Anyone who wants to use siafunds can just use standard wallet addresses. Anyone who wants to load their current siag addresses into the wallet can do so using siac.

`./siac wallet load siag`


lukechampine:
:tada:

DavidVorick:
build passed

Issue #748, opened 2015-08-28
siac set storage


DavidVorick:
`./siac host config totalstorage 100GB`

will set the total storage to 100 bytes.

PR #777, opened 2015-09-04
Add support for convenient sizes "100GB", "100GiB", etc

bitspill:
Fix issue #748

### DIFF
--- a/siac/hostcmd.go
+++ b/siac/hostcmd.go
@@ -62,6 +62,14 @@ func hostconfigcmd(param, value string) {
 		p.Mul(p, big.NewRat(1e24/1e9, 4320))
 		value = new(big.Int).Div(p.Num(), p.Denom()).String()
 	}
+	// parse sizes of form 10GB, 10TB, 1TiB etc
+	if param == "totalstorage" || param == "minfilesize" || param == "maxfilesize" {
+		var err error // must be pre-declared because value is
+		value, err = parseSize(value)
+		if err != nil {
+			fmt.Println("could not parse " + param)
+		}
+	}
 	err := post("/host/configure", param+"="+value)
 	if err != nil {
 		fmt.Println("Could not update host settings:", err)
--- a/siac/parse.go
+++ b/siac/parse.go
@@ -1,0 +1,77 @@
+package main
+
+import (
+	"errors"
+	"fmt"
+	"math/big"
+	"strings"
+)
+
+var errUnableToParseSize = errors.New("unable to parse size")
+
+// parseSize converts strings of form 10GB to a size in bytes. Fractional sizes
+// are truncated at the byte size. If parsing fails an error is returned.
+func parseSize(strSize string) (string, error) {
+	units := []struct {
+		suffix     string
+		multiplier int64
+	}{
+		{"kb", 1000},
+		{"mb", 1000000},
+		{"gb", 1000000000},
+		{"tb", 1000000000000},
+		{"kib", 1024},
+		{"mib", 1048576},
+		{"gib", 1073741824},
+		{"tib", 1099511627776},
+		{"b", 1}, // must be after others else it'll match on them all
+		{"", 1},  // no suffix is still a valid suffix
+	}
+
+	strSize = strings.ToLower(strSize)
+	for _, unit := range units {
+		if strings.HasSuffix(strSize, unit.suffix) {
+			r, ok := new(big.Rat).SetString(strings.TrimSuffix(strSize, unit.suffix))
+			if !ok {
+				return "", errUnableToParseSize
+			}
+			r.Mul(r, new(big.Rat).SetInt(big.NewInt(unit.multiplier)))
+			if !r.IsInt() {
+				f, _ := r.Float64()
+				return fmt.Sprintf("%d", int64(f)), nil
+			}
+			return r.RatString(), nil
+		}
+	}
+
+	return "", errUnableToParseSize
+}
+
+// coinUnits converts a siacoin amount to base units.
+func coinUnits(amount string) (string, error) {
+	units := []string{"pS", "nS", "uS", "mS", "SC", "KS", "MS", "GS", "TS"}
+	for i, unit := range units {
+		if strings.HasSuffix(amount, unit) {
+			// scan into big.Rat
+			r, ok := new(big.Rat).SetString(strings.TrimSuffix(amount, unit))
+			if !ok {
+				return "", errors.New("malformed amount")
+			}
+			// convert units
+			exp := 24 + 3*(int64(i)-4)
+			mag := new(big.Int).Exp(big.NewInt(10), big.NewInt(exp), nil)
+			r.Mul(r, new(big.Rat).SetInt(mag))
+			// r must be an integer at this point
+			if !r.IsInt() {
+				return "", errors.New("non-integer number of hastings")
+			}
+			return r.RatString(), nil
+		}
+	}
+	// check for hastings separately
+	if strings.HasSuffix(amount, "H") {
+		return strings.TrimSuffix(amount, "H"), nil
+	}
+
+	return "", errors.New("amount is missing units; run 'wallet --help' for a list of units")
+}
--- a/siac/parse.go
+++ b/siac/parse.go
@@ -16,14 +16,14 @@ func parseSize(strSize string) (string, error) {
 		suffix     string
 		multiplier int64
 	}{
-		{"kb", 1000},
-		{"mb", 1000000},
-		{"gb", 1000000000},
-		{"tb", 1000000000000},
-		{"kib", 1024},
-		{"mib", 1048576},
-		{"gib", 1073741824},
-		{"tib", 1099511627776},
+		{"kb", 1e3},
+		{"mb", 1e6},
+		{"gb", 1e9},
+		{"tb", 1e12},
+		{"kib", 1 << 10},
+		{"mib", 1 << 20},
+		{"gib", 1 << 30},
+		{"tib", 1 << 40},
 		{"b", 1}, // must be after others else it'll match on them all
 		{"", 1},  // no suffix is still a valid suffix
 	}
--- a/siac/parse_test.go
+++ b/siac/parse_test.go
@@ -1,0 +1,41 @@
+package main
+
+import (
+	"fmt"
+	"testing"
+)
+
+func TestParseSize(t *testing.T) {
+	tests := []struct {
+		in, out string
+		err     error
+	}{
+		{"1b", "1", nil},
+		{"1KB", "1000", nil},
+		{"1MB", "1000000", nil},
+		{"1GB", "1000000000", nil},
+		{"1TB", "1000000000000", nil},
+		{"1KiB", "1024", nil},
+		{"1MiB", "1048576", nil},
+		{"1GiB", "1073741824", nil},
+		{"1TiB", "1099511627776", nil},
+		{"", "", errUnableToParseSize},
+		{"123", "123", nil},
+		{"123TB", "123000000000000", nil},
+		{"123GiB", "132070244352", nil},
+		{"123BiB", "", errUnableToParseSize},
+		{"GB", "", errUnableToParseSize},
+		{"123G", "", errUnableToParseSize},
+		{"123B99", "", errUnableToParseSize},
+		{"12A3456", "", errUnableToParseSize},
+		{"1.23KB", "1230", nil},
+		{"1.234KB", "1234", nil},
+		{"1.2345KB", "1234", nil},
+	}
+	for _, test := range tests {
+		res, err := parseSize(test.in)
+		if res != test.out || err != test.err {
+			t.Error(fmt.Sprintf("parseSize(%v): expected %v %v, got %v %v", test.in, test.out, test.err, res, err))
+		}
+	}
+}
--- a/Makefile
+++ b/Makefile
@@ -63,9 +63,9 @@ clean:
 # 'test-long' runs more thorough tests which should not last more than a few
 # minutes.
 pkgs = ./api ./compatibility ./crypto ./encoding ./modules ./modules/consensus \
-       ./modules/explorer ./modules/gateway ./modules/host ./modules/hostdb \
-       ./modules/miner ./modules/renter ./modules/transactionpool \
-       ./modules/wallet ./persist ./siae ./types
+       ./modules/explorer ./modules/gateway ./modules/host ./modules/hostdb	   \
+       ./modules/miner ./modules/renter ./modules/transactionpool              \
+       ./modules/wallet  ./persist ./siac ./siae ./types
 test: clean fmt REBUILD
 	go test -short -tags='debug testing' -timeout=10s $(pkgs)
 test-v: clean fmt REBUILD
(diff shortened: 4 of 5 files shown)


DavidVorick:
Needs unit tests. Check out types/currency_test.go for some good examples on how to write unit tests for this type of PR.

I think it also makes sense to put this function (as well as the function for parsing siacoin values) into a separate 'parse.go' file. The file with the tests would be called 'parse_test.go'.

I think the examples in currency_test.go are sufficient, but if you want to learn more about go's test framework:

http://golang.org/pkg/testing/
https://blog.golang.org/examples

siac/hostcmd.go
```diff
@@ -51,6 +52,61 @@ Doing so will override the standard connectivity checks.`,
 	}
 )
 
+// parseSize converts strings of form 10GB to a size in bytes
+// If parsing fails the original value is returned unmodified
+func parseSize(value string) string {
+	var suffix string
+	var size int
```
DavidVorick: typically you declare function variables just before you use them. I'd probably put these directly above the first `if` statement.

siac/hostcmd.go
```diff
@@ -51,6 +52,61 @@ Doing so will override the standard connectivity checks.`,
 	}
 )
 
+// parseSize converts strings of form 10GB to a size in bytes
+// If parsing fails the original value is returned unmodified
+func parseSize(value string) string {
```
DavidVorick: `value` is an awkward name because it's fully generic. It's the same class of name as `data` or `input` and doesn't give you any indication of what it's supposed to be.

`strSize` is probably better.

siac/hostcmd.go
```diff
@@ -51,6 +52,61 @@ Doing so will override the standard connectivity checks.`,
...
+		if err != nil {
+			return value
+		}
+	} else {
+		// value does not appear to have a suffix, return as is
+		return value
+	}
+
+	switch strings.ToLower(suffix) {
+	case "tb":
```
DavidVorick: you can also do:

`case "tb", "TB":`

siac/hostcmd.go
bitspill: I didn't want the variables to get between the comment and the if statement so I placed it above, if you feel it's better the other way I can move it.

siac/hostcmd.go
bitspill: Agreed. I will change that. (I just pulled value since that was the variable name in the original function but doesn't make much sense out of context)

siac/hostcmd.go
bitspill: I chose ToLower so I wouldn't have to consider "Tb", "TIB", "Tib", etc. it seemed like there'd be too many variants to list explicitly and lowercase is cleaner

bitspill:
I'll start to move it to it's own file and put together some unit tests.

siac/hostcmd.go
DavidVorick: oh I missed the ToLower call

siac/hostcmd.go
DavidVorick: I would consider the variables to be a part of the `if` statement, it's not like they're actually in the way. If you move it and decide that it's too disruptive, you can leave it how it is.

bitspill:
My Go environment is hosed need a reinstall, I'm going to abuse Travis a bit then I'll squash it all down.

siac/parse.go
```diff
@@ -0,0 +1,63 @@
+package main
+
+import (
+	"fmt"
+	"strings"
+)
+
+// parseSize converts strings of form 10GB to a size in bytes
+// If parsing fails the original value is returned unmodified
```
DavidVorick: parseSize should return an error if parsing fails

DavidVorick:
I believe it's outside the scope of this PR, but at some point it would be nice to support '100 GB' as well as '100GB', but I think right now cobra interprets the space incorrectly.

DavidVorick:
Can you also move `coinUnits` (in walletcmd.go) into parse.go?

DavidVorick:
Now that siac has unit tests, you need to add `./siac` to the Makefile variable `pkgs`, which starts on line 64.

The packages are all sorted alphabetically. It's probably sufficient to put `./siac` after `./persist`

siac/parse.go
```diff
@@ -0,0 +1,93 @@
...
+		fallthrough
+	case "mib":
+		size *= 1024
+		fallthrough
+	case "kib":
+		size *= 1024
+		break
+	default:
+		// must not be a valid suffix, parse leading numbers
+		return fmt.Sprintf("%d", strSize), nil
```
DavidVorick: if it's not a valid suffix, an error should be returned

siac/parse.go
```diff
@@ -0,0 +1,93 @@
...
+	"strings"
+)
+
+// parseSize converts strings of form 10GB to a size in bytes
+// Parsing is flexible in that unnecessary trailing characters are discarded
+// If parsing fails an error is thrown
+func parseSize(strSize string) (string, error) {
+	// if the value is at least 2 or 3 characters it may have a suffix
+	// to detect a suffix a check is made that the character is outside
+	// the ascii range for numbers (48-57), letters are (65-90, 97-122)
```
DavidVorick: comments, in particular long comments (exceptions can be made for shorter comments) should have full punctuation, including capitals and periods.

siac/parse.go
```diff
@@ -0,0 +1,93 @@
...
+
+import (
+	"fmt"
+	"math/big"
+	"strings"
+)
+
+// parseSize converts strings of form 10GB to a size in bytes
+// Parsing is flexible in that unnecessary trailing characters are discarded
+// If parsing fails an error is thrown
```
DavidVorick: Docstring should have full puntuation.

Should also be wrapped to 80 characters. I'm not sure what text editor you are using, but I know vim and sublime text have a command you can use that will auto-wrap/format a block of text to 80 characters. In vim, you highlight the lines you want to wrap, and then press `gq`. I'm not sure what the command is in other text editors.

siac/parse.go
```diff
@@ -0,0 +1,90 @@
...
+import (
+	"fmt"
+	"math/big"
+	"strings"
+)
+
+// parseSize converts strings of form 10GB to a size in bytes, if it fails an error is returned.
+func parseSize(strSize string) (string, error) {
+	// If the value is at least 2 or 3 characters it may have a suffix. To verify it could be a suffix one of the
+	// characters are checked to see it is outside the ascii range for numbers (48-57), letters are (65-90, 97-122).
```
DavidVorick: Needs to be wrapped to 80 chars. This one is at 110 + tab (tabs can count for 2, 4, 8, doesn't matter)

bitspill:
`coinUnits` has been moved, Makefile alphabetized, comments adjusted (personal preference I like the comments no wider than the code below them), errors fixed (hopefully)

DavidVorick:
sorry to be so picky

DavidVorick:
Can you add a test for "12A3456"?

bitspill:
Comments should be formatted for real this time, my editor was set to 120 instead of 80.
New test added.

bitspill:
Muxadoooooooo

DavidVorick:
ah man, I should have caught this earlier.

You can't have a test named `TestGB` because the test relates explicitly to `parseSize`, and other tests may want to use a similar namespace.

You can do a find+replace for "Test" -> "TestParseSize" and that should finish it up. Everything else looks good to me, I'll wait to merge until @lukechampine gives his approval.

siac/parse_test.go
```diff
@@ -0,0 +1,141 @@
...
+	b, err := parseSize(a)
+	if b != "1024" && err == nil {
+		t.Error("ParseSize 1KiB failed")
+	}
+}
+
+func Test12A3456(t *testing.T) {
+	a := "12A3456"
+	b, err := parseSize(a)
+	if b != "" || err != nil {
```
DavidVorick: shouldn't err be non-nil?

siac/parse.go
```diff
@@ -0,0 +1,94 @@
...
+			return "", errors.New("unable to parse size")
+		}
+	} else if len(strSize) > 2 && strSize[len(strSize)-2] > 57 {
+		suffix = strSize[len(strSize)-2:] // KB, MB, GB, TB
+		_, err := fmt.Sscan(strSize[:len(strSize)-2], &size)
+		if err != nil {
+			return "", errors.New("unable to parse size")
+		}
+	} else {
+		return fmt.Sprintf("%d", strSize), nil
```
DavidVorick: error should be returned if %d is parsed to ""

siac/parse.go
```diff
@@ -0,0 +1,94 @@
...
+		size *= 1024
+		fallthrough
+	case "kib":
+		size *= 1024
+		break
+	default:
+		// must not be a valid suffix
+		return "", errors.New("unable to parse size")
+	}
+	return fmt.Sprintf("%d", size), nil
```
DavidVorick: ditto here

siac/parse_test.go
bitspill: Yes, I actually did that on emptystring test and short prefix. Fixed in the next push.

DavidVorick:
np, thanks for doing this

siac/parse.go
```diff
@@ -0,0 +1,104 @@
+package main
+
+import (
+	"errors"
+	"fmt"
+	"math/big"
+	"strings"
+)
+
+var ErrUnableToParseSize = errors.New("unable to parse size")
```
DavidVorick: we prefer lower-case (project is inconsistent about this because we only started enforcing the convention recently).

`errUnableToParseSize`.

Our general approach is:

1. return errors.New
if we need to check for a specific error (in testing or otherwise):
2. return errUnexportedErr (no docstring needed on error)
if external packages need to check for a specific error
3. return ErrExportedErr (docstring needed on error)

siac/parse_test.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+}
+
+func TestParseSpace12A3456(t *testing.T) {
+	a := "12A3456"
+	b, err := parseSize(a)
+	if b != "" || err == nil {
+		t.Error("ParseSize 12A3456 failed")
+	}
+}
```
lukechampine: these need not be separate tests. Write something of the form:
```go
func TestParseSize(t *testing.T) {
    tests := []struct{
        in, out string
        err error
    }{
        {"1MiB", "1048576", nil},
        {"foo", "", errParseSize},
    }
    for _, test := range tests {
        res, err := parseSize(test.in)
        if res != test.out || err != test.err {
            t.Error("parseSize(%v): expected %v %v, got %v %v", test.in, test.out, test.err, res, err)
        }
    }
}
```

siac/parse.go
```diff
@@ -0,0 +1,104 @@
...
+			return "", ErrUnableToParseSize
+		}
+	} else if len(strSize) > 2 && strSize[len(strSize)-2] > 57 {
+		suffix = strSize[len(strSize)-2:] // KB, MB, GB, TB
+		_, err := fmt.Sscan(strSize[:len(strSize)-2], &size)
+		if err != nil {
+			return "", ErrUnableToParseSize
+		}
+	} else {
+		ret := fmt.Sprintf("%d", strSize)
```
lukechampine: this format string is invalid for its input.

I would rewrite this function to closely parallel `coinUnits`. Note that `coinUnits` allows for decimal values, like 12.345SC. The logic will be almost identical except for the units. (and you'll need a separate loop for the IEC units)

siac/parse.go
```diff
@@ -0,0 +1,104 @@
...
+		fallthrough
+	case "gb":
+		size *= 1000
+		fallthrough
+	case "mb":
+		size *= 1000
+		fallthrough
+	case "kb":
+		size *= 1000
+		break
```
lukechampine: this is a clever use of `fallthrough`, but the conversion can be done more compactly.

bitspill:
Alright, there we go. I got my go environment re-installed and think I finished this off. 

For fractional sizes such as 1.2345 KB I elected to truncate at the byte level so the result would be 1234 bytes.

Let me know if I forgot anything or did something silly again.

lukechampine:
looks good! I like the use of an anonymous struct to pair suffixes with multipliers. :+1:

I'm not sure about the truncation; maybe we ought to return an error in that case, since it may cause surprising behavior. It seems unlikely to me that anyone would care about +/- 1 byte, but you never know.

siac/parse.go
```diff
@@ -0,0 +1,77 @@
...
+		multiplier int64
+	}{
+		{"kb", 1000},
+		{"mb", 1000000},
+		{"gb", 1000000000},
+		{"tb", 1000000000000},
+		{"kib", 1024},
+		{"mib", 1048576},
+		{"gib", 1073741824},
+		{"tib", 1099511627776},
```
lukechampine: can you use left-shift notation here? i.e. `1 << 40`

DavidVorick: seconded. It's much easier to read `1e9` or `460e6` than it is to try and count zeroes or remember what 10^7 is.

Similarly, `1 << 30` is friendlier than `1073741824`

DavidVorick:
I think it'd be cool to have this in 0.4.2, which means merging it now. We can do a second PR to fix the number stuff since everything else seems ready.

siac/parse.go
bitspill: Ah, yes. I forgot about those.

I tried 10^3, 10^6, 2^10, 2^20, but that's bitwise xor in go so didn't work

siac/parse.go
```diff
@@ -0,0 +1,77 @@
...
+			// r must be an integer at this point
+			if !r.IsInt() {
+				return "", errors.New("non-integer number of hastings")
+			}
+			return r.RatString(), nil
+		}
+	}
+	// check for hastings separately
+	if strings.HasSuffix(amount, "H") {
+		return strings.TrimSuffix(amount, "H"), nil
```
DavidVorick: Doesn't need to be handled in this PR since it's an issue that's existed for a while, but it looks like

`164363zzidididH` would be considered a valid string by this function. A test case should be added and the logic should be corrected. I'll probably open a separate issue for this.

siac/hostcmd.go
```diff
@@ -62,6 +62,14 @@ func hostconfigcmd(param, value string) {
 		p.Mul(p, big.NewRat(1e24/1e9, 4320))
 		value = new(big.Int).Div(p.Num(), p.Denom()).String()
 	}
+	// parse sizes of form 10GB, 10TB, 1TiB etc
+	if param == "totalstorage" || param == "minfilesize" || param == "maxfilesize" {
+		var err error // must be pre-declared because value is
```
bitspill: This is actually not true, and an unnecessary line, but at this point I don't know if I should make another commit just for that

DavidVorick: scoping in go can get annoying sometimes.

This PR is good enough, you're never really done polishing code and if you don't draw the line you'll only ever write 1 feature :P

siac/parse.go
lukechampine: yeah. Would be easy to try a `SetString` after trimming the suffix.
Of course, the string you gave would be rejected by the API anyway, but it doesn't hurt to validate on both sides.

PR #779, opened 2015-09-04
change xc command

DavidVorick:
The releases produced are now slightly more friendly to the UI build
process.


lukechampine:
looks like all that changed here was changing `pr` to `br`, is that right?

DavidVorick:
-bc now sorted alphabetically
-pv=v0.4.1
-br already existed, -pr got removed entirely

I think that's all the changes

PR #780, opened 2015-09-05
Better peer culling

lukechampine:
`makeOutboundConnections` has been split into two functions, `peerManager` and `nodeManager`.

`peerManager` keeps the gateway well-connected by forming new connections to peers until reaching 8 total connections.

`nodeManager` keeps the node list healthy by pinging nodes and aggressively culling those that are unresponsive.

Tests pass, and coverage is okay, but these functions really ought to be tested in the wild. My guess is that it will take longer to get peers now. `makeOutboundConnections` tried to form up to 100 connections before sleeping; `peerManager` only tries to form one. However, these changes should (eventually) improve the average node list, such that fewer connection attempts are needed.

### DIFF
--- a/modules/gateway/nodes.go
+++ b/modules/gateway/nodes.go
@@ -113,11 +113,11 @@ func (g *Gateway) sendAddress(conn modules.PeerConn) error {
 	return encoding.WriteObject(conn, g.Address())
 }
 
-// nodeManager tries to keep the Gateway's node list healthy. As long as the
-// Gateway has fewer than minNodeListSize nodes, it asks a random peer for
-// more nodes. It also continually pings nodes in order to establish their
+// threadedNodeManager tries to keep the Gateway's node list healthy. As long
+// as the Gateway has fewer than minNodeListSize nodes, it asks a random peer
+// for more nodes. It also continually pings nodes in order to establish their
 // connectivity. Unresponsive nodes are aggressively removed.
-func (g *Gateway) nodeManager() {
+func (g *Gateway) threadedNodeManager() {
 	for {
 		time.Sleep(5 * time.Second)
 
--- a/modules/gateway/nodes.go
+++ b/modules/gateway/nodes.go
@@ -144,18 +144,16 @@ func (g *Gateway) nodeManager() {
 
 		// try to connect
 		conn, err := net.DialTimeout("tcp", string(node), dialTimeout)
-		if netErr, ok := err.(net.Error); ok && netErr.Timeout() {
+		if err != nil {
 			id = g.mu.Lock()
 			g.removeNode(node)
 			g.save()
 			g.mu.Unlock(id)
-		} else if err != nil {
 			continue
 		}
 		// if connection succeeds, supply an unacceptable version to ensure
 		// they won't try to add us as a peer
 		encoding.WriteObject(conn, "0.0.0")
 		conn.Close()
-
 	}
 }
--- a/modules/gateway/nodes.go
+++ b/modules/gateway/nodes.go
@@ -155,5 +155,8 @@ func (g *Gateway) nodeManager() {
 		// they won't try to add us as a peer
 		encoding.WriteObject(conn, "0.0.0")
 		conn.Close()
+		// sleep for an extra 10 minutes after success; we don't want to spam
+		// connectable nodes
+		time.Sleep(10 * time.Minute)
 	}
 }
--- a/modules/gateway/nodes.go
+++ b/modules/gateway/nodes.go
@@ -3,6 +3,7 @@ package gateway
 import (
 	"errors"
 	"net"
+	"time"
 
 	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/encoding"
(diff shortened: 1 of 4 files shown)


modules/gateway/nodes.go
```diff
@@ -111,3 +112,48 @@ func (g *Gateway) sendAddress(conn modules.PeerConn) error {
 	}
 	return encoding.WriteObject(conn, g.Address())
 }
+
+// nodeManager tries to keep the Gateway's node list healthy. As long as the
+// Gateway has fewer than minNodeListSize nodes, it asks a random peer for
+// more nodes. It also continually pings nodes in order to establish their
+// connectivity. Unresponsive nodes are aggressively removed.
+func (g *Gateway) nodeManager() {
```
DavidVorick: `threadedNodeManager`

modules/gateway/nodes.go
```diff
@@ -111,3 +112,48 @@ func (g *Gateway) sendAddress(conn modules.PeerConn) error {
...
+	for {
+		time.Sleep(5 * time.Second)
```
DavidVorick: 5 seconds is probably too aggressive.

5 minutes?

modules/gateway/peers.go
```diff
@@ -233,37 +233,29 @@ func (g *Gateway) Disconnect(addr modules.NetAddress) error {
...
+		if err != nil {
+			// can't do much until we have nodes
+			continue
+		}
+
+		if numPeers >= wellConnectedThreshold {
+			sleepTime = time.Minute
+		} else {
+			sleepTime = 5 * time.Second
 		}
```
DavidVorick: I like this conditional

DavidVorick: seems weird to me though t hat you would have it at the end, and then have the sleep at the top.

Couldn't you just put a

```go
if np > wct {
    time.Sleep(60)
} else {
    time.Sleep(5)
}
```
right where the first sleep appears? then you can also drop the `sleepTime` declaration at the top

modules/gateway/peers.go
```diff
@@ -233,37 +233,29 @@ func (g *Gateway) Disconnect(addr modules.NetAddress) error {
...
 }
 
-// makeOutboundConnections tries to keep the Gateway well-connected. As long
-// as the Gateway is not well-connected, it tries to add random nodes as
-// peers. It sleeps when the Gateway becomes well-connected, or it has tried
-// more than 100 nodes.
-func (g *Gateway) makeOutboundConnections() {
+// peerManager tries to keep the Gateway well-connected. As long as the
+// Gateway is not well-connected, it tries to connect to random nodes.
+func (g *Gateway) peerManager() {
```
DavidVorick: `threadedPeerManager`

modules/gateway/peers.go
```diff
@@ -233,37 +233,29 @@ func (g *Gateway) Disconnect(addr modules.NetAddress) error {
...
+		}
+
+		if numPeers >= wellConnectedThreshold {
+			sleepTime = time.Minute
+		} else {
+			sleepTime = 5 * time.Second
 		}
-		time.Sleep(5 * time.Second)
+
+		g.Connect(addr)
```
DavidVorick: how do you feel about

`go g.Connect(addr)` ?

Then, if a peer takes 2 minutes to respond, you are still making other connections. Worst case you end up with 40 or so peers because they are all taking 2 minutes before being added. But that's still well below the 128 limit, and a highly unlikely in a real situation. Even in the black swan case it's still okay.

modules/gateway/peers.go
lukechampine: yeah, that seems better. I think originally I was planning some sort of exponential back-off, like:
```go
if numPeers >= wellConnectedThreshold {
    sleepTime *= 2
} else {
    sleepTime = 5 * time.Second
}
```

lukechampine:
In live testing, this performed quite well. It reached 8 peers within a few minutes (lots of failed connections), and is now culling the node list.
However, a 2 minute timeout severely limits the effectiveness of `nodeManager`. Perhaps, like `peerManager`, it should spawn connections in a goroutine. Or it could use a reduced timeout.

modules/gateway/nodes.go
DavidVorick: please change this function name to follow conventions

modules/gateway/peers.go
DavidVorick: please change this function name to follow conventions

modules/gateway/peers.go
```diff
@@ -233,35 +233,26 @@ func (g *Gateway) Disconnect(addr modules.NetAddress) error {
...
-		addr, err := g.randomPeer()
+		numPeers := len(g.peers)
+		addr, err := g.randomNode()
 		g.mu.RUnlock(id)
-		if build.Release != "testing" && err == nil && numNodes < minNodeListLen {
-			g.RPC(addr, "ShareNodes", g.requestNodes)
+		if numPeers >= wellConnectedThreshold {
+			time.Sleep(5 * time.Minute)
+			continue
+		}
```
DavidVorick: should you be grabbing the node before you sleep?

lukechampine: the node isn't used in this branch, due to the `continue`.
It might be clearer to grab the node after the sleep, but that would require two readlocks.

DavidVorick: can you add a comment explaining that?

PR #781, opened 2015-09-05
Moving addSignatures out of transactionbuilder

mtlynch:
Creating a separate file for the addSignatures function and making the
signature adding function a function pointer in transactionBuilder. This will
allow us to more easily test transactionBuilder, as the complex logic of
addSignatures can be mocked out.

This also adds a very limited unit test for addSignatures.

### DIFF
--- a/modules/wallet/signtransaction.go
+++ b/modules/wallet/signtransaction.go
@@ -1,0 +1,64 @@
+package wallet
+
+import (
+	"bytes"
+
+	"github.com/NebulousLabs/Sia/crypto"
+	"github.com/NebulousLabs/Sia/types"
+)
+
+type hashSigner func(crypto.Hash, crypto.SecretKey) (crypto.Signature, error)
+
+type signatureAdder func(*types.Transaction, types.CoveredFields, types.UnlockConditions, crypto.Hash, spendableKey) error
+
+// addSignatures will sign a transaction using a spendable key, with support
+// for multisig spendable keys. Because of the restricted input, the function
+// is compatible with both siacoin inputs and siafund inputs.
+func addSignatures(txn *types.Transaction, cf types.CoveredFields, uc types.UnlockConditions, parentID crypto.Hash, spendKey spendableKey) error {
+	return addSignaturesWithHashSigner(txn, cf, uc, parentID, spendKey, crypto.SignHash)
+}
+
+// addSignaturesWithHashSigner provides the same functionality as
+// addSignatures, but allows the caller to specify the hash signing function.
+func addSignaturesWithHashSigner(txn *types.Transaction, cf types.CoveredFields, uc types.UnlockConditions, parentID crypto.Hash, spendKey spendableKey, signHash hashSigner) error {
+	// Try to find the matching secret key for each public key - some public
+	// keys may not have a match. Some secret keys may be used multiple times,
+	// which is why public keys are used as the outer loop.
+	totalSignatures := uint64(0)
+	for i, siaPubKey := range uc.PublicKeys {
+		// Search for the matching secret key to the public key.
+		for j := range spendKey.SecretKeys {
+			pubKey := spendKey.SecretKeys[j].PublicKey()
+			if bytes.Compare(siaPubKey.Key, pubKey[:]) != 0 {
+				continue
+			}
+
+			// Found the right secret key, add a signature.
+			sig := types.TransactionSignature{
+				ParentID:       parentID,
+				CoveredFields:  cf,
+				PublicKeyIndex: uint64(i),
+			}
+			txn.TransactionSignatures = append(txn.TransactionSignatures, sig)
+			sigIndex := len(txn.TransactionSignatures) - 1
+			sigHash := txn.SigHash(sigIndex)
+			encodedSig, err := signHash(sigHash, spendKey.SecretKeys[j])
+			if err != nil {
+				return err
+			}
+			txn.TransactionSignatures[sigIndex].Signature = encodedSig[:]
+
+			// Count that the signature has been added, and break out of the
+			// secret key loop.
+			totalSignatures++
+			break
+		}
+
+		// If there are enough signatures to satisfy the unlock conditions,
+		// break out of the outer loop.
+		if totalSignatures == uc.SignaturesRequired {
+			break
+		}
+	}
+	return nil
+}
--- a/modules/wallet/signtransaction_test.go
+++ b/modules/wallet/signtransaction_test.go
@@ -1,0 +1,56 @@
+package wallet
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+
+	"github.com/NebulousLabs/Sia/crypto"
+	"github.com/NebulousLabs/Sia/types"
+)
+
+// mockHashSigner serves as a mock replacement for crypto.SignHash.
+type mockHashSigner struct {
+	mock.Mock
+}
+
+func (hs *mockHashSigner) SignHash(hash crypto.Hash, key crypto.SecretKey) (sig crypto.Signature, err error) {
+	args := hs.Called(hash, key)
+	return args.Get(0).(crypto.Signature), args.Error(1)
+}
+
+func generateMockSignatureKeys() (sk crypto.SecretKey, pk crypto.PublicKey) {
+	sk[0] = 0x0a
+	sk[32] = 0x0b
+	return sk, sk.PublicKey()
+}
+
+func publicKeyToSiaPublicKey(pk crypto.PublicKey) (spk types.SiaPublicKey) {
+	spk.Key = pk[:]
+	return spk
+}
+
+func TestAddSignaturesOnEmptyTransaction(t *testing.T) {
+	sk, pk := generateMockSignatureKeys()
+	txn := types.Transaction{}
+	cf := types.CoveredFields{}
+	uc := types.UnlockConditions{}
+	uc.PublicKeys = append(uc.PublicKeys, publicKeyToSiaPublicKey(pk))
+	parentID := crypto.Hash{}
+	spendKey := spendableKey{}
+	spendKey.SecretKeys = append(spendKey.SecretKeys, sk)
+
+	// Create a mock hash signer that always returns a signature of
+	// {0x0c, 0x00, 0x00, ...}
+	hs := new(mockHashSigner)
+	var mockSignature crypto.Signature
+	mockSignature[0] = 0x0c
+	hs.On("SignHash", mock.Anything, sk).Return(mockSignature, nil)
+
+	addSignaturesWithHashSigner(&txn, cf, uc, parentID, spendKey, hs.SignHash)
+
+	assert.Equal(t, 1, len(txn.TransactionSignatures))
+	assert.Equal(t, mockSignature[:], txn.TransactionSignatures[0].Signature)
+	hs.AssertExpectations(t)
+}
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,7 @@ dependencies:
 	go get -u github.com/laher/goxc
 	go get -u github.com/spf13/cobra
 	go get -u github.com/stretchr/graceful
+	go get -u github.com/stretchr/testify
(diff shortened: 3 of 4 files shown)


mtlynch:
The idea here is that constructing `Transaction` objects is complex, signing transactions is complex, so it's hard to test `transactionBuilder` when it's tightly coupled with `addSignatures` because there are so many inputs, outputs, and external dependencies.

Splitting out `addSignatures` should simplify testing of `transactionBuilder` because instead of calling `FundSiacoins` and verifying that the resulting transaction was both constructed and signed properly, we can mock out `transactionBuilder.addSignatures` and just test whether the transaction was constructed properly and that `transactionBuilder.addSignatures` was called with the correct parameters to get it signed.

DavidVorick:
(haven't read PR yet)

did you mean to pull this in to master, or into tbuilder?

mtlynch:
>did you mean to pull this in to master, or into tbuilder?

Into master. I'll want to later merge these into tbuilder as well, but they should go to master directly.

modules/wallet/signtransaction.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"bytes"
+
+	"github.com/NebulousLabs/Sia/crypto"
+	"github.com/NebulousLabs/Sia/types"
+)
+
+type hashSigner func(crypto.Hash, crypto.SecretKey) (crypto.Signature, error)
+
+type signatureAdder func(*types.Transaction, types.CoveredFields, types.UnlockConditions, crypto.Hash, spendableKey) error
```
DavidVorick: I don't understand the purpose of this object

modules/wallet/signtransaction.go
```diff
@@ -0,0 +1,64 @@
+package wallet
+
+import (
+	"bytes"
+
+	"github.com/NebulousLabs/Sia/crypto"
+	"github.com/NebulousLabs/Sia/types"
+)
+
+type hashSigner func(crypto.Hash, crypto.SecretKey) (crypto.Signature, error)
```
DavidVorick: I don't understand the purpose of this object either

DavidVorick: I'm guess that you are trying to avoid a dependency on crypto.SignHash... but why? It seems like unneeded extra complexity to me.

modules/wallet/signtransaction_test.go
```diff
@@ -0,0 +1,56 @@
...
+	hs := new(mockHashSigner)
+	var mockSignature crypto.Signature
+	mockSignature[0] = 0x0c
+	hs.On("SignHash", mock.Anything, sk).Return(mockSignature, nil)
+
+	addSignaturesWithHashSigner(&txn, cf, uc, parentID, spendKey, hs.SignHash)
+
+	assert.Equal(t, 1, len(txn.TransactionSignatures))
+	assert.Equal(t, mockSignature[:], txn.TransactionSignatures[0].Signature)
+	hs.AssertExpectations(t)
```
DavidVorick: I don't really see the benefit to this test. You added a lot of complexity to set it up, when all you are verifying is two lines?

I'm just not understanding the benefit to these complicated changes in our codebase. The layers of abstraction add a lot of overhead to understanding the system, and don't seem to come with a whole lot of payoff. Seems like it would have been just fine to use the real signature libraries implemented by the crypto package.

mtlynch:
Thanks for reviewing!

>I'm just not understanding the benefit to these complicated changes in our codebase. The layers of abstraction add a lot of overhead to understanding the system, and don't seem to come with a whole lot of payoff. Seems like it would have been just fine to use the real signature libraries implemented by the crypto package.

I think your question is equivalent to "why are unit tests worth the added complexity of indirect calls to dependencies?" Because a test that uses the `crypto` package directly would be an integration test. Here are two good answers to this question in general:
* [On StackOverflow](https://programmers.stackexchange.com/questions/65477/why-is-it-important-that-a-unit-test-not-test-dependencies)
* [On the Google testing blog](http://googletesting.blogspot.com/2015/04/just-say-no-to-more-end-to-end-tests.html) - Full disclosure, I work full-time at Google as a developer and I've seen very good results with our unit-test focused testing strategy

I'll also try to explain why I think it would be highly beneficial to the Sia codebase specifically. At a high level:
* Unit tests vastly reduce the scope of defects when a test fails.
 * Currently if a `wallet` test fails, there are so many things that could be responsible: the persist code, the miner code, the signature code, etc.
 * With unit tests, if the transaction signing unit test code fails, it can only be a problem in a) the transaction signing code or b) the unit test itself because those are the only things in the universe of possibility.
* It makes it easier for new developers to understand the codebase and contribute to it
 * Looking at the [`wallet` code](https://github.com/NebulousLabs/Sia/blob/fe95fb7149987d6e2985f3be1e5b9b23f1f8d75d/modules/wallet/wallet.go#L39) it has dependencies on `WalletPersist`, `ConsensusSet`, `TransactionPool`, and `Seed`. I'm a developer who wants to understand how `Wallet` interacts with the rest of the codebase and this seems very complex and difficult to wrap my head around.
 * Unit testing encourages loose coupling and well-defined interfaces. Part of what makes `Wallet` difficult to understand is that it's using very loosely-defined interfaces.
    * For example, it depends on the `TransactionPool` interface, but only uses a small subset of that interface (I think it only needs `TransactionPoolSubscribe` and `AcceptTransactionSet`). It's unlikely this would occur if it was unit tested because a developer wouldn't want bother mocking out the entire interface for just two methods and would instead define a more explicit interface between `Wallet` and `TransactionPool`
* It allows a very fast write-build-test loop.
 * Running the `wallet` package testing suite on my machine takes ~3s right now and the wallet tests are not very exhaustive. This will not scale well as `wallet`s dependent packages become more complex and we add tests for additional scenarios. We will have to address this by either writing tests that are not exhaustive or running tests less frequently because nobody wants to run a 90 second test after every change.

> You added a lot of complexity to set it up, when all you are verifying is two lines?

This is the first unit test for `addSignatures` and it is admittedly very simple. It alone does not provide a ton of validation, but it does verify one thing that I don't think any existing tests do, which is that `crypto.SignHash` is being called correctly (i.e. that the transactions are actually signed correctly).

I'm travelling at the moment, so I'm not able to test this, but I suspect that if I changed `addSignatures` to comment out the call to `crypto.SignHash` none of the existing tests would fail. The new test `TestAddSignaturesOnEmptyTransaction` would correctly flag an error because it would detect that the expectations on the mock hash signer were never met.

The options to do this using integration tests are:
* Hardcode the correct signature bytes into the test and make sure the production code returns the matching value, (e.g. `if signature[:] != {0xf3, 0xd9, 0x03, 0x82, ... [snip], 0x62} { t.Errorf("Signature does not match!") }`.
 * This is very brittle because if we ever change the signature implementation (or change the covered fields) we have to recalculate all the correct hashes and update all the tests.
* Relax correctness requirements and focus on consistency.
 * This seems to be what the codebase currently does in most cases, for example [seed_test.go](https://github.com/NebulousLabs/Sia/blob/fe95fb7149987d6e2985f3be1e5b9b23f1f8d75d/modules/wallet/seed_test.go) doesn't seem to be able to test whether the returned seed is actually initialized, but it just makes sure that the seed returned by `Encrypt` and the seed returned by `PrimarySeed` are consistent. This seems dangerous to me as I don't see a way for us to detect if we've introduced a bug that causes the code to generate seeds incorrectly or if seed initialization is skipped entirely.

It's also not possible with integration tests to test something like `crypto.SignHash` returning an error because it's so difficult for integration tests to trigger a failure at such a low level, but it is important to ensure that each component can gracefully handle failures from its dependencies. (`crypto.SignHash` is maybe not an ideal example because it *always* succeeds but this applies in general to calls that can fail in ways that are not easy to trigger.)

tl; dr - I think unit tests are very useful and can provide important benefits to Sia that no other type of testing can.

DavidVorick:
Thanks for taking the time to write all of that out. I appreciate that you are willing to fight me on the best way to organize/manage Sia. Hopefully Sia benefits.

I'm still digesting your comments, but you've made some good points. All three of the bullet points you made are things that I would like to improve about our existing codebase, and I can see how unit tests can help us move towards those ideals.

> I'm travelling at the moment, so I'm not able to test this, but I suspect that if I changed addSignatures to comment out the call to crypto.SignHash none of the existing tests would fail.

They certainly do fail :), a large number of them too. The wallet package panics on `TestIntegrationUserSuppliedEncryption`. It should also fail on some of the later tests. `TestSendSiacoins` should also fail.

> It's also not possible with integration tests to test something like crypto.SignHash returning an error because it's so difficult for integration tests to trigger a failure at such a low level, but it is important to ensure that each component can gracefully handle failures from its dependencies.

That's a great point. One big example is `_, err := rand.Read(data[:])`. We use that all over the place, but it's pretty much impossible for us to trigger an error from that call.

I will try to come back to this in a day or two with some refreshed thoughts.

DavidVorick:
wanted to let you know that you've convinced me of the value of separating dependencies the project, so that more local unit tests can be performed.

I'm still not sure that you've organized things in the best way, but when I get a breather (things got crazy when the price jumped 3.5x in one day) I'll work with you to try and come up with a system that can be used to start splitting dependencies out of modules and other parts of the project in an organized and consistent way.

Thanks for walking me through the value of unit testing, and teaching me more about testing in general.

mtlynch:
Great, that's good to hear. I'm glad you're open to challenging your existing dev practices. Let me know when you're ready to talk about breaking up dependencies. Congrats on the price jump!

DavidVorick:
I think the wallet might be too difficult of a place to start experimenting with this stuff - the wallet has a ton of dependencies and I'm really not sure how to start organizing them. It's very different from any development I've done in the past.

Perhaps we could start with the crypto package? Notable dependencies are crypto/rand, io, blake2b, encoding, and merkletree. Most functions only have 1 place to start.

I was wondering if we should have a file like `dependencies.go`, or if we should keep dependencies in the same file where they are used. You run into trouble when multiple files or functions share a dependency (such as a hasher).

I was also thinking that maybe the stateful object of the package could house the dependencies. For example, when you call `New`, it could add all the dependencies in their full form. Then, after calling object.Mock, you could write code to replace the dependencies with mocked versions. Mock would only be available in an _test file, so that the mock function does not appear in godoc or in production. This method sort of falls apart for packages that don't have a stateful object (like the crypto package).

The package that needs the most attention from a testing perspective is the consensus package. But that's also got a massive amount of dependencies, and would probably be a huge effort to isolate everything.

I'm still not sure the best way to approach it.

mtlynch:
I'm traveling for the next week, but I will try to take a pass at breaking up dependencies in the crypto package if I have some downtime or when I get back.

modules/wallet/signtransaction_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/NebulousLabs/Sia/crypto"
+	"github.com/NebulousLabs/Sia/types"
+)
+
+// mockHashSigner serves as a mock replacement for crypto.SignHash.
+type mockHashSigner struct {
+	mock.Mock
+}
+
+func (hs *mockHashSigner) SignHash(hash crypto.Hash, key crypto.SecretKey) (sig crypto.Signature, err error) {
```
DavidVorick: looks like the error returned by the SignHash function is actually useless. Given that signing a hash can't actually return an error, mocking should not be required. I can make a PR that changes the function definition of SignHash.

mtlynch: The mock serves two purposes:

1. Ensure that the dependency is being called in the expected fashion (e.g. called with correct parameters, called the correct number of times when that's relevant).
1. Ensure that the code behaves in expected fashion, given different results from the dependency (e.g. succeeds vs. fails).

Getting rid of the error parameter on `SignHash` means that (2) becomes unnecessary but (1) remains. For this case, I feel like (1) is the one we care about because otherwise it's difficult to test that `SignHash` is being called with the correct set of values to sign.

DavidVorick:
closing because I believe this is the wrong place to be focusing time currently. A lot of good discussion came from this PR, and a bunch of subsequent mocking PRs have followed. We mock code more frequently than we used to (though perhaps not as much as MTL might believe is optimal).

Thanks for this contribution, even though it's not getting merged.

PR #782, opened 2015-09-06
Hostdb

DavidVorick:
1. Proper, tested blocking of local hosts in hostdb
2. added a logger to the renter, and reorganized the persist stuff a tiny bit. Logger doesn't get used at all yet.


modules/netaddress.go
```diff
@@ -0,0 +1,81 @@
...
+	host := na.RemovePort()
+	// Check if the host is a valid ip address.
+	if net.ParseIP(host) != nil {
+		return true
+	}
+
+	// A regex pulled from
+	// http://stackoverflow.com/questions/106179/regular-expression-to-match-dns-hostname-or-ip-address
+	// to check for a valid hostname.
+	regHostname, err := regexp.Compile("^([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\\-]{0,61}[a-zA-Z0-9])(\\.([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\\-]{0,61}[a-zA-Z0-9]))*$")
```
lukechampine: o_o

I don't think we need to be validating hostnames like this. We already have a fool-proof test of an address's validity: we try to connect to it. For our purposes, I don't see why an invalid address needs to be treated any differently from a valid address that never responds to pings.

modules/netaddress.go
```diff
@@ -0,0 +1,81 @@
...
+	_, port, _ := net.SplitHostPort(string(na))
+	return port
+}
+
+// RemovePort removes the port from a NetAddress. Due to error checking, the
+// behavior is different from 'Host()' for IPAddresses that don't have a port.
+// This function should probably be merged with Host(), but I wasn't sure if
+// portions of the gateway relied on getting 'nil' if no port was present in
+// the NetAddress.
+func (na NetAddress) RemovePort() string {
```
lukechampine: as discussed, I think it's fine to merge this with `Host()`

modules/netaddress.go
DavidVorick: if you can prove that a host is invalid with a simple regex, is there any reason to need to ping it?

I suppose I should be more aware of the time complexity of running this regex, but presumably it's a lot faster than trying to ping someone.

lukechampine: what I meant was, we're going to ping it at some point anyway. So it's okay if an unconnectable address makes it into the hostdb or the node list, because it'll be pruned anyway.

lukechampine:
Should add `""` as a test case

PR #786, opened 2015-09-09
Hostdb Improvements

DavidVorick:
A logger was added, so far just a single logging statment.

A severe bug in the hostdb construction was fixed involving pointers.

### DIFF
--- a/modules/hostdb/hostdb.go
+++ b/modules/hostdb/hostdb.go
@@ -8,6 +8,7 @@ package hostdb
 
 import (
 	"errors"
+	"log"
 
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/modules/consensus"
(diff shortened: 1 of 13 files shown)


modules/hostdb/hostdb.go
```diff
@@ -26,16 +23,17 @@ const (
...
+	errNilConsensusSet = errors.New("consensus set cannot be nil")
 )
 
 // The HostDB is a database of potential hosts. It assigns a weight to each
 // host based on their hosting parameters, and then can select hosts at random
 // for uploading files.
 type HostDB struct {
-	consensusSet *consensus.ConsensusSet
-	gateway      modules.Gateway
+	consensusSet    *consensus.ConsensusSet
```
lukechampine: `modules.ConsensusSet`?

lukechampine:
looks good

modules/hostdb/hostdb.go
DavidVorick: oh yeah might as well fix that

DavidVorick:
ok, should be all better

PR #787, opened 2015-09-09
pricing management tweaks

DavidVorick:
branch name is not really relevant. There were multiple mistakes in the pricing process, all the way from estimating the price, to doing the upload, to selecting hosts.

This PR makes everything more sane, and makes the pricing actually match.

### DIFF
--- a/modules/host/host.go
+++ b/modules/host/host.go
@@ -186,11 +186,13 @@ func (h *Host) Info() modules.HostInfo {
 	}
 
 	// Calculate estimated competition (reported in per GB per month). Price
-	// calculated by taking the average of 8 ranomly selected weighted hosts.
+	// calculated by taking the average of hosts 8-15.
 	var averagePrice types.Currency
-	// TODO: 8 is the sample size - to be made a constant.
-	hosts := h.hostdb.RandomHosts(8)
-	for _, host := range hosts {
+	hosts := h.hostdb.RandomHosts(15)
+	for i, host := range hosts {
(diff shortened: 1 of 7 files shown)


modules/host/host.go
```diff
@@ -186,11 +186,13 @@ func (h *Host) Info() modules.HostInfo {
...
+	// calculated by taking the average of hosts 8-15.
 	var averagePrice types.Currency
-	// TODO: 8 is the sample size - to be made a constant.
-	hosts := h.hostdb.RandomHosts(8)
-	for _, host := range hosts {
+	hosts := h.hostdb.RandomHosts(15)
+	for i, host := range hosts {
+		if i < 8 {
+			continue
+		}
```
lukechampine: interesting -- can you explain this? Looks like if `RandomHosts` returns less than 8 hosts, the avgPrice will be zero.
also, if you check the length before the loop, you could write:
```go
for _, host := range hosts[8:] {
    averagePrice = averagePrice.Add(host.Price)
}
```
which seems cleaner

DavidVorick: The cheapest 8 are outliers, and not really important to being competitive. If you want to be both profitable and competitive, you want to be competing with the people closer to the edge.

Average price will be 0 if there are less than 8 hosts, it's a downside to this algorithm.

lukechampine: hmm. Maybe (eventually) this should be more like a median than mean. I imagine over time our "average" function will get more sophisticated.

PR #793, opened 2015-09-12
Refactoring user agent check and whitelisting Sia-Explorer agent

mtlynch:


mtlynch:
See corresponding change in Sia-Explorer: https://github.com/NebulousLabs/Sia-Explorer/pull/24

lukechampine:
No need for `Sia-Explorer`; it's fine to just use `Sia-Agent`. (`Sia-Miner` is preserved for compatibility reasons, but will probably be removed eventually)

PR #794, opened 2015-09-14
Reader/Writer encoding interfaces

lukechampine:
`MarshalSia` and `UnmarshalSia` now take an `io.Writer` and `io.Reader`, respectively. This makes them better aligned with `Encode` and `Decode`. More importantly, it removes the need for all manually-marshaled data to be prefaced with a length prefix. This allows us to start writing `MarshalSia` methods for types where speed is important, e.g. `types.Block` (implemented here). Previously, doing so would have broken compatibility due to the length prefix.

After implementing `(Un)MarshalSia` methods for `types.Block`, non-rigorous benchmarking reveals the following speedups when encoding and decoding an empty block:
```
Encode benchmark: 38.70 MB/s -> 47.87 MB/s (23.7% faster)
Decode benchmark: 41.69 MB/s -> 55.80 MB/s (33.9% faster)
```
I suspect this speedup will only be magnified when we start manually marshaling transactions, and will be more noticeable with blocks that contain real data.

### DIFF
--- a/encoding/marshal.go
+++ b/encoding/marshal.go
@@ -42,6 +42,16 @@ func (e *Encoder) Encode(v interface{}) error {
 	return e.encode(reflect.ValueOf(v))
 }
 
+// EncodeAll encodes a variable number of arguments.
+func (e *Encoder) EncodeAll(vs ...interface{}) error {
+	for i := range vs {
+		if err := e.Encode(vs[i]); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
 // write catches instances where short writes do not return an error.
 func (e *Encoder) write(p []byte) error {
 	n, err := e.w.Write(p)
@@ -193,6 +203,16 @@ func (d *Decoder) Decode(v interface{}) (err error) {
 
 	d.decode(pval.Elem())
 	return
+}
+
+// DecodeAll decodes a variable number of arguments.
+func (d *Decoder) DecodeAll(vs ...interface{}) error {
+	for i := range vs {
+		if err := d.Decode(vs[i]); err != nil {
+			return err
+		}
+	}
+	return nil
 }
 
 // readN reads n bytes and panics if the read fails.
--- a/encoding/marshal.go
+++ b/encoding/marshal.go
@@ -12,21 +12,6 @@ import (
 	"reflect"
 )
 
-// A SiaMarshaler can encode itself as a byte slice.
-type SiaMarshaler interface {
-	MarshalSia() []byte
-}
-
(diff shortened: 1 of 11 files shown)


encoding/marshal.go
```diff
@@ -12,14 +12,23 @@ import (
 	"reflect"
 )
 
-// A SiaMarshaler can encode itself as a byte slice.
+const (
+	maxDecodeLen = 10 * 1024 * 1024 // 10 MB
+	maxSliceLen  = 4 * 1024 * 1024  // 4 MB
```
DavidVorick: MiB

encoding/marshal_test.go
```diff
@@ -237,3 +250,34 @@ func TestReadWriteFile(t *testing.T) {
 		t.Error("expected error, got nil")
 	}
 }
+
+func BenchmarkEncode(b *testing.B) {
```
DavidVorick: Can you add a comment that includes historical performance info?

encoding/marshal_test.go
```diff
@@ -237,3 +250,34 @@ func TestReadWriteFile(t *testing.T) {
...
+			err := enc.Encode(testStructs[i])
+			if err != nil {
+				b.Fatal(err)
+			}
+		}
+	}
+	b.SetBytes(int64(buf.Len()))
+}
+
+func BenchmarkDecode(b *testing.B) {
```
DavidVorick: ditto

types/block_bench_test.go
```diff
@@ -6,6 +6,14 @@ import (
 	"github.com/NebulousLabs/Sia/encoding"
 )
 
+func BenchmarkEncodeBlock(b *testing.B) {
```
DavidVorick: can you include a comment with historical data? Also, can it be named `BenchmarkEncodeEmptyBlock`?  Same with the next function, the docstring has the right name but the actual function name is wrong.

DavidVorick:
looks like your rebase got tangled, you've got some repeat commits from master.

lukechampine:
A M A T E U R
H O U R

modules/consensus/accept_bench_test.go
```diff
@@ -12,7 +12,7 @@ import (
 // BenchmarkAcceptEmptyBlocks measures how quckly empty blocks are integrated
 // into the consensus set.
 //
-// David's Setup - 08-26-2015: 6.377 ms / op
+// i7-4770, 61ab2f5: 6.377 ms / op
```
DavidVorick: Processor doesn't mean much because it's an i/o bottleneck.

PR #797, opened 2015-09-17
More permissive Gateway

lukechampine:
The new `acceptConn` rules are detailed in the commit message of 3c6b222. Another important change is that only outbound connections will count towards the `wellConnectedThreshold`.

This latter change may be a problem, now that I think of it. The reason is that port-forwarded nodes are a scarce resource, and we don't want to overwhelm them with connections. Perhaps 3-5 outgoing connections would be healthier.

### DIFF
--- a/modules/gateway/gateway.go
+++ b/modules/gateway/gateway.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"os"
 
+	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/sync"
 )
@@ -83,10 +84,18 @@ func New(addr string, persistDir string) (g *Gateway, err error) {
 		log:        logger,
 	}
 
-	// Load the old peer list. If it doesn't exist, no problem, but if it does,
+	// Load the old node list. If it doesn't exist, no problem, but if it does,
 	// we want to know about any errors preventing us from loading it.
 	if loadErr := g.load(); loadErr != nil && !os.IsNotExist(loadErr) {
 		return nil, loadErr
+	}
+
+	// Add the bootstrap peers to the node list.
+	if build.Release != "testing" {
+		for _, addr := range modules.BootstrapPeers {
+			g.addNode(addr)
+		}
+		g.save()
 	}
 
 	// Create listener and set address.
--- a/modules/gateway/gateway.go
+++ b/modules/gateway/gateway.go
@@ -33,6 +33,9 @@ type Gateway struct {
 	// network.
 	nodes map[modules.NetAddress]struct{}
 
+	// closeChan is used to shut down the Gateway's goroutines.
+	closeChan chan struct{}
+
 	persistDir string
 
 	log *log.Logger
@@ -54,6 +57,8 @@ func (g *Gateway) Close() error {
 		return err
 	}
 	g.mu.RUnlock(id)
+	// send close signal
+	close(g.closeChan)
 	// clear the port mapping (no effect if UPnP not supported)
 	g.clearPort(g.myAddr.Port())
 	// shut down the listener
@@ -79,6 +84,7 @@ func New(addr string, persistDir string) (g *Gateway, err error) {
 		initRPCs:   make(map[string]modules.RPCFunc),
 		peers:      make(map[modules.NetAddress]*peer),
 		nodes:      make(map[modules.NetAddress]struct{}),
+		closeChan:  make(chan struct{}),
 		persistDir: persistDir,
 		mu:         sync.New(modules.SafeMutexDelay, 2),
 		log:        logger,
--- a/modules/gateway/peers.go
+++ b/modules/gateway/peers.go
@@ -104,20 +104,6 @@ func (g *Gateway) acceptConn(conn net.Conn) {
 	addr := modules.NetAddress(conn.RemoteAddr().String())
 	g.log.Printf("INFO: %v wants to connect", addr)
 
-	// don't connect to an IP address more than once
-	if build.Release != "testing" {
-		id := g.mu.RLock()
-		for p := range g.peers {
-			if p.Host() == addr.Host() {
-				g.mu.RUnlock(id)
-				conn.Close()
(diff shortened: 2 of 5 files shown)


modules/gateway/gateway.go
```diff
@@ -83,12 +84,20 @@ func New(addr string, persistDir string) (g *Gateway, err error) {
...
 
-	// Load the old peer list. If it doesn't exist, no problem, but if it does,
+	// Load the old node list. If it doesn't exist, no problem, but if it does,
 	// we want to know about any errors preventing us from loading it.
 	if loadErr := g.load(); loadErr != nil && !os.IsNotExist(loadErr) {
 		return nil, loadErr
 	}
 
+	// Add the bootstrap peers to the node list.
+	if build.Release != "testing" {
```
DavidVorick: why not during testing?

modules/gateway/peers.go
```diff
@@ -143,19 +129,26 @@ func (g *Gateway) acceptConn(conn net.Conn) {
...
 	if len(g.peers) >= fullyConnectedThreshold {
-		oldPeer, err := g.randomInboundPeer()
-		if err == nil {
-			g.peers[oldPeer].sess.Close()
-			delete(g.peers, oldPeer)
-			g.log.Printf("INFO: disconnected from %v to make room for %v", oldPeer, addr)
+		// first choose a random peer, preferably inbound
+		kick, err := g.randomInboundPeer()
+		if err != nil {
+			kick, _ = g.randomPeer()
```
DavidVorick: don't kick outbound peers to make room. Also, wouldn't this code never be reached anyway? If you are so full that you are kicking peers, you'll have inbound peers.

modules/gateway/peers.go
```diff
@@ -143,19 +129,26 @@ func (g *Gateway) acceptConn(conn net.Conn) {
...
+		}
+		// if another peer shares this IP, choose that one instead
+		for p := range g.peers {
+			if p.Host() == addr.Host() {
+				kick = p
+				break
+			}
 		}
+		g.peers[kick].sess.Close()
```
DavidVorick: an attacker could cause a high volume of churn by repeatedly connecting cycling through a set of 129 available peers. And how does this work if a bunch of people are connecting simultaneously? Could an attacker have you kicking 50+ peers per second?

DavidVorick:
upnp should make peers more available, we have that implemented now. Thus far, nobody has complained that they aren't able to get to 8 peers (unless they are stuck at 0), and in 99% of cases you get 8 outbound peers before you ever get your first inbound peer anyway.

Even 8 seems weak to me, and I think it's only worth keeping it that low to minimize total network overhead while our broadcasting + routing algorithms are naive.

modules/gateway/peers.go
DavidVorick: perhaps this could be avoided by making sure that `Close()` will not terminate a connection until it is at least 90 seconds old?

But then you also have to make sure that there isn't a race condition where you kick the same peer twice and end up with too many overall connections.

modules/gateway/gateway.go
lukechampine: Because then the peerManager will connect to the actual bootstrap nodes during tests. Also, some tests already assume that the node list will be empty.

modules/gateway/peers.go
lukechampine: it is highly unlikely. I think it could only happen if the user manually connected to 128 nodes. Perhaps in that case we should just reject all inbound connections?

modules/gateway/peers.go
lukechampine: I think it would be pretty easy to limit the gateway to 1 new connection every n seconds. That would reduce strain/churn by ensuring that you will stay connected for at least that long.

lukechampine:
https://github.com/bitcoin/bitcoin/pull/6014
long discussion on why more outbound connections is bad

modules/gateway/peers.go
DavidVorick: oh forgot about manually connecting.

Just need to make sure that the user's outbound peers are never tossed due to inbound peers replacing them. Blocking all inbound peers if you have 128 outbound peers is probably okay.

DavidVorick:
8 seems weak from a what-if-they-are-all-evil perspective, not a how-fast-and-efficiently-will-this-block-propagate perspective.

Many people would suggest that networking is the weakest link in Bitcoin. It's not an easy thing to fix though.

Right now I think the average long running peer does end up with like 40 connections (assuming ports forwarded).

DavidVorick:
Bitcoin also has stronger sybil defenses such as banning nodes on the same /24

modules/gateway/peers.go
lukechampine: if you have 128 connections and an inbound peer replaces an outbound peer, the next inbound connection will replace that inbound peer. So you'll never have a scenario where repeated inbound connections can kick out all your outbound peers.

DavidVorick: can you add that in a comment then? It's good enough but should be documented.

modules/gateway/peers.go
DavidVorick: That sounds good enough to me. Peers are rejected if enough time hasn't passed since the last one?

Since honest peers will continue to try connecting after timeouts, it should be okay to do that.

modules/gateway/peers.go
```diff
@@ -14,7 +14,10 @@ import (
 )
 
 const (
+	// the gateway will abort a connection attempt after this long
 	dialTimeout = 2 * time.Minute
+	// the gateway will sleep this long between incoming connections
+	acceptInterval = 3 * time.Second
```
DavidVorick: Is 3 seconds too short?

DavidVorick:
Looks good to me except I'm not sure that 3 seconds is long enough for the acceptInterval.

lukechampine:
The common case (where kicking becomes relevant) is going to be 8 outbound connections and 120 inbound connections. At maximum churn (1 new inbound connection every 3 seconds), the average connection duration for an inbound peer is 3 seconds * 83 connections = 4 minutes (assuming my math is correct: `(119/120)^83 < 0.5`)

Longer acceptIntervals are annoying because they significantly slow down testing.

PR #798, opened 2015-09-21
Fix the Consensus Database

DavidVorick:
Accepting a block now happens as a read tx followed by 2 write tx's. The first tx adds the block regardless of verification (to the block tree- happens before verification), and the second tx adds the block if it verifies and extends the longest chain (updating the whole consensus set in the process).

The database is now much faster.

Some obscure bugs were fixed. Pretty much every file got touched. I may have broken syncrhonization, our testing is not really sufficient for me to be certain.

I uncovered two poorly documented behaviors of boltdb:

1. Bucket.Stats() does not update within a tx, only after a tx is committed.
2. if you do a Bucket.ForEach and delete items from the bucket, you get nondeterministic behavior. While this isn't surprising, it's also not documented and was actually the source of the bug that took me almost 1 day to find.

The database is a lot cleaner. I added some tests, commented out others. Except for the synchronization stuff, I do feel good about the test coverage. I'll probably come back to this in another week or two to continue testing and optimization. There might be a few more latent bugs in the system, but overall I think we're unlikely to run into them unless someone is spending a lot of time trying to trigger them. The most important tests are all passing.


modules/consensus/accept_test.go
```diff
@@ -17,26 +16,29 @@ import (
 
 // TestIntegrationDoSBlockHandling checks that saved bad blocks are correctly ignored.
 func TestIntegrationDoSBlockHandling(t *testing.T) {
+	// TestIntegrationDoSBlockHandling catches a wide array of simple errors,
+	// and therefore is included in the short tests despite being someone
```
bitspill: someone -> somewhat ?

lukechampine:
technically the `Delete` behavior is documented, just not where you'd expect. `ForEach` uses a `Cursor` to iterate over the elements of a bucket, and the docstring for `Cursor` states:
```go
// Changing data while traversing with a cursor may cause it to be invalidated
// and return unexpected keys and/or values. You must reposition your cursor
// after mutating data.
```

Makefile
```diff
@@ -62,16 +62,17 @@ clean:
...
+       ./modules/explorer ./modules/gateway ./modules/host ./modules/hostdb \
+       ./modules/miner ./modules/renter ./modules/transactionpool \
+       ./modules/wallet ./persist ./siac ./siae ./types
+test: REBUILD
+	go test -short -tags='testing' -timeout=3s $(pkgs) -run $(run)
+test-v: REBUILD
+	go test -race -v -short -tags='debug testing' -timeout=15s $(pkgs) -run $(run)
 test-long: clean fmt REBUILD
-	go test -v -race -tags='testing debug' -timeout=300s $(pkgs)
+	go test -v -race -tags='testing debug' -timeout=300s $(pkgs) -run $(run)
```
lukechampine: probably best to use the `-run=` form for consistency with the other arguments.

now we just need to add a `tags` variable, and we'll have replicated `go test`'s functionality in full!

encoding/marshal_test.go
```diff
@@ -126,6 +126,9 @@ func TestEncode(t *testing.T) {
 }
 
 func TestDecode(t *testing.T) {
+	if testing.Short() {
+		t.SkipNow()
+	}
```
lukechampine: really? This is a slow test now?

modules/consensus/database_helpers_test.go
```diff
@@ -0,0 +1,476 @@
...
+	err := db.View(func(tx *bolt.Tx) error {
+		b := tx.Bucket(bucket)
+		if build.DEBUG && b == nil {
+			panic(errNilBucket)
+		}
+		return b.ForEach(fn)
+	})
+	if err != nil {
+		panic(err)
+	}
```
lukechampine: unqualified panic

there are a whole lot of functions here. Should grep around and make sure they are all being used. Some could probably be combined.

modules/consensus/accept_reorg_test.go
```diff
@@ -0,0 +1,377 @@
...
+	return &reorgSets{
+		cstMain:   cstMain,
+		cstAlt:    cstAlt,
+		cstBackup: cstBackup,
+	}
+}
+
+// save takes all of the blocks in cstMain and moves them to cstBackup.
+func (rs *reorgSets) save() {
+	for i := types.BlockHeight(1); i <= rs.cstMain.cs.dbBlockHeight(); i++ {
```
lukechampine: `dbBlockHeight` executes a read txn every time it's called. Probably best to call it outside the loop, unless the blockheight is always going to be small.

modules/consensus/accept_reorg_test.go
```diff
@@ -0,0 +1,377 @@
...
+	}
+
+	// Try to trigger consensus inconsistencies by doing a full reorg on the
+	// simple block.
+	rs.fullReorg()
+}
+
+/// BREAK ///
+/// BREAK ///
+/// BREAK ///
```
lukechampine: ?

modules/consensus/accept_test.go
```diff
@@ -1811,12 +1305,5 @@ func TestTaxHardfork(t *testing.T) {
...
-
-	// Run a siacoins check to make sure that order has been restored - note
-	// that the siacoin check will fail in the middle, and thus is commented
-	// out until after the hardfork.
-	err = cst.cs.checkSiacoins()
-	if err != nil {
-		t.Fatal(err)
-	}
 }
+*/
```
lukechampine: Going to be restored later? Seems like an important test.

modules/consensus/accept_txntypes_test.go
```diff
@@ -0,0 +1,688 @@
...
+// being closed at the same time.
+
+// TestRegressionFileContractMaturity creates a consensus set tester and uses
+// it to call testFileContractMaturity. In the past, bolt's ForEach function
+// had been used incorrectly, resulting in the incorrect processing of bulk
+// file contracts.
+
+/// BREAK ///
+/// BREAK ///
+/// BREAK ///
```
lukechampine: ?? what's with all the floating docstrings?

modules/consensus/accept_txntypes_test.go
```diff
@@ -0,0 +1,688 @@
...
+	_, err = cst.miner.AddBlock()
+	if err != nil {
+		return err
+	}
+
+	// Check that the siafund pool was increased.
+	var siafundPool types.Currency
+	err = cst.cs.db.Update(func(tx *bolt.Tx) error {
+		siafundPool = getSiafundPool(tx)
+		return nil
```
lukechampine: can use `dbGetSiafundPool`

modules/consensus/applytransaction.go
```diff
@@ -183,22 +143,18 @@ func (cs *ConsensusSet) applyStorageProofs(tx *bolt.Tx, pb *processedBlock, t ty
...
 // applyTxSiafundInputs takes all of the siafund inputs in a transaction and
 // applies them to the state, updating the diffs in the processed block.
-func (cs *ConsensusSet) applySiafundInputs(tx *bolt.Tx, pb *processedBlock, t types.Transaction) error {
+func applySiafundInputs(tx *bolt.Tx, pb *processedBlock, t types.Transaction) {
 	for _, sfi := range t.SiafundInputs {
 		// Calculate the volume of siacoins to put in the claim output.
 		sfo, err := getSiafundOutput(tx, sfi.ParentID)
 		if err != nil {
-			return err
+			panic(err)
```
lukechampine: unqualified panic

modules/consensus/applytransaction.go
```diff
@@ -247,47 +196,19 @@ func (cs *ConsensusSet) applySiafundOutputs(tx *bolt.Tx, pb *processedBlock, t t
...
-	}
-	return nil
+func applyTransaction(tx *bolt.Tx, pb *processedBlock, t types.Transaction) {
+	applySiacoinInputs(tx, pb, t)
+	applySiacoinOutputs(tx, pb, t)
+	applyFileContracts(tx, pb, t)
+	applyFileContractRevisions(tx, pb, t)
+	applyStorageProofs(tx, pb, t)
+	applySiafundInputs(tx, pb, t)
+	applySiafundOutputs(tx, pb, t)
```
lukechampine: :+1:

modules/consensus/consensusset.go
```diff
@@ -51,16 +42,25 @@ type ConsensusSet struct {
...
 
-	// TODO: add a persistDir
-	mu *sync.RWMutex
+	// checkingConsistency is a bool indicating whether or not a consistency
+	// check is in progress. The consistency check logic call itself, resulting
+	// in infinite loops. This bool prevents that while still allowing for full
+	// granularity consistency checks. Previously, consistency checks were only
+	// performed after a full reorg, but now they are performed after every
+	// block.
+	checkingConsistency bool
```
lukechampine: hmm. Seems like you could get rid of this if not for `checkRevertApply`. Could you possibly remove `checkRevertApply` from `checkConsistency`, and just perform the consensus set hash check in `revertToBlock`?

modules/consensus/consensusset_test.go
```diff
@@ -29,6 +31,27 @@ type consensusSetTester struct {
 	persistDir string
 }
 
+// randAddress returns a random address that is not spendable.
```
lukechampine: *probably* not spendable :P

types/filecontracts.go
```diff
@@ -108,15 +108,23 @@ func (fcid FileContractID) StorageProofOutputID(proofStatus ProofStatus, i uint6
...
+// PostTax returns the amount of currency remaining in a file contract payout
+// after tax.
+func PostTax(height BlockHeight, payout Currency) Currency {
+	// COMPATv0.4.0 - until the first 20,000 blocks have been archived, they
+	// will need to be handled in a special way.
+	if (height < 21e3 && build.Release == "standard") || (height < 10 && build.Release == "testing") {
+		return payout.Sub(payout.MulFloat(0.039).RoundDown(SiafundCount))
+	}
+	return payout.Sub(payout.MulTax().RoundDown(SiafundCount))
+}
```
lukechampine: isn't this just `payout.Sub(Tax(height, payout))`? Is a separate function warranted? I find the name a bit confusing too.

Also, maybe `Tax` could be a method of `Currency`?

modules/consensus/consistency.go
```diff
@@ -2,281 +2,331 @@ package consensus
...
-	}
+// checkSiacoinCount checks that the number of siacoins countable within the
+// consensus set equal the expected number of siacoins for the block height.
+func checkSiacoinCount(tx *bolt.Tx) {
+	// Iterate through all the buckets looking for the delayed siacoin output
+	// buckets, and check that they are for the correct heights.
+	var dscoSiacoins types.Currency
+	err := tx.ForEach(func(name []byte, b *bolt.Bucket) error {
+		// Check if the bucket is a delayed siacoin output bucket.
+		if !strings.HasPrefix(string(name), string(prefixDSCO)) {
```
lukechampine: there is a `bytes.HasPrefix`

modules/consensus/consistency.go
```diff
@@ -2,281 +2,331 @@ package consensus
...
+		fmt.Println("Claim:", claimSiacoins)
+		if totalSiacoins.Cmp(expectedSiacoins) < 0 {
+			fmt.Println("total:", totalSiacoins)
+			fmt.Println("expected:", expectedSiacoins)
+			fmt.Println("expected is bigger:", expectedSiacoins.Sub(totalSiacoins))
+		} else {
+			fmt.Println("total:", totalSiacoins)
+			fmt.Println("expected:", expectedSiacoins)
+			fmt.Println("total is bigger:", totalSiacoins.Sub(expectedSiacoins))
+		}
```
lukechampine: instead of printing directly, stick these in a string and call `panic` on the string

modules/consensus/consistency.go
```diff
@@ -2,281 +2,331 @@ package consensus
...
+	// another map to track which ids have appeared in the dsco set.
+	dscoTracker := make(map[types.BlockHeight]struct{})
+	idMap := make(map[types.SiacoinOutputID]struct{})
+
+	// Iterate through all the buckets looking for the delayed siacoin output
+	// buckets, and check that they are for the correct heights.
+	err := tx.ForEach(func(name []byte, b *bolt.Bucket) error {
+		// If the bucket is not a delayed siacoin output bucket or a file
+		// contract expiration bucket, skip.
+		if !strings.HasPrefix(string(name), string(prefixDSCO)) {
```
lukechampine: ditto

modules/consensus/consistency.go
```diff
@@ -2,281 +2,331 @@ package consensus
...
+			continue
+		}
+		_, exists := dscoTracker[i]
+		if !exists {
+			panic("missing a dsco bucket")
+		}
+		expectedBuckets++
+	}
+	if len(dscoTracker) != expectedBuckets {
+		panic("too many dsco buckets")
```
lukechampine: lots of new unqualified panics. Maybe I'm missing something?

modules/consensus/database.go
```diff
@@ -56,417 +86,254 @@ func openDB(filename string) (*setDB, error) {
...
-		panic(errNilBucket)
-	}
-	k := encoding.Marshal(key)
-	v := encoding.Marshal(value)
-	if build.DEBUG && b.Get(k) != nil {
-		panic(errRepeatInsert)
+	// Set the block height to -1, so the genesis block is at height 0.
+	blockHeight := tx.Bucket(BlockHeight)
+	underflow := types.BlockHeight(0)
+	err := blockHeight.Put(BlockHeight, encoding.Marshal(underflow-1))
```
lukechampine: seems hackish. Perhaps better to just add the height + id manually, instead of using `pushPath`.

modules/consensus/database.go
```diff
@@ -56,417 +86,254 @@ func openDB(filename string) (*setDB, error) {
...
 }
 
-// inBucket checks if an item with the given key is in the bucket
-func (db *setDB) inBucket(bucket []byte, key interface{}) bool {
-	exists, err := db.Exists(bucket, encoding.Marshal(key))
+// blockHeight returns the height of the blockchain.
+func blockHeight(tx *bolt.Tx) types.BlockHeight {
+	var height int
+	bh := tx.Bucket(BlockHeight)
+	err := encoding.Unmarshal(bh.Get(BlockHeight), &height)
```
lukechampine: isn't `BlockHeight` a `uint64`??

modules/consensus/database.go
```diff
@@ -56,417 +86,254 @@ func openDB(filename string) (*setDB, error) {
...
-	if err != nil {
+	// Decode the block - should never fail.
+	var pb processedBlock
+	err := encoding.Unmarshal(pbBytes, &pb)
+	if build.DEBUG && err != nil {
 		panic(err)
 	}
-	err = encoding.Unmarshal(idBytes, &id)
-	if err != nil {
+	return &pb, nil
```
lukechampine: why return a pointer?

modules/consensus/database.go
```diff
@@ -56,417 +86,254 @@ func openDB(filename string) (*setDB, error) {
...
+	heightBytes := bh.Get(BlockHeight)
+	var oldHeight types.BlockHeight
+	err := encoding.Unmarshal(heightBytes, &oldHeight)
 	if build.DEBUG && err != nil {
 		panic(err)
 	}
-	var pb processedBlock
-	err = encoding.Unmarshal(pbBytes, &pb)
+	newHeightBytes := encoding.Marshal(oldHeight + 1)
+	err = bh.Put(BlockHeight, newHeightBytes)
```
lukechampine: for our purposes, `Put` should never return an error:
>Put returns an error if the bucket was created from a read-only transaction, if the key is blank, if the key is too large, or if the value is too large.

modules/consensus/database.go
```diff
@@ -56,417 +86,254 @@ func openDB(filename string) (*setDB, error) {
...
-		bucket := tx.Bucket(BlockMap)
-		id := pb.Block.ID()
-		return bucket.Put(id[:], encoding.Marshal(*pb))
-	})
-	if err != nil {
+	// Remove the block from the path - make sure to remove the block at
+	// oldHeight.
+	bp := tx.Bucket(BlockPath)
+	err = bp.Delete(oldHeightBytes)
+	if build.DEBUG && err != nil {
```
lukechampine: similarly, `Delete` should never return an error:
>Delete returns an error if the bucket was created from a read-only transaction

modules/consensus/diffs_test.go
```diff
@@ -301,6 +308,7 @@ func TestCommitDelayedSiacoinOutputDiff(t *testing.T) {
 	dscod.Direction = modules.DiffApply                             // set the direction to apply
 	cst.cs.commitDelayedSiacoinOutputDiff(dscod, modules.DiffApply) // apply an already existing delayed output.
 }
+*/
```
lukechampine: lots of dead tests here. What's the plan?

modules/consensus/fork.go
```diff
@@ -6,117 +6,107 @@ import (
...
-			panic("state hash changed after an unsuccessful fork attempt")
-		}
+// forkBlockchain will move the consensus set onto the 'newBlock' fork. An
+// error will be returned if any of the blocks applied in the transition are
+// found to be invalid. forkBlockchain is atomic; the ConsensusSet is only
+// updated if the function returns nil.
+func (cs *ConsensusSet) forkBlockchain(tx *bolt.Tx, newBlock *processedBlock) (revertedBlocks, appliedBlocks []*processedBlock, err error) {
+	commonParent := backtrackToCurrentPath(tx, newBlock)[0]
+	revertedBlocks = cs.revertToBlock(tx, commonParent)
+	appliedBlocks, err = cs.applyUntilBlock(tx, newBlock)
```
lukechampine: :+1:

modules/consensus/database_helpers_test.go
```diff
@@ -0,0 +1,476 @@
...
+		}
+		return nil
+	})
+	if dbErr != nil {
+		panic(dbErr)
+	}
+	return dsco, err
+}
+
+/// BREAK ///
```
lukechampine: not a fan of these BREAKs...

modules/consensus/maintenance.go
```diff
@@ -31,26 +30,37 @@ func (cs *ConsensusSet) applyMinerPayouts(tx *bolt.Tx, pb *processedBlock) error
...
 // applyMaturedSiacoinOutputs goes through the list of siacoin outputs that
 // have matured and adds them to the consensus set. This also updates the block
 // node diff set.
-func (cs *ConsensusSet) applyMaturedSiacoinOutputs(tx *bolt.Tx, pb *processedBlock) error {
+func applyMaturedSiacoinOutputs(tx *bolt.Tx, pb *processedBlock) {
 	// Skip this step if the blockchain is not old enough to have maturing
 	// outputs.
-	if !(pb.Height > types.MaturityDelay) {
-		return nil
+	if !(pb.Height >= types.MaturityDelay) {
```
lukechampine: `pb.Height < types.MaturityDelay`?

modules/consensus/persist.go
```diff
@@ -1,136 +1,92 @@
...
-	// Prevent the miner payout for the genesis block from being spent
-	cs.db.addSiacoinOutputs(cs.blockRoot.Block.MinerPayoutID(0), types.SiacoinOutput{
-		Value:      types.CalculateCoinbase(0),
-		UnlockHash: types.UnlockHash{},
+	// Send all of the existing blocks to subscribers - temporary while
+	// subscribers don't have any persistence for block progress.
+	return cs.db.View(func(tx *bolt.Tx) error {
+		for i := types.BlockHeight(0); i <= height; i++ {
+			// Fetch the processed block at height 'i'.
+			id, err := getPath(tx, i)
```
lukechampine: could use a cursor to do the iteration here

modules/consensus/persist.go
```diff
@@ -1,136 +1,92 @@
...
+			if build.DEBUG && err != nil {
+				panic(err)
+			}
+			pb, err := getBlockMap(tx, id)
+			if build.DEBUG && err != nil {
+				panic(err)
+			}
+
+			// Send the block to subscribers.
+			cs.updateSubscribers(nil, []*processedBlock{pb})
```
lukechampine: I suppose you could batch these, e.g. 10 at a time, but I'm not sure that it would be any faster.

modules/consensus/synchronize.go
```diff
@@ -11,54 +11,76 @@ import (
...
+// found.
+func blockHistory(tx *bolt.Tx) (blockIDs [32]types.BlockID) {
+	height := blockHeight(tx)
 	step := types.BlockHeight(1)
-	for height := s.height(); ; height -= step {
-		// after 12, start doubling
-		knownBlocks = append(knownBlocks, s.db.getPath(height))
-		if len(knownBlocks) >= 12 {
+	// The final step is to include the genesis block, which is why `step <
+	// height` is used as opposed to `step <= height`.
```
lukechampine: `step < height` does not appear in this function

modules/consensus/maintenance.go
bitspill: I'm not sure if it applies in go, but in some cases if either side is NaN or null the >= would be false

modules/consensus/synchronize.go
```diff
@@ -11,54 +11,76 @@ import (
...
+		}
+		blockIDs[i] = blockID
+
+		// Determine the height of the next block to include and then increase
+		// the step size. The height must be decreased first to prevent
+		// underflow.
+		//
+		// `i >= 9` means that the first 10 blocks will be included, and then
+		// skipping will start.
+		if i >= 9 {
```
lukechampine: This only includes the first 9 blocks.

modules/consensus/synchronize.go
```diff
@@ -113,27 +126,39 @@ func (s *ConsensusSet) sendBlocks(conn modules.PeerConn) error {
...
+			if err != nil {
+				continue
+			}
+			pathID, err := getPath(tx, pb.Height)
+			if err != nil {
+				continue
+			}
+			if pathID != pb.Block.ID() {
+				continue
+			}
```
lukechampine: would probably help to have a function that fetched a block from current path by ID

modules/consensus/synchronize.go
```diff
@@ -113,27 +126,39 @@ func (s *ConsensusSet) sendBlocks(conn modules.PeerConn) error {
...
+			if pb.Height == blockHeight(tx) {
```
lukechampine: haven't you already calculated the blockHeight as `csHeight`?

also, would probably help to have a function that fetched a block from current path by ID

types/currency_test.go
```diff
@@ -222,6 +222,10 @@ func TestCurrencyEncoding(t *testing.T) {
 // TestNegativeCurrencyMulRat checks that negative numbers are rejected when
 // calling MulRat on the currency type.
 func TestNegativeCurrencyMulRat(t *testing.T) {
+	if testing.Short() {
+		t.SkipNow()
+	}
+
```
lukechampine: how are these possibly taking too long?

lukechampine:
Okay, done. Additional comments:

- I think it would be beneficial to cleanly separate the consensus logic from the database operations. A `consensus/database` subpackage could be used to achieve this, but then `database` would not be able to use any types from `consensus` due to the resulting import cycle. Still, it's worth investigating. The current implementation is tightly coupled to bolt.
- On a related note, the `setDB` type is a little weird. It's wrapping a `persist.BoltDatabase`, but should it? Do we still need `persist.BoltDatabase`, which has its own metadata and unused methods? Probably not.
- There's a lot of duplication going on with respect to the `db` functions, which just start a new txn and then call some other function internally. Would be nice if we could clean that up somehow.

modules/consensus/maintenance.go
lukechampine: Go tends to behave a lot more sanely than, e.g. JavaScript. These types are both standard integers, so no problems there.

Makefile
DavidVorick: yeah but 80% of the time you don't use the extra stuff, vs with `go test` you'd end up writing everything out every time.

modules/consensus/database_helpers_test.go
DavidVorick: any function of the setDB is deprecated, I'm still working on tossing them all.

modules/consensus/accept_reorg_test.go
DavidVorick: I could probably document that better. It's a break between the deprecated functions and the not-deprecated functions

modules/consensus/accept_test.go
DavidVorick: It will eventually be restored in some capacity, probably in the form of a new test that does the same thing but without using deprecated paradigms

modules/consensus/accept_reorg_test.go
DavidVorick: not-deprecated functions are on top

modules/consensus/accept_txntypes_test.go
DavidVorick: They are for functions I haven't written yet but would like to. They are basically TODOs, but hopefully things I will address before this PR is merged.

modules/consensus/accept_txntypes_test.go
DavidVorick: this is below the 'deprecated function' break.

modules/consensus/consensusset.go
DavidVorick: maybe, otoh I'd really rather perform the full suite of consistency checks after every atomic action (adding or removing a block)

types/filecontracts.go
DavidVorick: I thought about making tax a method of the currency, but really it's about file contracts and not about the currency type. If the currency type were a separate package, the Tax method would not follow it. So I didn't think it should be of the currency type.

PostTax is used as a function almost as much as Tax is, perhaps even more than Tax is. I'm open to a different name but think it's valuable as a function.

modules/consensus/consistency.go
DavidVorick: sloppy coding, would rather have all the panics be qualified

DavidVorick: not as worried about it in any TestXXX function, or testXXX function either. The consistency stuff should probably be qualified though.

modules/consensus/database.go
DavidVorick: I think because it starts at -1?

Actually, I was pretty sure I made it a `types.BlockHeight`, but I guess that was in a commit that I had to roll back. (I ended up throwing away 100+ lines of changes more than a dozen times because the unit tests starting failing - lots of stuff got implemented a few times)

modules/consensus/database.go
DavidVorick: processed block is always handled as a pointer. It's a pretty large object, and you want to make changes to it that persist across function calls.

modules/consensus/database.go
DavidVorick: still need to check the error, helps catch developer mistakes

modules/consensus/database.go
DavidVorick: and one you would want to catch with a panic :)

modules/consensus/diffs_test.go
DavidVorick: bring them back after refactoring them.

modules/consensus/database_helpers_test.go
DavidVorick: divides between deprecated and non-deprecated stuff, breaks should be gone by the end of the week.

modules/consensus/persist.go
DavidVorick: that might work

DavidVorick: as long as they are sorted correctly

modules/consensus/synchronize.go
DavidVorick: Old code did 13 blocks, followed by 19 exponential blocks, one of which was the genesis block. 2^18 is only 250k, which is not enough. Need to get into the millions.

9 is 10 blocks, genesis makes 11, 2^21 = 2 million, which is more acceptable.

types/currency_test.go
DavidVorick: They depend on the debug flag being set. testing.Short doesn't set the debug flag

modules/consensus/database.go
lukechampine: so mostly a convenience thing I guess. nbd, my ears just perk up when I see a function returning a pointer to a locally-allocated object.

DavidVorick:
Most of your comments are already addressed, except in the test files. There's no duplication in database.go, only in database_helpers_test.go, which is a test file. setDB is deprecated, but a lot of tests still use functions that are implemented on the setDB, can't get rid of it all at once so it's still there.

Tight coupling to boltdb seems pretty inevitable to me. Subpackage seems overkill. If you want to squeeze out performance, which we do, you need to be working as close to the metal as possible. Abstracting the db away harms that.

We could turn all of the functions inside of database.go into an interface, but I'd rather not do that work until it's necessary.

modules/consensus/synchronize.go
lukechampine: pretty sure the old code only had 12 linear blocks...and it's not fair to discount the genesis block, which is sort of like the 20th doubling, but capped to the actual height.

DavidVorick: I think there was an off-by-one that I found while testing.

lukechampine:
ok. I agree that a subpackage is overkill. I'm skeptical that abstraction would impact performance though. I'm imagining something along the lines of `persist.Database`, where you embed the underlying bolt type and then implement some helpful methods on top. It seems like `persist.Database` isn't really being used right now, so we should consider removing it.

DavidVorick:
Depends on how badly you want me to get rid of the deprecated stuff. I don't want to remove any of the empty docstrings, because they are serving as TODOs, but by and large I think that this PR is production-ready. With the exception of the comments you've made (which I haven't addressed yet), the non-test code should be clean + good, it's just the test code that needs work.

types/currency_test.go
lukechampine: yikes, that's confusing

modules/consensus/database.go
DavidVorick: I don't think any of it is particularly pretty, ideally I'd just be able to call bucket.KeyN =/

modules/consensus/synchronize.go
DavidVorick: I realized that I was confused about how this function works. You probably don't need to double all the way out to the genesis block, and are unlikely to ever farther than the 3rd element unless you've been victim to an isolation attack.

types/currency_test.go
lukechampine: I tried running `go test -short` with and without `debug`, and the time difference seemed pretty negligible, even on my laptop. No package took more than 100ms longer to complete with `debug`. What's the slowdown like on your laptop?

DavidVorick: 7.4 seconds vs. 7.3 seconds....

heh. Probably should have tested that.

DavidVorick:
I started stress-testing boltdb and it is not really holding up =/. 

Sometimes, and nondeterministically, I'll end up with an empty bucket, despite having a bunch of debug statements which strongly suggest that the bucket was never deleted nor was the final element ever removed. 

The boltdb repo iirc has multiple complaints of this form =/

DavidVorick:
Okay, this is very nearly ready to go. Need to write 2 more tests, then I'm greenlighting it.

1. Spend siafunds
2. block that maximizes siacoin output turmoil.

I'm not sure the best way to construct #2, I might settle for fuzzing instead. It would be cool if I could use the fuzzer on boltdb to chase out the error we're running into.

DavidVorick:
OK.

It's ready.

modules/consensus/database.go
DavidVorick: ah just kidding, it was a mistake. For some reason, didn't cause any test cases to fail. But the rest of the db uses a `types.BlockHeight` when working with that bucket.

modules/consensus/persist.go
DavidVorick: I'm concerned that `updateSubscribers` would call other functions like `ChildTarget` that also result in a `View` tx being opened, then you've got a tx being opened in a tx, and that might cause instability. Since they are both `View` txs, I don't think that there would be a problem but I'm not sure.

The stripped down version of the project demonstrates that this isn't the source of the errors I'm seeing.

modules/consensus/database.go
lukechampine: probably didn't fail because `Unmarshal` treated it as an `int64` and the values were never large enough to overflow into the sign bit.

PR #800, opened 2015-09-27
Refactors crypto/signatures.go to use explicit interfaces and adds unit tests

mtlynch:
Refactors crypto.GenerateSignatureKeys and crypto.DeterministicSignatureKeys
so that they now depend on explicit interfaces.

Adds unit tests to exercise the new Generate and GenerateDeterministic
functions.

### DIFF
--- a/crypto/signatures.go
+++ b/crypto/signatures.go
@@ -21,7 +21,6 @@ type (
 )
 
 var (
-	ErrNilInput         = errors.New("cannot use nil input")
 	ErrInvalidSignature = errors.New("invalid signature")
 	ErrRandUnexpected   = errors.New("unexpected result from random number generator")
 )
--- a/crypto/signatures.go
+++ b/crypto/signatures.go
@@ -23,25 +23,58 @@ type (
 var (
 	ErrNilInput         = errors.New("cannot use nil input")
 	ErrInvalidSignature = errors.New("invalid signature")
+	ErrRandUnexpected   = errors.New("unexpected result from random number generator")
 )
 
-// GenerateKeyPair creates a public-secret keypair that can be used to sign and
-// verify messages.
-func GenerateSignatureKeys() (sk SecretKey, pk PublicKey, err error) {
+// KeyPairGenerator is an interface that allows the caller to generate
+// public-secret key pairs.
+type KeyPairGenerator interface {
+	Generate() (SecretKey, PublicKey, error)
+	GenerateDetermistically(entropy [EntropySize]byte) (SecretKey, PublicKey)
+}
+
+// readBytesFunc is a function pointer that reads bytes into a buffer and
+// returns the total number of bytes written to the buffer.
+type readBytesFunc func([]byte) (int, error)
+
+// deriveEd25519KeyPairFunc is a function pointer that matches the signature of
+// ed25519.GenerateKey.
+type deriveEd25519KeyPairFunc func([EntropySize]byte) (ed25519.SecretKey, ed25519.PublicKey)
+
+// SignatureKeyGenerator is an implementation of KeyPairGenerator.
+type SignatureKeyGenerator struct {
+	readRandBytes readBytesFunc
+	deriveKeyPair deriveEd25519KeyPairFunc
+}
+
+// NewSignatureKeyGenerator creates a new SignatureKeyGenerator type that uses
+// random data and depends on the ed25519.GenerateKey function for deriving
+// key pairs.
+func NewSignatureKeyGenerator() SignatureKeyGenerator {
+	return SignatureKeyGenerator{rand.Read, ed25519.GenerateKey}
+}
+
+// Generate creates a public-secret keypair that can be used to sign and verify
+// messages.
+func (skg SignatureKeyGenerator) Generate() (sk SecretKey, pk PublicKey, err error) {
 	var entropy [EntropySize]byte
-	_, err = rand.Read(entropy[:])
+	written, err := skg.readRandBytes(entropy[:])
 	if err != nil {
 		return
 	}
+	if written != EntropySize {
+		// readRandBytes did not fill the buffer. This should never happen.
+		return sk, pk, ErrRandUnexpected
+	}
 
-	skPointer, pkPointer := ed25519.GenerateKey(entropy)
+	skPointer, pkPointer := skg.deriveKeyPair(entropy)
 	return *skPointer, *pkPointer, nil
 }
 
-// DeterministicSignatureKeys generates keys deterministically using the input
+// GenerateDeterministic generates keys deterministically using the input
 // entropy. The input entropy must be 32 bytes in length.
-func DeterministicSignatureKeys(entropy [EntropySize]byte) (SecretKey, PublicKey) {
-	skPointer, pkPointer := ed25519.GenerateKey(entropy)
+func (skg SignatureKeyGenerator) GenerateDeterministic(entropy [EntropySize]byte) (SecretKey, PublicKey) {
+	skPointer, pkPointer := skg.deriveKeyPair(entropy)
 	return *skPointer, *pkPointer
 }
 
--- a/crypto/signatures.go
+++ b/crypto/signatures.go
@@ -3,6 +3,7 @@ package crypto
 import (
 	"crypto/rand"
 	"errors"
+	"io"
 
 	"github.com/NebulousLabs/ed25519"
 )
@@ -25,58 +26,39 @@ var (
 	ErrInvalidSignature = errors.New("invalid signature")
 )
 
-// KeyPairGenerator allows the caller to generate public-secret key pairs.
-type KeyPairGenerator interface {
-	Generate() (SecretKey, PublicKey, error)
-	GenerateDeterministic(entropy [EntropySize]byte) (SecretKey, PublicKey)
-}
+type (
+	// keyDeriver allows the caller to generate a public-secret key pair based on
+	// provided entropy.
+	keyDeriver interface {
+		deriveKeyPair([EntropySize]byte) (ed25519.SecretKey, ed25519.PublicKey)
+	}
 
-// entropySource allows the caller to retrieve an array of bytes populated to
-// random values.
-type entropySource interface {
-	getEntropy() ([EntropySize]byte, error)
-}
-
-// keyDeriver allows the caller to generate a public-secret key pair based on
-// provided entropy.
-type keyDeriver interface {
-	deriveKeyPair([EntropySize]byte) (SecretKey, PublicKey)
-}
-
-// stdGenerator is an implementation of KeyPairGenerator, allowing the caller
-// to generate public-secret key pairs.
-type stdGenerator struct {
-	es entropySource
-	kd keyDeriver
-}
+	// stdGenerator is an implementation of KeyPairGenerator, allowing the caller
+	// to generate public-secret key pairs.
+	stdGenerator struct {
+		entropySource io.Reader
+		kd            keyDeriver
+	}
+)
 
 // Generate creates a public-secret keypair that can be used to sign and verify
 // messages.
 func (sg stdGenerator) Generate() (sk SecretKey, pk PublicKey, err error) {
-	entropy, err := sg.es.getEntropy()
+	var entropy [EntropySize]byte
+	_, err = sg.entropySource.Read(entropy[:])
 	if err != nil {
 		return
 	}
-	sk, pk = sg.kd.deriveKeyPair(entropy)
-	return sk, pk, nil
+
+	skPointer, pkPointer := sg.kd.deriveKeyPair(entropy)
+	return *skPointer, *pkPointer, nil
 }
 
 // GenerateDeterministic generates keys deterministically using the input
 // entropy. The input entropy must be 32 bytes in length.
 func (sg stdGenerator) GenerateDeterministic(entropy [EntropySize]byte) (SecretKey, PublicKey) {
-	return sg.kd.deriveKeyPair(entropy)
-}
-
-// randSource is an implementation of entropySource that uses rand.Read to
-// generate random bytes.
-type randSource struct{}
-
-// getEntropy returns an array of bytes with random values.
-func (rs randSource) getEntropy() (entropy [EntropySize]byte, err error) {
-	if _, err := rand.Read(entropy[:]); err != nil {
-		return entropy, err
-	}
-	return entropy, nil
+	skPointer, pkPointer := sg.kd.deriveKeyPair(entropy)
+	return *skPointer, *pkPointer
 }
 
 // ed25519Deriver is an implementation of keyDeriver that uses
@@ -85,13 +67,12 @@ type ed25519Deriver struct{}
 
 // deriveKeyPair derives a public-secret key pair derived from the provided
 // array of bytes.
-func (ed ed25519Deriver) deriveKeyPair(entropy [EntropySize]byte) (sk SecretKey, pk PublicKey) {
-	skPointer, pkPointer := ed25519.GenerateKey(entropy)
-	return *skPointer, *pkPointer
+func (ed ed25519Deriver) deriveKeyPair(entropy [EntropySize]byte) (ed25519.SecretKey, ed25519.PublicKey) {
+	return ed25519.GenerateKey(entropy)
 }
 
-// StdKeyGen is a KeyPairGenerator based on randSource and ed25519Deriver.
-var StdKeyGen KeyPairGenerator = stdGenerator{es: randSource{}, kd: ed25519Deriver{}}
+// StdKeyGen is a stdGenerator based on randSource and ed25519Deriver.
+var StdKeyGen stdGenerator = stdGenerator{entropySource: rand.Reader, kd: ed25519Deriver{}}
 
 // SignHash signs a message using a secret key.
 func SignHash(data Hash, sk SecretKey) (sig Signature, err error) {
--- a/crypto/signatures.go
+++ b/crypto/signatures.go
@@ -23,60 +23,75 @@ type (
 var (
 	ErrNilInput         = errors.New("cannot use nil input")
 	ErrInvalidSignature = errors.New("invalid signature")
-	ErrRandUnexpected   = errors.New("unexpected result from random number generator")
 )
 
-// KeyPairGenerator is an interface that allows the caller to generate
-// public-secret key pairs.
+// KeyPairGenerator allows the caller to generate public-secret key pairs.
 type KeyPairGenerator interface {
 	Generate() (SecretKey, PublicKey, error)
-	GenerateDetermistic(entropy [EntropySize]byte) (SecretKey, PublicKey)
+	GenerateDeterministic(entropy [EntropySize]byte) (SecretKey, PublicKey)
 }
 
-// readBytesFunc is a function pointer that reads bytes into a buffer and
-// returns the total number of bytes written to the buffer.
-type readBytesFunc func([]byte) (int, error)
-
-// deriveEd25519Func is a function pointer that matches the signature of
-// ed25519.GenerateKey.
-type deriveEd25519Func func([EntropySize]byte) (ed25519.SecretKey, ed25519.PublicKey)
-
-// SignatureKeyGenerator is an implementation of KeyPairGenerator.
-type SignatureKeyGenerator struct {
-	readRandBytes readBytesFunc
-	deriveKeyPair deriveEd25519Func
+// entropySource allows the caller to retrieve an array of bytes populated to
+// random values.
+type entropySource interface {
+	getEntropy() ([EntropySize]byte, error)
 }
 
-// NewSignatureKeyGenerator creates a new SignatureKeyGenerator type that uses
-// random data and depends on the ed25519.GenerateKey function for deriving
-// key pairs.
-func NewSignatureKeyGenerator() SignatureKeyGenerator {
-	return SignatureKeyGenerator{rand.Read, ed25519.GenerateKey}
+// keyDeriver allows the caller to generate a public-secret key pair based on
+// provided entropy.
+type keyDeriver interface {
+	deriveKeyPair([EntropySize]byte) (SecretKey, PublicKey)
+}
+
+// stdGenerator is an implementation of KeyPairGenerator, allowing the caller
+// to generate public-secret key pairs.
+type stdGenerator struct {
+	es entropySource
+	kd keyDeriver
 }
 
 // Generate creates a public-secret keypair that can be used to sign and verify
 // messages.
-func (skg SignatureKeyGenerator) Generate() (sk SecretKey, pk PublicKey, err error) {
-	var entropy [EntropySize]byte
-	written, err := skg.readRandBytes(entropy[:])
+func (sg stdGenerator) Generate() (sk SecretKey, pk PublicKey, err error) {
+	entropy, err := sg.es.getEntropy()
 	if err != nil {
 		return
 	}
-	if written != EntropySize {
-		// readRandBytes did not fill the buffer. This should never happen.
-		return sk, pk, ErrRandUnexpected
-	}
-
-	skPointer, pkPointer := skg.deriveKeyPair(entropy)
-	return *skPointer, *pkPointer, nil
+	sk, pk = sg.kd.deriveKeyPair(entropy)
+	return sk, pk, nil
 }
 
 // GenerateDeterministic generates keys deterministically using the input
 // entropy. The input entropy must be 32 bytes in length.
-func (skg SignatureKeyGenerator) GenerateDeterministic(entropy [EntropySize]byte) (SecretKey, PublicKey) {
-	skPointer, pkPointer := skg.deriveKeyPair(entropy)
+func (sg stdGenerator) GenerateDeterministic(entropy [EntropySize]byte) (SecretKey, PublicKey) {
+	return sg.kd.deriveKeyPair(entropy)
+}
+
+// randSource is an implementation of entropySource that uses rand.Read to
+// generate random bytes.
+type randSource struct{}
+
+// getEntropy returns an array of bytes with random values.
+func (rs randSource) getEntropy() (entropy [EntropySize]byte, err error) {
+	if _, err := rand.Read(entropy[:]); err != nil {
+		return entropy, err
+	}
+	return entropy, nil
+}
+
+// ed25519Deriver is an implementation of keyDeriver that uses
+// ed25519.GenerateKey to derive keys.
+type ed25519Deriver struct{}
+
+// deriveKeyPair derives a public-secret key pair derived from the provided
+// array of bytes.
+func (ed ed25519Deriver) deriveKeyPair(entropy [EntropySize]byte) (sk SecretKey, pk PublicKey) {
+	skPointer, pkPointer := ed25519.GenerateKey(entropy)
 	return *skPointer, *pkPointer
 }
+
+// StdKeyGen is a KeyPairGenerator based on randSource and ed25519Deriver.
+var StdKeyGen KeyPairGenerator = stdGenerator{es: randSource{}, kd: ed25519Deriver{}}
 
 // SignHash signs a message using a secret key.
 func SignHash(data Hash, sk SecretKey) (sig Signature, err error) {
--- a/crypto/signatures.go
+++ b/crypto/signatures.go
@@ -21,6 +21,7 @@ type (
 )
 
 var (
+	ErrNilInput         = errors.New("cannot use nil input")
 	ErrInvalidSignature = errors.New("invalid signature")
 	ErrRandUnexpected   = errors.New("unexpected result from random number generator")
 )
--- a/crypto/signatures.go
+++ b/crypto/signatures.go
@@ -37,14 +37,14 @@ type KeyPairGenerator interface {
 // returns the total number of bytes written to the buffer.
 type readBytesFunc func([]byte) (int, error)
 
-// deriveEd25519KeyPairFunc is a function pointer that matches the signature of
+// deriveEd25519Func is a function pointer that matches the signature of
 // ed25519.GenerateKey.
-type deriveEd25519KeyPairFunc func([EntropySize]byte) (ed25519.SecretKey, ed25519.PublicKey)
+type deriveEd25519Func func([EntropySize]byte) (ed25519.SecretKey, ed25519.PublicKey)
 
 // SignatureKeyGenerator is an implementation of KeyPairGenerator.
 type SignatureKeyGenerator struct {
 	readRandBytes readBytesFunc
-	deriveKeyPair deriveEd25519KeyPairFunc
+	deriveKeyPair deriveEd25519Func
 }
 
 // NewSignatureKeyGenerator creates a new SignatureKeyGenerator type that uses
--- a/crypto/signatures.go
+++ b/crypto/signatures.go
@@ -30,7 +30,7 @@ var (
 // public-secret key pairs.
 type KeyPairGenerator interface {
 	Generate() (SecretKey, PublicKey, error)
-	GenerateDetermistically(entropy [EntropySize]byte) (SecretKey, PublicKey)
+	GenerateDetermistic(entropy [EntropySize]byte) (SecretKey, PublicKey)
 }
 
 // readBytesFunc is a function pointer that reads bytes into a buffer and
--- a/crypto/signatures_test.go
+++ b/crypto/signatures_test.go
@@ -36,12 +36,6 @@ func (m *mockKeyDeriver) Derive(b [EntropySize]byte) (sk ed25519.SecretKey, pk e
 	return args.Get(0).(ed25519.SecretKey), args.Get(1).(ed25519.PublicKey)
 }
 
-// Create a testable version of SignatureKeyGenerator whose readRandBytes and
-// deriveKeyPair members are caller-specified.
-func NewTestableSignatureKeyGenerator(readBytes readBytesFunc, deriveKeyPair deriveEd25519KeyPairFunc) SignatureKeyGenerator {
-	return SignatureKeyGenerator{readBytes, deriveKeyPair}
-}
-
 // Test that the Generate method is properly calling its dependencies and
 // returning the expected key pair.
 func TestGenerateRandomKeyPair(t *testing.T) {
@@ -56,7 +50,7 @@ func TestGenerateRandomKeyPair(t *testing.T) {
 	mockEd22519.On("Derive", *new([EntropySize]byte)).Return(derivedSk, derivedPk)
 
 	// Create a SignatureKeyGenerator using mocks.
-	skg := NewTestableSignatureKeyGenerator(mockRandGenerator.Read, mockEd22519.Derive)
+	skg := SignatureKeyGenerator{mockRandGenerator.Read, mockEd22519.Derive}
 
 	// Create key pair.
 	skActual, pkActual, err := skg.Generate()
@@ -81,7 +75,7 @@ func TestGenerateRandomKeyPairFailsWhenRandFails(t *testing.T) {
 	mockEd22519 := new(mockKeyDeriver)
 	mockEd22519.On("Derive", *new([EntropySize]byte)).Return(derivedSk, derivedPk)
 
-	skg := NewTestableSignatureKeyGenerator(mockRandGenerator.Read, mockEd22519.Derive)
+	skg := SignatureKeyGenerator{mockRandGenerator.Read, mockEd22519.Derive}
 	_, _, err := skg.Generate()
 	assert.NotNil(t, err, "Generate should fail when readRandBytes fails")
 }
@@ -99,7 +93,7 @@ func TestGenerateRandomKeyPairFailsWhenRandWritesInsufficientBytes(t *testing.T)
 	mockEd22519 := new(mockKeyDeriver)
 	mockEd22519.On("Derive", *new([EntropySize]byte)).Return(derivedSk, derivedPk)
 
-	skg := NewTestableSignatureKeyGenerator(mockRandGenerator.Read, mockEd22519.Derive)
+	skg := SignatureKeyGenerator{mockRandGenerator.Read, mockEd22519.Derive}
 	_, _, err := skg.Generate()
 	assert.Equal(t, ErrRandUnexpected, err, "Generate should fail when readRandBytes writes insufficient bytes")
 }
@@ -118,7 +112,7 @@ func TestGenerateDeterministicKeyPair(t *testing.T) {
 	mockEd22519 := new(mockKeyDeriver)
 	mockEd22519.On("Derive", entropy).Return(derivedSk, derivedPk)
 
-	skg := NewTestableSignatureKeyGenerator(nil, mockEd22519.Derive)
+	skg := SignatureKeyGenerator{nil, mockEd22519.Derive}
 
 	// Create key pair.
 	skActual, pkActual := skg.GenerateDeterministic(entropy)
--- a/crypto/signatures_test.go
+++ b/crypto/signatures_test.go
@@ -25,7 +25,7 @@ func (m *mockRandGenerator) Read(b []byte) (n int, err error) {
 	return args.Int(0), args.Error(1)
 }
 
-// mockKeyDeriver is a mock implementation of deriveEd25519KeyPairFunc, allowing
+// mockKeyDeriver is a mock implementation of deriveEd25519Func, allowing
 // tests to mock out calls to ed25519.GenerateKey.
 type mockKeyDeriver struct {
 	mock.Mock
--- a/crypto/signatures_test.go
+++ b/crypto/signatures_test.go
@@ -2,58 +2,171 @@ package crypto
 
 import (
 	"crypto/rand"
+	"errors"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+
 	"github.com/NebulousLabs/Sia/encoding"
+	"github.com/NebulousLabs/ed25519"
 )
+
+// mockRandGenerator is a mock implementation of readBytesFunc, allowing tests
+// to mock out calls to rand.Read and use a deterministic implementation
+// instead.
+type mockRandGenerator struct {
+	mock.Mock
+}
+
+func (m *mockRandGenerator) Read(b []byte) (n int, err error) {
+	args := m.Called(b)
+	return args.Int(0), args.Error(1)
+}
+
+// mockKeyDeriver is a mock implementation of deriveEd25519KeyPairFunc, allowing
+// tests to mock out calls to ed25519.GenerateKey.
+type mockKeyDeriver struct {
+	mock.Mock
+}
+
+func (m *mockKeyDeriver) Derive(b [EntropySize]byte) (sk ed25519.SecretKey, pk ed25519.PublicKey) {
+	args := m.Called(b)
+	return args.Get(0).(ed25519.SecretKey), args.Get(1).(ed25519.PublicKey)
+}
+
+// Create a testable version of SignatureKeyGenerator whose readRandBytes and
+// deriveKeyPair members are caller-specified.
+func NewTestableSignatureKeyGenerator(readBytes readBytesFunc, deriveKeyPair deriveEd25519KeyPairFunc) SignatureKeyGenerator {
+	return SignatureKeyGenerator{readBytes, deriveKeyPair}
+}
+
+// Test that the Generate method is properly calling its dependencies and
+// returning the expected key pair.
+func TestGenerateRandomKeyPair(t *testing.T) {
+	// Create a mock random number generator (note that it does not actually
+	// modify its buffer.
+	mockRandGenerator := new(mockRandGenerator)
+	mockRandGenerator.On("Read", mock.Anything).Return(EntropySize, nil)
+
+	derivedSk := ed25519.SecretKey(new([SecretKeySize]byte))
+	derivedPk := ed25519.PublicKey(new([PublicKeySize]byte))
+	mockEd22519 := new(mockKeyDeriver)
+	mockEd22519.On("Derive", *new([EntropySize]byte)).Return(derivedSk, derivedPk)
+
+	// Create a SignatureKeyGenerator using mocks.
+	skg := NewTestableSignatureKeyGenerator(mockRandGenerator.Read, mockEd22519.Derive)
+
+	// Create key pair.
+	skActual, pkActual, err := skg.Generate()
+
+	// Verify that Generate satisfied expectations.
+	assert.Nil(t, err)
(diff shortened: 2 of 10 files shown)


lukechampine:
This PR is really doing two things: refactoring some crypto functions, and adding a whole host of new testing dependencies.

We should have a longer discussion about testing frameworks. In general, we are conservative with our imports, so I'm resistant to adding new tests that depend on `testify`. My feeling is that adopting a testing framework makes it harder for people to contribute, because they have to learn the framework first.
I don't find `testify` tests more readable, and I don't like their heavy use of reflection, variadic functions, etc. Particularly bad are the use of strings to specify function names in `mock.On`, and the use of opaque indices in expressions like `args.Int(0)`. I dislike the separation of `assert` and `require` to delineate `t.Error` and `t.Fatal`; we frequently use both, which would mean creating both `assert` and `require` objects in many tests. Reliance on `Equal` also seems dangerous, especially when dealing with structs and pointers. Better to be explicit about what you are comparing.

I can see the benefit of using `mock` to define the behavior of test objects. Essentially, it allows you to very purely test a function's logic without worrying about its dependencies. I'm just not sure it's worth the cost of a new framework. My gut says that you can replicate 90% of `mock`'s functionality using simple stateful test objects. `mock` requires you to substitute interfaces for objects anyway, so after you do that, the `mock` package is really just for convenience. See modules/renter/upload_test.go and modules/renter/download_test.go for examples of where we're already using mocks.

More specific to this PR:

- The `SignatureKeyGenerator` type looks strange. It's a struct, but its fields are both functions. That sounds like an interface to me!
- You don't need to define functions like `NewTestableSignatureKeyGenerator` to initialize a struct; just use an object literal.
- Use `object{}` instead of `*new(object)`.

The point of all this refactoring and new code, as far as I can tell, is to break up the tight coupling of the crypto functions and their source of randomness (currently `crypto/rand`). This would allow us to test code paths where the crypto functions fail. The problem is that we use crypto functions all over the place, and sometimes they are buried quite deeply in the call stack. For example, in this PR you modified modules/host/host.go to use your new `skg` object. But there's still no way to make `Generate` fail in the host tests!
I see two possible solutions. One is to add a source of randomness to a bunch of functions and/or objects. In the host example, that would mean passing some sort of `crypto.Source` interface to `host.New`. For something more involved, like the gateway, you'd want to make the source a field of the `Gateway` object. Then, during testing, you pass in dummy sources that can fail.
Alternatively, you could add some sort of global toggle to the `crypto` package that would cause its functions to start failing. But this gets tricky when you realize we call many of these functions concurrently. How do you know when to "flip the switch?"
Neither of these options seem particularly attractive to me, but the former is probably the "right" thing to do. After all, the `New` functions already take lots of interfaces. I just worry that there's a slippery slope here, where we wind up mocking anything that involves I/O, including the entire filesystem (there are actually multiple packages that do this, e.g. https://github.com/spf13/afero).

I'll let @DavidVorick weigh in with his thoughts as well.

crypto/signatures.go
```diff
@@ -23,25 +23,58 @@ type (
...
 )
 
-// GenerateKeyPair creates a public-secret keypair that can be used to sign and
-// verify messages.
-func GenerateSignatureKeys() (sk SecretKey, pk PublicKey, err error) {
+// KeyPairGenerator is an interface that allows the caller to generate
+// public-secret key pairs.
+type KeyPairGenerator interface {
+	Generate() (SecretKey, PublicKey, error)
+	GenerateDetermistically(entropy [EntropySize]byte) (SecretKey, PublicKey)
```
DavidVorick: `GenerateDeterministic`

crypto/signatures.go
```diff
@@ -23,25 +23,58 @@ type (
...
+}
+
+// readBytesFunc is a function pointer that reads bytes into a buffer and
+// returns the total number of bytes written to the buffer.
+type readBytesFunc func([]byte) (int, error)
```
DavidVorick: might be good to use a convention like `randRead`

Also, we might want to export this, so that other packages can use it without needing to define their own.

crypto/signatures.go
```diff
@@ -23,25 +23,58 @@ type (
...
+
+// deriveEd25519KeyPairFunc is a function pointer that matches the signature of
+// ed25519.GenerateKey.
+type deriveEd25519KeyPairFunc func([EntropySize]byte) (ed25519.SecretKey, ed25519.PublicKey)
```
DavidVorick: `Ed25519Generator` -> KeyPair is fairly well implied once you are naming a signature algorithm

crypto/signatures.go
```diff
@@ -23,25 +23,58 @@ type (
...
+
+// SignatureKeyGenerator is an implementation of KeyPairGenerator.
+type SignatureKeyGenerator struct {
+	readRandBytes readBytesFunc
+	deriveKeyPair deriveEd25519KeyPairFunc
+}
```
DavidVorick: this is just an interface :P

crypto/signatures.go
```diff
@@ -23,25 +23,58 @@ type (
...
+func (skg SignatureKeyGenerator) Generate() (sk SecretKey, pk PublicKey, err error) {
 	var entropy [EntropySize]byte
-	_, err = rand.Read(entropy[:])
+	written, err := skg.readRandBytes(entropy[:])
 	if err != nil {
 		return
 	}
+	if written != EntropySize {
+		// readRandBytes did not fill the buffer. This should never happen.
+		return sk, pk, ErrRandUnexpected
```
DavidVorick: rand.Read is guaranteed to return an error if `written != EntropySize`, but I guess we can't rely on that now that we're using an abstraction.

crypto/signatures.go
DavidVorick: nvm I see why this is needed

modules/consensus/accept_test.go
```diff
@@ -1349,7 +1351,7 @@ func (cst *consensusSetTester) testPaymentChannelBlocks() error {
 		// txn needs to be fully custom. To get a custom txn, manually create an
 		// address and then use the wallet to fund that address.
 		channelSize := types.NewCurrency64(10e3)
-		channelFundingSK, channelFundingPK, err := crypto.GenerateSignatureKeys()
```
DavidVorick: no need to change the library call.

in the crypto package, define a struct that counts as the 'generic skg':

```
var (
    genericSKG SKG = NewSKG()
)

func GenerateSignatureKeys() stuff {
    return gerenicSKG.Generate()
}
```

crypto/signatures.go
```diff
@@ -23,25 +23,58 @@ type (
...
+	readRandBytes readBytesFunc
+	deriveKeyPair deriveEd25519KeyPairFunc
+}
+
+// NewSignatureKeyGenerator creates a new SignatureKeyGenerator type that uses
+// random data and depends on the ed25519.GenerateKey function for deriving
+// key pairs.
+func NewSignatureKeyGenerator() SignatureKeyGenerator {
+	return SignatureKeyGenerator{rand.Read, ed25519.GenerateKey}
+}
```
DavidVorick: This function is not needed, instead have a `var GenericSignatureKeyGenerateor` defined at the top of the file.

crypto/signatures_test.go
```diff
@@ -2,58 +2,171 @@ package crypto
 
 import (
 	"crypto/rand"
+	"errors"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
```
DavidVorick: I dislike both `asset` and `require`, writing out the full set of things via  t.Error seems fine to me. Also iffy on mock, espeically for smaller stuff writing out the full mock manually seems sufficient.

I'm finally understanding what all this is supposed to look like.

mtlynch:
Thanks for taking the time to review, @lukechampine!

>I don't like their heavy use of reflection, variadic functions, etc. Particularly bad are the use of strings to specify function names in mock.On, and the use of opaque indices in expressions like args.Int(0)

Yeah, I agree with you that the guts are quite ugly. I'm coming off a lot of recent experience with Python's `mock` package which is *extremely* elegant, so seeing `testify/Mock` is depressing, though I think as a language Go is much more rigid and so it's much harder to fiddle around with internals to facilitate testing than it is in Python.

>I can see the benefit of using mock to define the behavior of test objects. Essentially, it allows you to very purely test a function's logic without worrying about its dependencies. I'm just not sure it's worth the cost of a new framework. My gut says that you can replicate 90% of mock's functionality using simple stateful test objects. mock requires you to substitute interfaces for objects anyway, so after you do that, the mock package is really just for convenience. See upload_test.go and download_test.go for examples of where we're already using mocks.

I don't know that we could do better rolling our own. What you're pointing to in upload_test.go and download_test.go seem to be test *fakes* rather than mocks. The difference is that test fakes are just a simpler, more testable implementation of the real object whereas mocks have programmable expectations (more detailed explanation [here](http://martinfowler.com/articles/mocksArentStubs.html)).

I think in a lot of cases in this code, we want to test with mocks, not fakes. In upload_test, for example, it doesn't look like there's validation that `testHost.addPiece` was actually called, which might be fine in that case. But for something like `SignatureKeyGenerator.Generate` it's a pretty big problem if it generated keys without ever calling `rand.Read` because how could its generated keys be secure if it never accessed its source of entropy?

A lot of times, mocks are valuable in making sure that the function was called in the right way. For example, when we have a function that makes a call to a signing interface, we want to verify not only that it calls that interface and properly handles a given result, but that it called that interface correctly (i.e. are we signing the correct data?).

We can implement our own mock features like creating a type that keeps a record of how its functions were called, but then we're basically reimplementing `testify/mock`.

>I don't find testify tests more readable

I think they're a vast improvement over manual equality checks and way more readable and easier to write. Every testing framework I've ever worked with has mocks that are some variant of `assert(expected, actual, message)` so maybe I'm just used to it, but Go's built-in testing seems primitive and crazy to me.

Why would you want to write this:

```
if (a != b) {
  t.Errorf("Expected a==b, but a=%v, b=%v", a, b)
}
```

When you can just write:

```
assert.Equal(a, b)
```

I can live without testify's assert if you guys don't like it, but I think it's way better.

>I dislike the separation of assert and require to delineate t.Error and t.Fatal; we frequently use both, which would mean creating both assert and require objects in many tests.

I don't know if this makes a difference to you but creating the objects is just a convenience. You can also call like this:

```
assert.Equals(t, a, b)
require.Equals(t, a, b)
```

>Reliance on Equal also seems dangerous, especially when dealing with structs and pointers. Better to be explicit about what you are comparing.

I'm not clear on what you mean here. Aren't the following two snippets equally precise about what they're comparing?

```
if (a != b) { t.Error("a!=b") }
```
```
assert.Equal(a, b)
```

>The SignatureKeyGenerator type looks strange. It's a struct, but its fields are both functions. That sounds like an interface to me!

@DavidVorick said the same thing, but then I think it made sense for him after he spent some time with it. It's a struct whose fields are function *pointers* which is different from an interface.

> You don't need to define functions like NewTestableSignatureKeyGenerator to initialize a struct; just use a literal.

True, fixed in c666d24.

> Prefer object{} to *new(object).

Fixed in 226ade0.

>The point of all this refactoring and new code, as far as I can tell, is to break up the tight coupling of the crypto functions and their source of randomness (currently crypto/rand).

This is exactly correct.

>This would allow us to test code paths where the crypto functions fail.

Correct, but only for the code where we successfully break up the tight coupling. In this case, it allows us to unit test code paths within `SignatureKeyGenerator` but it doesn't improve testability of `SignatureKeyGenerator`'s clients much.

> in this PR you modified modules/host/host.go to use your new skg object. But there's still no way to make Generate fail in the host tests!

Correct, you can't make `Generate` fail from modules/host/host.go, but that's beyond the scope of this PR as it's a separate problem that `host.go` is tightly coupled with `SignatureKeyGenerator.Generate` (rather than a dependency on an interface, like `KeyPairGenerator`).

> I just worry that there's a slippery slope here, where we wind up mocking anything that involves I/O, including the entire filesystem (there are actually multiple packages that do this, e.g. https://github.com/spf13/afero).

I don't want to mock anything that involves I/O. I want to mock **[everything](https://www.youtube.com/watch?v=KBKXu3Kg4yg)**! *Especially* the filesystem.

It's not as bad as you think...

> we use crypto functions all over the place, and sometimes they are buried quite deeply in the call stack.

We only mock out a unit's immediate dependencies, so if it's more than 1 level deep, we don't have to worry about it.

Let's take the example of host.go. Our production code might look like:

```
host.go : New()
 -> SignatureKeyGenerator.Generate (via KeyPairGenerator interface)
   -> rand.Read (via readBytesFunc)
```

But our test would just look like:

```
host.go : New()
 -> MockSignatureKeyGenerator.Generate (via KeyPairGenerator interface)
```

This is the benefit of loose coupling. Each class communicates with its dependencies through well defined interfaces and the lower level implementation details of those dependencies don't matter.

Back to filesystems: We *really* want to mock out filesystem calls. The filesystem is shared global state across tests so it's very easy for tests to taint one another via the filesystem (i.e. generate false positives or negatives). Plus it generally makes the test loop much slower.

crypto/signatures.go
mtlynch: We can be flexible if we want to. I was being especially rigorous in this PR about checking every possible behavior of `rand.Read` to demonstrate the value of breaking up dependencies and the increased coverage we can get that we can't get with other types of testing.

I don't know that I'd do this normally, given that it's unlikely we'll ever use any implementation other than `rand.Read` which provides this guarantee.

crypto/signatures_test.go
mtlynch: I'm beginning to think you and @lukechampine have Stockholm Syndrome. There's no other explanation for your preference for go's built in testing package, which to me is horrible. I've added more details on the long reply to Luke, but tl; dr - I think we need `mock`, I prefer `assert/require` but I defer to your preferences.

modules/consensus/accept_test.go
mtlynch: I'd actually prefer to avoid this if possible.

I eventually want to switch all the `skg` objects to `KeyPairGenerator` interfaces so that we break up tight coupling between `SignatureKeyGenerator` and its clients (and can therefore test its clients more easily by mocking out `KeyPairGenerator`).

It's not a huge deal because the code delta between the two solutions is small, but I feel like the current code is a step closer to where we want to be. I can change it to your suggestion if this is something you feel strongly about.

crypto/signatures.go
mtlynch: Good catch, fixed 6a2e405f88b491e8154db2225

crypto/signatures.go
mtlynch: True, fixed 518f0f1d23cdde0ce1b60435ffa7022228f653ed

mtlynch:
Thanks for the review @DavidVorick!

One more note for both of you regarding `testify/mock`. It's a lot cleaner when you're mocking out an interface rather than a function pointer. It doesn't have great support for mocking out functions (or maybe I'm just not aware of how to do it).

But interfaces are much better. For example, if we wanted to generate a mock for the `KeyPairGenerator` interface, we wouldn't have to write the ugly `m.Called(0)`, `args.Get(0)` code by hand because the [mockery tool](https://github.com/vektra/mockery) will just generate it for us and write it directly to its own file in a `mocks/` directory where its ugliness is isolated from the rest of our code.

crypto/signatures.go
DavidVorick: I'm fine with verbose error checking. I generally find it to be readable.

modules/consensus/accept_test.go
DavidVorick: It's how `crypto/rand` works. There's a `var Rand io.Reader` that provides cryptographic entropy. Calling `rand.Reader.Read()` and `rand.Read()` actually do the same exact thing - `rand.Read()` is just a passthrough function.

I think this style makes sense. There's no reason to force people to call `skg := crypto.NewSKG(); skg.Generate()`, calling `crypto.Generate()` should be sufficient, especially since that's practically speaking going to be 100% of the non-testing use cases.

DavidVorick:
```
if (a != b) {
  t.Errorf("Expected a==b, but a=%v, b=%v", a, b)
}
```

You've got a point there, that's pretty exhausting to write. Luke and I don't like change, and in general hold that the go standard library is quite powerful. We also have a strong aversion to dependencies.

DavidVorick:
https://groups.google.com/forum/#!topic/golang-nuts/qjzflfHpsIc

re: StockholmSyndrome:

Unlike the standard libraries for many languages, golang has a very strong standard library. The functions are highly composable, and while it sometimes leads to more verbose code, you end up escaping the clutches of opinionated libraries. I would definitely disagree that go's testing library is terrible, it's the best testing framework I've ever worked with. Setting up your first test is dirt simple.

https://groups.google.com/forum/#!topic/golang-nuts/UR4hEMehy1c

We are fully on board with the mocking, but still figuring out the best way to integrate it into a golang project. I believe Luke will be going into more detail about things that other go projects are doing. I think the `crypto` package is a great place to figure all of this out because there's not a whole lot of infrastructure to write, we can iterate pretty rapidly over the design schemes.

crypto/signatures.go
```diff
@@ -23,25 +23,58 @@ type (
...
+// KeyPairGenerator is an interface that allows the caller to generate
+// public-secret key pairs.
+type KeyPairGenerator interface {
+	Generate() (SecretKey, PublicKey, error)
+	GenerateDetermistic(entropy [EntropySize]byte) (SecretKey, PublicKey)
+}
+
+// readBytesFunc is a function pointer that reads bytes into a buffer and
+// returns the total number of bytes written to the buffer.
+type readBytesFunc func([]byte) (int, error)
```
DavidVorick: this is actually just an `io.Reader`, you shouldn't need to define this function. In the struct, just have an `io.Reader` as one of the fields.

crypto/signatures.go
```diff
@@ -23,25 +23,58 @@ type (
...
+
+// deriveEd25519Func is a function pointer that matches the signature of
+// ed25519.GenerateKey.
+type deriveEd25519Func func([EntropySize]byte) (ed25519.SecretKey, ed25519.PublicKey)
+
+// SignatureKeyGenerator is an implementation of KeyPairGenerator.
+type SignatureKeyGenerator struct {
+	readRandBytes readBytesFunc
```
DavidVorick: `entropySource io.Reader`

crypto/signatures.go
```diff
@@ -23,25 +23,58 @@ type (
 var (
 	ErrNilInput         = errors.New("cannot use nil input")
```
DavidVorick: ErrNilInput is no longer used

lukechampine:
Check out this talk by SoundCloud: https://www.youtube.com/watch?v=Y1-RLAl7iOI
(And the corresponding writeup: http://peter.bourgon.org/go-in-production/)

Their tl;dr is: "We tried a bunch of ways of doing things, but inevitably circled back to just using the standard library." Go is unlike most languages these days in that its standard library is really good -- it was designed by experienced Google engineers *specifically for use in production environments*. Whenever I find that my code is at odds with the standard library, I ask myself why they might have designed it that way. And 99% of the time, it turns out that they made the wiser decision.
For example, `package testing` does not include assert functions. Why not? Obviously they didn't just forget to implement them. They must have decided that they did more harm than good. As mentioned in the discussion David linked to, asserts aren't very composable, and encourage lazy error handling. See https://golang.org/doc/faq#testing_framework. In Go, the preferred style is to use table-driven tests, as seen in https://golang.org/src/fmt/fmt_test.go.

>I think as a language Go is much more rigid and so it's much harder to fiddle around with internals to facilitate testing than it is in Python.

It's a trade-off. Dynamic languages make it easy to swap things out at runtime, but the cost is that you lose type safety, which is hugely important for large projects.

>I don't know that we could do better rolling our own. What you're pointing to in upload_test.go and download_test.go seem to be test *fakes* rather than mocks... I think in a lot of cases in this code, we want to test with mocks, not fakes.

True, they are fakes. However, fakes are often sufficient to test function logic. In `TestErasureUpload`, we don't need to manually check that `addPiece` is called; if it isn't called, the data won't be "uploaded" and the test will fail.
I agree that sometimes mocks are useful, as in the `rand.Read` example you gave. But you don't need to bring in a whole testing framework just to confirm that a function was called. Just stick a `called` boolean in the test object. The meaning is clearer anyway; to someone unfamiliar with `mock`, it's not obvious that `AssertExpectations` verifies that a function was called.

>I'm not clear on what you mean here. Aren't the following two snippets equally precise about what they're comparing?

For types that support `==`, yes. But it's trickier when you're dealing with more complicated objects. For example, in some cases you might want to follow pointers, and compare what they're pointing to, but in another case you might want to compare the pointers themselves. Similarly, you might want to check that two slices contain the same elements, but not necessarily in the same order. So highly-generalized functions like `assert.Equal` are dangerous because their definition of equality may not match up with the programmer's expectations.
(FWIW, SoundCloud does advocate using `reflect.DeepEqual`, which I imagine is what `assert.Equal` does behind the scenes.)

>It's a struct whose fields are function pointers which is different from an interface.

Ok. It's starting to come together in my head. We want a `crypto.KeyPairGenerator` interface that we can pass to functions like `host.New`, so that we can mock it in `package host`. But `crypto.KeyPairGenerator` has its own dependencies, which we want to mock in `package crypto`. Your implementation makes a lot more sense now. Here's my stab at it, using interfaces instead of function pointers:
```go
// this will be mocked outside the crypto package
type KeyPairGenerator interface {
    Generate()
    GenerateDeterministic()
}

// these will be mocked inside the crypto package
type entropySource interface {
    getEntropy() (EntropySize]byte, error)
}
type keyDeriver interface {
    deriveKey([EntropySize]byte) (SecretKey, PublicKey)
}

// inside the crypto package, we only have one implementation of SignatureKeyGenerator
type stdGenerator struct{
    es entropySource
    kd keyDeriver
}
func (sg stdGenerator) Generate() { /* use es.getEntropy and kd.deriveKey */ }
func (sg stdGenerator) GenerateDeterministic() { /* use kd.deriveKey */ }

// now we define production implementations of entropySource and keyDeriver
type randSource struct{}
func (rs randSource) getEntropy() { /* use rand.Read */ }
type ed25519Deriver struct{}
func (ed ed25519Deriver) deriveKey() { /* use ed25519.GenerateKey */ }

// and lastly, the production implementation of SignatureKeyGenerator
var StdKeyGen SignatureKeyGenerator = stdGenerator{es: randSource{}, kd: ed25519Deriver{}}
```
Finally, inside the test files, we mock the `entropySource` and `keyDeriver` interfaces, possibly using a single type:
```go
type mockSourceAndDeriver struct {
    // fields necessary for mocking
}
func (m *mockSourceAndDeriver) getEntropy() { /* mock stuff */ }
func (m *mockSourceAndDeriver) deriveKey()  { /* mock stuff */ }
```
Perhaps it is excessive to mock out individual functions like this. The advantage is consistency and clarity: mocking works the same way, instead of sometimes being an interface and sometimes being a function pointer.

>the mockery tool will just generate it for us and write it directly to its own file in a mocks/ directory

No thanks. For now, we aren't doing any codegen, especially from an external package.

I will follow up when I get home with a plan for when, why, and how we should approach mocking and faking.

crypto/signatures.go
mtlynch: Fixed in 6a17584

crypto/signatures.go
mtlynch: I don't think that works because io.Reader is an interface that contains a function that matches this signature, but we need a function pointer for the signature. I tried changing the type of `SignatureKeyGenerator.readRandBytes` and I get:

```
crypto/signatures.go:54: cannot use "crypto/rand".Read (type func([]byte) (int, error)) as type io.Reader in field value:
        func([]byte) (int, error) does not implement io.Reader (missing Read method)
crypto/signatures.go:61: cannot call non-function skg.readRandBytes (type io.Reader)
crypto/signatures_test.go:53: cannot use mockRandGenerator.Read (type func([]byte) (int, error)) as type io.Reader in field value:
        func([]byte) (int, error) does not implement io.Reader (missing Read method)
crypto/signatures_test.go:78: cannot use mockRandGenerator.Read (type func([]byte) (int, error)) as type io.Reader in field value:
        func([]byte) (int, error) does not implement io.Reader (missing Read method)
crypto/signatures_test.go:96: cannot use mockRandGenerator.Read (type func([]byte) (int, error)) as type io.Reader in field value:
        func([]byte) (int, error) does not implement io.Reader (missing Read method)
```

crypto/signatures.go
mtlynch: Don't think this can work, see note above.

modules/consensus/accept_test.go
mtlynch: I'm not sure if we're talking about the same thing. It sounds like you're advocating this:

```
type SKGClient struct {
  // some members
}

func (c SKGClient) Foo() {
  sk, pk, err := crypto.GenerateSignatureKeys()
  ...
}
```

How do we test `SKGClient.Foo()` in that case? It's tightly bound to `crypto.GenerateSignatureKeys` so we can't mock it out.

I'm saying I'd ultimately like to get to this:

```
type SKGClient struct {
  // some members
  signatureKeyGenerator KeyPairGenerator
}

func (c SKGClient) Foo() {
  sk, pk, err := signatureKeyGenerator.Generate()
  ...
}
```

That way we can easily test `SKGClient` by initializing it with a mock object for `signatureKeyGenerator `.

I prefer what's currently in the PR because it's closer to this ultimate goal.

crypto/signatures.go
mtlynch: addressed in the note in accept_test.go

crypto/signatures.go
mtlynch: ...and backed out in 62f4f77

```
modules/renter/upload_test.go:18: undefined: crypto.ErrNilInput
```

mtlynch:
Thanks for writing that up @lukechampine!

You've convinced me to give up on `stretchr/testify` and try reimplementing with the native standard library. I guess with table-driven tests, the tedium of writing out asserts longhand is not that bad.

I still feel like mocks are going to be a pain in the ass to write by hand, but I'm willing to give it a try.

I like your rewrite of the interfaces to get rid of function pointers. I'll integrate that into this PR.

lukechampine:
Ah, that is a relief. I'm glad we were able to reach a satisfying middle ground. I want to make it very clear that we value your input; it's certainly shifted our opinions on the subject. I'm now convinced of the importance of mocking. The tests themselves are important, but even more important is how it forces you to be explicit with your dependencies. In a way, it's similar to the purely functional style of code you'd see in a language like Haskell:
```go
// FP enthusiasts would say this function is bad because it modifies an object
// that wasn't passed as an argument.
func foo(x int) int {
    return x + myGlobalVar
}
// better version:
func foo(x, y int) int {
    return x + y
}

// TDD enthusiasts would say this function is bad because it uses a dependency
// that wasn't passed as an argument.
func bar(x int) {
    mypackage.Use(x)
}
// better version:
func bar(x int, u User) {
    u.Use(x)
}
```
Obviously, there are sane limits to how far you should take this. For example, I don't see much payoff in mocking the functionality of `package encoding` (maybe you disagree?). But I'm much more onboard with the general idea after viewing it in this light. I even support your plan to mock the filesystem.

The biggest targets for dependency injection (if I'm using that term correctly...) are the modules. They are in decent shape already, because they interact with each other via interfaces, but we can go further. As discussed previously, this would mean adding things like `KeyPairGenerator` as an argument to `host.New`. I'm thinking we may want to simplify initialization by having `host.New` use the production `KeyPairGenerator`, and call a companion function (`newHost`?) which takes ALL the parameters. e.g.:
```go
func New(addr string, saveDir string) (*Host, error) {
    // create production modules, kpg, net.Listener, filesystem...
    return newHost(cs, hdb, tpool, wallet, kpg, l, fs, addr, saveDir)
}
```
And then we test by supplying mocked versions to `newHost`. As a consequence, `New` isn't really tested, but that's not so bad, really. I'm not sure if import cycles prevent us from calling `consensus.New` inside `host.New` -- that would be a bummer.

Some of these mocked interfaces, especially the I/O stuff, will likely be shared across multiple packages. I imagine we'll want to define them in `package build` to avoid reimplementing them in every package that needs them.

All this talk of dependencies is reminding me of one of my previous ideas regarding modules. Basically, instead of `host.New` taking a `modules.ConsensusSet`, it would take a `host.consensusSet`: a restricted interface containing only the methods that the host needs. This makes it easier to mock modules, which are generally very complex. It also removes the need to define all the module interfaces in a central location; instead, we treat the real ConsensusSet as the sum of the interfaces that it implements. The only real downside of this is that it can be nice to have all the methods enumerated in one place. Maybe we could compromise and leave the current interface definitions, while also slowly replacing them with restricted interfaces inside the modules.

From what I can tell, there are two big impediments to rigorous testing of our codebase:
1. Modules like the wallet, transaction pool, and consensus set are too complex to mock.
2. Most modules spawn goroutines, which are difficult for the caller to access (and therefore test).

I think we can reduce the severity of 1 by using restricted interfaces, and by mocking multiple modules from the same object. For example, it would be tough to create a mock wallet that creates outputs that the mock tpool and mock consensus set will accept. But if all three modules have access to the same underlying type, you can have the wallet stick its output in a field of the type, such that the tpool and consensus set know to accept it. (I'm probably vastly oversimplifying things here...maybe David can provide a more legitimate example.)

I'm less sure about how to approach 2. The Gateway in particular is tough to test, because it spawns a lot of goroutines, some of which loop every n seconds, and they're tested by sleeping for n+Δ seconds -- gross. It's really tough to deterministically test concurrent code without bloating the goroutines with extra parameters and code that "phones home" in some way or another. Open to suggestions here.

One dangerous result of mocking everything is that it can hide problems that would arise in production. For example, right now, David is wrestling with a strange bug in boltdb that causes entries to be deleted. If all our tests used a mocked in-memory database, we wouldn't be made aware of this bug until we ran it in production and encountered it at random. So we need to take care not to isolate ourselves too completely from our dependencies. I suggest we take a fresh look at integration testing in order to crack down on this class of bugs.

That's all I got. I'm excited to see how our module architecture shifts to enable more rigorous testing. Thanks for getting the ball rolling!

crypto/signatures_test.go
```diff
@@ -2,24 +2,138 @@ package crypto
...
+// Test that the Generate method is properly calling its dependencies and
+// returning the expected key pair.
+func TestGenerateRandomKeyPair(t *testing.T) {
+	var mockEntropy [EntropySize]byte
+	mockEntropy[0] = 0x0a
+	mockEntropy[EntropySize-1] = 0x0b
+	es := mockEntropySource{entropy: mockEntropy}
+
+	sk := SecretKey([SecretKeySize]byte{})
+	pk := PublicKey([PublicKeySize]byte{})
```
mtlynch: re-reading this now, we should initialize `sk` and `pk` to non-nil values so we get better confidence `stdGenerator` is setting these values correctly

mtlynch:
I've gotta head to bed but I'll hopefully have time to continue on this tomorrow. Looks like I introduced a race condition somewhere.

crypto/signatures_test.go
lukechampine: btw, can't you just use `var sk SecretKey` here instead of casting from a byte slice?

DavidVorick:
nah that's just muxado giving travis a bad time. We're not sure what causes it, but it fails infrequently. It's a medium priority to fix it.

modules/consensus/accept_test.go
DavidVorick: ok. It's not how the go standard library works, but it's good enough for me. I see the fear, though there's not much room to introduce error when you've just made a passthrough function.

crypto/signatures_test.go
```diff
@@ -65,11 +180,11 @@ func TestSigning(t *testing.T) {
...
 	for i := 0; i < iterations; i++ {
-		// Generate the keys.
-		sk, pk, err := GenerateSignatureKeys()
-		if err != nil {
-			t.Fatal(err)
-		}
+		// Create dummy key pair.
+		var entropy [EntropySize]byte
+		entropy[0] = 0x05
+		entropy[1] = 0x08
```
DavidVorick: can also just do `entropy[0] = 5` -> it's the same thing as 0x05

crypto/signatures.go
```diff
@@ -25,26 +25,74 @@ var (
...
+// random values.
+type entropySource interface {
+	getEntropy() ([EntropySize]byte, error)
+}
+
+// keyDeriver allows the caller to generate a public-secret key pair based on
+// provided entropy.
+type keyDeriver interface {
+	deriveKeyPair([EntropySize]byte) (SecretKey, PublicKey)
+}
```
DavidVorick: even though this is the same as `type keyDeriver func(...) (..., ...)`, I think wrapping in an interface is more idiomatic. I'm comfortable with either though.

crypto/signatures.go
```diff
@@ -25,26 +25,74 @@ var (
...
+
+// stdGenerator is an implementation of KeyPairGenerator, allowing the caller
+// to generate public-secret key pairs.
+type stdGenerator struct {
+	es entropySource
+	kd keyDeriver
+}
```
DavidVorick: you can sandwich this chain of types inside of the 

```
type(
)
```

declaration. I usually do that when there are lots of types.

lukechampine:
for future reference, commit messages should always be in the imperative voice, i.e. "fix foo" rather than "fixed foo," "fixes foo," etc.

crypto/signatures.go
```diff
@@ -25,26 +25,74 @@ var (
...
+// randSource is an implementation of entropySource that uses rand.Read to
+// generate random bytes.
+type randSource struct{}
+
+// getEntropy returns an array of bytes with random values.
+func (rs randSource) getEntropy() (entropy [EntropySize]byte, err error) {
+	if _, err := rand.Read(entropy[:]); err != nil {
+		return entropy, err
+	}
+	return entropy, nil
```
lukechampine: be consistent with named return values. In this case, I would leave off the names. I would also explicitly return `[EntropySize]byte{}` in the error case.

crypto/signatures.go
```diff
@@ -25,26 +25,74 @@ var (
...
+}
+
+// ed25519Deriver is an implementation of keyDeriver that uses
+// ed25519.GenerateKey to derive keys.
+type ed25519Deriver struct{}
+
+// deriveKeyPair derives a public-secret key pair derived from the provided
+// array of bytes.
+func (ed ed25519Deriver) deriveKeyPair(entropy [EntropySize]byte) (sk SecretKey, pk PublicKey) {
```
lukechampine: ditto, named return values not necessary

crypto/signatures_test.go
```diff
@@ -2,24 +2,138 @@ package crypto
...
 
+	// Verify the dependencies were called correctly
+	if !es.called {
+		t.Error("entropySource was never called.")
+	}
+	if !kd.called {
+		t.Error("keyDeriver was never called.")
+	}
+	if mockEntropy != kd.entropy {
+		t.Error("keyDeriver was called with the wrong entropy. expected = %v, actual = %v", mockEntropy, kd.entropy)
```
lukechampine: should be `Errorf`

DavidVorick:
I was experimenting with different ways of doing mocking, and I came up with this example. I think it's quite clean:

http://play.golang.org/p/8hJ1kaXdwL

crypto/signatures.go
mtlynch: I understand why we don't want named return values for `deriveKeyPair` but why wouldn't we want them in `getEntropy`? It otherwise forces us to add a `var entropy [EntropySize]byte` line.

DavidVorick: If the read is partial, you want to return all zeroes. It's possible that you get a partial read (like 3 bytes) and then an error gets returned. When that happens, you don't want to return an array filled with partial data (it's just one of our conventions - if there's an error then all return values should be empty)

DavidVorick: as for the named returns, I'd probably use a named return here.

lukechampine:
The whole point of this is that we want to create mock objects with the `DoubleRename` method. Without an interface that contains `DoubleRename`, we can't do that. The best we can do is create a `FileSystem` using mocked dependencies, but then we can't control the behavior of `DoubleRename`.

crypto/signatures.go
lukechampine: ah, I missed that reading into `entropy` would require another declaration

crypto/signatures.go
lukechampine: ditto

lukechampine:
David and I talked about it some more, and we have the following (hopefully final) suggestions:
- Remove the exported interface (`KeyPairGenerator`). Instead, we're going to define interfaces only where they're used. So `package host` will define its own interface for generating keys. This is nice because the host only ever uses `Generate`, not `GenerateDeterministic`, so it will only need to mock one function instead of two. We also plan to do this with the (currently exported) module interfaces.
- Modify `getEntropy` and `deriveKey` to be "pure." Essentially, we want to avoid wrapping our dependencies in more code, because then we have to jump through additional hoops to test them. For example, how do we test the production `getEntropy` method? It would be better to replace the `entropySource` method with a simple `io.Reader`, because that shifts code from `getEntropy` to `Generate`, which is what we really want to test.
- `deriveKey` is a little trickier. Since the `ed25519` package exports a function, not an object, the "purest" approach would be to just use the function instead of wrapping it in a new type in order to satisfy a new interface. However, this creates additional complications during testing. For now, I think it's okay to leave the wrapper, but it should revert to returning the `ed25519` types instead of `SecretKey` and `PublicKey`, to match the wrapped function as closely as possible. (`deriveKey` should be a single line, `return ed25519.GenerateKey(entropy)`

Apologies for how long we've drawn this out. We want to do this right the first time, so we're iterating pretty rapidly.

mtlynch:
> Obviously, there are sane limits to how far you should take this. For example, I don't see much payoff in mocking the functionality of package encoding (maybe you disagree?).

For me it depends on what functionality we're using. Like if we have a function that encodes to JSON, it's not necessary to mock out the encoding API because it's pretty easy to just check that the end result is the expected JSON. But for something like encoding to binary or base64, it's more of a pain to test the output and I think it makes sense to mock out the call there.

> Remove the exported interface (KeyPairGenerator). Instead, we're going to define interfaces only where they're used. That means package host will define its own interface for generating keys. This is nice because the host only ever uses Generate, not GenerateDeterministic, so it will only need to mock one function instead of two. We also plan to do this with the (currently exported) module interfaces.

I like this. I worry that it may create headaches if we have interfaces that are basically identical in each package and then we're redundantly redeclaring them, but we can always revisit this decision if that happens.

> Modify getEntropy and deriveKey to be "pure." Essentially, we want to avoid wrapping our dependencies in more code, because then we have to jump through additional hoops to test them. For example, how do we test the production getEntropy method? It would be better to replace the entropySource method with a simple io.Reader, because that shifts code from getEntropy to Generate, which is what we really want to test.

I think I know what you're saying. Don't we still have the same problem though because we need to wrap `rand.Read` in an type which we don't test? Or is there a way of treating `rand.Read` as an `io.Reader` directly? I'm going to try implementing this now and maybe it'll become clearer to me.

> deriveKey is a little trickier. Since the ed25519 package exports a function, not an object, the "purest" approach would be to just use the function instead of wrapping it in a new type in order to satisfy a new interface. However, this creates additional complications during testing. For now, I think it's okay to leave the wrapper, but it should revert to returning the ed25519 types instead of SecretKey and PublicKey, to match the wrapped function as closely as possible. In other words, deriveKey should be a single line, return ed25519.GenerateKey(entropy)

Makes sense. I don't think we need to worry about unit testing every single thing; there are naturally going to be gaps that are not practical to unit test. I think what you're suggesting as a way of minimizing untested logic is a good strategy.

>Apologies for how long we've drawn this out. We want to do this right the first time, so we're iterating pretty rapidly.

No problem. I think we're making a lot of progress and it's helping me get a better understanding of Go.

crypto/signatures_test.go
mtlynch: Fixed

crypto/signatures_test.go
mtlynch: I'm just used to assigning hex values to byte types, but I can do normal base-10 if you guys prefer. Fixed.

modules/consensus/accept_test.go
mtlynch: I'm open to a different strategy but I don't know of any alternative that allows us to unit test the signature generator's clients. How can we test it if the function calls `crypto.GenerateSignatureKeys` directly?

crypto/signatures_test.go
mtlynch: This no longer applies because the code has changed, but I think that's probably true. I'm still getting used to Go's typing / auto initialization / conversions.

crypto/signatures.go
mtlynch: Fixed

crypto/signatures.go
mtlynch: Good suggestion, fixed.

crypto/signatures.go
mtlynch: Fixed

mtlynch:
> Or is there a way of treating rand.Read as an io.Reader directly? I'm going to try implementing this now and maybe it'll become clearer to me.

I didn't realize that rand exported a `Reader` type, so I've used that. Switching to `io.Reader` makes it a lot cleaner, good suggestion.

crypto/signatures_test.go
```diff
@@ -1,25 +1,141 @@
...
-	sk, pk, err := GenerateSignatureKeys()
+	skActual, pkActual, err := g.Generate()
+
+	// Verify that we got back the expected results.
 	if err != nil {
-		t.Fatal(err)
+		t.Error(err)
+	}
+	if *sk != skActual {
+		t.Errorf("Generated secret key does not match expected! expected = %v, actual = %v", sk, skActual)
```
DavidVorick: A shortcut I've taken frequently is to write a non-helpful error message when the test fails, with the intention of writing a more helpful error message if needed in the future.

90% of the checks you make in a test are never triggered, which is why I'm comfortable doing that. The other reason I'm comfortable doing that is because the context of the test is usually pretty obvious, and you have to look at the test code that failed anyway.

As a side effect, you can sometimes tell which tests I had trouble getting to pass because those tests will have much more verbose error messages.

crypto/signatures_test.go
```diff
@@ -1,25 +1,141 @@
...
+
+func (kd *mockKeyDeriver) deriveKeyPair(entropy [EntropySize]byte) (ed25519.SecretKey, ed25519.PublicKey) {
+	kd.called = true
+	kd.entropy = entropy
+	return kd.sk, kd.pk
+}
+
+// Test that the Generate method is properly calling its dependencies and
+// returning the expected key pair.
+func TestGenerateRandomKeyPair(t *testing.T) {
```
DavidVorick: Naming convention: `TestUnitXXX` for unit tests. 

You can run `make cover-unit` to see the code coverage for the unit tests, and `make cover-integration` to see the coverage for the integration tests. And `make cover` will show you for all of them.

crypto/signatures_test.go
```diff
@@ -1,25 +1,141 @@
...
+}
+
+// Test that the Generate method fails if the call to entropy source fails
+func TestGenerateRandomKeyPairFailsWhenRandFails(t *testing.T) {
+	fr := failingReader{err: errors.New("mock error from entropy reader")}
+	g := stdGenerator{entropySource: &fr}
+	if _, _, err := g.Generate(); err == nil {
+		t.Error("Generate should fail when entropy source fails.")
+	}
+}
```
DavidVorick: Should also check that the keys which get returned are blank.

DavidVorick:
61 comments whew.

This is looking a lot better as a framework though. I'm sure we'll continue to make improvements to the mocking style over time but I think we have a strong foundation at this point.

crypto/signatures_test.go
mtlynch: The unit test should verify that the function under test satisfies its contract. Do you feel like returning blank keys on error is part of the function contract or an implementation detail?

I feel like it's an implementation detail and not a part of the contract. I think it's a reasonable practice to return `nil` values for the other return values on error, as it's extra assurance if a caller forgets to check an `error` return value, but I don't think that we want callers or unit tests to depend on this behavior.

My understanding is that when functions return a non-`nil` error value, their other return values are undefined unless otherwise specified. If you guys have an established convention that contradicts that, I'll change it, but if we're deciding precedent here, I'd like to treat other return values when the `error` value is non-`nil`.

What do you guys think?

lukechampine: hmm. I'm not sure on this one, but I think I side slightly more with mtlynch here, the reason being that only *our* functions/interfaces (not those defined in external packages) would adhere to the convention. So we can't ever enforce it for all `io.Reader`s, because that would conflict with its preexisting contract as defined in `package io`.
Of course, it's still good practice to return empty values, but we never want to *depend* on that happening, because in many cases its out of our hands.

DavidVorick: The test would just be to check that `Generate()` only returns blank values, regardless of what behaviors the dependencies might be expressing. It is how `Generate()` is supposed to operate, even though other code shouldn't rely on that.

It's a small point. No reason to hold up the PR for that

crypto/signatures_test.go
mtlynch: Fixed in b81cfbb

crypto/signatures_test.go
mtlynch: I think we're all agreed that it's not necessary in this PR, but it is something I think is good to come to agreement on in general.

I'm *pretty* sure we're in agreement that returning `nil` values when the `error` is non-`nil` is good practice, but not part of the function's contract.

It seems like @DavidVorick is saying this kind of testing is a "nice to have, but not necessary" but I think it's actually counterproductive to do this kind of testing and we should avoid it in our tests. Testing implementation details violates the TDD principle of ["test the interface, not the behavior"](http://googletesting.blogspot.com/2013/08/testing-on-toilet-test-behavior-not.html). A failing unit test should be a sign to the developer that they're changing the class/function's behavior, so I recommend avoiding practices that add noise to this signal.

mtlynch:
You guys know what's going on with this? Why is it saying there's all these changes that aren't there? I think I must have botched the merge from master, but I'm not sure what the issue is.

lukechampine:
Standard practice is to rebase PRs instead of merging. I suggest removing the merge commit and instead rebasing from the feature branch with `git rebase master`. If all else fails, you can use `git reflog` to revert to before the merge.

mtlynch:
Failing I think with the false positive muxado fail, but otherwise I think we're good to go.

DavidVorick:
Looks good to me

lukechampine:
if it's not too much trouble, I'd really prefer that you fix up the commit history first. Using git properly is a skill worth developing. Use `git reflog` and `git reset --hard` to revert to before the merge, then carefully rebase onto master, resolving the conflict in accept_test.go. If done correctly, you won't need any "fixup" or merge commits afterwards.

mtlynch:
Okay, should be all good now.

lukechampine:
Woo!

PR #802, opened 2015-10-03
Demoting locks

DavidVorick:
This PR changed shapes a few times.

I wrote a new 'demotelock' package, and switched both the consensus set and the transaction pool over to that.  It doesn't have deadlock safety, but I don't think that deadlock safety is necessary anymore for those packages. I feel a lot more comfortable debugging a call stack, and the number of threads in our call stack is not as imposing as it used to be. (less than a hundred as opposed to more than a thousand). That combined with much stricter locking conventions means a lot less headache.

Eventually we can probably get rid of the 'sync' package altogether.

I did end up adding testing for the sync package, but that was before I decided that we didn't need it anymore. Since the tests are already written... might as well keep them, especially since we have several packages still using the safelocks.

This PR also comments out the sanity check that causes bolt so much trouble.

This PR should be all that's needed (besides the version PR) to settle 0.4.3. Unlocking the wallet during IBD takes time, but doesn't freeze or crash the program, or trigger race conditions. Closing Sia unexpectedly will not corrupt the database.


modules/consensus/accept.go
```diff
@@ -86,9 +86,7 @@ func (cs *ConsensusSet) validHeader(tx *bolt.Tx, b types.Block) error {
 	if b.Timestamp > types.CurrentTimestamp()+types.FutureThreshold {
 		go func() {
 			time.Sleep(time.Duration(b.Timestamp-(types.CurrentTimestamp()+types.FutureThreshold)) * time.Second)
-			lockID := cs.mu.Lock()
-			defer cs.mu.Unlock(lockID)
-			cs.acceptBlock(b) // NOTE: Error is not handled.
+			go cs.AcceptBlock(b) // NOTE: Error is not handled.
```
lukechampine: no need to spawn an extra goroutine here

modules/consensus/consensusset.go
```diff
@@ -89,7 +89,7 @@ func New(gateway modules.Gateway, persistDir string) (*ConsensusSet, error) {
 		dosBlocks: make(map[types.BlockID]struct{}),
 
 		persistDir: persistDir,
-		mu:         sync.New(modules.SafeMutexDelay, 1),
+		mu:         demotemutex.DemoteMutex{},
```
lukechampine: initialization not necessary; you can just omit the field

modules/transactionpool/transactionpool.go
```diff
@@ -86,7 +87,7 @@ func New(cs modules.ConsensusSet, g modules.Gateway) (*TransactionPool, error) {
 		// change will then have an index of '0'.
 		consensusChangeIndex: -1,
 
-		mu: sync.New(modules.SafeMutexDelay, 5),
+		mu: demotemutex.DemoteMutex{},
```
lukechampine: ditto

PR #803, opened 2015-10-05
Refactor consensus for unit testing

mtlynch:


mtlynch:
I'm still working on this so hold off on merging (and I know I'm missing documentation in a few places), but I just wanted to check in with you guys to make sure you don't have any objection to the direction I'm going.

I had to add in `GetParentID` and `GetTimestamp` functions on `types.Block`, which I don't like because now callers can get the parent ID by referencing either the `b.ParentID` field or `b.GetParentID()` function. We need the function getter or else we can't create an interface for `types.Block` that does what `ConsensusSet` needs. Should I mark the `ParentID` and `Timestamp` fields deprecated in favor of the getters?

modules/consensus/accept.go
```diff
@@ -24,8 +25,40 @@ var (
...
+		CheckMinerPayouts(height types.BlockHeight) bool
+		CheckTarget(target types.Target) bool
+		GetTimestamp() types.Timestamp
+	}
+
+	// boltTxWrapper wraps a bolt.Tx so that it matches the dbBucket interface.
+	boltTxWrapper struct {
+		tx *bolt.Tx
+	}
+)
```
DavidVorick: This stuff should all be in database.go, definitely check that out before you do more database mocking.

lukechampine:
Why is a block interface needed? A block is a very simple object with no internal state. Its methods are really just for convenience; they could easily be defined as standalone functions.

modules/consensus/accept.go
mtlynch: Can you be more specific? I looked at database.go and I don't see functionality there that does the same thing as I'm doing here (i.e. defining interfaces to wrap `bolt.Tx`).

DavidVorick: Mostly just wanted you to see that all of the functions like 'getSiacoinOutput' are defined in database.go using a more generic 'tx' from boltdb. The interfaces `dbBucket` and `dbTx` should reside in database.go, since they are highly relevant to the database.

I am also a little confused by the `txBlock` interface. `ID` and `CheckMinerPayouts` are both functions operating directly on a block, seems weird that you would mock them. They don't seem like dependencies to me. A block doesn't seem like something you should be mocking.

mtlynch:
> Why is a block interface needed?

It's pretty hard to use `types.Block` for unit testing. I'd like it to be a dumb struct with simple convenience functions, but I don't really think that's what `types.Block` is.

Calling `ID()` for example results in calls to `crypto.HashObject` and `crypto.NewTree`, so bugs in those APIs affect unit tests for `ConsensusSet`, which runs counter to the goal of testing units in isolation.

As another example, one of the unit tests I want to write is that `validateHeader` returns `errBadMinerPayouts` when it's supposed to. Here's my thought process in doing that:
 1. Okay, looking at the implementation of `CheckMinerPayouts` it looks like it depends on the `MinerPayouts` field. That's not so bad because that's another field in `Block` anyway. It's a pain to generate an array of `SiacoinOutput`s but nobody said testing would be easy.
 1. Okay, the sum of the payouts has to match the result of `CalculateSubsidy`. How does that function work?
 1. Oh, that calls into `CalculateCoinbase`...
 1. And then once that callchain is done, we're back in `CalculateSubsidy` and that has to calculate the sum of miner fees on a set of `Transactions`. Ay!

So it's a lot of work just to control the return value of `CheckMinerPayouts`. With an interface, it's dead simple to control the return value.

DavidVorick:
Okay, that all makes sense. You might be able to reduce some mocking complexity by splitting apart `validHeader` and then testing each component separately. For example, checking the miner payouts is going to have more dependencies than everything else, you might not need a single interface that can mock the entirety of `validHeader`. On the other hand it's just a single `if` statement so I'm not sure it really makes sense to split it out.

Another worry is that you have to be careful with mocking structs in the `types` package, because an interface is going to be encoded differently by the `encoding` package, and that's going to break compatibility. But, you might be able to mock the functions inside the block, or change them so they take a block as input instead of being called directly on the block.

Thinking about it more, I definitely think it's a bad idea to make the block (or any of the `types` structs) an interface. If you need to mock a function like `CheckMinerPayouts`, it's probably better to change the the function from `func (b *Block) CheckMinerPayouts() bool` to `func CheckMinerPayouts(b *Block) bool`, which means you can make an interface that has the function as a dependency without needing to turn the block into an interface.

mtlynch:
>Thinking about it more, I definitely think it's a bad idea to make the block (or any of the types structs) an interface. If you need to mock a function like CheckMinerPayouts, it's probably better to change the the function from func (b *Block) CheckMinerPayouts() bool to func CheckMinerPayouts(b *Block) bool, which means you can make an interface that has the function as a dependency without needing to turn the block into an interface.

Okay that works.

mtlynch:
>Thinking about it more, I definitely think it's a bad idea to make the block (or any of the types structs) an interface. If you need to mock a function like CheckMinerPayouts, it's probably better to change the the function from func (b *Block) CheckMinerPayouts() bool to func CheckMinerPayouts(b *Block) bool, which means you can make an interface that has the function as a dependency without needing to turn the block into an interface.

Did this in pull request #808

modules/consensus/clock.go
```diff
@@ -0,0 +1,17 @@
...
+
+type clock interface {
+	Now() types.Timestamp
+}
+
+type stdClock struct{}
+
+func (c stdClock) Now() types.Timestamp {
+	return types.Timestamp(time.Now().Unix())
+}
```
DavidVorick: this should probably all be in timestamp.go, maybe renamed to time.go

mtlynch: done

modules/consensus/database.go
```diff
@@ -78,6 +78,29 @@ var (
...
+	// boltTxWrapper wraps a bolt.Tx so that it matches the dbBucket interface.
+	boltTxWrapper struct {
+		tx *bolt.Tx
+	}
+)
+
+// Bucket returns the dbBucket associated with the given bucket name.
+func (b boltTxWrapper) Bucket(name []byte) dbBucket {
+	return b.tx.Bucket(name)
+}
```
DavidVorick: Can I ask you to split the bucket changes into a separate PR from the clock changes?

I think that changes to how the database stuff works in the consensus package are both needed but also going to be very big. I'm not sure where to proceed with that, but `dbBucket` and `dbTx` seem like a good start.

types/timestamp.go
```diff
@@ -31,3 +31,17 @@ func (ts TimestampSlice) Less(i, j int) bool {
 func (ts TimestampSlice) Swap(i, j int) {
 	ts[i], ts[j] = ts[j], ts[i]
 }
+
+// Clock allows clients to retrieve the current time.
+type Clock interface {
```
DavidVorick: the `Clock` interface does not need to be exported

types/timestamp.go
```diff
@@ -31,3 +31,17 @@ func (ts TimestampSlice) Less(i, j int) bool {
...
+}
+
+// StdClock is an implementation of Clock that retrieves the current time using
+// the system time.
+type StdClock struct{}
+
+// Now retrieves the current timestamp.
+func (c StdClock) Now() Timestamp {
+	return Timestamp(time.Now().Unix())
+}
```
DavidVorick: You make a clock, but you don't use it in any tests.

Also, we should settle on a file layout. @lukechampine @mtlynch , this is my proposal:

```
package xxx

import (
    "stdlib"

    "external packages"

    "internal packages"
)

const (
    // All constants, grouped sensibly
)

var (
    // things that should be consts, but aren't type-compatible

    // errors

    // other stuff, grouped sensibly
)

type (
    // aliases

    // dependency/mocking interfaces + structs

    // types that implement the mocking interfaces (like `StdClock`)
)

// any functions on the mocking types (like `keyDeriver`)

// rest of file
```

You put all of the mocking stuff near the top.

types/timestamp.go
DavidVorick: (but the `Now` function does)

modules/consensus/database.go
```diff
@@ -78,6 +78,29 @@ var (
...
+		Get(key []byte) []byte
+	}
+
+	// dbTx represents a read-only transaction on the database that can be used
+	// for retrieving values.
+	dbTx interface {
+		Bucket(name []byte) dbBucket
+	}
+
+	// boltTxWrapper wraps a bolt.Tx so that it matches the dbBucket interface.
```
DavidVorick: I don't understand the reason for having this when you've got a dbTx interface that could be used.

types/timestamp.go
mtlynch: `modules/consensus/consensusset.go` needs this to be exported so that it can declare a `types.Clock` member. Is there a different way of achieving this?

types/timestamp.go
mtlynch: Yeah, that's a good point. We should get consistent file layouts.

I'll clarify that I don't believe it's useful to think about interfaces in terms of "mocking interfaces" or "non-mocking interfaces" because I don't think there is a distinction. I don't think it's possible for us to define the term "mocking interface" in a way that doesn't just mean the same as "interface." I think it's good design to have types call one another via interfaces rather than static types regardless of whether you create mock implementations of those interfaces for testing.

My suggestion:

```
// everything in your proposal up to the type ( ... ) section

type (
    // aliases

    // interfaces

    // structs
)

// functions of structs

// free functions (functions not tied to a struct)
```

mtlynch: > You make a clock, but you don't use it in any tests.

Can you clarify what you mean? `StdClock` does not have any unit tests because I don't think it's possible to unit test it (open to suggestions if you guys see a way though). It does have test coverage because `ConsensusSet` uses it and those tests hit `StdClock`.

modules/consensus/database.go
mtlynch: I expanded the comment to better explain in a6ca0ba035d8e3bfd12bd533a0f4f37d461835d4 (also `dbBucket` should have been `dbTx`). Let me know if that clarifies.

modules/consensus/database.go
mtlynch: Sure created an independent clock PR: #815

mtlynch: And I created a separate request for database logic: #816.

I'll keep this PR open as is (rather than stripping clock/database stuff out) and then before we merge, I can collapse the commits into a single commit to clean up the change history. This PR (803) is the one that makes use of the smaller PRs to break up all the dependencies in `validHeader` and refactors it so that we can write unit tests for it.

types/timestamp.go
DavidVorick: as long as it's in the same package, you don't need it to be exported. Exporting is for leaving the package, not leaving the file.

DavidVorick:
At least some of this PR has already been merged as it was split into other PRs.

mtlynch:
>At least some of this PR has already been merged as it was split into other PRs.

Correct. I'll rebase so the diffs are clearer.

modules/consensus.go
```diff
@@ -172,11 +172,11 @@ type (
...
 
-		// EarliestChildTimestamp returns the earliest timestamp that is acceptable
-		// on the current longest fork according to the consensus set. This is a
-		// required piece of information for the miner, who could otherwise be at
+		// MinimumValidChildTimestamp returns the earliest timestamp that is
+		// valid on the current longest fork according to the consensus set. This is
+		// a required piece of information for the miner, who could otherwise be at
 		// risk of mining invalid blocks.
-		EarliestChildTimestamp(types.BlockID) (types.Timestamp, bool)
+		MinimumValidChildTimestamp(types.BlockID) (types.Timestamp, bool)
```
DavidVorick: Do you feel strongly about this name change?

modules/consensus/accept.go
```diff
@@ -33,57 +30,29 @@ func (cs *ConsensusSet) validHeader(tx dbTx, b types.Block) error {
...
-	}
-
-	// Check if the block is in the near future, but too far to be acceptable.
-	// This is the last check because it's an expensive check, and not worth
-	// performing if the payouts are incorrect.
-	if b.Timestamp > cs.clock.Now()+types.FutureThreshold {
-		return errFutureTimestamp
-	}
-	return nil
+	return cs.blockValidator.ValidateBlock(b, minTimestamp, parent.ChildTarget, parent.Height+1)
```
DavidVorick: why did you break this into a separate function? It's still only looking at the block header

mtlynch: I did it in an effort to decompose the complexity of `validateHeader` into independent functions. Most of the logic in ConsensusSet.validateHeader now is concerned with gathering the necessary data and passing it off to `blockValidator.ValidateBlock` to check validity rules. Retrieving and deserializing data from boltdb feels very independent from applying validation rules to a Block, so it seems like they should be in separate functions.

Do you feel like the name is incorrect or that the refactoring is incorrect?

modules/consensus.go
mtlynch: Addressed this note in #822 as that's the PR for this change.

modules/consensus/accept.go
DavidVorick: the target check and the timestamp check are definitely part of header validation. The miner payouts check is part of block validation.

One thing we were careful to do in the implementation was to make sure that slow stuff was validated last. I believe there were some other non-obvious things we had to order in a certain way, so definitely be careful that you don't switch up the order. I tried to write tests to catch many of the issues but don't think I was sufficiently comprehensive.

DavidVorick: also, seems odd to me to call `validateBlock` from inside of `validateHeader`. Seems you would want to do it the other way around (from a high level sort of reasoning)

modules/consensus/accept_test.go
```diff
@@ -8,7 +8,79 @@ import (
 	"github.com/NebulousLabs/Sia/types"
 )
 
-// TestIntegrationDoSBlockHandling checks that saved bad blocks are correctly ignored.
+// mockDbBucket is an implementation of dbBucket for unit testing.
+type mockDbBucket struct{}
+
+func (bucket mockDbBucket) Get(key []byte) []byte {
```
DavidVorick: needs a docstring

modules/consensus/accept_test.go
```diff
@@ -8,7 +8,79 @@ import (
...
+	return nil
+}
+
+// mockDbTx is an implementation of dbTx for unit testing. It uses an in-memory
+// key/value store to mock a database.
+type mockDbTx struct {
+	buckets map[string]dbBucket
+}
+
+func (db mockDbTx) Bucket(name []byte) dbBucket {
```
DavidVorick: needs a docstring

modules/consensus/consensusset.go
```diff
@@ -100,12 +100,12 @@ func New(gateway modules.Gateway, persistDir string) (*ConsensusSet, error) {
 
 		dosBlocks: make(map[types.BlockID]struct{}),
 
-		clock: types.StdClock{},
-
 		marshaler: encoding.StdGenericMarshaler{},
 
 		blockRuleHelper: stdBlockRuleHelper{},
 
+		blockValidator: NewBlockValidator(),
```
DavidVorick: all of these fields I feel should be clustered - marshaler, blockRuleHelper, blockValidator

DavidVorick: also, 'RuleHelper' feels like a weak name, but I don't have any good alternate suggestions at the moment.

mtlynch: > all of these fields I feel should be clustered - marshaler, blockRuleHelper, blockValidator

Fixed: a1741d3

> also, 'RuleHelper' feels like a weak name, but I don't have any good alternate suggestions at the moment.

Yeah, I agree. Any time I have to use the name "Helper" I feel like I'm giving a suboptimal name, but I can't think of a better name. It's not a rule *generator*, not a rule *enforcer*, not a rule *validator*. It generates values that are needed for enforcing a rule, but I don't know a succinct way of describing that.

modules/consensus/accept.go
mtlynch: Sorry, I'm still not quite clear. Is the issue that you feel like the refactoring makes sense but the function names need to be revised? Or are you saying that even with clearer function names, the refactoring does not make sense?

The implementation in master is that `validHeader` validates both the block **and** the block header, correct?

Would it make sense to call both functions `validateBlock`? (i.e. `ConsensusSet.validateBlock` and `blockValidator.validateBlock`)?

modules/consensus/accept_test.go
```diff
@@ -8,7 +8,79 @@ import (
...
+
+	tx := mockDbTx{}
+	cs := ConsensusSet{dosBlocks: dosBlocks}
+	err := cs.validateHeader(tx, block)
+	if err != errDoSBlock {
+		t.Fatalf("expected to fail with errDoSBlock, got: %v", err)
+	}
+}
+
+func TestUnitValidateHeaderRejectsOrphanBlocks(t *testing.T) {
```
DavidVorick: needs a docstring

DavidVorick:
just need to add the docstrings and then this looks good.

mtlynch:
Added docstrings, will continue filling in the TODOs in a separate PR

modules/consensus/block_validation_test.go
```diff
@@ -6,6 +6,92 @@ import (
...
+	return nil
+}
+
+type mockClock struct {
+	now types.Timestamp
+}
+
+func (c mockClock) Now() types.Timestamp {
+	return c.now
+}
```
DavidVorick: These 4 functions also need docstrings :P

modules/consensus/block_validation.go
```diff
@@ -2,10 +2,85 @@ package consensus
...
+// stdBlockValidator is the standard implementation of blockValidator.
+type stdBlockValidator struct {
+	// clock is a Clock interface that indicates the current system time.
+	clock types.Clock
+
+	// marshaler encodes and decodes between objects and byte slices.
+	marshaler encoding.GenericMarshaler
+}
+
+func NewBlockValidator() stdBlockValidator {
```
DavidVorick: docstring is also missing here

modules/consensus/block_validation_test.go
```diff
@@ -6,6 +6,92 @@ import (
...
+	}
+	b := types.Block{
+		Timestamp: now + types.ExtremeFutureThreshold + 1,
+	}
+	err := blockValidator.ValidateBlock(b, minTimestamp, types.RootDepth, 0)
+	wantErr := errExtremeFutureTimestamp
+	if err != wantErr {
+		t.Errorf("got %v, want %v", err, wantErr)
+	}
+}
```
lukechampine: use table-driven testing to compress these into one or two functions. e.g.
```go
tests := []struct{
    validator    stdBlockValidator
    block        types.Block
    minTimestamp types.Timestamp
    target       types.Target
    err          error
}{/*...*/}
for i, test := range tests {
    err := test.validator.ValidateBlock(test.block, test.minTimestamp, test.target, 0)
    if err != test.err {
        t.Errorf("case %d: expected %v, got %v", i, test.err, err)
    }
}
```

modules/consensus/accept_test.go
```diff
@@ -8,7 +8,83 @@ import (
...
+
+	tx := mockDbTx{map[string]dbBucket{
+		string(BlockMap): mockDbBucket{},
+	}}
+	cs := ConsensusSet{dosBlocks: dosBlocks}
+	err := cs.validateHeader(tx, block)
+	if err != errOrphan {
+		t.Fatalf("expected to fail with errOrphan, got: %v", err)
+	}
+}
```
lukechampine: see comment below about potentially compressing these tests

modules/consensus/block_validation.go
mtlynch: Fixed in a75127a

modules/consensus/block_validation_test.go
mtlynch: Fixed in a75127a

modules/consensus/accept_test.go
mtlynch: Fixed in 658598c

modules/consensus/block_validation_test.go
mtlynch: Fixed in 40ad2a7

PR #804, opened 2015-10-06
Block api

DavidVorick:
added a call /consensus/block to get a block from the consensus set, height is a parameter.

Ranges are not allowed, mostly because I didn't want people to input a range exceeding 10,000, but maybe it's okay. Or maybe I could spend some more time on it and add the feature with a hardcoded limit to the request size.

Testing + documentation is included.

### DIFF
--- a/api/consensus.go
+++ b/api/consensus.go
@@ -1,19 +1,22 @@
 package api
 
 import (
-	"fmt"
 	"net/http"
 
-	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/types"
 )
 
-// The ConsensusGET struct contains general information about the consensus
-// set, with tags to support idiomatic json encodings.
+// ConsensusGET contains general information about the consensus set, with tags
+// to support idiomatic json encodings.
 type ConsensusGET struct {
 	Height       types.BlockHeight `json:"height"`
 	CurrentBlock types.BlockID     `json:"currentblock"`
 	Target       types.Target      `json:"target"`
+}
+
+// ConsensusBlockGET contains a block.
+type ConsensusBlockGET struct {
+	Block types.Block `json:"block"`
 }
 
 // consensusHandlerGET handles a GET request to /consensus.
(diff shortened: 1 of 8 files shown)


bitspill:
Muxado strikes again

```
=== RUN   TestRandomInboundPeer
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x30 pc=0x6f86db]
goroutine 210 [running]:
github.com/inconshreveable/muxado/proto/frame.(*BasicTransport).Write(0xc82019c700, 0xc8201eef98, 0x10, 0x10, 0xc8201a84a8, 0x0, 0x0)
	<autogenerated>:141 +0x9b

A lot more stuff follows...
```

api/consensus.go
```diff
@@ -1,35 +1,31 @@
...
 type ConsensusGET struct {
 	Height       types.BlockHeight `json:"height"`
 	CurrentBlock types.BlockID     `json:"currentblock"`
 	Target       types.Target      `json:"target"`
 }
 
+// ConsensusBlockGET contains a block.
+type ConsensusBlockGET struct {
+	Block types.Block `json:"block"`
+}
```
lukechampine: this seems a little odd. Why do you need a new type here? Or, if you do want a new type for consistency, why not embed the Block type?

api/consensus.go
```diff
@@ -45,3 +41,29 @@ func (srv *Server) consensusHandler(w http.ResponseWriter, req *http.Request) {
...
 	writeError(w, "unrecognized method when calling /consensus", http.StatusBadRequest)
 }
+
+// consensusBlockHandlerGET handles a GET request to /consensus/block.
+func (srv *Server) consensusBlockHandlerGET(w http.ResponseWriter, req *http.Request) {
+	height, err := scanBlockHeight(req.FormValue("height"))
+	if err != nil {
+		writeError(w, err.Error(), http.StatusBadRequest)
+		return
+	}
```
lukechampine: You should be able to `fmt.Sscan` directly into a `types.BlockHeight`. Since the type is unsigned, `Sscan` will throw an error if you supply a negative value.

api/consensus.go
DavidVorick: Enables us to add more fields if we desire.

DavidVorick:
That makes 3 fails in a row =/

@lukechampine muxado stuff should be fixed/addressed after the renter PR stuff.

DavidVorick:
it's passing now

PR #805, opened 2015-10-08
Repairing

lukechampine:
plus some other minor tweaks.

the `repair` function is tested and seems to do the job, but `threadedRepairFiles` hasn't been tested yet since it's still trying to create real hostUploaders. Some more mocking work will be necessary in order to facilitate testing. Also, `repair` is far too similar to `upload` for my liking. If I gave it some more thought I could probably reduce the overlap quite a bit.

The renter now stores a list of files that it is repairing, in the form of a map from nicknames -> paths on disk. Currently, files are repaired IFF they are uploaded by the user, and you can't disable repairing. The repair loop runs infinitely without any sort of `time.Sleep` delay, and it doesn't remove files even after they reach 100% redundancy (this is because we may add uptime checking later, so that if a host goes offline, those pieces will be repaired). These should probably be changed; just wanted to get some feedback first.

The renter API is going to overhauled soon to make it more RESTful. We'll also want to add an API call to toggle repairing on a file. Note that this call would require a path on disk to the original file (and we have to be careful there, because if the file doesn't actually contain the original data, you'll wind up corrupting the file pretty severely).

### DIFF
--- a/modules/gateway/peers_test.go
+++ b/modules/gateway/peers_test.go
@@ -12,12 +12,25 @@ import (
 	"github.com/NebulousLabs/Sia/modules"
 )
 
+// dummyConn implements the net.Conn interface, but does not carry any actual
+// data. It is passed to muxado, because passing nil results in segfaults.
+type dummyConn struct {
+	net.Conn
+}
+
+// muxado uses these methods when sending its GoAway signal
+func (dc *dummyConn) Write(p []byte) (int, error) { return len(p), nil }
+
+func (dc *dummyConn) Close() error { return nil }
+
+func (dc *dummyConn) SetWriteDeadline(time.Time) error { return nil }
+
 func TestAddPeer(t *testing.T) {
 	g := newTestingGateway("TestAddPeer", t)
 	defer g.Close()
 	id := g.mu.Lock()
 	defer g.mu.Unlock(id)
-	g.addPeer(&peer{addr: "foo", sess: muxado.Client(nil)})
+	g.addPeer(&peer{addr: "foo", sess: muxado.Client(new(dummyConn))})
 	if len(g.peers) != 1 {
 		t.Fatal("gateway did not add peer")
 	}
@@ -33,7 +46,7 @@ func TestRandomInboundPeer(t *testing.T) {
 		t.Fatal("expected errNoPeers, got", err)
 	}
 
-	g.addPeer(&peer{addr: "foo", sess: muxado.Client(nil), inbound: true})
+	g.addPeer(&peer{addr: "foo", sess: muxado.Client(new(dummyConn)), inbound: true})
 	if len(g.peers) != 1 {
 		t.Fatal("gateway did not add peer")
 	}
--- a/modules/renter/download_test.go
+++ b/modules/renter/download_test.go
@@ -4,13 +4,16 @@ import (
 	"bytes"
 	"crypto/rand"
 	"io"
+	"sync"
 	"testing"
 	"time"
 
 	"github.com/NebulousLabs/Sia/crypto"
+	"github.com/NebulousLabs/Sia/modules"
 )
 
 type testHost struct {
+	ip        modules.NetAddress
 	data      []byte
 	pieceMap  map[uint64][]pieceData // key is chunkIndex
 	pieceSize uint64
@@ -20,6 +23,8 @@ type testHost struct {
 	// used to simulate real-world conditions
 	delay    time.Duration // transfers will take this long
 	failRate int           // transfers will randomly fail with probability 1/failRate
+
+	sync.Mutex
 }
 
 func (h *testHost) pieces(chunkIndex uint64) []pieceData {
--- a/modules/renter/files.go
+++ b/modules/renter/files.go
@@ -83,7 +83,7 @@ func (f *file) Available() bool {
 // reaches 100%.
 func (f *file) UploadProgress() float32 {
 	totalBytes := f.pieceSize * uint64(f.erasureCode.NumPieces()) * f.numChunks()
-	return 100 * float32(atomic.LoadUint64(&f.bytesUploaded)) / float32(totalBytes)
+	return 100 * (float32(atomic.LoadUint64(&f.bytesUploaded)) / float32(totalBytes))
 }
 
 // Nickname returns the nickname of the file.
--- a/modules/renter/negotiate.go
+++ b/modules/renter/negotiate.go
@@ -46,8 +46,11 @@ func (hu *hostUploader) Close() error {
 	encoding.WriteObject(hu.conn, types.Transaction{})
 	hu.conn.Close()
 	// submit the most recent revision to the blockchain
-	hu.renter.tpool.AcceptTransactionSet([]types.Transaction{hu.lastTxn})
-	return nil
+	err := hu.renter.tpool.AcceptTransactionSet([]types.Transaction{hu.lastTxn})
+	if err != nil {
+		hu.renter.log.Println("Could not submit final contract revision:", err)
+	}
+	return err
 }
 
 // negotiateContract establishes a connection to a host and negotiates an
--- a/modules/renter/negotiate.go
+++ b/modules/renter/negotiate.go
@@ -52,7 +52,6 @@ func (hu *hostUploader) Close() error {
 
 // negotiateContract establishes a connection to a host and negotiates an
 // initial file contract according to the terms of the host.
-// TODO: better error messages. In many cases we simply return "no data"
 func (hu *hostUploader) negotiateContract(filesize uint64, duration types.BlockHeight) error {
 	conn, err := net.DialTimeout("tcp", string(hu.settings.IPAddress), 15*time.Second)
 	if err != nil {
@@ -77,14 +76,14 @@ func (hu *hostUploader) negotiateContract(filesize uint64, duration types.BlockH
 
 	// write rpcID
 	if err := encoding.WriteObject(conn, modules.RPCUpload); err != nil {
-		return err
+		return errors.New("couldn't initiate RPC: " + err.Error())
 	}
 
 	// read host key
 	// TODO: need to save this?
 	var hostPublicKey types.SiaPublicKey
 	if err := encoding.ReadObject(conn, &hostPublicKey, 256); err != nil {
-		return err
+		return errors.New("couldn't read host's public key: " + err.Error())
 	}
 
 	// create our own key by combining the renter entropy with the host key
@@ -98,7 +97,7 @@ func (hu *hostUploader) negotiateContract(filesize uint64, duration types.BlockH
 
 	// send our public key
 	if err := encoding.WriteObject(conn, ourPublicKey); err != nil {
-		return err
+		return errors.New("couldn't send our public key: " + err.Error())
 	}
 
 	// create unlock conditions
@@ -145,14 +144,14 @@ func (hu *hostUploader) negotiateContract(filesize uint64, duration types.BlockH
 	// send txn
 	if err := encoding.WriteObject(conn, txnSet); err != nil {
 		txnBuilder.Drop()
-		return err
+		return errors.New("couldn't send our proposed contract: " + err.Error())
 	}
 
 	// read back acceptance
 	var response string
 	if err := encoding.ReadObject(conn, &response, 128); err != nil {
 		txnBuilder.Drop()
-		return err
+		return errors.New("couldn't read the host's response to our proposed contract: " + err.Error())
 	}
 	if response != modules.AcceptResponse {
 		txnBuilder.Drop()
@@ -163,7 +162,7 @@ func (hu *hostUploader) negotiateContract(filesize uint64, duration types.BlockH
 	var hostTxnSet []types.Transaction
 	if err := encoding.ReadObject(conn, &hostTxnSet, types.BlockSizeLimit); err != nil {
 		txnBuilder.Drop()
-		return err
+		return errors.New("couldn't read the host's updated contract: " + err.Error())
 	}
 
 	// check that txn is okay. For now, no collateral will be added, so the
@@ -190,14 +189,14 @@ func (hu *hostUploader) negotiateContract(filesize uint64, duration types.BlockH
 	}
 	if err := encoding.WriteObject(conn, signedTxnSet); err != nil {
 		txnBuilder.Drop()
-		return err
+		return errors.New("couldn't send the contract signed by us: " + err.Error())
 	}
 
 	// read signed txn from host
 	var signedHostTxnSet []types.Transaction
 	if err := encoding.ReadObject(conn, &signedHostTxnSet, types.BlockSizeLimit); err != nil {
 		txnBuilder.Drop()
-		return err
+		return errors.New("couldn't read the contract signed by the host: " + err.Error())
 	}
 
 	// submit to blockchain
--- a/modules/renter/negotiate.go
+++ b/modules/renter/negotiate.go
@@ -52,6 +52,7 @@ func (hu *hostUploader) Close() error {
 
 // negotiateContract establishes a connection to a host and negotiates an
 // initial file contract according to the terms of the host.
+// TODO: better error messages. In many cases we simply return "no data"
 func (hu *hostUploader) negotiateContract(filesize uint64, duration types.BlockHeight) error {
 	conn, err := net.DialTimeout("tcp", string(hu.settings.IPAddress), 15*time.Second)
 	if err != nil {
--- a/modules/renter/negotiate.go
+++ b/modules/renter/negotiate.go
@@ -41,6 +41,10 @@ func (hu *hostUploader) fileContract() fileContract {
 	return hu.contract
 }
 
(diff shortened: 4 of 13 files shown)


DavidVorick:
I think at one point the plan was to actually make `repair` and `upload` run the same code. 'upload' would add a file to the repair queue and let the repair loop handle it from there. I don't see why it shouldn't be that way.

Busywaiting on the repair loop seems like a bad idea.

modules/renter/download_test.go
```diff
@@ -20,6 +23,8 @@ type testHost struct {
 	// used to simulate real-world conditions
 	delay    time.Duration // transfers will take this long
 	failRate int           // transfers will randomly fail with probability 1/failRate
+
+	sync.Mutex
```
DavidVorick: makes me pause a bit. If your testing structures are that complex, you might need to add testing for your testing :P

DavidVorick: not that there aren't equally complex testing constructions in other places, wanted to call it out anyway

modules/renter/persist_test.go
```diff
@@ -142,7 +142,9 @@ func TestRenterSaveLoad(t *testing.T) {
 	rt.renter.saveFile(f3)
 
 	// load should now load the files into memory.
+	id := rt.renter.mu.Lock()
 	err = rt.renter.load()
+	rt.renter.mu.Unlock(id)
```
DavidVorick: hmm. In general, I try to avoid using non-exported functions and fields when testing something, unless it's a unit test. I think it helps with the design and overall coverage.

(that's for making changes - for reading the state and verifying it has the expected values, I'll use unexported fields and functions)

modules/renter/repair.go
```diff
@@ -0,0 +1,142 @@
...
+// repair attempts to repair a file by uploading missing pieces to more hosts.
+func (f *file) repair(r io.ReaderAt, hosts []uploader) error {
+	// determine which chunks need to be repaired
+	// TODO: inefficient -- O(2n) on number of pieces
+	present := make([][]bool, f.numChunks())
+	for i := range present {
+		present[i] = make([]bool, f.erasureCode.NumPieces())
+	}
+	for _, fc := range f.contracts {
+		// TODO: ping host, and don't mark pieces if unresponsive?
```
DavidVorick: maybe. Probably better to have some level of uptime history on the host.

DavidVorick: on a side note, I was thinking about it yesterday, and the hostdb+renter should be merged. They are too dependent on eachother.

modules/renter/repair.go
```diff
@@ -0,0 +1,142 @@
...
+	"os"
+	"sync"
+	"sync/atomic"
+)
+
+// repair attempts to repair a file by uploading missing pieces to more hosts.
+func (f *file) repair(r io.ReaderAt, hosts []uploader) error {
+	// determine which chunks need to be repaired
+	// TODO: inefficient -- O(2n) on number of pieces
+	present := make([][]bool, f.numChunks())
```
DavidVorick: vague name. `availablePieces` might be better.

modules/renter/repair.go
```diff
@@ -0,0 +1,142 @@
...
+	for chunkIndex, pieceBools := range present {
+		for pieceIndex, ok := range pieceBools {
+			if !ok {
+				missing[uint64(chunkIndex)] = append(missing[uint64(chunkIndex)], uint64(pieceIndex))
+			}
+		}
+	}
+
+	// For each chunk with missing pieces, re-encode the chunk and upload each
+	// missing piece.
```
DavidVorick: should have a threshold. No need to re-encode everything to fix 1 missing piece. (Well, maybe when we have 12 pieces total, 1 piece is okay. But if we move on to 20+ then there should be some repair threshold)

modules/renter/repair.go
```diff
@@ -0,0 +1,142 @@
...
+			up := uploadPiece{pieces[pieceIndex], chunkIndex, pieceIndex}
+			go func(host uploader, up uploadPiece) {
+				err := host.addPiece(up)
+				if err == nil {
+					atomic.AddUint64(&f.bytesUploaded, uint64(len(up.data)))
+				}
+				wg.Done()
+			}(host, up)
+		}
+		wg.Wait()
```
DavidVorick: I do worry about connection volume, especially since sometimes this for loop will spawn just a few connections, and sometimes it'll spawn a larger number of connections

modules/renter/repair.go
```diff
@@ -0,0 +1,142 @@
...
+
+// threadedRepairUploads improves the health of files tracked by the renter by
+// reuploading their missing pieces. Multiple repair attempts may be necessary
+// before the file reaches full redundancy.
+func (r *Renter) threadedRepairUploads() {
+	for {
+		// make copy of r.repairing under lock
+		repairing := make(map[string]string)
+		id := r.mu.RLock()
+		for name, path := range r.repairing {
```
DavidVorick: `r.repairing` is a confusing name. To me it sounds like a status bool, as opposed to a list of files that are [currently being repaired?] [able to be repaired?] [toggled to be repaired?]

modules/renter/repair.go
```diff
@@ -0,0 +1,142 @@
...
+		for name, path := range repairing {
+			// retrieve file object
+			id = r.mu.RLock()
+			f, ok := r.files[name]
+			r.mu.RUnlock(id)
+			if !ok {
+				r.log.Printf("failed to repair %v: no longer tracking that file", name)
+				id = r.mu.Lock()
+				delete(r.repairing, name)
+				r.mu.Unlock(id)
```
DavidVorick: this seems janky, feels like there's a race condition waiting to happen.

Can't quite put my finger on how this might go wrong, but there's probably a better way.

DavidVorick: Maybe not in the first iteration, but the locks are all spread out, and this function is written with a few assumptions that aren't explicitly commented and those assumptions might get violated in later revisions of this code.

modules/renter/repair.go
```diff
@@ -0,0 +1,142 @@
...
+				continue
+			}
+
+			// build host list
+			var hosts []uploader
+			randHosts := r.hostDB.RandomHosts(f.erasureCode.NumPieces())
+			for i := range randHosts {
+				// TODO: use smarter duration
```
DavidVorick: might be difficult to decide what 'smarter' means -> I think it will largely depend on how people use.

modules/renter/repair.go
```diff
@@ -0,0 +1,142 @@
...
+				hostUploader, err := r.newHostUploader(randHosts[i], f.size, defaultDuration, f.masterKey)
+				if err != nil {
+					continue
+				}
+				defer hostUploader.Close()
+				hosts = append(hosts, hostUploader)
+			}
```
DavidVorick: You 100% need some sort of mechanism that removes hosts which are already storing pieces of a chunk. Might want to be doing host selection on the per-chunk level instead of the per-file level.

Using the random host selector a higher number of times is probably superior anyway.

modules/renter/repair.go
```diff
@@ -0,0 +1,142 @@
...
+
+			// repair
+			err = f.repair(handle, hosts)
+			if err != nil {
+				r.log.Printf("failed to repair %v: %v", name, err)
+				id = r.mu.Lock()
+				delete(r.repairing, name)
+				r.mu.Unlock(id)
```
DavidVorick: why delete the file from the repairing set if the repair fails? Seems overly-aggressive.

modules/renter/repair.go
```diff
@@ -0,0 +1,142 @@
...
+				continue
+			}
+
+			// open file handle
+			handle, err := os.Open(path)
+			if err != nil {
+				r.log.Printf("failed to repair %v: %v", name, err)
+				id = r.mu.Lock()
+				delete(r.repairing, name)
```
DavidVorick: might want to confirm a DNE. If the user is having filesystem troubles or something you probably don't want to be dropping the files from the repair list.

modules/renter/download_test.go
lukechampine: yeah. In this case, this issue is that `repair` (and `upload`, for that matter) call the methods of `uploader` concurrently. I don't see an easy way to get around that, but it should probably be noted in the docstring of the `uploader` interface.

modules/renter/repair.go
lukechampine: that seems sensible, since the renter is the only package using the hostdb anyway. Would make mocking easier as well. My only concern would be bloating of the renter package.

modules/renter/repair.go
lukechampine: idk, encoding a single chunk is pretty cheap, as long as the chunk size remains reasonable (i.e. under 50MB).

modules/renter/repair.go
lukechampine: in this case it maxes out at `min(len(hosts), f.erasureCode.NumPieces())`, which isn't so bad. At some point I think we'll want to run some tests to see if connection volume is causing significant problems. One solution would be to use some form of shared connection pool that caps the total number of connections made by all the modules.

modules/renter/repair.go
lukechampine: yeah, it's not very pretty. Ideally you could just throw a lock around the whole loop, but then `Upload` will block until it can acquire the lock...
I also played around with using a channel instead of a shared map, but that has its own problems (e.g. need to place files back in the channel)

modules/renter/repair.go
DavidVorick: wouldn't surprise me if the renter ends up as our biggest package. It's the most important piece after all.

modules/renter/repair.go
DavidVorick: Right but if every chunk in a 1TB file is missing a single piece (from the same host, as it were) it's not cheap anymore.

modules/renter/repair.go
lukechampine: yeah. One approach would be to try to match the duration of the existing pieces, so that they all expire together. Another would be to extend each piece as necessary, so the file never goes offline.

modules/renter/repair.go
DavidVorick: repair will only handle 1 file at at time? That also seems potentially more parallelizable. Not worried about that performance at this point though, it's fine to keep it at 1 at a time for the short term.

modules/renter/repair.go
lukechampine: `repair` can only return an error if reading from `handle` fails. Seems reasonable to treat this as a fatal error, since it almost certainly means that the file was deleted/moved/corrupted. From the user's perspective, they'll see that the file is no longer being repaired, and they'll have to reenable repairing by pointing it to the right file.

modules/renter/repair.go
lukechampine: on the other hand, if filesystem trouble prevents the file from being opened, the user should probably know about it. Otherwise (depending on the nature of the error) it could just fail to repair indefinitely.

modules/renter/repair.go
lukechampine: this seems like the trickiest part to get right. selecting hosts on a per-chunk basis breaks `repair`, because there's no way for `repair` to check if an `uploader` corresponds to a given host IP (short of using a type assertion). Perhaps we need to move more logic out of `repair` and into this function.

modules/renter/repair.go
DavidVorick: I've generally been in favor of a waterfall method. Pieces expire over the course of 6 weeks, for example. Then someone has up to 6 weeks to turn back on their repair code instead of things suddenly going missing.

DavidVorick: that has $$ overhead though

modules/renter/repair.go
DavidVorick: you can't add an IP method?

modules/renter/repair.go
DavidVorick: a repair loop in the UI could still be stateless, except for not knowing which files needed to be repaired.

But I'd rather have some channel or something that can be used to ferry errors to the user. Perhaps a logger that pipes through stderr or something.

modules/renter/repair.go
lukechampine: ah, good point

modules/gateway/peers_test.go
```diff
@@ -12,12 +12,25 @@ import (
 	"github.com/NebulousLabs/Sia/modules"
 )
 
+// dummyConn implements the net.Conn interface, but does not carry any actual
+// data. It is passed to muxado, because passing nil results in segfaults.
+type dummyConn struct {
```
DavidVorick: prefer the name mockConn or genericConn I think. Not a huge deal but we should probably settle on some naming conventions for mocking things.

mockConn -> a Conn with absolute minimum viable functionality, minimal statefullness, functions that just  return non-zero values and always err = nil.

mockSomethingConn -> a Conn that does Something. Maybe it has a stateful network, or can validate whether a signature was signed correctly by the calling function (by checking for a leading magic byte or something simple). mockVerifySigner for example.

modules/renter/files.go
```diff
@@ -83,7 +83,7 @@ func (f *file) Available() bool {
 // reaches 100%.
 func (f *file) UploadProgress() float32 {
 	totalBytes := f.pieceSize * uint64(f.erasureCode.NumPieces()) * f.numChunks()
-	return 100 * float32(atomic.LoadUint64(&f.bytesUploaded)) / float32(totalBytes)
+	return 100 * (float32(atomic.LoadUint64(&f.bytesUploaded)) / float32(totalBytes))
```
DavidVorick: are the parens required? Maybe for precision?

modules/renter/negotiate.go
```diff
@@ -189,14 +196,14 @@ func (hu *hostUploader) negotiateContract(filesize uint64, duration types.BlockH
...
-		return err
+		return errors.New("couldn't send the contract signed by us: " + err.Error())
 	}
 
 	// read signed txn from host
 	var signedHostTxnSet []types.Transaction
 	if err := encoding.ReadObject(conn, &signedHostTxnSet, types.BlockSizeLimit); err != nil {
 		txnBuilder.Drop()
-		return err
+		return errors.New("couldn't read the contract signed by the host: " + err.Error())
```
DavidVorick: that is useful.

modules/renter/files.go
lukechampine: yeah this was the 99.999% bug I ran into. Without the parens, 100*x/x != 100 for certain x.

modules/gateway/peers_test.go
lukechampine: A mock is a specific type of test object that has programmable expectations. This is more of a "stub" than a mock, at least going by the definitions in http://martinfowler.com/articles/mocksArentStubs.html. The "dummy-" prefix is just my go-to for objects that are the bare minimum.

modules/renter/repair.go
```diff
@@ -0,0 +1,198 @@
...
+		for _, h := range hosts {
+			for _, ip := range curHosts {
+				if ip == h.addr() {
+					continue outer
+				}
+			}
+			chunkHosts = append(chunkHosts, h)
+		}
+		// don't bother encoding if there aren't any hosts to upload to
+		if len(chunkHosts) == 0 {
```
DavidVorick: confusing variable name - has different meaning than the function `chunkHosts`

modules/renter/repair.go
```diff
@@ -0,0 +1,198 @@
...
+		}
+
+		// read chunk data
+		chunk := make([]byte, f.chunkSize())
+		_, err := r.ReadAt(chunk, int64(chunkIndex*f.chunkSize()))
+		if err != nil && err != io.EOF && err != io.ErrUnexpectedEOF {
+			return err
+		}
+
+		// encode
```
DavidVorick: comment is not needed, the function call you are making is `Encode`

modules/renter/repair.go
```diff
@@ -0,0 +1,198 @@
...
+				// TODO: use smarter duration
+				hostUploader, err := r.newHostUploader(randHosts[i], f.size, defaultDuration, f.masterKey)
+				if err != nil {
+					continue
+				}
+				defer hostUploader.Close()
+				hosts = append(hosts, hostUploader)
+			}
+
+			// repair
```
DavidVorick: same thing about the non-needed comment.

modules/renter/upload_test.go
```diff
@@ -7,12 +7,16 @@ import (
 	"time"
 
 	"github.com/NebulousLabs/Sia/crypto"
+	"github.com/NebulousLabs/Sia/modules"
 )
 
 func (h *testHost) addPiece(p uploadPiece) error {
```
DavidVorick: need docstring

modules/renter/upload_test.go
```diff
@@ -27,7 +31,18 @@ func (h *testHost) addPiece(p uploadPiece) error {
 	return nil
 }
 
-func (h *testHost) fileContract() fileContract { return fileContract{} }
+func (h *testHost) fileContract() fileContract {
```
DavidVorick: need docstring

PR #806, opened 2015-10-08
Add bitspill's node to bootstrap list

DavidVorick:
(with permission)


bitspill:
Confirming permission.

PR #807, opened 2015-10-08
Sia-Agent is only whitelisted agent

DavidVorick:
All the others are obsolete.

Turns out that the miner was never setting a user agent in the first place, there must have been some sort of bug that was allowing a blank useragent to go through.

Either that, or a blank user agent used to be allowed (which I would believe).

I've updated the miner code, which means we'll need a miner 1.0.3

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -39,14 +39,11 @@ func HttpPOST(url string, data string) (resp *http.Response, err error) {
 // the API.
 func handleHTTPRequest(mux *http.ServeMux, url string, handler http.HandlerFunc) {
 	mux.HandleFunc(url, func(w http.ResponseWriter, req *http.Request) {
-		permittedClients := []string{"Sia-Agent", "Sia-Miner", "Electron", "AtomShell"}
-		for _, pc := range permittedClients {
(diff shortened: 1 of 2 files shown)


api/api.go
```diff
@@ -39,14 +39,11 @@ func HttpPOST(url string, data string) (resp *http.Response, err error) {
 // the API.
 func handleHTTPRequest(mux *http.ServeMux, url string, handler http.HandlerFunc) {
 	mux.HandleFunc(url, func(w http.ResponseWriter, req *http.Request) {
-		permittedClients := []string{"Sia-Agent", "Sia-Miner", "Electron", "AtomShell"}
-		for _, pc := range permittedClients {
-			if strings.Contains(req.UserAgent(), pc) {
-				handler(w, req)
-				return
-			}
+		if !strings.Contains(req.UserAgent(), "Sia-Agent") {
```
lukechampine: could even be a simple `==` if you want maximum strictness. But I suppose this allows you to do things like `Sia-Agent-Miner` if you want.

DavidVorick: Was thinking `Sia-Agent - Sia-UI v0.4.3`

PR #808, opened 2015-10-09
Move block check functions to modules/consensus

mtlynch:
Move block checking functions to modules/consensus as part of an effort
to pull complexity out of types.Block and make code that depends on
types.Block more testable.


DavidVorick:
@mtlynch we fixed the muxado errors that caused the build to frequently fail. If you merge the latest changes it won't be an issue for your PR anymore

lukechampine:
muxado fix is in the repairing branch, so that needs to be merged first

DavidVorick:
Pretty simple change, looks good to me without any further changes.

mtlynch:
Passing now.

I changed it slightly because miner/blockmanager.go now depends on `CalculateSubsidy` so I've left that function where it is for now.

PR #809, opened 2015-10-09
Some architecture changes

DavidVorick:
Miner and wallet moved to using stdlib sync.

Transaction pool no longer auto-subscribes you to the consensus set.

Dependencies are sorted by how they are used.

CPU hashrate algorithm simplified.

TestMiner split into its own interface, tester objects all now use a test miner instead of a full miner.

Some testing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,23 +5,27 @@ all: install
 # dependencies installs all of the dependencies that are required for building
 # Sia.
 dependencies:
-	go install -race std
+	# Consensus Dependencies
 	go get -u github.com/NebulousLabs/demotemutex
 	go get -u github.com/NebulousLabs/ed25519
-	go get -u github.com/NebulousLabs/entropy-mnemonics
 	go get -u github.com/NebulousLabs/go-upnp
 	go get -u github.com/NebulousLabs/merkletree
-	go get -u github.com/bgentry/speakeasy
+	go get -u github.com/inconshreveable/muxado
 	go get -u github.com/boltdb/bolt
 	go get -u github.com/dchest/blake2b
+	go get -u golang.org/x/crypto/twofish
+	# Module + Daemon Dependencies
+	go get -u github.com/NebulousLabs/entropy-mnemonics
 	go get -u github.com/inconshreveable/go-update
-	go get -u github.com/inconshreveable/muxado
 	go get -u github.com/kardianos/osext
 	go get -u github.com/klauspost/reedsolomon
-	go get -u github.com/laher/goxc
-	go get -u github.com/spf13/cobra
 	go get -u github.com/stretchr/graceful
-	go get -u golang.org/x/crypto/twofish
+	# Frontend Dependencies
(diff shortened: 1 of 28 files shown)


Makefile
```diff
@@ -5,23 +5,27 @@ all: install
...
 dependencies:
-	go install -race std
+	# Consensus Dependencies
 	go get -u github.com/NebulousLabs/demotemutex
 	go get -u github.com/NebulousLabs/ed25519
-	go get -u github.com/NebulousLabs/entropy-mnemonics
 	go get -u github.com/NebulousLabs/go-upnp
 	go get -u github.com/NebulousLabs/merkletree
-	go get -u github.com/bgentry/speakeasy
+	go get -u github.com/inconshreveable/muxado
```
lukechampine: muxado and upnp are module dependencies

DavidVorick:
ok

modules/miner.go
```diff
@@ -66,8 +70,4 @@ type CPUMiner interface {
 type Miner interface {
 	BlockManager
 	CPUMiner
```
lukechampine: I'm confused by the various `Miner` types now. If we have a `TestMiner`, why is `Miner` also a `CPUMiner`? Isn't the `TestMiner` a CPU Miner? Is CPU mining even allowed/possible anymore? If not, shouldn't the `Miner` just be a `BlockManager`?

DavidVorick: Still in transition. Eventually, the server will be reconfigured so that it contains both a block manager and a cpu miner. But that will require making changes to 3-4 files within the api package. Seemed out of scope for this PR, I'm trying to keep things small.

The Miner interface is on the way out, just doing it in steps.

modules/miner/testminer.go
```diff
@@ -0,0 +1,96 @@
...
+	copy(header[48:], merkleRoot[:])
+
+	nonce := (*uint64)(unsafe.Pointer(&header[32]))
+	for i := 0; i < iterationsPerAttempt; i++ {
+		id := crypto.HashBytes(header)
+		if bytes.Compare(target[:], id[:]) >= 0 {
+			copy(b.Nonce[:], header[32:40])
+			return b, true
+		}
+		*nonce++
```
lukechampine: this probably makes little difference, but I found that you can speed up this increment nearly 3x in a simple benchmark:
```go
// Benchmark says: 2.40 ns/op
var header [80]byte
nonce := (*uint64)(unsafe.Pointer(&header[32]))
for i := 0; i < b.N; i++ {
    *nonce++ // note that this is equivalent to *nonce = *nonce + 1
}

// Benchmark says: 0.77 ns/op
var header [80]byte
nonce := (*uint64)(unsafe.Pointer(&header[32]))
var uintNonce uint64 // local variable that matches values of nonce
for i := 0; i < b.N; i++ {
    uintNonce++
    *nonce = uintNonce // only one deref instead of two
}
```
For comparison, just incrementing a local uint64 takes 0.41 ns/op.

PR #810, opened 2015-10-12
Miner

DavidVorick:
This refactors the actual piece that gives headers to the mining clients. I changed a few things and added a bunch of tests. There were a few bugs I found that could have been the source of the high stale rate.

It's hard to be certain that the stale rate bug has been fixed, becuase it's hard to set up a dev environment with multiple competing GPUs in a realistic latency-environment. But the testing is a lot more complete, and I imagine the stalerate will be much closer to where it should be.

### DIFF
--- a/modules/consensus.go
+++ b/modules/consensus.go
@@ -38,8 +38,9 @@ type (
 
 	// ConsensusSetDigestSubscriber receives digests about changes to the
 	// consensus set in the form of a list of ids of blocks that got reverted
-	// and a list of ids of blocks that got applied. The ordering is always all
-	// of the reverted blocks followed by all of the applied blocks.
+	// and a list of ids of blocks that got applied. The IDs are given in the
+	// order in which they were processed. Reverted blocks are always processed
+	// before applied blocks.
 	ConsensusSetDigestSubscriber interface {
 		// ProcessConsensusDigest sends a list of reverted blocks and applied
 		// blocks to the module. There may not be any reverted blocks, but
@@ -168,7 +169,7 @@ type (
 		// ConsensusSetDigestSubscribe subscribes a module to a digest of the
 		// changes in the consensus set. Immediately upon subscription, a
 		// digest containing the list of blocks from the starting id to the
-		// current block.
+		// current block is presented to the subscriber.
 		ConsensusSetDigestSubscribe(ConsensusSetDigestSubscriber)
 
 		// EarliestChildTimestamp returns the earliest timestamp that is acceptable
--- a/modules/consensus.go
+++ b/modules/consensus.go
@@ -33,6 +33,16 @@ type (
 		ProcessConsensusChange(ConsensusChange)
 	}
 
+	// ConsensusSetDigestSubscriber receives digests about changes to the
+	// consensus set in the form of a list of ids of blocks that got reverted
+	// and a list of ids of blocks that got applied. The ordering is always all
+	// of the reverted blocks followed by all of the applied blocks.
+	ConsensusSetDigestSubscriber interface {
+		// ProcessConsensusDigest sends a list of reverted blocks and applied
+		// blocks to the module.
+		ProcessConsensusDigest(revertedBlocks []types.BlockID, appliedBlocks []types.BlockID)
+	}
+
 	// A ConsensusChange enumerates a set of changes that occured to the consensus set.
 	ConsensusChange struct {
 		// RevertedBlocks is the list of blocks that were reverted by the change.
@@ -150,6 +160,11 @@ type (
 		// will be sent to the module via the 'ReceiveConsensusSetUpdate' function.
 		// This is a thread-safe way of managing updates.
 		ConsensusSetSubscribe(ConsensusSetSubscriber)
+
+		// ConsensusSetDigestSubscribe subscribes a module to a digest of the
+		// changes in the consensus set. Immediately upon subscription, a
+		// digest containing the current block will be provided.
+		ConsensusSetDigestSubscribe(ConsensusSetDigestSubscriber)
 
 		// EarliestChildTimestamp returns the earliest timestamp that is acceptable
 		// on the current longest fork according to the consensus set. This is a
--- a/modules/consensus.go
+++ b/modules/consensus.go
@@ -30,6 +30,8 @@ type (
 	ConsensusSetSubscriber interface {
 		// ProcessConsensusChange sends a consensus update to a module through
 		// a function call. Updates will always be sent in the correct order.
+		// There may not be any reverted blocks, but there will always be
+		// applied blocks.
 		ProcessConsensusChange(ConsensusChange)
 	}
 
@@ -39,7 +41,8 @@ type (
 	// of the reverted blocks followed by all of the applied blocks.
 	ConsensusSetDigestSubscriber interface {
 		// ProcessConsensusDigest sends a list of reverted blocks and applied
-		// blocks to the module.
+		// blocks to the module. There may not be any reverted blocks, but
+		// there will always be applied blocks.
 		ProcessConsensusDigest(revertedBlocks []types.BlockID, appliedBlocks []types.BlockID)
 	}
 
@@ -163,7 +166,8 @@ type (
 
 		// ConsensusSetDigestSubscribe subscribes a module to a digest of the
 		// changes in the consensus set. Immediately upon subscription, a
-		// digest containing the current block will be provided.
+		// digest containing the list of blocks from the starting id to the
+		// current block.
 		ConsensusSetDigestSubscribe(ConsensusSetDigestSubscriber)
 
 		// EarliestChildTimestamp returns the earliest timestamp that is acceptable
--- a/modules/consensus.go
+++ b/modules/consensus.go
@@ -17,6 +17,7 @@ const (
 
 var (
 	ErrBlockKnown        = errors.New("block already present in database")
+	ErrBlockUnsolved     = errors.New("block does not meet target")
 	ErrNonExtendingBlock = errors.New("block does not extend the longest fork")
 )
 
--- a/modules/miner/blockmanager.go
+++ b/modules/miner/blockmanager.go
@@ -5,9 +5,14 @@ import (
 	"errors"
 	"time"
 
+	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/types"
+)
+
+var (
+	errLateHeader = errors.New("header is old, block could not be recovered")
 )
 
 // blockForWork returns a block that is ready for nonce grinding, including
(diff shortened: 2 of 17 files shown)


modules/consensus.go
```diff
@@ -30,9 +30,22 @@ type (
...
 		// a function call. Updates will always be sent in the correct order.
+		// There may not be any reverted blocks, but there will always be
+		// applied blocks.
 		ProcessConsensusChange(ConsensusChange)
 	}
 
+	// ConsensusSetDigestSubscriber receives digests about changes to the
+	// consensus set in the form of a list of ids of blocks that got reverted
+	// and a list of ids of blocks that got applied. The ordering is always all
+	// of the reverted blocks followed by all of the applied blocks.
```
lukechampine: what is meant by "ordering" here? Aren't the reverted/applied blocks two separate arguments?

modules/consensus.go
```diff
@@ -151,6 +164,12 @@ type (
 		// This is a thread-safe way of managing updates.
 		ConsensusSetSubscribe(ConsensusSetSubscriber)
 
+		// ConsensusSetDigestSubscribe subscribes a module to a digest of the
+		// changes in the consensus set. Immediately upon subscription, a
+		// digest containing the list of blocks from the starting id to the
+		// current block.
```
lukechampine: sentence fragment

modules/consensus.go
```diff
@@ -30,9 +30,22 @@ type (
...
 
+	// ConsensusSetDigestSubscriber receives digests about changes to the
+	// consensus set in the form of a list of ids of blocks that got reverted
+	// and a list of ids of blocks that got applied. The ordering is always all
+	// of the reverted blocks followed by all of the applied blocks.
+	ConsensusSetDigestSubscriber interface {
+		// ProcessConsensusDigest sends a list of reverted blocks and applied
+		// blocks to the module. There may not be any reverted blocks, but
+		// there will always be applied blocks.
+		ProcessConsensusDigest(revertedBlocks []types.BlockID, appliedBlocks []types.BlockID)
```
lukechampine: this could simulated by `ProcessConsensusChange`, right? So is this just for efficiency reasons?

modules/miner/blockmanager.go
```diff
@@ -1,127 +1,115 @@
...
-	// The transactions are assembled by calling append(singleElem,
-	// existingSlic) because doing it the reverse way has some side effects,
-	// creating a race condition and ultimately changing the block hash for
-	// other parts of the program. This is related to the fact that slices are
-	// pointers, and not immutable objects. Use of the builtin `copy` function
-	// when passing objects like blocks around may fix this problem.
+
+	// Update the address + payouts.
+	_ = m.checkAddress() // Err is ignored - address generation failed but can't do anything about it (log maybe).
+	b.MinerPayouts = []types.SiacoinOutput{types.SiacoinOutput{Value: b.CalculateSubsidy(m.height + 1), UnlockHash: m.address}}
```
lukechampine: I believe you can omit the inner type here (`types.SiacoinOutput`)

modules/miner/blockmanager.go
```diff
@@ -1,127 +1,115 @@
...
-	} else {
-		// Set block to previous block, but create new arbData
-		block = m.blockMem[m.headerMem[m.memProgress-1]]
-		arbData, _ = crypto.RandBytes(types.SpecifierLen)
-		block.Transactions[0].ArbitraryData[0] = arbData
-		header = block.Header()
+	// Create a header from the source block - this may be a race condition,
+	// but I don't think so (underlying slice may be shared with other blocks
+	// accessible outside the miner).
+	var arbData [crypto.EntropySize]byte
```
lukechampine: seems a bit odd to use a byte array instead of a slice

modules/miner/blockmanager.go
```diff
@@ -155,21 +143,39 @@ func (m *Miner) SubmitBlock(b types.Block) error {
...
+	arbData, arbExists := m.arbDataMem[bh]
 	if !bExists || !arbExists {
-		err := errors.New("block header returned late - block was cleared from memory")
-		m.log.Println("ERROR:", err)
-		return err
+		m.log.Println("ERROR:", errLateHeader)
+		m.mu.Unlock()
+		return errLateHeader
+	} else {
+	}
```
lukechampine: ?

modules/miner/blockmanager_test.go
```diff
@@ -1,115 +1,227 @@
...
-	if !exists {
-		return nil, errors.New("Header is either invalid or too old")
+// solveHeader takes a block header as input and returns a solved block header
+// as output.
+func solveHeader(header types.BlockHeader, target types.Target) types.BlockHeader {
+	// Solve the header.
+	for {
+		// Incrememnt the nonce first to guarantee that a new header is formed
+		// - this helps check for pointer errors.
+		header.Nonce[0]++
```
lukechampine: this only increments the first byte, yes? Will it loop infinitely if 256 attempts are insufficient?

modules/consensus.go
DavidVorick: they are two separate arguments, but they form a single list.

rev1, rev2, app1, app2, app3

for example.

modules/miner/blockmanager_test.go
```diff
@@ -1,115 +1,227 @@
...
+		t.Fatal(err)
+	}
 
-	for i := 0; i < headerForWorkMemory; i++ {
-		headers[i], _, err = mt.miner.HeaderForWork()
+	// Create a suite of headers for imaginary parallel mining.
+	headerDepth := HeaderMemory / BlockMemory * 2
+	var solvedHeaders []types.BlockHeader
+	for i := 0; i < headerDepth; i++ {
+		header, target, err := mt.miner.HeaderForWork()
```
lukechampine: it looks like you know the length of the slice in advance here, so you could write:
```go
solvedHeader := make([]types.BlockHeader, headerDepth)
for i := range solvedHeaders {
    header, target, err := mt.miner.HeaderForWork()
    if err != nil {
        t.Fatal(err)
    }
    solvedHeaders[i] = solveHeader(header, target)
}
```

modules/consensus.go
DavidVorick: yes. loading block ids from disk + decoding them is cheaper than loading + decoding entire processed blocks.

Eventually, `ProcessConsensusChange` will actually be deprecated.

modules/miner/blockmanager.go
DavidVorick: lol fixed before I even read the comment.

It contained a debug statement

modules/miner/blockmanager_test.go
DavidVorick: each attempt has 50% pass rate

256 attempts: 1-in-2^128 chance of failure.

Cryptographic grade probability that this test will never fail

modules/miner/blockmanager_test.go
```diff
@@ -1,115 +1,227 @@
...
 	}
 
-	// Make sure Miner still has headerForWorkMemory headers stored
-	for i := 0; i < headerForWorkMemory; i++ {
-		_, exists := mt.miner.blockMem[headers[i]]
-		if !exists {
-			t.Error("Miner did not remember enough headers")
+	// Submit the headers randomly and make sure they are all considered valid.
+	for len(solvedHeaders) > 0 {
+		choice, err := crypto.RandIntn(len(solvedHeaders))
```
lukechampine: `crypto.Perm` might be useful here

modules/miner/miner.go
```diff
@@ -6,32 +6,27 @@ import (
...
+	HeaderMemory = 10000
+
+	// BlockMemory is the maximum number of blocks the miner will store
+	// Blocks take up to 2 megabytes of memory, which is why this number is
+	// limited.
+	BlockMemory = 50
+
+	// MaxSourceBlockAge is the maximum amount of time that is allowed to
+	// elapse between generating source blocks.
+	MaxSourceBlockAge = 60 * time.Second
```
lukechampine: is there a reason for exporting these?

modules/miner/blockmanager_test.go
lukechampine: gotcha. I didn't realize the pass rate was 50%

modules/consensus.go
lukechampine: That seems pretty heavily implied to me. I think the sentence actually makes things more confusing. Maybe something like: `The IDs are given in the order in which they were processed by the consensus set. Reverted blocks are always processed before applied blocks.`

modules/consensus.go
lukechampine: That sounds good to me, unless it would result in decoding a block multiple times

DavidVorick: Not any more than is already happening, at least not at startup. Live, a block may be decoded 5 or more times as opposed to just once, but we're talking about 30 milliseconds, which is not a problem since validation takes much longer than that anyway.

I'm focusing down the biggest bottlenecks for now.

modules/miner/blockmanager.go
DavidVorick: indeed

modules/miner/blockmanager.go
DavidVorick: pointers =/

The arb data gets manipulated frequently. If you use a slice, 'pass-by-value' means the underlying pointer is still pointed at the same data, which was causing incorrect block reassembly.

modules/miner/blockmanager_test.go
DavidVorick: That's a bad name for a function. Had no idea what it did without the docstring =/

Hope you don't mind if I change it.

modules/miner/miner.go
DavidVorick: yes

DavidVorick: lol. The reason is so that they make it into the godoc. When optimizing a miner, or something like a mining pool, knowing those values will be helpful for figuring out how many cards you can hook up to a single instance of siad.

So right now they aren't used anywhere, but their documentation is important because they directly impact users of the block manager.

modules/miner/blockmanager_test.go
DavidVorick: oh nvm the stdlib has a function called `Perm`. Huh. Guess I'll keep it.

DavidVorick:
ok, this is good to go

modules/miner/miner.go
lukechampine: hmm. Seems like a strange reason to me (why would an optimizer look at godoc? If they were serious enough to be optimizing specifically for this code, wouldn't they look at the source?) but whatevs

modules/miner/blockmanager_test.go
```diff
@@ -1,115 +1,222 @@
...
+	mt, err := createMinerTester("TestIntegrationManyHeaders")
+	if err != nil {
+		t.Fatal(err)
+	}
 
-	for i := 0; i < headerForWorkMemory; i++ {
-		headers[i], _, err = mt.miner.HeaderForWork()
+	// Create a suite of headers for imaginary parallel mining.
+	solvedHeaders := make([]types.BlockHeader, 0, HeaderMemory/BlockMemory*2)
+	for _ = range solvedHeaders {
```
lukechampine: this is still weird lol. why don't you want to index `solvedHeaders`?

DavidVorick: index is not used. append seems natural enough to me.

DavidVorick: oh, I guess you can strip another line out

PR #811, opened 2015-10-14
Renewing

lukechampine:
Files are now renewed before their contracts are set to expire. An offline check was implemented, but not used, since I think it'll be tricky to get right due to latency, and at this stage it may be overly aggressive anyway.

I also removed the `bytesUploaded` and `chunksUploaded` fields from the file object, since these can be calculated from the contract set. Unfortunately this meant adding a mutex, since the contracts are updated during the upload process. But then I realized I needed to add a mutex anyway, since the repair loop can now modify contracts in the background at any time. I worry about how much complexity this has introduced though. The renter does not feel very solid to me, and I take responsibility for that. A lot of the difficulty is tied to persistence; in-memory data structures have to be kept in sync with on-disk files. I hope that we can steadily improve the renter until it feels sturdy again.

This stuff hasn't been adequately tested. I'll be doing that tomorrow. I will also be enabling "lazy" contract formation, so that you can request as many hosts from the hostdb as you need, and you'll only form contracts with them once you actually start uploading data. The current upload algorithm doesn't ask for enough hosts, so I think this will result in fewer upload failures.

### DIFF
--- a/api/daemon_test.go
+++ b/api/daemon_test.go
@@ -57,7 +57,7 @@ func TestSignedUpdate(t *testing.T) {
 	}
 
 	// apply (bad signature)
-	resp, err := HttpGET("http://localhost" + st.server.apiServer.Addr + "/daemon/updates/apply?version=current")
+	resp, err := HttpGET("http://" + st.server.listener.Addr().String() + "/daemon/updates/apply?version=current")
 	if err != nil {
 		t.Fatal("GET failed:", err)
 	}
--- a/api/daemon_test.go
+++ b/api/daemon_test.go
@@ -51,8 +51,10 @@ func TestSignedUpdate(t *testing.T) {
 
 	// newer version
 	uh.version = "100.0"
-	st.getAPI("/daemon/updates/check", &info)
-	if !info.Available {
+	err = st.getAPI("/daemon/updates/check", &info)
+	if err != nil {
+		t.Error(err)
+	} else if !info.Available {
 		t.Error("new version should be available")
 	}
 
--- a/encoding/prefix.go
+++ b/encoding/prefix.go
@@ -1,14 +1,8 @@
 package encoding
 
 import (
-	"errors"
 	"fmt"
 	"io"
-)
-
-var (
-	errNoData    = errors.New("no data")
-	errBadPrefix = errors.New("could not read full length prefix")
 )
 
 // ReadPrefix reads an 8-byte length prefixes, followed by the number of bytes
@@ -16,13 +10,8 @@ var (
 // specified maximum length.
 func ReadPrefix(r io.Reader, maxLen uint64) ([]byte, error) {
 	prefix := make([]byte, 8)
-	if n, err := io.ReadFull(r, prefix); n == 0 {
-		if err == nil {
-			err = errNoData
-		}
+	if _, err := io.ReadFull(r, prefix); err != nil {
 		return nil, err
-	} else if err != nil {
-		return nil, errBadPrefix
 	}
 	dataLen := DecUint64(prefix)
 	if dataLen > maxLen {
--- a/encoding/prefix.go
+++ b/encoding/prefix.go
@@ -17,7 +17,10 @@ var (
 func ReadPrefix(r io.Reader, maxLen uint64) ([]byte, error) {
 	prefix := make([]byte, 8)
 	if n, err := io.ReadFull(r, prefix); n == 0 {
-		return nil, errNoData
+		if err == nil {
+			err = errNoData
+		}
+		return nil, err
 	} else if err != nil {
 		return nil, errBadPrefix
 	}
--- a/modules/renter/files.go
+++ b/modules/renter/files.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"os"
 	"path/filepath"
+	"sync"
 
 	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/modules"
@@ -28,6 +29,7 @@ type file struct {
 	erasureCode modules.ErasureCoder
 	pieceSize   uint64
 	mode        uint32 // actually an os.FileMode
+	mu          sync.RWMutex
 }
 
 // A fileContract is a contract covering an arbitrary number of file pieces.
@@ -69,6 +71,8 @@ func (f *file) numChunks() uint64 {
 
 // Available indicates whether the file is ready to be downloaded.
 func (f *file) Available() bool {
+	f.mu.RLock()
+	defer f.mu.RUnlock()
 	chunkPieces := make([]int, f.numChunks())
 	for _, fc := range f.contracts {
 		for _, p := range fc.Pieces {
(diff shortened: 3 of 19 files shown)


DavidVorick:
nil pointer dereference

modules/renter/files.go
```diff
@@ -75,15 +71,35 @@ func (f *file) numChunks() uint64 {
...
+		for _, p := range fc.Pieces {
+			chunkPieces[p.Chunk]++
+		}
+	}
+	for _, n := range chunkPieces {
+		if n < f.erasureCode.MinPieces() {
+			return false
+		}
+	}
+	return true
```
DavidVorick: :+1:

api/daemon_test.go
```diff
@@ -51,13 +51,15 @@ func TestSignedUpdate(t *testing.T) {
...
-	st.getAPI("/daemon/updates/check", &info)
-	if !info.Available {
+	err = st.getAPI("/daemon/updates/check", &info)
+	if err != nil {
+		t.Error(err)
+	} else if !info.Available {
 		t.Error("new version should be available")
 	}
 
 	// apply (bad signature)
```
DavidVorick: what does the comment mean?

encoding/prefix.go
```diff
@@ -17,7 +17,10 @@ var (
 func ReadPrefix(r io.Reader, maxLen uint64) ([]byte, error) {
 	prefix := make([]byte, 8)
 	if n, err := io.ReadFull(r, prefix); n == 0 {
-		return nil, errNoData
+		if err == nil {
+			err = errNoData
+		}
+		return nil, err
```
DavidVorick: what prompted this change?

DavidVorick: oh, sometimes the error is different than nodata I guess

modules/renter/negotiate.go
```diff
@@ -281,6 +286,9 @@ func (hu *hostUploader) addPiece(p uploadPiece) error {
 }
 
 func (hu *hostUploader) revise(fc types.FileContract, piece []byte, height types.BlockHeight) error {
```
DavidVorick: docstring

modules/renter/negotiate.go
```diff
@@ -374,6 +383,11 @@ func (hu *hostUploader) revise(fc types.FileContract, piece []byte, height types
 }
 
 func (r *Renter) newHostUploader(settings modules.HostSettings, filesize uint64, duration types.BlockHeight, masterKey crypto.TwofishKey) (*hostUploader, error) {
```
DavidVorick: docstring

modules/renter/repair.go
```diff
@@ -2,50 +2,117 @@ package renter
...
 	"github.com/NebulousLabs/Sia/modules"
+	"github.com/NebulousLabs/Sia/types"
 )
 
+const (
+	// When a file contract is within this many blocks of expiring, the renter
+	// will attempt to reupload the data covered by the contract.
+	renewThreshold = 2000
+
+	hostTimeout = 3 * time.Second // TODO: longer?
```
DavidVorick: at least 5, but honestly I'd prefer 10 or 15

modules/renter/repair.go
```diff
@@ -87,22 +157,23 @@ func (f *file) repair(r io.ReaderAt, pieceMap map[uint64][]uint64, hosts []uploa
...
-				if err == nil {
-					atomic.AddUint64(&f.bytesUploaded, uint64(len(up.data)))
-				}
-				wg.Done()
-			}(host, up)
+			//go func(host uploader, up uploadPiece) {
+			err := host.addPiece(up)
+			if err != nil {
+			} else {
+			}
```
DavidVorick: empty else statement

DavidVorick: oh. empty whole statement

modules/renter/repair.go
```diff
@@ -154,19 +237,38 @@ func (r *Renter) threadedRepairUploads() {
...
+				// probabilistically filter out known bad hosts
+				// unresponsive hosts will be selected with probability 1/(1+nFailures)
+				nFailures, ok := r.blacklist[h.IPAddress]
+				if n, _ := crypto.RandIntn(1 + nFailures); ok && n != 0 {
+					continue
+				}
+
 				// TODO: use smarter duration
-				hostUploader, err := r.newHostUploader(randHosts[i], f.size, defaultDuration, f.masterKey)
+				hostUploader, err := r.newHostUploader(h, totalsize, defaultDuration, f.masterKey)
```
DavidVorick: I think that `totalsize` is too big. You should only need `totalsize` / `numPieces`

modules/renter/repair.go
```diff
@@ -154,19 +237,38 @@ func (r *Renter) threadedRepairUploads() {
...
 			// build host list
+			totalsize := f.pieceSize * uint64(f.erasureCode.NumPieces()) * f.numChunks()
 			var hosts []uploader
-			randHosts := r.hostDB.RandomHosts(f.erasureCode.NumPieces())
-			for i := range randHosts {
+			randHosts := r.hostDB.RandomHosts(f.erasureCode.NumPieces() * 2)
+			for _, h := range randHosts {
+				// probabilistically filter out known bad hosts
+				// unresponsive hosts will be selected with probability 1/(1+nFailures)
+				nFailures, ok := r.blacklist[h.IPAddress]
```
DavidVorick: don't need to do this now, but blacklist should have both a numerator and denominator. Fail = add 1 to demonminator, success = add 1 to both numerator and denmoinator

modules/renter/repair.go
```diff
@@ -154,19 +237,38 @@ func (r *Renter) threadedRepairUploads() {
...
+				if n, _ := crypto.RandIntn(1 + nFailures); ok && n != 0 {
+					continue
+				}
+
 				// TODO: use smarter duration
-				hostUploader, err := r.newHostUploader(randHosts[i], f.size, defaultDuration, f.masterKey)
+				hostUploader, err := r.newHostUploader(h, totalsize, defaultDuration, f.masterKey)
 				if err != nil {
+					// penalize unresponsive hosts
+					if strings.Contains(err.Error(), "timeout") {
```
DavidVorick: would much prefer to whitelist errors than blacklist them. We really should be adding hosts to a blacklist if they cause errors.

Only time we shouldn't add them to a blacklist in the case of an error:

1. It's our fault, not theirs
2. There's no way for a dishonest host to trigger that error - it can only be triggered internally

#2 is really important for security.

We want to ignore errors for the two following cases:

1. txn daisy chain is too long
2. wallet, due to restarting, is attempting double spends

I don't think there are other scenarios that should be whitelisted. Those two can be prevented by having the wallet:

1. return an error if a long txn chain is detected
2. have some persistence for txns it makes

I'm guessing we can't get this by 0.4.4, but I hope you agree with the long term need for whitelisting errors as opposed to blacklisting errors.

api/daemon_test.go
lukechampine: means the apply call is being tested with a bad signature, which should cause a failure

encoding/prefix.go
lukechampine: it's actually related to the blacklist haha. The blacklist relies on getting timeout errors, but if a timeout error prevents you from reading any data, you'll get `no data` instead of `timeout`. I'm not sure why I thought this was a great idea in the first place, honestly

modules/renter/repair.go
lukechampine: whoops. That's what happens when you blindly delete a bunch of printlns

modules/renter/repair.go
lukechampine: yeah, I imagine it will become a lot more sophisticated once the renter and hostdb are merged. For now, the most important errors are timeout errors. We don't really need to care about the other ones, as long as they're fast. But timeout errors significantly slow down the repair loop.

lukechampine:
ok. This has been tested and seems to perform fairly well. I'm ready to ship this to users.

PR #812, opened 2015-10-16
Create interfaces for block marshalling

mtlynch:
Create interface for marshalling type.Block and unmarshalling
consensus.processedBlock. The explicit interfaces will facilitate
testing of the consensus package (specifically the validHeader
function).


mtlynch:
Next steps, I'd like to replace the `blockMarshaller`, `processedBlockUnmarshaller`, and `dosBlocks` members of `ConsensusSet` with a `blockValidator` interface like:

```
type blockValidator interface {
  validate(types.Block) error
}

type stdBlockValidator struct {
  dosBlocks map[types.BlockID]struct{}
  blockMarshaller blockMarshaller
  processedBlockUnmarshaller processedBlockUnmarshaller
}

func (v stdBlockValidator) validate { 
  // move over current contents of validHeader()
}
```

DavidVorick:
I'm not sure that a BlockUnmarshaller makes sense, using this strategy would lead to creating an typeUnmarshaller type for every single type that gets unmarshalled. This seems like the wrong approach to mocking the encoding package, as we have dozens of types that get encoded throughout Sia.

hmm.

DavidVorick:
It would probably be sufficient to make a generic mock of the encoding package, having a type that has a `Marshal` and `Unmarshal` function implemented. For the stub case, all objects would marshal to []byte{'1'} and all calls to unmarshal would just return whatever object was already being pointed at (maybe panicking or returning an error if 'nil' was passed).

mtlynch:
> It would probably be sufficient to make a generic mock of the encoding package, having a type that has a Marshal and Unmarshal function implemented. For the stub case, all objects would marshal to []byte{'1'} and all calls to unmarshal would just return whatever object was already being pointed at (maybe panicking or returning an error if 'nil' was passed).

That makes sense, pushing a commit now that does this.

modules/consensus/block_marshal.go
```diff
@@ -0,0 +1,35 @@
...
+	stdProcessedBlockUnmarshaller struct{}
+)
+
+func (m stdBlockMarshaller) Marshal(block types.Block) []byte {
+	return encoding.Marshal(block)
+}
+
+func (um stdProcessedBlockUnmarshaller) Unmarshal(blockBytes []byte, block *processedBlock) error {
+	return encoding.Unmarshal(blockBytes, block)
+}
```
DavidVorick: is this file still needed?

modules/consensus/consensusset.go
```diff
@@ -54,6 +55,9 @@ type ConsensusSet struct {
 	// block.
 	checkingConsistency bool
 
+	// marshaler encodes and decodes between objects and byte slices.
+	marshaler encoding.GenericMarshaler
```
DavidVorick: interesting approach. Probably a good way to do things

modules/consensus/accept.go
```diff
@@ -61,7 +60,7 @@ func (cs *ConsensusSet) validHeader(tx *bolt.Tx, b types.Block) error {
 	}
 
 	// Check that the block is below the size limit.
-	if uint64(len(encoding.Marshal(b))) > types.BlockSizeLimit {
+	if uint64(len(cs.marshaler.Marshal(b))) > types.BlockSizeLimit {
```
DavidVorick: This seems reasonable.

DavidVorick:
Looks good except for the block marshaller file that's still around

modules/consensus/block_marshal.go
mtlynch: Ah, good catch. Fixed.

mtlynch:
Thanks for reviewing. Fixed the issues you mentioned and rebased onto master.

encoding/marshal.go
```diff
@@ -21,20 +21,34 @@ var (
...
+
+	// StdGenericMarshaler is an implementation of GenericMarshaler that uses
+	// the encoding.Marshal and encoding.Unmarshal functions to perform
+	// its marshaling/unmarshaling.
+	StdGenericMarshaler struct{}
+
+	// An Encoder writes objects to an output stream.
+	Encoder struct {
+		w io.Writer
+	}
```
lukechampine: is this being used anywhere?

PR #813, opened 2015-10-18
Add download timeouts

lukechampine:
Otherwise they can stall indefinitely.
This really makes me want to implement parallel downloading, but I guess it'll have to wait.

### DIFF
--- a/crypto/merkle.go
+++ b/crypto/merkle.go
@@ -28,6 +28,23 @@ func (t MerkleTree) PushObject(obj interface{}) {
 	t.Push(encoding.Marshal(obj))
 }
 
+// ReadSegments reads segments from r into the tree. If EOF is encountered
+// mid-segment, the zero-padded segment is added to the tree and no error is
+// returned.
+func (t MerkleTree) ReadSegments(r io.Reader) error {
+	buf := make([]byte, SegmentSize)
+	for {
+		_, err := io.ReadFull(r, buf)
+		if err == io.EOF {
+			break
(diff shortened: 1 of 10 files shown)


crypto/merkle.go
```diff
@@ -28,6 +28,23 @@ func (t MerkleTree) PushObject(obj interface{}) {
...
+		_, err := io.ReadFull(r, buf)
+		if err == io.EOF {
+			break
+		} else if err != nil && err != io.ErrUnexpectedEOF {
+			return err
+		}
+		t.Push(buf)
+	}
+	return nil
+}
```
DavidVorick: skeeves me out a bit that there's no crypto-grade testing for this function

modules/renter/download.go
```diff
@@ -45,6 +45,8 @@ func (hf *hostFetcher) pieces(chunk uint64) []pieceData {
 
 // fetch downloads the piece specified by p.
 func (hf *hostFetcher) fetch(p pieceData) ([]byte, error) {
+	hf.conn.SetDeadline(time.Now().Add(1 * time.Minute)) // sufficient to transfer 4 MB over 500 kbps
```
DavidVorick: 500kbps is too optimistic. Especially considering hosts might be getting multiple pieces at the same time.

Is there a way to extend the deadline? Maybe you could extend the deadline by 3 minutes once the uploading starts?

I'm not sure what our average host connection is. But I know a lot of homes have only 1mbps up or 0.5 mbps up. I don't think that there are many that are too far below that, but multiple simultaneous connections might still cause issues.

PR #816, opened 2015-10-18
Add interface for bolt database logic

mtlynch:
Add interfaces to wrap boltdb types to facilitate testing.

Replace boltdb types with interfaces in validHeader function.

### DIFF
--- a/modules/consensus/accept.go
+++ b/modules/consensus/accept.go
@@ -24,8 +24,7 @@ var (
 )
 
 // validHeader does some early, low computation verification on the block.
-func (cs *ConsensusSet) validHeader(tx *bolt.Tx, b types.Block) error {
-
+func (cs *ConsensusSet) validHeader(tx dbTx, b types.Block) error {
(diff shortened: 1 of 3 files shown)


modules/consensus/accept.go
```diff
@@ -24,7 +24,7 @@ var (
 )
 
 // validHeader does some early, low computation verification on the block.
-func (cs *ConsensusSet) validHeader(tx *bolt.Tx, b types.Block) error {
+func (cs *ConsensusSet) validHeader(tx dbTx, b types.Block) error {
```
DavidVorick: dbTx strikes me as a weird name... but I guess it's fine. It's on the edge, but it's pretty obvious that dbTx stands for databaseTransaction... I think?

DavidVorick:
except for the merge conflicts, this looks good

mtlynch:
Thanks for reviewing!

Fixed the merge conflicts.

PR #817, opened 2015-10-22
Fix Explorer Metrics

DavidVorick:
Total metrics were completely off. They are still completely manipulatable, but at least the 'TotalContractCost' can be considered reliable.

testing to follow soon.

### DIFF
--- a/modules/explorer/persist.go
+++ b/modules/explorer/persist.go
@@ -1,0 +1,23 @@
+package explorer
+
+import (
+	"os"
+)
+
+// initPersist initializes the persistent structures of the explorer module.
+func (e *Explorer) initPersist() error {
+	// Make the persist directory
+	err := os.MkdirAll(e.persistDir, 0700)
+	if err != nil {
+		return err
+	}
+
+	// Initilize the database
+	db, err := openDB(e.persistDir + "/blocks.db")
+	if err != nil {
+		return err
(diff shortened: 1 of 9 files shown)


modules/explorer/persist.go
```diff
@@ -0,0 +1,23 @@
...
+// initPersist initializes the persistent structures of the explorer module.
+func (e *Explorer) initPersist() error {
+	// Make the persist directory
+	err := os.MkdirAll(e.persistDir, 0700)
+	if err != nil {
+		return err
+	}
+
+	// Initilize the database
+	db, err := openDB(e.persistDir + "/blocks.db")
```
lukechampine: `filepath.Join`?

modules/explorer/update.go
```diff
@@ -11,24 +11,65 @@ import (
...
-			e.activeContracts -= 1
+			e.activeContractCount -= 1
 			e.activeContractCost = e.activeContractCost.Sub(diff.FileContract.Payout)
-			e.activeContractSize -= diff.FileContract.FileSize
+			e.activeContractSize = e.activeContractSize.Sub(types.NewCurrency64(diff.FileContract.FileSize))
 		}
 	}
 
+	// Compute the
+
```
lukechampine: ?

PR #822, opened 2015-10-24
Add blockRuleHelper interface

mtlynch:
Add a blockRuleHelper interface to calculate the minimum valid child
timestamp. Rename earliestChildTimestamp to minimumValidChildTimestamp.

### DIFF
--- a/modules/consensus.go
+++ b/modules/consensus.go
@@ -172,11 +172,11 @@ type (
 		// current block is presented to the subscriber.
 		ConsensusSetDigestSubscribe(ConsensusSetDigestSubscriber)
 
-		// EarliestChildTimestamp returns the earliest timestamp that is acceptable
-		// on the current longest fork according to the consensus set. This is a
-		// required piece of information for the miner, who could otherwise be at
+		// MinimumValidChildTimestamp returns the earliest timestamp that is
+		// valid on the current longest fork according to the consensus set. This is
+		// a required piece of information for the miner, who could otherwise be at
 		// risk of mining invalid blocks.
-		EarliestChildTimestamp(types.BlockID) (types.Timestamp, bool)
+		MinimumValidChildTimestamp(types.BlockID) (types.Timestamp, bool)
 
 		// GenesisBlock returns the genesis block.
 		GenesisBlock() types.Block
--- a/modules/consensus/block_rules.go
+++ b/modules/consensus/block_rules.go
@@ -1,0 +1,51 @@
+package consensus
+
+import (
+	"sort"
+
+	"github.com/NebulousLabs/Sia/encoding"
+	"github.com/NebulousLabs/Sia/types"
+)
+
+// blockRuleHelper assists with block validity checks by calculating values
+// on blocks that are relevant to validity rules.
+type blockRuleHelper interface {
+	minimumValidChildTimestamp(dbBucket, *processedBlock) types.Timestamp
+}
+
+// stdBlockRuleHelper is the standard implementation of blockRuleHelper.
+type stdBlockRuleHelper struct{}
+
+// minimumValidChildTimestamp returns the earliest timestamp that a child node
+// can have while still being valid. See section 'Block Timestamps' in
+// Consensus.md.
+//
+// To boost performance, minimumValidChildTimestamp is passed a bucket that it
+// can use from inside of a boltdb transaction.
+func (rh stdBlockRuleHelper) minimumValidChildTimestamp(blockMap dbBucket, pb *processedBlock) types.Timestamp {
+	// Get the previous MedianTimestampWindow timestamps.
+	windowTimes := make(types.TimestampSlice, types.MedianTimestampWindow)
+	windowTimes[0] = pb.Block.Timestamp
+	parent := pb.Block.ParentID
+	for i := uint64(1); i < types.MedianTimestampWindow; i++ {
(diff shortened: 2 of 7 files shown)


mtlynch:
This supports the larger effort of testing the consensus package in #803 (this commit is merged there so you can see how this fits in).

modules/consensus/block_rules.go
```diff
@@ -0,0 +1,51 @@
...
+		// parent lies at the first 32 bytes, and the timestamp of the block
+		// lies at bytes 40-48.
+		parentBytes := blockMap.Get(parent[:])
+		copy(parent[:], parentBytes[:32])
+		windowTimes[i] = types.Timestamp(encoding.DecUint64(parentBytes[40:48]))
+	}
+	sort.Sort(windowTimes)
+
+	// Return the median of the sorted timestamps.
+	return windowTimes[len(windowTimes)/2]
```
mtlynch: This is copied from the original implementation but this is actually not the median. The median should take the average of the middle two elements if the set has an odd number of elements. Does this matter?

lukechampine: Correct, but in this case the number of elements is always `types.MedianTimestampWindow`, which is even.

DavidVorick: `MedianTimestampWindow` is 11. The median is being calculated correctly by this code. Changing it would be a hardfork.

mtlynch: Meant to say middle two elements when it's an **even** number of elements. But sounds like this is the implementation you want either way.

Other than that, any blocking issues in this PR?

DavidVorick: mostly that I'm travelling and don't have much internet access.

Hopefully I'll get a few hours of good connection within the next 2 days. I'll be home on Halloween.

mtlynch:
> I'll be home on Halloween.

Then if you have any notes, I will make them in costume.

modules/consensus.go
```diff
@@ -172,11 +172,11 @@ type (
...
 
-		// EarliestChildTimestamp returns the earliest timestamp that is acceptable
-		// on the current longest fork according to the consensus set. This is a
-		// required piece of information for the miner, who could otherwise be at
+		// MinimumValidChildTimestamp returns the earliest timestamp that is
+		// valid on the current longest fork according to the consensus set. This is
+		// a required piece of information for the miner, who could otherwise be at
 		// risk of mining invalid blocks.
-		EarliestChildTimestamp(types.BlockID) (types.Timestamp, bool)
+		MinimumValidChildTimestamp(types.BlockID) (types.Timestamp, bool)
```
mtlynch: > Do you feel strongly about this name change?

EarliestChildTimestamp is a misleading name to me because I think a reasonable reader would interpret that to be "of the blocks existing children, give me the timestamp of the earliest child." I think that MinimumValidChildTimestamp potentially also has this ambiguity, but I feel like it more clearly indicates the function's purpose.

I'm open to other names, but I feel like `MinimumValidChildTimestamp` is an improvement. I get that it's longer, but it's only 4 characters longer.

DavidVorick: okay, makes sense to me

modules/consensus/consensusset.go
```diff
@@ -61,6 +61,10 @@ type ConsensusSet struct {
 	// marshaler encodes and decodes between objects and byte slices.
 	marshaler encoding.GenericMarshaler
```
DavidVorick: probably okay to remove the newline here

modules/consensus/consensusset.go
```diff
@@ -61,6 +61,10 @@ type ConsensusSet struct {
...
```
DavidVorick: *this one

modules/consensus/consensusset.go
```diff
@@ -100,6 +104,8 @@ func New(gateway modules.Gateway, persistDir string) (*ConsensusSet, error) {
 
 		marshaler: encoding.StdGenericMarshaler{},
```
DavidVorick: same thing

DavidVorick:
overall, looks pretty good. We should probably keep all of the dependencies in one pack so that the data structure doesn't get too spread out.

modules/consensus/consensusset.go
mtlynch: You want it like this?

```go
 	// marshaler encodes and decodes between objects and byte slices.
 	marshaler encoding.GenericMarshaler
	// blockRuleHelper calculates values on blocks for block validity rule
	// enforcement.
	blockRuleHelper blockRuleHelper
```

I don't have a strong objection to it because it's just whitespace but it seems like a strange request because it would make the spacing inconsistent with the rest of the struct.

modules/consensus/consensusset.go
mtlynch: ditto re: my question/concern above

modules/consensus/consensusset.go
DavidVorick: I'm going to merge this because it looks good, I'll try and find a good example with regards to the data structure formatting

modules/consensus/block_rules.go
DavidVorick: https://github.com/NebulousLabs/Sia/blob/master/modules/wallet/wallet.go

I like the way the wallet struct is formatted. Lot of fields, but they are clustered into tight groups so the overall flow is pretty visible

PR #823, opened 2015-10-26
disable broken consistency check in release constants

DavidVorick:


DavidVorick:
did IBD, I'm at 4500 blocks with no problems yet.

DavidVorick:
The build failed on the first time around: "panic: nil dsco" -> hmm. Hasn't caused problems since. But maybe we should poke around more.

PR #827, opened 2015-10-27
Ban 0.3.3.x

lukechampine:
There's still some COMPAT code in the tpool that we should consider removing.


lukechampine:
hmm. Getting nondeterministic `panic: consensus checksum mismatch after reverting`

PR #828, opened 2015-10-27
bump version to v0.4.7

lukechampine:


modules/transactionpool.go
```diff
@@ -17,8 +17,7 @@ var (
 	ErrLargeTransactionSet     = errors.New("transaction set is too large for this transaction pool")
 	ErrInvalidArbPrefix        = errors.New("transaction contains non-standard arbitrary data")
 
-	PrefixNonSia    = types.Specifier{'N', 'o', 'n', 'S', 'i', 'a'}
-	PrefixStrNonSia = "NonSia" // COMPATv0.3.3.3
```
DavidVorick: I'm not sure that we should be stripping this, there are some hosts that announced....... ah well.

DavidVorick: I'm guessing that overall it's okay. People know to try re-annoucning if they don't appear in the block explorer.

PR #831, opened 2015-11-02
remove port-forward check

DavidVorick:


lukechampine:
I just checked out the docs for Transmission and apparently they use an external site to check port connectivity. Seems like that's the only reliable way to do it.

PR #832, opened 2015-11-02
Host cleanup

lukechampine:
Added a logger and various minor improvements. Notably, the host now saves the full revision transaction, instead of just modifying the original file contract object.
I'm 90% sure this won't cause compatibility problems, but I'm running this code on my local hosting node for a while to make sure. (Host could use a lot more tests...)

### DIFF
--- a/modules/consensus/consensusset.go
+++ b/modules/consensus/consensusset.go
@@ -142,7 +142,7 @@ func (cs *ConsensusSet) BlockAtHeight(height types.BlockHeight) (block types.Blo
 		exists = true
 		return nil
 	})
-	return
+	return block, exists
 }
 
 // ChildTarget returns the target for the child of a block.
@@ -156,7 +156,7 @@ func (cs *ConsensusSet) ChildTarget(id types.BlockID) (target types.Target, exis
 		exists = true
 		return nil
 	})
-	return
+	return target, exists
 }
 
 // Close safely closes the block database.
@@ -179,7 +179,7 @@ func (cs *ConsensusSet) EarliestChildTimestamp(id types.BlockID) (timestamp type
 		exists = true
 		return nil
 	})
-	return
+	return timestamp, exists
 }
 
 // GenesisBlock returns the genesis block.
@@ -204,7 +204,7 @@ func (cs *ConsensusSet) InCurrentPath(id types.BlockID) (inPath bool) {
 		inPath = pathID == id
 		return nil
 	})
-	return
+	return inPath
 }
 
 // StorageProofSegment returns the segment to be used in the storage proof for
@@ -214,5 +214,5 @@ func (cs *ConsensusSet) StorageProofSegment(fcid types.FileContractID) (index ui
 		index, err = storageProofSegment(tx, fcid)
 		return nil
 	})
-	return
+	return index, err
 }
--- a/modules/consensus/consensusset.go
+++ b/modules/consensus/consensusset.go
@@ -214,5 +214,5 @@ func (cs *ConsensusSet) StorageProofSegment(fcid types.FileContractID) (index ui
 		index, err = storageProofSegment(tx, fcid)
 		return nil
 	})
-	return index, err
+	return
 }
--- a/modules/consensus/consensusset.go
+++ b/modules/consensus/consensusset.go
@@ -142,7 +142,7 @@ func (cs *ConsensusSet) BlockAtHeight(height types.BlockHeight) (block types.Blo
 		exists = true
 		return nil
 	})
-	return block, exists
+	return
 }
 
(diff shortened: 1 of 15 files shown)


modules/consensus/consensusset.go
```diff
@@ -142,7 +142,7 @@ func (cs *ConsensusSet) BlockAtHeight(height types.BlockHeight) (block types.Blo
 		exists = true
 		return nil
 	})
-	return block, exists
+	return
```
DavidVorick: not fan

modules/consensus/consensusset.go
```diff
@@ -204,7 +204,7 @@ func (cs *ConsensusSet) InCurrentPath(id types.BlockID) (inPath bool) {
 		inPath = pathID == id
 		return nil
 	})
-	return inPath
+	return
```
DavidVorick: still not fan

modules/gateway/gateway_test.go
```diff
@@ -32,7 +32,7 @@ func TestAddress(t *testing.T) {
 		t.Fatal("Address does not return g.myAddr")
 	}
 	port := modules.NetAddress(g.listener.Addr().String()).Port()
-	expAddr := modules.NetAddress(net.JoinHostPort("::1", port))
+	expAddr := modules.NetAddress(net.JoinHostPort("::", port))
```
DavidVorick: ?

modules/host/announce.go
```diff
@@ -70,16 +56,15 @@ func (h *Host) Announce() error {
 	addr := h.myAddr
 	h.mu.RUnlock(lockID)
 
-	// Check that the host's ip address is both known and reachable.
-	if addr.Host() == "::1" {
+	// Check that the host's ip address is known.
+	if addr.IsLocal() {
```
DavidVorick: does this also catch things like 192.168?

modules/host/host.go
```diff
@@ -2,13 +2,12 @@ package host
...
 import (
 	"errors"
+	"log"
 	"net"
-	"os"
 	"sync"
 
 	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/modules"
-	"github.com/NebulousLabs/Sia/modules/consensus"
```
DavidVorick: woo

modules/host/host.go
```diff
@@ -2,13 +2,12 @@ package host
...
 	safesync "github.com/NebulousLabs/Sia/sync"
```
DavidVorick: probably okay to move the whole package to just "sync", safesync probably not needed anymore

modules/host/negotiate.go
```diff
@@ -16,6 +16,11 @@ import (
 	"github.com/NebulousLabs/Sia/types"
 )
 
+const (
+	// maxRevisionSize is the maximum number of bytes added in a single revision
+	maxRevisionSize = 100e6 // 100 MB
```
DavidVorick: 100e6 -> one of the first times I actually had trouble figuring out what something said on my small test. Didn't help that the '0's had dots in the middle

modules/host/negotiate.go
```diff
@@ -230,7 +239,7 @@ func (h *Host) rpcUpload(conn net.Conn) error {
 
 	// send doubly-signed transaction set
 	if err := encoding.WriteObject(conn, signedTxnSet); err != nil {
-		return err
+		return errors.New("couldn't write signed transaction set: " + err.Error())
```
DavidVorick: ah, wonderful

modules/netaddress.go
```diff
@@ -39,7 +39,7 @@ func (na NetAddress) IsLocal() bool {
 	if ip := net.ParseIP(host); ip != nil && ip.IsLoopback() {
 		return true
 	}
-	if host == "localhost" || host == "0.0.0.0" {
+	if host == "localhost" || host == "::" || host == "0.0.0.0" {
```
DavidVorick: the ever growing blacklist :P

DavidVorick:
I'm feeling increasingly strongly about needing an integration test that will make sure the host is getting paid and is submitting the file contract correctly. Too much changed here and I don't feel like our test suite is strong enough for me to be confident that nothing broke.

Hosting is one of the most important parts of the network, it makes sense to throw a ton of testing at it.

modules/gateway/gateway_test.go
lukechampine: consequence of a5007062cd

modules/host/announce.go
lukechampine: no, only `127.0.0.1` and `0.0.0.0`

DavidVorick: yeah probably, or extend the function to include the local ips

PR #834, opened 2015-11-04
Host timeouts

lukechampine:
[not ready just yet -- need to run live tests]

### DIFF
--- a/modules/host/host.go
+++ b/modules/host/host.go
@@ -31,9 +31,8 @@ type contractObligation struct {
 	LastRevisionTxn types.Transaction
 	Path            string // Where on disk the file is stored.
 
-	// each obligation needs a mutex to prevent simultaneous revisions to the
-	// same obligation
-	mu *sync.Mutex
+	// revisions must happen in serial
+	mu sync.Mutex
 }
 
 // A Host contains all the fields necessary for storing files for clients and
@@ -51,8 +50,8 @@ type Host struct {
 
 	// variables
 	blockHeight         types.BlockHeight
-	obligationsByID     map[types.FileContractID]contractObligation
-	obligationsByHeight map[types.BlockHeight][]contractObligation
+	obligationsByID     map[types.FileContractID]*contractObligation
+	obligationsByHeight map[types.BlockHeight][]*contractObligation
 	spaceRemaining      int64
 	fileCounter         int
 	profit              types.Currency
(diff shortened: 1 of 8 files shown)


modules/host/host.go
```diff
@@ -31,9 +31,8 @@ type contractObligation struct {
 	LastRevisionTxn types.Transaction
 	Path            string // Where on disk the file is stored.
 
-	// each obligation needs a mutex to prevent simultaneous revisions to the
-	// same obligation
-	mu *sync.Mutex
+	// revisions must happen in serial
```
DavidVorick: hmm. Then how do we parallelize it? Seems like it's actually a really difficult problem, I'm guessing you were running into problems earlier because you weren't using the mutex correctly. There were undectable dataraces.

'thread 1 reads host:rev1'
'thread 2 reads host:rev1'
[work happens]
'thread2 grabs lock, writes host:rev1:rev2a'
'thread1 grabs lock, writes host:rev1:rev2b'
---> problem, despite no actual data races. If you want to parallelize revisions, you'll need to be fancier about it.

DavidVorick: actually, seems like parallel revisions is just overall a bad idea

modules/host/negotiate.go
```diff
@@ -231,6 +238,9 @@ func (h *Host) rpcUpload(conn net.Conn) error {
 	err = h.tpool.AcceptTransactionSet(signedTxnSet)
 	if err == modules.ErrDuplicateTransactionSet {
 		// this can happen if the host is uploading to itself
+		//
+		// TODO: is it possible for renter to cause a collision, overwriting a
+		// previous file contract?
```
DavidVorick: doesn't seem likely? what is the data that might be colliding?

Assuming that there's a siacoin output being used somewhere, you've got a guarantee that siacoin outputs will never collide. The IDs of transactions could collide though, because it's just a hash of the contents. But presumably, if you've made a colliding transaction you've already put all of the interesting information on the blockchain.

modules/host/net.go
```diff
@@ -22,6 +23,9 @@ func (h *Host) listen() {
 // TODO: maintain compatibility
 func (h *Host) handleConn(conn net.Conn) {
 	defer conn.Close()
+	// Set an initial duration that is generous, but finite. RPCs can extend
+	// this if so desired.
+	conn.SetDeadline(time.Now().Add(5 * time.Minute))
```
DavidVorick: does the ability to extend lead to a DoS vector?

Also, 5 minutes is 110 kbps for a 4MB file. But I think that's fast enough. Was thinking about people with dialup-grade upload speeds, but honestly if your upspeed is that slow you're not really going to be able to make much use of Sia anyway.

lukechampine: extending needs to be managed carefully to prevent DoS, but we're always safer with it than without it. The tricky spots are wherever we call `SetDeadline` in a loop, which for now is only `rpcRevise`.

Now that I think of it, it *is* still possible to consume a host connection indefinitely, thanks to this branch: https://github.com/NebulousLabs/Sia/pull/834/files#diff-58607cff14906af27cb7324938c37732R339
I can address this in a later PR.

PR #835, opened 2015-11-04
Move HostDB inside renter

lukechampine:
nothing too fancy here. Mostly trying to move things without changing them significantly -- that can come later.

The most important change here is that the hostdb API calls are now routed through the renter. For the 0.5 release, they should be moved to the `/renter/` route instead of `/hostdb/`. This can be done as part of the renter API overhaul.


DavidVorick:
build failed

lukechampine:
with a billion data races.
if I had a nickel for every "fix tests" commit that failed... :P

DavidVorick:
is it necessarily a good idea to hide the dataraces by changing the behavior of the program during testing? I think that's bad practice.

api/api.go
```diff
@@ -81,10 +81,10 @@ func (srv *Server) initAPI() {
...
 
-	// HostDB API Calls - Unfinished
-	if srv.hostdb != nil {
-		handleHTTPRequest(mux, "/hostdb/hosts/active", srv.hostdbHostsActiveHandler)
-		handleHTTPRequest(mux, "/hostdb/hosts/all", srv.hostdbHostsAllHandler)
+	// HostDB API Calls - DEPRECATED
+	if srv.renter != nil {
+		handleHTTPRequest(mux, "/hostdb/hosts/active", srv.renterHostsActiveHandler)
+		handleHTTPRequest(mux, "/hostdb/hosts/all", srv.renterHostsAllHandler)
 	}
```
DavidVorick: what do you know... there are two hostdb calls after all, just that only one is documented in the API.

DavidVorick: make sure you update doc/API.md to reflect the changes in the working code.

modules/host.go
```diff
@@ -19,6 +19,34 @@ var (
 	RPCDownload = types.Specifier{'D', 'o', 'w', 'n', 'l', 'o', 'a', 'd', 0}
 )
 
+var (
+	PrefixHostAnnouncement = types.Specifier{'H', 'o', 's', 't', 'A', 'n', 'n', 'o', 'u', 'n', 'c', 'e', 'm', 'e', 'n', 't'}
```
DavidVorick: Needs docstring

modules/host/host.go
```diff
@@ -184,25 +179,6 @@ func (h *Host) Info() modules.HostInfo {
...
-		averagePrice = averagePrice.Add(host.Price)
-	}
-	if len(hosts) == 0 {
-		return info
-	}
-	averagePrice = averagePrice.Div(types.NewCurrency64(uint64(len(hosts))))
-	// HACK: 4320 is one month, and 1e9 is a GB. Price is reported as per GB
-	// per month.
-	estimatedCost := averagePrice.Mul(types.NewCurrency64(4320)).Mul(types.NewCurrency64(1e9))
-	info.Competition = estimatedCost
```
DavidVorick: hmm.

Maybe the host doesn't need access to the competitive layout of the network, and instead can just optimize it's behavior and pricing (people have been asking for automatic pricing tools) based on how much attention it's getting from users.

DavidVorick:
This PR actually makes me pretty uneasy. The hostdb showed signs of instability in the past, and I just don't feel like the testing is there to provide security around the idea that this merge was done correctly.

Are there more things we can do to verify that bugs haven't been introduced into the hostdb?

Is there a test that walks through making an announcement, checking the blockchain, and verifying that the host gets into the hostdb? I believe there is.

I'm definitely definitely getting nervous about 0.5.0. We've changed a lot of things since the last release, and even though it hasn't been that many commits it's been a bit of a carpet-bomb.

lukechampine:
wrt data races, I think it's fine if a test like `TestWeightedList` isn't also doing scanning in the background. Testing functions in isolation is good. I think it's a good idea to test that exported functions don't cause data races during scanning, but why go through the effort of adding a bunch of locking to tests like `TestWeightedList`?

More generally, I think we're going to want to implement this pattern (not always spawning background threads during testing) in other modules, like the gateway. These threads make it a pain to test certain functions in isolation, because they often rely on `time.Sleep`. That means they can nondeterministically screw up your module's state depending on how slow Travis feels like running that day. So now we have test functions that call `time.Sleep` as well...it's just gross.
This is sort of related to mocking as well. I think we're going to want a "minimal `New`" and a "production `New`" for each module, where the minimal version takes tons of mocks as arguments, and the production version just takes other modules. So the minimal `New` should instantiate as bare a module as possible, preferably just creating the object and calculating some initial conditions, and leaving it up to the tests (or production `New`) to spawn background threads, create loggers, subscribe to consensus, etc.

DavidVorick:
ok

DavidVorick:
merge conflicts

lukechampine:
resolved

PR #836, opened 2015-11-08
Don't connect to loopback addresses

lukechampine:


modules/gateway/peers.go
```diff
@@ -175,6 +175,9 @@ func (g *Gateway) Connect(addr modules.NetAddress) error {
 	if addr == g.Address() {
 		return errors.New("can't connect to our own address")
 	}
+	if build.Release != "testing" && addr.IsLoopback() {
```
lukechampine: On a related note, I'd like to eliminate as many `build.Release != "testing"` statements as possible. Mocking should make this more feasible.

DavidVorick: yeah, hadn't thought about using mocking to replace the constants.

'New' could include some sort of interface that defines functions for all of the constants.

lukechampine: not really what I meant, but that's an interesting idea. I was thinking more along the lines of "if you mock an object during testing, you can implement the testing behavior inside the object instead of checking for `"testing"` inside the real functions."

PR #838, opened 2015-11-10
Hosting integration test

lukechampine:
And lots of race condition fixes along the way (benign ones, for the most part).
This also required adding a configurable duration to the `/renter/upload` call, which in turn required adding a new field to `renter.json`...more than I would have liked, but compatibility should be ok.
The repair scheme now works like this (in theory): if you manually pass in a duration, the renter will calculate the resulting ending height and use that for all future contracts for that file. So if you specify `duration=10` at height 100, whenever the renter repairs the file, it will create contracts ending at height 110, and remove the file from the repair set after that height has been reached. But if you don't supply a duration (or if `duration=0`), the renter will renew the file indefinitely. Reasonable?

### DIFF
--- a/api/host_test.go
+++ b/api/host_test.go
@@ -1,0 +1,80 @@
+package api
+
+import (
+	"io/ioutil"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/NebulousLabs/Sia/build"
+	"github.com/NebulousLabs/Sia/crypto"
+	"github.com/NebulousLabs/Sia/types"
+)
+
+// TestIntegrationHosting tests that the host correctly receives payment for
+// hosting files.
+func TestIntegrationHosting(t *testing.T) {
+	if testing.Short() {
+		//t.SkipNow()
+	}
+
+	st, err := createServerTester("TestIntegrationHosting")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// announce the host
+	err = st.stdGetAPI("/host/announce?address=" + string(st.host.Address()))
+	if err != nil {
+		t.Fatal(err)
+	}
+	// we need to announce twice, or the renter will complain about not having enough hosts
+	loopAddr := "127.0.0.1:" + st.host.Address().Port()
+	err = st.stdGetAPI("/host/announce?address=" + loopAddr)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// wait for announcement to register
+	st.miner.AddBlock()
+	var hosts ActiveHosts
+	st.getAPI("/hostdb/hosts/active", &hosts)
+	if len(hosts.Hosts) == 0 {
+		t.Fatal("host announcement not seen")
+	}
+
+	// create a file
+	path := filepath.Join(build.SiaTestingDir, "api", "TestIntegrationHosting", "test.dat")
+	data, err := crypto.RandBytes(1024)
+	if err != nil {
+		t.Fatal(err)
+	}
+	err = ioutil.WriteFile(path, data, 0600)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// upload to host
+	err = st.stdGetAPI("/renter/files/upload?nickname=test&duration=10&source=" + path)
+	if err != nil {
+		t.Fatal(err)
+	}
+	var fi []FileInfo
+	for len(fi) != 1 || fi[0].UploadProgress != 100 {
+		st.getAPI("/renter/files/list", &fi)
+		time.Sleep(3 * time.Second)
+	}
+
+	// mine blocks until storage proof is complete
+	for i := 0; i < 20+int(types.MaturityDelay); i++ {
+		st.miner.AddBlock()
+	}
+
+	// check balance
+	var wi WalletGET
+	st.getAPI("/wallet", &wi)
+	withoutProofBal := "7499794770722000000001457682072"
+	if wi.ConfirmedSiacoinBalance.String() == withoutProofBal {
+		t.Fatal("host's balance was not affected")
+	}
+}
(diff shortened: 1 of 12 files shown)


api/host_test.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/NebulousLabs/Sia/build"
+	"github.com/NebulousLabs/Sia/crypto"
+	"github.com/NebulousLabs/Sia/types"
+)
+
+// TestIntegrationHosting tests that the host correctly receives payment for
+// hosting files.
+func TestIntegrationHosting(t *testing.T) {
+	if testing.Short() {
+		//t.SkipNow()
```
DavidVorick: .

api/host_test.go
```diff
@@ -0,0 +1,80 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// announce the host
+	err = st.stdGetAPI("/host/announce?address=" + string(st.host.Address()))
+	if err != nil {
+		t.Fatal(err)
+	}
+	// we need to announce twice, or the renter will complain about not having enough hosts
```
DavidVorick: this will be a bigger issue next time we upgrade the constants.

api/host_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+	// mine blocks until storage proof is complete
+	for i := 0; i < 20+int(types.MaturityDelay); i++ {
+		st.miner.AddBlock()
+	}
+
+	// check balance
+	var wi WalletGET
+	st.getAPI("/wallet", &wi)
+	withoutProofBal := "7499794770722000000001457682072"
```
DavidVorick: doesn't seem like a good way to go about it. How did you get this number? Should probably not be hardcoded, especially since you only error in the event that the balance equals what's above

api/host_test.go
lukechampine: whoop

api/host_test.go
lukechampine: yeah, it's a bit sloppy. It's the result of mining all the blocks plus paying for all the storage.

modules/host/negotiate.go
```diff
@@ -280,6 +280,10 @@ func (h *Host) rpcRevise(conn net.Conn) error {
 	if err := encoding.ReadObject(conn, &fcid, crypto.HashSize); err != nil {
 		return errors.New("couldn't read contract ID: " + err.Error())
 	}
+
+	// remove conn deadline while we wait for lock and rebuild the Merkle tree
+	conn.SetDeadline(time.Time{})
```
DavidVorick: does this take a significant chunk of time?

modules/renter/repair.go
```diff
@@ -192,23 +192,37 @@ func (r *Renter) threadedRepairUploads() {
...
+			}
+
+			// calculate duration
+			var duration types.BlockHeight
+			if meta.EndHeight == 0 {
+				duration = defaultDuration
+			} else if meta.EndHeight > height {
+				duration = meta.EndHeight - height
+			} else {
+				r.log.Printf("removing %v from repair set: storage period has ended", name)
```
DavidVorick: this interacts okay with the storage proof delay? The host is supposed to wait 6-20 blocks or so before submitting the storage proof, bad idea to submit immediately.

api/host_test.go
DavidVorick: is there some way you can do the addition?

modules/renter/repair.go
lukechampine: I was wondering about that. Originally I added a buffer of `types.StorageProofReorgDepth`, but I wasn't sure if that was correct either.

modules/host/negotiate.go
lukechampine: no, this call is non-blocking.

modules/renter/repair.go
DavidVorick: doesn't hurt to air on the side of deleting the files late.

Though, it does seem like the tests are passing okay.

api/host_test.go
lukechampine: Possibly. I think the most practical way would be something like:
```go
uploadFile()
oldBal := getbal()
for {
    addBlock()
    diff := getbal() - oldBal
    oldBal = getbal()
    if diff doesn't end in a bunch of zeros {
        // success!
    }
}
```

modules/host/negotiate.go
DavidVorick: I meant waiting for the lock and rebuilding the Merkle tree.

lukechampine: fighting for the obligation lock could take a while, but in practice it shouldn't happen. I haven't run Merkle benchmarks in a while, but it could definitely take a bit for a large file.

api/host_test.go
DavidVorick: how about this:

1. get the balance before the proof gets submitted
2. get the balance after the proof gets submitted, and subtract all mining income.
3. make sure that upon submitting the proof, the balance increased vs. where it would be with just the coinbases.

It still seems risky/sloppy to me, but at least it's narrower.

DavidVorick:
all checks have failed

also, is not rebased

DavidVorick:
build failing

lukechampine:
yes I am aware of that thank you

lukechampine:
the problem was a race condition during revision. It's those sneaky slices again; the renter used to perform the revision like so:
```
rev.NewValidProofOutputs = fc.ValidProofOutputs
// ...
rev.NewValidProofOutputs[0].Value = rev.NewValidProofOutputs[0].Value.Sub(piecePrice)
```
But this meant we were actually modifying the memory of `fc`! This has been fixed so that the renter revises the last revision, instead of the original file contract.

lukechampine:
hmm...branch protection refuses to merge? Not sure if I like that. I don't want to have to rebase every PR I make

PR #840, opened 2015-11-11
Explore

DavidVorick:


api/explorer.go
```diff
@@ -1,62 +1,334 @@
...
+	// ExplorerGET is the object returned as a response to a GET request to
+	// /explorer.
+	ExplorerGET struct {
+		// General consensus information.
+		Height            types.BlockHeight `json:"height"`
+		CurrentBlock      types.BlockID     `json:"currentblock"`
+		Target            types.Target      `json:"target"`
+		Difficulty        types.Currency    `json:"difficulty"`
+		MaturityTimestamp types.Timestamp   `json:"maturitytimestamp"`
+		Circulation       types.Currency    `json:"circulation"`
```
lukechampine: What is `Circulation`? I assume the total number of coins in existence?

DavidVorick: Yeah, do you have a suggestion for a different name?

api/explorer.go
```diff
@@ -1,62 +1,334 @@
...
+	// ExplorerHashGET is the object returned as a response to a GET request to
+	// /explorer/hash. The HashType will indicate whether the hash corresponds
+	// to a block id, a transaction id, a siacoin output id, a file contract
+	// id, or a siafund output id. In the case of a block id, 'Block' will be
+	// filled out and all the rest of the fields will be blank. In the case of
+	// a transaction id, 'Transaction' will be filled out and all the rest of
+	// the fields will be blank. For everything else, 'Transactions' and
+	// 'Blocks' will/may be filled out and everything else will be blank.
+	ExplorerHashGET struct {
+		HashType     string                `json:"hashtype"`
```
lukechampine: are the hashtypes all explicitly documented somewhere?

api/explorer.go
```diff
@@ -1,62 +1,334 @@
...
+	if len(txids) != 0 {
+		var txns []ExplorerTransaction
+		var blocks []ExplorerBlock
+		for _, txid := range txids {
+			block, height, exists := srv.explorer.Transaction(txid)
+			if !exists && build.DEBUG {
+				panic("explorer pointing to nonexistant txn")
+			}
+			if types.TransactionID(block.ID()) == txid {
+				blocks = append(blocks, ExplorerBlock{
```
lukechampine: why does all this logic happen in the API package instead of inside the explorer?

doc/API.md
```diff
@@ -73,6 +73,82 @@ struct {
...
+
+	transactioncount          uint64
+	siacoininputcount         uint64
+	siacoinoutputcount        uint64
+	filecontractcount         uint64
+	filecontractrevisioncount uint64
+	storageproofcount         uint64
+	siafundinputcount         uint64
+	siafundoutputcount        uint64
+	minrefeecount             uint64
```
lukechampine: typo

doc/API.md
```diff
@@ -73,6 +73,82 @@ struct {
...
+	 block        api.ExplorerBlock
+	 blocks       []api.ExplorerBlock
+	 transaction  api.ExplorerTransaction
+	 transactions []api.ExplorerTransaction
+}
+```
+`hashtype` indicates what type of hash it is. The options are 'blockid',
+'transactionid', 'unlockhash', 'siacoinoutputid', 'filecontractid',
+'siafundoutputid'. If the object is a block, only the 'block' field will be
+filled out. If the object is a transaction, only the and 'transaction' field
```
lukechampine: typo

modules/explorer.go
```diff
@@ -5,89 +5,77 @@ import (
...
 
-// Used for the BlockInfo call
 type (
-	ExplorerBlockData struct {
-		ID        types.BlockID   // The id hash of the block
-		Timestamp types.Timestamp // The timestamp on the block
-		Target    types.Target    // The target the block was mined for
-		Size      uint64          // The size in bytes of the marshalled block
-	}
+	// ExplorerStatistics returns a bunch of instantaneous statistics about the
```
lukechampine: what does instantaneous mean here?

modules/explorer.go
```diff
@@ -5,89 +5,77 @@ import (
...
+	// explorer at the current height.
+	ExplorerStatistics struct {
+		// General consensus information.
+		Height            types.BlockHeight
+		CurrentBlock      types.BlockID
+		Target            types.Target
+		Difficulty        types.Currency
+		MaturityTimestamp types.Timestamp
+		Circulation       types.Currency
```
lukechampine: I assume this struct is identical to the one in the API package. You could embed it and therefore avoid the verbose conversion, or add it as an explicit field of the API struct.

modules/explorer/explorer.go
```diff
@@ -18,27 +17,41 @@ var (
...
+	siacoinInputCount         uint64
+	siacoinOutputCount        uint64
+	fileContractCount         uint64
+	fileContractRevisionCount uint64
+	storageProofCount         uint64
+	siafundInputCount         uint64
+	siafundOutputCount        uint64
+	minerFeeCount             uint64
+	arbitraryDataCount        uint64
+	transactionSignatureCount uint64
```
lukechampine: less certain in this case, but you might be able to embed an `ExplorerStatistics` struct here, like the host does with `HostSettings`

modules/explorer/info.go
```diff
@@ -24,61 +21,130 @@ func totalCurrency(height types.BlockHeight) types.Currency {
...
+	currentBlock, exists := e.cs.BlockAtHeight(e.blockchainHeight)
+	if build.DEBUG && !exists {
+		panic("current block not found in consensus set")
 	}
-	if finish > e.blockchainHeight+1 {
-		return nil, errors.New("cannot get info on a block higher than the blockchain")
+	return modules.ExplorerStatistics{
+		Height:            e.blockchainHeight,
+		CurrentBlock:      e.currentBlock,
+		Target:            target,
```
lukechampine: then this could just be `return modules.ExplorerStatistics`

modules/explorer/info.go
```diff
@@ -24,61 +21,130 @@ func totalCurrency(height types.BlockHeight) types.Currency {
...
+// appear in the blockchain.
+func (e *Explorer) FileContractID(id types.FileContractID) []types.TransactionID {
+	txnMap, exists := e.fileContractIDs[id]
+	if !exists || len(txnMap) == 0 {
+		return nil
+	}
+	ids := make([]types.TransactionID, 0, len(txnMap))
+	for txid := range txnMap {
+		ids = append(ids, txid)
+	}
```
lukechampine: I'm fairly sure you could rewrite all of these functions as:
```go
var s []sliceType
for x := range e.map[id] {
    s = append(s, x)
}
return s
```
since the callers don't distinguish between `nil` and empty, and `!exists` just means `txnMap` will be empty. But it's probably best to be explicit rather than fancy

modules/explorer/update.go
```diff
@@ -14,36 +11,298 @@ func (e *Explorer) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+				e.siafundOutputCount--
+			}
+			for _ = range txn.MinerFees {
+				e.minerFeeCount--
+			}
+			for _ = range txn.ArbitraryData {
+				e.arbitraryDataCount--
+			}
+			for _ = range txn.TransactionSignatures {
+				e.transactionSignatureCount--
```
lukechampine: could just subtract the `len` but I guess this leaves room for doing fancier stuff in the future

modules/explorer/update.go
```diff
@@ -14,36 +11,298 @@ func (e *Explorer) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+			scoid := block.MinerPayoutID(uint64(j))
+			delete(e.siacoinOutputIDs[scoid], tbid)
+			delete(e.unlockHashes[payout.UnlockHash], tbid)
+		}
+
+		// Update cumulative stats for reverted transcations.
 		for _, txn := range block.Transactions {
-			for _, fc := range txn.FileContracts {
-				e.totalContractCount -= 1
+			// Add the transction to the list of active transactions.
```
lukechampine: remove?

lukechampine: remove?

modules/explorer/update.go
```diff
@@ -14,36 +11,298 @@ func (e *Explorer) ProcessConsensusChange(cc modules.ConsensusChange) {
...
-				e.totalContractCount += 1
+			// Add the transaction to the list of active transactions.
+			txid := txn.ID()
+			e.transactionCount++
+			e.transactionHashes[txid] = e.blockchainHeight
+
+			for _, sci := range txn.SiacoinInputs {
+				_, exists := e.siacoinOutputIDs[sci.ParentID]
+				if !exists {
+					panic("siacoin input without siacoin output")
```
lukechampine: bare panics?

modules/explorer/update.go
```diff
@@ -14,36 +11,298 @@ func (e *Explorer) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+				}
+				e.siacoinOutputIDs[scoid][txid] = struct{}{}
+				_, exists = e.unlockHashes[sco.UnlockHash]
+				if !exists {
+					e.unlockHashes[sco.UnlockHash] = make(map[types.TransactionID]struct{})
+				}
+				e.unlockHashes[sco.UnlockHash][txn.ID()] = struct{}{}
+				e.siacoinOutputCount++
+			}
+			for k, fc := range txn.FileContracts {
```
lukechampine: these vars have local scope, so you can just use `j` for all of them

api/explorer.go
lukechampine: hmm. `TotalCoins` maybe? A little closer to what the number actually represents.

lukechampine:
The "database vs in-memory" implementation detail seems like a good candidate for an interface. I feel like we intended to do that in the persist package a while ago but never got around to it. Anyway, it's still low-priority for now, but maybe something for another contributor to work on.

modules/explorer/explorer.go
DavidVorick: if they were embedded the stats would have to be exported

modules/explorer/update.go
DavidVorick: yeah I wanted to make it easy add in fancier metrics for each field. For example, it'd be pretty easy to count the number of host announcement through the arb data field.

modules/explorer/update.go
DavidVorick: 4, I get all the way to 'l' in some of the loops.

I probably should break this into more functions.

DavidVorick:
I agree, I nearly made things an interface. But decided I wanted to prioritize finishing quickly instead of coding up an interface.

modules/explorer/explorer.go
lukechampine: does it matter? The `HostSettings` are exported too.
If you really don't want to embed it, you can just add a `stats modules.ExplorerStatistics` field, it's the same deal. I'd just like to avoid the needless duplication of all these fields, which is potentially error prone

api/explorer.go
DavidVorick: The constructions for the api and the explorer are different. The explorer module is designed to provide the information to other Go programs. As such, it doesn't need to worry about providing extra information such as the block id. Just call block.ID() and you can get that information yourself.

The API on the other hand provides data to programs that don't have access to functions like ID(), and so needs to include a lot of extra information that's really irrelevant when using the go interfaces directly.

Additionally, it's okay to make a bunch of function calls. It's not okay though to make a bunch of API calls, because API calls are slow, especially if you aren't making calls from the same machine (the explorer is going to be exposed over the internet in many cases, which means making API calls is super slow). The API calls should provide a lot more information to minimize the number of calls that need to be made.

modules/explorer.go
DavidVorick: I was thinking more of the calculus sense of the word. It means the statistics as they appear at that very moment.

lukechampine: "current" is probably more appropriate in that case

DavidVorick:
I opted to be pedantic instead of embedding the explorer statistics. At some point we may prefer to calculate some of the statistics on-the-fly, and it really shouldn't be the case that you are exporting fields from an object if modifying those fields can corrupt the object.

DavidVorick:
Also, I'm still having trouble with nondeterministic build failures, both locally and remotely.

Makefile
```diff
@@ -70,8 +70,8 @@ clean:
 run = Test
 pkgs = ./api ./build ./compatibility ./crypto ./encoding ./modules ./modules/consensus \
        ./modules/explorer ./modules/gateway ./modules/host ./modules/miner \
-       ./modules/renter ./modules/renter/hostdb ./modules/transactionpool \
-       ./modules/wallet ./persist ./siac ./siae ./sync ./types
+       ./modules/renter ./modules/transactionpool ./modules/wallet \
```
lukechampine: you don't want to test the hostdb?

doc/API.md
```diff
@@ -23,6 +23,8 @@ Notes:
   it is represented as a base-10 string. You must use a "bignum" library to handle
   these values, or you risk losing precision.
 
+Example GET curl call:  `curl -A "Sia-Agent" /wallet/transactions?startheight=1&endheight=250`
+Example POST curl call: `curl -A "Sia-Agent" --data "amount=123&destination=abcd" /wallet/siacoins
```
lukechampine: need an extra newline here

DavidVorick:
ok. Assuming the build passes, this is approximately complete.

DavidVorick:
Need to get this merged asap

api/explorer.go
```diff
@@ -4,59 +4,365 @@ import (
...
+// block it appears in an uses that to build an explorer transaction.
+func buildExplorerTransaction(height types.BlockHeight, parent types.BlockID, txn types.Transaction) ExplorerTransaction {
+	var scoids []types.SiacoinOutputID
+	var fcids []types.FileContractID
+	var fcvpoidss [][]types.SiacoinOutputID
+	var fcmpoidss [][]types.SiacoinOutputID
+	var fcrvpoidss [][]types.SiacoinOutputID
+	var fcrmpoidss [][]types.SiacoinOutputID
+	var sfoids []types.SiafundOutputID
+	var sfcoids []types.SiacoinOutputID
```
lukechampine: why would you not just declare a `var et ExporerTransaction`???

doc/API.md
```diff
@@ -73,6 +76,98 @@ struct {
...
+(if the hash is known at all).
+
+Response:
+```
+struct {
+	 hashtype     string
+	 block        api.ExplorerBlock
+	 blocks       []api.ExplorerBlock
+	 transaction  api.ExplorerTransaction
+	 transactions []api.ExplorerTransaction
```
lukechampine: are these structs defined in API.md anywhere?

DavidVorick: they are not, but they point you to the source file where they get defined. I don't think you need to be more concrete than that. Similarly, we don't define types.Block in API.md.

api/explorer.go
DavidVorick: heh

types/block.go
```diff
@@ -56,6 +56,20 @@ func CalculateCoinbase(height BlockHeight) Currency {
...
+func CalculateNumSiacoins(height BlockHeight) Currency {
+	deflationBlocks := BlockHeight(InitialCoinbase - MinimumCoinbase)
+	avgDeflationSiacoins := CalculateCoinbase(0).Add(CalculateCoinbase(height)).Div(NewCurrency64(2))
+	if height <= deflationBlocks {
+		deflationSiacoins := avgDeflationSiacoins.Mul(NewCurrency64(uint64(height + 1)))
+		return deflationSiacoins
+	}
+	deflationSiacoins := avgDeflationSiacoins.Mul(NewCurrency64(uint64(deflationBlocks + 1)))
+	trailingSiacoins := NewCurrency64(uint64(height - deflationBlocks)).Mul(CalculateCoinbase(height))
+	return deflationSiacoins.Add(trailingSiacoins)
```
lukechampine: this seems a lot more error-prone than:
```go
var total types.Currency
for i := 0; i < height; i++ {
    total.Add(CalculateCoinbase(i))
}
return total

DavidVorick: The more complicated way is computationally more efficient. I don't want to be doing 500,000 bignum operations every time we are counting coins (once the blockchain gets that large). Even today, 25,000 bignum operations is going to be worth optimizing.

The test for this function verifies the result against the simpler calculation.

lukechampine: 1 million bignum operations takes half a second. It doesn't hurt having an optimized version, especially since the test verifies its correctness, but readability suffers for no perceivable gain.

DavidVorick: half a second is a pretty big deal when you are going to be performing that operation many times, especially in debug mode that operation is going to get run every block. Today, that's 300 million total bignum operations (an average of 12,500 per block for 25,000 blocks), or about 2.5 minutes.

I think this is definitely a place where the optimization is worth the tradeoff.

lukechampine: ah, didn't see that this was being called in the consistency check.
Still, as a matter of practice it's prudent to at least verify that the naive version is slow before taking the time to optimize it. My guess is that the run time of the naive version would be a small percentage of the total run time of `checkConsistency`.

PR #841, opened 2015-11-13
Auto-Renew + offline checks

lukechampine:
The repair loop will now automatically renew uploaded files. For now, all user-uploaded files are auto-renewed; we'll have to add a call to enable/disable this behavior. (Loaded .sia files are not renewed, since they are not in the renter's repair set.)
Contracts are renewed when they are less than 2000 blocks away, and are renewed for 6000 blocks in the future.

Hosts are also checked for availability, and their contracts will be replaced if they do not respond to a ping within 15 seconds. This check is both slow and heavy-handed, so we may want to consider leaving it disabled until it is more sophisticated.


DavidVorick:
:+1:

DavidVorick:
Does this slow down the upload process? Definitely a next priority is to get an upload algo that doesn't spend 70% of its life idling

DavidVorick:
(metrics as well)

lukechampine:
the availability check will definitely slow down uploading, especially if you are tracking lots of files. In practice I would expect it to add a few seconds to the loop for each file being tracked, so it will take longer for the repair loop to reach files that haven't finished uploading.
We could improve on this by prioritizing unfinished files. We should be doing this anyway, honestly; no reason not to.

DavidVorick:
You want to prioritize files that are about to die as well, but right now we're only refreshing files if they are still on disk in the first place.

The simplest answer is to use more threads. A more complicated but better solution would be to keep tabs on all of the hosts that have files and ping them in a separate loop, marking files as needing help after a number of their hosts have disappeared.

This may tie in well as the beginnings of a host-relationship architecture as well.

modules/renter/repair.go
```diff
@@ -35,24 +35,35 @@ func (f *file) chunkHosts(index uint64) []modules.NetAddress {
...
-	f.mu.Lock()
-	defer f.mu.Unlock()
-	var expired []types.FileContractID
-	for id, fc := range f.contracts {
-		if currentHeight >= fc.WindowStart {
-			expired = append(expired, id)
+// A repairMap is a mapping of chunks to pieces that need repair.
+type repairMap map[uint64][]uint64
+
+// subtract deletes the values in r that also appear in m, and returns r.
```
DavidVorick: this docstring didn't help a whole lot. Maybe you could add some more context?

modules/renter/repair.go
```diff
@@ -73,25 +84,31 @@ func (f *file) incompleteChunks() map[uint64][]uint64 {
...
 
-// expiringChunks returns a map of chunks whose pieces will expire within
-// 'renewThreshold' blocks.
-func (f *file) expiringChunks(currentHeight types.BlockHeight) map[uint64][]uint64 {
-	expiring := make(map[uint64][]uint64)
+// chunksBelow returns a map of chunks whose pieces will expire before the
+// desired endHeight. Importantly, it will not return pieces that have
+// non-expiring duplicates; this prevents the renter from repairing the same
+// chunks over and over.
+func (f *file) chunksBelow(endHeight types.BlockHeight) repairMap {
```
DavidVorick: name is confusing, but the docstring is pretty helpful

modules/renter/repair_test.go
```diff
@@ -77,6 +79,59 @@ func TestRepair(t *testing.T) {
...
+		10: {3},
+		12: {4, 6},
+	}
+	res := orig.subtract(sub)
+	if !reflect.DeepEqual(exp, res) {
+		t.Fatal("maps were merged incorrectly:", exp, res)
+	}
+}
+
+func TestChunksBelow(t *testing.T) {
```
DavidVorick: docstring

DavidVorick:
Need to disable the re-upping, uploading is already as slow as can be allowed (really it's too slow, critically slow), we can't have it slowing down even more.

We really do want that auto-repair loop running though. We'll have to work on the uploading algorithm this week. We should be able to get metric information into that PR as well.

PR #842, opened 2015-11-16
On-disk changelog

DavidVorick:
I implemented an on-disk changelog for the consensus set, and then switched the miner over to using it. I also reorganized the database files so that the actual database construction happens in one file, the consensus database stuff happens in another file, and the changelog stuff happens in a third file.

I tried to document everything.

This probably only needed to be about 7 commits, but I am pretty happy with where the code ended up.


modules/consensus/consensusdb.go
```diff
@@ -0,0 +1,517 @@
...
+	bh := tx.Bucket(BlockHeight)
+	err := encoding.Unmarshal(bh.Get(BlockHeight), &height)
+	if build.DEBUG && err != nil {
+		panic(err)
+	}
+
+	// Check that there was not an underflow on the height.
+	zeroHeight := types.BlockHeight(0)
+	if height > zeroHeight-1e9 {
+		panic(height)
```
lukechampine: bare panic

modules/consensus/consensusdb.go
```diff
@@ -0,0 +1,517 @@
...
+	}
+	return types.BlockHeight(height)
```
lukechampine: looks like it...

modules/miner/persist.go
```diff
@@ -4,8 +4,46 @@ import (
...
+	settingsFile = modules.MinerDir + ".json"
+)
+
+var (
+	settingsMetadata = persist.Metadata{"Miner Settings", "0.5.0"}
 )
 
+type (
+	// Persist contains all of the persistent miner data.
+	Persist struct {
```
lukechampine: why is this type exported?

modules/miner/update.go
```diff
@@ -8,24 +8,19 @@ import (
...
+	m.persist.Height += types.BlockHeight(len(cc.AppliedBlocks))
 
 	// Update the unsolved block.
 	var exists1, exists2 bool
-	m.unsolvedBlock.ParentID = appliedIDs[len(appliedIDs)-1]
-	m.target, exists1 = m.cs.ChildTarget(m.unsolvedBlock.ParentID)
-	m.unsolvedBlock.Timestamp, exists2 = m.cs.MinimumValidChildTimestamp(m.unsolvedBlock.ParentID)
+	m.persist.UnsolvedBlock.ParentID = cc.AppliedBlocks[len(cc.AppliedBlocks)-1].ID()
+	m.persist.Target, exists1 = m.cs.ChildTarget(m.persist.UnsolvedBlock.ParentID)
+	m.persist.UnsolvedBlock.Timestamp, exists2 = m.cs.MinimumValidChildTimestamp(m.persist.UnsolvedBlock.ParentID)
```
lukechampine: it's a little weird to have all these fields prefixed with `persist` now. Oh well

modules/miner/update.go
```diff
@@ -36,6 +31,13 @@ func (m *Miner) ProcessConsensusDigest(revertedIDs, appliedIDs []types.BlockID)
 	// There is a new parent block, the source block should be updated to keep
 	// the stale rate as low as possible.
 	m.newSourceBlock()
+	m.persist.RecentChange = cc.ID
+
+	// Save the new consensus information.
+	err := m.save()
+	if err != nil {
+		m.log.Println("ERROR: " + err.Error())
```
lukechampine: with `Println`, prefer `Println("ERROR:", err)`

modules/miner/update.go
```diff
@@ -61,5 +63,5 @@ func (m *Miner) ReceiveUpdatedUnconfirmedTransactions(unconfirmedTransactions []
 			break
 		}
 	}
-	m.unsolvedBlock.Transactions = unconfirmedTransactions[0 : i+1]
+	m.persist.UnsolvedBlock.Transactions = unconfirmedTransactions[0 : i+1]
```
lukechampine: 0 index can be omitted

modules/consensus/block_validation.go
```diff
@@ -41,6 +41,32 @@ func NewBlockValidator() stdBlockValidator {
...
+		}
+		payoutSum = payoutSum.Add(payout.Value)
+	}
+
+	// Compare the payouts to the subsidy.
+	subsidy := b.CalculateSubsidy(height)
+	if subsidy.Cmp(payoutSum) != 0 {
+		return false
+	}
+	return true
```
DavidVorick: lol this isn't even code from this PR. I just copy-pasted.

modules/consensus/block_validation.go
```diff
@@ -41,6 +41,32 @@ func NewBlockValidator() stdBlockValidator {
...
 
+// checkMinerPayouts compares a block's miner payouts to the block's subsidy and
+// returns true if they are equal.
+func checkMinerPayouts(b types.Block, height types.BlockHeight) bool {
+	// Add up the payouts and check that all values are legal.
+	var payoutSum types.Currency
+	for _, payout := range b.MinerPayouts {
+		if payout.Value.IsZero() {
+			return false
+		}
```
DavidVorick: no, it's checking that no individual payout is equal to zero

modules/consensus/block_validation.go
DavidVorick: nonetheless, I will fix it

modules/consensus/changelog.go
```diff
@@ -0,0 +1,162 @@
...
+		RevertedBlocks []types.BlockID
+		AppliedBlocks  []types.BlockID
+	}
+
+	// changeNode contains a change entry and a pointer to the next change
+	// entry, and is the object that gets stored in the database.
+	changeNode struct {
+		Entry changeEntry
+		Next  modules.ConsensusChangeID
+	}
```
DavidVorick: they could, but then you'd be relying on indices, which I felt was not safe enough, especially if people are going to be copying over their consensus folder from other directories or whatever sillyness they might get into. Instead, they get an ID which points to an exact changeEntry. If the changeEntry exists, then they are guaranteed to be reading something that's sensical. If it's just an index, there's not the same certainty that the element at that index is the element you are expecting.

modules/consensus/consensusdb.go
DavidVorick: seems likely

modules/consensus/consensusdb.go
DavidVorick: very likely

modules/miner/persist.go
DavidVorick: I asked myself the same thing, and then failed to do anything about it. heh

modules/miner/update.go
DavidVorick: yeah, I was unsure about that as well.

lukechampine:
damn. Host test failed. Looks like I need to do more debugging...

PR #843, opened 2015-11-17
don't allow panics in production

DavidVorick:


modules/explorer/update.go
```diff
@@ -145,11 +145,15 @@ func (e *Explorer) ProcessConsensusChange(cc modules.ConsensusChange) {
 				_, exists := e.siacoinOutputIDs[sci.ParentID]
 				if build.DEBUG && !exists {
 					panic("siacoin input without siacoin output")
+				} else if !exists {
+					e.siacoinOutputIDs[sci.ParentID] = make(map[types.TransactionID]struct{})
```
lukechampine: slightly cleaner imo:
```
if !exists {
    if build.DEBUG {
       panic()
    }
    make(map)
}
```

DavidVorick: it's temporary, the sanity check shouldn't be failing. Just didn't have time to fix it, want to get our website serving live stats

PR #844, opened 2015-11-17
allow direct access to the api of servers without wallets

DavidVorick:


api/api.go
```diff
@@ -37,9 +37,9 @@ func HttpPOST(url string, data string) (resp *http.Response, err error) {
 
 // handleHTTPRequest is a wrapper function that handles all incoming calls to
 // the API.
-func handleHTTPRequest(mux *http.ServeMux, url string, handler http.HandlerFunc) {
+func (srv *Server) handleHTTPRequest(mux *http.ServeMux, url string, handler http.HandlerFunc) {
 	mux.HandleFunc(url, func(w http.ResponseWriter, req *http.Request) {
-		if !strings.Contains(req.UserAgent(), "Sia-Agent") {
+		if !strings.Contains(req.UserAgent(), "Sia-Agent") && srv.wallet == nil && srv.renter == nil && srv.host == nil && srv.miner == nil {
```
lukechampine: So if I'm making a request from Firefox, but `srv.wallet != nil`, the request will be accepted?
What's the desired behavior here?

DavidVorick: oh, bad logic.

desired behavior is to let the request through if the wallet is nil

DavidVorick:
fixed now

lukechampine:
why do you need to reference the others? Isn't implied that if the server has a renter, host, or miner, it also has a wallet?

I assume this is to allow browsers to directly access a siae instance. Seems like what you really want is to only expose the `/explorer` routes, yeah? If you allow outside access, won't anyone be able to call `/daemon/stop`?

DavidVorick:
yeah, this was another quick one so that Mike Pierce could start working on the website (which should dynamically fetch the stats from the explorer).

A longer term solution I think would actually be to pass a flag to the server or have some sort of configuration that tells it what should be authenticated and what should not.

This user-agent thing is not a security fix regardless for the website because anyone can make calls with a modified user agent, Aaron will have to make sure via the webserver that only `/explorer` is exposed to the public.

DavidVorick:
Referencing the other modules isn't strictly necessary, but technically the server has no notion of the module dependency tree. You can pass in any set of modules you want, including modules that have their dependencies filled but got sent in as 'nil' to the server. It's the main program that does the assembly.

lukechampine:
alternatively, you could do something like:
```go
if !strings.Contains(req.UserAgent(), "Sia-Agent") && !strings.HasPrefix(req.Path, "/explorer")
```
To always expose the explorer routes.

DavidVorick:
I did consider it. I felt uneasy about always having certain modules exposed though, felt like that could leak information too. At this point, I don't think there's anything dangerous that could be leaked but... seemed best to not open that window.

lukechampine:
depending on the user to configure their nginx properly seems a lot more dangerous to me. Now anyone running a public explorer can have their daemon stopped if they aren't aware of this (undocumented!) vulnerability.
The explorer seems designed to be public (unlike the rest of the modules) so exposing it seems ok to me. The vast majority of users aren't running one anyway, and if they are it's only listening on localhost.

DavidVorick:
> depending on the user to configure their nginx properly seems more dangerous to me. Now anyone running a public explorer can have their daemon stopped if they aren't aware of this (undocumented!) vulnerability.

I agree, but the point was that we can't control it anyway. Unless we're going to set up full password-authentication in the api we can't stop this attack. I do want to set up proper api authentication but was planning on having Jordan do it once he starts.

DavidVorick:
on account of the build failing repeatedly for known nondeterministic reasons, I'm closing this PR and pointing to #842 instead, which contains the commit found in this PR

PR #848, opened 2015-11-18
Expand unit tests for validateHeader

mtlynch:
Expand unit tests for validateHeader to cover all code paths. Fix unit
tests so that they don't assume an ordering of validation checks in
validateHeader.

### DIFF
--- a/modules/consensus/accept_test.go
+++ b/modules/consensus/accept_test.go
@@ -10,10 +10,97 @@ import (
 	"github.com/NebulousLabs/Sia/types"
 )
 
-// mockDbBucket is an implementation of dbBucket for unit testing.
-type mockDbBucket struct {
-	values map[string][]byte
-}
+var (
+	// validateBlockParamsGot stores the parameters passed to the most recent call
+	// to mockBlockValidator.ValidateBlock.
+	validateBlockParamsGot validateBlockParams
+
+	mockValidBlock = types.Block{
+		Timestamp: 100,
+		ParentID:  mockParentID(),
+	}
+	mockInvalidBlock = types.Block{
+		Timestamp: 500,
+		ParentID:  mockParentID(),
+	}
+	// parentBlockSerialized is a mock serialized form of a processedBlock.
+	parentBlockSerialized = []byte{3, 2, 1}
+
+	parentBlockUnmarshaler = mockBlockMarshaler{
+		[]predefinedBlockUnmarshal{
+			{parentBlockSerialized, mockParent(), nil},
+		},
+	}
+
+	unmarshalFailedErr = errors.New("mock unmarshal failed")
+
+	failingBlockUnmarshaler = mockBlockMarshaler{
+		[]predefinedBlockUnmarshal{
+			{parentBlockSerialized, processedBlock{}, unmarshalFailedErr},
+		},
+	}
+
+	serializedParentBlockMap = []blockMapPair{
+		{mockValidBlock.ParentID[:], parentBlockSerialized},
+	}
+)
+
+type (
+	// mockDbBucket is an implementation of dbBucket for unit testing.
+	mockDbBucket struct {
+		values map[string][]byte
+	}
+
+	// mockDbTx is an implementation of dbTx for unit testing. It uses an
+	// in-memory key/value store to mock a database.
+	mockDbTx struct {
+		buckets map[string]dbBucket
+	}
+
+	// predefinedBlockUnmarshal is a predefined response from mockBlockMarshaler.
+	// It defines the unmarshaled processedBlock and error code that
+	// mockBlockMarshaler should return in response to an input serialized byte
+	// slice.
+	predefinedBlockUnmarshal struct {
+		serialized  []byte
+		unmarshaled processedBlock
+		err         error
+	}
+
+	// mockBlockMarshaler is an implementation of the encoding.GenericMarshaler
+	// interface for unit testing. It allows clients to specify mappings of
+	// serialized bytes into unmarshaled blocks.
+	mockBlockMarshaler struct {
+		p []predefinedBlockUnmarshal
+	}
+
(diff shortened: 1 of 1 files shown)


DavidVorick:
:)

Thanks, glad to see the coverage being increased. I've got a lot on my plate this week, but this is definitely good to see.

modules/consensus/accept_test.go
```diff
@@ -27,60 +36,249 @@ func (db mockDbTx) Bucket(name []byte) dbBucket {
...
+	}
 )
 
+// blockMapPair represents a key-value pair in the mock block map.
+type blockMapPair struct {
+	key []byte
+	val []byte
+}
+
 var validateHeaderTests = []struct {
```
DavidVorick: It's probably cleaner to define these inside of the function, I'm not crazy about having them be global.

lukechampine: yeah, standard practice for table-driven tests is to put them inside the testing function. Plus then you don't have to worry about naming collisions, so you can just name them all `tests`.

mtlynch: Fixed in 1dd22a8

modules/consensus/accept_test.go
```diff
@@ -27,60 +36,248 @@ func (db mockDbTx) Bucket(name []byte) dbBucket {
...
+// AddPredefinedUnmarshal adds a predefinedBlockUnmarshal to mockBlockMarshaler.
+func (m *mockBlockMarshaler) AddPredefinedUnmarshal(u predefinedBlockUnmarshal) {
+	m.p = append(m.p, u)
+}
+
+// mockBlockRuleHelper is an implementation of the blockRuleHelper interface for
+// unit testing.
+type mockBlockRuleHelper struct {
+	minTimestamp types.Timestamp
+}
```
DavidVorick: types should all go at the top, if you feel the need to break apart types and functions you should probably be creating additional files.

DavidVorick: generally speaking though the godoc tool is pretty good at helping to keep the files organized and readable, in this case I don't think multiple files are necessary.

DavidVorick: actually, it might be good to have a block_mock.go

modules/consensus/accept_test.go
```diff
@@ -27,60 +36,248 @@ func (db mockDbTx) Bucket(name []byte) dbBucket {
...
+	called       bool
+	b            types.Block
+	minTimestamp types.Timestamp
+	target       types.Target
+	height       types.BlockHeight
+}
+
+// validateBlockParamsGot stores the parameters passed to the most recent call
+// to mockBlockValidator.ValidateBlock.
+var validateBlockParamsGot validateBlockParams
```
DavidVorick: vars should go above types.

```
package

import

const

var

type

func
```

modules/consensus/accept_test.go
```diff
@@ -27,60 +36,248 @@ func (db mockDbTx) Bucket(name []byte) dbBucket {
...
+	}
+	// parentBlockSerialized is a mock serialized form of a processedBlock.
+	parentBlockSerialized = []byte{3, 2, 1}
 	// emptyDosBlocks is a dosBlocks map where no block is marked bad.
-	emptyDosBlocks = map[types.BlockID]struct{}{}
-	// emptyTx is a transaction associated with an empty block map.
-	emptyTx = mockDbTx{map[string]dbBucket{
-		string(BlockMap): mockDbBucket{},
-	}}
+	emptyDosBlocks         = map[types.BlockID]struct{}{}
```
DavidVorick: I don't think this variable is necessary, would much rather see it spelled out in the structs.

also, you should be using `make` when creating maps.

`= make(map[types.BlockID]struct{})`

modules/consensus/accept_test.go
```diff
@@ -27,60 +36,248 @@ func (db mockDbTx) Bucket(name []byte) dbBucket {
...
+	var mockTarget types.Target
+	mockTarget[0] = 56
+	parent.ChildTarget = mockTarget
+	return parent
+}
+
 var (
-	// validBlock is a mock valid Block.
-	validBlock = types.Block{Timestamp: 100}
+	validBlock = types.Block{
```
DavidVorick: perhaps mockValidBlock

modules/consensus/accept_test.go
```diff
@@ -27,60 +36,248 @@ func (db mockDbTx) Bucket(name []byte) dbBucket {
...
+		Timestamp: 100,
+		ParentID:  mockParentID(),
+	}
+	invalidBlock = types.Block{
```
DavidVorick: similarly, mockInvalidBlock

modules/consensus/accept_test.go
mtlynch: fixed

modules/consensus/accept_test.go
mtlynch: fixed

modules/consensus/accept_test.go
mtlynch: Ah, right forgot about this. Do we have this documented anywhere?

Fixed.

modules/consensus/accept_test.go
mtlynch: Fixed

modules/consensus/accept_test.go
mtlynch: Fixed, didn't create a `mock_block.go` as I wasn't sure what would make sense to go there, since we're not creating a mock for `types.block` but open to suggestions

mtlynch:
Thanks for the review! Updated based on your notes.

DavidVorick:
Ok looks good, needs a rebase though. If you are busy I can do the rebase later this week.

mtlynch:
Finished the rebase, should be set.

PR #849, opened 2015-11-18
Miner api

DavidVorick:
This changes the miner api so that it is properly RESTful. I haven't written any unit tests for the api, and we definitely need to make sure the Sia-GPU-Miner is still compatible with the changes.

Except for testing, this PR is complete.

### DIFF
--- a/api/miner_test.go
+++ b/api/miner_test.go
@@ -1,0 +1,40 @@
+package api
+
+import (
+	"testing"
+)
+
+// TestMinerGET checks the GET call to the /miner endpoint.
+func (srv *Server) TestMinerGET(t *testing.T) {
+	if testing.Short() {
+		t.SkipNow()
+	}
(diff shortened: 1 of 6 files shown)


api/miner_test.go
```diff
@@ -0,0 +1,150 @@
...
+	// the blockchain increases.
+	resp, err = HttpPOST("http://"+st.server.listener.Addr().String()+"/miner/header", string(header[:]))
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer resp.Body.Close()
+	time.Sleep(500 * time.Millisecond)
+	if st.cs.Height() != startingHeight+1 {
+		println(st.cs.Height())
+		println(startingHeight + 1)
```
lukechampine: left over?

DavidVorick:
Does this look good to you?

PR #850, opened 2015-11-20
protect daemon calls if explorer is built

DavidVorick:
This is a dirty hack that protects the daemon calls from being exposed if there is an explorer, that allows the API to be exposed to the web without needing to worry about someone hitting the `/daemon/stop` button.

The protection is only in place when there is no wallet.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -55,11 +55,13 @@ func (srv *Server) initAPI() {
 	srv.handleHTTPRequest(mux, "/", srv.unrecognizedCallHandler)
 
 	// Daemon API Calls - Unfinished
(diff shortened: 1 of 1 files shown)


lukechampine:
What about the gateway? Don't want to allow people to remove all the gateway's peers, or add their own malicious nodes

lukechampine:
I threw together a [proof-of-concept](https://github.com/lukechampine/siae) for a separate explorer binary. Took <30min. I think it's worth considering.

DavidVorick:
Might be worth doing. I'm still in favor of the two binaries being combined, I feel weird about having two daemons.

But maybe it's merited. They do different enough things are are used in sufficiently different concepts.

PR #851, opened 2015-11-24
Siae death

DavidVorick:
I replaced siae with some configuration in siad. I know you aren't a huge fan of all the knobs, but I think it's the right way to go about things. I'd rather have 1 daemon with a bunch of buttons than many daemons. One of the places I'm drawing this from is x264, a video compression program with many many knobs. But it makes sense to have all the different settings integrated with the same binary because the core function is the same: compressing video.

In the case of Sia, the core function is to integrate with the Sia network. Renting, hosting, and exploring are all extensions of the consensus mechanisms, and all make use of the protocol. They are also all things that are only interesting to subsets of our userbase.

Changes:

+ siae dropped
+ user-agent requirement changed to be a custom string (specified in api.NewServer)
+ limitedAPI flag added that will hide any sensitive api endpoints, including /daemon/stop
+ mild test reorganization
+ testing for the new user agent and limitedAPI flag

+ flag added to configure the required user agent in the explorer. `""` is valid, and will accept all user agents
+ flag added to hide sensitive api endpoints
+ flag added to enable the explorer


lukechampine:
>In the case of Sia, the core function is to integrate with the Sia network
This is crucially different from the x264 example. x264 is a program that does One Thing Well -- compressing video. Sia (in its current incarnation) is a collection of Things. If you could point to any One Thing that Sia is supposed to do well, it's store data. To enable that, we need consensus, money management, peer management... but the block explorer is in no way essential to what Sia is or does. It's a service built on top of the core library, just like http://siapulse.com/en_US/page/network.

The fact that so much trouble is required to separate the explorer from the rest of the daemon with flags, booleans, special authentication, etc. is evidence that they don't quite belong together. A separate binary is easier to secure, easier to deploy, and does One Thing Well.

lukechampine:
Summary of IRL conversation:
- I was confused about the end goal of the explorer, which is less like a traditional block explorer and more like a "extra-introspection" mode for the daemon. For example, the explorer might be able to give you a list of all the peers you were ever connected to, or how many transactions you make per day, or other miscellaneous information that normal users don't particularly care about. It's primarily a tool for developers and power users who want greater insight into what siad is doing.
- A fully-featured explorer is likely to be a big undertaking, because it requires gathering metrics from all the other modules, and it's not totally clear what that would look like. For example, the explorer could import all the other modules and regularly call their methods. Or the modules could all import the explorer and regularly call a `Log` method with their internal data. Or the explorer could be a separate binary that regularly pings the API to gather metrics. At any rate, it seems tricky to pull off without casting a large shadow over the entire codebase. I feel that we should dedicate some serious time to designing this, and leave it as a low-priority item for now.
- When it comes to serving as the backend of a more traditional, public-facing block explorer like `explore.sia.tech`, I am still in favor of a simple binary that is vastly reduced in scope. While the local explorer tracks personal metrics, this backend would only report globally-true values like the block height and the contents of each block. To this end, I proposed combining this backend and the current `Sia-Explorer` frontend into a single repo, served by a Go HTTP server.
- For future discussion, I suggest referring to the current explorer as "metrics" or the "introspector" and the backend as the "explorer" or "block explorer."
- This conversation is ongoing, but in the interest of getting the explorer site up and running, I am ok merging this.

DavidVorick:
thanks

PR #852, opened 2015-11-25
Host improvements

DavidVorick:
I set out to improve the host, did a little bit, and then dropped the HostInfo struct from the modules package.

When I did this, I broke the api. So this PR is actually mostly about upgrading the host API to match the RESTful paradigms that we've been aiming at with the other modules. I also switched up the GET/POST orientation for some of the miner and wallet calls to be more consistent with the rest of the naming. (only `/wallet/backup` is affected in the wallet)

The next PR will be some additional changes to the API code for the other modules - no functional changes, but I found a way to write the API that should be less error prone. It's a pretty small change and should be purely cosmetic. After that I'll continue with the host stuff.

### DIFF
--- a/api/host.go
+++ b/api/host.go
@@ -32,7 +32,6 @@ type (
 func (srv *Server) hostHandlerGET(w http.ResponseWriter, req *http.Request) {
 	settings := srv.host.Settings()
 	upcomingRevenue, revenue := srv.host.Revenue()
-	storageRemaining, numContracts := srv.host.Capacity()
 	hg := HostGET{
 		Collateral:   settings.Collateral,
 		IPAddress:    settings.IPAddress,
@@ -43,9 +42,9 @@ func (srv *Server) hostHandlerGET(w http.ResponseWriter, req *http.Request) {
 		UnlockHash:   settings.UnlockHash,
 		WindowSize:   settings.WindowSize,
 
-		NumContracts:     numContracts,
+		NumContracts:     srv.host.Contracts(),
 		Revenue:          revenue,
-		StorageRemaining: storageRemaining,
+		StorageRemaining: srv.host.Capacity(),
 		UpcomingRevenue:  upcomingRevenue,
 	}
 	writeJSON(w, hg)
--- a/api/host.go
+++ b/api/host.go
@@ -14,7 +14,7 @@ func (srv *Server) hostAnnounceHandler(w http.ResponseWriter, req *http.Request)
 	// user can override this check by manually specifying the address.
 	var err error
 	if addr := req.FormValue("address"); addr != "" {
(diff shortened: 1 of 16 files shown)


api/host.go
```diff
@@ -5,65 +5,113 @@ import (
...
 				return
 			}
-			any = true
 		}
 	}
-	if !any {
-		writeError(w, "No valid configuration fields specified", http.StatusBadRequest)
-		return
+	srv.host.SetSettings(settings)
+	writeSuccess(w)
```
lukechampine: is it a good idea to fail silently here?

modules/host.go
```diff
@@ -24,72 +31,67 @@ var (
...
+		// external ip address is unknown.
+		Announce() error
+
+		// AnnounceAddress announces the specified address on the blockchain.
+		AnnounceAddress(NetAddress) error
+
+		// Capacity returns the amount of storage still available on the
+		// machine, as well as the number of unresolved file contracts the host
+		// is responsible for. The amount of space remaining can be negative.
+		Capacity() (spaceRemaining int64, openContracts uint64)
```
lukechampine: why does `Capacity` return the number of unresolved file contracts?

modules/host/host.go
```diff
@@ -38,31 +38,30 @@ type contractObligation struct {
...
+	fileCounter         int
 	obligationsByID     map[types.FileContractID]*contractObligation
 	obligationsByHeight map[types.BlockHeight][]*contractObligation
-	spaceRemaining      int64
-	fileCounter         int
 	profit              types.Currency
+	spaceRemaining      int64
+
+	// Persistent settings.
+	blockHeight types.BlockHeight
```
lukechampine: blockHeight doesn't seem to fit here

api/host.go
DavidVorick: Seemed like an odd corner case to drop an error if there were no changes.

I don't feel strongly about it either way, just seemed like it added confusing code to handle an edge case I didn't feel strongly about.

I'll bring it back if you want.

modules/host/host.go
DavidVorick: might seem more relevant when the block height gets added to the persistence. This is a change-in-progress. Adding block height will actually require two fields, `height` and `recentChange`. This will allow the host to do more efficient subscription at startup (instead of decoding every block in the database, only the ones that the host has never seen before (likely none or only one) will need to be decoded and given to the host.

It's not a setting, and maybe belongs with the utilities or module dependencies, or maybe belongs by itself. I'm not going to change it in this PR, but if you still think it doesn't belong, I can have it and `recentChange` go somewhere else in the next one.

modules/host.go
DavidVorick: it's loosely related, but I'll make a separate function for it.

PR #853, opened 2015-11-26
more strict method handling in the api

DavidVorick:
The api now handles methods by using a chain of if-else statements that
have 1-2 lines of code in them, usually to call a more complex function
to handle the request.

This strictness is introduced to minimize errors from not correctly
returning. Previously, one needed to call 'return' after each call to a
different method handler, an action which is not fully intuitive and on
multiple occasions was forgotten, resulting in bugs in production code.

This new paradigm is both more visually pleasing and hopefully less
prone to errors. This style of control-flow handling should mostly be
restricted to the api.


lukechampine:
:+1:

PR #855, opened 2015-11-26
Db upgrade

DavidVorick:
2 important things happened:

1. database will perform a backup-and-replace if the existing database is out of date, this should help create a seamless upgrade experience for our users.

2. There is a segfault that happens in you are using legacy host files, I wrapped the offending line of code in a compatibility check. @lukechampine  should confirm that it is sufficient to resolve compatibility issues.

### DIFF
--- a/modules/consensus/database.go
+++ b/modules/consensus/database.go
@@ -5,6 +5,8 @@ package consensus
 
 import (
 	"errors"
+	"fmt"
+	"os"
 
 	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/encoding"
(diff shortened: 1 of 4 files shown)


modules/consensus/database.go
```diff
@@ -21,7 +23,7 @@ var (
 	errNonEmptyBucket = errors.New("cannot remove a map with objects still in it")
 
 	dbMetadata = persist.Metadata{
-		Version: "0.4.3",
+		Version: "0.5.0",
```
lukechampine: This should only be bumped if we are introducing breaking changes the to the DB. Is that the case?

DavidVorick: Yeah, the changelog is a breaking change.

If you feel strongly about it, I can implement a separate upgrade pathway that'll add the changelog without needing to redownload everything, but that's going to be a nontrivial change, and code only used once.

Right now, IBD takes less than 4 hours even on an HDD, and it's like 10 minutes on an SSD.

PR #856, opened 2015-11-27
Move contract responsibilities into HostDB

lukechampine:
This is part of a larger effort to transition from one-off contracts to persistent host relationships, managed by the HostDB.

Related goals: enable easy parallel uploads in the renter, simplify lock management, refactor with mocking in mind

Next steps: explore changing file memory layout (to enable simpler repair algorithms), implement persistent host relationships, parallelize uploads

### DIFF
--- a/api/host_test.go
+++ b/api/host_test.go
@@ -15,9 +15,7 @@ import (
 // TestIntegrationHosting tests that the host correctly receives payment for
 // hosting files.
 func TestIntegrationHosting(t *testing.T) {
-	if testing.Short() {
-		t.SkipNow()
-	}
+	t.Skip("Disabled due to hostdb managing contracts")
 
 	st, err := createServerTester("TestIntegrationHosting")
 	if err != nil {
--- a/modules/renter.go
+++ b/modules/renter.go
@@ -43,46 +43,23 @@ type FileUploadParams struct {
 	PieceSize   uint64
 }
 
-// FileInfo is an interface providing information about a file.
-type FileInfo interface {
-	// Available indicates whether the file is available for downloading or
-	// not.
-	Available() bool
-
-	// UploadProgress is a percentage indicating the progress of the file as
-	// it is being uploaded. This percentage is calculated internally (unlike
-	// DownloadInfo) because redundancy schemes complicate the definition of
-	// "progress." Since UploadProgress includes redundancy, files will almost
-	// certainly be Available before UploadProgress == 100.
-	UploadProgress() float32
-
-	// Nickname is the nickname of the file.
-	Nickname() string
-
-	// Filesize is the size of the file.
-	Filesize() uint64
-
-	// Expiration is the block height at which the file will expire.
-	Expiration() types.BlockHeight
+// FileInfo provides information about a file.
+type FileInfo struct {
+	Nickname       string
+	Filesize       uint64
+	Available      bool    // whether file can be downloaded
+	UploadProgress float32 // percentage of full redundancy
+	Expiration     types.BlockHeight
 }
 
-// DownloadInfo is an interface providing information about a file that has
-// been requested for download.
-type DownloadInfo interface {
-	// StartTime is when the download was initiated.
-	StartTime() time.Time
-
-	// Filesize is the size of the file being downloaded.
-	Filesize() uint64
-
-	// Received is the number of bytes downloaded so far.
-	Received() uint64
-
-	// Destination is the filepath that the file was downloaded into.
-	Destination() string
-
-	// Nickname is the identifier assigned to the file when it was uploaded.
-	Nickname() string
+// DownloadInfo provides information about a file that has been requested for
+// download.
+type DownloadInfo struct {
+	Nickname    string
+	Destination string
+	Filesize    uint64
+	Received    uint64 // bytes
+	StartTime   time.Time
 }
 
 // RentInfo contains a list of all files by nickname. (deprecated)
--- a/modules/renter/download_test.go
+++ b/modules/renter/download_test.go
@@ -4,42 +4,39 @@ import (
 	"bytes"
 	"crypto/rand"
 	"io"
-	"sync"
 	"testing"
 	"time"
 
 	"github.com/NebulousLabs/Sia/crypto"
-	"github.com/NebulousLabs/Sia/modules"
 )
 
-type testHost struct {
-	ip        modules.NetAddress
+// a testFetcher simulates a host. It implements the fetcher interface.
+type testFetcher struct {
 	data      []byte
-	pieceMap  map[uint64][]pieceData // key is chunkIndex
+	pieceMap  map[uint64][]pieceData
 	pieceSize uint64
-	nAttempt  int // total number of download attempts
-	nFetch    int // number of successful download attempts
+
+	nAttempt int // total number of download attempts
+	nFetch   int // number of successful download attempts
 
 	// used to simulate real-world conditions
 	delay    time.Duration // transfers will take this long
 	failRate int           // transfers will randomly fail with probability 1/failRate
-
-	sync.Mutex
 }
 
-func (h *testHost) pieces(chunkIndex uint64) []pieceData {
-	return h.pieceMap[chunkIndex]
+func (f *testFetcher) pieces(chunkIndex uint64) []pieceData {
+	return f.pieceMap[chunkIndex]
 }
 
-func (h *testHost) fetch(p pieceData) ([]byte, error) {
-	h.nAttempt++
-	time.Sleep(h.delay)
+func (f *testFetcher) fetch(p pieceData) ([]byte, error) {
+	f.nAttempt++
+	time.Sleep(f.delay)
 	// randomly fail
-	if n, _ := crypto.RandIntn(h.failRate); n == 0 {
+	if n, _ := crypto.RandIntn(f.failRate); n == 0 {
 		return nil, io.EOF
 	}
-	h.nFetch++
-	return h.data[p.Offset : p.Offset+h.pieceSize], nil
+	f.nFetch++
+	return f.data[p.Offset : p.Offset+f.pieceSize], nil
 }
 
 // TestErasureDownload tests parallel downloading of erasure-coded data.
@@ -63,7 +60,7 @@ func TestErasureDownload(t *testing.T) {
 	const pieceSize = 10
 	hosts := make([]fetcher, rsc.NumPieces())
 	for i := range hosts {
-		hosts[i] = &testHost{
+		hosts[i] = &testFetcher{
 			pieceMap:  make(map[uint64][]pieceData),
 			pieceSize: pieceSize,
 
@@ -72,9 +69,9 @@ func TestErasureDownload(t *testing.T) {
 		}
 	}
 	// make one host really slow
-	hosts[0].(*testHost).delay = 100 * time.Millisecond
+	hosts[0].(*testFetcher).delay = 100 * time.Millisecond
 	// make one host always fail
-	hosts[1].(*testHost).failRate = 1
+	hosts[1].(*testFetcher).failRate = 1
 
 	// upload data to hosts
 	r := bytes.NewReader(data) // makes chunking easier
@@ -92,7 +89,7 @@ func TestErasureDownload(t *testing.T) {
 			t.Fatal(err)
 		}
 		for j, p := range pieces {
-			host := hosts[j%len(hosts)].(*testHost) // distribute evenly
+			host := hosts[j%len(hosts)].(*testFetcher) // distribute evenly
 			host.pieceMap[i] = append(host.pieceMap[i], pieceData{
 				uint64(i),
 				uint64(j),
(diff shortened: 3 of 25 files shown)


api/host_test.go
```diff
@@ -15,9 +15,7 @@ import (
 // TestIntegrationHosting tests that the host correctly receives payment for
 // hosting files.
 func TestIntegrationHosting(t *testing.T) {
-	if testing.Short() {
-		t.SkipNow()
-	}
+	t.Skip("Disabled due to hostdb managing contracts")
```
DavidVorick: :(

modules/renter.go
```diff
@@ -43,46 +43,23 @@ type FileUploadParams struct {
...
-	// Expiration is the block height at which the file will expire.
-	Expiration() types.BlockHeight
+// FileInfo provides information about a file.
+type FileInfo struct {
+	Nickname       string
+	Filesize       uint64
+	Available      bool    // whether file can be downloaded
+	UploadProgress float32 // percentage of full redundancy
+	Expiration     types.BlockHeight
 }
```
DavidVorick: :+1:

modules/renter/download_test.go
```diff
@@ -4,42 +4,39 @@ import (
...
 	"testing"
 	"time"
 
 	"github.com/NebulousLabs/Sia/crypto"
-	"github.com/NebulousLabs/Sia/modules"
 )
 
-type testHost struct {
-	ip        modules.NetAddress
+// a testFetcher simulates a host. It implements the fetcher interface.
```
DavidVorick: // a testFetcher
=>
// testFetcher

go lint would complain about this if it was an exported function

modules/renter/hostdb/hostdb.go
```diff
@@ -16,10 +21,32 @@ const (
...
 // The HostDB is a database of potential hosts. It assigns a weight to each
 // host based on their hosting parameters, and then can select hosts at random
 // for uploading files.
 type HostDB struct {
+	// modules
+	wallet modules.Wallet
+	tpool  modules.TransactionPool
+
+	// resources
+	log *log.Logger
```
DavidVorick: I typically put the logger down by the mutex

modules/renter/hostdb/hostdb.go
```diff
@@ -16,10 +21,32 @@ const (
...
+
+	// variables
+	blockHeight   types.BlockHeight
+	contracts     map[types.FileContractID]hostContract
+	cachedAddress types.UnlockHash // to prevent excessive address creation
+
+	// constants
+	entropy    [32]byte
+	persistDir string
```
DavidVorick: I don't think it makes sense to separate variables and constants. Or rather, I don't think it fits the general organizational style of the rest of our large objects. Most are sorted by how they integrate with the package, and not by their specific behavior. (IE separating the modules from the resources)

modules/renter/hostdb/hostdb.go
```diff
@@ -38,18 +65,51 @@ type HostDB struct {
...
+
+// a hostContract includes the original contract made with a host, along with
+// the most recent revision.
+type hostContract struct {
+	ID              types.FileContractID
+	FileContract    types.FileContract
+	LastRevisionTxn types.Transaction
+
+	// revisions must happen in serial
+	mu sync.RWMutex // TODO: consider `json:"-"`
```
DavidVorick: what's with the TODO?

modules/renter/hostdb/negotiate.go
```diff
@@ -377,38 +390,39 @@ func (hu *hostUploader) revise(rev types.FileContractRevision, piece []byte, hei
...
 			return nil, err
 		}
-		r.cachedAddress = uc.UnlockHash()
+		hdb.cachedAddress = uc.UnlockHash()
 	}
 
-	// TODO: check for existing contract?
-	err := hu.negotiateContract(filesize, duration, r.cachedAddress)
+	const filesize = 1 << 32 // 4 GiB
+	const duration = 6000    // 6 weeks
```
DavidVorick: I would like to change this to 4320 (30 days)

DavidVorick: Also, what's with the 4GiB?

DavidVorick: also, we're trying to stay in base 10, would prefer 4e9

modules/renter/hostdb/persist.go
```diff
@@ -0,0 +1,72 @@
...
+			ID:              hc.ID,
+			FileContract:    hc.FileContract,
+			LastRevisionTxn: hc.LastRevisionTxn,
+		})
+	}
+	return persist.SaveFile(saveMetadata, data, filepath.Join(hdb.persistDir, persistFilename))
+
+}
+
+func (hdb *HostDB) load() error {
```
DavidVorick: save and load both need docstrings

modules/renter/hostdb/hostdb.go
```diff
@@ -58,5 +118,7 @@ func New() *HostDB {
 	}
 	go hdb.threadedScan()
 
-	return hdb
+	cs.ConsensusSetSubscribe(hdb)
```
DavidVorick: hostdb subscribes to the consensus set again? hmm.

I guess it's okay.

modules/renter/hostdb/hostdb.go
```diff
@@ -38,18 +65,51 @@ type HostDB struct {
...
+
+		contracts:   make(map[types.FileContractID]hostContract),
 		activeHosts: make(map[modules.NetAddress]*hostNode),
 		allHosts:    make(map[modules.NetAddress]*hostEntry),
 		scanPool:    make(chan *hostEntry, scanPoolSize),
 
-		mu: sync.New(modules.SafeMutexDelay, 1),
+		persistDir: persistDir,
+	}
+	_, err := rand.Read(hdb.entropy[:])
```
DavidVorick: I forget what the hostdb entropy is used for. Something about determinism with a random number generator?

modules/renter/hostdb/negotiate.go
```diff
@@ -418,9 +432,43 @@ func (r *Renter) newHostUploader(settings modules.HostSettings, filesize uint64,
...
+	// select random hosts from remaining set
+	randHosts := hdb.randomHosts(n)
+	// replace removed hosts
+	for _, entry := range oldEntries {
+		hdb.insertNode(entry)
+	}
+	hdb.mu.Unlock()
+
+	// negotiate contracts with each host
+	for _, host := range randHosts {
```
DavidVorick: at this point, contracts get re-negotiated every time? For now, that is okay. Just clarifying.

modules/renter/hostdb/persist.go
```diff
@@ -0,0 +1,72 @@
...
+		// to avoid race conditions involving the contract's mutex, copy it
+		// manually into a new object
+		data.Contracts = append(data.Contracts, hostContract{
+			ID:              hc.ID,
+			FileContract:    hc.FileContract,
+			LastRevisionTxn: hc.LastRevisionTxn,
+		})
+	}
+	return persist.SaveFile(saveMetadata, data, filepath.Join(hdb.persistDir, persistFilename))
+
```
DavidVorick: not needed newline

modules/renter/hostdb/persist.go
```diff
@@ -0,0 +1,72 @@
...
+	for _, hc := range hdb.contracts {
+		// to avoid race conditions involving the contract's mutex, copy it
+		// manually into a new object
+		data.Contracts = append(data.Contracts, hostContract{
+			ID:              hc.ID,
+			FileContract:    hc.FileContract,
+			LastRevisionTxn: hc.LastRevisionTxn,
+		})
+	}
+	return persist.SaveFile(saveMetadata, data, filepath.Join(hdb.persistDir, persistFilename))
```
DavidVorick: right now the only thing saved is file contract obligations?

DavidVorick:
There should probably just be a constant or something somewhere that makes contracts of different sizes depending on whether you are using testing, development, or release constants. That's why the build package is there.

I think 80 blocks is reasonable for development constants, and 10 blocks is reasonable for testing constants.

DavidVorick:
I know that I'm even less good at it than you, but I think we should move in the direction of making lengthier commit messages, with the idea that the commit messages are to be used by the code reviewer to figure out what all changed and why in the PR.

Less `commit -m` unless it's like a 3 line change.

modules/renter/download_test.go
lukechampine: nope; golint specifies that for structs, the docstring should begin with either `Foo` or `A Foo`. If run `grep "^// A "` you'll see lots of examples of this.

modules/renter/hostdb/hostdb.go
lukechampine: ah, that was to prevent `json.Marshal` from accessing the `mu` field, causing a data race...I haven't tested that it would even work as hoped, though, and there are more robust ways to handle that race anyway.

modules/renter/hostdb/negotiate.go
lukechampine: 4 GiB is arbitrary. Not sure how big these contracts should be. I suppose eventually we will want them to grow larger/smaller as needed.

modules/renter/hostdb/hostdb.go
lukechampine: yeah, I wasn't sure about this. But it's the hostdb that processes consensus changes, not the renter.

modules/renter/hostdb/hostdb.go
lukechampine: generating keys during negotiation:
```go
// create our own key by combining the renter entropy with the host key
entropy := crypto.HashAll(hu.hdb.entropy, hostPublicKey)
ourSK, ourPK := crypto.StdKeyGen.GenerateDeterministic(entropy)
```

modules/renter/hostdb/negotiate.go
lukechampine: yes, that is still the case. Which is unfortunate because it doesn't jive well with the hostdb picking the contract filesize+duration.

modules/renter/hostdb/persist.go
lukechampine: not obligations, but yeah. Is there something else you think is missing?

lukechampine:
Yeah. There are often little discrepancies that get buried in the middle of a commit with a terse message. And then they cause problems later...

Do you think the `duration` parameter should be removed from the upload API call? Or might it still be useful? I can think of one use case: My hostdb has made two file contracts. One expires in 10 blocks, and one expires in 6000. I want to upload a file for 10 blocks. So the hostdb should add that file to the former contract instead of the latter, right?
Managing these relationships is going to be tricky. Maybe even NP complete :P

modules/renter/download_test.go
DavidVorick: :<

maybe in a more recent version. I swear I've had problems with this in the past.

modules/renter/hostdb/negotiate.go
DavidVorick: Why isn't the size just 0? Can't you change the size at any time?

modules/renter/hostdb/hostdb.go
DavidVorick: Now that the consensus set has persistent subscriptions, we can use those and it shouldn't affect startup time.

modules/renter/hostdb/negotiate.go
DavidVorick: I'm still uncertain why it's necessary to pick the filesize.

modules/renter/hostdb/persist.go
DavidVorick: the entropy? I'm not sure if it's important, and I'm not sure why it needs to be deterministic. But I'm guessing we want to save it if we're going through the trouble of having it be deterministic in the first place.

DavidVorick:
At this point I don't think it's necessary to specify the duration in the API call. With the relationships, it's assumed that the duration is "a long time". And if uploaders eventually want a duration, it should be about automatically deleting a file from the renter, not about setting the actual length of the file contract.

modules/renter/hostdb/negotiate.go
lukechampine: this is the total capacity of the file contract

modules/renter/hostdb/hostdb.go
lukechampine: clarification: the renter doesn't subscribe. It only needs `cs` to call `cs.Height`.

modules/renter/hostdb/persist.go
lukechampine: oh! Good call.

modules/renter/hostdb/hostdb.go
DavidVorick: ?

huh. I guess it doesn't have any automatic actions that it needs to participate in.

modules/renter/hostdb/negotiate.go
DavidVorick: (IRL conversation: you need to pick the filesize to specify the maximum amount of data that can be uploaded in the contract)

PR #858, opened 2015-11-28
Add preprocessing for the configuration to Siad

### LINKED ISSUE #650: siad port flag
currently, there's an error if the port flag to siad does not have a colon (":9970" vs "9970").

Support for both formats should exist.

DavidVorick:
Specifically, preprocessing was added for portnumbers in siad.
Previously, a port number provided as an integer resulted as an error, a
prefixed ":" was required. Now, it is okay to provide just an integer as
the preprocesser will clean up the number.

This was done by adding a function `preprocessConfig` which can be
extended to clean up other configuration values in the future.

fixes #650

---

'siad' previously had no testing. In this commit, a test package was
added for 'siad' so that `preprocessConfig` could be unit tested.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ run = Test
 pkgs = ./api ./build ./compatibility ./crypto ./encoding ./modules ./modules/consensus \
        ./modules/explorer ./modules/gateway ./modules/host ./modules/renter/hostdb \
        ./modules/miner ./modules/renter ./modules/transactionpool ./modules/wallet \
-       ./persist ./siac ./sync ./types
+       ./persist ./siac ./siad ./sync ./types
 
 # fmt calls go fmt on all packages.
 fmt:
--- a/siad/daemon.go
+++ b/siad/daemon.go
@@ -3,7 +3,6 @@ package main
 import (
 	"fmt"
 	"path/filepath"
-	"runtime"
 	"strconv"
 	"time"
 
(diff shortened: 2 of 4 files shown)


lukechampine:
It's a weird situation because you can also pass in a full address (host:port). Maybe we should just disallow that and only support passing in a port number? Also, did you intentionally leave out the API addr?

At any rate, I think this works better as a pure function:
```go
// processAddr prepends a ":" to bare port numbers.
func processAddr(addr string) string {
	_, err := strconv.Atoi(addr)
	if err == nil {
		addr = ":" + addr
	}
	return addr
}

func startDaemon() error {
	// ...
	apiAddr := processAddr(config.Siad.APIAddr)
	rpcAddr := processAddr(config.Siad.RPCAddr)
	hostAddr := processAddr(config.Siad.HostAddr)
}
```
This makes the test simpler (no reliance on global vars) and we can easily add a `processConfig` at a later time by reusing the `processAddr` function.

DavidVorick:
I did intentionally leave out the api addr, though maybe I shouldn't have. Some people do like to expose their api over something other than localhost (to a private network, for example).  I should add it.

I like the idea of having a preprocess function, but maybe we aren't yet doing enough for it to be necessary.

lukechampine:
we don't need a generic processing function right now, and it's questionable whether we ever will. It is overcomplicating what should be a very simple test.

DavidVorick:
oh dear. Rebase gone sour

DavidVorick:
https://www.youtube.com/watch?v=GnIB8pjhQA8

PR #863, opened 2015-11-30
transactions fix

Mingling94:


DavidVorick:
I don't like this change.

PR #865, opened 2015-11-30
Host Pools

lukechampine:
check those lengthy commit messages! woo

This diff is hilariously useless, but the gist of it is that `repair.go` underwent some serious refactoring to make optimization easier. (Note how easy it would be to spawn multiple `r.threadedRepairFile` goroutines.)

Some unused code got tossed out, specifically code related to replacing expiring contracts and checking for offline hosts. We'll be adding a new "refresh contract" RPC to the host soon, so I figure that will be a good time to reevaluate how auto-renewing should be managed. At this stage, the HostDB is still making per-file contracts, so renewing an individual file is pretty straightforward. But this will change once the HostDB migrates to long-term host relationships, so things can get tricky.

Good news: as of this PR, chunks should finally be distributed "properly" across hosts. That is, hosts should never receive more than one piece of each chunk.

### DIFF
--- a/modules/renter/hostdb/negotiate.go
+++ b/modules/renter/hostdb/negotiate.go
@@ -443,36 +443,65 @@ func (hdb *HostDB) newHostUploader(settings modules.HostSettings) (*hostUploader
 	return hu, nil
 }
 
-// UniqueHosts will return up to 'n' unique hosts that are not in 'old'. Note
-// that this is a blocking call that performs network I/O.
-func (hdb *HostDB) UniqueHosts(n int, old []Uploader) (hosts []Uploader) {
-	// remove old hosts from tree
-	hdb.mu.Lock()
-	var oldEntries []*hostEntry
-	for _, host := range old {
-		node, exists := hdb.activeHosts[host.Address()]
-		if !exists {
-			continue
-		}
-		node.removeNode()
-		delete(hdb.activeHosts, host.Address()) // probably unnecessary
-		oldEntries = append(oldEntries, node.hostEntry)
-	}
-	// select random hosts from remaining set
-	randHosts := hdb.randomHosts(n)
-	// replace removed hosts
-	for _, entry := range oldEntries {
-		hdb.insertNode(entry)
-	}
-	hdb.mu.Unlock()
+// A HostPool is a collection of hosts used to upload a file.
+type HostPool interface {
+	// UniqueHosts will return up to 'n' unique hosts that are not in 'old'.
+	UniqueHosts(n int, old []modules.NetAddress) []Uploader
 
-	// negotiate contracts with each host
+	// Close terminates all connections in the host pool.
+	Close() error
+}
+
+type pool struct {
+	hosts []*hostUploader
+	hdb   *HostDB
+}
+
+func (p *pool) Close() error {
+	for _, h := range p.hosts {
+		h.Close()
+	}
+	return nil
+}
+
+// UniqueHosts will return up to 'n' unique hosts that are not in 'old'. Note
+// that this may require negotiating new contracts.
+func (p *pool) UniqueHosts(n int, old []modules.NetAddress) (hosts []Uploader) {
+	// first reuse existing connections
+outer:
+	for _, h := range p.hosts {
+		for _, ip := range old {
+			if h.Address() == ip {
+				continue outer
+			}
+		}
+		hosts = append(hosts, h)
+	}
+
+	// TODO: revise old, unfilled contracts
+
+	// form new contracts from randomly-picked nodes
+	p.hdb.mu.Lock()
+	randHosts := p.hdb.randomHosts(n, old)
+	p.hdb.mu.Unlock()
 	for _, host := range randHosts {
-		hostUploader, err := hdb.newHostUploader(host)
+		h, err := p.hdb.newHostUploader(host)
 		if err != nil {
 			continue
 		}
-		hosts = append(hosts, hostUploader)
+		hosts = append(hosts, h)
+		p.hosts = append(p.hosts, h)
 	}
 	return hosts
+}
+
+// NewPool returns an empty HostPool, unless the HostDB contains no hosts at
+// all.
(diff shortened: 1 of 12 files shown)


modules/renter/hostdb/negotiate.go
```diff
@@ -443,36 +443,75 @@ func (hdb *HostDB) newHostUploader(settings modules.HostSettings) (*hostUploader
...
+	// Close terminates all connections in the host pool.
+	Close() error
+}
+
+type pool struct {
+	hosts []*hostUploader
+	hdb   *HostDB
+}
+
+func (p *pool) Close() error {
```
DavidVorick: docstring

modules/renter/hostdb/negotiate.go
```diff
@@ -443,36 +443,75 @@ func (hdb *HostDB) newHostUploader(settings modules.HostSettings) (*hostUploader
...
+// A HostPool is a collection of hosts used to upload a file.
+type HostPool interface {
+	// UniqueHosts will return up to 'n' unique hosts that are not in 'old'.
+	UniqueHosts(n int, old []modules.NetAddress) []Uploader
+
+	// Close terminates all connections in the host pool.
+	Close() error
+}
+
+type pool struct {
```
DavidVorick: docstring

modules/renter/hostdb/negotiate.go
```diff
@@ -443,36 +443,75 @@ func (hdb *HostDB) newHostUploader(settings modules.HostSettings) (*hostUploader
...
-func (hdb *HostDB) UniqueHosts(n int, old []Uploader) (hosts []Uploader) {
-	// remove old hosts from tree
-	hdb.mu.Lock()
-	var oldEntries []*hostEntry
-	for _, host := range old {
-		node, exists := hdb.activeHosts[host.Address()]
-		if !exists {
-			continue
+// that this may require negotiating new contracts.
+func (p *pool) UniqueHosts(n int, old []modules.NetAddress) (hosts []Uploader) {
```
DavidVorick: 'old' is a confusing variable name, and the docstring should probably be more verbose, as this function gets exported from the package. Explaining how the unique hosts are selected and why is probably a good idea (how at a high level)

DavidVorick: does this function form contracts? If it does, that should definitely be specified in the docstring, as well as indicating whether it's network-blocking or not.

I think it's going to be good practice to generally not have any network blocking on exported functions, though it's not clear that there will always be a way to achieve that safely.

DavidVorick: derp the diff made it confusing to read

modules/renter/hostdb/negotiate.go
```diff
@@ -443,36 +443,75 @@ func (hdb *HostDB) newHostUploader(settings modules.HostSettings) (*hostUploader
...
-	// replace removed hosts
-	for _, entry := range oldEntries {
-		hdb.insertNode(entry)
 	}
-	hdb.mu.Unlock()
 
-	// negotiate contracts with each host
+	// TODO: revise old, unfilled contracts
+
+	// form new contracts from randomly-picked nodes
```
DavidVorick: I'm increasingly of the opinion that we should not batch multiple files into a single file contract until either:

A. we have a pressing need due to scalability issues
B. we have a good garbage collection strategy.
C. we have a good UX plan for letting users know that the data won't be coming down. An example of this strategy might involve maintaining a git-ian style history of the renter (almost like IPFS) so that deleted or changed files can be reverted to an earlier state. This gives us a good excuse to not garbage collect, but in itself would be a pretty huge undertaking.

lukechampine: agreed. We have enough "scalability runway" that we can move it low on the priority list.

modules/renter/hostdb/negotiate.go
lukechampine: What's the rationale for not doing network I/O on exported functions? Where would the I/O happen instead?

modules/renter/hostdb/negotiate.go
```diff
@@ -443,36 +443,75 @@ func (hdb *HostDB) newHostUploader(settings modules.HostSettings) (*hostUploader
...
+	p.hdb.mu.Lock()
+	randHosts := p.hdb.randomHosts(n*2, old)
+	p.hdb.mu.Unlock()
 	for _, host := range randHosts {
-		hostUploader, err := hdb.newHostUploader(host)
+		h, err := p.hdb.newHostUploader(host)
```
DavidVorick: maybe `hu`? I usually think of `h` as a `Host`, it's a little confusing.

Probably okay though.

modules/renter/hostdb/negotiate_test.go
```diff
@@ -259,3 +259,34 @@ func TestReviseContract(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+// TestUniqueHosts tests the UniqueHosts method of the pool type.
+// TODO: make pool safer to test, using real or mocked hostdb
```
DavidVorick: what do you mean by 'safer to test'? I am trying to stop merging TODOs into the master branch, either they should be done in the PR or it should be accepted that they don't really fit into the time/resource constraints of our development cycle.

modules/renter/repair.go
```diff
@@ -21,51 +18,99 @@ const (
...
-// chunkHosts returns the IPs of the hosts storing a given chunk.
-func (f *file) chunkHosts(index uint64) []modules.NetAddress {
-	var hosts []modules.NetAddress
-	for _, fc := range f.contracts {
-		for _, p := range fc.Pieces {
-			if p.Chunk == index {
-				hosts = append(hosts, fc.IP)
-				break
-			}
+// repair attempts to repair a chunk of f by uploading its pieces to more hosts.
```
DavidVorick: confusing: `chunk of f by`

lukechampine:
comments addressed

PR #867, opened 2015-12-01
Explorer buff

DavidVorick:
I added some fields to the explorer API that were needed to provide good information from the webpage. There is still one feature missing, and that's the estimated hashrate. I don't think I'm going to implement it in this iteration.

The code is kind of messy.

### DIFF
--- a/modules/explorer.go
+++ b/modules/explorer.go
@@ -15,11 +15,11 @@ type (
 	// were at a specific block.
 	BlockFacts struct {
 		BlockID           types.BlockID     `json:"blockid"`
-		Height            types.BlockHeight `json:"height"`
-		Target            types.Target      `json:"target"`
-		MaturityTimestamp types.Timestamp   `json:"maturitytimestamp"`
-		Difficutly        types.Currency    `json:"difficulty"`
+		Difficulty        types.Currency    `json:"difficulty"`
 		EstimatedHashrate types.Currency    `json:"estimatedhashrate"`
+		Height            types.BlockHeight `json:"height"`
+		MaturityTimestamp types.Timestamp   `json:"maturitytimestamp"`
+		Target            types.Target      `json:"target"`
 		TotalCoins        types.Currency    `json:"totalcoins"`
 
 		// Transaction type counts.
--- a/modules/explorer.go
+++ b/modules/explorer.go
@@ -11,74 +11,43 @@ const (
 )
 
 type (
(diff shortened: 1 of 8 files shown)


modules/explorer.go
```diff
@@ -98,14 +67,29 @@ type (
...
 		// SiacoinOutputID returns all of the transaction ids associated with
 		// the provided siacoin output id.
 		SiacoinOutputID(types.SiacoinOutputID) []types.TransactionID
 
+		// FileContractHistory returns the history associated with a file
+		// contract.  History is broken in to 3 elements, and a bool is
+		// returned indicating whether history exists as a part of each
+		// element. If the file contract itself does not exist, no other type
+		// of history will exist either.
+		FileContractHistory(types.FileContractID) (types.FileContract, []types.FileContractRevision, types.StorageProof, bool, bool, bool)
```
lukechampine: this is a bit messy...consider splitting into multiple functions or combining return values into a struct. Three "exists" bools is just too much.

DavidVorick:
updated

modules/explorer/info_test.go
```diff
@@ -4,15 +4,18 @@ import (
 	"testing"
 )
 
-// TestStatistics grabs the statistics object from the block explorer and
-// verifies that the data has been filled out.
-func TestStatistics(t *testing.T) {
-	et, err := createExplorerTester("TestStatistics")
+// TestImmediateBlockFacts grabs the block facts object from the block explorer
+// at the current height and verifies that the data has been filled out.
+func TestImmedieateBlockFacts(t *testing.T) {
```
lukechampine: typo

types/target.go
```diff
@@ -42,6 +42,11 @@ func (x Target) Cmp(y Target) int {
 	return x.Int().Cmp(y.Int())
 }
 
+// Difficulty returns the difficulty associated with a given target.
+func (t Target) Difficulty() Currency {
+	return NewCurrency(new(big.Int).Div(RootDepth.Int(), t.Int()))
+}
```
lukechampine: does this simplify math in other functions? Why does it return a `Currency`? Is this function being called anywhere?

DavidVorick: `Currency` should probably be renamed to `UInt` or something. `types.UInt` should be pretty informative. Maybe `SafeUInt`

DavidVorick:
Cannot be merged because it's missing a few commits

DavidVorick:
ok, should be good now

PR #868, opened 2015-12-09
Refactor + cleanup hostdb

lukechampine:
I know I literally just ranted about "misc" PRs, but this is (kind of) one of them. Some minor fixes, but mostly shifting code around, splitting up functions better (note the new `upload.go`), making things easier to understand+maintain. It's still fairly "warm" and subject to change, but this is definitely a step in the right direction.
Renewing will follow shortly after this.

### DIFF
--- a/modules/renter/hostdb/hostdb.go
+++ b/modules/renter/hostdb/hostdb.go
@@ -1,4 +1,4 @@
-// package hostdb provides a HostDB object that implements the renter.hostDB
+// Package hostdb provides a HostDB object that implements the renter.hostDB
 // interface. The blockchain is scanned for host announcements and hosts that
 // are found get added to the host database. The database continually scans the
 // set of hosts it has found and updates who is online.
--- a/modules/renter/hostdb/hostdb.go
+++ b/modules/renter/hostdb/hostdb.go
@@ -71,9 +71,6 @@ type hostContract struct {
 	FileContract    types.FileContract
 	LastRevisionTxn types.Transaction
 	SecretKey       crypto.SecretKey
-
-	// revisions must happen in serial
-	mu sync.RWMutex
 }
 
 // New creates and starts up a hostdb. The hostdb that gets returned will not
--- a/modules/renter/hostdb/hostdb.go
+++ b/modules/renter/hostdb/hostdb.go
@@ -69,6 +69,7 @@ type hostContract struct {
 	IP              modules.NetAddress
 	ID              types.FileContractID
 	FileContract    types.FileContract
+	LastRevision    types.FileContractRevision
 	LastRevisionTxn types.Transaction
 	SecretKey       crypto.SecretKey
 }
--- a/modules/renter/hostdb/hostdb.go
+++ b/modules/renter/hostdb/hostdb.go
@@ -66,6 +66,7 @@ type HostDB struct {
 // a hostContract includes the original contract made with a host, along with
 // the most recent revision.
 type hostContract struct {
+	IP              modules.NetAddress
 	ID              types.FileContractID
 	FileContract    types.FileContract
 	LastRevisionTxn types.Transaction
--- a/modules/renter/hostdb/negotiate.go
+++ b/modules/renter/hostdb/negotiate.go
@@ -262,6 +262,7 @@ func (hu *hostUploader) negotiateContract(filesize uint64, duration types.BlockH
 
 	hu.hdb.mu.Lock()
 	hu.hdb.contracts[fcid] = hostContract{
+		IP:           hu.settings.IPAddress,
 		ID:           fcid,
 		FileContract: fc,
 		LastRevisionTxn: types.Transaction{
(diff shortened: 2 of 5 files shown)


DavidVorick:
Even at current prices, 1000 SC / GB / Mo is $83.4 / TB / Mo, it's pretty steep

lukechampine:
yeah...people were complaining about 500 SC on Slack, but on second thought I'm not sure why. I can revert that pretty easily if you think it's warranted.

DavidVorick:
I think 500 will continue to be a reasonable ceiling, even if the price keeps falling.

modules/renter/hostdb/hostdb.go
```diff
@@ -66,13 +66,11 @@ type HostDB struct {
...
 // the most recent revision.
 type hostContract struct {
+	IP              modules.NetAddress
 	ID              types.FileContractID
 	FileContract    types.FileContract
 	LastRevisionTxn types.Transaction
 	SecretKey       crypto.SecretKey
-
-	// revisions must happen in serial
-	mu sync.RWMutex
```
DavidVorick: the mutex is gone?

lukechampine: yeah. It's kind of a logic error for two threads to be revising the same contract anyway. So now the onus is on the renter not to call `Upload` concurrently on the same `Uploader` (it already avoids doing this anyway). Removing the mutex decreases complexity and prevents a lot of annoying pass-by-value errors ([as seen in the host](https://github.com/NebulousLabs/Sia/blob/master/modules/host/persist.go#L40)).

modules/renter/hostdb/negotiate.go
```diff
@@ -220,304 +119,207 @@ func (hu *hostUploader) negotiateContract(filesize uint64, duration types.BlockH
...
+				NewRevisionNumber:     fc.RevisionNumber,
+				NewFileSize:           fc.FileSize,
+				NewFileMerkleRoot:     fc.FileMerkleRoot,
+				NewWindowStart:        fc.WindowStart,
+				NewWindowEnd:          fc.WindowEnd,
+				NewValidProofOutputs:  []types.SiacoinOutput{fc.ValidProofOutputs[0], fc.ValidProofOutputs[1]},
+				NewMissedProofOutputs: []types.SiacoinOutput{fc.MissedProofOutputs[0], fc.MissedProofOutputs[1]},
+				NewUnlockHash:         fc.UnlockHash,
+			}},
 		},
```
DavidVorick: do you also need the signatures?

modules/renter/hostdb/negotiate.go
```diff
@@ -220,304 +119,207 @@ func (hu *hostUploader) negotiateContract(filesize uint64, duration types.BlockH
...
-	hu.hdb.mu.RUnlock()
-	if !exists {
-		return 0, errors.New("no record of contract to revise")
+// newContract negotiates an initial file contract with the specified host
+// and returns a hostContract. The contract is also saved by the HostDB.
+func (hdb *HostDB) newContract(host modules.HostSettings) (hostContract, error) {
+	// reject hosts that are too expensive
+	if host.Price.Cmp(maxPrice) > 0 {
+		return hostContract{}, errTooExpensive
 	}
```
DavidVorick: seems odd to me that you'd be doing this check here, seems like the code belongs in the previous function or in a 'checkTerms' type of function.

modules/renter/hostdb/negotiate.go
```diff
@@ -220,304 +119,207 @@ func (hu *hostUploader) negotiateContract(filesize uint64, duration types.BlockH
...
 
-	// offset is old filesize
-	offset := hu.lastTxn.FileContractRevisions[0].NewFileSize
-
-	// revise the file contract
-	err := hu.revise(hu.lastTxn.FileContractRevisions[0], data, height)
-	if err != nil {
-		return 0, err
+	// get an address to use for negotiation
+	// TODO: use more than one shared address
```
DavidVorick: TODO should be removed

DavidVorick:
very exhausted, but I looked at most of it. Seems mostly moved around. We should touch base IRL and go through the now-updated architecture.

modules/renter/hostdb/negotiate.go
lukechampine: So, this is one of the ugly constructions that remains. When you revise, you want to revise the previous revision. But after the initial contract negotiation, there are no revisions yet. So a dummy "0th revision" is created for future code to build off of.
Another related bit of ugliness is that you have to reference the revision with `contract.LastRevisionTxn.FileContractRevisions[0]`, which is both verbose and risks an out-of-bounds error. One obvious way to rectify this would be to add a `LastRevision` field to the `hostContract` type, but then you're duplicating data and you have to be careful that it stays synchronized with the revision in `LastRevisionTxn`.

modules/renter/hostdb/negotiate.go
lukechampine: by "previous function" do you mean `negotiateContract`? That function doesn't have access to the host's price (just the total contract payout), so it would have to be passed as an additional parameter.
Perhaps better would be for the hostdb to not return expensive hosts from `randomHosts`?

PR #872, opened 2015-12-14
Implement Renew RPC and add renew logic

lukechampine:
`RPCRenew` is a new host RPC that allows a renter to renew a contract without reuploading. The renter now makes use of this RPC to renew files that are within 2000 blocks of expiring.

The `negotiateContract` functions for both the host and renter have been refactored so that they can be used for renewing as well as the initial negotiation process. It's not a perfect fit, but it saves a lot of duplicated code. This may not be possible if the `RPCRenew` protocol is changed in the future.

### DIFF
--- a/api/host_test.go
+++ b/api/host_test.go
@@ -15,8 +15,6 @@ import (
 // TestIntegrationHosting tests that the host correctly receives payment for
 // hosting files.
 func TestIntegrationHosting(t *testing.T) {
-	t.Skip("Disabled due to hostdb managing contracts")
-
 	st, err := createServerTester("TestIntegrationHosting")
 	if err != nil {
 		t.Fatal(err)
@@ -25,15 +23,6 @@ func TestIntegrationHosting(t *testing.T) {
 	// Announce the host.
 	announceValues := url.Values{}
 	announceValues.Set("address", string(st.host.NetAddress()))
-	err = st.stdPostAPI("/host/announce", announceValues)
-	if err != nil {
-		t.Fatal(err)
-	}
-	// Announce twice, otherwise the renter will throw a 'not enough hosts'
-	// error.
-	loopAddr := "127.0.0.1:" + st.host.NetAddress().Port()
-	announceValues = url.Values{}
-	announceValues.Set("address", loopAddr)
 	err = st.stdPostAPI("/host/announce", announceValues)
 	if err != nil {
 		t.Fatal(err)
@@ -63,8 +52,9 @@ func TestIntegrationHosting(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	// only one piece will be uploaded (10% at current redundancy)
 	var fi []FileInfo
-	for len(fi) != 1 || fi[0].UploadProgress != 100 {
+	for len(fi) != 1 || fi[0].UploadProgress != 10 {
 		st.getAPI("/renter/files/list", &fi)
 		time.Sleep(3 * time.Second)
 	}
@@ -80,7 +70,7 @@ func TestIntegrationHosting(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	expRevenue := "382129999999997570526"
+	expRevenue := "31844166666666464210"
 	if hg.Revenue.String() != expRevenue {
 		t.Fatalf("host's profit was not affected: expected %v, got %v", expRevenue, hg.Revenue)
 	}
--- a/modules/host/negotiate.go
+++ b/modules/host/negotiate.go
@@ -246,7 +246,6 @@ func (h *Host) negotiateContract(conn net.Conn, merkleRoot crypto.Hash, filename
 
 	// Add this contract to the host's list of obligations.
 	h.mu.Lock()
-	h.fileCounter++
 	co := &contractObligation{
 		ID:           contractTxn.FileContractID(0),
 		FileContract: contractTxn.FileContracts[0],
@@ -277,7 +276,8 @@ func (h *Host) rpcUpload(conn net.Conn) error {
 	}
 
 	h.mu.RLock()
-	filename := filepath.Join(h.persistDir, strconv.Itoa(h.fileCounter+1))
+	h.fileCounter++
+	filename := filepath.Join(h.persistDir, strconv.Itoa(h.fileCounter))
 	h.mu.RUnlock()
 
 	// negotiate expecting empty Merkle root
@@ -444,6 +444,24 @@ func (h *Host) rpcRenew(conn net.Conn) error {
 	defer obligation.mu.Unlock()
 	merkleRoot := obligation.LastRevisionTxn.FileContractRevisions[0].NewFileMerkleRoot
 
-	// reuse old obligation path
-	return h.negotiateContract(conn, merkleRoot, obligation.Path)
+	// copy over old file data
+	h.mu.RLock()
+	h.fileCounter++
+	filename := filepath.Join(h.persistDir, strconv.Itoa(h.fileCounter))
+	h.mu.RUnlock()
+
+	old, err := os.Open(obligation.Path)
+	if err != nil {
+		return err
+	}
+	renewed, err := os.Create(filename)
+	if err != nil {
+		return err
+	}
+	_, err = io.Copy(renewed, old)
+	if err != nil {
+		return err
+	}
+
+	return h.negotiateContract(conn, merkleRoot, filename)
 }
--- a/modules/host/negotiate.go
+++ b/modules/host/negotiate.go
@@ -77,9 +77,6 @@ func (h *Host) considerContract(txn types.Transaction, renterKey types.SiaPublic
 	case len(fc.MissedProofOutputs) != 2:
 		return errors.New("bad file contract missed proof outputs")
 
-	case !fc.ValidProofOutputs[1].Value.IsZero(), !fc.MissedProofOutputs[1].Value.IsZero():
-		return errors.New("file contract collateral is not zero")
-
 	case fc.ValidProofOutputs[1].UnlockHash != h.UnlockHash:
 		return errors.New("file contract valid proof output not sent to host")
 	case fc.MissedProofOutputs[1].UnlockHash != voidAddress:
--- a/modules/host/negotiate.go
+++ b/modules/host/negotiate.go
@@ -165,7 +165,11 @@ func (h *Host) considerRevision(txn types.Transaction, obligation contractObliga
 	return nil
 }
 
-func (h *Host) negotiateContract(conn net.Conn, merkleRoot crypto.Hash) error {
+// negotiateContract negotiates an initial file contract with a renter, and
+// adds the metadata to the host's obligation set. The merkleRoot and filename
+// arguments are provided to make negotiateContract usable with both rpcUpload
+// and rpcRenew.
+func (h *Host) negotiateContract(conn net.Conn, merkleRoot crypto.Hash, filename string) error {
 	// allow 5 minutes for contract negotiation
 	conn.SetDeadline(time.Now().Add(5 * time.Minute))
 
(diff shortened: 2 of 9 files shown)


DavidVorick:
While we are getting into the habit of upgrading our comment quality, we need to upgrade the testing quality of our PRs as well, I would like to see some unit and integration testing added to bring the coverage of this PR up to 100%, even if the host as a whole is kept below 100%

modules/renter/hostdb/negotiate.go
```diff
@@ -319,3 +312,70 @@ func newRevision(rev types.FileContractRevision, pieceLen uint64, merkleRoot cry
...
+	}
+	if err := encoding.WriteObject(conn, modules.RPCRenew); err != nil {
+		return types.FileContractID{}, errors.New("couldn't initiate RPC: " + err.Error())
+	}
+	if err := encoding.WriteObject(conn, fcid); err != nil {
+		return types.FileContractID{}, errors.New("couldn't send contract ID: " + err.Error())
+	}
+
+	// execute negotiation protocol
+	// NOTE: negotiateContract is responsible for closing the connection.
```
DavidVorick: seems confusing that you would pass a connection to a function, and then that function would close the connection

modules/renter/repair.go
```diff
@@ -150,6 +151,21 @@ func (f *file) chunkHosts(chunk uint64) []modules.NetAddress {
 	return old
 }
 
+// expiringContracts returns the contracts that will expire soon.
+// TODO: what if contract has fully expired?
```
DavidVorick: Hmm. Right now you can renew it as long as it is in the source location, but if is gone both locally and remotely, you should probably just leave it and let the UI figure out how to report "not repairing" + "0 blocks left"

lukechampine:
Brought back TestIntegrationHosting, which uncovered a few bugs

DavidVorick:
needs rebase

api/host_test.go
```diff
@@ -29,15 +27,6 @@ func TestIntegrationHosting(t *testing.T) {
...
 	}
-	// Announce twice, otherwise the renter will throw a 'not enough hosts'
-	// error.
-	loopAddr := "127.0.0.1:" + st.host.NetAddress().Port()
-	announceValues = url.Values{}
-	announceValues.Set("address", loopAddr)
-	err = st.stdPostAPI("/host/announce", announceValues)
-	if err != nil {
-		t.Fatal(err)
-	}
```
DavidVorick: whew. Glad to see this gone

modules/host/negotiate.go
```diff
@@ -156,7 +156,7 @@ func (h *Host) considerRevision(txn types.Transaction, obligation contractObliga
 		return errors.New("revision outputs do not sum to original payout")
 
 	// outputs should have been adjusted proportional to the new filesize
-	case rev.NewValidProofOutputs[1].Value.Cmp(minHostPrice) <= 0:
+	case rev.NewValidProofOutputs[1].Value.Cmp(minHostPrice) < 0:
```
DavidVorick: Not sure why it was `<=`, good catch

modules/host/negotiate.go
```diff
@@ -165,16 +165,13 @@ func (h *Host) considerRevision(txn types.Transaction, obligation contractObliga
...
-
-	// allow 1 minute for contract negotiation
-	conn.SetDeadline(time.Now().Add(1 * time.Minute))
+// negotiateContract negotiates an initial file contract with a renter, and
+// adds the metadata to the host's obligation set. The merkleRoot and filename
+// arguments are provided to make negotiateContract usable with both rpcUpload
+// and rpcRenew.
+func (h *Host) negotiateContract(conn net.Conn, merkleRoot crypto.Hash, filename string) error {
+	// allow 5 minutes for contract negotiation
+	conn.SetDeadline(time.Now().Add(5 * time.Minute))
```
DavidVorick: I'm wondering if this shouldn't be longer, which is making me think that we should have some way to track how long the average revision takes. long term thought

modules/host/negotiate.go
```diff
@@ -272,6 +267,23 @@ func (h *Host) rpcUpload(conn net.Conn) error {
 	return nil
 }
 
+// rpcUpload is an RPC that negotiates a file contract. Under the new scheme,
+// file contracts should not initially hold any data.
+func (h *Host) rpcUpload(conn net.Conn) error {
+	// Check that the host has grabbed an address from the wallet.
+	if h.UnlockHash == (types.UnlockHash{}) {
+		return errors.New("host needs an address; have you properly announced?")
```
DavidVorick: errors should not contain puncutation

modules/host/negotiate.go
```diff
@@ -409,3 +421,47 @@ func (h *Host) rpcRevise(conn net.Conn) error {
...
+	if err != nil {
+		return err
+	}
+	_, err = io.Copy(renewed, old)
+	if err != nil {
+		return err
+	}
+
+	return h.negotiateContract(conn, merkleRoot, filename)
+}
```
DavidVorick: This function looks good.

modules/renter/hostdb/negotiate.go
```diff
@@ -319,3 +311,70 @@ func newRevision(rev types.FileContractRevision, pieceLen uint64, merkleRoot cry
...
 }
+
+// Renew negotiates a new contract for data already stored with a host. It
+// returns the ID of the new contract. This is a blocking call that performs
+// network I/O.
+func (hdb *HostDB) Renew(fcid types.FileContractID, newEndHeight types.BlockHeight) (types.FileContractID, error) {
+	hdb.mu.RLock()
+	height := hdb.blockHeight
+	hc, ok := hdb.contracts[fcid]
+	host, eok := hdb.allHosts[hc.IP] // or activeHosts?
```
DavidVorick: allHosts is probably what you want.

modules/renter/hostdb/negotiate.go
```diff
@@ -319,3 +311,70 @@ func newRevision(rev types.FileContractRevision, pieceLen uint64, merkleRoot cry
...
+		FileSize:       hc.LastRevision.NewFileSize,
+		FileMerkleRoot: hc.LastRevision.NewFileMerkleRoot,
+		WindowStart:    newEndHeight,
+		WindowEnd:      newEndHeight + host.WindowSize,
+		Payout:         payout,
+		UnlockHash:     types.UnlockHash{}, // to be filled in by negotiateContract
+		// TODO: are these correct?
+		RevisionNumber:     0,
+		ValidProofOutputs:  hc.LastRevision.NewValidProofOutputs,
+		MissedProofOutputs: hc.LastRevision.NewMissedProofOutputs,
```
DavidVorick: RevisionNumber is correct, but you don't want to reuse the outputs from the previous FC otherwise you are going to be double paying for the data.

You want to reset these to what they would be if it were a new file contract, where the host is getting 0 and the renter is getting everything in a refund.

modules/renter/hostdb/negotiate.go
```diff
@@ -319,3 +311,70 @@ func newRevision(rev types.FileContractRevision, pieceLen uint64, merkleRoot cry
...
+	newContract, err := negotiateContract(conn, hc.IP, fc, txnBuilder, hdb.tpool)
+	if err != nil {
+		txnBuilder.Drop() // return unused outputs to wallet
+		return types.FileContractID{}, err
+	}
+
+	// update host contract
+	hdb.mu.Lock()
+	hdb.contracts[newContract.ID] = newContract
+	hdb.save()
```
DavidVorick: there should be some error handling here.

modules/renter/repair.go
DavidVorick: if the contract is already fully expired, then renew should not be attempted, instead it should be picked up by the repair loop.

modules/renter/repair.go
```diff
@@ -150,6 +151,21 @@ func (f *file) chunkHosts(chunk uint64) []modules.NetAddress {
...
+func (f *file) expiringContracts(height types.BlockHeight) []fileContract {
+	f.mu.RLock()
+	defer f.mu.RUnlock()
+
+	var expiring []fileContract
+	for _, fc := range f.contracts {
+		if height > fc.WindowStart-renewThreshold {
```
DavidVorick: does it also make sense to check if the host is online (by looking at a table in the hostdb, not by doing network I/O on-the-fly), and to check whether the contract has fully expired? I'm guessing the logic here is too simple.

modules/renter/repair.go
```diff
@@ -193,7 +209,11 @@ func (r *Renter) threadedRepairFile(name string, meta trackedFile) {
 
 	// create host pool
 	contractSize := f.pieceSize * f.numChunks() // each host gets one piece of each chunk
-	pool, err := r.hostDB.NewPool(contractSize, defaultDuration)
+	var duration types.BlockHeight = defaultDuration
+	if meta.EndHeight != 0 {
+		duration = meta.EndHeight - height
+	}
+	pool, err := r.hostDB.NewPool(contractSize, duration)
```
DavidVorick: `NewPool` does not create the file contracts, right?

modules/renter/upload.go
```diff
@@ -74,6 +70,15 @@ func (r *Renter) Upload(up modules.FileUploadParams) error {
...
 	}
+	// calculate end height
+	// TODO: using 0 as a special value is kind of hacky. Should probably have
+	// an explicit "renew" field.
+	var endHeight types.BlockHeight
+	if up.Duration == 0 {
+		endHeight = 0
+	} else {
+		endHeight = r.cs.Height() + up.Duration
+	}
```
DavidVorick: why do you use '0' as a special value? Why do you need a 'renew' field?

Please resolve this TODO in this PR

modules/renter/repair.go
lukechampine: correct, contracts are only created when `UniqueHosts` is called

modules/renter/upload.go
lukechampine: `duration=0` is used to indicate that the file should be indefinitely renewed. Seems preferable to make the call accept either `duration=n` or `renew=true` (or maybe both?), although this complicates the API somewhat.
I would actually like to do this in a separate PR, since it touches the API. It seems significant enough for it to stand on its own.

modules/renter/repair.go
lukechampine: I figure that will be a separate function, `f.offlineHosts`. Right now the hostdb does not do any uptime checking.

modules/host/negotiate.go
lukechampine: The hostdb actually only allows 30 seconds for contract negotiation, because more than that usually means the connection will time out. But I agree, it would be useful to know the average time required for each RPC.

modules/renter/hostdb/negotiate.go
lukechampine: shouldn't the host be getting the full value, since they're already storing the data?

DavidVorick: oh yeah, not sure what I was thinking.

You'd want the price to cover (filesize)*(duration)*(price), which should match the previous payouts assuming:
 
+ duration is the same
+ price is the same
+ the outputs weren't used as payment channels to pay for downloads

**So, I would definitely not use the outputs in the previous file contract, as it will probably cause future headaches. It should be calculated again using the size, duration, and price.**

I was also thinking that after the first renew, the second renew should be 6 months instead of just 30 days. Maybe 6 months is too long, but right now because we renew with only 2 weeks left, we have two big issues:

1. users who don't run Sia for 15 days straight have a non-zero probability of losing files.
2. users are paying 50% overhead for their renewing (14 days out of 30 are overlapping)

If we jump it to 6 months, or even just 3 months, we can extend the 2 weeks to 4 or 6 weeks and reduce the 50% overhead to something like 15%.

OTOH, users are not super price sensitive, and corporations can be given special setups that reduce the overheads (because a corporation can expect to run Sia 95% of the time, making the two weeks overkill).

Maybe we should leave it for now.

modules/renter/repair.go
DavidVorick: so, if you are calling the `expiringContracts` function, at what stage are the hosts pruned that are either offline or whose contracts have fully expired?

I suppose they could be managed by the hostdb, if the host is offline or the contract has fully expired, the hostdb just skips over the request (maybe returning an error?)

modules/renter/upload.go
DavidVorick: Ah, okay. For a while I was thinking that all files should just be renewed indefinitely, but I can see use cases where you'd want to automatically delete files after a number of days (example: CCTV footage)

Setting `duration=0` is definitely industry-standard for indicating infinity, but I think that it would cause edge cases, so I agree that an alternate approach should be used.

modules/renter/repair.go
lukechampine: Maybe something like:
1. Renew any expiring contracts. If a contract can't be renewed, add it's pieces to the repair set.
2. Check for offline hosts. Add their pieces to the repair set.
3. Check for unuploaded pieces. Add those pieces to the repair set.
4. Repair all the missing pieces.

lukechampine:
Let's get this merged soon. I have more changes in the pipeline

PR #874, opened 2015-12-16
Miner safety

DavidVorick:
The miner previously was not safely handling alterations to the consensus database. Now, if there are problems, instead of refusing to start the miner will correctly perform a rescan. Startup time will be a lot slower, but at least the program will start at all.

I wanted to make this the best PR possible, but unfortunately that resulted in me needing to implement several new features. Test coverage is still not 100% for this PR because I don't think it's the appropriate time or place to start mocking filesystems, but other than dependency mocking to trigger errors, this PR has full coverage.


modules/miner/miner.go
```diff
@@ -101,17 +117,32 @@ func New(cs modules.ConsensusSet, tpool modules.TransactionPool, w modules.Walle
...
+	} else if err != nil {
+		return nil, errors.New("miner subscription failed: " + err.Error())
 	}
-	m.cs.ConsensusSetPersistentSubscribe(m, m.persist.RecentChange)
+
 	m.tpool.TransactionPoolSubscribe(m)
 	return m, nil
 }
 
+// Close terminates all portions of the
```
lukechampine: docstring

modules/miner/update.go
```diff
@@ -7,14 +7,61 @@ import (
 	"github.com/NebulousLabs/Sia/types"
 )
 
+// threadedConsensusRescan resets all of the miner consensus variables and
+// scans the blockchain starting from the genesis block. This function is
+// typically only needed if an error during startup indicates that the miner
+// has desyncrhonized from the consensus set. An error will be sent down the
```
lukechampine: typo

persist/log.go
```diff
@@ -0,0 +1,30 @@
...
+
+// FileLogger is a wrapper for the standard library logger that enforces logging
+// into a file with the Sia-standard settings.
+type FileLogger struct {
+	*log.Logger
+	logFile *os.File
+}
+
+// CreateLogger returns a logger that can be closed.
+func CreateFileLogger(logFilename string) (*FileLogger, error) {
```
lukechampine: maybe just `NewLogger`?

persist/log.go
```diff
@@ -0,0 +1,30 @@
+package persist
+
+import (
+	"log"
+	"os"
+)
+
+// FileLogger is a wrapper for the standard library logger that enforces logging
+// into a file with the Sia-standard settings.
```
lukechampine: The two main advantages of the logger are: 1) it opens files in append mode, and 2) it provides a `Close` method. These aren't really Sia-specific features.

types/constants.go
```diff
@@ -77,7 +77,7 @@ func init() {
 		// environments than a humand can interact with.
 		BlockFrequency = 1 // As fast as possible
 		MaturityDelay = 3
-		GenesisTimestamp = CurrentTimestamp()
+		GenesisTimestamp = CurrentTimestamp() - 1e6
```
lukechampine: ?

lukechampine:
Looks good to me, my only major concern is with the organization of `newHandleErrInvalidConsensusChangeID` and `threadedConsensusRescan`. It seems strange that the latter communicates via a channel, because the former isn't calling it concurrently. I can't offer good advice because I'm not sure what your intention is here. Do you want to run `threadedConsensusRescan` in the background while startup continues? Do you intend it to be used in other contexts eventually?

modules/miner/miner.go
VoidingWarranties: ?

DavidVorick: oh, lol it's not finished

persist/log.go
DavidVorick: hmm. Okay

persist/log.go
DavidVorick: not quite. It also establishes the format that gets used, including choosing to display all timestamps in UTC.

types/constants.go
DavidVorick: where are you seeing the race condition? I'm not seeing anything in the race detector, nor do I think that this causes any type of datarace.

DavidVorick:
I did it to be pedantic. Rescanning in general (for modules) should be a concurrent process, because there's a lot of lock contention and because it could take several minutes. Even though the miner only ever does rescanning at startup, something like the wallet might need to do a live rescan (say, after loading a new address in) and I wanted this to be a good example of what rescanning should look like.

I have made some changes that make the rescanning more pedantic

DavidVorick:
ok, changes complete

lukechampine:
I'm not sure I like the idea of writing code like this just to serve as an example. It means that the code isn't really suited for the task it's trying to accomplish, which is confusing to the reader. I would remove the channel and try to combine `newHandleErrInvalidConsensusChangeID` and `threadedConsensusRescan`, but I might also add a comment like:
```go
// Since rescanning happens only at startup, it's ok for it to block.
// To rescan asynchronously, pass errors down a channel instead of
// returning them from a blocking function.
```
If you're adamant about keeping it I won't protest, but you gotta admit `newHandleErrInvalidConsensusChangeID` is a pretty awful name :P

The anonymous function in `threadedConsensusRescan` also seems unnecessary to me. I've used them in other places, but only to avoid excessive calls to `Close` or `Unlock`.

DavidVorick:
It's less about it being an example and more about it being the exact same for all modules. By making all modules have a more or less uniform structure, overhead for working with different modules is reduced.

The anonymous function I feel is completely appropriate, the locking and unlocking happens in a limited scope within the threaded function, and even we drop the anonymous function I would still want to keep the scoping:

```go
{
    Lock()
    // stuff
    Unlock()
}
```

This is because the locking and unlocking happens right in the middle of the function. The scoping makes it very easy to see where the locking applies, and the anonymous function makes it easy to see that the locking is handled correctly.

I'm particularly pedantic about the locking because we've had so many problems with it in the past. If you forget to call `Unlock` somewhere it can take hours to figure out what exactly is causing your deadlock. Mistakes are very expensive in terms of developer overhead when it comes to concurrency, so it makes sense to be very strict.

lukechampine:
I don't think it's a big issue unless the locking spans a lot of code; three or four lines can be verified with little effort. But if it's your preference I won't complain.
I wish the mutex type would provide an `UnderLock` function to make this dead-simple:
```go
func (m *Mutex) UnderLock(fn func()) {
    m.Lock()
    fn()
    m.Unlock()
}

m.UnderLock(func() {
    // do stuff
})
```
oh well.

DavidVorick:
`newHandleErrInvalidConsensusChangeID` is something I can adjust. It's there to follow the convention of not having more than three lines in your error checking code. I'm not sure the best way to deal with it though.

types/constants.go
lukechampine: me neither, I was just curious why you made this change

lukechampine:
It's a small enough function that I would consider inlining it, although with the channel it's a bit verbose. That's why I was leaning towards removing the channel. Without it, you have:
```go
if err == modules.ErrInvalidConsensusChangeID {
	err = m.consensusRescan()
	if err != nil {
		return nil, errors.New("miner startup failed - rescanning failed: " + err.Error())
	}
}
```

DavidVorick:
okay, I'll remove the threadedness

DavidVorick:
Alright, should be good at this point.

PR #875, opened 2015-12-18
Renter bugfixes

lukechampine:
Commit messages are self-explanatory

### DIFF
--- a/api/host_test.go
+++ b/api/host_test.go
@@ -15,6 +15,10 @@ import (
 // TestIntegrationHosting tests that the host correctly receives payment for
 // hosting files.
 func TestIntegrationHosting(t *testing.T) {
+	if testing.Short() {
+		t.SkipNow()
+	}
+
 	st, err := createServerTester("TestIntegrationHosting")
 	if err != nil {
 		t.Fatal(err)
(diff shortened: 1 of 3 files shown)


api/host_test.go
```diff
@@ -70,7 +75,7 @@ func TestIntegrationHosting(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	expRevenue := "31844166666666464210"
+	expRevenue := "31857777777777575235"
```
DavidVorick: why does this number keep changing?

modules/renter/hostdb/upload.go
```diff
@@ -85,6 +85,7 @@ func (hu *hostUploader) Upload(data []byte) (uint64, error) {
 		return 0, errors.New("contract has already ended")
 	}
 	piecePrice := types.NewCurrency64(uint64(len(data))).Mul(types.NewCurrency64(uint64(hu.contract.FileContract.WindowStart - height))).Mul(hu.price)
+	piecePrice = piecePrice.MulFloat(1.02) // COMPATv0.4.8 -- hosts reject exact prices
```
DavidVorick: oh

modules/renter/repair.go
```diff
@@ -49,9 +50,13 @@ func (f *file) repair(chunkIndex uint64, missingPieces []uint64, r io.ReaderAt,
 	var wg sync.WaitGroup
 	wg.Add(numPieces)
 	for i := 0; i < numPieces; i++ {
-		go func(host hostdb.Uploader, pieceIndex uint64, piece []byte) {
+		go func(i int) {
```
DavidVorick: do you need to pass in 'i' as an arg because this is a concurrent function? I guess so. But does that mean that the rest of the stuff is safe?

I guess it looks good to me, just want you to double check the correctness here.

PR #876, opened 2015-12-18
Add go vet to travis

lukechampine:


DavidVorick:
Need to add vet to the makefile. Perhaps as the last step in `make test-long`?

```
test-long: clean fmt REBUILD
	go test -v -race -tags='testing debug' -timeout=300s $(pkgs) -run=$(run)
        go vet $(pkgs)
```

lukechampine:
done

DavidVorick:
no, you still need it because the build won't fail

DavidVorick:
oh yeah, you can't just throw a `go vet` in travis, you have to wrap it in the `test` call

.travis.yml
```diff
@@ -6,8 +6,9 @@ go:
 install:
   - make dependencies
   - test -z "$(go fmt ./...)"
+  - go vet ./...
```
DavidVorick: `  - test -z "$(go vet ./...)"`

Makefile
```diff
@@ -72,7 +76,7 @@ test: REBUILD
 	go test -short -tags='debug testing' -timeout=3s $(pkgs) -run=$(run)
 test-v: REBUILD
 	go test -race -v -short -tags='debug testing' -timeout=15s $(pkgs) -run=$(run)
-test-long: clean fmt REBUILD
+test-long: clean fmt vet REBUILD
 	go test -v -race -tags='testing debug' -timeout=300s $(pkgs) -run=$(run)
```
DavidVorick: I would prefer if vet was placed at the bottom, otherwise there's not going to be a FAIL indicating that you missed something. If it's at the bottom, you can see the output. Maybe we should do the same for fmt so you can see what files got updated. Right now you don't ever see that because it's buried behind the output of 500 tests.

DavidVorick:
Discussed IRL: go vet actually returns an error code if it detects anything, so it'll fail without needing to run the `test -z $(go vet ./...)` line. So it's been removed from setup and just added to a part of `make test-long`

PR #877, opened 2015-12-18
refactor mocking in the crypto package

DavidVorick:
This broke compatibility in a few places, hence the large number of
files changed.

### DIFF
--- a/crypto/signatures.go
+++ b/crypto/signatures.go
@@ -1,78 +1,55 @@
 package crypto
 
 import (
-	"crypto/rand"
 	"errors"
-	"io"
 
 	"github.com/NebulousLabs/ed25519"
 )
 
 const (
-	EntropySize   = ed25519.EntropySize
+	// EntropySize defines the amount of entropy necessary to do secure
+	// cryptographic operations, in bytes.
+	EntropySize = ed25519.EntropySize
+
+	// PublicKeySize defines the size of public keys in bytes.
 	PublicKeySize = ed25519.PublicKeySize
+
+	// SecretKeySize defines the size of secret keys in bytes.
 	SecretKeySize = ed25519.SecretKeySize
+
+	// SignatureSize defines the size of signatures in bytes.
 	SignatureSize = ed25519.SignatureSize
 )
 
 type (
+	// PublicKey is an object that can be used to verify signatures.
 	PublicKey [ed25519.PublicKeySize]byte
+
+	// SecretKey can be used to sign data for the corresponding public key.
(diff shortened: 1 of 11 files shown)


lukechampine:
`Generate` feels a little ambiguous now that it's not a method on `StdKeyGen`; `crypto.GenerateKeyPair` would be more readable.

Other than that, my only thought is whether `generate` and `generateDeterministic` still need to be methods (rather than taking dependencies as arguments). The `sigKeyGen` type does not seem to add anything of value, since it has no state and does not satisfy an interface.

Also, as you mentioned, it may not be necessary to mock the `ed25519.GenerateKey` function. In that case, we could remove its corresponding interface and the `stdKeyDeriver` type, and just define
```go
func GenerateDeterministicKeyPair(entropy [EntropySize]byte) (SecretKey, PublicKey) {
	skPointer, pkPointer := ed25519.GenerateKey(entropy)
	return *skPointer, *pkPointer
}
```
which would be unit tested separately.

DavidVorick:
fixed up

lukechampine:
What are your thoughts on the above?

DavidVorick:
oh. read it too quickly.

The functions would take dependencies as arguments? And those dependencies would be in the form of interfaces? I still prefer using the struct I think, because the list of dependencies could get unfriendly.

Also, I am okay with leaving ed25519 as a formal dependency. It does make the test run faster, and it does add some purity, and it does make the dependencies of the function explicit.

lukechampine:
It only seems friendly here because you "cheat" a bit by defining `stdKeyGen` globally. Otherwise you would have to define it inside both of the `Generate` functions:
```go
func GenerateKeyPair() (SecretKey, PublicKey, error) {
    g := sigKeyGen{entropySource: rand.Reader, keyDeriver: &stdKeyDeriver{}}
    return g.generate()
}
```
As opposed to:
```go
func GenerateKeyPair() (SecretKey, PublicKey, error) {
    return generate(rand.Reader, &stdKeyDeriver{})
}
```
That `&stdKeyDeriver{}` is still kinda ugly...lol, now I'm flip-flopping too, and wondering if using a function instead of an interface isn't such a bad idea after all. Whatevs, we can continue to refine this as we start mocking more things.

DavidVorick:
huh. I hadn't thought of creating the `sigKeyGen` inside of the exported function.

I'm still wary of the closure, we should probably stick with interfaces.

Let's mock more things before we do anything else with the crypto package. Having some OS mocking would be good experience.

PR #878, opened 2015-12-18
New Upload API semantics

lukechampine:
The upload API now behaves like so:
- if `renew == true`, the file will be renewed indefinitely.
- if `renew != true && duration != 0`, the file will be uploaded for the
  duration specified.
- if `renew != true && duration = 0`, the file will be uploaded for the
  default duration.

Basically, instead of using a magic value to indicate renewal (`duration == 0`), we use an explicit separate field.

### DIFF
--- a/api/server.go
+++ b/api/server.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net"
 	"strings"
+	"time"
 
 	"github.com/stretchr/graceful"
 
@@ -77,5 +78,11 @@ func (srv *Server) Serve() error {
 	}
 
 	fmt.Println("\rCaught stop signal, quitting.")
+	return nil
+}
+
+func (srv *Server) Close() error {
+	// give graceful 1 second to shutdown
+	srv.apiServer.Stop(time.Second)
 	return nil
 }
--- a/api/server.go
+++ b/api/server.go
@@ -1,7 +1,6 @@
 package api
 
 import (
-	"fmt"
 	"net"
 	"strings"
 	"time"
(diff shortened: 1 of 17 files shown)


DavidVorick:
You aren't providing enough testing with your PRs.

lukechampine:
true. Perhaps it's time to buff up the renter's test coverage.

DavidVorick:
Not merging this until there's test coverage for the new behavior

lukechampine:
good call -- found a bug. Can you weigh in on the commit message for e9490c7?

Renewing is tested now, but in general the host/renter coverage is still unacceptably low. I'll be working to improve that this week.

api/server.go
```diff
@@ -76,6 +76,11 @@ func (srv *Server) Serve() error {
 		srv.wallet.Lock()
 	}
 
-	fmt.Println("\rCaught stop signal, quitting.")
+	return nil
+}
+
+func (srv *Server) Close() error {
```
DavidVorick: docstring

modules/renter/repair.go
```diff
@@ -13,13 +14,22 @@ import (
...
+var renewThreshold = func() types.BlockHeight {
+	switch build.Release {
+	case "testing":
+		return 20
+	case "dev":
+		return 200
+	default:
+		return 2000
+	}
+}()
```
DavidVorick: This is an odd construction

DavidVorick: I guess it's strange that you can have a variable that's a function which gets called.

PR #880, opened 2015-12-21
Fix padding bug in MerkleTree.ReadSegments

VoidingWarranties:
Two things were wrong with the method.
1) The method should not have been padding data.
2) The function claimed to be padding with 0s but unless there was only one leaf it was actually padding data with the end of the previous leaf.

The fix was to make the corresponding unexported function in the merkletree repo an exported method. See PR NebulousLabs/merkletree#10. MerkleTree.ReadSegments now wraps this method instead of reimplementing it.

### DIFF
--- a/crypto/merkle.go
+++ b/crypto/merkle.go
@@ -33,11 +33,7 @@ func (t MerkleTree) PushObject(obj interface{}) {
 // to the tree.  No error is returned unless err != io.EOF && err !=
 // io.errUnexpectedEOF
 func (t MerkleTree) ReadSegments(r io.Reader) error {
-	err := t.ReadAll(r, SegmentSize)
-	if err != nil {
-		return err
-	}
-	return nil
+	return t.ReadAll(r, SegmentSize)
 }
 
 // Root returns the Merkle root of all the objects pushed to the tree.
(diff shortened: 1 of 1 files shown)


lukechampine:
Since the `merkleTree` type embeds a `merkletree.Tree`, you can call any methods of the latter on the former. So it's probably best to remove the `ReadSegments` method and simply rename instances of `ReadSegments` to `ReadAll`.

VoidingWarranties:
@DavidVorick mentioned yesterday that the reason for wrapping `merkletree.Tree` in `MerkleTree` is to create an interface that is independent of the `merkletree.Tree` interface so that a different merkletree implementation can be swapped out without requiring changes around the codebase. I don't necessarily agree with this as `ReadSegments` is used in only two other places. Furthermore we already break this rule in consistency.go where we call `merkletree.Tree.Push`. Thoughts?

lukechampine:
I suppose that's pragmatic, although at this point the merkle tree code is very stable and unlikely to be radically changed anytime soon. A better reason is that `ReadAll` takes a segment size, and `ReadSegments` does not -- it abstracts away that detail.
At any rate, you can shorten the implementation to simply `return t.ReadAll(r, SegmentSize)`

VoidingWarranties:
D'oh. Fixed.

PR #881, opened 2015-12-21
Add tests to crypto package

VoidingWarranties:
Brings tests coverage for the crypto package from 74.2% to 91.6%. No mocking was used in the tests added. The remaining tests necessary to bring the package up to 100% coverage will require mocking.

**NOTE: TestPartialLastLeafReadSegments will fail (as it should) until PR #880 is merged.**

Tests added:
- discard_test.go
  - TestUnitSecureWipe
  - TestUnitSecureWipeEdgeCases
- encrypt_test.go
  - TestUnitCiphertextUnmarshalJSON
  - TestCiphertextMarshalling
- hash_test.go
  - TestUnitHashMarshalJSON
  - TestUnitHashUnmarshalJSON
- merkle_test.go
  - TestReadSegments
  - TestPartialLastLeafReadSegments
- rand_test.go
  - TestRandIntnPanics

Besides tests, `Hash.String` and `Hash.UnmarshalJSON` were modified to encode / decode hashes to / from strings using `hex.EncodeToString` and `hex.DecodeString`, respectively. The previous implementation used `fmt.Printf` and `fmt.Scanf`. Unlike the fmt package, the hex package does not use reflection. The intent of the code is also more clear with the hex package.

### DIFF
--- a/crypto/encrypt_test.go
+++ b/crypto/encrypt_test.go
@@ -7,17 +7,6 @@ import (
 	"testing"
 )
 
-var (
-	ciphertextMarshallingTests = []struct {
-		ct        Ciphertext
-		jsonBytes []byte
-	}{
-		{ct: Ciphertext(nil), jsonBytes: []byte("null")},
-		{ct: Ciphertext(""), jsonBytes: []byte(`""`)},
-		{ct: Ciphertext("a ciphertext"), jsonBytes: []byte(`"YSBjaXBoZXJ0ZXh0"`) /* base64 encoding of the Ciphertext */},
-	}
-)
-
 // TestTwofishEncryption checks that encryption and decryption works correctly.
 func TestTwofishEncryption(t *testing.T) {
 	// Get a key for encryption.
@@ -153,27 +142,8 @@ func TestTwofishEntropy(t *testing.T) {
 	}
 }
 
-// TestUnitCiphertextMarshalJSON tests that Ciphertext.MarshalJSON never fails,
-// because json.Marshal should nevef fail to encode a byte slice.
-func TestUnitCiphertextMarshalJSON(t *testing.T) {
-	for _, test := range ciphertextMarshallingTests {
-		ct := test.ct
-		expectedJSONBytes := test.jsonBytes
-
-		jsonBytes, err := ct.MarshalJSON()
-		if err != nil {
-			t.Fatal(err)
-		}
-		if !bytes.Equal(jsonBytes, expectedJSONBytes) {
-			t.Errorf("cipher text %#v encoded incorrectly: expected %q, got %q\n", ct, expectedJSONBytes, jsonBytes)
-		}
-	}
-}
-
-// TestUnitCiphertextUnmarshalJSON tests that Ciphertext.UnmarshalJSON correctly
-// fails on invalid JSON marshalled Ciphertext and doesn't fail on valid JSON
-// marshalled Ciphertext. Also tests that valid JSON marshalled Ciphertext
-// decodes to the correct JSON.
+// TestUnitCiphertextUnmarshalJSON tests that Ciphertext.UnmarshalJSON
+// correctly fails on invalid JSON marshalled Ciphertext.
 func TestUnitCiphertextUnmarshalJSON(t *testing.T) {
 	// Test unmarshalling invalid JSON.
 	invalidJSONBytes := [][]byte{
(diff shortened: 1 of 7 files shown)


crypto/encrypt_test.go
```diff
@@ -141,3 +141,69 @@ func TestTwofishEntropy(t *testing.T) {
...
+			copy(ct, expectedCt)
+		}
+
+		// Marshal ct to JSON.
+		jsonBytes, err := ct.MarshalJSON()
+		if err != nil {
+			t.Fatal(err)
+		}
+		if !bytes.Equal(jsonBytes, expectedJSONBytes) {
+			t.Fatalf("Ciphertext %#v marshalled incorrectly: expected %q, got %q\n", ct, expectedJSONBytes, jsonBytes)
```
lukechampine: what is the purpose of using `%#v` here (rather than `%v`)?

crypto/hash.go
```diff
@@ -74,19 +74,21 @@ func (h Hash) MarshalJSON() ([]byte, error) {
...
+	// *2 because there are 2 hex characters per byte.
+	// +2 because the encoded JSON string has a `"` added at the beginning and end.
 	if len(b) != HashSize*2+2 {
 		return ErrHashWrongLen
 	}
 
-	var hBytes []byte
-	_, err := fmt.Sscanf(string(b[1:len(b)-1]), "%x", &hBytes)
+	// b[1 : len(b)-1] cuts off the leading and trailing `"` in the JSON string.
+	hBytes, err := hex.DecodeString(string(b[1 : len(b)-1]))
```
lukechampine: :+1:

crypto/encrypt_test.go
VoidingWarranties: If %v is used then `Ciphertext(nil)` and `Ciphertext([]byte{})` will be printed identically as "Ciphertext [] ". Using %#v prints "crypto.Ciphertext(nil)" and "crypto.Ciphertext{}", respectively.

crypto/encrypt_test.go
```diff
@@ -141,3 +141,69 @@ func TestTwofishEntropy(t *testing.T) {
 		t.Error("supposedly high entropy ciphertext has been compressed!")
 	}
 }
+
+// TestUnitCiphertextUnmarshalJSON tests that Ciphertext.UnmarshalJSON
+// correctly fails on invalid JSON marshalled Ciphertext.
+func TestUnitCiphertextUnmarshalJSON(t *testing.T) {
```
DavidVorick: name is not complete, because this test is only about invalid json.

It's okay for the test function names to be long, it's better to be clear about what is being tested.

crypto/hash.go
```diff
@@ -74,19 +74,21 @@ func (h Hash) MarshalJSON() ([]byte, error) {
 
 // String prints the hash in hex.
 func (h Hash) String() string {
-	return fmt.Sprintf("%x", h[:])
+	return hex.EncodeToString(h[:])
```
DavidVorick: huh. cool

crypto/hash.go
```diff
@@ -74,19 +74,21 @@ func (h Hash) MarshalJSON() ([]byte, error) {
...
 		return ErrHashWrongLen
 	}
 
-	var hBytes []byte
-	_, err := fmt.Sscanf(string(b[1:len(b)-1]), "%x", &hBytes)
+	// b[1 : len(b)-1] cuts off the leading and trailing `"` in the JSON string.
+	hBytes, err := hex.DecodeString(string(b[1 : len(b)-1]))
 	if err != nil {
-		return errors.New("could not unmarshal crypto.Hash: " + err.Error())
+		return err
```
DavidVorick: don't remove the context on the error. For example, sometimes Sia will be running and suddenly you'll get a message like "EOF", and you're not sure if it's a file EOF, if it's a network EOF, a json thing, and you have no idea what package the message came from.

crypto/hash_test.go
```diff
@@ -1,22 +1,26 @@
...
+type (
+	// TestObject is a struct that's used for testing HashAll and HashObject. The
+	// fields have to be exported so the encoder can read them.
+	TestObject struct {
+		A int
+		B byte
+		C bool
+		D string
+	}
+)
```
DavidVorick: no need to wrap a single struct in a `type ()` identifier, but... at least it's consistent.

crypto/merkle_test.go
```diff
@@ -95,3 +95,63 @@ func TestPaddedStorageProof(t *testing.T) {
...
 	}
 }
+
+// TestReadSegments tests MerkleTree.ReadSegments by building a MerkleTree and
+// comparing its root with the root returned from ReaderMerkleRoot.
+func TestReadSegments(t *testing.T) {
+	// Generate leaf data.
+	numSegments := 7
+	data := make([]byte, numSegments*SegmentSize)
+	rand.Read(data)
```
DavidVorick: You should definitely add a test case where the data is not a multiple of `SegmentSize`.

crypto/rand_test.go
```diff
@@ -0,0 +1,18 @@
+package crypto
+
+import (
+	"testing"
+)
+
+// TestRandIntnPanics tests that RandIntn panics if n <= 0. The crypto/rand
+// package guarantees that rand.Int will panic if n <= 0, but other random
+// packages might not.
```
DavidVorick: No need to write a justification comment (the second sentence). If something is supposed to have a certain behavior, that's all the reason you need to test that the behavior actually exists.

crypto/merkle_test.go
VoidingWarranties: See `TestNonMultipleNumberOfSegmentsReadSegments`

crypto/hash.go
VoidingWarranties: Why do we add context to errors so sparingly? This is one of the few places I've seen that adds a context to the error. Should we be adding context to errors more often?

PR #882, opened 2015-12-21
Brief Host Changes

DavidVorick:
Commit messages explain most of what happened. Some basic cleanup before beginning to do some more serious host work.

### DIFF
--- a/modules/consensus/accept_test.go
+++ b/modules/consensus/accept_test.go
@@ -504,7 +504,17 @@ func TestFutureTimestampHandling(t *testing.T) {
 	time.Sleep(time.Second * 3) // 3 seconds, as the block was originally 2 seconds too far into the future.
 	_, err = cst.cs.dbGetBlockMap(solvedBlock.ID())
 	if err == errNilItem {
-		t.Fatalf("future block was not added to the consensus set after waiting the appropriate amount of time")
+		t.Errorf("future block was not added to the consensus set after waiting the appropriate amount of time")
+	}
+	time.Sleep(time.Second * 3)
+	_, err = cst.cs.dbGetBlockMap(solvedBlock.ID())
+	if err == errNilItem {
+		t.Errorf("Future block not added to consensus set.\nCurrent Timestamp %v\nFutureThreshold: %v\nBlock Timestamp %v\n", types.CurrentTimestamp(), types.FutureThreshold, block.Timestamp)
+		time.Sleep(time.Second * 3)
+		_, err = cst.cs.dbGetBlockMap(solvedBlock.ID())
+		if err == errNilItem {
+			t.Error("waiting double long did not help.")
+		}
 	}
 }
 
--- a/modules/host/update.go
+++ b/modules/host/update.go
@@ -12,6 +12,7 @@ import (
 func (h *Host) threadedDeleteObligation(obligation *contractObligation) {
 	h.mu.Lock()
 	defer h.mu.Unlock()
+
 	err := h.deallocate(obligation.Path)
 	if err != nil {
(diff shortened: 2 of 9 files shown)


modules/host/update.go
```diff
@@ -74,17 +75,29 @@ func (h *Host) ProcessConsensusChange(cc modules.ConsensusChange) {
...
 	defer h.mu.Unlock()
 
-	h.blockHeight += types.BlockHeight(len(cc.AppliedBlocks))
-	h.blockHeight -= types.BlockHeight(len(cc.RevertedBlocks))
+	// Adjust the height of the host. The host height is initialized to zero,
+	// but the genesis block is actually height zero. For the genesis block
+	// only, the height will be left at zero.
+	//
+	// Checking the height here eliminates the need to inialize the host to and
+	// underflowed types.BlockHeight.
```
lukechampine: typo

DavidVorick:
I'm not sure what I changed that's causing the integration test to fail, but the renter is no longer able to upload files.

@lukechampine ideas?

DavidVorick:
I've confirmed that it's my corrections to the way the host is calculating the height, but I'm not sure why that's significant.

DavidVorick:
The build failed a few times due to timestamping problems with consensus. I wasn't able to get to the root of the problem because by the time I had added debugging statements, the problems had stopped. Possibly a buggy clock on the Travis side, it wasn't code should be failing nondeterministically.

I've added some extra debugging info

modules/consensus/accept_test.go
```diff
@@ -504,7 +504,20 @@ func TestFutureTimestampHandling(t *testing.T) {
...
-		t.Fatalf("future block was not added to the consensus set after waiting the appropriate amount of time")
+		t.Errorf("future block was not added to the consensus set after waiting the appropriate amount of time")
+	}
+	time.Sleep(time.Second * 3)
+	_, err = cst.cs.dbGetBlockMap(solvedBlock.ID())
+	if err == errNilItem {
+		t.Error("future block was not added to the consensus set after waiting the appropriate amount of time.")
+		t.Error(types.CurrentTimestamp())
+		t.Error(types.FutureThreshold)
+		t.Error(block.Timestamp)
```
lukechampine: I prefer to combine this sort of thing into a single `Errorf`

modules/host/update.go
```diff
@@ -74,17 +75,29 @@ func (h *Host) ProcessConsensusChange(cc modules.ConsensusChange) {
...
 	defer h.mu.Unlock()
 
-	h.blockHeight += types.BlockHeight(len(cc.AppliedBlocks))
-	h.blockHeight -= types.BlockHeight(len(cc.RevertedBlocks))
+	// Adjust the height of the host. The host height is initialized to zero,
+	// but the genesis block is actually height zero. For the genesis block
+	// only, the height will be left at zero.
+	//
+	// Checking the height here eliminates the need to initialize the host to
+	// and underflowed types.BlockHeight.
```
lukechampine: this is still done in other modules, right?
kinda ugly how this is handled. It would be nice if we could come up with something cleaner, perhaps by altering the semantics of subscription to treat the genesis block specially.

DavidVorick: I think it's worse to treat the genesis block special. I think it'd be better if genesis block were height 1.

It's nice to have the genesis block get provided by a subscription because it's got diffs, otherwise you'd have to call ProcessGenesisBlock() or something, and that's clunky too. I still prefer the way it's set up here, though you are right it will end up appearing in every module.

DavidVorick:
updated

Issue #883, opened 2015-12-22
Ignoring errors with the blank identifier


VoidingWarranties:
Contrary to the developers conventions, there a quite a few places where errors are ignored using the blank identifier (use `errcheck -blank` to find them). In some places these are accompanied by an explanation, e.g. [crypto/encrypt.go](https://github.com/NebulousLabs/Sia/blob/f3272113464e8f4b71383c71978aed400b4dfd89/crypto/encrypt.go#L38-L39)
```go
// NOTE: NewCipher only returns an error if len(key) != 16, 24, or 32.
cipher, _ := twofish.NewCipher(key[:])
```
In others there is no explanation.

Not checking errors if we can safely assume none will occur is not the best but with a valid explanation it seems okay. But for every error ignored (that has an explanation) there should be a corresponding test to make sure that the assumption is true. In this case the test should verify that `len(key)` is 16, 24, or 32.

VoidingWarranties:
Then again it would be simpler and more consistent if errors are never ignored.

VoidingWarranties:
But then `TwofishKey.NewReader` and `TwofishKey.NewWriter` would have to return the errors too. So testing each assumption used to explain why an error is ignored seems like the better solution.

DavidVorick:
Yeah we've been back and forth on this a few times. If your function is already returning an error, then there's almost no excuse, with the exception of defer statements which can be annoying.

`defer obj.Close()` -> can return an error, but how do you check an error inside of a defer statement?

And, if we start enforcing that everything needs to be returning errors, that's pretty annoying especially when the behavior is clear that errors will never be returned. Then again, always checking anyway helps catch developer mistakes.

I think I'd rather have the functions start to return errors. Idk how much headache that will end up being though.

lukechampine:
IIRC the `NewReader` and `NewWriter` functions aren't used anywhere anymore. Originally they were used for encrypting streams (during upload/download), but now we break things into small in-memory chunks anyway so we use `EncryptBytes` instead.

[...]

PR #884, opened 2015-12-22
Crypto errcheck

VoidingWarranties:
Used `errcheck -blank github.com/NebulousLabs/Sia/crypto/...` to find places where errors weren't being checked, and checked them.

The remaining unchecked errors are listed below. Other than the first three, all of the remaining unchecked errors are in tests and are calls to standard library packages that we can assume return `nil` on a working system, so they can be ignored.

The first three unchecked errors below assume that an error can not occur. See issue #883. For these cases tests were added to verify these assumptions.

```
/Users/jordan/Projects/Sia/crypto/encrypt.go:40:10	cipher, _ := twofish.NewCipher(key[:])
/Users/jordan/Projects/Sia/crypto/encrypt.go:50:8	aead, _ := cipher.NewGCM(key.NewCipher())
/Users/jordan/Projects/Sia/crypto/encrypt.go:69:8	aead, _ := cipher.NewGCM(key.NewCipher())
/Users/jordan/Projects/Sia/crypto/encrypt_test.go:101:26	key.NewWriter(buf).Write(plaintext)
/Users/jordan/Projects/Sia/crypto/encrypt_test.go:110:25	key.NewReader(buf).Read(decrypted)
/Users/jordan/Projects/Sia/crypto/encrypt_test.go:145:11	zip.Close()
/Users/jordan/Projects/Sia/crypto/hash_test.go:54:11	rand.Read(data)
/Users/jordan/Projects/Sia/crypto/merkle_test.go:50:11	rand.Read(data)
/Users/jordan/Projects/Sia/crypto/merkle_test.go:83:11	rand.Read(data)
/Users/jordan/Projects/Sia/crypto/merkle_test.go:105:11	rand.Read(data)
/Users/jordan/Projects/Sia/crypto/merkle_test.go:136:11	rand.Read(data)
/Users/jordan/Projects/Sia/crypto/signatures_test.go:156:11	rand.Read(signedData[:])
/Users/jordan/Projects/Sia/crypto/signatures_test.go:197:12	rand.Read(randData[:])
```

### DIFF
--- a/crypto/encrypt.go
+++ b/crypto/encrypt.go
@@ -36,6 +36,7 @@ func GenerateTwofishKey() (key TwofishKey, err error) {
 // NewCipher creates a new Twofish cipher from the key.
 func (key TwofishKey) NewCipher() cipher.Block {
 	// NOTE: NewCipher only returns an error if len(key) != 16, 24, or 32.
+	//       Assumption tested in TestTwofishNewCipherAssumption.
 	cipher, _ := twofish.NewCipher(key[:])
 	return cipher
 }
@@ -45,6 +46,7 @@ func (key TwofishKey) NewCipher() cipher.Block {
 func (key TwofishKey) EncryptBytes(plaintext []byte) (Ciphertext, error) {
 	// Create the cipher.
 	// NOTE: NewGCM only returns an error if twofishCipher.BlockSize != 16.
+	//       Assumption tested in TestCipherNewGCMAssumption.
 	aead, _ := cipher.NewGCM(key.NewCipher())
 
 	// Create the nonce.
@@ -63,6 +65,7 @@ func (key TwofishKey) EncryptBytes(plaintext []byte) (Ciphertext, error) {
 func (key TwofishKey) DecryptBytes(ct Ciphertext) ([]byte, error) {
 	// Create the cipher.
 	// NOTE: NewGCM only returns an error if twofishCipher.BlockSize != 16.
+	//       Assumption tested in TestCipherNewGCMAssumption.
 	aead, _ := cipher.NewGCM(key.NewCipher())
 
 	// Check for a nonce.
--- a/crypto/encrypt.go
+++ b/crypto/encrypt.go
@@ -36,7 +36,6 @@ func GenerateTwofishKey() (key TwofishKey, err error) {
 // NewCipher creates a new Twofish cipher from the key.
 func (key TwofishKey) NewCipher() cipher.Block {
 	// NOTE: NewCipher only returns an error if len(key) != 16, 24, or 32.
-	//       Assumption tested in TestTwofishNewCipherAssumption.
 	cipher, _ := twofish.NewCipher(key[:])
 	return cipher
 }
@@ -46,7 +45,6 @@ func (key TwofishKey) NewCipher() cipher.Block {
 func (key TwofishKey) EncryptBytes(plaintext []byte) (Ciphertext, error) {
 	// Create the cipher.
 	// NOTE: NewGCM only returns an error if twofishCipher.BlockSize != 16.
-	//       Assumption tested in TestCipherNewGCMAssumption.
 	aead, _ := cipher.NewGCM(key.NewCipher())
 
 	// Create the nonce.
@@ -65,7 +63,6 @@ func (key TwofishKey) EncryptBytes(plaintext []byte) (Ciphertext, error) {
 func (key TwofishKey) DecryptBytes(ct Ciphertext) ([]byte, error) {
 	// Create the cipher.
 	// NOTE: NewGCM only returns an error if twofishCipher.BlockSize != 16.
-	//       Assumption tested in TestCipherNewGCMAssumption.
 	aead, _ := cipher.NewGCM(key.NewCipher())
 
 	// Check for a nonce.
--- a/crypto/encrypt_test.go
+++ b/crypto/encrypt_test.go
@@ -229,9 +229,9 @@ func TestTwofishNewCipherAssumption(t *testing.T) {
 		t.Fatal(err)
 	}
 	// Test key length.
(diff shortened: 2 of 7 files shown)


crypto/encrypt.go
```diff
@@ -36,6 +36,7 @@ func GenerateTwofishKey() (key TwofishKey, err error) {
 // NewCipher creates a new Twofish cipher from the key.
 func (key TwofishKey) NewCipher() cipher.Block {
 	// NOTE: NewCipher only returns an error if len(key) != 16, 24, or 32.
+	//       Assumption tested in TestTwofishNewCipherAssumption.
```
DavidVorick: this comment doesn't add anything, and also if you are treating your dependencies as a black box you can't be certain that there aren't other reasons for an error to be returned (such as 'NoEntropy' or 'CryptoDoesn'tWorkOnTuesdays').

crypto/encrypt_test.go
```diff
@@ -212,3 +218,31 @@ func TestCiphertextMarshalling(t *testing.T) {
 		}
 	}
 }
+
+func TestTwofishNewCipherAssumption(t *testing.T) {
```
DavidVorick: need a docstring on every function, including tests.

crypto/encrypt.go
DavidVorick: oh I think I misuderstood what you meant by 'tested the assumption'.

I'm guessing it's okay to not check the error here, since encryption is just math. Comment is still not needed.

crypto/rand_test.go
```diff
@@ -11,6 +11,14 @@ func TestRandIntnPanics(t *testing.T) {
...
+	_, err := RandIntn(0)
+	if err != nil {
+		t.Error("expected panic on n <= 0, not error")
+	}
+
+	_, err = RandIntn(-1)
+	if err != nil {
+		t.Error("expected panic on n <= 0, not error")
+	}
 }
```
DavidVorick: This test I believe is wrong, the RandIntn(-1) code is never reached, because the recover triggers instead, which returns the function.

In the recover, you need to call RandIntn(-1) and then you need to above the existing recover set up another recover:

```
defer {
    recover
}
defer {
    recover
    rand -1
}
rand 0
```

DavidVorick:
In general looks good, we're going to have to keep working on the best way to use errcheck going forward.

crypto/rand_test.go
lukechampine: yeah, `panic` semantics make this slightly annoying to test. You could use a helper function:
```go
func panics(fn func()) (panicked bool) {
    defer func() {
        panicked = (recover() != nil)
    }()
    fn()
    return
}
```
and then test with:
```go
if !panics(func() { RandIntn(0) }) {
    t.Error("RandIntn did not panic")
}
```

crypto/encrypt_test.go
```diff
@@ -212,3 +218,31 @@ func TestCiphertextMarshalling(t *testing.T) {
...
+	// Generate key.
+	key, err := GenerateTwofishKey()
+	if err != nil {
+		t.Fatal(err)
+	}
+	// Test that len(key) is 16, 24, or 32 as these are the only cases
+	// where twofish.NewCipher(key[:]) doesn't return an error.
+	len := len(key)
```
lukechampine: try to avoid using builtin functions as variable names. This would actually prevent you from calling `len` later on. `keyLen` is a better name

crypto/rand.go
```diff
@@ -21,12 +21,16 @@ func RandIntn(n int) (int, error) {
 
 // Perm returns, as a slice of n ints, a random permutation of the integers
 // [0,n).
-func Perm(n int) []int {
+func Perm(n int) ([]int, error) {
```
lukechampine: this makes me sad but I guess it's best practice

VoidingWarranties: I didn't like it at first either, but then I saw that the other rand functions also return errors.

VoidingWarranties:
Fixed. It would be nice if to add errcheck to Travis but of course it would need to ignore cases where there is a note explaining why the error is ignored.

crypto/rand_test.go
```diff
@@ -4,13 +4,26 @@ import (
...
 // TestRandIntnPanics tests that RandIntn panics if n <= 0.
 func TestRandIntnPanics(t *testing.T) {
-	defer func() {
-		if r := recover(); r == nil {
+	testNs := []int{0, -1}
+	for _, n := range testNs {
+		if !panics(func() {
+			_, err := RandIntn(n)
+			if err != nil {
+				t.Error("expected panic on n <= 0, not error")
```
lukechampine: this error is kinda redundant: if it's triggered, then the `t.Error` below will be triggered as well.

VoidingWarranties: Good catch. Fixed.

crypto/rand_test.go
DavidVorick: that could potentially go in the build package, looks like it would work to me, makes checking panics linear instead of a stack that's coded backwards.

crypto/rand_test.go
```diff
@@ -4,13 +4,24 @@ import (
...
-		if r := recover(); r == nil {
-			t.Error("expected panic for n <= 0")
-		}
+		panicked = (recover() != nil)
 	}()
-	RandIntn(0)
-	RandIntn(-1)
+	fn()
+	return
+}
```
DavidVorick: This is starting to go outside of the scope for this PR, but next PR can you move this function to the build package? `build.Panics`

Also, instead of returning an error you should return the panic message. `nil` means no panic, and then anything else is whatever object the panic threw. (panic throws an interface)

This pattern is used a lot in the `consensus` package, but without the helper function. No need to migrate it over proactively, but if you run into it feel free to update.

PR #885, opened 2015-12-24
Persistent Host Consensus

DavidVorick:
Host now saves information that it parses from the consensus set, and will do a rescan in the event of desynchronization. This should shave 5 seconds off startup time on fast computers, and about 25 seconds on slow computers (perhaps 20% of the current startup time).

I also reorganized the persistence management quite a bit, changed the default price, and added test coverage for all the things I touched (minus mocking coverage).


api/host_test.go
```diff
@@ -75,7 +74,7 @@ func TestIntegrationHosting(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	expRevenue := "31857777777777575235"
+	expRevenue := "15928888888855757473"
```
DavidVorick: default price cut in half, hence this value changing

modules/host/host.go
```diff
@@ -23,7 +25,7 @@ var (
 	// defaultPrice defines the starting price for hosts selling storage. We
 	// try to match a number that is both reasonably profitable and reasonably
 	// competitive.
-	defaultPrice = types.SiacoinPrecision.Div(types.NewCurrency64(4320e9 / 200)) // 200 SC / GB / Month
+	defaultPrice = types.SiacoinPrecision.Div(types.NewCurrency64(4320e9)).Mul(types.NewCurrency64(100)) // 100 SC / GB / Month
```
DavidVorick: felt it was better to keep the defaults more in line with the actual expected price on the network. Starting high isn't great.

modules/host/host.go
```diff
@@ -46,29 +51,54 @@ type Host struct {
...
-	secretKey   crypto.SecretKey
-	publicKey   types.SiaPublicKey
-	modules.HostSettings
+	// Statistics
+	profit types.Currency
 
 	// Utilities.
 	listener   net.Listener
-	log        *log.Logger
+	log        *persist.Logger
```
DavidVorick: switched to a persist.Logger to save some code, and to get consistency with the miner module

modules/host/network.go
```diff
@@ -0,0 +1,25 @@
...
+		return err
+	}
+	h.netAddress = modules.NetAddress(h.listener.Addr().String())
+
+	// Networking subroutines.
+	go h.forwardPort(h.netAddress.Port())
+	go h.learnHostname()
+	go h.listen()
+	return nil
+}
```
DavidVorick: I felt like this part of startup really deserved its own function

modules/host/persist.go
```diff
@@ -1,80 +1,165 @@
...
+		Key:       pk[:],
+	}
+
+	// Save the defaults to disk.
+	err = h.save()
+	if err != nil {
+		return err
+	}
+	return nil
+}
```
DavidVorick: I also felt like this part of startup deserved its own function

modules/host/persist.go
```diff
@@ -1,80 +1,165 @@
...
-		if len(obligation.LastRevisionTxn.FileContractRevisions) > 0 { // COMPATv0.4.8
+		// update spaceRemaining to account for the storage held by this
+		// obligation.
+		if len(obligation.LastRevisionTxn.FileContractRevisions) > 0 {
 			h.spaceRemaining -= int64(obligation.LastRevisionTxn.FileContractRevisions[0].NewFileSize)
 		}
 	}
-	h.secretKey = sHost.SecretKey
-	h.publicKey = sHost.PublicKey
+}
```
DavidVorick: obligation handling broken into separate functions, because the logic was complex enough to justify it.

modules/host/update.go
```diff
@@ -100,4 +100,14 @@ func (h *Host) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+
+	// Update the host's recent change pointer to point to the most recent
+	// change.
+	h.recentChange = cc.ID
+
+	// Save the host.
+	err := h.save()
+	if err != nil {
+		h.log.Println("ERROR: could not save during ProcessConsnesusChange:", err)
+	}
```
DavidVorick: now that persistence is being saved, host needs to save after every update.

modules/miner/miner_test.go
```diff
@@ -272,17 +271,15 @@ func TestIntegrationStartupRescan(t *testing.T) {
...
 	oldTarget := mt.miner.persist.Target
 
-	// Corrupt the miner, the corruption should be fixed by the rescan.
+	// Corrupt the miner and verify that a rescan repairs the corruption.
 	mt.miner.persist.RecentChange[0]++
 	mt.miner.persist.Height += 500
 	mt.miner.persist.Target[0]++
-
-	// Call rescan and block until the scan is complete.
+	mt.cs.Unsubscribe(mt.miner)
```
DavidVorick: I forgot to update this test when I took unsubscribing out of the miner rescan function (not needed, as it only happens at startup before there is a successful subscription)

modules/host/host.go
```diff
@@ -148,6 +158,38 @@ func (h *Host) Capacity() int64 {
...
+func (h *Host) Close() error {
+	// The order in which things are closed has been explicitly chosen to
+	// minimize turbulence in the event of an error.
+
+	// Save the latest host state.
+	err := func() error {
+		h.mu.Lock()
+		defer h.mu.Unlock()
+		return h.save()
+	}()
```
lukechampine: I really don't see why you need this pattern for a single line of under-lock code. It doesn't seem worth the trouble. I think it's saner to reserve the inline-`defer` pattern for large blocks of under-lock code where multiple calls to `Unlock` would be required, and to stick a comment above each of them so that the reader isn't baffled by a random inline function.

modules/host/persist.go
```diff
@@ -1,80 +1,165 @@
...
-		SecretKey:      h.secretKey,
-		PublicKey:      h.publicKey,
-	}
+// getObligations returns a slice containing all of the contract obligations
+// currently being tracked by the host.
+func (h *Host) getObligations() []contractObligation {
+	cos := make([]contractObligation, 0, len(h.obligationsByID))
 	for _, ob := range h.obligationsByID {
-		// to avoid race conditions involving the obligation's mutex, copy it
-		// manually into a new object
```
lukechampine: this comment is important

modules/host/persist.go
```diff
@@ -1,80 +1,165 @@
...
 		height := obligation.FileContract.WindowStart + StorageProofReorgDepth
+		// Sanity check - if the height is below the current height, then set
+		// the height to current height + 3. This makes sure that all file
+		// contracts will eventually be hit or garbage collected by the host,
+		// even if a bug means that they aren't acted upon at the right moment.
+		if build.DEBUG && height < h.blockHeight {
+			panic("host settings file is inconsistent")
+		} else if height < h.blockHeight {
+			height = h.blockHeight + 3
+		}
```
lukechampine: might be clearer:
```go
if height < h.blockHeight {
    if build.DEBUG {
        panic("host settings file is inconsistent")
    }
    height = h.blockHeight + 3
}
```
and why +3? Why not +2 or +1?

modules/host/host.go
```diff
@@ -84,57 +114,37 @@ func New(cs modules.ConsensusSet, tpool modules.TransactionPool, wallet modules.
...
 	// Load the old host data and initialize the logger.
-	err = h.initPersist()
+	err := h.initPersist()
 	if err != nil {
 		return nil, err
 	}
 
-	// Create listener and set address.
-	h.listener, err = net.Listen("tcp", addr)
+	// Get the host established on the network.
```
lukechampine: you really like that word huh

modules/host/persist.go
lukechampine: I don't like how spaghettified the initialization is becoming, you have to hunt and search through two or three different call chains to find where a particular field gets set.

modules/host/update.go
lukechampine: typo

modules/host/persist.go
DavidVorick: Maybe it's a difference of top-down vs. bottom-up organization. I didn't like how complicated the `New` function was, so I split it into parts that initialized the persistence (logger + settings), and parts that initialized the network components.

Maybe there's a better organization, but I think that splitting it up is better than keeping it all in the primary function. I think it makes sense to call `establishDefaults` after loading fails, as opposed to previously where the defaults were established and then overwritten by the `load` call (which I believe actually was causing a bug where one value wasn't properly being loaded, so instead was always the default).

Network stuff on the other hand doesn't have any persistence, and should be called every time.

Or maybe the host struct itself should be organized by how the fields get initialized. It's pretty clear that all of the module dependencies are set in `New`, but the defaults are split across Host context and file management, and stuff like the height is pretty confusing too.

modules/host/host.go
DavidVorick: It's more than one line of code under lock, just deceptive. The alternative:

```
h.mu.Lock()
err := h.save()
if err != nil {
    h.mu.Unlock() // prone to be forgotten! pattern common in project history!
    return err
}
h.mu.Unlock()
```

As you can see, that's actually 5 lines of code under lock, and it includes a pattern that has caused frustrating deadlocks in the past. The way I've done it is more verbose by only 1 line of code, but explicitly avoids a known problematic pattern.

I'm very strict with the locking stuff because it's easy to make a mistake and very difficult to debug even when the mistake is a harmless as forgetting to unlock before returning, especially if you forgot when returning an error, because that means you have to figure out that the deadlock is only repeatable in the event of the error.  We've lost several days at a time previously to locking mistakes and I'm eager to avoid that in the future.

I do want to make it a convention that you need an anonymous function anytime you do locking that would require making more than 1 call to unlock (most common example is to return an error), or any time that the lock is held for more than 5 lines.

oh, I guess you could:

```
h.mu.Lock()
err := save()
h.mu.Unlock()
if err != nil {
    return err
}
```

But I'm strongly opposed to anything that would require writing multiple calls to unlock for a single call of lock (due to branching), **strongly** in favor of clearing indicating the scope in which a lock is held.

modules/host/persist.go
DavidVorick: I don't think that it's correct, in fact it's not clear to me that you can't just call `cos = append(cos, ob)`. The obligation's mutex doesn't cover any of the data in the structure, only the actual act of doing a revision.

And if you are going to be doing revisions on the obligation, that's an indication to me that you should be passing the whole obligation by reference instead of by value.  Maybe it should be passed by reference?

I don't think there's a race condition here, though `go vet` will complain about passing a mutex by value.

modules/host/persist.go
DavidVorick: yeah probably better to nest them. I actually forgot the `else if` and just did `else` on the first pass, which caused the testing to fail.

DavidVorick: I didn't want to do `+1` because I was afraid of off-by-one errors somewhere. It doesn't really hurt to wait an extra block, and if the consensus stuff is going to be passing more blocks down the pipe, I thought there might be more reason to wait. I'm not really sure why 3 exactly, but it's not code that's supposed to be hit ever anyway. I guess you could run into problems if someone never ran their node for more than 3 blocks at a time.

DavidVorick:
In an attempt to 'de-spaghettify' the code, I ran into an issue. The persistent file is responsible for multiple functions, which means that 'init persist' ends up needing to handle multiple separate logical items, including establishing default settings and creating cryptographic keys for network communication, as well as managing existing file contracts.

I did my best though, let me know what you think now.

modules/host/persist.go
lukechampine: there is a race condition. It occurs whenever you are concurrently 1) copying a mutex (or passing it by value) and 2) calling `Lock` or `Unlock` on the mutex. This is because those functions use atomic operations on the fields of the mutex, and copying an object is (apparently) not an atomic operation. Perhaps the comments should have been clearer about that.
I'm not sure, but it's possible that the race could be avoided if we saved `ob` rather than `*ob`. The JSON marshaller will still access the mutex field, but it won't (I think) access any of its elements.

modules/host/host.go
lukechampine: I've always done the latter, e.g. in host/negotiate.go [here](https://github.com/NebulousLabs/Sia/blob/46ce037b89bb4e6ea1adbee1937e41969ac69715/modules/host/negotiate.go#L195), [here](https://github.com/NebulousLabs/Sia/blob/46ce037b89bb4e6ea1adbee1937e41969ac69715/modules/host/negotiate.go#L298), and [here](https://github.com/NebulousLabs/Sia/blob/46ce037b89bb4e6ea1adbee1937e41969ac69715/modules/host/negotiate.go#L346). I thought we figured out a while ago that a good pattern is "gather data under lock, branch out of lock." Sometimes this isn't feasible, but 90% of the time it is. As far as I'm aware, the former pattern in your comment is relatively uncommon in our code now.
I think a simple convention of "avoid branching under locks" is better than rewriting half of the project's locking code in an unintuitive new style.

VoidingWarranties: I think the latter approach (locking and unlocking with no branches in between) is a preferable style as long as there are a few number of lines (i.e. 1) between the lock and unlock. Otherwise it would be too easy to accidentally add a line that branches without unlocking.

DavidVorick: Yeah, the biggest thing I'm concerned about is the number of lines that happen under lock. It's very important to me that it's extremely obvious which code is under lock and which code is not. That's why we've got a few conventions including 'exported functions never under lock', 'non-exported functions always under lock', and 'threaded functions not under lock, but must be called in a gothread'.

I also think the 'gather data under lock, branch out of lock' paradigm is very good, but only as long as the scoping is obvious and explicit. I've changed it for this PR

modules/host/persist.go
DavidVorick: I thought that atomic operations meant that you didn't need a lock to avoid race conditions? The `sync` documentation doesn't say anything about needing to hold a lock while creating a copy of the lock.

lukechampine: here's a simple program that will trigger a race:
```go
package main

import "sync"

func deref(mu *sync.Mutex) {
	_ = *mu
}

func main() {
	var mu sync.Mutex
	go deref(&mu)
	mu.Lock()
}
```

atomic operations *do* mean that you don't need a lock; however, the tradeoff is that you *must* use atomic operations (e.g. `atomic.LoadInt32`) for both reading and writing. Dereferencing a pointer is not an atomic operation, hence the race.
`sync` doesn't say anything in the docstring for `Mutex`, but in the overview it states:
>Values containing the types defined in this package should not be copied.

DavidVorick:
The build failed because `TestRelayNodes` failed, not sure what's up with that.

PR #886, opened 2015-12-28
Add tests for safeFiles

VoidingWarranties:
~~Use `ioutil.TempFile` instead of adding a "_temp" suffix when creating a new `safeFile`.~~

Added tests as well.

### DIFF
--- a/persist/persist.go
+++ b/persist/persist.go
@@ -49,8 +49,17 @@ type safeFile struct {
 	finalName string
 }
 
-// Commit renames the file to the intended final filename.
+// Commit syncs the file and then renames it to the intended final filename.
+// Writing to the file after calling Commit will succeed but will write to the
+// final file location (sf.Name() will deceptively still point to the old file
+// location).  Therefore it is recommended that the file handle be closed
+// immediately after calling Commit. Note that the file must not be closed
+// before calling commit as this will cause the sync to fail.
 func (sf *safeFile) Commit() error {
+	err := sf.Sync()
+	if err != nil {
+		return err
+	}
 	return os.Rename(sf.finalName+"_temp", sf.finalName)
 }
 
--- a/persist/persist.go
+++ b/persist/persist.go
@@ -5,6 +5,7 @@ import (
 	"encoding/base32"
 	"errors"
 	"os"
+	"path/filepath"
 )
 
 const (
@@ -60,5 +61,14 @@ func NewSafeFile(filename string) (*safeFile, error) {
 	if err != nil {
 		return nil, err
 	}
-	return &safeFile{file, filename}, nil
+
+	// Get the absolute path of the filename so that calling os.Chdir in
+	// between calling NewSafeFile and calling safeFile.Commit does not change
+	// the final file path.
+	absFilename, err := filepath.Abs(filename)
+	if err != nil {
+		return nil, err
+	}
+
+	return &safeFile{file, absFilename}, nil
 }
--- a/persist/persist_test.go
+++ b/persist/persist_test.go
@@ -45,50 +45,52 @@ func TestUnitNewSafeFile(t *testing.T) {
 	}
 }
 
-// testSafeFileWithPath tests creating and committing safe files with a
-// specified path.
-func testSafeFileWithPath(filename string, t *testing.T) {
-	// Create safe file.
-	sf, err := NewSafeFile(filename)
+// TestSafeFile tests creating and committing safe files with both relative and
+// absolute paths.
+func TestSafeFile(t *testing.T) {
+	// Generate absolute path filename.
+	absPath := filepath.Join(os.TempDir(), "NewSafeFile test file"+RandomSuffix())
+	// Get relative path filename from absolute path.
+	wd, err := os.Getwd()
 	if err != nil {
 		t.Fatal(err)
 	}
-	// These two defers seem redundant but are not. If sf.Commit() fails to
-	// move the file then the first defer is necessary. Otherwise the second
-	// defer is necessary.
-	defer os.Remove(sf.Name())
-	defer os.Remove(sf.finalName)
-
-	// Check that the name of the file is not equal to the final name of the
-	// file.
-	if sf.Name() == sf.finalName {
-		t.Errorf("safeFile temporary filename and finalName are equivalent: %s\n", sf.Name())
-	}
-
-	sf.Close()
-	err = sf.Commit()
+	relPath, err := filepath.Rel(wd, absPath)
 	if err != nil {
 		t.Fatal(err)
 	}
 
-	// Check that commiting moved the file to the originally specified path.
-	_, err = os.Stat(filename)
-	if err != nil {
-		t.Fatal("safeFile not committed correctly.")
+	// Test creating and committing a safe file with each filename.
+	filenames := []string{absPath, relPath}
+	for _, filename := range filenames {
+		// Create safe file.
+		sf, err := NewSafeFile(filename)
+		if err != nil {
+			t.Fatal(err)
+		}
+		// These two defers seem redundant but are not. If sf.Commit() fails to
+		// move the file then the first defer is necessary. Otherwise the second
+		// defer is necessary.
+		defer os.Remove(sf.Name())
+		defer os.Remove(sf.finalName)
+
+		// Check that the name of the file is not equal to the final name of the
+		// file.
+		if sf.Name() == sf.finalName {
+			t.Errorf("safeFile created with filename: %s has temporary filename that is equivalent to finalName: %s\n", filename, sf.Name())
+		}
+
+		// Check that committing doesn't return an error.
+		sf.Close()
+		err = sf.Commit()
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		// Check that commiting moved the file to the originally specified path.
+		_, err = os.Stat(filename)
+		if err != nil {
+			t.Fatalf("safeFile created with filename: %s not committed correctly to: %s\n", filename, sf.finalName)
+		}
 	}
-}
-
-// TestAbsolutePathSafeFile tests that safe files created with an absolute path
-// are created and committed correctly.
-func TestAbsolutePathSafeFile(t *testing.T) {
-	filename := filepath.Join(os.TempDir(), "NewSafeFile test file"+RandomSuffix())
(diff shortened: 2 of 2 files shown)


lukechampine:
I'm not sure I agree with this. One advantage of `_temp` is that if a user complains of data loss, we can say "just rename x_temp to x." Also, the tmp dir is subject to cleanup by the OS, so those files may be lost after a reboot or similar event.

persist/persist_test.go
```diff
@@ -26,3 +26,69 @@ func TestIntegrationRandomSuffix(t *testing.T) {
...
+	if err != nil {
+		t.Fatal("safeFile not committed correctly.")
+	}
+}
+
+// TestAbsolutePathSafeFile tests that safe files created with an absolute path
+// are created and committed correctly.
+func TestAbsolutePathSafeFile(t *testing.T) {
+	filename := filepath.Join(os.TempDir(), "NewSafeFile test file"+RandomSuffix())
+	absFilename, _ := filepath.Abs(filename)
```
lukechampine: won't `filename` be an absolute path anyway? `os.TempDir` returns `/tmp` on Linux.

VoidingWarranties:
This is true, although the "_temp" file is likely to be incomplete. Isn't the purpose of safeFile to prevent incomplete data? Unless the user or Sia can fix up an incomplete file by filling in missing information.

persist/persist.go
```diff
@@ -50,15 +52,25 @@ type safeFile struct {
...
 		return nil, err
 	}
-	return &safeFile{file, filename}, nil
+
+	// Get the absolute path of the filename so that calling os.Chdir inbetween
+	// creating a safeFile and commiting does not change the final file path.
+	absFilename, err := filepath.Abs(filename)
+	if err != nil {
+		return nil, err
+	}
```
VoidingWarranties: Regardless of whether this PR is accepted or not I think this change could prevent potential bugs if `os.Chdir` is every inadvertantly called between `NewSafeFile` and `safeFile.Commit`.

persist/persist_test.go
VoidingWarranties: Probably. `os.TempDir` uses the `TMPDIR` environment variable which *could* be a relative path. Probably not though.

persist/persist_test.go
```diff
@@ -26,3 +26,69 @@ func TestIntegrationRandomSuffix(t *testing.T) {
...
+	sf.Close()
+	err = sf.Commit()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Check that commiting moved the file to the originally specified path.
+	_, err = os.Stat(filename)
+	if err != nil {
+		t.Fatal("safeFile not committed correctly.")
```
lukechampine: errors should not end in punctuation

persist/persist_test.go
```diff
@@ -26,3 +26,69 @@ func TestIntegrationRandomSuffix(t *testing.T) {
...
+	// Check that the name of the file is not equal to the final name of the
+	// file.
+	if sf.Name() == sf.finalName {
+		t.Errorf("safeFile temporary filename and finalName are equivalent: %s\n", sf.Name())
+	}
+}
+
+// testSafeFileWithPath tests creating and committing safe files with a
+// specified path.
+func testSafeFileWithPath(filename string, t *testing.T) {
```
lukechampine: we prefer to avoid passing `testing.T` around to functions because it can obscure the exact nature of the failure (e.g. multiple failing tests will cite the same line number). Instead, have the function return an error and handle the error in the actual testing function.

alternatively, these two tests might be better off as a table-driven test anyway.

VoidingWarranties:
Removed change to `ioutil.TempFile`.

lukechampine:
looks good except that the files aren't being created in `build.SiaTestingDir`. Also, you can just use the name of the test (e.g. `TestUnitNewSafeFile`) as the name of the file. Something like:
```go
NewSafeFile(filepath.Join(build.SiaTestingDir, "persist", "TestUnitNewSafeFile"))
```

persist/persist.go
```diff
@@ -60,5 +61,14 @@ func NewSafeFile(filename string) (*safeFile, error) {
...
 	}
-	return &safeFile{file, filename}, nil
+
+	// Get the absolute path of the filename so that calling os.Chdir in
+	// between calling NewSafeFile and calling safeFile.Commit does not change
+	// the final file path.
+	absFilename, err := filepath.Abs(filename)
+	if err != nil {
+		return nil, err
+	}
```
DavidVorick: :+1:

persist/persist_test.go
```diff
@@ -26,3 +26,71 @@ func TestIntegrationRandomSuffix(t *testing.T) {
 		file.Close()
 	}
 }
+
+// TestUnitNewSafeFile checks that a new file is created and that its name is
+// different than the finalName of the new safeFile.
+func TestUnitNewSafeFile(t *testing.T) {
+	// Create safe file.
+	sf, err := NewSafeFile("NewSafeFile test file" + RandomSuffix())
```
DavidVorick: like Luke said, this file needs to be created in the testing tmp dir

persist/persist_test.go
```diff
@@ -26,3 +26,71 @@ func TestIntegrationRandomSuffix(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	defer os.Remove(sf.Name())
+	defer sf.Close()
+
+	// Check that the name of the file is not equal to the final name of the
+	// file.
+	if sf.Name() == sf.finalName {
+		t.Errorf("safeFile temporary filename and finalName are equivalent: %s\n", sf.Name())
+	}
```
DavidVorick: This test seems awkward and implementation dependent.

The input string to `NewSafeFile` is what the final name should be, right? So

```go
finalName := filepath.Join(build.SiaTestingDir, persistDir, "Test...")
NewSafeFile(finalName)
// more stuff
if sf.Name() == finalName {
// more stuff
sf.Commit()
os.Open(finalName)
// check that the contents of the file are correct
```

persist/persist_test.go
```diff
@@ -26,3 +26,71 @@ func TestIntegrationRandomSuffix(t *testing.T) {
 		file.Close()
 	}
 }
+
+// TestUnitNewSafeFile checks that a new file is created and that its name is
+// different than the finalName of the new safeFile.
+func TestUnitNewSafeFile(t *testing.T) {
```
DavidVorick: technically not a unit test because you are using the actual os package, which is IO and slow.

persist/persist_test.go
```diff
@@ -26,3 +26,71 @@ func TestIntegrationRandomSuffix(t *testing.T) {
...
+	}
+	defer os.Remove(sf.Name())
+	defer sf.Close()
+
+	// Check that the name of the file is not equal to the final name of the
+	// file.
+	if sf.Name() == sf.finalName {
+		t.Errorf("safeFile temporary filename and finalName are equivalent: %s\n", sf.Name())
+	}
+}
```
DavidVorick: I might go as far as saying that `TestUnitNewSafeFile` is not necessary, because `TestSafeFile` is more thorough and is essentially just as fast.

persist/persist_test.go
```diff
@@ -26,3 +26,71 @@ func TestIntegrationRandomSuffix(t *testing.T) {
...
+// absolute paths.
+func TestSafeFile(t *testing.T) {
+	// Generate absolute path filename.
+	absPath := filepath.Join(os.TempDir(), "NewSafeFile test file"+RandomSuffix())
+	// Get relative path filename from absolute path.
+	wd, err := os.Getwd()
+	if err != nil {
+		t.Fatal(err)
+	}
+	relPath, err := filepath.Rel(wd, absPath)
```
DavidVorick: can you write a comment about why you are checking both the relative and the absolute path?

DavidVorick: Also, in this case, because the relative and the absolute path have nothing in common, you are going to end up with the same string.

persist/persist_test.go
```diff
@@ -26,3 +26,71 @@ func TestIntegrationRandomSuffix(t *testing.T) {
...
+		// Create safe file.
+		sf, err := NewSafeFile(filename)
+		if err != nil {
+			t.Fatal(err)
+		}
+		// These two defers seem redundant but are not. If sf.Commit() fails to
+		// move the file then the first defer is necessary. Otherwise the second
+		// defer is necessary.
+		defer os.Remove(sf.Name())
+		defer os.Remove(sf.finalName)
```
DavidVorick: seems clunky.

This might be a special case, but I think you'd want to leave the files around anyway, that way if something goes wrong during testing you can check out the tempdir and see what happened.

persist/persist_test.go
```diff
@@ -26,3 +26,71 @@ func TestIntegrationRandomSuffix(t *testing.T) {
...
+		err = sf.Commit()
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		// Check that commiting moved the file to the originally specified path.
+		_, err = os.Stat(filename)
+		if err != nil {
+			t.Fatalf("safeFile created with filename: %s not committed correctly to: %s\n", filename, sf.finalName)
+		}
```
DavidVorick: you should verify the contents of the file as well as just checking that it exists.

persist/persist_test.go
lukechampine: yeah, I was thinking along the same lines.

persist/persist_test.go
lukechampine: I think the relative path will be something like `../../../../tmp/foo`

VoidingWarranties: Correct, the relative path is relative to the working directory. I added a comment to clarify.

persist/persist_test.go
```diff
@@ -26,3 +28,123 @@ func TestIntegrationRandomSuffix(t *testing.T) {
...
+	}
+
+	// Check that the file exists and has same data that was written to it.
+	f, err := os.Open(absPath)
+	defer f.Close()
+	if err != nil {
+		t.Fatal(err)
+	}
+	dataRead := make([]byte, 11)
+	n, err := f.Read(dataRead)
```
lukechampine: don't forget that you need `io.ReadAll` to guard against partial reads!

In this case it doesn't matter though, you should just use `ioutil.ReadFile`.

PR #888, opened 2015-12-30
Fix potential partial read bug of random data

VoidingWarranties:
by replacing `entropySource.Read(entropy[:])` with `io.ReadFull(entropySource, entropy[:])`


DavidVorick:
props, that's subtle

PR #889, opened 2015-12-30
Define GenesisBlock in types package

lukechampine:
I opted for the smallest possible diff here, but one alternative would be to define `GenesisBlock` *instead* of `GenesisSiafundAllocation` (and maybe `GenesisTimestamp` as well), since those values can easily be obtained by via `GenesisBlock`.


DavidVorick:
hmm.

Low code churn is good.

PR #890, opened 2015-12-31
Fixed a couple API.md things

Mingling94:


doc/API.md
```diff
@@ -1087,7 +1085,6 @@ Response:
 	UploadProgress float32
 	Nickname       string
 	Filesize       uint64
-	TimeRemaining  types.BlockHeight (uint64)
```
lukechampine: `TimeRemaining` has been changed to `Expiration`, which is an absolute block height rather than a relative number.

Mingling94: Kay, reflected.

Shouldn't there be a `UploadHeight` or something to that effect to help determine `TimeRemaining` and/or things like: total rented time, total rent paid, etc?

lukechampine: Probably. We're working on adding more comprehensive metrics to the renter now.

PR #891, opened 2016-01-03
Revamp API

lukechampine:
- Remove `graceful`
- Add `httprouter`
- Tweak consensus, explorer, and gateway routes
- Remove `/daemon/update` routes
- Overhaul renter routes

`httprouter` is pretty great, a very minimal and high-performance library that gives us easy access to path parameters like `/wallet/transactions/<id>`. It also lets us easily enforce proper methods, so I was able to toss out hundreds of lines of method checking code.

All API calls should return structs with lowercase field names now. (No more bare return values.)

I've updated siac to implement these API changes, but Sia-UI (and the explorer) will need some adjustments. We can also discuss whether the changes to various calls seem reasonable.

More details in the commit messages.


api/api.go
```diff
@@ -35,120 +30,115 @@ func HttpPOST(url string, data string) (resp *http.Response, err error) {
 	return new(http.Client).Do(req)
 }
 
-// handleHTTPRequest is a wrapper function that handles all incoming calls to
-// the API.
-func (srv *Server) handleHTTPRequest(mux *http.ServeMux, url string, handler http.HandlerFunc) {
-	mux.HandleFunc(url, func(w http.ResponseWriter, req *http.Request) {
-		if !strings.Contains(req.UserAgent(), srv.requiredUserAgent) {
+// requireUserAgent is middleware that requires all requests to set a
+// UserAgent that contains the specified string.
```
DavidVorick: why is it called middleware?

api/api.go
```diff
@@ -35,120 +30,115 @@ func HttpPOST(url string, data string) (resp *http.Response, err error) {
...
 	})
 }
 
 // initAPI determines which functions handle each API call.
 func (srv *Server) initAPI() {
-	mux := http.NewServeMux()
-
-	// 404 Calls
-	srv.handleHTTPRequest(mux, "/", srv.unrecognizedCallHandler)
+	mux := httprouter.New()
```
DavidVorick: is mux still the best name?

api/api.go
```diff
@@ -35,120 +30,115 @@ func HttpPOST(url string, data string) (resp *http.Response, err error) {
...
 
 	// Daemon API Calls - Unfinished
-	srv.handleHTTPRequest(mux, "/daemon/constants", srv.daemonConstantsHandler)
-	srv.handleHTTPRequest(mux, "/daemon/version", srv.daemonVersionHandler)
-	srv.handleHTTPRequest(mux, "/daemon/stop", srv.daemonStopHandler)
-	srv.handleHTTPRequest(mux, "/daemon/updates/apply", srv.daemonUpdatesApplyHandler)
-	srv.handleHTTPRequest(mux, "/daemon/updates/check", srv.daemonUpdatesCheckHandler)
+	mux.GET("/daemon/constants", srv.daemonConstantsHandler)
+	mux.GET("/daemon/version", srv.daemonVersionHandler)
+	mux.POST("/daemon/stop", srv.daemonStopHandler)
```
DavidVorick: I believe stop is supposed to be a GET request.

DavidVorick: We decided to only make things POST if you are adding some type of entry, everything else is GET

api/api.go
```diff
@@ -35,120 +30,115 @@ func HttpPOST(url string, data string) (resp *http.Response, err error) {
...
 
 	// Consensus API Calls
 	if srv.cs != nil {
-		srv.handleHTTPRequest(mux, "/consensus", srv.consensusHandler)            // GET
-		srv.handleHTTPRequest(mux, "/consensus/block", srv.consensusBlockHandler) // GET
+		mux.GET("/consensus", srv.consensusHandler)
+		mux.GET("/consensus/blocks/:height", srv.consensusBlocksHandler)
```
DavidVorick: why can't it just be `/consensus/height`?

DavidVorick: and, wouldn't you want to make blocks singular, since you are only fetching 1 block?

api/api.go
```diff
@@ -35,120 +30,115 @@ func HttpPOST(url string, data string) (resp *http.Response, err error) {
...
 	}
 
 	// Explorer API Calls
 	if srv.explorer != nil {
-		srv.handleHTTPRequest(mux, "/explorer", srv.explorerHandler)            // GET
-		srv.handleHTTPRequest(mux, "/explorer/", srv.explorerHandler)           // $(hash) GET
-		srv.handleHTTPRequest(mux, "/explorer/block", srv.explorerBlockHandler) // GET
+		mux.GET("/explorer", srv.explorerHandler)
+		mux.GET("/explorer/blocks/:height", srv.explorerBlocksHandler)
+		mux.GET("/explorer/hash/:hash", srv.explorerHashHandler)
```
DavidVorick: Same here, I think it's fine to do `explorer/:height` and `explorer/:hash`

api/api.go
```diff
@@ -35,120 +30,115 @@ func HttpPOST(url string, data string) (resp *http.Response, err error) {
...
 	}
 
 	// Gateway API Calls - Unfinished
 	if srv.gateway != nil {
-		srv.handleHTTPRequest(mux, "/gateway/status", srv.gatewayStatusHandler)
-		srv.handleHTTPRequest(mux, "/gateway/peers/add", srv.gatewayPeersAddHandler)
-		srv.handleHTTPRequest(mux, "/gateway/peers/remove", srv.gatewayPeersRemoveHandler)
+		mux.GET("/gateway", srv.gatewayHandler)
+		mux.POST("/gateway/add/:addr", srv.gatewayAddHandler)
+		mux.POST("/gateway/remove/:addr", srv.gatewayRemoveHandler)
```
DavidVorick: I'm not familiar with using the ID in a POST request, I think we'd want to use the query string instead of the url path for adding or removing a peer. `/gateway/:addr` to GET the peer, but I'm not familiar with doing it that way to do a POST.

api/api.go
```diff
@@ -35,120 +30,115 @@ func HttpPOST(url string, data string) (resp *http.Response, err error) {
...
 	if srv.explorer != nil {
-		srv.handleHTTPRequest(mux, "/explorer", srv.explorerHandler)            // GET
-		srv.handleHTTPRequest(mux, "/explorer/", srv.explorerHandler)           // $(hash) GET
-		srv.handleHTTPRequest(mux, "/explorer/block", srv.explorerBlockHandler) // GET
+		mux.GET("/explorer", srv.explorerHandler)
+		mux.GET("/explorer/blocks/:height", srv.explorerBlocksHandler)
+		mux.GET("/explorer/hash/:hash", srv.explorerHashHandler)
 	}
 
 	// Gateway API Calls - Unfinished
```
DavidVorick: The unfinished comment is not true anymore, it seems like you've properly fixed up the API here.

api/api.go
```diff
@@ -35,120 +30,115 @@ func HttpPOST(url string, data string) (resp *http.Response, err error) {
...
+		mux.POST("/host/announce", srv.hostAnnounceHandler)
 	}
 
 	// HostDB API Calls - DEPRECATED
 	if srv.renter != nil {
-		srv.handleHTTPRequest(mux, "/hostdb/hosts/active", srv.renterHostsActiveHandler)
-		srv.handleHTTPRequest(mux, "/hostdb/hosts/all", srv.renterHostsAllHandler)
+		mux.HandlerFunc("GET", "/hostdb/hosts/active", srv.renterHostsActiveHandler)
+		mux.HandlerFunc("GET", "/hostdb/hosts/all", srv.renterHostsAllHandler)
 	}
```
DavidVorick: Do we want to preserve the deprecated calls into 0.5.0? Doesn't seem needed to me.

api/api.go
```diff
@@ -35,120 +30,115 @@ func HttpPOST(url string, data string) (resp *http.Response, err error) {
...
-		srv.handleHTTPRequest(mux, "/miner/stop", srv.minerStopHandler)                   // POST
-		srv.handleHTTPRequest(mux, "/miner/headerforwork", srv.minerHeaderforworkHandler) // COMPATv0.4.8
-		srv.handleHTTPRequest(mux, "/miner/submitheader", srv.minerSubmitheaderHandler)   // COMPATv0.4.8
+		mux.GET("/miner", srv.minerHandler)
+		mux.GET("/miner/header", srv.minerHeaderHandlerGET)
+		mux.POST("/miner/header", srv.minerHeaderHandlerPOST)
+		mux.POST("/miner/start", srv.minerStartHandler)
+		mux.POST("/miner/stop", srv.minerStopHandler)
+		mux.GET("/miner/headerforwork", srv.minerHeaderHandlerGET) // COMPATv0.4.8
+		mux.GET("/miner/submitheader", srv.minerHeaderHandlerPOST) // COMPATv0.4.8
```
DavidVorick: These do need to be preserved though because a huge part of the hashrate depends on these calls.

api/api.go
```diff
@@ -35,120 +30,115 @@ func HttpPOST(url string, data string) (resp *http.Response, err error) {
...
-		srv.handleHTTPRequest(mux, "/miner/header", srv.minerHeaderHandler)               // GET, POST
-		srv.handleHTTPRequest(mux, "/miner/start", srv.minerStartHandler)                 // POST
-		srv.handleHTTPRequest(mux, "/miner/stop", srv.minerStopHandler)                   // POST
-		srv.handleHTTPRequest(mux, "/miner/headerforwork", srv.minerHeaderforworkHandler) // COMPATv0.4.8
-		srv.handleHTTPRequest(mux, "/miner/submitheader", srv.minerSubmitheaderHandler)   // COMPATv0.4.8
+		mux.GET("/miner", srv.minerHandler)
+		mux.GET("/miner/header", srv.minerHeaderHandlerGET)
+		mux.POST("/miner/header", srv.minerHeaderHandlerPOST)
+		mux.POST("/miner/start", srv.minerStartHandler)
+		mux.POST("/miner/stop", srv.minerStopHandler)
```
DavidVorick: Start and Stop should be GET calls

api/api.go
```diff
@@ -35,120 +30,115 @@ func HttpPOST(url string, data string) (resp *http.Response, err error) {
...
+		mux.GET("/miner/headerforwork", srv.minerHeaderHandlerGET) // COMPATv0.4.8
+		mux.GET("/miner/submitheader", srv.minerHeaderHandlerPOST) // COMPATv0.4.8
 	}
 
 	// Renter API Calls - Unfinished
```
DavidVorick: Unfinished comment not needed anyore

api/api.go
```diff
@@ -35,120 +30,115 @@ func HttpPOST(url string, data string) (resp *http.Response, err error) {
...
+		mux.POST("/renter/loadascii", srv.renterLoadAsciiHandler)
+		mux.POST("/renter/rename", srv.renterRenameHandler)
+		mux.POST("/renter/delete/*path", srv.renterDeleteHandler)
+		mux.GET("/renter/download/*path", srv.renterDownloadHandler)
+		mux.GET("/renter/share/*path", srv.renterShareHandler)
+		mux.GET("/renter/shareascii/*path", srv.renterShareAsciiHandler)
+		mux.POST("/renter/upload/*path", srv.renterUploadHandler)
 	}
 
 	// TransactionPool API Calls - Unfinished
```
DavidVorick: unfinished comment not needed anymore

api/api.go
```diff
@@ -35,120 +30,115 @@ func HttpPOST(url string, data string) (resp *http.Response, err error) {
...
+		mux.GET("/renter/downloadqueue", srv.renterDownloadqueueHandler)
+		mux.GET("/renter/files", srv.renterFilesHandler)
+		mux.POST("/renter/load", srv.renterLoadHandler)
+		mux.POST("/renter/loadascii", srv.renterLoadAsciiHandler)
+		mux.POST("/renter/rename", srv.renterRenameHandler)
+		mux.POST("/renter/delete/*path", srv.renterDeleteHandler)
+		mux.GET("/renter/download/*path", srv.renterDownloadHandler)
+		mux.GET("/renter/share/*path", srv.renterShareHandler)
+		mux.GET("/renter/shareascii/*path", srv.renterShareAsciiHandler)
+		mux.POST("/renter/upload/*path", srv.renterUploadHandler)
```
DavidVorick: what is the difference between `*stuff` and `:stuff`?

api/api.go
```diff
@@ -35,120 +30,115 @@ func HttpPOST(url string, data string) (resp *http.Response, err error) {
...
 	}
```
DavidVorick: please keep everything alphabetically ordered by path unless there is an explicit thing separating the calls.

api/api.go
```diff
@@ -35,120 +30,115 @@ func HttpPOST(url string, data string) (resp *http.Response, err error) {
...
-		srv.handleHTTPRequest(mux, "/wallet/siafunds", srv.walletSiafundsHandler)          // POST
-		srv.handleHTTPRequest(mux, "/wallet/transaction/", srv.walletTransactionHandler)   // $(id) GET
-		srv.handleHTTPRequest(mux, "/wallet/transactions", srv.walletTransactionsHandler)  // GET
-		srv.handleHTTPRequest(mux, "/wallet/transactions/", srv.walletTransactionsHandler) // $(addr) GET
-		srv.handleHTTPRequest(mux, "/wallet/unlock", srv.walletUnlockHandler)              // POST
+		mux.GET("/wallet", srv.walletHandler)
+		mux.GET("/wallet/address", srv.walletAddressHandler)
+		mux.GET("/wallet/addresses", srv.walletAddressesHandler)
+		mux.GET("/wallet/backup", srv.walletBackupHandler)
+		mux.POST("/wallet/encrypt", srv.walletInitHandler) // COMPATv0.4.0
```
DavidVorick: This call should go on the bottom since it's not technically supported anymore

DavidVorick: We probably shouldn't remove it, it's been a while since it's been needed, but poloniex might still rely on it

api/api.go
```diff
@@ -35,120 +30,115 @@ func HttpPOST(url string, data string) (resp *http.Response, err error) {
...
-		srv.handleHTTPRequest(mux, "/wallet/lock", srv.walletLockHandler)                  // POST
-		srv.handleHTTPRequest(mux, "/wallet/seeds", srv.walletSeedsHandler)                // GET
-		srv.handleHTTPRequest(mux, "/wallet/siacoins", srv.walletSiacoinsHandler)          // POST
-		srv.handleHTTPRequest(mux, "/wallet/siafunds", srv.walletSiafundsHandler)          // POST
-		srv.handleHTTPRequest(mux, "/wallet/transaction/", srv.walletTransactionHandler)   // $(id) GET
-		srv.handleHTTPRequest(mux, "/wallet/transactions", srv.walletTransactionsHandler)  // GET
-		srv.handleHTTPRequest(mux, "/wallet/transactions/", srv.walletTransactionsHandler) // $(addr) GET
-		srv.handleHTTPRequest(mux, "/wallet/unlock", srv.walletUnlockHandler)              // POST
+		mux.GET("/wallet", srv.walletHandler)
+		mux.GET("/wallet/address", srv.walletAddressHandler)
```
DavidVorick: `/wallet/:address`

api/api.go
```diff
@@ -35,120 +30,115 @@ func HttpPOST(url string, data string) (resp *http.Response, err error) {
...
+		mux.GET("/wallet/addresses", srv.walletAddressesHandler)
+		mux.GET("/wallet/backup", srv.walletBackupHandler)
+		mux.POST("/wallet/encrypt", srv.walletInitHandler) // COMPATv0.4.0
+		mux.POST("/wallet/init", srv.walletInitHandler)
+		mux.POST("/wallet/load/033x", srv.walletLoad033xHandler)
+		mux.POST("/wallet/load/seed", srv.walletLoadSeedHandler)
+		mux.POST("/wallet/load/siag", srv.walletLoadSiagHandler)
```
DavidVorick: For these 3 load calls, I'm wondering if the `load` is necessary in the path. `mux.POST("/wallet/033x")` seems pretty clear to me

doc/API.md
```diff
@@ -27,13 +27,115 @@ Example GET curl call:  `curl -A "Sia-Agent" /wallet/transactions?startheight=1&
 
 Example POST curl call: `curl -A "Sia-Agent" --data "amount=123&destination=abcd" /wallet/siacoins
 
+Daemon
+------
+
+Queries:
+
+* /daemon/constants [GET]
+* /daemon/stop      [POST]
```
DavidVorick: GET

doc/API.md
```diff
@@ -27,13 +27,115 @@ Example GET curl call:  `curl -A "Sia-Agent" /wallet/transactions?startheight=1&
...
+
 Consensus
 ---------
 
 Queries:
 
-* /consensus       [GET]
-* /consensus/block [GET]
+* /consensus                 [GET]
+* /consensus/blocks/<height> [GET]
```
DavidVorick: this syntax conflicts with what's used elsewhere

doc/API.md
```diff
@@ -57,7 +159,7 @@ struct {
 'target' is the hash that needs to be met by a block for the block to be valid.
 The target is inversely proportional to the difficulty.
 
-#### /consensus/block [GET]
+#### /consensus/blocks/:height [GET]
```
DavidVorick: namely here

DavidVorick: also should either be `/consensus/:height` or `/consensus/block/:height`

doc/API.md
```diff
@@ -80,9 +182,9 @@ Explorer
 
 Queries:
 
-* /explorer         [GET]
-* /explorer/block   [GET]
-* /explorer/$(hash) [GET]
+* /explorer                 [GET]
+* /explorer/blocks/<height> [GET]
+* /explorer/hash/<hash>     [GET]
```
DavidVorick: I think that `/explorer/:hash` makes more sense

doc/API.md
```diff
@@ -137,15 +241,18 @@ struct {
 }
 ```
 
-#### /explorer/$(hash) [GET]
+#### /explorer/hash/<hash> [GET]
```
DavidVorick: ditto

api/api.go
DavidVorick: We should probably have a longer conversation about the renter api. I don't think it feels very refined at the moment.

DavidVorick:
Overall I like these changes. Some of the code changed too much for the diff to be much use, but I'll be looking at the raw files in a bit.

As far as I understand, you've now moved the whole api over to a RESTful paradigm? If so, that's a pretty significant milestone.

api/consensus.go
```diff
@@ -57,12 +50,3 @@ func (srv *Server) consensusBlockHandlerGET(w http.ResponseWriter, req *http.Req
 		Block: block,
```
DavidVorick: need to return a `ConsensusBlocksGET` if the call is `/consensus/blocks`

api/api.go
DavidVorick: I'm tempted to say that these handlers should all be suffixed by the request type that they handle, e.g. `srv.ConsensusBlocksHandlerGET`

api/daemon.go
```diff
@@ -44,170 +40,36 @@ bwIDAQAB
...
-// server operated by the developers. The most recent version is stored in
-// current/. The folder contains the files changed by the update, as well as a
-// MANIFEST file that contains the version number and a file listing. To check
-// for an update, we first read the version number from current/MANIFEST. If
-// the version is newer, we download and apply the files listed in the update
-// manifest.
-var updateURL = "http://23.239.14.98/releases/" + runtime.GOOS + "_" + runtime.GOARCH
-
 // SiaConstants is a struct listing all of the constants in use.
 type SiaConstants struct {
```
DavidVorick: should probably be `type DaemonConstantsGET` to fit in with the rest of the folder

api/daemon.go
```diff
@@ -216,19 +78,33 @@ func (srv *Server) daemonConstantsHandler(w http.ResponseWriter, req *http.Reque
...
 		RootTarget: types.RootTarget,
 		RootDepth:  types.RootDepth,
 
 		MaxAdjustmentUp:   types.MaxAdjustmentUp,
 		MaxAdjustmentDown: types.MaxAdjustmentDown,
+
+		SiacoinPrecision: types.SiacoinPrecision,
 	}
 
 	writeJSON(w, sc)
```
DavidVorick: can probably be `writeJSON(w, SiaConstants{
    stuff,
})`

api/api.go
lukechampine: "middleware" is something of a generic term (and one I don't really like -- sounds too much like "synergy" or "paradigm"), but in the Go space for whatever reason it means chaining handlers between the basic `ServeMux` and the actual application layer.

api/api.go
lukechampine: we could use `router` or `r` instead, but `mux` isn't inaccurate; it's still a single object that multiplexes multiple routes.

api/api.go
lukechampine: `/consensus/:height` reads well here, but when you actually called it it would look like `/consensus/0`, `/consensus/1`, etc., which seemed too opaque.
I use the plural "blocks" because I imagine the blockchain as a persistent set of resources that live under the `/consensus` route. In theory, you could call `/consensus/blocks` and get some sort of metadata about the whole set (maybe the current blockheight?).
I don't think the `GET` suffix is as helpful now that we don't have method-routing functions, but I'm not entirely averse to it. I believe I kept it wherever two methods were possible on the same route.

api/api.go
lukechampine: this would be ambiguous, unfortunately. The router has no way of distinguishing between `/explorer/0` and `/explorer/a1b2c3d4...`

api/api.go
lukechampine: It's a more RESTful organization. One of the basic tenets is that resources should be uniquely identified by a URL, and should be manipulated by using different verbs on that URL. So it makes sense if you think of a peer as a resource.
It's not perfect though, because we have decided to only use GET and POST. The most RESTful organization would be:
```
PUT    /gateway/:addr (adds a peer)
GET    /gateway/:addr (accesses a peer)
DELETE /gateway/:addr (removes a peer)
```

api/api.go
lukechampine: I wasn't sure how the hostdb stuff should be handled. Are we still using these calls anywhere? If not, the renter package can be cleaned up a bit too.

api/api.go
lukechampine: yep, I made sure not to change the semantics of any miner or wallet calls

api/api.go
lukechampine: `:stuff` matches one "noun" in a URL. If your route is `/:foo/:bar` and the URL is `/abc/123`, then `foo` will be `abc` and `bar` will be `123.
`*stuff` matches everything until the end of the URL. If the route were `/:foo/*bar`, the assignments would be the same, but if it were `/*foo`, then `foo` would be `/abc/123` (leading slash is included). For obvious reasons, these can only be used at the end of the route.
See the [httprouter godoc](https://godoc.org/github.com/julienschmidt/httprouter) for more.

api/api.go
lukechampine: only `delete` and `download` are out of place, right? I like grouping together the `*path` routes, but nbd

doc/API.md
lukechampine: yeah originally I was using the same syntax as `httprouter`, but I felt it would be too confusing to have both `:` and `*`, so I switched to `<>` for both.

api/daemon.go
lukechampine: I considered that but held off because `DaemonConstants` might be a misleading name.
Also, since these are constants, we could just initialize them all here instead of instantiating a new `SiaConstants` object on every call to `/daemon/constants`. i.e.:
```go
var SiaConstants = struct {
   ...
}{ ... }
```

api/api.go
lukechampine: hmm. `siac` uses POST (and has for a while), but I guess that only worked because we weren't enforcing methods for the daemon calls yet.

api/api.go
lukechampine: btw, here are the docs for the dropbox API as a reference: https://www.dropbox.com/developers-v1/core/docs#files-GET

api/api.go
lukechampine: this call doesn't take any parameters, it just returns a new address.

api/api.go
DavidVorick: I'd probably prefer router, given that there is no context for `mux` anymore.

api/api.go
DavidVorick: The miner start and miner stop calls also did not enforce any methods

api/api.go
DavidVorick: Either it should be `/consensus/blocks` & `/consensus/blocks?height=5`, or it should be `/consensus/block/5`. Blocks can also be identified by hash, so I don't think it makes a ton of sense to use a path. I'd probably prefer to keep the querystring, especially if we're planning on adding more stuff like the ability to return multiple blocks.

But, I think `/consensus/block?height=1` is sufficient. It's not clear to me that anything specifically is superior.

api/api.go
DavidVorick: Anything equal to 64 or 76 characters is a hash, you aren't going to have a height with that many characters. Easy for the method to interpret, even if the router itself can't tell.

api/api.go
DavidVorick: We do still need to be able to see a list of hosts, but I thought we were getting that from the renter now.

api/api.go
DavidVorick: oh, heh

doc/API.md
DavidVorick: I think it makes sense to use `<>` in the docs, though it does make the path/nickname stuff more confusing

api/api.go
DavidVorick: Even if you are thinking 1 url = 1 resource, height is not perfect because you can get different blocks at the same height if the chain reorgs. Only by-hash is really unique within Sia.

doc/API.md
lukechampine: another thing that occurred to me was how we should document path parameters. In this PR they are listed along with the URL parameters. I think that the distinction is fairly obvious, but perhaps it's not explicit enough.

api/api.go
lukechampine: ok, so we just want to make them renter routes instead. Other than that, do we want to change these calls at all?

api/api.go
DavidVorick: Maybe it should be `/gateway/:addr/action`

https://www.dropbox.com/developers-v1/core/docs#fileops-copy

The fileops uses the filepath as a parameter, not as a resource in the URL, just as a reference point. I'm not sure I like it that way. I don't know what is best.

api/api.go
lukechampine: not true -- see https://github.com/NebulousLabs/Sia/blob/master/api/miner.go#L45
I'll move them to GET though

api/api.go
lukechampine: I suppose that's fair, but they're not true resources in the first place because you can't PUT or DELETE them. I just like the idea that each conceptual "thing" (block, peer, address, etc.) is represented by a path, rather than a query string.
How about `/consensus/blockchain/<height>` (or just `chain`) for accessing blocks by height, and `/consensus/block/<id>` for accessing by id?

api/api.go
lukechampine: In that case you'd have something like `/explorer/:heightOrHash`, and you'd access blocks with `/explorer/5`. I think it's better to keep routes non-magical and easy to interpret.

api/api.go
DavidVorick: oh. huh

lukechampine:
\>all those outdated diffs
wew

should be good to go now, unless you feel strongly about `DaemonConstantsGET`

doc/API.md
```diff
@@ -27,13 +27,114 @@ Example GET curl call:  `curl -A "Sia-Agent" /wallet/transactions?startheight=1&
...
+Daemon
+------
+
+Queries:
+
+* /daemon/constants [GET]
+* /daemon/stop      [GET]
+* /daemon/version   [GET]
+
+#### /daemon/constants [GET]
```
Mingling94: I was wondering, would it be useful to have an Advanced Overview plugin that just displayed lots of data in JSON format from various information/status calls like this? It'd be a pretty easy thing to make

doc/API.md
```diff
@@ -360,9 +503,236 @@ Function: Submit a header that has passed the POW.
 
 Parameters:
 ```
-[]byte
+input []byte
+```
+The input byte array should be 80 bytes that form the solved block header. *Unlike most API calls, it should be written directly to the request body, not as a query parameter.*
```
Mingling94: Why is that?

doc/API.md
```diff
@@ -360,9 +503,236 @@ Function: Submit a header that has passed the POW.
...
+	}
+}
+```
+'nickname' is the nickname given to the file when it was uploaded.
+
+'filesize' is the size of the file in bytes.
+
+'available' indicates whether or not the file can be downloaded immediately.
+
+'uploadProgress' is the current upload percentage of the file, including
```
Mingling94: `uploadprogress` or `uploadProgress`?

doc/API.md
```diff
@@ -360,9 +503,236 @@ Function: Submit a header that has passed the POW.
...
+
+#### /renter/load [POST]
+
+Function: Load a .sia file into the renter.
+
+Parameters:
+```
+filename string
+```
+'filename' is the filepath of the .sia file that is being loaded.
```
Mingling94: Why not call it filepath?

doc/API.md
```diff
@@ -360,9 +503,236 @@ Function: Submit a header that has passed the POW.
...
+nickname string
+newname  string
+```
+'nickname' is the current name of the file entry.
+
+'newname' is the new name for the file entry.
+
+Response: standard.
+
+#### /renter/delete/<path> [POST]
```
Mingling94: 1. Does this mean that the path/nickname can be supplied **both** or **either** as a query parameter or in the call itself?

2. Unfortunately markdown is able to interpret some level of HTML as well, so a `<` `>` term is interpreted as an HTML tag and not rendered on Github and other platforms: 
selection_054

doc/API.md
```diff
@@ -360,9 +503,236 @@ Function: Submit a header that has passed the POW.
...
+
+Function: Deletes a renter file entry. Does not delete any downloads or
+original files, only the entry in the renter.
+
+Parameters:
+```
+path string
+```
+'path' is the nickname of the file.
```
Mingling94: If we're adopting the idea that the 'nickname' is the 'path', shouldn't we stick to one or the other? Unless/until file system structuring is on siad's side of things, isn't naming it 'path' unneeded obfuscation?

doc/API.md
```diff
@@ -198,38 +341,38 @@ struct {
 	upcomingrevenue   types.Currency (string)
 }
 ```
-`collateral` is the number of hastings per byte per block that are put up as
+'collateral' is the number of hastings per byte per block that are put up as
 collateral when making file contracts.
 
-`ipaddress` is the network address of the host.
+'ipaddress' is the network address of the host.
```
Mingling94: Why isn't this also 'address' like in the announce call?

doc/API.md
DavidVorick: I was thinking we could have a console or an api plugin, where you can type siac commands or input api calls and get back raw output.

doc/API.md
DavidVorick: the bytes are random, meaning you can't put the data in as a query string unless you encode them. Encoding them causes blowup, and the miner is typically waiting for the call to return. Even locally, calls like this can take milliseconds, which is a nontrivial amount of time to block the miner.

Basically, hashrates drop when we use querystrings for this call.

doc/API.md
DavidVorick: `uploadprogress`

doc/API.md
DavidVorick: it should probably be `filepath`, since we're looking up the filepath of a .sia that's being loaded. I don't think 'destination' makes any sense.

doc/API.md
DavidVorick: yeah it should just be called 'nickname' here.

'filepath' is any time we're talking about stuff on-disk. The on-disk filenames and paths don't need to map to the nickname structure.

doc/API.md
DavidVorick: it should be 'netaddress'

'address' is confusing because 'address' means cryptographic address or wallet address depending on context.

doc/API.md
lukechampine: lol, good catch. We could switch to `:` or use escapes: \<path\>

doc/API.md
lukechampine: I think we should consider transitioning from "nickname" to "path". We may move path handling to siad in order to facilitate operations like, e.g. deleting a folder. "Nickname" is unintuitive for a newcomer anyway -- "What's a nickname? I just want to drag a file from my desktop into Sia."

Also, worth noting that you won't actually use the name "path" on the API side; you just type `/renter/delete/foo/bar/baz`.

doc/API.md
Mingling94: Sorry reversed that, 
>Or since /renter/share/<path> uses 'destination', ought to call it 'source' to mirror that convention.

doc/API.md
Mingling94: Ah, it could stay as path for now as a sort of implicit `TODO:`

api/renter.go
```diff
@@ -189,3 +164,18 @@ func (srv *Server) renterFilesUploadHandler(w http.ResponseWriter, req *http.Req
 
 	writeSuccess(w)
 }
+
+// renterHostsActiveHandler handes the API call asking for the list of active
+// hosts.
+func (srv *Server) renterHostsActiveHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	writeJSON(w, ActiveHosts{
+		Hosts: srv.renter.ActiveHosts(),
```
Mingling94: Shouldn't this be documented in the API.md and lowercased?

doc/API.md
```diff
@@ -360,9 +503,236 @@ Function: Submit a header that has passed the POW.
...
+'file' is the ASCII-encoded .sia file that is being loaded.
+
+Response:
+```
+struct {
+	filesadded []string
+}
+```
+
+#### /renter/rename [POST]
```
Mingling94: Why not have this endpoint also have the 'nickname' or 'path' in the request url?

api/renter.go
lukechampine: it is lowercased (see the type definition) but yeah, API.md needs to be updated

doc/API.md
DavidVorick: Isn't this technically not a parameter? I feel like url variables should have a different name from 'param', that's for querystring varibles

Mingling94: I think a "url variable" boils down to the same thing as a querystring variable. "Query string" itself implies that that the key, value will be parsed and part of the request `http://localhost:9980/module/endpoint/variable?key1=value1&key2=value2`

PR #893, opened 2016-01-06
API touchups

lukechampine:
@Mingling94 pointed out a few pieces of missing info from API.md. I knew I'd miss something :P

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -164,8 +164,8 @@ Explorer
 Queries:
 
 * /explorer                 [GET]
-* /explorer/blocks/<height> [GET]
-* /explorer/hashes/<hash>   [GET]
+* /explorer/blocks/{height} [GET]
+* /explorer/hashes/{hash}   [GET]
 
 #### /explorer [GET]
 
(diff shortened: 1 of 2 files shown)


doc/API.md
```diff
@@ -647,7 +633,7 @@ path string
```
DavidVorick: mentioned elsewhere, but I feel like 'path' really isn't a "parameter", those are supposed to mean query string parameters. 'path' is a url parameter, which I feel like should have a different section.

It is pretty unambiguous, but... hmm

doc/API.md
```diff
@@ -711,6 +713,68 @@ source   string
...
+		minduration  types.BlockHeight (uint64)
+		maxduration  types.BlockHeight (uint64)
+		windowsize   types.BlockHeight (uint64)
+		price        types.Currency    (string)
+		collateral   types.Currency    (string)
+		unlockhash   types.UnlockHash  (string)
+	}
+}
+```
+See /renter/hosts/active for a description of each field.
```
Mingling94: */renter/hosts/all

lukechampine: dammit lol

Issue #894, opened 2016-01-06
Handling paths as nicknames


lukechampine:
The new API treats nicknames as paths. Even `/renter/upload/foo.txt` will create a file called `/foo.txt`, with a leading slash. What is the right way to handle these strings?

There are three areas that need attention:
- The API. I see no problems here (for now), but we might consider dropping the leading slash from nicknames.
- The renter. Currently the renter tries to save .sia files using their nickname as the filename, so with the new API it will be creating files with slashes in their names, which is a disaster. We probably want to perform some kind of escaping here. Alternatively, the .sia filenames don't *really* need to match their nicknames; they could be numbers (like the host) or something else entirely.
- The UI+siac. Things are pretty clean for the UI, since path handling happens behind the scenes, but in siac the nicknames are more in-your-face. For example, you'll need to include the leading slash with some operations, but not others (namely `rename`, since it takes a path as a URL parameter). We could play fast and loose here (tacking on leading slashes behind the scenes), or we could be strict and require an explicit leading slash in all calls.

A more radical alternative would be to ban `/` in nicknames altogether. This is basically taking the position that siad should not know anything about paths, and that only the UI should be responsible for manipulating/parsing paths, using some other special character in place of `/`. I can sympathize with this position, but I feel it will become increasing untenable as the software matures.

I plan to implement this stuff myself, but I'm looking for input as to what seems like a good design.

Mingling94:
* I think file nicknames/paths should drop the leading slash and enable freedom for nickname/paths. As I mentioned to you in slack, uploading a file with `/renter/upload/foo` with make it impossible to delete 

* Is it hard to allow a path/nickname parameter in the url to `/renter/load` and `/renter/loadascii` like for the other endpoints? This would seem the most congruent.

* Tack on the leading slashes behind the scenes IMO. siac's a UI just like the GUI and can aim to be more convenient for a user. Most CLI tools I can think of allow file paths to start with a file or folder name and do not require a `./`

DavidVorick:
Allowing slashes in filenames is very convenient, and I think the right way to do it. Forcing things to use another character strikes me as completely awful, and forcing the UI to handle the pathing also strikes me as awful.

[...]

PR #895, opened 2016-01-07
Renter Paths

lukechampine:
As discussed in #894:
- Leading slashes have been removed
- The renter creates the directory structure in its persistence folder.

One possible modification here would be to add a `siaroot` folder to the renter folder, under which the .sia directory structure would be stored. This would prevent confusion stemming from the .sia files living alongside renter.json, renter.log, and hostdb.log.

### DIFF
--- a/api/renter_test.go
+++ b/api/renter_test.go
@@ -3,13 +3,22 @@ package api
 import (
 	"io/ioutil"
 	"net/url"
-	"os"
 	"path/filepath"
 	"testing"
 
 	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/crypto"
+	"github.com/NebulousLabs/Sia/modules/renter"
 )
+
+// createRandFile creates a file on disk and fills it with random bytes.
+func createRandFile(path string) error {
+	data, err := crypto.RandBytes(1024)
+	if err != nil {
+		return err
+	}
+	return ioutil.WriteFile(path, data, 0600)
+}
 
 // TestRenterPaths tests that the /renter routes handle path parameters
 // properly.
@@ -30,13 +39,8 @@ func TestRenterPaths(t *testing.T) {
 	}
 
 	// Create a file.
-	stDir := filepath.Join(build.SiaTestingDir, "api", "TestRenterPaths")
-	path := filepath.Join(stDir, "test.dat")
-	data, err := crypto.RandBytes(1024)
-	if err != nil {
-		t.Fatal(err)
-	}
-	err = ioutil.WriteFile(path, data, 0600)
+	path := filepath.Join(build.SiaTestingDir, "api", "TestRenterPaths", "test.dat")
+	err = createRandFile(path)
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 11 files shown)


Mingling94:
* https://github.com/NebulousLabs/Sia/pull/891/files#r48834586
* https://github.com/NebulousLabs/Sia/pull/891/files#r48919196
* https://github.com/NebulousLabs/Sia/pull/891/files#r48987747
Does this need any other reflection in the API.md or is it fine the way it is?
I know you would do `/renter/delete/{path}` instead of `/renter/delete{/path}` so I guess that's the case?

Mingling94:
Also to be even more RESTful couldn't these
/renter/delete/{path} [POST]
/renter/download/{path} [GET]
/renter/rename/{path} [POST]
/renter/share/{path} [GET]
/renter/shareascii/{path} [GET]
/renter/upload/{path} [POST]
/renter/hosts/active [GET]
be these
/renter/file/{path} [GET]
/renter/file/{path} [DELETE]
/renter/file/{path} [POST]

DavidVorick:
We made a decision to only use GET and POST in our api, because for whatever reason some libraries only have good support for GET and POST request. We sacrifice some RESTfulness, but we increase compatibility and reduce the barrier to entry for interacting with our API, which I think is more important.

lukechampine:
Important note: folders behavior may be a little strange. Namely, if your renter folder looks like `foo/bar/baz.sia`, you can still upload files named `foo` and `foo/bar`, since the `.sia` prefix will be appended before saving. In other words, the renter skips folders when checking for naming conflicts.

DavidVorick:
What happens if I have a folder named `foo.sia`? Does the naming conflict get discovered?

DavidVorick:
We should have tests for naming conflicts

api/server_helpers_test.go
```diff
@@ -231,6 +232,26 @@ func (st *serverTester) coinAddress() string {
...
+		return err
+	}
+	// mine block
+	st.miner.AddBlock()
+	var hosts ActiveHosts
+	time.Sleep(1 * time.Second)
+	st.getAPI("/renter/hosts/active", &hosts)
+	if len(hosts.Hosts) == 0 {
+		return errors.New("host announcement not seen")
+	}
```
DavidVorick: can we make this a little more robust?

```
st.getAPI()
for i := 0; i < 50 && len(hosts.Hosts) == 0; i++ {
     time.Sleep(50*time.Millisecond)
    st.getAPI()
}
if len == 0 {
    return errors.New()
}

This allows us to bring the average case to a lower time, especially for faster machines, and increases tolerance for slow machines or rare situations that might take a longer time.

File: modules/renter/files.go [hunk 2]
```diff
@@ -63,7 +63,8 @@ func (f *file) chunkSize() uint64 {
 // numChunks returns the number of chunks that f was split into.
 func (f *file) numChunks() uint64 {
 	n := f.size / f.chunkSize()
-	if f.size%f.chunkSize() != 0 {
+	// last chunk may have padding
+	if f.size%f.chunkSize() != 0 || f.size == 0 {
```
DavidVorick: might be worth adding a comment about how this handles 0-byte files.

modules/renter/files_test.go
```diff
@@ -6,6 +6,33 @@ import (
...
+
+		{101, 10, 1, 11}, // padded
+		{101, 10, 2, 6},  // padded
+
+		{10, 100, 1, 1}, // larger piece than file
+		{0, 10, 1, 1},   // 0-length
+	}
+
+	for _, test := range tests {
+		rsc, _ := NewRSCode(test.piecesPerChunk, 1) // can't use 0
```
DavidVorick: Can we upgrade the tests to also mess with the parity?

I'm not completely understanding this test, but I thought that the redundancy was important to the number of chunks?

modules/renter/persist.go
```diff
@@ -195,32 +202,37 @@ func (r *Renter) save() error {
...
-	for _, path := range filenames {
-		// Skip non-sia files.
-		if filepath.Ext(path) != ShareExtension {
-			continue
-		}
-		file, err := os.Open(filepath.Join(r.persistDir, path))
+	// Recursively load all files found in renter directory. Errors
+	// encountered during loading are logged, but are not considered fatal.
+	err := filepath.Walk(r.persistDir, func(path string, info os.FileInfo, err error) error {
+		// These errors should be considered fatal.
```
DavidVorick: why should these errors be considered fatal? This comment is incomplete

modules/renter/persist.go
```diff
@@ -195,32 +202,37 @@ func (r *Renter) save() error {
...
 		if err != nil {
-			// maybe just skip?
 			return err
 		}
+
+		// Skip folders and non-sia files.
+		if info.IsDir() || filepath.Ext(path) != ShareExtension {
```
DavidVorick: Do we want to eventually have support in the renter for empty folders? Seems like a good idea to me.

api/renter_test.go
```diff
@@ -0,0 +1,67 @@
...
+		if filepath.Ext(path) == ".sia" {
+			rel, _ := filepath.Rel(stDir, path)
+			walkStr += rel
+		}
+		return nil
+	})
+	expWalkStr := "renter/foo/bar/test.sia"
+	if walkStr != expWalkStr {
+		t.Fatalf("Bad walk string: expected %v, got %v", expWalkStr, walkStr)
+	}
```
DavidVorick: I think that this is the wrong approach for checking that the test succeeded. Ideally, you'd be able to verfiy without needing to check the internal state, and just use the api.

Would it work to instead query the list of files known to the renter and verify that the file returned has '/` in it?

modules/renter/persist_test.go
```diff
@@ -161,3 +161,68 @@ func TestRenterSaveLoad(t *testing.T) {
...
+		}
+		rel, _ := filepath.Rel(rt.renter.persistDir, path) // strip testdir prefix
+		walkStr += rel
+		return nil
+	})
+	// walk will descend into foo/bar/, reading baz, bar, and finally foo
+	expWalkStr := (f3.name + ".sia") + (f2.name + ".sia") + (f1.name + ".sia")
+	if walkStr != expWalkStr {
+		t.Fatalf("Bad walk string: expected %v, got %v", expWalkStr, walkStr)
+	}
```
DavidVorick: this check makes sense, since we're testing the renter as opposed to the api

DavidVorick:
I would like to see a few more tests added, particularly for corner cases:

+ uploading a file when there is a name conflict
+ uploading a file when there is a name conflict with a folder
+ uploading a 0 byte file
+ uploading a file that's interpreted as having a leading slash

I think there was one more edge case that should be tested, but I've forgotten

modules/renter/files_test.go
lukechampine: redundancy is relevant only with regard to how it affects the number of pieces per chunk, i.e. the number of data pieces. So the parity is being manipulated (just between 1 and 2) via `piecesPerChunk`.

modules/renter/persist.go
lukechampine: empty folders are even trickier than empty files. At the moment we don't support creating folders at all (only creating files within folders), so I think it's fine for now.
Later I suppose we'll add support for folder operations, and decide what makes sense then. As usual, I plan to find out how Dropbox does it.

api/renter_test.go
lukechampine: yeah, I should probably verify that `/renter/files` list the file properly and that `/renter/download/<path>` works.

modules/renter/persist.go
DavidVorick: ok, that's fine with me. I would like to support empty folders in the future, (as well as full file permissions, which might be how we achieve empty folders), but for now it's okay

api/renter_test.go
DavidVorick: :+1:

lukechampine:
The behavior for the last test (uploading a file with a leading slash, e.g. `/renter/upload//foo/bar` is the same as without the slash, except that the nickname contains the leading slash. I was worried that it might create the directory structure at the root level (in `/foo`), but this doesn't happen because the renter calls `filepath.Join(r.persistDir, f.name)`, which removes the double slash in `siaDir/renter//foo/bar`.
Is this the proper behavior? I think it mostly makes sense, but it could cause some confusing behavior (like /renter/upload//foo/bar` overwriting the sia file of `/renter/upload/foo/bar`). Then again, people shouldn't be using leading slashes in the first place...

lukechampine:
Ok, trying to create a file with a folder naming conflict results in an error: `rename foo/bar.sia_temp foo/bar.sia: is a directory`. I think this is satisfactory, especially since folders ending in ".sia" are sure to be very rare. We can address this more robustly when we add first-class folder support.

DavidVorick:
Should return an error for files with a leading slash

api/renter_test.go
```diff
@@ -0,0 +1,124 @@
...
+	if err == renter.ErrDuplicateNickname {
+		t.Fatalf("expected %v, got %v", renter.ErrDuplicateNickname, err)
+	}
+
+	// Upload using nickname that conflicts with folder.
+	err = st.stdPostAPI("/renter/upload/foo/bar", uploadValues)
+	if err == nil {
+		t.Fatal("expecting conflict error, got nil")
+	}
+}
```
DavidVorick: :+1:

PR #896, opened 2016-01-08
Repair offline chunks

lukechampine:
This isn't super sophisticated but it should work.

Tests will be added soon, I'm opening the PR now for immediate feedback.

### DIFF
--- a/modules/renter/repair.go
+++ b/modules/renter/repair.go
@@ -185,21 +185,31 @@ func (f *file) expiringContracts(height types.BlockHeight) []fileContract {
 // offlineChunks returns the chunks belonging to "offline" hosts -- hosts that
 // do not meet uptime requirements. Importantly, only chunks missing more than
 // half their redundancy are returned.
-func (f *file) offlineChunks(activeHosts []modules.HostSettings) map[uint64][]uint64 {
+func (f *file) offlineChunks(hdb hostDB) map[uint64][]uint64 {
 	f.mu.RLock()
 	defer f.mu.RUnlock()
 
-	// helper function for determining if a host is offline, i.e. not in the
-	// set of active hosts.
+	// helper function for determining if a host is offline. A host is
+	// considered offline if it is in allHosts but not activeHosts.
+	allHosts, activeHosts := hdb.AllHosts(), hdb.ActiveHosts()
 	isOffline := func(addr modules.NetAddress) bool {
-		for _, host := range activeHosts {
+		for _, host := range allHosts {
 			if host.IPAddress == addr {
-				return false
+				for _, host := range activeHosts {
+					if host.IPAddress == addr {
+						// host in allHosts and activeHosts
+						return false
+					}
+				}
+				// host in allHosts but not activeHosts
+				return true
 			}
 		}
-		return true
+		// host not in allHosts
+		return false
 	}
 
+	// mark all pieces belonging to offline hosts.
 	offline := make(map[uint64][]uint64)
 	for _, fc := range f.contracts {
 		if isOffline(fc.IP) {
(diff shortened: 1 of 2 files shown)


modules/renter/repair.go
```diff
@@ -227,6 +273,18 @@ func (r *Renter) threadedRepairFile(name string, meta trackedFile) {
 		r.repairChunks(f, handle, badChunks, duration)
 	}
 
+	// repair offline chunks
+	if badChunks := f.offlineChunks(r.hostDB); len(badChunks) != 0 {
```
DavidVorick: should only be repairing if 1/2 the error margin on redundancy is gone.

So if there are 8 parity pieces, repair when 4+ are missing. This is to help keep churn low, and prevent from continually uploading to crappy hosts.

modules/renter/repair.go
```diff
@@ -182,6 +182,52 @@ func (f *file) expiringContracts(height types.BlockHeight) []fileContract {
...
 
+// offlineChunks returns the chunks belonging to "offline" hosts -- hosts that
+// do not meet uptime requirements. Importantly, only chunks missing more than
+// half their redundancy are returned.
+func (f *file) offlineChunks(hdb hostDB) map[uint64][]uint64 {
+	f.mu.RLock()
+	defer f.mu.RUnlock()
+
+	// helper function for determining if a host is offline. A host is
+	// considered offline if it is in allHosts but not activeHosts.
```
DavidVorick: // TODO: once the hostdb has persistence, only 'activeHosts' needs to be checked.

modules/renter/repair.go
```diff
@@ -182,6 +182,52 @@ func (f *file) expiringContracts(height types.BlockHeight) []fileContract {
...
+						return false
+					}
+				}
+				// host in allHosts but not activeHosts
+				return true
+			}
+		}
+		// host not in allHosts
+		return false
+	}
```
DavidVorick: this for loop is potentially large enough to affect performance. AllHosts can be as large as 500 (and I'm thinking about increasing that limit), and ActiveHosts can be as large as 200 (thinking about raising it to 300). 200*500 = 10,000.

So, probably okay, but starting to be on the limit of what is acceptable for nested loops. Once we are no longer checking both all hosts and active hosts, the logic gets a lot cleaner. It's probably fine for now

modules/renter/repair.go
```diff
@@ -182,6 +182,52 @@ func (f *file) expiringContracts(height types.BlockHeight) []fileContract {
...
+
+	// mark all pieces belonging to offline hosts.
+	offline := make(map[uint64][]uint64)
+	for _, fc := range f.contracts {
+		if isOffline(fc.IP) {
```
DavidVorick: oh wait you are doing 100,000 ops per host? =/

allHosts is likely to be full. activeHosts is likely to be 50-100 if we get the type of popularity we're hoping for.

modules/renter/repair.go
DavidVorick: oh nvm this is per chunk

DavidVorick:
You need better order notation I'm thinking. run a benchmark checking 1,000 files with 200 active hosts and 500 all hosts. Those numbers are all high, but represent likely real world numbers. Anything over a second won't fly. Even a second is a really long time, would rather have this loop finish in a few milliseconds.

modules/renter/repair.go
lukechampine: would be pretty simple to convert to a set (`map[modules.NetAddress]struct{}`). Might even be fewer lines of code.

lukechampine:
ok, optimized + tested :like:

modules/renter/repair.go
```diff
@@ -182,6 +182,51 @@ func (f *file) expiringContracts(height types.BlockHeight) []fileContract {
...
+	f.mu.RLock()
+	defer f.mu.RUnlock()
+
+	// helper function for determining if a host is offline. A host is
+	// considered offline if it is in AllHosts but not ActiveHosts. For good
+	// order notation, we covert AllHosts and ActiveHosts to a single map. The
+	// map lookup will return true if the host was in ActiveHosts.
+	//
+	// TODO: once the hostdb is persistent, checking for existence in
+	// ActiveHosts should be sufficient.
```
DavidVorick: I would like to make this a relatively high priority. It's got a bad code smell

PR #897, opened 2016-01-08
Restore support for renaming

lukechampine:


modules/renter/files.go
```diff
@@ -176,30 +176,29 @@ func (r *Renter) FileList() []modules.FileInfo {
...
+	file.name = newName
+	r.saveFile(file)
+	file.mu.Unlock()
 
-		r.save()
-		return nil
+	// Update the entries in the renter.
+	delete(r.files, currentName)
+	r.files[newName] = file
+	r.save()
```
DavidVorick: need to check the errors here

modules/renter/files_test.go
```diff
@@ -125,23 +125,21 @@ func TestRenterDeleteFile(t *testing.T) {
...
+	// Call delete on the previous name.
+	err = rt.renter.DeleteFile("1")
+	if err != ErrUnknownNickname {
+		t.Error("Expected ErrUnknownNickname, got", err)
+	}
+	// Call delete on the new name.
+	err = rt.renter.DeleteFile("one")
+	if err != nil {
+		t.Error(err)
+	}
```
DavidVorick: Need to add a check here that the disk operations correctly cleared the .sia files.

DavidVorick: and really, you should probably add a renter reboot as well, to make sure that the persistence doesn't revive any zombies or drop any renamed files.

modules/renter/files.go
```diff
@@ -176,30 +178,41 @@ func (r *Renter) FileList() []modules.FileInfo {
...
+	if err != nil {
+		return err
+	}
 
-	*/
+	// Delete the old .sia file.
+	// NOTE: proper error handling is difficult here. For example, if the
+	// removal fails, should the entry in r.files be preserved? For now we will
+	// keep things simple, but it is important that our approach feels
+	// intuitive/unsurprising and doesn't put the user's data at risk.
```
DavidVorick: restarting the client will cause it to read in any sia files in the directory structure

PR #898, opened 2016-01-10
Host Revamp

DavidVorick:
I don't know how this blew up so much, it was a lot simpler when I got started. I'm still not done testing, and there are a handful of other features I'd like to add, but I don't see those being ready by v0.5.0. It's time to call it quits on the feature adding until we're ready to start moving towards v1.0. I haven't finished writing tests yet, it's likely that some of the tests will turn up more bugs - some of the fancier logic still isn't checked.

I added some tracking to the RPCs. Now, the host will count each time any of the RPCs are hit, which will give us information about how hosts are being used. This is particularly useful as a natural way to tell if the ports are forwarded. A very low hit count would indicate that the host has low visibility. A high hit count would indicate that the ports are probably forwarded correctly. Currently, there's no way to access these fields from outside the package.

The logger package will now get upset if you call `Write` after you call `Close`, or if you call `Close` twice.

The host is now able to track consensus on the transactions that it has. It can tell if a transaction is not in the consensus set, and will know to resubmit the transaction. It can tell if it has been double spent, and will know to delete the obligation. It can tell if there was a reorg after it submitted a storage proof.

The host now cleanly closes. All goroutines will get to terminate before stuff like the logger is closed, and when `Close` terminates there should be no more threads running in the host. Additionally, exported functions that require host resources will return errors if they are called after the host has been closed.

Host now passes golint, and a travis check has been added that will run golint on the host.

Errcheck has been run on the host, not all offences have been removed due to `defer Closer` type stuff, but a few mistakes were found and corrected. @lukechampine may disagree with some of the error handling, so it's worth a close look + discussion. I'm thinking mostly about checking the errors on `conn.SetDeadline`

Things that could still be done:
+ rpc tracking is not exported
+ bandwidth ratelimiting
+ merkle tree caching
+ allowing edits to a file in random places (requires merkle tree caching)
+ paid bandwidth
+ toggle switch to stop the host from accepting new contracts
+ some of the informational functions could be consolidated, which may make mocking easier


api/host_test.go
```diff
@@ -61,7 +62,7 @@ func TestIntegrationHosting(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	expRevenue := "15928888888855757473"
+	expRevenue := "15307662222190387473"
```
DavidVorick: I don't know why this changed, but I'm guessing some of the buffering constants were edited or something

.travis.yml
```diff
@@ -6,6 +6,7 @@ go:
 install:
   - make dependencies
   - test -z "$(go fmt ./...)"
+  - test -z "$(golint -min_confidence=1.0 ./modules/host)"
```
lukechampine: I don't like this -- `golint` should be treated like `go vet`, meaning it should be run in the Makefile rather than Travis. Otherwise you won't know about a golint failure until you make a PR.

modules/host/announce.go
```diff
@@ -50,6 +50,14 @@ func (h *Host) announce(addr modules.NetAddress) error {
 // arbitrary data, signing the transaction, and submitting it to the
 // transaction pool.
 func (h *Host) Announce() error {
+	// As an exported function that needs resources, Annouce holds the resource
```
lukechampine: typo

lukechampine:
We could opt to say, "Behavior is undefined if Close has been called." Would save a lot of code.

modules/host/announce.go
```diff
@@ -57,14 +65,22 @@ func (h *Host) Announce() error {
...
 
 	return h.announce(addr)
 }
 
-// ForceAnnounce announces using the provided address.
+// AnnounceAddress submits a host announcement to the blockchain to announce a
+// specific address. No checks for validity are performed on the address.
 func (h *Host) AnnounceAddress(addr modules.NetAddress) error {
+	// As an exported function that needs resources, AnnouceAddress holds the
+	// resource lock throughout its operation.
```
lukechampine: typo.
Is this docstring gonna be in every function that uses resources...?

modules/host/download.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/NebulousLabs/Sia/encoding"
+	"github.com/NebulousLabs/Sia/modules"
+	"github.com/NebulousLabs/Sia/types"
+)
+
+const (
+	// tolerableDownloadSize specifies the maximum size that can be downloaded
+	// in a single request. 64MB is chosen because most requests should be at
+	// 4MB exactly and connections can be unstable beyond 200MB.
+	tolerableDownloadSize = 1 >> 26
```
lukechampine: LOL
`1 >> 26` is 0, buddy.

modules/host/download.go
```diff
@@ -0,0 +1,89 @@
...
+			return errors.New("request exceeds file bounds")
+		}
+		if request.Length > tolerableDownloadSize {
+			return errors.New("cannot download provided length")
+		}
+
+		// Write segment to conn.
+		err := conn.SetDeadline(time.Now().Add(5 * time.Minute)) // sufficient to transfer 4 MB over 100 kbps
+		if err != nil {
+			return err // TODO: Is this the correct way to handle this error?
```
lukechampine: reading the `net` code, it looks like an error here indicates some sort of OS-level error when accessing conn's file descriptor. So, unlikely, but theoretically possible, and I imagine it would mean any future I/O on the conn would also fail.

modules/host/download.go
DavidVorick: the download test is not written yet

DavidVorick: the API tests are passing. huh

modules/host/host.go
```diff
@@ -80,13 +73,32 @@ type Host struct {
...
+	atomicMalformedCalls uint64
+	atomicDownloadCalls  uint64
+	atomicRenewCalls     uint64
+	atomicReviseCalls    uint64
+	atomicSettingsCalls  uint64
+	atomicUploadCalls    uint64
+
+	// The resource lock is held by threaded functions for the duration of
+	// their operation. Functions should grab the resource lock as a read lock
+	// unless they are planning on manipulating the 'closed' variable.
```
lukechampine: why is `resourceLock` used for `closed` rather than `mu`?

modules/host/network.go
```diff
@@ -20,57 +21,116 @@ func (h *Host) initNetworking(address string) error {
...
-// handleConn handles an incoming connection to the host, typically an RPC.
-func (h *Host) handleConn(conn net.Conn) {
-	defer conn.Close()
+// threadedHandleConn handles an incoming connection to the host, typically an
+// RPC. threadedHandleConn is responsible for releasing a readlock on
+// host.resourceLock when all communications have completed.
+func (h *Host) threadedHandleConn(conn net.Conn) {
+	// All threaded functions are called holding the close lock, and are
+	// expected to keep it until they no longer need access to the host's
+	// resources.
```
lukechampine: wat??? why? This seems like a dangerous convention. Locking->Unlocking across function boundaries = bad plan

modules/host/network.go
```diff
@@ -20,57 +21,116 @@ func (h *Host) initNetworking(address string) error {
...
+func (h *Host) threadedListen() {
+	// All threaded are called holding a readlock, and must release the
+	// readlock upon terminating.
+	defer h.resourceLock.RUnlock()
+
+	// If the host has closed, some of the necessary resources will not be
+	// available, and therefore the host should not be receiving connections.
+	if h.closed {
+		return
+	}
```
lukechampine: isn't this impossible? `threadedListen` is only called on startup

modules/host/obligations.go
```diff
@@ -0,0 +1,264 @@
...
+
+	// confirmationRequirement is the number of blocks that the host is going
+	// to wait before assuming that a storage proof has successfully been
+	// confirmed by the blockchain.
+	confirmationRequirement = 12
+
+	// Booleans to indicate that a contract obligation has been successful or
+	// unsuccessful.
+	obligationSucceeded = true
+	obligationFailed    = false
```
lukechampine: these should be an enum, not a const

modules/host/obligations.go
```diff
@@ -0,0 +1,264 @@
...
+	co.OriginConfirmed = false
+	co.ProofConfirmed = false
+
+	// If there is no revision, then the final revision counts as being
+	// confirmed. Otherwise, the revision should not be considered as
+	// confirmed.
+	if co.hasRevision() {
+		co.RevisionConfirmed = false
+	} else {
+		co.RevisionConfirmed = true
```
lukechampine: `co.RevisionConfirmed = !co.hasRevision()`

modules/host/obligations.go
```diff
@@ -0,0 +1,264 @@
...
+
+// revisionNumber returns the operating revision number of the obligation.
+func (co *contractObligation) revisionNumber() uint64 {
+	if co.hasRevision() {
+		return co.RevisionTxn.FileContractRevisions[0].NewRevisionNumber
+	}
+	return co.OriginTxn.FileContracts[0].RevisionNumber
+}
+
+// txnsConfirmed indiactes whether the file contract and its latest revision
```
lukechampine: typo

modules/host/obligations.go
```diff
@@ -0,0 +1,264 @@
...
+// have been seen on the blockchain.
+func (co *contractObligation) txnsConfirmed() bool {
+	if !co.OriginConfirmed {
+		return false
+	}
+	if co.hasRevision() && !co.RevisionConfirmed {
+		return false
+	}
+	return true
```
lukechampine:
```go
if co.hasRevision() {
    return co.RevisionConfirmed
}
```

modules/host/obligations.go
```diff
@@ -0,0 +1,264 @@
...
+	// Add the obligation to the list of host obligations.
+	h.obligationsByID[co.ID] = co
+
+	// The host needs to make sure that the file contract has made it into the
+	// blockchain, and to make sure that a storage proof is submitted on time.
+	confirmHeight := h.blockHeight + resubmissionTimeout
+	h.actionItems[confirmHeight] = append(h.actionItems[confirmHeight], co)
+
+	// Update the statistics.
+	h.anticipatedRevenue = h.anticipatedRevenue.Add(co.value()) // Output at index 1 alone belongs to host.
```
lukechampine: I don't understand this comment

modules/host/obligations.go
```diff
@@ -0,0 +1,264 @@
...
+	h.spaceRemaining = h.spaceRemaining - int64(co.fileSize())
```
lukechampine: `-=`?

modules/host/obligations.go
```diff
@@ -0,0 +1,264 @@
...
+	if build.DEBUG && len(revisionTransaction.FileContractRevisions) != 1 {
+		panic("cannot revise obligation without a file contract revision")
+	}
+	obligation, exists := h.obligationsByID[revisionTransaction.FileContractRevisions[0].ParentID]
+	if build.DEBUG && !exists {
+		panic("cannot revise obligation - obligation not found")
+	}
+
+	// Update the host's statistics, and determine whether the host needs a new
+	// action item due to the revision window moving.
```
lukechampine: I don't see any determining here. To me that implies a branch.

modules/host/obligations.go
lukechampine: alternatively, define separate functions for removing successful and unsuccessful obligations

modules/host/persist.go
```diff
@@ -25,34 +26,42 @@ var persistMetadata = persist.Metadata{
...
+	Revenue     types.Currency
+
+	// RPC Tracking.
+	ErroredCalls   uint64
+	MalformedCalls uint64
+	DownloadCalls  uint64
+	RenewCalls     uint64
+	ReviseCalls    uint64
+	SettingsCalls  uint64
+	UploadCalls    uint64
```
lukechampine: grouping these into a "metrics" struct is probably a good idea -- could be used in the Host type as well.

modules/host/persist.go
```diff
@@ -142,6 +177,11 @@ func (h *Host) establishDefaults() error {
 	}
 	h.spaceRemaining = h.settings.TotalStorage
 
+	// Reduce the window size when testing.
+	if build.Release == "testing" {
+		h.settings.WindowSize = testingWindowSize
+	}
```
lukechampine: alternatively, define `testingWindowSize` conditionally based on `build.Release`, the way the renter does with its defaults. That way you don't need to worry about the discrepancy between `h.settings.WindowSize` and `testingWindowSize`.

modules/host/update.go
```diff
@@ -98,11 +107,127 @@ func (h *Host) threadedCreateStorageProof(obligation *contractObligation) {
...
+	}
+
+	// Check for indications that the obligation should be deleted.
+	if !co.txnsConfirmed() && co.windowStart() < h.blockHeight {
+		// The final file contract revision has not made it into the
+		// blockchain, and the window for submitting it has closed. The
+		// obligation can be deleted.
+		h.removeObligation(co, obligationFailed)
+		return
+	}
```
lukechampine: seems like a giant `switch` might be more appropriate here

modules/host/update.go
```diff
@@ -98,11 +107,127 @@ func (h *Host) threadedCreateStorageProof(obligation *contractObligation) {
...
+		// submitting the proof has closed. The obligation can be deleted.
+		h.removeObligation(co, obligationFailed)
+		return
+	}
+	if co.proofConfirmed() && co.windowEnd()+confirmationRequirement <= h.blockHeight {
+		// The storage proof has been successful, and has enough confirmations
+		// to be considered stable. Therefore the obligation can be deleted.
+		h.removeObligation(co, obligationSucceeded)
+		return
+	}
```
lukechampine: a giant `switch` might be more appropriate for this function, which is essentially a bunch of conditionals

modules/host/update.go
```diff
@@ -111,30 +236,85 @@ func (h *Host) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+			}
+		}
+
+		// Adjust the height of the host. The host height is initialized to
+		// zero, but the genesis block is actually height zero. For the genesis
+		// block only, the height will be left at zero.
+		//
+		// Checking the height here eliminates the need to initialize the host
+		// to and underflowed types.BlockHeight.
+		if h.blockHeight != 0 || cc.AppliedBlocks[len(cc.AppliedBlocks)-1].ID() != types.GenesisBlock.ID() {
```
lukechampine: change to `block.ID()`

modules/host/upload.go
```diff
@@ -13,53 +16,447 @@ import (
...
-	err := encoding.ReadObject(conn, &contractID, crypto.HashSize)
+const (
+	// maxRevisionSize is the maximum number of bytes added in a single revision
+	maxRevisionSize = 100e6 // 100 MB
+)
+
+var (
+	// HostCapacityErr indicates that a host does not have enough room on disk
+	// to accept more files.
+	HostCapacityErr = errors.New("host is at capacity and cannot take more files")
```
lukechampine: I thought we didn't want to expose this error

modules/host/upload.go
```diff
@@ -13,53 +16,447 @@ import (
...
+	return nil
+}
+
+// managedRPCUpload is an RPC that negotiates a file contract. Under the new
+// scheme, file contracts should not initially hold any data.
+func (h *Host) managedRPCUpload(conn net.Conn) error {
+	// Check that the host has grabbed an address from the wallet.
+	h.mu.RLock()
+	uh := h.settings.UnlockHash
+	h.fileCounter++ // Harmless to increment the file counter in the event of an error.
```
lukechampine: does this number get reported to the user anywhere? Or is it just used to name files?

modules/host/upload.go
```diff
@@ -13,53 +16,447 @@ import (
...
+			}
+
+			// check revision against original file contract
+			h.mu.RLock()
+			err = h.considerRevision(revTxn, obligation)
+			h.mu.RUnlock()
+			if err != nil {
+				// There is nothing that can be done if there is an error while
+				// writing to a connection.
+				_ = encoding.WriteObject(conn, err.Error())
```
lukechampine: should probably `return` instead of `continue` if the write fails

modules/host/upload.go
```diff
@@ -13,53 +16,447 @@ import (
...
+				continue // don't terminate loop; subsequent revisions may be okay
+			}
+
+			// indicate acceptance
+			if err := encoding.WriteObject(conn, modules.AcceptResponse); err != nil {
+				return errors.New("couldn't write acceptance: " + err.Error())
+			}
+
+			// read piece
+			// TODO: simultaneously read into tree and file
```
lukechampine: this TODO is probably a bad idea. If something goes wrong you'd need to truncate the file. However, it may be necessary if revisions get larger than a few dozen MB.

modules/host/upload_test.go
```diff
@@ -0,0 +1,202 @@
...
+	for _, ob := range ht.host.obligationsByID {
+		if ob.fileSize() >= datasize {
+			return nil
+		}
+	}
+	return errors.New("ht.uploadFile: upload failed")
+}
+
+// TestRPCUPload attempts to upload a file to the host, adding coverage to the
+// upload function.
```
lukechampine: `UPload` -> `Upload` -- and is it necessary to specify that a test adds coverage? I think my test docstrings are usually of the form "TestFoo tests the foo function by barring the baz."

modules/transactionpool.go
```diff
@@ -59,3 +59,18 @@ type TransactionPool interface {
 	// transaction pool changes, and should not subscribe to both.
 	TransactionPoolSubscribe(TransactionPoolSubscriber)
 }
+
+// ConsensusConflict implements the error interface, and indicates
```
lukechampine: incomplete docstring

persist/log.go
```diff
@@ -3,28 +3,65 @@ package persist
...
+	"github.com/NebulousLabs/Sia/build"
 )
 
+// logFileWrapper wraps a log file to perform sanity checks on the Write and
+// Close functions. An error will be returned if Close is called multiple times
+// or if Write is called after Close has been called.
+type logFileWrapper struct {
+	closed bool // Flag is set after Close is called.
+	file   *os.File
+}
```
lukechampine: why is this separate type used instead of modifying the existing logger type?

.travis.yml
DavidVorick: yeah. I'd rather use the Makefile, and the `pkgs` variable, but then you have to have golint passing for every package. Otherwise, it's going to check every package every time you run `make test-long`, which could cause slowdown especially if you are using `run=` to minimize the amount of time.

It's a temporary solution until we are golint compliant in all of our packages. Then I agree, it should be in the Makefile and not the travis file.

modules/host/announce.go
DavidVorick: wrote it a few times, but really I should talk about it in the conventions documentation and then remove it from here.

Sounds like we haven't finished determining exactly what this usage should look like, so I'll hold off on writing more conventions stuff

modules/host/download.go
DavidVorick: what's the best way to limit how many times the deadline can be extended? Finite `for` loop?

modules/host/host.go
DavidVorick: if you are holding `mu` alone it's not sufficient to guarantee that no new threads are going to be trying to grab resources - the resource lock is often grabbed before the thread is even initiated (though in hindsight, that might not be necessary). If the closing function is only holding `mu`, it might miss that another thread has grabbed the resource lock but not `mu`.

Basically, to guarantee safety you need to use the resource lock, and once you are using that there's no reason to also grab `mu`.

modules/host/network.go
DavidVorick: yeah thinking through it a bit more I don't think it's necessary for the lock to be grabbed across function boundaries.

But I'm also [tired] fairly confident that a waitgroup is different than what's happening here. In this case, we want to know (for close) when everyone has released all the resources, but we don't want them to grab them again after that.

Actually, I'm not so sure. A wait group might be enough. But that confuses the locking stuff.

DavidVorick: removed the cross-function-boundaries stuff

modules/host/network.go
DavidVorick: currently, yes. but if the host hits the bandwidth cap or something it might make sense to start toggling this. In the meantime, I'll turn this into a sanity check

modules/host/obligations.go
DavidVorick: Why enum? And why define separate functions?

I just wanted to make the code a bit clearer:

`myFunction(obligation, catsAndDogs, true)` ==> `myFunction(obligation, catsAndDogs, obligationSucceeded)`

modules/host/obligations.go
DavidVorick: I find my way to be more intuitive. I've added more comments.

If the first txn isn't confirmed, then the txns aren't confirmed
if the second txn exists but isn't confirmed, then the txns aren't confirmed
otherwise, the txns are confirmed.

modules/host/obligations.go
DavidVorick: oh. there used to be determining. Now it's more deterministic

modules/host/persist.go
DavidVorick: too much work for this PR, but in general I'm in favor of finding better ways to alter the constants for release vs dev vs testing.

Just don't want to think anymore tonight >.<

modules/host/update.go
DavidVorick: perhaps, though I'm not a huge fan of the idea of inserting so much logic and comments into a switch statement.

modules/host/upload.go
DavidVorick: comments noticeably more sparse later in the PR :P

DavidVorick: also you wrote the original logic here, I'll change it to return though

persist/log.go
DavidVorick: needed to implement the `Write` function.

The alternative was to implement the `Write` function on the type holding the log.Logger, but then when initializing the log.Logger you'd be passing the struct containing it as the interface, and that just seemed wrong.

So instead I opted for a new type on which to implement `Write`. And since I had to do that, I figured it made sense to implement close on it as well, so it could add some safety features. I even considered having `newLogFileWrapper` open the file itself, and still might do that, but I haven't convinced myself that it's the best strategy yet.

But I think it's the right thing to do.

.travis.yml
lukechampine: just use a separate variable (or no variable at all) instead of pkgs. It may be a while before all packages pass golint.

persist/log.go
lukechampine: right right, that makes sense. I guess I was pretty tired as well.

it seems like this type only exists to enable a specific sanity check though. Didn't you originally implement this to prevent a race condition? A race condition should be just as visible as a panic.

I'm actually starting to question whether we really need to close the log file. What does that do for us? It's only relevant during testing, and we certainly don't have to worry about opening too many file handles or something. Our primary focus should be shutting down all goroutines, because those can have cross-test side effects.
More importantly, using an *os.File in the persist logger makes it harder to mock. One of the advantages of a standard `log.Logger` is that you can log to an in-memory Writer, like a `bytes.Buffer`.

lukechampine:
My recommendation:
Drop the Close from persist.Logger. This also removes the need to worry about cleanly shutting down the entire host before closing the logger. As a result, you don't need `resourceLock` or `closed`, and the code is greatly simplified.
I think it's worth exploring "clean shutdown" in more depth, but I don't think this is the right time or place to do it. I want to get this release out ASAP, which means not getting bogged down in quarrels about how modules should shutdown. That functionality is not a priority for this release.

.travis.yml
DavidVorick: yeah but then you have to set that variable every time you run make if you want to avoid the lint call.

DavidVorick:
Clean shutdown is implemented in the host. No reason to back out now. And, clean shutdown is definitely something that should be pursued. Right now, when a module closes, it's frequently still got several goroutines that are trying to access closed resources. It might be in the middle of a communication like a file repair, and you definitely want the shutdown process to wait until everything has finished.

lukechampine:
>Clean shutdown is implemented in the host. No reason to back out now.

That argument is only valid if you assume that I agree with your decision+specific implementation, which I'm not sure I do. I think it's bad development practice to shoehorn new features into a PR like this, where there's strong pressure for me to merge in order to get the release out.

>Right now, when a module closes, it's frequently still got several goroutines that are trying to access closed resources. It might be in the middle of a communication like a file repair, and you definitely want the shutdown process to wait until everything has finished.

It's a complex topic, and the implementation may wind up looking different for different modules. That's why I'm not entirely comfortable accepting the implementation here -- it sets a precedent for how future modules should shutdown.

My current thoughts:
- The general shutdown pattern is to give each module a `closeChan`, and signal shutdown by closing the channel. All goroutines should `select` on this channel, and exit upon receiving `nil` from the `closeChan` (see https://blog.golang.org/pipelines).
- `Close` should block until all the goroutines have exited. To facilitate this, pass a `sync.WaitGroup` to each goroutine, and increment it by 1 every time a goroutine is spawned. This pattern is outlined [here](http://www.hydrogen18.com/blog/stopping-it-all-in-go.html) and [here](http://adampresley.com/2015/02/16/waiting-for-goroutines-to-finish-running-before-exiting.html).
- Finally, I am of the opinion that we shouldn't have to worry about the module being used after `Close` has been called. This can be avoided by managing the shutdown process properly -- namely, closing in reverse order (e.g. wallet, then consensus, not the other way around). We can enforce this more strictly by wiping the resources of the struct in `Close` (e.g. `host.obligationsByID = nil`), such that any future calls attempting to modify those resources will panic. Alternatively, we could lock the module's mutex at the end of the `Close` call, preventing anyone else from ever grabbing it.

These practices are easy to follow, but they leave the most important question unanswered: "For each goroutine, assuming the close signal has been sent, when/how should the goroutine actually exit?" This answer depends on the nature of the goroutine's work. For example, if the goroutine is trying to forward a port, it seems reasonable that it should give up as soon as the signal is received. But if it's in the middle of negotiating a contract, or submitting a transaction, the answer is less clear. Furthermore, goroutines often don't receive the close signal immediately; they must complete their current code path before looping and reading the channel. This can be a problem if the code path takes a long time. How long is it reasonable to make the user wait after they have closed the program? I don't know. I think it's something we'll have to address on a case-by-case basis.

PR #899, opened 2016-01-10
Renter testing + fixes

lukechampine:
- the hostDB can communicate with v0.4.8 hosts again. This was broken in 02bddf9b481fd85ed0dbe12dccacdc659d3dfeb5.
- siac was using POST for /renter/download (now GET)
- repair loop now runs in parallel. It spawns a thread for each file. This is probably doesn't scale well. ~~I'm working on capping it to n threads.~~ (done)
- `threadedRepairFile` checks that there are chunks/contracts in need of repair before opening the file handle.


modules/renter/persist_test.go
```diff
@@ -149,17 +193,6 @@ func TestRenterSaveLoad(t *testing.T) {
...
-	// Corrupt a renter file and try to reload it.
-	err = ioutil.WriteFile(filepath.Join(rt.renter.persistDir, "corrupt"+ShareExtension), []byte{1, 2, 3}, 0660)
-	if err != nil {
-		t.Fatal(err)
-	}
-
-	err = rt.renter.load()
-	if err == nil {
-		t.Fatal("expected error, got nil")
-	}
```
DavidVorick: Any reason you removed the corruption test?

modules/renter/repair_test.go
```diff
@@ -3,25 +3,72 @@ package renter
...
+	// used to simulate real-world conditions
+	delay    time.Duration // transfers will take this long
+	failRate int           // transfers will randomly fail with probability 1/failRate
+
+	sync.Mutex
+}
+
+func (h *testHost) Address() modules.NetAddress  { return h.ip }
+func (h *testHost) EndHeight() types.BlockHeight { return 0 }
+func (h *testHost) Close() error                 { return nil }
```
DavidVorick: Docstrings

modules/renter/repair_test.go
```diff
@@ -3,25 +3,72 @@ package renter
...
+
+func (h *testHost) ContractID() types.FileContractID {
```
DavidVorick: Docstring

modules/renter/repair_test.go
```diff
@@ -48,27 +98,48 @@ func TestRepair(t *testing.T) {
...
+	/*
+		for i, h := range hosts {
+			host := h.(*testHost)
+			pieces := 0
+			for _, p := range host.pieceMap {
+				pieces += len(p)
+			}
+			t.Logf("Host #: %d\tDelay: %v\t# Pieces: %v\t# Chunks: %d", i, host.delay, pieces, len(host.pieceMap))
+		}
+	*/
```
DavidVorick: Any reason that this is commented out?

modules/renter/persist_test.go
lukechampine: it no longer triggers an error; corrupted .sia files are logged, but don't cause `load` to fail.

modules/renter/repair_test.go
lukechampine: I used it to assess how efficient the repair function is. Might be useful in the future, but if you leave it uncommented it will add lots of logging output to `make test`, which I didn't want.

DavidVorick: Need to add a comment explaining that

PR #903, opened 2016-01-11
API changes

lukechampine:
phew

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -259,9 +259,9 @@ Gateway
 
 Queries:
 
-* /gateway               [GET]
-* /gateway/add/{addr}    [POST]
-* /gateway/remove/{addr} [POST]
+* /gateway                     [GET]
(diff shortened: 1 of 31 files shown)


doc/API.md
```diff
@@ -619,97 +623,106 @@ struct {
...
-'path' is the nickname that will be used to reference the file.
+'siapath' is the location where the file will reside in the renter.
+
+'source' is the location on disk of the file being uploaded.
+
+'duration' is the number of blocks for which the file will be available. If
+the renew parameter is true, this parameter will be ignored.
 
-'source' is the path to the file to be uploaded.
+'renew' indicates whether the file's contracts should be automatically renewed
```
Mingling94: How does this get cancelled? Should the UI provide both options and if so have a default duration or renew by default?

lukechampine: no way to cancel as of yet :/
soon the renter will added routes for configuring whether a file should be renewed/repaired.

PR #905, opened 2016-01-11
Host Touchups

DavidVorick:
I expanded a test and moved around how some of the initialization happens. This is setting the stage for some compatibility changes that are coming in my next PR. I wanted to make these changes in their own PR, since they don't actually involve any compatibility. The next PR will be pure compatibility.

### DIFF
--- a/modules/host/host_test.go
+++ b/modules/host/host_test.go
@@ -199,7 +199,7 @@ func TestHostInitialization(t *testing.T) {
 		t.Fatal(err)
 	}
 	if bht.host.blockHeight != 1 {
-		t.Fatal("block height did not increase correctly after first block mined")
+		t.Fatal("block height did not increase correctly after first block mined:", bht.host.blockHeight, 1)
 	}
 }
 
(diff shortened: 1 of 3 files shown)


modules/host/host_test.go
```diff
@@ -283,6 +283,32 @@ func TestSetAndGetSettings(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	rebootHost, err := New(ht.cs, ht.tpool, ht.wallet, ":0", filepath.Join(ht.persistDir, modules.HostDir))
+	if err != nil {
+		t.Fatal(err)
+	}
+	rebootSettings := rebootHost.Settings()
+	if settings.TotalStorage != rebootSettings.TotalStorage {
+		t.Error("settings GET did not return updated value")
```
lukechampine: what's with the `GET`?

modules/host/persist.go
```diff
@@ -16,8 +16,8 @@ const (
 	logFile = modules.HostDir + ".log"
 )
 
-// Variables indicating the metadata header and the version of the data that
-// has been saved to disk.
+// persistMetadata is the header that gets written to the persist file, and is
+// used to recognize other persisit files.
```
lukechampine: typo

modules/host/host_test.go
DavidVorick: GET is a typo

PR #906, opened 2016-01-11
Host compatibility

DavidVorick:
Plus some typo corrections from previous PRs. Minimal changes were needed to the actual logic to get compatibility working, but the compatibility file itself is kind of long. Some logic is required to allow the host to scan the consensus set and look for contracts and revisions before moving on.

### DIFF
--- a/modules/host/persist_compat.go
+++ b/modules/host/persist_compat.go
@@ -1,0 +1,142 @@
+package host
+
+import (
+	"os"
+	"path/filepath"
+
+	"github.com/NebulousLabs/Sia/build"
+	"github.com/NebulousLabs/Sia/crypto"
+	"github.com/NebulousLabs/Sia/modules"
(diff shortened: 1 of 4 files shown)


modules/host/persist_compat.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	err = h.initConsensusSubscription()
+	if err != nil {
+		return err
+	}
+
+	// Remove all obligations that have not had their origin transactions
+	// confirmed on the blockchain, and add action items for the rest.
+	for _, uo := range upgradedObligations {
+		if !uo.OriginConfirmed {
```
lukechampine: I don't see this field being set anywhere in `h.loadCompat04Obligations`

DavidVorick: it gets set in `initConsensusSubscription` if the file contract is found while scanning the consensus set

PR #909, opened 2016-01-12
move atomic fields in host

DavidVorick:
atomic fields need to go at the top of the struct for compatibility with
32bit architectures.


lukechampine:
Would be nice to have 32-bit testing. Could probably be set up with some effort, but it would double the build time.

PR #910, opened 2016-01-12
upgrade travis.ci

DavidVorick:
upgrade to go1.5.2, run build tests on mac.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,11 @@
 language: go
 
+os:
+  - linux
+  - osx
+
 go:
-  - 1.5.1
+  - 1.5.2
 
 install:
   - make dependencies


lukechampine:
this will double the build time, correct?

DavidVorick:
yes, is that a problem?

lukechampine:
slow builds are a problem, yeah.
not a horrible problem, but certainly an inconvenience. No one likes waiting for Travis.
But no harm in merging it for now, just to see how much slower it actually is.

DavidVorick:
I would be more than happy to give up build times to get coverage for every architecture covered in `make xc`. I think something that Bitcoin does is run full tests every night or something like that, and PRs only get partial tests. Then, if a PR breaks something, you still find out later without needing to wait forever on each PR.

But I'm pretty strongly behind the idea that a 60+ minute build time is acceptable if it means we can catch windows specific and 32bit specific bugs.

lukechampine:
I would support those builds for sure, but a 64-bit osx build seems kinda pointless. I don't recall ever encountering any osx-specific bugs.

DavidVorick:
We definitely have in the UI, but I don't think we've seen anything in golang repos yet. Nonetheless I do believe we have dependencies that run different code depending on the operating system (bolt, namely, but I think another one or two as well). That to me is enough to justify a matrix build on osx.

lukechampine:
I set Travis to run the builds in parallel, slowdown should be minimal :+1:

PR #912, opened 2016-01-13
Export host fields

DavidVorick:


api/host.go
```diff
@@ -24,16 +24,26 @@ type (
...
 		StorageRemaining int64          `json:"storageremaining"`
 		UpcomingRevenue  types.Currency `json:"upcomingrevenue"`
+
+		RPCErrorCalls        uint64 `json:"rpcerrorcalls"`
+		RPCUnrecognizedCalls uint64 `json:"rpcunrecognizedcalls"`
+		RPCDownloadCalls     uint64 `json:"rpcdownloadcalls"`
+		RPCRenewCalls        uint64 `json:"rpcrenewcalls"`
+		RPCReviseCalls       uint64 `json:"rpcrevisecalls"`
+		RPCSettingsCalls     uint64 `json:"rpcsettingscalls"`
+		RPCUploadCalls       uint64 `json:"rpcuploadcalls"`
```
lukechampine: why is this not a `modules.HostRPCTracking`?

modules/host.go
```diff
@@ -62,6 +62,18 @@ type (
 		UnlockHash   types.UnlockHash  `json:"unlockhash"`
 	}
 
+	// HostRPCTracking reports the quantity of each type of rpc call that has
+	// been made to the host.
+	HostRPCTracking struct {
```
lukechampine: can we rename `Tracking` to `Metrics`?

modules/host/host.go
```diff
@@ -66,13 +66,13 @@ var (
...
-	atomicReviseCalls    uint64
-	atomicSettingsCalls  uint64
-	atomicUploadCalls    uint64
+	atomicErroredCalls      uint64
+	atomicUnrecognizedCalls uint64
+	atomicDownloadCalls     uint64
+	atomicRenewCalls        uint64
+	atomicReviseCalls       uint64
+	atomicSettingsCalls     uint64
+	atomicUploadCalls       uint64
```
lukechampine: should be able to use a `modules.HostRPCTracking` here as well

api/host.go
DavidVorick: I figured that it was better to have a flat status struct. In general I think flat data structures are superior

modules/host.go
DavidVorick: Ok

modules/host/host.go
DavidVorick: maybe. But then you don't get to call the fields `atomic`. And your top level struct isn't flat, which I think makes for much cleaner code within a function.

`atomic.AddUint64(&h.atomicErroredCalls)`
`atomic.AddUint64(&h.atomicRPCTracking.ErroredCalls)`

I'd prefer the flat struct and some extra copying I think.

DavidVorick:
renamed the tracking fields. I want to leave the structs flat as opposed to using modules.HostRPCTracking inside of host.Host and api.HostGET

api/host.go
lukechampine: why
I doubt it makes any difference performance-wise. Grouping objects makes them easier to understand and manipulate.

PR #917, opened 2016-01-14
Uploading bug fixes

lukechampine:
Commits are pretty self-explanatory.

### DIFF
--- a/modules/renter/hostdb/scan.go
+++ b/modules/renter/hostdb/scan.go
@@ -138,8 +138,10 @@ func (hdb *HostDB) threadedProbeHosts() {
 
 		// Now that network communication is done, lock the hostdb to modify the
 		// host entry.
-		hdb.mu.Lock()
-		{
+		func() {
+			hdb.mu.Lock()
+			defer hdb.mu.Unlock()
+
 			// Regardless of whether the host responded, add it to allHosts.
 			if _, exists := hdb.allHosts[hostEntry.NetAddress]; !exists {
 				hdb.allHosts[hostEntry.NetAddress] = hostEntry
@@ -148,8 +150,7 @@ func (hdb *HostDB) threadedProbeHosts() {
 			// If the scan was unsuccessful, decrement the host's reliability.
 			if err != nil {
 				hdb.decrementReliability(hostEntry.NetAddress, UnreachablePenalty)
-				hdb.mu.Unlock()
-				continue
+				return
 			}
 
 			// Update the host settings, reliability, and weight. The old NetAddress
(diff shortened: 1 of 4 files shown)


DavidVorick:
I'm increasingly in favor of long term host relationships. It's a new and somewhat awkward paradigm for cryptocurrency users, but really isn't any different from paying $10/mo for dropbox - except in Sia you get back what you don't use.

modules/renter/hostdb/scan.go
```diff
@@ -153,11 +159,9 @@ func (hdb *HostDB) threadedProbeHosts() {
...
-			// been reached, add the host to the database.
-			_, exists1 := hdb.activeHosts[hostEntry.NetAddress]
-			_, exists2 := hdb.allHosts[hostEntry.NetAddress]
-			if !exists1 && exists2 && len(hdb.activeHosts) < MaxActiveHosts {
+			// If 'MaxActiveHosts' has not been reached, add the host to the
+			// activeHosts tree.
+			if _, exists := hdb.activeHosts[hostEntry.NetAddress]; !exists && len(hdb.activeHosts) < MaxActiveHosts {
 				hdb.insertNode(hostEntry)
 			}
 		}
```
DavidVorick: this locking scope should probably be a function, it's pretty large and has conditions where the lock needs to be released inside the scope. I think a safer construction would be the 

```
func() {
    lock
    defer unlock
}()
```

pattern. there's a lot of code here.

DavidVorick: same deal in `threadedScan()`

modules/renter/hostdb/upload.go
```diff
@@ -201,17 +202,36 @@ outer:
 		}
 	}
 
-	// form new contracts from randomly-picked nodes
+	// Also exclude hosts on the pool's blacklist, and hosts we're already
```
DavidVorick: it's awkward to start this comment with also, I'm not sure what the previous thought was.

Probably sufficient to `Exclude...`

DavidVorick: oh, exclude is an input. I see, but I still feel like the comment is awkward

modules/renter/hostdb/upload.go
```diff
@@ -201,17 +202,36 @@ outer:
...
 	for _, host := range randHosts {
 		contract, err := p.hdb.newContract(host, p.filesize, p.duration)
 		if err != nil {
+			p.hdb.log.Printf("couldn't form contract with %v: %v", host.NetAddress, err)
+			p.blacklist = append(p.blacklist, host.NetAddress)
 			continue
 		}
 		hu, err := p.hdb.newHostUploader(contract)
 		if err != nil {
+			p.hdb.log.Printf("couldn't create uploader for %v: %v", host.NetAddress, err)
```
DavidVorick: a log here is probably not needed, unless you are unable to make uploaders with anybody or some fractional amount of them.

DavidVorick:
All of my comments are nits, logic looks good. I'm glad the blacklist has been added.

I am worried about the renter making uploads before the full range of hosts have been scanned.

PR #918, opened 2016-01-14
Host off switch

DavidVorick:
This PR is not complete, just wanted to post it for review. Still needs testing, API support, and siac support.

Commit messages are pretty strong.

### DIFF
--- a/modules/host/upload.go
+++ b/modules/host/upload.go
@@ -252,11 +252,20 @@ func (h *Host) managedNegotiateContract(conn net.Conn, filesize uint64, merkleRo
 // scheme, file contracts should not initially hold any data.
 func (h *Host) managedRPCUpload(conn net.Conn) error {
 	// Check that the host has grabbed an address from the wallet.
-	h.mu.RLock()
+	h.mu.Lock()
+	acceptingContracts := h.acceptingContracts
 	uh := h.settings.UnlockHash
 	h.fileCounter++ // Harmless to increment the file counter in the event of an error.
 	filename := filepath.Join(h.persistDir, strconv.Itoa(int(h.fileCounter)))
-	h.mu.RUnlock()
+	h.mu.Unlock()
+
+	// Check that the host is accepting new file contracts.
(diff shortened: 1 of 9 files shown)


DavidVorick:
```
--- FAIL: TestIntegrationWalletTransactionGETid (0.01s)

	wallet_test.go:245: provided encryption key is incorrect
```

This needs to be inspected. It's nondeterministic.

DavidVorick:
```
--- FAIL: TestReceiveConsensusSetUpdate (0.48s)

	update_test.go:83: hostdb should have a host after getting a host announcement transcation
```
I've seen this nondeterministic error a few times as well.

DavidVorick:
```
--- FAIL: TestIntegrationRenewing (7.87s)

	host_test.go:114: 0

	host_test.go:115: uploading has failed
```

Another failure. Hmmm, I hope there's a common thread to all of these.

lukechampine:
Why not treat `AcceptingContracts` like any other HostSettings field? i.e., to toggle it on or off you would go through the existing `/host [POST]` call, with `/host?acceptingcontracts=false`.
It's not as user-friendly as a separate toggle, but that shouldn't matter since UIs can easily make it a separate call if so desired.

modules/host/upload.go
```diff
@@ -252,11 +252,20 @@ func (h *Host) managedNegotiateContract(conn net.Conn, filesize uint64, merkleRo
 // scheme, file contracts should not initially hold any data.
 func (h *Host) managedRPCUpload(conn net.Conn) error {
 	// Check that the host has grabbed an address from the wallet.
-	h.mu.RLock()
+	h.mu.Lock()
```
lukechampine: why do you need a write lock here?

DavidVorick:
derp. Designed this in the reverse order, added the switch first and later decided it should be a setting. Didn't even think through the fact that we can already adjust the settings at will.

We can drop the extra methods and api calls, this change has a lot more code than is needed.

PR #922, opened 2016-01-15
Off switch

lukechampine:
This is a replacement of #918.

I waffled on adding a separate command to siac, because I wasn't sure what to name it. If you come up with a good name, I'll add it.

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -336,10 +336,19 @@ struct {
 	unlockhash   types.UnlockHash  (string)
 	windowsize   types.BlockHeight (uint64)
 
+	acceptingcontracts bool
 	numcontracts       uint64
 	revenue            types.Currency (string)
 	storageremaining   int64
 	anticipatedrevenue types.Currency (string)
+
+	rpcerrorcalls        uint64
+	rpcunrecognizedcalls uint64
+	rpcdownloadcalls     uint64
+	rpcrenewcalls        uint64
+	rpcrevisecalls       uint64
+	RPCsettingscalls     uint64
+	rpcuploadcalls       uint64
 }
 ```
 'collateral' is the number of hastings per byte per block that are put up as
@@ -365,6 +374,10 @@ downtime, 40 blocks is recommended as an absolute minimum. The current network
 default is 288 blocks. The current software will break entirely below 20
 blocks, though in theory something as low as 6 blocks could be safe.
 
+'acceptingcontracts' indicates whether the host is accepting new file
+contracts, file contract revisions, and file contract renewals. A host that is
+not accepting these calls will still accept download requests, and will still
+submit storage proofs to the network.
 
 'numcontracts' is the number of active contracts that the host is engaged in.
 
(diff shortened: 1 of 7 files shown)


VoidingWarranties:
What about `siac host pause` and `siac host resume`?

lukechampine:
I like those better than `siac host on`/`siac host off`. Let's see what david thinks

DavidVorick:
it's better than on/off, but pause/resume is still poorly named - the host when 'paused' will still service uploads and downloads, and will still respond to other types of requests as well

doc/API.md
```diff
@@ -336,10 +336,19 @@ struct {
 	unlockhash   types.UnlockHash  (string)
 	windowsize   types.BlockHeight (uint64)
 
+	acceptingcontracts bool
 	numcontracts       uint64
 	revenue            types.Currency (string)
 	storageremaining   int64
 	anticipatedrevenue types.Currency (string)
```
DavidVorick: clusters should be alphabetically sorted, previous PR screwed this up, can you fix it?

modules/host.go
```diff
@@ -52,14 +52,15 @@ type (
...
-		UnlockHash   types.UnlockHash  `json:"unlockhash"`
+		AcceptingContracts bool              `json:"acceptingcontracts"`
+		NetAddress         NetAddress        `json:"netaddress"`
+		TotalStorage       int64             `json:"totalstorage"`
+		MinDuration        types.BlockHeight `json:"minduration"`
+		MaxDuration        types.BlockHeight `json:"maxduration"`
+		WindowSize         types.BlockHeight `json:"windowsize"`
+		Price              types.Currency    `json:"price"`
+		Collateral         types.Currency    `json:"collateral"`
+		UnlockHash         types.UnlockHash  `json:"unlockhash"`
```
DavidVorick: These should be sorted

modules/host/announce.go
```diff
@@ -78,5 +83,8 @@ func (h *Host) AnnounceAddress(addr modules.NetAddress) error {
 	if h.closed {
 		return errHostClosed
 	}
+
+	h.mu.Lock()
+	defer h.mu.Unlock()
```
DavidVorick: probably okay to throw this to the top, other functions in the host take the form

```
lock
defer unlock
resourcelock
defer resource unlock
if closed {
}

rest of function
```

DavidVorick:
I think that the current siac implementation is okay.

modules/host.go
lukechampine: why

lukechampine: lol why

VoidingWarranties:
If using the off switch won't be a common need then having a siac command for it is unnecessary. What are the common reasons a host would want to use the off switch?

DavidVorick:
Pretty much the only reason that someone would turn off the host is if they are planning on repurposing the storage at some point in the future. They'd want to finish out their current contracts and then get the space back.

I don't think it's a common use case. Siac should be able to disable the host, but I don't think a separate command is necessary.

lukechampine:
Agreed. I think `siac host set acceptingcontracts false` is sufficiently intuitive. Of course, it will be more user-friendly in Sia-UI.

DavidVorick:
`siac host set` sounds fine to me.

DavidVorick:
though we might want to think about solidifying the `siac` commands along with the api, to preserve compatibility with bash scripts and stuff

modules/host.go
DavidVorick: It makes the fields easier to find. I think every grouping should be sorted internally, and if you feel like there's a reason to break the sorting, then that's reason enough to break apart the grouping.

PR #923, opened 2016-01-15
HostDB mocking

lukechampine:
This PR serves to showcase what extensive mocking would look like in a module. It's not perfect (comments/naming/organization could be improved), but it's more-or-less how I envisioned it.
Nearly all of the hostdb's dependencies have been mocked. (The exceptions are: `filepath.Join`, `sync.RWMutex`, `os.MkdirAll`, `crypto.Perm`, `rand.Int`, `time.Now`, `types.Currency`, and possibly one or two others that I missed. These are pretty inconsequential, since most do not return errors.)

The `New`/`newHostDB` pattern in particular has made it very easy to unit test the hostdb's methods. Most of our module tests have fully initialize the modules, meaning they spawn goroutines, write to disk, mine blocks, etc., all just to test a fairly simple function. By testing with `newHostDB` (which takes all dependencies as parameters and has no side effects) instead of `New`, it becomes trivial to pass in only the dependencies that will be called and mock them accordingly. See the newly-added tests for examples.

Finally, @VoidingWarranties and I talked in person about modifying this code to reduce duplication of types. The basic premise is to define shared "stub" types for each module in the `modules` package instead defining "restricted interfaces" internal to each module. As an example, the hostdb now defines the following type:
```go
hdbTransactionPool interface {
    AcceptTransactionSet([]types.Transaction) error
}
```
which is a parameter to `newHostDB`. Since `modules.TransactionPool` implements `hdbTransactionPool`, it can be passed to `newHostDB`. And during testing, tests which require `AcceptTransactionSet` must define a type that implements `hdbTransactionPool`.
In Jordan's proposal, we would instead define a stub type for the transaction pool like so:
```go
// in modules/transactionpool.go
type TransactionPoolStub struct{}
func (TransactionPoolStub) AcceptTransactionSet([]types.Transaction) error { return nil }
func (TransactionPoolStub) IsStandardTransaction(types.Transaction) error  { return nil }
func (TransactionPoolStub) PurgeTransactionPool()                          { }
// ... etc
```
then, when testing the hostdb, we would define our testing type like so:
```go
type tpoolTest struct {
    modules.TransactionPoolStub
}
func (tpoolTest) AcceptTransactionSet([]types.Transaction) error {
    // ... test-specific logic ...
}
```
And finally, we would throw away the `hdbTransactionPool` type, and have `newHostDB` take a `modules.TransactionPool` instead. Since `tpoolTest` embeds `modules.TransactionPoolStub`, it inherits its methods and thus implements the full `modules.TransactionPool` interface. We can then override any methods as needed during testing.
Part of the rationale here is that we will likely want to define shared stubs anyway, for dependencies like `time.Sleep` and `net.Dial`.

The code reduction is compelling, but something doesn't sit quite right with me about this approach. I prefer to pass modules the smallest interface they require. Also, embedding types all over the place may also be confusing or error-prone.
I guess my primary objection is that I'm averse to writing generic code until at least two specific cases exist. I feel that we should adopt the pattern in this PR for now, and consider abstracting it only after we've mocked another module to a similar degree. It will be easier to see the parallel and evaluate what makes sense once the code is on the page instead of in our heads.

### DIFF
--- a/modules/renter/hostdb/dependencies.go
+++ b/modules/renter/hostdb/dependencies.go
@@ -59,16 +59,21 @@ type hdbWalletShim struct {
 func (ws *hdbWalletShim) NextAddress() (types.UnlockConditions, error) { return ws.w.NextAddress() }
 func (ws *hdbWalletShim) StartTransaction() hdbTransactionBuilder      { return ws.w.StartTransaction() }
 
+// stdDialer implements the hdbDialer interface via net.DialTimeout.
 type stdDialer struct{}
 
 func (d stdDialer) DialTimeout(addr modules.NetAddress, timeout time.Duration) (net.Conn, error) {
 	return net.DialTimeout("tcp", string(addr), timeout)
 }
 
+// stdSleeper implements the hdbSleeper interface via time.Sleep.
 type stdSleeper struct{}
 
 func (s stdSleeper) Sleep(d time.Duration) { time.Sleep(d) }
 
+// stdPersist implements the hdbPersister interface via persist.SaveFile and
+// persist.LoadFile. The metadata and filename required by these functions is
+// internal to stdPersist.
 type stdPersist struct {
 	meta     persist.Metadata
 	filename string
@@ -92,6 +97,7 @@ func newPersist(dir string) *stdPersist {
 	}
 }
 
+// newLogger creates a persist.Logger with the standard filename.
 func newLogger(dir string) (*persist.Logger, error) {
 	return persist.NewLogger(filepath.Join(dir, "hostdb.log"))
 }
--- a/modules/renter/hostdb/dependencies.go
+++ b/modules/renter/hostdb/dependencies.go
@@ -1,0 +1,51 @@
+package hostdb
+
+import (
+	"net"
+	"time"
+
+	"github.com/NebulousLabs/Sia/modules"
+	"github.com/NebulousLabs/Sia/types"
+)
+
+// These interfaces define the HostDB's dependencies. Using the smallest
+// interface possible makes it easier to mock these dependencies in testing.
+type (
+	hdbConsensusSet interface {
+		ConsensusSetSubscribe(modules.ConsensusSetSubscriber)
+	}
+	hdbTransactionBuilder interface {
+		AddArbitraryData([]byte) uint64
+		AddFileContract(types.FileContract) uint64
+		Drop()
+		FundSiacoins(types.Currency) error
+		Sign(bool) ([]types.Transaction, error)
+		View() (types.Transaction, []types.Transaction)
+	}
+	hdbWallet interface {
+		NextAddress() (types.UnlockConditions, error)
+		StartTransaction() hdbTransactionBuilder
+	}
+	hdbTransactionPool interface {
+		AcceptTransactionSet([]types.Transaction) error
+	}
+
+	hdbDialer interface {
+		DialTimeout(modules.NetAddress, time.Duration) (net.Conn, error)
+	}
+)
+
+// because hdbWallet is not directly compatible with modules.Wallet (wrong
+// type signature for StartTransaction), we must provide a shim type.
+type hdbWalletShim struct {
+	w modules.Wallet
+}
+
+func (ws *hdbWalletShim) NextAddress() (types.UnlockConditions, error) { return ws.w.NextAddress() }
+func (ws *hdbWalletShim) StartTransaction() hdbTransactionBuilder      { return ws.w.StartTransaction() }
+
+type stdDialer struct{}
+
+func (d stdDialer) DialTimeout(addr modules.NetAddress, timeout time.Duration) (net.Conn, error) {
+	return net.DialTimeout("tcp", string(addr), timeout)
+}
--- a/modules/renter/hostdb/dependencies.go
+++ b/modules/renter/hostdb/dependencies.go
@@ -16,6 +16,17 @@ type (
 	hdbConsensusSet interface {
 		ConsensusSetSubscribe(modules.ConsensusSetSubscriber)
 	}
+	// in order to restrict the modules.TransactionBuilder interface, we must
+	// provide a shim to bridge the gap between modules.Wallet and
+	// hdbTransactionBuilder.
+	hdbWalletShim interface {
+		NextAddress() (types.UnlockConditions, error)
+		StartTransaction() modules.TransactionBuilder
+	}
+	hdbWallet interface {
+		NextAddress() (types.UnlockConditions, error)
+		StartTransaction() hdbTransactionBuilder
+	}
 	hdbTransactionBuilder interface {
 		AddArbitraryData([]byte) uint64
 		AddFileContract(types.FileContract) uint64
@@ -23,10 +34,6 @@ type (
 		FundSiacoins(types.Currency) error
 		Sign(bool) ([]types.Transaction, error)
 		View() (types.Transaction, []types.Transaction)
-	}
-	hdbWallet interface {
-		NextAddress() (types.UnlockConditions, error)
-		StartTransaction() hdbTransactionBuilder
 	}
 	hdbTransactionPool interface {
 		AcceptTransactionSet([]types.Transaction) error
@@ -51,13 +58,13 @@ type (
 )
 
 // because hdbWallet is not directly compatible with modules.Wallet (wrong
-// type signature for StartTransaction), we must provide a shim type.
-type hdbWalletShim struct {
-	w modules.Wallet
+// type signature for StartTransaction), we must provide a bridge type.
+type hdbWalletBridge struct {
+	w hdbWalletShim
 }
 
-func (ws *hdbWalletShim) NextAddress() (types.UnlockConditions, error) { return ws.w.NextAddress() }
-func (ws *hdbWalletShim) StartTransaction() hdbTransactionBuilder      { return ws.w.StartTransaction() }
+func (ws *hdbWalletBridge) NextAddress() (types.UnlockConditions, error) { return ws.w.NextAddress() }
+func (ws *hdbWalletBridge) StartTransaction() hdbTransactionBuilder      { return ws.w.StartTransaction() }
 
 // stdDialer implements the hdbDialer interface via net.DialTimeout.
 type stdDialer struct{}
--- a/modules/renter/hostdb/dependencies.go
+++ b/modules/renter/hostdb/dependencies.go
@@ -33,6 +33,10 @@ type (
 	hdbDialer interface {
 		DialTimeout(modules.NetAddress, time.Duration) (net.Conn, error)
 	}
+
+	hdbSleeper interface {
+		Sleep(time.Duration)
+	}
 )
 
 // because hdbWallet is not directly compatible with modules.Wallet (wrong
@@ -49,3 +53,7 @@ type stdDialer struct{}
 func (d stdDialer) DialTimeout(addr modules.NetAddress, timeout time.Duration) (net.Conn, error) {
 	return net.DialTimeout("tcp", string(addr), timeout)
 }
+
+type stdSleeper struct{}
+
+func (s stdSleeper) Sleep(d time.Duration) { time.Sleep(d) }
--- a/modules/renter/hostdb/dependencies.go
+++ b/modules/renter/hostdb/dependencies.go
@@ -44,6 +44,10 @@ type (
 		save(hdbPersist) error
 		load(*hdbPersist) error
 	}
+
+	hdbLogger interface {
+		Println(...interface{})
+	}
 )
 
 // because hdbWallet is not directly compatible with modules.Wallet (wrong
@@ -86,4 +90,8 @@ func newPersist(dir string) *stdPersist {
 		},
 		filename: filepath.Join(dir, "hostdb.json"),
 	}
+}
+
+func newLogger(dir string) (*persist.Logger, error) {
+	return persist.NewLogger(filepath.Join(dir, "hostdb.log"))
 }
--- a/modules/renter/hostdb/dependencies.go
+++ b/modules/renter/hostdb/dependencies.go
@@ -13,21 +13,21 @@ import (
 // These interfaces define the HostDB's dependencies. Using the smallest
 // interface possible makes it easier to mock these dependencies in testing.
 type (
-	hdbConsensusSet interface {
+	consensusSet interface {
 		ConsensusSetSubscribe(modules.ConsensusSetSubscriber)
 	}
 	// in order to restrict the modules.TransactionBuilder interface, we must
 	// provide a shim to bridge the gap between modules.Wallet and
-	// hdbTransactionBuilder.
-	hdbWalletShim interface {
+	// transactionBuilder.
+	walletShim interface {
 		NextAddress() (types.UnlockConditions, error)
 		StartTransaction() modules.TransactionBuilder
 	}
-	hdbWallet interface {
+	wallet interface {
 		NextAddress() (types.UnlockConditions, error)
-		StartTransaction() hdbTransactionBuilder
+		StartTransaction() transactionBuilder
 	}
-	hdbTransactionBuilder interface {
+	transactionBuilder interface {
 		AddArbitraryData([]byte) uint64
 		AddFileContract(types.FileContract) uint64
 		Drop()
@@ -35,50 +35,50 @@ type (
 		Sign(bool) ([]types.Transaction, error)
 		View() (types.Transaction, []types.Transaction)
 	}
-	hdbTransactionPool interface {
+	transactionPool interface {
 		AcceptTransactionSet([]types.Transaction) error
 	}
 
-	hdbDialer interface {
+	dialer interface {
 		DialTimeout(modules.NetAddress, time.Duration) (net.Conn, error)
 	}
 
-	hdbSleeper interface {
+	sleeper interface {
 		Sleep(time.Duration)
 	}
 
-	hdbPersister interface {
+	persister interface {
 		save(hdbPersist) error
 		load(*hdbPersist) error
 	}
 
-	hdbLogger interface {
+	logger interface {
 		Println(...interface{})
 	}
 )
 
-// because hdbWallet is not directly compatible with modules.Wallet (wrong
+// because wallet is not directly compatible with modules.Wallet (wrong
 // type signature for StartTransaction), we must provide a bridge type.
-type hdbWalletBridge struct {
-	w hdbWalletShim
+type walletBridge struct {
+	w walletShim
 }
 
-func (ws *hdbWalletBridge) NextAddress() (types.UnlockConditions, error) { return ws.w.NextAddress() }
-func (ws *hdbWalletBridge) StartTransaction() hdbTransactionBuilder      { return ws.w.StartTransaction() }
+func (ws *walletBridge) NextAddress() (types.UnlockConditions, error) { return ws.w.NextAddress() }
+func (ws *walletBridge) StartTransaction() transactionBuilder         { return ws.w.StartTransaction() }
 
-// stdDialer implements the hdbDialer interface via net.DialTimeout.
+// stdDialer implements the dialer interface via net.DialTimeout.
 type stdDialer struct{}
 
 func (d stdDialer) DialTimeout(addr modules.NetAddress, timeout time.Duration) (net.Conn, error) {
 	return net.DialTimeout("tcp", string(addr), timeout)
 }
 
-// stdSleeper implements the hdbSleeper interface via time.Sleep.
+// stdSleeper implements the sleeper interface via time.Sleep.
 type stdSleeper struct{}
 
 func (s stdSleeper) Sleep(d time.Duration) { time.Sleep(d) }
 
-// stdPersist implements the hdbPersister interface via persist.SaveFile and
+// stdPersist implements the persister interface via persist.SaveFile and
 // persist.LoadFile. The metadata and filename required by these functions is
 // internal to stdPersist.
 type stdPersist struct {
--- a/modules/renter/hostdb/dependencies.go
+++ b/modules/renter/hostdb/dependencies.go
@@ -2,9 +2,11 @@ package hostdb
 
 import (
 	"net"
+	"path/filepath"
 	"time"
 
 	"github.com/NebulousLabs/Sia/modules"
+	"github.com/NebulousLabs/Sia/persist"
 	"github.com/NebulousLabs/Sia/types"
 )
 
@@ -37,6 +39,11 @@ type (
 	hdbSleeper interface {
 		Sleep(time.Duration)
 	}
+
+	hdbPersister interface {
+		save(hdbPersist) error
+		load(*hdbPersist) error
+	}
 )
 
 // because hdbWallet is not directly compatible with modules.Wallet (wrong
@@ -57,3 +64,26 @@ func (d stdDialer) DialTimeout(addr modules.NetAddress, timeout time.Duration) (
 type stdSleeper struct{}
 
 func (s stdSleeper) Sleep(d time.Duration) { time.Sleep(d) }
+
+type stdPersist struct {
+	meta     persist.Metadata
+	filename string
+}
+
+func (p *stdPersist) save(data hdbPersist) error {
+	return persist.SaveFile(p.meta, data, p.filename)
+}
+
+func (p *stdPersist) load(data *hdbPersist) error {
+	return persist.LoadFile(p.meta, data, p.filename)
+}
+
+func newPersist(dir string) *stdPersist {
+	return &stdPersist{
+		meta: persist.Metadata{
+			Header:  "HostDB Persistence",
+			Version: "0.5",
+		},
+		filename: filepath.Join(dir, "hostdb.json"),
+	}
+}
--- a/modules/renter/hostdb/hostdb.go
+++ b/modules/renter/hostdb/hostdb.go
@@ -27,45 +27,14 @@ var (
 	errNilTpool  = errors.New("cannot create hostdb with nil transaction pool")
 )
 
-// These interfaces define the HostDB's dependencies. Using the smallest
-// interface possible makes it easier to mock these dependencies in testing.
-type (
-	hdbConsensusSet interface {
-		ConsensusSetSubscribe(modules.ConsensusSetSubscriber)
-	}
-	hdbTransactionBuilder interface {
-		AddArbitraryData([]byte) uint64
-		AddFileContract(types.FileContract) uint64
-		Drop()
-		FundSiacoins(types.Currency) error
-		Sign(bool) ([]types.Transaction, error)
-		View() (types.Transaction, []types.Transaction)
-	}
-	hdbWallet interface {
-		NextAddress() (types.UnlockConditions, error)
-		StartTransaction() hdbTransactionBuilder
-	}
-	hdbTransactionPool interface {
-		AcceptTransactionSet([]types.Transaction) error
-	}
-)
-
-// because hdbWallet is not directly compatible with modules.Wallet (differing
-// type signatures for StartTransaction), we must provide a shim type.
-type hdbWalletShim struct {
-	w modules.Wallet
-}
-
-func (ws *hdbWalletShim) NextAddress() (types.UnlockConditions, error) { return ws.w.NextAddress() }
-func (ws *hdbWalletShim) StartTransaction() hdbTransactionBuilder      { return ws.w.StartTransaction() }
-
 // The HostDB is a database of potential hosts. It assigns a weight to each
 // host based on their hosting parameters, and then can select hosts at random
 // for uploading files.
 type HostDB struct {
-	// modules
+	// dependencies
 	wallet hdbWallet
 	tpool  hdbTransactionPool
+	dialer hdbDialer
 
 	// The hostTree is the root node of the tree that organizes hosts by
 	// weight. The tree is necessary for selecting weighted hosts at
@@ -119,7 +88,7 @@ func New(cs hdbConsensusSet, wallet modules.Wallet, tpool hdbTransactionPool, pe
 		return nil, errNilTpool
 	}
 
-	hdb, err := newHostDB(&hdbWalletShim{w: wallet}, tpool, persistDir)
+	hdb, err := newHostDB(&hdbWalletShim{w: wallet}, tpool, stdDialer{}, persistDir)
 	if err != nil {
 		return nil, err
 	}
@@ -142,10 +111,11 @@ func New(cs hdbConsensusSet, wallet modules.Wallet, tpool hdbTransactionPool, pe
 // newHostDB creates a HostDB using the provided dependencies. It does not
 // have any side effects (i.e. it does not spawn background threads, perform
 // I/O, or call stateful methods of its dependencies.)
-func newHostDB(wallet hdbWallet, tpool hdbTransactionPool, persistDir string) (*HostDB, error) {
+func newHostDB(w hdbWallet, tpool hdbTransactionPool, d hdbDialer, persistDir string) (*HostDB, error) {
 	hdb := &HostDB{
-		wallet: wallet,
+		wallet: w,
 		tpool:  tpool,
+		dialer: d,
 
 		contracts:   make(map[types.FileContractID]hostContract),
 		activeHosts: make(map[modules.NetAddress]*hostNode),
--- a/modules/renter/hostdb/hostdb.go
+++ b/modules/renter/hostdb/hostdb.go
@@ -36,6 +36,7 @@ type HostDB struct {
 	tpool   hdbTransactionPool
 	dialer  hdbDialer
 	sleeper hdbSleeper
+	persist hdbPersister
 
 	// The hostTree is the root node of the tree that organizes hosts by
 	// weight. The tree is necessary for selecting weighted hosts at
@@ -58,8 +59,6 @@ type HostDB struct {
 	blockHeight   types.BlockHeight
 	contracts     map[types.FileContractID]hostContract
 	cachedAddress types.UnlockHash // to prevent excessive address creation
-
-	persistDir string
 
 	log *log.Logger
 	mu  sync.RWMutex
@@ -89,11 +88,11 @@ func New(cs hdbConsensusSet, wallet modules.Wallet, tpool hdbTransactionPool, pe
 		return nil, errNilTpool
 	}
 
-	hdb, err := newHostDB(&hdbWalletShim{w: wallet}, tpool, stdDialer{}, stdSleeper{}, persistDir)
+	hdb, err := newHostDB(&hdbWalletShim{w: wallet}, tpool, stdDialer{}, stdSleeper{}, newPersist(persistDir))
 	if err != nil {
 		return nil, err
 	}
-	err = hdb.initPersist()
+	err = hdb.initPersist(persistDir)
 	if err != nil {
 		return nil, err
 	}
@@ -112,19 +111,18 @@ func New(cs hdbConsensusSet, wallet modules.Wallet, tpool hdbTransactionPool, pe
 // newHostDB creates a HostDB using the provided dependencies. It does not
 // have any side effects (i.e. it does not spawn background threads, perform
 // I/O, or call stateful methods of its dependencies.)
-func newHostDB(w hdbWallet, tpool hdbTransactionPool, d hdbDialer, s hdbSleeper, persistDir string) (*HostDB, error) {
+func newHostDB(w hdbWallet, tpool hdbTransactionPool, d hdbDialer, s hdbSleeper, p hdbPersister) (*HostDB, error) {
 	hdb := &HostDB{
 		wallet:  w,
 		tpool:   tpool,
 		dialer:  d,
 		sleeper: s,
+		persist: p,
 
 		contracts:   make(map[types.FileContractID]hostContract),
 		activeHosts: make(map[modules.NetAddress]*hostNode),
 		allHosts:    make(map[modules.NetAddress]*hostEntry),
 		scanPool:    make(chan *hostEntry, scanPoolSize),
-
-		persistDir: persistDir,
 	}
 
 	return hdb, nil
--- a/modules/renter/hostdb/hostdb.go
+++ b/modules/renter/hostdb/hostdb.go
@@ -75,9 +75,9 @@ type hostContract struct {
 	SecretKey       crypto.SecretKey
 }
 
-// New creates and starts up a hostdb. The hostdb that gets returned will not
-// have finished scanning the network or blockchain.
+// New returns a new HostDB.
 func New(cs consensusSet, wallet walletShim, tpool transactionPool, persistDir string) (*HostDB, error) {
+	// Check for nil inputs.
 	if cs == nil {
 		return nil, errNilCS
 	}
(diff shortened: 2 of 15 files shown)


modules/renter/hostdb/dependencies.go
```diff
@@ -0,0 +1,110 @@
...
+	hdbConsensusSet interface {
+		ConsensusSetSubscribe(modules.ConsensusSetSubscriber)
+	}
+	// in order to restrict the modules.TransactionBuilder interface, we must
+	// provide a shim to bridge the gap between modules.Wallet and
+	// hdbTransactionBuilder.
+	hdbWalletShim interface {
+		NextAddress() (types.UnlockConditions, error)
+		StartTransaction() modules.TransactionBuilder
+	}
```
DavidVorick: why do you need the shim?  The hdbTransactionBuilder is a modules.TransactionBuilder

modules/renter/hostdb/hostdb.go
```diff
@@ -22,19 +22,22 @@ const (
 )
 
 var (
-	errNilCS     = errors.New("cannot create renter with nil consensus set")
-	errNilWallet = errors.New("cannot create renter with nil wallet")
-	errNilTpool  = errors.New("cannot create renter with nil transaction pool")
+	errNilCS     = errors.New("cannot create hostdb with nil consensus set")
+	errNilWallet = errors.New("cannot create hostdb with nil wallet")
+	errNilTpool  = errors.New("cannot create hostdb with nil transaction pool")
```
DavidVorick: heh

modules/renter/hostdb/hostdb.go
```diff
@@ -22,19 +22,22 @@ const (
...
-	cs     modules.ConsensusSet
-	wallet modules.Wallet
-	tpool  modules.TransactionPool
+	// dependencies
+	wallet  hdbWallet
+	tpool   hdbTransactionPool
+	dialer  hdbDialer
+	sleeper hdbSleeper
+	persist hdbPersister
+	log     hdbLogger
```
DavidVorick: hmmm

DavidVorick: mocking sucks

modules/renter/hostdb/hostdb.go
```diff
@@ -115,3 +119,22 @@ func New(cs modules.ConsensusSet, wallet modules.Wallet, tpool modules.Transacti
...
+		sleeper: s,
+		persist: p,
+		log:     l,
+
+		contracts:   make(map[types.FileContractID]hostContract),
+		activeHosts: make(map[modules.NetAddress]*hostNode),
+		allHosts:    make(map[modules.NetAddress]*hostEntry),
+		scanPool:    make(chan *hostEntry, scanPoolSize),
+	}
+}
```
DavidVorick: is this needed?

DavidVorick: oh. hmmm. Man this is rough to look at.

modules/renter/hostdb/hostdb.go
```diff
@@ -115,3 +119,22 @@ func New(cs modules.ConsensusSet, wallet modules.Wallet, tpool modules.Transacti
 
 	return hdb, nil
 }
+
+// newHostDB creates a HostDB using the provided dependencies. Unlike New, it
+// does not have any side effects (i.e. it does not spawn background threads,
+// perform I/O, or call stateful methods of its dependencies.)
+func newHostDB(w hdbWallet, tp hdbTransactionPool, d hdbDialer, s hdbSleeper, p hdbPersister, l hdbLogger) *HostDB {
```
DavidVorick: I don't think I like the names `hdbWallet`, couldn't it just be `wallet`?

modules/renter/hostdb/hostdb_test.go
```diff
@@ -0,0 +1,83 @@
...
+	"testing"
+
+	"github.com/NebulousLabs/Sia/build"
+	"github.com/NebulousLabs/Sia/modules"
+	"github.com/NebulousLabs/Sia/types"
+)
+
+// newStub is used to test the New function. It implements all of the hostdb's
+// dependencies.
+type newStub struct{}
```
DavidVorick: that's nice

modules/renter/hostdb/hostentry_test.go
```diff
@@ -0,0 +1,110 @@
...
+	case <-hdb.scanPool:
+		t.Error("duplicate host was added to scan pool")
+	case <-time.After(10 * time.Millisecond):
+	}
+}
+
+// TestActiveHosts tests the ActiveHosts method.
+func TestActiveHosts(t *testing.T) {
+	// no dependencies necessary
+	hdb := newHostDB(nil, nil, nil, nil, nil, nil)
```
DavidVorick: huh

modules/renter/hostdb/hostentry_test.go
```diff
@@ -0,0 +1,110 @@
...
+	h2 := new(hostEntry)
+	h2.NetAddress = "bar"
+	hdb.activeHosts = map[modules.NetAddress]*hostNode{
+		h1.NetAddress: &hostNode{hostEntry: h1},
+		h2.NetAddress: &hostNode{hostEntry: h2},
+	}
+	if hosts := hdb.ActiveHosts(); len(hosts) != 2 {
+		t.Errorf("wrong number of hosts: expected %v, got %v", 2, len(hosts))
+	}
+}
```
DavidVorick: This test is too implementation dependent. Basically it checks that `len(activeHosts)` is the function being used to check for the number of active hosts. That doesn't really check the logic. It would probably be better to add the host via an adding call, and then to check the length.

modules/renter/hostdb/hostentry_test.go
```diff
@@ -0,0 +1,110 @@
...
+	h2.NetAddress = "bar"
+	h2.Price = types.NewCurrency64(300)
+	h2.weight = baseWeight
+	hdb.insertNode(h2)
+	if len(hdb.activeHosts) != 2 {
+		t.Error("host was not added:", hdb.activeHosts)
+	}
+	if avg := hdb.AveragePrice(); avg.Cmp(types.NewCurrency64(200)) != 0 {
+		t.Error("average of two hosts should be their sum/2:", avg)
+	}
```
DavidVorick: Test also seems incomplete. Isn't the average price function actually pretty complex? Would be better to set up a matrix of input-outputs I think.

modules/renter/hostdb/hostdb.go
VoidingWarranties: @lukechampine can you explain more why `newHostDB` cannot have any side effects? I thought the purpose of the `New` + `newHostDB` pattern was to separate dependencies; to make `New` and `newHostDB` depend only on the arguments they are passed.

Would calling `s.Sleep` or `l.Println` in `newHostDB` be allowed?

modules/renter/hostdb/dependencies.go
lukechampine: yes, but unfortunately Go will still complain because the type signatures do not match: http://play.golang.org/p/RKOpJ0_-9m

modules/renter/hostdb/hostdb.go
lukechampine: Originally I liked the idea of having no side effects because it made testing easier; no extra goroutines flying around or files written to disk. Another big advantage is that you can pass `nil` for most (or all) of the dependencies if you know your test isn't going to call them.

Now I'm reconsidering. If you test with just `newHostDB`, you could argue that you're not *really* testing a hostDB, since a real hostDB has goroutines running in the background.

Perhaps the better approach is to make `New` the tiny function and `newHostDB` the big one. e.g.:
```go
func New(cs hdbConsensusSet, wallet hdbWalletShim, tpool hdbTransactionPool,
    persistDir string) (*hostDB, error) {
    return newHostDB(cs, wallet, tpool, stdDialer{}, stdSleeper{}, stdPersister{},
        stdLogger{}, persistDir)
}

func newHostDB(cs hdbConsensusSet, w hdbWallet, tp hdbTransactionPool, d hdbDialer,
    s hdbSleeper, p hdbPersister, l hdbLogger) (*hostDB, error) {
   // .. initialize and call dependencies as needed ...
    return hdb, nil
}
```
This is kind of a bummer because you can no longer pass `nil` for the `os` dependency, since `newHostDB` will now call `MkdirAll`. But that seems like a small price to pay.

modules/renter/hostdb/hostdb.go
lukechampine: I suppose. I wanted to avoid confusion between the hostdb wallet and the modules wallet. But I could go either way.

modules/renter/hostdb/hostentry_test.go
lukechampine: It should probably check that the contents of `hosts` match the contents of `hdb.activeHosts`. I agree that just checking the length isn't sufficient.

modules/renter/hostdb/hostentry_test.go
lukechampine: doing so is kind of a pain due to the number of steps required to create a hostEntry, but it could be done. My real concern with this test is that it doesn't address the case where the hostdb contains more host than `AveragePrice`'s sample size (18).

modules/renter/hostdb/hostdb.go
VoidingWarranties: I like this better. As for passing `nil`, if we ever write a mock package this will be a moot point as you could just pass in a stubOS.

DavidVorick: I definitely think that if you are going to be mocking, you need to be actually hitting the mocks. If you are making a hostdb during testing it should be calling MkdirAll, but that could be a mocked function itself.

But you want to make sure that the logic is as close as possible to the way the hostdb normally runs. You remove the dependencies so you can test the hostdb in isolation, but you don't want to deconstruct how the hostdb itself functions.

lukechampine:
There is some disagreement as to whether tests should always been run with a "real" HostDB, i.e. one that has goroutines flying around and concurrently mutating state.
My opinion is that unit tests such as `TestActiveHosts` should not be required to use a "real" HostDB. Unit tests are intended to be run in isolation. Mandating that unit tests spawn the HostDB's scanning threads turns them into integration tests, and introduces extra code (and headaches) for little clear gain.
I see parallels here to dependency mocking; they both make you think carefully about *boundaries*. When you're in mocking mode, you are drawing a tight circle around your object. Anywhere the circle is broken (e.g. where a dependency is tightly coupled), you use an interface to close the circle. This allows you to test your object in isolation.
With unit tests, the circle is broken by impurity. You want to test the function in as pure an environment as possible, so that you can verify that it produces the correct output without interference from hidden state, concurrent mutations, nondeterminism, etc. When testing pure functions, we don't have to worry about this at all. But where we do have impure functions, it makes sense to control them to the greatest extent possible. For example, `ActiveHosts` could be converted to a pure function like so:
```go
func (hdb *HostDB) ActiveHosts() []modules.HostSettings {
    hdb.mu.RLock()
    defer hdb.mu.RUnlock()
    return activeHosts(hdb.activeHosts) // implementation omitted
}
```
Then we could test `activeHosts` without needing a HostDB at all. But in practice this is pattern is probably infeasible, as it doubles the number of functions we have to test.
However, in `TestActiveHosts` we can more-or-less replicate this purity by creating a HostDB with only `activeHosts`, initialized with testing values. While the purity guarantees aren't as strong, they are better than nothing.
So my general recommendation is:
- Within reason, try to unit-test every function, and do so in isolation. Use table-driven tests where possible, and use mocking to ensure that dependencies are called correctly.
- Note where each function interacts with other functions, and write integration tests for each interaction. In general, these interactions happen when one function writes to a module field and another reads from it -- state is being shared. Test that a write followed by a read produces the correct result. And don't forget that concurrency (namely channels and mutexes) almost always implies shared state.

As an example, `hdb.insertHost` sends a value down the `hdb.scanPool` channel, while `hdb.threadedProbeHosts` reads from it. So we need two unit tests and one integration test. First, we test `hdb.insertHost` in isolation, manually reading values from `hdb.scanPool` to confirm that the send is occurring. Then we test `hdb.threadedProbeHosts` by manually sending values down `hdb.scanPool` and confirm that it is reading and processing the values correctly. Finally, we write a test that spawns `hdb.threadedProbeHosts` and calls `hdb.insertHost`, and confirm that they interacted properly. Now we can say the `hdb.insertHost<--->hdb.threadedProbeHosts` pairing has been tested to the best of our ability. Other integration tests can include this pairing as part of an even larger test.

DavidVorick:
My biggest concern is how useful the tests will be. When you test a function, you need to have strong certainty that the values you are testing correctly map to all ranges of values that will be provided to the function in a production environment.

If you consider the state to be part of the input, then you need to make sure that when simplifying the behavior of the state, you are simplifying it such that the tests are still representative of what will happen when the state is not simplified. I am worried that certain types of mocking over-simplify things, and therefore reduce the utility of the test.

I'm not making a general statement about mocking or that unit testing is bad. I'm just stating that I've seen unit tests which I would estimate are nearly worthless, and I want to be sure that we are writing unit tests which have actual utility.

lukechampine:
I agree that *integration* tests should try to emulate real-world conditions. That is absolutely important. But I think that stipulation makes less sense for unit tests, for the reasons given above.
Perhaps our strategy should be to write integration tests first, and unit tests only as a means of improving test coverage or checking for bugs revealed by the integration tests.
Another thing to consider is that integration tests can be just as fallible as unit tests when it comes to simulating reality. As you pointed out earlier, full code coverage does not imply absence of bugs. Just like unit tests, integration tests can wind up testing only a subset of the inputs that are possible in real-world conditions. The only difference is that with an integration test, you are guaranteed to be testing at least *one* code path that will occur in production.

modules/renter/hostdb/dependencies.go
```diff
@@ -0,0 +1,110 @@
...
+			Version: "0.5",
+		},
+		filename: filepath.Join(dir, "hostdb.json"),
+	}
+}
+
+// newLogger creates a persist.Logger with the standard filename.
+func newLogger(dir string) (*persist.Logger, error) {
+	return persist.NewLogger(filepath.Join(dir, "hostdb.log"))
+}
```
DavidVorick: ok, my understanding of this file is that you defined all of the dependencies for the host object, wrapped them in minimim-possible interfaces, then created structs which satisfied all of the dependencies for production environments. 

I do think that it makes sense to break this stuff apart by the file it's used in. For example, `save()` and `load()` seem more like dependencies that would belong in `persist_dependencies.go` rather than in a global dependencies file. But maybe it makes sense to keep it all in one place.

modules/renter/hostdb/hostdb.go
```diff
@@ -22,19 +22,22 @@ const (
...
-	cs     modules.ConsensusSet
-	wallet modules.Wallet
-	tpool  modules.TransactionPool
+	// dependencies
+	wallet  wallet
+	tpool   transactionPool
+	dialer  dialer
+	sleeper sleeper
+	persist persister
+	log     logger
```
DavidVorick: It's not alphabetically ordered

Also, I'm not sure it makes sense to put the sleepr/dialer/persister/logger next to the wallet and tpool, I feel like it makes sense to keep the other modules separate.

modules/renter/hostdb/hostdb.go
```diff
@@ -75,9 +75,9 @@ type hostContract struct {
 	SecretKey       crypto.SecretKey
 }
 
-// New creates and starts up a hostdb. The hostdb that gets returned will not
-// have finished scanning the network or blockchain.
-func New(cs modules.ConsensusSet, wallet modules.Wallet, tpool modules.TransactionPool, persistDir string) (*HostDB, error) {
+// New returns a new HostDB.
+func New(cs consensusSet, wallet walletShim, tpool transactionPool, persistDir string) (*HostDB, error) {
```
DavidVorick: I don't think `walletShim` is the right name for the dependency. What you need is a wallet, not a shim. If you want to use the name shim, it should probably be on the other interface. It just seems weird to me that a dependency would be a shim.

modules/renter/hostdb/hostdb_test.go
```diff
@@ -0,0 +1,125 @@
...
+	f, err := os.OpenFile(filepath.Join(dir, "hostdb.log"), os.O_CREATE, 0000)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer f.Close()
+	_, err = New(stub, stub, stub, dir)
+	if err == nil {
+		t.Fatal("expected permissions error, got nil")
+	}
+}
```
DavidVorick: :+1:

modules/renter/hostdb/hostdb_test.go
```diff
@@ -0,0 +1,125 @@
...
+
+	// Corrupted logfile.
+	os.RemoveAll(filepath.Join(dir, "hostdb.log"))
+	f, err := os.OpenFile(filepath.Join(dir, "hostdb.log"), os.O_CREATE, 0000)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer f.Close()
+	_, err = New(stub, stub, stub, dir)
+	if err == nil {
```
DavidVorick: would prefer if the error checking looked for an explicit error, helps to ensure correctness in the error handling.

modules/renter/hostdb/hostdb_test.go
```diff
@@ -0,0 +1,125 @@
...
+	// Bad persistDir.
+	_, err = New(stub, stub, stub, "")
+	if err == nil {
+		t.Fatal("expected invalid directory, got nil")
+	}
+
+	// Corrupted persist file.
+	ioutil.WriteFile(filepath.Join(dir, "hostdb.json"), []byte{1, 2, 3}, 0666)
+	_, err = New(stub, stub, stub, dir)
+	if err == nil {
```
DavidVorick: would prefer if the error checking looked for an explicit error, helps to ensure correctness in the error handling.

modules/renter/hostdb/hostdb_test.go
```diff
@@ -0,0 +1,125 @@
...
+
+	// Nil transaction pool.
+	_, err = New(stub, stub, nil, dir)
+	if err != errNilTpool {
+		t.Fatalf("expected %v, got %v", errNilTpool, err)
+	}
+
+	// Bad persistDir.
+	_, err = New(stub, stub, stub, "")
+	if err == nil {
```
DavidVorick: would prefer if the error checking looked for an explicit error, helps to ensure correctness in the error handling.

modules/renter/hostdb/hostdb_test.go
```diff
@@ -0,0 +1,125 @@
...
+	bridge := walletBridge{shim}
+	bridge.NextAddress()
+	if !shim.nextAddressCalled {
+		t.Error("NextAddress was not called on the shim")
+	}
+	bridge.StartTransaction()
+	if !shim.startTxnCalled {
+		t.Error("StartTransaction was not called on the shim")
+	}
+}
```
DavidVorick: Although I don't think there's anything wrong or inherently useless about this test, I am struggling to to imagine that it would actually have much chance of catching real bugs. There's no reason to remove it though.

modules/renter/hostdb/hostentry_test.go
```diff
@@ -0,0 +1,150 @@
...
+// scanHostEntry method.
+func TestInsertHost(t *testing.T) {
+	hdb := bareHostDB()
+
+	// invalid host should not be scanned
+	hdb.insertHost(modules.HostSettings{NetAddress: "foo"})
+	select {
+	case <-hdb.scanPool:
+		t.Error("invalid host was added to scan pool")
+	case <-time.After(10 * time.Millisecond):
```
DavidVorick: I think more time makes more sense.

DavidVorick:
One thing I definitely want to see is the strict use of the prefixes `TestUnit` and `TestIntegration` in the tests for all packages where mocking and 100% coverage has been achieved. I think it's important to be able to see what code isn't touched by integration tests, so that a judgement call can be made deciding whether more integration tests are necessary or whether the unit tests are being thorough enough.

A good example would be uploading. If you have 100% code coverage on uploading but none of it is an integration test with the renter, I would feel like an integration test would be necessary to add. But you might not realize that you are missing that coverage because the unit tests are saying that all the code is reached. OTOH, if you see that integration testing is covering all code except some of the unlikely error handling, it's probably good enough to accept that only unit tests are exploring the error handling code.

modules/renter/hostdb/hostentry_test.go
```diff
@@ -0,0 +1,150 @@
...
+	if hosts := hdb.ActiveHosts(); len(hosts) != 0 {
+		t.Errorf("wrong number of hosts: expected %v, got %v", 0, len(hosts))
+	}
+
+	// with one host
+	h1 := new(hostEntry)
+	h1.NetAddress = "foo"
+	hdb.activeHosts = map[modules.NetAddress]*hostNode{
+		h1.NetAddress: &hostNode{hostEntry: h1},
+	}
```
DavidVorick: I don't think that this is sufficient. If you want to test activeHosts in this way, you should write an `insertActiveHost` function that does the insertion for you. Then, any time an active host is being inserted within the codebase, it should call `insertActiveHost`, and you should have unit tests on the other functions which increase the active host count to make sure that they are increasing the active host count by calling `insertActiveHost`.

So, now I do see the usefulness of the type of test I was complaining about above. It's a good way to make sure that functions aren't taking shortcuts which may increase coupling between different parts of the codebase.

modules/renter/hostdb/hostentry_test.go
```diff
@@ -0,0 +1,150 @@
...
+		t.Errorf("ActiveHosts returned wrong host: expected %v, got %v", h1.NetAddress, hosts[0].NetAddress)
+	}
+
+	// with multiple hosts
+	h2 := new(hostEntry)
+	h2.NetAddress = "bar"
+	hdb.activeHosts = map[modules.NetAddress]*hostNode{
+		h1.NetAddress: &hostNode{hostEntry: h1},
+		h2.NetAddress: &hostNode{hostEntry: h2},
+	}
```
DavidVorick: ditto here

modules/renter/hostdb/hostentry_test.go
DavidVorick: Nonetheless this test needs to be more thorough.

modules/renter/hostdb/hostentry_test.go
```diff
@@ -0,0 +1,150 @@
...
+		allHosts: map[modules.NetAddress]*hostEntry{
+			"foo:1234": &hostEntry{online: true},
+			"bar:1234": &hostEntry{online: false},
+			"baz:1234": &hostEntry{online: true},
+		},
+		activeHosts: map[modules.NetAddress]*hostNode{
+			"foo:1234": nil,
+		},
+		scanPool: make(chan *hostEntry),
+	}
```
DavidVorick: same thing here as above, really should be building this map by making function calls.

modules/renter/hostdb/negotiate.go
```diff
@@ -373,7 +373,7 @@ func (hdb *HostDB) Renew(fcid types.FileContractID, newEndHeight types.BlockHeig
 	txnBuilder := hdb.wallet.StartTransaction()
 
 	// initiate connection
-	conn, err := net.DialTimeout("tcp", string(hc.IP), 15*time.Second)
+	conn, err := hdb.dialer.DialTimeout(hc.IP, 15*time.Second)
```
DavidVorick: :+1:

modules/renter/hostdb/negotiate_test.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/NebulousLabs/Sia/modules/gateway"
 	"github.com/NebulousLabs/Sia/modules/miner"
 	"github.com/NebulousLabs/Sia/modules/transactionpool"
-	"github.com/NebulousLabs/Sia/modules/wallet"
+	modWallet "github.com/NebulousLabs/Sia/modules/wallet" // name conflicts with type
```
DavidVorick: should it be the import that gets renamed or the object that gets renamed? I think I would prefer that the import not be changed, because it's referred to as `wallet` everywhere else in the codebase.

modules/renter/hostdb/persist.go
```diff
@@ -1,37 +1,23 @@
...
-
-const persistFilename = "hostdb.json"
-
-var saveMetadata = persist.Metadata{
-	Header:  "HostDB Persistence",
-	Version: "0.5",
+// hdbPersist defines what HostDB data persists across sessions.
+type hdbPersist struct {
+	Contracts []hostContract
 }
```
DavidVorick: Is this new, or has it been here? Because if the hostdb is doing persistence stuff, it should definitely be remembering activeHosts and allHosts

modules/renter/hostdb/dependencies.go
```diff
@@ -0,0 +1,110 @@
...
+			Header:  "HostDB Persistence",
+			Version: "0.5",
+		},
+		filename: filepath.Join(dir, "hostdb.json"),
+	}
+}
+
+// newLogger creates a persist.Logger with the standard filename.
+func newLogger(dir string) (*persist.Logger, error) {
+	return persist.NewLogger(filepath.Join(dir, "hostdb.log"))
```
DavidVorick: I would like to see a startup message in the log at initialization. It looks like you got rid of that, but I find it useful for telling when a user has changed sessions.

DavidVorick: oh nvm that happens in the persist package already automatically. sorry

modules/renter/hostdb/persist.go
DavidVorick: in another PR though

modules/renter/hostdb/persist_test.go
```diff
@@ -0,0 +1,66 @@
...
+	"testing"
+
+	"github.com/NebulousLabs/Sia/build"
+	"github.com/NebulousLabs/Sia/types"
+)
+
+// memPersist implements the persister interface in-memory.
+type memPersist hdbPersist
+
+func (m *memPersist) save(data hdbPersist) error { *m = memPersist(data); return nil }
```
DavidVorick: is the space here needed? I'm thinking it's not.

modules/renter/hostdb/persist_test.go
```diff
@@ -0,0 +1,66 @@
...
+	// create hostdb with mocked persist dependency
+	hdb := bareHostDB()
+	hdb.persist = new(memPersist)
+
+	// add some fake contracts
+	hdb.contracts = map[types.FileContractID]hostContract{
+		{0}: {IP: "foo"},
+		{1}: {IP: "bar"},
+		{2}: {IP: "baz"},
+	}
```
DavidVorick: same thing here I think, should be filling out the hostdb via function calls and not direct memory manipulation. The reason is to make it easy to extend. If you start storing metadata on the stuff that gets inserted, or if some extra checking needs to be added or something, or for some reason you change the architecture so multiple fields need to be updated instead of just one, then you can make that change through the function call and not break any of the tests.

DavidVorick: maybe I'm being too pedantic

modules/renter/hostdb/scan.go
```diff
@@ -202,32 +202,20 @@ func (hdb *HostDB) threadedScan() {
...
+			// hosts.
 			n := InactiveHostCheckupQuantity
-			if len(random) < InactiveHostCheckupQuantity {
-				n = len(random)
+			if n > len(entries) {
+				n = len(entries)
 			}
-			for i := 0; i < n; i++ {
-				hdb.scanHostEntry(random[i])
+			perm, err := crypto.Perm(n)
```
DavidVorick: I don't like the name `perm`, makes me think of permanent. I think that the full name `permutation` is justified, simply due to the namespace collision.

modules/renter/hostdb/scan.go
```diff
@@ -241,12 +229,11 @@ func (hdb *HostDB) threadedScan() {
 		if err != nil {
 			if build.DEBUG {
 				panic(err)
```
DavidVorick: not sure if we should do this in one go, or do it piecemeal, but slowly we can start replacing `if DEBUG panic()` with `build.Critical()`, because build.Critical will handle not-panicking correctly.

modules/renter/hostdb/scan_test.go
```diff
@@ -0,0 +1,156 @@
...
+)
+
+// TestDecrementReliability tests the decrementReliability method.
+func TestDecrementReliability(t *testing.T) {
+	hdb := bareHostDB()
+
+	// Decrementing a non-existent host should be a no-op.
+	// NOTE: can't check any post-conditions here; only indication of correct
+	// behavior is that the test doesn't panic.
+	hdb.decrementReliability("foo", types.NewCurrency64(0))
```
DavidVorick: I feel like this is incorrect behavior. I think it makes sense to either trigger a `build.Critical` or return an error if `decrementReliability` gets called on a nonexistant host. I'm leaning towards `build.Critical`, as this indicates developer error and not user error.

modules/renter/hostdb/scan_test.go
```diff
@@ -0,0 +1,156 @@
...
+	// Decrement reliability to 0. This should remove the host from allHosts.
+	hdb.decrementReliability(h.NetAddress, h.reliability)
+	if _, ok := hdb.allHosts[h.NetAddress]; ok {
+		t.Error("decrementing did not remove host from allHosts")
+	}
+}
+
+// probeDialer is used to test the threadedProbeHosts method. A simple type
+// alias is used so that it can easily be redefined during testing, allowing
+// multiple behaviors to be tested.
```
DavidVorick: cool, found this comment to be very helpful

modules/renter/hostdb/scan_test.go
```diff
@@ -0,0 +1,156 @@
...
+	h.NetAddress = "foo"
+	h.reliability = baseWeight // enough to withstand a few failures
+
+	// define a helper function for running threadedProbeHosts. We send the
+	// hostEntry, close the channel, and then call threadedProbeHosts.
+	// threadedProbeHosts will receive the host, loop once, and return after
+	// seeing the channel has closed.
+	runProbe := func(h *hostEntry) {
+		hdb.scanPool <- h
+		close(hdb.scanPool)
```
DavidVorick: tbh I had no idea what this keyword did, and it confused me for a while. TIL

modules/renter/hostdb/scan_test.go
```diff
@@ -0,0 +1,156 @@
...
+		// reset hdb.scanPool
+		hdb.scanPool = make(chan *hostEntry, 1)
+	}
+
+	// make the dial fail
+	hdb.dialer = probeDialer(func(modules.NetAddress, time.Duration) (net.Conn, error) {
+		return nil, net.UnknownNetworkError("fail")
+	})
+	runProbe(h)
+	if _, ok := hdb.activeHosts[h.NetAddress]; ok {
```
DavidVorick: you can probably use an exported function here to count the number of hosts, instead of directly looking at the memory. This helps to keep the test behavior-oriented instead of implementation-oriented.

modules/renter/hostdb/scan_test.go
```diff
@@ -0,0 +1,156 @@
...
+			// write old host settings
+			encoding.WriteObject(ourConn, oldHostSettings{
+				NetAddress: "probed",
+			})
+			ourConn.Close()
+		}()
+		return theirConn, nil
+	})
+	runProbe(h)
+	if _, ok := hdb.activeHosts[h.NetAddress]; !ok {
```
DavidVorick: same thing, could probably use the call `hdb.ActiveHosts()`

modules/renter/hostdb/scan_test.go
```diff
@@ -0,0 +1,156 @@
...
+		t.Error("host was not added")
+	}
+
+	// TODO: respond with old host settings
```
DavidVorick: did you resolve this TODO? looks like you did

modules/renter/hostdb/scan_test.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+// TestThreadedScan tests the threadedScan method.
+func TestThreadedScan(t *testing.T) {
+	hdb := bareHostDB()
+	// use a real sleeper; this will prevent threadedScan from looping too
+	// quickly.
+	hdb.sleeper = stdSleeper{}
```
DavidVorick: you could also send in a sleeper that sleeps for like 1/3 time or something. Not sure that it's needed here, but it's a thought.

modules/renter/hostdb/scan_test.go
```diff
@@ -0,0 +1,156 @@
...
+	h.reliability = types.NewCurrency64(1)
+	hdb.activeHosts[h.NetAddress] = &hostNode{hostEntry: h}
+
+	// perform one scan
+	go hdb.threadedScan()
+
+	// host should be sent down scanPool
+	select {
+	case <-hdb.scanPool:
+	case <-time.After(100 * time.Millisecond):
```
DavidVorick: cleanup question: if you call `time.After` but never pull the object out of the channel, does that mean the channel exists forever or is it going to get correctly garbage collected?

modules/renter/hostdb/upload.go
```diff
@@ -211,7 +211,7 @@ outer:
 
 	// Ask the hostdb for random hosts. We always ask for at least 10, to
 	// avoid selecting the same uncooperative hosts over and over.
-	ask := n
+	ask := n * 2
```
DavidVorick: woo

modules/renter/hostdb/weightedlist_test.go
```diff
@@ -326,12 +329,6 @@ func TestRandomHosts(t *testing.T) {
 	if len(randHosts) != 3 {
 		t.Error("didn't get 3 hosts")
 	}
-	if len(hdb.activeHosts) != 3 {
-		t.Error("wrong number of hosts")
-	}
-	if hdb.hostTree.weight.Cmp(types.NewCurrency64(6)) != 0 {
-		t.Error("unexpected weight at initialization")
-	}
```
DavidVorick: why did you remove these?

persist/log.go
```diff
@@ -55,16 +53,16 @@ func NewLogger(logFilename string) (*Logger, error) {
...
 }
 
 // Close terminates the Logger.
 func (l *Logger) Close() error {
 	l.Println("SHUTDOWN: Logging has terminated.")
-	return l.logFileWrapper.Close()
+	return l.file.Close()
 }
 
 // Critical will panic if debug mode is activated, and will log the statement
```
DavidVorick: These changes really belong in a different PR

persist/log.go
```diff
@@ -8,44 +8,42 @@ import (
...
-// Close closes the log file wrapper.
-func (lfw *logFileWrapper) Close() error {
+// Close closes the file and sets the closed flag.
+func (cf *closeableFile) Close() error {
 	// Sanity check - close should not have been called yet.
-	if build.DEBUG && lfw.closed {
-		panic("cannot close the logger after it has been closed")
+	if build.DEBUG && cf.closed {
+		panic("cannot close the file; already closed")
+	}
```
DavidVorick: can switch this to `build.Critical`

persist/log.go
```diff
@@ -8,44 +8,42 @@ import (
...
+	// Ensure that all data has actually hit the disk.
+	if err := cf.Sync(); err != nil {
```
DavidVorick: Is the `Sync` needed? I would like to reserve calls to `Sync` for user-triggered stuff and must-be-consistent stuff.

persist/log.go
```diff
@@ -8,44 +8,42 @@ import (
...
 }
 
 // Write takes the input data and writes it to the file.
-func (lfw *logFileWrapper) Write(b []byte) (int, error) {
+func (cf *closeableFile) Write(b []byte) (int, error) {
 	// Sanity check - close should not have been called yet.
-	if build.DEBUG && lfw.closed {
-		panic("cannot write to the logger after it has been closed")
+	if build.DEBUG && cf.closed {
+		panic("cannot write to the file after it has been closed")
```
DavidVorick: `build.Critical`

modules/renter/hostdb/dependencies.go
lukechampine: that breaks if you use a dependency in multiple files. And I don't want to have to hunt through four different `_dependencies` files to find where something is defined.

modules/renter/hostdb/hostdb_test.go
lukechampine: getting 100% coverage took some doing, haha

modules/renter/hostdb/hostdb_test.go
lukechampine: yeah. Just a "might as well get 100%" test.

modules/renter/hostdb/hostentry_test.go
lukechampine: a full second would probably be fine. Even Travis shouldn't be that slow.

modules/renter/hostdb/negotiate_test.go
lukechampine: yeah, the wallet dependency is just a pain in general.
maybe I should just throw out the `txnBuilder` shim code. Stubbing out a full transaction builder isn't all that bad. I don't think we actually do it anywhere.

modules/renter/hostdb/persist_test.go
lukechampine: between the type and the method? Those are enforced by go fmt.

modules/renter/hostdb/scan_test.go
lukechampine: ಠ_ಠ

modules/renter/hostdb/scan_test.go
lukechampine: yeah I was pretty jazzed when I realized I could do that.

modules/renter/hostdb/scan_test.go
lukechampine: nah, I wrote this test before the host compatibility had been introduced, so testing that behavior was impossible. I'll add it now.

modules/renter/hostdb/scan_test.go
lukechampine: yes, as Jordan discovered recently, `time.After` leaks a channel. During production we'll want to ensure we aren't doing that anywhere (or at least minimizing it), but during testing I think it's totally benign unless you're calling it in a loop or something.

modules/renter/hostdb/weightedlist_test.go
lukechampine: they're not really testing the `RandomHosts` function, just the implementation details of the host tree.

persist/log.go
lukechampine: I would agree if it were being called after each log, but I think it's fine if it's only called during `Close`. We don't want to risk dropping log info.

persist/log.go
lukechampine: guilty as charged

modules/renter/hostdb/scan_test.go
DavidVorick: if you are feeling special, you could run `t.Fatal` in the error case, and then create a goroutine to receive on the channel in the normal case. Or is the channel still leaked even if you pull the object out?

lukechampine: Actually, it's less clear than I thought. `time.Tick` seems to be the main culprit of memory leaks. I think `time.After` does not leak, although its memory won't be freed until the `<-` goes out of scope. After that, the channel can no longer be accessed, so it can be freed.
The problem with `time.Tick` is that a goroutine is continuously sending down the channel, so it can never be freed. But with `time.After`, only one send ever occurs.

DavidVorick: oh, that's no so bad then. I'm impressed the GC is smart enough to collect the time.After channel when the <- leaves scope. That's cool.

lukechampine:
Most concerns addressed. There are a few unfortunate bits left (like `walletShim` and `modWallet`), but they can be resolved later. I want to move on to more pressing concerns, so if you have any final comments, they should be "must-haves."

modules/renter/hostdb/hostentry_test.go
DavidVorick: :[

modules/renter/hostdb/negotiate_test.go
DavidVorick: really would like to have this name changed. I do want to block on this issue.

modules/renter/hostdb/scan_test.go
DavidVorick: :[

DavidVorick: you should at least drop a TODO in the code to suggest that a `build.Critical` should be thrown at some point in the future.

modules/renter/hostdb/hostentry_test.go
DavidVorick: TODO here as well

DavidVorick:
So, should drop 2 TODOs and should change the need for 

`modWallet github.com/bla/bla/bla`

lukechampine:
doing so would require renaming the `wallet` type. What is a better name?

PR #925, opened 2016-01-16
Change the file contract deletion endpoint to a POST request

VoidingWarranties:
This matches the API for deleting a file entry from the renter, which is also a POST request. It is also my opinion that any requests that mutate state should be a POST request, and any non-state-mutating requests should be a GET request.


VoidingWarranties:
@DavidVorick did you have any concerns about this?

lukechampine:
It's fine except that the build failed. We really need to squash that nondeterminism...but in the meantime you can force it rerun with `git commit --amend --no-message && git push --force`

VoidingWarranties:
Rerunning test. FYI if you log into Travis CI there is a rerun button.

lukechampine:
good to know. But you can't do that from the command line :P

PR #928, opened 2016-01-20
Small explorer improvements

VoidingWarranties:
A few small improvements I found while trying to track down potential memory leaks. The only memory improvement was adding `delete(e.blockTargets, bid)`.

Also removed unnecessary lines that set fields on the explorer when reverting a block only to have the same fields reset when applying blocks later in the function.

All other changes fix style consistency.

### DIFF
--- a/modules/explorer/update.go
+++ b/modules/explorer/update.go
@@ -18,9 +18,9 @@ func (e *Explorer) ProcessConsensusChange(cc modules.ConsensusChange) {
 		tbid := types.TransactionID(bid)
 
 		// Update all of the explorer statistics.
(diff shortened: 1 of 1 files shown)


VoidingWarranties:
David said the assumption is correct, but if we do not verify the assumption I don't think we should use it. I'll revert that commit. There are other issues if we don't make that assumption though. Specifically `e.timestamp` and `e.target` should be the timestamp and target of the reverted block's parent instead of the timestamp and target of the reverted block.

DavidVorick:
needs to be rebased

VoidingWarranties:
Done

PR #930, opened 2016-01-20
Host sanity

DavidVorick:
The host makes some assumptions about the state and contents of file contract obligations. There are now sanity checks to verify that these assumptions always hold.

### DIFF
--- a/build/critical.go
+++ b/build/critical.go
@@ -1,12 +1,14 @@
 package build
 
 import (
+	"fmt"
 	"os"
 )
 
 // Critical will print a message to os.Stderr unless DEBUG has been set, in
 // which case panic will be called instead.
-func Critical(s string) {
+func Critical(v ...interface{}) {
+	s := fmt.Sprintln(v...)
 	os.Stderr.WriteString(s)
 	if DEBUG {
 		panic(s)
--- a/build/critical.go
+++ b/build/critical.go
@@ -1,0 +1,14 @@
+package build
+
+import (
+	"os"
+)
+
+// Critical will print a message to os.Stderr unless DEBUG has been set, in
+// which case panic will be called instead.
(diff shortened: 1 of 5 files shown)


modules/host/obligations.go
```diff
@@ -63,6 +65,57 @@ type contractObligation struct {
...
+//  - There are two 'validProofOutputs' and two 'missedProofOutputs' for each
+//    file contract or revision.
+//  - RevisionConfirmed is set to 'true' if there is no file contract revision.
+//  - The ID is non-zero.
+func (co *contractObligation) isSane() error {
+	// Check that there is a file contract, and that it has the correct number
+	// of valid and missed proof outputs.
+	fclen := len(co.OriginTransaction.FileContracts)
+	if fclen != 1 {
+		return errors.New("obligation has bad file contract count: " + strconv.Itoa(fclen))
```
lukechampine: cleaner to import `fmt` instead of `strconv` and use `fmt.Errorf` instead

build/critical.go
```diff
@@ -0,0 +1,14 @@
+package build
+
+import (
+	"os"
+)
+
+// Critical will print a message to os.Stderr unless DEBUG has been set, in
+// which case panic will be called instead.
+func Critical(s string) {
+	os.Stderr.WriteString(s)
```
lukechampine: I would use variadic arguments for this (like `log.Println`):
```go
    s := fmt.Sprintln(v...)
    os.Stderr.WriteString(s)
    if DEBUG {
        panic(s)
    }
}
```

DavidVorick: ok, will do that, for log.Critical as well

modules/host/obligations.go
DavidVorick: ok, can do

PR #933, opened 2016-01-21
Add siac renter commands for listing upload and download queues

VoidingWarranties:
`siac renter uploads` lists only files that are currently uploading.

`siac renter downloads [--history/-H]` lists only files that are currently downloading. The `--history` flag shows previously downloaded files in addition to the currently downloading files. There is no `--history` flag for `siac renter uploads` because the backend does not yet support this.

I also moved all `command.AddCommand(subCommand)` into `init()` functions in the respective command's file. This way the main function doesn't need to worry about adding all of the subcommands.

### DIFF
--- a/siac/rentercmd.go
+++ b/siac/rentercmd.go
@@ -155,44 +155,40 @@ func renterdownloadscmd() {
 		fmt.Println("Could not get download queue:", err)
 		return
 	}
-	func() {
-		// Filter out files that have been downloaded.
-		var downloading []modules.DownloadInfo
-		for _, file := range queue.Downloads {
-			if file.Received != file.Filesize {
-				downloading = append(downloading, file)
-			}
+	// Filter out files that have been downloaded.
+	var downloading []modules.DownloadInfo
+	for _, file := range queue.Downloads {
+		if file.Received != file.Filesize {
+			downloading = append(downloading, file)
 		}
-		if len(downloading) == 0 {
-			fmt.Println("No files are downloading.")
-			return
-		}
+	}
+	if len(downloading) == 0 {
+		fmt.Println("No files are downloading.")
+	} else {
 		fmt.Println("Downloading", len(downloading), "files:")
 		for _, file := range downloading {
 			fmt.Printf("%s: %5.1f%% %s -> %s\n", file.StartTime.Format("Jan 02 03:04 PM"), 100*float64(file.Received)/float64(file.Filesize), file.SiaPath, file.Destination)
 		}
-	}()
+	}
 	if !renterShowHistory {
 		return
 	}
-	func() {
-		fmt.Println()
-		// Filter out files that are downloading.
-		var downloaded []modules.DownloadInfo
-		for _, file := range queue.Downloads {
-			if file.Received == file.Filesize {
-				downloaded = append(downloaded, file)
-			}
+	fmt.Println()
+	// Filter out files that are downloading.
+	var downloaded []modules.DownloadInfo
+	for _, file := range queue.Downloads {
+		if file.Received == file.Filesize {
+			downloaded = append(downloaded, file)
 		}
-		if len(downloaded) == 0 {
-			fmt.Println("No files downloaded.")
-			return
-		}
+	}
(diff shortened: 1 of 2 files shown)


lukechampine:
hmm. Not sure how I feel about all those `init`s. I kind of like keeping the whole structure (and all the flags) in main.go.

siac/rentercmd.go
```diff
@@ -110,21 +130,107 @@ func abs(path string) string {
...
+
+func filterDownloadInfos(dis []modules.DownloadInfo, f func(modules.DownloadInfo) bool) []modules.DownloadInfo {
+	disFiltered := make([]modules.DownloadInfo, 0)
+	for _, di := range dis {
+		if f(di) {
+			disFiltered = append(disFiltered, di)
+		}
+	}
+	return disFiltered
+}
```
lukechampine: filtering functions are probably overkill unless we expect to use them for multiple commands.
shed a tear for Go's lack of generics, then move on.

also, for future reference, prefer `var slice []T` to `slice := make([]T, 0)` (unless it's going to be encoded as JSON)

VoidingWarranties:
My reasoning for using `init`s is because it's nice to have the flag variables in the same file as the functions that use them, and the flags in the same file as the the flag variables. But there's a benefit to having all of the flags and commands in one place too. I'll revert that commit.

VoidingWarranties:
Should be all good now.

siac/rentercmd.go
```diff
@@ -110,21 +118,81 @@ func abs(path string) string {
 	return abspath
 }
 
-func renterdownloadqueuecmd() {
+func renteruploadscmd() {
```
DavidVorick: needs a docstring

siac/rentercmd.go
```diff
@@ -110,21 +118,81 @@ func abs(path string) string {
...
+		fmt.Println("No files are uploading.")
+		return
+	}
+	fmt.Println("Uploading", len(filteredFiles), "files:")
+	for _, file := range filteredFiles {
+		fmt.Printf("%13s  %s (uploading, %0.2f%%)\n", filesizeUnits(int64(file.Filesize)), file.SiaPath, file.UploadProgress)
+	}
+}
+
+func renterdownloadscmd() {
```
DavidVorick: docstring

siac/rentercmd.go
```diff
@@ -110,21 +118,81 @@ func abs(path string) string {
...
 	var queue api.RenterDownloadQueue
 	err := getAPI("/renter/downloads", &queue)
 	if err != nil {
 		fmt.Println("Could not get download queue:", err)
 		return
 	}
-	if len(queue.Downloads) == 0 {
-		fmt.Println("No downloads to show.")
+	func() {
```
DavidVorick: explain the anonymous function

siac/rentercmd.go
```diff
@@ -110,21 +118,81 @@ func abs(path string) string {
...
+		}
+	}()
+	if !renterShowHistory {
 		return
 	}
-	fmt.Println("Download Queue:")
-	for _, file := range queue.Downloads {
-		fmt.Printf("%s: %5.1f%% %s -> %s\n", file.StartTime.Format("Jan 02 03:04 PM"), 100*float64(file.Received)/float64(file.Filesize), file.SiaPath, file.Destination)
-	}
+	func() {
```
DavidVorick: explain the anonymous function

siac/rentercmd.go
VoidingWarranties: None of the command functions (not just renter) have docstrings. I don't think they're necessary as the functions are very well described in the command initialization. If anything I'd say the only docstring needed would be something like "renteruploadscmd is the handler for the command renter uploads" but that's already obvious from the name of the function.

DavidVorick: it's more of a pedantic thing. All functions are supposed to have docstrings, including test functions. It helps with things like grep, and helps you orient yourself if you aren't familiar with the project as a whole.

A lot of these functions were created before we were strict with how we do things. Siac is also heroically undertested.

PR #934, opened 2016-01-22
Miner testing

DavidVorick:
There's not much interesting happening in this PR. I did a lot of live-testing on my GPU to try and figure out why miners weren't finding blocks, but wasn't having any luck.

I haven't yet experimented in adding transactions from the transaction pool, perhaps I should get tests that check what happens when transactions are being periodically added to the transaction pool, and therefore the blocks.

### DIFF
--- a/modules/miner/blockmanager.go
+++ b/modules/miner/blockmanager.go
@@ -65,7 +65,7 @@ func (m *Miner) newSourceBlock() {
 // will store the header in memory for a while, depending on the constants
 // 'HeaderMemory', 'BlockMemory', and 'MaxSourceBlockAge'. On the full network,
 // it is typically safe to assume that headers will be remembered for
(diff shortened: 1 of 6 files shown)


DavidVorick:
I've now done testing involving transactions. Found an error with host announcements, but no errors related to transactions.

I think the miners that were suspicious were just incorrect. But maybe we should consider adding an 'almost solved' feature so miners can see that they are mining correctly and just need to get lucky.

lukechampine:
still failing nondeterministically :/

DavidVorick:
I fixed one of the issues, and I think the most major issue.

```
--- FAIL: TestIntegrationWalletTransactionGETid (0.02s)

	wallet_test.go:245: provided encryption key is incorrect
```

^ that one has been a problem for a while. It's like 1-in-20

PR #935, opened 2016-01-22
Set AcceptingContracts in hostHandlerGET

VoidingWarranties:
Otherwise siac will not show the proper value for acceptingcontracts in
host config

### DIFF
--- a/api/host.go
+++ b/api/host.go
@@ -55,6 +55,7 @@ func (srv *Server) hostHandlerGET(w http.ResponseWriter, req *http.Request, _ ht
 		UnlockHash:   settings.UnlockHash,
 		WindowSize:   settings.WindowSize,
 
+		AcceptingContracts: settings.AcceptingContracts,
 		NumContracts:       srv.host.Contracts(),
 		LostRevenue:        lostRevenue,
 		Revenue:            revenue,


VoidingWarranties:
This bug was caused because the `AcceptingContracts` field was added to `HostGET` struct, but because go doesn't require all fields to be named when constructing a struct, it went unnoticed. If we didn't name fields this would have raised a compile time error. Instead, we could initialize `HostGet` like:
```go
hg := HostGET{
 	/*Collateral:*/   settings.Collateral,
 	/*NetAddress:*/   settings.NetAddress,
 	/*MaxDuration:*/  settings.MaxDuration,
 	/*MinDuration:*/  settings.MinDuration,
 	/*Price:*/        settings.Price,
 	/*TotalStorage:*/ settings.TotalStorage,
 	/*UnlockHash:*/   settings.UnlockHash,
 	/*WindowSize:*/   settings.WindowSize,
 
	/*AcceptingContracts:*/ settings.AcceptingContracts,
	/*NumContracts:*/       srv.host.Contracts(),
	/*LostRevenue:*/        lostRevenue,
	/*Revenue:*/            revenue,
	/*StorageRemaining:*/   srv.host.Capacity(),
	/*AnticipatedRevenue:*/ anticipatedRevenue,

	/*RPCErrorCalls:*/        rpcCalls.ErrorCalls,
	/*RPCUnrecognizedCalls:*/ rpcCalls.UnrecognizedCalls,
	/*RPCDownloadCalls:*/     rpcCalls.DownloadCalls,
	/*RPCRenewCalls:*/        rpcCalls.RenewCalls,
	/*RPCReviseCalls:*/       rpcCalls.ReviseCalls,
	/*RPCSettingsCalls:*/     rpcCalls.SettingsCalls,
	/*RPCUploadCalls:*/       rpcCalls.UploadCalls,
}
```

Not as pretty, but better than bugs like this going unnoticed.

DavidVorick:
Our linting tool mandates that you use named fields. If you don't, you can run into severe issues that aren't caught. Rather than just a missing field, you could be shifted a field.

PR #942, opened 2016-01-24
Remove status commands from siac

VoidingWarranties:
e.g. The command to get the status of the miner is `siac miner` instead of `siac miner status`.

@lukechampine and I discussed this earlier. Status commands are redundant and require more typing.

Also renamed hostdb command to `siac host list`.
And renamed `siac gateway status` to `siac gateway list` (also callable with `siac gateway`).

### DIFF
--- a/siac/gatewaycmd.go
+++ b/siac/gatewaycmd.go
@@ -13,7 +13,7 @@ var (
 		Use:   "gateway",
 		Short: "Perform gateway actions",
 		Long:  "Add or remove a peer, view the current peer list, or synchronize to the network.",
-		Run:   wrap(gatewaylistcmd),
+		Run:   wrap(gatewaycmd),
 	}
 
 	gatewayAddCmd = &cobra.Command{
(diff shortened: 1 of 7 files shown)


siac/gatewaycmd.go
```diff
@@ -56,11 +60,13 @@ func gatewayremovecmd(addr string) {
 	fmt.Println("Removed", addr, "from peer list.")
 }
 
-func gatewaystatuscmd() {
+// gatewaylistcmd is the handler for the command `siac gateway list`.
+// Prints a list of all peers.
```
lukechampine: it also prints the netaddress of the gateway...maybe that should be split into a separate command? `siac gateway` could show just the netaddress and the number of peers.

siac/hostcmd.go
```diff
@@ -147,3 +162,22 @@ RPC Stats:
 `, hg.NetAddress, hg.UnlockHash, hg.RPCErrorCalls, hg.RPCUnrecognizedCalls, hg.RPCDownloadCalls,
 		hg.RPCRenewCalls, hg.RPCReviseCalls, hg.RPCSettingsCalls, hg.RPCUploadCalls)
 }
+
+// hostlistcmd is the handler for the command `siac host list`.
+// Prints a list of active hosts.
+func hostlistcmd() {
+	info := new(api.ActiveHosts)
+	err := getAPI("/renter/hosts/active", info)
```
lukechampine: wasn't this functionality moved to hostdbcmd.go?

lukechampine:
I feel that `siac host list` may be a little confusing, since it doesn't pertain to the host. I don't see anything wrong with keeping `siac hostdb`.

DavidVorick:
I agree with Luke, `hostdb` and `host` are two completely separate things, and should not be confused. Especially because the hostdb and the host are each going to get more features and more configurability and are likely to further step on eachother's toes.

VoidingWarranties:
Reverted hostdb change.

PR #944, opened 2016-01-25
Rename renter file commands

VoidingWarranties:
delete -> rm
rename -> mv
list -> ls

Most people using siac will be familiar with these commands. It would also be nice to add more features to make these commands to more closely mimic their unix equivalents, such as `ls`ing a specific directory.

### DIFF
--- a/siac/rentercmd.go
+++ b/siac/rentercmd.go
@@ -44,10 +44,11 @@ var (
 	}
 
 	renterFilesDeleteCmd = &cobra.Command{
-		Use:   "delete [path]",
-		Short: "Delete a file",
-		Long:  "Delete a file. Does not delete the file on disk.",
-		Run:   wrap(renterfilesdeletecmd),
+		Use:     "delete [path]",
+		Aliases: []string{"rm"},
+		Short:   "Delete a file",
+		Long:    "Delete a file. Does not delete the file on disk.",
+		Run:     wrap(renterfilesdeletecmd),
 	}
 
 	renterFilesDownloadCmd = &cobra.Command{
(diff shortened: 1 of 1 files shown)


lukechampine:
:+1:

DavidVorick:
I am reluctant, but not opposed.

DavidVorick:
perhaps it makes sense to support both styles?

LuccoJ:
I don't think siac only runs on Unix-style systems...?
Generally speaking, self-descriptive commands seem better than cryptic ones, even if the cryptic ones are actually well-known to a certain class of users.

pmknutsen:
No reason to not support both styles.

VoidingWarranties:
Both styles work. Cobra has command aliasing so that should be easy.

lukechampine:
awesome, I was going to suggest aliases but I didn't know cobra explicitly supported them

VoidingWarranties:
It would be nice if aliases showed up in `siac renter --help`.
Maybe something like
```
$ siac renter --help
Upload, download, rename, delete, load, or share files.

Usage:
  siac renter [flags]
  siac renter [command]

Available Commands:
  delete / rm     Delete a file
  download        Download a file
  downloads       View the download queue
  list / ls       List the status of all files
  load            Load a .sia file
  loadascii       Load an ASCII-encoded .sia file
  rename / mv     Rename a file
  share           Export a file to a .sia for sharing
  shareascii      Prints an ASCII-encoded .sia file for sharing
  upload          Upload a file
  uploads         View the upload queue

Global Flags:
  -a, --addr string   which host/port to communicate with (i.e. the host/port siad is listening on) (default "localhost:9980")

Use "siac renter [command] --help" for more information about a command.
```

But instead they only show up in the individual command helps, e.g. `siac renter list --help`.

I think cobra lets you customize help messages. I'll see if I can add it in another PR.

siac/rentercmd.go
```diff
@@ -271,7 +274,7 @@ func renterfilesloadasciicmd(ascii string) {
 	}
 }
 
-// renterfilesrenamecmd is the handler for the command `siac renter rename [path] [newpath]`.
+// renterfilesrenamecmd is the handler for the command `siac renter mv [path] [newpath]`.
```
DavidVorick: comment should not talk about the aliases, should talk about the primary. Same for the other 3

VoidingWarranties:
Oops I thought I reverted that. Fixed.

siac/rentercmd.go
VoidingWarranties: Oops I thought I reverted that. Fixed.

PR #945, opened 2016-01-25
Fix host desynchronization

DavidVorick:
Most of the major logging errors we've been seeing in the host should be resolved now. Additionally, the host has been altered so that the log does not grow as quickly, with the slowdown being a factor of 200x. We can continue to experiment with loggings as we move forward. Some type of proper aggregation would be nice, but we need to make sure that the logs are reasonable in size under all circumstances.

The transaction pool will return a new type of error if the transaction is rejected on consensus-incompatibility grounds. I do believe that there is an error here, but I haven't had enough time to investigate.

When attempting a renew, the host used to not delete the copy it made if the negotiation failed for some reason. It's not tested, but the host should not correctly delete the file. It will be difficult to test while the renter renewal uses time-based renewing instead of push-based renewing (for example, checking every time there's a new block). I will try to write a test today though, this isn't behavior that I want to release untested.

There's some unfortunate stuff here, I'm trying to fix it, but this PR as-is should be safe enough for release.

### DIFF
--- a/modules/consensus/accept_bench_test.go
+++ b/modules/consensus/accept_bench_test.go
@@ -18,7 +18,7 @@ func BenchmarkAcceptEmptyBlocks(b *testing.B) {
 	if err != nil {
 		b.Fatal("Error creating tester: " + err.Error())
 	}
-	defer cst.closeCst()
+	defer cst.Close()
 
 	// Create an alternate testing consensus set, which does not
 	// have any subscribers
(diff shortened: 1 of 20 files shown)


DavidVorick:
OK, the bugs I'm aware of are fixed properly now.

There is still not a test to make sure that the renew -> delete code is working correctly. But, it's a pretty simple addition and the implementation appears correct. I'll try and get a test for it.

lukechampine:
build failed with:
`panic: threaded listen does not have access to host resources - is only called at startup`
but only on Mac.

PR #946, opened 2016-01-25
Siac error exit codes

VoidingWarranties:
Exit with a non 0 exit code on an invalid command, or error response from a valid command. Exits with exit code 64 for usage errors (taken from [sysexits.h](https://www.freebsd.org/cgi/man.cgi?query=sysexits)), and exit code 1 for general errors.

### DIFF
--- a/siac/hostcmd.go
+++ b/siac/hostcmd.go
@@ -56,7 +56,7 @@ func hostconfigcmd(param, value string) {
 		// convert price to hastings/byte/block
 		p, ok := new(big.Rat).SetString(value)
 		if !ok {
-			dieNoError("Could not parse price")
+			die("Could not parse price")
 		}
 		p.Mul(p, big.NewRat(1e24/1e9, 4320))
 		value = new(big.Int).Div(p.Num(), p.Denom()).String()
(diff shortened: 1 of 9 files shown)


siac/hostcmd.go
```diff
@@ -55,8 +56,7 @@ func hostconfigcmd(param, value string) {
 		// convert price to hastings/byte/block
 		p, ok := new(big.Rat).SetString(value)
 		if !ok {
-			fmt.Println("could not parse price")
-			return
+			dieNoError("Could not parse price")
```
lukechampine: why would an error not be returned here?

VoidingWarranties: `ok` is a `bool` not an error.

lukechampine: whoops, I was confused lol.
A better approach would be to make `die` variadic, like `Println`:
```go
func die(args ...interface{}) {
    fmt.Fprintln(os.Stderr, args...)
    os.Exit(exitCode)
}

Comment (VoidingWarranties) [inline siac/hostcmd.go]: That is much more elegant but it would require the callers to add "\nError:" in order to be equivalent to the current output.

What do you think about the following?
```go
func die(context string, err error) {
 	fmt.Fprintln(os.Stderr, context)
 	if (err != nil) {
		fmt.Fprintln(os.Stderr, "Error:", err)
	}
 	os.Exit(exitCodeGeneral)
 }
```

I'm in favor of keeping the current output style as it more closely matches cobra's output when you run an invalid command e.g. `siac foo`

lukechampine: maybe something like:
```go
// die writes its arguments to os.Stderr (separated by newlines), and
// then exits with the default error code.
func die(args ...interface{}) {
    for _, arg := range args {
        fmt.Fprintln(os.Stderr, arg)
    }
    os.Exit(exitCodeGeneral)
}
```
It is a bit unconventional though. In general I think it's best to leave formatting up to the caller, but this seems like a special case.

lukechampine: I don't see much value in prefacing errors with `Error:`. The usual Go style seems fine to me:
```go
$ siac host
Command failed: no response from daemon
```

siac/main.go
```diff
@@ -24,6 +25,13 @@ var (
 	renterShowHistory bool   // Show download history in addition to download queue.
 )
 
+// exit codes
+// inspired by sysexits.h
+const (
+	exitCodeGeneral = 1
+	exitCodeUsage   = 64
+)
```
DavidVorick: can you extend the comment here? It's not really enough, I don't want to go look up sysexits.h to understand why these numbers are chosen, especially if you are reserving the right to deviate in small ways (which I think is the smart thing to do).

PR #947, opened 2016-01-26
Release constants

DavidVorick:
Prepare for release.

More details in the commit messages.


modules/host/upload.go
```diff
@@ -18,7 +18,7 @@ import (
 
 const (
 	// maxRevisionSize is the maximum number of bytes added in a single revision
-	maxRevisionSize = 100e6 // 100 MB
+	maxRevisionSize = 1 << 28
```
lukechampine: need units

PR #950, opened 2016-01-26
save repairing file incrementally

lukechampine:
Files are now saved after each chunk is uploaded, and after each contract is renewed. (Previously they were only saved after repairing every incomplete chunk and renewing every expiring contract.) This should greatly reduce the frequency of "lost" file progress.


DavidVorick:
Thanks

PR #952, opened 2016-01-26
Deprecate instead of remove siac commands

VoidingWarranties:
The deprecated commands will not appear in help messages, and are only for backwards compatibility. They will still work but will print a message stating they are deprecated, and suggesting an alternative command to use instead.


DavidVorick:
We use the tag // DEPRECATED across the project, makes it both stand out and easy to grep.

And, we also typically tag a version as well. // DEPRECATED v0.5.0

That tells us when it's safe to remove the code.

PR #956, opened 2016-01-27
host will not record failure for unconfirmed obligations

DavidVorick:
If an obligation is doublespent or is otherwise prevented from making it
onto the blockchain, this is not really a failure for the host.

### DIFF
--- a/modules/host/obligations.go
+++ b/modules/host/obligations.go
@@ -16,11 +16,16 @@ const (
 	// resubmissionTimeout is the number of blocks that the host will wait
 	// before trying to resubmit a transaction to the blockchain.
 	resubmissionTimeout = 2
+)
 
-	// Booleans to indicate that a contract obligation has been successful or
-	// unsuccessful.
-	obligationSucceeded = true
-	obligationFailed    = false
+const (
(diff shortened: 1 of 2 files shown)


modules/host/obligations.go
```diff
@@ -16,11 +16,14 @@ const (
...
-	// Booleans to indicate that a contract obligation has been successful or
-	// unsuccessful.
-	obligationSucceeded = true
-	obligationFailed    = false
+const (
+	// Iota to indicate the status of an obligation upon removal.
+	obligationUnspecified = iota
+	obligationUnconfirmed = iota
+	obligationSucceeded   = iota
+	obligationFailed      = iota
```
lukechampine: only the first `iota` is needed.
the comment is kinda strange too

DavidVorick:
fixed.

DavidVorick:
uh, Github doesn't seem to be seeing the changes, though the cli assures me that the branch was pushed.

DavidVorick:
fixed

lukechampine:
why do the Mac builds keep failing, but not Linux? Very odd.

PR #957, opened 2016-01-27
Siad -M, --modules flag

VoidingWarranties:
Specify modules for siad to load using the `-M` or `--modules` flag. Modules are specified by the first letter of the name. E.g. the default modules loaded are consensus (c), gateway (g), host (h), miner (m), renter (r), transaction pool (t), and wallet (w). This is equivalent to `siad --modules cghmrtw`.

On startup a status message is printed as each module is loaded. E.g.
```
$ siad
Loading...
(1/7) Loading gateway...
(2/7) Loading consensus...
(3/7) Loading transaction pool...
(4/7) Loading wallet...
(5/7) Loading miner...
(6/7) Loading host...
(7/7) Loading renter...
Finished loading in 167.819140245 seconds
```

### DIFF
--- a/siad/daemon.go
+++ b/siad/daemon.go
@@ -3,6 +3,7 @@ package main
 import (
 	"errors"
 	"fmt"
+	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -47,7 +48,7 @@ func processConfig(config Config) (Config, error) {
 		invalidModules = strings.Replace(invalidModules, string(m), "", 1)
 	}
 	if len(invalidModules) > 0 {
-		return Config{}, errors.New("Unrecognized modules: " + invalidModules)
+		return Config{}, errors.New("Unable to parse --modules flag, unrecognized modules: " + invalidModules)
 	}
 	return config, nil
 }
@@ -61,7 +62,7 @@ func startDaemon(config Config) (err error) {
 	// Create all of the modules.
 	var g modules.Gateway
 	if strings.Contains(config.Siad.Modules, "g") {
-		fmt.Print("Loading gateway...")
+		fmt.Println("Loading gateway...")
 		g, err = gateway.New(config.Siad.RPCaddr, filepath.Join(config.Siad.SiaDir, modules.GatewayDir))
 		if err != nil {
 			return err
@@ -164,7 +165,7 @@ func startDaemon(config Config) (err error) {
 }
 
 // startDaemonCmd is a passthrough function for startDaemon.
-func startDaemonCmd(*cobra.Command, []string) {
+func startDaemonCmd(cmd *cobra.Command, _ []string) {
 	// Create the profiling directory if profiling is enabled.
 	if globalConfig.Siad.Profile {
 		go profile.StartContinuousProfile(globalConfig.Siad.ProfileDir)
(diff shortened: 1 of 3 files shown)


siad/daemon.go
```diff
@@ -34,80 +37,127 @@ func processNetAddr(addr string) string {
...
-	return config
+	config.Siad.Modules = strings.ToLower(config.Siad.Modules)
+	validModules := "cghmrtwe"
+	invalidModules := config.Siad.Modules
+	for _, m := range validModules {
+		invalidModules = strings.Replace(invalidModules, string(m), "", 1)
+	}
+	if len(invalidModules) > 0 {
+		return Config{}, errors.New("Unable to parse --modules flag, unrecognized modules: " + invalidModules)
+	}
```
lukechampine: I think this will break if you supplied the same letter twice. This is kind of a hairy problem. Hmm

VoidingWarranties: I think it should fail if the same letter is supplied twice. No reason to support typos.

siad/daemon.go
```diff
@@ -32,82 +35,141 @@ func processNetAddr(addr string) string {
...
+// invalid module character.
+func processModules(modules string) (string, error) {
+	modules = strings.ToLower(modules)
+	validModules := "cghmrtwe"
+	invalidModules := modules
+	for _, m := range validModules {
+		invalidModules = strings.Replace(invalidModules, string(m), "", 1)
+	}
+	if len(invalidModules) > 0 {
+		return "", errors.New("Unable to parse --modules flag, unrecognized modules: " + invalidModules)
```
lukechampine: this error will be returned if you supply duplicate modules. Kind of misleading the modules aren't unrecognized, just duplicates.

VoidingWarranties: I could clarify the error message. What do you think about "... unrecognized or duplicate modules..."?

lukechampine: that should be fine

siad/daemon_test.go
```diff
@@ -21,19 +21,67 @@ func TestUnitProcessNetAddr(t *testing.T) {
...
+		if out != testVal.out {
+			t.Errorf("processModules returned incorrect modules: expected %s, got %s\n", testVal.out, out)
+		}
+	}
+
+	// Test invalid modules.
+	invalidModules := []string{"abdfijklnopqsuvxyz", "cghmrtwez", "cz", "z"}
+	for _, invalidModule := range invalidModules {
+		_, err := processModules(invalidModule)
+		if err == nil {
```
DavidVorick: I think it's better practice to check for a specific error. At the top of the file with the logic, declare an error:

```
var (
   ErrBadModuleString = errors.New("Unable to parse --modules flag, unrecognized or duplicate modules")
)
```

I still don't have a clean way for doing that with custom strings though.

siad/daemon_test.go
```diff
@@ -21,19 +21,67 @@ func TestUnitProcessNetAddr(t *testing.T) {
...
+		if err != nil {
+			t.Error("processModules failed with error:", err)
+		}
+		if out != testVal.out {
+			t.Errorf("processModules returned incorrect modules: expected %s, got %s\n", testVal.out, out)
+		}
+	}
+
+	// Test invalid modules.
+	invalidModules := []string{"abdfijklnopqsuvxyz", "cghmrtwez", "cz", "z"}
```
DavidVorick: need to test for duplicates.

siad/daemon_test.go
```diff
@@ -52,4 +103,12 @@ func TestUnitProcessConfig(t *testing.T) {
 			t.Error("processing failure at check", i, 2)
 		}
 	}
+
+	// Test invalid configs.
+	invalidModule := "z"
+	config.Siad.Modules = invalidModule
+	_, err := processConfig(config)
+	if err == nil {
```
DavidVorick: I would prefer to see a specific error here. I know the std library has a function "os.IsErrNotExist()" or something similar, maybe we can manage customized errors that way. I think you would want to do it by making a specific type + interface, though it would be nice if there was a less cody way of accomplishing that.

DavidVorick:
only blocking issue is the test for duplicates.

siad/daemon_test.go
VoidingWarranties: The error isn't being percolated up, very far or to any other packages so I don't think this is necessary. I've thought about named errors with custom strings, but the only way I could come up with it is to use custom error types and do type assertions.

PR #964, opened 2016-01-31
Siac renter verbose filelist

VoidingWarranties:
`siac renter list [-v, --verbose]` prints a tabular file list with more information such as File size, Availability, Progress, Redundancy, Renewing, and Sia path,

Redundancy is the redundancy of the least redundant chunk.

### DIFF
--- a/modules/renter.go
+++ b/modules/renter.go
@@ -52,6 +52,7 @@ type FileInfo struct {
 	Filesize       uint64            `json:"filesize"`
 	Available      bool              `json:"available"`
 	Renewing       bool              `json:"renewing"`
+	Redundancy     float64           `json:"redundancy"`
 	UploadProgress float64           `json:"uploadprogress"`
 	Expiration     types.BlockHeight `json:"expiration"`
 }
--- a/modules/renter/files.go
+++ b/modules/renter/files.go
@@ -107,6 +107,21 @@ func (f *file) uploadProgress() float64 {
 	return 100 * (float64(uploaded) / float64(desired))
 }
 
+// redundancy returns the average redundancy of a file across all chunks.
+// Because average redundancy is calculated, a redundancy >= 1 does not mean
+// all chunks have been uploaded.
+func (f *file) redundancy() float64 {
+	f.mu.RLock()
+	defer f.mu.RUnlock()
+	var uploaded uint64
+	for _, fc := range f.contracts {
+		uploaded += uint64(len(fc.Pieces)) * f.pieceSize
+	}
+	// f.size is not the padded size.
+	paddedSize := f.chunkSize() * f.numChunks()
+	return float64(uploaded) / float64(paddedSize)
+}
+
 // expiration returns the lowest height at which any of the file's contracts
 // will expire.
 func (f *file) expiration() types.BlockHeight {
@@ -171,6 +186,7 @@ func (r *Renter) FileList() []modules.FileInfo {
 			SiaPath:        f.name,
 			Filesize:       f.size,
 			Available:      f.available(),
+			Redundancy:     f.redundancy(),
 			Renewing:       renewing,
 			UploadProgress: f.uploadProgress(),
 			Expiration:     f.expiration(),
--- a/modules/renter/files.go
+++ b/modules/renter/files.go
@@ -6,6 +6,7 @@ import (
 	"path/filepath"
 	"sync"
 
+	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/types"
@@ -107,13 +108,13 @@ func (f *file) uploadProgress() float64 {
 	return 100 * (float64(uploaded) / float64(desired))
 }
 
-// redundancy returns the redundancy of the least redundant chunk. -1 is
-// returned if the file has 0 chunks. A file becomes available when this
-// redundancy is >= 1. Assumes that every piece is unique within a file
-// contract.
+// redundancy returns the redundancy of the least redundant chunk. A file
+// becomes available when this redundancy is >= 1. Assumes that every piece is
+// unique within a file contract.
 func (f *file) redundancy() float64 {
 	if f.numChunks() == 0 {
-		return -1
+		build.Critical("cannot get redundancy of a file with 0 chunks")
+		return 0
 	}
 	piecesPerChunk := make([]int, f.numChunks())
 	for _, fc := range f.contracts {
--- a/modules/renter/files.go
+++ b/modules/renter/files.go
@@ -2,6 +2,7 @@ package renter
 
 import (
 	"errors"
+	"math"
 	"os"
 	"path/filepath"
 	"sync"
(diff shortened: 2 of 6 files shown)


modules/renter/files.go
```diff
@@ -107,6 +108,29 @@ func (f *file) uploadProgress() float64 {
...
 }
 
+// redundancy returns the redundancy of the least redundant chunk. A file
+// becomes available when this redundancy is >= 1. Assumes that every piece is
+// unique within a file contract.
+func (f *file) redundancy() float64 {
+	if f.numChunks() == 0 {
+		build.Critical("cannot get redundancy of a file with 0 chunks")
+		return 0
+	}
```
DavidVorick: What guarantees are there that you will never call redundancy on a file with 0 chunks? Is it true that all files, even empty files, are going to have at least one chunk?

modules/renter/files_test.go
```diff
@@ -59,6 +59,39 @@ func TestFileAvailable(t *testing.T) {
...
+			t.Error("expected 0 redundancy, got", r)
+		}
+
+		testRedundancies := []uint64{1, 6, 10}
+		for _, testR := range testRedundancies {
+			var fc fileContract
+			for i := uint64(0); i < f.numChunks(); i++ {
+				for j := uint64(0); j < testR; j++ {
+					fc.Pieces = append(fc.Pieces, pieceData{Chunk: i, Piece: j})
+				}
```
DavidVorick: loop nesting this deep makes the test difficult to understand. There's probably a better way to approach this test.

modules/renter/files_test.go
```diff
@@ -59,6 +59,39 @@ func TestFileAvailable(t *testing.T) {
...
+			}
+			f.contracts = map[types.FileContractID]fileContract{types.FileContractID{}: fc}
+			expectedR := float64(testR) / float64(nData)
+			r = f.redundancy()
+			if r != expectedR {
+				t.Errorf("expected %f redundancy, got %f", expectedR, r)
+			}
+		}
+	}
+}
```
DavidVorick: In general I find this test confusing and can't tell if it's covering all the test cases. Though I could look at it some more, I think it would be better if it were written to be more understandable on the first pass.

siac/main.go
```diff
@@ -160,6 +161,14 @@ func die(args ...interface{}) {
 	os.Exit(exitCodeGeneral)
 }
 
+// yesNo returns "Yes" if b is true, and "No" if b is false.
+func yesNo(b bool) string {
```
DavidVorick: There is probably a better name for this function. Maybe bToS? or just boolToString

lukechampine: I considered that as well, but my first assumption for `boolToString` would be `"true"` and `"false"`.

DavidVorick: `humanReadableBool`?

lukechampine: I don't see how that's any different.
`yesNo` is fine for now

DavidVorick:
requires rebase

modules/renter.go
```diff
@@ -52,6 +52,7 @@ type FileInfo struct {
 	Filesize       uint64            `json:"filesize"`
 	Available      bool              `json:"available"`
 	Renewing       bool              `json:"renewing"`
+	Redundancy     float64           `json:"redundancy"`
 	UploadProgress float64           `json:"uploadprogress"`
```
DavidVorick: is it redundant to have both `Redundancy` and `UploadProgress`? I feel like they are the same thing. If you are always displaying the progress of the minimal chunk, then you have the progress of everything has a whole.

Or, do they have different behaviors? If they do, this should be documented somewhere.

DavidVorick: such as in doc/API.md, as well as where the code is written

modules/renter/files.go
```diff
@@ -107,6 +108,34 @@ func (f *file) uploadProgress() float64 {
...
+// unique within a file contract. -1 is returned if the file has size 0.
+func (f *file) redundancy() float64 {
+	if f.size == 0 {
+		return -1
+	}
+	piecesPerChunk := make([]int, f.numChunks())
+	// We cannot get the redundancy if piecesPerChunk has 0 length because later
+	// in the function we use the first element of the slice.
+	if len(piecesPerChunk) == 0 {
+		build.Critical("cannot get redundancy of a file with 0 chunks")
```
DavidVorick: I think that this has to be changed, because it is Sia's plan to support 0 size files, which means they don't need to be uploaded, which means their redundancy is... undefined. Returning -1 or a `-` is probably the right thing to do here.

modules/renter.go
DavidVorick: It might be confusing to have each number represent something different. I think I'd prefer to use `Redundancy` alone, as it gives the user a picture that is more relevant to them.

modules/renter/files.go
```diff
@@ -107,6 +108,34 @@ func (f *file) uploadProgress() float64 {
...
+		for _, p := range fc.Pieces {
+			piecesPerChunk[p.Chunk]++
+		}
+	}
+	minPieces := piecesPerChunk[0]
+	for _, numPieces := range piecesPerChunk {
+		if numPieces < minPieces {
+			minPieces = numPieces
+		}
+	}
```
DavidVorick: I am worried about the runtime and memory costs for doing all of these operations for every file when something like `files` is queried through the API. I'm wondering if we shouldn't limit the API to serving one folder at a time.

VoidingWarranties: That would be in line with adding `ls` like features.

modules/renter/files.go
VoidingWarranties: The line a few above this one does return -1 for 0 sized files. I'll change this block to also return -1 instead of 0 to match the behavior of the block above.

Actually I was about to change it to return NaN to match the behavior of uploadProgress for 0 sized files, but that's another PR.

modules/renter.go
VoidingWarranties: They are redundant. But user's expect to see an upload progress. We could calculate the upload progress from the redundancy in the client, but then the target redundancy would need to be known to the client.

modules/renter/files.go
lukechampine: caching is trivial to add later. We could add a `cachedRedundancy` field to the file object and update it after each upload.

modules/renter.go
lukechampine: `UploadProgress` has higher granularity. With `Redundancy`, the step size is 1/TotalRedundancy, so about 5% with 24 pieces. With `UploadProgress`, it's technically 1 byte, but we only update after each chunk, so in reality it's 1/NumChunks (with redundancy).
So the issue is that with a sufficiently large file, its redundancy will appear stuck at 0 for a long time.

modules/renter/files.go
lukechampine: `NaN` sounds good to me

modules/renter.go
DavidVorick: oh, that's a good point. It'll say Upload 0 until the upload loop has gotten to every single chunk, at which point you are hopefully already at full availability and then some. So you jump from 0 to like 1.5 or something. Which is not acceptable UX. So I think it makes a lot of sense to have both in that context.

PR #965, opened 2016-02-01
Code Quality

DavidVorick:
While trying to get an overview of the host, I ran into a few things that I though should be fixed. Commits go into detail on all of the changes.

### DIFF
--- a/api/host_test.go
+++ b/api/host_test.go
@@ -65,7 +65,7 @@ func TestIntegrationHosting(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	expRevenue := "15307662222190387473"
+	expRevenue := "15307662222222127617"
 	if hg.Revenue.String() != expRevenue {
 		t.Fatalf("host's profit was not affected: expected %v, got %v", expRevenue, hg.Revenue)
 	}
--- a/modules/host.go
+++ b/modules/host.go
@@ -136,3 +136,36 @@ type (
 		Close() error
 	}
 )
+
+// StoragePriceToConsensus converts a human storage price, having the unit
+// 'Siacoins Per Month Per Terabyte', to a consensus storage price, having the
+// unit 'Hastings Per Block Per Byte'.
+func StoragePriceToConsensus(siacoinsMonthTB uint64) (hastingsBlockByte types.Currency) {
+	// Perform multiplcation first to preserve precision.
+	hastingsMonthTB := types.NewCurrency64(siacoinsMonthTB).Mul(types.SiacoinPrecision)
+	hastingsBlockTB := hastingsMonthTB.Div(types.NewCurrency64(4320))
+	hastingsBlockByte = hastingsBlockTB.Div(types.NewCurrency64(1e12))
+	return hastingsBlockByte
+}
+
+// StoragePriceToHuman converts a consensus storage price, having the unit
+// 'Hastings Per Block Per Byte', to a human storage price, having the unit
+// 'Siacoins Per Month Per Terabyte'. An error is returned if the result would
+// overflow a uint64. If the result is between 0 and 1, the value is rounded to
+// the nearest value.
+func StoragePriceToHuman(hastingsBlockByte types.Currency) (siacoinsMonthTB uint64, err error) {
+	// Perform multiplcation first to preserve precision.
+	hastingsMonthByte := hastingsBlockByte.Mul(types.NewCurrency64(4320))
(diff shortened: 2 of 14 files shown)


api/host_test.go
```diff
@@ -65,7 +65,7 @@ func TestIntegrationHosting(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	expRevenue := "15307662222190387473"
+	expRevenue := "15307662222222127617"
```
lukechampine: why did this change?

modules/host.go
```diff
@@ -128,3 +136,36 @@ type (
...
+	return hastingsBlockByte
+}
+
+// StoragePriceToHuman converts a consensus storage price, having the unit
+// 'Hastings Per Block Per Byte', to a human storage price, having the unit
+// 'Siacoins Per Month Per Terabyte'. An error is returned if the result would
+// overflow a uint64. If the result is between 0 and 1, the value is rounded to
+// the nearest value.
+func StoragePriceToHuman(hastingsBlockByte types.Currency) (siacoinsMonthTB uint64, err error) {
+	// Perform multiplcation first to preserve precision.
```
lukechampine: typo

modules/host_test.go
```diff
@@ -0,0 +1,101 @@
...
+	// because the file contract transaction is going to grow as the terms are
+	// negotiated and as signatures are added.
+	if MaxFileContractSetLen > TransactionSetSizeLimit-1e3 {
+		t.Fatal("MaxfileContractSetLen does not have a sensible value - should be smaller than the TransactionSetSizeLimit")
+	}
+
+}
+
+// TestUnitStoragePriceConversions checks the functions StoragePriceToHuman and
+// StoragePriceToConsensus, verifiying that they correclty convert between
```
lukechampine: two typos

modules/host_test.go
```diff
@@ -0,0 +1,101 @@
...
+			// Convert from types.SiacoinPrecision - simple result.
+			types.SiacoinPrecision,
+			4320e12,
+		},
+	}
+
+	// Run all of the trials.
+	for i, trial := range trials {
+		// Convert from the consensus result to the human result, and
+		// vice-versa. The transformations should be communitive, so both
```
lukechampine: typo. But the relationship is more inverse than commutative.

types/currency.go
```diff
@@ -151,15 +152,25 @@ func (x Currency) Sqrt() (c Currency) {
...
+// function is sometimes called on values that can be determined by users -
+// rather than have all user-facing points do input checking, the input
+// checking should happen at the base type. This minimizes the chances of a
+// rogue user causing a build.Critical to be triggered.
+func (c Currency) Uint64() (u uint64, err error) {
+	if c.Cmp(NewCurrency64(math.MaxUint64)) > 0 {
+		return 0, ErrUint64Overflow
+	}
+	return uint64(c.Big().Int64()), nil
+}
```
lukechampine: `c.Big().Uint64()`

types/currency_test.go
```diff
@@ -298,3 +299,30 @@ func TestNegativeNewCurrency(t *testing.T) {
 	negBig := big.NewInt(-1)
 	_ = NewCurrency(negBig)
 }
+
+// TestCurrencyUint64 tests that a currency is correctly converted to a uint64.
+func TestCurrencyUint64(t *testing.T) {
+	// Try a set of valid values.
+	attempts := []uint64{0, 1, 2, 3, 4, 25e3, math.MaxUint64}
```
lukechampine: not a great name. I usually just use `tests`, but you could also use `u64s`, `currency64s`, `values`, etc.

api/host_test.go
DavidVorick: it's mentioned in the commit message for human readable conversions. The conversion that's happening now has a higher precision (does all the multiplies first, then the divides second) than previously (do a mult, then div, then mult), which resulted in the value changing slightly.

modules/host_test.go
DavidVorick: ah man. Is there a `gospell`? heh

DavidVorick: hmm. I've been spelling commutative wrong my whole life. I thought for sure it was communative. I've probably been saying it wrong my whole life too.

types/currency.go
DavidVorick: heh

DavidVorick: and, my implementation is incorrect anyway as an int64 isn't as large as a uint64, so I definitely missed an edge case here during testing.

DavidVorick:
fixed up

PR #967, opened 2016-02-02
Moving things around

DavidVorick:
I moved some things around, and updated the profiling.

This PR is largely uninteresting.

### DIFF
--- a/modules/host/persist.go
+++ b/modules/host/persist.go
@@ -26,6 +26,15 @@ var persistMetadata = persist.Metadata{
 
 // persistence is the data that is kept when the host is restarted.
 type persistence struct {
+	// RPC Metrics.
+	ErroredCalls      uint64
+	UnrecognizedCalls uint64
+	DownloadCalls     uint64
+	RenewCalls        uint64
+	ReviseCalls       uint64
+	SettingsCalls     uint64
+	UploadCalls       uint64
+
 	// Consensus Tracking.
 	BlockHeight  types.BlockHeight
 	RecentChange modules.ConsensusChangeID
@@ -42,15 +51,6 @@ type persistence struct {
 	FileCounter int64
 	LostRevenue types.Currency
 	Revenue     types.Currency
-
-	// RPC Metrics.
-	ErroredCalls      uint64
-	UnrecognizedCalls uint64
-	DownloadCalls     uint64
-	RenewCalls        uint64
-	ReviseCalls       uint64
-	SettingsCalls     uint64
-	UploadCalls       uint64
 
 	// Utilities.
 	Settings modules.HostSettings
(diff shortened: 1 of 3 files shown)


modules/host/persist.go
```diff
@@ -201,27 +201,3 @@ func (h *Host) load() error {
...
-		return err
-	}
-
-	// Load the prior persistance structures.
-	err = h.load()
-	if err != nil {
-		return err
-	}
-	return nil
-}
```
DavidVorick: moved initPersist into New, flattening it. I think you asked me to do this in a much earlier PR, but I have only recently decided that it's the right thing to do.

modules/host/persist.go
```diff
@@ -161,15 +170,6 @@ func (h *Host) load() error {
...
 
-	// Copy over rpc tracking.
-	atomic.StoreUint64(&h.atomicErroredCalls, p.ErroredCalls)
-	atomic.StoreUint64(&h.atomicUnrecognizedCalls, p.UnrecognizedCalls)
-	atomic.StoreUint64(&h.atomicDownloadCalls, p.DownloadCalls)
-	atomic.StoreUint64(&h.atomicRenewCalls, p.RenewCalls)
-	atomic.StoreUint64(&h.atomicReviseCalls, p.ReviseCalls)
-	atomic.StoreUint64(&h.atomicSettingsCalls, p.SettingsCalls)
-	atomic.StoreUint64(&h.atomicUploadCalls, p.UploadCalls)
-
```
DavidVorick: just moved these to the top so they match the layout of the host struct

siad/daemon.go
```diff
@@ -189,7 +189,7 @@ func startDaemon(config Config) (err error) {
 func startDaemonCmd(cmd *cobra.Command, _ []string) {
 	// Create the profiling directory if profiling is enabled.
 	if globalConfig.Siad.Profile {
-		go profile.StartContinuousProfile(globalConfig.Siad.ProfileDir)
+		go profile.StartContinuousProfile(globalConfig.Siad.ProfileDir, false, false)
```
DavidVorick: by default the profiler won't grab anything. I'm almost thinking that we shouldn't have this be a flag, and just have it be a build tag thing. If you build with the 'profile' build tag, you'll get a profiler on port 10501, I don't see why you shouldn't also get the continuous profiling stuff and then the whole profile package can just be left out of the normal binaries.

profile/profile.go
```diff
@@ -80,35 +81,56 @@ func SaveMemProfile(profileDir, identifier string) error {
...
+
+			if memProfiling {
+				// If performing memory profiling, grab a profile of the
+				// current memory structure.
+				err = SaveMemProfile(profileDir, identifier)
+				if err != nil {
+					fmt.Println("memory logging failed:", err)
+				}
+			}
+
```
DavidVorick: I extended the profiling capabilities a bit, though honestly I haven't found these extensions to be that useful

DavidVorick: just the http/pprof stuff has been equivalent/sufficient.

DavidVorick: the more I think about it the more I want to remove these things. Hmm. I'll remove them.

profile/profile.go
```diff
@@ -80,35 +81,56 @@ func SaveMemProfile(profileDir, identifier string) error {
...
 		return
 	}
 
-	// Create a logger to infrequently log performance statistics.
+	// Continuously log statistics about the running Sia application.
 	go func() {
-		// Create a logger for the goroutine.
-		logFile, err := os.OpenFile(filepath.Join(profileDir, "continuousProfiling.log"), os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0660)
+		// Create the logger.
+		log, err := persist.NewLogger(filepath.Join(profileDir, "continuousProfiling.log"))
```
DavidVorick: switched from a custom logger to a persist.Logger. Simplifies things

siad/daemon.go
lukechampine: yeah, I don't think there's much value in providing profiling capability in our release binaries. Profiling is generally something you only do after you've identified a problem. (As opposed to logging, which you do before you identify the problem.)

lukechampine:
your rebase got messed up
probably because I amended https://github.com/NebulousLabs/Sia/commit/6623315af735247ccb4f8645b73ebee23724ae5a
sry

DavidVorick:
fixed

PR #969, opened 2016-02-03
Don't mask networking errors in the host logs

robvanmieghem:
Signed-off-by: Rob Van Mieghem <rob@vanmieghemcloud.com>


lukechampine:
good catch!

DavidVorick:
Ah this bug is my mistake. Thanks for cleaning it up!

PR #971, opened 2016-02-04
Check error returned from ConsensusSetPersistentSubscribe

VoidingWarranties:
Because all modules (but the miner) subscribe to the consensus set with an initial change id of the genesis block, it's only possible for an error to be returned by `ConsensusSetPersistentSubscribe` right now if `computeConsensusChange` fails. In the future these modules will use a more recent `ConsensusChangeID` which could return make `ConsensusSetPersistentSubscribe` return `ErrInvalidConsensusChangeID` if the module and the consensus package become desynchronized.

### DIFF
--- a/modules/explorer/explorer.go
+++ b/modules/explorer/explorer.go
@@ -128,7 +128,10 @@ func New(cs modules.ConsensusSet, persistDir string) (*Explorer, error) {
 		return nil, err
 	}
 
-	cs.ConsensusSetPersistentSubscribe(e, modules.ConsensusChangeID{})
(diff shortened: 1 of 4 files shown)


DavidVorick:
needs rebase

DavidVorick:
also host can also manage the error.

VoidingWarranties:
Rebased.

> also host can also manage the error.

What do you mean?

DavidVorick:
> Because all modules (but the miner) subscribe to the consensus set with an initial change id of the genesis block

Host also subscribes to the most recent height, it's not just the miner :)

PR #972, opened 2016-02-04
Sort files in `siac renter list` by sia path

### LINKED ISSUE #963: `siac renter` sort output by filename

VoidingWarranties:
Closes #963

I wasn't sure whether to sort in the client or the daemon. I decided on sorting in the client so that in the future the field sorted on can be specified without modifying the API.


lukechampine:
:+1:

PR #974, opened 2016-02-04
Reduce the number of unnecessary block broadcasts

VoidingWarranties:
It is only necessary to broadcast the last block in a chain of calls to
AcceptBlock. This commit changes the behavior of threadedReceiveBlocks
to only broadcast the last block received. This is especially helpful
during the initial blockchain download. Instead of rebroadcasting every
block, only the last block in the initial blockchain download is
broadcasted.

Here's the gateway log after an initial blockchain download. Only one RelayBlock broadcast :)
```
$ cat gateway.log | grep RelayBlock
2016/02/04 17:32:26.716491 rpc.go:133: INFO: broadcasting RPC "RelayBlock" to 8 peers
```

### DIFF
--- a/modules/consensus/accept.go
+++ b/modules/consensus/accept.go
@@ -120,9 +120,15 @@ func (cs *ConsensusSet) addBlockToTree(b types.Block) (ce changeEntry, err error
 	return ce, nil
 }
 
-// managedAcceptBlock accepts a block but does not broadcast it to any peers.
-// See comment for AcceptBlock. Typically AcceptBlock should be used. This
-// method should only be used when there would otherwise be multiple
+// managedAcceptBlock will try to add a block to the consensus set. If the
+// block does not extend the longest currently known chain, an error is
+// returned but the block is still kept in memory. If the block extends a fork
+// such that the fork becomes the longest currently known chain, the consensus
+// set will reorganize itself to recognize the new longest fork. Accepted
+// blocks are not relayed.
+//
+// Typically AcceptBlock should be used so that the accepted block is relayed.
+// This method is typically only be used when there would otherwise be multiple
 // consecutive calls to AcceptBlock with each successive call accepting the
 // child block of the previous call.
 func (cs *ConsensusSet) managedAcceptBlock(b types.Block) error {
@@ -183,10 +189,13 @@ func (cs *ConsensusSet) managedAcceptBlock(b types.Block) error {
 	return nil
 }
 
-// AcceptBlock will add a block to the state, forking the blockchain if it is
-// on a fork that is heavier than the current fork. If the block is accepted,
-// it will be relayed to connected peers. This function should only be called
-// for new, untrusted blocks.
+// AcceptBlock will try to add a block to the consensus set. If the block does
+// not extend the longest currently known chain, an error is returned but the
+// block is still kept in memory. If the block extends a fork such that the
+// fork becomes the longest currently known chain, the consensus set will
+// reorganize itself to recognize the new longest fork. If a block is accepted
+// without error, it will be relayed to all connected peers. This function
+// should only be called for new blocks.
 func (cs *ConsensusSet) AcceptBlock(b types.Block) error {
 	err := cs.managedAcceptBlock(b)
 	if err != nil {
--- a/modules/consensus/accept.go
+++ b/modules/consensus/accept.go
@@ -120,11 +120,12 @@ func (cs *ConsensusSet) addBlockToTree(b types.Block) (ce changeEntry, err error
 	return ce, nil
 }
 
-// AcceptBlock will add a block to the state, forking the blockchain if it is
-// on a fork that is heavier than the current fork. If the block is accepted,
-// it will be relayed to connected peers. This function should only be called
-// for new, untrusted blocks.
-func (cs *ConsensusSet) AcceptBlock(b types.Block) error {
+// acceptBlockNoBroadcast accepts a block but does not broadcast it to any
+// peers. See comment for AcceptBlock. Typically AcceptBlock should be used.
+// This method should only be used when there would otherwise be multiple
+// consecutive calls to AcceptBlock with each successive call accepting the
+// child block of the previous call.
+func (cs *ConsensusSet) acceptBlockNoBroadcast(b types.Block) error {
 	// Grab a lock on the consensus set. Lock is demoted later in the function,
 	// failure to unlock before returning an error will cause a deadlock.
 	cs.mu.Lock()
@@ -179,9 +180,19 @@ func (cs *ConsensusSet) AcceptBlock(b types.Block) error {
 	if len(changeEntry.AppliedBlocks) > 0 {
 		cs.readlockUpdateSubscribers(changeEntry)
 	}
+	return nil
+}
 
+// AcceptBlock will add a block to the state, forking the blockchain if it is
+// on a fork that is heavier than the current fork. If the block is accepted,
+// it will be relayed to connected peers. This function should only be called
+// for new, untrusted blocks.
+func (cs *ConsensusSet) AcceptBlock(b types.Block) error {
+	err := cs.acceptBlockNoBroadcast(b)
+	if err != nil {
+		return err
+	}
 	// Broadcast the new block to all peers.
 	go cs.gateway.Broadcast("RelayBlock", b)
-
 	return nil
 }
--- a/modules/consensus/accept.go
+++ b/modules/consensus/accept.go
@@ -120,12 +120,12 @@ func (cs *ConsensusSet) addBlockToTree(b types.Block) (ce changeEntry, err error
 	return ce, nil
 }
 
-// acceptBlockNoBroadcast accepts a block but does not broadcast it to any
-// peers. See comment for AcceptBlock. Typically AcceptBlock should be used.
-// This method should only be used when there would otherwise be multiple
+// managedAcceptBlock accepts a block but does not broadcast it to any peers.
+// See comment for AcceptBlock. Typically AcceptBlock should be used. This
+// method should only be used when there would otherwise be multiple
 // consecutive calls to AcceptBlock with each successive call accepting the
 // child block of the previous call.
-func (cs *ConsensusSet) acceptBlockNoBroadcast(b types.Block) error {
+func (cs *ConsensusSet) managedAcceptBlock(b types.Block) error {
 	// Grab a lock on the consensus set. Lock is demoted later in the function,
 	// failure to unlock before returning an error will cause a deadlock.
 	cs.mu.Lock()
@@ -188,7 +188,7 @@ func (cs *ConsensusSet) acceptBlockNoBroadcast(b types.Block) error {
 // it will be relayed to connected peers. This function should only be called
 // for new, untrusted blocks.
 func (cs *ConsensusSet) AcceptBlock(b types.Block) error {
-	err := cs.acceptBlockNoBroadcast(b)
+	err := cs.managedAcceptBlock(b)
 	if err != nil {
 		return err
 	}
--- a/modules/consensus/accept_test.go
+++ b/modules/consensus/accept_test.go
@@ -3,6 +3,7 @@ package consensus
 import (
 	"bytes"
 	"errors"
+	"sync"
 	"testing"
 	"time"
 
@@ -706,9 +707,12 @@ func TestTaxHardfork(t *testing.T) {
 type mockGateway struct {
 	modules.Gateway
 	numBroadcasts int
+	mu            sync.RWMutex
 }
 
 func (g *mockGateway) Broadcast(string, interface{}) {
+	g.mu.Lock()
+	defer g.mu.Unlock()
 	g.numBroadcasts++
 	return
 }
@@ -721,7 +725,9 @@ func TestAcceptBlockBroadcasts(t *testing.T) {
 		t.Fatal(err)
 	}
 	defer cst.Close()
-	mg := &mockGateway{Gateway: cst.cs.gateway}
+	mg := &mockGateway{
+		Gateway: cst.cs.gateway,
+	}
 	cst.cs.gateway = mg
 
 	// Test that Broadcast is called for valid blocks.
@@ -734,31 +740,44 @@ func TestAcceptBlockBroadcasts(t *testing.T) {
 	// wait on a channel because we don't know how many times broadcast has
 	// been called.
 	time.Sleep(1)
-	if mg.numBroadcasts != 1 {
+	mg.mu.RLock()
+	numBroadcasts := mg.numBroadcasts
+	mg.mu.RUnlock()
+	if numBroadcasts != 1 {
 		t.Errorf("expected AcceptBlock to broadcast a valid block 1 time, instead it broadcasted %d times", mg.numBroadcasts)
 	}
 
 	// Test that Broadcast is not called for invalid blocks.
+	mg.mu.Lock()
 	mg.numBroadcasts = 0
+	mg.mu.Unlock()
 	err = cst.cs.AcceptBlock(types.Block{})
 	if err == nil {
 		t.Fatal("expected AcceptBlock to error on an invalid block")
 	}
 	// Sleep one second to wait for a possible call to g.Broadcast.
 	time.Sleep(1)
-	if mg.numBroadcasts != 0 {
+	mg.mu.RLock()
+	numBroadcasts = mg.numBroadcasts
+	mg.mu.RUnlock()
+	if numBroadcasts != 0 {
 		t.Error("AcceptBlock broadcasted an invalid block")
 	}
 
 	// Test that Broadcast is not called in managedAcceptBlock.
+	mg.mu.Lock()
 	mg.numBroadcasts = 0
+	mg.mu.Unlock()
 	b, _ = cst.miner.FindBlock()
 	err = cst.cs.managedAcceptBlock(b)
 	if err != nil {
 		t.Fatal(err)
 	}
 	time.Sleep(1)
-	if mg.numBroadcasts != 0 {
+	mg.mu.RLock()
+	numBroadcasts = mg.numBroadcasts
+	mg.mu.RUnlock()
+	if numBroadcasts != 0 {
 		t.Errorf("expected managedAcceptBlock to not broadcast any blocks, instead it broadcasted %d times", mg.numBroadcasts)
 	}
 }
--- a/modules/consensus/accept_test.go
+++ b/modules/consensus/accept_test.go
@@ -701,3 +701,64 @@ func TestTaxHardfork(t *testing.T) {
 		t.Fatal("siafund pool was not increased correctly")
 	}
 }
+
+// mockGateway implements modules.Gateway to mock the Broadcast method.
+type mockGateway struct {
+	modules.Gateway
+	numBroadcasts int
+}
+
+func (g *mockGateway) Broadcast(string, interface{}) {
+	g.numBroadcasts++
+	return
+}
+
+// TestAcceptBlockBroadcasts tests that AcceptBlock broadcasts valid blocks and
+// that managedAcceptBlock does not.
+func TestAcceptBlockBroadcasts(t *testing.T) {
+	cst, err := blankConsensusSetTester("TestAcceptBlockBroadcasts")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer cst.Close()
+	mg := &mockGateway{Gateway: cst.cs.gateway}
+	cst.cs.gateway = mg
+
+	// Test that Broadcast is called for valid blocks.
+	b, _ := cst.miner.FindBlock()
+	err = cst.cs.AcceptBlock(b)
+	if err != nil {
+		t.Fatal(err)
+	}
+	// Sleep to wait for possible calls to Broadcast to complete. We cannot
+	// wait on a channel because we don't know how many times broadcast has
+	// been called.
+	time.Sleep(1)
+	if mg.numBroadcasts != 1 {
+		t.Errorf("expected AcceptBlock to broadcast a valid block 1 time, instead it broadcasted %d times", mg.numBroadcasts)
+	}
+
+	// Test that Broadcast is not called for invalid blocks.
+	mg.numBroadcasts = 0
+	err = cst.cs.AcceptBlock(types.Block{})
+	if err == nil {
+		t.Fatal("expected AcceptBlock to error on an invalid block")
+	}
+	// Sleep one second to wait for a possible call to g.Broadcast.
+	time.Sleep(1)
+	if mg.numBroadcasts != 0 {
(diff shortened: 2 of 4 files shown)


modules/consensus/accept_test.go
```diff
@@ -701,3 +701,64 @@ func TestTaxHardfork(t *testing.T) {
...
+	// Test that Broadcast is called for valid blocks.
+	b, _ := cst.miner.FindBlock()
+	err = cst.cs.AcceptBlock(b)
+	if err != nil {
+		t.Fatal(err)
+	}
+	// Sleep to wait for possible calls to Broadcast to complete. We cannot
+	// wait on a channel because we don't know how many times broadcast has
+	// been called.
+	time.Sleep(1)
```
VoidingWarranties: I can't come up with a better way to solve this other than Sleep. If you wait on a channel, you must wait on it x number of times, where x is the same number of times you expect `gateway.Broadcast` to be called. But then `g.channel <- struct{}{}` will block if Broadcast is called more times than the `<-g.channel` is called. The channel could be buffered instead, but then waiting on the channel would only block until the expected number of broadcasts, after which there could be additional broadcasts.

lukechampine: why would you not know how many times it has been called? Isn't it just once per block?

VoidingWarranties: We are testing how many times it is called, so we do not know. We only know how many times we expect it to be called.

lukechampine: hmm. I guess a sleep is ok for now. Make it at least `10 * time.Millisecond` though, since Travis is sometimes slow.

DavidVorick: Sleep is a bad way to do this. Also, I think it's worth noting that `time.Sleep(1)` will sleep for 1 nanosecond.

I don't think the consensus set should be calling `Broadcast` in a gothread. I think it makes sense to let the gateway handle the calls, and be able to block the consensus set or the transaction pool if it's backlogged.

Otherwise, you can either wait until Broadcast has been called at least once, but it's difficult to provide any sort of guarantee that it will be called exactly once. That sort of guarantee couldn't come from unit testing, it'd need to come from static code analysis.

modules/consensus/accept.go
```diff
@@ -179,9 +180,19 @@ func (cs *ConsensusSet) AcceptBlock(b types.Block) error {
 	if len(changeEntry.AppliedBlocks) > 0 {
 		cs.readlockUpdateSubscribers(changeEntry)
 	}
+	return nil
+}
 
+// AcceptBlock will add a block to the state, forking the blockchain if it is
+// on a fork that is heavier than the current fork. If the block is accepted,
+// it will be relayed to connected peers. This function should only be called
+// for new, untrusted blocks.
```
DavidVorick: Wording is not super clear. I realize it's what I wrote, but today I'd probably rewrite it.

```go
// AcceptBlock will try to add a block to the consensus set. If the block does not extend the longest currently known chain, an error is returned but the block is still kept in memory. If the block extends a fork such that the fork becomes the longest currently known chain, the consensus set will reorganize itself to recognize the new longest fork. If a block is accepted without error, it will be relayed to all connected peers. This function should only be called for new blocks.
```

modules/consensus/accept_test.go
```diff
@@ -701,3 +702,82 @@ func TestTaxHardfork(t *testing.T) {
...
 }
+
+// mockGateway implements modules.Gateway to mock the Broadcast method.
+type mockGateway struct {
+	modules.Gateway
+	numBroadcasts int
+	mu            sync.RWMutex
+}
+
+func (g *mockGateway) Broadcast(string, interface{}) {
```
DavidVorick: needs docstring, especially since it's not a stub it actually does something

modules/consensus/accept.go
```diff
@@ -120,11 +120,12 @@ func (cs *ConsensusSet) addBlockToTree(b types.Block) (ce changeEntry, err error
...
-// on a fork that is heavier than the current fork. If the block is accepted,
-// it will be relayed to connected peers. This function should only be called
-// for new, untrusted blocks.
-func (cs *ConsensusSet) AcceptBlock(b types.Block) error {
+// managedAcceptBlock accepts a block but does not broadcast it to any peers.
+// See comment for AcceptBlock. Typically AcceptBlock should be used. This
+// method should only be used when there would otherwise be multiple
+// consecutive calls to AcceptBlock with each successive call accepting the
+// child block of the previous call.
+func (cs *ConsensusSet) managedAcceptBlock(b types.Block) error {
```
DavidVorick: the docstring here is a little weird, because `managedAcceptBlock` is not an exported function, and it gets called by `AcceptBlock`.  I would probably copy the `AcceptBlock` docstring and then remove the part about the broadcast, and the part about only being called on new blocks.

modules/consensus/accept_test.go
lukechampine: although, from the docstring it seems like the test really just checks whether `Broadcast` is called at all. That's much easier to test for.

modules/consensus/synchronize.go
```diff
@@ -90,8 +90,14 @@ func (cs *ConsensusSet) threadedReceiveBlocks(conn modules.PeerConn) error {
...
-		for _, block := range newBlocks {
-			acceptErr := cs.AcceptBlock(block)
+		for i, block := range newBlocks {
+			// Only broadcast the last block.
+			var acceptErr error
+			if !moreAvailable && i == len(newBlocks)-1 {
+				acceptErr = cs.AcceptBlock(block)
+			} else {
+				acceptErr = cs.managedAcceptBlock(block)
+			}
```
DavidVorick: so... `moreAvailable` depends on an external party. This logic here could actually be abused by an attacker to prevent you from broadcasting a block that you are given, because the attacker could say 'moreAvailable' and then throw an error or something.

Instead, I'm thinking you should do a `defer Broadcast` at the top of the function that will cause you to broadcast the highest block in your consensus set IFF there was at least one successful call to `managedAcceptBlock`.

So maybe something like:

```go
// A comment here explaining what is going on and why
var chainExtended *bool
defer func() {
    if chainExtended {
        recentBlock = someting // there's code in the consensus set to fetch the most recent block. Might need to open a db tx.
        cs.gateway.Broadcast() // would be nice to know that you are using the same broadcast code as AcceptBlock
    }
}()
```

That way you know you are going to broadcast 1 block only, and that you are only going to broadcast if you extend the chain, and you are definitely going to broadcast after extending the chain.

DavidVorick: defining a `broadcastBlock` helper function in the gateway makes sense to me, because to help with network flooding issues we're going to be sending around headers instead of blocks when broadcasts occur. Replacing the direct broadcast with the wrapped broadcast means it'll be a cleaner transition when the time comes.

modules/consensus/accept_test.go
```diff
@@ -701,3 +702,82 @@ func TestTaxHardfork(t *testing.T) {
...
+
+	// Test that Broadcast is not called for invalid blocks.
+	mg.mu.Lock()
+	mg.numBroadcasts = 0
+	mg.mu.Unlock()
+	err = cst.cs.AcceptBlock(types.Block{})
+	if err == nil {
+		t.Fatal("expected AcceptBlock to error on an invalid block")
+	}
+	// Sleep one second to wait for a possible call to g.Broadcast.
```
DavidVorick: *Sleep one nanosecond

DavidVorick: this is basically equivalent to calling `runtime.Gosched()`

modules/consensus/accept_test.go
```diff
@@ -701,3 +702,82 @@ func TestTaxHardfork(t *testing.T) {
...
+	mg.mu.RUnlock()
+	if numBroadcasts != 0 {
+		t.Error("AcceptBlock broadcasted an invalid block")
+	}
+
+	// Test that Broadcast is not called in managedAcceptBlock.
+	mg.mu.Lock()
+	mg.numBroadcasts = 0
+	mg.mu.Unlock()
+	b, _ = cst.miner.FindBlock()
```
DavidVorick: You were asking earlier what the testminer does. `FindBlock` is a function of the testminer, and is only called during testing.

modules/consensus/synchronize_test.go
```diff
@@ -225,3 +225,69 @@ func TestBlockHistory(t *testing.T) {
 		}
 	}
 }
+
+// TestSendBlocksBroadcasts tests that the SendBlocks RPC call only Broadcasts
+// one block, no matter how many blocks are sent. In the case 0 blocks are
+// sent, tests that Broadcast is never called.
```
DavidVorick: you should also test the case where the consensus set is not extended, as no blocks should be broadcast in that event.

modules/consensus/synchronize_test.go
```diff
@@ -225,3 +225,69 @@ func TestBlockHistory(t *testing.T) {
...
+
+	tests := []struct {
+		blocksToMine          int
+		expectedNumBroadcasts int
+	}{
+		{0, 0},
+		{1, 1},
+		{2, 1},
+		{MaxCatchUpBlocks, 1},
+		{2 * MaxCatchUpBlocks, 1},
```
DavidVorick: might want to extend the definition of `MaxCatchUpBlocks

DavidVorick: also might want to add 2*MCUB+1 in addition to the tests you've already got.

DavidVorick: also, shouldn't `go vet` be complaining about the way you are initializing the struct? You are using unnamed fields.

lukechampine: `go vet` only complains when you do that with an exported struct type.

VoidingWarranties:
Updated. A few things to check out:
* `synchronize.go` broadcasts the last block manually inside a defer, so as to prevent against attacks preventing block broadcast.
* `accept_test.go` uses channel selects to test that `Broadcast` is called in `AcceptBlock` for valid blocks and that Broadcast is never called in `managedAcceptBlock`.
* `accept_test.go` and `synchronize_test.go` use `10 * time.Millisecond` for `time.Sleep` and `time.After` instead of `1`.
* `MaxCatchUpBlocks` is now `3` during testing.

I think that addresses most of the concerns. The only concerns left unaddressed are @DavidVorick's concerns about adding a `broadcastBlock` helper function and `Broadcast` being called in a go routine.
* `broadcastBlock` helper function was not added because `Broadcast` is only being called in two places, `threadedReceiveBlocks` and `AcceptBlock`.
* `Broadcast` is still being called in a go routine. This should be a separate PR.

modules/consensus/accept.go
```diff
@@ -120,11 +120,18 @@ func (cs *ConsensusSet) addBlockToTree(b types.Block) (ce changeEntry, err error
...
-func (cs *ConsensusSet) AcceptBlock(b types.Block) error {
+// managedAcceptBlock will try to add a block to the consensus set. If the
+// block does not extend the longest currently known chain, an error is
+// returned but the block is still kept in memory. If the block extends a fork
+// such that the fork becomes the longest currently known chain, the consensus
+// set will reorganize itself to recognize the new longest fork. Accepted
+// blocks are not relayed.
+//
+// Typically AcceptBlock should be used so that the accepted block is relayed.
+// This method is typically only be used when there would otherwise be multiple
```
DavidVorick: typically typicaly

modules/consensus/accept.go
```diff
@@ -179,9 +186,22 @@ func (cs *ConsensusSet) AcceptBlock(b types.Block) error {
...
+// reorganize itself to recognize the new longest fork. If a block is accepted
+// without error, it will be relayed to all connected peers. This function
+// should only be called for new blocks.
+func (cs *ConsensusSet) AcceptBlock(b types.Block) error {
+	err := cs.managedAcceptBlock(b)
+	if err != nil {
+		return err
+	}
 	// Broadcast the new block to all peers.
 	go cs.gateway.Broadcast("RelayBlock", b)
```
DavidVorick: another argument for not calling `Broadcast` in a goroutine is that you want to make sure blocks are being broadcast in the right order, lest network confusion ensue.

DavidVorick: Though, no lock is held right here by the consensus set, so there's no guarantee that things are happening in the right order.

modules/consensus/accept_test.go
```diff
@@ -701,3 +701,66 @@ func TestTaxHardfork(t *testing.T) {
...
+	cst, err := blankConsensusSetTester("TestAcceptBlockBroadcasts")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer cst.Close()
+	mg := &mockGatewayDoesBroadcast{
+		Gateway:         cst.cs.gateway,
+		broadcastCalled: make(chan struct{}),
+	}
+	cst.cs.gateway = mg
```
DavidVorick: I feel like this isn't the best way to go about things, it's probably better to call `new` again.

modules/consensus/synchronize.go
```diff
@@ -10,11 +10,16 @@ import (
...
+var (
 	// MaxCatchUpBlocks is the maxiumum number of blocks that can be given to
 	// the consensus set in a single iteration during the initial blockchain
 	// download.
-	MaxCatchUpBlocks = 10
+	MaxCatchUpBlocks = func() types.BlockHeight {
+		if build.Release == "testing" {
+			return 3
+		}
+		return 10
```
DavidVorick: we prefer to have all units explicitly stated. if build.Release == "dev", "standard", and then default to `panic()`

modules/consensus/synchronize_test.go
```diff
@@ -225,3 +227,87 @@ func TestBlockHistory(t *testing.T) {
...
+	tests := []struct {
+		blocksToMine          int
+		expectedNumBroadcasts int
+	}{
+		{0, 0},
+		{1, 1},
+		{2, 1},
+		{int(MaxCatchUpBlocks), 1},
+		{2 * int(MaxCatchUpBlocks), 1},
+		{2*int(MaxCatchUpBlocks) + 1, 1},
```
DavidVorick: these all violate `go vet` rules - need named values

lukechampine: they don't! Try it yourself. `go vet` only complains about unkeyed fields on imported types.

VoidingWarranties: Unkeyed fields for slices of anonymous structs seems to be commonly used in other tests.

modules/consensus/synchronize.go
VoidingWarranties: Why so explicit?

modules/consensus/accept_test.go
VoidingWarranties: I don't understand, what do you mean?

modules/consensus/accept.go
VoidingWarranties: If blocks are broadcast out of order peers should still synchronize on the next block. But I'm not sure how blocks can be broadcast out of order now that there is only one broadcast per `threadedReceiveBlocks`.

modules/consensus/synchronize_test.go
DavidVorick: oh. I had assumed go vet didn't complain because it didn't run on test code.

modules/consensus/synchronize.go
DavidVorick: Being explicit prevents mistakes, and makes it easier to tweak parameters if you are doing debugging or specific testing. It's a lot easier to memorize/read/lookup if you know that the definitions are consistent throughout the file, rather than trying to remember if only testing or only dev is special for this or that constant.

modules/consensus/accept.go
DavidVorick: if you are receiving multiple blocks at once, and your calls to AcceptBlock are happening faster than your calls to Broadcast, you'll end up with multiple broadcasts waiting before any have finished. Since it's happening concurrently, you have no guarantee about which order the broadcasts are happening in.

modules/consensus/accept_test.go
DavidVorick: You create a full object, and then replace the gateway of the full object with the mocked object. That to me says that there are going to be problems somewhere, because there's a lot of stuff happening under the hood where objects and whatnot could be getting passed around. So, the non-mocked gateway could be leaking into the consensus set and into the other parts of the consensus set tester. For example, the transaction pool is using a full gateway, which has the potential to cause problems.

modules/consensus/synchronize.go
```diff
@@ -91,8 +107,13 @@ func (cs *ConsensusSet) threadedReceiveBlocks(conn modules.PeerConn) error {
 
 		// Integrate the blocks into the consensus set.
 		for _, block := range newBlocks {
-			acceptErr := cs.AcceptBlock(block)
-
+			// Call managedAcceptBlock instead of AcceptBlock so as not to broadcast
+			// every block.
+			acceptErr := cs.managedAcceptBlock(block)
+			// Only broadcast the last block if a block was accepted.
```
DavidVorick: this comment is a little confusing, because it makes it seem like you are doing the broadcasting immediately, instead of setting a flag that will trigger a broadcast in the defer statement.

DavidVorick:
Overall this looks pretty good. The testing seems sufficient, though I haven't checked that it's at 100% coverage yet (including error handling), and I would like to hold that as a requirement for all code moving forward.

It's just the nitpicks left at this point. I think the `go Broadcast` is something that should be discussed at another time in another pull request. Probably doesn't matter as much once we're doing header-flooding instead of block-flooding.

modules/consensus/accept_test.go
VoidingWarranties: Good point. Although in this test we only want to mock the gateway for the consensus set, and we don't want our mocked gateway to be used by transaction pool. If transaction pool were to call Broadcast we might mistake it as consensus calling Broadcast.

modules/consensus/synchronize.go
```diff
@@ -76,13 +81,24 @@ func (cs *ConsensusSet) threadedReceiveBlocks(conn modules.PeerConn) error {
 		return err
 	}
 
+	// Broadcast the last block accepted. This functionality is in a defer to
+	// ensure that a block is always broadcast if any blocks are accepted. This
+	// is to stop an attacker from preventing block broadcasts.
+	chainExtended := false
+	defer func() {
+		if chainExtended {
+			currentBlock := cs.CurrentBlock()
```
VoidingWarranties: Potential bug... If the last block sent by the `SendBlocks` RPC doesn't extend the longest chain, shouldn't we still broadcast the block?

modules/consensus/synchronize.go
```diff
@@ -91,8 +107,13 @@ func (cs *ConsensusSet) threadedReceiveBlocks(conn modules.PeerConn) error {
...
 		for _, block := range newBlocks {
-			acceptErr := cs.AcceptBlock(block)
-
+			// Call managedAcceptBlock instead of AcceptBlock so as not to broadcast
+			// every block.
+			acceptErr := cs.managedAcceptBlock(block)
+			// Only broadcast the last block if a block was accepted.
+			if acceptErr == nil {
+				chainExtended = true
+			}
```
VoidingWarranties: Should this be placed after the
``` go
// ErrNonExtendingBlock must be ignored until headers-first block
// sharing is implemented, block already in database should also be
// ignored.
if acceptErr == modules.ErrNonExtendingBlock || acceptErr == modules.ErrBlockKnown {
	acceptErr = nil
}
```
block so that a non-extending and known blocks are still broadcast?

modules/consensus/synchronize.go
```diff
@@ -91,8 +115,13 @@ func (cs *ConsensusSet) threadedReceiveBlocks(conn modules.PeerConn) error {
...
 		for _, block := range newBlocks {
-			acceptErr := cs.AcceptBlock(block)
-
+			// Call managedAcceptBlock instead of AcceptBlock so as not to broadcast
+			// every block.
+			acceptErr := cs.managedAcceptBlock(block)
+			// Set a flag to indicate that we should broadcast the last block received.
+			if acceptErr == nil {
+				chainExtended = true
+			}
```
VoidingWarranties: Should this be placed after the
``` go
// ErrNonExtendingBlock must be ignored until headers-first block
// sharing is implemented, block already in database should also be
// ignored.
if acceptErr == modules.ErrNonExtendingBlock || acceptErr == modules.ErrBlockKnown {
	acceptErr = nil
}
```
block so that a non-extending and known blocks are still broadcast?

modules/consensus/synchronize.go
```diff
@@ -76,13 +87,26 @@ func (cs *ConsensusSet) threadedReceiveBlocks(conn modules.PeerConn) error {
...
 
+	// Broadcast the last block accepted. This functionality is in a defer to
+	// ensure that a block is always broadcast if any blocks are accepted. This
+	// is to stop an attacker from preventing block broadcasts.
+	chainExtended := false
+	defer func() {
+		if chainExtended {
+			// The last block received will be the current block since
+			// managedAcceptBlock only returns nil if a block extends the longest chain.
+			currentBlock := cs.CurrentBlock()
```
VoidingWarranties: Potential bug... If the last block sent by the `SendBlocks` RPC doesn't extend the longest chain, shouldn't we still broadcast the block?

modules/consensus/accept.go
VoidingWarranties: Check your spellcheck haha

modules/consensus/accept_test.go
DavidVorick: You can have the mocked gateway be a fully functioning gateway by adding the real call to Broadcast after you do the counting.

VoidingWarranties: Done

modules/consensus/synchronize.go
DavidVorick: no, you only want to annoy your peers if you've found a longer chain that they might not know about yet. If you don't get any extending blocks, there's no reason to tell your peers because they already know about the longest chain (either you broadcast it to them when you got the longest block, or they asked you about it when they connected to you).

modules/consensus/synchronize.go
DavidVorick: also no

modules/consensus/synchronize.go
```diff
@@ -10,11 +10,22 @@ import (
...
 	// download.
-	MaxCatchUpBlocks = 10
+	MaxCatchUpBlocks = func() types.BlockHeight {
+		switch build.Release {
+		case "testing":
+			return 3
+		case "standard":
+			return 10
+		case "dev":
+			return 10
```
DavidVorick: alphabetic order please :P

modules/consensus/synchronize.go
```diff
@@ -10,11 +10,22 @@ import (
...
+		default:
+			panic("unrecognized build.Release")
+		}
```
DavidVorick: also, when it's just 3 options you probably don't need a switch statement

PR #978, opened 2016-02-09
Remove compatibility for v0.4.6

VoidingWarranties:
AcceptTransactionSet was broadcasting both the RPC RelayTransactionSet and the RPC RelayTransaction to maintain compatibility with v0.4.6. But because rpcIDs are 8 bytes, the two rpcIDs collide.

Also add clarifying comment regarding RPC identifiers to `RegisterRPC`.

### DIFF
--- a/modules/transactionpool/transactionpool.go
+++ b/modules/transactionpool/transactionpool.go
@@ -88,6 +88,8 @@ func New(cs modules.ConsensusSet, g modules.Gateway) (*TransactionPool, error) {
 		consensusChangeIndex: -1,
 	}
 	// Register RPCs
+	// TODO: rename RelayTransactionSet so that the conflicting RPC
+	// RelayTransaction calls v0.4.6 clients and earlier are ignored.
 	g.RegisterRPC("RelayTransactionSet", tp.relayTransactionSet)
 
 	// Subscribe the transaction pool to the consensus set.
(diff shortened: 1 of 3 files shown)


modules/transactionpool/transactionpool.go
```diff
@@ -88,6 +88,8 @@ func New(cs modules.ConsensusSet, g modules.Gateway) (*TransactionPool, error) {
 		consensusChangeIndex: -1,
 	}
 	// Register RPCs
+	// TODO: rename RelayTransactionSet so that the conflicting RPC
+	// RelayTransaction calls v0.4.6 clients and earlier are ignored.
 	g.RegisterRPC("RelayTransactionSet", tp.relayTransactionSet)
```
DavidVorick: So, before we release v0.5.2 we'll need to add a new RPC "RelayTx2" that operates the same as the current rpc. The v0.5.2 client will need to use "RelayTx2" for all nodes that are v0.5.2 or greater, and "RelayTra" for nodes that are an earlier version. v0.5.2 will need to accept "RelayTra" as well as "RelayTx2", meaning it'll get a lot of bad transactions.

Then, once a significant portion of the network has upgraded to v0.5.2 or higher, we can release a version that ignores all incoming "RelayTra" calls, protecting the nodes from undesired communication. At this point, compatibility will be maintained via bridges - people running v0.5.2 who are able to talk to both versions. Eventually, all compatibility with nodes earlier than v0.5.2 will be broken, so the rollout must be slow enough that people have time to upgrade.

PR #981, opened 2016-02-17
Versioned broadcast

VoidingWarranties:
`BroadcastMinVersion` and `BroadcastMaxVersion` broadcast only to peers that are of at least or at most the specified version, respectively. This is necessary for the upcoming header broadcast PR which will need to broadcast blocks to older peers and headers to newer peers. It is also useful for general backwards compatibility considerations. As an example use, RelayTransactionSet is now only broadcast to peers that are at least v0.4.7, so as to limit the number of conflicting calls to the colliding old RPC name RelayTransaction.

I'm not too sure about the names, as `BroadcastMinVersion` makes it sound like it's broadcasting some version number.

### DIFF
--- a/api/gateway.go
+++ b/api/gateway.go
@@ -9,15 +9,15 @@ import (
 )
 
 type GatewayInfo struct {
-	NetAddress modules.NetAddress   `json:"netaddress"`
-	Peers      []modules.NetAddress `json:"peers"`
+	NetAddress modules.NetAddress `json:"netaddress"`
+	Peers      []modules.Peer     `json:"peers"`
 }
 
 // gatewayHandler handles the API call asking for the gatway status.
 func (srv *Server) gatewayHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
 	peers := srv.gateway.Peers()
 	if peers == nil {
-		peers = make([]modules.NetAddress, 0)
+		peers = make([]modules.Peer, 0)
 	}
 	writeJSON(w, GatewayInfo{srv.gateway.Address(), peers})
 }
--- a/api/gateway.go
+++ b/api/gateway.go
@@ -16,6 +16,9 @@ type GatewayInfo struct {
 // gatewayHandler handles the API call asking for the gatway status.
 func (srv *Server) gatewayHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
 	peers := srv.gateway.Peers()
+	// nil slices are marshalled as 'null' in JSON, whereas 0-length slices are
+	// marshalled as '[]'. The latter is preferred, indicating that the value
+	// exists but contains no elements.
 	if peers == nil {
 		peers = make([]modules.Peer, 0)
 	}
--- a/modules/consensus/accept.go
+++ b/modules/consensus/accept.go
@@ -202,6 +202,6 @@ func (cs *ConsensusSet) AcceptBlock(b types.Block) error {
 		return err
 	}
 	// Broadcast the new block to all peers.
-	go cs.gateway.Broadcast("RelayBlock", b)
+	go cs.gateway.Broadcast("RelayBlock", b, nil)
 	return nil
 }
--- a/modules/consensus/synchronize.go
+++ b/modules/consensus/synchronize.go
@@ -96,7 +96,8 @@ func (cs *ConsensusSet) threadedReceiveBlocks(conn modules.PeerConn) error {
 			// The last block received will be the current block since
 			// managedAcceptBlock only returns nil if a block extends the longest chain.
 			currentBlock := cs.CurrentBlock()
-			go cs.gateway.Broadcast("RelayBlock", currentBlock)
+			// TODO: only broadcast blocks to peers v0.5.1 and below and headers to peers v0.5.2 and above.
+			go cs.gateway.Broadcast("RelayBlock", currentBlock, nil)
 		}
 	}()
 
--- a/modules/consensus/synchronize.go
+++ b/modules/consensus/synchronize.go
@@ -96,7 +96,6 @@ func (cs *ConsensusSet) threadedReceiveBlocks(conn modules.PeerConn) error {
 			// The last block received will be the current block since
 			// managedAcceptBlock only returns nil if a block extends the longest chain.
 			currentBlock := cs.CurrentBlock()
-			// TODO: only broadcast blocks to peers v0.5.1 and below and headers to peers v0.5.2 and above.
 			go cs.gateway.Broadcast("RelayBlock", currentBlock, nil)
 		}
 	}()
--- a/modules/gateway.go
+++ b/modules/gateway.go
@@ -29,52 +29,54 @@ var (
 	}
 )
 
-// A PeerConn is the connection type used when communicating with peers during
-// an RPC. For now it is identical to a net.Conn.
-type PeerConn interface {
-	net.Conn
-}
+type (
+	// A PeerConn is the connection type used when communicating with peers during
+	// an RPC. For now it is identical to a net.Conn.
+	PeerConn interface {
+		net.Conn
+	}
 
-// RPCFunc is the type signature of functions that handle RPCs. It is used for
-// both the caller and the callee. RPCFuncs may perform locking. RPCFuncs may
-// close the connection early, and it is recommended that they do so to avoid
-// keeping the connection open after all necessary I/O has been performed.
-type RPCFunc func(PeerConn) error
+	// RPCFunc is the type signature of functions that handle RPCs. It is used for
+	// both the caller and the callee. RPCFuncs may perform locking. RPCFuncs may
+	// close the connection early, and it is recommended that they do so to avoid
+	// keeping the connection open after all necessary I/O has been performed.
+	RPCFunc func(PeerConn) error
 
-// A Gateway facilitates the interactions between the local node and remote
-// nodes (peers). It relays incoming blocks and transactions to local modules,
-// and broadcasts outgoing blocks and transactions to peers. In a broad sense,
-// it is responsible for ensuring that the local consensus set is consistent
-// with the "network" consensus set.
-type Gateway interface {
-	// Connect establishes a persistent connection to a peer.
-	Connect(NetAddress) error
+	// A Gateway facilitates the interactions between the local node and remote
+	// nodes (peers). It relays incoming blocks and transactions to local modules,
+	// and broadcasts outgoing blocks and transactions to peers. In a broad sense,
+	// it is responsible for ensuring that the local consensus set is consistent
+	// with the "network" consensus set.
+	Gateway interface {
+		// Connect establishes a persistent connection to a peer.
+		Connect(NetAddress) error
 
-	// Disconnect terminates a connection to a peer.
-	Disconnect(NetAddress) error
+		// Disconnect terminates a connection to a peer.
+		Disconnect(NetAddress) error
 
-	// Address returns the Gateway's address.
-	Address() NetAddress
+		// Address returns the Gateway's address.
+		Address() NetAddress
 
-	// Peers returns the addresses that the Gateway is currently connected to.
-	Peers() []NetAddress
+		// Peers returns the addresses that the Gateway is currently connected to.
+		Peers() []NetAddress
 
-	// RegisterRPC registers a function to handle incoming connections that
-	// supply the given RPC ID.
-	RegisterRPC(string, RPCFunc)
+		// RegisterRPC registers a function to handle incoming connections that
+		// supply the given RPC ID.
+		RegisterRPC(string, RPCFunc)
 
-	// RegisterConnectCall registers an RPC name and function to be called
-	// upon connecting to a peer.
-	RegisterConnectCall(string, RPCFunc)
+		// RegisterConnectCall registers an RPC name and function to be called
+		// upon connecting to a peer.
+		RegisterConnectCall(string, RPCFunc)
 
-	// RPC calls an RPC on the given address. RPC cannot be called on an
-	// address that the Gateway is not connected to.
-	RPC(NetAddress, string, RPCFunc) error
+		// RPC calls an RPC on the given address. RPC cannot be called on an
+		// address that the Gateway is not connected to.
+		RPC(NetAddress, string, RPCFunc) error
 
-	// Broadcast transmits obj, prefaced by the RPC name, to all of the
-	// Gateway's connected peers in parallel.
-	Broadcast(name string, obj interface{})
+		// Broadcast transmits obj, prefaced by the RPC name, to all of the
+		// Gateway's connected peers in parallel.
+		Broadcast(name string, obj interface{})
 
-	// Close safely stops the Gateway's listener process.
-	Close() error
-}
+		// Close safely stops the Gateway's listener process.
+		Close() error
+	}
+)
--- a/modules/gateway.go
+++ b/modules/gateway.go
@@ -30,6 +30,12 @@ var (
 )
 
 type (
+	// Peer contains all the info necessary to Broadcast to a peer.
+	Peer struct {
+		NetAddress NetAddress `json:"netaddress"`
+		Version    string     `json:"version"`
+	}
+
 	// A PeerConn is the connection type used when communicating with peers during
 	// an RPC. For now it is identical to a net.Conn.
 	PeerConn interface {
@@ -58,7 +64,7 @@ type (
 		Address() NetAddress
 
 		// Peers returns the addresses that the Gateway is currently connected to.
-		Peers() []NetAddress
+		Peers() []Peer
 
 		// RegisterRPC registers a function to handle incoming connections that
 		// supply the given RPC ID.
--- a/modules/gateway.go
+++ b/modules/gateway.go
@@ -79,8 +79,8 @@ type (
 		RPC(NetAddress, string, RPCFunc) error
 
 		// Broadcast transmits obj, prefaced by the RPC name, to all of the
-		// Gateway's connected peers in parallel.
-		Broadcast(name string, obj interface{})
+		// given peers in parallel.
+		Broadcast(name string, obj interface{}, peers []Peer)
 
 		// Close safely stops the Gateway's listener process.
 		Close() error
--- a/modules/gateway/peers.go
+++ b/modules/gateway/peers.go
@@ -26,6 +26,16 @@ const (
 	minNodeListLen = 100
 )
 
+var errPeerRejectedConn = errors.New("peer rejected connection")
+
+// insufficientVersionError indicates a peer's version is insufficient.
+type insufficientVersionError string
+
+// Error implements the error interface for insufficientVersionError.
+func (s insufficientVersionError) Error() string {
+	return "unacceptable version: " + string(s)
+}
+
 type peer struct {
 	addr    modules.NetAddress
 	sess    muxado.Session
@@ -211,7 +221,7 @@ func (g *Gateway) Connect(addr modules.NetAddress) error {
 	}
 	// decide whether to accept this version
 	if remoteVersion == "reject" {
-		return errors.New("peer rejected connection")
+		return errPeerRejectedConn
 	}
 	// Check that version is acceptable.
 	//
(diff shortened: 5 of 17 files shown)


modules/gateway/rpc.go
```diff
@@ -155,3 +154,50 @@ func (g *Gateway) Broadcast(name string, obj interface{}) {
...
+
+// BroadcastMinVersion calls an RPC on all of the peers in the Gateway's peer
+// list that are at least the specified version. The calls are run in parallel.
+// Broadcasts are restricted to "one-way" RPCs, which simply write an object
+// and disconnect. This is why Broadcast takes an interface{} instead of an
+// RPCFunc. BroadcastMinVersion can be used to ensure new RPC calls are not
+// made to old nodes that wouldn't understand them.
+func (g *Gateway) BroadcastMinVersion(name string, obj interface{}, minVersion string) {
+	id := g.mu.RLock()
+	defer g.mu.RUnlock(id)
```
lukechampine: won't this hold the lock during the broadcast?

lukechampine:
I'm not sure if this has any present use, but it might be cool to add a `Version` method to `PeerConn`. Then you could write version-branching code inside RPCs.

lukechampine:
Adding two new methods is the easiest way to implement this functionality, but I'm not sure it's the best. Here's one alternative:
```go
// in modules:
type Peer struct {
    NetAddress NetAddress
    Version    string
}

// in gateway:
func (g *Gateway) Broadcast(name string, obj interface{}, peers []modules.Peer) {
    if peers == nil {
        peers = g.Peers()
    }
    //...broadcast logic
}

// in transaction pool:
var peers []modules.Peer
for _, p := range gateway.Peers() {
    if build.VersionCmp(p.Version, "0.4.7") >= 0 {
        peers = append(peers, p)
    }
}
go tp.gateway.Broadcast("RelayTransactionSet", ts, peers)
```
Not saying this is the best solution -- it complicates things by introducing a new type -- but it does allow you to write arbitrarily complex `Broadcast`s *outside* the Gateway rather than inside. Generally speaking, I think I would prefer one extensible `Broadcast` to multiple specific `Broadcast`s. Thoughts?

modules/gateway/rpc.go
VoidingWarranties: Correct. This isn't much of a problem because `broadcastTo` does not need any locking of it's own. Are you concerned that the lock is held longer than needed, i.e. during the actual network writes? In that case yes, the lock is held longer than needed.

On the other hand, holding the lock for the entire broadcast ensures that broadcasts are guaranteed to execute in the order they are called.

But since this does not align with the behavior of the unversioned `Broadcast` method, I'll change it.

VoidingWarranties:
>I'm not sure if this has any present use, but it might be cool to add a `Version` method to `PeerConn`. Then you could write version-branching code inside RPCs.

I'd be worried that this would overly complicate RPCs. With that said, it would allow us to maintain backwards compatibility without adding additional RPCs or changing their names to include a version (such as `RelayTx2` we talked about offline last week). Since backwards compatibility is hard, I like this idea.

VoidingWarranties:
>Adding two new methods is the easiest way to implement this functionality, but I'm not sure it's the best. Here's one alternative:
...

:+1:

modules/gateway/rpc.go
DavidVorick: It's very, very important that module locks not be held during network calls. That's a severe DoS vector, and in general will have a massive impact on performance and usability.

DavidVorick:
I think it makes sense. Especially if your logic is more complex, like 'send it to 045 and 047 but not 046' or something like that. Or if you've got a blacklist that the consensus set is maintaining, or some other reason. I'm in favor of nixing the range call in favor of letting the module figure out which peers deserve a message.

lukechampine:
`Broadcast` is interesting in that it could easily be implemented outside the gateway, since it only calls exported functions. I suppose it's a useful helper though, and the implementation may become more complex, or we may wish to add metrics to `Broadcast`

VoidingWarranties:
Done. As a side effect of returning `[]modules.Peer` from `Gateway.Peers()` instead of `[]modules.NetworkAddress`, siac was printing out peers in as the raw `modules.Peer` structs (`{address, version}`), so I updated the command to print the address and version in a table like so
```
$ siac gateway list
1 active peers:
Address              Version
  176.9.59.110:9981  0.5.1
```

modules/gateway/peers.go
```diff
@@ -163,7 +164,7 @@ func (g *Gateway) acceptConn(conn net.Conn) {
 		g.log.Printf("INFO: disconnected from %v to make room for %v", kick, addr)
 	}
 	// add the peer
-	g.addPeer(&peer{addr: addr, sess: muxado.Server(conn), inbound: true})
+	g.addPeer(&peer{addr: addr, sess: muxado.Server(conn), inbound: true, version: remoteVersion})
```
lukechampine: this object is now large enough that I would split it across multiple lines

modules/gateway/peers.go
```diff
@@ -210,7 +212,7 @@ func (g *Gateway) Connect(addr modules.NetAddress) error {
 	g.log.Println("INFO: connected to new peer", addr)
 
 	id = g.mu.Lock()
-	g.addPeer(&peer{addr: addr, sess: muxado.Client(conn), inbound: false})
+	g.addPeer(&peer{addr: addr, sess: muxado.Client(conn), inbound: false, version: remoteVersion})
```
lukechampine: ditto

api/gateway.go
```diff
@@ -9,15 +9,15 @@ import (
...
+	NetAddress modules.NetAddress `json:"netaddress"`
+	Peers      []modules.Peer     `json:"peers"`
 }
 
 // gatewayHandler handles the API call asking for the gatway status.
 func (srv *Server) gatewayHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
 	peers := srv.gateway.Peers()
 	if peers == nil {
-		peers = make([]modules.NetAddress, 0)
+		peers = make([]modules.Peer, 0)
```
lukechampine: a comment about why we need a non-nil slice here would be good

api/gateway.go
```diff
@@ -9,15 +9,15 @@ import (
 )
 
 type GatewayInfo struct {
-	NetAddress modules.NetAddress   `json:"netaddress"`
-	Peers      []modules.NetAddress `json:"peers"`
+	NetAddress modules.NetAddress `json:"netaddress"`
+	Peers      []modules.Peer     `json:"peers"`
```
lukechampine: since this struct has changed, the api docs need to be updated.

VoidingWarranties:
Oops thought you were talking about something different, I'll fix this up tomorrow when I'm not dead tired.

modules/gateway.go
```diff
@@ -29,52 +29,60 @@ var (
...
-// an RPC. For now it is identical to a net.Conn.
-type PeerConn interface {
-	net.Conn
-}
+type (
+	// Peer contains all the info necessary to Broadcast to a peer.
+	Peer struct {
+		NetAddress NetAddress
+		Version    string
+	}
```
lukechampine: need json tags

api/gateway.go
```diff
@@ -9,15 +9,18 @@ import (
...
 }
 
 // gatewayHandler handles the API call asking for the gatway status.
 func (srv *Server) gatewayHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
 	peers := srv.gateway.Peers()
+	// Although go clients can unmarshal a nil slice to an empty slice, we cannot
+	// depend on javascript clients having the same behavior. To make it easier
+	// for javascript clients, we make a 0 length slice.
```
VoidingWarranties: I wasn't 100% sure about this comment.

api/gateway.go
VoidingWarranties: Done

api/gateway.go
lukechampine: more accurate would be:
```go
// nil slices are marshalled as 'null' in JSON, whereas 0-length slices are marshalled as '[]'.
// The latter is preferred, indicating that the value exists but contains no elements.
```

modules/consensus/accept.go
```diff
@@ -202,6 +202,6 @@ func (cs *ConsensusSet) AcceptBlock(b types.Block) error {
 		return err
 	}
 	// Broadcast the new block to all peers.
-	go cs.gateway.Broadcast("RelayBlock", b)
+	go cs.gateway.Broadcast("RelayBlock", b, nil)
```
DavidVorick: would like to see a check in the gateway code that throws an error or panic if a developer tries to use more than 8 characters. It's easy to forget that "RelayBlock" and "RelayBlock2" are actually the same call.

modules/consensus/synchronize.go
```diff
@@ -96,7 +96,8 @@ func (cs *ConsensusSet) threadedReceiveBlocks(conn modules.PeerConn) error {
 			// The last block received will be the current block since
 			// managedAcceptBlock only returns nil if a block extends the longest chain.
 			currentBlock := cs.CurrentBlock()
-			go cs.gateway.Broadcast("RelayBlock", currentBlock)
+			// TODO: only broadcast blocks to peers v0.5.1 and below and headers to peers v0.5.2 and above.
```
DavidVorick: please resolve TODOs before making pull requests

DavidVorick: oh, headers is not in this PR. hmmm.

I'm really trying to avoid adding todos of any form to the master branch,we are at the point where we should be resolving them.

VoidingWarranties: The TODO is already gone in the `header-broadcast` branch. That PR is ready to go as soon as this one is merged.

modules/gateway/peers.go
```diff
@@ -198,10 +204,11 @@ func (g *Gateway) Connect(addr modules.NetAddress) error {
...
 	if err := encoding.ReadObject(conn, &remoteVersion, maxAddrLength); err != nil {
 		return err
-	} else if remoteVersion == "reject" {
-		return errors.New("peer rejected connection")
 	}
 	// decide whether to accept this version
+	if remoteVersion == "reject" {
+		return errors.New("peer rejected connection")
+	}
 	if build.VersionCmp(remoteVersion, "0.3.3") < 0 {
```
DavidVorick: Can we bump this to 0.4.0?

DavidVorick: 0333 nodes are actually hardforked off the network at this point, we need to phase out their ability to obtain peers so that they realize they need to upgrade.

modules/gateway/rpc_test.go
```diff
@@ -181,22 +185,84 @@ func TestBroadcast(t *testing.T) {
...
 		return nil
 	})
 
-	g1.Broadcast("Recv", "foo")
-	<-doneChan
-	<-doneChan
+	// Test that calling broadcast with nil peers broadcasts to all peers.
+	g1.Broadcast("Recv", "foo", nil)
+	<-g2DoneChan
+	<-g3DoneChan
```
DavidVorick: we should have timeouts here

modules/gateway/rpc_test.go
```diff
@@ -181,22 +185,84 @@ func TestBroadcast(t *testing.T) {
...
 	if g2Payload != "foo" || g3Payload != "foo" {
 		t.Fatal("broadcast failed:", g2Payload, g3Payload)
 	}
+
+	// Test that broadcasting to all peers in g1.Peers() broadcasts to all peers.
+	peers := g1.Peers()
+	g1.Broadcast("Recv", "bar", peers)
+	<-g2DoneChan
+	<-g3DoneChan
```
DavidVorick: ditto

modules/gateway/rpc_test.go
```diff
@@ -181,22 +185,84 @@ func TestBroadcast(t *testing.T) {
...
+	// Test that broadcasting to only g2 does not broadcast to g3.
+	peers = make([]modules.Peer, 0)
+	for _, p := range g1.Peers() {
+		if p.NetAddress == g2.Address() {
+			peers = append(peers, p)
+			break
+		}
+	}
+	g1.Broadcast("Recv", "baz", peers)
+	<-g2DoneChan
```
DavidVorick: timeouts

modules/gateway/rpc_test.go
```diff
@@ -181,22 +185,84 @@ func TestBroadcast(t *testing.T) {
...
+	// Test that broadcasting to only g3 does not broadcast to g2.
+	peers = make([]modules.Peer, 0)
+	for _, p := range g1.Peers() {
+		if p.NetAddress == g3.Address() {
+			peers = append(peers, p)
+			break
+		}
+	}
+	g1.Broadcast("Recv", "qux", peers)
+	<-g3DoneChan
```
DavidVorick: timeouts

modules/consensus/accept.go
VoidingWarranties: I agree. It will also encourage RPC names like "RelayTxs" instead of "RelayTransactionSet"

modules/consensus/synchronize.go
DavidVorick: you can probably just remove this line. It's pretty inherent to your next PR

VoidingWarranties:
Should be all set.

modules/gateway/rpc_test.go
```diff
@@ -181,22 +185,108 @@ func TestBroadcast(t *testing.T) {
...
 
-	g1.Broadcast("Recv", "foo")
-	<-doneChan
-	<-doneChan
+	// Test that calling broadcast with nil peers broadcasts to all peers.
+	g1.Broadcast("Recv", "foo", nil)
+	go func() {
+		<-g2DoneChan
+		<-g3DoneChan
+		bothDoneChan <- struct{}{}
```
DavidVorick: good enough I guess

modules/gateway/peers.go
```diff
@@ -198,11 +204,14 @@ func (g *Gateway) Connect(addr modules.NetAddress) error {
 	var remoteVersion string
 	if err := encoding.ReadObject(conn, &remoteVersion, maxAddrLength); err != nil {
 		return err
-	} else if remoteVersion == "reject" {
-		return errors.New("peer rejected connection")
 	}
 	// decide whether to accept this version
-	if build.VersionCmp(remoteVersion, "0.3.3") < 0 {
+	if remoteVersion == "reject" {
+		return errors.New("peer rejected connection")
```
DavidVorick: code is not covered by testing

modules/gateway/peers.go
```diff
@@ -198,11 +204,14 @@ func (g *Gateway) Connect(addr modules.NetAddress) error {
...
+	}
+	// Reject peers < v0.4.0 as the previous version is v0.3.3 which is
+	// pre-hardfork.
+	if build.VersionCmp(remoteVersion, "0.4.0") < 0 {
 		conn.Close()
 		return errors.New("unacceptable version: " + remoteVersion)
```
DavidVorick: though it's not in the diff, I feel like it's close enough to code being changed that tests should be written to cover this.

modules/transactionpool/accept.go
```diff
@@ -286,7 +287,16 @@ func (tp *TransactionPool) AcceptTransactionSet(ts []types.Transaction) error {
...
 	// Notify subscribers and broadcast the transaction set.
-	go tp.gateway.Broadcast("RelayTransactionSet", ts)
+	// NOTE: The transaction set is only broadcast to v0.4.7 peers and above.
+	// v0.4.7-v0.5.1 broadcasted both transaction sets and individual transactions
+	// and those versions act as a bridge between v0.5.2+ and older versions.
+	var v047AndAbove []modules.Peer
+	for _, p := range tp.gateway.Peers() {
+		if build.VersionCmp(p.Version, "0.4.7") >= 0 {
+			v047AndAbove = append(v047AndAbove, p)
+		}
```
DavidVorick: code is not covered by testing

siac/gatewaycmd.go
```diff
@@ -108,7 +110,10 @@ func gatewaylistcmd() {
 		return
 	}
 	fmt.Println(len(info.Peers), "active peers:")
+	w := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
+	fmt.Fprintln(w, "Address\tVersion")
 	for _, peer := range info.Peers {
-		fmt.Println("\t", peer)
+		fmt.Fprintf(w, "  %v\t%v\n", peer.NetAddress, peer.Version)
 	}
+	w.Flush()
```
DavidVorick: code is not covered by testing

VoidingWarranties: Siac testing will take a lot of work and I don't want it to be a blocker for this PR. I have a branch for siac testing I've been working on for a bit. I'll finish it up after the next header-broadcast PR.

modules/transactionpool/accept.go
VoidingWarranties: Done

VoidingWarranties:
All testing except siac testing should be complete. The only commits left for review are commits 10bb440 and onward (the commits that add testing). I rebased so it's not clear that those are the only commits different from last time.

modules/gateway/peers.go
```diff
@@ -26,10 +24,37 @@ const (
 	minNodeListLen = 100
 )
 
+var (
+	// The gateway will sleep this long between incoming connections.
+	acceptInterval = func() time.Duration {
```
DavidVorick: @lukechampine can you explain this comment more? I'm not sure what it means.

modules/gateway/peers.go
```diff
@@ -121,10 +146,14 @@ func (g *Gateway) acceptConn(conn net.Conn) {
 		return
 	}
 
-	// check that version is acceptable
+	// Check that version is acceptable.
+	//
+	// Reject peers < v0.4.0 as the previous version is v0.3.3 which is
```
DavidVorick: *previous version is v0.3.3.3

modules/gateway/peers_test.go
```diff
@@ -183,6 +183,231 @@ func TestConnect(t *testing.T) {
...
+			remoteVersion:       "世界",
+			versionResponseWant: "reject",
+			msg:                 "acceptConn shouldn't accept a remote peer whose version is utf8 giberish",
+		},
+		// Test that acceptConn fails when the remote peer's version is < 0.4.0 (0).
+		{
+			remoteVersion:       "0",
+			versionResponseWant: "reject",
+			msg:                 "acceptConn shouldn't accept a remote peer whose version is 0",
+		},
```
DavidVorick: probably good to have 00 and 0.0.0 and 00.00.00, I doubt any errors will turn up but often spamming 0's or 9's will get you something.

DavidVorick:
branch is out of date with master, otherwise looks pretty good. Two non-blocking nits

modules/gateway/peers.go
lukechampine: The gateway has a tight loop where it accepts an incoming connection and then spawns a goroutine to handle it. To prevent the loop from running too quickly (an attack vector), we sleep for `acceptInterval` before accepting another connection.

DavidVorick: that actually could be a separate attack vector. We aren't operating at the same sensitivity as Bitcoin, but for Bitcoin a 3 second delay for miners is a really big deal, it puts them 0.5% up in orphan rate compared to someone who is skipping the 3 second delay.

This is fine for now, but overall we should definitely have a different method for handing DoS in the gateway.

lukechampine: the delay only applies when you are trying to connect to a peer, not when sending an RPC. So it shouldn't slow down block/txn propagation.

DavidVorick: does not apply to incoming rpcs either? if not, then there's no issue. Slow peer connection times isn't a concern.

DavidVorick:
(siac aside, code coverage looks good)

modules/gateway/peers.go
VoidingWarranties: It shouldn't apply to incoming RPCs. Only calls to `Connect`.

modules/gateway/peers_test.go
VoidingWarranties: Done

VoidingWarranties:
Done (added more test cases for `TestAcceptConnRejects` and `TestConnectRejects` and rebased).

DavidVorick:
ok, needs rebase. Then we're good.

PR #987, opened 2016-02-21
reduce frequency of miner saves

lukechampine:


modules/miner/blockmanager.go
```diff
@@ -142,14 +142,10 @@ func (m *Miner) managedSubmitBlock(b types.Block) error {
...
 	if err != nil {
-		err2 := m.save()
-		if err2 != nil {
-			err = errors.New(err.Error() + " -and- " + err2.Error())
-		}
 		return err
 	}
 	m.persist.Address = uc.UnlockHash()
-	return m.save()
+	return nil
```
DavidVorick: I was wrong, we need the save here because we remember which blocks have been submitted, so that we can persistently calculate the miner's orphan rate. Derp.

DavidVorick:
I do think that we should have some sort of guaranteed periodic save. Otherwise a user could build up a backlog of 1000+ consensus changes that the miner hasn't processed, which would probably cause several seconds of delay the next time they start up (though, it'd be a one time thing until they built up a new backlog).

It's not that big of a deal, especially if they are actually using the miner and finding blocks, but...

hmm.

DavidVorick:
oh, it saves on close. I guess that's good enough.

PR #989, opened 2016-02-22
Redesign the Host

DavidVorick:
This PR is the worst kind of PR. It mixes a bunch of changes together. When I started writing the code, I really wasn't sure where I was going with it. I had in my head though that I should be creating "Host 1.0", which included all of the bells and whistles which would be required to make something that would have all the security features that I keep saying are 'on the way'. I wanted to not be saying that anymore, I wanted to say 'implemented'.

So, not all of those features are done. The bulk of this PR is just switching up the storage model of the host. The new storage model uses multiple storage folders, divides file contracts into sectors, and tolerates disk failures.

I went ahead and redesigned the host interface. You will notice that the actual host no longer satisfies the new host interface, because I added a bunch of functions that haven't been implemented yet. I had to extend the transaction pool to provide fee estimation, but that fee estimation is not used anywhere yet.

Basically, this is still a giant work in progress. You can skim the api/* diffs, and the modules/host/* diffs. Be sure to check out modules/host.go

All of the rpc code is gone from the host at this point. It's also out-of-scope for this PR. This feature branch should not be merged until both the host and the renter are using the new RPCs.


lukechampine:
are we not attempting to get the build passing until the feature branch is ready for merge?

DavidVorick:
I am not concerned with getting the build passing at the moment, because the host has changed pretty drastically and it'll require the renter changing as well to get a lot of the tests working again (especially the api integration tests).

build/critical.go
```diff
@@ -9,7 +9,9 @@ import (
 // which case panic will be called instead.
 func Critical(v ...interface{}) {
 	s := fmt.Sprintln(v...)
-	os.Stderr.WriteString(s)
+	if Release != "testing" {
+		os.Stderr.WriteString(s)
+	}
```
lukechampine: I think this could be:
```go
if DEBUG {
    panic(s)
} else {
    os.Stderr.WriteString(s)
}
```
since `testing` (essentially) implies `DEBUG`, but not the other way around

modules/host.go
```diff
@@ -21,39 +19,43 @@ const (
...
-	// A DownloadRequest is used to retrieve a particular segment of a file from a
-	// host.
-	DownloadRequest struct {
-		Offset uint64
-		Length uint64
+	// BandwidthLimits set limits on the volume and speed of the uploading and
+	// downloading of the host. The limits have no bearings on the other
+	// modules. The data limits are in bytes per month, and the speed limits
+	// are in bytes per second.
+	BandwidthLimits struct {
```
lukechampine: maybe `HostBandwidthLimits`?
Why do these need to be exported anyway?

modules/host.go
```diff
@@ -83,49 +93,113 @@ type (
...
+
+		// RemoveStorageFolder will remove a storage folder from the host. All
+		// storage on the folder will be moved to other storage folders,
+		// meaning that no data will be lost. If the host is unable to save
+		// data, an error will be returned and the operation will be stopped.
+		RemoveStorageFolder(index int) error
+
+		// ResetStorageFolderHealth will reset the health statistics on a
+		// storage folder.
+		ResetStorageFolderHealth(index int) error
```
lukechampine: what is this used for?

modules/host.go
```diff
@@ -83,49 +93,113 @@ type (
...
-		// host will begin accepting contracts.
-		Announce() error
+		// Announce submits a host announcement to the blockchain.  After
+		// announcing, the host will begin accepting contracts.
+		Announce(NetAddress) error
+
+		// ConfigureSmartPrices configures whether the host will automatically
+		// adjust prices based on the network environment. The algorithms aim
+		// to be profit maximizing.
+		ConfigureSmartPrices(contractPricing bool, downloadPricing bool, storagePricing bool, uploadPricing bool) error
```
lukechampine: can save type declarations here: `contractPricing, downloadPricing, storagePricing, uploadPricing bool`

modules/host.go
```diff
@@ -83,49 +93,113 @@ type (
...
+		// FileContracts returns a list of file contracts that the host
+		// currently has open, along with the volume of data tracked by each
+		// file contract.
+		FileContracts() ([]types.FileContractID, []uint64)
 
-		// DeleteContract deletes a file contract. The revenue and collateral
-		// on the file contract will be lost, and the data will be removed.
-		DeleteContract(types.FileContractID) error
+		// Financials returns the financial statistics of the host.
+		Financials() HostFinancials
```
lukechampine: maybe rename to match `RPCMetrics`, e.g. `RevenueMetrics`

modules/host.go
```diff
@@ -83,49 +93,113 @@ type (
...
-		// representing current collateral at risk, and one representing total
-		// collateral lost.
-		Revenue() (unresolved, resolved, lost types.Currency)
-
 		// RPCMetrics returns information on the types of RPC calls that have
 		// been made to the host.
 		RPCMetrics() HostRPCMetrics
 
+		// SetBandwidthLimits does exactly that.
+		SetBandwidthLimits(altruisticLimits, pricedLimits BandwidthLimits)
```
lukechampine: seems more complex than the docstring implies. :P
what are "altruistic limits" and "pricedLimits"?

modules/host.go
```diff
@@ -134,23 +208,50 @@ type (
...
+		StorageFolderManager
 	}
 )
 
+// BandwidthPriceToConsensus converts a human bandwidth price, having the unit
+// 'Siacoins per Terabyte', to a consensus storage price, having the unit
+// 'Hastings per Byte'.
+func BandwidthPriceToConsensus(siacoinsTB uint64) (hastingsByte types.Currency) {
+	hastingsTB := types.NewCurrency64(siacoinsTB).Mul(types.SiacoinPrecision)
+	return hastingsTB.Div(types.NewCurrency64(1e12))
```
lukechampine: `Mul64` and `Div64` would be nice to have in a lot of code like this.
```go
return types.SiacoinPrecision.Mul64(siacoinsTB).Div64(1e12)
```

modules/host/announce_test.go
```diff
@@ -15,6 +14,7 @@ func TestAnnouncement(t *testing.T) {
 	if testing.Short() {
 		t.SkipNow()
 	}
+	t.Parallel()
```
lukechampine: I would have preferred to add these in a separate PR, but whatevs

modules/host/dependencies.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+// These interfaces define the Host's dependencies. Mocking implementation
+// complexity can be reduced by defining each dependency as the minimium
+// possible subset of the real dependency.
+type (
+	// dependencies defines all of the dependencies of the Host.
+	dependencies interface {
+		// Listen gives the host the ability to receive incoming connections.
+		Listen(string, string) (net.Listener, error)
```
lukechampine: I think these can all be lowercased

modules/host/host.go
```diff
@@ -14,43 +14,197 @@ import (
...
+	}
+
+	// defaultContractPrice defines the default price of creating a contract
+	// with the host. The default is set to 50 siacoins, which means that the
+	// opening file contract can have 5 siacoins put towards it, the file
+	// contract revision can have 15 siacoins put towards it, and the storage
+	// proof can have 15 siacoins put towards it, with 5 left over for the
+	// host, a sort of compensation for keeping and backing up the storage
+	// obligation data.
+	defaultContractPrice = types.NewCurrency64(50).Mul(types.SiacoinPrecision) // 50 siacoins
```
lukechampine: I don't understand what `ContractPrice` is used for. It doesn't appear to be used in this PR.

modules/host/host.go
```diff
@@ -14,43 +14,197 @@ import (
...
+			return 1 << 33 // 8 GiB
 		}
 		if build.Release == "standard" {
-			return 144
+			return 1 << 47 // 128 TiB
+		}
+		if build.Release == "testing" {
+			return 1 << 20 // 1 MiB
 		}
+		panic("unrecognized release constant in host - maximum storage folder size")
```
lukechampine: we could probably eliminate these checks with a simple `init` function in the build package:
```go
func init() {
    switch Release {
    case "dev", "standard", "testing":
    default:
        panic("unrecognized Release constant: " + Release)
    }
}
```

modules/host/host.go
```diff
@@ -75,35 +256,44 @@ type Host struct {
...
-	obligationsByID map[types.FileContractID]*contractObligation
+	sectorSalt crypto.Hash
+
+	// Storage Obligation Management - different from file management in that
+	// the storage obligation management is the new way of handling storage
+	// obligations. Is a replacement for the contract obligation logic, but the
+	// old logic is being kept for compatibility purposes.
+	//
+	// Storage is broken up into sectors. The sectors are distributed randomly
+	// across a set of storage folders, using rendevouz hashing to choose which
```
lukechampine: rendezvous hashing -- outdated?

modules/host/host.go
```diff
@@ -120,8 +310,74 @@ type Host struct {
...
+}
+
+// initDB will check that the database has been initialized and if not, will
+// initialize the database.
+func (h *Host) initDB() error {
+	return h.db.Update(func(tx *bolt.Tx) error {
+		// Return nil if the database is already initialized. The database can
+		// be safely assumed to be initialized if the storage obligation bucket
+		// exists.
+		bso := tx.Bucket(bucketStorageObligations)
```
lukechampine: isn't this what `CreateBucketIfNotExists` is for?

modules/host/host.go
```diff
@@ -135,45 +391,76 @@ func New(cs modules.ConsensusSet, tpool modules.TransactionPool, wallet modules.
...
 		return nil, err
 	}
 
 	return h, nil
 }
 
+// New returns an initialized Host.
+func New(cs modules.ConsensusSet, tpool modules.TransactionPool, wallet modules.Wallet, address string, persistDir string) (*Host, error) {
+	return newHost(productionDependencies{}, cs, tpool, wallet, address, persistDir)
+}
```
lukechampine: :+1:

modules/host/host.go
```diff
@@ -199,6 +486,12 @@ func (h *Host) Close() error {
 	h.closed = true
 	h.resourceLock.Unlock()
 
+	// Close the bolt database.
+	err = h.db.Close()
+	if err != nil {
+		return err
+	}
```
lukechampine: Should probably delay returning the error; don't want to stop short before clearing the port, saving the state, etc.

modules/host/host_consistency.go
```diff
@@ -0,0 +1,86 @@
+package host
+
+import (
+	"errors"
+)
+
+// consistency.go contians a bunch of consistency checks for the host. Because
```
lukechampine: typo

modules/host/info.go
```diff
@@ -57,21 +104,24 @@ func (h *Host) RPCMetrics() modules.HostRPCMetrics {
...
 	defer h.mu.Unlock()
+	// Resource lock is grabbed for the save function.
 	h.resourceLock.RLock()
 	defer h.resourceLock.RUnlock()
 	if h.closed {
 		return errHostClosed
 	}
 
-	// Check that the unlock hash was not changed.
+	// Check that none of the illegal fields have been modified.
```
lukechampine: perhaps the structs should be split up. It sounds like they have diverged significantly. It would be convenient if one could be embedded in the other, but perhaps not even that is possible.

lukechampine:
taking a break before I tackle the real meat of this PR...

modules/host/sector.go
```diff
@@ -0,0 +1,280 @@
...
+// 16PB of data.
+//
+// 12 bytes can be represented as a filepath using 16 base64 characters. This
+// keeps the filesize small and therefore limits the amount of load placed on
+// the filesystem when trying to manage hundreds of thousands or even tens of
+// millions of sectors in a single folder.
+func (h *Host) sectorID(sectorRootBytes []byte) []byte {
+	saltedRoot := crypto.HashAll(sectorRootBytes, h.sectorSalt)
+	id := make([]byte, 16)
+	base64.RawURLEncoding.Encode(id, saltedRoot[:12])
```
lukechampine: there is some coupling between the 12 and 16 here. Maybe use `base64.EncodedLen(12)` in place of the 16?

modules/host/sector.go
```diff
@@ -0,0 +1,280 @@
...
+
+	// Determine which storage folder is going to receive the new sector.
+	err := h.db.Update(func(tx *bolt.Tx) error {
+		// Check whether the sector is a virtual sector.
+		sectorKey := h.sectorID(sectorRoot[:])
+		bsu := tx.Bucket(bucketSectorUsage)
+		usageBytes := bsu.Get(sectorKey)
+		var usage sectorUsage
+		if usageBytes != nil {
+			// usageBytes != nil indicates that this sector is already a in the
```
lukechampine: typo

modules/host/sector.go
```diff
@@ -0,0 +1,280 @@
...
+		}
+
+		// Try adding the sector to disk. In the event of a failure, the host
+		// will try the next storage folder until there is either a success or
+		// until all options have been exhausted.
+		potentialFolders := h.storageFolders
+		emptiestFolder, emptiestIndex := emptiestStorageFolder(potentialFolders)
+		for emptiestFolder != nil {
+			sectorPath := filepath.Join(h.persistDir, emptiestFolder.uidString(), string(sectorKey))
+			err := ioutil.WriteFile(sectorPath, sectorData, 0700)
```
lukechampine: `WriteFile` does not delete the file if an error was encountered, which I assume you want to do

modules/host/sector.go
```diff
@@ -0,0 +1,280 @@
...
+					break
+				}
+			}
+			if i == len(usage.Expiry) {
+				return errors.New("removing a sector that doesn't seem to exist")
+			}
+			if i == len(usage.Expiry)-1 {
+				usage.Expiry = append(usage.Expiry[0:i])
+			} else {
+				usage.Expiry = append(usage.Expiry[0:i], usage.Expiry[i+1:]...)
```
lukechampine: you can combine these branches; `append(usage.Expiry[:i], usage.Expiry[i+1:])` works for the `i == len(usage.Expiry-1)` case

modules/host/sector.go
lukechampine: there is some coupling between the 12 and 16 here. Maybe use `base64.EncodedLen(12)` in place of the 16?

also, why does this function take a slice instead of a `Hash`?

modules/host/sector.go
lukechampine: typo

modules/host/sector.go
lukechampine: `WriteFile` does not delete the file if an error was encountered, which I assume you want to do. Just need to stick an `os.Remove` in the err branch.

modules/host/sector.go
lukechampine: you can combine these branches; `append(usage.Expiry[:i], usage.Expiry[i+1:])` works for the `i == len(usage.Expiry-1)` case

modules/host/sector.go
```diff
@@ -0,0 +1,280 @@
...
+
+		sectorPath := filepath.Join(h.persistDir, hex.EncodeToString(usage.StorageFolder), string(sectorKey))
+		err = os.Remove(sectorPath)
+		if err != nil {
+			return err
+		}
+
+		// Find the storage folder that contains the sector.
+		for _, sf := range h.storageFolders {
+			if bytes.Compare(sf.UID, usage.StorageFolder) == 0 {
```
lukechampine: use `bytes.Equal`

modules/host/storagefolders.go
```diff
@@ -0,0 +1,594 @@
+package host
+
+// storgaefolder.go is responsible for managing the storage folders within the
```
lukechampine: typo

modules/host/storagefolders.go
```diff
@@ -0,0 +1,594 @@
...
+
+// emptiestStorageFolder takes a set of storage folders and returns the storage
+// folder with the lowest utilization by percentage. 'nil' is returned if there
+// are no storage folders provided with sufficient free space for a sector.
+//
+// Refusing to return a storage folder that does not have enough space prevents
+// the host from overfilling a storage folder.
+func emptiestStorageFolder(sfs []*storageFolder) (*storageFolder, int) {
+	lowestUtilization := float64(2) // Set higher than the max of 1 to protect against floating point imprecision.
+	winningIndex := -1              // Set to impossible value to prevent unintentionally returning the wrong storage folder.
```
lukechampine: why use an index instead of a `*storageFolder`?
I think that would allow you do remove the `winner` boolean too. You check at the end if `winningFolder == nil`

modules/host/storagefolders.go
```diff
@@ -0,0 +1,594 @@
...
+		if err != nil {
+			return err
+		}
+
+		// Check for collsions. Check should be relatively inexpensive at all
+		// times, because the total number of storage folders is limited to
+		// 256.
+		safe := true
+		for _, sf := range h.storageFolders {
+			if bytes.Compare(newSF.UID, sf.UID) == 0 {
```
lukechampine: `bytes.Equal`

modules/host/storagefolders.go
```diff
@@ -0,0 +1,594 @@
...
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
+	// Create a list of available folders. As folders are filled up, this list
+	// will be pruned. Once all folders are full, the offload loop will quit
+	// and return with errIncompleteOffload.
+	availableFolders := make([]*storageFolder, 0, maximumStorageFolders)
```
lukechampine: probably overkill to preallocate here if `maximumStorageFolders` is small

modules/host/storagefolders.go
```diff
@@ -0,0 +1,594 @@
...
+	dataOffloaded := uint64(0)
+	for currentSectorID != nil && dataOffloaded < dataToOffload && len(availableFolders) > 0 {
+		err = h.db.Update(func(tx *bolt.Tx) error {
+			// Determine whether the sector needs to be moved.
+			var usage sectorUsage
+			err = json.Unmarshal(currentSectorBytes, &usage)
+			if err != nil {
+				return err
+			}
+			if bytes.Compare(usage.StorageFolder, offloadFolder.UID) != 0 {
```
lukechampine: `bytes.Equal`

modules/host/storagefolders.go
```diff
@@ -0,0 +1,594 @@
...
+				if err != nil {
+					// TODO: Document the write failure so that the user can be
+					// notified of shortcomings of his disk.
+
+					// Because the write failed, we should move on to the next
+					// storage folder and try that.
+					if emptiestIndex == len(potentialFolders)-1 {
+						potentialFolders = potentialFolders[:emptiestIndex-1]
+					} else {
+						potentialFolders = append(potentialFolders[0:emptiestIndex], potentialFolders[emptiestIndex+1:]...)
```
lukechampine: can be combined

modules/host/storagefolders.go
```diff
@@ -0,0 +1,594 @@
...
+	// Check that the removal folder exists, and create a shortcut to it.
+	if removalIndex >= len(h.storageFolders) || removalIndex < 0 {
+		return errBadStorageFolderIndex
+	}
+	removalFolder := h.storageFolders[removalIndex]
+
+	// Move all of the sectors in the storage folder to other storage folders.
+	usedSize := removalFolder.Size - removalFolder.SizeRemaining
+	offloadErr := h.offloadStorageFolder(removalFolder, usedSize)
+	// If 'force' is set, we want to ignore 'errIncopmleteOffload' and try to
```
lukechampine: typo

modules/host/storagefolders.go
```diff
@@ -0,0 +1,594 @@
...
+	// remove the storage folder anyway. For any other error, we want to halt
+	// and return the error.
+	if (offloadErr != nil && offloadErr != errIncompleteOffload) || (offloadErr != nil && !force) {
```
lukechampine: might be more clearly written as:
```go
if force && offloadErr == errIncompleteOffload {
    offloadErr = nil
}
if offloadErr != nil {
    return offloadErr
}
```

modules/host/storagefolders.go
```diff
@@ -0,0 +1,594 @@
...
+		return offloadErr
+	}
+
+	// Remove the storage folder from the host and then save the host.
+	oldStorageFolders := h.storageFolders
+	if removalIndex == len(h.storageFolders)-1 {
+		h.storageFolders = h.storageFolders[0:removalIndex]
+	} else {
+		h.storageFolders = append(h.storageFolders[0:removalIndex], h.storageFolders[removalIndex+1:]...)
```
lukechampine: combine

modules/host/storagefolders_smoke_test.go
```diff
@@ -0,0 +1,877 @@
...
+	if err != errSmallStorageFolder {
+		t.Fatal("expecting errSmallStorageFolder:", err)
+	}
+	// Try linking to a storage folder that does not exist.
+	err = ht.host.AddStorageFolder(storageFolderOne, minimumStorageFolderSize)
+	if err == nil {
+		t.Fatal("should not be able to link to a storage folder which does not exist")
+	}
+	// Try linking to a storage folder that's not a directory.
+	err = ioutil.WriteFile(storageFolderOne, make([]byte, minimumStorageFolderSize), 0700)
```
lukechampine: I noticed that `ioutil` is being used in a lot of places instead of `dependencies`

modules/host/storagefolders.go
```diff
@@ -0,0 +1,594 @@
...
+	}
+	if resizeFolder.Size == newSize {
+		return errNoResize
+	}
+
+	// Sectors do not need to be moved onto or away from the resize folder if
+	// the folder is growing, or if after being shrunk the folder still has
+	// enough storage to house all of the sectors it currently tracks.
+	resizeFolderSizeConsumed := resizeFolder.Size - resizeFolder.SizeRemaining
+	if resizeFolder.Size < newSize || resizeFolderSizeConsumed <= newSize {
```
lukechampine: `resizeFolderSizeConsumed` is strictly less than `resizeFolder.Size`, so you only need to check for `resizeFolderSizeConsumed <= newSize`

modules/host_test.go
```diff
@@ -99,3 +99,76 @@ func TestUnitStoragePriceConversions(t *testing.T) {
...
+	// Establish a series of trials for conversion.
+	trials := []struct {
+		consensus types.Currency
+		human     uint64
+	}{
+		{
+			// Convert 0.
+			types.ZeroCurrency,
+			0,
+		},
```
lukechampine: I usually do these on one line

lukechampine:
ok, pretty much done for now. I didn't review the tests very thoroughly

modules/host.go
DavidVorick: Make a call to `StorageFolders()` to see the list of storage folders by index.

DavidVorick: I could change it to use the UID, but I'm not convinced that's the best design. `AddStorageFolders` would then need to return the UID, and you'd potentially run into problems if you started deleting and re-adding storage folders - UID selection is random, and that might not be intuitive.

build/critical.go
DavidVorick: Don't need the `else`, a panic will stop the codeflow.

modules/host.go
DavidVorick: They aren't, and I've decided it's premature to be adding this function. It's the type of thing that should be pretty easy to tag on later.

modules/host.go
DavidVorick: ah whoops. Altruistic limits are the limits for giving away stuff for free. I'm expecting a lot of hosts early on to be comfortable doing X GB / mo of free bandwidth, just to encourage network adoption. Priced limits are the limits when people are paying for the bandwidth. Presumably the limits are going to be much higher when there is payment.

DavidVorick: This should probably be thought through better as well. I'm guessing hosts are willing to go over their limits if the price is high enough. Maybe there should be a slice of limits, with each indicating the acceptable price level at those limits? That would add a bit of programmability.

modules/host/dependencies.go
DavidVorick: `logClose` vs `bufferClose` if you run into namespace issues. Not sure how I feel about lower-casing, I guess it makes sense.

modules/host/host.go
DavidVorick: When you create a file contract, the host charges a flat fee for the creation of the file contract. This gives the host a fee budget to submit revisions and storage proofs. I believe that 50 is too high, and I'm worried that doing this will make it attractive for hosts to accept file contracts that they never plan on revising or don't have the storage to accommodate, just to get the flat bonus.

Might make sense to have some default where the host throws up some collateral or something. Not sure yet. If the host doesn't have a contract price, then the renter can be abusive by having a ton of small, underutilized file contracts and the host can't submit storage proofs or anything. A flat rate for creating file contracts makes sense, just need to make sure there's a limited amount of abuse that the host can get away with.

modules/host/host.go
DavidVorick: I've had these checks already catch mistakes, and your suggestion wouldn't help. My mistake: `iif build.Release == "tsting" {`. Having the check in every function is more to make sure that you spelt all the words correctly than it is to make sure the binary was compiled with a reasonable build flag.

That could probably be avoided with type safety, but then there's another thing that I'm using this to protect against. If we ever add a 4th type of constants ("testnet" for an obvious potential - dev constants are too aggressive for a global testnet), we'd want all the funcs to panic until we'd chosen sane values for each of them. You'd really want to give manual attention to each one to make sure you're not forgetting anything important.

So, I like the construction that's in place for multiple reasons.

modules/host/host.go
DavidVorick: yeah

modules/host/host.go
DavidVorick: It would work in this case, though it would stop working if you added extra code like inserting some data or something (like the genesis block). I think when I created this function I had an intention of adding some base data, but later changed the architecture so it wasn't needed.

modules/host/host_consistency.go
DavidVorick: one of many most likely.

I'll go through with spellcheck when the PR is in a later stage. I'm still adding a lot of comments and changing stuff around, not too worried about typos at this point.

Though, I guess it's good to call them as you see them, lest they get missed on the next pass.

modules/host/info.go
DavidVorick: perhaps. It would be nice to be able to add internal-only fields such as the auto-pricing bools.

modules/host/sector.go
DavidVorick: There's a good reason I had for using bytes as input instead of a hash, but it's currently escaping me. I think it had something to do with the call to `crypto.HashAll`, if you pass a hash in it'll have a different result because you end up prepending 8 bytes of `len` when it's a slice.

hmm, struggling to remember all the details though. It might not be needed anymore, definitely wish I had dropped a comment when I made it.

modules/host/sector.go
DavidVorick: you wrote this comment before I had integrated the dependency mocking into the core logic

DavidVorick: but, noted about needing to follow WriteFile with Remove

modules/host/storagefolders.go
DavidVorick: I think it's the same number of lines either way. The winning index gets returned.

DavidVorick: hmm, just looking at how it's used, there may be a cleaner way to structure this.

modules/host/storagefolders.go
DavidVorick: yeah, that does seem better.

DavidVorick:
This code is mostly still pre-review stage. More just wanted you to see the high level architecture. I lot of the messiness will probably be polished up as I start testing more aggressively. One thing I like about aiming at 100% test coverage is that it forces you to go back and reconsider everything you wrote previously, helping you to clean up messy code and add comments where you may have missed them.

modules/host/sector.go
lukechampine: yeah, definitely good to keep in mind that the encoding is different. So a slice might make sense if you think the length may change in the future.

modules/host.go
lukechampine: because you'd get two different UIDs when adding the same folder two different times? Hmm.
From an outsider perspective, what sticks out to me is that the index doesn't seem to carry any meaning. It implies that there's an ordering, but it's not obvious what that ordering is or why it would be useful. With a UID, it's clear that there's no hidden meaning; it's just a token used to reference a resource.
A UID would also let you use a map (of `*storageFolder`s) internally instead of a slice, which seems like it might be preferable, though I can't say for sure.

modules/host/dependencies.go
lukechampine: so, this is sort of a special case because none of the dependencies here are implemented by existing objects. For example, instead of using the logger's methods (e.g. `Println`, `Close`), you use `NewLogger`. And the other modules aren't included in this interface at all. If you wanted to combine their methods into this single interface, you might have some collisions.

DavidVorick:
OK. Test coverage on sectors.go and storagefolders.go is 91% and 95% respectively. The remaining errors are mostly json stuff and bolt stuff, and really shouldn't appear in production. I'm pretty burned out on this stuff, but maybe I'll get them up to 100% before we move this feature-branch into master.

I'm finished coding on the multi-host branch, I'll be moving to another branch to tackle/finish the storage obligations + storage obligation testing. Once the storage obligations are done, the host announcements need to be tackled, and then the negotiation protocols. The settings management needs to be reconsidered, and then consistency checks need to be added all over the place, potentially with repair tools.

DavidVorick:
I will add that the code should be pretty hardened at this point.

modules/host/host_errors_test.go
```diff
@@ -0,0 +1,202 @@
...
+	"github.com/NebulousLabs/bolt"
+)
+
+// dependencyErrMkdirAll is a dependency set that returns an error when MkdirAll
+// is called.
+type dependencyErrMkdirAll struct {
+	productionDependencies
+}
+
+func (dependencyErrMkdirAll) mkdirAll(_ string, _ os.FileMode) error {
```
lukechampine: underscores can be omitted in most function definitions

modules/host/host.go
```diff
@@ -135,47 +423,106 @@ func New(cs modules.ConsensusSet, tpool modules.TransactionPool, wallet modules.
...
 
 	return h, nil
 }
 
+// New returns an initialized Host.
+func New(cs modules.ConsensusSet, tpool modules.TransactionPool, wallet modules.Wallet, address string, persistDir string) (*Host, error) {
+	return newHost(productionDependencies{}, cs, tpool, wallet, address, persistDir)
+}
+
+// composeErrors will take two errors and compose them into a single errors
```
lukechampine: can be more than two.

separating with "and" might be confusing if the error string contains the word "and." Maybe a semicolon would be better.
```go
func compose(errs ...error) error {
	var errStrings []string
	for _, err := range errs {
		if err != nil {
			errStrings = append(errStrings, err.Error())
		}
	}
	return errors.New(strings.Join(errStrings, "; "))
}
```

modules/host/storagefolders_smoke_errors_test.go
```diff
@@ -0,0 +1,490 @@
...
+// readFile reads a file from the filesystem. The call will fail if reading
+// from a file that has a substring which matches the ffs list of broken
+// substrings.
+func (ffs faultyFS) readFile(s string) ([]byte, error) {
+	for _, bs := range ffs.brokenSubstrings {
+		if strings.Contains(s, bs) {
+			return nil, mockErrReadFile
+		}
+	}
+	return ioutil.ReadFile(s)
```
lukechampine: `ffs.productionDependencies.readFile(s)` might be more robust here

Issue #988, opened 2016-02-21
Add timeouts to all gateway operations


VoidingWarranties:
If a connection is dropped in the middle of an RPC the RPC func will hang forever. `Conn.SetDeadline` can be used to timeout a read/write to a conn.

lukechampine:
yup, the host already does this by setting a 5-minute deadline on every incoming call. RPCs can extend the deadline if they desire.

DavidVorick:
*can extend the deadlines a bounded number of times

DavidVorick:
status update on this?

PR #992, opened 2016-02-24
Header flooding

VoidingWarranties:
This PR implements header broadcasting.

I tried to keep the changes to existing consensus code as minimal as possible. The only real changes to existing code should be:
* `validateHeader` renamed to `validateHeaderAndBlock`
* `AcceptBlock` and `threadedReceiveBlocks` broadcast blocks to some peers and headers to others, depending on their versions.

Everything else is new code and only deals with header broadcasting.

`RelayHeader` RPC mirrors the `RelayBlock` RPC used for block flooding. Every call to `RelayBlock` is also replaced with a branching broadcast that broadcasts `RelayHeader` to new peers (v0.5.2 and newer) and `RelayBlock` to old peers.

`managedAcceptHeader` mirrors `managedAcceptBlock` and does all of the header validation. However, unlike its counterpart, it cannot accept anything into the blockchain as it does not have the block and header-first blockchain downloads haven't been implemented yet. Instead, it returns `nil` if the header is valid. It is then up to the caller of `managedAcceptHeader` ( `RelayHeader`) to download the block and call either `managedAcceptBlock` or `AcceptBlock` on it.

`Send1Blk` is analogous to `SendBlocks`, but for just one block. It is used to download a block corresponding to a validated block header. Peers can request blocks from other peers by the block id. The requesting peer then calls `AcceptBlock` on the received block.

I tried to keep the commits clean and separated, so going through the diffs commit by commit should be easy.

### DIFF
--- a/modules/consensus/accept.go
+++ b/modules/consensus/accept.go
@@ -198,7 +198,7 @@ func (cs *ConsensusSet) managedAcceptBlock(b types.Block) error {
 	err := cs.db.View(func(tx *bolt.Tx) error {
 		// Do not accept a block if the database is inconsistent.
 		if inconsistencyDetected(tx) {
-			return errors.New("inconsistent database")
+			return errInconsistentSet
 		}
 
 		// Do some relatively inexpensive checks to validate the header and block.
@@ -264,4 +264,26 @@ func (cs *ConsensusSet) AcceptBlock(b types.Block) error {
 	peers := cs.gateway.Peers()
 	go cs.gateway.Broadcast("RelayBlock", b, peers)
 	return nil
+}
+
+// managedAcceptHeader will validate a block header and see if it extends the
+// heaviest chain. If a header is valid and would extend the heaviest chain,
+// nil is returned. managedAcceptHeader does not fetch the corresponding block
+// and therefore can't accept the block itself. Just as managedAcceptBlock does
+// not broadcast the block, managedAcceptHeader does not broadcast the block
+// header.
+func (cs *ConsensusSet) managedAcceptHeader(h types.BlockHeader) error {
+	cs.mu.Lock()
+	defer cs.mu.Unlock()
+	// Start verification inside of a bolt View tx.
+	err := cs.db.View(func(tx *bolt.Tx) error {
+		// Do not accept a header if the database is inconsistent.
+		if inconsistencyDetected(tx) {
+			return errInconsistentSet
+		}
+
+		// Do some relatively inexpensive checks to validate the header
+		return cs.validateHeader(boltTxWrapper{tx}, h)
+	})
+	return err
 }
--- a/modules/consensus/accept.go
+++ b/modules/consensus/accept.go
@@ -108,6 +108,8 @@ func (cs *ConsensusSet) validateHeader(tx dbTx, h types.BlockHeader) error {
 	// TODO: check if the block is in the extreme or near future, and return
 	// errExtremeFutureTimestamp or errFutureTimestamp, respectively.
 
+	// TODO: check if the block is a non extending block.
+
 	return nil
 }
 
@@ -210,6 +212,7 @@ func (cs *ConsensusSet) managedAcceptBlock(b types.Block) error {
 			if err == errFutureTimestamp {
 				go func() {
 					time.Sleep(time.Duration(b.Timestamp-(types.CurrentTimestamp()+types.FutureThreshold)) * time.Second)
+					// TODO: log error returned by AcceptBlock if non-nill?
 					cs.AcceptBlock(b) // NOTE: Error is not handled.
 				}()
 			}
--- a/modules/consensus/accept.go
+++ b/modules/consensus/accept.go
@@ -23,7 +23,8 @@ var (
 // block. Callers should not assume that validation will happen in a particular
 // order.
 func (cs *ConsensusSet) validateHeaderAndBlock(tx dbTx, b types.Block) error {
-	// See if the block is known already.
+	// Check if the block is a DoS block - a known invalid block that is expensive
+	// to validate.
 	id := b.ID()
 	_, exists := cs.dosBlocks[id]
 	if exists {
@@ -56,7 +57,7 @@ func (cs *ConsensusSet) validateHeaderAndBlock(tx dbTx, b types.Block) error {
 	return cs.blockValidator.ValidateBlock(b, minTimestamp, parent.ChildTarget, parent.Height+1)
 }
 
-// checkTarget returns true if the header's ID meets the given target.
+// checkHeaderTarget returns true if the header's ID meets the given target.
 func checkHeaderTarget(h types.BlockHeader, target types.Target) bool {
 	blockHash := h.ID()
 	return bytes.Compare(target[:], blockHash[:]) >= 0
@@ -66,7 +67,8 @@ func checkHeaderTarget(h types.BlockHeader, target types.Target) bool {
 // to determine if the block should be downloaded. Callers should not assume
 // that validation will happen in a particular order.
 func (cs *ConsensusSet) validateHeader(tx dbTx, h types.BlockHeader) error {
-	// See if the block is known already.
+	// Check if the block is a DoS block - a known invalid block that is expensive
+	// to validate.
 	id := h.ID()
 	_, exists := cs.dosBlocks[id]
 	if exists {
@@ -99,7 +101,8 @@ func (cs *ConsensusSet) validateHeader(tx dbTx, h types.BlockHeader) error {
 		return modules.ErrBlockUnsolved
 	}
 
-	// TODO: check if the block is a non extending block.
+	// TODO: check if the block is a non extending block once headers-first
+	// downloads are implemented.
 
 	// Check that the timestamp is not too far in the past to be acceptable.
 	minTimestamp := cs.blockRuleHelper.minimumValidChildTimestamp(blockMap, &parent)
@@ -219,10 +222,20 @@ func (cs *ConsensusSet) managedAcceptBlock(b types.Block) error {
 			// If the block is in the near future, but too far to be acceptable, then
 			// save the block and add it to the consensus set after it is no longer
 			// too far in the future.
+			//
+			// TODO: an attacker could mine many blocks off the genesis block all in the
+			// future and we would spawn a goroutine per each block. To fix this, either
+			// ban peers that send lots of future blocks OR stop spawning goroutines
+			// after we are already waiting on a large number of future blocks.
+			//
+			// TODO: an attacker could broadcast a future block many times and we would
+			// spawn a goroutine for each broadcast. To fix this we should create a
+			// cache of future blocks, like we already do for DoS blocks, and only spawn
+			// a goroutine if we haven't already spawned one for that block.
 			if err == errFutureTimestamp {
 				go func() {
 					time.Sleep(time.Duration(b.Timestamp-(types.CurrentTimestamp()+types.FutureThreshold)) * time.Second)
-					// TODO: log error returned by AcceptBlock if non-nill?
+					// TODO: log error returned by AcceptBlock if non-nill
 					cs.AcceptBlock(b) // NOTE: Error is not handled.
 				}()
 			}
--- a/modules/consensus/accept.go
+++ b/modules/consensus/accept.go
@@ -1,6 +1,7 @@
 package consensus
 
 import (
+	"bytes"
 	"errors"
 	"time"
 
@@ -54,6 +55,12 @@ func (cs *ConsensusSet) validateHeaderAndBlock(tx dbTx, b types.Block) error {
 	minTimestamp := cs.blockRuleHelper.minimumValidChildTimestamp(blockMap, &parent)
 
 	return cs.blockValidator.ValidateBlock(b, minTimestamp, parent.ChildTarget, parent.Height+1)
+}
+
+// checkTarget returns true if the header's ID meets the given target.
+func checkHeaderTarget(h types.BlockHeader, target types.Target) bool {
+	blockHash := h.ID()
+	return bytes.Compare(target[:], blockHash[:]) >= 0
 }
 
 // addBlockToTree inserts a block into the blockNode tree by adding it to its
--- a/modules/consensus/accept.go
+++ b/modules/consensus/accept.go
@@ -45,7 +45,6 @@ func (cs *ConsensusSet) validateHeaderAndBlock(tx dbTx, b types.Block) error {
 	if parentBytes == nil {
 		return errOrphan
 	}
-
 	var parent processedBlock
 	err := cs.marshaler.Unmarshal(parentBytes, &parent)
 	if err != nil {
@@ -61,6 +60,55 @@ func (cs *ConsensusSet) validateHeaderAndBlock(tx dbTx, b types.Block) error {
 func checkHeaderTarget(h types.BlockHeader, target types.Target) bool {
 	blockHash := h.ID()
 	return bytes.Compare(target[:], blockHash[:]) >= 0
+}
+
+// validateHeader does some early, low computation verification on the header
+// to determine if the block should be downloaded. Callers should not assume
+// that validation will happen in a particular order.
+func (cs *ConsensusSet) validateHeader(tx dbTx, h types.BlockHeader) error {
+	// See if the block is known already.
+	id := h.ID()
+	_, exists := cs.dosBlocks[id]
+	if exists {
+		return errDoSBlock
+	}
+
+	// Check if the block is already known.
+	blockMap := tx.Bucket(BlockMap)
+	if blockMap == nil {
+		return errNoBlockMap
+	}
+	if blockMap.Get(id[:]) != nil {
+		return modules.ErrBlockKnown
+	}
+
+	// Check for the parent.
+	parentID := h.ParentID
+	parentBytes := blockMap.Get(parentID[:])
+	if parentBytes == nil {
+		return errOrphan
+	}
+	var parent processedBlock
+	err := cs.marshaler.Unmarshal(parentBytes, &parent)
+	if err != nil {
+		return err
+	}
+
+	// Check that the timestamp is not too far in the past to be acceptable.
+	minTimestamp := cs.blockRuleHelper.minimumValidChildTimestamp(blockMap, &parent)
+	if minTimestamp > h.Timestamp {
+		return errEarlyTimestamp
+	}
+
+	// Check that the target of the new block is sufficient.
+	if !checkHeaderTarget(h, parent.ChildTarget) {
+		return modules.ErrBlockUnsolved
+	}
+
+	// TODO: check if the block is in the extreme or near future, and return
+	// errExtremeFutureTimestamp or errFutureTimestamp, respectively.
+
+	return nil
 }
 
 // addBlockToTree inserts a block into the blockNode tree by adding it to its
--- a/modules/consensus/accept.go
+++ b/modules/consensus/accept.go
@@ -270,9 +270,17 @@ func (cs *ConsensusSet) AcceptBlock(b types.Block) error {
 	if err != nil {
 		return err
 	}
-	// Broadcast the new block to all peers.
-	peers := cs.gateway.Peers()
-	go cs.gateway.Broadcast("RelayBlock", b, peers)
+	// Broadcast the block to all peers <= v0.5.1 and block header to all peers > v0.5.1.
+	var relayBlockPeers, relayHeaderPeers []modules.Peer
+	for _, p := range cs.gateway.Peers() {
+		if build.VersionCmp(p.Version, "0.5.1") <= 0 {
+			relayBlockPeers = append(relayBlockPeers, p)
+		} else {
+			relayHeaderPeers = append(relayHeaderPeers, p)
+		}
+	}
+	go cs.gateway.Broadcast("RelayBlock", b, relayBlockPeers)
+	go cs.gateway.Broadcast("RelayHeader", b.Header(), relayHeaderPeers)
 	return nil
 }
 
--- a/modules/consensus/accept.go
+++ b/modules/consensus/accept.go
@@ -295,25 +295,3 @@ func (cs *ConsensusSet) AcceptBlock(b types.Block) error {
 	go cs.gateway.Broadcast("RelayHeader", b.Header(), relayHeaderPeers)
 	return nil
 }
-
-// managedAcceptHeader will validate a block header and see if it extends the
-// heaviest chain. If a header is valid and would extend the heaviest chain,
-// nil is returned. managedAcceptHeader does not fetch the corresponding block
-// and therefore can't accept the block itself. Just as managedAcceptBlock does
-// not broadcast the block, managedAcceptHeader does not broadcast the block
-// header.
-func (cs *ConsensusSet) managedAcceptHeader(h types.BlockHeader) error {
-	cs.mu.Lock()
-	defer cs.mu.Unlock()
-	// Start verification inside of a bolt View tx.
-	err := cs.db.View(func(tx *bolt.Tx) error {
-		// Do not accept a header if the database is inconsistent.
-		if inconsistencyDetected(tx) {
-			return errInconsistentSet
-		}
-
-		// Do some relatively inexpensive checks to validate the header
-		return cs.validateHeader(boltTxWrapper{tx}, h)
-	})
-	return err
-}
--- a/modules/consensus/accept.go
+++ b/modules/consensus/accept.go
@@ -94,21 +94,31 @@ func (cs *ConsensusSet) validateHeader(tx dbTx, h types.BlockHeader) error {
 		return err
 	}
 
+	// Check that the target of the new block is sufficient.
+	if !checkHeaderTarget(h, parent.ChildTarget) {
+		return modules.ErrBlockUnsolved
+	}
+
+	// TODO: check if the block is a non extending block.
+
 	// Check that the timestamp is not too far in the past to be acceptable.
 	minTimestamp := cs.blockRuleHelper.minimumValidChildTimestamp(blockMap, &parent)
 	if minTimestamp > h.Timestamp {
 		return errEarlyTimestamp
 	}
 
-	// Check that the target of the new block is sufficient.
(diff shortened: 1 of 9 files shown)


modules/consensus/synchronize.go
```diff
@@ -245,16 +257,106 @@ func (cs *ConsensusSet) relayBlock(conn modules.PeerConn) error {
...
+// the function it returns calls the exported method AcceptBlock.
+func (cs *ConsensusSet) threadedReceiveBlock(id types.BlockID) modules.RPCFunc {
+	managedFN := func(conn modules.PeerConn) error {
+		if err := encoding.WriteObject(conn, id); err != nil {
+			return err
+		}
+		var block types.Block
+		if err := encoding.ReadObject(conn, &block, types.BlockSizeLimit); err != nil {
+			return err
+		}
```
VoidingWarranties: If `Send1Blk` errors out (if, for example the peer doesn't have the requested block), this call will block indefinitely. This would be fixed by #988

modules/consensus/accept_test.go
```diff
@@ -417,6 +612,16 @@ func TestOrphanHandling(t *testing.T) {
 	if err != errOrphan {
 		t.Fatalf("expected %v, got %v", errOrphan, err)
 	}
+	// Try submitting an orphan block's header.
+	// TODO: why is the test above duplicated?
```
VoidingWarranties: No harm in duplicating the test, but was it purposeful?

modules/consensus/synchronize.go
```diff
@@ -245,16 +257,106 @@ func (cs *ConsensusSet) relayBlock(conn modules.PeerConn) error {
...
+		// TODO: are these sanity checks necessary?
+		pathID, err := getPath(tx, pb.Height)
+		if err != nil {
+			return err
+		}
+		if pathID != pb.Block.ID() {
+			return errors.New("pathID and processed block's ID do not match")
+		}
```
VoidingWarranties: What is the reason for these sanity checks? (they also exist in `sendBlocks`)

modules/consensus/synchronize_test.go
```diff
@@ -311,3 +316,488 @@ func TestSendBlocksBroadcastsOnce(t *testing.T) {
...
+				if err != nil {
+					t.Error(err)
+				}
+				fnDone <- struct{}{}
+			},
+			errWant: errNilItem,
+			msg:     "expected send1Blk to error with a nonexistent block id",
+		},
+		// TODO: Other sanity checks? e.g. test if getPath returns an error or pathID != pb.Block.ID().
+		// TODO: Test with an otherwise failing db?
```
VoidingWarranties: Wasn't sure how to easily test these.

modules/consensus/synchronize_test.go
```diff
@@ -311,3 +316,488 @@ func TestSendBlocksBroadcastsOnce(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer cst2.Close()
+
+	err = cst1.cs.gateway.Connect(cst2.cs.gateway.Address())
+	if err != nil {
+		t.Fatal(err)
+	}
+	err = cst2.cs.gateway.Connect(cst1.cs.gateway.Address()) // TODO: why is this Connect call necessary? Shouldn't they connect to eachother with one Connect?
```
VoidingWarranties: A gateway receiving a connection request should connect to the requesting gateway, but this doesn't happen here. Why?

modules/consensus/accept.go
```diff
@@ -55,6 +56,73 @@ func (cs *ConsensusSet) validateHeader(tx dbTx, b types.Block) error {
...
+	if err != nil {
+		return err
+	}
+
+	// Check that the target of the new block is sufficient.
+	if !checkHeaderTarget(h, parent.ChildTarget) {
+		return modules.ErrBlockUnsolved
+	}
+
+	// TODO: check if the block is a non extending block.
```
VoidingWarranties: I _think_ this can be determined with just a header, but the current implementation requires a `processedBlock` and we don't want processedBlocks if we don't have the whole block.

modules/consensus/accept.go
```diff
@@ -154,6 +222,7 @@ func (cs *ConsensusSet) managedAcceptBlock(b types.Block) error {
 			if err == errFutureTimestamp {
 				go func() {
 					time.Sleep(time.Duration(b.Timestamp-(types.CurrentTimestamp()+types.FutureThreshold)) * time.Second)
+					// TODO: log error returned by AcceptBlock if non-nill?
```
VoidingWarranties: Would be easy if the consensus had logging, but since it doesn't yet, adding it should be another PR.

DavidVorick:
should change rpc names to 8 chars, enforced by panics at runtime

modules/consensus/accept.go
DavidVorick: you can determine that with just a header, though you don't want to trust it because you don't actually know if the data for the block exists. And you don't want to skip downloading the block if it's non-extending because you need to be able to get reorgs.

A check here doesn't make sense until we're doing headers-first downloads.

modules/consensus/accept_test.go
DavidVorick: I remember that there was a good reason, but I forget why.

types/block.go
```diff
@@ -99,6 +103,13 @@ func (b Block) ID() BlockID {
 	return BlockID(crypto.HashObject(b.Header()))
 }
 
+// ID returns the ID of a Block, which is calculated by hashing the
+// concatenation of the block's parent's ID, nonce, and the result of the
+// b.MerkleRoot().
+func (h BlockHeader) ID() BlockID {
```
DavidVorick: The comment here is confusing. While correct, the logic just hashes the header, which doesn't seem to match the comment.

modules/consensus/synchronize.go
DavidVorick: that's bad. This should be tolerant and should not rely on the gateway forcing timeouts.

modules/consensus/synchronize.go
DavidVorick: This was to help guarantee that the path was being constructed correctly, and that the processed blocks were being saved correctly (instead of overwriting eachother). That's pretty well tested at this point, the sanity check can be safely removed I think.

modules/consensus/synchronize_test.go
DavidVorick: does it have something to do with inbound/outbound peers, or some level of mocking that's happening somewhere? Perhaps it's a bug.

types/block_test.go
```diff
@@ -113,6 +113,44 @@ func TestBlockID(t *testing.T) {
 	}
 }
 
+// TestHeaderID probes the ID function of the BlockHeader type.
+func TestHeaderID(t *testing.T) {
```
DavidVorick: might make sense to throw a sanity check somewhere in the consensus code to make sure that the block id and the header id always yield the same result. Then we'd get to verify it when running a debugging instance of 'siad', because it would check across a wide range of block types.

types/block.go
```diff
@@ -13,6 +13,10 @@ import (
 	"github.com/NebulousLabs/Sia/encoding"
 )
 
+const (
+	BlockHeaderSize = 80 // 32 (ParentID) + 8 (Nonce) + 8 (Timestamp) + 32 (MerkleRoot)
```
DavidVorick: needs a docstring

modules/gateway/rpc.go
```diff
@@ -114,6 +114,7 @@ func (g *Gateway) threadedHandleConn(conn modules.PeerConn) {
 		g.log.Printf("WARN: incoming conn %v requested unknown RPC \"%v\"", conn.RemoteAddr(), id)
 		return
 	}
+	g.log.Printf("INFO: incoming conn %v requested RPC \"%v\"", conn.RemoteAddr(), id)
```
DavidVorick: if you put this here the gateway.log is going to be too big. It's already gigabytes large for some longer-running nodes. this log statement can't go in a normal log (Luke likes the idea of a rolling log, though I'm not sure that's the best approach - it's probably good enough in this case to just not log at all)

DavidVorick: *if you really want the log, have it write the log in debug mode. bitcoind will make stupid-large logs (sometimes running into the hundreds of GBs) if you run it in debug/verbose mode. It's probably okay for us to take a similar approach. Non-debug logs only log when something unanticipated goes wrong (so, failed RPCs and random network drops would not merit logs), and debug logs just dump data.

modules/consensus/accept.go
```diff
@@ -55,6 +56,73 @@ func (cs *ConsensusSet) validateHeader(tx dbTx, b types.Block) error {
...
+	// the extreme future arrives, this block will no longer be a part of the
+	// longest fork because it will have been ignored by all of the miners.
+	if h.Timestamp > types.CurrentTimestamp()+types.ExtremeFutureThreshold {
+		return errExtremeFutureTimestamp
+	}
+
+	// Check if the block is in the near future, but too far to be acceptable.
+	if h.Timestamp > types.CurrentTimestamp()+types.FutureThreshold {
+		return errFutureTimestamp
+	}
```
DavidVorick: no need to have a check for errFutureTimestamp, that's logic that should be handled in acceptBlock only. You may want to add a comment explaining why. The reason is that you want to have the future blocks in memory, so that you can accept them the millisecond that they are no longer too far in the future to be acceptable. Maybe they will have reorged, maybe not. But you want the host to go chase down the block if it finds a header, even if it's a 'future' block.

modules/consensus/accept.go
DavidVorick: in that case leaving a TODO is fine.

modules/consensus/accept.go
```diff
@@ -201,8 +270,38 @@ func (cs *ConsensusSet) AcceptBlock(b types.Block) error {
...
+		// Do not accept a header if the database is inconsistent.
+		if inconsistencyDetected(tx) {
+			return errInconsistentSet
+		}
+
+		// Do some relatively inexpensive checks to validate the header
+		return cs.validateHeader(boltTxWrapper{tx}, h)
+	})
+	return err
+}
```
DavidVorick: why do you need a managed version of AcceptHeader? The reason you need a managed version of accept block is because the lock has to be demoted partway through the function, which means you can't wrap it in a lock. AcceptHeader on the other hand doesn't push any updates to the modules (nor does it even modify the consensus set), so you don't need to worry about demotion. I don't think this function needs to be here, nor does it need to be managed.

modules/consensus/synchronize.go
```diff
@@ -96,8 +98,17 @@ func (cs *ConsensusSet) threadedReceiveBlocks(conn modules.PeerConn) error {
...
+			var relayBlockPeers, relayHeaderPeers []modules.Peer
+			for _, p := range cs.gateway.Peers() {
+				if build.VersionCmp(p.Version, "0.5.1") <= 0 {
+					relayBlockPeers = append(relayBlockPeers, p)
+				} else {
+					relayHeaderPeers = append(relayHeaderPeers, p)
+				}
+			}
+			go cs.gateway.Broadcast("RelayBlock", currentBlock, relayBlockPeers)
+			go cs.gateway.Broadcast("RelayHeader", currentBlock.Header(), relayHeaderPeers)
```
DavidVorick: might be worth abstracting this code into a helper function, since it's written in a few places.

modules/consensus/synchronize.go
```diff
@@ -246,16 +257,106 @@ func (cs *ConsensusSet) relayBlock(conn modules.PeerConn) error {
...
+	switch err {
+	case errOrphan:
+		// If the header is an orphan, try to find the parents.
+		// TODO: log error returned if non-nill?
 		go cs.gateway.RPC(modules.NetAddress(conn.RemoteAddr().String()), "SendBlocks", cs.threadedReceiveBlocks)
+	case errFutureTimestamp:
+		// Ignore errFutureTimestamp so that the block is downloaded and passed to
+		// managedAcceptHeader, which will spawn a go thread that waits until the
+		// timestamp would be valid before trying to accept it again.
+		fallthrough
```
DavidVorick: since we're removing errFutureTimestamp from the set of errors that acceptHeader can return, there's no need for this case or fallthrough.

modules/consensus/synchronize.go
```diff
@@ -246,16 +257,106 @@ func (cs *ConsensusSet) relayBlock(conn modules.PeerConn) error {
...
 	return nil
 }
+
+// threadedReceiveBlock takes a block id and returns an RPCFunc that requests
+// that block and then calls AcceptBlock on it. The returned function should be
+// used as the calling end of the Send1Blk RPC. Note that although the function
+// itself does not do any locking, it is still prefixed with "threaded" because
+// the function it returns calls the exported method AcceptBlock.
+func (cs *ConsensusSet) threadedReceiveBlock(id types.BlockID) modules.RPCFunc {
+	managedFN := func(conn modules.PeerConn) error {
```
DavidVorick: `return func() error { ... }`

DavidVorick:
I haven't looked at the testing yet, otherwise I've finished making this round of comments.

modules/consensus/synchronize.go
VoidingWarranties: `SendBlocks` and `threadedReceiveBlocks` is also susceptible to this. The solution is simple enough I think. Something like:
```go
// send1Blk
err = cs.db.View(func(tx *bolt.Tx) error {
// ...
})
if err != nil {
	encoding.WriteObject(conn, types.Block{})
	return err
}
// ...

// sendBlocks
err = cs.db.View(func(tx *bolt.Tx) error {
// ...
})
if err != nil {
	encoding.WriteObject(conn, []types.Block)
	encoding.WriteObject(conn, false)
	return err
}
// ...
```
But then the errors from `encoding.WriteObject` are lost.

modules/consensus/synchronize.go
VoidingWarranties: I suppose you could request a block using `Send1Blk` that isn't in the longest chain. In that case the `pathID != pb.Block.ID()` would be true.

modules/consensus/synchronize.go
VoidingWarranties: Alternatively we could concatenate the errors like you did in #989 with `composeErrors`.

types/block_test.go
VoidingWarranties: Where would be a good place for the sanity check? This test does verify that `BlockHeader.ID()` and `Block.ID()` match for the test blocks FWIW.

DavidVorick: I'd probably throw the sanity check in validateHeaderAndBlock

all the test blocks is one thing, but all of the blocks in the whole consensus set is another level of assurance, especially because it's a sanity check that'll be run for every block that gets mined in all the integration tests as well

modules/consensus/synchronize.go
VoidingWarranties: Since RelayBlock will eventually be removed once enough peers upgrade I think it's okay to not use a helper function. This code is only duplicated in two places.

modules/consensus/synchronize.go
```diff
@@ -247,12 +256,16 @@ func (cs *ConsensusSet) relayBlock(conn modules.PeerConn) error {
...
 	err = cs.AcceptBlock(b)
 	if err == errOrphan {
 		// If the block is an orphan, try to find the parents. The block
 		// received from the peer is discarded and will be downloaded again if
 		// the parent is found.
+		//
+		// TODO: log error returned if non-nill? OR we could not use a goroutine and
+		// just return the result of the RPC call. Is it back practice to call an RPC
+		// from inside another RPC?
 		go cs.gateway.RPC(modules.NetAddress(conn.RemoteAddr().String()), "SendBlocks", cs.threadedReceiveBlocks)
```
VoidingWarranties: Could this line be
```go
return cs.gateway.RPC(modules.NetAddress(conn.RemoteAddr().String()), "SendBlocks", cs.threadedReceiveBlocks)
```
?

modules/consensus/synchronize.go
```diff
@@ -260,3 +273,96 @@ func (cs *ConsensusSet) relayBlock(conn modules.PeerConn) error {
...
+	})
+	cs.mu.RUnlock()
+	if err == errOrphan {
+		// If the header is an orphan, try to find the parents.
+		//
+		// TODO: log error returned if non-nill? OR we could not use a goroutine and
+		// just return the result of the RPC call. Is it back practice to call an RPC
+		// from inside another RPC?
+		go cs.gateway.RPC(modules.NetAddress(conn.RemoteAddr().String()), "SendBlocks", cs.threadedReceiveBlocks)
+		return nil
```
VoidingWarranties: Same for this. Could we return the result of the RPC call instead of spawning a goroutine?

VoidingWarranties:
Mostly done except for the few comments I just made.

Also it look like Github threw out all the comments @DavidVorick made on the individual commits because I rebased. :( I guess we should only make comments on the master diff.

modules/consensus/accept.go
```diff
@@ -140,20 +210,31 @@ func (cs *ConsensusSet) managedAcceptBlock(b types.Block) error {
...
+		err := cs.validateHeaderAndBlock(boltTxWrapper{tx}, b)
 		if err != nil {
 			// If the block is in the near future, but too far to be acceptable, then
 			// save the block and add it to the consensus set after it is no longer
 			// too far in the future.
+			//
+			// TODO: an attacker could mine many blocks off the genesis block all in the
+			// future and we would spawn a goroutine per each block. To fix this, either
+			// ban peers that send lots of future blocks OR stop spawning goroutines
+			// after we are already waiting on a large number of future blocks.
```
DavidVorick: probably both. You want a hard-limit on all resources that you use, I'd never advocate having a system that can spin up unlimited goroutines.

modules/consensus/accept.go
```diff
@@ -140,20 +210,31 @@ func (cs *ConsensusSet) managedAcceptBlock(b types.Block) error {
...
+			//
+			// TODO: an attacker could broadcast a future block many times and we would
+			// spawn a goroutine for each broadcast. To fix this we should create a
+			// cache of future blocks, like we already do for DoS blocks, and only spawn
+			// a goroutine if we haven't already spawned one for that block.
```
DavidVorick: and then start evicting blocks once there are 50. We'd evict the ones that are furthest in the future.

modules/consensus/accept_test.go
```diff
@@ -286,6 +311,181 @@ func TestUnitValidateHeader(t *testing.T) {
...
+	var h types.BlockHeader
+
+	tests := []struct {
+		target   types.Target
+		expected bool
+		msg      string
+	}{
+		{types.RootDepth, true, "checkHeaderTarget failed for a low target"},
+		{types.Target{}, false, "checkHeaderTarget passed for a high target"},
+		{types.Target(h.ID()), true, "checkHeaderTarget failed for a same target"},
```
DavidVorick: huh. I've always done this by manually mining a target that either did or didn't fit the criteria. The testing target is {128}, which means you have an exactly 50% chance of finding a valid header each attempt. Not too painful to write a loop that guarantees the correct outcome, it's cryptographically guaranteed to complete within 128 iterations.

The way you've done it is fine as well

modules/consensus/synchronize.go
```diff
@@ -96,8 +96,17 @@ func (cs *ConsensusSet) threadedReceiveBlocks(conn modules.PeerConn) error {
...
 			currentBlock := cs.CurrentBlock()
-			peers := cs.gateway.Peers()
-			go cs.gateway.Broadcast("RelayBlock", currentBlock, peers)
+			// Broadcast the block to all peers <= v0.5.1 and block header to all peers > v0.5.1
+			var relayBlockPeers, relayHeaderPeers []modules.Peer
+			for _, p := range cs.gateway.Peers() {
+				if build.VersionCmp(p.Version, "0.5.1") <= 0 {
+					relayBlockPeers = append(relayBlockPeers, p)
+				} else {
+					relayHeaderPeers = append(relayHeaderPeers, p)
```
DavidVorick: no coverage on this line of code

modules/consensus/synchronize.go
DavidVorick: I don't object to that.

DavidVorick: @lukechampine

modules/consensus/synchronize.go
DavidVorick: same for this

modules/consensus/synchronize.go
```diff
@@ -260,3 +273,96 @@ func (cs *ConsensusSet) relayBlock(conn modules.PeerConn) error {
...
+	} else if err != nil {
+		return err
+	}
+	// If the header is valid and extends the heaviest chain, fetch, accept it,
+	// and broadcast it.
+	//
+	// TODO: log error returned if non-nill? OR we could not use a goroutine and
+	// just return the result of the RPC call. Is it back practice to call an RPC
+	// from inside another RPC?
+	go cs.gateway.RPC(modules.NetAddress(conn.RemoteAddr().String()), "BlockID", cs.threadedReceiveBlock(h.ID()))
```
DavidVorick: and this

types/block_test.go
DavidVorick: actually, I see now that the definition of Block.ID is `crypto.HashObject(b.Header())`

sanity check is probably not needed

types/block.go
```diff
@@ -99,6 +105,11 @@ func (b Block) ID() BlockID {
 	return BlockID(crypto.HashObject(b.Header()))
 }
 
+// ID returns the ID of a Block, which is calculated by hashing the header.
+func (h BlockHeader) ID() BlockID {
```
DavidVorick: needs to be sorted by object. You've got a blockHeader function between a bunch of functions that operate on a block. I'd probably move it to above the block, though if you want to stick with the alphabetic convention you should move it below all the block functions.

Either one is fine with me, but I'd rather it not be in the middle of a bunch of block methods

modules/consensus/synchronize.go
```diff
@@ -260,3 +273,96 @@ func (cs *ConsensusSet) relayBlock(conn modules.PeerConn) error {
...
+		return err
+	}
+
+	// Start verification inside of a bolt View tx.
+	cs.mu.RLock()
+	err = cs.db.View(func(tx *bolt.Tx) error {
+		// Do not accept a header if the database is inconsistent.
+		if inconsistencyDetected(tx) {
+			return errInconsistentSet
+		}
```
DavidVorick: I don't think you need this. It's more of a startup sanity check, and was a bigger issue before we were atomically modifying the database.

modules/consensus/synchronize.go
```diff
@@ -260,3 +273,96 @@ func (cs *ConsensusSet) relayBlock(conn modules.PeerConn) error {
...
+		// TODO: log error returned if non-nill? OR we could not use a goroutine and
+		// just return the result of the RPC call. Is it back practice to call an RPC
+		// from inside another RPC?
+		go cs.gateway.RPC(modules.NetAddress(conn.RemoteAddr().String()), "SendBlocks", cs.threadedReceiveBlocks)
+		return nil
+	} else if err != nil {
+		return err
+	}
+	// If the header is valid and extends the heaviest chain, fetch, accept it,
+	// and broadcast it.
```
DavidVorick: too many  steps in the comment.

`If the header is valid, go get the corresponding block for full validation`.

modules/consensus/synchronize.go
```diff
@@ -260,3 +273,96 @@ func (cs *ConsensusSet) relayBlock(conn modules.PeerConn) error {
...
+	//
+	// TODO: log error returned if non-nill? OR we could not use a goroutine and
+	// just return the result of the RPC call. Is it back practice to call an RPC
+	// from inside another RPC?
+	go cs.gateway.RPC(modules.NetAddress(conn.RemoteAddr().String()), "BlockID", cs.threadedReceiveBlock(h.ID()))
+	return nil
+}
+
+// rpcBlockID is an RPC that sends the requested block to the requesting peer.
+func (cs *ConsensusSet) rpcBlockID(conn modules.PeerConn) error {
```
DavidVorick: I don't like this name, makes it seem like you are fetching a blockID. Maybe just `rpcBlock` is good enough.

modules/consensus/synchronize.go
VoidingWarranties: Actually this isn't a problem. When `Send1Blk` errors out the gateway closes the connection, which should cause any reads or writes on the other end to error out as well.

modules/consensus/synchronize.go
VoidingWarranties: What about `SendBlk` instead of `BlockID` or `Send1Blk`? I don't think `Block` makes it clear that not only are we fetching the block, since we are also accepting it into the consensus set?

modules/consensus/synchronize.go
VoidingWarranties: But all of those things are done. I want to make it very clear that the RPC to fetch the corresponding block to a block header will do all of those things.

modules/consensus/synchronize.go
VoidingWarranties: This code is just copy-pasted logic from the similar change we made to broadcasting `RelayTransactionSet` vs `RelayTransaction` which is already tested. If we're going to be removing this line in the future (once enough peers have upgraded) I'd rather not spend more time testing it.

modules/consensus/synchronize.go
VoidingWarranties: It becomes problematic if cyclical RPC calls are possible. In fact this chain of RPC calls is cyclical.

Also it prevents us from setting reasonable timeouts. If RPC "a" sets a timeout of 5 minutes, and RPC "b" calls RPC "a", there is no way for RPC "b" to know that it should extend its timeout by 5 minutes.

So I'll keep them called in goroutines for now and add logging later.

VoidingWarranties:
Not ready to merge yet, adding more integration tests.

VoidingWarranties:
Spun up a test node and it's working well. Should be good to merge.

DavidVorick:
I remember estimating that this would be a simple PR. and initially, it was. I'm glad we took the time to make it quality though.

DavidVorick:
As of this PR, you are going to be primarily in charge of the peer network, the gateway code, and the synchronization process. You've had to do a deep enough dive to figure all this out that you probably are more familiar with the code than either Luke or myself.

Congrats :)

PR #994, opened 2016-02-25
Broadcast to 0 peers on nil peers arg

VoidingWarranties:
It was too easy to write bugs where the gateway to broadcast to all peers if no peers matched the filtering criteria. For example the following would broadcast to all peers if there were no peers >=v0.4.7
```go
var v047AndAbove []modules.Peer
for _, p := range tp.gateway.Peers() {
        if build.VersionCmp(p.Version, "0.4.7") >= 0 {
                v047AndAbove = append(v047AndAbove, p)
        }
}
go tp.gateway.Broadcast("RelayTransactionSet", ts, v047AndAbove)
```

Also add the connection address to the log output when an RPC fails

### DIFF
--- a/modules/gateway/nodes.go
+++ b/modules/gateway/nodes.go
@@ -92,14 +92,25 @@ func (g *Gateway) relayNode(conn modules.PeerConn) error {
 		return err
 	}
 	// add node
-	id := g.mu.Lock()
-	defer g.mu.Unlock(id)
-	if err := g.addNode(addr); err != nil {
+	err := func() error {
+		// We wrap this logic in an anonymous function so we can defer Unlock to
+		// avoid managing locks across branching.
+		id := g.mu.Lock()
+		defer g.mu.Unlock(id)
+		if err := g.addNode(addr); err != nil {
+			return err
+		}
+		if err := g.save(); err != nil {
+			return err
+		}
+		return nil
+	}()
+	if err != nil {
 		return err
 	}
-	g.save()
 	// relay
-	go g.Broadcast("RelayNode", addr, nil)
+	peers := g.Peers()
+	go g.Broadcast("RelayNode", addr, peers)
 	return nil
 }
 
--- a/modules/gateway/rpc.go
+++ b/modules/gateway/rpc.go
@@ -122,7 +122,7 @@ func (g *Gateway) threadedHandleConn(conn modules.PeerConn) {
 		err = nil
 	}
 	if err != nil {
-		g.log.Printf("WARN: incoming RPC \"%v\" failed: %v", id, err)
+		g.log.Printf("WARN: incoming RPC \"%v\" from conn %v failed: %v", id, conn.RemoteAddr(), err)
 	}
 }
 
--- a/modules/gateway/rpc.go
+++ b/modules/gateway/rpc.go
@@ -126,15 +126,11 @@ func (g *Gateway) threadedHandleConn(conn modules.PeerConn) {
 	}
 }
 
-// Broadcast calls an RPC on all of the specified peers. If peers is nil, the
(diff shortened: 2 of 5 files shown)


modules/gateway/nodes.go
```diff
@@ -93,13 +93,15 @@ func (g *Gateway) relayNode(conn modules.PeerConn) error {
 	}
 	// add node
 	id := g.mu.Lock()
-	defer g.mu.Unlock(id)
 	if err := g.addNode(addr); err != nil {
+		g.mu.Unlock(id)
 		return err
 	}
 	g.save()
```
DavidVorick: this function doesn't return an error?

modules/gateway/rpc.go
```diff
@@ -122,19 +122,19 @@ func (g *Gateway) threadedHandleConn(conn modules.PeerConn) {
...
+// parallel. Broadcasts are restricted to "one-way" RPCs, which simply write an
+// object and disconnect. This is why Broadcast takes an interface{} instead of
+// an RPCFunc.
 func (g *Gateway) Broadcast(name string, obj interface{}, peers []modules.Peer) {
-	if peers == nil {
-		peers = g.Peers()
-	}
+	// NOTE: do not set peers to g.Peers() if peers is nil. It used to be a
+	// "feature" that peers would be set to all peers if nil, but this caused bugs
+	// that were difficult to find and even more difficult to test.
```
DavidVorick: This comment isn't really needed.

modules/gateway/rpc_test.go
```diff
@@ -289,4 +271,15 @@ func TestBroadcast(t *testing.T) {
...
+
+	// Test that calling broadcast with nil peers does not broadcast to g2 or g3.
+	g1.Broadcast("Recv", "foo", nil)
+	select {
+	case <-g2DoneChan:
+		t.Error("broadcast broadcasted to peers not in the peers arg")
+	case <-g3DoneChan:
+		t.Error("broadcast broadcasted to peers not in the peers arg")
+	case <-time.After(10 * time.Millisecond):
+		// Neither peer should receive a broadcast.
```
DavidVorick: this test covers the NOTE you wrote above, which further enforces the idea that the note isn't really needed.

modules/gateway/nodes.go
```diff
@@ -93,13 +93,15 @@ func (g *Gateway) relayNode(conn modules.PeerConn) error {
...
+	g.mu.Unlock(id)
```
DavidVorick: The locking here is concerning. Try:

```go
id := g.mu.Lock()
err := g.addNode(addr)
// err2 := g.save()?
g.save()
g.mu.Unlock(id)
if err != nil {
}
// if err2 != nil {}
```

DavidVorick: if you can't do that, then you should really wrap this in a func, which Luke is not a fan of but I very much dislike seeing locking happing across branches.

```go
err := func() error {
    id := g.mu.Lock()
    if err := g.addNode(addr); err != nil {
        return error
    }
   err = g.save()
   if err != nil {
        return err
    }
    return nil
}()
if err != nil {
    return err
}
```

But, really I think this can be avoided just be rethinking how the error handling should be happening here.

VoidingWarranties: Shouldn't we avoid calling `g.save()` if there is an error adding a peer?

DavidVorick: I don't see the harm, other than performance. Luke and I were discussing reducing the total save volume by as much as possible to help with performance, Sia currently goes to disk all the time.

VoidingWarranties: alternatively we could make `save` `managedSave`

VoidingWarranties:
Done

PR #995, opened 2016-02-27
Long-form Renter contracts

lukechampine:
This is the first major step towards the new Renter architecture, in which the Renter forms long-duration, large-filesize contracts and uses them to store multiple files. This should greatly improve the scalability issues observed with the current scheme (in which a separate contract is made for each file).

Though the change sounds simple in theory, it has far-reaching implications:
- Contracts can no longer be exactly sized
    * The user needs to be able to set an "allowance" that governs how much to spend on contracts
        * Changing the allowance should result in new contracts, but only in certain cases
    * We must gracefully handle situations where the user attempts to upload more data than their allowance can support
- We must be stricter about who we form contracts with, since the stakes are higher
- Contracts must be fully modifiable, supporting append, edit, and delete operations
    * We must store the Merkle tree of each contract; otherwise we'd have to re-hash the data of every file covered by a contract in order to revise
        * To keep Merkle trees small, we must store only the "4 MB hashes"
        * All files must use a 4 MB piece size
    * New protocols must be created for contract negotiation

Here's a high-level breakdown of what I've completed thus far:
(note: not all of those marked "complete" are covered in this PR! Sorry for the confusion.)
- [x] move hostdb code to new "contractor" package
- [x] remove concurrency, expiration, and offline checking from renter
- [x] modify contractor to use long-form contracts
	* simplest possible; uses hard-coded values
- [x] move HostPool to Renter
- [x] implement set-once allowance
- [x] add allowance API endpoints
- [x] add metrics for how much allowance has been used
- [ ] implement auto-renewing
	- [x] renew unchanged contracts
	- [ ] renew changed contracts (with new allowance)
	- [ ] form new contracts if renewing fails
- [ ] implement modifiable allowance
	- [x] Update on next renew cycle
	- [ ] Update immediately
- [x] design and implement new negotiation protocols
- [ ] design and implement new .sia format
- [ ] achieve satisfactory test coverage
- [ ] restore concurrency and expiration
	- [x] restore multi-piece concurrency
	- [ ] restore multi-chunk concurrency
- [x] delete file data after expiration
	* (partial; protocol not finished)
- [ ] implement data migration (for when too much redundancy is lost)
	- [ ] check for hosts going offline
	- [ ] check for hosts holding data hostage
	- [ ] check for hosts becoming too expensive
	* Note: contractor can't recover file data, only renter can
- [ ] add persistence to hostdb
- [ ] contractor needs to scan blockchain to confirm that contract txns and revision txns are submitted
	* needs to submit them itself if not submitted by host
	* can't make use of file contracts until they are seen in the blockchain

Unanswered questions:
- Should `contractor.Contracts()` return all of the contracts, or just the most recent contract formed with each host? If not the latter, how do we avoid revising outdated contracts?
- What do we do if we can't form contracts with enough hosts?


modules/renter/contractor/contractor.go
```diff
@@ -0,0 +1,154 @@
...
+	errNilCS           = errors.New("cannot create contractor with nil consensus set")
+	errNilWallet       = errors.New("cannot create contractor with nil wallet")
+	errNilTpool        = errors.New("cannot create contractor with nil transaction pool")
+	errUnknownContract = errors.New("no record of that contract")
+)
+
+// a Contract includes the original contract made with a host, along with
+// the most recent revision.
+type Contract struct {
+	IP              modules.NetAddress
```
DavidVorick: You should identify hosts by their public keys, not by their ip addresses, because the ip addresses can change and we plan to add support for dynmaic ips, hosts will update renters through either the blockchain or through some type of peer protocol.

modules/renter/contractor/contractor.go
```diff
@@ -0,0 +1,154 @@
...
+	if old.Hosts == 0 {
+		return c.formContracts(a)
+	}
+
+	// Otherwise, if the new allowance is "significantly different" (to be
+	// defined more precisely later), form intermediary contracts.
+	if a.Funds.Cmp(old.Funds) > 0 {
+		// not yet implemented
+		// c.formContracts(diff(a, old))
+	}
```
DavidVorick: should drop a TODO here so it's easy to find later once you're in 'go back and finish things' mode

modules/renter/contractor/contractor.go
```diff
@@ -0,0 +1,154 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Create Contractor using production dependencies.
+	return newContractor(cs, &walletBridge{w: wallet}, tpool, hdb, stdDialer{}, newPersist(persistDir), logger)
+}
+
+// newContractor creates a Contractor using the provided dependencies.
+func newContractor(cs consensusSet, w wallet, tp transactionPool, hdb hostDB, d dialer, p persister, l logger) (*Contractor, error) {
```
DavidVorick: I do think I prefer the option of using a single dependencies interface.

DavidVorick: non-blocking issue, we're still experimenting with the best way to do dependencies

modules/renter/contractor/contractor_test.go
```diff
@@ -0,0 +1,195 @@
...
+	_, err = New(stub, nil, stub, stub, dir)
+	if err != errNilWallet {
+		t.Fatalf("expected %v, got %v", errNilWallet, err)
+	}
+
+	// Nil transaction pool.
+	_, err = New(stub, stub, nil, stub, dir)
+	if err != errNilTpool {
+		t.Fatalf("expected %v, got %v", errNilTpool, err)
+	}
```
DavidVorick: did not test nil hostdb

modules/renter/contractor/contractor_test.go
```diff
@@ -0,0 +1,195 @@
...
+			{2}: Contract{ID: types.FileContractID{2}, IP: "bar"},
+			{3}: Contract{ID: types.FileContractID{3}, IP: "baz"},
+		},
+	}
+	for _, contract := range c.Contracts() {
+		if exp := c.contracts[contract.ID]; exp.IP != contract.IP {
+			t.Errorf("contract does not match: expected %v, got %v", exp.IP, contract.IP)
+		}
+	}
+}
```
DavidVorick: standard concerns of testing the implementation and not the behavior. I think it's best to explore a function like `Contracts()` in a smoke test, where you're creating a contractor and then trying out all of the behaviors together (with production dependencies). Then you know you were creating contracts the natural way, instead of hardcoding them in where they belonged.

modules/renter/contractor/dependencies.go
```diff
@@ -0,0 +1,106 @@
...
+	logger interface {
+		Println(...interface{})
+	}
+)
+
+// because wallet is not directly compatible with modules.Wallet (wrong
+// type signature for StartTransaction), we must provide a bridge type.
+type walletBridge struct {
+	w walletShim
+}
```
DavidVorick: I still find the bridge and shim stuff confusing. Maybe there is a better way? Though I don't have many ideas. again, non-blocking

modules/renter/contractor/editor.go
```diff
@@ -0,0 +1,209 @@
...
+	"github.com/NebulousLabs/Sia/crypto"
+	"github.com/NebulousLabs/Sia/encoding"
+	"github.com/NebulousLabs/Sia/modules"
+	"github.com/NebulousLabs/Sia/types"
+)
+
+const (
+	// SectorSize is the number of bytes in a sector.
+	SectorSize = 1 << 22 // 4 MiB
+)
```
DavidVorick: needs to be moved to the modules package. Though, it'll screw up my code if you do immediately, so instead you should drop a TODO here. I believe that my settings are 

4kb for testing
1MiB for dev
4Mib for release

modules/renter/contractor/editor.go
```diff
@@ -0,0 +1,209 @@
...
+
+// An Editor modifies a Contract by communicating with a host.
+type Editor interface {
```
DavidVorick: not very helpful comment, need to explain more, especially within the context of the renter and contractor

modules/renter/contractor/editor.go
```diff
@@ -0,0 +1,209 @@
...
+
+	// ContractID returns the FileContractID of the contract.
+	ContractID() types.FileContractID
+
+	// EndHeight returns the height at which the contract ends.
+	EndHeight() types.BlockHeight
+
+	// Close terminates the connection to the uploader.
+	Close() error
+}
```
DavidVorick: per our irl conversation, some of the function definitions should change here

modules/renter.go
```diff
@@ -67,9 +67,20 @@ type DownloadInfo struct {
 	StartTime   time.Time `json:"starttime"`
 }
 
+// An Allowance dictates how much the Renter is allowed to spend in a given
+// period.
+type Allowance struct {
+	Funds  types.Currency    `json:"funds"`
+	Hosts  int               `json:"hosts"`
```
DavidVorick: I think it makes sense to use a uint64 here, because our standard is to use uint64 anywhere that there is an unsigned number

modules/renter.go
```diff
@@ -67,9 +67,20 @@ type DownloadInfo struct {
...
+	Period types.BlockHeight `json:"period"`
+}
```
DavidVorick: Might want to add an extra parameter - `RefreshWindow`, which will dictate how far in advance the renter will call the 'renew' function. So if Period is 12 months and 'refresh window' is 6 weeks, it'll call renew after 10 months and 2 weeks (so there are 6 weeks or less remaining)

modules/renter/contractor/renew.go
```diff
@@ -0,0 +1,127 @@
...
+		return types.FileContractID{}, errTooExpensive
+	}
+
+	// get an address to use for negotiation
+	c.mu.Lock()
+	if c.cachedAddress == (types.UnlockHash{}) {
+		uc, err := c.wallet.NextAddress()
+		if err != nil {
+			c.mu.Unlock()
+			return types.FileContractID{}, err
```
DavidVorick: would much rather see this in an anon func, given that you return while holding a lock. The ability to use `defer Unlock` removes some of the stress associated with locking

modules/renter/contractor/renew.go
```diff
@@ -0,0 +1,127 @@
...
+			go func() {
+				defer wg.Done()
+				_, err := c.managedRenew(contract.ID, newHeight)
+				if err != nil {
+					c.log.Println("WARN: failed to renew contract", contract.ID, ":", err)
+				}
+			}()
+		}
+	}
+	c.mu.RUnlock()
```
DavidVorick: so, if I'm understanding correctly here, all the `managedRenew` calls are going to block before this unlock is called, and then afterwards Wait() is called to let all of the renewals take place. That should be in a comment somewhere.

I'm also not sure that it makes sense to lump all of the renewals into a single call? It seems like you could end up with multiple calls to `threadedRenewContracts` open at the same time, which could cause problems.

modules/renter/contractor/revise.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/NebulousLabs/Sia/crypto"
+	"github.com/NebulousLabs/Sia/encoding"
+	"github.com/NebulousLabs/Sia/modules"
+	"github.com/NebulousLabs/Sia/types"
+)
+
+// negotiateRevision sends the revision and new piece data to the host.
+func negotiateRevision(conn net.Conn, rev types.FileContractRevision, piece []byte, secretKey crypto.SecretKey) (types.Transaction, error) {
+	conn.SetDeadline(time.Now().Add(5 * time.Minute)) // sufficient to transfer 4 MB over 100 kbps
+	defer conn.SetDeadline(time.Time{})               // reset timeout after each revision
```
DavidVorick: should not ever be setting the timeout to infinity

modules/renter/contractor/update.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+// renewThreshold calculates an appropriate renewThreshold for a given period.
+func renewThreshold(period types.BlockHeight) types.BlockHeight {
+	threshold := period / 4
+	if threshold > 1000 {
+		threshold = 1000
+	}
+	return threshold
+}
```
DavidVorick: as suggested above, this should probably be set by the user

DavidVorick: also worth noting, because of the new support for virtual sectors, there's not a huge cost to renewing at half or even 75%, the cost is in the number of transactions that hit the blockchain, and not because you're going to be double-paying. I'd recommend at this point that the threshold default to 6 weeks, and that the period default to 12 weeks.

modules/renter/contractor/update.go
```diff
@@ -0,0 +1,33 @@
...
+// ProcessConsensusChange will be called by the consensus set every time there
+// is a change in the blockchain. Updates will always be called in order.
+func (c *Contractor) ProcessConsensusChange(cc modules.ConsensusChange) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	if c.blockHeight != 0 || cc.AppliedBlocks[len(cc.AppliedBlocks)-1].ID() != types.GenesisBlock.ID() {
+		c.blockHeight += types.BlockHeight(len(cc.AppliedBlocks))
+		c.blockHeight -= types.BlockHeight(len(cc.RevertedBlocks))
+	}
```
DavidVorick: I think that this logic is subtly wrong, but I forget why. I've started using:

```go
        for _, block := range cc.RevertedBlocks {
			if block.ID() != types.GenesisBlock.ID() {
				h.blockHeight--
			}
        }
        for _, block := range cc.RevertedBlocks {
			if block.ID() != types.GenesisBlock.ID() {
				h.blockHeight++
			}
        }
```

modules/renter/contractor/update.go
```diff
@@ -0,0 +1,33 @@
...
+
+	// renew/replace contracts
+	if c.blockHeight+renewThreshold(c.allowance.Period) >= c.renewHeight {
+		c.renewHeight += c.allowance.Period
+		go c.threadedRenewContracts(c.allowance, c.renewHeight)
+	}
```
DavidVorick: this looks good to me, except that I'm concerned you will end up with multiple calls to `threadedRenewContracts` in the pipeline.

modules/renter/pool.go
```diff
@@ -0,0 +1,101 @@
+package renter
+
+import (
+	"github.com/NebulousLabs/Sia/modules"
+	"github.com/NebulousLabs/Sia/modules/renter/contractor"
+)
+
+// A hostPool is a collection of hosts that can store data. New hosts are
+// drawn from a HostDB, and contracts are negotiated with them on demand.
+type hostPool struct {
```
DavidVorick: seems like the way you are using the host pool vs. the way it is documented have diverged. As the architecture has evolved, the host pool now serves a different purpose

modules/renter/upload_test.go
```diff
@@ -7,47 +7,37 @@ import (
...
+type uploadContractor struct {
+	stubContractor
+}
 
-// NewPool returns a new mock HostPool. Since uploadHostDB implements the
-// HostPool interface, it can simply return itself.
-func (hdb uploadHostDB) NewPool(uint64, types.BlockHeight) (hostdb.HostPool, error) {
-	return hdb, nil
+func (uploadContractor) Contracts() []contractor.Contract {
+	return make([]contractor.Contract, 24) // exact number shouldn't matter, as long as its large enough
```
DavidVorick: I am in favor of using a larger number than the max number of hosts typically chosen. Especially as we move down the line, having 30 hosts instead of 24 means that the renter can start making decisions like "choose 3 really cheap but unreliable hosts instead of 1 good but really expensive host". I realized it's actually the wrong strategy to pick a redundancy, instead you should pick a reliability (such as eleven 9's) and let the renter figure out the cheapest/best combination of hosts to hit that reliability. Which means you may need more than 6x redundancy, or may mean you only need 2 redundancy, just depends on how reliable vs. expensive the hosts are.

modules/renter.go
lukechampine: yeah. Currently I use the `renewThreshold` helper function for that, but it's better to be explicit.

modules/renter/contractor/renew.go
lukechampine: true, doing it serially makes things simpler. I figured that we wouldn't want to be blocked by unresponsive hosts, but we're talking about ~24 hosts * 30 second timeout; 12 minute worst-case scenario, which isn't a huge deal.

modules/renter/contractor/update.go
lukechampine: that's why `c.renewHeight` is incremented immediately. It should limit `threadedRenewContracts` to one call per renew cycle. So, only a problem when processing thousands of blocks at a time, which should be prevented by persistence.

modules/renter/contractor/contractor.go
lukechampine: I'm in favor of keeping interfaces small and separating concerns. I don't like that your `dependencies` struct has a method that returns a `*persist.Logger`. I would rather have the logger be its own dependency with its own interface, to reduce coupling. What do you do if you want to monitor the log output? If your logger is an interface, it's easy to write directly into a buffer. If it's a `persist.Logger`, not so much. You need to be aware of the implementation details of the concrete type. And now you still have a `log` field in your host!
It makes sense to group some functionality together, like the `os` functions. But glomming everything together is just bizarre to me.

modules/renter/contractor/dependencies.go
lukechampine: could be removed if I was ok mocking the full `TransactionBuilder` interface...which might not be so bad.

modules/renter/contractor/update.go
lukechampine: maybe we should add a `ConsensusChange.UpdateBlockheight(types.BlockHeight)` method? Annoying to reimplement the same (and apparently non-trivial) logic for every subscriber.

DavidVorick: would be even nicer if we had just decided that the genesis block was height 1 :p

I might make sense, except some modules do stuff between updating the height, and some modules don't. We could probably only standardize around the modules that don't change things every block, but I don't think that's very many of them.

VoidingWarranties: Shouldn't the GenesisBlock never be in reverted or applied blocks?

DavidVorick: nope. It gets sent as the first block to be applied, so that the modules can integrate all the siafund outputs that it creates.

VoidingWarranties: The explorer uses the loops
```go
for _, block := range cc.RevertedBlocks {
	h.blockHeight--
	//...
}

for _, block := range cc.RevertedBlocks {
	h.blockHeight++
	//...
}
```
So then is the height displayed by the explorer wrong?

DavidVorick: uh, yeah. That'll cause it to be off by one in the average situation.

DavidVorick: oh, the explorer probably uses a different trick. The explorer probably sets the initial height to -1 (underflowing it because it's unsigned), then lets the genesis block increment to 0. That works, but then you have an intentional underflow in your code and you need to do the underflow at initialization. I didn't think that was good code practice, so I phased it out in favor of the loops.

PR #997, opened 2016-02-27
Object signing helpers

lukechampine:
Add two functions to the crypto package: `WriteSignedObject` and `ReadSignedObject`. They function identicially to `encoding.WriteObject` and `encoding.ReadObject`, except that `WriteSignedObject` also writes a signature of the encoded object's hash, and `ReadSignedObject` reads and verifies that hash. These functions will be useful in the new host/renter protocol, which requires reading and writing signed data. Both have 100% code coverage.
(Question: could we sign the object directly, instead of its hash?)

There is one quirk to the implementation, which is that the signature is not included in the length prefix. Most objects sent over a network carry a length-prefix, so that the receiver knows how much data to expect. Since the signature size is constant, we can remove its length prefix, saving 8 bytes.

Also add proper testing of the `PublicKey` method.

### DIFF
--- a/crypto/rand_test.go
+++ b/crypto/rand_test.go
@@ -25,3 +25,32 @@ func TestRandIntnPanics(t *testing.T) {
 		t.Error("expected panic for n <= 0")
 	}
 }
+
+// TestPerm tests the Perm function.
+func TestPerm(t *testing.T) {
+	chars := "abcde" // string to be permuted
+	createPerm := func() string {
+		perm, err := Perm(len(chars))
+		if err != nil {
+			t.Fatal(err)
+		}
+		s := make([]byte, len(chars))
+		for i, j := range perm {
+			s[i] = chars[j]
+		}
+		return string(s)
+	}
+
+	// create (factorial(len(chars)) * 100) permutations
+	permCount := make(map[string]int)
+	for i := 0; i < 12000; i++ {
+		permCount[createPerm()]++
+	}
+
+	// we should have seen each permutation approx. 100 times
+	for p, n := range permCount {
+		if n < 50 || n > 150 {
+			t.Errorf("saw permutation %v times: %v", n, p)
+		}
+	}
+}
--- a/crypto/signatures.go
+++ b/crypto/signatures.go
@@ -82,10 +82,14 @@ func WriteSignedObject(w io.Writer, obj interface{}, sk SecretKey) error {
 // ReadSignedObject reads a length-prefixed object followed by its signature,
 // and verifies the signature.
 func ReadSignedObject(r io.Reader, obj interface{}, maxLen uint64, pk PublicKey) error {
-	// read the encoded object and signature
-	var encObj []byte
+	// read the encoded object
+	encObj, err := encoding.ReadPrefix(r, maxLen)
+	if err != nil {
+		return err
+	}
+	// read the signature
 	var sig Signature
-	err := encoding.NewDecoder(r).DecodeAll(&encObj, &sig)
+	err = encoding.NewDecoder(r).Decode(&sig)
 	if err != nil {
 		return err
 	}
--- a/crypto/signatures.go
+++ b/crypto/signatures.go
@@ -2,7 +2,9 @@ package crypto
 
 import (
 	"errors"
+	"io"
 
+	"github.com/NebulousLabs/Sia/encoding"
 	"github.com/NebulousLabs/ed25519"
 )
 
@@ -67,6 +69,33 @@ func VerifyHash(data Hash, pk PublicKey, sig Signature) error {
 		return errInvalidSignature
 	}
 	return nil
+}
+
+// WriteSignedData writes a length-prefixed object followed by its signature.
+func WriteSignedData(w io.Writer, obj interface{}, sk SecretKey) error {
+	b := encoding.Marshal(obj)
+	sig, _ := SignHash(HashBytes(b), sk) // no error possible
+	return encoding.NewEncoder(w).EncodeAll(b, sig)
+}
+
+// ReadSignedData reads a length-prefixed object followed by its signature,
+// and verifies the signature.
+func ReadSignedData(r io.Reader, obj interface{}, maxLen uint64, pk PublicKey) error {
+	// read the encoded data and signature
+	data, err := encoding.ReadPrefix(r, maxLen)
+	if err != nil {
+		return err
+	}
+	var sig Signature
+	if err := encoding.NewDecoder(r).Decode(&sig); err != nil {
+		return err
+	}
+	// verify the signature
+	if err := VerifyHash(HashBytes(data), pk, sig); err != nil {
+		return err
+	}
+	// decode the object
+	return encoding.Unmarshal(data, obj)
 }
 
 // PublicKey returns the public key that corresponds to a secret key.
--- a/crypto/signatures.go
+++ b/crypto/signatures.go
@@ -1,7 +1,9 @@
 package crypto
 
 import (
+	"bytes"
 	"errors"
+	"io"
 
 	"github.com/NebulousLabs/Sia/encoding"
 	"github.com/NebulousLabs/ed25519"
@@ -70,27 +72,43 @@ func VerifyHash(data Hash, pk PublicKey, sig Signature) error {
 	return nil
 }
 
+// WriteSignedObject writes a length-prefixed object followed by its signature.
+func WriteSignedObject(w io.Writer, obj interface{}, sk SecretKey) error {
+	encObj := encoding.Marshal(obj)
+	sig, _ := SignHash(HashBytes(encObj), sk) // no error possible
+	return encoding.NewEncoder(w).EncodeAll(encObj, sig)
+}
+
+// ReadSignedObject reads a length-prefixed object followed by its signature,
+// and verifies the signature.
+func ReadSignedObject(r io.Reader, obj interface{}, maxLen uint64, pk PublicKey) error {
+	// read the encoded object and signature
+	var encObj []byte
+	var sig Signature
+	err := encoding.NewDecoder(r).DecodeAll(&encObj, &sig)
+	if err != nil {
+		return err
+	}
+	// verify the signature
+	if err := VerifyHash(HashBytes(encObj), pk, sig); err != nil {
+		return err
+	}
+	// decode the object
+	return encoding.Unmarshal(encObj, obj)
+}
+
 // SignObject encodes an object and its signature.
 func SignObject(obj interface{}, sk SecretKey) []byte {
-	b := encoding.Marshal(obj)
-	sig, _ := SignHash(HashBytes(b), sk) // no error possible
-	return encoding.MarshalAll(b, sig)
+	b := new(bytes.Buffer)
+	WriteSignedObject(b, obj, sk) // no error possible with bytes.Buffer
+	return b.Bytes()
 }
 
 // VerifyObject decodes an object and verifies its signature.
 func VerifyObject(data []byte, obj interface{}, pk PublicKey) error {
-	// split data into encoded object and signature
-	var objData []byte
-	var sig Signature
-	if err := encoding.UnmarshalAll(data, &objData, &sig); err != nil {
-		return err
-	}
-	// verify the signature
-	if err := VerifyHash(HashBytes(objData), pk, sig); err != nil {
-		return err
-	}
-	// decode the object
-	return encoding.Unmarshal(objData, obj)
+	r := bytes.NewReader(data)
+	// since object is already in memory, no need to enforce a maxLen
+	return ReadSignedObject(r, obj, ^uint64(0), pk)
 }
 
 // PublicKey returns the public key that corresponds to a secret key.
--- a/crypto/signatures.go
+++ b/crypto/signatures.go
@@ -2,7 +2,6 @@ package crypto
 
 import (
 	"errors"
-	"io"
 
 	"github.com/NebulousLabs/Sia/encoding"
 	"github.com/NebulousLabs/ed25519"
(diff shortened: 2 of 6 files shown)


crypto/signatures.go
```diff
@@ -69,8 +71,35 @@ func VerifyHash(data Hash, pk PublicKey, sig Signature) error {
 	return nil
 }
 
+// WriteSignedData writes a length-prefixed object followed by its signature.
+func WriteSignedData(w io.Writer, obj interface{}, sk SecretKey) error {
+	b := encoding.Marshal(obj)
+	sig, _ := SignHash(HashBytes(b), sk) // no error possible
+	return encoding.NewEncoder(w).EncodeAll(b, sig)
+}
```
DavidVorick: I find it confusing that this is a  writer

crypto/signatures.go
```diff
@@ -69,8 +71,35 @@ func VerifyHash(data Hash, pk PublicKey, sig Signature) error {
...
+	if err := encoding.NewDecoder(r).Decode(&sig); err != nil {
+		return err
+	}
+	// verify the signature
+	if err := VerifyHash(HashBytes(data), pk, sig); err != nil {
+		return err
+	}
+	// decode the object
+	return encoding.Unmarshal(data, obj)
+}
```
DavidVorick: similarly, I find it confusing that this is a reader.

crypto/signatures.go
DavidVorick: It doesn't stream things. It takes all of the objects into memory, encodes them in memory, hashes them in memory, and then writes them.

I think this function would make more sense and have utility as a non-reader that just returns a []byte. Then the function could be used in non-reader scenarios like in transaction validation.

crypto/signatures_test.go
```diff
@@ -276,3 +277,67 @@ func TestIntegrationSigKeyGeneration(t *testing.T) {
...
+// TestPublicKey tests the PublicKey method
+func TestPublicKey(t *testing.T) {
+	sk, pk, err := GenerateKeyPair()
+	if err != nil {
+		t.Fatal(err)
+	}
+	if sk.PublicKey() != pk {
+		t.Fatal("PublicKey does not match actual public key:", pk, sk.PublicKey())
+	}
+}
```
DavidVorick: this function really isn't good enough. Should test a bunch of keys, not just one, and maybe should even do verification using the PulicKey().

lukechampine: That would be testing the functionality of the public key, which doesn't really have anything to do with the function being tested.
All `PublicKey()` does is take a secret key and return its corresponding public key. As long as the result matches the previously-generated `pk`, `PublicKey()` has done its job.
Nothing wrong with generating 5 or 10 keys rather than just one, but since it's random and it gets run on every build, it doesn't seem much more thorough to me.

crypto/signatures.go
lukechampine: yeah, I think I agree. It would be technically possible to stream it -- you'd use a `MultiWriter` to write to both `w` and the hasher -- but the efficiency gain would be minimal unless the object is very large.

encoding/marshal.go
```diff
@@ -347,6 +348,19 @@ func Unmarshal(b []byte, v interface{}) error {
 	return NewDecoder(r).Decode(v)
 }
 
+// UnmarshalAll decodes the encoded values in b and stores them in vs, which
+// must be pointers.
+func UnmarshalAll(b []byte, vs ...interface{}) error {
```
lukechampine: this function wound up not being used, but if we have a `MashalAll` it makes sense to have an `UnmarshalAll` for parity.

crypto/signatures_test.go
DavidVorick: was thinking more like 500 or 5000 each build.

crypto/rand_test.go
```diff
@@ -25,3 +25,32 @@ func TestRandIntnPanics(t *testing.T) {
...
+	for i := 0; i < 12000; i++ {
+		permCount[createPerm()]++
+	}
+
+	// we should have seen each permutation approx. 100 times
+	for p, n := range permCount {
+		if n < 50 || n > 150 {
+			t.Errorf("saw permutation %v times: %v", n, p)
+		}
+	}
```
DavidVorick: :+1:

encoding/marshal_test.go
```diff
@@ -216,18 +216,58 @@ func TestEncodeDecode(t *testing.T) {
 	}
 }
 
+func TestEncodeAll(t *testing.T) {
+	var expected []byte
+	for i := range testStructs {
+		expected = append(expected, Marshal(testStructs[i])...)
+	}
```
DavidVorick: This test depends on the correctness of `Marshal`, I think would be better avoided, at least there should be 1 or 2 tests where you manually create structs and hardcode the expected output. It's worth having both, but I've found that it's worth having hardcoded tests.

Even something as simple as `EncodeAll(uint64(12), [][2]byte{{1,2},{3,4}}) => [hardcoded result]` would make me feel better.

encoding/marshal_test.go
```diff
@@ -216,18 +216,58 @@ func TestEncodeDecode(t *testing.T) {
...
+		t.Errorf("expected %v, got %v", expected, b.Bytes())
+	}
+}
+
+func TestDecodeAll(t *testing.T) {
+	b := new(bytes.Buffer)
+	enc := NewEncoder(b)
+	for i := range testStructs {
+		enc.Encode(testStructs[i])
+	}
```
DavidVorick: A similar thing here. Instead of (or rather in addition to) encoding something and then decoding it, I think it makes sense to start with some hardcoded bytes and some data and verify that it decodes correctly.

encoding/marshal_test.go
```diff
@@ -216,18 +216,58 @@ func TestEncodeDecode(t *testing.T) {
...
+
+	var emptyStructs = []interface{}{&test0{}, &test1{}, &test2{}, &test3{}, &test4{}, &test5{}, &test6{}}
+	err := NewDecoder(b).DecodeAll(emptyStructs...)
+	if err != nil {
+		t.Error(err)
+	}
```
DavidVorick: This check is not good enough, you definitely need to write some comparisons to make sure the data was populated correctly. At the very least you should be checking correctness for test0 and test6 (if it's getting things wrong, it's most likely to be getting them wrong on the edges), but really you should be checking all of them.

encoding/marshal_test.go
```diff
@@ -216,18 +216,58 @@ func TestEncodeDecode(t *testing.T) {
...
+	for i := range testStructs {
+		b = append(b, Marshal(testStructs[i])...)
+	}
+
+	var emptyStructs = []interface{}{&test0{}, &test1{}, &test2{}, &test3{}, &test4{}, &test5{}, &test6{}}
+	err := UnmarshalAll(b, emptyStructs...)
+	if err != nil {
+		t.Error(err)
+	}
+}
```
DavidVorick: Same comments applied to MarshalAll and UnmarshalAll.

encoding/marshal.go
DavidVorick: agreed, parity makes sense

crypto/signatures_test.go
```diff
@@ -276,3 +276,59 @@ func TestIntegrationSigKeyGeneration(t *testing.T) {
...
+	// not enough object data
+	err = VerifyObject(signedObj[:0], &read, pk)
+	if err == nil || err.Error() != "could not decode type []uint8: EOF" {
+		t.Fatal("expected decode error, got", err)
+	}
+	// not enough signature data
+	err = VerifyObject(signedObj[:19], &read, pk)
+	if err == nil || err.Error() != "could not decode type crypto.Signature: EOF" {
+		t.Fatal("expected decode error, got", err)
+	}
```
DavidVorick: I would feel better if we wrapped this whole function in a `for` loop, especially given that the sign and verify functions have failed nondeterministically for previous implementations (to the tune of 1-in-200 times or so).

```
var j int
if testing.Short {
    j = 3
} else {
    j = 200
}
for i := 0; i < j; i++ {
    // whole test
}
```

crypto/signatures.go
```diff
@@ -69,8 +72,47 @@ func VerifyHash(data Hash, pk PublicKey, sig Signature) error {
...
+func SignObject(obj interface{}, sk SecretKey) []byte {
+	b := new(bytes.Buffer)
+	WriteSignedObject(b, obj, sk) // no error possible with bytes.Buffer
+	return b.Bytes()
+}
+
+// VerifyObject decodes an object and verifies its signature.
+func VerifyObject(data []byte, obj interface{}, pk PublicKey) error {
+	r := bytes.NewReader(data)
+	// since object is already in memory, no need to enforce a maxLen
```
DavidVorick: :+1:

crypto/signatures.go
```diff
@@ -69,8 +72,47 @@ func VerifyHash(data Hash, pk PublicKey, sig Signature) error {
 	return nil
 }
 
+// WriteSignedObject writes a length-prefixed object followed by its signature.
+func WriteSignedObject(w io.Writer, obj interface{}, sk SecretKey) error {
+	encObj := encoding.Marshal(obj)
+	sig, _ := SignHash(HashBytes(encObj), sk) // no error possible
+	return encoding.NewEncoder(w).EncodeAll(encObj, sig)
```
DavidVorick: Is this the best way to do it? Also, I'm thinking it might cause problems. Intuitively, I would have expected WriteSignedObject to look like the object bytes followed by the signature bytes. But instead, we have a bytes slice which decodes to the object bytes followed by the signature bytes. You get an extra 8-byte prefix that I don't think needs to be there.

DavidVorick: Another thing that might make sense is to put the signature first. I think that's how nacl does it.

crypto/signatures.go
```diff
@@ -69,8 +72,47 @@ func VerifyHash(data Hash, pk PublicKey, sig Signature) error {
...
+	err := encoding.NewDecoder(r).DecodeAll(&encObj, &sig)
+	if err != nil {
+		return err
+	}
+	// verify the signature
+	if err := VerifyHash(HashBytes(encObj), pk, sig); err != nil {
+		return err
+	}
+	// decode the object
+	return encoding.Unmarshal(encObj, obj)
```
DavidVorick: I guess it doesn't really simplify decoding much, but still I worry that the other way would be confusing when writing protocols like negotiation. (I assumed sig was first, adjust accordingly if putting sig second)

```go
objAndSigBytes, err := ioutil.ReadAll(r)
objBytes := objAndSigBytes[SignatureSize:]
err = VerifyHash(objBytes), pk, objBytes)
return encoding.Unmarshal(objBytes, obj)
```

DavidVorick:
I was looking at some of the other tests in the crypto package, and there's some yuckyness. For example, we somehow ended up with a for loop testing 200 signatures when we're deterministically setting the entropy? Not sure how that got through =/

DavidVorick:
(though that's out of scope for this pr)

crypto/signatures.go
lukechampine: The length prefix is necessary to know the number of bytes in the encoded object. This is why `encoding.WriteObject` and `encoding.ReadObject` use a length prefix. Almost every object sent over the wire is sent in this format. Not only does this allow you to reject large objects without having to read them, it also has the more subtle benefit of making decoding optional. That is, you can read the raw bytes with `ReadPrefix`, or skip straight to decoding them with `ReadObject`. Without the length prefix, the only way to get the raw bytes would be to decode into the object and then re-marshal it. `VerifyObject` needs to hash those bytes in order to check the signature, so it would wind up doing gross extra work.

Safe networking code requires length prefixes anyway, so if you drop the length prefix, you get this:
```go
err := encoding.WriteObject(conn, foo)
err = encoding.WritePrefix(conn, crypto.SignObject(bar))
var obj Object
bytes, err := encoding.ReadPrefix(conn)
err = crypto.VerifyObject(bytes, &obj)
```
Instead of:
```go
err := encoding.WriteObject(conn, foo)
err = crypto.WriteSignedObject(conn, bar)
var obj Object
err = crypto.ReadSignedObject(conn, &obj)
```

I'm fine with putting the signature at the beginning, especially if that's what other libraries do.

crypto/signatures.go
lukechampine: You can't use `ReadAll` on a network connection. It will try to read until the connection is closed.

DavidVorick: oh. did not realize that. I guess it does make sense to create the encObj then, so you can ignore my other comment as well.

lukechampine:
don't merge yet -- found a bug

encoding/marshal_test.go
```diff
@@ -216,18 +216,58 @@ func TestEncodeDecode(t *testing.T) {
...
 	}
 
-	expected := MarshalAll(testStructs...)
+	b := MarshalAll(testStructs...)
 	if !bytes.Equal(b, expected) {
 		t.Errorf("expected %v, got %v", expected, b)
 	}
 }
 
+func TestUnmarshalAll(t *testing.T) {
```
DavidVorick: docstring please

encoding/marshal_test.go
```diff
@@ -216,18 +216,58 @@ func TestEncodeDecode(t *testing.T) {
...
+	var emptyStructs = []interface{}{&test0{}, &test1{}, &test2{}, &test3{}, &test4{}, &test5{}, &test6{}}
+	err := NewDecoder(b).DecodeAll(emptyStructs...)
+	if err != nil {
+		t.Error(err)
+	}
+}
+
 func TestMarshalAll(t *testing.T) {
-	var b []byte
+	var expected []byte
```
DavidVorick: docstring on this guy too

encoding/marshal_test.go
```diff
@@ -216,18 +216,58 @@ func TestEncodeDecode(t *testing.T) {
 	}
 }
 
+func TestEncodeAll(t *testing.T) {
```
DavidVorick: docstring

encoding/marshal_test.go
```diff
@@ -216,18 +216,58 @@ func TestEncodeDecode(t *testing.T) {
...
+	}
+
+	b := new(bytes.Buffer)
+	NewEncoder(b).EncodeAll(testStructs...)
+	if !bytes.Equal(b.Bytes(), expected) {
+		t.Errorf("expected %v, got %v", expected, b.Bytes())
+	}
+}
+
+func TestDecodeAll(t *testing.T) {
```
DavidVorick: docstring

encoding/marshal_test.go
lukechampine: hmm. Since `EncodeAll` is really just a sequence of encodes, maybe I should use `Encode` instead of marshal.
A hardcoded test wouldn't hurt as well.

PR #1002, opened 2016-03-01
Sync bug fixes

VoidingWarranties:
This PR fixes 2 major and 1 minor bugs.

Major bugs:
* SendBlocks was sending the entire blockchain every time (tested in `TestRPCSendBlockSendsOnlyNecessaryBlocks`)
* SendBlocks would sometimes fail to send the last block (tested in `TestRPCSendBlocks`)

Minor bug:
* blockHistory included the genesis block twice sometimes (when height == step). This doesn't cause any problems, but no reason to not fix it. Even with this fix the genesis block is still included twice (once at blockHistory[0] and once at blockHistory[31]) if the only block is the genesis block, which should only happen during initial blockchain download.

This should dramatically improve syncing performance.

### DIFF
--- a/modules/consensus/synchronize.go
+++ b/modules/consensus/synchronize.go
@@ -176,6 +176,7 @@ func (cs *ConsensusSet) sendBlocks(conn modules.PeerConn) error {
 			found = true
 			// Start from the child of the common block.
 			start = pb.Height + 1
+			break
 		}
 		return nil
 	})
--- a/modules/consensus/synchronize.go
+++ b/modules/consensus/synchronize.go
@@ -215,7 +215,7 @@ func (cs *ConsensusSet) sendBlocks(conn modules.PeerConn) error {
 				}
 				blocks = append(blocks, pb.Block)
 			}
-			moreAvailable = start+MaxCatchUpBlocks < height
+			moreAvailable = start+MaxCatchUpBlocks <= height
 			start += MaxCatchUpBlocks
 			return nil
 		})
--- a/modules/consensus/synchronize.go
+++ b/modules/consensus/synchronize.go
@@ -56,7 +56,7 @@ func blockHistory(tx *bolt.Tx) (blockIDs [32]types.BlockID) {
 		if i >= 9 {
 			step *= 2
 		}
-		if height < step {
+		if height <= step {
 			break
 		}
 		height -= step
--- a/modules/consensus/synchronize_test.go
+++ b/modules/consensus/synchronize_test.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/NebulousLabs/Sia/build"
+	"github.com/NebulousLabs/Sia/encoding"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/modules/gateway"
 	"github.com/NebulousLabs/Sia/types"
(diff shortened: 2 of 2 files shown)


lukechampine:
neat.
This might also reduce load on the explorer, so we'll want to update that binary as soon as this is merged.

VoidingWarranties:
Unfortunately the explorer won't benefit until its peers are also using this version of SemdBlocks.

modules/consensus/synchronize.go
```diff
@@ -176,6 +176,7 @@ func (cs *ConsensusSet) sendBlocks(conn modules.PeerConn) error {
 			found = true
 			// Start from the child of the common block.
 			start = pb.Height + 1
+			break
```
DavidVorick: T.T

DavidVorick: how many gigabytes do you think this bug has eaten? I'm going to guess 25,000.

modules/consensus/synchronize_test.go
```diff
@@ -311,3 +315,197 @@ func TestSendBlocksBroadcastsOnce(t *testing.T) {
...
+		0,
+		MaxCatchUpBlocks - 2,
+		0,
+		2*MaxCatchUpBlocks + 1,
+		0,
+		1,
+		2,
+		10,
+		11,
+		12,
```
DavidVorick:
```go
if i >= 9 {
 	step *= 2
 }
```

so, you should actually replace these last bits with some sort of loop that'll just do every number from 0 to 50. I don't know how long that test will run, but even at 60 seconds that's an okay test. You should at least do every number from 0 to 9, and then also do (12),15,16,17, and also (23), 31,32,33. That will give you much better coverage on the edge cases than what you've got so far.

modules/consensus/synchronize_test.go
```diff
@@ -311,3 +315,197 @@ func TestSendBlocksBroadcastsOnce(t *testing.T) {
...
+			if err != nil {
+				t.Fatal(err)
+			}
+			err = cst.cs.managedAcceptBlock(b)
+			if err != nil {
+				t.Fatal(err)
+			}
+		}
+
+		err = cs.gateway.RPC(cst.cs.gateway.Address(), "SendBlocks", cs.threadedReceiveBlocks)
```
DavidVorick: Would be greatly reassuring if you could drop a sanity check here to make sure that the consensus set hasn't actually caught up yet.

It's also worth me pointing out that when DEBUG mode is running, the cost of validating a block is linear in the total number of blocks (because the consensus set has some very aggressive sanity checks when DEBUG is enabled), so you might want to consider resetting the original consensus set every test as well. (otherwise you'll get to a height of 50+, which starts to slow testing down a lot).

VoidingWarranties: What is the best way to reset a consensus set?

modules/consensus/synchronize_test.go
```diff
@@ -311,3 +315,197 @@ func TestSendBlocksBroadcastsOnce(t *testing.T) {
...
+		if cs.CurrentBlock().ID() != cst.cs.CurrentBlock().ID() {
+			t.Errorf("current block ids not equal after call to SendBlocks when peer was ahead by %v blocks: expected %v, got %v", n, cst.cs.CurrentBlock().ID(), cs.CurrentBlock().ID())
+		}
+	}
+}
+
+// TestRPCSendBlockSendsOnlyNecessaryBlocks tests that the SendBlocks RPC only
+// sends blocks that the caller does not have and that are part of the longest
+// chain.
+func TestRPCSendBlockSendsOnlyNecessaryBlocks(t *testing.T) {
```
DavidVorick: ah man. Where was this test 7 months ago?

modules/consensus/synchronize_test.go
```diff
@@ -311,3 +315,197 @@ func TestSendBlocksBroadcastsOnce(t *testing.T) {
...
+		err = cs.gateway.RPC(cst.cs.gateway.Address(), "SendBlocks", cs.threadedReceiveBlocks)
+		if err != nil {
+			t.Error(err)
+		}
+		if cs.Height() != cst.cs.Height() {
+			t.Errorf("heights not equal after call to SendBlocks when peer was ahead by %v blocks: expected %v, got %v", n, cst.cs.Height(), cs.Height())
+		}
+		if cs.CurrentBlock().ID() != cst.cs.CurrentBlock().ID() {
+			t.Errorf("current block ids not equal after call to SendBlocks when peer was ahead by %v blocks: expected %v, got %v", n, cst.cs.CurrentBlock().ID(), cs.CurrentBlock().ID())
+		}
```
DavidVorick: Is there some sort of mocking you can add somewhere that will let us see how many times AcceptBlock was called, and then you can verify that it was the correct number of times? Basically, some way that we can roll the next test into this one, since this one has broader coverage of the various cases.

THOUGH, you don't test reorging cases here, and you probably should.

modules/consensus/synchronize_test.go
DavidVorick: I mentioned in a later comment, but you should also add some sort of test case here that explores reorging. How well does SendBlocks operate when one set is at height 35, and one is at height 50, but the common block is at 32? 30? 1? 0?

A whole new set of juicyness, but it's coverage that I would like to see.

modules/consensus/synchronize_test.go
DavidVorick: call `Close()` and make a new one using a different persist directory. So, not a super easy procedure.

modules/consensus/synchronize.go
VoidingWarranties: Let's assume an average block size of 100kB, with an average network size of 100 peers. There are ~40,000 blocks. Let's also assume constant block size and number of peers.

`100kB * 100 peers * sum of integers from 0 to 40,000 ~= 8PB`

Wow. Did I get my math wrong?

modules/consensus/synchronize_test.go
VoidingWarranties: Can you explain how the code block you mentioned affects this test?
Can you also explain the reason for specifically testing the numbers (12),15,16,17,(23),31,32,33?

modules/consensus/synchronize_test.go
VoidingWarranties: Currently the test only takes 2.2s, so definitely not necessary yet. With your other suggestions for adding more cases it might be necessary.

modules/consensus/synchronize_test.go
DavidVorick: the code block determines which blocks get sent to express similarity. it'll send the most recent block, then the next most recent block, following the pattern: 0,1,2,3,4,5,6,7,8,16,32,64, etc.

The numbers I picked all fall on those boundaries. 1-9 will cover the whole leading boundaries, and then 15,16,17 cover the 16 boundary (+/- 1), and (12) covers something in the middle somewhat arbitratily picked. (I did choose an even number for 12, and a prime number for 23, just because even numbers and prime numbers are often handled by different codepaths). I'm basically  trying to apply general fuzzing strategies to the numbers that are being chosen for testing.

modules/consensus/synchronize.go
DavidVorick: yes you did :)

Average peer has about 10 friends during IBD. Each peer does IBD probably, once, though total we've probably done 10,000 IBDs. Assuming that they're downloading 35,000 blocks on average, and they catch up from block 0, and I know that blocks don't average 100kb, probably closer to 10kb (especially the early blocks are mostly empty). I think the size of the whole blockchain is around 500MB.

500MB * 10 peers * 10,000 IBDs is 50,000GB. But, that's not where the bug comes in. The bug happens because when there's a new block you ask all 10 peers for their blocks again, and they send them to you from height 0. So assuming it takes 3 hours to IBD, that's 18 blocks. So you download the entire consensus set as much as 180 times. Though really, these connections fail a lot so you probably get cut off.

So, somewhere between 50TB and 500TB, though I'm going to go ahead and expand that number because we did a lot of assuming. So, between 10TB and 1PB.

modules/consensus/synchronize_test.go
VoidingWarranties: It would be nice to combine the two, but I don't see any easy way to do it.

PR #1008, opened 2016-03-06
Gateway testing bug fix

VoidingWarranties:
There was a bug in testing that prevented an RPC from calling another RPC. This was because a gateway would connect to a peer with the address "[::]:port" (the unspecified IPV6 address), but during the RPC the `conn` would have the remote address "[::1]:port" (the localhost IPV6 address). The fix is to use "localhost:0" instead of ":0".

This bug only affected tests and the only files changed in this PR are test files.

### DIFF
--- a/api/server_helpers_test.go
+++ b/api/server_helpers_test.go
@@ -46,7 +46,7 @@ type serverTester struct {
 // server tester, without creating any directories or mining any blocks.
 func assembleServerTester(key crypto.TwofishKey, testdir string) (*serverTester, error) {
 	// Create the modules.
-	g, err := gateway.New(":0", filepath.Join(testdir, modules.GatewayDir))
+	g, err := gateway.New("localhost:0", filepath.Join(testdir, modules.GatewayDir))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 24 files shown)


api/server_helpers_test.go
```diff
@@ -88,7 +88,7 @@ func assembleServerTester(key crypto.TwofishKey, testdir string) (*serverTester,
 	if err != nil {
 		return nil, err
 	}
-	srv, err := NewServer(":0", "Sia-Agent", cs, e, g, h, m, r, tp, w)
+	srv, err := NewServer("localhost:0", "Sia-Agent", cs, e, g, h, m, r, tp, w)
```
lukechampine: this is for the API listener, not the gateway listener. Shouldn't cause any problems though

DavidVorick:
So, this is not a production bug? Nodes in production are not affected?

DavidVorick:
PR looks fine to me. Just glanced at it, but it's all testing so it's not really a risky merge.

api/server_helpers_test.go
VoidingWarranties: I changed every instance of ":0" just in case. I would encourage always using "localhost:0" in the future as well.

VoidingWarranties:
Correct, this is not a bug in production.

VoidingWarranties:
#992 cannot be merged until this is merged.

PR #1011, opened 2016-03-07
Remove readlock from managedRPCSettings

### LINKED ISSUE #1009: managedRPCSettings() can deadlock
I've seen managedRPCSettings() deadlock with master (fda4e228); and probably also with the 0.5.1 release. (On linux). The problem is seems to be that managedRPCSettings() takes a read lock on h.mu, and then calls h.Settings() which in turn also takes a read lock. The deadlock happens if a second goroutine attempts to take a write lock during in the interval between the two read locks; the write lock and the second read lock both start waiting.

VoidingWarranties:
The readlock wasn't necessary as managedRPCSettings only called the exported method Settings() which has it's own readlock. The extra readlock made it possible for a deadlock to occur if a different goroutine tried to obtain a writelock after managedRPCSettings's readlock but before Settings() readlock.

fixes #1009
thanks @david60!

### DIFF
--- a/modules/host/network.go
+++ b/modules/host/network.go
@@ -154,7 +154,5 @@ func (h *Host) threadedListen() {
 
 // managedRPCSettings is an rpc that returns the host's settings.
 func (h *Host) managedRPCSettings(conn net.Conn) error {
-	h.mu.RLock()
-	defer h.mu.RUnlock()
 	return encoding.WriteObject(conn, h.Settings())
 }


DavidVorick:
Just to explain this in the context of our conventions, modules are set up such that you are supposed to assume that all exported calls are thread-safe. You should also assume that any exported call is going to grab a lock on the object. We make these assumptions to simplify the modules and reduce the cognitive overhead when working with concurrency. If you've got a set of clean assumptions that you can make based on the function name, then you don't have to worry about trying to remember which functions grab locks and which don't. You've got clear context telling you whether or not it's safe to grab a lock or call a function.

In this case, we are calling `h.Settings()`, which is an exported call, which means we should assume it'll grab a lock. Because of that, we don't need to grab a lock in the function. We should have been able to see the problem without knowing the implementation of `h.Settings()`, because of our locking conventions.

I am wondering if we should be calling `h.settings()` instead, it seems odd/wrong to me that we might be making a call to an exported function from a non-exported function. It is managed, which means it's okay, but nonetheless I think the problem here is more that we were calling an exported function rather than we grabbed a lock without need. 'managed' functions should be able to grab locks, so it's confusing that you would want to remove locks that it was grabbing.

VoidingWarranties:
There are a few places in consensus synchronization code where exported methods are called from non-methods, in particular in RPCs. I suppose that is a unique case as RPCs are effectively exported methods callable by other peers.

Are you suggesting we enforce that un-exported methods cannot call exported methods in the same package? I think that would make locking easier to reason about and help avoid bugs like this.

lukechampine:
>Are you suggesting we enforce that un-exported methods cannot call exported methods in the same package?

That was the original convention, before we realized we needed things like `threaded` and `managed`. It does make things much easier to reason about in general.

My opinion is that `managed` means "the programmer is going to manage locking manually here, and take extra care to make sure it's correct." So I don't see any problem calling an exported function from a `managed` function (although a comment pointing out the implicit locking might be prudent).

DavidVorick:
> My opinion is that managed means: "The programmer is going to manage locking manually here. Conventions may be broken. Take extra care to avoid deadlocks." So I don't see any problem calling an exported function from a managed function

I mostly agree, except that you should only do it if absolutely needed. `managed` functions should stick to conventions as much as possible, and only break where necessary.

PR #1013, opened 2016-03-10
IBD improvements

VoidingWarranties:
Not complete. Still needs manual timeouts (muxado doesn't timeout correctly nor does it return proper `net.Error` errors) and tests.

### DIFF
--- a/modules/consensus/consensusset.go
+++ b/modules/consensus/consensusset.go
@@ -9,6 +9,7 @@ package consensus
 import (
 	"errors"
 
+	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/encoding"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/persist"
@@ -111,12 +112,21 @@ func New(gateway modules.Gateway, persistDir string) (*ConsensusSet, error) {
 		return nil, err
 	}
 
-	// Register RPCs
-	gateway.RegisterRPC("SendBlocks", cs.rpcSendBlocks)
-	gateway.RegisterRPC("RelayBlock", cs.rpcRelayBlock)
-	gateway.RegisterRPC("RelayHeader", cs.rpcRelayHeader)
-	gateway.RegisterRPC("SendBlk", cs.rpcSendBlk)
-	gateway.RegisterConnectCall("SendBlocks", cs.threadedReceiveBlocks)
+	go func() {
+		// Sync with the network. Don't sync if we are testing because typically we
+		// don't have any mock peers to synchronize with in testing.
+		// TODO: figure out a better way to conditionally do IBD. Otherwise this block will never be tested.
+		if build.Release != "testing" {
+			cs.threadedInitialBlockchainDownload()
+		}
+
+		// Register RPCs
+		gateway.RegisterRPC("SendBlocks", cs.rpcSendBlocks)
+		gateway.RegisterRPC("RelayBlock", cs.rpcRelayBlock)
+		gateway.RegisterRPC("RelayHeader", cs.rpcRelayHeader)
+		gateway.RegisterRPC("SendBlk", cs.rpcSendBlk)
+		gateway.RegisterConnectCall("SendBlocks", cs.threadedReceiveBlocks)
+	}()
 
 	return cs, nil
 }
--- a/modules/consensus/consensusset.go
+++ b/modules/consensus/consensusset.go
@@ -126,6 +126,8 @@ func New(gateway modules.Gateway, persistDir string) (*ConsensusSet, error) {
 		gateway.RegisterRPC("RelayHeader", cs.rpcRelayHeader)
 		gateway.RegisterRPC("SendBlk", cs.rpcSendBlk)
 		gateway.RegisterConnectCall("SendBlocks", cs.threadedReceiveBlocks)
+
+		// TODO: change flag to indicate IBD done.
 	}()
 
 	return cs, nil
--- a/modules/consensus/synchronize.go
+++ b/modules/consensus/synchronize.go
@@ -1,6 +1,10 @@
 package consensus
 
 import (
+	"errors"
+	"net"
+	"time"
+
 	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/encoding"
@@ -16,16 +20,31 @@ var (
 	// download.
 	MaxCatchUpBlocks = func() types.BlockHeight {
 		switch build.Release {
-		case "testing":
-			return 3
-		case "standard":
-			return 10
 		case "dev":
 			return 10
+		case "standard":
+			return 10
+		case "testing":
+			return 3
 		default:
 			panic("unrecognized build.Release")
 		}
 	}()
+	// sendBlocksTimeout is the timeout for the SendBlocks RPC.
+	sendBlocksTimeout = func() time.Duration {
+		switch build.Release {
+		case "dev":
+			return 5 * time.Minute
+		case "standard":
+			return 5 * time.Minute
+		case "testing":
+			return 5 * time.Second
+		default:
+			panic("unrecognized build.Release")
+		}
+	}()
+
+	errSendBlocksStalled = errors.New("SendBlocks RPC timed and never received any blocks")
 )
 
 // blockHistory returns up to 32 block ids, starting with recent blocks and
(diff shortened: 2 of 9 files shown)


modules/consensus/synchronize.go
```diff
@@ -72,9 +97,21 @@ func blockHistory(tx *bolt.Tx) (blockIDs [32]types.BlockID) {
...
+	// prevent any one peer from stalling IBD.
+	err := conn.SetDeadline(time.Now().Add(sendBlocksTimeout))
+	// Pipes do not support Set{,Read,Write}Deadline and should only be used in
+	// testing.
+	if build.Release != "testing" {
+		if opErr, ok := err.(*net.OpError); ok && opErr.Op == "set" && opErr.Net == "pipe" {
+			return err
+		}
+	}
+
```
VoidingWarranties: Ignore this. It's being replaced anyways.

modules/consensus/synchronize_test.go
```diff
@@ -546,10 +546,6 @@ func TestRPCSendBlockSendsOnlyNecessaryBlocks(t *testing.T) {
 
 	// Create the "remote" peer.
 	cst, err := blankConsensusSetTester("TestRPCSendBlockSendsOnlyNecessaryBlocks - remote")
-	if err != nil {
-		t.Fatal(err)
-	}
-	defer cst.Close()
```
VoidingWarranties: Oops. Rebase gone wrong.

modules/consensus/consensusset.go
```diff
@@ -111,12 +112,21 @@ func New(gateway modules.Gateway, persistDir string) (*ConsensusSet, error) {
...
-	// Register RPCs
-	gateway.RegisterRPC("SendBlocks", cs.rpcSendBlocks)
-	gateway.RegisterRPC("RelayBlock", cs.rpcRelayBlock)
-	gateway.RegisterRPC("RelayHeader", cs.rpcRelayHeader)
-	gateway.RegisterRPC("SendBlk", cs.rpcSendBlk)
-	gateway.RegisterConnectCall("SendBlocks", cs.threadedReceiveBlocks)
+	go func() {
+		// Sync with the network. Don't sync if we are testing because typically we
+		// don't have any mock peers to synchronize with in testing.
+		// TODO: figure out a better way to conditionally do IBD. Otherwise this block will never be tested.
```
DavidVorick: one thing you could do is pass a flag in, but that could get pretty annoying.

modules/consensus/synchronize.go
```diff
@@ -10,22 +14,43 @@ import (
 	"github.com/NebulousLabs/bolt"
 )
 
+const (
+	// minNumOutbound is the minimum number of outbound peers required before ibd
+	// is confident we are synced.
+	minNumOutbound = 4
```
DavidVorick: I would like to bump this to 5.

I would also like to have logic to boot peers who are not contributing new blocks.

modules/consensus/synchronize.go
```diff
@@ -10,22 +14,43 @@ import (
...
+		default:
+			panic("unrecognized build.Release")
+		}
+	}()
+	// sendBlocksTimeout is the timeout for the SendBlocks RPC.
+	sendBlocksTimeout = func() time.Duration {
+		switch build.Release {
 		case "dev":
-			return 10
+			return 5 * time.Minute
```
DavidVorick: dev machines should assume local networks, no reason to make the timeout a full 5 minutes. 40 seconds should be good.

modules/consensus/synchronize.go
```diff
@@ -10,22 +14,43 @@ import (
...
 var (
 	// MaxCatchUpBlocks is the maxiumum number of blocks that can be given to
 	// the consensus set in a single iteration during the initial blockchain
 	// download.
 	MaxCatchUpBlocks = func() types.BlockHeight {
 		switch build.Release {
-		case "testing":
-			return 3
+		case "dev":
+			return 10
```
DavidVorick: I would almost make this higher for dev, since the dev constants will have smaller blocks and be on local networks.

modules/consensus/synchronize.go
VoidingWarranties: Changed it during debugging. It was at 5 seconds.

modules/consensus/synchronize.go
```diff
@@ -72,9 +97,21 @@ func blockHistory(tx *bolt.Tx) (blockIDs [32]types.BlockID) {
...
+	// SendBlocks will timeout. This is by design so that IBD switches peers to
+	// prevent any one peer from stalling IBD.
+	err := conn.SetDeadline(time.Now().Add(sendBlocksTimeout))
+	// Pipes do not support Set{,Read,Write}Deadline and should only be used in
+	// testing.
+	if build.Release != "testing" {
+		if opErr, ok := err.(*net.OpError); ok && opErr.Op == "set" && opErr.Net == "pipe" {
+			return err
+		}
+	}
```
DavidVorick: I don't understand what's going on here.

VoidingWarranties: Look below you

modules/consensus/synchronize.go
```diff
@@ -355,3 +400,73 @@ func (cs *ConsensusSet) threadedReceiveBlock(id types.BlockID) modules.RPCFunc {
...
 }
+
+// threadedInitialBlockchainDownload performs the IBD on outbound peers. Blocks
+// are downloaded from one peer at a time in 5 minute intervals, so as to
+// prevent any one peer from significantly slowing down IBD.
+//
+// NOTE: IBD will succeed right now when each peer has a different blockchain.
+// The height and the block id of the remote peers' current blocks are not
+// checked to be the same. This can cause issues if you are connected to
+// outbound peers <= v0.5.1 that are stalled in IBD.
```
DavidVorick: if you have somehow gotten 5 outbound peers which are all stalled...

hmm. Maybe peers should always return `more available` until they've managed to not be stalled themselves. Of course, that opens them up to a deadlock situation where all the peers connected to eachother are stalled, but then the peers will fix this by booting eachother.

So, I'm in favor of always returning `more available` until you are caught up. This could cause weird edge case behavior though, which would mean more testing...

heh

VoidingWarranties:
How long is too long for a long test? TestInitialBlockchainDownloadDoneRules takes 44 seconds.

DavidVorick:
44 seconds is fine, keep the whole thing under 5 minutes if possible. But even 10 minutes isn't too bad.

modules/consensus.go
```diff
@@ -58,6 +58,11 @@ type (
 		// There may not be any reverted blocks, but there will always be
 		// applied blocks.
 		ProcessConsensusChange(ConsensusChange)
+		// ConsensusSynced is called once after the initial blockchain download
+		// finishes, to indicate that the ConsensusSet is synced with the network. It
+		// is assumed that the ConsensusSet remains synced after ConsensusSynced is
+		// called.
+		//ConsensusSynced()
```
DavidVorick: I don't think this is a good idea.

modules/consensus/subscribe.go
```diff
@@ -180,3 +180,14 @@ func (cs *ConsensusSet) Unsubscribe(subscriber modules.ConsensusSetSubscriber) {
...
+// threadedNotifySubscribersConsensusSynced informs all subscribers that the
+// consensus set is synced with the network. This is typically called after
+// initial blockchain download completes.
+/*
+func (cs *ConsensusSet) threadedNotifySubscribersConsensusSynced() {
+	for _, subscriber := range cs.subscribers {
+		subscriber.ConsensusSynced()
+	}
+}
+*/
```
DavidVorick: ditto

modules/consensus/synchronize.go
```diff
@@ -71,10 +110,35 @@ func blockHistory(tx *bolt.Tx) (blockIDs [32]types.BlockID) {
...
+	// testing.
+	if opErr, ok := err.(*net.OpError); ok && opErr.Op == "set" && opErr.Net == "pipe" && build.Release == "testing" {
+		err = nil
+	}
+	if err != nil {
+		return err
+	}
+	stalled := true
+	defer func() {
+		// KLUDGE: Timeout errors returned by muxado do not conform to the net.Error
```
DavidVorick: Can you use one of the existing keywords? TODO: is probably sufficient. If we have too many keywords, they become difficult to grep.

modules/consensus.go
VoidingWarranties: Oops I didn't mean to commit that

modules/consensus/subscribe.go
VoidingWarranties: Or that

VoidingWarranties:
TestInitialBlockchainDownloadDoneRules only takes so long because `minIBDWaitTime` is 10 seconds during testing. It needs to be 10 seconds so that large reorg tests have enough time to complete. However, this specific test doesn't involve any block transfers so we could manually set `minIBDWaitTime` to 2s for just this test. That would shorten this test to just 9 seconds.

DavidVorick:
No need to worry about it. Since we use `t.Parallel()` (or at least, we're migrating to using it) there shouldn't be any issues with long sleeps. As long as the whole suite (per package) is finishing inside of 5 minutes, there's nothing to worry about.

PR #1014, opened 2016-03-10
New contract protocol (Renter side)

lukechampine:
The protocol code is almost ready, but not quite. There is some tricky signature stuff that I wasn't comfortable implementing on my own.

This PR also adds some spending metrics, and persistent subscription.

Not yet implemented:
- If the connection drops during revision, the renter should immediately attempt to reconnect and resend the last action.
- Need to figure out how host pubkeys should be handled by the hostdb. Right now I'm just assuming a `PublicKey` field in the host settings.
- The contract protocol should be documented extensively somewhere, perhaps in the doc folder. We wrote it down earlier so it shouldn't be hard to convert to a more readable format.

### DIFF
--- a/modules/renter/contractor/contractor.go
+++ b/modules/renter/contractor/contractor.go
@@ -45,6 +45,8 @@ type Contractor struct {
 	cachedAddress types.UnlockHash // to prevent excessive address creation
 	contracts     map[types.FileContractID]Contract
 	renewHeight   types.BlockHeight // height at which to renew contracts
+	spentPeriod   types.Currency    // number of coins spent on file contracts this period
+	spentTotal    types.Currency    // number of coins spent on file contracts ever
 
 	mu sync.RWMutex
 }
@@ -54,6 +56,13 @@ func (c *Contractor) Allowance() modules.Allowance {
 	c.mu.RLock()
 	defer c.mu.RUnlock()
 	return c.allowance
+}
+
+// Spending returns the number of coins spent on file contracts.
+func (c *Contractor) Spending() (types.Currency, types.Currency) {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+	return c.spentPeriod, c.spentTotal
 }
 
 // SetAllowance sets the amount of money the Contractor is allowed to spend on
@@ -87,7 +96,7 @@ func (c *Contractor) SetAllowance(a modules.Allowance) error {
 	// Otherwise, if the new allowance is "significantly different" (to be
 	// defined more precisely later), form intermediary contracts.
 	if a.Funds.Cmp(old.Funds) > 0 {
-		// not yet implemented
+		// TODO: implement
 		// c.formContracts(diff(a, old))
 	}
 
--- a/modules/renter/contractor/contractor.go
+++ b/modules/renter/contractor/contractor.go
@@ -44,6 +44,7 @@ type Contractor struct {
 	blockHeight   types.BlockHeight
 	cachedAddress types.UnlockHash // to prevent excessive address creation
 	contracts     map[types.FileContractID]Contract
+	lastChange    modules.ConsensusChangeID
 	renewHeight   types.BlockHeight // height at which to renew contracts
 	spentPeriod   types.Currency    // number of coins spent on file contracts this period
 	spentTotal    types.Currency    // number of coins spent on file contracts ever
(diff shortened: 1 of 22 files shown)


DavidVorick:
10 days ago? :( I'm getting sloppy.

modules/renter/contractor/contractor.go
```diff
@@ -56,6 +59,13 @@ func (c *Contractor) Allowance() modules.Allowance {
 	return c.allowance
 }
 
+// Spending returns the number of coins spent on file contracts.
+func (c *Contractor) Spending() (types.Currency, types.Currency) {
```
DavidVorick: because you are returning two of the same type that mean different things, the return values should definitely be named in the function definition, just for clarity's sake.

modules/renter/contractor/contractor.go
```diff
@@ -152,7 +162,16 @@ func newContractor(cs consensusSet, w wallet, tp transactionPool, hdb hostDB, d
 		return nil, err
 	}
 
-	cs.ConsensusSetPersistentSubscribe(c, modules.ConsensusChangeID{})
+	err = cs.ConsensusSetPersistentSubscribe(c, c.lastChange)
+	if err == modules.ErrInvalidConsensusChangeID {
+		c.lastChange = modules.ConsensusChangeID{}
+		// ??? fix things ???
```
DavidVorick: the renter is going to be getting consensus set updates starting from the genesis block. So you need to reset any internal state in the renter that actively updates with the current block, so that the whole renter is consistent with a renter that's never seen any blocks before.

This is definitely something that you should be testing as well. It's worth breaking this logic into a separate `rescan()` function so that you can randomly call `rescan()` during your integration tests and make sure that nothing breaks when you do so.

Basically, have the expectation that the user could replace the consensus database at any time (well, it's probably safe to assume that they'll close `siad` first) and that this will trigger a rescan in the renter.

modules/renter/contractor/editor.go
```diff
@@ -84,25 +84,38 @@ func (he *hostEditor) Upload(data []byte) (uint64, error) {
 	he.contractor.mu.RLock()
 	height := he.contractor.blockHeight
 	he.contractor.mu.RUnlock()
-	if height > he.contract.FileContract.WindowStart {
+	if height >= he.contract.FileContract.WindowStart {
```
DavidVorick: Do you have a test that will fail if this logic is incorrect or off-by-one? Seems like an easy mistake to make.

modules/renter/contractor/editor.go
```diff
@@ -121,38 +134,61 @@ func (he *hostEditor) Upload(data []byte) (uint64, error) {
...
+		if h != root {
+			index = i
+		} else {
+			newRoots = append(newRoots, h)
+		}
+	}
+	if index == -1 {
+		return errors.New("no record of that sector root")
+	}
+	tree := crypto.NewCachedTree(0) // height is not relevant here
```
DavidVorick: should throw in the correct height anyway.

lukechampine:
TODO: hostdb need to return host pubkey, probably using a new type.

modules/renter/contractor/contractor.go
```diff
@@ -56,6 +59,13 @@ func (c *Contractor) Allowance() modules.Allowance {
 	return c.allowance
 }
 
+// Spending returns the number of coins spent on file contracts.
+func (c *Contractor) Spending() (period, total types.Currency) {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+	return c.spentPeriod, c.spentTotal
+}
```
DavidVorick: woo

DavidVorick: spendPeriod is an unhelpful name

DavidVorick: the word period is overloaded and can also mean 'total', lol.

modules/renter/contractor/contractor.go
DavidVorick: Can you drop a TODO here?

modules/renter/contractor/contractor.go
```diff
@@ -152,7 +162,16 @@ func newContractor(cs consensusSet, w wallet, tp transactionPool, hdb hostDB, d
 		return nil, err
 	}
 
-	cs.ConsensusSetPersistentSubscribe(c, modules.ConsensusChangeID{})
+	err = cs.ConsensusSetPersistentSubscribe(c, c.lastChange)
+	if err == modules.ErrInvalidConsensusChangeID {
+		c.lastChange = modules.ConsensusChangeID{}
+		// ??? fix things ???
+		// subscribe again using the new ID
+		err = cs.ConsensusSetPersistentSubscribe(c, c.lastChange)
```
DavidVorick: passing the error on to the next `if` bracket seems error prone

PR #1019, opened 2016-03-15
Fix Close bugs

### LINKED ISSUE #1015: make cover: to many open files
I'm guessing this is because log files are not closed in some modules.

```
$ make cover pkgs=./modules/consensus
rm -rf release doc/whitepaper.aux doc/whitepaper.log doc/whitepaper.pdf
--- FAIL: TestIntegrationBroadcastRelayHeader (0.02s)
	synchronize_test.go:1100: open /var/folders/7q/hm_qblpd4h303j504dfg2zk40000gn/T/SiaTesting/consensus/TestIntegrationBroadcastRelayHeader2/wallet/wallet.log: too many open files
--- FAIL: TestIntegrationRelaySynchronize (0.00s)
	synchronize_test.go:1155: error opening consensus database: open /var/folders/7q/hm_qblpd4h303j504dfg2zk40000gn/T/SiaTesting/consensus/TestRelaySynchronize1/consensus/consensus.db: too many open files
--- FAIL: TestThreadedReceiveBlocksStalls (0.00s)
	synchronize_test.go:1286: open /var/folders/7q/hm_qblpd4h303j504dfg2zk40000gn/T/SiaTesting/consensus/TestThreadedReceiveBlocksStalls/gateway/gateway.log: too many open files
--- FAIL: TestIntegrationSendBlocksStalls (0.00s)
	synchronize_test.go:1403: open /var/folders/7q/hm_qblpd4h303j504dfg2zk40000gn/T/SiaTesting/consensus/TestThreadedReceiveBlocksTimesout - local/gateway/gateway.log: too many open files
--- FAIL: TestTryValidTransactionSet (0.00s)
	validtransaction_test.go:17: open /var/folders/7q/hm_qblpd4h303j504dfg2zk40000gn/T/SiaTesting/consensus/TestValidTransaction/gateway/gateway.log: too many open files
--- FAIL: TestTryInvalidTransactionSet (0.00s)
	validtransaction_test.go:48: open /var/folders/7q/hm_qblpd4h303j504dfg2zk40000gn/T/SiaTesting/consensus/TestValidTransaction/gateway/gateway.log: too many open files
```

VoidingWarranties:
There were a few bugs pertaining to module closing that are fixed in this PR:
* module Close methods would return immediately on error instead of finishing cleanup
* not all modules were Closed on API shutdown
* the gateway did not disconnect from its peers on shutdown

and a few improvements:
* module Close errors are returned from `Server.Close` instead of `Server.Serve`
* `Server.Close` waits for `Server.Serve` to terminate before terminating itself. This ensures the server is completely shutdown when `Server.Close` is done.

Also, the gateway now uses `persist.Logger` instead of `log.Logger` to be consistent with other modules.

Closes #1015

### DIFF
--- a/api/server.go
+++ b/api/server.go
@@ -120,7 +120,11 @@ func (srv *Server) Close() error {
 			errStrs = append(errStrs, fmt.Sprintf("explorer err: %v", err))
 		}
 	}
-	// TODO: close miner
+	if srv.miner != nil {
+		if err := srv.miner.Close(); err != nil {
+			errStrs = append(errStrs, fmt.Sprintf("miner err: %v", err))
+		}
+	}
 	if srv.wallet != nil {
 		// TODO: close wallet and lock the wallet in the wallet's Close method.
 		if srv.wallet.Unlocked() {
--- a/api/server.go
+++ b/api/server.go
@@ -76,14 +76,20 @@ func (srv *Server) Serve() error {
 	}
 
 	// safely close each module
+	if srv.host != nil {
+		srv.host.Close()
+	}
+	if srv.explorer != nil {
+		srv.explorer.Close()
+	}
+	if srv.wallet != nil {
+		srv.wallet.Lock()
+	}
 	if srv.cs != nil {
 		srv.cs.Close()
 	}
 	if srv.gateway != nil {
 		srv.gateway.Close()
-	}
-	if srv.wallet != nil {
-		srv.wallet.Lock()
 	}
 
 	return nil
--- a/api/server.go
+++ b/api/server.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"errors"
 	"fmt"
 	"net"
 	"net/http"
@@ -67,31 +68,52 @@ func (srv *Server) Serve() error {
 		srv.listener.Close()
 	}()
 
+	var errStrs []string
+
 	// The server will run until an error is encountered or the listener is
 	// closed, via either the Close method or the signal handling above.
 	// Closing the listener will result in the benign error handled below.
 	err := srv.apiServer.Serve(srv.listener)
 	if err != nil && !strings.HasSuffix(err.Error(), "use of closed network connection") {
-		return err
+		errStrs = append(errStrs, fmt.Sprintf("serve err: %v", err))
 	}
 
 	// safely close each module
 	if srv.host != nil {
-		srv.host.Close()
+		if err := srv.host.Close(); err != nil {
+			errStrs = append(errStrs, fmt.Sprintf("host err: %v", err))
+		}
 	}
+	// TODO: close renter (which should close hostdb as well)
 	if srv.explorer != nil {
-		srv.explorer.Close()
+		if err := srv.explorer.Close(); err != nil {
+			errStrs = append(errStrs, fmt.Sprintf("explorer err: %v", err))
+		}
 	}
+	// TODO: close miner
 	if srv.wallet != nil {
-		srv.wallet.Lock()
+		// TODO: close wallet and lock the wallet in the wallet's Close method.
+		if srv.wallet.Unlocked() {
+			if err := srv.wallet.Lock(); err != nil {
+				errStrs = append(errStrs, fmt.Sprintf("wallet err: %v", err))
+			}
+		}
 	}
+	// TODO: close transaction pool
 	if srv.cs != nil {
-		srv.cs.Close()
+		if err := srv.cs.Close(); err != nil {
+			errStrs = append(errStrs, fmt.Sprintf("consensus err: %v", err))
+		}
 	}
 	if srv.gateway != nil {
-		srv.gateway.Close()
+		if err := srv.gateway.Close(); err != nil {
+			errStrs = append(errStrs, fmt.Sprintf("gateway err: %v", err))
+		}
 	}
 
+	if len(errStrs) > 0 {
+		return errors.New(strings.Join(errStrs, "\n"))
+	}
 	return nil
 }
 
--- a/api/server.go
+++ b/api/server.go
@@ -69,13 +69,19 @@ func (srv *Server) Serve() error {
 	defer srv.wg.Done()
 
 	// stop the server if a kill signal is caught
-	sigChan := make(chan os.Signal)
+	sigChan := make(chan os.Signal, 1)
 	signal.Notify(sigChan, os.Interrupt, os.Kill)
-	defer signal.Reset(os.Interrupt, os.Kill)
+	defer signal.Stop(sigChan)
+	stop := make(chan struct{})
+	defer close(stop)
 	go func() {
-		<-sigChan
-		fmt.Println("\rCaught stop signal, quitting...")
-		srv.listener.Close()
+		select {
+		case <-sigChan:
+			fmt.Println("\rCaught stop signal, quitting...")
+			srv.listener.Close()
+		case <-stop:
+			// Don't leave a dangling goroutine.
+		}
 	}()
 
 	// The server will run until an error is encountered or the listener is
--- a/api/server.go
+++ b/api/server.go
@@ -1,7 +1,6 @@
 package api
 
 import (
-	"errors"
 	"fmt"
 	"net"
 	"net/http"
@@ -10,6 +9,7 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/modules"
 )
 
@@ -96,11 +96,11 @@ func (srv *Server) Serve() error {
 
 // Close closes the Server's listener, causing the HTTP server to shut down.
 func (srv *Server) Close() error {
-	var errStrs []string
+	var errs []error
 
 	// Close the listener, which will cause Server.Serve() to return.
 	if err := srv.listener.Close(); err != nil {
-		errStrs = append(errStrs, fmt.Sprintf("listener err: %v", err))
+		errs = append(errs, fmt.Errorf("listener.Close failed: %v", err))
 	}
 
 	// Wait for Server.Serve() to exit. We wait so that it's guaranteed that the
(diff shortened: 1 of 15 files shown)


api/server.go
```diff
@@ -76,15 +76,21 @@ func (srv *Server) Serve() error {
...
 	// safely close each module
+	if srv.host != nil {
+		srv.host.Close()
+	}
+	if srv.explorer != nil {
+		srv.explorer.Close()
+	}
+	if srv.wallet != nil {
+		srv.wallet.Lock()
+	}
```
lukechampine: foobar

api/server.go
```diff
@@ -120,7 +120,11 @@ func (srv *Server) Close() error {
 			errStrs = append(errStrs, fmt.Sprintf("explorer err: %v", err))
 		}
 	}
-	// TODO: close miner
+	if srv.miner != nil {
+		if err := srv.miner.Close(); err != nil {
+			errStrs = append(errStrs, fmt.Sprintf("miner err: %v", err))
+		}
+	}
```
lukechampine: foo

api/server.go
```diff
@@ -74,22 +91,62 @@ func (srv *Server) Serve() error {
...
-		srv.gateway.Close()
+
+	// Wait for Server.Serve() to exit. We wait so that it's guaranteed that the
+	// server has completely closed after Close() returns. This is particularly
+	// useful during testing so that we don't exit a test before Serve() finishes.
+	srv.wg.Wait()
+
+	// Safely close each module.
+	if srv.host != nil {
+		if err := srv.host.Close(); err != nil {
```
lukechampine: try this:
```go
modules := []io.Closer{srv.host, srv.explorer, srv.miner...}
for _, m : = range modules {
    if m != nil {
        if err := m.Close(); err != nil {
            errs = append(errs, err)
        }
    }
}
```
although, then you lose the name in the error message. I suppose you could use a map for that, though...

modules/gateway/gateway.go
```diff
@@ -62,7 +68,30 @@ func (g *Gateway) Close() error {
...
+		errStrs = append(errStrs, fmt.Sprintf("listener error: %v", err))
+	}
+	// Disconnect from peers.
+	for _, p := range g.Peers() {
+		if err := g.Disconnect(p.NetAddress); err != nil {
+			errStrs = append(errStrs, fmt.Sprintf("peer error: %v", err))
+		}
+	}
+	// Sleep to give time for all go routines to exit. This is necessary because
+	// some go routines write to the logger so we must give them time to exit
```
lukechampine: goroutine is one word

modules/gateway/gateway.go
```diff
@@ -62,7 +68,30 @@ func (g *Gateway) Close() error {
...
+	// before closing the logger.
+	// TODO: block until go routines exit instead of sleeping.
```
lukechampine: goroutine is one word

modules/gateway/gateway.go
```diff
@@ -62,7 +68,30 @@ func (g *Gateway) Close() error {
...
+	time.Sleep(100 * time.Millisecond)
+	// Close the logger. The logger should be the last thing to shut down so that
+	// all other objects have access to logging while closing.
+	if err := g.log.Close(); err != nil {
+		errStrs = append(errStrs, fmt.Sprintf("log error: %v", err))
+	}
+
+	if len(errStrs) > 0 {
+		return errors.New(strings.Join(errStrs, "; "))
+	}
```
lukechampine: this pattern is becoming common enough that we should consider abstracting it.

modules/gateway/peers.go
```diff
@@ -273,16 +273,14 @@ func (g *Gateway) Connect(addr modules.NetAddress) error {
 // Disconnect terminates a connection to a peer and removes it from the
 // Gateway's peer list. The peer's address remains in the node list.
 func (g *Gateway) Disconnect(addr modules.NetAddress) error {
-	id := g.mu.RLock()
+	id := g.mu.Lock()
+	defer g.mu.Unlock(id)
```
lukechampine: `p.sess.Close()` is a blocking call (it sends a "GoAway" frame), so it shouldn't happen while holding a mutex.

modules/miner.go
```diff
@@ -71,4 +71,8 @@ type TestMiner interface {
 type Miner interface {
 	BlockManager
 	CPUMiner
+
+	// Close terminates all ongoing processes involving the miner, enabling garbage
+	// collection.
+	Close() error
```
lukechampine: `io.Closer` is probably fine

persist/log.go
```diff
@@ -13,11 +14,14 @@ import (
 // panic if they are called after the file has already been closed.
 type closeableFile struct {
 	*os.File
+	mu     sync.RWMutex
```
lukechampine: should be the last field.

api/server.go
```diff
@@ -55,16 +62,26 @@ func NewServer(APIaddr string, requiredUserAgent string, cs modules.ConsensusSet
...
+	defer srv.wg.Done()
+
 	// stop the server if a kill signal is caught
-	sigChan := make(chan os.Signal)
+	sigChan := make(chan os.Signal, 1)
 	signal.Notify(sigChan, os.Interrupt, os.Kill)
-	defer signal.Reset(os.Interrupt, os.Kill)
+	defer signal.Stop(sigChan)
+	stop := make(chan struct{})
+	defer func() { stop <- struct{}{} }()
```
lukechampine: `defer close(stop)`

modules/gateway/peers.go
VoidingWarranties: `p.sess.Close()` is a method on the `muxado.Session` type and shouldn't mutate the internal state of the gateway, right?

modules/gateway/gateway.go
VoidingWarranties: Hmmm I used this pattern a few places in this PR but looking at it again I'm not sure I like it. I also haven't seen any other packages use a similar pattern, although I haven't looked very hard. What do you think about returning just the first error?

```go
var returnErr error // or returnErr could be the func's named return value

if err := foo.Close(); err != nil && returnErr == nil {
	returnErr = err
}
if err := bar(); err != nil && returnErr == nil {
	returnErr = err
}

return returnErr
```

modules/gateway/peers.go
lukechampine: correct.

modules/gateway/gateway.go
lukechampine: I was thinking more like:
```go
func JoinErrors(errs []error, sep string) error {
    var strs []string
    for _, err := range errs {
        if err != nil {
            strs = append(strs, err.Error())
        }
    }
    if len(strs) == 0 {
        return nil
    }
    return errors.New(strings.Join(strs, sep))
}
```
Usage:
```go
var errs []error
errs = append(errs, foo.Close())
errs = append(errs, bar.Close())
return build.JoinErrors(errs)
```
Or, if you want context:
```go
var errs []error
if err := foo.Close(); err != nil {
    errs = append(errs, fmt.Errorf("foo failed: %v", err)
}
if err := bar.Close(); err != nil {
    errs = append(errs, fmt.Errorf("bar failed: %v", err)
}
return build.JoinErrors(errs)
```

api/server.go
VoidingWarranties: I like this, but right now the wallet doesn't have a close function, and it doesn't seem trivial to add because transaction pool doesn't have an Unsubscribe method.

To shutdown the modules in the reverse order they are created we can't use this method until the wallet has a close method.

modules/gateway/peers.go
VoidingWarranties: Reverted.

VoidingWarranties:
Done. Consensus still needs logging, but that's for another PR.

lukechampine:
Logging is out of scope for this PR. Move those commits to another branch.

persist/log.go
```diff
@@ -14,10 +15,13 @@ import (
 type closeableFile struct {
 	*os.File
 	closed bool
+	mu     sync.RWMutex
```
DavidVorick: Was this causing race conditions? Can you explain more?

modules/miner/miner.go
```diff
@@ -192,16 +194,27 @@ func New(cs modules.ConsensusSet, tpool modules.TransactionPool, w modules.Walle
 // Close terminates all ongoing processes involving the miner, enabling garbage
 // collection.
 func (m *Miner) Close() error {
+	var errStrs []string
```
DavidVorick: do you want to use JoinErrs here?

modules/gateway/persist.go
```diff
@@ -1,14 +1,19 @@
...
+const (
+	// nodesFile is the name of the file that contains all seen nodes.
+	nodesFile = "nodes.json"
+	// logFile is the name of the log file.
+	logFile = modules.GatewayDir + ".log"
+)
+
 var persistMetadata = persist.Metadata{
 	Header:  "Sia Node List",
 	Version: "0.3.3",
```
DavidVorick: did we want to change the version number here? Or do we want to do that after changing the RPC calls to types.Specifiers?

api/server.go
```diff
@@ -26,6 +28,11 @@ type Server struct {
 	apiServer         *http.Server
 	listener          net.Listener
 	requiredUserAgent string
+
+	// wg is used to block Close() from returning until Serve() has finished. A
+	// WaitGroup is used instead of a chan struct{} so that Close() can be called
+	// without necessarily calling Serve() first.
+	wg sync.WaitGroup
```
DavidVorick: I don't think this is necessary, you should be able to just call close after the server returns, right?

api/server.go
```diff
@@ -55,16 +62,26 @@ func NewServer(APIaddr string, requiredUserAgent string, cs modules.ConsensusSet
...
+	defer srv.wg.Done()
+
 	// stop the server if a kill signal is caught
-	sigChan := make(chan os.Signal)
+	sigChan := make(chan os.Signal, 1)
 	signal.Notify(sigChan, os.Interrupt, os.Kill)
-	defer signal.Reset(os.Interrupt, os.Kill)
+	defer signal.Stop(sigChan)
+	stop := make(chan struct{})
+	defer close(stop)
```
DavidVorick: this code is a mess, I'm guessing there's a much better way to handle it.

modules/gateway/persist.go
lukechampine: This version number is only for the persist file. It should only be changed if the layout of the persist file is changed.

api/server.go
lukechampine: Perhaps the signal handling code should be moved to `NewServer`?

api/server.go
```diff
@@ -55,16 +62,26 @@ func NewServer(APIaddr string, requiredUserAgent string, cs modules.ConsensusSet
...
 	go func() {
-		<-sigChan
-		fmt.Println("\rCaught stop signal, quitting...")
-		srv.listener.Close()
+		select {
+		case <-sigChan:
+			fmt.Println("\rCaught stop signal, quitting...")
+			srv.listener.Close()
```
lukechampine: is this sufficient? Shouldn't it be a full `srv.Close()`?

api/server.go
lukechampine: Well, currently `Serve` only returns *because* you called Close (or it caught a signal).

api/server.go
lukechampine: For reference, [my old shutdown proposal](https://github.com/NebulousLabs/Sia/pull/898#issuecomment-170403863).

persist/log.go
VoidingWarranties: `persist.Logger`s use `closeableFile`s. Therefore, logging from a goroutine caused a race condition.

PR #1020, opened 2016-03-15
Reorder siac gateway list so that address is last

VoidingWarranties:
and add an Outbound column

As per @lukechampine's suggestions

### DIFF
--- a/siac/gatewaycmd.go
+++ b/siac/gatewaycmd.go
@@ -111,9 +111,9 @@ func gatewaylistcmd() {
 	}
 	fmt.Println(len(info.Peers), "active peers:")
 	w := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
-	fmt.Fprintln(w, "Address\tVersion")
+	fmt.Fprintln(w, "Version\tOutbound\tAddress")
 	for _, peer := range info.Peers {
-		fmt.Fprintf(w, "  %v\t%v\n", peer.NetAddress, peer.Version)
+		fmt.Fprintf(w, "%v\t%v\t%v\n", peer.Version, yesNo(!peer.Inbound), peer.NetAddress)
 	}
 	w.Flush()
 }


siac/gatewaycmd.go
```diff
@@ -111,9 +111,9 @@ func gatewaylistcmd() {
 	}
 	fmt.Println(len(info.Peers), "active peers:")
 	w := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
-	fmt.Fprintln(w, "Address\tVersion")
+	fmt.Fprintln(w, "Version\tOutbound\tAddress")
 	for _, peer := range info.Peers {
-		fmt.Fprintf(w, "  %v\t%v\n", peer.NetAddress, peer.Version)
+		fmt.Fprintf(w, "  %v\t%v\t%v\n", peer.Version, yesNo(!peer.Inbound), peer.NetAddress)
```
lukechampine: should the two leading spaces be dropped?

VoidingWarranties: It looks a little weird without them. Before we printed using a tab writer we used a leading tab.

lukechampine: won't the spaces screw up the alignment though? e.g.
```
Version    Outbound    Address
  0.5.1    Yes         foo:9981
  0.4.8    No          bar:20200
```

VoidingWarranties: In the past I've right-aligned numbers and left-aligned strings (e.g. `siac renter list -v`). Of course the version number isn't actually right aligned right now, it just has the appearance of being right aligned. If you ignore the trailing vs. leading space, the alignment only matters if we ever have a double-digit version number.

With that in mind I think this:
```
Version    Outbound    Address
0.5.10    Yes         foo:9981
0.5.9    Yes         bar:9981
```
looks better than this:
```
Version    Outbound    Address
 0.5.10    Yes         foo:9981
  0.5.9    Yes         bar:9981
```
I'll remove the leading spaces.

PR #1021, opened 2016-03-15
Proper HostDB persistence

lukechampine:
The HostDB should no longer rescan every host on startup.
This is important for the contractor to function well.

### DIFF
--- a/modules/renter/hostdb/hostentry.go
+++ b/modules/renter/hostdb/hostentry.go
@@ -10,9 +10,9 @@ import (
 type hostEntry struct {
 	modules.HostDBEntry
 
-	weight      types.Currency
-	reliability types.Currency
-	online      bool
+	Weight      types.Currency
+	Reliability types.Currency
+	Online      bool
 }
 
 // insert adds a host entry to the state. The host will be inserted into the
@@ -36,7 +36,7 @@ func (hdb *HostDB) insertHost(host modules.HostDBEntry) {
 	// Create hostEntry and add to allHosts.
 	h := &hostEntry{
 		HostDBEntry: host,
-		reliability: DefaultReliability,
+		Reliability: DefaultReliability,
 	}
 	hdb.allHosts[host.NetAddress] = h
 
@@ -121,7 +121,7 @@ func (hdb *HostDB) IsOffline(addr modules.NetAddress) bool {
 		return false
 	}
 	if h, ok := hdb.allHosts[addr]; ok {
-		return !h.online
+		return !h.Online
 	}
 	return false
 }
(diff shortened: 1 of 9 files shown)


modules/renter/hostdb/persist.go
```diff
@@ -1,16 +1,42 @@
...
+	err := hdb.persist.load(&data)
+	if err != nil {
+		return err
+	}
+	for _, entry := range data.AllHosts {
+		hdb.allHosts[entry.NetAddress] = &entry
+	}
+	for _, entry := range data.ActiveHosts {
+		hdb.insertNode(&entry)
+	}
```
lukechampine: this is the only part I'm uncertain about. Maybe we should rescan the active hosts immediately?

DavidVorick: I guess we should, in case we have users that don't leave siad running long enough for a standard scan to happen randomly.

modules/renter/hostdb/hostentry.go
```diff
@@ -120,7 +120,7 @@ func (hdb *HostDB) IsOffline(addr modules.NetAddress) bool {
 		return false
 	}
 	if h, ok := hdb.allHosts[addr]; ok {
-		return !h.online
+		return !h.Online
 	}
 	// no record of the host; add it to the HostDB
 	hdb.insertHost(modules.HostSettings{NetAddress: addr})
```
DavidVorick: might not be a good choice, especially if the user is trying to keep a filter on which hosts are getting into the hostdb. On the other hand, insertHost might be the right place to add the filtering code, so maybe this is okay.

Should insertHost be called in a goroutine, I don't think  you want to block while it does whatever blackbox actions.

modules/renter/hostdb/persist.go
DavidVorick: Thought about this more, the whole rescanning logic is going to get something of an overhaul anyway, but what should really be happening here is you check a timestamp on the last rescan for each host and then decide if it's time to do another rescan based on randomness.

DavidVorick:
You need to add testing to make sure that the rescan code is running correctly.

modules/renter/hostdb/hostentry.go
lukechampine: If the renter checks a host once, it's probably going to check it again later. So it makes sense to add it to the hostdb at some point. I agree that this is a little implicit though.

`insertHost` is already non-blocking because it calls `scanHost`, which spawns a goroutine to send down a channel. Besides, since `insertHost` requires a lock, we'd need to re-lock the mutex inside this new goroutine. Bleh.

PR #1023, opened 2016-03-17
Consensus logging

VoidingWarranties:
Adds basic logging to consensus. None of the logging statements are wrapped in a `if build.DEBUG` as they are all logging errors that should not occur frequently.

### DIFF
--- a/modules/consensus/accept.go
+++ b/modules/consensus/accept.go
@@ -237,8 +237,10 @@ func (cs *ConsensusSet) managedAcceptBlock(b types.Block) error {
 			if err == errFutureTimestamp {
 				go func() {
 					time.Sleep(time.Duration(b.Timestamp-(types.CurrentTimestamp()+types.FutureThreshold)) * time.Second)
-					// TODO: log error returned by AcceptBlock if non-nill
-					cs.AcceptBlock(b) // NOTE: Error is not handled.
+					err := cs.AcceptBlock(b)
+					if err != nil {
+						cs.log.Println("WARN: failed to accept a future block:", err)
+					}
 				}()
 			}
 			return err
--- a/modules/consensus/accept.go
+++ b/modules/consensus/accept.go
@@ -239,7 +239,7 @@ func (cs *ConsensusSet) managedAcceptBlock(b types.Block) error {
 					time.Sleep(time.Duration(b.Timestamp-(types.CurrentTimestamp()+types.FutureThreshold)) * time.Second)
 					err := cs.AcceptBlock(b)
 					if err != nil {
-						cs.log.Println("WARN: failed to accept a future block:", err)
+						cs.log.Debugln("WARN: failed to accept a future block:", err)
 					}
 				}()
 			}
--- a/modules/consensus/consensusset.go
+++ b/modules/consensus/consensusset.go
@@ -8,6 +8,7 @@ package consensus
 
 import (
 	"errors"
+	"fmt"
 
 	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/encoding"
@@ -65,6 +66,7 @@ type ConsensusSet struct {
 
 	// Utilities
 	db         *persist.BoltDatabase
+	log        *persist.Logger
 	persistDir string
 	mu         demotemutex.DemoteMutex
 }
@@ -176,7 +178,15 @@ func (cs *ConsensusSet) ChildTarget(id types.BlockID) (target types.Target, exis
 func (cs *ConsensusSet) Close() error {
 	cs.mu.Lock()
 	defer cs.mu.Unlock()
-	return cs.db.Close()
+
+	var errs []error
+	if err := cs.db.Close(); err != nil {
+		errs = append(errs, fmt.Errorf("db.Close failed: %v", err))
+	}
+	if err := cs.log.Close(); err != nil {
+		errs = append(errs, fmt.Errorf("log.Close failed: %v", err))
+	}
+	return build.JoinErrors(errs, "; ")
 }
 
 // CurrentBlock returns the latest block in the heaviest known blockchain.
--- a/modules/consensus/synchronize.go
+++ b/modules/consensus/synchronize.go
@@ -331,7 +331,7 @@ func (cs *ConsensusSet) rpcRelayBlock(conn modules.PeerConn) error {
 		go func() {
 			err := cs.gateway.RPC(modules.NetAddress(conn.RemoteAddr().String()), "SendBlocks", cs.threadedReceiveBlocks)
 			if err != nil {
-				cs.log.Println("WARN: failed to get parents of orphan block:", err)
+				cs.log.Debugln("WARN: failed to get parents of orphan block:", err)
 			}
 		}()
 	}
@@ -362,7 +362,7 @@ func (cs *ConsensusSet) rpcRelayHeader(conn modules.PeerConn) error {
 		go func() {
 			err := cs.gateway.RPC(modules.NetAddress(conn.RemoteAddr().String()), "SendBlocks", cs.threadedReceiveBlocks)
 			if err != nil {
-				cs.log.Println("WARN: failed to get parents of orphan header:", err)
+				cs.log.Debugln("WARN: failed to get parents of orphan header:", err)
 			}
 		}()
 		return nil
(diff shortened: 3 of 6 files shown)


modules/consensus/consensusset.go
```diff
@@ -174,9 +175,13 @@ func (cs *ConsensusSet) ChildTarget(id types.BlockID) (target types.Target, exis
 
 // Close safely closes the block database.
 func (cs *ConsensusSet) Close() error {
+	// TODO: is this lock necessary?
 	cs.mu.Lock()
 	defer cs.mu.Unlock()
```
VoidingWarranties: Is it necessary or good practice to hold a lock during Close?

DavidVorick: if you are changing things, you should have a lock.

DavidVorick: furthermore, `Close` is exported, so people will be assuming that there's a lock. That doesn't necessarily mean that it needs a lock, but if it's a larger function it should have one so that devs don't make the wrong assumptions.

DavidVorick:
We definitely need some call to logging that will only log the message in DEBUG mode.

VoidingWarranties:
Agreed, more logging in DEBUG mode is good. I'm not very familiar with consensus code outside of synchronizing, so I don't know where else logging might be useful.

modules/consensus/accept.go
```diff
@@ -237,8 +237,10 @@ func (cs *ConsensusSet) managedAcceptBlock(b types.Block) error {
 			if err == errFutureTimestamp {
 				go func() {
 					time.Sleep(time.Duration(b.Timestamp-(types.CurrentTimestamp()+types.FutureThreshold)) * time.Second)
-					// TODO: log error returned by AcceptBlock if non-nill
-					cs.AcceptBlock(b) // NOTE: Error is not handled.
+					err := cs.AcceptBlock(b)
+					if err != nil {
+						cs.log.Println("WARN: failed to accept a future block:", err)
```
DavidVorick: this should be debug logged only.

modules/consensus/consensusset.go
```diff
@@ -176,7 +177,10 @@ func (cs *ConsensusSet) ChildTarget(id types.BlockID) (target types.Target, exis
 func (cs *ConsensusSet) Close() error {
 	cs.mu.Lock()
 	defer cs.mu.Unlock()
-	return cs.db.Close()
+	return build.JoinErrors([]error{
+		build.AddContext("log.Close failed:", cs.log.Close()),
+		build.AddContext("db.Close failed:", cs.db.Close()),
+	}, "; ")
```
DavidVorick: This is interesting, I'm not sure I like the construction. I like both the context and the fact that we're joining the errors instead of returning early, but not necessarily the actual way the code is constructed. You are returning a function that's taking an array of function calls that have function calls as arguments.

Doesn't seem like the right way to do things.

Also, if you put the last argument of `JoinErrors` as the first argument, you could make the array variadic, which I think is more appropriate for `JoinErrors` anyway.

modules/consensus/synchronize.go
```diff
@@ -327,9 +327,12 @@ func (cs *ConsensusSet) rpcRelayBlock(conn modules.PeerConn) error {
 		// If the block is an orphan, try to find the parents. The block
 		// received from the peer is discarded and will be downloaded again if
 		// the parent is found.
-		//
-		// TODO: log error returned if non-nill.
-		go cs.gateway.RPC(modules.NetAddress(conn.RemoteAddr().String()), "SendBlocks", cs.threadedReceiveBlocks)
+		go func() {
+			err := cs.gateway.RPC(modules.NetAddress(conn.RemoteAddr().String()), "SendBlocks", cs.threadedReceiveBlocks)
+			if err != nil {
+				cs.log.Println("WARN: failed to get parents of orphan block:", err)
```
DavidVorick: definitely needs to be logged only if DEBUG

modules/consensus/synchronize.go
```diff
@@ -355,18 +358,24 @@ func (cs *ConsensusSet) rpcRelayHeader(conn modules.PeerConn) error {
 	cs.mu.RUnlock()
 	if err == errOrphan {
 		// If the header is an orphan, try to find the parents.
-		//
-		// TODO: log error returned if non-nill.
-		go cs.gateway.RPC(modules.NetAddress(conn.RemoteAddr().String()), "SendBlocks", cs.threadedReceiveBlocks)
+		go func() {
+			err := cs.gateway.RPC(modules.NetAddress(conn.RemoteAddr().String()), "SendBlocks", cs.threadedReceiveBlocks)
+			if err != nil {
+				cs.log.Println("WARN: failed to get parents of orphan header:", err)
```
DavidVorick: same here

modules/consensus/synchronize.go
```diff
@@ -355,18 +358,24 @@ func (cs *ConsensusSet) rpcRelayHeader(conn modules.PeerConn) error {
...
 	}
 	// If the header is valid and extends the heaviest chain, fetch, accept it,
 	// and broadcast it.
-	//
-	// TODO: log error returned if non-nill.
-	go cs.gateway.RPC(modules.NetAddress(conn.RemoteAddr().String()), "SendBlk", cs.threadedReceiveBlock(h.ID()))
+	go func() {
+		err := cs.gateway.RPC(modules.NetAddress(conn.RemoteAddr().String()), "SendBlk", cs.threadedReceiveBlock(h.ID()))
+		if err != nil {
+			cs.log.Println("WARN: failed to get header's corresponding block:", err)
```
DavidVorick: aaaand same here

modules/consensus/synchronize.go
```diff
@@ -433,8 +442,9 @@ func (cs *ConsensusSet) threadedInitialBlockchainDownload() {
 	// Set the deadline 10 minutes in the future. After this deadline, we will say
 	// IBD is done as long as there is at least one outbound peer synced.
 	deadline := time.Now().Add(minIBDWaitTime)
+	numOutboundSynced := 0
 	for {
-		numOutboundSynced := 0
+		numOutboundSynced = 0
```
DavidVorick: can you explain why you moved `numOutboudSynced` outside the loop? I'm guessing it was a bug?

DavidVorick:
As you can see from my comments, pretty picky about what types of logging statements should be making production. Basically, only statements which indicate that the developer has done something wrong (like peers failing to disconnect) or that the user machine is having errors. Things that are triggered by bad peers, etc. shouldn't be logged unless in DEBUG mode.

One-time statements such as 'IBD complete with x peers' also make sense.

modules/consensus/consensusset.go
lukechampine: Variadic appeals to me too, although `build.JoinErrors` currently has nice symmetry with `strings.Join`.

Ooh, something else occurred to me: if `build.JoinErrors` only sees one non-nil error, it should return that error directly instead of creating a new error. This preserves the ability to check for things like `io.EOF`.

VoidingWarranties: >Ooh, something else occurred to me: if `build.JoinErrors` only sees one non-nil error, it should return that error directly instead of creating a new error. This preserves the ability to check for things like `io.EOF`.

That seems nice in theory, but I don't see the practical advantage. It encourages clients of `JoinErrors` or functions that use `JoinErrors` to rely on the behavior and trust that there is only one error being returned.

modules/consensus/synchronize.go
VoidingWarranties: Not a bug, it's needed for the last line outside the for loop
```go
cs.log.Printf("INFO: IBD done, synced with %v peers", numOutboundSynced)
```

modules/consensus/consensusset.go
lukechampine: hmm. Fair point. It seems like most of these joined errors go straight to logs or to the user anyway.

VoidingWarranties:
I added `Logger.Debug`, `Logger.Debugf`, and `Logger.Debugln`. There's not testing for them as `build.DEBUG` is a constant that's always true during testing.

I changed the noted log statements to use `Debugln`.

modules/consensus/consensusset.go
VoidingWarranties: A disadvantage of the `build.JoinErrors` pattern here is that you can only call functions that only return an error, and you cannot intersperse other function calls (such as locking / unlocking).

Then again, maybe it is good style to enforce that the components of the module be able to close easily like this, as it makes it easier to reason about.

VoidingWarranties:
Build is failing because of https://github.com/travis-ci/travis-ci/issues/5806

DavidVorick:
Don't we already only have a `Println` for the logger? I don't think adding all 3 was necessary, though strictly speaking I don't think it's harmful

PR #1024, opened 2016-03-18
add test for file contract revisions

DavidVorick:
I thought there was a bug here, but my test failed to turn anything up. And then I looked closer at the problems I was having with the host, and I realized the problem was not what I had originally diagnosed. Sorry.

Otoh, now we've got a nice test.

### DIFF
--- a/modules/transactionpool/accept_test.go
+++ b/modules/transactionpool/accept_test.go
@@ -382,3 +382,60 @@ func TestIntegrationNilAccept(t *testing.T) {
 		t.Error("no error returned when submitting nothing to the transaction pool")
 	}
 }
+
+// TestAcceptFCAndConflictingRevision checks that the transaction pool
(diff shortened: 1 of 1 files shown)


modules/transactionpool/accept_test.go
```diff
@@ -382,3 +382,60 @@ func TestIntegrationNilAccept(t *testing.T) {
 		t.Error("no error returned when submitting nothing to the transaction pool")
 	}
 }
+
+// TestAcceptFCAndConflictingRevision checks that the transaction pool
+// correctly accepts a file contract in a transaction set followed by a correct
+// revision to that file contract in the a following transaction set, with no
+// block separating them. This test was caught by a severe bug in production,
+// which had significantly impacted the renter's functioning in a negative way.
```
VoidingWarranties: Lies.

DavidVorick: darn

PR #1026, opened 2016-03-18
Shutdown API server only after /daemon/stop request completes

lukechampine:
Without this change, `siac stop` will print `no response from daemon`, even though the daemon shuts down correctly.


DavidVorick:
lol, wasn't your original PR like 40 lines before testing? Now you've got it down to 24 test included.

Well done.

PR #1027, opened 2016-03-19
Storage obligations

DavidVorick:
The host now manages storage obligations, hopefully in a way that is correct.

I have some very minimal testing that suggests it mostly works. I suspect that there are some fairly significant bugs, because the testing is minimal, but I wanted to get everything out there so we could start on the negotiation code and begin fixing the api tests.


crypto/merkle.go
```diff
@@ -95,6 +95,21 @@ func BuildReaderProof(r io.Reader, proofIndex uint64) (base []byte, hashSet []Ha
 	return
 }
 
+// BuildMerkleProof will build a proof that the segment at
```
DavidVorick: comment is incomplete

modules/host.go
```diff
@@ -53,6 +48,26 @@ var (
...
+		if build.Release == "dev" {
+			return 1 << 20 // 1 MiB
+		}
+		if build.Release == "standard" {
+			return 1 << 22 // 4 MiB
+		}
+		if build.Release == "testing" {
+			return 1 << 12 // 4 KiB
+		}
+		panic("unrecognized release constant in host - sectorSize")
```
DavidVorick: did we decide that there was a better way to do this? I forget. Oh, I think we were planning on adding type safety to catch misspellings at compile time.

modules/host.go
```diff
@@ -72,26 +87,26 @@ type (
...
 	// are in bytes per second.
 	HostBandwidthLimits struct {
-		DownloadDataLimit  uint64
-		DownloadSpeedLimit uint64
-		UploadDataLimit    uint64
-		UploadSpeedLimit   uint64
+		DownloadDataLimit  uint64 `json:"downloaddatalimit"`
+		DownloadSpeedLimit uint64 `json:"downloadspeedlimit"`
+		UploadDataLimit    uint64 `json:uploaddatalimit"`
+		UploadSpeedLimit   uint64 `json:uploadspeedlimit"`
```
DavidVorick: missing quotes

crypto/merkle.go
```diff
@@ -95,6 +95,21 @@ func BuildReaderProof(r io.Reader, proofIndex uint64) (base []byte, hashSet []Ha
...
+func BuildMerkleProof(sourceData []byte, proofIndex uint64) (proofSet [][]byte, err error) {
+	// Get the proof set for the sourceData and proofIndex.
+	tree := merkletree.New(NewHash())
+	err = tree.SetIndex(proofIndex)
+	if err != nil {
+		return nil, err
+	}
+	for i := 0; i < len(sourceData); i += SegmentSize {
+		tree.Push(sourceData[i : i+SegmentSize])
+	}
```
lukechampine: `BuildReaderProof(bytes.NewReader(sourceData, proofIndex)`?

modules/host/host.go
```diff
@@ -343,18 +126,41 @@ type Host struct {
 	resourceLock sync.RWMutex
 }
 
+// checkUnlockHash will check that the host has an unlock hash. If the host
+// does not have an unlock hash, an attempt will be made to get an unlock hash
+// from the wallet. That may fail due to the wallet being locked, in which case
+// an error is returned.
+func (h *Host) checkUnlockHash() error {
```
lukechampine: name seems a little weird, since "check" doesn't usually imply that the function will modify state.

modules/host/sector.go
```diff
@@ -220,6 +225,34 @@ func (h *Host) addSector(sectorRoot crypto.Hash, expiryHeight types.BlockHeight,
...
+		sectorKey := h.sectorID(sectorRoot[:])
+		sectorUsageBytes := bsu.Get(sectorKey)
+		if sectorUsageBytes == nil {
+			return errSectorNotFound
+		}
+		var su sectorUsage
+		err = json.Unmarshal(sectorUsageBytes, &su)
+		if sectorUsageBytes == nil {
+			return err
+		}
```
lukechampine: duplicate check

modules/host/storageobligations.go
```diff
@@ -1,14 +1,32 @@
...
+
+// TODO: Need to queue the action item for checking on the submission status of
+// the file contract revision. Also need to make sure that multiple actions are
+// being taken if needed.
+
+// TODO: Document mistakes that got made when constructing the Sia consensus
+// code. The first is not putting timeouts on keys including signatures? The
+// second was using floating point values in consensus code. The third is that
+// you can't build a concise storage proof (e.g. 0 bytes) for a file contract
+// with a size of 0.
```
lukechampine: seems like an odd place for this comment...

modules/host/storageobligations.go
```diff
@@ -131,27 +227,45 @@ func (so *storageObligation) isSane() error {
...
+	return
+}
+
+// proofDeadline returns the height by which the storage proof must be
+// submitted.
+func (so *storageObligation) proofDeadline() types.BlockHeight {
+	if len(so.RevisionTransactionSet) > 0 {
+		return so.RevisionTransactionSet[len(so.RevisionTransactionSet)-1].FileContractRevisions[0].NewWindowEnd
+	}
+	return so.OriginTransactionSet[len(so.OriginTransactionSet)-1].FileContracts[0].WindowEnd
```
lukechampine: this sort of logic crops up in the contractor as well. It might be worth defining a single type for handling contract+revision. Something like:
```go
type RevisedContract struct {
    OriginalContract types.FileContract
    LastRevision   types.FileContractRevision
}
```
I'm not sure what its methods would look like, but maybe:
```go
func (rc RevisedContract) AfterRevision() types.FileContract {
    // If there is a revision, apply it to the original contract.
    // Otherwise just return the original.
}
```

modules/host/storageobligations.go
```diff
@@ -203,18 +298,23 @@ func (h *Host) addStorageObligation(so *storageObligation) error {
...
-	if build.DEBUG {
-		for _, txn := range so.OriginTransactionSet {
-			err := txn.StandaloneValid(h.blockHeight)
-			if err != nil {
-				h.log.Critical("invalid transaction is being added in a storage obligation")
-			}
-		}
+	// Sanity check - There needs to be enough time left on the file contract
+	// for the host to safely submit the file contract revision.
+	if so.expiration()-resubmissionTimeout <= h.blockHeight {
```
lukechampine: I try to avoid subtracting block heights whenever possible to prevent underflows.

modules/host/storageobligations.go
```diff
@@ -235,35 +364,383 @@ func (h *Host) addStorageObligation(so *storageObligation) error {
...
+	// proof.
+	err0 := h.tpool.AcceptTransactionSet(so.OriginTransactionSet)
+	// The file contract was already submitted to the blockchain, need to check
+	// after the resubmission timeout that it was submitted successfully.
+	err1 := h.queueActionItem(h.blockHeight+resubmissionTimeout, soid)
+	// The storage proof should be submitted
+	err2 := h.queueActionItem(so.expiration()+resubmissionTimeout, soid)
+	err = composeErrors(err0, err1, err2)
+	if err != nil {
+		return composeErrors(err, h.removeStorageObligation(so, obligationRejected))
```
lukechampine: maybe use `build.JoinErrors` now?

modules/host/storageobligations.go
```diff
@@ -235,35 +364,383 @@ func (h *Host) addStorageObligation(so *storageObligation) error {
...
+	var err error
+	for i = range sectorsGained {
+		err = h.addSector(sectorsGained[i], so.expiration(), gainedSectorData[i])
+		if err != nil {
+			break
+		}
+	}
+	if err != nil {
+		// Because there was an error, all of the sectors that got added need
+		// to be revered.
```
lukechampine: reversed lol

modules/host/storageobligations.go
```diff
@@ -235,35 +364,383 @@ func (h *Host) addStorageObligation(so *storageObligation) error {
...
+		for j := 0; j < i; j++ {
+			// Error is not checkeed because there's nothing useful that can be
+			// done about an error.
+			_ = h.removeSector(sectorsGained[j], so.expiration())
```
lukechampine: couldn't this be done inside the loop?
```go
for i := range sectorsGained {
    err := h.addSector(sectorsGained[i], so.expiration(), gainedSectorData[i])
    if err != nil {
        // before returning the error, we need to revert all the sectors that were added
        for _, added := range sectorsGained[:i] {
            _ = h.removeSector(added, so.expiration())
        }
        return err
    }
}
```

modules/host/storageobligations.go
```diff
@@ -235,35 +364,383 @@ func (h *Host) addStorageObligation(so *storageObligation) error {
...
 		}
+		return tx.Bucket(bucketStorageObligations).Put(soid[:], soBytes)
+	})
+	if err != nil {
+		// Because there was an error, all of the sectors that got added need
+		// to be revered.
+		for i := range sectorsGained {
+			// Error is not checkeed because there's nothing useful that can be
+			// done about an error.
+			_ = h.removeSector(sectorsGained[i], so.expiration())
```
lukechampine: actually, since you do this a lot, you could define a `revertSectors([]crypto.Hash)` helper function. That would mean fewer lines in the error branch above, which is always good.

modules/host/storageobligations.go
```diff
@@ -235,35 +364,383 @@ func (h *Host) addStorageObligation(so *storageObligation) error {
...
+		// Error is not checked, we want to call remove on every sector even if
+		// there are problems - disk health information will be updated.
+		_ = h.removeSector(root, so.expiration())
+	}
+
+	// Update the host revenue metrics based on the status of the obligation.
+	if sos == obligationConfused {
+		h.log.Critical("storage obligation confused!")
+	}
+	if sos == obligationRejected {
```
lukechampine: `switch` is preferable because the values are exclusive (i.e. there's no possibility of fallthrough)

modules/host/update.go
```diff
@@ -18,7 +28,34 @@ func (h *Host) initRescan() error {
...
+			for k, soBytes := c.First(); soBytes != nil; k, soBytes = c.Next() {
+				var so storageObligation
+				err := json.Unmarshal(soBytes, &so)
+				if err != nil {
+					return err
+				}
+				so.OriginConfirmed = false
+				so.RevisionConfirmed = false
+				so.ProofConfirmed = false
+				allObligations = append(allObligations, &so)
```
lukechampine: pretty sure this is no good. `&so` will be the same for every obligation

modules/host/update.go
```diff
@@ -33,9 +70,18 @@ func (h *Host) initRescan() error {
...
-	// obligations.
-
+	// Re-queue all of the action items for the storage obligations.
+	for _, so := range allObligations {
+		soid := so.id()
+		err0 := h.tpool.AcceptTransactionSet(so.OriginTransactionSet)
+		err1 := h.queueActionItem(h.blockHeight+resubmissionTimeout, soid)
+		err2 := h.queueActionItem(so.expiration()-revisionSubmissionBuffer, soid)
+		err3 := h.queueActionItem(so.expiration()+resubmissionTimeout, soid)
+		err = composeErrors(err0, err1, err2, err3)
```
lukechampine: could be:
```go
err = composeErrors(
    h.tpool.Accept,
    h.queue,
    h.queue,
    ...
)
```

modules/host/update.go
```diff
@@ -58,27 +104,182 @@ func (h *Host) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+	// Wrap the whole parsing into a single large database tx to keep things
+	// efficient.
+	var actionItems []*storageObligation
+	err := h.db.Update(func(tx *bolt.Tx) error {
+		for _, block := range cc.RevertedBlocks {
+			// Look for transactions relevant to open storage obligations.
+			for i, txn := range block.Transactions {
+				// Check for file contracts.
+				if len(txn.FileContracts) > 0 {
+					for _ = range txn.FileContracts {
```
lukechampine: need proper index

modules/host/update.go
```diff
@@ -58,27 +104,182 @@ func (h *Host) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+						fcid := txn.FileContractID(uint64(i))
```
lukechampine: wrong `i`

modules/host/update.go
```diff
@@ -58,27 +104,182 @@ func (h *Host) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+			// the genesis block, height will already be at height 0 and should not
+			// update.
+			if block.ID() != types.GenesisBlock.ID() {
+				h.blockHeight++
+			}
+
+			// Handle any action items relevant to the current height.
+			bai := tx.Bucket(bucketActionItems)
+			heightBytes := make([]byte, 8)
+			binary.BigEndian.PutUint64(heightBytes, uint64(h.blockHeight))
```
lukechampine: is there a comment somewhere that explains why big endian is used? (I assume so that bolt orders them efficiently)

modules/host/update.go
```diff
@@ -58,27 +104,182 @@ func (h *Host) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+			existingItems := bai.Get(heightBytes)
+
+			// From the existing items, pull out a storage obligation.
+			obligationIDs := make([]types.FileContractID, len(existingItems)/crypto.HashSize)
+			for i := 0; i < len(existingItems); i += crypto.HashSize {
+				copy(obligationIDs[i/crypto.HashSize][:], existingItems[i:i+crypto.HashSize])
+			}
```
lukechampine: Safer to use the encoding package to store/retrieve these:
```go
// queueActionItem
existingItems = append(existingItems, id) // now a slice of IDs, rather than a byte slice
bai.Put(heightBytes, encoding.Marshal(existingItems)

// ProcessConsensusChange
idBytes := bai.Get(heightBytes)
if idBytes == nil {
    // handle error
}
var obligationIDs []types.FileContractID
err = encoding.Unmarshal(idBytes, &obligationIDs)
```

modules/host/update.go
```diff
@@ -58,27 +104,182 @@ func (h *Host) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+			bso := tx.Bucket(bucketStorageObligations)
+			for _, soid := range obligationIDs {
+				soBytes := bso.Get(soid[:])
+				var so storageObligation
```
lukechampine: don't you have `getStorageObligation` for this?

modules/host/update.go
```diff
@@ -58,27 +104,182 @@ func (h *Host) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+				err := json.Unmarshal(soBytes, &so)
```
lukechampine: don't you have `getStorageObligation` for this?

types/validtransaction.go
```diff
@@ -60,9 +60,13 @@ func (t Transaction) correctFileContracts(currentHeight BlockHeight) error {
 		}
 		outputPortion := PostTax(currentHeight, fc.Payout)
 		if validProofOutputSum.Cmp(outputPortion) != 0 {
+			println("valid proof problems")
 			return ErrFileContractOutputSumViolation
 		}
 		if missedProofOutputSum.Cmp(outputPortion) != 0 {
+			println("missed proof problems")
+			println(missedProofOutputSum.String())
+			println(outputPortion.String())
```
lukechampine: debugging

modules/host/update.go
```diff
@@ -58,27 +104,182 @@ func (h *Host) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+				if len(txn.FileContracts) > 0 {
+					for i := range txn.FileContracts {
+						fcid := txn.FileContractID(uint64(i))
+						so, err := getStorageObligation(tx, fcid)
+						if err != nil {
+							// The storage folder may not exist, or the disk
+							// may be having trouble. Either way, we ignore the
+							// problem. If the disk is having trouble, the user
+							// will have to perform a rescan.
+							continue
```
VoidingWarranties: It might be useful to log these errors before continuing.

modules/host/update.go
```diff
@@ -58,27 +104,182 @@ func (h *Host) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+				// Check for file contract revisions.
+				if len(txn.FileContractRevisions) > 0 {
+					for _, fcr := range txn.FileContractRevisions {
+						so, err := getStorageObligation(tx, fcr.ParentID)
+						if err != nil {
+							// The storage folder may not exist, or the disk
+							// may be having trouble. Either way, we ignore the
+							// problem. If the disk is having trouble, the user
+							// will have to perform a rescan.
+							continue
```
VoidingWarranties: Ditto for logging.

modules/host/update.go
```diff
@@ -58,27 +104,182 @@ func (h *Host) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+				// Check for storage proofs.
+				if len(txn.StorageProofs) > 0 {
+					for _, sp := range txn.StorageProofs {
+						so, err := getStorageObligation(tx, sp.ParentID)
+						if err != nil {
+							// The storage folder may not exist, or the disk
+							// may be having trouble. Either way, we ignore the
+							// problem. If the disk is having trouble, the user
+							// will have to perform a rescan.
+							continue
```
VoidingWarranties: log

modules/host/update.go
```diff
@@ -58,27 +104,182 @@ func (h *Host) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+				if len(txn.FileContracts) > 0 {
+					for _ = range txn.FileContracts {
+						fcid := txn.FileContractID(uint64(i))
+						so, err := getStorageObligation(tx, fcid)
+						if err != nil {
+							// The storage folder may not exist, or the disk
+							// may be having trouble. Either way, we ignore the
+							// problem. If the disk is having trouble, the user
+							// will have to perform a rescan.
+							continue
```
VoidingWarranties: log

modules/host/update.go
```diff
@@ -58,27 +104,182 @@ func (h *Host) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+						}
+						so.OriginConfirmed = false
+						err = putStorageObligation(tx, so)
+						if err != nil {
+							continue
```
VoidingWarranties: Why is this error ignored? It seems like `putStorageObligation` should only fail on a database error, in which case something is going wrong. The error should at least be logged. Same goes for all the other instances of `putStorageObligation`.

modules/host/host.go
DavidVorick: do you have suggestions for an alternative?

modules/host/storageobligations.go
DavidVorick: will clean this up over the next few days. There's clearly a lot of room for improvement in the host code and structure right now, but I'm trying to get it feature-complete before worrying too much about organization.

modules/host/storageobligations.go
DavidVorick: hmm. I wonder if I should add sanity checks here to guarantee that it'll panic on underflow.

modules/host/update.go
DavidVorick: I think there is, but doesn't hurt to have the same information in multiple places.

modules/host/update.go
DavidVorick: yeah, I was being lazy, probably shouldn't have been.

modules/host/update.go
DavidVorick: written after the fact, didn't migrate all the code over yet.

types/validtransaction.go
DavidVorick: aw man, I've tried to use `git diff` to prevent that from happening. I worry that they are in more places too though. This is definitely some of the sloppiest code I've written in a while.

modules/host/update.go
DavidVorick: if the database is having problems there are bigger issues. Mostly though I wanted to get this all written, there are a few places where the error handling isn't completely correct. But look at the size of this PR. If I had taken the time to do error handling everywhere we'd have added another week or more.

I will revisit this, but right now I think it's more important to get the api integration tests working again, especially since this code isn't consensus critical or anything.

modules/host/host.go
lukechampine: perhaps `getUnlockHash` or jst `unlockHash`, and it would return the new (or cached) unlock hash? Depends on how it's used, I guess.

modules/host/update.go
```diff
@@ -113,8 +113,8 @@ func (h *Host) ProcessConsensusChange(cc modules.ConsensusChange) {
 			for i, txn := range block.Transactions {
 				// Check for file contracts.
 				if len(txn.FileContracts) > 0 {
-					for _ = range txn.FileContracts {
-						fcid := txn.FileContractID(uint64(i))
+					for j = range txn.FileContracts {
```
lukechampine: it looks like you don't use the `i` in the `block.Transactions` range above, so you could use `i` instead. That's what you do in the `AppliedBlocks` loop.

modules/host/host.go
DavidVorick: it's used to check whether the host has an unlock hash, and if the host doesn't have one then it gives the host an unlock hash.

I think `checkUnlockHash` as a name makes sense, though I understand the concerns about the state-change.

modules/host/storageobligations.go
DavidVorick: A closer look revealed that the logic was incorrect, and also that there were ways to construct the conditional such that only addition was used.

thx

modules/host/storageobligations.go
DavidVorick: *removes hat*

modules/host/storageobligations.go
DavidVorick: will keep that in mind later. Since I'm planning on doing a fair amount of refactoring, I'm not too worried about specific refactoring in this PR. My guess is at some point I'll find a way to remove 30-40% of the code here while making everything both more readable and more intuitively correct.

DavidVorick:
My priorities are a little split on this PR. I don't want to spend too much time on it, and instead want to focus on getting to a point where our renter-host integration tests are working again. I've cleaned up a few things.

One big thing to note that I just did is change the crypto package to drop `ReaderMerkleRoot` and `BuildReaderProof` in favor of `MerkleRoot` and `MerkleProof`, after realizing that we weren't using a single reader in the entire codebase. Pretty much every single call was `ReaderMerkleRoot(bytes.NewReader(data))`.

The renter is broken now, but that's because modules/host.go has changed. I did update the renter to use the new crypto functions.

crypto/merkle.go
```diff
@@ -70,45 +92,39 @@ func CalculateLeaves(fileSize uint64) uint64 {
 	return numSegments
 }
 
-// ReaderMerkleRoot returns the Merkle root of a reader.
-func ReaderMerkleRoot(r io.Reader) (h Hash, err error) {
-	root, err := merkletree.ReaderRoot(r, NewHash(), SegmentSize)
-	if err != nil {
-		return
+// MerkleRoot returns the Merkle root of a reader.
+func MerkleRoot(b []byte) Hash {
```
lukechampine: comment is wrong.

I think I would prefer to continue using `merkletree.ReaderRoot` internally, because it lets us offload critical logic to the merkletree package, rather than reimplementing it here.

modules/host/storageobligations.go
```diff
@@ -654,29 +652,19 @@ func (h *Host) handleActionItem(so *storageObligation) {
...
-		proofSet, err := crypto.BuildMerkleProof(sectorBytes, sectorSegment)
-		if err != nil {
-			return
-		}
+		base, cachedHashSet := crypto.MerkleProof(sectorBytes, sectorSegment)
 
 		// Using the sector, build a cached root.
 		log2SectorSize := uint64(0)
-		for 1<<log2SectorSize < modules.SectorSize {
+		for 1<<log2SectorSize < (modules.SectorSize / crypto.SegmentSize) {
```
lukechampine: should be able to define `log2SectorSize` as a const, or at least a global var.

crypto/merkle.go
```diff
@@ -50,14 +46,40 @@ type CachedMerkleTree struct {
...
+	for i, proof := range proofSet[1:] {
+		copy(hashSet[i][:], proof)
+	}
+	return hashSet
+}
+
+// Push pushes a subtree Merkle root into a cached Merkle tree.
+func (ct *CachedMerkleTree) Push(h Hash) {
+	ct.CachedTree.Push(h[:])
+}
```
lukechampine: :+1:

crypto/merkle.go
```diff
@@ -1,8 +1,12 @@
 package crypto
 
-import (
-	"io"
+// TODO: It's likely that all of the readers in this file can be replaced with
+// simple byte slices, I believe that in the vast majority of cases things are
+// now handled one sector at a time, meaning that the callers are actually
+// creating new readers out of byte slices the majority of the time in order to
+// be able to interact with this file.
```
lukechampine: remove

crypto/merkle.go
```diff
@@ -70,29 +92,39 @@ func CalculateLeaves(fileSize uint64) uint64 {
...
-func ReaderMerkleRoot(r io.Reader) (h Hash, err error) {
-	root, err := merkletree.ReaderRoot(r, NewHash(), SegmentSize)
-	if err != nil {
-		return
+// MerkleRoot returns the Merkle root of a reader.
+func MerkleRoot(b []byte) Hash {
+	t := NewTree()
+	for len(b) >= SegmentSize {
+		t.Push(b[:SegmentSize])
+		b = b[SegmentSize:]
```
lukechampine: alternative:
```go
for i := 0; i < len(b); i += SegmentSize {
    t.Push(b[i:i+SegmentSize])
}
```
or:
```go
buf := bytes.NewBuffer(b)
for buf.Len() > 0 {
    t.Push(buf.Next(SegmentSize))
}
```
These let you omit the `if len(b) > 0` at the end.

DavidVorick: I find your use of `bytes.Buffer` to be pleasing.

DavidVorick: lol

DavidVorick:
ok, fixed up crypto/merkle.go some more. Added a test to make sure the `bytes.Buffer` was handling the bounds correctly, and as expected it was. Test coverage remains at 100% for crypto/merkle.go.

PR #1028, opened 2016-03-19
misc IBD fixes

VoidingWarranties:
Check for timeouts using muxado's timeout error instead of `net.Error.Timeout()`

Don't broadcast blocks in threadedReceiveBlocks during IBD. They are unnecessary and were caused by the frequent connection timeouts that are expected during IBD.

Also add a bunch of COMPAT comments that were previously missing.

### DIFF
--- a/modules/consensus/synchronize.go
+++ b/modules/consensus/synchronize.go
@@ -156,7 +156,7 @@ func (cs *ConsensusSet) threadedReceiveBlocks(conn modules.PeerConn) (returnErr 
 	// is to stop an attacker from preventing block broadcasts.
 	chainExtended := false
 	defer func() {
-		if chainExtended {
+		if chainExtended && cs.Synced() {
 			// The last block received will be the current block since
 			// managedAcceptBlock only returns nil if a block extends the longest chain.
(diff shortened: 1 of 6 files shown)


modules/consensus/synchronize.go
```diff
@@ -313,6 +313,7 @@ func (cs *ConsensusSet) rpcSendBlocks(conn modules.PeerConn) error {
 }
 
 // rpcRelayBlock is an RPC that accepts a block from a peer.
+// COMPATv0.5.1
```
DavidVorick: should include more information about why it's compatibility code.

DavidVorick:
> Don't broadcast blocks in threadedReceiveBlocks during IBD. They are unnecessary and were caused by the frequent connection timeouts that are expected during IBD.

We are still broadcasting the block at the end right?

VoidingWarranties:
>We are still broadcasting the block at the end right?

Depends on what you mean. If IBD is in progress, we do not broadcast any block, even the last one. If IBD has finished, then we broadcast the last block received.

DavidVorick:
ah, sounds good to me. Needs rebase.

PR #1032, opened 2016-03-22
Change IBD timeouts for 0.5.2 RC2

VoidingWarranties:
Should be reverted for 0.5.2 release

### DIFF
--- a/modules/consensus/synchronize.go
+++ b/modules/consensus/synchronize.go
@@ -51,13 +51,16 @@ var (
 		}
 	}()
 	// minIBDWaitTime is the time threadedInitialBlockchainDownload waits before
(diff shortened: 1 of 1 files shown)


modules/consensus/synchronize.go
```diff
@@ -56,7 +57,8 @@ var (
 		case "dev":
 			return 80 * time.Second
 		case "standard":
-			return 10 * time.Minute
+			// TODO: 90 minutes for 0.5.2 RC2. Change back to 10 minutes for release.
+			return 90 * time.Minute
```
DavidVorick: This should probably be kept at 90 minutes, even after RC2

DavidVorick:
needs rebasee

PR #1033, opened 2016-03-22
Misc test fixes

VoidingWarranties:
Should fix some tests that often failed.

### DIFF
--- a/api/daemon_test.go
+++ b/api/daemon_test.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"testing"
+	"time"
 
 	"github.com/NebulousLabs/Sia/build"
 )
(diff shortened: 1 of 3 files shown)


api/daemon_test.go
```diff
@@ -33,6 +34,9 @@ func TestStop(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	// Sleep to give time for server to close, as /daemon/stop will return success
+	// before Server.Close() is called.
+	time.Sleep(200 * time.Millisecond)
```
VoidingWarranties: Alternatively we could split up `Close` into `closeModules` and `closeServer` and change `daemonStopHandler` to
```go
srv.closeModules()
writeSuccess(w)
src.listener.Close()
```
and change `Close` to:
```go
srv.closeModules()
src.listener.Close()
```

DavidVorick: Seems too complicated for now. I think it makes sense to worry about the server more when we're prepping the API for v1.0. But at the moment we're still refactoring the host and renter.

PR #1038, opened 2016-03-25
Explorer Database

lukechampine:
I know this was gonna be Jordan's job, but I got tired of restarting the explorer and banged it out in two days. Sorry.

Previously, `siad -M cge` consumed about 2GB of memory, causing it to crash on many systems. Notably, it was crashing http://explore.sia.tech, filling up both its 2GB physical RAM and 2GB swap.

As of this PR, the explorer database now resides on disk, in a bolt database. `siad -M cge` consumes less than 10MB of RAM (!!). I found out that `htop` may be misleading for judging memory usage; it seems to include the mmap overhead of bolt. (The explorer database is about 2GB.) `pmap -x $(pidof siad)` gives more accurate results.

I tried to preserve the old logic as much as possible, but it's quite likely that a discrepancy or two has snuck in. I know for sure that the difficulty calculation has changed, and likely the estimated hashrate as well. I would appreciate some assistance in calculating those values correctly. The way we handle FileContractDiffs is pretty gross too. My gut says there's a much better way to do it.
The only major changes were to block facts. The old explorer kept a set of "current" facts, but the new model just maps each block to its facts. This broke one test (see TODO). Also, `blockFacts` now embeds a `modules.BlockFacts`. They only difference between the two is that `blockFacts` has an extra `Timestamp` field, and `modules.BlockFacts` has a redundant `Difficulty` field (can be calculated from the target).

Finally, I need to add code to handle subscription errors. Restarting from 0 seems sensible, but the existing database needs to be wiped first.

Oh yeah -- no new tests added. :I
I've run it locally and confirmed that it functions as expected, but obviously some rigorous testing would be better.

ALSO: I think I may have identified what was "pushing the explorer over the edge" and causing it to crash. If you ask it about the zero hash, it will report every transaction containing the zero hash, which is... a lot. Even with the DB, memory and CPU usage spike for multiple minutes.
We could triage this by intercepting the zero hash in the API, but I'm not sure what the best behavior is long-term.

### DIFF
--- a/api/explorer.go
+++ b/api/explorer.go
@@ -317,6 +317,13 @@ func (srv *Server) explorerHashHandler(w http.ResponseWriter, req *http.Request,
 		return
 	}
 
+	// TODO: lookups on the zero hash are too expensive to allow. Need a
+	// better way to handle this case.
+	if hash == (types.UnlockHash{}) {
+		writeError(w, "can't lookup the empty unlock hash", http.StatusBadRequest)
+		return
+	}
+
 	// Try the hash as an unlock hash. Unlock hash is checked last because
 	// unlock hashes do not have collision-free guarantees. Someone can create
 	// an unlock hash that collides with another object id. They will not be
--- a/modules/explorer/database.go
+++ b/modules/explorer/database.go
@@ -4,7 +4,6 @@ import (
 	"errors"
 
 	"github.com/NebulousLabs/Sia/encoding"
-	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/types"
 
 	"github.com/NebulousLabs/bolt"
@@ -152,38 +151,13 @@ func dbGetSiafundOutputTxnIDs(id types.SiafundOutputID, ids *[]types.Transaction
 
 // Internal bucket getters/setters
 
-// Set/Get block height
-func dbSetInternalBlockHeight(height types.BlockHeight) func(*bolt.Tx) error {
+func dbSetInternal(key []byte, val interface{}) func(*bolt.Tx) error {
 	return func(tx *bolt.Tx) error {
-		return tx.Bucket(bucketInternal).Put(internalBlockHeight, encoding.Marshal(height))
+		return tx.Bucket(bucketInternal).Put(key, encoding.Marshal(val))
 	}
 }
-func dbGetInternalBlockHeight(height *types.BlockHeight) func(*bolt.Tx) error {
+func dbGetInternal(key []byte, val interface{}) func(*bolt.Tx) error {
 	return func(tx *bolt.Tx) error {
-		return encoding.Unmarshal(tx.Bucket(bucketInternal).Get(internalBlockHeight), height)
-	}
-}
-
-// Set/Get recent change
-func dbSetInternalRecentChange(id modules.ConsensusChangeID) func(*bolt.Tx) error {
-	return func(tx *bolt.Tx) error {
-		return tx.Bucket(bucketInternal).Put(internalRecentChange, encoding.Marshal(id))
-	}
-}
-func dbGetInternalRecentChange(id *modules.ConsensusChangeID) func(*bolt.Tx) error {
-	return func(tx *bolt.Tx) error {
-		return encoding.Unmarshal(tx.Bucket(bucketInternal).Get(internalRecentChange), id)
-	}
-}
-
-// Set/Get difficulty
-func dbSetInternalDifficulty(difficulty types.Target) func(*bolt.Tx) error {
-	return func(tx *bolt.Tx) error {
-		return tx.Bucket(bucketInternal).Put(internalDifficulty, encoding.Marshal(difficulty))
-	}
-}
-func dbGetInternalDifficulty(difficulty *types.Target) func(*bolt.Tx) error {
-	return func(tx *bolt.Tx) error {
-		return encoding.Unmarshal(tx.Bucket(bucketInternal).Get(internalDifficulty), difficulty)
+		return encoding.Unmarshal(tx.Bucket(bucketInternal).Get(key), val)
 	}
 }
--- a/modules/explorer/database.go
+++ b/modules/explorer/database.go
@@ -14,18 +14,25 @@ var (
 
 	// database buckets
 	bucketBlockFacts            = []byte("BlockFacts")
-	bucketBlockHashes           = []byte("BlockHashes")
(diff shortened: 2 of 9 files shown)


modules/explorer/info.go
```diff
@@ -23,56 +24,27 @@ func (e *Explorer) Block(id types.BlockID) (types.Block, types.BlockHeight, bool
...
+	block, exists := e.cs.BlockAtHeight(height)
+	if !exists {
 		return modules.BlockFacts{}, false
 	}
 
-	// Grab the stats and return the facts.
-	bf := e.historicFacts[height]
-	var maturityTimestamp types.Timestamp
-	if height > types.MaturityDelay {
-		maturityTimestamp = e.historicFacts[height-types.MaturityDelay].timestamp
```
lukechampine: this special maturity logic was lost. Can you explain it?

modules/explorer/persist.go
```diff
@@ -11,5 +21,39 @@ func (e *Explorer) initPersist() error {
...
+		}
+		for _, b := range buckets {
+			_, err := tx.CreateBucketIfNotExists(b)
+			if err != nil {
+				return err
+			}
+		}
+		return nil
+	})
+
```
VoidingWarranties: Check error returned from `db.Update`?

modules/explorer/persist.go
```diff
@@ -11,5 +24,59 @@ func (e *Explorer) initPersist() error {
...
+			}
+			err := b.Put(d.key, d.val)
+			if err != nil {
+				return err
+			}
+		}
+
+		return nil
+	})
+
```
VoidingWarranties: Check error returned from `db.Update`?

api/explorer.go
```diff
@@ -317,6 +317,13 @@ func (srv *Server) explorerHashHandler(w http.ResponseWriter, req *http.Request,
 		return
 	}
 
+	// TODO: lookups on the zero hash are too expensive to allow. Need a
+	// better way to handle this case.
+	if hash != (types.UnlockHash{}) {
```
VoidingWarranties: Typo? Doesn't this ban everything *but* the zero hash?

VoidingWarranties:
Instead of banning the zero hash we could ban any lookup that would result in a very large number of results. Naively we could do:
```go
if bucket.Stats().KeyN > maxNumResults {
    return errBucketTooLarge
}
err = bucket.ForEach(func(k, v []byte)  error {
//...
})

```

for every occurrence of `bucket.ForEach`. Note sure if this is the best way, but it has the benefit of preventing someone from bringing down our explorer by creating many transactions with any particular unlock hash, not just the zero hash. `maxNumResults` is also then configurable.

api/explorer.go
lukechampine: whoops, good catch

DavidVorick:
I was thinking we should just throw the explorer onto a more powerful machine, give it 4GB of RAM and a more powerful CPU. I'm guessing you know this, but I'm not really ready for this PR, as we've got a bunch of other stuff to do, and this is definitely increasing our bug surface area before we've fixed all of the other bugs.

As a team, we need to stop implementing things until we've got better repo-wide testing.

lukechampine:
If I can fix the hashrate estimation, I'd at least like to run this on the staging site. It consumes so little memory that it shouldn't affect the stability of the current explorer.

lukechampine:
hashrate calculation should be more accurate now. It's still not identical to the current explorer though:

| Height | Old | New | Error |
|:---------|:------|:------|--------|
| 10000 | 560350946316 | 573642705124 | +2.372% |
| 20000 | 666536537025 | 676929199543 | +1.559% |
| 30000 | 339321673650 | 347801577794 | +2.499% |
| 40000 | 436082194618 | 445230600687 | +2.098% |

Could be as simple as an off-by-one error in the estimation window.

VoidingWarranties:
I believe the discrepancy is caused by the conditionals wrapping the old explorer hash rate logic:
```go
if e.blockchainHeight > hashrateEstimationBlocks {
    //...
}
```
The hash rates for the first `hashrateEstimationBlocks` are never added to the hash rate.

VoidingWarranties:
The difficulty calculation now matches the old explorer exactly, yes?

api/explorer.go
```diff
@@ -317,6 +317,13 @@ func (srv *Server) explorerHashHandler(w http.ResponseWriter, req *http.Request,
 		return
 	}
 
+	// TODO: lookups on the zero hash are too expensive to allow. Need a
+	// better way to handle this case.
+	if hash == (types.UnlockHash{}) {
```
VoidingWarranties: Why are there parens around `types.UnlockHash{}`?

modules/explorer/database.go
```diff
@@ -0,0 +1,105 @@
...
+//
+//   var height types.BlockHeight
+//   db.View(func(tx *bolt.Tx) error {
+//       bytes := tx.Bucket(bucketBlockIDs).Get(encoding.Marshal(id))
+//       return encoding.Unmarshal(bytes, &height)
+//   })
+
+// dbGetAndDecode returns a 'func(*bolt.Tx) error' that retrieves and decodes
+// a value from the specified bucket. If the value does not exist,
+// dbGetAndDecode returns errNotExist.
```
VoidingWarranties: *the function returned by dbGetAndDecode returns errNotExist.

modules/explorer/database.go
```diff
@@ -0,0 +1,105 @@
...
+		if valBytes == nil {
+			return errNotExist
+		}
+		return encoding.Unmarshal(valBytes, val)
+	}
+}
+
+// dbGetTransactionIDSet returns a 'func(*bolt.Tx) error' that decodes a
+// bucket of transaction IDs into a slice. If the bucket is nil,
+// dbGetTransactionIDSet returns errNotExist.
```
VoidingWarranties: Same critique as `dbGetAndDecode`

modules/explorer/database.go
VoidingWarranties: In the context of the other functions that return a `func(*bolt.Tx) error` this is clear and just the comment more wordy.

api/explorer.go
lukechampine: it's one of the few ambiguities in Go's grammar.

DavidVorick:
Can we change the estimated hashrate window from 60 blocks to 200 blocks?

DavidVorick:
ok. Good enough.

PR #1044, opened 2016-03-29
Negate siafund pool diff direction instead of hardcoding revert

VoidingWarranties:
From the lines above I think this was a typo. My understanding of consensus is lacking so I'm not sure what the impact of this bug was, if it even was a bug.

### DIFF
--- a/modules/consensus/subscribe.go
+++ b/modules/consensus/subscribe.go
@@ -42,7 +42,7 @@ func (cs *ConsensusSet) computeConsensusChange(tx *bolt.Tx, ce changeEntry) (cc 
 		}
 		for i := len(revertedBlock.SiafundPoolDiffs) - 1; i >= 0; i-- {
 			sfpd := revertedBlock.SiafundPoolDiffs[i]
-			sfpd.Direction = modules.DiffRevert
+			sfpd.Direction = !sfpd.Direction
 			cc.SiafundPoolDiffs = append(cc.SiafundPoolDiffs, sfpd)
 		}
 	}


DavidVorick:
now the hard part. Can you write a test that failed before the change and succeeded after the change? That might be something that I need to do.

I'm thinking you'd create some reorg (using the _reorg.go file) that has a bunch of siafund pool diffs, and then you'd have some made up module that uses the diffs to keep track of the siafund pool size, and then you'd test that the siafund pool size according to the made up module matches the siafund pool size in the consensus set.

DavidVorick:
Actually, we might want to make a more broad test for the consensus change system that does that with all of the values. Again, that's probably more my responsibility than yours.

DavidVorick:
Even more, we should run that fake module in debug mode.

DavidVorick:
Closing this due to lack of bandwidth, can make an issue for more testing if you want, but I'm pretty confident at the moment that the code here is correct.

VoidingWarranties:
@DavidVorick we talked offline about how siafund pool diffs are different than other diffs. Can you explain this a bit more?

DavidVorick:
Most diffs indicate that you should either delete or create an object. So, in the usual case, when going forward you follow the instructions (create or delete), and when going backwards you follow the reverse of the instructions (delete or create).

In the case of the siafund pool, you are replacing the old siafund pool value with the new siafund pool value. I think that the way this is coded has the siafund pool diff always being added such that you should just apply it.

It's poorly commented code, I'm sorry about that. Actual testing will be needed to be certain that there is or is not a bug, and that testing should be added at some point soon.

PR #1045, opened 2016-03-29
NetAddress fixup

### LINKED ISSUE #1040: Hard coded IPV6 loop back address
In the Gatway's sendAddress method.

VoidingWarranties:
Cleans up the NetAddress type. The main change is adding to the `IsValid()` method, which checks whether a NetAddress is a valid address. Valid is defined as being of the form "host:port" such that "host" is not a loopback address (except in testing) nor an unspecified IP address, and such that neither the host nor port are blank, contain white space, or contain the null character.

Previously some methods allowed loopback addresses in testing while others did not. This change makes loopback addresses always valid in testing only. This allows us to get rid of some kludges in the gateway testing.

Fixes #1040

### DIFF
--- a/api/host.go
+++ b/api/host.go
@@ -105,7 +105,11 @@ func (srv *Server) hostHandlerPOST(w http.ResponseWriter, req *http.Request, _ h
 			}
 		}
 	}
-	srv.host.SetInternalSettings(settings)
+	err := srv.host.SetInternalSettings(settings)
+	if err != nil {
+		writeError(w, err.Error(), http.StatusBadRequest)
+		return
+	}
 	writeSuccess(w)
 }
 
--- a/modules/gateway/gateway_test.go
+++ b/modules/gateway/gateway_test.go
@@ -17,11 +17,6 @@ func newTestingGateway(name string, t *testing.T) *Gateway {
 	if err != nil {
 		t.Fatal(err)
 	}
-	// Manually add myAddr as a node. This is necessary because g.addNode
-	// rejects loopback addresses.
-	id := g.mu.Lock()
-	g.nodes[g.myAddr] = struct{}{}
-	g.mu.Unlock(id)
 	return g
 }
 
--- a/modules/gateway/nodes.go
+++ b/modules/gateway/nodes.go
@@ -117,7 +117,7 @@ func (g *Gateway) relayNode(conn modules.PeerConn) error {
 // sendAddress is the calling end of the RelayNode RPC.
 func (g *Gateway) sendAddress(conn modules.PeerConn) error {
 	// don't send if we aren't connectible
-	if g.Address().Host() == "::1" {
+	if g.Address().IsLoopback() {
 		return errors.New("can't send address without knowing external IP")
 	}
 	return encoding.WriteObject(conn, g.Address())
--- a/modules/gateway/nodes.go
+++ b/modules/gateway/nodes.go
@@ -75,7 +75,10 @@ func (g *Gateway) requestNodes(conn modules.PeerConn) error {
 	}
 	id := g.mu.Lock()
 	for _, node := range nodes {
-		g.addNode(node)
+		err := g.addNode(node)
+		if err != nil {
+			g.log.Printf("WARN: peer '%v' send the invalid addr '%v'", conn.RemoteAddr(), node)
+		}
 	}
 	g.save()
 	g.mu.Unlock(id)
--- a/modules/gateway/nodes.go
+++ b/modules/gateway/nodes.go
@@ -22,8 +22,8 @@ func (g *Gateway) addNode(addr modules.NetAddress) error {
 		return errors.New("node already added")
 	} else if net.ParseIP(addr.Host()) == nil {
 		return errors.New("address is not routable: " + string(addr))
-	} else if net.ParseIP(addr.Host()).IsLoopback() {
-		return errors.New("cannot add loopback address")
+	} else if addr.IsValid() != nil {
+		return errors.New("address is not valid: " + string(addr))
 	}
 	g.nodes[addr] = struct{}{}
 	return nil
@@ -117,7 +117,7 @@ func (g *Gateway) relayNode(conn modules.PeerConn) error {
 // sendAddress is the calling end of the RelayNode RPC.
 func (g *Gateway) sendAddress(conn modules.PeerConn) error {
 	// don't send if we aren't connectible
-	if g.Address().IsLoopback() {
+	if g.Address().IsValid() != nil {
 		return errors.New("can't send address without knowing external IP")
 	}
 	return encoding.WriteObject(conn, g.Address())
--- a/modules/host/upnp.go
+++ b/modules/host/upnp.go
@@ -66,7 +66,7 @@ func (h *Host) managedLearnHostname() {
 	defer h.mu.Unlock()
 	autoAddress := modules.NetAddress(net.JoinHostPort(hostname, h.port))
 	if err := autoAddress.IsValid(); err != nil {
-		h.log.Printf("WARN: discovered net address '%v' is invalid: %v", autoAddress, err)
+		h.log.Printf("WARN: discovered hostname %q is invalid: %v", autoAddress, err)
 		return
 	}
 	if autoAddress == h.autoAddress && h.announced {
--- a/modules/host/upnp.go
+++ b/modules/host/upnp.go
@@ -65,6 +65,10 @@ func (h *Host) managedLearnHostname() {
 	h.mu.Lock()
 	defer h.mu.Unlock()
 	autoAddress := modules.NetAddress(net.JoinHostPort(hostname, h.port))
+	if err := autoAddress.IsValid(); err != nil {
+		h.log.Printf("WARN: discovered net address '%v' is invalid: %v", autoAddress, err)
+		return
+	}
 	if autoAddress == h.autoAddress && h.announced {
 		// Nothing to do - the auto address has not changed and the previous
 		// annoucement was successful.
--- a/modules/netaddress.go
+++ b/modules/netaddress.go
@@ -62,13 +62,25 @@ func (na NetAddress) isLoopback() bool {
 // testing). Valid IPv4 addresses, IPv6 addresses, and hostnames are detailed
 // in RFCs 791, 2460, and 952, respectively.
 func (na NetAddress) IsValid() error {
-	if build.Release != "testing" && na.isLoopback() {
-		return errors.New("host is a loopback address")
-	}
-
 	host, port, err := net.SplitHostPort(string(na))
 	if err != nil {
 		return err
+	}
+
+	portInt, err := strconv.Atoi(port)
+	if err != nil {
+		return errors.New("port is not an integer")
+	} else if portInt < 1 || portInt > 65535 {
+		return errors.New("port is invalid")
+	}
+
+	// This check must come after the valid port check so that a host such as
+	// "localhost:badport" will fail.
+	if na.isLoopback() {
+		if build.Release == "testing" {
+			return nil
+		}
+		return errors.New("host is a loopback address")
 	}
 
 	// First try to parse host as an IP address; if that fails, assume it is a
@@ -78,12 +90,24 @@ func (na NetAddress) IsValid() error {
 			return errors.New("host is the unspecified address")
 		}
 	} else {
+		// Hostnames can have a trailing dot (which indicates that the hostname is
+		// fully qualified), but we ignore it for validation purposes.
+		if strings.HasSuffix(host, ".") {
+			host = host[:len(host)-1]
+		}
 		if len(host) < 1 || len(host) > 253 {
 			return errors.New("invalid hostname length")
 		}
-		for _, label := range strings.Split(host, ".") {
+		labels := strings.Split(host, ".")
+		if len(labels) == 1 {
+			return errors.New("unqualified hostname")
+		}
+		for _, label := range labels {
 			if len(label) < 1 || len(label) > 63 {
 				return errors.New("hostname contains label with invalid length")
+			}
+			if strings.HasPrefix(label, "-") || strings.HasSuffix(label, "-") {
+				return errors.New("hostname contains label that starts or ends with a hyphen")
 			}
 			for _, r := range strings.ToLower(label) {
 				isLetter := 'a' <= r && r <= 'z'
@@ -94,13 +118,6 @@ func (na NetAddress) IsValid() error {
 				}
 			}
 		}
-	}
-
-	portInt, err := strconv.Atoi(port)
-	if err != nil {
-		return errors.New("port is not an integer")
-	} else if portInt < 1 || portInt > 65535 {
-		return errors.New("port is invalid")
 	}
 
 	return nil
--- a/modules/netaddress.go
+++ b/modules/netaddress.go
@@ -3,7 +3,8 @@ package modules
 import (
 	"errors"
 	"net"
-	"unicode"
+	"strconv"
+	"strings"
 
 	"github.com/NebulousLabs/Sia/build"
 )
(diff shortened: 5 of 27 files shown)


modules/gateway/nodes.go
```diff
@@ -117,7 +117,7 @@ func (g *Gateway) relayNode(conn modules.PeerConn) error {
 // sendAddress is the calling end of the RelayNode RPC.
 func (g *Gateway) sendAddress(conn modules.PeerConn) error {
 	// don't send if we aren't connectible
-	if g.Address().Host() == "::1" {
+	if g.Address().IsLoopback() {
```
VoidingWarranties: We probably want to allow sending a loopback address on dev / testing releases so that we can have multiple nodes on the same machine.

VoidingWarranties: Although at the moment `Gateway.learnHostname()` is responsible for calling `g.RegisterConnectCall("RelayNode", g.sendAddress)`, which it never does during testing.

modules/netaddress.go
```diff
@@ -36,10 +36,10 @@ func (na NetAddress) IsLoopback() bool {
 	}
 
 	host := na.Host()
-	if ip := net.ParseIP(host); ip != nil && ip.IsLoopback() {
+	if ip := net.ParseIP(host); ip != nil && (ip.IsLoopback() || ip.IsUnspecified()) {
```
VoidingWarranties: This is confusing as a caller of `IsLoopback` would not expect it to return true for unspecified addresses. `IsLoopback` seems to be used as a check for validity. It seems to me it would be better to add
```go
ip := net.ParseIP(na.Host())
if ip.IsLoopback() || ip.IsUnspecified{
    return false
}
```
to the `NetAddress.IsValid` method.

lukechampine: agreed

modules/netaddress.go
```diff
@@ -36,10 +36,10 @@ func (na NetAddress) IsLoopback() bool {
...
 		return true
 	}
-	if host == "localhost" || host == "::" || host == "0.0.0.0" {
+	if host == "localhost" {
```
DavidVorick: we had these here because of explicit quirks found in production. There should have been a comment explaining that, but you can't remove these like that, you probably want to replace them with a function.

We should also have a regression test for these changes.

modules/netaddress.go
DavidVorick: I would like to see this cleaned up in this PR

DavidVorick:
I would like to see testing to handle all of the cases that get changed in this PR.

modules/netaddress.go
VoidingWarranties: They are replaced with the `ip.IsUnspecified()` call a few lines above.

DavidVorick:
Per-discussion, I think this PR should be exchanged for a `modules.NewNetAddress() error`, and the call to 'New' can do all of the error checking. It should clean things up quite a bit and add a bunch of safety.

VoidingWarranties:
So I didn't implement the `NewNetAddress` func because it in order to encourage its use you would need to make `NetAddress` a struct with a single, non exported, string field. However, we can't do this because it would break compatibility with RPCs that send `NetAddress`es and any other places that decode encoded `NetAddress`es.

So if we keep the type the same it would be up to the developer to remember to use `NewNetAddress` instead of simple assignment. But the developer could just as easily remember to check the `IsValid` method instead.

To make conditionally allowing loopback addresses during testing I made `IsValid` return an error instead of a bool. Now instead of
```go
if !addr.IsValid() {
    if build.Release != "testing" || !addr.IsLoopback() {
        return
    }
}
```
you can do
```go
    if build.Release != "testing" || err != modules.ErrLoopbackAddr {
        return
    }
}
```

Really you should never be checking `IsLoopback()` and should use the return value of `IsValid()` instead so that you catch other errors. This is especially true since previously `IsLoopback()` was used to check for the unspecified address, which is now only checked in `IsValid`.

DavidVorick:
> However, we can't do this because it would break compatibility with RPCs that send NetAddresses and any other places that decode encoded NetAddresses.

Pretty sure that 

`type NetAddress string`

and

`type NetAddress struct {
    addr string
}`

will encode to exactly the same bits

DavidVorick:
oh, unless we are using json

lukechampine:
regardless, we could write marshalling/unmarshalling methods to ensure compatibility.

VoidingWarranties:
A malicious peer would still be able to send a malformed address so we have to check `IsValid` anyways.

lukechampine:
you can do decode-time validation in the unmarshalling method.
```go
func (na *NetAddress) UnmarshalJSON(b []byte) error {
    var addr string
    err := json.Unmarshal(b, &addr)
    if err != nil {
        return err
    }
    *na, err = NewNetAddress(addr)
    return err
}
```

VoidingWarranties:
Using `NewNetAddress` would force us to always allow (or disallow) loopback addresses during testing. In some places we currently disallow loopback addresses during testing and in others we allow it.

DavidVorick:
Consistency would be a good thing

DavidVorick:
Are we planning on closing this PR and replacing it with the netaddress stuff?

modules/netaddress_test.go
```diff
@@ -1,34 +1,91 @@
...
+		// Valid addresses.
+		"foo:bar",
+		"hn.com:8811",
+		"[::2]:bar",
+		"111.111.111.111:111",
+		"12.34.45.64:7777",
+		"世界:bar",
+		"bar:世界",
+		"世:界",
+		`":"`, // yeah, okay
```
lukechampine: ಠ_ಠ

seems easier to me to call `strconv.Atoi` on the port.
```go
portInt, err := strconv.Atoi(port)
if err != nil || portInt < 0 || portInt > 65535 {
    return errors.New("invalid port")
}
```
this would catch unicode stuff too

modules/netaddress.go
```diff
@@ -26,27 +33,60 @@ func (na NetAddress) Port() string {
...
+
+	for _, runeValue := range host {
+		if unicode.IsSpace(runeValue) || runeValue == 0 {
+			return errors.New("host has invalid characters")
+		}
+	}
+	if host == "" {
+		return errors.New("host is blank")
+	}
+	if ip := net.ParseIP(host); ip != nil && ip.IsUnspecified() {
```
lukechampine: `net.ParseIP` should also catch unicode/blank errors

modules/netaddress.go
```diff
@@ -26,27 +33,60 @@ func (na NetAddress) Port() string {
...
+	}
+
+	for _, runeValue := range port {
+		if unicode.IsSpace(runeValue) || runeValue == 0 {
+			return errors.New("port has invalid characters")
+		}
+	}
+	if port == "" {
+		return errors.New("port is blank")
+	}
```
lukechampine: seems easier to me to call `strconv.Atoi` on the port.
```go
portInt, err := strconv.Atoi(port)
if err != nil || portInt < 0 || portInt > 65535 {
    return errors.New("invalid port")
}
```
this would catch unicode stuff too.

lukechampine:
From Wikipedia:

>Hostnames are composed of series of labels concatenated with dots, as are all domain names. For example, "en.wikipedia.org" is a hostname. Each label must be between 1 and 63 characters long,[2] and the entire hostname (including the delimiting dots but not a trailing dot) has a maximum of 253 ASCII characters.[3]
>
>The Internet standards (Requests for Comments) for protocols mandate that component hostname labels may contain only the ASCII letters 'a' through 'z' (in a case-insensitive manner), the digits '0' through '9', and the hyphen ('-'). The original specification of hostnames in RFC 952, mandated that labels could not start with a digit or with a hyphen, and must not end with a hyphen. However, a subsequent specification (RFC 1123) permitted hostname labels to start with digits. No other symbols, punctuation characters, or white space are permitted.

I pushed a commit that implements these changes.

VoidingWarranties:
Thanks for the addition. Should we even allow unqualified hostnames?

DavidVorick:
Is there a reason that we need to? Probably no reason to allow them.

lukechampine:
I believe `localhost` is an unqualified hostname.

VoidingWarranties:
True, but I think it's the only unqualified hostname that should be allowed. It makes the implementation *slightly* uglier, but I think it's still reasonably clean.

modules/netaddress_test.go
```diff
@@ -45,20 +50,27 @@ var (
 		// Unspecified address
 		"[::]:bar",
 		"0.0.0.0:bar",
+		// Invalid hostnames
+		"unqualifiedhost:123",
+		"Yo-Amazon-we-are-really-happy-for-you-and-we-will-let-you-finish-but-Sia-is-the-best-cloud-storage-of-all-time-of-all-time-of-all-time-of-all-time-of-all-time-of-all-time-of-all-time-of-all-time-of-all-time-of-all-time-of-all-time-of-all-time-of-all-time:123",
```
lukechampine: `foo-bar` is sufficient, lol

modules/netaddress.go
```diff
@@ -81,7 +94,14 @@ func (na NetAddress) IsValid() error {
 		if len(host) < 1 || len(host) > 253 {
 			return errors.New("invalid hostname length")
 		}
-		for _, label := range strings.Split(host, ".") {
+		if host[0] == '-' {
+			return errors.New("hostnames cannot start with a hyphen")
```
lukechampine: as per the RFC, labels can't begin *or end* with a hyphen. I would use `if strings.HasPrefix(host, "-") || strings.HasSuffix(host, "-")`

lukechampine:
Requiring qualification broke some other tests, probably because they used hostnames like `"foo"`

modules/netaddress_test.go
```diff
@@ -1,34 +1,111 @@
...
+		"世界:123",
+		"bar:世界",
+		"世:界",
+		`":"`,
+		// Unspecified address
+		"[::]:bar",
+		"0.0.0.0:bar",
+		// Invalid hostnames
+		"unqualifiedhost:123",
+		"Yo-Amazon-we-are-really-happy-for-you-and-we-will-let-you-finish-but-Sia-is-the-best-cloud-storage-of-all-time-of-all-time-of-all-time-of-all-time-of-all-time-of-all-time-of-all-time-of-all-time-of-all-time-of-all-time-of-all-time-of-all-time-of-all-time:123",
```
lukechampine: `foo-bar` is sufficient, lol

modules/netaddress.go
```diff
@@ -26,27 +34,87 @@ func (na NetAddress) Port() string {
...
+		if ip.IsUnspecified() {
+			return errors.New("host is the unspecified address")
+		}
+	} else {
+		if len(host) < 1 || len(host) > 253 {
+			return errors.New("invalid hostname length")
+		}
+		if host[0] == '-' {
+			return errors.New("hostnames cannot start with a hyphen")
+		}
```
lukechampine: as per the RFC, labels can't begin *or end* with a hyphen. I would use `if strings.HasPrefix(host, "-") || strings.HasSuffix(host, "-")`

modules/netaddress_test.go
DavidVorick: testing the character limit

DavidVorick: I like it lol

VoidingWarranties:
Rebased and fixed up the previous commit to properly check for leading or trailing hyphens in hostname labels.

modules/gateway/gateway_test.go
```diff
@@ -17,11 +17,6 @@ func newTestingGateway(name string, t *testing.T) *Gateway {
 	if err != nil {
 		t.Fatal(err)
 	}
-	// Manually add myAddr as a node. This is necessary because g.addNode
-	// rejects loopback addresses.
-	id := g.mu.Lock()
-	g.nodes[g.myAddr] = struct{}{}
-	g.mu.Unlock(id)
```
lukechampine: makes me think: wherever we have a comment like: "this is necessary because," it would be ideal to include an assertion that it is indeed the case. In this case, we would assert that `g.nodes` does not already contain `g.myAddr`.

modules/host/upnp.go
```diff
@@ -65,6 +65,10 @@ func (h *Host) managedLearnHostname() {
 	h.mu.Lock()
 	defer h.mu.Unlock()
 	autoAddress := modules.NetAddress(net.JoinHostPort(hostname, h.port))
+	if err := autoAddress.IsValid(); err != nil {
+		h.log.Printf("WARN: discovered net address '%v' is invalid: %v", autoAddress, err)
```
lukechampine: for future reference, you can use `%q` to specify a quoted string.

modules/gateway/gateway_test.go
VoidingWarranties: Good idea. Although in this specific case I think the assertion would be slightly different. It's not that we are assuming `g.nodes` contains `g.myAddr` (it shouldn't, this was just a kludge), but rather we were assuming `addNode` would reject loopback addresses.
```go
if err := g.addNode("localhost:123"); err == nil {
    t.Fatal("assumption invalidated - addNode accepted a loopback address:", err)
}
```

VoidingWarranties:
Added address validation to API calls, fixed up a few bugs in testing, and rebased. Should be good to go.

api/gateway.go
```diff
@@ -41,6 +45,10 @@ func (srv *Server) gatewayAddHandler(w http.ResponseWriter, req *http.Request, p
 // gatewayRemoveHandler handles the API call to remove a peer from the gateway.
 func (srv *Server) gatewayRemoveHandler(w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
 	addr := modules.NetAddress(ps.ByName("netaddress"))
+	if err := addr.IsValid(); err != nil {
+		writeErorr(w, "invalid address: "+err.Error(), http.StatusBadRequest)
+		return
+	}
```
lukechampine: these seem unnecessary. `Connect` will fail if the address is not dialable, and `Disconnect` will fail if the address is not in the peer list. I don't see why further validation is necessary for either.

VoidingWarranties: I think it's simpler to do validation as soon as the address is created, that way validation isn't required in every method that takes a NetAddress as an argument. In this case it is true that the call will fail because the address isn't dialable, but I don't think we can always rely on this behavior unless we enforce that all exported methods that take a NetAddress also validate it.

I'm not sure which is better. What do you think?

As a side note if you pass an un-dialable NetAddress to `Connect` it will add it to it's peer list before it returns an error. This should probably be fixed regardless.

VoidingWarranties: Oops, didn't mean to make this convo collapse when I fixed the typo.

lukechampine: Validation is not required in many places. It is required only when an address is going to be stored previously being dialed. AFAIK, that only occurs in two places: in creating/reading host announcements, and in the gateway's node sharing.
Regardless, I don't think validation should be performed in the API package. The API package strives to be a lightweight wrapper around the module objects. It should introduce as little additional logic as possible.

>As a side note if you pass an un-dialable NetAddress to Connect it will add it to it's peer list before it returns an error.

I don't think this is the case. Maybe you were confused by these lines:
```go
	id := g.mu.RLock()
	_, exists := g.peers[addr]
	g.mu.RUnlock(id)
	if exists {
		return errors.New("peer already added")
	}
```

VoidingWarranties: Oops, you were right. I forgot that map access does not insert an object there with the zero value like it does in C++. Old habits.

VoidingWarranties: >Regardless, I don't think validation should be performed in the API package. The API package strives to be a lightweight wrapper around the module objects. It should introduce as little additional logic as possible.

What about validation for host announce with specified address? Should we validate the address in `host.AnnounceAddress`?

VoidingWarranties:
Removed address validation from API calls, and added validation to `host.SetInternalSettings`.

api/host.go
```diff
@@ -105,7 +105,12 @@ func (srv *Server) hostHandlerPOST(w http.ResponseWriter, req *http.Request, _ h
 			}
 		}
 	}
-	srv.host.SetInternalSettings(settings)
+	err := srv.host.SetInternalSettings(settings)
+	if err != nil {
+		// TODO: error returned is not necessarily because of a bad request.
+		writeError(w, err.Error(), http.StatusBadRequest)
```
lukechampine: this is a common problem in our API, though the distinction is not very important. I don't think anyone checks the exact 400 code; only whether it's 200 or not.

VoidingWarranties: I don't see an easy solution either. Oh well.

modules/netaddress.go
```diff
@@ -15,6 +20,9 @@ func (na NetAddress) Host() string {
 	if err != nil {
 		return ""
 	}
+	if na.IsValid() != nil {
+		return ""
+	}
```
lukechampine: hmm, not sure about these. `Host` and `Port` are just convenience wrappers more than rigorous methods. It's sort of implicit that the netaddress is valid.

VoidingWarranties: I added this check to discourage use of partial addresses. Like before how we did hostname discovery by initially setting the NetAddress to `localhost:port` and then later modifying only the hostname. This seemed to be the cause of a few bugs so I figured discouraging it was worth it.

If you think they shouldn't be there I'll remove them.

PR #1046, opened 2016-03-30
Fix error handling in computeConsensusChange

VoidingWarranties:
Named return values obfuscated errors like this. Named return values are used extensively in the consensus module so it might be a good idea to go through and remove all named return values.

### DIFF
--- a/modules/consensus/subscribe.go
+++ b/modules/consensus/subscribe.go
@@ -1,7 +1,6 @@
 package consensus
 
 import (
-	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/modules"
 
 	"github.com/NebulousLabs/bolt"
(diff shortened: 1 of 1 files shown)


VoidingWarranties:
Bump

modules/consensus/subscribe.go
```diff
@@ -1,20 +1,22 @@
...
 	"github.com/NebulousLabs/bolt"
 )
 
 // computeConsensusChange computes the consensus change from the change entry
 // at index 'i' in the change log. If i is out of bounds, an error is returned.
-func (cs *ConsensusSet) computeConsensusChange(tx *bolt.Tx, ce changeEntry) (cc modules.ConsensusChange, err error) {
-	cc.ID = ce.ID()
+func (cs *ConsensusSet) computeConsensusChange(tx *bolt.Tx, ce changeEntry) (modules.ConsensusChange, error) {
+	cc := modules.ConsensusChange{
+		ID: ce.ID(),
```
DavidVorick: was it just not computing the id before?

DavidVorick: 0.o

VoidingWarranties: No, this part is functionally the same.

PR #1050, opened 2016-04-05
Don't fsync during testing

VoidingWarranties:
as it can take too long and causes tests to fail in Travis

### DIFF
--- a/persist/log.go
+++ b/persist/log.go
@@ -26,9 +26,13 @@ func (cf *closeableFile) Close() error {
 	if cf.closed {
 		build.Critical("cannot close the file; already closed")
 	}
-	// Ensure that all data has actually hit the disk.
-	if err := cf.Sync(); err != nil {
-		return err
+	// Don't sync during testing as it can take too long and causes tests to fail.
+	// TODO: mock the os package so that Sync() is a no-op during testing.
+	if build.Release != "testing" {
+		// Ensure that all data has actually hit the disk.
+		if err := cf.Sync(); err != nil {
+			return err
+		}
 	}
 	cf.closed = true
 	return cf.File.Close()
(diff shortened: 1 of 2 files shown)


DavidVorick:
This change makes me very nervous, because I'm not sure what code relies on the fsync happening.

@lukechampine can you think of any examples where this might cause problems, or do you think it's mostly a safe change?

I do agree that fsync should be skipped during testing for most cases, but this skips fsync in all cases, and I'm not sure if that's going to cause issues in some packages or potentially add footguns down the road.

lukechampine:
I don't foresee this causing any problems. I'm not even sure if we could write a test that checked whether a given file had been `Fsync`'d or not. By design, the filesystem tries to make that behavior invisible.
`Fsync` is basically just a way for us to cover our ass when writing really important information to disk. It should be used sparingly, because it incurs a significant performance penalty when writing to e.g. a network drive. So I'm in favor of this change.

lukechampine:
Related reading: Theodore T'so [claims](http://thunk.org/tytso/blog/2009/03/15/dont-fear-the-fsync/) that the performance impact shouldn't be super noticeable for "normal workloads," but I don't think running our test suite on Travis is a normal workload. It involves syncing a large number of small files. Ideally, we'd be able to prove that removing the `Fsync`s significantly reduced timeout failures. I'm not sure if this has been shown yet.

VoidingWarranties:
We could reasonably show that by re-running this PR's tests enough times without observing a panic from a short test hanging during a sync. How many times would we need to run it to be confident? I don't know. It's been happening with noticeable frequency ever since we started properly closing the gateway.

lukechampine:
The `consensus` and `gateway` tests have started timing out on my machine, but with this PR they drop below 3s again. `consensus` in particular drops by over half.

PR #1051, opened 2016-04-06
Make state-modifying API calls POSTs

VoidingWarranties:
This PR is intended to serve as an area for discussion about API improvements, not just this specific suggested improvement. As we approach v1.0.0 we need to finalize the API. The change in this PR (so far) is to make every state-modifying call a POST and all other calls a GET.

Another proposed API change we have talked about offline before is adding more HTTP verbs. DELETE in particular would be more semantic for the renter / hostdb / gateway when deleting or removing files / hosts / peers.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -49,7 +49,8 @@ func (srv *Server) initAPI() {
 	// Daemon API Calls
 	router.GET("/daemon/constants", srv.daemonConstantsHandler)
 	router.GET("/daemon/version", srv.daemonVersionHandler)
-	router.GET("/daemon/stop", srv.daemonStopHandler)
+	router.GET("/daemon/stop", srv.daemonStopHandler) // COMPATv0.5.2
+	router.POST("/daemon/stop", srv.daemonStopHandler)
 
 	// Consensus API Calls
 	if srv.cs != nil {
@@ -83,8 +84,10 @@ func (srv *Server) initAPI() {
 		router.GET("/miner", srv.minerHandler)
 		router.GET("/miner/header", srv.minerHeaderHandlerGET)
 		router.POST("/miner/header", srv.minerHeaderHandlerPOST)
-		router.GET("/miner/start", srv.minerStartHandler)
-		router.GET("/miner/stop", srv.minerStopHandler)
+		router.GET("/miner/start", srv.minerStartHandler) // COMPATv0.5.2
+		router.GET("/miner/stop", srv.minerStopHandler)   // COMPATv0.5.2
+		router.POST("/miner/start", srv.minerStartHandler)
+		router.POST("/miner/stop", srv.minerStopHandler)
 		router.GET("/miner/headerforwork", srv.minerHeaderHandlerGET)  // COMPATv0.4.8
 		router.POST("/miner/submitheader", srv.minerHeaderHandlerPOST) // COMPATv0.4.8
 	}


lukechampine:
I like the idea of more HTTP verbs in theory, but the reality is that 1) they are generally less supported than GET and POST, and 2) our architecture doesn't map perfectly onto REST, leading to lots of bikeshedding around whether a given route should be a PUT or POST, whether params should go in the path or query string or request body, etc.
My preference would be to stick with GET + POST, but perhaps add PUT and DELETE routes alongside POSTs where they are a better fit (e.g. `/files/delete`)

VoidingWarranties:
Supporting multiple verbs for the same action is a good alternative.

"Other HTTP verbs aren't as well supported" has come up before, but I have yet to come across a language that doesn't support them.

I don't think full REST is a good fit for our API as it requires many more things than just the HTTP verbs that seem unnecessary for our application, particularly the "discoverability" requirement of REST.

lukechampine:
Other verbs are indeed supported by all languages, but GET and POST are often easier to use than others. Compare:
```go
resp, err := http.Get("http://example.com/")
```
```go
req, _ := http.NewRequest(http.MethodDelete, "http://example.com/", nil)
resp, err := http.DefaultClient.Do(req)
```
Not sure whether this extends to other languages.
But yeah, no harm in supporting both.

DavidVorick:
There has been a lot of debate about what verbs to support, and there was also a lot of debate about whether to make /daemon/start+stop GET or POST. We've definitely had complaints before about other verbs not being well supported. Technically supported, but not as easily manipulated through the stdlibs. I do think this extends across most languages.

As Luke mentioned, we're not really a RESTful application, REST was built for a completely different paradigm than what we're using it for. /service/stop is not something you'd ever see on an actual internet platform. So we decided that cutting corners is okay, and that we should make the API as easy and comfortable as possible, with the emphasis being on consistency and easy-to-learn, instead of don't-need-to-learn, given that we've got some unique stuff that you're going to need to learn anyway.

We decided that POST should be used for posting data, and GET should be used for everything else. You POST a file, you POST your settings, and you use POST instead of DELETE. For all of everything else, you use GET, even if it's state changing.

NACK on this PR. I think we're largely happy with the paradigm we've chosen, acknowledging that it's imperfect but also that we're not going to get it perfect, and that REST is really not the ideal tool for the job. It's just really well supported by most languages (at least for the GET and POST verbs)

DavidVorick:
I'd also rather not have 2 ways to write the same code. "no harm in supporting both" -> more bug surface area, more places to get confused when developers are looking at eachothers code, and it's more places for other libraries to have bugs that result in users asking us to change things. Also, you'll get more things like "well, you support DELETE for this action, but why not this other action? Can you support it for both?"

In general, I'm in favor of being restrictive. `go fmt` is a great example of being restrictive in a highly beneficial way.

VoidingWarranties:
@DavidVorick can you clarify the differentiation between our uses of POST an GET a bit more? What counts as data in order for it to be a POST vs a GET? Do parameters count as data?

Specifically the endpoint `/renter/delete/*siapath` is a POST. Is the siapath considered the data in this case?

DavidVorick:
well the idea is that you're deleting a file, so the POST is about the file. I'm not sure I agree with using POST in place of DELETE, but that's the convention that's currently in place.

This is a conversation we can have when we're to the step of finalizing the API. Before that point, it's probably better not to worry about it. I'm thinking we'll be at that point in about 35 days.

Issue #1025, opened 2016-03-18
New .sia format


lukechampine:
With the transition to hash-based retrieval, we must design a new .sia format. The current format is a standard binary encoding of the `Renter.file` type:

``` go
struct {
    Name      string
    Size      uint64
    MasterKey crypto.SecretKey
    PieceSize uint64
    Mode      os.FileMode

    // generally the string "Reed-Solomon", plus data and parity integers
    ErasureCode modules.ErasureCode

    Contracts []struct {
        ID          types.FileContractID
        IP          modules.NetAddress
        WindowStart types.BlockHeight

        Pieces []struct {
                Chunk  uint64
                Piece  uint64
                Offset uint64
        }
    }
}
```

The simplest possible change would be to replace the `Offset` field with the piece's Merkle root. However, this would bloat the filesize significantly.
To reduce size, one strategy would be to store hashes in a flat array. This allows us to determine the `Chunk` and `Piece` values based on the array index. As a consequence of this, we would need to use a lookup table so that the hashes could efficiently reference their corresponding contract. The resulting format would look something like this:

``` go
struct {
    // name, encryption key, erasure code...

    Contracts []struct {
        ID          types.FileContractID
        IP          modules.NetAddress
        WindowStart types.BlockHeight
    }

    Sectors []struct {
        MerkleRoot crypto.Hash
        Contract   uint64 // indexes the Contracts field
    }
}
```

Unfortunately, this only saves 8 bytes per sector, since all we've done is replace the `Chunk` and `Piece` values with `Contract`. If we use a `uint16` for `Contract`, we save another 6 bytes, bringing the total increase down to 34/24 = 1.41.
Another opportunity for saving bytes is the `ID` field, which (IIRC) is not necessary for retrieval in the new protocol. This would save 32 bytes per contract, which works out to less than a kilobyte total for our current redundancy parameters. Since the ID is used elsewhere in the Renter (e.g. when renewing), removing it from the .sia format would mean storing it in the Renter persistence. So I'm opposed to removing the ID field.
We also need a way to deal with missing sectors. We could either use an empty Merkle root, or change the sector type to `[]*struct`, with `nil` indicating a missing sector (this adds one byte to each sector). I'm not sure which one I prefer. The latter saves lots of bytes when sectors are missing, but wastes bytes when they are not. Shouldn't .sia files be treated as though all their sectors will eventually be uploaded? I don't know.

[...]

PR #1052, opened 2016-04-08
Create dotsia package

lukechampine:
`dotsia` is a package that defines the (new) .sia format and provides functions for encoding/decoding them.

The full specification is in format.go. Here's some design decisions I made:
- After favoring .zip for a while, I finally decide to switch to .tar.gz (my original choice). While .zip is convenient in some aspects, the tradeoffs weren't worth it. Specifically, .zip compresses each file individually, which is a huge loss since all of our files have the same structure. Also, reading a .zip requires that you [know the size in advance](https://golang.org/pkg/archive/zip/#NewReader) (I know, right??).
- I chose not to use any Sia-specific types in the package, to make things as straightforward as possible. This decision may seem a little strange, but hear me out: doing so reduces coupling between the format and the Sia repo. For example, I could have used a `modules.NetAddress` instead of a `string` for the `hostAddress` field. But if we ever decided to add a marshalling method to `NetAddress` (unlikely, I know), it could change how the `dotsia` package encoded that field, which would invisibly break compatibility. Besides, fewer dependencies is always a nice thing.
- I waffled on what to do about the `MasterKey` and `ErasureCode` fields. They are tricky because they need to be generic, which means you can't handle them inside the `dotsia` package. Ideally you would first decode the `name` field and branch on that, but JSON decodes everything at once. You could use `json.RawMessage` to get around this, but it's a pain for callers to work with. Ultimately I just stuck with the default type of `map[string]interface{}`. Since these fields will generally be dead-simple anyway, it shouldn't be too painful to do the necessary typecasting.
- The memory layout of the `File` struct has not changed much, despite the potential savings described in #1025. I think it's favorable to simply rely on compression to minimize any structural differences. It would be prudent, though, to compare different compressed structures with real-world data.
- I put the new package inside the `modules/renter` folder, but it could alternatively live at the top level, or even in its own repo. This might make sense given its decoupling from the rest of the Sia code, but I'm not sure.

Next step is to integrate this package with the renter, replacing most of its persist code in the process. (Actually, next step is to get this branch building...)

### DIFF
--- a/modules/renter/dotsia/format.go
+++ b/modules/renter/dotsia/format.go
@@ -25,14 +25,11 @@ import (
 	"bytes"
 	"compress/gzip"
 	"encoding/base64"
+	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"io"
 	"os"
-
-	"github.com/NebulousLabs/Sia/crypto"
-	"github.com/NebulousLabs/Sia/modules"
-	"github.com/NebulousLabs/Sia/types"
 )
 
 const (
@@ -46,6 +43,7 @@ const (
 var (
 	ErrNotSiaFile   = errors.New("not a .sia file")
 	ErrIncompatible = errors.New("file is not compatible with version " + Version)
+	ErrWrongLen     = errors.New("Hex string is not 64 bytes long")
 
 	currentMetadata = Metadata{
 		Header:  Header,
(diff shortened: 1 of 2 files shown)


modules/renter/dotsia/format.go
```diff
@@ -0,0 +1,392 @@
...
+
+Integer fields must not contain negative values.
+
+The "permissions" field is encoded as a decimal number (not octal, or a
+symbolic string) and must not exceed 511 (0777 in octal).
+
+The "path" string must be an absolute Unix-style path; that is, it must begin
+with a leading slash and use '/' as its separator. Additionally, paths must
+not end in a slash, and must not contain any occurences of the current
+directory (.) or parent directory (..) elements.
```
lukechampine: This could be changed to a generic "name," such that path-style names would be optional. But I think it's common enough to enforce in the standard. I believe the .zip format also does this.

DavidVorick:
Will look at this in depth more tomorrow, but in general I support the motion to decouple things. Anything going into the .sia needs to be JSON, and the .sia file needs to be easily extensible.

I don't think it really makes sense to decouple the sia specific types anywhere that it is complicating implementation though. I haven't seen all of the back-bending that's required, so maybe it's an unmerited concern, but after 1.0 we're essentially committing to keeping backwards compatible support for all of our types over the next two years or more. (JSON means we can extend fields, but not deprecate them). So I do think compatibility concerns are not the issue. Decoupling is always a good thing though, and choosing to not use Sia types is certainly a decoupling decision.

modules/renter/dotsia/format.go
```diff
@@ -0,0 +1,390 @@
...
+populated. A File only contains metadata, not the actual file as uploaded
+to the Sia network; thus, metadata pertaining to the uploaded file, such as
+its path and mode bits, are specified inside the File, not the tar header.
+
+The first entry in the tar archive is a special metadata file. It is a JSON
+object corresponding to the Metadata type exported by this package. This
+object contains a version string, which indicates the version of the .sia
+format used. At this time, the .sia format has no promise of backwards or
+forwards compatibility, except that the version field will never be removed
+from the metadata object.
```
DavidVorick: as of 1.0 it would be nice to provide both backwards and forwards compatibility guarantees.

Backwards compatibility will be a requirement.

modules/renter/dotsia/format.go
```diff
@@ -0,0 +1,390 @@
...
+
+The JSON encoding tries to be flexible in allowing arbitrary encryption and
+encoding schemes. As such, the "masterKey" and "erasureCode" fields are
+encoded as generic JSON objects (in Go, a map[string]interface{}). However,
+these objects must always contain a "name" field that identifies the scheme
+used. They may not be null.
+
+Integer fields must not contain negative values.
```
DavidVorick: Seems overly restrictive, I don't think this is a good idea.

modules/renter/dotsia/format.go
```diff
@@ -0,0 +1,390 @@
...
+
+The "permissions" field is encoded as a decimal number (not octal, or a
+symbolic string) and must not exceed 511 (0777 in octal).
```
DavidVorick: we are not allowing directories?

DavidVorick: I think we should allow directories.

modules/renter/dotsia/format.go
DavidVorick: if you assume a leading slash, you can make it so that it's compatible with generic names

modules/renter/dotsia/format.go
```diff
@@ -0,0 +1,390 @@
...
+
+The "path" string must be an absolute Unix-style path; that is, it must begin
+with a leading slash and use '/' as its separator. Additionally, paths must
+not end in a slash, and must not contain any occurences of the current
+directory (.) or parent directory (..) elements.
+
+The "contracts" array must not be null, but it may be empty. The same rule
+applies to the "sectors" array inside the contract object.
```
DavidVorick: Contract information is stored in the '.sia'? I thought we didn't need that (contractor handles contract information, .sia handles hashes/sectors + host ids)

lukechampine:
Now that I think of it, the contract ID changes when you renew, right? So maybe it doesn't make much sense to include the contract ID after all.
Come to think of it... if a host's IP is subject to change as well, how can you figure out what address to request the data from? Use a pubkey instead and look up the latest host announcement using that key?

DavidVorick:
Yeah, you'd probably want to include both the host's ip address and the hosts pubkey. We can reasonably assume that the receiver is going to have a full client. The tricky part is the paid downloads, you are only going to be able to pay for the downloads if you have a file contract open with that host.

We could set up a lightning-style network where hosts will relay payments to eachother, but that's a completely separate can of worms. But, it's definitely something that we could do in the future. In the near term, I think it's safe to assume that people will frequently have enough overlapping hosts to share a file. We could also do things like upload shared files at higher redundancy to a broader range of hosts, which increases the chance that renters are going to have payment channels open with enough hosts.

A final, and centralized, solution would be to have a server like Sunfish that acts as a hub to assist the payments. Then we don't need to worry about all the p2p routing.

These are all things that would take time to support, therefore we can't pursue them by 1.0. The one thing that might be easy enough is to just upload the shared files at an improved redundancy + host propagation.

DavidVorick:
Can you point this PR at master? It's preventing me from deleting `renter-host-upgrades`

PR #1053, opened 2016-04-09
Fix build

lukechampine:
Host-related code in the API and siac needs to be updated for the new paradigm. I imagine that will mean displaying metrics about each folder or something. This PR just gets the build passing again.

In the future, use a tool like `gorename` when changing common types. Much less hassle.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ os:
   - osx
 
 go:
-  - 1.5.3
+  - 1.6
 
 before_install:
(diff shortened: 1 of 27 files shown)


lukechampine:
>`modules/host/storageobligations_test.go:19: types.UnlockHash composite literal uses unkeyed fields`

very weird. Can't reproduce it locally.

lukechampine:
>modules/host/storageobligations_test.go:19: types.UnlockHash composite literal uses unkeyed fields

weird. Can't reproduce that locally.

PR #1057, opened 2016-04-11
Host revise

DavidVorick:
It may compile, but that doesn't mean it's going to actually work. But, I did outline the rest of the revision, found some shortcomings in the form contract protocol, and hopefully got things closer to fully functional.

In the spirit of continuing to keep our heads down, I haven't added any testing or anything fancy. This code is probably a disaster, but that's according to our current development plan.

(for outsiders: given that this is a feature branch, we've chosen to ignore testing for the time being and just get things working at all. We're in the middle of a very focused sprint to get things done, and we believe that things will get a lot easier once the general concept is working, even if it's not tested)

### DIFF
--- a/modules/host/negotiaterevisecontract.go
+++ b/modules/host/negotiaterevisecontract.go
@@ -1,0 +1,337 @@
+package host
+
+// TODO: The revision transaction does need to be sent, because it needs to
+// contain the transaction signatures. Furthermore, the 'WholeTransaction' flag
+// on the transaction signatures needs to be set to false, something that the
+// negotiation protocol needs to check.
+
+// TODO: Since we're gathering untrusted input, need to check for both
+// overflows and nil values.
+
+import (
+	"errors"
+	"net"
+	"time"
+
+	"github.com/NebulousLabs/Sia/crypto"
+	"github.com/NebulousLabs/Sia/encoding"
+	"github.com/NebulousLabs/Sia/modules"
+	"github.com/NebulousLabs/Sia/types"
+
+	"github.com/NebulousLabs/bolt"
+)
+
+var (
+	// errBadModificationIndex is returned if the renter requests a change on a
+	// sector root that is not in the file contract.
+	errBadModificationIndex = errors.New("renter has made a modification that points to a nonexistant sector")
+
+	// errBatchCountExceeded is returned if the renter tries to send a batch of
+	// modifications that is larger than the maximum allowed batch.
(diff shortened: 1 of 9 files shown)


modules/host/negotiaterevisecontract.go
```diff
@@ -0,0 +1,337 @@
...
+	// The renter will either accept or reject the settings + revision
+	// transaction.
+	var acceptStr string
+	err = encoding.ReadObject(conn, &acceptStr, modules.MaxErrorSize)
+	if err != nil {
+		return false, err
+	}
+	if acceptStr != modules.AcceptResponse {
+		return false, errors.New(acceptStr)
+	}
```
lukechampine: probably deserves a helper function if you're going to define one for `rejectNegotiation` as well

modules/host/negotiaterevisecontract.go
```diff
@@ -0,0 +1,337 @@
...
+
+	// The renter is now going to send a batch of modifications followed by and
+	// update file contract revision. Read the number of modifications being
+	// sent by the renter.
+	var modificationCount uint64
+	err = encoding.ReadObject(conn, &modificationCount, 8)
```
lukechampine: On the renter side this is simply:
```go
	// send 'delete' action
	err := encoding.WriteObject(he.conn, []modules.RevisionAction{{
		Type:        modules.ActionDelete,
		SectorIndex: uint64(index),
	}})
```
Is there any reason why you couldn't do the same?
```go
var modifications []modules.RevisionAction
err = encoding.ReadObject(conn, &modifications, maxBatchSize)
if err != nil {
    return false, err
} else if len(modifications) > settings.MaxBatchCount {
    return false, errBatchCountExceeded
}

for _, m := range modifications {
    switch m.Type {
    case modules.ActionDelete:
    // ...
    }
}
```

modules/host/negotiaterevisecontract.go
```diff
@@ -0,0 +1,337 @@
...
+	// Revision is acceptable, write an acceptance string.
+	err = encoding.WriteObject(conn, modules.AcceptResponse)
+	if err != nil {
+		return false, err
+	}
+
+	// Renter will now send the transaction signatures for the file contract,
+	// followed by an indication of whether another iteration is preferred.
+	var renterSig types.TransactionSignature
+	var another bool
```
lukechampine: thor

This is a little confusing though lol. Maybe `shouldContinue` or `renterWantsContinue`?

PR #1060, opened 2016-04-12
Host revise

DavidVorick:
The host announcement protocol has undergone some changes, and the code for managing it has been cleaned up.

The hostdb should be correctly recognizing announcements from the host, but there are no integration tests yet.

### DIFF
--- a/modules/negotiate.go
+++ b/modules/negotiate.go
@@ -1,0 +1,260 @@
+package modules
+
+import (
+	"bytes"
+	"errors"
+	"time"
+
+	"github.com/NebulousLabs/Sia/build"
+	"github.com/NebulousLabs/Sia/crypto"
+	"github.com/NebulousLabs/Sia/encoding"
+	"github.com/NebulousLabs/Sia/types"
+)
+
+const (
+	// AcceptResponse is the response given to an RPC call to indicate
+	// acceptance. (Any other string indicates rejection, and describes the
+	// reason for rejection.)
+	AcceptResponse = "accept"
+
+	// NegotiateFileContractTime defines the amount of time that the renter and
+	// host have to negotiate a file contract. The time is set high enough that
+	// a node behind Tor has a reasonable chance at making the multiple
+	// required round trips to complete the negotiation.
+	NegotiateFileContractTime = 360 * time.Second
+
+	// NegotiateFileContractRevisionTime defines the minimum amount of time
+	// that the renter and host have to negotiate a file contract revision. The
+	// time is set high enough that a full 4MB can be piped through a
+	// connection that is running over Tor.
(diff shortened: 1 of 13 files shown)


modules/negotiate.go
```diff
@@ -0,0 +1,260 @@
...
+
+	// Read the signature out of the reader.
+	var sig crypto.Signature
+	_, err = r.Read(sig[:])
+	if err != nil {
+		return HostAnnouncement{}, err
+	}
+	// Verify the signature.
+	var pk crypto.PublicKey
+	copy(pk[:], ha.PublicKey.Key)
```
lukechampine: would be nice if we had a method for converting a `types.SiaPublicKey` to a `crypto.PublicKey`. It would check both the specifier and the length of the key slice.

modules/negotiate.go
```diff
@@ -0,0 +1,260 @@
...
+		SectorIndex uint64
+		Offset      uint64
+		Data        []byte
+	}
+)
+
+// CreateAnnouncement will take a host announcement and encode it, returning
+// the exact []byte that should be added to the arbitrary data of a
+// transaction.
+func CreateAnnouncement(addr NetAddress, pk types.SiaPublicKey, sk crypto.SecretKey) (signedAnnouncement []byte, err error) {
```
lukechampine: the types are a little weird here. Mixing `SiaPublicKey` and `crypto.SecretKey`, and no mention of the `HostAnnouncement` type...
I dunno, there isn't a super clear way I can think of improving it, other than perhaps replacing the `SiaPublicKey` with a `crypto.PublicKey` (and doing the conversion internally).

modules/negotiate.go
DavidVorick: noted, but put on hold.

modules/negotiate.go
DavidVorick: The `HostAnnouncement` type is used by the hostdb, but seems like it's not used anywhere else. I'm not sure that it does need to be exported, except that it is going into the blockchain in a specific way, so it's probably nice that the type is appearing in our godoc. I don't think it would be too hard to pull the host announcement out of the hostdb, seems like it's just passthrough anyway.

I'm inclined to leave it for now. I was hoping to leave things open so that we could switch to the bitcoin curve + schnorr, and the libraries to do that are now available. We should talk about that more at some point.

I do wish we didn't have to mix those types.

lukechampine: It is? I thought it was only using it for decoding purposes, for which it ought to use `DecodeAnnouncement` instead.

Having the type in the godoc is nice I guess, but it's not a full byte-level specification anyway because it doesn't detail the signature part.

PR #1061, opened 2016-04-12
Vendor muxado

VoidingWarranties:
muxado's WIP branch has been merged (yay!) and has made some breaking changes. For the short term, vendor muxado until we have time to go through all the changes.

Build will fail until NebulousLabs/muxado#1 is merged


DavidVorick:
ah

PR #1064, opened 2016-04-13
siad flag cleanup

VoidingWarranties:
I noticed that we were using global flags in siad instead of command specific flags (`siad version --help` would show all the flags even though none of them applied to the version command).

I also deprecated the shorthands for infrequently used flags.

Finally I added a `siad modules` command to print help info regarding modules available for use with the `-M` or `--modules` flag.

### DIFF
--- a/siad/main.go
+++ b/siad/main.go
@@ -79,6 +79,17 @@ func main() {
 	root.Flags().StringVarP(&globalConfig.Siad.RPCaddr, "rpc-addr", "r", ":9981", "which port the gateway listens on")
 	root.Flags().StringVarP(&globalConfig.Siad.Modules, "modules", "M", "cghmrtw", "enabled modules")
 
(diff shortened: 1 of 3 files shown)


VoidingWarranties:
The descriptions of the modules could use review. I had a hard time concisely summarizing some of the modules.

siad/main.go
```diff
@@ -52,6 +52,67 @@ func versionCmd(*cobra.Command, []string) {
...
+		siad -M g
+Consensus Set (c):
+	The consensus set manages everything related to consensus and keeps the
+	blockchain in sync with the rest of the network.
+	The consensus set requires the gateway to be running.
+	Example:
+		siad -M gc
+Transaction Pool (t):
+	The transaction pool manages unconfirmed transactions.
+	The trannsaction pool requires the consensus set to be running.
```
lukechampine: typo

PR #1065, opened 2016-04-14
API authentication

VoidingWarranties:
Authentication is implemented with [HTTP basic auth](https://tools.ietf.org/html/rfc2617#section-2). Passwords are sent in plaintext. Therefore API authentication should only be used if the API is only exposed over localhost. In the future, TLS can be used to encrypt API calls which will make it safe to use API authentication elsewhere.

It is safe to send passwords in plaintext over localhost as only root should be able to sniff localhost.

Authentication is enabled in siad by default. The `--no-password` flag can be used to disable it.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -70,15 +70,35 @@ func requireUserAgent(h http.Handler, ua string) http.Handler {
 	})
 }
 
-// initAPI determines which functions handle each API call.
-func (srv *Server) initAPI() {
+// requirePassword is middleware that requires a request to authenticate with a
+// password using HTTP basic auth. Usernames are ignored. Empty passwords
+// indicate no authentication is required.
+func requirePassword(h httprouter.Handle, password string) httprouter.Handle {
+	// An empty password is equivalent to no password.
+	if password == "" {
+		return h
+	}
+	return func(w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
+		_, pass, ok := req.BasicAuth()
+		if !ok || pass != password {
+			w.Header().Set("WWW-Authenticate", "Basic realm=\"SiaAPI\"")
+			writeError(w, "API authentication failed.", http.StatusUnauthorized)
+			return
+		}
+		h(w, req, ps)
+	}
+}
+
+// initAPI determines which functions handle each API call. An empty string as
+// the password indicates no password.
+func (srv *Server) initAPI(password string) {
 	router := httprouter.New()
 	router.NotFound = http.HandlerFunc(srv.unrecognizedCallHandler) // custom 404
 
 	// Daemon API Calls
 	router.GET("/daemon/constants", srv.daemonConstantsHandler)
 	router.GET("/daemon/version", srv.daemonVersionHandler)
-	router.GET("/daemon/stop", srv.daemonStopHandler)
+	router.GET("/daemon/stop", requirePassword(srv.daemonStopHandler, password))
 
 	// Consensus API Calls
 	if srv.cs != nil {
@@ -95,34 +115,35 @@ func (srv *Server) initAPI() {
 	// Gateway API Calls
 	if srv.gateway != nil {
 		router.GET("/gateway", srv.gatewayHandler)
-		router.POST("/gateway/add/:netaddress", srv.gatewayAddHandler)
-		router.POST("/gateway/remove/:netaddress", srv.gatewayRemoveHandler)
+		router.POST("/gateway/add/:netaddress", requirePassword(srv.gatewayAddHandler, password))
+		router.POST("/gateway/remove/:netaddress", requirePassword(srv.gatewayRemoveHandler, password))
 	}
 
 	// Host API Calls
 	if srv.host != nil {
 		// Calls directly pertaining to the host.
-		router.GET("/host", srv.hostHandlerGET)                // Get a bunch of information about the host.
-		router.POST("/host", srv.hostHandlerPOST)              // Set HostInternalSettings.
-		router.POST("/host/announce", srv.hostAnnounceHandler) // Announce the host, optionally on a specific address.
+		router.GET("/host", srv.hostHandlerGET)                                           // Get a bunch of information about the host.
+		router.POST("/host", requirePassword(srv.hostHandlerPOST, password))              // Set HostInternalSettings.
+		router.POST("/host/announce", requirePassword(srv.hostAnnounceHandler, password)) // Announce the host, optionally on a specific address.
 
 		// Calls pertaining to the storage manager that the host uses.
 		router.GET("/storage", srv.storageHandler)
-		router.POST("/storage/folders/add/*folder", srv.storageFoldersAddHandler)
-		router.POST("/storage/folders/remove/*folder", srv.storageFoldersRemoveHandler)
-		router.POST("/storage/folders/resize/*folder", srv.storageFoldersResizeHandler)
-		router.POST("/storage/sectors/delete/:merkleroot", srv.storageSectorsDeleteHandler)
+		router.POST("/storage/folders/add/*folder", requirePassword(srv.storageFoldersAddHandler, password))
+		router.POST("/storage/folders/remove/*folder", requirePassword(srv.storageFoldersRemoveHandler, password))
+		router.POST("/storage/folders/resize/*folder", requirePassword(srv.storageFoldersResizeHandler, password))
+		router.POST("/storage/sectors/delete/:merkleroot", requirePassword(srv.storageSectorsDeleteHandler, password))
 	}
 
 	// Miner API Calls
 	if srv.miner != nil {
 		router.GET("/miner", srv.minerHandler)
-		router.GET("/miner/header", srv.minerHeaderHandlerGET)
-		router.POST("/miner/header", srv.minerHeaderHandlerPOST)
-		router.GET("/miner/start", srv.minerStartHandler)
-		router.GET("/miner/stop", srv.minerStopHandler)
-		router.GET("/miner/headerforwork", srv.minerHeaderHandlerGET)  // COMPATv0.4.8
-		router.POST("/miner/submitheader", srv.minerHeaderHandlerPOST) // COMPATv0.4.8
+		router.GET("/miner/header", requirePassword(srv.minerHeaderHandlerGET, password))
+		router.POST("/miner/header", requirePassword(srv.minerHeaderHandlerPOST, password))
+		router.GET("/miner/start", requirePassword(srv.minerStartHandler, password))
+		router.GET("/miner/stop", requirePassword(srv.minerStopHandler, password))
+		// TODO: doesn't adding authentication break compatibility?
+		router.GET("/miner/headerforwork", requirePassword(srv.minerHeaderHandlerGET, password))  // COMPATv0.4.8
+		router.POST("/miner/submitheader", requirePassword(srv.minerHeaderHandlerPOST, password)) // COMPATv0.4.8
 	}
 
 	// Renter API Calls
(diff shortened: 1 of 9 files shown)


VoidingWarranties:
As discussed in person, this makes siac clumsy to use as you need to enter a password for every command. It would be easier to use if only specific API calls require a password. I propose making only state modifying API calls require a password.

VoidingWarranties:
Next step: add an API authentication cookie similar to https://github.com/bitcoin/bitcoin/pull/6388/

VoidingWarranties:
One problem with this PR is that people will likely want to have their password persist, whereas this implementation require the user to input the password on every startup of siad. If Sia had a config file we could store it there (optionally hashed and salted), but we don't.

To make it easier for user's to set the same password we could include a flag like
`--password=<password>`. I specifically avoided this in this PR because it would encourage users to create startup scripts that stored the password in plaintext. Instead, we could do something similar to Bitcoin's `-rpcauth` flag, and let the user specify a salt and hash of the password,
e.g. `--api-auth=<salt>:<hash>`

See https://github.com/bitcoin/bitcoin/pull/7044 for discussion on Bitcoin's RPC authentication and their implementation.

VoidingWarranties:
Hmmm, both of these solutions work well for API authentication, but the plan was to make the API authentication password also used to encrypt modules' persist files and optionally the wallet encryption password. These two solutions wouldn't work for that.

api/api.go
```diff
@@ -66,43 +117,44 @@ func (srv *Server) initAPI() {
...
-		router.GET("/miner/stop", srv.minerStopHandler)
-		router.GET("/miner/headerforwork", srv.minerHeaderHandlerGET)  // COMPATv0.4.8
-		router.POST("/miner/submitheader", srv.minerHeaderHandlerPOST) // COMPATv0.4.8
+		router.GET("/miner/header", requirePassword(srv.minerHeaderHandlerGET, password))
+		router.POST("/miner/header", requirePassword(srv.minerHeaderHandlerPOST, password))
+		router.GET("/miner/start", requirePassword(srv.minerStartHandler, password))
+		router.GET("/miner/stop", requirePassword(srv.minerStopHandler, password))
+		// TODO: doesn't adding authentication break compatibility?
+		router.GET("/miner/headerforwork", requirePassword(srv.minerHeaderHandlerGET, password))  // COMPATv0.4.8
+		router.POST("/miner/submitheader", requirePassword(srv.minerHeaderHandlerPOST, password)) // COMPATv0.4.8
```
VoidingWarranties: If we are breaking compatibility by adding authentication, we might as well get rid of the endpoints, no?

api/api.go
```diff
@@ -116,22 +168,23 @@ func (srv *Server) initAPI() {
...
-		router.POST("/wallet/siafunds", srv.walletSiafundsHandler)
-		router.POST("/wallet/siagkey", srv.walletSiagkeyHandler)
+		router.GET("/wallet/backup", requirePassword(srv.walletBackupHandler, password))
+		router.POST("/wallet/init", requirePassword(srv.walletInitHandler, password))
+		router.POST("/wallet/lock", requirePassword(srv.walletLockHandler, password))
+		router.POST("/wallet/seed", requirePassword(srv.walletSeedHandler, password))
+		router.GET("/wallet/seeds", requirePassword(srv.walletSeedsHandler, password))
+		router.POST("/wallet/siacoins", requirePassword(srv.walletSiacoinsHandler, password))
+		router.POST("/wallet/siafunds", requirePassword(srv.walletSiafundsHandler, password))
+		router.POST("/wallet/siagkey", requirePassword(srv.walletSiagkeyHandler, password))
```
lukechampine: the usual way to do middleware is to wrap the router, not the individual handlers.

see below:
```go
 	// Apply UserAgent middleware and create HTTP server
 	uaRouter := requireUserAgent(router, srv.requiredUserAgent)
 	srv.apiServer = &http.Server{Handler: uaRouter}
```
this would become:
```go
 	// Apply UserAgent and authentication middleware and create HTTP server
 	uaRouter := requireUserAgent(router, srv.requiredUserAgent)
 	uaAuthRouter := requirePassword(uaRouter, password)
 	srv.apiServer = &http.Server{Handler: uaAuthRouter}
```
and `requirePassword` would need to be modified accordingly.
speaking of which, do we still need the UserAgent restriction if we have authentication?

I suppose the difference here is that we aren't applying auth to every endpoint, just the vast majority. We could use a whitelist, I suppose...or just require authentication for every call.

lukechampine:
btw, since `bitcoind` makes you specify `rpcuser` and `rpcpassword` in plaintext, aren't they trivially discoverable?

api/api.go
DavidVorick: The user agent is no longer strictly necessary once we are doing authentication. I'm trying to think if there are any edge cases where it would still be good to have it, generally I like to be conservative/prudent.

DavidVorick:
yes, I think so. =/

api/api.go
VoidingWarranties: If you wrap the router you require authentication for every endpoint. The reason for only requiring auth for some endpoints is to improve the UX.

VoidingWarranties:
Yes, bitcoin allows you to specify the password in plaintext. We definitely shouldn't do that.

api/api.go
VoidingWarranties: Does bitcoin have a similar UserAgent restriction? Once password auth is enabled by default it might make sense to remove it.

api/api.go
```diff
@@ -8,24 +8,55 @@ import (
...
-// HttpGET is a utility function for making http get requests to sia with a whitelisted user-agent
+// HttpGET is a utility function for making http get requests to sia with a
+// whitelisted user-agent. A non-2xx response does not return an error.
 func HttpGET(url string) (resp *http.Response, err error) {
 	req, err := http.NewRequest("GET", url, nil)
 	if err != nil {
 		return nil, err
 	}
-	req.Header.Add("User-Agent", "Sia-Agent")
+	req.Header.Set("User-Agent", "Sia-Agent")
```
DavidVorick: What is the difference?

api/api.go
DavidVorick: makes sense to me. Especially considering how badly we broke compatibility with 060, I don't see much of a reason to keep these endpoints around.

api/api.go
DavidVorick: Is it that significant if we wrap every route? It would definitely improve privacy surface area.

I don't think it's that bad to have this level of customization, though it does require you to think about passwords every time you make a change. It adds complexity to the codebase and should be considered carefully. If we choose to expose endpoints, it's not really something that we can reverse easily in the future. But the other way is easier, if you decide that something doesn't need to be secret, you can still accept a password on the call.

api/api.go
```diff
@@ -116,22 +168,23 @@ func (srv *Server) initAPI() {
...
+		router.POST("/wallet/siafunds", requirePassword(srv.walletSiafundsHandler, password))
+		router.POST("/wallet/siagkey", requirePassword(srv.walletSiagkeyHandler, password))
 		router.GET("/wallet/transaction/:id", srv.walletTransactionHandler)
 		router.GET("/wallet/transactions", srv.walletTransactionsHandler)
 		router.GET("/wallet/transactions/:addr", srv.walletTransactionsAddrHandler)
-		router.POST("/wallet/unlock", srv.walletUnlockHandler)
-		router.POST("/wallet/encrypt", srv.walletInitHandler) // COMPATv0.4.0
+		router.POST("/wallet/unlock", requirePassword(srv.walletUnlockHandler, password))
+		// TODO: doesn't authentication break compatibility?
+		router.POST("/wallet/encrypt", requirePassword(srv.walletInitHandler, password)) // COMPATv0.4.0
```
DavidVorick: as stated above, adding auth means that we can remove the compat calls.

api/api.go
```diff
@@ -116,22 +168,23 @@ func (srv *Server) initAPI() {
...
-		router.POST("/wallet/seed", srv.walletSeedHandler)
-		router.GET("/wallet/seeds", srv.walletSeedsHandler)
-		router.POST("/wallet/siacoins", srv.walletSiacoinsHandler)
-		router.POST("/wallet/siafunds", srv.walletSiafundsHandler)
-		router.POST("/wallet/siagkey", srv.walletSiagkeyHandler)
+		router.GET("/wallet/backup", requirePassword(srv.walletBackupHandler, password))
+		router.POST("/wallet/init", requirePassword(srv.walletInitHandler, password))
+		router.POST("/wallet/lock", requirePassword(srv.walletLockHandler, password))
+		router.POST("/wallet/seed", requirePassword(srv.walletSeedHandler, password))
+		router.GET("/wallet/seeds", requirePassword(srv.walletSeedsHandler, password))
```
DavidVorick: if you are doing GET on /wallet/seeds, you should be doing POST on /wallet/seeds as well

siac/main.go
```diff
@@ -93,6 +106,18 @@ func apiPost(call, vals string) (*http.Response, error) {
...
+		// Prompt for password and retry request with authentication.
+		password, err := speakeasy.Ask("API password: ")
+		if err != nil {
+			return nil, err
+		}
+		resp, err = api.HttpPOSTAuthenticated("http://"+addr+call, vals, password)
+		if err != nil {
+			return nil, errors.New("no response from daemon - authentication failed")
+		}
+	}
```
DavidVorick: hmm, if `siac` is going to be asking for a password every time you run it, I can see why you would want to leave the information api calls exposed.

siad/daemon.go
```diff
@@ -67,6 +68,25 @@ func processConfig(config Config) (Config, error) {
...
 func startDaemon(config Config) (err error) {
+	// Prompt user for API password.
+	var password string
+	if config.Siad.AuthenticateAPI {
+		password, err = speakeasy.Ask("Enter API password: ")
+		if err != nil {
+			return err
+		}
+		if password == "" {
+			return errors.New("password cannot be blank, don't use the --authenticate-api flag instead")
```
DavidVorick: wording here is weird.

"password cannot be blank" is probably enough.

DavidVorick:
Overall this PR looks good, want to spend a little bit more time thinking through the implications of what we're doing, given that it's both a substantial and permanent change.

I don't think this was the right place to change the `/wallet/seed` call, you should move that to a different PR. More, you should not change it until we're on the 'finalize api' part of the release cycle, because it is going to break compatibility with siac and '/wallet/seed POST' is a pretty important call. (this is fixed by updating 'siac' as well, but now you're writing even more code and it makes even more sense to be a different PR)

api/api.go
lukechampine: The header uses a `map[string][]string` internally, not a `map[string]string`. So a header can have multiple values associated with it. `Set` ensures that ours is the only value, whereas `Add` appends to the `[]string`.

api/api.go
VoidingWarranties: What do you mean?

VoidingWarranties:
>
I don't think this was the right place to change the `/wallet/seed` call, you should move that to a different PR. More, you should not change it until we're on the 'finalize api' part of the release cycle, because it is going to break compatibility with siac and '/wallet/seed POST' is a pretty important call. (this is fixed by updating 'siac' as well, but now you're writing even more code and it makes even more sense to be a different PR)

I don't think I changed anything? Maybe I did accidentally...

VoidingWarranties:
Note that you can supply the password to the `--addr` flag so that `siac` never prompts for a password.
Example:
```
siac stop --addr :<api_password>@localhost:9980
```
Of course doing this would expose your password in plaintext in bash history.

siac/main.go
VoidingWarranties: Yes this was my main reasoning for leaving non-sensitive API calls exposed. It's habbit for me to check up on Sia (make sure it's staying in sync, etc...) every once in a while using `siac` and it would be a pain to type the password every time.

api/api.go
VoidingWarranties: Yes, it should result in identical headers unless the user agent was being silently set elsewhere.

api/api.go
DavidVorick: it looks like in this PR you changed the /wallet/seed call to /wallet/seeds?

VoidingWarranties: Both endpoints already exist :P

DavidVorick:
re: /wallet/seeds, I read the diff wrong, you only changed the comments, all good.

api/api.go
DavidVorick: never mind I was reading the diff wrong.

api/api.go
```diff
@@ -66,34 +117,33 @@ func (srv *Server) initAPI() {
...
 	}
 
 	// Host API Calls
 	if srv.host != nil {
 		// Calls directly pertaining to the host.
-		router.GET("/host", srv.hostHandlerGET)                // Get a bunch of information about the host.
-		router.POST("/host", srv.hostHandlerPOST)              // Set HostInternalSettings.
-		router.POST("/host/announce", srv.hostAnnounceHandler) // Announce the host, optionally on a specific address.
+		// TODO: does the /host GET endpoint expose private information? if so it should require authentication.
+		router.GET("/host", srv.hostHandlerGET)                                           // Get a bunch of information about the host.
```
VoidingWarranties: Should this endpoint be authenticated?

api/api.go
```diff
@@ -66,34 +117,33 @@ func (srv *Server) initAPI() {
...
+		router.POST("/host", requirePassword(srv.hostHandlerPOST, password))              // Set HostInternalSettings.
+		router.POST("/host/announce", requirePassword(srv.hostAnnounceHandler, password)) // Announce the host, optionally on a specific address.
 
 		// Calls pertaining to the storage manager that the host uses.
-		router.GET("/storage", srv.storageHandler)
-		router.POST("/storage/folders/add/*folder", srv.storageFoldersAddHandler)
-		router.POST("/storage/folders/remove/*folder", srv.storageFoldersRemoveHandler)
-		router.POST("/storage/folders/resize/*folder", srv.storageFoldersResizeHandler)
-		router.POST("/storage/sectors/delete/:merkleroot", srv.storageSectorsDeleteHandler)
+		router.GET("/storage", srv.storageHandler) // TODO: does the /storage GET endpoint expose private information? if so it should require authentication.
```
VoidingWarranties: Should this endpoint be authenticated?

DavidVorick: nah

api/api.go
DavidVorick: nah

VoidingWarranties:
Moved to #1101 so that these changes are merged into a staging branch that can be merged into master after v0.6.0 has been released.

api/api.go
```diff
@@ -103,15 +150,15 @@ func (srv *Server) initAPI() {
 		router.GET("/renter/allowance", srv.renterAllowanceHandlerGET)
 		router.POST("/renter/allowance", srv.renterAllowanceHandlerPOST)
```
lukechampine: needs auth

PR #1067, opened 2016-04-14
Get FormContract working

lukechampine:
The `FormContract` RPC now works. Others are a work in progress. The first host+renter integration test has been written. Some tweaks to the protocol were necessary.
Collateral handling is pretty painful and dangerously opaque. Ideally it could be simplified, but if not, we should at least move some of the logic into a function (perhaps in modules/negotiate.go) and document it extensively.

### DIFF
--- a/modules/host/consts.go
+++ b/modules/host/consts.go
@@ -76,6 +76,10 @@ var (
 	// opportunity cost in being a host.
 	defaultCollateral = types.NewCurrency64(50) // 50 SC / GB / Month
 
+	defaultCollateralFraction = types.NewCurrency64(650e3)
+
+	defaultMaxCollateral = types.NewCurrency64(10000).Mul(types.SiacoinPrecision)
+
 	// defaultWindowSize is the size of the proof of storage window requested
 	// by the host. The host will not delete any obligations until the window
 	// has closed and buried under several confirmations. For release builds,
--- a/modules/host/negotiate.go
+++ b/modules/host/negotiate.go
@@ -8,6 +8,11 @@ import (
 	"github.com/NebulousLabs/Sia/modules"
 )
 
+// acceptNegotiation writes the 'accept' response to conn.
+func acceptNegotiation(conn net.Conn) error {
+	return encoding.WriteObject(conn, modules.AcceptResponse)
+}
+
 // rejectNegotiation will write a rejection response to the connection and
 // return the input error composed with the error received from writing to the
 // connection.
--- a/modules/host/negotiate.go
+++ b/modules/host/negotiate.go
@@ -1,25 +1,10 @@
 package host
 
 import (
-	"net"
 	"sync/atomic"
 
-	"github.com/NebulousLabs/Sia/encoding"
 	"github.com/NebulousLabs/Sia/modules"
 )
-
-// acceptNegotiation writes the 'accept' response to conn.
-func acceptNegotiation(conn net.Conn) error {
-	return encoding.WriteObject(conn, modules.AcceptResponse)
-}
-
-// rejectNegotiation will write a rejection response to the connection and
-// return the input error composed with the error received from writing to the
-// connection.
-func rejectNegotiation(conn net.Conn, err error) error {
-	writeErr := encoding.WriteObject(conn, err.Error())
-	return composeErrors(err, writeErr)
-}
 
 // NetAddress returns the address at which the host can be reached.
 func (h *Host) NetAddress() modules.NetAddress {
--- a/modules/host/negotiateformcontract.go
+++ b/modules/host/negotiateformcontract.go
@@ -182,26 +182,26 @@ func (h *Host) managedFinalizeContract(builder modules.TransactionBuilder, rente
 // file contract, creating a storage obligation and submitting the contract to
 // the blockchain.
 func (h *Host) managedRPCFormContract(conn net.Conn) error {
-	// Set the negotiation deadline.
-	conn.SetDeadline(time.Now().Add(modules.NegotiateFileContractTime))
-
 	// Send the host settings to the renter. If the host is not accepting new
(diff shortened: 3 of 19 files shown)


modules/host/consts.go
```diff
@@ -76,6 +76,10 @@ var (
 	// opportunity cost in being a host.
 	defaultCollateral = types.NewCurrency64(50) // 50 SC / GB / Month
 
+	defaultCollateralFraction = types.NewCurrency64(650e3)
+
+	defaultMaxCollateral = types.NewCurrency64(10000).Mul(types.SiacoinPrecision)
```
DavidVorick: Given that requesting collateral from the host is so expensive, especially when file contracts are only 12 weeks long (the renter pays a 3.9% fee to lock the host out of funds for 12 weeks? that's a really expensive attack, especially since the renter has to lock itself out of half the funds for 12 weeks as well), I don't think we really need to worry about attackers here. It's just too expensive to be practical, and the host can get around the problem simply by getting more coins. The host is not losing money at all, it'll come back, which means that the only risks here are potential profit from interest, and the risk inherent to holding a cryptocurrency.

A bigger fear would be that we have a lot of renters who try out early on, lock up a bunch of host funds for a few weeks, but then never use the storage because they get disinterested and go back to DropBox. So, maybe early on it makes sense to have a small default max collateral, but I think that as time passes I think we'll be pretty easily able to jump it quite a bit.

modules/host/negotiate.go
```diff
@@ -8,6 +8,11 @@ import (
 	"github.com/NebulousLabs/Sia/modules"
 )
 
+// acceptNegotiation writes the 'accept' response to conn.
+func acceptNegotiation(conn net.Conn) error {
+	return encoding.WriteObject(conn, modules.AcceptResponse)
+}
```
DavidVorick: do we want to move this into modules/negotiate.go? Then the renter would have access to it as well.

DavidVorick: also, we could add an error/panic to rejectNegotiation if you try to write an error that is too large.

modules/host/negotiateformcontract.go
```diff
@@ -166,7 +148,15 @@ func (h *Host) managedFinalizeContract(builder modules.TransactionBuilder, rente
...
 	}
+	err = h.lockStorageObligation(so)
+	if err != nil {
+		return nil, err
+	}
 	err = h.addStorageObligation(so)
+	err2 := h.unlockStorageObligation(so)
+	if err2 != nil {
+		return nil, err2
+	}
```
DavidVorick: This was code that I wrote, so it's my fault, but this cluster is pretty messy. Maybe there's a better organization? At the very least, err2->err and err->addObligationErr or something

modules/host/negotiateformcontract.go
```diff
@@ -192,27 +182,27 @@ func (h *Host) managedFinalizeContract(builder modules.TransactionBuilder, rente
...
+	h.mu.RUnlock()
 	if !settings.AcceptingContracts {
-		// The host is not accepting contracts, the connection can be closed.
-		// The renter has been given enough information to understand that the
-		// connection is going to be closed.
 		return nil
 	}
 
+	// Set the negotiation deadline.
+	conn.SetDeadline(time.Now().Add(modules.NegotiateFileContractTime))
```
DavidVorick: deadline is now set below the settings management?

DavidVorick: oh, I guess the settings adds its own deadline. got it.

So, I'm increasingly convinced that it would be beneficial for the host to recognize repeat renters. The biggest reason is repeat collateral, simply because collateral is such a sensitive resource. Mostly, I'm cautious about users who come in, try the system, and then leave, consuming a relatively large amount of host collateral without actually uploading many files. If we have 20x as many users as hosts, and each user is talking to a host, the host may find that they actually need a massive amount of coins to properly manage the collateral.

A tool such as renter recognition would go a long way. I'm not sure the best way to approach that, but we should keep that in mind moving forward.

modules/renter/contractor/negotiate.go
```diff
@@ -71,48 +78,116 @@ func negotiateContract(conn net.Conn, host modules.HostDBEntry, fc types.FileCon
...
+	if acceptStr != modules.AcceptResponse {
+		return Contract{}, errors.New(acceptStr)
+	}
+	var hostSigs []types.TransactionSignature
+	if err := encoding.ReadObject(conn, &hostSigs, 2e3); err != nil {
+		return Contract{}, errors.New("couldn't read the host's signatures: " + err.Error())
+	}
+	for _, sig := range hostSigs {
+		txnBuilder.AddTransactionSignature(sig)
+	}
```
DavidVorick: I'm not going to spend a ton of time reviewing the networking code for now, I think we're still tweaking things. Once it feels like the networking code has hit a more or less resting state (including all RPCs implemented), then I'll look at it more closely.

Also, I do think that we should try to get the renew rpc working this release. I said earlier that we didn't need it, but we're in the RPC headspace so we should probably knock them all out at once.

modules/host/negotiate.go
lukechampine: yes to both, I think.

modules/negotiate_test.go
```diff
@@ -90,3 +90,30 @@ func TestAnnouncementHandling(t *testing.T) {
...
+	err = WriteNegotiationRejection(buf, ErrLowBalance)
+	if err != ErrLowBalance {
+		t.Fatal(err)
+	}
+	err = ReadNegotiationAcceptance(buf)
+	// can't compare to ErrLowBalance directly; contents are the same, but pointer is different
+	if err == nil || err.Error() != ErrLowBalance.Error() {
+		t.Fatal(err)
+	}
+}
```
DavidVorick: ^^

PR #1069, opened 2016-04-14
Cleanup

DavidVorick:
Some really quick cleanup of the modules and host package. Wanted to make a separate PR in the interest of keeping things cleaner than I have been recently. Basically nothing interesting is happening here.


modules/negotiate.go
```diff
@@ -192,8 +197,18 @@ type (
 		Version        string `json:"version"`
 	}
 
-	// A RevisionAction is a description of an edit to be performed on a
-	// contract.
+	// A RevisionAction is a description of an edit to be performed on a file
+	// contract. Three types are allowed, 'ActionDelecte', 'ActionInsert', and
```
lukechampine: typo

PR #1071, opened 2016-04-15
Some Improvements

DavidVorick:
I really meant to PR this as soon as the form contract negotiation code was cleaned up, but then I fell down a rabbit hole when I told myself 'well, I'll just clean up this one netAddress thing'. I should have committed code before doing that, and I'm sorry that this isn't organized better. It was pretty avoidable.

That said, I do think that this PR contains a significant number of improvements to the code quality and overall structure of the host. I will explain how the netAddress stuff works:

The host has two fields for the net address. The first is `host.autoAddress` and the second is `host.settings.NetAddress`. The settings address is only set if the user has set it manually. They are both saved in the host persistence. If the settings address is not set, then the host will fall back to the auto address, which it gets by periodically (every 30 minutes) checking its hostname. If at any point the hostname changes, the host will add another host announcement to the blockchain.

The host will not automatically announce to the blockchain when the address is set manually, but the host can report whether or not an announcement has been submitted to the transaction pool. This is a bit short of being able to tell whether the announcement made it to the blockchain, but since there are fees I'm not sure what the best response is. The host does make announcements with a fairly high transaction fee, so I would think it likely that the host does succeed.

Then again, the existing transaction pool is not that intelligent.


doc/File Contract Negotiation.md
```diff
@@ -59,37 +59,41 @@ Settings Request
...
+# This minimizes the surface area of data for a malicious party to manipulate,
+# which means less verification code, which means less chances of having a bug in
+# the verification code.
+#
+# The renter pays for the siafund fee on the host's collateral and contract
+# fee. If a renter opens a file contract and then never uses it, the host does
+# not lose money. This does put the renter at risk, as they may open up a file
+# contract and then watch the host leave, but the renter is spreading the risk
+# over communications with many hosts, and has a reputation system that will help
+# ensure that the renter is only dealing with upstanding hosts.
```
lukechampine: this does not render well in markdown, lol

DavidVorick: oh. They are supposed to be comments. I guess that's the '%' character.

DavidVorick: double oh. I guess there's not an easy way to make comments in markdown. I'll have to figure out some other way to get this information across.

modules/host/negotiatesettings.go
```diff
@@ -27,14 +27,21 @@ func (h *Host) managedRPCSettings(conn net.Conn) error {
 	// Set the negotiation deadline.
 	conn.SetDeadline(time.Now().Add(modules.NegotiateSettingsTime))
 
-	h.mu.RLock()
+	h.mu.Lock()
+	h.revisionNumber++
```
lukechampine: why do you bump the revision number every time someone requests your settings?

modules/host/announce.go
```diff
@@ -63,26 +68,22 @@ func (h *Host) announce(addr modules.NetAddress) error {
...
-		return errors.New("can't announce without knowing external IP")
+	// Determine whether to use the settings.NetAddress or autoAddress.
+	if h.settings.NetAddress != "" {
+		return h.announce(h.settings.NetAddress)
 	}
-
-	h.mu.Lock()
-	defer h.mu.Unlock()
-	return h.announce(addr)
+	if (h.autoAddress == "" || h.autoAddress.IsLoopback()) && build.Release != "testing" {
```
lukechampine: do you need to worry about it being a loopback address?

modules/host/network.go
```diff
@@ -5,21 +5,57 @@ import (
...
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/types"
 )
 
+// threadedUpdateHostname periodically runs 'managedLearnHostname', which
+// checks if the host's hostname has changed, and makes an updated host
+// announcement if so.
+//
+// TODO: This thread doesn't actually have clean shutdown, because the sleep is
+// outside of of the resource lock.
```
lukechampine: would be easiest with a shutdown channel:
```go
select {
case <-time.After(30 * time.Minute):
case <-h.shutdown:
    return
}
```

modules/host/network.go
```diff
@@ -5,21 +5,57 @@ import (
...
-	h.netAddress = modules.NetAddress(h.listener.Addr().String())
+	h.mu.Lock()
+	h.port = modules.NetAddress(h.listener.Addr().String()).Port()
+	if build.Release == "testing" {
+		// TODO: I'm not really sure if this is causing errors because the
+		// tests are bad, or because the design is bad. But what seems to be
+		// happening is that the host is realizing it has an empty autoAddress,
+		// and this is somehow confusing the contractor. Bears further
+		// investigation.
+		h.autoAddress = modules.NetAddress(":" + h.port)
```
lukechampine: safer to do `net.JoinHostPort("localhost", h.port)`

modules/host/announce.go
DavidVorick: I was wondering this too. Supposedly the hostname discovery will not return a loopback address, so I think I can remove that clause.

modules/host/negotiatesettings.go
DavidVorick: then you are certain that you didn't forget.

We have a 64bit number of counting available to us, so there's no risk of running out.

The settings are going to be adjusting rapidly either way. The host is going to have bandwidth load balancing at some point, and the bandwidth balancer might update the price once a minute or more. Better to make sure that the settings number is always going up, than to end up in a situation where the host seems to be giving out conflicting settings information.

modules/host/network.go
DavidVorick: Yeah, was thinking that. But then I need to add another field to the host, and I'm not sure how it's going to interact with the other functions/routines that don't seem to be closing cleanly right now.

I'm going to come back to this after all the features are written, then I can compose a closing strategy that takes into account everything that's happening in the host.

modules/host/network.go
DavidVorick: still need to launch the listener. I'm actually not really certain what all should be happening in a production vs. a dev environment. For example, the port management is actually not synchronized right now. It's possible to run the host and close it, and get a race condition where the host clears the port and then forwards the port, which is undesired behavior.

I don't know when is the best time to add the port to the hostname, and I'm not really sure why it needs to happen here to get the contractor tests to pass, but I do know that if you remove the code the hostdb will stop adding the host to the database (maybe because it ignores host announcements where the net address is the empty string?).

Like, for testing purposes should I just ignore all of the hostname and port stuff? Should I somehow use dependency management/mocking to give the tester control over the apparent state of the host?

Maybe it would be better to have the hostname lookup "succeed" and set the autoAddress to "foo:1234" during testing. Or be configurable. I think the best solution is probably to control the hostname lookup during testing, which should give us the most flexibility to test things and be certain that a test is actually able to catch problems that might occur in production.

lukechampine: you could mock the upnp dependency, I suppose, or the `myExternalIP` function.

The hostdb will add the host to `allHosts`, but it won't be able to scan it if the net address is empty, so it will never be added to `activeHosts`.

DavidVorick: your proposal, `net.Join(...)`, restricts testing to only happening over localhost. I do think that's a reasonable thing to do, so I'm going to go ahead and do it.

DavidVorick:
fixed up

lukechampine:
hmm. Looks like it's causing the hostdb to ignore the host.

PR #1074, opened 2016-04-16
Add 'RecentRevision' to the list of host RPCs

DavidVorick:
'RecentRevision' allows a renter to request a recent file contract
revision from the host. The host will return the revision and the
signatures for the revision. A helper function has been added to the
modules package which verifies that the revision + signatures are valid
and follow protocol.

### DIFF
--- a/modules/host/host.go
+++ b/modules/host/host.go
@@ -30,6 +30,21 @@ package host
 // TODO: 'announced' doesn't tell you if the announcement made it to the
 // blockchain.
 
+// TODO: Need to make sure that the revision exchange for the renter and the
+// host is being handled correctly. For the host, it's not so difficult. The
+// host need only send the most recent revision every time. But, the host
+// should not sign a revision unless the renter has explicitly signed such that
+// the 'WholeTransaction' fields cover only the revision and that the
+// signatures for the revision don't depend on anything else. The renter needs
+// to verify the same when checking on a file contract revision from the host.
+// If the host has submitted a file contract revision where the signatures have
+// signed the whole file contract, there is an issue.
+
+// TODO: there is a mistake in the file contract revision rpc, the host, if it
+// does not have the right file contract id, should be returning an error there
+// to the renter (and not just to it's calling function without informing the
+// renter what's up).
+
 // TODO: clean up all of the magic numbers in the host.
 
 // TODO: host_test.go has commented out tests.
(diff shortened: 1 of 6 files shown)


DavidVorick:
RPCRecentRevision operates much like RPCSettings in that it's meant to be chained into other RPCs such as Revise, Download, and Renew. I created this after I started working on Download and realized that there was a nontrivial amount of code which would be replicated over the RPCs.

As an added bonus, there is now a way for the renter to resynchronize with the host in the event of a failed connection, that way the renter can determine whether its recent revision was successful. It doesn't help in the case where the host goes offline in that key time where the host has maybe accepted the file contracts but hasn't sent the signatures yet. Though rare, it would leave the renter in an interesting limbo that the renter should be able to handle gracefully.

modules/host/host.go
```diff
@@ -30,6 +30,21 @@ package host
...
+// the 'WholeTransaction' fields cover only the revision and that the
+// signatures for the revision don't depend on anything else. The renter needs
+// to verify the same when checking on a file contract revision from the host.
+// If the host has submitted a file contract revision where the signatures have
+// signed the whole file contract, there is an issue.
+
+// TODO: there is a mistake in the file contract revision rpc, the host, if it
+// does not have the right file contract id, should be returning an error there
+// to the renter (and not just to it's calling function without informing the
+// renter what's up).
```
DavidVorick: these todos mostly serve as triggers for me to remember to check various things. In this case, I don't think the contents are 100% relevant anymore as this PR solves some of the things listed, but I still haven't done all the actions mentioned. Improvements will be ongoing.

modules/host/negotiaterecentrevision.go
```diff
@@ -0,0 +1,74 @@
...
+	// Fetch the storage obligation with the file contract revision
+	// transaction.
+	var so *storageObligation
+	err = h.db.Update(func(tx *bolt.Tx) error {
+		fso, err := getStorageObligation(tx, fcid)
+		so = &fso
+		return err
+	})
+	if err != nil {
+		return types.FileContractID{}, composeErrors(err, modules.WriteNegotiationRejection(conn, err))
```
lukechampine: doesn't `WriteNegotiationRejection` already compose the errors?

modules/host/negotiaterecentrevision.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	// Sanity check - verify that the host has a valid revision and set of
+	// signatures.
+	h.mu.RLock()
+	blockHeight := h.blockHeight
+	h.mu.RUnlock()
+	err = modules.VerifyFileContractRevisionTransactionSignatures(recentRevision, revisionSigs, blockHeight)
+	if err != nil {
+		h.log.Critical("host is inconsistend, bad file contract revision transaction", err)
```
lukechampine: spurdo

typo

lukechampine:
`RevisionRequest` is a pretty generic name. In the contractor code I usually refer to it as `LastRevision`, but `RecentRevision` seems okay too.

modules/host/negotiaterecentrevision.go
DavidVorick: oh, right

modules/host/negotiaterecentrevision.go
DavidVorick: I laughed at this harder than I would like to admit

PR #1076, opened 2016-04-19
Download

DavidVorick:
So, now the host supports a download rpc. I'm feeling a little frazzled, because I feel like I still haven't finalized RPCRevise.

I'm going to be prioritizing that now that RPCDownload seems mostly complete. Once everything else feels finalized, then I'll move on to RPCRenew. Then we'll be done with the networking stuff.

### DIFF
--- a/modules/host/negotiatedownload.go
+++ b/modules/host/negotiatedownload.go
@@ -201,6 +201,18 @@ func (h *Host) managedRPCDownload(conn net.Conn) error {
 		return err
 	}
 
+	// Lock the storage obligation during the revision.
+	err = h.lockStorageObligation(so)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		err = h.unlockStorageObligation(so)
+		if err != nil {
+			h.log.Critical(err)
+		}
+	}()
+
 	// Perform a loop that will allow downloads to happen until the maximum
 	// time for a single connection has been reached.
 	for time.Now().Before(startTime.Add(1200 * time.Second)) {
--- a/modules/host/negotiatedownload.go
+++ b/modules/host/negotiatedownload.go
@@ -1,12 +1,25 @@
 package host
 
 import (
+	"errors"
 	"net"
 	"time"
 
+	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/encoding"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/types"
+)
+
+var (
+	// errLargeDownloadBatch is returned if the renter requests a download
+	// batch that exceeds the maximum batch size that the host will
+	// accomondate.
+	errLargeDownloadBatch = errors.New("a batch of download requests has exceeded the maximum allowed download size")
+
+	// errRequestOutOfBounds is returned when a download request is made which
+	// asks for elements of a sector which do not exist.
+	errRequestOutOfBounds = errors.New("a download request has been made that exceeds the sector boundaries")
 )
 
 // managedDownloadIteration is responsible for managing a single iteration of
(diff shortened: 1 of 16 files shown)


modules/host/negotiatedownload.go
```diff
@@ -0,0 +1,230 @@
...
+			sectorData, err := h.readSector(request.MerkleRoot)
+			if err != nil {
+				return err
+			}
+			payload = append(payload, sectorData[request.Offset:request.Offset+request.Length])
+		}
+
+		// Read the transaction signature from the renter.
+		var renterSignature types.TransactionSignature
+		err = encoding.ReadObject(conn, &renterSignature, 16e3)
```
lukechampine: in revise, the renter sends the revision, host sends acceptance, then renter sends signatures. Here, the renter sends the revision and the signatures, then host sends acceptance. Can we standardize on one?

DavidVorick: yeah, the first

modules/host/negotiatedownload.go
```diff
@@ -0,0 +1,230 @@
...
+
+var (
+	// errLargeDownloadBatch is returned if the renter requests a download
+	// batch that exceeds the maximum batch size that the host will
+	// accomondate.
+	errLargeDownloadBatch = errors.New("a batch of download requests has exceeded the maximum allowed download size")
+
+	// errRequestOutOfBounds is returned when a download request is made which
+	// asks for elements of a sector which do not exist.
+	errRequestOutOfBounds = errors.New("a download request has been made that exceeds the sector boundaries")
```
lukechampine: passive voice is uncommon for error messages. Usually it's more like "download request exceeds maximum batch size" and "download request has invalid bounds"

modules/host/negotiatedownload.go
```diff
@@ -0,0 +1,230 @@
...
+			return errors.New("unstable revision")
+		}
+		if paymentRevision.NewWindowStart != existingRevision.NewWindowStart {
+			return errors.New("unstable revision")
+		}
+		if paymentRevision.NewWindowEnd != existingRevision.NewWindowEnd {
+			return errors.New("unstable revision")
+		}
+		if paymentRevision.NewUnlockHash != existingRevision.NewUnlockHash {
+			return errors.New("unstable revision")
```
lukechampine: these errors won't be very helpful for debugging...

modules/host/negotiateformcontract.go
```diff
@@ -130,14 +125,54 @@ func (h *Host) managedAddCollateral(txnSet []types.Transaction, settings modules
...
+		ParentID: contractTxn.FileContractID(0),
+		UnlockConditions: types.UnlockConditions{
+			PublicKeys: []types.SiaPublicKey{
+				{
+					Algorithm: types.SignatureEd25519,
+					Key:       renterPK[:],
+				},
+				{
+					Algorithm: types.SignatureEd25519,
+					Key:       hostPK[:],
```
lukechampine: couldn't you just use `h.publicKey`?

modules/host/negotiaterecentrevision.go
```diff
@@ -12,43 +14,55 @@ import (
...
 	var revisionSigs []types.TransactionSignature
 	for _, sig := range revisionTxn.TransactionSignatures {
+		// Checking for just the parent id is sufficient, an over-signed file
+		// contract is invalid.
 		if sig.ParentID == crypto.Hash(fcid) {
 			revisionSigs = append(revisionSigs, sig)
 		}
 	}
 
+	// Verify that the challegne response matches public key.
```
lukechampine: typo

modules/host/negotiaterecentrevision.go
```diff
@@ -57,18 +71,63 @@ func (h *Host) managedRPCRecentRevision(conn net.Conn) (types.FileContractID, er
...
+	if err != nil {
+		return types.FileContractID{}, nil, err
+	}
+
+	// Send a challenge to the renter to verify that the renter has write
+	// access to the revision being opened.
+	var challenge crypto.Hash
+	_, err = rand.Read(challenge[:])
+	if err != nil {
+		return types.FileContractID{}, nil, modules.WriteNegotiationRejection(conn, err)
```
lukechampine: renter is not expecting a response here

modules/host/negotiaterevisecontract.go
```diff
@@ -91,74 +90,80 @@ func (h *Host) managedRevisionIteration(conn net.Conn, so *storageObligation) er
...
-			sectorsGained = append(sectorsGained, newRoot)
-			gainedSectorData = append(gainedSectorData, sector)
-			so.SectorRoots[modification.SectorIndex] = newRoot
-		} else {
-			return errUnknownModification
 		}
+		return nil
+	}()
+	if err != nil {
+		return modules.WriteNegotiationRejection(conn, err)
```
lukechampine: renter is not expecting a response here

modules/host/negotiatedownload.go
DavidVorick: it's not done, they were thrown there quickly so you would have a reference to write your code against. Fixing up in more detail now.

modules/host/negotiateformcontract.go
DavidVorick: no, needs to be a SigEd25519 public key because the host needs to be able to recognize the algorithm during the challenge->response.

=/ kind of locked into ed25519 at this point.

modules/host/negotiaterevisecontract.go
DavidVorick: it is now :P

modules/host/negotiateformcontract.go
lukechampine: Doesn't the host know what algorithm the renter used in the original transaction? If so, it can create an algorithm-specific challenge. Of course, there's no need to support anything but ed25519 for the time being, but it seems preferable to make the RPC algorithm-agnostic.

DavidVorick: yes, but that requires the host to actually know the algorithm. Right now the only algorithm that the host knows is ed25519

lukechampine: presumably, the host will reject a file contract if the renter uses an unrecognized algorithm. So it can guarantee that it will always be able to create proofs.

PR #1077, opened 2016-04-21
Renter download

lukechampine:
Ok, there's more than just download going on here...

There's a race condition with the storageObligation locks :I
Seems like you need to grab a host lock before you check the storageObligation lock map.


DavidVorick:
oh, you definitely do, that's my fault. Do you want to fix it, or should I? Make sure that unlock is called correctly, and that you apply it to all 3 functions that grab a storage obligation lock (formcontract, revisecontract, and download)

PR #1078, opened 2016-04-21
Storage manager

DavidVorick:
This PR splits out the storage folder and sector management from the rest of the host, putting it in its own package. I have not finished updating the documentation, and I have not finished uncommenting all of the tests, the only test that remains is the smoke test. I uncommented most of the other tests, and did not run into any problems. All of the host tests are passing to the full extent that they were before this move happened.

Because the functions are now part of an exported external module, some of the concurrency conventions are being violated in the host (calling exported functions while under lock). I will clean that up in the coming week.

The host and storage layer are now properly separated, with an interface defining everything that needs to be happening for someone to write a separate one. Anyone looking to optimize or tweak the storage layer should have an easier time doing so, and will not threaten the other functions of the host by poking around.


modules/host/storagemanager/sector.go
```diff
@@ -40,7 +40,7 @@ import (
 // read from the filesystem, which means the problem manifests at the lowest
 // level, the sector level. Because data is missing, there is no 'repair'
 // operation that can be supported. The sector usage database should match the
-// storage obligation database, and should be patched if there's a mismatch.
+// storage obligation database, and should be patched if there's a mismatcsm.
```
lukechampine: why

lukechampine:
`expiryDate` is definitely confusing.

modules/host/storagemanager/sector.go
DavidVorick: NOOOOOOOOOOOOOOOOOOOOOOOOOOOOOO

DavidVorick: why I can't I hold all these hands?

PR #1079, opened 2016-04-21
Renter API + siac

lukechampine:
I forgot I already wrote the API code. So this is just siac stuff mainly.

We should go through the `formContract` code together and make sure it isn't doing anything really stupid, keeping in mind that some naive/suboptimal behavior is permissible for the RC.


DavidVorick:
Are these all of the changes that need to be made? Overall there is not much happening here.

PR #1081, opened 2016-04-22
Host api

DavidVorick:
In this pull request, I add some routes for the host and the storage manager. These routes should be everything that is necessary for v0.6.0-rc1.

I also restructured modules/host.go a little, mostly to simplify things as much as possible. Overall, I'm quite happy with modules/host.go and modules/storagemanager.go (except for the expiryHeight thing), and this is just more polish on top of what was already pretty good.

Siac + testing to come shortly.

### DIFF
--- a/api/host.go
+++ b/api/host.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"errors"
 	"fmt"
 	"net/http"
 
@@ -8,6 +9,13 @@ import (
 	"github.com/NebulousLabs/Sia/modules"
 
 	"github.com/julienschmidt/httprouter"
+)
+
+var (
+	// errStorageFolderNotFound is returned if a call is made looking for a
+	// storage folder which does not appear to exist within the storage
+	// manager.
+	errStorageFolderNotFound = errors.New("storage folder with the provided path could not be found")
 )
 
 type (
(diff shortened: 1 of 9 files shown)


api/host.go
```diff
@@ -1,153 +1,227 @@
...
+	folderIndex, err := folderIndex(folderPath, storageFolders)
+	if err != nil {
+		writeError(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	var force bool
+	_, err = fmt.Sscan(req.FormValue("force"), &force)
+	if err != nil {
+		writeError(w, err.Error(), http.StatusBadRequest)
```
lukechampine: this means `force` is required in all requests. Better would be:
```go
force := false
if req.FormValue("force") != "" {
    _, err = fmt.Sscan(req.FormValue("force"), &force)
}
```
or even simply:
```go
force := req.FormValue("force") == "true"
```
which is the approach used by the renter upload API call.

api/host.go
```diff
@@ -1,153 +1,227 @@
...
+// folderIndex determines the index of the storage folder with the provided
+// path.
+func folderIndex(folderPath string, storageFolders []modules.StorageFolderMetadata) (int, error) {
+	for i, sf := range storageFolders {
+		if sf.Path == folderPath {
+			return i, nil
+		}
+	}
+	return -1, errStorageFolderNotFound
+}
```
lukechampine: seeks like a hack to do this in the API server rather than in the host, but I can understand if you don't want to rewrite all the host methods right now

DavidVorick: hmm. I would rather have the interface do it by index, but maybe there's no justification for that. I feel like, programatically, it's easier to manage than to know the paths of all the folders that you've set up.

lukechampine: from an outsider perspective, a map from path -> folder object doesn't seem any harder to manage than a slice of folder objects. But I'm not familiar enough with the implementation to know for sure.

PR #1084, opened 2016-04-22
misc fixes for an integration test

DavidVorick:
I tried writing an integraiton test where the host was uploading a file
to the renter. This was to test the host, but as a first step the renter
needs to be working. I got stuck when the file would never get beyond
0%.

After the siac stuff is ready, we should work together to get the integration tests passing.

### DIFF
--- a/api/server_helpers_test.go
+++ b/api/server_helpers_test.go
@@ -186,6 +186,21 @@ func createServerTester(name string) (*serverTester, error) {
 		}
 	}
 
+	// Announce the host.
+	err = st.announceHost()
+	if err != nil {
+		return nil, err
+	}
+
+	// Set an allowance for the renter.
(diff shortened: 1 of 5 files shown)


api/server_helpers_test.go
```diff
@@ -186,6 +186,21 @@ func createServerTester(name string) (*serverTester, error) {
...
 
+	// Announce the host.
+	err = st.announceHost()
+	if err != nil {
+		return nil, err
+	}
+
+	// Set an allowance for the renter.
+	allowanceValues := url.Values{}
+	allowanceValues.Set("funds", "10000000000000000000000000000")
```
lukechampine: if you want something more readable, you can use a `types.Currency` and its `String` method.

PR #1085, opened 2016-04-22
add siac host commands

lukechampine:


siac/hostcmd.go
```diff
@@ -45,8 +49,137 @@ You may also supply a specific address to be announced, e.g.:
...
+		return
+	}
+	w := tabwriter.NewWriter(os.Stdout, 0, 0, 4, ' ', 0)
+	fmt.Fprintln(w, "\nUsed\tCapacity\t% Used\tPath")
+	for _, folder := range sg.StorageFolderMetadata {
+		curSize := int64(folder.Capacity - folder.CapacityRemaining)
+		pctUsed := 100 * (float64(curSize) / float64(folder.Capacity))
+		fmt.Fprintf(w, "%s\t%s\t%.2f\t%s\n", filesizeUnits(curSize), filesizeUnits(int64(folder.Capacity)), pctUsed, folder.Path)
+	}
+	w.Flush()
```
lukechampine: Example output:
```
Host info:
	Storage:      20.000 GB (0 B used)
	Price:        19999 SC per TB per month
	Max Duration: 25920 Blocks

	Accepting Contracts: No
	Anticipated Revenue: 0
	Revenue:             0
	Lost Revenue:        0

	Net Address: 

RPC Stats:
	Error Calls:        0
	Unrecognized Calls: 1
	Download Calls:     0
	Renew Calls:        0
	Revise Calls:       0
	Settings Calls:     0
	FormContract Calls: 0

Used    Capacity     % Used    Path
0 B     10.000 GB    0.00      /home/luke/foo
0 B     10.000 GB    0.00      /home/luke/bar
```

siac/hostcmd.go
```diff
@@ -45,8 +49,137 @@ You may also supply a specific address to be announced, e.g.:
...
+		Long: `Change how much data a storage folder should store. If the new size is less
+than what the folder is currently storing, data will be distributed across the
+other storage folders.`,
+		Run: wrap(hostfolderresizecmd),
+	}
+
+	hostSectorCmd = &cobra.Command{
+		Use:   "sector",
+		Short: "Add or delete a sector",
+		Long:  "Add or delete a sector. Note that deleting a sector may impact host revenue.",
```
DavidVorick: There is no 'add sector' support right now.

lukechampine: true, but I felt like just `Delete a sector` would be misleading.

siac/hostcmd.go
DavidVorick: Example output:
```
Host info:
	Storage:      20.000 GB (0 B used)
	Price:        19999 SC per TB per month
	Max Duration: 25920 Blocks

	Accepting Contracts: No
	Anticipated Revenue: 0
	Revenue:             0
	Lost Revenue:        0

	Net Address: 

RPC Stats:
	Error Calls:        0
	Unrecognized Calls: 1
	Download Calls:     0
	Renew Calls:        0
	Revise Calls:       0
	Settings Calls:     0
	FormContract Calls: 0

Storage Stats:
    Used    Capacity     % Used    Path
    0 B     10.000 GB    0.00      /home/luke/foo
    0 B     10.000 GB    0.00      /home/luke/bar
```

DavidVorick: Jordan does not like my example

siac/hostcmd.go
```diff
@@ -45,8 +49,137 @@ You may also supply a specific address to be announced, e.g.:
...
+		Long: `Delete a sector, identified by its Merkle root. Note that deleting a
+sector may impact host revenue.`,
+		Run: wrap(hostsectordeletecmd),
+	}
 )
 
+// hostcmd is the handler for the command `siac host`.
+// Prints info about the host and its storage folders.
+func hostcmd() {
+	hg := new(api.HostGET)
```
VoidingWarranties: Why use new here instead of `var hg api.HostGET`?

siac/hostcmd.go
```diff
@@ -45,8 +49,137 @@ You may also supply a specific address to be announced, e.g.:
...
+	accept := yesNo(is.AcceptingContracts)
+	// convert price to SC/TB/mo
+	price, err := modules.StoragePriceToHuman(is.MinimumStoragePrice)
+	if err != nil {
+		price = ^uint64(0)
+	}
+	totalRevenue := fm.ContractCompensation.
+		Add(fm.StorageRevenue).
+		Add(fm.DownloadBandwidthRevenue).
+		Add(fm.UploadBandwidthRevenue)
```
VoidingWarranties: I didn't know you could do this, cool!

siac/hostcmd.go
```diff
@@ -45,8 +49,137 @@ You may also supply a specific address to be announced, e.g.:
...
+		totalstorage += folder.Capacity
+		storageremaining += folder.CapacityRemaining
+	}
+
+	// convert accepting bool
+	accept := yesNo(is.AcceptingContracts)
+	// convert price to SC/TB/mo
+	price, err := modules.StoragePriceToHuman(is.MinimumStoragePrice)
+	if err != nil {
+		price = ^uint64(0)
```
VoidingWarranties: Why do we use the max price on error?

siac/hostcmd.go
DavidVorick: `new` gives you a pointer, if you do `var hg api.HostGET` you get the actual object, and if you do `var hg *api.HostGET` you get a nil pointer. The way above gives you a pointer to an object that is initialized.

siac/hostcmd.go
DavidVorick: There should be a comment here, or better yet an explicit error check. Right now, the only error that can be returned by the `StoragePriceToHuman` function is if there are more coins than can be represented by a uint64, but that's not obvious just by looking at this code.

siac/hostcmd.go
DavidVorick: yeah it was new to me too

siac/hostcmd.go
lukechampine: you don't need a pointer though. In most cases we should be doing `var hg api.HostGET`.

siac/hostcmd.go
lukechampine: it's more common in dot-heavy languages like javascript. They like to abuse this pattern to chain tons and tons of actions together.

siac/hostcmd.go
VoidingWarranties: But here you are both using `new` and using `&hg` to take the address of the value when passing it to `getAPI`. Isn't this a pointer to a pointer? Why is this necessary?

siac/hostcmd.go
VoidingWarranties: +1 for explicit error check

siac/hostcmd.go
lukechampine: huh. Right you are. JSON's unmarshalling rules prevent this causing problems, but it's definitely wrong

PR #1091, opened 2016-04-24
Make fsync optional when saving a file

DavidVorick:
Fsync is really slow, to the point that it was causing signficant
performance issues. Instead of file saving calling fsync by default for
everything, fsync is only called for user action and for really
important action that pertains to monetary changes or changes to network
contracts. The result should be a significantly faster siad.

Further investigation has led me to believe that the bottleneck for some of the startup is actually not I/O, and not fsync, but instead is hashing. I also remember making this conclusion before. Regardless, we do a lot less fsync now, only during important calls. This might be worth looking over to figure out where I made what decisions, and to dispute any that you disagree with.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,7 @@ vet: release-std
 	go vet $(pkgs)
 
 # will always run on some packages for a while.
-lintpkgs = ./modules/host ./modules ./modules/renter/hostdb ./modules/renter/contractor
+lintpkgs = ./modules/host ./modules ./modules/renter/hostdb ./modules/renter/contractor ./persist
 lint:
 	@for package in $(lintpkgs); do                           \
 		golint -min_confidence=1.0 $$package                  \
--- a/modules/gateway/gateway.go
+++ b/modules/gateway/gateway.go
@@ -58,7 +58,7 @@ func (g *Gateway) Close() error {
 
 	// save the latest gateway state
 	id := g.mu.RLock()
-	if err := g.save(); err != nil {
+	if err := g.save(true); err != nil {
 		errs = append(errs, fmt.Errorf("save failed: %v", err))
 	}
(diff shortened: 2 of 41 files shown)


lukechampine:
Seeing "magic booleans" all over the place makes me sad. I might actually prefer explicit `saveSync` methods to a boolean, given that sync'd saves are now fairly uncommon.

As discussed, I think fsync should primarily be used when failure to write to disk could result in a loss of coins (including losing track of resources which we spent coins on, like contracts). Saving on `Close` doesn't seem strictly necessary, though I suppose efficiency is not a major concern during shutdown.

We should consider the worst-case scenario: siad is mid-API call when power failure occurs. What will the user do when they reboot? What behavior will they expect? Bear in mind that we already have atomic files, so recovery should usually be possible. ~~The singular benefit of `fsync` is that we can say: "If the call returned `nil`, the data has hit the disk."~~ Actually, we don't even get that much. `fsync` flushes data to the drive, but the drive may maintain its own internal buffers, which will be lost in the event of power failure (though not system crash).

Things like peer lists, metrics, logs, etc. are not fsync-worthy. I'll let you worry about the wallet, but as for the gateway, renter, contractor, and hostdb, I think fsync is only necessary for saving new contracts and new files.

DavidVorick:
I thought having a bool was cleaner, but maybe it would be better to define a new function in every module. We're looking at the difference between a boolean added to all the save functions, or to doubling up the save functions so that there are two save functions for each module. 

I tried to follow this strategy, it closely matches what you've described: fsync for file contract stuff (host + renter), during shutdown, and if the user explicitly makes a change (like SetSettings). Fsync is not that slow, it's only when you start calling it every block for a 40,000 block IBD or rescan that it becomes a bottleneck.

PR #1092, opened 2016-04-24
performance and correctness updates to miner

DavidVorick:
During IBD and loading, some hashing in the miner module was causing
visible slowdown. While investigating other potential sources of
slowdown, I also discovered some locking misuse.

Removing the hashes resulted in about a 10% speedup for the miner accepting an average sized block (where average was defined by using the real chain). `FindBlock` had an outright deadlock in it.  An unchecked error is now checked because the miner now has a logger that can record the error.


lukechampine:
Maybe `types` should define `GenesisBlockID` so that we never need to hash it more than once. grep says that the host, hostdb, and contractor all hash the genesis block on every update.

DavidVorick:
I think that would be worthwhile.

PR #1093, opened 2016-04-24
fix locking in consensus package

DavidVorick:
There were a few locking bugs in the consensus package, places where
the database was being accessed outside of a read lock.

There was one critical place in particular, which was using a readlock
followed by a writelock, when it really needed to be using a single
writelock -> demotelock. This place is likely the cause of the consensus
desynchronizations between the miner + consensus package, has likely
impacted other modules, and since fixing it I have not been able to
trigger the behavior in question.

### DIFF
--- a/modules/consensus/subscribe.go
+++ b/modules/consensus/subscribe.go
@@ -153,18 +153,17 @@ func (cs *ConsensusSet) initializePersistentSubscribe(subscriber modules.Consens
 // As a special case, using an empty id as the start will have all the changes
 // sent to the modules starting with the genesis block.
 func (cs *ConsensusSet) ConsensusSetPersistentSubscribe(subscriber modules.ConsensusSetSubscriber, start modules.ConsensusChangeID) error {
+	cs.mu.Lock()
+	cs.subscribers = append(cs.subscribers, subscriber)
+	cs.mu.Demote()
+	defer cs.mu.DemotedUnlock()
+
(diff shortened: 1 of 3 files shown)


modules/consensus/subscribe.go
```diff
@@ -153,18 +153,17 @@ func (cs *ConsensusSet) initializePersistentSubscribe(subscriber modules.Consens
...
 	if err != nil {
 		return err
 	}
-
 	// Only add the module as a subscriber if there was no error.
-	cs.mu.Lock()
-	cs.subscribers = append(cs.subscribers, subscriber)
-	cs.mu.Unlock()
 	return nil
 }
```
DavidVorick: This function is likely the cause of the problems users have been reporting with the miner being desynchronized from the consensus set. A subtle race condition where the function was letting go of the lock before adding a subscriber to the subscribers list, meaning that the subscriber could miss entire consensus changes. This would cause the behavior we were observing.

initializePersistentSubscribe can take a long time, especially if the machine is slow or if the subscriber is missing multiple updates (more likely for the miner because we removed the 'save on every block' function from the miner, meaning at startup the miner might receive quite a few for not having saved during the last session), and in that time `AcceptBlock` may request a write lock. When the readlock releases, the write lock from `AcceptBlock` will get priority over the write lock in this function, meaning that a block (or multiple blocks) would make it through the consensus set, and the subscriber would never get those updates.

Now fixed :)

PR #1095, opened 2016-04-24
Consensus synchronize

DavidVorick:
I'm trying to track down the travis bug where the consensus synchronize is failing nondeterministically. This PR is not ready to be merged, and I'll probably be pushing to it a bunch of times and rebuilding a bunch of times until I can figure out why  the  test is failing.


DavidVorick:
Update: I think I have tracked down the problem, this should eliminate our most common nondeterministic build failure.

PR #1097, opened 2016-04-24
Add api call for HostExternalSettings

DavidVorick:
Pretty straightforward.

### DIFF
--- a/modules/host/negotiatesettings.go
+++ b/modules/host/negotiatesettings.go
@@ -12,26 +12,20 @@ import (
 // capacity returns the amount of storage still available on the machine. The
 // amount can be negative if the total capacity was reduced to below the active
 // capacity.
-func (h *Host) capacity() (total, remaining uint64, err error) {
+func (h *Host) capacity() (total, remaining uint64) {
 	// Total storage can be computed by summing the size of all the storage
 	// folders.
-	sfs, err := h.StorageFolders()
-	if err != nil {
-		return 0, 0, err
-	}
+	sfs := h.StorageFolders()
 	for _, sf := range sfs {
 		total += sf.Capacity
 		remaining += sf.CapacityRemaining
 	}
-	return total, remaining, nil
+	return total, remaining
 }
 
 // externalSettings compiles and returns the external settings for the host.
-func (h *Host) externalSettings() (modules.HostExternalSettings, error) {
-	totalStorage, remainingStorage, err := h.capacity()
-	if err != nil {
-		return modules.HostExternalSettings{}, err
-	}
+func (h *Host) externalSettings() modules.HostExternalSettings {
+	totalStorage, remainingStorage := h.capacity()
 	var netAddr modules.NetAddress
 	if h.settings.NetAddress != "" {
 		netAddr = h.settings.NetAddress
@@ -61,7 +55,7 @@ func (h *Host) externalSettings() (modules.HostExternalSettings, error) {
 
 		RevisionNumber: h.revisionNumber,
 		Version:        build.Version,
-	}, nil
+	}
 }
 
 // managedRPCSettings is an rpc that returns the host's settings.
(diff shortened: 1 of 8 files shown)


modules/host/negotiatesettings.go
```diff
@@ -26,54 +26,60 @@ func (h *Host) capacity() (total, remaining uint64, err error) {
 	return total, remaining, nil
 }
 
+// externalSettings compiles and returns the external settings for the host.
+func (h *Host) externalSettings() (modules.HostExternalSettings, error) {
+	totalStorage, remainingStorage, err := h.capacity()
+	if err != nil {
+		return modules.HostExternalSettings{}, err
+	}
```
lukechampine: hmm, seems strange that returning the external settings can fail. I feel like it should be possible to track capacity without needing to read from disk every time.

DavidVorick: uh, yeah. I didn't think too much of it, but I might be able to drop that error.

DavidVorick: Oh. So, there are actually a handful of errors here, but the biggest is that the call to `StorageFolders` returns more than capacity, it calls `filepath.EvalSymlinks` to return the the path of the storage folder. It also grabs a resource lock but I don't think it needs to do that, the function doesn't actually seem to use any resources that can close.

If we can figure out some other way to handle an error from `filepath.EvalSymlinks`, then we should be able to drop the error. Capacity itself is tracked in memory. I guess instead of calling EvalSymlinks during `StorageFolders()` I could store the path when the directory gets created.

I'm wondering if using symlinks was the right decision. In hindsight, it seems a little strange, but I do think that you get to keep the portability and everything. Right now, the host does all operations inside of its own folder, and when you add a storage folder it makes a symlink to the folder that you added, and then forgets that the original ever existed. Which is why you need to evaluate the symlinks to remember where the original storage folder was.

I suppose we could have the whole thing be in a .json file, but I think using symlinks made a lot of the operations a lot simpler. Plus, in a file manager, all of your host data is going to be in one place (via symlinks) which makes everything a lot easier to browse. Not that there's any reason to do that beyond natural curiosity, but still it's a nice bonus.

DavidVorick: Another approach to ditching the error is to just return an error string ("problems parsing folder") as the name if there's an error. Then you still return all of the rest of the information. That's definitely more graceful than not being able to see the status of *any* of your storage folders if only one of them is having problems.

lukechampine: symlinks do seem a little strange. If I make a folder `/bar` and tell the host to track it, I would expect the host to have some internal state that contains `/bar`. As for browsing, I'm skeptical that there's much utility to being able to navigate to all your storage folders from one place. Why would you want to browse your storage folders anyway? And even if I did want to browse them, my first thought would be to run `siac host` and then navigate to the listed folders, not `cd` to my Sia/host directory and follow a bunch of symlinks.

The latter scenario is a great example of where a generic `Either` type would be useful, i.e. `Either<StorageFolder, error>`. Then you could return a slice of those. We could simulate it by making a new type, or return a slice of pointers (`nil` meaning an unspecified error occurred, though it would be logged), or return a slice of errors, or return a single error alongside the valid entries. I think my preference would be the latter, using a custom error type so that the valid entries could be recovered:
```go
sfs, err := host.StorageFolders()
if sfe, ok := err.(*host.StorageFolderError); ok {
    for _, f := range sfe.ErrFolders {
        fmt.Println("Warning: could not parse folder", f)
    }
} else if err != nil {
    return err
}
// this path will still be reached if the error is a StorageFolderError
for _, f := range sfe.Folders {
    fmt.Println(f)
}
```

DavidVorick:
Does our API support returning an error alongside a bunch of valid data? It's not a paradigm we've used yet anywhere in the codebase.

lukechampine:
true, I was only thinking about the internal API, not the HTTP API. I don't think returning both data and an error is a viable option for the API.

PR #1098, opened 2016-04-25
Readme build instructions

DavidVorick:
Grabbed the commit from @mtlynch and then made some tweaks. It does seem like we can only get the build to pass right now for branches that are directly on the Sia repo, which means we'll have to manually transport any pull requests =/

We'll have to figure out a better solution for that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,10 +27,10 @@ record, we advise against using it as a sole means of storing important data.
 
 This release comes with 2 binaries, siad and siac. siad is a background
 service, or "daemon," that runs the Sia protocol, and siac is a client that is
-used to interact with siad. Siad exposes an API on 'localhost:9980' which can
-be used to interact with the daemon. There is a front-end program called Sia-UI
-which can be used to interact with the daemon in a more user-friendly way.
-Documentation on the API can be found in doc/API.md.
+used to interact with siad. Siad exposes an HTTP API on 'localhost:9980' which
+can be used to interact with the daemon. There is a front-end program called
+Sia-UI which can be used to interact with the daemon in a more user-friendly
+way. Documentation on the API can be found in doc/API.md.
 
 Usage
 -----
@@ -50,24 +50,22 @@ Building From Source
 --------------------
 
 To build from source, [Go 1.6 must be installed](https://golang.org/doc/install)
-on the system. Then run the following command to build and install:
+on the system. Then simply use `go get`:
 
 ```
-go get -u github.com/NebulousLabs/Sia
+go get github.com/NebulousLabs/Sia/...
 ```
 
-To start `siad`, the Sia daemon process, run:
+This will download the Sia repo to your `$GOPATH/src` folder, and install the
+`siad` and `siac` binaries in your `$GOPATH/bin` folder.
 
-```
-$GOPATH/bin/siad
-```
-
-`siad` is a daemon which runs in the background and collaborates with the Sia
-network. The daemon can be controlled via an http api, or by `siac`, a command
-line utility that can be run from a separate shell:
-```
-$GOPATH/bin/siac
-```
+To stay up-to-date, you can run the previous `go get` command with the `-u`
+flag. Alternatively, you can use the Makefile provided in this repo. Run
+`git pull origin master` to pull the latest changes, and `make release-std` to
+build the new binaries. You can also run `make test` and `make test-long` to
+run the short and full test suites, respectively. Finally, `make cover` will
+generate code coverage reports for each package; they are stored in the `cover`
+folder and can be viewed in your browser.
 
 Troubleshooting
 ---------------
--- a/README.md
+++ b/README.md
@@ -46,6 +46,32 @@ access." You can now run `siac` in a separate command prompt to interact with
 siad. From here, you can send money, mine blocks, upload and download
 files, and advertise yourself as a host.
 
+Building From Source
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -53,10 +53,8 @@ To build from source, [Go 1.6 must be installed](https://golang.org/doc/install)
 on the system. Then run the following to build:
 
 ```
-go get github.com/NebulousLabs/Sia
-cd  $GOPATH/src/github.com/NebulousLabs/Sia/
-make dependencies
-make
+go get -u github.com/NebulousLabs/Sia
+go install github.com/NebulousLabs/Sia/...
```
lukechampine: `go get` runs `go install automatically

lukechampine:
We could vendor the requisite packages, so that we no longer need to fetch them from google.

README.md
```diff
@@ -46,6 +46,30 @@ access." You can now run `siac` in a separate command prompt to interact with
...
 
+Building From Source
+--------------------
+
+To build from source, [Go 1.6 must be installed](https://golang.org/doc/install)
+on the system. Then run the following to build:
+
+```
+go get -u github.com/NebulousLabs/Sia
+go install github.com/NebulousLabs/Sia/...
```
lukechampine: `go get` runs `go install` automatically

DavidVorick:
I am not in favor of that option, I hope there's something better. But I do agree that it's better than having 100% of third party PRs fail, so if we don't find another option I'm willing to do that.

DavidVorick:
updated to remove the install call

README.md
```diff
@@ -46,6 +46,29 @@ access." You can now run `siac` in a separate command prompt to interact with
...
 files, and advertise yourself as a host.
 
+Building From Source
+--------------------
+
+To build from source, [Go 1.6 must be installed](https://golang.org/doc/install)
+on the system. Then run the following command to build and install:
+
+```
+go get -u github.com/NebulousLabs/Sia
```
lukechampine: this won't work. You need the `...`

VoidingWarranties:
I've heard people suggest an all or nothing approach to vendoring ([source](https://blog.gopheracademy.com/advent-2015/vendor-folder/)), although I'm not entirely sure what their reasoning is.

DavidVorick:
@lukechampine it's a good idea to run `go get -u` on your first build, because if you've got any old dependencies I don't think go will automatically update them. I think at this point we've vendored every package that's ever made a breaking change, so it shouldn't be a problem here but it doesn't hurt to make sure everything's up to date before pulling down Sia.

DavidVorick:
your other changes look good

lukechampine:
fair enough

DavidVorick:
rebased on to master.

4 commits to update the readme with build instructions. lol

README.md
```diff
@@ -46,6 +46,27 @@ access." You can now run `siac` in a separate command prompt to interact with
...
+```
+go get -u github.com/NebulousLabs/Sia/...
+```
+
+This will download the Sia repo to your `$GOPATH/src` folder, and install the
+`siad` and `siac` binaries in your `$GOPATH/bin` folder.
+
+To stay up-to-date, run the previous `go get` command again. Alternatively, you
+can use the Makefile provided in this repo. Run `git pull origin master` to
+pull the latest changes, and `make release-std` to build the new binaries. You
```
mtlynch: On a clean install of Ubuntu 14 server, `make release-std` fails

```
$ make release-std
go install -a ./api ./build ./compatibility ./crypto ./encoding ./modules ./modules/consensus ./modules/explorer ./modules/gateway ./modules/host ./modules/host/storagemanager ./modules/renter/hostdb ./modules/renter/contractor ./modules/miner ./modules/renter ./modules/wallet ./modules/transactionpool ./persist ./siac ./siad ./sync ./types
go install runtime/internal/sys: open /usr/local/go/pkg/linux_amd64/runtime/internal/sys.a: permission denied
make: *** [release-std] Error 1
```

Is that expected? `make` works fine.

lukechampine: looks like you need to change permissions on `$GOPATH/pkg`

DavidVorick: What version of go was that? It looks like the permissions of either your $GOROOT or $GOPATH are not allowing you to build files into them, both should be writeable by your user.

mtlynch: This is on Go 1.6.2 following the [standard install instructions](https://golang.org/doc/install). To write `/usr/local` on Ubuntu 14 server, you have to be root, so extracting the tar needs to run as sudo.

Then that file in the archive is not writable except by owner:

```
$ tar -ztvf go1.6.2.linux-amd64.tar.gz | grep pkg/linux_amd64/runtime/internal/sys.a
-rw-r--r-- 0/0            3026 2016-04-19 19:51 go/pkg/linux_amd64/runtime/internal/sys.a
```

lukechampine: does it work without `-a`?

mtlynch: Yeah, that works

DavidVorick: so now the question... what do we do to prevent other users from running into this problem? Do we remove the `-a` flag, or is there something more subtle? I do want to make sure that anyone running `make release` is going to be getting a fully updated binary.

lukechampine: According to https://github.com/golang/go/commit/4dab6d01f12591f256d36b32cd6480ef679458f1, the `-a` flag should no longer be necessary.

PR #1100, opened 2016-04-25
Make fsync optional when saving a file

DavidVorick:
Fsync is really slow, to the point that it was causing signficant
performance issues. Instead of file saving calling fsync by default for
everything, fsync is only called for user action and for really
important action that pertains to monetary changes or changes to network
contracts. The result should be a significantly faster siad.

### DIFF
--- a/persist/boltdb.go
+++ b/persist/boltdb.go
@@ -42,8 +42,8 @@ func (db *BoltDatabase) checkMetadata(md Metadata) error {
 	return err
 }
 
-// updatbMetadata will set the contents of the metadata bucket to be
-// what is stored inside the metadata argument
+// updateMetadata will set the contents of the metadata bucket to the values
+// in db.Metadata.
 func (db *BoltDatabase) updateMetadata(tx *bolt.Tx) error {
 	bucket, err := tx.CreateBucketIfNotExists([]byte("Metadata"))
 	if err != nil {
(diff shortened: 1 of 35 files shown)


DavidVorick:
This PR is obviously at odds with #1091 - only one of them should be merged.

persist/boltdb.go
```diff
@@ -60,7 +42,25 @@ func (db *BoltDatabase) checkMetadata(md Metadata) error {
 	return err
 }
 
-// CloseDatabase saves the bolt database to a file, and updates metadata
+// updatbMetadata will set the contents of the metadata bucket to be
```
lukechampine: typo

lukechampine:
The obvious downside to this approach is the code duplication. We don't want people updating `save` and forgetting to update `saveSync` as well. Its easy enough to consolidate, but then we wind up with three `save` functions per module, which is sad. Still, I think I prefer this approach.

persist/json.go
```diff
@@ -64,16 +78,16 @@ func SaveFile(meta Metadata, data interface{}, filename string) error {
...
-	file, err := os.Open(filename)
+// SaveFileSync atomically saves json data to a file and then syncs to disk.
+func SaveFileSync(meta Metadata, data interface{}, filename string) error {
+	file, err := NewSafeFile(filename)
 	if err != nil {
 		return err
 	}
 	defer file.Close()
-	err = Load(meta, data, file)
+	err = Save(meta, data, file)
```
lukechampine: this is the most deceptive diff I've ever seen, lol

DavidVorick: Holy cow. That's terrible!

PR #1101, opened 2016-04-25
API authentication

VoidingWarranties:
Moving #1065 here so that branch is merged into staging branch instead of master so as not to disturb v0.6.0 RCs.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -141,9 +141,6 @@ func (srv *Server) initAPI(password string) {
 		router.POST("/miner/header", requirePassword(srv.minerHeaderHandlerPOST, password))
 		router.GET("/miner/start", requirePassword(srv.minerStartHandler, password))
 		router.GET("/miner/stop", requirePassword(srv.minerStopHandler, password))
-		// TODO: doesn't adding authentication break compatibility?
-		router.GET("/miner/headerforwork", requirePassword(srv.minerHeaderHandlerGET, password))  // COMPATv0.4.8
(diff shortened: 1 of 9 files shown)


VoidingWarranties:
There was some debate as to what endpoints to authenticate. There are three options:
1) authenticate everything
2) authenticate state modifying calls and calls that expose sensitive data (like wallet seeds)
3) authenticate endpoints likely to be attacked (pretty much only wallet endpoints)

This PR originally authenticated every endpoint, but it was annoying to have to type the password every time for simple status calls.

Now only the state modifying endpoints and endpoints that expose sensitive data are authenticated. However, @lukechampine brought up the point that the more barrier there is to making API calls, the less likely people are to enable API authentication. Since API authentication will be default in v1.0 we don't want to encourage people to disable authentication.

Therefore as part of API finalization we should decide what endpoints should require authentication. Endpoints that don't provide financial incentive to attack (or are prohibitively difficult for such an attack to be worthwhile) shouldn't require authentication (by default, at least) in order to make the API as usable as possible.

DavidVorick:
After some thought, I don't think we need to password protect the miner endpoints

DavidVorick:
We want to make things as simple as possible for users. I like the idea of "unsecured", "reasonably secure", and "maximally secure" as three options that we present, where unsecured has all security options disabled, 'reasonably secure' is a very convenience oriented setup that only protects things which are highly vulnerable, and in ways that are minimally intrusive, and then 'maximally secure' which gives you the best coverage possible at moderate (but not heavy or extreme) cost to convenience.

PR #1103, opened 2016-04-26
Tpool Fast Start

DavidVorick:
To support the transaction pool's lack of need to actually be aware of any consensus history, I have changed the ConsensusSetSubscribe function to take a few special cases, one of which won't give you any consensus changes, but subscribe you to the beginning. Meaning you may have missed a bunch of updates, but you may not care (as is the case with the transaction pool)

The diff is pretty big because I touched up a few other things as well.

### DIFF
--- a/modules/consensus.go
+++ b/modules/consensus.go
@@ -21,6 +21,16 @@ const (
 )
 
 var (
+	// ConsensusChangeBeginning is a special consensus change id that tells the
+	// consensus set to provide all consensus changes starting from the very
+	// first diff, which includes the genesis block diff.
+	ConsensusChangeBeginning = ConsensusChangeID{}
+
+	// ConsensusChangeRecent is a special consensus change id that tells the
+	// consensus set to provide the most recent consensus change, instead of
+	// starting from a specific value (which many not be known to the caller).
+	ConsensusChangeRecent = ConsensusChangeID{1}
+
 	// ErrBlockKnown is an error indicating that a block is already in the
 	// database.
 	ErrBlockKnown = errors.New("block already present in database")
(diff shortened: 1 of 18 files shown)


modules/consensus/subscribe.go
```diff
@@ -97,25 +97,31 @@ func (cs *ConsensusSet) readlockUpdateSubscribers(ce changeEntry) {
...
-		if start == (modules.ConsensusChangeID{}) {
-			// Special case: if 'start' is blank, create an initial node
-			// pointing to the genesis block. The subscriber will recieve the
-			// diffs for all blocks in the consensus set, including the genesis
-			// block.
+		if start == modules.ConsensusChangeBeginning {
+			// Special case: for modules.ConsensusChangeBeginning, create an
+			// initial node pointing to the genesis block. The subscriber will
+			// recieve the diffs for all blocks in the consensus set, including
+			// the genesis block.
```
lukechampine: I prefer `switch` in these scenarios

modules/consensus/subscribe.go
```diff
@@ -97,25 +97,31 @@ func (cs *ConsensusSet) readlockUpdateSubscribers(ce changeEntry) {
...
 			entry = cs.genesisEntry()
 			exists = true
+		} else if start == modules.ConsensusChangeRecent {
```
lukechampine: I prefer `switch` in these scenarios

modules/consensus.go
```diff
@@ -21,6 +21,16 @@ const (
...
 var (
+	// ConsensusChangeBeginning is a special consensus change id that tells the
+	// consensus set to provide all consensus changes starting from the very
+	// first diff, which includes the genesis block diff.
+	ConsensusChangeBeginning = ConsensusChangeID{}
+
+	// ConsensusChangeRecent is a special consensus change id that tells the
+	// consensus set to provide the most recent consensus change, instead of
+	// starting from a specific value (which many not be known to the caller).
+	ConsensusChangeRecent = ConsensusChangeID{1}
```
lukechampine: maybe `First` and `Last`?

PR #1104, opened 2016-04-26
Fix host config command

lukechampine:
As discussed, we should think about how to make detailed information about each setting readily available to the user.

Proposals:
- man page
- doc/API.md
- `siac define` command


VoidingWarranties:
I think it is appropriate to have the long text of a command be ~long~ and very descriptive.

DavidVorick:
Also, I think we need to make sure that any design and configurability be simple enough that users are not overwhelmed. If you need 350 words to say how something works, it's probably not designed well enough.

This does apply to a few of the host settings. I'll work on simplifying them.

PR #1106, opened 2016-04-26
hostdb and contractor now save consensus updates correctly

DavidVorick:
Prior to this PR, the hostdb never saved anywhere at all, so all the persistence code was actually completely untouched in production.

The contractor would not save when it got new consensus updates, which means that if you never used it you would have to rescan the blockchain at startup every time.

Now fixed :)

As of this PR, starting siad on my machine takes 0.2 seconds, which is... amazingly refreshing. I believe everyone will really appreciate having responsive startup.

### DIFF
--- a/modules/renter/hostdb/scan.go
+++ b/modules/renter/hostdb/scan.go
@@ -77,7 +77,7 @@ func (hdb *HostDB) decrementReliability(addr modules.NetAddress, penalty types.C
 	}
 }
 
-// threadedProbeHost tries to fetch the settings of a host. If successful, the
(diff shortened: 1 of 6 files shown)


modules/renter/hostdb/scan.go
```diff
@@ -175,6 +176,7 @@ func (hdb *HostDB) threadedScan() {
 			for _, randIndex := range hostOrder {
 				hdb.scanHostEntry(entries[randIndex])
 			}
+			hdb.save()
```
lukechampine: I don't think this `save` accomplishes anything, since the scans run concurrently.

PR #1113, opened 2016-04-27
Host siac fixes

lukechampine:


siac/parse.go
```diff
@@ -47,8 +60,37 @@ func parseSize(strSize string) (string, error) {
...
+
+	// iterate until we find a unit greater than c
+	mag := pico
+	unit := ""
+	for _, unit = range []string{"pS", "nS", "uS", "mS", "SC", "KS", "MS", "GS", "TS"} {
+		if c.Cmp(mag.Mul(types.NewCurrency64(1e3))) < 0 {
+			break
+		} else if unit != "TS" {
+			// don't want to perform this multiply on the last iter; that
+			// would give us 1.2345 TB instead of 1234.5 TB
```
DavidVorick: TB -> TS

siac/hostcmd.go
```diff
@@ -160,7 +160,8 @@ func hostcmd() {
 	Revenue:             %v
 	Lost Revenue:        %v
 `, filesizeUnits(int64(totalstorage)), filesizeUnits(int64(totalstorage-storageremaining)),
-		price, is.MaxDuration, accept, totalPotentialRevenue, totalRevenue, fm.LostRevenue)
+		price, is.MaxDuration, accept, currencyUnits(totalPotentialRevenue),
+		currencyUnits(totalRevenue), currencyUnits(fm.LostRevenue))
```
DavidVorick: There's a field missing here which is lost collateral. Lost revenue is stuff that the renter would have paid you but didn't, meaning no real harm done just disappointing, and lost collateral is money you outright threw into the void.

siac/hostcmd.go
```diff
@@ -175,7 +176,7 @@ RPC Stats:
 	Revise Calls:       %v
 	Settings Calls:     %v
 	FormContract Calls: %v
-`, is.NetAddress, nm.ErrorCalls, nm.UnrecognizedCalls, nm.DownloadCalls,
+`, nm.NetAddress, nm.ErrorCalls, nm.UnrecognizedCalls, nm.DownloadCalls,
```
DavidVorick: too simple, you want the user to be able to know whether they are seeing the auto address or the net address... I think

Perhaps it's enough to display a boolean indicating whether the field has been set manually by the user. `Address: Automatic` vs `Address: Manual`, or maybe `Net Address: 12.34.56.78 (automatic)`.

Tricky to do this in a way that easily conveys what happens if they set it manually vs. set it automatically. Maybe it should be labeled 'static' vs 'dynamic', where 'dynamic' is the auto address, and 'static' is manually configured address.

siac/parse.go
```diff
@@ -47,8 +60,37 @@ func parseSize(strSize string) (string, error) {
...
+			mag = mag.Mul(types.NewCurrency64(1e3))
+		}
+	}
+
+	num := new(big.Rat).SetInt(c.Big())
+	denom := new(big.Rat).SetInt(mag.Big())
+	res, _ := new(big.Rat).Mul(num, denom.Inv(denom)).Float64()
+
+	return fmt.Sprintf("%.5g %s", res, unit)
+}
```
DavidVorick: Might also want some sane rounding in these functions. Not sure if the rounding is happening now, but previously 19999999999999 -> 19999 SC, where really you should be displaying 20000

siac/parse_test.go
```diff
@@ -33,9 +35,40 @@ func TestParseSize(t *testing.T) {
...
+	tests := []struct {
+		in, out string
+	}{
+		{"1", "1 H"},
+		{"1000", "1000 H"},
+		{"100000000000", "100000000000 H"},
+		{"1000000000000", "1 pS"},
+		{"1234560000000", "1.2346 pS"},
+		{"12345600000000", "12.346 pS"},
+		{"123456000000000", "123.46 pS"},
```
DavidVorick: Is 5 too many? Maybe it would be better to have 3 or 4

siac/parse.go
DavidVorick: oh I see you've got a test for it

siac/hostcmd.go
lukechampine: how can I tell if the address is auto or not? Compare to internalsettings?

siac/parse_test.go
lukechampine: 4 is probably enough. 3 doesn't seem sufficient because if you have a 3-digit number you have no decimal at all.

siac/hostcmd.go
DavidVorick: if the internal settings is `""`, then it's auto, otherwise it's manual.

PR #1117, opened 2016-04-28
simple hostdb fix

DavidVorick:
everything I said out-of-band today proposing why the persistence was failing was wrong. I was too quick to make assumptions.

All that was really happening was that pointers were being grabbed from a `for _, object :=` loop, which means that every object actually had the same address, which means the final object overwrote all of the others.

This bug has cropped up a few times, so this is a friendly reminder not to use pointers when doing a `for _, object := ` loop, and if you need pointer then you need to index it via a `for i := range objects` loop


lukechampine:
weird that `TestSaveLoad` didn't catch this

PR #1120, opened 2016-04-29
Upload fixes

lukechampine:
2721300 was the worst offender here, but e19e23b also contributed.

### DIFF
--- a/modules/renter/contractor/formcontract.go
+++ b/modules/renter/contractor/formcontract.go
@@ -363,6 +363,7 @@ func (c *Contractor) formContracts(a modules.Allowance) error {
 			// negotiate with because they'll probably be more expensive than
 			// we can afford.
 			c.log.Println("WARN: failed to negotiate contract:", h.NetAddress, err)
+			continue
 		}
 		if numContracts++; numContracts >= a.Hosts {
 			break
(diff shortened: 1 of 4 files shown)


modules/renter/contractor/formcontract.go
```diff
@@ -363,6 +367,7 @@ func (c *Contractor) formContracts(a modules.Allowance) error {
 			// negotiate with because they'll probably be more expensive than
 			// we can afford.
 			c.log.Println("WARN: failed to negotiate contract:", h.NetAddress, err)
```
DavidVorick: Debugln, it's something that will probably happen a lot and is only interesting to developers.

DavidVorick: A renter in the wild was reported to have a log of 1.5GB, too much for non-dev machines. Target should be less than 1mb per day for all of `siad` on the logs, ideally more like 10k per day max for all of `siad`.

Dev machines should try to log as little as possible, but up to 1GB per day for all of `siad` is okay. If the information can help us improve the code, the disk cost is not really a concern.

DavidVorick:
excited to be testing these fixes, I think this will be enough to release rc2

modules/renter/contractor/formcontract.go
lukechampine: On the contrary, `formContract` should be called only when setting the allowance, and maybe when renewing.
The reason the renter log gets so big is because it will try to repair something, fails, logs the error, and then try again 5 seconds later. That's definitely bad.

modules/renter/contractor/formcontract.go
```diff
@@ -363,6 +367,7 @@ func (c *Contractor) formContracts(a modules.Allowance) error {
...
+			continue
 		}
 		if numContracts++; numContracts >= a.Hosts {
 			break
```
VoidingWarranties: Don't we have insufficient hosts if `numContracts < a.Hosts` when the for loop exits? Should we return an error in that case?

PR #1124, opened 2016-04-29
Renter metrics

lukechampine:
Adds spending metrics to the renter, including storage, upload, download, and total paid into contracts. (From these, you can calculate how much will be returned to the renter when its contracts end.)

For now, only historical totals are given. So you can't view how much you spent within, e.g. the last renewal period. I don't think contracts are ever cleared from the contractor (even after they expire). We should address these issues when we implement renewal.

I don't like threading calls through the renter to the contractor. I'm eager to make the contractor a "first-class citizen" with its own API routes. This would allow you to easily inspect individual contracts. It's also a necessity for low-level renter integrations.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -98,10 +98,11 @@ func (srv *Server) initAPI() {
 
 	// Renter API Calls
 	if srv.renter != nil {
-		router.GET("/renter/downloads", srv.renterDownloadsHandler)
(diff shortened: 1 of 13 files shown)


DavidVorick:
nice and simple.

Looking forward to contractor as a first class citizen as well, but might make sense to push that off until after 1.0.

PR #1125, opened 2016-04-30
v0.6.0-rc2

lukechampine:


DavidVorick:
Definitely there should be a flag to prevent the host from setting announce to true, so people don't need to run two commands. No rush.

PR #1126, opened 2016-04-30
fix rc recommendedHosts

lukechampine:


DavidVorick:
only 5 hosts online T.T


just need to get one more, this is fine.

PR #1129, opened 2016-05-02
siac host units

lukechampine:


siac/hostcmd.go
```diff
@@ -141,11 +141,8 @@ func hostcmd() {
 
 	// convert accepting bool
 	accept := yesNo(is.AcceptingContracts)
-	// convert price to SC/TB/mo
-	price, err := modules.StoragePriceToHuman(is.MinimumStoragePrice)
-	if err != nil {
-		price = ^uint64(0)
-	}
+	// convert price from bytes/block to TB/Month
+	price := currencyUnits(is.MinimumStoragePrice.Mul(types.NewCurrency64(4320e12)))
```
VoidingWarranties: Can you write a comment explaining 4320e12?

DavidVorick: Even better, can you use the function in the modules package that has thorough testing? It's there so that we don't get mistakes like these where people omit units or make the calculation out of order.

DavidVorick: why is `StoragePriceToHuman` failing? That means that somewhere something else has bad units, changing it here is not the right action

DavidVorick: (I am poking around, seems like I problem with the host)

siac/hostdbcmd.go
```diff
@@ -30,6 +30,7 @@ func hostdbcmd() {
 	}
 	fmt.Println("Active hosts:")
 	for _, host := range info.Hosts {
-		fmt.Printf("\t%v - %v SC / GB / Mo\n", host.NetAddress, host.StoragePrice.Mul(types.NewCurrency64(4320e9)).Div(types.SiacoinPrecision))
+		price := host.StoragePrice.Mul(types.NewCurrency64(4320e12))
+		fmt.Printf("\t%v - %v / TB / Month\n", host.NetAddress, currencyUnits(price))
```
DavidVorick: All these conversions should be using the `modules` defined functions, if they aren't correct it means that there's a problem somewhere else in the code.

DavidVorick:
This PR is suddenly a lot bigger :P

modules/host.go
```diff
@@ -9,6 +9,14 @@ const (
 	HostDir = "host"
 )
 
+var (
+	// BytesPerTerabyte is conversion rate between bytes and terabytes.
+	BytesPerTerabyte = types.NewCurrency64(1e12)
+
+	// BlockBytesPerMonthTerabyte is the conversion rate between block-bytes and month-TB.
+	BlockBytesPerMonthTerabyte = BytesPerTerabyte.Mul64(4320)
```
DavidVorick: hmm. I guess it makes sense.

modules/host.go
```diff
@@ -111,61 +119,3 @@ type (
...
-		// therefore 0 should be returned.
-		return 0, nil
-	}
-	if hastingsMonthTB.Cmp(types.SiacoinPrecision) < 0 {
-		// The result of the final division is going to be greater than or
-		// equal to 0.5, but less than 1, therefore 1 should be returned.
-		return 1, nil
-	}
-	return hastingsMonthTB.Div(types.SiacoinPrecision).Uint64()
-}
```
DavidVorick: oh.

PR #1134, opened 2016-05-10
Renter usability

lukechampine:
mostly better error reporting.

### DIFF
--- a/modules/renter/contractor/downloader.go
+++ b/modules/renter/contractor/downloader.go
@@ -129,7 +129,9 @@ func (c *Contractor) Downloader(contract Contract) (Downloader, error) {
 	if !ok {
 		return nil, errors.New("no record of that host")
 	}
-	if host.StoragePrice.Cmp(maxPrice) > 0 {
+	// download price should be no more than 3x the cost to store the same
+	// amount of data for a month
+	if host.DownloadBandwidthPrice.Cmp(host.StoragePrice.Mul64(3*4320)) > 0 {
 		return nil, errTooExpensive
 	}
 
--- a/modules/renter/contractor/downloader.go
+++ b/modules/renter/contractor/downloader.go
@@ -129,9 +129,7 @@ func (c *Contractor) Downloader(contract Contract) (Downloader, error) {
 	if !ok {
 		return nil, errors.New("no record of that host")
 	}
-	// download price should be no more than 3x the cost to store the same
-	// amount of data for a month
-	if host.DownloadBandwidthPrice.Cmp(host.StoragePrice.Mul64(3*4320)) > 0 {
+	if host.DownloadBandwidthPrice.Cmp(maxDownloadPrice) > 0 {
 		return nil, errTooExpensive
 	}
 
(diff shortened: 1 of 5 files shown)


modules/renter/contractor/formcontract.go
```diff
@@ -354,20 +356,27 @@ func (c *Contractor) formContracts(a modules.Allowance) error {
...
 		}
 		if numContracts++; numContracts >= a.Hosts {
 			break
 		}
 	}
+	// If we couldn't form any contracts, return an error. Otherwise, just log
+	// the failures.
+	// TODO: is there a better way to handle failure here? Should we prefer an
+	// all-or-nothing approach? We can't pick new hosts to negotiate with
+	// because they'll probably be more expensive than we can afford.
```
DavidVorick: I am in favor of an all-or-nothing approach. Especially because we want to have some buffer room for repairs.

modules/renter/contractor/downloader.go
```diff
@@ -129,7 +129,9 @@ func (c *Contractor) Downloader(contract Contract) (Downloader, error) {
 	if !ok {
 		return nil, errors.New("no record of that host")
 	}
-	if host.StoragePrice.Cmp(maxPrice) > 0 {
```
DavidVorick: You still need to check that the storage price is small enough

modules/renter/contractor/downloader.go
```diff
@@ -129,7 +129,9 @@ func (c *Contractor) Downloader(contract Contract) (Downloader, error) {
...
+	// download price should be no more than 3x the cost to store the same
+	// amount of data for a month
+	if host.DownloadBandwidthPrice.Cmp(host.StoragePrice.Mul64(3*4320)) > 0 {
```
DavidVorick: max should be 3x the max storage price, not 3x the host's storage price

modules/renter/contractor/downloader.go
lukechampine: when you're downloading? why?

modules/renter/contractor/formcontract.go
DavidVorick: oh, but on the other hand the contracts are going to go through regardless, because you made them before knowing that the others would fail.

DavidVorick: Uploads should probably be blocked until there are enough contracts to get to a safe redundancy.

modules/renter/contractor/downloader.go
DavidVorick: oh did not realize that this is just dowloading.

modules/renter/contractor/formcontract.go
lukechampine: it's difficult, but it could be possible: you'd negotiate all your contracts in parallel, but stop right before the end of the protocol. Then when all of them are ready to be finalized, you complete the protocol with all hosts simultaneously. There's still a possibility of failure, but it's smaller. Would be a pain to code though.

DavidVorick: sounds error prone, probably better to have a handful of incomplete contracts. Most likely a few will fail, but if the hostdb is doing its job well you shouldn't see a catastrophic amount of failures.

Issue #1135, opened 2016-05-10
Host stale obligation critical error


lukechampine:
The linode host is reporting this error:
```
CRITICAL: revision submission window was not verified before trying to modify a storage obligation
```

The relevant code:
```go
func (h *Host) modifyStorageObligation(so *storageObligation, sectorsRemoved []crypto.Hash, sectorsGained []crypto.Hash, gainedSectorData [][]byte) error {
    // Sanity check - obligation should be under lock while being modified.
    soid := so.id()
    _, exists := h.lockedStorageObligations[soid]
    if !exists {
        h.log.Critical("modifyStorageObligation called with an obligation that is not locked")
    }
    // Sanity check - there needs to be enough time to submit the file contract
    // revision to the blockchain.
    if so.expiration()-revisionSubmissionBuffer <= h.blockHeight {
        h.log.Critical("revision submission window was not verified before trying to modify a storage obligation")
        return errNoBuffer
    }
```

DavidVorick:
heh, this check does not appear in either place that it's supposed to.

Issue #1086, opened 2016-04-22
Host adds duplicate storage folders


lukechampine:
Should return an error

PR #1136, opened 2016-05-10
Duplicate storage folders

DavidVorick:
Fix two issues with the host, #1086 and #1135

### DIFF
--- a/modules/host/negotiaterevisecontract.go
+++ b/modules/host/negotiaterevisecontract.go
@@ -29,6 +29,12 @@ var (
 	// data which creates a sector that is larger than what the host uses.
 	errLargeSector = errors.New("renter has sent a sector that exceeds the host's sector size")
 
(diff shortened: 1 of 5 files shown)


modules/host/negotiaterevisecontract.go
```diff
@@ -29,6 +29,12 @@ var (
 	// data which creates a sector that is larger than what the host uses.
 	errLargeSector = errors.New("renter has sent a sector that exceeds the host's sector size")
 
+	// errLateRevision is returned if the renter is attempting to revise a
+	// revision after the revision deadline. The host needs time to submit the
+	// final revision to the blockchain to guarantee payment, and therefore
+	// will not accept revisions once the window start is too close.
+	errLateRevision = errors.New("renter is attempting to revise a revision which the host has closed")
```
lukechampine: use of "closed" is slightly confusing, but nbd

PR #1139, opened 2016-05-10
switch storage folder to query param

lukechampine:
this was necessary because Windows-style absolute paths don't work well as URL paths.

absolute paths on Windows don't have a leading slash. So we either get:
```
/storage/folders/addC:/foo/bar
```
which is obviously invalid, or we manually add the slash:
```
/storage/folders/add/C:/foo/bar
```
which works, but now the API sees the path as /C:/foo/bar, which isn't valid. Using a query param seems like the simplest option.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -79,9 +79,9 @@ func (srv *Server) initAPI() {
 
 		// Calls pertaining to the storage manager that the host uses.
 		router.GET("/storage", srv.storageHandler)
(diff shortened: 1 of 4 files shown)


DavidVorick:
weird that this did not break more tests.

lukechampine:
it only affects the API, and the API has a helper function for that specific call. Everyone else calls the method directly.

PR #1141, opened 2016-05-12
Separate protocol code from Contractor

lukechampine:
The renter-host protocol is now fully encapsulated in `renter/contractor/proto`. The basic idea is that `proto` is a lightweight interface to the renter-host protocol; it should do minimal sanity checking, should be stateless (no logging/persistence), and generally assume that the caller knows what they're doing. This is the first step towards a major separation of concerns in the renter module. The various concerns are as follows:

Protocol tier:
- Speak the renter-host protocol with hosts over a network connection
- Provide simple abstractions that facilitate creating, revising, and renewing file contracts, as well as downloading stored data

Relationship tier:
- Manage the user's allowance, enforcing spending limits and forming contracts of appropriate size and duration
- Intelligently select hosts from the host database (Long-term, favor hosts who are faster, high collateral, etc.)
- Allow the user to whitelist or blacklist hosts
- Automatically renew file contracts before they expire
- Manage a collection of potentially overlapping file contracts, distributing data amongst them

File tier:
- Provide a file abstraction for securely and redundantly uploading a file
- Encrypt files during upload and decrypt them during download
- Use erasure coding to store files across multiple hosts
- Monitor file health and reupload file data when necessary

Browser tier:
- Provide a file system abstraction for managing files in a "Sia virtual drive"
- Facilitate sharing and loading files via the .sia format

As of this PR, the protocol tier is handled by `renter/contractor/proto`, the relationship tier is handled by `/renter/contractor`, and the file+browser tier are handled by `renter`. The browser tier is relatively simple, so it may not need to be split from the file tier anytime soon.

I think this new organization will make it easier to incrementally improve the renter. It will also be easier for 3rd parties to hook into the renter at their desired level of abstraction. For example, someone who wants direct control over their redundancy could replace the file tier with their own implementation.

The downside to this approach is that it requires careful thought about how this tiered hierarchy should be expressed in the code base, especially with respect to the `Renter` interface and the API. Should all of the tiers' methods be crammed into one giant interface, or should each tier have its own interface? (Do all those interfaces need to live in `modules`, or can they live under `renter`?) What should be hard-coded, and what should be user-configurable? How will the various tiers be initialized? How should subscription be handled? etc.

So far, I've come up with two approaches. One is to replace the `Renter` interface with a struct, containing one field per tier. The API would then access the various tiers with e.g. `srv.renter.Browser.Rename`. Equivalently, `Renter` could remain an interface, with one method per tier.
The other approach is for the `Renter` interface to only contain the Browser tier methods, plus a method that provides access to the File tier object. The File tier object would in turn provide access to the Relationship object, and from there the Protocol tier.

My preference is for the first approach, but I think there is more room for improvement. The fundamental problem is that various users will want to interact with the renter on various levels of abstraction. Granted, we aren't obligated to make things super friendly for those users; if they have to import the package directly instead of talking to an HTTP API, that's not the end of the world.

This code is definitely a work-in-progress in some areas. Perhaps it should be merged into a feature branch.


modules/renter/repair.go
```diff
@@ -87,9 +86,19 @@ func (f *file) repair(chunkIndex uint64, missingPieces []uint64, r io.ReaderAt,
...
-	wg.Wait()
+	var errs []error
+	for i := 0; i < numPieces; i++ {
+		err := <-errChan
+		if err != nil {
+			errs = append(errs, err)
+		}
+	}
+	if len(errs) == numPieces {
+		return fmt.Errorf("could not upload any sectors:\n%v", build.JoinErrors(errs, "\n"))
```
DavidVorick: Even if not everyone had an error, it would be good to `log.Debugln` all of the other errors. For renters that are offline, this may create massive debug log files, but that's okay.

DavidVorick: After we add some sort of exponential backoff to the renter for trying hosts which have been offline previously, it should improve.

lukechampine: I'm thinking `proto` should have its own error type:
```go
type Error struct  {
    error
    transient bool
}

func IsTransient(err error) {
    e, ok := err.(*Error)
    return ok && e.transient
}
```
Then the renter can explicitly check for transient errors, and it will know not to retry uploading after a non-transient error (e.g. `contract has ended`). Transient errors should be handled with exponential backoff, as you suggested.
I suppose the utility of this depends on how many non-transient errors there are. I think it's worth it if there's more than two.

DavidVorick: I think it needs more thought. Seems kind of bolted-on as presented above.

modules/renter/contractor/contractor.go
```diff
@@ -6,36 +6,22 @@ import (
 	"path/filepath"
 	"sync"
 
-	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/modules"
+	"github.com/NebulousLabs/Sia/modules/renter/contractor/proto"
```
DavidVorick: too much nesting. Fine for now but we should figure out a better way to structure it. Probably file to have it all in /renter

modules/renter/contractor/editor.go
```diff
@@ -84,216 +59,68 @@ func (he *hostEditor) EndHeight() types.BlockHeight { return he.contract.FileCon
...
-		SectorIndex: uint64(index),
-		Offset:      offset,
-		Data:        newData,
-	}}
-	rev := newModifyRevision(he.contract.LastRevision, merkleRoot, sectorPrice)
-
-	// run the revision iteration
-	if err := he.runRevisionIteration(actions, rev, newRoots, height); err != nil {
+	oldUploadSpending := he.editor.UploadSpending
+	contract, err := he.editor.Modify(oldRoot, newRoot, offset, newData)
```
DavidVorick: Might want to name `Modify` to `Revise` to keep parity with the rest of the codebase.

lukechampine: `Modify` is the specific revision action type that overwrites sector data

modules/renter/contractor/contractor.go
lukechampine: you mean `/renter/proto` instead of `/renter/contractor/proto`? That seems sensible.

modules/renter/contractor/formcontract.go
```diff
@@ -3,311 +3,59 @@ package contractor
...
 	// reject hosts that are too expensive
 	if host.StoragePrice.Cmp(maxStoragePrice) > 0 {
-		return Contract{}, errTooExpensive
+		return proto.Contract{}, errTooExpensive
 	}
 
-	// get an address to use for negotiation
+	// TODO: move this outside this function?
 	c.mu.Lock()
+	// get an address to use for negotiation
```
DavidVorick: Either you need to do the locked outside of `newContract`, or you need to name the function `managedNewContract`

modules/renter/contractor/formcontract.go
```diff
@@ -382,7 +128,7 @@ func (c *Contractor) formContracts(a modules.Allowance) error {
 		c.log.Printf("WARN: failed to form desired number of contracts (wanted %v, got %v): %v", a.Hosts, numContracts, strings.Join(errs, "\n"))
 	}
 	c.mu.Lock()
-	c.renewHeight = endHeight
+	c.renewHeight = c.blockHeight + a.Period // TODO: this risks desync
```
DavidVorick: What needs to happen so that desync is no longer risked? I feel like this is a pretty important problem.

modules/renter/contractor/formcontract.go
lukechampine: come to think of it, `cachedAddress` may no longer be necessary, not that we aren't creating a new contract for each file. Worst case, you'll get a handful of wasted addresses when you try to set an allowance.

modules/renter/contractor/formcontract.go
lukechampine: `desync` here means the contracts could end at different heights than `c.renewHeight`. I'm not sure why I changed this. The old method of passing `endHeight` to `newContract` seems fine.

modules/renter/contractor/formcontract.go
DavidVorick: Assuming the user is changing the allowance once per month, the renter ideally uses less than 1000 addresses per year. 2500 is not so bad either, but over 10,000 starts to become problematic.

Renewing 50 contracts every 6 weeks comes to around 500 addresses per year. So 1-5 wasted per contract when setting the allowance is acceptable.

PR #1143, opened 2016-05-12
ignore legacy settings calls

DavidVorick:


modules/negotiate.go
```diff
@@ -156,6 +156,9 @@ var (
 	// RPCSettings is the specifier for requesting settings from the host.
 	RPCSettings = types.Specifier{'S', 'e', 't', 't', 'i', 'n', 'g', 's', 2}
 
+	// RPCSettingsDeprecated is a specifier for a deprecated settings request.
+	RPCSettingsDeprecated = types.Specifier{'S', 'e', 't', 't', 'i', 'n', 'g', 's'}
```
lukechampine: since it's deprecated, could probably be defined in `host` rather than `modules`. idk

DavidVorick: I debated it. For now it's okay in `modules` I think. Renter may care for some reason. Maybe not.

DavidVorick:
updated

PR #1144, opened 2016-05-12
Remove the gateway's RelayNode RPC

VoidingWarranties:
The RelayNode RPC allowed any node to add an arbitrary address to the node list of any other node. Even worse, nodes would re-broadcast this new node and so it would quickly propagate across the network. This could be abused by an attacker to maliciously pollute the node lists of all nodes on the network (an eclipse attack).

Now nodes share their port on connect so that the receiving node can add the connecting node to its node list and connect back to them later.

A notable feature necessitated by this change is that inbound connections now store the peer's dialback address instead of the actual `conn.RemoteAddr()`, and therefore `siac gateway list` will show the dialable address of inbound peers.

Also some miscellaneous cleanup.

### DIFF
--- a/modules/gateway/gateway.go
+++ b/modules/gateway/gateway.go
@@ -119,7 +119,6 @@ func New(addr string, persistDir string) (g *Gateway, err error) {
 
 	// Register RPCs.
 	g.RegisterRPC("ShareNodes", g.shareNodes)
-	g.RegisterRPC("RelayNode", g.relayNode)
 	g.RegisterConnectCall("ShareNodes", g.requestNodes)
 
 	// Load the old node list. If it doesn't exist, no problem, but if it does,
--- a/modules/gateway/gateway.go
+++ b/modules/gateway/gateway.go
@@ -23,6 +23,7 @@ var (
 type Gateway struct {
 	listener net.Listener
 	myAddr   modules.NetAddress
+	port     string
 
 	// handlers are the RPCs that the Gateway can handle.
 	handlers map[rpcID]modules.RPCFunc
@@ -143,9 +144,9 @@ func New(addr string, persistDir string) (g *Gateway, err error) {
 	if err != nil {
 		return
 	}
-	_, port, portErr := net.SplitHostPort(g.listener.Addr().String())
-	if portErr != nil {
-		return nil, portErr
+	_, g.port, err = net.SplitHostPort(g.listener.Addr().String())
+	if err != nil {
+		return nil, err
 	}
 	if build.Release == "testing" {
 		g.myAddr = modules.NetAddress(g.listener.Addr().String())
@@ -154,10 +155,10 @@ func New(addr string, persistDir string) (g *Gateway, err error) {
 	g.log.Println("INFO: gateway created, started logging")
 
 	// Forward the RPC port, if possible.
-	go g.forwardPort(port)
+	go g.forwardPort(g.port)
 
 	// Learn our external IP.
-	go g.learnHostname(port)
+	go g.learnHostname()
 
 	// Spawn the peer and node managers. These will attempt to keep the peer
 	// and node lists healthy.
--- a/modules/gateway/nodes.go
+++ b/modules/gateway/nodes.go
@@ -12,18 +12,22 @@ import (
 
 const (
 	maxSharedNodes = 10
-	maxAddrLength  = 100
+	maxAddrLength  = 100 // TODO: max NetAddress length is 254 for the hostname (including the trailing dot, 253 without) + 1 for the : + 5 for the port.
 	minPeers       = 3
+)
+
+var (
+	errNodeExists = errors.New("node already added")
 )
 
 // addNode adds an address to the set of nodes on the network.
 func (g *Gateway) addNode(addr modules.NetAddress) error {
 	if _, exists := g.nodes[addr]; exists {
-		return errors.New("node already added")
-	} else if net.ParseIP(addr.Host()) == nil {
-		return errors.New("address is not routable: " + string(addr))
+		return errNodeExists
 	} else if addr.IsValid() != nil {
 		return errors.New("address is not valid: " + string(addr))
+	} else if net.ParseIP(addr.Host()) == nil {
+		return errors.New("address must be an IP address: " + string(addr))
 	}
 	g.nodes[addr] = struct{}{}
 	return nil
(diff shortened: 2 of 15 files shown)


modules/gateway/gateway.go
```diff
@@ -58,7 +59,13 @@ func (g *Gateway) Close() error {
 
 	// save the latest gateway state
 	id := g.mu.RLock()
-	if err := g.saveSync(); err != nil {
+	var err error
+	if build.Release == "testing" {
+		err = g.save()
+	} else {
+		err = g.saveSync()
+	}
```
DavidVorick: not necessary, `saveSync()` already skips the Sync if testing

modules/gateway/nodes.go
```diff
@@ -12,18 +12,25 @@ import (
 
 const (
 	maxSharedNodes = 10
-	maxAddrLength  = 100
+	maxAddrLength  = 100 // TODO: max NetAddress length is 254 for the hostname (including the trailing dot, 253 without) + 1 for the : + 5 for the port.
```
DavidVorick: why didn't you just increase it?

modules/gateway/nodes_test.go
```diff
@@ -222,9 +224,9 @@ func TestRelayNodes(t *testing.T) {
 
 	// g2 should have received g3's address from g1
```
DavidVorick: this comment is incorrect, g3 should have received g2's address from g1. Logic is correct.

modules/gateway/peers.go
```diff
@@ -162,6 +163,40 @@ func (g *Gateway) acceptConn(conn net.Conn) {
...
+		var dialbackPort string
+		err := encoding.ReadObject(conn, &dialbackPort, 13) // Max port # is 65535 (5 digits long) + 8 byte string length prefix
+		if err != nil {
+			conn.Close()
+			g.log.Printf("INFO: could not read remote peer's (%v) port: %v", addr, err)
+			return
+		}
+		if _, err := strconv.Atoi(dialbackPort); err != nil {
+			conn.Close()
+			g.log.Printf("INFO: peer (%v) sent an invalid dialback port: %v", addr, err)
```
DavidVorick: Generally speaking we don't want to `Printf` messages that can be induced by byzantine peers. In this case, the message can be manipulated by the peer being malicious, therefore the logging should be debug only.  Same for some of the other `log.Print` calls in this PR

modules/gateway/peers.go
```diff
@@ -187,17 +222,34 @@ func (g *Gateway) acceptConn(conn net.Conn) {
...
+		p.NetAddress = remoteAddr
+	}
+	g.addPeer(&p)
+	// addNode must be called after addPeer so that threadedPeerManager doesn't
+	// try to Connect to the node before we do (although in this particular case
+	// it doesn't matter as a lock is held over both calls).
+	var err error
+	if build.VersionCmp(remoteVersion, "0.6.1") >= 0 {
+		err = g.addNode(remoteAddr)
+		// TODO: call g.save?
```
DavidVorick: probably a good idea.

DavidVorick:
Ok, I'm going to say what I understand the new protocol does, and you correct me if I'm wrong about anything.

1. When two peers connect, the old peer adds the new peer to their nodelist. The new peer obviously already has the old peer in their nodelist.
2. If you don't have enough peers, you can ask one of your existing peers for a selection from their nodelist. Their nodelist is pruned over time, and therefore is likely to have a high percentage of nodes that are currently online.
3. You don't have multiple nodes from the same hostname / IP address in your nodelist at the same time. This is tricky, because the fact could be disguised by using a bunch of hostnames. The goal is to mitigate Sybil attacks. Not a huge concern for now, but we should start doing bucketing at some point to fight this.

modules/gateway/gateway.go
VoidingWarranties: I don't think it does. It used to, but the recent PR that made syncing optional un-did that.

modules/gateway/peers.go
VoidingWarranties: Hmmm that's an interesting concern. I mean the purpose of logging is to be able to debug when things go wrong, right? If we didn't log anything that byzantine peers could manipulate I feel like we'd never log anything at all.

VoidingWarranties:
First two are correct. We currently allow multiple nodes with the same IP (non-IP hostnames aren't allowed). But like you said we can easily change this in the future.

modules/gateway/gateway.go
DavidVorick: Oh. I still don't think you should be doing this at this level. If you don't think we should be syncing during testing, let's handle it in the persist package and not here.

modules/gateway/peers.go
DavidVorick: To me, purpose of logging is to give the user a clean history of the things that their client has done. Stuff like 'started up, made 5 contracts, renewed on X date for Y coins', etc. Cluttering it with a bunch of debugging statements is counter productive unless you are doing debugging, in which case `Debugln` will give you what you want.

If we let byzantine users control production machine logging, they can fill users disks in a DoS attack. Definitely not a good thing.

modules/gateway/peers.go
```diff
@@ -162,6 +163,40 @@ func (g *Gateway) acceptConn(conn net.Conn) {
...
+			conn.Close()
+			g.log.Printf("INFO: peer's address (%v) is invalid: %v", remoteAddr, err)
+			return
+		}
+		id := g.mu.Lock()
+		_, exists := g.peers[remoteAddr]
+		g.mu.Unlock(id)
+		if exists {
+			conn.Close()
+			g.log.Printf("INFO: already connected to a peer on that address (%v), is the peer lying about their port?", remoteAddr)
```
lukechampine: speculation is not necessary

modules/gateway/peers.go
```diff
@@ -162,6 +163,40 @@ func (g *Gateway) acceptConn(conn net.Conn) {
 		return
 	}
 
+	// Read the peer's port that we can dial them back on. Peers older than v0.6.1
+	// will only be discovered by newer peers via the ShareNodes RPC.
+	var remoteAddr modules.NetAddress
+	if build.VersionCmp(remoteVersion, "0.6.1") >= 0 {
```
lukechampine: this branch is larger enough that it might warrant its own function. e.g.
```go
switch {
case build.VersionCmp(remoteVersion, "0.6.1"):
    return g.addPeer(blah, blah, blah)
default:
    return g.addOldPeer(blah, blah, blah)
}
```
depends on how ugly the solution winds up being

PR #1148, opened 2016-05-13
Clean termination of revision loop

lukechampine:
The host now sends `modules.StopResponse` to the renter when it wishes to gracefully terminate the revision loop. The renter will respect this by disconnecting without error.

I still don't like this. It's bizarre that proper use of the host-renter protocol should require repeatedly disconnecting and reconnecting for no reason. There are many reasons why a host would want to terminate the revision loop: shutdown, changed settings, DoS concerns. But disconnecting for no reason forces the renter to reconnect, and this requires ugly and confusing logic. (The situation is even worse when downloading: due to the way the concerns are separated, there's no easy way to reconnect at all. This means that **files taking longer than 20 minutes to download cannot currently be downloaded**.)
We should ask what we stand to gain from enforcing this hard limit. Assume that the renter will immediately try to reconnect. The gain here is that we have a brief period between disconnect and reconnect, where we can use some criteria to judge whether we should accept the reconnection. But at present (to my knowledge) *no such criteria is implemented*, and if we wanted to implement it, we could do so inside the revision loop, rather than outside.

In short: gracefully terminating with `StopResponse` is fine, but the implication that the renter should immediately reconnect is not.

### DIFF
--- a/modules/host/negotiaterevisecontract.go
+++ b/modules/host/negotiaterevisecontract.go
@@ -106,7 +106,7 @@ var (
 // managedRevisionIteration handles one iteration of the revision loop. As a
 // performance optimization, multiple iterations of revisions are allowed to be
 // made over the same connection.
-func (h *Host) managedRevisionIteration(conn net.Conn, so *storageObligation) error {
+func (h *Host) managedRevisionIteration(conn net.Conn, so *storageObligation, finalIter bool) error {
 	// Send the settings to the renter. The host will keep going even if it is
 	// not accepting contracts, because in this case the contract already
 	// exists.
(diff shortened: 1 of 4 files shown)


DavidVorick:
It's worth noting that it's already the case that files taking longer than 20 minutes to download cannot currently be downloaded. Persistent downloads are going to be required regardless, as large downloads could take multiple days and there are no guarantees that a single connection would remain open long enough to pull in that much data.

I think I agree that there's no reason for the constant connect-disconnect cycle, though at the moment clean shutdown in the host fully depends on it - the resourceLock is grabbed when a network connection is opened and is not released until the network connection is closed. It's things of that sort that make me think it's good to have a hard limit on the bound - there may be other behaviors that we're missing that could go awry if a function is allowed to be alive forever. It's annoying, but I believe it's good security practice. It's difficult to be certain that you've implemented everything correctly, and definitely good to have safety features which can course-correct if something goes wrong. The host is more complex than we are able to guarantee behavior about, and allowing a state-heavy function to exist forever may have consequences that we haven't considered, especially because it was designed thinking that the function would be guaranteed to reset after 30 minutes.

lukechampine:
We should work to rectify this. This would not be acceptable if other clients were talking to the host.

PR #1149, opened 2016-05-14
implement file contract renew host-side

DavidVorick:
This code is a combination of `FormContract` and `ReviseContract` code, though there is some unique stuff with regards to how payouts are managed.

### DIFF
--- a/modules/host/negotiateformcontract.go
+++ b/modules/host/negotiateformcontract.go
@@ -48,7 +48,7 @@ var (
 
 	// errEmptyFileContractTransactionSet is returned if the renter provides a
 	// nil file contract transaction set during file contract negotiation.
-	errEmptyFileContractTransactionSet = errors.New("file contract transaction set is nil - invalid!")
+	errEmptyFileContractTransactionSet = errors.New("file contract transaction set is empty")
 
 	// errLowHostPayout is returned if the host is not getting paid enough in
 	// the file contract to cover the contract price.
@@ -84,8 +84,7 @@ var (
 // contractCollateral returns the amount of collateral that the host is
 // expected to add to the file contract based on the payout of the file
 // contract and based on the host settings.
-func contractCollateral(settings modules.HostInternalSettings, txnSet []types.Transaction) types.Currency {
-	fc := txnSet[len(txnSet)-1].FileContracts[0]
+func contractCollateral(settings modules.HostInternalSettings, fc types.FileContract) types.Currency {
 	return fc.ValidProofOutputs[1].Value.Sub(settings.MinimumContractPrice)
 }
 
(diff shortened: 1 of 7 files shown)


modules/host/negotiateformcontract.go
```diff
@@ -43,6 +46,10 @@ var (
 	// which is longer than the host's maximum duration.
 	errDurationTooLong = errors.New("file contract has a duration which exceeds the duration permitted by the host")
 
+	// errEmptyFileContractTransactionSet is returned if the renter provides a
+	// nil file contract transaction set during file contract negotiation.
+	errEmptyFileContractTransactionSet = errors.New("file contract transaction set is nil - invalid!")
```
lukechampine: `file contract transaction set is nil` is fine. Errors aren't supposed to have punctuation.

DavidVorick:
Does the rest of this look alright?

modules/host/negotiaterenewcontract.go
```diff
@@ -0,0 +1,315 @@
...
+func renewBaseCollateral(so *storageObligation, settings modules.HostExternalSettings, txnSet []types.Transaction) types.Currency {
+	fc := txnSet[len(txnSet)-1].FileContracts[0]
+	timeExtension := fc.WindowEnd - so.proofDeadline()
+	return settings.Collateral.Mul64(fc.FileSize).Mul64(uint64(timeExtension))
+}
+
+// renewBasePrice returns the base cost of the storage in the file contract,
+// using the host external settings and the starting file contract.
+func renewBasePrice(so *storageObligation, settings modules.HostExternalSettings, txnSet []types.Transaction) types.Currency {
+	fc := txnSet[len(txnSet)-1].FileContracts[0]
```
lukechampine: why do these functions take `txnSet` rather than `fc` directly?

modules/host/negotiaterenewcontract.go
```diff
@@ -0,0 +1,315 @@
...
+		return err
+	}
+	err = encoding.ReadObject(conn, &renterPK, modules.NegotiateMaxSiaPubkeySize)
+	if err != nil {
+		return err
+	}
+
+	h.mu.RLock()
+	settings := h.externalSettings()
+	h.mu.RUnlock()
```
lukechampine: what happens if the user changes their settings between `h.managedRPCSettings` and `h.externalSettings()`? I assume the host may reject the renewal, or send terms that the renter will reject. This behavior seems fine to me, just something to be aware of.

modules/host/negotiaterenewcontract.go
```diff
@@ -0,0 +1,315 @@
...
+		return errEmptyFileContractTransactionSet
+	}
+	// Check that the transaction set has a file contract.
+	if len(txnSet[len(txnSet)-1].FileContracts) < 1 {
+		return errNoFileContract
+	}
+
+	h.mu.RLock()
+	blockHeight := h.blockHeight
+	externalSettings := h.externalSettings()
```
lukechampine: since these were read earlier in `managedRPCRenewContract`, maybe pass them as an arg?

lukechampine:
Yeah, seems almost identical to `FormContract`, which is good

modules/host/negotiaterenewcontract.go
DavidVorick: I thought about sending the settings around starting from the beginning of the connection, but this is a rare race condition and it will be handled mostly gracefully. The host will either reject the renter's file contract, or nothing unusual will happen. Host will only reject the renter if the price goes up, nothing happens if the price goes down.

This is something that can reasonably be addressed later. Might be a larger issue once the host is automatically adjusting prices to meet supply + demand.

DavidVorick:
ok, I should be able to get to all of this before bed tonight.

PR #1150, opened 2016-05-17
Host logging

DavidVorick:
I believe I know why the host is not reporting revenue, I believe it's a significant issue (the host is actually not getting revenue, not merely misreporting it), and I believe that fixing it requires changing the transaction pool.

I think I can have it done tonight.

### DIFF
--- a/modules/host/storageobligations.go
+++ b/modules/host/storageobligations.go
@@ -539,6 +539,7 @@ func (h *Host) removeStorageObligation(so *storageObligation, sos storageObligat
 	}
 	if sos == obligationRejected {
 		// Remove the obligation statistics as potential risk and income.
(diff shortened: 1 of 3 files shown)


modules/host/storageobligations.go
```diff
@@ -594,25 +597,31 @@ func (h *Host) handleActionItem(so *storageObligation) {
...
+			// Check if the transaction is invalid with the current consensus set.
+			// If so, the transaction is highly unlikely to ever be confirmed, and
+			// the storage obligation should be removed. This check should come
+			// after logging the errror so that the function can quit.
+			//
+			// TODO: If the host or tpool is behind consensus, might be difficult
+			// to have certainty about the issue. If some but not all of the
+			// parents are confirmed, might be some difficulty.
+			_, t := err.(modules.ConsensusConflict)
+			if t {
```
lukechampine: usually this is written:
```go
if _, ok := err.(modules.ConsensusConflict); ok {
```

PR #1151, opened 2016-05-17
Renter renew protocol

lukechampine:
Also remove MaxCollateralFraction

### DIFF
--- a/modules/renter/contractor/host_integration_test.go
+++ b/modules/renter/contractor/host_integration_test.go
@@ -507,7 +507,7 @@ func TestIntegrationRenew(t *testing.T) {
 		t.Fatal(err)
 	}
 	// insert the sector
-	_, err = editor.Upload(data)
+	root, err := editor.Upload(data)
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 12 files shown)


modules/renter/contractor/host_integration_test.go
```diff
@@ -472,3 +472,54 @@ func TestIntegrationModify(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// renew the contract
+	contract = c.contracts[contract.ID]
+	_, err = c.managedRenew(contract, modules.SectorSize*10, c.blockHeight+200)
+	if err != nil {
+		t.Fatal(err)
+	}
```
DavidVorick: I would like to see this test extended, so that you do some checking on the values of the renewed contract. Might be good to do a download using the renewed contract as well, just to make sure that everything is at least working.

modules/renter/proto/renew.go
```diff
@@ -0,0 +1,247 @@
...
+
+	// add miner fee
+	txnBuilder.AddMinerFee(fee)
+
+	// create initial transaction set
+	txn, parentTxns := txnBuilder.View()
+	txnSet := append(parentTxns, txn)
+
+	// initiate connection
+	conn, err := net.DialTimeout("tcp", string(host.NetAddress), 15*time.Second)
```
DavidVorick: 15 seconds?

lukechampine: The hostdb scan times out in 5 seconds, so it's a moot point. We can extend it across the board, but I think it's out of scope for this PR.

We may want to add special accommodations for Tor users. Long timeouts degrade the average user experience.

modules/renter/contractor/host_integration_test.go
lukechampine: done.

btw, do you think these tests make more sense in `proto`? The contractor hides some irrelevant details like the tpool, txnBuilder, and renterAddress, but at the end of the day, the core functionality being tested lies in the `proto` functions, not the contractor ones.

PR #1152, opened 2016-05-17
Don't sync persist files to disk in testing

VoidingWarranties:
this was originally done in #1050 but was undone in #1100

### DIFF
--- a/persist/log.go
+++ b/persist/log.go
@@ -85,9 +85,13 @@ func (cf *closeableFile) Close() error {
 		build.Critical("cannot close the file; already closed")
 	}
 
-	// Ensure that all data has actually hit the disk.
-	if err := cf.Sync(); err != nil {
-		return err
+	// Ensure that all data has actually hit the disk. Don't sync during testing
+	// as it can take too long and causes tests to fail.
+	// TODO: mock the os package so that Sync() is a no-op during testing.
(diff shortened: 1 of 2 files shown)


DavidVorick:
What is the difference in performance? Now that we are syncing less frequently, I don't think it should make a huge difference in the amount of time a test takes to run.

Especially when running tests on different architectures, I think it's good to keep things as full featured as possible. If there's a problem calling `Sync` on FreeBSD, for example, that's not something we're going to catch until production if our tests don't do any syncing ever. Or there might be other obscure bugs that would show up in testing but only if sync is enabled. This keeps with the general strategy of being as close to the real world as possible.

We don't currently have CI for FreeBSD (or Windows for that matter), but it's definitely something I would like to set up at some point, and I would like the test suite to be as comprehensive as possible.

What are the benchmarks on running the long tests with and without this code? If it's within a factor of 2, I don't think it's a big deal. Our long tests take less than 10 minutes, which is faster than most projects of this size can even compile (yay golang).

DavidVorick:
So it's an issue for the short tests?

VoidingWarranties:
Yes

DavidVorick:
Okay, pending further thought my preference would be to move non-unit tests to be long instead of short. That would fix the problem, and keep the short tests running fast.

PR #1154, opened 2016-05-19
Fix Transaction Pool Bug

DavidVorick:
The transaction pool was previously experiencing a bug where transaction sets with parents that had been accepted to the blockchain would be rejected as double-spends. This was preventing hosts from getting paid. The transaction pool now keeps a record of all the transactions that have been confirmed on the blockchain, and will remove the parents from a transaction set that have been confirmed.

This behavior should be tested in an rc, but I think this PR is the last code change we need before v0.6.0, the rest should just be adjusting the constants.

### DIFF
--- a/modules/gateway/rpc.go
+++ b/modules/gateway/rpc.go
@@ -61,7 +61,7 @@ func (g *Gateway) RPC(addr modules.NetAddress, name string, fn modules.RPCFunc) 
 func (g *Gateway) RegisterRPC(name string, fn modules.RPCFunc) {
 	id := g.mu.Lock()
 	defer g.mu.Unlock(id)
-	if build.DEBUG {
+	if build.DEBUG && build.Release != "testing" {
 		if _, ok := g.handlers[handlerName(name)]; ok {
 			panic("refusing to overwrite RPC " + name)
 		}
--- a/modules/transactionpool/persist.go
+++ b/modules/transactionpool/persist.go
@@ -5,7 +5,6 @@ import (
 	"os"
 	"path/filepath"
 
-	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/persist"
 	"github.com/NebulousLabs/Sia/types"
@@ -60,7 +59,7 @@ func (tp *TransactionPool) initPersist() error {
 	}
 
 	// Create the database and get the most recent consensus change.
-	cc := modules.ConsensusChangeBeginning
+	var cc modules.ConsensusChangeID
 	err = tp.db.Update(func(tx *bolt.Tx) error {
 		// Create the database buckets.
 		buckets := [][]byte{
@@ -126,16 +125,12 @@ func (tp *TransactionPool) transactionConfirmed(tx *bolt.Tx, id types.Transactio
 	if confirmedBytes == nil {
 		return false
 	}
-	if confirmedBytes[0] == 1 {
-		return true
-	}
-	build.Critical("transaction database has an illegal value for a txid")
-	return false
+	return true
 }
 
 // addTransaction adds a transaction to the list of confirmed transactions.
 func (tp *TransactionPool) addTransaction(tx *bolt.Tx, id types.TransactionID) error {
-	return tx.Bucket(bucketConfirmedTransactions).Put(id[:], []byte{1})
+	return tx.Bucket(bucketConfirmedTransactions).Put(id[:], []byte{})
 }
 
 // deleteTransaction deletes a transaction from the list of confirmed
--- a/modules/transactionpool/persist.go
+++ b/modules/transactionpool/persist.go
@@ -1,0 +1,145 @@
+package transactionpool
+
+import (
+	"errors"
+	"os"
+	"path/filepath"
+
+	"github.com/NebulousLabs/Sia/build"
+	"github.com/NebulousLabs/Sia/modules"
+	"github.com/NebulousLabs/Sia/persist"
+	"github.com/NebulousLabs/Sia/types"
+
+	"github.com/NebulousLabs/bolt"
+)
+
+var (
+	// bucketRecentConsensusChange holds the most recent consensus change seen
+	// by the transaction pool.
+	bucketRecentConsensusChange = []byte("RecentConsensusChange")
+
+	// bucketConfirmedTransactions holds the ids of every transaction that has
+	// been confirmed on the blockchain.
+	bucketConfirmedTransactions = []byte("ConfirmedTransactions")
+
+	// errNilConsensusChange is returned if there is no consensus change in the
+	// database.
+	errNilConsensusChange = errors.New("no consensus change found")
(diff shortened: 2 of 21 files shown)


modules/gateway/rpc.go
```diff
@@ -61,7 +61,7 @@ func (g *Gateway) RPC(addr modules.NetAddress, name string, fn modules.RPCFunc)
 func (g *Gateway) RegisterRPC(name string, fn modules.RPCFunc) {
 	id := g.mu.Lock()
 	defer g.mu.Unlock(id)
-	if build.DEBUG {
+	if build.DEBUG && build.Release != "testing" {
```
VoidingWarranties: Why ignore this check in testing?

modules/transactionpool/transactionpool.go
```diff
@@ -80,28 +93,34 @@ func New(cs modules.ConsensusSet, g modules.Gateway) (*TransactionPool, error) {
 		knownObjects:        make(map[ObjectID]TransactionSetID),
 		transactionSets:     make(map[TransactionSetID][]types.Transaction),
 		transactionSetDiffs: make(map[TransactionSetID]modules.ConsensusChange),
+
+		persistDir: persistDir,
 	}
-	// Register RPCs
-	// TODO: rename RelayTransactionSet so that the conflicting RPC
-	// RelayTransaction calls v0.4.6 clients and earlier are ignored.
```
VoidingWarranties: Isn't this TODO still valid?

modules/transactionpool/persist.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	// Open the database file.
+	tp.db, err = persist.OpenDatabase(dbMetadata, filepath.Join(tp.persistDir, dbFilename))
+	if err != nil {
+		return err
+	}
+
+	// Create the database and get the most recent consensus change.
+	cc := modules.ConsensusChangeBeginning
```
VoidingWarranties: Should this be `var cc modules.ConsensusChangeID`?

lukechampine: a common Go idiom is "make the zero value useful," so I would also prefer that style. We do this with currency as well: `var c types.Currency` is 0 coins. However, sometimes it's convenient to have an explicit variable for the zero value, as we do with `types.ZeroCurrency`.

modules/gateway/rpc.go
lukechampine: sometimes it's convenient to overwrite an RPC for mocking purposes. I don't see any tests in this PR that do that, though

DavidVorick: The rescan tests requires me to close and re-open the tpool a few times. The tpool calls RegisterRPC during `New`, causing panics because there's no way to disconnect an RPC from the gateway. I should drop a TODO here before this gets merged.

 Ideally, modules would be able to de-register their RPCs upon closing. Well, doubly important because the object is not going to be GC'd if the gateway is still pointing to it, and something might panic if you close a module but the gateway still has an RPC registered to it.

modules/transactionpool/persist.go
```diff
@@ -0,0 +1,145 @@
...
+	if confirmedBytes[0] == 1 {
+		return true
+	}
+	build.Critical("transaction database has an illegal value for a txid")
+	return false
+}
+
+// addTransaction adds a transaction to the list of confirmed transactions.
+func (tp *TransactionPool) addTransaction(tx *bolt.Tx, id types.TransactionID) error {
+	return tx.Bucket(bucketConfirmedTransactions).Put(id[:], []byte{1})
```
lukechampine: I wish there was a cleaner way of creating sets (in BoltDB and Go).

Technically, you can use an empty non-nil byte slice here (`[]byte{}`). From the Bolt docs:
> It's important to note that you can have a zero-length value set to a key which is different than the key not existing.
Then you can simply return `tx.Bucket(bucketConfirmedTransactions).Get(id[:]) != nil` in `transactionConfirmed`. Checking for illegal values seems unnecessary, but that's just me.

modules/gateway/rpc.go
VoidingWarranties: 👍  for having a de-register RPC method

lukechampine: Good point. Deregistering should be very easy to implement.

modules/transactionpool/persist.go
VoidingWarranties: In this case the `cc` variable is also being overwritten before it is ever read.

modules/gateway/rpc.go
DavidVorick: I made an issue instead of a TODO

modules/transactionpool/transactionpool.go
DavidVorick: technically, but there aren't many old nodes anymore. Ecosystem has done a good job of upgrading.

modules/transactionpool/persist.go
DavidVorick: `ConsensusChangeBeginning` has a meaning which is not necessarily corresponding to anything in particular. Having it be the 0 value is probably an anti-pattern, you can trigger unexpected behavior on accident.

I should map it to a different value.

modules/transactionpool/persist.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	// Subscribe to the consensus set using the most recent consensus change.
+	err = tp.consensusSet.ConsensusSetSubscribe(tp, cc)
+	if err == modules.ErrInvalidConsensusChangeID {
+		// Reset and rescan because the consensus set does not recognize the
+		// provided consensus change id.
+		resetErr := tp.db.Update(func(tx *bolt.Tx) error {
+			return tp.resetDB(tx)
+		})
```
lukechampine: `tp.db.Update(tp.resetDB)`?

lukechampine:
Curious: is there a way to accomplish this without needing to save every txid in the whole blockchain? I hope so.

DavidVorick:
Depends on where you are willing to push the problem. If you want to call `TryTransactionSet` 2^n times, you don't need to save every txid. If you do intelligent txn analysis, you can bring 2^n down to n^2 I believe, but that's still incredibly costly, because each operation is very expensive.

If you want the modules to pay attention to their own transaction dependencies by tracking the blockchain, you don't need to do this at all. But what you've done is increase the responsibility of the modules in a way that requires them to understand how transaction dependencies work. To me that sounds like the responsibility of the transaction pool.

Having this will make other tpool optimizations easier in the future as well.

modules/transactionpool/persist.go
DavidVorick: hmm. That's almost in the realm of 'too clever'. It's just not the way I'm used to seeing bolt used, it would throw me off.

VoidingWarranties:
```
A consensus rescan has been requested. This can take a while.
Rescan has hit height 0 out of 0
```
appears a bunch throughout the test suite. Can it be silenced in testing?

DavidVorick:
sure

Issue #1156, opened 2016-05-19
Gateway, deregister RPCs


DavidVorick:
Modules should de-register their RPCs from the gateway upon shutdown. Right now you could do that by overwriting an RPC with an empty function, but only during testing.

this is part of clean shutdown, and absolutely required if we want to do things like in-place module rebooting (which is something that makes testing a lot easier)

PR #1157, opened 2016-05-19
Unregister RPC and ConnectCall methods

VoidingWarranties:
Fixes issue #1156 brought up in #1154.

This PR adds two new methods to the gateway `UnregisterRPC` and `UnregisterConnectCall`.

Without these methods the gateway will maintain references to RPCFuncs passed into RegisterRPC and RegisterConnectCall forever. This prevented the GC from cleaning up modules that were closed separately from the gateway. These two methods should primarily be used in the `Close()` method of modules.

### DIFF
--- a/modules/gateway/rpc.go
+++ b/modules/gateway/rpc.go
@@ -69,6 +69,19 @@ func (g *Gateway) RegisterRPC(name string, fn modules.RPCFunc) {
 	g.handlers[handlerName(name)] = fn
 }
 
+// UnregisterRPC unregisters an RPC and removes the corresponding RPCFunc from
+// g.handlers. Future calls to the RPC by peers will fail.
+func (g *Gateway) UnregisterRPC(name string) {
+	id := g.mu.Lock()
+	defer g.mu.Unlock(id)
+	if build.DEBUG {
+		if _, ok := g.handlers[handlerName(name)]; !ok {
+			panic("RPC is not registered: " + name)
+		}
+	}
+	delete(g.handlers, handlerName(name))
+}
+
 // RegisterConnectCall registers a name and RPCFunc to be called on a peer
 // upon connecting.
 func (g *Gateway) RegisterConnectCall(name string, fn modules.RPCFunc) {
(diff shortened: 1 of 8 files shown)


modules/gateway/rpc.go
```diff
@@ -69,6 +69,19 @@ func (g *Gateway) RegisterRPC(name string, fn modules.RPCFunc) {
...
+// UnregisterRPC unregisters an RPC and removes the corresponding RPCFunc from
+// g.handlers. Future calls to the RPC by peers will fail.
+func (g *Gateway) UnregisterRPC(name string) {
+	id := g.mu.Lock()
+	defer g.mu.Unlock(id)
+	if build.DEBUG {
+		if _, ok := g.handlers[handlerName(name)]; !ok {
+			panic("RPC is not registered: " + name)
+		}
+	}
```
VoidingWarranties: The intention of panicing here was to catch typos, but panicing is dangerous as an RPC may be registered only under certain conditions. In such a case there is no way to know through gateway's exported methods whether an RPC has been registered or not.

lukechampine:
be sure to revert the `&& build.Release != "testing"` that David added in his last PR

modules/gateway/rpc.go
DavidVorick: `build.Critical` -> panics `if DEBUG` otherwise prints a strong message to the terminal.

modules/gateway/rpc.go
```diff
@@ -82,6 +95,20 @@ func (g *Gateway) RegisterConnectCall(name string, fn modules.RPCFunc) {
...
 
+// UnregisterConnectCall unregisters an on-connect call and removes the
+// corresponding RPCFunc from g.initRPCs. Future connections to peers will not
+// trigger the RPC to be called on them.
+func (g *Gateway) UnregisterConnectCall(name string) {
+	id := g.mu.Lock()
+	defer g.mu.Unlock(id)
+	if build.DEBUG {
+		if _, ok := g.initRPCs[name]; !ok {
+			panic("ConnectCall is not registered: " + name)
```
DavidVorick: ditto for using `build.Critical`

PR #1158, opened 2016-05-19
Add /renter/contracts route

lukechampine:
To enable this, the `proto.Contract` type had to be moved to `modules`. This is an interesting side effect of requiring module interfaces. Since `proto` imports `modules`, the `Renter` interface cannot return a `proto` type. If the interface were not defined in `modules`, this circular import could be avoided. However, organizationally, I can see the benefit of defining all the module-related types in one place.

The `IP` field was also renamed to `NetAddress` for consistency. This will break existing contractor.json compatibility, but this seems excusable since contractor.json has only existed in RCs thus far.


lukechampine:
Sample `siac renter contracts` output:
```
Contracts:
Host                    Value     Data      End Height  ID
60.194.192.32:9982      1.711 KS  2.382 GB  50725       a8145d15416157a4a80cee3993dbd93cc9338061ce78f3d7eb7cd1e9b8d20935
148.251.221.163:9982    1.713 KS  2.047 GB  50725       c65d114afd1d44065f0d2ac25240b5d4d3a8aa6e2f6334834745d8e3b7b3a334
46.72.126.178:9982      1.711 KS  2.454 GB  50725       1288ff0c4c68609196730b773785c589c072bdeba98157a4252315e678429b0d
sia.pylon.org.uk:10982  1.711 KS  2.613 GB  50725       0770580b877e67e2063eaa8236c8fe75ce0615b84bc77da9a74c6ed9275acc77
95.147.183.112:9982      1.71 KS  2.747 GB  50725       054bf9ec90cfc39aade31eccf83b91ea8ddfa6a7a2ff68d68c5f4c9a33b15c99
71.71.203.175:9982      1.711 KS  2.332 GB  50725       59ac3bdc85b1b090055e55175babc5d8d9c9440db6dee3e9e1e5e88ef6bcbdaa
```

modules/renter/proto/negotiate.go
```diff
@@ -114,3 +114,116 @@ func verifyRecentRevision(conn net.Conn, contract Contract) error {
...
+	}
+	var hostSig types.TransactionSignature
+	if err := encoding.ReadObject(conn, &hostSig, 16e3); err != nil {
+		return types.Transaction{}, errors.New("couldn't read host's signature: " + err.Error())
+	}
+
+	// add the signature to the transaction and verify it
+	// NOTE: we can fake the blockheight here because it doesn't affect
+	// verification; it just needs to be above the fork height and below the
+	// contract expiration (which was checked earlier).
```
DavidVorick: `StandaloneValid` is influenced by an exact height, because we changed the fee at I think block 21,000. In production, it's not significant but if you aren't providing an exact height you might see some weird things in testing if the height is close to testing fork height.

DavidVorick: oh nvm that's only an issue for FileContracts, not revisions

lukechampine: cool, I was wondering about the testing constants as well

PR #1161, opened 2016-05-20
hostdb recognizes dynamic public keys

DavidVorick:
This is a temporary fix, as the hostdb should really be doing host
lookups by public key instead of by ip address. But, for v0.6.0 the
hostdb is going to continue looking up hosts by ip address. Now, it can
moderately gracefully handle a host changing public keys.

### DIFF
--- a/modules/renter/hostdb/scan.go
+++ b/modules/renter/hostdb/scan.go
@@ -5,6 +5,7 @@ package hostdb
 // settings of the hosts.
 
 import (
+	"bytes"
 	"crypto/rand"
 	"math/big"
 	"time"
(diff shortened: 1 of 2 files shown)


modules/renter/hostdb/scan.go
```diff
@@ -110,9 +113,16 @@ func (hdb *HostDB) threadedProbeHosts() {
...
+			priorHost, exists := hdb.allHosts[hostEntry.NetAddress]
+			if !exists {
 				hdb.allHosts[hostEntry.NetAddress] = hostEntry
 			}
+			if exists && !bytes.Equal(priorHost.PublicKey.Key, hostEntry.PublicKey.Key) {
+				// The public key in the announcement doesn't match the public
+				// key we have locally, and the host did not respond - can't be
+				// sure we have the right host, don't decrement reliability.
+				return
+			}
```
lukechampine: comments says `host did not respond`, but there's no `err != nil` here. Seems like this won't add ever add a host to `activeHosts` if the keys mismatch.

PR #1162, opened 2016-05-20
remove misplaced sanity check

DavidVorick:


modules/host/negotiaterecentrevision.go
```diff
@@ -62,17 +62,6 @@ func (h *Host) verifyChallengeResponse(fcid types.FileContractID, challenge cryp
 	if err != nil {
 		return nil, types.FileContractRevision{}, nil, err
 	}
-
-	// Sanity check - verify that the host has a valid revision and set of
-	// signatures.
-	h.mu.RLock()
-	blockHeight := h.blockHeight
-	h.mu.RUnlock()
-	err = modules.VerifyFileContractRevisionTransactionSignatures(recentRevision, revisionSigs, blockHeight)
```
lukechampine: what makes this misplaced? The renter calls this function on the received values.

DavidVorick: There were some errors being returned regarding the revision having been expired. From the host side, I see no reason to report an error if that's true. The renter should be allowed to request a recent revision for something that can't actually be submitted to the tpool anymore (presumably it was already submitted to the tpool).

PR #1164, opened 2016-05-20
Reduce gateway logging

### LINKED ISSUE #1047: Reduce gateway logging
Users have reported gateway logs on the order of gigabytes. There are lots of unnecessary logs like: `WARN: incoming RPC "RelayNod" from conn <peer> failed: node already added`.

VoidingWarranties:
Specifically, move "lost connection to peer" logging to Debugf as it could be triggered by a malicious peer.

And replace instances of RemoteAddr in logging with RPCAddr so that the logs are more easily compared with the gateway peer list.

Fixes #1047

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -14,7 +14,7 @@ func HttpGET(url string) (resp *http.Response, err error) {
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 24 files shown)


VoidingWarranties:
Closing and reopening on master.

PR #1166, opened 2016-05-20
Add IBD progress estimation

### LINKED ISSUE #929: siad progress bar for initial blockchain synchronization
It's annoying to keep running `siac status`.

VoidingWarranties:
New users might not know about https://explore.sia.tech so this PR provides a rough estimation in siac of IBD progress.

The blockchain height is estimated using the block time (10 minutes) and the time passed since a block at a known height and time (block 50000 was chosen because it is fairly recent and I'm assuming the hash rate has been stable recently).

It should be trivial to do something similar in the UI.

Example output:
```
$ siac
Synced: No
Height: 34253
Progress (estimated): 67%
```

Closes #929

### DIFF
--- a/siac/consensuscmd.go
+++ b/siac/consensuscmd.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"time"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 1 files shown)


lukechampine:
Neat! Would be nice to have a similar construct for the wallet, but there's no easy equivalent.

Unfortunately, it seems that our builds are now deterministically failing on Travis.

VoidingWarranties:
Closing and reopening on master.

PR #1170, opened 2016-05-21
upgrade host to v0.6.0 constants

DavidVorick:


modules/host/storageobligations.go
```diff
@@ -764,7 +764,7 @@ func (h *Host) handleActionItem(so *storageObligation) {
 
 		// Queue another action item to check whether there the storage proof
 		// got confirmed.
-		err = h.queueActionItem(h.blockHeight+types.BlockHeight(storageProofConfirmations), so.id())
+		err = h.queueActionItem(so.proofDeadline(), so.id())
```
lukechampine: makes me uneasy changing host stuff so close to release, but this seems simple enough that I trust it doesn't need live testing

DavidVorick: all this does is decide when the storage proof is confirmed. Instead of waiting `storageProofConfirmations` after the current height, it just waits until the `proofDeadline`, which in most cases will be a longer wait, unless for some reason the storage proof only got accepted at the last minute.

DavidVorick: none of the logic for actually submitting the storage proof has changed.

PR #1177, opened 2016-05-23
Return a descriptive error from /wallet/transactions

johnathanhowell:
This PR changes /wallet/transactions to return a descriptive error message if the `startheight` or `endheight` query parameters don't exist.  Previously, a strconv.ParseInt error would be returned, now `startheight and endheight must be provided to a /wallet/transactions call.` is returned, along with a `400 bad request` HTTP status code.

### DIFF
--- a/api/wallet.go
+++ b/api/wallet.go
@@ -371,12 +371,12 @@ func (srv *Server) walletTransactionsHandler(w http.ResponseWriter, req *http.Re
 	// Get the start and end blocks.
 	start, err := strconv.Atoi(startheightStr)
 	if err != nil {
-		writeError(w, "error after call to /wallet/transactions: "+err.Error(), http.StatusBadRequest)
+		writeError(w, "parsing integer value for parameter `startheight` failed: "+err.Error(), http.StatusBadRequest)
 		return
 	}
 	end, err := strconv.Atoi(endheightStr)
 	if err != nil {
-		writeError(w, "error after call to /wallet/transactions: "+err.Error(), http.StatusBadRequest)
+		writeError(w, "parsing integer value for parameter `endheight` failed: "+err.Error(), http.StatusBadRequest)
 		return
 	}
 	confirmedTxns, err := srv.wallet.Transactions(types.BlockHeight(start), types.BlockHeight(end))
--- a/api/wallet.go
+++ b/api/wallet.go
@@ -363,13 +363,20 @@ func (srv *Server) walletTransactionHandler(w http.ResponseWriter, req *http.Req
 
 // walletTransactionsHandler handles API calls to /wallet/transactions.
 func (srv *Server) walletTransactionsHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	// If 'startheight' or 'endheight' query parameters don't exist,
+	// return a descriptive error message along with a HTTP 400 Bad Request response code.
+	startheightStr, endheightStr := req.FormValue("startheight"), req.FormValue("endheight")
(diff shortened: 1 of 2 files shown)


api/wallet.go
```diff
@@ -371,13 +371,20 @@ func (srv *Server) walletTransactionHandler(w http.ResponseWriter, req *http.Req
...
+	startheightStr, endheightStr := req.FormValue("startheight"), req.FormValue("endheight")
+	if startheightStr == "" || endheightStr == "" {
+		writeError(w, "startheight and endheight must be provided to a /wallet/transactions call.", http.StatusBadRequest)
+		return
+	}
 	// Get the start and end blocks.
-	start, err := strconv.Atoi(req.FormValue("startheight"))
+	start, err := strconv.Atoi(startheightStr)
 	if err != nil {
 		writeError(w, "error after call to /wallet/transactions: "+err.Error(), http.StatusBadRequest)
```
VoidingWarranties: It might be helpful to change the wording of this error to be more descriptive as well. Same goes for parsing end height.

lukechampine: yeah. The contents of `err.Error()` won't be that useful. A hard-coded error message would be better here

api/wallet.go
```diff
@@ -371,13 +371,20 @@ func (srv *Server) walletTransactionHandler(w http.ResponseWriter, req *http.Req
 
 // walletTransactionsHandler handles API calls to /wallet/transactions.
 func (srv *Server) walletTransactionsHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	// If 'startheight' or 'endheight' query parameters don't exist,
+	// return a descriptive error message along with a HTTP 400 Bad Request response code.
```
lukechampine: generally we leave out this sort of comment, since it follows pretty clearly from the code

api/wallet_test.go
```diff
@@ -264,6 +264,15 @@ func TestIntegrationWalletTransactionGETid(t *testing.T) {
 	if len(wtg.UnconfirmedTransactions) != 0 {
 		t.Error("expecting 0 unconfirmed transactions")
 	}
+	// A call to /wallet/transactions without startheight and endheight parameters
+	// should return a descriptive error message.
+	err = st.getAPI("/wallet/transactions", &wtg)
+	if err == nil {
+		t.Error("expecting /wallet/transactions call with empty parameters to error")
+	}
+	if err.Error() != "startheight and endheight must be provided to a /wallet/transactions call.\n" {
```
lukechampine: this will still panic if `err == nil`, because `t.Error` doesn't stop execution. Prefer:
```go
if err == nil || err.Error() != "blah blah blah" {

}
```

lukechampine:
We should explore the use of `net/http/httptest` for easier HTTP testing. Low priority for now.

api/wallet.go
VoidingWarranties: I think the error string can still be useful, as `Atoi` can return one of a few different errors. The "error after call to /wallet/transactions:" string seems unnecessary though as callers should know what endpoint they're requesting.

DavidVorick:
What do we need to change to prevent these errors? I thought we were vendoring twofish now?

VoidingWarranties:
Just need to rebase and push again

DavidVorick:
sorry for the turmoil, but it needs a rebase again. Side effect of having lots of PRs coming in.

DavidVorick:
after you do the rebase, you should drop a comment so that we know to merge it, otherwise we won't get a notification that it's ready.

johnathanhowell:
ACK.  Just rebased.

PR #1178, opened 2016-05-23
Gateway clean shutdown

VoidingWarranties:
I chose to put the `wg.Add(1)` `wg.Done()`s on the calling side because they are not necessarily called from goroutines. The godocs also suggest putting the `Add` outside the goroutine. It also makes it easy to visually check that all goroutines have a corresponding `Add` and `Done`.

### DIFF
--- a/modules/gateway/gateway.go
+++ b/modules/gateway/gateway.go
@@ -11,7 +11,6 @@ import (
 	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/persist"
-	siasync "github.com/NebulousLabs/Sia/sync"
 )
 
 var (
@@ -44,13 +43,13 @@ type Gateway struct {
 	persistDir string
 
 	log *persist.Logger
-	mu  *siasync.RWMutex
+	mu  sync.RWMutex
 }
 
 // Address returns the NetAddress of the Gateway.
 func (g *Gateway) Address() modules.NetAddress {
-	id := g.mu.RLock()
-	defer g.mu.RUnlock(id)
+	g.mu.RLock()
+	defer g.mu.RUnlock()
 	return g.myAddr
 }
 
@@ -64,17 +63,17 @@ func (g *Gateway) Close() error {
 	g.UnregisterRPC("ShareNodes")
 	g.UnregisterConnectCall("ShareNodes")
 	// save the latest gateway state
-	id := g.mu.RLock()
+	g.mu.RLock()
 	if err := g.saveSync(); err != nil {
 		errs = append(errs, fmt.Errorf("save failed: %v", err))
 	}
-	g.mu.RUnlock(id)
+	g.mu.RUnlock()
 	// send close signal
 	close(g.closeChan)
 	// clear the port mapping (no effect if UPnP not supported)
-	id = g.mu.RLock()
+	g.mu.RLock()
 	g.clearPort(g.myAddr.Port())
-	g.mu.RUnlock(id)
+	g.mu.RUnlock()
 	// shut down the listener
 	if err := g.listener.Close(); err != nil {
 		errs = append(errs, fmt.Errorf("listener.Close failed: %v", err))
@@ -111,7 +110,6 @@ func New(addr string, persistDir string) (g *Gateway, err error) {
 		nodes:      make(map[modules.NetAddress]struct{}),
 		closeChan:  make(chan struct{}),
 		persistDir: persistDir,
-		mu:         siasync.New(modules.SafeMutexDelay, 2),
 	}
 
 	// Create the logger.
--- a/modules/gateway/gateway.go
+++ b/modules/gateway/gateway.go
@@ -6,12 +6,12 @@ import (
 	"net"
 	"os"
 	"path/filepath"
-	"time"
+	"sync"
 
 	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/persist"
-	"github.com/NebulousLabs/Sia/sync"
+	siasync "github.com/NebulousLabs/Sia/sync"
 )
 
 var (
@@ -38,11 +38,13 @@ type Gateway struct {
 
 	// closeChan is used to shut down the Gateway's goroutines.
 	closeChan chan struct{}
+	// closeWG is used to wait for all goroutines to exit before returning from Close().
+	closeWG sync.WaitGroup
 
 	persistDir string
 
 	log *persist.Logger
-	mu  *sync.RWMutex
+	mu  *siasync.RWMutex
 }
 
 // Address returns the NetAddress of the Gateway.
@@ -83,11 +85,8 @@ func (g *Gateway) Close() error {
 			errs = append(errs, fmt.Errorf("Disconnect failed: %v", err))
 		}
 	}
-	// Sleep to give time for all goroutines to exit. This is necessary because
-	// some goroutines write to the logger so we must give them time to exit
-	// before closing the logger.
-	// TODO: block until goroutines exit instead of sleeping.
-	time.Sleep(100 * time.Millisecond)
+	// Wait for goroutines to exit.
+	g.closeWG.Wait()
 	// Close the logger. The logger should be the last thing to shut down so that
 	// all other objects have access to logging while closing.
 	if err := g.log.Close(); err != nil {
@@ -112,7 +111,7 @@ func New(addr string, persistDir string) (g *Gateway, err error) {
 		nodes:      make(map[modules.NetAddress]struct{}),
 		closeChan:  make(chan struct{}),
 		persistDir: persistDir,
-		mu:         sync.New(modules.SafeMutexDelay, 2),
+		mu:         siasync.New(modules.SafeMutexDelay, 2),
 	}
 
 	// Create the logger.
@@ -159,18 +158,38 @@ func New(addr string, persistDir string) (g *Gateway, err error) {
 	g.log.Println("INFO: gateway created, started logging")
 
 	// Forward the RPC port, if possible.
-	go g.forwardPort(port)
+	g.closeWG.Add(1)
+	go func() {
+		defer g.closeWG.Done()
+		g.forwardPort(port)
+	}()
 
 	// Learn our external IP.
-	go g.learnHostname(port)
+	g.closeWG.Add(1)
+	go func() {
+		defer g.closeWG.Done()
+		g.learnHostname(port)
+	}()
 
 	// Spawn the peer and node managers. These will attempt to keep the peer
 	// and node lists healthy.
-	go g.threadedPeerManager()
-	go g.threadedNodeManager()
+	g.closeWG.Add(1)
+	go func() {
+		defer g.closeWG.Done()
+		g.threadedPeerManager()
+	}()
+	g.closeWG.Add(1)
+	go func() {
+		defer g.closeWG.Done()
+		g.threadedNodeManager()
+	}()
 
 	// Spawn the primary listener.
-	go g.listen()
+	g.closeWG.Add(1)
+	go func() {
+		defer g.closeWG.Done()
+		g.listen()
+	}()
 
 	return
 }
--- a/modules/gateway/gateway.go
+++ b/modules/gateway/gateway.go
@@ -133,6 +133,14 @@ func New(addr string, persistDir string) (g *Gateway, err error) {
 	if err != nil {
 		return
 	}
+	// Automatically close the listener when g.threads.Stop() is called.
+	g.threads.OnStop(func() {
+		err := g.listener.Close()
+		if err != nil {
+			g.log.Println("WARN: closing the listener failed:", err)
+		}
+	})
+
 	_, port, portErr := net.SplitHostPort(g.listener.Addr().String())
 	if portErr != nil {
 		return nil, portErr
--- a/modules/gateway/gateway.go
+++ b/modules/gateway/gateway.go
@@ -11,6 +11,7 @@ import (
 	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/persist"
+	siasync "github.com/NebulousLabs/Sia/sync"
 )
 
 var (
@@ -35,10 +36,9 @@ type Gateway struct {
 	// network.
 	nodes map[modules.NetAddress]struct{}
 
-	// closeChan is used to shut down the Gateway's goroutines.
-	closeChan chan struct{}
-	// closeWG is used to wait for all goroutines to exit before returning from Close().
-	closeWG sync.WaitGroup
+	// threads is used to signal the Gateway's goroutines to shut down and to wait
+	// for all goroutines to exit before returning from Close().
+	threads siasync.ThreadGroup
 
 	persistDir string
 
@@ -55,8 +55,11 @@ func (g *Gateway) Address() modules.NetAddress {
 
 // Close saves the state of the Gateway and stops its listener process.
 func (g *Gateway) Close() error {
-	var errs []error
+	if err := g.threads.Stop(); err != nil {
+		return err
+	}
 
+	var errs []error
 	// Unregister RPCs. Not strictly necessary for clean shutdown in this specific
 	// case, as the handlers should only contain references to the gateway itself,
 	// but do it anyways as an example for other modules to follow.
@@ -68,24 +71,10 @@ func (g *Gateway) Close() error {
 		errs = append(errs, fmt.Errorf("save failed: %v", err))
 	}
 	g.mu.RUnlock()
-	// send close signal
-	close(g.closeChan)
 	// clear the port mapping (no effect if UPnP not supported)
 	g.mu.RLock()
 	g.clearPort(g.myAddr.Port())
 	g.mu.RUnlock()
-	// shut down the listener
-	if err := g.listener.Close(); err != nil {
-		errs = append(errs, fmt.Errorf("listener.Close failed: %v", err))
-	}
-	// Disconnect from peers.
-	for _, p := range g.Peers() {
-		if err := g.Disconnect(p.NetAddress); err != nil {
-			errs = append(errs, fmt.Errorf("Disconnect failed: %v", err))
-		}
-	}
-	// Wait for goroutines to exit.
-	g.closeWG.Wait()
 	// Close the logger. The logger should be the last thing to shut down so that
 	// all other objects have access to logging while closing.
 	if err := g.log.Close(); err != nil {
@@ -108,7 +97,6 @@ func New(addr string, persistDir string) (g *Gateway, err error) {
 		initRPCs:   make(map[string]modules.RPCFunc),
 		peers:      make(map[modules.NetAddress]*peer),
 		nodes:      make(map[modules.NetAddress]struct{}),
-		closeChan:  make(chan struct{}),
 		persistDir: persistDir,
 	}
 
@@ -156,38 +144,30 @@ func New(addr string, persistDir string) (g *Gateway, err error) {
 	g.log.Println("INFO: gateway created, started logging")
 
 	// Forward the RPC port, if possible.
-	g.closeWG.Add(1)
 	go func() {
-		defer g.closeWG.Done()
+		if err := g.threads.Add(); err != nil {
+			return
+		}
+		defer g.threads.Done()
 		g.forwardPort(port)
 	}()
 
 	// Learn our external IP.
-	g.closeWG.Add(1)
 	go func() {
-		defer g.closeWG.Done()
+		if err := g.threads.Add(); err != nil {
+			return
+		}
+		defer g.threads.Done()
 		g.learnHostname(port)
 	}()
 
 	// Spawn the peer and node managers. These will attempt to keep the peer
 	// and node lists healthy.
-	g.closeWG.Add(1)
-	go func() {
-		defer g.closeWG.Done()
-		g.threadedPeerManager()
-	}()
-	g.closeWG.Add(1)
-	go func() {
-		defer g.closeWG.Done()
-		g.threadedNodeManager()
-	}()
+	go g.threadedPeerManager()
+	go g.threadedNodeManager()
 
 	// Spawn the primary listener.
-	g.closeWG.Add(1)
-	go func() {
-		defer g.closeWG.Done()
-		g.listen()
-	}()
+	go g.threadedListen()
 
 	return
 }
--- a/modules/gateway/peers.go
+++ b/modules/gateway/peers.go
@@ -13,8 +13,6 @@ import (
 )
 
 const (
-	// the gateway will abort a connection attempt after this long
-	dialTimeout = 2 * time.Minute
 	// the gateway will not accept inbound connections above this threshold
 	fullyConnectedThreshold = 128
 	// the gateway will ask for more addresses below this threshold
@@ -31,6 +29,19 @@ var (
 			return 3 * time.Second
 		case "testing":
 			return 10 * time.Millisecond
+		default:
+			panic("unrecognized build.Release")
+		}
+	}()
+	// the gateway will abort a connection attempt after this long
+	dialTimeout = func() time.Duration {
+		switch build.Release {
+		case "dev":
+			return 2 * time.Minute
+		case "standard":
+			return 2 * time.Minute
+		case "testing":
+			return 2 * time.Second
 		default:
 			panic("unrecognized build.Release")
 		}
--- a/modules/gateway/peers.go
+++ b/modules/gateway/peers.go
@@ -72,7 +72,11 @@ func (p *peer) accept() (modules.PeerConn, error) {
 // to handle its requests.
 func (g *Gateway) addPeer(p *peer) {
 	g.peers[p.NetAddress] = p
-	go g.listenPeer(p)
+	g.closeWG.Add(1)
+	go func() {
+		defer g.closeWG.Done()
+		g.listenPeer(p)
+	}()
 }
 
 // randomPeer returns a random peer from the gateway's peer list.
@@ -118,7 +122,11 @@ func (g *Gateway) listen() {
 			return
 		}
 
-		go g.acceptConn(conn)
+		g.closeWG.Add(1)
+		go func() {
+			defer g.closeWG.Done()
+			g.acceptConn(conn)
+		}()
 
 		// Sleep after each accept. This limits the rate at which the Gateway
 		// will accept new connections. The intent here is to prevent new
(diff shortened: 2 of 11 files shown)


VoidingWarranties:
Forcing all tests that use `newTestingGateway` to be long tests is probably a little heavy handed, but it's the result of `Sync` sometimes taking a few seconds on Travis. This was discussed in #1152.

lukechampine:
I'm worried about threads calling `wg.Add(1)` concurrently with `wg.Wait()`. I tested this in the hostdb and it caused a race condition.

It seems to me that the most rigorous way of handling this would be a simple convention:
The function that spawns a goroutine is responsible for ensuring that the goroutine exits during shutdown.
This means that a function like `listen`, which spawns goroutines for each listener, is responsible for ensuring that those goroutines have exited before `listen` itself exits. It may therefore use its own internal waitgroup:
```go
func (g *Gateway) listen() {
	var wg sync.WaitGroup
	defer wg.Wait()
	for {
		conn, err := g.listener.Accept()
		if err != nil {
			return
		}
		wg.Add(1)
		go func() {
			g.acceptConn(conn)
			wg.Done()
		}()
		select {
		case <-time.After(acceptInterval):
		case <-g.closeChan:
			return
		}
	}
}
```
This might be complicated by functions that spawn goroutines that live longer than their parent. (Although, we should strive to avoid that pattern outside of the `New` functions.)

VoidingWarranties:
That pattern is great and works well in most places, except as you mentioned when the goroutine outlives the parent. The only place I can find where that is desired is in `Connect`, which calls all of the on-connect RPCs on the peer. Specifically the consensus' `SendBlocks` on-connect RPC could cause the call to `Connect` to not return for 5 minutes.

In this case I'm not sure if it's better to not track the goroutine with any wait group or to use the `g.closeWG` wait group. Because `Connect` is an exported method that isn't called very often internally (only by `threadedPeerManager` infrequently), it seems safe-ish to use the object-global wait group.

VoidingWarranties:
Hmmm `addPeer` also uses go routines that outlive it. I suppose using this pattern in some places is better than not using it at all.

lukechampine:
It seems the problem is primarily with exported functions. Any exported function is a "parent," because it was spawned from the outside, not by `New`. Hmm.

DavidVorick:
The way I handled this in the host was with the resource lock. All threads had a finite amount of time that they would hold the resource lock. Infinite threads had periodic moments where they release and re-grab the resource lock. If the resource bool is set to false, the threads know that shutdown is occurring and they exit.

All exported functions also reference this shutdown variable. Upon closing, you lock the resource lock, meaning you'll block until everyone has released the resource lock, and then you can set the variable to 'false', knowing that nobody is actively holding it and that exported functions will return an error and exit immediately.

modules/gateway/gateway.go
```diff
@@ -6,12 +6,12 @@ import (
...
 	"os"
 	"path/filepath"
-	"time"
+	"sync"
 
 	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/persist"
-	"github.com/NebulousLabs/Sia/sync"
+	siasync "github.com/NebulousLabs/Sia/sync"
```
DavidVorick: We probably don't need safelocks in the gateway anymore, you can replace them with standard mutexes.

VoidingWarranties:
Updated to use the new `ThreadGroup` type. I wasn't sure which methods should be responsible for calling `threads.Add` and `threads.Done`. Is every exported method responsible for calling these? Or maybe every exported method that can return an error? As of now it's not consistent. Some exported methods that are frequently called from goroutines do call `threads.Add` and `threads.Done` whereas others that aren't usually called from goroutines do not. I haven't measured the added cost of calling these methods, but it might be negligible enough to blindly call them in every exported method. Unexported methods that begin with `threaded` are also responsible for calling `threads.Add` and `threads.Done`.

DavidVorick:
You need to add `ThreadGroup` use to every single function or thread that needs a resource which can close. So, if it's a function which uses the logger, you need to use the ThreadGroup, etc.

modules/gateway/gateway.go
```diff
@@ -36,58 +37,45 @@ type Gateway struct {
...
 }
 
 // Address returns the NetAddress of the Gateway.
 func (g *Gateway) Address() modules.NetAddress {
-	id := g.mu.RLock()
-	defer g.mu.RUnlock(id)
+	g.mu.RLock()
+	defer g.mu.RUnlock()
 	return g.myAddr
 }
```
DavidVorick: As an example. you don't need to use the ThreadGroup here because there's no use of a closeable resource in this thread.

modules/gateway/peers.go
```diff
@@ -109,27 +120,53 @@ func (g *Gateway) randomInboundPeer() (modules.NetAddress, error) {
...
+	if g.threads.Add() != nil {
+		return
+	}
+	go func() {
+		defer g.threads.Done()
+		<-g.threads.StopChan()
+		if err := g.listener.Close(); err != nil {
+			g.log.Printf("WARN: listener.Close failed: %v", err)
+		}
+	}()
```
DavidVorick: It took me a little bit looking to understand what was happening here, and then another bit of looking to certain that it was correct. It does look correct, but there might be a cleaner/clearer way to accomplish what this code does.

modules/gateway/gateway.go
```diff
@@ -159,18 +145,26 @@ func New(addr string, persistDir string) (g *Gateway, err error) {
...
+	go func() {
+		defer g.threads.Done()
+		g.forwardPort(port)
+	}()
 
 	// Learn our external IP.
-	go g.learnHostname(port)
+	g.threads.Add()
+	go func() {
+		defer g.threads.Done()
```
lukechampine: unlike a normal waitgroup, it's safe to do the `Add` inside the thread. I think this should be the preferred style.

modules/gateway/peers.go
lukechampine: The `g.listener.Close()` could be moved to `gateway.Close`, before the `g.threads.Stop()`. The question is: should `g.threads.Stop()` be the sole way of stopping things? If so, we need to do things like the above. If not, we can perform other shutdown-related activity in `Close` methods. My preference is the latter; it's less surprising.

modules/gateway/peers.go
```diff
@@ -257,31 +299,33 @@ func (g *Gateway) Connect(addr modules.NetAddress) error {
...
+	g.mu.Unlock()
 
 	// call initRPCs
-	id = g.mu.RLock()
+	g.mu.RLock()
 	for name, fn := range g.initRPCs {
-		go g.RPC(addr, name, fn)
+		go func(name string, fn modules.RPCFunc) {
+			g.RPC(addr, name, fn)
+		}(name, fn)
```
lukechampine: interesting that this got mangled. Looks like a statement was added in the body, then removed later?

modules/gateway/rpc.go
```diff
@@ -93,17 +98,34 @@ func (g *Gateway) RegisterConnectCall(name string, fn modules.RPCFunc) {
...
+	defer g.threads.Done()
+
+	// Disconnect from the peer when the stop signal is received.
+	go func() {
+		if g.threads.Add() != nil {
+			return
+		}
+		defer g.threads.Done()
+		<-g.threads.StopChan()
+		if err := p.sess.Close(); err != nil {
```
lukechampine: ditto for this. Easier to loop through sessions in `gateway.Close`

lukechampine:
Test failures indicate that there's a deadlock somewhere

modules/gateway/peers.go
DavidVorick: I prefer having `g.threads.Stop()` be the first thing that you do when shutting down. It doesn't need to be the sole way of closing things, but in this case it's the only way to guarantee that `listener.Close` doesn't get called multiple times.

modules/gateway/gateway.go
DavidVorick: That is my understanding as well. I find it odd that a normal waitgroup does not support concurrent calls to `Add()`

lukechampine: it sort-of does, but it introduces a race condition, because you can `Add` right when `Wait` thinks it's finished. With `ThreadGroup`, calls to `Add` during `Stop` return immediately with an error, so there's no race condition.

VoidingWarranties: The advantage of doing the `Add` outside the goroutine is that you do not need to check the return value, because you know that the gateway cannot be closed inside the `New` function.

modules/gateway/peers.go
VoidingWarranties: Correct. I kept the function body because the next PR will introduce error logging for the RPC and keeping the body will make the merge easier (the `gateway-logging` PR originally made on the `v061` branch)

modules/gateway/rpc.go
VoidingWarranties: This code has a hidden bug too. `Add` is called inside the goroutine, which means that this goroutine might only start after the gateway has been closed, preventing the peers from disconnecting. This might be the cause of the deadlock. Closing the sessions in `gateway.Close` is easier to implement bug free and is a lot cleaner.

VoidingWarranties:
Closing the listener and peer sessions in `gateway.Close` is problematic, as concurrent calls to `gateway.Close` could attempt to close the listener and peer sessions multiple times. This might work, but according to @lukechampine closing multiple times is often undefined.

Adding a `if g.threads.IsStopped() { return errClosed }` at the top of `gateway.Close()` wouldn't help, as you would need to hold a lock from the `IsStopped` check until the call to `Stop`.

It is a little uglier, but it is also nice to have every `threaded` method responsible for closing itself, either with a `select` using the `StopChan` or by spawning a goroutine and waiting for the `StopChan`.

VoidingWarranties:
Deadlock fixed in 86346d9703ffd4dc11d8c328be8ca8fc36ad0769, but there are still some non-deterministic deadlocks in the host. It appears that writing bug free clean shutdown code is not easy.

modules/gateway/gateway.go
```diff
@@ -6,17 +6,18 @@ import (
...
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/persist"
-	"github.com/NebulousLabs/Sia/sync"
+	siasync "github.com/NebulousLabs/Sia/sync"
 )
 
 var (
 	errNoPeers     = errors.New("no peers")
 	errUnreachable = errors.New("peer did not respond to ping")
+	errClosed      = errors.New("Gateway is closed")
```
DavidVorick: probably better to use the error from the ThreadGroup, and then let the higher level process specify the module.

modules/gateway/gateway.go
```diff
@@ -36,58 +37,45 @@ type Gateway struct {
...
+	g.mu.RLock()
+	defer g.mu.RUnlock()
 	return g.myAddr
 }
 
 // Close saves the state of the Gateway and stops its listener process.
 func (g *Gateway) Close() error {
-	var errs []error
+	if g.threads.Stop() != nil {
+		return errClosed
```
DavidVorick: e.g. just return the error here.

modules/gateway/gateway.go
VoidingWarranties: Fixed.

modules/gateway/gateway.go
```diff
@@ -159,18 +144,26 @@ func New(addr string, persistDir string) (g *Gateway, err error) {
 	g.log.Println("INFO: gateway created, started logging")
 
 	// Forward the RPC port, if possible.
-	go g.forwardPort(port)
+	g.threads.Add()
+	go func() {
+		defer g.threads.Done()
+		g.forwardPort(port)
```
DavidVorick: I believe this function should be called `threadedForwardPort`, given how it is being used here.

modules/gateway/gateway.go
DavidVorick: Always check your error values, doesn't matter if you know it can't return an error. Knowing that something can't return an error is an implementation detail,which means your code is brittle against implementation changes.

We've had enough bugs crop up from unchecked errors that I feel we need to be pedantic about it.

modules/gateway/peers.go
```diff
@@ -35,6 +33,19 @@ var (
 			panic("unrecognized build.Release")
 		}
 	}()
+	// the gateway will abort a connection attempt after this long
+	dialTimeout = func() time.Duration {
+		switch build.Release {
+		case "dev":
+			return 2 * time.Minute
```
DavidVorick: dev is meant to be on your local machine, 10 seconds is definitely enough for a timeout.

DavidVorick: as in, the whole network is on your local machine.

modules/gateway/peers.go
```diff
@@ -315,20 +371,25 @@ func (g *Gateway) threadedPeerManager() {
 		// spawn a goroutine and sleep for five seconds. This allows us to
 		// continue making connections if the node is unresponsive.
 		if err == nil {
-			go g.Connect(addr)
+			go func() {
+				connectErr := g.Connect(addr)
+				if connectErr != nil {
+					g.log.Println("WARN: automatic connect failed:", connectErr)
```
DavidVorick: Better to have this be a Debugln I think.

modules/gateway/peers.go
```diff
@@ -315,20 +371,25 @@ func (g *Gateway) threadedPeerManager() {
...
+				}
+			}()
 		}
 		select {
 		case <-time.After(5 * time.Second):
```
DavidVorick: 5 seconds strikes me as a really short timeout.

modules/gateway/rpc.go
```diff
@@ -93,17 +98,37 @@ func (g *Gateway) RegisterConnectCall(name string, fn modules.RPCFunc) {
...
+	if g.threads.Add() != nil {
+		return
+	}
+	defer g.threads.Done()
+
+	// Disconnect from the peer when the stop signal is received.
+	if g.threads.Add() != nil {
+		return
+	}
+	go func() {
```
DavidVorick: Does this code take into account situations where you disconnect from the peer before `Close()` is called? What happens to this goroutine if your peer leaves early?

I think right now it's a memory leak.

modules/gateway/rpc.go
```diff
@@ -162,16 +198,24 @@ func (g *Gateway) Broadcast(name string, obj interface{}, peers []modules.Peer)
...
 		go func(addr modules.NetAddress) {
+			defer wg.Done()
 			err := g.RPC(addr, name, fn)
 			if err != nil {
+				g.log.Debugf("WARN: broadcasting RPC %q to peer %q failed (attempting again in 10 seconds): %v", name, addr, err)
 				// try one more time before giving up
-				time.Sleep(10 * time.Second)
-				g.RPC(addr, name, fn)
+				select {
+				case <-time.After(10 * time.Second):
```
DavidVorick: 10 seconds is both not long enough and is a magical number. Can we create a const for this value? Also, the gateway I think has enough files that it deserves a `consts.go`, but that should be done in a different PR.

modules/gateway/gateway.go
VoidingWarranties: I've been using the `threaded` prefix to indicate that the method needs to be called from a goroutine, not that it _should_ be called from a goroutine. Same goes for `learnHostname`.

modules/gateway/peers.go
VoidingWarranties: I can't remember why I added this commit. I think it was because a test was failing and I suspected it might have been due to a long `dialTimeout`. This change really belongs in another PR. I'll remove it from this PR since there's a bunch of constants to update and they can all go in the same PR.

modules/gateway/rpc.go
VoidingWarranties: Good catch.

modules/gateway/gateway.go
DavidVorick: In the host, `forwardPort` changes the state of the host, mostly to grab the port number. I thought the gateway did the same thing, but it seems like it doesn't.

DavidVorick: However, it is `g.forwardPort`, and there's a pretty strict convention that you need to assume unexported methods on an object hold a lock. So, either you need to change it to `g.threadedForwardPort`, `g.ForwardPort`, or `forwardPort` (last one is not a method on the gateway).

Definitely the way it's being used here is incorrect, you can't call an unexported method on the gateway in a goroutine unless said goroutine wraps the call in a lock.

Given that no state is changing, I'd probably prefer `forwardPort`

VoidingWarranties: I thought unexported methods that don't begin with `threaded` or `managed` do **not** hold a lock.

DavidVorick:
looks good to me

DavidVorick:
I take that back, the build failure for the go module is a deadlock - the threadgroup is blocking after a call to `Stop`, but there doesn't seem to be any thread which is waiting to call `Done`, which means there is somewhere that you are calling `Add` which is not being cleaned up correctly.

```
panic: test timed out after 5m0s

goroutine 243 [running]:

panic(0x968a80, 0xc82011d460)

	/home/travis/.gimme/versions/go1.6.linux.amd64/src/runtime/panic.go:464 +0x3ff

testing.startAlarm.func1()

	/home/travis/.gimme/versions/go1.6.linux.amd64/src/testing/testing.go:725 +0x184

created by time.goFunc

	/home/travis/.gimme/versions/go1.6.linux.amd64/src/time/sleep.go:129 +0x6e

goroutine 1 [chan receive]:

testing.RunTests(0xbe6118, 0xe7b780, 0x1d, 0x1d, 0x1)

	/home/travis/.gimme/versions/go1.6.linux.amd64/src/testing/testing.go:583 +0xb39

testing.(*M).Run(0xc820051f00, 0xead6d5)

	/home/travis/.gimme/versions/go1.6.linux.amd64/src/testing/testing.go:515 +0x11e

main.main()

	github.com/NebulousLabs/Sia/modules/gateway/_test/_testmain.go:110 +0x211

goroutine 17 [syscall, locked to thread]:

runtime.goexit()

	/home/travis/.gimme/versions/go1.6.linux.amd64/src/runtime/asm_amd64.s:1998 +0x1

goroutine 143 [IO wait]:

net.runtime_pollWait(0x2b967bf60db8, 0x77, 0xc82004d020)

	/home/travis/.gimme/versions/go1.6.linux.amd64/src/runtime/netpoll.go:160 +0x63

net.(*pollDesc).Wait(0xc820118370, 0x77, 0x0, 0x0)

	/home/travis/.gimme/versions/go1.6.linux.amd64/src/net/fd_poll_runtime.go:73 +0x56

net.(*pollDesc).WaitWrite(0xc820118370, 0x0, 0x0)

	/home/travis/.gimme/versions/go1.6.linux.amd64/src/net/fd_poll_runtime.go:82 +0x44

net.(*netFD).connect(0xc820118310, 0x0, 0x0, 0x2b967bf60190, 0xc8201ce5e0, 0xecedffb48, 0x3b2b14f8, 0xe90d20, 0x0, 0x0, ...)

	/home/travis/.gimme/versions/go1.6.linux.amd64/src/net/fd_unix.go:127 +0x2e7

net.(*netFD).dial(0xc820118310, 0x2b967bf61248, 0x0, 0x2b967bf61248, 0xc8201ca660, 0xecedffb48, 0xc83b2b14f8, 0xe90d20, 0x0, 0x0, ...)

	/home/travis/.gimme/versions/go1.6.linux.amd64/src/net/sock_posix.go:137 +0x429

net.socket(0xae5958, 0x3, 0x2, 0x1, 0x0, 0xc8201ca600, 0x2b967bf61248, 0x0, 0x2b967bf61248, 0xc8201ca660, ...)

	/home/travis/.gimme/versions/go1.6.linux.amd64/src/net/sock_posix.go:89 +0x507

net.internetSocket(0xae5958, 0x3, 0x2b967bf61248, 0x0, 0x2b967bf61248, 0xc8201ca660, 0xecedffb48, 0x3b2b14f8, 0xe90d20, 0x1, ...)

	/home/travis/.gimme/versions/go1.6.linux.amd64/src/net/ipsock_posix.go:161 +0x164

net.dialTCP(0xae5958, 0x3, 0x0, 0xc8201ca660, 0xecedffb48, 0x3b2b14f8, 0xe90d20, 0x0, 0x20000004455c0, 0x0, ...)

	/home/travis/.gimme/versions/go1.6.linux.amd64/src/net/tcpsock_posix.go:171 +0x13c

net.dialSingle(0xc82001c120, 0x2b967bf611e0, 0xc8201ca660, 0xecedffb48, 0x3b2b14f8, 0xe90d20, 0x0, 0x0, 0x0, 0x0)

	/home/travis/.gimme/versions/go1.6.linux.amd64/src/net/dial.go:371 +0x50d

net.dialSerial.func1(0xecedffb48, 0x3b2b14f8, 0xe90d20, 0x0, 0x0, 0x0, 0x0)

	/home/travis/.gimme/versions/go1.6.linux.amd64/src/net/dial.go:343 +0xb5

net.dial(0xae5958, 0x3, 0x2b967bf611e0, 0xc8201ca660, 0xc82004d990, 0xecedffb48, 0x3b2b14f8, 0xe90d20, 0x0, 0x0, ...)

	/home/travis/.gimme/versions/go1.6.linux.amd64/src/net/fd_unix.go:40 +0x7a

net.dialSerial(0xc82001c120, 0xc82011cb30, 0x1, 0x1, 0x0, 0x0, 0x0, 0x0, 0x0)

	/home/travis/.gimme/versions/go1.6.linux.amd64/src/net/dial.go:345 +0xa82

net.(*Dialer).Dial(0xc82004dc30, 0xae5958, 0x3, 0xb377e0, 0x14, 0x0, 0x0, 0x0, 0x0)

	/home/travis/.gimme/versions/go1.6.linux.amd64/src/net/dial.go:239 +0x66d

net.DialTimeout(0xae5958, 0x3, 0xb377e0, 0x14, 0x1bf08eb000, 0x0, 0x0, 0x0, 0x0)

	/home/travis/.gimme/versions/go1.6.linux.amd64/src/net/dial.go:200 +0xb4

github.com/NebulousLabs/Sia/modules/gateway.(*Gateway).Connect(0xc8201e6180, 0xb377e0, 0x14, 0x0, 0x0)

	/home/travis/gopath/src/github.com/NebulousLabs/Sia/modules/gateway/peers.go:239 +0x5ef

github.com/NebulousLabs/Sia/modules/gateway.(*Gateway).threadedPeerManager.func1(0xc8201e6180, 0xb377e0, 0x14)

	/home/travis/gopath/src/github.com/NebulousLabs/Sia/modules/gateway/peers.go:352 +0x6b

created by github.com/NebulousLabs/Sia/modules/gateway.(*Gateway).threadedPeerManager

	/home/travis/gopath/src/github.com/NebulousLabs/Sia/modules/gateway/peers.go:356 +0x38f

goroutine 202 [semacquire]:

sync.runtime_Semacquire(0xc8201e61e8)

	/home/travis/.gimme/versions/go1.6.linux.amd64/src/runtime/sema.go:47 +0x26

sync.(*WaitGroup).Wait(0xc8201e61dc)

	/home/travis/.gimme/versions/go1.6.linux.amd64/src/sync/waitgroup.go:127 +0x118

github.com/NebulousLabs/Sia/sync.(*ThreadGroup).Stop(0xc8201e61c0, 0x0, 0x0)

	/home/travis/gopath/src/github.com/NebulousLabs/Sia/sync/threadgroup.go:82 +0x1bc

github.com/NebulousLabs/Sia/modules/gateway.(*Gateway).Close(0xc8201e6180, 0x0, 0x0)

	/home/travis/gopath/src/github.com/NebulousLabs/Sia/modules/gateway/gateway.go:58 +0x7e

github.com/NebulousLabs/Sia/modules/gateway.TestConnect(0xc8201e2360)

	/home/travis/gopath/src/github.com/NebulousLabs/Sia/modules/gateway/peers_test.go:206 +0xc40

testing.tRunner(0xc8201e2360, 0xe7b8a0)

	/home/travis/.gimme/versions/go1.6.linux.amd64/src/testing/testing.go:473 +0xdd

created by testing.RunTests

	/home/travis/.gimme/versions/go1.6.linux.amd64/src/testing/testing.go:582 +0xae3

goroutine 180 [semacquire]:

sync.runtime_Syncsemacquire(0xc8201760d0)

	/home/travis/.gimme/versions/go1.6.linux.amd64/src/runtime/sema.go:241 +0x201

sync.(*Cond).Wait(0xc8201760c0)

	/home/travis/.gimme/versions/go1.6.linux.amd64/src/sync/cond.go:63 +0xb3

github.com/NebulousLabs/muxado/proto/buffer.(*Inbound).Read(0xc8201e0060, 0xc82011d1a8, 0x4, 0x8, 0x0, 0x0, 0x0)

	/home/travis/gopath/src/github.com/NebulousLabs/muxado/proto/buffer/inbound.go:155 +0x697

github.com/NebulousLabs/muxado/proto.(*Stream).Read(0xc82001e100, 0xc82011d1a8, 0x4, 0x8, 0x0, 0x0, 0x0)

	/home/travis/gopath/src/github.com/NebulousLabs/muxado/proto/stream.go:73 +0x79

io.ReadAtLeast(0x2b967bf65a48, 0xc82001e100, 0xc82011d1a8, 0x4, 0x8, 0x4, 0x0, 0x0, 0x0)

	/home/travis/.gimme/versions/go1.6.linux.amd64/src/io/io.go:297 +0x119

io.ReadFull(0x2b967bf65a48, 0xc82001e100, 0xc82011d1a8, 0x4, 0x8, 0x80, 0x0, 0x0)

	/home/travis/.gimme/versions/go1.6.linux.amd64/src/io/io.go:315 +0x77

encoding/binary.Read(0x2b967bf65a48, 0xc82001e100, 0x2b967bfbf0d8, 0xeae4c0, 0x94dfc0, 0xc82011d1a0, 0x0, 0x0)

	/home/travis/.gimme/versions/go1.6.linux.amd64/src/encoding/binary/binary.go:152 +0x16d

github.com/NebulousLabs/muxado/proto/ext.(*Heartbeat).request(0xc8201ee4e0)

	/home/travis/gopath/src/github.com/NebulousLabs/muxado/proto/ext/heartbeat.go:114 +0x3c0

created by github.com/NebulousLabs/muxado/proto/ext.(*Heartbeat).Start

	/home/travis/gopath/src/github.com/NebulousLabs/muxado/proto/ext/heartbeat.go:47 +0xce

FAIL	github.com/NebulousLabs/Sia/modules/gateway	300.020s
```

VoidingWarranties:
So _thats_ why I changed the `dialTimeout`. It's blocking in `Connect` on
```go
conn, err := net.DialTimeout("tcp", string(addr), dialTimeout)
```
I'll investigate, I think it's a bug in the test and changing the `dialTimeout` just hides it.

lukechampine:
Would be nice if `ThreadGroup` could track the line number of each `Add` and tell you which one hadn't returned.

lukechampine:
oh dammit

VoidingWarranties:
Ahh it's the `threadedPeerManager` trying to connect to the a dummy node. Ideally we wouldn't use dummy nodes, and we'd use real gateways instead, but it's a common pattern in testing so reducing `dialTimeout` is a better solution.

DavidVorick:
I'm being mean with this one, given that it seemed to introduce a lot of nondeterministic test failures initially. If the next set of tests passes I'll merge it.

PR #1180, opened 2016-05-24
Vendor twofish

lukechampine:
This is primarily to remove the rate limiting constraint imposed by `golang.org/x/crypto`. Vendoring means we no longer need to do the `gitcookies` dance to evade the rate limit, which was a problem for external pull requests.


VoidingWarranties:
Should we remove the Travis environment variable? I set it up with the Travis CLI, which now appears to be broken on my machine so I'm not entirely sure how to do that.

lukechampine:
done

PR #1186, opened 2016-05-25
sync.ThreadGroup

lukechampine:
Example usage:
```go
type module struct {
	threads sync.ThreadGroup
	log     *persist.Logger
}

func NewModule() module {
	return module{
		log: persist.NewLogger(),
	}
}

func (m *module) DoSomething() error {
	err := m.threads.Add(1)
	if err != nil {
		return errors.New("Module is closed")
	}
	defer m.threads.Done()

	for {
		select {
		case <-time.After(1 * time.Second):
			// do some work

		case <-m.threads.StopChan():
			return nil
		}
	}

	// alternatively:
	for {
		if m.threads.IsStopped() {
			return nil
		}
		// do some work
	}
}

func (m *module) Close() error {
	err := m.threads.Stop()
	if err != nil {
		return errors.New("Module already closed")
	}
	return m.log.Close()
}
```

### DIFF
--- a/sync/threadgroup.go
+++ b/sync/threadgroup.go
@@ -1,0 +1,67 @@
+package sync
+
+import (
+	"errors"
+	"sync"
+)
+
+// ErrStopped is returned by ThreadGroup methods if Stop has already been
+// called.
+var ErrStopped = errors.New("ThreadGroup already stopped")
+
+// ThreadGroup is a sync.WaitGroup with additional functionality for
+// facilitating clean shutdown. Namely, it provides a StopChan method for
+// notifying callers when shutdown occurrs. Another key difference is that a
+// ThreadGroup is only intended be used once; as such, its Add and Stop
+// methods return errors if Stop has already been called.
+type ThreadGroup struct {
+	stopChan chan struct{}
+	chanOnce sync.Once
+	wg       sync.WaitGroup
+}
+
+// StopChan provides read-only access to the ThreadGroup's stopChan. Callers
+// should select on StopChan in order to interrupt long-running reads (such as
+// time.After).
+func (tg *ThreadGroup) StopChan() <-chan struct{} {
+	// Initialize tg.stopChan if it is nil; this makes an unitialized
+	// ThreadGroup valid. (Otherwise, a NewThreadGroup function would be
+	// necessary.)
+	tg.chanOnce.Do(func() { tg.stopChan = make(chan struct{}) })
+	return tg.stopChan
+}
+
+// IsStopped returns true if Stop has been called.
+func (tg *ThreadGroup) IsStopped() bool {
+	select {
+	case <-tg.StopChan():
+		return true
+	default:
+		return false
+	}
+}
+
+// Add adds delta to the ThreadGroup counter.
+func (tg *ThreadGroup) Add(delta int) error {
+	if tg.IsStopped() {
+		return ErrStopped
+	}
+	tg.wg.Add(delta)
+	return nil
+}
+
+// Done decrements the ThreadGroup counter.
+func (tg *ThreadGroup) Done() {
+	tg.wg.Done()
+}
+
+// Stop closes the Threadgroup's stopChan and blocks until the counter is
+// zero.
+func (tg *ThreadGroup) Stop() error {
+	if tg.IsStopped() {
+		return ErrStopped
+	}
+	close(tg.stopChan)
+	tg.wg.Wait()
+	return nil
+}
--- a/sync/threadgroup.go
+++ b/sync/threadgroup.go
@@ -17,6 +17,7 @@ var ErrStopped = errors.New("ThreadGroup already stopped")
 type ThreadGroup struct {
 	stopChan chan struct{}
 	chanOnce sync.Once
+	mu       sync.Mutex
 	wg       sync.WaitGroup
 }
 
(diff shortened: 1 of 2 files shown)


VoidingWarranties:
Why wrap a `sync.WaitGroup` instead of using an atomic counter? You wouldn't need a mutex with the atomic counter.

Something like
```go
// ...

func (tg *ThreadGroup) Add() error {
        if tg.IsStopped() {
                return ErrStopped
        }
        atomic.AddInt32(&tg.counter, 1)
        return nil
}

func (tg *ThreadGroup) Done() {
        if atomic.LoadInt32(&tg.counter) == 0 {
                panic("negative ThreadGroup counter")
        }
        atomic.AddInt32(&tg.counter, -1)
        select {
        case tg.semaChan <- struct{}{}:
        default:
        }
}

func (tg *ThreadGroup) Stop() error {
        if tg.IsStopped() {
                return ErrStopped
        }
        close(tg.stopChan)
        tg.semaChan = make(chan struct{})
        for atomic.LoadInt32(&tg.counter) > 0 {
                <-tg.semaChan
        }
        return nil
}
```

lukechampine:
I'm not totally sure what you are suggesting, but the `semaChan` construction will certainly fail. Sends down a nil channel panic, so you can't initialize it only in `Stop`.
If you wanted to do away with the `WaitGroup`, the proper replacement would be a `sync.Cond`, woken up by `Add` and `Done`.

Mostly I like using a `WaitGroup` because it makes it clear that we are extending that base idea, instead of making something completely new.

VoidingWarranties:
The channel send is in a select with a default, so it will not panic.

lukechampine:
Whoops, you are correct, I was thinking of a send to a closed channel.

VoidingWarranties:
Aw you're right. And I thought I was being clever, oh well.

lukechampine:
FWIW, when I added a mutex around your `semaChan`, it did pass my tests. But at that point we're just reimplementing a `WaitGroup` for no gain.

VoidingWarranties:
The `WaitGroup` has a lot of unneeded complexity for our simpler use case. I doubt the speedup would be significant enough to be worth it though. And it might actually make it slower since the `WaitGroup` implementation uses an unexported semaphore from the runtime package.

lukechampine:
That's fair; `WaitGroup` guards against improper reuse and concurrent `Wait`s, neither of which apply to our use case. Still, I like the clarity of the current design: a `WaitGroup` for waiting plus a `stopChan` for signalling, and the rest is just implementation details.

sync/threadgroup.go
```diff
@@ -0,0 +1,73 @@
...
+	if tg.IsStopped() {
+		return ErrStopped
+	}
+	tg.wg.Add(delta)
+	return nil
+}
+
+// Done decrements the ThreadGroup counter.
+func (tg *ThreadGroup) Done() {
+	tg.wg.Done()
```
DavidVorick: This will panic if you go negative?

sync/threadgroup_test.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	// try to trigger the race detector
+	tg = new(ThreadGroup)
+	go tg.IsStopped()
+	go tg.Add(1)
+	err := tg.Stop()
+	if err != nil {
+		t.Fatal(err)
+	}
```
DavidVorick: This last little bit probably belongs in a new test.

sync/threadgroup.go
lukechampine: yes. We can convert it to a `build.Critical` if you want. Any mismatch between `Add` and `Done` is definitely a developer mistake.

DavidVorick: I'm split. `build.Critical` would be good, but I don't think an outright panic is that bad either.

VoidingWarranties: `Add` can also panic if the delta is negative, so if you convert it to a `built.Critical` here convert it in `Add` as well!

sync/threadgroup.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+}
+
+// Add adds delta to the ThreadGroup counter.
+func (tg *ThreadGroup) Add(delta int) error {
+	tg.mu.Lock()
+	defer tg.mu.Unlock()
+	if tg.IsStopped() {
+		return ErrStopped
+	}
```
VoidingWarranties: It's valid to have a negative delta, in which case you wouldn't want the lock or the `IsStopped` check. What about only locking and checking `IsStopped` if the delta is positive?

lukechampine: Negative deltas are valid for a `WaitGroup`, but I don't see why we would want (or need) them for a `ThreadGroup`. A negative delta should just result in an error.

lukechampine: Alternatively, we could remove the `delta` argument entirely, as you have proposed. I would be okay with that.

VoidingWarranties: I don't see why you would ever want to use a negative delta (other than -1) with a `WaitGroup` either, but I haven't used `WaitGroup`s very often so maybe I just haven't come across the corresponding pattern.

Like you said our use case is rather specific so I'm okay with removing the `delta` argument until we come across a case that necessitates it.

DavidVorick: same here, fine to remove the delta

DavidVorick:
Out of date with master, otherwise ACK.

Waiting on ack from @VoidingWarranties, then I'll merge

VoidingWarranties:
ACK. I'll use this in the gateway's clean shutdown as soon as this is merged.

PR #1188, opened 2016-05-26
Boltdb TestOpenDatabase function

mnsl:
Implements new TestOpenDatabase function that tests the OpenDatabase function and BoltDatabase Close method in package persist.  Checks cases where database is empty, full, and has not yet been created.

### DIFF
--- a/persist/boltdb_test.go
+++ b/persist/boltdb_test.go
@@ -1,0 +1,167 @@
+package persist
+
+import (
+	"math/rand"
+	"os"
+	"testing"
+
+	"github.com/NebulousLabs/Sia/build"
+	"github.com/NebulousLabs/bolt"
+)
+
+const numTestInputs = 25
+
+var dbVersions = []string{"0.0.0", "7.0.4", "asdf"}
+
+type testInput struct {
+	dbMetadata Metadata
+	dbFilename string
+}
+
+var testInputs = make([]testInput, numTestInputs)
+
+var testBuckets = [][]byte{
+	[]byte("FakeBucket"),
+	[]byte("FakeBucket123"),
+	[]byte("FakeBucket123!@#$"),
+	[]byte("Another Fake Bucket"),
+	[]byte("FakeBucket" + RandomSuffix()),
+	[]byte("_"),
+	[]byte(" asdf"),
+}
+
+// TestOpenDatabase tests calling OpenDatabase on the following types of
+// databases:
+// - a database that has not yet been created
+// - an existing empty database
+// - an existing nonempty database
+// Along the way, it also tests calling Close on:
+// - a newly-created database
+// - a newly-filled database
+// - a newly-emptied database
+func TestOpenDatabase(t *testing.T) {
+	// Create a folder for the database file. If a folder by that name exists
+	// already, it will be replaced by an empty folder.
+	testdir := build.TempDir(persistDir, "TestOpenNewDatabase")
+	err := os.MkdirAll(testdir, 0700)
+	if err != nil {
+		t.Fatal(err)
+	}
+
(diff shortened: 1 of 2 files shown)


persist/boltdb_test.go
```diff
@@ -0,0 +1,167 @@
...
+
+var testBuckets = [][]byte{
+	[]byte("FakeBucket"),
+	[]byte("FakeBucket123"),
+	[]byte("FakeBucket123!@#$"),
+	[]byte("Another Fake Bucket"),
+	[]byte("FakeBucket" + RandomSuffix()),
+	[]byte("_"),
+	[]byte(" asdf"),
+}
```
DavidVorick: Since these variables and costs are only used inside of one test, it's better to declare them inside that test, especially since they don't really have broader application.

persist/boltdb_test.go
```diff
@@ -0,0 +1,167 @@
...
+func TestOpenDatabase(t *testing.T) {
+	// Create a folder for the database file. If a folder by that name exists
+	// already, it will be replaced by an empty folder.
+	testdir := build.TempDir(persistDir, "TestOpenNewDatabase")
+	err := os.MkdirAll(testdir, 0700)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Generate ten random testInputs for building database files.
```
DavidVorick: better not to use a word like 'ten', because sometimes you change the variable but forget the constant :) (same problem you fixed above - probably better to remove the number altogether here too)

persist/boltdb_test.go
```diff
@@ -0,0 +1,167 @@
...
+		testInputs[i] = in
+	}
+
+	// Run the same tests as in TestOpenDatabase for each testInput
+	for _, in := range testInputs {
+		// Create a new database.
+		db, err := OpenDatabase(in.dbMetadata, in.dbFilename)
+		if err != nil {
+			t.Errorf("calling OpenDatabase on a new database failed for input %v (error below)", in)
+			t.Fatal(err)
```
lukechampine: why split this into two calls? `t.Fatalf` should be fine

persist/boltdb_test.go
```diff
@@ -0,0 +1,167 @@
...
+		// Empty every bucket in the database.
+		err = db.Update(func(tx *bolt.Tx) error {
+			for _, testBucket := range testBuckets {
+				b := tx.Bucket(testBucket)
+				err := b.ForEach(func(k, v []byte) error {
+					err := b.Delete(k)
+					if err != nil {
+						return err
+					}
+					return nil
```
lukechampine: simplify to `return b.Delete(k)`

persist/boltdb_test.go
lukechampine: Also, note that you can inline the struct definition:
```go
testInputs := make([]struct{
    meta Metadata
    filename string
}, numTestInputs)
```

persist/boltdb_test.go
```diff
@@ -0,0 +1,167 @@
...
+	// Create a folder for the database file. If a folder by that name exists
+	// already, it will be replaced by an empty folder.
+	testdir := build.TempDir(persistDir, "TestOpenNewDatabase")
+	err := os.MkdirAll(testdir, 0700)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Generate ten random testInputs for building database files.
+	for i := 0; i < numTestInputs; i++ {
```
lukechampine: when assigning to elements of a slice, use `for i := range slice`

mnsl:
Addresses code review comments, except for Luke's note about the weird table driven test format. The randomized testInputs will be replaced with hardcoded ones in a future pull request in order to reimplement via anonymous struct.

PR #1196, opened 2016-05-27
Host clean shutdown

DavidVorick:
First commit does a brief cleanup, the second commit switches from the resourceLock to a ThreadGroup.

### DIFF
--- a/modules/host/host.go
+++ b/modules/host/host.go
@@ -33,9 +33,6 @@ package host
 // TODO: merge the network interfaces stuff, don't forget to include the
 // 'announced' variable as one of the outputs.
 
-// TODO: check that the host is doing proper clean shudown, especially
-// network.go, a couple of problems with clean shutdown in network.go.
-
 // TODO: 'announced' doesn't tell you if the announcement made it to the
 // blockchain.
 
@@ -80,6 +77,7 @@ import (
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/modules/host/storagemanager"
 	"github.com/NebulousLabs/Sia/persist"
+	siasync "github.com/NebulousLabs/Sia/sync"
 	"github.com/NebulousLabs/Sia/types"
 )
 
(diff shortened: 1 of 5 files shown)


modules/host/host.go
```diff
@@ -245,13 +235,15 @@ func newHost(dependencies dependencies, cs modules.ConsensusSet, tpool modules.T
 		persistDir: persistDir,
 	}
 
+	// Call stop in the event of a partial startup.
```
lukechampine: interesting. I think in an ideal startup, you wouldn't spawn any goroutines until all potential failures had been accounted for. But that isn't always possible with a complex object.
Maybe you should go a step further and write a `h.tearDown()` helper? This would close the log and db as well, but only if they were non-nil.

lukechampine:
I think `cs.Unsubscribe` should happen before `tg.Stop`, for the same reasons as `Accept`.

In my view, the proper order of `Close` is:
1. Prevent new threads from being spawned
2. Stop currently running threads
3. Close resources that threads were accessing

It's a nice convention for `tg.Stop` to always be first, but it doesn't correspond to how I believe shutdown should happen. However, I admit that there is probably no difference in real-world behavior.

modules/host/host.go
DavidVorick: perhaps in another PR. I'm not confident about the best way to make sure that startup is clean.

DavidVorick:
Cleanly unsubscribing before calling stop would require calling `Add` and `Done` wrapping the unsubscribe call. Subscription isn't another thread, it's another object holding the host and occasionally calling functions. Unsubscribe doesn't really fall into any of the 3 bullet points that you mentioned.  Perhaps the consensus set could be considered a resource, but it's not owned by any thread that belongs to the host.

lukechampine:
>Cleanly unsubscribing before calling stop would require calling `Add` and `Done` wrapping the unsubscribe call.

I don't see why. But it is rather late, so maybe I'm missing something.

Unsubscribing falls into point 1: it prevents new calls to `ProcessConsensusChange` during shutdown. It's true of course that these new calls will fail immediately (because `Add` will return an error), but why not prevent them from happening in the first place?

DavidVorick:
It is to account for multiple calls to `Close`. Calling `Unsubscribe` twice on the consensus set is undefined behavior. I don't think it would cause problems, but in the spirit of best practice we should avoid having any undefined behavior in the codebase.

Issue #1102, opened 2016-04-26
Nondeterministic build failure in TestMultiSectorStorageObligationStack


DavidVorick:
```
--- FAIL: TestMultiSectorStorageObligationStack (0.73s)

	storageobligations_smoke_test.go:541: storage obligation is not saying that the storage proof was confirmed on the blockchain

--- PASS: TestStorageObligationID (0.00s)
```

DavidVorick:
also fixed

PR #1197, opened 2016-05-27
add extra logging to the host submitting storage proofs

DavidVorick:
adds information that will potentially be helpful in #1102 

@VoidingWarranties reported that this failure was occurring frequently for him. I can't reproduce it frequently, but the one time I was able to trigger it I didn't have enough logging information to get to the bottom. For me, it happens maybe 1-in-50 tries. With this debugging information, I should be able to get farther next time I see the failure.


DavidVorick:
*this debugging information is good to have regardless of the specific bug I am targeting, and should remain in the codebase even after the bug is resolved.

PR #1199, opened 2016-05-27
More persist/boltdb.go testing

mnsl:
Fixes up TestOpenDatabase:
- there is now a broader variety of testInputs and testBuckets
- testInputs are now hardcoded rather than randomly generated
- testInputs can thus now be defined via anonymous struct, so type testInput can be removed
- filepath argument to OpenDatabase is now the full filepath of the database (previously just the file name, dbFilename, was passed in, which meant that the database was created in the wrong directory)

Introduces TestBadFileModeOpenDatabase, which checks that a file with the wrong filemode (less than 0600) will cause OpenDatabase to return a permissions error.

### DIFF
--- a/persist/boltdb_test.go
+++ b/persist/boltdb_test.go
@@ -258,10 +258,7 @@ func TestErrTxNotWritable(t *testing.T) {
 		}
 		// Should return an error because updateMetadata is being called from
 		// a read-only transaction.
-		err = db.View(func(tx *bolt.Tx) error {
-			err = boltDB.updateMetadata(tx)
-			return err
-		})
+		err = db.View(boltDB.updateMetadata)
 		if err != bolt.ErrTxNotWritable {
 			t.Errorf("updateMetadata returned wrong error for input %v, filename %v; expected tx not writable, got %v", in.md, dbFilename, err)
 		}
--- a/persist/boltdb_test.go
+++ b/persist/boltdb_test.go
@@ -1,5 +1,3 @@
-// Borrowed weird strings from https://github.com/minimaxir/big-list-of-naughty-strings
-
 package persist
 
 import (
@@ -12,7 +10,11 @@ import (
 	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/bolt"
 )
-
+ 
+// testInputs and testFilenames are global variables because most tests require
+// a variety of metadata and filename inputs (although only TestCheckMetadata 
+// and TestIntegratedCheckMetadata use testInput.newMd and testInput.err). 
+// Weird strings are from https://github.com/minimaxir/big-ltist-of-naughty-strings
 var (
     testInputs = []struct{
         md		Metadata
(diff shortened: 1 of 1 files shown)


mnsl:
New changes:
* Rename TestBadFilemodeOpenDatabase to TestErrPermissionOpenDatabase.

* Introduce several new functions:
  * `TestErrTxNotWritable` (tests that updateMetadata returns an error when called from a read-only transaction)
  * `TestErrDatabaseNotOpen (tests that checkMetadata returns an error when called on a closed BoltDatabase)
  * `TestErrCheckMetadata` (tests that checkMetadata(md) returns current error when called on a database whose metadata does not match md)
  * `TestErrIntegratedCheckMetadata` (tests that checkMetadata returns correct errors when called within OpenDatabase on a database whose metadata does not match md)

* Make global testInputs and testFilenames structs, since tests require the same sorts of inputs

persist/boltdb_test.go
```diff
@@ -28,72 +76,71 @@ func TestOpenDatabase(t *testing.T) {
...
 				if err != nil {
+					t.Fatalf("db.Update failed on bucket name %v for metadata %v, filename %v; error was %v", testBucket, in.md, dbFilename, err)
 					return err
 				}
 			}
 			return nil
 		})
 		if err != nil {
-			t.Fatal(err)
+		}
```
lukechampine: error not handled?

persist/boltdb_test.go
```diff
@@ -149,14 +196,256 @@ func TestOpenDatabase(t *testing.T) {
...
+			t.Fatal(err)
+		}
+
+		boltDB := &BoltDatabase{
+			Metadata: in.md,
+			DB:       db,
+		}
+
+		tx, err := db.Begin(false)
+		// Should return an error since tx is a read-only transaction.
```
lukechampine: use `db.View` for the read-only equivalent of `db.Update`

persist/boltdb_test.go
mnsl: Weird, sorry about that.

mnsl: Although then the database won't be closed properly.  Perhaps t.Fatal(err) is best.

lukechampine: A simple `t.Fatal(err)` is fine here. In practice, we don't really care about cleanup in the event of test failure, since test data is stored in `/tmp` and gets wiped on each run anyway.
Similarly, using `t.Error` instead of `t.Fatal` is indeed better practice in some cases, but most of the time it's fine to be lazy and use `t.Fatal`. In either case, we're going to fix the problem; it's like the distinction between your compiler emitting 10 errors or only the first one encountered.

mnsl:
Latest changes address Luke's comments:
* Fix error handling in OpenDatabase
* Use View() instead of Begin() and Commit() in TestErrTxNotWritable

persist/boltdb_test.go
```diff
@@ -28,74 +76,78 @@ func TestOpenDatabase(t *testing.T) {
...
 		}
 
+		// Make sure CreateBucketIfNotExists method handles invalid (nil)
+		// bucket name.
+		err = db.Update(func(tx *bolt.Tx) error {
+			_, err := tx.CreateBucketIfNotExists(nil)
+			return err
+		})
+
+		//
```
DavidVorick: blank comment

persist/boltdb_test.go
```diff
@@ -149,14 +203,261 @@ func TestOpenDatabase(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		boltDB := &BoltDatabase{
+			Metadata: in.md,
+			DB:       db,
+		}
+
+		err = db.View(func(tx *bolt.Tx) error {
```
DavidVorick: Generally better to favor fewer newlines in your code - more compact code means more fits on the screen, and if you are clumping ideas together well readability will improve.

I would generally pull out the newline here, and a couple of the others in this file.

mnsl:
Latest update removes extraneous new lines and that empty comment David spotted

persist/boltdb_test.go
```diff
@@ -145,18 +180,227 @@ func TestOpenDatabase(t *testing.T) {
...
+		boltDB := &BoltDatabase{
+			Metadata: in.md,
+			DB:       db,
+		}
+		// Should return an error because updateMetadata is being called from
+		// a read-only transaction.
+		err = db.View(func(tx *bolt.Tx) error {
+			err = boltDB.updateMetadata(tx)
+			return err
+		})
```
VoidingWarranties: Could this could be
```go
err = db.View(boltDB.updateMetadata)
```

mnsl: Good catch, I don't see why not

mnsl:
db.View now takes updateMetadata directly as an argument

PR #1201, opened 2016-05-29
Better 'siac host -v' output

DavidVorick:


siac/hostcmd.go
```diff
@@ -152,35 +152,50 @@ func hostcmd() {
...
-		if is.NetAddress == "" {
-			netaddr += " (automatically determined)"
-		} else {
-			netaddr += " (manually specified)"
-		}
 		fmt.Printf(`
-	Net Address: %v
+Host Settings:
+	AcceptingContracts:  %v
+	MaxDuration:          %v
```
lukechampine: misaligned

lukechampine: Is there a reason for not adding spaces to these settings? I assume it's so that you can easily map them to the corresponding API parameter.The capitalization is wrong anyway, though. Maybe we should lowercase the argument to `siac host config`.

siac/hostcmd.go
```diff
@@ -152,35 +152,50 @@ func hostcmd() {
...
-		currencyUnits(fm.LockedStorageCollateral), currencyUnits(fm.RiskedStorageCollateral),
-		currencyUnits(totalRevenue), currencyUnits(fm.LostRevenue),
-		currencyUnits(fm.LostStorageCollateral))
+	// determine the display method for the net address.
+	netaddr := es.NetAddress
+	if is.NetAddress == "" {
+		netaddr += " (automatically determined)"
+	} else {
+		netaddr += " (manually specified)"
+	}
```
VoidingWarranties: Why move this outside the conditional if you aren't displaying it non-verbose output?

DavidVorick: At some point I was displaying it in the non-verbose output, but I felt like it would make the future lives of programmers easier if this code were outside.

siac/hostcmd.go
DavidVorick: I wasn't sure the best convention to follow, so I just mimicked the struct I was looking at.

You are right though, either they should be english words or they should be the json fields. My preference would be the json fields.

DavidVorick:
2016-05-31-135819_663x824_scrot

DavidVorick:
oh, seems like calling `currencyUnits` is not enough, really need to do the type conversions as well on the bandwidth, collateral, and storage.

siac/hostcmd.go
```diff
@@ -152,35 +152,49 @@ func hostcmd() {
...
 		// describe net address
-		netaddr := es.NetAddress
-		if is.NetAddress == "" {
-			netaddr += " (automatically determined)"
-		} else {
-			netaddr += " (manually specified)"
-		}
 		fmt.Printf(`
-	Net Address: %v
+Host Internal Settings:
```
VoidingWarranties: There's an extra newline before `Host Internal Settings:`

DavidVorick:
attempt number 2:

2016-05-31-145947_653x849_scrot

VoidingWarranties:
MaxDuration in `siac host` has blocks as units, but maxduration in `siac host -v` has weeks as units (but it doesn't display the units to the user).

DavidVorick:
2016-05-31-193809_584x826_scrot

fixed

PR #1202, opened 2016-05-30
Fix deadlock in host shutdown

VoidingWarranties:
The first commit in this PR fixes the deadlock. The rest of the commits are suggestions. I ask that you evaluate them individually.

The deadlock was possible because it was possible for the listener to start before the host is closed, but for the closer-goroutine to start after the host is closed. This is fixed by calling `Add` outside of the goroutine. The exact same deadlock existed in the gateway in #1178. `Add` should always be called outside of the goroutine when starting a closer-goroutine.

### DIFF
--- a/modules/host/network.go
+++ b/modules/host/network.go
@@ -98,9 +98,17 @@ func (h *Host) initNetworking(address string) (err error) {
 // threadedHandleConn handles an incoming connection to the host, typically an
 // RPC.
 func (h *Host) threadedHandleConn(conn net.Conn) {
-	// We defer the conn.Close so that the conn is still closed if `Add` returns
-	// an error.
-	defer conn.Close()
+	// Close the conn on host.Close or when the method terminates, whichever comes
+	// first.
+	connCloseChan := make(chan struct{})
+	defer close(connCloseChan)
+	go func() {
+		select {
+		case <-h.tg.StopChan():
+		case <-connCloseChan:
+		}
+		conn.Close()
+	}()
 
 	err := h.tg.Add()
 	if err != nil {
--- a/modules/host/network.go
+++ b/modules/host/network.go
@@ -49,14 +49,14 @@ func (h *Host) initNetworking(address string) (err error) {
 		return err
 	}
 	// Automatically close the listener when h.tg.Stop() is called.
+	err = h.tg.Add()
+	if err != nil {
+		return err
+	}
 	go func() {
-		err := h.tg.Add()
-		if err != nil {
-			return
-		}
 		defer h.tg.Done()
 		<-h.tg.StopChan()
-		err = h.listener.Close()
+		err := h.listener.Close()
 		if err != nil {
 			h.log.Println("Closing the listener failed:", err)
 		}
--- a/modules/host/network.go
+++ b/modules/host/network.go
@@ -98,6 +98,10 @@ func (h *Host) initNetworking(address string) (err error) {
 // threadedHandleConn handles an incoming connection to the host, typically an
 // RPC.
 func (h *Host) threadedHandleConn(conn net.Conn) {
+	// We defer the conn.Close so that the conn is still closed if `Add` returns
+	// an error.
+	defer conn.Close()
+
 	err := h.tg.Add()
 	if err != nil {
 		return
@@ -111,7 +115,6 @@ func (h *Host) threadedHandleConn(conn net.Conn) {
 		h.log.Println("WARN: could not set deadline on connection:", err)
 		return
 	}
-	defer conn.Close()
 
 	// Read a specifier indicating which action is being called.
 	var id types.Specifier
--- a/modules/host/network.go
+++ b/modules/host/network.go
@@ -191,7 +191,6 @@ func (h *Host) threadedListen() {
 			return
 		}
 
-		// Grab the resource lock before creating a goroutine.
 		go h.threadedHandleConn(conn)
 	}
 }
--- a/modules/host/network.go
+++ b/modules/host/network.go
@@ -49,18 +49,12 @@ func (h *Host) initNetworking(address string) (err error) {
 		return err
 	}
 	// Automatically close the listener when h.tg.Stop() is called.
-	err = h.tg.Add()
-	if err != nil {
-		return err
-	}
-	go func() {
-		defer h.tg.Done()
-		<-h.tg.StopChan()
+	h.tg.OnStop(func() {
 		err := h.listener.Close()
 		if err != nil {
-			h.log.Println("Closing the listener failed:", err)
+			h.log.Println("WARN: closing the listener failed:", err)
 		}
-	}()
+	})
 
 	// Set the port.
 	_, port, err := net.SplitHostPort(h.listener.Addr().String())
--- a/modules/host/network.go
+++ b/modules/host/network.go
@@ -18,14 +18,14 @@ var rpcSettingsDeprecated = types.Specifier{'S', 'e', 't', 't', 'i', 'n', 'g', '
 // checks if the host's hostname has changed, and makes an updated host
 // announcement if so.
 func (h *Host) threadedUpdateHostname() {
-	for {
-		err := h.tg.Add()
-		if err != nil {
-			return
-		}
-		h.managedLearnHostname()
-		h.tg.Done()
+	err := h.tg.Add()
+	if err != nil {
+		return
+	}
+	defer h.tg.Done()
 
+	for {
+		h.managedLearnHostname()
 		// Wait 30 minutes to check again. If the hostname is changing
 		// regularly (more than once a week), we want the host to be able to be
 		// seen as having 95% uptime. Every minute that the announcement is


DavidVorick:
I know you just made a bunch of PRs, but if you want stuff to be evaluated individually you should make a separate PR for each.

It's okay to have a bunch of PRs.

modules/host/network.go
```diff
@@ -49,11 +49,11 @@ func (h *Host) initNetworking(address string) (err error) {
...
 	// Automatically close the listener when h.tg.Stop() is called.
+	err = h.tg.Add()
+	if err != nil {
+		return err
+	}
 	go func() {
-		err := h.tg.Add()
-		if err != nil {
-			return
-		}
```
DavidVorick: Are you sure that this causes a deadlock?

1. This thread is spawned, but put in the queue, `Add()` is not called yet.
2. Host closes, full clean shutdown occurs
3. This thread is reached. `Add()` returns an error, thread closes.

No deadlock, as far as I can tell. This goroutine will not shut down until after full clean shutdown, but there's no deadlock.

DavidVorick: fwiw I do agree with the change, since it means the goroutine is killed properly.

modules/host/network.go
```diff
@@ -174,6 +161,20 @@ func (h *Host) threadedListen() {
...
+	}
+	go func() {
+		defer h.tg.Done()
+		<-h.tg.StopChan()
+		err = h.listener.Close()
+		if err != nil {
+			h.log.Println("Closing the listener failed:", err)
+		}
+	}()
+
```
DavidVorick: This thread is not the thread that creates the listener, but this piece of code is the code that creates the listener, therefore this code definitely doesn't make sense here. It also brings back the non-clean shutdown bug that was fixed in the first commit.

modules/host/network.go
```diff
@@ -18,14 +18,14 @@ var rpcSettingsDeprecated = types.Specifier{'S', 'e', 't', 't', 'i', 'n', 'g', '
...
+	defer h.tg.Done()
+
 	for {
-		err := h.tg.Add()
-		if err != nil {
-			return
-		}
 		h.managedLearnHostname()
-		h.tg.Done()
-
```
DavidVorick: hmm. Originally I grouped it this was so that we would not have a long running hold on `tg`. As it happens now, the thread will hold the `tg` indefinitely, something that seemed not great. The way it currently is does not adhere to maximum clean shutdown, but it does mean that we lose a long running hold on `tg`.

modules/host/network.go
```diff
@@ -85,9 +85,18 @@ func (h *Host) initNetworking(address string) (err error) {
...
+		// We defer the conn.Close so that the conn is still closed if `Add` returns
+		// an error.
+		defer conn.Close()
+		err := h.tg.Add()
+		if err != nil {
+			return
+		}
+		defer h.tg.Done()
+		<-h.tg.StopChan()
+	}()
```
DavidVorick: I'm not sure that this code appropriately addresses the problem. We still seem to have an issue where it could take up to X minutes to shutdown?

modules/host/network.go
VoidingWarranties: The deadlock is less of a deadlock and more of a "the listener never closes and therefore `Close` never exits".

VoidingWarranties: Host tests were timing out non-deterministically on Close. I believe this was the cause. It might not be a deadlock in the original code (ungrouped `listener.Close` and `listener.Accept`) if the execution order of goroutines is guaranteed to be the same as the calling order, but I'm not sure this is true.

modules/host/network.go
VoidingWarranties: This does not re-introduce the bug as far as I can tell. Can you explain your reasoning more?

This is in contrast to the paradigm that every parent should be responsible for killing it's children. This paradigm works well when the children never outlive the parent, but that is not true here. When the children outlive the parent I think a better paradigm is for the thread to be responsible for killing itself.

modules/host/network.go
VoidingWarranties: `conn.Read` and `conn.Write` will return immediately when the conn is closed.

modules/host/network.go
VoidingWarranties: The effective change this commit introduces is very tiny. The thread will still exit immediately on `Close` because the thread selects on the `StopChan`. I do not think you could write a test to tell the two apart.

modules/host/network.go
lukechampine: why is this function called with `defer` instead of `go`?
This is a tricky problem, because you don't want the goroutine to linger if the connection closes gracefully. The easiest approach I can see is:
```go
go func() {
    <-h.tg.StopChan()
    conn.Close()
}()
```
This takes advantage of the fact that `net.Conn.Close` can be safely called twice.

VoidingWarranties: Oops, definitely meant it to be `go` not `defer`. Why remove the `Add` and `Done` from the goroutine?

I didn't know that a `net.Conn` can be safely closed twice. Regardless it's still possible to only close it once.
```go
connCloseChan := make(chan struct{})
defer func() { close(connCloseChan) }()
go func() {
    defer conn.Close()
    err := h.tg.Add()
    if err != nil {
        return
    }
    defer h.tg.Done()
    select {
    case <-h.tg.StopChan():
    case <-connCloseChan:
    }
}
```

lukechampine: I modified `ThreadGroup` to address this issue. Take a look: https://github.com/NebulousLabs/Sia/pull/1205

lukechampine: hmm... actually, this doesn't perfectly fit the use case. It works great for `net.Listener`s, but not so much for `net.Conn`s. If you naively register every conn, then the ThreadGroup will wind up storing a reference to every conn you've ever made. Definitely bad.

VoidingWarranties:
```go
defer h.tg.UnregisterCloser(conn)
```

VoidingWarranties: The same problem exists for the naive goroutine too. I'm not sure which would tax the system more but neither is good.

lukechampine: as for the `Add` and `Done`, they aren't strictly necessary here. As discussed, `Close` can be called multiple times without issue, so the only real risk is the channel read blocking forever. But since `tg.Stop` closes the channel, it is guaranteed to return with `tg.Stop`.

VoidingWarranties: Does something like this look good to you?
```go
connCloseChan := make(chan struct{})
defer func() { close(connCloseChan) }()
go func() {
    select {
    case <-h.tg.StopChan():
    case <-connCloseChan:
    }
    conn.Close()
}()
```

lukechampine: yeah, that looks good.

modules/host/network.go
DavidVorick: Goroutines are not guaranteed to be run in the same order that they are created.

This still wouldn't cause a timeout during shutdown though. Either this thread is has called `Add` which means the `Close` thread is going to block, or this thread has not yet called `Add` which means the `Close` thread is not going to block. What you've done is guarantee that `Add` has been called, which can only increase the length of time that it takes for shutdown to occur.

The host tests are timing out during the short tests or the long tests? If there is a deadlock in the host, it's not here.

In the parent PR, can you provide some of the dumps associated with the timeouts? It might just be the short vs. long test thing again.

modules/host/network.go
DavidVorick: 1. The listener is started.
2. This thread is spawned, but no code is executed.
3. Close is called, the thread does not block because `Add` has not been called yet.

DavidVorick: I was thinking you could move to starting up the listener into this thread, but really you want to make sure that the `host` object is not returned until the listener is running. Currently the listener is spawned in a thread that will block the return of the host object, and that's the way it should stay.

This code definitely belongs next to the listener, because the job of this code is to close the listener, not to close `threadedListen`. `threadedListen` is a separate entity that chose to close based on the uptime of the listener. If you don't like that then you should be making it's shutdown more explicitly depend on the `StopChan` instead of on the death of the listener, but I don't think that's necessary.

NACK on this code change.

modules/host/network.go
DavidVorick: no you probably could not. Depends on how many goroutines are running I would think.

DavidVorick: It was more significant when the host was using resource locks instead of thread groups, because there was no shutdown chan meaning that the lock could be held for up to thirty minutes in a row. That was definitely important to fix, but the `stopChan` makes it okay for the lock grab to be long-running again.

modules/host/network.go
```diff
@@ -98,6 +85,18 @@ func (h *Host) initNetworking(address string) (err error) {
 // threadedHandleConn handles an incoming connection to the host, typically an
 // RPC.
 func (h *Host) threadedHandleConn(conn net.Conn) {
+	// Close the conn on host.Close or when the method terminates, whichever comes
+	// first.
+	connCloseChan := make(chan struct{})
+	defer func() { close(connCloseChan) }()
```
DavidVorick: you can't just `defer close(connCloseChan)`?

DavidVorick:
This PR looks good, except for separating the code that closes the listener from the code that creates the listener. Those definitely need to be next to eachother.

modules/host/network.go
VoidingWarranties: Oops, good catch

modules/host/network.go
VoidingWarranties: It's not the `Close` thread that causes the timeout, it's the `threadedHandleConn` thread that timesout. I wrote a test to demonstrate.
```go
func TestHostDeadlockBug(t *testing.T) {
        if testing.Short() {
                t.SkipNow()
        }

        ht, err := blankHostTester("TestHostDeadlockBug")
        if err != nil {
                t.Fatal(err)
        }
        err = ht.host.Close()
        if err != nil {
                t.Fatal(err)
        }

        for i := 0; i < 10000; i++ {
                h, _ := New(ht.cs, ht.tpool, ht.wallet, "localhost:0", filepath.Join(build.SiaTestingDir, "TestHostDeadlockBug", modules.HostDir))
                h.Close()
        }
}
```
And here's the test output
```
$ make test-long pkgs=./modules/host run=TestHostDeadlockBug
rm -rf release doc/whitepaper.aux doc/whitepaper.log doc/whitepaper.pdf
gofmt -s -l -w ./modules/host
go install ./modules/host
go vet ./modules/host
go test -v -race -tags='testing debug' -timeout=300s ./modules/host -run=TestHostDeadlockBug
=== RUN   TestHostDeadlockBug
panic: test timed out after 5m0s

goroutine 2035 [running]:
testing.startAlarm.func1()
        /usr/local/Cellar/go/1.5.3/libexec/src/testing/testing.go:703 +0x16b
created by time.goFunc
        /usr/local/Cellar/go/1.5.3/libexec/src/time/sleep.go:129 +0x6e

goroutine 1 [chan receive, 4 minutes]:
testing.RunTests(0x8a5eb0, 0xb4e820, 0x15, 0x15, 0x1)
        /usr/local/Cellar/go/1.5.3/libexec/src/testing/testing.go:562 +0xafa
testing.(*M).Run(0xc82003ff10, 0xc82014c770)
        /usr/local/Cellar/go/1.5.3/libexec/src/testing/testing.go:494 +0xe5
main.main()
        github.com/NebulousLabs/Sia/modules/host/_test/_testmain.go:94 +0x210

goroutine 17 [syscall, 4 minutes, locked to thread]:
runtime.goexit()
        /usr/local/Cellar/go/1.5.3/libexec/src/runtime/asm_amd64.s:1721 +0x1

goroutine 6 [semacquire, 4 minutes]:
sync.runtime_Semacquire(0xc8200d4ed0)
        /usr/local/Cellar/go/1.5.3/libexec/src/runtime/sema.go:43 +0x26
sync.(*WaitGroup).Wait(0xc8200d4ec4)
        /usr/local/Cellar/go/1.5.3/libexec/src/sync/waitgroup.go:126 +0x118
github.com/NebulousLabs/Sia/sync.(*ThreadGroup).Stop(0xc8200d4ea8, 0x0, 0x0)
        /Users/jordan/golang/src/github.com/NebulousLabs/Sia/sync/threadgroup.go:71 +0x10a
github.com/NebulousLabs/Sia/modules/host.(*Host).Close(0xc8200d4a00, 0x0, 0x0)
        /Users/jordan/golang/src/github.com/NebulousLabs/Sia/modules/host/host.go:310 +0x7a
github.com/NebulousLabs/Sia/modules/host.TestHostDeadlockBug(0xc82008e1b0)
        /Users/jordan/golang/src/github.com/NebulousLabs/Sia/modules/host/host_test.go:119 +0x4dc
testing.tRunner(0xc82008e1b0, 0xb4e8e0)
        /usr/local/Cellar/go/1.5.3/libexec/src/testing/testing.go:456 +0xdd
created by testing.RunTests
        /usr/local/Cellar/go/1.5.3/libexec/src/testing/testing.go:561 +0xaa4

goroutine 7 [sleep]:
time.Sleep(0x4a817c800)
        /usr/local/Cellar/go/1.5.3/libexec/src/runtime/time.go:59 +0xf9
github.com/NebulousLabs/Sia/sync.(*RWMutex).threadedDeadlockFinder(0xc82000bf80)
        /Users/jordan/golang/src/github.com/NebulousLabs/Sia/sync/lock.go:78 +0x728
created by github.com/NebulousLabs/Sia/sync.New
        /Users/jordan/golang/src/github.com/NebulousLabs/Sia/sync/lock.go:47 +0x127

goroutine 10 [select]:
github.com/NebulousLabs/Sia/modules/gateway.(*Gateway).threadedPeerManager(0xc8200f5030)
        /Users/jordan/golang/src/github.com/NebulousLabs/Sia/modules/gateway/peers.go:320 +0x429
created by github.com/NebulousLabs/Sia/modules/gateway.New
        /Users/jordan/golang/src/github.com/NebulousLabs/Sia/modules/gateway/gateway.go:169 +0xc36

goroutine 11 [select]:
github.com/NebulousLabs/Sia/modules/gateway.(*Gateway).threadedNodeManager(0xc8200f5030)
        /Users/jordan/golang/src/github.com/NebulousLabs/Sia/modules/gateway/nodes.go:135 +0x653
created by github.com/NebulousLabs/Sia/modules/gateway.New
        /Users/jordan/golang/src/github.com/NebulousLabs/Sia/modules/gateway/gateway.go:170 +0xc5b

goroutine 12 [IO wait, 4 minutes]:
net.runtime_pollWait(0x2586238, 0x72, 0x0)
        /usr/local/Cellar/go/1.5.3/libexec/src/runtime/netpoll.go:157 +0x63
net.(*pollDesc).Wait(0xc8200f5100, 0x72, 0x0, 0x0)
        /usr/local/Cellar/go/1.5.3/libexec/src/net/fd_poll_runtime.go:73 +0x56
net.(*pollDesc).WaitRead(0xc8200f5100, 0x0, 0x0)
        /usr/local/Cellar/go/1.5.3/libexec/src/net/fd_poll_runtime.go:78 +0x44
net.(*netFD).accept(0xc8200f50a0, 0x0, 0x25c0000, 0xc82017e060)
        /usr/local/Cellar/go/1.5.3/libexec/src/net/fd_unix.go:408 +0x2f6
net.(*TCPListener).AcceptTCP(0xc82002c140, 0x0, 0x0, 0x0)
        /usr/local/Cellar/go/1.5.3/libexec/src/net/tcpsock_posix.go:254 +0x77
net.(*TCPListener).Accept(0xc82002c140, 0x0, 0x0, 0x0, 0x0)
        /usr/local/Cellar/go/1.5.3/libexec/src/net/tcpsock_posix.go:264 +0x4b
github.com/NebulousLabs/Sia/modules/gateway.(*Gateway).listen(0xc8200f5030)
        /Users/jordan/golang/src/github.com/NebulousLabs/Sia/modules/gateway/peers.go:116 +0x5d
created by github.com/NebulousLabs/Sia/modules/gateway.New
        /Users/jordan/golang/src/github.com/NebulousLabs/Sia/modules/gateway/gateway.go:173 +0xc80

goroutine 2075 [IO wait, 4 minutes]:
net.runtime_pollWait(0x25857b8, 0x72, 0x0)
        /usr/local/Cellar/go/1.5.3/libexec/src/runtime/netpoll.go:157 +0x63
net.(*pollDesc).Wait(0xc8201de060, 0x72, 0x0, 0x0)
        /usr/local/Cellar/go/1.5.3/libexec/src/net/fd_poll_runtime.go:73 +0x56
net.(*pollDesc).WaitRead(0xc8201de060, 0x0, 0x0)
        /usr/local/Cellar/go/1.5.3/libexec/src/net/fd_poll_runtime.go:78 +0x44
net.(*netFD).accept(0xc8201de000, 0x0, 0x25c0000, 0xc8201ce080)
        /usr/local/Cellar/go/1.5.3/libexec/src/net/fd_unix.go:408 +0x2f6
net.(*TCPListener).AcceptTCP(0xc820218000, 0x545d0, 0x0, 0x0)
        /usr/local/Cellar/go/1.5.3/libexec/src/net/tcpsock_posix.go:254 +0x77
net.(*TCPListener).Accept(0xc820218000, 0x0, 0x0, 0x0, 0x0)
        /usr/local/Cellar/go/1.5.3/libexec/src/net/tcpsock_posix.go:264 +0x4b
github.com/NebulousLabs/Sia/modules/host.(*Host).threadedListen(0xc8200d4a00)
        /Users/jordan/golang/src/github.com/NebulousLabs/Sia/modules/host/network.go:178 +0xf7
created by github.com/NebulousLabs/Sia/modules/host.(*Host).initNetworking
        /Users/jordan/golang/src/github.com/NebulousLabs/Sia/modules/host/network.go:94 +0x3f2
exit status 2
FAIL    github.com/NebulousLabs/Sia/modules/host        300.129s
make: *** [test-long] Error 1
```

modules/host/network.go
VoidingWarranties: I'm sorry I'm still failing to see how this reintroduces the bug.

You bring up a good point about placing the shutdown of the listener next to the creation of the listener, rather than the `listener.Accept`.

This pattern is in contrast to the shutdown of the `conn` created by `listener.Accept`. If we were to follow this pattern we would move `conn.Close` to `threadedListen` like so
```go
// ...
conn, err := h.listener.Accept()
if err != nil {
    return
}

// Close the conn on host.Close or when threadedHandleConn terminates,
// whichever comes first.
connCloseChan := make(chan struct{})
go func() {
    select {
    case <-h.tg.StopChan():
    case <-connCloseChan:
    }
    conn.Close()
}()

go func() {
    h.threadedHandleConn(conn)
    close(connCloseChan)
}()
// ...
```
It's different in that it is possible and perfectly fine for multiple goroutines to call and wait on `listener.Accept` simultaneously, but it is probably an error if multiple goroutines are calling `conn.Read` or `conn.Write` simultaneously.

NACK agreed.

modules/host/network.go
DavidVorick: What I'm seeing in the dump:

+ thread frozen on host.go:310 (`tg.Stop()`)
+ thread frozen on network.go:178 - looks like the file changed b/c network.go:178 in this PR points to the start of the function. Is it blocking on `tg.Add()` or another call?

If it's closing on `tg.Add()` it's an implementation error in `tg`, because `tg` should never hang on `tg.Add()`.

DavidVorick: OH, it's blocking on `listener.Accept` because the listener closer thread was not created in time. You're right, that's a deadlock.

DavidVorick:
Needs rebase, otherwise LGTM. Can confirm that this PR fixes a deadlock that currently exists in the host.

VoidingWarranties:
Rebased and also removed the listener closer-coroutine in favor if `ThreadGroup.OnStop`.

DavidVorick:
34 comments, a diff size of 40. lol

PR #1205, opened 2016-05-30
add RegisterCloser to ThreadGroup

lukechampine:
This makes it clean and easy to close `net.Conn`s and `net.Listener`s during shutdown. No extra goroutines needed.

My only concern is that it's a little opaque. But I think the tradeoff is worth it.

### DIFF
--- a/sync/threadgroup.go
+++ b/sync/threadgroup.go
@@ -78,6 +78,7 @@ func (tg *ThreadGroup) Stop() error {
 	for _, c := range tg.closers {
 		c.Close()
 	}
+	tg.closers = nil
 	tg.mu.Unlock()
 	tg.wg.Wait()
 	return nil
--- a/sync/threadgroup.go
+++ b/sync/threadgroup.go
@@ -2,7 +2,6 @@ package sync
 
 import (
 	"errors"
-	"io"
 	"sync"
 )
 
@@ -16,17 +15,17 @@ var ErrStopped = errors.New("ThreadGroup already stopped")
 // ThreadGroup is only intended be used once; as such, its Add and Stop
 // methods return errors if Stop has already been called.
 //
-// During shutdown, it is common to close resources such as net.Conns and
-// net.Listeners. Typically, this would require spawning a goroutine to wait
-// on the ThreadGroup's StopChan and then close the resource. To make this
-// more convenient, ThreadGroup provides a RegisterCloser method. io.Closers
-// registered in this manner will be automatically closed when Stop is called.
+// During shutdown, it is common to close resources such as net.Listeners.
+// Typically, this would require spawning a goroutine to wait on the
+// ThreadGroup's StopChan and then close the resource. To make this more
+// convenient, ThreadGroup provides an OnStop method. Functions passed to
+// OnStop will be called automatically when Stop is called.
 type ThreadGroup struct {
 	stopChan chan struct{}
 	chanOnce sync.Once
 	mu       sync.Mutex
 	wg       sync.WaitGroup
-	closers  []io.Closer
+	stopFns  []func()
 }
 
 // StopChan provides read-only access to the ThreadGroup's stopChan. Callers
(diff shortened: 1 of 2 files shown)


VoidingWarranties:
Cool! I like it.

sync/threadgroup.go
```diff
@@ -58,16 +66,33 @@ func (tg *ThreadGroup) Done() {
...
+	}
+	tg.closers = nil
 	tg.mu.Unlock()
 	tg.wg.Wait()
 	return nil
 }
+
+// RegisterCloser adds an io.Closer to the ThreadGroups closer set. Members of
+// the set will be closed when Stop is called. Note that this means the errors
+// returned by Close are not recoverable.
```
DavidVorick: Also note that this method should be exclusively called for threads which will be alive until `Close` is called. Good for `net.Listener`, not good for `net.Conn`. Also good for things like the renter repair loop and the hostdb scan loop.

sync/threadgroup.go
```diff
@@ -14,11 +15,18 @@ var ErrStopped = errors.New("ThreadGroup already stopped")
 // notifying callers when shutdown occurrs. Another key difference is that a
 // ThreadGroup is only intended be used once; as such, its Add and Stop
 // methods return errors if Stop has already been called.
+//
+// During shutdown, it is common to close resources such as net.Conns and
```
DavidVorick: not good for `net.Conn`

sync/threadgroup.go
```diff
@@ -14,11 +15,18 @@ var ErrStopped = errors.New("ThreadGroup already stopped")
...
+// on the ThreadGroup's StopChan and then close the resource. To make this
+// more convenient, ThreadGroup provides a RegisterCloser method. io.Closers
+// registered in this manner will be automatically closed when Stop is called.
 type ThreadGroup struct {
 	stopChan chan struct{}
 	chanOnce sync.Once
 	mu       sync.Mutex
 	wg       sync.WaitGroup
+	closers  []io.Closer
 }
```
DavidVorick: Just thinking about memory and garbage collection... I'm not sure that there's an easy way to tell when an object is garbage collected, but if you are holding a slice of `closers` it means that none of those objects will be GC'd until the `tg` is GC'd, even if the `tg` is already closed.

If everything is running smoothly, the `tg` will be GC'd shortly after it is shutdown, but we may want to wipe the list of closers anyway, to minimize the damage in the event of memory leaks.

sync/threadgroup.go
```diff
@@ -58,16 +66,33 @@ func (tg *ThreadGroup) Done() {
...
 	tg.mu.Lock()
 	if tg.IsStopped() {
 		tg.mu.Unlock()
 		return ErrStopped
 	}
 	close(tg.stopChan)
+	for _, c := range tg.closers {
+		c.Close()
+	}
+	tg.closers = nil
```
DavidVorick: oh nvm

VoidingWarranties:
This could be used with `net.Conn`s if there was a method such as `CloseCloserEarly(io.Closer)` which removed the closer from the set of closers and then closed it. Then you could write `threadedAcceptConn` methods like so
```go
func (m *Module) threadedAcceptConn(conn net.Conn) {
    m.threads.RegisterCloser(conn)
    defer m.threads.CloseCloserEarly(conn)
    if err := m.threads.Add(); err != nil {
        return err
    }
    defer m.threads.Done()
}
```

lukechampine:
@DavidVorick what do you think of making this more generic, i.e.:
```go
func (tg *ThreadGroup) OnStop(fn func()) error {
    tg.stoppers = append(tg.stoppers, fn) // with requisite locking, of course
}

func (tg *ThreadGroup) Stop() error {
    for _, fn := range tg.stoppers {
        fn()
    }
}

m.threads.OnStop(func() { m.listener.Close })
```

This would also make it possible to capture the listener error.

DavidVorick:
Generic stop function looks good to me. It makes it easier to ensure that resources are cleaned up at  the time that they are created, which reduces cognitive load. Though, I would probably have them get stopped in reverse order from how they are created. For example, you need to create the logger before you create the database, and you need to shutdown the logger after you shutdown the database, so you couldn't use `OnStop` for both the logger and the database unless you separate the closing of the logger from the creating of the logger, somewhat defeating the point.

Would also be nice to see a helper function for this code:

```
	// Forward the RPC port, if possible.
	g.threads.Add()
	go func() {
		defer g.threads.Done()
		g.forwardPort(port)
	}()
```

VoidingWarranties:
👍  for calling the `OnClose` functions in reverse order. Although offloading everything to these `OnClose` functions results in the `modules.Close` methods becoming just
```go
func (m *Module) Close error {
    if err := m.threads.Stop(); err != nil {
        return err
    }
}
```
which hides what is being closed. On the other hand it is nice to pair the `OnClose` calls with the creation of the closeable objects.

DavidVorick:
Ah derp I made the comment and then merged it before the change could happen.

@lukechampine do you mind fixing this and opening another PR? My mistake.

DavidVorick:
@VoidingWarranties Modules are pretty complex, I'm guessing that shutdown is always going to be more than just a call to `m.threads.Stop`, but I definitely think it's superior to set up an objects death at the same time you create it. Much less bug prone.

Also, `m.threads` is a pretty bad name imho, because it's not a set of threads it's more like a method for tracking what threads still need to be closed. I think `tg` is okay because we already use `mu` for `mutex`, `cs` for consensus set, and I think a few other common truncations.

PR #1207, opened 2016-05-30
Smarter SetAllowance + Auto-renew

lukechampine:
`SetAllowance` will now only form contracts as needed, to make up the difference between the current number of contracts and the desired number.
This new scheme may result in "heterogenous" contracts, i.e. a set of contracts ending at different heights.

Contracts are renewed as soon as the blockheight enters their "renew window." The amount of funds allocated per contract is calculated such that each renewed contract can support the same number of sectors.

To prevent slow rescans, contracts are only checked for renewal when `cs.Synced() == true`.

### DIFF
--- a/modules/host/negotiaterenewcontract.go
+++ b/modules/host/negotiaterenewcontract.go
@@ -241,7 +241,7 @@ func (h *Host) managedVerifyRenewedContract(so *storageObligation, txnSet []type
 	}
 	// The WindowEnd for the new file contract must be further in the future
 	// than the WindowEnd for the existing file contract.
-	if fc.WindowEnd <= so.proofDeadline() {
+	if fc.WindowEnd < so.proofDeadline() {
 		return errRenewDoesNotExtend
 	}
 
--- a/modules/renter/contractor/allowance.go
+++ b/modules/renter/contractor/allowance.go
@@ -101,7 +101,8 @@ func (c *Contractor) SetAllowance(a modules.Allowance) error {
 	// should not change either
 	endHeight := c.blockHeight + a.Period
 	if a.Period == c.allowance.Period && len(c.contracts) > 0 {
-		endHeight = c.contractEndHeight()
+		// COMPAT v0.6.0 - old hosts require end height increase by at least 1
+		endHeight = c.contractEndHeight() + 1
 	}
 	c.mu.RUnlock()
 
--- a/modules/renter/contractor/allowance.go
+++ b/modules/renter/contractor/allowance.go
@@ -72,12 +72,22 @@ func (c *Contractor) SetAllowance(a modules.Allowance) error {
 
 	c.mu.RLock()
 	shouldRenew := a.Period != c.allowance.Period || a.Funds.Cmp(c.allowance.Funds) != 0
+	shouldWait := c.blockHeight+a.Period < c.contractEndHeight()
 	remaining := int(a.Hosts) - len(c.contracts)
 	c.mu.RUnlock()
 
-	// if no contracts need renewing, run special-case logic instead
 	if !shouldRenew {
+		// If no contracts need renewing, just form new contracts.
 		return c.managedFormAllowanceContracts(remaining, numSectors, a)
+	} else if shouldWait {
+		// If the new period would result in an earlier endHeight, we can't
+		// renew; instead, set the allowance without modifying any contracts.
+		// They will be renewed with the new allowance when they expire.
+		c.mu.Lock()
+		c.allowance = a
+		err = c.saveSync()
+		c.mu.Unlock()
+		return err
 	}
 
 	c.mu.RLock()
--- a/modules/renter/contractor/allowance.go
+++ b/modules/renter/contractor/allowance.go
@@ -8,6 +8,13 @@ import (
 	"github.com/NebulousLabs/Sia/types"
 )
 
+var (
+	errAllowanceNoHosts    = errors.New("hosts must be non-zero")
+	errAllowanceZeroPeriod = errors.New("period must be non-zero")
+	errAllowanceZeroWindow = errors.New("renew window must be non-zero")
+	errAllowanceWindowSize = errors.New("renew window must be less than period")
+)
+
 // contractEndHeight returns the height at which the Contractor's contracts
 // end. If there are no contracts, it returns zero.
 func (c *Contractor) contractEndHeight() types.BlockHeight {
@@ -46,13 +53,13 @@ func (c *Contractor) contractEndHeight() types.BlockHeight {
 func (c *Contractor) SetAllowance(a modules.Allowance) error {
 	// sanity checks
 	if a.Hosts == 0 {
-		return errors.New("hosts must be non-zero")
+		return errAllowanceNoHosts
 	} else if a.Period == 0 {
-		return errors.New("period must be non-zero")
+		return errAllowanceZeroPeriod
 	} else if a.RenewWindow == 0 {
-		return errors.New("renew window must be non-zero")
+		return errAllowanceZeroWindow
 	} else if a.RenewWindow >= a.Period {
-		return errors.New("renew window must be less than period")
+		return errAllowanceWindowSize
 	}
 
 	// check that allowance is sufficient to store at least one sector
--- a/modules/renter/contractor/allowance.go
+++ b/modules/renter/contractor/allowance.go
@@ -3,9 +3,29 @@ package contractor
 import (
 	"errors"
 
+	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/types"
 )
+
+// contractEndHeight returns the height at which the Contractor's contracts
+// end. If there are no contracts, it returns zero.
+func (c *Contractor) contractEndHeight() types.BlockHeight {
+	var endHeight types.BlockHeight
+	for _, contract := range c.contracts {
+		endHeight = contract.EndHeight()
+		break
+	}
+	// sanity check: all contracts should have same EndHeight
+	if build.DEBUG {
+		for _, contract := range c.contracts {
+			if contract.EndHeight() != endHeight {
+				build.Critical("all contracts should have EndHeight", endHeight, "-- got", contract.EndHeight())
+			}
+		}
+	}
+	return endHeight
+}
 
 // SetAllowance sets the amount of money the Contractor is allowed to spend on
 // contracts over a given time period, divided among the number of hosts
@@ -63,11 +83,8 @@ func (c *Contractor) SetAllowance(a modules.Allowance) error {
 	// calculate new endHeight; if the period has not changed, the endHeight
 	// should not change either
 	endHeight := c.blockHeight + a.Period
-	if a.Period == c.allowance.Period {
-		for _, contract := range c.contracts {
-			endHeight = contract.EndHeight()
-			break
-		}
+	if a.Period == c.allowance.Period && len(c.contracts) > 0 {
+		endHeight = c.contractEndHeight()
 	}
 	c.mu.RUnlock()
 
@@ -130,9 +147,8 @@ func (c *Contractor) managedFormAllowanceContracts(n int, numSectors uint64, a m
 	// should be a full period.
 	c.mu.RLock()
 	endHeight := c.blockHeight + a.Period
-	for _, contract := range c.contracts {
-		endHeight = contract.EndHeight()
-		break
+	if len(c.contracts) > 0 {
+		endHeight = c.contractEndHeight()
 	}
 	c.mu.RUnlock()
 
--- a/modules/renter/contractor/allowance.go
+++ b/modules/renter/contractor/allowance.go
@@ -1,0 +1,161 @@
+package contractor
+
+import (
+	"errors"
+
+	"github.com/NebulousLabs/Sia/modules"
+	"github.com/NebulousLabs/Sia/types"
+)
+
+// SetAllowance sets the amount of money the Contractor is allowed to spend on
+// contracts over a given time period, divided among the number of hosts
+// specified. Note that Contractor can start forming contracts as soon as
+// SetAllowance is called; that is, it may block.
+//
+// In most cases, SetAllowance will renew existing contracts instead of
+// forming new ones. This preserves the data on those hosts. When this occurs,
+// the renewed contracts will atomically replace their previous versions. If
+// SetAllowance is interrupted, renewed contracts may be lost, though the
+// allocated funds will eventually be returned,
+//
+// TODO: can an Editor or Downloader be used across renewals?
+// TODO: will hosts allow renewing the same contract twice?
+//
+// NOTE: At this time, transaction fees are not counted towards the allowance.
+// This means the contractor may spend more than allowance.Funds.
+func (c *Contractor) SetAllowance(a modules.Allowance) error {
+	// sanity checks
+	if a.Hosts == 0 {
+		return errors.New("hosts must be non-zero")
+	} else if a.Period == 0 {
+		return errors.New("period must be non-zero")
+	} else if a.RenewWindow == 0 {
+		return errors.New("renew window must be non-zero")
+	} else if a.RenewWindow >= a.Period {
+		return errors.New("renew window must be less than period")
+	}
+
+	// check that allowance is sufficient to store at least one sector
+	numSectors, err := maxSectors(a, c.hdb)
+	if err != nil {
+		return err
+	} else if numSectors == 0 {
+		return errInsufficientAllowance
+	}
+
+	c.mu.RLock()
+	shouldRenew := a.Period != c.allowance.Period || a.Funds.Cmp(c.allowance.Funds) != 0
+	remaining := int(a.Hosts) - len(c.contracts)
+	c.mu.RUnlock()
+
+	// if no contracts need renewing, run special-case logic instead
+	if !shouldRenew {
+		return c.managedFormAllowanceContracts(remaining, numSectors, a)
+	}
+
+	c.mu.RLock()
+	// gather contracts to renew
+	var renewSet []modules.RenterContract
+	for _, contract := range c.contracts {
+		renewSet = append(renewSet, contract)
+	}
+
+	// calculate new endHeight; if the period has not changed, the endHeight
+	// should not change either
+	endHeight := c.blockHeight + a.Period
+	if a.Period == c.allowance.Period {
+		for _, contract := range c.contracts {
+			endHeight = contract.EndHeight()
+			break
+		}
+	}
+	c.mu.RUnlock()
+
+	// renew existing contracts with new allowance parameters
+	newContracts := make(map[types.FileContractID]modules.RenterContract)
+	for _, contract := range renewSet {
+		newContract, err := c.managedRenew(contract, numSectors, endHeight)
+		if err != nil {
+			c.log.Printf("WARN: failed to renew contract with %v; a new contract will be formed in its place", contract.NetAddress)
+			remaining++
+			continue
+		}
+		newContracts[newContract.ID] = newContract
+		if len(newContracts) >= int(a.Hosts) {
+			break
+		}
+	}
+
+	// if we did not renew enough contracts, form new ones
+	if remaining > 0 {
+		formed, err := c.managedFormContracts(remaining, numSectors, endHeight)
+		if err != nil {
+			return err
+		}
+		for _, contract := range formed {
+			newContracts[contract.ID] = contract
+		}
+	}
+
+	// if we weren't able to form anything, return an error
+	if len(newContracts) == 0 {
+		return errors.New("unable to form or renew any contracts")
+	}
+
+	// Set the allowance and replace the contract set
+	c.mu.Lock()
+	c.allowance = a
+	c.contracts = newContracts
+	// update metrics
+	var spending types.Currency
+	for _, contract := range c.contracts {
+		spending = spending.Add(contract.RenterFunds())
+	}
+	c.financialMetrics.ContractSpending = spending
+	err = c.saveSync()
+	c.mu.Unlock()
+
+	return err
+}
+
+// managedFormAllowanceContracts handles the special case where no contracts
+// need to be renewed when setting the allowance.
+func (c *Contractor) managedFormAllowanceContracts(n int, numSectors uint64, a modules.Allowance) error {
+	if n <= 0 {
+		return nil
+	}
+
+	// if we're forming contracts but not renewing, the new contracts should
+	// have the same endHeight as the existing ones. Otherwise, the endHeight
+	// should be a full period.
+	c.mu.RLock()
+	endHeight := c.blockHeight + a.Period
+	for _, contract := range c.contracts {
+		endHeight = contract.EndHeight()
+		break
+	}
+	c.mu.RUnlock()
+
+	// form the contracts
+	formed, err := c.managedFormContracts(n, numSectors, endHeight)
+	if err != nil {
+		return err
+	}
+
+	// Set the allowance and replace the contract set
+	c.mu.Lock()
+	c.allowance = a
+	for _, contract := range formed {
+		c.contracts[contract.ID] = contract
+	}
+	// update metrics
+	var spending types.Currency
+	for _, contract := range c.contracts {
+		spending = spending.Add(contract.RenterFunds())
+	}
+	c.financialMetrics.ContractSpending = spending
+	err = c.saveSync()
+	c.mu.Unlock()
+
+	return err
+}
--- a/modules/renter/contractor/contractor.go
+++ b/modules/renter/contractor/contractor.go
@@ -56,6 +56,9 @@ func (c *Contractor) FinancialMetrics() modules.RenterFinancialMetrics {
 // contracts over a given time period, divided among the number of hosts
 // specified. Note that Contractor can start forming contracts as soon as
 // SetAllowance is called; that is, it may block.
+//
+// NOTE: At this time, transaction fees are not counted towards the allowance.
+// This means the contractor may spend more than allowance.Funds.
 func (c *Contractor) SetAllowance(a modules.Allowance) error {
 	// sanity checks
 	if a.Hosts == 0 {
@@ -68,35 +71,43 @@ func (c *Contractor) SetAllowance(a modules.Allowance) error {
 		return errors.New("renew window must be less than period")
 	}
 
-	err := c.formContracts(a)
-	if err != nil {
-		return err
+	c.mu.RLock()
+	// calculate number of new contracts to form
+	var nContracts int
+	if a.Hosts > uint64(len(c.contracts)) {
+		nContracts = int(a.Hosts) - len(c.contracts)
+	}
+
+	// calculate end height of new contracts
+	endHeight := c.blockHeight + a.Period
+
+	// calculate how much can be spent on new contracts
+	var currentSpent types.Currency
+	for _, contract := range c.contracts {
+		currentSpent = currentSpent.Add(contract.FileContract.ValidProofOutputs[0].Value)
+	}
+	c.mu.RUnlock()
+
+	// only form contracts if we need to and have enough funds to do so
+	if nContracts > 0 {
+		if a.Funds.Cmp(currentSpent) > 0 {
+			funds := a.Funds.Sub(currentSpent)
+			err := c.managedFormContracts(nContracts, funds, endHeight)
+			if err != nil {
+				return err
+			}
+		} else {
+			c.log.Println("WARN: want to form more contracts, but new allowance is too small")
+		}
 	}
 
 	// Set the allowance.
 	c.mu.Lock()
 	c.allowance = a
-	err = c.saveSync()
+	err := c.saveSync()
 	c.mu.Unlock()
 
 	return err
-
-	/*
-		// If this is the first time the allowance has been set, form contracts
-		// immediately.
-		if old.Hosts == 0 {
-			return c.formContracts(a)
-		}
-
-		// Otherwise, if the new allowance is "significantly different" (to be
-		// defined more precisely later), form intermediary contracts.
-		if a.Funds.Cmp(old.Funds) > 0 {
-			// TODO: implement
-			// c.formContracts(diff(a, old))
-		}
-
-		return nil
-	*/
 }
 
 // Contracts returns the contracts formed by the contractor.
--- a/modules/renter/contractor/contractor.go
+++ b/modules/renter/contractor/contractor.go
@@ -78,27 +78,13 @@ func (c *Contractor) SetAllowance(a modules.Allowance) error {
 	if a.Hosts > uint64(len(c.contracts)) {
 		nContracts = int(a.Hosts) - len(c.contracts)
 	}
-
-	// calculate end height of new contracts
-	endHeight := c.blockHeight + a.Period
-
-	// calculate how much can be spent on new contracts
-	var currentSpent types.Currency
-	for _, contract := range c.contracts {
-		currentSpent = currentSpent.Add(contract.FileContract.ValidProofOutputs[0].Value)
-	}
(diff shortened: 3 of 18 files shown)


api/renter.go
```diff
@@ -103,14 +103,18 @@ func (srv *Server) renterAllowanceHandlerPOST(w http.ResponseWriter, req *http.R
 	// 	writeError(w, "Couldn't parse renewwindow: "+err.Error(), http.StatusBadRequest)
 	// 	return
 	// }
+	renewWindow := period / 4
```
DavidVorick: I would prefer period/2, because there's not much cost to doing it that way, and it means that people can forget about Sia for up to 6 weeks at a time. 6 weeks is a period used by most private bittorrent trackers, and it seems to work pretty well for them.

If it were only 3 weeks I'd worry that people would forget and lose their files.

lukechampine: ok, reverted

modules/renter/contractor/contractor_test.go
```diff
@@ -130,28 +132,81 @@ type stubHostDB struct{}
...
-	c := &Contractor{
-		// an empty hostDB ensures that calls to formContracts will always fail
-		hdb: stubHostDB{},
+// TestIntegrationSetAllowance tests the SetAllowance method.
+func TestIntegrationSetAllowance(t *testing.T) {
+	if testing.Short() {
+		t.SkipNow()
+	}
+	// create testing trio
+	h, c, m, err := newTestingTrio("TestIntegrationAutoRenew")
```
DavidVorick: I don't like the name of this function

modules/renter/contractor/formcontract.go
```diff
@@ -53,66 +87,61 @@ func (c *Contractor) managedNewContract(host modules.HostDBEntry, filesize uint6
...
+	var exclude []modules.NetAddress
+	for _, contract := range c.contracts {
+		exclude = append(exclude, contract.NetAddress)
 	}
-	// Calculate average host price.
-	var sum types.Currency
-	for _, h := range hosts {
-		sum = sum.Add(h.StoragePrice)
+	c.mu.RUnlock()
+	hosts := c.hdb.RandomHosts(nRandomHosts, exclude)
```
DavidVorick: good to see the exclusion being used

DavidVorick:
To summarize discussion that happened offline:

The renter should be able to give users access to a larger allowance immediately, which requires creating new contracts if a larger allowance is specified. This should of course be accompanied by a test.

There was also some discussion about what should happen if the renter reduces the period of the contract, such that the new renew point is sooner than the previous renew point. Because we want to make sure that contracts always end at the same height, the best thing to do is probably to renew immediately, just the same as we would with a bigger allowance. This includes if the allowance has stayed the same or decreased.

modules/renter/contractor/contractor.go
```diff
@@ -81,41 +76,76 @@ func (c *Contractor) SetAllowance(a modules.Allowance) error {
...
+	// check if allowance has different period or funds; if so, we should
+	// renew existing contracts with the new allowance.
+	// TODO: compare numSectors instead of allowance.Funds?
+	var renewSet []modules.RenterContract
+	if a.Period != c.allowance.Period || a.Funds.Cmp(c.allowance.Funds) != 0 {
+		for _, contract := range c.contracts {
+			renewSet = append(renewSet, contract)
+		}
+		// TODO: dangerous -- figure out a safer approach later
+		c.contracts = map[types.FileContractID]modules.RenterContract{}
```
lukechampine: In the new approach, as soon as you set a different allowance, all the old contracts are "deleted" (moved from `c.contracts` to `renewSet`). This gives us a clean slate to work with, so that we know all contracts in `c.contracts` were formed with the same allowance.
The downside is that if you shutdown before `SetAllowance` returns, you might have deleted a contract without renewing it. So when you start up again, you'll need to form a new contract instead of renewing.

Another unresolved issue is that contracts can still be "heterogeneous" with respect to their end height. That is, if you call `SetAllowance` once and it forms 3 out of 6 contracts, and then call it again 10 blocks later, you'll wind up with 3 contracts ending at one height and 3 ending at another. I don't think this is a huge issue, but it's worth fixing for the sake of robustness.

DavidVorick: Hmm, I don't think you should be deleting contracts until the renew for that contract has completed. Ideally upon restart it would pick up where it left off, including renewing to a homogeneous height (same height as the previous ones were already renewed to)

Definitely lots of edge cases here, more than I was expecting.

lukechampine: The problem is that in the meantime, people are calling `c.Contracts()` and `ProcessConsensusChange` may be trying to auto-renew. So the old contracts cannot be left in `c.contracts` while `SetAllowance` is running. They have to be atomically replaced.
This suggests an alternative, in which you would incrementally renew contracts and store them in some `tempContracts` map, and then atomically swap `tempContracts` for `c.contracts`. But this is even worse, because if you exit early you lose all of the renewed contracts.
To rectify this, you could move `tempContracts` into the contractor and save it to disk. But there are still problems! Namely, until you do the atomic swap, `ProcessConsensusChange` will still be looking at the old contracts, and thus it may try to renew them, which is wasteful at best.
So yes, lots of edge cases :P

This brings up a good question: what happens when you try to renew a contract and revise it at the same time? Will the host reject one of the operations? This scenario will occur if the user sets a new allowance while files are being uploaded.

DavidVorick: You can pretty easily atomically update the allowance. And you can also (more work, but easy work) easily keep track of all the in-progress rewnewals happening at the moment of an updated allowance. Then, when doing a renew, you can verify against the allowance that was atomically updated.

So the logic flow probably looks something like this:

1. atomically update allowance
2. all renewals go into a queue - both from changing the allowance and from ProcessConsensusChange, and upon being insert they check for conflicts (is there already a renewal for this contract in the queue, and if so it should be replaced)
3. some other thread pops stuff out of the queue and updates the set of contracts.

That should be safe

lukechampine:
I think we should consider restricting the `/renter/allowance` API call to a single settable value: the number of bytes the user wants to store. (The number of funds is also reasonable, but what if the value of the siacoin changes? Users care about bytes, not coins.)
The other allowance parameters are extraneous. `Hosts` certainly doesn't need to be settable; it should just be the number of pieces in the default erasure coding. Most users won't care about `Period` either, and certainly not `RenewWindow`.

To complement this, we would add `form`/`renew` endpoints to the renter to allow for manual contract management.

DavidVorick:
That creates a really huge gap on the customization spectrum. Either you you are choosing the number of bytes, or you are manually managing contracts. That's definitely too much of a gap.

When you are forming contracts based on a byte count, you are glossing over significant facts, such as how much downloading might be expected from the contract, volatility in the price of storage, volatility in the price of the siacoin, and the fact that you will lose a certain number of hosts each renew cycle.

I definitely don't want to see the 'pick a number of bytes' abstraction in the API.

VoidingWarranties:
A lack of a bytes abstraction in the API will be a hindrance to third party devs. I'm worried that this will offload the abstraction to the majority of applications built on top of our platform. I'm also worried that this bytes abstraction is complex enough to be difficult or bothersome to write without our domain specific knowledge.

With that said the bytes abstraction requires a lot of user interaction. What happens if the price of storage increased 2x? Certainly the user would want to know about that and have a chance to stop the contracts from renewing. Then again I suppose that is only necessary for applications that let the user interact directly with Sia. I imagine the majority of applications that could be built on Sia would only use Sia as a backend, and would not expose these details to the user and would instead consider the 2x increase in price as a cost of business.

So for the renter plugin in the GUI the bytes abstraction is necessary, whether in the API or the GUI. What other plugins or applications might want this abstraction?
* [pmknutsen's GNOME shell Sia extension](https://github.com/pmknutsen/gnome-shell-sia) and similar applications, for example a Sia FUSE extension.
* Siaboxes (NAS with Sia support built in) might want this abstraction if they want to provide renting abilities to the user. Although I imagine most Siaboxes would be primarily used for hosting.

All other use cases I can think of would not need the bytes abstraction, as they would presumably store multiple user's files in the same file contracts.

lukechampine:
I think I'm with David on this one. If you want to use bytes instead of coins, then you really need three args: bytes stored, bytes uploaded, and bytes downloaded.
It shouldn't be terribly difficult for API consumers to calculate an allowance if they know the average host settings, which we plan to make accessible.

lukechampine:
Actually, another thought occurred to me, more in favor of bytes:
Internally, `SetAllowance` uses `allowance.Funds` and a sampling of hosts to determine how much money to allocate to each host. Currently, it do so by first calculating `allowance.Funds` / (averageSectorPrice * allowance.Hosts)`, which yields the number of sectors that the allowance can support. It then forms contracts with the hosts, multiplying `numSectors` by the storage price for that host.
But this leaves upload/download price entirely out of the picture! In order to properly account for them, `SetAllowance` would have to know how many uploads/downloads the allowance can support -- but also what fraction of the allowance should be spent on uploads and downloads. How do we determine this? Wouldn't the user need to supply these values as well?

lukechampine:
The new approach is to atomically replace the contract set only when all contracts have been formed/renewed.
This has the disadvantage that quitting siad mid-`SetAllowance` may result in loss of contracts. However, I think this is the better of the two naive implementations; the old implementation deleted the old contracts immediately, which is much more surprising to the user.
Later, if necessary, we can add support for resuming `SetAllowance` across sessions. But given that it should only take a minute or two, and the fact that no coins are at risk (they are "frozen" for the contract duration, but not lost) this feels like a low priority.
It would also be quite easy to parallelize the RPCs.

I also encountered a bug(?) where the host would reject renewals with the same `WindowStart`. These renewals will occur if the user changes the allowance funds, but not the period. I could have subverted this by always bumping the `WindowStart` by 1, but this seemed more correct.

lukechampine:
Metrics reporting could use some improvements. Currently, `ContractsSpending` will contain the sum of `c.contracts`. This may not give a correct indication of how many funds are locked up in contracts.

DavidVorick:
I'm **very strongly** in favor of keeping it as a monetary allowance. Things like price changes, download behavior, and upload behavior are too difficult to predict. We don't want the API to be doing that level of management for the user. In the UI it makes sense to me to have this really blunt tool, but if you are developing on the API you need to have an understanding that creating contracts, uploading files, downloading files, and keeping storage around for a long time all incur different costs. Abstracting that away is taking too much power away from our developers and slamming them into a single paradigm that assumes a single usage pattern.

The whole point of supporting third party development is to give the ecosystem a way to cater to usage patterns and use cases that we don't have the resources to cater to directly.

If you are a GUI user, I don't want you to have to learn all the things about contracts and upload costs and whatnot, we want the least friction possible between hearing about Sia and performing your first upload. If you are an API developer, I'm assuming that you are building for some use case which is poorly covered by the GUI, which means you probably have different requirements and patterns. I want our API developers to have a good understanding of the mechanics behind setting an allowance, spending money to upload and storage files, spending money to renew, and spending money to download files.

---

Switching out the set of contracts atomically seems good enough. It's not ideal, ideally you'd be able to resume between sessions, but we can add that later as things evolve.

lukechampine:
People who want precise control shouldn't be using `SetAllowance` anyway, since it doesn't let you pick your own hosts.
Perhaps instead of packaging everything into one super-smart `SetAllowance` route, we could split the functionality across multiple calls, and let the user decide how much "smartness" they want to rely on. e.g.:

| Route | Params | Description |
|---------|------------|---------------|
| GET `/renter/contracts` |  | List contracts |
| GET `/renter/contracts/:id` |  | View contract info |
| POST `/renter/contracts/:id` | `autorenew` | Toggle auto-renew (?) |
| POST `/renter/contracts/form` | `host`, `funds`, `endHeight` | Form a contract |
| POST `/renter/contracts/renew` | `id`, `funds`, `endHeight` | Renew a contract |
| GET `/renter/hosts` | `n`, `sort` | List hosts |

Given these routes, you could emulate `SetAllowance` via:
1. GET `/renter/hosts?n=24&sort=price`
2. Calculate how much to spend with each host
3. POST `/renter/contracts/form` (for each host)
4. POST `/renter/contracts/:id?autorenew=true` (for each contract)

I like this in principle because it is less "magic" and more "do one thing well," without sacrificing too much in the way of usability. The `autorenew` bit is a little ugly, but besides that the routes are all pretty straightforward.
The downside is that we offload some work to the caller. For example, step 2 requires them to know which host fields to use when calculating cost, and they have to remember to use big ints. But the larger problem is what to do when contract formation fails. You need to pick new hosts and try again, and you may need some state to persist across sessions. On the other hand, the UI might be better suited to handle these sorts of errors anyway. If initial formation fails, you can display a dialogue explaining what happened and give the user options for what to do next: Try to form contracts with more expensive hosts? Abort and try again later? Decrease allowance? This flexibility would be difficult to implement if everything were handled on the backend.

In general, when we design this stuff it's probably a good idea to bear in the mind the distinction between [Simple and Easy](https://github.com/matthiasn/talk-transcripts/blob/master/Hickey_Rich/SimpleMadeEasy.md).

DavidVorick:
I think that the difference between picking the amount of money you are spending and picking your own hosts is pretty huge. There are lots of reasons to have the renter do the host picking - the renter has a far better idea of which hosts are reliable, etc. It's something I can see being exposed later, but in the meantime I'd rather have filters that get passed in to block certain hosts, and maybe block certain prices or speeds.

And then we can start talking about optimizations like giving more money to certain hosts because we know that we will download from them more frequently. That's an optimization you'd expect from the renter, not from the person utilizing the API.

I also don't think it makes sense to have the developer managing renew schedules. As we've seen from implementing it ourselves, it's a pretty difficult thing to get right.

No part of managing files on Sia is simple. Getting a file safely onto Sia means making connections to multiple different hosts and giving it to them. Doing it cheaply means applying erasure coding and distinguishing between the hosts based on price, reliability, and utility. Getting scalability means pre-paying and utilizing file contract revisions. If you are in the UK, following the law means adding geographic filtering.

We should shape the API based on the use cases of the developers who will be using the API, keeping in mind that we can always expose lower level functions like what the contractor does at a later date. We need to think about what types of things we should be supporting early on, and what types of things we'd add support for later. We're only going to have a successful ecosystem emerge if people can figure out safe ways to get files on to Sia in ways that make sense for their application.

---

The simplest use case is uploading a file. It makes sense to have a route that gives you as little friction as possible on the road from getting a brand new Sia to uploading a file. In that case, you might even have the daemon automatically set an allowance and upload contracts based on that, refusing to get started unless enough coins were in the wallet. And then the renter can automatically manage renewals and per-host budget increases based on usage patterns.

The original idea behind setting the allowance was not to estimate the amount of coins you are supposed to allocate, but rather to put a leash on the renter and make sure it did not spend more than a specific amount. And this is how it should be presented to the user. They should not be trying to guess their usage, that's a pretty good way to make sure that they don't use it very much - they'll be thinking about spending money.

I think we all got lost in the weeds on this one. The renter should be doing all of the heavy lifting on the estimation and management. At some point, we'll want to expose lower level APIs for power users so they can control things like exact redundancy, and perhaps also do their own host selection, but that's not something that we need to support at the moment.

As far as the API goes, and also as far as the UI goes, I would like to see the allowance be used as a money field, and as a maximum instead of an estimate. Users should be encouraged to over-estimate, because they get their money back. And, as the renter evolves, I think it would make sense that the renter doesn't even use the whole allowance when forming contracts, instead forming contracts based on usage patterns and doing renewals when they start to run out of money, keeping tabs on the allowance.

An increased allowance should not result in renewals (it's okay for now), but instead renewals should always be performed on an as-needed basis. After finishing an action, some check should be made to see if the balance is running low, the same way that a check is made at the end of each block to see if time is running low.

modules/host/negotiaterenewcontract.go
```diff
@@ -241,7 +241,7 @@ func (h *Host) managedVerifyRenewedContract(so *storageObligation, txnSet []type
 	}
 	// The WindowEnd for the new file contract must be further in the future
 	// than the WindowEnd for the existing file contract.
-	if fc.WindowEnd <= so.proofDeadline() {
+	if fc.WindowEnd < so.proofDeadline() {
```
DavidVorick: Uh oh.

This breaks compatibility with all v0.6.0 hosts. I see why you did it but you can't utilize it until everyone upgrades to 1.0, otherwise contract renewals won't work on our network.

DavidVorick: Does it mess things up too badly if you just increase the height by '1' every time?

modules/renter/contractor/allowance.go
```diff
@@ -0,0 +1,161 @@
...
+// SetAllowance is called; that is, it may block.
+//
+// In most cases, SetAllowance will renew existing contracts instead of
+// forming new ones. This preserves the data on those hosts. When this occurs,
+// the renewed contracts will atomically replace their previous versions. If
+// SetAllowance is interrupted, renewed contracts may be lost, though the
+// allocated funds will eventually be returned,
+//
+// TODO: can an Editor or Downloader be used across renewals?
+// TODO: will hosts allow renewing the same contract twice?
```
DavidVorick: I don't think hosts like it when you touch a contract which has already been renewed.

DavidVorick: hmmm, actually I changed my mind. I think its' fine. They keep everything separate, and they track redundancy on their own.

DavidVorick: Easiest way to check is write a test.

modules/renter/contractor/allowance.go
```diff
@@ -0,0 +1,161 @@
...
+	// check that allowance is sufficient to store at least one sector
+	numSectors, err := maxSectors(a, c.hdb)
+	if err != nil {
+		return err
+	} else if numSectors == 0 {
+		return errInsufficientAllowance
+	}
+
+	c.mu.RLock()
+	shouldRenew := a.Period != c.allowance.Period || a.Funds.Cmp(c.allowance.Funds) != 0
```
DavidVorick: If I understand this correctly, you should renew if the contractor period is not equal to the allowance period, in an attempt to get them to the same spot?

modules/renter/contractor/allowance.go
```diff
@@ -0,0 +1,161 @@
...
+	remaining := int(a.Hosts) - len(c.contracts)
+	c.mu.RUnlock()
+
+	// if no contracts need renewing, run special-case logic instead
+	if !shouldRenew {
+		return c.managedFormAllowanceContracts(remaining, numSectors, a)
+	}
+
+	c.mu.RLock()
```
DavidVorick: I think you should break this into two functions. If the code for handling the allowance is completely separate depending on `shouldRenew`, it'd probably be more accessable to do:

```
if !shouldRenew {
    return case1()
}
return case2()
```

Would probably make testing nicer as well.

modules/renter/contractor/allowance.go
```diff
@@ -0,0 +1,161 @@
...
+
+	// if we're forming contracts but not renewing, the new contracts should
+	// have the same endHeight as the existing ones. Otherwise, the endHeight
+	// should be a full period.
+	c.mu.RLock()
+	endHeight := c.blockHeight + a.Period
+	for _, contract := range c.contracts {
+		endHeight = contract.EndHeight()
+		break
+	}
```
DavidVorick: This is an odd loop. Is there a better way to do the same thing?

modules/renter/contractor/allowance.go
```diff
@@ -0,0 +1,161 @@
...
+		c.contracts[contract.ID] = contract
+	}
+	// update metrics
+	var spending types.Currency
+	for _, contract := range c.contracts {
+		spending = spending.Add(contract.RenterFunds())
+	}
+	c.financialMetrics.ContractSpending = spending
+	err = c.saveSync()
+	c.mu.Unlock()
```
DavidVorick: So it looks like the atomic swap thing is happening in this code. And the eventual plan (post 1.0) is to have some sort of saved field where contracts that were successfully formed but not swapped in get restored and you finish whatever didn't get finished the first time?

DavidVorick: basically resumable renew

modules/host/negotiaterenewcontract.go
lukechampine: I don't think it would be too bad. Can we do both and just phase out the +1 later?

modules/renter/contractor/allowance.go
lukechampine: correct, or if the amount of funds has changed.
I suppose it ought to be `a.Period > c.allowance.Period`, since hosts will reject a renewal with a lower end height. Can you confirm?

modules/renter/contractor/allowance.go
lukechampine: yeah, it's weird because it's a map. A slice would make it easier. Another option would be to keep the endHeight in a separate contractor variable, or to make it a method.

In general I'm uneasy with the assumption that all contracts will have the same endHeight. Feels vulnerable to bugs.

modules/renter/contractor/allowance.go
lukechampine: correct.

modules/host/negotiaterenewcontract.go
DavidVorick: Yeah, that would be best. But run all of your testing with the `<=`, which means cluttering the code a bit but we need to make sure that the testing is handling the correct case.

modules/renter/contractor/allowance.go
DavidVorick: Best to verify your assumptions by writing a test. It's easy for other developers to not realize they are assumptions and break things if there's no test to inform them that they've broken compatibility.

I'm pretty sure you are correct though, hosts will reject any renewals that don't have a strictly greater height (until your change). I'm not 100% sure why this is the case.

modules/renter/contractor/allowance.go
DavidVorick: Might be good to throw in a sanity check then. If you are doing atomic swaps it shouldn't be a problem, but it's better to have code which reassures you it's not a problem then to have stuff silently failing, especially since you might forget about this 4 months down the line when some change we make starts causing problems.

modules/renter/contractor/allowance.go
```diff
@@ -0,0 +1,161 @@
...
+	// gather contracts to renew
+	var renewSet []modules.RenterContract
+	for _, contract := range c.contracts {
+		renewSet = append(renewSet, contract)
+	}
+
+	// calculate new endHeight; if the period has not changed, the endHeight
+	// should not change either
+	endHeight := c.blockHeight + a.Period
+	if a.Period == c.allowance.Period {
```
DavidVorick: This is another spot of code that seems brittle to me. You've got a dangling assumption, which is that the endHeight should only change if the period has changed. The coupling between the assumption and execution are loose.

I think it's sort of the same vein as some of the stuff we do with clean shutdown. While it's more intuitive to have a `New` method and a `Close` method that each define all the startup and shutdown functions, it decouples things that are supposed to be tightly coupled. So instead we use thread groups to define an object's shutdown right after we define its startup, that way behavior which is tightly coupled is directed by code which is also tightly coupled.

modules/renter/contractor/allowance.go
```diff
@@ -0,0 +1,161 @@
...
+			continue
+		}
+		newContracts[newContract.ID] = newContract
+		if len(newContracts) >= int(a.Hosts) {
+			break
+		}
+	}
+
+	// if we did not renew enough contracts, form new ones
+	if remaining > 0 {
```
DavidVorick: Kind of a confusing variable name

modules/renter/contractor/contractor_test.go
```diff
@@ -130,28 +132,128 @@ type stubHostDB struct{}
...
-	err := c.SetAllowance(modules.Allowance{Funds: types.NewCurrency64(1), Period: 0, Hosts: 3})
+	var a modules.Allowance
+	err = c.SetAllowance(a)
 	if err == nil {
 		t.Error("expected error, got nil")
 	}
-	err = c.SetAllowance(modules.Allowance{Funds: types.NewCurrency64(1), Period: 2, Hosts: 0})
+	a.Hosts = 1
+	err = c.SetAllowance(a)
 	if err == nil {
```
DavidVorick: usually better to check for an explicit error, otherwise future changes to the code may cause this test to fail in a different way, resulting in silently lost testing coverage.

DavidVorick: Alternatively, you could have something that is 100% correct except for the Hosts value, and it's equally effective.

modules/renter/contractor/contractor_test.go
```diff
@@ -130,28 +132,128 @@ type stubHostDB struct{}
...
+	// set same allowance; should no-op
+	err = c.SetAllowance(a)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(c.contracts) != 1 {
+		t.Fatal("expected 1 contract, got", len(c.contracts))
+	}
+
+	// reannounce host on different IP (easier than creating a new host)
```
DavidVorick: This test will stop working once we switch to identifying hosts by public key. Probably worthwhile to make a `addiitonalHost` helper function - I'm guessing that thorough testing will want to be able to spin up many hosts in multiple different tests.

modules/renter/contractor/allowance.go
lukechampine: That's not an assumption, it's part of the intended behavior of `SetAllowance`: if the period hasn't changed, the contract length shouldn't be extended.

It would be easy to remove this, though, such that changing the allowance would always renew to `c.blockHeight + a.Period`. I wasn't sure if that was the right approach.

modules/renter/contractor/allowance.go
lukechampine: Initially I agreed, but after I started implementing it, I changed my mind, mostly because `case2` is not clear-cut the way `case1` is. `case1` can easily be named `managedFormAllowanceContracts`, but in `case2`, you need to both renew and potentially form new contracts. So `managedRenewAllowanceContracts` would be a misnomer, and I can't think of a better name.

Alternatively, you could further split things up. Something like:
```go
if !shouldRenew {
    return c.managedFormAllowanceContracts()
}
n, err := c.managedRenewAllowanceContracts()
if err != nil {
    return err
} else if n > 0 {
    return c.formContracts(n)
}
return nil
```
I think it's fine the way it is though.

lukechampine:
btw: contractor coverage increased from 62.7% to 84.7%.

modules/host/negotiaterenewcontract.go
lukechampine: Ran new tests with `<=`, confirmed that fc37f810b8cbf69c5ac7251eb3baaf1543245c68 fixes the problem.

lukechampine:
>I think that the difference between picking the amount of money you are spending and picking your own hosts is pretty huge. There are lots of reasons to have the renter do the host picking - the renter has a far better idea of which hosts are reliable, etc.

The idea is that a `/hosts` endpoint would do the heavy lifting in terms of choosing good hosts. It would also give you flexibility to choose `cheap`, `fast`, `best`, etc.

>I also don't think it makes sense to have the developer managing renew schedules. As we've seen from implementing it ourselves, it's a pretty difficult thing to get right.

Most of the difficulty lies in the error handling -- specifically, the fact that we can't just abort as soon as we get an error. We have to make decisions on behalf of the caller, like whether to form new contracts if renewal fails. And those decisions are opaque; the caller doesn't know how much money is being spent or why, and if something goes wrong, there's often nowhere to report the error except the log.


>We should shape the API based on the use cases of the developers who will be using the API, keeping in mind that we can always expose lower level functions like what the contractor does at a later date.

True, but we can't un-expose `SetAllowance`. And if we finalize it, how much of it gets covered by the compatibility promise? Certainly the input parameters and the return values, but what about the behavior? Are we bound to continue supporting the quirks of `SetAllowance` in perpetuity, or are we allowed to say `SetAllowance sets the maximum amount of coins the renter may spend in a given period; it may or may not form or renew contracts, and its specific effect on the user's wallet and contract set is subject to change`?


Anyway, as I've said, I'm ok with having high-level contract formation behavior. I think part of the problem is that we're conflating `SetAllowance` with `IntelligentlyFormContractsAndAutomaticallyRenewThem`. We should reframe the situation: the contractor intelligently form contracts and automatically renews them. That's its job. The allowance only serves to inform the contractor of its spending limit. The point being: `SetAllowance` should not be treated as a call that causes contracts to be formed or renewed. This is usually a side effect, but it is not the primary function. The documentation should make it clear that the contractor will form and renew contracts in an opaque manner. And likewise, the `SetAllowance` documentation should make no indication that it can control when contracts are formed or renewed, how much is spent on each contract, or who the contracts are made with. I think I'm fine having this sort of call finalized, though perhaps it should be renamed to `SetSettings` or something equally ambiguous, if the plan is to potentially expand its functionality.

---

The renter-contractor stack has a long way to go. It's pretty daunting. A big concern is that we still don't have any repair functionality. If too many hosts reject your renewal, you risk losing data. This is especially dangerous given the current `SetAllowance` behavior, since it will blindly form new contracts to replace failed renewals. That means that if all your renewals are rejected due to some default host setting (e.g. the new height wasn't high enough), it may form new contracts to replace every old contract, and boom, all your data is gone. How should this be handled?

DavidVorick:
> Most of the difficulty lies in the error handling -- specifically, the fact that we can't just abort as soon as we get an error. We have to make decisions on behalf of the caller, like whether to form new contracts if renewal fails. And those decisions are opaque; the caller doesn't know how much money is being spent or why, and if something goes wrong, there's often nowhere to report the error except the log.

And what to do if some succeed and others don't. When some succeed you've already blown funds on the ones that succeeded. To me the way forward there is pretty clear: keep making new contracts until you've got enough to sustain the files. I don't think that's an opaque decision. When you decide to create a set of contracts for X siacoins, it should complete the task.

> The idea is that a /hosts endpoint would do the heavy lifting in terms of choosing good hosts. It would also give you flexibility to prioritize by cheap, fast, best, etc., or to use a blacklist or whitelist.

You could use a `/hosts` endpoint to establish what sorts of filters get applied to the hostdb. I don't see a problem with adding that later.

> True, but we can't un-expose SetAllowance. And if we finalize it, how much of it gets covered by the compatibility promise?

I'm thinking that instead of `SetAllowance`, we should probably have something like a `RenterSettings` struct, because what we're really doing is trying to control the behavior of the renter the same ways that we try to control the behavior of the host. You're not really setting an allowance, you're establishing a behavior pattern for the renter to form contracts, renew contracts, and repair contracts. Then you can add a cheater-field that's something like `contractmanagement: automatic`, which gives us wiggle room to extend this functionality later. Then you can redact the contract management code simply by setting it to `manual` or `automatic-EU` or `automatic-IBM-needs-this-specific-behavior-pattern-that-we-programmed-for-them` or whatever.

And then it's pretty easy to deny calls to the contractor if contract management has not been set to `manual`.

> SetAllowance sets the maximum amount of coins the renter may spend in a given period; it may or may not form or renew contracts, and its specific effect on the user's wallet and contract set is subject to change

The promise of the automatic contract manager can be as simple as 'we'll make sure you can still upload and download, and that we won't spend more than X coins per month. All other behavior undefined'.

DavidVorick:
> The renter-contractor stack has a long way to go. It's pretty daunting. A big concern is that we still don't have any repair functionality. If too many hosts reject your renewal, you risk losing data. This is especially dangerous given the current SetAllowance behavior, since it will blindly form new contracts to replace failed renewals. That means that if all your renewals are rejected due to some default host setting (e.g. the new height wasn't high enough), it may form new contracts to replace every old contract, and boom, all your data is gone. How should this be handled?

You've got time before the existing contracts expire. You need to do a download and reupload. I don't think repair in the naive case is very difficult, it's just taxing to the bandwidth.

lukechampine:
anything left that needs fixing in this PR?
My plan is to first fix the desync problem, and then refactor the contractor to use `SetSettings` instead of `SetAllowance`. In doing so, I plan to move the contract formation code out of `SetSettings`. Instead, it will be managed by a background thread, which will simply receive signals from `SetSettings` and `ProcessConsensusChange` when it might need to perform some work.

Issue #1192, opened 2016-05-27
Gateway Security Vulnerability


DavidVorick:
The details of this security vulnerability are being kept private until the fix has been released. @VoidingWarranties is well aware of both the problem and the solution, and has already made significant steps in fixing the vulnerability.

PR #1221, opened 2016-06-03
Gateway version handshake

VoidingWarranties:
This PR moves the version handshake network code from `Connect` and `threadedAcceptConn` to `connectVersionHandshake` and `acceptConnVersionHandshake`, respectively. The primary reason for doing this is to be able to pass the version that is sent in the handshake as an argument to the function. This is useful as `build.Version` is a const and is difficult to mock.

This PR is a prerequisite for the testing required in #1192.

### DIFF
--- a/modules/gateway/peers.go
+++ b/modules/gateway/peers.go
@@ -67,6 +67,14 @@ func (s insufficientVersionError) Error() string {
 	return "unacceptable version: " + string(s)
 }
 
+// invalidVersionError indicates a peer's version is not a valid version number.
+type invalidVersionError string
+
+// Error implements the error interface for invalidVersionError.
+func (s invalidVersionError) Error() string {
+	return "invalid version: " + string(s)
+}
+
 type peer struct {
 	modules.Peer
 	sess muxado.Session
@@ -216,7 +224,7 @@ func (g *Gateway) threadedAcceptConn(conn net.Conn) {
 // acceptableVersion returns an error if the version is unacceptable.
 func acceptableVersion(version string) error {
 	if !build.IsVersion(version) {
-		return insufficientVersionError(version)
+		return invalidVersionError(version)
 	}
 	if build.VersionCmp(version, minAcceptableVersion) < 0 {
 		return insufficientVersionError(version)
--- a/modules/gateway/peers.go
+++ b/modules/gateway/peers.go
@@ -2,6 +2,7 @@ package gateway
 
 import (
 	"errors"
+	"fmt"
 	"net"
 	"time"
 
@@ -15,8 +16,16 @@ import (
 const (
 	// the gateway will not accept inbound connections above this threshold
 	fullyConnectedThreshold = 128
+
 	// the gateway will ask for more addresses below this threshold
 	minNodeListLen = 100
+
+	// minAcceptableVersion is the version below which the gateway will refuse to
+	// connect to peers and reject connection attempts.
+	//
+	// Reject peers < v0.4.0 as the previous version is v0.3.3 which is
+	// pre-hardfork.
+	minAcceptableVersion = "0.4.0"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


modules/gateway/peers.go
```diff
@@ -17,6 +18,15 @@ const (
...
 	// the gateway will ask for more addresses below this threshold
 	minNodeListLen = 100
+	// minAcceptableVersion is the version below which the gateway will refuse to
+	// connect to peers and reject connection attempts.
+	//
+	// Reject peers < v0.4.0 as the previous version is v0.3.3 which is
+	// pre-hardfork.
+	//
+	// NOTE: this version must be bumped whenever the gateway or consensus
+	// breaks compatibility.
```
DavidVorick: NOTE is not necessary, also I would prefer if there were newlines between the consts.

modules/gateway/peers.go
```diff
@@ -58,6 +68,14 @@ func (s insufficientVersionError) Error() string {
 	return "unacceptable version: " + string(s)
 }
 
+// invalidVersionError indicates a peer's version is not a valid version number.
+type invalidVersionError string
+
+// Error implements the error interface for invalidVersionError.
+func (s invalidVersionError) Error() string {
+	return "invalid version: " + string(s)
+}
```
DavidVorick: It seems to me like this error is only being used to be returned and for testing. I think it's probably okay to remove this type and just call `errors.New()` for most of what you are returning.

modules/gateway/peers.go
```diff
@@ -225,6 +222,61 @@ func (g *Gateway) threadedAcceptConn(conn net.Conn) {
...
+			return "", fmt.Errorf("failed to write reject: %v", err)
+		}
+		return "", err
+	}
+	// Send our version.
+	if err := encoding.WriteObject(conn, version); err != nil {
+		return "", fmt.Errorf("failed to write version: %v", err)
+	}
+	return remoteVersion, nil
+}
```
DavidVorick: These three functions look pretty good to me. I wish there was some way we could escape needing to pass in the version string as an argument, but it doesn't seem easily avoidable.

modules/gateway/peers.go
VoidingWarranties: The `insufficientVersionError` caught a few errors in testing in the past. I'd prefer to keep this type so as to make testing easier and catch similar errors in the future.

modules/gateway/peers.go
```diff
@@ -251,29 +303,11 @@ func (g *Gateway) Connect(addr modules.NetAddress) error {
...
-	// Reject peers < v0.4.0 as the previous version is v0.3.3 which is
-	// pre-hardfork.
-	//
-	// NOTE: this version must be bumped whenever the gateway or consensus
-	// breaks compatibility.
-	if build.VersionCmp(remoteVersion, "0.4.0") < 0 {
+	remoteVersion, err := connectVersionHandshake(conn, build.Version)
+	if err != nil {
+		// TODO: should this be muxado.Client(conn).Close()?
 		conn.Close()
```
VoidingWarranties: @lukechampine, do you know how `conn.Close()` differs from
```go
muxado.Client(conn).Close()
```

VoidingWarranties:
Updated const comments and spacing and rebased.

modules/gateway/peers.go
VoidingWarranties: The error returned by `session.Accept()` on `Close` is the same whether the connection is closed with `conn.Close()` or `muxado.Client(conn).Close()`. Since we'll be switching to YAMUX / SPDY / HTTP2 soon anyways it is probably easier to continue closing with `conn.Close()`.

lukechampine: `Client.Close` sends a "GoAway" frame. It's basically just a more graceful way of shutting down a muxado connection. During `Connect`, you haven't agreed to be muxado peers yet, so a normal `conn.Close` is fine.

VoidingWarranties: Ah okay. Currently peer rejection signaling is one way (only a peer receiving a connection request can send a rejection signal, by sending "reject" as its version) so the peer initiating the connection might have `addPeer` on the peer before `threadedAcceptConn` has a chance to disconnect. Regardless I'll stick with `conn.Close`.

PR #1223, opened 2016-06-05
Finalize gateway API

VoidingWarranties:
add & remove isn't as clear as connect & disconnect, especially since the peer remains in the node list on disconnect. add & remove could be used in the future to add and remove addresses from the node list.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -115,8 +115,11 @@ func (srv *Server) initAPI(password string) {
 	// Gateway API Calls
 	if srv.gateway != nil {
 		router.GET("/gateway", srv.gatewayHandler)
-		router.POST("/gateway/add/:netaddress", requirePassword(srv.gatewayAddHandler, password))
-		router.POST("/gateway/remove/:netaddress", requirePassword(srv.gatewayRemoveHandler, password))
+		router.POST("/gateway/connect/:netaddress", requirePassword(srv.gatewayConnectHandler, password))
+		router.POST("/gateway/disconnect/:netaddress", requirePassword(srv.gatewayDisconnectHandler, password))
+		// COMPATv0.6.0
+		router.POST("/gateway/add/:netaddress", requirePassword(srv.gatewayConnectHandler, password))
+		router.POST("/gateway/remove/:netaddress", requirePassword(srv.gatewayDisconnectHandler, password))
 	}
 
 	// Host API Calls
--- a/api/api.go
+++ b/api/api.go
@@ -117,9 +117,6 @@ func (srv *Server) initAPI(password string) {
 		router.GET("/gateway", srv.gatewayHandler)
 		router.POST("/gateway/connect/:netaddress", requirePassword(srv.gatewayConnectHandler, password))
 		router.POST("/gateway/disconnect/:netaddress", requirePassword(srv.gatewayDisconnectHandler, password))
-		// COMPATv0.6.0
-		router.POST("/gateway/add/:netaddress", requirePassword(srv.gatewayConnectHandler, password))
-		router.POST("/gateway/remove/:netaddress", requirePassword(srv.gatewayDisconnectHandler, password))
 	}
 
 	// Host API Calls
--- a/doc/API.md
+++ b/doc/API.md
@@ -263,11 +263,11 @@ Gateway
 
 Queries:
 
-* /gateway                     [GET]
-* /gateway/add/{netaddress}    [POST]
-* /gateway/remove/{netaddress} [POST]
+* /gateway                         [GET]
+* /gateway/connect/{netaddress}    [POST]
+* /gateway/disconnect/{netaddress} [POST]
 
-#### /gateway
+#### /gateway [GET]
 
 Function: Returns information about the gateway, including the list of peers.
 
@@ -290,22 +290,24 @@ address and the port Sia is listening on.
 'peers' is a list of the network addresses and versions of peers that the
 Gateway is currently connected to.
 
-#### /gateway/add/{netaddress} [POST]
+#### /gateway/connect/{netaddress} [POST]
 
-Function: Adds a peer to the gateway.
+Function: Connects the gateway to a peer. The peer is added to the node list if
+it not already present.
 
 Parameters:
 ```
 netaddress string
 ```
-'netaddress' should be a reachable hostname + port number, typically of the
+'netaddress' should be a reachable ip + port number, typically of the
 form "a.b.c.d:xxxx".
 
 Response: standard
 
-#### /gateway/remove/{netaddress} [POST]
+#### /gateway/disconnect/{netaddress} [POST]
 
-Function: Will remove a peer from the gateway.
+Function: Disconnects the gateway from the peer. The peer remains in the node
+list.
 
 Parameters:
 ```
--- a/doc/API.md
+++ b/doc/API.md
@@ -27,6 +27,17 @@ Example GET curl call:  `curl -A "Sia-Agent" /wallet/transactions?startheight=1&
 
 Example POST curl call: `curl -A "Sia-Agent" --data "amount=123&destination=abcd" /wallet/siacoins
 
+Table of contents:
+- [Daemon](#daemon)
+- [Consensus](#consensus)
+- [Explorer](#explorer)
+- [Gateway](#gateway)
+- [Host](#host)
+- [Miner](#miner)
+- [Renter](#renter)
+- [Transaction Pool](#transaction-pool)
+- [Wallet](#wallet)
+
 Daemon
 ------
 
--- a/doc/API.md
+++ b/doc/API.md
@@ -287,62 +287,66 @@ provided hash.
 Gateway
 -------
 
-Queries:
+| Route | HTTP verb |
+| ----- | --------- |
+| [/gateway](#gateway-get) | GET |
+| [/gateway/connect/{netaddress}](#gatewayconnectnetaddress-post) | POST |
+| [/gateway/disconnect/{netaddress}](#gatewaydisconnectnetaddress-post) | POST |
 
-* /gateway                         [GET]
-* /gateway/connect/{netaddress}    [POST]
-* /gateway/disconnect/{netaddress} [POST]
+For examples and detailed descriptions of request and response parameters,
+refer to [Gateway.md](/doc/api/Gateway.md).
 
 #### /gateway [GET]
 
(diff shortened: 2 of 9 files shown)


siac/gatewaycmd.go
```diff
@@ -14,22 +14,36 @@ var (
...
+		Short: "Disconnect from a peer",
+		Long:  "Disconnect from a peer. Does not remove the peer from the node list.",
+		Run:   wrap(gatewaydisconnectcmd),
+	}
+
+	// COMPATv0.6.0
+	gatewayDeprecatedRemoveCmd = &cobra.Command{
+		Use:        "remove [address]",
+		Deprecated: "use `siac gateway disconnect` instead.",
+		Run:        wrap(gatewaydisconnectcmd),
```
lukechampine: I think you can use an alias here, like we do for `renter ls`

VoidingWarranties: If we want to remove the `remove` command eventually we should deprecate it instead of aliasing it so that the user is warned that its deprecated when they try to use it.

api/api.go
```diff
@@ -115,8 +115,11 @@ func (srv *Server) initAPI(password string) {
 	// Gateway API Calls
 	if srv.gateway != nil {
 		router.GET("/gateway", srv.gatewayHandler)
-		router.POST("/gateway/add/:netaddress", requirePassword(srv.gatewayAddHandler, password))
-		router.POST("/gateway/remove/:netaddress", requirePassword(srv.gatewayRemoveHandler, password))
+		router.POST("/gateway/connect/:netaddress", requirePassword(srv.gatewayConnectHandler, password))
+		router.POST("/gateway/disconnect/:netaddress", requirePassword(srv.gatewayDisconnectHandler, password))
+		// COMPATv0.6.0
+		router.POST("/gateway/add/:netaddress", requirePassword(srv.gatewayConnectHandler, password))
+		router.POST("/gateway/remove/:netaddress", requirePassword(srv.gatewayDisconnectHandler, password))
```
DavidVorick: When we switch to 1.0 we should remove all COMPAT calls in the api

DavidVorick:
As a part of this pull request, I would like to see a file created in `doc/api` called `Gateway.md` which details all of the gateway calls and what they do. They should document all of the input fields, all of the output fields, and as you create this file you should spellcheck everything.

Don't forget to spellcheck objects that get used. For example, in the host there's the `HostInternalSettings` object, which has JSON that is unnamed in the api package. To spellcheck the API's `HostInternalSettings` related calls, you'd have to look at the code in the `modules` package where `HostInternalSettings` is defined.

VoidingWarranties:
How do you want `doc/api/Gateway.md` to be different from the gateway section in `doc/API.md`?

DavidVorick:
Sorry for not answering the question earlier. In an attempt to finalize the host api, I've got an example, checkout the documentation changes in #1234 

I'm not particularly attached to any method, except that I insist that the whole project be consistent.

There are a few major differences to the way that the host api doc has been structured:

+ Provide a list of host routes in API.md, along with a list of fields and parameters, but not the full description or explanation.
+ Link to the full description
+ Put the full description in doc/api/Host.md instead of in doc/API.md
+ Explain what a field does inline instead of explaining it at the bottom. The documentation for a field is now closer to the field, which hopefully improves readability.

A table of contents sounds reasonable.

The way I've set up the host means that there's one reference sheet with all of the routes and their fields, but it's very dense (I think this is a good thing). The full descriptions, if you need them, are available in separate files.

Benefits: breaks out a full host.md description,

VoidingWarranties:
Updated the gateway section of `doc/API.md` and added `doc/api/Gateway.md`. I thought a lot about the layout of these, so please let me know if you have any thoughts.

VoidingWarranties:
Notably the structs have changed from Golang syntax to JSON (really javascript) syntax. Clients of the API are unlikely to be using Golang, and the API docs should be language agnostic.

Also thanks to @johnathanhowell for providing feedback offline.

api/api.go
DavidVorick: This PR is intended to be a part of preparing for 1.0... remove these calls :P

doc/API.md
```diff
@@ -27,6 +27,32 @@ Example GET curl call:  `curl -A "Sia-Agent" /wallet/transactions?startheight=1&
...
+
+#### Success
+
+The standard response indicating the request was successfully processed is HTTP
+status code 204.
+
+#### Error
+
+The standard error response indicating the request failed for any reason, is a
+4xx HTTP status code with the error message as plain text.
```
DavidVorick: I believe that the error message should be returned in json, that way additional objects can be added for compat support or other weirdness.

This isn't the link I was thinking of, but it does talk about error design:  https://apigee.com/about/blog/technology/restful-api-design-what-about-errors

doc/API.md
```diff
@@ -27,6 +27,32 @@ Example GET curl call:  `curl -A "Sia-Agent" /wallet/transactions?startheight=1&
...
+- [Daemon](#daemon)
+- [Consensus](#consensus)
+- [Explorer](#explorer)
+- [Gateway](#gateway)
+- [Host](#host)
+- [Miner](#miner)
+- [Renter](#renter)
+- [Transaction Pool](#transaction-pool)
+- [Wallet](#wallet)
+
```
DavidVorick: Table of contents looks good

doc/API.md
```diff
@@ -261,60 +287,66 @@ provided hash.
...
-	peers      []struct {
-                netaddress string
-                version    string
-                inbound    bool
-        }
+| Route | HTTP verb |
+| ----- | --------- |
+| [/gateway](#gateway-get) | GET |
+| [/gateway/connect/{netaddress}](#gatewayconnectnetaddress-post) | POST |
+| [/gateway/disconnect/{netaddress}](#gatewaydisconnectnetaddress-post) | POST |
```
DavidVorick: Ideally this would be cleaner in markdown, can you align the bars? I believe it will still format correctly.

doc/api/Gateway.md
```diff
@@ -0,0 +1,163 @@
...
+
+        // inbound is true when the peer initiated the connection. This field
+        // is exposed as outbound peers are generally trusted more than inbound
+        // peers, as inbound peers are easily manipulated by an adversary.
+        "inbound":    Boolean
+    }
+}
+```
+
+###### [Example](#gateway-info-example)
```
DavidVorick: The doc is really busy. I like most of the information, I like the examples and stuff, but there are a lot of elements and it's confusing to navigate.

One of the things that might help is to move the link to the example up to the line with the route definition. e.g.

`#### /gateway [GET] [(Example)](#gateway-info-example)`

DavidVorick: 2016-06-15-200926_1789x1320_scrot

DavidVorick: line 1: bold
line 2:  normal, with `code` at the front
line 3: faded bold
line 4-12: block
line 13: link
line 14: bold
line 15-16: normal, leading `with code`
line 17: faded bold
line 18 `code`
line 19: faded bold
line 20: normal
line 21: link

In 21 lines you've got 12 format changes. That's really difficult for me to read. 

From what I can tell, each route currently has a maximum of 7 different style changes:

route
route description
parameters title
parameters javascript
response title
response javascript / text
example

Ideally we could get this down to 3 or 4. I'm not sure we can get that low, but maybe you've got ideas. Perhaps:

route + example link
description
parameters javascript
response javascript

doc/api/Gateway.md
```diff
@@ -0,0 +1,163 @@
...
+| [/gateway](#gateway-get) | GET |
+| [/gateway/connect/{netaddress}](#gatewayconnectnetaddress-post) | POST |
+| [/gateway/disconnect/{netaddress}](#gatewaydisconnectnetaddress-post) | POST |
+
+Examples
+--------
+
+* [gateway info](#gateway-info-example)
+* [connecting to a peer](#connect-example)
+* [disconnecting from a peer](#disconnect-example)
```
DavidVorick: If you are planning on having an example for every route, you probably should not link to them all. For the gateway it's not so bad because there are only three, but it does add to the cluttered feeling.

Or maybe you can add an example field to the index table

```
| Route | HTTP verb | Example |
| ----- | --------- | ------- |
| [/gateway](#gateway-get) | GET | [gateway info](#gateway-info-example) |
| [/gateway/connect/{netaddress}](#gatewayconnectnetaddress-post) | POST | [connecting to a peer](#connect-example) |
| [/gateway/disconnect/{netaddress}](#gatewaydisconnectnetaddress-post) | POST | [disconnecting from a peer](#disconnect-example)| 
```

siac/main.go
```diff
@@ -238,10 +238,14 @@ func main() {
...
 
 	root.AddCommand(gatewayCmd)
-	gatewayCmd.AddCommand(gatewayAddCmd, gatewayRemoveCmd, gatewayAddressCmd, gatewayListCmd)
+	gatewayCmd.AddCommand(gatewayConnectCmd, gatewayDisconnectCmd, gatewayAddressCmd, gatewayListCmd)
 
 	root.AddCommand(consensusCmd)
 
+	// COMPATv0.6.0
+	gatewayCmd.AddCommand(gatewayDeprecatedAddCmd)
+	gatewayCmd.AddCommand(gatewayDeprecatedRemoveCmd)
```
DavidVorick: remove compatibility code. 1.0 is brutal :)

DavidVorick: It's fine to do it this way because people are expecting a moderate to substantial change with this release.

DavidVorick:
Overall the documentation looks good, there's a lot of good information and I like that we now have examples, but the way that the information is presented is still pretty cluttered and confusing.

Given that we're working in markdown, there's only so much we can do to address that. It's probably going to be cluttered no matter what. But hopefully we can get a step or two better than where it's at.

Either way, definitely a big improvement over where it was before.

doc/API.md
VoidingWarranties: Here's a long discussion about JSON error design. https://github.com/json-api/json-api/issues/7

doc/api/Gateway.md
VoidingWarranties: I think there's an advantage of having the example link next to each route description (or in the title), as a developer is more likely to want to look at an example when they are looking through the request or response specs.

VoidingWarranties:
Updated with better styling, removed COMPAT code, and JSON errors.

PR #1224, opened 2016-06-07
Clean shutdown of miner

mnsl:
Adds `tg *siasync.ThreadGroup` field to the Miner type; m.tg.Stop() is called at the start of m.Close(), automatically unsubscribing the miner from its consensus set and transaction pool and blocking until the m.threadedMine() goroutine is done.

To unsubscribe the miner from the transaction pool, I've added an exported Unsubscribe method to TransactionPool (akin to the ConsensusSet Unsubscribe method).

### DIFF
--- a/api/miner_test.go
+++ b/api/miner_test.go
@@ -73,7 +73,7 @@ func TestIntegrationMinerStartStop(t *testing.T) {
 		t.Fatal(err)
 	}
 	if !mg.CPUMining {
-		t.Error("cpu is not reporting through the api that it is mining.")
+		t.Error("cpu is not reporting through the api that it is mining")
 	}
 
 	// Stop the cpu miner and wait for the stop call to go through.
@@ -83,7 +83,7 @@ func TestIntegrationMinerStartStop(t *testing.T) {
 	}
 	time.Sleep(100 * time.Millisecond)
 	if st.server.miner.CPUMining() {
-		t.Error("cpu miner is reporting that it is not on")
+		t.Error("cpu miner is reporting that it is on after being stopped")
 	}
 
 	// Check the numbers through the status api call.
@@ -92,7 +92,7 @@ func TestIntegrationMinerStartStop(t *testing.T) {
 		t.Fatal(err)
 	}
 	if mg.CPUMining {
-		t.Error("cpu is not reporting through the api that it is mining.")
+		t.Error("cpu is not reporting through the api that it is mining")
 	}
 }
 
--- a/modules/miner/blockmanager.go
+++ b/modules/miner/blockmanager.go
@@ -69,6 +69,11 @@ func (m *Miner) newSourceBlock() {
 // it is typically safe to assume that headers will be remembered for
 // min(10 minutes, 10e3 requests).
 func (m *Miner) HeaderForWork() (types.BlockHeader, types.Target, error) {
+	if err := m.tg.Add(); err != nil {
+		return types.BlockHeader{}, types.Target{}, err
+	}
+	defer m.tg.Done()
+
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
@@ -158,6 +163,11 @@ func (m *Miner) managedSubmitBlock(b types.Block) error {
 
 // SubmitHeader accepts a block header.
 func (m *Miner) SubmitHeader(bh types.BlockHeader) error {
+	if err := m.tg.Add(); err != nil {
+		return err
+	}
+	defer m.tg.Done()
+
 	// Because a call to managedSubmitBlock is required at the end of this
 	// function, the first part needs to be wrapped in an anonymous function
 	// for lock safety.
--- a/modules/miner/cpuminer.go
+++ b/modules/miner/cpuminer.go
@@ -2,11 +2,18 @@ package miner
 
 import (
 	"time"
+
+	"github.com/NebulousLabs/Sia/build"
 )
 
 // threadedMine starts a gothread that does CPU mining. threadedMine is the
 // only function that should be setting the mining flag to true.
 func (m *Miner) threadedMine() {
+	if err := m.tg.Add(); err != nil {
+		return
+	}
+	defer m.tg.Done()
+
 	// There should not be another thread mining, and mining should be enabled.
 	m.mu.Lock()
 	if m.mining || !m.miningOn {
@@ -54,6 +61,11 @@ func (m *Miner) threadedMine() {
 
 // CPUHashrate returns an estimated cpu hashrate.
 func (m *Miner) CPUHashrate() int {
+	if err := m.tg.Add(); err != nil {
+		build.Critical(err)
+	}
+	defer m.tg.Done()
+
 	m.mu.Lock()
 	defer m.mu.Unlock()
 	return int(m.hashRate)
@@ -61,6 +73,11 @@ func (m *Miner) CPUHashrate() int {
 
 // CPUMining indicates whether the cpu miner is running.
 func (m *Miner) CPUMining() bool {
+	if err := m.tg.Add(); err != nil {
+		build.Critical(err)
+	}
+	defer m.tg.Done()
+
 	m.mu.Lock()
 	defer m.mu.Unlock()
 	return m.miningOn
@@ -69,6 +86,11 @@ func (m *Miner) CPUMining() bool {
 // StartCPUMining will start a single threaded cpu miner. If the miner is
 // already running, nothing will happen.
 func (m *Miner) StartCPUMining() {
+	if err := m.tg.Add(); err != nil {
+		build.Critical(err)
+	}
+	defer m.tg.Done()
+
 	m.mu.Lock()
 	defer m.mu.Unlock()
 	m.miningOn = true
@@ -78,6 +100,11 @@ func (m *Miner) StartCPUMining() {
 // StopCPUMining will stop the cpu miner. If the cpu miner is already stopped,
 // nothing will happen.
 func (m *Miner) StopCPUMining() {
+	if err := m.tg.Add(); err != nil {
+		build.Critical(err)
+	}
+	defer m.tg.Done()
+
 	m.mu.Lock()
 	defer m.mu.Unlock()
 	m.hashRate = 0
--- a/modules/miner/cpuminer.go
+++ b/modules/miner/cpuminer.go
@@ -27,8 +27,15 @@ func (m *Miner) threadedMine() {
 	// occurring.
 	cycleStart := time.Now()
 	for {
-		// Kill the thread if mining has been turned off.
 		m.mu.Lock()
+		// Kill the thread if the miner's threadgroup is stopped.
+		if m.tg.IsStopped() {
+			m.miningOn = false
+			m.mining = false
+			m.mu.Unlock()
+			return
+		}
+		// Kill the thread if mining has been turned off.
 		if !m.miningOn {
 			m.mining = false
 			m.mu.Unlock()
--- a/modules/miner/miner.go
+++ b/modules/miner/miner.go
@@ -206,6 +206,7 @@ func (m *Miner) Close() error {
 	defer m.mu.Unlock()
 
 	m.cs.Unsubscribe(m)
+	m.tpool.Unsubscribe(m)
 
 	var errs []error
 	if err := m.saveSync(); err != nil {
--- a/modules/miner/miner.go
+++ b/modules/miner/miner.go
@@ -10,6 +10,7 @@ import (
 	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/persist"
+	siasync "github.com/NebulousLabs/Sia/sync"
 	"github.com/NebulousLabs/Sia/types"
 )
 
@@ -104,6 +105,9 @@ type Miner struct {
 	mu         sync.RWMutex
 	persist    persistence
 	persistDir string
+	// tg signals the Miner's goroutines to shut down and blocks until all
+	// goroutines have exited before returning from Close().
+	tg siasync.ThreadGroup
 }
 
 // startupRescan will rescan the blockchain in the event that the miner
@@ -194,6 +198,10 @@ func New(cs modules.ConsensusSet, tpool modules.TransactionPool, w modules.Walle
 // Close terminates all ongoing processes involving the miner, enabling garbage
 // collection.
 func (m *Miner) Close() error {
+	if err := m.tg.Stop(); err != nil {
+		return err
+	}
+
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
--- a/modules/miner/miner.go
+++ b/modules/miner/miner.go
@@ -235,6 +235,11 @@ func (m *Miner) checkAddress() error {
 // BlocksMined returns the number of good blocks and stale blocks that have
 // been mined by the miner.
 func (m *Miner) BlocksMined() (goodBlocks, staleBlocks int) {
+	if err := m.tg.Add(); err != nil {
+		build.Critical(err)
+	}
+	defer m.tg.Done()
+
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
--- a/modules/transactionpool/subscribe_test.go
+++ b/modules/transactionpool/subscribe_test.go
@@ -13,7 +13,7 @@ type mockSubscriber struct {
 }
 
 // ReceiveUpdatedUnconfirmedTransactions receives transactinos from the
-// transaction pool and stores them in the order they were received. 
+// transaction pool and stores them in the order they were received.
 // This method allows *mockSubscriber to satisfy the
 // modules.TransactionPoolSubscriber interface.
 func (ms *mockSubscriber) ReceiveUpdatedUnconfirmedTransactions(txns []types.Transaction, _ modules.ConsensusChange) {
@@ -28,7 +28,7 @@ func TestSubscription(t *testing.T) {
 		t.Skip()
 	}
 
-	tpt, err  := createTpoolTester("TestUnsubscribe")
+	tpt, err := createTpoolTester("TestUnsubscribe")
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -63,11 +63,10 @@ func TestSubscription(t *testing.T) {
 	if len(ms.txns) != numTxns {
 		t.Errorf("mock subscriber should've received %v transactions; received %v instead", numTxns, len(ms.txns))
 	}
-	
+
 	numSubscribers := len(tpt.tpool.subscribers)
 	tpt.tpool.Unsubscribe(&ms)
 	if len(tpt.tpool.subscribers) != numSubscribers-1 {
 		t.Error("transaction pool failed to unsubscribe mock subscriber")
 	}
 }
-
--- a/modules/transactionpool/subscribe_test.go
+++ b/modules/transactionpool/subscribe_test.go
@@ -1,0 +1,73 @@
+package transactionpool
+
+import (
+	"github.com/NebulousLabs/Sia/modules"
+	"github.com/NebulousLabs/Sia/types"
+	"testing"
+)
+
+// mockSubscriber receives transactions from the transaction pool it is
+// subscribed to, retaining them in the order they were received.
+type mockSubscriber struct {
+	txns []types.Transaction
+}
+
+// ReceiveUpdatedUnconfirmedTransactions receives transactinos from the
+// transaction pool and stores them in the order they were received. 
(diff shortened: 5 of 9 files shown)


modules/miner/miner.go
```diff
@@ -104,6 +105,9 @@ type Miner struct {
 	mu         sync.RWMutex
 	persist    persistence
 	persistDir string
+	// tg signals the Miner's goroutines to shut down and blocks until
+	// all goroutines have exited before returning from Close().
+	tg *siasync.ThreadGroup
```
VoidingWarranties: If you don't use a pointer you don't need to use `new`

modules/miner/miner.go
```diff
@@ -194,18 +210,21 @@ func New(cs modules.ConsensusSet, tpool modules.TransactionPool, w modules.Walle
 // Close terminates all ongoing processes involving the miner, enabling garbage
 // collection.
 func (m *Miner) Close() error {
+	if err := m.tg.Stop(); err != nil {
+		return err
+	}
+
 	m.mu.Lock()
 	defer m.mu.Unlock()
```
VoidingWarranties: Only a read lock is necessary here, and only while calling `saveSync`. I don't think it matters much though.

modules/miner/miner.go
```diff
@@ -180,8 +184,20 @@ func New(cs modules.ConsensusSet, tpool modules.TransactionPool, w modules.Walle
 		return nil, errors.New("miner subscription failed: " + err.Error())
 	}
 
+	// Automatically unsubscribe the miner from its transaction pool
```
DavidVorick: I feel like there should be no newline here because the unsubscribing of the consensus set is directly paired to the subscribing to the consensus set.

DavidVorick:
You need to add testing for the unsubscription from the transaction pool.

modules/miner/miner.go
DavidVorick: No you want a `Lock` when closing because you are shutting down resources.

DavidVorick:
The exported functions need to return an error if they are called after the miner has been closed

modules/miner/miner.go
DavidVorick: oh, also I think the miner lock is just a normal lock. No readlock because the miner uses the lock a lot and I wanted it to be fast.

VoidingWarranties:
Can you close the corresponding issue in one of the commit messages so that it's automatically closed?

mnsl:
Latest changes address code review comments:
* add Unsubscribe method to TransactionPool in modules/transactionpool/subscribe.go
* add `modules/transactionpool/subscribe_test.go` file with `mockSubscriber` interface and `TestUnsubscribe` function
* introduce miner.errClosed error 
* make exported `*Miner` methods write `build.Critical(errClosed)` when the miner is closed

api/miner_test.go
```diff
@@ -35,10 +35,12 @@ func TestIntegrationMinerGET(t *testing.T) {
 	if mg.StaleBlocksMined != staleBlocksMined {
 		t.Error("stale blocks mined is incorrect")
 	}
-	if mg.CPUHashrate != st.server.miner.CPUHashrate() {
+	CPUHashrate := st.server.miner.CPUHashrate()
+	if mg.CPUHashrate != CPUHashrate {
```
mnsl: Oops, this is a holdover from when I was making miner.CPUMining() return an error.  Will return to original in next commit.

api/miner_test.go
```diff
@@ -35,10 +35,12 @@ func TestIntegrationMinerGET(t *testing.T) {
...
 		t.Error("mismatched cpu hashrate")
 	}
-	if mg.CPUMining != st.server.miner.CPUMining() {
+	CPUMining := st.server.miner.CPUMining()
+	if mg.CPUMining != CPUMining {
```
mnsl: same as above

api/miner_test.go
```diff
@@ -62,7 +64,8 @@ func TestIntegrationMinerStartStop(t *testing.T) {
 		t.Fatal(err)
 	}
 	time.Sleep(100 * time.Millisecond)
-	if !st.server.miner.CPUMining() {
+	CPUMining := st.server.miner.CPUMining()
+	if !CPUMining {
```
mnsl: same as above

api/miner_test.go
```diff
@@ -82,8 +85,9 @@ func TestIntegrationMinerStartStop(t *testing.T) {
 		t.Fatal(err)
 	}
 	time.Sleep(100 * time.Millisecond)
-	if st.server.miner.CPUMining() {
-		t.Error("cpu miner is reporting that it is not on")
+	CPUMining = st.server.miner.CPUMining()
+	if CPUMining {
+		t.Error("cpu miner is reporting that it is on after being stopped")
```
mnsl: same as above

api/miner_test.go
```diff
@@ -83,7 +83,7 @@ func TestIntegrationMinerStartStop(t *testing.T) {
 	}
 	time.Sleep(100 * time.Millisecond)
 	if st.server.miner.CPUMining() {
-		t.Error("cpu miner is reporting that it is not on")
+		t.Error("cpu miner is reporting that it is on after being stopped")
```
mnsl: pretty sure this error message was wrong

modules/transactionpool/subscribe_test.go
```diff
@@ -0,0 +1,65 @@
+package transactionpool
+
+import (
+	"github.com/NebulousLabs/Sia/modules"
+	"github.com/NebulousLabs/Sia/types"
+	"testing"
+)
+
```
mnsl: Sorry, forgot to add comments.  Will fix pronto

modules/miner/cpuminer.go
```diff
@@ -2,11 +2,18 @@ package miner
...
 )
 
 // threadedMine starts a gothread that does CPU mining. threadedMine is the
 // only function that should be setting the mining flag to true.
 func (m *Miner) threadedMine() {
+	if err := m.tg.Add(); err != nil {
+		return
+	}
+	defer m.tg.Done()
+
```
mnsl: Should this occur before or after `m.mu.Lock()`?

lukechampine: the lock does not protect the threadgroup, so it can be called with or without a lock. Generally these should be the first 4 lines of each tracked function.

mnsl: Including functions that don't call goroutines?

VoidingWarranties: These should also be the first four lines of exported methods that use closable resources, such as printing to the log.

modules/miner/cpuminer.go
```diff
@@ -56,13 +63,19 @@ func (m *Miner) threadedMine() {
 func (m *Miner) CPUHashrate() int {
 	m.mu.Lock()
 	defer m.mu.Unlock()
+	if m.tg.IsStopped() {
+		build.Critical(errClosed)
+	}
```
VoidingWarranties: This can be the first three lines, before the lock.

mnsl:
Latest updates:
* move `if m.tg.IsStopped()` checks outside of lock
* add `if m.tg.Add() .... defer m.tg.Done()` to exported functions that use closable resources

mnsl:
This PR and #1230 are both failing due to the same [data race](https://travis-ci.org/NebulousLabs/Sia/jobs/136316860#L581) in the gateway module involving `gateway.(*Gateway).threadedListenPeer` and `gateway.(*Gateway).threadedAcceptConn`.  It's nondeterministic (about half the time).

modules/miner/miner.go
mnsl: Is it OK that the `m.tg.stopFns` (`m.cs.Unsubscribe` and `m.tpool.Unsubscribe`) are called outside of the lock?

VoidingWarranties:
The data race is in the version of muxado we use. It should be fixed when #1231 is complete.

modules/miner/miner.go
lukechampine: @VoidingWarranties for future reference, `saveSync` should be called with a write lock. The persist file should be treated as an extension of the object's state, which means you need a write lock when writing to it. This makes more sense if you frame it in terms of "only one goroutine should write to the file at a time."

modules/transactionpool/subscribe_test.go
```diff
@@ -0,0 +1,78 @@
...
+// subscribed to, retaining them in the order they were received.
+type mockSubscriber struct {
+	txns []types.Transaction
+}
+
+// newMockSubscriber returns a blank mockSubscriber that is ready to subscribe
+// to a transaction pool.
+func newMockSubscriber() mockSubscriber {
+	return mockSubscriber{}
+}
```
lukechampine: constructors should only be used when some initialization needs to be performed.

modules/transactionpool/subscribe_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+	// Create a mock subscriber and subscribe it to the transaction pool.
+	ms := newMockSubscriber()
+	tpt.tpool.TransactionPoolSubscribe(&ms)
+	if len(ms.txns) != 0 {
+		t.Fatalf("mock subscriber has received %v transactions; shouldn't have received any yet", len(ms.txns))
+	}
+
+	// Create a valid transaction set and check that the mock subscriber's
+	// transaction list is updated.
```
lukechampine: seems like this tests more than just `Unsubscribe`, which is fine -- just change the name to `TestSubscription` (or similar) and update the docstring.

modules/miner/blockmanager.go
```diff
@@ -69,6 +70,9 @@ func (m *Miner) newSourceBlock() {
 // it is typically safe to assume that headers will be remembered for
 // min(10 minutes, 10e3 requests).
 func (m *Miner) HeaderForWork() (types.BlockHeader, types.Target, error) {
+	if m.tg.IsStopped() {
+		build.Critical(errClosed)
+	}
```
DavidVorick: `HeaderForWork` has an error in the function definition, you should be returning `errMinerClosed` here.

modules/miner/blockmanager.go
```diff
@@ -158,6 +162,14 @@ func (m *Miner) managedSubmitBlock(b types.Block) error {
 
 // SubmitHeader accepts a block header.
 func (m *Miner) SubmitHeader(bh types.BlockHeader) error {
+	if err := m.tg.Add(); err != nil {
```
DavidVorick: `IsStopped` check should happen before you add to the `tg`, and since this function returns an error you should return `errMinerClosed` if the miner is closed.

modules/miner/cpuminer.go
DavidVorick: @lukechampine not quite correct.

`threaded` indicates that this function should always be called as a goroutine (e.g. `go threadedMine()`), which means that it's possible the miner closes between the time this function is called and the time that this function executes.

So, the first lines of code here should actually be to check whether the miner is closed or not. @mnsl it's not a critical error if this function gets called after the miner is closed - it's possible during normal use for this function to be called, and then for the miner to close, and then for this function to execute. So if the miner is closed when this function starts, you should just `return`, silently refusing to start the miner.

DavidVorick: There's a second bit of trickiness related to this function, and I'm not sure that the testing coverage is complete enough to detect problems. I'm pretty sure that if you were to call `Close` right now while the cpu miner was running, it'd never actually shut down. You should write a test to verify this.

`Stop` cannot return until `m.tg.Done()` has returned, which won't happen until after this function terminates. And this function won't terminate if `m.miningOn`. The simplest fix is to change the mining statement to `if m.mining || !m.miningOn || m.tg.IsStopped() {`, and then you'll also want to add the line `m.miningOn = false` to the `Close` function.

modules/miner/miner.go
```diff
@@ -229,6 +240,10 @@ func (m *Miner) BlocksMined() (goodBlocks, staleBlocks int) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
+	if m.tg.IsStopped() {
+		build.Critical(errClosed)
+	}
```
DavidVorick: you can throw this above the `Lock` call.

DavidVorick:
I think there's a problem here. Imagine that the ConsensusSet has a new block and is locked at the moment that `miner.Close` is called. ConsensusSet is not going to let you call `Unsubscribe` until it has finished processing the block, which includes calling `ProcesseConsensusChange` on the miner. Depending on what is called in the `Stop` function, you may lose resources before `ProcessConsnesusChange` is called. `IsStopped` is set immediately, so if you are checking `IsStopped` during `ProcessConsensusChange` you are going to potentially have problems.

One option is to call `Unsubscribe` before you call `Stop`. And the other is to call `Unsubscribe` as one of the first calls inside of the `Stop` function. Still thinking about other options, but it seems to make things pretty messy. There's probably a clean way to deal with all of this, just not sure yet.

The problem seems unique to `Unsubscribe` functions, as that should be the only situation where a function might be accessing the miner after `Close`. I think it's cleanly solved by making sure subscription is the last thing that happens during startup, and the first thing that happens during shutdown. The issue there is that then the `ProcessConsensusChange` functions either shouldn't check whether the miner is stopped, or they should just close silently (without calling `build.Critical`) if the miner is stopped, because unless you call `Unsubscribe` before `IsStopped` gets set to `true`, the function may get confused.

@lukechampine what do you think?

modules/miner/blockmanager.go
lukechampine: The typical pattern is to check the return value of `tg.Add`. If you used `IsStopped` first, you'd still need to check the return value of `tg.Add` anyway. See https://github.com/NebulousLabs/Sia/blob/master/modules/gateway/peers.go#L142

modules/miner/miner.go
lukechampine: shouldn't this be a `tg.Add`?

lukechampine:
Because `Subscribe` is a source of new threads, `Unsubscribe` should be called as a stopFn. This is in line with the strategy of eliminating new threads before proceeding with shutdown.

DavidVorick:
If it's called as a stopFn there's a deadlock risk that you have to consider as a developer. I guess you have to consider it anyway. A call to `Unsubscribe` may not be able to complete until a locking call is made to the miner, which means that `Unsubscribe` needs to be called when the miner is not locked.

Which means we need to either assume that all calls within the `tg` assume that the miner is not locked (which I guess makes sense), or we need a special case around it. I guess it's an exported thread.

DavidVorick:
But that means that you can't assume that `Process` will never be called when `IsStopped()` is returning true. If you want that assurance, you need either a `tg.BeforeStop` or you need to call `Unsubscribe` before you call `Stop()` in the closing function, which means decoupling the subscription code from the unsubscription code (not preferred).

modules/miner/miner.go
DavidVorick: You need both. First you check if it's stopped. If it is, you either return an error or call `build.Critical`. Then you do the `tg.Add(); defer tg.Done()`. You need both.

DavidVorick:
Oh nvm, I completely forgot that `tg.Add` will return an error if `IsStopped`. So, yeah, all you need is the `Add(); Done()` calls, and not the check for `IsStopped`. My mistake.

modules/miner/blockmanager.go
DavidVorick: gah forgot that `tg.Add` checked `IsStopped` for you.

modules/miner/blockmanager.go
DavidVorick: As Luke pointed out, simpling calling `m.tg.Add` is enough because it will check `IsStopped` inside of that function. Then you can return whatever error it returns.

modules/miner/blockmanager.go
```diff
@@ -158,6 +162,14 @@ func (m *Miner) managedSubmitBlock(b types.Block) error {
...
+		return err
+	}
+	defer m.tg.Done()
+	if m.tg.IsStopped() {
+		build.Critical(errClosed)
+	}
```
DavidVorick: Same here, no need to both call `tg.Add()` and check `IsStopped()` - my mistake.

modules/miner/miner.go
DavidVorick: I'm wrong.

modules/miner/cpuminer.go
DavidVorick: I'm re-writing some previous comments because they had an incorrect understanding of how `tg` works.

Definitely okay to call `tg.Add()` without holding a miner lock.

modules/miner/miner.go
```diff
@@ -194,10 +199,15 @@ func New(cs modules.ConsensusSet, tpool modules.TransactionPool, w modules.Walle
 // Close terminates all ongoing processes involving the miner, enabling garbage
 // collection.
 func (m *Miner) Close() error {
+	if err := m.tg.Stop(); err != nil {
```
DavidVorick: `m.tg.Stop()` depends on the cpu miner being turned off to close correctly. To ensure that, you should set `m.miningOn = false` before calling `m.tg.Stop()`. You should also write a test that starts the miner, turns on cpu mining, and then calls `Close` before turning off cpu mining. If all goes well, the cpu miner should close before `Stop()` returns. You can check that the cpu miner has closed by looking at the `m.mining` variable (not to be confused with `m.miningOn` - `m.mining` tells you whether or not the miner is *actually* mining, and `m.miningOn` tells you whether or not the miner is *supposed* to be mining.

DavidVorick: Hmm. Alternatively you might want to just check the value of `tg.IsStopped()` inside of the `threadedMining` thread.

lukechampine: I think the latter is the correct approach. Threads should use `tg` as a point of synchronization for knowing when to shut down.

mnsl: At the top of `threadedMine` is

```go
if err := m.tg.Add(); err != nil {
    return
}
```

Is this sufficient?  Or should `tg.IsStopped()` be checked explicitly in the for loop or something?

DavidVorick: You need to check `tg.IsStopped()` every iteration of the loop, otherwise there's a deadlock.

Easy way to see is to write a test that turns on the mining, then calls `miner.Close()`. `miner.Close()` will never complete, at least as far as I can tell, it'll just hang. (you should write this test regardless)

Inside the loop, there's a place where it checks the value of `miningOn` and then returns. In that same place, you should add a check for `tg.IsStopped` as well.

mnsl: Yeah, can confirm that `miner.Close` only completes when `tg.IsStopped()` is checked within the loop. 

Should the test I wrote to confirm this be included in the PR?  See below.  It's slow as is, and 

```go
func TestMinerCloseDeadlock(t *testing.T) {
	if testing.Short() {
		t.SkipNow()
	}
	mt, err := createMinerTester("TestMinerCloseDeadlock")
	if err != nil {
		t.Fatal(err)
	}

	// Turn miner on.
	mt.miner.StartCPUMining()

        // Without this wait, Close() is called on the miner before StartCPUMining() 
        // manages to call 'go threadedMine()'.  If this function is going to be included
        // in the Sia test suite, I should find a real fix since this still sets off warnings 
        // about data races.
	time.Sleep(time.Second)

	// Call close without stopping the miner.  Since 'tg.IsStopped' is
        // with each iteration of the loop in 'threadedMine', the miner
        // should close on the next iteration of said loop.  Without this
        // check, Close would hang forever.
	if err = mt.miner.Close(); err != nil {
		t.Fatal(err)
	}
}
```

DavidVorick: yeah, that's a good test to keep in the codebase.

mnsl:
* fix up commit history and a few comments
* remove unnecessary `IsStopped` checks
* kill `threadedMine` if `m.tg.IsStopped` (implement `TestMinerCloseDeadlock` to confirm that this is necessary)

mnsl:
rebased, should be ready to go

modules/miner/cpuminer.go
```diff
@@ -20,8 +27,15 @@ func (m *Miner) threadedMine() {
 	// occurring.
 	cycleStart := time.Now()
 	for {
-		// Kill the thread if mining has been turned off.
 		m.mu.Lock()
+		// Kill the thread if the miner's threadgroup is stopped.
+		if m.tg.IsStopped() {
+			m.miningOn = false
+			m.mining = false
```
VoidingWarranties: If we are in the process of closing setting these values to false shouldn't matter, right? Returning immediately would let you put the `IsStopped` check outside of the lock.

lukechampine: behavior of exported methods after shutdown is undefined (at least when they can't return an error), so for correctness' sake, `CPUMining` should return false after shutdown.

PR #1227, opened 2016-06-08
Share dialback port on Connect in the gateway

VoidingWarranties:
The gateway now shares a port that you can dial it back on when it connects to you. Consequently, inbound peers are now identified by this dialback port instead of `conn.RemoteAddr()`. Therefore `conn.RemoteAddr()` should no longer be used to call an RPC from within an RPC.

### DIFF
--- a/build/version.go
+++ b/build/version.go
@@ -6,7 +6,7 @@ import (
 )
 
 // Version is the current version of siad.
-const Version = "0.6.0"
+const Version = "1.0.0"
 
 // IsVersion returns whether str is a valid version number.
 func IsVersion(str string) bool {
--- a/build/version.go
+++ b/build/version.go
@@ -5,8 +5,17 @@ import (
 	"strings"
 )
 
-// Version is the current version of siad.
-const Version = "1.0.0"
+const (
+	// Version is the current version of siad.
+	Version = "1.0.0"
+
+	// MaxEncodedVersionLength is the maximum length of a version string encoded
+	// with the encode package. 100 is much larger than any version number we send
+	// now, but it allows us to send additional information in the version string
+	// later if we choose. For example appending the version string with the HEAD
+	// commit hash.
+	MaxEncodedVersionLength = 100
+)
 
 // IsVersion returns whether str is a valid version number.
 func IsVersion(str string) bool {
--- a/modules/gateway/peers.go
+++ b/modules/gateway/peers.go
@@ -178,7 +178,6 @@ func (g *Gateway) threadedAcceptConn(conn net.Conn) {
 	remoteVersion, err := acceptConnVersionHandshake(conn, build.Version)
 	if err != nil {
 		g.log.Debugf("INFO: %v wanted to connect but version handshake failed: %v", addr, err)
-		// TODO: should this be muxado.Server(conn).Close()?
 		conn.Close()
 		return
 	}
@@ -256,7 +255,7 @@ func connectVersionHandshake(conn net.Conn, version string) (remoteVersion strin
 
 // acceptConnVersionHandshake performs the version handshake and should be
 // called on the side accepting a connection request. The remote version is
-// only returned in err == nil.
+// only returned if err == nil.
 func acceptConnVersionHandshake(conn net.Conn, version string) (remoteVersion string, err error) {
 	// Read remote version.
 	if err := encoding.ReadObject(conn, &remoteVersion, maxAddrLength); err != nil {
@@ -304,7 +303,6 @@ func (g *Gateway) Connect(addr modules.NetAddress) error {
 	}
 	remoteVersion, err := connectVersionHandshake(conn, build.Version)
 	if err != nil {
-		// TODO: should this be muxado.Client(conn).Close()?
 		conn.Close()
 		return err
 	}
--- a/modules/gateway/peers.go
+++ b/modules/gateway/peers.go
@@ -85,7 +85,7 @@ func (p *peer) open() (modules.PeerConn, error) {
 	if err != nil {
 		return nil, err
 	}
-	return &peerConn{conn}, nil
+	return &peerConn{conn, p.NetAddress}, nil
 }
 
 func (p *peer) accept() (modules.PeerConn, error) {
@@ -93,7 +93,7 @@ func (p *peer) accept() (modules.PeerConn, error) {
 	if err != nil {
 		return nil, err
 	}
-	return &peerConn{conn}, nil
+	return &peerConn{conn, p.NetAddress}, nil
 }
 
 // addPeer adds a peer to the Gateway's peer list and spawns a listener thread
--- a/modules/gateway/peers.go
+++ b/modules/gateway/peers.go
@@ -320,7 +320,15 @@ func (g *Gateway) Connect(addr modules.NetAddress) error {
 		},
 		sess: muxado.Client(conn),
 	})
+	// addNode must be called after addPeer so that threadedPeerManager doesn't
+	// try to Connect to the node before we do (although in this particular case
+	// it doesn't matter as a lock is held over both calls).
+	err = g.addNode(addr)
 	g.mu.Unlock()
+	if err != nil && err != errNodeExists {
+		g.Disconnect(addr)
+		return err
+	}
 
 	// call initRPCs
 	g.mu.RLock()
--- a/modules/gateway/peers.go
+++ b/modules/gateway/peers.go
@@ -182,10 +182,84 @@ func (g *Gateway) threadedAcceptConn(conn net.Conn) {
 		return
 	}
 
+	if build.VersionCmp(remoteVersion, "1.0.0") < 0 {
+		err = g.managedAcceptConnOldPeer(conn, remoteVersion)
+	} else {
+		err = g.managedAcceptConnNewPeer(conn, remoteVersion)
+	}
+	if err != nil {
+		g.log.Debugf("INFO: %v wanted to connect, but failed: %v", addr, err)
+		conn.Close()
+		return
+	}
+
+	g.log.Debugf("INFO: accepted connection from new peer %v (v%v)", addr, remoteVersion)
+}
+
+// managedAcceptConnOldPeer accepts a connection request from peers < v1.0.0.
+// The requesting peer is added as a peer, but is not added to the node list
+// (older peers do not share their dialback address). The peer is only added if
+// a nil error is returned.
+func (g *Gateway) managedAcceptConnOldPeer(conn net.Conn, remoteVersion string) error {
+	addr := modules.NetAddress(conn.RemoteAddr().String())
+
+	g.mu.Lock()
+	defer g.mu.Unlock()
+
+	g.acceptPeer(&peer{
+		Peer: modules.Peer{
+			NetAddress: addr,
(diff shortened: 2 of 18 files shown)


mnsl:
Good job

modules/gateway/peers.go
```diff
@@ -173,20 +174,104 @@ func (g *Gateway) threadedAcceptConn(conn net.Conn) {
...
+// (older peers do not share their dialback address). The peer is only added if
+// a nil error is returned.
+func (g *Gateway) managedAcceptConnOldPeer(conn net.Conn, remoteVersion string) error {
+	addr := modules.NetAddress(conn.RemoteAddr().String())
+
+	g.mu.Lock()
+	defer g.mu.Unlock()
+
+	// Kick peers to make room if necessary.
+	g.kickPeersToMakeRoomFor(addr)
```
DavidVorick: This could have a better name lol. perhaps `g.kickAPeer`?

DavidVorick: oh, that function decides whether or not you need to kick anyone? Hmm.

It's a bit of a weird abstraction. Maybe `prepareForPeer`

lukechampine: definitely weird. I think you lose important context when you factor it out into a helper function. Maybe have a `g.acceptPeer` or similar that does both the `kick` and the `addPeer`, since they are common to both the old and new `managedAcceptConn`.

modules/gateway/peers.go
```diff
@@ -173,20 +174,104 @@ func (g *Gateway) threadedAcceptConn(conn net.Conn) {
...
 		conn.Close()
 		return
 	}
 
+	switch {
+	case build.VersionCmp(remoteVersion, "0.6.1") < 0:
+		err = g.managedAcceptConnOldPeer(conn, remoteVersion)
+	default:
+		err = g.managedAcceptConnNewPeer(conn, remoteVersion)
+	}
```
lukechampine: `switch` is overkill, stick with `if else` for now. This is COMPAT code anyway and will eventually be removed.

modules/gateway/peers.go
```diff
@@ -304,29 +474,31 @@ func (g *Gateway) Connect(addr modules.NetAddress) error {
...
 		return err
 	}
 
-	g.log.Println("INFO: connected to new peer", addr)
+	switch {
+	case build.VersionCmp(remoteVersion, "0.6.1") < 0:
+		err = g.managedConnectOldPeer(conn, remoteVersion, addr)
+	default:
+		err = g.managedConnectNewPeer(conn, remoteVersion, addr)
+	}
```
lukechampine: ditto

modules/gateway/peers.go
```diff
@@ -207,18 +292,41 @@ func (g *Gateway) threadedAcceptConn(conn net.Conn) {
...
+	if err != nil {
+		return "", fmt.Errorf("could not read remote peer's (%v) port: %v", addr, err)
+	}
+	if _, err := strconv.Atoi(dialbackPort); err != nil {
+		return "", fmt.Errorf("peer (%v) sent an invalid dialback port: %v", addr, err)
+	}
+	remoteAddr = modules.NetAddress(net.JoinHostPort(addr.Host(), dialbackPort))
+	// This check should only fail if dialbackPort == 0, which we could check
+	// manually, but there's no harm in validating the entire address.
+	if err := remoteAddr.IsValid(); err != nil {
```
lukechampine: doesn't `IsValid` obviate the need for the earlier `strconv.Atoi`? It also checks that the port's range is correct.

modules/gateway/peers.go
```diff
@@ -207,18 +292,41 @@ func (g *Gateway) threadedAcceptConn(conn net.Conn) {
...
-		sess: muxado.Server(conn),
-	})
-	g.mu.Unlock()
+}
+
+// acceptConnPortHandshake performs the port handshake and should be called on
+// the side accepting a connection request. The remote address is only returned
+// if err == nil.
+func acceptConnPortHandshake(conn net.Conn) (remoteAddr modules.NetAddress, err error) {
+	addr := modules.NetAddress(conn.RemoteAddr().String())
```
lukechampine: use `net.SplitHostPort` to get the host directly

modules/gateway/upnp.go
```diff
@@ -61,7 +66,7 @@ func (g *Gateway) learnHostname(port string) {
 		return
 	}
 
-	addr := modules.NetAddress(net.JoinHostPort(host, port))
+	addr := modules.NetAddress(net.JoinHostPort(host, g.port))
```
lukechampine: needs an `RLock`

modules/gateway/peers.go
VoidingWarranties: The string is sent as a port and I was worried that an adversary could send a port string that changed the host. Something similar to the peer sending the port "::x:9981" to which would append "::x" to the host portion of `remoteAddr`. This specific example isn't possible because IPV6 addresses must be enclosed with square brackets. It's probably overkill, but that's why the check is there.

Alternatively we replace the `Atoi` check with `if remoteAddr.Host() != addr.Host()`. Or instead of sending a port string we send a port integer.

modules/gateway/upnp.go
VoidingWarranties: `g.port` is immutable.

modules/gateway/peers.go
```diff
@@ -207,18 +292,41 @@ func (g *Gateway) threadedAcceptConn(conn net.Conn) {
...
+
+	// Read the peer's port that we can dial them back on.
+	var dialbackPort string
+	err = encoding.ReadObject(conn, &dialbackPort, 13) // Max port # is 65535 (5 digits long) + 8 byte string length prefix
+	if err != nil {
+		return "", fmt.Errorf("could not read remote peer's (%v) port: %v", addr, err)
```
VoidingWarranties: including the peer's address in the error here is redundant as `threadedAcceptConn` logs the peer's address along with the error.

build/version.go
```diff
@@ -5,8 +5,17 @@ import (
 	"strings"
 )
 
-// Version is the current version of siad.
-const Version = "0.6.0"
+const (
+	// Version is the current version of siad.
+	Version = "0.6.1"
```
VoidingWarranties: No v0.6.1, bump to v1.0.0 instead. Make sure to replace reference to v0.6.1 with v1.0.0

modules/gateway/upnp.go
DavidVorick: Still needs an RLock, for convention immutable fields either need to be named as such (`g.immutablePort`) or they need to be under lock when being accessed/modified.

Otherwise the developer needs to remember whether or not the field is immutable, and that greatly increases the barrier to entry for navigating something as complicated as the gateway.

VoidingWarranties:
Fixed up as per reviewer suggestions. Rebased and good to go.

modules/gateway/upnp.go
lukechampine: Despite its convenience, I think I'd rather not have an `immutable` convention, since it would be a lie. Immutable fields seem to be pretty rare anyway; the only other one that comes to mind is `persistDir`.

Also, consider that during testing, you may want to mutate the "immutable" field. If locks are not used when accessing the field, then it becomes impossible to mutate them without triggering a data race.

VoidingWarranties:
`threadedForwardPort` now takes the port as an argument to avoid locking. Added read locks in all other places `g.port` is used.

VoidingWarranties:
Rebased and good to go.

modules/gateway/peers.go
```diff
@@ -290,6 +458,9 @@ func (g *Gateway) Connect(addr modules.NetAddress) error {
 	if err := addr.IsValid(); err != nil {
 		return errors.New("can't connect to invalid address")
 	}
+	if net.ParseIP(addr.Host()) == nil {
+		return errors.New("address must be an IP address")
+	}
```
lukechampine: must it? why?

modules/gateway/upnp.go
```diff
@@ -38,10 +38,15 @@ func myExternalIP() (string, error) {
 	return string(buf[:n-1]), nil
 }
 
-// learnHostname discovers the external IP of the Gateway. Once the IP has
-// been discovered, it registers the ShareNodes RPC to be called on new
+// threadedLearnHostname discovers the external IP of the Gateway. Once the IP
+// has been discovered, it registers the ShareNodes RPC to be called on new
 // connections, advertising the IP to other nodes.
-func (g *Gateway) learnHostname(port string) {
+func (g *Gateway) threadedLearnHostname() {
```
lukechampine: why doesn't this take a `port string`?
Looks like it may have been altered back when the `g.threads.Add` was done outside the function.

modules/gateway/peers.go
VoidingWarranties: There's no reason to support non-IP addresses that I can think of. Discovered hostnames can only be IP addresses (unless UPnP supports non-IP hostname discovery; I don't think it does).

It would be deceptive for `Connect` to allow a non-IP address, as the peer it is connecting to would only be aware of its IP address. It is this IP address that the receiving peer adds to its node list. That is, unless `PortHandshake` becomes `AddressHandshake` which isn't unreasonable. But then you would want to make sure that the address the peer claims is it's address matches the `conn.RemoteAddr()`, by doing a DNS lookup.

It's possible, but adds more complexity than I think is necessary. I don't see any advantage to supporting non-IP addresses, especially since there is no current mechanism to set a peer's hostname as there is in the host module.

modules/gateway/upnp.go
DavidVorick: @VoidingWarranties and I went back and forth on it a bit. In general, it can be seen as the gateway that's forwarding it's port, which is why it makes sense for this function to be a method of the gateway. This point is hammered home more when you add the logging calls, which necessarily need the gateway's logger.

Since you're already using the gateway, it makes sense to also use the port within the gateway instead of passed as an argument. And since you don't want the gateway to close if this function hasn't shut down.

VoidingWarranties: I'd be willing to change it back if only to make the `RLock` unnecessary, and to make the argument signature match `threadedForwardPort`.

DavidVorick: I am against it. I don't think it makes sense for `threadedForwardPort` to have the port as an argument, you have the entire gateway which includes the port.

If it's a stateful function you should use the stateful paradigm. Otherwise there's confusion about whether or not the input port is in fact the same port as the gateway port.

lukechampine: I think I slightly prefer the current version, since it's a little strange for a function named `learnHostname` to take a port as an argument.

lukechampine:
Another NDF:
```
--- FAIL: TestIntegrationBroadcastRelayHeader (1.74s)
	synchronize_test.go:1155: RelayHeader didn't broadcast a valid block header
```
Travis is having a bad day, especially on Mac

lukechampine:
Now it's failing with:
```
=== RUN   TestSimpleInitialBlockchainDownload
panic: cannot write to the file after it has been closed
```

VoidingWarranties:
@lukechampine that will be fixed in #1266. Maybe it's best to wait until that is closed until merging this PR.

PR #1230, opened 2016-06-08
Implement Close methods for tpoolTester and hostTester; close miner in consensusSetTester.Close()

### LINKED ISSUE #1209: Close host tester in tests
By implementing the `Close` method on the `hostTester` type. `consensusSetTester.Close` is an existing example.

mnsl:
Implements `Close()` methods for `tpoolTester` and `hostTester`.  Adds `io.Closer` to the `TestMiner` interface so that `tpoolTester.Close()` and `hostTester.Close()` can close the miner.

These methods are necessary for clean shutdown during testing.  A later pull request will implement `modules.Wallet.Close()`.

Resolves #1209.

### DIFF
--- a/modules/host/host_test.go
+++ b/modules/host/host_test.go
@@ -199,6 +199,21 @@ func newHostTester(name string) (*hostTester, error) {
 	return ht, nil
 }
 
+// Close safely closes the hostTester. It panics if err != nil because there
+// isn't a good way to errcheck when deferring a close.
+func (ht *hostTester) Close() error {
+	errs := []error{
+		ht.cs.Close(),
+		ht.gateway.Close(),
+		ht.tpool.Close(),
+		ht.miner.Close(),
+	}
+	if err := build.JoinErrors(errs, "; "); err != nil {
+		panic(err)
+	}
+	return nil
+}
+
 // TestHostInitialization checks that the host initializes to sensible default
 // values.
 func TestHostInitialization(t *testing.T) {
--- a/modules/host/host_test.go
+++ b/modules/host/host_test.go
@@ -226,6 +226,7 @@ func TestHostInitialization(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	defer bht.Close()
 	if bht.host.blockHeight != 0 {
 		t.Error("host initialized to the wrong block height")
 	}
@@ -256,6 +257,7 @@ func TestHostMultiClose(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	defer ht.Close()
 
 	err = ht.host.Close()
 	if err != nil {
@@ -281,6 +283,7 @@ func TestNilValues(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	defer ht.Close()
 
 	hostDir := filepath.Join(ht.persistDir, modules.HostDir)
 	_, err = New(nil, ht.tpool, ht.wallet, "localhost:0", hostDir)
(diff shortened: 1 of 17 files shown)


lukechampine:
this commit history could be a little tidier. The two `add io.Closer to modules.Wallet interface` commits should be one; use `git commit --amend --no-edit`. (I have a git alias for this since I use it all the time.)
Also, undoing a commit (as with the wallet stuff) is usually performed with `git revert`. However, it's rare to revert a commit in a pull request. We prefer to simply remove the original commit. The easiest way to do this is `git reset --hard HEAD^`, but if it was a few commits back, use `git rebase -i` and simply delete the line corresponding to the commit.

modules/miner.go
```diff
@@ -67,6 +67,12 @@ type TestMiner interface {
 	// which amounts to trying a few thousand different nonces. SolveBlock is
 	// primarily used for testing.
 	SolveBlock(types.Block, types.Target) (types.Block, bool)
+
+	// TestMiner and Miner implement the io.Closer interface with the same
+	// Close() method. Miner needs to be an io.Closer for the
+	// api.Server.Close() method; TestMiner needs to be an io.Closer for
+	// clean shutdown during testing.
+	io.Closer
```
lukechampine: our style has been to write the method signature explicitly: `Close() error`. It's not a huge deal, but it signifies that calling `Close` on a module is not really the same as closing a file or a connection. As you noted in the comment, `Close` is about clean shutdown, not releasing a single resource.

mnsl:
latest changes:
* clean up commit history
* use `Close() error` instead of `io.Closer` in interface definitions

mnsl:
rebased, should be ready to go

modules/host/host_test.go
```diff
@@ -199,6 +199,20 @@ func newHostTester(name string) (*hostTester, error) {
 	return ht, nil
 }
 
+// Close safely closes the hostTester. It panics if err != nil because there
+// isn't a good way to errcheck when deferring a close.
+func (ht *hostTester) Close() error {
+	errs := []error{
+		ht.cs.Close(),
+		ht.gateway.Close(),
+		ht.tpool.Close(),
```
DavidVorick: forgot the miner

VoidingWarranties:
Can you add all the necessary `hostTester.Close` calls in tests.

Also the `TpoolTester` in `TestIntegrationLargeTransactions`, `TestSubscription`, and `TestArbDataOnly` need to be closed.

Otherwise looks good.

mnsl:
Latest changes:
* add `defer Close()` after all `createHostTester` and `blankHostTester` calls
* close `TpoolTester`s in `TestIntegrationLargeTransactions`, `TestSubscription`, and `TestArbDataOnly`

Should be ready to go.

lukechampine:
hmm, NDF in TestIntegrationRPCSendBlocks

DavidVorick:
I'm not familiar with that one, but part of me wants to say that #1227 fixes the problem

VoidingWarranties:
This specific failure was because of a read timeout. I suspect this is because running all the tests in parallel can significantly slow them down, especially once that involve a lot of I/O like this one does.

DavidVorick:
Hmm. We make have been too aggressive with the `t.Parallel()` then. It makes sense for tests with lots of idle sleep time (e.g. `TestFutureBlock` type tests), but perhaps doesn't make sense for tests that are time sensitive.

I'm pretty sure that go runs all serial tests before it starts any of the parallel tests. Should be good enough to just remove `t.Parallel()` for any tests that look like they might cause problems if they don't have full control of the cpu / disk.

lukechampine:
I will say that `t.Parallel` caused a test to fail for me because I was accidentally running it in the same directory as another test :|
We should definitely be careful when using it though, especially because it's likely to exacerbate any load-based NDFs

mnsl:
Rebased, ready to go

mnsl:
Rebased, ready to go

PR #1234, opened 2016-06-11
attempt to finalize the host api

DavidVorick:
I changed the names of a couple of fields, including the JSON.

Please take a close look at all of the spellings for the fields in the API. Once we finalize this, we will not be able to correct misspelled fields.

I'm thinking about creating an official bug bounty for finding misspelled fields. It's a good candidate for a bug bounty because it is a very specific, easily reviewed process for determining whether or not the 'bug' is valid and deserving, and there is a finite window (must be submitted before June 27th) that the bounty can be claimed.


DavidVorick:
There has been a notable change - instead of using 'Minimum' we are now using 'Min', to match 'Maximum' vs. 'Max' that's seen in the now-finalized renter-host protocol.

Another notable change is that some of the fields were missing words in the json transcription.

I don't think that this change breaks compatibility in any way, as only the internal settings are changed. Renters upgrading from v0.6.0 to v1.0.0 may find some of their settings zeroed out, we might want to include some compatibility code to prevent that.

doc/API.md
```diff
@@ -326,94 +326,90 @@ Queries:
 * /host/announce                [POST]
 * /host/delete/{filecontractid} [POST]
 
+[Full Description][api/Host.md]
```
lukechampine: need parens, not brackets

doc/API.md
```diff
@@ -1393,3 +1360,5 @@ encryptionpassword string
 frequently, the encryption password is the same as the primary wallet seed.
 
 Response: standard
+
+[api/Host.md]https://github.com/NebulousLabs/Sia/blob/master/doc/api/Host.md
```
lukechampine: need a colon after the declaration (see Jane's guides)

lukechampine:
otherwise LGTM

DavidVorick:
pushed the fixes. @VoidingWarranties has been working on a new documentation style/format that this PR doesn't yet conform to, but in the interest of having the API route changes merged, I'm going to nominate that the full doc changes be fixed in another PR.

Which means this PR should be ready to go.

Issue #1096, opened 2016-04-24
Nondeterministic failure in TestAutoRevisionSubmission


DavidVorick:
```
--- FAIL: TestAutoRevisionSubmission (4.31s)

	storageobligations_smoke_test.go:675: storage obligation is not saying that the storage proof was confirmed on the blockchain

--- PASS: TestStorageProof (3.39s)
```

VoidingWarranties:
Any update on this? This test is still failing frequently on Travis.

DavidVorick:
How frequently? I have not looked into it yet.

DavidVorick:
this is fixed

PR #1235, opened 2016-06-11
hardfork at block 100,000

DavidVorick:
The verify storage proof code is incorrect, missing the case where the
final segment was the correct length. This can be corrected with a
hardfork that has been scheduled for block 100,000.

We may want to consider anything else that might be on the hardfork wishlist since we're doing another hardfork.

All v1.0.0+ versions will have the hardfork code, seems reasonable to me.

### DIFF
--- a/crypto/merkle.go
+++ b/crypto/merkle.go
@@ -124,10 +124,12 @@ func MerkleProof(b []byte, proofIndex uint64) (base []byte, hashSet []Hash) {
 
 	// Get the proof and convert it to a base + hash set.
 	_, proof, _, _ := t.Prove()
-	base = proof[0]
-	hashSet = make([]Hash, len(proof)-1)
-	for i, p := range proof[1:] {
-		copy(hashSet[i][:], p)
+	if len(proof) > 0 {
+		base = proof[0]
+		hashSet = make([]Hash, len(proof)-1)
+		for i, p := range proof[1:] {
+			copy(hashSet[i][:], p)
+		}
 	}
 	return base, hashSet
 }
--- a/crypto/merkle.go
+++ b/crypto/merkle.go
@@ -124,12 +124,15 @@ func MerkleProof(b []byte, proofIndex uint64) (base []byte, hashSet []Hash) {
 
 	// Get the proof and convert it to a base + hash set.
 	_, proof, _, _ := t.Prove()
-	if len(proof) > 0 {
-		base = proof[0]
-		hashSet = make([]Hash, len(proof)-1)
-		for i, p := range proof[1:] {
-			copy(hashSet[i][:], p)
-		}
+	if len(proof) == 0 {
+		// There's no proof, because there's no data. Return blank values.
+		return nil, nil
+	}
+
+	base = proof[0]
+	hashSet = make([]Hash, len(proof)-1)
+	for i, p := range proof[1:] {
+		copy(hashSet[i][:], p)
 	}
 	return base, hashSet
 }
--- a/modules/consensus/validtransaction.go
+++ b/modules/consensus/validtransaction.go
@@ -144,7 +144,7 @@ func validStorageProofs(tx *bolt.Tx, t types.Transaction) error {
 			segmentLen = uint64(crypto.SegmentSize)
 		}
 		// HARDFORK 100,000
-		if (build.Release == "standard" && blockHeight(tx) < 100e3) || (build.Release == "testing" && blockHeight(tx) >= 10) {
+		if (build.Release == "standard" && blockHeight(tx) < 100e3) || (build.Release == "testing" && blockHeight(tx) < 10) {
 			segmentLen = uint64(crypto.SegmentSize)
 			if segmentIndex == leaves-1 {
 				segmentLen = fc.FileSize % crypto.SegmentSize
@@ -158,8 +158,21 @@ func validStorageProofs(tx *bolt.Tx, t types.Transaction) error {
 			segmentIndex,
 			fc.FileMerkleRoot,
 		)
-		if !verified {
+		if !verified && fc.FileSize > 0 {
 			return errInvalidStorageProof
+		}
+
+		// HARDFORK 100,000
+		//
+		// Originally, it was impossible to provide a storage proof for data of
+		// length zero. A hardfork was added triggering at block 100,000 to
+		// enable an optimization where hosts could submit empty storage proofs
+		// for files of size 0, saving space on the blockchain in conditions
+		// where the renter is content.
+		if (build.Release == "standard" && blockHeight(tx) < 100e3) || (build.Release == "testing" && blockHeight(tx) < 10) {
+			if !verified {
+				return errInvalidStorageProof
+			}
 		}
 	}
 
--- a/modules/consensus/validtransaction.go
+++ b/modules/consensus/validtransaction.go
@@ -114,16 +114,41 @@ func validStorageProofs(tx *bolt.Tx, t types.Transaction) error {
 		}
 		leaves := crypto.CalculateLeaves(fc.FileSize)
 		segmentLen := uint64(crypto.SegmentSize)
+
+		// If this segment chosen is the final segment, it should only be as
+		// long as necessary to complete the filesize.
 		if segmentIndex == leaves-1 {
 			segmentLen = fc.FileSize % crypto.SegmentSize
 		}
+		if segmentLen == 0 {
+			segmentLen = uint64(crypto.SegmentSize)
+		}
 
-		// COMPATv0.4.0
+		// HARDFORK 21,000
 		//
-		// Fixing the padding situation resulted in a hardfork. The below code
-		// will stop the hardfork from triggering before block 21,000.
+		// Originally, the code used the entire segment to verify the
+		// correctness of the storage proof. This made the code incompatible
+		// with data sizes that did not fill an entire segment.
+		//
+		// This was patched with a hardfork in block 21,000. The new code made
+		// it possible to perform successful storage proofs on the final
+		// segment of a file if the final segment was not crypto.SegmentSize
+		// bytes.
+		//
+		// Unfortunately, a new bug was introduced where storage proofs on the
+		// final segment would fail if the final segment was selected and was
+		// crypto.SegmentSize bytes, because the segmentLen would be set to 0
+		// instead of crypto.SegmentSize, due to an error with the modulus
+		// math. This new error has been fixed with the block 100,000 hardfork.
 		if (build.Release == "standard" && blockHeight(tx) < 21e3) || (build.Release == "testing" && blockHeight(tx) < 10) {
 			segmentLen = uint64(crypto.SegmentSize)
+		}
+		// HARDFORK 100,000
+		if (build.Release == "standard" && blockHeight(tx) < 100e3) || (build.Release == "testing" && blockHeight(tx) >= 10) {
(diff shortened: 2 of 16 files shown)


DavidVorick:
A hardfork height of 100,000 blocks gives us something like 8 months to get everyone to upgrade.

DavidVorick:
Should also be noted that this fixes one of the nodeterministic test failures.

modules/consensus/validtransaction.go
```diff
@@ -126,6 +126,15 @@ func validStorageProofs(tx *bolt.Tx, t types.Transaction) error {
...
 		}
 
+		// COMPATv1.0.0
+		//
+		// One hardfork apparently wasn't enough, the adjusted hardfork code
+		// ALSO had errors in it. A new hardfork has been tentatively scheduled
+		// for block 100,000.
+		if ((build.Release == "standard" && blockHeight(tx) < 100e3) || (build.Release == "testing" && blockHeight(tx) < 10)) && segmentLen == 0 {
+			segmentLen = uint64(crypto.SegmentSize)
+		}
```
lukechampine: I'm confused by this code. It seems like if `fc.FileSize%crypto.SegmentSize == 0` (and `fc.FileSize != 0`), we want `segmentLen = crypto.SegmentSize`. But this logic only gets triggered below the hardfork height. Shouldn't it be the other way around?

I think a good convention would be to thoroughly document the behavior of the code before the hardfork. 
We may also want to use consts for the hardfork heights.

```go
var hardforkHeightSegment1, hardforkHeightSegment2 = func() (types.BlockHeight, types.BlockHeight) {
	switch build.Release {
	case "standard":
		return 21e3, 100e3
	case "testing":
		return 10, 10
	}
}()
```

```go
segmentLen := crypto.SegmentSize
if segmentIndex == leaves-1 && fc.FileSize%crypto.SegmentSize != 0 {
	segmentLen = fc.FileSize % crypto.SegmentSize
}

if blockHeight(tx) < hardforkHeightSegment1 {
	// HARDFORK 21,000
	//
	// The original behavior was to use the entire segment for a storage
	// proof. This is unacceptable because the last segment may not
	// contain crypto.SegmentSize bytes of data, but will be padded with
	// zeros because sp.Segment is an array. This means that a segment
	// ending in one zero is indistinguishable from a segment ending in 64
	// zeros.
	segmentLen = crypto.SegmentSize
} else if blockHeight(tx) < hardforkHeightSegment2 {
	// HARDFORK 100,000
	//
	// A bug was introduced when fixing the previous hardfork: it did not
	// account for the case where fc.FileSize is an even multiple of
	// crypto.SegmentSize. In this case, segmentLen should be
	// crypto.SegmentSize, but was erroneously set to zero.
	segmentLen = fc.FileSize % crypto.SegmentSize
}
```

DavidVorick:
Don't merge this until we've put all of the hardfork code that we want in place, iirc there were a few other small things that were worth addressing

DavidVorick:
Also don't merge until there's better testing.

modules/consensus/validtransaction.go
```diff
@@ -120,11 +120,37 @@ func validStorageProofs(tx *bolt.Tx, t types.Transaction) error {
...
+		// Ironically, we now only use complete segments in our storage proof.
+		// Had we never made the original hardfork, the current implementation
+		// would be bug free. Unfortunately, a second hardfork was necessary to
+		// fix the new bug.
 		if (build.Release == "standard" && blockHeight(tx) < 21e3) || (build.Release == "testing" && blockHeight(tx) < 10) {
 			segmentLen = uint64(crypto.SegmentSize)
 		}
+		// COMPATv1.0.0
+		//
+		// HARDFORK 100,000
```
lukechampine: I think `COMPAT` is misleading because we sometimes remove `COMPAT` code after enough versions. `HARDFORK` should serve as its own identifier that makes it clear that you can never remove this code.

modules/consensus/validtransaction.go
```diff
@@ -120,11 +120,37 @@ func validStorageProofs(tx *bolt.Tx, t types.Transaction) error {
...
+		//
+		// This code addresses a hardfork error where the segment size would be
+		// set to 0 if segmentIndex == leaves-1. Prior to this hardfork,
+		// certain choices of segmentIndex would ensure that a host was
+		// completely unable to produce a satisfactory storage proof.
+		if ((build.Release == "standard" && blockHeight(tx) >= 100e3) || (build.Release == "testing" && blockHeight(tx) >= 10)) && segmentLen == 0 {
+			segmentLen = uint64(crypto.SegmentSize)
+		}
```
lukechampine: hardfork code should always be `blockHeight < threshold`. The current "correct" behavior should be in the main path, outside `if` statements.

modules/consensus/validtransaction.go
```diff
@@ -120,11 +120,37 @@ func validStorageProofs(tx *bolt.Tx, t types.Transaction) error {
...
+		// implementations use padding to complete the segments. The
+		// introduction of this hardfork made it possible to use uneven file
+		// sizes such that the last segment was incomplete, but introduced a
+		// new bug where a full final segment made a storage proof impossible
+		// if the final segment is selected.
+		//
+		// Ironically, we now only use complete segments in our storage proof.
+		// Had we never made the original hardfork, the current implementation
+		// would be bug free. Unfortunately, a second hardfork was necessary to
+		// fix the new bug.
```
lukechampine: this is unnecessarily verbose and you use "ironically" twice. Just explain the prior behavior and why it was wrong.

DavidVorick: only meant to use it once, guess I was distracted

DavidVorick: It did seem like in your comment though you wanted it to be verbose, and I do think that it's good to leave these lessons lying around. Hopefully someone can learn from the mistakes.

modules/consensus/validtransaction.go
```diff
@@ -114,17 +114,41 @@ func validStorageProofs(tx *bolt.Tx, t types.Transaction) error {
...
+		// instead of crypto.SegmentSize, due to an error with the modulus
+		// math. This new error has been fixed with the block 100,000 hardfork.
 		if (build.Release == "standard" && blockHeight(tx) < 21e3) || (build.Release == "testing" && blockHeight(tx) < 10) {
 			segmentLen = uint64(crypto.SegmentSize)
 		}
+		// HARDFORK 100,000
+		if (build.Release == "standard" && blockHeight(tx) < 100e3) || (build.Release == "testing" && blockHeight(tx) >= 10) {
+			if segmentIndex == leaves-1 && segmentLen == uint64(crypto.SegmentSize) && fc.FileSize > 0 {
+				segmentLen = 0
+			}
```
lukechampine: imo this path should just reproduce the original behavior:
```go
segmentLen = uint64(crypto.SegmentSize)
if segmentIndex == leaves-1 {
    segmentLen = fc.FileSize % crypto.SegmentSize
}
```
It's the easiest way to ensure that the behavior was not altered. Otherwise you risk coupling it to the code outside the `if`.

modules/consensus/validtransaction.go
lukechampine: I agree, the goal should be to prevent people from making the same mistake again. The comments should be as specific as possible. Even "This made the code incompatible with data sizes that did not fill an entire segment" could be more descriptive -- *why* is it incompatible? etc.

crypto/merkle.go
```diff
@@ -124,10 +124,12 @@ func MerkleProof(b []byte, proofIndex uint64) (base []byte, hashSet []Hash) {
 
 	// Get the proof and convert it to a base + hash set.
 	_, proof, _, _ := t.Prove()
-	base = proof[0]
-	hashSet = make([]Hash, len(proof)-1)
-	for i, p := range proof[1:] {
-		copy(hashSet[i][:], p)
+	if len(proof) > 0 {
```
lukechampine: is it possible for `proof` to have less than one element? If it only happens when `len(b) == 0`, perhaps it would be better to check for that case explicitly.
Even if not, I think edge cases should be handled "off the main path," i.e.
```go
if len(proof) == 0 {
    return nil, nil
}
```

DavidVorick: it's only a problem when `len(b) == 0` as far as I can see. But it was causing a panic during testing. We could probably throw that check at the top if you think that's cleaner.

lukechampine: since the code relies on `proof[0]` not panicking, I think it makes sense to check `len(proof)` rather than `len(b)`.

modules/consensus/validtransaction.go
```diff
@@ -133,9 +158,22 @@ func validStorageProofs(tx *bolt.Tx, t types.Transaction) error {
 			segmentIndex,
 			fc.FileMerkleRoot,
 		)
-		if !verified {
+		if !verified && fc.FileSize > 0 {
```
VoidingWarranties: Shouldn't this conditional be after the following conditional?

DavidVorick: It looks okay to me, what is the problem you are seeing?

DavidVorick:
Fixes #1096, not sure how to get the auto-close

DavidVorick:
Should be noted that, though the diff is huge, it's mostly just changes to the testing I made. I brought back some tests, fixed up some tests, made nondeterministic failures less likely in travis.

The actual diff here is very small. A tweak to `crypto.MerkleProof`, and a small number of lines introducing hardforking changes to `consensus.validtransaction.go`.

I have not finished brining back all of the testing that I feel is necessary to be comfortable with the hardfork, and there is some more testing that I wish to write. I also need to verify that blockchain synchronization can still occur, indicating more thoroughly that the proper behavior is happening for blocks before height 100,000.

modules/consensus/validtransaction.go
VoidingWarranties: The code is functionally correct, but it's confusing to see a `HARDFORK` block as the last block in a function.

To build on top of what @lukechampine said:
>hardfork code should always be `blockHeight < threshold`. The current "correct" behavior should be in the main path, outside `if` statements.

I would add to that and say that a `HARDFORK` block should come before the "correct" behavior. If the `if !verified && fc.FileSize > 0 {` block is the "correct" behavior, it would go after the `HARDFORK` block.

For this particular case both orderings are functionally equivalent because we are adding a condition, not removing a condition. For example, if the hardfork was to remove a condition the `HARDFORK` block would need to come before the "correct" behavior. Consider this example
```go
// HARDFORK
if (build.Release == "standard" && blockHeight(tx) < 100e3) || (build.Release == "testing" && blockHeight(tx) < 10) {
    if !verified && SOME_OLD_CONDITION {
        return errInvalidStorageProof
    }
}

if !verified { // Note that SOME_OLD_CONDITION was removed from the "correct" behavior.
    return errInvalidStorageProof
}
```

Swapping the order of these blocks would not result in the desired behavior.

DavidVorick:
Additional testing revealed a pretty significant bug - pre-hardfork code was actually incorrect.

I realized the correct way to make sure that pre-hardfork behavior was identical was to... copy-paste into a new function, and then call this function if the height was below 100e3. Should have figured that out sooner, that's how they deal with stuff like this in Bitcoin.

It's much easier to be certain about correctness when doing it this way, and has the added benefit of not cluttering the post-hardfork code with pre-hardfork branching.

DavidVorick:
should be ready to merge

lukechampine:
looks like your build never ran for some reason

PR #1239, opened 2016-06-15
Improve 'ActiveHosts' behavior

DavidVorick:
First, I changed it so that 'ActiveHosts' sorts the hosts by preference. As we add more preferences like geography and reliability, the UI doesn't need to worry about price estimation as much.

Second, I extended the `/renter/hosts/active` call to support limiting the number of hosts returned to 'numhosts'. This field is optional, and has testing.

I think, in another PR, we should add a field that just gives a price estimate, instead of having the UI do the price estimate. The renter knows a lot better what it's behavior is going to be.

### DIFF
--- a/api/renter.go
+++ b/api/renter.go
@@ -242,8 +242,28 @@ func (srv *Server) renterUploadHandler(w http.ResponseWriter, req *http.Request,
 // renterHostsActiveHandler handes the API call asking for the list of active
 // hosts.
 func (srv *Server) renterHostsActiveHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	var numHosts uint64
+	hosts := srv.renter.ActiveHosts()
+
+	if req.FormValue("numhosts") == "" {
+		// Default value for 'numhosts' is all of them.
+		numHosts = uint64(len(hosts))
+	} else {
+		// Parse the value for 'numhosts'.
+		_, err := fmt.Sscan(req.FormValue("numhosts"), &numHosts)
+		if err != nil {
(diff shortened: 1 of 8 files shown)


api/renter.go
```diff
@@ -242,8 +242,25 @@ func (srv *Server) renterUploadHandler(w http.ResponseWriter, req *http.Request,
...
+
+	var numHosts uint64
+	_, err := fmt.Sscan(req.FormValue("numhosts"), &numHosts)
+	if err != nil {
+		writeError(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+	if numHosts > uint64(len(hosts)) {
+		numHosts = uint64(len(hosts))
+	}
```
lukechampine: I'd prefer to see these two cases combined, such that there's only one `writeJSON`. `numhosts == ""` should be a special case that sets `numHosts` to `len(hosts)`.

doc/API.md
```diff
@@ -819,9 +819,15 @@ Response: standard.
...
 #### /renter/hosts/active [GET]
 
-Function: Lists all of the active hosts known to the renter.
+Function: Lists all of the active hosts known to the renter, sorted by
+preference.
 
-Parameters: none
+Parameters:
+```
+numhosts uint64 // Optional
```
lukechampine: do we have a convention for specifying optional parameters?

DavidVorick: It's more consistent in the new architecture that @VoidingWarranties and I discussed offline a few days ago. This is mostly a placeholder until we switch to that format.

The new doc architecture clearly marks optional parameters.

DavidVorick:
fixed, rebased, build passed

PR #1246, opened 2016-06-16
host now accepts renewals that don't extend the deadline

DavidVorick:
This is missing an integration test with the renter that verifies renewing at a low height is okay.


lukechampine:
Note to self: when enough hosts are running this code, the contractor can renew contracts with a lesser period immediately, instead of waiting until the old ones expire.

PR #1248, opened 2016-06-16
encapsulate Allowance in RenterSettings

lukechampine:
The renter is now structured similarly to the host; the `/renter/allowance` endpoints have been merged with `/renter`, and the backend has been updated accordingly.

There is one discrepancy that warrants consideration: Whereas the host has an `announce` endpoint that "activates" the settings, the renter has no such corresponding call. This means that the renter will attempt to start forming contracts even if you only set `funds` or `period` (i.e., not both). Clearly this doesn't make much sense, so the current behavior of `/renter` is to only allow calls that specify both parameters. Similarly, siac still uses `renter allowance` instead of `renter config`, because the former forces you to provide both parameters.
Bottom line: we may want to consider adding an endpoint that instructs the renter to put the settings into effect.

Later I would like to tweak the renter internals such that `SetSettings` merely informs the contractor of the new allowance, instead of commanding it to form/renew contracts. One downside of this is that contract formation errors won't be immediately returned by the `SetSettings` call; they will need to be bubbled up in some other fashion. However, this is already the case for automatically scheduled renewals, so clearly we need a better approach anyway.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -145,9 +145,8 @@ func (srv *Server) initAPI(password string) {
 
 	// Renter API Calls
 	if srv.renter != nil {
-		router.GET("/renter", srv.renterHandler)
-		router.GET("/renter/allowance", srv.renterAllowanceHandlerGET)
-		router.POST("/renter/allowance", requirePassword(srv.renterAllowanceHandlerPOST, password))
+		router.GET("/renter", srv.renterHandlerGET)
+		router.POST("/renter", requirePassword(srv.renterHandlerPOST, password))
 		router.GET("/renter/contracts", srv.renterContractsHandler)
 		router.GET("/renter/downloads", srv.renterDownloadsHandler)
 		router.GET("/renter/files", srv.renterFilesHandler)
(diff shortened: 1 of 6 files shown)


DavidVorick:
There should probably be an 'active renter' call. The host has `acceptingcontracts` which is more or less the same. Announcing merely tells people you exist, you can still disable the host after announcing.

It's not clear how one would disable the renter, but I guess it would just stop renewing contracts, stop doing uploads, and stop doing downloads. Might as well keep the files though in case the user changes their mind.

DavidVorick:
Looks good, it's all pretty straightforward

lukechampine:
maybe a setting like `automanage=true`. Disabling `automanage` would stop the renter from renewing, but I don't see why it should prevent it from uploading or downloading as well. Well, I suppose there's a complication if the renter is automatically renewing in response to uploads/downloads...hmm

DavidVorick:
I would think that if you turn the renter off, it's because you don't want it spending money.

Not sure what the best UX is here, but I don't think it needs to block this PR.

DavidVorick:
Allowance of 0 is a good way in the meantime to make sure contracts don't get formed.

Issue #1243, opened 2016-06-15
API standard success and error responses


VoidingWarranties:
The standard success response is `{ "Success": true }` which does not conform to the capitalization standard. Is this a standard response or are we free to change the capitalization?

Is there a standard error response format we should be conforming to?

VoidingWarranties:
HTTP status code 204 might be more semantic than the success JSON.

VoidingWarranties:
Closed in #1249

PR #1249, opened 2016-06-16
API standard success & error responses

VoidingWarranties:
This PR changes the standard success & error responses as described in #1243.

The APIError type is meant to provide descriptive, useful errors to API clients. Simply returning an error string is often not useful. An error string is the minimum, but other information such as which parameter caused the request to fail should also be included. Only the message field has been implemented in this commit as the other fields are non-trivial to implement. Additional fields can be added to APIError type in the future without breaking compatibility.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -8,6 +8,26 @@ import (
 	"github.com/julienschmidt/httprouter"
 )
 
+// APIError is a type that is encoded as JSON and returned in an API response
+// in the event of an error. Only the Message field is required. More fields
+// may be added to this struct in the future for better error reporting.
+type APIError struct {
+	// Message describes the error in English. Typically it is set to
+	// `err.Error()`. This field is required.
+	Message string `json:"message"`
+
+	// TODO: add a Param field with the (omitempty option in the json tag)
+	// to indicate that the error was caused by an invalid, missing, or
+	// incorrect parameter. This is not trivial as the API does not
+	// currently do parameter validation itself. For example, the
+	// /gateway/connect endpoint relies on the gateway.Connect method to
+	// validate the netaddress. However, this prevents the API from knowing
+	// whether an error returned by gateway.Connect is because of a
+	// connection error or an invalid netaddress parameter. Validating
+	// parameters in the API is not sufficient, as a parameter's value may
+	// be valid or invalid depending on the current state of a module.
+}
+
 // HttpGET is a utility function for making http get requests to sia with a
 // whitelisted user-agent. A non-2xx response does not return an error.
 func HttpGET(url string) (resp *http.Response, err error) {
@@ -63,7 +83,7 @@ func HttpPOSTAuthenticated(url string, data string, password string) (resp *http
 func requireUserAgent(h http.Handler, ua string) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
 		if !strings.Contains(req.UserAgent(), ua) {
-			writeError(w, "Browser access disabled due to security vulnerability. Use Sia-UI or siac.", http.StatusBadRequest)
+			writeError(w, APIError{"Browser access disabled due to security vulnerability. Use Sia-UI or siac."}, http.StatusBadRequest)
 			return
 		}
 		h.ServeHTTP(w, req)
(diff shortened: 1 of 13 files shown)


lukechampine:
Looks like this is causing `TestStop` to fail.

VoidingWarranties:
I can't figure out a trivial fix for the test. If you have ideas, lmk. Otherwise I'm putting this PR on the back burner for now in favor of finishing the API docs.

VoidingWarranties:
Need to update siac to use the new success and error responses

DavidVorick:
If I'm understanding this PR correctly, we've changed the success code to provide no content, but to use the error code '204'  - "Success but no content"

And we've changed the error stuff to return JSON instead of a string. You can identify that it's error JSON because there will be an error code, either 4xx or 5xx. The json value containing the error string is called "message", and there's room to add more fields in the future which might give the developer additional information about what went wrong.

That seems good to me, except for using "message" to return an error. Though you have the clue of the 4xx or 5xx, I feel like we should be more explicit and have the field containing the error message named "error", or "errormessage". (though I prefer just "error"). That change will probably save some bugs in the future for our api devs, if I had to guess.

VoidingWarranties:
@DavidVorick we talked offline and agreed to change the `message` field to `error`. I looked at a few more widely used APIs to see what they do, and the standard does seem to be `message`. The fact that it's an error is indicated by the non-2xx error code. See [Stripe's API docs](https://stripe.com/docs/api#errors) and [GitHub's API docs](https://developer.github.com/v3/#client-errors)

VoidingWarranties:
@lukechampine the test has been commented out for now. See the commit "Disable TestStop..." for details.

VoidingWarranties:
It looks like OS X builds aren't starting on Travis again. The build passed on Linux and should be good to go.

api/api.go
```diff
@@ -8,6 +8,32 @@ import (
 	"github.com/julienschmidt/httprouter"
 )
 
+// Error is a type that is encoded as JSON and returned in an API response in
+// the event of an error. Only the Message field is required. More fields may
+// be added to this struct in the future for better error reporting.
+type Error struct {
+	// Message describes the error in English. Typically it is set to
+	// `err.Error()`. This field is requried.
```
lukechampine: typo

api/api.go
```diff
@@ -8,6 +8,32 @@ import (
...
+	Message string `json:"message"`
+
+	// TODO: add a Param field with the (noemit option in the json tag) to
```
lukechampine: I assume you mean `omitempty`

api/api.go
```diff
@@ -217,8 +247,9 @@ func writeJSON(w http.ResponseWriter, obj interface{}) {
 	}
 }
 
-// writeSuccess writes the success json object ({"Success":true}) to the
-// ResponseWriter
+// writeSuccess writes the HTTP header with status 204 No Content to the
+// ResponseWriter. writeSuccess should only be used to indicate that the
+// requested action succeeded AND there is no data to return.
```
lukechampine: :+1:

api/server_helpers_test.go
```diff
@@ -337,6 +337,25 @@ func createExplorerServerTester(name string) (*serverTester, error) {
...
+// non-2xx. If an error message or Error is included in the response, it is
+// decoded and returned as the error.
+func decodeErrorResponse(resp *http.Response) error {
+	if resp.StatusCode < 200 || resp.StatusCode > 299 {
+		respErr, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return err
+		}
+		var apiErr Error
+		err = json.Unmarshal(respErr, &apiErr)
```
lukechampine: use `json.NewDecoder(resp.Body)`

api/server_helpers_test.go
```diff
@@ -337,6 +337,25 @@ func createExplorerServerTester(name string) (*serverTester, error) {
 	return st, nil
 }
 
+// decodeErrorResponse returns an error if the response's status code is
+// non-2xx. If an error message or Error is included in the response, it is
+// decoded and returned as the error.
+func decodeErrorResponse(resp *http.Response) error {
+	if resp.StatusCode < 200 || resp.StatusCode > 299 {
```
lukechampine: this function is a little weird, since it may not actually decode anything at all. I'd prefer the conditional to be outside the function; something like:
```go
if non2xx(resp.StatusCode) {
    return decodeError(resp)
}
```

VoidingWarranties:
Should be good to go now. The branch is out of date with master by once commit, but the commit is unrelated and this PR should be able to be merged without a problem. Otherwise I can rebase.

PR #1251, opened 2016-06-17
Fix broken links in API doc

VoidingWarranties:
~~And use bootstrap IP in API examples.~~

### DIFF
--- a/doc/api/Gateway.md
+++ b/doc/api/Gateway.md
@@ -22,13 +22,13 @@ peers on its own.
 Index
 -----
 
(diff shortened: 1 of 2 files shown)


doc/api/Gateway.md
```diff
@@ -150,7 +150,7 @@ Examples
 
 ###### Request
 ```
-/gateway/disconnect/123.456.789.0:123
+/gateway/disconnect/23.239.14.98:9981
```
lukechampine: I don't think we should be using bootstrap addresses in the docs. The old IP was clearly a placeholder; this is not.

PR #1252, opened 2016-06-17
Fix siac gateway connect / disconnect

VoidingWarranties:
broken by #1223. The lack of testing in siac finally bites us


DavidVorick:
There are a few other problems with misspellings in the siac host command as well I believe. Luckily not a huge problem, especially once there's auto-updating it'll be pretty easy to push out minor fixes like that.

PR #1254, opened 2016-06-17
API typos

lukechampine:
I grepped through every json tag in the repo and checked that they were equivalent to the lowercased field name. These were the only inconsistencies.
I did not thoroughly check for typos that may have occurred in both the tag and the field name.


VoidingWarranties:
ACK

We should have a competition to see who can find the most bugs using only regex.

VoidingWarranties:
Never mind, host financial metrics are saved to the host's persist. This will break compatibility with loading old persist files.

DavidVorick:
hmm. I guess we need some compatibility code at startup that can check for old potential revenue.

Shouldn't be too bad.

lukechampine:
it's just a pain because you need to define a whole struct with all the same fields, the only difference being the typo.

I wish there was a json tag that would enable decoding from multiple field names

DavidVorick:
@lukechampine no need. Just make a compatibility struct that has a single field `potentialestoragerevenue` and load the json object into it. All the extra fields get ignored. If that field is non-zero and `potentialstoragerevenue` is zero, you just overwrite the new with the old.

So you only need to define a compat struct with a single field. I will do it, since it's in the host.

DavidVorick:
For this PR, please remove the host field from the changes. I will fix the host field in a separate PR.

lukechampine:
ah, true. That is a little nicer, I suppose. You could even define the struct type inline rather than declaring it at the top level.

lukechampine:
without that commit, this would be a one-line PR. I'd rather you fix the host by pushing to this branch.

modules/host/persist.go
```diff
@@ -171,6 +171,44 @@ func (h *Host) load() error {
...
+	}
+	if compatPersistence.Settings.MinDownloadBandwidthPrice.Cmp(types.ZeroCurrency) > 0 && p.Settings.MinDownloadBandwidthPrice.Cmp(types.ZeroCurrency) == 0 {
+		h.settings.MinDownloadBandwidthPrice = compatPersistence.Settings.MinDownloadBandwidthPrice
+	}
+	if compatPersistence.Settings.MinStoragePrice.Cmp(types.ZeroCurrency) > 0 && p.Settings.MinStoragePrice.Cmp(types.ZeroCurrency) == 0 {
+		h.settings.MinStoragePrice = compatPersistence.Settings.MinStoragePrice
+	}
+	if compatPersistence.Settings.MinUploadBandwidthPrice.Cmp(types.ZeroCurrency) > 0 && p.Settings.MinUploadBandwidthPrice.Cmp(types.ZeroCurrency) == 0 {
+		h.settings.MinUploadBandwidthPrice = compatPersistence.Settings.MinUploadBandwidthPrice
+	}
```
DavidVorick: Maybe I should pop this into a separate function.

DavidVorick:
fixed, rebased, build passed

modules/host/persist_compat_1.0.0_test.go
```diff
@@ -0,0 +1,69 @@
...
+	t.Parallel()
+	ht, err := newHostTester("TestHostPersistCompat100")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Close the host and then swap out the persist file for the one that is
+	// being used for testing.
+	ht.host.Close()
+	err = build.CopyFile("persist_compat_1.0.0.json", filepath.Join(ht.host.persistDir, settingsFile))
```
lukechampine: the json file should be placed in a folder called `testdata`. The `go` tool will ignore such files. This is the convention used by the standard library, e.g. https://golang.org/src/image

modules/host/persist.go
```diff
@@ -187,32 +208,27 @@ func (h *Host) load() error {
...
 	}
-	err = h.dependencies.loadFile(persistMetadata, &compatPersistence, filepath.Join(h.persistDir, settingsFile))
+	err := h.dependencies.loadFile(persistMetadata, &compatPersistence, filepath.Join(h.persistDir, settingsFile))
 	if err != nil {
 		return err
 	}
 	// Load the compat values, but only if the compat values are non-zero and
 	// the real values are zero.
-	if compatPersistence.FinancialMetrics.PotentialStorageRevenue.Cmp(types.ZeroCurrency) > 0 && p.FinancialMetrics.PotentialStorageRevenue.Cmp(types.ZeroCurrency) == 0 {
+	if compatPersistence.FinancialMetrics.PotentialStorageRevenue.Cmp(types.ZeroCurrency) > 0 && p.FinancialMetrics.PotentialStorageRevenue.IsZero() {
```
lukechampine: use `!IsZero` instead of `Cmp(types.ZeroCurrency) > 0`. Currency values can't be negative.

DavidVorick:
fixed, rebased, build passed.

Worth noting that the compatibility package also has files that could be moved to a testdata dir, but can probably be ignored for now - it's been that way for a while.

PR #1255, opened 2016-06-17
Correct JSON case for HostDBEntry

VoidingWarranties:
Are JSONified HostDBEntries ever sent over the network? This would break compatibility if they are.


lukechampine:
good catch. No, they are not sent over the network. This will still break compatibility with old hostdbs though.

VoidingWarranties:
Oh, cool!

PR #1257, opened 2016-06-18
Renter resync

lukechampine:
The renter and host can "desynchronize" if a disconnect occurs at a specific part of the revision protocol. Specifically, if the renter disconnects after sending the revision to the host, but before receiving the host's signatures, it will not know for sure whether the host has applied the revision or not.

One solution to this would be for the host to accept references to either the "current" or "previous" revision (according to the host). However, this requires the host to store up to one additional segment per contract.

An alternative (implemented here) is for the renter to save the revision to disk before sending it to the host for approval. If a disconnect then occurs and the renter reconnects, it will first attempt to synchronize with the "old" revision, and then try the "new" revision if that fails. Note that no additional storage requirement is imposed here, because the renter is not the party storing the data.

### DIFF
--- a/modules/renter/contractor/host_integration_test.go
+++ b/modules/renter/contractor/host_integration_test.go
@@ -15,6 +15,7 @@ import (
 	"github.com/NebulousLabs/Sia/modules/host"
 	"github.com/NebulousLabs/Sia/modules/miner"
 	"github.com/NebulousLabs/Sia/modules/renter/hostdb"
+	"github.com/NebulousLabs/Sia/modules/renter/proto"
 	"github.com/NebulousLabs/Sia/modules/transactionpool"
 	modWallet "github.com/NebulousLabs/Sia/modules/wallet"
 	"github.com/NebulousLabs/Sia/types"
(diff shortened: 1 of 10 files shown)


api/explorer.go
```diff
@@ -164,7 +164,7 @@ func (srv *Server) buildExplorerTransaction(height types.BlockHeight, parent typ
 	}
 
 	for _, sfi := range txn.SiafundInputs {
-		et.SiaClaimOutputIDs = append(et.SiaClaimOutputIDs, sfi.ParentID.SiaClaimOutputID())
+		et.SiafundClaimOutputIDs = append(et.SiafundClaimOutputIDs, sfi.ParentID.SiaClaimOutputID())
```
DavidVorick: looks like you grabbed some commits from your other PR

modules/host.go
```diff
@@ -35,7 +35,7 @@ type (
 		LockedStorageCollateral types.Currency `json:"lockedstoragecollateral"`
 		LostRevenue             types.Currency `json:"lostrevenue"`
 		LostStorageCollateral   types.Currency `json:"loststoragecollateral"`
-		PotentialStorageRevenue types.Currency `json:"potentialestoragerevenue"`
+		PotentialStorageRevenue types.Currency `json:"potentialstoragerevenue"`
```
DavidVorick: ditto

modules/renter/proto/downloader.go
```diff
@@ -35,6 +37,17 @@ func (hd *Downloader) Sector(root crypto.Hash) (modules.RenterContract, []byte,
...
+	// create the download revision
+	rev := newDownloadRevision(hd.contract.LastRevision, sectorPrice)
+
+	// if a SaveFn was provided, call it before doing any further I/O, because
+	// that's when we're most likely to experience failure
+	if hd.SaveFn != nil {
+		if err := hd.SaveFn(rev); err != nil {
+			return modules.RenterContract{}, nil, errors.New("failed to save unsigned revision: " + err.Error())
+		}
+	}
```
DavidVorick: This is a little confusing. Why might the safeFn be nil?

modules/renter/proto/negotiate.go
```diff
@@ -105,7 +104,7 @@ func verifyRecentRevision(conn net.Conn, contract modules.RenterContract) error
 	// check that revision number matches; if it does, do a more thorough
 	// check by comparing unlock hashes.
 	if lastRevision.NewRevisionNumber != contract.LastRevision.NewRevisionNumber {
-		return fmt.Errorf("our revision number (%v) does not match the host's (%v)", contract.LastRevision.NewRevisionNumber, lastRevision.NewRevisionNumber)
+		return &recentRevisionError{contract.LastRevision.NewRevisionNumber, lastRevision.NewRevisionNumber}
```
DavidVorick: The unlock hashes and signatures should be checked before the revision number, that way if a recent revision error is returned the caller can know that the only problem is the recent revision, doesn't have to worry about anything else.

Also, I might be missing something, but shouldn't the recent revision error also be returning the host signature? For the renter to be consistent, it should be saving the signatures.

DavidVorick:
The biggest thing I'm worried about here is the signatures. Need to make sure that the revision update doesn't happen until the signatures + unlock hashes have been verified, and then you need to make sure that the renter's local copy also has both of the signatures.

When the desync happens, it happens before the renter gets the host signatures. I don't think your test accurately reflects that.

modules/renter/contractor/host_integration_test.go
```diff
@@ -573,3 +574,103 @@ func TestIntegrationRenew(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	_, err = c2.Downloader(badContract)
+	if !proto.IsRevisionMismatch(err) {
+		t.Fatal(err)
+	}
+
+	// add cachedRevision
+	c2.mu.Lock()
+	c2.cachedRevisions[contract.ID] = contract.LastRevision
```
DavidVorick: This is cheating, because `contract.LastRevision` has all the signatures, but the renter in reality will not.

Would be a lot better if we could somehow get the connection to fail at just the right moment. I don't really have a good idea of how to do that though without making the test extremely brittle. There might be some way to write a 'bad host' that intentionally disconnects before writing the signatures down the pipe.

Hmm. Maybe we could add a:

```go
if build.Release == "testing" {
    // might be incorrect
    //
    // roll a random number [0,4]
    randNum := crypto.RandInt(4)
    if randNum == 4 {
        return nil // Quit without writing the host signature, but let the host commit the change
    }
}
```

DavidVorick: That might ruin some of our existing tests though if they don't know to retry a few times if the first attempts don't succeed.

modules/renter/proto/downloader.go
lukechampine: currently, `Downloader` and `Editor` do not require that a SaveFn is provided. Alternatively, we could force callers to supply it as an argument to `NewEditor`/`NewDownloader`, or we could supply a default `SaveFn` that simply returns `nil`.

modules/renter/proto/negotiate.go
lukechampine: If the contractor sees a recent revision error, it will start over (i.e. close the connection and make a new RPC) using the saved revision.

One failing of the current implementation is that it does not reconstruct the revision transaction in the event of a desync. However, this transaction is only stored so that it might be submitted to the blockchain later, and as you pointed out, this is rarely in the renter's interest anyway. Also, the revision transaction will be correctly updated as soon as the next revision completes.

modules/renter/proto/negotiate.go
```diff
@@ -102,12 +101,12 @@ func verifyRecentRevision(conn net.Conn, contract modules.RenterContract) error
...
-	if lastRevision.NewRevisionNumber != contract.LastRevision.NewRevisionNumber {
-		return fmt.Errorf("our revision number (%v) does not match the host's (%v)", contract.LastRevision.NewRevisionNumber, lastRevision.NewRevisionNumber)
-	} else if lastRevision.UnlockConditions.UnlockHash() != contract.LastRevision.UnlockConditions.UnlockHash() {
+	// Check that the unlock hashes match; if they do not, something is
+	// seriously wrong. Otherwise, check that the revision numbers match.
+	if lastRevision.UnlockConditions.UnlockHash() != contract.LastRevision.UnlockConditions.UnlockHash() {
 		return errors.New("unlock conditions do not match")
+	} else if lastRevision.NewRevisionNumber != contract.LastRevision.NewRevisionNumber {
+		return &recentRevisionError{contract.LastRevision.NewRevisionNumber, lastRevision.NewRevisionNumber}
 	}
```
DavidVorick: I think you still need to call `VerifyFileContractRevisionTransactionSignatures` before you return a recent revision error.

Need to make sure that if the revision supplied by the host has invalid signatures, the renter catches them.

modules/renter/contractor/host_integration_test.go
```diff
@@ -573,3 +574,103 @@ func TestIntegrationRenew(t *testing.T) {
...
+		t.Fatal("expected revision mismatch, got", err)
+	}
+	_, err = c.Downloader(badContract)
+	if !proto.IsRevisionMismatch(err) {
+		t.Fatal("expected revision mismatch, got", err)
+	}
+
+	// add cachedRevision
+	c.mu.Lock()
+	c.cachedRevisions[contract.ID] = contract.LastRevision
```
DavidVorick: This is still an issue, because on the live network the cached revision is not going to have the signatures, but here the last revision does have the signatures.

Unless I'm missing something.

modules/renter/proto/negotiate.go
lukechampine: it will. After `recentRevisionError` is returned, `verifyRecentRevision` will be called again with the other revision, and the renter will only proceed if it returns `nil`.

modules/renter/contractor/host_integration_test.go
lukechampine: I think you are confusing "revision" with "revision transaction." The revision itself does not carry the signatures in question.

DavidVorick: oh, it's just loading the revision without the signatures.

ok.

PR #1259, opened 2016-06-18
switch the host persistence to correctly load compatibility fields

DavidVorick:
There were some incorrectly named fields in the pre-verion 1.0 releases
of the host. The names have since been corrected, but compatibility code
needed to be written to make sure that old hosts doing an upgrade would
not experience any issues due to the renaming.


DavidVorick:
Please double check the spelling for me.

modules/host/persist.go
```diff
@@ -171,6 +171,44 @@ func (h *Host) load() error {
...
+		FinancialMetrics struct {
+			PotentialStorageRevenue types.Currency `json:"potentialerevenue"`
+		}
+		Settings struct {
+			MinContractPrice          types.Currency `json:"contractprice"`
+			MinDownloadBandwidthPrice types.Currency `json:"minimumdownloadbandwidthprice"`
+			MinStoragePrice           types.Currency `json:"storageprice"`
+			MinUploadBandwidthPrice   types.Currency `json:"minimumuploadbandwidthprice"`
+		}
+	}
```
DavidVorick: These values don't reflect what is in master branch, but they do reflect what is in the v0.6.0 release.

DavidVorick: I used git blame to find other fields where compatibility had been accidentally been broken. I'm not sure the best way to be 100% sure that others weren't missed.

I only checked the host fields, not fields from other modules.

There may be some chaos this upgrade.

lukechampine: shouldn't this be `COMPAT v0.6.0`?

modules/host/persist.go
```diff
@@ -171,6 +171,44 @@ func (h *Host) load() error {
...
+	err = h.dependencies.loadFile(persistMetadata, &compatPersistence, filepath.Join(h.persistDir, settingsFile))
+	if err != nil {
+		return err
+	}
+	// Load the compat values, but only if the compat values are non-zero and
+	// the real values are zero.
+	if compatPersistence.FinancialMetrics.PotentialStorageRevenue.Cmp(types.ZeroCurrency) > 0 && p.FinancialMetrics.PotentialStorageRevenue.Cmp(types.ZeroCurrency) == 0 {
```
lukechampine: use `IsZero` to make these more readable and concise

modules/host/persist.go
DavidVorick: I thought the convention was to label the COMPAT for the version where the change was introduced, to give us an idea of how long people have had to upgrade away from the change.

modules/host/persist.go
DavidVorick: forgot we had that function.

DavidVorick:
pushed into #1254

PR #1261, opened 2016-06-18
hostdb only returns hosts if they are accepting contracts

DavidVorick:


modules/renter/hostdb/scan_test.go
```diff
@@ -112,7 +113,8 @@ func TestThreadedProbeHosts(t *testing.T) {
 			encoding.ReadObject(ourConn, new(types.Specifier), types.SpecifierLen)
 			// write host settings
 			crypto.WriteSignedObject(ourConn, modules.HostExternalSettings{
-				NetAddress: "probed",
+				AcceptingContracts: true,
+				NetAddress:         "probed",
 			}, sk)
 			ourConn.Close()
 		}()
```
DavidVorick: These 4 lines are the only non-testing changes.

modules/renter/hostdb/weightedlist.go
```diff
@@ -183,13 +183,18 @@ func (hdb *HostDB) RandomHosts(n int, ignore []modules.NetAddress) (hosts []modu
...
 		node, err := hdb.hostTree.nodeAtWeight(types.NewCurrency(randWeight))
 		if err != nil {
+			build.Critical("nodeAtWeight is returning and error:", err)
 			break
 		}
-		hosts = append(hosts, node.hostEntry.HostDBEntry)
+		// Only return the host if they are accepting contracts.
+		if node.hostEntry.HostDBEntry.AcceptingContracts {
+			hosts = append(hosts, node.hostEntry.HostDBEntry)
+		}
```
lukechampine: you mean these four lines? :P

lukechampine:
From the name alone, it's not entirely clear that `RandomHosts` will only return active hosts. I think it's fine for now, but something to be aware of.

PR #1264, opened 2016-06-19
add contract count to host financial metrics + siac

DavidVorick:


modules/host/storageobligations.go
```diff
@@ -372,6 +372,7 @@ func (h *Host) addStorageObligation(so *storageObligation) error {
 
 	// Update the host financial metrics with regards to this storage
 	// obligation.
+	h.financialMetrics.ContractCount++
```
lukechampine: seems a little odd to maintain the current ContractCount in memory, but load it from `Stats().KeyN`...
I dunno, it's fine.

DavidVorick: Keeping it in memory means you don't have to do database I/O every time you want the value.

Originally I had it as just in memory, but then I realized that anyone upgrading from a lower version would get some weird values.

PR #1265, opened 2016-06-20
add a host competitive price to siac

DavidVorick:
renter module is required to get the estimation.

### DIFF
--- a/siac/hostcmd.go
+++ b/siac/hostcmd.go
@@ -117,14 +117,31 @@ sector may impact host revenue.`,
 // Prints info about the host and its storage folders.
 func hostcmd() {
 	hg := new(api.HostGET)
-	err := getAPI("/host", &hg)
+	err := getAPI("/host", hg)
 	if err != nil {
 		die("Could not fetch host settings:", err)
 	}
 	sg := new(api.StorageGET)
-	err = getAPI("/storage", &sg)
+	err = getAPI("/storage", sg)
 	if err != nil {
 		die("Could not fetch storage info:", err)
+	}
+
+	// Determine the competitive price string.
+	ah := new(api.ActiveHosts)
+	var competitivePrice string
+	err = getAPI("/renter/hosts/active?numhosts=24", ah)
+	if err != nil || len(ah.Hosts) == 0 {
+		competitivePrice = "Unavailable"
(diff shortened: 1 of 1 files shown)


siac/hostcmd.go
```diff
@@ -117,16 +117,30 @@ sector may impact host revenue.`,
...
+	var competitivePrice string
+	err = getAPI("/renter/hosts/active?numhosts=24", ah)
+	if err != nil {
+		competitivePrice = "Unavailable"
+	} else {
+		var sum types.Currency
+		for _, host := range ah.Hosts {
+			sum = sum.Add(host.StoragePrice)
+		}
+		competitivePrice = currencyUnits(sum.Div64(26).Mul(modules.BlockBytesPerMonthTerabyte)) // 24 hosts were selected, but the division was 26 to lowball the average slightly.
```
lukechampine: the call might not return 24 hosts. Should divide by number of hosts and then `MulFloat` by some fudge factor.
As a consequence, you need to avoid divide-by-zero by checking for `len(ah.Hosts) == 0` in the above case. I recommend:
```go
if err != nil || len(ah.Hosts) == 0 {
    competitivePrice = "Unavailable"
} else {
   // ...
}
```

lukechampine:
sorry for the merge commit, I didn't think I was going to suggest any changes. Just push --force to remove it.

siac/hostcmd.go
DavidVorick: good catch

siac/hostcmd.go
```diff
@@ -117,16 +117,33 @@ sector may impact host revenue.`,
...
 		die("Could not fetch storage info:", err)
 	}
 
+	// Determine the competitive price string.
+	ah := new(api.ActiveHosts)
+	var competitivePrice string
+	err = getAPI("/renter/hosts/active?numhosts=24", ah)
+	if err != nil {
+		competitivePrice = "Unavailable"
+	} else if len(ah.Hosts) > 0 {
```
lukechampine: if `err == nil` but `len(ah.Hosts) == 0`, this will result in `competitivePrice = ""`

DavidVorick: sloppy

thanks

PR #1266, opened 2016-06-20
Add clean close to ConsensusSet

### LINKED ISSUE #1253: ConsensusSet needs clean close
I encountered this non-deterministic test failure which would be fixed using a `sync.ThreadGroup`.
```
panic: cannot write to the file after it has been closed
goroutine 2279 [running]:
panic(0x6659c0, 0xc820398a10)
	/Users/travis/.gimme/versions/go1.6.darwin.amd64/src/runtime/panic.go:464 +0x3ff
github.com/NebulousLabs/Sia/build.Critical(0xc8205a3d38, 0x1, 0x1)
	/Users/travis/gopath/src/github.com/NebulousLabs/Sia/build/critical.go:13 +0xa2
github.com/NebulousLabs/Sia/persist.(*closeableFile).Write(0xc820454f60, 0xc820423000, 0x6f, 0x80, 0x0, 0x0, 0x0)
	/Users/travis/gopath/src/github.com/NebulousLabs/Sia/persist/log.go:102 +0x1bf
log.(*Logger).Output(0xc82074c870, 0x2, 0xc82029f180, 0x40, 0x0, 0x0)
	/Users/travis/.gimme/versions/go1.6.darwin.amd64/src/log/log.go:166 +0x482
github.com/NebulousLabs/Sia/persist.(*Logger).Debugln(0xc820751640, 0xc8205a3f50, 0x2, 0x2)
	/Users/travis/gopath/src/github.com/NebulousLabs/Sia/persist/log.go:48 +0x97
github.com/NebulousLabs/Sia/modules/consensus.(*ConsensusSet).rpcRelayHeader.func2(0xc820229400, 0x21d0710, 0xc820341b20)
	/Users/travis/gopath/src/github.com/NebulousLabs/Sia/modules/consensus/synchronize.go:385 +0x30d
created by github.com/NebulousLabs/Sia/modules/consensus.(*ConsensusSet).rpcRelayHeader
	/Users/travis/gopath/src/github.com/NebulousLabs/Sia/modules/consensus/synchronize.go:387 +0x2e4
FAIL	github.com/NebulousLabs/Sia/modules/consensus	40.242s
```

VoidingWarranties:
Closes #1253

### DIFF
--- a/modules/consensus/accept.go
+++ b/modules/consensus/accept.go
@@ -236,6 +236,11 @@ func (cs *ConsensusSet) managedAcceptBlock(b types.Block) error {
 			// a new block to the cache.
 			if err == errFutureTimestamp {
 				go func() {
+					if cs.tg.Add() != nil {
+						return
+					}
+					defer cs.tg.Done()
+
 					time.Sleep(time.Duration(b.Timestamp-(types.CurrentTimestamp()+types.FutureThreshold)) * time.Second)
 					err := cs.AcceptBlock(b)
 					if err != nil {
--- a/modules/consensus/consensusset.go
+++ b/modules/consensus/consensusset.go
@@ -152,6 +152,15 @@ func New(gateway modules.Gateway, persistDir string) (*ConsensusSet, error) {
 		gateway.RegisterRPC("SendBlk", cs.rpcSendBlk)
 		gateway.RegisterConnectCall("SendBlocks", cs.threadedReceiveBlocks)
 
+		// Unregister RPCs on Close.
+		cs.tg.OnStop(func() {
+			cs.gateway.UnregisterRPC("SendBlocks")
+			cs.gateway.UnregisterRPC("RelayBlock") // COMPATv0.5.1
+			cs.gateway.UnregisterRPC("RelayHeader")
+			cs.gateway.UnregisterRPC("SendBlk")
+			cs.gateway.UnregisterConnectCall("SendBlocks")
+		})
+
 		// Mark that we are synced with the network.
 		cs.mu.Lock()
 		cs.synced = true
@@ -201,14 +210,6 @@ func (cs *ConsensusSet) Close() error {
 
 	cs.mu.Lock()
 	defer cs.mu.Unlock()
-
-	if cs.synced {
-		cs.gateway.UnregisterRPC("SendBlocks")
-		cs.gateway.UnregisterRPC("RelayBlock") // COMPATv0.5.1
-		cs.gateway.UnregisterRPC("RelayHeader")
-		cs.gateway.UnregisterRPC("SendBlk")
-		cs.gateway.UnregisterConnectCall("SendBlocks")
-	}
 
 	var errs []error
 	if err := cs.db.Close(); err != nil {
--- a/modules/consensus/consensusset.go
+++ b/modules/consensus/consensusset.go
@@ -146,10 +146,10 @@ func New(gateway modules.Gateway, persistDir string) (*ConsensusSet, error) {
 		}
 
 		// Register RPCs
-		gateway.RegisterRPC("SendBlocks", cs.rpcSendBlocks)
-		gateway.RegisterRPC("RelayBlock", cs.rpcRelayBlock) // COMPATv0.5.1
-		gateway.RegisterRPC("RelayHeader", cs.rpcRelayHeader)
-		gateway.RegisterRPC("SendBlk", cs.rpcSendBlk)
+		gateway.RegisterRPC("SendBlocks", cs.threadedRPCSendBlocks)
+		gateway.RegisterRPC("RelayBlock", cs.threadedRPCRelayBlock) // COMPATv0.5.1
+		gateway.RegisterRPC("RelayHeader", cs.threadedRPCRelayHeader)
+		gateway.RegisterRPC("SendBlk", cs.threadedRPCSendBlk)
 		gateway.RegisterConnectCall("SendBlocks", cs.threadedReceiveBlocks)
 
 		// Unregister RPCs on Close.
--- a/modules/consensus/consensusset.go
+++ b/modules/consensus/consensusset.go
@@ -14,6 +14,7 @@ import (
 	"github.com/NebulousLabs/Sia/encoding"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/persist"
+	"github.com/NebulousLabs/Sia/sync"
 	"github.com/NebulousLabs/Sia/types"
 
 	"github.com/NebulousLabs/bolt"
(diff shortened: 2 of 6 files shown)


VoidingWarranties:
I did not add `cs.tg.Add` to exported methods that use the database, as bolt returns `bolt. ErrDatabaseReadOnly` when you try to view or update a closed database.

VoidingWarranties:
Although it may be preferable to use `cs.tg.Add` in exported methods that use the database **and also return an error** so that `sync.ErrStopped` is returned instaed of `bolt. ErrDatabaseNotOpen`.

modules/consensus/consensusset.go
```diff
@@ -145,6 +152,14 @@ func New(gateway modules.Gateway, persistDir string) (*ConsensusSet, error) {
 		gateway.RegisterRPC("SendBlk", cs.rpcSendBlk)
 		gateway.RegisterConnectCall("SendBlocks", cs.threadedReceiveBlocks)
 
+		cs.tg.OnStop(func() {
```
DavidVorick: Can you add a comment `// Unregister RPCs on shutdown`?

modules/consensus/synchronize.go
```diff
@@ -237,6 +242,11 @@ func (cs *ConsensusSet) threadedReceiveBlocks(conn modules.PeerConn) (returnErr
 // that BlockHeight onwards are returned. It also sends a boolean indicating
 // whether more blocks are available.
 func (cs *ConsensusSet) rpcSendBlocks(conn modules.PeerConn) error {
+	if err := cs.tg.Add(); err != nil {
+		return err
+	}
+	defer cs.tg.Done()
```
DavidVorick: should either be `threadedRPCSendBlocks` or should not have the `tg` code. Same applies to the other RPCs.

modules/consensus/synchronize.go
```diff
@@ -380,6 +405,11 @@ func (cs *ConsensusSet) rpcRelayHeader(conn modules.PeerConn) error {
 	if err == errOrphan {
 		// If the header is an orphan, try to find the parents.
 		go func() {
```
DavidVorick: Is there a reason that we want the handling of `errOrphan` to be done in a separate goroutine?

modules/consensus/synchronize.go
```diff
@@ -392,6 +422,11 @@ func (cs *ConsensusSet) rpcRelayHeader(conn modules.PeerConn) error {
 	// If the header is valid and extends the heaviest chain, fetch, accept it,
 	// and broadcast it.
 	go func() {
+		if cs.tg.Add() != nil {
```
DavidVorick: same here, is there a reason that we want this handling done in a separate thread? It's already in a separate thread and the current thread has nothing more to do.

Does it close the conn with the remote gateway?

modules/consensus/synchronize.go
VoidingWarranties: All RPCs should have `tg` code so it's redundant to have `threadedRPC`

DavidVorick: I believe the host has them as `threadedRPC`, if you make it so that the prefix `rpc` implies `threaded` you've added to the vocabulary and introduced a risk of functions in the gateway that prefix with `rpc` but aren't actually RPCs, something like `rpcFactory` or whatever (forgive the crappy example).

Better to have a smaller number of keywords.

modules/consensus/accept.go
```diff
@@ -236,6 +236,11 @@ func (cs *ConsensusSet) managedAcceptBlock(b types.Block) error {
 			// a new block to the cache.
 			if err == errFutureTimestamp {
 				go func() {
+					if cs.tg.Add() != nil {
+						return
+					}
+					defer cs.tg.Done()
+
 					time.Sleep(time.Duration(b.Timestamp-(types.CurrentTimestamp()+types.FutureThreshold)) * time.Second)
```
lukechampine: Need to use a `select` here with `time.After`; otherwise the goroutine will run for a long time before shutdown can complete.

modules/consensus/consensusset.go
```diff
@@ -131,6 +133,11 @@ func New(gateway modules.Gateway, persistDir string) (*ConsensusSet, error) {
 	}
 
 	go func() {
+		if cs.tg.Add() != nil {
+			return
+		}
+		defer cs.tg.Done()
```
lukechampine: ditto; this means `Close` will not return until `cs.threadedInitialBlockchainDownload` has finished. That function should be modified to exit early (blocking for a maximum of one iteration) during shutdown. Specifically, it should check for `tg.IsStopped`, and it should `select` on `time.After` instead of sleeping.

modules/consensus/consensusset.go
```diff
@@ -139,12 +146,21 @@ func New(gateway modules.Gateway, persistDir string) (*ConsensusSet, error) {
...
 
+		// Unregister RPCs on Close.
+		cs.tg.OnStop(func() {
+			cs.gateway.UnregisterRPC("SendBlocks")
+			cs.gateway.UnregisterRPC("RelayBlock") // COMPATv0.5.1
+			cs.gateway.UnregisterRPC("RelayHeader")
+			cs.gateway.UnregisterRPC("SendBlk")
+			cs.gateway.UnregisterConnectCall("SendBlocks")
+		})
+
```
VoidingWarranties: This doesn't work. If `Close` is called after the `RegisterRPC` calls but before this `OnStop` call, the `OnStop` call will return immediately without unregistering the RPCs.

VoidingWarranties: You could move these `UnregisterRPC` calls back to the `Close` method, but then an adversary could prevent the module from closing by repeatedly calling RPCs on us.

The real problem is that registering the RPCs and setting `synced` to true is not atomic. But it could be made atomic by moving the write lock to surround the calls to `RegisterRPC` too.

lukechampine: Interesting, and it seems like this would apply to many uses of `OnStop`. I think you could get around this like so:
```go
unregister := func() { /* blah blah blah */ }
err = cs.tg.OnStop(unregister)
if err != nil {
    // already stopped
    unregister()
}
```
But this raises the question: why not embed this logic directly in `OnStop`? After all, if the threadgroup is stopped, we want the function to execute! So I propose we modify `OnStop` like so:
```go
func (tg *ThreadGroup) OnStop(fn func()) {
	if tg.IsStopped() {
		fn() // call function immediately instead of returning an error
	} else {
		tg.mu.Lock()
		tg.stopFns = append(tg.stopFns, fn)
		tg.mu.Unlock()
	}
}
```

DavidVorick:
I'm closing this in favor of implementing it myself. This PR has a number of mistakes related to locking, will go through later to point out examples. This PR was written  before the conception of the Flush call, which changes the approaches that must be used.

PR #1270, opened 2016-06-20
remove t.Parallel from networking tests

DavidVorick:
The networking tests wait until a network action is complete, but have a
timeout in case the action doesn't complete in time. Running the tests
in parallel makes it much more likely that the tests don't complete in
time because other tests are hogging all of the cpu and disk resources.

For tests which have sensitive timeouts, t.Parallel() has been removed.

### DIFF
--- a/modules/consensus/synchronize_test.go
+++ b/modules/consensus/synchronize_test.go
@@ -27,7 +27,6 @@ func TestSynchronize(t *testing.T) {
 	if testing.Short() {
 		t.SkipNow()
 	}
-	t.Parallel()
 	cst1, err := createConsensusSetTester("TestSynchronize1")
 	if err != nil {
 		t.Fatal(err)
@@ -108,7 +107,6 @@ func TestBlockHistory(t *testing.T) {
 	if testing.Short() {
 		t.SkipNow()
 	}
-	t.Parallel()
 
 	cst, err := createConsensusSetTester("TestBlockHistory")
 	if err != nil {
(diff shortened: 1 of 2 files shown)


DavidVorick:
should note that most of the consensus tests still utilize `t.Parallel()`

modules/consensus/synchronize_test.go
```diff
@@ -108,7 +107,6 @@ func TestBlockHistory(t *testing.T) {
 	if testing.Short() {
 		t.SkipNow()
 	}
-	t.Parallel()
```
VoidingWarranties: I don't think this test is a networking test nor does it have timeouts

VoidingWarranties:
This increases the consensus test time by 30 seconds. A small price for more determinism.

modules/consensus/synchronize_test.go
mnsl: `TestSynchronize` has been timing out for me today when run in parallel (running the test on its own is fine).  The test times out on line 84 while waiting for `cst1.cs.dbBlockHeight()` to catch up to `cst2.cs.dbBlockHeight()` after reconnecting the gateways.

DavidVorick: It was causing my laptop to time out a bunch as well, I think that something with the I/O caching and screwing up Bolt would be related. Or maybe the scheduler was just having trouble.

Post this PR, we should continue to pay attention to the timeouts. My 8-core desktop + SSD definitely speeds up when we have every test running t.Parallel, but travis is a lot more resource constrained. It might we worth only subjecting a test to t.Parallel if it spends a significant amount of time idling.

VoidingWarranties: Oh I thought it was `TestBlockHistory`. Weird, it must be a bug in GitHub.

screen shot 2016-06-20 at 9 20 22 pm

PR #1272, opened 2016-06-20
Close a few stray consensusSetTesters

mnsl:
This will hopefully clean testing up a bit

### DIFF
--- a/modules/consensus/changelog_test.go
+++ b/modules/consensus/changelog_test.go
@@ -21,6 +21,7 @@ func TestIntegrationChangeLog(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	defer cst.Close()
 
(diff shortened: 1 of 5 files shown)


mnsl:
Both builds failed. 
Linux didn't even make it past the short tests; it looks like there was a problem encrypting the `explorerTester`'s wallet in `explorer.TestImmediateBlockFacts`.  The PR only changes `consensus` tests, so perhaps this is a new non-deterministic failure.  I've never seen it before and it doesn't show up in any of the issues.

On OS X, two `consensus` tests fail: `TestIntegrationRelaySynchronize` and `TestIntegrationRPCSendBlocks`.  Both tests are in `synchronize_test.go`, which was unchanged in the PR and has been the source of many recent NDFs.  Hopefully #1270 (just merged) will resolve the problem.

mnsl:
Rebased, tests passed

PR #1273, opened 2016-06-21
Renter/Contractor tweaks for 1.0

lukechampine:


modules/renter/contractor/renew.go
```diff
@@ -104,7 +94,5 @@ func (c *Contractor) managedRenewContracts() {
 	}
 	err = c.saveSync()
 	c.mu.Unlock()
-	if err != nil {
-		c.log.Println("WARN: failed to save the contractor:", err)
```
DavidVorick: the contractor has died of dysentery

lukechampine: let the record show that David is losing it a little

PR #1274, opened 2016-06-21
implement muxado v2 transition

lukechampine:
yamux was the original choice, but produced "duplicate stream id" errors during `TestCallingRPCFromRPC`. I haven't yet identified the root cause of this, but it seems to be a "cross-network race condition" that can occur if the client and server initiate a stream at the same time.

This diff turned out to be pleasantly tiny thanks to Jordan's recent reorganization that cleanly split the old version code from the new.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,7 @@ dependencies:
 	go get -u github.com/NebulousLabs/entropy-mnemonics
 	go get -u github.com/NebulousLabs/go-upnp
 	go get -u github.com/NebulousLabs/muxado
+	go get -u github.com/inconshreveable/muxado
 	go get -u github.com/klauspost/reedsolomon
 	go get -u github.com/julienschmidt/httprouter
 	# Frontend Dependencies
--- a/modules/gateway/peers.go
+++ b/modules/gateway/peers.go
@@ -10,7 +10,6 @@ import (
 	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/encoding"
 	"github.com/NebulousLabs/Sia/modules"
-	"github.com/NebulousLabs/muxado"
 )
 
 const (
@@ -77,7 +76,7 @@ func (s invalidVersionError) Error() string {
 
 type peer struct {
 	modules.Peer
-	sess muxado.Session
+	sess streamSession
 }
 
 func (p *peer) open() (modules.PeerConn, error) {
@@ -212,7 +211,7 @@ func (g *Gateway) managedAcceptConnOldPeer(conn net.Conn, remoteVersion string) 
 			Inbound:    true,
 			Version:    remoteVersion,
 		},
-		sess: muxado.Server(conn),
+		sess: newMuxadoV1Server(conn),
 	})
 
 	return nil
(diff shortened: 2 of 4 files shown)


DavidVorick:
The linux build took 1 minute to realize that `make` failed, and the OSX build took 5 minutes to realize that `make` failed. OSX Travis seems to have gotten some significant downgrades.

DavidVorick:
I would like to know why the yamux stuff isn't working, but the fact that it's having trouble straight out of the gate is pretty significant. And it looks like spdy has been deprecated in favor of http2?

DavidVorick:
Can you explain in more detail what's happening here? It seems like a separate instance of muxado gets spun up for each incoming connection?

Do you need two different ports? Why or why not, and what's stopping us from just having direct connections to our peers?

lukechampine:
I tried to wire up spdystream as well, just to see if it would pass, but they have an entirely different interface that would require a fair amount of refactoring. It also requires you to specify an `http.Header` when you create a stream, which is something of a red flag.

When you connect to a peer, you call `muxado.Server` on the connection and the peer calls `muxado.Client`. (There's no meaningful distinction between the `Server` and `Client`; all that matters it that there's one of each.) This creates a new "session" that orchestrates streams. Streams are an abstraction that allow you to behave as though you could do a separate `net.Dial` for every RPC. In reality, all data is being sent over a single TCP connection, i.e. between one port on the client and one port on the server. There are a few benefits to this: the overhead of creating/destroying a TCP connection is amortized; you don't need to worry about running out of ports; and you can communicate with peers behind NAT. The latter is the most important feature, since many peers are behind NAT, meaning you cannot directly `net.Dial` them. However, if they dial you, you can respond using the same TCP connection. That's why it's important to sustain that connection, because if it drops, you can no longer talk to that peer.

Speaking of which: I don't think muxado v2 performs keep-alive (heartbeats) by default. It should be simple to restore that functionality though, since they specifically provide a wrapper for it.

lukechampine:
`TestIntegrationSendBlocksStalls` seems to time out with this change.

VoidingWarranties:
`threadedReceiveBlocks` and `threadedInitialBlockchainDownload` necessarily depended on implementation details of the old muxado. Specifically they relied on read / write timeout errors being the string "Read timeout" and "Write timeout", respectively. This was necessary because the old muxado didn't support the `net.Error` interface (https://github.com/inconshreveable/muxado/issues/14). It appears muxado v2 also doesn't support the `net.Error` interface. It also seems like muxado v2 doesn't support read timeouts, only write timeouts.

VoidingWarranties:
It looks like the lack of a read timeout is the cause of the test timeout. `threadedReceiveBlocks` never returns from its first read.

VoidingWarranties:
Yamux supports read / write timeouts (both are returned as `yamux.ErrTimeout`. They don't conform to the `net.Error` interface, but it would be easy to make them conform or do a similar kludge as we did with muxado.

Wrapping the `Read` and `Write` methods of `modules.PeerConn` is an easy solution.
```go
func (pc peerConn) Read(...) {
    n, err := pc.Conn.Read(...)
    if pc.muxer == muxerMuxadoV1 && (err.Error() == "Read timeout" || err.Error() == "Write timeout") {
        return n, netError{
            err: err,
            timeout: true,
        }
    } else if pc.muxer == muxerYamux && err.Error() == "i/o deadline reached" {
        return n, netError{
            err: err,
            timeout: true,
        }
    }
    return n, err
}
```

lukechampine:
that is pretty bogus. Maybe we ought to switch back to yamux and get to the bottom of the "duplicate stream" bug, then.

DavidVorick:
closing this as my understanding is that the goal is to pursue yamux.

PR #1278, opened 2016-06-22
Implement Close for the Wallet

mnsl:
The wallet needs a `Close` method for clean shutdown during testing and serving.

### DIFF
--- a/api/server.go
+++ b/api/server.go
@@ -114,18 +114,6 @@ func (srv *Server) Close() error {
 	srv.wg.Wait()
 
 	// Safely close each module.
-	// TODO: close transaction pool
-
-	// wallet has special closing mechanics
-	if srv.wallet != nil {
-		// TODO: close wallet and lock the wallet in the wallet's Close method.
-		if srv.wallet.Unlocked() {
-			if err := srv.wallet.Lock(); err != nil {
-				errs = append(errs, fmt.Errorf("wallet.Lock failed: %v", err))
-			}
-		}
-	}
-
 	mods := []struct {
 		name string
 		c    io.Closer
@@ -134,6 +122,8 @@ func (srv *Server) Close() error {
 		{"renter", srv.renter},
 		{"explorer", srv.explorer},
 		{"miner", srv.miner},
+		{"wallet", srv.wallet},
+		{"tpool", srv.tpool},
 		{"consensus", srv.cs},
 		{"gateway", srv.gateway},
 	}
--- a/modules/wallet/update.go
+++ b/modules/wallet/update.go
@@ -209,6 +209,13 @@ func (w *Wallet) applyHistory(cc modules.ConsensusChange) {
 // ProcessConsensusChange parses a consensus change to update the set of
 // confirmed outputs known to the wallet.
 func (w *Wallet) ProcessConsensusChange(cc modules.ConsensusChange) {
+	if err := w.tg.Add(); err != nil {
+		// The wallet should gracefully reject updates from the consensus set
+		// or transaction pool that are sent after the wallet's Close method
+		// has closed the wallet's ThreadGroup.
+		return
+	}
+	defer w.tg.Done()
 	w.mu.Lock()
 	defer w.mu.Unlock()
 	w.updateConfirmedSet(cc)
@@ -219,12 +226,18 @@ func (w *Wallet) ProcessConsensusChange(cc modules.ConsensusChange) {
 // ReceiveUpdatedUnconfirmedTransactions updates the wallet's unconfirmed
 // transaction set.
 func (w *Wallet) ReceiveUpdatedUnconfirmedTransactions(txns []types.Transaction, _ modules.ConsensusChange) {
+	if err := w.tg.Add(); err != nil {
+		// Gracefully reject transactions if the wallet's Close method has
+		// closed the wallet's ThreadGroup already.
+		return
+	}
+	defer w.tg.Done()
 	w.mu.Lock()
 	defer w.mu.Unlock()
 
 	w.unconfirmedProcessedTransactions = nil
 	for _, txn := range txns {
-		// To save on  code complexity, relveancy is determined while building
+		// To save on code complexity, relevancy is determined while building
 		// up the wallet transaction.
 		relevant := false
 		pt := modules.ProcessedTransaction{
--- a/modules/wallet/wallet.go
+++ b/modules/wallet/wallet.go
@@ -5,9 +5,11 @@ package wallet
 
 import (
 	"errors"
+	"fmt"
 	"sort"
 	"sync"
 
+	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/persist"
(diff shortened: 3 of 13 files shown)


api/server.go
```diff
@@ -134,6 +122,10 @@ func (srv *Server) Close() error {
 		{"renter", srv.renter},
 		{"explorer", srv.explorer},
 		{"miner", srv.miner},
+		{"wallet", srv.wallet},
+		//  #1230 makes the tpoolTester interface an io.Closer;
+		//  waiting for merge.
+		//	{"tpool", srv.tpool},
```
lukechampine: needs update

modules/wallet/wallet_test.go
```diff
@@ -132,8 +132,16 @@ func createBlankWalletTester(name string) (*walletTester, error) {
...
 func (wt *walletTester) closeWt() {
-	err := wt.gateway.Close()
-	if err != nil {
+	errs := []error{
+		wt.gateway.Close(),
+		wt.cs.Close(),
+		// tpoolTester.Close and TestMiner.Close are implemented in #1230;
+		// waiting for merge.
+		// wt.tpool.Close(),
+		//	wt.miner.Close(),
```
lukechampine: ditto

mnsl:
fixed, ready to go

mnsl:
* Updated now that #1230 has been merged (`TestMiner` and `tpoolTester` now have `Close` methods).
* Added `TestClose` to test the wallet's new `Close` method.  It's pretty basic.

api/server.go
```diff
@@ -114,18 +114,6 @@ func (srv *Server) Close() error {
...
-	// wallet has special closing mechanics
-	if srv.wallet != nil {
-		// TODO: close wallet and lock the wallet in the wallet's Close method.
-		if srv.wallet.Unlocked() {
-			if err := srv.wallet.Lock(); err != nil {
-				errs = append(errs, fmt.Errorf("wallet.Lock failed: %v", err))
-			}
-		}
-	}
-
```
VoidingWarranties: :+1:

modules/wallet/wallet.go
```diff
@@ -137,6 +139,29 @@ func New(cs modules.ConsensusSet, tpool modules.TransactionPool, persistDir stri
...
+// Close terminates all ongoing processes involving the wallet, enabling
+// garbage collection.
+func (w *Wallet) Close() error {
+	var errs []error
+	// Lock the wallet outside of mu.Lock because Lock uses its own mu.Lock.
+	if w.Unlocked() {
+		if err := w.Lock(); err != nil {
+			errs = append(errs, err)
+		}
+	}
```
VoidingWarranties: This does not gaurantee that the wallet is locked at the time `Close()` returns as you could unlock the wallet in another goroutine after this block has executed, but before the `Lock` is obtained. This is probably extremely unlikely given the wallet's unlock time.

You could fix it with a `sync.ThreadGroup`, but I'm not sure that this is a semantic use of a `ThreadGroup`.

mnsl: Latest push fixes this by adding a `ThreadGroup`

mnsl:
Added a `ThreadGroup` to the wallet to prevent the vulnerability Jordan pointed out (above).  

I wasn't sure which methods were/weren't worth tracking with the `ThreadGroup`, so I may have been more liberal than necessary.

modules/wallet/update.go
```diff
@@ -219,6 +223,10 @@ func (w *Wallet) ProcessConsensusChange(cc modules.ConsensusChange) {
 // ReceiveUpdatedUnconfirmedTransactions updates the wallet's unconfirmed
 // transaction set.
 func (w *Wallet) ReceiveUpdatedUnconfirmedTransactions(txns []types.Transaction, _ modules.ConsensusChange) {
+	if err := w.tg.Add(); err != nil {
+		panic(err)
+	}
+	defer w.tg.Done()
```
lukechampine: we don't allow bare panics in our code. Use `build.Critical` (which panics in debug mode and logs otherwise) or wrap the panic in `if build.DEBUG`

DavidVorick: I'm pretty sure you can trigger this codepath anyway, it should just gracefully return. There's a race condition with how the consensus updates and tpool updates work.

mnsl: There are a lot of bare panics already in the wallet code, should that be fixed in a later PR?  

David, do you mean that it's not necessary to check `err := w.tg.Add()` here at all?  

And why do the consensus set and transaction pool both need to update the wallet, especially if it introduces a race condition?

lukechampine: The wallet only unsubscribes from the consensus and tpool after `tg.Stop` returns. In the interim, the consensus or tpool could send another update, which the wallet should gracefully reject. That's why David is advocating a `return` here instead of a `panic`.

mnsl: OK, I see

modules/wallet/update.go
```diff
@@ -219,12 +226,16 @@ func (w *Wallet) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+		build.Critical(err)
+	}
+	defer w.tg.Done()
 	w.mu.Lock()
 	defer w.mu.Unlock()
 
 	w.unconfirmedProcessedTransactions = nil
 	for _, txn := range txns {
-		// To save on  code complexity, relveancy is determined while building
+		// To save on  code complexity, relevancy is determined while building
```
VoidingWarranties: There's two spaces before "code complexity"

VoidingWarranties:
There are a few more methods that need to make use of the `ThreadGroup`. For the wallet specifically I think it's okay to use the `ThreadGroup` more liberally because it handles your money. It is problematic if a client thinks they've manipulated the wallet's state, but actually haven't because it's closed.

* `Lock` because it uses `w.log`.
* `SendSiacoins` and `SendSiafunds` because it's probably a bad idea to try to send siacoins/funds when the wallet is closed. Also I'm surprised this method doesn't have logging. It seems to me like it should log whenever you send make a transaction, but maybe @DavidVorick could comment on that.
* `CreateBackup` and `LoadBackup` anything dealing with the filesystem should probably fail when the wallet is closed. I know `LoadBackup` is commented out but when we eventually implement this we might forget to add it to the thread group.
* `MergeWallet` same reason as above.
* `NextAddress` manipulates the wallet's filesystem state.
* `LoadSeed`, `LoadSiagKeys`, `Load033xWallet` same reason as above.

ACK once these are added to the thread group.

mnsl:
I don't think we can add `Lock` to the threadgroup -- `Lock` is called after the threadgroup is stopped in `Wallet.Close`.

mnsl:
The threadgroup now tracks all exported functions that change the wallet's state, except for `Lock` (which is called in `Close` after the threadgroup has already been stopped).

modules/wallet/update.go
```diff
@@ -219,12 +226,16 @@ func (w *Wallet) ProcessConsensusChange(cc modules.ConsensusChange) {
 // ReceiveUpdatedUnconfirmedTransactions updates the wallet's unconfirmed
 // transaction set.
 func (w *Wallet) ReceiveUpdatedUnconfirmedTransactions(txns []types.Transaction, _ modules.ConsensusChange) {
+	if err := w.tg.Add(); err != nil {
+		build.Critical(err)
```
lukechampine: shouldn't this be a gracefully rejection too, like `ProcessconsensusChange`?

VoidingWarranties:
It's not critical to add `Lock` to the thread group for safety, but it would be nice to add it for correctness. Otherwise calling `Lock` after the wallet is closed will result in a `build.Critical`. You can lock the wallet from an `OnStop` and then adding it to the thread group is no longer a problem.

PR #1280, opened 2016-06-23
Change OnStop behavior when called on a stopped ThreadGroup

mnsl:
`tg.OnStop(fn)` now calls `fn` immediately if `tg` is already closed rather than just returning `ErrClosed`.
Note that `OnStop` no longer returns an error.

See discussion in #1266 for context.

### DIFF
--- a/sync/threadgroup_test.go
+++ b/sync/threadgroup_test.go
@@ -54,16 +54,12 @@ func TestThreadGroupStop(t *testing.T) {
 		t.Error("IsStopped returns false on stopped ThreadGroup")
 	}
 
-	// Add, Stop, and RegisterCloser should return errors
+	// Add and Stop should return errors
 	err = tg.Add()
 	if err != ErrStopped {
 		t.Error("expected ErrStopped, got", err)
 	}
 	err = tg.Stop()
-	if err != ErrStopped {
-		t.Error("expected ErrStopped, got", err)
-	}
-	err = tg.OnStop(nil)
 	if err != ErrStopped {
 		t.Error("expected ErrStopped, got", err)
 	}
@@ -136,10 +132,7 @@ func TestThreadGroupOnStop(t *testing.T) {
 
 	// create ThreadGroup and register the closer
 	var tg ThreadGroup
-	err = tg.OnStop(func() { l.Close() })
-	if err != nil {
-		t.Fatal(err)
-	}
+	tg.OnStop(func() { l.Close() })
 
 	// send on channel when listener is closed
 	var closed bool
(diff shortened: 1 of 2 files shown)


lukechampine:
Can you add a test that confirms that if the threadgroup is closed, `fn` will be called immediately? Something like:
```go
var tg ThreadGroup
var closed bool
tg.OnClose(func() { closed = true })
// closed should still be false
if closed {
    t.Fatal("close function should not have been called yet")
}
tg.Stop()
if !closed {
    t.Fatal("close function should have been called")
}

// if Stop has already been called, the close function should be called immediately
closed = false
tg.OnClose(func() { closed = true })
if !closed {
    t.Fatal("close function should have been called immediately")
}
```

mnsl:
Latest change implements `TestThreadGroupClosedOnStop` to check that `OnStop` behaves as expected when called on a stopped `ThreadGroup`

sync/threadgroup_test.go
```diff
@@ -173,6 +166,40 @@ func TestThreadGroupRace(t *testing.T) {
...
+	}
+
+	// this close function (which calls a goroutine) should also be called
+	// immediately since Stop has been called already
+	closedChan := make(chan struct{})
+	tg.OnStop(func() { go func() { closedChan <- struct{}{} }() })
+	select {
+	case <-closedChan:
+		break
+	case <-time.After(10 * time.Second):
```
VoidingWarranties: Since this is a short test the time should be < 3 seconds.

sync/threadgroup_test.go
```diff
@@ -173,6 +166,40 @@ func TestThreadGroupRace(t *testing.T) {
...
+	closed = false
+	tg.OnStop(func() { closed = true })
+	if !closed {
+		t.Fatal("close function should have been called immediately")
+	}
+
+	// this close function (which calls a goroutine) should also be called
+	// immediately since Stop has been called already
+	closedChan := make(chan struct{})
+	tg.OnStop(func() { go func() { closedChan <- struct{}{} }() })
```
VoidingWarranties: Channel receives unblock as soon as a channel is closed so you could write this line as
```go
tg.OnStop(func() { go close(closedChan) })
```
code golf

sync/threadgroup_test.go
DavidVorick: there should be no `time.Sleep` in a short test. The short tests are meant to be well under a second per package. Better to just make this a long test.

lukechampine: more to the point, I don't really see what this test adds that the previous test does not cover.

mnsl: yep, I'll get rid of it

mnsl:
Removed redundant test case in `TestThreadGroupClosedOnStop`

PR #1283, opened 2016-06-24
Renter 1.0 fixes

lukechampine:
The renter now caps how much collateral it will spend on a contract.
Old hosts will reject this (for revisions), so the cap is wrapped with a version conditional.

Also fix a bug that caused hostEditors to not report their host's NetAddress. This resulted in upload stalling because the host would not be properly removed from the host pool.

### DIFF
--- a/modules/renter/contractor/formcontract.go
+++ b/modules/renter/contractor/formcontract.go
@@ -12,9 +12,11 @@ import (
 
 var (
 	// the contractor will not form contracts above this price
-	maxStoragePrice = types.SiacoinPrecision.Mul64(500e3).Mul(modules.BlockBytesPerMonthTerabyte) // 500k SC / TB / Month
+	maxStoragePrice = types.SiacoinPrecision.Mul64(500e3).Div(modules.BlockBytesPerMonthTerabyte) // 500k SC / TB / Month
 	// the contractor will not download data above this price (3x the maximum monthly storage price)
 	maxDownloadPrice = maxStoragePrice.Mul64(3 * 4320)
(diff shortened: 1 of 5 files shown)


modules/renter/contractor/formcontract.go
```diff
@@ -12,9 +12,11 @@ import (
 
 var (
 	// the contractor will not form contracts above this price
-	maxStoragePrice = types.SiacoinPrecision.Mul64(500e3).Mul(modules.BlockBytesPerMonthTerabyte) // 500k SC / TB / Month
+	maxStoragePrice = types.SiacoinPrecision.Mul64(500e3).Div(modules.BlockBytesPerMonthTerabyte) // 500k SC / TB / Month
```
DavidVorick: oops

modules/renter/contractor/renew.go
```diff
@@ -18,6 +18,10 @@ func (c *Contractor) managedRenew(contract modules.RenterContract, numSectors ui
 	} else if host.StoragePrice.Cmp(maxStoragePrice) > 0 {
 		return modules.RenterContract{}, errTooExpensive
 	}
+	// cap host.MaxCollateral
+	if host.MaxCollateral.Cmp(maxCollateral) > 0 {
+		host.MaxCollateral = maxCollateral
+	}
```
DavidVorick: and this only applies to the new hosts? Not quite following how the old hosts end up not running this code.

DavidVorick: oh nvm, realized this is renew not revise. Get those two confused.

PR #1284, opened 2016-06-24
fix some misc hostdb errors

DavidVorick:
Should fix the 'negative currency' reports - problem was due to a race
condition.

Tree is sorted to have better structure.

Some style adjustments

Some consistency code was added during debugging. While useful, right
now it's not actually used anywhere, so it's commented out. Will
probably come in handy later, so I left it in this commit.

### DIFF
--- a/modules/renter/hostdb/persist.go
+++ b/modules/renter/hostdb/persist.go
@@ -45,7 +45,7 @@ func (hdb *HostDB) load() error {
 		hdb.allHosts[data.AllHosts[i].NetAddress] = &data.AllHosts[i]
 	}
 	for i := range data.ActiveHosts {
-		hdb.insertNode(&data.AllHosts[i])
+		hdb.insertNode(&data.ActiveHosts[i])
 	}
 	hdb.lastChange = data.LastChange
 	return nil
--- a/modules/renter/hostdb/scan.go
+++ b/modules/renter/hostdb/scan.go
@@ -67,8 +67,8 @@ func (hdb *HostDB) decrementReliability(addr modules.NetAddress, penalty types.C
 	// database.
 	node, exists := hdb.activeHosts[addr]
 	if exists {
-		delete(hdb.activeHosts, entry.NetAddress)
 		node.removeNode()
+		delete(hdb.activeHosts, entry.NetAddress)
 	}
 
 	// If the reliability has fallen to 0, remove the host from the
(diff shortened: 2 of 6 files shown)


modules/renter/hostdb/scan.go
```diff
@@ -142,7 +142,7 @@ func (hdb *HostDB) threadedProbeHosts() {
 
 			// If 'maxActiveHosts' has not been reached, add the host to the
 			// activeHosts tree.
-			if _, exists := hdb.activeHosts[hostEntry.NetAddress]; !exists && len(hdb.activeHosts) < maxActiveHosts {
+			if _, exists := hdb.activeHosts[hostEntry.NetAddress]; exists || len(hdb.activeHosts) < maxActiveHosts {
```
DavidVorick: This guy was causing the negative currency errors and host tree restructuring. Lack of safety on the hostdb tree can be blamed for this.

Adding safety would be a pretty major refactor.

modules/renter/hostdb/persist.go
```diff
@@ -45,7 +45,7 @@ func (hdb *HostDB) load() error {
 		hdb.allHosts[data.AllHosts[i].NetAddress] = &data.AllHosts[i]
 	}
 	for i := range data.ActiveHosts {
-		hdb.insertNode(&data.AllHosts[i])
+		hdb.insertNode(&data.ActiveHosts[i])
```
lukechampine: :sob:

modules/renter/hostdb/weightedlist_consistency_test.go
```diff
@@ -0,0 +1,82 @@
...
+		fmt.Println("No Left")
+	}
+	if hn.right != nil {
+		fmt.Println("Node Right")
+		hn.right.Print()
+	} else {
+		fmt.Println("No Right")
+	}
+}
+*/
```
lukechampine: what's the deal with these? If they're actually helpful, they shouldn't be commented out.

DavidVorick: They are helpful for debugging, but not necessarily for testing. For example, printing a node is something you'd never do during testing.

modules/renter/hostdb/persist.go
DavidVorick: yeah totally my fault too

modules/renter/hostdb/weightedlist_consistency_test.go
lukechampine: just seems weird to me to purposely include commented-out code.

DavidVorick: I could change it to `String`. I can also uncomment them if you prefer.

PR #1285, opened 2016-06-24
renter no longer needs wallet

lukechampine:
Since revising contracts does not require wallet access, the renter no longer needs to call any methods of the wallet.


DavidVorick:
Build is failing with a race condition, probably worth investigating + fixing before a merge.

lukechampine:
fixed. I'm not sure why this didn't trigger earlier.

PR #1290, opened 2016-06-25
ConsensusSubscribe bugfix

### LINKED ISSUE #1287: problems during and after host consensus rescan
Hi,

I noticed a couple of effects during or after the host does a consensus rescan. (I don't have the exact sequence of events which triggered the original rescan, possibly I triggered it by removing the consensus set).

(i) after the rescan the host apparently advanced in block height by 2 for each block received. Seems to be because ConsensusSetSubscribe() leaves the host subscribed even if it returns an error. (i.e. in host's initConsensusSubscription() the ErrInvalidConsensusChangeID error is used to trigger the rescan, after which the host subscribes again, leaving it subscribed twice).

(ii) sometimes I saw some panics during the rescan, from within initRescan(). Apparently due to the requeue of action items for all storageobligations at the end of initRescan, where it can call queueActionItem with a height below the current blockHeight. (i.e. so.expiration()-revisionSubmissionBuffer for an older storageobligation).

Sorry I don't have proper errors to report for this ticket. I hope it helps anyway.

David

DavidVorick:
The commit chain is cleaner and easier to understand than the complete diff.

fixes #1287, as well as a deadlock in the contractor that will cause all of siad to freeze (deadlock exists in RC1 and RC2). 

This bug is pretty significant, this PR should be high priority.

### DIFF
--- a/modules/consensus.go
+++ b/modules/consensus.go
@@ -110,6 +110,10 @@ type (
 		// SiafundPoolDiffs are the siafund pool diffs that were applied to the
 		// consensus set in the recent change.
 		SiafundPoolDiffs []SiafundPoolDiff
+
+		// Synced indicates whether or not the ConsensusSet is synced with its
+		// peers.
+		Synced bool
 	}
 
 	// A SiacoinOutputDiff indicates the addition or removal of a SiacoinOutput in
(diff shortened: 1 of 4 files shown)


modules/consensus.go
```diff
@@ -110,6 +110,10 @@ type (
 		// SiafundPoolDiffs are the siafund pool diffs that were applied to the
 		// consensus set in the recent change.
 		SiafundPoolDiffs []SiafundPoolDiff
+
+		// Synced indicates whether or not the ConsensusSet is synced with its
+		// peers.
+		Synced bool
```
lukechampine: did we confirm that this doesn't screw up any persistence stuff?

DavidVorick: Just confirmed with some testing, doesn't screw up persistence stuff. Looked at the code as well.

PR #1292, opened 2016-06-27
Don't allow non-localhost api address bind without `--disable-security`

johnathanhowell:


siad/main.go
```diff
@@ -29,6 +29,7 @@ type Config struct {
 		APIaddr  string
 		RPCaddr  string
 		HostAddr string
+		AllowAPIBind bool
```
lukechampine: `go fmt`

modules/netaddress.go
```diff
@@ -59,6 +59,10 @@ func (na NetAddress) isLoopback() bool {
 	}
 	return false
 }
+// exported form of isLoopback
+func (na NetAddress) IsLoopback() bool {
+	return na.isLoopback()
+}
```
lukechampine: The proper thing to do here is to replace `isLoopback` with `IsLoopback`. There are very few instances that you need to change, so no excuses.

johnathanhowell: ack, wasn't sure if we wanted both.

siad/daemon.go
```diff
@@ -57,6 +57,10 @@ func processModules(modules string) (string, error) {
 func processConfig(config Config) (Config, error) {
 	var err error
 	config.Siad.APIaddr = processNetAddr(config.Siad.APIaddr)
+	apiaddr := modules.NetAddress(config.Siad.APIaddr)
+	if apiaddr.Host() != "" && !apiaddr.IsLoopback() && !config.Siad.AllowAPIBind {
+		return Config{}, errors.New("ERROR: You must pass --disable-security to bind Siad to a non-localhost address.")
```
lukechampine: errors should not start with a capital or end in punctuation. The `ERROR:` is also redundant.

lukechampine: I would structure it like this:
```go
if !config.Siad.AllowAPIBind {
    addr := modules.NetAddress(config.Siad.APIaddr)
    if !addr.IsLoopback() && addr.Host() != "" {
        return Config{}, errors.New("you must pass --disable-security to bind Siad to a non-localhost address")
    }
}
```
Although the `.Host() != ""` really shouldn't be necessary; that seems like a failing of `IsLoopback`.

siad/main.go
```diff
@@ -157,6 +158,7 @@ func main() {
 	root.Flags().StringVarP(&globalConfig.Siad.RPCaddr, "rpc-addr", "", ":9981", "which port the gateway listens on")
 	root.Flags().StringVarP(&globalConfig.Siad.Modules, "modules", "M", "cghmrtw", "enabled modules, see 'siad modules' for more info")
 	root.Flags().BoolVarP(&globalConfig.Siad.AuthenticateAPI, "authenticate-api", "", false, "enable API password protection")
+	root.Flags().BoolVarP(&globalConfig.Siad.AllowAPIBind, "disable-security", "", false, "allow siad to listen on a non-localhost address (DANGEROUS)")
```
DavidVorick: given that this only applies to the api, I would mark it as "disable-api-security".

I was also thinking we could have a flag to enable the user to view the wallet seeds. Right now, if the wallet is unlocked you can fetch the wallet seeds over the API. This should be disabled by default, would prevent a lot of the repeated attacks that we're seeing I think.

PR #1296, opened 2016-06-27
Finalize the hostdb

VoidingWarranties:
Move hostdb API endpoints from /renter/hosts to /hostdb

**This PR is incomplete.** The API documentation for the hostdb still needs to be updated to the new style. I'm waiting to update the documentation until I get comments about my concern below.

I was conflicted about whether to call the endpoint `/hostdb` or `/hosts`. `/hostdb` is more in line with naming endpoints after their modules (is it still the plan to make the hostdb its own module in the future?) but is potentially more confusing at first glance. Host DB might be confused as the database hosts use to store renter's files, instead of a database of hosts. `/hosts` make more intuitive sense, especially the full endpoint paths `/hosts/active` and `/hosts/all`) but is more easily mistaken for the `/host` endpoint.

I know this is nit-picky but since the API is being finalized I didn't want to skip over any considerations other people might have.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -163,8 +163,9 @@ func (srv *Server) initAPI(password string) {
 		router.POST("/renter/rename/*siapath", requirePassword(srv.renterRenameHandler, password))
 		router.POST("/renter/upload/*siapath", requirePassword(srv.renterUploadHandler, password))
 
-		router.GET("/renter/hosts/active", srv.renterHostsActiveHandler)
-		router.GET("/renter/hosts/all", srv.renterHostsAllHandler)
+		// HostDB endpoints.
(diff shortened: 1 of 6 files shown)


mnsl:
Can I ask why we're switching from `/renter/hosts` in the first place?  I agree with your concerns about `hostdb` and `hosts`, though I think `hostdb` is better.

VoidingWarranties:
My understanding is that the hostdb could be (is being?) used by the host to provide competitive price recommendations. It's also standard practice to make sure you're in the hostdb after your first time setting up your host.

mnsl:
Fair enough.  I vote `hostdb`

PR #1299, opened 2016-06-28
Fix JSON fieldname typos

mnsl:
Potentiale is not a word @VoidingWarranties

### DIFF
--- a/modules/host.go
+++ b/modules/host.go
@@ -36,7 +36,7 @@ type (
 		LockedStorageCollateral types.Currency `json:"lockedstoragecollateral"`
 		LostRevenue             types.Currency `json:"lostrevenue"`
 		LostStorageCollateral   types.Currency `json:"loststoragecollateral"`
-		PotentialStorageRevenue types.Currency `json:"potentialestoragerevenue"`
+		PotentialStorageRevenue types.Currency `json:"potentialstoragerevenue"`
 		RiskedStorageCollateral types.Currency `json:"riskedstoragecollateral"`
 		StorageRevenue          types.Currency `json:"storagerevenue"`
 		TransactionFeeExpenses  types.Currency `json:"transactionfeeexpenses"`


modules/host.go
```diff
@@ -36,7 +36,7 @@ type (
 		LockedStorageCollateral types.Currency `json:"lockedstoragecollateral"`
 		LostRevenue             types.Currency `json:"lostrevenue"`
 		LostStorageCollateral   types.Currency `json:"loststoragecollateral"`
-		PotentialStorageRevenue types.Currency `json:"potentialestoragerevenue"`
+		PotentialStorageRevenue types.Currency `json:"potentialstoragerevenue"`
```
DavidVorick: Didn't we already fix this?

modules/host/persist.go
```diff
@@ -209,7 +209,7 @@ func (h *Host) load() error {
 func (h *Host) loadCompat(p *persistence) error {
 	var compatPersistence struct {
 		FinancialMetrics struct {
-			PotentialStorageRevenue types.Currency `json:"potentialerevenue"`
+			PotentialStorageRevenue types.Currency `json:"potentialrevenue"`
```
DavidVorick: this is actually not a typo, this one is here on purpose as compatibility code to a previous typo.

mnsl: Oop, should have tested before pushing sorry

modules/host.go
VoidingWarranties: We fixed it in 11a0e34c8a8484b558fe6117fa1083c3fdef08e4 but then it looks like the change was reverted in 12f298da4f45c60acf39876c711547899d4ef310. Probably a bad merge conflict resolution.

lukechampine: yeah... weird.
stepping through the commits of #1254, it looks like the typo was restored by a faulty rebase.
In 11a0e34c8a8484b558fe6117fa1083c3fdef08e4 *you* remove the typo, and in 12f298da4f45c60acf39876c711547899d4ef310, I add it back!

PR #1301, opened 2016-06-28
remove 32bit from make xc

DavidVorick:


lukechampine:
what's the story with `arm`?

DavidVorick:
oh, we should be targeting arm64 but not arm

PR #1303, opened 2016-06-28
remove arm

DavidVorick:


lukechampine:
strong commit name :muscle:

DavidVorick:
rmarm

easye:
Does this mean siad will no longer run on a Raspberry Pi?

DavidVorick:
@easye Sia has been unable to run on 32bit architectures for a while. The limitation is the 2GB address space. The database we use does a nmap which, as of around block 50,000, required more than 2GB of virtual space.

There's not an easy fix to this. The latest iteration of the raspi is a 64bit architecture, but I don't think there's a linux distribution available yet that runs on arm64.

Fornax96:
According to this page Ubuntu Server ARM 16.04 runs on 64 bit 
http://www.ubuntu.com/download/server/arm

easye:
@DavidVorick Well, fortunately the RPi3 has a 64bit mode…

easye:
If anybody wants a Docker container for siad on ARM, lemme know.

PR #1304, opened 2016-06-28
restore update functionality

lukechampine:
`siad` and `siac` can now be automatically updated (again).

GitHub is used as the update channel. Later, we may add some fallback channels.

Binaries are signed using the `developerKey` provided in `daemon.go`. All future releases must include signatures for the `siad` and `siac` binaries. In addition, I will be signing all my commits from now on, and other core devs should as well. I will also be signing the release tags. Since a tag references all the code at a given commit hash, signing it amounts to signing the entire codebase at that time. This provides verification for users who prefer to build from source.

Only .zip releases are supported. We can restore .tar.gz support later, but I don't view it as a high priority.
Releases must be tagged with a "clean" semver, e.g. `v1.2.3`. Suffixes such as `-beta` will confuse the update code.

`siac` is assumed to be located in the same folder as `siad`. If `siac` is not present, an error will be returned. (This is a limitation of `go-update`, which we can subvert if necessary.)

This PR is missing documentation for the update routes.

### DIFF
--- a/api/daemon.go
+++ b/api/daemon.go
@@ -149,8 +149,9 @@ func updateToRelease(release githubRelease) error {
 	if err != nil {
 		return err
 	}
-	// release is small enough to store in memory
-	content, err := ioutil.ReadAll(resp.Body)
+	// release should be small enough to store in memory (<10 MiB); use
+	// LimitReader to ensure we don't download more than 32 MiB
+	content, err := ioutil.ReadAll(io.LimitReader(resp.Body, 1<<25))
 	resp.Body.Close()
 	if err != nil {
 		return err
--- a/api/daemon.go
+++ b/api/daemon.go
@@ -1,14 +1,71 @@
 package api
 
 import (
+	"archive/zip"
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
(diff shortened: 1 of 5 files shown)


api/daemon.go
```diff
@@ -1,15 +1,72 @@
...
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
 	"math/big"
 	"net/http"
+	"path"
+	"path/filepath"
+	"runtime"
```
DavidVorick: oh my

lukechampine: I know, it takes like 12+ imports to accomplish this. Feels bad man

api/daemon.go
```diff
@@ -40,32 +97,149 @@ bwIDAQAB
...
+		return errors.New("couldn't find download URL for " + releaseName)
+	}
+
+	// download release archive
+	resp, err := http.Get(downloadURL)
+	if err != nil {
+		return err
+	}
+	// release is small enough to store in memory
+	content, err := ioutil.ReadAll(resp.Body)
```
DavidVorick: use an `io.LimitReader`

PR #1309, opened 2016-06-28
Finalize storage API

VoidingWarranties:
Docs are incomplete but will do for now. I'll fix the docs up in another PR when I wake up. The primary change is moving `/storage` -> `/host/storage`.

### DIFF
--- a/api/host.go
+++ b/api/host.go
@@ -32,7 +32,7 @@ type (
 	// to /host/storage - a bunch of information about the status of storage
 	// management on the host.
 	StorageGET struct {
-		StorageFolderMetadata []modules.StorageFolderMetadata
+		Folders []modules.StorageFolderMetadata `json:"folders"`
 	}
 )
 
@@ -124,9 +124,9 @@ func (srv *Server) hostAnnounceHandler(w http.ResponseWriter, req *http.Request,
 // storageHandler returns a bunch of information about storage management on
 // the host.
 func (srv *Server) storageHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
-	sfs := srv.host.StorageFolders()
+	folders := srv.host.StorageFolders()
 	sg := StorageGET{
-		StorageFolderMetadata: sfs,
+		Folders: folders,
 	}
 	writeJSON(w, sg)
 }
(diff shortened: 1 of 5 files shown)


doc/API.md
```diff
@@ -480,6 +485,92 @@ netaddress string // Optional
...
+Parameters: none
+
+Response:
+```javascript
+{
+  "folders": [
+    {
+      "path": "/home/foo/bar",
+      "capacity": 50000000000, // bytes
+      "capacityremaining": 100000, // bytes
```
DavidVorick: comments should be aligned as they would be from go fmt

DavidVorick:
forgot to run `make`

api/host.go
```diff
@@ -32,7 +32,7 @@ type (
 	// to /host/storage - a bunch of information about the status of storage
 	// management on the host.
 	StorageGET struct {
-		Folders []modules.StorageFolderMetadata `json:"folders"`
+		FolderMetadata []modules.StorageFolderMetadata `json:"foldermetadata"`
```
VoidingWarranties: I think it's advantageous to have json tags that are one word long. Consider the hostdb endpoints that return an object with a single "host" field.

doc/API.md
VoidingWarranties: I was thinking about this earlier. I kept them unaligned as the response is specified in JSON not go, and it doesn't seem to be a convention to align JSON fields in API docs. But aligned fields are easier to read so :+1:

api/host.go
DavidVorick: yeah but I feel like 'Folders' is insufficiently clear. You aren't returning the actual folders, only information about the folders. You can't use the folders you receive to see the contents or make any changes.

VoidingWarranties: FolderMetadata doesn't sound right to me. I think it's because the plurality of folder is ambiguous. It makes it sounds like it's a single folder's metadata.

lukechampine: what would it mean to return the "actual folders?"
I think "folders" is clear enough.

droghio: From an API usage perspective it seems clearer to have a call that reads something like `storage.folders` rather than `storage.StorageFolderMetadata` when querying the storage folders' paths/sizes. Just my 2¢.

DavidVorick: currently it would read storage.foldermetadata

PR #1313, opened 2016-06-28
Specify GitHub API version when checking for updates

VoidingWarranties:
>**Important**: The default version of the API may change in the future. If you're building an application and care about the stability of the API, be sure to request a specific version in the `Accept` header as shown in the examples below.

https://developer.github.com/v3/media/#request-specific-version

I haven't tested this change yet.


lukechampine:
I've manually confirmed that the response produced is identical to the response without the version header set. (Which is expected)

PR #1316, opened 2016-06-28
wrap /renter/contracts type

lukechampine:
This prevents the API from marshalling all of the contract Merkle roots, which can be a non-trivial amount of data (and is rarely useful).

### DIFF
--- a/api/renter.go
+++ b/api/renter.go
@@ -37,11 +37,11 @@ type (
 
 	// RenterContract represents a contract formed by the renter.
 	RenterContract struct {
-		FileContract types.FileContract         `json:"filecontract"`
-		ID           types.FileContractID       `json:"id"`
-		LastRevision types.FileContractRevision `json:"lastrevision"`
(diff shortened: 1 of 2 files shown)


api/renter.go
```diff
@@ -35,9 +35,18 @@ type (
 		FinancialMetrics modules.RenterFinancialMetrics `json:"financialmetrics"`
 	}
 
+	// RenterContract represents a contract formed by the renter.
+	RenterContract struct {
+		FileContract types.FileContract         `json:"filecontract"`
+		ID           types.FileContractID       `json:"id"`
+		LastRevision types.FileContractRevision `json:"lastrevision"`
+		NetAddress   modules.NetAddress         `json:"netaddress"`
+		Size         uint64                     `json:"size"`
```
lukechampine: open to feedback on which fields should be included. Specifically, I opted to precalculate `Size` instead of returning the number of Merkle roots. I think this is what 99% of API consumers care about.

PR #1317, opened 2016-06-28
fix (again) corruption condition in the hostdb

DavidVorick:
There was a problem in the hostdb previously where the host tree would
corrupt after a scan because the weight would change but the tree would
not restructure to account for the change in the weight.

That bug was supposed to be fixed by re-adding the node to the tree
after the change. But, it was an insufficient move due to the pointer
unsafety of the tree. Adding the node again is prefixed by a call to
remove the existing entry. The existing entry however has a different
weight than when it was inserted, causing corruption upon removal.

This change is not tested, tests to follow in future commits.

### DIFF
--- a/modules/renter/hostdb/scan.go
+++ b/modules/renter/hostdb/scan.go
@@ -78,6 +78,56 @@ func (hdb *HostDB) decrementReliability(addr modules.NetAddress, penalty types.C
 	}
 }
 
+// managedUpdateEntry updates an entry in the hostdb after a scan has taken
+// place.
+func (hdb *HostDB) managedUpdateEntry(hostEntry *hostEntry, settings modules.HostExternalSettings, err error) {
+	hdb.mu.Lock()
+	defer hdb.mu.Unlock()
+
+	// Regardless of whether the host responded, add it to allHosts.
+	priorHost, exists := hdb.allHosts[hostEntry.NetAddress]
+	if !exists {
+		hdb.allHosts[hostEntry.NetAddress] = hostEntry
+	}
+
+	// If the scan was unsuccessful, decrement the host's reliability.
+	if err != nil {
+		if exists && bytes.Equal(priorHost.PublicKey.Key, hostEntry.PublicKey.Key) {
+			// Only decrement the reliability if the public key in the
+			// hostdb matches the public key in the host announcement -
+			// the failure may just be a failed signature, indicating
+			// the wrong public key.
+			hdb.decrementReliability(hostEntry.NetAddress, UnreachablePenalty)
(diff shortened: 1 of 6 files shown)


DavidVorick:
@david60 can I ask you to code-review this? It should fix the hostdb problems you have reported, including an extra test to make sure things are happening correctly.

david60:
@DavidVorick sorry, I can only contribute sporadically. If you want to keep this open for a while I'll make a review when I can.

modules/renter/hostdb/scan.go
```diff
@@ -78,6 +78,56 @@ func (hdb *HostDB) decrementReliability(addr modules.NetAddress, penalty types.C
...
+// managedUpdateEntry updates an entry in the hostdb after a scan has taken
+// place.
+func (hdb *HostDB) managedUpdateEntry(hostEntry *hostEntry, settings modules.HostExternalSettings, err error) {
+	hdb.mu.Lock()
+	defer hdb.mu.Unlock()
+
+	// Regardless of whether the host responded, add it to allHosts.
+	priorHost, exists := hdb.allHosts[hostEntry.NetAddress]
+	if !exists {
+		hdb.allHosts[hostEntry.NetAddress] = hostEntry
```
david60: I think the conditional "hdb.allHosts[hostEntry.NetAddress] = hostEntry" assignment here may be causing a problem. (Not that this is related to the issue we're patching in this ticket - and this is how it pas previously, not changed by this patch).

I think this can leave the AllHosts out of date with respect to the ActiveHosts. The idea is that when AllHosts/ActiveHosts have been loaded from the persistent store, the hostEntry pointer in allHosts and the one in the node within activeHosts (for the same network address) point to separate (but similar) structures. The conditional assigment here means that subsequent updates in settings discovered during the scan may not be set in the hostEntry accessible via allHosts. Maybe the assignment should be done after the "if err != null {" bock, and done unconditionally (i.e. done even if allHosts already has the network address).

This may be involved in ticket #1320; if the renter uses an out of date payment or collateral rate setting when making a contract revision or renewal the host may reject it due to incorrect payment. I've seen some requests at my host that might be consistent with that. (The values sent in the revision or renewal corresponded to a price or collateral rate setting I had since changed).

I'll open a new ticket if I can write a test to show the problem..

david60:
Hi. Fix for problem where the weights in the host tree became inconsistent seems to be fixed. The new TestThreadedProbeHostsCorruption fails without the patch and passes with it. I did notice another potential problem nearly in managedUpdateEntry, see comment. (probably unrelated to the weights issue).

PR #1325, opened 2016-06-30
Require paths to be absolute when adding a storage folder

VoidingWarranties:
When writing the API docs for the storage endpoints I realized it might be confusing to add a relative path using the API, as the path would be relative to siad. This is especially confusing if you then restart siad in a different directory.

Another option is to convert the path to an absolute path, so that it's full path is shown in the list of storage folders.

### DIFF
--- a/modules/host/storagemanager/storagefolders.go
+++ b/modules/host/storagemanager/storagefolders.go
@@ -111,6 +111,9 @@ var (
 	// errStorageFolderNotFolder is returned if a storage folder gets added
 	// that is not a folder.
 	errStorageFolderNotFolder = errors.New("must use to an existing folder")
+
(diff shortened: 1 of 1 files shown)


modules/host/storagemanager/storagefolders.go
```diff
@@ -111,6 +111,9 @@ var (
 	// errStorageFolderNotFolder is returned if a storage folder gets added
 	// that is not a folder.
 	errStorageFolderNotFolder = errors.New("must use to an existing folder")
```
mnsl: `errStorageFolderNotFolder` message should be `"must use an existing folder"`

VoidingWarranties:
Typo fixed and rebased.

PR #1329, opened 2016-07-01
force api authentication if --disable-api-security

DavidVorick:
siac has no support for api authentication, and I would like to write
some additional tests around the logic. Would also like to convert the
error messages in to explicit variables the way that we have for most of
the modules.

### DIFF
--- a/modules/consensus/synchronize_test.go
+++ b/modules/consensus/synchronize_test.go
@@ -1196,7 +1196,7 @@ func TestIntegrationRelaySynchronize(t *testing.T) {
 
 	// Spin until the connection calls have completed.
 	for i := 0; i < 100; i++ {
-		time.Sleep(50 * time.Millisecond)
+		time.Sleep(150 * time.Millisecond)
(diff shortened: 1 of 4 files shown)


DavidVorick:
this should have more tests before being merged

DavidVorick:
This is necessary to protect miners. The miners in our ecosystem, even after upgrading to 1.0, are still largely using the `--disable-api-security` flag without setting up any sort of additional security, leaving themselves vulnerable to having their coins stolen. Then the miners blame us, and the thieves sell the coins, and the price drops, and the traders get mad too.

It's not really our fault, but I think we can at least do a little more to protect our miners.

VoidingWarranties:
siac does support API authentication

DavidVorick:
running `siac -h` does not display anything which indicates support for api authentication

lukechampine:
It could be clearer, yeah. IIRC the current behavior is that siac will try to make the call without auth, then fallback to a password prompt if it gets an "auth required" response.

DavidVorick:
oh hmm, that does allow you to protect the password with speakeasy

DavidVorick:
should be good now

PR #1332, opened 2016-07-02
require unlock hashes to include the checksum

### LINKED ISSUE #1232: Wallet should not accept sending addresses without a checksum

DavidVorick:
fix #1232

### DIFF
--- a/api/explorer.go
+++ b/api/explorer.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 
 	"github.com/NebulousLabs/Sia/build"
+	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/types"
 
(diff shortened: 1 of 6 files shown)


lukechampine:
looks good, I like that `scanAddress` is no longer overloaded.

DavidVorick:
I am concerned that there will be roadbumps in the explorer, before updating it we should run a few hashes through the explorer (or rather, write some tests that do the same) to make sure that they are working correctly.

Same with the host, the host was doing a `scanAddress` at some point and we should write a test to make sure that it can scan the hashes correctly.

I left it out of this PR because the whole API is pretty embarrassingly under-tested. @mnsl after you fix the stopgap stuff for the hostdb this is a good place to pick up. The API is a great place to test Sia as a whole because you're pretty much forced into exclusively integration tests. Plus, because of our preserve-compatibility requirements, the testing is not brittle, and therefore useful when making architectural changes to the modules.

PR #1333, opened 2016-07-03
Improve host logging, fix negative contract count error

DavidVorick:
Added some more logging to the host so we can get a good idea of what is happening behind the scenes as the number of contracts in the host increases and decreases.


modules/host/storageobligations.go
```diff
@@ -398,6 +401,7 @@ func (h *Host) addStorageObligation(so *storageObligation) error {
 	err3 := h.queueActionItem(so.expiration()+resubmissionTimeout, soid)
 	err = composeErrors(err0, err1, err2, err3)
 	if err != nil {
+		h.log.Println("Error with transaction set, redacting obligaation, id", so.id())
```
lukechampine: typo

PR #1335, opened 2016-07-04
store a status string in the storage obligations 

DavidVorick:
When a storage obligation is completed, a status string is added to the
obligation stored on disk. This allows hosts to view their contracts
historically and see why they failed in the event of a failure, which
will help with debugging/troubleshooting

I'm not entirely certain that a string is the best move. An alternate
option has us using error codes. A string is nice because it lets you
see the actual meaning of the error directly in the database. But if we
ever change the errors, the errors could be lost from the codebase.

I think a string is more likely to make sense in the long run, but it
does mean you could be left with a string that you are unable to grep.

### DIFF
--- a/modules/host/storageobligations.go
+++ b/modules/host/storageobligations.go
@@ -140,6 +140,7 @@ type storageObligation struct {
 	RevisionConfirmed bool
 	ProofConfirmed    bool
 	ObligationStatus  storageObligationStatus
+	StatusMessage     string
 }
 
 // getStorageObligation fetches a storage obligation from the database tx.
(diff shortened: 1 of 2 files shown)


DavidVorick:
this PR builds on top of the improved host logging PR.

I'm a little uncertain that this is the correct thing to do, but it will definitely help with debugging for failed storage contracts, once the host supports a way to serve all contracts over the api.

lukechampine:
Hmm. I'm not sure either. `StatusMessage` seems like it could be `Error`, left blank to indicate success. Or, you could set `StatusMessage` more thoroughly, e.g. with a message to indicate that the host is waiting to submit the storage proof. But I suppose you can already infer that from the height...

I guess I don't see why this is better than just parsing the log output, which should contain all of these errors already.

DavidVorick:
It makes it easier to present the obligation as a whole to the user.

DavidVorick:
closing this for now, as the storage obligation logic may be getting a facelift. For ACID correctness, it needs to be inside the `contractmanager`, which means splitting it out of the host. When moving it over, it's likely that I'll add a lot of safety to the code which currently isn't there.

PR #1338, opened 2016-07-05
Use contract and bandwidth prices to calculate host weight

mnsl:
`calculateHostWeight` now takes into account `hostEntry`'s `DownloadBandwidthPrice`, `UploadBandwidthPrice`, and `ContractPrice` fields; previously only `hostEntry.StoragePrice` was considered.

This PR also renames the `hostDB`'s `AveragePrice` method to `AverageContractPrice` for the sake of clarity (especially since contract and storage price have been [sometimes](https://github.com/NebulousLabs/Sia/pull/1317/commits/3e68f1bf27a0cf2c289ca995bc8b2231916bae28) been mixed up).

### DIFF
--- a/modules/renter/hostdb/hostweight.go
+++ b/modules/renter/hostdb/hostweight.go
@@ -8,7 +8,7 @@ import (
 
 var (
 	// Because most weights would otherwise be fractional, we set the base
-	// weight to 10^80 to give ourselves lots of precision when determing the
+	// weight to 10^150 to give ourselves lots of precision when determing the
 	// weight of a host
 	baseWeight = types.NewCurrency(new(big.Int).Exp(big.NewInt(10), big.NewInt(150), nil))
 )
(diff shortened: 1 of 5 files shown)


modules/renter/hostdb/hostweight.go
```diff
@@ -8,16 +8,22 @@ import (
...
 )
 
 // calculateHostWeight returns the weight of a host according to the settings of
 // the host database entry. Currently, only the price is considered.
 func calculateHostWeight(entry hostEntry) (weight types.Currency) {
+	// In the weighted price, the download bandwidth price is multiplied by
+	// numDownloads, which is our guess at the median number of downloads per
+	// file (some files may go untouched, but others may be requested
+	// frequently).
+	numDownloads := types.NewCurrency(big.NewInt(50))
```
lukechampine: `types.NewCurrency64`

mnsl: Ah, of course.  Thanks.

modules/renter/hostdb/hostweight.go
```diff
@@ -8,16 +8,22 @@ import (
...
 )
 
 // calculateHostWeight returns the weight of a host according to the settings of
 // the host database entry. Currently, only the price is considered.
 func calculateHostWeight(entry hostEntry) (weight types.Currency) {
+	// In the weighted price, the download bandwidth price is multiplied by
+	// numDownloads, which is our guess at the median number of downloads per
+	// file (some files may go untouched, but others may be requested
+	// many times).
+	const numDownloads = uint64(50)
```
lukechampine: `consts` never need explicit types.\* Not a big deal though.

\*I'm not 100% sure this is true, but the only counterexample I could think of was:
```go
type foo int

func (f foo) times2() foo {
    return f*2
}

func main() {
	const f foo = 3
	println(f.times2())
}
```
Even here, you could typecast at the call site (and I believe this incurs no overhead):
```go
func main() {
	const f = 3
	println(foo(f).times2())
}
```

DavidVorick:
`AverageContractPrice` => don't we want to be returning the average storage price?

mnsl:
Good question @DavidVorick -- also as far as I can tell, that method's not actually used anywhere

lukechampine:
In that case, I recommend we remove the method. For the purpose of allowance estimation, we may want to restore it later in some form or another, but this would likely be done in the contractor, not the hostdb.

PR #1345, opened 2016-07-07
A Ragtag Band of Changes

DavidVorick:
This is a work in progress. Originally this PR was aimed at creating a broad test that created multiple server testers and linked them all together. But getting the synchronization right proved to be problematic. So I started digging into ways to get things to syncrhonize, and found a pretty serious synchronizing issue with the host. Further down the rabbit hole I modified sync.ThreadGroup pretty heavily.

I'm making this PR early so that you guys can review the changes to the thread group and give your feedback. Pretty much nothing here is tested, other than that it gets used by the existing tests and they are all passing.

Highlight of the really important things and why I did them:

1. `OnStop` got broken into `BeforeStop` and `AfterStop`. `BeforeStop` functions will be called after the thread group is closed, but before `wg.Wait()` is called. `AfterStop` will be called after `wg.Wait()` is called. There's a pretty clear use case. Previously, `OnStop` had the same functionality as `BeforeStop`. This of course has issues, as if you are using the thread group to close the logger, there may be functions open (not yet finished because we haven't called `Wait`) which depend on the logger. And on the other end of the spectrum, there is the listener closing, which to function correctly definitely needs to happen before `wg.Wait`.

2. Got rid of IsStopped because it wasn't really in use, and because it interfered with some of the locking changes I made. The locking changes were necessary to get the concurrency model correct, though I haven't yet written the tests to demonstrate this. We can bring back IsStopped if it's really needed, but for the time being using the StopChan has been enough. Only the CPU miner was using it.

3. Added the `Pause` and `Resume` functions. These are going to be very useful during testing, and are going to enable us to break free of the rigid and restrictive concurrency pattern that the modules are currently forced to follow. The biggest reason that we enforced the rigid concurrency pattern is because it was very hard to test the modules when all of them were at different heights and you had no way of knowing when they would all be done processing the most recent block. Now you can use `Pause` and `Resume` to guarantee that all chain-connected events have completed within a module.

For example, in the host, storage obligations are submitted to the transaction pool asynchronously. But, a single call connects the storage obligation submission goroutine to the goroutine that spawned it, meaning that if you call `host.tg.Pause` after an action which should cause the obligation submission goroutine to be spawned, you can know that the goroutine will also be finished by the time Pause stops blocking.

I predict that this will be enormously helpful during testing.

While `Pause` is in effect, no thread or module will be able to get access to a `tg.Add` call, it will block until the pause has been resumed.

Getting pause working required the addition of two more functions, `AddPermanent` and `DonePermanent`. This is because threads such as the listener threads are only spawned once, and interfere with `Pause` because they will not release their waitgroup lock until `Stop` is called and the thread is killed. We don't want to kill them during `Pause`, so we actually put them in a separate waitgroup. There are probably alternate ways to achieve the same goal.

That puts the current ThreadGroup implementation at 10 exported functions, which is definitely on the high side. That said, I feel pretty strongly about `Pause` and `Resume`, they are going to be very helpful for testing asynchronous actions.

---

Getting my simulated-full-network test working is going to require adding proper pausing to the consensus set, which was part of the inspriation for making this. But I also needed it to properly test the host.

---

For the rest of this PR, the host still needs some lookovers, especially with regards to the storage obligation handling and the action item handling. I want to get the ecosystem test that I wrote to be passing, and I definitely need to add proper testing to the ThreadGroup changes and to the TryLock that I created.

Following this PR I will be looking further into either the safety of the tbuilder or the dependency sending from the renter for the renter-host protocol. The renter is very likely not sending all dependent transactions, and is also very likely to need an extension to the transaction pool to be able to make the sends.


modules/miner/testminer.go
```diff
@@ -89,22 +113,5 @@ func (m *Miner) FindBlock() (types.Block, error) {
...
-		if bytes.Compare(target[:], id[:]) >= 0 {
-			copy(b.Nonce[:], header[32:40])
-			return b, true
-		}
-		*(*uint64)(unsafe.Pointer(&header[32])) = nonce
-		nonce++
-	}
-	return b, false
+	return solveBlock(b, target)
 }
```
mnsl: Why is this change necessary?  Will `SolveBlock` eventually do more than just wrap `solveBlock`?

sync/threadgroup.go
```diff
@@ -53,46 +46,120 @@ func (tg *ThreadGroup) IsStopped() bool {
...
-		fn()
-	} else {
-		tg.stopFns = append(tg.stopFns, fn)
-		tg.mu.Unlock()
-	}
+// StopChan provides read-only access to the ThreadGroup's stopChan. Callers
+// should select on StopChan in order to interrupt long-running reads (such as
+// time.After).
+func (tg *ThreadGroup) StopChan() <-chan struct{} {
+	// Initialize tg.stopChan if it is nil; this makes an unitialized
```
mnsl: typo:  uninitialized

sync/threadgroup_test.go
```diff
@@ -37,23 +37,13 @@ func TestThreadGroup(t *testing.T) {
 // TestThreadGroupStop tests the behavior of a ThreadGroup after Stop has been
 // called.
 func TestThreadGroupStop(t *testing.T) {
+	// Create a thread group and stop it.
 	var tg ThreadGroup
-
-	// IsStopped should return false
-	if tg.IsStopped() {
```
mnsl: Why not test `isStopped` here?

sync/threadgroup_test.go
```diff
@@ -37,23 +37,13 @@ func TestThreadGroup(t *testing.T) {
...
 	err := tg.Stop()
 	if err != nil {
 		t.Fatal(err)
 	}
 
-	// IsStopped should return true
-	if !tg.IsStopped() {
-		t.Error("IsStopped returns false on stopped ThreadGroup")
-	}
-
```
mnsl: ditto

sync/threadgroup_test.go
```diff
@@ -104,12 +94,6 @@ func TestThreadGroupOnce(t *testing.T) {
 	}
 
 	tg = new(ThreadGroup)
-	tg.IsStopped()
```
mnsl: ditto, why not test `isStopped` here?

sync/trylock.go
```diff
@@ -0,0 +1,62 @@
...
+
+// Lock grabs a lock on the TryMutex, blocking until the lock is obtained.
+func (tm *TryMutex) Lock() {
+	tm.outer.Lock()
+	defer tm.outer.Unlock()
+
+	// The inner lock needs to be grabbed before 'tm.locked' is set to true.
+	tm.inner.Lock()
+	if tm.locked {
+		panic("bug in TryLock")
```
lukechampine: name is incorrect. `"'locked' flag should be false after obtaining lock"` might be more helpful.

sync/trylock.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+	tm.locked = true
+}
+
+// TryLock grabs a lock on the TryMutex, returning an error if the mutex is
+// already locked.
+func (tm *TryMutex) TryLock() error {
```
lukechampine: a bool seems like the more appropriate return value here, although it means you need to create the error yourself:
```go
if !mu.TryLock() {
    return errObligationLocked
}
```

sync/threadgroup.go
```diff
@@ -53,46 +46,125 @@ func (tg *ThreadGroup) IsStopped() bool {
...
+	tg.mu.Lock()
+	defer tg.mu.Unlock()
+
+	if tg.isStopped() {
+		fn()
+		return
+	}
+	tg.beforeStopFns = append(tg.beforeStopFns, fn)
+}
+
```
mnsl: `BeforeWait`, `AfterWait`, `beforeWaitFns`, and `afterWaitFns` might be better names, since the `stopChan` is closed before the `beforeStopFns` are called.

modules/gateway/gateway_test.go
```diff
@@ -69,6 +69,7 @@ func TestAddress(t *testing.T) {
 	}
 }
 
+// TestPeers checks that two gateways are able to connect to eachother.
```
mnsl: eachother -> each other

modules/host/negotiaterevisecontract.go
```diff
@@ -288,14 +288,14 @@ func (h *Host) managedRPCReviseContract(conn net.Conn) error {
...
 	if err != nil {
 		return err
 	}
 	defer func() {
 		h.mu.Lock()
-		err = h.unlockStorageObligation(so)
+		h.unlockStorageObligation(so.id())
 		h.mu.Unlock()
 		if err != nil {
 			h.log.Critical(err)
```
mnsl: checking the wrong `err` here, since `h.unlockStorageObligation(so.id())` doesn't return an error anymore

modules/host/storageobligations_smoke_test.go
```diff
@@ -199,18 +193,15 @@ func TestSingleSectorStorageObligationStack(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	err = ht.host.lockStorageObligation(so)
+	ht.host.lockStorageObligation(so.id())
 	if err != nil {
 		t.Fatal(err)
```
mnsl: checking the wrong `err` here, since `h.unlockStorageObligation(so.id())` doesn't return an error anymore

modules/host/storageobligations_smoke_test.go
```diff
@@ -494,18 +485,15 @@ func TestMultiSectorStorageObligationStack(t *testing.T) {
 			NewUnlockHash:         types.UnlockConditions{}.UnlockHash(),
 		}},
 	}}
-	err = ht.host.lockStorageObligation(so)
+	ht.host.lockStorageObligation(so.id())
```
mnsl: checking the wrong `err` here, since `h.lockStorageObligation(so.id())` doesn't return an error anymore

lukechampine:
What is the difference between `AfterStop` and simply calling the function after `tg.Stop`?

In general, I dislike the idea of creating highly sophisticated, bespoke concurrency constructs. The ThreadGroup is now much more complex. Although it is more general, in practice it is harder to use and harder to verify.
(`TryMutex` seems ok, since it's a relatively simple extension of `Mutex` and its semantics are clear from its name.)

I haven't quite wrapped my head around `Pause` and `Resume`, but my initial feeling is that they don't map 1-1 to the behavior you're actually trying to achieve. As I understand it, you are using `Pause` to block until a certain action has completed. But `Pause` only gives you the ability to block at points where `tg.Add` is called. (It also blocks calls to `tg.Stop`, `tg.Add`, `tg.Done`, etc. which may surprise the user and cause deadlocks, although this seems unlikely.) I feel like it would be preferable to have an explicit signal inserted wherever you want to observe behavior, in the form of a channel or a function call. This requires adding fields to the object being tested, which sucks. But at least it's clear what you're trying to do, and you don't have to rely on calls to `tg.Add` aligning with what you're trying to test. (Again, I'm not 100% sure what your intended use case is, so this critique could be invalid.)
Another thing that stuck out to me was that `Pause` and `Resume` were always called in direct sequence -- could they be combined into a single function?
Perhaps I'll come around to the idea if you provide a more concrete example of how they would be used.

---

Recently I've been wondering if this sort of fine-grained, introspective testing is the best way to go. To me, the upload/download problems we've experienced represent a failure of our current testing methodology -- and we put a *lot* of effort into writing these tests! What if instead, we had spent our time setting up a proper testnet and doing manual smoke tests? That is, a checklist like:

- Create a wallet
- Mine coins into the wallet
- Set an allowance
- Upload a file
- Download the file
- Wait for file to be renewed
- Download the file again
- Delete the file
- Set a different allowance
- ...

It's impossible to know for sure, but my guess is that we would have been more in touch with the bugs that directly affect users. How long do you think the `MerkleRoots` renewal bug would have survived if we were doing these sorts of tests? And it's not just bugs, but the user experience as a whole, which automated testing can't really capture. We'd have a better sense of how fast uploads/downloads are, where the pain-points are, etc. And constant usage of the software would strongly motivate us to fix those annoyances (i.e. dogfooding). Another benefit is that we could escape this miasma of setup code, teardown code, instrumentation/introspection tools, *mocking*... we've sunk so many hours into writing code that ultimately did not impact users at all.
The downside is that more subtle bugs may have gone undetected for longer; also, fewer tests means it's harder to pinpoint the root cause of the bug, which could prolong debugging. Manual testing is also prone to human error and laziness, so we would probably do it much less frequently. I think it's obvious that rigorous unit testing is absolutely important for some code (e.g. consensus). And there's no reason why you can't do both. But I think that for non-critical code, we should consider manual integration testing in lieu of writing a bunch of brittle unit tests that will never fail anyway.

sync/threadgroup_test.go
DavidVorick: Oh duh

DavidVorick: Will add it back

modules/miner/testminer.go
DavidVorick: SolveBlock is an exported function of the miner, which means by convention you assume it grabs a lock. SolveBlock needs to be called by another function while under a lock, which by convention is illegal.

Eliminating exceptions makes the code less brittle and less error prone

sync/threadgroup.go
DavidVorick: Was thinking about that. It requires knowing the internals, but you really need to know those anyway to use the thread group correctly.

Will probably change the names to your suggestions

DavidVorick:
Jordan is in the middle of setting up a more sophisoticated way for us to do broad production testing. Manual testing is definitely not the way to go, other than  dogfooding, but we don't need all the automatic tests to be in go either, we could have a bash script for example, or since go makes it so easy to use the api, a 'siat' that will interactively walk you through a huge set of user actions, letting you inspect stuff halfway through if you wish.

Some of this stuff is basically impossible to smoke test though. Once we add more reputation, for example. We'll need to continue evolving our test strategies, but it needs to be systemic and not ad-hoc.

My reaction right now is to set up a giant in-testing network (ecosystem_test.go) that runs a bunch of nodes and synchronizes them, and simulate realistic situations like one host not having transactions or being a few blocks behind. Setting up that test though requires more control over the node synchronization, because you don't want to check for a result until the processing is complete.

Adding a function to `AfterStop` gives you the luxury of opening and closing a resource at the same time, something we've already discussed as a big benefit to safe coding.

Pause  makes sense because it's a very simple insertion. If you are doing clean shutdown correctly, pausing does not require any changes to your threading strategy. You call pause, and then all open processes will complete and new ones will block, same as if you call stop. And now you've got a completed, consistent view into the module state. If it's not complete or freezes, that means you have errors in your clean shutdown code.

You can basically think of pause like an early clean shutdown command, with a resume action attached to it.

DavidVorick:
Another good test might be to spin up a bunch of tester objects, have them all mine and rent and host with no synchronization controls, set the difficulty to something sane (so all nodes are mining with similar power and nobody is getting all the blocks), then letting it get to a certain height and seeing if it succeeds.

A lot of the dirtiness though just can't be replicated well in testing, at least not without better tooling. For example, problems have been caused by hosts disconnecting right before sending a signature, but after propagating the transaction. Problems can be caused by sending money transactions at the same time as making file contracts. Our sterile test environments don't have things like that. Even manual testing is unlikely to accurately replicate a user trying to do a bunch of things at once.

To me that just means we don't have sufficient concurrency safety in our project. We will likely to continue to miss important bugs while users are doing things we didn't think about. I think the best way forward is now emphasis on the testing of the safely features and failure modes.  I don't think we would have seen all these issues on a smaller network with experienced users.

DavidVorick:
Ah hit the wrong button

api/server_helpers_test.go
```diff
@@ -260,6 +260,28 @@ func assembleExplorerServerTester(testdir string) (*serverTester, error) {
...
 }
 
+// blankServerTester creates a server tester object that is ready for testing,
+// without any money in the wallet.
+func blankServerTester(name string) (*serverTester, error) {
+	// createServerTester is expensive, and therefore should not be called
+	// during short tests.
+	if testing.Short() {
+		panic("createServerTester called during short tests")
+	}
```
mnsl: `createServerTester` -> `blankServerTester`

sync/threadgroup_test.go
DavidVorick: Adding this code back turned up an error.

Good catch :+1:

api/ecosystem_test.go
```diff
@@ -0,0 +1,238 @@
...
+	chainTip, err := synchronizationCheck(allTesters)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Mine a block from each node, to give the node money in the wallet that
+	// is recognized by the shared chain.
+	for i := range allTesters {
+		// Wait until the current tester has 'leadBlockID' as its current
+		// block, to make sure the network is buildling a community chain
```
mnsl: buildling -> building

api/ecosystem_test.go
```diff
@@ -0,0 +1,238 @@
...
+	err = stRenter1.stdPostAPI("/gateway/connect/"+string(ggSTR2.NetAddress), nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	err = stRenter2.stdPostAPI("/gateway/connect/"+string(ggSTL.NetAddress), nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Connectivity check - all nodes should by synchronized to the leader's
```
mnsl: by -> be

api/ecosystem_test.go
```diff
@@ -0,0 +1,238 @@
...
+	allTesters := []*serverTester{stLeader, stHost1, stHost2, stHost3, stHost4, stRenter1, stRenter2}
+	chainTip, err := synchronizationCheck(allTesters)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Mine a block from each node, to give the node money in the wallet that
+	// is recognized by the shared chain.
+	for i := range allTesters {
+		// Wait until the current tester has 'leadBlockID' as its current
```
mnsl: probably clearer to just call it `chainTip` instead of 'leadBlockID' since the latter isn't used anywhere else

DavidVorick:
Ok, this guy should be ~done.

ecosystem_test.go is still fully disabled, and the way the host does storage obligation locking is still incomplete. But those shall be fixed in another PR.

This code should be good enough to RC. There's significantly expanded testing, at least one critical error removed from the host, a greater tolerance for block + txn propagation mismatches (which should reduce the number of consensus conflicts), and I think at least one deadlock got eliminated as well.

Looking forward to continuing the expansion of testing.

david60:
@DavidVorick Hi. Nice catch, I noticed the fix the race between the reading and writing of the storageobligation between ProcessConsensusChange & threadedHandleActionItem and an ongoing modification of the obligation. :)

I think there may still an issue with the updates of the OriginConfirmed (and possibly RevisionConfirmed & ProofConfirmed) within ProcessConsensusChange: If there is a modification of the this contract ongoing the change of OriginConfirmed might be lost when the revision is written.

lukechampine:
I'm seeing nondeterministic timeouts. I believe they are in TestSingleSectorStorageObligationStack.

DavidVorick:
There's a deadlock somewhere. I also realized that the `ThreadGroup` implementation is not safe, so I'm going to have to modify that somewhat significantly. For our purposes, it's not enough to be using a `sync.WaitGroup`, we'll have to do the incrementing and decrementing manually. This is because we need to support daisy-chaining.

@david60 the locking around the storageobligation is unfortunately incomplete. A lock should be grabbed as soon as the `recentrevision` call happens (after the renter passes the cryptographic challenge), and it should not be released until full negotiation is done. I believe that will solve the remaining issues.

I was going to do that in another pull request, but it might make sense to do it here, though I feel bad about having so much code packed into a single PR.

DavidVorick:
```
tg.Add()
go func() {
    tg.Flush()
}
go func() {
    time.Sleep(time.Second)
    tg.Add()
    go func() {
        time.Sleep(time.Second)
        tg.Done()
    }
    tg.Done()
}
```
First add is called. It's not going to return until a different gothread has run. That gothread calls another gothread protected by an add and a done. Flush is called in the middle of all this. Flush should not get stuck despite waiting for all of the calls to Add and Done to finish. Not all of the Add functions will have been called by the time Flush is called.

DavidVorick:
Alternatively, this chaining could be tracked manually by whatever, using a waitgroup that is external to the thread group.

If you support daisy chaining, you lose the guarantee that flush will finish, because external functions can continue calling 'Add'. There'd need to be a requirement that 'Add' will never be called inside of another call to 'Add'. Maybe that is preferable, but it will make certain parts of the host more complex.

DavidVorick:
hmm, that's probably preferable. I'll discard the support for daisy chaining.

david60:
@DavidVorick the update of, e.g. OriginConfirmed, for storageobligations in ProcessConsensusChange() isn't protected by requiring the lock on the storageobligation, is it?  (If that's the case, then maybe a new ticket/PR is better, because it's not something introduced in this PR)

modules/consensus/consensusset.go
```diff
@@ -211,6 +211,9 @@ func (cs *ConsensusSet) Close() error {
 
 // CurrentBlock returns the latest block in the heaviest known blockchain.
 func (cs *ConsensusSet) CurrentBlock() (block types.Block) {
+	cs.mu.RLock()
+	defer cs.mu.RUnlock()
```
lukechampine: do you need this on the other exported functions as well? I assume it's protecting `cs.db`

modules/host/storageobligations.go
```diff
@@ -732,7 +757,9 @@ func (h *Host) handleActionItem(so *storageObligation) {
 		sectorIndex := segmentIndex / (modules.SectorSize / crypto.SegmentSize)
 		// Pull the corresponding sector into memory.
 		sectorRoot := so.SectorRoots[sectorIndex]
+		h.mu.Lock()
 		sectorBytes, err := h.ReadSector(sectorRoot)
+		h.mu.Unlock()
```
lukechampine: why does this require a write lock?

modules/host/update.go
```diff
@@ -97,23 +101,36 @@ func (h *Host) initConsensusSubscription() error {
...
-	h.mu.Lock()
-	defer h.mu.Unlock()
 	err := h.tg.Add()
 	if err != nil {
 		return
 	}
-	defer h.tg.Done()
+	wg := new(sync.WaitGroup)
+	defer func() {
+		wg.Wait()
```
lukechampine: this is far away from where the waitgroup is actually used. I would move it down to the `for` loop and call `wg.Wait` directly instead of in the `defer`.

sync/threadgroup.go
```diff
@@ -9,90 +9,142 @@ import (
...
+// is no longer useful.
 //
-// During shutdown, it is common to close resources such as net.Listeners.
-// Typically, this would require spawning a goroutine to wait on the
-// ThreadGroup's StopChan and then close the resource. To make this more
-// convenient, ThreadGroup provides an OnStop method. Functions passed to
-// OnStop will be called automatically when Stop is called.
+// Nested calls to Add() are a deadlock risk. It is safe to call Add() and
+// Stop() and Flush() concurrently so long as there are no nested calls to
+// Add().
```
lukechampine: need to clarify what a "nested call to Add" is.

sync/threadgroup.go
```diff
@@ -9,90 +9,142 @@ import (
...
 		return ErrStopped
 	}
 	tg.wg.Add(1)
 	return nil
 }
 
-// Done decrements the ThreadGroup counter.
+// AfterStop ensures that a function will be called after Stop() has been
+// called and after all running routines have called Done(). The functions will
+// be called in reverse order to how they were added, simiar to defer. If
```
lukechampine: typo

modules/host/update.go
DavidVorick: The `wg.Wait()` call needs to happen after `h.mu.Unlock()`. Because defers are called in reverse, if we are going to `defer wg.Wait()`, then that defer statement needs to come before `defer h.mu.Unlock()`, so we'd need to do manual unlocking in the rest of the function.

I will add more comments to explain this, since it is far away from the other code it is coupled to.

modules/consensus/consensusset.go
DavidVorick: That ends up being difficult to answer. It is protecting the `cs.db`, but if we add locking to the other exported functions, we'll get deadlocks. Now that we have a `ThreadGroup`, a lot of this stuff can be repaired more easily, but going in to depth on everything is going to be a pull request of its own.

=/

modules/host/storageobligations.go
DavidVorick: I think I was just being over-eager with the locking. I don't think we even need a readlock here.

sync/threadgroup.go
DavidVorick: I don't see it

mnsl: line 60, simiar -> similar

modules/host/update.go
lukechampine: seems like you could avoid that by moving the `Unlock` to after the `Update`.

sync/trymutex.go
```diff
@@ -0,0 +1,63 @@
+package sync
+
+import (
+	"sync"
+	"time"
+)
+
+// TryMutex provides a mutex that allows you to attempt to grab a mutex, and
+// then fail if the mutex is either not grabbed immediately or is
```
lukechampine: incomplete comment

Issue #1346, opened 2016-07-07
renter failed when needing to use the cached revision during upload


david60:
Hi,

I saw my renter and a host go through this sequence:

While uploading, the renter sends data and revision number X to host. Host accepts; but sends WriteNegotiationStop, and then the transactions signatures. (Stop because I think it had reached its time limit for revision iterations).

Renter treats the stop as an error and considers this revision failed. So at this point it has revision X-1 in the contact, and revision X in the cachedRevisions. It attempts to connect to the host again, using revision X-1, but fails with revision mismatch. It then uses the cached revision, and connects and sends revision X+1. The host rejects this because the file merkle root is bad. The host's storageobligation is still at revision X. But the renter[correction] has revision X-1 persisted in the contract and now only has revision X+1 in the cached revisions. So it is not able to successfully connect to the host again to revise the contract any more.

It seems a problem may be that when the cached revision is applied, the contract's MerkleRoots list is then inconsistent with the revision that it got from the revision cache.

A separate issue might be the handling of the "stop" from host to renter during the revision iteration.

lukechampine:
Good catch! It looks like you are correct, the renter calls `modules.ReadNegotiationAcceptance` but does not have a special case for `modules.ErrStopResponse`. The proper behavior would be to complete the revision, but close the connection. I will implement a fix asap.

PR #1347, opened 2016-07-07
Handle StopResponse correctly in proto

lukechampine:
As reported in #1346, the renter protocol was not properly handling `StopResponse` during revision. The correct behavior is to apply the revision and then gracefully terminate. Instead, the renter treated `StopResponse` as an error and exited early without applying the revision. This caused the renter and host to desync.

Note that in this implementation, the renter will not know that the host has closed the connection until it tries to perform another revision operation. An alternative approach would bubble the `ErrStopResponse` error all the way up to `Upload`, `Delete`, etc., and the caller would have to check for this error specifically and handle it correctly -- namely, they would have to recognize that the other return values were still valid, despite `err != nil`. I opted not to take this approach, because there is little benefit and the renter needs to be safeguarded against random host disconnects anyway.

On that note, this PR should also fix a subtle bug affecting cached revisions. Previously, `SaveFn` was called before doing any network I/O. This meant that if the host had silently disconnected, the renter would cache the revision before finding out. By itself, this shouldn't cause the renter to desync (since it will first try the current revision, not the cached revision), but it may have compounded other issues.

### DIFF
--- a/modules/renter/proto/downloader.go
+++ b/modules/renter/proto/downloader.go
@@ -40,14 +40,6 @@ func (hd *Downloader) Sector(root crypto.Hash) (modules.RenterContract, []byte, 
 	// create the download revision
 	rev := newDownloadRevision(hd.contract.LastRevision, sectorPrice)
 
-	// if a SaveFn was provided, call it before doing any further I/O, because
-	// that's when we're most likely to experience failure
-	if hd.SaveFn != nil {
-		if err := hd.SaveFn(rev); err != nil {
-			return modules.RenterContract{}, nil, errors.New("failed to save unsigned revision: " + err.Error())
-		}
-	}
-
 	// initiate download by confirming host settings
 	if err := startDownload(hd.conn, hd.host); err != nil {
 		return modules.RenterContract{}, nil, err
@@ -64,13 +56,11 @@ func (hd *Downloader) Sector(root crypto.Hash) (modules.RenterContract, []byte, 
 	}
 
 	// send the revision to the host for approval
-	signedTxn, err := negotiateRevision(hd.conn, rev, hd.contract.SecretKey)
+	signedTxn, err := negotiateRevision(hd.conn, rev, hd.contract.SecretKey, hd.SaveFn)
 	if err == modules.ErrStopResponse {
 		// if host gracefully closed, close our connection as well; this will
-		// cause the next operation will fail. We also need to set SaveFn to
-		// nil to prevent saving a revision that the host can't see.
-		he.conn.Close()
-		he.SaveFn = nil
+		// cause the next operation will fail
+		hd.conn.Close()
 	} else if err != nil {
 		return modules.RenterContract{}, nil, err
 	}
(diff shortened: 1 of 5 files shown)


lukechampine:
@david60 also pointed out that since cached revisions are not accompanied by their Merkle roots, desync can still occur, since the renter will be using the old revision but the new Merkle roots. I will address this in a separate PR.

modules/renter/proto/downloader.go
```diff
@@ -65,7 +65,13 @@ func (hd *Downloader) Sector(root crypto.Hash) (modules.RenterContract, []byte,
 
 	// send the revision to the host for approval
 	signedTxn, err := negotiateRevision(hd.conn, rev, hd.contract.SecretKey)
-	if err != nil {
+	if err == modules.ErrStopResponse {
+		// if host gracefully closed, close our connection as well; this will
+		// cause the next operation will fail. We also need to set SaveFn to
```
mnsl: 'will fail' -> 'to fail'

modules/renter/proto/editor.go
```diff
@@ -82,7 +82,13 @@ func (he *Editor) runRevisionIteration(actions []modules.RevisionAction, rev typ
 
 	// send revision to host and exchange signatures
 	signedTxn, err := negotiateRevision(he.conn, rev, he.contract.SecretKey)
-	if err != nil {
+	if err == modules.ErrStopResponse {
+		// if host gracefully closed, close our connection as well; this will
+		// cause the next operation will fail. We also need to set SaveFn to
```
mnsl: ditto

modules/renter/proto/downloader.go
```diff
@@ -65,7 +65,13 @@ func (hd *Downloader) Sector(root crypto.Hash) (modules.RenterContract, []byte,
...
+		// nil to prevent saving a revision that the host can't see.
+		he.conn.Close()
+		he.SaveFn = nil
```
mnsl: Why not return here?

lukechampine: good catch. Since the download transfer happens after the exchange, you don't want to close until you've received the data. (For other revisions, the action is sent along with the revision, so this doesn't apply.)

The host download code, however, does not currently support sending `StopResponse` after timing out. David should tackle this.

DavidVorick:
Can you write some tests (preferably in the api package) that demonstrate the failure before the change, and demonstrate the correction after the change?

Need to get into the habit of wrapping every piece of functionality in a test.

lukechampine:
There isn't an easy way to simulate a `StopResponse`. The host sends it only after `iteratedConnectionTime` has elapsed, so a standard integration test is out, unless we alter `iteratedConnectionTime` during testing -- but even then, we're only testing the behavior indirectly, and the test would take a while. Alternatively, I could simulate the entire host-side of the protocol, but that's not trivial either (a `host/proto` package might make it easier).

The most straightforward test would be to test the `negotiateRevision` function, simulating the host's side of the connection. I think that's the approach I will use.

PR #1349, opened 2016-07-08
API testing

mnsl:
* Fixes typos & a faulty error check in `TestRenterConflicts`
* Implements `TestRenterHostsAllHandler` to test `renterHostsAllHandler`, which handles the call to `hostdb/all`.
* Introduces type `AllHosts` to make testing `renterHostsAllHandler` more intuitive (see commit 9710357 for the code without the `AllHosts` type)

Upcoming: test that `renterHostsActiveHandler` and `renterHostsAllHandler` work correctly for a network with multiple hosts.

### DIFF
--- a/api/renter.go
+++ b/api/renter.go
@@ -73,6 +73,11 @@ type (
 	ActiveHosts struct {
 		Hosts []modules.HostDBEntry `json:"hosts"`
 	}
+
+	// AllHosts lists all hosts that the renter is aware of.
+	AllHosts struct {
+		Hosts []modules.HostDBEntry `json:"hosts"`
+	}
 )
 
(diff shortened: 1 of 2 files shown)


api/renter.go
```diff
@@ -73,6 +73,11 @@ type (
 	ActiveHosts struct {
 		Hosts []modules.HostDBEntry `json:"hosts"`
 	}
+
+	// AllHosts lists all hosts that the renter is aware of.
+	AllHosts struct {
+		Hosts []modules.HostDBEntry `json:"allhosts"`
```
lukechampine: json names should always match field names. I think `hosts` is fine here.

PR #1353, opened 2016-07-12
Save merkle roots

lukechampine:
One alternative that just occurred to me is defining `cachedRevision` in the proto package. Then `SaveFn` could take a `cachedRevision` as an argument instead of two separate args. Not sure though.

### DIFF
--- a/modules/renter/contractor/contractor.go
+++ b/modules/renter/contractor/contractor.go
@@ -6,6 +6,7 @@ import (
 	"path/filepath"
 	"sync"
 
+	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/persist"
 	"github.com/NebulousLabs/Sia/types"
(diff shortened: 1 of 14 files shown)


modules/renter/contractor/contractor.go
```diff
@@ -17,6 +17,11 @@ var (
 	errNilTpool  = errors.New("cannot create contractor with nil transaction pool")
 )
 
+type cachedRevision struct {
+	revision    types.FileContractRevision
+	merkleRoots []crypto.Hash
```
DavidVorick: for now this is fine, but we should note that merkleRoots may end up being very large, dozens or hundreds of MB depending on how many sectors have been uploaded. Long term, this should probably be a diff over the existing saved merkleRoots.

DavidVorick:
`make` failed

modules/renter/contractor/contractor.go
lukechampine: 100MB of Merkle roots would mean a 13 TB file. I don't think we need to worry about that for a while.
It is true, though, that caching a revision for every file effectively doubles the data that the renter stores locally. So a diff is reasonable. I'll leave a TODO.

PR #1355, opened 2016-07-13
use most recent contract when downloading

lukechampine:
This should alleviate some download problems.

### DIFF
--- a/modules/renter/contractor/contractor.go
+++ b/modules/renter/contractor/contractor.go
@@ -64,6 +64,18 @@ func (c *Contractor) FinancialMetrics() modules.RenterFinancialMetrics {
 	return c.financialMetrics
 }
 
+// Contract returns the latest contract formed with the specified host.
+func (c *Contractor) Contract(hostAddr modules.NetAddress) (modules.RenterContract, bool) {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+	for _, c := range c.contracts {
+		if c.NetAddress == hostAddr {
+			return c, true
+		}
+	}
+	return modules.RenterContract{}, false
+}
+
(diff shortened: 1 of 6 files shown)


DavidVorick:
Is there a lock of some sort to make sure that you can't have a revision and a renewal open at the same time?

DavidVorick:
Also, this code needs a test

lukechampine:
That is a separate issue. Currently we just rely on the host lock. Renewals will be rejected until editing stops.

lukechampine:
It's difficult to test this in any meaningful way because of how the download code is constructed. I could write a test for `Contracts` pretty easily but I doubt it would be useful.

DavidVorick:
Having a test for Contracts is better than having a component of the code which is untested.

I don't think that there will be any problems with the download code using the wrong contract, unless there's somehow going to be multiple processes looking at the same contract concurrently. The payment is still being transferred, so  the host is happy. And the host will recognize the sector root even if the sector root is not in the old file contract, because the host download code is willing to serve any sector that it has on disk (which is why all renters need to encrypt their data - the host will let anyone look at it).

Which opens up a good question. What do we do if a user wants to both upload and download from a contract at the same time? Per the contract locking, a user should only be trying to do one at a time.

Eventually it's going to be bad for the renter if the renter is relying on the host locking, because the renter is going to reduce the reputation of a host every time that the host denies a request. If the renter is getting denied because of its own locking issues though, the renter is going to have a set of host reputations that don't line up well with reality.

DavidVorick:
modules/renter/contractor/contractor_test.go:118: possible formatting directive in Error call

modules/renter/contractor/contractor_test.go:120: possible formatting directive in Error cal

PR #1356, opened 2016-07-13
better storage obligation safety

DavidVorick:
The host now locks storage obligations before ever pulling them from the
database, and keeps the locks until the host is no longer going to be
adding them to the database. This resolves some issues relating to
storage obligation corruption, which can cause errors like 'Consensus
Conflict' and can cause the host to report the wrong revision number, or
to have the wrong sector roots, etc.

### DIFF
--- a/modules/host/update.go
+++ b/modules/host/update.go
@@ -113,32 +113,14 @@ func (h *Host) initConsensusSubscription() error {
 // ProcessConsensusChange will be called by the consensus set every time there
 // is a change to the blockchain.
 func (h *Host) ProcessConsensusChange(cc modules.ConsensusChange) {
-	// ProcessConsensusChange spawns multiple threads which will wait until
-	// storage obligations are unlocked and then submit storage proofs to
-	// blockchain for those obligations. The ThreadGroup needs to wait for
-	// these threads to terminate. These threads also require a host lock,
-	//
-	// We use a local sync.WaitGroup to track the progress of these threads.
-	// Because these threads require a host lock, wg.Wait() should not be
-	// called until after the host lock has been released. And then tg.Done()
-	// should not be called until wg.Wait() returns.
-	//
-	// Because the host lock is released in a defer statement, wg.Wait() must
-	// be called in a defer statement that comes before the deferred
-	// h.mu.Unlock(). This is why the code for finishing the wait group is
-	// declared far above the place where threads are added to the wait group.
+	// Add is called at the beginning of the function, but Done cannot be
+	// called until all of the threads spawned by this function have also
+	// terminated. This function should not block while these threads wait to
+	// terminate.
 	err := h.tg.Add()
 	if err != nil {
 		return
 	}
-	wg := new(sync.WaitGroup)
-	defer func() {
-		wg.Wait()
-		h.tg.Done()
-	}()
-
-	// Host needs to unlock before wg.Wait() is called, so that the other
-	// threads may access the host lock.
 	lockID := h.mu.Lock()
 	defer h.mu.Unlock(lockID)
 
@@ -313,7 +295,25 @@ func (h *Host) ProcessConsensusChange(cc modules.ConsensusChange) {
 		h.log.Println(err)
 	}
 
-	// Handle the list of action items.
+	// Handle the list of action items. Action items require host locks and
+	// potentially require waiting for long periods of time while various
+	// network communications finish. To prevent the host lock from being held
+	// while the network I/O is blocking the action item execution, the action
+	// items are handled in their own thread. A wait group is used to prevent
+	// the host from releasing ThreadGroup control until all threads that this
+	// function have spawned have completed.
+	wg := new(sync.WaitGroup)
+	// The host should not release thread group control until all spawned
+	// threads have completed execution, though waiting for them should not
+	// block this function. The host should also not release thread group
+	// control until this function has terminated, hence the use of 'defer
+	// func() { go func() }' to coordinate waiting and done-ing.
+	defer func() {
+		go func() {
+			wg.Wait()
+			h.tg.Done()
+		}()
+	}()
 	for i := range actionItems {
 		// Add the action item to the wait group outside of the threaded call.
 		// The call to wg.Done() was established at the beginning of the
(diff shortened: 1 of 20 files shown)


lukechampine:
This causes a NDF in my contractor integration test:
```
--- FAIL: TestResync (2.52s)
	host_integration_test.go:652: host did not accept revision request: the requested file contract is currently locked
```

DavidVorick:
The lock timeout in the test may not be long enough.

lukechampine:
same scenario, this time in a different test:
```
--- FAIL: TestIntegrationDelete (7.61s)
	host_integration_test.go:356: host did not accept revision request: the requested file contract is currently locked
```

DavidVorick:
I think I'm going to have to add some custom printlns, been unable to get this to fail locally

DavidVorick:
ok, I've added some code that should improve the safety and add a lot of output. Hopefully it tells us what is going wrong.

DavidVorick:
Problem discovered and fixed. Travis should be happy now.

A deadlock witnessed in the wild caused me to add safelocks to the host. The locking in the host is not all that clean, so I think it's a good idea to have them anyway, at least for a short period of time. I'm thinking about adding a safety feature to the TryMutex as well, that will let you set a max duration that a lock can be held. I don't think it would be difficult to implement (easier than for safelocks - the hard part about the safelocks is that the readlocks mean multiple threads can have a lock at once. For the TryMutex, only one lock can be held, so if it's timing out, it is pretty easy to know who is the culprit.

Anyway, this PR should be good to go, safety in the TryMutex is for another PR.

lukechampine:
mac still exhibiting NDF on TestResync

DavidVorick:
Found it for real this time

DavidVorick:
Recent failure appears to be due to unsafety in the way that the gateway uses ThreadGroups - right now it's possible for the gateway to end up with a single thread that calls Add twice. That's a completely separate PR.

modules/host/update.go
```diff
@@ -313,7 +295,25 @@ func (h *Host) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+	// function have spawned have completed.
+	wg := new(sync.WaitGroup)
+	// The host should not release thread group control until all spawned
+	// threads have completed execution, though waiting for them should not
+	// block this function. The host should also not release thread group
+	// control until this function has terminated, hence the use of 'defer
+	// func() { go func() }' to coordinate waiting and done-ing.
+	defer func() {
+		go func() {
+			wg.Wait()
```
lukechampine: calling `wg.Wait` in a goroutine seems dangerous. It means that during resync, you'll be spawning `threadedHandleActionItem` goroutines at an unbounded rate. It's probably fine if you don't have tons of action items, but it makes me a little uneasy.
Would it be equivalent to simply call `tg.Add` in each `threadedHandleActionItem`, and remove the `WaitGroup`? Or is it important that each set of action items is processed atomically?

DavidVorick: It's not unbounded, because you will only spawn a thread to handle an action item if there is an action item, and since the host is locked you know that action items won't be getting added while you are spinning up the threads.

There is a call to `tg.Add()` at the beginning of `ProcessConsensusChange`, which means all future calls will be blocked. It's important for `Flush()` to wait until all action items have been handled, otherwise you aren't going to know (during testing) if the storage proof has been submitted yet. The purpose of `Flush()` is definitely to make it easy to know that certain actions have completed.

I could potentially see issues if some 'attacker' has formed 10,000 file contracts with you that all expire in the same, but they pay a small price for each contract they form, and that's an attack vector regardless of how the threading is operating.

So basically, every action item that you have to deal with this block would have been created prior to the call to `ProcessConsensusChange` and would mean at least a little bit of income each.

***if it's not clear, `Flush()` and `Stop()` will block all incoming calls to `tg.Add()` which occurs at the top of the `ProcessConsensusChange` function. So even if the call to `ProcessConsensusChange` completes, you know that there will not be another one to add more action items.

lukechampine: I guess I only meant "unbounded" in the sense that you don't have any explicit control over how many goroutines will be running simultaneously; the only bound is the number of action items. In practice it will probably be a small number.
The reason it's weird is that the `tg.Add` and `tg.Done` are happening in separate threads. That's definitely convention-breaking. But I don't see an easy way around it if you're explicitly coding with `Flush()` in mind.
For example, this feels much clearer to me:
```go
func ProcessConsensusChange() {
	tg.Add()
	defer tg.Done()

	// collect action items...

	// handling action items takes time; spawn a new goroutine so we don't block
	go h.threadedHandleActionItems(actionItems)
}

func (h *Host) threadedHandleActionItems(actionItems []types.FileContractID) {
	tg.Add()
	defer tg.Done()

	// handle action items in parallel
	var wg sync.WaitGroup
	wg.Add(len(actionItems))
	for _, item := range actionItems {
		go h.threadedHandleActionItem(item, &wg)
	}
	wg.Wait()
}
```
...but then you need some way of delaying the call to `Flush()` until the `Add` in `threadedHandleActionItems`  has been called.

DavidVorick: Yeah it's just weird because you don't want to block the ProcessConsensusChange function while it's child threads are running (which is part of the reason they are child threads). In terms of closeness, you are still doing `Add; defer Done`, it's just `Add; defer go Done`. It allows you to cleanly encapsulate a bunch of child threads into the same thread group without needing the parent thread to outlive all of the children.

imo it's good design to have concurrency where your parent doesn't need to stick around. It's a little different than how most people approach the problem but I think it allows you to approach a broader range of problems in a more natural way.

lukechampine: I think there are a couple modules that want to avoid blocking ProcessConsensusChange. Usually the problem is that later blocks may invalidate previous blocks (e.g. the contractor trying to renew contracts before being fully synced).

Different concurrency strategies are appropriate for different problems. When you're just using concurrency to speed things up, the parent/child approach makes a lot of sense. But a lot of our code treats goroutines as independent actors that talk to each other. We should read up on the Smalltalk/Erlang approach to concurrency. I bet there's some good ideas to be found there.

DavidVorick: yeah I guess I am really just passing the parent responsibility to another thread.

PR #1357, opened 2016-07-13
fix siac units

DavidVorick:
`siac host -v` has been displaying 'Weeks' as the unit for 'windowsize',
but 'Hours' is the correct human-friendly value. The window size is now
displayed in hours.

`siac host config collateral` had been setting the value incorrectly,
converting using 'Bytes per Terabyte' instead of 'Bytes per terabyte
month'. The correct conversion is now used.

`siac host config -v` has displaying 'int' as the unit for
'maxdownloadbatchsize' and 'maxrevisebatchsize', but the unit is
actually bytes. This has been fixed as well.

### DIFF
--- a/siac/hostcmd.go
+++ b/siac/hostcmd.go
@@ -29,8 +29,8 @@ var (
 Available settings:
      acceptingcontracts:   boolean
      maxduration:          blocks
-     maxdownloadbatchsize: int
-     maxrevisebatchsize:   int
+     maxdownloadbatchsize: bytes
+     maxrevisebatchsize:   bytes
      netaddress:           string
      windowsize:           blocks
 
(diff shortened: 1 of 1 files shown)


lukechampine:
why hours? That seems like a weird unit. Shouldn't it actually be a `time.Duration`? Then it could be printed as hours, days, weeks, whatever, similar to filesize and currency.

DavidVorick:
Because hours is the timescale that's relevant. Renters will reject anything over 48 hours, and anything under 2 hours is pretty risky. Long term optimal values are likely between 3 and 6 hours.

lukechampine:
ok. We should consider using `time.Duration` for time-based values though. Might simplify some parsing.

PR #1360, opened 2016-07-13
tune some of the transaction pool fee estimation values

DavidVorick:


modules/host/storageobligations.go
```diff
@@ -786,7 +786,7 @@ func (h *Host) threadedHandleActionItem(soid types.FileContractID, wg *sync.Wait
 
 		// Create and build the transaction with the storage proof.
 		builder := h.wallet.StartTransaction()
-		feeRecommendation, _ := h.tpool.FeeEstimation()
+		_, feeRecommendation := h.tpool.FeeEstimation()
 		if so.value().Cmp(feeRecommendation) < 0 {
 			// There's no sense submitting the storage proof of the fee is more
```
mnsl: of -> if

modules/transactionpool/transactionpool.go
```diff
@@ -121,7 +121,10 @@ func (tp *TransactionPool) FeeEstimation() (min, max types.Currency) {
 	// TODO: The fee estimation tool should look at the recent blocks and use
 	// them to guage what sort of fee should be required, as opposed to just
 	// guessing blindly.
-	return types.SiacoinPrecision.Mul64(10).Div64(1e3), types.SiacoinPrecision.Mul64(25).Div64(1e3)
+	//
+	// TODO: The current minimum has been reduced significantly to account for
+	// legacy renters that are not correctly adding transaction fees.
+	return types.SiacoinPrecision.Mul64(1).Div64(1e3), types.SiacoinPrecision.Mul64(25).Div64(1e3)
```
mnsl: Need to change comment for `FeeEstimation` in modules/transactionpool.go (line 79) to reflect new minimum

lukechampine: would be nice to have an explanation of the `Div64(1e3)` as well. Otherwise `Mul64(1)` looks very weird

DavidVorick: the comment in modules/transactionpool.go only discusses a number of blocks, not an actual fee.

PR #1361, opened 2016-07-14
fix download bug

lukechampine:
Be very careful when writing `for _, x := range xs` loops.

### DIFF
--- a/modules/renter/download.go
+++ b/modules/renter/download.go
@@ -59,10 +59,10 @@ func (hf *hostFetcher) fetch(p pieceData) ([]byte, error) {
 }
 
 // newHostFetcher creates a new hostFetcher.
-func newHostFetcher(d contractor.Downloader, fc fileContract, masterKey crypto.TwofishKey) *hostFetcher {
+func newHostFetcher(d contractor.Downloader, pieces []pieceData, masterKey crypto.TwofishKey) *hostFetcher {
 	// make piece map
 	pieceMap := make(map[uint64][]pieceData)
-	for _, p := range fc.Pieces {
+	for _, p := range pieces {
 		pieceMap[p.Chunk] = append(pieceMap[p.Chunk], p)
 	}
 	return &hostFetcher{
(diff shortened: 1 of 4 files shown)


siac/rentercmd.go
```diff
@@ -245,18 +245,18 @@ func rentersetallowancecmd(amount, period string) {
...
+
+func (s byValue) Len() int      { return len(s) }
+func (s byValue) Swap(i, j int) { s[i], s[j] = s[j], s[i] }
+func (s byValue) Less(i, j int) bool {
+	cmp := s[i].RenterFunds.Cmp(s[j].RenterFunds)
+	if cmp == 0 {
 		return s[i].NetAddress < s[j].NetAddress
 	}
-	return hi > hj
+	return cmp > 0
```
DavidVorick: I am guessing that you decided to sort things by value because all contracts should be terminating on ~the same block (maybe a few apart)?

siac/rentercmd.go
```diff
@@ -245,18 +245,18 @@ func rentersetallowancecmd(amount, period string) {
...
-type byHeight []api.RenterContract
-
-func (s byHeight) Len() int      { return len(s) }
-func (s byHeight) Swap(i, j int) { s[i], s[j] = s[j], s[i] }
-func (s byHeight) Less(i, j int) bool {
-	hi, hj := s[i].EndHeight, s[j].EndHeight
-	if hi == hj {
+// byValue sorts contracts by their value in siacoins, high to low. If two
+// contracts have the same value, they are sorted by their host's address.
+type byValue []api.RenterContract
```
DavidVorick: the name here is kind of weird. A type `byValue` doesn't really make sense unless it's in the context of the sort function.

modules/renter/download.go
```diff
@@ -191,33 +191,41 @@ func (r *Renter) Download(path, destination string) error {
 
 	// Look up the most recent contract for each host.
 	// NOTE: this assumes that only one contract is made with each host.
-	contracts := make(map[*fileContract]modules.RenterContract)
+	var contractPieces []struct {
+		contract modules.RenterContract
+		pieces   []pieceData
+	}
 	file.mu.RLock()
 	for _, fc := range file.contracts {
```
DavidVorick: seems a little strange that you chose to keep the `for _, obj := range` construction instead of just switching to the `for i := range` construction, given that you needed to add like 20 lines of code to make it happen.

DavidVorick:
You need to add a regression test that proves the error existed in the first place, and then demonstrates that the error has been fixed.

modules/renter/download.go
DavidVorick: I guess using a slice instead of a map makes sense.

DavidVorick:
The sorting code should really also have a test.

siac/rentercmd.go
lukechampine: the contract set should always have the same end height uniformly. If it doesn't, there's a bug.

modules/renter/download.go
lukechampine: it would still be ugly:
```go
for id := range file.contracts {
    fc := file.contracts[id]
    contracts[&fc] = c
}
```
and a newcomer might not understand the rationale.
Mostly I didn't like the old map anyway; using a `*fileContract` as a map key was basically just a way to avoid defining a new type.

siac/rentercmd.go
lukechampine: agreed, but that seems to be the standard for sorting types. And it helps that the type is defined close to where it's used.

lukechampine:
To me, this is another indicator that our current testing strategy is insufficient. The `Download` function had good test coverage (including the buggy code), and was subject to both unit tests and an integration test in the api package. But none of them found the bug. Why? Because our testing assumptions did not match real-world conditions. Specifically, in this case the bug only appears when uploading to more than one host with an erasure code that requires more than one piece.
Is that oversight my fault? Sure. But it's unreasonable to assume that we'll always be able to anticipate these sorts of bugs and/or perfectly replicate real-world conditions during testing. The point is: manual testing uncovered the bug immediately, and I had no trouble locating the cause and fixing it. I think we undervalue that sort of testing.

DavidVorick:
We don't do enough dogfooding, if that's what you mean.

But, the fact that we don't have a test which does a download when a file has been uploaded to more than one host with more than one piece per host demonstrates a huge gap in our automated testing.

ecosystem_test.go should help with these broader tests.

I spent 10 hours yesterday doing manual testing on our live network. I'm not saying that we don't do enough maual testing, but it is very costly. Each iteration of the manual test rook me about 40 minutes where an automated test would take less than three.

We should be using the manual testing to highlight gaps in the automated testing that we then close. If our manual testing is finding bugs it means the automated tests need expansion. And that's fine because we're definitely at a point where we don't know how big or what nature the gaps are in our automated testing.

lukechampine:
any blocking issues here?

PR #1363, opened 2016-07-15
separate ThreadGroup locking by function

DavidVorick:
The threadgroup has two separate things that it is protecting when it
locks the mutex. One group of things is to block the execution of other
threads, in the case of 'Add', 'Flush', and 'Stop'. The other group of
things is to protect the ThreadGroup state variables, namely 'onStopFns'
and 'afterStopFns'.

Previously, you would get a deadlock with the following:

tg.Add()
go func {
	tg.Stop()
}
time.Sleep(1 * time.Second)
tg.OnStop()
tg.Done()

The deadlock happens because tg.Stop() is blocking the OnStop call while
it waits for the tg.Add to return, but the tg.Add will not return until
tg.OnStop can get the lock.

I don't see any reason for this behavior to be forbidden. You want to be
able to do something like:

tg.Add()
tg.RegisterRPC()
tg.OnStop(func() {
	tg.UnregisterRPC()
})
tg.Done()

and currently that's not safe code if you are assuming that Stop() could
be called at any time.

### DIFF
--- a/sync/threadgroup.go
+++ b/sync/threadgroup.go
@@ -26,8 +26,9 @@ type ThreadGroup struct {
 	afterStopFns []func()
 
 	once     sync.Once
-	mu       sync.Mutex
 	stopChan chan struct{}
+	bmu      sync.Mutex // Ensures blocking between calls to 'Add', 'Flush', and 'Stop'
+	mu       sync.Mutex // Protects the 'onStopFns' and 'afterStopFns' variable
 	wg       sync.WaitGroup
 }
 
@@ -49,8 +50,8 @@ func (tg *ThreadGroup) isStopped() bool {
 
 // Add increments the thread group counter.
 func (tg *ThreadGroup) Add() error {
-	tg.mu.Lock()
-	defer tg.mu.Unlock()
+	tg.bmu.Lock()
+	defer tg.bmu.Unlock()
 
 	if tg.isStopped() {
 		return ErrStopped
@@ -105,8 +106,8 @@ func (tg *ThreadGroup) Done() {
 // called 'tg.Done'. This in effect 'flushes' the module, letting it complete
 // any tasks that are open before taking on new ones.
 func (tg *ThreadGroup) Flush() error {
-	tg.mu.Lock()
-	defer tg.mu.Unlock()
+	tg.bmu.Lock()
+	defer tg.bmu.Unlock()
 
 	if tg.isStopped() {
 		return ErrStopped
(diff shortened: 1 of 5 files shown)


sync/threadgroup.go
```diff
@@ -26,8 +26,9 @@ type ThreadGroup struct {
 	afterStopFns []func()
 
 	once     sync.Once
-	mu       sync.Mutex
 	stopChan chan struct{}
+	bmu      sync.Mutex // Ensures blocking between calls to 'Add', 'Flush', and 'Stop'
```
DavidVorick: open to a better name

sync/threadgroup_test.go
```diff
@@ -416,6 +416,57 @@ func TestThreadGroupSiaExample(t *testing.T) {
...
+	<-addChan
+	err := tg.Stop()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if data != 20 {
+		t.Error("20 calls were made to increment data, but value is", data)
+	}
+}
```
DavidVorick: The thread control in this function is pretty cool. No sleeps needed, and the execution is properly synchronized.

lukechampine:
```
--- FAIL: TestIntegrationBroadcastRelayHeader (1.90s)
	synchronize_test.go:1156: RelayHeader didn't broadcast a valid block header
```
have we seen this one before?

DavidVorick:
No, but it looks like it's timeout related. The test is currently only waiting 200ms. If Travis is a shared VM (which I'm pretty sure it is), it's reasonable to think that you might lose 200ms or things might slow down enough for the wait to be missed.

In the short term, I've increased the timeout, but in the long term we should be able to use `Flush()` to skip the timeouts altogether. Though, this test uses a mocked gateway and I'm not sure how that changes things. It seems like the gateway was only mocked so that we could tell when broadcast had been called, but `Flush()` would solve that problem too. Removing the mock means more realistic test coverage, and also it should mean less brittle test coverage.

@VoidingWarranties can you comment on this?

DavidVorick:
```
A lock was held for too long, id '26'. Call stack:

	File: '/Users/travis/gopath/src/github.com/NebulousLabs/Sia/modules/host/storageobligations.go:674'

	File: '/Users/travis/.gimme/versions/go1.6.darwin.amd64/src/runtime/asm_amd64.s:1998'

	File: ':0'
```

The host locks, does some disk I/O, then unlocks. There's no deadlock here, but the lock is timing out, probably because Travis is taking forever. I'm pretty sure at this point I've found the deadlocks in the host that I was looking for, so it might be worth removing the safelocks to get the build passing again. Also they're pretty heavy as far as system resources go.

DavidVorick:
The other NDF is due to a nested `Add` call in the gateway. I've already got a branch open to work on the gateway ThreadGroup management, not sure though if it's worth addressing here. I guess we'll see how frequently the gateway fails.

The problem:

`gateway.Broadcast` spins up a bunch of goroutines that call `gateway.RPC`. `gateway.Broadcast` calls `tg.Add` and then does not call `tg.Done` until all of the `gateway.RPC` calls have returned. The `gateway.RPC` calls also call `tg.Add`. The easiest solution is to break the `RPC` function into an exported and unexported version, though there might be something more elegant that we can do as well.

DavidVorick:
A song from my brother:

"99 bugs in the Sia codebase
99 bugs in the code
take one down, pass it around,
127 bugs in the Sia codebase!"

lukechampine:
When we're down to 0 bugs, we must be extra careful not to fix any more, lest we end up with 4294967295 bugs

DavidVorick:
The circle of life

JoshVorick:
lol

PR #1364, opened 2016-07-15
Consensus threadgroup

DavidVorick:
This PR adds a ThreadGroup to the consensus set. Note that it does not fully implement clean shutdown. It does however provide enough syncing that the ecosystem test can start working.

This thread may experience NDFs. I think most of them are fixed by my threadgroup-safety PR. The rest will be fixed by a ThreadGroup related PR that's in-progress for the gateway.

@mnsl, and anyone else, I ask that you not touch the ecosystem test. I'm still working on it, just wanted to have the proof-of-concept out. It's fine to copy it into a different test and then extend that, if you've got some tests you want to implement.

### DIFF
--- a/modules/consensus/accept.go
+++ b/modules/consensus/accept.go
@@ -19,6 +19,21 @@ var (
 	errOrphan          = errors.New("block has no known parent")
 )
 
+// managedBroadcastBlock will broadcast a block to the consensus set's peers.
+func (cs *ConsensusSet) managedBroadcastBlock(b types.Block) {
+	// COMPATv0.5.1 - broadcast the block to all peers <= v0.5.1 and block header to all peers > v0.5.1.
+	var relayBlockPeers, relayHeaderPeers []modules.Peer
+	for _, p := range cs.gateway.Peers() {
+		if build.VersionCmp(p.Version, "0.5.1") <= 0 {
+			relayBlockPeers = append(relayBlockPeers, p)
+		} else {
+			relayHeaderPeers = append(relayHeaderPeers, p)
+		}
+	}
+	go cs.gateway.Broadcast("RelayBlock", b, relayBlockPeers)
+	go cs.gateway.Broadcast("RelayHeader", b.Header(), relayHeaderPeers)
+}
+
 // validateHeaderAndBlock does some early, low computation verification on the
 // block. Callers should not assume that validation will happen in a particular
 // order.
@@ -237,10 +252,11 @@ func (cs *ConsensusSet) managedAcceptBlock(b types.Block) error {
 			if err == errFutureTimestamp {
 				go func() {
 					time.Sleep(time.Duration(b.Timestamp-(types.CurrentTimestamp()+types.FutureThreshold)) * time.Second)
-					err := cs.AcceptBlock(b)
+					err := cs.managedAcceptBlock(b)
 					if err != nil {
 						cs.log.Debugln("WARN: failed to accept a future block:", err)
 					}
+					cs.managedBroadcastBlock(b)
 				}()
 			}
 			return err
@@ -283,20 +299,16 @@ func (cs *ConsensusSet) managedAcceptBlock(b types.Block) error {
 // without error, it will be relayed to all connected peers. This function
 // should only be called for new blocks.
 func (cs *ConsensusSet) AcceptBlock(b types.Block) error {
-	err := cs.managedAcceptBlock(b)
+	err := cs.tg.Add()
 	if err != nil {
 		return err
 	}
-	// COMPATv0.5.1 - broadcast the block to all peers <= v0.5.1 and block header to all peers > v0.5.1.
-	var relayBlockPeers, relayHeaderPeers []modules.Peer
-	for _, p := range cs.gateway.Peers() {
-		if build.VersionCmp(p.Version, "0.5.1") <= 0 {
-			relayBlockPeers = append(relayBlockPeers, p)
-		} else {
-			relayHeaderPeers = append(relayHeaderPeers, p)
-		}
+	defer cs.tg.Done()
+
+	err = cs.managedAcceptBlock(b)
+	if err != nil {
+		return err
 	}
-	go cs.gateway.Broadcast("RelayBlock", b, relayBlockPeers)
-	go cs.gateway.Broadcast("RelayHeader", b.Header(), relayHeaderPeers)
+	cs.managedBroadcastBlock(b)
 	return nil
 }
--- a/modules/consensus/consensusset.go
+++ b/modules/consensus/consensusset.go
@@ -14,6 +14,7 @@ import (
 	"github.com/NebulousLabs/Sia/encoding"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/persist"
+	"github.com/NebulousLabs/Sia/sync"
 	"github.com/NebulousLabs/Sia/types"
 
 	"github.com/NebulousLabs/bolt"
@@ -81,6 +82,7 @@ type ConsensusSet struct {
 	log        *persist.Logger
 	mu         demotemutex.DemoteMutex
 	persistDir string
+	tg         sync.ThreadGroup
 }
 
 // New returns a new ConsensusSet, containing at least the genesis block. If
@@ -144,6 +146,16 @@ func New(gateway modules.Gateway, persistDir string) (*ConsensusSet, error) {
 		gateway.RegisterRPC("RelayHeader", cs.rpcRelayHeader)
 		gateway.RegisterRPC("SendBlk", cs.rpcSendBlk)
 		gateway.RegisterConnectCall("SendBlocks", cs.threadedReceiveBlocks)
+		cs.tg.OnStop(func() {
+			cs.mu.Lock()
+			defer cs.mu.Unlock()
+
+			cs.gateway.UnregisterRPC("SendBlocks")
+			cs.gateway.UnregisterRPC("RelayBlock")
+			cs.gateway.UnregisterRPC("RelayHeader")
+			cs.gateway.UnregisterRPC("SendBlk")
+			cs.gateway.UnregisterConnectCall("SendBlocks")
+		})
 
 		// Mark that we are synced with the network.
 		cs.mu.Lock()
@@ -188,15 +200,9 @@ func (cs *ConsensusSet) ChildTarget(id types.BlockID) (target types.Target, exis
 
 // Close safely closes the block database.
 func (cs *ConsensusSet) Close() error {
-	cs.mu.Lock()
-	defer cs.mu.Unlock()
-
-	if cs.synced {
-		cs.gateway.UnregisterRPC("SendBlocks")
-		cs.gateway.UnregisterRPC("RelayBlock") // COMPATv0.5.1
-		cs.gateway.UnregisterRPC("RelayHeader")
-		cs.gateway.UnregisterRPC("SendBlk")
-		cs.gateway.UnregisterConnectCall("SendBlocks")
+	err := cs.tg.Stop()
+	if err != nil {
+		return err
 	}
 
 	var errs []error
--- a/modules/consensus/consensusset.go
+++ b/modules/consensus/consensusset.go
@@ -143,7 +143,7 @@ func New(gateway modules.Gateway, persistDir string) (*ConsensusSet, error) {
 		// Register RPCs
 		gateway.RegisterRPC("SendBlocks", cs.rpcSendBlocks)
 		gateway.RegisterRPC("RelayBlock", cs.rpcRelayBlock) // COMPATv0.5.1
-		gateway.RegisterRPC("RelayHeader", cs.rpcRelayHeader)
+		gateway.RegisterRPC("RelayHeader", cs.threadedRPCRelayHeader)
 		gateway.RegisterRPC("SendBlk", cs.rpcSendBlk)
 		gateway.RegisterConnectCall("SendBlocks", cs.threadedReceiveBlocks)
 		cs.tg.OnStop(func() {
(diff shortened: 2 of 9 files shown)


modules/consensus/accept.go
```diff
@@ -19,6 +19,21 @@ var (
 	errOrphan          = errors.New("block has no known parent")
 )
 
+// managedBroadcastBlock will broadcast a block to the consensus set's peers.
+func (cs *ConsensusSet) managedBroadcastBlock(b types.Block) {
+	// COMPATv0.5.1 - broadcast the block to all peers <= v0.5.1 and block header to all peers > v0.5.1.
```
lukechampine: when do we plan to remove this code?

modules/consensus/consensusset.go
```diff
@@ -141,9 +143,19 @@ func New(gateway modules.Gateway, persistDir string) (*ConsensusSet, error) {
 		// Register RPCs
 		gateway.RegisterRPC("SendBlocks", cs.rpcSendBlocks)
 		gateway.RegisterRPC("RelayBlock", cs.rpcRelayBlock) // COMPATv0.5.1
-		gateway.RegisterRPC("RelayHeader", cs.rpcRelayHeader)
+		gateway.RegisterRPC("RelayHeader", cs.threadedRPCRelayHeader)
 		gateway.RegisterRPC("SendBlk", cs.rpcSendBlk)
 		gateway.RegisterConnectCall("SendBlocks", cs.threadedReceiveBlocks)
+		cs.tg.OnStop(func() {
+			cs.mu.Lock()
+			defer cs.mu.Unlock()
```
lukechampine: why is a lock needed here?

modules/consensus/synchronize.go
```diff
@@ -389,13 +403,18 @@ func (cs *ConsensusSet) rpcRelayHeader(conn modules.PeerConn) error {
...
+	// because an exported call to the gateway is used, which is a deadlock
+	// risk given that rpcRelayHeader is called from the gateway.
+	wg.Add(1)
 	go func() {
-		err := cs.gateway.RPC(conn.RPCAddr(), "SendBlk", cs.threadedReceiveBlock(h.ID()))
+		err = cs.gateway.RPC(conn.RPCAddr(), "SendBlk", cs.threadedReceiveBlock(h.ID()))
 		if err != nil {
 			cs.log.Debugln("WARN: failed to get header's corresponding block:", err)
 		}
+		wg.Done()
```
lukechampine: this is bizarre. You don't need a WaitGroup. Just move the `tg.Done` inside this function. It's fine as long as it's properly documented.

modules/consensus/consensusset.go
```diff
@@ -222,6 +228,12 @@ func (cs *ConsensusSet) CurrentBlock() (block types.Block) {
 	return block
 }
 
+// Flush will block until the consensus set has finished all in-progress
+// routines.
+func (cs *ConsensusSet) Flush() error {
+	return cs.tg.Flush()
+}
```
lukechampine: I feel like we shouldn't need this level of introspection in order to do proper integration testing. This creates some pretty tight coupling.

modules/consensus/accept.go
DavidVorick: I almost removed it. But it's a separate concern for a different PR. 

It's probably prudent to wait a little longer.

modules/consensus/consensusset.go
DavidVorick: By convention, a lock here is a deadlock risk because exported functions are being  called below it.

Oops.

DavidVorick: Oh whoops, wrong package. Thought this was the gateway.

You should try removing the locks, on my computer it causes memory faults in bolt??? I am not sure what's going on there, but I should probably investigate further.

modules/consensus/consensusset.go
DavidVorick: It's the only way to get strong reliability on the asynchronous stuff though. Otherwise you make a block and have no idea when it's propagated to the miner, or to the other consensus sets, etc.

I think it's imperfect anyway though.

modules/consensus/synchronize.go
DavidVorick: Can't call tg.Add multiple times from the same thread

DavidVorick: If Stop or Flush is called between the two calls, deadlock.

DavidVorick: WaitGroups have a similar limitation - you can't call wg.Add and wg.Wait at the same time, which is a feature we require for clean shutdown.

lukechampine: It seems like you need to track the goroutine in the `err == errOrphan` branch as well? (Hidden from the diff.) If that's the case, a WaitGroup makes more sense, since the WaitGroup counter might be 2 instead of 1.

modules/consensus/consensusset.go
lukechampine: You can spin on `cs.CurrentBlock` to wait for the block. Or, if you want cleaner synchronization, subscribe to the consensus set:
```go
type subscriberFn func(modules.ConsensusChange)

func (fn subscriberFn) ProcessConsensusChange(cc modules.ConsensusChange) {
    fn(cc)
}

func TestFoo() {
    cs := consensusSet.New()
    blockNotify := make(chan struct{})
    cs.Subscribe(subscriberFn(func(modules.ConsensusChange) {
        blockNotify <- struct{}{}
    }))
    // wait for block
    <-blockNotify
}
```
The above is probably insufficient because you need a corresponding channel read for each block, but with a few tweaks I'm sure it could serve your purposes.

DavidVorick: `cs.CurrentBlock` only needs a ReadLock, which means that it won't wait until `ProcessConsensusChange` has been called on all the modules. That's the problem I was having with the ecosystem test. You don't know when all the modules have finished receiving a block.

Currently, `ProcessConsensusChange` is called on the modules, blocks until they finish processing, and it calls on the modules in the order that they subscribed in. You can use those implementation details to create the strategy above, where you become the last subscriber and then block until the last subscriber has received the block, but then you force the implementation detail that `ProcessConsensusChange` must be blocking at it must call the modules in a knowable order.

I would very much like to go back to asynchronous calls to `ProcessConsensusChange`. It prevents the consensus set from being blocked for potentially long periods of time, and allows for more parallelism in a daemon that currently tends to peg one core. (our bottleneck would still be boltdb though, and I'm not sure increased parallelism would do that much to help it, even on SSDs. We could test it).

Consensus set aside, it's also important for knowing the progress of other modules. For example, the gateway. The consensus set runs, calls `Broadcast`, then... how do we know when the broadcast call has finished? We could spin on the peers that were supposed to receive the block. If there's an error in the sending though, you won't know whether the sender never finished sending or whether the receiver somehow didn't process it. With `Flush()` you can at least be certain that the sender thinks they finished sending.

And same thing with the host. Right now we don't have functions in the TransactionPool that make it easy to tell whether a storage proof has been submitted. We'd have to grab the list of transactions (or maybe subscribe), then iterate through them and identify the storage proof. And even then, we're not sure if it's the right storage proof, and if we're not finding it we're not sure whether we aren't waiting long enough or if the host was having some other issue. But if we expose a `Flush()` call we know that the host at least submitted the storage proof (or it thinks that it did). And we don't need to iterate through the transaction pool to verify that the process is done.

I'm pretty strongly convinced that `Flush()` will make a lot of our testing difficulties go away. It'll be much easier to set up tests that involve large numbers of asynchronous moving parts - something that we very much struggle with today.

modules/consensus/synchronize.go
DavidVorick: Oh, I see what you are saying about the WaitGroup being unnecessary. If we ignore the goroutine that gets spawned in the event of an error, we have a case where a single goroutine is being spawned at the end of execution of the function, meaning you can just add the `tg.Done` inside of the goroutine instead of in a defer statement at the top.

But that's only true because only a single goroutine is spawned, and it gets spawned at the very end of the function. If it was spawned before the end, you'd still need a WaitGroup because you need to know that both the parent and the child thread have completed.

As for the goroutine that gets spawned in the error - I missed it, good catch.

I think I would still prefer to use the `defer go Wait Done` pattern because it applies more broadly. While somewhat unnecessary, it's a recognizable pattern for managing synchronization between functions that spawn multiple goroutines.

DavidVorick:
```
go install -race -tags='dev debug profile' ./api ./build ./compatibility ./crypto ./encoding ./modules ./modules/consensus ./modules/explorer ./modules/gateway ./modules/host ./modules/host/storagemanager ./modules/renter ./modules/renter/contractor ./modules/renter/hostdb ./modules/renter/proto ./modules/miner ./modules/wallet ./modules/transactionpool ./persist ./siac ./siad ./sync ./types

# github.com/NebulousLabs/Sia/siac

siac/hostcmd.go:361: cmd.Usage undefined (type *cobra.Command has no field or method Usage)

siac/main.go:204: cmd.Usage undefined (type *cobra.Command has no field or method Usage)

make: *** [install] Error 2

The command "make" failed and exited with 2 during .

Your build has been stopped.
```

huh?

modules/consensus/consensusset.go
DavidVorick: *It is my plan to expose a `Flush()` function on all modules, to accompany the `Close()` function which already exists in all modules.

modules/consensus/consensusset.go
DavidVorick: I'm having trouble reasoning about whether this should be `OnStop` or `AfterStop`. What exactly happens when the RPCs get removed? Does that just mean that incoming connections are blocked? It shouldn't affect anything else right? Or would calling it early somehow disrupt an ongoing network call?

If we call `OnStop` instead of `AfterStop`, it means the de-registration will happen while there are RPCs in progress. Is that okay?

DavidVorick:
```
2016/07/18 16:01:10 http: Accept error: accept tcp 127.0.0.1:49473: accept: too many open files; retrying in 5ms

--- FAIL: TestRenterHostsAllHandler (0.63s)

	renter_test.go:231: Get http://127.0.0.1:49473/hostdb/all: net/http: HTTP/1 transport connection broken: readLoopPeekFailLocked: EOF

=== RUN   TestIntegrationHostAndRent

--- FAIL: TestIntegrationHostAndRent (0.31s)

	renterhost_test.go:29: open /var/folders/gw/_2jq29095y7b__wtby9dg_5h0000gn/T/SiaTesting/api/TestIntegrationHostAndRent/renter: too many open files

=== RUN   TestExplorerPreset

--- FAIL: TestExplorerPreset (0.00s)

	server_test.go:16: open /var/folders/gw/_2jq29095y7b__wtby9dg_5h0000gn/T/SiaTesting/api/TestExplorerPreset/gateway/gateway.log: too many open files

=== RUN   TestReloading

--- FAIL: TestReloading (0.00s)

	server_test.go:40: open /var/folders/gw/_2jq29095y7b__wtby9dg_5h0000gn/T/SiaTesting/api/TestReloading/gateway/gateway.log: too many open files

=== RUN   TestAuthentication

--- FAIL: TestAuthentication (0.00s)

	server_test.go:71: open /var/folders/gw/_2jq29095y7b__wtby9dg_5h0000gn/T/SiaTesting/authenticated-api/TestAuthentication/gateway/gateway.log: too many open files

=== RUN   TestIntegrationWalletGETEncrypted

--- FAIL: TestIntegrationWalletGETEncrypted (0.00s)

	wallet_test.go:29: Failed to create gateway: open /var/folders/gw/_2jq29095y7b__wtby9dg_5h0000gn/T/SiaTesting/api/TestIntegrationWalletGETEncrypted/gateway/gateway.log: too many open files

=== RUN   TestIntegrationWalletBlankEncrypt

--- FAIL: TestIntegrationWalletBlankEncrypt (0.00s)

	wallet_test.go:94: open /var/folders/gw/_2jq29095y7b__wtby9dg_5h0000gn/T/SiaTesting/api/TestIntegrationWalletBlankEncrypt/gateway/gateway.log: too many open files

=== RUN   TestIntegrationWalletGETSiacoins

--- FAIL: TestIntegrationWalletGETSiacoins (0.00s)

	wallet_test.go:156: open /var/folders/gw/_2jq29095y7b__wtby9dg_5h0000gn/T/SiaTesting/api/TestIntegrationWalletGETSiacoins/gateway/gateway.log: too many open files

=== RUN   TestIntegrationWalletTransactionGETid

--- FAIL: TestIntegrationWalletTransactionGETid (0.00s)

	wallet_test.go:250: open /var/folders/gw/_2jq29095y7b__wtby9dg_5h0000gn/T/SiaTesting/api/TestIntegrationWalletTransactionGETid/gateway/gateway.log: too many open files
```

^ Is this because we are running too many tests in parallel? Is it because clean shutdown is not implemented in enough places?

We might need to switch to nightly builds or a new CI, it's definitely frustrating to be battling Travis all the time. I don't have any plans to reduce the number of files that we have open at a given time, we're just now talking about tests that spin up multiple nodes simultaneously, they'll all have files that go with them.

modules/consensus/consensusset.go
lukechampine: unregistering won't affect in-progress RPCs. In some cases, an RPC calls another RPC, so there's a "race condition" where RPC1 succeeds, but then RPC2 fails (because the recipient unregistered it). But this should be benign as long as the RPC error is handled correctly.

DavidVorick:
gah still running into the 'too many open files' problem. Maybe it's a Travis thing after all?

lukechampine:
did you rebase on top of the other PR?

DavidVorick:
I removed mac from the travis build. I liked having it but the 'too many files' error seemed to be inherent to running 6 daemons simultaneously. Given that each will be running logs and databases for most modules, that's probably like 50 files or something. I'm not sure what the limit is on the mac builds but it seems to be extremely low.

We'll have to explore other CI options at some point. For now, just Linux should be good enough.

lukechampine:
We can use appveyor for Windows builds. It's free for open source, like Travis, so there's no harm in trying. If it frequently fails nondeterministically or the builds are really slow, we can remove it.

PR #1365, opened 2016-07-17
fix broken cobra command

DavidVorick:


DavidVorick:
Merging immediately so I can get the other builds passing. Any comments/complaints can be handled retroactively.

siac/hostcmd.go
```diff
@@ -358,7 +358,7 @@ func hostannouncecmd(cmd *cobra.Command, args []string) {
 	case 1:
 		err = post("/host/announce", "netaddress="+args[0])
 	default:
-		cmd.Usage()
+		cmd.UsageFunc()
```
lukechampine: should be `c.UsageFunc()(c)`

DavidVorick: What is the difference? I ran 'siac' to make sure that the usage output was correct

lukechampine: are you sure you rebuilt the binary? `cmd.UsageFunc()` just returns a function, it doesn't print anything.

DavidVorick: Oh, looks like I didn't rebuild the binary.

>.< derp

PR #1369, opened 2016-07-19
Daemon update testing & UX

mnsl:
* Previously, calling `siac update` on an up-to-date daemon yielded
```
Already up to date.
Updated to version 1.0.0! Restart siad now.
```
rather than just 
```
Already up to date.
```

* Implements `TestUpdate` to check that an API call to `/daemon/update` only signals that an update is available if the build version is behind the latest version (which for a test build means that an update should never be available).

### DIFF
--- a/api/daemon.go
+++ b/api/daemon.go
@@ -22,6 +22,8 @@ import (
 	"github.com/kardianos/osext"
 )
 
+var errEmptyUpdateResponse = errors.New("API call to https://api.github.com/repos/NebulousLabs/Sia/releases/latest is returning an empty response")
+
 // SiaConstants is a struct listing all of the constants in use.
 type SiaConstants struct {
 	GenesisTimestamp      types.Timestamp   `json:"genesistimestamp"`
@@ -113,6 +115,9 @@ func fetchLatestRelease() (githubRelease, error) {
 	err = json.NewDecoder(resp.Body).Decode(&release)
 	if err != nil {
 		return githubRelease{}, err
+	}
+	if release.TagName == "" && len(release.Assets) == 0 {
+		return githubRelease{}, errEmptyUpdateResponse
 	}
 	return release, nil
 }
--- a/api/daemon.go
+++ b/api/daemon.go
@@ -57,7 +57,7 @@ type UpdateInfo struct {
 	Version   string `json:"version"`
 }
 
-// githubRelease represents of the JSON returned by the GitHub release API
+// githubRelease represents some of the JSON returned by the GitHub release API
 // endpoint. Only the fields relevant to updating are included.
 type githubRelease struct {
 	TagName string `json:"tag_name"`
--- a/api/daemon_test.go
+++ b/api/daemon_test.go
@@ -24,6 +24,27 @@ func TestVersion(t *testing.T) {
 	}
 }
 
+// TestUpdate checks that /daemon/update correctly asserts that an update is
+// not available for the daemon (since the test build is always up to date).
+func TestUpdate(t *testing.T) {
+	if testing.Short() {
+		t.SkipNow()
+	}
+	st, err := createServerTester("TestUpdate")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer st.server.Close()
+
+	var update UpdateInfo
+	if err = st.getAPI("/daemon/update", &update); err != nil {
+		t.Fatal(err)
+	}
+	if update.Available && build.Version == update.Version {
+		t.Fatal("daemon should not have an update available")
+	}
+}
+
 /*
 // TODO: enable this test again once proper daemon shutdown is implemented (shutting down modules and listener separately).
 // TestStop tests the /daemon/stop handler.
--- a/api/daemon_test.go
+++ b/api/daemon_test.go
@@ -39,15 +39,31 @@ func TestUpdate(t *testing.T) {
 
 	var update UpdateInfo
 	if err = st.getAPI("/daemon/update", &update); err != nil {
-		// Notify tester that the API call failed, but allow testing to continue.
-		// Otherwise you have to be online to run tests.
+		// Skip test if the GitHub API call fails, since this seems to be a bug
+		// with Travis Darwin.
 		if strings.HasSuffix(err.Error(), errEmptyUpdateResponse.Error()) {
 			t.Skip(err)
 		}
+		// Skip test if testing machine is offline.
+		if strings.HasSuffix(err.Error(), "no such host") {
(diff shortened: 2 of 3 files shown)


mnsl:
Travis Darwin failed twice with this stack trace (no such failure in Travis Linux or my Mac).

The call to [https://api.github.com/repos/NebulousLabs/Sia/releases/latest](https://api.github.com/repos/NebulousLabs/Sia/releases/latest) in `api.fetchLatestUpdate` seems not to be retrieving the expected text file.

Stack trace:
```
2016/07/19 04:13:28 http: panic serving 127.0.0.1:49365: runtime error: slice bounds out of range
goroutine 86 [running]:
net/http.(*conn).serve.func1(0xc820284200)
	/Users/travis/.gimme/versions/go1.6.darwin.amd64/src/net/http/server.go:1389 +0xd2
panic(0x854d40, 0xc82000e110)
	/Users/travis/.gimme/versions/go1.6.darwin.amd64/src/runtime/panic.go:426 +0x521
github.com/NebulousLabs/Sia/api.(*Server).daemonUpdateHandlerGET(0xc8200c4180, 0x1c10af0, 0xc82021e8f0, 0xc820328000, 0x0, 0x0, 0x0)
	/Users/travis/gopath/src/github.com/NebulousLabs/Sia/api/daemon.go:222 +0x23e
github.com/NebulousLabs/Sia/api.(*Server).(github.com/NebulousLabs/Sia/api.daemonUpdateHandlerGET)-fm(0x1c10af0, 0xc82021e8f0, 0xc820328000, 0x0, 0x0, 0x0)
	/Users/travis/gopath/src/github.com/NebulousLabs/Sia/api/api.go:127 +0x6a
github.com/julienschmidt/httprouter.(*Router).ServeHTTP(0xc820444240, 0x1c10af0, 0xc82021e8f0, 0xc820328000)
	/Users/travis/gopath/src/github.com/julienschmidt/httprouter/router.go:344 +0x232
github.com/NebulousLabs/Sia/api.requireUserAgent.func1(0x1c10af0, 0xc82021e8f0, 0xc820328000)
	/Users/travis/gopath/src/github.com/NebulousLabs/Sia/api/api.go:95 +0x11a
net/http.HandlerFunc.ServeHTTP(0xc8204a8270, 0x1c10af0, 0xc82021e8f0, 0xc820328000)
	/Users/travis/.gimme/versions/go1.6.darwin.amd64/src/net/http/server.go:1618 +0x48
net/http.serverHandler.ServeHTTP(0xc820426080, 0x1c10af0, 0xc82021e8f0, 0xc820328000)
	/Users/travis/.gimme/versions/go1.6.darwin.amd64/src/net/http/server.go:2081 +0x207
net/http.(*conn).serve(0xc820284200)
	/Users/travis/.gimme/versions/go1.6.darwin.amd64/src/net/http/server.go:1472 +0x1566
created by net/http.(*Server).Serve
	/Users/travis/.gimme/versions/go1.6.darwin.amd64/src/net/http/server.go:2137 +0x4d2
--- FAIL: TestUpdate (1.66s)
	daemon_test.go:41: Get http://127.0.0.1:49364/daemon/update: EOF
```

api/daemon.go
```diff
@@ -114,6 +114,7 @@ func fetchLatestRelease() (githubRelease, error) {
 	if err != nil {
 		return githubRelease{}, err
 	}
+	fmt.Printf("[FOR TESTING]\nrelease: %v\n", release)
```
lukechampine: ?

DavidVorick:
It looks like `fetchLatestRelease` is returning `err = nil`, but also an empty string for `release.TagName`.

Not sure why that might be, @lukechampine?

`fetchLatestRelease` is a network call, which means that's it's going to fail on a reasonably regular basis. The Travis build probably shouldn't rely on that webpage to be available to test the API call, especially because we want to be able to run the test suite when offline.

We should have some way to mock the response if `build.Release == "testing"`, and then have some service that probes the various things we need to be online, so we can get alerts if it goes down. (that service can come later)

siac/daemoncmd.go
```diff
@@ -50,6 +50,7 @@ func updatecmd() {
 	}
 	if !update.Available {
 		fmt.Println("Already up to date.")
+		return
```
DavidVorick: :+1:

api/daemon.go
mnsl: see the Travis output for Mac vs. Linux: 

Travis Darwin: 
`release: { []}`


Travis Linux & my Mac:
`release: {v1.0.0 [{Sia-v1.0.0-darwin-amd64.zip https://github.com/NebulousLabs/Sia/releases/download/v1.0.0/Sia-v1.0.0-darwin-amd64.zip} {Sia-v1.0.0-linux-amd64.tar.gz https://github.com/NebulousLabs/Sia/releases/download/v1.0.0/Sia-v1.0.0-linux-amd64.tar.gz} {Sia-v1.0.0-windows-amd64.zip https://github.com/NebulousLabs/Sia/releases/download/v1.0.0/Sia-v1.0.0-windows-amd64.zip}]}`

mnsl: This failure should be reported as an error -- will add a check in `fetchLatestRelease`.
Will change `TestUpdate` so that it prints this error if it occurs but still allows the build to pass.

api/daemon_test.go
```diff
@@ -24,6 +25,32 @@ func TestVersion(t *testing.T) {
...
+	if err = st.getAPI("/daemon/update", &update); err != nil {
+		// Notify tester that the API call failed, but allow testing to continue.
+		// Otherwise you have to be online to run tests.
+		if strings.HasSuffix(err.Error(), errEmptyUpdateResponse.Error()) {
+			t.Skip(err)
+		}
+		t.Fatal(err)
+	}
+	if update.Available && build.Version == update.Version {
+		t.Fatal("daemon should not have an update available")
```
lukechampine: seems like `update.Available` should be checked separately from the version strings.

DavidVorick:
Does this test fail if the testing machine is offline?

mnsl:
Travis Linux failed on an unrelated test.
```
--- FAIL: TestIntegrationRPCSendBlocks (34.28s)
	synchronize_test.go:494: Read timeout
	synchronize_test.go:506: SendBlocks should catch up when the remote CS is ahead: expected height 23, got 9
	synchronize_test.go:509: SendBlocks should catch up when the remote CS is ahead: remote and local CSTs have different current blocks
```

lukechampine:
not sure why Travis is getting empty responses. When I run the test locally, it works fine

mnsl:
Same for me.  Also, Travis Linux was getting the correct response until this latest build (only Travis Darwin was getting an empty response).  It's also not the same as the error for an offline machine.

PR #1372, opened 2016-07-19
Test Parallel Transaction Builders

DavidVorick:
I had a suspicion that the wallet would create conflicting transactions and double spends if you used multiple transaction builders at the same time.

This testing pretty thoroughly explores that, and it seems that the wallet does not create conflicting transactions, but is working correctly. Which means the errors in the wild we are seeing are due to some other issue.

This PR only contains a small change to the actual code (first commit), most of it is just testing that didn't fail.

### DIFF
--- a/modules/wallet.go
+++ b/modules/wallet.go
@@ -38,7 +38,7 @@ var (
 
 	// ErrIncompleteTransactions is returned if the wallet has incomplete
 	// transactions being built that are using all of the current outputs, and
-	// therefore the wallet is unable to spend money despite it not techincally
+	// therefore the wallet is unable to spend money despite it not technically
 	// being 'unconfirmed' yet.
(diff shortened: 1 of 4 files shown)


modules/wallet.go
```diff
@@ -36,9 +36,11 @@ var (
 	// complete the desired action.
 	ErrLowBalance = errors.New("insufficient balance")
 
-	// ErrPotentialDoubleSpend is returned when the wallet is uncertain whether
-	// a spend is going to be legal or not.
-	ErrPotentialDoubleSpend = errors.New("wallet has coins spent in unconfirmed transactions - not enough remaining coins to complete transaction")
+	// ErrIncompleteTransactions is returned if the wallet has incomplete
+	// transactions being built that are using all of the current outputs, and
+	// therefore the wallet is unable to spend money despite it not techincally
```
mnsl: techincally -> technically

PR #1373, opened 2016-07-20
set responses to close

DavidVorick:
HTTP connections are persistent by default, unless you specify in the
header that the server is supposed to close the connection once the call
has finished.

This was causing 'too many files open' errors in our Travis build
system.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -42,7 +42,7 @@ func HttpGET(url string) (resp *http.Response, err error) {
 		return nil, err
 	}
 	req.Header.Set("User-Agent", "Sia-Agent")
-	return new(http.Client).Do(req)
+	return http.DefaultClient.Do(req)
 }
 
 // HttpGETAuthenticated is a utility function for making authenticated http get
@@ -55,7 +55,7 @@ func HttpGETAuthenticated(url string, password string) (resp *http.Response, err
 	}
 	req.Header.Set("User-Agent", "Sia-Agent")
 	req.SetBasicAuth("", password)
-	return new(http.Client).Do(req)
+	return http.DefaultClient.Do(req)
 }
 
 // HttpPOST is a utility function for making post requests to sia with a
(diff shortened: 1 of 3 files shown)


DavidVorick:
If all goes well, this should fix all of the 'too many files' errors that we are seeing in Travis.

lukechampine:
This could hurt latency during normal operation, although only the UI would be affected (siac already has to create a new connection each time). Might want to wrap in `testing` blocks.

DavidVorick:
I don't think that the UI uses these calls, it uses a JS library to talk to the API. It might affect `siac`, but `siac` is supposed to be making one-off calls anyway.

Any call that's initiating a connection should also be closing it, if you want a persistent connection for performance reasons, you should probably be using a different function.

Unless, we want to create an init-style function that establishes a connection with the daemon and then re-uses it between calls, and then closes when the daemon closes. That would also prevent the 'too many files' situation, but it's enough of a change that it would need further testing and would be a separate PR.

lukechampine:
ah, you're right, the UI is not affected.

In practice I doubt there will be noticeable overhead because the connection is local.

lukechampine:
`http.Transport` provides a `DisableKeepAlives` field. Behind the scenes, it appears to effect the same behavior as this PR (setting the header field). I think `DisableKeepAlives` is more explicit, but it's also a bit more code.

DavidVorick:
From the godoc:

>  Clients and Transports are safe for concurrent use by multiple goroutines and for efficiency should only be created once and re-used. 

ended up being a pretty simple change to make a global client and just use it all the time, the PR has changed entirely.

lukechampine:
ah, that is even better. http already provides this variable though:
```go
var DefaultClient = &Client{}
```
So we can do `return http.DefaultClient.Do(req)`

DavidVorick:
Fixed. Notably, the build has not failed for 'too many files' during any of these changes. Given that it was failing almost every try in other PRs, it seems very likely that we've fixed the problem.

PR #1375, opened 2016-07-20
don't check AcceptingContracts during revision

lukechampine:


DavidVorick:
Merging this early so we can build, but it needs a test demonstrating that it wasn't working before and is working now.

PR #1377, opened 2016-07-21
Verify API security when host is blank

VoidingWarranties:
And do not verify API security in DEBUG mode. See commit message for reasoning.

### DIFF
--- a/siad/daemon.go
+++ b/siad/daemon.go
@@ -33,7 +33,7 @@ func verifyAPISecurity(config Config) error {
 	// --disable-api-security flag has been used.
 	if !config.Siad.AllowAPIBind {
 		addr := modules.NetAddress(config.Siad.APIaddr)
-		if !addr.IsLoopback() && addr.Host() != "" {
+		if !addr.IsLoopback() {
 			return errors.New("you must pass --disable-api-security to bind Siad to a non-localhost address")
 		}
 		return nil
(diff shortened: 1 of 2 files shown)


VoidingWarranties:
Do not merge yet

VoidingWarranties:
Oh, I overlooked that disabling the check in DEBUG would break the tests. Hmm, maybe the convenience provided isn't enough reason.

DavidVorick:
We need to be able to test this function, you can't disable it in testing.

Also, definitely need to extend the tests to show that the flag warning will trigger for `--api-addr :xxxx`

DavidVorick:
Also, is it really that bad to add the --diable-api-security flag when using a docker container? Definitely not worth it.

VoidingWarranties:
### Done
* Rebased
* Removed the commit that skipped security check in DEBUG
* Fixed tests
* Added an additional message to the error output if the host is blank e.g.
```
$ siad -d sia_test/ --api-addr 9980
an empty host will listen on all interfaces, did you mean localhost:9980?
you must pass --disable-api-security to bind Siad to a non-localhost address
```

### To do
* [ ] Consider special casing the `--api-addr` argument to default to `localhost` if the host is not specified.
    - should `x` default to `localhost:x`?
    - should `:x` default to `localhost:x`? This would make it impossible to listen on all interfaces. I'm not aware of a valid use case for this, but there might be.
* [ ] Update `TestUnitProcessNetAddr`, `TestUnitProcessConfig`, and `TestVerifyAPISecurity` accordingly.

PR #1378, opened 2016-07-21
Contract compensation, host logging

DavidVorick:
Added an extra line of display to `siac host -v`, and added a bunch of calls to `h.log.Debugln` in the event of errors during negotiation. Currently, most errors did get logged but without any context.


modules/host/negotiateformcontract.go
```diff
@@ -128,6 +129,7 @@ func (h *Host) managedRPCFormContract(conn net.Conn) error {
 	// Send the host settings to the renter.
 	err := h.managedRPCSettings(conn)
 	if err != nil {
+		h.log.Debugln("RPCSettings call failed during form contract:", err)
 		return err
```
lukechampine: why not just add this message to the error?

DavidVorick: Is the question general to all the logging statements, or just this one?

By making a logging statement, we get the line of code that is declaring the logging. A lot of the messages across the files are very similar, so having the exact file location, while fragile to changes in the codebase, give us a more complete idea of what is happening, especially if the message itself ends up being not nearly specific enough

lukechampine: I've found that if you add enough context, it's easy to grep for the exact error. For example, the `proto` error code looks like this:

```go
if err := encoding.WriteObject(conn, rev); err != nil {
	return errors.New("couldn't send revision: " + err.Error())
}
if err := modules.ReadNegotiationAcceptance(conn); err != nil {
	return errors.New("host did not accept revision: " + err.Error())
}
if err := encoding.WriteObject(conn, signedTxn.TransactionSignatures[0]); err != nil {
	return errors.New("couldn't send transaction signature: " + err.Error())
}
```

The only downside here is that you can't check for sentinel errors. For that, we'd want to use something like `pkg/errors` to wrap and unwrap the underlying error in context.

DavidVorick:
I have added a new file, 'errors.go', to experiment with a new type of error reporting and logging. negotiatedownload.go has full error reporting implemented, including a large amount of context being added to errors.

Logging will log the first 250 of each type of error, instead of letting one type of error dominate the log and then having all the other errors get blocked. Nondeterministic logging is not yet implemented, but it would be pretty easy to log errors nondeterministically based on type (1-in-5 for internal errors, 1-in-20 for communication errors, 1-in-50 for connection errors, etc.)

DavidVorick:
There's a pretty significant amount of testing that needs to be added, and also the probabilistic logging for normal servers, but then this PR is ready to go.

modules/host/errors.go
```diff
@@ -0,0 +1,137 @@
...
+// mangedLogError will take an error and log it to the host, depending on the
+// type of error and whether or not the DEBUG flag has been set.
+func (h *Host) managedLogError(err error) {
+	// Determine the error type and the number of errors we've seen of that
+	// type previously.
+	var num uint64
+	switch err.(type) {
+	case ErrorCommunication:
+		atomic.AddUint64(&h.atomicCommunicationErrors, 1)
+		num = atomic.LoadUint64(&h.atomicCommunicationErrors)
```
lukechampine: `atomic.AddUint64` returns the new value

modules/host/negotiatedownload.go
```diff
@@ -93,7 +35,7 @@ func (h *Host) managedDownloadIteration(conn net.Conn, so *storageObligation) er
 	// The renter will either accept or reject the host's settings.
 	err = modules.ReadNegotiationAcceptance(conn)
 	if err != nil {
-		return err
+		return extendErr("renter rejected host settings: ", ErrorCommunication(err.Error()))
```
lukechampine: at first I thought this would result in:
`renter rejected host settings: communication error: foo`
but after rereading `extendErr` I see it's actually:
`communication error: renter rejected host settings: foo`
isn't this the same as:
`ErrorCommunication(fmt.Sprint("renter rejected host settings:", err))`
I suppose the advantage of `extendErr` is that if you extend an error multiple times, it will retain the same type. You'd need to be aware of this, though. For example, in the code above, if `ReadNegotiationAcceptance` returned `ErrorConnection`, it would be overwritten by `ErrorCommunication`.

DavidVorick: For the most part, I tried to be conscious of two things when writing errors:

1. the error being returned may already be typed, but only if being returned by an internal-method call. So, `modules.ReadNegotiationAcceptance` won't return a typed error b/c it's not in the host package
2. the calling function is going to add context, so you don't need to. For example, if RPCFormContract calls RPCSettings, and RPCSettings returns an error, you'll get context up the whole stack. RPCSettings will extend with 'could not write secret key', RPCFormContract will extend with 'error during RPCSettings', and network.go will extend with 'error during RPCFormContract', so the full error ends up being:

"connection error: error during RPCFormContract: error during RPCSettings: could not write secret key: EOF" or something of that sort. You basically get the whole call stack in the form of context.

DavidVorick:
I have finished testing the error handling, found and fixed some bugs while doing the testing.

To get the testing completed, it was necessary to change the behaviour of `log.Debugln` slightly, I needed to add a string that marks the log as a debugging statement. This is probably a good change for the project regardless.

persist/log.go
```diff
@@ -45,7 +45,7 @@ func (l *Logger) Debugf(format string, v ...interface{}) {
 // it is a no-op.
 func (l *Logger) Debugln(v ...interface{}) {
 	if build.DEBUG {
-		l.Output(2, fmt.Sprintln(v...))
+		l.Output(2, "DEBUG: "+fmt.Sprintln(v...))
```
lukechampine: typically we include a `"WARN:"` or `"ERROR:"` in the log message. Perhaps `"[DEBUG]"` is more suitable

modules/host/errors_test.go
```diff
@@ -0,0 +1,330 @@
...
+	// Try extending ErrorCommunication.
+	err = ErrorCommunication("err")
+	err2 = extendErr("extend: ", err)
+	if err2.Error() != "communication error: extend: err" {
+		t.Error("extending ErrorCommunication did not occur correctly:", err2.Error())
+	}
+	switch err2.(type) {
+	case ErrorCommunication:
+	default:
+		t.Error("extended error did not preserve error type")
```
lukechampine: use `if _, ok := err2.(ErrorCommunication); !ok`

modules/host/errors_test.go
```diff
@@ -0,0 +1,330 @@
...
+
+	"github.com/NebulousLabs/Sia/modules"
+)
+
+// countFileLines is a helper function that will count the number of lines in a
+// file, based on the number of '\n' characters. countFileLines will load the
+// file into memory using ioutil.ReadAll.
+//
+// countFileLines will ignore all lines with the string 'DEBUG' in it.
+func countFileLines(filepath string) (uint64, error) {
```
lukechampine: alternatively, you could replace the host's logger with a mock:
```go
type linecountWriter int

func (count *linecountWriter) Write(p []byte) (int, error) {
    count += bytes.Count(p, []byte("\n"))
    return len(p), nil
}

var count linecountWriter
h.log = persist.NewLogger(count)
```

DavidVorick: I considered it, but you run into complications because the logging function needs access to other parts of the host state, namely the counters.

DavidVorick: Alternatively we could just have the logger itself do all the probabilistic logging, it could look for a special error interface that has Type() and Probability() functions

lukechampine: >the logging function needs access to other parts of the host state, namely the counters

I don't follow. Swapping out the logger shouldn't affect the behavior of `managedLogError`

DavidVorick: no but it might screw something else up, especially if you don't replace all of the methods. A logger is a type, not an interface, which makes it more difficult to swap out without side effects.

It probably wouldn't be too bad in this case, but the code to do it without mocking is already in place, and imo if you don't need to mock it's better to get your coverage without it.

DavidVorick:
Fixed the small stuff, would rather focus on other things than mock the logger when the test suite is already doing its job.

DavidVorick:
Also added some more context: the host will now report the RemoteAddr of the renter that originated a negotiation error, so we can  tell when an error is just one broken renter on repeat vs. a bunch of different renters all having the same problems.

modules/host/errors_test.go
lukechampine: you're swapping out the writer, not the logger. All it means is that whatever you write to the logger gets swallowed instead of written to disk.

DavidVorick: oh, I see. I misunderstood.

DavidVorick:
Found a pretty serious deadlock, went ahead and popped the fix in for that. Also changed once instance of `save` to `saveSync` because it's a really important operation, and it only ever happens one time per host (even after reboots, it will not run again).

I believe at this point there's still a negative currency error for the host.

DavidVorick:
Adding safelocks to the host has caused a lot of problems, because as it happens the host I/O is not being threaded well. They are detecting fake deadlocks, resolving them in a somewhat brutal way, and introducing race conditions.

I'll be removing them in this PR because the changes needed to resolve the I/O problems are a lot more significant. I'm actively working on those fixes, but the changes are pretty substantial. (basically a blanket rewrite of the storage manager).

DavidVorick:
This should be ready to go, is there anything blocking?

lukechampine:
I guess I'm just wondering what the value is in tracking all the different kinds of errors. Is that information useful? Is it actionable?

I like the general concept of adding context to errors; I think [pkg/errors](https://github.com/pkg/errors) does it best. But you can reduce the complexity of this PR quite a bit if you strip out the different error types. You'd still need a new error type for probabilistic logging, but you could apply it only as needed, i.e. only to errors which seem to be very frequent.

If you'd rather not spend more time with this, I can merge it now and we'll circle back to it later

DavidVorick:
Yeah, let's come back to it, it's good enough for now. Parsing a log file is I/O, reading an atomic value is very fast. Next PR removes the counts from persistence, so you get reset to zero each reboot. Much more actionable that way imo, been using the logs and error dashboard this week to debug the hosts

lukechampine:
>Parsing a log file is I/O

I was assuming the dashboard would be an external program, not internal to the host. If you're just reading log lines as they come in (instead of reading the entire log file each time), the I/O is comparable to the I/O required to make an API call.

One benefit of the counting is that every error is counted, but not all errors are logged (due to probabilistic logging). But that seems like a pretty minor benefit.

PR #1379, opened 2016-07-26
properly return the latest block facts from /explorer

johnathanhowell:
This PR adds `LatestBlockFacts()`, which uses the explorer's internal block height to get the latest `BlockFacts`.  Previously, the consensus set's block height was used, which sometimes caused empty block facts to be returned because the explorer's block facts are not always up to date with the latest consensus set.

### DIFF
--- a/modules/explorer/info.go
+++ b/modules/explorer/info.go
@@ -1,8 +1,12 @@
 package explorer
 
 import (
+	"errors"
+	"github.com/NebulousLabs/Sia/build"
+	"github.com/NebulousLabs/Sia/encoding"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/types"
+	"github.com/NebulousLabs/bolt"
 )
 
 // Block takes a block ID and finds the corresponding block, provided that the
@@ -40,22 +44,28 @@ func (e *Explorer) BlockFacts(height types.BlockHeight) (modules.BlockFacts, boo
 
 // LatestBlockFacts returns a set of statistics about the blockchain as they appeared
 // at the latest block height in the explorer's consensus set.
-func (e *Explorer) LatestBlockFacts() (modules.BlockFacts, bool) {
-	var height types.BlockHeight
-	err := e.db.View(dbGetInternal(internalBlockHeight, &height))
-	if err != nil {
-		return modules.BlockFacts{}, false
-	}
-	block, exists := e.cs.BlockAtHeight(height)
-	if !exists {
-		return modules.BlockFacts{}, false
-	}
+func (e *Explorer) LatestBlockFacts() modules.BlockFacts {
 	var bf blockFacts
-	err = e.db.View(dbGetAndDecode(bucketBlockFacts, block.ID(), &bf))
+	err := e.db.View(func(tx *bolt.Tx) error {
+		var height types.BlockHeight
+		err := encoding.Unmarshal(tx.Bucket(bucketInternal).Get(internalBlockHeight), &height)
+		if err != nil {
+			return err
+		}
+		block, exists := e.cs.BlockAtHeight(height)
+		if !exists {
+			build.Critical(errors.New("latest explorer block doesnt exist in consensus set"))
+		}
+		bfBytes := tx.Bucket(bucketBlockFacts).Get(encoding.Marshal(block.ID()))
+		if bfBytes == nil {
+			build.Critical(errors.New("block facts for the latest explorer block dont exist"))
+		}
+		return encoding.Unmarshal(bfBytes, &bf)
+	})
 	if err != nil {
-		return modules.BlockFacts{}, false
+		build.Critical(err)
 	}
-	return bf.BlockFacts, true
+	return bf.BlockFacts
 }
 
 // Transaction takes a transaction ID and finds the block containing the
--- a/modules/explorer/info.go
+++ b/modules/explorer/info.go
@@ -38,6 +38,26 @@ func (e *Explorer) BlockFacts(height types.BlockHeight) (modules.BlockFacts, boo
 	return bf.BlockFacts, true
 }
 
+// LatestBlockFacts returns a set of statistics about the blockchain as they appeared
+// at the latest block height in the explorer's consensus set.
+func (e *Explorer) LatestBlockFacts() (modules.BlockFacts, bool) {
+	var height types.BlockHeight
+	err := e.db.View(dbGetInternal(internalBlockHeight, &height))
+	if err != nil {
+		return modules.BlockFacts{}, false
+	}
+	block, exists := e.cs.BlockAtHeight(height)
+	if !exists {
+		return modules.BlockFacts{}, false
+	}
+	var bf blockFacts
+	err = e.db.View(dbGetAndDecode(bucketBlockFacts, block.ID(), &bf))
+	if err != nil {
+		return modules.BlockFacts{}, false
+	}
+	return bf.BlockFacts, true
+}
+
 // Transaction takes a transaction ID and finds the block containing the
 // transaction. Because of the miner payouts, the transaction ID might be a
 // block ID. To find the transaction, iterate through the block.
--- a/modules/explorer/info.go
+++ b/modules/explorer/info.go
@@ -1,10 +1,7 @@
 package explorer
 
 import (
-	"errors"
-	
 	"github.com/NebulousLabs/Sia/build"
-	"github.com/NebulousLabs/Sia/encoding"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/types"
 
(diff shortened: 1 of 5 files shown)


DavidVorick:
hmm. Did you observe this error in the wild? The consensus set will hold a lock until all of the modules have the most recent update. I guess it only holds a readlock though, which could cause problems.

Regardless, I have plans so that the consensus set will not block while all the modules process blocks, which should help with performance as you get much better parallelism around I/O bottlenecks and the like.

modules/explorer/info.go
```diff
@@ -38,6 +38,26 @@ func (e *Explorer) BlockFacts(height types.BlockHeight) (modules.BlockFacts, boo
 	return bf.BlockFacts, true
 }
 
+// LatestBlockFacts returns a set of statistics about the blockchain as they appeared
+// at the latest block height in the explorer's consensus set.
+func (e *Explorer) LatestBlockFacts() (modules.BlockFacts, bool) {
```
DavidVorick: bit weird to return false here, the latest block should always exist

modules/explorer/info.go
```diff
@@ -38,6 +38,26 @@ func (e *Explorer) BlockFacts(height types.BlockHeight) (modules.BlockFacts, boo
...
+	err := e.db.View(dbGetInternal(internalBlockHeight, &height))
+	if err != nil {
+		return modules.BlockFacts{}, false
+	}
+	block, exists := e.cs.BlockAtHeight(height)
+	if !exists {
+		return modules.BlockFacts{}, false
+	}
+	var bf blockFacts
+	err = e.db.View(dbGetAndDecode(bucketBlockFacts, block.ID(), &bf))
```
DavidVorick: You want to lump these two database transactions into a single transaction, bolt calls to `View` are expensive. Calls to `Update` are substantially more expensive, but even View is pretty slow. You can make the `e.cs.BlockAtHeight` call from within the View call:

`e.db.View(func(tx *bolt.Tx) {
    stuff1;
    stuff2;
})

DavidVorick:
lol you are changing the PR as I'm writing comments

modules/explorer/info.go
johnathanhowell: Yeah, @lukechampine and I had a brief offline conversation about this.  Would returning an error be saner?

modules/explorer/info.go
```diff
@@ -38,6 +38,17 @@ func (e *Explorer) BlockFacts(height types.BlockHeight) (modules.BlockFacts, boo
...
 
+// LatestBlockFacts returns a set of statistics about the blockchain as they appeared
+// at the latest block height in the explorer's consensus set.
+func (e *Explorer) LatestBlockFacts() (modules.BlockFacts, bool) {
+	var height types.BlockHeight
+	err := e.db.View(dbGetInternal(internalBlockHeight, &height))
+	if err != nil {
+		return modules.BlockFacts{}, false
+	}
+	return e.BlockFacts(height)
```
DavidVorick: don't call an exported function from within an exported function, the concurrency conventions we use can result in deadlocks by doing that.

Within an exported function, you can safely assume full control of a mutex, assume that nobody else holding one has called you, etc. `defer Unlock` is a common pattern, and a person deciding to call that shouldn't be frustrated by having to deal with another exported call from inside the function, if at all possible.

modules/explorer/info.go
DavidVorick: I'd return just the block facts, then `build.Critical` if you don't get them.

modules/explorer/info.go
johnathanhowell: ACK.  I'm just going to roll back this commit then

modules/explorer/info.go
DavidVorick: not sure if you saw this earlier, but you should be combining these db transactions in to a single transaction.

johnathanhowell:
Latest commit addresses your comments re: one transaction and build.Critical, and updates TestImmediateBlockFacts to use `LatestBlockFacts`.

modules/explorer/info.go
```diff
@@ -1,8 +1,12 @@
 package explorer
 
 import (
+	"errors"
+	"github.com/NebulousLabs/Sia/build"
+	"github.com/NebulousLabs/Sia/encoding"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/types"
+	"github.com/NebulousLabs/bolt"
```
lukechampine: convention is to separate imports like so:
```go
 import (
	"errors"

)
```
i.e. first stdlib, then internal, then external

modules/explorer/info.go
```diff
@@ -38,6 +42,32 @@ func (e *Explorer) BlockFacts(height types.BlockHeight) (modules.BlockFacts, boo
 	return bf.BlockFacts, true
 }
 
+// LatestBlockFacts returns a set of statistics about the blockchain as they appeared
+// at the latest block height in the explorer's consensus set.
+func (e *Explorer) LatestBlockFacts() modules.BlockFacts {
+	var bf blockFacts
+	err := e.db.View(func(tx *bolt.Tx) error {
+		var height types.BlockHeight
+		err := encoding.Unmarshal(tx.Bucket(bucketInternal).Get(internalBlockHeight), &height)
```
lukechampine:
```go
err := dbGetInternal(internalBlockHeight, &height)(tx)
```

modules/explorer/info.go
```diff
@@ -38,6 +42,32 @@ func (e *Explorer) BlockFacts(height types.BlockHeight) (modules.BlockFacts, boo
...
+		}
+		block, exists := e.cs.BlockAtHeight(height)
+		if !exists {
+			build.Critical(errors.New("latest explorer block doesnt exist in consensus set"))
+		}
+		bfBytes := tx.Bucket(bucketBlockFacts).Get(encoding.Marshal(block.ID()))
+		if bfBytes == nil {
+			build.Critical(errors.New("block facts for the latest explorer block dont exist"))
+		}
+		return encoding.Unmarshal(bfBytes, &bf)
```
lukechampine:
```go
return dbGetAndDecode(bucketBlockFacts, block.ID(), &bf)(tx)
```

lukechampine: actually this function could just be:
```go
var height types.BlockHeight
err := e.db.View(dbGetInternal(internalBlockHeight, &height))
if err != nil {
    build.Critical(err)
}
return e.BlockFacts(height)
```
DRY and whatnot. The only downside is that this means you're doing two db transactions instead of one.

DavidVorick: You missed some of my earlier comments, definitely we should avoid calling exported functions within exported functions wherever possible, it's a deadlock risk given our concurrency style.

modules/explorer/info.go
```diff
@@ -38,6 +42,32 @@ func (e *Explorer) BlockFacts(height types.BlockHeight) (modules.BlockFacts, boo
 	return bf.BlockFacts, true
 }
 
+// LatestBlockFacts returns a set of statistics about the blockchain as they appeared
+// at the latest block height in the explorer's consensus set.
+func (e *Explorer) LatestBlockFacts() modules.BlockFacts {
+	var bf blockFacts
```
DavidVorick: safety tells me that we should be grabbing a readlock on the explorer here, with a deferred unlock. Bolt has some special things that might make it unnecessary, but then you are assuming that all stateful accesses in this function are bolt, and that's cognitive overhead that's just invite race-condition and dead-lock based edge cases in the future.

modules/explorer/info.go
```diff
@@ -38,6 +42,32 @@ func (e *Explorer) BlockFacts(height types.BlockHeight) (modules.BlockFacts, boo
...
+	err := e.db.View(func(tx *bolt.Tx) error {
+		var height types.BlockHeight
+		err := encoding.Unmarshal(tx.Bucket(bucketInternal).Get(internalBlockHeight), &height)
+		if err != nil {
+			return err
+		}
+		block, exists := e.cs.BlockAtHeight(height)
```
DavidVorick: so... present module design (which is bad, and I wish to fix soon) means that it's a deadlock risk to call an exported consensus function while holding an explorer lock, which we should be holding. I believe though that the explorer has an internal way of looking up the block at a given height.

DavidVorick: based on the state of the database, I don't think there's a way to avoid this. Given that we frequently load the most recent block facts for a block (it's the homepage), we really shouldn't have to go to disk to figure out the id of the most recent block, it should just be kept in memory as the consensus set updates.

modules/explorer/info.go
lukechampine: The explorer does not have its own mutex. It doesn't store anything in memory at all. bolt's internal locking is sufficient. This may change if we later decide to cache things in memory.

lukechampine: If you want to mimic the existing paradigm, the closest approximation would be to create an unexported function `dbBlockFacts`:
```go
func (e *Explorer) dbGetBlockFacts(height types.BlockHeight) func(*bolt.Tx) error {
    // ...
}
```
Then `BlockFacts` and `LatestBlockFacts` would both call this function from within a bolt transaction, which serves the role of a mutex.

PR #1382, opened 2016-07-26
Clarify username in API authentication docs

VoidingWarranties:
The username required for API authentication wasn't documented anywhere, and was confusing users (http://forum.sia.tech/topic/302/what-s-the-authenticate-api-username-rfc-2617).

The username isn't actually checked in the API authentication, but I didn't mention that as it's feasible to imagine an authentication scheme where different users have different permissions for different endpoints.

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -56,9 +56,15 @@ Authentication
 
 API authentication can be enabled with the `--authenticate-api` siad flag.
 Authentication is HTTP Basic Authentication as described in
-[RFC 2617](https://tools.ietf.org/html/rfc2617). The flag does not enforce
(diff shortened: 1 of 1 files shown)


lukechampine:
```
--- FAIL: TestUpdate (0.47s)
	daemon_test.go:59: the /daemon/update API call is reporting a version that is not up to date but says no update is available
```
This is failing because the version number has been bumped, but the release hasn't been made yet. Maybe this test needs some modification. @mnsl 
The alternative would be to not bump the version number until the release has been made, but that's not acceptable because releases need to be tagged, and the tag needs to be after the version bump.

PR #1387, opened 2016-07-28
Modify siad  to allow non-localhost API port when security is disabled

### LINKED ISSUE #1386: --disable-api-security does not disable API security
The change in behavior of the `--disable-api-security` introduced in v1.0.1 is unintuitive and breaks user scenarios.

In 1.0.1, if the user passes `--disable-api-security` siad requires the user to add an API password. This is confusing because this does not seem to honor the user's request to disable the security measures on the API port; it just applies different security requirements.

This requirement breaks scenarios like [Sia via Docker](http://mtlynch.io/sia-via-docker/). It is not possible for anything outside of the Docker container to communicate with the container's localhost interface. Even if the user tries to add an API password, it's klunky to script because siad expects the password to be entered by an interactive user rather than as part of a script.

mtlynch:
Changes the behavior of the siad command line processor to allow the user to
bind to a non-local interfact to listen for API calls when the user has passed
the --disable-api-security flag.

Fixes #1386.

### DIFF
--- a/siad/daemon.go
+++ b/siad/daemon.go
@@ -42,11 +42,6 @@ func verifyAPISecurity(config Config) error {
 		return nil
 	}
 
-	// If the --disable-api-security flag is used, enforce that
-	// --authenticate-api must also be used.
-	if config.Siad.AllowAPIBind && !config.Siad.AuthenticateAPI {
-		return errors.New("cannot use --disable-api-security without setting an api password")
-	}
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


DavidVorick:
Sorry to leave this open for almost a year. Going to stand behind the assessment that it's better to force unique situations to work a little harder than to put everyday users at risk.

The only reason I'm being stubborn here is because we've had miners actually lose money to attackers for improper configuration, and the safety you are removing was responsible for stopping the thefts.

PR #1392, opened 2016-07-29
attempt to limit bolt memory usage

DavidVorick:
I think this PR may limit the amount of memory that siad is using by a factor of 2 or 3. Having Sia sit at 800MB is a lot better than 2GB.

I'm not completely sure though, testing has been unclear/incomplete, trying it on my live nodes. Wanted @VoidingWarranties to give it a shot on his nodes as well, see if it makes much of a difference.


DavidVorick:
*if memory usage does not decrease substantially then this PR is not worth merging.

lukechampine:
what's the status on this?

DavidVorick:
I have been pretty sucked into the host I/O stuff, haven't had time to verify whether or not RAM usage is reduced. Would be very helpful if someone else could run the code on their machine, compare RAM usage.

The easiest test I can think of is to take a fully synced machine with a new wallet (no renter or host or miner), then unlock the wallet. On the master branch, the RAM usage should be much higher than on this branch. If that is the case, we merge it. If RAM usage is high or moderate on both, we don't merge it.

DavidVorick:
Further testing suggests that this PR does not reduce memory usage.

PR #1393, opened 2016-07-29
Add a database to the Wallet

lukechampine:
This is the first in a series of PRs that will be merged into a wallet feature branch. The feature branch is being employed to prevent breaking compatibility until all of the features have been implemented.

The high-level goals for this feature branch are:
- [ ] Instant wallet unlocking
- [ ] Ability to create wallet from a seed directly (instead of creating, then loading)
- [ ] Preserve compatibility with old wallets (by automatically converting to new format on first load)


modules/wallet/database.go
```diff
@@ -0,0 +1,32 @@
...
+
+func dbPutHistoricOutput(tx *bolt.Tx, id types.OutputID, c types.Currency) error {
+	return tx.Bucket(bucketHistoricOutputs).Put(encoding.Marshal(id), encoding.Marshal(c))
+}
+
+func dbGetHistoricOutput(tx *bolt.Tx, id types.OutputID) (c types.Currency, err error) {
+	b := tx.Bucket(bucketHistoricOutputs).Get(encoding.Marshal(id))
+	err = encoding.Unmarshal(b, &c)
+	return
+}
```
DavidVorick: docstrings

modules/wallet/persist.go
```diff
@@ -8,12 +8,14 @@ import (
...
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/persist"
+
+	"github.com/NebulousLabs/bolt"
 )
 
 const (
-	logFile            = modules.WalletDir + ".log"
-	settingsFileSuffix = ".json"
-	settingsFile       = modules.WalletDir + settingsFileSuffix
```
DavidVorick: heh. I guess I went overboard lol

modules/wallet/persist.go
```diff
@@ -124,6 +154,13 @@ func (w *Wallet) initPersist() error {
 	if err != nil {
 		return err
 	}
+
+	// Open the database.
+	err = w.openDB(filepath.Join(w.persistDir, dbFile))
+	if err != nil {
+		return err
+	}
```
DavidVorick: while we're here, it might make sense to create a threadgroup to handling the closing of the db. Maybe out of scope for this PR though.

modules/wallet/seed_test.go
```diff
@@ -108,6 +108,8 @@ func TestLoadSeed(t *testing.T) {
 		t.Error("AllSeeds returned the wrong seed")
 	}
 
+	wt.wallet.db.Close() // db must be closed before reuse
```
DavidVorick: What's going on here? Need to expand the comment

DavidVorick: oh hmm. Is it just an artifact of the transition being incomplete?

DavidVorick:
Looks like a good start. One thing that has me worried is data corruption. Namely, there are two files in the wallet now, the db and the json, and they can desynchronize during a power failure. e.g. the db might update with the latest consensus change, but the json may not have finished writing to disk. Then the db will get the consensus change a second time.

Either we need to move all of the wallet persistence in to the database, or we need to create a write-ahead-log. The first is probably easier, WALs are pretty complicated.

modules/wallet/persist.go
lukechampine: wallet already has a threadgroup

lukechampine:
I think storing everything in the db is fine. It might feel a little weird to store single items in the database, but it's preferable to having two persist files.

modules/wallet/seed_test.go
lukechampine: it's because the test calls `New` in the same directory twice. Without the `Close`, you'd get a timeout error when trying to open the database.
I first tried `wt.wallet.Close`, but sometimes the test wants to compare one wallet to another, which it can't do after `Close` has been called.

modules/wallet/database.go
lukechampine: ok, but first I need to know what the historic outputs are actually used for :P

DavidVorick: When money is sent to you, it'll often come from an output that's not yours. Having the output does not tell you the value though, you must have a list of outputs -> values.

An alternative method here would be to have the cs provide that in the consensus change.

DavidVorick: Claim starts are the same thing, but they tell you the claim start of a siafund output.

lukechampine: thanks, added

DavidVorick:
Once we get there, I'd just have a single key "WalletPersist" or something that maps to the walletPersist struct. That's pretty easy, and doesn't require going to disk multiple times to write stuff.

I think where bolt loses performance is that often times it needs to go to disk multiple times to do a read or a write.

modules/wallet/persist.go
```diff
@@ -27,40 +29,45 @@ var (
 		Header:  "Wallet Seed",
 		Version: "0.4.0",
 	}
+
+	dbMetadata = persist.Metadata{
+		Header:  "Wallet Database",
+		Version: "1.0.0",
```
DavidVorick: 1.1.0

DavidVorick: We're going to be adding routes related to the wallet, namely some better restore-wallet routes, therefore the version needs to bump.

modules/wallet/persist.go
```diff
@@ -103,6 +110,29 @@ func (w *Wallet) initSettings() error {
...
+		}
+		for _, b := range buckets {
+			_, err := tx.CreateBucketIfNotExists(b)
+			if err != nil {
+				return err
+			}
+		}
+		return nil
+	})
+	return err
```
DavidVorick: This blob of code here is likely to grow. It's fine for now but should keep in mind what the structure will be turning into. Ideally, any boltdb code is written such that it's easy to reason about the performance impact.

DavidVorick:
1. Need unit tests for all the `dbPut` and `dbGet` functions that have been created.
2. Need to close the db via a threadgroup. The tests that currently close the db manually should erase the wallet and load a new one using the same persistence folder.
3. Need unit tests for `openDB`
4. Should be adding context to all errors, it's a lifesaver when looking at user logs

I'm also nervous about all the changes happening to update.go. Does the testing have any sanity checks or consistency checks? The fact that we don't even have an update_test.go is pretty worrying. I guess that's a problem for a different PR, but since we're changing code that is financially critical, testing is going to play a very big role in this branch.

modules/wallet/database_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"testing"
+
+	"github.com/NebulousLabs/Sia/types"
+
+	"github.com/NebulousLabs/bolt"
+)
+
+func TestDBHistoricHelpers(t *testing.T) {
```
DavidVorick: docstring.

Test looks good otherwise.

PR #1394, opened 2016-08-02
API Absolute Paths

### LINKED ISSUE #1385: Sia API Path Expansion Inconsistent
It looks as if the API does not support tilde expansion. This seems inconsistent given the API supports relative paths. Also paths are relative to `siad` which could be in an obscure location; does it make sense to allow relative paths at all?

If a user runs siac from a different directory than siad was launched from this could lead to confusion. This could be particularly true if a user tries to use a relative path in either Sia-UI's Terminal Plugin or with a siad instance started by the Sia-UI.

Below are examples of the current behavior.

### Endpoint - /wallet/load/033x

*Absolute Path* - Works
````
$ curl -sA Sia-Agent http://localhost:9980/wallet/033x --data "encryptionpassword=$SIA_PWD&source=/Users/John/Desktop/wallet.dat"  | python -m json.tool
{
    "message": "error when calling /wallet/033x: old wallet has no new seeds"
}
````

*Paths Relative to Siad* - Works
````
$ curl -sA Sia-Agent http://localhost:9980/wallet/033x --data "encryptionpassword=$SIA_PWD&source=../../../../Desktop/wallet.dat"  | python -m json.tool
{
    "message": "error when calling /wallet/033x: old wallet has no new seeds"
}
````

*Tilde Expansion* - **Fails**
````
$ curl -sA Sia-Agent http://localhost:9980/wallet/033x --data "encryptionpassword=$SIA_PWD&source=~/Desktop/wallet.dat"  | python -m json.tool
{
    "message": "error when calling /wallet/033x: open ~/Desktop/wallet.dat: no such file or directory"
}
````

### Endpoint -  /wallet/load/siagkey

Unfortunately I didn't have a siafund file to test with, so the following calls either found the file and crashed while reading it, or could not find the file.

*Absolute Path* - Works (Finds the file)
````
$ curl -sA Sia-Agent http://localhost:9980/wallet/siagkey --data "encryptionpassword=$SIA_PWD&keyfiles=/Users/John/Desktop/wallet.dat"  | python -m json.tool
{
    "message": "error when calling /wallet/siagkey: error while reading /Users/John/Desktop/wallet.dat: could not decode type wallet.SiagKeyPair: length 9728765351815533699 exceeds maxLen of 5000000"
}
````

*Paths Relative to Siad* - Works (Finds the file)
````
$ curl -sA Sia-Agent http://localhost:9980/wallet/siagkey --data "encryptionpassword=$SIA_PWD&keyfiles=../../../../Desktop/wallet.dat"  | python -m json.tool
{
    "message": "error when calling /wallet/siagkey: error while reading ../../../../Desktop/wallet.dat: could not decode type wallet.SiagKeyPair: length 9728765351815533699 exceeds maxLen of 5000000"
}
````

*Tilde Expansion* - **Fails** (Cannot find the file)
````
$ curl -sA Sia-Agent http://localhost:9980/wallet/siagkey --data "encryptionpassword=$SIA_PWD&keyfiles=~/Desktop/wallet.dat"  | python -m json.tool
{
    "message": "error when calling /wallet/siagkey: open ~/Desktop/wallet.dat: no such file or directory"
}
````

droghio:
Validates that file paths sent to the renter and wallet apis are absolute paths. Throws an error otherwise.

The host module has implemented this behavior since pr #1325. I am not sure if we should move this check to the api to or the wallet/host checks into the modules so in the mean time I have left the host api as is.

Fixes #1385.

### DIFF
--- a/api/renter.go
+++ b/api/renter.go
@@ -3,8 +3,8 @@ package api
 import (
 	"fmt"
 	"net/http"
-	"strings"
 	"path/filepath"
+	"strings"
 
 	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/modules"
@@ -161,8 +161,8 @@ func (srv *Server) renterDownloadsHandler(w http.ResponseWriter, _ *http.Request
 func (srv *Server) renterLoadHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
 	source := req.FormValue("source")
 	if !filepath.IsAbs(source) {
-	    writeError(w, Error{"source must be an absolute path"}, http.StatusBadRequest)
-	    return
+		writeError(w, Error{"source must be an absolute path"}, http.StatusBadRequest)
+		return
 	}
 
 	files, err := srv.renter.LoadSharedFiles(source)
@@ -222,8 +222,8 @@ func (srv *Server) renterDownloadHandler(w http.ResponseWriter, req *http.Reques
 	destination := req.FormValue("destination")
 	// Check that the destination path is absolute.
 	if !filepath.IsAbs(destination) {
-	    writeError(w, Error{"destination must be an absolute path"}, http.StatusBadRequest)
-	    return
+		writeError(w, Error{"destination must be an absolute path"}, http.StatusBadRequest)
+		return
 	}
 
 	err := srv.renter.Download(strings.TrimPrefix(ps.ByName("siapath"), "/"), destination)
@@ -241,8 +241,8 @@ func (srv *Server) renterShareHandler(w http.ResponseWriter, req *http.Request, 
 	destination := req.FormValue("destination")
 	// Check that the destination path is absolute.
 	if !filepath.IsAbs(destination) {
-	    writeError(w, Error{"destination must be an absolute path"}, http.StatusBadRequest)
-	    return
+		writeError(w, Error{"destination must be an absolute path"}, http.StatusBadRequest)
+		return
 	}
 
 	err := srv.renter.ShareFiles(strings.Split(req.FormValue("siapaths"), ","), destination)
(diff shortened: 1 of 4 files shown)


mnsl:
Maybe include tests, e.g.
```go
func TestRenterRelativePathError(t *testing.T) {
	if testing.Short() {
		t.SkipNow()
	}
	st, err := createServerTester("TestRenterRelativePathError")
	if err != nil {
		t.Fatal(err)
	}
	defer st.server.Close()

	// Announce the host.
	if err := st.announceHost(); err != nil {
		t.Fatal(err)
	}

	// Create a file in the present working directory.
	if err = createRandFile("test.dat", 1024); err != nil {
		t.Fatal(err)
	}

	// Upload to host.
	uploadValues := url.Values{}
	uploadValues.Set("source", "test.dat")
	err = st.stdPostAPI("/renter/upload/foo/bar/test", uploadValues)
        if err == nil || err.Error() != "source must be an absolute path" {
		t.Fatal(err)
	}
        // etc....
}

If this seems like too much of a hassle, I'm happy to do the rest

DavidVorick:
Testing needs to accompany all PRs, but wouldn't be difficult to add a commit from @mnsl to this branch

lukechampine:
unfortunately this branch is on drogo's fork, so mnsl can't push to it directly

droghio:
Makes sense, @mnsl gave a great example of a test and I'll work on extending it to cover all modified API routes. In the mean time I also gave @mnsl push permissions to my repo, just in case.

droghio:
Added tests and rebased to latest master. The hosts module still implements an absolute path check in the module instead of the API. Not sure if this should be moved/duplicated in the api or left as is, but I feel where we perform these checks should be consistent.

PR #1398, opened 2016-08-05
test host API calls

mnsl:
This PR tests that API calls to add, resize and remove host storage folders are handled properly, including various flavors of invalid calls.

Several errors from the `storagemanager` module are now exported, and I added a couple of exported functions so that the tests could access `storagemanager`'s `minimumStorageFolderSize`and `maximumStorageFolderSize` constants.

I also fixed some typos.

TODO:  
* a couple more tests to cover all error handling by the add/resize storage folder API calls
* test the API call to delete a specific storage sector

### DIFF
--- a/api/host_test.go
+++ b/api/host_test.go
@@ -419,8 +419,141 @@ func TestResizeFolderNoPath(t *testing.T) {
 	resizeValues := url.Values{}
 	resizeValues.Set("newsize", mediumSizeFolderString)
 	err = st.stdPostAPI("/host/storage/folders/resize", resizeValues)
-	if err == nil || err.Error() != "path parameter is required" {
-		t.Fatalf("expected error to be path parameter is required; got %v", err)
+	if err == nil || err.Error() != errNoPath.Error() {
+		t.Fatalf("expected error to be %v; got %v", errNoPath, err)
+	}
+}
+
+// TestRemoveEmptyStorageFolder checks that removing an empty storage folder
+// succeeds -- even if the host is left with zero storage folders.
+func TestRemoveEmptyStorageFolder(t *testing.T) {
+	if testing.Short() {
+		t.SkipNow()
+	}
+
+	st, err := createServerTester("TestRemoveEmptyStorageFolder")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer st.server.Close()
+
+	// Set up a storage folder for the host.
+	if err := st.setHostStorage(); err != nil {
+		t.Fatal(err)
+	}
+
+	// Try to delete the host's empty storage folder.
+	removeValues := url.Values{}
+	removeValues.Set("path", st.dir)
+	if err = st.stdPostAPI("/host/storage/folders/remove", removeValues); err != nil {
+		t.Fatal(err)
+	}
+}
+
+// TestRemoveStorageFolderError checks that invalid calls to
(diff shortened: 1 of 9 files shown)


api/host_test.go
```diff
@@ -1,5 +1,441 @@
...
+	// Announce the host and start accepting contracts.
+	if err := st.announceHost(); err != nil {
+		t.Fatal(err)
+	}
+	if err := st.acceptContracts(); err != nil {
+		t.Fatal(err)
+	}
+	if err := st.setHostStorage(); err != nil {
+		t.Fatal(err)
+	}
```
mnsl: Should a host be allowed to announce itself before it has a storage folder?

DavidVorick: hmm, not really. But I could see situations where you want the announcement on the chain before you'd actually be ready to receive calls.

An error here sort of makes sense, but I think that in terms of power and potential future problems, this is really a warning that should be handled by the front-end, not the back end. We don't want to predict too heavily how third parties will want to use the system.

api/host_test.go
```diff
@@ -1,5 +1,485 @@
...
+			}
+		}
+	}
+}
+
+// TestResizeNonemptyStorageFolder tests that invalid and valid calls to resize
+// a storage folder with one sector filled are properly handled.
+// Ideally, we would also test a very full storage folder (including the case
+// where the host tries to resize to a size smaller than the amount of data
+// in the folder), but that would be a very expensive test.
```
DavidVorick: If you have a storage folder that is twice as large as the minimum size allowed, it wouldn't be that expensive. The hardest part is probably that you need a renter to do the uploading, as there's no way in the API to upload data to the host without using a renter.

sectors during testing are set to 4kb.

api/host_test.go
```diff
@@ -1,5 +1,485 @@
...
+// has been set to "true".
+func TestForceRemoveStorageFolder(t *testing.T) {
+	if testing.Short() {
+		t.SkipNow()
+	}
+
+	st, err := createServerTester("TestForceRemoveStorageFolder")
+	if err != nil {
+		t.Fatal(err)
+	}
```
DavidVorick: need to close the server tester

api/host_test.go
```diff
@@ -1,5 +1,485 @@
...
+// /host/storage/folders/remove fail with the appropriate error.
+func TestRemoveStorageFolderError(t *testing.T) {
+	if testing.Short() {
+		t.SkipNow()
+	}
+
+	st, err := createServerTester("TestRemoveStorageFolderErr")
+	if err != nil {
+		t.Fatal(err)
+	}
```
DavidVorick: need to close the server tester

DavidVorick:
Overall these tests are very good. These are exactly the types of tests that I'm hoping to have cover all the modules. The logic and test types are cleanly separated, the code is pretty straightforward, and the underlying modules are probed using exclusively the API, which means that it's checking that both the modules are conforming to specification and that the API does not have any bugs.

mnsl:
Appveyor is reporting that `TestHostAutoRevisionSubmission` in the host module was killed because it ran for too long (10 min).

lukechampine:
seems like a NDF. I'll restart the build

DavidVorick:
This is happening in Jordan's PR as well.

Issue #1399, opened 2016-08-05
siad does not cleanly close each module


johnathanhowell:
Currently, signal handling is done in `api/server.go`, here : https://github.com/NebulousLabs/Sia/blob/master/api/server.go#L84.  Siad calls this function and expects it to block, returning when Serve() has completed. Therefore, server.Serve() returning means the program ends.  Because of this, srv.Close progresses to `srv.wg.Wait()`, and the program ends, failing to safely close each module.

PR #1400, opened 2016-08-05
wait for Close() to finish before returning from Serve()

johnathanhowell:
This is a hotfix for #1399.

### DIFF
--- a/api/server.go
+++ b/api/server.go
@@ -30,9 +30,7 @@ type Server struct {
 	listener          net.Listener
 	requiredUserAgent string
 
-	// wg is used to block Close() from returning until Serve() has finished. A
-	// WaitGroup is used instead of a chan struct{} so that Close() can be called
-	// without necessarily calling Serve() first.
+	// wg is used to block Serve() from returning until Close() has finished.
 	wg sync.WaitGroup
 }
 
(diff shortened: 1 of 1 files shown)


johnathanhowell:
Note that this PR reveals another issue which is that `gateway.Close()` hangs (deadlock?)

DavidVorick:
Do you have more details about the deadlock? Like a thread dump from the test suite? gateway.Close() is at risk of deadlock because there are places where it does nested calls to `tg.Add()`. Removing them should fix most of the deadlocks, but I'm not sure how much restructuring will be required.

api/server.go
```diff
@@ -93,6 +87,10 @@ func (srv *Server) Serve() error {
 	// closed, via either the Close method or the signal handling above.
 	// Closing the listener will result in the benign error handled below.
 	err := srv.apiServer.Serve(srv.listener)
+
+	// Wait for srv.Close to finish before returning.
+	srv.wg.Wait()
```
DavidVorick: You call `Wait()` before handling the error? Also, why is the waiting performed in the `Serve()` thread instead of the `Close()` thread?

I'm not following the shutdown design very well here. `Close()` needs to block until all processes have completed, which would include the `Serve()` process.

johnathanhowell:
I'm closing this PR in favor of fixing this bug properly in an upcoming PR which moves the server and signal handling to `siad`.

PR #1401, opened 2016-08-08
Replace the Host Storage Manager

DavidVorick:
After doing some benchmarks it became clear that speeds on the old storage manager were not acceptable. Unfortunately they were 1-2 orders of magnitude below acceptable, and in particular operations such as renewing a storage obligation were taking so long that it because unlikely such a process would be possible in the wild with any sort of large storage obligation.

We don't have any sort of stress testing right now, and even just a little bit would have revealed this problem.

I spent a few days learning everything there is to know about disks and I/O and memory and latency and bottlenecks, then took it upon myself to replace boltdb in the storage manager with a custom, hand-crafted database. This database makes a single disk operation, and does batching that matches acceptable latencies within the host.

The new contract manager is incomplete, as it doesn't actually store any data yet. Most of the optimizations haven't been implemented, because it  takes a lot of work to create a custom ACID database. But, the guts of the database are complete and extending it to serve new functions should be pretty easy.

I have implemented a single operation - AddStorageFolder, and then written a boatload of tests around AddStorageFolder to more or less demonstrate that the ACID transactions are functional. Included in this testing is some more mocking, including a new type of mocking I'll label 'disrupt mocking'. There's a disrupt() dependency that in production always returns `false`. But, the code has paths such that when it returns `true`, either logic is skipped or a function is terminated or somehow otherwise the code-flow of the program is disrupted.

That made it easy to do things like kill `AddStorageFolder` halfway through.

A big part of the diff is commented out code that I copy-pasted from the storagemanager to help guide the new functions - the old storagemanager had a bad database but extensive fault-tolerance, and when I was doing the replacement I wanted to make sure that I wasn't dropping any tolerance. I'll continue to remove the commented out code as I re-implement it.

All told, I've spent probably 80 hours on this so far. Hopefully it'll speed things up enough to make the investment pay off. I'm pretty sure there will be massive performance gains.


DavidVorick:
It should be noted that the new code isn't actually running, and isn't used at all by `siad`. That will come eventually.

It's also worth noting that compatibility with the previous storagemanager is completely broken, so there will need to be some compatibility code written to transition existing hosts to the new contract manager. That will also come eventually.

Might be worth releasing some RCs before the compatibility code is written, so that we can see how the hosts perform in the wild.

lukechampine:
For posterity, can you give a high-level overview of how the current storage manager works, and how the new manager will work?

DavidVorick:
The current storage manager keeps a list of sectors in a bolt database instance, storing them by a 12 byte key, and giving them a value indicating the number of virtual sectors represented by the physical sector. The sectors are stored in the filesystem by a filename. Also stored in the database is a list of the expiration heights of all the virtual sectors, this list grows in size every time a virtual sector is added, and shrinks every time a virtual sector is removed.

Updating a file contract revision right now requires opening up the bolt database and doing a bunch of Update actions. This is slow. It also doesn't scale well to larger hosts. Multiple people have raised concerns about the structure of the filesystem. Namely, a multi-terabyte host is going to end up with millions of files in a single directory, and if bolt is doing a binary search along the keys in a sparse database, we're looking at dozens of disk reads to do a single sector lookup. The sector metadata must be updated every time a sector is added or removed.

The new contract manager removes bolt from the equation and uses a custom ACID database instead. Once I started thinking in terms of ACID I realized that we actually have a lot of issues in the existing host (and likely in other modules as well) relating to consistency and durability. In general, isolation was not a design goal at all, which means that parallelism is unsafe (for the most part we handle that with locking - all major operations would hold a lock until an fsync had completed, which means that there was effectively no parallelism - lots of disk blocking all the time).

The majority of this pull request sets up a framework for an ACID database using a write-ahead-logger. It's worth checking out contractmanager.go and storagefolders.go as your first stops, but really the only important stuff there is the struct definitions. Keep them open in separate tabs while you dive into writeaheadlog.go, which is where the real juice is.

The WAL only currently supports a single operation, adding a storage folder. The hard part is applying ACID principles to the storage folder addition without impacting performance too heavily.  The general flow for using the WAL is as follows:

1. lock the WAL, assemble a consistent state, and based on the consistent state determine whether to continue or return an error (for example `errRepeatStorageFolder`). Assembling the consistent state needs to happen using the two parts of the WAL - the committed state and the uncommitted state. The committed state is held in the contract manager. If you have a lock on the WAL, and a lock on the contract manager, then you are guaranteed that all of the state in the contract manager has been successfully and permanently committed to disk, and will be  there after restart. The uncommitted state may or may not be committed all the way to disk - if there's no power interruption, it'll definitely be on disk. If there is a power interruption, maybe it won't be on disk, but at least you are guaranteed that if the uncommitted changes don't make it to disk, neither will the changes you are about to make.
2. append a change to the WAL. This tells the WAL 'commit this to disk next time you get a chance'. This uses the function `wal.appendChange`
3. Grab a channel (which tells you when the commit was successful) and then unlock the WAL, giving other threads a chance to suggest changes to the WAL.
4. When the channel closes, you know that your change has been permanently committed to disk.

When you have the WAL lock, nothing else can make changes to the WAL. So release it quickly! Ideally you only have it for 1 or 2 milliseconds, so that hundreds of changes can be made to the WAL every second. The WAL has a `threadedSyncLoop` which will occasionally commit all the pending changes together.

The sync loop can take a long time to commit, because the sync loop calls `fsync()` multiple times. Once after it does a COW write of the log to disk - it writes a bunch of changes to disk that say 'this is what I'm about to do', and then it executes those changes. Those changes may touch multiple files on multiple hard drives, so each of those files incurs an `fsync()`. It's okay though, because all of the syncing is done in parallel, which means it should take about as long as a single worst-case `fsync()`. Note: this parallelism is not actually implemented as of this PR because only one file is in use. That'll change as soon as AddSector is complete.

Now, it needs to be noted here that the changes which get added to the WAL should be idempotent. Ideally, you'd also be able to apply them in any order. The method for achieving this depends on the action being performed, but it basically comes down to converting a call like 'Add 1 to myVar'  to 'Set myVar to 12'. Then, if there are multiple calls (e.g. Set 12 and Set 13), you figure out which ones are redundant (the WAL has a specific ordering, so you do some preprocessing to remove all the redundant actions) and remove them. Then you can execute everything that remains in parallel. If you lose power, the WAL is still there, and so you just repeat the process. Some of the actions may have already been committed before power was lost, but because everything is idempotent, it's okay that the actions are potentially applied multiple times.

Some operations, such as 'AddStorageFolder', can take a long time to commit. Long running operations should definitely not hold a lock on the WAL, but they may affect other operations. For example, in the case of 'AddStorageFolder' you need to know whether or not a storage folder of the same name is in the middle of being added, or if the maximum number of storage folders allowed has been exceeded. So you add a note to the WAL/state that a storage folder addition is in progress, but the heavy lifting has not been completed. Because this long running process creates files, you want to be able to clean them up if it's interrupted (by power loss) before it commits all the way. So you end up waiting for two Sync() cycles before AddStorageFolder will return, but that's okay because it's a slow operation anyway, and you ensure that the parallelism of the WAL is not affected at all. (it's only blocking the WAL for very small periods of time while it does expensive operations).

There are a few other instances of performance back-bending here. For example, all changes given to the WAL start writing to the WAL file immediately. Sync() is faster if most of the data has already been written. The goal is to spend only a fraction of your time actually Syncing, because while waiting for a Sync all other actions on the WAL are blocked entirely. Writing immediately means that the call to Sync the WAL will execute faster.

But not that much faster, because you still have to commit everything after the WAL syncs, and that usually involves writing as much data as the WAL. Except, in the case of adding new storage folders, it can go much faster because the adds may be occurring on multiple disks, which means parallel writes.

There's another quirk where the WAL is strictly append-only. That means that if you commit to adding a storage folder, and then the storage folder fails (errNoDiskSpace or something), you can't un-commit to adding the storage folder. Instead, you commit to the fact that the storage folder addition failed.

I think that more or less explains everything. The real meat of this PR is wrtieaheadlog.go and storagefolderadd.go, and together they are about 750 lines of code, not anywhere near as scary as the diff indicates.

DavidVorick:
Just wanted to say again, outside of a wall of text, that the meat of this PR is writeaheadlog.go and storagefolderadd.go, which together are less than 800 lines of code, not so scary compared to the +3000 that this PR is claiming.

modules/host/contractmanager/consts.go
```diff
@@ -106,13 +112,13 @@ var (
...
 	// bitfield defines which sectors are available, and the bitfield must be
-	// constructed 1 uint32 at a time (4 bytes, 32 bits, or 32 sectors).
+	// constructed 1 uint64 at a time (8 bytes, 64 bits, or 64 sectors).
 	//
-	// This corresponds to a granularity of 32 MiB on the production network,
-	// which relative to the TiBs of storage that hosts are expected to
-	// provide, is a large amount of granularity.
-	storageFolderGranularity = uint64(32)
+	// This corresponds to a granularity of 256 MiB on the production network,
+	// which is a high granluarity relative the to the TiBs of storage that
```
lukechampine: typo

modules/host/contractmanager/dependencies.go
```diff
@@ -71,7 +71,9 @@ type (
 	// file implements all of the methods that can be called on an os.File.
 	file interface {
 		io.Closer
+		io.Reader
 		io.Writer
+		Seek(int64, int) (int64, error)
```
lukechampine: You could replace three of these with a `io.ReadWriteCloser` or a `io.ReadWriteSeeker`

modules/host/contractmanager/sectoradd.go
```diff
@@ -0,0 +1,216 @@
...
+		}
+	}
+	// If nothing was found by the end of the array, a wraparound is needed.
+	if i == len(usage) {
+		for i = 0; i < start; i++ {
+			if usage[i] != math.MaxUint64 {
+				break
+			}
+		}
+	}
```
lukechampine: these loops could be combined using a modulo, but that might incur unacceptable overhead

modules/host/contractmanager/sectoradd.go
```diff
@@ -0,0 +1,216 @@
...
+		i = i >> 4
+	}
+	return r + bval[i]
+}
+
+// randFreeSector will take a usage array and find a random free sector within
+// the usage array. If one is found, the bit will be flipped atomically in the
+// usage array to indicate that the sector is no longer available. The uint64
+// indicates the index of the sector within the usage array.
+func randFreeSector(usage []uint64) uint32 {
```
lukechampine: outdated comment? I don't see the bit being flipped anywhere

modules/host/contractmanager/sectoradd.go
```diff
@@ -0,0 +1,216 @@
...
+func (wal *writeAheadLog) commitAddSector(sa sectorAdd) {
+	sf := wal.cm.storageFolders[sa.Folder]
+	lookupTableSize := len(sf.Usage) * 64 * 16
+
+	// Write the sector metadata to disk.
+	writeData := make([]byte, 16)
+	copy(writeData, sa.ID[:])
+	binary.LittleEndian.PutUint16(writeData[12:14], sa.Count)
+	binary.LittleEndian.PutUint16(writeData[14:], sa.Folder)
+	_, err := sf.file.Seek(16*int64(sa.Index), 0)
```
lukechampine: `binary.Write` may be appropriate here:
```go
_, err := sf.file.Seek(16*int64(sa.Index), 0)
meta := struct{
    ID [12]byte
    Count uint16
    Folder uint16
}{sa.ID, sa.Count, sa.Folder}
err = binary.Write(sf.file, binary.LittleEndian, meta)
```
Most likely, you would define the `meta` type separately. Then you could use `binary.Read` to decode it.

You could also combine the two `uint16`s into one `uint64` and use the `encoding` package. Then you only waste 4 bytes. I'm not a big fan of bitpacking though

modules/host/contractmanager/sectoradd.go
```diff
@@ -0,0 +1,216 @@
...
+	// Write the sector to disk. The write only needs to happen if the count is
+	// equal to 1, otherwise the data can be assumed to already be there.
+	if sa.Count == 1 {
+		_, err = sf.file.Seek(int64(modules.SectorSize)*int64(sa.Index)+int64(lookupTableSize), 0)
+		if err != nil {
+			wal.cm.log.Println("ERROR: unable to seek to sector data when adding sector")
+			sf.FailedWrites += 1
+			return
+		}
+		_, err = sf.file.Write(sa.Data)
```
lukechampine: `os.File` supports `WriteAt`, which uses a different syscall (`pwrite`). It's only good for one-off writes, though; IIRC, `WriteAt` does not affect the write position of `Write`.

modules/host/contractmanager/storagefolderadd.go
```diff
@@ -6,9 +6,42 @@ import (
...
+	// additions identified by their path to the storage folder that is being
+	// added.
+	usfMap := make(map[uint16]*storageFolder)
+	for _, sc := range scs {
+		for _, sf := range sc.UnfinishedStorageFolderAdditions {
+			usfMap[sf.Index] = sf
+		}
+		for _, sf := range sc.StorageFolderAdditions {
+			delete(usfMap, sf.Index)
+		}
```
lukechampine: how does the ordering work here? Is looks like if you have these state changes:
```
scs[0] = {
    Unfinished: [1, 2, 3]
    Additions: [3, 4]
}
scs[1] = {
    Unfinished: [4, 5]
    Additions: [1]
}
```
Then `scs[1]` will delete addition 1, but `scs[0]` will not delete addition 4 (making the final state `[2, 5]`). That behavior seems reasonable to me, but I wanted to double-check. Is it not possible for `scs[0]` to reference an unfinished addition in `scs[1]`?

modules/host/contractmanager/storagefolderadd.go
```diff
@@ -148,43 +179,33 @@ func (wal *writeAheadLog) managedAddStorageFolder(sf *storageFolder) error {
...
 	}
-	_, err = file.Write(finalBytes)
+	_, err = sf.file.Write(finalBytes)
 	if err != nil {
 		return build.ExtendErr("could not allocate storage folder", err)
 	}
-	err = file.Sync()
+	err = sf.file.Sync()
 	if err != nil {
 		return build.ExtendErr("could not syncronize allocated storage folder", err)
```
lukechampine: synchronize

modules/host/contractmanager/writeaheadlog.go
```diff
@@ -137,170 +159,58 @@ func (wal *writeAheadLog) appendChange(sc stateChange) error {
...
-	defer file.Close()
-
+// recoverWAL will read a previous WAL and re-commit all of the changes inside,
+// restoring the program to consistency after an unclean shutdown.
+func (wal *writeAheadLog) recoverWAL(walFile file) error {
 	// Read the WAL metadata to make sure that the version is correct.
-	decoder := json.NewDecoder(file)
-	err = readWALMetadata(file, decoder)
+	decoder := json.NewDecoder(walFile)
+	err := readWALMetadata(decoder)
```
lukechampine: why create the decoder outside `readWALMetadata`?

modules/host/contractmanager/writeaheadlogsync.go
```diff
@@ -0,0 +1,281 @@
...
+		if err != nil {
+			wal.cm.log.Critical("ERROR: unable to sync the contract manager settings:", err)
+			panic("unable to sync contract manager settings, crashing to avoid data corruption")
+		}
+		err = wal.fileSettingsTmp.Close()
+		if err != nil {
+			wal.cm.log.Println("unable to close the temporary contract manager settings file:", err)
+		}
+		// COW the settings file.
+		err = os.Rename(tmpFilename, filename)
```
lukechampine: COW means sharing a resource by pointer initially, and transparently allocating a copy when the first modification is performed

modules/host/contractmanager/writeaheadlogsync.go
```diff
@@ -0,0 +1,281 @@
...
+	// log file will be created shortly.
+	err = os.Remove(walTmpName)
+	if err != nil {
+		wal.cm.log.Println("ERROR: unable to remove temporary write-ahead-log after successful commit-to-log:", err)
+	}
+	// Any module waiting for a synchronization has now had the changes
+	// permanently committed, as of the WAL COW operation. Signal that syncing
+	// is complete.
+	close(wal.syncChan)
+	wal.syncChan = make(chan struct{})
```
lukechampine: `sync.Cond` might be a better choice for repeated signalling

modules/host/contractmanager/writeaheadlogsync.go
```diff
@@ -0,0 +1,281 @@
...
+			close(syncLoopStopped)
+			return
+		case <-time.After(syncInterval):
+			// Commit all of the changes in the WAL to disk, and then apply the
+			// changes. Measure how long it takes to apply the changes, and use
+			// that to steer the amount of time that should be waited to the
+			// next sync.
+			wal.mu.Lock()
+			wal.commit()
+			wal.mu.Unlock()
```
lukechampine: `syncInterval` no longer seems to be modified

modules/host/contractmanager/sectoradd.go
DavidVorick: 2 uint16s is a uint32. A tight encoding is important to keep the file as small as possible - reduces seek and sync times

modules/host/contractmanager/writeaheadlogsync.go
DavidVorick: The lock is to protect `wal.commit()` lol, which does a ton of state modifications.

lukechampine: I was referring to the comment, which is outdated now

lukechampine:
needs rebase (which will probably fix appveyor failures as well)

modules/host/contractmanager/storagefolderadd.go
DavidVorick: If there's an addition, it'll appear in a strictly later scs element than the corresponding 'unfinished'. The state change array is like a set of transactions in a block - you process them in order and they can depend on parents, but parents can't depend on children.

DavidVorick: given that, I don't think your example makes sense.

modules/host/contractmanager/writeaheadlogsync.go
DavidVorick: oh. I have misused the term copy-on-write all over the documentation. Guess I have to clean that up.

crypto/rand.go
```diff
@@ -3,32 +3,106 @@ package crypto
...
+// systems, hashing out of a preset RNG pool will be substantially faster than
+// using crypto/rand, which relies on syscalls.
+func init() {
+	// Use 64 bytes in case the first 32 aren't completely random.
+	base := make([]byte, 64)
+	_, err := rand.Read(base)
+	if err != nil {
+		panic("unable to set entropy for the crypto package")
+	}
+	entropyBase = HashObject(base)
```
lukechampine: no need to encode this; just use `HashBytes`

crypto/rand.go
```diff
@@ -3,32 +3,106 @@ package crypto
...
+		// Fetch and update the counter values before executing the hash.
+		var newCounter, newCounter2 uint64
+		newCounter = atomic.AddUint64(&counter, 1)
+		if newCounter == 0 {
+			newCounter2 = atomic.AddUint64(&counter2, 1)
+		} else {
+			newCounter2 = atomic.LoadUint64(&counter2)
+		}
+		// Grab some entropy using the unique counter set.
+		entropy := HashAll(newCounter, newCounter2, entropyBase)
```
lukechampine: this will incur encoding overhead. I suggest preallocating an array of the proper size and using `binary` to encode the `uint64`s directly:
```go
func Read(b []byte) {
    var arena [16 + HashSize]byte // two uint64s plus a hash
    copy(arena[16:], entropyBase) // same for every iteration
    n := len(b)
    for i := 0; i < n; i += HashSize {
        c1, c2 := incrementCounters()
        binary.LittleEndian.PutUint64(c1, arena[:8])
        binary.LittleEndian.PutUint64(c2, arena[8:16])
        copy(b[i:], HashBytes(arena[:]))
    }
}
```

crypto/rand.go
```diff
@@ -3,32 +3,106 @@ package crypto
...
+
+		// Fill out 'b'.
+		copy(b[i:], entropy[:])
```
lukechampine: if you have `Read`, `RandBytes` is reduced to a convenience function:
```go
func RandBytes(n int) ([]byte, error) {
    b := make([]byte, n)
    Read(b)
    return n, nil
}
```

crypto/rand.go
```diff
@@ -3,32 +3,106 @@ package crypto
...
+	}
+	// Grab some entropy using the unique counter set.
+	entropy := HashAll(newCounter, newCounter2, entropyBase)
+
+	// Convert the first 24 bytes into a big.Int, then grab the modulus of n.
+	// 24 bytes means that there is at least 16 bytes of overflow, which means
+	// early numbers are favored by less than 1-in-2^128 - a cryptographically
+	// safe preference.
+	b := new(big.Int)
+	b.SetBytes(entropy[:24])
```
lukechampine: we should probably stick with `rand.Int`. We can simply seed it with our `crypto.Reader` instead of `rand.Reader`.

crypto/rand.go
```diff
@@ -3,32 +3,106 @@ package crypto
...
+	n := len(b)
+	for i := 0; i < n; i += HashSize {
+		// Fetch and update the counter values before executing the hash.
+		var newCounter, newCounter2 uint64
+		newCounter = atomic.AddUint64(&counter, 1)
+		if newCounter == 0 {
+			newCounter2 = atomic.AddUint64(&counter2, 1)
+		} else {
+			newCounter2 = atomic.LoadUint64(&counter2)
+		}
```
lukechampine: I'm not super confident in my ability to use `sync/atomic`, but this looks fine. Worst-case scenario, this might happen:
```
0. counter = ^uint64(0), counter2 = 0
1. thread 1 adds 1 to counter 1; its newCounter == 0
2. thread 2 adds 1 to counter 1; its newCounter == 1
3. thread 2 loads counter2; its newCounter2 == 0
4. thread 1 adds 1 to counter2; its newCounter2 == 1
```
in this scenario, thread 1 sees (0, 1) and thread 2 sees (1, 0). In other words, thread2 will reuse a previous value. However, this can only happen once per uint64-address-space, which we do not expect to exhaust. If we only hashed slices of length 1, we would still need to hash 18 exabytes of data before overflowing the uint64.

crypto/rand_bench_test.go
```diff
@@ -0,0 +1,90 @@
...
+// BenchmarkRandRead64K benchmarks the speed of rand.Read for larger slices.
+func BenchmarkRandRead64K(b *testing.B) {
+	buf := make([]byte, 64e3)
+	for i := 0; i < b.N; i++ {
+		_, err := rand.Read(buf)
+		if err != nil {
+			b.Fatal(err)
+		}
+	}
+}
```
lukechampine: Might be a good place to use sub-benchmarks:
```go
func BenchmarkRead(b *testing.B) {
	small := make([]byte, 32)
	large := make([]byte, 64e3)

	b.Run("read-32", func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			Read(small)
		}
	})

	b.Run("rand-read-32", func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			rand.Read(small)
		}
	})

	b.Run("read-64k", func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			Read(large)
		}
	})

	b.Run("rand-read-64k", func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			rand.Read(large)
		}
	})
}
```

DavidVorick: I did not realize this was possible.

modules/host/contractmanager/storagefoldershrink_bench_test.go
```diff
@@ -0,0 +1,38 @@
...
+// BenchmarkUnsetManySectors checks that unsetting a bunch of sectors
+// individually takes an acceptable amount of time. Millions should be possible
+// in well under a second.
+//
+// My laptop is doing 10e6 in 1.2 seconds. This is on the edge of too slow, but
+// overall it's close enough.
+func BenchmarkUnsetManySectors(b *testing.B) {
+	// Create a uint64 that has all of the bits set.
+	base := uint64(0)
+	base--
```
lukechampine: `^uint64(0)`?

PR #1407, opened 2016-08-10
move server to Siad, start siad server immediately, properly close all modules.

### LINKED ISSUE #1399: siad does not cleanly close each module

johnathanhowell:
This PR moves api.Server to the `siad` binary.  `package api` has been changed to export an `API` type, which fulfills the `http.Handler` interface.  `Server` in siad serves this handler, along with its own handler which controls `siad` (`/daemon` routes).  This PR fixes https://github.com/NebulousLabs/Sia/issues/1399, as it does a deferred close for each module.  It also opens the door for 'instant' server startup; future PRs can build off of this one and provide module loading statuses to API callers.

Fixes #1399


DavidVorick:
the test suite doesn't compile

johnathanhowell:
fixed, made a mistake when manually resolving some merge conflicts

VoidingWarranties:
Does `siad` hang when quitting with this PR because of the gateway deadlock?

api/api.go
```diff
@@ -115,120 +117,145 @@ func requirePassword(h httprouter.Handle, password string) httprouter.Handle {
...
+		host:     h,
+		miner:    m,
+		renter:   r,
+		tpool:    tp,
+		wallet:   w,
+	}
 
+	// Register API handlers
+	router := httprouter.New()
+	router.NotFound = http.HandlerFunc(api.unrecognizedCallHandler)
```
VoidingWarranties: Missing newline after this line.

johnathanhowell:
Yes.  It does eventually exit, but it takes ~30 seconds to 5 minutes from what I've observed.

VoidingWarranties:
If it does eventually exit then I doubt it's a deadlock. I would guess that it is calls to `net.DialTimeout` in `threadedNodeManager` and `Connect`. The current timeout is 2 minutes, maybe we could reduce it.

DavidVorick:
Rather than reduce the timeout, you should select on a channel:

```
select{
case <-shudownChan:
     // exit early
case <-thingyTimesOut:
    // close the connection
}
```

johnathanhowell:
note that closing yields (non-determinsitcally)

```
Caught stop signal, quitting...
Critical error: cannot write to the file after it has been closed
Please submit a bug report here: https://github.com/NebulousLabs/Sia/issues
```

VoidingWarranties:
>Rather than reduce the timeout, you should select on a channel:
>
>```
select{
case <-shudownChan:
     // exit early
case <-thingyTimesOut:
    // close the connection
}
```

I don't know of a way to do this with `net.DialTimeout`. We'd have to implement our own `DialTimeout` that exited early on `shutdownChan`.

**Comment by VoidingWarranties** [comment]
You didn't touch this with your PR, but `startDaemon` has the wrong docstring.

File: siad/daemon.go [hunk 2]
```diff
@@ -125,6 +139,7 @@ func startDaemon(config Config) (err error) {
 		if err != nil {
 			return err
 		}
+		defer g.Close()
```
VoidingWarranties: What was your reason for using multiple `defer`s instead of one `defer srv.Close()`?

johnathanhowell: `srv` does not do module initialization.  Closing should be handled by the initializer, in this case `siad`.

VoidingWarranties: Ah okay, I was confused because the implementation of `Server` in server_helpers_test.go does close the modules.

johnathanhowell: Yeah.  There's a separate type `Server` defined there, that does close the modules.  It's only used for testing now.

VoidingWarranties:
>note that closing yields (non-determinsitcally)
>
>```
Caught stop signal, quitting...
Critical error: cannot write to the file after it has been closed
Please submit a bug report here: https://github.com/NebulousLabs/Sia/issues
```

Is this error seen in testing or in release builds? Since the errors from each `module.Close()` are ignored by the defer I wonder where this is coming from.

**Comment by johnathanhowell** [comment]
I observed this in both testing and release builds.

**Comment by DavidVorick** [comment]
It's a `build.Critical` message you don't have to check the error for it to print. I thought we added line numbers / call stacks to our `build.Critical` calls but I guess we haven't done that yet. @johnathanhowell could you sneak that into this pull request? The call you want I believe is `runtime.Caller`, there's a decent example of it happening in https://github.com/NebulousLabs/Sia/blob/master/sync/lock.go

Speaking of which, we should not be ignoring the errors in the Close calls, and now we don't have to! We can add a logger to the daemon that doesn't close until all of the modules have finished closing. If the modules return an error, they can be logged in the logger.

**Comment by VoidingWarranties** [comment]
Oh right, it's a `build.Critical` from within a module. That makes sense. Line numbers or Luke's pretty print stack trace would be very helpful.

**Comment by johnathanhowell** [comment]
I snuck `debug.PrintStack` into build.Critical.

**Comment by lukechampine** [comment]
>I don't know of a way to do this with net.DialTimeout. We'd have to implement our own DialTimeout that exited early on shutdownChan.

Dialer has a `cancel` channel: https://golang.org/pkg/net/#Dialer
You would wait on the shutdown channel and then close the `cancel` channel. You'd have to be careful to stop waiting on the shutdown channel when the Dial actually completes, though. Tricky.

File: api/api.go [hunk 3]
```diff
@@ -115,120 +117,146 @@ func requirePassword(h httprouter.Handle, password string) httprouter.Handle {
...
+}
+
+// api.ServeHTTP implements the http.Handler interface.
+func (api *API) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	api.router.ServeHTTP(w, r)
+}
+
+// New creates a new Sia API from the provided modules.
+// The API will require authentication using HTTP basic auth for certain endpoints
+// if the suppliecd password is not the empty string.  Usernames are ignored for authentication.
```
lukechampine: wrap comments to 80 chars. There's a typo here too

api/api.go
```diff
@@ -115,120 +117,146 @@ func requirePassword(h httprouter.Handle, password string) httprouter.Handle {
...
-	uaRouter := requireUserAgent(router, srv.requiredUserAgent)
-	srv.apiServer = &http.Server{Handler: uaRouter}
+	// Apply UserAgent middleware and return the API
+	api.router = requireUserAgent(router, requiredUserAgent)
+	return api
 }
 
 // unrecognizedCallHandler handles calls to unknown pages (404).
-func (srv *Server) unrecognizedCallHandler(w http.ResponseWriter, req *http.Request) {
+func (api *API) unrecognizedCallHandler(w http.ResponseWriter, req *http.Request) {
```
lukechampine: this doesn't need to be a method of `API`

johnathanhowell:
@lukechampine thoughts on exporting `writeJSON`, `writeError`, `writeSuccess`?

johnathanhowell:
also, should the shutdown fixes be part of this PR?

api/server_helpers_test.go
```diff
@@ -24,6 +29,102 @@ import (
...
+func NewServer(APIaddr string, requiredUserAgent string, requiredPassword string, cs modules.ConsensusSet, e modules.Explorer, g modules.Gateway, h modules.Host, m modules.Miner, r modules.Renter, tp modules.TransactionPool, w modules.Wallet) (*Server, error) {
+	l, err := net.Listen("tcp", APIaddr)
+	if err != nil {
+		return nil, err
+	}
+
+	a := New(requiredUserAgent, requiredPassword, cs, e, g, h, m, r, tp, w)
+	serv := &http.Server{
+		Handler: a,
+	}
```
lukechampine: inline this in the `Server` struct for better readability (no `serv`/`srv` distinction)

lukechampine:
don't make tiny commits for each PR comment. Wait until I finish reviewing and then break all the changes into commits as needed. Otherwise travis and appveyor will be constantly rerunning

siac/daemoncmd.go
```diff
@@ -31,6 +29,13 @@ var (
 	}
 )
 
+type (
+	UpdateInfo struct {
+		Available bool   `json:"available"`
+		Version   string `json:"version"`
+	}
+)
```
lukechampine: don't need parens here, and the type doesn't need to be exported.

siad/server.go
```diff
@@ -302,3 +377,61 @@ func (srv *Server) daemonStopHandler(w http.ResponseWriter, _ *http.Request, _ h
...
+
+	// Create the Server
+	mux := http.NewServeMux()
+	siadServ := &Server{
+		mux:      mux,
+		listener: l,
+	}
+	siadServ.httpServer = &http.Server{
+		Handler: mux,
+	}
```
lukechampine: can be inlined

lukechampine:
ok, all set. Other comments:
`siac` should import `siad` types to decode its responses.
`api` should export its helper methods.

DavidVorick:
@lukechampine alternatively you can make 12 commits for 12 PR comments, just don't push them until you've got them all. Travis/AppVeyor will only run on the most recent commit

DavidVorick:
With regards to PrintStack, I think we want to print to stdout while PrintStack will print to Stderr.

The error that we are seeing seems to relate to the fact that `OnStop` calls get registered in gofuncs, with the assumption being that they will be able to run before the `AfterStop` funcs have run. But because they are set up in goroutines, there's no guarantee that they'll run before the `AfterStop` calls, especially since `managedForwardPort` can take a while to execute.

So... that means that there'll need to be some concurrency added somewhere to guarantee that the logger is still open before the upnp stuff commits. I don't think it'll be that hard, but I might want to delay it until my current PR goes through, as I think I've touched a moderate amount of host code.

siad/daemon.go
```diff
@@ -115,7 +118,19 @@ func startDaemon(config Config) (err error) {
...
 
-	// Create all of the modules.
+	// Create the server and start serving daemon routes immediately.
+	fmt.Printf("(0/%d) Loading siad...\n", len(config.Siad.Modules))
+	srv, err := NewServer(config.Siad.APIaddr, config.Siad.RequiredUserAgent, config.APIPassword)
+	if err != nil {
+		return err
+	}
+
+	serverrs := make(chan error)
```
DavidVorick: servErrs please

siad/daemon.go
```diff
@@ -202,31 +237,28 @@ func startDaemon(config Config) (err error) {
...
-		perm, err := crypto.Perm(len(modules.BootstrapPeers))
-		if err != nil {
-			return err
-		}
-		for _, i := range perm[:3] {
-			go g.Connect(modules.BootstrapPeers[i])
-		}
-	}
+	// connect the API to the server
+	srv.mux.Handle("/", a)
```
DavidVorick: ok, so as of this PR, only one handler is being managed by the mux? Which means that this PR doesn't implement the startup status stuff? (ability to at least expose over the API as much granularity as "gateway loaded... consensus loaded... etc., even if we don't have progress within each module)

DavidVorick: it's fine to leave that to another PR just trying to make sure I'm not missing anything.

siad/server.go
```diff
@@ -302,3 +312,61 @@ func (srv *Server) daemonStopHandler(w http.ResponseWriter, _ *http.Request, _ h
...
+	srv := &Server{
+		mux:      mux,
+		listener: l,
+		httpServer: &http.Server{
+			Handler: mux,
+		},
+	}
+
+	// Register siad routes
+	srv.mux.Handle("/daemon/", api.RequireUserAgent(srv.daemonHandler(requiredPassword), requiredUserAgent))
```
DavidVorick: oh jk apparently I did miss something

lukechampine:
needs rebase to fix merge conflicts.

It's a bummer that siac can't import the `UpdateInfo` type from siad. Maybe splitting the API like this wasn't the best idea. The alternative would be to pass the daemon object to `api.New` and have it handle the routes. That's ugly too, but for different reasons.
I guess we'll leave it for now.

johnathanhowell:
rebased

DavidVorick:
I stand behind splitting the server from the API, it's going to make other things in the future easier.

I think it's pretty important that the daemon, responsible for spinning up the modules, also be responsible for shutting them down. It also makes sense that the daemon would declare and handle its own API. And, we need some way to get the module startup statuses to the user before they have all finished starting, especially if there's a rebase.

The API package as it stands could not take a module until it had finished loading, because the API package takes interfaces. Seems much cleaner to leave it that way and expose an alternate API to provide startup details (namely, the daemon API. I'm not sure where it would be pulling its information from, but perhaps a status channel or something that you give to the modules when you call 'New').

VoidingWarranties:
>With regards to PrintStack, I think we want to print to stdout while PrintStack will print to Stderr.

But this is an error so shouldn't we be printing to stderr?

lukechampine:
definitely agree that splitting the `Server` object out of the API package was the right decision. The `/daemon` routes are just an unfortunate side effect. If we were starting from scratch, I'm not sure I would include routes like `/stop` or `/update` anyway. But the cat's out of the bag, so we have to work within the boundaries we've set.

api/api.go
```diff
@@ -108,152 +110,178 @@ func requirePassword(h httprouter.Handle, password string) httprouter.Handle {
...
-	router.GET("/daemon/stop", requirePassword(srv.daemonStopHandler, password))
+	router http.Handler
+}
+
+// api.ServeHTTP implements the http.Handler interface.
+func (api *API) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	api.router.ServeHTTP(w, r)
+}
+
+// New creates a new Sia APi from the provided modules.  The API will require
```
mnsl: APi -> API

build/critical.go
```diff
@@ -13,4 +14,5 @@ func Critical(v ...interface{}) {
 		panic(s)
 	}
 	os.Stderr.WriteString(s)
+	debug.PrintStack()
```
mnsl: Are we keeping this in or was it just for testing this PR?  Probably useful generally

DavidVorick: We want to keep it but not using `PrintStack`, as that goes to stderr and I think stdout is the proper place for the output.

VoidingWarranties: @DavidVorick can you elaborate more on your preference for stdout vs. stderr?

lukechampine:
Re: dial timeouts, Go 1.7 has added a [`DialContext`](https://golang.org/pkg/net/#Dialer.DialContext) function, which would be more idiomatic (although I don't quite grok contexts yet)

build/critical.go
DavidVorick: @VoidingWarranties I am under the impression that the message is most likely to get to the user's eyes if we print it to stdout instead of stderr. If `Critical` is being called, it's due to a bug in the code and we need the highest chance possible of realizing that it has been triggered.

lukechampine: did you know that `println` prints to stderr?
stderr is the proper place for error output. The streams are merged on the terminal. In some cases, stderr might be _more_ likely to be see, e.g. when piping stdout to a file, the stderr output will not be piped and instead displayed in the terminal.

DavidVorick: hmm. If that's the case, then `PrintStack` is probably okay.

PR #1408, opened 2016-08-10
Add more fields to wallet

lukechampine:
Add siafundOutputs, siacoinOutputs, and spentOutputs to the database.

Also improve unlock speed by determining transaction relevancy first.

### DIFF
--- a/modules/wallet/update.go
+++ b/modules/wallet/update.go
@@ -13,7 +13,7 @@ import (
 
 // updateConfirmedSet uses a consensus change to update the confirmed set of
 // outputs as understood by the wallet.
-func (w *Wallet) updateConfirmedSet(cc modules.ConsensusChange) {
+func (w *Wallet) updateConfirmedSet(tx *bolt.Tx, cc modules.ConsensusChange) error {
 	for _, diff := range cc.SiacoinOutputDiffs {
 		// Verify that the diff is relevant to the wallet.
 		_, exists := w.keys[diff.SiacoinOutput.UnlockHash]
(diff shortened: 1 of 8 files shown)


modules/wallet/update.go
```diff
@@ -4,54 +4,52 @@ import (
...
 
-	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/types"
 
 	"github.com/NebulousLabs/bolt"
 )
 
+// isWalletAddress is a helper function that an UnlockHash is derived from
+// one of the wallet's spendable keys.
```
DavidVorick: confusing wording, I think you missed a word

DavidVorick:
So far, so good. What work remains in getting instant wallet unlocking?

DavidVorick:
Seems like you just need to be storing the processedTransactions? (next PR of course)

lukechampine:
The fields in the current wallet persistence need to be moved over as well. And I need to write compatibility code for loading that persistence file into the db.

There's also the question of the seeds/keys: for security reasons, we can't store the keys in the database, correct? (Or could they be stored in encrypted form?)

DavidVorick:
The seeds/keys is something that we can tackle in the next step. They might need to go in the database for the purposes of atomicity, but you'd just store them the way they are currently being stored on disk.

I'm thinking though that the keys should remain in-memory, because going to disk is really expensive and even if you've got 5000 keys they aren't going to consume that much RAM.

lukechampine:
each `spendableKey` is going to be ~100 bytes, so 5000 keys would be <500kb. I worry about users who have many more keys than that, though.

PR #1409, opened 2016-08-15
Update Miner API docs

VoidingWarranties:
Only significant non-styling difference is the addition of the visual depiction of the header encoding in /doc/api/Miner.md#byte-response

### DIFF
--- a/doc/api/Miner.md
+++ b/doc/api/Miner.md
@@ -84,10 +84,12 @@ possible nonces result in a header with a hash less than the target, call
 root. The above process can then be repeated for the new block header.
 
 The other fields can generally be ignored. The parent block ID field is the
-hash of the parent block's header. The timestamp is the time at which the block
-was mined. The merkle root is the merkle root of a merkle tree consisting of
-the timestamp, the miner outputs (one leaf per payout), and the transactions
-(one leaf per transaction).
+hash of the parent block's header. Modifying this field will result in an
+orphan block. The timestamp is the time at which the block was mined and is set
+by the Sia Daemon. Modifying this field can result in invalid block. The merkle
+root is the merkle root of a merkle tree consisting of the timestamp, the miner
+outputs (one leaf per payout), and the transactions (one leaf per transaction).
+Modifying this field will result in an invalid block.
 
 | Field           | Byte range within response | Byte range within header |
 | --------------- | -------------------------- | ------------------------ |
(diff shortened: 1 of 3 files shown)


doc/api/Miner.md
```diff
@@ -0,0 +1,108 @@
...
+0-31: target
+
+32-111: header
+
+```
+xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx (returned bytes)
+tttttttttttttttttttttttttttttttt (target)
+                                hhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh (header)
+                                                                nnnnnnnn (nonce)
+```
```
lukechampine: I feel like there's a standard way of displaying something like this, e.g. https://www.google.com/search?q=tcp+header&tbm=isch

I haven't been able to identify the name of such a specification, though

doc/api/Miner.md
```diff
@@ -0,0 +1,108 @@
...
+
```
DavidVorick: The 'Byte Response' needs to be more thorough. Notably, it should explain that a miner should be hashing the header, comparing the result to the target, and then submitting the header if the result is lower than the target. If the result is NOT lower than the target, the nonce should be increased / changed and the hash should be tried again. The nonce is only 64bits, which means sometimes you may flip through the entire nonce (this would take 2^64 attempts) without finding a block. If that happens, another call to /miner/header [GET] must be made. The new header returned by the miner will have a different Merkle root, which means that it will have a different hashing result than the previous header, even for the same nonce.

It should also be explicitly mentioned that the nonce is bytes [32-40] within the header, or [64-72] overall.

Because we are working with raw binary, it may make sense to go ahead and define the location and function of the timestamp and MerkleRoot as well. While a miner/developer doesn't technically need to know these values to be able to mine effectively, more complete knowledge usually means that you can develop in more comfort, and it also means that you are less likely to monkey with something that you shouldn't be touching. (there's less of a sense of "I wonder what will happen..." because you already know what will happen)

doc/api/Miner.md
DavidVorick: I find this one to be pretty readable:

https://3.bp.blogspot.com/_DZhhZp9of5I/TGz5YA5sLDI/AAAAAAAAC1k/9SESBl_yv6Y/s1600/CIE-L006-TCP-Header.png

doc/api/Miner.md
VoidingWarranties: > I feel like there's a standard way of displaying something like this, e.g. https://www.google.com/search?q=tcp+header&tbm=isch
>
>I haven't been able to identify the name of such a specification, though

GitHub's markdown unfortunately doesn't support entries spanning multiple columns. Alternatively we could write the raw HTML, but then the raw text file wouldn't be very readable.

VoidingWarranties: @DavidVorick 

>The 'Byte Response' needs to be more thorough.

:+1:

>It should also be explicitly mentioned that the nonce is bytes [32-40] within the header, or [64-72] overall.

This is already mentioned (except that it's [32-39] and [64-71])

>Because we are working with raw binary, it may make sense to go ahead and define the location and function of the timestamp and MerkleRoot as well.

:+1:

lukechampine: you could also create the table in ASCII (using ```). This is the approach used in [RFCs](https://tools.ietf.org/html/rfc793#page-15).

VoidingWarranties:
I added field descriptions and added an ASCII table for visualizing the byte layout. I left in the old ASCII visualization for comparison. Thew new style is over 100 lines longer but is more readable. What do you think?

doc/api/Miner.md
```diff
@@ -0,0 +1,133 @@
...
+header's bytes, and comparing the resulting hash to the target. The block with
+that nonce is valid if the hash is less than the target. If none of the 2^64
+possible nonces result in a header with a hash less than the target, call
+`/miner/header [GET]` again to get a new block header with a different merkle
+root. The above process can then be repeated for the new block header.
+
+The other fields can generally be ignored. The parent block ID field is the
+hash of the parent block's header. The timestamp is the time at which the block
+was mined. The merkle root is the merkle root of a merkle tree consisting of
+the timestamp, the miner outputs (one leaf per payout), and the transactions
```
DavidVorick: merkle tree does not contain the timestamp, only the miner payouts and the transactions:

```go

// MerkleRoot calculates the Merkle root of a Block. The leaves of the Merkle
// tree are composed of the Timestamp, the miner outputs (one leaf per
// payout), and the transactions (one leaf per transaction).
func (b Block) MerkleRoot() crypto.Hash {
	tree := crypto.NewTree()
	for _, payout := range b.MinerPayouts {
		tree.PushObject(payout)
	}
	for _, txn := range b.Transactions {
		tree.PushObject(txn)
	}
	return tree.Root()
}
```

doc/api/Miner.md
```diff
@@ -0,0 +1,133 @@
...
+Blocks are mined by repeatedly changing the nonce of the header, hashing the
+header's bytes, and comparing the resulting hash to the target. The block with
+that nonce is valid if the hash is less than the target. If none of the 2^64
+possible nonces result in a header with a hash less than the target, call
+`/miner/header [GET]` again to get a new block header with a different merkle
+root. The above process can then be repeated for the new block header.
+
+The other fields can generally be ignored. The parent block ID field is the
+hash of the parent block's header. The timestamp is the time at which the block
+was mined. The merkle root is the merkle root of a merkle tree consisting of
```
DavidVorick: You should mention that the daemon will pick a correct timestamp, and that modifying the timestamp can result in invalid blocks

doc/api/Miner.md
```diff
@@ -0,0 +1,133 @@
...
+
+provides a block header that is ready to be grinded on for work.
+
+###### Byte Response
+
+For efficiency the header for work is returned as a raw byte encoding of the
+header, rather than encoded to JSON.
+
+Blocks are mined by repeatedly changing the nonce of the header, hashing the
+header's bytes, and comparing the resulting hash to the target. The block with
```
DavidVorick: should we describe what the comparison should look like? (e.g. describe an algorithm for comparing them? Or is that too much hand holding?)

doc/api/Miner.md
```diff
@@ -0,0 +1,133 @@
...
+that nonce is valid if the hash is less than the target. If none of the 2^64
+possible nonces result in a header with a hash less than the target, call
+`/miner/header [GET]` again to get a new block header with a different merkle
+root. The above process can then be repeated for the new block header.
+
+The other fields can generally be ignored. The parent block ID field is the
+hash of the parent block's header. The timestamp is the time at which the block
+was mined. The merkle root is the merkle root of a merkle tree consisting of
+the timestamp, the miner outputs (one leaf per payout), and the transactions
+(one leaf per transaction).
```
DavidVorick: Might also be worth mentioning that modifying the Merkle root will definitely result in an invalid block.

DavidVorick:
The way the ascii table is presented, I prefer the old style. If following the RFC style, each row is supposed to represent new bytes I believe. So you'd have 4 rows of 32 bytes instead of 2 rows of 112 bytes. But in representing it that way you don't have a good way of distinguishing between the header and the target, you're just slapping them on top of eachother.

I think for this one I will lean towards the first style (e.g. tttttttttttthhhhhhhhhhhh etc)

doc/api/Miner.md
VoidingWarranties: Ah okay, I was going off of the docstring of the method.

DavidVorick: Oh. Do you mind fixing the docstring and including that in this PR?

doc/api/Miner.md
VoidingWarranties: It would be nice information to have but seems like a bit of hand holding and a little out of place in the API docs.

VoidingWarranties:
Updated with @DavidVorick's suggestions.

PR #1411, opened 2016-08-15
Use {En,De}codeAll in {M,Unm}arshalAll, respectively

VoidingWarranties:
Thank you @lukechampine for clarifying this for me.

### DIFF
--- a/encoding/marshal.go
+++ b/encoding/marshal.go
@@ -165,10 +165,9 @@ func Marshal(v interface{}) []byte {
 func MarshalAll(vs ...interface{}) []byte {
 	b := new(bytes.Buffer)
 	enc := NewEncoder(b)
-	// can't encode with EncodeAll (type information is lost)
-	for _, v := range vs {
-		enc.Encode(v)
-	}
+	// Error from EncodeAll is ignored as encoding cannot fail when writing
+	// to a bytes.Buffer.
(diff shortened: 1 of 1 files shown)


encoding/marshal.go
```diff
@@ -165,10 +165,9 @@ func Marshal(v interface{}) []byte {
 func MarshalAll(vs ...interface{}) []byte {
 	b := new(bytes.Buffer)
 	enc := NewEncoder(b)
-	// can't encode with EncodeAll (type information is lost)
-	for _, v := range vs {
-		enc.Encode(v)
-	}
+	// Error from EncodeAll is ignored as encoding cannot fail when writing
+	// to a bytes.Buffer.
+	_ = enc.EncodeAll(vs...)
```
VoidingWarranties: The blank identifier is used so that errcheck won't throw a warning.

DavidVorick:
Can you write a benchmark around this? I want to be sure that this code is not a speed regression.

DavidVorick:
```
--- SKIP: TestUpdate (0.44s)

	daemon_test.go:45: Failed to fetch latest release: API call to https://api.github.com/repos/NebulousLabs/Sia/releases/latest is returning an empty response
```

@mnsl did we ever find a solution for this? ^

DavidVorick:
Oh nvm I see that the test was skipped/passed

PR #1412, opened 2016-08-16
Shorter HashAll

lukechampine:
inspired by #1411 :wink:


DavidVorick:
man... seems like the host is not having a good time with appveyor. Looks like a deadlock exists somewhere, this happens often enough.

lukechampine:
probably ok to merge as long as one CI passes and the failure is clearly not related to the PR

PR #1414, opened 2016-08-16
Host logging panic

DavidVorick:
This PR was ultimately inspired by a fix which could have been 5 lines, but while in the code I found some stuff which needed cleanup. And then I felt compelled to write a regression test.

Regression test is passing, shutdown for the host tester is cleaner, and some of the files are now sorted alphabetically.

### DIFF
--- a/modules/host/dependencies.go
+++ b/modules/host/dependencies.go
@@ -30,6 +30,12 @@ var (
 type (
 	// dependencies defines all of the dependencies of the Host.
 	dependencies interface {
+		// disrupt can be inserted in the code as a way to inject problems,
+		// such as a network call that take 10 minutes or a disk write that
+		// never completes. distrupt will return true if the disruption is
+		// forcibly triggered. In production, disrupt will always return false.
+		disrupt(string) bool
+
 		// listen gives the host the ability to receive incoming connections.
 		listen(string, string) (net.Listener, error)
 
@@ -70,6 +76,12 @@ type (
 	// dependencies using full featured libraries.
 	productionDependencies struct{}
 )
+
+// disrupt will always return false, but can be over-written during testing to
+// trigger disruptions.
+func (productionDependencies) disrupt(string) bool {
+	return false
+}
 
 // listen gives the host the ability to receive incoming connections.
 func (productionDependencies) listen(s1, s2 string) (net.Listener, error) {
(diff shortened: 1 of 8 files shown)


DavidVorick:
*

The host was panicking sometimes because shutdown would trigger before the gothread managing port forwarding had properly initialized. The fix is to have the port management thread call `h.tg.Add`

modules/host/host_test.go
```diff
@@ -398,6 +406,10 @@ func TestSetAndGetInternalSettings(t *testing.T) {
 	if rebootSettings.NetAddress != settings.NetAddress {
 		t.Error("settings retrieval did not return updated value")
 	}
+
+	// Set ht.host to 'h' so that the 'ht.Close()' method will close everything
```
mnsl: 'h' -> 'rebootHost' ?

modules/host/dependencies.go
```diff
@@ -30,6 +30,12 @@ var (
 type (
 	// dependencies defines all of the dependencies of the Host.
 	dependencies interface {
+		// disrupt can be inserted in the code as a way to inject problems,
+		// such as a network call that take 10 minutes or a disk write that
+		// never completes. distrupt will return true if the disruption is
```
mnsl: distrupt -> disrupt

modules/host/network_test.go
```diff
@@ -1,5 +1,61 @@
 package host
 
+import (
+	"testing"
+	"time"
+)
+
+// blockingPortForward is a depedency set that causes the host port froward
```
mnsl: froward -> forward

modules/host/network_test.go
```diff
@@ -1,5 +1,61 @@
...
+	// thread control regarding upnp and shutdown.
+	err = ht.Close()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// The trailing sleep is needed to catch the previously existing error
+	// where the host was not shutting down correctly. Currently, the extra
+	// sleep does nothing, but in the regression a logging panic would occur.
+	time.Sleep(time.Second * 12)
```
lukechampine: does it need to be that long? Why not 7 seconds / 5 seconds?

DavidVorick: It needs to outlive the `disrupt` action that causes 10 seconds of sleeping. The `disrupt` action needs live until the host has had enough time to shutdown entirely. I can probably reduce each though.

PR #1415, opened 2016-08-16
add tracing to failing test

DavidVorick:
There is a host test that is timing out frequently in appveyor. On my machine the test takes 1.5 seconds, and I can't repro the issue.

This PR currently doesn't fix anything, but it will give me better insight into what's going wrong in the appveyor build.

I suspect that this bug is causing problems in production.

### DIFF
--- a/modules/renter/hostdb/scan.go
+++ b/modules/renter/hostdb/scan.go
@@ -136,10 +136,16 @@ func (hdb *HostDB) threadedProbeHosts() {
 	for hostEntry := range hdb.scanPool {
 		// Request settings from the queued host entry.
 		// TODO: use dialer.Cancel to shutdown quickly
-		hdb.log.Debugln("Scanning", hostEntry.NetAddress, hostEntry.PublicKey)
+		//
+		// A readlock is necessary when viewing the elements of the host entry.
+		hdb.mu.RLock()
+		netAddr := hostEntry.NetAddress
+		pubKey := hostEntry.PublicKey
+		hdb.mu.RUnlock()
+		hdb.log.Debugln("Scanning", netAddr, pubKey)
 		var settings modules.HostExternalSettings
 		err := func() error {
-			conn, err := hdb.dialer.DialTimeout(hostEntry.NetAddress, hostRequestTimeout)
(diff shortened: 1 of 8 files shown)


DavidVorick:
I believe I have resolved the issue, which related to jobs running out of order. The first commit actually resolves a separate ndf. This PR should be ready to merge.

There is a separate issue that I found related to storage obligation management. That one cuts a bit deeper and will need a separate, larger PR. Not as large as the contractmanager PR, but it will be a similar type of overhaul. The existing code for making sure that host obligations make it to the blockchain is pretty awful.

modules/renter/hostdb/scan.go
```diff
@@ -136,10 +136,16 @@ func (hdb *HostDB) threadedProbeHosts() {
 	for hostEntry := range hdb.scanPool {
 		// Request settings from the queued host entry.
 		// TODO: use dialer.Cancel to shutdown quickly
-		hdb.log.Debugln("Scanning", hostEntry.NetAddress, hostEntry.PublicKey)
+		//
+		// A readlock is necessary when viewing the elements of the host entry.
+		hdb.mu.RLock()
+		netAddr := hostEntry.NetAddress
+		pubKey := hostEntry.PublicKey
+		hdb.mu.RUnlock()
```
lukechampine: that's definitely weird... `scanPool` should probably send a `hostEntry` rather than a `*hostEntry` so you don't need to worry about it being modified. Why would an entry's netaddress or pubkey change though?

DavidVorick: scanPool needs a *hostEntry so that it can modify the fields lol. We could add more safety with a getter and a setter, but I don't want to do too much of that without remodeling the entire concurrency scheme of the hostdb.

The hostdb is one of our earliest forays into high-concurrency (having multiple threads running all the time that all modify the state). There are a lot of mistakes and warts that are evidence of our inexperience.

DavidVorick: The netaddress is allowed to change, a host can change IP addresses or domain names, as the renter is supposed to recognize the host by its pubkey.

Currently the renter does not, but this is a preparation for the day that it does.

PR #1416, opened 2016-08-16
persist.Save: use Encoder.SetIndent

lukechampine:
This requires Go 1.7, so we can hold off on merging until people have updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ files, and advertise yourself as a host.
 Building From Source
 --------------------
 
-To build from source, [Go 1.6 must be installed](https://golang.org/doc/install)
+To build from source, [Go 1.7 must be installed](https://golang.org/doc/install)
 on the system. Then simply use `go get`:
 
 ```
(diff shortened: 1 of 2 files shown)


DavidVorick:
What is the point of this?

lukechampine:
prevents marshalling the entire structure in memory. So, a slight optimization for large persist files. But it's probably too early to be forcing people to update to 1.7

VoidingWarranties:
Given how IO heavy Sia is this could have a huge impact. It would be cool to see a benchmark of this.

Can we add this to a future Sia milestone (perhaps [v1.1.0](https://github.com/NebulousLabs/Sia/milestone/8))?

DavidVorick:
Any benchmark demonstrating a significant speedup in I/O will prioritized heavily.

That said, I don't suspect that this would cause much of a speedup.

lukechampine:
agreed. It doesn't reduce the amount of I/O done, only the amount of memory used.

Issue #1395, opened 2016-08-02
Renter API Docs Unclear on Sia File Commands


droghio:
Based on the renter api docs the sia file calls look available. Since these are currently disabled in the router it might not be a bad idea to mark them as currently unavailable in the docs.

Relevant Endpoints:
```
POST /renter/load
POST /renter/loadascii
GET /renter/share
GET /renter/shareascii
```

PR #1417, opened 2016-08-17
Update Renter API docs

VoidingWarranties:


doc/API.md
```diff
@@ -711,253 +711,190 @@ The input byte array should be 80 bytes that form the solved block header. *Unli
...
+| ------------------------------------------------------------- | --------- |
+| [/renter](#renter-get)                                        | GET       |
+| [/renter](#renter-post)                                       | POST      |
+| [/renter/contracts](#rentercontracts-get)                     | GET       |
+| [/renter/downloads](#renterdownloads-get)                     | GET       |
+| [/renter/files](#renterfiles-get)                             | GET       |
+| [/renter/delete/___*siapath___](#renterdeletesiapath-post)    | POST      |
+| [/renter/download/___*siapath___](#renterdownloadsiapath-get) | GET       |
+| [/renter/rename/___*siapath___](#renterrenamesiapath-post)    | POST      |
+| [/renter/upload/___*siapath___](#renteruploadsiapath-post)    | POST      |
```
VoidingWarranties: Should routes be in alphabetical order or grouped by similarity?

doc/API.md
```diff
@@ -711,253 +711,190 @@ The input byte array should be 80 bytes that form the solved block header. *Unli
...
+###### Response
+standard success or error response. See
+[API.md#standard-responses](/doc/API.md#standard-responses).
 
-Function: Lists all files in the download queue.
+#### /renter/contracts [GET]
 
-Parameters: none
+// TODO: is this description correct?
+returns all contracts that have been formed by the renter.
```
VoidingWarranties: Does this route return all contracts that have ever been formed by the renter or only currently used & open contracts?

doc/API.md
```diff
@@ -711,253 +711,190 @@ The input byte array should be 80 bytes that form the solved block header. *Unli
...
+###### JSON Response [(with comments)](/doc/api/Renter.md#json-response-1)
+```javascript
+{
+  "contracts": [
+    {
+      "endheight":   50000, // block height
+      "id":          "1234567890abcdef0123456789abcdef0123456789abcdef0123456789abcdef",
+      "netaddress":  "12.34.56.78:9",
+      "renterfunds": "1234", // hastings
+      "size":        8192    // bytes TODO: is this the right unit?
```
VoidingWarranties: Bytes?

doc/API.md
```diff
@@ -711,253 +711,190 @@ The input byte array should be 80 bytes that form the solved block header. *Unli
...
-		uploadprogress float64
-		expiration     types.BlockHeight (uint64)
-	}
+###### JSON Response [(with comments)](/doc/api/Renter.md#json-response-2)
+```javascript
+{
+  "downloads": [
+    {
+      "siapath":     "foo/bar.txt",
+      "destination": "/home/users/alice",
```
VoidingWarranties: Should be "/home/users/alice/bar.txt"

doc/API.md
```diff
@@ -711,253 +711,190 @@ The input byte array should be 80 bytes that form the solved block header. *Unli
...
+      "filesize":    8192,                  // bytes
+      "received":    4096,                  // bytes
+      "starttime":   "2009-11-10T23:00:00Z" // RFC 3339 time TODO: why is this not a types.Timestamp?
```
VoidingWarranties: This is more of a question for when we get around to API V2... Was there a reason to use `time.Time` instead of `types.Timestamp` which is used more commonly in the rest of the API? `types.Timestamp`s are most often used in blocks and transactions, which importantly need to all have the same timestamp regardless of timezone.

doc/API.md
```diff
@@ -711,253 +711,190 @@ The input byte array should be 80 bytes that form the solved block header. *Unli
...
+###### Response
+standard success or error response. See
+[API.md#standard-responses](/doc/API.md#standard-responses).
 
-#### /renter/share [GET]
+#### /renter/download/___*siapath___ [GET]
 
-Function: Create a .sia file that can be shared with other people.
+// TODO: does this call still block until the file has been downloaded?
+downloads a file to the local filesystem.
```
VoidingWarranties: I vaguely remember this behavior changing recently, what did we decide on?

doc/API.md
```diff
@@ -711,253 +711,190 @@ The input byte array should be 80 bytes that form the solved block header. *Unli
...
+[API.md#standard-responses](/doc/API.md#standard-responses).
 
-#### /renter/delete/{siapath} [POST]
+#### /renter/rename/___*siapath___ [POST]
 
-Function: Deletes a renter file entry. Does not delete any downloads or
-original files, only the entry in the renter.
+renames a file. Does not rename any downloads or source files, only renames the
+entry in the renter.
+// TODO: what happens if newsiapath already exists? What happens if siapath doesn't exist?
```
VoidingWarranties: Not sure what the current behavior is, but it makes sense both of these should return errors if they don't already.

doc/api/Renter.md
```diff
@@ -0,0 +1,277 @@
...
+      "funds": "1234", // hastings
+
+      // Number of hosts that contracts will be formed with.
+      "hosts":24,
+
+      // Duration of contracts formed, in number of blocks.
+      "period": 6048, // blocks
+
+      // TODO: write a description for this field.
+      "renewwindow": 3024 // blocks
```
VoidingWarranties: @lukechampine and @DavidVorick what would be a good description for this field?

doc/api/Renter.md
```diff
@@ -0,0 +1,277 @@
...
+      "endheight": 50000, // block height
+
+      // ID of the file contract.
+      "id": "1234567890abcdef0123456789abcdef0123456789abcdef0123456789abcdef",
+
+      // Address of the host the file contract was formed with.
+      "netaddress": "12.34.56.78:9",
+
+      // TODO: ??? remaining funds? initial funds?
+      "renterfunds": "1234", // hastings
```
VoidingWarranties: @lukechampine and @DavidVorick what would be a good description for this field?

doc/api/Renter.md
```diff
@@ -0,0 +1,277 @@
...
+
+      // Size of the file contract. TODO: elaborate.
+      "size": 1024 // bytes TODO: is this the right unit?
```
VoidingWarranties: @lukechampine and @DavidVorick this seems like it could have a better description.

doc/api/Renter.md
mnsl: If the current blockheight + the renewal window >= the height the contract is scheduled to end, the contract is renewed automatically.  Must be nonzero.

doc/API.md
lukechampine: yes. It is the number of sectors * the sector size (4MiB)

doc/api/Renter.md
mnsl: Yes, bytes is correct.  Also we mean the size of the file being uploaded, not the contract itself.. should clarify

doc/API.md
lukechampine: `types.Timestamp` is a bit unwieldy, since it's just a large number. `time.Time` is easier to process in other languages and is meant for human consumption. It's a pretty minor distinction though.

doc/API.md
lukechampine: it still blocks AFAIK

doc/api/Renter.md
lukechampine: remaining funds. It's the amount of money that the renter has left to spend on uploads/downloads

doc/api/Renter.md
lukechampine: Number of bytes that have been uploaded to the host

doc/API.md
DavidVorick: If the group gets too large it makes sense to break it apart into two smaller groups. Each group should be sorted alphabetically, of course.

doc/api/Renter.md
DavidVorick: Basically if there are less than 'renewwindow' blocks left in the contract, the contract is renewed

doc/api/Renter.md
DavidVorick: @lukechampine this is the size of the file contract, post padding right? So it's always going to be a factor of modules.SectorSize, and may be larger than the sum of the sizes of all the files uploaded to the host (especially if you are uploading lots of small files)?

doc/API.md
DavidVorick: ?

VoidingWarranties:
This PR closes #1395.

doc/api/Renter.md
lukechampine: correct, it will always be a multiple of the sector size, and it may be padded. But it is the number of bytes uploaded.

(well, not exactly -- if you modify the data later, you have to upload those new bytes as well, but the size won't change. So maybe a better description is warranted.)

doc/API.md
VoidingWarranties: Presumably the destination should include the full path to the file, not just the parent directory's path.

doc/api/Renter.md
```diff
@@ -0,0 +1,277 @@
...
+
+modify settings that control the renter's behavior.
+
+###### Query String Parameters
+```
+// Number of hastings allocated for file contracts in the given period.
+funds // (optional) hastings
+
+// Duration of contracts formed.
+period // (optional) block height
```
VoidingWarranties: Not optional

doc/api/Renter.md
```diff
@@ -0,0 +1,277 @@
...
+```
+
+#### /renter [POST]
+
+modify settings that control the renter's behavior.
+
+###### Query String Parameters
+```
+// Number of hastings allocated for file contracts in the given period.
+funds // (optional) hastings
```
VoidingWarranties: Not optional

doc/API.md
```diff
@@ -711,253 +711,190 @@ The input byte array should be 80 bytes that form the solved block header. *Unli
...
-Function: Sets the contract allowance.
-
-Parameters: none
+###### Query String Parameters [(with comments)](/doc/api/Renter.md#query-string-parameters)
 ```
-funds  types.Currency    (string)
-hosts  uint64
-period types.BlockHeight (uint64)
+funds  // (optional) hastings
+hosts  // (optional)
```
VoidingWarranties: Cannot set hosts

doc/API.md
```diff
@@ -711,253 +711,190 @@ The input byte array should be 80 bytes that form the solved block header. *Unli
...
+period // (optional) block height
```
VoidingWarranties: Not optional

doc/API.md
```diff
@@ -698,253 +698,189 @@ description of the byte encoding.
...
-* /renter/download/{siapath} [GET]
-* /renter/rename/{siapath}   [POST]
-* /renter/upload/{siapath}   [POST]
+| Route                                                         | HTTP verb |
+| ------------------------------------------------------------- | --------- |
+| [/renter](#renter-get)                                        | GET       |
+| [/renter](#renter-post)                                       | POST      |
+| [/renter/contracts](#rentercontracts-get)                     | GET       |
+| [/renter/downloads](#renterdownloads-get)                     | GET       |
+| [/renter/files](#renterfiles-get)                             | GET       |
```
lukechampine: this link appears to be broken

doc/api/Renter.md
```diff
@@ -0,0 +1,280 @@
...
+```
+
+###### Response
+standard success or error response. See
+[API.md#standard-responses](/doc/API.md#standard-responses).
+
+#### /renter/contracts [GET]
+
+// TODO: is this description correct?
+returns all contracts that have been formed by the renter.
```
lukechampine: I think it returns only the active contracts. If a contract expires, it will be deleted from the renter and no longer appears.

doc/api/Renter.md
```diff
@@ -0,0 +1,280 @@
...
+
+###### JSON Response
+// TODO: are these field descriptions correct?
```
lukechampine: yep

doc/api/Renter.md
```diff
@@ -0,0 +1,280 @@
...
+
+###### Response
+standard success or error response. See
+[API.md#standard-responses](/doc/API.md#standard-responses).
+
+#### /renter/rename/___*siapath___ [POST]
+
+renames a file. Does not rename any downloads or source files, only renames the
+entry in the renter.
+// TODO: what happens if newsiapath already exists? What happens if siapath doesn't exist?
```
lukechampine: errors in both cases

doc/API.md
lukechampine: see other comment

doc/API.md
lukechampine: see other comment

VoidingWarranties:
Rebased and should be good to go.

doc/API.md
VoidingWarranties: Good catch. Fixed! (change isn't in this diff)

lukechampine:
LGTM

PR #1419, opened 2016-08-17
add `api.Client` API communication methods

johnathanhowell:
This PR adds `api.Client`, an interface for communicating with a Sia API.  It handles creating the request with the proper user agent and url, basic authentication, and decoding.  Here's an example usage:

```
c := api.NewClient("localhost:9980", "")
var gatewayInfo api.GatewayGET
err := c.Get("/gateway", &gatewayInfo)
if err != nil {
   // handle this error
}
fmt.Println(gatewayInfo)
```

this should provide a more elegant and easy-to-grok interface for Golang devs to use when they use the Sia API.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -57,20 +57,15 @@ func (c *Client) Get(resource string, obj interface{}) error {
 		return apiErr
 	}
 
-	if res.StatusCode == http.StatusNoContent {
-		return Error{"expecting a response, but API returned status code 204 No Content"}
-	}
-
-	err = json.NewDecoder(res.Body).Decode(obj)
-	if err != nil {
-		return err
+	if res.StatusCode != http.StatusNoContent {
+		return json.NewDecoder(res.Body).Decode(obj)
 	}
 	return nil
 }
 
 // POST makes a POST request to the resource at `resource`, using `data` as the
-// request body.
-func (c *Client) Post(resource string, data string) error {
+// request body.  The response, if provided, will be decoded into `obj`.
+func (c *Client) Post(resource string, data string, obj interface{}) error {
 	url := "http://" + c.address + resource
 	req, err := http.NewRequest("POST", url, strings.NewReader(data))
 	if err != nil {
@@ -100,5 +95,8 @@ func (c *Client) Post(resource string, data string) error {
 		return apiErr
 	}
 
+	if res.StatusCode != http.StatusNoContent {
+		return json.NewDecoder(res.Body).Decode(&obj)
+	}
 	return nil
 }
--- a/api/client.go
+++ b/api/client.go
@@ -57,7 +57,7 @@ func (c *Client) Get(resource string, obj interface{}) error {
 		return apiErr
 	}
 
-	if res.StatusCode != http.StatusNoContent {
+	if res.StatusCode != http.StatusNoContent && obj != nil {
 		return json.NewDecoder(res.Body).Decode(obj)
 	}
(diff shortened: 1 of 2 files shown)


lukechampine:
(For context: the plan is to deprecate the existing `api.HttpX` functions. Eventually, we may add helper methods to `api.Client`, e.g. `c.Gateway() (api.GatewayInfo, error)`)

api/client.go
```diff
@@ -0,0 +1,102 @@
...
+
+	if res.StatusCode != http.StatusNoContent && obj != nil {
+		return json.NewDecoder(res.Body).Decode(obj)
+	}
+	return nil
+}
+
+// POST makes a POST request to the resource at `resource`, using `data` as the
+// request body.  The response, if provided, will be decoded into `obj`.
+func (c *Client) Post(resource string, data string, obj interface{}) error {
```
VoidingWarranties: Docstring incorrect (POST -> Post)

VoidingWarranties:
It would be helpful for clients of these methods to know the HTTP status code of the response. Specifically this would be useful for differentiating between client and server errors. The easiest way would be to return another value from `Get()` and `Post()`, but we could also add a `statuscode` field to `api.Error`.

johnathanhowell:
What's the use case for returning status codes?

lukechampine:
status code seems like it belongs in `api.Error` anyway.

api/client.go
```diff
@@ -0,0 +1,102 @@
...
+		address:  address,
+		password: password,
+	}
+}
+
+// Get requests the resource at `resource` and decodes it into `obj`, returning an
+// error if requesting or decoding the resource fails.  A non-2xx status code
+// constitutes a request failure.
+func (c *Client) Get(resource string, obj interface{}) error {
+	url := "http://" + c.address + resource
```
lukechampine: might be safest to use `path.Join` here, not sure. That way, `resource` can either be `foo/bar` or `/foo/bar`.

api/client.go
```diff
@@ -0,0 +1,102 @@
...
+		err = json.NewDecoder(res.Body).Decode(&apiErr)
+		if err != nil {
+			return err
+		}
+		return apiErr
+	}
+
+	if res.StatusCode != http.StatusNoContent && obj != nil {
+		return json.NewDecoder(res.Body).Decode(&obj)
+	}
```
lukechampine: as per the [docstring](https://golang.org/pkg/net/http/#Response), `res.Body` should always be fully read even if you are discarding its content. This allows the connection to be reused.

To keep the code clean, you could do this in the `defer`:
```go
defer func() {
    io.Copy(ioutil.Discard, res.Body)
    res.Body.Close()
}()
```

api/client.go
johnathanhowell: `path.Join` doesn't work with URLs.  You could do something like this:

```
u, err := url.Parse("http://foo")
u.Path = path.Join(u.Path, "bar")
resourceUrl := u.String() // http://foo/bar
``

api/client.go
johnathanhowell: I definitely prefer the `defer` approach.

api/client.go
lukechampine: why, because of the `// -> /` reduction?

it's a non-issue though, I think callers know to begin routes with a leading slash.

DavidVorick: might there be other edge cases?

PR #1420, opened 2016-08-17
Add ProcessedTransactions to wallet db

lukechampine:
This only adds one new field to the db, but it's a weird field. I also reorganized the db helper functions.

Relying on proper transaction ordering feels brittle. Eventually I plan to cache the transactions in memory, and use a map so that transactions can be removed based on their ID, not their index. (You could do this now, but it would require lots of db I/O).

### DIFF
--- a/modules/wallet/database.go
+++ b/modules/wallet/database.go
@@ -1,6 +1,8 @@
 package wallet
 
 import (
+	"reflect"
+
 	"github.com/NebulousLabs/Sia/encoding"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/types"
@@ -40,6 +42,27 @@ func dbGet(b *bolt.Bucket, key, val interface{}) error {
 // dbDelete is a helper function for deleting a marshalled key/value pair.
 func dbDelete(b *bolt.Bucket, key interface{}) error {
 	return b.Delete(encoding.Marshal(key))
+}
+
+// dbForEach is a helper function for iterating over a bucket and calling fn
+// on each entry. fn must be a function with two parameters.
+func dbForEach(b *bolt.Bucket, fn interface{}) error {
+	// check function type
+	fnVal, fnTyp := reflect.ValueOf(fn), reflect.TypeOf(fn)
+	if fnTyp.Kind() != reflect.Func || fnTyp.NumIn() != 2 {
+		panic("bad fn type: needed func(a, b), got " + fnTyp.String())
+	}
+
+	return b.ForEach(func(keyBytes, valBytes []byte) error {
+		key, val := reflect.New(fnTyp.In(0)), reflect.New(fnTyp.In(1))
+		if err := encoding.Unmarshal(keyBytes, key.Interface()); err != nil {
+			return err
+		} else if err := encoding.Unmarshal(valBytes, val.Interface()); err != nil {
+			return err
+		}
+		fnVal.Call([]reflect.Value{key.Elem(), val.Elem()})
+		return nil
+	})
 }
 
 // dbPutHistoricClaimStart stores the number of siafunds corresponding to the
--- a/modules/wallet/database.go
+++ b/modules/wallet/database.go
@@ -93,6 +93,19 @@ func dbDeleteProcessedTransaction(tx *bolt.Tx, id types.TransactionID) error {
 	return dbDelete(tx.Bucket(bucketProcessedTransactions), id)
 }
 
+// dbForEachProcessedTransaction iterates over the ProcessedTransactions
+// bucket, calling fn on each entry.
+func dbForEachProcessedTransaction(tx *bolt.Tx, fn func(modules.ProcessedTransaction)) error {
+	return tx.Bucket(bucketProcessedTransactions).ForEach(func(_, val []byte) error {
+		var pt modules.ProcessedTransaction
+		if err := encoding.Unmarshal(val, &pt); err != nil {
+			return err
+		}
+		fn(pt)
+		return nil
+	})
+}
+
 // dbPutSiacoinOutput stores the output corresponding to the specified
 // SiacoinOutputID.
 func dbPutSiacoinOutput(tx *bolt.Tx, id types.SiacoinOutputID, output types.SiacoinOutput) error {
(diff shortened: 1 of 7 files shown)


modules/wallet/database.go
```diff
@@ -1,21 +1,51 @@
...
+	bucketSiacoinOutputs = []byte("bucketSiacoinOutputs")
+	// bucketSiacoinOutputs maps a SiafundOutputID to its SiafundOutput. Only
+	// outputs that the wallet controls are stored. The wallet uses these
+	// outputs to fund transactions.
+	bucketSiafundOutputs = []byte("bucketSiafundOutputs")
+	// bucketSpentOutputs maps an OutputID to the height at which it was
+	// spent. Only outputs spent by the wallet are stored. The wallet tracks
+	// these outputs so that it can reuse them if they are not confirmed on
+	// the blockchain.
+	bucketSpentOutputs = []byte("bucketSpentOutputs")
```
lukechampine: this description doesn't jive with the way `SpentOutputs` is used in the code. Currently, only the transaction builder uses it, and it uses it only to check that it is not reusing outputs it has already spent. Is that it's true purpose? Does this field need to be in the db at all?

modules/wallet/transactions.go
```diff
@@ -89,19 +104,27 @@ func (w *Wallet) Transactions(startHeight, endHeight types.BlockHeight) (pts []m
...
+		for key, val := c.First(); key != nil; key, val = c.Next() {
+			var pt modules.ProcessedTransaction
+			if err := encoding.Unmarshal(val, &pt); err != nil {
+				return err
+			}
+			if pt.ConfirmationHeight < startHeight {
+				continue
+			} else if pt.ConfirmationHeight > endHeight {
+				// transactions are stored in chronological order, so we can
+				// break as soon as we are above endHeight
```
lukechampine: ideally we would seek to `startHeight` first, but this is not possible because the key is a sequence integer, not a blockheight. Using a blockheight would mean storing a `[]Transaction` as the value rather than a single transaction. I'm not too worried about this though because this code will likely be obsoleted by caching anyway.

modules/wallet/database.go
DavidVorick: > The wallet tracks these outputs so that it can reuse them if they are not confirmed on the blockchain.

This statement is correct, though maybe confusing. The alternative to tracking the height at which an output was spent is to delete the output. The transaction builder will not spend an output that has been spent (but not confirmed on-blockchain) in the past 80 blocks. After 80 blocks, it'll reuse that output again.

It would probably be better if this field was in the database rather than just in memory. If it's just in memory then you lose the info at restart, which could cause other nodes problems and result in you accidentally double-spending.

modules/wallet/database.go
```diff
@@ -38,91 +68,115 @@ func dbDelete(b *bolt.Bucket, key interface{}) error {
...
-// it can determine the value of outputs that were not created
-// by the wallet.
 func dbPutHistoricOutput(tx *bolt.Tx, id types.OutputID, c types.Currency) error {
 	return dbPut(tx.Bucket(bucketHistoricOutputs), id, c)
 }
-
-// dbGetHistoricOutput retrieves the number of coins corresponding to the
-// specified OutputID. The wallet uses this mapping to determine the value of
-// outputs that were not created by the wallet.
 func dbGetHistoricOutput(tx *bolt.Tx, id types.OutputID) (c types.Currency, err error) {
```
DavidVorick: why did you remove the docstrings? Please have docstrings on all functions

modules/wallet/database.go
```diff
@@ -38,91 +68,115 @@ func dbDelete(b *bolt.Bucket, key interface{}) error {
...
+	key, err := b.NextSequence()
+	if err != nil {
+		return err
+	}
+	// big-endian is used so that the keys are properly sorted
+	keyBytes := make([]byte, 8)
+	binary.BigEndian.PutUint64(keyBytes, key)
+	return b.Put(keyBytes, encoding.Marshal(pt))
+}
+func dbGetLastProcessedTransaction(tx *bolt.Tx) (pt modules.ProcessedTransaction, err error) {
```
DavidVorick: same here, strongly prefer to see docstrings on every function

modules/wallet/database.go
```diff
@@ -38,91 +68,115 @@ func dbDelete(b *bolt.Bucket, key interface{}) error {
...
+		key, val := reflect.New(fnTyp.In(0)), reflect.New(fnTyp.In(1))
+		if err := encoding.Unmarshal(keyBytes, key.Interface()); err != nil {
+			return err
+		} else if err := encoding.Unmarshal(valBytes, val.Interface()); err != nil {
+			return err
+		}
+		fnVal.Call([]reflect.Value{key.Elem(), val.Elem()})
+		return nil
+	})
+}
```
DavidVorick: I don't like this function, I think I would prefer to see a `ForEach` defined on each type that gets used in the codebase, similar to the other helper functions in this file. The reflection here is complicated, and is performed in a loop.

Maybe it's just because I'm pretty unfamiliar with reflection, but I don't find this code to be readable.

modules/wallet/transactions.go
DavidVorick: this basically makes the cacheing a requirement though. Doing 10,000 seeks + reads on a bolt database is going to take a while, especially on an HDD.

modules/wallet/database.go
lukechampine: ah, ok, I was confused because I didn't see the outputs being deleted anywhere. That seems slightly better to me, since otherwise `spentOutputs` will grow forever.

modules/wallet/database.go
lukechampine: The docstrings would have no value. It's plainly obvious that they are type-safe wrappers around `dbPut`, `dbGet`, etc. Those functions all have good docstrings. Furthermore, the purpose of each bucket is now explained at the top of the file. There's nothing to say in the docstring other than to describe what bucket and types it operates on -- which is given in the function definition.
Docstrings are always required on exported functions, but in rare cases, they can be omitted on unexported functions. The [stubs in the renter tests](https://github.com/NebulousLabs/Sia/blob/master/modules/renter/renter_test.go#L167) are a good example of where a single docstring makes more sense than docstrings on each individual function.

modules/wallet/database.go
lukechampine: Since these functions are more complicated, I'm ok adding docstrings to them

modules/wallet/database.go
lukechampine: The reflection is pretty simple compared to most reflection. It's used in three places: 1) to check that the function has the correct type; 2) to allocate values to decode into; 3) to call the function.

I actually did initially define separate functions; see 0dc9cab for the diff. Though a reflection-based approach may be less readable, defining a separate `ForEach` for each bucket has its downsides as well. The biggest one is that the reader cannot easily confirm that each `ForEach` function is doing the same thing. Even if they look broadly similar, there may be subtle differences. And of course, more code means more places for a bug to creep in.

modules/wallet/database.go
DavidVorick: I thought they got deleted in the consensus update loop, but maybe they don't get deleted at all.

modules/wallet/database.go
lukechampine: I think the conclusion here was: yes, it's ugly, but it almost certainly won't be merged into master; the buckets will be stored in memory, allowing for a traditional `for` loop.

DavidVorick: ok

PR #1422, opened 2016-08-18
Gateway concurrency

DavidVorick:
This pull request is most easily digested going commit-to-commit. They are pretty cleanly separated.

I went through and fixed up a lot of the concurrency patterns in the gateway. None of the changes have proper tests, but all of the old tests are still passing. Several deadlocks have been removed, and the biggest offenders in terms of quick-shutdown have been removed.

Some things remain on my todo list:

1. `threadedListenPeer` is not fixed up yet, it violates the 'don't hold a `tg.Add()` for long periods of time` rule, and has some other weirdness to the concurrency that I don't like.
2. `managedConnectNewPeer`, `managedConnectOldPeer`, `managedAcceptConnectNewPeer`, and `managedAcceptConnectOldPeer` all don't support quick-shutdown. I may have slipped in a goroutine here or there that is not being correctly closed with regards to these functions, can clean that up tomorrow.
3. `threadedHandleConn` does not yet support clean shutdown.

I am considering abandoning the pursuit of clean shutdown for the code that does not have it yet. The biggest offenders have already been addressed. That would just leave `threadedListenPeer` on my todo list.

Oh, and testing, as a lot of this code is not tested well, even though the whole test suite is passing.

### DIFF
--- a/modules/gateway.go
+++ b/modules/gateway.go
@@ -8,12 +8,8 @@ const (
 	// GatewayDir is the name of the directory used to store the gateway's
 	// persistent data.
 	GatewayDir = "gateway"
-	// WellConnectedThreshold is the number of outbound connections at which the
-	// gateway will not attempt to make new outbound connections.
-	WellConnectedThreshold = 8
 )
 
-// TODO: Move this and its functionality into the gateway package.
 var (
 	// BootstrapPeers is a list of peers that can be used to find other peers -
 	// when a client first connects to the network, the only options for
--- a/modules/gateway/conn.go
+++ b/modules/gateway/conn.go
@@ -17,3 +17,14 @@ type peerConn struct {
 func (pc peerConn) RPCAddr() modules.NetAddress {
 	return pc.dialbackAddr
 }
+
+// dial will dial the input address and return a connection. dial appropriately
+// handles things like clean shutdown, fast shutdown, and chooses the correct
+// communication protocol.
+func (g *Gateway) dial(addr modules.NetAddress) (net.Conn, error) {
+	dialer := &net.Dialer{
+		Timeout: dialTimeout,
+		Cancel:  g.threads.StopChan(),
+	}
+	return dialer.Dial("tcp", string(addr))
+}
--- a/modules/gateway/consts.go
+++ b/modules/gateway/consts.go
@@ -7,6 +7,10 @@ import (
 )
 
 const (
+	// handshakeUpgradeVersion is the version where the gateway handshake RPC
+	// was altered to include adiitional information transfer.
+	handshakeUpgradeVersion = "1.0.0"
+
 	// minAcceptableVersion is the version below which the gateway will refuse to
 	// connect to peers and reject connection attempts.
 	//
--- a/modules/gateway/consts.go
+++ b/modules/gateway/consts.go
@@ -18,7 +18,18 @@ const (
 var (
 	// maxSharedNodes defines the number of nodes that will be shared between
 	// peers when they are expanding their node lists.
-	maxSharedNodes uint64 = 10
+	maxSharedNodes = func() uint64 {
+		switch build.Release {
+		case "dev":
+			return 5
+		case "standard":
+			return 10
+		case "testing":
+			return 3
+		default:
+			panic("unrecognized build.Release in healthyNodeListLen")
+		}
+	}()
 
 	// nodePurgeDelay defines the amount of time that is waited between each
 	// iteration of the node purge loop.
--- a/modules/gateway/consts.go
+++ b/modules/gateway/consts.go
@@ -1,0 +1,159 @@
+package gateway
+
+import (
+	"time"
+
+	"github.com/NebulousLabs/Sia/build"
+)
+
+const (
+	// minAcceptableVersion is the version below which the gateway will refuse to
+	// connect to peers and reject connection attempts.
+	//
+	// Reject peers < v0.4.0 as the previous version is v0.3.3 which is
+	// pre-hardfork.
+	minAcceptableVersion = "0.4.0"
+)
+
+var (
+	// maxSharedNodes defines the number of nodes that will be shared between
+	// peers when they are expanding their node lists.
+	maxSharedNodes uint64 = 10
+
+	// nodePurgeDelay defines the amount of time that is waited between each
+	// iteration of the node purge loop.
+	nodePurgeDelay = func() time.Duration {
+		switch build.Release {
+		case "dev":
+			return 20 * time.Second
+		case "standard":
+			return 10 * time.Minute
+		case "testing":
+			return 2 * time.Second
+		default:
+			panic("unrecognized build.Release in nodePurgeDelay")
+		}
+	}()
+
+	// healthyNodeListLen defines the number of nodes that the gateway must
+	// have in the node list before it will stop asking peers for more nodes.
+	healthyNodeListLen = func() int {
+		switch build.Release {
+		case "dev":
+			return 30
+		case "standard":
+			return 200
+		case "testing":
+			return 15
+		default:
+			panic("unrecognized build.Release in healthyNodeListLen")
+		}
+	}()
+
+	// pruneNodeListLen defines the number of nodes that the gateway must have
+	// to be pruning nodes from the node list.
+	pruneNodeListLen = func() int {
+		switch build.Release {
+		case "dev":
+			return 15
+		case "standard":
+			return 50
+		case "testing":
+			return 10
+		default:
+			panic("unrecognized build.Release in pruneNodeListLen")
+		}
+	}()
+)
+
+var (
+	// fullyConnectedThreshold defines the number of peers that the gateway can
+	// have before it stops accepting inbound connections.
+	fullyConnectedThreshold = func() int {
+		switch build.Release {
+		case "dev":
+			return 20
+		case "standard":
+			return 128
+		case "testing":
+			return 10
+		default:
+			panic("unrecognized build.Release in fullyConnectedThreshold")
+		}
+	}()
+
+	// noPeersDelay defines the amount of time that is waited between
+	// iterations of the peer formation loop if the gateway is well connected.
+	noPeersDelay = func() time.Duration {
+		switch build.Release {
+		case "dev":
+			return 10 * time.Second
+		case "standard":
+			return 20 * time.Second
+		case "testing":
+			return 3 * time.Second
+		default:
+			panic("unrecognized build.Release in noPeersDelay")
+		}
+	}()
+
+	// wellConnectedDelay defines the amount of time that is waited between
+	// iterations of the peer formation loop if the gateway is well connected.
+	wellConnectedDelay = func() time.Duration {
+		switch build.Release {
+		case "dev":
+			return 1 * time.Minute
+		case "standard":
+			return 5 * time.Minute
+		case "testing":
+			return 5 * time.Second
+		default:
+			panic("unrecognized build.Release in wellConnectedDelay")
+		}
+	}()
+
+	// wellConnectedThreshold is the number of outbound connections at which
+	// the gateway will not attempt to make new outbound connections.
+	wellConnectedThreshold = func() int {
+		switch build.Release {
+		case "dev":
+			return 5
+		case "standard":
+			return 8
+		case "testing":
+			return 3
+		default:
+			panic("unrecognized build.Release in wellConnectedThreshold")
+		}
+	}()
+)
+
+var (
+	// The gateway will sleep this long between incoming connections.
+	acceptInterval = func() time.Duration {
+		switch build.Release {
+		case "dev":
+			return 3 * time.Second
+		case "standard":
+			return 3 * time.Second
+		case "testing":
+			return 10 * time.Millisecond
+		default:
+			panic("unrecognized build.Release in acceptInterval")
+		}
+	}()
+
+	// the gateway will abort a connection attempt after this long
+	dialTimeout = func() time.Duration {
+		switch build.Release {
+		case "dev":
+			return 45 * time.Second
+		case "standard":
+			return 2 * time.Minute
+		case "testing":
+			return 2 * time.Second
+		default:
+			panic("unrecognized build.Release in dialTimeout")
+		}
+	}()
+)
--- a/modules/gateway/consts.go
+++ b/modules/gateway/consts.go
@@ -40,7 +40,7 @@ var (
 		case "standard":
 			return 10 * time.Minute
 		case "testing":
-			return 2 * time.Second
+			return 1 * time.Second
 		default:
 			panic("unrecognized build.Release in nodePurgeDelay")
 		}
@@ -135,7 +135,7 @@ var (
 		case "standard":
 			return 5 * time.Minute
 		case "testing":
-			return 5 * time.Second
+			return 2 * time.Second
 		default:
 			panic("unrecognized build.Release in wellConnectedDelay")
 		}
--- a/modules/gateway/consts.go
+++ b/modules/gateway/consts.go
@@ -16,6 +16,21 @@ const (
 )
 
 var (
+	// healthyNodeListLen defines the number of nodes that the gateway must
+	// have in the node list before it will stop asking peers for more nodes.
+	healthyNodeListLen = func() int {
+		switch build.Release {
+		case "dev":
+			return 30
+		case "standard":
+			return 200
+		case "testing":
+			return 15
+		default:
+			panic("unrecognized build.Release in healthyNodeListLen")
+		}
+	}()
+
 	// maxSharedNodes defines the number of nodes that will be shared between
 	// peers when they are expanding their node lists.
 	maxSharedNodes = func() uint64 {
@@ -40,24 +55,24 @@ var (
 		case "standard":
 			return 10 * time.Minute
 		case "testing":
-			return 1 * time.Second
+			return 500 * time.Millisecond
 		default:
 			panic("unrecognized build.Release in nodePurgeDelay")
 		}
 	}()
 
-	// healthyNodeListLen defines the number of nodes that the gateway must
-	// have in the node list before it will stop asking peers for more nodes.
-	healthyNodeListLen = func() int {
+	// nodeListDelay defines the amount of time that is waited between each
+	// iteration of the node list loop.
+	nodeListDelay = func() time.Duration {
 		switch build.Release {
 		case "dev":
-			return 30
+			return 3 * time.Second
 		case "standard":
-			return 200
+			return 5 * time.Second
 		case "testing":
-			return 15
+			return 500 * time.Millisecond
 		default:
-			panic("unrecognized build.Release in healthyNodeListLen")
+			panic("unrecognized build.Release in nodePurgeDelay")
 		}
 	}()
 
@@ -88,7 +103,7 @@ var (
 		case "standard":
 			return 5 * time.Second
 		case "testing":
-			return 1 * time.Second
+			return 500 * time.Millisecond
 		default:
 			panic("unrecognized build.Release in wellConnectedDelay")
 		}
@@ -135,7 +150,7 @@ var (
 		case "standard":
 			return 5 * time.Minute
 		case "testing":
-			return 2 * time.Second
+			return 3 * time.Second
 		default:
 			panic("unrecognized build.Release in wellConnectedDelay")
 		}
--- a/modules/gateway/consts.go
+++ b/modules/gateway/consts.go
@@ -191,11 +191,11 @@ var (
 	dialTimeout = func() time.Duration {
 		switch build.Release {
 		case "dev":
-			return 45 * time.Second
+			return 20 * time.Second
 		case "standard":
 			return 2 * time.Minute
 		case "testing":
-			return 2 * time.Second
+			return 500 * time.Millisecond
 		default:
 			panic("unrecognized build.Release in dialTimeout")
 		}
--- a/modules/gateway/consts.go
+++ b/modules/gateway/consts.go
@@ -165,7 +165,7 @@ var (
 		case "standard":
 			return 8
 		case "testing":
-			return 3
+			return 4
 		default:
 			panic("unrecognized build.Release in wellConnectedThreshold")
 		}
--- a/modules/gateway/consts.go
+++ b/modules/gateway/consts.go
@@ -67,6 +67,22 @@ var (
 )
 
 var (
+	// acquiringPeersDelay defines the amount of time that is waited between
+	// iterations of the peer acquisition loop if the gateway is actively
+	// forming new connections with peers.
+	acquiringPeersDelay = func() time.Duration {
+		switch build.Release {
+		case "dev":
+			return 3 * time.Minute
+		case "standard":
+			return 5 * time.Second
+		case "testing":
+			return 1 * time.Second
+		default:
+			panic("unrecognized build.Release in wellConnectedDelay")
+		}
+	}()
+
 	// fullyConnectedThreshold defines the number of peers that the gateway can
 	// have before it stops accepting inbound connections.
 	fullyConnectedThreshold = func() int {
@@ -83,7 +99,8 @@ var (
 	}()
 
 	// noPeersDelay defines the amount of time that is waited between
-	// iterations of the peer formation loop if the gateway is well connected.
+	// iterations of the peer acquisition loop if the gateway does not have
+	// enough nodes.
 	noPeersDelay = func() time.Duration {
 		switch build.Release {
 		case "dev":
@@ -98,7 +115,8 @@ var (
 	}()
 
 	// wellConnectedDelay defines the amount of time that is waited between
-	// iterations of the peer formation loop if the gateway is well connected.
+	// iterations of the peer acquisition loop if the gateway is well
+	// connected.
 	wellConnectedDelay = func() time.Duration {
 		switch build.Release {
 		case "dev":
--- a/modules/gateway/gateway.go
+++ b/modules/gateway/gateway.go
@@ -148,16 +148,30 @@ func New(addr string, persistDir string) (g *Gateway, err error) {
 	// Spawn the peer connection listener.
 	go g.threadedListen(threadedListenClosedChan)
 
-	// Spwan the peer manager and provide tools for ensuring clean shutdown.
+	// Spawn the peer manager and provide tools for ensuring clean shutdown.
 	peerManagerClosedChan := make(chan struct{})
 	g.threads.OnStop(func() {
 		<-peerManagerClosedChan
 	})
-	go g.threadedPeerManager(peerManagerClosedChan)
+	go g.permanentPeerManager(peerManagerClosedChan)
 
+	// Spawn the node manager and provide tools for ensuring clean shudown.
+	nodeManagerClosedChan := make(chan struct{})
+	g.threads.OnStop(func() {
+		<-nodeManagerClosedChan
+	})
+	go g.permanentNodeManager(nodeManagerClosedChan)
+
+	// Spawn the node purger and provide tools for ensuring clean shutdown.
+	nodePurgerClosedChan := make(chan struct{})
+	g.threads.OnStop(func() {
+		<-nodePurgerClosedChan
+	})
+	go g.permanentNodePurger(nodePurgerClosedChan)
+
+	// Spawn threads to take care of port forwarding and hostname discovery.
 	go g.threadedForwardPort(g.port)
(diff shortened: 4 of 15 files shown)


modules/gateway/gateway.go
```diff
@@ -64,8 +64,6 @@ func (g *Gateway) Close() error {
 	// Unregister RPCs. Not strictly necessary for clean shutdown in this specific
 	// case, as the handlers should only contain references to the gateway itself,
 	// but do it anyways as an example for other modules to follow.
-	g.UnregisterRPC("ShareNodes")
-	g.UnregisterConnectCall("ShareNodes")
```
mnsl: need to remove comment as well

mnsl: never mind, fixed in a later commit

DavidVorick:
Testing aside, I have finished this PR for now. I decided not to write code to interrupt the peer handshake during gateway shutdown, which means quick-shutdown has not fully been realized. But it should be pretty rare that the gateway is closed mid-handshake anyway, especially if it's been running for a while.

modules/gateway/nodes.go
```diff
@@ -43,15 +44,26 @@ func (g *Gateway) removeNode(addr modules.NetAddress) error {
...
+		return err
+	}
+	for node := range g.nodes {
+		if r <= 0 {
+			return node, nil
 		}
+		r--
 	}
 
 	return "", errNoPeers
```
mnsl: this extra return statement is unnecessary, since `r` is less than `len(g.nodes)` by definition, so the function will return at some point in the for loop

DavidVorick: what if before the loop starts len(g.nodes) == 0?

DavidVorick: oh nvm that's covered at the top of the function

modules/gateway/gateway.go
lukechampine: that's one of the downsides of stepping through commit-by-commit :/

VoidingWarranties:
Can we make the concurrency rules more explicit? For example you mention that methods should not hold a `tg.Add` for long periods of time. Why is this so?

modules/gateway.go
```diff
@@ -8,12 +8,8 @@ const (
 	// GatewayDir is the name of the directory used to store the gateway's
 	// persistent data.
 	GatewayDir = "gateway"
-	// WellConnectedThreshold is the number of outbound connections at which the
-	// gateway will not attempt to make new outbound connections.
-	WellConnectedThreshold = 8
 )
 
-// TODO: Move this and its functionality into the gateway package.
```
VoidingWarranties: Why remove this TODO?

DavidVorick: ah mistake. I tried to execute the TODO, then decided it was out-of-scope for this PR. It's on my 'let's do this next PR' list.

modules/gateway/upnp.go
```diff
@@ -109,15 +109,22 @@ func (g *Gateway) threadedForwardPort(port string) {
 	}
 
 	g.log.Println("INFO: successfully forwarded port", port)
+
+	// Establish port-clearing at shutdown.
+	g.threads.AfterStop(func() {
+		g.managedClearPort(g.myAddr.Port())
```
VoidingWarranties: It isn't safe to use `g.myAddr.Port()` as the hostname may not have been discovered yet, in which case `myAddr` will be the empty string. We used to set `myAddr` to ":port", but we changed this behavior when we cleaned up the `NetAddress` and implemented better `IsValid` checking.

DavidVorick:
@VoidingWarranties I'll write a blog post on it, and then update the docs for our concurrency patterns. You don't want to hold `tg.Add` for long periods of time because it'll block `tg.Flush()` calls. 'Long periods' means longer than one logical function. e.g. a peer that will potentially be connected indefinitely. If it's something like a 35 second database operation, but it's definitely finite, then it's okay to hold the `tg.Add` for the whole operation.

I added some testing that does a lot of gateway spinning up. The upside is that it's more clear that gateway shutdown and deadlock prevention are working well. The downside is that the muxado error is much more common now.

modules/gateway/upnp.go
```diff
@@ -109,15 +109,22 @@ func (g *Gateway) threadedForwardPort(port string) {
...
+	})
 }
 
-// clearPort removes a port mapping from the router.
-func (g *Gateway) clearPort(port string) {
+// managedClearPort removes a port mapping from the router.
+func (g *Gateway) managedClearPort(port string) {
```
VoidingWarranties: Why the name change? I thought the managed prefix is used to indicate the method does it's own locking?

modules/gateway/upnp.go
```diff
@@ -77,6 +76,8 @@ func (g *Gateway) threadedLearnHostname() {
 		return
 	}
 
+	// TODO: Is this safe? What happens to processes or nodes that have already
+	// request the IP address?
```
VoidingWarranties: Interestingly the only place we use `myAddr` in the gateway is in `addNode` to ensure we don't add ourself as a node. Perhaps `gateway.Address()` should return an error or bool if the hostname hasn't been discovered yet.

But to get to the point, this is currently safe in the current code, but it may be necessary to check `myAddr.IsValid()` before using `myAddr` in future code.

modules/gateway/nodes.go
```diff
@@ -43,15 +44,26 @@ func (g *Gateway) removeNode(addr modules.NetAddress) error {
 	return nil
 }
 
+// randomNode returns a random node from the gateway. An error can be returned
```
VoidingWarranties: Missing a trailing period.

VoidingWarranties:
>I'll write a blog post on it, and then update the docs for our concurrency patterns. You don't want to hold tg.Add for long periods of time because it'll block tg.Flush() calls. 'Long periods' means longer than one logical function. e.g. a peer that will potentially be connected indefinitely. If it's something like a 35 second database operation, but it's definitely finite, then it's okay to hold the tg.Add for the whole operation.

:+1: it would also be helpful to include this in our `/doc/`s for contributors and new hires.

>I added some testing that does a lot of gateway spinning up. The upside is that it's more clear that gateway shutdown and deadlock prevention are working well. The downside is that the muxado error is much more common now.

Perhaps we should skip these tests once this PR has been merged.

modules/gateway/peers.go
```diff
@@ -442,14 +442,9 @@ func (g *Gateway) managedConnectNewPeer(conn net.Conn, remoteVersion string, rem
...
-// Gateway's peer list.
-func (g *Gateway) Connect(addr modules.NetAddress) error {
-	if err := g.threads.Add(); err != nil {
-		return err
-	}
-	defer g.threads.Done()
-
+// managedConnect establishes a persistent connection to a peer, and adds it to
+// the Gateway's peer list.
+func (g *Gateway) managedConnect(addr modules.NetAddress) error {
```
VoidingWarranties: Why separate `Connect` into `Connect` and `managedConnect`? The only other place `managedConnect` is used outside of `Connect` it calls `tg.Add(); defer tg.Done()` just like `Connect` does.

modules/gateway/rpc.go
```diff
@@ -10,8 +10,11 @@ import (
 	"github.com/NebulousLabs/Sia/modules"
 )
 
+// rpcID is an 8-byte signature that is added to all RPCs to tell the gatway
+// what to do with the RPC.
 type rpcID [8]byte
 
+// String returns a string representation of an rpcID.
```
VoidingWarranties: It would be helpful to mention that empty elements of `rpcID`s will be encoded as spaces.

modules/gateway/peers.go
```diff
@@ -493,7 +493,7 @@ func (g *Gateway) managedConnect(addr modules.NetAddress) error {
 			}
 			defer g.threads.Done()
 
-			err := g.RPC(addr, name, fn)
+			err := g.managedRPC(addr, name, fn)
```
VoidingWarranties: I don't think this was a potential deadlock as it was called from a goroutine. But the change fits best practices I think.

modules/gateway/nodes.go
```diff
@@ -102,68 +102,126 @@ func (g *Gateway) requestNodes(conn modules.PeerConn) error {
...
+		// Get a random node for scanning.
 		g.mu.RLock()
 		numNodes := len(g.nodes)
-		peer, err := g.randomPeer()
+		node, err := g.randomNode()
 		g.mu.RUnlock()
 		if err != nil {
-			// can't do much until we have peers
+			// Most likely the error indicates that there are no nodes in the
+			// node list. Wait until the next iteration to try again.
```
VoidingWarranties: It would be helpful to assert this assumption by logging the error if it is not `errNoPeers`.

VoidingWarranties: It might be helpful to assert this assumption by logging the error if it is not `errNoPeers`.

modules/gateway/upnp.go
DavidVorick: Oh! I think the host is experiencing the same problem!

Thanks, will fix both

modules/gateway/upnp.go
DavidVorick: It doesn't need a lock, which is equivalent to it doing it's own locking. Functionally, 'mamaged' tells you not to call it holding a lock. It can take a while too, so you really don't want to be holding a lock

modules/gateway/nodes.go
```diff
@@ -102,68 +102,126 @@ func (g *Gateway) requestNodes(conn modules.PeerConn) error {
...
 		conn.Close()
-		// sleep for an extra 10 minutes after success; we don't want to spam
-		// connectable nodes
+	}
+}
+
+// permanentNodeManager tries to keep the Gateway's node list healthy. As long
+// as the Gateway has fewer than minNodeListSize nodes, it asks a random peer
+// for more nodes. It also continually pings nodes in order to establish their
+// connectivity. Unresponsive nodes are aggressively removed.
```
VoidingWarranties: With the addition of `permanentNodePurger` the last sentence is no longer true.

modules/gateway/nodes.go
```diff
@@ -102,68 +102,126 @@ func (g *Gateway) requestNodes(conn modules.PeerConn) error {
...
+
+		g.mu.RLock()
+		numNodes := len(g.nodes)
+		peer, err := g.randomPeer()
+		g.mu.RUnlock()
+		if err != nil {
+			// Most commonly the error indicates that there are no peers yet.
+			// Waiting through the next iteration gives the gateway time to
+			// bootstrap.
+			continue
```
VoidingWarranties: Same comment as the one in `permanentNodePurger` (consider logging the error if it's not `errNoPeers`)

modules/gateway/nodes.go
```diff
@@ -102,68 +102,126 @@ func (g *Gateway) requestNodes(conn modules.PeerConn) error {
...
+		}
+
+		// Determine whether there are a satisfactory number of nodes in the
+		// nodelist. If there are not, use the random peer from earlier to
+		// expand the node list.
+		if numNodes < minNodeListLen {
+			// TODO: should this be done in a go func? It's a blocking
+			// function.
```
VoidingWarranties: If it's not in a goroutine and we only have 1 peer then we would repeatedly call the RPC on the peer every 5 seconds. Probably best to keep it as a blocking call, but set a relatively short deadline in `g.requestNodes`.

modules/gateway/peers_test.go
```diff
@@ -653,6 +653,8 @@ func TestDisconnect(t *testing.T) {
 		if err != nil {
 			t.Fatal("accept failed:", err)
```
VoidingWarranties: Not changed in this PR, but `t.Fatal` and `t.Error` cannot be called from a goroutine. There are probably many other places we do this, but my regex-foo is not strong enough to find them easily.

modules/gateway/peers.go
```diff
@@ -172,6 +172,7 @@ func (g *Gateway) threadedListen(closeChan chan struct{}) {
 // threadedAcceptConn adds a connecting node as a peer.
 func (g *Gateway) threadedAcceptConn(conn net.Conn) {
 	if g.threads.Add() != nil {
+		conn.Close()
```
VoidingWarranties: Good catch!

modules/gateway/nodes.go
```diff
@@ -142,28 +142,35 @@ func (g *Gateway) permanentNodePurger(closeChan chan struct{}) {
...
-			Timeout: dialTimeout,
-			Cancel:  cancelDialChan,
-		}
-		go func() {
-			select {
-			case <-cancelDialChan:
-			case <-g.threads.StopChan():
-				close(cancelDialChan)
+		var conn net.Conn
+		func() {
```
VoidingWarranties: Can you elaborate on the addition on the anonymous function? The only difference I see is where `close(dialDoneChan)` is called.

modules/gateway/peers.go
```diff
@@ -177,6 +177,22 @@ func (g *Gateway) threadedAcceptConn(conn net.Conn) {
...
+	go func() {
+		select {
+		case <-connectionSuccessChan:
+			// After the connection is successful, alternate code is in place
+			// to close the connection. Do nothing.
+		case <-g.threads.StopChan():
+			// The gateway is shutting down, but the connection has not
+			// finished forming. Interrupt the connection formation so the
+			// gateway can shut down quickly.
+			conn.Close()
```
VoidingWarranties: Alternatively we could disconnect from all peers in `gateway.Close`. I thought we already did this, but maybe that was an idea that I never implemented.

modules/gateway/gateway.go
```diff
@@ -93,6 +106,15 @@ func New(addr string, persistDir string) (g *Gateway, err error) {
 		}
 	})
 
+	// Establish that the peerTG must complete shutdown before the primary
+	// thread group completes shutdown.
+	g.threads.OnStop(func() {
+		err = g.peerTG.Stop()
```
VoidingWarranties: Should be `err := g.peerTG.Stop()`

modules/gateway/rpc.go
```diff
@@ -141,28 +154,27 @@ func (g *Gateway) threadedListenPeer(p *peer) {
...
-		return
-	}
-	defer g.threads.Done()
-
 	for {
 		conn, err := p.accept()
 		if err != nil {
-			g.log.Println("WARN: lost connection to peer", p.NetAddress)
-			return
+			g.log.Debugln("Peer connection closed:", p.NetAddress)
```
VoidingWarranties: Why remove the "WARN"?

VoidingWarranties:
I'm looking forward to being able to use `g.tg.Flush` in testing instead of all those ugly sleeps.

modules/gateway/nodes.go
DavidVorick: Compiler complains without it. Though it seems to be unreachable code, the compiler believes that you need it, so it's needed just to get the build passing.

modules/gateway/upnp.go
DavidVorick: hmm. Not sure I like that setup. But I guess we should leave it for now.

modules/gateway/nodes.go
DavidVorick: looks like I spaced out and forgot to finish the sentence.

modules/gateway/peers.go
DavidVorick: You should be assuming that an exported function is going to call `tg.Add`. There are two places in this PR where manageConnect gets called, and I think one is already under the influence of a `tg.Add`.

modules/gateway/peers.go
DavidVorick: definitely a potential deadlock - right above the call the thread calls `g.threads.Add()`, and inside of the blocking call to `g.RPC` is another call to `g.threads.Add()` - deadlock!

modules/gateway/peers_test.go
DavidVorick: ah jeez, that's really annoying.

DavidVorick: fixed all of the offenders in the gateway, not sure about the other packages though (commit incoming).

modules/gateway/rpc.go
DavidVorick: It's a common thing for peers to disconnect, not really something that you need to be paying attention to unless you are losing most of your peers.

DavidVorick:
In this case `tg.Flush` can't fix a lot of them, because `Flush()` has no bearing over continuous operations such as 'get more peers' or 'purge a set of peers'.

DavidVorick:
OK, this should have all of the tests that I want now. Everything seems to be running smoothly, even in the face of a fully connected network with overloaded nodes.

VoidingWarranties:
I wish GitHub had better reviewing tools. Going through commit by commit is convenient for the reviewer, but I ended up commenting on a bunch of outdated diffs. Some of my comments were invalidated by later commits so I deleted them, but others are still valid but are collapsed by default because it is an outdated diff.

Would you rather I go through and make the comments on the net-diff page so that they don't show as outdated until they are updated?

Maybe in the future it would be best to look through the commits one by one but always comment on the net-diff page.

modules/gateway/peers.go
VoidingWarranties: Ah okay. I was thinking you could get rid of `tg.Add(); defer tg.Done()` in the code below and instead call `Connect`. But then the call to `g.log.Debugln` could panic when the gateway is closing. 

For reference here is the relevant code where `managedConnect` is called in peers.go.
```go
 +		// Try connecting to that peer in a goroutine.
 +		go func() {
 +			if err := g.threads.Add(); err != nil {
 +				return
 +			}
 +			defer g.threads.Done()
 +
 +			err := g.managedConnect(addr)
 +			if err != nil {
 +				g.log.Debugln("WARN: automatic connect failed:", err)
 +			}
 +		}()
```

modules/gateway/peers.go
VoidingWarranties: Ohhhhhh this is the same pattern as the `managedConnect` issue above. Got it.

DavidVorick:
Github could definitely be better in this regard.

I would code review by having the commit-by-commit in one window, and the net-diff in another window. Then I'd use the per-commit to walk through the changes and orient myself, but make any comments on the net-diff. The net-diff would also inform you that something had been changed again before you made the comment.

modules/gateway/gateway_test.go
```diff
@@ -5,24 +5,30 @@ import (
...
+	siasync "github.com/NebulousLabs/Sia/sync"
 )
 
 // newTestingGateway returns a gateway read to use in a testing environment.
 func newTestingGateway(name string, t *testing.T) *Gateway {
 	g, err := New("localhost:0", build.TempDir("gateway", name))
 	if err != nil {
-		t.Fatal(err)
+		// TODO: the proper thing to do here is to return an error and not even
+		// take a `testing.T` as an arguement. Calling t.Fatal is insufficient
```
lukechampine: typo

why not remove the argument then?

DavidVorick: Lots of code would change

modules/gateway/nodes.go
```diff
@@ -43,17 +39,28 @@ func (g *Gateway) removeNode(addr modules.NetAddress) error {
...
-		}
+	if len(g.nodes) == 0 {
+		return "", errNoPeers
 	}
 
+	// Select a random peer. Note that the algorithm below is roughly linear in
+	// the number of nodes known by the gateway, and this number can approach
+	// every node on the network. If the network gets large, this algorithm
+	// will either need to be refactored, or more likely a cap on the size of
+	// g.nodes will need to be added.
```
lukechampine: I bet there's some horrible (i.e. reflection-based) hack that could return a cryptographically random map value in constant time.
At first I thought you could use `reflect.ValueOf(myMap).MapKeys()`, since it returns a slice (which you can randomly index). But in fact this is even slower than what we're doing now, because the [implementation](https://golang.org/src/reflect/value.go?s=32392:32424#L1052) iterates through all the map keys _and_ allocates a new slice for them.
So you'd probably have to do something **really** gross involving unsafe access to the map's buckets :P

modules/gateway/nodes.go
```diff
@@ -90,68 +97,134 @@ func (g *Gateway) requestNodes(conn modules.PeerConn) error {
...
+				select {
+				case <-dialDoneChan:
+				case <-g.threads.StopChan():
+					close(cancelDialChan)
+				}
+			}()
+			dialer := &net.Dialer{
+				Timeout: dialTimeout,
+				Cancel:  cancelDialChan,
+			}
```
lukechampine: this is both ugly enough and useful enough that I think it should be split into a separate function. I think you would want this behavior in `g.Connect` as well, right?

modules/gateway/nodes.go
```diff
@@ -90,68 +97,134 @@ func (g *Gateway) requestNodes(conn modules.PeerConn) error {
...
+				g.log.Debugf("INFO: removing node %q because dialing it failed: %v", node, err)
+				return
 			}
+
+			// If connection succeeds, supply an unacceptable version so that we
+			// will not be added as a peer.
+			//
+			// NOTE: this is a somewhat clunky way of specifying that you didn't
+			// actually want a connection.
+			encoding.WriteObject(conn, "0.0.0")
```
lukechampine: it's also annoying because it results in lots of `"WARN: peer tried to connect, but their version (0.0.0) was unacceptable"` log lines. We'll probably want to support this "ping" behavior next time we change the gateway protocol.

modules/gateway/nodes.go
```diff
@@ -90,68 +97,134 @@ func (g *Gateway) requestNodes(conn modules.PeerConn) error {
...
-		case <-time.After(10 * time.Minute):
-		case <-g.threads.StopChan():
-			return
+		// Determine whether there are a satisfactory number of nodes in the
+		// nodelist. If there are not, use the random peer from earlier to
+		// expand the node list.
+		if numNodes < healthyNodeListLen {
+			// TODO: should this be done in a go func? It's a blocking
+			// function.
+			err := g.managedRPC(peer, "ShareNodes", g.requestNodes)
```
lukechampine: I think the ideal behavior would be to cancel the RPC after a reasonable timeout, but we don't currently support that. Blocking seems slightly better than non-blocking to me because it's easier to reason about and the slowdown isn't a big deal.
However, we have a far bigger problem, which is that our RPCs (or at least `requestNodes`, not sure about others) do not call `SetDeadline` on the conn. That means they can potentially hang indefinitely! That should be fixed in this PR.

modules/gateway/peers.go
```diff
@@ -459,34 +412,55 @@ func (g *Gateway) Connect(addr modules.NetAddress) error {
...
-	} else {
-		err = g.managedConnectNewPeer(conn, remoteVersion, addr)
-	}
+		// Perform peer initialization.
+		remoteVersion, err := connectVersionHandshake(conn, build.Version)
+		if err != nil {
+			conn.Close()
+			return err
+		}
+		if build.VersionCmp(remoteVersion, "1.0.0") < 0 {
```
lukechampine: we might want to make this a constant (`minAcceptableVersion`?)

modules/gateway/peers.go
```diff
@@ -509,84 +483,110 @@ func (g *Gateway) Connect(addr modules.NetAddress) error {
...
+		g.mu.RUnlock()
+		// If there was an error, log the error and then wait a while before
+		// trying again.
+		if err != nil {
+			select {
+			case <-time.After(noPeersDelay):
+			case <-g.threads.StopChan():
+				// Interrupt the thread if the shutdown signal is issued.
+				return
+			}
```
lukechampine: this is a bit verbose, but the alternative:
```go
if interrupted := sleepLightly(noPeersDelay, g.threads.StopChan()); interrupted {
    return
}
```
...feels too clever and obscures what's happening behind the scenes.

modules/gateway/peers.go
```diff
@@ -509,84 +483,110 @@ func (g *Gateway) Connect(addr modules.NetAddress) error {
...
+	if err := g.threads.Add(); err != nil {
+		return err
+	}
+	defer g.threads.Done()
+	return g.managedConnect(addr)
+}
+
+// Disconnect terminates a connection to a peer and removes it from the
+// Gateway's peer list. The peer's address remains in the node list.
+func (g *Gateway) Disconnect(addr modules.NetAddress) error {
```
lukechampine: why do we need a `managedConnect` but not a `managedDisconnect`?

modules/gateway/rpc.go
```diff
@@ -10,8 +10,11 @@ import (
 	"github.com/NebulousLabs/Sia/modules"
 )
 
+// rpcID is an 8-byte signature that is added to all RPCs to tell the gatway
+// what to do with the RPC. Empty elements of rpcID will be encoded as spaces.
```
lukechampine: the encoding comment makes more sense next to the `String` method. It might actually be misleading, because empty bytes are encoded as `0x00` on the wire. The space encoding is only to make log messages more readable.

modules/gateway/nodes.go
DavidVorick: in how many places does this need to be addressed?

DavidVorick: basically anywhere we are getting a conn, we need to call conn.SetDeadline() ?

DavidVorick: between the peers that have connections open all the time, and the RPCs which are individual, I'm not sure which things should have deadlines or how to add them.

modules/gateway/peers.go
DavidVorick: we already have one that's set to "0.4.0"

modules/gateway/peers.go
DavidVorick: we could just make it a method of the `ThreadGroup` :}

modules/gateway/nodes.go
lukechampine: The only `conn` we don't want to set a deadline on is the underlying connection to the peer. When we send or receive an RPC, a new "virtual" `conn` is created by muxado. We want to set a deadline on all of those. `grep` says we are not calling `SetDeadline` _anywhere_ in the gateway o.o

I'm surprised that the gateway is not already doing this, even on the receiving side. It would be really easy: just follow the host's example of setting a generous-but-finite deadline as soon as you accept the conn: https://github.com/NebulousLabs/Sia/blob/master/modules/gateway/rpc.go#L160
This may be a source of some gateway "deadlocks" or inefficiency.

modules/gateway/consts.go
```diff
@@ -7,6 +7,10 @@ import (
 )
 
 const (
+	// handshakeUpgradeVersion is the version where the gateway handshake RPC
+	// was altered to include adiitional information transfer.
```
lukechampine: typo

modules/gateway/nodes.go
```diff
@@ -122,9 +123,13 @@ func (g *Gateway) permanentNodePurger(closeChan chan struct{}) {
 		numNodes := len(g.nodes)
 		node, err := g.randomNode()
 		g.mu.RUnlock()
-		if err != nil {
-			// Most likely the error indicates that there are no nodes in the
-			// node list. Wait until the next iteration to try again.
+		if err == errNoNodes {
+			// errNoNodes is a common error that will be resovled by the
```
lukechampine: typo

modules/gateway/nodes.go
```diff
@@ -135,45 +140,28 @@ func (g *Gateway) permanentNodePurger(closeChan chan struct{}) {
...
-				g.removeNode(node)
-				g.save()
-				g.mu.Unlock()
-				g.log.Debugf("INFO: removing node %q because dialing it failed: %v", node, err)
-				return
-			}
+		// remove them from the node list.
+		conn, err := g.dial(node)
+		if err != nil {
+			// NOTE: an error may be returned if the dial is cancelled
```
lukechampine: "canceled"  is annoyingly the correct spelling, but it looks weird to me so I don't care if you change it

modules/gateway/conn.go
```diff
@@ -17,3 +17,14 @@ type peerConn struct {
...
 	return pc.dialbackAddr
 }
+
+// dial will dial the input address and return a connection. dial appropriately
+// handles things like clean shutdown, fast shutdown, and chooses the correct
+// communication protocol.
+func (g *Gateway) dial(addr modules.NetAddress) (net.Conn, error) {
+	dialer := &net.Dialer{
+		Timeout: dialTimeout,
+		Cancel:  g.threads.StopChan(),
```
lukechampine: whoa, does that work? can't believe I missed that :stuck_out_tongue_closed_eyes:

DavidVorick:
PR should be good to go now, the infinite-RPC is going to have to be another PR.

PR #1423, opened 2016-08-19
Update host API docs

### LINKED ISSUE #718: Break unwiedly api doc into multiple files
The api doc is now very large and should probably be broken up into a separate file per module.

Right now the wallet and consensus are the only ones that are implemented, so this is a bit more of a long term issue.

I'm open to other ideas for breaking up the api doc. Perhaps we could put it on learn.siacoin.com instead of github? Then we can use javascript to make it fancier.

VoidingWarranties:
Closes #718.

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -253,81 +253,79 @@ refer to [Host.md](/doc/api/Host.md).
 
 fetches status information about the host.
 
-// TODO: convert to example JSON response and add units.
 ###### JSON Response [(with comments)](/doc/api/Host.md#json-response)
-```go
-struct {
-	externalsettings {
-		acceptingcontracts   bool
-		maxdownloadbatchsize uint64
-		maxduration          types.BlockHeight (uint64)
-		maxrevisebatchsize   uint64
-		netaddress           modules.NetAddress (string)
-		remainingstorage     uint64
-		sectorsize           uint64
-		totalstorage         uint64
-		unlockhash           types.UnlockHash (string)
-		windowsize           types.BlockHeight (uint64)
+```javascript
+{
+  "externalsettings": {
+    "acceptingcontracts":   true,
+    "maxdownloadbatchsize": 17825792, // bytes
+    "maxduration":          25920,    // blocks
+    "maxrevisebatchsize":   17825792, // bytes
+    "netaddress":           "123.456.789.0:9982",
+    "remainingstorage":     35000000000, // bytes
+    "sectorsize":           4194304,     // bytes
+    "totalstorage":         35000000000, // bytes
+    "unlockhash":           "0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789ab",
+    "windowsize":           144, // blocks
 
-		collateral    types.Currency (string)
-		maxcollateral types.Currency (string)
+    "collateral":    "57870370370",                     // hastings / byte / block
+    "maxcollateral": "100000000000000000000000000000",  // hastings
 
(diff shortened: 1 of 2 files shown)


doc/API.md
```diff
@@ -332,27 +332,26 @@ standard success or error response. See
...
 
 #### /host [GET]
 
-Function: Fetches status information about the host.
-
-Parameters: none
+fetches status information about the host.
 
-Response:
+// TODO: convert to example JSON response and add units.
```
DavidVorick: Dangling TODO

doc/API.md
```diff
@@ -332,27 +332,26 @@ standard success or error response. See
...
+###### JSON Response [(with comments)](/doc/api/Host.md#json-response)
```
DavidVorick: One of the structs here has a docstring, which looks pretty out of place.

doc/API.md
```diff
@@ -493,64 +494,71 @@ Response:
...
+and be unable to provide a storage proof on that sector. This endpoint is for
+removing the data entirely, and will remove instances of the sector appearing
+at all heights. The primary purpose is to comply with legal requests to remove
+data.
 
-Path Parameters
+###### Path Parameters [(with comments)](/doc/api/Host.md#path-parameters)
 ```
-{merkleroot} // Required
+:merkleroot
```
DavidVorick: The leading colon looks intentional but wanted to double check

doc/api/Host.md
```diff
@@ -1,20 +1,42 @@
...
+| [/host/storage/folders/resize](#hoststoragefoldersresize-post)                        | POST      |
+| [/host/storage/sectors/delete/___:merkleroot___](#hoststoragesectorsdeletemerkleroot) | POST      |
+
+#### /host [GET]
 
-Parameters: none
+fetches status information about the host.
 
-Response:
+// TODO: convert to example JSON response and add units.
```
DavidVorick: Dangling TODO

doc/API.md
VoidingWarranties: Yes, want to make it clear that the colon should not be included in the path.

doc/API.md
```diff
@@ -453,27 +452,29 @@ minstorageprice           types.Currency (string) // Optional
 minuploadbandwidthprice   types.Currency (string) // Optional
 ```
```
VoidingWarranties: Go types don't match style of other module's API docs.

doc/api/Host.md
```diff
@@ -1,452 +1,536 @@
...
+
+    // The version of external settings being used. This field helps
+    // coordinate updates while preserving compatibility with older nodes.
+    "version": "1.0.0"
+  },
+
+  // The financial status of the host.
+  "financialmetrics": {
+    // Number of open file contracts.
+    "contractcount": 2,
```
VoidingWarranties: Is `contractcount` the number of open file contracts or the total number of contracts ever?

DavidVorick: I believe that currently it's the total number of contracts ever. Not sure if it's best to leave it that way, but there is the compatibility promise to think about.

VoidingWarranties: For reference the corresponding commit that added this field is 4fa09ed31580ca543ce8ee7df64595e94e426ace

I'm not familiar with the host, but `ContractCount` is decremented in the `removeStorageObligation` method, which to me indicated that it might be only the number of current contracts.

VoidingWarranties: Good to go or no?

PR #1427, opened 2016-08-23
Consensus shutdown

DavidVorick:
I tried to keep the commits clean here. Clean shutdown still is not complete for the consensus set, which means we may still see logging panics after this gets merged. But I went ahead and made the PR because I touched up some of the IBD code while I was digging around.

### DIFF
--- a/modules/consensus.go
+++ b/modules/consensus.go
@@ -111,6 +111,15 @@ type (
 		// consensus set in the recent change.
 		SiafundPoolDiffs []SiafundPoolDiff
 
+		// ChildTarget defines the target of any block that would be the child
+		// of the block most recently appended to the consensus set.
+		ChildTarget types.Target
+
+		// MinimumValidChildTimestamp defines the minimum allowed timestamp for
+		// any block that is the child of the block most recently appended to
+		// the consensus set.
+		MinimumValidChildTimestamp types.Timestamp
+
 		// Synced indicates whether or not the ConsensusSet is synced with its
 		// peers.
 		Synced bool
--- a/modules/consensus/consensusset.go
+++ b/modules/consensus/consensusset.go
@@ -8,7 +8,6 @@ package consensus
 
 import (
 	"errors"
-	"fmt"
 
 	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/encoding"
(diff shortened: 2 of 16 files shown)


modules/consensus/consensusset.go
```diff
@@ -132,13 +132,28 @@ func New(gateway modules.Gateway, persistDir string) (*ConsensusSet, error) {
...
+			// without a goroutine is okay.
+			err = cs.threadedInitialBlockchainDownload()
+			if err != nil {
+				return
+			}
 		}
 
+		// threadedInitialBlockchainDownload will release the threadgroup 'Add'
+		// it was holding, so another needs to be grabbed to finish off this
+		// goroutine.
```
lukechampine: is there a race condition if `Stop` is called between ibd releasing the `Add` and the new `Add`?

hmm, looks like it's harmless; it would only prevent RPCs from being registered. Although with `Flush` it might be more complicated, right?

modules/consensus/consensusset.go
```diff
@@ -197,6 +196,13 @@ func (cs *ConsensusSet) BlockAtHeight(height types.BlockHeight) (block types.Blo
 
 // ChildTarget returns the target for the child of a block.
 func (cs *ConsensusSet) ChildTarget(id types.BlockID) (target types.Target, exists bool) {
+	// A call to a close database can cause undefined behavior.
```
lukechampine: close*d*

modules/consensus/synchronize.go
```diff
@@ -553,7 +553,7 @@ func (cs *ConsensusSet) threadedInitialBlockchainDownload() error {
 		//
 		// It is possible that our peers are malicous and will lie to us about
 		// being synced, which is why we take the majority after 10 minutes.
-		if numOutboundSynced >= minNumOutbound || (numOutboundSynced > numOutboundNotSynced && time.Now().After(deadline)) {
+		if numOutboundSynced > numOutboundNotSynced && (numOutboundSynced >= minNumOutbound || time.Now().After(deadline)) {
```
lukechampine: need to update the comment

lukechampine:
solid commits, easy to review :+1:

modules/consensus/consensusset.go
DavidVorick: Shouldn't be. If Shutdown is called early, the RPCs will never be registered and then never cleaned up. If Flush is called, the RPCs may not be registered upon completion.

If you want control over waiting until the RPCs are registered, you'll need a separate thread group (or maybe just a WaitGroup). IBD can be a very long process, it does not make sense to block shutdown until IBD has completed.

Maybe it does, if you add in some intelligent clean-shutdown code.

modules/consensus/synchronize.go
```diff
@@ -497,34 +505,55 @@ func (cs *ConsensusSet) threadedInitialBlockchainDownload() {
...
-		// we are synced until we've been doing ibd for 10 minutes and we are synced
-		// with at least one peer.
-		if numOutboundSynced >= minNumOutbound || (numOutboundSynced > 0 && time.Now().After(deadline)) {
+		// If we have minNumOutbound peers synced, we are done. Otherwise,
+		// don't say we are synced until we've been doing ibd for 10 minutes
+		// and we have strictly more synced peers than not synced peers.
+		//
+		// It is possible that our peers are malicous and will lie to us about
+		// being synced, which is why we take the majority after 10 minutes.
+		if numOutboundSynced > numOutboundNotSynced && (numOutboundSynced >= minNumOutbound || time.Now().After(deadline)) {
```
lukechampine: need to update comment

DavidVorick:
I have finished with clean shutdown. There seems to be a NDF in one of the synchronize tests still, and there's also an NDF related to occasional boltdb faults that is still being investigated.

modules/consensus.go
```diff
@@ -111,6 +111,15 @@ type (
...
 		SiafundPoolDiffs []SiafundPoolDiff
 
+		// ChildTarget defines the target of any block that would be the child
+		// of the block most recently appended to the consensus set.
+		ChildTarget types.Target
+
+		// MinimumValidChildTimestamp defines the minimum allowed timestamp for
+		// any block that is the child of the block most recently appended to
+		// the consensus set.
+		MinimumValidChildTimestamp types.Timestamp
```
lukechampine: does this break any compatibility in the consensus persistence?

DavidVorick: It doesn't, the consensus change fields don't get saved to disk anywhere.

PR #1428, opened 2016-08-24
Gateway tweaks

DavidVorick:
Overall, I was looking to solve the problem where gateways on the same machine were unable to connect to eachother except during testing.

That problem should be approximately solved now, though the build is still failing.

This is under-tested, but a bunch of other small safety features have been added. There is still some work to go.

### DIFF
--- a/modules/gateway/peers.go
+++ b/modules/gateway/peers.go
@@ -529,7 +529,7 @@ func (g *Gateway) permanentPeerManager(closedChan chan struct{}) {
 		// outbound peers to be local peers.
 		if err != nil || (numOutboundPeers > 0 && addr.IsLocal() && build.Release != "testing") {
 			select {
-			case <-time.After(noPeersDelay):
+			case <-time.After(noNodesDelay):
 			case <-g.threads.StopChan():
 				// Interrupt the thread if the shutdown signal is issued.
 				return
@@ -551,6 +551,13 @@ func (g *Gateway) permanentPeerManager(closedChan chan struct{}) {
 				// the outbound peers instead of allow the attacker to pick out
 				// the peers for us. Because we have made the selection, it is
 				// okay to set the peer as an outbound peer.
+				//
+				// The nodelist size check ensures that an attacker can't flood
+				// a new node with a bunch of inbound requests. Doing so would
+				// result in a nodelist that's entirely full of attacker nodes.
+				// There's not much we can do about that anyway, but at least
+				// we can hold off making attacker nodes 'outbound' peers until
+				// our nodelist has had time to fill up naturally.
 				g.mu.Lock()
 				p, exists := g.peers[addr]
 				if exists {
--- a/modules/gateway/peers.go
+++ b/modules/gateway/peers.go
@@ -192,33 +192,31 @@ func (g *Gateway) managedAcceptConnNewPeer(conn net.Conn, remoteVersion string) 
 		return err
 	}
 
-	err = func() error {
-		g.mu.Lock()
-		defer g.mu.Unlock()
-
-		// Don't accept a connection from a peer we're already connected to.
-		if _, exists := g.peers[remoteAddr]; exists {
-			return fmt.Errorf("already connected to a peer on that address: %v", remoteAddr)
-		}
-
-		// Accept the peer.
-		g.acceptPeer(&peer{
-			Peer: modules.Peer{
-				NetAddress: remoteAddr,
-				Inbound:    true,
-				Version:    remoteVersion,
-			},
-			sess: muxado.Server(conn),
-		})
-		return nil
-	}()
-	if err != nil {
-		return err
+	// Attempt to add the peer to the node list. If the add is successful and
+	// the address is a local address, mark the peer as a local peer.
+	local := false
+	err = g.managedAddUntrustedNode(remoteAddr)
+	if err != nil && remoteAddr.IsLocal() {
+		local = true
 	}
 
-	// Ignore the error from adding the untrusted node, does not matter either
-	// way.
-	g.managedAddUntrustedNode(remoteAddr)
+	g.mu.Lock()
+	defer g.mu.Unlock()
+
+	// Don't accept a connection from a peer we're already connected to.
+	if _, exists := g.peers[remoteAddr]; exists {
+		return fmt.Errorf("already connected to a peer on that address: %v", remoteAddr)
+	}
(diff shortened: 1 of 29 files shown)


DavidVorick:
added some more stuff. At this point it's closer to completion, although the testing is still not there. Also, the build still isn't passing due to the gateway being more effective at finding peers (this is disrupting the consensus tests, which try to set up gateways with no outbound peers, but the gateways will now find outbound peers if they have peers at all)

DavidVorick:
ok, everything should be working now.

There's still only a minimum of testing, though the most important tests I think are covered. I'd like some more time to look at it before deciding that everything is working. Setting things up on the ant-farm will help us further appraise the strength of this code.

DavidVorick:
Just added two commits which should result in the  gateway and the consensus set doing bootstrapping correctly.

modules/gateway/peers.go
```diff
@@ -184,62 +177,79 @@ func (g *Gateway) managedAcceptConnNewPeer(conn net.Conn, remoteVersion string)
 		return err
 	}
 
+	// Attempt to add the peer to the node list. If the add is successful and
+	// the address is a local address, mark the peer as a local peer.
+	local := false
+	err = g.managedAddUntrustedNode(remoteAddr)
+	if err != nil && remoteAddr.IsLocal() {
+		local = true
```
lukechampine: why does `local` depend on `err != nil`?

modules/gateway/peers.go
```diff
@@ -184,62 +177,79 @@ func (g *Gateway) managedAcceptConnNewPeer(conn net.Conn, remoteVersion string)
...
+			continue
 		}
-		// if another peer shares this IP, choose that one instead
-		for addr := range g.peers {
-			if addr.Host() == p.NetAddress.Host() {
-				kick = addr
-				break
-			}
+		if p.Local {
+			continue
```
lukechampine: `if !p.Inbound || p.Local`?

modules/netaddress.go
```diff
@@ -60,32 +61,83 @@ func (na NetAddress) IsLoopback() bool {
...
+
+// IsStdValid returns an error if the NetAddress is invalid. A valid NetAddress
 // is of the form "host:port", such that "host" is either a valid IPv4/IPv6
 // address or a valid hostname, and "port" is an integer in the range
-// [1,65535]. Furthermore, "host" may not be a loopback address (except during
-// testing). Valid IPv4 addresses, IPv6 addresses, and hostnames are detailed
+// [1,65535]. Valid IPv4 addresses, IPv6 addresses, and hostnames are detailed
 // in RFCs 791, 2460, and 952, respectively.
-func (na NetAddress) IsValid() error {
+func (na NetAddress) IsStdValid() error {
```
lukechampine: not a very helpful name. Looking forward to this being phased out

modules/netaddress.go
```diff
@@ -60,32 +61,83 @@ func (na NetAddress) IsLoopback() bool {
...
+	range2High := net.ParseIP("172.31.255.255").To16()
+	range3Low := net.ParseIP("192.168.0.0").To16()
+	range3High := net.ParseIP("192.168.255.255").To16()
+	range4Low := net.ParseIP("fd00:0000:0000:0000:0000:0000:0000:0000")
+	range4High := net.ParseIP("fdff:ffff:ffff:ffff:ffff:ffff:ffff:ffff")
+
+	// Return true if ip16 falls between any of the above defined ranges.
+	if bytes.Compare(range1Low, ip16) <= 0 && bytes.Compare(ip16, range1High) <= 0 {
+		return true
+	}
```
lukechampine: where is this code from?

lukechampine: I think you want something like this:
```go
localCIDRs := []string{
    "10.0.0.0/8",
    "192.168.0.0/16",
    // ... etc.
}
for _, cidr := range localCIDRs {
    _, ipnet, _ := net.ParseCIDR(cidr)
    if ipnet.Contains(ip) {
        return true
    }
}
return false
```

modules/gateway/peers.go
DavidVorick: I'm pretty sure that you can lie about what your remote address is. If you say that you are a local node, you get special treatment in terms of being kicked. Don't want to do that without verifying that you are reachable at the specified address.

modules/netaddress.go
DavidVorick: Yeah, `IsStdValid` will be renamed back to Valid.

modules/netaddress.go
DavidVorick: yeah, passes all of the tests as well.

modules/gateway/peers.go
VoidingWarranties: I don't know of any way of manipulating your remote address over TCP. Is this really possible?

lukechampine: According to [this answer](http://security.stackexchange.com/questions/37481/is-it-possible-to-pass-tcp-handshake-with-spoofed-ip-address), you can only get away with spoofing your IP if you control a router "close to the target." Of course, if you control such a router, you can probably do much worse things than that. But there's no harm in guarding against it unless it severely impacts other code.

VoidingWarranties:
A few questions...

1. Why is it desirable to connect multiple nodes on the same machine over the loopback address? Is this only for the antfarm?
2. Why kick local peers? More generally, why do we treat local peers differently from non-local peers? It seems like this makes testing the gateway's peer behavior difficult on the antfarm.

DavidVorick:
We don't kick local peers, we protect them.

```
		// Do not kick outbound peers or local peers.
		if !p.Inbound || p.Local {
			continue
		}

		// Prefer kicking a peer with the same hostname.
		if addr.Host() == p.NetAddress.Host() {
			addrs = []modules.NetAddress{addr}
			break
		}
		addrs = append(addrs, addr)
```

'same hostname' -> same hostname as the inbound peer, not same hostname as yourself.

DavidVorick:
You want multiple nodes to connect to eachother on the same machine for testing purposes. It's not just the antfarm, any sort of testing that you are doing on a single machine is going to require connecting a bunch of nodes on the same machine together.

There's no practical reason to stop that from happening. The limits that we put on the gateway are primarily due to concerns around bandwidth bottlenecks (e.g. broadcasting a 2MB block to 128 peers). When you are communicating over the loopback, it's not going to be a problem.

PR #1429, opened 2016-08-25
Move last of wallet persistence into DB

lukechampine:
In addition, preloading has been removed.


DavidVorick:
Does that mean that the wallet is now performing instant-unlock?

lukechampine:
Almost. The wallet was never storing a `ConsensusChangeID`, so I need to add that to the database. I could add that to this PR; shouldn't be much code

DavidVorick:
seems good. Will be fun to see a working example of an instant-unlock wallet.

siac/walletcmd.go
```diff
@@ -385,7 +385,6 @@ func walletunlockcmd() {
 	if err != nil {
 		die("Reading password failed:", err)
 	}
-	fmt.Println("Unlocking the wallet. This may take several minutes...")
```
johnathanhowell: :tada: 

I'll remove this warning from the UI too.

DavidVorick: wallet is not merged yet

DavidVorick: (this is not a PR onto the master branch)

johnathanhowell: Ah yeah.  I'll remove the warning once the wallet is fully merged.

modules/host/persist.go
```diff
@@ -95,10 +95,7 @@ func (h *Host) establishDefaults() error {
 		return err
 	}
 	h.secretKey = sk
-	h.publicKey = types.SiaPublicKey{
-		Algorithm: types.SignatureEd25519,
-		Key:       pk[:],
-	}
+	h.publicKey = types.Ed25519PublicKey(pk)
```
DavidVorick: Interesting. I guess this is good, though the types package now has a crapload of functions.

But... probably not a bad thing if they are all well documented. It is the best starting place for learning the Sia codebase.

modules/wallet/encrypt.go
```diff
@@ -26,129 +32,244 @@ func uidEncryptionKey(masterKey crypto.TwofishKey, uid uniqueID) crypto.TwofishK
...
-					done := w.subscribed
-					w.mu.RUnlock()
-					if done {
-						println("\nDone!")
-						break
+		err = w.cs.ConsensusSetSubscribe(w, lastChange)
+		if err == modules.ErrInvalidConsensusChangeID {
+			// something went wrong; resubscribe from the beginning and spawn a
+			// goroutine to display rescan progress
+			if build.Release != "testing" {
```
DavidVorick: weird. I realize it's been here the whole time, but I'm not sure why.

DavidVorick: oh nvm, it only wraps the println code

DavidVorick:
Overall looks pretty good, doesn't seem like anything too dramatic is happening here.

PR #1430, opened 2016-08-27
add gateway timeouts

DavidVorick:


DavidVorick:
I also cleaned up all TODOs in the gateway and added golint. I think all of the major things have been addressed, all that remains is more documentation.

modules/gateway/conn.go
```diff
@@ -26,5 +27,10 @@ func (g *Gateway) dial(addr modules.NetAddress) (net.Conn, error) {
 		Timeout: dialTimeout,
 		Cancel:  g.threads.StopChan(),
 	}
-	return dialer.Dial("tcp", string(addr))
+	conn, err := dialer.Dial("tcp", string(addr))
+	if err != nil {
+		return nil, err
+	}
+	conn.SetDeadline(time.Now().Add(connStdDeadline))
```
VoidingWarranties: Is it safe / necessary to set a deadline for a connection that will be used to create a muxado session?

VoidingWarranties:
A small nitpick with the commits. It makes reviewing a little cumbersome if there are frequent commits that fix previous mistakes (bd21891469d90dc182c8a73f8ead824ae35db0e1 and 431526cb2a199bdad5abb8f05dbb6e11773e270e). Could you do an amend or fixup instead?

DavidVorick:
I amend them if they are the most recent commit, but since we are doing signing I try as much as possible not to rebase, because it destroys the signatures

lukechampine:
why is it a problem to overwrite a signature?

modules/gateway/conn.go
VoidingWarranties: Oops, I missed the `conn.SetDeadline(time.Time{})` added in another commit to reset the deadline in `Connect`.

build/critical.go
```diff
@@ -6,13 +6,28 @@ import (
 	"runtime/debug"
 )
 
-// Critical will print a message to os.Stderr unless DEBUG has been set, in
-// which case panic will be called instead.
+// Critical should be called if a sanity check has failed, indicating developer
+// error. Severe is called with an extended message guiding the user to the
+// issue tracker on Github. If the program does not panic, the call stack for
+// the running goroutine is printed to help determine the error.
```
lukechampine: "Severe" -> "Critical"?

build/critical.go
```diff
@@ -6,13 +6,28 @@ import (
...
+	os.Stderr.WriteString(s)
 	if DEBUG {
 		panic(s)
 	}
-	os.Stderr.WriteString(s)
 	debug.PrintStack()
 }
+
+// Severe will print a message to os.Stderr unless DEBUG has been set, in which
+// case panic will be called instead. Severe should be called in situations
```
lukechampine: not instead; both will happen

modules/gateway.go
```diff
@@ -17,15 +19,26 @@ var (
...
+				"68.55.10.144:9981",
+				"109.206.33.225:9981",
+				"79.51.183.132:9981",
+				"176.9.59.110:9981",
+			}
+		case "testing":
+			return nil
+		default:
+			panic("unrecognized build.Release constant in BootstrapPeers")
+		}
```
lukechampine: wondering if a helper like this would be worth it:
```go
type Var struct {
    Dev, Standard, Testing interface{}
}

func Select(v Var) interface{} {
    switch Release {
    case "dev":
        return v.Dev
    case "standard":
        return v.Standard
    case "testing":
        return v.Testing
    default:
        panic("unknown Release constant: " + Release)
    }
}
```
```go
var BootstrapPeers = build.Select(build.Var{
    Dev:      nil,
    Standard: []NetAddress{"foo"},
    Testing:  nil,
}).([]NetAddress)
```

Unfortunately, the above code has a bug: If you run it with `build.Release == "testing"`, it will panic with the message `interface conversion: interface is nil, not []NetAddress`. So the `nil`s need to be replaced with `[]NetAddress(nil)` or `[]NetAddress{}`. This is only a problem with slices/pointers/maps/channels, which are rarely used as build vars.

modules/gateway/gateway.go
```diff
@@ -1,5 +1,115 @@
...
+// form connections with other nodes, and then uses those connections to
+// participate in the flood network. The primary vector for an attacker to
+// achieve an eclipse attack is node list domination. If a gateway's nodelist
+// is heavily dominated by attacking nodes, then when the gateway chooses to
+// make random connections the gateway is at risk of selecting only attacker
+// nodes.
+//
+// The gateway defends itself from these attacks by minimizing the amount of
+// control that an attacker has over the node list and peer list. The first
+// major defense is that the gateway maintais 8 'outbound' relationships, which
```
lukechampine: typo

modules/gateway/gateway.go
```diff
@@ -1,5 +1,115 @@
...
+// also be attacker nodes, but they are less likely to be attacker nodes
+// because you chose them, instead of them choosing you.
+//
+// If the gateway forms too many connections, the gateway will allow incoming
+// connections by kicking an existing peer. But, to limit the amount of control
+// that an attacker may have, only inbound peers are selected to be kicked.
+// Furthermore, to increase the difficulty of attack, if a new inbound
+// connection shares the same IP address as an existing connection, the shared
+// connection is the connection that gets dropped (unless that connection is an
+// inbound connection).
```
lukechampine: is "inbound" correct here? Seems like you might have meant "local" or "outbound" instead

modules/gateway/gateway.go
```diff
@@ -1,5 +1,115 @@
...
+// after they successfully form a connection with the gateway. To limit the
+// attacker's ability to add nodes to the nodelist, connections are
+// ratelimited. An attacker with lots of IP addresses still has the ability to
+// fill up the nodelist, however getting 90% dominance of the nodelist requries
+// forming thousands of connections, which will take hours or days. By that
+// time, the attacked node should already have its set of outbound peers,
+// limiting the amount of damage that the attacker can do.
+//
+// To limit DNS-based tomfoolry, nodes are only added to the nodelist if their
+// connection information takes the form of an IP address.
```
lukechampine: tomfoolery lol

modules/gateway/gateway.go
```diff
@@ -1,5 +1,115 @@
...
+//
+// Some research has been done on Bitcoin's flood networks. The more relevant
+// research has been listed below. The papers listed first are more relevant.
+//     Eclipse Attacks on Bitcoin's Peer-to-Peer Network (Heilman, Kendler, Zohar, Goldberg)
+//     Stubborn Mining: Generalizing Selfish Mining and Combining with an Eclipse Attack (Nayak, Kumar, Miller, Shi)
+//     An Overview of BGP Hijacking (https://www.bishopfox.com/blog/2015/08/an-overview-of-bgp-hijacking/)
```
lukechampine: solid writeup :+1:

modules/gateway/nodes.go
```diff
@@ -102,33 +102,54 @@ func (g *Gateway) randomNode() (modules.NetAddress, error) {
...
+				break
+			}
+		}
+	}()
 	return encoding.WriteObject(conn, nodes)
 }
 
 // requestNodes is the calling end of the ShareNodes RPC.
 func (g *Gateway) requestNodes(conn modules.PeerConn) error {
+	conn.SetDeadline(time.Now().Add(connStdDeadline))
```
lukechampine: in renter/proto I use a helper function:
```go
func extendDeadline(conn net.Conn, d time.Duration) { _ = conn.SetDeadline(time.Now().Add(d)) }
```
This also takes care of `errcheck` warnings

DavidVorick:
fixed the comments, should be good to go now.

PR #1431, opened 2016-08-28
Fix API path parameter docs

VoidingWarranties:
to match httprouter style specification
e.g. `/foo/*bar` and `/foo/:bar`


lukechampine:
LGTM

PR #1432, opened 2016-09-08
Host port fixing

DavidVorick:
added some logs, and set it up so that it only clears the port if the port was forwarded successfully in the first place.

### DIFF
--- a/modules/host/upnp.go
+++ b/modules/host/upnp.go
@@ -106,7 +106,7 @@ func (h *Host) managedForwardPort(port string) error {
 
 	d, err := upnp.Discover()
 	if err != nil {
-		h.log.Printf("WARN: could not automatically forward port %s: no UPnP-enabled devices found: %v", port, err)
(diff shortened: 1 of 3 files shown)


modules/host/upnp.go
```diff
@@ -99,20 +99,19 @@ func (h *Host) managedForwardPort() error {
...
-	port := h.port
-	h.mu.RUnlock()
 	portInt, err := strconv.Atoi(port)
 	if err != nil {
 		return err
 	}
 
 	d, err := upnp.Discover()
 	if err != nil {
+		h.log.Printf("WARN: could not automatically forward port %s: no UPnP-enabled devices found: %v", port, err)
```
lukechampine: `Discover` already returns an error message that says `"no UPnP-enabled gateway found", so this log might read weird. Just printing the error should be ok

DavidVorick:
ok, fixed

PR #1433, opened 2016-09-09
Seed initialization and sweeping

lukechampine:
This PR adds seed initialization (creating a new wallet from an existing seed) and seed sweeping (scanning the blockchain for outputs generated by a seed, and transferring them to the wallet).

`InitFromSeed` works by scanning the blockchain for uses of the seed. It records the highest seedIndex seen, and sets the wallet's `PrimarySeedProgress` to that value (plus a 10% fudge factor). Note that this means the blockchain must be fully synced before calling `InitFromSeed`. When the wallet is unlocked for the first time, another full rescan will be triggered so that the wallet can record all its outputs, processed txns, etc. (Obviously there is room for an optimization here; we should only need to scan the blockchain once.) Scanning the blockchain takes a while, and there is currently no visual indication that `InitFromSeed` is making progress. We should at least add terminal output similar to the unlock output. `InitFromSeed` also needs a test (whoops).

`SweepSeed` similarly scans the blockchain for uses of the seed. It then creates a new transaction, containing inputs for each seen output, and an output belonging to the wallet. Outputs less than the "dustThreshold" are discarded. Currently, siacoin outputs and miner payouts are supported, but not siafunds. (I didn't trust myself to get the code right with those.) As with `InitFromSeed`, the blockchain must be fully synced and the process can take a while.


modules/wallet/scan.go
```diff
@@ -0,0 +1,195 @@
...
+		if err := cs.ConsensusSetSubscribe(s, modules.ConsensusChangeBeginning); err != nil {
+			return err
+		}
+		if s.largestIndexSeen < s.numKeys()/2 {
+			cs.Unsubscribe(s)
+			return nil
+		}
+		// double number of keys generated each iteration, capping so that we
+		// do not exceed maxScanKeys
+		numKeys *= 2
```
lukechampine: as discussed IRL, this should be changed to multiply by 4. Otherwise, because we check for the upper half, we would check every newly-generated key.

modules/wallet/seed.go
```diff
@@ -216,3 +241,79 @@ func (w *Wallet) LoadSeed(masterKey crypto.TwofishKey, seed modules.Seed) error
...
+		txn.SiacoinInputs = append(txn.SiacoinInputs, types.SiacoinInput{
+			ParentID:         types.SiacoinOutputID(output.id),
+			UnlockConditions: sk.UnlockConditions, // TODO: check timelock, etc.
+		})
+		// add a signature for the input
+		swept = swept.Add(output.value)
+	}
+
+	estTxnSize := len(s.siacoinOutputs) * 350 // since we use maxFee, lowballing is ok
+	estFee := maxFee.Mul64(uint64(estTxnSize))
```
mnsl: The same magic number (350) is being used in two places, this time without explanation.

Also, seems like `estFee` should be renamed to `sweepFee`, since it's the actual fee that the wallet is charged for performing the sweep.  Or am I missing something?

modules/wallet/seed.go
```diff
@@ -216,3 +241,79 @@ func (w *Wallet) LoadSeed(masterKey crypto.TwofishKey, seed modules.Seed) error
 	defer w.mu.Unlock()
 	return w.loadSeed(masterKey, seed)
 }
+
+// SweepSeed scans the blockchain for outputs generated from seed and create a
+// transaction that transfers them to the wallet. Note that this incurs a
+// transaction fee. It returns the total value of the outputs, minus the fee.
+// TODO: support SiafundOutputs too
+func (w *Wallet) SweepSeed(masterKey crypto.TwofishKey, seed modules.Seed) (payout types.Currency, err error) {
```
mnsl: `masterKey` is never used in `SweepSeed`

modules/wallet/seed.go
```diff
@@ -42,6 +44,27 @@ func generateSpendableKey(seed modules.Seed, index uint64) spendableKey {
...
 }
 
+// generateKeys generates n keys from seed, starting from index start.
+func generateKeys(seed modules.Seed, start, n uint64) []spendableKey {
+	// generate in parallel, one goroutine per core.
+	keys := make([]spendableKey, n)
+	var wg sync.WaitGroup
+	wg.Add(runtime.NumCPU())
+	for cpu := 0; cpu < runtime.NumCPU(); cpu++ {
+		go func(start uint64) {
```
mnsl: Perhaps rename this local parameter `start` to `offset` to avoid confusion with the outer function's `start` parameter.

modules/wallet/scan.go
```diff
@@ -0,0 +1,195 @@
...
+				// if a seed output is spent, remove it from the output set
+				id := types.OutputID(sci.ParentID)
+				if _, exists := s.siacoinOutputs[id]; exists {
+					delete(s.siacoinOutputs, id)
+				}
+			}
+		}
+	}
+
+	// update largestIndexSeen
```
mnsl: `largestIndexSeen` -> `s.largestIndexSeen`

DavidVorick:
I don't remember closing this, even by accident, and eve worse it's not letting me reopen it?

Did I accidentally delete the branch it was being PR'd into? I think I may have.

DavidVorick:
ok, restored

DavidVorick:
hmm. Restore may not have all the most recent commits? Were the merge conflicts there before? @lukechampine terribly sorry, hoping you've got the means to fix this easily.

lukechampine:
weird, for some reason the `wallet` branch was outdated. I restored it from my local copy.

modules/wallet/seed.go
lukechampine: ah, I see what you mean. I used "est" to indicate that it was derived from the estimated size of the transaction. It could be renamed to "sweepFee" or "txnFee" but I don't think it matters much.

DavidVorick:
What happened what I accidentally deleted the `wallet` branch on github, which auto-closed this PR. I restored the wallet from my local machine but it had been long enough since my last `git fetch` that I didn't have all of the commits.

modules/wallet/seed.go
DavidVorick: I think it's worth changing, you want it to be clear what you are doing.

modules/wallet/encrypt.go
```diff
@@ -52,30 +52,15 @@ func checkMasterKey(tx *bolt.Tx, masterKey crypto.TwofishKey) error {
...
 		// Check if the wallet encryption key has already been set.
 		if wb.Get(keyEncryptionVerification) != nil {
 			return errReencrypt
 		}
 
-		// If the input key is blank, use the hash of the seed to create the
-		// master key. Otherwise, use the input key.
-		if masterKey == (crypto.TwofishKey{}) {
-			masterKey = crypto.TwofishKey(crypto.HashObject(seed))
-		}
```
DavidVorick: What's up with skipping this step?

DavidVorick: oh I see you moved the code.

To me it makes it feel like it's more likely that initEncryption could be used incorrectly. hmm

modules/wallet/encrypt.go
```diff
@@ -311,23 +296,67 @@ func (w *Wallet) Encrypted() bool {
...
+// synced.
+func (w *Wallet) InitFromSeed(masterKey crypto.TwofishKey, seed modules.Seed) error {
+	if err := w.tg.Add(); err != nil {
+		return err
+	}
+	defer w.tg.Done()
+
+	if !w.cs.Synced() {
+		return errors.New("cannot init from seed until blockchain is synced")
+	}
```
DavidVorick: We should write a test around it, but I don't think that this would actually be necessary. As new outputs come in, all of the seed addresses will be checked.

There are two different concerns though. The first is that I'm not sure how the locking will work, seems to me like rescans tend to be super slow during IBD. Like, they will progress at a slower rate than the rate at which the chain is being extended, so a rescan can take many hours. Second, I'm not sure what would happen if you unsubscribed halfway through IBD, though I think it would be okay.

DavidVorick: From a UX perspective, I think we would want to let people init their seeds before syncing completes. The rescan makes this difficult though.

DavidVorick: I think most of the slowness is derived from the fact that the consensus set is pretty much continuously under lock during IBD, and when rescanning modules like the hostdb need to requently call consensus-locking functions, meaning things get slowed down a lot by all of the contention.

DavidVorick: I guess it would be worth profiling. Ideally the ConsensusChange would contain all the data you would potentially need already (since it's derived directly from a processed block, which is pretty much the most complete thing you can have), that way you don't need to ask the consensus set for more info you can just pull it from the CC

modules/wallet/persist_legacy.go
```diff
@@ -1,177 +0,0 @@
...
-// new wallet that contains all of the addresses from each. This is useful for
-// loading backups.
-func (w *Wallet) MergeWallet(filepath string) error {
-	err := w.loadWallet(filepath)
-	if err != nil {
-		return err
-	}
-	return w.save()
-}
-*/
```
DavidVorick: I don't think you should be dropping this file, it's a good template for the functions that we need to support for compatibility reasons when we make the release. We need to be able to support the loading of old wallet files, even dating back to pre 0.3.3.x, because I know there are at least a handful of users who still have coins in those files. The wallet seed wasn't introduced until 0.4.0.

modules/wallet/scan.go
```diff
@@ -0,0 +1,198 @@
...
+	case "testing":
+		return 1e5
+	default:
+		panic("unrecognized build.Release")
+	}
+}()
+
+var errMaxKeys = fmt.Errorf("refused to generate more than %v keys from seed", maxScanKeys)
+
+type scannedOutput struct {
```
DavidVorick: docstring

modules/wallet/scan.go
```diff
@@ -0,0 +1,198 @@
...
+	value     types.Currency
+	seedIndex uint64
+}
+
+// A seedScanner scans the blockchain for addresses that belong to a given
+// seed.
+type seedScanner struct {
+	blockheight      types.BlockHeight
+	dustThreshold    types.Currency              // minimum value of outputs to be included
+	keys             map[types.UnlockHash]uint64 // map address to seed index
```
DavidVorick: haven't looked at much of the rest of the code, but is it necessary to store the whole unlock hash when doing the scan? That's going to add something like 60MB to the memory-cost of this operation for 1e6 seeds, and 6GB of memory cost for 100e6 seeds.

Because the settings go all the way up to 100e6 seeds, the memory usage needs to be thought through a bit more. Even at 1e6 seeds it's going to be an expensive operation.

DavidVorick: Thinking a little more, we're probably going to want to benchmark this stuff on a full Sia instance. We can create a wallet that puts outputs at every 500e3 places and then see how expensive the rescan is for each retry. I'm guessing it becomes intolerable pretty quickly. (maybe around 10e6)

modules/wallet/scan.go
```diff
@@ -0,0 +1,198 @@
...
+func (s *seedScanner) generateKeys(n uint64) {
+	initialProgress := s.numKeys()
+	for i, k := range generateKeys(s.seed, initialProgress, n) {
+		s.keys[k.UnlockConditions.UnlockHash()] = initialProgress + uint64(i)
+	}
+}
+
+// ProcessConsensusChange scans the blockchain for information relevant to the
+// seedScanner.
+func (s *seedScanner) ProcessConsensusChange(cc modules.ConsensusChange) {
```
DavidVorick: Does this mean that it does not interfere at all with the wallet's PCC? That's an interesting choice, and it makes me worry that the seed scanner could either get ahead of the wallet or fall behind the wallet.

If a block happens between the seed scanner doing its rescan and the wallet receiving the outputs, you may actually miss money. I'm not fully sure how this is set up but it seems like a pitfall.

modules/wallet/scan.go
```diff
@@ -0,0 +1,198 @@
...
+	s.blockheight -= types.BlockHeight(len(cc.RevertedBlocks))
+	// update outputs
```
DavidVorick: aren't you going to have problems if you save an output that got reverted and then double-spent? A lot of user wallets are going to have double spends in them because the wallet for a long time would not remember which outputs it had already used if you restarted it. In fact I'm not even sure 1.0.0 solved that issue.

modules/wallet/scan.go
```diff
@@ -0,0 +1,198 @@
...
+			// mandatory if txn0 can spend an output created in txn1
+			for _, sci := range txn.SiacoinInputs {
+				// if a seed output is spent, remove it from the output set
+				id := types.OutputID(sci.ParentID)
+				if _, exists := s.siacoinOutputs[id]; exists {
+					delete(s.siacoinOutputs, id)
+				}
+			}
+		}
+	}
```
DavidVorick: I don't think you need to be looking at the actual blocks to achieve this, I also don't think you need to be tracking the miner outputs yourself. Because doing this stuff has a lot of surface area for bugs, it's all handled in the ConsensusChange object for you. You can just iterate through `cc.SiacoinOutputDiffs` and `cc.SiafundOutputDiffs`, ignoring the blocks altogether.

modules/wallet/scan.go
```diff
@@ -0,0 +1,198 @@
...
+			for _, sco := range txn.SiacoinOutputs {
+				addrs = append(addrs, sco.UnlockHash)
+			}
+			for _, sfi := range txn.SiafundInputs {
+				addrs = append(addrs, sfi.UnlockConditions.UnlockHash())
+			}
+			for _, sfo := range txn.SiafundOutputs {
+				addrs = append(addrs, sfo.UnlockHash)
+			}
+		}
```
DavidVorick: all of these unlock hashes should be appearing in cc.SiacoinOutputDiffs, I don't think you need to do this extra scanning.

modules/wallet/scan.go
```diff
@@ -0,0 +1,198 @@
...
+		numKeys *= 4
+		if numKeys > maxScanKeys-s.numKeys() {
+			numKeys = maxScanKeys - s.numKeys()
+		}
+	}
+	cs.Unsubscribe(s)
+	return errMaxKeys
+}
+
+func newSeedScanner(seed modules.Seed) *seedScanner {
```
DavidVorick: docstring

modules/wallet/scan.go
```diff
@@ -0,0 +1,198 @@
...
+			return nil
+		}
+		// quadruple number of keys generated each iteration, capping so that we
+		// do not exceed maxScanKeys
+		numKeys *= 4
+		if numKeys > maxScanKeys-s.numKeys() {
+			numKeys = maxScanKeys - s.numKeys()
+		}
+	}
+	cs.Unsubscribe(s)
```
DavidVorick: I am pretty confident that using a separate subscription from the wallet is going to cause headaches and missed outputs at least a few times for the users. The wallet and the scanner need to be on the same page.

modules/wallet/seed.go
```diff
@@ -42,6 +44,27 @@ func generateSpendableKey(seed modules.Seed, index uint64) spendableKey {
...
+	var wg sync.WaitGroup
+	wg.Add(runtime.NumCPU())
+	for cpu := 0; cpu < runtime.NumCPU(); cpu++ {
+		go func(offset uint64) {
+			defer wg.Done()
+			for i := offset; i < n; i += uint64(runtime.NumCPU()) {
+				// NOTE: don't bother trying to optimize generateSpendableKey;
+				// profiling shows that ed25519 key generation consumes far
+				// more CPU time than encoding or hashing.
+				keys[i] = generateSpendableKey(seed, i)
```
DavidVorick: this is not surprising, but perhaps it's worth including the profiling results here, along with a date. It's possible that at some point we switch to a different ed25519 library, or that more crypto-based hardware accelleration becomes available.

DavidVorick: I know Intel has been adding more and more support for this each release cycle

modules/wallet/encrypt.go
lukechampine: >As new outputs come in, all of the seed addresses will be checked.

The problem is that you need to generate keys in order to identify outputs, but you don't know how many keys to generate until you've scanned the whole blockchain.
The alternative is to do something like `SweepSeed` and pregenerate a ton of keys, hoping that you generate enough to catch them all in one go, and generating more + rescanning if necessary. That's probably what we'll want to do in the long run, but the code is a lot trickier.

modules/wallet/persist_legacy.go
lukechampine: Noted, I'll reference it when I write the compatibility code. But there isn't much sense in keeping it around if it's all commented out, especially since it will be 100% obsolete in a few weeks.

Speaking of which, when I write the compat code, I should use the new approach of defining functions in the `compatibility` package, right?

modules/wallet/scan.go
lukechampine: you need the unlock hashes to compare against the addresses that appear in the blockchain. Perhaps you could optimize by truncating the hash and then doing a follow-up full comparison if you get a match. i.e.:
```go
type truncatedHash [8]byte

keys map[truncatedHash]uint64

// check if addr is a seed address
if index, ok := keys[addr]; ok {
    // partial match; check for full match
    fullHash := generateKey(seed, index).UnlockHash()
    if fullHash == addr {
        // full match
    }
}
```

modules/wallet/scan.go
lukechampine: ah, that's good to know.

modules/wallet/scan.go
lukechampine: possibly. Can you lay out the exact scenario? I'm pretty sure I understand, but I want to confirm.
In any case, it should be easy to delete outputs seen in reverted blocks.

modules/wallet/scan.go
lukechampine: An important detail it that the seedScanner doesn't "stick around" like the wallet does; it only exists inside the scope of `SweepSeed` (and `InitFromSeed`). I'm not sure what you mean by "miss money."

You are right to be wary, since we haven't used this pattern before. I like it because it cleanly separates the scanner state from the wallet state, but of course there are tradeoffs. I'll be more confident once `SweepSeed` and `InitFromSeed` have passed both automated testing and real-world usage.

modules/wallet/encrypt.go
DavidVorick: Is there any reason you couldn't reuse the `SweepSeed` code, except you just don't save the output data? Or even if you do, you just don't make the transaction?

modules/wallet/persist_legacy.go
DavidVorick: well, I'm actually referring to a different type of compatibility code.

The `compatibility` package is for upgrading persist files between `siad` versions, so everything in that package should approximately follow the method of `compatV1.0.2ToV1.0.3(persistFolder)`.

I don't think it makes as much sense to have the wallet loading code in the compatibility package, because the method of loading old wallets is going to change if the internals of the wallet change. It's not something you can easily chain because you aren't changing the 033x file itself, you're just using the contents to learn some unlock hashes which you are supposed to be tracking.

modules/wallet/scan.go
DavidVorick: let's punt it for the time being but definitely something we need to test + optimize before merging in to master.

modules/wallet/scan.go
DavidVorick: wallet height: 70,000
<scan starts>
<scan finishes up to 70,000>
<wallet gets a CC, jumps to height 70,001>
<scan finishes merging new outputs into wallet database>

in that code arrangement, the new outputs never got scanned for block 70,001, because the two subscriptions were running concurrently.

lukechampine: The assumption is that the seeds have not been used recently, so as long as you have finished IBD, you should find all of their outputs. The off-by-one scenario could happen, but is it any different from an output appearing 2 blocks after you do the scan, or 200, or a year later?

DavidVorick: > The assumption is that the seeds have not been used recently

I don't think that's a safe assumption, for example a lot of miners will have many seeds active at once, and they may be trying to merge them all into one wallet without shutting off any of their miners.

If you miss the output in block 70,001, you will never see it, which means it's the same as never receiving the money.

modules/wallet/scan.go
DavidVorick: user makes file contract
user restarts wallet
user makes file contract. Wallet is dumb so it uses the same outputs as the previous file contract

Two blocks get found an propagated at the same time. One has one set of FCs, the other has the other set of FCs. First one makes it into your consensus set, then the other wins the race so you reorg. Now you've got both in your history. Wallet during sees both and tries to spend both.

Something to that effect. The FC transaction create refunds, it's the refund that the wallet is trying to sweep.

modules/wallet/scan.go
lukechampine: So what's the solution? To continue tracking the seed after it's been swept? That doesn't match my understanding of what a "sweep" is; I've been treating it as a single operation. In that way, it's distinct from loading a seed.

DavidVorick: ah I guess you are right, it shouldn't be that important if you are doing a sweep instead of a load. But there needs to be some warning/info that the wallet will not be tracking the outputs found in the sweep.

modules/wallet/encrypt.go
lukechampine: that is precisely what I do, haha. `InitFromSeed` creates a `seedScanner`, scans the blockchain, and then uses `s.largestIndexSeen`. The `seedScanner` will still accumulate the outputs, but `InitFromSeed` ignores them.

modules/wallet/persist_legacy.go
lukechampine: ah, gotcha -- imports are not the same as upgrades.
It seems like the most important thing is to document how 033x wallets were saved. I will make sure that information is included when I restore 033x loading.

lukechampine:
What do you think about giving the user the option to manually specify the SeedProgress of an imported seed? If you know you've used less than 1 million keys, then its much faster to set that as your progress instead of scanning the whole blockchain. On the other hand, if you overestimate by a large factor, then you will have to wait longer for all the keys to be generated, and the wallet in general will run a bit slower. Still, I think it would make sense to at least make it an option for power users.

DavidVorick:
Maybe into the API, but if you present it to users it will confuse them. A rescan is going to take a long time either way.

It may make sense to have it be an input with a recommended value in the docstring, but I would probably not put it in the GUI. Maybe an optional command-specific flag for the subcommand in siac.

modules/wallet/scan.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/NebulousLabs/Sia/modules"
+	"github.com/NebulousLabs/Sia/types"
+)
+
+// numInitialKeys is the number of keys generated by the seedScanner before
+// scanning the blockchain for the first time.
+var numInitialKeys = func() uint64 {
+	switch build.Release {
+	case "dev":
+		return 1e4
```
DavidVorick: 10e3

PR #1435, opened 2016-09-12
Gateway fixes

DavidVorick:


modules/gateway/peers.go
```diff
@@ -514,7 +514,7 @@ func (g *Gateway) permanentPeerManager(closedChan chan struct{}) {
 		// Also wait a while and try again if there is at least one outbound
 		// peer and the selected peer is local. We do not want all of our
 		// outbound peers to be local peers.
-		if err != nil || (numOutboundPeers > 0 && addr.IsLocal() && build.Release != "testing") {
+		if err != nil || (numOutboundPeers >= maxLocalOutboundPeers && addr.IsLocal() && build.Release != "testing") {
```
VoidingWarranties: Can you explain this change? `numOutboundPeers` seems to be the total number of outbound peers, local or not. Why compare this to `maxLocalOutboundPeers`?

DavidVorick:
did a rebase and it wiped out all of my signatures. Is there a command that makes it easy to re-sign all the commits when you rebase?

johnathanhowell:
Not that I'm aware of, though on my system rebasing re-signs each commit

lukechampine:
Are you signing manually with `-S`, or did you configure git to sign by default? I did the latter and, like John, rebasing signs every commit.

modules/gateway/conn.go
```diff
@@ -24,13 +24,13 @@ func (pc peerConn) RPCAddr() modules.NetAddress {
...
-		Cancel:  g.threads.StopChan(),
+		Cancel:   g.threads.StopChan(),
+		Deadline: time.Now().Add(connStdDeadline),
+		Timeout:  dialTimeout,
 	}
 	conn, err := dialer.Dial("tcp", string(addr))
 	if err != nil {
 		return nil, err
 	}
-	conn.SetDeadline(time.Now().Add(connStdDeadline))
```
lukechampine: I don't think these are equivalent. The `Dial` docstring says:

>// Deadline is the absolute point in time after which dials
        // will fail. If Timeout is set, it may fail earlier.
        // Zero means no deadline, or dependent on the operating system
        // as with the Timeout option.
        Deadline time.Time

So this deadline only applies to the dial operation, not the connection. Weird since `Timeout` seems like the same thing, but that's what it says.

DavidVorick:
I've been manually sigining with -S, rebase doesn't have a -S option, which I guess explains everything.

modules/gateway/conn.go
DavidVorick: Making this changed fixed a bug where the gateway's were having a substantial amount of i/o timeouts. I'm not sure where I read that I should be doing that, but I think it was in google-gropus. I think it said that `Dial` we reuse conns, and so you want to set the timeout in the dialer instead of on the conn itself.

Now I'm second guessing myself. John can back me up that it solved a bunch of i/o timeout warnings in the gateway.

DavidVorick: oh, I was wrong. It's only when you are doing http transports that you shouldn't be setting a timeout on the conn itself.

I may have fixed the problem via some of the other changes, I added some bounds to the number of concurrent outbound connections the gateway will have, which I think made things a lot easier on the antfarm.

source of wrong information: gropus.google.com/forum/#!topic/golang-nuts/VncYuUsEMvY

PR #1437, opened 2016-09-14
Fix some small typo's in the documentation

robvanmieghem:


doc/Consensus.md
```diff
@@ -91,11 +91,11 @@ There are future plans to enable sidechain compatibility with Sia. This would
 allow other currencies such as Bitcoin to be spent in all the same places that
 the Siacoin can be spent.
 
-Marshalling
+Marshaling
 -----------
 
 Many of the Sia types need to be hashed at some point, which requires having a
-consistent algorithm for marshalling types into a set of bytes that can be
+consistent algorithm for marshaling types into a set of bytes that can be
```
DavidVorick: https://en.wikipedia.org/wiki/Marshalling_(computer_science)

I believe Marshalling is correct here

DavidVorick:
Looks good, except for the marshalling correction

robvanmieghem:
@DavidVorick You are right, seems like both versions are correct. I removed that change.

PR #1439, opened 2016-09-19
restore gateway default conn timeout behavior

DavidVorick:
Originally, the behavior was altered because it was assumed that the
gateway could be reusing conns during Dial. This was a misunderstanding,
some http frameworks will reuse conns but the standard dialer.Dial will
not (from my best understanding).

It turns out that other techniques were responsible for fixing the
gateway desync issues, not this particular change.

I have run some tests in the antfarm, and it seems like reverting this
change has not re-introduced the desynchronization that we were
struggling with earlier.


DavidVorick:
hmm. I may have made this PR too early. Some extra checks that I ran are now demonstrating desyncing again.

DavidVorick:
Further testing shows that desynchronization can still happen, though it seems to be related to faulty renter code. I was able to replicate the desynchronization on all branches.

PR #1440, opened 2016-09-21
Hostdb debugging

DavidVorick:
As this is a debugging PR, the commits are all over the place because I was poking around and tuning/tweaking.

generally speaking, I added more/better logging, including a String() method for types.SiaPublicKey. I changed the weird 10,000 gothread queue into a slice-based queue, which should make life easier for the scheduler. I swapped out some potentially buggy height update code for some known height update code. I added some shutdown statements so we can see what's going wrong when `siad` is taking forever to shutdown. Thus far it seems to be the host.

I don't think I solved any real problems here, other than adding more logging which allows us to better see what's going wrong if a host cannot be contacted.

### DIFF
--- a/modules/renter/hostdb/scan.go
+++ b/modules/renter/hostdb/scan.go
@@ -23,11 +23,11 @@ const (
 	minScanSleep     = 1 * time.Hour
 
 	maxActiveHosts              = 500
-	inactiveHostCheckupQuantity = 250
+	inactiveHostCheckupQuantity = 2000
 
 	maxSettingsLen = 2e3
 
-	hostRequestTimeout = 5 * time.Second
+	hostRequestTimeout = 60 * time.Second
 
 	// scanningThreads is the number of threads that will be probing hosts for
 	// their settings and checking for reliability.
@@ -110,6 +110,9 @@ func (hdb *HostDB) managedUpdateEntry(entry *hostEntry, newSettings modules.Host
 	if exists {
 		existingNode.removeNode()
 		delete(hdb.activeHosts, entry.NetAddress)
+	} else if len(hdb.activeHosts) > maxActiveHosts {
+		// We already have the maximum number of active hosts, do not add more.
+		return
 	}
 
 	// Update the host settings, reliability, and weight. The old NetAddress
@@ -117,13 +120,14 @@ func (hdb *HostDB) managedUpdateEntry(entry *hostEntry, newSettings modules.Host
 	newSettings.NetAddress = entry.HostExternalSettings.NetAddress
 	entry.HostExternalSettings = newSettings
 	entry.Reliability = MaxReliability
-	entry.Weight = calculateHostWeight(*entry)
 	entry.Online = true
+	entry.Weight = calculateHostWeight(*entry)
+	hdb.insertNode(entry)
 
-	// If 'maxActiveHosts' has not been reached, add the host to the
-	// activeHosts tree.
-	if _, exists := hdb.activeHosts[entry.NetAddress]; exists || len(hdb.activeHosts) < maxActiveHosts {
-		hdb.insertNode(entry)
+	// Sanity check - the node should be in the hostdb now.
+	_, exists = hdb.activeHosts[entry.NetAddress]
+	if !exists {
+		hdb.log.Critical("Host was not added to the list of active hosts after the entry was updated.")
 	}
 	hdb.save()
 }
--- a/modules/renter/hostdb/scan.go
+++ b/modules/renter/hostdb/scan.go
@@ -28,11 +28,11 @@ const (
 
 	maxSettingsLen = 2e3
 
-	hostRequestTimeout = 60 * time.Second
+	hostRequestTimeout = 30 * time.Second
 
 	// scanningThreads is the number of threads that will be probing hosts for
 	// their settings and checking for reliability.
-	scanningThreads = 5
+	scanningThreads = 12
 )
 
 // Reliability is a measure of a host's uptime.
@@ -185,13 +185,11 @@ func (hdb *HostDB) threadedProbeHosts() {
 		netAddr := hostEntry.NetAddress
 		pubKey := hostEntry.PublicKey
 		hdb.mu.RUnlock()
-		hdb.log.Debugln("Scanning", netAddr, fmt.Sprintf("%x", pubKey.Key))
+		hdb.log.Debugln("Scanning", netAddr, pubKey)
 		var settings modules.HostExternalSettings
-		connFail := false
 		err := func() error {
 			conn, err := hdb.dialer.DialTimeout(netAddr, hostRequestTimeout)
 			if err != nil {
-				connFail = true
 				return err
 			}
 			defer conn.Close()
@@ -204,7 +202,7 @@ func (hdb *HostDB) threadedProbeHosts() {
 			return crypto.ReadSignedObject(conn, &settings, maxSettingsLen, pubkey)
 		}()
 		if err != nil {
-			hdb.log.Debugln("Scanning", netAddr, pubKey, "failed, status:", connFail, "error:", err)
+			hdb.log.Debugln("Scanning", netAddr, pubKey, "failed:", err)
 		} else {
 			hdb.log.Debugln("Scanning", netAddr, pubKey, "succeeded")
 		}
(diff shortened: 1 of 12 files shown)


modules/renter/hostdb/update.go
```diff
@@ -34,22 +34,43 @@ func (hdb *HostDB) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+	// Update the hostdb's understanding of the block height.
+	for _, block := range cc.RevertedBlocks {
+		// Only doing the block check if the height is above zero saves hashing
+		// and saves a nontrivial amount of time during IBD.
+		if hdb.blockHeight > 0 || block.ID() != types.GenesisID {
+			hdb.blockHeight--
+		} else if hdb.blockHeight != 0 {
+			// Sanity check - if the current block is the genesis block, the
+			// miner height should be set to zero.
+			hdb.log.Critical("Miner has detected a genesis block, but the height of the miner is set to ", hdb.blockHeight)
```
lukechampine: miner -> hostdb

modules/renter/hostdb/update.go
```diff
@@ -34,22 +34,43 @@ func (hdb *HostDB) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+	}
+	for _, block := range cc.AppliedBlocks {
+		// Only doing the block check if the height is above zero saves hashing
+		// and saves a nontrivial amount of time during IBD.
+		if hdb.blockHeight > 0 || block.ID() != types.GenesisID {
+			hdb.blockHeight++
+		} else if hdb.blockHeight != 0 {
+			// Sanity check - if the current block is the genesis block, the
+			// miner height should be set to zero.
+			hdb.log.Critical("Miner has detected a genesis block, but the height of the miner is set to ", hdb.blockHeight)
```
lukechampine: ditto

modules/transactionpool/persist.go
```diff
@@ -53,6 +55,9 @@ func (tp *TransactionPool) initPersist() error {
 	}
 
 	// Open the database file.
+	if build.DEBUG && build.Release != "testing" {
+		fmt.Println("\tTransaction Pool: opening database")
+	}
```
lukechampine: is this leftover code from a debugging session?

modules/transactionpool/transactionpool.go
```diff
@@ -102,9 +104,12 @@ func New(cs modules.ConsensusSet, g modules.Gateway, persistDir string) (*Transa
 	if err != nil {
 		return nil, err
 	}
+	if build.DEBUG && build.Release != "testing" {
+		fmt.Println("\tTransaction Pool: persist initialization complete")
+	}
 
 	// Register RPCs
-	g.RegisterRPC("RelayTransactionSet", tp.relayTransactionSet)
+	go g.RegisterRPC("RelayTransactionSet", tp.relayTransactionSet)
```
lukechampine: why a goroutine? doesn't seem worth it

types/signatures.go
```diff
@@ -389,3 +390,10 @@ func (t *Transaction) validSignatures(currentHeight BlockHeight) error {
 
 	return nil
 }
+
+// String defines how to print a SiaPublicKey - hex is used to keep things
+// compact during logging. The key type prefix and lack of a checksum help to
+// separate it from a sia address.
+func (spk *SiaPublicKey) String() string {
+	return spk.Algorithm.String() + ":" + fmt.Sprintf("%x", spk.Key)
```
lukechampine: :+1:

modules/renter/hostdb/scan.go
```diff
@@ -46,14 +46,54 @@ var (
 // The gofunc is created inside of this function to eliminate the burden of
 // needing to remember to call 'go addHostToScanPool'.
 func (hdb *HostDB) scanHostEntry(entry *hostEntry) {
+	// Add the entry to a waitlist, then check if any thread is currently
+	// emptying the waitlist. If not, spawn a thread to empty the waitlist.
```
lukechampine: instead of spawning a new thread as needed, why not just spawn a single thread at startup?

modules/renter/hostdb/scan.go
```diff
@@ -46,14 +46,54 @@ var (
...
+			if len(hdb.scanList) == 0 {
+				// Scan list is empty, can exit. Let the world know that nobody
+				// is emptying the scan list anymore.
+				hdb.scanWait = false
+				hdb.mu.Unlock()
+				return
+			}
+			// Get the next host, shrink the scan list.
+			entry := hdb.scanList[0]
+			hdb.scanList = hdb.scanList[1:]
```
lukechampine: this prevents memory reuse; taking from the end of the slice would be more efficient. However, this would reverse the iteration order (the most recently inserted entry would be scanned first). That might be acceptable, not sure.

modules/renter/hostdb/scan.go
lukechampine: also, the function's docstring needs to be updated

modules/transactionpool/persist.go
DavidVorick: No it's meant to stay there for the time being. The transcation pool occasionally hangs at startup (indefinitely) and I'm trying to figure out where in the code it gets stuck.

modules/transactionpool/transactionpool.go
DavidVorick: transaction pool sometimes hangs at startup, felt like it might make a difference.

modules/renter/hostdb/scan.go
DavidVorick: This was (I felt) the quickest way to handle this. The hostdb is one of the earlier modules, so it doesn't have much of the safety and it really doesn't have any of the clean shutdown.

Doing this properly will take a few days at least, and I want to be focusing my energy on the renter at this point, because for the most part the hostdb is performing just fine.

modules/renter/hostdb/scan.go
DavidVorick: I was considering it. I'm not sure how the slice-append algo works, but if it only doubles based on the current size of the slice (instead of total historic usage) it'll basically be allocating memory every time there's an append, getting no advantage from amortization.

DavidVorick:
Yeah I was thinking the same thing, hostdb scan pool hosts can be dropped if they never make it from the scan pool into the list of all hosts.

Fixing that is going to require more major changes. The hostdb mostly works well enough right now, I don't think it's the best use of time to be fixing all this stuff. The host + renter are definitely much worse in terms of practical usability.

modules/renter/hostdb/scan.go
lukechampine: I think that is correct, yeah, unless there are multiple appends before anything is consumed.

modules/transactionpool/transactionpool.go
lukechampine: did it help?
my concern is that you have a race condition if you try to call the `RelayTransactionSet` RPC immediately after creating the transaction pool, though I'm sure this would only occur during testing and in race circumstances.

modules/transactionpool/persist.go
DavidVorick: It only prints if you are running the debugging build, and I think it's useful to have this information coming from anyone running a `siad` instance - if it's having trouble starting up you want to know more about why.

It's not an overwhelming amount of information, and the bug is rare so we have a better chance of catching it if the code is in more places.

modules/transactionpool/transactionpool.go
DavidVorick: Oh, that would be an issue. I'll switch it back.

modules/transactionpool/persist.go
lukechampine: I don't think this is the best way to debug the problem. Println debugging is not effective against hard-to-reproduce bugs. I suggest running with profiling enabled so that you can inspect the blocking profile live when it hangs. That should tell you exactly where it got stuck.

DavidVorick: we could set it up so that the profiler is running in DEBUG mode, I'm not sure how expensive that would be, but should help.

DavidVorick: The transaction pool doesn't have a logger, heh.

lukechampine:
hmm. GitHub used to collapse PR comments (marking them as outdated) after you changed the code. Looks like the new review system doesn't, which is a bummer. Makes it hard to tell what's left to do.

DavidVorick:
otoh it would occasionally collapse things that hadn't been addressed yet due to weird diff parsing.

I'm pretty sure I've addressed everything, including punting the tpool debugging to another PR.

PR #1442, opened 2016-09-21
Improve siac's `wallet load` ux

johnathanhowell:
Numerous users have reported being confused by the `wallet load` ux, expecting the `Password: ` input to be the wallet password of the seed, 033x, or siag that they are loading.  This PR adds some interactivity that should hopefully make it more clear that the user should provide their _current_ wallet password.

### DIFF
--- a/siac/walletcmd.go
+++ b/siac/walletcmd.go
@@ -61,7 +61,7 @@ By default the wallet encryption / unlock password is the same as the generated 
 	}
 
 	walletLoad033xCmd = &cobra.Command{
-		Use:   "033x",
+		Use:   "033x [filepath]",
 		Short: "Load a v0.3.3.x wallet",
 		Long:  "Load a v0.3.3.x wallet into the current wallet",
 		Run:   wrap(walletload033xcmd),
@@ -75,11 +75,11 @@ By default the wallet encryption / unlock password is the same as the generated 
 	}
 
 	walletLoadSiagCmd = &cobra.Command{
-		Use:   `siag [filepaths]`,
-		Short: "Load a siag keyset into the wallet",
-		Long: `Load a set of siag keys into the wallet - typically used for siafunds.
-Example: 'siac wallet load siag key1.siakey,key2.siakey'`,
-		Run: wrap(walletloadsiagcmd),
+		Use:     `siag [filepath,...]`,
+		Short:   "Load siag key(s) into the wallet",
+		Long:    "Load siag key(s) into the wallet - typically used for siafunds.",
+		Example: "siac wallet load siag key1.siakey,key2.siakey",
+		Run:     wrap(walletloadsiagcmd),
 	}
 
 	walletLockCmd = &cobra.Command{
--- a/siac/walletcmd.go
+++ b/siac/walletcmd.go
@@ -3,7 +3,6 @@ package main
 import (
 	"fmt"
 	"math/big"
-	"time"
 
 	"github.com/bgentry/speakeasy"
 	"github.com/spf13/cobra"
@@ -62,7 +61,7 @@ By default the wallet encryption / unlock password is the same as the generated 
 	}
 
 	walletLoad033xCmd = &cobra.Command{
-		Use:   "033x [filepath]",
+		Use:   "033x",
 		Short: "Load a v0.3.3.x wallet",
 		Long:  "Load a v0.3.3.x wallet into the current wallet",
 		Run:   wrap(walletload033xcmd),
@@ -193,11 +192,11 @@ func walletinitcmd() {
 }
 
 // walletload033xcmd loads a v0.3.3.x wallet into the current wallet.
-func walletload033xcmd(source string) {
-	fmt.Println("Loading 033x...")
-	time.Sleep(time.Second * 3)
-	fmt.Println("Load successful.  Encrypting with current encryption password")
-	password, err := speakeasy.Ask("Need Password: ")
+func walletload033xcmd() {
+	fmt.Print("Enter 033x Location: ")
+	var source string
+	fmt.Scanln(&source)
+	password, err := speakeasy.Ask("Current Password: ")
 	if err != nil {
 		die("Reading password failed:", err)
 	}
@@ -211,16 +210,13 @@ func walletload033xcmd(source string) {
 
 // walletloadseedcmd adds a seed to the wallet's list of seeds
 func walletloadseedcmd() {
-	fmt.Println("Loading seed...")
-	time.Sleep(time.Second * 3)
-	fmt.Println("Load successful.  Encrypting with current encryption password")
-	password, err := speakeasy.Ask("Need Password: ")
-	if err != nil {
-		die("Reading password failed:", err)
-	}
 	seed, err := speakeasy.Ask("New Seed: ")
 	if err != nil {
 		die("Reading seed failed:", err)
+	}
+	password, err := speakeasy.Ask("Current Password: ")
+	if err != nil {
+		die("Reading password failed:", err)
 	}
 	qs := fmt.Sprintf("encryptionpassword=%s&seed=%s&dictionary=%s", password, seed, "english")
 	err = post("/wallet/seed", qs)
@@ -231,11 +227,11 @@ func walletloadseedcmd() {
 }
 
 // walletloadsiagcmd loads a siag key set into the wallet.
-func walletloadsiagcmd(keyfiles string) {
-	fmt.Println("Loading siag...")
-	time.Sleep(time.Second * 3)
-	fmt.Println("Load successful.  Encrypting with current encryption password")
-	password, err := speakeasy.Ask("Need Password: ")
+func walletloadsiagcmd() {
+	fmt.Print("Enter Siag Location: ")
+	var keyfiles string
+	fmt.Scanln(&keyfiles)
+	password, err := speakeasy.Ask("Current Password: ")
 	if err != nil {
 		die("Reading password failed:", err)
 	}
--- a/siac/walletcmd.go
+++ b/siac/walletcmd.go
@@ -3,6 +3,7 @@ package main
 import (
 	"fmt"
 	"math/big"
+	"time"
 
 	"github.com/bgentry/speakeasy"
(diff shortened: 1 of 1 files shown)


DavidVorick:
LGTM

siac/walletcmd.go
```diff
@@ -193,7 +194,10 @@ func walletinitcmd() {
 
 // walletload033xcmd loads a v0.3.3.x wallet into the current wallet.
 func walletload033xcmd(source string) {
-	password, err := speakeasy.Ask("Wallet password: ")
+	fmt.Println("Loading 033x...")
+	time.Sleep(time.Second * 3)
+	fmt.Println("Load successful.  Encrypting with current encryption password")
+	password, err := speakeasy.Ask("Need Password: ")
```
lukechampine: I don't understand this behavior. Why is there a 3-second delay? Why lie to the user?

I'd prefer something like:
```
$ siac wallet load 033x wallet.dat
Your current wallet password is needed to encrypt the loaded data.
Password:
Loaded wallet.dat
```

siac/walletcmd.go
```diff
@@ -225,7 +232,10 @@ func walletloadseedcmd() {
 
 // walletloadsiagcmd loads a siag key set into the wallet.
 func walletloadsiagcmd(keyfiles string) {
-	password, err := speakeasy.Ask("Wallet password: ")
+	fmt.Println("Loading siag...")
+	time.Sleep(time.Second * 3)
+	fmt.Println("Load successful.  Encrypting with current encryption password")
```
lukechampine: We use one space after periods. That goes for comments as well.

siac/walletcmd.go
```diff
@@ -207,7 +211,10 @@ func walletload033xcmd(source string) {
 
 // walletloadseedcmd adds a seed to the wallet's list of seeds
 func walletloadseedcmd() {
-	password, err := speakeasy.Ask("Wallet password: ")
+	fmt.Println("Loading seed...")
+	time.Sleep(time.Second * 3)
+	fmt.Println("Load successful.  Encrypting with current encryption password")
+	password, err := speakeasy.Ask("Need Password: ")
```
lukechampine: this prints "load successful" before it even asks you for the thing it's loading!

crazy pills

siac/walletcmd.go
DavidVorick: Idea was to make it more clear that the password needed was the encryption password to the current wallet.

johnathanhowell:
This branch has been updated to prompt for the new seed, 033x, or siag key first, then prompt for the current wallet password:

```
`New Seed: `
`Current Password: `

`Enter Siag Location: `
`Current Password: `

`Enter 033x Location: `
`Current Password: `
```

this should help the confusion around this part of the ux

siac/walletcmd.go
```diff
@@ -224,8 +227,11 @@ func walletloadseedcmd() {
 }
 
 // walletloadsiagcmd loads a siag key set into the wallet.
-func walletloadsiagcmd(keyfiles string) {
-	password, err := speakeasy.Ask("Wallet password: ")
+func walletloadsiagcmd() {
+	fmt.Print("Enter Siag Location: ")
```
DavidVorick: This prompt doesn't make it clear that the user is supposed to specify multiple files.

DavidVorick:
I'm still not sure that 'Current Password' is clear enough. Would be nice to dry-run this on a few new users and see if they can figure out what password they are supposed to provide. Maybe 'Current Wallet Password:', or maybe just explain in full:

'We need to encrypt the new data using the current wallet password, please provide:'

DavidVorick:
looking for thoughts from @lukechampine and @VoidingWarranties

siac/walletcmd.go
```diff
@@ -61,7 +61,7 @@ By default the wallet encryption / unlock password is the same as the generated
 	}
 
 	walletLoad033xCmd = &cobra.Command{
-		Use:   "033x [filepath]",
+		Use:   "033x",
```
VoidingWarranties: If you remove `[filepath]` here, consider also removing `[filepaths]` from `load siag`. Or inversely, consider keeping `[filepath]` here.

siac/walletcmd.go
```diff
@@ -192,8 +192,11 @@ func walletinitcmd() {
 }
 
 // walletload033xcmd loads a v0.3.3.x wallet into the current wallet.
-func walletload033xcmd(source string) {
-	password, err := speakeasy.Ask("Wallet password: ")
+func walletload033xcmd() {
+	fmt.Print("Enter 033x Location: ")
+	var source string
+	fmt.Scanln(&source)
```
VoidingWarranties: Can you explain the rationale behind this change a bit more? It doesn't seem like more interactivity necessarily improves clarity here, especially since the confusion came from wallet password vs. seed, not the filepath of the wallet being loaded.

Then again this pattern more closely matches the pattern of `siac wallet load seed`. However doing so gets rid of bash completion which to me is a deal breaker.

siac/walletcmd.go
```diff
@@ -192,8 +192,11 @@ func walletinitcmd() {
 }
 
 // walletload033xcmd loads a v0.3.3.x wallet into the current wallet.
-func walletload033xcmd(source string) {
-	password, err := speakeasy.Ask("Wallet password: ")
+func walletload033xcmd() {
+	fmt.Print("Enter 033x Location: ")
```
VoidingWarranties: In the rest of siac we only capitalize the first word of a prompt.

siac/walletcmd.go
VoidingWarranties: Same comment as before (capitalizing first word only)

VoidingWarranties: More than just making it plural, perhaps also provide a short example `file1, file2, ...`

siac/walletcmd.go
```diff
@@ -224,8 +227,11 @@ func walletloadseedcmd() {
 }
 
 // walletloadsiagcmd loads a siag key set into the wallet.
-func walletloadsiagcmd(keyfiles string) {
-	password, err := speakeasy.Ask("Wallet password: ")
+func walletloadsiagcmd() {
+	fmt.Print("Enter Siag Location: ")
+	var keyfiles string
+	fmt.Scanln(&keyfiles)
+	password, err := speakeasy.Ask("Current Password: ")
```
VoidingWarranties: Same comment as before (capitalizing first word only)

siac/walletcmd.go
```diff
@@ -224,8 +227,11 @@ func walletloadseedcmd() {
 }
```
VoidingWarranties: Not touched by this PR, but `fmt.Println("Added Key")` could me reworded to be more consistent with the other success indications. It's especially confusing because the word key is used instead of seed.

Even the existing success messages could be improved. Maybe our resident grammar master could comment, but @lukechampine "Wallet loading successful." doesn't seem like a proper sentence.

siac/walletcmd.go
lukechampine: I agree. More interactivity is generally a bad thing because it makes siac harder to script.

siac/walletcmd.go
lukechampine: I think it's implying "[The] Wallet loading [was] successful," but "Loaded 0.3.3.x wallet [filename]" would be more clear.

siac/walletcmd.go
johnathanhowell: The rationale for doing it this way was to match the way we load seeds, ie interactively.  I guess it doesn't really make sense since the 033x and siag loaders take a file location, not the actual data.

johnathanhowell:
Updated with the changes requested, removing the interactivity in siag/033x load functions and adding david's suggested `Enter password...` text.

siac/walletcmd.go
DavidVorick: the usage was chaged to exclude the filepath, definitely the correct option is to leave it off, and remove it from `siag` as well.

VoidingWarranties: I agree it makes sense to leave it off the the filepath is specified interactively. But if the filepath is not specified interactively, shouldn't it be kept?

VoidingWarranties:
Do we want to get this into the upcoming RC? If so there's a few minor comments that still haven't been addressed. Otherwise LGTM.

VoidingWarranties:
Cleaned up the usage. @DavidVorick can you check my changes to the `load siag` usage?

siac/walletcmd.go
```diff
@@ -75,11 +75,11 @@ By default the wallet encryption / unlock password is the same as the generated
 	}
 
 	walletLoadSiagCmd = &cobra.Command{
-		Use:   `siag [filepaths]`,
-		Short: "Load a siag keyset into the wallet",
-		Long: `Load a set of siag keys into the wallet - typically used for siafunds.
-Example: 'siac wallet load siag key1.siakey,key2.siakey'`,
-		Run: wrap(walletloadsiagcmd),
+		Use:     `siag [filepath,...]`,
```
DavidVorick: I don't think this is the right way to specify that multiple filepaths can be used. Qubes does it this way: https://www.qubes-os.org/doc/vm-tools/qvm-copy-to-vm/ (note that their whole style is different from ours, therefore not the best example)

Another alternative seems to be `siag [filepath]...`

DavidVorick:
Looks good except for the `siag [filepath,...]`, which seems a bit awkward to me. I couldn't find any example usage strings that I really liked, so I'm okay merging it as-is at the moment, though I feel like we can do better here.

siac/walletcmd.go
DavidVorick: docopt.org

VoidingWarranties:
I agree it looks clunky. I was copying `cat` and `ls`'s styles.
`usage: cat [-benstuv] [file ...]`
`usage: ls [-ABCFGHLOPRSTUWabcdefghiklmnopqrstuwx1] [file ...]`
but used a comma instead of a space as files must be comma separated without a space.

DavidVorick:
they do? That's clunky as well. Not sure what I was thinking when I wrote that.

Should we leave it then, or do you think it's worth parsing the input files in a more sane way?

PR #1444, opened 2016-09-27
replace goxc with release.sh

lukechampine:
goxc is overkill for our purposes.

Note that you will need the developer private key in order to sign binaries. (Well, technically you can sign them with any key.)

Maybe `siac` should include a command for verifying signed release binaries.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+set -e
+
+# version and private key are supplied as arguments
+version="$1"
+keyfile="$2"
+if [[ -z $version || -z $keyfile ]]; then
+	echo "Usage: $0 VERSION KEYFILE"
+	exit 1
+fi
+
+# check for keyfile before proceeding
+if [ ! -f $keyfile ]; then
+    echo "Key file not found: $keyfile"
+    exit 1
+fi
+
+for os in darwin linux windows; do
+	echo Packaging ${os}...
+	# create workspace
+	root=$(pwd)
+	folder=$root/release/Sia-$version-$os-amd64
+	rm -rf $folder
+	mkdir -p $folder
+	# compile and sign binaries
+	for pkg in siac siad; do
+		bin=$pkg
+		if [ "$os" == "windows" ]; then
+			bin=${pkg}.exe
+		fi
+		GOOS=${os} go build -o $folder/$bin ./$pkg
+		openssl dgst -sha256 -sign $keyfile -out $folder/${bin}.sig $folder/$bin
+	done
+	# add other artifacts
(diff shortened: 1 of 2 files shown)


DavidVorick:
can you add a `make` command that runs the shell script? Might seem silly but I find it easier to navigate a project when everything is in one place. The Makefile seems to be that place for Sia.

DavidVorick:
do you think it would be easy enough to add something that would verify the dev key from the build script? (for the developer to know they used the right key, not for the user to know that the binary is legit)

also, would be nice if you could pull the version out of build/version.go. There's probably some sort of sed command you can use. I guess you still need an arg to pass 'RC1', etc.

DavidVorick:
(this comment was supposed to be first, not sure what happened)

putting the verification tool in the binary that you are trying to protect is probably more harmful, as someone distributing modified binaries only needs to change out the dev key to get that command to succeed. Then the user might run 'verify' thinking that they are protected when they actually have a malicious binary.

lukechampine:
>do you think it would be easy enough to add something that would verify the dev key from the build script?

You could sign a hash or something and compare to a hard-coded result. I'm not sure what that code would look like (not familiar with `openssl`). Personally I don't think it's worth the trouble; how many keys do you have lying around in PEM format? I'll consider it if it's trivial to do.

>would be nice if you could pull the version out of build/version.go. There's probably some sort of sed command you can use. I guess you still need an arg to pass 'RC1', etc.

That would be kinda brittle, and as you point out, you still need to specify the full semver. Also, passing the version as an argument lets you do things like:
```bash
$ ./release.sh fornax-debug key.pem
Packaging darwin...
Packaging linux...
Packaging windows...
$ ls release
Sia-fornax-debug-darwin-amd64.zip  Sia-fornax-debug-linux-amd64.zip  Sia-fornax-debug-windows-amd64.zip
```

release.sh
```diff
@@ -14,6 +14,11 @@ if [ ! -f $keyfile ]; then
     echo "Key file not found: $keyfile"
     exit 1
 fi
+keysum=$(sha256sum $keyfile | cut -c -64)
```
VoidingWarranties: OS X does not have `sha256sum`. Could we use `shasum -a 256` instead?

johnathanhowell:
should I build a similar tool for `Sia-UI`?

DavidVorick:
> Personally I don't think it's worth the trouble

It's worth the trouble b/c it's embarassing for the project if you get the signature wrong. Lots of users would get concerned and then we'd have to clarify that everything is okay, not canary, etc. etc. Better to have good hygiene.

I guess it doesn't need to go in the makefile if the only thing it does is make a cross-platform, distirbutable release.

PR #1445, opened 2016-09-30
Account for transaction fees when allocating sectors

DavidVorick:
The contractor will now account for transaction fees and for contract fees when it is allocating sectors, which means it should be much less likely to allocate too much storage and much less likely to over-charge substantially.

I also cut the fee by a factor of 5, because we don't need fees as high as 25 SC / kb, especially not after the price spike.

### DIFF
--- a/modules/renter/contractor/formcontract.go
+++ b/modules/renter/contractor/formcontract.go
@@ -60,7 +60,7 @@ func maxSectors(a modules.Allowance, hdb hostDB, tp transactionPool) (uint64, er
 	_, feeEstimation := tp.FeeEstimation()
 	costForTxnFees := types.NewCurrency64(estimatedFileContractTransactionSize).Mul(feeEstimation).Mul64(a.Hosts)
 	// Check for potential divide by zero
-	if a.Funds.Cmp(costForTxnFees.Add(costForContracts)) < 0 {
+	if a.Funds.Cmp(costForTxnFees.Add(costForContracts)) <= 0 {
 		return 0, errors.New("allowance is not large enough to cover the fees associated with creating file contracts")
 	}
 	sectorFunds := a.Funds.Sub(costForTxnFees).Sub(costForContracts)
@@ -68,7 +68,7 @@ func maxSectors(a modules.Allowance, hdb hostDB, tp transactionPool) (uint64, er
 	// Divide total funds by cost per sector.
 	numSectors, err := sectorFunds.Div(costPerSector).Uint64()
 	if err != nil {
-		return 0, errors.New("error when totaling number of sectors that can be bought with an allowance: "+err.Error())
+		return 0, errors.New("error when totaling number of sectors that can be bought with an allowance: " + err.Error())
 	}
 	return numSectors, nil
 }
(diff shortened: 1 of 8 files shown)


modules/renter/contractor/formcontract.go
```diff
@@ -28,34 +28,47 @@ var (
 
 // maxSectors is the estimated maximum number of sectors that the allowance
 // can support.
-func maxSectors(a modules.Allowance, hdb hostDB) (uint64, error) {
-	if a.Hosts == 0 || a.Period == 0 {
+func maxSectors(a modules.Allowance, hdb hostDB, tp transactionPool) (uint64, error) {
+	if a.Hosts <= 0 || a.Period <= 0 {
```
lukechampine: these are unsigned integers

modules/renter/contractor/formcontract.go
```diff
@@ -28,34 +28,47 @@ var (
...
+	averageSectorPrice := sectorSum.Div64(uint64(len(hosts)))
+	averageContractPrice := contractCostSum.Div64(uint64(len(hosts)))
+	costPerSector := averageSectorPrice.Mul64(a.Hosts).Mul64(modules.SectorSize).Mul64(uint64(a.Period))
+	costForContracts := averageContractPrice.Mul64(a.Hosts)
+
+	// Subtract fees for creating the file contracts from the allowance.
+	_, feeEstimation := tp.FeeEstimation()
+	costForTxnFees := types.NewCurrency64(estimatedFileContractTransactionSize).Mul(feeEstimation).Mul64(a.Hosts)
+	// Check for potential divide by zero
+	if a.Funds.Cmp(costForTxnFees.Add(costForContracts)) < 0 {
```
lukechampine: should be `<= 0`

modules/renter/contractor/formcontract.go
DavidVorick: it's equivalent either way. But when it's written like this you can be certain it's correct even if you don't know whether the integers are signed or not.

modules/renter/contractor/formcontract.go
DavidVorick: I had only intended to return an error if there was a risk of triggering a panic, but it's worth asking - if you can't even afford 1 sector (which could mean that you have >0 money, but just barely), should that also return an error?

DavidVorick: And, what if you have enough to buy 1 sector, but only on 1/3 of the requested hosts?

DavidVorick: Hmm. I'm guessing a user would be pretty upset if they set an allowance, payed 3 KS in fees, then got just 2-3 sectors - 50MiB of data. Perhaps we should return an error for anything less than... like 500 MiB of data or something.

lukechampine: If a.funds == cost+fee, you'll get a divide by zero later because you subtract them.

I think the average user would rather not spend more on fees than on actual storage costs, but it's hard to apply that sort of rule to everyone. In my ideal client, there would be a popup slider thing where you could adjust either the desired cost or the desired storage, and it would also highlight how much was being spent on fees.

lukechampine: wrt errors, I'm inclined to say that an error should only be returned if you are unable to perform the calculation. If you can't continue because you don't have enough money, that means you can afford zero sectors. So I think I would just return zero instead of an error.

DavidVorick: ok. The errFeesMoreThanSectors should be returned at another layer of the code. Missed the divide by zero.

PR #1449, opened 2016-10-04
imporoved error detection and handling in host

DavidVorick:
The host is now better at detecting errors - detects some errors it was
previously missing, and has eliminated some false positives. Also adds more
context to some of the errors that it does detect, which will make debugging
easier.

### DIFF
--- a/modules/host/negotiatedownload.go
+++ b/modules/host/negotiatedownload.go
@@ -1,6 +1,7 @@
 package host
 
 import (
+	"fmt"
 	"net"
 	"time"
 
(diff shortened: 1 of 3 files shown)


modules/host/negotiatedownload.go
```diff
@@ -157,22 +158,42 @@ func verifyPaymentRevision(existingRevision, paymentRevision types.FileContractR
...
+	// Determine the amount that was transfered from the renter.
+	if paymentRevision.NewValidProofOutputs[0].Value.Cmp(existingRevision.NewValidProofOutputs[0].Value) > 0 {
+		return extendErr("renter increased its valid proof output: ", errHighRenterValidOutput)
 	}
-	// The new revenue goes into the host's valid outputs.
-	if existingRevision.NewValidProofOutputs[1].Value.Add(expectedTransfer).Cmp(paymentRevision.NewValidProofOutputs[1].Value) < 0 {
-		return errLowHostValidOutput
+	fromRenter := existingRevision.NewValidProofOutputs[0].Value.Sub(paymentRevision.NewValidProofOutputs[0].Value)
+	// Verify that enough money was transfered.
+	if fromRenter.Cmp(expectedTransfer) < 0 {
```
lukechampine: splitting this into two separate `if`s is good

lukechampine:
looks good, I'm going to run it on the antfarm and see if it causes any new errors

modules/host/negotiatedownload.go
DavidVorick: I don't know why I was so stubborn about it before. Clearly the code was hard to think about, as we both reviewed it and okay'd it but it was wrong. Should have made the code more intuitive from the start.

lukechampine:
getting a few of these:
```
host expected to post at most 6978370370325708800 collateral, but contract has host posting 7017206518473608269:
```
this is with `collateral *= 0.98` btw

PR #1451, opened 2016-10-06
Prevent concurrent contract revision

lukechampine:
Also, downloaders are now thread-safe and cached by the contractor. They are referenced by their ID rather than the full contract object. This means less contract objects floating around and therefore less chance of saving/revising the wrong contract. It also opens the door for a future change: the contractor should be able to invalidate any active downloaders/editors as soon as it starts renewing their underlying contract. Generally speaking, the contractor should be the only package that deals with contracts directly.

One downside of caching, however, is that it complicates the question of when the connection should be closed. In this PR, I added a counter to the downloader object. Each time another goroutine starts using the downloader, the counter is incremented, and each time `Close` is called, the counter is decremented. `Close` only closes the underlying connection when the counter reaches zero.

Another complication is timeouts. The download RPC has a hard deadline, after which it will close. The existing download code essentially ignores this possibility, but the likelihood of hitting the deadline increases if connections are being cached and reused. To address this, the download code should attempt to reconnect when the deadline is reached. Ideally, it would detect timeout errors and only retry in that case, but it may work fine to simply retry after any failure.

My next step is to give downloads a way to interrupt uploads in the renter. I will use a simplistic approach because it appears the uploading process may shortly be subject to more dramatic changes.

### DIFF
--- a/modules/renter/contractor/contractor.go
+++ b/modules/renter/contractor/contractor.go
@@ -43,6 +43,7 @@ type Contractor struct {
 	blockHeight     types.BlockHeight
 	cachedRevisions map[types.FileContractID]cachedRevision
 	contracts       map[types.FileContractID]modules.RenterContract
+	downloaders     map[types.FileContractID]*hostDownloader
 	lastChange      modules.ConsensusChangeID
 	renewHeight     types.BlockHeight             // height at which to renew contracts
 	revising        map[types.FileContractID]bool // prevent overlapping revisions
@@ -133,6 +134,7 @@ func newContractor(cs consensusSet, w wallet, tp transactionPool, hdb hostDB, p 
 
 		cachedRevisions: make(map[types.FileContractID]cachedRevision),
 		contracts:       make(map[types.FileContractID]modules.RenterContract),
+		downloaders:     make(map[types.FileContractID]*hostDownloader),
 		revising:        make(map[types.FileContractID]bool),
 	}
 
--- a/modules/renter/contractor/contractor.go
+++ b/modules/renter/contractor/contractor.go
@@ -44,7 +44,8 @@ type Contractor struct {
 	cachedRevisions map[types.FileContractID]cachedRevision
 	contracts       map[types.FileContractID]modules.RenterContract
 	lastChange      modules.ConsensusChangeID
-	renewHeight     types.BlockHeight // height at which to renew contracts
+	renewHeight     types.BlockHeight             // height at which to renew contracts
+	revising        map[types.FileContractID]bool // prevent overlapping revisions
 
 	financialMetrics modules.RenterFinancialMetrics
 
@@ -132,6 +133,7 @@ func newContractor(cs consensusSet, w wallet, tp transactionPool, hdb hostDB, p 
 
 		cachedRevisions: make(map[types.FileContractID]cachedRevision),
 		contracts:       make(map[types.FileContractID]modules.RenterContract),
+		revising:        make(map[types.FileContractID]bool),
 	}
 
 	// Load the prior persistence structures.
--- a/modules/renter/contractor/downloader.go
+++ b/modules/renter/contractor/downloader.go
@@ -6,6 +6,7 @@ import (
 	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/modules/renter/proto"
+	"github.com/NebulousLabs/Sia/types"
 )
 
 // An Downloader retrieves sectors from with a host. It requests one sector at
@@ -27,6 +28,7 @@ type Downloader interface {
 type hostDownloader struct {
 	downloader *proto.Downloader
 	contractor *Contractor
+	contractID types.FileContractID
 }
 
 // Sector retrieves the sector with the specified Merkle root, and revises
@@ -51,7 +53,13 @@ func (hd *hostDownloader) Sector(root crypto.Hash) ([]byte, error) {
 
 // Close cleanly terminates the download loop with the host and closes the
 // connection.
-func (hd *hostDownloader) Close() error { return hd.downloader.Close() }
+func (hd *hostDownloader) Close() error {
+	// release revising lock
+	hd.contractor.mu.Lock()
+	delete(hd.contractor.revising, hd.contractID)
+	hd.contractor.mu.Unlock()
+	return hd.downloader.Close()
+}
 
 // Downloader initiates the download request loop with a host, and returns a
 // Downloader.
@@ -70,6 +78,21 @@ func (c *Contractor) Downloader(contract modules.RenterContract) (Downloader, er
 		return nil, errTooExpensive
 	}
 
+	// acquire revising lock
+	c.mu.Lock()
+	alreadyRevising := c.revising[contract.ID]
+	if alreadyRevising {
+		c.mu.Unlock()
+		return nil, errors.New("already revising that contract")
+	}
+	c.revising[contract.ID] = true
+	c.mu.Unlock()
+	releaseLock := func() {
+		c.mu.Lock()
+		delete(c.revising, contract.ID)
+		c.mu.Unlock()
+	}
+
 	// create downloader
 	d, err := proto.NewDownloader(host, contract)
 	if proto.IsRevisionMismatch(err) {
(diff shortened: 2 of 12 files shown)


DavidVorick:
These commits are all labeled at 9 days ago?

modules/renter/contractor/contractor.go
```diff
@@ -133,6 +134,7 @@ func newContractor(cs consensusSet, w wallet, tp transactionPool, hdb hostDB, p
 
 		cachedRevisions: make(map[types.FileContractID]cachedRevision),
 		contracts:       make(map[types.FileContractID]modules.RenterContract),
+		downloaders:     make(map[types.FileContractID]*hostDownloader),
```
DavidVorick: What happens if the file contract id changes due to a renew? Host key might be a better key, but then if you're basically limited to exactly 1 contract per host, which might be limiting but ultimately I think we could work around any problems that it might introduce.

At some point in the future we're going to want to be able to use the funds from previous contracts (renewed contracts can still have money in them), but that seems like something the contractor should be worrying about, and not the renter. Since this is just an ID for the renter, it's probably okay to limit the renter to seeing one element per host. Especially since it's just a pointer and the renter is allowed to have multiple of those elements out at a time.

modules/renter/contractor/downloader.go
```diff
@@ -51,25 +58,63 @@ func (hd *hostDownloader) Sector(root crypto.Hash) ([]byte, error) {
...
+
+	if haveDownloader {
+		// increment number of clients
+		cachedDownloader.mu.Lock()
+		cachedDownloader.clients++
+		cachedDownloader.mu.Unlock()
+		return cachedDownloader, nil
+	} else if !haveContract {
+		return nil, errors.New("no record of that contract")
+	} else if height > contract.EndHeight() {
```
DavidVorick: This is going to result in bugs, b/c the host rejects contracts within some distance of the end height, such that it has time to submit the file contract revisions to the blockchain. I forget where that height is set, and maybe the host should be using the distance between WindowStart and WindowEnd to take care of things, but the current state of things is that you should probably not be modifying a contract within that period.

I looked, and it seems like the host will wait 20 blocks during dev, 4 blocks during testing, and 288 blocks in production. Might seem high, but I wanted to set the value high to be forgiving to our hosts in the early days, especially while we have so few and while they aren't actually making that much money anyway.

And,

DavidVorick: It's probably best to address this by adding a new function to the contract `RevisionLastCall` or something to that effect, so that the logic can be implemented in one place and used all over.

modules/renter/contractor/downloader.go
```diff
@@ -51,25 +58,63 @@ func (hd *hostDownloader) Sector(root crypto.Hash) ([]byte, error) {
...
+	} else if host.DownloadBandwidthPrice.Cmp(maxDownloadPrice) > 0 {
 		return nil, errTooExpensive
 	}
 
+	// acquire revising lock
+	c.mu.Lock()
+	alreadyRevising := c.revising[contract.ID]
+	if alreadyRevising {
+		c.mu.Unlock()
+		return nil, errors.New("already revising that contract")
```
DavidVorick: hmm. So if a download is already in progress, you still get an error returned for this contract?

Seems like it should block until the revising resource becomes available. And if that's going to be in place, clean-shutdown needs to be considered, since you don't want to wait for all of your open downloads to complete before things close all the way.

DavidVorick: oh wait were we going to handle that in the next PR?

modules/renter/contractor/editor.go
```diff
@@ -63,7 +63,13 @@ func (he *hostEditor) EndHeight() types.BlockHeight { return he.contract.EndHeig
 
 // Close cleanly terminates the revision loop with the host and closes the
 // connection.
-func (he *hostEditor) Close() error { return he.editor.Close() }
+func (he *hostEditor) Close() error {
+	// release revising lock
+	he.contractor.mu.Lock()
+	delete(he.contractor.revising, he.contract.ID)
+	he.contractor.mu.Unlock()
+	return he.editor.Close()
```
DavidVorick: lol a long chain of close calls happening here.

modules/renter/contractor/editor.go
```diff
@@ -145,6 +151,21 @@ func (c *Contractor) Editor(contract modules.RenterContract) (Editor, error) {
...
+		c.mu.Unlock()
+		return nil, errors.New("already revising that contract")
+	}
+	c.revising[contract.ID] = true
+	c.mu.Unlock()
+	releaseLock := func() {
+		c.mu.Lock()
+		delete(c.revising, contract.ID)
+		c.mu.Unlock()
+	}
```
DavidVorick: this violates our conventions. There's way too much code separating the acquisition and release of the lock, so it needs to be wrapped in a defer or somehow otherwise managed such that it's obvious that the resource is released correctly.

modules/renter/contractor/host_integration_test.go
```diff
@@ -741,3 +742,78 @@ func TestResync(t *testing.T) {
...
+	d4, err := c.Downloader(contract.ID)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// downloaders should match
+	if d4 == d1 {
+		t.Fatal("downloader should not have been cached after all clients were closed")
+	}
+}
```
DavidVorick: should let them finish the downloads and close them all the way, then verify that Close happened correctly.

lukechampine: they don't do any downloading, but I guess I can check that they are not in `c.downloaders` after closing.

modules/renter/contractor/contractor.go
lukechampine: a renew should invalidate any existing downloaders. The contractor will know both the old and new contract IDs, so it can easily loop through `downloaders` and invalidate any IDs that were renewed. This hasn't been implemented yet though.

modules/renter/contractor/downloader.go
lukechampine: even if you match the host settings, if you're desynched by a block the host may still reject it. And AFAIK, the 288-block buffer is specific to our host implementation and not part of the renter-host protocol. Practically speaking, it doesn't matter much if we reject it early on the renter side or wait until the host rejects it, so I'd prefer to be permissive on the renter side as long as there's a chance the host will accept it.

modules/renter/contractor/downloader.go
lukechampine: yeah, it's still necessary because of the download-upload interaction. This PR fixes the problem of downloads stepping on each other's toes, but you still need to prevent downloaders when there's an active uploader.

modules/renter/contractor/editor.go
lukechampine: yeah, the `proto.Editor` -> `contractor.Editor` chain is unfortunate. The former does all the real work, the latter just tracks the contracts and metrics (and, eventually, adds thread-safety).

modules/renter/contractor/editor.go
```diff
@@ -145,6 +151,21 @@ func (c *Contractor) Editor(contract modules.RenterContract) (Editor, error) {
 		}
 	}
 
+	// acquire revising lock
+	c.mu.Lock()
+	alreadyRevising := c.revising[contract.ID]
+	if alreadyRevising {
+		c.mu.Unlock()
+		return nil, errors.New("already revising that contract")
```
lukechampine: a `sync.Cond` might be appropriate here. That would allow us to efficiently wait for the revision to become available. You get the nice blocking properties without having to spawn background loops that send `Editor`s down channels and whatnot.

modules/renter/contractor/editor.go
lukechampine: hmm. I could do one of these:
```go
defer func() {
    if err != nil {
        c.mu.Lock()
        delete(c.revising, contract.ID)
        c.mu.Unlock()
    }
}()

DavidVorick: That's probably better. Have to make sure you are handling errors correctly, but you have to do that anyway.

modules/renter/contractor/contractor.go
lukechampine: ok. I'm gonna leave it for now because the change is more involved than you might think (a lot of downloader code only has access to to the ID, not the host key). I do agree that we should move this field (and possibly a few others) to using the host key.

PR #1454, opened 2016-10-18
Rc2 tweaks

DavidVorick:
A quick change to the host to change sector file permissions from 0700 to 0400, tests seem to be coping with it well. Also added a logging statement to the storage manager if a sector write fails, as currently that failure is recorded but no reason is given for the failure.

Finally, I updated the hostdb to de-prioritize hosts that don't have much storage remaining. This is because those hosts have limited utility, especially given that contract fees are so high. I'm not sure if I calibrated the numbers correctly, the first penalty hits at 200GB (it's not too severe), and then below 50GB you get hit really hard. Below 1GB you are essentially ignored as a host.

I'm re-running the hostdb on my local machine to make sure this doesn't affect prices too dramatically. I'm not actually sure what percentage of hosts have more than 200GB available, but my initial checks suggest it's well above 50%, which is where it needs to be for this change to not be disruptive.

### DIFF
--- a/modules/renter/hostdb/hostweight.go
+++ b/modules/renter/hostdb/hostweight.go
@@ -3,6 +3,7 @@ package hostdb
 import (
 	"math/big"
 
+	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/types"
 )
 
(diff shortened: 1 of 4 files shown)


modules/renter/hostdb/hostweight.go
```diff
@@ -47,6 +63,29 @@ func calculateHostWeight(entry hostEntry) (weight types.Currency) {
...
+		weight = weight.Div64(5) // 95x total penalty
+	}
+	if entry.RemainingStorage < 10*requiredStorage {
+		weight = weight.Div64(6) // 570x total penalty
+	}
+	if entry.RemainingStorage < 5*requiredStorage {
+		weight = weight.Div64(10) // 5700x total penalty
+	}
+	if entry.RemainingStorage < requiredStorage {
+		weight = types.NewCurrency64(1) // Worthelss host.
```
lukechampine: typo

there's probably a cleaner function for this weighting, but maybe not worth worrying about right now

DavidVorick:
Addressed comments via rebase, also added an extra commit to tweak the default values.

- default collateral is being increased b/c hosts now have more expectations around needing to have money
- default storage price has been reduced because the current defaults are no longer competitive. Also, I'm expecting the price of the coin to jump following our campaigns over the next two weeks.
- default upload price reduced to near-zero, because hosts should be happy that they are getting storage, and generally speaking they will have plenty of downstream (which is what is consumed on the host-side)
- revision submission buffer dropped to one day to reflect the size of the storage window. Hosts are expected to have at least some uptime every day if they run zero-risk of losing contracts, seems reasonable to me. The occasional day or two of downtime will affect revenue, but will not make the host unprofitable.

PR #1457, opened 2016-10-24
cache editors for reuse

lukechampine:


modules/renter/contractor/downloader.go
```diff
@@ -109,11 +111,15 @@ func (c *Contractor) Downloader(id types.FileContractID) (Downloader, error) {
...
-	}
+
+	// release lock early if function returns an error
+	defer func() {
+		if err != nil {
+			c.mu.Lock()
+			delete(c.revising, contract.ID)
+			c.mu.Unlock()
+		}
+	}()
```
DavidVorick: This code had not already been converted? I thought we made this change already.

lukechampine: yeah, turns out I only fixed it for the editor, not the downloader

PR #1458, opened 2016-10-24
invalidate editors/downloaders during renew

lukechampine:
`managedRenewContracts` will now mark each editor/downloader as invalid, then grab the latest revision of each contract and renew them. During renewal, attempting to create an editor or downloader will return an error. Trying to use an invalid editor/downloader will also result in an error, so the renter will just continually attempt to create editors and fail until the renewal has completed.

I'm not super happy with the locking here; a redesign is probably warranted. Specifically, the interaction between `managedRenewContracts` and `editor.Close` is ugly because `Close` locks the contractor. Ideally, editors/downloaders wouldn't lock the contractor at all, but that requires an alternative method of updating the contract after each revision. We can revisit this later.

### DIFF
--- a/modules/renter/contractor/contractor.go
+++ b/modules/renter/contractor/contractor.go
@@ -47,6 +47,7 @@ type Contractor struct {
 	editors         map[types.FileContractID]*hostEditor
 	lastChange      modules.ConsensusChangeID
 	renewHeight     types.BlockHeight             // height at which to renew contracts
+	renewing        map[types.FileContractID]bool // prevent revising during renewal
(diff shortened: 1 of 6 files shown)


DavidVorick:
This needs to be run through the ant-farm

lukechampine:
antfarm discovered a deadlock, which I eventually fixed (after much weeping and gnashing of teeth). antfarm is performing well now, except that renter files are not updated after renewals, causing them to fail.

the locking design of the contractor should be refactored, it's a mess. I think per-contract locks would help.

PR #1459, opened 2016-10-25
Expand Host Contracts Visibility

DavidVorick:
The host module now has a method `Contracts()` which is supported by the API, and by `siac`. It's all experimental, and we're not committing to compatibility for this. We want to get the first commit into the master branch though, I guess I'll open that in a different PR.

With this command you can see all of the contracts that the host has formed and has failed to form, as well as some information like how much potential revenue there is, and how many sectors there are.

### DIFF
--- a/modules/host/host.go
+++ b/modules/host/host.go
@@ -64,6 +64,7 @@ package host
 // TODO: update_test.go has commented out tests.
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"net"
@@ -77,6 +78,8 @@ import (
 	"github.com/NebulousLabs/Sia/persist"
 	siasync "github.com/NebulousLabs/Sia/sync"
 	"github.com/NebulousLabs/Sia/types"
+
+	"github.com/NebulousLabs/bolt"
 )
 
 const (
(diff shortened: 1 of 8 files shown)


modules/host/host.go
```diff
@@ -318,6 +321,68 @@ func (h *Host) Close() error {
...
+			hc.LockedCollateral = so.LockedCollateral
+			hc.DownloadRevenue = so.PotentialDownloadRevenue
+			hc.RiskedCollateral = so.RiskedCollateral
+			hc.StorageRevenue = so.PotentialStorageRevenue
+			hc.TransactionFeesPaid = so.TransactionFeesAdded
+			hc.UploadRevenue = so.PotentialUploadRevenue
+
+			hc.FileContractConfirmed = so.OriginConfirmed
+			hc.FileContractRevisionConfirmed = so.RevisionConfirmed
+			hc.StorageProofConfirmed = so.ProofConfirmed
```
lukechampine: maybe use `hc := modules.HostContract{x, y, z...}` to set all these fields?

lukechampine:
can this be closed in light of the new contract manager?

DavidVorick:
Not yet, it's got some features that are actually really important + useful for debugging. This PR is open because it's the only way to inspect contract state in production. The other PR doesn't have anything like this yet.

DavidVorick:
closing this, have full plans to replace it entirely while doing live-testing on the hosts that I'm running.

PR #1460, opened 2016-10-25
ContractCount refers to active contracts

DavidVorick:
this commit is in another PR, but as we aren't intending to merge that PR tonight, I wanted to pull this into its own PR. The commit hashes should match.


DavidVorick:
commit hashes don't match anymore b/c I had to deal with a ndf on appveyor >.<

PR #1461, opened 2016-10-25
host can batch sector add calls during renew

DavidVorick:
A severe performance bottleneck previously existed during renew, meaning that
hosts with slower drives would stall out / fail if renewing file contracts
larger than 1 GB. This removes that bottleneck, hosts should now be good for
file contracts up to 100GB.

### DIFF
--- a/modules/host/storagemanager/sector.go
+++ b/modules/host/storagemanager/sector.go
@@ -74,6 +74,10 @@ var (
 	// hovering around 95% capacity and rarely over 98% or under 90% capacity.
 	errInsufficientStorageForSector = errors.New("not enough storage remaining to accept sector")
 
+	// errNonVirtualSectorInBatch is thrown if a batch-add for sectors has a
+	// physical sector in it.
+	errNonVirtualSectorInBatch = errors.New("sector added in batch which is not a virtual sector")
+
 	// errMaxVirtualSectors is returned when a sector cannot be added because
 	// the maximum number of virtual sectors for that sector id already exist.
 	errMaxVirtualSectors = errors.New("sector collides with a physical sector that already has the maximum allowed number of virtual sectors")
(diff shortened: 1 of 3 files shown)


modules/host/storagemanager/sector.go
```diff
@@ -238,6 +242,112 @@ func (sm *StorageManager) AddSector(sectorRoot crypto.Hash, expiryHeight types.B
...
+				break
+			}
+			err = bsu.Put(sm.sectorID(root[:]), usageBytes)
+			if err != nil {
+				break
+			}
+		}
+		if err != nil {
+			// Need to remove the virtual sectors that are added, as there has
+			// been an error.
```
lukechampine: couldn't you induce a rollback by returning an error inside the bolt function? Seems safer than writing the rollback logic manually. Of course, you can only do this if the bolt function has no other side effects.
```go
if len(usage.Expiry) >= maximumVirtualSectors {
	// returning non-nil error, so whole tx will be reverted
	return errMaxVirtualSectors
}
```

DavidVorick: oh yeah, forgot that this function entirely happens within the bolt Update call. Pretty sure there are no side-effects, because every sector is supposed to be virtual.

derp, could have saved my self some time.

modules/host/storagemanager/sector.go
```diff
@@ -238,6 +242,60 @@ func (sm *StorageManager) AddSector(sectorRoot crypto.Hash, expiryHeight types.B
...
+			// return an error. The host handles virtual sectors differently
+			// from physical sectors and therefore needs to limit the number of
+			// times that the same data can be uploaded to the host. For
+			// renters that are properly using encryption and are using
+			// sane/reasonable file contract renewal practices, this limit will
+			// never be reached (sane behavior will cause 3-5 at an absolute
+			// maximum, but the limit is substantially higher).
+			if len(usage.Expiry) >= maximumVirtualSectors {
+				err = errMaxVirtualSectors
+				break
```
lukechampine: this sets the error, but then breaks, and the function returns nil

DavidVorick: good catch

PR #1462, opened 2016-10-25
Renter download loop

DavidVorick:
I added a loop to the renter download code so that large files could be downloaded despite host timeouts.

### DIFF
--- a/modules/renter/contractor/downloader.go
+++ b/modules/renter/contractor/downloader.go
@@ -3,8 +3,10 @@ package contractor
 import (
 	"errors"
 	"sync"
+	"time"
 
 	"github.com/NebulousLabs/Sia/crypto"
+	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/modules/renter/proto"
 	"github.com/NebulousLabs/Sia/types"
 )
@@ -28,12 +30,14 @@ type Downloader interface {
 // It implements the Downloader interface. hostDownloaders are safe for use by
 // multiple goroutines.
 type hostDownloader struct {
-	clients    int // safe to Close when 0
-	contractID types.FileContractID
-	contractor *Contractor
-	downloader *proto.Downloader
-	invalid    bool // true if invalidate has been called
-	mu         sync.Mutex
+	clients      int // safe to Close when 0
+	contractID   types.FileContractID
+	contractor   *Contractor
+	downloader   *proto.Downloader
+	hostSettings modules.HostExternalSettings
+	invalid      bool   // true if invalidate has been called
+	speed        uint64 // Bytes per second.
+	mu           sync.Mutex
 }
 
 // invalidate sets the invalid flag and closes the underlying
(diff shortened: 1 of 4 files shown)


DavidVorick:
ah whoops, grabbed the batch commit.

modules/renter/contractor/downloader.go
```diff
@@ -145,10 +170,11 @@ func (c *Contractor) Downloader(id types.FileContractID) (_ Downloader, err erro
...
 	hd := &hostDownloader{
-		clients:    1,
-		contractID: contract.ID,
-		contractor: c,
-		downloader: d,
+		clients:      1,
+		contractID:   contract.ID,
+		contractor:   c,
+		downloader:   d,
+		hostSettings: host.HostExternalSettings,
```
lukechampine: including the whole `Host` might be best, especially for things like debugging. easy to change later though

modules/renter/download.go
```diff
@@ -121,16 +121,15 @@ func (d *download) getPiece(chunkIndex, pieceIndex uint64) []byte {
 // run performs the actual download. It spawns one worker per host, and
 // instructs them to sequentially download chunks. It then writes the
 // recovered chunks to w.
-func (d *download) run(w io.Writer) error {
-	var received uint64
-	for i := uint64(0); received < d.fileSize; i++ {
+func (d *download) run(chunkIndex uint64, received uint64, w io.Writer) (uint64, uint64, error) {
```
lukechampine: what do these uint64s represent?

DavidVorick: `run` has an internal loop to track which chunk it's currently on. Because it now gets called multiple times per iteration, that tracking needs to be made external and provided as input

modules/renter/download.go
```diff
@@ -256,22 +207,91 @@ func (r *Renter) Download(path, destination string) error {
...
+			d.hosts = hosts
+			r.mu.Unlock(lockID)
+			defer func() {
+				// Clear the set of hosts from the downloader.
+			}()
+
+			// Perform download.
+			chunkIndex, received, err = d.run(chunkIndex, received, f)
+			if err != nil {
+				return errors.New("download iteration did not complete")
```
lukechampine: `run` can return severe (i.e. unrecoverable) errors, such as being unable to write to the file handle. It's fine for now; those errors should be very rare.

Instead of a sentinel error, I would prefer to return a `(bool, error)`, where the bool indicates whether another iteration is required. Then the logic below becomes:
```go
if finished {
    break
}
if err != nil {
    // severe error; sleep before trying again
    time.Sleep(time.Second * 90)
}
```

modules/renter/download.go
```diff
@@ -256,22 +207,91 @@ func (r *Renter) Download(path, destination string) error {
...
+			// Check that this host set is sufficient to download the file.
+			err := checkHosts(hosts, file.erasureCode.MinPieces(), file.numChunks())
+			if err != nil {
+				return err
+			}
 
+			// We have a new set of hosts, update the downloader to reflect that.
+			lockID := r.mu.Lock()
+			// TODO: Not quite sure how thread safe this is.
+			d.hosts = hosts
```
lukechampine: `d.hosts` is never accessed by goroutines other than the download loop, so a lock shouldn't be necessary here.

modules/renter/download.go
```diff
@@ -256,22 +207,91 @@ func (r *Renter) Download(path, destination string) error {
...
+			r.mu.Unlock(lockID)
+			defer func() {
+				// Clear the set of hosts from the downloader.
```
lukechampine: seems like this was left over? `d.hosts = hosts` will clear the set, and they are closed with a `defer` in the loop above.

modules/renter/download.go
lukechampine: sorry, I meant the returned values. They should be named, or the docstring should at least be updated

modules/renter/download.go
DavidVorick: filesystem errors are probably not that rare, I'll add better handling

PR #1488, opened 2016-10-31
Wallet addrs

### LINKED ISSUE #1066: Negative addresses remaining in wallet
```
$ siac wallet seeds
Primary Seed: ***
Addresses Remaining -16891
All Seeds:
***
```

lukechampine:
Fixes #1066. `/wallet/seeds` will now always report 1 address remaining. This value should be ignored.

We could alternatively have it return the seed progress in place of `AddressesRemaining`, but this would change the semantics of the field without changing its name, so it's probably a bad idea.

### DIFF
--- a/api/wallet.go
+++ b/api/wallet.go
@@ -277,7 +277,7 @@ func (api *API) walletSeedsHandler(w http.ResponseWriter, req *http.Request, _ h
 	}
 
 	// Get the primary seed information.
-	primarySeed, progress, err := api.wallet.PrimarySeed()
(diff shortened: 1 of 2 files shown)


DavidVorick:
for some reason it looks like the appveyor build did not run here. Not that any of this code should cause appveyor problems, but weird that it didn't run.

DavidVorick:
ah shoot we needed to update the API documentation.

lukechampine:
It isn't running because this because we haven't configure it to build the wallet branch. I'll fix that

PR #1492, opened 2016-11-02
Include encoding errors in API response

### LINKED ISSUE #1342: API encoding errors
If the API fails to encode a response, no details about the encoding failure are retrievable. The error should be included in the HTTP error response. `build.Critical` is another reasonable option, since response encoding is something that should really never fail.
Also note that since these errors are not JSON, siac fails to properly parse them.

VoidingWarranties:
Closes #1342 

@lukechampine lmk what you think of manually encoding the JSON response.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 	"strings"
 
+	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/modules"
 
 	"github.com/julienschmidt/httprouter"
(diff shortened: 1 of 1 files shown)


lukechampine:
My feeling is that `WriteError` should never fail unless we somehow break the way `api.Error` is marshaled. `WriteJSON` is much more susceptible to failure because it is marshaling an arbitrary type. So I think my preference would be to give up entirely in `WriteError` (just call `build.Critical`, since it's a developer error) and call `WriteError` from `WriteJSON` if `WriteJSON` fails.

lukechampine:
I take it back -- in either case, encoding should never fail. If it does, it's a developer error, so we should use `build.Critical`. There's no reason to send a response at all because we're in "undefined behavior" territory -- something is egregiously wrong and needs to be fixed ASAP. So I would drop the response writing and just leave the `build.Critical`s, along with a comment explaining that it can only occur in the event of developer error.

VoidingWarranties:
Agreed. Updated with @lukechampine's suggested change.

PR #1493, opened 2016-11-02
update siac version to show daemon version and siac version

### LINKED ISSUE #1490: siac version should show the daemon version as well
If siac is unable to connect to the daemon, it does not report an error and it does not show the daemon version.

johnathanhowell:
Fixes #1490

### DIFF
--- a/siac/daemoncmd.go
+++ b/siac/daemoncmd.go
@@ -48,13 +48,13 @@ type daemonVersion struct {
 
 // version prints the version of siac and siad.
 func versioncmd() {
+	fmt.Println("Sia Client v" + build.Version)
 	var versioninfo daemonVersion
 	err := getAPI("/daemon/version", &versioninfo)
 	if err != nil {
 		fmt.Println("Could not get daemon version:", err)
 		return
 	}
-	fmt.Println("Sia Client v" + build.Version)
 	fmt.Println("Sia Daemon v" + versioninfo.Version)
 }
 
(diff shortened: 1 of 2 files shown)


siac/daemoncmd.go
```diff
@@ -27,13 +28,36 @@ var (
...
+
+// version prints the version of siac and siad.
+func versioncmd() {
+	var versioninfo daemonVersion
+	err := getAPI("/daemon/version", &versioninfo)
+	if err != nil {
+		fmt.Println("Could not get daemon version:", err)
+		return
+	}
+	fmt.Println("Sia Client v" + build.Version)
```
VoidingWarranties: If the Sia Daemon is not running, this command will exit before displaying `siac`'s version. Perhaps this line could be moved to the top of the func so as to always print the `siac` version.

johnathanhowell: yeah, that would be better.

siac/daemoncmd.go
```diff
@@ -27,13 +28,36 @@ var (
...
 )
 
 type updateInfo struct {
 	Available bool   `json:"available"`
 	Version   string `json:"version"`
 }
 
+type daemonVersion struct {
+	Version string
+}
```
VoidingWarranties: Also, I don't think it's strictly necessary, but it might be good practice to have the JSON tags for the version field as well.

VoidingWarranties: Ah, right!

PR #1494, opened 2016-11-03
Sweep Siafunds

lukechampine:
Previously `SweepSeed` did not support sweeping siafunds. This PR adds support for sweeping both coins and funds.

When sweeping coins, it is assumed that the user will want to abort the sweep if the value of the coins is less than the cost of the transaction fees. But when sweeping funds, it is assumed the user will never want to abort the sweep for this reason, since funds are much more valuable than coins.

A related complication arises if only funds are found in the scan. When sweeping coins, we can subtract the transaction fee from the total value swept. This is convenient because it means we don't need to touch the existing wallet outputs at all; you can sweep coins even if your balance is zero.
However, this is not the case if only funds are found (or if the value of coins found is less than the transaction fee). In this case, we must use the existing wallet outputs to pay for the transaction fee. To make this easier, I first refactored `SweepSeed` to use a transaction builder before adding siafund support.

### DIFF
--- a/modules/wallet/seed.go
+++ b/modules/wallet/seed.go
@@ -376,15 +376,15 @@ func (w *Wallet) SweepSeed(seed modules.Seed) (coins, funds types.Currency, err 
 		sk := generateSpendableKey(seed, sfo.seedIndex)
 		addSignatures(&txn, types.FullCoveredFields, sk.UnlockConditions, crypto.Hash(sfo.id), sk)
 	}
-	// if inputs were added by the transaction builder, sign those as well
+	// Usually, all the inputs will come from swept outputs. However, there is
+	// an edge case in which inputs will be added from the wallet. To cover
+	// this case, we iterate through the SiacoinInputs and add a signature for
+	// any input that belongs to the wallet.
 	w.mu.RLock()
 	for _, input := range txn.SiacoinInputs {
-		key, ok := w.keys[input.UnlockConditions.UnlockHash()]
-		if !ok {
-			w.mu.RUnlock()
-			return types.Currency{}, types.Currency{}, errors.New("transaction builder added an input that it cannot sign")
+		if key, ok := w.keys[input.UnlockConditions.UnlockHash()]; ok {
+			addSignatures(&txn, types.FullCoveredFields, input.UnlockConditions, crypto.Hash(input.ParentID), key)
 		}
-		addSignatures(&txn, types.FullCoveredFields, input.UnlockConditions, crypto.Hash(input.ParentID), key)
 	}
 	w.mu.RUnlock()
 
--- a/modules/wallet/seed.go
+++ b/modules/wallet/seed.go
@@ -349,7 +349,7 @@ func (w *Wallet) SweepSeed(seed modules.Seed) (coins, funds types.Currency, err 
 		})
 		err = tb.FundSiacoins(estFee)
 		if err != nil {
-			return types.Currency{}, types.Currency{}, err
+			return types.Currency{}, types.Currency{}, errors.New("couldn't pay transaction fee on swept funds: " + err.Error())
 		}
 
 	case !coins.IsZero() && !funds.IsZero():
--- a/modules/wallet/seed.go
+++ b/modules/wallet/seed.go
@@ -254,14 +254,14 @@ func (w *Wallet) LoadSeed(masterKey crypto.TwofishKey, seed modules.Seed) error 
 // transaction that transfers them to the wallet. Note that this incurs a
 // transaction fee. It returns the total value of the outputs, minus the fee.
 // TODO: support SiafundOutputs too
-func (w *Wallet) SweepSeed(seed modules.Seed) (payout types.Currency, err error) {
+func (w *Wallet) SweepSeed(seed modules.Seed) (coins, funds types.Currency, err error) {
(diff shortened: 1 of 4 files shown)


modules/wallet/seed.go
```diff
@@ -287,45 +287,98 @@ func (w *Wallet) SweepSeed(seed modules.Seed) (payout types.Currency, err error)
...
+		// if we're sweeping funds but not coins, add a siafund output for
+		// them. This is tricky because we still need to pay for the
+		// transaction fee, but we can't simply subtract the fee from the
+		// output value like we can with swept coins. Instead, we need to fund
+		// the fee using the existing wallet balance.
+		tb.AddSiafundOutput(types.SiafundOutput{
+			Value:      funds,
+			UnlockHash: uc.UnlockHash(),
+		})
+		err = tb.FundSiacoins(estFee)
```
lukechampine: it occurs to me now that even if `coins < estFee`, you could still use the value of the coins to offset a portion of the fee. I don't know if it's significant enough to worry about though.

DavidVorick: Probably better to tell the user that they need more SC, you don't want SF stuck in limbo

modules/wallet/seed.go
```diff
@@ -287,45 +287,109 @@ func (w *Wallet) SweepSeed(seed modules.Seed) (payout types.Currency, err error)
...
+		if err != nil {
+			return types.Currency{}, types.Currency{}, err
```
DavidVorick: should be adding context to this error. It's probably something like "output not found" or so. Doesn't really help us figure out where the code failed.

DavidVorick:
Can you add two more tests for sweeping siafunds?

1. A test where siafunds have been spent a few times over. This makes sure that the code properly handles sweeping already-spent siafunds. (send them to yourself a few times, and send some to the void a few times).

2. A test where the seed being swept has both siafunds and siacoins. And perhaps make this two tests, one where there are not enough siacoins to cover the fees, and a second one where there are definitely enough siacoins to cover the fees.

modules/wallet/transactionbuilder.go
```diff
@@ -555,7 +558,10 @@ func (tb *transactionBuilder) Sign(wholeTransaction bool) ([]types.Transaction,
 	}
 	for _, inputIndex := range tb.siafundInputs {
 		input := tb.transaction.SiafundInputs[inputIndex]
-		key := tb.wallet.keys[input.UnlockConditions.UnlockHash()]
+		key, ok := tb.wallet.keys[input.UnlockConditions.UnlockHash()]
+		if !ok {
+			return nil, errors.New("transaction builder added an input that it cannot sign")
```
DavidVorick: Is this a codepath that you have reached? I agree with the code being here, just wondering if you added it because of issues, or if you're just programming defensively.

modules/wallet/seed_test.go
```diff
@@ -278,57 +278,71 @@ func TestSweepSeedFunds(t *testing.T) {
...
+			t.Fatal(err)
+		}
+		wt.addBlockNoPayout()
+	}
+	// send some funds to the void
+	_, err = wt.wallet.SendSiafunds(types.NewCurrency64(10), types.UnlockHash{})
+	if err != nil {
+		t.Fatal(err)
+	}
+	wt.addBlockNoPayout()
```
DavidVorick: Would actually like to see this split into its own test. As far as work goes, you can do most of the heavy lifting via copy-paste.

I know that results in a ton of testing code, but it makes implementation easy, and more importantly you end up with tests that can have focused goals + purposes, which later on makes debugging and modification a million times easier.

In the old storage manager I had a bunch of really large tests, basically doing what you've just done above but taking it to a bigger extreme (some tests 300+ lines, some 900+ lines). Then when something broke, it was basically necessary to read the whole test to figure out where things had gone wrong, and then changing anything could potentially affect the entire rest of the test.

In host2 I've found it to be much more manageable to break them apart. The resulting files are massive, but when a test breaks its usually only 1 or 2 of them, and so far I've only had to spend more than a few minutes fixing it up just 1 time.

It's more important for the host because I'm doing a ton of dependency injection, but now it's just 1 dependency injection per test. But I think the same applies to the wallet nonetheless.

I think we will prefer maintaining these tests when there's tons of duplicate code but each test individually is just large enough to accomplish a single goal.

lukechampine: do you think it's worth trying to achieve some DRYness with helper functions, like we've done in the past, or just copy-paste everything?

Another incentive to combine tests is to avoid doing the same work twice, e.g. if a test takes 8 seconds to set up. But you can usually mitigate that via `t.Parallel`.

lukechampine: also, I assume you mean:
- One test that just sweeps siafunds
- One test that sends funds to itself+void, and then sweeps

DavidVorick: That is what I mean.

DRY is good in testing but only if you aren't too creative about it. If you end up changing the setup code, you risk breaking any test that's depending on it. But if it's all copy-paste, that risk isn't there. It also means having to learn more helper functions which might be spread across multiple files and might have their own helper functions (api package testing is like this), and that can get annoying / confusing / frustrating if you are just trying to understand / fix a single test.

PR #1496, opened 2016-11-08
Add API endpoints for InitFromSeed and SweepSeed

lukechampine:
open to other suggestions for the endpoint names and siac commands. I'm not a huge fan of either of them.


DavidVorick:
Would it make sense to add a 'Seed' field to the `/wallet/init` endpoint? Would we want to use `/wallet/init/new` and `wallet/init/seed`? I think I am most in favor of just adding a field to the existing endpoint.

For sweep, `wallet/sweep` seems okay to me. @VoidingWarranties what are your thoughts on the endpoints?

api/wallet.go
```diff
@@ -65,6 +65,13 @@ type (
 		AllSeeds           []string `json:"allseeds"`
 	}
 
+	// WalletSweepPOST contains the coins and funds discovered by a call to
+	// /wallet/sweep.
```
DavidVorick: comment should make it a little more clear that this is the return value, not the input value.

Though iirc by convention inputs are always path parameters

api/wallet.go
```diff
@@ -203,6 +210,32 @@ func (api *API) walletInitHandler(w http.ResponseWriter, req *http.Request, _ ht
...
+		return
+	}
+
+	err = api.wallet.InitFromSeed(encryptionKey, seed)
+	if err != nil {
+		WriteError(w, Error{"error when calling /wallet/initseed: " + err.Error()}, http.StatusBadRequest)
+		return
+	}
+	WriteSuccess(w)
+}
```
DavidVorick: Need to write testing around this, I do not think that this function will work.

Remember to add testing to every PR with code changes.

api/wallet.go
```diff
@@ -364,6 +397,30 @@ func (api *API) walletSiafundsHandler(w http.ResponseWriter, req *http.Request,
...
+	coins, funds, err := api.wallet.SweepSeed(seed)
+	if err != nil {
+		WriteError(w, Error{"error when calling /wallet/sweep: " + modules.ErrBadEncryptionKey.Error()}, http.StatusBadRequest)
+		return
+	}
+	WriteJSON(w, WalletSweepPOST{
+		Coins: coins,
+		Funds: funds,
+	})
+}
```
DavidVorick: If I understand correctly, there is currently no support for sweeping a key (such as a siag key)?

Do we want to have those under the same endpoints, or do we want to specify them using two separate endpoints? `/wallet/sweep/seed` and `/wallet/sweep/siag` and `/wallet/sweep/033x`?

We should be able to sweep all types of keys that we support loading, which I believe is the above 3.

lukechampine:
Agreed re: `/wallet/init` taking `seed` as a param and `/wallet/sweep` having multiple subroutes. I forgot that siag and 033x keys could be swept as well.

VoidingWarranties:
Also agreed with adding field to `/wallet/init`. In the documentation we should note the version number the field was added in.

doc/api/Wallet.md
```diff
@@ -168,10 +167,14 @@ standard success or error response. See
...
 #### /wallet/init [POST]
 
-initializes the wallet. After the wallet has been initialized once, it does not
-need to be initialized again, and future calls to /wallet/init will return an
-error. The encryption password is provided by the api call. If the password is
-blank, then the password will be set to the same as the seed.
+initializes the wallet. After the wallet has been initialized once, it does
+not need to be initialized again, and future calls to /wallet/init will return
+an error. The encryption password is provided by the api call. If the password
+is blank, then the password will be set to the same as the seed.  If the seed
```
VoidingWarranties: Two spaces after period.

VoidingWarranties:
Can you please update `/doc/API.md` as well?

DavidVorick:
re `/wallet/init`, I might backtrack on the idea of adding the seed as an additional field.

Because, we want to be able to init from

+ An existing seed
+ An older backup file
+ A newer backup file

`/wallet/init/seed`
`/wallet/init/backup`

It's probably better if we can have logic figure out which backup file is being loaded. iirc the backup files are versioned, so this should be easy to do.

DavidVorick:
backup files have a list of seeds + keys

modules/wallet.go
```diff
@@ -295,6 +292,22 @@ type (
...
+		// along with a uint64 indicating how many addresses may be safely
+		// generated from the seed.
+		PrimarySeed() (Seed, uint64, error)
+
+		// SweepSeed scans the blockchain for outputs generated from seed and
+		// creates a transaction that transfers them to the wallet. Note that
+		// this incurs a transaction fee. It returns the total value of the
+		// outputs, minus the fee. If only siafunds were found, the fee is
+		// deducted from the wallet.
+		SweepSeed(seed Seed) (coins, funds types.Currency, err error)
```
DavidVorick: :+1:

DavidVorick:
cover-walletinitseedhandler
cover-walletsweepseedhandler

Shouldn't be too hard to get coverage on these lines.

api/wallet_test.go
```diff
@@ -240,6 +325,71 @@ func TestIntegrationWalletGETSiacoins(t *testing.T) {
...
+	if err == nil {
+		t.Fatal("expected error, got nil")
+	}
+
+	// Call /wallet/sweep/seed with an invalid seed
+	qs.Set("seed", "foo")
+	err = st.postAPI("/wallet/sweep/seed", qs, &wsp)
+	if err == nil {
+		t.Fatal("expected error, got nil")
+	}
```
DavidVorick: In the future this should probably be a separate test, as it's polluting this test by having + executing on multiple ideas. Much easier to debug problems with tests if each test has a single thing that it's focused on contributing.

PR #1498, opened 2016-11-15
Sweep 033x and siag keys

lukechampine:
I considered combining `seedScanner` and `keyScanner`, but decided against it for two reasons: 1) they are sufficiently different in functionality (`seedScanner` also tracks `largestIndexSeen`) and 2) seeds are now the primary basis of wallet keys, so it makes sense for them to have their own code path, separate from 033x and siag (which are special cases).


modules/wallet/scan.go
```diff
@@ -1,15 +1,20 @@
...
 	"github.com/NebulousLabs/Sia/build"
+	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/types"
+	"github.com/NebulousLabs/bolt"
 )
 
 const scanMultiplier = 4 // how many more keys to generate after each scan iteration
 
+const outputSize = 350 // approx. size in bytes of an output and accompanying signature
```
DavidVorick: This is an excellent number to have.

Also, ouch. If the encoding had been better designed, it'd be closer to 100 bytes. The problem is all the 8-byte prefixes for empty arrays in the transaction object and signature object. I like it's like 10 each. I would consider that more a shortcoming of the transaction layout than the encoding scheme.

DavidVorick:
There doesn't seem to be testing in the wallet package, just in the api package.

modules/wallet/scan.go
lukechampine: More precise testing shows that the value is closer to 200 bytes, with a small amount of variance (1-3 bytes) depending on the value of the output. 250 is probably a safe value.

lukechampine:
Closing this for now. If there is sufficient interest we can reopen it. I'll leave the branch on GitHub (and I have a local copy) so we can easily restore it.

PR #1499, opened 2016-11-16
Add Cmp64, Equals, and Equals64

lukechampine:
Curious to see what people think about these. Took like 2 minutes to make these changes using `gofmt -w -r` so I won't be offended if we decide to reject them. If we do decide to add them, I'll write tests for the new functions.

I think there's a good case for `Cmp64`, but less so for `Equals`. The reason being that `Equals` exists alongside `Cmp`, which could be confusing since it's just a special case. We could go all the way and replace `Cmp` with a handful of comparison functions, but then you need separate functions for both `<` and `<=`. So you'd wind up with:
- `Equals`
- `LessThan`
- `LessThanOrEqualTo` (`AtMost`?)
- `GreaterThan`
- `GreaterThanOrEqualTo` (`AtLeast`?)

...plus `64` variants for all of them. That feels like too much, but I'm not totally sure. As far as readability goes, I'm sure we can agree that `x.Equals(y)` is much nicer than `x.Cmp(y) == 0`.

### DIFF
--- a/api/wallet_test.go
+++ b/api/wallet_test.go
@@ -174,10 +174,10 @@ func TestIntegrationWalletGETSiacoins(t *testing.T) {
 	if wg.ConfirmedSiacoinBalance.Cmp(types.CalculateCoinbase(1)) != 0 {
 		t.Error("reported wallet balance does not reflect the single block that has been mined")
 	}
-	if wg.UnconfirmedOutgoingSiacoins.Cmp(types.NewCurrency64(0)) != 0 {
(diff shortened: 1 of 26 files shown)


DavidVorick:
Bytes package supports Compare and Equal, and nothing else iirc, so there is a precedent.

lukechampine:
I didn't write the tests yet lol

DavidVorick:
ah oops, you are right.

DavidVorick:
please follow up with some tests, especially since this package has already done a pretty good job of being at 100% coverage.

PR #1500, opened 2016-11-16
implement LTS release fetching

lukechampine:
note the assumptions made about the lts tag format.

### DIFF
--- a/siad/server.go
+++ b/siad/server.go
@@ -129,7 +129,7 @@ func latestLTS(releases []githubRelease) (githubRelease, error) {
 	// filter the releases to exclude non-LTS releases
 	var ltsReleases []githubRelease
 	for _, r := range releases {
-		if strings.Contains(r.TagName, "lts") {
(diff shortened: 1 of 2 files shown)


DavidVorick:
Seems good to me. If we sort alphabetically:

lts-v1.3.2
lts-v1.0.4
v1.4.3
v1.4.2
v1.4.1
etc

Makes the LTS really easy to find.

lukechampine:
Another thing I realized is that this commit will need to be added to every lts branch. Alternatively, we could have only a single lts branch. That would be fine if LTSs never overlap.

DavidVorick:
I was definitely thinking that we only have a single LTS branch.

Large upgrades (e.g. 1.0.4 -> 1.3.1) might be rough, but if we're careful with how we backport, I think we can avoid headaches.

DavidVorick:
I think we should change the parent branch to `lts`, directly forked from our current master.

lukechampine:
ok. After this PR is merged I'll rename the branch. (Easier than closing+reopening)

PR #1501, opened 2016-11-17
Replace the Host Storage Manager

DavidVorick:
Not everything is complete, and there's no upgrade path from host -> host2. So it's going into staging instead of master.

We should merge the wallet branch into staging as well. Then we can all be working on the same base as we charge towards v1.1.


modules/host/contractmanager/storagefoldershrink_bench_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+	// Benchmark how long it takes to unset the bits.
+	for i := 0; i < b.N; i++ {
+		// Create a usage array with all bits set.
+		b.StopTimer()
+		usageArray := make([]uint64, 10e6)
+		for i := 0; i < len(usageArray); i++ {
+			usageArray[i] = base
+		}
+		b.StartTimer()
```
lukechampine: can probably stick this outside the loop (and add `b.ResetTimer()` after). The speed of the `&` operation shouldn't be affected by whether the bits are set or unset. So perhaps you don't need to set each element to `base` either.

DavidVorick: It might be optimized out though if a known zero value is being set to zero again. I'd rather keep it this way.

lukechampine: I highly doubt that, but I suppose it's not impossible. By the same token, the compiler might optimize your bit-setting loop to a single `&`.
I have long been curious about how aggressively the compiler will optimize benchmarks. It seems like a Sufficiently Smart Compiler could easily see that `usageArray` isn't being used anywhere, and decide to remove all the code being benchmarked. I think this actually does happen for really trivial code. Nowadays I frequently use `_ = fn()` to indicate that the function call shouldn't be optimized out.

Anyway, you should at least move the allocation out of the loop. There's no reason to be allocating 10e6 bytes inside a tight benchmarking loop. It won't affect the timer, but it still makes the benchmark run more slowly.

modules/host/contractmanager/storagefoldershrink_bench_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+		// Set all of the bits to zero.
+		for j := 0; j < len(usageArray); j++ {
+			// Set each bit to zero.
+			for k := 0; k < storageFolderGranularity; k++ {
+				usageElement := usageArray[j]
+				usageElementUpdated := usageElement & (^(1 << uint64(k)))
+				usageArray[j] = usageElementUpdated
```
lukechampine: you can do `usageArray[j] &^= 1 << uint64(k)` here

DavidVorick:
uh oh

DavidVorick:
Fixed - was just a problem with the staging branch being really out of date

lukechampine:
As discussed, I would like to see the test files compressed and extracted at test time, or perhaps even generated at test time. If possible, their entropy should be low so that they compress well. Right now they are cluttering the diff (can't even load the table of contents)

modules/host/persist_compat_1.2.0_test.go
```diff
@@ -0,0 +1,122 @@
...
+	oldCapacity := smPersist.StorageFolders[0].Size + smPersist.StorageFolders[1].Size
+	oldCapacityRemaining := smPersist.StorageFolders[0].SizeRemaining + smPersist.StorageFolders[1].SizeRemaining
+	oldUsed := oldCapacity - oldCapacityRemaining
+
+	// Create the symlink to point to the storage folder.
+	err = os.Symlink(filepath.Join(legacyHost, "storageFolderOne"), filepath.Join(legacyHost, v112StorageManagerDir, "66"))
+	if err != nil {
+		t.Fatal(err)
+	}
+	err = os.Symlink(filepath.Join(legacyHost, "storageFolderTwo"), filepath.Join(legacyHost, v112StorageManagerDir, "04"))
```
lukechampine: I thought we were getting rid of symlinks because they don't work on windows?

sync/tryrwmutex.go
```diff
@@ -0,0 +1,103 @@
...
+	"sync"
+	"sync/atomic"
+)
+
+const (
+	readOffset  = uint64(1)
+	writeOffset = uint64(1 << 20)
+	tryOffset   = uint64(1 << 40)
+
+	tryMask = uint64(1099511627775) // equivalent to setting the first 39 bits to '1'.
```
lukechampine: hex is better for this purpose. This is actually FFFFFFFFFF, whereas the comment states that it is 7FFFFFFFFF. You could also use `(1<<39)-1`.
It seems like this code works fine as-is, though.

sync/tryrwmutex.go
```diff
@@ -0,0 +1,103 @@
...
+func (tm *TryRWMutex) TryRLock() bool {
+	// Signal that a read lock is waiting, and that there is contention from a
+	// Try call.
+	v := atomic.AddUint64(&tm.lock, readOffset+tryOffset)
+	// Mask the try offset when performing the comparison.
+	v = v & tryMask
+	if v > writeOffset {
+		// If there is a write lock waiting, revert the signal and return
+		// false.
+		atomic.AddUint64(&tm.lock, ^(readOffset + tryOffset - 1))
```
lukechampine: all this atomic code makes me a bit uneasy, but as long as it's properly tested and not burning CPU I can't complain

crypto/rand.go
```diff
@@ -3,32 +3,106 @@ package crypto
...
+		// Fetch and update the counter values before executing the hash.
+		var newCounter, newCounter2 uint64
+		newCounter = atomic.AddUint64(&counter, 1)
+		if newCounter == 0 {
+			newCounter2 = atomic.AddUint64(&counter2, 1)
+		} else {
+			newCounter2 = atomic.LoadUint64(&counter2)
+		}
+		// Grab some entropy using the unique counter set.
+		entropy := HashAll(newCounter, newCounter2, entropyBase)
```
DavidVorick: The mutexes will run more noticeably slower than atomic if you have a dozen threads all concurrently calling Read I think

sync/tryrwmutex.go
DavidVorick: Wow I don't even remember writing this code.

Also, seems like our Sync stuff should be pulled out into a separate package. Or maybe each sync tool pulled out separately.

I bet a lot of go devs would like these things.

We should also pull `Flush` from the TG, I think we've given up on using that. It's a nice concept but in terms of practical engineering it's just not worthwhile. Plus Flush is the whole reason we can't do nested Add calls, which otherwise I think would be supported.

modules/host/persist_compat_1.2.0_test.go
DavidVorick: This is only in the test code lol. You have to make symlinks to mimic the old storage manager which uses symlinks.

crypto/rand.go
lukechampine: ah, good point. I wasn't benchmarking a high-contention scenario

sync/tryrwmutex.go
lukechampine: Agreed. We may also want to pull the new rand code out into a separate package, and make it slightly more configurable (e.g. allow using any hash, not just blake2b, and allow creating your own `Reader` instance instead of needing to use the global one).

modules/host/contractmanager/storagefolder.go
```diff
@@ -0,0 +1,370 @@
...
+		panic(err)
+	}
+
+	// Find the first element of the array that is not completely full.
+	var i int
+	for i = start; i < len(usage); i++ {
+		if usage[i] != math.MaxUint64 {
+			break
+		}
+	}
```
DavidVorick: Randomness is only for performance. If we start at the beginning every time we will have to scan approximately all full sectors before finding an opening. But if we start in a random place, and gaps are approx. randomly distributed, we only have to scan (fill %) items. Unlike truly random choices, this has a firm upper bound on scan time.

crypto/rand.go
```diff
@@ -3,32 +3,106 @@ package crypto
...
+	return b, nil
 }
 
 // RandIntn returns a non-negative random integer in the range [0,n). It panics
 // if n <= 0.
 func RandIntn(n int) (int, error) {
-	r, err := rand.Int(rand.Reader, big.NewInt(int64(n)))
-	return int(r.Int64()), err
+	if n <= 0 {
+		panic("RandIntn must be called with a positive, nonzero number")
```
DavidVorick: you can go for it. I left it returning an error b/c I did not want this diff to be too huge. :>

sync/tryrwmutex.go
DavidVorick: I don't see why that would be necessary - random numbers are random numbers. Blake2b is just an implementation detail, and really if we were pulling it out into a separate package we'd use SHA2 b/c SHA2 has hardware acceleration support on a lot of machines.

Users shouldn't need to worry about implementation details.

modules/host/contractmanager/contractmanager_bench_test.go
```diff
@@ -0,0 +1,89 @@
...
+func BenchmarkStorageFolders(b *testing.B) {
+	// Create a massive usage array, matching a 96 TiB storage folder on disk.
+	// The array is a bit-array, so 24e6 sectors (96 TiB) is represented by
+	// 375e3 usage elements.
+	usage := make([]uint64, 375e3)
+
+	// Fill the folder to ~99.99% capacity, which will degrade performance.
+	for i := 0; i < 23999e3; i++ {
+		randFreeSector(usage)
+	}
```
lukechampine: looks like this isn't actually setting the bits?

DavidVorick: When it was written that function did set the bits. I guess it needs to update

DavidVorick:
No appveyor build?

But, should be good to go. I screwed up the commit history again, the final commit is accidentally amended into the merge-with-master commit. Hopefully forgivable as it's basically just spellchecking.

lukechampine:
To get appveyor to build I think you just need to add it to the appveyor.yml. We should definitely do that because otherwise we have little means to test windows compatibility

DavidVorick:
I can't figure out why travis is stuck? The Makefile won't complete?

lukechampine:
yeah I think that happens when you PR into a branch that isn't up to date with master. I updated the staging branch so it should work now.

modules/host/persist_compat_1.2.0_test.go
lukechampine: seems like this is causing the test to fail on windows. Not sure how we handled symlink testing on windows previously.

lukechampine:
the 300 files are still in the git history because they were added and deleted in separate commits. We should rebase this branch to remove the commit where they were added. I can handle this if you want, but if you have unpushed changes to this branch you should probably do it yourself

DavidVorick:
As long as the build passes I'm all set. I've screwed up the git here a couple of time, I'll let you handle pulling out the 300 files that we don't need.

PR #1504, opened 2016-11-23
Uptime monitoring

lukechampine:
The contractor now spawns a background loop that regularly deletes any contracts whose hosts have gone offline. Since the contractor already tries to form new contracts when it dips below `allowance.Hosts`, no further work is needed to ensure that the contractor will replace the offline hosts with new contracts. It is then the renter's job to reupload data to the new contracts.

This code relies on `hdb.IsOffline` being a good indicator of whether a host is truly offline. The current `IsOffline` algorithm is too simple and will result in contracts being deleted too aggressively. For example, if the hostdb is turned off for a week, and a host is momentarily offline when the hostdb turns on, then it will consider the host to be offline.

### DIFF
--- a/modules/renter/contractor/contractor.go
+++ b/modules/renter/contractor/contractor.go
@@ -85,14 +85,12 @@ func (c *Contractor) Contract(hostAddr modules.NetAddress) (modules.RenterContra
 	return modules.RenterContract{}, false
 }
 
-// Contracts returns the contracts formed by the contractor.
+// Contracts returns the contracts formed by the contractor. Only contracts
+// formed with currently online hosts are returned.
 func (c *Contractor) Contracts() (cs []modules.RenterContract) {
 	c.mu.RLock()
 	defer c.mu.RUnlock()
-	for _, c := range c.contracts {
-		cs = append(cs, c)
-	}
-	return
+	return c.onlineContracts()
 }
 
 // resolveID returns the ID of the most recent renewal of id.
@@ -167,9 +165,6 @@ func newContractor(cs consensusSet, w wallet, tp transactionPool, hdb hostDB, p 
 	if err != nil {
 		return nil, errors.New("contractor subscription failed: " + err.Error())
 	}
-
-	// spawn uptime loop
-	go c.threadedMonitorUptime()
 
 	return c, nil
 }
--- a/modules/renter/contractor/contractor.go
+++ b/modules/renter/contractor/contractor.go
@@ -168,5 +168,8 @@ func newContractor(cs consensusSet, w wallet, tp transactionPool, hdb hostDB, p 
 		return nil, errors.New("contractor subscription failed: " + err.Error())
 	}
 
+	// spawn uptime loop
+	go c.threadedMonitorUptime()
+
 	return c, nil
 }
--- a/modules/renter/contractor/renew.go
+++ b/modules/renter/contractor/renew.go
@@ -59,8 +59,11 @@ func (c *Contractor) managedRenew(contract modules.RenterContract, numSectors ui
 func (c *Contractor) managedRenewContracts() error {
 	c.mu.RLock()
 	// Renew contracts when they enter the renew window.
+	// NOTE: offline contracts are not considered here, since we may have
+	// replaced them (and we probably won't be able to connect to their host
+	// anyway)
 	var renewSet []types.FileContractID
-	for _, contract := range c.contracts {
+	for _, contract := range c.onlineContracts() {
 		if c.blockHeight+c.allowance.RenewWindow >= contract.EndHeight() {
 			renewSet = append(renewSet, contract.ID)
 		}
--- a/modules/renter/contractor/update.go
+++ b/modules/renter/contractor/update.go
@@ -44,7 +44,7 @@ func (c *Contractor) ProcessConsensusChange(cc modules.ConsensusChange) {
 
 	// only attempt contract formation/renewal if we are synced
 	// (harmless if not synced, since hosts will reject our renewal attempts,
-	// (but very slow)
+	// but very slow)
 	if cc.Synced {
 		go func() {
 			// only one goroutine should be editing contracts at a time
--- a/modules/renter/contractor/update.go
+++ b/modules/renter/contractor/update.go
@@ -53,16 +53,16 @@ func (c *Contractor) ProcessConsensusChange(cc modules.ConsensusChange) {
 			}
 			defer c.editLock.Unlock()
 
-			// renew any contracts that have entered the renew window
+			// renew any (online) contracts that have entered the renew window
 			err := c.managedRenewContracts()
 			if err != nil {
 				c.log.Debugln("WARN: failed to renew contracts after processing a consensus chage:", err)
 			}
 
-			// if we don't have enough contracts, form new ones
+			// if we don't have enough (online) contracts, form new ones
 			c.mu.RLock()
 			a := c.allowance
-			remaining := int(a.Hosts) - len(c.contracts)
+			remaining := int(a.Hosts) - len(c.onlineContracts())
 			c.mu.RUnlock()
 			if remaining <= 0 {
 				return
--- a/modules/renter/contractor/uptime.go
+++ b/modules/renter/contractor/uptime.go
@@ -41,13 +41,11 @@ func (c *Contractor) threadedMonitorUptime() {
 		// delete contracts with bad hosts. When a new block arrives,
 		// ProcessConsensusChange will take care of forming new contracts as
 		// needed.
-		c.editLock.Lock()
 		c.mu.Lock()
 		for _, contract := range badContracts {
 			delete(c.contracts, contract.ID)
 		}
 		c.mu.Unlock()
-		c.editLock.Unlock()
 		c.log.Printf("INFO: deleted %v contracts because their hosts were offline", len(badContracts))
 	}
 }
--- a/modules/renter/contractor/uptime.go
+++ b/modules/renter/contractor/uptime.go
@@ -1,0 +1,43 @@
+package contractor
+
+import (
+	"time"
+)
+
+const (
+	// uptimeInterval specifies how frequently hosts are checked.
+	uptimeInterval = 30 * time.Minute
+)
+
+// threadedMonitorUptime regularly checks host uptime, and deletes contracts
+// whose hosts fall below a minimum uptime threshold.
+func (c *Contractor) threadedMonitorUptime() {
+	for range time.Tick(uptimeInterval) {
+		// get current contracts
+		contracts := c.Contracts()
+
+		// identify hosts with poor uptime
+		badContracts := contracts[:0] // reuse memory during filter
+		for _, contract := range contracts {
+			if c.hdb.IsOffline(contract.NetAddress) {
+				badContracts = append(badContracts, contract)
+			}
+		}
+		if len(badContracts) == 0 {
+			// nothing to do
+			continue
+		}
+
+		// delete contracts with bad hosts. When a new block arrives,
+		// ProcessConsensusChange will take care of forming new contracts as
+		// needed.
+		c.editLock.Lock()
+		c.mu.Lock()
+		for _, contract := range badContracts {
+			delete(c.contracts, contract.ID)
+		}
+		c.mu.Unlock()
+		c.editLock.Unlock()
+		c.log.Println("INFO: deleted %v contracts because their hosts were offline", len(badContracts))
+	}
+}
--- a/modules/renter/contractor/uptime.go
+++ b/modules/renter/contractor/uptime.go
@@ -28,6 +28,14 @@ var uptimeWindow = func() time.Duration {
 // IsOffline indicates whether a contract's host should be considered offline,
 // based on its scan metrics.
 func (c *Contractor) IsOffline(addr modules.NetAddress) bool {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+	return c.isOffline(addr)
+}
+
+// isOffline indicates whether a contract's host should be considered offline,
+// based on its scan metrics.
+func (c *Contractor) isOffline(addr modules.NetAddress) bool {
 	// lookup the contract's host
 	host, ok := c.hdb.Host(addr)
 	if !ok {
@@ -69,12 +77,12 @@ func (c *Contractor) IsOffline(addr modules.NetAddress) bool {
 	return windowEnd.Sub(windowStart) >= uptimeWindow
 }
 
-// threadedMonitorUptime regularly checks host uptime, and deletes contracts
-// whose hosts fall below a minimum uptime threshold.
+// onlineContracts returns the subset of the Contractor's contracts whose
+// hosts are considered online.
 func (c *Contractor) onlineContracts() []modules.RenterContract {
 	var cs []modules.RenterContract
 	for _, contract := range c.contracts {
-		if !c.IsOffline(contract.NetAddress) {
+		if !c.isOffline(contract.NetAddress) {
 			cs = append(cs, contract)
 		}
 	}
--- a/modules/renter/contractor/uptime.go
+++ b/modules/renter/contractor/uptime.go
@@ -7,61 +7,54 @@ import (
 	"github.com/NebulousLabs/Sia/modules"
 )
 
-// uptimeInterval specifies how frequently hosts are checked.
-var (
-	uptimeInterval = func() time.Duration {
-		switch build.Release {
-		case "dev":
-			return 1 * time.Minute
-		case "standard":
-			return 30 * time.Minute
-		case "testing":
-			return 100 * time.Millisecond
-		}
-		panic("undefined uptimeInterval")
-	}()
+// uptimeMinScans is the minimum number of scans required to judge whether a
+// host is offline or not.
+const uptimeMinScans = 3
 
-	// uptimeWindow specifies the duration in which host uptime is checked.
-	uptimeWindow = func() time.Duration {
-		switch build.Release {
-		case "dev":
-			return 30 * time.Minute
-		case "standard":
-			return 7 * 24 * time.Hour // 1 week
-		case "testing":
-			return 15 * time.Second
-		}
-		panic("undefined uptimeWindow")
-	}()
-)
+// uptimeWindow specifies the duration in which host uptime is checked.
+var uptimeWindow = func() time.Duration {
+	switch build.Release {
+	case "dev":
+		return 30 * time.Minute
+	case "standard":
+		return 7 * 24 * time.Hour // 1 week
+	case "testing":
+		return 15 * time.Second
+	}
+	panic("undefined uptimeWindow")
+}()
 
 // isOffline decides whether a host should be considered offline, based on its
 // scan metrics.
 func isOffline(host modules.HostDBEntry) bool {
 	// consider a host offline if:
-	// 1) The host has been scanned at least 3 times in the past uptimeWindow, and
-	// 2) The scans occurred over a period of at least 1/3 the uptimeWindow, and
-	// 3) None of the scans succeeded
-	var pastWeek []modules.HostDBScan
-	for _, scan := range host.ScanHistory {
-		if time.Since(scan.Timestamp) < uptimeWindow {
-			if scan.Success {
-				// at least one scan in the uptimeWindow succeeded
-				return false
-			}
-			pastWeek = append(pastWeek, scan)
-		}
-	}
-	if len(pastWeek) < 3 {
+	// 1) The host has been scanned at least three times, and
+	// 2) The three most recent scans have all failed, and
+	// 3) The time between the most recent scan and the last successful scan
+	//    (or first scan) is at least uptimeWindow
+	numScans := len(host.ScanHistory)
+	if numScans < uptimeMinScans {
 		// not enough data to make a fair judgment
 		return false
 	}
-	if pastWeek[len(pastWeek)-1].Timestamp.Sub(pastWeek[0].Timestamp) < uptimeWindow/3 {
-		// scans were not sufficiently far apart to make a fair judgment
-		return false
+	// NOTE: ScanHistory is ordered from oldest-newest
+	recent := host.ScanHistory[numScans-uptimeMinScans:]
+	for _, scan := range recent {
+		if scan.Success {
+			// one of the scans succeeded
+			return false
+		}
 	}
-	// all conditions satisfied
-	return true
+	// initialize window bounds
+	windowStart, windowEnd := host.ScanHistory[0].Timestamp, host.ScanHistory[numScans-1].Timestamp
+	// iterate from newest-oldest, seeking to last successful scan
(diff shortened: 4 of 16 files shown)


modules/renter/contractor/uptime.go
```diff
@@ -0,0 +1,51 @@
...
+		return 30 * time.Minute
+	case "testing":
+		return 100 * time.Millisecond
+	}
+	panic("undefined uptimeInterval")
+}()
+
+// threadedMonitorUptime regularly checks host uptime, and deletes contracts
+// whose hosts fall below a minimum uptime threshold.
+func (c *Contractor) threadedMonitorUptime() {
```
DavidVorick: Is deleting the best course of action? I feel like it might be better just to mark the contract as 'unusable'. Then the signal is there for the renter that new contracts need to be created, but you are still able to talk to the host in the future if it comes back.

modules/renter/contractor/uptime.go
```diff
@@ -0,0 +1,51 @@
...
+	for range time.Tick(uptimeInterval) {
+		// get current contracts
+		contracts := c.Contracts()
+
+		// identify hosts with poor uptime
+		badContracts := contracts[:0] // reuse memory during filter
```
DavidVorick: this is on the boarder of too clever. I see that there are no issues but I had to look more closely to be certain. I think falls outside of the 'obviously correct' standard that we typically aim for.

modules/renter/hostdb/hostentry.go
```diff
@@ -113,19 +115,43 @@ func (hdb *HostDB) AverageContractPrice() types.Currency {
...
+		lastSeen, lastScanned = entry.LastSeen, entry.LastScanned
+	}
+	hdb.mu.RUnlock()
+	if !ok {
 		return false
 	}
-	if h, ok := hdb.allHosts[addr]; ok {
-		return !h.Online
+
+	if time.Since(lastScanned) > uptimeThreshold {
```
DavidVorick: Scanning could take a few minutes. Do we want the function to block for that time?

I think better in this case to 

```
if time.Since(lastSeen) > time.Since(lastScanned).Add(time.Hour * 24 * 3) {
    return false
}
```
Which will return true only if the last scan was at least 3 days following the last sighting.

DavidVorick: actually, might that be sufficient for the whole function?

lukechampine: blocking isn't a big deal since the contractor is calling it in a background thread. And it's probably good because it means contract hosts will be scanned more frequently than other hosts in the hostdb.

regardless, I don't think two data points are really enough here. We want something more like "was successfully scanned 8/10 times over the last 3 days." Otherwise the false positive rate is just too high.

DavidVorick: By implement it this way you are requiring the contractor to call it in a background thread, and I don't think our implementation of the contractor is stable enough to justify limitations in upstream.

IsOffline is a query function, I don't think I would expect a function like that to block for network calls. I would generally expect it to just look at the existing state.

modules/renter/contractor/uptime.go
```diff
@@ -0,0 +1,94 @@
...
+			return 100 * time.Millisecond
+		}
+		panic("undefined uptimeInterval")
+	}()
+
+	// uptimeWindow specifies the duration in which host uptime is checked.
+	uptimeWindow = func() time.Duration {
+		switch build.Release {
+		case "dev":
+			return 24 * time.Hour // 1 day
```
DavidVorick: 1 day is too long, 'dev' constants are what you use in the antfarm. We don't want to wait a day to realize that a host has gone offline if we are testing the handling of offline hosts.

Probably 30 minutes is better.

modules/renter/contractor/uptime.go
```diff
@@ -0,0 +1,94 @@
...
+		case "standard":
+			return 7 * 24 * time.Hour // 1 week
+		case "testing":
+			return 1 * time.Minute // 1 minute
```
DavidVorick: same here, a minute is a really long time when running automated tests. Perhaps 15 seconds?

modules/renter/contractor/uptime.go
```diff
@@ -0,0 +1,94 @@
...
+		}
+	}
+	// need at least 3 scans to make a fair judgment
+	if len(lastWeek) < 3 {
+		return false
+	}
+	// return true if all 3 most recent scans failed
+	// NOTE: lastWeek is ordered from oldest to newest
+	lastWeek = lastWeek[len(lastWeek)-3:]
+	return !lastWeek[0].Success && !lastWeek[1].Success && !lastWeek[2].Success
```
DavidVorick: we discussed offline, but wanted to comment here so it's visible.

Consider a host offline if:

1) There have been no instances of uptime in the entire past week.
2) There are a minimum of 3 data points total.

DavidVorick: *

2. There are a minimum of 3 data points total AND the 3 most recent data points all indicate that the host is offline.

modules/renter/contractor/contractor.go
```diff
@@ -168,5 +168,8 @@ func newContractor(cs consensusSet, w wallet, tp transactionPool, hdb hostDB, p
 		return nil, errors.New("contractor subscription failed: " + err.Error())
 	}
 
+	// spawn uptime loop
+	go c.threadedMonitorUptime()
```
DavidVorick: Oh, rather than do it in the update function, since we aren't actually deleting contracts, we might just want to change the logic that the renter uses when it's counting the number of contracts that it currently has.

modules/renter/contractor/uptime.go
```diff
@@ -0,0 +1,94 @@
...
+		// ProcessConsensusChange will take care of forming new contracts as
+		// needed.
+		c.mu.Lock()
+		for _, contract := range badContracts {
+			delete(c.contracts, contract.ID)
+		}
+		c.mu.Unlock()
+		c.log.Printf("INFO: deleted %v contracts because their hosts were offline", len(badContracts))
+	}
+}
```
DavidVorick: Because we decided not to delete contracts, it's probably better to move this logic to the contract formation loop, where it counts the number of contracts it has by  using the above logic instead of by calling len(c.contracts)

modules/renter/repair.go
```diff
@@ -30,6 +30,19 @@ var renewThreshold = func() types.BlockHeight {
...
+// uptimeWindow specifies the duration in which host uptime is checked.
+var uptimeWindow = func() time.Duration {
+	switch build.Release {
+	case "dev":
+		return 24 * time.Hour // 1 day
+	case "standard":
+		return 7 * 24 * time.Hour // 1 week
+	case "testing":
+		return 1 * time.Minute // 1 minute
+	}
```
DavidVorick: same changes apply to these constants as apply to the contractor constants - day is too long for "dev", etc.

modules/renter/repair.go
```diff
@@ -185,6 +198,28 @@ func (f *file) expiringContracts(height types.BlockHeight) []fileContract {
...
+	// need at least 3 scans to make a fair judgment
+	if len(lastWeek) < 3 {
+		return false
+	}
+	// return true if all 3 most recent scans failed
+	// NOTE: lastWeek is ordered from oldest to newest
+	lastWeek = lastWeek[len(lastWeek)-3:]
+	return !lastWeek[0].Success && !lastWeek[1].Success && !lastWeek[2].Success
+}
```
DavidVorick: same high-level changes here as apply to the contractor

lukechampine:
Summary of new changes:
- Uptime constants have been decreased.
- Uptime algorithm is now:
>consider a host offline if:
>  1) The host has been scanned at least 3 times in the past week, and
>  2) The scans occurred over a period of at least 1/3rd of a week, and
>  3) None of the scans succeeded.
- Offline checking now happens during `ProcessConsensusChange`, i.e. when the contractor is trying to renew or form contracts.
- Offline contracts are no longer counted during renewal or repair. That is, the contractor will not try to renew contracts marked offline, and it will ignore offline contracts when calculating how many it needs to form to satisfy the allowance. This means that as soon as a contract goes offline, the contractor will form a new one in its place. If necessary, we can modify this so that the contractor only forms new contracts after a certain number go offline.
- The allowance code has not yet been modified to "leave room" for contract replacement.

DavidVorick:
It's okay to make contracts one by one as hosts go offline. Data should only be downloaded + reuploaded when a certain number of pieces are missing (to conserve bandwidth), but the contract stuff doesn't have the same limitation, so it's okay to make new contracts one at a time.

modules/renter/contractor/uptime_test.go
```diff
@@ -0,0 +1,131 @@
...
+	"time"
+
+	"github.com/NebulousLabs/Sia/build"
+	"github.com/NebulousLabs/Sia/modules"
+	"github.com/NebulousLabs/Sia/types"
+)
+
+// uptimeHostDB overrides an existing hostDB so that it always returns
+// isOffline == true for a specified address.
+type uptimeHostDB struct {
```
DavidVorick: should have a different name - 'uptimeHostDB' doesn't give me the impression that it always returns isOffline == true.

modules/renter/contractor/uptime_test.go
```diff
@@ -0,0 +1,131 @@
...
+	hostDB
+	addr modules.NetAddress
+}
+
+func (u uptimeHostDB) Host(addr modules.NetAddress) (modules.HostDBEntry, bool) {
```
DavidVorick: This should have a docstring. Repeating yourself in comments I feel is okay because it makes it easier to navigate - otherwise the reader has to hunt around more to figure out what a function does.

In this case it's nontrivial relative to the name. It doesn't just return a Host, it returns an offline host. That's definitely significant enough to merit a docstring.

modules/renter/contractor/uptime_test.go
```diff
@@ -0,0 +1,131 @@
...
+		// no data
+		{nil, false},
+		// not enough data
+		{[]modules.HostDBScan{oldBadScan, newGoodScan}, false},
+		// data covers small range
+		{[]modules.HostDBScan{oldBadScan, oldBadScan, oldBadScan}, false},
+		// data covers large range, but at least 1 scan succeded
+		{[]modules.HostDBScan{oldBadScan, newGoodScan, currentBadScan}, false},
+		// data covers large range, no scans succeded
+		{[]modules.HostDBScan{oldBadScan, newBadScan, currentBadScan}, true},
```
DavidVorick: and also bad, bad, bad, good, where good is not within uptimeWindow

modules/renter/repair.go
```diff
@@ -185,6 +202,39 @@ func (f *file) expiringContracts(height types.BlockHeight) []fileContract {
...
+	windowStart, windowEnd := host.ScanHistory[0].Timestamp, host.ScanHistory[numScans-1].Timestamp
+	// iterate from newest-oldest, seeking to last successful scan
+	for i := numScans - 1; i >= 0; i-- {
+		if scan := host.ScanHistory[i]; scan.Success {
+			windowStart = scan.Timestamp
+			break
+		}
+	}
+	return windowEnd.Sub(windowStart) >= uptimeWindow
+}
```
DavidVorick: hmm.

Given how this is used, I'm going to  say that this should be established as an exported method of the contract. "Is this contract still useful at all?" seems like a question that the contractor should be answering, given that it's the one collecting the metrics on contracts and whatnot.

The renter will have separate sorts of 'should I use this contract right now' functions, (e.g. 'no' if it has failed to perform an upload in the past 2 hours or if it has been unbearably slow).

With regards to my previous position on duplicating this code, I had assumed we were talking about duplicating it between the hostdb and contractor, where I feel it is being used in sufficiently different ways to justify separate implementations.

modules/renter/repair.go
```diff
@@ -195,10 +245,12 @@ func (f *file) offlineChunks(hdb hostDB) map[uint64][]uint64 {
...
 	offline := make(map[uint64][]uint64)
 	for _, fc := range f.contracts {
-		if hdb.IsOffline(fc.IP) {
-			for _, p := range fc.Pieces {
-				offline[p.Chunk] = append(offline[p.Chunk], p.Piece)
-			}
+		host, ok := hdb.Host(fc.IP)
+		if !ok || !isOffline(host) {
+			continue
+		}
```
DavidVorick: I'm not sure how big of an implementation change it would be, but it should be asking the contractor here instead of the hostdb.

modules/renter/repair_test.go
```diff
@@ -154,29 +154,29 @@ func TestRepair(t *testing.T) {
 	*/
 }
 
-// offlineHostDB is a mocked hostDB, used for testing the offlineChunks method
-// of the file type. It is implemented as a map from NetAddresses to booleans,
-// where the bool indicates whether the host is active.
+// offlineHostDB overrides an existing hostDB so that it always returns
+// isOffline == false for a specified address.
 type offlineHostDB struct {
```
DavidVorick: lol it has a more appropriate name here

modules/renter/contractor/uptime.go
```diff
@@ -0,0 +1,76 @@
...
+	// 1) The host has been scanned at least three times, and
+	// 2) The three most recent scans have all failed, and
+	// 3) The time between the most recent scan and the last successful scan
+	//    (or first scan) is at least uptimeWindow
+	numScans := len(host.ScanHistory)
+	if numScans < uptimeMinScans {
+		// not enough data to make a fair judgment
+		return false
+	}
+	// NOTE: ScanHistory is ordered from oldest-newest
```
lukechampine: `time.Time` is a Unix timestamp so it isn't affected by timezones. AFAICT, the worst case scenario is that the user sets their clock ahead by a significant amount (i.e. a large fraction of the uptimeWindow) and then performs an unsuccessful scan. This could cause the host to be marked offline if the jump is sufficiently large. (The host would still need to have been offline for its last 3 scans, though.) That's kinda-sorta a problem, but there isn't really anything we can do about it.

DavidVorick: I would feel a lot better if we were validating that the timestamps were monotonically increasing, it's not that much extra code and it protects against potential funky behavior. I don't think we should rely on the user not adjusting their hardware clocks, and we definitely shouldn't rely on undefined behavior being acceptably incorrect.

lukechampine: ok, consider this case:

- user scans host three times within an hour. All three scans fail, but the host is not marked offline because the scans did not happen over a sufficient window.
- user sets their clock back 1 week
- user scans again, this time successfully
- scan is inserted in sorted order, so it appears 1 week before the first three scans
- now we have three recent failed scans, and the last successful scan was over a week ago. The host is marked offline.

So now a host can be marked offline even if the user _just scanned it successfully!_ Clearly this behavior is not correct either.

It's misleading to say that this code is "relying" on the user not adjusting their clock, as though most of our code did not make such assumptions. We implicitly rely on the user not to do hundreds of things. It's not our responsibility to anticipate and guard against every way the user might modify their system. We can only focus on the common cases. For example, it's not uncommon for users to hard-power-off their systems, so we guard against that by using atomic files, transactional databases, etc. But it is very uncommon for a user to set their `ulimit` to a number small enough to prevent siad from opening its persist files. When that happens, all we can say is "well, it's not reasonable to expect siad to work if you do that." Likewise, it's not reasonable to expect a distributed consensus application to function correctly if your clock is set a week in the past.

modules/renter/repair.go
lukechampine: A method of the contractor, you mean? Should it still be called `IsOffline`? In its current incarnation, `isOffline` only needs the `modules.HostDBEntry`, so it seems strange to make it part of the contractor. I can see your reasoning though if we expand it to a more general ruling about whether the contract is useful at all.

modules/renter/contractor/uptime.go
DavidVorick: your case is valid.

But it doesn't change the fact that you are assuming that data has been sorted when it hasn't actually been sorted. To me that says it could cause weird things to happen, which makes me worry about exploits. Undefined behavior is surface area for problems + attacks.

Yes, you are right that we mostly assume that the user is being sane. But clocks adjust all the time, whether via NTP or manual intervention. My bigger problem is that the code is making decisions based on assumptions that may be invalid.

And I don't mean assumptions like 'there's at least X free space on disk', I mean assumptions that could cause the incorrect code path to be followed, resulting in a cascade of inconsistent state.

lukechampine: They adjust on the order of a few seconds, not days. It's only a problem if the discrepancy is enough to reorder the scans, which would mean two scans happened within seconds of each other.

Anyway, I thought of another potential issue: when determining the scan window, we call `windowEnd.Sub(windowStart)`. If the scans are not in order, it's possible for the subtraction to overflow.
In light of this, I am willing to sort the scans, despite my belief that such problems are highly unlikely to occur, and would be benign even if they did. I will also add some documentation and a debug check.

DavidVorick: "I thought of another potential issue: when determining the scan window, we call windowEnd.Sub(windowStart)"

This is exactly the type of thing that makes me really paranoid about that sort of assumption. It's possible that there are also other potential problems, or that there will be other potential problems in the future after we make additional modifications. It opens up a vector to cause inconsistent state, and that inconsistent state could potentially lead to other inconsistent state, which at some point could lead to a more serious bug.

I agree that we can't protect against users intentionally doing bad things to their program, but we should at least make it so that the state will not become inconsistent - instead it should crash or return an error (or be designed so that it can't become inconsistent).

modules/renter/contractor/uptime.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+	return windowEnd.Sub(windowStart) >= uptimeWindow
+}
+
+// threadedMonitorUptime regularly checks host uptime, and deletes contracts
+// whose hosts fall below a minimum uptime threshold.
+func (c *Contractor) onlineContracts() []modules.RenterContract {
+	var cs []modules.RenterContract
+	for _, contract := range c.contracts {
+		if !c.IsOffline(contract.NetAddress) {
```
DavidVorick: This is I believe technically fine, but it's a violation of convention.

Exported functions are assumed to manage their own locks, and unexported functions are assumed to be under an exclusive lock.

IsOffline as best I can tell is threadsafe without a lock, but someone changing its implementation in the future may run into unexpected problems if they expect that it's only ever called from functions that don't have locks.

modules/renter/contractor/contractor.go
```diff
@@ -85,14 +85,12 @@ func (c *Contractor) Contract(hostAddr modules.NetAddress) (modules.RenterContra
...
+// Contracts returns the contracts formed by the contractor. Only contracts
+// formed with currently online hosts are returned.
 func (c *Contractor) Contracts() (cs []modules.RenterContract) {
 	c.mu.RLock()
 	defer c.mu.RUnlock()
-	for _, c := range c.contracts {
-		cs = append(cs, c)
-	}
-	return
+	return c.onlineContracts()
```
DavidVorick: by convention lowercase functions assume that they are being called with an exclusive lock, but here you've only got a readlock.

DavidVorick: it's technically correct here but strict conventions make it easier to modify in the future

modules/renter/contractor/renew.go
```diff
@@ -59,8 +59,11 @@ func (c *Contractor) managedRenew(contract modules.RenterContract, numSectors ui
 func (c *Contractor) managedRenewContracts() error {
 	c.mu.RLock()
 	// Renew contracts when they enter the renew window.
+	// NOTE: offline contracts are not considered here, since we may have
+	// replaced them (and we probably won't be able to connect to their host
+	// anyway)
 	var renewSet []types.FileContractID
-	for _, contract := range c.contracts {
+	for _, contract := range c.onlineContracts() {
```
DavidVorick: ditto here - onlineContracts being called with a readlock, should be getting called under an exclusive lock.

modules/renter/contractor/contractor.go
lukechampine: >lowercase functions assume that they are being called with an exclusive lock

what? I don't recall any such convention. It's certainly not in our `Developers.md`. You wrap unexported functions with readlocks in both the (old) host and consensus set. When are you supposed to use readlocks if you can't call unexported functions with them?

modules/renter/contractor/update.go
```diff
@@ -53,16 +53,16 @@ func (c *Contractor) ProcessConsensusChange(cc modules.ConsensusChange) {
...
 			if err != nil {
 				c.log.Debugln("WARN: failed to renew contracts after processing a consensus chage:", err)
 			}
 
-			// if we don't have enough contracts, form new ones
+			// if we don't have enough (online) contracts, form new ones
 			c.mu.RLock()
 			a := c.allowance
-			remaining := int(a.Hosts) - len(c.contracts)
+			remaining := int(a.Hosts) - len(c.onlineContracts())
```
lukechampine: I've never heard of this rule. Our only official convention regarding unexported function locking is:

>Non-exported functions should not do any locking, unless they have a special prefix to the name (explained below). The responsibility for thread-safety comes from the exported functions which call the non-exported functions.

Nothing in there about it needing to be a write lock.

DavidVorick: oh hmm.

I guess that's sufficient if you trust the race detector. You can't trigger a deadlock this way, only a race condition.

At some point in my head I switched it from not doing locking to requiring a write lock.

DavidVorick:
Ok, I'm going to backpedal on convention violations regarding the unexported function, particularly since violation can at worst cause a race condition, and between the testing + antfarm + race detector we seem to be pretty good finding + repairing race conditions.

But calling `IsOffline()` from `onlineContracts()` is a more serious violation, because it risks a deadlock and those are substantially harder to debug. We have in the past lost multiple days of development just tracking down a single deadlock.

lukechampine:
agreed re: `IsOffline`, that is definitely a violation. As for the write-lock, I will clarify that I'm open to adopting the convention; I can see the benefit, but I want to give it some more thought first. I was just surprised because I wasn't previously aware of it.

DavidVorick:
There's another type of race condition that has bitten us a few times, and it's the type that is not caught usually by the race detector.

r.lock()
obj := r.someObj()
r.unlock()

// update obj

r.lock()
r.updateObj(obj)
r.unlock()

The problem above is that the field is being updated under lock, but the object has persisted outside of the lock. Some other updateObj call may have happened, and now you've changed your object and introduced a race condition, but fundamental data was never at risk so the race detector doesn't realize that two threads have made things inconsistent.

I don't think we have conventions that protect against that. I think that rust has a lot of good protections against stuff like this, but we aren't using rust.

Anyway, something to think about for the future, PR looks good.

lukechampine:
Yeah, I guess this is what drives people towards more fine-grained locking. In my experience, wherever I've seen the pattern you describe, it's because we're doing some blocking operation between the two locks. So we introduce an (undetectable) race condition in order to avoid holding a lock for a while. Perhaps our conventions should be updated to encourage safety first, and speed later.

PR #1507, opened 2016-11-30
Renter pools

DavidVorick:
None of the tests are passing, this PR is being opened mostly for a high-level pass on what's happening.

Basically, there's a worker pool now in the renter. Each worker has a contract and can perform uploads and downloads with the host on the other side of the contract.

The download queue will download 1 file at a time, with a moderate amount of parallelism. There is not really much randomization happening, which might mean it's somewhat inferior to the previous download scheme. However, I believe that the previous download scheme was entirely serial, whereas this download scheme will download an entire chunk in parallel, and it'll also download multiple chunks at the same time.

I would like to expand it so that it can also download multiple files at a time, but the code needs more work before getting there.

The uploading has much stronger parallelization now. The previous scheme was bottlenecked by the slowest host, basically meaning that uploads could move at-best as fast as 24x the slowest host, though I think in practice long timeouts meant that things usually moved very slowly, especially when trying to get to the first redundancy.

Things are still technically bottlenecked by the slowest host, however faster hosts will move onto the next chunk instead of waiting for the slower hosts. The time to reach 1x or 2x redundancy should very dramatically reduced.

What happens right now is a giant matrix of work is created when doing an upload. Every chunk that might need attention is looked at, and there's a grid that sort of remembers how much redundancy each chunk has already. 4 workers are set upon each chunk, meaning the chunk gains 1x redundancy at a time. When parallelizing, the algorithm waits until there are enough workers available and then it scans through and finds a chunk that is missing pieces from at least 4 of the available workers. Then it performs those uploads.

There's still a ton of room for optimization, and the giant upload matrices don't scale very well. But things should be moving a lot faster now vs. earlier. Once the bugs are worked out, at least.


modules/renter/renter.go
```diff
@@ -123,36 +122,110 @@ func newRenter(cs modules.ConsensusSet, tpool modules.TransactionPool, hdb hostD
...
+		if !found {
+			err = r.addWorker(nc.ID)
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	// Find all elements in oldContracts that are not in newContracts and kill
+	// their workers.
```
lukechampine: I think this can be cleaned up by using maps:
```go
newContracts := make(map[types.FilecontractID]struct{})
for _, id := range r.hostContractor.Contracts() {
    newContracts[id] = struct{}{}
}
oldContracts := r.workerPool // just need ID

// add a worker for each newContract that is not in oldContracts
for id := range newContracts {
    if _, ok := oldContracts[id]; !ok {
        r.addWorker(id)
    }
}
// remove any workers that are in oldContracts but not newContracts
for id := range oldContracts {
    if _, ok := newContracts[id]; !ok {
        r.retireWorker(id)
    }
}
```

modules/renter/worker.go
```diff
@@ -0,0 +1,245 @@
...
+		}
+	}
+	contract.Pieces = append(contract.Pieces, pieceData{
+		Chunk:      uw.chunkIndex,
+		Piece:      uw.pieceIndex,
+		MerkleRoot: root,
+	})
+	uw.file.contracts[w.contractID] = contract
+	uw.file.mu.Unlock()
+	w.renter.saveFile(uw.file)
```
lukechampine: wondering if this code could be moved out of the worker. It would be nice if the worker didn't need to worry about updating file metadata.

modules/renter/download.go
```diff
@@ -1,182 +1,290 @@
...
-	// make piece map
-	pieceMap := make(map[uint64][]pieceData)
-	for _, p := range pieces {
-		pieceMap[p.Chunk] = append(pieceMap[p.Chunk], p)
+// managedDownloadIteraiton downloads a chunk from the next available file.
+func (r *Renter) managedDownloadIteration() {
+	// Get the set of available workers.
+	availableWorkers := make([]*worker, 0)
+	id := r.mu.RLock()
+	for _, worker := range r.workerPool {
```
lukechampine: `availableWorkers := make([]*worker, 0, len(r.workerPool)`

modules/renter/download.go
```diff
@@ -1,182 +1,290 @@
...
+		// Signal that the download has errored and quit.
+		nextDownload.downloadFinished <- errInsufficientHosts
+		return
+	}
+
+	// TODO: Exclude the most expensive workers, and perhaps the slowest as
+	// well.
+
+	// Open a file handle for the download.
+	fileDest, err := os.OpenFile(nextDownload.destination, os.O_CREATE|os.O_RDWR|os.O_TRUNC, 0600)
```
lukechampine: Need to use `file.perm` for the permission bits, with 0666 as a sane default

modules/renter/download.go
```diff
@@ -1,182 +1,290 @@
...
+	if err != nil {
+		nextDownload.downloadFinished <- err
+		return
+	}
+
+	// Grab each chunk from the download.
+	chunkDownloads := make([]chunkDownload, nextDownload.fileSize/nextDownload.chunkSize)
```
lukechampine: there should be a `file.numChunks()` method for calculating this. Simple division will produce the wrong value unless the file size is an even multiple of the chunk size.

modules/renter/download.go
```diff
@@ -1,182 +1,290 @@
...
+			// working on this chunk.
+			for i := range availableWorkers {
+				_, exists := chunkDownloads[chunkIndex].workerAttempts[availableWorkers[i].contractID]
+				if !exists {
+					dw := downloadWork{
+						dataRoot:   pieceSet[chunkIndex][availableWorkers[i].contractID].MerkleRoot,
+						chunkIndex: chunkIndex,
+						resultChan: resultChan,
+					}
+					chunkDownloads[chunkIndex].workerAttempts[availableWorkers[i].contractID] = pieceSet[chunkIndex][availableWorkers[i].contractID].Piece
```
lukechampine: what was that JPL rule about no more than two indirections per line...? :P

modules/renter/download.go
```diff
@@ -208,94 +310,14 @@ func (r *Renter) Download(path, destination string) error {
 	}
```
lukechampine: some dead code here

modules/renter/download.go
lukechampine: Looks like `fileDest` isn't ever closed

modules/renter/download.go
```diff
@@ -1,182 +1,290 @@
...
 
-// A download is a file download that has been queued by the renter.
-type download struct {
-	// NOTE: received is the first field to ensure 64-bit alignment, which is
-	// required for atomic operations.
-	received   uint64
-	chunkIndex uint64
+	var activeDownloads int
+	var chunkIndex uint64
+	incompleteChunks := make([]uint64, 0)
```
lukechampine: use `var x []T` instead of `x := make([]T, 0)`

modules/renter/download.go
```diff
@@ -1,182 +1,290 @@
...
+type chunkDownload struct {
+	// Map of workers who have attempted downloads for this chunk, from the
+	// worker id to the index of the piece they are trying to download.
+	workerAttempts map[types.FileContractID]uint64
 
-	// fetch returns the data specified by piece metadata.
-	fetch(pieceData) ([]byte, error)
+	// List of completed pieces, mapping from the index of the piece to the
+	// data.
+	completedPieces map[uint64][]byte
```
lukechampine: This should be a slice, since the number of pieces is small. Pieces are also ordered, and the ordering is required for proper erasure-code recovery.

modules/renter/download.go
```diff
@@ -208,94 +310,14 @@ func (r *Renter) Download(path, destination string) error {
...
-			progressDeadline = time.Now().Add(30 * time.Minute)
-		}
-		// if we haven't made any progress in 30 minutes, give up
-		if time.Now().After(progressDeadline) {
-			return errors.New("no progress in 30 minutes; giving up")
-		}
-	}
-	return nil
+	// Block until the download has completed.
+	return <-d.downloadFinished
```
lukechampine: better to just return `d.downloadFinished`; that makes it dead simple for the API to add a non-blocking version of the `/download` route

modules/renter/repair.go
```diff
@@ -26,323 +25,374 @@ var renewThreshold = func() types.BlockHeight {
 	case "dev":
 		return 200
 	default:
-		return 144 * 7 * 3 // 3 weeks - to soon be 6 weeks.
+		return 144 * 7 * 6
```
lukechampine: `// 6 weeks`

modules/renter/repair.go
```diff
@@ -26,323 +25,374 @@ var renewThreshold = func() types.BlockHeight {
...
 	}
-	// filter out chunks missing less than half of their redundancy
-	filtered := make(map[uint64][]uint64)
-	for chunk, pieces := range offline {
-		if len(pieces) > f.erasureCode.NumPieces()/2 {
-			filtered[chunk] = pieces
+	if maxGaps == 0 {
+		// There is no work to do. Sleep for 15 minutes, or until there has
+		// been a new upload. Then iterate to make a new repair matrix and
+		// check again.
```
lukechampine: not sure I understand this. It looks like it will just exit the `if` block without constructing a new matrix.

modules/renter/repair.go
```diff
@@ -26,323 +25,374 @@ var renewThreshold = func() types.BlockHeight {
...
-func (f *file) repair(chunkIndex uint64, missingPieces []uint64, r io.ReaderAt, hosts []contractor.Editor) error {
-	// read chunk data and encode
-	chunk := make([]byte, f.chunkSize())
-	_, err := r.ReadAt(chunk, int64(chunkIndex*f.chunkSize()))
-	if err != nil && err != io.EOF && err != io.ErrUnexpectedEOF {
-		return err
+	// If the file is not being tracked, don't add it to the repair matrix.
+	_, exists := r.tracking[file.name]
+	if exists {
+		return
```
lukechampine: shouldn't this be in `createRepairMatrix`? Then `addFileToRepairMatrix` doesn't need to be a method of the Renter.

modules/renter/repair.go
```diff
@@ -26,323 +25,374 @@ var renewThreshold = func() types.BlockHeight {
...
-		return err
+
+	// For each chunk, create a map from the chunk to the pieces that it
+	// has, and to the contracts that have that chunk.
+	chunkCount := file.numChunks()
+	contractMap := make(map[uint64][]types.FileContractID)
+	pieceMap := make(map[uint64][]uint64)
+	for i := uint64(0); i < chunkCount; i++ {
+		contractMap[i] = make([]types.FileContractID, 0)
+		pieceMap[i] = make([]uint64, 0)
```
lukechampine: this loop isn't necessary. It doesn't initialize anything.

modules/renter/renter.go
DavidVorick: ok, I can do this.

modules/renter/repair.go
```diff
@@ -26,323 +25,374 @@ var renewThreshold = func() types.BlockHeight {
...
-				return
+	// Iterate through each chunk and, if there are gaps, add the inverse
+	// to the repair matrix.
+	for i := uint64(0); i < chunkCount; i++ {
+		if len(pieceMap[i]) < file.erasureCode.NumPieces() {
+			// Find the gaps in the pieces and contracts.
+			potentialPieceGaps := make(map[uint64]struct{})
+			for j := 0; j < file.erasureCode.NumPieces(); i++ {
+				potentialPieceGaps[i] = struct{}{}
+			}
```
lukechampine: you got your `i` and `j` mixed up.

`make([]bool, file.erasureCode.NumPieces())` might be a better data structure for this purpose.

modules/renter/renter.go
DavidVorick: I think I'm going to split this code into a function called 'updateWorkerPool', because you'll need to run this code any time that the contracts in the contractor changes, which can happen without the settings being set directly.

DavidVorick: (e.g. renewing, e.g. replacing offline hosts)

modules/renter/renter.go
```diff
@@ -123,36 +122,110 @@ func newRenter(cs modules.ConsensusSet, tpool modules.TransactionPool, hdb hostD
...
+	// Spin up the workers for the work pool.
+	contracts := hc.Contracts()
+	for _, c := range contracts {
+		err := r.addWorker(c.ID)
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	go r.threadedRepairLoop()
```
lukechampine: seems like `threadedDownloadLoop` isn't being spawned anywhere. Weren't we planning to merge it with the repair loop anyway? I thought that was the point of having workers, so that they could be shared between uploads/downloads.

modules/renter/repair.go
```diff
@@ -26,323 +25,374 @@ var renewThreshold = func() types.BlockHeight {
...
 
-// incompleteChunks returns a map of chunks containing pieces that have not
-// been uploaded.
-func (f *file) incompleteChunks() map[uint64][]uint64 {
-	f.mu.RLock()
-	defer f.mu.RUnlock()
+			// Record the number of gaps that this chunk has, which makes
+			// blocking-related decisions easier.
+			gapCounts[gaps.numGaps] = gapCounts[gaps.numGaps] + 1
```
lukechampine: just use `++`

modules/renter/repair.go
```diff
@@ -26,323 +25,374 @@ var renewThreshold = func() types.BlockHeight {
...
-	if !ok {
-		logAndRemove("removing %v from repair set: no longer tracking that file", name)
-		return
-	}
+			// Parse the filename and chunk index from the repair
+			// matrix key.
+			chunkIndexBytes := chunkID[:8]
+			filename := chunkID[8:]
+			chunkIndex := binary.LittleEndian.Uint64([]byte(chunkIndexBytes))
+			file, exists := r.files[filename]
```
lukechampine: this is a little bizarre

modules/renter/repair.go
```diff
@@ -26,323 +25,374 @@ var renewThreshold = func() types.BlockHeight {
...
+			chunkName := string(chunkPrefix) + file.name
 
-	incomplete := make(map[uint64][]uint64)
-	for chunkIndex, pieceBools := range present {
-		for pieceIndex, ok := range pieceBools {
-			if !ok {
-				incomplete[uint64(chunkIndex)] = append(incomplete[uint64(chunkIndex)], uint64(pieceIndex))
-			}
+			// Add the chunk to the repair matrix.
+			repairMatrix[chunkName] = &gaps
```
lukechampine: why use a pointer? `chunkGaps` is just a wrapper around two pointers (and an int, I guess)

modules/renter/renter.go
DavidVorick: I think I just forgot to spawn it.

The download loop is separate from the repair loop because the events that wake the download loop are different from the events that wake the repair loop.

The workers can handle both downloads and uploads being fed to them at the same time, so it's safe to have both loops running in parallel. That simplifies needing to worry about whether or not a worker is in use. The worker will block on the upload channel if there is an active download, and will block on the download channel if there is an active upload. The worker, once ready for more work, will prioritize downloads over uploads.

modules/renter/repair.go
```diff
@@ -26,323 +25,374 @@ var renewThreshold = func() types.BlockHeight {
...
-			expiring = append(expiring, fc)
+func (r *Renter) managedRepairIteration() {
+	// Create the initial set of workers that are used to perform
+	// uploading.
+	availableWorkers := make(map[types.FileContractID]struct{})
+	id := r.mu.RLock()
+	for id, worker := range r.workerPool {
+		// Ignore workers that have had an upload failure in the past two
+		// hours.
+		if worker.recentUploadFailure.Add(time.Minute * 120).Before(time.Now()) {
```
lukechampine: `if time.Since(worker.recentUploadFailure) < time.Minute * 120`

modules/renter/repair.go
```diff
@@ -26,323 +25,374 @@ var renewThreshold = func() types.BlockHeight {
...
+						}
+					}
+				}
+				// Update the gap counts if they have been affected in any
+				// way.
+				if len(chunkGaps.contracts) < len(chunkGaps.pieces) && len(chunkGaps.contracts) < chunkGaps.numGaps {
+					oldNumGaps := chunkGaps.numGaps
+					chunkGaps.numGaps = len(chunkGaps.contracts)
+					gapCounts[oldNumGaps] = gapCounts[oldNumGaps] - 1
+					gapCounts[chunkGaps.numGaps] = gapCounts[chunkGaps.numGaps] + 1
```
lukechampine: use `--` and `++`

modules/renter/repair.go
```diff
@@ -26,323 +25,374 @@ var renewThreshold = func() types.BlockHeight {
...
-				offline[p.Chunk] = append(offline[p.Chunk], p.Piece)
-			}
+	// Create the repair matrix. The repair matrix is a set of chunks,
+	// linked from chunk id to the set of hosts that do not have that
+	// chunk.
+	repairMatrix, gapCounts := r.createRepairMatrix(availableWorkers)
+	maxGaps := 0
+	for i := 1; i < len(gapCounts); i++ {
+		if gapCounts[i] > 0 {
+			maxGaps = i
```
lukechampine: I'm confused by `gapCounts`. What is it a `map[int]int` if you don't ever use the second `int`? Could it be a `[]bool`? Could the "maximum" code be handled inside `createRepairMatrix`, so that it just returns `maxGaps`?

modules/renter/repair.go
```diff
@@ -26,323 +25,374 @@ var renewThreshold = func() types.BlockHeight {
...
+			}
+			// encrypt pieces
+			for i := range pieces {
+				key := deriveKey(file.masterKey, chunkIndex, uint64(i))
+				pieces[i], err = key.EncryptBytes(pieces[i])
+				if err != nil {
+					// TODO: Manage this error.
+					continue
+				}
+			}
```
lukechampine: this is probably worth splitting into a helper function

lukechampine:
My general impression of this is that it would benefit from a few well-named helper functions/methods, and some intermediate variables during complex manipulations. The download code is especially dense and hard to understand.

modules/renter/repair.go
DavidVorick: I generally like being explicit about making things b/c I often `exists` a lot, also for slices I feel like there's a pretty strong difference between an empty slice and `nil`. I don't think I use either of those things in this function, but it still bothers me to have a map that's missing elements when I expect them to be there.

modules/renter/repair.go
DavidVorick: because that int is really important

DavidVorick: Mostly it makes it easier to think about if I don't have to worry about the object getting copied in any way. The entire repair loop happens on a single thread, so there's no safety to worry about with objects created within the repair loop.

modules/renter/repair.go
DavidVorick: ah there are supposed to be return statements here, got the control flow messed up.

modules/renter/worker.go
DavidVorick: I felt it was a lot easier to handle here. The repair loop itself receives the completed jobs all down the same channel, and so doesn't have a whole lot of context about what exactly the change was. It complicates the repair loop a lot to have to think about more than just figuring out how to reassign workers.

I will probably move the download / file write code into the worker as well. It'll keep the repair loop and download loop focused on assigning work, instead of handling completed jobs by... completing them more? I feel like part of the job of the worker is to update all the persist, etc.

modules/renter/download.go
```diff
@@ -1,182 +1,290 @@
...
-		pieceMap:   pieceMap,
-		masterKey:  masterKey,
+	r.mu.RUnlock(id)
+
+	// Grab the next file to be downloaded.
+	id = r.mu.RLock()
+	if len(r.downloadQueue) == 0 {
+		// TODO: Block until either there's another download, or until
+		// shutdown, or until enough time has passed that the workers should be
+		// reassembled?
```
DavidVorick: yeah, the way it's currently coded it will spin. I have not fully decided the best way to handle blocking here.

modules/renter/download.go
DavidVorick: heh.

I don't even know what that line does at this point

modules/renter/download.go
DavidVorick: Pieces are not appended in order.

DavidVorick: I guess you are saying that the chunk map should be built up here instead of where it is done later in the code. I'll think about it, but that feels like over-optimization to me.

modules/renter/download.go
DavidVorick: I'll write a TODO for it.

modules/renter/repair.go
DavidVorick: I don't think I realized that.

modules/renter/repair.go
DavidVorick: you do use the second int. The second int tells you how far you are from zero

DavidVorick: Once it reaches zero, you start requiring more workers before you  go search for a chunk in the repair matrix.

DavidVorick:
I would have thought the upload code would be harder to understand. The download code doesn't even support downloading multiple files at the same time lol.

But yes, I definitely plan on adding more helper functions, this was just a first pass.

modules/renter/download.go
lukechampine: Instead of appending, you assign them to their proper index in the `[][]byte`. This is the format expected by `Recover`

api/renterhost_test.go
```diff
@@ -158,7 +158,7 @@ func TestHostAndRent(t *testing.T) {
 	// Mine blocks until the host recognizes profit. The host will wait for 12
 	// blocks after the storage window has closed to report the profit, a total
 	// of 40 blocks should be mined.
-	t.Skip("TODO: NEED TO GET THE CONTRACT STUFF WORKING AGAIN????")
+	t.Skip("TODO: NEED TO GET THE CONTRACT STUFF WORKING AGAIN")
```
DavidVorick: Ah darn.

DavidVorick:
Unfortunately this pull request ended up touching a lot of packages. But it fixes up a good handful of TODOs and does not create much new technical debt. Where new technical debt has been created, it is because greater technical debt was cleaned up previously.

Sorry that the commits are all over the place. I will walk through the diffs and leave comments regarding what changed and why.

api/client_test.go
```diff
@@ -6,10 +6,11 @@ import (
 
 // TestApiClient tests that the API client connects to the server tester and
 // can call and decode routes correctly.
-func TestIntegrationApiClient(t *testing.T) {
+func TestApiClient(t *testing.T) {
 	if testing.Short() {
 		t.SkipNow()
 	}
+	t.Parallel()
```
DavidVorick: API tests were running too slowly, and I was relying primarily on API tests because I know that we plan to change more of the renter + contractor internals in the near future (mainly the persist stuff). While I was adding t.Parallel() to everything, I also removed the obsoleted 'Integration' word from the testing names, where I ran into it. I was not super thorough.

api/ecosystem_test.go
```diff
@@ -3,15 +3,192 @@ package api
...
+		values := url.Values{}
+		values.Set("path", st.dir)
+		values.Set("size", "1048576")
+		err := st.stdPostAPI("/host/storage/folders/add", values)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
DavidVorick: I added a ton of ecosystem helper functions which make it easier to set up and command fleets of hosts and renters. This was needed for the current testing, as the renter now has a lot more parallelism and I needed to make sure that it could handle a lot of conditions. Notably missing from the current test suite are situations where hosts go offline.

api/renter.go
```diff
@@ -8,25 +11,57 @@ import (
...
+	}).(float64)
+
+	// requiredRenewWindow establishes the minimum allowed renew window for the
+	// renter settings. This minimum is here to prevent users from shooting
+	// themselves in the foot.
+	requiredRenewWindow = build.Select(build.Var{
+		Standard: types.BlockHeight(288),
+		Dev:      types.BlockHeight(1),
+		Testing:  types.BlockHeight(1),
+	}).(types.BlockHeight)
```
DavidVorick: I added API support for choosing redundancy and host count. Along with this API support, I added some constants to specify what the default values would be, and what the minimum allowed values would be. Minimum values are obviously easy to circumvent by recompiling, but if you are advanced enough to do that then you deserve whatever footguns you create for yourself.

api/renter.go
```diff
@@ -275,17 +331,57 @@ func (api *API) renterUploadHandler(w http.ResponseWriter, req *http.Request, ps
...
+			return
+		}
+
+		// Create the erasure coder.
+		ec, err = renter.NewRSCode(dataPieces, parityPieces)
+		if err != nil {
+			WriteError(w, Error{"Unable to encode file using the provided parameters: " + err.Error()}, http.StatusBadRequest)
+			return
+		}
+	}
```
DavidVorick: Should be noted that the testing here is not really complete. It's probably in our interest not to merge this until coverage is at 100%.

build/errors.go
```diff
@@ -5,6 +5,39 @@ import (
...
+
+// ExtendErr will return a new error which extends the input error with a
+// string. If the input error is nil, then 'nil' will be returned, discarding
+// the input string.
+func ExtendErr(s string, err error) error {
+	if err == nil {
+		return nil
+	}
+	return errors.New(s + ": " + err.Error())
+}
```
DavidVorick: Looks like I did not bring over the testing for these.

I have these in the host branch and was not intending to merge them into master until the host branch was ready, but... well I get what I deserve. These really should have been a separate PR the whole time. If you push hard enough I'll probably cave and actually create a separate PR for them.

DavidVorick: I find these functions to be incredibly useful for building coherent error stacks.

modules/renter/contractor/uptime.go
```diff
@@ -27,22 +28,33 @@ var uptimeWindow = func() time.Duration {
...
+	// Get the net address associated with the contract.
+	//
+	// TODO: This should eventually be updated to query the host by public key.
+	contract, exists := c.contracts[id]
+	if !exists {
+		return false
+	}
+	addr := contract.NetAddress
+	// Look up the host by address.
+	//
```
DavidVorick: beh. Cleaning that up too.

modules/renter/contractor/uptime_test.go
```diff
@@ -136,21 +144,26 @@ func TestIsOffline(t *testing.T) {
...
-		// recent scan was good (and within uptimeWindow of oldBadScan, but that shouldn't matter)
-		{[]modules.HostDBScan{oldBadScan, oldBadScan, oldBadScan, oldGoodScan}, false},
-		// recent scan was good (and outside uptimeWindow of oldBadScan, but that shouldn't matter)
-		{[]modules.HostDBScan{oldBadScan, oldBadScan, oldBadScan, currentGoodScan}, false},
+		// old scan was good, recent scans are bad.
+		{[]modules.HostDBScan{oldGoodScan, newBadScan, newBadScan, currentBadScan}, true},
+		// recent scan was good, with many recent bad scans.
+		{[]modules.HostDBScan{oldBadScan, newGoodScan, newBadScan, currentBadScan, currentBadScan}, false},
+		// recent scan was good, old scans were bad.
+		{[]modules.HostDBScan{oldBadScan, newBadScan, currentBadScan, currentGoodScan}, false},
```
DavidVorick: Added some extra tests here because I felt the existing testing was incomplete. Implementation seems solid, but you should double check that the tests make sense.

modules/renter/files.go
```diff
@@ -25,13 +25,14 @@ var (
...
 	contracts   map[types.FileContractID]fileContract
-	masterKey   crypto.TwofishKey
-	erasureCode modules.ErasureCoder
-	pieceSize   uint64
-	mode        uint32 // actually an os.FileMode
-	mu          sync.RWMutex
+	masterKey   crypto.TwofishKey    // Static - can be accessed without lock.
+	erasureCode modules.ErasureCoder // Static - can be accessed without lock.
+	pieceSize   uint64               // Static - can be accessed without lock.
+	mode        uint32               // actually an os.FileMode
```
DavidVorick: I'm 95% sure that I did not change any of the persist stuff for the renter. But maybe I missed something on accident.

DavidVorick:
Looks like when I made the API tests all run in parallel I triggered a bunch of NDFs that have been there for a while. I'm not sure how much I want to clean them up, I may de-parallelize the worst offenders and then tackle them in another pull request.

My guess though is that at least a few of the NDFs are the result of latent bugs, so we should plan on getting to them at some point in the near future.

DavidVorick:
Besides the NDF stuff, this PR I think is basically good to go. It's probably worth throwing it through the antfarm.

modules/renter/repair.go
```diff
@@ -1,365 +1,473 @@
...
+
+		// Update the number of gaps for this chunk.
+		numGaps := chunkStatus.numGaps(rs)
+		rs.gapCounts[chunkStatus.recordedGaps]--
+		rs.gapCounts[numGaps]++
+		chunkStatus.recordedGaps = numGaps
+
+		// Update the set of useful workers and the set of missing pieces.
+		missingPieces = missingPieces[1:]
+		usefulWorkers = usefulWorkers[1:]
```
lukechampine: more readable if this is at the top of the loop:
```go
id, usefulWorkers := usefulWorkers[0], usefulWorkers[1:]
p, missingPieces := missingPieces[0], missingPieces[1:]
```

modules/renter/repair.go
```diff
@@ -1,365 +1,473 @@
...
-		f.mu.RUnlock()
-		if err != nil {
-			// If saving failed for this chunk, it will probably fail for the
-			// next chunk as well. Better to try again on the next cycle.
-			r.log.Printf("failed to save repaired file %v: %v", f.name, err)
+		// Chill out for an extra 5 minutes before going through the files
+		// again.
+		select {
+		case <-time.After(time.Minute * 5):
+		case <-r.tg.StopChan():
```
lukechampine: I wrote a nice [helper package](https://github.com/lukechampine/sleep/blob/master/sleep.go) for this sort of thing:
```go
interrupted := sleep.Lightly(5 * time.Minute, r.tg.StopChan())
if interrupted {
    return
}
```
(not that we should be importing it, but this pattern seems common enough that we may want to duplicate it)

modules/renter/worker.go
```diff
@@ -0,0 +1,287 @@
...
+
+// upload will perform some upload work.
+func (w *worker) upload(uw uploadWork) {
+	e, err := w.renter.hostContractor.Editor(w.contractID)
+	if err != nil {
+		w.recentUploadFailure = time.Now()
+		select {
+		case uw.resultChan <- finishedUpload{uw.chunkID, crypto.Hash{}, err, uw.pieceIndex, w.contractID}:
+		case <-w.renter.tg.StopChan():
+		}
```
lukechampine: may want to define a helper for this. You can do it inline with `w.upload` so that it captures its variables:
```go
func (w *worker) upload(uw uploadWork) {
    sendResult := func(u finishedUpload) {
        select {
        case uw.resultChan <- u:
        case <-w.renter.tg.StopChan():
        }
    }
    // ...
```
Also, I noticed that `w.download` doesn't have an equivalent for stopping early.

modules/renter/worker.go
```diff
@@ -0,0 +1,287 @@
...
+
+// work will perform one unit of work, exiting early if there is a kill signal
+// given before work is completed.
+func (w *worker) work() {
+	// Check for priority downloads.
+	select {
+	case d := <-w.priorityDownloadChan:
+		w.download(d)
+		return
+	default:
```
lukechampine: should probably check `w.killChan` and `tg.StopChan` first

api/renter.go
```diff
@@ -275,17 +331,57 @@ func (api *API) renterUploadHandler(w http.ResponseWriter, req *http.Request, ps
...
+		_, err = fmt.Sscan(req.FormValue("paritypieces"), &parityPieces)
+		if err != nil {
+			WriteError(w, Error{"Unable to read parameter 'paritypieces': " + err.Error()}, http.StatusBadRequest)
+			return
+		}
+
+		// Verify that sane values for parityPieces and redundancy are being
+		// supplied.
+		if parityPieces < requiredParityPieces {
+			WriteError(w, Error{fmt.Sprintf("A minimum of %v parity pieces is required, but %v parity pieces requested.", parityPieces, requiredParityPieces)}, http.StatusBadRequest)
```
DavidVorick: I don't think we add context to our api errors. But it's something I could see us doing in the future.

Open to explore this more, but during this PR I was assuming that the api error would be directly fed to the user.

DavidVorick: I guess you could at least add context explaining what call failed.

build/errors.go
DavidVorick: I think we would need to replace a couple of functions. Seems like something we can do in a later PR.

DavidVorick:
The high data churn only happens during the last 15% or so of the file being uploaded, because it's doing substantially more erasure coding and encryption during the last 15% than during first 85%. I think I can make some adjustments that will help a lot.

doc/api/Renter.md
```diff
@@ -269,8 +281,15 @@ uploads a file to the network from the local filesystem.
 
 ###### Query String Parameters
 ```
+// The number of data pieces to use when erasure coding the file.
+datapieces // int
+
+// The number of parity pieces to use when erasure coding the file. Total
+// redundancy of the file is (datapieces+paritypieces)/datapieces.
+paritypieces // int
```
DavidVorick: huh

DavidVorick: this comment confused me lol. I think you just meant to say 'good job' though. At first I thought I had misformatted the comment, but I don't think there's a mistake here.

modules/renter/contractor/uptime_test.go
```diff
@@ -104,12 +104,20 @@ func TestIntegrationReplaceOffline(t *testing.T) {
...
+	if len(m.hosts) == 0 {
+		return modules.HostDBEntry{}, false
+	}
+	if len(m.hosts) != 1 {
+		panic("oops")
+	}
+	for _, host := range m.hosts {
+		return host, true
+	}
+	return modules.HostDBEntry{}, false
```
DavidVorick: hosts in the test don't have an addr anymore. So, no matter what addr you pass to the function, it'll return the only host in the map, panicking if there's more than one host in the map.

it's hacky

modules/renter/worker.go
DavidVorick: We use this pattern in a lot of places. Is there some way we could make it more generic?

e.g. tg.SendStuffOrQuitEarly(chan, data)

DavidVorick: and, the download should probably have an equivalent.

modules/renter/worker.go
DavidVorick: They get checked in the work loop, immediately before `work()` is called.

api/renter.go
lukechampine: we do in siac: `call to /foo/bar failed: [error]`

doc/api/Renter.md
lukechampine: haha yeah, I was just happy to see that you updated the docs to reflect the API change

modules/renter/contractor/uptime_test.go
lukechampine: why don't you just change the map key to a fcid?

modules/renter/worker.go
lukechampine: I wish haha, but there isn't really a way to do a generic `select`, unless you want to invoke the chaos gods of `reflect`. By which I mean: I could probably figure out a way to do it, but it wouldn't be pretty.

modules/renter/worker.go
lukechampine: is it possible to interrupt `w.upload`/`w.download`? Otherwise you're stuck waiting for a transfer to complete before shutdown will finish. Not terrible, but annoying.

I guess if you wanted to do it properly, you'd need to `Close` the actual connection and handle that gracefully, which is probably nontrivial.

modules/renter/worker.go
lukechampine:
```go
func abortableSend(sendCh, abortCh, data interface{}) {
    case1 := reflect.SelectCase{
        Dir:  reflect.SelectSend,
        Chan: reflect.ValueOf(sendCh),
        Send: reflect.ValueOf(data),
    }
    case2 := reflect.SelectCase{
        Dir:  reflect.SelectRecv,
        Chan: reflect.ValueOf(abortCh),
    }
    reflect.Select([]reflect.SelectCase{case1, case2})
}
```

DavidVorick:
Some notes on the current state:

1. All direct-to-code comments addressed
2. Pretty clear that there's some sort of memory leak in the download code, causing *significant* issues.
3. The download code hierarchy is pretty not-cool. Not sure how easy it will be to refactor without rewriting a ton.
4. There's a problem with the proto package where it's not properly handling revision number mismatches. Uploads and downloads both are being affected in significant ways due to this.
5. Best I can tell, the download code is ignorant as to whether the target worker actually has the data or not. Meaning if you've got 30 contracts and you are downloading a file that's only halfway uploaded, you'll do a lot of "hey do you have this piece" -> "no lol you never uploaded to me" types of interactions.

DavidVorick:
Other than the proto issues, the upload code seems to be working smoothly.

lukechampine:
can you elaborate on the proto bug?

DavidVorick:
`upload err: our revision number (464) does not match the host's (465)` -> this has happened to more than half of my contracts. I thought we had code to gracefully recover from this, (cached revisions), but seemingly not. Are the cached revisions persistent?

DavidVorick:
Worth noting that uploads and downloads can be running at the same time, or at the very least the renter could be switching between them rapidly. Which means the cached revisions may fail or something? I'm not sure.

Do the cached revisions get Fsync'd? Though, I haven't had any power disruptions so that shouldn't actually matter.

lukechampine:
I would drop some `println`s in after each revision (upload or download) containing the FCID and revision number. Then you could see if they are interfering with each other. Although, the contractor already has safeguards to prevent you from uploading to a contract if you're already downloading, and vis versa.

modules/renter/download.go
```diff
@@ -136,6 +132,9 @@ func newDownload(f *file, destination string) *download {
 
 		downloadFinished: make(chan error),
 	}
+	// Allocate the piece size and progress bar so that the
+	d.reportedPieceSize = d.fileSize / (d.numChunks * uint64(d.erasureCode.MinPieces()))
```
lukechampine: comment

DavidVorick: drat

modules/renter/download.go
```diff
@@ -136,6 +132,9 @@ func newDownload(f *file, destination string) *download {
...
+	d.atomicDataReceived = d.fileSize - (d.reportedPieceSize * d.numChunks * uint64(d.erasureCode.MinPieces()))
```
lukechampine: this means that downloads will show a few bytes as received before anything happens?

DavidVorick: yes. potentially up to but not including the total number of pieces in the file.

DavidVorick: *data pieces

modules/renter/download.go
```diff
@@ -1,182 +1,552 @@
...
-	// fetch returns the data specified by piece metadata.
-	fetch(pieceData) ([]byte, error)
-}
+		// completedPieces contains information about which pieces have been
+		// successfully downloaded.
+		//
+		// workerAttempts contains a list of workers that are able to fetch a
+		// piece of the chunk, mapped to an indication of whether or not they
+		// have tried to fetch a piece of the chunk.
+		completedPieces map[uint64][]byte
```
DavidVorick: I needed a quick way to see how many pieces had been added. Not sure if I still need that, but everything is working as-is for now.

modules/renter/download.go
```diff
@@ -1,182 +1,552 @@
...
+		// Syncrhonization tools.
+		downloadFinished chan error
+		mu               sync.Mutex
+	}
+
+	// downloadState tracks all of the stateful information within the download
+	// loop, primarily used to simplify the use of helper functions. There is
+	// no thread safety with the download state, as it is only ever accessed by
+	// the primary download loop thread.
+	downloadState struct {
```
DavidVorick: logging is the biggest thing I think. But overall the hierarchy is pretty thoroughly overlapped. It would take a pretty big re-write to fix that though.

modules/renter/download.go
```diff
@@ -1,182 +1,552 @@
...
+		incompleteChunk.download.mu.Lock()
+		downloadComplete := incompleteChunk.download.downloadComplete
+		incompleteChunk.download.mu.Unlock()
+		if downloadComplete {
+			// The download has most likely failed. No need to complete this
+			// chunk.
+			ds.activePieces-- // for the current incomplete chunk
+			for range incompleteChunk.completedPieces {
+				ds.activePieces-- // for the completed chunk
+			}
```
DavidVorick: I am not sure what happened here.

api/renter.go
```diff
@@ -175,11 +175,11 @@ func (api *API) renterHandlerPOST(w http.ResponseWriter, req *http.Request, _ ht
 	if req.FormValue("hosts") != "" {
 		_, err := fmt.Sscan(req.FormValue("hosts"), &hosts)
 		if err != nil {
-			WriteError(w, Error{"Couldn't parse hosts: " + err.Error()}, http.StatusBadRequest)
+			WriteError(w, Error{"unable to parse hosts: " + err.Error()}, http.StatusBadRequest)
 			return
 		}
 		if hosts < requiredHosts {
-			WriteError(w, Error{fmt.Sprintf("Insufficient number of hosts, need at least %v but have %v.", recommendedHosts, hosts)}, http.StatusBadRequest)
+			WriteError(w, Error{fmt.Sprintf("insuccificent number of hosts, need at least %v but have %v", recommendedHosts, hosts)}, http.StatusBadRequest)
```
lukechampine: lol

modules/renter/download.go
```diff
@@ -1,182 +1,552 @@
...
+		// unless no more workers exist who can download pieces for that chunk,
+		// in which case the download has failed.
+		//
+		// resultChan is the channel that is used to receive completed worker
+		// downloads.
+		activePieces     int
+		activeWorkers    map[types.FileContractID]struct{}
+		availableWorkers []*worker
+		incompleteChunks []*chunkDownload
+		resultChan       chan finishedDownload
```
lukechampine: there are only three channel types: send-only, receive-only, and bidirectional.

PR #1508, opened 2016-11-30
build.Critical if decrementReliability is called on nonexistent host in hostdb

johnathanhowell:


modules/renter/hostdb/scan.go
```diff
@@ -99,7 +99,7 @@ func (hdb *HostDB) decrementReliability(addr modules.NetAddress, penalty types.C
 	// Look up the entry and decrement the reliability.
 	entry, exists := hdb.allHosts[addr]
 	if !exists {
-		// TODO: should panic here
+		build.Critical("host to be decremented did not exist in hostdb")
```
DavidVorick: This statement is a TODO because there's a bug where the code is getting hit frequently I believe. We had a hard time finding it, and so chose to leave the bug rather than to build.Critical, otherwise our daemons would be crashing all the time.

Basically, need to fix the bugs before we can merge this.

DavidVorick: it does scream that there are problems though, with the idea that the user should make bug reports about them.

They are too common and too poorly understood to put into production. And all of our nodes usually run in DEBUG mode so that we can find the tricky stuff.

DavidVorick:
Merged this as we intend to fix this by the next version.

Hopefully it doesn't cause to many problems for people who build directly from master, on the other hand it might help us get to the bottom of the problem faster.

PR #1510, opened 2016-12-06
Contractor leeway

lukechampine:
The contractor will now form 30 contracts by default, and spend 50% less money on them. I have not added testing yet because I would like some feedback on how the lowballing was implemented. In short, I modified `maxSectors` to return half its previous value. The contractor calls `maxSectors` whenever it is calculating how much to spend on storage, so this seemed like an easy way to implement lowballing at the root, rather than modifying every function that forms or renews contracts. Still, I am wary that this may be too clever.

### DIFF
--- a/modules/renter/contractor/allowance.go
+++ b/modules/renter/contractor/allowance.go
@@ -59,11 +59,16 @@ func (c *Contractor) SetAllowance(a modules.Allowance) error {
 		return errAllowanceWindowSize
 	}
 
-	// check that allowance is sufficient to store at least one sector
-	numSectors, err := maxSectors(a, c.hdb, c.tpool)
+	// calculate the maximum sectors this allowance can store
+	max, err := maxSectors(a, c.hdb, c.tpool)
 	if err != nil {
 		return err
-	} else if numSectors == 0 {
+	}
+	// Only allocate half as many sectors as the max. This leaves some leeway
+	// for replacing contracts, transaction fees, etc.
+	numSectors := max / 2
+	// check that this is sufficient to store at least one sector
+	if numSectors == 0 {
 		return ErrInsufficientAllowance
 	}
 
--- a/modules/renter/contractor/renew.go
+++ b/modules/renter/contractor/renew.go
@@ -11,8 +11,8 @@ import (
 )
 
 // managedRenew negotiates a new contract for data already stored with a host.
-// It returns the new contract. This is a blocking call that
-// performs network I/O.
+// It returns the new contract. This is a blocking call that performs network
+// I/O.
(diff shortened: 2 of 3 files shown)


DavidVorick:
It is too clever, maxSectors should tell you the maximum number of sectors.

I'm also trying to think of other optimizations we would want in the future, without getting in the way of those.

For example, you probably want to allocate more money for downloads to hosts with cheap+fast bandwidth, and less money for downloads to other hosts. This is an optimization that we won't be adding in the near future, but it'd be great if we built the current architecture to be flexible enough to support that in the future.

DavidVorick:
Thinking a bit more, where we ultimately want to get to is probably this flow:

1. renter gets an allowance with some indication of distribution needs (geography, download frequency, etc.)
2. hostdb trees are created to match the distribution needs, likely 1 tree per set of needs
3. hosts are selected from the trees. e.g. 1.2x redundancy selected from the 'cheap downloads' pool, 0.4x redundancy selected from the 'cheap + Asia', 1x redundancy selected from 'cheap + US', and 0.5x redundancy selected from 'cheap + Europe', etc.
4. The renter figures out how much money to put in each contract.
    - estimates a storage, download, fee, etc. cost for each contract, gets a fixed value for each contract
    - adds all the fixed values together to get a 'total proportion' cost
    - multiplies that against the allowance to get the actual amount of money to fund each contract
    - multiplies that final value against some 'hold X% in reserve for repairs and stuff' value.
5. The renter then has the contractor create each contract individually with the explicit values for the contract.

I know this is not how the code is currently divided. But if we agree that the above is a good final target (feel free to critique), then we can figure out how best introduce changes into the current codebase to move that direction.

lukechampine:
The big changes I see there are the hostdb creating multiple trees and the renter taking responsibility for choosing hosts. My inclination is to avoid a design where the renter is responsible for choosing hosts, forming contracts, renewing contracts, uploading data, downloading data, monitoring uptime, _and_ providing a high-level "file" abstraction to the user. I would very much like to divide these responsibilities across multiple packages. The problem is that they are all intricately bound up in each other. For example, I can imagine having a "downloader" package that is solely responsible for implementing the download algorithm. But the hostdb needs to be updated using metrics generated by the downloader, and the contractor needs to know when a host's reliability falls too low so that it can form a replacement contract, and the "repair-er" needs to know when the contractor forms a new contract so that it can replicate the data, and when the data is replicated, the file metadata needs to swap out the old host for the new host, and then the downloader needs to be informed of the new host so that it can download from them! And there are multiple cyclic interactions like this going on simultaneously. So we either need to design things very carefully, or resign ourselves to throwing everything in one giant package.

DavidVorick:
This now needs some testing. The PR that I am working on for the renter - to get the testing I need to verify that the parallelism is working correctly has me adding back the ability for the user to set the values for the recommendedHosts, so that change can be removed from this PR to prevent conflicts.

PR #1511, opened 2016-12-06
Build vars

lukechampine:
This is a demonstration of how we might clean up our myriad release-dependent variables. The declarations are more compact, and no longer require a separate `panic` call. The downside is that a type assertion is required due to `Select` returning an `interface{}`. This type assertion will be hard to forget though, since any attempt to use the variable as e.g. an `int` will cause a compilation error.

I opted to only replace the vars in the miner package in order to avoid merge conflicts. If we decide to go ahead, we should also update the consensus set, gateway, transaction pool, and siad packages. The other packages are currently being overhauled, so we should delay updating them until their feature branches have been merged. Fortunately, it's easy to roll this out incrementally, since the existing method of selecting a build variable will continue to work as before.

EDIT: it occurred to me that we probably introduced some merge conflicts when we merged the Cmp64 PR. Drat :/

### DIFF
--- a/build/var.go
+++ b/build/var.go
@@ -1,0 +1,27 @@
+package build
+
+// A Var represents a variable whose value depends on which Release is being
+// compiled. None of the fields may be nil, and all fields must have the same
+// underlying type.
+type Var struct {
+	Standard interface{}
+	Dev      interface{}
+	Testing  interface{}
+}
+
+// Select returns the field of v that corresponds to the current Release.
+func Select(v Var) interface{} {
+	if v.Standard == nil || v.Dev == nil || v.Testing == nil {
+		panic("nil value in build variable")
+	}
+	switch Release {
+	case "standard":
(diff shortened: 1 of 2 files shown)


DavidVorick:
I think this looks good to me. We'll have to put this PR on hold though until we get the wallet stuff and host stuff merged in.

lukechampine:
Well, like I said, it's easy to do it incrementally. We can merge it now (well, after adding consensus/tpool/siad) and fixup the wallet/host/renter later, or we can wait until everything is merged and then fixup everything in a big PR. I'm in favor of the former because it means one less open PR to worry about; the only disadvantage is that we won't be uniformly using `built.Select` everywhere, which may be confusing.

DavidVorick:
it's probably okay. The Linux kernel for example does a lot of incremental upgrades like this.

PR #1512, opened 2016-12-07
More Build.Vars

lukechampine:
Also prevent unkeyed `Var` literals, and make `Select` more restrictive wrt assignability,

One unintended consequence is that you can't write:
```go
var foo = build.Select(build.Var{
    Standard: 1,
    Dev:      2,
    Testing:  3,
}.(uint64)
```
Because the numbers will be interpreted as `int`s, causing the type assertion to fail. In the old pattern, they were interpreted as constants, so they could be automatically converted to the function's return type. Now you have to explicitly cast them:
```go
var foo = build.Select(build.Var{
    Standard: uint64(1),
    Dev:      uint64(2),
    Testing:  uint64(3),
}.(uint64)
```
Which is unfortunate, but not the end of the world.

As discussed, the other modules will remain unconverted for now.


build/var.go
```diff
@@ -1,19 +1,40 @@
 package build
 
+import "reflect"
```
DavidVorick:
```go
import (
    "reflect"
)
```

PR #1513, opened 2016-12-08
add hostdb subpackage `hosttree` for storing and retreiving weighted hosts

johnathanhowell:
This PR introduces the `hosttree` subpackage to the renter's hostdb. This package exposes a minimal interface for inserting, deleting, modifying, and fetching hosts from the tree. A later PR will implement this package with the `hostdb`. Most of the testing from the previous `weightedlist` implementation has been preserved. Hosts are now indexed by Public Key instead of NetAddress.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ dependencies:
 run = Test
 pkgs = ./api ./build ./compatibility ./crypto ./encoding ./modules ./modules/consensus \
        ./modules/explorer ./modules/gateway ./modules/host ./modules/host/storagemanager \
-       ./modules/renter ./modules/renter/contractor ./modules/renter/hostdb ./modules/renter/proto \
+       ./modules/renter ./modules/renter/contractor ./modules/renter/hostdb ./modules/renter/hostdb/hosttree ./modules/renter/proto \
        ./modules/miner ./modules/wallet ./modules/transactionpool ./persist ./siac ./siad ./sync ./types
 
 # fmt calls go fmt on all packages.
--- a/Makefile
+++ b/Makefile
@@ -29,10 +29,12 @@ dependencies:
 # pkgs changes which packages the makefile calls operate on. run changes which
 # tests are run during testing.
 run = Test
-pkgs = ./api ./build ./compatibility ./crypto ./encoding ./modules ./modules/consensus \
-       ./modules/explorer ./modules/gateway ./modules/host ./modules/host/storagemanager \
-       ./modules/renter ./modules/renter/contractor ./modules/renter/hostdb ./modules/renter/hostdb/hosttree ./modules/renter/proto \
-       ./modules/miner ./modules/wallet ./modules/transactionpool ./persist ./siac ./siad ./sync ./types
+pkgs = ./api ./build ./compatibility ./crypto ./encoding ./modules \
+       ./modules/consensus ./modules/explorer ./modules/gateway ./modules/host \
+       ./modules/host/storagemanager ./modules/renter ./modules/renter/contractor \
+       ./modules/renter/hostdb ./modules/renter/hostdb/hosttree ./modules/renter/proto \
+       ./modules/miner ./modules/wallet ./modules/transactionpool ./persist ./siac \
+       ./siad ./sync ./types 
 
 # fmt calls go fmt on all packages.
 fmt:
--- a/modules/renter/hostdb/hosttree/hosttree.go
+++ b/modules/renter/hostdb/hosttree/hosttree.go
@@ -46,8 +46,8 @@ type (
 		// hosts is a map of public keys to nodes.
 		hosts map[string]*node
 
-		// weight applies a weight to a hostEntry
-		weight WeightFunc
+		// weightFn calculates the weight of a hostEntry
+		weightFn WeightFunc
 
 		mu sync.Mutex
 	}
@@ -91,8 +91,8 @@ func New(wf WeightFunc) *HostTree {
 		root: &node{
 			count: 1,
 		},
-		weight: wf,
-		hosts:  make(map[string]*node),
+		weightFn: wf,
+		hosts:    make(map[string]*node),
 	}
 }
 
@@ -189,7 +189,7 @@ func (ht *HostTree) Insert(hdbe modules.HostDBEntry) error {
 
 	entry := &hostEntry{
 		HostDBEntry: hdbe,
-		weight:      ht.weight(hdbe),
+		weight:      ht.weightFn(hdbe),
 	}
 
 	if _, exists := ht.hosts[string(entry.PublicKey.Key)]; exists {
@@ -232,7 +232,7 @@ func (ht *HostTree) Modify(hdbe modules.HostDBEntry) error {
 
 	entry := &hostEntry{
 		HostDBEntry: hdbe,
-		weight:      ht.weight(hdbe),
+		weight:      ht.weightFn(hdbe),
 	}
 
 	_, node = ht.root.recursiveInsert(entry)
--- a/modules/renter/hostdb/hosttree/hosttree.go
+++ b/modules/renter/hostdb/hosttree/hosttree.go
@@ -97,7 +97,6 @@ func (n *node) recursiveInsert(entry *HostEntry) (nodesAdded int, newnode *node)
 		n.entry = entry
 		n.taken = true
 		n.weight = entry.Weight
-		nodesAdded = 0
 		newnode = n
 		return
 	}
@@ -251,7 +250,10 @@ func (ht *HostTree) Fetch(n int, ignore []types.SiaPublicKey) ([]modules.HostDBE
 	}
 
 	for _, entry := range removedEntries {
-		ht.Insert(entry)
+		err := ht.Insert(entry)
+		if err != nil {
+			return hosts, err
+		}
 	}
 
 	return hosts, nil
--- a/modules/renter/hostdb/hosttree/hosttree.go
+++ b/modules/renter/hostdb/hosttree/hosttree.go
@@ -3,6 +3,7 @@ package hosttree
 import (
 	"crypto/rand"
 	"errors"
+	"sync"
 
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/types"
@@ -40,6 +41,8 @@ type (
 
 		// hosts is a map of public keys to nodes.
 		hosts map[string]*node
+
+		mu sync.Mutex
 	}
 
 	// HostEntry is an entry in the host tree.
@@ -175,6 +178,9 @@ func (n *node) remove() {
 // Insert inserts the entry provided to `entry` into the host tree. Insert will
 // return an error if the input host already exists.
 func (ht *HostTree) Insert(entry *HostEntry) error {
+	ht.mu.Lock()
+	defer ht.mu.Unlock()
+
 	if _, exists := ht.hosts[string(entry.PublicKey.Key)]; exists {
 		return ErrHostExists
 	}
@@ -187,6 +193,9 @@ func (ht *HostTree) Insert(entry *HostEntry) error {
 
 // Remove removes the host with the public key provided by `pk`.
 func (ht *HostTree) Remove(pk types.SiaPublicKey) error {
+	ht.mu.Lock()
+	defer ht.mu.Unlock()
+
 	node, exists := ht.hosts[string(pk.Key)]
 	if !exists {
 		return ErrNoSuchHost
@@ -200,6 +209,9 @@ func (ht *HostTree) Remove(pk types.SiaPublicKey) error {
 // Modify updates a host entry at the given public key, replacing the old entry
 // with the entry provided by `newEntry`.
 func (ht *HostTree) Modify(entry *HostEntry) error {
+	ht.mu.Lock()
+	defer ht.mu.Unlock()
+
 	node, exists := ht.hosts[string(entry.PublicKey.Key)]
 	if !exists {
 		return ErrNoSuchHost
@@ -217,6 +229,9 @@ func (ht *HostTree) Modify(entry *HostEntry) error {
 // The hosts that are returned first have the higher priority. Hosts passed to
 // `ignore` will not be considered; pass `nil` if no blacklist is desired.
 func (ht *HostTree) Fetch(n int, ignore []types.SiaPublicKey) ([]modules.HostDBEntry, error) {
+	ht.mu.Lock()
+	defer ht.mu.Unlock()
+
 	var hosts []modules.HostDBEntry
 	var removedEntries []*HostEntry
 
@@ -250,10 +265,9 @@ func (ht *HostTree) Fetch(n int, ignore []types.SiaPublicKey) ([]modules.HostDBE
 	}
 
 	for _, entry := range removedEntries {
-		err := ht.Insert(entry)
-		if err != nil {
-			return hosts, err
-		}
+		_, node := ht.root.recursiveInsert(entry)
+
+		ht.hosts[string(entry.PublicKey.Key)] = node
 	}
 
 	return hosts, nil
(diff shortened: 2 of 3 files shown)


modules/renter/hostdb/hosttree/hosttree.go
```diff
@@ -0,0 +1,258 @@
...
+	// not exist in the tree.
+	ErrNoSuchHost = errors.New("no host with specified public key")
+)
+
+type (
+	// HostTree is a data structure that contains a weighted tree of
+	// `HostEntries`, and provides methods for inserting, retreiving, and
+	// modifing entries.
+	HostTree struct {
+		*node
```
lukechampine: probably should be named `root`

modules/renter/hostdb/hosttree/hosttree.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+	node.remove()
+	delete(ht.hosts, pk.String())
+
+	return nil
+}
+
+// Modify updates a host entry at the given public key, replacing the old entry
+// with the entry provided by `newEntry`.
+func (ht *HostTree) Modify(pk types.SiaPublicKey, newEntry *HostEntry) error {
```
lukechampine: why does this function need to take a public key, if the public key is also in `newEntry`?

modules/renter/hostdb/hosttree/hosttree.go
```diff
@@ -0,0 +1,258 @@
...
+
+	// Return the root entry.
+	return n, nil
+}
+
+// remove takes a node and removes it from the tree by climbing through the
+// list of parents. remove does not delete nodes.
+func (n *node) remove() {
+	n.weight = n.weight.Sub(n.entry.Weight)
+	n.taken = false
```
lukechampine: why are nodes not deleted, but rather marked as "not taken?" Is it an efficiency thing?

modules/renter/hostdb/hosttree/hosttree.go
```diff
@@ -0,0 +1,258 @@
...
+			count: 1,
+		},
+		hosts: make(map[string]*node),
+	}
+}
+
+// recursiveInsert inserts an entry into the appropriate place in the tree. The
+// running time of recursiveInsert is log(n) in the maximum number of elements
+// that have ever been in the tree.
+func (n *node) recursiveInsert(entry *HostEntry) (nodesAdded int, newnode *node) {
```
lukechampine: `nodesAdded` doesn't seem to be used anywhere.

modules/renter/hostdb/hosttree/hosttree.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+
+	// node is a node in the tree.
+	node struct {
+		parent *node
+		left   *node
+		right  *node
+
+		count int
+		taken bool
```
lukechampine: these could use docstrings

modules/renter/hostdb/hosttree/hosttree.go
```diff
@@ -0,0 +1,258 @@
...
+		hosts map[string]*node
+	}
+
+	// HostEntry is an entry in the host tree.
+	HostEntry struct {
+		modules.HostDBEntry
+
+		FirstSeen   types.BlockHeight
+		Weight      types.Currency
+		Reliability types.Currency
```
lukechampine: These could use docstrings. Also, does the caller need to know what the Weight of an entry is?

DavidVorick:
Host tree should have the same locking conventions as all other NebulousLabs packages - threadsafe.

It also provides more room for performance in the future. An external package can at-best lock->exportedcall->unlock, but internally it can only hold the lock as long as it needs it. Not that I expect it to be much of an issue, but this package definitely needs thread safety.

lukechampine:
You could also argue the reverse: if the caller knows that it will never use hosttree concurrently, then it doesn't need to hold any locks at all. Whereas if the hosttree is threadsafe, it will do internal locking even if it doesn't need to.

Another advantage of external locking is that you can compose the hosttree's methods atomically. e.g.
```go
mu.Lock()
old := hosttree.Fetch(1)
old.Weight = old.Weight.Add64(10)
hosttree.Modify(old)
mu.Unlock()
```
whereas if `Fetch` and `Modify` are doing internal locking, you can't guarantee atomicity. You would need to hold a separate lock anyway.

modules/renter/hostdb/hosttree/hosttree.go
johnathanhowell: This function should definitely just take an `entry`.

modules/renter/hostdb/hosttree/hosttree.go
johnathanhowell: it's used to increment `n.count` at the bottom of this function.

modules/renter/hostdb/hosttree/hosttree.go
johnathanhowell: the caller should not need to know the weight of an entry. That field should likely be unexported.

modules/renter/hostdb/hosttree/hosttree.go
lukechampine: ah lol, I was only looking at the non-recursive calls

DavidVorick:
The hostdb performs a bunch of scans in parallel and when the scans finish it interacts with the hosttree, there's inherent concurrency.

Makefile
```diff
@@ -31,7 +31,7 @@ dependencies:
 run = Test
 pkgs = ./api ./build ./compatibility ./crypto ./encoding ./modules ./modules/consensus \
        ./modules/explorer ./modules/gateway ./modules/host ./modules/host/storagemanager \
-       ./modules/renter ./modules/renter/contractor ./modules/renter/hostdb ./modules/renter/proto \
+       ./modules/renter ./modules/renter/contractor ./modules/renter/hostdb ./modules/renter/hostdb/hosttree ./modules/renter/proto \
```
DavidVorick: sticks out too far, have to re-sort

DavidVorick: it's annoying, but you'll live :)

modules/renter/hostdb/hosttree/hosttree.go
```diff
@@ -0,0 +1,274 @@
...
+	// should always have a non-nil entry, unless they have been Delete()ed.
+	ErrNilEntry = errors.New("node has a nil entry")
+
+	// ErrHostExists is returned if an Insert is called with a public key that
+	// already exists in the tree.
+	ErrHostExists = errors.New("host already exists in the tree")
+
+	// ErrNoSuchHost is returned if Remove is called with a public key that does
+	// not exist in the tree.
+	ErrNoSuchHost = errors.New("no host with specified public key")
```
DavidVorick: Usually the errors don't get exported - cleans up the godoc.

modules/renter/hostdb/hosttree/hosttree.go
```diff
@@ -0,0 +1,274 @@
...
+		mu sync.Mutex
+	}
+
+	// HostEntry is an entry in the host tree.
+	HostEntry struct {
+		modules.HostDBEntry
+		weight types.Currency
+
+		FirstSeen   types.BlockHeight
+		Reliability types.Currency
```
DavidVorick: This should be replaced by a history of scans. Luke's recent contractor PR has some of this stuff.

The 'Reliability' stuff we do in the hostdb is pretty gross and was never meant to be a long term solution.

modules/renter/hostdb/hosttree/hosttree.go
johnathanhowell: Isn't it useful to have these errors exported for the caller? They can check for the exact type of the returned error eg `if err == hosttree.ErrNoSuchHost` etc

johnathanhowell:
After our short offline discussion about this package, the following changes have been made:

- `hostEntry` is now an unexported type.
- `New()` now takes one argument, a `WeightFunc` (`func (modules.HostDBEntry) types.Currency`), which allows the caller to safely define how the `HostDBEntries` are weighted in the tree.

modules/renter/hostdb/hosttree/hosttree_test.go
```diff
@@ -160,6 +161,33 @@ func TestHostTree(t *testing.T) {
...
+		}
+	}()
+	go func() {
+		time.Sleep(time.Millisecond * 5)
+		_, err := tree.Fetch(2, nil)
+		if err != nil {
+			t.Fatal(err)
+		}
+	}()
+}
```
DavidVorick: My preferred approach:

1. have a consistency check that can easily identify inconsistencies in the state. I believe we already have one of those for the host tree.

2. spin up like 100 gothreads in a for loop using a thread group. Have the gothreads running on an infinite loop and randomly either inserting, fetching, modifying, or using one of the other exported functions. You can use a map internal to the thread so it knows if it has inserted something before (therefore can fetch, modify, or delete that thing).

3. Let it run for a few seconds

4. tg.Stop() -> run the inconsistency check.

That will pretty much tell you if it's thread safe or not. This I feel does not get as much code coverage.

modules/renter/hostdb/hosttree/hosttree.go
DavidVorick: You would think, but in practice it's just clutter.

Unless you find yourself thinking "man, it'd be really nice if the exact error were exposed", there's no reason to export your errors. There are a few places in the code where we switch based on error (e.g. in the gateway I think we handle timeouts differently from other types of errors), but it's only a few and usually you are already going out of your way to get the two packages to communicate properly.

modules/renter/hostdb/hosttree/hosttree.go
```diff
@@ -0,0 +1,290 @@
...
+	HostTree struct {
+		root *node
+
+		// hosts is a map of public keys to nodes.
+		hosts map[string]*node
+
+		// wf applies a weight to a hostEntry
+		wf WeightFunc
+
+		mu sync.Mutex
```
DavidVorick: `wf` is not an acceptable name, because it's not a common acronym. `mu` and `tg` are special cases because they are ubiquitous throughout the project - pretty much every module has both a `mu` and a `tg`. But only the host tree has a weightFunc, so it should be spelled out.

Also, probably okay to group the weightFunc and mu together.

modules/renter/hostdb/hosttree/hosttree.go
```diff
@@ -0,0 +1,290 @@
...
+)
+
+type (
+	// WeightFunc is a function used to weight a given HostDBEntry in the tree.
+	WeightFunc func(modules.HostDBEntry) types.Currency
+
+	// HostTree is a data structure that contains a weighted tree of
+	// `HostEntries`, and provides methods for inserting, retreiving, and
+	// modifing entries.
+	HostTree struct {
```
DavidVorick: For whatever reason, we standardized on using ' instead of \` in our docstrings. Also, definitely redundant to call HostTree a data structure.

Maybe "A HostTree is used to store and select host database entries. Each HostTree is initialized with a weight func that is able to assign a weight to each entry. The entries can be selected at random, weighted by the weight func."

modules/renter/hostdb/hosttree/hosttree.go
```diff
@@ -0,0 +1,290 @@
...
+
+	ht.hosts[string(entry.PublicKey.Key)] = node
+	return nil
+}
+
+// Fetch grabs a random `n` hosts from the tree. There will be no repeats, but
+// the length of the slice returned may be less than `n`, and may even be zero.
+// The hosts that are returned first have the higher priority. Hosts passed to
+// `ignore` will not be considered; pass `nil` if no blacklist is desired.
+func (ht *HostTree) Fetch(n int, ignore []types.SiaPublicKey) ([]modules.HostDBEntry, error) {
```
DavidVorick: Should probably be called `RandomSelect()`

modules/renter/hostdb/hosttree/hosttree.go
lukechampine: `weightFn` is fine, or maybe even `weight`. `ht.weight(entry)` reads ok.

modules/renter/hostdb/hosttree/hosttree.go
lukechampine: I usually leave out the `''` unless I'm referring to a variable name that isn't easy to parse. The standard library doesn't use them at all. For example:
```go
// Abs returns the absolute value of x.
func Abs(x float64) float64
```
Here it's pretty obvious what `x` is referring to, so you don't need any special marks.
```go
// Abs returns an absolute representation of path.
func Abs(path string) string
```
Here it's a little less obvious, because `path` doesn't stick out as much as `x`. Still, it's passable.

Types, especially exported ones, stand out pretty clearly so they can be typed verbatim. You just need to use your best judgment to decide whether a name will confuse the reader. For example, in docstring of `Fetch`:
```go
// Hosts passed to `ignore` will not be considered.
func (ht *HostTree) Fetch(n int, ignore []types.SiaPublicKey)
```
Here, it would be confusing if `ignore` weren't quoted somehow. So surrounding it in single quotes is recommended.

johnathanhowell:
Latest commits address review comments. Commit messages and diffs should be somewhat easy to parse.

modules/renter/hostdb/hosttree/hosttree.go
```diff
@@ -10,43 +10,44 @@ import (
...
 	HostTree struct {
 		root *node
 
 		// hosts is a map of public keys to nodes.
 		hosts map[string]*node
 
-		// wf applies a weight to a hostEntry
-		wf WeightFunc
+		// weight applies a weight to a hostEntry
+		weight WeightFunc
```
lukechampine: more accurate to say `weight calculates the weight of a hostEntry`

modules/renter/hostdb/hosttree/hosttree.go
DavidVorick: weightFn seems good. With just weight it seems like the weight of the tree, not a method used to evaluate the weight of its entries

modules/renter/hostdb/hosttree/hosttree_test.go
```diff
@@ -195,7 +195,7 @@ func TestHostTreeParallel(t *testing.T) {
 
 			for {
 				select {
-				case <-time.After(time.Second * 50):
+				case <-time.After(time.Millisecond * 50):
```
DavidVorick: lol

modules/renter/hostdb/hosttree/hosttree_test.go
```diff
@@ -0,0 +1,565 @@
...
+	var tg siasync.ThreadGroup
+	nthreads := 100
+	nelements := 0
+	var mu sync.Mutex
+	for i := 0; i < nthreads; i++ {
+		go func() {
+			tg.Add()
+			defer tg.Done()
+
+			inserted := make(map[string]modules.HostDBEntry)
```
DavidVorick: mildly unrealistic because two threads are never working on the same element in the tree at the same time. Given how the tree is implemented, I don't think this is actually an issue, the parallelism should still cause a lot of damage if the implementation is correct.

But generally you want your threads to have data conflicts as well as codepath conflicts when testing.

For this, that might be difficult to do because you'd need to wrap the map in a lock and that might be sufficient to make the threads effectively single threaded. 

It's probably fine as-is, I really don't expect the parallelism code that we added to cause issues.

modules/renter/hostdb/hosttree/hosttree_test.go
```diff
@@ -0,0 +1,565 @@
...
+					return &entry
+				}
+				return nil
+			}
+
+			for {
+				select {
+				case <-time.After(time.Millisecond * 50):
+				case <-tg.StopChan():
+					return
```
johnathanhowell: I just wanted to throttle this loop a bit so this test wouldn't be extremely stressful.

DavidVorick: it's supposed to be extremely stressful lol

PR #1514, opened 2016-12-08
Add individual contract metrics

lukechampine:
The contractor (and by extension, the renter) now returns per-contract metrics, defined by `modules.RenterContractMetrics`.

Unresolved questions:
- How do we group per-contract metrics according to their allowance period?
- How should these metrics be reported in the API?
- What should be done with the existing `RenterFinancialMetrics` type?

re: grouping: The easiest approach right now would be to use `contract.EndHeight()`. Using the start height makes more intuitive sense for the user (since it's the height at which their funds were locked away), but currently the contractor doesn't store any info about when the allowance period began. I suppose you could just subtract the period from the end height, but then what if the allowance period is changed?

### DIFF
--- a/api/renterhost_test.go
+++ b/api/renterhost_test.go
@@ -249,3 +249,111 @@ func TestIntegrationUploadDownload(t *testing.T) {
 		t.Fatal("the uploading is not succeeding for some reason:", rf.Files[0], rf.Files[1])
 	}
 }
+
+// TestIntegrationRenterMetrics tests that the renter's metrics are properly
+// reported after uploading and downloading.
+func TestIntegrationRenterMetrics(t *testing.T) {
+	if testing.Short() {
+		t.SkipNow()
+	}
+	st, err := createServerTester("TestIntegrationRenterMetrics")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer st.server.Close()
+
+	// Announce the host and start accepting contracts.
+	err = st.announceHost()
+	if err != nil {
+		t.Fatal(err)
+	}
+	err = st.acceptContracts()
+	if err != nil {
+		t.Fatal(err)
+	}
+	err = st.setHostStorage()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Renter metrics should be zero.
+	var metrics RenterGET
+	err = st.getAPI("/renter", &metrics)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !metrics.FinancialMetrics.ContractSpending.IsZero() || len(metrics.ContractMetrics) > 0 {
+		t.Error("renter should not have formed any contracts yet:", metrics.FinancialMetrics)
+	}
+
+	// Set an allowance for the renter, allowing a contract to be formed.
+	allowanceValues := url.Values{}
+	testFunds := "10000000000000000000000000000" // 10k SC
+	testPeriod := "5"
+	allowanceValues.Set("funds", testFunds)
+	allowanceValues.Set("period", testPeriod)
+	err = st.stdPostAPI("/renter", allowanceValues)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Metrics should reflect new allowance.
+	err = st.getAPI("/renter", &metrics)
+	if err != nil {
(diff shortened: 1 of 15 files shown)


DavidVorick:
That feels hacky to me, and like there would be a lot of edge cases where you end up with weird or confusing reporting.

How difficult would it be to add another persistent field? Are the contracts being stored in Json? We'd lose information on the existing contracts, but as a substitute we could subtract `period` from them the first time if the value is '0'. (I guess a sort of compatibility code).

Depending, this might be a good place to pilot a compatibility upgrade, because the change is relatively small compared to what we'll be doing for the host and renter down the line.

DavidVorick:
On the other hand, now is probably not the time to be doing persist changes.

hmm. Maybe we'll have to call it 'good enough' for now to set the start point to contract.EndHeight()-period. I don't think using just EndHeight() is sufficient.

Most people won't be changing their period. That's going to have to be good enough I think until we can get the creation date into the contract.

modules/renter/contractor/editor.go
```diff
@@ -117,12 +117,18 @@ func (he *hostEditor) Upload(data []byte) (crypto.Hash, error) {
...
-	he.contractor.financialMetrics.StorageSpending = he.contractor.financialMetrics.StorageSpending.Add(storageDelta)
-	he.contractor.contracts[contract.ID] = contract
-	he.contractor.saveSync()
-	he.contractor.mu.Unlock()
+	c := he.contractor
+	c.mu.Lock()
+	metrics := c.contractMetrics[contract.ID]
+	metrics.UploadSpending = metrics.UploadSpending.Add(uploadDelta)
+	metrics.StorageSpending = metrics.StorageSpending.Add(storageDelta)
+	metrics.Unspent = metrics.Unspent.Sub(uploadDelta.Add(storageDelta))
```
DavidVorick: I think it would be less confusing to do `metrics.Unspend.Sub(ud).Sub(sd)` -> seeing `Add(storageDelta)` is confusing when you are expecting the value to be subtracted out.

lukechampine:
Ok, added heights to the metrics and added them to the API. At this point, the UI can display metrics about each contract and group contracts according to their allowance. I decided to simply include the entire allowance in each contract metric. However, you still need the "period start height" as well in order to group the contracts according to period. It would be nice if we could include this in the `Allowance` struct somehow, but that seemed like too big of a change.

doc/API.md
```diff
@@ -638,11 +638,33 @@ returns the current settings along with metrics on the renter's spending.
...
+    "uploadspending":       "5678"  // hastings
+  },
+  "contractmetrics": [
+    {
+      "allowance": {
+        "funds":       "1234", // hastings
+        "hosts":       24,
+        "period":      6048, // blocks
+        "renewwindow": 3024  // blocks
+      },
```
DavidVorick: This should be removed, the allowance is going to get more complicated and I think it will end up polluting the contract metrics. (e.g., allowance can have host blacklists or whitelists or ip ranges or emphasis on performance, etc)

api/renterhost_test.go
```diff
@@ -249,3 +249,111 @@ func TestIntegrationUploadDownload(t *testing.T) {
 		t.Fatal("the uploading is not succeeding for some reason:", rf.Files[0], rf.Files[1])
 	}
 }
+
+// TestIntegrationRenterMetrics tests that the renter's metrics are properly
+// reported after uploading and downloading.
+func TestIntegrationRenterMetrics(t *testing.T) {
```
DavidVorick: fine for now but we've dropped the 'Integration' and 'Unit' conventions, because we don't actually use them.

sync/threadgroup_test.go
```diff
@@ -417,7 +417,7 @@ func TestThreadGroupSiaExample(t *testing.T) {
 		tg.Done()
 	}()
 	tg.Stop()
-	if !threadFinished || !listenerCleanedUp || !fileClosed {
+	if !threadFinished2 || !listenerCleanedUp || !fileClosed {
```
DavidVorick: How did you find this?

DavidVorick: nice find

PR #1516, opened 2016-12-09
add thread saftey to TransactionList() call

DavidVorick:
We have apparently missed for a long time the fact that this call is not actually thread safe.


DavidVorick:
Hmm. Trying to figure out the best way to get this into both the `lts` branch and the `master` branch without conflicts.

DavidVorick:
oh well

PR #1520, opened 2016-12-15
Save cachedRevisions properly

lukechampine:
The contractor's `cachedRevisions` map was not being properly saved. Because the fields of the `cachedRevision` type were not exported, the JSON encoder skipped over them when saving.

Testing should have caught this, but the `TestSaveLoad` test was deficient. Not only did it not check the `cachedRevisions` field, it also failed to clear the contractor's fields between saving and loading.

### DIFF
--- a/modules/renter/contractor/contractor.go
+++ b/modules/renter/contractor/contractor.go
@@ -24,8 +24,8 @@ var (
 // as a safeguard against desynchronizing with the host.
 // TODO: save a diff of the Merkle roots instead of all of them.
 type cachedRevision struct {
-	revision    types.FileContractRevision
-	merkleRoots []crypto.Hash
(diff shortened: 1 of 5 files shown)


DavidVorick:
I would like to suggest that one of the reasons the testing was not accurate is because you are creating the contractor by hand instead of using `newContractRevision` or even better the exported functions. If you are using `New` and `Close` it's much less likely that data will be leftover from the previous run, because you'll have an entirely new object.

It's also more likely to catch discontinuities, and it's also more malleable when you start adding new fields and features.

lukechampine:
indeed, fully closing and recreating the contractor would be more true-to-life. It still would not have caught the error though, since the cachedRevisions field wasn't being checked at all.

PR #1522, opened 2016-12-16
Renter period metrics

lukechampine:
I didn't update the API docs because I'm not sure this is how we want to organize things. I'm not seeing the need for a separate `FinancialMetrics` object if we have `PeriodMetrics`. We can't outright remove those fields, but we could deprecate them.

There was also a suggestion that we record the set of contract IDs in the `PeriodMetrics` to enable easy lookup.

Also, the allowance code seemed to have some bugs around setting `periodStart`.

### DIFF
--- a/modules/renter.go
+++ b/modules/renter.go
@@ -92,10 +92,35 @@ type RenterFinancialMetrics struct {
 	DownloadSpending types.Currency `json:"downloadspending"`
 	StorageSpending  types.Currency `json:"storagespending"`
 	UploadSpending   types.Currency `json:"uploadspending"`
+}
 
-	// AllowancePeriodStart is the blockheight at which the current allowance
-	// period began.
-	AllowancePeriodStart types.BlockHeight `json:"allowanceperiodstart"`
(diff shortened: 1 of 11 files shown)


DavidVorick:
best way to probe for bugs is to write tests around the cases that give you uncertainty

modules/renter.go
```diff
@@ -92,10 +92,35 @@ type RenterFinancialMetrics struct {
...
+	PeriodStart types.BlockHeight `json:"periodstart"`
+
+	// The amount of money spent on fees during the period.
+	TxnFeeSpending      types.Currency `json:"txnfeespending"`
+	ContractFeeSpending types.Currency `json:"contractfeespending"`
+	SiafundFeeSpending  types.Currency `json:"siafundfeespending"`
+
+	// The amount of money allocated for uploads, downloads, and storage
+	// during the period.
+	ContractAllocations types.Currency `json:"contractallocations"`
```
DavidVorick: Meaning, money that was locked into transactions? Comment string can probably be made a little more clear.

lukechampine:
Closing in favor of a simpler approach. See #1523.

PR #1524, opened 2016-12-17
Unspent allowance

lukechampine:
This took longer than expected because I had to track down a bug in the metrics reporting. The commit is pretty self-explanatory.

The only open question here is when to reset the metrics. I believe they should be reset as soon as we renew the first contract in a new allowance period, but how do we know when that is? We may need to keep track of `periodStart`, as before.

### DIFF
--- a/api/renter.go
+++ b/api/renter.go
@@ -32,8 +32,24 @@ var (
 type (
 	// RenterGET contains various renter metrics.
 	RenterGET struct {
-		Settings         modules.RenterSettings         `json:"settings"`
-		FinancialMetrics modules.RenterFinancialMetrics `json:"financialmetrics"`
+		Settings         modules.RenterSettings `json:"settings"`
+		FinancialMetrics RenterFinancialMetrics `json:"financialmetrics"`
+		CurrentPeriod    types.BlockHeight      `json:"currentperiod"`
+	}
+
+	// RenterFinancialMetrics contains metrics about how much the Renter has
+	// spent on storage, uploads, and downloads.
+	RenterFinancialMetrics struct {
+		// Amount of money in the allowance spent on file contracts including
+		// fees.
+		ContractSpending types.Currency `json:"contractspending"`
+
+		DownloadSpending types.Currency `json:"downloadspending"`
+		StorageSpending  types.Currency `json:"storagespending"`
+		UploadSpending   types.Currency `json:"uploadspending"`
+
+		// Amount of money in the allowance that has not been spent.
+		Unspent types.Currency `json:"unspent"`
 	}
 
 	// RenterContract represents a contract formed by the renter.
@@ -83,9 +99,32 @@ type (
 
 // renterHandlerGET handles the API call to /renter.
 func (api *API) renterHandlerGET(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	// calculate financial metrics from contracts
+	periodStart := api.renter.CurrentPeriod()
+	var fm RenterFinancialMetrics
+	for _, c := range api.renter.Contracts() {
+		if c.StartHeight < periodStart {
+			continue
+		}
+		fm.ContractSpending = fm.ContractSpending.Add(c.TotalCost)
+		fm.DownloadSpending = fm.DownloadSpending.Add(c.DownloadSpending)
+		fm.UploadSpending = fm.UploadSpending.Add(c.UploadSpending)
+		fm.StorageSpending = fm.StorageSpending.Add(c.StorageSpending)
+		fm.Unspent = fm.Unspent.Add(c.RenterFunds())
+	}
+	settings := api.renter.Settings()
+	fm.Unspent = fm.Unspent.Add(settings.Allowance.Funds)
+	if fm.Unspent.Cmp(fm.ContractSpending) < 0 {
+		// avoid negative currency error
+		fm.Unspent = types.ZeroCurrency
+	} else {
+		fm.Unspent = fm.Unspent.Sub(fm.ContractSpending)
+	}
+
 	WriteJSON(w, RenterGET{
-		Settings:         api.renter.Settings(),
-		FinancialMetrics: api.renter.FinancialMetrics(),
+		Settings:         settings,
+		FinancialMetrics: fm,
+		CurrentPeriod:    periodStart,
 	})
 }
 
--- a/api/renter.go
+++ b/api/renter.go
@@ -99,9 +99,11 @@ type (
 
 // renterHandlerGET handles the API call to /renter.
 func (api *API) renterHandlerGET(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
-	// calculate financial metrics from contracts
+	settings := api.renter.Settings()
 	periodStart := api.renter.CurrentPeriod()
+	// calculate financial metrics from contracts
 	var fm RenterFinancialMetrics
+	fm.Unspent = settings.Allowance.Funds
 	for _, c := range api.renter.Contracts() {
 		if c.StartHeight < periodStart {
 			continue
@@ -110,15 +112,11 @@ func (api *API) renterHandlerGET(w http.ResponseWriter, req *http.Request, _ htt
 		fm.DownloadSpending = fm.DownloadSpending.Add(c.DownloadSpending)
 		fm.UploadSpending = fm.UploadSpending.Add(c.UploadSpending)
 		fm.StorageSpending = fm.StorageSpending.Add(c.StorageSpending)
-		fm.Unspent = fm.Unspent.Add(c.RenterFunds())
-	}
-	settings := api.renter.Settings()
-	fm.Unspent = fm.Unspent.Add(settings.Allowance.Funds)
-	if fm.Unspent.Cmp(fm.ContractSpending) < 0 {
-		// avoid negative currency error
-		fm.Unspent = types.ZeroCurrency
-	} else {
-		fm.Unspent = fm.Unspent.Sub(fm.ContractSpending)
+		// total unspent is:
+		//    allowance - (cost to form contracts) + (money left in contracts)
+		if fm.Unspent.Add(c.RenterFunds()).Cmp(c.TotalCost) > 0 {
+			fm.Unspent = fm.Unspent.Add(c.RenterFunds()).Sub(c.TotalCost)
+		}
 	}
 
 	WriteJSON(w, RenterGET{
--- a/modules/renter.go
+++ b/modules/renter.go
@@ -79,21 +79,6 @@ type RenterSettings struct {
 	Allowance Allowance `json:"allowance"`
 }
 
-// RenterFinancialMetrics contains metrics about how much the Renter has
-// spent on storage, uploads, and downloads.
-type RenterFinancialMetrics struct {
-	// ContractSpending is how much the Renter has paid into file contracts
-	// formed with hosts. Note that some of this money may be returned to the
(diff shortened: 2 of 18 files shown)


modules/renter/contractor/contractor.go
```diff
@@ -70,6 +70,23 @@ func (c *Contractor) Allowance() modules.Allowance {
...
+	unspent := c.allowance.Funds
+	for _, contract := range c.contracts {
+		unspent = unspent.Add(contract.RenterFunds())
+	}
+	if unspent.Cmp(c.financialMetrics.ContractSpending) < 0 {
+		// avoid negative currency error
+		unspent = types.ZeroCurrency
+	} else {
+		unspent = unspent.Sub(c.financialMetrics.ContractSpending)
+	}
```
DavidVorick: :+1:

modules/renter/contractor/renew.go
```diff
@@ -76,6 +84,13 @@ func (c *Contractor) managedRenewContracts() error {
 
 	c.log.Printf("renewing %v contracts", len(renewSet))
 
+	// If this is the first time we are renewing contracts in a new allowance
+	// period, reset the spending metrics.
+	// TODO: how do we determine this?
+	if true {
```
DavidVorick: Ideally you don't have to figure this out, whatever function is figuring out how much is unspent will just add up all the costs of all the contracts in the current period. Which would mean that from a contract you need to be able to determine:

- how much it cost
- what period it belongs in (based on the block it was created in)

----------

If that approach ends up being too big of a change, you'd know you are in a new period by having some way to check what period you are currently in. If (currentPeriod.StartHeight+periodLength < currentHeight) then you are moving to a new period. You need to reset the tracking and set the currentPeriod.StartHeight = currentHeight.

DavidVorick: periodLength = allowancePeriod-renewWindow

DavidVorick:
I think the overall best thing to do is to track metrics per-contract, and then have some high level method of figuring out which period each contract belongs in and aggregating them from that. The amount of computation required to aggregate 100s of contracts is completely negligible unless you need to recalculate it multiple times per second. This makes errors less likely because you don't have state which can become inconsistent.

Seems like that might be out of reach for the time being. In that case you'd want multiple checks for whether it's time to reset the period, I think including a check at startup. A check at startup means that a user can resolve inconsistent state by restarting Sia. Not ideal, but important if we want to make sure that users are always able to renew their files.

lukechampine:
>I think the overall best thing to do is to track metrics per-contract, and then have some high level method of figuring out which period each contract belongs in and aggregating them from that.

Agreed. Grouping by period is the only tricky thing about it. I'm leaning towards the approach where we don't explicitly store any period-level metrics, but instead have a `Period` struct containing the start height and a list of contract IDs. From that, it's easy to compute aggregate metrics.

Another thought just occurred to me: Instead of figuring out the period start and summing forward from there, why not simply sum backwards until we hit a contract formed over 1 period ago? After all, the purpose of the allowance is to cap the amount of spending within *any* period-length window.

DavidVorick:
No, it's not to cap the amount of spending within any period-length window, because that's not how people are used to billing cycles operating.

Imagine having a budget of $100 per month which resets on the 1st of every month. You can spend $80 on the 28th of one month and $90 on the 2nd of the next month and not break your budgeting, though there is a 1 week window where you spent at least $170.

I believe people will be expecting it to work that way for Sia as well.

lukechampine:
Ok, this PR has been totally revised (again). Metrics are now stored on a per-contract basis in `modules.RenterContract`. As a result, metrics collection has been greatly simplified, since the `proto` package operates on this type directly. (Previously, we were threading metrics from `proto` up to `contractor`.)

In order to aggregate the contract metrics, the contractor gained a field (and method) called `CurrentPeriod`, which holds the start height of the current period. To calculate metrics for the current period, simply iterate through all the contracts and only count the ones whose `StartHeight` is above `CurrentPeriod`.

The only tricky part of this approach is compatibility. When users load the new contractor for the first time, their `CurrentPeriod` will be `0` and the metrics of their contracts will be empty. I addressed `CurrentPeriod` by subtracting the allowance period from the highest contract `EndHeight`. We haven't yet decided how to handle the empty contract metrics. Note, however, that since their `StartHeight` will be `0`, they won't be counted in the aggregate period metrics. This is probably a good thing.

api/renter.go
```diff
@@ -32,8 +32,24 @@ var (
...
+		// Amount of money in the allowance spent on file contracts including
+		// fees.
+		ContractSpending types.Currency `json:"contractspending"`
+
+		DownloadSpending types.Currency `json:"downloadspending"`
+		StorageSpending  types.Currency `json:"storagespending"`
+		UploadSpending   types.Currency `json:"uploadspending"`
+
+		// Amount of money in the allowance that has not been spent.
+		Unspent types.Currency `json:"unspent"`
```
DavidVorick: This is a good place for this field.

api/renter.go
```diff
@@ -83,9 +99,32 @@ type (
...
+		fm.Unspent = fm.Unspent.Add(c.RenterFunds())
+	}
+	settings := api.renter.Settings()
+	fm.Unspent = fm.Unspent.Add(settings.Allowance.Funds)
+	if fm.Unspent.Cmp(fm.ContractSpending) < 0 {
+		// avoid negative currency error
+		fm.Unspent = types.ZeroCurrency
+	} else {
+		fm.Unspent = fm.Unspent.Sub(fm.ContractSpending)
+	}
```
DavidVorick: This does not look correct to me, I do not think that it accurately reports funds which have not ever been allocated into contracts.

DavidVorick: oh wait I see.

you add the unspent of each contract, then you add all of the funds of the allowance, then finally you subtract out the total spent on contracts.

Ok, that does seem correct. But I think it would help if there were more comments explaining that.

modules/renter.go
```diff
@@ -157,6 +152,10 @@ type Renter interface {
 	// Contracts returns the contracts formed by the renter.
 	Contracts() []RenterContract
 
+	// CurrentPeriod returns the height at which the current allowance period
+	// began.
+	CurrentPeriod() types.BlockHeight
```
DavidVorick: A name like `CurrentPeriod()` makes me expect a range, but I guess this is okay.

modules/renter/contractor/downloader.go
```diff
@@ -72,34 +71,19 @@ func (hd *hostDownloader) Sector(root crypto.Hash) ([]byte, error) {
...
 	return sector, nil
 }
 
-// Speed returns the most recent download speed of this host, in bytes per
-// second.
-func (hd *hostDownloader) Speed() uint64 {
-	hd.mu.Lock()
-	defer hd.mu.Unlock()
-	return hd.speed
-}
```
DavidVorick: I presume we are only taking this out for the current PR?

Next PR should have reporting for the hosts, at least enough that we can tell when a host is complete garbage and needs to be replaced.

api/renter.go
lukechampine: Might be more obvious to write it like this:
```go
fm.Unspent = settings.Allowance.Funds
for _, c := range contracts {
    // subtract the total amount we spent on the contract, but add back
    // the amount in the contract that is still spendable
    fm.Unspent = fm.Unspent.Sub(c.TotalCost).Add(c.RenterFunds())
}
```
(of course, you still need a negative check for the `Sub`)

modules/renter/contractor/downloader.go
lukechampine: yeah I removed it because it wasn't used anywhere. The code is simple enough that to restore afterwards. It's also wrong: it calculates speed using `time / sectorsize` instead of the other way around.

modules/renter/contractor/persist.go
```diff
@@ -53,7 +53,18 @@ func (c *Contractor) load() error {
 	for _, contract := range data.Contracts {
 		c.contracts[contract.ID] = contract
 	}
-	c.financialMetrics = data.FinancialMetrics
+	c.currentPeriod = data.CurrentPeriod
+	if c.currentPeriod == 0 {
+		// If loading old persist, current period will be unknown. Best we can
+		// do is guess based on contracts + allowance.
```
DavidVorick: hmm. This works, but it's dangerous code because it doesn't define itself as compatibility code.

/ideally/ (maybe this is out of scope) we'd define a new version number in the file and then create a function in the compatibility package that can upgrade this file to having a period.

Minimally we need to mark that this code is here for compatibility reasons, and given that it's `COMPATv1.0.4-lts` we won't be able to pull it out ever - only we'll be able to throw it in the compat package.

Which... I think is fine. As long as there is a comment clearly outlining the potential problem and the solution, we should have an easy enough time moving it to the compatibility package later.

modules/renter/contractor/update.go
```diff
@@ -35,6 +35,11 @@ func (c *Contractor) ProcessConsensusChange(cc modules.ConsensusChange) {
 		c.log.Debugln("INFO: deleted expired contract", id)
 	}
 
+	// if we have entered the next period, update currentPeriod
+	if c.blockHeight > c.currentPeriod+c.allowance.Period {
+		c.currentPeriod += c.allowance.Period
+	}
```
DavidVorick:
```go
cycleLen := c.allowance.Period - c.allowance.RenewWindow
if c.blockHeight > c.currentPeriod+cycleLen {
    c.currentPeriod += cycleLen
}
```

api/renter.go
DavidVorick: I think that would be more obvious.

modules/renter/contractor/persist.go
lukechampine: IIRC our current versioning scheme is to only bump the persist version when the data is no longer convertible. The persist package takes the simple approach of checking for basic version equality, rather than a more nuanced comparison (e.g `>=v1.0.4`).

When it comes time to write the compatibility code, we should probably switch to using `build.Version` instead of individual persist versions. Then the compat code can define ranges of versions that are compatible.

modules/renter/contractor/allowance.go
```diff
@@ -147,12 +148,10 @@ func (c *Contractor) SetAllowance(a modules.Allowance) error {
 	c.mu.Lock()
 	c.allowance = a
 	c.contracts = newContracts
-	// update metrics
-	var spending types.Currency
-	for _, contract := range c.contracts {
-		spending = spending.Add(contract.RenterFunds())
+	// if the currentPeriod was previously unset, set it now
+	if c.currentPeriod == 0 {
+		c.currentPeriod = periodStart
```
DavidVorick: should it only be updating to a new period if there isn't currently a period? Should it update any time that the renewWindow changes?

hmm. I guess that's probably too much detail/stress for this PR. We'll leave it for now but I'm thinking that we'll be coming back to it in the future.

lukechampine: not sure. Maybe it should only reset if you replace all your contracts. The allowance has a lot of edge cases.

DavidVorick: The period should only reset if the user is expecting it to reset, which I don't think is directly related to how many contracts are being replaced. I'm trying to imagine what a sane outcome would be if I extended the period. Am I expecting to enter a new period, or am I expecting the current period to roll over?

I'm sort of thinking that the user would not be expecting to enter a new period.

And same if the period was shortened. If you bring it from 30 days down to 7, you probably want it to start from the previous start, and then walk forward by 7 until the current period is hit. Or maybe you do want it to reset completely?

Not sure, but I think it's okay to punt it for now.

PR #1525, opened 2016-12-20
migrate `hostdb` to the from `weightedlist` to the new `hosttree` package

johnathanhowell:
This PR removes `weightedlist` from the `hostdb` and replaces it with the new `hosttree` package. Hosts are still stored in the `activeHosts` map by the NetAddress, since changing fully to public keys requires changing the `HostDB` interface, which is a much larger change. A subsequent PR will be opened, likely after the renter pools branch is merged, which removes `activeHosts` and changes the HostDB to use public keys.


johnathanhowell:
Some contractor testing had to change because of these changes, since the `hostTree` stores hosts by publicKey it is no longer sufficient to simply re-announce the same host on a different netaddress. I've changed `newTestingTrio` to take a number of hosts to create, `nhosts`, and return a slice of hosts (with unique keypairs, obviously.)

modules/renter.go
```diff
@@ -141,6 +141,8 @@ type HostDBEntry struct {
 	// ScanHistory is the set of scans performed on the host. It should always
 	// be ordered according to the scan's Timestamp, oldest to newest.
 	ScanHistory HostDBScans
```
DavidVorick: I know that this is not from this PR, but I think here is bad form, should just be []hdbs

The hdbss type is purely for sorting.

modules/renter/contractor/host_integration_test.go
```diff
@@ -104,9 +105,9 @@ func newTestingContractor(testdir string, cs modules.ConsensusSet, tp modules.Tr
 	return New(cs, w, tp, hdb, filepath.Join(testdir, "contractor"))
 }
 
-// newTestingTrio creates a Host, Contractor, and TestMiner that can be used
-// for testing host/renter interactions.
-func newTestingTrio(name string) (modules.Host, *Contractor, modules.TestMiner, error) {
+// newTestingTrio a Contractor, TestMiner, and a slice of hosts that can be
+// used for testing host/renter interactions.
+func newTestingTrio(name string, nhosts int) ([]modules.Host, *Contractor, modules.TestMiner, error) {
```
DavidVorick: @lukechampine what are your thoughts on this change?

modules/renter.go
lukechampine: yeah, my bad. Typically the pattern is to name the sorting type `ByX`. So we'd write:
```go
sort.Sort(modules.ByTimestamp(entry.ScanHistory))
```
The name becomes pretty ugly though, since `ByTimestamp` is ambiguous -- it ought to be `HostDBScansByTimestamp` or something.

lukechampine: btw, in Go 1.8 we can do:
```go
sort.Slice(entry.ScanHistory, func(i, j int) bool {
    return entry.ScanHistory[i].Timestamp.Before(entry.ScanHistory[j].Timestamp)
})
```

modules/renter/contractor/host_integration_test.go
lukechampine: I don't really like it; it's not a trio if there's multiple hosts. AFAICT, the only place we create multiple hosts is `TestIntegrationSetAllowance`, where we create 2 instead of 1. So I would prefer to revert the change, and instead either create the extra host manually or add a new helper function for creating multiple hosts.

johnathanhowell: I agree that this change is ugly, I'll revert it and create an extra host in `TestIntegrationSetAllowance`.

modules/renter.go
DavidVorick: yeah that's what we should probably end up doing.

modules/renter/hostdb/hostentry_test.go
```diff
@@ -88,26 +114,17 @@ func TestAverageContractPrice(t *testing.T) {
...
-	h2.NetAddress = "bar"
+	h2 := makeHostDBEntry()
 	h2.ContractPrice = types.NewCurrency64(300)
-	h2.Weight = baseWeight
-	h2.AcceptingContracts = true
-	hdb.insertNode(h2)
-	if len(hdb.activeHosts) != 2 {
-		t.Error("host was not added:", hdb.activeHosts)
-	}
+	hdb.hostTree.Insert(h2)
```
DavidVorick: ^_^ very nice

modules/renter/hostdb/hostweight.go
```diff
@@ -31,113 +33,115 @@ var (
...
-	//
-	// In the future, the renter should be able to track average user behavior
-	// and adjust accordingly. This flexibility will be added later.
-	adjustedContractPrice := entry.ContractPrice.Div64(6048).Div64(10e9) // Adjust contract price to match 10GB for 6 weeks.
-	adjustedUploadPrice := entry.UploadBandwidthPrice.Div64(24192)       // Adjust upload price to match a single upload over 24 weeks.
-	adjustedDownloadPrice := entry.DownloadBandwidthPrice.Div64(12096)   // Adjust download price to match one download over 12 weeks.
-	siafundFee := adjustedContractPrice.Add(adjustedUploadPrice).Add(adjustedDownloadPrice).Add(entry.Collateral).MulTax()
-	totalPrice := entry.StoragePrice.Add(adjustedContractPrice).Add(adjustedUploadPrice).Add(adjustedDownloadPrice).Add(siafundFee)
+func (hdb *HostDB) calculateHostWeight() hosttree.WeightFunc {
+	return func(entry modules.HostDBEntry) types.Currency {
```
DavidVorick: I don't like the tabbing here, would prefer to see the weight func defined separately, then `return weightFunc` or whatever. I don't think it makes sense to be inline here, it's a massive amount of code.

modules/renter/hostdb/hostweight.go
```diff
@@ -31,113 +33,115 @@ var (
...
+			}
+			if age < 1000 {
+				weight = weight.Div64(4) // 64x total
+			}
+			if age < 288 {
+				weight = weight.Div64(10) // 640x total
+			}
+		} else {
+			// Shouldn't happen, but the usecase is covered anyway.
+			weight = weight.Div64(1000) // Because something weird is happening, don't trust this host very much.
```
DavidVorick: Should probably be 10e3, not 1e3. Should definitely use e notation since it's got >= three zeroes

modules/renter/hostdb/hostweight.go
```diff
@@ -31,113 +33,115 @@ var (
...
-		if age < 288 {
-			weight = weight.Div64(10) // 640x total
+		if entry.RemainingStorage < 10*requiredStorage {
+			weight = weight.Div64(6) // 570x total penalty
+		}
+		if entry.RemainingStorage < 5*requiredStorage {
+			weight = weight.Div64(10) // 5,700x total penalty
+		}
+		if entry.RemainingStorage < requiredStorage {
+			weight = weight.Div64(100) // 570,000 total penalty
```
DavidVorick: missing an x lol

modules/renter/hostdb/hostdb.go
```diff
@@ -107,11 +108,13 @@ func newHostDB(cs consensusSet, d dialer, s sleeper, p persister, l *persist.Log
 		log:     l,
 
 		// TODO: should index by pubkey, not ip
-		activeHosts: make(map[modules.NetAddress]*hostNode),
+		activeHosts: make(map[modules.NetAddress]*hostEntry),
 		allHosts:    make(map[modules.NetAddress]*hostEntry),
 		scanPool:    make(chan *hostEntry, scanPoolSize),
 	}
 
+	hdb.hostTree = hosttree.New(hdb.calculateHostWeight())
```
DavidVorick: shouldn't you be able to just specify `hdb.calculateHostWeight`, instead of having a function that returns a function? You can just define the function and then specify it here.

modules/renter/hostdb/persist.go
```diff
@@ -17,8 +17,8 @@ func (hdb *HostDB) persistData() hdbPersist {
 	for _, entry := range hdb.allHosts {
 		data.AllHosts = append(data.AllHosts, *entry)
 	}
-	for _, node := range hdb.activeHosts {
-		data.ActiveHosts = append(data.ActiveHosts, *node.hostEntry)
+	for _, entry := range hdb.activeHosts {
+		data.ActiveHosts = append(data.ActiveHosts, *entry)
```
DavidVorick: just want to confirm that this saves everything important.

Can you verify that there is a test for saving and loading hosts, including things like the scan history?

modules/renter/hostdb/hostdb.go
johnathanhowell: The reason for this is that the host weight function needs access to the hostdb's state.

lukechampine: :+1:

modules/renter/hostdb/scan.go
```diff
@@ -107,10 +107,10 @@ func (hdb *HostDB) decrementReliability(addr modules.NetAddress, penalty types.C
 
 	// If the entry is in the active database, remove it from the active
 	// database.
-	node, exists := hdb.activeHosts[addr]
+	existingEntry, exists := hdb.activeHosts[addr]
```
DavidVorick: so the entries still have the reliability cruft, instead of using the scan data?

modules/renter/hostdb/persist.go
johnathanhowell: yes, I've confirmed that `TestSaveLoad` fails after modifying these lines.

modules/renter/hostdb/scan.go
johnathanhowell: yeah, ideally `activeHosts` would not exist but that requires a much larger change.

modules/renter/hostdb/hostdb.go
lukechampine: @DavidVorick is right -- you can just pass `hdb.calculateHostWeight` if you modify it to satisfy the `weightFunc` type. It should be:

```go
func (hdb *HostDB) calculateHostWeight(entry modules.HostDBEntry) types.Currency {
    // use hdb.blockheight internally
}
```

johnathanhowell: ah yeah, that's a lot cleaner.

PR #1526, opened 2016-12-21
Archive contracts instead of deleting them

lukechampine:
Contracts are now moved to a new contractor map, `oldContracts`, instead of being deleted. This map is not used anywhere, but it may be used in the future (e.g. for calculating historic metrics).

An alternative would be to store _all_ contracts (including current ones) in this new map. I haven't thought through what the tradeoffs of that approach are.

Also, a new "secret" method has been added to the contractor, `AllContracts`. This method returns all the contracts in the current period, whereas `Contracts` filters out offline contracts. It's "secret" because it's not actually part of the contractor (or renter) interface; the API uses a type assertion to access it. This is because the method isn't intended to stick around.

I am much more in favor of moving the offline filtering into the renter, and having `Contracts` return the offline contracts. Since determining the offline status of a host doesn't require any unexported contractor functionality, the renter could define its own `onlineContracts` helper function and call it wherever we call `r.hostContractor.Contracts`, which is only 3 places. The only reason not to do this is to avoid code duplication between the renter and contractor. (The contract still needs to filter out offline contracts so that it can replace them.) We could, however, move this functionality to the `modules` package.

### DIFF
--- a/api/renter.go
+++ b/api/renter.go
@@ -136,10 +136,14 @@ type (
 func (api *API) renterHandlerGET(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
 	settings := api.renter.Settings()
 	periodStart := api.renter.CurrentPeriod()
-	// calculate financial metrics from contracts
+	// calculate financial metrics from contracts. We use the special
+	// AllContracts method to include contracts that are offline.
 	var fm RenterFinancialMetrics
 	fm.Unspent = settings.Allowance.Funds
-	for _, c := range api.renter.Contracts() {
+	contracts := api.renter.(interface {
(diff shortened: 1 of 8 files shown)


DavidVorick:
In all likelihood the renter and contractor will end up with different definitions of offline anyway. The renter more cares about the past 3 hours, while the contractor doesn't consider something offline until we haven't seen it for a week.

Ultimately it makes sense to do the offline filtering separately and have one function returning everything. I am just eager to have something release-ready.

DavidVorick:
There is no testing here.

Hmm. Might actually take a lot to verify that the historic contracts are surviving a renew cycle.

PR #1528, opened 2016-12-21
adjust constants given new capabilities

DavidVorick:
The renter is a lot more stable now, and has the ability to replace offline
hosts. This means we can safely reduce the redundacny from 6x to 4x, which will
help upload times. We have also increased the total number of hosts in use from
24 to 32, which will increase overall reliability relative to the redundancy.
Filanny, we have increased the default number of hosts to 42, which is 10 more
hosts than are needed to get a chunk to full redundancy. This means that the
renter can largely tolerate a lot of failing hosts without the upload being
impacted. It also means a greater percentage of hosts on the network will be
getting data, which will help with host encouragement.

### DIFF
--- a/api/renter.go
+++ b/api/renter.go
@@ -22,7 +22,7 @@ var (
 	// contracts with if the value is not specified explicity in the call to
 	// SetSettings.
 	recommendedHosts = build.Select(build.Var{
-		Standard: uint64(30),
+		Standard: uint64(42),
 		Dev:      uint64(2),
 		Testing:  uint64(1),
 	}).(uint64)
@@ -31,7 +31,7 @@ var (
 	// the renter settings for the renter settings to be valid. This minimum is
 	// there to prevent users from shooting themselves in the foot.
 	requiredHosts = build.Select(build.Var{
-		Standard: uint64(24),
+		Standard: uint64(20),
 		Dev:      uint64(1),
 		Testing:  uint64(1),
 	}).(uint64)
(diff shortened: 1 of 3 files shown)


api/renter.go
```diff
@@ -22,7 +22,7 @@ var (
 	// contracts with if the value is not specified explicity in the call to
 	// SetSettings.
 	recommendedHosts = build.Select(build.Var{
-		Standard: uint64(30),
+		Standard: uint64(42),
```
DavidVorick: 10 more than the default number of hosts used when uploading a file. That means much better upload speeds all the way through to the end, also means more hosts getting contracts + files overall.

api/renter.go
```diff
@@ -31,7 +31,7 @@ var (
 	// the renter settings for the renter settings to be valid. This minimum is
 	// there to prevent users from shooting themselves in the foot.
 	requiredHosts = build.Select(build.Var{
-		Standard: uint64(24),
+		Standard: uint64(20),
```
DavidVorick: people tinkering around with their own constants are probably okay with just 20 hosts.

modules/renter/repair.go
```diff
@@ -17,7 +17,7 @@ import (
 
 // TODO: Move to a consts file.
 const uploadFailureCooldown = time.Hour
-const minPiecesRepair = 6
+const minPiecesRepair = 5
```
DavidVorick: This optimizes for network performance slightly at the cost of encryption overhead. But encryption overhead is not very high at the moment, network performance still seems to be the bottleneck on the tail end of uploading.

PR #1531, opened 2017-01-03
Update r.tracking in RenameFile

lukechampine:
Previously, renaming a file would cause that file to stop uploading.

As discussed OOB, going forward, all file-related info should be stored in one place in the renter. The "right thing to do" in this case is probably to move the `RepairPath` field to the `file` type, rather than maintaining a separate map in the renter.

### DIFF
--- a/api/renter_test.go
+++ b/api/renter_test.go
@@ -7,6 +7,7 @@ import (
 	"strconv"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/crypto"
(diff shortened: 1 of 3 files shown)


DavidVorick:
Need an identical PR into LTS. Leaving this open until that is created.

DavidVorick:
This needs testing, preferably in the api

lukechampine:
The core behavior change was already tested, but I added further testing in the API package to confirm that uploads and downloads continue to work after renaming.

DavidVorick:
Testing needs to be added for all bugfixes, to verify that the bug is properly fixed and also that it says fixed in the future.

PR #1533, opened 2017-01-04
Support allowance canceling

lukechampine:
The allowance can now be canceled via `SetAllowance(Allowance{})`. Doing so will cause all active contracts to be immediately archived. Archived contracts cannot be used to create Editors or Downloaders, and they are not renewed. And since the new value of `allowance.Hosts` is 0, the contractor will not attempt to form any new contracts.

Supporting this change in the API required relaxing the restrictions on `requiredHosts` and `requiredRenewWindow`.

An important consideration here is that `SetAllowance(Allowance{})` must invalidate all active editors and downloaders before archiving contracts and returning. Otherwise, an editor could save its updated contract to `c.contracts` after the previous version had been archived.

The test could be more comprehensive. It might be good enough to monitor the `Unspent` metric.

### DIFF
--- a/api/renterhost_test.go
+++ b/api/renterhost_test.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"net/url"
 	"path/filepath"
+	"strings"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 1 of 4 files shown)


api/renterhost_test.go
```diff
@@ -607,3 +608,84 @@ func TestRenterUploadDownload(t *testing.T) {
...
+	allowanceValues.Set("funds", testFunds)
+	allowanceValues.Set("period", testPeriod)
+	err = st.stdPostAPI("/renter", allowanceValues)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Create a file.
+	path := filepath.Join(st.dir, "test.dat")
+	err = createRandFile(path, 1024)
```
DavidVorick: needs to be a lot bigger if you want to check that uploads are being interrupted I think.

DavidVorick: oh seems that you don't check it was interrupted

DavidVorick:
Guess we need to try this in the UI now

PR #1535, opened 2017-01-10
add TestRenterUploadDelete

lukechampine:


api/renterhost_test.go
```diff
@@ -607,3 +607,98 @@ func TestRenterUploadDownload(t *testing.T) {
...
+	allowanceValues.Set("funds", testFunds)
+	allowanceValues.Set("period", testPeriod)
+	err = st.stdPostAPI("/renter", allowanceValues)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Create a file.
+	path := filepath.Join(st.dir, "test.dat")
+	err = createRandFile(path, 1024)
```
DavidVorick: this file needs to be a lot bigger, I don't think it's at risk of race condition when it is this small.

Maybe do 10 MB.

api/renterhost_test.go
```diff
@@ -607,3 +607,98 @@ func TestRenterUploadDownload(t *testing.T) {
...
+		time.Sleep(100 * time.Millisecond)
+	}
+	if len(rf.Files) != 1 {
+		t.Fatal("file is not being registered:", rf.Files)
+	}
+
+	// In parallel, upload another file and delete the first file.
+	path2 := filepath.Join(st.dir, "test2.dat")
+	test2Size := modules.SectorSize*2 + 1
+	err = createRandFile(path2, int(test2Size))
```
DavidVorick: should create both rand files before any uploading starts, as this will take longer than anything else.

Issue #1536, opened 2017-01-10
don't allow allowance to be set while still syncing


DavidVorick:
If you aren't caught up, you'll spend too much money and your first contracts won't last very long, or you will just get errors

johnathanhowell:
fixed in #1537

PR #1537, opened 2017-01-10
require consensusset to be synced to set allowance

johnathanhowell:
this PR adds another check to the contractor's `SetAllowance` method which checks of the consensus set is synced before attempting to set an allowance.

Addresses #1536


DavidVorick:
Is there an easy way to test this?

lukechampine:
could be done with a mocked consensus set.
Interestingly, the fact that none of the integration tests failed means that the contractor is synced in all of them.

DavidVorick:
I think we are pretty careful about that actually

PR #1540, opened 2017-01-11
wallet defragmentation

johnathanhowell:
This PR introduces wallet defragmentation, in order to solve the problems noticed by some users, notably Poloniex, where wallets with many outputs would fail to send coins with `transaction too large`. This PR adds `defragWallet()`, which is called after every block and sends the sum of the 15 largest outputs to itself.

### DIFF
--- a/modules/wallet/defrag.go
+++ b/modules/wallet/defrag.go
@@ -41,7 +41,7 @@ func (w *Wallet) defragWallet() {
 		totalOutputValue = totalOutputValue.Add(output.Value)
 	}
 
-	w.log.Printf("defragmenting wallet: %v outputs, %v total value\n", len(defragOutputs), totalOutputValue)
+	w.log.Printf("performing wallet defragmentation transaction: %v outputs, %v total value\n", len(defragOutputs), totalOutputValue)
 
 	// grab a new address from the wallet
 	addr, err := w.nextPrimarySeedAddress()
@@ -55,7 +55,8 @@ func (w *Wallet) defragWallet() {
 	go func() {
 		txns, err := w.SendSiacoins(totalOutputValue, addr.UnlockHash())
 		if err != nil {
-			w.log.Println("error after call to SendSiacoins: ", err)
+			w.log.Printf("error sending defrag transaction with value %v: %v\n ", totalOutputValue, err)
+			return
 		}
 
 		w.log.Println("defrag transaction size: ", len(encoding.Marshal(txns)))
--- a/modules/wallet/defrag.go
+++ b/modules/wallet/defrag.go
@@ -41,12 +41,10 @@ func (w *Wallet) defragWallet() {
 		totalOutputValue = totalOutputValue.Add(output.Value)
 	}
 
-	w.log.Printf("performing wallet defragmentation transaction: %v outputs, %v total value\n", len(defragOutputs), totalOutputValue)
-
 	// grab a new address from the wallet
 	addr, err := w.nextPrimarySeedAddress()
 	if err != nil {
-		w.log.Println("error getting an address for defrag: ", err)
+		w.log.Println("Error getting an address for defragmentation: ", err)
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


modules/wallet/defrag.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	var so sortedOutputs
+	for scoid, sco := range w.siacoinOutputs {
+		so.ids = append(so.ids, scoid)
+		so.outputs = append(so.outputs, sco)
+	}
+	sort.Sort(sort.Reverse(so))
+
+	defragOutputs := so.outputs[:15]
```
DavidVorick: 15 should be constant

modules/wallet/defrag.go
```diff
@@ -0,0 +1,51 @@
...
+// defragWallet computes the sum of the 15 largest outputs in the wallet and
+// sends that sum to itself, effectively defragmenting the wallet. This defrag
+// operation is only performed if the wallet has greater than defragThreshold
+// outputs.
+func (w *Wallet) defragWallet() {
+	if len(w.siacoinOutputs) < defragThreshold {
+		return
+	}
+
+	var so sortedOutputs
```
DavidVorick: comments

modules/wallet/defrag.go
DavidVorick: also comments

modules/wallet/defrag.go
```diff
@@ -0,0 +1,51 @@
...
+
+	w.log.Printf("defragmenting wallet: %v outputs, %v total value\n", len(defragOutputs), totalOutputValue)
+
+	var dest types.UnlockHash
+	for h := range w.keys {
+		dest = h
+		break
+	}
+
+	go func() {
```
DavidVorick: comments. Why is this in a gofunc?

modules/wallet/defrag.go
```diff
@@ -0,0 +1,51 @@
...
+	// defragThreshold is the number of outputs a wallet is allowed before it is
+	// defragmented.
+	defragThreshold = 20
+)
+
+// defragWallet computes the sum of the 15 largest outputs in the wallet and
+// sends that sum to itself, effectively defragmenting the wallet. This defrag
+// operation is only performed if the wallet has greater than defragThreshold
+// outputs.
+func (w *Wallet) defragWallet() {
```
DavidVorick: There should probably be some check that the previous defrag operation succeeded / got into a block before making another one.

lukechampine: In the contractor I use a `TryMutex` to accomplish this: https://github.com/NebulousLabs/Sia/blob/master/modules/renter/contractor/update.go#L66
Not sure if it's the best pattern but it seems to work.

johnathanhowell: I think using a trylock would only prevent multiple defrags from running concurrently, it wouldn't tell you if the previous defrag transaction was accepted into the block

DavidVorick: the error is a graceful/benign one, so we decided check not needed

modules/wallet/defrag.go
```diff
@@ -0,0 +1,64 @@
...
+		w.log.Println("error getting an address for defrag: ", err)
+		return
+	}
+
+	// send the sum of the outputs to this wallet. This operation is done in a
+	// goroutine since defragWallet() is called under lock.
+	go func() {
+		txns, err := w.SendSiacoins(totalOutputValue, addr.UnlockHash())
+		if err != nil {
+			w.log.Printf("error sending defrag transaction with value %v: %v\n ", totalOutputValue, err)
```
DavidVorick: These log statements should have full punctuation, they are being displayed to end users!

DavidVorick: "Attempted to defragement wallet but failed. Defragment size: %v. Error: %v.", totalOutputValue, err

modules/wallet/defrag.go
```diff
@@ -0,0 +1,64 @@
...
+	sort.Sort(sort.Reverse(so))
+
+	// choose the defragBatchSizeth largest outputs from the wallet and sum them
+	defragOutputs := so.outputs[:defragBatchSize]
+	var totalOutputValue types.Currency
+	for _, output := range defragOutputs {
+		totalOutputValue = totalOutputValue.Add(output.Value)
+	}
+
+	w.log.Printf("performing wallet defragmentation transaction: %v outputs, %v total value\n", len(defragOutputs), totalOutputValue)
```
DavidVorick: Needs full punctuation

DavidVorick: But, really this log should be merged with the one below, no need to make two logging statements per defrag, just one is fine (one if success, a different one for each type of error (I think only 1 type possible right now))

johnathanhowell:
ok, I've made the defrag only print one log per cycle, and made the logs more human-readable and informative as requested.

PR #1541, opened 2017-01-11
only build.critical for json encoding errors

johnathanhowell:
Previously, `WriteJSON` and `WriteError` would cause a build.critical if the remote caller hung up the connection.


DavidVorick:
Iguess not easy to make a regtest here

PR #1542, opened 2017-01-11
Fix RenterFunds panic

lukechampine:
Archived ("old") contracts were being incorrectly loaded as active contracts in the contractor persist. This, in combination with the special metrics contract, is probably what caused `RenterFunds` to panic. Regardless, additional safety has been added to `RenterFunds` as well.

### DIFF
--- a/modules/renter.go
+++ b/modules/renter.go
@@ -4,6 +4,7 @@ import (
 	"io"
 	"time"
 
+	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/types"
 )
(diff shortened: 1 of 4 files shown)


DavidVorick:
Would be nice to have a reg test demonstrating the problem

modules/renter.go
```diff
@@ -135,6 +136,10 @@ func (rc *RenterContract) EndHeight() types.BlockHeight {
 // RenterFunds returns the funds remaining in the contract's Renter payout as
 // of the most recent revision.
 func (rc *RenterContract) RenterFunds() types.Currency {
+	if len(rc.LastRevision.NewValidProofOutputs) < 2 {
+		build.Critical("malformed RenterContract:", rc)
+		return types.ZeroCurrency
+	}
```
DavidVorick: :+1:

DavidVorick:
otherwise LGTM

lukechampine:
done

PR #1543, opened 2017-01-11
ignore dust outputs in defrag, only defrag if the wallet is unlocked and the consensus set is synced

johnathanhowell:


DavidVorick:
Needs tests.

Otherwise LGTM

johnathanhowell:
done

modules/wallet/defrag.go
```diff
@@ -21,14 +21,22 @@ const (
 // operation is only performed if the wallet has greater than defragThreshold
 // outputs.
 func (w *Wallet) defragWallet() {
-	if len(w.siacoinOutputs) < defragThreshold {
+	// accumulate a map of non-dust outputs
+	nonDustOutputs := make(map[types.SiacoinOutputID]types.SiacoinOutput)
+	for id, output := range w.siacoinOutputs {
+		if output.Value.Cmp(types.SiacoinPrecision) > 0 {
```
DavidVorick: This should be turned into a constant and put in constants.go

johnathanhowell:
I added the dustValue const, but there's a few more changes that I'll be doing to the defrag algorithm as discussed OOB with @DavidVorick. Don't merge until those are completed.

modules/wallet/consts.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"github.com/NebulousLabs/Sia/types"
+)
+
+var (
+	// DustValue is the value below which a types.Currency is considered Dust.
+	dustValue = func() types.Currency {
+		return types.SiacoinPrecision
+	}
```
lukechampine: a function???

johnathanhowell: woops, meant to add `}()`

DavidVorick: I would like to have it as a function actually. Though, of the form `func dustValue() types.Currency {`. This is because, unlike for the constants, this is something that I expect to be computed at runtime and to change as the transaction fees change.

Eventually, the transaction fees are going to be calculated dynamically, and when that transition is made the dust outputs should also be determined dynamically.

johnathanhowell: alright, I updated consts.go to have `func dustValue() types.Currency`.

DavidVorick:
should be good now

PR #1545, opened 2017-01-15
fix reporting for aggregate metrics

lukechampine:
The "empty metrics" bug was caused by this bit of code in the API:
```go
for _, c := range contracts {
	if c.StartHeight < periodStart {
		continue
	}
	// ...
}
```
The special metrics contract didn't have a start height, so the loop was skipping over it. This was fixed by setting the contract start height to `c.periodStart + 1`.
Later in the loop, `RenterFunds` is called on the contract. Since the special metrics contract doesn't have a proper revision, this caused a panic. This was fixed by adding zero-valued `ValidProofOutput`s.

During debugging, it occurred to me that old (pre-upgrade) contracts won't have a start height either. Thus they too will be skipped by the loop in the API. This means that any new uploads/downloads using the old contracts will not cause the metrics to change. I'm not sure what the best fix for this is.

### DIFF
--- a/modules/renter/contractor/persist.go
+++ b/modules/renter/contractor/persist.go
@@ -23,7 +23,7 @@ type contractorPersist struct {
 		DownloadSpending types.Currency `json:"downloadspending"`
 		StorageSpending  types.Currency `json:"storagespending"`
 		UploadSpending   types.Currency `json:"uploadspending"`
-	}
+	} `json:",omitempty"`
 }
 
 // persistData returns the data in the Contractor that will be saved to disk.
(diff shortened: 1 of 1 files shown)


DavidVorick:
seems that vet is failing

lukechampine:
ah, fixed.
Appveyor didn't fail because it doesn't run go vet

lukechampine:
Thinking more on this, we should set `c.StartHeight` for old contracts the same way we do with the metrics contract. Otherwise they won't be reflected properly in the metrics call. I'll open a PR for this shortly.

PR #1547, opened 2017-01-15
add progress to siac renter download

lukechampine:
Perhaps later on we can add an animated download bar, e.g. https://github.com/cheggaaa/pb

### DIFF
--- a/siac/rentercmd.go
+++ b/siac/rentercmd.go
@@ -6,6 +6,7 @@ import (
 	"path/filepath"
 	"sort"
 	"text/tabwriter"
+	"time"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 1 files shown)


lukechampine:
looks like we're hitting an internal error on appveyor:

>ERR Error running script (call to f_5b8e3b60e050400311bc8b6ef2cbba37bc0b9cb3): @user_script:1: @user_script: 1: -OOM command not allowed when used memory > 'maxmemory'.

lukechampine:
Sample output:
```
luke@mjolnir:~$ siac renter download foo.mkv bar.mkv
Downloading... 100.0% of 266.36 MB, 45s elapsed, 47.31 Mbps     
Downloaded 'foo.mkv' to /home/luke/bar.mkv.
```

lukechampine:
travis keeps failing :/ lots of "block does not extend the longest fork" errors

DavidVorick:
That bug has been around for a while, but ultimately pretty elusive because whenever we start debugging inevitably it becomes unreproducible.

What is the max memory? Perhaps we are running out b/c something in the download code is using a ton of RAM?

PR #1548, opened 2017-01-16
penalize workers exponentially for consecutive failures

DavidVorick:
also significantly reduce the penalty for a first time failure.

### DIFF
--- a/modules/renter/repair.go
+++ b/modules/renter/repair.go
@@ -16,7 +16,8 @@ import (
 )
 
 // TODO: Move to a consts file.
-const uploadFailureCooldown = time.Hour
+const uploadFailureCooldown = time.Second * 61 // Prime to avoid intersecting with regular events.
+const maxConsecutivePenalty = 10               // Limit the number of doublings to prevent overflows.
 const minPiecesRepair = 5
 
 var (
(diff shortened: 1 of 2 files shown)


modules/renter/repair.go
```diff
@@ -193,7 +194,11 @@ func (r *Renter) managedRepairIteration(rs *repairState) {
 		}
 
 		// Ignore workers that have had an upload failure recently.
-		if time.Since(worker.recentUploadFailure) < uploadFailureCooldown {
+		penalty := uint64(worker.consecutiveUploadFailures)
+		if worker.consecutiveUploadFailures > maxConsecutivePenalty {
+			penalty = uint64(maxConsecutivePenalty)
+		}
+		if time.Since(worker.recentUploadFailure) < uploadFailureCooldown*(1<<penalty) {
```
lukechampine: seems slightly brittle to use `<<` since it can only be used for power-of-2 exponentiation. On the other hand, doubling each time seems reasonable enough that it isn't likely to change in the future.

Alternatively, you could move the cooldown time into the worker struct. Then instead of counting failures, you could multiply the cooldown by some factor. That would allow you to exponentiate without requiring a loop.

Regardless, it would be good to have a comment here explaining that the cooldown is exponential. I would also prefer that the "max" value be a duration (e.g. `24*time.Hour`) instead of an exponent; makes it easier to tweak later.

DavidVorick: Easier to tweak but more annoying to implement. This overall felt pretty clean.

If we add the cooldown to the worker, then we need to change the recent failure mechanic from being a recent failure time to a 'cooldown expires time'.

I'll touch it up.

I'm also not really sure how to test this, I don't think we have a good framework yet for putting the renter with offline hosts.

DavidVorick: oh yeah. If you don't put a reasonable max on the number of doublings, you quickly achieve overflow. That's why I used a doubling-count instead of a max time.

lukechampine: I don't understand -- won't the outcome be the same whether you use a maximum number of doublings or a maximum duration? The longest possible duration is about 290 years so you shouldn't be running into overflows.

DavidVorick: no, because we let the consecutive failures count continue to increase even after we've reached the timeout. If we want to prevent overflows, we have to stop incrementing the consecutive failures after we've reached the maximum timeout.

The worker's job should not be to manage what time timeout is, that should be the uploaders job. The worker just reports the status and then the uploader decides which workers are useful. We don't want to cross the logic.

PR #1549, opened 2017-01-16
fake startheight for old contracts

lukechampine:
Follow up to #1545. This is important for the v1.1.0 transition because many people won't form contracts under v1.1.0 for a while. By faking the startheight of old contracts, we ensure that their metrics (upload/download/storage spending) are properly reported. Otherwise they won't be counted because the metrics logic only looks at contracts formed during the current period.

### DIFF
--- a/modules/renter/contractor/persist.go
+++ b/modules/renter/contractor/persist.go
@@ -62,9 +62,6 @@ func (c *Contractor) load() error {
 	for _, rev := range data.CachedRevisions {
 		c.cachedRevisions[rev.Revision.ParentID] = rev
 	}
-	for _, contract := range data.Contracts {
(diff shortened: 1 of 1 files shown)


DavidVorick:
That brings us to another interesting point - because getting to full 4x redundancy requires 32 contracts, people who don't form new contracts immedately after upgrading will actually be stuck at 3x or more likely ~2x (due to offline/full/etc hosts) redundancy.

Should the UI / CLI be blasting warnings if there are not enough contracts and if the allowance needs to be reformed?

DavidVorick:
Would be good if we had an antfarm tests to verify that upgrades are going smoothly. I think that's the most likely thing to cause problems in this upcoming release.

PR #1555, opened 2017-01-19
fix tpool locking bug

lukechampine:
There are two things going on here:

- `ConsensusChange` now has a `TryTransactionSet` field, which is an unexported (i.e. unlocked) version of the consensus set's `TryTransactionSet` function. This allows the transaction pool to call `TryTransactionSet` even under lock, namely during `ProcessConsensusChange`.
- The consensus set now has an exported method called `LockedTryTransactionSet`, which read-locks the consensus set and calls the supplied function, passing it an unlocked version of `TryTransactionSet`. However, this function is not part of the `ConsensusSet` interface. This allows the transaction pool to call `TryTransactionSet` multiple times without unlocking the consensus set. More importantly, it means the transaction pool can lock the consensus set first, before locking itself. This prevents a deadlock.

The effect of this is that the defrag test now passes. We may also see less NDFs.

### DIFF
--- a/modules/consensus.go
+++ b/modules/consensus.go
@@ -123,6 +123,12 @@ type (
 		// Synced indicates whether or not the ConsensusSet is synced with its
 		// peers.
 		Synced bool
+
+		// TryTransactionSet is an unlocked version of
+		// ConsensusSet.TryTransactionSet. This allows the TryTransactionSet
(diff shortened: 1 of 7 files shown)


DavidVorick:
yikes, but glad we have it fixed up.

PR #1556, opened 2017-01-21
Prune nodes more rapidly when there are many

DavidVorick:
The gateway was having a problem with a large node list, where many of the nodes
on that list were unconnectable, but the pruning was not happening fast enough
to keep the node list reasonably useful. Now, the gateway will prune nodes at a
more rapid rate when there are a lot of nodes.

This is an incomplete fix, we also need to extend the dial bit to do the full
handshake, merely seeing that a node is online is not enough if that node is not
running a recent version.

-----

This PR needs two additional things, first it needs an extension of the `g.dial` bit in the purge loop to verify that the other node is in fact a node running Sia, and on a recent enough version. Secondly, it needs testing.

### DIFF
--- a/modules/gateway/consts.go
+++ b/modules/gateway/consts.go
@@ -41,9 +41,9 @@ var (
 	// healthyNodeListLen defines the number of nodes that the gateway must
 	// have in the node list before it will stop asking peers for more nodes.
 	healthyNodeListLen = build.Select(build.Var{
-		Standard: 200,
-		Dev:      30,
-		Testing:  15,
+		Standard: int(200),
+		Dev:      int(30),
+		Testing:  int(15),
 	}).(int)
 
 	// maxSharedNodes defines the number of nodes that will be shared between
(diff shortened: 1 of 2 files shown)


modules/gateway/consts.go
```diff
@@ -68,6 +77,14 @@ var (
...
 		Testing:  10,
 	}).(int)
+
+	// quickPruneListLen defines the number of nodes that the gateway must have
+	// to be pruning nodes quickly from the node list.
+	quickPruneListLen = build.Select(build.Var{
+		Standard: 250,
+		Dev:      40,
+		Testing:  20,
+	}).(int)
```
lukechampine: best practice to explicitly cast these to `int` (i.e. `int(250)`, `int(40)`). Otherwise someone might come along and think they could change the `int` at the bottom to a `uint64` or something, which they can't -- it'll panic.

DavidVorick: It will? I was just copying the other constants, why is this one different?

lukechampine: it has to do with how constants are handled. If you have a declaration like this:
```go
const foo uint64 = 7
```
Then Go knows that `7` is supposed to interpreted as a `uint64`. But if you write this:
```go
var foo interface{} = 7
```
...the compiler just has to guess. And for integers, it always guesses `int`.

Since the fields of `build.Var` are all `interface{}`s, at compile time the integers are converted to `int` before they are assigned. Then we do a type assertion on the return value of `build.Select`. In this case, it returns an `int`, so the assertion works. But if we changed it to `uint64`, it would fail.
This is one of the unfortunate downsides of the `build.Select`/`build.Var` design.

DavidVorick: got it, will fix

DavidVorick:
comments addressed.

PR #1558, opened 2017-01-23
Add siac export command to support leaderboard

lukechampine:
This involved adding a field to the `api.RenterContract` struct. Since it's JSON, there shouldn't be any compatibility issues with the current frontends.

### DIFF
--- a/api/consensus.go
+++ b/api/consensus.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"encoding/json"
 	"net/http"
 
 	"github.com/NebulousLabs/Sia/types"
@@ -27,4 +28,21 @@ func (api *API) consensusHandler(w http.ResponseWriter, req *http.Request, _ htt
 		CurrentBlock: cbid,
 		Target:       currentTarget,
 	})
+}
+
+// consensusValidateTransactionSetHandler handles the API calls to
+// /consensus/validate/transactionset.
+func (api *API) consensusValidateTransactionSetHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	var txnset []types.Transaction
+	err := json.NewDecoder(req.Body).Decode(&txnset)
+	if err != nil {
+		WriteError(w, Error{"could not decode transaction set: " + err.Error()}, http.StatusBadRequest)
+		return
+	}
+	_, err = api.cs.TryTransactionSet(txnset)
+	if err != nil {
+		WriteError(w, Error{"transaction set validation failed: " + err.Error()}, http.StatusBadRequest)
+		return
+	}
+	WriteSuccess(w)
 }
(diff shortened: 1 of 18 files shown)


lukechampine:
I also added a commit to omit empty transaction fields from the JSON. When exporting my 44 contracts, this reduces the filesize from 100KB to 70KB. It should also significantly reduce the amount of JSON returned by calls like `/wallet/transactions`.
Changing an encoding is always risky business, especially in the `types` package, but I think it's acceptable for JSON encodings. We never hash JSON, after all. However, it may be an issue for Sia-UI if there is code that explicitly checks for `null`, since now those fields will be `undefined` instead. cc @johnathanhowell

lukechampine:
Reverted after further discussion. We can't break the API this way (the commit removes fields that were previously present in API responses).

DavidVorick:
You may be able to get away with calling omitempty on the new field

lukechampine:
true, but the new field isn't ever empty. The next step would be to convert the `types.Transaction` to a new api-specific type that omits empty fields, but that's just confusing.

DavidVorick:
Probably best just not to use the tag anywhere in the API, bandwidth and spacing typically isn't the bottleneck anyway. A few hundred kb in a large request is okay.

api/consensus.go
```diff
@@ -28,3 +29,20 @@ func (api *API) consensusHandler(w http.ResponseWriter, req *http.Request, _ htt
 		Target:       currentTarget,
 	})
 }
+
+// consensusValidateTransactionSetHandler handles the API calls to
+// /consensus/validate/transactionset.
+func (api *API) consensusValidateTransactionSetHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
```
DavidVorick: convention is to only captialize where the `/`s are

lukechampine:
keep getting an appveyor error:
```
--- FAIL: TestIntegrationLoad1of1Siag (0.48s)
	unseeded_test.go:47: rename C:\Users\appveyor\AppData\Local\Temp\1\SiaTesting\wallet\TestIntegrationLoad1of1Siag\wallet\wallet.json_temp C:\Users\appveyor\AppData\Local\Temp\1\SiaTesting\wallet\TestIntegrationLoad1of1Siag\wallet\wallet.json: The process cannot access the file because it is being used by another process.
```

doc/API.md
```diff
@@ -176,6 +177,19 @@ returns information about the consensus set, such as the current block height.
 }
 ```
 
+#### /consensus/validate/transactionset [POST]
+
+validates a set of transactions using the current utxo set.
```
DavidVorick: Determines whether a transaction set is valid using the current state of the network. A valid transaction could be included in the next block if mined.

PR #1563, opened 2017-01-28
HostDB tabbed print

lolsteve:
Modified the `hostdb` command to print tabbed and show the number of hosts. I tried to follow the `gateway list` format.

selection_034


DavidVorick:
This looks good to me. I am almost embarrassed that we did not have it already. Thanks!

PR #1564, opened 2017-01-29
Hostdb fixes

DavidVorick:
Originally this PR was meant to fix up some of the scanning issues in the hostdb. I went ahead and stepped it forward a lot though. This PR has insufficient testing, but is otherwise approximately complete and ready for code review.

Hosts now are only tracked through the host tree, and all hosts are in the host tree. Scanning is now based on hosts with the highest score, and downtime can dramatically affect score. The old persist was repurposed to fit the new style, re-using the 'AllHosts' field of the old persist and basically ignoring the existing 'ActiveHosts' field, as we have a new way to determine whether a host is active or inactive.

It's not clear to me that this code handles all edge cases gracefully, I guess this is a part of testing.

I discovered some pretty severe deadlock possiblities in the host, introduced by our wallet fixes. It's not clear to me how much work is needed to fix this, but it's likely a decent amount.

A new API call has been added that allows renters to view the expected cost of various actions on the network.

The number of hosts being used to upload data has been increased to 50. I expect this to be the final value for the number of contracts that the renter will form by default, at least for consumers. This is because larger values incur heavy contract fees.

The redundancy parameters have been changed to 10-of-30, which I predict will be pretty stable. Eventually, as the network scales, I think we will switch to 20-of-40, and then stop there for consumers. This will happen when we are better at picking faster, more stable hosts, and we are confident that 10 extra hosts is enough to keep uploads running at high speed for the whole 2x redundancy. We are definitely not there yet.


modules/renter/contractor/persist.go
```diff
@@ -120,6 +133,26 @@ func (c *Contractor) load() error {
...
+	// known. Use the hostdb to make sure the relationship set is fully filled
+	// out.
+	allHosts := c.hdb.AllHosts()
+	hmap := make(map[modules.NetAddress]modules.HostDBEntry)
+	// Iterate backwards (lowest score to highest) so that in the event of
+	// duplicate hosts for a netaddress, the highest score host is the one in
+	// the map.
+	for i := len(allHosts) - 1; i >= 0; i-- {
+		hmap[allHosts[i].NetAddress] = allHosts[i]
+	}
```
lukechampine: interesting, but maybe too clever, since it relies on allHosts being properly sorted. You can do this explicitly at the cost of verbosity:
```go
for _, h := range c.hdb.AllHosts() {
    // in the event of a duplicate address, use the host with the highest score
    if dup, exists := hmap[h.NetAddress]; !exists || dup.Score < h.Score {
        hmap[h.NetAddress] = h
    }
}
```
Although for that to work, you'd need an external way of determining the score...

modules/renter/contractor/uptime_test.go
```diff
@@ -13,14 +13,14 @@ import (
...
 }
 
 // Host returns the host with address addr. If addr matches hdb.addr, the
 // host's scan history will be modified to make the host appear offline.
-func (hdb offlineHostDB) Host(addr modules.NetAddress) (modules.HostDBEntry, bool) {
-	host, ok := hdb.hostDB.Host(addr)
-	if ok && addr == hdb.addr {
+func (hdb offlineHostDB) Host(spk types.SiaPublicKey) (modules.HostDBEntry, bool) {
+	host, ok := hdb.hostDB.Host(spk)
+	if ok && string(spk.Key) == string(hdb.spk.Key) {
```
lukechampine: `bytes.Equal`

modules/renter/hostdb/hostdb.go
```diff
@@ -52,31 +51,27 @@ type HostDB struct {
...
 	scanWait bool
+	online   bool
 
 	blockHeight types.BlockHeight
 	lastChange  modules.ConsensusChangeID
 }
 
 // New returns a new HostDB.
-func New(cs consensusSet, persistDir string) (*HostDB, error) {
+func New(g modules.Gateway, cs consensusSet, persistDir string) (*HostDB, error) {
```
lukechampine: should use a new interface defined in dependencies.go. Looks like you only need:
```go
type gateway interface {
    Peers() []modules.NetAddress
}
```

modules/renter/hostdb/hostdb_test.go
```diff
@@ -3,58 +3,115 @@ package hostdb
...
 	}
 	hdb.hostTree = hosttree.New(hdb.calculateHostWeight)
 	return hdb
 }
 
 // newStub is used to test the New function. It implements all of the hostdb's
 // dependencies.
-type newStub struct{}
+type newStub struct {
+	*gateway.Gateway
```
lukechampine: replace this with a stub `Peers` method, as with `ConsensusSetSubscribe`

modules/renter/hostdb/hosttree/hosttree.go
```diff
@@ -181,6 +185,24 @@ func (n *node) remove() {
...
+// All returns all of the hosts in the host tree, sorted by weight.
+func (ht *HostTree) All() []modules.HostDBEntry {
+	ht.mu.Lock()
+	defer ht.mu.Unlock()
+
+	var he hostEntries
+	for _, node := range ht.hosts {
+		he = append(he, *node.entry)
+	}
+	sort.Sort(he)
```
lukechampine: convention is to name the sorting type after the sorting criteria, e.g. `sort.Sort(byWeight(he))`

modules/renter/hostdb/hosttree/sort.go
```diff
@@ -0,0 +1,7 @@
+package hosttree
+
+type hostEntries []hostEntry
+
+func (he hostEntries) Len() int           { return len(he) }
+func (he hostEntries) Less(i, j int) bool { return he[i].weight.Cmp(he[j].weight) < 0 }
+func (he hostEntries) Swap(i, j int)      { he[i], he[j] = he[j], he[i] }
```
lukechampine: does this need its own file?

modules/renter/hostdb/online.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"time"
+)
+
+// online.go performs regular checks against the gateway to see if there is
+// internet connectivity. The check is performed by looking for at least one
+// non-local peer in the peer list. If there is no internet connectivity, scans
+// must be stopped lest we penalize otherwise online hosts.
+
+func (hdb *HostDB) threadedOnlineCheck() {
```
lukechampine: I searched around and it seems like the only robust alternative to this is to ping a high-uptime server like `8.8.8.8`. Obviously there are disadvantages to that, but at least we wouldn't need to thread a new dependency through all this code. It could also be used as a fallback, though it's probably a bad sign if you don't have any peers.

modules/renter/contractor/contractor.go
```diff
@@ -41,27 +41,29 @@ type Contractor struct {
...
+	revising    map[types.FileContractID]bool // prevent overlapping revisions
+
 	cachedRevisions map[types.FileContractID]cachedRevision
 	contracts       map[types.FileContractID]modules.RenterContract
-	currentPeriod   types.BlockHeight
-	downloaders     map[types.FileContractID]*hostDownloader
-	editors         map[types.FileContractID]*hostEditor
-	lastChange      modules.ConsensusChangeID
 	oldContracts    map[types.FileContractID]modules.RenterContract
+	relationships   map[types.FileContractID]types.SiaPublicKey
```
DavidVorick: I determined that the compatibility issues would be harder to work around.

This map is temporary, just like the oldContracts map. I'm procedurally migrating everything to just using pubkeys, as we do this in more parts of the codebase we may accumulate a few more helper-maps, but then finally we'll be able to drop them all at the same time.

types/signatures.go
```diff
@@ -391,6 +393,21 @@ func (t *Transaction) validSignatures(currentHeight BlockHeight) error {
 	return nil
 }
 
+// LoadString is the inverse of SiaPublicKey.String().
+func (spk *SiaPublicKey) LoadString(s string) {
```
DavidVorick: hostdb requires having the whole SiaPublicKey.

modules/renter/contractor/persist.go
DavidVorick: If it's not in the AllHosts definition that they will be returned in sorted order, I will add it. It's much more useful to see how the hosts stack up.

modules/renter/hostdb/hostdb.go
DavidVorick: I really don't like that style of dependency management, because it adds a lot of annoyance when you decide that you want some other gateway function as well. We don't want an interface that can tell us what peers we have, we want the whole entire gateway that does all the gateway things. Some alternate interface which has a Peers() method but isn't a gateway (maybe it's a kadmelia implementation or something) wouldn't be the same, because it doesn't have all the same standards for aggressively hunting non-local outbound peers, and might not use the same keepalive algorithms.

modules/renter/hostdb/hostdb_test.go
DavidVorick: I am in the process of pulling all of the stubs out of the hostdb, they make it very high-friction to change any internals.

modules/renter/hostdb/hosttree/hosttree.go
DavidVorick: has that been released yet?

modules/renter/hostdb/online.go
DavidVorick: The depedencies are already there, the hostdb already needs a consensus set which needs a gateway. While they are now more visible, it doesn't change the dependency graph too much.

I don't think it's likely to cause problems, because the gateway inherently aggressively looks for nonlocal outbound peers. Would be a very different network if somehow it didn't need to be doing that.

modules/renter/contractor/persist.go
lukechampine: that's ok, but it's only a weak guarantee. Reminds me of the [HostDBScan thing](https://github.com/NebulousLabs/Sia/blob/master/modules/renter/contractor/uptime.go#L58) where we decided to add a sanity check. But we can't do a sanity check here because we don't have access to the sorting criteria.

Perhaps `AllHosts` should do the filtering internally, so that it is guaranteed to return only one host per NetAddress?

modules/renter/hostdb/hostdb_test.go
```diff
@@ -3,58 +3,115 @@ package hostdb
...
+}
 
 // consensus set stubs
 func (newStub) ConsensusSetSubscribe(modules.ConsensusSetSubscriber, modules.ConsensusChangeID) error {
 	return nil
 }
 
+// newHDBTester returns a tester object wrapping a HostDB and some extra
+// information for testing.
+func newHDBTester(name string) (*hdbTester, error) {
```
lukechampine: this function doesn't appear to be used

modules/renter/hostdb/hostdb.go
lukechampine: Why do you say we want the entire gateway? We only want to determine whether we are online. The downside of using a full gateway is that now you need to supply a full gateway whenever you want to construct a HostDB for testing. You can't easily mock the gateway in order to test `threadedOnlineCheck`.

I agree that it shouldn't have a `Peers` method though. The fact that we use the gateway's peers to determine online status is an implementation detail. What you really want is a `HasInternetConnection() bool` method. That makes it clear what your intention is. Then wrap the full gateway in a type that uses `g.Peers()` to implement that method.

I know that restrictive typing is a pain, but it pays off in the long run by minimizing surface area and clarifying intent. Think of it as a form of documentation.

modules/renter/hostdb/hosttree/hosttree.go
lukechampine: you are thinking of `sort.Slice`. I'm just suggesting renaming `hostEntries` to `byWeight`.

modules/renter/hostdb/scan.go
```diff
@@ -254,6 +194,26 @@ func (hdb *HostDB) threadedProbeHosts() {
 		case <-hdb.tg.StopChan():
 			return
 		case hostEntry := <-hdb.scanPool:
+			// Block the scan until the host is online.
```
lukechampine: what happens if the hostdb is deleted and a full blockchain rescan occurs without internet? Do the host announcements accumulate in a queue somewhere, or do they get dropped?

types/signatures.go
lukechampine: I don't understand what you mean by that

modules/renter/contractor/contractor.go
lukechampine: can you elaborate on the compatibility issues? Couldn't we just add a check for "if contract doesn't have a pubkey, look it up in the hostdb?"

modules/renter/contractor/persist.go
DavidVorick: No because this is really only a hack around the renter's dependency on the NetAddress type. Really we should not be filtering by NetAddress at all, everything should be pubkey based.

I would write a sanity check and also a test but the weight function is not exported.

modules/renter/hostdb/hostdb.go
DavidVorick: You already know that I disagree with mocking entire dependencies for testing. You want the side effects and interfereing processes, because those effects will be there during production. If it makes the test hard to write or NDF, it's often a bug upstream, and then you are glad you found the bug. Supplying a full gateway is easy when you've got a helper function in the fashion of the contract manager testing. You ask for the root dependencies and then if you need you can overwrite some of them individually to trigger certain codepaths.

I agree that really we want a `HasInternetConnection` method as opposed to a `Peers` method. I guess we can extend the gateway to support that method.

I'm not sure exactly what you mean when you say restrictive typing, but if you are talking about redefining every dependency as a new interface which is the subset of functions that you need, I am pretty strongly against the practice. I feel like it makes testing more difficult, and it also leads you to believe that you can easily swap in some other object to satisfy the interface when the reality is that these modules are designed to depend on eachother, not on eachother's subsets.

modules/renter/hostdb/hosttree/hosttree.go
DavidVorick: ah

modules/renter/hostdb/hostdb_test.go
DavidVorick: still working on the testing, there are some unpushed commits using this function

modules/renter/hostdb/scan.go
DavidVorick: I may not have pushed the code yet, but they get inserted into the host tree immediately upon being discovered in the blockchain.

A full rescan won't occur at this point in the code though. By the time any of the scan loops have launched, any hdb startup like a rescan has already completed.

doc/API.md
```diff
@@ -722,6 +723,21 @@ lists the status of all files.
...
+
+lists the estimated prices of performing various network operations.
+
+###### JSON Response [(with comments)](/doc/api/Renter.md#json-response-4)
+```javascript
+{
+  "downloadterabyte": "1234", // hastings
+  "formcontracts":    "1234", // hastings
+  "storageterabyte":  "1234", // hastings
+  "uploadterabyte":   "1234", // hastings
```
DavidVorick: it's important here because we're normalizing to unusual units. Sia has a defined set of units that it uses for everything - hastings, bytes, blocks. But here we are departing from that, so I felt that we should be explicit. Especially because it's pretty arbitrary.

doc/API.md
```diff
@@ -722,6 +769,21 @@ lists the status of all files.
...
+#### /renter/prices [GET]
+
+lists the estimated prices of performing various network operations.
+
+###### JSON Response [(with comments)](/doc/api/Renter.md#json-response-4)
+```javascript
+{
+  "downloadterabyte": "1234", // hastings
+  "formcontracts":    "1234", // hastings
+  "storageterabyte":  "1234", // hastings
```
DavidVorick: I am wondering if this should have the word 'month' in it.

doc/api/HostDB.md
```diff
@@ -166,6 +167,119 @@ any particular order, and the order may change in subsequent calls.
...
+    // minimum size of window that the host will accept in a file contract.
+    "windowsize": 144,
+
+    // Public key used to identify and verify hosts.
+    "publickey": {
+      // Algorithm used for signing and verification. Typically "ed25519".
+      "algorithm": "ed25519",
+
+      // Key used to verify signed host messages.
+      "key": "RW50cm9weSBpc24ndCB3aGF0IGl0IHVzZWQgdG8gYmU="
```
DavidVorick: note that this here is displayed in base64

doc/api/HostDB.md
```diff
@@ -288,3 +402,46 @@ Examples
   ]
 }
 ```
+
+#### Hosts
+
+###### Request
+```
+/hostdb/hosts/ed25519:1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef
```
DavidVorick: note that this here is a string.

DavidVorick: should we return the string value of the public key from the API? I'm guessing that we should. We should probably add it as an extra field to the entry.

doc/api/HostDB.md
```diff
@@ -18,10 +18,11 @@ identifies hosts by their public key and keeps track of metrics such as price.
...
 
-| Request                                     | HTTP Verb | Examples                      |
-| ------------------------------------------- | --------- | ----------------------------- |
-| [/hostdb/active](#hostdbactive-get-example) | GET       | [Active hosts](#active-hosts) |
-| [/hostdb/all](#hostdball-get-example)       | GET       | [All hosts](#all-hosts)       |
+| Request                                                 | HTTP Verb | Examples                      |
+| ------------------------------------------------------- | --------- | ----------------------------- |
+| [/hostdb/active](#hostdbactive-get-example)             | GET       | [Active hosts](#active-hosts) |
+| [/hostdb/all](#hostdball-get-example)                   | GET       | [All hosts](#all-hosts)       |
+| [/hostdb/hosts/___:pubkey___](#hostdbhosts-get-example) | GET       | [Hosts](#hosts)               |
```
DavidVorick: Because the other calls are /hostdb/active, I was considering just making this /hostdb/:pubkey - wasn't sure though.

DavidVorick:
The 'wip' commits come from when I switch machines, they are easily squashed in the case of this PR. I tried pretty hard to keep concerns separated, and it only really got messy at the end when I started changing the dependencies.

PR should be logically complete, though I'm going to be adding a bunch of testing over the next few days.

modules/host/upnp.go
```diff
@@ -69,12 +78,14 @@ func (h *Host) managedLearnHostname() {
 	// has a storage obligation. If the host is not accepting contracts and has
 	// no open contracts, there is no reason to notify anyone that the host's
 	// address has changed.
-	if h.settings.AcceptingContracts || h.financialMetrics.ContractCount > 0 {
-		err = h.announce(autoAddress)
+	if hAcc || hCC > 0 {
```
DavidVorick: laziness is a disease.

If the variable is used more than 3-4 lines from where it's declared, I'll change it.

modules/renter/contractor/contractor.go
DavidVorick: well, you can't look anything up in the hostdb without a pubkey because the host only keys things in terms of pubkeys now.

I'm cautious to add anything to `modules.RenterContract` because I'm not sure if its used in the persist anywhere.

DavidVorick:
re: PR hygiene

+ will squash commits + rebase after PR is fully ready to go, I like looking at the build history for Travis + Appveyor, don't want to throw that out.
+ adding the API calls could/should have been done in a separate PR I think, otoh I've been using them for testing, and I don't think I would have pushed through this PR without them coded up.

doc/api/HostDB.md
lukechampine: should specify that in the docs

doc/api/HostDB.md
lukechampine: ah, because the caller needs it to construct the request?

It is a little confusing to have three different representations of the pubkey flying around. In the string, it's hex; in JSON, it's base64; and in memory, it's binary.

DavidVorick: If I could go back in time, there would only ever be the hex/string version. The base64 version was never meant to be human readable, and it only slipped through because we didn't think that anyone would ever be directly looking at pubkeys.

api/hostdb_test.go
```diff
@@ -0,0 +1,217 @@
...
+	if err = st.getAPI("/hostdb/all", &ah); err != nil {
+		t.Fatal(err)
+	}
+	if len(ah.Hosts) != 1 {
+		t.Fatalf("expected 1 host, got %v", len(ah.Hosts))
+	}
+}
+
+// TestHostDBHostsHandler checks that the hosts handler is easily able to return
+func TestHostDBHostsHandler(t *testing.T) {
```
DavidVorick: This test is new, the others in this file are copy-paste

api/host.go
```diff
@@ -71,35 +72,127 @@ func (api *API) hostHandlerGET(w http.ResponseWriter, req *http.Request, _ httpr
...
+	}
+	if req.FormValue("minuploadbandwidthprice") != "" {
+		var x types.Currency
+		_, err := fmt.Sscan(req.FormValue("minuploadbandwidthprice"), &x)
+		if err != nil {
+			WriteError(w, Error{"Malformed minuploadbandwidthprice"}, http.StatusBadRequest)
+			return
+		}
+		settings.MinUploadBandwidthPrice = x
+	}
```
DavidVorick: should be checked for spelling.

api/hostdb.go
```diff
@@ -0,0 +1,84 @@
...
+		WriteError(w, Error{"requested host does not exist"}, http.StatusBadRequest)
+		return
+	}
+	breakdown := api.renter.ScoreBreakdown(entry)
+
+	WriteJSON(w, HostdbHostsGET{
+		Entry:          entry,
+		ScoreBreakdown: breakdown,
+	})
+}
```
DavidVorick: This function is new, the others are copy-paste.

api/renter.go
```diff
@@ -22,7 +22,7 @@ var (
 	// contracts with if the value is not specified explicity in the call to
 	// SetSettings.
 	recommendedHosts = build.Select(build.Var{
-		Standard: uint64(42),
+		Standard: uint64(50),
```
DavidVorick: note the upgrade to using 50 hosts. The long-term plan is to top at 50 hosts. At 82 hosts visible to the network, we're basically on the fringe territory of safe. We can afford up to 32 hosts who are too expensive or otherwise being dumb. Sia is on a good growth trajectory though, and we've been above 70 hosts for a very long time.

I think it's fine.

42 -> 50 means that we have at all times an extra 8 hosts to use for speedboosts, etc. File redundancy was also changed to using 30 total hosts instead of 32, so it means that the final 1x redundancy will always be able to use at least 20x parallelism. It will be the 20 slowest hosts, but it's still 20 hosts. Previously it was just 10. And further, of those 10, we were seeing 8 completely unusable in the common case. So really it was only 3 hosts in use in the common case. Here we should see that number go from 3 -> 12 in the common case.

This PR also fixes some of the bugs that were causing hosts to be unreachable (can't use a host that's not in the hostdb). So that '8 in the common case' should also be going down automatically.

Long story short, this combined with a number of other changes should mean that people don't have problems anymore seeing slow uploads on the tail of a file.

api/renter.go
```diff
@@ -302,6 +292,12 @@ func (api *API) renterFilesHandler(w http.ResponseWriter, req *http.Request, _ h
 	})
 }
 
+// renterPricesHandler reports the expected costs of various actions given the
+// renter settings and the set of available hosts.
+func (api *API) renterPricesHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	WriteJSON(w, api.renter.PriceEstimation())
+}
```
DavidVorick: Most API functions have their own struct defined. In this particular case I didn't see any reason that we might have extra fields that would not appear in the PriceEstimation struct. OTOH, if we eventually decide to change how something is encoded or have some compatibility departure, perhaps it is best that the struct get redeifined.

Actually, I think I will take a commit to properly define it as all the other functions do. =/ heh

api/renter_test.go
```diff
@@ -767,3 +645,377 @@ func TestRenterRelativePathErrorDownload(t *testing.T) {
...
+		t.Log(rpeSingle.FormContracts)
+		t.Error("price did not drop from single to multi")
+	}
+	if !(rpeMulti.StorageTerabyteMonth.Cmp(rpeSingle.StorageTerabyteMonth) > 0) {
+		t.Error("price did not drop from single to multi")
+	}
+	if !(rpeMulti.UploadTerabyte.Cmp(rpeSingle.UploadTerabyte) > 0) {
+		t.Error("price did not drop from single to multi")
+	}
+}
```
DavidVorick: I felt kind of bad because these tests take like 8 seconds to run, but they caught bugs in I think 4 different packages, so I don't feel bad after all.

build/critical.go
```diff
@@ -12,11 +12,11 @@ import (
 // the running goroutine is printed to help determine the error.
 func Critical(v ...interface{}) {
 	s := "Critical error: " + fmt.Sprintln(v...) + "Please submit a bug report here: https://github.com/NebulousLabs/Sia/issues\n"
+	debug.PrintStack()
```
DavidVorick: I think we should test that the UI is capturing this output correctly, at the moment I don't actually think that it is.

doc/api/HostDB.md
DavidVorick: I will probably add a field "pubkeystring" so that other implementers don't need to muck around with a base64->hex conversion following the Sia conventions.

doc/api/Renter.md
```diff
@@ -211,6 +212,32 @@ lists the status of all files.
...
+	  // contracts.
+      "formcontracts": "1234", // hastings
+
+      // The estimated cost of storing one terabyte of data on the network for
+	  // a month, including accounting for redundancy.
+      "storageterabytemonth": "1234", // hastings
+
+	  // The estimated cost of uploading one terabyte of data to the network,
+	  // including accounting for redundancy.
+      "uploadterabyte": "1234", // hastings
```
DavidVorick: yes, it's currently hard-coded. We can't look at the renter settings because the renter doesn't have a default setting for file redundancy.

This adds some friction and misunderstanding, but I don't want to mislead people by showing a non-redundant price and then oops-it's-3x-the-price-we-said.

modules/host.go
```diff
@@ -104,6 +104,9 @@ type (
 		// have been made to the host.
 		NetworkMetrics() HostNetworkMetrics
 
+		// PublicKey returns the public key of the host.
+		PublicKey() types.SiaPublicKey
```
DavidVorick: you may expect this function to be in the HostExternalSettings but we explicitly chose to have it be separate because we didn't want the renter code accidentally reading off a new pubkey and thinking it was the same host (which would be a nasty way to bypass Sybil attack defenses we have)

modules/host/announce.go
```diff
@@ -16,22 +16,27 @@ var (
 	errUnknownAddress = errors.New("host cannot announce, does not seem to have a valid address.")
 )
 
-// announce creates an announcement transaction and submits it to the network.
-func (h *Host) announce(addr modules.NetAddress) error {
+// managedAnnounce creates an announcement transaction and submits it to the network.
+func (h *Host) managedAnnounce(addr modules.NetAddress) error {
```
DavidVorick: This should be in another PR, but I had finished coding it before I realized that the problem was in more places than just this.

Basically it's a violation of conventions to call `AcceptTransactionSet` while the host is under lock (locking an external package while retaining a lock over your own package is not allowed) and in this case the violation caused a deadlock.

Unfortunately, the host contract negotiation stuff has this problem in a much more deeply embedded way =/. Wallet also appears to have this problem, and in each case they cause deadlocks.

modules/host/consts.go
```diff
@@ -45,13 +45,13 @@ var (
 	// keep the data even if the price of siacoin fluctuates, the price of raw
 	// storage fluctuates, or the host realizes that there is unexpected
 	// opportunity cost in being a host.
-	defaultCollateral = types.SiacoinPrecision.Mul64(500).Div(modules.BlockBytesPerMonthTerabyte) // 500 SC / TB / Month
+	defaultCollateral = types.SiacoinPrecision.Mul64(100).Div(modules.BlockBytesPerMonthTerabyte) // 100 SC / TB / Month
```
DavidVorick: lowered the default collateral because a lot of hosts complaining about how much collateral they were using.

this default will definitely be going back up. But while we're below 100-200 hosts I want to make becoming a host on the more friendly side - especially while there's actually not much uploading happening.

modules/host/consts.go
```diff
@@ -45,13 +45,13 @@ var (
...
 
 	// defaultCollateralBudget defines the maximum number of siacoins that the
 	// host is going to allocate towards collateral. The number has been chosen
 	// as a number that is large, but not so large that someone would be
 	// furious for losing access to it for a few weeks.
-	defaultCollateralBudget = types.SiacoinPrecision.Mul64(2e6)
+	defaultCollateralBudget = types.SiacoinPrecision.Mul64(1e6)
```
DavidVorick: not sure what host has 1 million siacoins set aside, but I didn't want to set the default lower because I don't think it's exposed in either case.

modules/host/consts.go
```diff
@@ -89,12 +89,12 @@ var (
 	// away by only a few hundred file contracts. As the ecosystem matures, it
 	// is expected that the safe default for this value will increase quite a
 	// bit.
-	defaultMaxCollateral = types.SiacoinPrecision.Mul64(100e3)
+	defaultMaxCollateral = types.SiacoinPrecision.Mul64(5e3)
```
DavidVorick: again this is so someone can't drain the entire wallet in a single contract.

modules/host/consts.go
```diff
@@ -89,12 +89,12 @@ var (
...
 
 	// defaultStoragePrice defines the starting price for hosts selling
 	// storage. We try to match a number that is both reasonably profitable and
 	// reasonably competitive.
-	defaultStoragePrice = types.SiacoinPrecision.Mul64(500).Div(modules.BlockBytesPerMonthTerabyte) // 500 SC / TB / Month
+	defaultStoragePrice = types.SiacoinPrecision.Mul64(750).Div(modules.BlockBytesPerMonthTerabyte) // 750 SC / TB / Month
```
DavidVorick: meh felt like raising the default price was an okay choice. Might bite us if the siacoin price skyrockets.

DavidVorick: mostly I just want hosts to be seeing income.

modules/host/update.go
```diff
@@ -71,17 +71,22 @@ func (h *Host) initRescan() error {
...
 			h.log.Println("dropping storage obligation during rescan, id", so.id())
-			return composeErrors(err, h.removeStorageObligation(so, obligationRejected))
 		}
+
+		go func(i int) {
+			err := h.tpool.AcceptTransactionSet(allObligations[i].OriginTransactionSet)
+			if err != nil {
+				h.log.Println("Unable to submit contract transaction set after rescan:", soid)
+			}
+		}(i)
```
DavidVorick: more yuck. This also relates to the deadlock thing. Tests I don't think will pass without this fix in particular.

modules/host/upnp.go
```diff
@@ -23,9 +23,18 @@ func (h *Host) managedLearnHostname() {
...
+	// Fetch a group of host vars that will be used to dictate the logic of the
+	// function.
 	h.mu.RLock()
 	netAddr := h.settings.NetAddress
+	hPort := h.port
+	hAuto := h.autoAddress
+	hAnn := h.announced
+	hAcc := h.settings.AcceptingContracts
+	hCC := h.financialMetrics.ContractCount
 	h.mu.RUnlock()
```
DavidVorick: Same bother-ness with the deadlocking. Had to completely turn around the locking strategy b/c this function makes and submits a transaction.

modules/renter.go
```diff
@@ -56,46 +75,64 @@ type FileInfo struct {
...
-// RenterSettings control the behavior of the Renter.
-type RenterSettings struct {
-	Allowance Allowance `json:"allowance"`
-}
-
 // A HostDBEntry represents one host entry in the Renter's host DB. It
 // aggregates the host's external settings and metrics with its public key.
+//
+// TODO: WE FORGOT THE JSON STRINGS FOR SOME OF THE ELEMENTS! It's used in both
+// persist and over the API.
```
DavidVorick: Special attention here. I'm not sure what compatibility is broken when I added these new json things. But mostly I think it just means that hosts will lose their scan history, and their 'FirstSeen' field. We should probably add some compat code to recover the FirstSeen value, but maybe not that important.

DavidVorick: if we don't add compat code, a bunch of hosts who are only a week old may suddenly look very old. New hosts will get the right first seen value added. It's not a huge componenet of the score, and it'll only be an issue for the first few weeks after someone upgrades from v1.1.0. These values iirc don't even exist in v1.0.4, so in that sense it's already an issue for nodes that upgrade.

modules/renter.go
```diff
@@ -56,46 +75,64 @@ type FileInfo struct {
...
+// significantly from machine to machine.
+type HostScoreBreakdown struct {
+	AgeAdjustment              float64 `json:"ageadjustment"`
+	BurnAdjustment             float64 `json:"burnadjustment"`
+	CollateralAdjustment       float64 `json:"collateraladjustment"`
+	PriceAdjustment            float64 `json:"pricesmultiplier"`
+	StorageRemainingAdjustment float64 `json:"storageremainingadjustment"`
+	UptimeAdjustment           float64 `json:"uptimeadjustment"`
+	VersionAdjustment          float64 `json:"versionadjustment"`
+}
```
DavidVorick: I decided that everything in this struct would either be a float or a bool. I could have added a final score, but decided not to. If we ever deprecate a float, it'll always return '1'. If we deprecate a bool, it'll always return 'false'.

The bools we have time to change our mind about. But basically they'd be things like 'IsEurope' or 'IsHIPPACertified', stuff like that. If we deprecate a category, nothing is in that category anymore. Generally, a bool does not affect your score except to disqualify you from a bucket. For example, we may require that there be at least 2 hosts on every continent.

That's how I decided to handle compatibility here.

modules/renter.go
lukechampine: I actually don't think this breaks compatibility: https://play.golang.org/p/R5hrFPYjyW

modules/renter.go
```diff
@@ -56,46 +75,64 @@ type FileInfo struct {
...
+// various operations on the network.
+type RenterPriceEstimation struct {
+	// The cost of downloading 1 TB of data.
+	DownloadTerabyte types.Currency `json:"downloadterabyte"`
+
+	// The cost of forming a set of contracts using the defaults.
+	FormContracts types.Currency `json:"formcontracts"`
+
+	// The cost of storing 1 TB for a month, including redundancy.
+	StorageTerabyteMonth types.Currency `json:"storageterabyte"`
```
DavidVorick: I should change the json here.

modules/renter/hostdb/consts.go
```diff
@@ -0,0 +1,36 @@
...
+	hostScanDeadline   = 60 * time.Second
+)
+
+var (
+	// hostCheckupQuantity specifies the number of hosts that get scanned every
+	// time there is a regular scanning operation.
+	hostCheckupQuantity = build.Select(build.Var{
+		Standard: int(250),
+		Dev:      int(4),
+		Testing:  int(5),
```
DavidVorick: I am reconsidering these values. I wanted them to be low enough that in testing we could surpass them with online hosts. I'm not actually sure how many hosts we can reasonably run simultaneously on a single machine at this time.

modules/renter/hostdb/consts.go
```diff
@@ -0,0 +1,36 @@
...
+	}).(int)
+
+	// scanningThreads is the number of threads that will be probing hosts for
+	// their settings and checking for reliability.
+	scanningThreads = build.Select(build.Var{
+		Standard: int(25),
+		Dev:      int(6),
+		Testing:  int(3),
+	}).(int)
```
DavidVorick: hmm. Definitely silly that we have 6 scanning threads for only 4 hosts. Maybe the number of scanning threads should be 3 for both? I want enough that race conditions can be spotted.

I guess ideally we'd find way to run antfarms with 20+ hosts on them.

modules/renter/hostdb/dependencies.go
```diff
@@ -12,63 +11,34 @@ import (
...
-		save(hdbPersist) error
-		saveSync(hdbPersist) error
-		load(*hdbPersist) error
+	dependencies interface {
+		dialTimeout(modules.NetAddress, time.Duration) (net.Conn, error)
+		disrupt(string) bool
+		loadFile(persist.Metadata, interface{}, string) error
+		saveFile(persist.Metadata, interface{}, string) error
+		saveFileSync(persist.Metadata, interface{}, string) error
+		sleep(time.Duration)
```
DavidVorick: I still feel bad about this. The persister object was actually a pretty good idea, made some things a lot nicer. I got rid of it because it was unfamiliar, but in the future would be open to bringing things like that back. Though it did strike me as too much on the side of clever.

modules/renter/hostdb/hostdb.go
```diff
@@ -35,137 +37,177 @@ const (
...
 		go hdb.threadedProbeHosts()
 	}
-	go hdb.threadedScan()
+
+	// Spawn the scan loop during production, but allow it to be disrupted
+	// during testing. Primary reason is so that we can fill the hostdb with
+	// fake hosts and not have them marked as offline as the scanloop operates.
+	if !hdb.deps.disrupt("disableScanLoop") {
+		go hdb.threadedScan()
+	}
```
DavidVorick: This is essentially the alternative to the barehostdb. Disable just the features that you need to disable, as opposed to everything.

modules/renter/hostdb/hostdb_test.go
```diff
@@ -1,111 +1,381 @@
...
+	prodDependencies
+}
+
+// Send a disrupt signal to the quitAfterLoad codebreak.
+func (disableScanLoopDeps) disrupt(s string) bool {
+	if s == "disableScanLoop" {
+		return true
+	}
+	return false
+}
```
DavidVorick: 11 loc to mock out a dependency.

modules/renter/upload.go
```diff
@@ -37,7 +37,7 @@ var (
 		case "dev":
 			return 1
 		case "standard":
-			return 24
+			return 20
```
DavidVorick: files are now 10-of-30

siac/hostcmd.go
```diff
@@ -130,9 +130,9 @@ func hostcmd() {
 	}
 
 	// Determine the competitive price string.
-	ah := new(api.ActiveHosts)
+	ah := new(api.HostdbActiveGET)
 	var competitivePrice string
-	err = getAPI("/hostdb/active?numhosts=24", ah)
+	err = getAPI("/hostdb/active?numhosts=32", ah)
```
DavidVorick: I need to change this to 50

DavidVorick: nvm I didn't realize this was for the host competitive price estimation, and not for the renter side of things. Host competitive price estimator intentionally undershoots a bit to account for potential other factors that may affect host score.

DavidVorick:
uh. Looks like I screwed up a rebase =/

DavidVorick:
I went ahead and wrote some new siac commands to get a glimpse into the true depths of the hostdb. But, enough is enough, so I've decided to hold off on PR'ing them until this gets merged. I've also got a handful of other improvements for the hostdb, but I will hold off on those as well.

So basically, except for testing the renter relationships stuff and for fixing a bug that seems to be in the hostdb (not tracking height correctly), I plan to push everything into the next PR. I maybe should have split this into a feature branch, though we already have 2 of those that aren't doing so well.

doc/api/Renter.md
```diff
@@ -211,6 +212,32 @@ lists the status of all files.
...
+```javascript
+{
+      // The estimated cost of downloading one terabyte of data from the
+	  // network.
+      "downloadterabyte": "1234", // hastings
+
+	  // The estimated cost of forming a set of contracts on the network. This
+	  // cost also applies to the estimated cost of renewing the renter's set of
+	  // contracts.
+      "formcontracts": "1234", // hastings
```
DavidVorick: they will have to account for that themselves, or we can extend the api to take in a proposed set of parameters.

types/signatures.go
DavidVorick: it's used in the api because you query a host by its public key as a string.

modules/renter/contractor/persist.go
DavidVorick: I'm dumb. Sorting things already sorts them from lowest to highest. I was doing this backwards. Oops.

DavidVorick: I really do wish we could throw a sanity check here but there's no way to pull the weight of an entry from the hostdb and I'm not convinced its worth expanding the interface just for a sanity check.

modules/renter/hostdb/hosttree/sort.go
DavidVorick: Normally I'd say no but there are only 3 files total in the package, so it seemed alright.

modules/renter/hostdb/hostweight.go
```diff
@@ -41,104 +128,227 @@ func (hdb *HostDB) calculateHostWeight(entry modules.HostDBEntry) types.Currency
...
+// elements of the host's overall score.
+func (hdb *HostDB) ScoreBreakdown(entry modules.HostDBEntry) modules.HostScoreBreakdown {
+	return modules.HostScoreBreakdown{
+		AgeAdjustment:              hdb.lifetimeAdjustments(entry),
+		BurnAdjustment:             1,
+		CollateralAdjustment:       hdb.collateralAdjustments(entry),
+		PriceAdjustment:            hdb.priceAdjustments(entry),
+		StorageRemainingAdjustment: storageRemainingAdjustments(entry),
+		UptimeAdjustment:           hdb.uptimeAdjustments(entry),
+		VersionAdjustment:          versionAdjustments(entry),
```
lukechampine: do these functions need to be methods of the hostDB? If not, is there good reason to believe that they will eventually need to be?

modules/renter/hostdb/online.go
```diff
@@ -0,0 +1,39 @@
...
+	err := hdb.tg.Add()
+	if err != nil {
+		return
+	}
+	defer hdb.tg.Done()
+
+	for {
+		// Every 30 seconds, check the online status and update the online
+		// field.
+		peers := hdb.gateway.Peers()
```
lukechampine: I think we are going to want to add a `gateway.HasInternetConnection` method. The gateway itself needs such a method because otherwise it will prune nodes while offline (I think -- there may already be protections against this). Can be done in a separate PR though

api/hostdb.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/julienschmidt/httprouter"
+)
+
+type (
+	// ExtendedHostDBEntry is an extension to modules.HostDBEntry that includes
+	// the string representation of the public key, otherwise presented as two
+	// fields, a string and a base64 encoded byte slice.
+	ExtendedHostDBEntry struct {
+		modules.HostDBEntry
+		PublicKeyString string `json:"publickeystring"`
```
DavidVorick: yeah the problem with JSON is not the api, it's that the frontend needs some way to let the user pick out hosts individually. That could be done by giving each host their own button, and I'm sure some people will do it that way, but for example in siac you definitely cannot give each host a button. You'd have to give them a number and then somehow know that the number corresponded to something exact in the... well it just doesn't work in a stateless app like siac.

I'm not happy that we sometimes present public keys in the json as base64 and now sometimes as strings, but I definitely feel that strings is the right move, and wish we had started there. Just like we chose for addresses to be hex strings.

doc/API.md
```diff
@@ -737,6 +787,21 @@ lists the status of all files.
...
+
+lists the estimated prices of performing various storage and data operations.
+
+###### JSON Response [(with comments)](/doc/api/Renter.md#json-response-4)
+```javascript
+{
+  "downloadterabyte":      "1234", // hastings
+  "formcontracts":         "1234", // hastings
+  "storageterabytemonth":  "1234", // hastings
+  "uploadterabyte":        "1234", // hastings
```
DavidVorick: I agree that this is a controversial choice. I stand behind it, but not very strongly.

doc/api/HostDB.md
```diff
@@ -166,6 +167,130 @@ any particular order, and the order may change in subsequent calls.
...
+  // A set of scores as determined by the renter. Generally, the host's final
+  // final score is all of the values multiplied together. Modified renters may
+  // have additional criteria that they use to judge a host, or may ignore
+  // certin criteia. In general, these fields should only be used as a loose
+  // guide for the score of a host, as every renter sees the world differently
+  // and uses different metrics to evaluate hosts.
+  "scorebreakdown": {
+    // The multiplier that gets applied to the host based on how long it has
+	// been a host. Older hosts typically have a lower penalty.
+    "ageadjustment":              0.1234,
```
DavidVorick: I just didn't want it to be interpreted as a canonical value. While right now you can get a good idea for host score by multiplying all the floats together, there's no api promise that it will continue to work that way.

I would make the promise, but things will get messy when we add host buckets. (geo stuff, etc.)

modules/renter/hostdb/hostweight.go
```diff
@@ -41,104 +128,227 @@ func (hdb *HostDB) calculateHostWeight(entry modules.HostDBEntry) types.Currency
...
+	// 90%  uptime = 0.51
+	// 85%  uptime = 0.16
+	// 80%  uptime = 0.03
+	// 75%  uptime = 0.005
+	// 70%  uptime = 0.001
+	// 50%  uptime = 0.000002
+	uptimePenalty := float64(1)
+	for i := float64(1); i > uptimeRatio && i > 0.80; i -= 0.01 {
+		uptimePenalty *= uptimeRatio
+	}
```
DavidVorick: originally this was happening with types.Currency instead of float. I guess I also didn't realize the math package operated on floats.

modules/renter/hostdb/hostweight.go
DavidVorick: logging critical errors is the only reason, I believe.

modules/renter/hostdb/online.go
DavidVorick: I agree with this. But, separate PR probably makes the most sense.

modules/renter.go
DavidVorick: it will always be a float. Bools can be '1' for true and '0' for false

lukechampine:
might need to rebase this to fix the failing test

DavidVorick:
it is rebased =/

DavidVorick:
hmm it's also not a parallel test :<

DavidVorick:
so it probably is a more fundamental issue with the test, thing is, it's a consensus test not a gateway test.

DavidVorick:
I'm going to just `t.Skip()` this test, we should come back to it though. I don't like the way that it changes constants mid-test heh (Jordan wrote it).

DavidVorick:
No printlns? NDF.

Printlns? No NDF.

that's life I guess.

Hopefully we can get the build to fail with the println commits I've added. Otherwise this PR is ready to go I think, just need to rebase out the println commits. Shouldn't be too bad.

PR #1566, opened 2017-01-30
Probe handshake

lukechampine:
@DavidVorick, please double-check d47d81f. I'm not sure if we want `err == nil` or `err != nil`, but the comment implies that it should be `err == nil`.

### DIFF
--- a/modules/gateway/nodes.go
+++ b/modules/gateway/nodes.go
@@ -5,7 +5,6 @@ import (
 	"net"
 	"time"
 
-	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/encoding"
 	"github.com/NebulousLabs/Sia/modules"
@@ -35,11 +34,6 @@ func (g *Gateway) addNode(addr modules.NetAddress) error {
 // pingNode verifies that there is a reachable node at the provided address
 // by performing the Sia gateway handshake protocol.
 func (g *Gateway) pingNode(addr modules.NetAddress) error {
-	// Performing the ping during testing does not work.
-	if build.Release == "testing" {
-		return nil
-	}
-
 	// Ping the untrusted node to see whether or not there's actually a
 	// reachable node at the provided address.
 	conn, err := g.dial(addr)
--- a/modules/gateway/nodes.go
+++ b/modules/gateway/nodes.go
@@ -47,10 +47,10 @@ func (g *Gateway) pingNode(addr modules.NetAddress) error {
 	// NOTE: this is a somewhat clunky way of specifying that you didn't
 	// actually want a connection.
 	_, err = connectVersionHandshake(conn, "0.0.0")
-	if err != errPeerRejectedConn {
-		g.log.Debugln("ERROR: version handshake should have resulted in rejection; got", err)
+	if err == errPeerRejectedConn {
+		err = nil // we expect this error
 	}
-	return nil
+	return err
 }
 
 // removeNode will remove a node from the gateway.
--- a/modules/gateway/nodes.go
+++ b/modules/gateway/nodes.go
@@ -2,6 +2,7 @@ package gateway
 
 import (
 	"errors"
+	"fmt"
 	"net"
 	"time"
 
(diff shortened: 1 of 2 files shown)


DavidVorick:
I haven't looked at it in depth yet, but is there some way that we could check with testing that it should be `err == nil` vs `err != nil`?

The probe stuff is probably pretty hard to check, but can you think of testing around that as well?

DavidVorick:
If we can't automate it, we should at least be able to test by spinning up multiple nodes on our machine in real time (using the non-testing constants) and connect to local nodes manually, and also connect to remote nodes. Then we use the API or just normal printlns to check that it's behaving correctly.

I trust that sort of test more than I trust just reading the code. Seems like you are right and that it's currently not working and that you have fixed it, but given that the gateway has been stable I definitely want test-oriented proof that you've solved a bug instead of created one.

Ideally we do find a way to make an automated test for this, but barring that a manual test is sufficient.

modules/gateway/nodes.go
```diff
@@ -53,14 +54,9 @@ func (g *Gateway) managedAddUntrustedNode(addr modules.NetAddress) error {
...
-	encoding.WriteObject(conn, "0.0.0")
-	var reject string
-	err = encoding.ReadObject(conn, &reject, build.MaxEncodedVersionLength)
-	if err != nil {
-		g.log.Debugln("ERROR: version handshake ping terminated unexpectedly:", err)
-	}
-	if reject != "reject" {
-		g.log.Debugln("WARN: peer does not seem to have correctly rejected our ping:", reject)
+	_, err = connectVersionHandshake(conn, "0.0.0")
+	if err != errPeerRejectedConn {
```
DavidVorick: do we know that the rejection is because of the version? If so, the variable name should probably be made more specific.

I guess that if they are returning a Sia error string, they are probably an active Sia node, so this should be good enough.

DavidVorick:
Except for the manual test, I think this code overall looks good.

modules/gateway/nodes.go
lukechampine: during the version handshake, first we send our version ("0.0.0"), and then the peer replies with their version. If our version is unacceptable to the peer (and it always will be, unless they are running custom code), they will reply with the string "reject" instead of their actual version. The connecting node recognizes this as a special case and returns `errPeerRejectedConn`.

Technically the peer can reject our handshake for any reason, not just because our version was bad. For example, they could reject new peers instead of kicking out old ones. I think the variable name is ok for now.

lukechampine:
I confirmed with manual testing that peers are correctly identified as local with the new code.
The old code was actually exhibiting some strange behavior. It relied on the error from `managedAddUntrustedNode` to decide whether to count a peer as local, but that function can fail even if the ping was successful. Specifically, it returns an error if the NetAddress is already in the gateway's peer list. So it was a weird situation where a connectible peer would only be marked as local if they were already in your node list.

I've refactor some of the connect/accept code. The biggest change is that the connecting functions no longer call `managedAddUntrustedNode`. They don't need to, because we already know that the address is connectible. I also removed this call from `managedAcceptConnNewPeer`, the rationale being that the supplied address will eventually be purged anyway. We can easily revert this commit if you think it's the wrong move.

DavidVorick:
I remember that there were also concerns about attackers spoofing their source IP. So they'd connect as one IP but change the headers to make you ping another. Something like that. Regardless, the check was there for a good reason iirc.

modules/gateway/nodes.go
```diff
@@ -32,45 +31,26 @@ func (g *Gateway) addNode(addr modules.NetAddress) error {
...
-	defer g.mu.Unlock()
-	err = g.addNode(addr)
-	if err != nil {
-		return err
+	_, err = connectVersionHandshake(conn, "0.0.0")
+	if err != errPeerRejectedConn {
+		g.log.Debugln("ERROR: version handshake should have resulted in rejection; got", err)
 	}
-	return g.save()
+	return nil
```
DavidVorick: Here you return `nil` even if the error provided by the peer was not the expected peer. Shouldn't you be returning an error so the peer gets rejected?

modules/gateway/peers.go
```diff
@@ -167,8 +167,8 @@ func (g *Gateway) managedAcceptConnOldPeer(conn net.Conn, remoteVersion string)
 		},
 		sess: muxado.Server(conn),
 	})
-
-	return nil
+	g.addNode(addr)
+	return g.save()
```
DavidVorick: I am wondering if we still need to support peers with version less than v1.0.0. I know we do not have many left.

Not this PR, just noticing that we still support them and probably don't have to. Just like we did for v0.3.3.x peers, we should probably start pushing <v1 nodes off of the network.

DavidVorick: It has been about 7 months since releasing v.1.0.0 btw.

modules/gateway/nodes.go
lukechampine: The "different address" case is actually handled earlier, in `acceptConnPortHandshake`. We do indeed reject the peer if the addresses differ.

`pingNode` is only concerned with whether or not we were able to connect to the supplied address. This is not guaranteed -- their ports need to be forwarded. In other words, if we rejected here, only port-forwarded peers would be able to connect to each other.

modules/gateway/peers.go
lukechampine: :+1:

modules/gateway/nodes.go
lukechampine: oops, thought this comment was on `managedAcceptConnNewPeer`. You are right, I think we want to check for the specific error.
However, it's probably alright if the handshake returns `nil`. That just means the remote peer is being a little weird and not rejecting peers that advertise as 0.0.0. There might be a legitimate reason to do so.

PR #1570, opened 2017-02-03
Don't break downloads when renewing

lukechampine:
This fixes a bug that caused files to become unretrievable after renewing or changing the allowance. It also fixes a deadlock in the renter.

### DIFF
--- a/api/renterhost_test.go
+++ b/api/renterhost_test.go
@@ -799,3 +799,107 @@ func TestRenterParallelDelete(t *testing.T) {
 		t.Fatal("file was not deleted properly:", rf.Files)
 	}
 }
+
+// TestRenterRenew sets up an integration test where a renter renews a
+// contract with a host.
+func TestRenterRenew(t *testing.T) {
+	if testing.Short() {
+		t.SkipNow()
+	}
+	t.Parallel()
+	st, err := createServerTester("TestRenterRenew")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer st.server.Close()
+
+	// Announce the host and start accepting contracts.
+	err = st.announceHost()
+	if err != nil {
+		t.Fatal(err)
+	}
+	err = st.acceptContracts()
+	if err != nil {
+		t.Fatal(err)
+	}
+	err = st.setHostStorage()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Set an allowance for the renter, allowing a contract to be formed.
+	allowanceValues := url.Values{}
+	testFunds := "10000000000000000000000000000" // 10k SC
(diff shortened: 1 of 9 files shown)


modules/renter/renter.go
```diff
@@ -177,6 +180,16 @@ func newRenter(cs modules.ConsensusSet, tpool modules.TransactionPool, hdb hostD
...
 	go r.threadedQueueRepairs()
+
+	// Kill workers on shutdown.
+	r.tg.OnStop(func() {
+		id := r.mu.RLock()
+		for _, worker := range r.workerPool {
+			close(worker.killChan)
+		}
+		r.mu.RUnlock(id)
+	})
```
DavidVorick: This code is not needed, as each worker also listens for a tg.StopChan() signal.

But it does add clarity to have it here.

lukechampine: I think there is actually a deadlock.
The worker loop looks like:
```go
for {
		select {
		case <-w.killChan:
			return
		default:
		}
		if w.renter.tg.Add() != nil {
			return
		}
		w.work()
		w.renter.tg.Done()
}
```
But `tg.Add` and `tg.Stop` both lock `tg.bmu`, so `tg.Add` will block until shutdown finishes.

DavidVorick:
This needs a test demonstrating what was broken before + what is fixed now. You can probably set something up in the API using the ecosystem tests.

api/renterhost_test.go
```diff
@@ -799,3 +799,93 @@ func TestRenterParallelDelete(t *testing.T) {
...
+		st.miner.AddBlock()
+	}
+	// Wait for the contract to be renewed.
+	time.Sleep(time.Second)
+
+	// Try downloading the file.
+	downpath := filepath.Join(st.dir, "testdown.dat")
+	err = st.stdGetAPI("/renter/download/test?destination=" + downpath)
+	if err != nil {
+		t.Fatal(err)
```
lukechampine: needless to say, we should have written this test a while ago. It would have caught the bug while you were writing the worker code, instead of much later.

DavidVorick:
Can I ask you to write additional tests around a few more edge cases?

1. Set the allowance to a little longer (so you have 10 blocks wiggle room instead of 5)
2. after setting the allowance, change it to have [change], then let it do it's thing, then try re-downloading
3. mine blocks until the period runs through again (an automatic one to follow the manual one you just did) and download again 
Changes:

{larger allowance, smaller allowance, same allowance}, {larger period, smaller period, same period}

That's 9 scenarios in total. Not sure if easy to make a table, but at the very least should should be able to just copy-paste between them and change the numbers around.

PR #1571, opened 2017-02-05
always decode Currency into new memory

lukechampine:
Decoding into existing memory caused problems with global variables, because decoding into a global affects all other users of the global. This bug didn't rear its head until now because we very rarely decode into global variables. It also is not triggered when decoding into `types.ZeroCurrency`, since the internal slice is `nil`. In other words, it can only affect shared state when decoding modifies the previously-allocated memory of an internal slice.

Originally I thought we would need to need to resort to `build.Critical` here, but that turns out to be unnecessary. Allocating a new Currency for each decode works just fine.

There is one caveat, though, which is that you can still modify a global variable by decoding into it explicitly. For example, this code will change the global value of `types.SiacoinPrecision`:
```go
fmt.Sscan("7", &types.SiacoinPrecision)
```
However, this code is quite obviously wrong, and not much different from writing something like `types.SiacoinPrecision = types.ZeroCurrency`. In both cases it is clear that the variable itself is being modified, so I think this is acceptable.

### DIFF
--- a/types/currency.go
+++ b/types/currency.go
@@ -236,7 +236,9 @@ func (c *Currency) UnmarshalSia(r io.Reader) error {
 	if err != nil {
 		return err
 	}
-	c.i.SetBytes(b)
+	var dec Currency
+	dec.i.SetBytes(b)
+	*c = dec
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


DavidVorick:
so this fixes the Sscan stuff entirely?

lukechampine:
As far as I can tell, yes. The `Scan` method is used for any sort of `fmt.Scan` function (`Sscan`, `Sscanf`, etc.) so all cases should be fixed. Ditto for `UnmarshalSia`.

PR #1572, opened 2017-02-06
add /renter/downloadasync endpoint for ansychronous downloading

johnathanhowell:


api/renter.go
```diff
@@ -334,6 +334,20 @@ func (api *API) renterDownloadHandler(w http.ResponseWriter, req *http.Request,
...
+// renterDownloadAsyncHandler handles the API call to download a file asynchronously.
+func (api *API) renterDownloadAsyncHandler(w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
+	destination := req.FormValue("destination")
+	// Check that the destination path is absolute.
+	if !filepath.IsAbs(destination) {
+		WriteError(w, Error{"destination must be an absolute path"}, http.StatusBadRequest)
+		return
+	}
+
+	go api.renter.Download(strings.TrimPrefix(ps.ByName("siapath"), "/"), destination)
```
lukechampine: One thing to consider here is error handling. The `DownloadInfo` struct doesn't contain an error field, so there's no way for the caller to tell if an async download failed.

We should add an `Err` field. We may also want to modify the `Download` call itself to be async, and then have the synchronous API call do polling. On the other hand, that would probably cause quite a few tests to fail...

DavidVorick:
Not allowed to change the API without updating the docs and writing tests around the new endpoints that demonstrate the endpoints are working correctly.

DavidVorick:
Also for this case in particular I'm wondering if it should be async on the back end as well. We could add a renter func QueueDownload, just like we have QueueUpload.

DavidVorick:
This may conflict with the hostdb-fixes branch, but other than that I don't think it'd be a very big change.

api/renter_test.go
```diff
@@ -30,6 +30,100 @@ func createRandFile(path string, size int) error {
...
+		t.Fatal("the uploading is not succeeding for some reason:", rf.Files[0])
+	}
+
+	// download the file asynchronously
+	downpath := filepath.Join(st.dir, "asyncdown.dat")
+	err = st.getAPI("/renter/downloadasync/test.dat?destination="+downpath, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+
```
DavidVorick: download should be big enough that you can run a check here to verify that the download has not yet completed, before you spin to check that the download has completed. This test is not great atm because you could swap in the synchronous download call above and the test would still pass.

johnathanhowell:
I improved the test per @DavidVorick's suggestions and verified that the test fails with the `/download` endpoint and succeeds with `/downloadasync`

doc/API.md
```diff
@@ -770,6 +771,24 @@ destination
...
+#### /renter/downloadasync/___*siapath___ [GET]
+
+downloads a file to the local filesystem. The call will return immediately.
+
+###### Path Parameters [(with comments)](/doc/api/Renter.md#path-parameters-1)
+```
+*siapath
+```
+
+###### Query String Parameters [(with comments)](/doc/api/Renter.md#query-string-parameters-1)
```
DavidVorick: These links I think point to the /download/ example, need to swap the '1' for a '2'.

Also looks like the rename and upload stuff point to the same path parameters as well, those need to be updated and pointed to the correct place. (may as well since we're worried about it)

doc/api/Renter.md
```diff
@@ -250,6 +251,24 @@ destination
...
+#### /renter/downloadasync/___*siapath___ [GET]
+
+downloads a file to the local filesystem. The call will return immediately.
+
+###### Path Parameters [(with comments)](/doc/api/Renter.md#path-parameters-1)
+```
+*siapath
+```
+
+###### Query String Parameters [(with comments)](/doc/api/Renter.md#query-string-parameters-1)
```
DavidVorick: Same deal here I believe, the links are pointing to the wrong place.

DavidVorick: oh jk this file isn't supposed to have links at all.

johnathanhowell:
I updated the API docs to the proper format

lukechampine:
Adding an `Error` field to `DownloadInfo` is mandatory. I'm not sure how tricky it will be to integrate with the current renter code, but it has to be done. Also, since `Download` can currently return an error, we need to convert it to stick the error in a `DownloadInfo` instead.
I can do this myself and push to this branch if that's easiest.

modules/renter/downloadqueue.go
```diff
@@ -60,6 +60,9 @@ func (r *Renter) DownloadQueue() []modules.DownloadInfo {
 			StartTime:   d.startTime,
 		}
 		downloads[i].Received = atomic.LoadUint64(&d.atomicDataReceived)
+		if d.downloadErr != nil {
+			downloads[i].Error = d.downloadErr.Error()
+		}
```
lukechampine: This is close but there is one edge case to consider: `renter.Download` can return an error, meaning the download will not be placed in the queue, but `downloadasync` will still return success.

johnathanhowell: It looks to me like d.downloadErr will always be set by the worker thread before returning from the `Download` call, though this isn't immediately obvious

johnathanhowell: I added a test to verify this behavior

lukechampine: That is true except for one case: when the file doesn't exist.
There is a TODO in `Download` that reads:
```
// TODO: Eventually just return the channel to the error instead of the
// error itself.
select {
	case err := <-d.downloadFinished:
	return err
```

Perhaps we should tackle this now. There aren't many callsites to update. Then the "file not found" error would look like:
```go
	lockID := r.mu.RLock()
	file, exists := r.files[path]
	r.mu.RUnlock(lockID)
	if !exists {
		ch := make(chan error, 1)
		ch <- errors.New("no file with that path")
		return ch
	}
```

modules/renter/downloadqueue.go
```diff
@@ -32,16 +36,11 @@ func (r *Renter) Download(path, destination string) error {
...
-	case err := <-d.downloadFinished:
-		return err
-	case <-r.tg.StopChan():
-		return errors.New("download interrupted by shutdown")
-	}
+	go func() {
+		errchan <- <-d.downloadFinished
+	}()
+
+	return errchan
```
lukechampine: `return d.downloadFinished` lol

modules/renter/download.go
```diff
@@ -81,8 +81,8 @@ type (
 		siapath           string
 
 		// Syncrhonization tools.
-		downloadFinished chan error
-		mu               sync.Mutex
+		downloadFinished chan struct{}
+		mu               sync.RWMutex
```
lukechampine: leaving this as a normal mutex is actually ok. A RWMutex is advantageous only when you're reading a lot, and in parallel. Since we only read in the call to `Err` (which is not called multiple times in parallel), there is little benefit.

modules/renter/download.go
```diff
@@ -174,7 +182,7 @@ func (d *download) fail(err error) {
 
 	d.downloadComplete = true
 	d.downloadErr = err
-	d.downloadFinished <- err
+	d.downloadFinished <- struct{}{}
```
lukechampine: close instead of sending a struct

modules/renter/download.go
```diff
@@ -263,7 +271,7 @@ func (cd *chunkDownload) recoverChunk() error {
 	if nowComplete {
 		// Signal that the download is complete.
 		cd.download.downloadComplete = true
-		cd.download.downloadFinished <- nil
+		cd.download.downloadFinished <- struct{}{}
```
lukechampine: ditto

lukechampine:
build should pass once you rebase

PR #1574, opened 2017-02-07
Some random stuff

DavidVorick:
My nodes tonight were having trouble finding peers and I was trying to figure out why. I didn't really get to the bottom of it, but I no longer thing that it's because the nodelist is mostly bad. Connection attempts seem to succeed, and then the peer gets tossed after a minute or so, perhaps less.

Maybe there is a bug in the communication code that makes it incompatible with certain versions? Seems like I'm only able to connect to v1.1.0, v1.0.0, and v1.0.1 for any lengthy period of time. Still investigating.

It might also be that all of our peers with open ports and high-propagation are saturated, having 128 peers each. Maybe we should increase the maximum? Or maybe we just announce to #contributors that it's a problem and ask people to open their ports? I don't want to do this though until we confirm that the problem is over-saturation. The errors don't help us on that front though.

### DIFF
--- a/modules/gateway/peersmanager.go
+++ b/modules/gateway/peersmanager.go
@@ -35,6 +35,9 @@ func (g *Gateway) managedPeerManagerConnect(addr modules.NetAddress) {
 		g.mu.Unlock()
 	} else if err != nil {
 		g.log.Debugf("[PMC] [ERROR] [%v] WARN: removing peer because automatic connect failed: %v\n", addr, err)
+		g.mu.Lock()
+		g.removeNode(addr)
+		g.mu.Unlock()
 	} else {
 		g.log.Debugf("[PMC] [SUCCESS] [%v] peer successfully added", addr)
 	}
--- a/modules/gateway/peersmanager.go
+++ b/modules/gateway/peersmanager.go
@@ -8,7 +8,7 @@ import (
 // managedPeerManagerConnect is a blocking function which tries to connect to
 // the input addreess as a peer.
 func (g *Gateway) managedPeerManagerConnect(addr modules.NetAddress) {
-	g.log.Debugln("[PMC] Attempting connection to", addr)
+	g.log.Debugf("[PMC] [%v] Attempting connection", addr)
 	err := g.managedConnect(addr)
 	if err == errPeerExists {
 		// This peer is already connected to us. Safety around the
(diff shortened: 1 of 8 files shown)


DavidVorick:
nevermind I seem to be getting high enough connection diversity. So I'm not sure what exactly is causing all of the failures. But the most common error by far is just 'Session closed'

lukechampine:
I believe `Session Closed` occurs when the remote peer terminates the connection without warning. Unfortunately, we don't have a good idea of why they are terminating. Even if we had access to the logs, there aren't many logging statements that would help identify the cause (I checked the bootstrap linode).

modules/gateway/peersmanager.go
```diff
@@ -30,15 +30,20 @@ func (g *Gateway) managedPeerManagerConnect(addr modules.NetAddress) {
 			// race condition could mean that the peer was disconnected
 			// before this code block was reached.
 			p.Inbound = false
-			g.log.Debugln("[PMC] [SUCCESS] existing peer has been converted to outbound peer:", addr)
-		} else {
-			g.log.Debugln("[PMC] errPeerExists was returned, but by the time the peer was to be set to 'outbound', it no longer existed.")
+			g.log.Debugf("[PMC] [SUCCESS] [%v] existing peer has been converted to outbound peer", addr)
```
lukechampine: we should be more consistent with the use of "peer" vs "node". A node is an address that you know about. A peer is a node that you have connected to. This log statement could simply read "connected to node".

modules/gateway/peersmanager.go
```diff
@@ -30,15 +30,20 @@ func (g *Gateway) managedPeerManagerConnect(addr modules.NetAddress) {
...
 		}
 		g.mu.Unlock()
 	} else if err != nil {
-		g.log.Debugf("[PMC] WARN: automatic connect to %v failed: %v\n", addr, err)
+		g.log.Debugf("[PMC] [ERROR] [%v] WARN: removing peer because automatic connect failed: %v\n", addr, err)
```
lukechampine: ditto, should read "removing node"

siad/main.go
```diff
@@ -132,6 +132,9 @@ Explorer (e):
 
 // main establishes a set of commands and flags using the cobra package.
 func main() {
+	if build.DEBUG {
+		fmt.Println("Running with debugging enabled")
+	}
```
lukechampine: alternatively, we could add `-debug` to the version string printed by `siad version`.

modules/gateway/peersmanager.go
DavidVorick: this is in the peer manager code. In this case an actual existing peer has been converted from an inbound peer to an outbound peer. The logging statement is pedantically correct.

modules/gateway/peersmanager.go
DavidVorick: this logging statement should indeed read "removing node"

siad/main.go
DavidVorick: I'm nervous what that would do for compatibility, do not want to change the API. And, just changing the output of `siad version` is not preferred b/c sometimes you aren't sure if a long-running siad instance is debug or not.

modules/gateway/peersmanager.go
lukechampine: oops, my mistake

siad/main.go
lukechampine: On second consideration, I'm in favor of adding new fields to the `DaemonVersion` struct. Then we'd have the freedom to do this:
```go
type DaemonVersion struct {
    Version string
    Release string // dev, testing, standard
    Debug   bool
    // dynamically added at build time
    CommitSHA string
    BuildDate time.Time
}
```
This would help us diagnose remote machines because we could quickly identify exactly what commit was compiled, and when.

lukechampine: I had a hunch that on Linux, it should be possible to access the binary of a currently-running process, even if the original binary was deleted. And indeed, it is possible, though it takes some work.

1. Get the PID of the running process
2. Open `/proc/$pid/maps` and locate the line containing the path of the original binary
3. Copy the inode number from that line (last column before the path)
4. Get the device descriptor of the filesystem containing the inode
5. Run `debugfs -w $device -R 'ln <$inode> /path/to/file` (may require `sudo`)

The binary is now available at `/path/to/file`. (An important quirk is that the path is relative to the device descriptor mount. So if you had `/home` mounted on `/dev/sda3`, the file will show up in `/home/path/to/file`.)

Pretty cool. Also pretty impractical :P

DavidVorick: Let's do the easy thing for now and we can extend the API in another PR. But having 'version', 'release', and also 'race' would be nice. As well as 'Debug'.

PR #1575, opened 2017-02-08
Prune unreachable bootstrap nodes

lukechampine:
Also fix a bug affecting gateway connections. Previously, the gateway would try to ping the remote address during the connection protocol. The dial timeout is 2 minutes, so the peer can appear unresponsive for up to 2 minutes before RPCs start succeeding. This seems to confuse muxado, leading to dropped connections. When I scanned the bootstrap list, only pre-v1.0.3 peers were consistently reachable. (The ping was introduced in v1.0.3.) However, it is also clearly possible to successfully connect to (and accept connections from) nodes of any version. So it's unclear what exactly causes the failure.


modules/gateway.go
```diff
@@ -22,76 +22,38 @@ var (
...
 			"82.196.11.170:9981",
 			"82.196.5.50:9981",
 			"82.220.99.82:9981",
-			"83.76.19.197:10981",
 			"85.255.197.69:9981",
-			"87.236.27.155:12487",
-			"87.98.216.46:9981",
-			"88.109.8.173:9981",
-			"91.121.183.171:9981",
-			"95.211.203.138:9981",
```
DavidVorick: worth pointing out that every node that remains has been there for 4 months or more, as that's when the list was last added to. But I only added nodes which had already been around for a while.

Which means we have a decent list of long term nodes :)

modules/gateway/peers.go
```diff
@@ -205,10 +201,19 @@ func (g *Gateway) managedAcceptConnNewPeer(conn net.Conn, remoteVersion string)
...
+	// remoteAddr to our node list after accepting the peer.
+	go func() {
+		err := g.pingNode(remoteAddr)
+		if err == nil {
+			g.mu.Lock()
+			g.addNode(remoteAddr)
+			g.save()
+			g.mu.Unlock()
+		}
+	}()
```
DavidVorick: Can you extend the comments here to explain why this needs to be done synchronously instead of asynchronously? I believe you that there's a bug but it should be explained so we don't switch it back in the future.

Ideally there'd be a test to run with it as well, but it seems like the problem doesn't really happen on localhost.

lukechampine:
I've been seeing this NDF a lot more frequently lately:
```
--- FAIL: TestInitialBlockchainDownloadDoneRules (6.62s)
	synchronize_ibd_test.go:415: threadedInitialBlockchainDownload never finished with 2 synced peers and 1 non-synced peer
```
probably worth investigating

DavidVorick:
Was just looking that. I've added it to my notes.

DavidVorick:
it's seeming not an NDF at this point, might need to be fixed as a part of this PR.

PR #1576, opened 2017-02-08
Defrag fixup

DavidVorick:
The biggest change here is the second commit, fixes an NDF plus a lot of unnecessary address use.


DavidVorick:
fourth commit fixes a second wallet ndf as well.

PR #1578, opened 2017-02-10
Hostdb new siac

DavidVorick:
couple of extensions to siac to support the new api endpoints made in the previous PR.

These are helpful for exploring the host network and figuring out what is and is not going according to plan.

### DIFF
--- a/siac/rentercmd.go
+++ b/siac/rentercmd.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/NebulousLabs/Sia/api"
 	"github.com/NebulousLabs/Sia/modules"
+	"github.com/NebulousLabs/Sia/types"
 )
 
 var (
@@ -98,6 +99,13 @@ have a reasonable number (>30) of hosts in your hostdb.`,
 		Short: "Upload a file",
 		Long:  "Upload a file to [path] on the Sia network.",
 		Run:   wrap(renterfilesuploadcmd),
+	}
+
+	renterPricesCmd = &cobra.Command{
+		Use:   "prices",
+		Short: "Display the price of storage and bandwidth",
+		Long:  "Display the estimated prices of storing files, retrieving files, and creating a set of contracts",
+		Run:   wrap(renterpricescmd),
 	}
 )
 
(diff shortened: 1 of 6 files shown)


siac/rentercmd.go
```diff
@@ -411,3 +419,21 @@ func renterfilesuploadcmd(source, path string) {
...
+	if err != nil {
+		die("Could not read the renter prices:", err)
+	}
+
+	fmt.Println("Renter Prices (estimated):")
+	w := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
+	fmt.Fprintln(w, "\tFees for Creating a Set of Contracts:\t", rpg.FormContracts.Div(types.SiacoinPrecision))
+	fmt.Fprintln(w, "\tDownload 1 TB:\t", rpg.DownloadTerabyte.Div(types.SiacoinPrecision))
+	fmt.Fprintln(w, "\tStore 1 TB for 1 Month:\t", rpg.StorageTerabyteMonth.Div(types.SiacoinPrecision))
+	fmt.Fprintln(w, "\tUpload 1 TB:\t", rpg.UploadTerabyte.Div(types.SiacoinPrecision))
```
lukechampine: don't we have a method for formatting currency values?

DavidVorick:
this is ready

DavidVorick:
Went ahead and made a few additional changes:

1. /hostdb/active will now only return hosts who are accepting contracts
2. dropped the 'siac hostdb all' command
3. added support for 'siac hostdb -v', which will display the set of offline hosts, inactive hosts, and active hosts
4. Removed pubkey from 'siac hostdb' display
5. 'siac hostdb -v' will display both pubkey and uptime
6. entries are given numbers when printed out. Note that they start high and count down to 1, because the most valuable hosts are displayed at the bottom.

DavidVorick:
I am considering expanding the /hostdb endpoint to have not just `Hosts` as a field (all hosts ranked by weight), but also 'InactiveHosts' and 'ActiveHosts', also ranked by weight, including support for the arg that limits the number of hosts returned.

This would make life easier for api consumer such as siapulse.com

Haven't don't anything with it yet though.

lukechampine:
what determines whether a host is active? If it can be easily determined from one of the struct fields, I don't see much reason to split it out separately in the API. If it requires looping over the scan history or something, then there's more of a case for it.

DavidVorick:
active: most recent scan was successful && accepting contracts
inactive: most recent scan was successful
offline: everything else

lukechampine:
Could we add an `Active` field to `ExtendedHostDBEntry`? I think that is preferable to grouping things server-side.

DavidVorick:
let's do it in a different PR though. lol I want these siac commands on master branch b/c I am using them a lot for live testing.

DavidVorick:
But, we could add the fields 'Active bool' and 'Online bool' and 'Forsaken bool' or something like that. (forsaken = offline for so long that we don't think we will ever see them again)

PR #1579, opened 2017-02-10
Continued hostdb improvements

DavidVorick:
This is a very big upgrade for the hostdb in terms of fuctionality.

1. Upgrade time has been reduced from 10k-50k seconds to 100-1k seconds. Definitely a big deal.
2. When scanning a bunch of hosts from the blockchain, duplicates are ignored.
3. The hostdb more gracefully handles ip address changes than before. My hostdb is now finding 87 ipv4 hosts, and there are at least 4 known ipv6 hosts, meaning that the total number of discoverable hosts has been increased to at least 91, which is substantially higher than any number previously available.
4. The logging is slightly better.

There's no automated testing for this yet, but I've been doing manual testing all night, and I'm very pleased with the results.

### DIFF
--- a/modules/renter/hostdb/hostdb.go
+++ b/modules/renter/hostdb/hostdb.go
@@ -17,6 +17,7 @@ package hostdb
 
 import (
 	"errors"
+	"fmt"
 	"os"
 	"path/filepath"
 	"sync"
@@ -110,6 +111,12 @@ func newHostDB(g modules.Gateway, cs modules.ConsensusSet, persistDir string, de
 		return nil, err
 	}
 	hdb.log = logger
+	hdb.tg.AfterStop(func() {
+		if err := hdb.log.Close(); err != nil {
+			// Resort to println as the logger is in an uncertain state.
+			fmt.Println("Failed to close the hostdb logger:", err)
+		}
+	})
 
 	// The host tree is used to manage hosts and query them at random.
 	hdb.hostTree = hosttree.New(hdb.calculateHostWeight)
@@ -119,6 +126,13 @@ func newHostDB(g modules.Gateway, cs modules.ConsensusSet, persistDir string, de
 	if err != nil && !os.IsNotExist(err) {
 		return nil, err
 	}
+
+	// Loading is complete, establish the save loop.
+	saveLoopShutdown := make(chan struct{})
+	go hdb.threadedSaveLoop(saveLoopShutdown)
+	hdb.tg.OnStop(func() {
+		close(saveLoopShutdown)
+	})
 
 	// Don't perform the remaining startup in the presence of a quitAfterLoad
 	// disruption.
--- a/modules/renter/hostdb/hostdb.go
+++ b/modules/renter/hostdb/hostdb.go
@@ -62,8 +62,10 @@ type HostDB struct {
 
 	// the scanPool is a set of hosts that need to be scanned. There are a
 	// handful of goroutines constantly waiting on the channel for hosts to
-	// scan.
+	// scan. The scan map is used to prevent duplicates from entering the scan
+	// pool.
 	scanList []modules.HostDBEntry
+	scanMap  map[string]struct{}
 	scanPool chan modules.HostDBEntry
 	scanWait bool
 	online   bool
@@ -96,6 +98,7 @@ func newHostDB(g modules.Gateway, cs modules.ConsensusSet, persistDir string, de
 		gateway:    g,
 		persistDir: persistDir,
 
+		scanMap:  make(map[string]struct{}),
 		scanPool: make(chan modules.HostDBEntry, scanPoolSize),
 	}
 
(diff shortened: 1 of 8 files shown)


modules/renter/hostdb/persist.go
```diff
@@ -75,3 +71,27 @@ func (hdb *HostDB) load() error {
 	hdb.lastChange = data.LastChange
 	return nil
 }
+
+// threadedSaveLoop saves the hostdb to disk every 2 minutes, also saving when
+// given the shutdown signal.
+func (hdb *HostDB) threadedSaveLoop(shutChan chan struct{}) {
```
lukechampine: why can't this just use `tg.StopChan`?

modules/renter/hostdb/persist.go
```diff
@@ -75,3 +71,27 @@ func (hdb *HostDB) load() error {
...
+			}
+			return
+		case <-time.After(saveFrequency):
+			hdb.mu.Lock()
+			err := hdb.saveSync()
+			hdb.mu.Unlock()
+			if err != nil {
+				hdb.log.Println("Difficulties saving the hostdb:", err)
+			}
+		}
```
lukechampine: maybe cleaner:
```go
for {
    hdb.saveSync() // plus locking, logging

    select {
    case <-shutChan:
        return
    case <-time.After(saveFrequency):
    }
}
```
`time.After` leaks, but it's nbd

modules/renter/hostdb/persist.go
```diff
@@ -75,3 +71,27 @@ func (hdb *HostDB) load() error {
...
+	}
```
lukechampine: maybe cleaner:
```go
for {
    var shutdown bool
    select {
    case <-shutChan:
        shutdown = true
    case <-time.After(saveFrequency):
    }
    
    hdb.saveSync() // plus locking, logging

    if shutdown {
        return
    }
}
```

modules/renter/hostdb/persist.go
DavidVorick: Want to make sure the final save happens after everything else has shut down.

lukechampine: could you stick it in a `tg.AfterStop` function then?

DavidVorick: I could but I felt this was cleaner. NBD to me either way.

DavidVorick: I plan on repeating this pattern in other modules so we should figure out how we want to code it.

I think repeating this in other places will speed up IBD

lukechampine: IMO if you want to express "always save right before shutdown," the clearest way to do it is in `AfterStop` (or in the `Close` function itself :P)

This seems like a good change to roll out to other modules. Losing 2 minutes of data isn't a big deal as long as it can be recovered by resubscribing. There isn't much data that we absolutely have to `saveSync` as soon as it arrives. Just contracts and some wallet stuff I imagine.

DavidVorick: heh the code I wrote is incorrect anyway, because the `Stop` function isn't blocked, it'll continue immediately after closing the channel and not wait for the save to happen.

I'll switch it to your method.

modules/renter/hostdb/hostweight.go
```diff
@@ -47,10 +44,7 @@ var (
 
 	// priceExponentiation is the number of times that the weight is divided by
 	// the price.
-	//
-	// NOTE: Changing this value upwards means that the baseWeight will need to
-	// be increased.
-	priceExponentiation = 4
+	priceExponentiation = 5
```
DavidVorick: I didn't want to do this because it makes certain attacks easier, but we've taken steps even then to limit them as much as possible. More importantly though, we don't have nearly enough hosts, and without the filter being this sharp some expensive ones can still make the top-50.

modules/renter/hostdb/update.go
```diff
@@ -42,7 +42,15 @@ func (hdb *HostDB) insertBlockchainHost(host modules.HostDBEntry) {
...
+		// Replace the netaddress with the most recently announced netaddress.
+		// Also replace the FirstSeen value with the current block height if
+		// the first seen value has been set to zero (no hosts actually have a
+		// first seen height of zero, but due to rescans hosts can end up with
+		// a zero-value FirstSeen field.
 		oldEntry.NetAddress = host.NetAddress
+		if oldEntry.FirstSeen == 0 {
+			oldEntry.FirstSeen = hdb.blockHeight
+		}
 		err := hdb.hostTree.Modify(oldEntry)
```
lukechampine: off topic, but `Modify` looks weird to me since it only takes one argument. Maybe `Update` would be more clear.

modules/renter/hostdb/hostdb.go
```diff
@@ -119,6 +123,17 @@ func newHostDB(g modules.Gateway, cs modules.ConsensusSet, persistDir string, de
 	if err != nil && !os.IsNotExist(err) {
 		return nil, err
 	}
+	hdb.tg.AfterStop(func() {
+		hdb.mu.Lock()
+		err := hdb.saveSync()
+		hdb.mu.Unlock()
```
lukechampine: just wondering, is it necessary to lock in `AfterStop` functions? In theory this should be the only goroutine running, right?

PR #1581, opened 2017-02-13
use cachedRevision if renew fails

lukechampine:
I'm getting frequent deadlocks when testing the api package locally. Seems like a shutdown thing.


lukechampine:
Might have to tackle the deadlock in a different PR. I haven't been able to make much headway

PR #1583, opened 2017-02-14
add validateSiapath to validate renter siapaths on upload

johnathanhowell:
This PR fixes an issue where by specifying a siapath with `../`, users could write .sia files outside of their renter directory anywhere on their filesystem. I also moved the existing validation to the `validateSiapath` function and added testing around it.

### DIFF
--- a/modules/renter/upload.go
+++ b/modules/renter/upload.go
@@ -49,7 +49,7 @@ var (
 // ../ is disallowed to prevent directory traversal,
 // and paths must not begin with / or be empty.
 func validateSiapath(siapath string) error {
-	if strings.HasPrefix(siapath, "/") {
+	if strings.HasPrefix(siapath, "/") || strings.HasPrefix(siapath, "./") {
 		return errors.New("nicknames cannot begin with /")
 	}
 
--- a/modules/renter/upload.go
+++ b/modules/renter/upload.go
@@ -61,6 +61,10 @@ func validateSiapath(siapath string) error {
 		return errors.New("directory traversal is not allowed")
 	}
 
+	if strings.Contains(siapath, "./") {
+		return errors.New("siapath contains invalid characters")
+	}
+
 	return nil
 }
 
--- a/modules/renter/upload.go
+++ b/modules/renter/upload.go
@@ -45,15 +45,31 @@ var (
 	}()
 )
 
+// validateSiapath checks that a Siapath is a legal filename.
+// ../ is disallowed to prevent directory traversal,
+// and paths must not begin with / or be empty.
+func validateSiapath(siapath string) error {
+	if strings.HasPrefix(siapath, "/") {
+		return errors.New("nicknames cannot begin with /")
+	}
+
+	if siapath == "" {
+		return ErrEmptyFilename
+	}
+
+	if strings.Contains(siapath, "../") {
+		return errors.New("directory traversal is not allowed")
+	}
+
+	return nil
+}
+
 // Upload instructs the renter to start tracking a file. The renter will
 // automatically upload and repair tracked files using a background loop.
 func (r *Renter) Upload(up modules.FileUploadParams) error {
 	// Enforce nickname rules.
-	if strings.HasPrefix(up.SiaPath, "/") {
-		return errors.New("nicknames cannot begin with /")
-	}
-	if up.SiaPath == "" {
-		return ErrEmptyFilename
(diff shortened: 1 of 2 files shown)


DavidVorick:
This needs tests

johnathanhowell:
oh, I wrote tests but did not check them in lol

modules/renter/upload_test.go
```diff
@@ -0,0 +1,33 @@
...
+	}{
+		{"valid/siapath", true},
+		{"../../../directory/traversal", false},
+		{"testpath", true},
+		{"valid/siapath/../with/directory/traversal", false},
+		{"validpath/test", true},
+		{"..validpath/..test", true},
+		{"./invalid/path", false},
+		{"test/path", true},
+		{"/leading/slash", false},
```
DavidVorick: do we currently support/require a leading slash? I thought we needed it?

modules/renter/upload_test.go
```diff
@@ -0,0 +1,33 @@
...
+// TestRenterSiapathValidate verifies that the validateSiapath function correctly validates SiaPaths.
+func TestRenterSiapathValidate(t *testing.T) {
+	var pathtests = []struct {
+		in    string
+		valid bool
+	}{
+		{"valid/siapath", true},
+		{"../../../directory/traversal", false},
+		{"testpath", true},
+		{"valid/siapath/../with/directory/traversal", false},
```
DavidVorick: is this something we want to be false? Seems okay to me, though I can see fears regarding edges cases as overriding the convenience of having it.

modules/renter/upload_test.go
johnathanhowell: we currently require that siapaths do not have a leading slash

DavidVorick: (leading slash would be parsed as the root siapath directory, not the root system directory)

(will be more useful when we add proper directory support)

DavidVorick: oh. I guess we will leave it for now and consider adding it back in the future.

modules/renter/upload_test.go
johnathanhowell: If we allow some directory traversal, it opens up the possibility of users inadvertantly overwriting their siafiles, take for example:

`folder/siafile.png`
`folder/folder2/../siafile.png`

these two siapaths resolve on the filesystem to the same file and would cause an overwrite condition

I think it's safer just to disallow directory traversal entirely, I don't see a good reason to have it.

modules/renter/upload.go
```diff
@@ -45,15 +45,31 @@ var (
...
+		return errors.New("nicknames cannot begin with /")
+	}
+
+	if siapath == "" {
+		return ErrEmptyFilename
+	}
+
+	if strings.Contains(siapath, "../") {
+		return errors.New("directory traversal is not allowed")
+	}
```
lukechampine: A cleaner way might be:
```go
if path.Clean(siapath) != siapath {
    return errors.New("illegal path")
}
```
however, this will also reject paths like `///foo///bar` (-> `/foo/bar`). Not sure if this is desirable.

johnathanhowell: manually testing this showed that path.Clean("../../../directory/traversal") == "../../../directory/traversal"

modules/renter/upload_test.go
lukechampine: I don't think it's all that convenient anyway. Why would a user specify such a path? Personally I only use `..` when I want to reference a directory relative to the one I'm in now, but that doesn't apply to siapaths.

modules/renter/upload.go
lukechampine: bogus. I guess that behavior makes sense though.

lukechampine: can we catch `./` as well though? Seems bad that `foo/./bar` can collide with `foo/bar`

johnathanhowell: yeah, we should probably scan the whole string for it using strings.Contains

johnathanhowell: Additionally, how does the renter deal with siapaths that have backslashes instead of forwardslashes?

modules/renter/upload_test.go
DavidVorick: It is convenient if you are doing directory traversal, or have otherwise written a script to do stuff

modules/renter/upload.go
DavidVorick: I don't think they are allowed

PR #1584, opened 2017-02-14
Testing around dynamic ip addresses

DavidVorick:
I did some testing to make sure the renter+host+contractor are all following dynamic IPs correctly. The contractor was not, but I fixed that with some very small code changes. One of the tests I wrote I could not get to pass, even when I was keeping the IP static. @lukechampine could you look at the test I wrote with t.Skip() and see if there's any obvious reason that the file is not downloadable at the end of the test?

With the exception of that final test (exploring that renew functions correctly with dynamic IPs), I think it's safe to establish that the renter can now follow around hosts with dynamic ip addresses. This should dramatically increase the reliability on our network, I believe that we were shedding a dozen or more hosts (out of 42) due to this, and now we won't be losing them.

### DIFF
--- a/modules/renter/contractor/downloader.go
+++ b/modules/renter/contractor/downloader.go
@@ -167,6 +167,13 @@ func (c *Contractor) Downloader(id types.FileContractID) (_ Downloader, err erro
 		}
 	}
 
+	// Update the contract to the most recent net address for the host.
+	host, exists := c.hdb.Host(contract.HostPublicKey)
+	if !exists {
+		return nil, errors.New("unable to find the contract's host in the host database")
+	}
+	contract.NetAddress = host.NetAddress
+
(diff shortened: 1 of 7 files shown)


modules/renter/contractor/downloader.go
```diff
@@ -167,6 +167,13 @@ func (c *Contractor) Downloader(id types.FileContractID) (_ Downloader, err erro
 		}
 	}
 
+	// Update the contract to the most recent net address for the host.
+	host, exists := c.hdb.Host(contract.HostPublicKey)
```
lukechampine: we already get the host above lol

DavidVorick:
Manage to get the test suite working. Extended it to support a wider range of testing, and also got everything in working shape.

PR #1585, opened 2017-02-16
faster, compatible encoding scheme for []crypto.Hash

DavidVorick:
couldn't help myself. This should be a lot smaller (base64 instead of base16, no separators), compatible (loads the old method if the new load attempt fails), and faster (only encoding 1 object instead of N).

Benchmarking on a production Sia client.

### DIFF
--- a/modules/renter.go
+++ b/modules/renter.go
@@ -1,6 +1,7 @@
 package modules
 
 import (
+	"encoding/json"
 	"io"
 	"time"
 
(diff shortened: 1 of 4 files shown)


DavidVorick:
result: it's smaller but also more CPU intensive, so a net negative.

I want to look into optimizing it a bit more.

DavidVorick:
I added some benchmarks and... I was wrong. It is faster, by more than a factor of two. I'm not sure what the server was doing that made me think it was slower, but I can confirm that the server is actually running faster now. Before the update it was going at maybe 17mbps and now it's doing about 27mbps... for now at least.

Not a significant boost by any means though. I'm looking forward to the PR that avoids encoding 100% of the metadata every update. :)

DavidVorick:
:< I am seeing the repeated situation where all tests pass in the API but then for some reason the suite gets a fail? Not sure what could be going on here, but maybe we should ask upstream.

modules/renter.go
```diff
@@ -147,6 +148,44 @@ func (s HostDBScans) Len() int           { return len(s) }
...
+type MerkleRootSet []crypto.Hash
+
+// MarshalJSON defines a JSON encoding for a MerkleRootSet.
+func (mrs MerkleRootSet) MarshalJSON() ([]byte, error) {
+	// Copy the whole array into a giant byte slice and then encode that.
+	fullBytes := make([]byte, crypto.HashSize*len(mrs))
+	for i := range mrs {
+		copy(fullBytes[i*crypto.HashSize:(i+1)*crypto.HashSize], mrs[i][:])
+	}
+	return json.Marshal(fullBytes)
```
lukechampine: I tried speeding this up by doing the marshaling manually, but it had little impact. The actual bottleneck here is that we call `json.Marshal(MerkleRootSet)` instead of `MerkleRootSet.MarshalJSON()`. In a benchmark of marshaling 100 hashes, the latter is about 4x faster.

lukechampine:
Won't this be negated by the move to log-based persist, since we only add one hash at a time? If we switch a single root's encoding to base64, it's not as compact, but it would be applicable to single-hash updates as well.

modules/renter.go
DavidVorick: lol I also tried doing the base64 manually. Probably should have mentioned that and saved you the trouble.

We could speed it up a lot if we defined MarshalJSON on the whole struct, though that makes compatibility more annoying. But then we get to bypass all of the reflection.

DavidVorick:
we'd also lose most of the speed boost I think, most of the speed increase I believe is the fact that we're just working with one slice.

But also, I don't think we'll be moving away from storing them as 1 giant base64 all that soon. First we need a WAL, then we need confidence that it's the biggest bottleneck, then we need some scheme for storing them on disk in a way that is easy to random-access and also not super intrusive.

I'm guessing we will merge host+wallet pr's before we get that far

DavidVorick:
This PR is not too much code and provides basically an immediate 35% speedup to anyone storing more than 100GB of data, I think that's a good reason to merge it.

PR #1587, opened 2017-02-18
Contractor journal

lukechampine:
My biggest PR in a while.

Recently it became clear that the contractor was not scaling well when storing lots of data. This was because it re-marshaled its entire persist data on every save -- and we save every time we revise a contract. Our tests showed that this amounted to about 35MB of data being encoded and written to disk (with fsync).

To address this, I wrote a json-based transaction journal. The journal records updates to the contracts instead of rewriting the whole contract. These updates are applied to the original object at regular intervals. The primary goal is to reduce latency when uploading/downloading, so for now, only those actions take advantage of the journal.

Things got a little ugly because the contractor persist object changed: it now uses string-keyed maps instead of slices in a few places.

Left to do:
- fix bugs in journal (see TODOs in journal.go)
- add metadata to journal format
- test extensively on antfarm before even attempting with real contracts

### DIFF
--- a/api/renterhost_test.go
+++ b/api/renterhost_test.go
@@ -1011,3 +1011,126 @@ func TestRenterAllowance(t *testing.T) {
 		}
 	}
 }
+
+// TestHostAndRentReload sets up an integration test where a host and renter
+// do basic uploads and downloads, with an intervening shutdown+startup.
+func TestHostAndRentReload(t *testing.T) {
+	if testing.Short() {
+		t.SkipNow()
+	}
+	t.Parallel()
+	st, err := createServerTester("TestHostAndRentReload")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Announce the host and start accepting contracts.
+	err = st.announceHost()
+	if err != nil {
+		t.Fatal(err)
+	}
+	err = st.acceptContracts()
+	if err != nil {
+		t.Fatal(err)
+	}
+	err = st.setHostStorage()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Set an allowance for the renter, allowing a contract to be formed.
+	allowanceValues := url.Values{}
+	testFunds := "10000000000000000000000000000" // 10k SC
+	testPeriod := "10"
+	allowanceValues.Set("funds", testFunds)
+	allowanceValues.Set("period", testPeriod)
+	err = st.stdPostAPI("/renter", allowanceValues)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Create a file.
+	path := filepath.Join(st.dir, "test.dat")
+	err = createRandFile(path, 1024)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Upload the file to the renter.
+	uploadValues := url.Values{}
+	uploadValues.Set("source", path)
+	err = st.stdPostAPI("/renter/upload/test", uploadValues)
+	if err != nil {
+		t.Fatal(err)
+	}
+	// Only one piece will be uploaded (10% at current redundancy).
+	var rf RenterFiles
+	for i := 0; i < 200 && (len(rf.Files) != 1 || rf.Files[0].UploadProgress < 10); i++ {
+		st.getAPI("/renter/files", &rf)
+		time.Sleep(100 * time.Millisecond)
+	}
+	if len(rf.Files) != 1 || rf.Files[0].UploadProgress < 10 {
+		t.Fatal("the uploading is not succeeding for some reason:", rf.Files[0])
+	}
+
+	// Try downloading the file.
+	downpath := filepath.Join(st.dir, "testdown.dat")
+	err = st.stdGetAPI("/renter/download/test?destination=" + downpath)
+	if err != nil {
+		t.Fatal(err)
+	}
+	// Check that the download has the right contents.
+	orig, err := ioutil.ReadFile(path)
+	if err != nil {
+		t.Fatal(err)
+	}
+	download, err := ioutil.ReadFile(downpath)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if bytes.Compare(orig, download) != 0 {
+		t.Fatal("data mismatch when downloading a file")
+	}
+
+	// The renter's downloads queue should have 1 entry now.
+	var queue RenterDownloadQueue
+	if err = st.getAPI("/renter/downloads", &queue); err != nil {
+		t.Fatal(err)
+	}
+	if len(queue.Downloads) != 1 {
+		t.Fatalf("expected renter to have 1 download in the queue; got %v", len(queue.Downloads))
+	}
+
+	// close and reopen the server
+	err = st.server.Close()
+	if err != nil {
+		t.Fatal(err)
+	}
+	st, err = assembleServerTester(st.walletKey, st.dir)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer st.server.Close()
+	err = st.announceHost()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Try downloading the file.
+	err = st.stdGetAPI("/renter/download/test?destination=" + downpath)
+	if err != nil {
+		t.Fatal(err)
+	}
+	// Check that the download has the right contents.
+	orig, err = ioutil.ReadFile(path)
+	if err != nil {
+		t.Fatal(err)
+	}
+	download, err = ioutil.ReadFile(downpath)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if bytes.Compare(orig, download) != 0 {
+		t.Fatal("data mismatch when downloading a file")
+	}
+}
--- a/modules/renter/contractor/contractor.go
+++ b/modules/renter/contractor/contractor.go
@@ -28,8 +28,8 @@ var (
 // as a safeguard against desynchronizing with the host.
 // TODO: save a diff of the Merkle roots instead of all of them.
 type cachedRevision struct {
-	Revision    types.FileContractRevision
-	MerkleRoots modules.MerkleRootSet
+	Revision    types.FileContractRevision `json:"revision"`
+	MerkleRoots modules.MerkleRootSet      `json:"merkleroots"`
 }
 
 // A Contractor negotiates, revises, renews, and provides access to file
--- a/modules/renter/contractor/contractor.go
+++ b/modules/renter/contractor/contractor.go
@@ -124,6 +124,12 @@ func (c *Contractor) ResolveID(id types.FileContractID) types.FileContractID {
 	return id
 }
 
+// Close closes the Contractor.
+func (c *Contractor) Close() error {
+	c.log.Close()
+	return c.persist.Close()
+}
+
 // New returns a new Contractor.
 func New(cs consensusSet, wallet walletShim, tpool transactionPool, hdb hostDB, persistDir string) (*Contractor, error) {
 	// Check for nil inputs.
--- a/modules/renter/contractor/dependencies.go
+++ b/modules/renter/contractor/dependencies.go
@@ -68,9 +68,8 @@ type walletBridge struct {
 func (ws *walletBridge) NextAddress() (types.UnlockConditions, error) { return ws.w.NextAddress() }
 func (ws *walletBridge) StartTransaction() transactionBuilder         { return ws.w.StartTransaction() }
 
-// stdPersist implements the persister interface via jj.OpenJournal and
-// jj.CheckPoint. The filename required by these functions is internal to
-// stdPersist.
+// stdPersist implements the persister interface via the journal type. The
+// filename required by these functions is internal to stdPersist.
 type stdPersist struct {
 	journal  *journal
 	filename string
--- a/modules/renter/contractor/dependencies.go
+++ b/modules/renter/contractor/dependencies.go
@@ -53,7 +53,7 @@ type (
 
 	persister interface {
 		save(contractorPersist) error
-		update(...interface{}) error
+		update(...journalUpdate) error
 		load(*contractorPersist) error
 	}
 )
@@ -75,21 +75,19 @@ type stdPersist struct {
 	filename string
 }
 
-func (p stdPersist) save(data contractorPersist) error {
+func (p *stdPersist) save(data contractorPersist) error {
 	if p.journal == nil {
 		var err error
-		p.journal, err = openJournal(p.filename, data)
-		return err
+		p.journal, err = newJournal(p.filename)
+		if err != nil {
+			return err
+		}
 	}
 	return p.journal.checkpoint(data)
 }
 
-func (p *stdPersist) update(us ...interface{}) error {
-	var updates []journalUpdate
-	for i := 0; i < len(us); i += 2 {
-		updates = append(updates, newJournalUpdate(us[i].(string), us[i+1].(interface{})))
-	}
-	return p.journal.update(updates)
+func (p *stdPersist) update(us ...journalUpdate) error {
+	return p.journal.update(us)
 }
 
 func (p *stdPersist) load(data *contractorPersist) error {
--- a/modules/renter/contractor/dependencies.go
+++ b/modules/renter/contractor/dependencies.go
@@ -78,10 +78,8 @@ type stdPersist struct {
 func (p *stdPersist) save(data contractorPersist) error {
 	if p.journal == nil {
 		var err error
-		p.journal, err = newJournal(p.filename)
-		if err != nil {
-			return err
-		}
+		p.journal, err = newJournal(p.filename, data)
+		return err
 	}
 	return p.journal.checkpoint(data)
 }
@@ -95,7 +93,11 @@ func (p *stdPersist) load(data *contractorPersist) error {
 	p.journal, err = openJournal(p.filename, data)
 	if err != nil {
 		// try loading old persist
-		return loadv110persist(filepath.Dir(p.filename), data)
+		err = loadv110persist(filepath.Dir(p.filename), data)
+		if err != nil {
+			return err
+		}
+		p.journal, err = newJournal(p.filename, *data)
 	}
 	return err
 }
--- a/modules/renter/contractor/dependencies.go
+++ b/modules/renter/contractor/dependencies.go
@@ -4,7 +4,6 @@ import (
 	"path/filepath"
 
 	"github.com/NebulousLabs/Sia/modules"
-	"github.com/NebulousLabs/Sia/persist"
 	"github.com/NebulousLabs/Sia/types"
 )
 
@@ -54,7 +53,7 @@ type (
 
 	persister interface {
 		save(contractorPersist) error
-		saveSync(contractorPersist) error
+		update(...interface{}) error
 		load(*contractorPersist) error
 	}
 )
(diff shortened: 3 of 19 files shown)


DavidVorick:
Hooray. Eventually the hope is that the journal will turn into a full write ahead log for managing all sorts of complex persist updates that hopefully iteratively boost scalability and speed as we add them. I think this is a pretty good start, and while it may not iterate for a while, I'm looking forward to testing it, using it, and expanding it.

api/renterhost_test.go
```diff
@@ -1011,3 +1011,126 @@ func TestRenterAllowance(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	download, err = ioutil.ReadFile(downpath)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if bytes.Compare(orig, download) != 0 {
+		t.Fatal("data mismatch when downloading a file")
+	}
+}
```
DavidVorick: Would like to see additional tests that put the restart in various places.

modules/renter/contractor/contractor.go
```diff
@@ -28,8 +28,8 @@ var (
 // as a safeguard against desynchronizing with the host.
 // TODO: save a diff of the Merkle roots instead of all of them.
```
DavidVorick: has this TODO been resolved?

modules/renter/contractor/dependencies.go
```diff
@@ -68,32 +67,43 @@ type walletBridge struct {
...
 
-func (p *stdPersist) save(data contractorPersist) error {
-	return persist.SaveFile(p.meta, data, p.filename)
+func (p stdPersist) save(data contractorPersist) error {
+	if p.journal == nil {
+		var err error
+		p.journal, err = openJournal(p.filename, data)
+		return err
+	}
+	return p.journal.checkpoint(data)
```
DavidVorick: This is a sync operation? I would like to move the project in the direction of always syncing.

modules/renter/contractor/dependencies.go
```diff
@@ -68,32 +67,43 @@ type walletBridge struct {
...
 }
 
-func (p *stdPersist) saveSync(data contractorPersist) error {
-	return persist.SaveFileSync(p.meta, data, p.filename)
+func (p *stdPersist) update(us ...interface{}) error {
+	var updates []journalUpdate
+	for i := 0; i < len(us); i += 2 {
```
DavidVorick: need a length check here

DavidVorick: and probably a build.Critical

modules/renter/contractor/dependencies.go
```diff
@@ -68,32 +67,43 @@ type walletBridge struct {
...
+	return p.journal.update(updates)
 }
 
 func (p *stdPersist) load(data *contractorPersist) error {
-	return persist.LoadFile(p.meta, data, p.filename)
+	var err error
+	p.journal, err = openJournal(p.filename, data)
+	if err != nil {
+		// try loading old persist
+		return loadv110persist(filepath.Dir(p.filename), data)
```
DavidVorick: I am wondering if this should be thrown into the compatibility package as something that can do an in-place upgrade from v052 to v111 - meta version is v052 btw

modules/renter/contractor/dependencies.go
```diff
@@ -68,32 +67,43 @@ type walletBridge struct {
...
 }
 
 func newPersist(dir string) *stdPersist {
 	return &stdPersist{
-		meta: persist.Metadata{
-			Header:  "Contractor Persistence",
-			Version: "0.5.2",
-		},
-		filename: filepath.Join(dir, "contractor.json"),
+		filename: filepath.Join(dir, "contractor.journal"),
```
DavidVorick: need to add a version number here for sure.

DavidVorick: preferably a persist.Metadata - they help when browsing butchered files as well because they have titles

modules/renter/contractor/dependencies.go
lukechampine: what I would really like to do is move all the existing COMPAT code out of `load` and into `loadv110persist`. But there's a dilemma: the COMPAT code requires access to the contractor (since it does stuff like rescanning), so it can't happen in this method. But it also can't happen in `c.load`, because `c.load` doesn't have access to the persist directory. (It is hidden by the `persister` dependency.)

modules/renter/contractor/dependencies.go
lukechampine: yes, checkpoint syncs and atomically replaces the file

modules/renter/contractor/contractor.go
lukechampine: hmm, I guess it has. woo

modules/renter/contractor/dependencies.go
DavidVorick: oh yeah I guess that throws a wrench into things. we can probably at least have a compat.go though

modules/renter/contractor/downloader.go
```diff
@@ -79,7 +80,12 @@ func (hd *hostDownloader) Sector(root crypto.Hash) ([]byte, error) {
 
 	hd.contractor.mu.Lock()
 	hd.contractor.contracts[contract.ID] = contract
-	hd.contractor.saveSync()
+	cpath := fmt.Sprintf("contracts.%s", contract.ID.String())
+	hd.contractor.persist.update(
+		cpath+".lastrevision", contract.LastRevision,
+		cpath+".lastrevisiontxn", contract.LastRevisionTxn,
+		cpath+".downloadspending", contract.DownloadSpending,
+	)
```
DavidVorick: Is there any way we can be certain that this covers the whole spectrum of changes?

We're sort of changing paradigms here from something where you can expect all fields to be saved simultaneously to where you have to be explicit any time that you change something.

I know that it's verbose, but for safety reasons I think I'd prefer the updates to always cover all fields except the big ones.

DavidVorick:
would be great if there was some way to test the cached revisions from the api package.

oh hmm. We could `if build.Testing && randInt(10) { dropConn }` in key places within the host. I'm not sure how gracefully our existing tests would handle that, but actually I think reasonably gracefully.

modules/renter/contractor/journal.go
```diff
@@ -0,0 +1,492 @@
+package contractor
+
+// The contractor achieves efficient persistence using a JSON transaction
+// journal. It enables efficient ACID transactions on JSON objects.
+//
+// Each journal represents a single JSON object. The object is serialized as
+// an "initial object" followed by a series of update sets, one per line. Each
+// update specifies a field and a modification. See the journalUpdate type for
+// a full specification.
```
DavidVorick: In a full WAL, updates serialized into json in a batchable way sounds quite reasonable.

I think we'll be able to keep a lot of this code.

modules/renter/contractor/downloader.go
DavidVorick: The other way we could add safety is by moving the fields themselves into the journal. Then the only way to change them is through the journal, so anything you forget  to updates will start causing bugs immediately instead of being silent until the next restart.

DavidVorick: The standard way to make that workable I think would be to add getters and setters to every field, which is awful naively, but if you grouped fields well maybe it would not be so bad. Of course, the list of getters and setters would only grow with time.

modules/renter/contractor/journal.go
```diff
@@ -0,0 +1,492 @@
...
+	return journalUpdate{
+		Path:  path,
+		Value: &rm,
+	}
+}
+
+// rewritePath replaces the value at path in json with val. The returned slice
+// may share underlying memory with json. If path is malformed, the original
+// json is returned.
+func rewritePath(json []byte, path string, val []byte) []byte {
```
DavidVorick: definitely need to be returning an error or dropping a build.Critical for malformed stuff (probably both). Silently dropping updates is not okay.

modules/renter/contractor/downloader.go
DavidVorick: hmm.... actually not great because update == fsync and we don't actually want to fsync every time we change a field.

modules/renter/contractor/journal.go
```diff
@@ -0,0 +1,492 @@
...
+		if i > 0 {
+			buf = append(buf, ',')
+		}
+		buf = append(buf, `{"p":"`...)
+		buf = append(buf, u.Path...)
+		buf = append(buf, `","v":`...)
+		buf = append(buf, *u.Value...)
+		buf = append(buf, '}')
+	}
+	buf = append(buf, ']', '\n')
```
DavidVorick: I feel pretty motivated to tag a checksum to the end here. I feel that it's realistic to get corruption that still has a close bracket at the end. The motivation mostly comes though from just how bad things could be in an invalid update got applied to the state on load.

modules/renter/contractor/journal.go
```diff
@@ -0,0 +1,492 @@
...
+	if _, err := j.f.Write(buf); err != nil {
+		return err
+	}
+	return j.f.Sync()
```
DavidVorick: This is a future thing, but for max performance:

1. update has to be done under a lock.
2. sync does not have to be done under a lock
3. function should not return until sync is complete

and as a confounding variable, you can't have multiple threads interacting with the same part of the log at the same time, but if you are going to have any good performance they must be able to modify different contractors in parallel. I don't think we are that far along to be worrying about this, but it's on the list of future headaches.

Also this is a great counterargument to the getters/setters thing. The race conditions could get really difficult to manage, especially in a performance sensitive way.

modules/renter/contractor/journal.go
```diff
@@ -0,0 +1,492 @@
...
+// Checkpoint refreshes the journal with a new initial object. It syncs the
+// underlying file before returning.
+func (j *journal) checkpoint(obj interface{}) error {
+	// write to a new temp file
+	//
+	// TODO: a separate file may not be necessary. We could use an update with
+	// path "" instead, and then overwrite the beginning of the file and
+	// truncate. If the overwrite fails, we still have the full rewrite update
+	// left at the end. Just need to be careful not to overflow into the
+	// update if the new object is large.
```
DavidVorick: Too premature, it doubles the amount of data you write to disk. Also I think risky or even unsafe because what happens if there is corruption at the head of the file? You have to be able to read through it to find the non-corrupt part. And then I think you would not want to run the truncate operation until after syncing the write at the head, but the write at the head is quite potentially going to obliterate data anyway.

It's probably too clever, and best to wait to try to scrap the move operations until the logis properly a fully separate file.

DavidVorick:
Maybe too much at the moment, but now that we're no longer building a general solution to work with any json object, we can reasonably drop the string parsing for setters for each type of update. And right now I think that's only the fc operations, as everything else just does a full flush (fine as it's not happening several times per second)

DavidVorick:
I would like to keep the updates arbitrarily extensible though. Like right now it's super easy to add another field to an update. Not sure the best way to achieve this with setters.

modules/renter/contractor/journal.go
```diff
@@ -0,0 +1,492 @@
...
+// update set may be only partially written. Partially written update sets are
+// simply ignored when reading the journal. Individual updates may also be
+// ignored if they are malformed, though other updates in the set may be
+// applied. See the journalUpdate docstring for an explanation of malformed updates.
+
+// TODO:
+// - handle case sensitivity
+// - handle null vs. empty array
+//    - if index is 0, just do a wholesale replace with [val]
+// - handle inserting new keys into object
```
DavidVorick: I think switching to strict setters eliminates all of these issues. Not sure how jarring of a change that would be though.

DavidVorick: hmm but then each setter function needs a name, because instead of writing path based updates to disk you are writing arbitrary objects. So you need the setters to have string names so the loader function knows what code to throw it through.

modules/renter/contractor/journal.go
lukechampine: yup, especially if we are moving to hardcoded update types. Part of the reason the update type is so forgiving is that it has to deal with arbitrary JSON.

lukechampine:
Specific update types seem like the right way to go. I will explore what that looks like.

lukechampine:
Using specific updates, each revision results in about 2KB synced to disk. This is a 15,000-fold improvement over the current state of things, assuming a 30MB contractor.json :P
There is certainly room for improvement in that figure as well. By my estimates, we could get it as low as 300 bytes. But that's easy enough to add later, when we are confident in the safety of this scheme.

Benchmarking shows that `update` is still quite slow, and that the slowdown is entirely due to syncing. With syncing, each update takes about 11ms. Without syncing, that drops to <100us, and the JSON encoding becomes the bottleneck.

Left to do:
- Make updates idempotent
- Add a header to the journal with version metadata
- Add checksums to each update set

modules/renter/contractor/editor.go
```diff
@@ -116,6 +116,7 @@ func (he *hostEditor) Upload(data []byte) (crypto.Hash, error) {
 	he.contractor.persist.update(updateUploadRevision{
 		NewRevisionTxn:     contract.LastRevisionTxn,
 		NewSectorRoot:      sectorRoot,
+		NewSectorIndex:     len(contract.MerkleRoots) - 1,
```
DavidVorick: double checking, don't you want this to be len(contract.MerkleRoots) - so you append, instead of replace the most recent value?

DavidVorick:
>  With syncing, each update takes about 11ms.

That's an SSD number, HDDs will be slower.

When we split out the log, I'm pretty sure we'll be able to do staggered syncs, meaning 10 pieces all trying to sync at the same time would be able to do it in 11-30ms instead of in 110ms.

11ms is still a huge improvement over where it is right now. I'm currently able to do about 1 piece per second. With this improvement, I'm hoping it'll jump to 5 or 10.

modules/renter/contractor/journal.go
```diff
@@ -0,0 +1,332 @@
...
+	if err != nil {
+		return nil, err
+	}
+	// write metadata
+	enc := json.NewEncoder(f)
+	if err := enc.Encode(journalMeta); err != nil {
+		return nil, err
+	}
+	// write empty contractorPersist
+	if err := enc.Encode(contractorPersist{}); err != nil {
```
DavidVorick: Is the initial value actually empty for the contractor?

modules/renter/contractor/journal.go
```diff
@@ -0,0 +1,332 @@
...
+}
+
+func (u updateUploadRevision) apply(data *contractorPersist) {
+	if len(u.NewRevisionTxn.FileContractRevisions) == 0 {
+		build.Critical("updateUploadRevision is missing its FileContractRevision")
+		return
+	}
+	rev := u.NewRevisionTxn.FileContractRevisions[0]
+	c := data.Contracts[rev.ParentID.String()]
+	if u.NewSectorIndex != len(c.MerkleRoots) {
```
DavidVorick: `if u.NewSectorIndes > len(c.MerkleRoots) {`

if we break it into multiple files, it's possible that the merkle roots are longer at this point, but we don't catch up until the future updates. Best to fix the sanity check to reflect that situation now, even though it technically can't happen in the single-file scenario

DavidVorick: oh huh. Once we add delete it's actually possible that the roots aren't out this far.

imagine it's `append -> delete -> delete -> delete`, then you load. You'll be trying to put a root in a spot that doesn't exist. In that case, you'd probably just want to ignore the root altogether, with an understanding that it will be deleted anyway before load is complete.

I guess this sanity check just doesn't make sense.

modules/renter/contractor/journal.go
```diff
@@ -0,0 +1,332 @@
...
+		build.Critical(fmt.Sprintf("bad sector index for %v: expected %v, got %v", rev.ParentID, len(c.MerkleRoots), u.NewSectorIndex))
+		return
+	}
+	c.LastRevisionTxn = u.NewRevisionTxn
+	c.LastRevision = rev
+	c.MerkleRoots = append(c.MerkleRoots, u.NewSectorRoot) // TODO: make this idempotent
```
DavidVorick: easy:

```
if len(c.MerkleRoots) == u.NewSectorIndex {
    append
} else if len(c.MerkleRoots) > u.NewSectorIndex {
    c.MerkleRoots[u.NewSectorIndex] = root
} else {
    // this is going to be deleted by a future update anyway, don't do anything
}
```

modules/renter/contractor/journal.go
```diff
@@ -0,0 +1,332 @@
...
+
+type updateCachedUploadRevision struct {
+	Revision    types.FileContractRevision `json:"revision"`
+	SectorRoot  crypto.Hash                `json:"sectorroot"`
+	SectorIndex int                        `json:"sectorindex"`
+}
+
+func (u updateCachedUploadRevision) apply(data *contractorPersist) {
+	c := data.CachedRevisions[u.Revision.ParentID.String()]
+	if u.SectorIndex != len(c.MerkleRoots) {
```
DavidVorick: same deal with this sanity check

modules/renter/contractor/journal.go
```diff
@@ -0,0 +1,332 @@
...
+		build.Critical(fmt.Sprintf("bad sector index for %v: expected %v, got %v", u.Revision.ParentID, len(c.MerkleRoots), u.SectorIndex))
+		return
+	}
+	c.Revision = u.Revision
+	c.MerkleRoots = append(c.MerkleRoots, u.SectorRoot)
```
DavidVorick: need to do the same thing here as you did above

DavidVorick:
It's not super clear to me how the cached revisions are working, nor that they are safe.

Can you put some comments somewhere explaining the process, why it's safe, and then add some DEBUG code to the `save()` / `checkpoint()` functions to make sure that the cached revisions are in fact being saved correctly?

modules/renter/contractor/formcontract.go
```diff
@@ -113,7 +114,8 @@ func (c *Contractor) managedNewContract(host modules.HostDBEntry, numSectors uin
 
 	contractValue := contract.RenterFunds()
 	c.log.Printf("Formed contract with %v for %v SC", host.NetAddress, contractValue.Div(types.SiacoinPrecision))
-
+	contract.MerkleRoots = []crypto.Hash{}
+	c.cachedRevisions[contract.ID] = cachedRevision{contract.LastRevision, contract.MerkleRoots}
```
lukechampine: oh, this was a temporary measure that the old (generic) journal required. I can remove it now

modules/renter/contractor/editor.go
lukechampine: it's a little confusing, but what happens is that `contract.MerkleRoots` gets appended to before this call. So the length is already 1+ the old length.

modules/renter/contractor/journal.go
```diff
@@ -0,0 +1,332 @@
+package contractor
+
+// The contractor achieves efficient persistence using a JSON transaction
+// journal. It enables efficient ACID transactions on JSON objects.
+//
+// The journal represents a single JSON object, containing all of the
+// contractor's persisted data. The object is serialized as an "initial
+// object" followed by a series of update sets, one per line. Each update
```
lukechampine: what do you mean by human-readable? They seem pretty readable to me as-is.

modules/renter/contractor/journal.go
lukechampine: I thought we weren't allowing deletion anymore

modules/renter/contractor/journal.go
DavidVorick: Indented JSON, maybe I just misunderstood the comment though

modules/renter/contractor/journal.go
DavidVorick: Just until we add the new delete operation. It will be supported at some point, you just have to do swap-and-pop instead of random delete.

modules/renter/contractor/journal.go
lukechampine: ah. One nice property of having each update on its own line is that in the event of corruption, you can seek directly to the next line. I don't do this currently, but I probably should. Otherwise if you fail to read one update, you may fail to read good updates on subsequent lines.

lukechampine: Another thing that would greatly improve readability would be to use `json:",omitempty"` on all our fields. In my tests this seemed to shrink the journal by 10-15%. But there are API compatibility concerns: for anything exposed to the API, we would need to preserve the current behavior.

modules/renter/contractor/journal.go
lukechampine: Pretty much. The only non-empty fields are the ones that get filled in by `openJournal`. I figured that it would be necessary to check for `null` whenever you open the journal, so it's redundant to initialize them in `newJournal`. I suppose it wouldn't hurt though.

modules/renter/contractor/journal.go
```diff
@@ -0,0 +1,332 @@
...
+		return nil, err
+	}
+
+	// decode metadata
+	dec := json.NewDecoder(f)
+	var meta persist.Metadata
+	if err = dec.Decode(&meta); err != nil {
+		return nil, err
+	} else if meta.Version != journalMeta.Version {
+		return nil, errors.New("incompatible version")
```
lukechampine: the persist package currently returns `ErrBadVersion` and `ErrBadHeader`, which are equally unhelpful. They do allow you to check for a specific error though. That may be more useful when we implement a more robust compatibility pipeline

modules/renter/contractor/journal.go
lukechampine: why does the scenario above lead to putting a root in a spot that doesn't exist? It will looks like this:
```
initial state: [1, 2, 3]
update: set 3 4
update: delete 3
update: delete 2
update: delete 1
```
when loading, the set (append) will be applied, resulting in `[1, 2, 3, 4]`. Then the three deletes are applied, resulting in `[1]`. In other words, the deletes can't be applied before the append.

modules/renter/contractor/journal.go
DavidVorick: If you run into a corrupted update, everything after it is also corrupt because the process is linear. Imagine that you've got two appends and one is corrupt. Then your second append is going to give you an improper state.

If you hit a corrupt entry, nothing after it should be assumed to have sync'd. You should stop reading, even if the later checksums match.

modules/renter/contractor/contractor.go
DavidVorick: bump

modules/renter/contractor/journal.go
```diff
@@ -55,6 +56,27 @@ func (j *journal) update(us updateSet) error {
...
+				panic("CachedRevisions mismatch: " + fmt.Sprint(c, c2))
+			}
+		}
+		for id, c := range data.Contracts {
+			if c2, ok := data2.Contracts[id]; ok && !reflect.DeepEqual(c, c2) {
+				panic("Contracts mismatch: " + fmt.Sprint(c, c2))
+			}
+		}
+		j2.Close()
+	}
```
DavidVorick: :+1:

DavidVorick:
Now that the sanity checks are in place, if you get this working on the antfarm I'll feel comfortable throwing it on the server. Then we can see what sort of speedups we're getting.

lukechampine:
ok, working in the antfarm now, including upload+download after a full renew cycle.
The only thing left to test here is that it can successfully load the old contractor.json.

DavidVorick:
I will try it out on the server as soon as the renew cycle completes :)

DavidVorick:
I ran it on the server, and... VROOOM. Went from about 12mbps sustained upload to doing about 170mbps sustained upload. Those numbers are not exaggerated, they are the production values that I've been measuring.

So, basically this seems good to go. And also, a major, major improvement. Very excited to have this make the v1.1.1 release.

PR #1590, opened 2017-02-22
Erasure lock

DavidVorick:
Each commit pretty much stands on its own, they could be separate PRs. Also pretty easy to exclude anything that's controversial.

### DIFF
--- a/api/ecosystem_test.go
+++ b/api/ecosystem_test.go
@@ -265,7 +265,6 @@ func TestHostPoorConnectivity(t *testing.T) {
 	if testing.Short() {
 		t.SkipNow()
 	}
-	t.Parallel()
 
 	// Create the various nodes that will be forming the simulated ecosystem of
 	// this test.
--- a/api/hostdb_test.go
+++ b/api/hostdb_test.go
@@ -27,7 +27,6 @@ func TestHostDBHostsActiveHandler(t *testing.T) {
 	if testing.Short() {
 		t.SkipNow()
 	}
-	t.Parallel()
 	st, err := createServerTester("TestHostDBHostsActiveHandler")
 	if err != nil {
 		t.Fatal(err)
@@ -117,7 +116,6 @@ func TestHostDBHostsAllHandler(t *testing.T) {
 	if testing.Short() {
 		t.SkipNow()
 	}
-	t.Parallel()
 	st, err := createServerTester("TestHostDBHostsAllHandler")
 	if err != nil {
 		t.Fatal(err)
@@ -149,7 +147,6 @@ func TestHostDBHostsHandler(t *testing.T) {
 	if testing.Short() {
 		t.SkipNow()
 	}
-	t.Parallel()
 	st, err := createServerTester("TestHostDBHostsHandler")
 	if err != nil {
 		t.Fatal(err)
@@ -320,6 +317,7 @@ func TestHostDBScanOnlineOffline(t *testing.T) {
 	if testing.Short() {
 		t.SkipNow()
 	}
+	t.Parallel()
 	st, err := createServerTester("TestHostDBScanOnlineOffline")
 	if err != nil {
 		t.Fatal(err)
--- a/api/renterhost_test.go
+++ b/api/renterhost_test.go
@@ -24,7 +24,6 @@ func TestHostAndRentVanilla(t *testing.T) {
 	if testing.Short() {
 		t.SkipNow()
 	}
-	t.Parallel()
 	st, err := createServerTester("TestHostAndRentVanilla")
 	if err != nil {
 		t.Fatal(err)
@@ -164,7 +163,6 @@ func TestHostAndRentMultiHost(t *testing.T) {
 	if testing.Short() {
 		t.SkipNow()
 	}
-	t.Parallel()
 	st, err := createServerTester("TestHostAndRentMultiHost-Host1andRenter")
 	if err != nil {
 		t.Fatal(err)
@@ -280,7 +278,6 @@ func TestHostAndRentManyFiles(t *testing.T) {
 	if testing.Short() {
 		t.SkipNow()
 	}
-	t.Parallel()
 	st, err := createServerTester("TestHostAndRentManyFiles-Host1andRenter")
 	if err != nil {
 		t.Fatal(err)
@@ -490,7 +487,6 @@ func TestRenterUploadDownload(t *testing.T) {
 	if testing.Short() {
 		t.SkipNow()
 	}
-	t.Parallel()
 	st, err := createServerTester("TestRenterUploadDownload")
 	if err != nil {
 		t.Fatal(err)
@@ -617,7 +613,6 @@ func TestRenterCancelAllowance(t *testing.T) {
 	if testing.Short() {
 		t.SkipNow()
 	}
-	t.Parallel()
 	st, err := createServerTester("TestRenterCancelAllowance")
 	if err != nil {
 		t.Fatal(err)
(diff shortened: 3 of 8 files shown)


modules/renter/repair.go
```diff
@@ -336,6 +330,12 @@ func (r *Renter) managedScheduleChunkRepair(rs *repairState, chunkID chunkID, ch
 		missingPieces = missingPieces[:len(usefulWorkers)]
 	}
 
+	// Erasure code the pieces.
+	pieces, err := file.erasureCode.Encode(chunkData, missingPieces)
+	if err != nil {
+		return build.ExtendErr("unable to erasure code chunk data", err)
+	}
```
lukechampine: wouldn't it be easier to just `nil` out the unneeded pieces here, instead of complicating the `Encode` interface?

modules/renter/erasure.go
```diff
@@ -43,6 +53,11 @@ func (rs *rsCode) Encode(data []byte, needed []uint64) ([][]byte, error) {
 		unneeded = append(unneeded, uint64(i))
 	}
 
+	// Encoding is very memory haevy, ensure that only one encode or decode
```
lukechampine: heavy

modules/renter/erasure.go
```diff
@@ -67,6 +82,11 @@ func (rs *rsCode) Encode(data []byte, needed []uint64) ([][]byte, error) {
 // Encode (length and order must be preserved), but with missing elements
 // set to nil.
 func (rs *rsCode) Recover(pieces [][]byte, n uint64, w io.Writer) error {
+	// Encoding is very memory haevy, ensure that only one encode or decode
```
lukechampine: heavy

siad/daemon.go
```diff
@@ -305,6 +306,16 @@ func startDaemonCmd(cmd *cobra.Command, _ []string) {
...
 	}
 
+	// Establish a loop that will clear memory every 30 seconds. This will help
+	// keep 'siad' as lightweight as possible. Can be removed once peak usage is
+	// below 1 GiB.
+	go func() {
+		for {
+			debug.FreeOSMemory()
+			time.Sleep(time.Second * 30)
+		}
```
lukechampine: for reference: from what I can tell, the runtime will do this every 5 minutes by default

profile/profile.go
```diff
@@ -107,6 +108,8 @@ func StartContinuousProfile(profileDir string) {
 			// keeps the size of the log small while still providing lots of
 			// information.
 			StopCPUProfile()
+			runtime.GC()
+			debug.FreeOSMemory()
```
lukechampine: I'm not sure this is desirable. I don't think the standard profiling tools run GC before collecting stats.

siad/daemon.go
DavidVorick: I believe that is correct. I was thinking about putting this even at like 10 seconds. The GC operation only takes a couple of milliseconds I believe. Mostly this is to keep the RAM usage as low as possible.

profile/profile.go
DavidVorick: yeah I think I agree here, probably not the right choice. I'll pull out the commit.

modules/renter/repair.go
DavidVorick: I wanted the memory to be released before the erasure code lock was released.

DavidVorick: the GC runs later, so it's kind of moot. I wish there was a way to Free() it immediately.

lukechampine: That isn't guaranteed to happen anyway unless you force a GC cycle. I don't think the complexity is worth it.

What we really want is a shared buffer that `Encode` will write to, e.g.:

```go
var erasureBuffer [][]byte // initialized to necessary capacity
var erasureLock sync.Mutex

func repair() {
    erasureLock.Lock()
    pieces := erasureCode.Encode(chunkData, erasureBuffer)
    upload(pieces)
    erasureLock.Unlock()
}
```
This would put a hard limit on the amount of memory allocated for erasure coding, and it would also eliminate the need to `nil` unused pieces.

DavidVorick: you'd just add some extra information to the object to indicate whether the upload on that data had completed or not.

I'll pull the locks from the erasure coding package, they don't really seem to be helping memory usage at all.

DavidVorick:
Added one extra commit to get more through logged profile stats. That dragged in a new formatting dependency. If we don't want it, we can probably just copy-paste the code into the file where it's used. We only use 2 functions, and I imagine that their implementations are simple.

DavidVorick:
bump

profile/profile.go
```diff
@@ -103,18 +110,55 @@ func StartContinuousProfile(profileDir string) {
...
+
 			// Sleep for an exponential amount of time each iteration, this
 			// keeps the size of the log small while still providing lots of
 			// information.
-			StopCPUProfile()
-			SaveMemProfile(profileDir, "continuousProfilingMem")
 			StartCPUProfile(profileDir, "continuousProfilingCPU")
 			time.Sleep(sleepTime)
-			sleepTime = time.Duration(1.5 * float64(sleepTime))
+			sleepTime = time.Duration(1 * float64(sleepTime))
```
lukechampine: seems like this line could just be deleted?

profile/profile.go
```diff
@@ -103,18 +110,55 @@ func StartContinuousProfile(profileDir string) {
...
+			fmt.Fprintln(w, "MCacheSys:\t", humanize.Bytes(m.MCacheSys))
+			fmt.Fprintln(w, "MCacheInuse:\t", humanize.Bytes(m.MCacheInuse))
+			w.Flush()
+			log.Println(string(b.Bytes()))
+
 			// Sleep for an exponential amount of time each iteration, this
 			// keeps the size of the log small while still providing lots of
 			// information.
-			StopCPUProfile()
-			SaveMemProfile(profileDir, "continuousProfilingMem")
```
lukechampine: StopCPUProfile was removed, but there's still a StartCPUProfile after it

profile/profile.go
```diff
@@ -103,18 +110,55 @@ func StartContinuousProfile(profileDir string) {
...
+			fmt.Fprintln(w, "HeapInuse:\t", humanize.Bytes(m.HeapInuse))
+			fmt.Fprintln(w)
+			fmt.Fprintln(w, "StackSys:\t", humanize.Bytes(m.StackSys))
+			fmt.Fprintln(w, "StackInuse:\t", humanize.Bytes(m.StackInuse))
+			fmt.Fprintln(w)
+			fmt.Fprintln(w, "MSpanSys:\t", humanize.Bytes(m.MSpanSys))
+			fmt.Fprintln(w, "MSpanInuse:\t", humanize.Bytes(m.MSpanInuse))
+			fmt.Fprintln(w)
+			fmt.Fprintln(w, "MCacheSys:\t", humanize.Bytes(m.MCacheSys))
+			fmt.Fprintln(w, "MCacheInuse:\t", humanize.Bytes(m.MCacheInuse))
```
lukechampine: Is this the best format for this data? I figured you would want it in some sort of tabular form so that you can easily see how a given value changes over time. E.g.
```
Goroutines    Alloc
100           10
150           20
125           30
```
etc.

DavidVorick: I haven't found it too difficult to read. Normally you are trying to map the memory sizes to events such as the wallet, more than mapping them over time.

But really, I'm more interested in seeing the numbers as they were most recently than as they were historically at this point.

This format keeps the status quo, it's how they were already displayed I just added more values and made them human readable.

DavidVorick: I'm not really happy that this commit adds another dependency, going to strip it out too.

These other 3 should have made the RC, next time I will split separate commits in to separate PRs.

DavidVorick:
Went ahead and removed the memory-free commit as well. I don't think it was ultimately helping all that much.

PR #1593, opened 2017-02-24
Renter File Handling

DavidVorick:
1. Start repairing files immediately upon startup
2. Don't repair files that we aren't tracking.


DavidVorick:
failing due to a race condition, merging these commits with the race condition commits.

PR #1596, opened 2017-02-28
Scan tuning

DavidVorick:
I changed some of the scanning properties. With this tuning, we should be good for release.

### DIFF
--- a/modules/renter/hostdb/consts.go
+++ b/modules/renter/hostdb/consts.go
@@ -9,11 +9,11 @@ import (
 const (
 	// maxHostDowntime specifies the maximum amount of time that a host is
 	// allowed to be offline while still being in the hostdb.
-	maxHostDowntime = 30 * 24 * time.Hour
+	maxHostDowntime = 10 * 24 * time.Hour
 
 	// minScans specifies the number of scans that a host should have before the
 	// scans start getting compressed.
-	minScans = 20
+	minScans = 12
 
 	// maxSettingsLen indicates how long in bytes the host settings field is
 	// allowed to be before being ignored as a DoS attempt.
@@ -43,7 +43,7 @@ var (
 	// scanningThreads is the number of threads that will be probing hosts for
 	// their settings and checking for reliability.
 	scanningThreads = build.Select(build.Var{
-		Standard: int(40),
+		Standard: int(20),
 		Dev:      int(4),
 		Testing:  int(3),
 	}).(int)
@@ -61,7 +61,7 @@ var (
 	// maxScanSleep is the maximum amount of time that the hostdb will sleep
 	// between performing scans of the hosts.
 	maxScanSleep = build.Select(build.Var{
-		Standard: time.Hour * 4,
+		Standard: time.Hour * 8,
 		Dev:      time.Minute * 10,
 		Testing:  time.Second * 15,
 	}).(time.Duration)
--- a/modules/renter/hostdb/hostdb.go
+++ b/modules/renter/hostdb/hostdb.go
@@ -107,7 +107,9 @@ func newHostDB(g modules.Gateway, cs modules.ConsensusSet, persistDir string, de
 	hdb.hostTree = hosttree.New(hdb.calculateHostWeight)
 
 	// Load the prior persistence structures.
+	hdb.mu.Lock()
 	err = hdb.load()
+	hdb.mu.Unlock()
 	if err != nil && !os.IsNotExist(err) {
 		return nil, err
 	}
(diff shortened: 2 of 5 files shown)


modules/renter/hostdb/consts.go
```diff
@@ -9,11 +9,11 @@ import (
 const (
 	// maxHostDowntime specifies the maximum amount of time that a host is
 	// allowed to be offline while still being in the hostdb.
-	maxHostDowntime = 30 * 24 * time.Hour
+	maxHostDowntime = 10 * 24 * time.Hour
```
DavidVorick: hosts get pruned + compressed after 10 days, not 30.

DavidVorick: this will help clear out the hostdb faster. If a host has been offline for that long, they need to re-announce to get found by renters.

modules/renter/hostdb/consts.go
```diff
@@ -9,11 +9,11 @@ import (
...
 
 	// minScans specifies the number of scans that a host should have before the
 	// scans start getting compressed.
-	minScans = 20
+	minScans = 12
```
DavidVorick: hosts only need 12 scans before being pruned. This will help speed up compression and pruning.

modules/renter/hostdb/consts.go
```diff
@@ -43,7 +43,7 @@ var (
 	// scanningThreads is the number of threads that will be probing hosts for
 	// their settings and checking for reliability.
 	scanningThreads = build.Select(build.Var{
-		Standard: int(40),
+		Standard: int(20),
```
DavidVorick: back down to 20 scanning threads. I worry that too many concurrent connections makes scans more likely to fail. Also now that hosts are added to hostdb right away and guaranteed to be scanned, it's not as important to scan hosts quickly.

modules/renter/hostdb/consts.go
```diff
@@ -61,7 +61,7 @@ var (
 	// maxScanSleep is the maximum amount of time that the hostdb will sleep
 	// between performing scans of the hosts.
 	maxScanSleep = build.Select(build.Var{
-		Standard: time.Hour * 4,
+		Standard: time.Hour * 8,
```
DavidVorick: wait up to 8 hours between scans instead of just 4. Didn't seem to be necessary to be doing 9 scans a day on average. Now it's more like 5.

modules/renter/hostdb/hostweight.go
```diff
@@ -47,18 +47,11 @@ var (
...
-			return 1e3
-		default:
-			panic("incorrect/missing value for requiredStorage constant")
-		}
-	}()
+	requiredStorage = build.Select(build.Var{
+		Standard: uint64(20e9),
+		Dev:      uint64(1e6),
+		Testing:  uint64(1e3),
+	}).(uint64)
```
DavidVorick: required storage raised to 20GB, meaning to escape the storage penalty altogether you need to have 4 TiB of data.

modules/renter/hostdb/hostweight.go
```diff
@@ -162,20 +155,29 @@ func storageRemainingAdjustments(entry modules.HostDBEntry) float64 {
...
-		base = base / 5 // 96x total penalty
+		base = base / 5 // 3,600x total penalty
 	}
 	if entry.RemainingStorage < 5*requiredStorage {
-		base = base / 10 // 960x total penalty
+		base = base / 5 // 14,400x total penalty
 	}
 	if entry.RemainingStorage < requiredStorage {
-		base = base / 10 // 9600x total penalty
+		base = base / 5 // 72,000x total penalty
```
DavidVorick: Penalties scale a lot harder now. This is particularly because the faster uploads means that the smaller hosts are actually filling up all the way, which is bad for renters.

modules/renter/hostdb/persist.go
```diff
@@ -66,6 +66,11 @@ func (hdb *HostDB) load() error {
 		if err != nil {
 			hdb.log.Debugln("ERROR: could not insert host while loading:", host.NetAddress)
 		}
+
+		// Make sure that all hosts have gone through the initial scanning.
+		if len(host.ScanHistory) < 2 {
+			hdb.queueScan(host)
+		}
```
DavidVorick: scan hosts if they haven't been scanned before. This makes sure that you don't miss out on good hosts if you shutdown siad too quickly after your first load.

modules/renter/hostdb/hostdb.go
```diff
@@ -107,7 +107,9 @@ func newHostDB(g modules.Gateway, cs modules.ConsensusSet, persistDir string, de
 	hdb.hostTree = hosttree.New(hdb.calculateHostWeight)
 
 	// Load the prior persistence structures.
+	hdb.mu.Lock()
 	err = hdb.load()
+	hdb.mu.Unlock()
```
lukechampine: this seems weird. Why is locking required in a constructor?

DavidVorick: Spins up scanning threads

DavidVorick: If you look at the queue scan function, you see it launches a thread that was triggering a race condition

lukechampine: `load` spinning up goroutines makes me uneasy. A function called `load` should just do one thing: load data into the `hdb` object. You could avoid the locking by moving the call to `queueScan` into `newHostDB`, or perhaps `threadedScan`.

In the interest of getting the release out, I'll merge this, but I don't like it.

DavidVorick: It's an efficiency thing. The load function already touches all 1000 hosts. If we were to call queue scan elsewhere we'd need to touch all 1000 hosts again, or have the load function return a list of hosts that need to be queued

PR #1597, opened 2017-03-03
change penalty structure for host version

DavidVorick:
this fixes the ndf we're seeing on the master branch

### DIFF
--- a/modules/renter/hostdb/hostweight.go
+++ b/modules/renter/hostdb/hostweight.go
@@ -184,6 +184,9 @@ func storageRemainingAdjustments(entry modules.HostDBEntry) float64 {
 // version reported by the host.
 func versionAdjustments(entry modules.HostDBEntry) float64 {
 	base := float64(1)
+	if build.VersionCmp(entry.Version, "1.1.2") < 0 {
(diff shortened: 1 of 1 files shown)


modules/renter/hostdb/hostweight.go
```diff
@@ -184,6 +184,9 @@ func storageRemainingAdjustments(entry modules.HostDBEntry) float64 {
 // version reported by the host.
 func versionAdjustments(entry modules.HostDBEntry) float64 {
 	base := float64(1)
+	if build.VersionCmp(entry.Version, "1.1.2") < 0 {
+		base = base * 0.99999 // Safety value to make sure we update the version penalties every time we update the host.
+	}
```
lukechampine: why exactly does this cause problems?

DavidVorick: There's a test that makes sure the value is being used. The default is '1', so if the reported result is 1 then the code is probably getting skipped.

Also it's good to choose hosts with the most recent versions because they will have fewer bugs, so I want this problem to crop up every time we change the version if I forget to adjust this value

PR #1600, opened 2017-03-03
Upgrade Travis to Go 1.8 and use t.Name

lukechampine:
As of Go 1.8, `testing.T` and `testing.B` have a `Name` method. Previously we have cut-and-paste the name of the test into strings, but this is error prone and has caused NDFs in the past. Going forward, we should use `t.Name()` in all new tests.
I updated all existing tests except for those in the host package. I will add those as well if @DavidVorick does not think the changes will conflict too badly with the new storage manager code.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ os:
   - linux
 
 go:
-  - 1.6
+  - 1.8
 
 install:
   - make dependencies
--- a/api/client_test.go
+++ b/api/client_test.go
@@ -11,7 +11,7 @@ func TestApiClient(t *testing.T) {
 		t.SkipNow()
 	}
 	t.Parallel()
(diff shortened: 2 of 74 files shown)


DavidVorick:
Prefer if you hold back on the host package. Bad copy paste here has caused many bugs, glad we don't need to worry about it anymore

DavidVorick:
Worth noting that go1.8 is now a requirement for building Sia. At some point it might be worth waiting a few months before using new go releases so that package managers have time to catch up. Definitely annoys me about javascript projects that I've always got to have the most recent version of node.

lukechampine:
True. I was a bit eager for this because it also gives us clean shutdown in the server and more ergonomic sorting. I'll get to those at some point. I don't think there's anything amazing slated for Go 1.9, so we should be stable for another year.

PR #1602, opened 2017-03-10
Fix critical bug preventing hosts from submitting storage proofs

DavidVorick:
Not sure why I didn't look into this sooner. Shame on me.


lukechampine:
AppVeyor seems to be failing with a "write after close" error

DavidVorick:
travis CI failed with a contractor error as well, something complaining about merkle roots not matching.

Looking into that as well.

DavidVorick:
all better

PR #1604, opened 2017-03-10
Change command naming of `siac renter export contracts`

RNabel:
New name: `siac renter export rankings-data` and expand the help messages. Change made to clarify that `renter export contracts` is intended to be used to upload contract details to [https://rankings.sia.tech/](https://rankings.sia.tech/).

### DIFF
--- a/siac/export.go
+++ b/siac/export.go
@@ -16,13 +16,14 @@ var (
 		Use:   "export",
 		Short: "export renter data to various formats",
 		Long:  "Export renter data in various formats.",
-		// Run field not provided; export requires a subcommand
+		// Run field not provided; export requires a subcommand.
 	}
 
 	renterExportContractsCmd = &cobra.Command{
-		Use:   "contracts [destination]",
-		Short: "export the renter's contracts",
-		Long:  "Export the renter's current contract set in JSON format to the specified file.",
+		Use:   "rankings-data [destination]",
+		Short: "export the renter's contracts for import to `https://rankings.sia.tech/`",
+		Long:  "Export the renter's current contract set in JSON format to the specified " +
+			"file. Intended for upload to `https://rankings.sia.tech/`.",
 		Run:   wrap(renterexportcontractscmd),
 	}
 )
--- a/siac/export.go
+++ b/siac/export.go
@@ -20,7 +20,7 @@ var (
 	}
 
 	renterExportContractsCmd = &cobra.Command{
-		Use:   "rankings-data [destination]",
+		Use:   "contract-txns [destination]",
 		Short: "export the renter's contracts for import to `https://rankings.sia.tech/`",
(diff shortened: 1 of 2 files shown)


lukechampine:
Neat, thanks for this! I agree that the old name want very clear. Are there any other suggestions for how to name this call? `rankings-data` is ok but I don't like that it has a hyphen (can be confused with flags). Maybe `renter export contracttxns`?
Also, we should update the name of the associated command variable and function name.

RNabel:
No problem. I went with `rankings-data` as this was one of the suggestions from David (on slack). Happy to change it to `contracttxns`, the main point was to make it clear that the JSON content is intended for the rankings page, which should be obvious for from the description.

If there are no other suggestions, I'll change it to `contracttxns` and update the variable name.

lukechampine:
I'll wait for @DavidVorick's input. I prefer `contracttxns` because it describes the data itself rather than one specific use for that data, but it's not a huge deal. The main thing we want to avoid is that `export contracts` implied that you could import those contracts on a different machine and access your data.

DavidVorick:
What is the aversion to the dash? Seems pretty standard to me. Flags only have dashes in the front, preceded by a space in all examples that I can bring to mind.

`contract-txns` or `contracttxns` seems fine to me, but I prefer the dash I think

lukechampine:
either is fine with me. We don't currently have any commands containing a dash, but they certainly exist in other software. There is only one other siac command that contains multiple words, `setallowance`, with no dash. But `setallowance` is clearly more readable than `contracttxns`, which perhaps is why we didn't feel the need for a dash. @RNabel, please update the name as you see fit (dash or no dash) and this should be good to merge.

siac/export.go
```diff
@@ -16,14 +16,15 @@ var (
...
 	renterExportContractsCmd = &cobra.Command{
-		Use:   "contracts [destination]",
-		Short: "export the renter's contracts",
-		Long:  "Export the renter's current contract set in JSON format to the specified file.",
-		Run:   wrap(renterexportcontractscmd),
+		Use:   "contract-txns [destination]",
+		Short: "export the renter's contracts for import to `https://rankings.sia.tech/`",
+		Long: "Export the renter's current contract set in JSON format to the specified " +
+			"file. Intended for upload to `https://rankings.sia.tech/`.",
+		Run: wrap(renterexportcontractscmd),
```
lukechampine: this function (`renterexportcontractscmd`) should be renamed to `renterexportcontracttxnscmd`, and the `renterExportContractsCmd` variable should be renamed to `renterExportContractTxnsCmd`

RNabel: Ah, yes, of course. Please see my most recent push. Should be fixed now.

lukechampine:
Woo! Thanks.

PR #1609, opened 2017-03-17
Compat code for new wallet

lukechampine:
This PR adds upgrade code for migrating from the old wallet.json to the new wallet.db.

Not visible in this PR is the merge commit (66bcadf5) I made to update the wallet branch to the latest master. So far I have not been able to get the branch to pass the `TestRentAndHostReload` api test. I would appreciate some help in tracking down why exactly the test fails. The error message is:
```
renterhost_test.go:1172: wallet has coins spent in incomplete transactions - not enough remaining coins
```
I suspected defragging was the cause, but disabling defragging does not fix the problem.

### DIFF
--- a/modules/wallet/persist.go
+++ b/modules/wallet/persist.go
@@ -167,9 +167,10 @@ func (w *Wallet) convertPersist(dbFilename, compatFilename string) error {
 		tx.Bucket(bucketWallet).Put(keyUID, data.UID[:])
 		tx.Bucket(bucketWallet).Put(keyEncryptionVerification, data.EncryptionVerification)
 		tx.Bucket(bucketWallet).Put(keyPrimarySeedFile, encoding.Marshal(data.PrimarySeedFile))
-		dbPutPrimarySeedProgress(tx, data.PrimarySeedProgress)
 		tx.Bucket(bucketWallet).Put(keyAuxiliarySeedFiles, encoding.Marshal(data.AuxiliarySeedFiles))
 		tx.Bucket(bucketWallet).Put(keySpendableKeyFiles, encoding.Marshal(data.UnseededKeys))
+		// old wallets had a "preload depth" of 25
+		dbPutPrimarySeedProgress(tx, data.PrimarySeedProgress+25)
 
 		// set consensus height and CCID to zero so that a full rescan is
 		// triggered
--- a/modules/wallet/persist.go
+++ b/modules/wallet/persist.go
@@ -8,6 +8,7 @@ import (
 	"path/filepath"
 
 	"github.com/NebulousLabs/Sia/crypto"
+	"github.com/NebulousLabs/Sia/encoding"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/persist"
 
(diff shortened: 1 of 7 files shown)


DavidVorick:
Defragging makes sure to leave a dozen or so outputs available which are not transactions.

The message is a good clue though I'm guessing you misread it. "Incomplete transactions" are different from unconfirmed transactions. Somewhere there's a transaction builder which has reserved outputs but not spent them. Either that builder has hanged, or it failed and someone forgot to call `Drop`. 

It could be some of the code you added, or it could be the allowance stuff, or maybe even another process entirely.

DavidVorick:
Might be worth grepping that error message to make sure it's an incomplete builder

modules/renter/hostdb/hostweight.go
```diff
@@ -186,7 +186,7 @@ func storageRemainingAdjustments(entry modules.HostDBEntry) float64 {
 // version reported by the host.
 func versionAdjustments(entry modules.HostDBEntry) float64 {
 	base := float64(1)
-	if build.VersionCmp(entry.Version, "1.1.2") < 0 {
+	if build.VersionCmp(entry.Version, "1.1.2") <= 0 {
```
DavidVorick: ?

modules/wallet/persist.go
```diff
@@ -117,6 +137,59 @@ func (w *Wallet) CreateBackup(backupFilepath string) error {
 	return w.createBackup(f)
 }
 
+type compatPersist struct {
```
DavidVorick: Need a version number in all compat type names. It'll make life a lot easier down the road.

modules/wallet/persist.go
```diff
@@ -117,6 +137,59 @@ func (w *Wallet) CreateBackup(backupFilepath string) error {
...
+	UnseededKeys           []spendableKeyFile
+}
+
+var compatMeta = persist.Metadata{
+	Header:  "Wallet Settings",
+	Version: "0.4.0",
+}
+
+// convertPersist converts an old wallet.json file to a wallet.db database.
+func (w *Wallet) convertPersist(dbFilename, compatFilename string) error {
```
DavidVorick: Definitely need version numbers here, both a from and a to

modules/renter/hostdb/hostweight.go
lukechampine: I actually pushed this commit into master because it was causing all builds to fail. I'm not sure what the proper fix is. Is the expectation that we'll update this code after every release?

DavidVorick: yes, the expectation is that we modify the code after every release, because each release potentially adds important changes which change how much the newer hosts should be prioritized.

v1.1.2 for example adds a critical bugfix, and so hosts running older  than 1.1.2 should be penalized pretty heavily.

So, this fix was definitely an incomplete fix.

lukechampine: ok. The reason this wasn't caught before was that I usually push version bump commits directly to master. If I had made a PR, the build would have failed. We can fix this up when it's time to make the next release.

PR #1612, opened 2017-03-21
add ExtractTarGz

lukechampine:


DavidVorick:
Should add a tiny testdata .tar.gz

lukechampine:
I did tho. `testdata/test.tar.gz` is a tiny archive with three files named `1`, `2`, and `3`

DavidVorick:
Oh, it was hidden.

PR #1615, opened 2017-03-23
crypto changes

lukechampine:
I iterated on the new crypto randomness generator and it now runs at ~100MB/s. In addition, many crypto functions no longer return errors. This resulted in some wide-ranging cleanups. `crypto/rand` is no longer used anywhere in the project except to seed the new generator. I also removed some old mocking code while I was at it. See [rand.go](https://github.com/NebulousLabs/Sia/blob/e042dac2cfd6f78cd6d216a5c7f734508ac96734/crypto/rand.go) for the interesting new stuff.
I was pleased to see that the ratio of deletions to additions is about 3:1, which corresponds to one
```
if err != nil {
    return err
}
```
block being deleted for each line added.


lukechampine:
hmmm... The build can't complete because `go-upnp` hasn't been updated to use this code, but we can't update `go-upnp` until this is merged into master...
`go-upnp` should probably just use `(math/rand).Intn` instead of our crypto package.

DavidVorick:
Actually a should just create a package called fastrand or something, and pull our rand code into that.

Why does go-upnp need randomness? I am always slow to use math/rand.

crypto/rand.go
```diff
@@ -2,109 +2,103 @@ package crypto
...
-	entropyBase = HashObject(base)
+// A randReader produces random values via repeated hashing. The entropy field
+// is the concatenation of an initial seed and a 128-bit counter. Each time
+// the entropy is hashed, the counter is incremented.
+type randReader struct {
+	entropy [16 + HashSize]byte
+	h       hash.Hash
+	buf     [32]byte
+	mu      sync.Mutex
 }
```
DavidVorick: There is no reason for this object. It's not something anyone should ever have to customize. Random numbers are random numbers, what goes on under the hood doesn't matter as long as what comes out is secure.

crypto/rand.go
```diff
@@ -2,109 +2,103 @@ package crypto
...
+		r.h.Reset()
+		r.h.Write(r.entropy[:])
+		r.h.Sum(r.buf[:0])
 
 		// Fill out 'b'.
-		copy(b[i:], entropy[:])
+		n += copy(b[n:], r.buf[:])
 	}
-	return b, nil
+	r.mu.Unlock()
```
DavidVorick: would prefer to see atomic operations here, there are situations where the rng is going to be used in parallel, and there's no reason that it shouldn't do speedups when you use it in parallel. This lock is held during all the hashing, meaning parallelism offers no advantages.

A single atomic uint64 is good enough, then when it gets up above a certain value (2^63 for example) you mix in more entropy and then reset it. That can happen safely in a separate goroutine.

crypto/rand.go
DavidVorick: I guess it's needed for the global reader. But we still shouldn't need to be setting our own hash. Also that could cause problems because different hashes output different sized results, and yet your `buf` and `entropy` fields are fixed sizes.

DavidVorick:
Can't help myself, writing some benchmarks and improving on this design.

Definitely worth pulling this out into its own package once we're done, thing is going to be blazing.

DavidVorick:
Just pushed a commit which improves on the benchmarks. In particular, the contention benchmark that was written didn't measure well the impact of contention, because the background threads were sleeping a lot and you were only measuring the performance of a single thread.

The new benchmarks send out several goroutines at once, and wait for them to all come back. The `b.Bytes` value is set to the total amount of data read in all of the goroutines.

The new results:
```
go test -tags='debug testing' -timeout=500s -run=XXX -bench=. ./crypto
BenchmarkRandIntn-4           	 2000000	       731 ns/op
BenchmarkRandIntnLarge-4      	 2000000	       834 ns/op
BenchmarkRead32-4             	 3000000	       336 ns/op	  95.19 MB/s
BenchmarkRead64K-4            	    2000	    938615 ns/op	  68.19 MB/s
BenchmarkRead4Threads-4       	  100000	     16608 ns/op	   7.71 MB/s
BenchmarkRead4Threads64k-4    	     300	   4057354 ns/op	  63.10 MB/s
BenchmarkRead64Threads-4      	   10000	    285338 ns/op	   7.18 MB/s
BenchmarkRead64Threads64k-4   	      20	  76976246 ns/op	  53.21 MB/s
BenchmarkReadCrypto32-4       	  300000	      6158 ns/op	   5.20 MB/s
BenchmarkReadCrypto64K-4      	     200	   6036572 ns/op	  10.60 MB/s
BenchmarkPerm32-4             	  100000	     22195 ns/op
BenchmarkPermLarge4k-4        	    1000	   2894932 ns/op
PASS
ok  	github.com/NebulousLabs/Sia/crypto	26.565s
```

In particular, you can see that doing tiny reads with multiple threads resulted in a pretty dramatic slowdown, less than 8% of the performance of just doing it with one thread. I don't think that slowdown is genuine though, I'm guessing most of the penalty comes from spinning up the goroutines, which is probably more expensive than calling read once. Looking at the larger reads gives a better picture of what's happening.

When doing large reads, the slowdown was a lot less significant. We were getting 63 MB/s vs. 68MB/s. Note that this is not performance per thread, it's performance overall. Adding parallelism to the reader slows down global throughput.

Finally, I was concerned that my changes would introduce race conditions, so before creating my changes I added a large concurrency test to make sure that using the Reader a ton wouldn't cause any issues.

DavidVorick:
My computer was under load for the above benchmarks, they are garbage.

lukechampine:
I had originally pulled the code out into a separate repo, but decided that we could split it out after merging it here. Guess I should have stuck with the original plan. I will create `fastrand` (that's what I named it too lol) and open a PR with your improvements. This PR will then be updated to import `fastrand`.

DavidVorick:
I know the plan is to open a new repo, but now we have fast-crypto, buffered-crypto, and the new fastrand package you are planning on creating.

Will you just be pulling the rand.go file from the buffered-crypto branch?

lukechampine:
My plan was to create the repo using my original code, and then open a PR on it that contains your improvements. That way, we can continue to optimize the code in the PR, but `fastrand` will still be an importable package (i.e. can be used in `go-upnp` and `Sia`).

DavidVorick:
I am happy with the optimizations as they stand.

Someone suggested on IRC that chacha might be faster than blake2b. I don't actually think so (both can do about 3 cycles per byte when fully optimized), but I also don't feel it's worth messing around with further.

The next big thing we can do here is upgrade to the faster blake2b (my vote would be the std lib, but if minio is benching a lot faster than no reason not to), as blake2b is currently the biggest bottleneck by quite a bit.

DavidVorick:
Maybe too much headache, but since we went ahead and made the fastrand repo... might as well strip out all of the rand stuff and just pull the rand.go file out entirely from the repo.

PR #1619, opened 2017-03-25
do not write out whole file when creating new storage folder

DavidVorick:
also remove TODOs related to directory syncing, as I don't think there's an easy way to do this.


lukechampine:
should updated the comment on managedAddStorageFolder.

Looks like there is a package for getting the disk space: https://github.com/ricochet2200/go-disk-usage
It uses the `syscall` package though; we would want to use https://godoc.org/golang.org/x/sys instead. Could easily be a subfolder inside `contractmanager`

PR #1621, opened 2017-03-26
Add bash completion command.

### LINKED ISSUE #1613: Tab completion when using siac
It would be nice to be able to use tab completion when navigating the siac commands.

RNabel:
Adds a command allowing the creation of a bash completion script, as well as instructions on how to use it. Closes #1613.

The command syntax is:
```bash
siac bash-completion [path]
```

This command creates a file which needs to be copied to the bash completion folder (e.g. `/etc/bash_completion.d/`)

### DIFF
--- a/siac/bashcomplcmd.go
+++ b/siac/bashcomplcmd.go
@@ -1,0 +1,25 @@
+package main
+
+import "github.com/spf13/cobra"
+
+var (
+	bashcomplCmd = &cobra.Command{
+		Use:   "bash-completion [path]",
+		Short: "Creates bash completion file.",
+		Long: "Creates a bash completion file at the specified " +
+			"location.\n\n" +
+
+			"Note: Bash completions will only work with the " +
+			"prefix with which the script is created (e.g. " +
+			"`./siac` or `siac`).\n\n" +
+
+			"Once created, the file has to be moved to the bash " +
+			"completion script folder - usually " +
+			"`/etc/bash_completion.d/`.",
+		Run: wrap(bashcomplcmd),
+	}
+)
+
+func bashcomplcmd(path string) {
+	rootCmd.GenBashCompletionFile(path)
+}
--- a/siac/main.go
+++ b/siac/main.go
@@ -16,16 +16,19 @@ import (
 	"github.com/NebulousLabs/Sia/build"
 )
 
-// flags
 var (
+	// Flags.
 	addr              string // override default API address
 	initPassword      bool   // supply a custom password when creating a wallet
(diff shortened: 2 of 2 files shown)


lukechampine:
Excellent! Just need to run `gofmt`. Most Go devs have some sort of on-save hook set up so that `gofmt` is run automatically. There are plugins to do this on all mainstream editors, or you can use a git pre-commit hook.

siac/bashcomplcmd.go
```diff
@@ -0,0 +1,20 @@
...
+
+var (
+	bashcomplCmd = &cobra.Command{
+		Use:   "bash-completion [path]",
+		Short: "Creates bash completion file.",
+		Long:  `Creates a bash completion file at the specified location.
+
+Note: Bash completions will only work with the prefix with which the script is created (e.g. './siac' or 'siac').
+
+Once created, the file has to be moved to the bash completion script folder - usually /etc/bash_completion.d/`,
```
lukechampine: These should be wrapped to 80 chars.

Could we just write to `/etc/bash_completion.d/` by default?

RNabel: Unfortunately not, unless `/etc/bash_completion.d/` can be edited by the current user. If the user calls `sudo siac bash-completion ...` then all completions will only work with the prefix `sudo siac` and not just `siac`.

RNabel:
@lukechampine Done!

siac/bashcomplcmd.go
```diff
@@ -0,0 +1,25 @@
...
+		Long: "Creates a bash completion file at the specified " +
+			"location.\n\n" +
+
+			"Note: Bash completions will only work with the " +
+			"prefix with which the script is created (e.g. " +
+			"`./siac` or `siac`).\n\n" +
+
+			"Once created, the file has to be moved to the bash " +
+			"completion script folder - usually " +
+			"`/etc/bash_completion.d/`.",
```
lukechampine: this is fine, but typically we use raw string literals (` `` `) for multi-line strings

RNabel: Noted.

vevarm:
I pack this into the debian package of sia, which will be released any week now ;). Any @RNabel would you consider adding also a manpage generation, I assume it would be very similar to bash completion.

vevarm:
After reviewing what changes you did, with this to _lean on how hard can it be_. Ill give it a try myself.

RNabel:
Haha, good luck! PM me (rnabel) on slack if you need an extra pair of eyes.

Issue #1620, opened 2017-03-26
[feature] siac renter contracts view <contractid>


vevarm:
Would be nice with some inspector for the current active contacts. For the purpose of educating the community this would be great, I think.

RNabel:
@vevarm just added a PR for this, let me know what you think.

PR #1624, opened 2017-03-27
Add contract details command

RNabel:
Adds an endpoint for contract details, and a `siac` command to view details of a contract. Adresses #1620.

The endpoint is at:
`GET /renter/contract/*contract-id`. It returns a json-serialized `modules.RenterContract` object.

The siac command is:
`siac renter contracts view <contract-id>`

The output is currently:
```
OVERVIEW
========
End height:             107385
Host:                   x.x.x.x:9982
ID:                     xxxxxxxxxx7c56f90b7cc0a8ef31c27792bd2bb8cc4d5c468ae5426bec5

FILE CONTRACT
=============
File Size:              0 B

SPENDING
========
Renter funds:           2.329 SC
Upload Spending:        7.565 mS
```

This is a work in progress - let me know what else should be included. See [here](https://github.com/NebulousLabs/Sia/blob/master/modules/renter.go#L192) which information can be printed.

### DIFF
--- a/api/renter.go
+++ b/api/renter.go
@@ -89,12 +89,33 @@ type (
 
 	// RenterContract represents a contract formed by the renter.
 	RenterContract struct {
-		EndHeight       types.BlockHeight    `json:"endheight"`
-		ID              types.FileContractID `json:"id"`
-		LastTransaction types.Transaction    `json:"lasttransaction"`
-		NetAddress      modules.NetAddress   `json:"netaddress"`
-		RenterFunds     types.Currency       `json:"renterfunds"`
-		Size            uint64               `json:"size"`
+		// Amount of contract funds that have been spent on downloads.
+		DownloadSpending types.Currency `json:"downloadspending"`
+		// Block height that the file contract ends on.
+		EndHeight types.BlockHeight `json:"endheight"`
+		// Fees paid in order to form the file contract.
+		Fees types.Currency `json:"fees"`
+		// Public key of the host the contract was formed with.
+		HostPublicKey types.SiaPublicKey `json:"hostpublickey"`
+		// ID of the file contract.
+		ID types.FileContractID `json:"id"`
+		// A signed transaction containing the most recent contract revision.
+		LastTransaction types.Transaction `json:"lasttransaction"`
+		// Address of the host the file contract was formed with.
+		NetAddress modules.NetAddress `json:"netaddress"`
+		// Remaining funds left for the renter to spend on uploads & downloads.
+		RenterFunds types.Currency `json:"renterfunds"`
+		// Size of the file contract, which is typically equal to the number of
+		// bytes that have been uploaded to the host.
+		Size uint64 `json:"size"`
+		// Block height that the file contract began on.
+		StartHeight types.BlockHeight `json:"startheight"`
+		// Amount of contract funds that have been spent on storage.
+		StorageSpending types.Currency `json:"StorageSpending"`
+		// Total cost to the wallet of forming the file contract.
+		TotalCost types.Currency `json:"totalcost"`
+		// Amount of contract funds that have been spent on uploads.
+		UploadSpending types.Currency `json:"uploadspending"`
 	}
 
 	// RenterContracts contains the renter's contracts.
@@ -232,12 +253,19 @@ func (api *API) renterContractsHandler(w http.ResponseWriter, _ *http.Request, _
 	contracts := []RenterContract{}
 	for _, c := range api.renter.Contracts() {
 		contracts = append(contracts, RenterContract{
-			EndHeight:       c.EndHeight(),
-			ID:              c.ID,
-			NetAddress:      c.NetAddress,
-			LastTransaction: c.LastRevisionTxn,
-			RenterFunds:     c.RenterFunds(),
-			Size:            c.LastRevision.NewFileSize,
+			DownloadSpending: c.DownloadSpending,
+			EndHeight:        c.EndHeight(),
+			Fees:             c.TxnFee.Add(c.SiafundFee).Add(c.ContractFee),
+			HostPublicKey:    c.HostPublicKey,
+			ID:               c.ID,
+			LastTransaction:  c.LastRevisionTxn,
+			NetAddress:       c.NetAddress,
+			RenterFunds:      c.RenterFunds(),
+			Size:             c.LastRevision.NewFileSize,
+			StartHeight:      c.StartHeight,
+			StorageSpending:  c.StorageSpending,
+			TotalCost:        c.TotalCost,
+			UploadSpending:   c.UploadSpending,
 		})
 	}
 	WriteJSON(w, RenterContracts{
--- a/doc/API.md
+++ b/doc/API.md
@@ -737,12 +737,49 @@ returns active contracts. Expired contracts are not included.
 {
   "contracts": [
     {
-      "endheight":       50000, // block height
-      "id":              "1234567890abcdef0123456789abcdef0123456789abcdef0123456789abcdef",
-      "lasttransaction": {}, // types.Transaction
-      "netaddress":      "12.34.56.78:9",
-      "renterfunds":     "1234", // hastings
-      "size":            8192    // bytes
+      // Amount of contract funds that have been spent on downloads.
+      "downloadspending": "1234", // hastings
+
+      // Block height that the file contract ends on.
+      "endheight": 50000, // block height
+
+      // Fees paid in order to form the file contract.
+      "fees": "1234", // hastings
+
+      // Public key of the host the contract was formed with.
+      "hostpublickey": {
+        "algorithm": "ed25519",
+        "key": "RW50cm9weSBpc24ndCB3aGF0IGl0IHVzZWQgdG8gYmU="
+      },
+
+      // ID of the file contract.
+      "id": "1234567890abcdef0123456789abcdef0123456789abcdef0123456789abcdef",
+
+      // A signed transaction containing the most recent contract revision.
+      "lasttransaction": {},
+
+      // Address of the host the file contract was formed with.
+      "netaddress": "12.34.56.78:9",
+
+      // Remaining funds left for the renter to spend on uploads & downloads.
+      "renterfunds": "1234", // hastings
+
+      // Size of the file contract, which is typically equal to the number of
+      // bytes that have been uploaded to the host.
+      "size": 8192, // bytes
+
+      // Block height that the file contract began on.
+      "startheight": 50000, // block height
+
+      // Amount of contract funds that have been spent on storage.
+      "storagespending": "1234", // hastings
+
+      // Total cost to the wallet of forming the file contract.
+      // This includes both the fees and the funds allocated in the contract.
+      "totalcost": "1234", // hastings
+
+      // Amount of contract funds that have been spent on uploads.
+      "uploadspending": "1234" // hastings
     }
   ]
 }
--- a/siac/main.go
+++ b/siac/main.go
@@ -267,6 +267,9 @@ func main() {
 		renterContractsCmd, renterFilesListCmd, renterFilesRenameCmd,
 		renterFilesUploadCmd, renterUploadsCmd, renterExportCmd,
 		renterPricesCmd)
+
+	renterContractsCmd.AddCommand(renterContractsViewCmd)
+
 	renterCmd.Flags().BoolVarP(&renterListVerbose, "verbose", "v", false, "Show additional file info such as redundancy")
 	renterDownloadsCmd.Flags().BoolVarP(&renterShowHistory, "history", "H", false, "Show download history in addition to the download queue")
 	renterFilesListCmd.Flags().BoolVarP(&renterListVerbose, "verbose", "v", false, "Show additional file info such as redundancy")
--- a/siac/rentercmd.go
+++ b/siac/rentercmd.go
@@ -62,6 +62,13 @@ have a reasonable number (>30) of hosts in your hostdb.`,
 		Run:   wrap(rentercontractscmd),
 	}
 
+	renterContractsViewCmd = &cobra.Command{
+		Use:   "view [contract-id]",
+		Short: "View details of the specified contract",
+		Long:  "View all details available of the specified contract.",
+		Run:   wrap(rentercontractsviewcmd),
+	}
+
 	renterFilesDeleteCmd = &cobra.Command{
 		Use:     "delete [path]",
 		Aliases: []string{"rm"},
@@ -292,6 +299,48 @@ func rentercontractscmd() {
 			c.ID)
 	}
 	w.Flush()
+}
+
+// rentercontractsviewcmd is the handler for the command `siac renter contracts <id>`.
+// It lists details of a specific contract.
+func rentercontractsviewcmd(cid string) {
+	var rc api.RenterContracts
+	err := getAPI("/renter/contracts", &rc)
+	if err != nil {
+		die("Could not get contract details: ", err)
+	}
+
+	for _, rc := range rc.Contracts {
+		if rc.ID.String() == cid {
+			fmt.Printf(`
+Contract %v
+Host: %v (Public Key: %v)
+
+Start Height: %v
+End Height:   %v
+
+Total cost:        %v (Fees: %v)
+Funds Allocated:   %v
+Upload Spending:   %v
+Storage Spending:  %v
+Download Spending: %v
+Remaining Funds:   %v
+
+File Size: %v
+`, rc.ID, rc.NetAddress, rc.HostPublicKey, rc.StartHeight, rc.EndHeight,
+				currencyUnits(rc.TotalCost),
+				currencyUnits(rc.Fees),
+				currencyUnits(rc.TotalCost.Sub(rc.Fees)),
+				currencyUnits(rc.UploadSpending),
+				currencyUnits(rc.StorageSpending),
+				currencyUnits(rc.DownloadSpending),
+				currencyUnits(rc.RenterFunds),
+				filesizeUnits(int64(rc.Size)))
+			return
+		}
+	}
+
+	fmt.Println("Contract not found")
 }
 
 // renterfilesdeletecmd is the handler for the command `siac renter delete [path]`.
--- a/modules/renter.go
+++ b/modules/renter.go
@@ -188,7 +188,7 @@ func (mrs *MerkleRootSet) UnmarshalJSON(b []byte) error {
 }
 
 // A RenterContract contains all the metadata necessary to revise or renew a
-// file contract.
+// file contract. See `api.RenterContract` for field information.
 type RenterContract struct {
 	FileContract    types.FileContract         `json:"filecontract"`
 	HostPublicKey   types.SiaPublicKey         `json:"hostpublickey"`
--- a/modules/renter/contractor/persist.go
+++ b/modules/renter/contractor/persist.go
@@ -94,6 +94,14 @@ func (c *Contractor) load() error {
 		if contract.StartHeight == 0 {
 			contract.StartHeight = c.currentPeriod + 1
 		}
+		// COMPATv1.1.2
+		// Old versions calculated the TotalCost field incorrectly, omitting
+		// the transaction fee. Recompute the TotalCost from scratch using the
+		// original allocated funds and fees.
+		if len(contract.FileContract.ValidProofOutputs) > 0 {
+			contract.TotalCost = contract.FileContract.ValidProofOutputs[0].Value.
+				Add(contract.TxnFee).Add(contract.SiafundFee).Add(contract.ContractFee)
+		}
 		c.contracts[contract.ID] = contract
 	}
 


api/api.go
```diff
@@ -207,6 +207,7 @@ func New(requiredUserAgent string, requiredPassword string, cs modules.Consensus
 		router.GET("/renter", api.renterHandlerGET)
 		router.POST("/renter", RequirePassword(api.renterHandlerPOST, requiredPassword))
 		router.GET("/renter/contracts", api.renterContractsHandler)
+		router.GET("/renter/contract/*contractid", api.renterContractDetailHandler)
```
lukechampine: if you use `:contractid`, the resulting variable won't have a leading `/`

api/renter.go
```diff
@@ -245,6 +245,29 @@ func (api *API) renterContractsHandler(w http.ResponseWriter, _ *http.Request, _
...
+		}
+	}
+
+	if ind > -1 {
+		// Success - return the contract details.
+		WriteJSON(w, carr[ind])
+	} else {
+		// Failure.
+		WriteError(w, Error{"Contract ID was not found."}, http.StatusBadRequest)
+	}
```
lukechampine: I would clean this up like so:
```go
cid := ps.ByName("contractid")
for _, c := range api.renter.Contracts() {
    if c.ID.String() == cid {
        WriteJSON(w, c)
        return
    }
}
WriteError(w, Error{"no record of contract " + cid}, http.StatusBadRequest)
```

siac/rentercmd.go
```diff
@@ -294,6 +301,35 @@ func rentercontractscmd() {
...
+	out = fmt.Sprintf("%s\nFile Size:\t\t%v", out, filesizeUnits(int64(rc.FileContract.FileSize)))
+	// TODO: finish this section.
+
+	// Spending.
+	out = fmt.Sprintf("%s\n\nSPENDING\n========", out)
+	out = fmt.Sprintf("%s\nRenter funds:\t\t%s", out, currencyUnits(rc.RenterFunds()))
+	out = fmt.Sprintf("%s\nUpload Spending:\t%v", out, currencyUnits(rc.UploadSpending))
+	// TODO: finish this section.
+
+	fmt.Println(out)
```
lukechampine: I think I would like to see:

```go
fmt.Printf(`
Contract %v
Host: %v (Public Key: %v)

Start Height: %v
End Height:   %v (%v blocks remaining)

Total Cost:        %v (Fees: %v)
Funds Allocated:   %v
Upload Spending:   %v
Storage Spending:  %v
Download Spending: %v
Remaining Funds:   %v

File Size: %v
`, args...)
```

Example:
```
Contract xxxxxxxxxx7c56f90b7cc0a8ef31c27792bd2bb8cc4d5c468ae5426bec5
Host: x.x.x.x:9982 (Public Key: ed25519:pubkeypubkeypubkey...)

Start Height: 103456
End Height:   123456 (20000 blocks remaining)

Total Cost:        2 KS (Fees: 500 SC)
Funds Allocated:   1.5 KS
Upload Spending:   300 SC
Storage Spending:  200 SC
Download Spending: 200 SC
Remaining Funds:   700 SC

File Size: 10 GB
```
@vevarm, please chime in with what sort of output you'd like to see.

api/renter.go
RNabel: Done.

api/api.go
RNabel: Just to clarify: do you think it would more sense to use a URL param for this? e.g.
`/renter/contract?contractdid=abcxxxxxx`

siac/rentercmd.go
RNabel: Done, new version:

```
Contract xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
Host: x.x.x.x:9982 (Public Key: {ed25519 [64 242 174 152 179 181 235 206 30 242 64 36 229 95 74 72 206 234 233 162 187 50 187 153 120 175 119 7 103 92 166 97]})

Start Height: 94425
End Height:   107385 (12960 blocks remaining)

Total cost:        30.22 SC (Fees: 10.24 SC)
Funds Allocated:   19.98 SC
Upload Spending:   1.009 mS
Storage Spending:  229.3 mS
Download Spending: 9.456 mS
Remaining Funds:   8.754 SC

File Size: 100.66 MB
```

api/api.go
DavidVorick: it's the convention to have the folder be plural?

e.g. `/renter/contracts/:id` and not `/renter/contract/:id`?

Is it possible / does it make sense to merge the `/renter/contracts` endpoint with the new endpoint? Or can the middleware not handle that?

@RNabel just a warning, because we're committed to maintaining full compatibility with the API for multiple years, this PR is probably going to get a lot of nit-picking, back-and-forth, etc. I think this is a good endpoint to add to the API, but the standards we have for extending the API are very high.

api/renter.go
```diff
@@ -245,6 +245,23 @@ func (api *API) renterContractsHandler(w http.ResponseWriter, _ *http.Request, _
 	})
 }
 
+func (api *API) renterContractDetailHandler(w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
```
DavidVorick: convention is to match the handler name to the route. So `/renter/contracts/:id` would become `renterContractsIdHandler`, where a capital letter indicates a `/`. Since we haven't fully settled on the route name yet, find to leave it.

This also needs a docstring, even if the docstring is pretty simple.

siac/main.go
```diff
@@ -262,6 +262,9 @@ func main() {
 		renterContractsCmd, renterFilesListCmd, renterFilesRenameCmd,
 		renterFilesUploadCmd, renterUploadsCmd, renterExportCmd,
 		renterPricesCmd)
+
+	renterContractsCmd.AddCommand(renterContractDetailsCmd)
```
DavidVorick: I believe same thing here, the command name should be derived directly from the route name. Makes navigating 100 api calls a lot easier. Use comments to indicate what the command does if you don't think the name is nearly expressive enough.

DavidVorick: oh nvm this is siac, the cmd name follows the actual command sequence you type into siac.

DavidVorick:
Best to hold off until we've settled on the full details of the call, but all new api calls must:

1. have an integration test in the api package demonstrating that they work, and demonstrating that they handle edge cases well

2. Be fully documented in doc/API.md and the corresponding doc/api/$module.md files.

api/api.go
RNabel: Honestly no idea re the middleware - do you want me to make it `contracts`? Haven't used golang much so this'll be a bit of an adventure.

@DavidVorick Makes perfect sense - I'm happy to put the time in to get this merged.

lukechampine: hold on -- does `/renter/contracts` already return the set of active contracts? If so, is there a need for a separate endpoint? Why not have `siac` iterate through them?

DavidVorick: oh look at that, it does lol. Yeah, and it's a small enough payload that `siac` should be able to just iterate through them.

lukechampine: There's also the question of whether the endpoint returns just the active contract set, or all historical contracts. I'm assuming that the latter is preferred, if it isn't the current behavior.

RNabel: So: 

1. `api.RenterContract` needs to be extended. Is there a reason not to expose everything in `modules.RenterContract`?
0. @lukechampine We could make this the default behaviour and attach a field to `api.RenterContract` to indicate whether a contract is historical or current. Thoughts?

DavidVorick: If we are going to be feeding historical contracts out of this endpoint it should probably take a parameter to specify if historical should be included or not, as the historical contracts could be a lot more data - thousands even depending on how your renter is configured

lukechampine: >Is there a reason not to expose everything in `modules.RenterContract`?

The main reason is that it's easy to add new fields to an API, but hard to remove them. So we prefer to be as conservative as possible when adding new fields. That's why we typically define new types local to the `api` package that are subsets of the types passed around internally.

vevarm:
@RNabel Cool you started working on this, I looked at the type model and tried to make sense of it, I dont think I get it. 

Maybe I got this model all mapped wrong in my brain, but this is kind of what I am expecing: 
What are the terms on this contracts. Are these contracts so different to real world contracts?  
List all agreed fees/tariffs of various kind. 
List all bought units of above tariff/fees so far
Sum it into subtotals and total. 

Rather spam too much info rather than too little at this stage I think.

siac/rentercmd.go
```diff
@@ -353,9 +398,15 @@ func renterfilesdownloadcmd(path, destination string) {
...
-func (s bySiaPath) Less(i, j int) bool { return s[i].SiaPath < s[j].SiaPath }
+func (s bySiaPath) Len() int {
+	return len(s)
+}
+func (s bySiaPath) Swap(i, j int) {
+	s[i], s[j] = s[j], s[i]
+}
+func (s bySiaPath) Less(i, j int) bool {
+	return s[i].SiaPath < s[j].SiaPath
+}
```
lukechampine: it's convention to keep these methods all on one line because they are concise and tend to be (nearly) the same for all `sort.Sort` implementers.
If you are interested, now that we have upgraded to Go 1.8, we can replace a fair amount of our sorting types with [`sort.Slice`](https://golang.org/pkg/sort/#Slice).

siac/rentercmd.go
lukechampine: Hmm, looks like the pubkey is not printing properly. `types.SiaPublicKey` has a `String` method, which should be triggered by the `%v` directive.

siac/rentercmd.go
RNabel: Oh, this must have been `gofmt`. Is there a way to run `gofmt` only on the changes made?

lukechampine: This is not due to `gofmt` as far as I can tell. Running `gofmt` (and `go fmt`, for good measure) on `rentercmd.go` does not affect these functions on my machine. It's possible that you are running an additional formatter.

RNabel: This is now fixed. You are right, it doesn't seem to be `gofmt` - no idea how this happened then.

api/api.go
RNabel: @lukechampine, that makes sense. Did you consider to version the API? This would solve this problem somewhat.

@DavidVorick What about adding an `include_historical` flag whose default value is `false`? This way the current behaviour is not changed.

DavidVorick: per convention, the name of the flag would be `includehistorical`.

We will add a version 2 of the api at somepoint, but definitely it does not make sense at this point. I'm guessing it will not make sense for at least 12 - 18 more months, potentially longer

DavidVorick: * I think an `includehistorical` flag is a good way to add historical contracts to this call.

lukechampine: @RNabel: I think these are the fields we should add to `api.RenterContract`:

- StartHeight
- HostKey
- Spending (upload, download, storage)
- Costs (TotalCost, Fees)

It is technically possible to extract `HostKey` from `LastTxn` already, but it's a bit of a chore.

RNabel: Cool. I'll:

- Add the fields to `api.RenterContract`
- Remove the endpoint
- Implement the new logic to select a single transaction in `siac`

DavidVorick:
Quick review of what's been decided:

Code should be updated to extend the existing API endpoint instead of adding a new endpoint. Fields that are missing but are relevant to the user can be added without fear of breaking compatibility.

That's probably good for now. We also discussed handing historic contracts, but I believe that should be handled in a separate PR if it is needed.

RNabel:
Updates made to:

- Remove endpoint
- Extend `api.RenterContract`
- Add logic in `siac`
- Align `siac` handler function name with entered command

Current output example:
```bash
$ siac renter contracts view 7xxxxxxxxxxxxxxxxxxxxxxxbce5f94a182ed673e0761ddb08f4e74eceea8556320
```

```
Contract 7xxxxxxxxxxxxxxxxxxxxxxxbce5f94a182ed673e0761ddb08f4e74eceea8556320
Host: 77.81.121.89:9982 (Public Key: {ed25519 [xx 120 86 63 203 59 196 106 195 173 203 246 82 120 77 50 187 191 209 101 xx 59 24 xx 120 xx 53 xx 38 5 146 147]})

Start Height: 98704
End Height:   111662 (12958 blocks remaining)

Total cost:        29.84 SC (Fees: 10.24 SC)
Funds Allocated:   19.6 SC
Upload Spending:   37.82 mS
Storage Spending:  11.47 mS
Download Spending: 0 H
Remaining Funds:   6.194 SC

File Size: 37.75 MB
```

If the ID is not present, the following error is returned:

```
Error: Contract with id <entered id> could not be found.
```

api/renter.go
```diff
@@ -232,12 +239,19 @@ func (api *API) renterContractsHandler(w http.ResponseWriter, _ *http.Request, _
...
 		contracts = append(contracts, RenterContract{
-			EndHeight:       c.EndHeight(),
-			ID:              c.ID,
-			NetAddress:      c.NetAddress,
-			LastTransaction: c.LastRevisionTxn,
-			RenterFunds:     c.RenterFunds(),
-			Size:            c.LastRevision.NewFileSize,
+			DownloadSpending: c.DownloadSpending,
+			EndHeight:        c.EndHeight(),
+			Fees:             c.TxnFee,
```
lukechampine: should be `c.TxnFee.Add(c.SiafundFee).Add(c.ContractFee)`

siac/rentercmd.go
```diff
@@ -294,6 +301,48 @@ func rentercontractscmd() {
...
+				currencyUnits(rc.UploadSpending),
+				currencyUnits(rc.StorageSpending),
+				currencyUnits(rc.DownloadSpending),
+				currencyUnits(rc.RenterFunds),
+				filesizeUnits(int64(rc.Size)))
+			return
+		}
+	}
+
+	fmt.Printf("Error: Contract with id %s could not be found.\n", cid)
```
lukechampine: kind of a nitpick, but we generally don't preface errors in siac. Something like `Contract not found` or `No record of that contract` is sufficient because the user already knows what the ID is.

DavidVorick:
`Error: Contract with id <entered id> could not be found.`

^ probably not necessary to include the entered id in the error message, as it will be available in the bash history. I guess if it has weird characters or can only parse a fraction of the characters or something then printing the id to the terminal would make sense.

api/renter.go
RNabel: Updated, but the new output doesn't seem to add up.

Before fix:
```
Total cost:        29.84 SC (Fees: 10.24 SC)
Funds Allocated:   19.6 SC
```

After fix:
```
Total cost:        29.84 SC (Fees: 33.84 SC)
Funds Allocated:   29.84 SC
```
(i.e. Fees > TotalCost, thus TotalCost.Sub(Fees) < 0 and the subtraction results in the smaller of the two values). 

Just as a sanity check:
Total cost is the overall cost of the File Contract including upload/download spending storage spending etc. + fees.
Remaining funds is the difference between the Transaction Inputs and Outputs allocated as of the last transaction revision. Is this data part of the Transaction or the File Contract? Or is there a different Outputs field in a File Contract?

Do you know what needs to be changed here?

siac/rentercmd.go
RNabel: Fixed.

RNabel:
Made the changes as requested. However, now the Fees and TotalCost don't add up anymore (see reply to @lukechampine's comment above).  Please let me know if you have any idea what's going on there.

lukechampine:
Ah, turns out there was a bug in the reporting: `TotalCost` did not include the txn fee. (btw, `TotalCost` is the total amount of coins the renter spent in order to form the contract -- fees + money allocated in the payment channel.) I fixed this in #1664, so if you rebase onto master once that's merged, you should see the correct figures.

RNabel:
@lukechampine Just rebased and re-run the code in the PR and still get the same issue where TotalCost is less than the fees. Do you have any other pointers re how this can be fixed?

lukechampine:
unfortunately `TotalCost` is precomputed when the contract is created, so won't be able to see the fix unless you form new contracts.
I'm fairly sure that this is good to merge, but I will run it on some test contracts just to confirm that the `TotalCost` issue is actually fixed.

RNabel:
Oh of course haha
Updates to the docs in this PR or a separate one?

api/renter.go
```diff
@@ -232,12 +239,19 @@ func (api *API) renterContractsHandler(w http.ResponseWriter, _ *http.Request, _
...
-			NetAddress:      c.NetAddress,
-			LastTransaction: c.LastRevisionTxn,
-			RenterFunds:     c.RenterFunds(),
-			Size:            c.LastRevision.NewFileSize,
+			DownloadSpending: c.DownloadSpending,
+			EndHeight:        c.EndHeight(),
+			Fees:             c.TxnFee.Add(c.SiafundFee).Add(c.ContractFee),
+			HostPublicKey:    c.HostPublicKey,
+			ID:               c.ID,
+			LastTransaction:  c.LastRevisionTxn,
```
lukechampine: this field shouldn't be included

lukechampine:
preferably this one. Honestly, these descriptions should be added to the `modules.RenterContract` fields as well:
```
{
  "contracts": [
    {
      // Amount of contract funds that have been spent on downloads.
      "downloadspending": "1234", // hastings

      // Block height that the file contract ends on.
      "endheight": 50000, // block height

      // Fees paid in order to form the file contract.
      "fees": "1234", // hastings

      // Public key of the host the contract was formed with.
      "hostpublickey": {
        "algorithm": "ed25519",
        "key": "RW50cm9weSBpc24ndCB3aGF0IGl0IHVzZWQgdG8gYmU="
      },

      // ID of the file contract.
      "id": "1234567890abcdef0123456789abcdef0123456789abcdef0123456789abcdef",

      // A signed transaction containing the most recent contract revision.
      "lasttransaction": {},

      // Address of the host the file contract was formed with.
      "netaddress": "12.34.56.78:9",

      // Remaining funds left for the renter to spend on uploads & downloads.
      "renterfunds": "1234", // hastings

      // Size of the file contract, which is typically equal to the number of
      // bytes that have been uploaded to the host.
      "size": 8192, // bytes

      // Block height that the file contract began on.
      "startheight": 50000, // block height

      // Amount of contract funds that have been spent on storage.
      "storagespending": "1234", // hastings

      // Total cost to the wallet of forming the file contract.
      // This includes both the fees and the funds allocated in the contract.
      "totalcost": "1234", // hastings

      // Amount of contract funds that have been spent on uploads.
      "uploadspending": "1234" // hastings
    }
  ]
}
```

RNabel:
@lukechampine yeah, that description would have been super useful - I'll make sure to add it to this PR.

RNabel:
Documentation added to both `API.md` and the `api.RenterContract` struct.

lukechampine:
`Blocks Remaining` seems more useful, but you won't be able to display it unless you're synced. Let's just leave it out entirely for now. It's easy enough to call `siac` to see the current height and compare that to the `End Height` of the contract.

RNabel:
Took the `Remaining blocks` part out. Lmk if anything else needs changing.

lukechampine:
just this comment: https://github.com/NebulousLabs/Sia/pull/1624#issuecomment-292288613

then this should be good to go

RNabel:
cool, shall I add this to `siac` or `siad`?

lukechampine:
probably `siad`; otherwise all API clients would have to add the same fix. I would do it as close to the source as possible, in the `contractor.load` function:
```diff
diff --git a/modules/renter/contractor/persist.go b/modules/renter/contractor/persist.go
index 530ccfd..a491966 100644
--- a/modules/renter/contractor/persist.go
+++ b/modules/renter/contractor/persist.go
@@ -94,6 +94,12 @@ func (c *Contractor) load() error {
                if contract.StartHeight == 0 {
                        contract.StartHeight = c.currentPeriod + 1
                }
+               // COMPATv1.1.2
+               // Old versions calculated the TotalCost field incorrectly, omitting
+               // the transaction fee. Recompute the TotalCost from scratch using the
+               // original allocated funds and fees.
+               contract.TotalCost = contract.FileContract.ValidProofOutputs[0].Value.
+                       Add(contract.TxnFee).Add(contract.SiafundFee).Add(contract.ContractFee)
                c.contracts[contract.ID] = contract
        }
 ```

You should test this locally first to make sure it's correct, though.

File: doc/API.md [hunk 10]
```diff
@@ -737,12 +737,49 @@ returns active contracts. Expired contracts are not included.
...
-      "renterfunds":     "1234", // hastings
-      "size":            8192    // bytes
+      // Amount of contract funds that have been spent on downloads.
+      "downloadspending": "1234", // hastings
+
+      // Block height that the file contract ends on.
+      "endheight": 50000, // block height
+
+      // Fees paid in order to form the file contract.
+      "fees": "1234", // hastings
```
DavidVorick: @lukechampine: does this include (on the backend) the contract price paid to the host? Does that count as a fee?

doc/API.md
```diff
@@ -737,12 +737,49 @@ returns active contracts. Expired contracts are not included.
...
+      "id": "1234567890abcdef0123456789abcdef0123456789abcdef0123456789abcdef",
+
+      // A signed transaction containing the most recent contract revision.
+      "lasttransaction": {},
+
+      // Address of the host the file contract was formed with.
+      "netaddress": "12.34.56.78:9",
+
+      // Remaining funds left for the renter to spend on uploads & downloads.
+      "renterfunds": "1234", // hastings
```
DavidVorick: Is renterfunds the best name?

doc/API.md
```diff
@@ -737,12 +737,49 @@ returns active contracts. Expired contracts are not included.
...
+
+      // Block height that the file contract began on.
+      "startheight": 50000, // block height
+
+      // Amount of contract funds that have been spent on storage.
+      "storagespending": "1234", // hastings
+
+      // Total cost to the wallet of forming the file contract.
+      // This includes both the fees and the funds allocated in the contract.
+      "totalcost": "1234", // hastings
```
DavidVorick: Does this include the contract price paid to the host?

lukechampine: yes, the intent is that after you form a contract, your wallet will have `totalcost` fewer coins in it.

doc/API.md
lukechampine: yep, it is recorded in the `ContractFee` field

doc/API.md
lukechampine: perhaps not, but it's already in the API

DavidVorick: Ah. Darn.

RNabel:
Alright, updated and tested. The numbers add up now, but the output seems fee-heavy (just saying this in case either of you can spot anything wrong with the numbers):

```
Total cost:        45.15 SC (Fees: 38.9 SC)                                                                                                                            
Funds Allocated:   6.245 SC                                                                                                                                            
Upload Spending:   12.61 mS                                                                                                                                            
Storage Spending:  218.5 mS                                                                                                                                            
Download Spending: 10.09 mS                                                                                                                                            
Remaining Funds:   6.004 SC                                                                                                                                            
                                                                                                                                                                      
File Size: 29.36 MB
```

DavidVorick:
Fees are going to be:

+ Transaction fee (usually around 10 SC)
+ Contract fee (usually 15-30 SC, depending on host)
+ Siafund fees (3.9% of total contract cost - including the collateral).

Given that the total cost is 45.15 SC, transaction fees + contract fee are going to be 25-40, it seems normal to me.

I'd expect the total fee to be closer to 10% for a large contract (500+ SC).

lukechampine:
Woo! Thanks for this PR @RNabel, I think power users will really appreciate it.

DavidVorick:
Sorry for missing this earlier, but we need a test in the API package demonstrating that the call works as intended. This is as much to protect our compatibility promise in the future as it is to verify that there are no bugs.

RNabel:
@DavidVorick since the endpoint added fields, and fields are not exhaustively tested in the existing test what test case would you recommend? I could add checks to [this section](https://github.com/NebulousLabs/Sia/blob/master/api/renter_test.go#L439).

DavidVorick:
The policy is the exhaustively test the new fields, but if you want to go the extra mile and test the existing fields as well, we do eventually want to have an api where every endpoint, field, and parameter has a basic test demonstrating that the api performs according to specification.

PR #1626, opened 2017-03-29
don't scan local hosts found on the blockchain

DavidVorick:


johnathanhowell:
This would cause issues in the future with local antfarms, perhaps this should be in a `if build.Release == "standard"` branch?

DavidVorick:
Fixed

PR #1628, opened 2017-03-30
Faster Wallet IBD/rescan

lukechampine:
The wallet now interacts with the database via a single global bolt.Tx, which is committed every 2 minutes or when durability is required (e.g. before reporting the balance). In testing on my SSD, this sped up a full rescan about 5x. This sort of speedup is important because some of the new wallet functionality (e.g. `InitSeed` and `SweepSeed`) requires rescanning.

A further step to take here would be to remove the references to `w.dbTx` in a bunch of places and instead make it implicit. The only reason I decided against this was that it makes it very difficult to work with any sort of bolt.Tx that isn't `w.dbTx`.

### DIFF
--- a/modules/wallet/database.go
+++ b/modules/wallet/database.go
@@ -3,6 +3,7 @@ package wallet
 import (
 	"encoding/binary"
 	"reflect"
+	"time"
 
 	"github.com/NebulousLabs/Sia/encoding"
 	"github.com/NebulousLabs/Sia/modules"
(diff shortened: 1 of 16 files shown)


modules/wallet/database.go
```diff
@@ -66,6 +67,42 @@ var (
...
+	if err != nil {
+		w.log.Severe("ERROR: failed to apply database update:", err)
+		w.dbTx.Rollback()
+	}
+	// begin a new tx
+	w.dbTx, err = w.db.Begin(true)
+	if err != nil {
+		w.log.Severe("ERROR: failed to start database update:", err)
+	}
+}
```
DavidVorick: This looks good. Would be great to see an identical system in the tpool, we could probably shave 2-3 hours off of IBD time for HDDs, and 20-30 minutes for SSDs.

lukechampine:
Antfarm seems to perform fine with these changes.

DavidVorick:
Including with the big spender and little spender jobs? They use the wallet more than the other jobs by far. The renter and host also use it, but the code coverage is a lot different.

lukechampine:
I included `bigspender` but not `littlesupplier` because the latter requires major changes to the antfarm.

DavidVorick:
I've decided that we can merge this, but the `bigspender` can't operate without the `littlespender`, so your test probably did not add much real coverage.

PR #1630, opened 2017-03-31
add rpc documentation

GlenDC:
This PR aims to add documentation for the RPC code, and make it more clear what calls are available, and how they work. This should also make it easier in the future to write clients in other languages.

I'm not sure if this initial state is good enough, please collaborate and give feedback where needed :)

I am not sure if I should omit the (incomplete) type info or not, for now it's there, but if you rather have me remote that part, I'll do so.

### DIFF
--- a/doc/RPC.md
+++ b/doc/RPC.md
@@ -1,0 +1,127 @@
+RPC
+===
+
+Sia peers communicate with each other via Remote Procedure Calls. An RPC consists of a unique ID and a pair of functions, one on the calling end and one on the receiving end. After the ID is written/received, both peers hand the connection off to their respective functions. Typically, the calling end writes an object to the receiver and then reads a response.
+
+RPC IDs are always 8 bytes and contain a human-readable name for the RPC. If the name is shorter than 8 bytes, the remainder is padded with zeros. If the name is longer than 8 bytes, it is truncated.
+
+### Call Listing
+
+Unless otherwise specified, these calls follow a request/response pattern and use the [encoding](./Encoding.md) package to serialize data.
+
+**All data received via RPC should be considered untrusted and potentially malicious.**
+
+#### ShareNodes
+
+ShareNodes requests peer addresses from a peer.
+The gateway calls this RPC regularly to update its list of potential peers.
+
+ID: `"ShareNod"`
+
+Request: None
+
+Response:
+
+```go
+[]modules.NetAddress
+```
+
+Recommendations:
+
++ Requesting peers should limit the request to 3000 bytes.
++ Responding peers should send no more than 10 peers, and should not send peers that are unlikely to be reachable.
+
+#### SendBlocks
+
+SendBlocks requests blocks from a peer. The blocks are added to the requesting peer's blockchain, and optionally rebroadcast to other peers. Unlike most RPCs, the SendBlocks call is a loop of requests and responses that continues until the responding peer has no more blocks to send.
+
+ID: `"SendBloc"`
+
+Request:
+
+```go
+// Exponentially-spaced IDs of most-recently-seen blocks,
+// ordered from most recent to least recent.
+// Less than 32 elements may be present, but the last element
+// (index 31) is always the ID of the genesis block.
+[32]types.BlockID
+```
+
+Response:
+
+```go
+struct {
+   // sequential list of blocks, beginning with the first
+   // block in the main chain not seen by the requesting peer.
+   blocks []types.Block
+   // true if the responding peer can send more blocks
+   more bool
+}
+```
+
+Recommendations:
+
++ Requesting peers should limit the request to 20MB.
++ Responding peers should identify the most recent BlockID that is in their blockchain, and send up to 10 blocks following that block.
++ Responding peers should set `more = true` if they have not sent the most recent block in their chain.
+
+#### RelayHeader
+
+RelayHeader sends a block header ID to a peer, with the expectation that the peer will relay the ID to its own peers.
+
+ID: `"RelayHea"`
+
+Request:
+
+```go
+types.BlockHeader
+```
+
+Response: None
+
+Recommendation:
+
++ Requesting (sending) peers should call this RPC on all of their peers as soon as they mine or receive a block via `SendBlocks` or `SendBlk`.
++ Responding (receiving) peers should use the `SendBlk` RPC to download the actual block content. If the block is an orphan, `SendBlocks` should be used to discover the block's parent(s).
++ Responding peers should not rebroadcast the received ID until they have downloaded and verified the actual block.
+
+#### SendBlk
+
+SendBlk requests a block's contents from a peer, given the block's ID.
+
+ID: `"SendBlk\0"`
+
+Request:
+
+```go
+types.BlockID
+```
+
+Response:
+
+```go
+types.Block
+```
+
++ Requesting peers should limit the received block to 2 MB (the maximum block size).
++ Requesting peers should broadcast the block's ID using `RelayHeader` once the received block has been verified.
++ Responding peers may simply close the connection if the block ID does not match a known block.
+
+#### RelayTransactionSet
+
+RelayTransactionSet sends a transaction set to a peer.
+
+ID: `"RelayTra"`
+
+Request:
+
+```go
+[]types.Transaction
+```
+
+Response: None
+
+Recommendations:
+
++ Requesting peers should limit the request to 2 MB (the maximum block size).
++ Responding peers should broadcast the received transaction set once it has been verified.


doc/RPC.md
```diff
@@ -0,0 +1,170 @@
+RPC
+===
+
+Remote Procedure Calls are used within Sia to trigger actiosn between server and client.
+Most RPCs are send to all connected peers of a gateway using a broadcast call,
+while some RPCs have a Client and Server version,
+to provide bidirectional one-to-one communication.
```
lukechampine: My stab at an intro:

>Sia peers communicate with each other via Remote Procedure Calls. An RPC consists of a unique ID and a pair of functions, one on the calling end and one on the receiving end. After the ID is written/received, both peers hand the connection off to their respective functions. Typically, the calling end writes an object to the receiver and then reads a response.
>
>RPC IDs are always 8 bytes and contain a human-readable name for the RPC. If the name is shorter than 8 bytes, the remainder is padded with zeros. If the name is longer than 8 bytes, it is truncated.

doc/RPC.md
```diff
@@ -0,0 +1,170 @@
...
+#### ShareNodes
+
+Used as a one-to-one operation.
+
+Registered in the `gateway` module.
+
+Request Package:
+
+```
+[ []byte{"ShareNod"} | []string{nodesNetAddresses} ]
```
lukechampine: We should model these after the API docs. Something like:

>### Call Listing
>
>Unless otherwise, specified, these calls follow a request/response pattern and use the [encoding](./Encoding.md) package to serialize data.
>
>**All data received via RPC should be considered untrusted and potentially malicious.**
>
>#### ShareNodes
>
>ShareNodes requests peer addresses from a peer. The gateway calls this RPC regularly to update its list of potential peers.
>
>ID: `"ShareNod"`
>
>Request: None
>
>Response:
> 
>```go
>[]modules.NetAddress
>```
>
>Recommendations:
>
>- Requesting peers should limit the response to 3000 bytes.
>- Responding peers should send no more than 10 peers, and should not send peers that are unlikely to be reachable.

doc/RPC.md
```diff
@@ -0,0 +1,170 @@
...
+
+Registered in the `transactionpool` module.
+
+Request Content:
+
+```
+[ []byte{"RelayTra"} | []types.Transaction ]
+```
+
+### Encoded Types
```
lukechampine: this section can be dropped

doc/RPC.md
```diff
@@ -0,0 +1,170 @@
...
+
+```
+[ []byte{"SendBloc"} | blockhashes: [32]hexEncodedString ]
+```
+
+Reply Package:
+
+```
+[ []types.Block | boolean ]
+```
```
lukechampine: >#### SendBlocks
>
>SendBlocks requests blocks from a peer. The blocks are added to the requesting peer's blockchain, and optionally rebroadcast to other peers. Unlike most RPCs, the SendBlocks call is a loop of requests and responses that continues until the responding peer has no more blocks to send.
>
>ID: `"SendBloc`
>
>Request:
>
>```go
>// Exponentially-spaced IDs of most-recently-seen blocks,
>// ordered from most recent to least recent.
>// Less than 32 elements may be present, but the last element
>// (index 31) is always the ID of the genesis block.
>[32]types.BlockID
>```
>
>Response:
> 
>```go
>struct {
>    // sequential list of blocks, beginning with the first
>    // block in the main chain not seen by the requesting peer.
>    blocks []types.Block
>    // true if the responding peer can send more blocks
>    more bool
>}
>```
>
>Recommendations:
>
>- Requesting peers should limit the response to 20MB.
>- Responding peers should identify the most recent BlockID that is in their blockchain, and send up to 10 blocks following that block.
>- Responding peers should set `more = true` if they have not sent the most recent block in their chain.

doc/RPC.md
GlenDC: Thank you, this is definitely a great improvement! :)

doc/RPC.md
GlenDC: Thanks, I'll do my best to reformat the call listing, following your excellent examples :)

doc/RPC.md
```diff
@@ -0,0 +1,140 @@
+RPC
+===
+
+Sia peers communicate with each other via Remote Procedure Calls. An RPC consists of a unique ID and a pair of functions, one on the calling end and one on the receiving end. After the ID is written/received, both peers hand the connection off to their respective functions. Typically, the calling end writes an object to the receiver and then reads a response.
+
+RPC IDs are always 8 bytes and contain a human-readable name for the RPC. If the name is shorter than 8 bytes, the remainder is padded with zeros. If the name is longer than 8 bytes, it is truncated.
+
+### Call Listing
+
+Unless otherwise, specified, these calls follow a request/response pattern and use the [encoding](https://github.com/NebulousLabs/Sia/pull/Encoding.md) package to serialize data.
```
lukechampine: `./Encoding.md` should work fine as a link

doc/RPC.md
```diff
@@ -0,0 +1,140 @@
...
++ Requesting peers should limit the request to 20MB.
++ Responding peers should identify the most recent BlockID that is in their blockchain, and send up to 10 blocks following that block.
++ Responding peers should set `more = true` if they have not sent the most recent block in their chain.
+
+#### RelayBlock
+
+RelayBlock receives a block from a peer, which sent it as a broadcast.
+If the received block is an orphan,
+it will try to find its parents, using the `SendBlocks` RPC.
+Each receiving peer broadcasts the received block to all its peers as well.
```
lukechampine: Stuff about rebroadcasting should be moved to the `Recommendations` section. I think the descriptions should be written from the perspective of the peer that initiated the call. So this would become:

>RelayBlock sends a block to a peer, with the expectation that the peer will relay the block to its own peers.
>This RPC is deprecated as of v0.5.1.

Actually, @DavidVorick, should we just strip out the `RelayBlock` RPC entirely? It's been over a year since we deprecated it.

DavidVorick: Yeah, it's probably okay. I think we even deprecated it before 1.0

GlenDC: Want me to remove the code as well as the documentation? I wouldn't mind.

DavidVorick: Sure, but do the code changes in a different pull request.

GlenDC:
Does this require more changes?

lukechampine:
yeah, I have some minor comments for the sections I didn't get to previously. So far it's looking good though

GlenDC:
Cool @lukechampine, feel free to push those changes yourself to this branch though, I think as a owner of this repo you can do that. Might be quicker than going back and forward :) And makes this a real collaborative PR 👍

lukechampine:
I think only you can push to the branch since it's on your repo. We sometimes push to each other's PRs on Sia because the branches are all part of the NebulousLabs repo.

doc/RPC.md
```diff
@@ -0,0 +1,125 @@
...
+types.BlockHeader
+```
+
+Response: None
+
+Recommendation:
+
++ Requesting peers should send this message as a broadcast to all its peers.
++ Requesting peers should look for the parents of a block, in case the given header is an orphan, using the `SendBlocks` RPC.
++ Responding peers should send the matching block back to the requesting peer, in case the header is valid, using the `SendBlk` RPC.
```
lukechampine: >RelayHeader sends a block header ID to a peer, with the expectation that the peer will relay the ID to its own peers.

>Recommendations:
>
>- Requesting (sending) peers should call this RPC on all of their peers as soon as they mine a block or receive a block via `SendBlocks` or `SendBlk`.
>- Responding (receiving) peers should use the `SendBlk` RPC to download the actual block contents. If the block is an orphan, `SendBlocks` should be used to discover the block's parent(s).
>- Responding peers should not rebroadcast the received ID until they have downloaded and verified the actual block.

doc/RPC.md
```diff
@@ -0,0 +1,125 @@
...
+
+```go
+types.BlockID
+```
+
+Response:
+
+```go
+types.Block
+```
```
lukechampine: >SendBlk requests a block's contents from a peer, given the block's ID.

>Recommendations:
>
>- Requesting peers should limit the received block to 2 MB (the maximum block size).
>- Requesting peers should broadcast the block's ID using `RelayHeader` once the received block has been verified.
>- Responding peers may simply close the connection if the block ID does not match a known block.

doc/RPC.md
```diff
@@ -0,0 +1,125 @@
...
+[]types.Transaction
+```
+
+Response: None
+
+Recommendations:
+
++ Requesting peers should limit the request to 2 MB.
++ Requesting peers should send this message as a broadcast to all its peers.
++ Responding peers should broadcast the received transactions usin the same RPC to all its peers as well, in case the transactions are valid.
```
lukechampine: >RelayTransactionSet sends a transaction set to a peer.

>Recommendations:
>
>- Requesting peers should limit the request to 2 MB (the maximum block size)
>- Responding peers should broadcast the received transaction set once it has been verified.

GlenDC:
@lukechampine ah, didn't know that. Somehow I thought that you were also able to push to it. Might have misread that at some point, a couple of months ago.

I added the changes you requested.

doc/RPC.md
```diff
@@ -0,0 +1,127 @@
+RPC
+===
+
+Sia peers communicate with each other via Remote Procedure Calls. An RPC consists of a unique ID and a pair of functions, one on the calling end and one on the receiving end. After the ID is written/received, both peers hand the connection off to their respective functions. Typically, the calling end writes an object to the receiver and then reads a response.
+
+RPC IDs are always 8 bytes and contain a human-readable name for the RPC. If the name is shorter than 8 bytes, the remainder is padded with zeros. If the name is longer than 8 bytes, it is truncated.
+
+### Call Listing
+
+Unless otherwise, specified, these calls follow a request/response pattern and use the [encoding](./Encoding.md) package to serialize data.
```
lukechampine: unnecessary comma after "otherwise"

GlenDC:
Removed comma as requested.

DavidVorick:
Woo

PR #1636, opened 2017-04-01
Rpc deadlines

DavidVorick:
Add deadlines to RPCs that don't have them.

We should probably have a safety deadline in the gateway just to be sure, that will automatically put a 20 minute deadline on every conn or something like that.

I'm guessing that this PR will fix the 'too many open sockets' problems that we are seeing, as well as some of the absurd amounts of goroutines we are seeing, as well as help clean up some of the memory usage.

### DIFF
--- a/modules/consensus/synchronize.go
+++ b/modules/consensus/synchronize.go
@@ -31,11 +31,27 @@ var (
 		Testing:  types.BlockHeight(3),
 	}).(types.BlockHeight)
 
-	// sendBlocksTimeout is the timeout for the SendBlocks RPC.
+	// sendBlocksTimeout is the timeout for the SendBlocks RPC, and for the
+	// depricated RelayBlock RPC.
 	sendBlocksTimeout = build.Select(build.Var{
 		Standard: 5 * time.Minute,
 		Dev:      40 * time.Second,
 		Testing:  5 * time.Second,
+	}).(time.Duration)
+
+	// sendBlkTimeout is the timeout for the SendBlocks RPC, and for the
+	// depricated RelayBlock RPC.
+	sendBlkTimeout = build.Select(build.Var{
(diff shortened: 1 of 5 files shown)


lukechampine:
Huh, I figured the gateway would already be setting a default deadline.

The err handling code for `SetDeadline` is ugly. I suggest we add a method to `modules.PeerConn` called `SetTimeout`:
```go
// gateway/conn.go
func (pc peerConn) SetTimeout(d time.Duration) error {
	if p.LocalAddr().String() == "pipe" && build.Release == "testing" {
		// Pipes do not support SetDeadline and should only be used in testing.
		return nil
	}
	return conn.SetDeadline(time.Now().Add(d))
}
```

modules/consensus/synchronize.go
```diff
@@ -31,13 +31,29 @@ var (
...
+	// depricated RelayBlock RPC.
+	sendBlkTimeout = build.Select(build.Var{
+		Standard: 4 * time.Minute,
+		Dev:      30 * time.Second,
+		Testing:  4 * time.Second,
+	}).(time.Duration)
+
+	// relayHeaderTimeout is the timeout for the RelayHeader RPC.
+	relayHeaderTimeout = build.Select(build.Var{
+		Standard: 3 * time.Minute,
```
lukechampine: 3 minutes to relay a header? Seems very generous for transferring 32 bytes.

DavidVorick: It is rather generous, but the problem is not the data volume it's the round trip time. Especially nodes in China need a lot of extra time to get their packets around.

DavidVorick:
Sure, I'll try it.

Gateway doesn't have a default timeout, but I'm pushing a commit that gives one to RPCs.

PR #1637, opened 2017-04-02
Tpool quick shutdown

DavidVorick:
Wanted to add quick shutdown to the transaction pool, but it didn't have a threadgroup, or even a logger. So I added a threadgroup and a logger, and then added quick shutdown.

### DIFF
--- a/modules/transactionpool/persist.go
+++ b/modules/transactionpool/persist.go
@@ -2,9 +2,11 @@ package transactionpool
 
 import (
 	"errors"
+	"fmt"
 	"os"
 	"path/filepath"
 
+	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/persist"
 	"github.com/NebulousLabs/Sia/types"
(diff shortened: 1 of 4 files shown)


modules/transactionpool/persist.go
```diff
@@ -95,9 +115,22 @@ func (tp *TransactionPool) initPersist() error {
...
 		}
-		return tp.consensusSet.ConsensusSetSubscribe(tp, modules.ConsensusChangeBeginning)
+		freshScanErr := tp.consensusSet.ConsensusSetSubscribe(tp, modules.ConsensusChangeBeginning)
+		if freshScanErr != nil {
+			return freshScanErr
+		}
+		tp.tg.OnStop(func() {
+			tp.consensusSet.Unsubscribe(tp)
+		})
+		return nil
```
lukechampine: typically I write this like:
```go
err = tp.consensusSet.ConsensusSetSubscribe(tp, modules.ConsensusChangeBeginning)
```
so that control flow can then proceed to the standard error handling. Avoids code duplication, although I can see it being a little surprising.

PR #1639, opened 2017-04-02
fix double-close on Editor/Downloader

lukechampine:
Seems more robust than #1638 
`Close` is supposed to be idempotent so I think this is the Right Thing To Do.

### DIFF
--- a/modules/renter/proto/downloader.go
+++ b/modules/renter/proto/downloader.go
@@ -3,6 +3,7 @@ package proto
 import (
 	"errors"
 	"net"
+	"sync"
 	"time"
 
 	"github.com/NebulousLabs/Sia/build"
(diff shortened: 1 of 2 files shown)


DavidVorick:
I considered this approach, and I don't think this code is wrong or shouldn't be merged, but I do worry that the renter code is causing more problems than just a double close the way that it is structured. Or more that it's just doing improper resource management with the multithreading and the editors.

So, I approve of this PR, but also think the other PR should be considered as well - this PR isn't solving the problem (improper resource usage), it's just moving to a best-practice that makes the system more tolerant to mistakes.

lukechampine:
No reason we can't merge both, I suppose. I agree that this PR does not solve the underlying problem. (Although the convention in a lot of Go code seems to be that you don't need to worry about calling `Close` multiple times. Kind of strange when you contrast it to `close(channel)` panicking...)

PR #1640, opened 2017-04-02
establish clean + quick shutdown for conns in the consensus set

DavidVorick:
Combine this with the tpool PR and we should have Sia to a place where it can shut down in 10 seconds every time.


DavidVorick:
same failure that I tried to fix in the tpool PR, once that is merged I can rebase this on onto it and hopefully the NDF goes away.

DavidVorick:
rebased + fixed up

PR #1641, opened 2017-04-02
add sessionheader, extending the handshake protocol

GlenDC:
This PR adds the concept of a sessionheader, extending the handshake protocol:

+ connecting peers are now ensured to share the same genesis blockID;
+ connecting peers are now ensured not to connect to itself;
+ pinging between newer clients is no longer a hack;

It does this by having a new `connectSessionHandshake` that comes right after the `connectVersionHandshake`, which exchanges the genesisID, gatewayID and if the peer wants to connect (eg this is false when the peer was just pinging).

This is a pretty invasive feature, but should be fully backwards compatible. Well that's the goal at least. Would love feedback on this, and make this a thing into master, together, if possible.

Review Request: @DavidVorick @lukechampine


modules/gateway/peers.go
```diff
@@ -326,6 +351,36 @@ func connectVersionHandshake(conn net.Conn, version string) (remoteVersion strin
...
+		return
+	}
+	// ensure that their genesisID is the same as ours
+	if bytes.Compare(header.GenesisID[:], remoteHeader.GenesisID[:]) != 0 {
+		err = errPeerRejectedConn
+		return
+	}
+	// ensure that we're not connecting to ourselves
+	if bytes.Compare(header.UniqueID[:], remoteHeader.UniqueID[:]) == 0 {
+		err = errPeerExists
```
lukechampine: should be a `errOurAddress`

modules/gateway/peers.go
```diff
@@ -348,6 +403,45 @@ func acceptConnVersionHandshake(conn net.Conn, version string) (remoteVersion st
...
+	}
+	// ensure that we're not connecting to ourselves
+	if bytes.Compare(header.UniqueID[:], remoteHeader.UniqueID[:]) == 0 {
+		header.WantConn = false
+		if err := encoding.WriteObject(conn, header); err != nil {
+			return false, fmt.Errorf("failed to reject header: %v", err)
+		}
+		err = errPeerExists
+		return
+	}
```
lukechampine: define a helper function for this since it's used in two functions:
```go
// acceptableHeader returns an error if the remote header indicates that
// a connection is impossible.
func acceptableHeader(our, their sessionHeader) error {
    if their.GenesisID != our.GenesisID {
        return errPeerGenesisID
    } else if their.UniqueID == our.UniqueID {
        return errOurAddress
    } else if !their.WantConn {
        return errPeerRejected
    }
    return nil
}
```

Then `connectSessionHandshake` becomes:
```go
func connectSessionHandshake(conn net.Conn, header sessionHeader) error {
    // Write our header.
    if err := encoding.WriteObject(conn, header); err != nil {
        return fmt.Errorf("failed to write header: %v", err)
    }
    // Read remote header.
    var remoteHeader sessionHeader
    if err := encoding.ReadObject(conn, &remoteHeader, build.MaxEncodedVersionLength); err != nil {
        return fmt.Errorf("failed to read remote header: %v", err)
    }
    return acceptableHeader(header, remoteHeader)
}
```
And `acceptConnSessionHandshake` becomes:
```go
func acceptConnSessionHandshake(conn net.Conn, header sessionHeader) error {
    var remoteHeader sessionHeader
    // Read remote header.
    if err := encoding.ReadObject(conn, &remoteHeader, build.MaxEncodedVersionLength); err != nil {
	return fmt.Errorf("failed to read remote header: %v", err)
    }
    // Write our header.
    if err := encoding.WriteObject(conn, header); err != nil {
        return fmt.Errorf("failed to write header: %v", err)
    }
    return acceptableHeader(header, remoteHeader)
}
```

modules/gateway/peers.go
```diff
@@ -348,6 +403,45 @@ func acceptConnVersionHandshake(conn net.Conn, version string) (remoteVersion st
 	return remoteVersion, nil
 }
 
+// acceptConnSessionHandshake performs the session handshake and should be
+// called on the side accepting a connection request.
+// if err == nil, wantConn indicates if a connection is wanted on both sides.
+func acceptConnSessionHandshake(conn net.Conn, header sessionHeader) (wantConn bool, err error) {
+	var remoteHeader sessionHeader
+	// Read remote header.
+	if err = encoding.ReadObject(conn, &remoteHeader, build.MaxEncodedVersionLength); err != nil {
```
lukechampine: we need a separate constant for the maximum length. It should be `sizeof(BlockID) + sizeof(gatewayID) + sizeof(bool)` == `32 + 8 + 1` = `41`. Define it in `gateway/consts.go`

modules/gateway/peers.go
```diff
@@ -437,6 +531,23 @@ func (g *Gateway) managedConnect(addr modules.NetAddress) error {
...
+		if err != nil {
+			conn.Close()
+			return err
+		}
+
+		if !wantConn {
+			conn.Close()
+			return errPeerRejectedConn
+		}
+	}
```
lukechampine: this should be an explicit `if-else` chain:
```go
if build.VersionCmp(remoteVersion, sessionHandshakeUpgradeVersion) >= 0 {
    // new peer
} else if build.VersionCmp(remoteVersion, handshakeUpgradeVersion) >= 0 {
    // old peer
} else {
    // really old peer
}
```
this requires duplicating some code but it's less prone to bugs.

modules/gateway/peers.go
```diff
@@ -437,6 +531,23 @@ func (g *Gateway) managedConnect(addr modules.NetAddress) error {
 		conn.Close()
 		return err
 	}
+
+	// Ensure peer is not connection to itself,
+	// the peers are on the same network, and a connection is wanted.
+	if build.VersionCmp(remoteVersion, sessionHandshakeUpgradeVersion) >= 0 {
+		wantConn := true
+		wantConn, err = connectSessionHandshake(conn, g.sessionHeader(wantConn))
```
lukechampine: `wantConn, err := connectSessionHandshake(conn, g.sessionHeader(wantConn))`

modules/gateway/gateway.go
```diff
@@ -209,6 +189,15 @@ func (g *Gateway) Address() modules.NetAddress {
...
 }
 
+// sessionHeader returns the header for the session of this gateway
+func (g *Gateway) sessionHeader(wantConn bool) sessionHeader {
+	return sessionHeader{
+		GenesisID: types.GenesisID,
+		UniqueID:  g.id,
+		WantConn:  wantConn,
+	}
+}
```
lukechampine: hmm. I'm not sure about this. I feel like functions like `connectSessionHandshake` shouldn't have to take a whole `sessionHeader`, since they already know two out of three things (the genesis ID and whether they want a connection). So perhaps they should just take a `gatewayID` instead of a full `sessionHeader`. The only downside is that during testing, you can't change the genesis ID that they'll send/expect. But I don't think that's a huge deal.

GlenDC:
I ran the test that Travis failed on, and it seems to work here. I think this is one of those cases where muxado acts up, am I right?

GlenDC:
Ah yes, I was working from a forked directory at first, but then it kinda broke everything, and I guess it only really works for go libraries with no inter-dependencies. I did exactly what you did, but never realized that those wrong repo links were in there, as my IDE adds them for me. Sorry for that.

modules/gateway/gateway.go
```diff
@@ -81,15 +81,7 @@ import (
...
-// TODO: When peers connect to eachother, and when they add nodes to the node
-// list, there is no verification that the peers are running on the same Sia
-// network, something that will be problematic if we set up a large testnet.
-// It's already problematic, and currently only solved by using a different set
-// of bootstrap nodes. If there is any cross-polination (which an attacker
-// could do pretty easily), the gateways will not clean up over time, which
-// will degrade the quality of the flood network as the two networks will
-// continously flood eachother with irrelevant information. Additionally, there
-// is no public key exhcange, so communications cannot be effectively encrypted
+// TODO: There is no public key exhcange, so communications cannot be effectively encrypted
```
GlenDC: Oh nice, I quite like that comment, I will recheck my comment adds/mods. I also like to keep them short, but am never too exact about the 80, as I usually just guess. I will make sure that in your codebase it all first in 80 chars. Does this also count for other code, or just comments?

lukechampine:
Yeah, Travis sometimes fails due to non-deterministic errors, mostly because it runs so slowly that our timeouts get triggered. If that happens, just drop a comment and I will restart the build until it passes.

modules/gateway/gateway.go
```diff
@@ -188,6 +158,16 @@ type Gateway struct {
...
+
+	// Unique ID
+	id gatewayID
+}
+
+type gatewayID [8]byte
+
+func randGatewayID() (id gatewayID, err error) {
+	_, err = rand.Read(id[:])
+	return
```
GlenDC: Do you just want to do `err = rand.Read(g.id[:])` directly in the `New` function then?

modules/gateway/gateway.go
```diff
@@ -237,6 +226,12 @@ func New(addr string, bootstrap bool, persistDir string) (*Gateway, error) {
 		persistDir: persistDir,
 	}
 
+	// Set Unique GatewayID
+	g.id, err = randGatewayID()
+	if err != nil {
+		return nil, fmt.Errorf("couldn't generate unique gatewayID: %s", err)
+	}
```
GlenDC: oh ok, I guess this replaces my previous comment.

modules/gateway/gateway.go
lukechampine: it's not a super-hard rule -- if there is some code that is very indented, we are more relaxed about wrapping it. However, every comment at the top level (e.g. any docstring on a function, var, const, type) should be wrapped to 80 chars.
Most editors have a way to do this automatically. It's definitely not something you want to do by hand.

modules/gateway/peers.go
```diff
@@ -326,6 +351,36 @@ func connectVersionHandshake(conn net.Conn, version string) (remoteVersion strin
...
+		err = errPeerRejectedConn
+		return
+	}
+	// ensure that we're not connecting to ourselves
+	if bytes.Compare(header.UniqueID[:], remoteHeader.UniqueID[:]) == 0 {
+		err = errPeerExists
+		return
+	}
+	// return if both sesions want a connection
+	wantConn = header.WantConn && remoteHeader.WantConn
```
GlenDC: I changed this behavior back and forward quite a few times during development, sometimes returning exactly what you're now asking. Eventually I settled on returning the header. But I think you're right, there is never really any need to know the other information after we've settled the information is matching.

GlenDC:
Ok, and thanks for the review. Will just on it in a minute. When I'm done with my `smux` branch, would that help the reproducibility a lot, or are there other things causing issues like this?

GlenDC:
@lukechampine changed the stuff you requested.
Ran `make test && make test-long` with success.

modules/gateway/consts.go
```diff
@@ -8,9 +8,14 @@ import (
...
 const (
 	// handshakeUpgradeVersion is the version where the gateway handshake RPC
-	// was altered to include adiitional information transfer.
+	// was altered to include additional information transfer.
 	handshakeUpgradeVersion = "1.0.0"
 
+	// sessionHandshakeUpgradeVersion is the version where the gateway handshake RPC
+	// was altered to include the ID of the genesis block, the gateway's unique ID,
+	// and whether a connection is desired.
+	sessionHandshakeUpgradeVersion = "1.1.2"
```
lukechampine: 1.1.2 is the current release. The handshake upgrade will be released in 1.2.0

GlenDC: is 1.2 the version that is released next?

modules/gateway/nodes.go
```diff
@@ -33,24 +35,33 @@ func (g *Gateway) addNode(addr modules.NetAddress) error {
...
+	if err != nil {
+		return
 	}
-	return err
+
+	if build.VersionCmp(remoteVersion, sessionHandshakeUpgradeVersion) < 0 {
+		return // for older versions, this is where pinging ends
+	}
+
+	// if both peers are >= 1.1.2,
```
lukechampine: 1.2.0

modules/gateway/peers.go
```diff
@@ -39,6 +40,16 @@ type peer struct {
 	sess muxado.Session
 }
 
+// sessionHeader is used to ensure the peers are on the same network,
+// the peer is not connecting to itself.
+// when used to purge peers, wantConn can be set to false,
+// such that this peer won't be added as a node on the receiving end
+type sessionHeader struct {
```
lukechampine: >sessionHeader is sent after the initial version exchange. It prevents peers on different blockchains from connecting to each other, and prevents the gateway from connecting to itself. The receiving peer can set WantConn to false to refuse the connection, and the initiating peer can set WantConn to false if they merely want to confirm that a node is online.

modules/gateway/peers.go
```diff
@@ -126,6 +137,16 @@ func (g *Gateway) threadedAcceptConn(conn net.Conn) {
 		return
 	}
 
+	if build.VersionCmp(remoteVersion, sessionHandshakeUpgradeVersion) >= 0 {
+		wantConnect := true
+		err = acceptConnSessionHandshake(conn, g.id, wantConnect)
```
lukechampine: I think `wantConn` is implied to always be true for `acceptConnSessionHandshake`, so it doesn't need to take it as an argument. If we want to explicitly reject a peer we could use a new function called `reject` instead of `accept`.

modules/gateway/peers.go
```diff
@@ -304,6 +325,19 @@ func acceptableVersion(version string) error {
 	return nil
 }
 
+// acceptableSessionHeader returns an error
+// if the remote header indicates that a connection is impossible.
+func acceptableSessionHeader(ours, theirs *sessionHeader) error {
```
lukechampine: don't pass `sessionHeader` by pointer. It's small enough that there's no reason to worry about copying it around.

modules/gateway/peers.go
```diff
@@ -348,6 +408,32 @@ func acceptConnVersionHandshake(conn net.Conn, version string) (remoteVersion st
 	return remoteVersion, nil
 }
 
+// acceptConnSessionHandshake performs the session handshake and should be
+// called on the side accepting a connection request.
+// if err == nil, wantConn indicates if a connection is wanted on both sides.
+func acceptConnSessionHandshake(conn net.Conn, uniqueID gatewayID, wantConn bool) (err error) {
```
lukechampine: not a big deal, but we typically avoid named return values unless the meaning of the value is not obvious. Even then, we don't like naked `return` statements.

modules/gateway/peers.go
```diff
@@ -437,10 +523,17 @@ func (g *Gateway) managedConnect(addr modules.NetAddress) error {
...
-	if build.VersionCmp(remoteVersion, handshakeUpgradeVersion) < 0 {
-		err = g.managedConnectOldPeer(conn, remoteVersion, addr)
-	} else {
+
+	if build.VersionCmp(remoteVersion, sessionHandshakeUpgradeVersion) >= 0 {
+		wantConnect := true
+		err = connectSessionHandshake(conn, g.id, wantConnect)
+		if err == nil {
+			err = g.managedConnectNewPeer(conn, remoteVersion, addr)
+		}
```
lukechampine: hmm. This is ugly. We really try to avoid writing `if err == nil` because `err != nil` is so much more common. I would rather this look like:
```go
if build.VersionCmp(remoteVersion, sessionHandshakeUpgradeVersion) >= 0 {
    err = g.managedConnectv120Peer(conn, remoteVersion, addr)
} else if build.VersionCmp(remoteVersion, handshakeUpgradeVersion) >= 0 {
    err = g.managedConnectv100Peer(conn, remoteVersion, addr)
} else {
    err = g.managedConnectOldPeer(conn, remoteVersion, addr)
}
```

modules/gateway/consts.go
lukechampine: that's the plan, yeah.
Changing this string might break your tests, not sure. In that case I suppose we should just push a commit to master that updates the version number to 1.2.0.

GlenDC: Could I do that as part of this PR?

GlenDC:
Added your requested changes, and bumped version in a separate commit.

modules/gateway/nodes.go
```diff
@@ -33,24 +35,33 @@ func (g *Gateway) addNode(addr modules.NetAddress) error {
...
-	//
-	// NOTE: this is a somewhat clunky way of specifying that you didn't
-	// actually want a connection.
-	_, err = connectVersionHandshake(conn, "0.0.0")
-	if err == errPeerRejectedConn {
-		err = nil // we expect this error
+
+	// in order for 2 peers tho connect,
+	// their version should be compatible
+	remoteVersion, err := connectVersionHandshake(conn, build.Version)
```
lukechampine: >Read the node's version.

modules/gateway/peers.go
```diff
@@ -326,6 +361,29 @@ func connectVersionHandshake(conn net.Conn, version string) (remoteVersion strin
 	return remoteVersion, nil
 }
 
+// connectSessionHandshake performs the sessionHeader handshake and should be called
+// on the side making the connection request, whether that's for pure pinging reasons or not.
+// a connection is only possible if err == nil
```
lukechampine: wrap

modules/gateway/peers.go
```diff
@@ -304,6 +326,19 @@ func acceptableVersion(version string) error {
 	return nil
 }
 
+// acceptableSessionHeader returns an error
+// if the remote header indicates that a connection is impossible.
```
lukechampine: wrap

modules/gateway/peers.go
lukechampine: also the last sentence can be removed, it's implied

lukechampine:
Instead of bumping the version number in this PR, you could alter the tests like so:
```go
func TestHandshake(t *testing.T) {
    if build.VersionCmp(build.Version, sessionHandshakeVersion) < 0 {
        t.Skipf("Test requires build.Version >= %v", sessionHandshakeVersion)
    }
}
```
we usually don't bump the version number until just prior to the release.

modules/gateway/peers.go
```diff
@@ -39,6 +40,18 @@ type peer struct {
 	sess muxado.Session
 }
 
+// sessionHeader is sent after the initial version exchange.
+// It prevents peers on different blockchains from connecting to each other,
+// and prevents the gateway from connecting to itself.
+// The receiving peer can set WantConn to false to refuse the connection,
+// and the initiating peer can set WantConn to false
+// if they merely want to confirm that a node is online.
```
DavidVorick: should be noted that most text editors have some way to auto-wrap content to 80 characters. In vim, you can do this by highlighting the content in visual mode and then typing `gq`

lukechampine:
I was too hasty re: v1.2.0. We are already pushing out enough major changes in v1.2.0, so the new gateway protocol will be postponed until the release after that, probably v1.2.1.

modules/gateway/peers.go
GlenDC: What you mean? This is within 80 character limits? (66 to be precise)

modules/gateway/peers.go
GlenDC: this is within 80 characters? Or do you mean that you want to wrap exactly at 80 chars?

modules/gateway/nodes.go
GlenDC: How would that be possible? We get the version from this handshake?

modules/gateway/peers.go
GlenDC: Never mind, using Vim, I now see what you mean. I use VSCode for my day-to-day golang programming, only use vim for certain quick edits. Will check if vscode has such an option/plugin available.

DavidVorick:
I mentioned earlier, but this PR touches super critical parts of the code. The QA that we need to do to be certain that nothing breaks and compatibility is preserved is a lot, the bar is really high for touching the gateway.

We have to make sure that it's done correctly before we merge it, and that's going to take time even if there are no bugs.

modules/gateway/nodes.go
lukechampine: yeah, I just mean the comment should be simplified.

GlenDC:
@DavidVorick how would these QA tests take place?

DavidVorick:
https://github.com/NebulousLabs/Sia-Ant-Farm

We'll spin up an ant-farm that has 2-3 nodes of v1.1.2, 2-3 nodes of v1.2.0, and 2-3 nodes of [gateway upgrades]. We'll give each of the nodes different jobs, making sure that there's at least 1 renter or host in each batch. Then we'll watch them perform for a few hours and make sure everyone is able to talk to eachother.

We'll repeat without the v1.2.0 nodes, and then again but this time without the v1.1.2 nodes.

Finally, we'll take some of the gateway code and upgrade our servers, and see if they are able to collaborate the network without issues.

Right now our plate is full dealing with bugfixes and stuff in v1.2.0, so the gateway QA is going to have to wait until v1.2.0 is more stable. If you want to start messing around with the antfarm, we can help you get it running in slack. Probably the best channel would be #app-dev, as I believe there are several people in that channel who would be interested in following along. (the ant farm is a great way to test Sia apps as well as Sia itself).

modules/gateway/gateway.go
```diff
@@ -81,16 +81,9 @@ import (
...
-// It's already problematic, and currently only solved by using a different set
-// of bootstrap nodes. If there is any cross-polination (which an attacker
-// could do pretty easily), the gateways will not clean up over time, which
-// will degrade the quality of the flood network as the two networks will
-// continously flood eachother with irrelevant information. Additionally, there
-// is no public key exhcange, so communications cannot be effectively encrypted
-// or authenticated. The nodes must have some way to share keys.
+// TODO: There is no public key exhcange,
+// so communications cannot be effectively encrypted or authenticated.
+// The nodes must have some way to share keys.
```
DavidVorick: comment is incorrectly formatted. Most text editors have some shortcut or hotkey for highlighting a block of text and then snapping it to an 80 line word wrap (and will leave the `// ` parts of the comment intact and in the correct places). In vim, you can highlight the block of text in visual mode and then press `gq`

modules/gateway/nodes.go
```diff
@@ -33,24 +35,32 @@ func (g *Gateway) addNode(addr modules.NetAddress) error {
...
+
+	// Read the node's version.
+	remoteVersion, err := connectVersionHandshake(conn, build.Version)
+	if err != nil {
+		return
 	}
-	return err
+
+	if build.VersionCmp(remoteVersion, sessionHandshakeUpgradeVersion) < 0 {
+		return // for older versions, this is where pinging ends
```
DavidVorick: @lukechampine is this going to cause problems since the old peer is going to expect the connection to be maintained? We are not writing 0.0.0 as our version anymore, so we need to make sure that we hang up some other way.

modules/gateway/nodes.go
```diff
@@ -33,24 +35,32 @@ func (g *Gateway) addNode(addr modules.NetAddress) error {
...
+	}
+
+	// if both peers are >= 1.2.0,
+	// we'll also guarantee that the connection is between 2 different peers,
+	// and the peers are on the same network
```
DavidVorick: comment incorrectly formatted, newline at the top is not needed.

modules/gateway/nodes.go
```diff
@@ -33,24 +35,32 @@ func (g *Gateway) addNode(addr modules.NetAddress) error {
...
+	wantConnect := false
+	// we're pinging, so we don't care if other side wants to connect or not
```
DavidVorick: I'd probably put the variable declaration below this comment, or even put this comment on the same line as the variable declaration.

non-blocking issue, no big deal either way.

modules/gateway/peers.go
```diff
@@ -370,9 +449,9 @@ func (g *Gateway) managedConnectOldPeer(conn net.Conn, remoteVersion string, rem
 	return g.save()
 }
 
-// managedConnectNewPeer connects to peers >= v1.0.0. The peer is added as a
+// managedConnectNewPeer connects to peers >= v1.0.0 and < v1.2.0. The peer is added as a
 // node and a peer. The peer is only added if a nil error is returned.
-func (g *Gateway) managedConnectNewPeer(conn net.Conn, remoteVersion string, remoteAddr modules.NetAddress) error {
+func (g *Gateway) managedConnectv100Peer(conn net.Conn, remoteVersion string, remoteAddr modules.NetAddress) error {
```
DavidVorick: :+1:

DavidVorick:
Overall, looks pretty good. I wonder if there's some way we can write an integration test within the package (not needing the antfarm) to demonstrate compatibility between the versions. Only things I can think of right now though either require siad binaries, or some ugly bash scripts that will checkout our release tags and build multiple binaries on the fly.

But that's not necessarily sufficient anyway, because old versions will also be running different dependencies and were compiled with different compilers (namely, older compiler versions). The only real authentic way to do it would be to use our actual release binaries.

modules/gateway/nodes.go
lukechampine: At first I thought we could mitigate this by checking the remote version and writing `0.0.0` if they were below the upgrade version. But that's not possible because we send our version first and then read the remote version.

Regardless, I don't think this will cause any serious issues. The old peer will perceive it as a connection error. Worst case, it adds a line to the old peer's gateway.log, but I think that's only enabled on debug builds.

lukechampine:
seems like this is in pretty good shape (just needs some formatting tweaks), but needs more real-world testing. I would like to see:

- Nodes running the new code connecting to old nodes on the mainnet
- Nodes running the new code connecting to each other on the mainnet
- A mixture of old/new nodes interacting in antfarm tests

I hope to run these tests myself over the next few days, but it would be nice if @GlenDC could independently confirm that this code works on the mainnet.

Higher-level: to minimize the amount of compatibility code, we probably want to merge this and the `smux` change into the same release. To that end, we should merge both into a feature branch, and then merge the feature branch into a future release. Unfortunately I carelessly deleted the `staging` branch, so @GlenDC will have to reopen the `smux` PR. I suggest `gateway-proto` as the name of the feature branch.

modules/gateway/nodes.go
GlenDC: So I keep this as it is?

GlenDC:
@lukechampine I guess you mean that you guys want to merge this into a new branch called `gateway-proto`, rather than `master`? I'll re-open the `smux` PR, no problem :)

lukechampine:
@GlenDC: please fixup the merge conflicts (and the comment formatting) and this should be ready for merge, after which we will run some live tests

PR #1642, opened 2017-04-02
Faster type encoding/hashing

lukechampine:
The main thrust of this PR is a new file, `types/encoding.go`, which contains `MarshalSia` methods for many of the types in the `types` package. We may want to move the existing `MarshalSia` methods into this file.
The motivation for this new file is that adding `MarshalSia` methods to types can transparently speed up their encodings. These methods also also be fairly lengthy (depending on the number of fields in the type), which is why I defined them all in a separate file instead of alongside their type declarations, like other methods are. Generally, `MarshalSia` methods are a form of optimization and not relevant to the semantics of the type. This is different from methods like `(BlockID).MarshalJSON`, which serves to override the standard JSON encoding of the `BlockID` type (marshalling it as a hex string instead of an array of integers).

Another potential optimization I identified was to calculate `(Block).MerkleRoot` more directly. Currently we encode each field and write it out to a `bytes.Buffer`, and then pass it off to `(merkletree).Push` for hashing. We could eliminate the intermediate buffer by encoding directly into the hash, and then using the hashes to construct a `merkletree.CachedTree`. However, I felt this was too complicated and brittle to be worth it.

One of the big takeaways here is that a substantial portion of encoding "slowness" -- 20% to 30% -- comes from allocating tons of very tiny objects, which puts pressure on the garbage collector. One potential way to alleviate this is to use a `sync.Pool`. The `fmt` package employs this strategy.

EDIT: I tried using `sync.Pool` to allocate buffers for encoding uint64s, but there was no tangible gain. :/


DavidVorick:
if we are going to have an `encoding.go` we should move all of the existing MarshalSia methods to it. And we should probably also move all of the String FromString MarshalJSON methods to it as well.

crypto/hash.go
```diff
@@ -41,8 +41,14 @@ func NewHash() hash.Hash {
...
-func HashAll(objs ...interface{}) Hash {
-	return HashBytes(encoding.MarshalAll(objs...))
+func HashAll(objs ...interface{}) (hash Hash) {
+	h := NewHash()
+	enc := encoding.NewEncoder(h)
+	for _, obj := range objs {
+		enc.Encode(obj)
+	}
+	h.Sum(hash[:0])
+	return
```
DavidVorick: huh. This is both faster and bit-for-bit identical? I would probably feel better with a test or two that just compares some fixed inputs to some expected hash outputs. Not too difficult to set that up as a table test I think.

crypto/hash.go
```diff
@@ -52,8 +58,11 @@ func HashBytes(data []byte) Hash {
 
 // HashObject takes an object as input, encodes it using the encoding package,
 // and then hashes the result.
-func HashObject(obj interface{}) Hash {
-	return HashBytes(encoding.Marshal(obj))
+func HashObject(obj interface{}) (hash Hash) {
+	h := NewHash()
+	encoding.NewEncoder(h).Encode(obj)
+	h.Sum(hash[:0])
+	return
```
DavidVorick: Ditto here. The test is more for ensuring compatibility if we ever make future changes.

types/encoding.go
```diff
@@ -0,0 +1,153 @@
...
+	binary.LittleEndian.PutUint64(length, uint64(len((t.TransactionSignatures))))
+	w.Write(length)
+	for i := range t.TransactionSignatures {
+		err := t.TransactionSignatures[i].MarshalSia(w)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
DavidVorick: This is definitely a case where I would want to see some tests that prove equivalence for full types. It'd be relatively easy  to just miss one field here.

types/encoding.go
```diff
@@ -0,0 +1,153 @@
...
+		cf.SiacoinOutputs,
+		cf.FileContracts,
+		cf.FileContractRevisions,
+		cf.StorageProofs,
+		cf.SiafundInputs,
+		cf.SiafundOutputs,
+		cf.MinerFees,
+		cf.ArbitraryData,
+		cf.TransactionSignatures,
+	}
```
DavidVorick: This one as well, though it does look complete to me.

DavidVorick: I'm pretty sure that if you messed this up (we should actually verify that by intentionally messing it up), the chain would be unable to sync, and you'd have a node that was stuck at a very low block height.

DavidVorick:
Overall this looks good to me. Just need some extra reassurance that no dumb errors slipped in.

lukechampine:
agreed re: comparing against known good values. I did this to some extent while writing the methods (computed genesis block ID and compared against a hard-coded string), but did not formalize it in a real test. I will also attempt to sync a node with these changes.

DavidVorick:
You should also attempt to sync a node with changes you know to be incorrect. Perhaps screw up the storage proof or something simple. I'm curious how far it would make it before failing.

lukechampine:
Attempted sync with invalid storage proof encoding. It got to ~7000 blocks and then IBD started failing with "block does not meet target," which is what you would expect if the encoding of a block differs.

Syncing with this PR now, and no issues as of height 30000. btw, if you start syncing with a slow peer, will you be bottlenecked for the entire IBD? Or do we fetch blocks in parallel?

DavidVorick:
You fetch blocks from one peer at a time, but you move to the next peer after five minutes. The result is certainly suboptimal, but you won't get bottlenecked by a single per, because they only have you for 5 minutes.

encoding/integers.go
```diff
@@ -33,3 +34,14 @@ func DecUint64(b []byte) uint64 {
...
 	return binary.LittleEndian.Uint64(b2)
 }
+
+// WriteUint64 writes u to w.
+func WriteUint64(w io.Writer, u uint64) error {
+	_, err := w.Write(EncUint64(u))
+	return err
+}
+
+// WriteUint64 writes i to w.
```
GlenDC: WriteInt

PR #1643, opened 2017-04-02
change host announce mechanics

DavidVorick:
Host will check addresses are valid before attempting to make an announce call.
Hosts will also update the internal address to match the announcement that they
are making.

I think that we have gone back and forth on this before, but it's pretty clear
that users expect their most recent announcement to be the one that they keep.
Also, it does not make sense to be announcing a netaddress that you do not plan
to use as your permanent address, as the hostdb only remembers the most recent
address that has crossed the blockchain.

### DIFF
--- a/modules/host/announce.go
+++ b/modules/host/announce.go
@@ -3,6 +3,7 @@ package host
 import (
 	"errors"
 
+	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/modules"
 )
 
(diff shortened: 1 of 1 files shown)


modules/host/announce.go
```diff
@@ -72,37 +73,75 @@ func (h *Host) managedAnnounce(addr modules.NetAddress) error {
...
+		return errors.New("announcement requested with local net address")
+	}
+
+	// Address has cleared inspection, perform the announcement.
+	return h.managedAnnounce(annAddr)
 }
 
 // AnnounceAddress submits a host announcement to the blockchain to announce a
-// specific address. No checks for validity are performed on the address.
+// specific address.
```
lukechampine: should note that the address will be set as the host's new address

DavidVorick:
updated

PR #1647, opened 2017-04-04
move renter consts to consts.go

lukechampine:


lukechampine:
appveyor does not seem to like the new test constants. Hmm.

lukechampine:
okay. I've narrowed it down to `repairQueueInterval`. Appveyor does not like it when the `repairQueueInterval` is short enough that it runs during the test.
That actually seems like a bug worth investigating. It's weird that it only happens on appveyor, but maybe it can be triggered on Travis as well if we reduce the interval even more.

DavidVorick:
That does seem like something worth investigating.  I'll merge it for now but I'll also make an issue for it.

lukechampine:
nooo those ugly commits

DavidVorick:
ah shoot. sorry about that

PR #1648, opened 2017-04-04
persist inbound info together with node address

GlenDC:
I'm not sure if my execution is perfect here, but the idea is there. Feedback needed I think. The goal is to store the inbound information along with the NetAddress, rather then just the NetAddress.

First I thought a small change could be done, where I'm just serializing the current outbound peers. However the problem is that all peers are disconnected before we serialize, thus that is not possible. Than I noticed that while we do store the nodes, we were storing them as empty structs, so why not store some extra information along with it?

This approach does mean that the update is a bit more invasive than intended, but should still be safe for merging once we agree on all modifications.

I also bumped the gateway persist version, as the format changed a bit. I do support the previous version, so that this node information is not lost after the upgrade.

### DIFF
--- a/modules/gateway/nodes.go
+++ b/modules/gateway/nodes.go
@@ -16,18 +16,29 @@ var (
 	errOurAddress = errors.New("can't add our own address")
 )
 
+type node struct {
+	Inbound bool `json:"inbound"`
+}
+
 // addNode adds an address to the set of nodes on the network.
-func (g *Gateway) addNode(addr modules.NetAddress) error {
+func (g *Gateway) addNode(addr modules.NetAddress, inbound bool) error {
 	if addr == g.myAddr {
 		return errOurAddress
-	} else if _, exists := g.nodes[addr]; exists {
+	} else if node, exists := g.nodes[addr]; exists {
+		if node.Inbound != inbound {
+			node.Inbound = inbound
+			return nil
+		}
+
 		return errNodeExists
 	} else if addr.IsStdValid() != nil {
 		return errors.New("address is not valid: " + string(addr))
 	} else if net.ParseIP(addr.Host()) == nil {
 		return errors.New("address must be an IP address: " + string(addr))
 	}
-	g.nodes[addr] = struct{}{}
+	g.nodes[addr] = &node{
+		Inbound: inbound,
+	}
 	return nil
 }
 
@@ -133,9 +144,10 @@ func (g *Gateway) requestNodes(conn modules.PeerConn) error {
 		return err
 	}
 
+	const inbound = true
 	g.mu.Lock()
 	for _, node := range nodes {
-		err := g.addNode(node)
+		err := g.addNode(node, inbound)
 		if err != nil && err != errNodeExists && err != errOurAddress {
 			g.log.Printf("WARN: peer '%v' sent the invalid addr '%v'", conn.RPCAddr(), node)
 		}
--- a/modules/gateway/persist.go
+++ b/modules/gateway/persist.go
@@ -19,30 +19,81 @@ const (
 // gateway persist file.
 var persistMetadata = persist.Metadata{
 	Header:  "Sia Node List",
+	Version: "0.4.0",
+}
+
+// previousPersistMetadata contains the header and version strings that
+// identify the previous gateway persist file. It allows us to load files in a
+// backward-compatible way.
+var previousPersistMetadata = persist.Metadata{
+	Header:  "Sia Node List",
 	Version: "0.3.3",
 }
 
+type nodePersist struct {
+	NetAddress modules.NetAddress `json:"netaddress"`
+	Inbound    bool               `json:"inbound"`
+}
+
+type gatewayPersist struct {
+	Nodes []nodePersist `json:"nodes"`
+}
+
+// TODO: ensure that we actually have all outbound nodes,
+//       as it seems that peers disconnect
+
 // persistData returns the data in the Gateway that will be saved to disk.
-func (g *Gateway) persistData() (nodes []modules.NetAddress) {
-	for node := range g.nodes {
-		nodes = append(nodes, node)
+func (g *Gateway) persistData() (data gatewayPersist) {
+	for address, node := range g.nodes {
+		data.Nodes = append(data.Nodes, nodePersist{
+			NetAddress: address,
+			Inbound:    node.Inbound,
+		})
 	}
+
 	return
 }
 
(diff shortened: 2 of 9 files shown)


lukechampine:
just to clarify, is the idea here that when the gateway restarts, it would prioritize nodes with `inbound = false` when choosing nodes to reconnect to?
Perhaps we could find a better of way of saving the current peers. That seems more in line with the goal of restoring previously-connected peers on startup.

GlenDC:
Yes that's the goal indeed @lukechampine. As we know outbound nodes are more likely to be trustworthy.

> Perhaps we could find a better of way of saving the current peers. That seems more in line with the goal of restoring previously-connected peers on startup.

Do you have any suggestions or directions?

lukechampine:
I take it back; just prioritizing the nodes you made outbound connections to seems fine. And we probably want to be saving nodes as an object anyway (rather than simply their netaddress), since it makes it easier to add/remove fields in the future. I will take a closer look at this tomorrow.

modules/gateway/nodes.go
```diff
@@ -16,18 +16,29 @@ var (
 	errOurAddress = errors.New("can't add our own address")
 )
 
+type node struct {
```
DavidVorick: needs a docstring

modules/gateway/nodes.go
```diff
@@ -16,18 +16,29 @@ var (
...
-func (g *Gateway) addNode(addr modules.NetAddress) error {
+func (g *Gateway) addNode(addr modules.NetAddress, inbound bool) error {
 	if addr == g.myAddr {
 		return errOurAddress
-	} else if _, exists := g.nodes[addr]; exists {
+	} else if node, exists := g.nodes[addr]; exists {
+		if node.Inbound != inbound {
+			node.Inbound = inbound
+			return nil
+		}
```
DavidVorick: This whole bit of code I think could be rewritten as:

`node.Inbound = inbound; return nil`

But I'm not sure that's the desired behavior. Once we've marked a node as an outbound node, we should probably keep them marked as `outbound` permanently. Additionally, upon startup we should start by connecting to our existing outbound peers instead of just connecting around randomly. If we already have a set of trusted outbound peers, we should stick to that set instead of connecting to potential attackers.

modules/gateway/nodes.go
```diff
@@ -133,9 +144,10 @@ func (g *Gateway) requestNodes(conn modules.PeerConn) error {
 		return err
 	}
 
+	const inbound = true
```
DavidVorick: Curious to set this as a const. I think it's the right thing to do, but I don't think it's a pattern we've used anywhere else.

DavidVorick:
So, one thing that I definitely don't like here is that the variable being saved is `Inbound`. That means that the default is going to be `Inbound: false`, which means that if there's a programmer error a peer which should otherwise be set to inbound is going to be set to outbound.

But it's definitely safer behavior to have programmer error favor setting peers to inbound.

modules/gateway/persist.go
```diff
@@ -19,30 +19,81 @@ const (
 // gateway persist file.
 var persistMetadata = persist.Metadata{
 	Header:  "Sia Node List",
+	Version: "0.4.0",
```
DavidVorick: Metadata version should only be bumped if compatibility is broken with the previous file. And, I'm pretty sure that we can pull this off without breaking compatibility. Further, when we do increment the version, we increment it to the version of Sia that it's going to be released in, so we would set the new version to `"1.2.1"` in this case.

But, as long as the file is already JSON (which it is), we should be able to preserve compatibility by adding new named fields to correspond with the new features. They will be read in as blank, and so set to default values.

modules/gateway/persist.go
```diff
@@ -19,30 +19,81 @@ const (
...
+}
+
+// previousPersistMetadata contains the header and version strings that
+// identify the previous gateway persist file. It allows us to load files in a
+// backward-compatible way.
+var previousPersistMetadata = persist.Metadata{
```
DavidVorick: Again, don't think a version change is necessary here, but if it was, the correct name for this struct would be `v033PersistMetadata`

modules/gateway/persist.go
```diff
@@ -19,30 +19,81 @@ const (
...
+	Header:  "Sia Node List",
 	Version: "0.3.3",
 }
 
+type nodePersist struct {
+	NetAddress modules.NetAddress `json:"netaddress"`
+	Inbound    bool               `json:"inbound"`
+}
+
+type gatewayPersist struct {
```
DavidVorick: These need docstrings

modules/gateway/nodes.go
lukechampine: If Go supported `const` on all types then I would definitely prefer defining everything as `const` unless it needs to be a variable. But since in practice only a few types can be `const`, we typically default to `var` except for certain top-level declarations.

modules/gateway/persist.go
```diff
@@ -19,30 +19,81 @@ const (
...
 		return err
 	}
-	for _, node := range nodes {
-		err := g.addNode(node)
+
+	// add saved nodes
+	for _, node := range data.Nodes {
+		if !node.Inbound {
+			// if outbound node, try to connect to it
+			err := g.Connect(node.NetAddress)
```
lukechampine: I don't think this belongs in the `load` function. Load should just initialize the `Gateway` object. The connecting can be done inside `gateway.New`. Actually, why not just have `threadedPeerManager` prioritize outbound nodes when it's selecting nodes to connect to?

modules/gateway/nodes.go
```diff
@@ -16,18 +16,29 @@ var (
 	errOurAddress = errors.New("can't add our own address")
 )
 
+type node struct {
+	Inbound bool `json:"inbound"`
```
lukechampine: A json tag makes sense if this object is being saved to disk, but it looks like you have the `nodePersist` type for that.

Given that `addNode` is now a bit ugly (taking two args), I propose that we define:
```go
type node struct { 
    Addr     modules.NetAddress `json:"addr"`
    Outbound bool        `json:"outbound"`
}
```
And have `g.addNode` take a `node` (maybe a `*node`, not sure) instead of two args.
Lastly, I think it's fine to save this object in the gateway persist directly (instead of having a separate `nodePersist` type). The `g.nodes` map should be flattened to a `[]node` during the save.

lukechampine:
Moving this to #1758

PR #1663, opened 2017-04-06
remove data protection panics

DavidVorick:
When running not in DEBUG mode, this removal will give us more visibility into
the errors that are happening during normal operation, hopefully enabling us to
more easily discover the root cause.


DavidVorick:
This is because I failed to source the problem. We'll still get SEVERE error messages, but the host won't crash itself. The SEVERE messages will hopefully be numerous enough (multiple instead of just one) that we can still dial in on the problem and fix it.

It is questionable whether crashing at these points was the best decision anyway.

PR #1665, opened 2017-04-06
Added manpage for siad.1 oh what a horror to edit with nroff

vevarm:
So this is the first version of a siad manpage. 

This is an experiment possibly its too hard to maintain and in that case I suggest it to be deleted. But lets give it a try.  

//Cheers

### DIFF
--- a/doc/man/siad.1
+++ b/doc/man/siad.1
@@ -1,0 +1,84 @@
+.\"Created with GNOME Manpages Editor Wizard
+.\"http://sourceforge.net/projects/gmanedit2
+.TH siad 1 "April 6, 2017" "" "siad"
+
+.SH NAME
+
+.B siad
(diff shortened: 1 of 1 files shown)


lukechampine:
Looks like cobra has support for generating man pages: https://github.com/spf13/cobra/blob/6b74a60562f5c1c920299b8f02d153e16f4897fc/doc/man_docs.md

I guess you've done most of the work already, but this would be useful for keeping it up to date.

vevarm:
@lukechampine Keeping it up to date is the hard part. As I understand it cobra is the tool to define and parse commandline, right. I think this is DEFINITELY the way to go if it works ok. I have no problem you scrap what has been done so far.

vevarm:
I dont think I will look into this cobra mysefl in near time, feels to much work to get myself going in a go-lang dev environment. Maybe later.

PR #1669, opened 2017-04-07
Add offset, length, async, http response parameters to download.

RNabel:
- Creates a new `siad` endpoint at `/renter/downloadchunk/<path>?destination=<dst>&chunkindex=<cind>`
- Adds a new `siac` command: `siac renter downloadchunk [path] [chunk index] [destination]` (analog to the order of args in `renter download`)
- Refactors the `download.go` and `downloadqueue.go` to minimize duplicated logic.
- Extends the `download` struct to store metadata for the special case of a chunk download. It could be changed to handle a whole file download as a special case of a chunk download, but this would add a lot of complexity.

This could be the basis for repairing files which are not stored locally.

### DIFF
--- a/api/renter.go
+++ b/api/renter.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/julienschmidt/httprouter"
 	"strconv"
+	"time"
 )
 
 const (
@@ -130,7 +131,7 @@ type (
 
 	// DownloadQueue contains the renter's download queue.
 	RenterDownloadQueue struct {
-		Downloads []modules.DownloadInfo `json:"downloads"`
+		Downloads []DownloadInfo `json:"downloads"`
 	}
 
 	// RenterFiles lists the files known to the renter.
@@ -152,6 +153,15 @@ type (
 	// RenterShareASCII contains an ASCII-encoded .sia file.
 	RenterShareASCII struct {
 		ASCIIsia string `json:"asciisia"`
+	}
+
+	DownloadInfo struct {
+		SiaPath     string    `json:"siapath"`
+		Destination string    `json:"destination"`
+		Filesize    uint64    `json:"filesize"`
+		Received    uint64    `json:"received"`
+		StartTime   time.Time `json:"starttime"`
+		Error       string    `json:"error"`
 	}
 )
 
@@ -280,6 +290,12 @@ func (api *API) renterContractsHandler(w http.ResponseWriter, _ *http.Request, _
 
 // renterDownloadsHandler handles the API call to request the download queue.
 func (api *API) renterDownloadsHandler(w http.ResponseWriter, _ *http.Request, _ httprouter.Params) {
+	//dlq := api.renter.DownloadQueue()
+	//for i := range dlq {
+	//	el := dlq[i]
+	//
+	//} TODO: translate a modules.DownloadInfo into a api.DownloadInfo object.
+
 	WriteJSON(w, RenterDownloadQueue{
 		Downloads: api.renter.DownloadQueue(),
 	})
@@ -406,15 +422,22 @@ func parseDownloadParameters(w http.ResponseWriter, req *http.Request, ps httpro
 	asyncparam := req.FormValue("async")
 
 	// Parse the offset and length parameters. TODO(rnabel): Handle empty string.
-	offset, err := strconv.ParseUint(offsetparam, 10, 64)
-	if err != nil {
-		return nil, &Error{"could not decode the offset as uint64: " +
-			err.Error()}
+	var offset, length uint64
+	var err error
+	if len(offsetparam) > 0 {
+		offset, err = strconv.ParseUint(offsetparam, 10, 64)
+		if err != nil {
+			return nil, &Error{"could not decode the offset as uint64: " +
+				err.Error()}
+		}
 	}
-	length, err := strconv.ParseUint(lengthparam, 10, 64)
-	if err != nil {
-		return nil, &Error{"could not decode the length as uint64: " +
-			err.Error()}
+	if len(lengthparam) > 0 {
+		length, err = strconv.ParseUint(lengthparam, 10, 64)
+		if err != nil {
+			return nil, &Error{"could not decode the length as uint64: " +
+				err.Error()}
+		}
+
 	}
 	// Verify that if either offset or length have been provided that both were provided.
 	offparampassed := len(offsetparam) > 0
--- a/api/renter.go
+++ b/api/renter.go
@@ -481,6 +481,10 @@ func (api *API) parseAndValidateDownloadParameters(w http.ResponseWriter, req *h
 		return nil, errmsg
 	}
 
+	if async && httpresp {
+		return nil, &Error{"only one of the async and httpresp flags can be specified."}
+	}
+
 	siapath := strings.TrimPrefix(ps.ByName("siapath"), "/") // Sia file name.
 
 	// Lookup the file associated with the nickname.
--- a/api/renter.go
+++ b/api/renter.go
@@ -358,6 +358,7 @@ func (api *API) renterDownloadHandler(w http.ResponseWriter, req *http.Request, 
 	p, errmsg := parseDownloadParameters(w, req, ps)
 	if errmsg != nil {
 		WriteError(w, *errmsg, http.StatusBadRequest)
+		return
 	}
 
 	if p.Async { // Create goroutine if `async` param set.
--- a/api/renter.go
+++ b/api/renter.go
@@ -410,15 +410,12 @@ func (api *API) renterDownloadHandler(w http.ResponseWriter, req *http.Request, 
 
 // renterDownloadAsyncHandler handles the API call to download a file asynchronously.
 func (api *API) renterDownloadAsyncHandler(w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
+	// Set async flag to true.
+	req.Form.Add("async", "true")
 	p, errmsg := api.parseAndValidateDownloadParameters(w, req, ps)
 	if errmsg != nil {
 		WriteError(w, *errmsg, http.StatusBadRequest)
 	}
-
-	// Set async param to true.
-	// This parameter may be used in the future so it is set for safety reasons,
-	// its intended use is to determine whether to call DownloadSection in a goroutine.
-	p.Async = true
 
 	go api.renter.DownloadSection(p)
 
--- a/api/renter.go
+++ b/api/renter.go
@@ -386,9 +386,9 @@ func (api *API) renterDownloadHandler(w http.ResponseWriter, req *http.Request, 
 	}
 
 	if params.Async { // Create goroutine if `async` param set.
-		go api.renter.DownloadSection(params)
+		go api.renter.Download(params)
 	} else {
-		err := api.renter.DownloadSection(params)
+		err := api.renter.Download(params)
 		if err != nil {
 			WriteError(w, Error{"download failed: " + err.Error()}, http.StatusInternalServerError)
 			return
--- a/api/renter.go
+++ b/api/renter.go
@@ -444,8 +444,7 @@ func (api *API) parseDownloadParameters(w http.ResponseWriter, req *http.Request
 	if len(lengthparam) > 0 {
 		length, err = strconv.ParseUint(lengthparam, 10, 64)
 		if err != nil {
-			return nil, &Error{"could not decode the length as uint64: " +
-				err.Error()}
+			return nil, build.ExtendErr("could not decode the length as uint64: ", err)
 		}
 	}
 	// Verify that if either offset or length have been provided that both were provided.
--- a/api/renter.go
+++ b/api/renter.go
@@ -151,6 +151,7 @@ type (
 		ASCIIsia string `json:"asciisia"`
 	}
 
+	// DownloadInfo contains all client-facing information of a file.
 	DownloadInfo struct {
 		SiaPath     string    `json:"siapath"`
 		Destination string    `json:"destination"`
@@ -612,6 +613,7 @@ func (api *API) renterUploadHandler(w http.ResponseWriter, req *http.Request, ps
 	WriteSuccess(w)
 }
 
+// stringToBool converts "true" and "false" strings to their respective boolean value and returns an error if conversion is not possible.
 func stringToBool(param string) (bool, *Error) {
 	// Parse the async parameter.
 	var out bool
--- a/api/renter.go
+++ b/api/renter.go
@@ -403,7 +403,9 @@ func (api *API) renterDownloadHandler(w http.ResponseWriter, req *http.Request, 
 		}
 	}
 
-	WriteSuccess(w)
+	if !p.Httpresp {
+		WriteSuccess(w)
+	}
 }
 
 // renterDownloadAsyncHandler handles the API call to download a file asynchronously.
--- a/api/renter.go
+++ b/api/renter.go
@@ -19,10 +19,6 @@ import (
 	"time"
 )
 
-const (
-	maxUint64 = ^uint64(0)
-)
-
 var (
 	// recommendedHosts is the number of hosts that the renter will form
 	// contracts with if the value is not specified explicitly in the call to
@@ -290,14 +286,33 @@ func (api *API) renterContractsHandler(w http.ResponseWriter, _ *http.Request, _
 
 // renterDownloadsHandler handles the API call to request the download queue.
 func (api *API) renterDownloadsHandler(w http.ResponseWriter, _ *http.Request, _ httprouter.Params) {
-	//dlq := api.renter.DownloadQueue()
-	//for i := range dlq {
-	//	el := dlq[i]
-	//
-	//} TODO: translate a modules.DownloadInfo into a api.DownloadInfo object.
+	dlq := api.renter.DownloadQueue()
 
+	// Translate []modules.DownloadInfo to []api.DownloadInfo.
+	downloads := make([]DownloadInfo, len(dlq))
+	for i := range dlq {
+		d := dlq[len(dlq)-i-1]
+
+		// Find the destination of the download.
+		dstwriter, ok := (d.Destination).(*modules.DownloadFileWriter)
+		var dst string
+		if ok {
+			dst = dstwriter.Location
+		} else {
+			dst = "httpresp"
+		}
+
+		downloads[i] = DownloadInfo{
+			SiaPath:     d.SiaPath,
+			Destination: dst,
+			Filesize:    d.Filesize,
+			StartTime:   d.StartTime,
+			Received:    d.Received,
+			Error:       d.Error,
+		}
+	}
 	WriteJSON(w, RenterDownloadQueue{
-		Downloads: api.renter.DownloadQueue(),
+		Downloads: downloads,
 	})
 }
 
@@ -371,7 +386,7 @@ func (api *API) renterDeleteHandler(w http.ResponseWriter, req *http.Request, ps
 
 // renterDownloadHandler handles the API call to download a file.
 func (api *API) renterDownloadHandler(w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
-	p, errmsg := parseDownloadParameters(w, req, ps)
+	p, errmsg := api.parseAndValidateDownloadParameters(w, req, ps)
 	if errmsg != nil {
 		WriteError(w, *errmsg, http.StatusBadRequest)
 		return
@@ -392,7 +407,7 @@ func (api *API) renterDownloadHandler(w http.ResponseWriter, req *http.Request, 
 
 // renterDownloadAsyncHandler handles the API call to download a file asynchronously.
 func (api *API) renterDownloadAsyncHandler(w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
-	p, errmsg := parseDownloadParameters(w, req, ps)
+	p, errmsg := api.parseAndValidateDownloadParameters(w, req, ps)
 	if errmsg != nil {
 		WriteError(w, *errmsg, http.StatusBadRequest)
 	}
@@ -407,7 +422,7 @@ func (api *API) renterDownloadAsyncHandler(w http.ResponseWriter, req *http.Requ
 	WriteSuccess(w)
 }
 
-func parseDownloadParameters(w http.ResponseWriter, req *http.Request, ps httprouter.Params) (*modules.RenterDownloadParameters, *Error) {
+func (api *API) parseAndValidateDownloadParameters(w http.ResponseWriter, req *http.Request, ps httprouter.Params) (*modules.RenterDownloadParameters, *Error) {
 	destination := req.FormValue("destination")
 
 	// The offset and length in bytes.
@@ -467,9 +482,15 @@ func parseDownloadParameters(w http.ResponseWriter, req *http.Request, ps httpro
 
 	siapath := strings.TrimPrefix(ps.ByName("siapath"), "/") // Sia file name.
 
+	// Lookup the file associated with the nickname.
+	file, exists := api.renter.GetFile(siapath)
+	if !exists {
+		return nil, &Error{Message: "file could not be found"}
+	}
+
 	if !offparampassed { // Determine if entire file is to be downloaded.
 		offset = 0
-		length = maxUint64
+		length = file.Filesize
 	}
 
 	// Instantiate the correct DownloadWriter implementation
--- a/api/renter.go
+++ b/api/renter.go
@@ -15,6 +15,7 @@ import (
 	"github.com/NebulousLabs/Sia/types"
 
 	"github.com/julienschmidt/httprouter"
+	"strconv"
 )
 
 var (
@@ -358,6 +359,25 @@ func (api *API) renterDownloadHandler(w http.ResponseWriter, req *http.Request, 
 	}
 
 	err := api.renter.Download(strings.TrimPrefix(ps.ByName("siapath"), "/"), destination)
+	if err != nil {
+		WriteError(w, Error{"download failed: " + err.Error()}, http.StatusInternalServerError)
+		return
+	}
+
+	WriteSuccess(w)
+}
+
+// renterDownloadchunkHandler handles the API call to download a specific chunk from a file.
+func (api *API) renterDownloadchunkHandler(w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
+	destination := req.FormValue("destination")
+	cindparam := req.FormValue("chunkindex")
+
+	cindex, err := strconv.ParseUint(cindparam, 10, 64)
+	if err != nil {
+		WriteError(w, Error{"could not decode the chunk index as uint64: " + err.Error()}, http.StatusBadRequest)
+	}
+
+	err = api.renter.DownloadChunk(strings.TrimPrefix(ps.ByName("siapath"), "/"), destination, cindex)
 	if err != nil {
 		WriteError(w, Error{"download failed: " + err.Error()}, http.StatusInternalServerError)
 		return
--- a/api/renter.go
+++ b/api/renter.go
@@ -505,7 +505,7 @@ func (api *API) parseAndValidateDownloadParameters(w http.ResponseWriter, req *h
 		if !filepath.IsAbs(destination) {
 			return nil, &Error{"destination must be an absolute path"}
 		}
-		dw = modules.NewDownloadFileWriter(destination)
+		dw = modules.NewDownloadFileWriter(destination, offset, length)
 	}
 
 	return &modules.RenterDownloadParameters{
--- a/api/renter.go
+++ b/api/renter.go
@@ -402,7 +402,7 @@ func parseDownloadParameters(w http.ResponseWriter, req *http.Request, ps httpro
 
 	// Determines whether to return on completion of download or straight away.
 	// If httprespparam is present, this parameter is ignored.
-	asyncparam := req.FormValue("httpresp")
+	asyncparam := req.FormValue("async")
 
 	// Parse the offset and length parameters. TODO(rnabel): Handle empty string.
 	offset, err := strconv.ParseUint(offsetparam, 10, 64)
--- a/api/renter.go
+++ b/api/renter.go
@@ -410,11 +410,18 @@ func (api *API) renterDownloadHandler(w http.ResponseWriter, req *http.Request, 
 
 // renterDownloadAsyncHandler handles the API call to download a file asynchronously.
 func (api *API) renterDownloadAsyncHandler(w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
+	// Ensure that the async flag is not set to false.
+	if strings.Contains(req.URL.RawQuery, "async=false") {
+		WriteError(w, Error{"/downloadasync does not support allow requests containing `async=false`."}, http.StatusBadRequest)
+		return
+	}
+
 	// Set async flag to true.
-	req.Form.Add("async", "true")
+	req.URL.RawQuery += "&async=true"
 	p, errmsg := api.parseAndValidateDownloadParameters(w, req, ps)
 	if errmsg != nil {
 		WriteError(w, *errmsg, http.StatusBadRequest)
+		return
 	}
 
 	go api.renter.DownloadSection(p)
@@ -436,7 +443,7 @@ func (api *API) parseAndValidateDownloadParameters(w http.ResponseWriter, req *h
 	// If httprespparam is present, this parameter is ignored.
 	asyncparam := req.FormValue("async")
 
-	// Parse the offset and length parameters. TODO(rnabel): Handle empty string.
+	// Parse the offset and length parameters.
 	var offset, length uint64
 	var err error
 	if len(offsetparam) > 0 {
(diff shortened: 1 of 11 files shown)


DavidVorick:
Preliminary review looks good. Probably don't need a new API endpoint, instead can add a parameter to the existing /download call. I am also thinking it might make more sense to have 'offset' and 'size' instead of a chunk index. The advantage to that is that we can more easily support random access in the future.

Immediately, the back end can re-interpret a range+offset as a set of chunks, and then download the ones needed and truncate as necessary. In the future, when downloading partial chunks and partial sectors is allowed, we can optimize further without changing the API

lukechampine:
If we decide to keep the new API endpoint, I suggest making it write the chunk to the response body instead of to a file. This would make it trivial to perform streaming downloads.

fwiw, I think I would prefer a new endpoint. There are different goals and requirements for low-level vs. high-level contract access, so it makes sense to separate them.
In that vein, I can see the advantages of using offset+length instead of chunk index. For example, it would facilitate packing multiple files into a single chunk. On the other hand, it kind of hides the cost of doing that: even if you only want 3 bytes of a chunk, behind the scenes we have to download the whole thing. There's also the issue of end-of-file chunk padding. If the caller specifies offset+length, then they are responsible for knowing the original file size and trimming off the padding, whereas if they specify a chunk index, the backend can know that it's the last chunk and trim it accordingly.

DavidVorick:
> behind the scenes we have to download the whole thing

This will only be true for a little while, the protocol was written to support partial sector downloads. It may even be something we can support quickly, I believe the hosts already support partial downloads.

We should consider the API design in terms of what makes sense in 6 months when all the easy/obvious optimizations are in place. 

Ability to stream also seems to me like something you could toggle with a bool. If toggled, response in body, otherwise sent to a file.

I'm not sure how easily we can re-write the backend to pull a chunk steam to a response body.  But wouldn't be difficult if we did the auto-repair stuff well (just recycle that code)

RNabel:
@DavidVorick agreed, adding the parameters to an existing endpoint probably makes the API easier to use and less convoluted.

Re: offset+length, totally for it, initially thought that this may be implemented client-side. As @lukechampine suggested, this glosses over the actual cost of a download, which is why I felt like exposing the chunks directly is the way to go. Thinking about it now, letting the low-level abstractions bleed through the API may be confusing for someone new to Sia.

Re: response in the body, was thinking about that too. It would probably require the output `io.Writer` to be passed around in the `download` object, and the `finishedChunks` list be changed from `[]pieceSet` to `map[int]pieceSet` to be able to take a slice of a file with as little added complexity as possible (e.g. most loops would stay the same, taking the mapped integer as chunk index instead of the vanilla list index as currently).

### Action items (please let me know if you agree with these):

- Merge the additional API endpoint into the existing `/download` endpoint
- Add a response-in-body `bool`
- Change from chunk num to offset+length requests
- Extend `download` to store the `io.Writer` that the output is written to. Haven't done much web stuff in golang so if this is unlikely to work with the current setup, please let me know.
- Work out what abstraction is required for downloading lists of chunks (e.g. change from `[]pieceSet` to `map[int]pieceSet`)

lukechampine:
I still believe this would be simpler with its own endpoint. The conditional "write to file"/"write to body" is a red flag for me. If we overload an endpoint with too much flexibility, you have to manage the interactions of all permutations of its flags, which can get hairy. For example, if we add a "download in order" flag, we need to explicitly define (and probably document) how that interacts with writing to the body.

Re `io.Writer`: since individual chunks are small, it should be sufficient to just pass the chunk around in memory and let the API handler write it to the response. e.g.:
```go
func downloadHandler(w http.ResponseWriter, req *http.Request) {
    chunk, err := renter.DownloadChunk(req.FormValue("chunkindex"))
    if err != nil {
        WriteError(w, err)
        return
    }
    w.Write(chunk)
}
```
(I just realized that this brings up another question: what do we write to `w` if the download fails?)

RNabel:
This is not ready for a full review, but if you have comments re the changes I listed, please let me know.

gowithplanb:
> what do we write to w if the download fails?
The function/call should give an error indicating what failed and report that the file/object/chunk was not available. The download itself could also fail, or take too much time. Requestors need to be aware of this and act accordingly.

api/renter.go
```diff
@@ -350,49 +387,156 @@ func (api *API) renterDeleteHandler(w http.ResponseWriter, req *http.Request, ps
...
 }
 
 // renterDownloadAsyncHandler handles the API call to download a file asynchronously.
 func (api *API) renterDownloadAsyncHandler(w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
-	destination := req.FormValue("destination")
-	// Check that the destination path is absolute.
-	if !filepath.IsAbs(destination) {
-		WriteError(w, Error{"destination must be an absolute path"}, http.StatusBadRequest)
+	// Ensure that the async flag is not set to false.
+	if strings.Contains(req.URL.RawQuery, "async=false") {
```
RNabel: This means that the destination cannot include the string `async=false`. I'm open for change suggestions.

modules/renter.go
```diff
@@ -50,12 +54,92 @@ type Allowance struct {
...
+		firstByteIndex: int(offset),
+		length:         int(length),
+		buffer:         make([]byte, length),
+		offsets:        make(map[int]int),
+	}
+}
+
+// WriteAt buffers parts of the file until the entire file can be
+// flushed to the client. Returns the number of bytes written or an error.
+func (dw *DownloadHttpWriter) WriteAt(b []byte, off int64) (int, error) {
```
lukechampine: In the worst case, this will result in buffering the entire file in memory. This might work fine for downloading a single chunk (although chunks can still be quite large), but later someone else might come along and assume that `DownloadHttpWriter` could be used to stream a whole file.

More generally: there is a lot of code sharing going on here that makes me uneasy. I think `Download` and `DownloadSection` are fundamentally different operations, so sharing code between them is a bonus, not a requirement.
Perhaps we could have some function that downloaded a chunk and held it in memory, and then `Download` would write it out to disk and `DownloadSection` would write it out to the HTTP response. But if the current renter design makes that difficult or impractical, I'd rather we just write the `DownloadSection` logic separately. Long-term, we will want to optimize `DownloadSection` in ways that aren't applicable to `Download`. And I think @DavidVorick is itching to redesign the download code anyway, so all the more reason to keep them separate.

api/renter.go
```diff
@@ -275,8 +287,33 @@ func (api *API) renterContractsHandler(w http.ResponseWriter, _ *http.Request, _
...
+	for i := range dlq {
+		d := dlq[len(dlq)-i-1]
+
+		// Find the destination of the download.
+		dstwriter, ok := (d.Destination).(*modules.DownloadFileWriter)
+		var dst string
+		if ok {
+			dst = dstwriter.Location
+		} else {
+			dst = "httpresp"
```
DavidVorick: Typically we like to have a separate bool indicating special use. Instead of setting the dstwriter.Location to a magic value, to get httpresp downloads you should have both an empty dstwriter.Location (which we might prefer to call Destination) and a bool indicating that it's going to be streamed. If the streaming bool is set and the destination is not empty, an error is returned because that's an improper request.

api/renter.go
```diff
@@ -350,49 +387,141 @@ func (api *API) renterDeleteHandler(w http.ResponseWriter, req *http.Request, ps
 
 // renterDownloadHandler handles the API call to download a file.
 func (api *API) renterDownloadHandler(w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
-	destination := req.FormValue("destination")
-	// Check that the destination path is absolute.
-	if !filepath.IsAbs(destination) {
-		WriteError(w, Error{"destination must be an absolute path"}, http.StatusBadRequest)
+	p, errmsg := api.parseAndValidateDownloadParameters(w, req, ps)
```
DavidVorick: `p` is not a typical name, would prefer the name `params` which is more commonly used throughout the project.

api/renter.go
```diff
@@ -350,49 +387,156 @@ func (api *API) renterDeleteHandler(w http.ResponseWriter, req *http.Request, ps
...
+	asyncparam := req.FormValue("async")
+
+	// Parse the offset and length parameters.
+	var offset, length uint64
+	var err error
+	if len(offsetparam) > 0 {
+		offset, err = strconv.ParseUint(offsetparam, 10, 64)
+		if err != nil {
+			return nil, &Error{"could not decode the offset as uint64: " +
+				err.Error()}
```
DavidVorick: generally we won't make strings like this wrap to multiple lines.

There's a function called `ExtendErr` in the build package that we use when extending errors. There are vague plans in the future to have `build.ExtendErr` return a different error type which remembers each individual underlying error. That will make it easy for us to check equivalence, for example with something like `buildErr.Is(modules.SpecificError)`, such that `Is` will return true if the original error was `modules.SpecificError` or has ever been extended by `modules.SpecificError`.

We don't have it yet, but it's something we've been wanting for a while. At that point we'll probably break it all out in to a separate error package as well.

`return nil, build.ExtendErr("could not decode the offset as unit64", err)`

DavidVorick: anyway, would prefer you use `build.ExtendErr` here

api/renter.go
```diff
@@ -350,49 +387,156 @@ func (api *API) renterDeleteHandler(w http.ResponseWriter, req *http.Request, ps
...
 		}
 	}
+	if len(lengthparam) > 0 {
+		length, err = strconv.ParseUint(lengthparam, 10, 64)
+		if err != nil {
+			return nil, &Error{"could not decode the length as uint64: " +
+				err.Error()}
```
DavidVorick: same anywhere you are adding context to an error.

api/renter.go
```diff
@@ -350,49 +387,156 @@ func (api *API) renterDeleteHandler(w http.ResponseWriter, req *http.Request, ps
...
+		}
+	}
+	// Verify that if either offset or length have been provided that both were provided.
+	offparampassed := len(offsetparam) > 0
+	lenparampassed := len(lengthparam) > 0
+	if (offparampassed || lenparampassed) &&
+		!(offparampassed && lenparampassed) {
```
DavidVorick: this if statement can definitely be one line. We generally try not to wrap code because it tends to reduce readability. There are of course exceptions, but I think that's generally not until you are approaching 150 chars.

api/renter.go
```diff
@@ -350,49 +387,156 @@ func (api *API) renterDeleteHandler(w http.ResponseWriter, req *http.Request, ps
...
+	// Parse the async parameter.
+	async, errmsg := stringToBool(asyncparam)
+	if errmsg != nil {
+		return nil, errmsg
+	}
 
-	WriteSuccess(w)
+	if async && httpresp {
+		return nil, &Error{"only one of the async and httpresp flags can be specified."}
+	}
```
DavidVorick: should also verify that if `httpresp` is set, `destination` is also not set.

DavidVorick: and I guess that if `async` is set, `destination` is also set

api/renter.go
```diff
@@ -350,49 +387,156 @@ func (api *API) renterDeleteHandler(w http.ResponseWriter, req *http.Request, ps
...
+
+	siapath := strings.TrimPrefix(ps.ByName("siapath"), "/") // Sia file name.
+
+	// Lookup the file associated with the nickname.
+	file, exists := api.renter.GetFile(siapath)
+	if !exists {
+		return nil, &Error{Message: "download failed: no file with that path"}
```
DavidVorick: probably worth printing which path was being requested. This will help us if the GUI for example is hitting this error and the user is showing us the error - we'll know better what's going on.

api/renter.go
```diff
@@ -350,49 +387,156 @@ func (api *API) renterDeleteHandler(w http.ResponseWriter, req *http.Request, ps
...
+	}
+
+	if !offparampassed { // Determine if entire file is to be downloaded.
+		offset = 0
+		length = file.Filesize
+	}
```
DavidVorick: Should move this code so it appears immediately after you validate the offset and length parameters

DavidVorick: comment can appear above the code instead of alongside the if statement

api/renter.go
```diff
@@ -489,3 +633,19 @@ func (api *API) renterUploadHandler(w http.ResponseWriter, req *http.Request, ps
 	}
 	WriteSuccess(w)
 }
+
+// stringToBool converts "true" and "false" strings to their respective boolean value and returns an error if conversion is not possible.
+func stringToBool(param string) (bool, *Error) {
```
DavidVorick: this belongs in `scan.go`

api/renter.go
```diff
@@ -489,3 +633,19 @@ func (api *API) renterUploadHandler(w http.ResponseWriter, req *http.Request, ps
...
+	// Parse the async parameter.
+	var out bool
+	switch {
+	case param == "true":
+		out = true
+	case len(param) == 0 || param == "false":
+		out = false
+	default:
+		return false, &Error{"asyncparam has to be empty, \"true\" or \"false\""}
```
DavidVorick: error needs to be more generic, perhaps "boolean parameters need to take the form "true" or "false""

api/renter.go
```diff
@@ -350,49 +387,156 @@ func (api *API) renterDeleteHandler(w http.ResponseWriter, req *http.Request, ps
...
+	// Instantiate the correct DownloadWriter implementation
+	// (e.g. content written to file or response body).
+	var dw modules.DownloadWriter
+	if httpresp {
+		dw = modules.NewDownloadHttpWriter(w, offset, length)
+	} else {
+		// Ensure that destination is valid beforehands.
+		// Check that the destination path is absolute.
+		if !filepath.IsAbs(destination) {
+			return nil, &Error{"destination must be an absolute path"}
```
DavidVorick: @lukechampine should these errors be pointers? Or just stack objects?

api/renter.go
```diff
@@ -350,49 +387,141 @@ func (api *API) renterDeleteHandler(w http.ResponseWriter, req *http.Request, ps
...
+	if errmsg != nil {
+		WriteError(w, *errmsg, http.StatusBadRequest)
 		return
 	}
 
-	err := api.renter.Download(strings.TrimPrefix(ps.ByName("siapath"), "/"), destination)
-	if err != nil {
-		WriteError(w, Error{"download failed: " + err.Error()}, http.StatusInternalServerError)
-		return
+	if p.Async { // Create goroutine if `async` param set.
```
DavidVorick: I think that the input validation should probably be done alongside the use. When you validate separately, the likelihood of missing something goes up, especially if in the future this part of the code is modified. The requirements may change, and then the writer and reviewer may forget to update the validator. But if the validation happens alongside the use, then during review it's a lot easier to realize that you've written code which invalidates your validation.

api/renter.go
DavidVorick: Per my comment below, I think it would be good to separate the parsing and the validating. Parsing everything in one place makes sense to me, but I think that validation should happen alongside the use (makes it easier during both writing, review, and future modification to verify that validation + use match up correctly, b/c the code is all in the same place)

api/renter.go
```diff
@@ -350,49 +387,141 @@ func (api *API) renterDeleteHandler(w http.ResponseWriter, req *http.Request, ps
...
+		err := api.renter.DownloadSection(p)
+		if err != nil {
+			WriteError(w, Error{"download failed: " + err.Error()}, http.StatusInternalServerError)
+			return
+		}
 	}
 
-	WriteSuccess(w)
+	if !p.Httpresp {
+		WriteSuccess(w)
```
DavidVorick: This comment should be in the code lol

DavidVorick: Best practice (not sure if we follow it everywhere yet) is to `return` immediately after calling `WriteSuccess` or `WriteError`, even if the function ends directly below. This ensures that future extensions to the function will not inadvertently introduce an error, and also it's easier for reviewers to see that the WriteX calls are being used correctly.

api/renter.go
DavidVorick: `if req.FormValue("async") == "false" {`

api/renter.go
```diff
@@ -350,49 +387,156 @@ func (api *API) renterDeleteHandler(w http.ResponseWriter, req *http.Request, ps
...
+	// Ensure that the async flag is not set to false.
+	if strings.Contains(req.URL.RawQuery, "async=false") {
+		WriteError(w, Error{"/downloadasync does not support allow requests containing `async=false`."}, http.StatusBadRequest)
 		return
 	}
 
-	siapath := strings.TrimPrefix(ps.ByName("siapath"), "/")
-	fileExists := false
+	// Set async flag to true.
+	req.URL.RawQuery += "&async=true"
```
DavidVorick: This strikes me as the wrong way to tag on the async flag.

api/renter.go
```diff
@@ -350,49 +387,156 @@ func (api *API) renterDeleteHandler(w http.ResponseWriter, req *http.Request, ps
...
+	p, errmsg := api.parseAndValidateDownloadParameters(w, req, ps)
+	if errmsg != nil {
+		WriteError(w, *errmsg, http.StatusBadRequest)
+		return
+	}
+
+	go api.renter.DownloadSection(p)
+
+	WriteSuccess(w)
+}
```
DavidVorick: @lukechampine looks like we are essentially deprecating `/downloadasync` in favor of adding an async flag to `/download`? What were the reasons for not just doing it this way in the first place?

Fwiw I am in favor of keeping all these flags and such under the same endpoint. I know you think that this is going to complicate the interactions, but all the parameters I've seen here don't seem to complicate things too much. For example, a full download really is just a special case of an offset+len download.

Permutations of flags is pretty easy to manage so far. Basic intuition can tell you which flags shouldn't be used together (for example, can't set a destination and also set the file to be streamed), and simple validation makes sure that nothing is inappropriately interacting.

modules/renter.go
DavidVorick: `Download` is just a special case of `DownloadSection` where offset=0 and len=filesize. I don't think they are fundamentally different operations at all.

Any optimizations we want to add to `DownloadSection` will also apply to `Download`. Small files are eventually going to be fitting inside of chunks, and if there's a large file with some amount of padding in the final chunk, you want to be able to skip the download of that padding, meaning you are going to be sending offsets to the host anyway.

Overall I think the logical combining of these operations actually does make a lot of sense.

api/renter.go
DavidVorick: you gave an example of a `downloadinorder` flag. We'd handle that by defaulting the flag to `true` (meaning it's only false if explicitly written false), and then making sure that if the user sets the flag to false, they don't set it along side http streaming. Out-of-order downloads just don't make sense if you are doing streaming (it's possible, but would overly complicate the api).

api/renter.go
```diff
@@ -350,49 +387,138 @@ func (api *API) renterDeleteHandler(w http.ResponseWriter, req *http.Request, ps
...
+	}
+
+	if async && httpresp {
+		return nil, &Error{"only one of the async and httpresp flags can be specified."}
+	}
+
+	siapath := strings.TrimPrefix(ps.ByName("siapath"), "/") // Sia file name.
+
+	// Lookup the file associated with the nickname.
+	file, exists := api.renter.GetFile(siapath)
```
DavidVorick: golang typically doesn't like using the word `Get`. Simply calling the function `File` I think is more idiomatic.

api/renter_test.go
```diff
@@ -9,16 +9,19 @@ import (
...
 
+	"bytes"
+	"fmt"
 	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/modules/renter"
 	"github.com/NebulousLabs/Sia/modules/renter/contractor"
 	"github.com/NebulousLabs/Sia/types"
 	"github.com/NebulousLabs/fastrand"
+	"os"
```
DavidVorick: stdlib imports go in the top section, Sia libs go in the middle section, external libs go in the bottom section (including non-Sia NebulousLabs libs)

api/renter_test.go
```diff
@@ -101,63 +216,227 @@ func TestRenterDownloadError(t *testing.T) {
...
+
+func TestRenterDownloadHttpRespOffsetAndLengthManyChunks(t *testing.T) {
+	filesize := int64(modules.SectorSize) * 5
+	setupDownloadTest(t, filesize, 150, 3*filesize/4, true)
+}
+
+func TestRenterDownloadHttpRespOffsetAndLengthManyChunksSubsetOfChunks(t *testing.T) {
+	filesize := int64(modules.SectorSize) * 5
+	setupDownloadTest(t, filesize, 150, 1*filesize/4, true)
+}
```
DavidVorick: These should all be converted to a single larger table-driven test. Here's an example:

https://github.com/NebulousLabs/Sia/blob/master/modules/renter/erasure_test.go#L27

modules/renter.go
lukechampine: To clarify, I mean that streaming vs. parallel optimizations are going to be necessarily different. I agree that `Download` and `DownloadSection` will share optimizations. However, as currently written, `DownloadSection` is design to meet a streaming use-case, whereas `Download` is not. If we want to merge their functionality, we need to figure out a way to bridge that gap. One design would be to have both `Download` and `DownloadSection` support streaming or parallel downloads, e.g. with a `stream` flag.

api/renter.go
lukechampine: also, use `` to write string literals containing `"`

modules/renter.go
RNabel: **Re: Buffer**: I've got a pretty straight forward fix for the buffering problem, should get it out by EOD. I had a working version of this in previous commits but chose to take it out because the tests simulate a too-perfect environment (e.g. chunks always arrive in order because they are scheduled in order).

**Re: Download overlap**: The actual downloading logic hasn't changed, just the destination of the data. The initial layout of the section download was completely separate (see initial reviews), which introduced many corner cases, made the code unwieldy and prone to logic duplication.
@lukechampine could you elaborate on what you mean by streaming vs parallel optimizations? Downloads are currently performed in parallel and order is only introduced by the order in which chunks are scheduled for download (which is currently linear).
If a `stream` flag were to be introduced to speed up streaming downloads it could be used with file-based and streaming downloads alike needing changes only to be made in one place.

DavidVorick: Yeah, basically to convert the current renter from parallel -> stream, the only thing you need to do is send the finished chunk to a merger instead of to a file. The merger will block until the finished chunk is written to the stream, and it won't write the finished chunk to the stream until all previous chunks are written.

You may want to add some logic that more carefully selects how slow flights of hosts perform downloads, but it would be fairly easy to set a flag such that slow hosts ignore streaming chunk downloads.

Overall, I  think 90% of the optimizations are going to apply to both cases, and it'd be better to have the code heavily overlap with some flags rather than have two separate, but nearly identical, and both highly complex, sets of code to manage the downloads.

api/renter.go
```diff
@@ -350,49 +387,156 @@ func (api *API) renterDeleteHandler(w http.ResponseWriter, req *http.Request, ps
...
 	}
 
-	WriteSuccess(w)
+	if !p.Httpresp {
+		WriteSuccess(w)
+	}
 }
 
 // renterDownloadAsyncHandler handles the API call to download a file asynchronously.
 func (api *API) renterDownloadAsyncHandler(w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
```
DavidVorick: I'm thinking the best thing to do here is to modify the request to expressly set async to true, and then pass it to the other handler. This function basically only needs to check for `async=false` and then tag on `async=true` if it's not a already there.

modules/renter.go
```diff
@@ -50,12 +54,92 @@ type Allowance struct {
...
+// DownloadFileWriter is a file-backed implementation of DownloadWriter.
+type DownloadFileWriter struct {
+	f        *os.File
+	Location string
+	offset   uint64
+}
+
+// NewDownloadFileWriter creates a new instance of a DownloadWriter backed by the file named.
+func NewDownloadFileWriter(fname string, offset, length uint64) *DownloadFileWriter {
+	l, _ := os.OpenFile(fname, os.O_CREATE|os.O_WRONLY, defaultFilePerm)
```
johnathanhowell: should the `DownloadFileWriter` have a `Close()` method that closes this handle?

RNabel: ah yes. I'll track how many bytes were written and will close it in `WriteAt()` once `length` bytes have been written.
Alternatively, do you think `DownloadWriter` should implement `io.Closer` and have an explicit `Close()` method?

johnathanhowell: Having a `Close()` method is probably overkill for the `DownloadWriter` interface, I think closing once `length` bytes have written is sufficient.

RNabel: Okay, I'll add a `Close()` method then :)

DavidVorick: What happens if you call `WriteAt()` out of order with the DownloadFileWriter? Going to disk makes sense, but if you are doing it in a stream, it's going to need to block so that `siad` doesn't consume more memory than it is allowed to (if it returns, the calling thread will send off another flight of downloads, consuming an extra 40MB of RAM).

But if it's blocking, it means that you have to make sure that all previous chunks are at the very least in-progress, which puts an unexpected constraint on the download algorithm. It's a constraint that we already fulfill, but one that we'll at least need to be explicit about in the future.

But, there's also some optimization based stuff that we might lose. One thing we are going to do to speed up downloads is filter hosts into fast and slow groups. If you are doing an in-order download, the slow group could bottleneck all the fast groups, because the WriteAt calls will not return until the slow groups complete. This could significantly impact overall performance, which means if you are doing sequential downloads, you may want to disable the slowest groups and just use the fast hosts (depending on how much RAM the user is allowing you to use. If you've got more than 2GB to work with it's probably not going to matter).

RNabel: @DavidVorick The DownloadFileWriter's WriteAt() functions supports out-of-order writes and is a direct copy of the file-backed logic currently used [here](https://github.com/NebulousLabs/Sia/blob/9abac979dddb66f1de924eadcd51709fc23b975f/modules/renter/download.go#L242).

The file-backed WriteAt() doesn't block. If the current file length is less than `offset` the underlying file is padded with `0` bytes.

modules/renter.go
```diff
@@ -50,12 +54,96 @@ type Allowance struct {
...
+	Destination DownloadWriter `json:"destination"`
+	Filesize    uint64         `json:"filesize"`
+	Received    uint64         `json:"received"`
+	StartTime   time.Time      `json:"starttime"`
+	Error       string         `json:"error"`
+}
+
+// DownloadWriter provides an interface which all output writers have to implement.
+type DownloadWriter interface {
+	WriteAt(b []byte, off int64) (int, error)
```
DavidVorick: John and I talked a bit and we feel that `DownloadWriter` is the wrong abstraction for what we want to achieve. The limitations of `WriteAt` don't fit the constraints that we have to work with when doing massively parallel downloads under memory constrained situations, especially if we are trying to eek out every ounce of performance and cost-efficiency.

What I think instead should be done is have the downloader only operate one chunk at a time, and then when the downloader is finished fetching that chunk, it'll return it.

`DownloadChunk(index, offset uint64) ([]byte, error)`

Then the caller is responsible for managing what happens with the data afterwords. And at some point I can see us extending the DownloadChunk function to have priority as well or something (streaming would want a more consistent latency at the very least, and maybe even to get download-boosted on every chunk vs. just the final ones).

I think the `WriteAt` function is just going to cause lots of headaches as we try to optimize around different situations. Seems much more flexible to grab one chunk at a time and then let the caller handle optimizing, ordering, memory constraints, etc.

modules/renter.go
```diff
@@ -50,12 +54,92 @@ type Allowance struct {
...
 // download.
 type DownloadInfo struct {
-	SiaPath     string    `json:"siapath"`
-	Destination string    `json:"destination"`
-	Filesize    uint64    `json:"filesize"`
-	Received    uint64    `json:"received"`
-	StartTime   time.Time `json:"starttime"`
-	Error       string    `json:"error"`
+	SiaPath     string         `json:"siapath"`
+	Destination DownloadWriter `json:"destination"`
```
DavidVorick: Unless doing a type switch, caller is not going to be able to see the fields like Location. Given that this struct is only ever used for returning information about a download to the caller, seems strange to me to have this be an interface. Especially because this struct is going to be fed over the API.

modules/renter/download.go
```diff
@@ -16,10 +15,10 @@ import (
 	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/types"
+	"math"
```
DavidVorick: incorrectly clustered import.

modules/renter/download.go
```diff
@@ -63,15 +62,17 @@ type (
...
+		length             uint64
 
 		// Timestamp information.
 		completeTime time.Time
 		startTime    time.Time
 
 		// Static information about the file - can be read without a lock.
 		chunkSize         uint64
-		destination       string
+		destination       modules.DownloadWriter
```
DavidVorick: I think it actually makes sense to transition this here to a per-chunk thing. Basically, a `download` as managed by the core download loop should only be doing one chunk at a time. Should be able to do offset at length within the chunk (so you can download partial chunks), but because we've now got 3 completely different download scenarios (repair, file download, stream download) and they are all going to want to organize their chunks and priorities a bit differently.

modules/renter/download.go
```diff
@@ -63,20 +62,22 @@ type (
...
 		erasureCode       modules.ErasureCoder
 		fileSize          uint64
 		masterKey         crypto.TwofishKey
 		numChunks         uint64
-		pieceSet          []map[types.FileContractID]pieceData
+		pieceSet          map[int]map[types.FileContractID]pieceData
```
DavidVorick: what's the goal in creating this nested map? Generally speaking that's got a bad code smell, very easy to forget to initialize maps and stuff, or to miss indexes, etc.

modules/renter/download.go
```diff
@@ -116,35 +117,59 @@ type (
...
+	d.reportedPieceSize = dlSize / (numChunks * uint64(d.erasureCode.MinPieces()))
+	d.atomicDataReceived = dlSize - (d.reportedPieceSize * numChunks * uint64(d.erasureCode.MinPieces()))
 
 	// Assemble the piece set for the download.
-	d.pieceSet = make([]map[types.FileContractID]pieceData, f.numChunks())
-	for i := range d.pieceSet {
+	d.pieceSet = make(map[int]map[types.FileContractID]pieceData)
+	for i := range d.finishedChunks {
 		d.pieceSet[i] = make(map[types.FileContractID]pieceData)
 	}
```
DavidVorick: still not sure I see the value of doing it this way over using a full array.

modules/renter/download.go
```diff
@@ -230,35 +258,43 @@ func (cd *chunkDownload) recoverChunk() error {
 		return build.ExtendErr("unable to recover chunk", err)
 	}
 
-	// Open a file handle for the download.
-	fileDest, err := os.OpenFile(cd.download.destination, os.O_CREATE|os.O_WRONLY, defaultFilePerm)
-	if err != nil {
-		return build.ExtendErr("unable to open download destination", err)
+	var result = recoverWriter.Bytes()
+
+	// Calculate the offset. If the offset is within the chunk, the requested offset is passed, otherwise the offset of the chunk within the overall file is passed.
```
DavidVorick: comment needs to be wrapped.

modules/renter/downloadqueue.go
```diff
@@ -4,16 +4,16 @@ import (
...
+	"fmt"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/types"
 )
 
-// Download downloads a file, identified by its path, to the destination
-// specified.
-func (r *Renter) Download(path, destination string) error {
+// DownloadSection performs a file download according to the download parameters passed.
+func (r *Renter) DownloadSection(p *modules.RenterDownloadParameters) error {
```
DavidVorick: Fine to continue calling it `Download`

modules/renter/files.go
```diff
@@ -214,6 +214,20 @@ func (r *Renter) FileList() []modules.FileInfo {
...
+// GetFile returns file information about the file with the provided name and returns false if the file does not exist.
+func (r *Renter) GetFile(name string) (*modules.FileInfo, bool) {
+	// Lookup the file associated with the nickname.
+	files := r.FileList()
+
+	for _, f := range files {
+		if f.SiaPath == name {
+			return &f, true
+		}
+	}
```
DavidVorick: There's got to be a cleaner way to do this.

api/renter_test.go
RNabel: Done.

modules/renter/download.go
RNabel: The array type works as all chunks were downloaded (e.g. numbered 0 through N). Since `offset+length` downloads require a subset of chunks to be downloaded, a map allows tracking only the chunks that are actually required.

modules/renter/files.go
RNabel: Function removed in its entirety.

modules/renter/downloadqueue.go
RNabel: Name updated.

modules/renter/download.go
RNabel: Fixed.

modules/renter/download.go
RNabel: Fixed.

DavidVorick:
I think all that you have to do is call `Read` on the `http.ResponseWriter` as the chunks come in. When the read call returns, you know that it's okay to free the memory and grab the next chunk. You should have multiple chunks buffered (or at least one extra one) for larger downloads, so there shouldn't be any gap time.

DavidVorick:
This PR has been merged into #1778

PR #1674, opened 2017-04-08
add score information for active hosts in ./siac hostdb -v

DavidVorick:
This PR is a PR that is helping to lead up to an overhaul of how the allowance works. I'm programming the allowance to switch away from hosts if they become uncompetitive, which requires understanding what counts as competitive.

This PR makes it a lot easier to understand how competitive the hosts are relative to eachother, given the current scoring methods.

### DIFF
--- a/siac/hostdbcmd.go
+++ b/siac/hostdbcmd.go
@@ -191,7 +191,7 @@ func hostdbcmd() {
 		// Grab the host at the 1/5th point and use it as the reference. (it's
 		// like using the median, except at the 1/5th point instead of the 1/2
 		// point.)
-		referenceScore := float64(1)
+		referenceScore := big.NewFloat(1)
 		if len(activeHosts) > 0 {
 			referenceIndex := len(activeHosts) / 5
 			hostInfo := new(api.HostdbHostsGET)
@@ -199,7 +199,7 @@ func hostdbcmd() {
 			if err != nil {
 				die("Could not fetch provided host:", err)
 			}
-			referenceScore, _ = big.NewFloat(1).SetInt(hostInfo.ScoreBreakdown.Score.Big()).Float64()
+			referenceScore = big.NewFloat(1).SetInt(hostInfo.ScoreBreakdown.Score.Big())
 		}
 
 		fmt.Println()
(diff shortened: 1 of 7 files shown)


siac/hostdbcmd.go
```diff
@@ -227,9 +229,18 @@ func hostdbcmd() {
...
 			}
 
+			// Grab the score information for the active hosts.
+			hostInfo := new(api.HostdbHostsGET)
+			err := getAPI("/hostdb/hosts/"+host.PublicKeyString, hostInfo)
+			if err != nil {
+				die("Could not fetch provided host:", err)
+			}
+			score, _ := big.NewFloat(1).SetInt(hostInfo.ScoreBreakdown.Score.Big()).Float64()
+			score /= 1e70
```
lukechampine: `1e70` seems a bit magic

DavidVorick: It's very magic. And the worst part is, I don't know of a good way to de-magic it, except maybe to squish it with a logarithm. But then users might get confused about just how terrible a low score is.

DavidVorick:
Ok, updated this PR to use normalization instead of blindly dividing by 1e70. The normalization uses a type of median, which means that as the hostdb changes the displayed score for a host may change even though it's absolute score has not changed. This is fine though, because the scores are purely relative anyway.

API documentation has been updated appropriately as well.

siac/hostdbcmd.go
```diff
@@ -183,21 +184,35 @@ func hostdbcmd() {
...
+		// point.)
+		referenceScore := float64(1)
+		if len(activeHosts) > 0 {
+			referenceIndex := len(activeHosts) / 5
+			hostInfo := new(api.HostdbHostsGET)
+			err := getAPI("/hostdb/hosts/"+activeHosts[referenceIndex].PublicKeyString, hostInfo)
+			if err != nil {
+				die("Could not fetch provided host:", err)
+			}
+			referenceScore, _ = big.NewFloat(1).SetInt(hostInfo.ScoreBreakdown.Score.Big()).Float64()
```
lukechampine: nbd, but `new(big.Float)` is our usual style

also, seems like you are converting a `big.Int` to a `big.Float` and then a `float64`, then converting other `big.Int`s to `big.Floats` and then `float64`s and then dividing them by the first `float64`. Couldn't you do the division on `big.Float`s directly? `big.Float` supports the standard printf directives, so you don't even have to change your format.

DavidVorick: oh, I guess that works. Didn't realize that there was no need to convert to a float64.

DavidVorick: hmm I don't see a way to div floats

PR #1675, opened 2017-04-10
Don't kick outbound or local peers on accept

VoidingWarranties:
Hello old friends.

Fixes a bug where outbound and local peers could be kicked, and where no peers would be kicked if the peer being accepted was local.

### DIFF
--- a/modules/gateway/peers.go
+++ b/modules/gateway/peers.go
@@ -226,9 +226,9 @@ func (g *Gateway) acceptPeer(p *peer) {
 	// Select a peer to kick. Outbound peers and local peers are not
 	// available to be kicked.
 	var addrs []modules.NetAddress
-	for addr := range g.peers {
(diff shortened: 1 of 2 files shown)


modules/gateway/peers.go
```diff
@@ -226,9 +226,9 @@ func (g *Gateway) acceptPeer(p *peer) {
 	// Select a peer to kick. Outbound peers and local peers are not
 	// available to be kicked.
 	var addrs []modules.NetAddress
-	for addr := range g.peers {
+	for addr, peer := range g.peers {
 		// Do not kick outbound peers or local peers.
-		if !p.Inbound || p.Local {
+		if !peer.Inbound || peer.Local {
```
lukechampine: whoa, derp

lukechampine:
Thanks Jordan!

PR #1677, opened 2017-04-10
siac: add recursive directory upload

lukechampine:
This is untested. I recommend testing on both Windows and Linux to confirm that the path separator doesn't cause problems.

### DIFF
--- a/siac/rentercmd.go
+++ b/siac/rentercmd.go
@@ -491,6 +491,7 @@ func renterfilesuploadcmd(source, path string) {
 		for _, file := range files {
 			fpath, _ := filepath.Rel(source, file)
 			fpath = filepath.Join(path, fpath)
+			fpath = filepath.ToSlash(fpath)
 			err = post("/renter/upload/"+fpath, "source="+abs(file))
 			if err != nil {
 				die("Could not upload file:", err)
@@ -508,7 +509,7 @@ func renterfilesuploadcmd(source, path string) {
 }
 
 // renterpricescmd is the handler for the command `siac renter prices`, which
-// displays the pirces of various storage operations.
+// displays the prices of various storage operations.
 func renterpricescmd() {
 	var rpg api.RenterPricesGET
 	err := getAPI("/renter/prices", &rpg)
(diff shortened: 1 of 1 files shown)


DavidVorick:
no appveyor?

DavidVorick:
build failed btw

lukechampine:
I abused appveyor a little testing this. Caught at least one bug. Only thing I'm unsure about is converting Windows paths to siapaths. For example, if you upload directory `foo` on Linux, you get:
```
POST /renter/upload/foo/bar
POST /renter/upload/foo/baz
POST /renter/upload/foo/quux
```
but on Windows you get:
```
POST /renter/upload/foo\bar
POST /renter/upload/foo\baz
POST /renter/upload/foo\quux
```
These don't result in an error, but shouldn't they? The existing `upload` command doesn't seem to convert the path at all, so this behavior has existed for a while on Windows with apparently no complaints. It's easy enough to do the conversion, I'm just curious why it hasn't been an issue. @johnathanhowell, can you give any input on how Sia-UI handles this? From a cursory look, I can see you use `posix` filepaths in a couple places.

johnathanhowell:
Sia-UI uses only forward slashes in the calls to `/renter/upload`, parsing the required path by taking the basename of the file to upload and joining it with the current working siapath in the file browser.

DavidVorick:
I think we wanted the api to only accept normal/standard slashes. I might remember that we have safeties on the Sia backend as well, but I think general/best/suggested practice is for the frontend to convert to all forward slashes for siapaths.

But for the sourcefile I think it makes sense to use the native slashes.

siac/rentercmd.go
```diff
@@ -491,6 +491,7 @@ func renterfilesuploadcmd(source, path string) {
 		for _, file := range files {
 			fpath, _ := filepath.Rel(source, file)
 			fpath = filepath.Join(path, fpath)
+			fpath = filepath.ToSlash(fpath)
```
lukechampine: looking at the source, this seems to do a naive `string.Replace`. I feel like that wouldn't work on escaped path separators, but apparently it's good enough for the stdlib.

PR #1678, opened 2017-04-10
rescan after loading seed

lukechampine:
It's a bummer that we need to rescan in so many different cases. I think a better walletdb design could remove the need to rescan at the cost of storing more data. For example, if we stored a mapping from every outputID to the blockID where it was spent, we could simply rescan this set (requesting blocks from consensus as needed) to find all the outputs and transactions associated with a given seed.

### DIFF
--- a/api/wallet_test.go
+++ b/api/wallet_test.go
@@ -12,6 +12,7 @@ import (
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/modules/consensus"
 	"github.com/NebulousLabs/Sia/modules/gateway"
+	"github.com/NebulousLabs/Sia/modules/miner"
 	"github.com/NebulousLabs/Sia/modules/transactionpool"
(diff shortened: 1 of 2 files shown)


DavidVorick:
Need an API test demonstrating that this functionality is working as designed.

api/wallet_test.go
```diff
@@ -456,6 +457,68 @@ func TestIntegrationWalletSweepSeedPOST(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	for i := types.BlockHeight(0); i <= types.MaturityDelay; i++ {
+		m.AddBlock()
+	}
+
+	// Record starting balances.
+	oldBal, _, _ := st.wallet.ConfirmedBalance()
+	w2bal, _, _ := w2.ConfirmedBalance()
```
DavidVorick: should drop a sanity check here that `w2bal` != 0. Otherwise whole test looks good.

PR #1686, opened 2017-04-11
find all keys when loading a seed

lukechampine:


modules/wallet/seed.go
```diff
@@ -229,16 +187,59 @@ func (w *Wallet) LoadSeed(masterKey crypto.TwofishKey, seed modules.Seed) error
...
+		}
+	}
+	w.mu.RUnlock()
+
+	// scan blockchain to determine how many keys to generate for the seed
+	s := newSeedScanner(seed)
+	if err := s.scan(w.cs); err != nil {
+		return err
+	}
+	// Add 10% as a buffer because there is little reason not to.
```
DavidVorick: should also pitch the advantage lol - we don't want to risk that the user have addresses in the wild which have not been used yet, but might be used soon (e.g. a donation address or something).

modules/wallet/seed.go
```diff
@@ -229,16 +187,60 @@ func (w *Wallet) LoadSeed(masterKey crypto.TwofishKey, seed modules.Seed) error
...
+
+	// scan blockchain to determine how many keys to generate for the seed
+	s := newSeedScanner(seed)
+	if err := s.scan(w.cs); err != nil {
+		return err
+	}
+	// Add 10% as a buffer because the seed may have addresses in the wild
+	// that have not appeared in the blockchain yet.
+	seedProgress := s.largestIndexSeen + 1
+	seedProgress += seedProgress / 10
```
DavidVorick: can you add a log here with both the largest seen and the final value for seedProgress?

I would like to check in my wallet what these values ended up being.

DavidVorick:
doesn't compile

DavidVorick:
Should be good to go, logs seem to be working as desired.

modules/wallet/scan.go
```diff
@@ -144,8 +155,9 @@ func (s *seedScanner) scan(cs modules.ConsensusSet) error {
 		if err := cs.ConsensusSetSubscribe(s, modules.ConsensusChangeBeginning); err != nil {
 			return err
 		}
+		cs.Unsubscribe(s)
 		if s.largestIndexSeen < s.numKeys()/2 {
-			cs.Unsubscribe(s)
+			s.log.Println("Seed scanning completed, highest key index found at", s.largestIndexSeen)
```
lukechampine: We already print a log message like this in `InitFromSeed` and `LoadSeed`. Also, sometimes you don't care about the largest index, as in `SweepSeed`.

DavidVorick:
fixed

PR #1687, opened 2017-04-11
add `ConnectabilityState` and `WorkingState` fields to the host

johnathanhowell:
This PR moves the functionality described in https://github.com/NebulousLabs/Sia-UI/pull/530 to the back-end, exposing two new API fields, `ConnectabilityState` and `WorkingState`. The commits are pretty easy to follow and should be well commented.

### DIFF
--- a/doc/api/Host.md
+++ b/doc/api/Host.md
@@ -312,7 +312,18 @@ fetches status information about the host.
     // The number of times that a renter has attempted to use an
     // unrecognized call. Larger numbers typically indicate buggy software.
     "unrecognizedcalls": 6
-  }
+  },
+
+	// Information about the health of the host.
+
+	// connectabilitystate is one of "checking", "connectable",
+	// or "not connectable", and indicates if the host can connect to
+	// itself on its configured NetAddress.
+	"connectabilitystate": "checking",
+
+	// workingstate is one of "checking", "working", or "not working"
+	// and indicates if the host is being actively used by renters.
+	"workingstate": "checking"
 }
 ```
 
--- a/doc/api/Host.md
+++ b/doc/api/Host.md
@@ -314,16 +314,16 @@ fetches status information about the host.
     "unrecognizedcalls": 6
   },
 
-	// Information about the health of the host.
+  // Information about the health of the host.
 
-	// connectabilitystate is one of "checking", "connectable",
-	// or "not connectable", and indicates if the host can connect to
-	// itself on its configured NetAddress.
-	"connectabilitystate": "checking",
+  // connectabilitystate is one of "checking", "connectable",
+  // or "not connectable", and indicates if the host can connect to
+  // itself on its configured NetAddress.
+  "connectabilitystate": "checking",
 
-	// workingstate is one of "checking", "working", or "not working"
-	// and indicates if the host is being actively used by renters.
-	"workingstate": "checking"
+  // workingstate is one of "checking", "working", or "not working"
+  // and indicates if the host is being actively used by renters.
+  "workingstate": "checking"
 }
 ```
 
--- a/doc/api/Host.md
+++ b/doc/api/Host.md
@@ -316,14 +316,14 @@ fetches status information about the host.
 
   // Information about the health of the host.
 
-  // connectabilitystate is one of "checking", "connectable",
+  // connectabilitystatus is one of "checking", "connectable",
   // or "not connectable", and indicates if the host can connect to
   // itself on its configured NetAddress.
-  "connectabilitystate": "checking",
+  "connectabilitystatus": "checking",
 
-  // workingstate is one of "checking", "working", or "not working"
+  // workingstatus is one of "checking", "working", or "not working"
   // and indicates if the host is being actively used by renters.
-  "workingstate": "checking"
+  "workingstatus": "checking"
 }
 ```
 
--- a/modules/host.go
+++ b/modules/host.go
@@ -15,6 +15,35 @@ var (
 
 	// BlockBytesPerMonthTerabyte is the conversion rate between block-bytes and month-TB.
 	BlockBytesPerMonthTerabyte = BytesPerTerabyte.Mul64(4320)
+
+	// HostWorkingStatusChecking is returned from WorkingStatus() if the host is
+	// still determining if it is working, that is, if settings calls are
+	// incrementing.
+	HostWorkingStatusChecking = HostWorkingStatus("checking")
+
+	// HostWorkingStatusNotWorking is returned from WorkingStatus() if the host
+	// has not received any settings calls over the duration of
+	// workingStatusFrequency.
+	HostWorkingStatusNotWorking = HostWorkingStatus("not working")
+
+	// HostWorkingStatusWorking is returned from WorkingStatus() if the host has
+	// received more than workingThreshold settings calls over the duration of
+	// workingStatusFrequency.
+	HostWorkingStatusWorking = HostWorkingStatus("working")
+
+	// HostConnectabilityStatusChecking is returned from ConnectabilityStatus()
+	// if the host is still determining if it is connectable.
+	HostConnectabilityStatusChecking = HostConnectabilityStatus("checking")
+
+	// HostConnectabilityStatusConnectable is returned from
+	// ConnectabilityStatus() if the host is connectable at its configured
+	// netaddress.
+	HostConnectabilityStatusConnectable = HostConnectabilityStatus("connectable")
+
+	// HostConnectabilityStatusNotConnectable is returned from
+	// ConnectabilityStatus() if the host is not connectable at its configured
+	// netaddress.
+	HostConnectabilityStatusNotConnectable = HostConnectabilityStatus("not connectable")
 )
 
 type (
--- a/modules/host.go
+++ b/modules/host.go
@@ -31,7 +31,7 @@ type (
 		ContractCompensation          types.Currency `json:"contractcompensation"`
 		PotentialContractCompensation types.Currency `json:"potentialcontractcompensation"`
 
-		// Metrics related to storage proofs, collateral, and submitting //
+		// Metrics related to storage proofs, collateral, and submitting
 		// transactions to the blockchain.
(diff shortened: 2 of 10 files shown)


DavidVorick:
Needs a test in the api package as well.

johnathanhowell:
ACK, I added two, one for each field

modules/host/network.go
```diff
@@ -49,6 +49,75 @@ func (h *Host) threadedUpdateHostname(closeChan chan struct{}) {
...
+
+		h.mu.Lock()
+		if err == nil {
+			h.connectabilityState = ConnectabilityStateConnectable
+		} else {
+			h.connectabilityState = ConnectabilityStateNotConnectable
+		}
+		h.mu.Unlock()
+
+		conn.Close()
```
DavidVorick: panic is happening here in the appveyor tests.

doc/api/Host.md
```diff
@@ -312,7 +312,18 @@ fetches status information about the host.
     // The number of times that a renter has attempted to use an
     // unrecognized call. Larger numbers typically indicate buggy software.
     "unrecognizedcalls": 6
-  }
+  },
```
DavidVorick: is this brace tabbed out correctly? Should it even be here?

modules/host.go
```diff
@@ -92,6 +92,12 @@ type (
 		ObligationStatus    uint64 `json:"obligationstatus"`
 	}
 
+	// HostWorkingState reports the working state of a host.
+	HostWorkingState string
+
+	// HostConnectabilityState reports the connectability state of a host.
+	HostConnectabilityState string
```
DavidVorick: These need to have well defined values that they can be. String is not enough, need to enumerate all possible results.

DavidVorick: *in the docstring, like you did for the api

DavidVorick: Hmm. In other modules I think we actually defined consts for each possible result. e.g. `statusChecking HostConnectabilityStatus = "checking"`

And, I'm thinking 'Status' is a better suffix than 'State', @lukechampine thoughts on that?

modules/host/consts.go
```diff
@@ -31,6 +31,33 @@ const (
...
+	// host is still determining if it is connectable.
+	ConnectabilityStateChecking = "checking"
+
+	// ConnectabilityStateConnectable is returned from ConnectabilityState() if
+	// the host is connectable at its configured netaddress.
+	ConnectabilityStateConnectable = "connectable"
+
+	// ConnectabilityStateNotConnectable is returned from ConnectabilityState()
+	// if the host is not connectable at its configured netaddress.
+	ConnectabilityStateNotConnectable = "not connectable"
```
DavidVorick: oh derp, I'm dumb. Lol

modules/host.go
DavidVorick: nvm I see that you did define constants for each potential value. Still they need to be enumerated in this docstring though.

modules/host/consts.go
```diff
@@ -104,6 +131,26 @@ var (
...
+
+	// workingStateThreshold defines how many settings calls must occur over the
+	// workingStateFrequency for the host to be considered working.
+	workingStateThreshold = uint64(3)
+
+	// connectablityCheckFrequency defines how often the host's connectability
+	// check is run.
+	connectabilityCheckFrequency = build.Select(build.Var{
+		Standard: time.Minute * 3,
+		Dev:      time.Minute,
```
DavidVorick: can probably put the values here higher. Maybe 10 minutes for Standard and 3 minutes for Dev?

DavidVorick: or maybe even use the workingStateFrequency values, those seem good too.

modules/host/network.go
```diff
@@ -49,6 +49,75 @@ func (h *Host) threadedUpdateHostname(closeChan chan struct{}) {
...
+	defer close(closeChan)
+
+	for {
+		prevSettingsCalls := atomic.LoadUint64(&h.atomicSettingsCalls)
+
+		select {
+		case <-h.tg.StopChan():
+			return
+		case <-time.After(workingStateFrequency):
+		}
```
DavidVorick: :+1:

modules/host/network.go
```diff
@@ -49,6 +49,75 @@ func (h *Host) threadedUpdateHostname(closeChan chan struct{}) {
...
+		autoAddr := h.autoAddress
+		userAddr := h.settings.NetAddress
+		h.mu.RUnlock()
+
+		activeAddr := autoAddr
+		if userAddr != "" {
+			activeAddr = userAddr
+		}
+
+		conn, err := net.Dial("tcp", string(activeAddr))
```
DavidVorick: Need to use the quick-shutdown pattern on dial calls.

https://github.com/NebulousLabs/Sia/blob/master/modules/host/network.go#L141

^ example. It's pretty gross, but we've not found a better way to do it that's equally complete. Anything else we've figured out so far has been just as gross.

modules/host/network.go
DavidVorick: if you use the quick shutdown pattern linked above, it should be okay.

modules/host/network.go
```diff
@@ -49,6 +49,75 @@ func (h *Host) threadedUpdateHostname(closeChan chan struct{}) {
...
+	for {
+		prevSettingsCalls := atomic.LoadUint64(&h.atomicSettingsCalls)
+
+		select {
+		case <-h.tg.StopChan():
+			return
+		case <-time.After(workingStateFrequency):
+		}
+
+		settingsCalls := atomic.LoadUint64(&h.atomicSettingsCalls)
```
DavidVorick: can probably drop the newlines buffering the soft-sleep pattern. The whole idea of get calls, sleep, get calls again is a single thought.

johnathanhowell:
Latest commits address review comments, I also change the field names from `State` to `Status`, because `Status` is more prevalent in the codebase currently.

modules/host/network.go
johnathanhowell: we actually don't need to do anything with this conn, so I think it's sufficient to simply close it immediately here.

DavidVorick: The problem is that `Dial` can take a while. Also I linked to the wrong example.

DavidVorick: https://github.com/NebulousLabs/Sia/blob/master/modules/gateway/conn.go#L36

johnathanhowell: Ah, that makes more sense, thanks

modules/host/network.go
```diff
@@ -49,6 +49,78 @@ func (h *Host) threadedUpdateHostname(closeChan chan struct{}) {
...
+		dialer := &net.Dialer{
+			Cancel:  h.tg.StopChan(),
+			Timeout: connectabilityCheckTimeout,
+		}
+		conn, err := dialer.Dial("tcp", string(activeAddr))
+		if err != nil {
+			h.mu.Lock()
+			h.connectabilityStatus = ConnectabilityStatusNotConnectable
+			h.mu.Unlock()
+			continue
```
DavidVorick: continue here is after the `conn.Close()`, need to make sure you call close before you continue

johnathanhowell: but this is in the `err != nil` branch, `conn` is nil

DavidVorick: oh, whoops. You are right.

DavidVorick:
LGTM

modules/host/consts.go
```diff
@@ -104,6 +131,38 @@ var (
...
+		Testing:  time.Second * 30,
+	}).(time.Duration)
+
+	// workingStatusThreshold defines how many settings calls must occur over the
+	// workingStatusFrequency for the host to be considered working.
+	workingStatusThreshold = build.Select(build.Var{
+		Standard: uint64(3),
+		Dev:      uint64(1),
+		Testing:  uint64(1),
+	}).(uint64)
```
lukechampine: why does this ever need to be > 1?

modules/host/consts.go
```diff
@@ -31,6 +31,33 @@ const (
...
+	// host is still determining if it is connectable.
+	ConnectabilityStatusChecking = "checking"
+
+	// ConnectabilityStatusConnectable is returned from ConnectabilityStatus() if
+	// the host is connectable at its configured netaddress.
+	ConnectabilityStatusConnectable = "connectable"
+
+	// ConnectabilityStatusNotConnectable is returned from ConnectabilityStatus()
+	// if the host is not connectable at its configured netaddress.
+	ConnectabilityStatusNotConnectable = "not connectable"
```
lukechampine: shouldn't these be defined in the `modules` package, not `host`?

modules/host/network.go
lukechampine: maybe simpler as:
```go
var status modules.HostConnectabilityStatus
if err != nil {
    status = modules.HostConnectabilityStatusNotConnectable
} else {
    conn.Close()
    status = modules.HostConnectabilityStatusNotConnectable
}
h.mu.Lock()
h.connectabilityStatus = status
m.mu.Unlock()
```

modules/host/consts.go
DavidVorick: Guess so, given that the API relies on them to match a certain specification.

modules/host/consts.go
DavidVorick: Sometimes the number will increment because you are pinging yourself. It's not supposed to happen, but legacy users that upgrade to this functionality may have themselves in the hostdb via a local route.

modules/host/consts.go
johnathanhowell: `modules.ConnectabilityStatusNotConnectable` isn't that clear, would it have to be `modules.HostConnectabilityStatusNotConnectable`?

modules/host/network.go
johnathanhowell: definitely more readable that way

modules/host/consts.go
DavidVorick: yes, it would need to be. All structs in the modules have names prefixed by the most relevant module. e.g. `modules.RenterContract`

PR #1689, opened 2017-04-12
API return score breakdown

S-anasol:


DavidVorick:
I almost did this change myself, but there's a few things holding me back.

1. Getting the score breakdown of a host requires some more computation, and there is already an api call to get the score breakdown on a single host.
2. It makes the call to get an individual host redundant, as it now has a HostdbScoreBreakdown in two places.

I'm undecided if this is a change we want.

@lukechampine ?

DavidVorick:
@S-anasol can you describe your use case for it, and whether it would be prohibitive to make the call once per host you want the score breakdown of (which is already supported)

lukechampine:
I think it's fine to have one call that returns general information about every host and a separate call that returns specific information about one host. I suppose it's important to consider the theoretical size of the hostdb. Will it ever be large enough that we will need to paginate the response? If so, we should lean towards returning as little info about each host as possible.

DavidVorick:
Yeah. And I also think that the ScoreBreakdown is likely to increase substantially in computational complexity. Today it's maybe a few microseconds per host, but as we get fancy we could see the calculation rising into the multi-millsecond range, which would be a problem if the database grows to 5k+ hosts.

S-anasol:
using this to avoid 100+ additional requests for each host.
https://github.com/S-anasol/siahub/blob/master/app/Console/Commands/Hosts.php#L46-L60

DavidVorick:
You will probably have to get around this by caching the information per-host and then updating the score once every 10 minutes or something. Host scores should not be changing that fast anyway. In the long term, providing a new score breakdown for each host every api call is going to slow down the daemon too much.

S-anasol:
image

PR #1690, opened 2017-04-13
add wallet reinitialization (--force)

johnathanhowell:
This PR adds a new option to the `/wallet/init` endpoint, `force`, which when set to `true` will reinitialize the wallet, even if it has already been encrypted. This is useful for users who wish to destroy their wallet and start over, without meddling with persist data.

### DIFF
--- a/api/wallet.go
+++ b/api/wallet.go
@@ -190,7 +190,14 @@ func (api *API) walletInitHandler(w http.ResponseWriter, req *http.Request, _ ht
 	if req.FormValue("encryptionpassword") != "" {
 		encryptionKey = crypto.TwofishKey(crypto.HashObject(req.FormValue("encryptionpassword")))
 	}
-	seed, err := api.wallet.Encrypt(encryptionKey)
+
+	var seed modules.Seed
+	var err error
+	if req.FormValue("force") == "true" {
+		seed, err = api.wallet.Reencrypt(encryptionKey)
+	} else {
+		seed, err = api.wallet.Encrypt(encryptionKey)
+	}
 	if err != nil {
 		WriteError(w, Error{"error when calling /wallet/init: " + err.Error()}, http.StatusBadRequest)
 		return
--- a/api/wallet.go
+++ b/api/wallet.go
@@ -199,6 +199,10 @@ func (api *API) walletInitHandler(w http.ResponseWriter, req *http.Request, _ ht
 		}
 	}
 	seed, err := api.wallet.Encrypt(encryptionKey)
+	if err != nil {
+		WriteError(w, Error{"error when calling /wallet/init: " + err.Error()}, http.StatusBadRequest)
+		return
+	}
 
 	dictID := mnemonics.DictionaryID(req.FormValue("dictionary"))
 	if dictID == "" {
--- a/api/wallet.go
+++ b/api/wallet.go
@@ -191,17 +191,14 @@ func (api *API) walletInitHandler(w http.ResponseWriter, req *http.Request, _ ht
 		encryptionKey = crypto.TwofishKey(crypto.HashObject(req.FormValue("encryptionpassword")))
 	}
 
-	var seed modules.Seed
-	var err error
 	if req.FormValue("force") == "true" {
-		seed, err = api.wallet.Reencrypt(encryptionKey)
-	} else {
-		seed, err = api.wallet.Encrypt(encryptionKey)
+		err := api.wallet.Reset()
+		if err != nil {
+			WriteError(w, Error{"error when calling /wallet/init: " + err.Error()}, http.StatusBadRequest)
+			return
+		}
 	}
-	if err != nil {
-		WriteError(w, Error{"error when calling /wallet/init: " + err.Error()}, http.StatusBadRequest)
-		return
-	}
+	seed, err := api.wallet.Encrypt(encryptionKey)
 
 	dictID := mnemonics.DictionaryID(req.FormValue("dictionary"))
 	if dictID == "" {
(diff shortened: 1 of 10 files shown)


api/wallet.go
```diff
@@ -190,7 +190,14 @@ func (api *API) walletInitHandler(w http.ResponseWriter, req *http.Request, _ ht
 	if req.FormValue("encryptionpassword") != "" {
 		encryptionKey = crypto.TwofishKey(crypto.HashObject(req.FormValue("encryptionpassword")))
 	}
-	seed, err := api.wallet.Encrypt(encryptionKey)
+
+	var seed modules.Seed
+	var err error
+	if req.FormValue("force") == "true" {
+		seed, err = api.wallet.Reencrypt(encryptionKey)
```
lukechampine: here's an alternative API:
```go
if req.FormValue("force") == "true" {
    // revert the wallet to an uninitialized state
    api.wallet.Clear()
}
seed, err := api.wallet.Encrypt(encryptionKey)
```

lukechampine: The advantage of this design is that it can be used to add a `"force"` flag to the `InitFromSeed` endpoint as well.

modules/wallet/database.go
```diff
@@ -206,6 +207,39 @@ func dbDeleteSpentOutput(tx *bolt.Tx, id types.OutputID) error {
...
+	for _, bucket := range dbBuckets {
+		err := tx.DeleteBucket(bucket)
+		if err != nil {
+			return err
+		}
+		_, err = tx.CreateBucket(bucket)
+		if err != nil {
+			return err
+		}
+	}
```
DavidVorick: Would like to see this in the wild, to see if the database blows up too much.

DavidVorick: *meaning the db size

modules/wallet/encrypt_test.go
```diff
@@ -284,3 +284,32 @@ func TestInitFromSeed(t *testing.T) {
...
+
+	var encryptionKey crypto.TwofishKey
+	seed, err := wt.wallet.Reencrypt(encryptionKey)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	walletKey := crypto.TwofishKey(crypto.HashObject(seed))
+	postEncryptionTesting(wt.miner, wt.wallet, walletKey)
+}
```
DavidVorick: Need to try funding the wallet and sending money after re-init. Need to verify that the internal state is still collected enough to be producing real results.

DavidVorick: And really, should do a couple rounds of spending money before reinit is called as well.

DavidVorick: Ah man. And, it's a PITA, but _really_ should be spending money that was collect from expired file contract outputs as well. Which actually probably needs to be done from the API package.

I think for this release we can just drop a TODO in the API package for that much complexity. This package should probably do some sending and receiving within this package on each side of the re-encrypt (which means having multiple wallets), and then a TODO in the API package about an ecosystem test that reinits one of the wallets halfway. Or maybe that'd make more sense as an antfarm job.

johnathanhowell: `postEncryptionTesting` unlocks/locks the wallet, mines coins, and spends, though it only does one `SendSiacoins` call

api/wallet.go
lukechampine: `Reset` might be a slightly better name

modules/wallet/encrypt_test.go
DavidVorick: oh, I see. Ok, so I think for this PR it's sufficient to:

+ add a call to postEncryptionTesting after encrypting the wallet the first time
+ add a second send to postEncryptionTesting that drains the wallet entirely. You may need to change the first send if it's already draining the wallet

Ideally we'd find a way to slip in receiving coins as well, but I think that the way it's coded up, just the change address from the first send (as long as it's not the whole wallet) will be enough to get the same code coverage. But that's implementation-specific coverage instead of blackbox testing coverage.

api/wallet.go
johnathanhowell: Yeah, this design is definitely better. I think `Reset` is probably ideal

modules/wallet/database.go
lukechampine: isn't it only a problem if the existing wallet was huge? Even if bolt doesn't shrink its DB, it will still reuse existing space, right?

DavidVorick: That's why I want to see this tested out. The wallet db is 1 GB, if Bolt doesn't efficiently reuse the space then users doing multiple re-inits could end up with several extra GB on disk.

It's easy enough to test.

johnathanhowell: Ok, so doing one reinit led to a filesize increase of `wallet.db` from 815MB -> 898MB

johnathanhowell: subsequent init increased filesize to 1.2G

DavidVorick: I'd say that's an okay overhead. Just wanted to be sure.

lukechampine: interesting. I wrote a test program that did:
```go
for {
    createBucketWith1024Elements(db)
    deleteBucket(db)
    printDBSize(db)
}
```
and it never grew larger than 128KiB

johnathanhowell: Yeah, after several new reinitializations it doesn't seem to be growing at all.

api/wallet.go
```diff
@@ -190,11 +190,15 @@ func (api *API) walletInitHandler(w http.ResponseWriter, req *http.Request, _ ht
...
-		return
+
+	if req.FormValue("force") == "true" {
+		err := api.wallet.Reset()
+		if err != nil {
+			WriteError(w, Error{"error when calling /wallet/init: " + err.Error()}, http.StatusBadRequest)
+			return
+		}
 	}
+	seed, err := api.wallet.Encrypt(encryptionKey)
```
lukechampine: error not checked

PR #1691, opened 2017-04-13
Fees

DavidVorick:
Adjust the transaction fees and contract fees by a lot to make renting more economical. We'll have to ping all of the hosts and let them know that it's safe to reduce the contract fees.

### DIFF
--- a/modules/transactionpool/transactionpool.go
+++ b/modules/transactionpool/transactionpool.go
@@ -135,7 +135,7 @@ func (tp *TransactionPool) FeeEstimation() (min, max types.Currency) {
 	// a much lower value, which means hosts would be incompatible if the
 	// minimum recommended were set to 10. The value has been set to 1, which
 	// should be okay temporarily while the renters are given time to upgrade.
-	return types.SiacoinPrecision.Mul64(1).Div64(1e3), types.SiacoinPrecision.Mul64(5).Div64(1e3)
+	return types.SiacoinPrecision.Mul64(1).Div64(3).Div64(1e3), types.SiacoinPrecision.Mul64(1).Div64(1e3)
 }
 
 // TransactionList returns a list of all transactions in the transaction pool.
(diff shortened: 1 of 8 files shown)


modules/transactionpool/transactionpool.go
```diff
@@ -135,7 +135,7 @@ func (tp *TransactionPool) FeeEstimation() (min, max types.Currency) {
 	// a much lower value, which means hosts would be incompatible if the
 	// minimum recommended were set to 10. The value has been set to 1, which
 	// should be okay temporarily while the renters are given time to upgrade.
-	return types.SiacoinPrecision.Mul64(1).Div64(1e3), types.SiacoinPrecision.Mul64(5).Div64(1e3)
+	return types.SiacoinPrecision.Mul64(1).Div64(3).Div64(1e3), types.SiacoinPrecision.Mul64(1).Div64(1e3)
```
lukechampine: this is a bit weird looking but I understand why: 1 SC * (1/3) / 1000. If you reduced it to 1 SC / 3000 it wouldn't be as intuitive.

DavidVorick: yes, that's correct

lukechampine:
looks like there's a test that needs updating

DavidVorick:
should be fixed now.

lukechampine:
`TestIntegrationCheckMinerFees` is still failing

DavidVorick:
Ok, should be better now. Worth noting that existing transaction pools will still have the old limit set, but this should not be an issue as the transaction pool should rarely hit a full 500kb - that's between 200 and 800 file contracts, and it's something like 20 defrag transactions.

This change would not be safe on a busier network, but I think we can afford it. I also think it likely that the miners will be wanting to upgrade to the instant wallet unlock, and also to take advantage of the new gateway changes that improve propagation.

PR #1694, opened 2017-04-13
add some persistence benchmarks to test various systems

DavidVorick:


DavidVorick:
Secretly, I wrote these so I could see what the results were on appveyor. Since I had to write them anyway, now we have an easy thing for users to run on various filesystems and setups that will give us information about how the building blocks of the contractmanager perform.

And these are building blocks we will of course be reusing as we update the persist stuff in the other packages.

persist/disk_test.go
```diff
@@ -0,0 +1,309 @@
...
+	}
+
+	// Close the file.
+	err = f.Close()
+	if err != nil {
+		b.Fatal(err)
+	}
+
+	// Check the sequential read speed.
+	for i := 0; i < b.N; i++ {
```
lukechampine: no `b.ResetTimer` before this?

persist/disk_test.go
```diff
@@ -0,0 +1,309 @@
...
+		if err != nil {
+			b.Fatal(err)
+		}
+
+		// Close the file.
+		err = f.Close()
+		if err != nil {
+			b.Fatal(err)
+		}
+		b.StartTimer()
```
lukechampine: why all the complex initialization?
```go
data := fastrand.Bytes(1 << 28)
b.ResetTimer()
for i := 0; i < b.N; i++ {
    b.StopTimer()
    ioutil.WriteFile(filename, data, 0666)
    b.StartTimer()
```

DavidVorick: 1. copy-paste
2. that's 256MiB of memory that you just alloacted. one step at a time keeps mem requirements low, unless you're using a ramdisk for /tmp, but then that's your problem.

lukechampine: fwiw benchmarks don't run in parallel so mem usage isn't a big deal. But as long as the benchmark is accurately measuring what we want to measure, it doesn't matter.

DavidVorick: That's true. 256 MiB doesn't really mean much next to the 3.5 GiB of the api package.

DavidVorick:
This PR has been a challenge. But I think I got it now.

persist/disk_test.go
DavidVorick: going to leave it because it seems to be working, but also because having the writes be done incrementally means you can run private/personal benchmarks that push the data volumes way higher.

DavidVorick:
This is ready for merge

PR #1695, opened 2017-04-13
resolve divide by zero in siac

DavidVorick:


DavidVorick:
added another commit to eliminate a confusing 'total score' report. Normalizations can change from query to query, which means a normalized total score could jerk around a lot. Better I thought to just remove it entirely, and let the user compare the total score by looking at all the hosts. Score, all by itself, does not have much meaning.

lukechampine:
I forget, why does this code use `big.Float` instead of `big.Rat`?

DavidVorick:
I.... am not sure.

DavidVorick:
will update I guess

PR #1698, opened 2017-04-14
Add piece per chunk multiplier to chunk download.

RNabel:
1. Adds a multiplier determining the maximum number of pieces per chunk to be downloaded concurrently relative to the minimum required,
2. Adds a limiter to the number of chunks to which this new maximum is applied. The limiter controls that the speed up is only applied to the last N chunks.

This PR has resulted in single-chunk download times to go from ~19s to as low as 2.2s (multiplier val: 2.5). The current multiplier results in speedups of about 3-4x to <5s.

Disclaimer: these measurements were made on a Gbit network with direct connection to the internet backbone.

### DIFF
--- a/modules/renter/download.go
+++ b/modules/renter/download.go
@@ -7,6 +7,7 @@ package renter
 import (
 	"bytes"
 	"errors"
+	"math"
 	"os"
 	"sync"
 	"sync/atomic"
@@ -38,6 +39,29 @@ var (
 		Dev:      int(10),
 		Testing:  int(5),
 	}).(int)
+
+	// maxActivePPCMultiplier determines the maximum of pieces downloaded
+	// concurrently per chunk, relative to the minimum number required.
+	// E.g. if the minimum is 10 pieces, as in 10-of-30 erasure coding,
+	// and multiplier value of 1.2, 12 pieces will be downloaded rather
+	// than 10.
+	// This is an effective measure to reduce latency, as chunk latency
+	// is the min( set( all piece latencies ) ).
+	maxActivePPCMultiplier = build.Select(build.Var{
+		Standard: float32(1.2),
+		Dev:      float32(1.2),
+		Testing:  float32(1.2),
+	}).(float32)
+
+	// numFastChunks determines how many pieces will be downloaded with
+	// maxActivePPCMultiplier. Once a downloaded file has numFastChunks
+	// remaining all chunks will use a higher number of concurrent piece
+	// downloads.
+	numFastChunks = build.Select(build.Var{
+		Standard: int(4),
+		Dev:      int(4),
+		Testing:  int(4),
+	}).(int)
 )
 
 type (
@@ -55,6 +79,10 @@ type (
 		// have tried to fetch a piece of the chunk.
 		completedPieces map[uint64][]byte
 		workerAttempts  map[types.FileContractID]bool
+
+		assignedPieceCount int // Tracks the number of pieces assigned to a worker.
+		maxPieceCount      int // Tracks the maximum number of pieces assigned to a worked.
+		minPieceCount      int // Tracks the minimum number of pieces required to recover the chunk.
 	}
 
 	// A download is a file download that has been queued by the renter.
@@ -80,7 +108,7 @@ type (
 		reportedPieceSize uint64
 		siapath           string
 
-		// Syncrhonization tools.
+		// Synchronization tools.
 		downloadFinished chan struct{}
 		mu               sync.Mutex
 	}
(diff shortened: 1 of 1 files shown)


RNabel:
The tests are currently exploding because of [this check](https://github.com/NebulousLabs/Sia/blob/master/modules/renter/download.go#L319). Should be an easy fix.

DavidVorick:
This is an amazingly simple change for such a big payoff. We talked about adding it to v1.2.0, but the testing timeline I feel is just too short.

It will be in v1.2.1 for sure though, and we don't plan on taking long to get that out.

modules/renter/download.go
```diff
@@ -493,11 +518,26 @@ func (r *Renter) managedScheduleNewChunks(ds *downloadState) {
 			continue
 		}
 
+		// Determine whether the download should be sped up.
+		su := (nextChunk.download.numChunks - nextChunk.index) <= uint64(numFastChunks)
+		var np = nextChunk.download.erasureCode.MinPieces()
```
DavidVorick: variable names need to read like english. There are exceptions for 'obvious context', but I don't think that applies here.

For example, `cst := newConsensusSetTester` - there's an obvious mapping from cst to ConsensusSetTester. But above, `su` doesn't obviously map to anything.

Same with `np`, it's a bit closer because you have the context of 'MinPieces', but I think still difficult to read.

Also, convention would be to do `np := nextChunk...`, instead of declaring the variable with `var`.

modules/renter/download.go
```diff
@@ -493,11 +518,26 @@ func (r *Renter) managedScheduleNewChunks(ds *downloadState) {
...
+			max := nextChunk.download.erasureCode.NumPieces()
+			n := int(math.Ceil(float64(min) * float64(maxActivePPCMultiplier)))
+			if n > max { // Cap the number of pieces downloaded at the total number of pieces.
+				n = max
+			}
+
+			np = n
+		}
+
+		fmt.Printf("Pieces concurrently downloaded: %d\n", np)
```
DavidVorick: I assume this was leftover from debugging :)

modules/renter/download.go
```diff
@@ -493,11 +518,26 @@ func (r *Renter) managedScheduleNewChunks(ds *downloadState) {
...
 		// Add an incomplete chunk entry for every piece of the download.
-		for i := 0; i < nextChunk.download.erasureCode.MinPieces(); i++ {
+		for i := 0; i < np; i++ {
 			ds.incompleteChunks = append(ds.incompleteChunks, nextChunk)
 		}
-		ds.activePieces += nextChunk.download.erasureCode.MinPieces()
+		ds.activePieces += np
```
DavidVorick: I'm pretty amazed that this update was really this easy. Looks like there's at least one corner case remaining though.

DavidVorick:
The test suite we have is showing a corner case where adding the speedup appears to result in a failed download because there are not enough hosts. The download checker looks at the number of pieces it needs to fetch vs. the number of candidate hosts it has, and will fail if it does not have enough candidate hosts.

It needs to realize that it has enough hosts to download the file, just not enough to perform the speed boost.

modules/renter/download.go
RNabel: Thank you for taking the time to get me up to speed on golang best practice :)

modules/renter/download.go
RNabel: Yep, removed in the newest revision.

modules/renter/download.go
```diff
@@ -549,8 +611,7 @@ func (r *Renter) managedWaitOnDownloadWork(ds *downloadState) {
 	// If the chunk has completed, perform chunk recovery.
 	if len(cd.completedPieces) == cd.download.erasureCode.MinPieces() {
 		err := cd.recoverChunk()
-		ds.activePieces -= len(cd.completedPieces)
-		cd.completedPieces = make(map[uint64][]byte)
```
RNabel: @DavidVorick could removing this line have other negative consequences? This commit is currently relying on the reference eventually being garbage collected.

As far as I gather from the surrounding code, this line was needed because `len(cd.completedPieces)` was deducted from `ds.activePieces` elsewhere. Since all references to `len(cd.completedPieces)` was replaced with `cd.assignedPieceCount` resetting the map doesn't seem necessary any more.

modules/renter/download.go
```diff
@@ -295,13 +323,34 @@ func (r *Renter) addDownloadToChunkQueue(d *download) {
 			continue
 		}
 
+		// Calculate number of pieces to add to the download list.
+		// If the chunk's download speed should be sped up, the piece
+		// count will be raised above the minimum required.
+		// speedupchunk is true if it is the Nth last chunk where N is defined by `numFastChunks`.
```
DavidVorick: Looks like the comment formatting for all the bulk comments is off. If you use vim, you can get it correct easily by highlighting the block comment in visual mode, and then pressing `gq`. Sublime text has an equivalent, and most code-editors also have some sort of equivalent.

modules/renter/download.go
DavidVorick: To be honest, this code is not super well organized and I'm not certain whether this will cause problems or not.

The best way to tell would be an integration test that downloads:

+ a multi-chunk file with enough hosts to utilize the extra pieces
+ a multi-chunk file with seemingly enough hosts to utilize the extra pieces, except some of the hosts have gone offline so really you only have exactly enough hosts available
+ a file with more chunks than hosts that can utilize the full paralleism of the downloader

If it can do all these things, then this code is probably not worse than the existing code.

To be honest, the existing download code is pretty sloppy (my own design) and could use some design iteration. Mostly, it doesn't seem right to me that this file would need to be a whole 500 lines of code - we should be able to achieve everything it does in a lot fewer lines than that (I think, anyway).

modules/renter/download.go
RNabel: I'm using neither, and it is rendering ok for me. The comments are in-line with the rest of the block and (mostly) less than 80 characters long. Shall I change them to a block comment, or a concatenate them to a single line?

modules/renter/download.go
RNabel: Thanks for letting me know. I was thinking about rewriting this file but don't have the time right now. Maybe we could discuss possible changes to this at some point in the coming weeks. In the meantime, I'll add a couple of tests to check these cases.

lukechampine:
The renter code has changed significantly, and we'll be continuing to overhaul it over the coming weeks. I don't think it makes sense to push this PR through.

PR #1699, opened 2017-04-14
debug log block size after each block

johnathanhowell:


modules/consensus/block_validation_test.go
```diff
@@ -78,7 +78,7 @@ func TestUnitValidateBlock(t *testing.T) {
 				now: tt.now,
 			},
 		}
-		err := blockValidator.ValidateBlock(b, tt.minTimestamp, types.RootDepth, 0)
+		err := blockValidator.ValidateBlock(b, tt.minTimestamp, types.RootDepth, 0, nil)
```
DavidVorick: Bold move

modules/consensus/block_validation.go
```diff
@@ -63,7 +64,7 @@ func checkTarget(b types.Block, target types.Target) bool {
 // ValidateBlock validates a block against a minimum timestamp, a block target,
 // and a block height. Returns nil if the block is valid and an appropriate
 // error otherwise.
-func (bv stdBlockValidator) ValidateBlock(b types.Block, minTimestamp types.Timestamp, target types.Target, height types.BlockHeight) error {
+func (bv stdBlockValidator) ValidateBlock(b types.Block, minTimestamp types.Timestamp, target types.Target, height types.BlockHeight, log *persist.Logger) error {
```
lukechampine: Why not just add the log to the `stdBlockValidator` struct?

johnathanhowell: I thought it best not to add more state to the `stdBlockValidator` struct, but maybe that's misguided

lukechampine: I guess the real question is: should the logging happen inside `ValidateBlock`, or could it be done by the caller?

johnathanhowell: The reasoning for logging inside `ValidateBlock` is that we're already computing the block size here, doing it in the caller would incur another Marshal.

lukechampine: I wouldn't worry about that too much, especially if you wrap the block in `if build.DEBUG` so that you don't incur the encoding cost unless you're running debug.

On that note, I think it might be worthwhile to add a `EncodedLen` function to the `encoding` package. It could just count bytes instead of actually writing anything, so you wouldn't need to allocate memory. I'm not sure how common `len(encoding.Marshal(x))` is, though.

lukechampine: answer: somewhat common:
```
modules/transactionpool/accept.go
225:		tp.transactionListSize -= len(encoding.Marshal(conflictSet))
243:	tp.transactionListSize += len(encoding.Marshal(superset))
300:	tp.transactionListSize += len(encoding.Marshal(ts))

modules/transactionpool/standard.go
66:	if len(encoding.Marshal(t)) > modules.TransactionSizeLimit {
120:		totalSize += len(encoding.Marshal(ts[i]))

modules/host/storageobligations.go
718:		txnSize := uint64(len(encoding.MarshalAll(so.RevisionTransactionSet)) + 300)
804:		txnSize := uint64(len(encoding.Marshal(sp)) + 300)

modules/transactionpool.go
133:	size := len(encoding.Marshal(ts))

modules/miner/update.go
75:		remainingSize -= len(encoding.Marshal(unconfirmedTransactions[i]))

types/validtransaction.go
116:	if uint64(len(encoding.Marshal(t))) > BlockSizeLimit-5e3 {
```

DavidVorick: Debug builds are already too slow, at some point we do need to look into optimizing those, I would like to see all devs running debug all the time, but today that's not feasible

lukechampine: preliminary testing shows that an `EncodedLen` function would only be slightly faster:
```
BenchmarkLen-4          1000000   1465 ns/op   54.58 MB/s   472 B/op   27 allocs/op
BenchmarkLenMarshal-4 	1000000   1798 ns/op   44.48 MB/s   712 B/op   28 allocs/op
```

The quickest way to speed this up would be to add an `EncodedLen` to the `Block` type directly, which would in turn depend on an `EncodedLen` method of the `Transaction` type. These could conceivably bypass a lot of the encoding and therefore run much faster.

My guess is that `len(encoding.Marshal(x))` isn't much of a bottleneck though.

PR #1700, opened 2017-04-14
debug log transaction sets

johnathanhowell:


modules/transactionpool/accept.go
```diff
@@ -237,7 +237,18 @@ func (tp *TransactionPool) handleConflicts(ts []types.Transaction, conflicts []T
...
+	tp.transactionListSize += tsetSize
+
+	// debug logging
+	if build.DEBUG {
+		tp.log.Debugf("accepted transaction superset %v, size: %vB\n", setID, tsetSize)
+		for i, t := range superset {
+			tp.log.Debugf("superset transaction %v size: %vB\n", i, len(encoding.Marshal(t)))
+		}
+		tp.log.Debugf("tpool size is %vB after accepting transaction superset\n", tp.transactionListSize)
+	}
```
DavidVorick: This should all be one log. If you want multiple lines, use \n with \t\t\t for the next lines.

DavidVorick: Though, these logs aren't needed at all, only the other ones. Same comment to them though.

johnathanhowell: if these logs are removed, the transaction sets will never be logged if there is a conflict (acceptTransactionSet has the `if len(conflicts) > 0 return tp.handleConflicts..` branch)

DavidVorick: Oh I forgot it was structured that way. Ok, looks good then. But does still need to be a single log call, even if multiple lines.

PR #1701, opened 2017-04-14
Faster host upgrades + storage folder operations

DavidVorick:
PR is not finished - namely the test suite is broken now b/c it measures whether or not `WriteAt` gets called, but now we use `Truncate` instead, so the suite gets upset because it doesn't think we are writing out the file.

First reordered the upgrade process so that people with massive storage arrays can interrupt the growth phase without being forced to write out a full 500TB over NFS.

Second switched from doing 0-writes to doing incremental truncates. Some quick searching informed me that calling 'Truncate' is the correct way to create sparse files, and that writing zeroes is a lot slower. Updated my benchmarks PR as well, which now shows that calling Truncate is pretty fast. I guess I should also write a benchmark for tail-0 writing.

This new PR should provide great speedups for filesystems that support sparse files without impacting the performance or progress-ness of filesystems that don't support sparse files.

Made this PR quickly so that it could get reviewed faster.

### DIFF
--- a/modules/host/contractmanager/consts.go
+++ b/modules/host/contractmanager/consts.go
@@ -39,6 +39,11 @@ const (
 )
 
 const (
+	// folderAllocationStepSize is the amount of data that gets allocated at a
+	// time when writing out the sparse sector file during a storageFolderAdd or
+	// a storageFolderGrow.
+	folderAllocationStepSize = 50e6
+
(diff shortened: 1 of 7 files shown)


lukechampine:
failing with:
```
--- FAIL: TestHostDBHostsHandler (0.55s)
	hostdb_test.go:234: One value in host score breakdown
```

DavidVorick:
ah man we missed the version bump.

PR #1703, opened 2017-04-15
Tomfoolery

VoidingWarranties:
This PR is intended to look benign, but isn't :) Can you see the vulnerability this change introduces? (hint: look at why TestAcceptPeer fails) It's an admittedly contrived example, but nonetheless possible. What do you think?

### DIFF
--- a/modules/gateway/peers.go
+++ b/modules/gateway/peers.go
@@ -225,19 +225,20 @@ func (g *Gateway) acceptPeer(p *peer) {
 
 	// Select a peer to kick. Outbound peers and local peers are not
 	// available to be kicked.
+	acceptingHost := p.NetAddress.Host()
 	var addrs []modules.NetAddress
-	for addr, peer := range g.peers {
+	for _, р := range g.peers {
 		// Do not kick outbound peers or local peers.
-		if !peer.Inbound || peer.Local {
+		if !p.Inbound || p.Local {
 			continue
 		}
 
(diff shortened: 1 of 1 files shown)


lukechampine:
wow, that is sneaky. At first I thought `p` was shadowing the `p *peer` argument, but that seemed too easy. Then I flipped back because it worked when I changed the variable name. Then I got REALLY wigged out when I started getting `undefined: р in р.NetAddress errors`. I thought maybe there was some lexical scoping rule in Go that I wasn't aware of. I won't give it away so David and John have a chance to try.

DavidVorick:
Going to admit here, @lukechampine had to tell me what was wrong. I stared at it for probably half an hour without figuring it out.

1. Thank god we have a test to check correctness
2. I'm sure there are places in our code where you could get something this awful through without triggering a build failure.

Bravo, this is really clever. Thanks for sharing :)

lukechampine:
Inspired by this, I'm going to write a linter that checks for suspicious unicode. Easiest approach would be to simply ban any character larger than a certain value, but that could be overreaching for non-english programmers. Maybe that's acceptable, not sure.

VoidingWarranties:
The linter is great! Thanks for writing it. It's much better than my kludgy grep-foo. I'm glad that homographic attacks have been getting more attention recently, although this is still the first example I've seen of a homographic attack applied to source code.

PR #1709, opened 2017-04-16
Fix spelling errors in source comments and output messages

mtlynch:
Fix spelling errors via the misspell tool. Excluded fixes to "marshalled" and
"marshalling" because they seemed ~to~ too deeply ingrained in the code.

This is easily repeatable:

```
go get -u github.com/client9/misspell/cmd/misspell
$GOPATH/bin/misspell -locale US -w -i "marshalled,marshalling,Marshalling,Marshalled" .
```

### DIFF
--- a/api/renter.go
+++ b/api/renter.go
@@ -19,7 +19,7 @@ import (
 
 var (
 	// recommendedHosts is the number of hosts that the renter will form
-	// contracts with if the value is not specified explicity in the call to
+	// contracts with if the value is not specified explicitly in the call to
 	// SetSettings.
 	recommendedHosts = build.Select(build.Var{
 		Standard: uint64(50),
--- a/api/server_helpers_test.go
+++ b/api/server_helpers_test.go
@@ -400,7 +400,7 @@ func createServerTester(name string) (*serverTester, error) {
 
 // createAuthenticatedServerTester creates an authenticated server tester
 // object that is ready for testing, including money in the wallet and all
-// modules initalized.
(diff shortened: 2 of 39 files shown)


lukechampine:
Thanks! At one point I had `misspell` set up as a pre-commit hook, but I lost it after a system refresh (and of course, pre-commit hooks only work on new code, not existing code). I seem to recall `misspell` not being comprehensive enough for my liking, but it's certainly better than nothing.

lukechampine:
also, I'm not opposed to fixing `marshalling` and `marshalled` as long as they aren't part of API.

mtlynch:
>Thanks! At one point I had misspell set up as a pre-commit hook, but I lost it after a system refresh

Have you considered checking in your hooks? I generally just check them in and have an `enable_hooks` command that other developers can use to turn my hooks on in their client, e.g.:

https://github.com/mtlynch/gofn-prosper/tree/master/hooks

>and of course, pre-commit hooks only work on new code, not existing code

I'm not sure what you mean. They're just arbitrary scripts, so can't they run over whatever code you want them to run over?

lukechampine:
ah, you're right, I forgot that it was just a quirk of my setup that it only examined the code in the commit. I'll look into checking in `misspell` (and perhaps other hooks as well).

DavidVorick:
I am also okay with fixing the Marshall/Marshal thing, though I'm pretty sure that both spellings are technically correct.

PR #1710, opened 2017-04-16
add glyphcheck to Travis

lukechampine:
Once this is merged, #1703 should fail. :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ go:
 install:
   - make dependencies
   - test -z "$(go fmt ./...)"
+  - glyphcheck ./...
   - make
 
 script: make test && make test-long && make cover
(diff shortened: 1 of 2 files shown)


DavidVorick:
#1703 already fails.

DavidVorick:
Glyphcheck repo needs a test suite. Also, does it only check for the homoglyphs, but leaves all other glyphs behind as okay?

lukechampine:
I suppose it already fails, but now it will fail earlier.
It only checks for homoglyphs. This reduces the number of false positives in non-english code. Perhaps we could use a flag to enable strict ASCII-only mode.

VoidingWarranties:
#1703 only fails because I wrote the test for it the week prior ;)

DavidVorick:
pretty sure it does not check the dependencies. It is a good thought that we should be vendoring them and checking them, but that's too big of a change for now.

Fwiw I'm pretty sure @johnathanhowell ran this check over a ton of repos and nothing came up. I know that's not the same as checking every dependency every time something changes, but it's some initial piece of mind.

I think we're good to merge this for now, and we can make an issue to automate checking our deps from a different PR.

PR #1719, opened 2017-04-18
update siac consensus progress estimation

johnathanhowell:
This PR updates the consensus progress estimation algorithm, using 9 minutes as the block time since blocks generally happen a bit faster than 10 minutes and it's better to estimate a lower progress than a higher progress from a UX perspective.

### DIFF
--- a/siac/consensuscmd_test.go
+++ b/siac/consensuscmd_test.go
@@ -17,27 +17,27 @@ func TestEstimatedHeightAt(t *testing.T) {
 		// Test on the same block that is used to estimate the height
 		{
 			time.Date(2017, time.April, 13, 23, 29, 49, 0, time.UTC),
-			1e5,
+			100e3,
 		},
 		// 4 minutes later
 		{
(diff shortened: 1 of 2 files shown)


siac/consensuscmd_test.go
```diff
@@ -16,28 +16,28 @@ func TestEstimatedHeightAt(t *testing.T) {
...
-			5e4,
+			time.Date(2017, time.April, 13, 23, 29, 49, 0, time.UTC),
+			1e5,
 		},
 		// 4 minutes later
 		{
-			time.Date(2016, time.May, 11, 19, 37, 0, 0, time.UTC),
-			5e4,
+			time.Date(2017, time.April, 13, 23, 33, 49, 0, time.UTC),
+			1e5,
```
DavidVorick: 100e3, same for all the others.

Remember that we do our constants in sets of 3 (e3, e6, e9, etc.) because it maps easily to the thousand/million/billion/etc that people are already familiar with.

PR #1720, opened 2017-04-18
add missing close(finishedChan) to threadedReceiveBlocks

johnathanhowell:


lukechampine:
eesh, nice catch! Does this mean those conns were just hanging around until shutdown?

johnathanhowell:
Seems like it

DavidVorick:
The deadline should have killed it after long enough.

lukechampine:
A deadline will cause the next read/write operation to return an error, but I'm not sure if it actually cleans up the underlying connection.

PR #1721, opened 2017-04-18
unsubscribe when locking wallet

lukechampine:
This fixes a bug where locking the wallet caused the wallet to miss blocks.

This also seems like a good place to discuss the wallet memory usage. Using a global bolt tx speeds up IBD and rescans, but causes transaction data to build up in memory. A shorter commit interval alleviates this somewhat. In my testing, committing the transaction pool (which uses the same db model) once per second caused a peak of 100MB RAM usage, while committing after each block reduced that to 4MB. Committing after each block is prohibitively slow, but if fsync is disabled on the db it runs slightly faster than committing once per second. I suggest that we explore disabling fsync during IBD and rescans. It seems like fsync can be enabled and disabled at will during execution, i.e. we don't need to close and reopen the database.

### DIFF
--- a/modules/wallet/encrypt.go
+++ b/modules/wallet/encrypt.go
@@ -400,6 +400,10 @@ func (w *Wallet) Lock() error {
 	}
 	w.log.Println("INFO: Locking wallet.")
 
+	// Unsubscribe from Consensus Set and Transaction Pool
+	w.cs.Unsubscribe(w)
+	w.tpool.Unsubscribe(w)
+
 	// Wipe all of the seeds and secret keys. They will be replaced upon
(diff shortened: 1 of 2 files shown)


DavidVorick:
Pretty sure that you aren't supposed to unsubscribe when the wallet is locked. The wallet has all of the pubkeys in memory still and should be perfectly able to recognize incoming inputs.

So we'd need to make sure that it's not double-subscribing.

DavidVorick:
I'm firmly against disabling fsync, we're playing with people's money. If the wallet database gets corrupted, it's a big mess for everyone.

lukechampine:
looks like `Unlock` already checks that we are not subscribed before subscribing. So there wasn't ever a bug in the first place.

PR #1723, opened 2017-04-19
more realistic host connectability check

avahowell:
This PR adds a more realistic host connectability checking algorithm. Instead of simply attempting to dial itself and setting its status based on the success or failure of that Dial, this PR adds a new rpc `CheckHost`. `threadedTrackConnectabilityStatus` selects at random two outbound peers and asks them to check the host's netaddress, using `CheckHost`. If both peers agree on the host's ConnectabilityStatus, the host will set its ConnectabilityStatus accordingly.

### DIFF
--- a/modules/host/network.go
+++ b/modules/host/network.go
@@ -24,6 +24,7 @@ import (
 	"github.com/NebulousLabs/Sia/encoding"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/types"
+	"github.com/NebulousLabs/fastrand"
 )
 
 // rpcSettingsDeprecated is a specifier for a deprecated settings request.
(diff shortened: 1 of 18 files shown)


modules/host/network.go
```diff
@@ -80,18 +82,111 @@ func (h *Host) threadedTrackWorkingStatus(closeChan chan struct{}) {
...
+
+	var checkAddress modules.NetAddress
+	err = encoding.ReadObject(conn, &checkAddress, 128)
+	if err != nil {
+		return err
+	}
+
+	// check that the checkAddress resolves to at least one of the remote's IP
+	// addresses.
+	checkHost, _, err := net.SplitHostPort(string(checkAddress))
```
lukechampine: `checkAddress.Host()`

modules/host/network.go
```diff
@@ -80,18 +82,111 @@ func (h *Host) threadedTrackWorkingStatus(closeChan chan struct{}) {
...
+		conn.Close()
+	}()
+	err = h.tg.Add()
+	if err != nil {
+		return err
+	}
+	defer h.tg.Done()
+
+	var checkAddress modules.NetAddress
+	err = encoding.ReadObject(conn, &checkAddress, 128)
```
lukechampine: I think we have a const for this

DavidVorick:
I'm going to put this guy on hold for this release. I think the probability of significant error / bug is way too high, and we don't have time to run it through the test paces.

This can come out in the next non-hotfix release.

PR #1724, opened 2017-04-19
Wallet sync

lukechampine:
This does not include syncing the tpool every 100 blocks. The tpool doesn't track the current height so I wasn't sure how to implement something equivalent. I suppose one option is to sync probabilistically via `fastrand.Intn(100) == 0`.

### DIFF
--- a/modules/wallet/encrypt.go
+++ b/modules/wallet/encrypt.go
@@ -325,6 +325,9 @@ func (w *Wallet) Reset() error {
 		return errUnencryptedWallet
 	}
 
+	w.cs.Unsubscribe(w)
+	w.tpool.Unsubscribe(w)
+
 	err := dbReset(w.dbTx)
 	if err != nil {
 		return err
--- a/modules/wallet/update.go
+++ b/modules/wallet/update.go
@@ -1,14 +1,24 @@
 package wallet
 
 import (
+	"bytes"
 	"fmt"
 	"math"
+	"sort"
 
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/types"
 
 	"github.com/NebulousLabs/bolt"
 )
+
+// historicOutput defines a historic output as recognized by the wallet. This
+// struct is primarily used to sort the historic outputs before inserting them
+// into the bolt database.
+type historicOutput struct {
+	id  types.OutputID
+	val types.Currency
+}
 
 // isWalletAddress is a helper function that checks if an UnlockHash is
 // derived from one of the wallet's spendable keys.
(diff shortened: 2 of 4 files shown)


DavidVorick:
you could just add a stateful counter to whatever object holds the tx, then increment it every time you use it. No need to involve the block height.

modules/wallet/update.go
```diff
@@ -300,6 +297,12 @@ func (w *Wallet) ProcessConsensusChange(cc modules.ConsensusChange) {
 		w.log.Println("ERROR: failed to update consensus change ID:", err)
 	}
 
+	if height, err := dbGetConsensusHeight(w.dbTx); err == nil && height%100 == 0 {
+		// sync every 100 blocks to prevent running transaction from growing
+		// too large
+		w.syncDB()
+	}
```
DavidVorick: This adds an I/O operation every single block. While not as bad as a sync, it's probably better to just keep a counter in memory.

DavidVorick: I'm pretty sure you already fetch the height at some point in one of the above helper functions. Maybe you can pass it through to this point?

Also, there's a big miss here. You aren't counting blocks, you are checking the height. A consensus change can include multiple blocks, and I know at least one contributor is working on code that intends to lump multiple blocks together (resulting in a 60% speedup for IBD). If those blocks come 5 at a time with the offset 51, 56, 61, etc., then this will never trigger.

lukechampine:
ok, I think I will proceed with wrapping the `Tx` type. If it goes well, we may want to consider doing the wrapping in the `persist` package.

DavidVorick:
Can you unlock​/init a wallet on old+new code and report the difference in RES?

lukechampine:
conclusions:
- forcing a sync every 100 updates doesn't seem to have a huge impact on memory usage
- sorting before insertion has a moderate impact on write speed

Just wondering: is there any way to get the value of an output from consensus? Because if consensus could export a method for that, even an expensive one, it would remove the need for the wallet to track every output.

DavidVorick:
Consensus doesn't store the historical outputs, once you spend it it'll delete it from it's database. So the wallet needs a separate database to preserve historic information that the consensus package doesn't keep.

Same for the transaction pool, the tpool needs a list of all transaction ids so it knows which transactions have already been confirmed. Consensus db doesn't store this list.

modules/wallet/encrypt.go
```diff
@@ -325,6 +325,9 @@ func (w *Wallet) Reset() error {
 		return errUnencryptedWallet
 	}
 
+	w.cs.Unsubscribe(w)
+	w.tpool.Unsubscribe(w)
```
DavidVorick: Is this supposed to be here? I thought we determined it was okay.

DavidVorick: oh nvm, this is Reset(), not Lock()

PR #1731, opened 2017-04-19
add -c --cancel flag to siac allowance

johnathanhowell:


lukechampine:
I feel like this makes more sense as a subcommand. It's a bit weird because we use `siac renter setallowance` to change the allowance, but `siac renter allowance cancel` seems most natural for canceling. Hmm.

DavidVorick:
I wasn't a huge fan of `setallowance` in the first place.

johnathanhowell:
Latest commit changes this to a subcommand.

siac/rentercmd.go
```diff
@@ -257,6 +255,15 @@ func renterallowancecmd() {
 `, currencyUnits(allowance.Funds), allowance.Period)
 }
 
+// renterallowancecancelcmd cancels the current allowance.
+func renterallowancecancelcmd() {
+	err := post("/renter", "hosts=0&funds=0&period=0&renewwindow=0")
+	if err != nil {
+		die("error cancelling allowance:", err)
+	}
+	fmt.Println("successfully cancelled allowance.")
```
lukechampine: Should be capitalized. "Allowance cancelled" is fine.

lukechampine:
LGTM

PR #1733, opened 2017-04-20
don't hold lock continuously during subscribe

lukechampine:


DavidVorick:
Should be noted that we did this because thousands of goroutines were piling up while the wallet spent hours unlocking.

The other PR should reduce wallet unlocking by some margin (maybe it runs 50% faster? hard to tell), and this PR should prevent the system from clogging up on scheduling issues, and from falling behind on things like transactions.

Hopefully together these PRs fix up some of the wallet instability that users are still seeing.

PR #1734, opened 2017-04-21
New persist

DavidVorick:
man... diff was too aggresive. json.go is a completely new file, jsonold.go is the old json.go, and if you want to read the code in a more sane way, you should just view it as a standalone file.

Still writing more comprehensive tests, but I wanted to see how the checksum parser worked on Windows. I'm wondering if a different approach to newlines will screw up my byte-parsing.


DavidVorick:
This PR is intended to address the file corruption error that has been occasionally wiping out users files after a hard shutdown. This new method has more safety around it, and notably does not use `os.Rename` at all to preserve safety.

modules/host/contractmanager/writeaheadlog.go
```diff
@@ -318,9 +318,19 @@ func (wal *writeAheadLog) load() error {
...
+	}
+	enc := json.NewEncoder(wal.fileSettingsTmp)
+	if err := enc.Encode(settingsMetadata.Header); err != nil {
+		build.ExtendErr("unable to write header to settings temp file", err)
+	}
+	if err := enc.Encode(settingsMetadata.Version); err != nil {
+		build.ExtendErr("unable to write version to settings temp file", err)
+	}
+	if _, err = wal.fileSettingsTmp.Write(b); err != nil {
+		build.ExtendErr("unable to write data settings temp file", err)
```
lukechampine: what's the difference here?

modules/host/contractmanager/writeaheadlogsync.go
```diff
@@ -178,7 +189,7 @@ func (wal *writeAheadLog) commit() {
 		if err != nil {
 			wal.cm.log.Severe("ERROR: unable to create write-ahead-log:", err)
 		}
-		// Write the metadata into the WAL.
+		// Write the mmetaetadata into the WAL.
```
lukechampine: :|

persist/json.go
```diff
@@ -1,93 +1,238 @@
...
 	if version != meta.Version {
 		return ErrBadVersion
 	}
-	if err := dec.Decode(data); err != nil {
-		return err
+
+	// Read everything else.
+	remainingBytes, err := ioutil.ReadAll(dec.Buffered())
+	if err != nil {
+		return build.ExtendErr("unable to read persisted json object data", err)
```
lukechampine: why not keep using the decoder?

lukechampine: oic, it's so you can calculate the checksum. Could probably achieve the same using a `io.TeeReader` but this is fine

persist/json.go
```diff
@@ -1,93 +1,238 @@
...
+	}()
 	if err != nil {
 		return err
 	}
-	defer file.Close()
-	err = Save(meta, data, file)
+
+	// Write out the data to the real file, with a sync.
+	err = func() (err error) {
+		file, err := os.OpenFile(filename, os.O_RDWR|os.O_TRUNC|os.O_CREATE, 0600)
```
lukechampine: this is just a normal `os.Create`

persist/json.go
```diff
@@ -1,93 +1,238 @@
...
+	} else {
+		// Cryptographic checksum failed, try interpreting a manual checksum.
+		var manualChecksum string
+		err := json.Unmarshal(remainingBytes[:8], &manualChecksum)
+		if err == nil && manualChecksum == "manual" {
+			// Manual checksum is proper. Update the remaining data to exclude
+			// the manual checksum.
+			remainingBytes = remainingBytes[9:]
+		}
+	}
```
lukechampine: doesn't seem like the "checksum exists, but is wrong" case is handled explicitly

DavidVorick: it's handled by providing the data unchecksummed to the json decoder. This is needed to preserve compatibility with old versions calling `SaveFile` on an object that's just a string.

Not that we used it that way anywhere, but I figured this was the best way of handling things.

modules/host/contractmanager/writeaheadlog.go
DavidVorick: not exactly sure what you are asking, but the primary difference is that the Sync isn't going to be called until the next iteration of the sync loop. That could be 500ms away. Sync returns a lot faster if the data is already written to disk, so we're actually writing it early to minimize the total amount of downtime (the contractmanager is stop-the-world while the syncs are happening - if we can do most of the disk writes ahead of time, it'll be stopped for less time).

modules/host/contractmanager/writeaheadlogsync.go
DavidVorick: I don't know if some combination of find-and-replace was involved, but this is the funniest typo I've seen in a while.

modules/miner/persist.go
```diff
@@ -65,15 +68,35 @@ func (m *Miner) initPersist() error {
...
+			if err != nil {
+				return
+			}
+			defer m.tg.Done()
+
+			m.mu.Lock()
+			err = m.saveSync()
+			if err != nil {
+				m.log.Println("ERROR: Unable to save miner persist:", err)
+			}
```
johnathanhowell: This was already merged looks like the lock is never released here

DavidVorick: Looks like you are right. Wish there were static tools to detect this type of mistake

PR #1737, opened 2017-04-22
ratelimit incoming RPCs

DavidVorick:


modules/gateway/consts.go
```diff
@@ -73,6 +73,15 @@ var (
 		Testing:  500 * time.Millisecond,
 	}).(time.Duration)
 
+	// peerRPCDelay defines the amount of time waited between each RPC accepted
+	// from a peer. Without this delay, a peer can force us to spin up thousands
+	// of goroutines per second.
+	peerRPCDelay = build.Select(build.Var{
+		Standard: 1 * time.Second,
+		Dev:      1 * time.Second,
+		Testing:  100 * time.Millisecond,
```
lukechampine: 100ms might be too slow for some tests

modules/gateway/rpc.go
```diff
@@ -173,6 +173,10 @@ func (g *Gateway) threadedListenPeer(p *peer) {
 		// The handler is responsible for closing the connection, though a
 		// default deadline has been set.
 		go g.threadedHandleConn(conn)
+
+		if !g.managedSleep(peerRPCDelay) {
```
DavidVorick: This is per-RPC.

There should be a test, yes. Will work on that in a bit.

DavidVorick: * sorry, this ratelimit is per-peer. So one peer cannot drown out the other peers.

We have another ratelimit that allows only one peer to connect per 6 seconds, and that could potentially be drowned out by an adversary. But it only affects inbound connections. One option we thought of was to require a PoW in the handshake.

DavidVorick:
I'm not sure how seriously the new limits will affect the gateway on main-net, so I'm going to be trying it out today. We should also run an antfarm with this update to make sure that peers can still function normally, send transactions, etc.

DavidVorick:
The test I pushed is failing, but seems like more muxado issues and not a result of anything in the gateway code. I've also tested this code in production to confirm that everything works.

Will be good to switch away from muxado asap. I think we'll see a lot of issues disappear upon doing so.

lukechampine:
Is this good to merge then? Or will it cause future builds to fail? If so, we can wrap it in a `t.Skip("skipped until muxado is dropped")` or something.

DavidVorick:
Yeah. Assuming the build passes this should be good to merge.

modules/host/consts.go
```diff
@@ -254,6 +254,21 @@ var (
...
+			return time.Millisecond * 10
+		}
+		if build.Release == "standard" {
+			return time.Millisecond * 50
+		}
+		if build.Release == "testing" {
+			return time.Millisecond
+		}
+		panic("unrecognized release constant in host - obligationLockTimeout")
+	}()
```
lukechampine: `build.Select`?

DavidVorick: This matches all of the other constants in the host. Figured we'd update them all together in a different PR.

DavidVorick:
looks like ndf build failure, but otherwise lgtm

PR #1747, opened 2017-04-26
Remove historic buckets from wallet database

lukechampine:
The big win here is that fresh wallets no longer need to scan the full blockchain during unlock; they subscribe from `modules.ConsensusChangeRecent`. The wallet db should also shrink considerably; I would estimate that over 95% of its current size is just storing historic outputs. This will go a long way towards decreasing Sia's memory footprint.

The downside is that `ProcessedInputs` only have a valid `Value` field if they were funded by a wallet address. The API documentation has been updated to reflect this change. This doesn't break any known code because `siac` and `Sia-UI` already filter out any inputs that don't come from a wallet address. Worth noting that the Bitcoin wallet also only shows input values for wallet-controlled addresses. For users needing to inspect the value of every input, tools such as the explorer are available.

### DIFF
--- a/modules/wallet/encrypt.go
+++ b/modules/wallet/encrypt.go
@@ -193,29 +193,48 @@ func (w *Wallet) managedUnlock(masterKey crypto.TwofishKey) error {
 	subscribed := w.subscribed
 	w.mu.RUnlock()
 	if !subscribed {
-		// Subscription can take a while, so spawn a goroutine to print the
-		// wallet height every few seconds. (If subscription completes
-		// quickly, nothing will be printed.)
-		done := make(chan struct{})
-		go w.rescanMessage(done)
-		defer close(done)
+		if primarySeedProgress == 0 {
+			// If the wallet has not been used, we don't need to scan the entire
+			// blockchain looking for relevant transactions. Instead, just grab
+			// the current height and subscribe to future blocks.
+			err = w.cs.ConsensusSetSubscribe(w, modules.ConsensusChangeRecent)
+			if err != nil {
+				return fmt.Errorf("wallet subscription failed: %v", err)
+			}
+			w.mu.Lock()
+			err = dbPutConsensusHeight(w.dbTx, w.cs.Height())
+			w.mu.Unlock()
+			if err != nil {
+				return fmt.Errorf("wallet subscription failed: %v", err)
+			}
+		} else {
+			// If the wallet has been used, we need to scan the entire
+			// blockchain. This can take a while, so spawn a goroutine to
+			// print the wallet height every few seconds. (If subscription
+			// completes quickly, nothing will be printed.)
+			done := make(chan struct{})
+			go w.rescanMessage(done)
+			defer close(done)
 
-		err = w.cs.ConsensusSetSubscribe(w, lastChange)
-		if err == modules.ErrInvalidConsensusChangeID {
-			// something went wrong; resubscribe from the beginning
-			err = dbPutConsensusChangeID(w.dbTx, modules.ConsensusChangeBeginning)
-			if err != nil {
-				return fmt.Errorf("failed to reset db during rescan: %v", err)
+			err = w.cs.ConsensusSetSubscribe(w, lastChange)
+			if err == modules.ErrInvalidConsensusChangeID {
+				// something went wrong; resubscribe from the beginning
+				err = dbPutConsensusChangeID(w.dbTx, modules.ConsensusChangeBeginning)
+				if err != nil {
+					return fmt.Errorf("failed to reset db during rescan: %v", err)
+				}
+				err = dbPutConsensusHeight(w.dbTx, 0)
+				if err != nil {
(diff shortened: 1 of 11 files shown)


lukechampine:
The host version weighting needs to be updated.

modules/wallet/encrypt.go
```diff
@@ -193,29 +193,48 @@ func (w *Wallet) managedUnlock(masterKey crypto.TwofishKey) error {
...
+			// the current height and subscribe to future blocks.
+			err = w.cs.ConsensusSetSubscribe(w, modules.ConsensusChangeRecent)
 			if err != nil {
-				return fmt.Errorf("failed to reset db during rescan: %v", err)
+				return fmt.Errorf("wallet subscription failed: %v", err)
 			}
-			err = dbPutConsensusHeight(w.dbTx, 0)
+			w.mu.Lock()
+			err = dbPutConsensusHeight(w.dbTx, w.cs.Height())
+			w.mu.Unlock()
```
lukechampine: this part feels janky. When a new block arrives, does `cs.Height` return the new height before or after updating subscribers? If it's after, this should be safe. It it's before, then this could result in an off-by-one in the height. Would it make more sense to set the height before subscribing?

lukechampine: Of course, the perfect solution to these questions would be to include the blockheight in the consensus change. I may implement this myself.

DavidVorick: you shouldn't be calling `cs.Height()` if you are trying to stay synchronized with the most recent consensus change.

The cc doesn't return the value of the inputs? I would be worried about breaking compatibility with other apps that have already been written. For example, by making this change we may break our exchanges.

We can't suddenly start serving values of '0' out of the api where there used to be non-zero values. We have a lot of people (including several exchanges) using the wallet api and we don't know which calls they are depending on, but we do know that often it's more of the calls than we thought necessary.

lukechampine: In theory the cc could provide the input values, since it knows the value of every unspent output, right? I assume that would require modifying the consensus persist structures though. Same with the height.

I think it's worth contacting the exchanges to see if this will actually break anything. Given that Bitcoin does not provide the value of every input, I would be surprised if the exchanges required it. I don't want to tell users that Sia is eating up gigabytes of RAM+disk and wasting hours of their time to store values that they will never see or care about because we _think_ that someone might need them.

DavidVorick: You are assuming that we have a reliable way to talk to our exchange devs. We don't.

DavidVorick:
Still need to fix the height thing. Doesn't the height get updated while the consensus change is being processed though? You can track it by counting the blocks.

modules/wallet/update.go
```diff
@@ -317,7 +294,7 @@ func (w *Wallet) ProcessConsensusChange(cc modules.ConsensusChange) {
 	if err := w.revertHistory(w.dbTx, cc.RevertedBlocks); err != nil {
 		w.log.Println("ERROR: failed to revert consensus change:", err)
 	}
-	if err := w.applyHistory(w.dbTx, cc.AppliedBlocks); err != nil {
```
DavidVorick: should be able to drop a quick function here to count the blocks, that will give you an accurate height for the wallet I think.

lukechampine:
Before this PR, the wallet only updated its height via `ProcessConsensusChange`. But for the instant-unlock shortcut, it calls `cs.Height()` instead of rescanning the whole chain just to count the number of blocks. After that, the height should be incremented as usual by `ProcessConsensusChange`. The problem is that a block could arrive between the call to `cs.Subscribe` and the call to `cs.Height` . The block would then be processed with the wrong height. On the other hand, if we call `cs.Height` before `cs.Subscribe` and a block arrives between them, the wallet would miss that block, making its height wrong again.
Of course, during the period, the wallet is totally fresh, so there is no chance of us seeing a block relevant to us until after `Unlock` returns (which is why we can skip the full rescan in the first place). So I think it's probably ok to have the wrong blockheight for the short period between `cs.Subscribe` and `cs.Height`.

lukechampine:
Reverted the instant unlocking of a fresh wallet. On my SSD, fresh unlock takes 3.5 minutes. The resulting wallet db is 32KB (probably bolt's initial db size)

DavidVorick:
Can you get the stats for an HDD?

lukechampine:
HDD is sadly much slower: 45 minutes to unlock a fresh wallet. I collected some IO stats and it appears to have read ~4GB and written 1.2GB. Not sure why it's writing so much to disk, but it could have something to do with the fact that the consensus wasn't fully synced.

DavidVorick:
I think this looks good. Sad about the HDD stats, but still a huge improvement. We can dive deeper into it later, for now this needs testing:

wallet tests to prove historic values are calculated correctly
API tests to make sure those historic values still make it into the processed transactions

lukechampine:
is this good to go?

DavidVorick:
Sorry for delay, I just need to find time to write the validation tests

PR #1748, opened 2017-04-28
Contract management

DavidVorick:
I have not gotten to the testing yet, and my notebook has a list of about 20 TODOs that I think will need to be done before this can be merged. But I tried to keep the change as simple as this:

Merge all contract management (tracking if a contract had money, renewing if not enough money, renewing if expiring, retiring if expired, forming new contracts, etc.) into a single loop in the codebase that gets called every time there's a new block. The loop is exclusive, so if you call the loop while another instance of the loop is running, the second call will just silently fail (as the first call is doing all the dirty work).

Some other changes ended up coming through as well. I tried to keep it minimal, but:

+ forming contracts and renewing contracts now has a cancelChan for the dial operation
+ allowance.Funds now reports the amount the host is spending on its active contracts. Trying to set the value to something else is a no-op. Another PR will juggle around what the values mean and figure out how to best preserve compatibility with the API
+ contract formation is a lot smarter. Looks at funds remaining, checks for price gouging, checks for other issues, is more aware of contract usefulness, etc.
+ contract formation is now one at a time, instead of all or nothing

There are two really big things missing here:

+ contractor has no way to provide feedback to the hostdb about failed connections and stuff. If a host is not responding to connections, or drops an upload, or isn't doing downloads, the host needs to have its score appropriately reduced. This will help protect renters against crappy hosts
+ the allowance stuff needs to be handled better. It's not okay to be completely ignoring the 'allowance.Funds' value that the user is setting. What's likely going to happen is that the existing fields are going to be deprecated, but we will map them onto new fields. I'll discuss more details when that PR arrives, but I'm thinking of 'allowance.MaxSpending' as replacing what is currently 'allowance.Funds'. In the API package, it'll translate one field into the other to preserve compatibility. And then what this PR is using as 'Funds' will probably be renamed to 'allowance.ActiveSpending' or something like that.


DavidVorick:
I am sorry for the abysmal commit history. But rebasing was not going to work as there were big conflicts in early commits, the set of merges covers something like 100 commits total.

lukechampine:
might be best to just squash these into one or two commits. They aren't likely to be separable anyway. Just going off the commit messages, they look more like checkpoints than atomic units.

modules/renter/contractor/allowance.go
```diff
@@ -20,40 +18,59 @@ var (
...
+		c.managedUnlockContract(contract.ID)
 	}
-	return endHeight
+
+	// reset currentPeriod and archive all contracts
+	c.mu.Lock()
+	c.allowance = a
+	c.currentPeriod = 0
+	c.mu.Unlock()
+	return
```
johnathanhowell: this `return` seems unnecessary :stuck_out_tongue:

modules/renter/contractor/consts.go
```diff
@@ -4,15 +4,46 @@ import (
...
+	// allowed to have before the host is considered to be unacceptable. The
+	// least competitive of the set of competitive hosts is selected, and its
+	// score is slased by a factor of 'badScoreForgiveness'.
+	badScoreForgiveness = 25
+
 	// estimatedFileContractTransactionSize provides the estimated size of
 	// the average file contract in bytes.
 	estimatedFileContractTransactionSize = 1200
+
+	// misiingMinScans is the minimum number of scans required to judge whether
```
johnathanhowell: s/misiing/missing

modules/renter/contractor/contracts.go
```diff
@@ -0,0 +1,439 @@
...
+	"sort"
+	"time"
+
+	"github.com/NebulousLabs/Sia/build"
+	"github.com/NebulousLabs/Sia/modules"
+	"github.com/NebulousLabs/Sia/sync"
+	"github.com/NebulousLabs/Sia/types"
+)
+
+// callateralFromCost calculates the appropriate collateral to expect from the
```
johnathanhowell: s/callateral/collateral

modules/renter/contractor/contracts.go
```diff
@@ -0,0 +1,439 @@
...
+		c.contractLocks[id] = new(sync.TryMutex)
+	}
+	return c.contractLocks[id].TryLock()
+}
+
+// managedUnlockContract will unlock a locked contract.
+func (c *Contractor) managedUnlockContract(id types.FileContractID) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.contractLocks[id].Unlock()
```
johnathanhowell: Should this be checking for `exists`?

modules/renter/contractor/renew.go
```diff
@@ -57,133 +57,46 @@ func (c *Contractor) managedRenew(contract modules.RenterContract, numSectors ui
...
+	// Archive the old contract.
+	c.oldContracts[contract.ID] = contract
+	// Delete the old contract.
+	delete(c.contracts, contract.ID)
+	// Insert the new contract.
+	c.contracts[newContract.ID] = newContract
+	// Map from the old contract to the new contract.
+	c.renewedIDs[contract.ID] = newContract.ID
+	// Transfer the current cached revision to the new contract id.
+	c.cachedRevisions[newContract.ID] = c.cachedRevisions[contract.ID] // TODO: Is this necessary, won't the revision numbers, etc. be off anyway?
```
lukechampine: yeah, that does seem unnecessary. It was explicitly added though (see https://github.com/NebulousLabs/Sia/commit/523f0b2c22bcfa24b1e9006e9c358a76b3c25cfc), so I must have thought it was necessary for some reason. Better commit messages would have helped here :/
Check if removing it breaks anything, I guess.

modules/renter/renter.go
```diff
@@ -84,7 +84,7 @@ type hostContractor interface {
 	Close() error
 
 	// Contract returns the latest contract formed with the specified host.
-	Contract(modules.NetAddress) (modules.RenterContract, bool)
+	Contract(types.FileContractID) (modules.RenterContract, bool)
```
lukechampine: docstring needs updating

modules/renter/proto/renew.go
```diff
@@ -91,11 +85,15 @@ func Renew(contract modules.RenterContract, params ContractParams, txnBuilder tr
...
+	dialer := &net.Dialer{
+		Cancel:  cancelChan,
+		Timeout: dialHostTimeout,
+	}
+	conn, err := dialer.Dial("tcp", string(host.NetAddress))
 	if err != nil {
 		return modules.RenterContract{}, err
 	}
-	defer func() { _ = conn.Close() }()
+	defer conn.Close()
```
lukechampine: heh, I think this was to placate `errcheck`

modules/renter/repair.go
```diff
@@ -126,7 +126,11 @@ func (r *Renter) addFileToRepairState(rs *repairState, file *file) {
 		// Check whether this contract is offline. Even if the contract is
 		// offline, we want to record that the chunk has attempted to use this
 		// contract.
-		offline := r.hostContractor.IsOffline(contract.ID)
+		var offline bool
+		renterContract, exists := r.hostContractor.Contract(contract.ID)
+		if exists && renterContract.InGoodStanding {
+			offline = true
+		}
```
lukechampine: `offline := exists && renterContract.InGoodStanding`

wait, why is it marked offline if the contract is in good standing?

modules/renter.go
```diff
@@ -192,14 +192,18 @@ func (mrs *MerkleRootSet) UnmarshalJSON(b []byte) error {
 // A RenterContract contains all the metadata necessary to revise or renew a
 // file contract. See `api.RenterContract` for field information.
 type RenterContract struct {
+	HostPublicKey types.SiaPublicKey   `json:"hostpublickey"`
+	ID            types.FileContractID `json:"id"`
+	NetAddress    NetAddress           `json:"netaddress"`
+	SecretKey     crypto.SecretKey     `json:"secretkey"`
+
+	InGoodStanding  bool `json:"ingoodstanding"`  // if false, host is due for replacement
+	UsefulForUpload bool `json:"usefulforupload"` // if false, host cannot be used to upload
```
lukechampine: looks like these aren't set to false anywhere. I'm assuming that will be added in later commits.

modules/renter/contractor/contracts.go
DavidVorick: yeah I guess, but we'll be calling `build.Critical` anyway.

modules/renter.go
DavidVorick: no there are places in contracts.go where each is set to false

modules/renter/repair.go
DavidVorick: ah whoops, should be `!exists || !InGoodStanding`

lukechampine: or `online := exists && InGoodStanding`

modules/renter.go
lukechampine: oops, I forgot to expand that diff lol

modules/renter/repair.go
DavidVorick: going to keep it named offline, want to disturb this code as little as possible. It'll be given a facelift soon enough. Can do `offline := !exists || !InGodStanding` though.

modules/renter/contractor/contracts.go
```diff
@@ -0,0 +1,445 @@
...
+	// not need to be exactly the best performers in order to be acceptable.
+	lowestScore = lowestScore.Div64(badScoreForgiveness)
+
+	// Now that we've finished interacting with external packages, lock the
+	// contractor and begin marking bad contracts.
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	// Check each host for various conditions that would cause them to be
+	contracts := make([]modules.RenterContract, 0, len(c.contracts))
```
lukechampine: to be...

modules/renter/contractor/contracts.go
```diff
@@ -0,0 +1,445 @@
...
+	for _, contract := range c.contracts {
+		contracts = append(contracts, contract)
+	}
+	for _, contract := range contracts {
```
lukechampine: what is the purpose of copying all the contracts into a slice?

modules/renter/contractor/contracts.go
```diff
@@ -0,0 +1,445 @@
...
+		if i >= len(hosts) {
+			break
+		}
+		if hosts[i].RemainingStorage < storageRemainingThreshold {
+			hosts = append(hosts[:i], hosts[i+1:]...)
+		} else {
+			i++
+		}
+
+	}
```
lukechampine:
```go
var hosts []modules.HostDBEntry
for _, host := range c.hdb.RandomHosts(selectionSize, existingHosts) {
    if host.RemainingStorage < storageRemainingThreshold {
        hosts = append(hosts, host)
    }
}
```

modules/renter/contractor/contracts.go
```diff
@@ -0,0 +1,445 @@
...
+
+	// Form contracts with the hosts until we have the desired number of total
+	// contracts.
+	i = 0
+	for numGoodContracts < wantedHosts && i < len(hosts) {
```
lukechampine: I would probably do `for _, host := range hosts` and then add the `numGoodContracts < wantedHosts` check inside the loop. That also lets you skip the last sleep.

modules/renter/contractor/contracts.go
DavidVorick: The contracts in the contractor map are not pointers, they are objects, and you can't directly update an element of an object while it is in a map, you have to create a copy and then insert the copy into the map.

Not something I was aware of until this PR.

modules/renter/contractor/contracts.go
DavidVorick: can't skip the last sleep, it's to allow transactions to propagate the network.

modules/renter/contractor/contracts.go
lukechampine: Right, Go won't let you do `m["foo"].bar = 2` because `"foo"` may not be in the map. Interestingly, it is legal to do `m["baz"] += 2` because that's short for `m["baz"] = m["baz"] + 2`, and `m["baz"]` will return `0` if the key doesn't exist, so the expression is still well-defined.

Anyway, couldn't you still do:
```go
for _, contract := range contracts {
    contract.Field++
    contracts[contract.ID] = contract
}
```

modules/renter/contractor/contracts.go
lukechampine: I meant the "final" sleep that you do after all the contracts have been formed. We need to sleep in between transactions, but not after they've all been formed, right?

modules/renter.go
```diff
@@ -41,10 +41,39 @@ type ErasureCoder interface {
...
+
+	// Spending is the amount of money that the renter has spent on contracts in
+	// the current billing cylce.
+	Spending types.Currency `json:"spending"`
+
+	// UploadHosts defines the number of hosts that the renter needs to keep
+	// available for upload. The actual number of contracts may be higher than
+	// this, as the renter may have hosts that it cannot upload to, or otherwise
+	// that the renter does not consider reliable.
+	UploadHosts uint64 `json:"uploadhosts"`
```
DavidVorick: Please review the new allowance setup. Rest of the code has not been updated to reflect this. Keeping compatibility in the API is going to be annoying, but we basically just have to map the following fields:

Funds -> Budget
Hosts -> UploadHosts
RenewWindow -> GracePeriod
Period -> (BillingPeriod = Period-GracePeriod)

`Spending` and `ActiveFunding` are both fields that cannot be set by the user. I'm not sure exactly how to work this, but I think we'll probably want to switch to a SetSettings type of function like we have in the other modules.

modules/renter.go
```diff
@@ -41,10 +41,39 @@ type ErasureCoder interface {
...
+	// contracts and files for the renter will be lost. This means that the
+	// maximum amount of downtime that a renter can have without risk of losing
+	// files is 'GracePeriod', and the maximum amount of downtime a renter can
+	// have before being guaranteed to lose files is 'BillingPeriod +
+	// GracePeriod'.
+	GracePeriod types.BlockHeight `json:"graceperiod"`
+
+	// Spending is the amount of money that the renter has spent on contracts in
+	// the current billing cylce.
+	Spending types.Currency `json:"spending"`
```
johnathanhowell: Can you clarify the difference between `ActiveFunding` and `Spending`?

DavidVorick:
Closing, as a separate chain of pull requests have covered the most significant functionality here.

Issue #1752, opened 2017-04-28
[enhencement] add cobra manpage generation command for siac


vevarm:
I will create a pr for this one.

lukechampine:
Implemented in #1753

PR #1753, opened 2017-04-28
[enhencement] add cobra manpage generation command for siac

vevarm:
I am kind of newbie on this so take it easy on me. 

https://github.com/NebulousLabs/Sia/issues/1752

//Dinkel

### DIFF
--- a/siac/main.go
+++ b/siac/main.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/bgentry/speakeasy"
 	"github.com/spf13/cobra"
+
 	"github.com/NebulousLabs/Sia/api"
 	"github.com/NebulousLabs/Sia/build"
 )
--- a/siac/main.go
+++ b/siac/main.go
@@ -11,7 +11,7 @@ import (
 
 	"github.com/bgentry/speakeasy"
 	"github.com/spf13/cobra"
-
+	"github.com/spf13/cobra/doc"
 	"github.com/NebulousLabs/Sia/api"
 	"github.com/NebulousLabs/Sia/build"
 )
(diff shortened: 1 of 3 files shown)


vevarm:
Also possibly this should be refactored into. 

siac gen manpage
siac gen bash-completion

or similar

siac/main.go
```diff
@@ -11,7 +11,6 @@ import (
 
 	"github.com/bgentry/speakeasy"
 	"github.com/spf13/cobra"
-
```
johnathanhowell: this newline should be here, we like to have a newline between the non-`NebulousLabs` dependencies

siac/mangencmd.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+var (
+	mangenCmd = &cobra.Command{
+		Use:   "man-generation [path]",
+		Short: "Creates unix style manpages.",
+		Long: "Creates a man pages at the specified " +
+			"directory.\n\n",
+
+		Run: wrap(mangencmd),
```
lukechampine: no extra newline necessary. Also, `Long` comments should use ` rather than ". Follow the style of other commands in the same package.

siac/mangencmd.go
```diff
@@ -0,0 +1,28 @@
...
+
+func mangencmd(path string) {
+	header := &doc.GenManHeader{
+		Section: "1",
+		Manual:  "siac Manual",
+		Source:  "",
+	}
+
+
+	doc.GenManTree(rootCmd, header, path)
```
lukechampine: looks like gofmt wasn't run on this

siac/main.go
vevarm: Yep, corrected this was unintended and clumpsy.

siac/mangencmd.go
vevarm: Corrected.

vevarm:
I saw build broke due to secondary depency to github.com/cpuguy83/go-md2man/md2man from cobra. I dont know where this import should be placed.

DavidVorick:
Just need to modify `make dependencies` in the makefile to `go get` that package.

lukechampine:
It should be:
```
go get -u github.com/spf13/cobra/...
```
(i.e. with trailing `...`)

I'm kinda surprised that Go doesn't grab that package by default though. When I checked my gopath, I already had `md2man`. Maybe cobra reorganized their packages recently?

vevarm:
the ... seems to have done it, now it builds.

PR #1758, opened 2017-05-01
Add node priority

lukechampine:
Picking up where #1648 left off, `node` is now a proper type. It contains a `Priority` bool to indicate that the node should be prioritized when choosing new peers. I wasn't satisfied with calling it `Outbound` or `Inbound`, but I don't think `Priority` is an amazing name either. Maybe we can come up with something better.

@GlenDC, I added you as the author of the first commit here to note your contributions in #1648. Hope that's alright. The main departure from #1648 is that `addNode` still takes just a `NetAddress`, and `g.prioritizeNode` was added to set the priority of a node. This seemed reasonable because we may add a node to the node list long before we actually connect to it (and thus want to prioritize it for future connections).

The current priority algorithm is to pick a random node three times, returning the node if it is prioritized, and otherwise falling back to a final random node. "Three" is a magic number here and should probably be proportional to the size of the list, e.g. `len(g.nodes)/3`. An alternative would be to select randomly from all the prioritized nodes first, and then fallback to the non-prioritized nodes -- but I felt this could cause issues if there were only 1 or 2 prioritized nodes and neither of them were reachable.

This could use more testing.


GlenDC:
@lukechampine I'm happy that you continued with this. I'm really swamped with work these days, so sorry that I couldn't continue this. Great that it's getting some love from you guys!

modules/gateway/nodes.go
```diff
@@ -18,6 +18,18 @@ var (
 	errPeerGenesisID = errors.New("peer has different genesis ID")
 )
 
+// A node represents a potential peer on the Sia network. The Prioritize field
+// indicates whether the node should be prioritized when forming new
+// connections.
+type node struct {
+	NetAddress modules.NetAddress `json:"netaddress"`
+	Prioritize bool               `json:"prioritize"`
```
DavidVorick: Outbound definitely makes more sense to me

modules/gateway/gateway.go
```diff
@@ -271,6 +264,8 @@ func New(addr string, bootstrap bool, persistDir string) (*Gateway, error) {
 			if err != nil && err != errNodeExists {
 				g.log.Printf("WARN: failed to add the bootstrap node '%v': %v", addr, err)
 			}
+			// All bootstrap nodes should be prioritized.
+			g.prioritizeNode(addr)
```
DavidVorick: I don't think you should prioritize the bootstrap node, they are obvious targets for attack.

modules/gateway/peers.go
```diff
@@ -499,33 +479,7 @@ func (g *Gateway) managedConnectv120Peer(conn net.Conn, remoteVersion string, re
...
-		},
-		sess: muxado.Client(conn),
-	})
-	// Add the peer to the node list. We can ignore the error: addNode
-	// validates the address and checks for duplicates, but we don't care
-	// about duplicates and we have already validated the address by
-	// connecting to it.
-	g.addNode(remoteAddr)
-	return g.saveSync()
+	return g.managedConnectv100Peer(conn, remoteVersion, remoteAddr)
```
DavidVorick: Was this just a bunch of duplicated code?

modules/gateway/peersmanager.go
```diff
@@ -60,6 +60,25 @@ func (g *Gateway) numOutboundPeers() (numOutboundPeers int) {
...
+		return "", errNoPeers
+	}
+
+	// Try three times to select a prioritized node.
+	for i := 0; i < 3; i++ {
+		addr, _ := g.randomNode()
+		if g.nodes[addr].Prioritize {
+			return addr, nil
+		}
+	}
```
DavidVorick: I think it would be far more secure to scan the whole list and specifically pull out the outbound ones, and then choose randomly from those. If we have connectable outbound peers, we should reuse them, and only fall back to our inbound peers if all remaining outbound connections have failed but we still need more peers.

Outbound is likely to be a tiny percentage of the list, almost worth putting in a separate place in memory.

modules/gateway/nodes.go
lukechampine: "Outbound" feels vague to me. What is an "outbound node?" What we want to express is: "A node that we, at some point, successfully initiated a connection to and retained as a peer." But I couldn't come up with a terse way of saying that. `WasOutboundPeer` is wordy but at least more accurate.

If we do go with `Oubound`, I think we should also convert the `Inbound` field of the `peer` object` to `Outbound`.

modules/gateway/peers.go
lukechampine: yeah. I like this reorg because it makes it clear that v120 is just v100 plus a new handshake.

However, if we merge the smux stuff in at the same time, v120 will no longer be able to share code with v100, so the DRY breaks. Also, I need to remember to update all the v120 references to v130 (assuming our target for the new gateway code is v1.3.0)

modules/gateway/peersmanager.go
lukechampine: ok, so something like:
```go
func permanentPeerManager() {
    // try the previous outbound nodes first
    for _, node := range g.nodes {
        if node.WasOutboundPeer {
            tryConnect(node)
        }
    }

    // afterwards, use random selection
    for {
        tryConnect(g.randomPotentialPeer())
        sleep()
    }
}
```
I'm also thinking that `g.randomPotentialPeer` should filter out any addrs that we're already connected to.

modules/gateway/peers.go
DavidVorick: I believe that the plan is to merge smux in the same release

modules/gateway/peersmanager.go
DavidVorick: Yeah was also just thinking that we should make sure not to double connect.

Above code is approximate okay, but needs more correct randomness

modules/gateway/nodes.go
DavidVorick: Ah yeah, Outbound is pretty vague. The real thing that's important is "we selected them randomly, but at least we know they were not explicitly chosen by an adversary. Also we are careful to keep the number of these selected peers very small."

I was thinking 'trusted', but that really indicates that the user specified them as a trusted peer explicitly. 

'Anchor' I think matches the literature on this stuff.

lukechampine: I considered `Trusted` too but rejected it for the same reason -- too strong a word.
I'll go with `WasOutboundPeer` for now.

modules/gateway/peersmanager.go
```diff
@@ -73,6 +74,29 @@ func (g *Gateway) permanentPeerManager(closedChan chan struct{}) {
...
+	// connection. The generator will first draw randomly from
+	// previousOutboundPeers, and then fallback to random node selection.
+	perm := fastrand.Perm(len(previousOutboundPeers))
+	nextNode := func() (addr modules.NetAddress, err error) {
+		if len(perm) > 0 {
+			addr, perm = previousOutboundPeers[perm[0]], perm[1:]
+			return addr, nil
+		}
+		return g.randomNode()
+	}
```
lukechampine: this is a little fancy, but the alternative would be to either duplicate the following `for` loop or mangle it to accommodate both sources.

modules/gateway/nodes.go
```diff
@@ -18,6 +18,16 @@ var (
 	errPeerGenesisID = errors.New("peer has different genesis ID")
 )
 
+// A node represents a potential peer on the Sia network.
+type node struct {
+	NetAddress      modules.NetAddress `json:"netaddress"`
+	WasOutboundPeer bool               `json:"wasoutboundpeer"`
+}
+
+func (n node) String() string {
```
DavidVorick: docstring pls

modules/gateway/persist.go
```diff
@@ -78,3 +76,22 @@ func (g *Gateway) threadedSaveLoop() {
...
+		return err
+	}
+	for _, addr := range nodes {
+		err := g.addNode(addr)
+		if err != nil {
+			g.log.Printf("WARN: error loading node '%v' from persist: %v", addr, err)
+		}
+	}
+	return nil
+}
```
DavidVorick: Code looks good but it needs a test demonstrating that the gateway is able to load legacy persist correctly. My preference would be for using a file in the `testdata` dir and verifying it loads correctly. No reason not to just pull a gateway persist from an existing node.

modules/gateway/peersmanager.go
DavidVorick: both sources? Isn't `nextNode()` only called in one place?

More importantly though, you actually want the loop that creates `previousOutboundPeers` to be inside the inifinite loop of this function (line 100). That's because it runs for a long time, potentially days, during which your outbound peers may go online and offline frequently.

Currently, the `previousOutboundPeers` loop will only ever collect peers that you had as outbound when you first booted. But it's very likely that you'll pick up a few more outbound peers as you run, and you want to make sure they are added to the current pool you are selecting from.

lukechampine: Okay, lmk what you think of the change I just pushed.

modules/gateway/peersmanager.go
```diff
@@ -73,71 +72,97 @@ func (g *Gateway) permanentPeerManager(closedChan chan struct{}) {
...
+
+	// swap the outbound nodes to the front of the list
+	numOutbound := 0
+	for i, node := range nodes {
+		if g.nodes[node].WasOutboundPeer {
+			nodes[numOutbound], nodes[i] = nodes[i], nodes[numOutbound]
+			numOutbound++
+		}
+	}
+	return nodes
```
DavidVorick: This code should have a quick test showing it works as implemented. Pretty sure it does, but it's an easy test.

DavidVorick:
PR needs an overall test in the gateway package showing that the gateway will prefer to connect to it's previous outbound peers at start, and probably another to show it'll keep to its previous outbound peers if they have low uptime. (Say have 2 peers that take turns going offline, if the node has those two as outbound initially it should switch between them as they take turns rather than pulling in a new node)

lukechampine:
ok, tests added

modules/gateway/peersmanager.go
```diff
@@ -31,6 +31,11 @@ func (g *Gateway) managedPeerManagerConnect(addr modules.NetAddress) {
 			// race condition could mean that the peer was disconnected
 			// before this code block was reached.
 			p.Inbound = false
+			if n, ok := g.nodes[p.NetAddress]; ok && !n.WasOutboundPeer {
+				n.WasOutboundPeer = true
+				g.nodes[n.NetAddress] = n
+				g.saveSync()
```
DavidVorick: don't need to saveSync for automated actions, it'll get picked up by the save loop after `saveFrequency` anyway.

modules/gateway/peers_test.go
```diff
@@ -1087,30 +1087,22 @@ func TestPeerManagerOutboundSave(t *testing.T) {
...
 			for _, node := range g.nodes {
 				if node.WasOutboundPeer {
 					outboundNodes++
 				}
 			}
 			g.mu.RUnlock()
-
-			if outboundPeers < wellConnectedThreshold {
-				return errors.New("not enough outbound peers: " + strconv.Itoa(outboundPeers))
-			}
```
DavidVorick: Do we still need this code removed? I'm thinking not.

lukechampine: test fails nondeterministically if it's there. Sometimes a peer will get stuck at 2 or 3 peers, and each time it tries to connect the other peer will reject it as being already connected. Worth looking into more I guess, but I debugged it for a while and got nowhere.

DavidVorick: yeah, probably points to a bug, perhaps even one of the bugs related to peers not connecting well. I guess I'll poke at it.

modules/gateway/persist.go
lukechampine: Added in https://github.com/NebulousLabs/Sia/commit/de43b04b95e8cee1e9e3b5d49c2472455b9c839a

modules/gateway/peersmanager.go
lukechampine: Added in https://github.com/NebulousLabs/Sia/commit/4a09c3770ca518a9bc3032ed4a7e2e0b64b6b4c8

lukechampine:
I think I've found the root cause. Basically, if a peer hard-disconnects from you (without properly shutting down the muxado session), you won't know it until the next time you call an RPC on them. So I think what's happening in the test is that peers are getting disconnected, but not realizing it. The failure mode is that one gateway will keep trying to connect to another, but the other gateway rejects it because it thinks it's already connected.

To rectify this I modified `managedRPC` so that it disconnects from the peer immediately if it can't initiate a new stream. Next I added some hack code to the peermanager that calls the ShareNodes RPC in a tight loop on random peers. This has the effect of quickly noticing when a peer has disconnected from us.

I've run the test a dozen times now and can no longer trigger a failure. However, calling RPCs in a tight loop is clearly a bit of a hack. I'm not sure how best to address this, but it shouldn't be a big deal in real-world scenarios.

DavidVorick:
seems like it's just a matter of setting a smaller heartbeat interval and heartbeat timeout in muxado/smux

lukechampine:
If you get a chance, you can test this by changing the heartbeat values in muxado. They are hard-coded so you need to modify the source, but that will at least allow you to confirm that changing the heartbeat fixes it. You will likely need to run the test a couple times before it fails.

lukechampine:
We should merge this, as it's blocking the smux stuff. The disconnect bug should be fixed, but it's been running on live nodes for ages now and doesn't appear to cause serious problems.

Issue #1763, opened 2017-05-03
[enhancement] add static siad manpages


vevarm:
Had a discussion with nemo and we agreed best to create static manpages for siad since it does not move so much. Will make a pull request for it.

PR #1764, opened 2017-05-03
Add static manpages for siad

vevarm:
https://github.com/NebulousLabs/Sia/issues/1763


vevarm:
May I ask why 1 out of 2 checks fails? I just added documentation.

lukechampine:
Our tests fail non-deterministically due to a race condition in our stream multiplexing library. We are working to replace it in the next release, but in the meantime we have to restart builds from time to time.

vevarm:
muxado bug?

lukechampine:
yup

doc/manpages/siad.1
```diff
@@ -0,0 +1,156 @@
...
+siad version
+.br
+siad modules
+.br
+siad [OPTIONS]
+
+.SH DESCRIPTION
+.PP
+The Sia daemon, siad, is a background process that does all of the heavy lifting for Sia. The daemon downloads the blockchain, uploads files, manages redundancy, monitors hosts for uptime, to name a few of the many tasks that it performs.
+The daemon is composed of modules, and each modules provides a set of primary functions on the network.
```
lukechampine: This section has to be written manually, right? That's why I figured we would want the siad manpages to be static instead of autogenerated, because the autogenerated version wouldn't have a custom description. But if cobra can autogenerate manpages that contain a custom description, we can probably stick with that.

vevarm:
background was that I could not tdchnically figure out howto feed corba manpage generation a directory, siac and siad differs in corba structure. this was discussed in #contrbutors 1st  may. After some posts nemo writes:

nemo:
for siad we probably want a static and mostly explanatory man page, since it doesn't have many options and they rarely change. siac commands are added much more frequently so auto-generating the man page is useful.

vevarm:
I have manully added alot regarding modules basically the printout from the siad modules command. Cobra is completly unaware of the options for the modules. That is the main downside if we would make it autogen instead. I can tell you got better understanding of cobra when working on siac and if you would redo the work then siad would be modelled differently, but it is what it is. 

I have no strong opinion, we can do whatever you like. Switch to autogen again, but make up your mind.

vevarm:
Also I think we can add this intro text into the generation, somehow I expect there be a way since it such an obvoius thing to do.

lukechampine:
Yeah, I think if we can add the intro to the generation, it's good enough. Using autogen may mean that we lose the modules descriptions, but that's not too bad. See https://godoc.org/github.com/spf13/cobra/doc. I will try to get this working and credit you in the commit.

lukechampine:
I wasn't able to find an easy way to specify a custom description. I think this is sufficient for now.

PR #1767, opened 2017-05-04
Gracefully handle absent storage folders

DavidVorick:
Wanted some comments and ideas on the overall design. This PR (assuming it builds - didn't check b/c 5am) adds code to gracefully handle the situation where siad loads and a storage folder is not available.

Today the behavior is to delete that storage folder from memory, including all of the metadata. This means the host will lose that storage folder if they load siad at boot and the host starts up before the storage folder is available. v1.1.2 and earlier would fail silently, dropping all historic data as well and then putting the new data in the wrong location.

New behavior basically will still report the storage folder to the user, and will still retain all metadata, but will return some gross error statistics, and it will also prevent all operations involving the storage folder (reads + writes, namely).

I could also add some sort of background check which regularly checks whether the storage folder has been added and then logs an error if not, and marks the storage folder as available (opening all needed file handles) if it does? That part isn't implemented yet.

This PR could have more safety, though not quite sure how best to achieve it. Mostly this moves the host in the direction of robustness.

Side note: right now we have a json bitfield being encoded as a series of uint64s. Json prints numbers as a string, meaning that we end up with things taking 20+ bytes each (plus formatting overhead - burdens the CPU) when they should really only be taking up 8 bytes each (and ideally no formatting overhead). Not sure if there's an easy encoding fix here but @lukechampine it'd make the host run a lot smoother, and also it'd make the disks last longer. A better solution would not have us writing out a giant bitfield in JSON every half second, but that's a more difficult change to make.

### DIFF
--- a/modules/host/contractmanager/contractmanager.go
+++ b/modules/host/contractmanager/contractmanager.go
@@ -20,9 +20,15 @@ package contractmanager
 // TODO: Re-write the WAL to not need to do group syncing, and also to not need
 // to use the rename call at all.
 
+// TODO: When a storage folder is missing, operations on the sectors in that
+// storage folder (Add, Remove, Delete, etc.) may result in corruption and
+// inconsistent internal state for the contractor. For now, this is fine because
+// it's a rare situation, but it should be addressed eventually.
+
 import (
 	"errors"
 	"path/filepath"
+	"sync/atomic"
 
 	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/crypto"
@@ -165,6 +171,13 @@ func newContractManager(dependencies dependencies, persistDir string) (*Contract
 		defer cm.wal.mu.Unlock()
 
 		for _, sf := range cm.storageFolders {
+			// No storage folder to close if the folder is not available.
+			if atomic.LoadUint64(&sf.atomicUnavailable) == 1 {
+				// File handles will either already be closed or may even be
+				// nil.
+				continue
+			}
+
 			err = sf.metadataFile.Close()
 			if err != nil {
 				cm.log.Println("Error closing the storage folder file handle", err)
@@ -178,7 +191,9 @@ func newContractManager(dependencies dependencies, persistDir string) (*Contract
 
 	// The sector location data is loaded last. Any corruption that happened
 	// during unclean shutdown has already been fixed by the WAL.
-	cm.loadSectorLocations()
+	for _, sf := range cm.storageFolders {
+		cm.loadSectorLocations(sf)
+	}
 
 	// Launch the sync loop that periodically flushes changes from the WAL to
 	// disk.
--- a/modules/host/contractmanager/contractmanager.go
+++ b/modules/host/contractmanager/contractmanager.go
@@ -192,6 +192,12 @@ func newContractManager(dependencies dependencies, persistDir string) (*Contract
 	// The sector location data is loaded last. Any corruption that happened
 	// during unclean shutdown has already been fixed by the WAL.
 	for _, sf := range cm.storageFolders {
+		if atomic.LoadUint64(&sf.atomicUnavailable) == 1 {
+			// Metadata unavailable, just count the number of sectors instead of
+			// loading them.
+			sf.sectors = uint64(len(usageSectors(sf.usage)))
+			continue
+		}
 		cm.loadSectorLocations(sf)
 	}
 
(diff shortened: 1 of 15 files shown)


modules/host/contractmanager/contractmanager.go
```diff
@@ -165,6 +165,13 @@ func newContractManager(dependencies dependencies, persistDir string) (*Contract
 		defer cm.wal.mu.Unlock()
 
 		for _, sf := range cm.storageFolders {
+			// No storage folder to close if the folder is not available.
+			if atomic.LoadUint64(&sf.atomicAvailable) == 1 {
```
johnathanhowell: it's a bit confusing to me that `sf.atomicAvailable == 1` means the folder is unavailable and not `sf.atomicAvailable == 0`

modules/host/contractmanager/persist_test.go
```diff
@@ -0,0 +1,137 @@
...
+	defer cmt.panicClose()
+
+	// Add a storage folder to the contract manager tester.
+	storageFolderDir := filepath.Join(cmt.persistDir, "storageFolderOne")
+	// Create the storage folder dir.
+	err = os.MkdirAll(storageFolderDir, 0700)
+	if err != nil {
+		t.Fatal(err)
+	}
+	err = cmt.cm.AddStorageFolder(storageFoldeRDir, modules.SecotrSize*storageFolderGranularity*2)
```
johnathanhowell: sp

modules/host/contractmanager/contractmanager.go
johnathanhowell: ah, looks like it is `atomicUnavailable` elsewhere in this PR

modules/host/contractmanager/sector.go
```diff
@@ -147,6 +147,10 @@ func (cm *ContractManager) ReadSector(root crypto.Hash) ([]byte, error) {
 		cm.log.Critical("Unable to load storage folder despite having sector metadata")
 		return nil, ErrSectorNotFound
 	}
+	if atomic.LoadUint64(&sf.atmoicUnavailable) == 1 {
```
lukechampine: I don't know how you can live without autocomplete on identifiers lol. It probably saves me thousands of keystrokes per day.

modules/host/contractmanager/storagefolder.go
```diff
@@ -90,6 +90,11 @@ type storageFolder struct {
 	atomicSuccessfulReads  uint64
 	atomicSuccessfulWrites uint64
 
+	// Atomic bool indicating whether or not the storage folder is available. If
+	// the storage folder is not available, it will still be loaded but return
+	// an error if it is queried.
+	atomicUnavailable uint64 // uint64 for alignment
```
lukechampine: the size of a field doesn't impact its alignment. This field is 8-byte aligned because all the previous fields are 8 bytes.

lukechampine:
If the rest of the file is JSON then the most compact encoding is a `[]byte` encoded as a base-64 string.

modules/host/contractmanager/contractmanager.go
DavidVorick: Yeah changed the name halfway through

modules/host/contractmanager/sector.go
DavidVorick: Maybe time to invest

modules/host/contractmanager/storagefolder.go
DavidVorick: Alignment of fields that might be added later

DavidVorick:
This commit is ready. I'm intending to add another commit where a background thread continuously scans for the storage folder, bringing it back up automatically if it finds it.

DavidVorick:
This guy should be ready to go now as well. The diff is almost 80% tests.

johnathanhowell:
Looks like one of your tests isn't compiling

johnathanhowell:
Seems a bit weird to me to accomplish this functionality by setting `failedreads` and `failedwrites` to `999999`, why not just add an extra bit to `StorageFolderMetadata` (`unavailable`?)

DavidVorick:
This method doesn't change the API, which means we can signal to users running the current UI that something is amiss. Otherwise we'd need to

+ extend api
+ extend siac
+ extend ui

Which is all fine, but we're crunched for time right now, and this seemed close enough.

johnathanhowell:
rebased, hopefully this fixes the NDFs on appveyor

PR #1770, opened 2017-05-05
do not count offline contracts in redundancy reporting

johnathanhowell:
Code and commits should be relatively easy to follow. This adds a field `Offline` to the `fileContract`s and changes `f.redundancy()` so it doesn't count pieces from offline contracts. This should more accurately reflect the file's redundancy on the network.

### DIFF
--- a/api/renterhost_test.go
+++ b/api/renterhost_test.go
@@ -1329,5 +1329,137 @@ func TestHostAndRenterRenewInterrupt(t *testing.T) {
 	if bytes.Compare(orig, download) != 0 {
 		t.Fatal("data mismatch when downloading a file")
 	}
+}
 
+// TestRedundancyReporting verifies that redundancy reporting is accurate if
+// contracts become offline.
+func TestRedundancyReporting(t *testing.T) {
+	if testing.Short() {
+		t.SkipNow()
+	}
+	st, err := createServerTester(t.Name())
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer st.server.Close()
+	stH1, err := blankServerTester(t.Name() + " - Host 2")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer stH1.server.Close()
+	testGroup := []*serverTester{st, stH1}
+
+	// Connect the testers to eachother so that they are all on the same
+	// blockchain.
+	err = fullyConnectNodes(testGroup)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Make sure that every wallet has money in it.
+	err = fundAllNodes(testGroup)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Add storage to every host.
+	err = addStorageToAllHosts(testGroup)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Announce every host.
+	err = announceAllHosts(testGroup)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Set an allowance with two hosts.
+	allowanceValues := url.Values{}
+	allowanceValues.Set("funds", "50000000000000000000000000000") // 50k SC
+	allowanceValues.Set("hosts", "2")
+	allowanceValues.Set("period", "10")
+	err = st.stdPostAPI("/renter", allowanceValues)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Create a file to upload.
+	filesize := int(45678)
+	path := filepath.Join(st.dir, "test.dat")
+	err = createRandFile(path, filesize)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// upload the file
+	uploadValues := url.Values{}
+	uploadValues.Set("source", path)
+	err = st.stdPostAPI("/renter/upload/test", uploadValues)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// redundancy should reach 2
+	success := false
+	var rf RenterFiles
+	for start := time.Now(); time.Since(start) < time.Minute; time.Sleep(time.Millisecond * 100) {
+		st.getAPI("/renter/files", &rf)
+		if len(rf.Files) >= 1 && rf.Files[0].Redundancy == 2 {
+			success = true
+			break
+		}
+	}
+	if !success {
+		t.Fatal("file upload did not complete after one minute")
+	}
+
+	// take down one of the hosts
+	stH1.server.Close()
+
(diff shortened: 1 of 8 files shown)


johnathanhowell:
as @DavidVorick pointed out, this will have to be changed once the contractor changes are merged (IsOffline is gone), but only slightly, and those changes would have had to occurred anyway.

lukechampine:
should we hold off on reviewing this until the contractor changes are merged, then?

johnathanhowell:
This is mostly testing, the critical logic is very small and easy to update

api/renterhost_test.go
```diff
@@ -1326,5 +1326,137 @@ func TestHostAndRenterRenewInterrupt(t *testing.T) {
...
+	for start := time.Now(); time.Since(start) < time.Minute; time.Sleep(time.Millisecond * 100) {
+		st.getAPI("/renter/files", &rf)
+		if len(rf.Files) >= 1 && rf.Files[0].Redundancy == 1 {
+			success = true
+			break
+		}
+	}
+	if !success {
+		t.Fatal("file redundancy did not decrement after closing host after one minute")
+	}
```
DavidVorick: There's a helper function in this package for that.

api/renterhost_test.go
```diff
@@ -1326,5 +1326,137 @@ func TestHostAndRenterRenewInterrupt(t *testing.T) {
...
+	defer stH1.server.Close()
+
+	// add a few blocks, without this this test fails with `wallet has coins spent in incomplete transactions`
+	for i := 0; i < 5; i++ {
+		_, err = stH1.miner.AddBlock()
+		if err != nil {
+			t.Fatal(err)
+		}
+	}
```
DavidVorick: Should only need one block, need to push through the transaction before you can make another.

api/renterhost_test.go
```diff
@@ -1326,5 +1326,137 @@ func TestHostAndRenterRenewInterrupt(t *testing.T) {
...
+	err = addStorageToAllHosts(testGroup)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Announce every host.
+	err = announceAllHosts(testGroup)
+	if err != nil {
+		t.Fatal(err)
+	}
```
DavidVorick: Can pull the newlines between these setup operations

api/renterhost_test.go
johnathanhowell: I thought so too, but I can confirm that this test fails without calling AddBlock() at least 4 times before announcing

DavidVorick: Uh. Hmm. That probably is a bug that other users are bumping into then. Might be one of our elusive wallet bugs. Can you see why the wallet is returning that error, and why one block doesn't resolve it?

DavidVorick:
Yeah I think this will not be so bad to update.

One thing that worries me here is that the contractor has a rather loose definition of 'IsOffline' and later 'InGoodStanding' - it'll give hosts a few days before demoting them. But, if a bunch of hosts on your particular file are not 'InGoodStanding' technically, but haven't been online enough to earn the demotion yet, you may still want to do the repair.

So the renter probably needs a little more resolution than that, maybe we add a variable like 'CurrenltyOnline' or 'Online', so that if a file does end up with 75% of its spare pieces offline but "InGoodStanding" it knows to do some repairs anyway.

That I think can be out of scope for this PR. Maybe just add a TODO or something.

DavidVorick:
Re: review, I'm pretty sure the changes will be basic. We're just switching from `IsOffline()` to `InGoodStanding`, the core logic should otherwise be the same.

So we can merge this, and the resulting conflicts should be fine to fix in the contractor

johnathanhowell:
latest commit changes this PR to add a `offlineContracts` map to the renter, which gets persisted. `fileContracts` were persisted using the encoding package, so adding the `Offline` field broke persist from old contracts. Latest commit fixes that.

modules/renter/files.go
```diff
@@ -112,8 +112,9 @@ func (f *file) uploadProgress() float64 {
 
 // redundancy returns the redundancy of the least redundant chunk. A file
 // becomes available when this redundancy is >= 1. Assumes that every piece is
-// unique within a file contract. -1 is returned if the file has size 0.
-func (f *file) redundancy() float64 {
+// unique within a file contract. -1 is returned if the file has size 0. It
+// takes one argument, a map of offline contracts for this file.
+func (f *file) redundancy(offlineContracts map[types.FileContractID]bool) float64 {
```
johnathanhowell: should `available()` be updated like this also, so it doesn't count pieces from offline contracts?

johnathanhowell:
Should be ready for verification tests

modules/renter/consts.go
```diff
@@ -32,6 +32,6 @@ var (
 	repairQueueInterval = build.Select(build.Var{
 		Dev:      30 * time.Second,
 		Standard: time.Minute * 15,
-		Testing:  time.Minute * 15,
+		Testing:  10 * time.Second,
```
lukechampine: we should be consistent about this -- `10 * time.Second` or `time.Second * 10`? I prefer the former, and it's more prevalent in the codebase, the biggest exception being `host/consts.go`.

modules/renter/files.go
```diff
@@ -203,8 +212,8 @@ func (r *Renter) FileList() []modules.FileInfo {
 		files = append(files, modules.FileInfo{
 			SiaPath:        f.name,
 			Filesize:       f.size,
-			Available:      f.available(),
-			Redundancy:     f.redundancy(),
+			Available:      f.available(r.offlineContracts),
+			Redundancy:     f.redundancy(r.offlineContracts),
```
lukechampine: could we do this instead?
```go
f.available(r.hostContractor.IsOffline)
```
```go
func (f *file) available(isOffline func(types.FileContractID) bool) bool {
    for {
        if isOffline(fc.ID) {
            continue
        }
    }
}
```
That would remove the need to store the offline state on disk, right?

johnathanhowell: Yeah. Good call.

modules/wallet/wallet.go
```diff
@@ -71,6 +71,8 @@ type Wallet struct {
 	seeds []modules.Seed
 	keys  map[types.UnlockHash]spendableKey
 
+	activeScans map[string]uint64
+
```
lukechampine: looks like this slipped in there

johnathanhowell: oops! fixed, and rebased to the latest master

PR #1778, opened 2017-05-08
remote file repair

johnathanhowell:
This PR leverages RNabel's contributions in #1669 and implements a download-to-upload strategy in case chunks need to be repaired but the file cannot be opened locally, removing the requirement for users to keep a copy of their file locally.


johnathanhowell:
Note that the test in this PR depends on #1770 to pass.

api/renter.go
```diff
@@ -350,49 +379,131 @@ func (api *API) renterDeleteHandler(w http.ResponseWriter, req *http.Request, ps
...
+			return
+		}
 	}
 
-	WriteSuccess(w)
+	if !params.Httpresp {
+		// `httpresp=true` causes writes to w before this line is run, automatically
+		// adding `200 Status OK` code to response. Calling this results in a
+		// multiple calls to WriteHeaders() errors.
+		WriteSuccess(w)
```
DavidVorick: solid comment.

Could probably be outside the `if` braces, but nbd. Definitely readable and understandable either way.

api/renter.go
```diff
@@ -350,49 +379,131 @@ func (api *API) renterDeleteHandler(w http.ResponseWriter, req *http.Request, ps
...
 	}
+	if len(lengthparam) > 0 {
+		length, err = strconv.ParseUint(lengthparam, 10, 64)
+		if err != nil {
+			return nil, build.ExtendErr("could not decode the length as uint64: ", err)
+		}
+	}
+	// Verify that if either offset or length have been provided that both were provided.
+	offparampassed := len(offsetparam) > 0
+	lenparampassed := len(lengthparam) > 0
```
DavidVorick: there's no code verifying that either both are len 0 or neither are len 0 here. I'm split on whether to consider that a parsing error or not, but it seems fine to handle that case here instead of letting something else figure out that you have an offset param but no len param.

DavidVorick: hmm I guess you actually shouldn't worry about that here, let the other code deal with it.

In that case, comment needs to be changed because it says 'verify'

api/renter.go
```diff
@@ -350,49 +379,131 @@ func (api *API) renterDeleteHandler(w http.ResponseWriter, req *http.Request, ps
...
+	// Parse the async parameter.
+	async, err := stringToBool(asyncparam)
+	if err != nil {
+		return nil, build.ExtendErr("async parameter could not be parsed", err)
+	}
+
+	// Verify that either async or httpresp are true but not both.
+	if async && httpresp {
+		return nil, errors.New("only one of the async and httpresp flags can be specified.")
 	}
```
DavidVorick: wrong place, just do the parsing not the verification logic. Verify before you act on the data, that way the verification and action are paired - makes it harder to forget to change the other if you decide to change one.

api/renter.go
```diff
@@ -350,49 +379,131 @@ func (api *API) renterDeleteHandler(w http.ResponseWriter, req *http.Request, ps
...
+	// Ensure that destination is defined if async is set to true.
+	if async && destination == "" {
+		return nil, errors.New("`destination` must be defined if `async=true`")
+	}
 
-	WriteSuccess(w)
+	// Ensure destination is not set if httpresp is defined.
+	if httpresp && destination != "" {
+		return nil, errors.New("`destination` cannot be defined if `httpresp=true`")
+	}
```
DavidVorick: same deal here

api/renter.go
```diff
@@ -350,49 +379,131 @@ func (api *API) renterDeleteHandler(w http.ResponseWriter, req *http.Request, ps
...
+	if httpresp {
+		dw = modules.NewDownloadHttpWriter(w, offset, length)
+	} else {
+		// Ensure that destination is valid beforehands.
+		// Check that the destination path is absolute.
+		if !filepath.IsAbs(destination) {
+			return nil, &Error{"destination must be an absolute path"}
+		}
+		dw = modules.NewDownloadFileWriter(destination, offset, length)
+	}
```
DavidVorick: oh hmm. Do we want the download writer to be created in the parse function? Should we consider a new name for the parse function?

api/renter.go
```diff
@@ -350,49 +379,131 @@ func (api *API) renterDeleteHandler(w http.ResponseWriter, req *http.Request, ps
...
+
+	return &modules.RenterDownloadParameters{
+		Async:        async,
+		DlWriter:     dw,
+		Httpresp:     httpresp,
+		Length:       length,
+		LengthPassed: lenparampassed,
+		Offset:       offset,
+		OffsetPassed: offparampassed,
```
DavidVorick: Seems weird to me to pass bools to indicate whether the length and offset were provided or if they are just 0-value. @lukechampine perhaps you have an idea for how else we could handle this? api doesn't know the actual length of the file. Perhaps '0' could be interpreted as 'whole file'? Offset unset being interpreted as '0' would be correct, and length of '0' doesn't make any sense, so repurposing it doesn't seem too bad.

johnathanhowell: I'm definitely in favor of removing the bools here and just using the zero value, this was part of RNabel's diff that I missed

api/renter.go
```diff
@@ -350,49 +379,131 @@ func (api *API) renterDeleteHandler(w http.ResponseWriter, req *http.Request, ps
...
+	var dw modules.DownloadWriter
+	if httpresp {
+		dw = modules.NewDownloadHttpWriter(w, offset, length)
+	} else {
+		// Ensure that destination is valid beforehands.
+		// Check that the destination path is absolute.
+		if !filepath.IsAbs(destination) {
+			return nil, &Error{"destination must be an absolute path"}
+		}
+		dw = modules.NewDownloadFileWriter(destination, offset, length)
```
DavidVorick: haven't gotten down to the code yet, but you probably actually want these things being created by the renter. The renter internals are going to have different optimizations for each type of writer, and you want the renter to be able to take clues and such to figure out the best way to structure the downloads.

e.g. httpwriter really doesn't want to use slow hosts, while filewriter wants to use as many hosts as it has memory to use.

api/renter.go
johnathanhowell: props to @RNabel

api/renter.go
johnathanhowell: `DownloadWriters` are currently just a simple interface used by the renter as a destination for the data, basically just a drop-in replacement for the `os.File` that the renter used to write to. Maybe we should pass a flag in the `download` struct that tweaks the algorithm the renter uses to download the file? Then you could use different download strategies with any `DownloadWriter`.

modules/renter.go
```diff
@@ -50,12 +55,154 @@ type Allowance struct {
...
+	Filesize    uint64         `json:"filesize"`
+	Received    uint64         `json:"received"`
+	StartTime   time.Time      `json:"starttime"`
+	Error       string         `json:"error"`
+}
+
+// DownloadWriter provides an interface which all output writers have to implement.
+type DownloadWriter interface {
+	WriteAt(b []byte, off int64) (int, error)
+	Location() string
```
DavidVorick: The `Location()` method is confusing to me. How is it used?

johnathanhowell: Agreed, it was confusing to me too. It's used to indicate the destination of the download writer, plugged into the `DownloadInfo` struct and returned from the `downloads` endpoint in the api.

DavidVorick: Does it make sense for the destination to be a part of the download writer? Or should it just be a separate field altogether? Maybe it does make sense.

modules/renter.go
```diff
@@ -50,12 +55,154 @@ type Allowance struct {
...
+}
+
+// DownloadBufferWriter is a buffer-backed implementation of DownloadWriter.
+type DownloadBufferWriter struct {
```
DavidVorick: I'd go ahead and move all of these implementations directly into the renter. I don't see any reason for them to be defined at the modules layer, especially b/c I imagine that there will be a lot of implementation specific context that we add to them later.

modules/renter.go
DavidVorick: The biggest thing I don't like here is that the `Location` has different meanings depending on what you are returning.

api/renter.go
RNabel: I just noticed that there's a grammatical mistake in there. The last "a" should be taken out since "errors" is plural.

api/renter.go
RNabel: Just FYI, I refactored this code to move all checks into the download function in download.go that can be moved there. These lines could be moved into the download handler functions since that is where they are used.

api/renter.go
RNabel: Yeah, I initially did not have these as all values were validated when they were parsed. Since there are 3 states for these bools (passed+true, passed+false, not provided) I had to add a boolean to perform the verification next to their respective usage.

modules/renter.go
RNabel: The download logic in siac relies on the `downloads` endpoint providing a location parameter that matches the destination provided in the download command. Thus, I had to keep this information around. Since the `DownloadWriter` is the destination for all data, it seemed fitting to use it as the place that provides this information instead. 

Having it in DownloadWriter also means that the DownloadWriter implementation defines what the value is (if it is hard-coded as is the case with the Http writer). If it is stored in the struct, this information would somehow have to be supplied when the struct is created.

johnathanhowell: I came to the same conclusion as @RNabel, I think having this as a part of the DownloadWriter interface makes sense. Perhaps `Destination()` would be a better name, though.

RNabel: Maybe it could be called `String()`? It's kind of like a `toString` operation.

johnathanhowell: I think that's what it originally was, but I changed it because it seemed like `String()` should return a string representation of the data, not a string representing where the file was downloaded to.

RNabel: Ah ok, I didn't realize that's the case. `Destination()` sounds like a good idea then.

api/renter.go
lukechampine: `length = 0 -> whole file` seems like a reasonable sentinel to me.

RNabel: Nice. So calls to `/download` with `offset=X` and no `length` would download the entire file starting at that offset? 

Similarly, if `offset` is not specified it could default it to `0` and downloads the entire file (or the first `length` bytes).

lukechampine: it will default to 0 anyway because Go's types are zero-valued if not initialized. So if you provide a length and no offset, it'll be interpreted as length=X offset=0.

As for offset=X length=0, that seems like a bit of a weird case, but I suppose it could be useful for resuming downloads.

johnathanhowell:
Latest commits address review comments, and also fix a case where `/renter/downloadasync` would fail silently (if there is no file with that path).

RNabel:
I just ran a few experiments using my original PR's code, and noticed that it is missing a feature:

When a website requests video it uses a `Range: bytes=START-FINISH` header. It would make sense to add this to this commit, so that siad provides standard webserver semantics.

See the docs for more info: 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests

RNabel:
@johnathanhowell Re HTTP ranged requests: a working (and pretty crude) implementation can be found in the latest commit to #1669.

It basically just a) responds with the filesize to HEAD requests, and 2) handles `Range: start-end` header value.

DavidVorick:
You don't need to wait for each block if you are mining like 5. You can mine 5 of them all in a row from the same miner, then only wait for the last one. This should be faster.

api/renter.go
```diff
@@ -275,8 +287,24 @@ func (api *API) renterContractsHandler(w http.ResponseWriter, _ *http.Request, _
 
 // renterDownloadsHandler handles the API call to request the download queue.
 func (api *API) renterDownloadsHandler(w http.ResponseWriter, _ *http.Request, _ httprouter.Params) {
+	dlq := api.renter.DownloadQueue()
+
+	// Translate []modules.DownloadInfo to []api.DownloadInfo.
+	downloads := make([]DownloadInfo, len(dlq))
+	for i := range dlq {
+		d := dlq[len(dlq)-i-1]
```
lukechampine: is the intent here to reverse? How about instead, keep the order of `DownloadQueue`, and then use `sort.Slice` to order `downloads` as desired?

api/renter.go
```diff
@@ -350,49 +378,118 @@ func (api *API) renterDeleteHandler(w http.ResponseWriter, req *http.Request, ps
...
 
-	siapath := strings.TrimPrefix(ps.ByName("siapath"), "/")
-	fileExists := false
+	api.renterDownloadHandler(w, req, ps)
+}
+
+// parseDownloadParameters parses the download parameters passed to the
+// /renter/download endpoint. Validation of these parameters is done by the
+// renter.
+func (api *API) parseDownloadParameters(w http.ResponseWriter, req *http.Request, ps httprouter.Params) (*modules.RenterDownloadParameters, error) {
```
lukechampine: doesn't look like this needs to be a method of `API`

api/renter.go
```diff
@@ -350,49 +378,118 @@ func (api *API) renterDeleteHandler(w http.ResponseWriter, req *http.Request, ps
...
+	var offset, length uint64
+	var err error
+	if len(offsetparam) > 0 {
+		offset, err = strconv.ParseUint(offsetparam, 10, 64)
+		if err != nil {
+			return nil, build.ExtendErr("could not decode the offset as uint64: ", err)
+		}
+	}
+	if len(lengthparam) > 0 {
+		length, err = strconv.ParseUint(lengthparam, 10, 64)
```
lukechampine: use `fmt.Sscan` for these

api/renter.go
```diff
@@ -350,49 +378,118 @@ func (api *API) renterDeleteHandler(w http.ResponseWriter, req *http.Request, ps
...
-	// Check that the destination path is absolute.
-	if !filepath.IsAbs(destination) {
-		WriteError(w, Error{"destination must be an absolute path"}, http.StatusBadRequest)
+	// Ensure that the async flag is not set to any other value than "true" or "".
+	async := req.FormValue("async")
+	if async == "false" || async != "true" && async != "" {
+		WriteError(w, Error{"/downloadasync require the `async` parameter to be either equal to `true` or be empty."}, http.StatusBadRequest)
 		return
+	} else if async == "" { // Add async if it is not already set to `true`.
+		req.Form["async"] = []string{"true"}
```
lukechampine: use `req.Form.Set("async", "true")`

api/renter.go
```diff
@@ -350,49 +378,118 @@ func (api *API) renterDeleteHandler(w http.ResponseWriter, req *http.Request, ps
...
 // renterDownloadAsyncHandler handles the API call to download a file asynchronously.
 func (api *API) renterDownloadAsyncHandler(w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
-	destination := req.FormValue("destination")
-	// Check that the destination path is absolute.
-	if !filepath.IsAbs(destination) {
-		WriteError(w, Error{"destination must be an absolute path"}, http.StatusBadRequest)
+	// Ensure that the async flag is not set to any other value than "true" or "".
+	async := req.FormValue("async")
+	if async == "false" || async != "true" && async != "" {
+		WriteError(w, Error{"/downloadasync require the `async` parameter to be either equal to `true` or be empty."}, http.StatusBadRequest)
```
lukechampine: don't really see the point of this branch. Could probably just drop it.

api/scan.go
```diff
@@ -35,3 +36,20 @@ func scanHash(s string) (h crypto.Hash, err error) {
 	}
 	return h, nil
 }
+
+// stringToBool converts "true" and "false" strings to their respective
+// boolean value and returns an error if conversion is not possible.
+func stringToBool(param string) (bool, error) {
+	// Parse the async parameter.
```
lukechampine: leftover comment

api/scan.go
```diff
@@ -35,3 +36,20 @@ func scanHash(s string) (h crypto.Hash, err error) {
...
+	var out bool
+	switch {
+	case param == "true":
+		out = true
+	case len(param) == 0 || param == "false":
+		out = false
```
lukechampine: these branches can just `return true, nil` and `return false, nil`

modules/renter.go
```diff
@@ -251,8 +258,8 @@ type Renter interface {
 	// DeleteFile deletes a file entry from the renter.
 	DeleteFile(path string) error
 
-	// Download downloads a file to the given destination.
-	Download(path, destination string) error
+	// DownloadSection performs a download according to the parameters passed, including downloads of `offset` and `length` type.
+	Download(params *RenterDownloadParameters) error
```
lukechampine: this doesn't need to take a pointer. Also, comment should be wrapped

modules/renter.go
```diff
@@ -297,3 +304,13 @@ type Renter interface {
 	// Upload uploads a file using the input parameters.
 	Upload(FileUploadParams) error
 }
+
+// RenterDownloadParameters contains all parameters that can be passed to the `/download` endpoint.
```
lukechampine: the modules package shouldn't reference API endpoints

modules/renter/download.go
```diff
@@ -116,35 +121,58 @@ type (
...
-		chunkSize:   f.chunkSize(),
-		destination: destination,
-		erasureCode: f.erasureCode,
-		fileSize:    f.size,
-		masterKey:   f.masterKey,
-		numChunks:   f.numChunks(),
-		siapath:     f.name,
+	// Calculate chunks to download.
+	min_chunk := uint64(math.Floor(float64(offset / f.chunkSize())))
+	max_chunk := uint64(math.Floor(float64((offset + length) / f.chunkSize())))
```
lukechampine: `math.Floor` is unnecessary, these will be properly truncated by integer division.

modules/renter/download.go
```diff
@@ -116,35 +121,58 @@ type (
...
 )
 
-// newDownload initializes and returns a download object.
-func (r *Renter) newDownload(f *file, destination string, currentContracts map[modules.NetAddress]types.FileContractID) *download {
-	d := &download{
-		finishedChunks: make([]bool, f.numChunks()),
+// newSectionDownload initialises and returns a download object for the specified chunk.
+func (r *Renter) newSectionDownload(f *file, destination modules.DownloadWriter, currentContracts map[modules.NetAddress]types.FileContractID, offset, length uint64) *download {
+	d := &download{}
+	d.initDownload(f, destination)
```
lukechampine: why not `d := newDownload(f, destination)`?

modules/renter/download.go
```diff
@@ -230,35 +261,45 @@ func (cd *chunkDownload) recoverChunk() error {
...
-		return build.ExtendErr("unable to open download destination", err)
+	var result = recoverWriter.Bytes()
+
+	// Calculate the offset. If the offset is within the chunk, the
+	// requested offset is passed, otherwise the offset of the chunk
+	// within the overall file is passed.
+	chunkBaseAddress := cd.index * cd.download.chunkSize
+	chunkTopAddress := chunkBaseAddress + cd.download.chunkSize - 1
+	var off = chunkBaseAddress
+	var lowerBound = 0
```
lukechampine: our style is to use `:=` instead of `var =`

modules/renter/download.go
```diff
@@ -63,20 +65,23 @@ type (
 		atomicDataReceived uint64
 		downloadComplete   bool
 		downloadErr        error
-		finishedChunks     []bool
+		finishedChunks     map[int]bool
```
lukechampine: this should probably use a `uint64` instead of an `int`

modules/renter/download.go
```diff
@@ -116,35 +121,58 @@ type (
...
-		masterKey:   f.masterKey,
-		numChunks:   f.numChunks(),
-		siapath:     f.name,
+	// Calculate chunks to download.
+	min_chunk := uint64(math.Floor(float64(offset / f.chunkSize())))
+	max_chunk := uint64(math.Floor(float64((offset + length) / f.chunkSize())))
 
-		downloadFinished: make(chan struct{}),
-	}
+	makeRange(int(min_chunk), int(max_chunk+1), &d.finishedChunks)
```
lukechampine: this function call should be inlined. It's only used once and its effect isn't immediately obvious.

modules/renter.go
```diff
@@ -50,12 +51,18 @@ type Allowance struct {
...
+	Filesize    uint64         `json:"filesize"`
+	Received    uint64         `json:"received"`
+	StartTime   time.Time      `json:"starttime"`
+	Error       string         `json:"error"`
+}
+
+// DownloadWriter provides an interface which all output writers have to implement.
+type DownloadWriter interface {
+	WriteAt(b []byte, off int64) (int, error)
+	Destination() string
```
lukechampine: what is the purpose of this method?

modules/renter/download.go
```diff
@@ -586,3 +625,145 @@ func (r *Renter) threadedDownloadLoop() {
...
+func (dw *DownloadBufferWriter) WriteAt(bytes []byte, off int64) (int, error) {
+	if len(bytes)+int(off) > len(dw.data) {
+		return 0, errors.New("write at specified offset exceeds buffer size")
+	}
+
+	i := 0
+	for _, b := range bytes {
+		dw.data[off+int64(i)] = b
+		i++
+	}
```
lukechampine: `i := copy(dw.data[off:], bytes)`

modules/renter/download.go
```diff
@@ -586,3 +625,145 @@ func (r *Renter) threadedDownloadLoop() {
...
+
+// WriteAt writes the passed bytes at the specified offset.
+func (dw *DownloadFileWriter) WriteAt(b []byte, off int64) (int, error) {
+	fileOffset := off - int64(dw.offset)
+
+	r, err := dw.f.WriteAt(b, fileOffset)
+	if err != nil {
+		build.ExtendErr("unable to write to download destination", err)
+	}
+	dw.f.Sync()
```
lukechampine: I don't think it's necessary to sync after each write when downloading a file.

modules/renter/download.go
```diff
@@ -586,3 +625,145 @@ func (r *Renter) threadedDownloadLoop() {
...
+
+	return r, err
+}
+
+// DownloadHttpWriter is a http response writer-backed implementation of DownloadWriter.
+// The writer writes all content that is written to the current `offset` directly to the ResponseWriter,
+// and buffers all content that is written at other offsets.
+// After every write to the ResponseWriter the `offset` and `length` fields are updated, and buffer content written until
```
lukechampine: needs to be wrapped

modules/renter/downloadqueue.go
```diff
@@ -44,6 +88,10 @@ func (r *Renter) Download(path, destination string) error {
 	}
 }
 
+//func (r *Renter) DownloadChunk(f *file, index, offset, length uint64) ([]byte, error) {
+//	return
+//}
```
lukechampine: remove

modules/renter/files.go
```diff
@@ -66,7 +66,7 @@ func (f *file) chunkSize() uint64 {
 	return f.pieceSize * uint64(f.erasureCode.MinPieces())
 }
 
-// numChunks returns the number of chunks that f was split into.
+// dlChunks returns the number of chunks that f was split into.
 func (f *file) numChunks() uint64 {
```
lukechampine: wasn't renamed

modules/renter/files_test.go
```diff
@@ -8,7 +8,7 @@ import (
 	"github.com/NebulousLabs/Sia/types"
 )
 
-// TestFileNumChunks checks the numChunks method of the file type.
+// TestFileNumChunks checks the dlChunks method of the file type.
```
lukechampine: wasn't renamed

siac/rentercmd.go
```diff
@@ -96,6 +96,13 @@ have a reasonable number (>30) of hosts in your hostdb.`,
 		Run:   wrap(renterfilesdownloadcmd),
 	}
 
+	renterDownloadSectionCmd = &cobra.Command{
+		Use:   "downloadsection [path] [offset] [length] [destination]",
+		Short: "Download a chunk",
+		Long:  "Download a specific chunk from a previously uploaded file.",
+		Run:   wrap(renterfilesdownloadsectioncmd),
```
lukechampine: do we need to be supporting this in siac? siac is for humans, not machines.

siac/rentercmd.go
```diff
@@ -381,37 +388,7 @@ func renterfilesdeletecmd(path string) {
...
-				elapsed := time.Since(d.StartTime)
-				elapsed -= elapsed % time.Second // round to nearest second
-				mbps := (float64(d.Received*8) / 1e6) / time.Since(d.StartTime).Seconds()
-				fmt.Printf("\rDownloading... %5.1f%% of %v, %v elapsed, %.2f Mbps    ", pct, filesizeUnits(int64(d.Filesize)), elapsed, mbps)
-			}
-		}
-	}()
+	go downloadprogress(done, path)
 
 	err := get("/renter/download/" + path + "?destination=" + destination)
```
lukechampine: should this be using `downloadasync`?

api/renter.go
DavidVorick: Does dlq have a promise on what order things will be returned in? Seems like the type of thing that might get optimized and jumbled as we continue to pursue performance (this stuff is likely to end up in a bunch of maps and random-access disk locations)

api/renter.go
```diff
@@ -350,49 +378,118 @@ func (api *API) renterDeleteHandler(w http.ResponseWriter, req *http.Request, ps
...
+		go func() {
+			errchan <- api.renter.Download(params)
+		}()
+		select {
+		case err = <-errchan:
+			if err != nil {
+				WriteError(w, Error{"download failed: " + err.Error()}, http.StatusInternalServerError)
+				return
+			}
+		case <-time.After(time.Second * 5):
```
DavidVorick: Five seconds is definitely way too long for an async call. It should be waiting maybe 100ms, but even that is probably too much. 20ms if we have faith in the program being able to validate quickly.

We could always add more strict execution. Rather than wait until a certain timeout, we could block on a channel that the call sends down after validation is complete. But let's save that for another PR.

api/renter.go
DavidVorick: if you are calling the async endpoint and also setting the async param to something other  than "true", you've made a mistake. The API should catch things like that. Seems reasonable enough to me that it would happen.

modules/renter.go
```diff
@@ -297,3 +304,13 @@ type Renter interface {
 	// Upload uploads a file using the input parameters.
 	Upload(FileUploadParams) error
 }
+
+// RenterDownloadParameters contains all parameters that can be passed to the `/download` endpoint.
+type RenterDownloadParameters struct {
+	Async       bool
+	Httpwriter  http.ResponseWriter
```
lukechampine: why does this need to be a `http.ResponseWriter`? Why not just an `io.Writer`?

api/renter.go
```diff
@@ -350,49 +378,118 @@ func (api *API) renterDeleteHandler(w http.ResponseWriter, req *http.Request, ps
...
+
+	dp := &modules.RenterDownloadParameters{
+		Destination: destination,
+		Async:       async,
+		Length:      length,
+		Offset:      offset,
+		Siapath:     siapath,
+	}
+
+	if httpresp {
```
DavidVorick: no space here, it's the same idea of filling out `dp`

api/renter.go
lukechampine: I guess we can use `stringToBool` here

appveyor.yml
```diff
@@ -21,5 +21,5 @@ install:
 
 build_script:
   - go test -tags="testing debug" -short ./...
-  - go test -tags="testing debug" -v -race ./...
-  - go test -tags="testing debug" -v -timeout=500s -run=XXX -bench=. ./...
+  - go test -tags="testing debug" -v -timeout=1200s -race ./...
```
DavidVorick: travis I know will time-out if there's no output for 10 minutes, and I also know for some reason the test package won't always flush the buffers inside of 10 minutes if you have a bunch of tests running in parallel. Not sure why, but this could impact appveyor somehow as well. I guess we'd need to refer to the appveyor documentation.

modules/renter.go
DavidVorick: yeah, it's the api that exposes the download parameters, not the download parameters that follow the api commands.

modules/renter/consts.go
```diff
@@ -34,4 +34,12 @@ var (
 		Standard: time.Minute * 15,
 		Testing:  10 * time.Second,
 	}).(time.Duration)
+
+	// maxChunkCacheSize determines the maximum number of chunks that will be
+	// cached in memory.
+	maxChunkCacheSize = build.Select(build.Var{
```
DavidVorick: is this a replacement of the download pieces thing? The combination of download pieces and upload pieces?

DavidVorick: This probably should have been in a different PR, but I guess we have it now.

modules/renter/download.go
DavidVorick: This map seems really memory intensive to me, especially given that progress should be approximately linear. I'm probably worrying too much about memory footprint at this point though.

DavidVorick: maybe I'm interpreting this wrong and the map is kept relatively sparse.

modules/renter/consts.go
johnathanhowell: This is an optimzation for the download-and-reupload algorithm: up to `maxChunkCacheSize` downloaded chunks (used for repair) will be cached in memory in order to reduce the amount of redundant downloading.

modules/renter.go
johnathanhowell: This purpose of this method is to provide a human-readable string representation of the destination which the Download Writer writes to, ie, `httpresp` or `buffer`, `file`, etc.

api/renter_test.go
```diff
@@ -76,11 +74,146 @@ func TestRenterDownloadError(t *testing.T) {
...
+
+	return st, path
+}
+
+func waitForDownloadToComplete(t *testing.T, st *serverTester, siapath string, errmsg string) {
+	var rdq RenterDownloadQueue
+
+	// download should eventually complete
+	success := false
+	for start := time.Now(); time.Since(start) < 30*time.Second; time.Sleep(time.Millisecond * 10) {
```
lukechampine: use `retry`?

api/renter_test.go
```diff
@@ -76,11 +74,146 @@ func TestRenterDownloadError(t *testing.T) {
...
+		}
+	}
+
+	// should have correct length
+	if int64(downbytes.Len()) != length {
+		return errors.New(fmt.Sprintf("downloaded file has incorrect size: %d, %d expected.", downbytes.Len(), length))
+	}
+
+	// should be byte-for-byte equal to the original uploaded file
+	if bytes.Compare(originalBytes.Bytes(), downbytes.Bytes()) != 0 {
```
lukechampine: `bytes.Equal`

api/renter_test.go
johnathanhowell: this func is only used once anyway, should really be inlined with `retry`

api/renter_test.go
```diff
@@ -101,8 +234,118 @@ func TestRenterDownloadError(t *testing.T) {
 	}
 }
 
-// TestRenterAsyncDownloadError tests that the /renter/asyncdownload route sets the download's error field if it fails.
-func TestRenterAsyncDownloadError(t *testing.T) {
+// TestValidDownloads tests valid and boundary parameter combinations.
+func TestValidDownloads(t *testing.T) {
```
lukechampine: this test takes over a minute to run. Maybe drop some test cases or run them in parallel.

api/renter_test.go
```diff
@@ -101,8 +234,118 @@ func TestRenterDownloadError(t *testing.T) {
...
+	if testing.Short() {
+		t.SkipNow()
+	}
+	t.Parallel()
+
+	blk := int64(modules.SectorSize)
```
lukechampine: not a very helpful variable name

api/renterhost_test.go
```diff
@@ -19,6 +20,217 @@ import (
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Check that the download has the right contents.
+	downloaded, err := ioutil.ReadFile(downloadPath)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if bytes.Compare(orig, downloaded) != 0 {
```
lukechampine: `bytes.Equal`

api/renterhost_test.go
```diff
@@ -19,6 +20,217 @@ import (
 	"github.com/NebulousLabs/Sia/types"
 )
 
+// TestRemoteFileRepair verifies that if a trackedFile unavailable locally by
+// being deleted, the repair loop will download the necessary chunks from the
+// living hosts
+func TestRemoteFileRepair(t *testing.T) {
```
lukechampine: great test :+1:

api/renterhost_test.go
```diff
@@ -969,9 +1181,12 @@ func TestRenterRenew(t *testing.T) {
 // TestRenterAllowance sets up an integration test where a renter attempts to
 // download a file after changing the allowance.
 func TestRenterAllowance(t *testing.T) {
+	t.Skip("bypassing NDF")
```
lukechampine: still necessary?

johnathanhowell: I believe so

siac/rentercmd.go
johnathanhowell: I'm against having this in siac

siac/rentercmd.go
johnathanhowell: I don't see a reason for it to be, we want this command to block until the download is finished, using downloadasync here would just introduce a bunch of ugly polling

siac/rentercmd.go
lukechampine: @DavidVorick @RNabel thoughts? We can always add this later if there's a clear need for it.

PR #1779, opened 2017-05-08
Api tests fix

DavidVorick:
Tried to fix up the API tests. This is a work in progress, though if the build is passing I think that it is mergeable. It's just a bunch of small stuff that I ran into while trying to improve our testing.


DavidVorick:
Looks like these cleaned up nicely, except that now the muxado race condition triggers every time.

We have another test that's commented out currently for the same reason - always triggers the muxado bug.

Transition to smux looks better every day. This is tangental, but the heartbeat in muxado right now is way too fast, it's harming our users in China. We have to make it more forgiving if peer connections are going to last over the GFC.

DavidVorick:
build should start passing once the muxado PRs are merged.

DavidVorick:
I ~think that I've gotten everything right finally. Benefits:

+ tests now take less than 2 GB of RAM
+ muxado ndfs should be gone
+ a lot of tests will be somewhat faster
+ the test tooling has been improved slightly.

If the build passes, this should be ready to merge. Not sure that there are any verification tests to write here, as this is basically just all work on the test suite itself.

PR #1783, opened 2017-05-09
fix InitFromSeed concurrency bug

johnathanhowell:
This PR fixes https://github.com/NebulousLabs/Sia-UI/issues/583 and https://github.com/NebulousLabs/Sia-UI/issues/581, the wallet was not holding a lock during the duration of the InitFromSeed call, causing users to unlock their wallets while the seed sweeper was still running, resulting in an incorrect seed progress value.

### DIFF
--- a/modules/wallet/encrypt_test.go
+++ b/modules/wallet/encrypt_test.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
+	"time"
 
 	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/crypto"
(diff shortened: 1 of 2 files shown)


DavidVorick:
In general don't we suffer from a situation where multiple concurrent calls that affect the scanner don't properly exclude eachother? And aren't these types of calls the things you wouldn't want to stack up? For example, if I call init-force multiple times, I think right now what would happen is one would block until the other finished, and then when the other finished the one would start, which could result in the wallet hanging for days if the user got frustrated and tried the same action multiple times after seeing no results.

We should probably have a separate trylock in the wallet that results in certain actions returning an error if the trylock indicates that they are already being performed elsewhere. The `Sia/sync` package already has a trylock implementation.

modules/wallet/encrypt_test.go
```diff
@@ -248,6 +249,50 @@ func TestLock(t *testing.T) {
...
+	// pause for 10ms to allow the seed sweeper to start
+	time.Sleep(time.Millisecond * 10)
+
+	// unlock
+	err = w.Unlock(crypto.TwofishKey(crypto.HashObject(seed)))
+	if err != nil {
+		t.Fatal(err)
+	}
+	// starting balance should match the original wallet
+	newBal, _, _ := w.ConfirmedBalance()
```
lukechampine: have you confirmed that this fails with improper locking? (and race detector enabled?)

lukechampine:
Agreed re: separate lock. It should be easy enough to add a `"Already unlocking"` codepath or similar. But this PR should at least fix the main issue in the meantime.

modules/wallet/encrypt_test.go
johnathanhowell: Yes

modules/wallet/encrypt_test.go
```diff
@@ -248,6 +249,50 @@ func TestLock(t *testing.T) {
...
+	w, err := New(wt.cs, wt.tpool, dir)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// spawn an initfromseed goroutine
+	go w.InitFromSeed(crypto.TwofishKey{}, seed)
+
+	// pause for 10ms to allow the seed sweeper to start
+	time.Sleep(time.Millisecond * 10)
```
lukechampine: Since the seed scanner processes blocks by subscribing to `w.cs`, you could override the `Subscribe` method to e.g. wait on a channel before sending the first block. Then start unlocking before sending down the channel.

PR #1788, opened 2017-05-10
add trylock to wallet

lukechampine:
This should prevent double-subscriptions and accidentally queuing multiple load / sweep operations.

### DIFF
--- a/modules/wallet/encrypt.go
+++ b/modules/wallet/encrypt.go
@@ -55,7 +55,7 @@ func checkMasterKey(tx *bolt.Tx, masterKey crypto.TwofishKey) error {
 }
 
 // initEncryption initializes and encrypts the primary SeedFile.
-func (w *Wallet) initEncryption(masterKey crypto.TwofishKey, seed modules.Seed) (modules.Seed, error) {
+func (w *Wallet) initEncryption(masterKey crypto.TwofishKey, seed modules.Seed, progress uint64) (modules.Seed, error) {
 	wb := w.dbTx.Bucket(bucketWallet)
 	// Check if the wallet encryption key has already been set.
 	if wb.Get(keyEncryptionVerification) != nil {
@@ -70,7 +70,7 @@ func (w *Wallet) initEncryption(masterKey crypto.TwofishKey, seed modules.Seed) 
 	if err != nil {
 		return modules.Seed{}, err
 	}
-	err = wb.Put(keyPrimarySeedProgress, encoding.Marshal(uint64(0)))
+	err = wb.Put(keyPrimarySeedProgress, encoding.Marshal(progress))
 	if err != nil {
 		return modules.Seed{}, err
 	}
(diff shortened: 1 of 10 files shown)


modules/wallet/encrypt.go
```diff
@@ -362,6 +363,11 @@ func (w *Wallet) InitFromSeed(masterKey crypto.TwofishKey, seed modules.Seed) er
 		masterKey = crypto.TwofishKey(crypto.HashObject(seed))
 	}
 
+	if !w.scanLock.TryLock() {
+		return errScanInProgress
+	}
+	defer w.scanLock.Unlock()
+
 	w.mu.Lock()
 	defer w.mu.Unlock()
```
johnathanhowell: are we still holding this lock for the duration of InitFromSeed?

modules/wallet/encrypt_test.go
```diff
@@ -293,6 +302,45 @@ func TestInitFromSeedParallelUnlock(t *testing.T) {
...
+	// lock the wallet
+	wt.wallet.Lock()
+
+	// spawn an unlock goroutine
+	done := make(chan struct{})
+	go func() {
+		// acquire the write lock so that Unlock acquires the trymutex, but
+		// cannot proceed further
+		wt.wallet.mu.Lock()
+		wt.wallet.Unlock(wt.walletMasterKey)
```
johnathanhowell: shouldn't this error be getting checked? IIRC calling t.Fatal in a goroutine is a race condition so maybe the best would be to change `done` to a `chan error`, pass this error down the channel and check it in the main thread

modules/wallet/encrypt.go
lukechampine: yeah, didn't see much reason not to. It may not be necessary though -- seems reasonable that you should be able to check your balance and such during init. @DavidVorick, do you have any guidance on locking interactions here? Should `scanLock` always be acquired first, or after, or what?

DavidVorick:
I've been thinking about this a lot, I think we really need to update 'siac' and the UI (which means the API too) to report that there's a scan in progress if you request the balance while a scan is in progress. Which means that we should probably have an atomic variable that you set + unset when you do a scan, that the /wallet endpoint (and any other relevant endpoint - send siacoins, etc.) can check when working with the user.

If the user has a scan in progress, they really should be informed that a scan is in progress. I think that's a pretty big UX deal. Being able to also report what % complete the scan is would be great, but I think not as required as simply being able to report that a scan is happening.

lukechampine:
Added `Rescanning` to the API.
Adding a `Progress` field shouldn't be too bad. Need to add a `scanProgress` field to wallet, then update it inside scanning operation. The annoying part is that some scans use the `seedScanner` while others (like `Unlock`) do not. So the easiest way to implement this may be with a subscription wrapper:
```go
type subscriberWithHeight struct {
    blockheight types.BlockHeight
    subscriber  modules.ConsensusSetSubscriber
}

func (s *subscriberWithHeight) ProcessConsensusChange(cc modules.ConsensusChange) {
    s.blockheight += types.BlockHeight(len(cc.AppliedBlocks) - len(cc.RevertedBlocks))
    s.subscriber.ProcessConsensusChange(cc)
}
```

DavidVorick:
LGTM

PR #1797, opened 2017-05-15
profile.go: update function descriptions

starius:
Function names in the documentation started with lower case letters, but functions themselves started with uppercase letters.


lukechampine:
Thanks! `golint` should catch this but I guess this package isn't being linted yet.

PR #1803, opened 2017-05-17
Add ScanHeight to Wallet

lukechampine:
siad no longer prints the current height during unlocking. Instead, it is printed by siac. Currently only `siac wallet unlock` prints the scan height, but this could easily be added to `load seed` and `sweep seed` as well.

Side note: should the api field be `Rescanning` or just `Scanning`? `Rescanning` doesn't make sense if you're unlocking for the first time.

### DIFF
--- a/modules/wallet.go
+++ b/modules/wallet.go
@@ -364,8 +364,8 @@ type (
 		RegisterTransaction(t types.Transaction, parents []types.Transaction) TransactionBuilder
 
 		// Rescanning reports whether the wallet is currently rescanning the
-		// blockchain.
-		Rescanning() bool
+		// blockchain, and if so, the highest blockheight scanned.
+		Rescanning() (bool, types.BlockHeight)
 
 		// StartTransaction is a convenience method that calls
 		// RegisterTransaction(types.Transaction{}, nil)
--- a/modules/wallet/encrypt.go
+++ b/modules/wallet/encrypt.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"sync/atomic"
 
 	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/crypto"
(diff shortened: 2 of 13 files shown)


modules/wallet.go
```diff
@@ -364,8 +364,8 @@ type (
 		RegisterTransaction(t types.Transaction, parents []types.Transaction) TransactionBuilder
 
 		// Rescanning reports whether the wallet is currently rescanning the
-		// blockchain.
-		Rescanning() bool
+		// blockchain, and if so, the highest blockheight scanned.
+		Rescanning() (bool, types.BlockHeight)
```
lukechampine: waffled on whether this should be two separate functions, or maybe `(types.BlockHeight, bool)` to match the "comma ok" idiom.

siac/walletcmd.go
```diff
@@ -457,10 +463,31 @@ func walletunlockcmd() {
...
+			return
+
+		// otherwise, display scan height every 3 seconds
+		case <-time.After(3 * time.Second):
+			var wg api.WalletGET
+			err := getAPI("/wallet", &wg)
+			if err != nil || !wg.Rescanning {
+				continue
+			}
+			fmt.Printf("\rScanned to height %v...", wg.ScanHeight)
```
lukechampine: currently we print the height, but printing a percentage would be easy too. If we want to get fancy there's https://github.com/cheggaaa/pb

modules/wallet.go
DavidVorick: (blockheight, bool) would be my preference, though the name doesn't quite fit in that case. Maybe change to RescanProgress ?

siac/walletcmd.go
DavidVorick: just height is fine for now. cli users especially should not have an issue seeing the exact height.

modules/wallet/encrypt.go
```diff
@@ -122,6 +122,12 @@ func (w *Wallet) managedUnlock(masterKey crypto.TwofishKey) error {
 		// lastChange
 		lastChange = dbGetConsensusChangeID(w.dbTx)
 
+		// currentHeight
+		currentHeight, err = dbGetConsensusHeight(w.dbTx)
```
DavidVorick: :+1:

lukechampine: this is definitely worth testing, since it's the only place where `scanHeight` doesn't start from 0

PR #1804, opened 2017-05-17
dont run entire test suite twice in appveyor.yml

johnathanhowell:


lukechampine:
fortunately this has only been in place for 2 weeks

PR #1806, opened 2017-05-17
add option --trace

starius:


lukechampine:
will this create a trace for the entire lifetime of the program? Because that's a ton of data. In my experience you only want to grab a 5-second window or so, using `/debug/pprof/trace?seconds=5`

DavidVorick:
First traces will be 20 seconds long, then they will increase in length exponentially as the program continues running.

One of the big goals of this trace is to figure out where all the dead OS threads are coming from. Since they trigger nondeterministically after days, trace seemed like the best way to catch the issue.

DavidVorick:
I am using this PR in production, and it seems to be working as expected.

The code looks like you will get overlapping behavior if you run both `--trace` and also `--profile` at the same time. If this is ill-advised, there should be some checker in `siad` that returns an error to a user who sets incompatible flags. Otherwise, the function needs to be updated so that running both at the same time does not result in two loops being created that each log the `ReadMemStats` stuff. Then again, it does log them to two different directories.

I am also thinking that the trace files should just go in the same dir as the other profiles - tracing imo is a type of profiling, but open to disagreement here if you do disagree.

Finally, since we've got two separate flags for tracing and profiling, it might make sense to break cpu and memory profiling into separate flags as well. Currently they both trigger with the same flag because I was being lazy, but in the documentation it states that you can get less accurate/useful results if you run the two together. There are still reasons to run them both together, so I'd like to leave that as an option, but when trying to get a really good understanding of the memory or cpu, it should be run alone.

starius:
> will this create a trace for the entire lifetime of the program? Because that's a ton of data.
> First traces will be 20 seconds long, then they will increase in length exponentially as the program continues running.

Maybe the length should be capped by ~10 minutes to prevent too large trace files. `go tool trace` takes a lot of resources on large inputs.

> Finally, since we've got two separate flags for tracing and profiling, it might make sense to break cpu and memory profiling into separate flags as well.

Does it make sense to make a single option `--profile=cpu,memory,trace` ?

> The code looks like you will get overlapping behavior if you run both --trace and also --profile at the same time.

I enabled them both and saw parts of one in output generated by another. I think we should at least check that at most one of (CPU profiling, trace) is enabled. I think memory info dumping can work together with other two - it is being written only once in a while, so it has no permanent effect.

DavidVorick:
I think --profile=x,y,z sounds good to me. Perhaps it makes sense to force devs to only enable one at a time, it's really the recommended way to do things.

DavidVorick:
Capping trace to 10 or 20 minutes seems like a good idea. Can we have the output go into the `profile` folder though?

DavidVorick:
Update on this?

profile/profile.go
```diff
@@ -81,34 +84,60 @@ func SaveMemProfile(profileDir, identifier string) error {
...
+func StopTrace() {
+	traceLock.Lock()
+	if traceActive {
+		trace.Stop()
+		traceActive = false
+	}
+	traceLock.Unlock()
+}
+
+func startContinuousLog(dir string, restart func()) {
```
DavidVorick: function needs a docstring

siad/daemon.go
```diff
@@ -77,16 +77,32 @@ func processModules(modules string) (string, error) {
 	return modules, nil
 }
 
+// processProfiles checks that the flags given for profiling are valid.
+func processProfile(profile string) (string, error) {
```
DavidVorick: docstring and function have a different name, and a better name for this is probably proceesProfileFlags

DavidVorick:
Two small nits remain, otherwise this is ready to go.

DavidVorick:
Need to run `go fmt`, or in general run `make test-long` before pushing code - `make test-long` will run `go fmt` and so linting and vetting as well, so you can catch the problems before the build runs.

starius:
`go fmt` was run.

By the way, `make test-long` reported no errors because `profile` is skipped in `pkgs` variable in `Makefile`. Maybe we should replace the hardcoded list with `./...` there?

DavidVorick:
The hardcoded list is needed to get the cover command to work I think. Good to know that profile wasn't in the Makefile. I'm strongly in favor of continuing to use the hardcoded list, simply because some of the commands require it, and situations like these bring attention to the fact that we didn't update the hardcoded list.

We can go ahead and update that in a separate PR though.

lukechampine:
I was in favor of `./...` originally, but it has one downside for me, which is that it will try to build all the little helper scripts I keep in my Sia directory.

profile/profile.go
```diff
@@ -81,40 +84,94 @@ func SaveMemProfile(profileDir, identifier string) error {
...
 			// keeps the size of the log small while still providing lots of
 			// information.
-			StopCPUProfile()
-			SaveMemProfile(profileDir, "continuousProfilingMem")
-			StartCPUProfile(profileDir, "continuousProfilingCPU")
+			restart()
 			time.Sleep(sleepTime)
 			sleepTime = time.Duration(1.5 * float64(sleepTime))
-
+			if sleepCap != 0*time.Second && sleepTime > sleepCap {
```
lukechampine: you can compare `sleepCap` to `0` directly, since it's an untyped constant.

starius:
By the way, the list for `cover` can be generated in runtime using `go list ./...`
Helper scripts can be moved to a directory that is filtered out (like `... | grep -v /scripts`)

PR #1808, opened 2017-05-17
fix TestRenterAllowance

johnathanhowell:
This actually fixes a fairly substantial bug in contractor persist (the journal would not be checkpointed after it was first created), in addition to fixing the appveyor failure on TestRenterAllowance


johnathanhowell:
looks like my first pass was successful. Weird that the travis test wasn't failing as well, perhaps yet another example of how windows is more picky when it comes to doing safe IO

johnathanhowell:
I'm writing a regression test for this, don't merge until complete

DavidVorick:
I still really don't see why this is resolving issues? How is this functionally different from skipping the rename? Is it because if we don't use the rename, we can end up with a corrupted header for the journal?

PR #1813, opened 2017-05-19
fix NDFs in host status tests, close host testers

johnathanhowell:
This PR adds more flexibility to the host status tests, fixing an observed NDF. Also properly closes the hosttesters that were previously unclosed.

### DIFF
--- a/modules/host/network_test.go
+++ b/modules/host/network_test.go
@@ -66,37 +66,43 @@ func TestHostWorkingStatus(t *testing.T) {
 		t.SkipNow()
 	}
 	t.Parallel()
+
 	ht, err := newHostTester(t.Name())
 	if err != nil {
 		t.Fatal(err)
 	}
+	defer ht.Close()
 
 	if ht.host.WorkingStatus() != modules.HostWorkingStatusChecking {
 		t.Fatal("expected working state to initially be modules.HostWorkingStatusChecking")
 	}
 
-	// Simulate some setting calls, and see if the host picks up on it.
-	atomic.AddUint64(&ht.host.atomicSettingsCalls, workingStatusThreshold+1)
-	time.Sleep(workingStatusFirstCheck + time.Second)
-	if ht.host.WorkingStatus() != modules.HostWorkingStatusWorking {
-		t.Fatal("expected host working status to be modules.HostWorkingStatusWorking after incrementing status calls")
-	}
+	for i := 0; i < 5; i++ {
+		// Simulate some setting calls, and see if the host picks up on it.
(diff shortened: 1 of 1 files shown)


modules/host/network_test.go
```diff
@@ -66,37 +66,43 @@ func TestHostWorkingStatus(t *testing.T) {
...
+		// Simulate some setting calls, and see if the host picks up on it.
+		atomic.AddUint64(&ht.host.atomicSettingsCalls, workingStatusThreshold+1)
 
-	// No more settings calls, host should believe it is not working now.
-	time.Sleep(workingStatusFrequency + time.Second)
-	if ht.host.WorkingStatus() != modules.HostWorkingStatusNotWorking {
-		t.Fatal("expected host working status to be modules.HostWorkingStatusNotWorking after waiting workingStatusFrequency with no settings calls")
-	}
+		success := false
+		for start := time.Now(); time.Since(start) < 30*time.Second; time.Sleep(time.Millisecond * 10) {
```
DavidVorick: can use build.Retry for this loop pattern

modules/host/network_test.go
```diff
@@ -66,37 +66,43 @@ func TestHostWorkingStatus(t *testing.T) {
...
-	time.Sleep(workingStatusFirstCheck + time.Second)
-	if ht.host.WorkingStatus() != modules.HostWorkingStatusNotWorking {
-		t.Fatal("expected host working status to be modules.HostWorkingStatusWorking after incrementing status calls")
-	}
-	time.Sleep(workingStatusFrequency - workingStatusFirstCheck)
-	if ht.host.WorkingStatus() != modules.HostWorkingStatusWorking {
-		t.Fatal("expected host working status to be modules.HostWorkingStatusWorking after incrementing status calls")
+		// make no settins calls, host should flip back to NotWorking
+		success = false
+		for start := time.Now(); time.Since(start) < 30*time.Second; time.Sleep(time.Millisecond * 10) {
```
DavidVorick: ditto

modules/host/network_test.go
lukechampine: looks like `build.Retry` never made it into master?

lukechampine: ah, `build.Retry` was added in the #1758. I can push that commit to master if you want. Even if we merge #1758, `build.Retry` won't make it into master until we merge the gateway-proto branch.

johnathanhowell: ah, I forgot that we exported that pattern

DavidVorick: Ok, sounds good to me. It's useful 😛

PR #1814, opened 2017-05-19
fix contractor journal sanity check race condition

johnathanhowell:
This PR fixes a rare, nondeterministic race in the renter's contractor where the sanity check would sometimes fail. Instead of reloading the journal from disk and sanity checking against the loading data, the contractor now has a lock-protected `data contractorPersist` field that is continuously updated under lock and used for the sanity check.

### DIFF
--- a/modules/renter/contractor/journal.go
+++ b/modules/renter/contractor/journal.go
@@ -125,28 +125,18 @@ func (j *journal) Close() error {
 // newJournal creates a new journal, using data as the initial object.
 func newJournal(filename string, data contractorPersist) (*journal, error) {
 	// safely create the journal
-	tmp, err := os.Create(filename + "_tmp")
+	f, err := os.Create(filename)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


lukechampine:
don't we *want* it to be reloading from disk though? I thought that was the purpose of the sanity check -- to confirm that the data on disk matches the data in memory.

DavidVorick:
Yeah lol the point was to make sure that the data on disk matched the data in memory.

DavidVorick:
I think the real issue here is that the contractor lock is grabbed before the journal lock, but the journal lock needs to be grabbed first to ensure data correctness

DavidVorick:
Worth noting that the contract manager has only one lock, we might want to replicate that. Actions affecting the log also affect the status of the manager correctness, so you make sure the manager cannot change while the log does and vice versa

johnathanhowell:
Ah, that makes sense. I'll amend this commit to fix the locking patterns instead

PR #1820, opened 2017-05-21
add `/wallet/changekey` API endpoint for change the wallet's encryption key

johnathanhowell:
This PR adds the ability to change the wallet's encryption password, using a new EncryptionManager method `ChangeKey`. It also adds a `/wallet/changekey` api endpoint so API consumers can change their wallet passwords without having to initialize a new wallet.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -257,6 +257,7 @@ func New(requiredUserAgent string, requiredPassword string, cs modules.Consensus
 		router.GET("/wallet/transactions/:addr", api.walletTransactionsAddrHandler)
 		router.GET("/wallet/verify/address/:addr", api.walletVerifyAddressHandler)
 		router.POST("/wallet/unlock", RequirePassword(api.walletUnlockHandler, requiredPassword))
+		router.POST("/wallet/changekey", RequirePassword(api.walletChangeKeyHandler, requiredPassword))
 	}
 
 	// Apply UserAgent middleware and return the API
--- a/api/api.go
+++ b/api/api.go
@@ -257,7 +257,7 @@ func New(requiredUserAgent string, requiredPassword string, cs modules.Consensus
 		router.GET("/wallet/transactions/:addr", api.walletTransactionsAddrHandler)
 		router.GET("/wallet/verify/address/:addr", api.walletVerifyAddressHandler)
 		router.POST("/wallet/unlock", RequirePassword(api.walletUnlockHandler, requiredPassword))
-		router.POST("/wallet/changekey", RequirePassword(api.walletChangeKeyHandler, requiredPassword))
+		router.POST("/wallet/changepassword", RequirePassword(api.walletChangePasswordHandler, requiredPassword))
 	}
 
 	// Apply UserAgent middleware and return the API
--- a/api/wallet.go
+++ b/api/wallet.go
@@ -538,7 +538,7 @@ func (api *API) walletChangeKeyHandler(w http.ResponseWriter, req *http.Request,
 	var newKey crypto.TwofishKey
 
 	originalKeys := encryptionKeys(req.FormValue("encryptionpassword"))
-	if len(potentialOriginalKeys) != 1 {
+	if len(originalKeys) != 1 {
 		WriteError(w, Error{"expected one encryption key passed to encryptionpassword"}, http.StatusBadRequest)
 		return
 	}
--- a/api/wallet.go
+++ b/api/wallet.go
@@ -535,12 +535,12 @@ func (api *API) walletUnlockHandler(w http.ResponseWriter, req *http.Request, _ 
 // walletChangePasswordHandler handles API calls to /wallet/changepassword
 func (api *API) walletChangePasswordHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
 	var newKey crypto.TwofishKey
-	newKeys := encryptionKeys(req.FormValue("newpassword"))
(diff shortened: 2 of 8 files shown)


doc/API.md
```diff
@@ -964,6 +964,7 @@ Wallet
 | [/wallet/transactions/___:addr___](#wallettransactionsaddr-get) | GET       |
 | [/wallet/unlock](#walletunlock-post)                            | POST      |
 | [/wallet/verify/address/:___addr___](#walletverifyaddress-get)  | GET       |
+| [/wallet/changekey](#walletchangekey-post)                      | POST      |
```
DavidVorick: for api, do /wallet/changepassword

api/api.go
```diff
@@ -257,6 +257,7 @@ func New(requiredUserAgent string, requiredPassword string, cs modules.Consensus
 		router.GET("/wallet/transactions/:addr", api.walletTransactionsAddrHandler)
 		router.GET("/wallet/verify/address/:addr", api.walletVerifyAddressHandler)
 		router.POST("/wallet/unlock", RequirePassword(api.walletUnlockHandler, requiredPassword))
+		router.POST("/wallet/changekey", RequirePassword(api.walletChangeKeyHandler, requiredPassword))
```
DavidVorick: same, /wallet/changepassword

johnathanhowell: ACK

api/wallet.go
```diff
@@ -532,6 +532,33 @@ func (api *API) walletUnlockHandler(w http.ResponseWriter, req *http.Request, _
...
+		return
+	}
+	originalKey = originalKeys[0]
+
+	newKeys := encryptionKeys(req.FormValue("newpassword"))
+	if len(newKeys) != 1 {
+		WriteError(w, Error{"expected one encryption key to be passed to newpassword"}, http.StatusBadRequest)
+		return
+	}
+	newKey = newKeys[0]
```
DavidVorick: Technically the api promise here is that you'll check every key in the set of keys. Since this is creating a new key, probably best to return an error if there is more than one key.

johnathanhowell: That's what `if len(newKeys) !=1` does

api/wallet.go
```diff
@@ -532,6 +532,33 @@ func (api *API) walletUnlockHandler(w http.ResponseWriter, req *http.Request, _
...
+func (api *API) walletChangeKeyHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	var originalKey crypto.TwofishKey
+	var newKey crypto.TwofishKey
+
+	originalKeys := encryptionKeys(req.FormValue("encryptionpassword"))
+	if len(originalKeys) != 1 {
+		WriteError(w, Error{"expected one encryption key passed to encryptionpassword"}, http.StatusBadRequest)
+		return
+	}
+	originalKey = originalKeys[0]
```
DavidVorick: promise here is to check every key. If a key does not work, try the next one. I forget why we had to do it this way, but I think it was related to not knowing whether a provided key is a nmemonic or just a string.

DavidVorick: regardless I don't think avoidable that you should try will all keys, one by one, until one does not return an error.

johnathanhowell: ah, ok. I'll update this

api/wallet.go
DavidVorick: just kidding, here I think it's fine to just use the first key.

api/wallet.go
```diff
@@ -532,6 +532,33 @@ func (api *API) walletUnlockHandler(w http.ResponseWriter, req *http.Request, _
 	WriteError(w, Error{"error when calling /wallet/unlock: " + modules.ErrBadEncryptionKey.Error()}, http.StatusBadRequest)
 }
 
+// walletChangeKeyHandler handles API calls to /wallet/changekey.
+func (api *API) walletChangeKeyHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	var originalKey crypto.TwofishKey
+	var newKey crypto.TwofishKey
+
+	originalKeys := encryptionKeys(req.FormValue("encryptionpassword"))
+	if len(originalKeys) != 1 {
```
DavidVorick: if the original password was a mnemonic you will definitely get at least 2 results, need to try them both instead of returning an error.

johnathanhowell: Yeah, I just pushed a commit that does the same thing as /wallet/unlock (iterate through each key, break if one succeeds)

modules/wallet/encrypt.go
```diff
@@ -413,6 +413,151 @@ func (w *Wallet) Lock() error {
...
+func (w *Wallet) ChangeKey(masterKey crypto.TwofishKey, newKey crypto.TwofishKey) error {
+	if err := w.tg.Add(); err != nil {
+		return err
+	}
+	defer w.tg.Done()
+
+	if !w.scanLock.TryLock() {
+		return errScanInProgress
+	}
+	defer w.scanLock.Unlock()
```
lukechampine: Is this necessary?

I'm also wondering if perhaps we should require the wallet to be locked before changing the password. I don't have a clear justification for this, only that it could automatically rule out a certain class of bugs.

johnathanhowell: I had the same thought while writing this PR, but decided against it for UX reasons.

johnathanhowell: Also, I dont think the trylock is necessary. I'll remove it

PR #1823, opened 2017-05-22
add changepassword support to siac

DavidVorick:


siac/walletcmd.go
```diff
@@ -162,6 +169,9 @@ will be sent to your wallet.`,
 
 const askPasswordText = "We need to encrypt the new data using the current wallet password, please provide: "
 
+const newPasswordText = "Please provide the current password of the wallet: "
+const oldPasswordText = "Please provide the new password for the wallet: "
+
```
lukechampine: Surely "Current password" and "New password" are clear enough? Users are familiar with password rotation.

`askPasswordText` is a bit long-winded too, although that's a less familiar situation.

johnathanhowell: agreed, `Current password` and `New password` are definitely enough here.

johnathanhowell: Also, the wording on these is flipped.

johnathanhowell:
latest commits fixup the prompt messages (oldpassword and newpassword had swapped semantic meanings), fixes an unintended branch in `walletChangePasswordHandler`, and improves TestWalletChangePasswordDeep, verifying that passwords with spaces work correctly.

PR #1824, opened 2017-05-22
consensus: download and process blocks in parallel

starius:
Previous implementation did only one thing at a time: download blocks | process blocks. The new one downloads up to 1000 blocks in advance in order to reduce chances of blocking on network.


modules/consensus/synchronize.go
```diff
@@ -175,39 +176,72 @@ func (cs *ConsensusSet) managedReceiveBlocks(conn modules.PeerConn) (returnErr e
...
+			// Send blocks over the channel.
+			blocksChan <- newBlocks
+		}
+	}()
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		func() {
+			for newBlocks := range blocksChan {
```
johnathanhowell: this needs to `select{}` on the blocksChan and `cs.tg.StopChan()`, in order for clean shutdown not to be blocked.

starius: Done. I added new changes as separate commits temporarily. I'll squash them afterwards.

Now it fails with the following error:
```
synchronize_ibd_test.go:71: could not solve block using limited hashing power
```

starius: Ah, I was testing with `go test` but Travis tests with `go test -v -race -tags='testing debug' -timeout=500s`. The later command succeeds here as well. So the "limited hashing power" issue is unrelated.

starius:
Squashed and rebased.

modules/consensus/synchronize.go
```diff
@@ -175,39 +176,82 @@ func (cs *ConsensusSet) managedReceiveBlocks(conn modules.PeerConn) (returnErr e
...
-			if acceptErr == modules.ErrNonExtendingBlock || acceptErr == modules.ErrBlockKnown {
-				acceptErr = nil
+			if err := encoding.ReadObject(conn, &moreAvailable, 1); err != nil {
+				downloadErr = err
+				return
 			}
-			if acceptErr != nil {
-				return acceptErr
+			// Send blocks over the channel.
+			blocksChan <- newBlocks
```
johnathanhowell: I think you may need to `select{}` for StopChan() here as well, to cover the (unlikely) case where `blocksChan` is full

starius: If `blocksChan` is full, then when `cs.tg.StopChan` is triggered, the precessing goroutine calls `atomic.StoreInt32(&processingFailed, 1)`, returns from the internal function and then reads all remaining items from the channel, so the downloading goroutine (that was blocked in `blocksChan <- newBlocks`) is unblocked, goes to the next iteration, detects that `processingFailed == 1` and quits.

modules/consensus/synchronize.go
```diff
@@ -175,39 +176,82 @@ func (cs *ConsensusSet) managedReceiveBlocks(conn modules.PeerConn) (returnErr e
...
-	for moreAvailable {
-		// Read a slice of blocks from the wire.
-		var newBlocks []types.Block
-		if err := encoding.ReadObject(conn, &newBlocks, uint64(MaxCatchUpBlocks)*types.BlockSizeLimit); err != nil {
-			return err
-		}
-		if err := encoding.ReadObject(conn, &moreAvailable, 1); err != nil {
-			return err
-		}
+	blocksChan := make(chan []types.Block, 1000)
```
DavidVorick: Max block size is 2mb, queuing 1000 blocks is not good. I think at best we could queue 100, but even that makes me uncomfortable. If you want to get fancier, we could track the size and then limit it to 50mb or somethin. And we could also add a settings endpoint to the consensus api to make the 50mb number configurable.

modules/consensus/synchronize.go
```diff
@@ -175,39 +176,82 @@ func (cs *ConsensusSet) managedReceiveBlocks(conn modules.PeerConn) (returnErr e
...
+					if !has {
+						return
+					}
+				}
+				// Integrate the blocks into the consensus set.
+				for _, block := range newBlocks {
+					stalled = false
+					// Call managedAcceptBlock instead of AcceptBlock so as not to broadcast
+					// every block.
+					acceptErr = cs.managedAcceptBlock(block)
```
DavidVorick: for the record, on HDDs there is basically a constant penalty of between 100ms and 250ms for every time we call  `managedAcceptBlock` because that's just the cost of the `bolt.Update` call that happens. We could make this a lot better by having it accept multiple blocks. Bolt is slow, some some of the larger blocks probably take more than 100ms anyway, but some of the smaller, earlier blocks probably only take like 10 or 20ms to process all by themselves, to be able to strip 100ms per block from those blocks would speed up IBD a lot.

DavidVorick: This is of course a completely different pr, I'm just pointing out places where speed can really be improved.

DavidVorick:
Overall I like this change, though I need to think more about how our rotating code (we rotate peers every 5 minutes or something) affects these changes.

Have you done a full ibd on an hdd to test the speedup? I'm guessing (b/c you did not address the primary bottleneck on hdd) that the speedup is a lot more impressive on an ssd and on a ramdisk than on an hdd. But more I want to make sure that the peer rotation code doesn't get stuck during ibd with these changes.

DavidVorick:
I think that the peer rotation code should be okay, actually.

starius:
> b/c you did not address the primary bottleneck on hdd

I did it but kept it for another PR.

> Have you done a full ibd on an hdd to test the speedup?

I did but can't find it now :-)
I'll redo it.

starius:
I tested this PR in GCP machines with 15G RAM and 4 CPU, located in us-west1-b. This PR has no or nagative effect of IBD time there:

hdd-before
real    621m3.500s
user    79m21.116s
sys     6m20.388s

hdd-after
real    627m11.598s
user    68m6.336s
sys     5m38.028s


ssd-before
real    232m40.071s
user    64m35.192s
sys     4m39.252s

ssh-after
real    242m59.424s
user    64m45.752s
sys     4m58.132s

We should test it on machines with worse network connection to see the benefit of the PR.

PR #1831, opened 2017-05-23
fix host contract count increasing on restart

johnathanhowell:
This fixes the bug where the host's `FinancialMetrics.ContractCount` would incorrectly increase after each restart. Diff and commits should be easy to follow.


modules/host/persist.go
```diff
@@ -172,26 +171,6 @@ func (h *Host) load() error {
 		return err
 	}
 
-	// Get the contract count by observing all of the incomplete storage
-	// obligations in the database.
-	err = h.db.View(func(tx *bolt.Tx) error {
```
johnathanhowell: we could also set `h.financialMetrics.ContractCount = 0` here instead of removing this database scan, not sure which is more correct.

lukechampine:
@DavidVorick, is anything blocking this from merge? LGTM

modules/host/persist.go
DavidVorick: The database scan was added because other accounting in the host is incorrect. I'd much prefer to keep this, and set h.financialMetrics.ContractCount to 0 at start.

The scan I think we'll be adding other features to. The whole host is a mess right now and I think we need to inject some sanity. Recalculating values at startup is going to be more reliable than keeping some state around and hoping it never corrupts.

At least temporarily anyway.

johnathanhowell:
Updated per @DavidVorick's suggestions. The database scan is still performed, but ContractCount is cleared before it happens.

api/hostdb_test.go
```diff
@@ -617,21 +618,25 @@ func TestHostDBAndRenterDownloadDynamicIPs(t *testing.T) {
...
+				return err
+			}
+			// Try downloading the file.
+			// Check that the download has the right contents.
+			download, err = ioutil.ReadFile(downpath)
+			if err != nil {
+				return err
+			}
+			if bytes.Compare(orig, download) != 0 {
+				return err
```
lukechampine: `err` could be nil here; need to return a new error. Also, use `bytes.Equal`

PR #1834, opened 2017-05-25
fix contractor.IsOffline

johnathanhowell:
Previously, the contractor's `IsOffline` method would report that contracts that did not exist in the contractor database were online.


lukechampine:
I think the rationale here was to avoid punishing a host if we don't actually have any records. It made a little more sense when (IIRC) `IsOffline` took a host pubkey as an argument instead of a contract ID. But yeah, if the contract doesn't exist, something's definitely wrong.

DavidVorick:
We've also added a few other protections related to resetting the hostdb

lukechampine:
LGTM

PR #1842, opened 2017-05-26
Pass -s and -w ldflags to reduce binary size.

moimael:
Those flags remove DWARF table used by go debugger in release mode.

**Before:**
siad: **16.5MB**
siac: **9.6MB**

**After:**
siad: **11.1MB**
siac: **6.5MB**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,7 +64,7 @@ release:
 release-race:
 	go install -race -tags='debug profile' $(pkgs)
 release-std:
-	go install $(pkgs)
+	go install -ldflags='-s -w' $(pkgs)
 
 # clean removes all directories that get automatically created during
 # development.
--- a/release.sh
+++ b/release.sh
@@ -36,7 +36,7 @@ for os in darwin linux windows; do
 		if [ "$os" == "windows" ]; then
 			bin=${pkg}.exe
 		fi
-		GOOS=${os} go build -o $folder/$bin ./$pkg
+		GOOS=${os} go build -ldflags="-s -w" -o $folder/$bin ./$pkg
 		openssl dgst -sha256 -sign $keyfile -out $folder/${bin}.sig $folder/$bin
 		# verify signature
 		if [[ -n $pubkeyfile ]]; then


DavidVorick:
Rejecting this for now. Sia is still rather unstable and we need those crash reports. We also do some live debugging, and I'm sure that stripping the binaries would make life harder for us when new bugs are discovered in the wild.

Worth considering when things are more stable, but our binaries are already less than 50MB together. For me, that's plenty small. (UI downloads could certainly still be improved)

moimael:
But it doesn't strip stacktraces it's only useful if you use the debugger in production release which is not the case. See here for more details :https://blog.filippo.io/shrink-your-go-binaries-with-this-one-weird-trick/

DavidVorick:
ok. Waiting for @lukechampine to comment (he's better with go tool stuff than I am)

lukechampine:
I think this is probably safe, although I would like to do some more research before approving. Also, I don't think binary size is a huge deal; the precompiled downloads are already compressed, so there's no gain there. @moimael, do you have a specific use case in mind?

DavidVorick:
re-opening as it seems this may be okay to merge after all.

lukechampine:
LGTM. People who want to debug the binary can build it themselves.

DavidVorick:
I'm still not confident that it's the right choice. What functionality exactly is lost by removing these? And if it's not expected to be necessary, why is it default? Will gdb still work? Trace? `--profile`?

I guess I just really don't understand what the data is doing that we've stripped out.

moimael:
Basically you cannot attach a debugger to release binary. You still get stacktraces in case of errors. It removes debug symbols (only for release, not in dev).

DavidVorick:
I have twice attached a debugger to a non-debug binary, but that's actually not that many times, for all the problems I've worked through in production. Maybe this is okay as a change then.

I am just nervous about that one time that you really wish that you had it.

moimael:
Maybe what's needed here is a proper error reporting solution like sentry or mixpanel ?

PR #1843, opened 2017-05-26
Add support for getting and sending raw transactions

DavidVorick:
Now you can grab the raw transaction associated with an id, and you can re-broadcast it as well. This will help exchanges that have started struggling with the mempool.


api/transactionpool.go
```diff
@@ -3,17 +3,80 @@ package api
...
+	}
+
+	txn, parents, exists := api.tpool.Transaction(txid)
+	if !exists {
+		WriteError(w, Error{"transaction not found in transaction pool"}, http.StatusBadRequest)
+		return
+	}
+	WriteJSON(w, TpoolRawGET{
+		ID:          txid,
+		Parents:     encoding.Marshal(parents),
```
DavidVorick: parents may be 'ni', would that cause a problem here?

modules/transactionpool.go
```diff
@@ -99,9 +99,17 @@ type TransactionPool interface {
...
+	// corresponding to the provided transaction id.
+	Transaction(id types.TransactionID) (txn types.Transaction, unconfirmedParents []types.Transaction, exists bool)
+
 	// Unsubscribe removes a subscriber from the transaction pool.
 	// This is necessary for clean shutdown of the miner.
 	Unsubscribe(TransactionPoolSubscriber)
+
+	// Broadcast broadcasts a transaction set to all of the transaction pool's
+	// peers.
+	Broadcast(ts []types.Transaction)
```
DavidVorick: alphabetically sorted, please

api/transactionpool.go
```diff
@@ -3,17 +3,80 @@ package api
...
+		WriteError(w, Error{"error decoding transaction:" + err.Error()}, http.StatusBadRequest)
+		return
+	}
+
+	txnSet := append(parents, txn)
+
+	api.tpool.Broadcast(txnSet)
+	err = api.tpool.AcceptTransactionSet(txnSet)
+	if err != nil {
+		WriteError(w, Error{"error accepting transaction set:" + err.Error()}, http.StatusBadRequest)
```
DavidVorick: This error needs to ignore the 'duplicate' error.

modules/transactionpool/accept.go
```diff
@@ -373,5 +373,7 @@ func (tp *TransactionPool) relayTransactionSet(conn modules.PeerConn) error {
 	if err != nil {
 		return err
 	}
+	// TODO: Should probably be the 'accept' one. ('Accept' will
+	// auto-broadcast, but this should not broadcast bad ones or duplicates.
```
DavidVorick: can remove the todo, replaced by separate broadcast method

modules/transactionpool/transactionpool.go
```diff
@@ -151,3 +153,80 @@ func (tp *TransactionPool) TransactionList() []types.Transaction {
...
+	for _, tSet := range tp.transactionSets {
+		for _, t := range tSet {
+			if t.ID() == id {
+				txn = t
+				exists = true
+			}
+		}
+	}
+
+	// collect parent IDs
```
DavidVorick: not good enough, also not needed. You don't just want the parents, you want all the grandparents, etc. as well. But you already have all those, they are cluseterd in to transactionSets. So all you really need to do with a transaction is use the transaction set as the set of parents. The transaction set is ordered correctly, so if the txn is not the last item of the set, you can just use the elements prior.

There's an edge case where you may end up with unneeded parents. So to be completely correct you have to loop through the parents in reverse order and check if they are in fact necessary. And if so, you expand the set of relevant outputs, and continue the loop. Overall, it should be pretty fast to do that, it should be a linear time operation.

DavidVorick:
still needs:

+ api docs
+ api test
+ test for the `Broadcast` function (probably both in api and in tpool)
+ fix the way you select parents

I have a validation test in mind but this PR isn't ready for it yet.

modules/transactionpool/transactionpool.go
johnathanhowell: What determines if a parent is necessary?

api/transactionpool.go
johnathanhowell: No, I manually tested and verified this behavior with the encoding package

johnathanhowell:
Updated with review comments.

api/transactionpool_test.go
```diff
@@ -0,0 +1,120 @@
...
+		t.Fatal(err)
+	}
+	postValues := url.Values{}
+	postValues.Set("parents", string(trg.Parents))
+	postValues.Set("transaction", string(trg.Transaction))
+	err = st.stdPostAPI("/tpool/raw", postValues)
+	if err != nil {
+		t.Fatal(err)
+	}
+}
```
DavidVorick: Test doesn't verify that the broadcast bit actually re-broadcasts.

doc/API.md
```diff
@@ -1354,3 +1355,43 @@ newpassword
...
+
+returns the ID for the requested transaction and its raw encoded parents and transaction data.
+
+###### JSON Response [(with comments)](/doc/api/Transactionpool.md#json-response)
+```javascript
+{
+	// id of the transaction
+	"id": "1234567890abcdef0123456789abcdef0123456789abcdef0123456789abcdef",
+	// raw transaction data
+	"transaction": "AAAAAQID",
```
DavidVorick: This is not enough bytes to be a real transaction

johnathanhowell: Yeah, I will add some more

doc/api/Transactionpool.md
```diff
@@ -0,0 +1,56 @@
...
+
+returns the ID for the requested transaction and its raw encoded parents and transaction data.
+
+###### JSON Response [(with comments)](/doc/api/Transactionpool.md#json-response)
+```javascript
+{
+	// id of the transaction
+	"id": "1234567890abcdef0123456789abcdef0123456789abcdef0123456789abcdef",
+	// raw transaction data
+	"transaction": "AAAAAQID",
```
DavidVorick: should have real results here

doc/api/Transactionpool.md
```diff
@@ -0,0 +1,56 @@
...
+	"parents": "TWFuIGlzIGRpc3Rpbmd1aXNoZWQsIG5vdCBvbmx5IGJ5IGhpcyByZWFzb24sIGJ1dCBieSB0aGlz",
```
DavidVorick: probably okay to have this one return a 'nil' result in the example

DavidVorick: docs should probably indicate that 'parents' will sometimes be nil.

doc/API.md
```diff
@@ -1354,3 +1355,43 @@ newpassword
...
+
+#### /tpool/raw [POST]
+
+submits a raw transaction to the transaction pool, broadcasting it to the transaction pool's peers.
+
+###### Query String Parameters [(with comments)](/doc/api/Transactionpool.md#query-string-parameters)
+
+```
+parents string // raw encoded transaction parents
+transaction string // raw encoded transaction
```
DavidVorick: line up comments please

lukechampine:
Why are we using  base64 encoding for this? Wouldn't JSON work just fine?

johnathanhowell:
The original spec called for returning Transactions and Parents as a marshalled byte slice, which the `encoding/json` package automatically encodes as base64. We could return the transaction set as a `[]types.Transaction, types.Transaction` encoded to JSON instead, which seems more correct to me. @DavidVorick can probably expand on why a `[]byte` was chosen instead of a `[]types.Transaction`

lukechampine:
We definitely want the JSON as a []byte. Using Sia's internal encoding makes interop a huge headache. It would mean that api consumers would have to reimplement the encoding package in order to construct transactions.

AFAIK the only endpoint that exposes the encoding package is the miner one, but that's a special case because it's high performance and the manipulations are simple (just overwriting bytes in-place)

DavidVorick:
@lukechampine that's why it's called the raw endpoint. You aren't supposed to do anything with the data besides get and post, and you aren't supposed to manipulate it in between.

Giving users access to the actual transactions is a different endpoint.

api/transactionpool.go
```diff
@@ -1,19 +1,93 @@
...
+	TpoolRawGET struct {
+		ID          types.TransactionID `json:"id"`
+		Parents     string              `json:"parents"`
+		Transaction string              `json:"transaction"`
+	}
 )
 
-type TransactionPoolGET struct {
-	Transactions []types.Transaction `json:"transactions"`
+func decodeTransactionID(txidStr string) (types.TransactionID, error) {
```
DavidVorick: docstring on all functions please

DavidVorick:
I thought about the raw vs. json for a while longer, but I still feel convinced that raw is the right way to present this. The purpose of this endpoint is so that you can fetch and post transactions, not so you can see inside of the transaction pool. Seeing inside of the transaction pool would likely involve a completely different structure for the api.

lukechampine:
Why would you want to have a separate get and post for transactions if they're opaque? Is this solely to support resubmitting existing txns? If so, why not just have a /resubmit route?

I don't think "raw" necessarily implies opaque. Why not use JSON? What is the downside?

PR #1847, opened 2017-05-28
Improvements to Transaction Pool (tpool) Module

willmadison:
This is just a placeholder for our conversation about work to be done to improve Sia's transaction pool such that it can safely support larger pools and ultimately be configurable by the user (long term the goal is to improve the minimum fee algorithm, add support for replace by fee and cP4P, and wallet based fee bumping all of which is dependent on an improved tpool).


lukechampine:
Since there are multiple features we want to add, let's move this to an issue (or perhaps a project @DavidVorick) rather than a pull request. We like to keep our list of PRs small. Each feature will likely warrant its own PR.

DavidVorick:
Sorry, we did not talk about this. @lukechampine I've directed people looking at contributing to open pull requests directly, because issues lately have been high in volume. Every pull request represents work in progress, and will give everyone a short list of what everyone else is working on.

I'm expecting to have 3-5 new contributors all working on new things, and this was my choice for product management.

Sorry that we did not discuss it first.

DavidVorick:
For the formatting, we like to keep it 

stdlib

sialib

otherlib

Which goimports seems to not prefer. The file that got changed has them in the wrong order anyway, but the new changes do not match our other files

lukechampine:
@DavidVorick I believe goimports does not actively split your imports except to separate stdlib from non-stdlib. If you manually insert extra line breaks, goimports will respect them.

PR #1852, opened 2017-05-29
fix TestWalletRescanning NDF

johnathanhowell:
This PR addresses an observed NDF in TestWalletRescanning by mining more blocks and sleeping less.


DavidVorick:
ouch. One of the issues here is that with our current set of DEBUG flags, adding blocks is quadratic but rescanning is only linear.

We might be able to fix that. I'm always nervous to disable sanity checks, but perhaps we can make the quadratic one a probabilistic check.

PR #1853, opened 2017-05-29
Update progress cap from 99 to 100

VoidingWarranties:
Why? Otherwise from 99% to 100% the progress siac reports will always be printed as 99.0% even if progress is being made.


johnathanhowell:
Is this the preferred way of display progress now? Previously the reasoning was to avoid giving users the impression that the software was stalled by capping progress at 99% (it doesn't really make much sense for a long-running process to be at 100% to a user, it should be complete!).

The UI is smarter about this case, when it reaches <2% progress it switches the progress bar over to a generic 'Synchronizing' visualization, but it still caps progress at 99%. Should that be removed?

Issue #1855, opened 2017-05-30
Renter tries to upload directory node


alexf91:
I have a directory with a few files in it:
```
foodir
├── bardir
│     └── zoofile
├── barfile1
├── barfile2
└── barfile3
```

If I upload this directory using `siac renter upload foodir foodir`, not only the files, but also the directory nodes are uploaded. The upload of the nodes never finishes.

```
siac renter list                
Tracking 6 files:
   4.1 KB  foodir (uploading, 0.00%)
   4.1 KB  foodir/bardir (uploading, 0.00%)
      0 B  foodir/bardir/zoofile
      0 B  foodir/barfile1
      0 B  foodir/barfile2
      0 B  foodir/barfile3
```

Related to #913 and #1334. Recursive upload works, but inodes are not ignored.

lukechampine:
This should be a simple fix. I believe we just need to check for `info.IsDir` when walking the directory.

However, it also seems like the backend is not handling directories properly either. It should return an error if you attempt to upload one.

johnathanhowell:
Fixed in #1856

Issue #1334, opened 2016-07-03
Unexpected directory upload behavior


mnsl:
When a user attempts to upload a directory, they expect to upload the contents of that directory.  Therefore siac should warn that recursive directory upload is not yet implemented for siac (see #913), instead of uploading the directory itself, which is just a list of inodes and their assigned names.

We should implement recursive directory upload ASAP, but in the meantime we should at least warn the user that it is not yet implemented.

DavidVorick:
it's probably easy enough to do recursive directory uploads that we don't need to warn the user.

lukechampine:
fixed in https://github.com/NebulousLabs/Sia/pull/1677

Issue #913, opened 2016-01-13
siac recursive directory upload


DavidVorick:
siac should have a command that will upload an entire directory

lukechampine:
hmm. I suppose uploading a folder is common enough behavior that we could expect most frontends to support it, hence the justification for doing it on the backend.
We could support this without changing the API, just detect if the source path is a folder and branch on that.
There are a few annoying edge cases here that we would need to manage carefully. I can think of three off the bat: 1) dotfiles, 2) empty folders, 3) directory loops (via symlink). We'd probably include dotfiles by default and add a flag for excluding them. Can't do much about empty folders for now. At least Sia-UI can fake an empty folder, but if its offloading the word to the backend, it won't know which folders to fake. And lastly, we'd need to ensure our `filepath.Walk` detects loops somehow.

VoidingWarranties:
I think symlinks should be ignored by default, with an option to follow them potentially added as a feature in the future. This is `cp`'s and other filesystem utils default behavior.

DavidVorick:
I'm pretty sure filepath.Walk avoids symlinks automatically as well. It's default behavior is probably sane enough that it'll do all the edge cases we need properly.

DavidVorick:
would like to see this following the shortlist for 1.0. It might have to wait, but we should have this done in 1.1 if it's not done by 1.0.

VoidingWarranties:
@johnathanhowell and I were talking about this. Specifically about the expected behavior when you do not have enough space to upload the entire directory. It would be helpful to have a check at the beginning to see if there is enough space to upload the entire directory, and to fail if there is not enough space. Otherwise it would not be obvious to the user which files were successfully uploaded, especially if the directory has a high branching factor or deep branches.

DavidVorick:
Seems reasonable to scan the whole directory before doing any uploading.

lukechampine:
added in https://github.com/NebulousLabs/Sia/pull/1677

PR #1856, opened 2017-05-30
do not upload inodes in renter

johnathanhowell:
See #1855, #913, #1334. `renter.Upload()` now disallows uploading directories, and `siac renter upload` will skip over directories in its Walk.

### DIFF
--- a/modules/renter/upload.go
+++ b/modules/renter/upload.go
@@ -72,13 +72,7 @@ func validateSiapath(siapath string) error {
 // validateSource verifies that a sourcePath meets the
 // requirements for upload.
 func validateSource(sourcePath string) error {
-	f, err := os.Open(sourcePath)
-	if err != nil {
-		return err
-	}
-	defer f.Close()
-
-	finfo, err := f.Stat()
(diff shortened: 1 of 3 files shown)


modules/renter/upload.go
```diff
@@ -68,6 +69,26 @@ func validateSiapath(siapath string) error {
...
+// validateSource verifies that a sourcePath meets the
+// requirements for upload.
+func validateSource(sourcePath string) error {
+	f, err := os.Open(sourcePath)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+
+	finfo, err := f.Stat()
```
lukechampine: Can just use `os.Stat` instead of opening and closing the file.

lukechampine: `validateSource` is also a bit generic for this namespace. Could we inline it?

lukechampine: Oops, thought this was in the API package. The name is fine for the renter package

johnathanhowell:
Re: "inode", "empty inode" is the correct term, but directory is clearer. I'll update the wording.

PR #1859, opened 2017-05-31
add logging to the wallet

DavidVorick:
Now there are logging statements any time that a wallet-relevant transaction is added to consensus, or reverted from consensus.

This PR is partially to help Poloniex, so there is a log of transactions that may or may not have confirmed.

### DIFF
--- a/types/encoding.go
+++ b/types/encoding.go
@@ -115,6 +115,35 @@ func (c *Currency) UnmarshalSia(r io.Reader) error {
 	return nil
 }
 
+// HumanString prints the Currency using human readable units. The unit used
+// will be the largest unit that results in a value greater than 1. The value is
+// rounded to 4 significant digits.
+func (c Currency) HumanString() string {
+	pico := SiacoinPrecision.Div64(1e12)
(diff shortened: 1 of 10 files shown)


types/encoding.go
```diff
@@ -115,6 +115,35 @@ func (c *Currency) UnmarshalSia(r io.Reader) error {
 	return nil
 }
 
+// HumanString prints the Currency using human readable units. The unit used
+// will be the largest unit that results in a value greater than 1. The value is
+// rounded to 4 significant digits.
+func (c Currency) HumanString() string {
```
johnathanhowell: This should probably have unit tests to verify its correctness given various currency amounts (just copy the `TestCurrencyUnits` test from siac)

Siac should probably be updated to use this method also

DavidVorick: added tests. Decided to leave `siac` as it is - displaying numbers through siac and through logs do not naturally require the same formatting. While they use the same formatting right now, we may in the future decide that different levels of precision, etc. are required for each situation.

In light of that, I'm going to leave them separate.

PR #1865, opened 2017-06-01
try reverted transactions in tpool Update()

johnathanhowell:
previously, transactions that were reverted did not get retried. Now every reverted transaction will be tried, oldest to newest, and if `acceptTransactionSet` succeeds, the reverted transaction will be added back to the pool.

### DIFF
--- a/modules/transactionpool/update.go
+++ b/modules/transactionpool/update.go
@@ -105,7 +105,9 @@ func (tp *TransactionPool) ProcessConsensusChange(cc modules.ConsensusChange) {
 	// processing consensus changes. Overall, the locking is pretty fragile and
 	// more rules need to be put in place.
 	for _, set := range unconfirmedSets {
-		tp.acceptTransactionSet(set, cc.TryTransactionSet) // Error is not checked.
+		for _, txn := range set {
+			tp.acceptTransactionSet([]types.Transaction{txn}, cc.TryTransactionSet) // Error is not checked.
+		}
 	}
 
 	// Inform subscribers that an update has executed.
--- a/modules/transactionpool/update.go
+++ b/modules/transactionpool/update.go
@@ -52,20 +52,21 @@ func (tp *TransactionPool) ProcessConsensusChange(cc modules.ConsensusChange) {
 		}
 	}
 
-	// TODO: Right now, transactions that were reverted to not get saved and
-	// retried, because some transactions such as storage proofs might be
-	// illegal, and there's no good way to preserve dependencies when illegal
-	// transactions are suddenly involved.
-	//
-	// One potential solution is to have modules manually do resubmission if
-	// something goes wrong. Another is to have the transaction pool remember
-	// recent transaction sets on the off chance that they become valid again
-	// due to a reorg.
-	//
-	// Another option is to scan through the blocks transactions one at a time
-	// check if they are valid. If so, lump them in a set with the next guy.
-	// When they stop being valid, you've found a guy to throw away. It's n^2
-	// in the number of transactions in the block.
+	// Scan through the reverted blocks and try to add each transaction in the
+	// block to the transactionSet. If the add succeeds, add the transaction back
+	// to the transaction pool.
+	for _, block := range cc.RevertedBlocks {
+		for _, txn := range block.Transactions {
+			err = tp.acceptTransactionSet([]types.Transaction{txn}, cc.TryTransactionSet)
+			if err == nil {
+				txids[txn.ID()] = struct{}{}
+				err = tp.addTransaction(tp.dbTx, txn.ID())
+				if err != nil {
+					tp.log.Println("ERROR: could not add a transaction:", err)
+				}
+			}
+		}
+	}
 
 	// Save all of the current unconfirmed transaction sets into a list.
 	var unconfirmedSets [][]types.Transaction
--- a/modules/transactionpool/update.go
+++ b/modules/transactionpool/update.go
@@ -52,22 +52,6 @@ func (tp *TransactionPool) ProcessConsensusChange(cc modules.ConsensusChange) {
 		}
 	}
 
-	// Scan through the reverted blocks and try to add each transaction in the
-	// block to the transactionSet. If the add succeeds, add the transaction back
-	// to the transaction pool.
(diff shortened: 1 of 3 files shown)


modules/transactionpool/update_test.go
```diff
@@ -40,3 +41,94 @@ func TestArbDataOnly(t *testing.T) {
...
+	success := false
+	for start := time.Now(); time.Since(start) < time.Minute; time.Sleep(time.Millisecond * 100) {
+		if tpt.cs.CurrentBlock().ID() == tpt2.cs.CurrentBlock().ID() {
+			success = true
+			break
+		}
+	}
+	if !success {
+		t.Fatal("testers did not have the same block height after one minute")
+	}
```
DavidVorick: This is pretty clean on it's own but standard way to do it is build.Retry

DavidVorick:
Test ran too long on appveyor. At some point we should look into consolidating some of the longer tests. For now I guess switching to 15 minutes would be enough

DavidVorick:
I realized that this code is not actually correct. The tranasactions in the existing tpool and the sets in the existing tpool may depend on the transactions that got reverted from the blocks. That dependency is why we call `purge()` before figuring out which sets are still valid in the tpool.

You have to call `purge()` first, before you try to add the reverted transactions back into the tpool. I've gone ahead and fixed this up.

modules/transactionpool/update.go
```diff
@@ -52,20 +52,21 @@ func (tp *TransactionPool) ProcessConsensusChange(cc modules.ConsensusChange) {
...
-	// in the number of transactions in the block.
+	// Scan through the reverted blocks and try to add each transaction in the
+	// block to the transactionSet. If the add succeeds, add the transaction back
+	// to the transaction pool.
+	for _, block := range cc.RevertedBlocks {
+		for _, txn := range block.Transactions {
+			err = tp.acceptTransactionSet([]types.Transaction{txn}, cc.TryTransactionSet)
+			if err == nil {
+				txids[txn.ID()] = struct{}{}
+				err = tp.addTransaction(tp.dbTx, txn.ID())
```
DavidVorick: `addTransaction` is only for transactions that have been confirmed on-chain.

modules/transactionpool/update.go
```diff
@@ -52,20 +52,21 @@ func (tp *TransactionPool) ProcessConsensusChange(cc modules.ConsensusChange) {
...
-	// When they stop being valid, you've found a guy to throw away. It's n^2
-	// in the number of transactions in the block.
+	// Scan through the reverted blocks and try to add each transaction in the
+	// block to the transactionSet. If the add succeeds, add the transaction back
+	// to the transaction pool.
+	for _, block := range cc.RevertedBlocks {
+		for _, txn := range block.Transactions {
+			err = tp.acceptTransactionSet([]types.Transaction{txn}, cc.TryTransactionSet)
+			if err == nil {
+				txids[txn.ID()] = struct{}{}
```
DavidVorick: txids are only txids of transactions that have been confirmed on-chain

modules/transactionpool/update.go
```diff
@@ -52,20 +52,21 @@ func (tp *TransactionPool) ProcessConsensusChange(cc modules.ConsensusChange) {
...
-	// Another option is to scan through the blocks transactions one at a time
-	// check if they are valid. If so, lump them in a set with the next guy.
-	// When they stop being valid, you've found a guy to throw away. It's n^2
-	// in the number of transactions in the block.
+	// Scan through the reverted blocks and try to add each transaction in the
+	// block to the transactionSet. If the add succeeds, add the transaction back
+	// to the transaction pool.
+	for _, block := range cc.RevertedBlocks {
+		for _, txn := range block.Transactions {
+			err = tp.acceptTransactionSet([]types.Transaction{txn}, cc.TryTransactionSet)
```
DavidVorick: no need to check this error, there is nothing to do either way

modules/transactionpool/update.go
```diff
@@ -52,20 +52,21 @@ func (tp *TransactionPool) ProcessConsensusChange(cc modules.ConsensusChange) {
...
-	// due to a reorg.
-	//
-	// Another option is to scan through the blocks transactions one at a time
-	// check if they are valid. If so, lump them in a set with the next guy.
-	// When they stop being valid, you've found a guy to throw away. It's n^2
-	// in the number of transactions in the block.
+	// Scan through the reverted blocks and try to add each transaction in the
+	// block to the transactionSet. If the add succeeds, add the transaction back
+	// to the transaction pool.
+	for _, block := range cc.RevertedBlocks {
```
DavidVorick: also, looks like this is not going in the correct order. You should be crawling through the reverted blocks in reverse order.

modules/transactionpool/update_test.go
```diff
@@ -40,3 +41,94 @@ func TestArbDataOnly(t *testing.T) {
...
+
+	// verify the transaction pool still has the reorged txns
+	for _, txnSet := range txnSets {
+		for _, txn := range txnSet {
+			if !tpt.tpool.transactionConfirmed(tpt.tpool.dbTx, txn.ID()) {
```
DavidVorick: `transactionConfirmed` tells you whether or not the transaction is confirmed on the blockchain, which is different from telling you whether  the transaction is in the transaction pool.

DavidVorick:
Except for any NDFs, I've cleaned this up and it should be ready for merge.

modules/transactionpool/update_test.go
```diff
@@ -40,3 +41,104 @@ func TestArbDataOnly(t *testing.T) {
...
+		}
+	}
+	if !success {
+		t.Fatal("testers did not have the same block height after one minute")
+	}
+
+	// verify the transaction pool still has the reorged txns
+	for _, txnSet := range txnSets {
+		for _, txn := range txnSet {
+			_, _, exists := tpt.tpool.Transaction(txn.ID())
```
johnathanhowell: should we be checking that `tpt2` has `txn.ID()` as well?

DavidVorick: it's not going to, because you don't broadcast your existing tpool when someone joins the network.

I guess that's something we should consider, but it's out of scope for this PR either way.

PR #1869, opened 2017-06-02
Update Explorer to track file contract payouts

ChrisSchinnerl:


modules/explorer/update.go
```diff
@@ -222,6 +220,26 @@ func (e *Explorer) ProcessConsensusChange(cc modules.ConsensusChange) {
 			}
 		}
 
+		// Update stats for according to SiacoinOutputDiffs
+		for _, scod := range cc.SiacoinOutputDiffs {
+			scoid := scod.ID
+			sco := scod.SiacoinOutput
+
+			if scod.Direction == true {
```
lukechampine: use `modules.DiffApply` instead of `true` here, to better show the intent (ditto for the other predicate below)

modules/explorer/update.go
```diff
@@ -330,8 +354,8 @@ func dbAddFileContractID(tx *bolt.Tx, id types.FileContractID, txid types.Transa
 	mustPutSet(b, txid)
 }
 func dbRemoveFileContractID(tx *bolt.Tx, id types.FileContractID, txid types.TransactionID) {
-	// TODO: delete bucket when it becomes empty
 	mustDelete(tx.Bucket(bucketFileContractIDs).Bucket(encoding.Marshal(id)), txid)
+	deleteEmptyNestedBucket(tx.Bucket(bucketFileContractIDs), encoding.Marshal(id))
```
lukechampine: this name doesn't feel clear enough to me. Perhaps move the conditional to the callsite, i.e.:
```go
func dbRemoveFileContractID(tx *bolt.Tx, id types.FileContractID, txid types.TransactionID) {
    // might be a cleaner way to write this...
    bucket := tx.Bucket(bucketFileContractIDs).Bucket(encoding.Marshal(id))
    mustDelete(bucket, txid)
    if bucketIsEmpty(bucket) {
        tx.Bucket(bucketFileContractIDs).DeleteBucket(encoding.Marshal(id))
    }
}

func bucketIsEmpty(bucket *bolt.Bucket) bool {
    k, _ := bucket.Cursor().First()
    return k == nil
}
```
In general, I prefer having the control flow explicit at the callsite rather than hidden inside helper functions (with `assertNil` being an obvious exception :P)

ChrisSchinnerl: Yes I guess you are right. That name is bad ^^ Seems strange that there is no built in bucket.IsEmpty function

modules/explorer/info_test.go
```diff
@@ -72,3 +74,190 @@ func TestBlockFacts(t *testing.T) {
...
+		panic(err)
+	}
+
+	err = et.tpool.AcceptTransactionSet(tSet)
+	if err != nil {
+		panic(err)
+	}
+	_, err = et.miner.AddBlock()
+	if err != nil {
+		panic(err)
```
lukechampine: Use `t.Fatal` for these, not `panic`

modules/explorer/info_test.go
```diff
@@ -72,3 +74,190 @@ func TestBlockFacts(t *testing.T) {
...
+	// the filecontract was removed from the consensus set
+	outputs, err := et.explorer.FileContractPayouts(fcid)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if len(outputs) != len(fc.ValidProofOutputs) {
+		t.Error("Incorrect number of outputs returned")
+		t.Error("Expecting -> ", len(fc.MissedProofOutputs))
+		t.Error("But was -> ", len(outputs))
```
lukechampine: Use a single line here with Errorf (`expected %v, got %v`)

modules/explorer/update.go
```diff
@@ -222,6 +220,26 @@ func (e *Explorer) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+
+			if scod.Direction == modules.DiffApply {
+				dbAddSiacoinOutput(tx, scoid, sco)
+			}
+		}
+
+		// Update stats for according to SiafundOutputDiffs
+		for _, sfod := range cc.SiafundOutputDiffs {
+			sfoid := sfod.ID
+			sfo := sfod.SiafundOutput
```
lukechampine: These vars don't seem necessary

modules/explorer/info_test.go
ChrisSchinnerl: Ok I will change all panic calls to t.Fatal. Is there a time when panic is prefered over t.Fatal? I noticed that there are lots of calls to panic in accept_txntypes_test.go

lukechampine: Panicking is pretty much a last resort. If it's present in other tests, it's because we were less careful back then. :P sometimes it can be convenient if you don't have access to the testing object.

PR #1870, opened 2017-06-02
prune unconfirmed transactions older than 12 blocks from the tpool

johnathanhowell:
This PR updates the transaction pool to prune transactions that have been in the pool for more than 12 blocks.

### DIFF
--- a/modules/host/network_test.go
+++ b/modules/host/network_test.go
@@ -73,9 +73,10 @@ func TestHostWorkingStatus(t *testing.T) {
 	}
 	defer ht.Close()
 
-	// this causes an ndf, because it relies on the host tester starting up and
-	// fully returning faster than the first check, which isnt always the case.
-	// Disabled for now.
+	// TODO: this causes an ndf, because it relies on the host tester starting up
+	// and fully returning faster than the first check, which isnt always the
+	// case.  This check is disabled for now, but can be fixed by using the
+	// disrupt() pattern.
 	// if ht.host.WorkingStatus() != modules.HostWorkingStatusChecking {
 	// 	t.Fatal("expected working state to initially be modules.HostWorkingStatusChecking")
 	// }
@@ -122,9 +123,10 @@ func TestHostConnectabilityStatus(t *testing.T) {
 	}
 	defer ht.Close()
 
-	// this causes an ndf, because it relies on the host tester starting up and
-	// fully returning faster than the first check, which isnt always the case.
-	// Disabled for now.
+	// TODO: this causes an ndf, because it relies on the host tester starting up
+	// and fully returning faster than the first check, which isnt always the
+	// case.  This check is disabled for now, but can be fixed by using the
+	// disrupt() pattern.
 	// if ht.host.ConnectabilityStatus() != modules.HostConnectabilityStatusChecking {
 	// 		t.Fatal("expected connectability state to initially be ConnectablityStateChecking")
 	// }
--- a/modules/host/network_test.go
+++ b/modules/host/network_test.go
@@ -73,9 +73,12 @@ func TestHostWorkingStatus(t *testing.T) {
 	}
 	defer ht.Close()
 
-	if ht.host.WorkingStatus() != modules.HostWorkingStatusChecking {
-		t.Fatal("expected working state to initially be modules.HostWorkingStatusChecking")
-	}
+	// this causes an ndf, because it relies on the host tester starting up and
+	// fully returning faster than the first check, which isnt always the case.
+	// Disabled for now.
+	// if ht.host.WorkingStatus() != modules.HostWorkingStatusChecking {
+	// 	t.Fatal("expected working state to initially be modules.HostWorkingStatusChecking")
+	// }
 
 	for i := 0; i < 5; i++ {
 		// Simulate some setting calls, and see if the host picks up on it.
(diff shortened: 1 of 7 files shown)


modules/host/network_test.go
```diff
@@ -73,9 +73,12 @@ func TestHostWorkingStatus(t *testing.T) {
...
 
-	if ht.host.WorkingStatus() != modules.HostWorkingStatusChecking {
-		t.Fatal("expected working state to initially be modules.HostWorkingStatusChecking")
-	}
+	// this causes an ndf, because it relies on the host tester starting up and
+	// fully returning faster than the first check, which isnt always the case.
+	// Disabled for now.
+	// if ht.host.WorkingStatus() != modules.HostWorkingStatusChecking {
+	// 	t.Fatal("expected working state to initially be modules.HostWorkingStatusChecking")
+	// }
```
DavidVorick: This is where we could insert a disrupt dependency to make sure that the check doesn't run until the host is full started up, and to also make sure that this check doesn't happen until the first check has run. Since this is all inside of the host package, we can do this pretty easily.

modules/host/network_test.go
```diff
@@ -119,9 +122,12 @@ func TestHostConnectabilityStatus(t *testing.T) {
...
 
-	if ht.host.ConnectabilityStatus() != modules.HostConnectabilityStatusChecking {
-		t.Fatal("expected connectability state to initially be ConnectablityStateChecking")
-	}
+	// this causes an ndf, because it relies on the host tester starting up and
+	// fully returning faster than the first check, which isnt always the case.
+	// Disabled for now.
+	// if ht.host.ConnectabilityStatus() != modules.HostConnectabilityStatusChecking {
+	// 		t.Fatal("expected connectability state to initially be ConnectablityStateChecking")
+	// }
```
DavidVorick: ditto

modules/transactionpool/accept.go
```diff
@@ -311,6 +311,15 @@ func (tp *TransactionPool) acceptTransactionSet(ts []types.Transaction, txnFn fu
 	tp.transactionSetDiffs[setID] = cc
 	tsetSize := len(encoding.Marshal(ts))
 	tp.transactionListSize += tsetSize
+	blockHeight, err := tp.getBlockHeight(tp.dbTx)
```
DavidVorick: we should probably keep this value in memory. Let's extend the putBlockHeight function to also set the in-memory block height in the tpool to the same value, and then read that instead of going to disk.

DavidVorick: oh, looks like you already have the height in memory, let's read from that.

modules/transactionpool/update.go
```diff
@@ -39,6 +46,46 @@ func (tp *TransactionPool) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+	}
+
+	// prune transactions older than maxTxnAge.
+	pruneTxns := make(map[types.TransactionID]struct{})
+	for txid, seenHeight := range tp.transactionHeights {
+		if tp.blockHeight-seenHeight > maxTxnAge {
+			delete(tp.transactionHeights, txid)
+			pruneTxns[txid] = struct{}{}
+		}
+	}
```
DavidVorick: hmm. I'm worried that scrolling through the entire list of transctions every block is not going to scale very well if we ever put the tpool at millions of transactions. (which just be a few hundred MBs).

I'm trying to think of more scalable ways to do this. Linear time stuff works up to thousands, but at tens of thousands you need to start reconsidering.

DavidVorick: Also, I don't think there's any need to make a map here.

modules/transactionpool/update.go
```diff
@@ -39,6 +46,46 @@ func (tp *TransactionPool) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+	for id, tSet := range tp.transactionSets {
+		var validTxns []types.Transaction
+		for _, txn := range tSet {
+			_, shouldPrune := pruneTxns[txn.ID()]
+			if !shouldPrune {
```
DavidVorick: instead of checking `shouldPrune`, can't you just check tp.transactionHeights inline?

modules/transactionpool/update.go
DavidVorick: I just realized that we already call `acceptTransactionSet` on every single transaction every single block, so already we're doing way worse than a map lookup.

modules/transactionpool/update.go
johnathanhowell: definitely

modules/transactionpool/update.go
```diff
@@ -39,6 +46,46 @@ func (tp *TransactionPool) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+				for _, diff := range setDiffs.FileContractDiffs {
+					delete(tp.knownObjects, ObjectID(diff.ID))
+				}
+				for _, diff := range setDiffs.SiafundOutputDiffs {
+					delete(tp.knownObjects, ObjectID(diff.ID))
+				}
+			}
+			delete(tp.transactionSetDiffs, id)
+		}
+	}
```
DavidVorick: all the transactions get thrown out at `t.purge()` already anyway, there's no need to manually extract them. Instead, after the purge happens, just scroll through the list of `unconfirmedSets` and check the transactions

johnathanhowell: Gotcha. That makes this much cleaner.

DavidVorick:
test looks good

modules/host/network_test.go
DavidVorick: please drop a TODO here so we can grep for it later

DavidVorick:
LGTM

Fornax96:
Is this a good idea? What if a real fee market starts to exist, will people's transactions just get dropped after 2 hours? In bitcoin when I send a transaction with a low fee I expect it to get confirmed in about a week or more. 

I hope you will implement a proper mempool in the future that doesn't just drop transactions out of the blue (unless they are invalid).

Fornax96:
I would personally only prune transactions after 7 days _at least_.

DavidVorick:
You have to balance that with memory concerns. If a transaction isn't confirming after a few hours it should be up to the software to keep trying again until the transaction becomes invalid, including fee bumping if necessary.

We are going to make these parameters tuneable, but ultimately the job of the mempool is to help smooth out huge usage spikes, not to make sure that every transaction gets confirmed after sitting in memory for multiple days.

The solutions to handle intentionally-anemic fees (by those such as yourself, willing to wait a few days to get a 80% discount or whatever) should be client-side, not network-side. We will continue to extend the client to be useful for this. It might also make sense to have third party fee-rebroadcasting services, who maybe take a tiny cut in return for rebroadcasting your transaction every few hours until it gets confirmed.

PR #1876, opened 2017-06-03
add API support for send-to-many

lukechampine:
This allows users to post a set of amounts and addresses to `/wallet/siacoins`, the same endpoint currently used for single amount+address pairs. The rationale is to make it easy to split a large input among many outputs without requiring a separate transaction for each (which is slower and can cause fragmentation).
To create a send-to-many transaction, post a list of comma-separated amounts/addrs in the `amounts`/`destinations` form values, instead of the typical `amount`/`destination` form values. The backing logic is implemented in a new `SendSiacoinsMulti` method of the wallet.

PR'ing this early to invite discussion. Should we create a new route for send-to-many, instead of overloading `/wallet/siacoins`? Should the api handler construct the transaction itself using a transaction builder? And how should this PR interact with #1849?

### DIFF
--- a/api/wallet.go
+++ b/api/wallet.go
@@ -367,46 +367,37 @@ func (api *API) walletSeedsHandler(w http.ResponseWriter, req *http.Request, _ h
 
 // walletSiacoinsHandler handles API calls to /wallet/siacoins.
 func (api *API) walletSiacoinsHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
-	var txns []types.Transaction
-	if req.FormValue("destination") != "" {
-		// single amount + destination
-		amount, ok := scanAmount(req.FormValue("amount"))
+	amountStrs := strings.Split(req.FormValue("amount"), ",")
+	destStrs := strings.Split(req.FormValue("destination"), ",")
+	amounts := make([]types.Currency, len(amountStrs))
+	dests := make([]types.UnlockHash, len(destStrs))
+	for i, amountStr := range amountStrs {
+		var ok bool
+		amounts[i], ok = scanAmount(amountStr)
 		if !ok {
-			WriteError(w, Error{"could not read 'amount' from POST call to /wallet/siacoins"}, http.StatusBadRequest)
+			WriteError(w, Error{"could not read amount from POST call to /wallet/siacoins"}, http.StatusBadRequest)
 			return
 		}
-		dest, err := scanAddress(req.FormValue("destination"))
+	}
+	var err error
+	for i, destStr := range destStrs {
+		dests[i], err = scanAddress(destStr)
 		if err != nil {
-			WriteError(w, Error{"error after call to /wallet/siacoins: " + err.Error()}, http.StatusBadRequest)
+			WriteError(w, Error{"could not read address from POST call to /wallet/siacoins"}, http.StatusBadRequest)
 			return
 		}
-		txns, err = api.wallet.SendSiacoins(amount, dest)
+	}
+
+	var txns []types.Transaction
+	if len(amounts) == 1 && len(dests) == 1 {
+		// single amount + destination
+		txns, err = api.wallet.SendSiacoins(amounts[0], dests[0])
 		if err != nil {
 			WriteError(w, Error{"error after call to /wallet/siacoins: " + err.Error()}, http.StatusInternalServerError)
 			return
 		}
 	} else {
 		// multiple amounts + destinations
-		amountStrs := strings.Split(req.FormValue("amounts"), ",")
-		destStrs := strings.Split(req.FormValue("destinations"), ",")
-		amounts := make([]types.Currency, len(amountStrs))
-		dests := make([]types.UnlockHash, len(destStrs))
-		for i, amountStr := range amountStrs {
-			var ok bool
-			amounts[i], ok = scanAmount(amountStr)
-			if !ok {
-				WriteError(w, Error{"could not read amount from POST call to /wallet/siacoins"}, http.StatusBadRequest)
-				return
-			}
-		}
-		var err error
-		for i, destStr := range destStrs {
-			dests[i], err = scanAddress(destStr)
-			if err != nil {
-				WriteError(w, Error{"could not read address from POST call to /wallet/siacoins"}, http.StatusBadRequest)
-				return
-			}
-		}
 		txns, err = api.wallet.SendSiacoinsMulti(amounts, dests)
 		if err != nil {
 			WriteError(w, Error{"error after call to /wallet/siacoins: " + err.Error()}, http.StatusInternalServerError)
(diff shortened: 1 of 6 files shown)


DavidVorick:
Worth noting that this call is going to make life easier, and also make the blockchain smaller, for parties like exchanges and mining pools that frequently have to perform many payouts per block.

Overloading /wallet/siacoins seems okay to me.

Seems also reasonable that you could still set a fee on this set. In both cases, you are only setting a single fee. Let's keep the other PR hanging though until the fee market is implemented. I've got the first half of it done, just getting the tests working again before PR'ing it.

api/wallet.go
```diff
@@ -367,22 +367,53 @@ func (api *API) walletSeedsHandler(w http.ResponseWriter, req *http.Request, _ h
...
-	if !ok {
-		WriteError(w, Error{"could not read 'amount' from POST call to /wallet/siacoins"}, http.StatusBadRequest)
-		return
-	}
-	dest, err := scanAddress(req.FormValue("destination"))
-	if err != nil {
-		WriteError(w, Error{"error after call to /wallet/siacoins: " + err.Error()}, http.StatusBadRequest)
-		return
+	var txns []types.Transaction
+	if req.FormValue("destination") != "" {
```
DavidVorick: I'm almost in favor of having them be the same. What happens if you write only a single amount + destination inthe 'amounts' and 'destinations' fields? Do you get a txn that is meaningfully different from just using 'amount' and 'destination' ?

api/wallet.go
```diff
@@ -367,22 +367,53 @@ func (api *API) walletSeedsHandler(w http.ResponseWriter, req *http.Request, _ h
...
+		}
+		txns, err = api.wallet.SendSiacoins(amount, dest)
+		if err != nil {
+			WriteError(w, Error{"error after call to /wallet/siacoins: " + err.Error()}, http.StatusInternalServerError)
+			return
+		}
+	} else {
+		// multiple amounts + destinations
+		amountStrs := strings.Split(req.FormValue("amounts"), ",")
+		destStrs := strings.Split(req.FormValue("destinations"), ",")
```
DavidVorick: need to check that the length is the same here

DavidVorick: hmm, I guess we could let SendSiacoinsMulti do the checking, but I'm not 100% sure what is best.

modules/wallet/money.go
```diff
@@ -111,6 +113,62 @@ func (w *Wallet) SendSiacoins(amount types.Currency, dest types.UnlockHash) ([]t
...
+	}
+	if len(amounts) != len(dests) {
+		return nil, errors.New("number of amounts and destinations must match")
+	}
+
+	txnBuilder := w.StartTransaction()
+
+	// Add estimated transaction fee.
+	_, tpoolFee := w.tpool.FeeEstimation()
+	tpoolFee = tpoolFee.Mul64(750 * uint64(len(dests))) // Estimated transaction size in bytes
```
DavidVorick: can no longer use 750 as a reasonable guess. Should make some trial transactions to see. A send-to-5, send-to-50, etc.

I'm guessing it'll take somewhere around 150 bytes per destination+amount, but  really we should measure that.

modules/wallet/money.go
```diff
@@ -111,6 +113,62 @@ func (w *Wallet) SendSiacoins(amount types.Currency, dest types.UnlockHash) ([]t
...
+		return nil, err
+	}
+	defer w.tg.Done()
+	if !w.unlocked {
+		w.log.Println("Attempt to send coins has failed - wallet is locked")
+		return nil, modules.ErrLockedWallet
+	}
+	if len(amounts) != len(dests) {
+		return nil, errors.New("number of amounts and destinations must match")
+	}
```
DavidVorick: need to check that the number of amounts is not larger than what can reasonably fit in an IsStandard transaction - 32kb. We should probably target 24kb or so, to give wiggleroom for weirdness (signatures may be large, or you may be spending tons of outputs, or something).

DavidVorick:
Overall looks pretty good in terms of design.

modules/wallet/money.go
DavidVorick: *plus 600 base just for calling `FundSiacoins` and having all the other txn trimmings

api/wallet.go
lukechampine: Hmm. That seems ok to me. We'd probably still need a separate wallet method though. Ideally you could modify `SendSiacoins` to accept multiple addrs, but that would require updating every occurence of `SendSiacoins` in the codebase. Not worth it.

api/wallet.go
lukechampine: Currently `SendSiacoinsMulti` handles it. Another possibility is:
```go
type SendDestination struct {
    Amount types.Currency
    Addr   types.UnlockHash
}

func SendSiacoinsMulti(dests []SendDestination) {
    // ...
}
```
Then it would be impossible to mismatch the number of each in the call to `SendSiacoinsMulti`. This would imply that the api would have to be responsible for ensuring that the numbers of amounts and addresses match.

I'm not inclined to go this route though, mostly because I'm not sure what to name the new type, and because it's not as readable.

api/wallet_test.go
```diff
@@ -1437,3 +1438,82 @@ func TestWalletChangePassword(t *testing.T) {
 		t.Error("wallet is not unlocked")
 	}
 }
+
+// TestWalletSiacoins tests the /wallet/siacoins endpoint, including sending
+// to multiple addresses.
```
DavidVorick: This test only checks that send-to-many works, does not check that single sends work. Granted, we use single sends all over the place for API testing I don't think it's something that strictly needs a separate test, but docstring is misleading.

api/wallet_test.go
```diff
@@ -1437,3 +1438,82 @@ func TestWalletChangePassword(t *testing.T) {
...
+	for i, w := range wallets[1:] {
+		var wg WalletGET
+		err = w.getAPI("/wallet", &wg)
+		if err != nil {
+			t.Fatal(err)
+		}
+		if !wg.ConfirmedSiacoinBalance.Equals(sendAmount) {
+			t.Errorf("wallet %d should have %v coins, has %v", i+2, sendAmount, wg.ConfirmedSiacoinBalance)
+		}
+	}
```
DavidVorick: should probably have a second component of this test that tries to do a send of 250 in size, since that's the limit advertised by the api docs. You can just send all 250 to the same wallet, pull 250 addresses out in a loop.

DavidVorick: oh, I can do that for the validation testing

api/wallet_test.go
```diff
@@ -1437,3 +1438,82 @@ func TestWalletChangePassword(t *testing.T) {
...
+
+	// mine blocks until the send is confirmed
+	for i := types.BlockHeight(0); i < types.MaturityDelay; i++ {
+		_, err := st.miner.AddBlock()
+		if err != nil {
+			t.Fatal(err)
+		}
+	}
+	// allow some time for blocks to propagate
+	time.Sleep(time.Second)
```
DavidVorick: This sleep will result in NDFs, there is a helper function `waitForBlock` that you can use instead.

DavidVorick: oh, actually in this case you'd want `synchronizationCheck`

PR #1877, opened 2017-06-04
Add market based transaction accepting and fee estimation

DavidVorick:
This code is the first step to adding a fee market. There are a few TODOs scattered around.

### DIFF
--- a/modules/transactionpool.go
+++ b/modules/transactionpool.go
@@ -82,10 +82,6 @@ type TransactionPool interface {
 	// within 10 blocks.
 	FeeEstimation() (minimumRecommended, maximumRecommended types.Currency)
 
-	// IsStandardTransaction returns `err = nil` if the transaction is
-	// standard, otherwise it returns an error explaining what is not standard.
-	IsStandardTransaction(types.Transaction) error
-
 	// PurgeTransactionPool is a temporary function available to the miner. In
 	// the event that a miner mines an unacceptable block, the transaction pool
 	// will be purged to clear out the transaction pool and get rid of the
--- a/modules/transactionpool/accept.go
+++ b/modules/transactionpool/accept.go
@@ -127,7 +127,7 @@ func (tp *TransactionPool) checkTransactionSetComposition(ts []types.Transaction
 	// fly.
 
 	// Check that all transactions follow 'Standard.md' guidelines.
-	setSize, err := tp.IsStandardTransactionSet(ts)
+	setSize, err := isStandardTransactionSet(ts)
 	if err != nil {
 		return 0, err
 	}
--- a/modules/transactionpool/accept.go
+++ b/modules/transactionpool/accept.go
@@ -6,6 +6,7 @@ package transactionpool
 import (
 	"errors"
 	"fmt"
+	"math"
 	"time"
 
 	"github.com/NebulousLabs/Sia/build"
@@ -98,13 +99,8 @@ func (tp *TransactionPool) requiredFeesToExtendTpool() types.Currency {
 	}
 
 	// Calculate the fee required to bump out the size of the transaction pool.
-	numer := TransactionPoolSizeTarget
-	denom := tp.transactionListSize
-	for i := 1; i < TransactionPoolExponentiation; i++ {
-		numer *= TransactionPoolSizeTarget
-		denom *= tp.transactionListSize
-	}
-	feeFactor := float64(numer) / float64(denom)
+	ratioToTarget := float64(tp.transactionListSize) / TransactionPoolSizeTarget
+	feeFactor := math.Pow(ratioToTarget, TransactionPoolExponentiation)
 	return types.SiacoinPrecision.MulFloat(feeFactor).Div64(1000)
 }
 
--- a/modules/transactionpool/accept.go
+++ b/modules/transactionpool/accept.go
@@ -16,18 +16,20 @@ import (
 )
 
 const (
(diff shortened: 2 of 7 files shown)


modules/transactionpool/accept.go
```diff
@@ -85,51 +87,37 @@ func relatedObjectIDs(ts []types.Transaction) []ObjectID {
...
+		return types.ZeroCurrency
+	}
+
+	// Calculate the fee required to bump out the size of the transaction pool.
+	numer := TransactionPoolSizeTarget
+	denom := tp.transactionListSize
+	for i := 1; i < TransactionPoolExponentiation; i++ {
+		numer *= TransactionPoolSizeTarget
+		denom *= tp.transactionListSize
+	}
```
lukechampine: `math.Pow` is probably better here. Also, won't `numer` be a constant, since it's derived solely from other constants?

modules/transactionpool.go
```diff
@@ -82,10 +82,6 @@ type TransactionPool interface {
 	// within 10 blocks.
 	FeeEstimation() (minimumRecommended, maximumRecommended types.Currency)
 
-	// IsStandardTransaction returns `err = nil` if the transaction is
-	// standard, otherwise it returns an error explaining what is not standard.
-	IsStandardTransaction(types.Transaction) error
-
```
lukechampine: Woo, shrinking an interface! I guess we just never use this anywhere? Does it still need to be an exported method, then?

modules/transactionpool/accept.go
lukechampine:
```go
ratioToTarget := TransactionPoolSizeTarget / float64(tp.transactionListSize)
feeFactor := math.Pow(ratioToTarget, TransactionPoolExponentiation)
```
I'm confused -- doesn't this mean that as `tp.transactionListSize` increases, the required fee *decreases*? Shouldn't it be the other way around?

DavidVorick: ah whoops I flipped them.

modules/transactionpool.go
DavidVorick: guess not, went ahead and fixed that up.

modules/transactionpool/accept.go
```diff
@@ -85,51 +88,32 @@ func relatedObjectIDs(ts []types.Transaction) []ObjectID {
...
+	// are no fees.
+	if tp.transactionListSize < TransactionPoolSizeForFee {
+		return types.ZeroCurrency
 	}
-	return nil
+
+	// Calculate the fee required to bump out the size of the transaction pool.
+	ratioToTarget := float64(tp.transactionListSize) / TransactionPoolSizeTarget
+	feeFactor := math.Pow(ratioToTarget, TransactionPoolExponentiation)
+	return types.SiacoinPrecision.MulFloat(feeFactor).Div64(1000)
```
lukechampine: The `Div(1000)` is a bit magic; maybe clearer with a comment and separate var:
```go
targetFee := types.SiacoinPrecision.Div64(1000) // 1 SC / KB
return targetFee.MulFloat(feeFactor)
```
not a blocking issue though

modules/transactionpool/transactionpool.go
```diff
@@ -129,18 +129,50 @@ func (tp *TransactionPool) Close() error {
...
-	// really should look more like 10 SC per kb. But, legacy renters are using
-	// a much lower value, which means hosts would be incompatible if the
-	// minimum recommended were set to 10. The value has been set to 1, which
-	// should be okay temporarily while the renters are given time to upgrade.
-	return types.SiacoinPrecision.Mul64(1).Div64(20).Div64(1e3), types.SiacoinPrecision.Mul64(1).Div64(1e3) // TODO: Adjust down once miners have upgraded.
+	// The third method just has hardcoded minimums as a sanity check. In the
+	// event of empty blocks, there should still be some fees being added to the
+	// chain.
+
+	// TODO: Implement the blockchain method.
```
lukechampine: will this be implemented for this PR?

DavidVorick: Plan is to implement in the next PR to keep this PR small

types/transaction_helpers.go
```diff
@@ -0,0 +1,110 @@
...
+//  o   o
+//   \ /
+//    o
+//   /|\
+//   \| \
+//    o  x // 'x' transactions are symbolic, not actually created
+//    |
+//    x
+//
+func TransactionGraph(sourceOutput SiacoinOutputID, sources []int, dests []int, outputValues []Currency, fees []Currency) ([]Transaction, error) {
```
lukechampine: why not create a new type instead of four separate slices?
```go
type TransactionGraphEdge struct {
    Input      *TransactionGraphEdge
    Output     *TransactionGraphEdge
    Value, Fee types.Currency
}
```
Then you wire up the edges yourself, at the callsite, and the `TransactionGraph` function just converts them into a real transaction.
```go
edges := make([]types.TransactionGraphEdge, 3)
edges[0].Value = types.SiacoinPrecision
edges[0].Output = &edges[1]
edges[1].Value = types.SiacoinPrecision
edges[1].Output = &edges[2]
types.TransactionGraph(sourceOutput, edges)
```
(definitely some kinks in this that I haven't fully thought out, but you get the idea)

DavidVorick: thinking about it. Then we get another type in our godoc, which I'm not super happy with. But 4 slices that are all tuples is pretty gross too.

4 slices has been easy enough to work with, though it produces pretty ugly code.

PR #1878, opened 2017-06-04
consensus: apply 10 blocks in one Bolt transaction

starius:
Bolt transaction involved fsync calls which are major time consumers during initial blocks download (IBD) according to trace tool. Applying many blocks (namely 10 - the size of newBlocks during IBD) in a single Bolt transaction results in a speedup.

Speedup numbers coming soon.

### DIFF
--- a/modules/consensus/accept.go
+++ b/modules/consensus/accept.go
@@ -189,7 +189,7 @@ func (cs *ConsensusSet) addBlockToTree(tx *bolt.Tx, b types.Block, parent *proce
 // This method is typically only be used when there would otherwise be multiple
 // consecutive calls to AcceptBlock with each successive call accepting the
 // child block of the previous call.
-func (cs *ConsensusSet) managedAcceptBlocks(blocks []types.Block) (bool, error) {
+func (cs *ConsensusSet) managedAcceptBlocks(blocks []types.Block) error {
 	// Grab a lock on the consensus set.
 	cs.mu.Lock()
 	defer cs.mu.Unlock()
@@ -197,7 +197,7 @@ func (cs *ConsensusSet) managedAcceptBlocks(blocks []types.Block) (bool, error) 
 	// Make sure that blocks are consecutive.
 	for i, b := range blocks {
 		if i > 0 && b.Header().ParentID != blocks[i-1].ID() {
-			return false, errOrphan
+			return errOrphan
 		}
 	}
 
@@ -278,26 +278,20 @@ func (cs *ConsensusSet) managedAcceptBlocks(blocks []types.Block) (bool, error) 
 		}
 		return nil
 	})
-	chainExtended := len(changes.AppliedBlocks) > 0
 	if err != nil {
-		return chainExtended, err
+		return err
 	}
 	// The order is important here: subscribers must be updated after
 	// the check for fatal errors (which result in rollback) but before
 	// the check for other errors which don't result in rollback.
-	if chainExtended {
+	if len(changes.AppliedBlocks) > 0 {
 		cs.readlockUpdateSubscribers(changes)
 	}
-	if err2 != nil {
-		return chainExtended, err2
-	}
-
-	return chainExtended, nil
+	return err2
 }
 
 func (cs *ConsensusSet) managedAcceptBlock(b types.Block) error {
-	_, err := cs.managedAcceptBlocks([]types.Block{b})
-	return err
+	return cs.managedAcceptBlocks([]types.Block{b})
 }
 
 // AcceptBlock will try to add a block to the consensus set. If the block does
--- a/modules/consensus/accept.go
+++ b/modules/consensus/accept.go
@@ -195,7 +195,7 @@ func (cs *ConsensusSet) addBlockToTree(tx *bolt.Tx, b types.Block, parent *proce
 // This method is typically only be used when there would otherwise be multiple
 // consecutive calls to AcceptBlock with each successive call accepting the
 // child block of the previous call.
-func (cs *ConsensusSet) managedAcceptBlocks(blocks []types.Block) error {
+func (cs *ConsensusSet) managedAcceptBlocks(blocks []types.Block) (bool, error) {
 	// Grab a lock on the consensus set.
 	cs.mu.Lock()
 	defer cs.mu.Unlock()
@@ -203,13 +203,14 @@ func (cs *ConsensusSet) managedAcceptBlocks(blocks []types.Block) error {
 	// Make sure that blocks are consecutive.
 	for i, b := range blocks {
 		if i > 0 && b.Header().ParentID != blocks[i-1].ID() {
-			return errOrphan
+			return false, errOrphan
 		}
 	}
 
 	var changes changeEntry
 	var err2 error
 
+	chainExtended := false
 	err := cs.db.Update(func(tx *bolt.Tx) error {
 		// Do not accept a block if the database is inconsistent.
 		if inconsistencyDetected(tx) {
@@ -278,24 +279,26 @@ func (cs *ConsensusSet) managedAcceptBlocks(blocks []types.Block) error {
 			}
 			changes.RevertedBlocks = append(changes.RevertedBlocks, changeEntry.RevertedBlocks...)
 			changes.AppliedBlocks = append(changes.AppliedBlocks, changeEntry.AppliedBlocks...)
+			chainExtended = true
 		}
 		return nil
 	})
 	if err != nil {
-		return err
+		return chainExtended, err
 	}
 	if len(changes.AppliedBlocks) > 0 {
 		cs.readlockUpdateSubscribers(changes)
 	}
 	if err2 != nil {
-		return err2
+		return chainExtended, err2
 	}
 
-	return nil
+	return chainExtended, nil
 }
 
 func (cs *ConsensusSet) managedAcceptBlock(b types.Block) error {
-	return cs.managedAcceptBlocks([]types.Block{b})
+	_, err := cs.managedAcceptBlocks([]types.Block{b})
+	return err
 }
 
 // AcceptBlock will try to add a block to the consensus set. If the block does
--- a/modules/consensus/accept.go
+++ b/modules/consensus/accept.go
@@ -204,7 +204,6 @@ func (cs *ConsensusSet) managedAcceptBlocks(blocks []types.Block) (bool, error) 
 	var changes changeEntry
 	var err2 error
 
-	chainExtended := false
 	err := cs.db.Update(func(tx *bolt.Tx) error {
 		// Do not accept a block if the database is inconsistent.
 		if inconsistencyDetected(tx) {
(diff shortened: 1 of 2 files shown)


starius:
I fixed a bug in my PR in https://github.com/NebulousLabs/Sia/pull/1878/commits/0c7517ab03270177dc58b8977fd09a428d074dc4 but it is still failing:

```
=== RUN   TestIntegrationRPCSendBlocks
--- FAIL: TestIntegrationRPCSendBlocks (11.72s)
	synchronize_test.go:494: block has no known parent
	synchronize_test.go:506: SendBlocks should reorg the local CS when the remote CS's chain is longer: expected height 10, got 5
	synchronize_test.go:509: SendBlocks should reorg the local CS when the remote CS's chain is longer: remote and local CSTs have different current blocks
	synchronize_test.go:494: block has no known parent
	synchronize_test.go:506: SendBlocks should reorg the local CS when the remote CS's chain is longer: expected height 20, got 15
	synchronize_test.go:509: SendBlocks should reorg the local CS when the remote CS's chain is longer: remote and local CSTs have different current blocks
	synchronize_test.go:506: SendBlocks should reorg the local CS when the remote CS's chain is longer: expected height 2, got 1
	synchronize_test.go:509: SendBlocks should reorg the local CS when the remote CS's chain is longer: remote and local CSTs have different current blocks
	synchronize_test.go:506: SendBlocks should reorg the local CS when the remote CS's chain is longer: expected height 3, got 1
	synchronize_test.go:509: SendBlocks should reorg the local CS when the remote CS's chain is longer: remote and local CSTs have different current blocks
	synchronize_test.go:494: block has no known parent
	synchronize_test.go:506: SendBlocks should reorg the local CS when the remote CS's chain is longer: expected height 4, got 1
	synchronize_test.go:509: SendBlocks should reorg the local CS when the remote CS's chain is longer: remote and local CSTs have different current blocks
	synchronize_test.go:494: block has no known parent
	synchronize_test.go:506: SendBlocks should reorg the local CS when the remote CS's chain is longer: expected height 7, got 4
	synchronize_test.go:509: SendBlocks should reorg the local CS when the remote CS's chain is longer: remote and local CSTs have different current blocks
```
and other. Any idea why?

starius:
One more error found: `modules.ErrNonExtendingBlock` caused a transaction rollback, but it shouldn't.

starius:
I fixed with problem with `modules.ErrNonExtendingBlock` but it still fails: `validateHeaderAndBlock` returns `errOrphan` for a block which is a child of a block that doesn't extend main blockchain. I'm looking into why `blockMap.Get` returns `nil` in this case.

starius:
I found and fixed more bugs. Now all tests pass locally.

lukechampine:
Thanks, I've been meaning to get back to this. Do we have any benchmarks showing speed up?

starius:
I remember that the initial version of this did a significant speedup. Unfortunately the exact numbers are lost in the Slack history.

I tried to redo speedup measurements against latest head (e57306d) and found that IBD is broken: https://github.com/NebulousLabs/Sia/issues/1958

lukechampine:
well that's no good. Likely a tpool issue. We'll have to use `git bisect` to identify which commit introduced the bug.

lukechampine:
@starius you should be able to do IBD on the `ibd-fix` branch

starius:
Preliminary results of the benchmark.
Machines: GCP machines with 15G RAM and 4 CPU, located in us-west1-b.

Base code version: latest master (dfaf1612a1552714d2d14ba6d6760cd73c53d80e) + https://github.com/NebulousLabs/Sia/pull/1963

```
hdd-before:
still running, real time is 560.
Height: 103580
Progress (estimated): 93.5%

hdd-after:
real    168m1.142s
user    62m0.836s
sys     2m56.192s

ssd-before:
real    263m36.443s
user    65m47.592s
sys     5m15.636s

ssd-after:
real    175m44.029s
user    57m7.016s
sys     2m29.544s
```

DavidVorick:
From over 10 hours to under 3. This is definitely a very good thing. I will try to review it today, we're almost done with the v1.3.0 scalability stuff.

modules/consensus/accept.go
```diff
@@ -137,29 +137,20 @@ func (cs *ConsensusSet) validateHeader(tx dbTx, h types.BlockHeader) error {
...
-// committed. Switching to a managed tx through bolt will make this complexity
+// addBlockToTree might need to modify the database while returning an error
+// on the block. Such errors are handled outside of the transaction by the
+// caller. Switching to a managed tx through bolt will make this complexity
 // unneeded.
-func (cs *ConsensusSet) addBlockToTree(b types.Block) (ce changeEntry, err error) {
-	var nonExtending bool
-	err = cs.db.Update(func(tx *bolt.Tx) error {
+func (cs *ConsensusSet) addBlockToTree(tx *bolt.Tx, b types.Block) (ce changeEntry, err error) {
+	err = func() error {
```
DavidVorick: I don't think this func needs to be here anymore, the error catching has been cleaned up some already, and since we've don't that I don't really see a reason to leave this func.

modules/consensus/accept.go
```diff
@@ -207,78 +195,110 @@ func (cs *ConsensusSet) addBlockToTree(b types.Block) (ce changeEntry, err error
...
-func (cs *ConsensusSet) managedAcceptBlock(b types.Block) error {
-	// Grab a lock on the consensus set. Lock is demoted later in the function,
-	// failure to unlock before returning an error will cause a deadlock.
+func (cs *ConsensusSet) managedAcceptBlocks(blocks []types.Block) (bool, error) {
+	// Grab a lock on the consensus set.
 	cs.mu.Lock()
+	defer cs.mu.Unlock()
+
+	// Make sure that blocks are consecutive.
+	for i, b := range blocks {
```
DavidVorick: This is an interesting requirement, are you sure that it's necessary? Submitting a bunch of blocks in random order should actually be okay here too, though that's something we'd want to test.

modules/consensus/accept.go
```diff
@@ -207,78 +195,110 @@ func (cs *ConsensusSet) addBlockToTree(b types.Block) (ce changeEntry, err error
 // This method is typically only be used when there would otherwise be multiple
 // consecutive calls to AcceptBlock with each successive call accepting the
 // child block of the previous call.
-func (cs *ConsensusSet) managedAcceptBlock(b types.Block) error {
-	// Grab a lock on the consensus set. Lock is demoted later in the function,
-	// failure to unlock before returning an error will cause a deadlock.
+func (cs *ConsensusSet) managedAcceptBlocks(blocks []types.Block) (bool, error) {
```
DavidVorick: Looks like we need to reconsider  the interplay between the bool and the error - if we are accepting a whole batch of blocks, the blocks could individually have different errors, and some of them may extend the chain while others do not.

In a reorg situation, you could end up with a bunch of blocks that don't extend the chain followed by only one that does. I'm not sure how that plays with the rest of the code, but my initial leaning is to return an `[]error` instead of a `bool, error`, that way you can return the errors for each block individually, and then handle them as you would have if you had submitted the blocks in a loop.

modules/consensus/accept.go
```diff
@@ -207,78 +195,110 @@ func (cs *ConsensusSet) addBlockToTree(b types.Block) (ce changeEntry, err error
...
+				continue
+			} else if err != nil {
+				return err
+			}
+			// If appliedBlocks is 0, revertedBlocks will also be 0.
+			if build.DEBUG && len(changeEntry.AppliedBlocks) == 0 && len(changeEntry.RevertedBlocks) != 0 {
+				panic("appliedBlocks and revertedBlocks are mismatched!")
+			}
+			changes.RevertedBlocks = append(changes.RevertedBlocks, changeEntry.RevertedBlocks...)
+			changes.AppliedBlocks = append(changes.AppliedBlocks, changeEntry.AppliedBlocks...)
```
DavidVorick: Hmmm. I'm not sure you can do this, the consensus changes have a pretty strict requirement that all reverted blocks must follow all applied blocks.

The requirement you have that all blocks must be in a row may make this change okay. Bundling consensus changes all together like this would also help contribute to performance improvements, I think.

I need to think about it some more, but I don't think this is robust enough. Need to be certain that the change you are building represents a single path from the original head block to the new head block (rewinding any blocks that got removed, and then adding on only the blocks that got added).

DavidVorick: I think this code would be okay the way it is currently written, but then we have to keep the requirement that grouped blocks always be in order, and once we do something like headers-first implementation, I'm not sure we're going to want that requirement.

modules/consensus/accept.go
DavidVorick: For the time being, I think that this is a good requirement to have. It's not strictly necessary, and will probably be changed once we are doing headers-first downloading and parallel block downloading, but if we get rid of it now there will be more code complexity needed to maintain correctness.

modules/consensus/accept.go
```diff
@@ -207,78 +195,110 @@ func (cs *ConsensusSet) addBlockToTree(b types.Block) (ce changeEntry, err error
...
+			// not extend the current blockchain, however the changes from
+			// addBlockToTree should be committed (which means 'nil' must
+			// be returned). A flag is set to indicate that modules.ErrNonExtending
+			// should be returned.
+			if err == modules.ErrNonExtendingBlock || err == modules.ErrBlockKnown {
+				err2 = err
+				continue
+			} else if err != nil {
+				return err
+			}
```
DavidVorick: If there is a series of valid blocks followed by an invalid block, all of them will get rolled back.

DavidVorick: I think that's okay, you'd have to re-download the valid ones from a new peer, but I don't think you'd get stuck.

modules/consensus/accept.go
starius: Fixed

modules/consensus/accept.go
starius: I'm keeping the check taking into account the fact that it is needed to apply changes in groups (https://github.com/NebulousLabs/Sia/pull/1878#discussion_r123674511).

modules/consensus/accept.go
starius: I considered this interface, but it seems to be excessively complex (e.g., what to return in case of error in the middle of the group in the check for consecutiveness I added) but doesn't provide add information for the caller. If we encounter a non-fatal error (ErrBlockKnown, ErrNonExtendingBlock), we switch to next block; in case of a fatal error we don't consider subsequent blocks because they are children of a block causing a fatal error.

Basically we don't even need `bool` here (it is added in a separate commit but everything works before that commit). It was added to inform a caller (`managedReceiveBlocks`) if a blockchain was extended in case of a non-fatal error.

modules/consensus/accept.go
starius: I'm keeping the check that grouped blocks always be in order.

By the way, what is the purpose of headers-first? IBD is not a parallel process anyway: even if blocks are downloaded in advance, they are applied one by one. Downloading is not a bottleneck. Headers-first could help to implement Electrum-like client for Sia. But that usecase will probably require new RPCs to be added anyway.

modules/consensus/accept.go
starius: Agreed. Note that the order of updating subscribers and the checks is important in the end of this function. I added a comment about this.

modules/consensus/accept.go
```diff
@@ -276,14 +275,14 @@ func (cs *ConsensusSet) managedAcceptBlocks(blocks []types.Block) (bool, error)
 			// See https://github.com/NebulousLabs/Sia/pull/1878#discussion_r123674511
 			changes.RevertedBlocks = append(changes.RevertedBlocks, changeEntry.RevertedBlocks...)
 			changes.AppliedBlocks = append(changes.AppliedBlocks, changeEntry.AppliedBlocks...)
-			chainExtended = true
 		}
 		return nil
 	})
+	chainExtended := len(changes.AppliedBlocks) > 0
```
DavidVorick: ooh, that's clever. It's also correct :)

modules/consensus/accept.go
starius: The `bool` was removed. We want to broadcast only changes that extent the longest chain to prevent DoS through extensions of the genesis blocks. Changes that extent the longest chain result in `nil` error here if all blocks are valid (and if some are invalid, the client switches to another peer).

starius:
Replaced with https://github.com/NebulousLabs/Sia/pull/1994

PR #1884, opened 2017-06-04
synchronize_test: log testcase name

starius:


modules/consensus/synchronize_test.go
```diff
@@ -438,12 +438,12 @@ func TestIntegrationRPCSendBlocks(t *testing.T) {
 		// Create the "remote" peer.
 		remoteCST, err := blankConsensusSetTester(filepath.Join(t.Name()+" - remote", strconv.Itoa(i)))
 		if err != nil {
-			t.Fatal(err)
+			t.Fatalf("%v: %s", tt.msg, err)
```
lukechampine: I think `%s` doesn't work for the `error` type. We typically use `%v` everywhere unless we need specific formatting, e.g. specifying the precision of a float

starius: %s works. From https://golang.org/pkg/fmt/

>  If an operand implements the error interface, the Error method will be invoked to convert the object to a string, which will then be formatted as required by the verb (if any).

But if %v is used in other parts of the code, it makes sense to use it here as well. I'll amend my commit.

lukechampine: interesting. There is another reason to use `%v` though, which is that `%s` produces garbage output if `err == nil`:
```go
fmt.Printf("%s", error(nil))
// output:
// %!s(<nil>)
```

starius: I didn't know that %v helps with the garbage output. It makes sense to use %v.

I amended the commit.

lukechampine:
may want to add the test case index as well, since I notice that the message is identical for some of these.

We should explore using [subtests](https://blog.golang.org/subtests) for all table-driven tests. Doing so essentially forces you to give a name to each test case, and the names are then printed by default.

starius:
> may want to add the test case index as well, since I notice that the message is identical for some of these.

Done

> We should explore using subtests for all table-driven tests.

Looks good to me! I didn't know about subtests, thanks!
I think the transition to subtests should happen in all files at once to maintain consistency.

PR #1889, opened 2017-06-05
use types.SiacoinOutput for /wallet/siacoins param

lukechampine:
The (destination, amount) pairs are now encoded as a JSON array in the query string.
I realized that `types.SiacoinOutput` already has the desired structure, so I used it here. This also cleans up the `SendSiacoinsMulti` implementation.

### DIFF
--- a/api/wallet.go
+++ b/api/wallet.go
@@ -371,6 +371,11 @@ func (api *API) walletSiacoinsHandler(w http.ResponseWriter, req *http.Request, 
 	var txns []types.Transaction
 	if req.FormValue("outputs") != "" {
 		// multiple amounts + destinations
+		if req.FormValue("amount") != "" || req.FormValue("destination") != "" {
+			WriteError(w, Error{"cannot supply both 'outputs' and single amount+destination pair"}, http.StatusInternalServerError)
+			return
+		}
+
 		var outputs []types.SiacoinOutput
 		err := json.Unmarshal([]byte(req.FormValue("outputs")), &outputs)
 		if err != nil {
--- a/api/wallet.go
+++ b/api/wallet.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"encoding/json"
 	"net/http"
 	"path/filepath"
 	"strconv"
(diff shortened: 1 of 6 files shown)


api/wallet.go
```diff
@@ -367,42 +368,39 @@ func (api *API) walletSeedsHandler(w http.ResponseWriter, req *http.Request, _ h
...
-	for i, amountStr := range amountStrs {
-		var ok bool
-		amounts[i], ok = scanAmount(amountStr)
-		if !ok {
-			WriteError(w, Error{"could not read amount from POST call to /wallet/siacoins"}, http.StatusBadRequest)
+	var txns []types.Transaction
+	if req.FormValue("outputs") != "" {
+		// multiple amounts + destinations
+		var outputs []types.SiacoinOutput
+		err := json.Unmarshal([]byte(req.FormValue("outputs")), &outputs)
```
DavidVorick: need to verify before parsing that 'amount' and 'destination' are both empty - don't want a user filling out both and getting confused.

doc/api/Wallet.md
```diff
@@ -314,20 +314,24 @@ dictionary
...
 
-Function: Send siacoins to a set of addresses. The outputs are arbitrarily
-selected from addresses in the wallet. Each address is paired with an amount;
-the number of amounts must match the number of addresses. The number of pairs
-should not exceed 250; this may result in a transaction too large to fit in
-the transaction pool.
+Function: Send siacoins to an address or set of addresses. The outputs are
+arbitrarily selected from addresses in the wallet. If 'outputs' is supplied,
+'amount' and 'destination' must be empty. The number of outputs should not
+exceed 250; this may result in a transaction too large to fit in the
```
DavidVorick: I think we can safely set 250 to 400

doc/api/Wallet.md
```diff
@@ -314,20 +314,24 @@ dictionary
...
+// Number of hastings being sent. A hasting is the smallest unit in Sia. There
+// are 10^24 hastings in a siacoin.
 amount      // hastings
 
-// Comma-separated list of addresses that are receiving coins.
+// Address that is receiving the coins.
 destination // address
+
+// JSON array of outputs. The structure of each output is:
+// {"unlockhash": "<destination>", "value": "<amount>"}
```
DavidVorick: :+1+

PR #1891, opened 2017-06-05
Fix typo

eolszewski:


lukechampine:
This is intentional; `siad` is the daemon that serves our HTTP API. But it's not the only possible "Sia API."

PR #1897, opened 2017-06-07
Implement smux transition

lukechampine:
This does not include switching the handshake to JSON.

I'm rerunning the test that failed (nondeterministically) in #1758 to see if it still fails with smux. If so, we'll tackle it here.


build/version.go
```diff
@@ -7,7 +7,7 @@ import (
 
 const (
 	// Version is the current version of siad.
-	Version = "1.2.1"
+	Version = "1.3.0"
```
lukechampine: this is necessary to make the gateway tests actually use smux

modules/gateway/stream.go
```diff
@@ -0,0 +1,79 @@
+package gateway
+
+import (
+	"net"
+
+	"github.com/NebulousLabs/Sia/build"
+	"github.com/NebulousLabs/muxado"
+	"github.com/xtaci/smux"
```
lukechampine: not sure whether we want to fork this (like muxado) or vendor it (like twofish). I think vendoring is considered best practice now, and a lot of tooling is being written for it, so my inclination is to vendor.

DavidVorick: Vendoring makes Debian maintainers happier too

lukechampine:
update: I've confirmed that the bug is caused by two gateways attempting to connect to each other at the same time. Adding a global mutex to serialize calls to `Connect` fixes the bug. Of course, in real life there isn't any way to prevent this scenario. Instead, we could add a mutex so that the gateway is never both connecting and accepting a connection at the same time. A related problem is that the connection protocol does not allow the gateway to reject a connection after it reads the remote's port number; the remote will assume that the connection was successful, while the other end has already hung up. We should modify the handshake so that *after* all relevant information has been sent, the gateways confirm that they want to continue the connection. I recommend that we do this in tandem with the transition to JSON.

lukechampine:
I suggest that we merge this, after which I'll open a PR that:

- switches handshake to JSON
- changes handshake sequence
- uses UniqueID to prevent double-connecting

The new handshake sequence is:

1. Connecting peer sends JSON-encoded `sessionHeader`, which includes GenesisID, UniqueID, and port number
2. Receiving peer sends a string indicating acceptance, or an error indicating why the connection was rejected
3. If receiving peer accepts, it also sends its own `sessionHeader`
4. Connecting peer responds with accept/reject string

Instead of the `WantConn` boolean, the connecting peer will send an error in step 4 if it just wants to ping instead of a full connection.

Additionally, the gateway will maintain a map of `UniqueID`s (possibly as the key for its `peer` map, instead of the net address). This will be used to prevent double-connect scenarios, especially the recently discovered race where two gateways connect to each other at the same time. Of course, we still need to check the IP/hostname to prevent a single IP (or later, /16 block) from monopolizing the peer list.

DavidVorick:
Double connecting is more trivially fixed with a lock that prevents you from accepting and forming connections at the same time. That lock will also prevent two peers from connecting to eachother at the same time

lukechampine:
Sorry, that is what I meant by "double connecting". I guess there are two situations that could refer to.

I don't think a simple mutex will prevent it. You need to check your peer list for duplicates too, because you can't simply assume that current peers will never try to dial you again (they could be running non-standard code). And if you have the duplicate-checking code, there's little reason to add a new mutex, because the dup check could already use an exclusive lock on the existing mutex.

modules/renter/hostdb/hostweight.go
```diff
@@ -195,9 +195,15 @@ func storageRemainingAdjustments(entry modules.HostDBEntry) float64 {
 // version reported by the host.
 func versionAdjustments(entry modules.HostDBEntry) float64 {
 	base := float64(1)
-	if build.VersionCmp(entry.Version, "1.2.2") < 0 {
+	if build.VersionCmp(entry.Version, "1.4.0") < 0 {
```
DavidVorick: 1.3.1 please

DavidVorick:
Pretty much looks good. I'm expecting to be doing a ton of testing on the interactions between the versions before we merge the gateway branch, so I'm not nearly as worried about testing here.

PR #1898, opened 2017-06-07
Tpool block estimation

DavidVorick:
Making this before it's cleaned up or tested so that the general designs and ideas can be reviewed.

Also because the github diff UI is really nice, heh.

### DIFF
--- a/modules/transactionpool/transactionpool.go
+++ b/modules/transactionpool/transactionpool.go
@@ -13,17 +13,7 @@ import (
 	"github.com/NebulousLabs/Sia/types"
 )
 
-const (
-	dbFilename = "transactionpool.db"
-	logFile    = "transactionpool.log"
-)
-
 var (
-	dbMetadata = persist.Metadata{
-		Header:  "Sia Transaction Pool DB",
-		Version: "0.6.0",
-	}
-
 	errNilCS      = errors.New("transaction pool cannot initialize with a nil consensus set")
 	errNilGateway = errors.New("transaction pool cannot initialize with a nil gateway")
 )
@@ -157,12 +147,12 @@ func (tp *TransactionPool) FeeEstimation() (min, max types.Currency) {
 
 	// Set the minimum fee to the numbers recommended by the blockchain.
 	min = tp.recentMedianFee
-	max = tp.recentMedianFee.Mul64(2)
+	max = tp.recentMedianFee.Mul64(maxMultiplier)
 
 	// Method two: use 'requiredFeesToExtendPool'.
 	required := tp.requiredFeesToExtendTpool()
-	requiredMin := required.MulFloat(1.2) // Clear the local requirement by a little bit.
-	requiredMax := required.MulFloat(3)   // Clear the local requirement by a lot.
+	requiredMin := required.MulFloat(minExtendMultiplier) // Clear the local requirement by a little bit.
+	requiredMax := requiredMin.MulFloat(maxMultiplier)    // Clear the local requirement by a lot.
 	if min.Cmp(requiredMin) < 0 {
 		min = requiredMin
 	}
@@ -174,8 +164,8 @@ func (tp *TransactionPool) FeeEstimation() (min, max types.Currency) {
 	if min.Cmp(TpoolSaneMinFee) < 0 {
 		min = TpoolSaneMinFee
 	}
-	if max.Cmp(TpoolSaneMinFee.Mul64(3)) < 0 {
-		max = TpoolSaneMinFee.Mul64(3)
+	if max.Cmp(TpoolSaneMinFee.Mul64(maxMultiplier)) < 0 {
+		max = TpoolSaneMinFee.Mul64(maxMultiplier)
 	}
 
 	return
--- a/modules/transactionpool/transactionpool.go
+++ b/modules/transactionpool/transactionpool.go
@@ -38,8 +38,16 @@ type (
 
 	// feeSummary desribes the fee structure of a transaction.
 	feeSummary struct {
-		fee  types.Currency // SC per byte
-		size uint64
+		Fee  types.Currency // SC per byte
+		Size uint64
+	}
+
+	// medianPersist is the json object that gets stored in the database so that
+	// the transaction pool can persist its block based fee estimations.
+	medianPersist struct {
+		RecentConfirmedFees []feeSummary
+		TxnsPerBlock        []uint64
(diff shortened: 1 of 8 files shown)


modules/transactionpool/transactionpool.go
```diff
@@ -60,13 +60,17 @@ type (
...
-		blockHeight         types.BlockHeight
 		// TODO: Write a consistency check comparing transactionSets,
 		// transactionSetDiffs.
 		//
 		// TODO: Write a consistency check making sure that all unconfirmedIDs
 		// point to the right place, and that all UnconfirmedIDs are accounted for.
 
+		blockHeight         types.BlockHeight
+		recentConfirmedFees []feeSummary
+		txnsPerBlock        []uint64       // the number of txns in each of the blocks
```
lukechampine: does this grow unbounded? i.e. will it end up with 100k elements? Or does it operate on a much smaller window?

DavidVorick: 10 blocks

modules/transactionpool/update.go
```diff
@@ -33,17 +36,87 @@ func (tp *TransactionPool) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+			tailOffset := uint64(0)
+			for i := 0; i < len(tp.txnsPerBlock)-1; i++ {
+				tailOffset += tp.txnsPerBlock[i]
+			}
+
+			// Strip out all of the transactions in this block.
+			tp.recentConfirmedFees = tp.recentConfirmedFees[:tailOffset]
+			// Strip off the tail offset.
+			tp.txnsPerBlock = tp.txnsPerBlock[:len(tp.txnsPerBlock)-1]
+		}
```
DavidVorick: This feels like a lot of code for it's effective function - pop the most recent block off of the list of blocks.

modules/transactionpool/update.go
```diff
@@ -33,17 +36,87 @@ func (tp *TransactionPool) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+			totalSize += size
+		}
+		// Add an extra zero-fee tranasction for any unused block space.
+		remaining := types.BlockSizeLimit - totalSize
+		tp.recentConfirmedFees = append(tp.recentConfirmedFees, feeSummary{
+			Fee:  types.ZeroCurrency,
+			Size: remaining, // fine if remaining is zero.
+		})
+		// Mark the number of fee transactions in this block.
+		tp.txnsPerBlock = append(tp.txnsPerBlock, uint64(len(block.Transactions)+1))
```
DavidVorick: Same here, feels like a lot of code when all we're doing is adding all of transactions in the recent block into the set of recent transactions.

modules/transactionpool/update.go
```diff
@@ -33,17 +36,87 @@ func (tp *TransactionPool) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+		})
+		// Scroll through the sorted fees until hitting the median.
+		var progress uint64
+		for i := range tp.recentConfirmedFees {
+			progress += tp.recentConfirmedFees[i].Size
+			if progress > (uint64(len(tp.txnsPerBlock))*types.BlockSizeLimit)/2 {
+				tp.recentMedianFee = tp.recentConfirmedFees[i].Fee
+				break
+			}
+		}
```
DavidVorick: This sort happens every block. For values larger than 10 it's pretty unrealistic, because it's going to slow down block validation and increase orphan rate. The huge gain of a linear sort I think is almost worthwhile, but really I'd like to see some sort of algorithm that can do sublinear median finding (modulo some error bound, b/c I think actual sublinear median finding is impossible).

modules/transactionpool/update.go
```diff
@@ -33,17 +36,87 @@ func (tp *TransactionPool) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+			tp.txnsPerBlock = tp.txnsPerBlock[:len(tp.txnsPerBlock)-1]
+		}
 	}
 	for _, block := range cc.AppliedBlocks {
 		if tp.blockHeight > 0 || block.ID() != types.GenesisID {
 			tp.blockHeight++
 		}
 		for _, txn := range block.Transactions {
-			err := tp.addTransaction(tp.dbTx, txn.ID())
+			err := tp.putTransaction(tp.dbTx, txn.ID())
```
johnathanhowell: :+1:, this name is a lot clearer

modules/transactionpool/update.go
```diff
@@ -33,17 +32,87 @@ func (tp *TransactionPool) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+		// applied blocks than reverted blocks, and if there aren't (a height
+		// decreasing reorg), there will be more than 10 applied blocks.
+		if len(tp.txnsPerBlock) > 0 {
+			tailOffset := uint64(0)
+			for i := 0; i < len(tp.txnsPerBlock)-1; i++ {
+				tailOffset += tp.txnsPerBlock[i]
+			}
+
+			// Strip out all of the transactions in this block.
+			tp.recentConfirmedFees = tp.recentConfirmedFees[:tailOffset]
```
lukechampine: it's not clear to me how `recentConfirmedFees` and `tailOffset` are related. Probably want a sanity check to ensure that `tailOffset <= len(tp.recentConfirmedFees)`

modules/transactionpool/update.go
```diff
@@ -33,17 +32,87 @@ func (tp *TransactionPool) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+
+		// Pull the transactions out of the fee summary. For estimating only
+		// over 10 blocks, it is extermely likely that there will be more
+		// applied blocks than reverted blocks, and if there aren't (a height
+		// decreasing reorg), there will be more than 10 applied blocks.
+		if len(tp.txnsPerBlock) > 0 {
+			tailOffset := uint64(0)
+			for i := 0; i < len(tp.txnsPerBlock)-1; i++ {
+				tailOffset += tp.txnsPerBlock[i]
+			}
```
lukechampine: `tailOffset` isn't the best name; not clear what this value represents

modules/transactionpool/update.go
lukechampine: could it be that `recentConfirmedFees` isn't binned by block? Seems like some code would be cleaner if the fees were grouped by block. I admit I don't understand the full design though.

PR #1901, opened 2017-06-09
Clean dialer shutdown in contractor

ChrisSchinnerl:
Pass the StopChan() to the dialer upon creation to stop connection before shutting down

### DIFF
--- a/modules/renter/contractor/formcontract.go
+++ b/modules/renter/contractor/formcontract.go
@@ -104,7 +104,7 @@ func (c *Contractor) managedNewContract(host modules.HostDBEntry, numSectors uin
 	// create transaction builder
 	txnBuilder := c.wallet.StartTransaction()
 
-	contract, err := proto.FormContract(params, txnBuilder, c.tpool)
+	contract, err := proto.FormContract(params, txnBuilder, c.tpool, c.tg.StopChan())
 	if err != nil {
 		txnBuilder.Drop()
 		return modules.RenterContract{}, err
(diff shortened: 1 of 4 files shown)


johnathanhowell:
LGTM

lukechampine:
thanks!

DavidVorick:
Looks good to me, not sure what @lukechampine thinks.

@ChrisSchinnerl a good follow-up to this PR (make a different PR) would be to address all the places it says `time.Second * 15`. In production, we want latencies that are long enough to both support crossing the firewall of China, and also to support nodes running on Tor. 15 seconds dial timeout is not long enough, should be closer to 60.

At the same time, we want much shorter timeouts during testing I think. So we can define a new const in `consts.go` to specify different timeouts depending on dev, standard, and testing (another example already in consts.go). We'd probably want 10 seconds, 60 seconds, and 1 second respectively.

For context, 'dev' constants are intended to be used on closed testnets, where we drop all the constants down low to make it more punishing to the software. The values are generally unsafe for release, but we test them at unsafe levels so that we can feel more confident that, if the tests pass, they'll also still work at the safer levels.

'standard' is of course the values that we use in our production releases.

'testing' is meant for on-same-machine testing, and is primarily geared towards moving things as fast as possible, so those numbers tend to be substantially lower.

ChrisSchinnerl:
@DavidVorick Do you mean only 15 seconds timeouts specifically? I can only find 3 of those in the whole project where 2 of them are only contained in constants (defaultScanSleep, maxScanSleep) and 1 of them is the one in host/negotiate.go:264

DavidVorick:
I was referring specifically to the two timeouts in this PR, but generally speaking any sleeps the program does should have different durations for each build mode.

Most of them already do this, but some of them do not. In production, we should never be making a call with a timeout shorter than 60 seconds, and really if there are multiple round trips in the communication or a lot of data it should be 2 or 3 minutes.

PR #1902, opened 2017-06-09
add timeout for downloading chunks in repair loop

johnathanhowell:


modules/renter/consts.go
```diff
@@ -42,4 +42,13 @@ var (
...
 	}).(int)
+
+	// chunkDownloadTimeout defines the maximum amount of time to wait for a
+	// chunk download to finish before returning in the download-to-upload repair
+	// loop
+	chunkDownloadTimeout = build.Select(build.Var{
+		Dev:      time.Minute,
+		Standard: time.Minute,
+		Testing:  20 * time.Second,
+	}).(time.Duration)
```
lukechampine: is 1 minute a reasonable timeout? Transferring one sector (4MB) per minute is 500kbps.

johnathanhowell: Definitely not. I think 20 minutes would be more reasonable?

modules/renter/consts.go
```diff
@@ -42,4 +42,13 @@ var (
 		Standard: 30,
 		Testing:  60,
 	}).(int)
+
+	// chunkDownloadTimeout defines the maximum amount of time to wait for a
+	// chunk download to finish before returning in the download-to-upload repair
+	// loop
+	chunkDownloadTimeout = build.Select(build.Var{
+		Dev:      15 * time.Minute,
```
DavidVorick: Dev chunks are much smaller, only 256kb iirc. 2 minutes is probably good

PR #1905, opened 2017-06-09
Optimize Block.MerkleRoot by replacing tree.PushObject calls

marcinja:
Optimized MerkleRoot computation for transactionpool tests. Specifically running the profiler for TestFeeEstimation shows ~900 MB used for runtime.convT2E which can be reduced to ~415 MB with the changes.

Before: 
before

After:
after

### DIFF
--- a/types/block.go
+++ b/types/block.go
@@ -3,7 +3,11 @@ package types
 // block.go defines the Block type for Sia, and provides some helper functions
 // for working with blocks.
 
-import "github.com/NebulousLabs/Sia/crypto"
+import (
+	"bytes"
+
+	"github.com/NebulousLabs/Sia/crypto"
+)
 
 const (
 	// BlockHeaderSize is the size, in bytes, of a block header.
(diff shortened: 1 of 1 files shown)


types/block.go
```diff
@@ -108,10 +112,14 @@ func (b Block) ID() BlockID {
...
-		tree.PushObject(payout)
+		var b bytes.Buffer
+		payout.MarshalSia(&b)
+		tree.Push(b.Bytes())
 	}
 	for _, txn := range b.Transactions {
-		tree.PushObject(txn)
+		var b bytes.Buffer
+		txn.MarshalSia(&b)
+		tree.Push(b.Bytes())
```
lukechampine: I think you can do even better here. Create the buffer outside the loop and use `buf.Reset` to clear it. That way, it retains the underlying capacity, so it doesn't need to resize every time you write to it.

marcinja:
Before: correct-before

After: correct-after

runtime.convT2E size still approximately the same (as expected)

PR #1906, opened 2017-06-09
Create hostdb scanning routines on demand

ChrisSchinnerl:
HostDB scanning routines should be created on demand until a max number of threads is reached. They should exit if they are no longer needed.


modules/renter/hostdb/consts.go
```diff
@@ -42,7 +42,7 @@ var (
 
 	// scanningThreads is the number of threads that will be probing hosts for
 	// their settings and checking for reliability.
-	scanningThreads = build.Select(build.Var{
+	maxScanningThreads = build.Select(build.Var{
 		Standard: int(20),
```
DavidVorick: Since they are now spinning up and down, I think we can double this number.

modules/renter/hostdb/scan.go
```diff
@@ -242,40 +250,52 @@ func (hdb *HostDB) managedScanHost(entry modules.HostDBEntry) {
 	hdb.mu.Unlock()
 }
 
+// runs scan on a host
+func (hdb *HostDB) threadedProbeHost(hostEntry modules.HostDBEntry) {
+	// Block the scan until the host is online.
```
DavidVorick: Comment should be 'block until hostdb has internet connectivity'

modules/renter/hostdb/scan.go
```diff
@@ -242,40 +250,52 @@ func (hdb *HostDB) managedScanHost(entry modules.HostDBEntry) {
 	hdb.mu.Unlock()
 }
 
+// runs scan on a host
+func (hdb *HostDB) threadedProbeHost(hostEntry modules.HostDBEntry) {
```
DavidVorick: The comment on a function needs to be a full sentence, and needs to take the form 'funcName blah blah blah...'

modules/renter/hostdb/scan.go
```diff
@@ -242,40 +250,52 @@ func (hdb *HostDB) managedScanHost(entry modules.HostDBEntry) {
...
+	// Block the scan until the host is online.
+	for {
+		hdb.mu.RLock()
+		online := hdb.online
+		hdb.mu.RUnlock()
+		if online {
+			break
+		}
+	}
```
DavidVorick: This is a very tight loop that will lock up the CPU. Need to add in the 30 second soft sleeps that were in the previous implementation​ of this loop.

ChrisSchinnerl:
Please excuse my commit spamming but there was still a potential race condition in my code and it felt a bit hacky ;P I noticed that the scanPool is only used in one function so I thought it is save to remove it from the hostdb and create a new one each time the threads are started to enable a clean shutdown of the pool by closing the channel instead of reading from 3 channels. (1 for shutdown, 1 for cancelling the pool and the scanPool)

modules/renter/hostdb/consts.go
```diff
@@ -59,10 +59,10 @@ var (
...
 	// scanningThreads is the number of threads that will be probing hosts for
 	// their settings and checking for reliability.
-	scanningThreads = build.Select(build.Var{
-		Standard: int(20),
-		Dev:      int(4),
-		Testing:  int(3),
+	maxScanningThreads = build.Select(build.Var{
+		Standard: int(40),
+		Dev:      int(8),
+		Testing:  int(6),
```
DavidVorick: Dev and testing should be kept to 3 and 4 - those numbers are because it's tough to spin up enough hosts to exceed them. Standard is kept to 40 to limit system load (your new code makes it acceptable to raise to 40)

modules/renter/hostdb/scan.go
```diff
@@ -69,12 +73,18 @@ func (hdb *HostDB) queueScan(entry modules.HostDBEntry) {
 			if exists {
 				entry = recentEntry
 			}
+
+			// Create new worker thread
+			if hdb.scanningThreads < maxScanningThreads {
+				hdb.scanningThreads++
+				go hdb.threadedProbeHosts(scanPool)
```
lukechampine: The increment + decrement should be near each other in the source code. So either add the decrement here, like so:
```go
hdb.scanningThreads++
go func() {
    hdb.threadedProbeHosts(scanPool)
    hdb.mu.Lock()
    hdb.scanningThreads--
    hdb.mu.Unlock()
}()
```
or move the increment into `threadedProbeHosts`. I think the former makes it easier to verify correctness. As written, `scanningThreads` will not be properly decremented if `threadedProbeHosts` returns early.

PR #1907, opened 2017-06-09
Optimize Transaction.ID() by adding direct calls to MarshalSia.

marcinja:
Reduces runtime.convT2E memory usage to by ~90MB

With PR #1905 it goes down to ~300MB.

Before:
before
correct-before

After:
txid-after
txid-after-t25

### DIFF
--- a/types/transactions.go
+++ b/types/transactions.go
@@ -7,6 +7,7 @@ package types
 
 import (
 	"errors"
+	"io"
 
 	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/encoding"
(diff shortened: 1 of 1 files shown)


lukechampine:
huh, only 90 MB reduction? What does `--alloc_objects` show?

types/transactions.go
```diff
@@ -143,17 +144,51 @@ type (
...
+	encoding.WriteInt(h, len((t.ArbitraryData)))
+	for i := range t.ArbitraryData {
+		encoding.WritePrefix(h, t.ArbitraryData[i])
+	}
+	encoding.WriteInt(h, len((t.TransactionSignatures)))
+	for i := range t.TransactionSignatures {
+		t.TransactionSignatures[i].MarshalSia(h)
+	}
+	h.Sum(hash[:0])
+	return TransactionID(hash)
```
lukechampine: more direct:
```go
var txid TransactionID
h.Sum(txid[:0])
return txid
```

types/transactions.go
```diff
@@ -143,17 +144,51 @@ type (
...
+	encoding.WriteInt(h, len((t.MinerFees)))
+	for i := range t.MinerFees {
+		t.MinerFees[i].MarshalSia(h)
+	}
+	encoding.WriteInt(h, len((t.ArbitraryData)))
+	for i := range t.ArbitraryData {
+		encoding.WritePrefix(h, t.ArbitraryData[i])
+	}
+	encoding.WriteInt(h, len((t.TransactionSignatures)))
+	for i := range t.TransactionSignatures {
```
lukechampine: typically we use the `for _, x := range xs` form unless we need the index. Although in this case, the readability doesn't matter so much, so `i` is fine.

DavidVorick: I'm pretty sure that `_, x := range` does more memory copying, and slightly more allocation (but mostly copying).

lukechampine: I was skeptical of this; seems like an easy thing to optimize. I tried a microbenchmark:
```go
xs := make([]int, 1e6)
sum := 0
for i := range xs {
	sum += xs[i]
}

// vs.

xs := make([]int, 1e6)
sum := 0
for _, n := range xs {
	sum += n
}
```
The former does in fact run faster. The generated assembly is almost identical except for one section:
```
4f392c: LEAQ 0x1(DI), R8	      
4f3939: MOVQ R8, DI	  

vs:

4f3930: MOVQ 0(AX), R9	        
4f3933: ADDQ $0x8, AX	         
4f3937: INCQ R8	         
```  
It appears that when using the index, the compiler uses `LEAQ` to load the value directly, whereas when using the value, it uses `ADDQ` and `INCQ` to calculate the memory offset. So the number of allocations and the number of copies are the same, but the latter is indeed marginally slower.

DavidVorick: The biggest reason is because when you do

`for i := range myArray`, you get to access each element in `myArray` directly, so there's no copying or anything. When you do

`for _, obj := range myArray`, it allocates new space for `obj` and copies each element of the array into `obj` when you access it inside the loop. So if you modify obj, it won't modify the actual memory in `myArray` (note: if `obj` contains pointers, then you can still change values that will also affect the values seen by `myArray`).

So the primary reason that it's faster (from a high level, anyway) is because you get to skip on 1 allocation of size `obj` and you get to skip on `len(myArray)` copy calls of size `obj`. It doesn't create much GC pressure but it does create a lot of copying issues.

That's also why you can't do this:

```
for _, obj := range myArray {
    go manipulateObj(&obj)
}
```

You'll get a bunch of race conditions because each goroutine is pointing at the same spot in memory.

lukechampine: You can see from the asm that in most cases it doesn't actually allocate or copy anything though. The compiler sees that it can safely use the existing memory. The only reason there's a performance difference is that it computes the memory offset less efficiently (unclear why).

Regardless, clarity comes first, so we should use the value form unless the index is needed or (as in your example) you need a pointer to the value. It's less verbose and communicates the intent more clearly.

marcinja:
Using the --alloc-objects flag:
alloc-obj

lukechampine:
Copying our convo from Slack:

Looks like there are still quite a few calls coming from `SiacoinOutputID`. I notice a lot of code shared between `Transaction.ID()` and `SiacoinOutputID`, `FileContractID`, etc. We should efficiently refactor those. I suggest adding a helper like `(t Transaction) marshalSiaNoSignatures(w io.Writer) error`. Then you can do:
```go
func (t Transaction) ID() (id TransactionID) {
	h := crypto.NewHash()
	t.marshalSiaNoSignatures(h)
	h.Sum(id[:0])
	return
}

func (t Transaction) SiacoinOutputID(i uint64) (id SiacoinOutputID) {
	h := crypto.NewHash()
	h.Write(SpecifierSiacoinOutput[:])
	t.marshalSiaNoSignatures(h)
	encoding.WriteUint64(h, i)
	h.Sum(id[:0])
	return
}
```
etc.

marcinja:
This brings the convT2E memory allocation to about 740MB

PR #1910, opened 2017-06-10
New Block Fee Estimation Algo

DavidVorick:
Bunch of insight while writing up our new transaction fee mechanism for review.

1. There's no point in considering the highest-value fees in each block when guessing the required fee. The ~lowest value fee in each block was enough to get in that block, so that's pretty much all you should be considering. Of course, lowest is a bit dangerous, so we grab the fee located in the 75%-ile for each block, meaning there are about 500kb of transactions with lower fees, and about 1.5mb of transactions with higher fees in the block. This makes malice harder, while also not getting you an unneeded value.

2. Sia makes very heavy use of the child-pays-for-parent scheme. This means that for any set of transactions, the naive median fee is rather likely to be '0', which is not very helpful because it's not actually 0. So this PR includes a scheme that lumps transaction sets together and appropriately considers cp4p.

The result is a new block-based fee estimation that is way better than the one we merged yesterday.

### DIFF
--- a/modules/transactionpool/update.go
+++ b/modules/transactionpool/update.go
@@ -11,6 +11,38 @@ import (
 // findSets takes a bunch of transactions (presumably from a block) and finds
 // all of the separate transaction sets within it. Set does not check for
 // conflicts.
+//
+// The algorithm goes through one transaction at a time. All of the outputs of
+// that transaction are added to the objMap, pointing to the transaction to
+// indicate that the transaction contains those outputs. The transaction is
+// assigned an integer id (each transaction will have a unique id) and added to
+// the txMap.
+//
+// The transaction's inputs are then checked against the objMap to see if there
+// are any parents of the transaction in the graph. If there are, the
+// transaction is added to the parent set instead of its own set. If not, the
+// transaction is added as its own set.
+//
+// The forwards map contains a list of ints indicating when a transaction has
+// been merged with a set. When a transaction gets merged with a parent set, its
+// integer id gets added to the forwards map, indicating that the transaction is
+// no longer in its own set, but instead has been merged with other sets.
+//
+// Some transactions will have parents from multiple distinct sets. If a
+// transaction has parents in multiple distinct sets, those sets get merged
+// together and the transaction gets added to the result. One of the sets is
+// nominated (arbitrarily) as the official set, and the integer id of the other
+// set and the new transaction get forwarded to the official set.
+//
+// TODO: Set merging currently occurs any time that there is a child. But
+// really, it should only occur if the child increases the average fee value of
+// the set that it is merging with (which it will if and only if it has a higher
+// average fee than that set). If the child has multiple parent sets, it should
+// be compared with the parent set that has the lowest fee value. Then, after it
+// is merged with that parent, the result should be merged with the next
+// lowest-fee parent set if and only if the new set has a higher average fee
+// than the parent set. And this continues until either all of the sets have
+// been merged, or until the remaining parent sets have higher values.
 func findSets(ts []types.Transaction) [][]types.Transaction {
 	// txMap marks what set each transaction is in. If two sets get combined,
 	// this number will not be updated. The 'forwards' map defined further on
--- a/modules/transactionpool/update.go
+++ b/modules/transactionpool/update.go
@@ -1,9 +1,9 @@
 package transactionpool
 
 import (
+	"bytes"
 	"sort"
 
-	"github.com/NebulousLabs/Sia/encoding"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/types"
 )
@@ -215,8 +215,11 @@ func (tp *TransactionPool) ProcessConsensusChange(cc modules.ConsensusChange) {
 			// Compile the fees for this set.
 			var feeSum types.Currency
 			var sizeSum int
+			b := new(bytes.Buffer)
 			for _, txn := range set {
-				sizeSum += len(encoding.Marshal(txn))
+				txn.MarshalSia(b)
+				sizeSum += b.Len()
+				b.Reset()
 				for _, fee := range txn.MinerFees {
 					feeSum = feeSum.Add(fee)
 				}
--- a/modules/transactionpool/update.go
+++ b/modules/transactionpool/update.go
@@ -37,16 +37,9 @@ func (tp *TransactionPool) ProcessConsensusChange(cc modules.ConsensusChange) {
 		// over 10 blocks, it is extermely likely that there will be more
 		// applied blocks than reverted blocks, and if there aren't (a height
 		// decreasing reorg), there will be more than 10 applied blocks.
-		if len(tp.txnsPerBlock) > 0 {
-			tailOffset := uint64(0)
-			for i := 0; i < len(tp.txnsPerBlock)-1; i++ {
-				tailOffset += tp.txnsPerBlock[i]
-			}
-
+		if len(tp.recentMedians) > 0 {
 			// Strip out all of the transactions in this block.
-			tp.recentConfirmedFees = tp.recentConfirmedFees[:tailOffset]
-			// Strip off the tail offset.
-			tp.txnsPerBlock = tp.txnsPerBlock[:len(tp.txnsPerBlock)-1]
+			tp.recentMedians = tp.recentMedians[:len(tp.recentMedians)-1]
 		}
 	}
 	for _, block := range cc.AppliedBlocks {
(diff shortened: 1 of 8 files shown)


modules/transactionpool/update.go
```diff
@@ -60,60 +171,69 @@ func (tp *TransactionPool) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+		txnSets := findSets(block.Transactions)
+		for _, set := range txnSets {
+			// Compile the fees for this set.
 			var feeSum types.Currency
-			size := uint64(len(encoding.Marshal(txn)))
-			for _, fee := range txn.MinerFees {
-				feeSum = feeSum.Add(fee)
+			var sizeSum int
+			for _, txn := range set {
+				sizeSum += len(encoding.Marshal(txn))
```
lukechampine: would `sizeSum := len(encoding.Marshal(set))` work? How precise does it need to be? (Encoding `set` will add an 8-byte length prefix to the front.)

modules/transactionpool/update_test.go
```diff
@@ -8,6 +8,134 @@ import (
...
+		t.Fatal("there should be two sets")
+	}
+	lens = make(map[int]struct{})
+	lens[len(sets[0])] = struct{}{}
+	lens[len(sets[1])] = struct{}{}
+	lens[len(sets[2])] = struct{}{}
+	_, exists1 = lens[graph1Size]
+	_, exists2 = lens[graph2Size]
+	_, exists3 = lens[graph3Size]
+	if !exists1 || !exists2 || !exists3 {
```
lukechampine: This is a weird construction. I assume it's because the order of `sets` is undefined? Instead maybe do:
```go
lens := []int{len(sets[0]), len(sets[1]), len(sets[2])}
sort.Ints(lens)
if exp := []int{graph1Size, graph2Size, graph3Size}; !reflect.DeepEqual(lens, exp) {
   t.Fatal("sets have wrong counts: got %v, expected %v", lens, exp)
}
```

lukechampine: also, seems like the test only checks that `findSets` is returning slices of appropriate length. Is there an easy way to inspect the content of the sets and confirm that they match the associated graph?

modules/transactionpool/update.go
```diff
@@ -8,6 +8,124 @@ import (
 	"github.com/NebulousLabs/Sia/types"
 )
 
+// findSets takes a bunch of transactions (presumably from a block) and finds
+// all of the separate transaction sets within it. Set does not check for
+// conflicts.
+func findSets(ts []types.Transaction) [][]types.Transaction {
```
lukechampine: what is defined as a "set" here? Transactions that depend on each other?

modules/transactionpool/update_test.go
DavidVorick: hmm. I think the map solution is fine, it's faster for large sets at the very least (linear vs. n log n). No need to reflect, you could just sort both the graph sizes and the results, and then see if they match.

modules/transactionpool/update.go
DavidVorick: yeah, we take all the transactions that have dependencies and lump them in with their parents.

DavidVorick: That lets us get a median that appropriately accounts for child-pays-for-parent.

modules/transactionpool/update_test.go
lukechampine: Faster? We're talking about n < 1000 here. It could be quadratic for all I care.
`reflect.DeepEqual` is shorthand for "seeing if they match"
Maps also have the issue of duplicate insertions. If you have two sets of the same size, the map won't tell you if only one was inserted. (Granted, you protect against this by checking the length of `sets`, but you don't need to do that if you use `reflect.DeepEqual`.)

modules/transactionpool/update.go
lukechampine: just spitballing, but would this code be any simpler if we mandated that for any transaction set, transaction[i] can only reference transaction[j] if j < i? How easy would it be to detect transactions that didn't conform to this?
I'm wondering if placing a little restriction on txn ordering could reduce the computational load on the transaction pool.

modules/transactionpool/update.go
```diff
@@ -8,6 +8,124 @@ import (
...
+			objMap[ObjectID(fcid)] = tid
+		}
+		for j := range t.FileContractRevisions {
+			fcid := t.FileContractRevisions[j].ParentID
+			objMap[ObjectID(fcid)] = tid
+		}
+		for j := range t.SiafundOutputs {
+			sfoid := t.SiafundOutputID(uint64(j))
+			objMap[ObjectID(sfoid)] = tid
+		}
```
lukechampine: don't these outputs require a number of confirmations before they can be spent?

modules/transactionpool/update.go
lukechampine: I feel like this is a graph theory problem: "given a set of edges, partition the edges into a set of connected graphs". But I can't find any literature about it.

modules/transactionpool/update.go
DavidVorick: not the siafund outputs, only the siacoin claims.

modules/transactionpool/update.go
DavidVorick: needs to be exact, these numbers are used by the miner for building blocks.

We're already crawling through all the transactions to get the miner fees, might as well encode them at the same time. If there's a meaningful performance difference for some reason we can encode the whole set and then subtract 8l.

Though I guess if we want to start being performance sensitive I should probably be doing the MarshalSia trick here to avoid incurring the T2E cost.

modules/transactionpool/update.go
DavidVorick: Ok, I can add a comment explaining the algorithm. I should note that the algorithm is not complete yet, and is also not likely to stay in this package as the miner needs to use the exact same algorithm.

I believe that the way it currently works is a graph theory problem, the solution I have here is linear time which is definitely the best O-notation you can get, as you minimally need to look at all the transaction edges anyway.

The extension that needs to happen here is that we need to split off child subsets with lower fees, and that's going to complicate the code a bit. It'll still be linear time though. Basically, instead of greedily combining sets, we'll only combine sets if the combining increases the average fee value of the parent set.

lukechampine: How does this (pseudocode) algo compare? I believe it is O(n)
```go
// Approach: Map each output to the transaction containing it. Use this to map
// each txn to its children, and to determine which txns are "roots," i.e.
// have no parents within the set. Finally, recursively collect the children
// of each root txn.

// map each output to its txn
var parentIDs map[output]int
for i, txn := range txns {
	for _, output := range txn.outputs {
		parentIDs[output] = i
	}
}

// map each txn to its children
var hasParentInSet []bool
var children []map[int]struct{}
for i, txn := range txns {
	for _, input := range txn.inputs {
		if parent, ok := parentIDs[input]; ok {
			hasParentInSet[i] = true
			children[parent][i] = struct{}{}
		}
	}
}

// identify txns without parents in the set
var noparents []int
for i, hasParents := range hasParentInSet {
	if !hasParents {
		noparents = append(noparents, i)
	}
}

// recursively collect children of each txn in 'noparents'
var sets [][]types.Transaction
for _, i := range noparents {
	set := []types.Transaction{txns[i]}
	toCheck := children[i]
	var seen map[int]bool
	for len(toCheck) > 0 {
		j, toCheck = toCheck[0], toCheck[1:]
		set = append(set, txns[j])
		seen[j] = true
		for child := range children[j] {
			if !seen[child] {
				toCheck = append(toCheck, child)
			}
		}
	}
	sets = append(sets, set)
}
```

modules/transactionpool/update.go
lukechampine: ok, wasn't sure because of this comment:
>```
>// Don't need to add anything for the file contract outputs, storage
>// proof outputs, siafund claim outputs; these outputs are not
>// allowed to be spent until 50 confirmations.
>```

modules/transactionpool/update.go
lukechampine: I doubt the performance impact would be very large. If we need maximum performance we'd want to write a custom `(Transaction).MarshalSiaLen` method that calculates the length without allocating or copying any memory.

modules/transactionpool/update_test.go
DavidVorick: As a pattern, I think it's good to be concerned about performance. Especially now that we are working on improving performance throughout our codebase. There are places where we use hundreds of graphs.

But, you are right, the doubling up thing means that the maps are not as accurate as sorting.

DavidVorick:
Fixed up

PR #1912, opened 2017-06-11
Add consts.go for proto package with connTimeout constant

ChrisSchinnerl:


modules/renter/proto/consts.go
```diff
@@ -0,0 +1,17 @@
...
+	"github.com/NebulousLabs/Sia/build"
+)
+
+var (
+	// connTimeout determines the number of seconds the dialer will wait
+	// for a connect to complete
+	connTimeout = build.Select(build.Var{
+		Dev:      10 * time.Second,
+		Standard: 60 * time.Second,
+		Testing:  1 * time.Second,
```
lukechampine: 1 second appears to be much too short; appveyor tests are failing all over the place. Try bumping it up to 5 or 10.

PR #1914, opened 2017-06-11
Add look ahead safety to wallet

ChrisSchinnerl:
Add look ahead safety to wallet by generating 5000 + 10% more addresses than the current "SeedProgress" and looking for transactions on those addresses. If one of those addresses receives coins the SeedProgress and buffer will be updated accordingly. If one address past the first 1000 in the buffer receives coins a full rescan will be initiated.

### DIFF
--- a/modules/wallet/seed.go
+++ b/modules/wallet/seed.go
@@ -93,20 +93,14 @@ func decryptSeedFile(masterKey crypto.TwofishKey, sf seedFile) (seed modules.See
 	return seed, nil
 }
 
-// maxFutureKeys is a helper function from a given start index
-// which usually is the current primarySeedProgress
-func maxFutureKeys(start uint64) uint64 {
-	return start + 5000 + start/10
-}
-
-// updateFutureKeys creates future keys up to a maximum of maxKeys keys
-func (w *Wallet) updateFutureKeys(start uint64) {
+// regenerateLookahead creates future keys up to a maximum of maxKeys keys
+func (w *Wallet) regenerateLookahead(start uint64) {
 	// Check how many keys need to be generated
-	maxKeys := maxFutureKeys(start)
-	existingKeys := uint64(len(w.futureKeys))
+	maxKeys := maxLookahead(start)
+	existingKeys := uint64(len(w.lookahead))
 
-	for _, k := range generateKeys(w.primarySeed, start+existingKeys, maxKeys-existingKeys) {
-		w.futureKeys[k.UnlockConditions.UnlockHash()] = true
+	for i, k := range generateKeys(w.primarySeed, start+existingKeys, maxKeys-existingKeys) {
+		w.lookahead[k.UnlockConditions.UnlockHash()] = start + existingKeys + uint64(i)
 	}
 }
 
@@ -139,8 +133,8 @@ func (w *Wallet) nextPrimarySeedAddress(tx *bolt.Tx) (types.UnlockConditions, er
 	w.keys[spendableKey.UnlockConditions.UnlockHash()] = spendableKey
 
 	// Remove new key from the future keys and update them according to new progress
-	delete(w.futureKeys, spendableKey.UnlockConditions.UnlockHash())
-	w.updateFutureKeys(progress + 1)
+	delete(w.lookahead, spendableKey.UnlockConditions.UnlockHash())
+	w.regenerateLookahead(progress + 1)
 
 	return spendableKey.UnlockConditions, nil
 }
--- a/modules/wallet/seed.go
+++ b/modules/wallet/seed.go
@@ -93,6 +93,23 @@ func decryptSeedFile(masterKey crypto.TwofishKey, sf seedFile) (seed modules.See
 	return seed, nil
 }
 
+// maxFutureKeys is a helper function from a given start index
+// which usually is the current primarySeedProgress
+func maxFutureKeys(start uint64) uint64 {
+	return start + 5000 + start/10
+}
+
+// updateFutureKeys creates future keys up to a maximum of maxKeys keys
+func (w *Wallet) updateFutureKeys(start uint64) {
+	// Check how many keys need to be generated
+	maxKeys := maxFutureKeys(start)
+	existingKeys := uint64(len(w.futureKeys))
+
+	for _, k := range generateKeys(w.primarySeed, start+existingKeys, maxKeys-existingKeys) {
+		w.futureKeys[k.UnlockConditions.UnlockHash()] = true
+	}
+}
+
 // integrateSeed generates n spendableKeys from the seed and loads them into
 // the wallet.
 func (w *Wallet) integrateSeed(seed modules.Seed, n uint64) {
@@ -120,6 +137,11 @@ func (w *Wallet) nextPrimarySeedAddress(tx *bolt.Tx) (types.UnlockConditions, er
 	// conditions.
 	spendableKey := generateSpendableKey(w.primarySeed, progress)
 	w.keys[spendableKey.UnlockConditions.UnlockHash()] = spendableKey
+
+	// Remove new key from the future keys and update them according to new progress
+	delete(w.futureKeys, spendableKey.UnlockConditions.UnlockHash())
+	w.updateFutureKeys(progress + 1)
+
 	return spendableKey.UnlockConditions, nil
 }
 
--- a/modules/wallet/seed.go
+++ b/modules/wallet/seed.go
@@ -217,10 +217,10 @@ func (w *Wallet) LoadSeed(masterKey crypto.TwofishKey, seed modules.Seed) error 
 	if err := s.scan(w.cs); err != nil {
 		return err
 	}
-	// Add 10% as a buffer because the seed may have addresses in the wild
+	// Add 4% as a buffer because the seed may have addresses in the wild
 	// that have not appeared in the blockchain yet.
-	seedProgress := s.largestIndexSeen + 1
-	seedProgress += seedProgress / 10
+	seedProgress := s.largestIndexSeen + 500
+	seedProgress += seedProgress / 25
 	w.log.Printf("INFO: found key index %v in blockchain. Setting auxiliary seed progress to %v", s.largestIndexSeen, seedProgress)
 
 	err := func() error {
--- a/modules/wallet/update.go
+++ b/modules/wallet/update.go
@@ -10,18 +10,9 @@ import (
 	"github.com/NebulousLabs/bolt"
 )
 
-// historicOutput defines a historic output as recognized by the wallet. This
-// struct is primarily used to sort the historic outputs before inserting them
-// into the bolt database.
-type historicOutput struct {
-	id  types.OutputID
-	val types.Currency
-}
-
 // threadedResetSubscriptions unsubscribes the wallet from the consensus set and transaction pool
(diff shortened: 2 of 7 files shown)


modules/wallet/wallet.go
```diff
@@ -68,8 +68,9 @@ type Wallet struct {
 	// from the seeds, when checking new outputs or spending outputs, the seeds
 	// are not referenced at all. The seeds are only stored so that the user
 	// may access them.
-	seeds []modules.Seed
-	keys  map[types.UnlockHash]spendableKey
+	seeds      []modules.Seed
+	keys       map[types.UnlockHash]spendableKey
+	futureKeys map[types.UnlockHash]bool
```
lukechampine: I'm thinking this should map to the seed index, not a bool, so that you can easily jump forward to the correct index when you see a future key.

modules/wallet/seed.go
```diff
@@ -120,6 +137,11 @@ func (w *Wallet) nextPrimarySeedAddress(tx *bolt.Tx) (types.UnlockConditions, er
 	// conditions.
 	spendableKey := generateSpendableKey(w.primarySeed, progress)
 	w.keys[spendableKey.UnlockConditions.UnlockHash()] = spendableKey
+
+	// Remove new key from the future keys and update them according to new progress
+	delete(w.futureKeys, spendableKey.UnlockConditions.UnlockHash())
+	w.updateFutureKeys(progress + 1)
```
lukechampine: since you operate directly on `w.futureKeys` for the deletion, I think it's cleaner to do the same for the insertion:
```go
nextKey := generateSpendableKey(w.primarySeed, progress+1)
w.futureKeys[nextKey.UnlockConditions.UnlockHash()] = progress + 1
```

modules/wallet/update.go
```diff
@@ -18,10 +18,60 @@ type historicOutput struct {
...
+}
+
 // isWalletAddress is a helper function that checks if an UnlockHash is
-// derived from one of the wallet's spendable keys.
+// derived from one of the wallet's spendable keys or future keys.
 func (w *Wallet) isWalletAddress(uh types.UnlockHash) bool {
 	_, exists := w.keys[uh]
+	if !exists {
+		exists = w.isFutureWalletAddress(uh)
+	}
```
lukechampine: Is `isFutureWalletAddress` going to be useful on its own? If not, I would just inline it here:
```go
func (w *Wallet) isWalletAddress(uh types.UnlockHash) bool {
	_, exists := w.keys[uh]
	if exists {
		return true
	}
	// check future keys
	index, exists := w.futureKeys[uh]
	if exists {
		// advance the lookahead accordingly
		w.advanceSeedLookahead(index)
	}
	return exists
}
```
where `advanceSeedLookahead` adds all the keys in the range `currentSeedProgress -> index` to `w.keys`, deletes them from `w.futureKeys`, and adds all the keys in the range `(currentSeedProgress + lookahead) -> (index + lookahead)` to `w.futureKeys`.
I'm not sure how best to structure the lookahead stuff; it might be more readable to write something like `regenerateLookahead(min, max)` instead. But broadly speaking, I think "lookahead" communicates the purpose more directly than "future keys."

modules/wallet/update.go
```diff
@@ -18,10 +18,60 @@ type historicOutput struct {
...
+		}
+
+		if _, exists := w.keys[uh]; exists {
+			foundKey = true
+			break
+		}
+	}
+
+	// if future key was not among the next 1000 keys rescan the blockchain
+	if !foundKey {
```
lukechampine: under what circumstances would this happen? I figure that if the key isn't in the lookahead, you don't know that it's one of your keys anyway, so how would you know to rescan the blockchain?

In the extreme case (`InitFromSeed`), we basically just want to use a huge lookahead (think 1 million keys) so that the probability of a rescan is as low as possible. Currently, during `InitFromSeed`, we rescan if the highest index seen in the blockchain was in the higher 50% of the keys we generated. I'm not sure what the rescan rules should be for the lookahead, but it will likely be a similar construction. I think in general we can rely on the assumption that addresses will be used in roughly monotonically-increasing order -- that is, keys with a higher index will appear later in the blockchain, and if they don't, the discrepancy will be no greater than ~1000 keys. 

My thinking is: there's a bit of performance cost associated with a large lookahead, but the cost of a rescan is vastly greater. So we should aim for a lookahead that reduces the probability of a rescan as much as possible without impacting performance too severely. To estimate this, it'd be good to have some benchmark numbers showing how long it takes to generate a new key.

Finally, again wrt `InitFromSeed`, we don't want to have to keep using a huge lookahead forever. Ideally, once we were confident that we had found all the uses of the seed, we would drop the lookahead down to a much lower value (i.e. the default). But when can we be sure that we've found all the uses? Naively, we can say that we've seen all uses once we've done a full scan of a synced blockchain. In other words, we should be able to drop down to the default lookahead as soon as the call to `cs.ConsensusSetSubscribe` completes.

lukechampine:
Overall, I'm pleased with how this is progressing. :+1:

modules/wallet/update.go
ChrisSchinnerl: @DavidVorick suggested that the lookahead is 5000 keys + 10% of the current seedProgress and if the key is not found in the first 1000 keys of the lookahead then the blockchain should be scanned again. That's why I included that if statement and also part of the reason for the `updateFutureKeys` function since the size of the lookahead is dynamic due to the 10%.

If that is not necessary I could fill up the futureKeys map to 5000 and only increase it if a new primary seed address is generated.

modules/wallet/update.go
```diff
@@ -18,10 +18,47 @@ type historicOutput struct {
 	val types.Currency
 }
 
+// advanceSeedLookahead is a helper function that checks if an UnlockHash is
+// derived from one of the wallet's lookahead keys. If it is all lookahead keys up to
+// this one will be added to the spendable keys
+func (w *Wallet) advanceSeedLookahead(index uint64) {
```
lukechampine: comment is not accurate

ChrisSchinnerl: sry forgot about that one. I guess I should write comments after the code.

ChrisSchinnerl:
@lukechampine I am not sure about how to proceed
`advanceSeedLookahead` seems to be needed every time `isWalletAddress` is called but it feels wrong to call it in there. It also feels wrong to add an additional return value to `isWalletAddress` and place an `if` after each call to `isWalletAddress` which checks if a lookahead advancement is necessary. 
Another option would probably be to replace direct access to `keys` with a getter which sets the seed accordingly on demand whenever a key is retrieved. 

I am also not sure about the extra 10% of lookahead. I think a constant size might be enough and as you mentioned before it allows for simpler code in `nextPrimarySeedAddress` since only 1 lookahead key would be generated every time it is called. `regenerateLookahead` would still be needed to be called upon wallet unlock though to ensure it is filled.

ChrisSchinnerl:
Do you think the rescan should be initialized in `ProcessConsensusChange` ? Currently it is in `advanceSeedLookahead` which seems simpler to me than passing another return value from `advanceSeedLookahead` to `updateLookahead` and check it in `ProcessConsensusChange`

lukechampine:
it may be *easier*, but I'm not sure it's simpler. The distinction is important. I worry that `advanceSeedLookahead` doesn't imply the side-effect of potentially rescanning. `advanceSeedLookahead` sounds like a helper function, certainly not something I would expect to spawn a goroutine. I generally prefer to spawn goroutines higher up in the call tree, since that's where the decisions are made, rather than inside helper functions.

It's also complicated by the fact that currently, `advanceSeedLookahead` is only called in one place. If we were calling it in multiple places, we'd have a better sense of whether it should be responsible for spawning the rescan goroutine. Until then, we can only rely on our best judgment. My vote is to have `updateLookahead` return a bool indicating whether a rescan is needed and handle it in `ProcessConsensusChange`. We already have a precedent for spawning goroutines in `ProcessConsensusChange`, e.g. [in the contractor](https://github.com/NebulousLabs/Sia/blob/master/modules/renter/contractor/update.go#L65).

ChrisSchinnerl:
Understood, I changed it. Do you have any idea what might cause those race conditions?

lukechampine:
Looks unrelated to this PR. Should be fixed though, I'll make a note of it.

lukechampine:
Ok, this is looking good. I only request one more small change: `TestAdvanceLookaheadNoRescan` should generate the addresses to send to manually, rather than pulling them directly from the wallet's lookahead map. This will ensure, at a high level, that the lookahead code functions as expected when sending to an address generated on another machine.

modules/wallet/update.go
```diff
@@ -10,21 +10,93 @@ import (
...
-// historicOutput defines a historic output as recognized by the wallet. This
-// struct is primarily used to sort the historic outputs before inserting them
-// into the bolt database.
-type historicOutput struct {
-	id  types.OutputID
-	val types.Currency
+// threadedResetSubscriptions unsubscribes the wallet from the consensus set and transaction pool
+// and subscribes again.
+func (w *Wallet) threadedResetSubscriptions() error {
+	if !w.scanLock.TryLock() {
```
DavidVorick: This operation needs to interact with the thread group.

DavidVorick: Looks like both the update functions are missing threadgroup interactions as well.

Right now, if you close the wallet, this will be left running in the background. If the wallet is shutting down correctly, it'll close the log, which means any action which gets logged will cause a panic if there's a subscription running in the background.

DavidVorick: I guess that we don't strictly need to interact with the threadgroup here, actually. Grabbing an `Add` on the threadgroup would stall shutdown until the whole rescan finishes, which is not what we want I think. I'm not sure actually how to get a clean interrupt on the subscribe function, but it seems like something that we would want.

DavidVorick:
This looks good to me. I thought it needed a threadgroup interaction, but it looks like there are actually deeper-seated problems there. I'm going to merge this, and we'll have to remember to dive into the wallet's update + rescan practices with regards to clean shutdown.

I think the wallet is the only module that will unsubscribe and then re-subscribe to the consensus set. None of the other modules ever need to do that, so quick shutdown is unique to the wallet in that sense.

@ChrisSchinnerl a good next pull request I think would be to wrap the update functions of the wallet (receiving consensus set and transaction pool updates) with the threadgroup calls. Those functions should definitely quit if the threadgroup has stopped.

And in a separate pull request (because it's a separate concern) we need some way to subscribe to the consensus set that allows us to interrupt it in the event of early shutdown.

lukechampine:
>A good next pull request I think would be to wrap the update functions of the wallet (receiving consensus set and transaction pool updates) with the threadgroup calls. Those functions should definitely quit if the threadgroup has stopped.

Do you mean adding a `if err := w.tg.Add()` to the top of `ProcessConsensusChange`? None of the other modules currently do this, and now I'm wondering why. Perhaps it's because modules call `cs.Unsubscribe` during shutdown, and that's assumed to be good enough.

ChrisSchinnerl:
Seems like `cs.Unsubscribe` won't be enough since `managedInitializeSubscribe` will call `ProcessConsensusChange` on every missed change before actually adding the subscriber to the list of subscribers. Wrapping it in threadgroup calls might work but `computeConsensusChange` will still be called for every missed change. If that is not costly it might be fine.

PR #1915, opened 2017-06-11
switch to diff-based updates for tpool

DavidVorick:
This PR builds off of #1910, and is an attempt to greatly improve the
performance of Sia when the transaction pool is full. These upgrades will help
to improve the stability of Sia, and also allow us to bump out the transaction
pool to 5mb or even 20mb if the optimizations are robust enough. I'm
continuing to boost performance and test edge cases before we increase the
size of the transaction pool, but this should be ready soon. Even a jump to 5mb
would really help the transaction pool pressure that's on our network, but really
I'm targeting 20mb.

Transaction pool was previously updating subscribers on the new state of the
transaction pool by sending the entire transaction pool to them every single
time a new transaction got added. This was incredibly slow, especially
considering all of the IDs and stuff that was happening behind the scenes.

Now only the diffs are sent whenever there's a transaction pool update, instead
of sending everything. This has resulted in huge speedups that should allow us
to safely increase the transaction pool from 2mb to something larger.

### DIFF
--- a/modules/consensus/validtransaction.go
+++ b/modules/consensus/validtransaction.go
@@ -52,6 +52,8 @@ func validSiacoins(tx *bolt.Tx, t types.Transaction) error {
 		inputSum = inputSum.Add(sco.Value)
 	}
 	if !inputSum.Equals(t.SiacoinOutputSum()) {
+		println(inputSum.String())
+		println(t.SiacoinOutputSum().String())
 		return errSiacoinInputOutputMismatch
 	}
 	return nil
--- a/modules/consensus/validtransaction.go
+++ b/modules/consensus/validtransaction.go
@@ -52,8 +52,6 @@ func validSiacoins(tx *bolt.Tx, t types.Transaction) error {
 		inputSum = inputSum.Add(sco.Value)
 	}
 	if !inputSum.Equals(t.SiacoinOutputSum()) {
-		println(inputSum.String())
-		println(t.SiacoinOutputSum().String())
 		return errSiacoinInputOutputMismatch
 	}
 	return nil
--- a/modules/miner/update.go
+++ b/modules/miner/update.go
@@ -1,6 +1,8 @@
 package miner
 
 import (
+	"sort"
+
 	"github.com/NebulousLabs/Sia/encoding"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/types"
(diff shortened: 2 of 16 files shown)


DavidVorick:
It looks like we still really aren't going fast enough to increase the size of the tpool. I think the rest of the issues here should be cleaner and faster though. I'll keep poking at performance.

modules/miner/update.go
```diff
@@ -52,26 +54,76 @@ func (m *Miner) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+		}
+		delete(m.fullSets, diff.ID)
+	}
+
+	// Split the new sets and add the splits to the list of transactions we pull
+	// form.
+	for _, diff := range diffs {
+		// Skip the reverted diffs, we only care about the applied ones.
+		if diff.Direction == modules.DiffRevert {
+			continue
```
lukechampine: maybe we want separate `Applied` and `Reverted` diffs, like consensus?

modules/wallet/update.go
```diff
@@ -344,66 +344,109 @@ func (w *Wallet) ProcessConsensusChange(cc modules.ConsensusChange) {
...
-	spentSiacoinOutputs := make(map[types.SiacoinOutputID]types.SiacoinOutput)
-	for _, diff := range cc.SiacoinOutputDiffs {
+	// Do the pruning first. If there are any pruned transactions, we will need
+	// to re-allocate the whole processed transactions array.
+	droppedTransactions := make(map[types.TransactionID]struct{})
+	for _, diff := range diffs {
 		if diff.Direction == modules.DiffRevert {
-			// revert means spent
-			spentSiacoinOutputs[diff.ID] = diff.SiacoinOutput
+			txids := w.unconfirmedSets[diff.ID]
```
lukechampine: I don't see a corresponding place where `unconfirmedSets` is written to

DavidVorick: oops. Was causing build failures at least though.

DavidVorick:
Added a stress test. Results of stress test gave me enough confidence to boost mempool baseline to 3mb.

This should be fine on the network. The way the fee gating works is that 3mb is the point where you have to start paying more than 1 SC per kb, which is going to come out to around 1 cent per transaction. At 12mb you'll be paying around 40 cents per transaction, and you'll be in reasonable DoS territory. But you'll have needed to make 24,000 transactions to get there, which is ~$5,000 (if spent optimally) that miners are going to scoop up in about 1 hour.

I'll keep poking at the performance tomorrow and see if we can't give ourselves more margin.

modules/consensus/validtransaction.go
```diff
@@ -52,6 +52,8 @@ func validSiacoins(tx *bolt.Tx, t types.Transaction) error {
 		inputSum = inputSum.Add(sco.Value)
 	}
 	if !inputSum.Equals(t.SiacoinOutputSum()) {
+		println(inputSum.String())
+		println(t.SiacoinOutputSum().String())
```
lukechampine: leftovers

modules/wallet/update.go
```diff
@@ -344,66 +344,109 @@ func (w *Wallet) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+			// only create a ProcessedTransaction if txn is relevant
+			if !relevant {
+				continue
+			}
+
+			pt := modules.ProcessedTransaction{
+				Transaction:           txn,
+				TransactionID:         uts.IDs[i],
+				ConfirmationHeight:    types.BlockHeight(math.MaxUint64),
+				ConfirmationTimestamp: types.Timestamp(math.MaxUint64),
```
lukechampine: I think `math.MaxUint64` is an untyped constant so an explicit cast isn't necessary. NBD

PR #1922, opened 2017-06-12
add /host/estimatescore to estimate the host's HostDB score

johnathanhowell:
This PR adds a new endpoint to the API, `/host/estimatescore`, enabling API consumers to get an idea of their estimated HostDB score with their host's current settings.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -184,6 +184,7 @@ func New(requiredUserAgent string, requiredPassword string, cs modules.Consensus
 		router.GET("/host", api.hostHandlerGET)                                                   // Get the host status.
 		router.POST("/host", RequirePassword(api.hostHandlerPOST, requiredPassword))              // Change the settings of the host.
 		router.POST("/host/announce", RequirePassword(api.hostAnnounceHandler, requiredPassword)) // Announce the host to the network.
+		router.GET("/host/estimatescore", api.hostEstimateScoreGET)
 
 		// Calls pertaining to the storage manager that the host uses.
 		router.GET("/host/storage", api.storageHandler)
--- a/api/api.go
+++ b/api/api.go
@@ -184,7 +184,7 @@ func New(requiredUserAgent string, requiredPassword string, cs modules.Consensus
 		router.GET("/host", api.hostHandlerGET)                                                   // Get the host status.
 		router.POST("/host", RequirePassword(api.hostHandlerPOST, requiredPassword))              // Change the settings of the host.
 		router.POST("/host/announce", RequirePassword(api.hostAnnounceHandler, requiredPassword)) // Announce the host to the network.
-		router.GET("/host/estimatescore", api.hostEstimateScoreGET)
+		router.POST("/host/estimatescore", api.hostEstimateScorePOST)
 
 		// Calls pertaining to the storage manager that the host uses.
 		router.GET("/host/storage", api.storageHandler)
--- a/api/api.go
+++ b/api/api.go
@@ -184,7 +184,7 @@ func New(requiredUserAgent string, requiredPassword string, cs modules.Consensus
 		router.GET("/host", api.hostHandlerGET)                                                   // Get the host status.
 		router.POST("/host", RequirePassword(api.hostHandlerPOST, requiredPassword))              // Change the settings of the host.
 		router.POST("/host/announce", RequirePassword(api.hostAnnounceHandler, requiredPassword)) // Announce the host to the network.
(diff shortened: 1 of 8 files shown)


johnathanhowell:
Latest commits add a `ConversionRate` field, used to indicate how often your host will be contracted by renters given the provided settings.

api/api.go
```diff
@@ -184,6 +184,7 @@ func New(requiredUserAgent string, requiredPassword string, cs modules.Consensus
 		router.GET("/host", api.hostHandlerGET)                                                   // Get the host status.
 		router.POST("/host", RequirePassword(api.hostHandlerPOST, requiredPassword))              // Change the settings of the host.
 		router.POST("/host/announce", RequirePassword(api.hostAnnounceHandler, requiredPassword)) // Announce the host to the network.
+		router.POST("/host/estimatescore", api.hostEstimateScorePOST)
```
lukechampine: shouldn't this be a `GET`?

api/host.go
```diff
@@ -166,40 +167,109 @@ func (api *API) hostHandlerPOST(w http.ResponseWriter, req *http.Request, _ http
...
+
+	// compute the host's conversion rate, that is, how likely it is to be
+	// selected by renters for use in contracts, by finding its position in the
+	// sorted list of active hosts
+	hosts := api.renter.ActiveHosts()
+	if !exists {
+		hosts = append(hosts, entry)
+	}
+	sort.Slice(hosts, func(i, j int) bool {
+		return api.renter.ScoreBreakdown(hosts[i]).Score.Cmp(api.renter.ScoreBreakdown(hosts[j]).Score) == 1
```
lukechampine: use `>= 0` instead of `== 1`

api/host.go
```diff
@@ -166,40 +167,109 @@ func (api *API) hostHandlerPOST(w http.ResponseWriter, req *http.Request, _ http
...
 	}
 
-	err := api.host.SetInternalSettings(settings)
+	return settings, nil
+}
+
+// hostEstimateScoreGET handles the POST request to /host/estimatescore and
+// computes an estimated HostDB score for the provided settings.
+func (api *API) hostEstimateScorePOST(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	entry, exists := api.renter.Host(api.host.PublicKey())
```
lukechampine: what is the purpose of fetching `entry`? Don't you basically overwrite the whole entry later?

johnathanhowell: We overwrite the settings with the previous settings combined with the settings provided to the endpoint, but the uptime is left unchanged. This is so the estimate takes into account your current uptime.

Come to think of it, that does actually pose a problem. We learned recently that certain network configurations do not allow you to probe your own external IP address, meaning that users with that problem would have an unfairly low estimation of their score.

Not sure what the best way of dealing with that is, we could simply ignore uptime and not fetch the entry.

lukechampine: yeah, I think uptime should be ignored, since it can't be accurately measured. But then how do we estimate the ranking? Perhaps the uptime of every host should be set to some sane value before their score is calculated.

S-anasol: Or maybe add option to set (expected) uptime in %? 👨‍🚀 
Optional ofc.

modules/renter/hostdb/hostweight.go
```diff
@@ -360,6 +361,64 @@ func (hdb *HostDB) calculateHostWeight(entry modules.HostDBEntry) types.Currency
...
+	sort.Slice(rankedHosts, func(i, j int) bool {
+		return rankedHosts[i].score.Cmp(rankedHosts[j].score) >= 0
+	})
+	var rank float64
+	for i, host := range rankedHosts {
+		if host.pk == entry.PublicKey.String() {
+			rank = float64(i)
+		}
+	}
+	conversionRate := math.Max(100-((rank/50)*100), 0)
```
lukechampine: 50 is kind of magic here

api/host_test.go
```diff
@@ -49,7 +51,115 @@ var (
...
+
+	tests := []struct {
+		price          types.Currency
+		conversionRate float64
+	}{
+		{types.SiacoinPrecision, 100},
+		{types.SiacoinPrecision.Mul64(30000), 98},
+		{types.SiacoinPrecision.Mul64(50000), 96},
+		{types.SiacoinPrecision.Mul64(70000), 94},
+		{types.SiacoinPrecision.Mul64(60000000), 94},
```
lukechampine: I'm assuming this test will break if we change the scoring algo. Kind of annoying, but not much we can do about it for now.

PR #1935, opened 2017-06-14
Fast method for calculating transaction size

lukechampine:
The transaction pool makes frequent use of the size of encoded transactions. Currently, we determine the size of a transaction by writing its encoded form to a buffer and returning the length of the buffer. Even though the transaction type has a custom `MarshalSia` method, this is still wasteful; it is slow and puts pressure on the garbage collector.

The transaction type (and a few others) now have a `MarshalSiaSize` method that returns the encoded size without encoding anything. These methods are essentially just iterated arithmetic, so they run very fast. Benchmarks for encoding `types.GenesisBlock.Transactions[0]`:

```
BenchmarkTransactionMarshalSiaSize-4   	2000000     918 ns/op      0 B/op     0 allocs/op
BenchmarkTransactionMarshalSiaLen-4    	 100000   12078 ns/op  11144 B/op   205 allocs/op
```
(`BenchmarkTransactionMarshalSiaLen` calls `len(encoding.Marshal(txn))`)
For small transactions, the effect is more pronounced: up to 40x faster.

I also added `MarshalSia` methods for `types.FileContract` and `types.FileContractRevision`. This definitely speeds up the encoding of a transaction, but the effect isn't visible when encoding the genesis transaction (since it doesn't contain any file contracts).

### DIFF
--- a/types/encoding.go
+++ b/types/encoding.go
@@ -72,6 +72,22 @@ func (cf CoveredFields) MarshalSia(w io.Writer) error {
 	return nil
 }
 
+// MarshalSiaSize returns the encoded size of cf.
+func (cf CoveredFields) MarshalSiaSize() (size int) {
+	size += 1 // WholeTransaction
+	size += 8 + len(cf.SiacoinInputs)*8
+	size += 8 + len(cf.SiacoinOutputs)*8
+	size += 8 + len(cf.FileContracts)*8
+	size += 8 + len(cf.FileContractRevisions)*8
+	size += 8 + len(cf.StorageProofs)*8
+	size += 8 + len(cf.SiafundInputs)*8
+	size += 8 + len(cf.SiafundOutputs)*8
+	size += 8 + len(cf.MinerFees)*8
+	size += 8 + len(cf.ArbitraryData)*8
+	size += 8 + len(cf.TransactionSignatures)*8
+	return
+}
+
 // MarshalJSON implements the json.Marshaler interface.
 func (c Currency) MarshalJSON() ([]byte, error) {
 	// Must enclosed the value in quotes; otherwise JS will convert it to a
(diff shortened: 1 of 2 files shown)


types/encoding.go
```diff
@@ -355,6 +479,62 @@ func (t Transaction) MarshalSia(w io.Writer) error {
 	return nil
 }
 
+// MarshalSiaSize returns the encoded size of t.
+func (t Transaction) MarshalSiaSize() (size int) {
+	size += 8
+	for _, sci := range t.SiacoinInputs {
+		size += len(sci.ParentID)
+		size += sci.UnlockConditions.MarshalSiaSize()
+	}
```
lukechampine: I tried manually unrolling these like so:
```go
size += len(t.SiacoinInputs) * len(SiacoinOutputID{})
for _, sci := range t.SiacoinInputs {
    size += sci.UnlockConditions.MarshalSiaSize()
}
// etc.
```
but there was no difference in the benchmarks. It seems that the compiler optimizes this sort of code well.

types/encoding.go
```diff
@@ -103,6 +119,31 @@ func (c Currency) MarshalSia(w io.Writer) error {
...
+	const (
+		_m    = ^big.Word(0)
+		_logS = _m>>8&1 + _m>>16&1 + _m>>32&1
+		_S    = 1 << _logS // number of bytes per big.Word
+	)
+
+	// start with the number of Words * number of bytes per Word, then
+	// subtract trailing bytes that are 0
+	bits := c.i.Bits()
+	size := len(bits) * _S
```
lukechampine: I also tried calculating the size like so:
```go
bitlen := c.i.BitLen()
size := (bitlen / _S)
if bitlen%_S != 0 {
	size++
}
return 8 + size
```
but surprisingly, this is slower (10ns vs 6ns), and the slowness isn't caused by the modulo. I guess `BitLen` does needless work calculating the exact number of bits, when really we just care about the number of bytes.

DavidVorick:
Is the tpool len stuff merged? I think not. Will wait on this until it is, then would like to see a probabilistic debug sanity check that compares the size of an encoded transaction to the size of the len call.

To verify correctness of the new MarshalSia calls, we probably only need to ibd. If they are wrong, I really doubt ibd on the main chain will complete.

PR #1937, opened 2017-06-15
Update the way miner selects transactions out of the tpool

marcinja:


modules/consensus/validtransaction.go
```diff
@@ -52,6 +52,8 @@ func validSiacoins(tx *bolt.Tx, t types.Transaction) error {
 		inputSum = inputSum.Add(sco.Value)
 	}
 	if !inputSum.Equals(t.SiacoinOutputSum()) {
+		println(inputSum.String())
+		println(t.SiacoinOutputSum().String())
```
DavidVorick: leftover printlns

modules/miner/miner.go
```diff
@@ -149,6 +168,23 @@ func New(cs modules.ConsensusSet, tpool modules.TransactionPool, w modules.Walle
 		arbDataMem: make(map[types.BlockHeader][crypto.EntropySize]byte),
 		headerMem:  make([]types.BlockHeader, HeaderMemory),
 
+		fullSets:  make(map[modules.TransactionSetID][]int),
+		splitSets: make(map[splitSetID]*splitSet),
+
+		blockMapHeap: &mapHeap{
```
DavidVorick: no space here - when initializing structs we try to match the spacing of their definitions.

modules/miner/miner.go
```diff
@@ -149,6 +168,23 @@ func New(cs modules.ConsensusSet, tpool modules.TransactionPool, w modules.Walle
...
+			selectID: make(map[splitSetID]*mapElement),
+			data:     make([]*mapElement, 0),
+			size:     uint64(0),
+			minHeap:  true,
+		},
+
+		overflowMapHeap: &mapHeap{
```
DavidVorick: no space here either

modules/miner/splitsetheap.go
```diff
@@ -0,0 +1,213 @@
...
+		// Swap i and j, then continue.
+		mh.Swap(i, j)
+		j = i
+	}
+}
+
+// down maintains the heap condition by checking that the children of the element
+// at index i are less than the element at i (as defined by Less()). If so,
+// it swaps them, and continues down the heap until the heap condition is satisfied.
+func (mh *mapHeap) down(i0, n int) bool {
```
DavidVorick: typically we put unexported functions at the top of the file

modules/miner/splitsetheap.go
```diff
@@ -0,0 +1,213 @@
...
+	if !mh.down(i, mh.Len()) {
+		mh.up(i)
+	}
+}
+
+// up maintains the heap condition by checking if the element at index j is
+// less than its parent (as defined by Less()). If so it swaps them, so that the
+// element at index j goes 'up' the heap. It continues until the heap condition
+// is satisfied again.
+func (mh *mapHeap) up(j int) {
```
DavidVorick: unexported functions above the exported functions

modules/miner/splitsetheap.go
DavidVorick: * above the exported functions, but below the consts, variables, and types

modules/wallet/update.go
```diff
@@ -344,66 +344,109 @@ func (w *Wallet) ProcessConsensusChange(cc modules.ConsensusChange) {
...
-			pt.Outputs = append(pt.Outputs, modules.ProcessedOutput{
-				ID:             types.OutputID(txn.SiacoinOutputID(uint64(i))),
-				FundType:       types.SpecifierSiacoinOutput,
-				MaturityHeight: types.BlockHeight(math.MaxUint64),
-				WalletAddress:  w.isWalletAddress(sco.UnlockHash),
-				RelatedAddress: sco.UnlockHash,
-				Value:          sco.Value,
-			})
+	// Scroll through all of the diffs and add any new transactions.
+	for _, uts := range diff.AppliedTransactions {
```
DavidVorick: uts needs to be renamed, I think it stands for 'unconfirmedTransactionSet' or something, but that's not clear. (I think I wrote this code, but it still needs to be fixed)

modules/miner/splitsetheap.go
```diff
@@ -0,0 +1,213 @@
+package miner
+
+// mapElements are stored in a mapHeap.
+// The index refers to the location of the splitSet in the underlying slice
+// used to represent the heap.
+type mapElement struct {
```
DavidVorick: This can all be one block, no need for that newline after the first sentence.

modules/miner/splitsetheap.go
```diff
@@ -0,0 +1,213 @@
...
+	set   *splitSet
+	id    splitSetID
+	index int
+}
+
+// MapHeap is a heap of splitSets (compared by averageFee).
+// The minHeap bool specifies whether it is a min-heap or max-heap.
```
DavidVorick: Same here, can all be one block with no newlines.

modules/miner/splitsetheap_test.go
```diff
@@ -0,0 +1,450 @@
...
+		selectID: make(map[splitSetID]*mapElement),
+		data:     make([]*mapElement, 0),
+		size:     0,
+		minHeap:  true,
+	}
+
+	max.Init()
+	min.Init()
+
+	for i := 0; i < 1000; i++ {
```
DavidVorick: for a more valid test, you want to add these elements randomly.

you can import the `github.com/NebulousLabs/fastrand` package and use the `Perm` function. `Perm(1000)` will give you 1000 integers in the range [0,1000), but they will be in random order. Then you can range through that list to get random heap insertion here.

modules/miner/splitsetheap_test.go
```diff
@@ -0,0 +1,450 @@
...
+		if minPop.set.averageFee.Cmp(types.SiacoinPrecision.Mul64(uint64(i))) != 0 {
+			t.Error("Unexpected currency value in result from min-heap pop.")
+		}
+	}
+}
+
+// TestMapHeapSimpleDiffOrder performs the same test as TestMapHeapSimple but
+// pushes the half of elements with larger averageFee values first, and the
+// smaller sized fee elements last. Then the sequence of pops is checked.
+func TestMapHeapSimpleDiffOrder(t *testing.T) {
```
DavidVorick: I believe using `Perm` on the above tests removes the need for this test.

modules/miner/splitsetheap_test.go
```diff
@@ -0,0 +1,450 @@
...
+		size:     0,
+		minHeap:  true,
+	}
+
+	max.Init()
+	min.Init()
+
+	var expectedSize uint64
+
+	for i := 0; i < 1000; i++ {
```
DavidVorick: same here, put the min in random order.

modules/miner/splitsetheap_test.go
```diff
@@ -0,0 +1,450 @@
...
+			t.Error("Unexpected set size in result from max-heap pop.")
+		}
+
+		if minPop.set.size != uint64(100*i) {
+			t.Log(i)
+			t.Error("Unexpected set size in result from min-heap pop.")
+		}
+
+	}
+}
```
DavidVorick: Would be nice if this test had some stage where you fill the heap to like 100 elements, then you popped 10, pushed 10, popped 5, pushed 5, etc. This makes sure that the heap is not merely staying consistent simply because you are pushing all in a row and then popping all in a row.

modules/miner/splitsetheap_test.go
```diff
@@ -0,0 +1,450 @@
...
+
+// TestMapHeapRemoveBySetID pushes a sequence of elements onto
+// a max-heap and min-heap. Then it removes a random element using
+// its splitSetID, and checks that it has been removed.
+func TestMapHeapRemoveBySetID(t *testing.T) {
```
DavidVorick: Looks like you do not check that the size of the heap stays consistent when removing by id. Need to be testing for all of the stateful conditions of the heap when testing specific operations.

modules/miner/splitsetheap_test.go
```diff
@@ -0,0 +1,450 @@
...
+	}
+	if inMaxMap {
+		t.Error("Element found in max map(s) after being removed by splitSetID")
+	}
+
+}
+
+// TestMapHeapPeek test the Peek method. First, on an empty heap Peek should
+// return false. Then it checks that Peek returns the same result as the next Pop.
+func TestMapHeapPeek(t *testing.T) {
```
DavidVorick: same here, should be pretty easy to tag on a check that the heap size does not change from a peek.

modules/miner/update.go
```diff
@@ -1,11 +1,205 @@
 package miner
 
 import (
-	"github.com/NebulousLabs/Sia/encoding"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/types"
 )
 
+func (m *Miner) getNewSplitSets(diff *modules.TransactionPoolDiff) []*mapElement {
```
DavidVorick: function needs a docstring

modules/miner/update.go
```diff
@@ -1,11 +1,205 @@
...
+			}
+			break
+		}
+	}
+	delete(m.splitSets, id)
+}
+
+// Change the UnsolvedBlock so that it  has exactly those transactions in the
+// blockMapHeap.
+func (m *Miner) adjustUnsolvedBlock() {
```
DavidVorick: functions need to be sorted alphabetically

modules/miner/update.go
```diff
@@ -1,11 +1,205 @@
...
+}
+
+// addNewTxns adds new unconfirmed transactions to the miner's transaction
+// selection.
+func (m *Miner) addNewTxns(diff *modules.TransactionPoolDiff) {
+	// Get new splitSets (in form of mapElement)
+	newElements := m.getNewSplitSets(diff)
+
+	// Place each elem in one of the MapHeaps
+	for _, elem := range newElements {
```
DavidVorick: more efficient to use the construction `for i := 0; i < len(newElements); i++ {` - does less memory copying.

modules/miner/update.go
```diff
@@ -1,11 +1,205 @@
...
+	// Check if removing sets from the blockMapHeap will be worth it.
+	// bottomSets will hold  the lowest fee sets from the blockMapHeap
+	bottomSets := make([]*mapElement, 0)
+	var sizeOfBottomSets uint64
+	var averageFeeOfBottomSets types.Currency
+
+	// While the heap cannot fit this set s, and while the (weighted)
+	// average fee for the lowest sets from the block is less than
+	// the fee for the set s, continue removing from the heap.
+	for {
```
DavidVorick: general design here - it's not immediately obvious that this loop is going to exit. It's also not clear that the variables declared above the loop are part of the loop state, because they are separated by a newline. Should take the form:

```
// comment
vars := myVars
for {
```

In loop design, you want to try to make it obvious that the loop is going to exit.

modules/miner/update.go
```diff
@@ -1,11 +1,205 @@
...
+		_, exists := m.blockMapHeap.Peek()
+		// If the blockMapHeap is empty, push all elements removed from it
+		// back in, and place the candidate set into the overflow.
+		if !exists {
```
DavidVorick: I think you can go ahead and wrap this in a sanity check. Blocks are 2mb, transaction sets aren't going to be larger than 250kb. You should never empty the block trying to figure out if a txn set is going to fit.

modules/miner/update.go
```diff
@@ -1,11 +1,205 @@
...
+			// Put back in transactions removed.
+			for _, v := range bottomSets {
+				m.blockMapHeap.Push(v)
+			}
+			// Go to the next candidate set.
+			break
+		}
+		nextSet := m.blockMapHeap.Pop()
+
+		bottomSets = append(bottomSets, nextSet)
```
DavidVorick: seems like it's one idea to fetch the next set and add it to the bottom set.

Also really seems like one idea to check if there is a set, then grab it, then add it to the bottom set. So this can probably all be one block of code.

modules/miner/update.go
```diff
@@ -1,11 +1,205 @@
...
+
+		// Calculating fees to compare total fee from those sets removed
+		// and the current set s.
+		totalFeeFromNextSet := nextSet.set.averageFee.Mul64(nextSet.set.size)
+		totalBottomFees := averageFeeOfBottomSets.Mul64(sizeOfBottomSets).Add(totalFeeFromNextSet)
+		sizeOfBottomSets += nextSet.set.size
+		averageFeeOfBottomSets := totalBottomFees.Div64(sizeOfBottomSets)
```
DavidVorick: An average is going to have rounding errors. Instead, you should keep the total size and total fees for both as the state, and then compute the total average when you need it.

DavidVorick: I guess that means we need to change the splitSet definition. Hopefully that doesn't disrupt too much of the code

DavidVorick: You should probably check, before doing anything else, whether the average fee of the set you popped is higher than the average fee of the set you have to put on. Because if it is, you can stop execution right there, you know that it's not going to work in your favor anymore.

DavidVorick: oh. We don't need to change the splitSet definition, you shouldn't ever be merging average fees at all. Though technically more optimal to do it this way, you create weird stateful situations and you might end up at local minima.

modules/miner/update.go
```diff
@@ -1,11 +1,205 @@
...
+
+		// If the average fee of the bottom sets from the block is higher
+		// than the fee from this candidate set, put the candidate into the
+		// overflow MapHeap.
+		if averageFeeOfBottomSets.Cmp(candidateSet.averageFee) == 1 {
```
DavidVorick: Things get convoluted I think if you focus primarily on the average fee of the bottom sets as a whole. Instead, you should allow it to be memoryless. Pop the lowest fees of a block until you either find a higher average fee for just the one set (at which point you give up), or until you have enough room to add the candidate set, and then you drop all the fees at once.

This way you can preserve the property that the lowest average fee in the block is higher than the highest average fee in the leftovers.

modules/miner/update.go
```diff
@@ -1,11 +1,205 @@
...
+			// Put transaction sets from bottom back into the blockMapHeap.
+			for _, v := range bottomSets {
+				m.blockMapHeap.Push(v)
+			}
+			// Go to the next candidate set.
+			break
+		}
+
+		// Check if the candidateSet can fit in the block.
+		if m.blockMapHeap.size-sizeOfBottomSets+candidateSet.size < types.BlockSizeLimit-5e3 {
```
DavidVorick: This condition should probably lead the `for` loop. Then you can drop the starter code you have at the top. You just start the loop right away by checking if there's enough room in the block, then you push it in, handle all the bottom sets (of which there may be zero), and then return. This makes it obvious that the for loop has a plain exit condition.

Next piece of the `for` loop is the sanity check followed by the alternative condition. Sanity check verifies that the block has transactions to pop off, and alternative condition checks that the next element in the block has a lower average fee than the transaction you are trying to put in the block. If it doesn't, you restore the block structure and exit.

I think that would overall be a lot cleaner approach to this function.

modules/miner/update.go
```diff
@@ -1,11 +1,205 @@
...
+		m.overflowMapHeap.RemoveSetByID(id)
+		return
+	}
+
+	if inBlockMapHeap {
+		// Remove from blockMapHeap.
+		m.blockMapHeap.RemoveSetByID(id)
+
+		// Promote sets from overflow heap to block if possible.
+		for {
```
DavidVorick: don't default to infinite loops. Try to structure your loops so the exit condition is in the `for` declaration.

modules/miner/update.go
```diff
@@ -1,11 +1,205 @@
...
+			overflowElem, canPromote := m.overflowMapHeap.Peek()
+			if canPromote && m.blockMapHeap.size+overflowElem.set.size < types.BlockSizeLimit-5e3 {
```
DavidVorick: right here you have the exit condition - it's the inverse of this `if` statement. Just move it up into the loop declaration.

modules/miner/update.go
```diff
@@ -1,11 +1,205 @@
...
+		// as well.
+		splitSetIndexes := m.fullSets[id]
+		for _, ss := range splitSetIndexes {
+			m.deleteMapElementTxns(splitSetID(ss))
+		}
+		delete(m.fullSets, id)
+	}
+}
+
+func (m *Miner) deleteMapElementTxns(id splitSetID) {
```
DavidVorick: function needs a docstring

modules/miner/update.go
```diff
@@ -1,11 +1,205 @@
...
+			overflowElem, canPromote := m.overflowMapHeap.Peek()
+			if canPromote && m.blockMapHeap.size+overflowElem.set.size < types.BlockSizeLimit-5e3 {
+				promotedElem := m.overflowMapHeap.Pop()
+				m.blockMapHeap.Push(promotedElem)
+				continue
+			}
+			break
+		}
+	}
+	delete(m.splitSets, id)
```
DavidVorick: This line of code is not reached if `inOverflowMapHeap`, is that a bug?

DavidVorick: perhaps this should be set up at the top as a defer statement

DavidVorick: or maybe the `inOverflow/Block` blocks need to be an `if else` instead of consecutive `if`

modules/miner/update.go
```diff
@@ -1,11 +1,205 @@
...
+		m.persist.UnsolvedBlock.Transactions = make([]types.Transaction, 0, newCap)
+	} else {
+		m.persist.UnsolvedBlock.Transactions = m.persist.UnsolvedBlock.Transactions[:0]
+	}
+
+	for _, elem := range m.blockMapHeap.selectID {
+		set := elem.set
+		m.persist.UnsolvedBlock.Transactions = append(m.persist.UnsolvedBlock.Transactions, set.transactions...)
+	}
+}
```
DavidVorick: This function doesn't have any comments.

DavidVorick: Overall the logic seems sound though.

modules/transactionpool/transactionpool_test.go
```diff
@@ -336,3 +336,327 @@ func TestFeeEstimation(t *testing.T) {
...
+	// Mine blocks until the tpool is gone.
+	for tpt.tpool.transactionListSize > 0 {
+		_, err := tpt.miner.AddBlock()
+		if err != nil {
+			t.Fatal(err)
+		}
+	}
+}
+
+func TestHeapFees(t *testing.T) {
```
DavidVorick: Even tests need docstrings

modules/transactionpool/transactionpool_test.go
```diff
@@ -336,3 +336,327 @@ func TestFeeEstimation(t *testing.T) {
...
+		t.SkipNow()
+	}
+	tpt, err := createTpoolTester(t.Name())
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer tpt.Close()
+
+	// Mine a few more blocks to get some extra funding.
+	for i := 0; i < 5; i++ {
```
DavidVorick: Try needing less than 5, it takes a lot of time to make a block when you are in testing mode. Lots of safety checks to verify that consensus doesn't have any bugs. If you do need a full five, explain why in the comment.

modules/transactionpool/transactionpool_test.go
```diff
@@ -336,3 +336,327 @@ func TestFeeEstimation(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+	}
+	// Make a bunch of transactions with high fees. A bunch with low fees.
+	// See that the next block has all high fee txns.
+	// Iterate over block txns and add up fees
+	// Prepare a bunch of outputs for a series of graphs to fill up the
+	// transaction pool.
+	coinFrac := types.SiacoinPrecision
```
DavidVorick: Comments should apply to code blocks, this comment + code block is completely disconnected from whatever is actually happening.

modules/transactionpool/transactionpool_test.go
```diff
@@ -336,3 +336,327 @@ func TestFeeEstimation(t *testing.T) {
...
+	}
+
+	// Mine the graph setup in the consensus set so that the graph outputs are
+	// transaction sets.
+	_, err = tpt.miner.AddBlock()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	finalTxn := txns[len(txns)-1]
```
DavidVorick: I'd probably the three code blocks above into one, it's all the same idea that we're creating a bunch of setup outputs to use for the graph.

modules/transactionpool/transactionpool_test.go
```diff
@@ -336,3 +336,327 @@ func TestFeeEstimation(t *testing.T) {
...
+		err := tpt.tpool.AcceptTransactionSet([]types.Transaction{graph[0]})
+		if err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	block, err := tpt.miner.AddBlock()
+	if err != nil {
+		t.Fatal(err)
+	}
```
DavidVorick: should merge this with the block above and explain that you are confirming the head of each graph so that the rest of the transactions in the graph will be viewed as individual transactions.

modules/transactionpool/transactionpool_test.go
```diff
@@ -336,3 +336,327 @@ func TestFeeEstimation(t *testing.T) {
...
+	avgFee1 := totalFee1.Div64(uint64(numTxns1))
+
+	// Mine the next block so we can check the transactions inside
+	block, err = tpt.miner.AddBlock()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	var totalFee2 types.Currency
+	//var expectedFee types.Currency // Calculate this
```
DavidVorick: probably should not have this commented out

DavidVorick:
Overall I would say that this PR looks pretty good, but a lot of the code style could be cleaner. All the comments should help figure out what needs to be improved, and then we can get this code merged.

lukechampine:
why are david's commits included in this branch? Does it need to be rebased onto master?

modules/miner/miner.go
```diff
@@ -85,9 +87,11 @@ type Miner struct {
...
 
 	// Transaction pool variables.
-	fullSets   map[modules.TransactionSetID][]int
-	setCounter int
-	splitSets  map[int]*splitSet
+	fullSets        map[modules.TransactionSetID][]int
+	setCounter      int
+	splitSets       map[splitSetID]*splitSet
+	blockMapHeap    *mapHeap
+	overflowMapHeap *mapHeap
```
DavidVorick: need to be organized alphabetically if they are all clustered together.

modules/miner/miner.go
```diff
@@ -165,7 +169,19 @@ func New(cs modules.ConsensusSet, tpool modules.TransactionPool, w modules.Walle
 		headerMem:  make([]types.BlockHeader, HeaderMemory),
 
 		fullSets:  make(map[modules.TransactionSetID][]int),
-		splitSets: make(map[int]*splitSet),
+		splitSets: make(map[splitSetID]*splitSet),
+		blockMapHeap: &mapHeap{
+			selectID: make(map[splitSetID]*mapElement),
+			data:     make([]*mapElement, 0),
+			size:     uint64(0),
```
DavidVorick: The `uint64` cast is implicit, and isn't needed. You can just set it to 0.

The zero is also implicit, and isn't needed. You can just omit this line altogether.

modules/miner/miner.go
```diff
@@ -165,7 +169,19 @@ func New(cs modules.ConsensusSet, tpool modules.TransactionPool, w modules.Walle
...
+			minHeap:  true,
+		},
+		overflowMapHeap: &mapHeap{
+			selectID: make(map[splitSetID]*mapElement),
+			data:     make([]*mapElement, 0),
+			size:     uint64(0),
```
DavidVorick: same here

modules/miner/splitsetheap.go
```diff
@@ -0,0 +1,211 @@
...
+			break
+		}
+
+		//j is the left child of i.
+		j := j1
+
+		// If the right child (j2) of the element at index i (the sibling of j),
+		// is within the bounds of the heap and satisfies
+		if j2 := j1 + 1; j2 < n && !mh.Less(j1, j2) {
+
```
DavidVorick: I realize that you just copied this code, but this newline is weird and should be removed.

modules/miner/splitsetheap_test.go
```diff
@@ -0,0 +1,395 @@
...
+	for i := 0; i < 1000; i++ {
+		maxPop := max.Pop()
+		minPop := min.Pop()
+
+		if maxPop.set.size != uint64(100*(999-i)) {
+			t.Log(i)
+			t.Error("Unexpected set size in result from max-heap pop.")
+		}
+
+		if minPop.set.size != uint64(100*i) {
```
DavidVorick: can do these checks in a row, without the newline.

modules/miner/splitsetheap_test.go
```diff
@@ -0,0 +1,395 @@
...
+	}
+
+	min := &mapHeap{
+		selectID: make(map[splitSetID]*mapElement),
+		data:     make([]*mapElement, 0),
+		size:     0,
+		minHeap:  true,
+	}
+
+	max.Init()
```
DavidVorick: can do the declaration and init all at once, no need for the two newlines that spread this code out

modules/miner/splitsetheap_test.go
```diff
@@ -0,0 +1,395 @@
...
+	minSizeAfter := min.size
+	maxSizeAfter := max.size
+
+	if minSizeBefore-minRemovedSetSize != minSizeAfter {
+		t.Error("unexpected difference in size after removing from min heap.")
+	}
+
+	if maxSizeBefore-maxRemovedSetSize != maxSizeAfter {
+		t.Error("unexpected difference in size after removing from max heap.")
+	}
```
DavidVorick: spacing here is also over-zealous

modules/miner/splitsetheap_test.go
```diff
@@ -0,0 +1,395 @@
...
+			removedFromMin = false
+			break
+		}
+	}
+
+	if !removedFromMin {
+		t.Error("Element found in  min heap(s) after being removed by splitSetID.")
+	}
+
+	if !removedFromMax {
```
DavidVorick: (and so on)

modules/miner/splitsetheap_test.go
```diff
@@ -0,0 +1,395 @@
...
+		if maxPop.set.averageFee.Cmp(maxPeek.set.averageFee) != 0 {
+			t.Error("Unexpected currency value in result from max-heap Peek.")
+		}
+
+		if minPop.set.averageFee.Cmp(minPeek.set.averageFee) != 0 {
+			t.Error("Unexpected currency value in result from min-heap Peek.")
+		}
+	}
+
+}
```
DavidVorick: This test also has spacing issues

modules/miner/update.go
```diff
@@ -28,11 +27,138 @@ func (m *Miner) addNewTxns(diff *modules.TransactionPoolDiff) {
...
 		}
+
+		elem := &mapElement{
+			set:   s,
+			id:    splitSetID(m.setCounter),
+			index: 0,
+		}
+
+		m.splitSets[splitSetID(m.setCounter)] = s
+		newElements = append(newElements, elem)
```
DavidVorick: I think it's bad design to add the splitSet to the miner state (`m.splitSets`) but then not add the newElements to the global state at the same time. You want all of your dependent state modifications to happen at the same time. Probably better to return an array of splitSets alongside the array of mapElements, and fill out `m.splitSets` in the same function that you add the map elements to the heaps.

DavidVorick: The goal here is to make life easy for yourself as the programmer and for reviewers and other programmers. You want everything that's related all happening in one place, and then you split off the unrelated or fully-independent bits into separate functions.

modules/miner/update.go
```diff
@@ -28,11 +27,138 @@ func (m *Miner) addNewTxns(diff *modules.TransactionPoolDiff) {
...
+			m.overflowMapHeap.Push(elem)
+			// Put transaction sets from bottom back into the blockMapHeap.
+			for _, v := range bottomSets {
+				m.blockMapHeap.Push(v)
+			}
+			// Go to the next candidate set.
+			break
+		}
+
+	}
```
DavidVorick: uneeded newline here

modules/miner/update.go
```diff
@@ -28,11 +27,138 @@ func (m *Miner) addNewTxns(diff *modules.TransactionPoolDiff) {
...
+
+	// While the heap cannot fit this set s, and while the (weighted) average
+	// fee for the lowest sets from the block is less than the fee for the set
+	// s, continue removing from the heap. The block heap doesn't have enough
+	// space for this transaction. Check if removing sets from the blockMapHeap
+	// will be worth it. bottomSets will hold  the lowest fee sets from the
+	// blockMapHeap
+	bottomSets := make([]*mapElement, 0)
+	var sizeOfBottomSets uint64
+	var averageFeeOfBottomSets types.Currency
```
DavidVorick: this should probably be `feesOfBottomSets` to maintain precision

modules/miner/update.go
```diff
@@ -28,11 +27,138 @@ func (m *Miner) addNewTxns(diff *modules.TransactionPoolDiff) {
...
+	for {
+		// Check if the candidateSet can fit in the block.
+		if m.blockMapHeap.size-sizeOfBottomSets+candidateSet.size < types.BlockSizeLimit-5e3 {
+			// Place candidate into block,
+			m.blockMapHeap.Push(elem)
+
+			// Places transactions removed from block heap into
+			// the overflow heap.
+			for _, v := range bottomSets {
```
DavidVorick: grammatical typo

modules/miner/update.go
```diff
@@ -28,11 +27,138 @@ func (m *Miner) addNewTxns(diff *modules.TransactionPoolDiff) {
...
+		// happen since transaction sets are much smaller than the max block
+		// size.
+		_, exists := m.blockMapHeap.Peek()
+		if !exists {
+			m.overflowMapHeap.Push(elem)
+			// Put back in transactions removed.
+			for _, v := range bottomSets {
+				m.blockMapHeap.Push(v)
+			}
+			// Go to the next candidate set.
```
DavidVorick: comment is confusing because the function has no context of multiple candidate sets, all it knows is that it is finishing up the current candidate set.

modules/miner/update.go
```diff
@@ -28,11 +27,138 @@ func (m *Miner) addNewTxns(diff *modules.TransactionPoolDiff) {
...
+		// the fee from this candidate set, put the candidate into the overflow
+		// MapHeap.
+		if averageFeeOfBottomSets.Cmp(candidateSet.averageFee) == 1 {
+			// CandidateSet goes into the overflow.
+			m.overflowMapHeap.Push(elem)
+			// Put transaction sets from bottom back into the blockMapHeap.
+			for _, v := range bottomSets {
+				m.blockMapHeap.Push(v)
+			}
+			// Go to the next candidate set.
```
DavidVorick: ditto with this comment

modules/miner/update.go
```diff
@@ -28,11 +27,138 @@ func (m *Miner) addNewTxns(diff *modules.TransactionPoolDiff) {
...
+
+// addNewTxns adds new unconfirmed transactions to the miner's transaction
+// selection.
+func (m *Miner) addNewTxns(diff *modules.TransactionPoolDiff) {
+	// Get new splitSets (in form of mapElement)
+	newElements := m.getNewSplitSets(diff)
+
+	// Place each elem in one of the MapHeaps.
+	for i := 0; i < len(newElements); i++ {
+		m.addMapElementTxns(newElements[i])
```
DavidVorick: this is where I would also add the split sets to the state, you want to do it all together.

modules/miner/update.go
```diff
@@ -144,5 +253,5 @@ func (m *Miner) ReceiveUpdatedUnconfirmedTransactions(diff *modules.TransactionP
 
 	m.deleteReverts(diff)
 	m.addNewTxns(diff)
-	m.pickNewTransactions(diff)
+	m.adjustUnsolvedBlock()
 }
```
DavidVorick: Overall this file is looking a lot cleaner from the first pass, code quality and organization is definitely much higher.

modules/transactionpool/transactionpool_test.go
```diff
@@ -473,3 +473,195 @@ func TestTpoolScalability(t *testing.T) {
...
+
+	if minFee1.Cmp(maxFee2) < 0 {
+		t.Error("Expected lowest fee from first block to be greater than or equal to than highest fee from second block.")
+	}
+
+	if maxFee1.Cmp(minFee1) <= 0 {
+		t.Error("Expected highest fee from first block to be greater than lowest fee from first block.")
+	}
+
+	if maxFee2.Cmp(minFee2) <= 0 {
```
DavidVorick: Can drop all of the newlines between these checks, no need to spread them out. Pretty simple actions and it's all the single idea of checking that the state is what it's supposed to be.

DavidVorick:
This is looking much better, overall I'd say it's pretty close to being merged.

DavidVorick:
*should also merge or rebase in the master branch again. We've fixed some more NDFs since your last commit.

modules/miner/update.go
marcinja: So each time we add a mapElement we'll have to iterate through the transaction set and compute it's total fee and instead of using the average fee? What about changing the splitSet definition to have a `totalFee` field instead of `averageFee` like mentioned before? That way since it already stores its own size we can still compute a splitSet's average fee precisely and quickly.

marcinja:
By the way, `TestHeapFees` gets stuck in the tpool deadlock (as seen in the Travis CI build and when I just ran `make test-vlong`) but passes when run standalone

modules/miner/update.go
DavidVorick: That seems alright to me. Let's punt it to another PR though.

modules/miner/miner.go
```diff
@@ -165,7 +169,17 @@ func New(cs modules.ConsensusSet, tpool modules.TransactionPool, w modules.Walle
...
+		blockMapHeap: &mapHeap{
+			selectID: make(map[splitSetID]*mapElement),
+			data:     make([]*mapElement, 0),
+			minHeap:  true,
+		},
+		overflowMapHeap: &mapHeap{
+			selectID: make(map[splitSetID]*mapElement),
+			data:     make([]*mapElement, 0),
+			minHeap:  false,
+		},
```
lukechampine: maybe have a `newMapHeap` constructor? Also, it's not common to call `make` with `0` length. You can use `nil` instead or omit the field entirely.

modules/miner/splitsetheap.go
```diff
@@ -0,0 +1,210 @@
...
+		// Swap with the child and continue down the heap.
+		mh.Swap(i, j)
+		i = j
+	}
+	return i > i0
+}
+
+// Len returns the number of items stored in the heap. It implements the sort
+// interface.
+func (mh mapHeap) Len() int {
```
lukechampine: is this interface actually used? I'm not seeing any calls to `sort.Sort`.

modules/miner/splitsetheap.go
```diff
@@ -0,0 +1,210 @@
...
+	mh.data[i], mh.data[j] = mh.data[j], mh.data[i]
+
+	// Change values in slice to correct indices. Note that the same mapeElement
+	// pointer is in the map also, so we only have to mutate it in one place.
+	mh.data[i].index = i
+	mh.data[j].index = j
+}
+
+// Push puts an element onto the heap and maintains the heap condition.
+func (mh *mapHeap) Push(elem *mapElement) {
```
lukechampine: not a big deal, but all functions of `mapHeap` should probably be lowercased unless they're satisfying an interface (e.g. `sort.Interface`)

modules/miner/splitsetheap.go
```diff
@@ -0,0 +1,210 @@
...
+	return elem
+}
+
+// RemoveSetByID removes an element from the MapHeap using only the splitSetID.
+func (mh *mapHeap) RemoveSetByID(s splitSetID) *mapElement {
+	// Get index into data at which the element is stored.
+	i := mh.selectID[s].index
+
+	//Remove it from the heap using the Go library.
+	return mh.Remove(i)
```
lukechampine: Go library?

Ah, I gather that the first iteration of `mapHeap` was a wrapper around `container/heap`? That explains some things.

modules/miner/update.go
```diff
@@ -28,11 +26,139 @@ func (m *Miner) addNewTxns(diff *modules.TransactionPoolDiff) {
...
+// blockMapHeap.
+func (m *Miner) adjustUnsolvedBlock() {
+	numTxns := 0
+	for _, elem := range m.blockMapHeap.selectID {
+		numTxns += len(elem.set.transactions)
+	}
+	// If the transactions that need to be added don't fit in the block,
+	// increase the size of the block by a constant factor to be more efficient.
+	if numTxns > cap(m.persist.UnsolvedBlock.Transactions) {
+		newCap := cap(m.persist.UnsolvedBlock.Transactions) * 6 / 5
```
lukechampine: what is the initial size of `Transactions`?
FYI the default Go slice behavior is to double until 1024 elements, and grow by 25% thereafter.

lukechampine:
I'm curious about the performance impact of using `container/heap`, and also the degree to which it simplifies the code. If the code is a lot simpler and the cost is one or two `interface{}` conversions, it may be worth it. What are the hot paths in the miner code? Do we need to add more benchmarks to the testing suite? Is memory usage a concern, or should we lean towards allocating more to reduce CPU time? I guess overall I'm thinking: modifying the miner to use a heap is one change, and optimizing that heap is another. I think I'd rather see the heap change be as simple as possible, as long as it's reasonably performant. We can go hard on profiling+optimization as a follow up.

DavidVorick:
There were more reasons than performance for copying the heap in fully, namely it gives us a lot more flexibility around the architecture. This PR ended up not utilizing that flexibility as much as expected, but the follow-up will likely use it more.

DavidVorick:
There were more reasons than performance for copying the heap in fully, namely it gives us a lot more flexibility around the architecture. This PR ended up not utilizing that flexibility as much as expected, but the follow-up will likely use it more.

I definitely think we should keep it how it is, especially since we've already gone through the trouble.

PR #1940, opened 2017-06-16
Add the packaging metadata to build the Sia snap

elopio:
This package will let you publish the latest sia in the Ubuntu store, and from there reach many users on all the supported Ubuntu versions, and more Linux distributions in progress.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -8,7 +8,7 @@ description: |
   users to safely store their data with hosts that they do not know or trust.
 
 grade: devel # must be 'stable' to release into candidate/stable channels
-confinement: strict #
+confinement: strict
 
 apps:
   daemon:
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,29 @@
+name: sia
+version: git
+summary: Blockchain-based marketplace for file storage
+description: |
+  Sia is a new decentralized cloud storage platform that radically alters the
+  landscape of cloud storage. By leveraging smart contracts, client-side
+  encryption, and sophisticated redundancy (via Reed-Solomon codes), Sia allows
+  users to safely store their data with hosts that they do not know or trust.
+
+grade: devel # must be 'stable' to release into candidate/stable channels
+confinement: strict #
+
+apps:
+  siad:
+    command: siad
+    plugs: [network, network-bind]
+  client:
+    command: siac
+    plugs: [network]
+    aliases: [siac]
+
+parts:
+  sia:
+    source: .
+    plugin: go
+    go-importpath: github.com/NebulousLabs/Sia
+    after: [go]
+  go:
+    source-tag: go1.8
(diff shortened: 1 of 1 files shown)


elopio:
Hello!

Here at ubuntu we are working a new packaging and delivery system. If you haven't heard about snaps before, here's a lot of info: https://snapcraft.io/

I used Sia to [test our new snapcraft release](https://forum.snapcraft.io/t/in-progress-snapcraft-2-31/682/8), and it worked nicely so here it is for you. I love the idea of distributed storage, and snaps are a great fit because they are an extra layer of security, your users will be autoupdated to the latest revision, and if there are problems with the update an automatic rollback will happen. Also, publishing it in the store will make it trivial for the million of ubuntu users to install it and contribute with the network.

To test this in an Ubuntu 16.04 machine:

    $ sudo apt install git snapcraft
    $ git clone https://github.com/elopio/Sia
    $ cd Sia
    $ git checkout snapcraft
    $ snapcraft
    $ sudo snap install *.snap --dangerous
    $ sia.daemon &
    $ sia.client consensus

(`--dangerous` because the snap you build doesn't come signed from the store. Once you publish it, your users will just have to do `sudo snap install Sia`).

If you have any questions, please let me know.

pura vida.

lukechampine:
Interesting, thanks for doing this. I have heard of snaps but haven't investigated the details of how they work.

I guess my only question is: how do we go about publishing the Sia snap? It looks like the recommend behavior is to automatically publish a new snap every time we push to master. We break master fairly frequently -- could we manually control when to publish instead?

elopio:
Thanks for looking at it @lukechampine.

The most simple way to publish is to go to https://build.snapcraft.io/ and enable continuous delivery. But, the snap will land in the edge channel. The idea is that only your early adopters and the ubuntu community of testers will install Sia from edge, to help you identify bugs earlier. Then you design your pipeline as you want, to move the snap from edge to beta, candidate and finally to stable. Only the snaps in stable are visible on user searches, so feel free to experiment right now with the edge channel.

About Sia-UI, that would be perfect to have in the store. You put a desktop file, and then it will be visible in the software center. Many users just searching for storage solutions will be able to install it from there with one click. In this snap I used only one part. You can have multiple parts, one for the UI, and another for the daemon. The problem here will be with the autostart of the daemon, because we don't yet support user daemons, and I found this problem with Sia today: https://forum.snapcraft.io/t/dac-override-denied-in-the-sia-snap/1017

Come join us in the forum. It will be great to have your perspective there, and your requirements will help us a lot defining the priorities of our roadmap. As a quick solution for the deamon in the Sia UI, I think it will be possible to write a script that launches the daemon before launching the window, we can help you there.

lukechampine:
@elopio: I talked this over with @DavidVorick and we decided to merge this. We probably won't be promoting anything to `stable` anytime soon, but there's little downside to having siad available in the edge channel. Thanks!

elopio:
@lukechampine nice! Now that you have it in master, you can automate the delivery to edge with a couple of clicks in https://build.snapcraft.io

Once it's in edge, we can start a call for testing with the Ubuntu community of testers.

PR #1942, opened 2017-06-16
Some NDFs

DavidVorick:


lukechampine:
Is this a WIP?

DavidVorick:
It should be complete at this point

DavidVorick:
Pushed 2 more ndf fixes because they were within reach.

lukechampine:
I've restarted this 4 times and both builds have failed, for different reasons, every time :|

DavidVorick:
We just have to keep grinding through them I think. Frustrating that the tests are so much more reliable locally, but I think we'll get through these soon enough. We're doing a good job fixing a few every PR. There are a finite number.

PR #1951, opened 2017-06-17
Seed sweep considers transaction size limit

ChrisSchinnerl:
If seed sweep is done on a seed with too many outputs it fails due to the sweep not considering the transaction size limit. This PR should fix that.

### DIFF
--- a/modules/wallet/seed.go
+++ b/modules/wallet/seed.go
@@ -338,35 +338,34 @@ func (w *Wallet) SweepSeed(seed modules.Seed) (coins, funds types.Currency, err 
 		return types.Currency{}, types.Currency{}, errors.New("nothing to sweep")
 	}
 
-	for len(s.siacoinOutputs) > 0 || len(s.siafundOutputs) > 0 {
-		// Create temporary maps and var for the current transaction
-		siacoinOutputs := make(map[types.SiacoinOutputID]scannedOutput)
-		siafundOutputs := make(map[types.SiafundOutputID]scannedOutput)
-		var currentCoins, currentFunds types.Currency
+	// Flatten map to slice
+	var siacoinOutputs, siafundOutputs []scannedOutput
+	for _, sco := range s.siacoinOutputs {
+		siacoinOutputs = append(siacoinOutputs, sco)
+	}
+	for _, sfo := range s.siafundOutputs {
+		siafundOutputs = append(siafundOutputs, sfo)
+	}
 
-		// Fill those maps with up to maxOutputs outputs from the scanner
-		counter := 0
-		for id, output := range s.siacoinOutputs {
-			if counter == maxOutputs {
-				break
-			}
-			siacoinOutputs[id] = output
-			delete(s.siacoinOutputs, id)
-			counter++
-		}
-		for id, output := range s.siafundOutputs {
-			if counter == maxOutputs {
-				break
-			}
-			siafundOutputs[id] = output
-			delete(s.siafundOutputs, id)
-			counter++
-		}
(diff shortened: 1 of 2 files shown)


modules/wallet/seed.go
```diff
@@ -326,123 +326,155 @@ func (w *Wallet) SweepSeed(seed modules.Seed) (coins, funds types.Currency, err
...
-			ParentID:         types.SiafundOutputID(output.id),
-			UnlockConditions: sk.UnlockConditions,
-		})
-		// add a signature for the input
-		sweptFunds = sweptFunds.Add(output.value)
-	}
+	for len(s.siacoinOutputs) > 0 || len(s.siafundOutputs) > 0 {
+		// Create temporary maps and var for the current transaction
+		siacoinOutputs := make(map[types.SiacoinOutputID]scannedOutput)
+		siafundOutputs := make(map[types.SiafundOutputID]scannedOutput)
```
lukechampine: Would this work instead? It looks like we don't need the `OutputID`, so we can flatten the `map` to a `[]scannedOutput`:
```go
var siacoinOutputs, siafundOutputs []scannedOutput
for _, sco := range s.siacoinOutputs {
	siacoinOutputs = append(siacoinOutputs, sco)
}
for _, sfo := range s.siafundOutputs {
	siafundOutputs = append(siafundOutputs, sfo)
}
```
Then you can iterate through them like so:
```go
for len(siacoinOutputs) > 0 || len(siafundOutputs) > 0 {
	// process up to maxOutputs siacoinOutputs
	txnSiacoinOutputs := make([]scannedOutput, maxOutputs)
	n := copy(txnSiacoinOutputs, siacoinOutputs)
	txnSiacoinOutputs = txnSiacoinOutputs[:n]
	siacoinOutputs = siacoinOutputs[n:]

	// process up to (maxOutputs-n) siafundOutputs
	txnSiafundOutputs := make([]scannedOutput, maxOutputs-n)
	n = copy(txnSiafundOutputs, siafundOutputs)
	txnSiafundOutputs = txnSiafundOutputs[:n]
	siafundOutputs = siafundOutputs[n:]

	// ... use txnSiacoinOutputs and txnSiafundOutputs
}
```

lukechampine: I think your way works fine too, but perhaps would be cleaner with slices instead of maps, i.e.
```go
for len(s.siacoinOutputs) > 0 || len(s.siafundOutputs) > 0 {
	var siacoinOutputs, siafundOutputs []scannedOutput
	for id, output := range s.siacoinOutputs {
		if len(siacoinOutputs) > maxOutputs {
			break
		}
		siacoinOutputs = append(siacoinOutputs, output)
		delete(s.siacoinOutputs, id)
	}
	for id, output := range s.siafundOutputs {
		if len(siacoinOutputs)+len(siafundOutputs) > maxOutputs {
			break
		}
		siafundOutputs = append(siafundOutputs, output)
		delete(s.siafundOutputs, id)
	}

	// ... use siacoinOutputs and siafundOutputs
}
```

ChrisSchinnerl: Yes, of course. You are right.

ChrisSchinnerl: I changed it to your first suggestion since it doesn't require that much nesting.

lukechampine:
Awesome! Can you confirm that the updated test fails using the old single-txn SweepSeed? (no need to push any code for this, just run it locally and comment with the result)

ChrisSchinnerl:
yep, it fails as expected with `seed_test.go:450: transaction is too large for this transaction pool`

PR #1959, opened 2017-06-19
implement difficulty hardfork (no tests)

DavidVorick:
This is the long-awaited difficulty hardfork that has been holding back the v1.3.0 betas. Please do not merge until I've had time to write more tests and go over this with a fine toothed comb.

The design and algorithm are tested, simulated, and well thought through, including edge cases and adversarial situations. The design itself should be ready to rock, it's more a question of whether the implementation matches the design or if there's some egregious mistake somewhere.

### DIFF
--- a/modules/consensus/consensusdb.go
+++ b/modules/consensus/consensusdb.go
@@ -41,6 +41,13 @@ var (
 	// blocks in the current path.
 	BlockPath = []byte("BlockPath")
 
+	// BucketOak is the database bucket that contains all of the fields related
+	// to the oak difficulty adjustment algorithm. The cumulative difficulty and
+	// time values are stored for each block id, and then the key "OakInit"
+	// contains the value "true" if the oak fields have been properly
+	// initialized.
+	BucketOak = []byte("Oak")
+
 	// Consistency is a database bucket with a flag indicating whether
 	// inconsistencies within the database have been detected.
 	Consistency = []byte("Consistency")
@@ -60,6 +67,18 @@ var (
 	// SiafundPool is a database bucket storing the current value of the
 	// siafund pool.
 	SiafundPool = []byte("SiafundPool")
+)
+
+var (
+	// FieldOakInit is a field in BucketOak that gets set to "true" after the
+	// oak initialiation process has completed.
+	FieldOakInit = []byte("OakInit")
+)
+
+var (
+	// ValueOakInit is the value that the oak init field is set to if the oak
+	// difficulty adjustment fields have been correctly intialized.
+	ValueOakInit = []byte("true")
 )
 
 // createConsensusObjects initialzes the consensus portions of the database.
--- a/modules/consensus/difficulty.go
+++ b/modules/consensus/difficulty.go
@@ -1,0 +1,204 @@
+package consensus
+
+import (
+	"bytes"
+	"encoding/binary"
+	"math/big"
+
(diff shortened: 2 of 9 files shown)


modules/consensus/difficulty.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	// For each value, first multiply by the decay, and then add in the new
+	// delta.
+	newTotalTime = (prevTotalTime * types.OakDifficultyDecayNum / types.OakDifficultyDecayDenom) + (int64(currentTimestamp) - int64(parentTimestamp))
+	newTotalTarget = prevTotalTarget.MulDifficulty(big.NewRat(types.OakDifficultyDecayNum, types.OakDifficultyDecayDenom)).AddDifficulties(targetOfCurrentBlock)
+
+	// Store the new total time and total target in the database at the
+	// appropriate id.
+	bytes := make([]byte, 72, 72)
```
lukechampine: you can omit the capacity if it's the same as the length

types/constants.go
```diff
@@ -39,6 +45,11 @@ var (
 	// The GenesisID is used in many places. Calculating it once saves lots of
 	// redundant computation.
 	GenesisID BlockID
+
+	// Oak hardfork constants.
+	OakHardforkBlock     BlockHeight
+	OakDifficultyMaxRise = big.NewRat(1004, 1000)
+	OakDifficultyMaxDrop = big.NewRat(1000, 1004)
```
lukechampine: do these belong in `types`? I thought previous hardfork values lived in `consensus`. I feel like they make more sense there.

modules/consensus/consensusdb.go
```diff
@@ -62,6 +69,18 @@ var (
...
+	// FieldOakInit is a field in BucketOak that gets set to "true" after the
+	// oak initialiation process has completed.
+	FieldOakInit = []byte("OakInit")
+)
+
+var (
+	// ValueOakInit is the value that the oak init field is set to if the oak
+	// difficulty adjustment fields have been correctly intialized.
+	ValueOakInit = []byte("true")
+)
```
lukechampine: why are these exported? (why are any of the bucket names exported?)

types/constants.go
DavidVorick: The existing difficulty adjustment values live in `types` already, using the names `MaxAdjustmentUp` and `MaxAdjustmentDown`.

Things belong in the `types` package if other, non-consensus packages might need to see the constants, because they can't necessarily import the consensus package (in practice they can).

I'll think about package placement a bit more, but I think the  types package is the right place for the hardfork height.

modules/consensus/consensusdb.go
DavidVorick: I am not sure, I don't think the bucket names need to be exported. But if we un-export them, that should happen in another PR. This PR has it because it's consistent.

types/constants.go
lukechampine: I guess it's consistent with functions like [`types.Tax`](https://github.com/NebulousLabs/Sia/blob/2a58b7e9a314173ae0d8390a66f94227feab194a/types/filecontracts.go#L121) including hardfork heights

DavidVorick: I'm going to leave it in `types` because that's where all of the other consensus tuning parameters are. Makes sense to me to keep everything together. I'll clean up the comments at least though.

PR #1960, opened 2017-06-19
fix tpool syncDB race condition

lukechampine:


DavidVorick:
Wish I knew a good way to test this, but I don't. It's just updating something outside of conventions to follow conventions though, and we've already got 1000 tests that interact with this code.

PR #1961, opened 2017-06-19
sanity checks around code optimizations

DavidVorick:
One of my sanity checks found a bug, that's also fixed here.

### DIFF
--- a/types/transactions.go
+++ b/types/transactions.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"io"
 
+	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/encoding"
 )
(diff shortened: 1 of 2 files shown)


types/transactions.go
```diff
@@ -186,19 +187,34 @@ func (t Transaction) marshalSiaNoSignatures(w io.Writer) {
...
+			t.SiafundInputs,
+			t.SiafundOutputs,
+			t.MinerFees,
+			t.ArbitraryData,
+		))
+
+		if verify != txid {
+			println(txid.String())
+			println(verify.String())
+			panic("transaction id marshalling is incorrect")
```
lukechampine: I don't think dumping the strings helps much here, they don't really contain any actionable information

DavidVorick: ah this is just leftover. It contains actionable information if it's something like all '0's, which was the case on my first attempt at this sanity check. And then after that I was dumping the raw encoded bytes for both formats to figure out what else was incorrect.

Was not my intention to leave the printlns here though.

PR #1963, opened 2017-06-19
Ibd fix

lukechampine:


Makefile
```diff
@@ -76,7 +76,7 @@ test:
 test-v:
 	go test -race -v -short -tags='debug testing' -timeout=15s $(pkgs) -run=$(run)
 test-long: clean fmt vet lint
-	go test -v -race -tags='testing debug' -timeout=500s $(pkgs) -run=$(run)
+	go test -v -race -tags='testing debug' -timeout=800s $(pkgs) -run=$(run)
```
DavidVorick: This might cause problems on travis

DavidVorick: if there is no output for 600s Travis will kill the build.

lukechampine: I think Travis auto-fails if there's no output after a certain amount of time (10 mins?) but our tests should print frequently enough to prevent it

lukechampine:
I shaved ~100 seconds off the api tests, but the tpool tests still routinely exceed 500 seconds. Not sure how we want to handle those.

DavidVorick:
I'd `t.Skip` the scalability test. It's less of a correctness test and more of a 'this needs to run faster' test. Marcin's branch should have those tests running a lot faster.

PR #1969, opened 2017-06-21
Host Scalability Update

ChrisSchinnerl:
This PR is meant to increase the scalability of the host to achieve 1TB file contracts. The first step is to modify the contractmanager WAL to not rely on a tmp log since `os.Rename` is not atomic. Once the WAL is fixed the two host databases need to be merged into a single one.

### DIFF
--- a/modules/host/contractmanager/consts.go
+++ b/modules/host/contractmanager/consts.go
@@ -33,11 +33,6 @@ const (
 	// walFile is the name of the file that is used to save the write ahead log
 	// for the contract manager.
 	walFile = "contractmanager.wal"
-
-	// walFileTmp is used for incomplete writes to the WAL. Data could be
-	// interrupted by power outages, etc., and is therefore written to a
-	// temporary file before being atomically renamed to the correct name.
-	walFileTmp = "contractmanager.wal_temp"
 )
 
 const (
--- a/modules/host/contractmanager/writeaheadlog.go
+++ b/modules/host/contractmanager/writeaheadlog.go
@@ -1,18 +1,34 @@
 package contractmanager
 
 import (
+	"bytes"
+	"encoding/binary"
 	"encoding/json"
 	"errors"
-	"io"
 	"os"
 	"path/filepath"
 	"sync"
 
 	"github.com/NebulousLabs/Sia/build"
+	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/persist"
+	"log"
 )
 
 type (
+	// walHeader is the header of the wal file
+	//
+	// lengthMD defines the changeLength (in bytes) of the metadata section right after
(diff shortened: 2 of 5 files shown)


modules/host/contractmanager/writeaheadlog.go
```diff
@@ -104,34 +129,137 @@ type (
...
+	buffer := bytes.NewBuffer(bytesPrefix)
+	var prefix stateChangePrefix
+
+	err = binary.Read(buffer, binary.LittleEndian, &prefix)
+	if err != nil {
+		return 0, build.ExtendErr("could not decode stateChange prefix", err)
+	}
+
+	// Read the state change
+	bytesChange := make([]byte, prefix.ChangeLength)
```
ChrisSchinnerl: Just noticed that this is actually quite dangerous. Might have to check that `prefix.ChangeLength` is below `offset+wal.header.OffsetSC` to prevent a panic

modules/host/contractmanager/consts.go
```diff
@@ -76,6 +71,13 @@ var (
 		Header:  "Sia Contract Manager WAL",
 		Version: "1.2.0",
 	}
+
+	// maxWalSize defines the maximum size in bytes of the WAL
+	maxWalSize = build.Select(build.Var{
+		Dev:      int64(5e+5),
+		Standard: int64(5e+6),
+		Testing:  int64(5e+4),
```
DavidVorick: for readability, we only use powers of 3 in our exponentiation. That makes match the phonetics (50e3 = 50 thousand :: 5e4 = 5 ten-thousands?). In my experience, it's much easier to read. So you'd have 50e3, 500e3, and 5e6 for your values above.

modules/host/contractmanager/writeaheadlog.go
```diff
@@ -225,22 +450,24 @@ func (wal *writeAheadLog) recoverWAL(walFile file) error {
...
-		if err == nil {
-			// The uncommitted changes are loaded into memory using a simple
-			// append, because the tmp WAL file has not been created yet, and
-			// will not be created until the sync loop is spawned. The sync
-			// loop spawner will make sure that the uncommitted changes are
-			// written to the tmp WAL file.
-			wal.commitChange(sc)
-			scs = append(scs, sc)
+	offset := headerLength() + header.LengthMD
+	for offset < maxWalSize {
```
ChrisSchinnerl: This loop might need to be changed to run until an error occured. There is a slim chance that the reset which is initialized at 80% of `maxWalSize` doesn't finish in time. That means that the WAL grows larger than `maxWalSize` and is cut off in this loop. 
Of course other checks that use maxWalSize would have to be modified accordingly since maxWalSize wouldn't be the actual maxWalSize anymore but rather a soft limit.

PR #1973, opened 2017-06-21
optimize double-load of parent in addBlockToTree

lukechampine:


modules/consensus/accept.go
```diff
@@ -28,39 +28,42 @@ func (cs *ConsensusSet) managedBroadcastBlock(b types.Block) {
 // validateHeaderAndBlock does some early, low computation verification on the
 // block. Callers should not assume that validation will happen in a particular
 // order.
-func (cs *ConsensusSet) validateHeaderAndBlock(tx dbTx, b types.Block) error {
+func (cs *ConsensusSet) validateHeaderAndBlock(tx dbTx, b types.Block) (parent processedBlock, err error) {
```
DavidVorick: should just return a pointer here

PR #1978, opened 2017-06-22
File sharing feature

tlightsky:
implement this feature by these steps:
#1 use shareAscii to share file,contracts with publickey 
#2 use loadShareAscii to load share info on another renter 
#3 check publickeys is sufficient to download or not ,if not form new download only contracts
#4 use renter download to download loaded file (with setallowance before)

### DIFF
--- a/api/ecosystem_helpers_test.go
+++ b/api/ecosystem_helpers_test.go
@@ -44,6 +44,10 @@ func announceAllHosts(sts []*serverTester) error {
 		return err
 	}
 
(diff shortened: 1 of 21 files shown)


CarloPalinckx:
You should consider turning those vars into consts

tlightsky:
rebased to current master

tlightsky:
rebased to current master

tlightsky:
seems a wrong rebase happen, try to fix it

Issue #1970, opened 2017-06-21
Can't bind siad to interface/address


D4rk4:
siad -M cghrtw --rpc-addr 10.0.1.240:9981 --host-addr 10.0.1.240:9982

netstat -tapn | grep siad
tcp        0      0 127.0.0.1:9980          0.0.0.0:*               LISTEN      29987/siad
tcp        0      0 10.0.1.240:9981         0.0.0.0:*               LISTEN      29987/siad
tcp        0      0 10.0.1.240:9982         0.0.0.0:*               LISTEN      29987/siad
tcp        0      1 10.0.1.250:5864         24.27.103.185:9982      SYN_SENT    29987/siad
tcp        0      1 10.0.1.250:24330        100.64.62.174:9982      SYN_SENT    29987/siad
tcp        0      1 10.0.1.250:60610        68.118.232.232:9982     SYN_SENT    29987/siad
tcp        0      1 10.0.1.250:61884        101.200.214.115:9981    SYN_SENT    29987/siad
tcp        0      1 10.0.1.250:1676         192.222.228.186:9982    SYN_SENT    29987/siad
tcp        0      1 10.0.1.250:5532         79.119.155.56:9982      SYN_SENT    29987/siad
tcp        0      1 10.0.1.250:48356        99.91.195.81:9982       SYN_SENT    29987/siad
tcp        0      1 10.0.1.250:26208        81.64.128.68:9982       SYN_SENT    29987/siad
tcp        0      1 10.0.1.250:7238         90.87.224.114:9982      SYN_SENT    29987/siad
tcp        0      1 10.0.1.250:20770        96.231.239.120:9982     SYN_SENT    29987/siad
tcp        0      0 10.0.1.250:3158         122.150.251.159:9982    ESTABLISHED 29987/siad
tcp        0      1 10.0.1.250:29888        79.172.204.10:9981      SYN_SENT    29987/siad
tcp        0      1 10.0.1.250:4106         73.71.41.138:9982       SYN_SENT    29987/siad
tcp        0      1 10.0.1.250:47304        95.86.33.52:9982        SYN_SENT    29987/siad
tcp        0      1 10.0.1.250:28214        195.211.84.149:9982     SYN_SENT    29987/siad
tcp        0      1 10.0.1.250:38482        94.123.92.8:9982        SYN_SENT    29987/siad
tcp        0     24 10.0.1.250:55518        99.27.120.250:9982      ESTABLISHED 29987/siad
[...]

PR #1988, opened 2017-06-23
fix 1970

nf404:
fix https://github.com/NebulousLabs/Sia/issues/1970


modules/gateway/conn.go
```diff
@@ -26,6 +26,7 @@ func (g *Gateway) dial(addr modules.NetAddress) (net.Conn, error) {
 	dialer := &net.Dialer{
 		Cancel:  g.threads.StopChan(),
 		Timeout: dialTimeout,
+		LocalAddr: config.Siad.HostAddr,
```
lukechampine: close! But I think what we want is more like `LocalAddr: g.myAddr.Host()`. The part I'm not sure about is that (I believe) we need to specify the port as well. Maybe `:0` works, like it does with `net.Listen`?

PR #1994, opened 2017-06-24
Convert manageAcceptBlocks to accept multiple blocks, for faster blockchain download

DavidVorick:
Thanks to @starius for the base code that I am building off of.


DavidVorick:
derp. Did not mean to PR this without fixing it up further. I don't think it works yet.

@starius I'm going to be using this to clean up the merge conflict that resulted when @lukechampine added a performance optimization. We spaced and didn't realize we were stepping directly on your toes, and the result was pretty messy.

I'm hoping to have this done in the next few hours, so we can get your huge performance boost into the codebase.

DavidVorick:
Code itself is ready to go I think, including a bunch of sanity checks to verify that we're doing more or less what we were doing previously, and to make sure delicate behaviors are correct.

Cleaned up a few things as well, including several mistakes that could lead to DoS attacks and things that would be in a similar vein to eclipse attacks. Nothing super serious, but better to have it resolved.

I still need to write a test that's dedicated to probing the various edge cases involved with submitting multiple blocks, can't merge until that test is ready, but it should be prepared soon.

modules/consensus/accept.go
```diff
@@ -207,80 +217,153 @@ func (cs *ConsensusSet) addBlockToTree(b types.Block, parent *processedBlock) (c
...
+	if setErr != nil {
+		// Check if any blocks were valid.
+		if len(validBlocks) < 1 {
+			// Nothing more to do, the first block was invalid.
+			return false, setErr
+		}
+
+		// At least some of the blocks were valid. Add the valid blocks before
+		// returning, since we've already done the downloading and header
+		// validation.
```
lukechampine: is there a DoS attack wherein the attacker sends the maximum number of blocks, with the last being invalid, forcing consensus to redo a bunch of work?

lukechampine:
LGTM, just one question above.

@starius going forward, please avoid prefixing commits with `fixup:` since it is too similar to the special `fixup!` git directive. We typically don't use prefixes like this, but if it's your preferred style, something like `fix:` is probably ok.

starius:
@lukechampine Thanks! I'll take this into account. Actually I planned to squash them using `git rebase -i` but decided to keep to make code changes traceable.

PR #2004, opened 2017-06-25
Fix a few NDFs`

DavidVorick:
2 NDFs fixed, one given more diagnostics.


DavidVorick:
Both builds passed on the first try. I'll take that as a sign that we're finally near the end of this ndf mess.

PR #2009, opened 2017-06-26
don't use float math to convert bytes to blocks

starius:


lukechampine:
meant to fix this before it was merged in, thanks for the catch

PR #2016, opened 2017-06-26
ConsensusSetSubscribe abort 

ChrisSchinnerl:
Added a param to ConsensusSetSubscribe to allow it to be aborted if required.


modules/explorer/explorer.go
```diff
@@ -77,7 +77,7 @@ func New(cs modules.ConsensusSet, persistDir string) (*Explorer, error) {
 		return nil, err
 	}
 
-	err = cs.ConsensusSetSubscribe(e, recentChange)
+	err = cs.ConsensusSetSubscribe(e, recentChange, nil)
```
ChrisSchinnerl: Not sure about this one. Maybe we could pass the `sigChan` directly? Doesn't feel like a clean solution though.

modules/host/announce_test.go
```diff
@@ -45,7 +45,7 @@ func newAnnouncementFinder(cs modules.ConsensusSet) (*announcementFinder, error)
 	af := &announcementFinder{
 		cs: cs,
 	}
-	err := cs.ConsensusSetSubscribe(af, modules.ConsensusChangeBeginning)
+	err := cs.ConsensusSetSubscribe(af, modules.ConsensusChangeBeginning, nil)
```
ChrisSchinnerl: Didn't pass the channel here because I don't think that it really matters that much for those tests.

modules/consensus/subscribe.go
```diff
@@ -169,12 +172,17 @@ func (cs *ConsensusSet) managedInitializeSubscribe(subscriber modules.ConsensusS
 		cs.mu.RLock()
 		err = cs.db.View(func(tx *bolt.Tx) error {
 			for i := 0; i < 100 && exists; i++ {
-				cc, err := cs.computeConsensusChange(tx, entry)
-				if err != nil {
-					return err
+				select {
+				case <-cancel:
+					return errors.New("aborting managedInitializeSubscribe")
```
lukechampine: Is an error appropriate here? If there shutdown is intentional, I don't see why it should result in an error

ChrisSchinnerl: Yeah I wasn't sure about that one. I thought that since most of the time the channel that is passed is the `StopChan` of a threadgroup it might make sense to abort instantly right after calling `ConsensusSetSubscribe` instead of executing some other code like subscribing to the tpool which might take a few more seconds. Of course semantically it would make more sense to return `nil`.

lukechampine: ah, I see. Hmm.
We don't have a clear convention for this because `StopChan` usually just causes a goroutine to exit, not a blocking call to return early. The only other instance I could find is in the renter:

```go
// wait for the download to complete and return the data
select {
case <-d.downloadFinished:
	return buf.Bytes(), d.Err()
case <-r.tg.StopChan():
	return nil, errors.New("chunk download interrupted by shutdown")
case <-time.After(chunkDownloadTimeout):
	return nil, errors.New("chunk download timed out")
}
```
so returning an error might be okay after all. Still, I think we'd at least want it to be an exported variable (e.g. `siasync.ErrStopped`) so that the caller can check for it explicitly.

ChrisSchinnerl: Ok I will change that. Should I use `siasync.ErrStopped` or would you prefer a `consensus.ErrSubscribeCanceled` in subscribe.go?

lukechampine: I'd like @DavidVorick to chime in here before merging, but in the meantime I think `siasync.ErrStopped` is slightly preferable.

ChrisSchinnerl: Done, but for some reason it won't show up here although I pushed it to the [branch](https://github.com/NebulousLabs/Sia/compare/master...ChrisSchinnerl:CSSubscribeInterrupt?expand=1)

lukechampine: hmm, same thing happened to me recently. Might be a bug with GitHub.

modules/explorer/explorer.go
lukechampine: `nil` is fine for now. We'll give the explorer a threadgroup later.

PR #2018, opened 2017-06-26
close conn if StopChan signals shutdown

DavidVorick:


modules/consensus/synchronize.go
```diff
@@ -127,6 +127,19 @@ func (cs *ConsensusSet) managedReceiveBlocks(conn modules.PeerConn) (returnErr e
...
 	}
+	finishedChan := make(chan struct{})
+	defer close(finishedChan)
+	go func() {
+		select {
+		case <-cs.tg.StopChan():
+		case <-finishedChan:
+		}
+		conn.Close()
+	}()
```
lukechampine: huh. So previously, this conn was never being closed?

DavidVorick: it had a timeout, but otherwise I don't think it was being closed.

Maybe this is the source of the 10,000 system threads? Except, those didn't time out ever. I'm not sure.

PR #2025, opened 2017-06-26
specify units in /tpool/fee API docs

johnathanhowell:


lukechampine:
aren't units typically specified in the comment next to the field?

johnathanhowell:
yeah, I updated this PR to match that style

PR #2031, opened 2017-06-27
host: read only needed part of sector from disk

starius:


modules/host/negotiatedownload.go
```diff
@@ -89,11 +89,11 @@ func (h *Host) managedDownloadIteration(conn net.Conn, so *storageObligation) er
 
 		// Load the sectors and build the data payload.
 		for _, request := range requests {
-			sectorData, err := h.ReadSector(request.MerkleRoot)
+			sectorData, err := h.ReadSectorPart(request.MerkleRoot, uint32(request.Offset), uint32(request.Length))
```
lukechampine: why the conversion between `uint32` and `uint64`? Probably best to use `uint64` throughout, only converting to `int64` in the call to `ReadAt`.

tbenz9:
@starius are you still interested in working on this PR?  If not, I suggest we close it so it's not stalled indefinitely.

PR #2032, opened 2017-06-27
Improve visual format of siac host param README.md section

rudibs:
lost preformat indicators losing tabulation on last adjustment


lukechampine:
this doesn't appear to render as a proper markdown table. I don't think they can be enclosed in ```

rudibs:
any better Luke? 
anybetter

Issue #1990, opened 2017-06-23
Explorer database generation


S-anasol:
1. Extremely slow, even on SSD(12+ hrs).

2. Result random, after generation explorer shows incorrect info. Like zero numbers, zero block heights etc.
In most tries explorer data correct until 95-105k height, but after shows zero values(but with correct(?) transactions etc).

Some additional info for @starius in russian(:trollface:)  : https://gist.github.com/S-anasol/53043dffd2e006e5a39dce4213646bac

This issue for @starius, keep it open please 👍 
Or you can help 😃

tbenz9:
Could be related to #2074

PR #2046, opened 2017-06-28
Consensus blocks api

S-anasol:
Mini real-time blocks explorer.

Since internal block explorer unusable, we need at least possibility to get 'raw' blocks from siad.
I know that explorer have similar function, but:
1. Explorer unusable, very long and incorrect DB build. #1990
2. Explorer block api unavailable with disabled explorer module([you dont say.jpg](https://vignette1.wikia.nocookie.net/anime-arts/images/1/1e/You_dont_say.jpg/revision/latest?cb=20130224165024)).

Also i believe that blocks api should be in consensus module.


S-anasol:
Response example: https://gist.github.com/S-anasol/3fb462145ee8c13f16cf80f081b5642a

Human readable view: 
image

S-anasol:
No commets?
let me know if you want some additions.

But this is must have and useful 👍 
https://explorer.siahub.info/ powered by this PR

api/api.go
```diff
@@ -162,6 +162,7 @@ func New(requiredUserAgent string, requiredPassword string, cs modules.Consensus
 	if api.cs != nil {
 		router.GET("/consensus", api.consensusHandler)
 		router.POST("/consensus/validate/transactionset", api.consensusValidateTransactionsetHandler)
+		router.GET("/consensus/blocks/:height", api.consensusBlocksHandler)
```
DavidVorick: Wondering if we should let the user request by both height and id, or if just height is enough

api/consensus.go
```diff
@@ -19,6 +21,55 @@ type ConsensusGET struct {
 	Difficulty   types.Currency    `json:"difficulty"`
 }
 
+type ConsensusFileContract struct {
```
DavidVorick: needs a docstring

api/consensus.go
```diff
@@ -19,6 +21,55 @@ type ConsensusGET struct {
...
+	WindowStart        types.BlockHeight              `json:"windowstart"`
+	WindowEnd          types.BlockHeight              `json:"windowend"`
+	Payout             types.Currency                 `json:"payout"`
+	ValidProofOutputs  map[string]types.SiacoinOutput `json:"validproofoutputs"`
+	MissedProofOutputs map[string]types.SiacoinOutput `json:"missedproofoutputs"`
+	UnlockHash         types.UnlockHash               `json:"unlockhash"`
+	RevisionNumber     uint64                         `json:"revisionnumber"`
+}
+
+type ConsensusFileContractRevision struct {
```
DavidVorick: docstring

api/consensus.go
```diff
@@ -19,6 +21,55 @@ type ConsensusGET struct {
...
+	NewFileSize           uint64                         `json:"newfilesize"`
+	NewFileMerkleRoot     crypto.Hash                    `json:"newfilemerkleroot"`
+	NewWindowStart        types.BlockHeight              `json:"newwindowstart"`
+	NewWindowEnd          types.BlockHeight              `json:"newwindowend"`
+	NewValidProofOutputs  map[string]types.SiacoinOutput `json:"newvalidproofoutputs"`
+	NewMissedProofOutputs map[string]types.SiacoinOutput `json:"newmissedproofoutputs"`
+	NewUnlockHash         types.UnlockHash               `json:"newunlockhash"`
+}
+
+type ConsensusTransaction struct {
```
DavidVorick: docstring

api/consensus.go
```diff
@@ -19,6 +21,55 @@ type ConsensusGET struct {
...
+	SiacoinInputs         map[string]types.SiacoinInput            `json:"siacoininputs"`
+	SiacoinOutputs        map[string]types.SiacoinOutput           `json:"siacoinoutputs"`
+	FileContracts         map[string]ConsensusFileContract         `json:"filecontracts"`
+	FileContractRevisions map[string]ConsensusFileContractRevision `json:"filecontractrevisions"`
```
DavidVorick: Why are these newly defined types instead of just using the types package?

api/consensus.go
```diff
@@ -19,6 +21,55 @@ type ConsensusGET struct {
...
 }
 
+type ConsensusFileContract struct {
+	FileSize           uint64                         `json:"filesize"`
+	FileMerkleRoot     crypto.Hash                    `json:"filemerkleroot"`
+	WindowStart        types.BlockHeight              `json:"windowstart"`
+	WindowEnd          types.BlockHeight              `json:"windowend"`
+	Payout             types.Currency                 `json:"payout"`
+	ValidProofOutputs  map[string]types.SiacoinOutput `json:"validproofoutputs"`
+	MissedProofOutputs map[string]types.SiacoinOutput `json:"missedproofoutputs"`
```
DavidVorick: I don't think using a map here is the right move, maps in go are randomly ordered, but valid and missed proofs have a specific order, and the order is meaningful.

DavidVorick: (renter is first, host is second, void is third)

api/consensus.go
S-anasol: need this for child maps with hash as key
`ValidProofOutputs  map[string]types.SiacoinOutput`

but actually not sure that we need these output/input hashes, it should be in blockchain in usual way too as simple out/in not in contract?

api/consensus.go
S-anasol: comment above should answer this too :)

If we dont need this maps at all, it can be removed and used types package.

api/api.go
S-anasol: since at zero point we dont have any hash - easier way to get first data is just height of block.

Optionally ofc can be queried by hash, but i see no need for me.

DavidVorick:
All changes to the API require two things:

1. Updates to doc/API.md and the corresponding doc/api/module.md, explaining the full route, all the fields and their meanings, and providing basic example usages.

2. A test in the api package demonstrating that the route works. Doesn't need to be super comprehensive but does need to show that the fields returned have non-zero values and that basic error handling (incorrect height, id, etc.) is in place.

I'm not sure what the reason was for making the valid and missed proofs part of a map, but I think we should look at a different way of accomplishing your goal. I think it would be much cleaner if the base types were not changed, and any extra/augmenting information should be in a separate data structure.

api/consensus.go
DavidVorick: Is there any particular reason that you can't just use a `types.Block` to define the whole data structure? I don't see why we need these custom definitions, but I'm also not fully sure what all you are using it for.

S-anasol: using custom structures with `map[string]` instead of types with `map[]` to fill it with any type hashes to make possibility index it with external tools.

Sia internally generate hash every time when it need, like this:
https://github.com/NebulousLabs/Sia/blob/master/types/transactions.go#L217-L248

So default structures doesn't contain hash of anything.

Check this out sia explorer vs my indexer: 
https://github.com/S-anasol/SiaConsensusIndexer/blob/master/update.php#L58-L119

https://github.com/NebulousLabs/Sia/blob/master/modules/explorer/update.go#L59

Sia Explorer uses internal functions to calculate hash(ID) for data array,
So making indexer in other language with full copy of Sia hashes calculation is complicated and overheaded.

I hope understandable explanation :)

S-anasol: Final result of my indexer

image

S-anasol: Maybe there is better way to get same result, but i don't know it :)

api/consensus.go
DavidVorick: Ok. My very strong preference is that we just return a normal types.Block. If someone does end up needing the hashes to be computed for them, we can return all the ids (txids, block id, output ids, etc.) in a separate struct that mirrors the block, except it only has ids as fields. But if nobody is using it, we don't need it.

tbenz9:
@S-anasol are you still interested in working on this PR or should we close it?

PR #2056, opened 2017-06-30
Update the way host feedback works

ChrisSchinnerl:
This PR is meant to improve existing measurements and add new ones like latency, speed and responsiveness to requests.

The first step is to add "InteractionReliability" by counting all the successful and failed interactions with a host (upload, download, delete, modify, form new contract, renew contract and scan) and adding them up to historic values which decay at a certain rate after each new block.

### DIFF
--- a/modules/renter/contractor/contractor_test.go
+++ b/modules/renter/contractor/contractor_test.go
@@ -34,6 +34,8 @@ func (newStub) AllHosts() []modules.HostDBEntry                                 
 func (newStub) ActiveHosts() []modules.HostDBEntry                              { return nil }
 func (newStub) Host(types.SiaPublicKey) (settings modules.HostDBEntry, ok bool) { return }
 func (newStub) RandomHosts(int, []types.SiaPublicKey) []modules.HostDBEntry     { return nil }
+func (newStub) IncrementSuccessfulInteractions(key types.SiaPublicKey)          {}
+func (newStub) IncrementFailedInteractions(key types.SiaPublicKey)              {}
 
 // TestNew tests the New function.
 func TestNew(t *testing.T) {
@@ -186,6 +188,8 @@ func (stubHostDB) ActiveHosts() (hs []modules.HostDBEntry)                      
 func (stubHostDB) Host(types.SiaPublicKey) (h modules.HostDBEntry, ok bool)         { return }
 func (stubHostDB) PublicKey() (spk types.SiaPublicKey)                              { return }
 func (stubHostDB) RandomHosts(int, []types.SiaPublicKey) (hs []modules.HostDBEntry) { return }
+func (stubHostDB) IncrementSuccessfulInteractions(key types.SiaPublicKey)           {}
+func (stubHostDB) IncrementFailedInteractions(key types.SiaPublicKey)               {}
 
 // TestIntegrationSetAllowance tests the SetAllowance method.
 func TestIntegrationSetAllowance(t *testing.T) {
--- a/modules/renter/contractor/contractor_test.go
+++ b/modules/renter/contractor/contractor_test.go
@@ -34,8 +34,8 @@ func (newStub) AllHosts() []modules.HostDBEntry                                 
 func (newStub) ActiveHosts() []modules.HostDBEntry                              { return nil }
 func (newStub) Host(types.SiaPublicKey) (settings modules.HostDBEntry, ok bool) { return }
 func (newStub) RandomHosts(int, []types.SiaPublicKey) []modules.HostDBEntry     { return nil }
-func (newStub) IncrementSuccessfulInteractions(key types.SiaPublicKey)          {}
-func (newStub) IncrementFailedInteractions(key types.SiaPublicKey)              {}
+func (newStub) IncrementSuccessfulInteractions(key types.SiaPublicKey)          { return }
+func (newStub) IncrementFailedInteractions(key types.SiaPublicKey)              { return }
 
 // TestNew tests the New function.
 func TestNew(t *testing.T) {
@@ -188,8 +188,8 @@ func (stubHostDB) ActiveHosts() (hs []modules.HostDBEntry)                      
 func (stubHostDB) Host(types.SiaPublicKey) (h modules.HostDBEntry, ok bool)         { return }
 func (stubHostDB) PublicKey() (spk types.SiaPublicKey)                              { return }
 func (stubHostDB) RandomHosts(int, []types.SiaPublicKey) (hs []modules.HostDBEntry) { return }
-func (stubHostDB) IncrementSuccessfulInteractions(key types.SiaPublicKey)           {}
-func (stubHostDB) IncrementFailedInteractions(key types.SiaPublicKey)               {}
+func (stubHostDB) IncrementSuccessfulInteractions(key types.SiaPublicKey)           { return }
+func (stubHostDB) IncrementFailedInteractions(key types.SiaPublicKey)               { return }
 
 // TestIntegrationSetAllowance tests the SetAllowance method.
 func TestIntegrationSetAllowance(t *testing.T) {
--- a/modules/renter/contractor/downloader.go
+++ b/modules/renter/contractor/downloader.go
@@ -175,6 +175,15 @@ func (c *Contractor) Downloader(id types.FileContractID, cancel <-chan struct{})
 		}
 	}
 
+	// Increase Successful/Failed interactions accordingly
+	defer func() {
+		if err != nil {
+			c.hdb.IncrementFailedInteractions(contract.HostPublicKey)
+		} else {
+			c.hdb.IncrementSuccessfulInteractions(contract.HostPublicKey)
+		}
+	}()
+
 	// create downloader
 	d, err := proto.NewDownloader(host, contract, cancel)
 	if proto.IsRevisionMismatch(err) {
--- a/modules/renter/contractor/downloader.go
+++ b/modules/renter/contractor/downloader.go
@@ -175,17 +175,8 @@ func (c *Contractor) Downloader(id types.FileContractID, cancel <-chan struct{})
 		}
 	}
 
-	// Increase Successful/Failed interactions accordingly
-	defer func() {
-		if err != nil {
-			c.hdb.IncrementFailedInteractions(contract.HostPublicKey)
-		} else {
-			c.hdb.IncrementSuccessfulInteractions(contract.HostPublicKey)
-		}
-	}()
-
 	// create downloader
-	d, err := proto.NewDownloader(host, contract, cancel)
+	d, err := proto.NewDownloader(host, contract, c.hdb, cancel)
 	if proto.IsRevisionMismatch(err) {
 		// try again with the cached revision
 		c.mu.RLock()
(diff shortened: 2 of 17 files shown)


modules/renter/contractor/contractor_test.go
```diff
@@ -186,6 +188,8 @@ func (stubHostDB) ActiveHosts() (hs []modules.HostDBEntry)
 func (stubHostDB) Host(types.SiaPublicKey) (h modules.HostDBEntry, ok bool)         { return }
 func (stubHostDB) PublicKey() (spk types.SiaPublicKey)                              { return }
 func (stubHostDB) RandomHosts(int, []types.SiaPublicKey) (hs []modules.HostDBEntry) { return }
+func (stubHostDB) IncrementSuccessfulInteractions(key types.SiaPublicKey)           {}
+func (stubHostDB) IncrementFailedInteractions(key types.SiaPublicKey)               {}
```
DavidVorick: I didn't even know this style was legal code.

modules/renter/contractor/editor.go
```diff
@@ -104,9 +104,19 @@ func (he *hostEditor) Close() error {
 }
 
 // Upload negotiates a revision that adds a sector to a file contract.
-func (he *hostEditor) Upload(data []byte) (crypto.Hash, error) {
+func (he *hostEditor) Upload(data []byte) (_ crypto.Hash, err error) {
```
DavidVorick: Having this change alone makes me nervous, because you are depending on the values of these variables in a `defer` check. But I'm pretty sure if you do something like `return errors.New("an error")` the deferred code will miss the fact that you returned an error. Shadowed variables could also be an issue.

So, all the rest of the return statements in this function need to be updated to just say `return`, I think.

modules/renter/contractor/contractor_test.go
ChrisSchinnerl: oh.. didn't even notice I forgot the return. I will change that to match the style of the other declarations.

modules/renter/contractor/editor.go
ChrisSchinnerl: To be honest the fact that checking a value after a return is possible in go kind of blew my mind. I noticed this syntax the first time in the contractor's editor.go where it says `// release lock early if function returns an error` and I programmed a minimal example to check that this really works.
`func check() (i int) {
	defer func() {
		fmt.Println(i)
	}()
	i = 0
	return 1
}`
Which prints 1. I would still change it to a simple `return` if you want.

modules/renter/contractor/editor.go
```diff
@@ -130,9 +140,19 @@ func (he *hostEditor) Upload(data []byte) (crypto.Hash, error) {
...
+	defer func() {
+		if err != nil {
+			he.contractor.hdb.IncrementFailedInteractions(he.contract.HostPublicKey)
+		} else {
+			he.contractor.hdb.IncrementSuccessfulInteractions(he.contract.HostPublicKey)
+		}
+	}()
+
 	if he.invalid {
 		return errInvalidEditor
```
DavidVorick: This check needs to happen after the deferred func, because it's not a fault of the host - `he.invalid` indicates I believe that the renter has done a renewal or is otherwise no longer using the contract. So we wouldn't want to increment the reliability stats at all.

modules/renter/hostdb/update.go
```diff
@@ -29,6 +29,23 @@ func findHostAnnouncements(b types.Block) (announcements []modules.HostDBEntry)
 	return
 }
 
+// updateHistoricInteractions adds the recent interactions to the historic interactions
+// and progresses the decay
+func (hdb *HostDB) updateHistoricInteractions() {
+	hosts := hdb.hostTree.All()
+	for _, host := range hosts {
```
DavidVorick: I'm wondering if this loop is going to be too slow when we have more hosts. Iterating through 1000 objects once per block is not so bad, but perhaps it'd be better to have a global decay value or something.

I ~think we are safe to leave it this way for now, but we might want to tag a TODO indicating performance concerns.

DavidVorick:
Looks good, except that the deferred statements are too heavy handed - they don't properly handle cases where errors may not be the fault of the host, and they don't properly check that the right error is being returned.

So those all need to be revisited. The core logic in the hostdb looks good to me, except for the performance concern I brought up. The fix for this performance issue is not a small amount of work I  think, so maybe we have to live with it.

DavidVorick:
The general fix though I think would be to keep a saved value in the hostdbentry indicating the block height that it was last updated. Then, when you update it for some other reason, if the block height has changed, you can run an `update` function on the host that correctly fills out all of the fields

modules/renter/hostdb/scan.go
```diff
@@ -230,9 +230,14 @@ func (hdb *HostDB) managedScanHost(entry modules.HostDBEntry) {
 	}()
 	if err != nil {
 		hdb.log.Debugf("Scan of host at %v failed: %v", netAddr, err)
+		// Increment failed host interactions
+		entry.RecentFailedInteractions++
```
ChrisSchinnerl: Forgot that those are not calling the helper.

DavidVorick:
I stand corrected, if an error is declared at the top of the function, and you return `errors.New`, the returned value will update before the `defer` statements run, resolving one of my fears related to this pull request.

modules/renter/contractor/editor.go
DavidVorick: I stand corrected

modules/renter/contractor/formcontract.go
```diff
@@ -67,7 +67,16 @@ func maxSectors(a modules.Allowance, hdb hostDB, tp transactionPool) (uint64, er
...
-func (c *Contractor) managedNewContract(host modules.HostDBEntry, numSectors uint64, endHeight types.BlockHeight) (modules.RenterContract, error) {
+func (c *Contractor) managedNewContract(host modules.HostDBEntry, numSectors uint64, endHeight types.BlockHeight) (_ modules.RenterContract, err error) {
+	// Increase Successful/Failed interactions accordingly
+	defer func() {
+		if err != nil {
+			c.hdb.IncrementFailedInteractions(host.PublicKey)
+		} else {
+			c.hdb.IncrementSuccessfulInteractions(host.PublicKey)
+		}
+	}()
```
DavidVorick: This is still happening too early. It's not a failed interaction technically if the host is too expensive (a separate part of the logic handles that case), and it's definitely not a failed interaction if the wallet returns an error for `c.wallet.NextAddress()`

modules/renter/contractor/renew.go
```diff
@@ -12,7 +12,16 @@ import (
...
-func (c *Contractor) managedRenew(contract modules.RenterContract, numSectors uint64, newEndHeight types.BlockHeight) (modules.RenterContract, error) {
+func (c *Contractor) managedRenew(contract modules.RenterContract, numSectors uint64, newEndHeight types.BlockHeight) (_ modules.RenterContract, err error) {
+	// Increase Successful/Failed interactions accordingly
+	defer func() {
+		if err != nil {
+			c.hdb.IncrementFailedInteractions(contract.HostPublicKey)
+		} else {
+			c.hdb.IncrementSuccessfulInteractions(contract.HostPublicKey)
+		}
+	}()
```
DavidVorick: Same here, this function should be deferred until after the networking starts.

modules/renter/hostdb/scan.go
```diff
@@ -228,11 +228,19 @@ func (hdb *HostDB) managedScanHost(entry modules.HostDBEntry) {
 		copy(pubkey[:], pubKey.Key)
 		return crypto.ReadSignedObject(conn, &settings, maxSettingsLen, pubkey)
 	}()
+	// Update historic interactions of entry if necessary
+	updateHostsHistoricInteractions(&entry, hdb.cs.Height())
```
DavidVorick: This looks good

DavidVorick:
Ok, still a few places where the `defer` statements need to be moved around, and then also we need some testing. I may pick up and finish this PR tomorrow so that we can get through the price gouging stuff in the contractor - I think there will be a merge conflict otherwise.

ChrisSchinnerl:
Yes, sorry. I ran out of time last night so I focused on the performance of the update but I am taking a look at it right now.

ChrisSchinnerl:
The error handling should be more accurate now except for the calls to `Dial` which will still be counted towards the failed interactions. This could also be a bit tricky to fix since there are many different reasons for that to fail which might be either the renter's or host's fault.

DavidVorick:
@ChrisSchinnerl the hostdb actually has a field `online` - you can resolve a lot of the `Dial` false negatives by checking checking the `online` field when adding the negative feedback. If the hostdb is not online, just assume a false negative and don't record anything.

ChrisSchinnerl:
Ahh ok thanks! I will do that.

modules/renter/hostdb/hostdb.go
```diff
@@ -238,3 +272,41 @@ func (hdb *HostDB) Host(spk types.SiaPublicKey) (modules.HostDBEntry, bool) {
...
+
+// IncrementFailedInteractions increments the number of failed interactions with a host for a given key
+func (hdb *HostDB) IncrementFailedInteractions(key types.SiaPublicKey) {
+	hdb.mu.Lock()
+	defer hdb.mu.Unlock()
+
+	host, haveHost := hdb.hostTree.Select(key)
+	if !haveHost {
+		return
+	}
```
DavidVorick: Change to `if !haveHost || !hdb.online {` - this will prevent the host from being docked due to connection issues on the renter side. It's not foolproof b/c the `online` value is only updated every 30 seconds or so, but if we ever update the `online` value to be more responsive, we'll get the benefit automatically.

modules/renter/proto/editor.go
```diff
@@ -255,12 +265,22 @@ func (he *Editor) Modify(oldRoot, newRoot crypto.Hash, offset uint64, newData []
...
+func NewEditor(host modules.HostDBEntry, contract modules.RenterContract, currentHeight types.BlockHeight, hdb hostDB, cancel <-chan struct{}) (_ *Editor, err error) {
 	// check that contract has enough value to support an upload
 	if len(contract.LastRevision.NewValidProofOutputs) != 2 {
 		return nil, errors.New("invalid contract")
 	}
 
+	// Increase Successful/Failed interactions accordingly
+	defer func() {
+		// a revision mismatch is not necessarily the host's fault
+		if err != nil && !IsRevisionMismatch(err) {
```
DavidVorick: hmm. A revision mismatch can be the host's fault too though. Generally speaking the renter is constructed so that as long as the host is honest, there will not be a revision mismatch. It's probably fine to count a mismatch as a failed interaction

modules/renter/contractor/downloader.go
```diff
@@ -189,7 +189,11 @@ func (c *Contractor) Downloader(id types.FileContractID, cancel <-chan struct{})
 		}
 		c.log.Printf("host %v has different revision for %v; retrying with cached revision", contract.NetAddress, contract.ID)
 		contract.LastRevision = cached.Revision
-		d, err = proto.NewDownloader(host, contract, cancel)
+		d, err = proto.NewDownloader(host, contract, c.hdb, cancel)
+		// needs to be handled separately since a revision mismatch is not automatically a failed interaction
+		if proto.IsRevisionMismatch(err) {
+			c.hdb.IncrementFailedInteractions(host.PublicKey)
+		}
```
DavidVorick: oh I see how you are handling this. This is correct.

modules/renter/hostdb/hostdb.go
```diff
@@ -17,6 +17,7 @@ import (
 	"github.com/NebulousLabs/Sia/persist"
 	siasync "github.com/NebulousLabs/Sia/sync"
 	"github.com/NebulousLabs/Sia/types"
+	"math"
```
DavidVorick: needs to be added to the first batch of imports. Sorting is in 3 batches: standard library imports, followed by internal imports (github.com/NebulousLabs/Sia/*) followed by external imports (github.com/NebulousLabs/fastrand for example)

PR #2061, opened 2017-07-01
repair concurrency related ndf in persist package

DavidVorick:


lukechampine:
couldn't we manually trigger the panic by adding an entry to `activeFiles` in the test? I would prefer that to inserting a `runtime.Gosched` call into `SaveJSON`

DavidVorick:
hmm. That treads into the realm of implementation-specific testing though

lukechampine:
closing because the relevant test was removed

PR #2062, opened 2017-07-01
skip test with poor concurrency design

DavidVorick:


lukechampine:
`TestUnlockConcurrent` isn't great either. I think both can be safely removed at this point. The cost of NDF outweighs their usefulness.

PR #2066, opened 2017-07-03
New gateway handshake

lukechampine:
This is a pretty messy refactoring. It would be nice if we could split the compat code into a separate file or something. Even then, `peers.go` is too long for my liking.

One uncertainty I had was around hostname validation. In the new handshake, the gateways exchange their full NetAddresses. They will then check that the "claimed" NetAddress hostname matches the hostname of `conn.RemoteAddr`. (The rationale for this is to prevent a gateway from tricking you into DDoSing an IP other than their own.)
However, while running the antfarm, connections would fail because the connecting gateway thinks its NetAddress is its external IP, but the accepting gateway sees `conn.RemoteAddr == 127.0.0.1`.
I rectified this by having the gateway send `conn.LocalAddr` instead of its external NetAddress, but I'm not sure if that solution will work for all network architectures. There may be a valid instance where the "true" hostname doesn't match the hostname returned by `conn.RemoteAddr`. For now, I'm erring on the side of preventing DDoS.

### DIFF
--- a/build/version.go
+++ b/build/version.go
@@ -7,7 +7,7 @@ import (
 
 const (
 	// Version is the current version of siad.
-	Version = "1.3.0"
+	Version = "1.4.0"
 
 	// MaxEncodedVersionLength is the maximum length of a version string encoded
 	// with the encode package. 100 is much larger than any version number we send
--- a/build/version.go
+++ b/build/version.go
@@ -7,7 +7,7 @@ import (
 
 const (
 	// Version is the current version of siad.
-	Version = "1.4.0"
+	Version = "1.3.0"
 
 	// MaxEncodedVersionLength is the maximum length of a version string encoded
 	// with the encode package. 100 is much larger than any version number we send
--- a/modules/gateway/consts.go
+++ b/modules/gateway/consts.go
@@ -15,7 +15,7 @@ const (
 	// was altered to include the ID of the genesis block, the gateway's
 	// unique ID, and whether a connection is desired. This version also uses
 	// smux instead of muxado for stream multiplexing.
-	sessionUpgradeVersion = "1.3.0"
+	sessionUpgradeVersion = "1.4.0"
 
 	// maxLocalOutbound is currently set to 3, meaning the gateway will not
 	// consider a local node to be an outbound peer if the gateway already has
--- a/modules/gateway/consts.go
+++ b/modules/gateway/consts.go
@@ -16,7 +16,7 @@ const (
 	// was altered to include the ID of the genesis block, the gateway's
 	// unique ID, and whether a connection is desired. This version also uses
 	// smux instead of muxado for stream multiplexing.
-	sessionUpgradeVersion = "1.4.0"
+	sessionUpgradeVersion = "1.3.0"
 
 	// maxLocalOutbound is currently set to 3, meaning the gateway will not
 	// consider a local node to be an outbound peer if the gateway already has
--- a/modules/gateway/consts.go
+++ b/modules/gateway/consts.go
@@ -4,6 +4,7 @@ import (
 	"time"
 
 	"github.com/NebulousLabs/Sia/build"
+	"github.com/NebulousLabs/Sia/modules"
 )
 
 const (
@@ -33,10 +34,9 @@ const (
 	// pre-hardfork.
 	minAcceptableVersion = "0.4.0"
 
-	// EncodedSessionHeaderLength is the length of a session header encoded
-	// with the encode package.
-	// sizeof(BlockID) + sizeof(gatewayID) + sizeof(bool) == 32 + 8 + 1 = 41
-	EncodedSessionHeaderLength = 41
+	// maxEncodedSessionHeaderSize is the maximum allowed size of an encoded
+	// sessionHeader object.
+	maxEncodedSessionHeaderSize = 40 + modules.MaxEncodedNetAddressLength
 
 	// saveFrequency defines how often the gateway saves its persistence.
 	saveFrequency = time.Minute * 2
--- a/modules/gateway/peers.go
+++ b/modules/gateway/peers.go
@@ -41,13 +41,11 @@ type peer struct {
 
 // sessionHeader is sent after the initial version exchange. It prevents peers
 // on different blockchains from connecting to each other, and prevents the
-// gateway from connecting to itself. The receiving peer can set WantConn to
-// false to refuse the connection, and the initiating peer can set WantConn to
-// false if they merely want to confirm that a node is online.
+// gateway from connecting to itself.
 type sessionHeader struct {
-	GenesisID types.BlockID
-	UniqueID  gatewayID
-	WantConn  bool
+	GenesisID  types.BlockID
+	UniqueID   gatewayID
+	NetAddress modules.NetAddress
 }
 
 func (p *peer) open() (modules.PeerConn, error) {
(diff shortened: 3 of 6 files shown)


build/version.go
```diff
@@ -7,7 +7,7 @@ import (
 
 const (
 	// Version is the current version of siad.
-	Version = "1.3.0"
+	Version = "1.4.0"
```
DavidVorick: 1.3.0 is fine I think, we're going to merge this branch for v1.3.0 I think.

modules/gateway/consts.go
```diff
@@ -15,7 +16,7 @@ const (
 	// was altered to include the ID of the genesis block, the gateway's
 	// unique ID, and whether a connection is desired. This version also uses
 	// smux instead of muxado for stream multiplexing.
-	sessionUpgradeVersion = "1.3.0"
+	sessionUpgradeVersion = "1.4.0"
```
DavidVorick: v1.3.0 is fine

modules/gateway/peers.go
```diff
@@ -162,40 +153,96 @@ func (g *Gateway) threadedAcceptConn(conn net.Conn) {
...
+	}
+	return nil
+}
 
-	g.mu.Lock()
-	defer g.mu.Unlock()
+// managedAcceptConnv100Peer accepts connection requests from peers >= v1.4.0.
+// The requesting peer is added as a node and a peer. The peer is only added if
+// a nil error is returned.
+func (g *Gateway) managedAcceptConnv140Peer(conn net.Conn, remoteVersion string) error {
```
DavidVorick: v130

modules/gateway/peers.go
```diff
@@ -403,82 +400,73 @@ func acceptConnVersionHandshake(conn net.Conn, version string) (remoteVersion st
...
+	}
 
-	if err := g.saveSync(); err != nil {
-		g.log.Println("ERROR: Unable to save new outbound peer to gateway:", err)
+	return remoteHeader, nil
+}
+
+// managedConnectv140Peer connects to peers >= v1.4.0. The peer is added as a
+// node and a peer. The peer is only added if a nil error is returned.
+func (g *Gateway) managedConnectv140Peer(conn net.Conn, remoteVersion string, remoteAddr modules.NetAddress) error {
```
DavidVorick: v1.3.0

modules/renter/hostdb/hostweight.go
```diff
@@ -195,9 +195,12 @@ func storageRemainingAdjustments(entry modules.HostDBEntry) float64 {
...
 func versionAdjustments(entry modules.HostDBEntry) float64 {
 	base := float64(1)
-	if build.VersionCmp(entry.Version, "1.3.1") < 0 {
+	if build.VersionCmp(entry.Version, "1.4.1") < 0 {
 		base = base * 0.99999 // Safety value to make sure we update the version penalties every time we update the host.
 	}
+	if build.VersionCmp(entry.Version, "1.3.1") < 0 {
+		base = base * 0.9
+	}
 	if build.VersionCmp(entry.Version, "1.3.0") < 0 {
```
DavidVorick: version repairs

modules/gateway/peers_test.go
```diff
@@ -236,38 +234,46 @@ func TestListen(t *testing.T) {
...
-	err = connectPortHandshake(conn, addr.Port())
+	_, err = readSessionHeader(conn, header)
 	if err != nil {
 		t.Fatal(err)
 	}
 
 	// g should add the peer
-	var ok bool
-	for !ok {
+	err = build.Retry(10, 20*time.Millisecond, func() error {
```
DavidVorick: you are probably going to want more time than this, Travis can sleep a thread for as long as 5 seconds iirc

modules/gateway/peers_test.go
```diff
@@ -236,38 +234,46 @@ func TestListen(t *testing.T) {
...
+		t.Fatal(err)
 	}
 
+	// Disconnect. Now that connection has been established, need to shutdown
+	// via the stream multiplexer.
 	newClientStream(conn, build.Version).Close()
 
 	// g should remove the peer
-	for ok {
+	err = build.Retry(10, 20*time.Millisecond, func() error {
```
DavidVorick: same here

MrStrong:
Hi I'm a new user. 

So I've got a node running at home and a test node running on AWS, and both are behind NATs. I've come accross the situation where after running:

`gateway connect aws.external.ip.addr:9981` from home, or `gateway connect home.external.ip.addr:9981` on AWS
both return the following result

`Could not add peer: peer rejected our header: claimed hostname (local.ip.addr) does not match conn.RemoteAddr (external.ip.addr)`

lukechampine:
hmm, does the home node have its ports forwarded? Does `siac gateway address` display the correct (external) address?

Perhaps in the handshake we should just ignore the claimed address and use the `RemoteAddr`. We could still log the difference with a warning or something.

MrStrong:
yes, forwarding is working fine and accepting connections, and `gateway address` does return my correct external address, on both hosts.

I was wondering if the sessionHeader was getting its host (claimed) IP address from a different place to the gateway module.


Yeah, thinking about the original merge is it even important to check the claimed host IP in the Sia header? If an attacker was going to forge that why not spoof the src addr in the IP header too. Not knowing much how the protocol works, but if an attacker did spoof their IP and receiving node would quickly drop the traffic since the handshake would fail? And some rate limiting can be used to mitigate spamming.

One more question you say hostname merge request comment, but as far as I can tell there shouldn't be any domain names here?

lukechampine:
The claimed IP is determined first by asking your router (via UPnP), and if that fails, dialing a third-party service (http://myexternalip.com/raw). It seems there a few legitimate network setups that would cause this claimed IP to not match the `RemoteAddr` as seen by the other peer.

I think the right move for now is to remove the check, and just use `RemoteAddr.Host() + ClaimedAddr.Port()`. (We definitely need to use the claimed port.)

I'm also wondering if this is related to https://github.com/NebulousLabs/Sia/issues/1970. Would be nice if we could fix that as well.

MrStrong:
sorry if i'm going off onto a tangent, but really do we even need to be transmitting any connection data in the Sia session header, shouldn't it just a be a unique client identifier (like a cookie in HTTP)? 
Since the TCP connection already contains all the necessary information, ie IP and port, the connection state can't even exist without it :P 

Re the attack vector, a client that changes IP or port will break the TCP connection and a new one will need to be recreated, so again I'm thinking the connection details are redundant to include in the sia header. 
I agree it is a valid concern but my understanding says that this is all already taken care of in TCP, no need to duplicate the work

lukechampine:
We need to send the port number so that the peer can dial back later. When you connect, they see the port you're connecting from, not the port you're listening on. For the same reason, I can imagine a scenario where the IP/hostname you're making connections from isn't the same as the IP you're listening for new connections on.

It's also worth pointing out that if you're just trying to force someone to ping another IP, you can accomplish that by sending your peers fake peer lists. The session header feels slightly different because you're specifically saying "this is **my** address" rather than "this is **an** address," but maybe that distinction doesn't matter much in practice.

dNetGuru:
@lukechampine I had the same problem as MrStrong running 1.3.0 however the problem persisted when I updated my client to 1.3.1 and then noticed that I could **only** manually connect to 1.3.1 hosts so I ended up digging into the source code to see why this is happening...

Much to my surprise I found that you are sending the interface address in the ```sessionHeader``` by calling ```conn.LocalAddr()``` (https://github.com/NebulousLabs/Sia/blob/master/modules/gateway/peers.go#L440 is one instance, there are 3 in total of which only two matter) which in case of a NATed address will be an invalid address and the remote host will reject it since the remote host sees the public IP.
To remedy this I replaced the local address with the public address (which will be the address that is visible to a remote node) and I can now connect to all node versions (https://github.com/dNetGuru/Sia/commit/a3dc2cc38b7cf420086ed86e071f608f7ac1e2db).

PR #2069, opened 2017-07-04
Add txn size limit to hardfork

lukechampine:
I also snuck some encoding optimizations in. Mostly just calling `MarshalSia` methods directly where we weren't before.

Regrettably, `MarshalSiaSize` had a bug; it was 8 bytes short for each `TransactionSignature`. This was fixed on my local `txn-size` branch but was not pushed before that PR was merged. Good call on adding the sanity check during debug mode, otherwise I wouldn't have caught this.

### DIFF
--- a/types/constants.go
+++ b/types/constants.go
@@ -42,11 +42,12 @@ var (
 
 	// Oak hardfork constants. Oak is the name of the difficulty algorithm for
 	// Sia following a hardfork at block 135e3.
-	OakHardforkBlock BlockHeight
-	OakDecayNum      int64
-	OakDecayDenom    int64
-	OakMaxRise       *big.Rat
-	OakMaxDrop       *big.Rat
+	OakHardforkBlock        BlockHeight
+	OakDecayNum             int64
+	OakDecayDenom           int64
+	OakMaxRise              *big.Rat
+	OakMaxDrop              *big.Rat
+	OakHardforkTxnSizeLimit = uint64(65536) // 64 KiB
 )
 
 // init checks which build constant is in place and initializes the variables
--- a/types/constants.go
+++ b/types/constants.go
@@ -47,7 +47,7 @@ var (
 	OakDecayDenom           int64
 	OakMaxRise              *big.Rat
 	OakMaxDrop              *big.Rat
-	OakHardforkTxnSizeLimit = uint64(65536) // 64 KiB
+	OakHardforkTxnSizeLimit = uint64(64e3) // 64 KB
 )
 
 // init checks which build constant is in place and initializes the variables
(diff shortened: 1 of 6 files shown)


types/constants.go
```diff
@@ -42,11 +42,12 @@ var (
...
-	OakDecayNum      int64
-	OakDecayDenom    int64
-	OakMaxRise       *big.Rat
-	OakMaxDrop       *big.Rat
+	OakHardforkBlock        BlockHeight
+	OakDecayNum             int64
+	OakDecayDenom           int64
+	OakMaxRise              *big.Rat
+	OakMaxDrop              *big.Rat
+	OakHardforkTxnSizeLimit = uint64(65536) // 64 KiB
```
DavidVorick: Convention for sizes is to use base 10. It's not really relevant, but given that our blocks are 2e6, transactions should probably be 64e3

types/encoding.go
```diff
@@ -19,7 +19,17 @@ func (b Block) MarshalSia(w io.Writer) error {
...
+	for i := range b.MinerPayouts {
+		b.MinerPayouts[i].MarshalSia(w)
+	}
+	encoding.WriteInt(w, len(b.Transactions))
+	for i := range b.Transactions {
+		if err := b.Transactions[i].MarshalSia(w); err != nil {
+			return err
+		}
+	}
+	return nil
```
DavidVorick: Can we add a sanity check here to verify equivalence? Our sanity checks have caught 2 bugs at this point, that's more than enough reason to keep adding them.

DavidVorick: That should also catch any mistakes with the coveredFields implementation.

DavidVorick:
Just those 2 comments, then this should be good to merge.

types/encoding.go
lukechampine: Unfortunately we can't (easily) add sanity checks to `MarshalSia` methods. The obvious way to do it is to compare to `encoding.Marshal`, but `encoding.Marshal` will call the object's `MarshalSia` method if it's defined, so you get an infinite loop.

In addition, this method writes directly to `w`. In order to do a sanity check, we would need to write everything to a byte slice, compare it to the `encoding.Marshal` output, and then flush it to `w`. It's doable to but it negates some of the performance gain.

DavidVorick: Can define a MarshalSiaOld function that has the identical implementation to the previous implementation of MarshalSia. Then you can call it and have it pass back a writer. If !DEBUG, all it does is pass back the same writer you sent it, and nothing more. If DEBUG, it'll implement a writer that does a sanity check on the byte stream.

Something like that. Again, we've caught enough bugs with these, and there's enough code change in this PR (covered fields as well as block as well as sighash) that I think we really need a sanity check.

PR #2070, opened 2017-07-04
Automatic host price adjustments

ChrisSchinnerl:


modules/host/consts.go
```diff
@@ -53,11 +53,9 @@ var (
...
 	defaultCollateralBudget = types.SiacoinPrecision.Mul64(100e3)
 
-	// defaultContractPrice defines the default price of creating a contract
-	// with the host. The default is set to 30 siacoins, which the file
-	// contract revision can have 15 siacoins put towards it, and the storage
-	// proof can have 15 siacoins put towards it.
-	defaultContractPrice = types.SiacoinPrecision.Mul64(3) // 3 siacoins
+	// with the host. The default is set to 0.05 siacoins and will be adjusted
+	// automatically if it needs to be higher
+	defaultContractPrice = types.SiacoinPrecision.Mul64(5).Div64(100) // 0.05 siacoins
```
Cinerar: I think this defaultContractPrice is designed to cover transactionFees.  maybe they should be somehow tied?

modules/host/host_test.go
```diff
@@ -422,6 +422,56 @@ func TestSetAndGetInternalSettings(t *testing.T) {
 	ht.host = rebootHost
 }
 
+// TestRecentContractPricesLength checks
```
johnathanhowell: docstring

modules/host/consts.go
DavidVorick: Originally it was, now I'm thinking the default can just be set to zero. The purpose of this pull request (haven't reviewed it yet) iirc is to make it so that the host will automatically bump the contract price if mincontractprice set by the settings is too low to cover the user fees. This protects the host from losing money just for forming contracts, while allowing the default to always be 'as low as practically possible'

modules/host/consts.go
```diff
@@ -53,11 +53,9 @@ var (
 	// furious for losing access to it for a few weeks.
 	defaultCollateralBudget = types.SiacoinPrecision.Mul64(100e3)
 
-	// defaultContractPrice defines the default price of creating a contract
-	// with the host. The default is set to 30 siacoins, which the file
-	// contract revision can have 15 siacoins put towards it, and the storage
-	// proof can have 15 siacoins put towards it.
-	defaultContractPrice = types.SiacoinPrecision.Mul64(3) // 3 siacoins
+	// with the host. The default is set to 0 siacoins and will be adjusted
```
avahowell: seems like there's a line missing from this docstring (the line starting with `// defaultContractPrice defines...`)

modules/host/host_test.go
```diff
@@ -422,6 +422,76 @@ func TestSetAndGetInternalSettings(t *testing.T) {
...
+	ht.miner.AddBlock()
+
+	// check if recentContractPrices grew by 1 entry
+	if len(ht.host.recentContractPrices) != initialLength+1 {
+		t.Error("recentContractPrices doesn't contain additional entry after new block")
+	}
+
+	// add 150 more blocks
+	for i := 0; i < 150; i++ {
+		ht.miner.AddBlock()
```
avahowell: missing error check here

DavidVorick:
Closing in favor of #2771

PR #2073, opened 2017-07-04
Contractor Combining

DavidVorick:
This PR is the first step in eliminating the price gouging issue that we're seeing with our hosts. I've tried to keep things functionally equivalent at this point to how they used to run, which means that we haven't solved all that many problems.

But, instead of having 3 separate loops responsible for creating contracts, I've moved them into a single loop that handles everything.

Creating an allowance is no longer a blocking call. This broke a lot of tests, but I have since fixed them. There may be an NDF or two lying around still.

This is the first stage, and should be merged on its own to prevent this PR from blowing up the way my last PR did. The next step will be to add a IsGoodForUpload and IsGoodForRenew field to the RenterContract, which will allow us to disable hosts that are evil.

That should do it for v1.3.0. There's still a lot more things we should be doing with the allowance management, but it'll have to wait until v1.3.1


modules/renter/contractor/update_test.go
```diff
@@ -142,6 +153,15 @@ func TestIntegrationRenewInvalidate(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	err = build.Retry(50, 100*time.Millisecond, func() error {
+		if len(c.Contracts()) == 0 {
+			return errors.New("contracts were not formed")
+		}
+		return nil
+	})
```
lukechampine: This is common enough to warrant a `waitForContracts` helper function

modules/renter/contractor/contracts.go
```diff
@@ -0,0 +1,395 @@
...
+		// Quit the loop if we've replaced all needed contracts.
+		neededContracts--
+		if neededContracts < 1 {
+			break
+		}
+
+		// Soft sleep before making the next contract.
+		select {
+		case <-c.tg.StopChan():
+			return
```
lukechampine: This is common enough that we should consider a `tg.SoftSleep` method

modules/renter/contractor/contracts.go
```diff
@@ -0,0 +1,395 @@
...
+
+	// Count the number of contracts, and add more where they are needed.
+	//
+	// TODO: Use the IsGoodForUpload logic.
+	c.mu.Lock()
+	neededContracts := int(c.allowance.Hosts) - len(c.onlineContracts())
+	c.mu.Unlock()
+	// Nothing to do if there are no needed contracts.
+	if neededContracts < 1 {
+		return
```
lukechampine: Nitpick, but I prefer <= 0 here.

modules/renter/contractor/contracts.go
```diff
@@ -0,0 +1,395 @@
...
+	}
+
+	// Loop through the contracts and renew them one-by-one.
+	for _, id := range renewSet {
+		// Quit / return in the event of shutdown.
+		select {
+		case <-c.tg.StopChan():
+			return
+		default:
+		}
```
lukechampine: Given that we soft-sleep at the bottom of the loop, is this necessary?

DavidVorick: Yeah if shudown is called sometime between starting to form a contract and hitting this part of the loop, we'd want to quit immediately.

DavidVorick: Maybe we could add a if tg.Closed() { return } type thing, which brings it from 5 lines down to 3

modules/renter/contractor/update_test.go
DavidVorick: it'd need to be implemented in multiple packages, but perhaps worthwhile still

modules/renter/contractor/contracts.go
DavidVorick: I am in agreement. Perhaps in another PR, perhaps the same one where we drop the `Flush` function and transit to the NebulousLabs/threadgroup package.

modules/renter/contractor/contracts.go
lukechampine: we already have a `tg.isStopped` method. It used to be exported but then we unexported it. I forget why.

modules/renter/contractor/update_test.go
lukechampine: One version for the contractor and one for the API should be ok

modules/renter/contractor/contracts.go
```diff
@@ -0,0 +1,388 @@
...
+	// Quit in the event of shutdown.
+	select {
+	case <-c.tg.StopChan():
+		return
+	default:
+	}
+
+	// Count the number of contracts, and add more where they are needed.
+	//
+	// TODO: Use the IsGoodForUpload logic.
```
tlightsky: cool, i'll rebase to this pr and then use this new flag
now the 1 sector contract strategy seems not work

PR #2080, opened 2017-07-05
Add API call to retrieve raw consensus change

ChrisSchinnerl:
This PR adds the `"/consensus/change/:id"` GET API call.
It takes a hex encoded consensus change id and returns a raw consensus change + the id of the next change if there is any. Otherwise it returns the 32 byte long zero id.
The first change can be retrieved providing the zero id.

### DIFF
--- a/api/consensus.go
+++ b/api/consensus.go
@@ -6,6 +6,8 @@ import (
 
 	"github.com/NebulousLabs/Sia/types"
 
+	"fmt"
+	"github.com/NebulousLabs/Sia/modules"
 	"github.com/julienschmidt/httprouter"
 )
 
@@ -17,6 +19,21 @@ type ConsensusGET struct {
 	CurrentBlock types.BlockID     `json:"currentblock"`
 	Target       types.Target      `json:"target"`
 	Difficulty   types.Currency    `json:"difficulty"`
+}
+
+// ConsensusChangeGet contains a raw consensus set with tags to support idiomatic json encodings.
+type ConsensusChangeGet struct {
+	ID                         string                             `json:"id"`
+	NextID                     string                             `json:"nextid"`
+	RevertedBlocks             []types.Block                      `json:"revertedblocks"`
+	AppliedBlocks              []types.Block                      `json:"appliedblocks"`
+	SiacoinOutputDiffs         []modules.SiacoinOutputDiff        `json:"siacoinoutputdiffs"`
+	FileContractDiffs          []modules.FileContractDiff         `json:"filecontractdiffs"`
+	SiafundOutputDiffs         []modules.SiafundOutputDiff        `json:"siafundoutputdiffs"`
+	DelayedSiacoinOutputDiffs  []modules.DelayedSiacoinOutputDiff `json:"delayedsiacoinoutputdiffs"`
+	SiafundPoolDiffs           []modules.SiafundPoolDiff          `json:"siafundpooldiffs"`
+	ChildTarget                types.Target                       `json:"childtarget"`
+	MinimumValidChildTimestamp types.Timestamp                    `json:"minimumvalidchildtimestamp"`
 }
 
 // consensusHandler handles the API calls to /consensus.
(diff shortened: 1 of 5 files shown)


api/consensus.go
```diff
@@ -6,6 +6,8 @@ import (
 
 	"github.com/NebulousLabs/Sia/types"
 
+	"fmt"
+	"github.com/NebulousLabs/Sia/modules"
```
lukechampine: imports should be grouped like so:
```
import (
    "stdlib"

    "sub-packages of NebulousLabs/Sia"

    "external packages"
)
```

api/consensus.go
```diff
@@ -19,6 +21,21 @@ type ConsensusGET struct {
 	Difficulty   types.Currency    `json:"difficulty"`
 }
 
+// ConsensusChangeGet contains a raw consensus set with tags to support idiomatic json encodings.
+type ConsensusChangeGet struct {
+	ID                         string                             `json:"id"`
+	NextID                     string                             `json:"nextid"`
```
lukechampine: better to define a `MarshalJSON` method on `modules.ConsensusChangeID`

modules/consensus.go
```diff
@@ -243,6 +243,11 @@ type (
 		// allowing for garbage collection and rescanning. If the subscriber is
 		// not found in the subscriber database, no action is taken.
 		Unsubscribe(ConsensusSetSubscriber)
+
+		// GetConsensusChange returns the ConsensusChange that corresponds to
+		// a given ConsensusChangeID. Primarily used for API calls as modules
+		// should Subscribe to the Consensus Set instead
+		GetConsensusChange(id ConsensusChangeID) (ConsensusChange, ConsensusChangeID, error)
```
lukechampine: Go convention is to not include "Get" in getter methods. (i.e. this should just be `ConsensusChange`)

modules/consensus/consensusset.go
```diff
@@ -358,3 +358,45 @@ func (cs *ConsensusSet) StorageProofSegment(fcid types.FileContractID) (index ui
 	})
 	return index, err
 }
+
+// GetConsensusChange returns the ConsensusChange that corresponds to a given ConsensusChangeID. Primarily used for API
+// calls as modules should Subscribe to the Consensus Set instead
+func (cs *ConsensusSet) GetConsensusChange(id modules.ConsensusChangeID) (change modules.ConsensusChange,
+	next modules.ConsensusChangeID, err error) {
```
lukechampine: we have a convention of never splitting the function declaration across lines, no matter how long they are. :P

api/consensus.go
ChrisSchinnerl: ahhh sry I know, but I keep forgetting to double check after the IDE adds an import -.-

modules/consensus/consensusset.go
ChrisSchinnerl: challenge accepted

modules/consensus.go
```diff
@@ -260,3 +267,32 @@ func (cc ConsensusChange) Append(cc2 ConsensusChange) ConsensusChange {
...
+		return crypto.ErrHashWrongLen
+	}
+
+	// b[1 : len(b)-1] cuts off the leading and trailing `"` in the JSON string.
+	hBytes, err := hex.DecodeString(string(b[1 : len(b)-1]))
+	if err != nil {
+		return errors.New("could not unmarshal ConsensusChangeID: " + err.Error())
+	}
+	copy(ccid[:], hBytes)
+	return nil
```
ChrisSchinnerl: Is there a way in Go to avoid this by calling the marshalling functions of the `crypto.Hash` ? I did some searching but it seems that named types in Go are strictly separated and don't like casting.

lukechampine: indeed, you can do:
```go
id, err := scanHash(ps.ByName("id"))
if err != nil {
    // handle err
}
ccid := modules.ConsensusChangeID(id)
```
or if you want to get fancy:
```go
var id modules.ConsensusChangeID
err := (*crypto.Hash)(&id).LoadString(ps.ByName("id"))
```

ChrisSchinnerl: ahhh thanks. I didn't see `scanHash`.

TimonPeng:
How to get a consensus change id?

ChrisSchinnerl:
@TimonPeng by providing a "zero id" where all digits of the 32 byte hex encoded hash are 0.

TimonPeng:
Is the `0000000000000000000000000000000000000000000000000000000000000000 ` here is parent id?

ChrisSchinnerl:
`0000000000000000000000000000000000000000000000000000000000000000` is simply used to indicate that you request the first consensus change. In the following calls it will be replaced by the id returned from the first call.

TimonPeng:
I run it successfully. Can I get or compute consensus change id from parent id, or between two parent ids?

ChrisSchinnerl:
No, similar to a module subscribing to the consensus in siad, you should retrieve changes in the correct order. Of course you could store retrieved ids to use this call on any change you already visited.

TimonPeng:
Is a bug here? parent id is not the right block id.

`http://localhost:9980/consensus/change/5c8a8370795eceb5d615259cea2b60043db310517da1a8e985afd47daf5d1a36`

TimonPeng:
LGTM.

PR #2081, opened 2017-07-05
Update hostdb scan quanity/sleep time

S-anasol:
Current default settings cant handle hostdb grow.
So hostdb have a lot outdated data.

Adjusted settings for siahub wallet.
Enough to keep fresh db for now.

After update we have big fall since many hosts not scanned fast/properly due big hostdb.
https://siahub.info/network


S-anasol:
Actually problem is that scan doesnt cover DB size and every scan get only top N hosts.

https://github.com/NebulousLabs/Sia/blob/ec3fbc0dd35bc2ad0b476ab0c96d6e4ec3317045/modules/renter/hostdb/scan.go#L304

So in this moment we have some good hosts that actually online and work, but not scanned due quanity limit.

modules/renter/hostdb/consts.go
```diff
@@ -39,15 +39,15 @@ var (
 	// hostCheckupQuantity specifies the number of hosts that get scanned every
 	// time there is a regular scanning operation.
 	hostCheckupQuantity = build.Select(build.Var{
-		Standard: int(200),
+		Standard: int(500),
```
S-anasol: actually outdated limit, it will not cover all online hosts already

Cinerar:
I believe this is not the best way to go.  (as a temporary fix i think this is ok)
As long as i understand this aproach can leave best host (the one that changed settings recently) with outdated info and almoust never scan it again.  

Limiting scanned hosts centralize the network, as long as only best hosts will be regulary rescanned. 
I believe that Score of the host can be tied to period of rescan. This way even not so good hosts will be rescanned sooner or later and there will be not so much network overload.

DavidVorick:
I'm going to go ahead and close this as a pull request, I think a full discussion is needed when searching for better options. At 1000 hosts or so, I think we're above what we should be scanning every block. I like the 200 hosts limit per 8 hours I think, we probably just need a more intelligent way to scan them.

The most important thing is that we're aware of all the hosts that might be interesting to us.

PR #2085, opened 2017-07-05
[WIP] FIX #1603; Add customizable log directory

### LINKED ISSUE #1603: [enhencement] allow for siad to write logs to /var/log or elsewhere
It would be nice if the baselogdir for siad could be set as an arg on commandline. 

Cheers.

Eriner:
I've authored a small patch here, but would like to get your input before moving forward, as I'm not all that familiar with the project. FWIW, the patch works as expected, minus argument/input path validation/creation, which is something I'd like your input on.

My understanding, after looking at some of the code for a few minutes, was the best place to place persist "setter" was in the `daemon.go` file. I had originally attempted to place it at the beginning of the `startDaemon` function, after it calls `processConfig`, but this caused a race due to the `go profile.StartContinuousProfile()` routine calling `NewFileLogger()` before the `LogDir` var was set. Thus, I had to place it before that call, in the `startDaemonCmd` function. If this is an issue or there's a better way to go about this, please let me know.

As I mention in the WIP patch TODO, I know path/directory validation and creation is currently missing (it will fail on a non-existent directory) and the native string concat for the filepath is naïve -- this is just a proof-of-concept and to get feedback to see if I'm on the right track here.

### DIFF
--- a/modules/host/contractmanager/dependencies.go
+++ b/modules/host/contractmanager/dependencies.go
@@ -16,7 +16,6 @@ import (
 var (
 	mockErrListen       = errors.New("simulated Listen failure")
 	mockErrLoadFile     = errors.New("simulated LoadFile failure")
-	mockErrMkdirAll     = errors.New("simulated MkdirAll failure")
 	mockErrNewLogger    = errors.New("simulated NewLogger failure")
 	mockErrOpenDatabase = errors.New("simulated OpenDatabase failure")
 	mockErrReadFile     = errors.New("simulated ReadFile failure")
@@ -62,10 +61,6 @@ type (
 
 		// loadFile allows the host to load a persistence structure form disk.
 		loadFile(persist.Metadata, interface{}, string) error
-
-		// mkdirAll gives the host the ability to create chains of folders
-		// within the filesystem.
-		mkdirAll(string, os.FileMode) error
 
 		// newLogger creates a logger that the host can use to log messages and
 		// write critical statements.
@@ -186,12 +181,6 @@ func (pd *productionDependencies) init() {
 // loadFile allows the host to load a persistence structure form disk.
 func (productionDependencies) loadFile(m persist.Metadata, i interface{}, s string) error {
 	return persist.LoadJSON(m, i, s)
-}
-
-// mkdirAll gives the host the ability to create chains of folders within the
-// filesystem.
-func (productionDependencies) mkdirAll(s string, fm os.FileMode) error {
-	return os.MkdirAll(s, fm)
 }
 
 // newLogger creates a logger that the host can use to log messages and write
--- a/modules/host/dependencies.go
+++ b/modules/host/dependencies.go
@@ -15,7 +15,6 @@ import (
 var (
 	mockErrListen       = errors.New("simulated Listen failure")
 	mockErrLoadFile     = errors.New("simulated LoadFile failure")
-	mockErrMkdirAll     = errors.New("simulated MkdirAll failure")
 	mockErrNewLogger    = errors.New("simulated NewLogger failure")
 	mockErrOpenDatabase = errors.New("simulated OpenDatabase failure")
 	mockErrReadFile     = errors.New("simulated ReadFile failure")
@@ -41,10 +40,6 @@ type (
 
 		// loadFile allows the host to load a persistence structure form disk.
 		loadFile(persist.Metadata, interface{}, string) error
-
-		// mkdirAll gives the host the ability to create chains of folders
-		// within the filesystem.
-		mkdirAll(string, os.FileMode) error
 
 		// newLogger creates a logger that the host can use to log messages and
 		// write critical statements.
@@ -91,12 +86,6 @@ func (productionDependencies) listen(s1, s2 string) (net.Listener, error) {
 // loadFile allows the host to load a persistence structure form disk.
 func (productionDependencies) loadFile(m persist.Metadata, i interface{}, s string) error {
 	return persist.LoadJSON(m, i, s)
-}
-
-// mkdirAll gives the host the ability to create chains of folders within the
-// filesystem.
-func (productionDependencies) mkdirAll(s string, fm os.FileMode) error {
-	return os.MkdirAll(s, fm)
 }
 
 // newLogger creates a logger that the host can use to log messages and write
--- a/modules/host/host_errors_test.go
+++ b/modules/host/host_errors_test.go
@@ -2,7 +2,6 @@ package host
 
 import (
 	"net"
-	"os"
 	"path/filepath"
 	"strings"
 	"testing"
@@ -10,46 +9,6 @@ import (
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/persist"
 )
-
-// dependencyErrMkdirAll is a dependency set that returns an error when MkdirAll
-// is called.
-type dependencyErrMkdirAll struct {
-	productionDependencies
-}
-
-func (dependencyErrMkdirAll) mkdirAll(string, os.FileMode) error {
-	return mockErrMkdirAll
-}
-
-// TestHostFailedMkdirAll initializes the host using a call to MkdirAll that
-// will fail.
-func TestHostFailedMkdirAll(t *testing.T) {
-	if testing.Short() {
-		t.SkipNow()
-	}
-	t.Parallel()
-	ht, err := blankHostTester("TestHostFailedMkdirAll")
-	if err != nil {
-		t.Fatal(err)
-	}
-	defer ht.Close()
-
-	err = ht.host.Close()
-	if err != nil {
-		t.Fatal(err)
-	}
-	ht.host, err = newHost(dependencyErrMkdirAll{}, ht.cs, ht.tpool, ht.wallet, "localhost:0", filepath.Join(ht.persistDir, modules.HostDir))
-	if err != mockErrMkdirAll {
-		t.Fatal(err)
-	}
-	// Set ht.host to something non-nil - nil was returned because startup was
-	// incomplete. If ht.host is nil at the end of the function, the ht.Close()
-	// operation will fail.
-	ht.host, err = newHost(productionDependencies{}, ht.cs, ht.tpool, ht.wallet, "localhost:0", filepath.Join(ht.persistDir, modules.HostDir))
-	if err != nil {
-		t.Fatal(err)
-	}
-}
 
 // dependencyErrNewLogger is a dependency set that returns an error when
 // NewLogger is called.


lukechampine:
Hi, sorry it took a while for us to get to this.
I think that rather than patching `NewFileLogger` (which is kind of sneaky/implicit), we should patch the calls to `NewFileLogger` to supply a different path. The current code for each module looks like:
```go
func New(..., persistDir string) (*Module, error) {
    logPath := filepath.Join(persistDir, "module.log")
    log, err := persist.NewFileLogger(logPath)
}
```
If we have a global `LogDir` variable, we could instead do:
```go
func New(..., persistDir string) (*Module, error) {
    var logPath string
    if persist.LogDir != "" {
        logPath := filepath.Join(persist.LogDir, "module.log")    
    } else {
        logPath := filepath.Join(persistDir, "module.log")
    }
    log, err := persist.NewFileLogger(logPath)
}
```
This requires more code, but it's more explicit and less likely to break in surprising ways. Let me know what you think.

Eriner:
@lukechampine 

> This requires more code, but it's more explicit and less likely to break in surprising ways. Let me know what you think.

I can make that change if that's what we want.

That rationale (having to change the callers vs the called function) was the reason I didn't do it that way in the first place. I figured that since we were setting a global flag to determine the path, the change should be applied globally (as in the `NewFileLogger` func) and the callers should otherwise operate independently. I'd even take it a step further and suggest that the `os.MkdirAll` calls to create the directory _also_ be handled by the `NewFileLogger` func instead of relying on the caller to implement that, as the code will be the same for literally every implementation. 

Making that change just means adding code duplication, and although I guess it makes it more explicit in the view of the caller, it's just about shuffling the code around and duplicating it everywhere a `NewFileLogger` call is made. Because it will have to happen on literally every call to that function, it makes more sense to me to make that a part of the function itself, which is what I've done. Ultimately up to you, and I'll make any change that's necessary. 

I would actually go the other way and advocate passing the filename `"module.log"` and `persistDir` strings back to `NewFileLogger` and let it handle constructing the filepath, rather than doing it inline in the caller. This would mean you have universal code applied to all modules (less code duplication), and if there is a bug, it would be consistent and you'd only have to fix it in one place. And based on the implementations I've seen, the caller funcs are returning upon `err != nil` for both `os.MkdirAll` and `NewFileLogger`, so nothing would change in that regard if the error is returned and checked by a single call to `NewFileLogger`.

ex:
```go
func (r *Renter) initPersist() error {
	// Create the perist directory if it does not yet exist.
	err := os.MkdirAll(r.persistDir, 0700)
	if err != nil {
		return err
	}

	// Initialize the logger.
	r.log, err = persist.NewFileLogger(filepath.Join(r.persistDir, logFile))
	if err != nil {
		return err
	}

	// Load the prior persistence structures.
	err = r.load()
	if err != nil && !os.IsNotExist(err) {
		return err
	}
	return nil
}

```

This could be simplified to:
```go
func (r *Renter) initPersist() error {
	// Initialize the logger.
	r.log, err = persist.NewFileLogger(r.persistDir, logFile)
	if err != nil {
		return err
	}

	// Load the prior persistence structures.
	err = r.load()
	if err != nil && !os.IsNotExist(err) {
		return err
	}
	return nil
}
```

In this case, `NewFileLogger` will handle global/non-global path manipulation, create the dir, and return the logger.

> and less likely to break in surprising ways

I don't really know how this is possible, and I think forgetting to add this check when calling `NewFileLogger` is more likely, which will result in the "global log dir" functionality not being applied for that call. Same goes for `os.MkdirAll` calls. Ultimately the code has to go somewhere, but unless there is a demonstrable bug in the implementation, I don't think shuffling it around and increasing code duplication for the purpose of preventing a theoretical "break in surprising ways" is a sufficient justification.

lukechampine:
hmm. I don't like implicitly rewriting paths, but I don't like tons of duplicated code either. I guess it's fine as written, as long as we can guarantee that there won't be collisions, i.e. two modules trying to create log files with the same name.

I'll add some minor comments and then we should be good to go.

siad/daemon.go
```diff
@@ -316,6 +317,14 @@ func startDaemon(config Config) (err error) {
 
 // startDaemonCmd is a passthrough function for startDaemon.
 func startDaemonCmd(cmd *cobra.Command, _ []string) {
+
+	// Set the global log directory, if provided. This must be
+	// done before profile.StartContinuousProfile is called, else
+	// there is a race on access.
+	if globalConfig.Siad.LogDir != "" {
+		persist.LogDir = globalConfig.Siad.LogDir
```
lukechampine: create `LogDir` here:
```go
err := os.MkdirAll(persist.LogDir, 0700)
if err != nil {
    die(err)
}
```

persist/log.go
```diff
@@ -119,6 +125,22 @@ func (cf *closeableFile) Write(b []byte) (int, error) {
...
+	//     - this seems to exist already for profile directories and the like,
+	//       is there file utility code in this repo I can rely on?
+	//   * once the above is done, don't overrwrite logFilename if these conditions
+	//     fail, and instead just write to the standard log path.
+	//     - it would be ideal if we could return an error here, noting the fallback,
+	//       however none of the callers of this function are prepared for that type
+	//       of error, meaning we need to decide to either modify the callers, or just
+	//       call an fmt.Println()
+	if LogDir != "" {
+		logFilename = LogDir + "/" + filepath.Base(logFilename)
```
lukechampine:
```go
filepath.Join(LogDir, filepath.Base(logFilename))
```

persist/log.go
```diff
@@ -119,6 +125,22 @@ func (cf *closeableFile) Write(b []byte) (int, error) {
...
+	//   * validate input to create path
+	//   * check if directory exists, else attempt to create
+	//     - this seems to exist already for profile directories and the like,
+	//       is there file utility code in this repo I can rely on?
+	//   * once the above is done, don't overrwrite logFilename if these conditions
+	//     fail, and instead just write to the standard log path.
+	//     - it would be ideal if we could return an error here, noting the fallback,
+	//       however none of the callers of this function are prepared for that type
+	//       of error, meaning we need to decide to either modify the callers, or just
+	//       call an fmt.Println()
```
lukechampine: `LogDir` should be created in the startup code (see other comment), so you don't need to create it in `NewFileLogger` or implement any fallback code. We'd definitely prefer to fail with an error here instead of writing to an unintended path.

Eriner:
@lukechampine comments look good. Thoughts on moving the `filepath.Join`'s/path building from the calling funcs to `NewFileLogger`?

DavidVorick:
Hi, I'm strongly in favor of leaving the filepath at the callsite. A lot of projects get in to trouble when they start adding global variables and hidden assumptions to their function calls. In the code you posted above, which made the `os.MkdirAll` implicit, that I consider to be a bad change. Yes, we have a lot of `MkdirAll` calls in our codebase, but that's because we are making a lot of folders, each which serve a very different purpose.

One thing I like about Go is that it works hard to encourage you to be explicit with everything you do. You check every error, which seems like code repetition, but really, it highlights very clearly every place that an error is possible, which is a godsend when you are working with unfamiliar code or hunting for a bug.

Global variables should be avoided as much as possible. We make a few exceptions with our build flags, but even then we are moving away from this as well.

lukechampine:
I'm not sure we can avoid a global variable at this point, unless we add a `logDir` or `logger` argument to each module constructor. I'm not opposed to that, but it would be a big change.

Eriner:
> One thing I like about Go is that it works hard to encourage you to be explicit with everything you do. You check every error, which seems like code repetition, but really, it highlights very clearly every place that an error is possible, which is a godsend when you are working with unfamiliar code or hunting for a bug.

This doesn't address anything being "wrong" with implicit directory creation by passing the path to the function? `if err != nil { return err }` still applies, so error handling isn't changed in any way.

> I'm not sure we can avoid a global variable at this point, unless we add a `logDir` or `logger` argument to each module constructor. I'm not opposed to that, but it would be a big change.

Again back to my initial comment, this was indeed the reason I added a global; without it, we'd have to change every call to the logger func.

lukechampine:
I don't think @DavidVorick was referring to any code in this PR, just making a point about Go's explicitness in general.

I guess the main reason for pushback is that when you pass a filepath to a function, you expect it to use that path, not some other path. It's ok to break a convention like that sometimes, but you need a really good reason.
As I said earlier, I think I'd be ok with changing the path as long as we're careful about it. In particular, that behavior would definitely need to be documented in the `NewFileLogger` docstring.

Eriner:
> As I said earlier, I think I'd be ok with changing the path as long as we're careful about it. In particular, that behavior would definitely need to be documented in the NewFileLogger docstring.

Of course, the docstring would need to be updated to reflect the change. 

> I guess the main reason for pushback is that when you pass a filepath to a function

I guess my comment here is if the change I proposed above (moving dir creation and management into `NewFileLogger` instead of duplicating that code inline in each caller) is acceptable, we're no longer passing a "filepath", but rather a logfile name **and** a _default_ filepath. Docstring and variable names in `NewFileLogger` could/should be changed to reflect this change as well.

Eriner:
## Update

I've made the proposed changes to `NewFileLogger` in the latest set of commits, as well as minor persist log test changes.
I've also removed all calls to `os.Mkdir`, which is now handled by `NewFileLogger`. I also removed the mkdirAll dependency interface func,
as it wasn't used anywhere after these changes.
Additionally, I've refactored all uses of `NewFileLogger` to match the new format.

Tests are passing, including `test-long`.

Against master at this PR's branch point, +68/-165, net -98 LOC. Unsubstantial decrease to siad (dev) binary size (~%0.0657). No change to siac (dev) binary size.

## Additional notes

### changes to persist/log_test.go

The `TestLogger` func had to be changed to construct the file path in the same way that `NewFileLogger` now does,
as it attempts to `ioutil.ReadFile`.
  
### "initPersist" implementations differed in how they returned errors from `os.MkdirAll`

Previously, errors returned by `os.MkdirAll` were handled differently based on the context of where it was called. Now, the same error `os.MkdirAll` would have returned is instead returned by `persist.NewFileLogger`. As long as the desired method for handling that error is the same, this should not impact how the error is handled.

Examples of (previous) variations in `os.MkdirAll` error handling:

```go
func (cs *ConsensusSet) initPersist() error {
       // Create the consensus directory.
       err := os.MkdirAll(cs.persistDir, 0700)
       if err != nil {
               return err
       }
```

```go
// Create the perist directory if it does not yet exist.
err = dependencies.mkdirAll(cm.persistDir, 0700)
if err != nil {
        return nil, build.ExtendErr("error while creating the persist directory for the contract manager", err)
}
```

```go
// Create the folder for all of the profiling results.
err := os.MkdirAll(dir, 0700)
if err != nil {
        fmt.Println(err)
        return
}
```

These variations in how `os.MkdirAll` errors are handled houldn't matter; these same variations can/should handled by the err check post ca

persist/log.go
Eriner: Check the current implementation now, as the `os.Mkdir` call should apply in one place for both the passed (default) directory and the `LogDir`.

lukechampine:
I don't think `NewFileLogger` should be creating the folder. It's not obvious that the responsibility for creating the folder lies with `NewFileLogger`. For comparison, `os.Create` doesn't create the folder where the file will be created; neither does `ioutil.WriteFile`.

My own preference is that these directories should all be created by the function that creates the modules. e.g.:
```go
consensusDir := filepath.Join(config.Siad.SiaDir, "consensus")
err := os.MkdirAll(consensusDir, 0700)
consensusLogger, err := persist.NewFileLogger(config.Siad.LogDir, "consensus.log")
cs, err := cs, err = consensus.New(consensusDir, consensusLogger)
```
Regardless, we should all be in agreement before any more code is written.

Eriner:
Right, but that's the same way that it was done before the latest three commits, just moving that same redundant code into a different function. One of the benefits of the change is the reduction in code redundancy. You could have `os.MkdirAll` calls in front of every call to `NewFileLogger`, but since that was/would be the case, why not place it inside `NewFileLogger`?

lukechampine:
There are lots of competing concerns when it comes to writing good code, and the DRY principle is just one of them. Specifically, in this case, DRY is at odds with making the function's behavior obvious and composable. Calling `Mkdir` inside `NewFileLogger` is [easy, but not simple](https://www.infoq.com/presentations/Simple-Made-Easy). It's a sort of inversion of responsibility; the function is setting up the environment that it _should_ have been called in in the first place.

Another way of looking at it: the log file isn't the only file that this applies to. The wallet database also needs its parent directory to exist, the gateway's `nodes.json` needs its parent directory to exist, etc. If we called `Mkdir` in every helper function that created a file, that would result in a fair amount of code duplication as well.

Besides, we can shift the responsibility around without abandoning DRY completely. For example, we could create the directories in `startDaemon` like so:
```go
moduleNames := []string{"gateway", "consensus", "wallet", ...}
for _, name := range moduleNames {
    // error handling omitted
    os.MkdirAll(filepath.Join(config.Siad.SiaDir, name))
}
```
That way, we're only calling `Mkdir` in one place.

Thanks for your patience on this. It feels a little silly to have so much back-and-forth over a relatively small change, but as I'm sure you know, sometimes programming is like that. ¯\\_(ツ)\_/¯

Eriner:
That's an argument I'm amenable to. I just don't think repeating `os.MkdirAll` everywhere prefacing the call to `NewFileLogger` makes sense, and any arguments where the solution is "just shift it over here and repeat it for every module" just seemed silly to me.

Eriner:
If that approach, or something similar, is what we end up doing, the commits thus far are still worthwhile and valid, FWIW. Just have to move the `MkdirAll` calls from `NewFileLogger` into somewhere else where it handles that (for all modules). This would also mean that there are likely additional calls to `os.MkdirAll` that could be removed, as per your example for the wallet db and the gateway.

Eriner:
If there are changes I can make to move this PR along, please let me know and I will re-open a new PR with these commits and the ammended changes, however until that happens I'm going to go ahead and close this PR and remove my repo (trying to do some personal spring cleaning).

Let me know how/if to proceed. Thanks.

PR #2086, opened 2017-07-06
add logic to switch away from bad hosts

DavidVorick:
I do not have tests yet for the code here, but this contains the fixes that will allow renters to automatically avoid hosts which no longer have attractive or competitive scores.

I am about to connect the host reliability metrics that @ChrisSchinnerl added in a recent pull request to the scoring system, which means that the contractor will also eventually move past any contracts that are consistently failing during uploads or downloads.


modules/renter/contractor/contractor_test.go
```diff
@@ -345,7 +351,10 @@ func TestIntegrationSetAllowance(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if len(c.contracts) != 2 {
+	c.mu.Lock()
+	clen := len(c.contracts)
+	c.mu.Unlock()
+	if clen != 2 {
 		t.Fatal("expected 2 contracts, got", len(c.contracts))
```
johnathanhowell: shouldn't this be `clen`?

modules/renter.go
```diff
@@ -224,6 +224,11 @@ type RenterContract struct {
 	ContractFee types.Currency `json:"contractfee"`
 	TxnFee      types.Currency `json:"txnfee"`
 	SiafundFee  types.Currency `json:"siafundfee"`
+
+	// GoodForUpload indicates whether the contract should be used to upload new
+	// data or not.
+	GoodForUpload bool
```
tlightsky: gooood

modules/renter/contractor/contracts.go
```diff
@@ -356,12 +482,14 @@ func (c *Contractor) threadedContractMaintenance() {
...
 	// contracts.
 	for _, host := range hosts {
-		// Attempt forming a contract with htis host.
+		// Attempt forming a contract with this host.
 		newContract, err := c.managedNewContract(host, numSectors, endHeight)
 		if err != nil {
 			c.log.Printf("Attempted to form a contract with %v, but if failed: %v\n", host.NetAddress, err)
 			continue
 		}
+		// newContract.GoodForUpload = true
```
johnathanhowell: should this be commented out??

modules/renter/contractor/contracts.go
```diff
@@ -83,6 +83,104 @@ func (c *Contractor) contractEndHeight() types.BlockHeight {
...
+	c.mu.RLock()
+	hostCount := int(c.allowance.Hosts)
+	c.mu.RUnlock()
+	hosts := c.hdb.RandomHosts(hostCount+minScoreHostBuffer, nil)
+
+	// Find the lowest score of this batch of hosts.
+	if len(hosts) <= 0 {
+		return
+	}
+	lowestScore := c.hdb.ScoreBreakdown(hosts[0]).Score
```
johnathanhowell: This nondeterministically causes an `index out of range` panic on my machine

modules/renter/contractor/contracts.go
```diff
@@ -83,6 +84,104 @@ func (c *Contractor) contractEndHeight() types.BlockHeight {
...
+	c.mu.RLock()
+	hostCount := int(c.allowance.Hosts)
+	c.mu.RUnlock()
+	if hostCount <= 0 {
+		return
+	}
+	hosts := c.hdb.RandomHosts(hostCount+minScoreHostBuffer, nil)
+
+	// Find the lowest score of this batch of hosts.
+	lowestScore := c.hdb.ScoreBreakdown(hosts[0]).Score
```
johnathanhowell: This nondeterministically causes a `index out of range` panic on my machine. I think we need a `if len(hosts)` branch before this line

DavidVorick:
T.T I just want the build to pass T.T

modules/renter/contractor/contracts.go
```diff
@@ -83,6 +84,119 @@ func (c *Contractor) contractEndHeight() types.BlockHeight {
...
+	}
+	hosts := c.hdb.RandomHosts(hostCount+minScoreHostBuffer, nil)
+	if len(hosts) <= 0 {
+		return
+	}
+
+	// Find the lowest score of this batch of hosts.
+	lowestScore := c.hdb.ScoreBreakdown(hosts[0]).Score
+	for i := 1; i < len(hosts); i++ {
+		score := c.hdb.ScoreBreakdown(hosts[i]).Score
```
lukechampine: `for _, h := range hosts[1:]` also works

modules/renter/contractor/contracts.go
```diff
@@ -83,6 +84,119 @@ func (c *Contractor) contractEndHeight() types.BlockHeight {
...
+		// Contract has no utility if the score is poor.
+		if c.hdb.ScoreBreakdown(host).Score.Cmp(minScore) < 0 {
+			contracts[i].GoodForUpload = false
+			contracts[i].GoodForRenew = false
+			continue
+		}
+		// Contract has no utility if the host is offline.
+		c.mu.Lock()
+		offline := c.isOffline(contracts[i].ID)
+		c.mu.Unlock()
```
lukechampine: why does this require a write-lock?

DavidVorick: convention mandates that any call to an unexported function elsewhere in the module requires a writelock.

modules/renter/contractor/contracts.go
DavidVorick: less memory efficient.

modules/renter/contractor/contracts.go
lukechampine: why?

api/renter.go
```diff
@@ -278,6 +284,9 @@ func (api *API) renterContractsHandler(w http.ResponseWriter, _ *http.Request, _
 			StorageSpending:  c.StorageSpending,
 			TotalCost:        c.TotalCost,
 			UploadSpending:   c.UploadSpending,
+
+			GoodForRenew:  c.GoodForRenew,
+			GoodForUpload: c.GoodForUpload,
```
DavidVorick: ah whoops, this was for testing

DavidVorick:
I feel like I can't even be mad at this point. 25 failed commits in a row.

A lot of the struggle was because the existing code simply wasn't handling the renew process correctly. Doubling up on hosts, failing to repair dead contracts, etc etc. Basically I wrote a test that dug a lot deeper than any other test we have, and it exposed a lot of issues.

DavidVorick:
Download code is still loose, but this is ready to merge

PR #2092, opened 2017-07-07
use new contractor fields in renter

johnathanhowell:


modules/renter/files.go
```diff
@@ -204,15 +204,23 @@ func (r *Renter) FileList() []modules.FileInfo {
 	lockID := r.mu.RLock()
 	defer r.mu.RUnlock(lockID)
 
+	isOffline := func(id types.FileContractID) bool {
+		offline := r.hostContractor.IsOffline(id)
+		contract, exists := r.hostContractor.ContractByID(id)
+		if !exists {
+			return offline
```
DavidVorick: should probably be `return true`

modules/renter/files.go
```diff
@@ -204,15 +204,23 @@ func (r *Renter) FileList() []modules.FileInfo {
 	lockID := r.mu.RLock()
 	defer r.mu.RUnlock(lockID)
 
+	isOffline := func(id types.FileContractID) bool {
+		offline := r.hostContractor.IsOffline(id)
+		contract, exists := r.hostContractor.ContractByID(id)
```
DavidVorick: should not be calling exported methods while holding a lock , you've got a readlock on the renter right now and you're making exported calls to the contractor

modules/renter/contractor/contractor.go
```diff
@@ -84,6 +84,12 @@ func (c *Contractor) Contract(hostAddr modules.NetAddress) (modules.RenterContra
 	return modules.RenterContract{}, false
 }
 
+// ContractByID returns the contract with the id specified, if it exists.
+func (c *Contractor) ContractByID(id types.FileContractID) (modules.RenterContract, bool) {
+	contract, exists := c.contracts[id]
+	return contract, exists
```
DavidVorick: shouldn't this be causing race conditions? It's an exported function with no threading safety

johnathanhowell: Yeah, needs a mutex lol this PR was put together pretty hastily

johnathanhowell:
Updated this PR to fix the concurrency patterns. `make test-long` is failing every test that involves uploading, because the branch under `if GoodForUpload {` in the renter is never being hit (contracts are never being marked GoodForUpload in testing)

johnathanhowell:
Rebased on to the latest `host-score-check` changes and tests are passing :)

modules/renter/files.go
```diff
@@ -201,25 +201,34 @@ func (r *Renter) DeleteFile(nickname string) error {
...
-
-	files := make([]modules.FileInfo, 0, len(r.files))
 	for _, f := range r.files {
+		files = append(files, f)
+	}
+	r.mu.RUnlock(lockID)
+
+	isOffline := func(id types.FileContractID) bool {
+		return r.hostContractor.IsOffline(id)
+
```
DavidVorick: extra newline?

modules/renter/files.go
```diff
@@ -201,25 +201,34 @@ func (r *Renter) DeleteFile(nickname string) error {
...
 			Filesize:       f.size,
-			Available:      f.available(r.hostContractor.IsOffline),
-			Redundancy:     f.redundancy(r.hostContractor.IsOffline),
 			Renewing:       renewing,
+			Available:      f.available(isOffline),
+			Redundancy:     f.redundancy(isOffline),
 			UploadProgress: f.uploadProgress(),
 			Expiration:     f.expiration(),
 		})
 		f.mu.RUnlock()
```
DavidVorick: This means that the hostContractor and the file can't be locked at the same time from other functions. Are sure that's the right approach?

I'm still not super happy with the way this is constructed.

PR #2096, opened 2017-07-08
Add new RPC for revising contracts

ChrisSchinnerl:
This PR is work in progress.

Once finished a new RPC should be added that handles all operations that require a new contract revision. (Download, Insert, Modify and Delete)
This also means that the renter's downloader will be merged into the editor and the host's negotiatedownload will be merged into negotiaterevisecontract.


lukechampine:
Wow, I did not expect to see this so soon!

There's one important tweak that I think we should make: the renter should send the signed revision before it sends the revision request, and the `Data` field should be sent last. This allows the host to check that all the metadata checks out before it starts reading the 4MB sector. As a result, it can reject revisions early if there's a problem, so you don't need to wait for a whole sector to transfer (consuming time and bandwidth) before rejecting the revision.

ChrisSchinnerl:
Well, it still needs some work. The renter part is not done yet but I wanted to keep you updated  to get some feedback before I start the renter part tomorrow.
I will separate the request in 2 structs and send them separately.

ChrisSchinnerl:
@lukechampine I changed the protocol in the last commit. Is that what you had in mind? The code is still a bit messy and needs cleanup but the most of the work for that PR should be done.

lukechampine:
Closing this for now so that we can focus, but let's keep the branch around (or you can just keep it locally) so that we can revisit it later.

PR #2101, opened 2017-07-09
repair broken consensus change distribution

DavidVorick:
By combining consensus changes before distributing them when downloading multiple blocks at a time, we ended up providing subscribers with consensus change IDs that did not match the original consensus change ID. The result was a lot of the modules started resetting themselves.

This fixes that, I should have thought about it on the first pass, and I remember being skeptical that it was the right way to go.

Code has been simplified, and now it should be correct.

We need to replace RC3 with this code ASAP, it's a pretty massive usability fix.

### DIFF
--- a/modules/consensus/accept_test.go
+++ b/modules/consensus/accept_test.go
@@ -9,6 +9,8 @@ import (
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/persist"
 	"github.com/NebulousLabs/Sia/types"
+
+	"github.com/NebulousLabs/bolt"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


DavidVorick:
Actually, let's keep this open until I have a regression test for it.

DavidVorick:
Should be good to go, including regtests.

modules/consensus/accept_test.go
```diff
@@ -1145,7 +1148,22 @@ func TestChainedAcceptBlock(t *testing.T) {
...
-	if bcs.changes != 3 || bcs.appliedBlocks != int(cst2.cs.Height()+1) || bcs.revertedBlocks != 0 {
+	if bcs.appliedBlocks != int(cst2.cs.Height()+1) || bcs.revertedBlocks != 0 {
 		t.Error("consensus changes do not seem to be getting passed to subscribers correctly")
 	}
+
+	// Check that every change recorded in 'bcs' is also available in the
+	// consensus set.
+	for _, change := range bcs.changes {
+		err := cst2.cs.db.Update(func(tx *bolt.Tx) error {
+			_, exists := getEntry(tx, change)
```
lukechampine: shouldn't this be a `View`? And shouldn't the `for` loop be inside it?

lukechampine: oh, it's in a test so not a big deal

DavidVorick: ah yeah, little lazy about the performance here. Maybe adds 1 second to the test, shouldn't be too bad.

PR #2102, opened 2017-07-10
Add SPV consensus

nitronick600:
Overall, this is conceptually what I think should be done.  There are some improvements to be made, though I'm not unsure about added complexity. This implementation forces full nodes to have to maintain a block consensus and a header consensus.  The "blockMap" and "headerMap" would ideally be a merged DB but there are back-compatibility complications along with added implementation complexity.

### DIFF
--- a/modules/consensus/accept.go
+++ b/modules/consensus/accept.go
@@ -18,20 +18,22 @@ var (
 	errInconsistentSet = errors.New("consensus set is not in a consistent state")
 	errOrphan          = errors.New("block has no known parent")
 	errNonLinearChain  = errors.New("block set is not a contiguous chain")
+	errNoHdrMap        = errors.New("header map is not in database")
 )
 
 // managedBroadcastBlock will broadcast a block to the consensus set's peers.
-func (cs *ConsensusSet) managedBroadcastBlock(b types.Block) {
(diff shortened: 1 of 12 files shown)


nitronick600:
I will make changes to the Docs as well if we decide this is the correct approach.

modules/consensus/accept.go
```diff
@@ -177,6 +117,45 @@ func (cs *ConsensusSet) addBlockToTree(tx *bolt.Tx, b types.Block, parent *proce
 	return ce, nil
 }
 
+// addBlockToTree inserts a block into the blockNode tree by adding it to its
```
nitronick600: will revert this

DavidVorick:
closing this, I don't feel it's the right approach. I think if we want to pursue SPV, the right route to take is trusted explorer nodes that your wallet talks to and works with. Those trusted nodes can be some trusted third party service, or those trusted nodes can be your desktop computer at home, or your own server instance in the cloud.

Starting from that point, we can build out lite clients to be decreasingly trustful, but still always dependent on some trusted central source for information like transactions coming from blocks.

I want to use this trusted model because SPV boils down to the same trust model, but it's less apparent and a lot of people think that spv nodes count as full, trustless blockchain nodes, something that I think is dangerous for cryptocurrencies.

PR #2104, opened 2017-07-10
replace rawJSON type with standard json.RawMessage

aiorla:
Resolves TODO that was waiting for two Go 1.8 fixes. [Link](https://github.com/golang/go/blob/master/src/encoding/json/stream.go#L244) (last 2 commits)


lukechampine:
thanks!

PR #2105, opened 2017-07-10
add reasonable timeouts to the API server

johnathanhowell:
Previously, `siad`'s API server was missing [timeouts](https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/) for its HTTP connections. This could lead to it leaking conns if clients are slow, disappearing, or unreliable and is a plausible cause for the `too many open files` errors users have been reporting.

### DIFF
--- a/siad/server.go
+++ b/siad/server.go
@@ -16,6 +16,7 @@ import (
 	"runtime"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/NebulousLabs/Sia/api"
 	"github.com/NebulousLabs/Sia/build"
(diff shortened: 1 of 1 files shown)


siad/server.go
```diff
@@ -382,6 +383,23 @@ func NewServer(bindAddr, requiredUserAgent, requiredPassword string) (*Server, e
...
 			Handler: mux,
+
+			// set reasonable timeout windows for requests, to prevent the Sia API
+			// server from leaking file descriptors due to slow, disappearing, or
+			// unreliable API clients.
+
+			// ReadTimeout defines the maximum amount of time allowed to fully read
+			// the request body. This timeout is applied to every handler in the
+			// server.
+			ReadTimeout: time.Minute,
```
lukechampine: might this be a problem if we're doing something like downloading a file over HTTP? Can the timeout be overridden on a per-request basis?

I think what we probably want most is a strict `IdleTimeout`; that should clear up any hung requests.

johnathanhowell: Downloading a file over http shouldn't be a problem; the request body will have already been read, it's the `WriteTimeout` that would be an issue in that case, which is why it is not included in this PR.

lukechampine: gotcha. I'd still like to see a shorter `IdleTimeout` though, probably <30 seconds

johnathanhowell: ACK, I decreased the `IdleTimeout` to 20 seconds.

DavidVorick: Would like to see all of these be friendly to tunneling over Tor, though it's an uncommon use case, I see no danger in supporting it. That means 2m-5m for each value, please update

PR #2108, opened 2017-07-11
File sharing feature

tlightsky:
an incorrect rebase close the pr before: https://github.com/NebulousLabs/Sia/pull/1978

now i rebase all the code to newest master and make a pr again

some problem still unsolved though:

* how many contracts should be formed(if a file is already avaiable,should we make more to download it more faster?)
* how long the contract should be(same with allowance? what to do if no allowance, default a week?)
* should we renew the `download only contract`
* should we use the money directly from wallet(without confirm from user?)


lukechampine:
Closing this because the renter architecture is being overhauled over the coming weeks. We appreciate your efforts, but filesharing needs to be incorporated into the new architecture from the ground up if it's going to meet our standards.

PR #2119, opened 2017-07-12
Remove duplicated comment and reformat with two references

petabytestorage:
this is ferris/rudi enjoying trying to fix docs :slightly_smiling_face:

### DIFF
--- a/modules/consensus/synchronize.go
+++ b/modules/consensus/synchronize.go
@@ -417,7 +417,7 @@ func (cs *ConsensusSet) threadedRPCRelayHeader(conn modules.PeerConn) error {
 		return cs.validateHeader(boltTxWrapper{tx}, h)
 	})
 	cs.mu.RUnlock()
-	// WARNING Orphan Multithreading Logic (dangerous area)
+	// WARN: orphan multithreading logic (dangerous areas, see below)
 	//
 	// If the header is valid and extends the heaviest chain, fetch the
 	// corresponding block. Call needs to be made in a separate goroutine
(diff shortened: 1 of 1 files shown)


modules/consensus/synchronize.go
```diff
@@ -417,17 +417,19 @@ func (cs *ConsensusSet) threadedRPCRelayHeader(conn modules.PeerConn) error {
...
-		// be made in a separate goroutine as execution requires calling an
-		// exported gateway method - threadedRPCRelayHeader was likely called
-		// from an exported gateway function.
-		//
-		// NOTE: In general this is bad design. Rather than recycling other
-		// calls, the whole protocol should have been kept in a single RPC.
-		// Because it is not, we have to do weird threading to prevent
-		// deadlocks, and we also have to be concerned every time the code in
-		// managedReceiveBlocks is adjusted.
+	// WARNING Orphan Multithreading Logic (dangerous area)
```
DavidVorick: needs a colon: `WARNING:`

I think though we typically use `WARN:` in the codebase, so it should probably just be the 4 letters and a colon.

modules/consensus/synchronize.go
```diff
@@ -441,16 +443,7 @@ func (cs *ConsensusSet) threadedRPCRelayHeader(conn modules.PeerConn) error {
...
-	// corresponding block. Call needs to be made in a separate goroutine
-	// because an exported call to the gateway is used, which is a deadlock
-	// risk given that rpcRelayHeader is called from the gateway.
-	//
-	// NOTE: In general this is bad design. Rather than recycling other calls,
-	// the whole protocol should have been kept in a single RPC. Because it is
-	// not, we have to do weird threading to prevent deadlocks, and we also
-	// have to be concerned every time the code in managedReceiveBlock is
-	// adjusted.
+	// warning orphan multithreading logic case #2
```
DavidVorick: Should probably also be `WARN: orphan multi...`

petabytestorage:
thanks @DavidVorick all fixed now i believe

PR #2125, opened 2017-07-12
Gateway Upgrades

DavidVorick:
This is a merge of the gateway staging branch, which addresses the points of the "Gateway Upgrades" project. The goals of the project were:

- [x]   Version exchange should provide a genesis block + self-identifying nonce

-  [x]  Gateway should use self-identifying nonce to avoid connecting to itself

-  [x]  Gateway needs to be using ~~yamux~~ smux instead of muxado

- [ ]  More research should be done on what we can do to prevent eclispe attacks, DoS attacks, and other network-based attacks.

As of this PR, the gateway uses smux, with muxado still employed for communicating with older gateways. Muxado will eventually be phased out, but this cannot happen until we drop support for LTS v1.0.4. Unlike muxado, smux has been vendored into the repo.

Gateways v1.3.0 and above will use a new handshake that includes their genesis block ID and a unique ID. This handshake requires fewer roundtrips than the previous protocol, and allows each gateway to send an error message to the other instead of hacks like silently closing the connection. For example, when a gateway wants to ping another gateway, it will send a `modules.StopResponse` error, which is handled as a special case by the recipient. (This mimics the use of `StopResponse` in the renter-host protocol to initiate graceful termination.)

Lastly, the gateway now stores more information persistently about each node, including whether the node has ever been connected to as an outbound peer. This information can be used to improve the gateway's resilience to certain attacks.


modules/consensus/synchronize.go
```diff
@@ -143,12 +144,7 @@ func (cs *ConsensusSet) managedReceiveBlocks(conn modules.PeerConn) (returnErr e
 	// needs to be chosen.
 	stalled := true
 	defer func() {
-		// TODO: Timeout errors returned by muxado do not conform to the net.Error
-		// interface and therefore we cannot check if the error is a timeout using
-		// the Timeout() method. Once muxado issue #14 is resolved change the below
-		// condition to:
-		//     if netErr, ok := returnErr.(net.Error); ok && netErr.Timeout() && stalled { ... }
-		if stalled && returnErr != nil && (returnErr.Error() == "Read timeout" || returnErr.Error() == "Write timeout") {
+		if netErr, ok := returnErr.(net.Error); ok && netErr.Timeout() && stalled {
```
lukechampine: needs backwards compatibility with muxado

modules/gateway/peers.go
```diff
@@ -142,40 +153,96 @@ func (g *Gateway) threadedAcceptConn(conn net.Conn) {
...
+	ourHeader := sessionHeader{
+		GenesisID:  types.GenesisID,
+		UniqueID:   g.id,
+		NetAddress: modules.NetAddress(net.JoinHostPort(host, g.port)),
+	}
+	remoteHeader, err := readSessionHeader(conn, ourHeader)
+	if err != nil {
+		return err
+	}
+	if err := writeSessionHeader(conn, ourHeader); err != nil {
```
lukechampine: names aren't super clear -- `writeSessionHeader` also reads error. Maybe `exchangeRemoteSessionHeader`

modules/gateway/peers.go
```diff
@@ -347,66 +400,71 @@ func acceptConnVersionHandshake(conn net.Conn, version string) (remoteVersion st
...
+		return fmt.Errorf("failed to read header acceptance: %v", err)
+	} else if response == modules.StopResponse {
+		return errors.New("peer did not want a connection")
+	} else if response != modules.AcceptResponse {
+		return fmt.Errorf("peer rejected our header: %v", response)
+	}
+	return nil
+}
+
+func readSessionHeader(conn net.Conn, ourHeader sessionHeader) (sessionHeader, error) {
```
lukechampine: add comments when renaming

modules/gateway/persist.go
```diff
@@ -20,29 +20,27 @@ const (
...
 }
 
 // load loads the Gateway's persistent data from disk.
 func (g *Gateway) load() error {
-	var nodes []modules.NetAddress
+	var nodes []*node
 	err := persist.LoadJSON(persistMetadata, &nodes, filepath.Join(g.persistDir, nodesFile))
 	if err != nil {
-		return err
+		// COMPATv1.2.1
```
lukechampine: should be 1.3.0

modules/gateway/rpc.go
```diff
@@ -45,6 +45,15 @@ func (g *Gateway) managedRPC(addr modules.NetAddress, name string, fn modules.RP
...
 	if err != nil {
+		// peer probably disconnected without sending a shutdown signal;
+		// disconnect from them
+		g.log.Debugf("Could not initiate RPC with %v; disconnecting", addr)
+		g.mu.Lock()
+		delete(g.peers, addr)
+		g.mu.Unlock()
+		if derr := peer.sess.Close(); err != nil {
+			g.log.Debugf("Could not cleanly disconnect from %v: %v", addr, derr)
+		}
```
lukechampine: maybe deduplicate via `g.managedDisconnect`

lukechampine: close before deleting from `g.peers`

PR #2126, opened 2017-07-13
more performant hostdb locking in interaction methods

johnathanhowell:


DavidVorick:
Closing this for now, I do not think it's the bottleneck anywhere. I'm also not confident what the consistency model is between the hostdb and the hosttree, I thought you were supposed to have the hostdb locked when manipulating the hosttree. Maybe not, but since this is non-critical and also controvertial, closing for now

PR #2127, opened 2017-07-13
add cleanCloseHandler: return from http handler immediately if remote hangs up

johnathanhowell:
This PR adds a middleware which ensures that if an API caller hangs up the API request early, before the underlying handler finishes, the connection will be fully closed. Previously, the connection would stay half-open (`CLOSE_WAIT`) until the underlying handler finished, which could cause a buildup of useless open sockets.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -118,6 +118,23 @@ func RequirePassword(h httprouter.Handle, password string) httprouter.Handle {
 	}
 }
 
+// cleanCloseHandler wraps the entire API, ensuring that underlying conns are
+// not leaked if the rmeote end closes the connection before the underlying
+// handler finishes.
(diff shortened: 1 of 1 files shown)


api/api.go
```diff
@@ -118,6 +120,33 @@ func RequirePassword(h httprouter.Handle, password string) httprouter.Handle {
...
+		case <-done:
+		case <-r.Context().Done():
+		}
+
+		// Sanity check - thread should not take more than an hour to return.
+		select {
+		case <-done:
+		case <-time.After(time.Minute * 60):
+			debug.PrintStack()
+			build.Critical("api call is taking more than 20 minutes to return")
```
lukechampine: doesn't `build.Critical` already print the stack?

DavidVorick: oh yeah

PR #2141, opened 2017-07-14
efficiently remove transactions from miner unsolved block

marcinja:
Before:
```
=== RUN   TestFeeEstimation
--- PASS: TestFeeEstimation (45.80s)
=== RUN   TestTpoolScalability
--- PASS: TestTpoolScalability (54.17s)
=== RUN   TestHeapFees
--- PASS: TestHeapFees (59.54s)
```
After:
```
=== RUN   TestFeeEstimation
--- PASS: TestFeeEstimation (47.18s)
=== RUN   TestTpoolScalability
--- PASS: TestTpoolScalability (30.35s)
=== RUN   TestHeapFees
--- PASS: TestHeapFees (32.76s)
```

Edit: `TestBigTpool` is probably too big. It fills up the tpool with at least 3 blocks worth of transactions (~6MB+), takes an enormous amount of time (150+ seconds) and is a lot of lines of code. It might be a good idea to define some helper functions inside the test to make it easier to read.


marcinja:
Running `TestBigTpool` with different values for `numGraphsPerChunk` confirms that there's probably some O(n^2) code in the tpool. At `numGraphsPerChunk = 500` the test runs in ~40 sec and creates around 3 MB of transactions. `numGraphsPerChunk = 1000` gives ~150 sec and ~6MB txns. `numGraphsPerChunk = 1500` gives ~360 sec and ~9MB txns.

I did some profiling at `numGraphsPerChunk = 1500`, these are the things that stand out the most.

total-time
update-subs

Here are some snippets from the `(pprof) weblist` output under the `(*TransactionPool).updateSubscribersTransactions` method.

```
 Total:       8.33s   3.88mins (flat, cum) 57.60%
```

```
     15        1.45s     30.39s 	for id := range tp.subscriberSets {
     16            .          . 		// The transaction set is still in the transaction pool, no need to
     17            .          . 		// create an update.
     18        1.16s   1.45mins 		_, exists := tp.transactionSets[id]
```

```
     34        4.75s     31.20s 	for id, set := range tp.transactionSets {
     35        950ms   1.30mins 		_, exists := tp.subscriberSets[id]
```

### Solutions
In this test (and other transaction pool tests) we create testers that accept thousands of transaction sets in a short period of time. One way to make this more efficient would be to add a buffer in the transaction pool that stores accepted transaction sets and waits a short amount of time before updating its subscribers with all the sets stored in that buffer.


It'd be useful to find out what the normal number of calls to `(*TransactionPool).updateSubscribersTransactions` a `siad` process usually makes and whether or not such a change would make a difference there. If nodes frequently receive many transaction sets in a short amount of time from their peers then a change like this might help with efficiency.

modules/miner/update.go
```diff
@@ -193,12 +170,155 @@ func (m *Miner) deleteMapElementTxns(id splitSetID) {
...
 
 		// Promote sets from overflow heap to block if possible.
-		for overflowElem, canPromote := m.overflowMapHeap.peek(); canPromote && m.blockMapHeap.size+overflowElem.set.size < types.BlockSizeLimit-5e3; {
-			promotedElem := m.overflowMapHeap.pop()
-			m.blockMapHeap.push(promotedElem)
+		for overflowElem, canPromote := m.peekAtOverflow(); canPromote && m.blockMapHeap.size+overflowElem.set.size < types.BlockSizeLimit-5e3; {
+			promotedElem := m.popFromOverflow()
+			m.pushToBlock(promotedElem)
+		}
+		m.removeSplitSetFromUnsolvedBlock(id)
```
DavidVorick: Can you leave a comment explaining why you chose to remove the element from the block after you pushed elements into the block? Naively this seems backwards because you end up with an over-full block as a midstate, which could mean the block has to add more elements.

If there's a good reason to do it in this order, a comment is needed.

modules/miner/update.go
```diff
@@ -193,12 +170,155 @@ func (m *Miner) deleteMapElementTxns(id splitSetID) {
...
+		setsFixed[setID] = struct{}{}
+	}
+}
+
+// removeTxFromUnsolvedBlock removes the given transaction by either swapping it
+// with the transaction at the end of the slice or, if the transaction to be
+// removed is the last transaction in the block, just shrinking the slice. It
+// returns the transaction ID of the last element in the block prior to the
+// swap/removal taking place.
+func (m *Miner) removeTxFromUnsolvedBlock(id types.TransactionID) types.TransactionID {
```
DavidVorick: sort the functions alphabetically

modules/transactionpool/transactionpool_test.go
```diff
@@ -655,3 +655,660 @@ func TestHeapFees(t *testing.T) {
...
+			t.Error(err)
+		}
+		outputTxns1 = append(outputTxns1, txns)
+	}
+
+	counter := 1
+	var graphs [][]types.Transaction
+	for _, output := range outputTxns1 {
+		finalTxn := output[len(output)-1]
+		for i := 0; i < 500; i++ { //500 is the the number of outputs
```
DavidVorick: leave a space after `//` in the comments

modules/transactionpool/transactionpool_test.go
```diff
@@ -655,3 +655,660 @@ func TestHeapFees(t *testing.T) {
...
+			graph, err := types.TransactionGraph(finalTxn.SiacoinOutputID(uint64(i)), edges)
+			if err != nil {
+				t.Fatal(err)
+			}
+			graphs = append(graphs, graph)
+			counter++
+		}
+	}
+	////////////////////////////////////////////////////////////////////////////
+	//						Chunk 2
```
DavidVorick: just one space between `//` and `Chunk 2`

modules/transactionpool/transactionpool_test.go
```diff
@@ -655,3 +655,660 @@ func TestHeapFees(t *testing.T) {
...
+			graph, err := types.TransactionGraph(finalTxn.SiacoinOutputID(uint64(i)), edges)
+			if err != nil {
+				t.Fatal(err)
+			}
+			graphs2 = append(graphs2, graph)
+			counter++
+		}
+	}
+	////////////////////////////////////////////////////////////////////////////
+	//						Chunk 3
```
DavidVorick: same

modules/transactionpool/transactionpool_test.go
```diff
@@ -655,3 +655,660 @@ func TestHeapFees(t *testing.T) {
...
+			t.Error(err)
+		}
+		outputTxns3 = append(outputTxns3, txns)
+	}
+
+	counter = 1
+	var graphs3 [][]types.Transaction
+	for _, output := range outputTxns3 {
+		finalTxn := output[len(output)-1]
+		for i := 0; i < 500; i++ { //500 is the the number of outputs
```
DavidVorick: space

modules/transactionpool/transactionpool_test.go
```diff
@@ -655,3 +655,660 @@ func TestHeapFees(t *testing.T) {
...
+	finishThird := fastrand.Perm(numGraphsPerChunk / 2)
+	for _, i := range finishThird {
+		graph := graphs3[i+(numGraphsPerChunk/2)]
+		for _, txn := range graph[1:] {
+			err := tpt.tpool.AcceptTransactionSet([]types.Transaction{txn})
+			if err != nil {
+				t.Fatal(err)
+			}
+		}
+	}
```
DavidVorick: I'm thinking there's a better way to do this. Append all of the sets into a single large set, then do the first 250 as one `Perm`, do the second 500 as a second `Perm`, the third 500 as a third `Perm`, and the final 250 as a fourth `Perm`.

Since `Perm` I think only gives you 0-n, you'd have to do `perm+250` for the second set, `perm+750` for the third set, etc.

The result is a little different than what you have implemented above, but I think is overall a better mix.

DavidVorick:
Overall this looks pretty clean, just a few changes requested.

PR #2152, opened 2017-07-19
add host interaction weight to the hostdb

DavidVorick:
This was originally several nicer commits, but my .git folder corrupted somehow :/

The hostdb was already tracking the reliability of hosts, but this reliability was not actually being factored into the scores of the hosts.

Now it is factored in, although siac doesn't display it in the score breakdown I believe, still need a commit to extend siac.

### DIFF
--- a/modules/renter.go
+++ b/modules/renter.go
@@ -98,10 +98,10 @@ type HostDBEntry struct {
 	HistoricUptime   time.Duration `json:"historicuptime"`
 	ScanHistory      HostDBScans   `json:"scanhistory"`
 
-	HistoricFailedInteractions     uint64 `json:"historicFailedInteractions"`
-	HistoricSuccessfulInteractions uint64 `json:"historicSuccessfulInteractions"`
-	RecentFailedInteractions       uint64 `json:"recentFailedInteractions"`
-	RecentSuccessfulInteractions   uint64 `json:"recentSuccessfulInteractions"`
+	HistoricFailedInteractions     uint64 `json:"historicfailedinteractions"`
+	HistoricSuccessfulInteractions uint64 `json:"historicsuccessfulinteractions"`
+	RecentFailedInteractions       uint64 `json:"recentfailedinteractions"`
+	RecentSuccessfulInteractions   uint64 `json:"recentsuccessfulinteractions"`
 
 	LastHistoricUpdate types.BlockHeight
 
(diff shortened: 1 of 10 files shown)


modules/renter.go
```diff
@@ -98,10 +98,10 @@ type HostDBEntry struct {
 	HistoricUptime   time.Duration `json:"historicuptime"`
 	ScanHistory      HostDBScans   `json:"scanhistory"`
 
-	HistoricSuccessfulInteractions uint64 `json:"historicSuccessfulInteractions"`
 	HistoricFailedInteractions     uint64 `json:"historicFailedInteractions"`
-	RecentSuccessfulInteractions   uint64 `json:"recentSuccessfulInteractions"`
+	HistoricSuccessfulInteractions uint64 `json:"historicSuccessfulInteractions"`
 	RecentFailedInteractions       uint64 `json:"recentFailedInteractions"`
+	RecentSuccessfulInteractions   uint64 `json:"recentSuccessfulInteractions"`
```
lukechampine: eesh, somehow I missed that these fields are using camelCase instead of our conventional lowercase. As long as these haven't been exposed in the API yet, we could change them without breaking compatibility. Go ignores case when decoding.

modules/renter/hostdb/hostentry.go
```diff
@@ -0,0 +1,115 @@
...
+// than one block passed since the last update. This should be called every time
+// before the recent interactions are updated.  if passedTime is e.g. 10, this
+// means that the recent interactions were updated 10 blocks ago but never
+// since. So we need to apply the decay of 1 block before we append the recent
+// interactions from 10 blocks ago and then apply the decay of 9 more blocks in
+// which the recent interactions have been 0
+func updateHostHistoricInteractions(host *modules.HostDBEntry, bh types.BlockHeight) {
+	passedTime := bh - host.LastHistoricUpdate
+	if passedTime == 0 {
+		// no time passed. nothing to do.
```
lukechampine: might need a sanity check for `bh < host.LastHistoricUpdate`

DavidVorick: hmm. If the consensus db reset or something, this could actually cause issues. Good catch

DavidVorick:
Test should be fixed now. Also added some `siac` output to display the reliability better. Increased the impact of having low reliability. Changed uint64 to float for better precision around small numbers (which they typically are).

This will help with the download and download-repair issues that I'm seeing, but it's not a complete fix.

I think a complete fix will involve making the download code smart enough to prefer hosts that are cheaper and have strong download track records. There's a TODO spot in the codebase to allow this, so I don't think it'll be too much effort to add.

PR #2157, opened 2017-07-19
add file repair regression test

johnathanhowell:
This PR adds a regression test which verifies that for a file which is available on-disk locally, the renter will not download chunks to repair that file.


api/renterhost_test.go
```diff
@@ -21,6 +21,97 @@ import (
...
+
+	// download spending should not have increased
+	var rg RenterGET
+	err = st.getAPI("/renter", &rg)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if rg.FinancialMetrics.DownloadSpending.Cmp(types.NewCurrency64(0)) > 0 {
+		t.Fatalf("expected no download spending, got %v instead\n", rg.FinancialMetrics.DownloadSpending)
+	}
```
DavidVorick: need to extend this test twice:

1. delete a host and wait for redundancy to restore, then check balance

2. delete file locally and then delete a host and wait for redundancy to restore, then check balance.

This test is incomplete without that.

PR #2161, opened 2017-07-20
fix active pieces bug

lukechampine:


api/renterhost_test.go
```diff
@@ -1741,3 +1742,177 @@ func TestRedundancyReporting(t *testing.T) {
...
+			return nil
+		}
+		return errors.New("file redundancy not decremented: " + fmt.Sprint(rf.Files[0].Redundancy))
+	})
+	if err != nil {
+		t.Log(err)
+	}
+
+	// verify that the download fails
+	downloadPath = filepath.Join(st.dir, "test-downloaded-verify2.dat")
```
johnathanhowell: this is the same downloadPath as the last check, should probably be different

api/renterhost_test.go
```diff
@@ -1741,3 +1742,177 @@ func TestRedundancyReporting(t *testing.T) {
...
```
johnathanhowell: ditto

modules/renter/download.go
```diff
@@ -419,6 +419,7 @@ loop:
 		downloadComplete := incompleteChunk.download.downloadComplete
 		incompleteChunk.download.mu.Unlock()
 		if downloadComplete {
+
```
johnathanhowell: extra newline?

PR #2162, opened 2017-07-20
fix active pieces log formatting

lukechampine:
we may want to keep this open for a bit, since there's at least one report that the active pieces bug isn't quite fixed yet.


DavidVorick:
There are multiple reports that the active pieces bug is not yet fixed. I haven't tracked it down yet, but it seems to be dropping entire chunks at a time. Which means it's scheduling a whole set of pieces to download, but then for some reason that whole set is getting dropped.

PR #2165, opened 2017-07-21
compute diff in acceptTransactionSet more efficiently.

marcinja:
Computes the diff in `acceptTransactionSet` directly rather than by iterating through the whole transaction pool like in `updateSubscribersTransactions`. Working on a similar change for `ProcessConsensusChange` so that the entire transaction pool doesn't have to be remade after every consensus change.

### DIFF
--- a/modules/transactionpool/accept.go
+++ b/modules/transactionpool/accept.go
@@ -101,8 +101,12 @@ func (tp *TransactionPool) checkTransactionSetComposition(ts []types.Transaction
 }
 
 // handleConflicts detects whether the conflicts in the transaction pool are
-// legal children of the new transaction pool set or not.
+// legal children of the new transaction pool set or not. If the conflict can be
+// handled itt also creates a TransactionPoolDiff by adding the given
+// transaction set to the transaction pool, and removing any transaction sets
+// that conflict with it. It then updates all subscribers with that diff.
 func (tp *TransactionPool) handleConflicts(ts []types.Transaction, conflicts []TransactionSetID, txnFn func([]types.Transaction) (modules.ConsensusChange, error)) error {
+	diff := new(modules.TransactionPoolDiff)
 	// Create a list of all the transaction ids that compose the set of
 	// conflicts.
 	conflictMap := make(map[types.TransactionID]TransactionSetID)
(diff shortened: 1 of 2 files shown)


modules/transactionpool/accept.go
```diff
@@ -306,6 +337,30 @@ func (tp *TransactionPool) acceptTransactionSet(ts []types.Transaction, txnFn fu
...
+		Transactions: ts,
+	}
+	// Add this diff to our set of subscriber diffs.
+	tp.subscriberSets[setID] = ut
+	diff := new(modules.TransactionPoolDiff)
+	diff.AppliedTransactions = append(diff.AppliedTransactions, ut)
+	for _, subscriber := range tp.subscribers {
+		subscriber.ReceiveUpdatedUnconfirmedTransactions(diff)
+	}
+
```
ChrisSchinnerl: This looks good to me. The only thing that is bothering me a little is, that `handleConflicts` implicitly accepts the tset using almost the same code. Maybe we could simplify the code by checking for conflicts before checking the composition, fees etc. in `acceptTransactionSet`. 
If there are no conflicts we continue with `ts` and if there were,`handleConflicts` creates and returns the `superset` which is then checked and used to create the diff instead of `ts`.

modules/transactionpool/accept.go
```diff
@@ -284,7 +314,8 @@ func (tp *TransactionPool) acceptTransactionSet(ts []types.Transaction, txnFn fu
 		}
 	}
 	if len(conflicts) > 0 {
-		return tp.handleConflicts(ts, conflicts, txnFn)
+		err := tp.handleConflicts(ts, conflicts, txnFn)
+		return err
```
ChrisSchinnerl: I personally prefer the return in the same line. Do we have a convention for that?

modules/transactionpool/accept.go
```diff
@@ -101,8 +101,12 @@ func (tp *TransactionPool) checkTransactionSetComposition(ts []types.Transaction
 }
 
 // handleConflicts detects whether the conflicts in the transaction pool are
-// legal children of the new transaction pool set or not.
+// legal children of the new transaction pool set or not. If the conflict can be
+// handled itt also creates a TransactionPoolDiff by adding the given
```
ChrisSchinnerl: "itt"

modules/transactionpool/accept.go
```diff
@@ -198,24 +202,45 @@ func (tp *TransactionPool) handleConflicts(ts []types.Transaction, conflicts []T
...
-	for _, diff := range cc.FileContractDiffs {
-		tp.knownObjects[ObjectID(diff.ID)] = setID
+	for _, txDiff := range cc.FileContractDiffs {
+		tp.knownObjects[ObjectID(txDiff.ID)] = setID
 	}
-	for _, diff := range cc.SiafundOutputDiffs {
-		tp.knownObjects[ObjectID(diff.ID)] = setID
+	for _, txDiff := range cc.SiafundOutputDiffs {
+		tp.knownObjects[ObjectID(txDiff.ID)] = setID
 	}
```
ChrisSchinnerl: @DavidVorick are we sure that we are not missing ids of the deduplicated set here by not using `relatedObjectIDs` ?

PR #2171, opened 2017-07-22
Added missing property 'local' in Gateway's peer response

mharkus:
This commit will add the 'local' property in the /gateway's sample response. Although I don't know the description of it.

### DIFF
--- a/doc/api/Gateway.md
+++ b/doc/api/Gateway.md
@@ -55,7 +55,8 @@ returns information about the gateway, including the list of connected peers.
         // peers, as inbound peers are easily manipulated by an adversary.
         "inbound":    Boolean,
 
-        // what does local mean?
+        // local is true if the peer's IP address belongs to a local address
+        // range such as 192.168.x.x or 127.x.x.x
         "local":      Boolean
     }
 }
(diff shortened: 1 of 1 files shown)


doc/api/Gateway.md
```diff
@@ -53,7 +53,10 @@ returns information about the gateway, including the list of connected peers.
         // inbound is true when the peer initiated the connection. This field
         // is exposed as outbound peers are generally trusted more than inbound
         // peers, as inbound peers are easily manipulated by an adversary.
-        "inbound":    Boolean
+        "inbound":    Boolean,
+
+        // what does local mean?
+        "local":      Boolean
```
petabytestorage: connected through the `loopback` interface

lukechampine: the precise definition is here: https://github.com/NebulousLabs/Sia/blob/master/modules/netaddress.go#L63

So "local" can mean another node on your LAN, e.g. `10.0.0.5`, not just another node on the same computer

petabytestorage: good point. those "private ip address ranges" are from rfc 1918 (v4) and rfc 4193 (v6).

lukechampine:
thanks!

PR #2184, opened 2017-07-25
Update Makefile

S-anasol:
image


lukechampine:
this dependency is annoying. We could strip it out if we vendored cobra.

PR #2191, opened 2017-07-27
get files detail

tlightsky:
it's a new api that could know exact which chunk is stored on which host,
and every chunk's health

i really want more info that about which piece is actively being processed by which worker and all the progress of the worker

@DavidVorick

### DIFF
--- a/modules/renter.go
+++ b/modules/renter.go
@@ -102,6 +102,13 @@ type FileDetail struct {
 	IsOffline bool       `json:"isoffline"`
 }
 
+// FileDetails represents a sortable slice of FileDetail.
+type FileDetails []*FileDetail
+
+func (s FileDetails) Len() int           { return len(s) }
+func (s FileDetails) Less(i, j int) bool { return s[i].IP < s[j].IP }
+func (s FileDetails) Swap(i, j int)      { s[i], s[j] = s[j], s[i] }
+
 // A HostDBEntry represents one host entry in the Renter's host DB. It
 // aggregates the host's external settings and metrics with its public key.
 type HostDBEntry struct {
--- a/modules/renter.go
+++ b/modules/renter.go
@@ -93,21 +93,26 @@ type FileDetailInfo struct {
 	Redundancy     float64           `json:"redundancy"`
 	UploadProgress float64           `json:"uploadprogress"`
 	Expiration     types.BlockHeight `json:"expiration"`
-	Details        [][][]*FileDetail `json:"details"`
+	Details        *FileDetail       `json:"details"`
 }
 
-// FileDetail is the detail of pieces
-type FileDetail struct {
+type HostDetail struct {
 	IP        NetAddress `json:"host"`
 	IsOffline bool       `json:"isoffline"`
 }
 
-// FileDetails represents a sortable slice of FileDetail.
-type FileDetails []*FileDetail
+// FileDetail is the detail of pieces
+type FileDetail struct {
+	Hosts  []*HostDetail `json:"hosts"`
+	Chunks [][][]int     `json:"chunks"`
+}
 
-func (s FileDetails) Len() int           { return len(s) }
-func (s FileDetails) Less(i, j int) bool { return s[i].IP < s[j].IP }
-func (s FileDetails) Swap(i, j int)      { s[i], s[j] = s[j], s[i] }
+// FileDetails represents a sortable slice of FileDetail.
+// type FileDetails []*FileDetail
+
+// func (s FileDetails) Len() int           { return len(s) }
+// func (s FileDetails) Less(i, j int) bool { return s[i].IP < s[j].IP }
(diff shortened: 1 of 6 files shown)


tlightsky:
the host info in some exact pieces keeps changing which might imply there is more than one host hold a piece which i have not noticed before

lukechampine:
I'm ok with the general idea of this, but it needs to be done carefully.

The main issue is that this is a LOT of data. Currently, you are sending `numChunks * numHosts` `FileDetail` objects. Each `FileDetail` object will be about 40 bytes in JSON. So if you are storing a 1 TB file across 50 hosts, that's (1 TB / 4 MB) * 50 `FileDetail` objects (about 12 million), which encodes to 500 MB of JSON.

Even if we used a binary encoding, a 1 TB file is going to require many MB to describe. it's just a lot of data, there's no way around it. So if we're going to do this, the API call needs to have some form of pagination. Probably this will mean requesting a range of chunk indexes.

I suggest separating the host information from the chunk information. Something like:
```json
{
    "hosts": [
        {"pubkey": "foo", "isoffline": false},
        {"pubkey": "bar", "isoffline": false}
    ],
    "chunks": [
        [0, -1, 1],
        [0, 1, -1]
    ]
}
```
here, `chunks` is an array of chunks, where each chunk is an array of pieces, and each element in the piece array is the index of a host in `hosts`. -1 means the piece isn't stored on any host. Also note that we're trying to switch from `netaddress` to `pubkey`.

For a 1 TB file on 50 hosts, there are about 250,000 chunks, and each element of `chunks` is about 140 bytes of JSON, for a total of 35 MB. That's still probably too much to send in one request, but pagination will help.

tlightsky:
- i think pagination is reasonable
- `separating the host information from the chunk information` is like a light weight compression which i also agree
- use `pubkey` rather than `netaddress` might not obvious to end user, maybe i can keep them both cause there would be 50 atmost in recent version(i think)

so the steps maybe like:

1. separate the host information from the chunk information
2. add pubkey to result
3. separate file details to each file request rather than a big filesdetail request
4. pagination each file detail request

@lukechampine look forward to any further suggestion :)

tlightsky:
there are 6 chunks for a file with 45678 bytes in test, a little strange

modules/renter.go
```diff
@@ -83,6 +83,38 @@ type FileInfo struct {
...
+// first dimension of details is chunk id, second is chunk redundency hosts
+type FileDetailInfo struct {
+	SiaPath        string            `json:"siapath"`
+	Filesize       uint64            `json:"filesize"`
+	Available      bool              `json:"available"`
+	Renewing       bool              `json:"renewing"`
+	Redundancy     float64           `json:"redundancy"`
+	UploadProgress float64           `json:"uploadprogress"`
+	Expiration     types.BlockHeight `json:"expiration"`
+	Details        *FileDetail       `json:"details"`
```
lukechampine: This doesn't need to be a pointer. Also, `Details` isn't a very descriptive name.

I'm wondering if perhaps we should get rid of the duplication between `FileInfo` and `FileDetailInfo`. It's not too bad if you have to call two separate API endpoints. I think I would prefer:
```go
type FileDetailsInfo struct {
    Hosts []HostDetail `json:"hosts"`
}
```

modules/renter/files.go
```diff
@@ -146,6 +149,55 @@ func (f *file) redundancy(isOffline func(types.FileContractID) bool) float64 {
...
+				IP:        fc.IP,
+				IsOffline: isoffline,
+			}
+			hostMap[string(hd.IP)] = len(fd.Hosts)
+			fd.Hosts = append(fd.Hosts, hd)
+		}
+
+		for _, p := range fc.Pieces {
+			if fd.Chunks[p.Chunk][p.Piece] == nil {
+				fd.Chunks[p.Chunk][p.Piece] = make([]int, 0)
```
lukechampine: instead do:
```go
// need to create a zero-length slice so that it is JSON-encoded as [], not null
fd.Chunks[p.Chunk][p.Piece] = []int{}
```

modules/renter/files.go
```diff
@@ -240,6 +292,72 @@ func (r *Renter) FileList() []modules.FileInfo {
...
+		offline := r.hostContractor.IsOffline(id)
+		contract, exists := r.hostContractor.ContractByID(id)
+		if !exists {
+			return true
+		}
+		return offline || !contract.GoodForRenew
+	}
+
+	fd := file.detail(isOffline)
+	p := paginater.New(len(fd.Chunks), pagingNum, current, 5)
```
lukechampine: I don't think we need a new `paginater` dependency. The math is pretty simple.

tlightsky:
@lukechampine sorry, on a trip(bad wifi connection) recently.
i'll try to fix those problems

lukechampine:
Closing because the renter architecture is being overhauled over the coming weeks. The low-level file semantics will likely change in a way that's incompatible with this PR.

PR #2210, opened 2017-08-01
[WIP] add memlogging options

marcinja:


api/daemon.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"net/http"
+	"strings"
+
+	"github.com/NebulousLabs/Sia/build"
+	"github.com/julienschmidt/httprouter"
+)
+
+func (api *API) memloggingGET(w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
```
DavidVorick: docstring

api/daemon.go
```diff
@@ -0,0 +1,34 @@
...
+	if build.MEMLOGGING {
+		w.Write([]byte("memlogging enabled"))
+	} else {
+		w.Write([]byte("memlogging disabled"))
+	}
+}
+
+func (api *API) memloggingPOST(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
```
DavidVorick: docstring

api/daemon.go
```diff
@@ -0,0 +1,34 @@
...
+	val := strings.ToLower(req.FormValue("set")) // response is case-insensitive
```
DavidVorick: call it "active", unless @lukechampine has a better idea

siad/main.go
```diff
@@ -168,6 +170,7 @@ func main() {
 	root.Flags().StringVarP(&globalConfig.Siad.Modules, "modules", "M", "cghrtw", "enabled modules, see 'siad modules' for more info")
 	root.Flags().BoolVarP(&globalConfig.Siad.AuthenticateAPI, "authenticate-api", "", false, "enable API password protection")
 	root.Flags().BoolVarP(&globalConfig.Siad.AllowAPIBind, "disable-api-security", "", false, "allow siad to listen on a non-localhost address (DANGEROUS)")
+	root.Flags().BoolVarP(&globalConfig.Siad.MemLogging, "enable-memlogging", "", false, "enable the logging of the sizes of various module data structures")
```
DavidVorick: since it's in the api we don't need a flag anymore

lukechampine: *should* this be in the API? Is there a technical reason why it can't be just a `siad` flag? That makes more sense to me than adding a debugging hook to the public API.

DavidVorick: It's not documented, therefore not a part of the compatibility promise.

Multiple people have been asking for runtime debugging options, and I'm inclined to agree with them actually. It's very annoying when you have something interesting happen to your node but it was running without the profile or debug flags, or you want a logging message that is DEBUG only. So I think it's actually better to be moving a lot of this stuff to the runtime, so you can disable and enable it in real time.

lukechampine: ok, I see the advantage of being able to enable it later on without having to start `siad` with a special flag.
what about moving it to the `siad` API (along with `/stop`, etc.)?

marcinja:
I removed the `siad` flag. The new api calls are defined in the `siac` directory along with the other daemon api calls.  I also changed it so that it's now a JSON response like the other calls.

in action:

```
marcinja@mbp:~ $ siac daemon memlogging 
Memlogging Active:  true
marcinja@mbp:~ $ siac daemon memlogging false
marcinja@mbp:~ $ siac daemon memlogging      
Memlogging Active:  false
```

siac/main.go
```diff
@@ -102,7 +102,9 @@ func apiGet(call string) (*http.Response, error) {
 // if the response status is not 2xx.
 func getAPI(call string, obj interface{}) error {
 	resp, err := apiGet(call)
+
 	if err != nil {
+		println("HERESS")
```
marcinja: will remove this

build/memlogging.go
```diff
@@ -0,0 +1,4 @@
+package build
+
+// LOGMEMORY enables memory logging of module data structures if true
+var MEMLOGGING = false
```
DavidVorick: All caps is only for constants. This can just be called MemLogging I think

modules/miner/persist.go
```diff
@@ -76,6 +77,19 @@ func (m *Miner) saveSync() error {
 	return persist.SaveJSON(settingsMetadata, m.persist, filepath.Join(m.persistDir, settingsFile))
 }
 
+func (m *Miner) statsLogger() {
```
DavidVorick: needs a docstring, don't forget it

modules/transactionpool/persist.go
```diff
@@ -27,6 +27,18 @@ var (
 	errNilFeeMedian = errors.New("no fee median found")
 )
 
+func (tp *TransactionPool) statsLogger() {
```
DavidVorick: docstring

modules/wallet/database.go
```diff
@@ -58,6 +59,17 @@ var (
 	errNoKey = errors.New("key does not exist")
 )
 
+func (w *Wallet) statsLogger() {
```
DavidVorick: docstring

siac/daemoncmd.go
```diff
@@ -46,6 +60,34 @@ type daemonVersion struct {
...
+
+	case 1:
+		err := post("/daemon/memlogging", "active="+args[0])
+		if err != nil {
+			die("Could not update memlogging settings:", err)
+		}
+	default:
+		fmt.Println("Error: expected 0 or 1 params to this command")
+	}
+}
```
DavidVorick: docstrings on this whole file

siad/server.go
```diff
@@ -366,6 +372,29 @@ func (srv *Server) daemonHandler(password string) http.Handler {
...
+// memlogging on if true, off if false.
+func (srv *Server) memloggingPOST(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	val := strings.ToLower(req.FormValue("active")) // response is case-insensitive
+	if val == "true" {
+		build.MEMLOGGING = true
+		api.WriteSuccess(w)
+		return
+	}
+	if val == "false" {
+		build.MEMLOGGING = false
```
DavidVorick: This is going to cause a race condition, because you are writing to this variable outside of a lock, and then reading it in the loggers inside of a lock. Cleanest way to avoid this I think is to just use atomics, and then instead of having the build variable be a bool, have it be a 1 or a 0.

The package for atomics is `sync/atomic`. Convention requires you prefix the name with the word 'atomic', so the global variable name becomes `AtomicMemLogging`. Use `atomic.StoreUint64` and `atomic.LoadUint64` to read and set the value for the memlogging.

DavidVorick:
You need a test demonstrating that the memlogging is running, and that it is being enabled and disabled by the api.

To check for logging statements in the api, what we do is is close the file, then load it, then search for a specific string that you know appears only in the logging message you were looking for. The test doesn't have to be super comprehensive, just prove the logging is running correctly in a single package.

This test written against the current code should complain about a race condition.

tbenz9:
@marcinja is this ready for another review?  If you fix the file conflicts I can make sure a dev takes a look in the next few days.

PR #2214, opened 2017-08-02
add sanity check subscriber to tpool

marcinja:
This PR adds a new subscriber to the tpool which creates and maintains a map of `transactionSets` from subscriber diffs. This confirms that the transaction pool is correctly altering its own internal state consistently the the diffs it sends out to subscribers.

### DIFF
--- a/modules/transactionpool/sanitychecksubscriber.go
+++ b/modules/transactionpool/sanitychecksubscriber.go
@@ -1,0 +1,70 @@
+package transactionpool
+
+import (
+	"github.com/NebulousLabs/Sia/modules"
+	"github.com/NebulousLabs/Sia/types"
+	"github.com/NebulousLabs/fastrand"
+)
+
+type sanityCheckSubscriber struct {
+	transactionSets map[TransactionSetID][]types.Transaction
+}
+
+// newSanityCheckSubscriber returns a new sanityCheckSubscriber with no
+// transaction sets.
+func newSanityCheckSubscriber() *sanityCheckSubscriber {
+	return &sanityCheckSubscriber{
+		transactionSets: make(map[TransactionSetID][]types.Transaction),
+	}
+}
+
+// ReceiveUpdatedUnconfirmedTransactions updates the sanityCheckSubscriber's
+// transactionSets using the diff sent from the tpool. It is needed to satisfy
+// the TransactionPoolSubscriber interface.
+func (s *sanityCheckSubscriber) ReceiveUpdatedUnconfirmedTransactions(diff *modules.TransactionPoolDiff) {
+	for _, setID := range diff.RevertedTransactions {
+		delete(s.transactionSets, TransactionSetID(setID))
+	}
+	for _, unconfirmedTxnSet := range diff.AppliedTransactions {
+		s.transactionSets[TransactionSetID(unconfirmedTxnSet.ID)] = unconfirmedTxnSet.Transactions
+	}
+}
+
+// subscriberSanityCheck performs a sanity check on the transaction pool. It
+// panics if the map of transaction sets in the subscriber's state is not
+// exactly the same as the map of transaction sets in the transaction pool.
+func (tp *TransactionPool) subscriberSanityCheck(s *sanityCheckSubscriber) {
+	// 1/30 chance of running this check because it is expensive
+	if fastrand.Intn(30) != 0 {
+		return
+	}
+
+	if len(tp.transactionSets) != len(s.transactionSets) {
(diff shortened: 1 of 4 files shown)


modules/transactionpool/sanitychecksubscriber.go
```diff
@@ -0,0 +1,71 @@
...
+		s.transactionSets[TransactionSetID(unconfirmedTxnSet.ID)] = unconfirmedTxnSet.Transactions
+	}
+}
+
+// subscriberSanityCheck performs a sanity check on the transaction pool. It
+// panics if the map of transaction sets in the subscriber's state is not
+// exactly the same as the map of transaction sets in the transaction pool.
+func (tp *TransactionPool) subscriberSanityCheck() {
+	// 1/30 chance of running this check because it is expensive
+	if fastrand.Intn(30) != 0 || !build.DEBUG {
```
lukechampine: Go has short-circuiting `if` statements, so put the `!build.DEBUG` first

modules/transactionpool/sanitychecksubscriber.go
```diff
@@ -0,0 +1,71 @@
+package transactionpool
+
+import (
+	"github.com/NebulousLabs/Sia/build"
+	"github.com/NebulousLabs/Sia/modules"
+	"github.com/NebulousLabs/Sia/types"
+	"github.com/NebulousLabs/fastrand"
+)
+
+type sanityCheckSubscriber struct {
```
DavidVorick: docstring

modules/transactionpool/sanitychecksubscriber.go
```diff
@@ -0,0 +1,71 @@
+package transactionpool
+
+import (
```
DavidVorick: would be good to have a comment here explaining the purpose of the file, e.g.

```
package transactionpool

// sanitychecksubscriber.go is a tool used during debugging to verify that
// etc etc etc

import (
```

modules/transactionpool/sanitychecksubscriber.go
```diff
@@ -0,0 +1,71 @@
...
+		subscriberSet, ok := tp.sanityCheck.transactionSets[tpoolSetID]
+		if !ok {
+			// Doesn't contain a set the tpool contains.
+			panic("sanityCheckSubscriber doesn't contain same transaction set as tpool")
+		}
+
+		if len(tpoolSet) != len(subscriberSet) {
+			panic("sanityCheckSubscriber txn set has different size than corresponding set in tpool")
+		}
+		// Check that both sets contain the exact same transactions
```
DavidVorick: comment is in the wrong spot

modules/transactionpool/sanitychecksubscriber.go
```diff
@@ -0,0 +1,71 @@
...
+		}
+		for _, txn := range subscriberSet {
+			_, ok := tpoolTxns[txn.ID()]
+			if !ok {
+				// Doesn't contain a transacion the tpool contains.
+				panic("sanityCheckSubscriber doesn't contain the same transaction in the same set as tpool")
+			}
+		}
+	}
+}
```
DavidVorick: Overall, logic looks great

modules/transactionpool/transactionpool.go
```diff
@@ -62,6 +63,7 @@ type (
 		// transaction pool, all prior consensus changes are sent to the new
 		// subscriber.
 		subscribers []modules.TransactionPoolSubscriber
+		sanityCheck *sanityCheckSubscriber
```
DavidVorick: would pick a different name for this variable. Perhaps 'basicSubscriber'

modules/transactionpool/transactionpool.go
```diff
@@ -97,6 +99,12 @@ func New(cs modules.ConsensusSet, g modules.Gateway, persistDir string) (*Transa
 		persistDir: persistDir,
 	}
 
+	if build.DEBUG {
+		sub := newSanityCheckSubscriber()
+		tp.subscribers = append(tp.subscribers, sub)
+		tp.sanityCheck = sub
```
DavidVorick: Have `newSanityCheckSubscriber` take a tpool as an argument and then have it call `Subscribe` on the tpool, instead of doing it manually after initialization.

DavidVorick:
One last change. Instead of `basicSubscriber` and `sanityCheckSubscriber` and `sanitycheck.go` etc, can we use the name `diffConsistencySubscriber` everywhere? I think this is the most complete, it tells you what type of sanity check it is, it checks that the diffs are consistent with the state.

PR #2242, opened 2017-08-16
update defragFee to use transaction pool fee estimates

### LINKED ISSUE #2235: Does defrag function still need 10SC Tx fee?
The Tx fee for normal transactions have been lowered a lot, but defrag transactions still use 10SC as a fee. Is that correct or shold they also be adapted?

avahowell:
Closes #2235

### DIFF
--- a/modules/wallet/consts.go
+++ b/modules/wallet/consts.go
@@ -46,14 +46,9 @@ func dustValue() types.Currency {
 
 // defragFee is the miner fee paid to miners when performing a defrag
 // transaction.
-//
-// TODO: These need to be functions of the wallet that interact with the
-// transaction pool.
-func defragFee() types.Currency {
-	// 35 outputs at an estimated 250 bytes needed per output means about a 10kb
-	// total transaction, much larger than your average transaction. So you need
-	// a lot of fees.
-	return types.SiacoinPrecision.Mul64(10)
+func (w *Wallet) defragFee(txnSize uint64) types.Currency {
+	minFee, _ := w.tpool.FeeEstimation()
+	return minFee.Mul64(txnSize)
 }
 
 func init() {
--- a/modules/wallet/consts.go
+++ b/modules/wallet/consts.go
@@ -44,13 +44,6 @@ func dustValue() types.Currency {
 	return types.SiacoinPrecision
 }
 
-// defragFee is the miner fee paid to miners when performing a defrag
-// transaction.
-func (w *Wallet) defragFee(txnSize uint64) types.Currency {
-	minFee, _ := w.tpool.FeeEstimation()
-	return minFee.Mul64(txnSize)
-}
-
 func init() {
 	// Sanity check - the defrag threshold needs to be higher than the batch
 	// size plus the start index.
--- a/modules/wallet/defrag.go
+++ b/modules/wallet/defrag.go
@@ -14,7 +14,7 @@ var (
 
 // createDefragTransaction creates a transaction that spends multiple existing
 // wallet outputs into a single new address.
-func (w *Wallet) createDefragTransaction() ([]types.Transaction, error) {
+func (w *Wallet) createDefragTransaction(fee types.Currency) ([]types.Transaction, error) {
 	consensusHeight, err := dbGetConsensusHeight(w.dbTx)
 	if err != nil {
 		return nil, err
@@ -78,7 +78,6 @@ func (w *Wallet) createDefragTransaction() ([]types.Transaction, error) {
 	}
 
 	// Create the defrag transaction.
-	fee := defragFee()
 	refundAddr, err := w.nextPrimarySeedAddress(w.dbTx)
 	if err != nil {
 		return nil, err
@@ -130,9 +129,16 @@ func (w *Wallet) threadedDefragWallet() {
 		w.mu.Unlock()
 		return
 	}
+	w.mu.Unlock()
+
+	// get the required fee for the defrag transaction
+	// 35 outputs at an estimated 250 bytes needed per output means about a 10kb
+	// total transaction.
+	fee := w.defragFee(10000)
 
 	// Create the defrag transaction.
-	txnSet, err := w.createDefragTransaction()
+	w.mu.Lock()
+	txnSet, err := w.createDefragTransaction(fee)
 	w.mu.Unlock()
 	if err == errDefragNotNeeded {
 		// benign
--- a/modules/wallet/defrag.go
+++ b/modules/wallet/defrag.go
@@ -15,6 +15,8 @@ var (
 // managedCreateDefragTransaction creates a transaction that spends multiple existing
 // wallet outputs into a single new address.
 func (w *Wallet) managedCreateDefragTransaction() ([]types.Transaction, error) {
+	minFee, _ := w.tpool.FeeEstimation()
+
 	w.mu.Lock()
 	defer w.mu.Unlock()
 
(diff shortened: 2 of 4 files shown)


modules/wallet/defrag.go
```diff
@@ -130,9 +129,16 @@ func (w *Wallet) threadedDefragWallet() {
 		w.mu.Unlock()
 		return
 	}
+	w.mu.Unlock()
+
+	// get the required fee for the defrag transaction
+	// 35 outputs at an estimated 250 bytes needed per output means about a 10kb
+	// total transaction.
+	fee := w.defragFee(10000)
```
DavidVorick: Does this even merit a helper function anymore? I think this code can just be inlined. Comment also needs to be formatted. Also, it's probably worth doing the full multiplication here. E.g. numOutputs * sizeAvgOutput instead of hardcoding 10,000.

Ideally the transaction builder itself could figure out how many bytes are needed to sign a transaction, but that's more complexity than we need at the moment.

avahowell: Without this being a helper function, testing this functionality is more difficult and prone to nondeterminism. Having it be a helper function like this allows us to have a test that cleanly verifies the defrag fee is correct.

I updated the PR to include the full multiplication and formatted the comment.

modules/wallet/consts.go
```diff
@@ -46,14 +46,9 @@ func dustValue() types.Currency {
...
-// transaction pool.
-func defragFee() types.Currency {
-	// 35 outputs at an estimated 250 bytes needed per output means about a 10kb
-	// total transaction, much larger than your average transaction. So you need
-	// a lot of fees.
-	return types.SiacoinPrecision.Mul64(10)
+func (w *Wallet) defragFee(txnSize uint64) types.Currency {
+	minFee, _ := w.tpool.FeeEstimation()
+	return minFee.Mul64(txnSize)
 }
```
DavidVorick: can delete this now that it's not used anywhere

modules/wallet/defrag_test.go
```diff
@@ -198,3 +199,45 @@ func TestDefragOutputExhaustion(t *testing.T) {
...
+func (t *mockTpool) Close() error                                      { return nil }
+func (t *mockTpool) FeeEstimation() (types.Currency, types.Currency)   { return t.minFee, t.maxFee }
+func (t *mockTpool) ProcessConsensusChange(cc modules.ConsensusChange) {}
+func (t *mockTpool) PurgeTransactionPool()                             {}
+func (t *mockTpool) Transaction(id types.TransactionID) (types.Transaction, []types.Transaction, bool) {
+	return types.Transaction{}, []types.Transaction{}, true
+}
+func (t *mockTpool) TransactionList() []types.Transaction                                  { return []types.Transaction{} }
+func (t *mockTpool) TransactionPoolSubscribe(subscriber modules.TransactionPoolSubscriber) {}
+func (t *mockTpool) Unsubscribe(subscriber modules.TransactionPoolSubscriber)              {}
```
DavidVorick: You should be able to use a real tpool as the base for your mock tpool, right?  Then you just overwrite the FeeEstimation function, and it should be a lot cleaner overall.

modules/wallet/consts.go
avahowell: huh? This is still called from the tests and from `threadedDefragWallet`, for the reasons outlined in the last comment

modules/wallet/defrag.go
```diff
@@ -130,9 +129,17 @@ func (w *Wallet) threadedDefragWallet() {
...
+	// get the required fee for the defrag transaction.  35 outputs at an
+	// estimated 250 bytes needed per output means about a 10kb total
+	// transaction.
+	sizeAvgOutput := uint64(250) // bytes
+	fee := w.defragFee(sizeAvgOutput * defragBatchSize)
 
 	// Create the defrag transaction.
-	txnSet, err := w.createDefragTransaction()
+	w.mu.Lock()
+	txnSet, err := w.createDefragTransaction(fee)
```
lukechampine: why is the fee calculated outside `createDefragTransaction`? Wouldn't the function be in a better position to calculate the fee, since it knows how many outputs it's defragging?

avahowell: the fee is calculated outside of `createDefragTransaction` because fee calculation requires calling an exported method (`tpool.FeeEstimation()`), and `createDefragTransaction` is called under the wallet's lock.

modules/wallet/consts.go
DavidVorick: The function as it stands violates locking convention, which is that unexported functions of a module need to be called under lock. But if you do that, this function calls out to another module, which is a deadlock risk. Needs to be named managedDefragFee at least.

But this code is identical to all code that calculates fees, I don't see why it needs a special helper function. I believe this function has been made obsolete, and the code should just be inlined everywhere.

modules/wallet/defrag.go
lukechampine: agreed, `managedCreateDefragTransaction` seems fine to me

avahowell:
Rebased and updated with latest review comments

modules/wallet/defrag.go
```diff
@@ -78,11 +81,18 @@ func (w *Wallet) createDefragTransaction() ([]types.Transaction, error) {
...
 	if err != nil {
 		return nil, err
 	}
+
+	// get the transaction fee
+	w.mu.Unlock()
+	minFee, _ := w.tpool.FeeEstimation()
+	sizeAvgOutput := uint64(250)
+	fee := minFee.Mul64(sizeAvgOutput * defragBatchSize)
+	w.mu.Lock()
```
DavidVorick: I worry about patterns like these. When you see a `lock; defer unlock;` at the top of a function, you generally make assumptions about things that you can modify, and about consistency. The fact that it actually releases the lock midway through the function can be a problem, especially if you write code that's expecting values in the wallet not to change.

DavidVorick: It's definitely a much safer pattern to grab the `minFee` before you lock the wallet at all.

modules/wallet/defrag.go
```diff
@@ -12,9 +12,17 @@ var (
...
 
-// createDefragTransaction creates a transaction that spends multiple existing
+// managedCreateDefragTransaction creates a transaction that spends multiple existing
 // wallet outputs into a single new address.
-func (w *Wallet) createDefragTransaction() ([]types.Transaction, error) {
+func (w *Wallet) managedCreateDefragTransaction() ([]types.Transaction, error) {
+	// get the transaction fee
+	minFee, _ := w.tpool.FeeEstimation()
+	sizeAvgOutput := uint64(250)
+	fee := minFee.Mul64(sizeAvgOutput * defragBatchSize)
```
DavidVorick: Sorry to keep iterating on this, but this is another pattern that makes it easier to slip bugs into the code. The fee here is calculated completely separately from the place where the transactions are created, which means verifying correctness requires jumping around in the codebase. We want to write code so that pieces of code which are dependent on eachother appear right next to eachother.

So the part where we grab `minFee` should be up at the top of the function, but then the part where we calculate the actual fee should be down by the transaction builder.

avahowell: Gotcha. Amended the last commit with the suggested changes.

modules/wallet/defrag.go
```diff
@@ -130,10 +139,10 @@ func (w *Wallet) threadedDefragWallet() {
 		w.mu.Unlock()
 		return
 	}
+	w.mu.Unlock()
```
lukechampine: this can change to an `RLock` now:
```go
w.mu.RLock()
unlocked := w.unlocked
w.mu.RUnlock()
if !unlocked {
    // Can't defrag if the wallet is locked.
    return
}
```

PR #2245, opened 2017-08-17
Add removeTransaction method to tpool

marcinja:
The `removeTransaction` method should be used in a new `ProcessConsensusChange` method for the tpool to remove any transaction that conflicts with the consensus set and all transactions that depend on its outputs. This PR still needs its own tests and should be tested with #2214 also. 

Right now this method can't actually be called anywhere usefully.

### DIFF
--- a/modules/transactionpool/update.go
+++ b/modules/transactionpool/update.go
@@ -458,6 +458,8 @@ func (tp *TransactionPool) findObjectLocations(oid ObjectID) ([]*objectLocation,
 	return locations, true
 }
 
+// removeTransaction removes a transaction and any transactions that spend its
+// outputs from the tpool.
 func (tp *TransactionPool) removeTransaction(txn types.Transaction, cc *modules.ConsensusChange) {
 	// removedTxnIndices is used to keep track of which need to be removed from
 	// which sets. Sets are changed (and a diff is created) once all changes
--- a/modules/transactionpool/update.go
+++ b/modules/transactionpool/update.go
@@ -387,3 +387,185 @@ func (tp *TransactionPool) PurgeTransactionPool() {
 	tp.purge()
 	tp.mu.Unlock()
 }
+
+// objectLocations are used to pinpoint the exact locations of an object in a
+// transaction set so as to avoid recomputing ObjectIDs when searching for
+// objects.
+type objectLocation struct {
+	setID   TransactionSetID
+	txIndex int
+}
+
+// findObjectLocation returns the location of an object from its object id and
+// true if it has been found. If the object's set or transaction is not found,
+// the returned bool will be false.
+func (tp *TransactionPool) findObjectLocation(oid ObjectID) ([]*objectLocation, bool) {
+	setID, ok := tp.knownObjects[oid]
+	if !ok {
+		return nil, false
+	}
+	set := tp.transactionSets[setID]
+
+	var location *objectLocation
+	location.setID = setID
+	// Iterate over all objects in all transactions the set until you find
+	// the transaction containing this object.
+	for i := 0; i < len(set); i++ {
+		for _, sci := range set[i].SiacoinInputs {
+			id := ObjectID(sci.ParentID)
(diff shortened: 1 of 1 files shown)


modules/transactionpool/update.go
```diff
@@ -387,3 +388,245 @@ func (tp *TransactionPool) PurgeTransactionPool() {
...
+// transaction set so as to avoid recomputing ObjectIDs when searching for
+// objects.
+type objectLocation struct {
+	setID   TransactionSetID
+	txIndex int
+}
+
+// findObjectLocations finds the locations that reference the given object as an
+// input.
+func (tp *TransactionPool) findObjectLocations(oid ObjectID) ([]*objectLocation, bool) {
```
marcinja: This method's implementation will need to be changed along with changes to transaction set types, since the goal for the tpool is to have child sets separate from parent sets if the fees work out that way.

modules/transactionpool/update.go
```diff
@@ -387,3 +388,245 @@ func (tp *TransactionPool) PurgeTransactionPool() {
...
+		removalStack = removalStack[:n-1]
+
+		// Get the location of the object just popped from the stack.
+		// TODO: this needs to be a func that gets locations from inputs that refer to
+		// outputs in the stack
+		nextObjectLocations, ok := objectLocations[nextObjectID]
+		if !ok {
+			nextObjectLocations, ok = tp.findObjectLocations(nextObjectID)
+			if !ok {
+				//TODO: Decide: Is this correct behavior? should this ever happen?
```
marcinja: If an object was marked for removal, and it cannot be found in the tpool then something crazy has probably happened. Not sure what the correct behavior here should be.

modules/transactionpool/update.go
```diff
@@ -387,3 +388,245 @@ func (tp *TransactionPool) PurgeTransactionPool() {
...
+				removedObjects[id] = struct{}{}
+				removalStack = append(removalStack, id)
+			}
+		}
+
+		// TODO: Make sure all outputs are accounted for properly in file
+		// contract related fields. Do `NewValidProofOutputs` and
+		// `NewMissedProofOutputs` fields of a FileContractRevision need to be
+		// added to the removal stack as well?
+		for i := range txn.FileContracts {
```
marcinja: This `TODO` needs to be addressed. I don't know how the `SiacoinOutputs` of a `FileContract` should be treated here.

modules/transactionpool/update.go
```diff
@@ -387,3 +388,245 @@ func (tp *TransactionPool) PurgeTransactionPool() {
...
+		tp.transactionSetDiffs[TransactionSetID(unconfirmedSet.ID)] = cc
+	}
+
+	// Delete all removed objects from tpool state.
+	for object, _ := range removedObjects {
+		delete(tp.knownObjects, object)
+	}
+
+	// Update subscribers on changes that have been made.
+	for _, subscriber := range tp.subscribers {
```
marcinja: It might be better to return a diff to whatever method calls this one (i.e. `ProcessConsensusChange`), instead of handing it off to subscribers here.

tbenz9:
@marcinja I'm a little confused about the purpose of this PR.  Since the devs haven't commented in 3 months I'd guess they aren't interested in this PR in it's current form.  If you're proposing a solution it's usually best practice to create a GitHub issue to discuss a possible solution before opening a PR.  I would recommend we close this PR (keep the branch), and continue the discussion in a GitHub Issue.

DavidVorick:
This is a PR that I commissioned while @marcinja was interning for us. We haven't given it enough attention because the renter really needs work more than the transaction pool.

The removeTransaction function will be helpful during reorgs and will also help us prepare to add replaceByFee, both of which will be necessary as Sia gets more on-chain transaction volume.

Right now, the on-chain transaction volume is low enough that this is low priority. If we see a big uptick again though, we'll be coming back to this pretty quickly.

PR #2254, opened 2017-08-23
remove remoteAddr check

lukechampine:
This fixes https://github.com/NebulousLabs/Sia/pull/2066#issuecomment-320294721

This shouldn't affect the threat model. Dropping this check makes it easier for a peer to make you ping an IP other than their own, but that was already possible via other means.

c.f. https://github.com/NebulousLabs/Sia/issues/1970


starius:
This check seems to prevent connecting nodes from Tor, because Tor may choose different exit nodes for http://myexternalip.com/raw and for the Sia node. Also it should be documented that Sia sends IP determined by UPnP to external world as it can deanonymize Tor users. Also Sia nodes can use each other to discover self IPs instead of depending on a 3d party service (myexternalip.com).

PR #2263, opened 2017-08-26
fix for repair loop deadlock

ChrisSchinnerl:


modules/renter/renter.go
```diff
@@ -161,6 +161,9 @@ type Renter struct {
 	mu             *sync.RWMutex
 	tg             *sync.ThreadGroup
 	tpool          modules.TransactionPool
+
+	// Testing utilities
+	debugFlags map[string]bool
```
DavidVorick: As mentioned in the WAL, this pattern should be replaced with an explicit set of fields/bools that have names and comments explaining what the flags do. That makes it a lot easier to get oriented when you are first reading the codebase.

modules/renter.go
```diff
@@ -314,6 +314,9 @@ type Renter interface {
 	// SetSettings sets the Renter's settings.
 	SetSettings(RenterSettings) error
 
+	// SetDebugFlag sets a debug flag to induce a certain behaviour in the renter
+	SetDebugFlag(flag string)
```
DavidVorick: This is just something you use in testing, right? Instead of having it be a function of the renter, we probably want to do dependency injection. When you create the renter, you can pass in a dependency like the 'disrupt()' dependency that we have in the contract manager, which will cause the code to behave in strategic ways for certain tests.

This function though is cluttering our specification, and also makes it a lot harder to re-implement the renter for someone who wants to take a stab at re-writing the module in a way that is compatible with the API.

modules/renter/repair.go
```diff
@@ -201,6 +208,14 @@ func (r *Renter) managedRepairIteration(rs *repairState) {
 		}
 	}
 
+	// Accept work non-blocking if there is any
```
DavidVorick: This code is not non-blocking, but I don't think we want it to be non-blocking either. The code looks good just the comment is incorrect.

modules/renter.go
ChrisSchinnerl: Yes, it is. I wanted to create a working testcase with minimal changes + a fix before implementing an injection system like the one in the contract manager. This would also replace the debug flags with a dependency object. Same thing for the wal of course.

modules/renter/repair.go
```diff
@@ -253,14 +268,31 @@ func (r *Renter) managedRepairIteration(rs *repairState) {
...
+		dc, downloading := rs.downloadingChunks[chunkID]
+		if downloading {
+			// Check if download finished
+			select {
+			default:
+				// Check if download timed out
+				if time.Now().After(dc.startTime.Add(chunkDownloadTimeout)) {
+					r.log.Println("Download of chunk to repair timed out")
+					chunksToDelete = append(chunksToDelete, chunkID)
+					delete(rs.downloadingChunks, chunkID)
```
DavidVorick: Do we need to fail the download here, or can we safely assume that the download was failed when the chunk failed?

modules/renter/repair.go
```diff
@@ -253,14 +268,31 @@ func (r *Renter) managedRepairIteration(rs *repairState) {
...
+				}
+				continue
+			case <-dc.d.downloadFinished:
+				// Download finished, continue repairing the chunk.
+			}
 		}
-		// Update the number of gaps for this chunk.
+
+		// Don't update gaps for downloaded chunks twice
```
DavidVorick: Where do the download gaps get updated? It's not a big deal if the code is working, because we're scrapping it anyway, but generally speaking we should have everything happening all in one place so that it's easier to keep track of in the future.

But again, this code is all going into a dumpster in a few weeks anyway, we've got a much better way of doing this. So, we can relax a bit here and just be happy that it's not buggy.

ChrisSchinnerl: Right below. They are still updated but the comment was moved inside an if statement.

modules/renter/repair.go
ChrisSchinnerl: As far as I can tell it doesn't make a difference. Letting it fail will only set a boolean value and close the channel. Since we also delete it from `rs.downloadingChunks` nothing is holding a reference to that download anyway. I am still going to let it fail for completeness though.

Issue #2260, opened 2017-08-25
Host: randomness tests for the stored data


starius:
Attackers can upload non-encrypted illegal files to hosts, which can be dangerous for host operators from legal perspective. (The uploaded data can be forbidden to be stored in the country.) Even if the host operator is clean as a whistle, law enforcement people often lack education in this area and can still seize the hardware and arrest the operator (this [happened](http://web.archive.org/web/20141004142101/http://raided4tor.cryto.net/) with Tor operators).

Hosts can reduce this risk by doing randomness tests of the uploaded data and reject if it seems to be non-encrypted. We can use the fact that the encrypted data can't be distinguished from random data reliably. As a starting point we can check that `net/http.DetectContentType` returns `"application/octet-stream"` for the data. Also we can measure Shannon Entropy and reject data if the entropy value is low.

lukechampine:
Here's a simple function that returns the randomness (as a value between 0 and 1) of a given `[]byte`:
```go
func ShannonEntropy(data []byte) float64 {
	frq := make(map[byte]int)
	for _, i := range data {
		frq[i]++
	}

	var sum float64
	for _, v := range frq {
		f := float64(v) / float64(len(data))
		sum -= f * math.Log2(f)
	}

	return sum / 8
}
```
It returns 0 when `data` is a single repeated byte, and 0.99999 when `data` 4MB of `crypto/rand`. So 0.999 seems like a reasonable cut-off. If the cut-off is too low, a malicious renter could store a bit of plaintext alongside a lot of randomness. In that case, maybe a "sliding window" approach makes more sense?

starius:
I tried this function with a video file of format:
RIFF (little-endian) data, AVI, 720 x 400, 23.98 fps, video: XviD, audio: MPEG-1 Layer 3 (stereo, 48000 Hz)
In the middle of a file, the function returned 0.99918 for 1 megabyte of data.
Also, `DetectContentType` returned `"application/octet-stream"`.

lukechampine:
Hmm. I guess in practice it will be difficult (theoretically impossible) to distinguish between data that's encrypted and data that's compressed. A sliding window still might work, since even a compressed format will have a low-entropy header or footer.

lukechampine:
won't a zlib of a zlib-compressed (but unencrypted) file be larger than the original?
maybe we could check for the magic header bytes for a number of formats. But then you run the risk of false positives...

petabytestorage:
[...]

PR #2264, opened 2017-08-27
host: reject low entropy data

starius:
fix https://github.com/NebulousLabs/Sia/issues/2260

### DIFF
--- a/modules/host/negotiaterevisecontract.go
+++ b/modules/host/negotiaterevisecontract.go
@@ -2,6 +2,7 @@ package host
 
 import (
 	"fmt"
+	"math"
 	"net"
 	"time"
 
@@ -92,6 +93,9 @@ func (h *Host) managedRevisionIteration(conn net.Conn, so *storageObligation, fi
 				if uint64(len(modification.Data)) != modules.SectorSize {
 					return errBadSectorSize
 				}
+				if err := randomnessTest(modification.Data); err != nil {
+					return err
+				}
 
 				// Update finances.
 				blocksRemaining := so.proofDeadline() - blockHeight
@@ -120,6 +124,10 @@ func (h *Host) managedRevisionIteration(conn net.Conn, so *storageObligation, fi
 					return extendErr("could not read sector: ", ErrorInternal(err.Error()))
 				}
 				copy(sector[modification.Offset:], modification.Data)
+
+				if err := randomnessTest(sector); err != nil {
+					return err
+				}
 
 				// Update finances.
 				bandwidthRevenue = bandwidthRevenue.Add(settings.MinUploadBandwidthPrice.Mul64(uint64(len(modification.Data))))
@@ -322,5 +330,30 @@ func verifyRevision(so storageObligation, revision types.FileContractRevision, b
 		return errBadFileMerkleRoot
 	}
 
+	return nil
+}
+
+// shannonEntropy returns Shannon entropy of the data.
+func shannonEntropy(data []byte) float64 {
+	var frq [256]int
+	for _, v := range data {
+		frq[v]++
+	}
+	var sum float64
+	den := float64(len(data))
+	for _, v := range frq {
+		if v > 0 {
+			f := float64(v) / den
+			sum -= f * math.Log2(f)
+		}
+	}
+	return sum / 8
+}
+
+// randomnessTest checks that the data looks random.
+func randomnessTest(data []byte) error {
+	if shannonEntropy(data) < modules.MinimumEntropy {
+		return errLowEntropy
+	}
 	return nil
 }
--- a/modules/host/negotiaterevisecontract_test.go
+++ b/modules/host/negotiaterevisecontract_test.go
@@ -1,0 +1,88 @@
+package host
+
+import (
+	"testing"
+)
+
+func TestShannonEntropy(t *testing.T) {
+	t.Parallel()
+	uniform := make([]byte, 256)
+	for i := 0; i < 256; i++ {
+		uniform[i] = byte(i)
+	}
+	almost := make([]byte, 256)
+	for i := 0; i < 256-10; i++ {
+		almost[i] = byte(i)
+	}
+	testCases := []struct {
+		name      string
+		input     []byte
+		low, high float64
+	}{
+		{
+			name:  "all zeros",
+			input: make([]byte, 100),
+			low:   0.0,
+			high:  0.00000001,
+		},
+		{
+			name:  "uniform distrinution",
+			input: uniform,
+			low:   0.9999999,
+			high:  1.0,
+		},
+		{
+			name:  "almost uniform distrinution",
+			input: almost,
+			low:   0.98,
+			high:  0.99,
+		},
+	}
+	for _, ts := range testCases {
+		got := shannonEntropy(ts.input)
+		if got < ts.low || got > ts.high {
+			t.Errorf("shannonEntropy(%s): %#v, want between %#v and %#v", ts.name, got, ts.low, ts.high)
+		}
+	}
+}
+
+func TestRandomnessTest(t *testing.T) {
+	t.Parallel()
+	uniform := make([]byte, 256)
+	for i := 0; i < 256; i++ {
+		uniform[i] = byte(i)
+	}
+	almost := make([]byte, 256)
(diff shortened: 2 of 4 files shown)


modules/host/negotiaterevisecontract.go
```diff
@@ -92,6 +93,9 @@ func (h *Host) managedRevisionIteration(conn net.Conn, so *storageObligation, fi
 				if uint64(len(modification.Data)) != modules.SectorSize {
 					return errBadSectorSize
 				}
+				if err := randomnessTest(modification.Data); err != nil {
+					return err
+				}
```
lukechampine: I think inlining this function is clearer:
```go
if shannonEntry(modification.Data) < modules.MinimumEntropy {
    return errLowEntropy
}
```

modules/host/negotiaterevisecontract.go
```diff
@@ -121,6 +125,10 @@ func (h *Host) managedRevisionIteration(conn net.Conn, so *storageObligation, fi
 				}
 				copy(sector[modification.Offset:], modification.Data)
 
+				if err := randomnessTest(sector); err != nil {
+					return err
+				}
```
lukechampine: ditto

modules/host/negotiaterevisecontract.go
```diff
@@ -324,3 +332,28 @@ func verifyRevision(so storageObligation, revision types.FileContractRevision, b
...
+	return sum / 8
+}
+
+// randomnessTest checks that the data looks random.
+func randomnessTest(data []byte) error {
+	if shannonEntropy(data) < modules.MinimumEntropy {
+		return errLowEntropy
+	}
+	return nil
+}
```
lukechampine: this function can be removed

modules/host/negotiaterevisecontract_test.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+	for _, ts := range testCases {
+		got := shannonEntropy(ts.input)
+		if got < ts.low || got > ts.high {
+			t.Errorf("shannonEntropy(%s): %#v, want between %#v and %#v", ts.name, got, ts.low, ts.high)
+		}
+	}
+}
+
+func TestRandomnessTest(t *testing.T) {
```
lukechampine: don't need this test

modules/host/negotiaterevisecontract_test.go
```diff
@@ -0,0 +1,88 @@
...
+			input: uniform,
+			low:   0.9999999,
+			high:  1.0,
+		},
+		{
+			name:  "almost uniform distrinution",
+			input: almost,
+			low:   0.98,
+			high:  0.99,
+		},
```
lukechampine: should also test random data (using `fastrand.Bytes()`

modules/host/negotiaterevisecontract_test.go
```diff
@@ -0,0 +1,88 @@
...
+		uniform[i] = byte(i)
+	}
+	almost := make([]byte, 256)
+	for i := 0; i < 256-10; i++ {
+		almost[i] = byte(i)
+	}
+	testCases := []struct {
+		name      string
+		input     []byte
+		low, high float64
```
lukechampine: might be better to use `exp, delta float64` and then test with:
```go
if math.Abs(got - ts.exp) > ts.delta {
    // fail
}
```

modules/host/negotiaterevisecontract.go
starius: This is reserved for the future: if we decide to add more randomness tests or change existing ones, we will change `randomnessTest` function instead of modifying the inlined implementation in two places (there is a risk that we update only one place and forget to update another).

modules/host/negotiaterevisecontract_test.go
```diff
@@ -0,0 +1,88 @@
...
+			name:  "all zeros",
+			input: make([]byte, 100),
+			low:   0.0,
+			high:  0.00000001,
+		},
+		{
+			name:  "uniform distrinution",
+			input: uniform,
+			low:   0.9999999,
+			high:  1.0,
```
lukechampine: it's a little surprising that 1,2,3... would yield such a high result, but I guess it makes sense under the definition of Shannon entropy. Still, makes me wonder if a compression test is warranted.

modules/host/negotiaterevisecontract_test.go
starius: Added, but `math/rand` was used instead of `fastrand` to make the input the same in all test invocations (and the test - reproducible).

modules/host/negotiaterevisecontract_test.go
starius: Done. Also delta is the same for all the test cases.

modules/host/negotiaterevisecontract_test.go
starius: gzip extends the length of {0, 1, 2, ..., 256} from 256 bytes to 284 bytes even with the best compression level: https://play.golang.org/p/uJSjjQTepx (Probably gzip's algorithm does not recognize this kind of pattern.)

I think we do not need to catch this sequence to fix #2260 because media files (which we hunt for) are unlikely to have this sequence.

modules/host/negotiaterevisecontract.go
starius: For example, if we decide to add compression test, this function will be a good place to put it.

modules/host/negotiaterevisecontract_test.go
lukechampine: interesting, thanks for running the test. In that case, I think Shannon entropy is good enough for now.

DavidVorick:
I'm not sure this is something we want. There are plenty of legitimate reasons to store unencrypted data on Sia, for example if you are running a CDN that's targeting low power devices, you may not want those devices doing any decryption, as that's going to suck up their limited CPU and battery and harm user experience.

And that's not going to stop attackers from uploading harmful data anyway. All you have to do is share the encryption key, and then you are back to where you started.

This pull request harms user experience in pursuit of stopping a type of upload that is information theoretically impossible to stop.

lukechampine:
I agree in principle, but before rejecting this I'd like to have at least a general sense of:

- Is there any legal distinction (at least in the U.S. and E.U.) between storing unencrypted data vs. encrypted data?
- Given the choice, would hosts opt-in to rejecting unencrypted data?

rudibs:
I doubt it matters, but on zlib: a) it's probably more accurate and stable
than the 0-order shannon entropy based measures
b) to the contrary, the fixed small ratio expansion of gzip/zlib is
precisely due to the fact that it has a "failsafe" mode that detects random
data. so this is a solved problem for decades now.

starius:
I have very little knowledge on the laws. U.S. Code § 230 - "Protection for private blocking and screening of offensive material" has a section about "Good Samaritan" (a provider that attempts to block offensive content). Can it be applicable here?

Can we make this a host option: whether or not it accepts unencrypted data? Then host operators can select a proper value based on jurisdiction and personal preferences.

rudibs:
I disagree on the "mathematical impossibility" point.  Google allows people to search for porn. Google makes a legally significant attempt to prevent youngsters from accidentally being exposed to porn. The standard technique for "protecting against youngsters" is not super secure (turn off safe search) but is legally all we expect because the problem is in practical terms impossible to solve for reasons of liberty and freedom / authoritarianism and enforcement not simple logic. So I believe if an attacker uploads an encrypted porn video then published the key it is still a lot better than something where you do not need to put two pieces together intentionally. This level of discouragement is normally all we can expect.

modules/host/negotiaterevisecontract.go
petabytestorage: It makes more sense to provide the compressed size in bits not a ratio.  Also the size before compression in bits. Ratio does not provide the same amount of information by itself.

starius:
I changed my mind on this approach after reading [this](https://trac.torproject.org/projects/tor/ticket/23270). In particular the following argument from @nmathewson:

> It has potential, in some jurisdictions, to open relay operators to legal liability for things that they choose _not_ to censor. (IANAL, TINLA) 

This argument is applicable to this PR.
I think the PR will bring more harm than profit.

PR #2270, opened 2017-08-29
fix cleanCloseHandler memory leak

lukechampine:


DavidVorick:
did not realize this would cause issues

PR #2273, opened 2017-08-30
renew contracts early if their funds are exhausted

avahowell:
This PR updates the contractor so that it will renew contracts prematurely (before their end height) if they are about to run out of funds. Previously, the contracts would be useless until the next renew period.

### DIFF
--- a/api/renter_test.go
+++ b/api/renter_test.go
@@ -1989,11 +1989,12 @@ func TestExhaustedContracts(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	// verify that the window did not change
-	newWindowStart := st.renter.Contracts()[0].LastRevision.NewWindowStart
-	newWindowEnd := st.renter.Contracts()[0].LastRevision.NewWindowEnd
-	startHeight := st.renter.Contracts()[0].StartHeight
-	endHeight := st.renter.Contracts()[0].EndHeight()
+	// verify that the window did not change and the contract ID did change
+	newContract := st.renter.Contracts()[0]
+	newWindowStart := newContract.LastRevision.NewWindowStart
+	newWindowEnd := newContract.LastRevision.NewWindowEnd
+	startHeight := newContract.StartHeight
+	endHeight := newContract.EndHeight()
 	if startHeight != initialContract.StartHeight {
 		t.Fatal("contract start height changed, wanted", initialContract.StartHeight, "got", startHeight)
 	}
@@ -2005,6 +2006,9 @@ func TestExhaustedContracts(t *testing.T) {
 	}
 	if newWindowStart != initialContract.LastRevision.NewWindowStart {
 		t.Fatal("contract windowStart changed, wanted", initialContract.LastRevision.NewWindowStart, "got", newWindowStart)
+	}
+	if newContract.ID == initialContract.ID {
+		t.Fatal("renew did not occur")
 	}
 }
 
--- a/api/renter_test.go
+++ b/api/renter_test.go
@@ -1972,7 +1972,7 @@ func TestExhaustedContracts(t *testing.T) {
 	uploadValues.Set("renew", "true")
 	uploadValues.Set("datapieces", "1")
 	uploadValues.Set("paritypieces", "1")
-	err = st.stdPostAPI("/renter/upload"+tmpfile.Name(), uploadValues)
+	err = st.stdPostAPI("/renter/upload/"+filepath.Base(tmpfile.Name()), uploadValues)
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -2052,7 +2052,7 @@ func TestAdversarialPriceRenewal(t *testing.T) {
 	uploadValues.Set("renew", "true")
 	uploadValues.Set("datapieces", "1")
 	uploadValues.Set("paritypieces", "1")
-	err = st.stdPostAPI("/renter/upload"+tmpfile.Name(), uploadValues)
+	err = st.stdPostAPI("/renter/upload/"+filepath.Base(tmpfile.Name()), uploadValues)
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/api/renter_test.go
+++ b/api/renter_test.go
@@ -1923,8 +1923,11 @@ func TestExhaustedContracts(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	// Increase the storage and bandwidth prices to drain the allowance faster.
 	settings := st.host.InternalSettings()
-	settings.MinUploadBandwidthPrice = types.SiacoinPrecision.Div64(200)
+	settings.MinUploadBandwidthPrice = settings.MinUploadBandwidthPrice.Mul64(2)
+	settings.MinStoragePrice = settings.MinUploadBandwidthPrice.Mul64(2)
+	settings.MaxDuration = 1e6 // set a high max duration to allow an expensive storage price
 	err = st.host.SetInternalSettings(settings)
 	if err != nil {
 		t.Fatal(err)
@@ -1936,8 +1939,8 @@ func TestExhaustedContracts(t *testing.T) {
 
 	// Set an allowance for the renter, allowing a contract to be formed.
 	allowanceValues := url.Values{}
-	testPeriod := "1000"
-	allowanceValues.Set("funds", types.SiacoinPrecision.Mul64(2000).String())
+	testPeriod := "950000" // large period to cause an expensive test, exhausting the allowance faster
+	allowanceValues.Set("funds", types.SiacoinPrecision.Mul64(5).String())
 	allowanceValues.Set("period", testPeriod)
 	err = st.stdPostAPI("/renter", allowanceValues)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


modules/renter/contractor/consts.go
```diff
@@ -54,6 +54,11 @@ var (
 		Standard: 10,
 		Testing:  1,
 	}).(int)
+
+	// minContractFundRenewalThreshold defines the ratio of remaining funds to
+	// total contract cost below which the contractor will prematurely renew a
+	// contract.
+	minContractFundRenewalThreshold = float64(0.10) // 10%
```
DavidVorick: 3% is better. 10% is a wide margin.

modules/renter/contractor/consts.go
```diff
@@ -62,7 +67,11 @@ var (
 	maxCollateral    = types.SiacoinPrecision.Mul64(1e3) // 1k SC
 	maxDownloadPrice = maxStoragePrice.Mul64(3 * 4320)
 	maxStoragePrice  = types.SiacoinPrecision.Mul64(30e3).Div(modules.BlockBytesPerMonthTerabyte) // 30k SC / TB / Month
-	maxUploadPrice   = maxStoragePrice.Mul64(4320)
+	maxUploadPrice   = build.Select(build.Var{
+		Standard: maxStoragePrice.Mul64(4320),
+		Dev:      maxStoragePrice.Mul64(4320),
+		Testing:  types.SiacoinPrecision,
```
DavidVorick: It's hard to tell by reading this whether the testing max price is much smaller or much larger, it should probably match the others but be 4320e6 or something

api/renter_test.go
```diff
@@ -1899,3 +1899,212 @@ func TestContractorHostRemoval(t *testing.T) {
...
+	testPeriod := "1000"
+	allowanceValues.Set("funds", types.SiacoinPrecision.Mul64(1000).String())
+	allowanceValues.Set("period", testPeriod)
+	err = st.stdPostAPI("/renter", allowanceValues)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// wait until we have a contract
+	err = retry(500, time.Millisecond*50, func() error {
```
DavidVorick: phasing this out for `build.Retry`

DavidVorick: it's the same function just accessible to everyone instead of silo'd to the api package

api/renter_test.go
```diff
@@ -1899,3 +1899,212 @@ func TestContractorHostRemoval(t *testing.T) {
...
+	initialContract := st.renter.Contracts()[0]
+
+	// upload a file. the high upload cost will cause the underlying contract to
+	// require premature renewal. If premature renewal never happens, the upload
+	// will never complete.
+	tmpfile, err := ioutil.TempFile("", t.Name())
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.Remove(tmpfile.Name())
```
DavidVorick: The testing environment wipes all persisted files every time a test cycle runs, no need to use a temp file. I believe we have a helper function `createRandFile` that you can use here.

api/renter_test.go
```diff
@@ -1899,3 +1899,212 @@ func TestContractorHostRemoval(t *testing.T) {
...
+	_, err = io.CopyN(tmpfile, fastrand.Reader, int64(modules.SectorSize))
```
DavidVorick: I'm not sure if I'm missing something, but I think you would want several sectors, where the contract has enough funds to upload a few sectors, but not all of them. Right now it looks like it just uploads one sector, and I'm not sure how an upload could get stuck halfway through if there's only one host and one sector.

api/renter_test.go
avahowell: Ah yeah, glad we have that merged

api/renter_test.go
avahowell: This test sets the upload bandwidth cost very high, high enough so that the contract gets exhausted part-way through uploading just one sector.

api/renter_test.go
```diff
@@ -1899,3 +1899,212 @@ func TestContractorHostRemoval(t *testing.T) {
...
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// verify that the window did not change
+	newWindowStart := st.renter.Contracts()[0].LastRevision.NewWindowStart
+	newWindowEnd := st.renter.Contracts()[0].LastRevision.NewWindowEnd
+	startHeight := st.renter.Contracts()[0].StartHeight
+	endHeight := st.renter.Contracts()[0].EndHeight()
```
DavidVorick: you should also verify that the file contract id changed. I'm not sure actually if the api exposes that, but the test should have some way to know whether or not an actual renewal happened (this validates the test)

api/renter_test.go
```diff
@@ -1899,3 +1899,212 @@ func TestContractorHostRemoval(t *testing.T) {
...
+		t.Fatal("contract windowEnd changed, wanted", initialContract.LastRevision.NewWindowEnd, "got", newWindowEnd)
+	}
+	if newWindowStart != initialContract.LastRevision.NewWindowStart {
+		t.Fatal("contract windowStart changed, wanted", initialContract.LastRevision.NewWindowStart, "got", newWindowStart)
+	}
+}
+
+// TestAdversarialPriceRenewal verifies that host cannot maliciously raise
+// their storage price in order to trigger a premature file contract renewal.
+func TestAdversarialPriceRenewal(t *testing.T) {
```
DavidVorick: This is a good test, I am glad to see it.

modules/renter/contractor/contracts.go
```diff
@@ -343,20 +341,38 @@ func (c *Contractor) threadedContractMaintenance() {
...
+				blockBytes := types.NewCurrency64(modules.SectorSize * uint64(contract.EndHeight()-c.blockHeight))
+				sectorStoragePrice := host.StoragePrice.Mul(blockBytes)
+				sectorBandwidthPrice := host.UploadBandwidthPrice.Mul64(modules.SectorSize)
+				sectorPrice := sectorStoragePrice.Add(sectorBandwidthPrice)
+				percentRemaining, _ := big.NewRat(0, 1).SetFrac(contract.RenterFunds().Big(), contract.TotalCost.Big()).Float64()
+				if contract.RenterFunds().Cmp(sectorPrice) < 0 || percentRemaining < minContractFundRenewalThreshold {
+					renewSet = append(renewSet, contract.ID)
+					refundSet[contract.ID] = struct{}{}
+				}
+			}
```
DavidVorick: This block has lots of heavily indented code. You can probably clean things up with different control flow. For example:

```go
if !contract.GoodForRenew {
    continue
}
if c.height+c.a.window > ctc.endHeight {
    append()
    continue
}

// now check for funds exhaustion
```

modules/renter/contractor/contracts.go
```diff
@@ -343,20 +341,38 @@ func (c *Contractor) threadedContractMaintenance() {
...
+				renewSet = append(renewSet, contract.ID)
+			} else {
+				// check if the contract has exhausted its funding and requires premature renewal.
+				c.mu.RUnlock()
+				host, _ := c.hdb.Host(contract.HostPublicKey)
+				c.mu.RLock()
+				if host.StoragePrice.Cmp(maxStoragePrice) > 0 || host.UploadBandwidthPrice.Cmp(maxUploadPrice) > 0 {
+					// skip this host if its prices are too high
+					continue
+				}
```
DavidVorick: It doesn't hurt to have this code here, but I believe that by the time this loop is running, another loop previously has already cycled through all of the contracts and marked the ones that no longer have acceptable prices.

modules/renter/contractor/consts.go
```diff
@@ -62,7 +67,11 @@ var (
 	maxCollateral    = types.SiacoinPrecision.Mul64(1e3) // 1k SC
 	maxDownloadPrice = maxStoragePrice.Mul64(3 * 4320)
 	maxStoragePrice  = types.SiacoinPrecision.Mul64(30e3).Div(modules.BlockBytesPerMonthTerabyte) // 30k SC / TB / Month
-	maxUploadPrice   = maxStoragePrice.Mul64(4320)
+	maxUploadPrice   = build.Select(build.Var{
+		Standard: maxStoragePrice.Mul64(4320),
+		Dev:      maxStoragePrice.Mul64(4320),
+		Testing:  maxStoragePrice.Mul64(4320e8),
```
DavidVorick: I feel like the better approach is to set a smaller allowance during testing, rather than allow a very high upload price to be set as part of our testing constants.

modules/renter/contractor/contracts.go
```diff
@@ -265,15 +265,13 @@ func (c *Contractor) managedRenew(contract modules.RenterContract, numSectors ui
 	}
 
 	// create contract params
-	c.mu.RLock()
 	params := proto.ContractParams{
 		Host:          host,
 		Filesize:      numSectors * modules.SectorSize,
-		StartHeight:   c.blockHeight,
+		StartHeight:   newStartHeight,
```
DavidVorick: any particular reason for changing the start height to be different from the current block height? I think that the old method of using the block height was okay. Unless I'm remembering these fields wrong, 'StartHeight' is supposed to indicate the block that was used to form the contract.

modules/renter/contractor/contracts.go
```diff
@@ -343,20 +341,41 @@ func (c *Contractor) threadedContractMaintenance() {
 	// Renew any contracts that need to be renewed.
 	c.mu.RLock()
 	var renewSet []types.FileContractID
+	refundSet := make(map[types.FileContractID]struct{})
```
DavidVorick: Can you add a comment here explaining the purpose of `refundSet`? It's not immediately clear.

modules/renter/contractor/contractor_test.go
```diff
@@ -198,6 +198,114 @@ func (stubHostDB) ScoreBreakdown(modules.HostDBEntry) modules.HostScoreBreakdown
 	return modules.HostScoreBreakdown{}
 }
 
+// TestAllowanceOverspend verifies that the contractor will not spend more
+// than the allowance if contracts need to be renewed early.
+func TestAllowanceOverspend(t *testing.T) {
+	t.Skip("FIXME: test requires better money management to pass")
```
lukechampine: is the idea that this will be un-skipped after a future PR is merged?

DavidVorick: Yeah, I think right now we don't actually have controls to prevent the allowance from over spending. That'll be my next PR after this is merged.

Issue #2266, opened 2017-08-28
Proposal: Wallet Contexts


avahowell:
This proposal outlines a change to the wallet to help make the wallet's behavior easier to understand, and to more easily facilitate certain behaviors that depend on the wallet. 

## Motivation

Users want to easily understand every transaction that occurs in their wallet. In Sia, there are more reasons a transaction can occur than other cryptocurrencies. Users can receive and send Siacoin of their own volition, which is currently relatively easy to understand, but the software also transacts on behalf of users in order to form contracts and transfer data.

Currently, these non-user-initiated transactions appear exactly the same to users as user-initiated transactions. This is confusing behavior: "why is Sia spending my money?", "What are all these small transactions?" are common sentiments expressed on reddit, the forum, and elsewhere.  

In addition, there is a use case for establishing 'sub-wallets' or contextual wallets outside of being able to tag transactions related to a certain behavior. An example use case for this would be server operators wishing to serve many wallet users from a single application running a single Sia node. This isn't currently possible, but would be with the proposed changes.

A major motivation for this change is the Allowance UX improvement proposal:
https://github.com/NebulousLabs/Sia/issues/2246. This proposal outlines a change to the wallet which would easily facilitate that behavior.

## High Level Overview

Each transaction in the wallet should be tagged with metadata. This metadata should indicate the `context` from which the transaction originated. Every transaction in the wallet should have an associated `context`, in effect creating contextualized 'sub-wallets' within the wallet module. Every wallet API call should be updated to take a `context` parameter. `Contexts` have a configurable balance, and if a transaction would result in lower than the configured limit the wallet will return an `insufficient balance` error.

#### Wallet API calls with context:

`GET /wallet/transactions` -> returns every transaction in the wallet.  
`GET /wallet/transactions?context="contractor"` -> returns every transaction in the wallet's `contractor` context.  

`POST /wallet/siacoins` -> sends siacoins.  
`POST /wallet/siacoins?context="contractor"` -> sends siacoins, from the wallet's `contractor` context.  

Other existing wallet API calls can be updated similarly to support contexts. Two new API calls are hereby proposed:

`GET /wallet/contexts` -> list every context currently tracked by the wallet, and their balances.
`POST /wallet/contexts` -> add or update an existing context's spending limit.

One field should be added to the `modules.ProcessedTransaction` type, `Context`. The new `ProcessedTransaction` type therefore has the following signature:

```go
ProcessedTransaction struct {
[...]

Issue #2246, opened 2017-08-19
Proposal: Allowance UX Improvement


avahowell:
# Motivation

The behavior of Sia's "allowance" abstraction, the mechanism by which the core use case of Sia is exposed, is currently very hard to understand from a user's perspective. The main point of confusion is the manner in which money is deducted from the wallet to form contracts: instead of a single transaction exactly the size of the allowance set appearing as a deduction in their transaction list, they see dozens/hundreds of small deductions as the contractor forms contracts. In addition, there are cases where the contractor can inadvertently spend more than the allotted amount. This proposal outlines a simple change to the allowance mechanism that will greatly improve user experience. 

# Goals

- The entire allowance amount should be deducted once from the wallet in a single transaction that exactly matches the amount the user specified, once per renew period.
- Contractor should never spend more than the allowance allows for in the given period.
- Contractor should accurately report upload, download, and storage spending for the current period. 

# Solution

The contractor currently interfaces with the user's wallet module, in order to build the transactions required for contract formation, renewal, upload, and download. As previously mentioned, this leads to undesirable user experience: it's not clear when and where the allowance is deducted from the user's wallet, or which transactions represent which actions.

The contractor should embed a secondary wallet module. When a user sets an allowance, a transaction will be created that sends the exact sum of the allowance from the user's wallet to the contractor subwallet, and the contractor will use this subwallet to create the transactions used to form and renew contracts. The allowance amount should be sent from the main wallet to the subwallet once per allowance period. If a period elapses and there are unspent allowance funds, the unspent amount should be sent back to the main wallet. In addition, the subwallet's seed should be deterministically generated from the main wallet's seed.

The API should expose endpoints that give an accurate overview of how the money allotted to the contractor is being spent in that period. Sia-UI will implement a segmented progress bar which displays clearly how the amount allocated to the contractor is being spent.

# Considerations

Currently there are performance concerns surrounding running another wallet module that must sync with the consensus set. Some of these may be mitigated by omitting the initial rescan for the contractor wallet where appropriate. A small performance penalty may be acceptable for significant UX gains, and focusing some effort on a more performant wallet is desirable regardless.

tlightsky:
it would be better that it gives more exact how much storage size if,
i even don't know how much space i rent after the allowance setup

eddiewang:
Having a second wallet module greatly increases the complexity of the application, although I could see why having a wallet used only by the contractor could be an interesting security feature, on top of the obvious UX improvements.

[...]

PR #2278, opened 2017-08-30
add wallet context support

avahowell:
This PR adds support in the wallet for contexts, as described in #2266. This lays the groundwork for the UX improvements outlined in #2246.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -122,7 +122,6 @@ func New(requiredUserAgent string, requiredPassword string, cs modules.Consensus
 
 	// Register API handlers
 	api.buildHttpRoutes(requiredUserAgent, requiredPassword)
-
 	return api
 }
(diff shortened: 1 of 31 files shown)


lukechampine:
do we need api support for this yet? I think I'd prefer to leave that until later. Also, I think it's fine if `SendSiacoins` always uses the default context. If you want to use a custom context, use a transactionbuilder.

DavidVorick:
Part of the purpose of writing proposals is to get reviews on the high-level ideas and design before writing the code, that way if substantial changes are made, you don't have to redo the coding bits

lukechampine:
ah yeah, I should have commented on the proposal earlier

avahowell:
Rebased and resolved merge conflicts.

PR #2280, opened 2017-08-31
close file handle in DownloadFileWriter

### LINKED ISSUE #2279: siad doesn't close file properly after download
Using Sia-UI 1.3 on Windows 10.

After downloading a file, it is still held open by siad after it's fully downloaded.

Can reproduce by just downloading a file with Sia-UI or curl and then use Process explorer to view open files (or just try to open the file in another process (read only is fine, but no write access)).

avahowell:
Fixes #2279. Also fixes a case where we didn't check the error when opening a file handle for a download.

### DIFF
--- a/modules/renter/download.go
+++ b/modules/renter/download.go
@@ -680,16 +680,23 @@ type DownloadFileWriter struct {
 	f        *os.File
 	location string
 	offset   uint64
+	written  uint64
+	length   uint64
 }
 
 // NewDownloadFileWriter creates a new instance of a DownloadWriter backed by the file named.
-func NewDownloadFileWriter(fname string, offset, length uint64) *DownloadFileWriter {
-	l, _ := os.OpenFile(fname, os.O_CREATE|os.O_WRONLY, defaultFilePerm)
+func NewDownloadFileWriter(fname string, offset, length uint64) (*DownloadFileWriter, error) {
+	l, err := os.OpenFile(fname, os.O_CREATE|os.O_WRONLY, defaultFilePerm)
+	if err != nil {
+		return nil, err
+	}
 	return &DownloadFileWriter{
 		f:        l,
 		location: fname,
 		offset:   offset,
-	}
+		written:  0,
+		length:   length,
+	}, nil
 }
 
 // Destination implements the Location method of the DownloadWriter interface
@@ -700,7 +707,15 @@ func (dw *DownloadFileWriter) Destination() string {
 
 // WriteAt writes the passed bytes at the specified offset.
 func (dw *DownloadFileWriter) WriteAt(b []byte, off int64) (int, error) {
-	return dw.f.WriteAt(b, off-int64(dw.offset))
+	n, err := dw.f.WriteAt(b, off-int64(dw.offset))
+	if err != nil {
+		return n, err
+	}
+	dw.written += uint64(n)
+	if dw.written == dw.length {
+		return n, dw.f.Close()
+	}
+	return n, err
 }
 
 // DownloadHttpWriter is a http response writer-backed implementation of
--- a/modules/renter/download.go
+++ b/modules/renter/download.go
@@ -707,12 +707,15 @@ func (dw *DownloadFileWriter) Destination() string {
 
 // WriteAt writes the passed bytes at the specified offset.
 func (dw *DownloadFileWriter) WriteAt(b []byte, off int64) (int, error) {
+	if dw.written+uint64(len(b)) > dw.length {
+		build.Critical("DownloadFileWriter write exceeds file length")
+	}
 	n, err := dw.f.WriteAt(b, off-int64(dw.offset))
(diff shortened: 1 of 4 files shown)


modules/renter/download.go
```diff
@@ -700,7 +707,15 @@ func (dw *DownloadFileWriter) Destination() string {
...
 func (dw *DownloadFileWriter) WriteAt(b []byte, off int64) (int, error) {
-	return dw.f.WriteAt(b, off-int64(dw.offset))
+	n, err := dw.f.WriteAt(b, off-int64(dw.offset))
+	if err != nil {
+		return n, err
+	}
+	dw.written += uint64(n)
+	if dw.written == dw.length {
+		return n, dw.f.Close()
+	}
```
lukechampine: should probably be `dw.written >= dw.length`. Also, should we be truncated `b` so that we never write more than `dw.length` bytes total?

avahowell: Hmm, should we truncate, or return an error if writing `b` would exceed `length`?

lukechampine: IIUC it's a developer error, so `build.Critical` would be appropriate.

DavidVorick:
Looks like there's a test failing to due file writer out of bounds errors

avahowell:
Should be all set now.

modules/renter/download.go
```diff
@@ -700,7 +707,18 @@ func (dw *DownloadFileWriter) Destination() string {
...
+	n, err := dw.f.WriteAt(b, off-int64(dw.offset))
+	if err != nil {
+		return n, err
+	}
+	dw.written += uint64(n)
+	if dw.written >= dw.length {
+		return n, dw.f.Close()
+	}
+	return n, err
```
DavidVorick: I think that calling `Close` inside of the `WriteAt` function is the wrong move. Whatever is calling `WriteAt` should be the method calling `Close`, which would more appropriately match the other interfaces in the golang language.

DavidVorick:
The test looks good, and the added error checking looks good, but when creating these objects we should be calling `Close` on them as though they were standard files. It's really strange behavior to have a call to `WriteAt` close the file. If it's past the end of the file or there's nothing left to write, it's fine for `WriteAt` to return an error, but I think where we really want to be calling `Close` here is in the `download.fail()` function, and in the `chunkDownload.recoverChunk()` function, which as far as I can tell are the two places where a download is completed.

avahowell:
Agreed, I'll update this PR and add a `Close()` Method to the `modules.DownloadWriter` interface.

avahowell:
Ready for final review

modules/renter/download.go
```diff
@@ -217,6 +217,7 @@ func (d *download) fail(err error) {
 	d.downloadComplete = true
 	d.downloadErr = err
 	close(d.downloadFinished)
+	d.destination.Close()
```
DavidVorick: should probably be `return d.dest.Close()`

modules/renter/download.go
```diff
@@ -317,6 +318,7 @@ func (cd *chunkDownload) recoverChunk() error {
 		// Signal that the download is complete.
 		cd.download.downloadComplete = true
 		close(cd.download.downloadFinished)
+		cd.download.destination.Close()
```
DavidVorick: should error check here

DavidVorick:
PR looks good except for the unchecked errors on the close methods.

modules/renter/download.go
```diff
@@ -217,6 +217,10 @@ func (d *download) fail(err error) {
 	d.downloadComplete = true
 	d.downloadErr = err
 	close(d.downloadFinished)
+	closeErr := d.destination.Close()
+	if closeErr != nil {
+		build.Severe("error closing download destination: " + closeErr.Error())
```
DavidVorick: We'd probably just log the error. If we don't have access to a logger, we can just write a TODO or something - this code is going to get mixed up during our 15 sprints for v1.3.2. I didn't realize that we're already in the fail function. build.Severe is definitely incorrect, because an error here can happen that is not a fault of the programmer nor a reason that the user cannot continue with future actions.

Sorry about that.

modules/renter/download.go
```diff
@@ -317,6 +321,7 @@ func (cd *chunkDownload) recoverChunk() error {
 		// Signal that the download is complete.
 		cd.download.downloadComplete = true
 		close(cd.download.downloadFinished)
+		return cd.download.destination.Close()
```
DavidVorick: This is not a good code pattern, because you've got a return statement in a nested situation. If at some point in the future we were to modify this function to add things at the end, we may not realize that it's not running in this block of code. So here we'd definitely want the full `if err != nil` block, instead of just returning whatever value.

DavidVorick:
one more round

avahowell:
amended and force-pushed to address latest review comments

PR #2282, opened 2017-08-31
Reduce decoding allocations

lukechampine:
The decoder operates using the `io.Reader` abstraction, which means that if it e.g. wants to decode a length prefix, it must allocate 8 bytes, read into the byte slice, and decode the result. This is unavoidable if the underlying `io.Reader` is an `*os.File` or a `net.Conn`. However, in the majority of cases, we are decoding a byte slice that has already been allocated, as in `Unmarshal`:
```go
func Unmarshal(b []byte, v interface{}) error {
	r := bytes.NewReader(b)
	return NewDecoder(r).Decode(v)
}
```
In this case, we should be able to skip the `Read` call and work directly with the data present in `b`.

To implement this, I added a type assertion that checks if the underlying `io.Reader` is a `bytes.Buffer`. If it is, we can call the `Next` method to return the next `n` bytes without allocating. Otherwise, we fallback to the allocation-based code.

Profiling suggests that this saves a few GB of allocations per rescan. At some point we should fully implement `UnmarshalSia` for `types.Block` so that it's fast, reflection-free, and keeps allocations to a minimum.

### DIFF
--- a/encoding/marshal.go
+++ b/encoding/marshal.go
@@ -13,8 +13,8 @@ import (
 )
 
 const (
-	maxDecodeLen = 12e6 // 12 MB
-	maxSliceLen  = 5e6  // 5 MB
+	MaxDecodeSize = 12e6 // 12 MB
+	MaxSliceSize  = 5e6  // 5 MB
 )
 
 var (
@@ -185,7 +185,7 @@ type Decoder struct {
 func (d *Decoder) Read(p []byte) (int, error) {
 	n, err := d.r.Read(p)
 	// enforce an absolute maximum size limit
-	if d.n += n; d.n > maxDecodeLen {
+	if d.n += n; d.n > MaxDecodeSize {
 		panic("encoded type exceeds size limit")
 	}
 	return n, err
@@ -233,7 +233,7 @@ func (d *Decoder) readN(n int) []byte {
 		if len(b) != n {
 			panic(io.ErrUnexpectedEOF)
 		}
-		if d.n += n; d.n > maxDecodeLen {
+		if d.n += n; d.n > MaxDecodeSize {
 			panic("encoded type exceeds size limit")
 		}
 		return b
@@ -286,7 +286,7 @@ func (d *Decoder) decode(val reflect.Value) {
 		val.SetUint(DecUint64(d.readN(8)))
 	case reflect.String:
 		strLen := DecUint64(d.readN(8))
-		if strLen > maxSliceLen {
+		if strLen > MaxSliceSize {
 			panic("string is too large")
 		}
 		val.SetString(string(d.readN(int(strLen))))
@@ -296,7 +296,7 @@ func (d *Decoder) decode(val reflect.Value) {
 		sliceLen := DecUint64(d.readN(8))
 		// sanity-check the sliceLen, otherwise you can crash a peer by making
 		// them allocate a massive slice
-		if sliceLen > 1<<31-1 || sliceLen*uint64(val.Type().Elem().Size()) > maxSliceLen {
+		if sliceLen > 1<<31-1 || sliceLen*uint64(val.Type().Elem().Size()) > MaxSliceSize {
 			panic("slice is too large")
 		} else if sliceLen == 0 {
 			return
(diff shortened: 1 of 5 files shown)


DavidVorick:
I think I'd prefer a complete solution via a fully custom `UnmarshalSia` for `types.Block`. Can you write a benchmark that demonstrates a performance improvement?

lukechampine:
I benchmarked by creating and unlocking a wallet from scratch (rescanning ~120,000 blocks).

Naive decoder: 6m2s, 58.8 GB GB allocated
Custom decoder: 4m44s, 48.6 GB allocated

i.e. ~20% faster unlock speed and ~20% less memory allocated.

Profiling indicates that most of the remaining allocations are "necessary," e.g. allocating the various slices in the `types.Transaction` struct. We could eliminate these in theory (consensus could reuse the memory of previously-decoded blocks) but it would be ugly and questionably safe.

I used a somewhat novel strategy to minimize allocations: threading a `decHelper` struct through methods that take an `io.Reader`. Each `UnmarshalSia` method calls `decoder` on the `io.Reader`; if the underlying type is a `decHelper`, it returns it, otherwise it creates and returns a new `decHelper`. This also allows me to use helper methods within the various `UnmarshalSia` methods. The error handling approach is monadic, accumulating errors internally and returning them at the end with `d.Err()`. Probably this should be reworked to store the first non-nil error (instead of the most recent error), and to return immediately from all methods if the stored error is non-nil.

Finally, this code is a bit less safe than the reflection-based decoder in one regard: it doesn't protect against unacceptably large slice lengths. That is, if an encoded object tells you that it's a slice with 2^64 elements, this new code will go right ahead and try to allocate it, whereas the reflection-based decoder will not. I plan to restore this bit of sanity checking, but it will take a bit of work.

rudibs:
i wonder if that unacceptably large slice length guard difference represents a network split attack vector for e.g. double spending

DavidVorick:
I was looking for a golang benchmark lol.

Though, shaving more than 60 seconds off of the full unlock time is unquestionably very impressive and worth merging. We should still have a golang benchmark for block encoding though. We do have one for empty blocks, though they don't seem to have been affected by the full 20% that you are reporting for natural blocks (which is perhaps not surprising).

DavidVorick:
@rudibs if you change the longest allowed slice, then yes that introduces a hardfork. In this case though, it's only being added to the string field, which as best I can remember there are no strings in the consensus types, so we should be okay.

encoding/marshal.go
```diff
@@ -228,6 +228,16 @@ func (d *Decoder) DecodeAll(vs ...interface{}) error {
 
 // readN reads n bytes and panics if the read fails.
 func (d *Decoder) readN(n int) []byte {
+	if buf, ok := d.r.(*bytes.Buffer); ok {
+		b := buf.Next(n)
+		if len(b) != n {
+			panic(io.ErrUnexpectedEOF)
+		}
+		if d.n += n; d.n > maxDecodeLen {
```
DavidVorick: Was this check already in place elsewhere?

DavidVorick:
Overall this looks good, but just like with the other MarshalSia implementations we wrote, we need a sanity check in here that verifies the marshalling pattern matches the old code exactly - this is consensus code and mistakes will cause hardforks.

lukechampine:
It just occurred to me that I can use the `decoder` trick to reduce allocations in `MarshalSia` too. Right now one of the largest sources of allocations is `encoding.WriteUint64`, because it has to allocate a new slice for each uint64 it writes. But if we're threading an `encoder` type through the `MarshalSia` methods, we can reuse a single buffer to encode all the writes. I'll do that in a follow up PR.

encoding/marshal.go
lukechampine: yep, it's checked in the decoder's `Read` method, which was called by `readN`

lukechampine:
Benchmarks:
```
old:

BenchmarkDecodeEmptyBlock-4	 2000000    750 ns/op	 85.26 MB/s    6 allocs/op
BenchmarkDecodeHeavyBlock-4	  100000  19735 ns/op 	 72.87 MB/s  167 allocs/op

new:

BenchmarkDecodeEmptyBlock-4	 5000000    384 ns/op	166.45 MB/s    3 allocs/op
BenchmarkDecodeHeavyBlock-4	  300000   4398 ns/op	326.93 MB/s   51 allocs/op
```

The new test adds coverage for all the new `UnmarshalSia` methods.

rudibs:
Great logic David and Luke!!  You really explained it well thanks.

lukechampine:
ok, maximum sizes are now enforced. This should be ready for final review

PR #2285, opened 2017-09-02
Reduce encoding allocations

lukechampine:
Follow up to #2282, implementing a similar strategy for reducing allocations. This should have a huge effect on our total number of allocations, and the resulting amount of time we spend in GC. The reason for this is that we hash lots of data, and hashing requires encoding, and encoding requires allocations. The two biggest offenders are `encoding.WriteUint64` (for writing length prefixes) and `Currency.MarshalSia` (which makes a copy of the bytes before writing them). I was able to nullify both of these by writing length prefixes to a reusable buffer and by doing some bitmath on the raw Currency bytes to avoid the copy. The end result is that most encodings require only a single allocation, even for very complex objects.

Microbenchmark:
```
old:

BenchmarkEncodeEmptyBlock-4	10000000    127 ns/op   500.19 MB/s     4 allocs/op
BenchmarkEncodeHeavyBlock-4	  500000   3330 ns/op   455.85 MB/s   102 allocs/op

new:

BenchmarkEncodeEmptyBlock-4	10000000    127 ns/op   501.06 MB/s     2 allocs/op
BenchmarkEncodeHeavyBlock-4	 1000000   1812 ns/op   837.71 MB/s     3 allocs/op
```

And a more realistic benchmark, unlocking a fresh wallet:

Master branch: 50.0 GB GB allocated (640 million objects)
PR branch: 43.6 GB allocated (470 million objects)

This wasn't as dramatic as I was hoping, but I think it's because rescanning is mostly a decode-heavy operation.

The block encoding already includes sanity checks, and it hasn't complained during testing or during the unlocks. I can sync a full node again just to be sure, though.

### DIFF
--- a/types/encoding.go
+++ b/types/encoding.go
@@ -96,9 +96,8 @@ func encoder(w io.Writer) *encHelper {
 		return e
 	}
 	return &encHelper{
-		w: w,
-		// 32 bytes is large enough for everything except StorageProof.Segment
-		buf: make([]byte, 32),
+		w:   w,
+		buf: make([]byte, 64), // large enough for everything but ArbitraryData
 	}
 }
 
--- a/types/encoding.go
+++ b/types/encoding.go
@@ -2,6 +2,7 @@ package types
 
 import (
 	"bytes"
+	"encoding/binary"
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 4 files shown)


types/encoding.go
```diff
@@ -301,7 +375,26 @@ func (c *Currency) UnmarshalJSON(b []byte) error {
...
+		_S    = 1 << _logS // number of bytes per big.Word
+	)
+
+	e.reset()
+	bits := c.i.Bits()
+	off := (len(bits) * _S) - size
+	for j := off; j < len(bits)*_S; j++ {
+		d := bits[j/_S] >> (64 - 8*(1+(uint(j)%_S)))
+		e.append(byte(d))
+	}
```
lukechampine: this is pretty esoteric. I'd like to clean it up a bit at some point (reviewers are welcome to take a stab), but it works.

lukechampine:
neat -- turns out the stdlib jpeg encoder uses a similar strategy: https://golang.org/src/image/jpeg/writer.go#L216

DavidVorick:
Did you do a full sync to double check that you didn't break any of the encoding?

lukechampine:
fully synced with no issues.

PR #2286, opened 2017-09-02
refactoring of log2SectorSize variable

starius:


lukechampine:
might be cleaner to define a `newCachedTree` helper function, since that seems to be the only use for log2SectorSize.

starius:
I created a `NewCachedTree` helper function in the `modules` package.

DavidVorick:
The cached trees of the current size work well for contracts that are hundreds of GBs but I think it's reasonable to assume in the future we'll have contracts in the TBs range, or even tens to hundreds of TBs, at which point we'll want another layer of cacheing.

This solution isn't nearly broad enough for our long term plans

starius:
Currently log2SectorSize's definition is inlined in two places. It depends only on SectorSize, so I moved its definition to the same place where SectorSize is defined. If SectorSize becomes dynamic in future, log2SectorSize will be a function instead of a variable/constant.

PR #2304, opened 2017-09-10
host API: stop leaking if the host has the contract with ID sent by renter

starius:


modules/renter/contractor/host_integration_test.go
```diff
@@ -1024,3 +1026,65 @@ func TestIntegrationCachedRenew(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func TestContractPresenceLeak(t *testing.T) {
```
avahowell: missing docstring

starius: Fixed

Issue #2292, opened 2017-09-05
Better Handling of Dust Thresholds


DavidVorick:
Currently Sia does not include dust outputs when it calculates your balance. This dust-excluded balance is the balance that the UI displays to the users. The transaction history however does not indicate ever that a transaction has outputs which are below the dust threshold, which means summing up the balances of your transaction history will result in a different balance than what is in the UI for users with small numbers of SC or high numbers of dust transactions (rare).

One tricky bit is that the dust threshold changes over time, so if you poll the endpoint and see that the dust threshold has moved significantly (>33% vs. last check), you should update all the transactions because some which were not previously dust may now be dust, or vice-versa.

This issue is best tackled after https://github.com/NebulousLabs/Sia/issues/2291 is complete in the Sia repo.

avahowell:
Should we close this, now that #2305 is merged?

aiorla:
I think they are different, #2305 made the dust threshold variable and this issue is more about using it to mark the transactions that are considered dust. I am actually going to tackle #2290 right now (if no one is interested) that as David explains can be seen as a previous step. After that I will probably will try this one.

tbenz9:
@aiorla now that #2291 has been merged in, are you planning on working on a fix for this issue?  :)

aiorla:
@tbenz9 not really, I don't have much time now. Feel free to take it!

MSevey:
@DavidVorick  should I take this into account with #3087 ?  If it has an impact on what the wallet balance is and therefore how the spend reporting should add up then it probably makes sense to include it.

Issue #2291, opened 2017-09-05
Add API Endpoint for Dust Threshold


DavidVorick:
We need to add an endpoint to the wallet API that queries the dust threshold of the wallet. This is good for the UI to know so that it can mark which transactions have outputs that are dust outputs, effectively warning the user that the transaction is not going to be counted for its full balance in the total balance.

How the dust threshold is managed is really a wallet implementation decision, which is why we shouldn't just tell people to take the /fee endpoint and multiply by 3.

Like all API endpoint changes, any pull request to add this needs to update doc/API.md, doc/api/wallet.md, and needs to have a test definitively proving that the dust threshold endpoint is displaying sane values and is functioning correctly.

PR #2305, opened 2017-09-11
Make dust threshold variable

### LINKED ISSUE #2290: Code wallet.dustValue() to use transactionpool.FeeEstimation
Currently the helper function `dustValue()` is hardcoded to 1 SC. We should re-code it to use the `transactionpool.FeeEstimation()`, and return the dust value equal to `3*minFee`. This means the dust threshold changes over time.

The tricky thing about this is that `dustValue` calls out to another module, which means that it can't be called internally while the wallet is under lock. So it should probably be called `managedDustValue`, even though it doesn't technically act upon a stateful object. If a stateful object uses that function, the result is the same, potential deadlock unless the locking is handled correctly.

The reviewer of the pull request to fix this issue needs to be careful to ensure that all uses of the `dustValue` function have been updated to correctly deal with locking.

aiorla:
This PR fixes #2290. I am actually not sure if I have understood correctly what  "avoiding calling internally the function while the wallet is under lock" meant there, I am just starting to get familiar with the codebase and Go. (hopefully to help more than hinder).

### DIFF
--- a/modules/wallet/defrag.go
+++ b/modules/wallet/defrag.go
@@ -15,6 +15,7 @@ var (
 // managedCreateDefragTransaction creates a transaction that spends multiple existing
 // wallet outputs into a single new address.
 func (w *Wallet) managedCreateDefragTransaction() ([]types.Transaction, error) {
+	dustThreshold := w.managedDustThreshold()
 	minFee, _ := w.tpool.FeeEstimation()
 
 	w.mu.Lock()
@@ -28,7 +29,7 @@ func (w *Wallet) managedCreateDefragTransaction() ([]types.Transaction, error) {
 	// Collect a value-sorted set of siacoin outputs.
 	var so sortedOutputs
 	err = dbForEachSiacoinOutput(w.dbTx, func(scoid types.SiacoinOutputID, sco types.SiacoinOutput) {
-		if w.checkOutput(w.dbTx, consensusHeight, scoid, sco) == nil {
+		if w.checkOutput(w.dbTx, consensusHeight, scoid, sco, dustThreshold) == nil {
 			so.ids = append(so.ids, scoid)
 			so.outputs = append(so.outputs, sco)
 		}
--- a/modules/wallet/defrag.go
+++ b/modules/wallet/defrag.go
@@ -15,6 +15,7 @@ var (
 // managedCreateDefragTransaction creates a transaction that spends multiple existing
 // wallet outputs into a single new address.
 func (w *Wallet) managedCreateDefragTransaction() ([]types.Transaction, error) {
+	// dustThreshold and minFee have to be obtained separate from the lock
 	dustThreshold := w.managedDustThreshold()
 	minFee, _ := w.tpool.FeeEstimation()
 
--- a/modules/wallet/money.go
+++ b/modules/wallet/money.go
@@ -22,6 +22,7 @@ func (w *Wallet) managedDustThreshold() types.Currency {
 // ConfirmedBalance returns the balance of the wallet according to all of the
 // confirmed transactions.
 func (w *Wallet) ConfirmedBalance() (siacoinBalance types.Currency, siafundBalance types.Currency, siafundClaimBalance types.Currency) {
+	// dustThreshold has to be obtained separate from the lock
 	dustThreshold := w.managedDustThreshold()
 
 	w.mu.Lock()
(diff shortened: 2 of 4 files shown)


modules/wallet/money.go
```diff
@@ -13,17 +13,25 @@ type sortedOutputs struct {
...
+// managedDustThreshold returns the quantity below which a Currency is considered to be Dust.
+func (w *Wallet) managedDustThreshold() types.Currency {
+	minFee, _ := w.tpool.FeeEstimation()
+	return minFee.Mul64(3)
+}
+
 // ConfirmedBalance returns the balance of the wallet according to all of the
 // confirmed transactions.
 func (w *Wallet) ConfirmedBalance() (siacoinBalance types.Currency, siafundBalance types.Currency, siafundClaimBalance types.Currency) {
+	dustThreshold := w.managedDustThreshold()
```
DavidVorick: probably worth adding a quick comment to each of these lone `dustThreshold :=` calls

modules/wallet/defrag.go
```diff
@@ -15,6 +15,7 @@ var (
 // managedCreateDefragTransaction creates a transaction that spends multiple existing
 // wallet outputs into a single new address.
 func (w *Wallet) managedCreateDefragTransaction() ([]types.Transaction, error) {
+	dustThreshold := w.managedDustThreshold()
 	minFee, _ := w.tpool.FeeEstimation()
```
DavidVorick: same here, probably worth indicating that these fields are needed after the lock is acquired, but have to be acquired separate from the lock.

DavidVorick:
Overall looks pretty good, I think you were able to preserve all of our concurrency conventions and avoid any concurrency related bugs. Probably just want to add a few comments, even though everything you are doing is common patterns. We like to keep the codebase commented well enough that anyone can drop in anywhere and still mostly feel like they know what's going on, instead of requiring users to learn all of our conventions before being able to read and review our code.

modules/wallet/money.go
```diff
@@ -13,17 +13,25 @@ type sortedOutputs struct {
 	outputs []types.SiacoinOutput
 }
 
+// managedDustThreshold returns the quantity below which a Currency is considered to be Dust.
+func (w *Wallet) managedDustThreshold() types.Currency {
+	minFee, _ := w.tpool.FeeEstimation()
+	return minFee.Mul64(3)
```
avahowell: Why does this need a special helper function? Should we follow the pattern discussed in the (defragThreshold PR)[https://github.com/NebulousLabs/Sia/pull/2242#discussion-diff-133795634R49] and inline this?

modules/wallet/defrag.go
avahowell: You already have the fee here (with `minFee`), you can avoid the additional tpool call and just:

```go
dustThreshold := minFee.Mul64(3)
```

modules/wallet/money.go
aiorla: Ups, I did not see that. I still think the helper is useful as it is called 4 times in 3 different files and `dustThreshold` is seen as an entity as issues #2291 & #2292 point. But you decide.

avahowell: I think in this case it probably makes sense to have a separate helper, since the `dustThreshold` value is useful in multiple places throughout the wallet.

modules/wallet/defrag.go
aiorla: @avahowell I did not inline the function because of my other comment and because it seemed more clear. But if it has to be inlined or the extra call is a performance issue I'll change it right away.

aiorla:
Comments added.

modules/wallet/money.go
DavidVorick: Unlike the fee calculations, which will always be using the `FeeEstimation` from the tpool, the dust threshold doesn't need to be inherently 3x the minFee. That's what it is in Bitcoin, but you could make a case for 1.5x as well, and potentially you could even have a more elaborate situation that takes into account historic fee lows, using some other method than FeeEstimation.

So the TL;DR is that I'm not really confident that this is the optimal way to do the dust threshold,  but I am confident that doing it this way is better than the way we were previously doing it.

DavidVorick: For example, if we make the dust threshold more of a long-term value (2x the smallest value seen in the past 12,000 blocks), then we could have the defrag function refuse to run if the `FeeEstimation` is not close enough to the cost of performing the defrag on that particular output. That makes the defragger smarter, but also requires altering this function in a way that is relevant to everyone else using it.

PR #2306, opened 2017-09-11
add second void output to file contracts

lukechampine:
Currently, file contracts have a single "void" output, where coins are burnt to prevent certain misaligned-incentive attacks (e.g. the renter DoS'ing the host). During a revision, missed payout funds are moved from the renter to void, and missed payout collateral is moved from the host to the void. This works perfectly well, but it has one unfortunate side-effect: the void output contains a mixture of renter and host coins. The reason this is unfortunate is that, were the coins cleanly separated, we could trivially compute the original number of coins paid in by the renter and the host (i.e. the original values of `ValidProofOutputs`). If we separate the coins by adding a new void output to `MissedProofOutputs`, and specify that index 2 is the "renter void" and index 3 is the "host void", then we can calculate the original funds via:

```go
missed := contract.LastRevision.NewMissedProofOutputs
origRenterFunds := missed[0] + missed[2]
origHostFunds := missed[1] + missed[3]
```

For reporting purposes, it's important that we know how much the renter originally paid into a contract, so currently we track this via separate metadata in the `modules.RenterContract` object. But there is no way to determine this value simply by looking at the most recent revision of the contract. You would have to lookup the parent ID in the blockchain.

This doesn't help Sia in the short-term, since we'll need to maintain compatibility for a while with hosts that don't support 2 void payouts. But down the road, I think it will be a useful property to have.

---

This code is passing tests despite making only small changes to the host, which I find somewhat surprising. It's possible that the host is not fully validating the missed proof outputs when accepting contract revisions. This warrants further investigation.

### DIFF
--- a/modules/renter/proto/formcontract.go
+++ b/modules/renter/proto/formcontract.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"net"
 
+	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/encoding"
 	"github.com/NebulousLabs/Sia/modules"
@@ -73,6 +74,16 @@ func FormContract(params ContractParams, txnBuilder transactionBuilder, tpool tr
 			// Once we start doing revisions, we'll move some coins to the host and some to the void.
 			{Value: types.ZeroCurrency, UnlockHash: types.UnlockHash{}},
 		},
+	}
+	if build.VersionCmp(host.Version, "1.3.0") >= 0 {
+		// After version 1.3.1, file contracts will have an extra void output
+		// in MissedProofOutputs. The renter moves voided coins to
+		// MissedProofOutputs[2], and the host moves voided coins to
+		// MissedProofOutputs[3]. This allows us to calculate the original
+		// amount of funds paid into the contract by the renter and host.
+		fc.MissedProofOutputs = append(fc.MissedProofOutputs, types.SiacoinOutput{
+			Value: types.ZeroCurrency, UnlockHash: types.UnlockHash{},
+		})
 	}
 
 	// Calculate transaction fee.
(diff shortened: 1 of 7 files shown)


modules/renter/proto/formcontract.go
```diff
@@ -74,6 +75,16 @@ func FormContract(params ContractParams, txnBuilder transactionBuilder, tpool tr
 			{Value: types.ZeroCurrency, UnlockHash: types.UnlockHash{}},
 		},
 	}
+	if build.VersionCmp(host.Version, "1.3.0") >= 0 {
```
lukechampine: for testing reasons this is 1.3.0. We should bump it to v1.3.1 after confirming that the tests pass

modules/renter/proto/renew.go
```diff
@@ -69,6 +70,18 @@ func Renew(contract modules.RenterContract, params ContractParams, txnBuilder tr
 			{Value: basePrice.Add(baseCollateral), UnlockHash: types.UnlockHash{}},
 		},
 	}
+	if build.VersionCmp(host.Version, "1.3.0") >= 0 {
```
lukechampine: ditto

DavidVorick:
I am hesitant to do this because it adds data to the blockchain, which is a scarce resource. On the other hand, it does make it easier to recover ones spending history in the event of local machine loss, which is definitely a good thing. It also makes some of the messier, easier to screw-up logic in the renter very clean, which is also a good thing.

But it means we're doing a protocol upgrade, which is a sticky thing in itself.

lukechampine:
looks like the second output adds about 50 bytes to the txn. The size of a file contract txn is ~1kb, so that's a 5% increase. Not much, but not completely negligible either. I think it's worth it; if nothing else, it's in line with our goal of making Sia developer-friendly.

DavidVorick:
Closing this for now, tradeoffs are uncertain and we're trying to stay away from changing the protocol for the time being.

lukechampine:
Ok. Perhaps we can revisit when implementing seed-based recovery.

PR #2310, opened 2017-09-12
cancel upnp discovery

lukechampine:


modules/gateway/upnp.go
```diff
@@ -94,6 +104,15 @@ func (g *Gateway) threadedForwardPort(port string) {
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	go func() {
+		select {
+		case <-g.threads.StopChan():
+			cancel()
+		case <-ctx.Done():
+		}
+	}()
 	d, err := upnp.Discover()
```
ChrisSchinnerl: Shouldn't this be `d, err := upnp.DiscoverCtx(ctx)`? Seems like you only changed it the first time and forgot the other 5.

lukechampine: derp, good catch

PR #2313, opened 2017-09-12
Add DustThreshold to /wallet API endpoint

### LINKED ISSUE #2291: Add API Endpoint for Dust Threshold

aiorla:
This PR fixes #2291. Two comments/questions:
- `TestWalletDustThreshold` is an adaptation of `TestTransactionPoolFee` to test the api endpoint like previous tests. It creates a simple test server and checks if the api endpoint result equals the direct call one but I am not sure if it clears the _sane value_ requirement that is said in the issue.
Is a different test needed or just to check its value in the real blockchain?
- Is adding this threshold value as an extra line in the `wallet balance` output of `siac` a good a idea? (after the estimated fee)

### DIFF
--- a/api/wallet.go
+++ b/api/wallet.go
@@ -28,6 +28,8 @@ type (
 
 		SiafundBalance      types.Currency `json:"siafundbalance"`
 		SiacoinClaimBalance types.Currency `json:"siacoinclaimbalance"`
+
+		DustThreshold types.Currency `json:"dustthreshold"`
 	}
 
 	// WalletAddressGET contains an address returned by a GET call to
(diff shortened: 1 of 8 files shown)


aiorla:
Discord comments fixed. @DavidVorick

DavidVorick:
Looks good to me

PR #2321, opened 2017-09-15
add RPC call Metadata requesting sector ids

starius:
Currently it only supports requesting entire list of sector ids.
In future it is planned to request a slice of sector ids by
specifying [begin, end) indices. The slice is accompanied with
a proof, that can be verified against the contract merkle root.

The implementation of proofs for slices is in review:
NebulousLabs/merkletree#17

That is why currently we support only the case begin=0 end=size.
In this case proof is empty. That is why the protocol in forward
compatible with the future plan.


lukechampine:
The sector roots will comprise a non-trivial amount of data -- enough that the host will likely want to charge for them. So this may require a more involved change.
I'm going to close this for now because it's a long term change, and one that we'd probably want to implement ourselves.

PR #2322, opened 2017-09-15
limit parallel downloads for auto repair

ChrisSchinnerl:
This PR should fix the high memory usage when repairing files that are not present on disk.

### DIFF
--- a/api/renterhost_test.go
+++ b/api/renterhost_test.go
@@ -2383,7 +2383,7 @@ func TestRemoteFileRepairMassive(t *testing.T) {
 	}
 
 	// Create a file to upload.
-	filesize := int(40000)
+	filesize := int(4000)
 	path := filepath.Join(st.dir, "test.dat")
 	err = createRandFile(path, filesize)
 	if err != nil {
--- a/api/renterhost_test.go
+++ b/api/renterhost_test.go
@@ -2331,3 +2331,192 @@ func TestRepairLoopBlocking(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+// TestRemoteFileRepairMassive is similar to TestRemoteFileRepair but uploads
+// more files to find potential deadlocks or crashes
+func TestRemoteFileRepairMassive(t *testing.T) {
+	if testing.Short() || !build.VLONG {
+		t.SkipNow()
+	}
+	st, err := createServerTester(t.Name())
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer st.server.Close()
+	stH1, err := blankServerTester(t.Name() + " - Host 1")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer stH1.server.Close()
+	testGroup := []*serverTester{st, stH1}
+
+	// Connect the testers to eachother so that they are all on the same
+	// blockchain.
+	err = fullyConnectNodes(testGroup)
+	if err != nil {
+		t.Fatal(err)
+	}
+	// Make sure that every wallet has money in it.
+	err = fundAllNodes(testGroup)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Add storage to every host.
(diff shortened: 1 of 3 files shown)


modules/renter/repair.go
```diff
@@ -290,14 +288,7 @@ func (r *Renter) managedRepairIteration(rs *repairState) {
...
 		}
 
-		// Don't update gaps for downloaded chunks twice
 		numGaps := chunkStatus.numGaps(rs)
-		if !downloading {
-			// Update the number of gaps for this chunk.
-			rs.gapCounts[chunkStatus.recordedGaps]--
-			rs.gapCounts[numGaps]++
-			chunkStatus.recordedGaps = numGaps
-		}
```
DavidVorick: pretty sure that this code is critical code, I don't think you want it removed.

DavidVorick: good rule of thumb for unstable opaque code is to assume that everything is important in delicate an unknown ways. I don't really remember why I put this here,  but I do remember that the gap count stuff was both tricky and also critical

modules/renter/repair.go
```diff
@@ -352,6 +343,11 @@ func (r *Renter) managedRepairIteration(rs *repairState) {
 // managedDownloadChunkData downloads the requested chunk from Sia, for use in
 // the repair loop.
 func (r *Renter) managedDownloadChunkData(rs *repairState, file *file, offset uint64, chunkIndex uint64, chunkID chunkID) ([]byte, error) {
+	// Don't initiate too many downloads to avoid using up all memory
+	if len(rs.downloadingChunks) >= maxScheduledDownloads {
+		return nil, nil
+	}
```
DavidVorick: I don't think we have any tests that try to repair 5 chunks at the same time, so though the test suite is passing, this construction is not being verified at all.

modules/renter/repair.go
```diff
@@ -290,14 +292,11 @@ func (r *Renter) managedRepairIteration(rs *repairState) {
...
-		// Don't update gaps for downloaded chunks twice
+		// Update the number of gaps for this chunk.
 		numGaps := chunkStatus.numGaps(rs)
-		if !downloading {
-			// Update the number of gaps for this chunk.
-			rs.gapCounts[chunkStatus.recordedGaps]--
-			rs.gapCounts[numGaps]++
-			chunkStatus.recordedGaps = numGaps
-		}
+		rs.gapCounts[chunkStatus.recordedGaps]--
```
avahowell: what's the rationale for moving this outside of the `if !downloading` branch?

api/renterhost_test.go
```diff
@@ -2331,3 +2331,192 @@ func TestRepairLoopBlocking(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+// TestRemoteFileRepairMassive is similar to TestRemoteFileRepair but uploads
+// more files to find potential deadlocks or crashes
+func TestRemoteFileRepairMassive(t *testing.T) {
```
avahowell: 👍

api/renterhost_test.go
```diff
@@ -2331,3 +2331,192 @@ func TestRepairLoopBlocking(t *testing.T) {
...
+	}
+
+	// add a few new blocks in order to cause the renter to form contracts with the new host
+	for i := 0; i < 10; i++ {
+		b, err := stNewHost.miner.AddBlock()
+		if err != nil {
+			t.Fatal(err)
+		}
+		for _, tester := range testGroup {
+			err = waitForBlock(b.ID(), tester)
```
DavidVorick: can just run `synchronizationCheck` here I'm pretty sure it's equivalent to this loop

ChrisSchinnerl: Are you sure? I copied that code from the other testcase and this loop is also in there. `synchronizationCheck` is already alled a few lines above. (line 2461)

PR #2324, opened 2017-09-15
bump version to v1.3.1

lukechampine:


modules/renter/hostdb/hostweight.go
```diff
@@ -236,6 +236,9 @@ func versionAdjustments(entry modules.HostDBEntry) float64 {
 	if build.VersionCmp(entry.Version, "1.4.0") < 0 {
 		base = base * 0.99999 // Safety value to make sure we update the version penalties every time we update the host.
 	}
+	if build.VersionCmp(entry.Version, "1.3.1") < 0 {
+		base = base * 0.9
```
DavidVorick: 0.6 - some of the changes we made affect host-renter relationships, and it's definitely better to be talking to v131 hosts than other hosts.

lukechampine:
fixed

PR #2332, opened 2017-09-18
add string spellchecking to the build

mtlynch:
Adding a spellcheck step to the build to catch misspelled words in comments and
output strings.

Fixing all of the existing spelling errors.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,7 @@ dependencies:
 	go get -u github.com/spf13/cobra/...
 	# Developer Dependencies
 	go install -race std
+	go get -u github.com/client9/misspell/cmd/misspell
 	go get -u github.com/golang/lint/golint
 	go get -u github.com/NebulousLabs/glyphcheck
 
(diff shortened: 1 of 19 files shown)


lukechampine:
From what I recall, `misspell` doesn't catch *all* spelling errors, just most of the common ones.
I could see this being frustrating for people who make a PR and only find out much later that the build failed due to a spelling error. On the other hand, I really do prefer the code to be free from spelling errors. Ideally everyone would run this as a pre-commit hook, but there's no way to enforce that.

I also notice that a lot of the misspellings are just British variants. Not sure if enforcing American spelling is necessary. Is there a way we can accept both?

mtlynch:
>I could see this being frustrating for people who make a PR and only find out much later that the build failed due to a spelling error. On the other hand, I really do prefer the code to be free from spelling errors. Ideally everyone would run this as a pre-commit hook, but there's no way to enforce that.

Up to you guys. If you want, I can pull out the `Makefile` changes and just leave the spelling corrections.

>I also notice that a lot of the misspellings are just British variants. Not sure if enforcing American spelling is necessary. Is there a way we can accept both?

Sure, I removed the US locale preference, so now it accepts US or UK spellings. As a check, I also changed it back to "optimi**s**ing**" in `vendor/github.com/pkg/errors/bench_test.go` and `misspell` is cool with it.

mtlynch:
ping

lukechampine:
I think we can accept this, but without `spellcheck` being required on `test-long` and `test-vlong`. I'm on  the fence about it, but we can always decide later whether to require it for PRs and such. For now, just having it as a makefile target is sufficient.

mtlynch:
Okay, updated.

lukechampine:
thanks!

PR #2338, opened 2017-09-19
fix decoding bug

lukechampine:
credit to @avahowell for finding this via fuzzing

### DIFF
--- a/types/encoding.go
+++ b/types/encoding.go
@@ -163,7 +163,7 @@ func (d *decHelper) NextUint64() uint64 {
 	if d.err != nil {
 		return 0
 	}
-	d.Read(d.buf[:])
+	io.ReadFull(d, d.buf[:])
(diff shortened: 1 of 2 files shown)


types/encoding.go
```diff
@@ -163,14 +163,17 @@ func (d *decHelper) NextUint64() uint64 {
 	if d.err != nil {
 		return 0
 	}
-	d.Read(d.buf[:])
+	io.ReadFull(d, d.buf[:])
```
avahowell: should we be checking the error returned here?

lukechampine: the error is checked internally by the `d.Read` method. I suppose it doesn't hurt though.

lukechampine:
rearranged the error checking so that the methods of `decoder` always check for `d.err` before returning

PR #2344, opened 2017-09-21
Fix re-open journal if rename fails

sgraves66:
If journal rename fails, the fd becomes invalid and journal processing will never recover. Antivirus software on Windows causes the rename to intermittently fail.
- Added 5 retry attempts on failed rename
- Check for invalid handle before trying to close

### DIFF
--- a/modules/renter/contractor/journal.go
+++ b/modules/renter/contractor/journal.go
@@ -19,7 +19,6 @@ package contractor
 // simply ignored when reading the journal.
 
 import (
-	"build"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -27,6 +26,7 @@ import (
 	"os"
 	"reflect"
 	"syscall"
+	"time"
 
 	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/crypto"
@@ -117,7 +117,9 @@ func (j *journal) checkpoint(data contractorPersist) error {
 	}
 
 	// Attempt rename up to 5 times to prevent issues with anti-virus software
-	err = build.Retry(5, 1000*Millisecond, func() error { return os.Rename(tmp.Name(), j.filename) })
+	err = build.Retry(5, 100*time.Millisecond, func() error {
+		return os.Rename(tmp.Name(), j.filename)
+	})
 	if err != nil {
 		return err
 	}
--- a/modules/renter/contractor/journal.go
+++ b/modules/renter/contractor/journal.go
@@ -25,7 +25,6 @@ import (
 	"io"
 	"os"
 	"reflect"
-	"syscall"
 	"time"
 
 	"github.com/NebulousLabs/Sia/build"
@@ -107,21 +106,17 @@ func (j *journal) checkpoint(data contractorPersist) error {
 	if err := tmp.Close(); err != nil {
 		return err
 	}
-
-	// If rename or close failed in a previous call, fd will be invalid,
-	//	so don't try to close
-	if fd := j.f.Fd(); fd != uintptr(syscall.InvalidHandle) {
-		if err := j.f.Close(); err != nil {
-			return err
-		}
+	if err := j.f.Close(); err != nil {
+		return err
 	}
 
-	// Attempt rename up to 5 times to prevent issues with anti-virus software
+	// Attempt rename up to 5 times to prevent issues with anti-virus software.
+	// If after 5 attempts file cannot be renamed, raise critical failure
 	err = build.Retry(5, 100*time.Millisecond, func() error {
 		return os.Rename(tmp.Name(), j.filename)
 	})
 	if err != nil {
-		return err
+		build.Critical("Unable to rename contractor.journal:", err)
 	}
 
 	// Reopen the journal.
--- a/modules/renter/contractor/journal.go
+++ b/modules/renter/contractor/journal.go
@@ -25,6 +25,7 @@ import (
 	"io"
 	"os"
 	"reflect"
+	"syscall"
 
 	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/crypto"
@@ -105,10 +106,22 @@ func (j *journal) checkpoint(data contractorPersist) error {
 	if err := tmp.Close(); err != nil {
 		return err
 	}
-	if err := j.f.Close(); err != nil {
-		return err
+
+	// If rename or close failed in a previous call, fd will be invalid, 
+	//	so don't try to close
+	if fd := j.f.Fd(); fd != uintptr(syscall.InvalidHandle) {
+		if err := j.f.Close(); err != nil {
+			return err
+		}
 	}
-	if err := os.Rename(tmp.Name(), j.filename); err != nil {
+
+	// Attempt rename up to 5 times to prevent issues with anti-virus software
+	renamed := false
+	for i := 0; !renamed && i < 5; i++ {
+		err := os.Rename(tmp.Name(), j.filename)
+		renamed = err == nil
+	}
+	if err != nil {
 		return err
 	}
 
--- a/modules/renter/contractor/journal.go
+++ b/modules/renter/contractor/journal.go
@@ -19,6 +19,7 @@ package contractor
 // simply ignored when reading the journal.
 
 import (
+	"build"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -107,7 +108,7 @@ func (j *journal) checkpoint(data contractorPersist) error {
 		return err
 	}
 
-	// If rename or close failed in a previous call, fd will be invalid, 
+	// If rename or close failed in a previous call, fd will be invalid,
 	//	so don't try to close
 	if fd := j.f.Fd(); fd != uintptr(syscall.InvalidHandle) {
 		if err := j.f.Close(); err != nil {
@@ -116,11 +117,7 @@ func (j *journal) checkpoint(data contractorPersist) error {
 	}
 
 	// Attempt rename up to 5 times to prevent issues with anti-virus software
-	renamed := false
-	for i := 0; !renamed && i < 5; i++ {
-		err := os.Rename(tmp.Name(), j.filename)
-		renamed = err == nil
-	}
+	err = build.Retry(5, 1000*Millisecond, func() error { return os.Rename(tmp.Name(), j.filename) })
 	if err != nil {
 		return err
 	}


modules/renter/contractor/journal.go
```diff
@@ -105,10 +106,22 @@ func (j *journal) checkpoint(data contractorPersist) error {
...
+		}
 	}
-	if err := os.Rename(tmp.Name(), j.filename); err != nil {
+
+	// Attempt rename up to 5 times to prevent issues with anti-virus software
+	renamed := false
+	for i := 0; !renamed && i < 5; i++ {
+		err := os.Rename(tmp.Name(), j.filename)
+		renamed = err == nil
+	}
```
lukechampine: we have a `build.Retry` function (from our internal `build` package) that may be useful here

sgraves66:
Thanks. I'll take a look at including it. Guess my only question is will a single failed copy cause any issues somewhere else in the code? 

My work laptop cannot launch siad at all due to this. It seems to work fine with the fix, just making sure that it doesn't introduce something more.

modules/renter/contractor/journal.go
```diff
@@ -105,10 +107,18 @@ func (j *journal) checkpoint(data contractorPersist) error {
...
 		return err
 	}
-	if err := j.f.Close(); err != nil {
-		return err
+
+	// If rename or close failed in a previous call, fd will be invalid,
+	//	so don't try to close
+	if fd := j.f.Fd(); fd != uintptr(syscall.InvalidHandle) {
+		if err := j.f.Close(); err != nil {
+			return err
```
lukechampine: shouldn't `Close` always work unless `j.f == nil`?
I think we can actually ignore the error here, since we call `Sync` beforehand. i.e.
```go
if j.f. != nil {
    j.f.Close()
}
```

modules/renter/contractor/journal.go
```diff
@@ -19,12 +19,14 @@ package contractor
 // simply ignored when reading the journal.
 
 import (
+	"build"
```
lukechampine: should be `github.com/NebulousLabs/Sia/build`

modules/renter/contractor/journal.go
```diff
@@ -105,10 +107,18 @@ func (j *journal) checkpoint(data contractorPersist) error {
...
+	//	so don't try to close
+	if fd := j.f.Fd(); fd != uintptr(syscall.InvalidHandle) {
+		if err := j.f.Close(); err != nil {
+			return err
+		}
 	}
-	if err := os.Rename(tmp.Name(), j.filename); err != nil {
+
+	// Attempt rename up to 5 times to prevent issues with anti-virus software
+	err = build.Retry(5, 1000*Millisecond, func() error { return os.Rename(tmp.Name(), j.filename) })
```
lukechampine: 1 second might be excessive, 100ms should be ok. Also, prefer to put the function body on its own line

modules/renter/contractor/journal.go
sgraves66: I guess it depends on what should be logged in consensus.  Sorry if syntax is incorrect. f.n is never nil, just the fd is set to 'syscall.InvalidHandle'.
- Option 1 - Ignore Close error altogether
```
j.f.Close()
if err := os.Rename(tmp.Name(), j.filename); err != nil {
   return err;
}
```
Invalid handle will not be logged, but rename will log 'Access is denied', which could be misleading as the root cause is invalid handle.


- Option 2 - Only rename if Close() is successful
```
if err := j.f.Close(); err == nil {
      if err := os.Rename(tmp.Name(), j.filename); err != nil {
       return err;
    }
}
```
Only rename errors are logged, but have no way of knowing there's an issue if Close() returns an error


- Option 3 - Log root cause if either Close() or Rename() fail
```
if err := j.f.Close(); err != nil {
   return err;
} else if err := os.Rename(tmp.Name(), j.filename); err != nil {
       return err;
    }
}
```

modules/renter/contractor/journal.go
sgraves66: Sure. 5 retries with 1 second delay is just my goto on Windows. It was working with no delay, so should be fine.

sgraves66:
```
err = build.Retry(5, 100*Millisecond, 
   func() error { return os.Rename(tmp.Name(), j.filename) })
```
or
```
err = build.Retry(5, 100*Millisecond, 
		func() error {
			 return os.Rename(tmp.Name(), j.filename) 
		})
```

avahowell: I think

```go
err = build.Retry(5, 100*Millisecond, func() error {
  return os.Rename(tmp.Name(), j.filename) 
})
```

is preferred

modules/renter/contractor/journal.go
lukechampine: Do we know *why* the fd gets set to `InvalidHandle`? Somehow antivirus is interfering? Dropping down to the `syscall` interface is more low-level than I'd prefer. Maybe if we can figure out the root cause, we can handle the problem at a higher level.

Why does `Rename` return `Access is denied` if the handle is invalid?

sgraves66: It becomes invalid because the rename fails and the reopen isn't called. The next call starts with a closed handle while expecting an open one. Reopen is never being called again when rename failed.

sgraves66: close sets the handle to invalid on success.

sgraves66: as for access is denied, looks like that's what is returned by go's file system calls when trying to rename given an invalid fd.

sgraves66: essentially, reopen has to be called if either close or rename fails. otherwise, stuck in an invalid fd loop.

sgraves66: stuck because close returns an error when trying to close an invalid handle.

lukechampine: so in principle, this shouldn't happen as long as the `Rename` (eventually) succeeds? Maybe the addition of `build.Retry` will be enough to prevent the `InvalidHandle` bug. If the `Rename` can't succeed after 5 retries, that's probably sufficient cause to crash the program. We definitely don't want it to be silently failing in the background.

sgraves66: I would suspect drive failure at that point or some other process has opened a handle. contractor.log fills to tens of megabytes if i leave to it's own devices. siad does eventually execute an orderly shutdown in >5 minutes or so.

sgraves66: if rename succeeds, it's fine. out of 3 win instances only one with McAfee AV ever gets into this state.

lukechampine: ok. My preference is to treat the `j.f.Close()` error normally (i.e. how it was treated before the PR), and if the `build.Retry` error is not nil, crash the program using `build.Critical("Unable to save contractor.log:", err)`. Are you able to test that on your McAfee machine? How reproducible is this?

sgraves66: and actually rename could fail as long as the failure isn't due to invalid fd. root of it is closing an invalid handle. if it's valid, the next call will simply close it and try the rename again.

sgraves66: immediately reproducible. it never launches again.

sgraves66:
Let me elaborate a bit on the re-open retry. Not only anti-virus/malware software, but also backup/sync apps can cause open's to fail. I had to exclude certain locations in Sia's data directory in CrashPlan because of open file failures in siad as well. To generalize, any external process that open's a file that Sia expects to be readable and/or writable can cause issues if you don't implement a retry. 

Windows is specifically a bit more touchy than *nix as typically O_EXCL is the only flag on *nix that could cause something like this. Windows introduces share mode flags FILE_SHARE_READ, FILE_SHARE_WRITE, and FILE_SHARE_DELETE. Specifying no flags is essentially the same as O_EXCL. 

This [article](https://msdn.microsoft.com/en-us/library/windows/desktop/aa363874(v=vs.85).aspx) will explain what can and cannot occur based on how a file is originally opened given each combination of share flags. Retrying on failed open/create will help prevent unnecessary crashes or invalid application states caused by unexpected internal/external file access.

sgraves66:
Yes, contractor.log is same for me. rename fails and then starts spewing same errors.

DavidVorick:
@sgraves66 do you only experience this issue with the contractor journal? I'm wondering if there are also potentially problems with our persist package. How consistently do you see issues? If there are issues with other files, would you be able to check your logs and verify that they do or do not exist?

Here is our persist code: https://github.com/NebulousLabs/Sia/blob/master/persist/json.go

I ask because we are working on a new method for saving which has much higher performance than other techniques we've been using. But I don't understand this anti-virus bug very well yet and I want to make sure that our new method of securely saving files isn't going to have negative conflicts with the anti-virus.

sgraves66:
So far, only this one particular file. Due to the frequency of changes, seems more likely to fail on open. Some AV software will exclusively open a file while it is being scanned, so any additional open requests will immediately fail. I'm addressing Windows more than *NIX here, but the solution is common for both. Historically, file opens will undoubtedly fail at some point because of av/malware/backup software, poorly designed text editors, etc. The best I can do here is pass along a generalization. If it's being opened at all, it's best to implement a retry.

Say 98% success first time open. Typically, only one additional retry is necessary to bring it up to 99%. And 99.5% success third time. 5 retries is my personal preference for catching the remaining. In doing so consistently, you immediately reduce the risk of crashing and invalid states caused by failed opens that are most likely recoverable events. There's really no performance cost if the first open succeeds. Depending upon how long you wait between retries, it should be a trivial hit anyways.

sgraves66:
Just looked through the persist code. Any time you repeatedly open, write and close, there's cause for performance degradation. If you can open once, flush when you really need to ensure data is written to disk, and close on shutdown that would be preferable in most cases. It would also alleviate issues with failed open as you only need one to succeed throughout the lifetime of your app. Not to mention ultimate control over who else can access it, when/how, etc. While you're  running, I doubt that external processes should even be allowed anything other then read access anyways.
Maybe encapsulate all file ops inside a high level TextFile/JsonFile class? At least you would have only one place to make performance modifications. Keep most frequently used in a map for quick retrieval. Let destructor/GC handle the close.

sgraves66:
But it doesn't solve issues with corruption like renaming does. I really don't use text files in this way for live data anymore. There are plenty of libraries available for storing data, relational or otherwise, that would work quite well. I have no experience with go, so not sure what's available.

Issue #2110, opened 2017-07-11
Storage proof revenue is incorrect in host.log


nielscastien:
The value that is shown with a successfull or failed storage proof is the Potential revenue of the host.

Line 558 in storageobligation.go should be changed to:

`h.log.Printf("Successfully submitted a storage proof. Revenue is %v.\n", so.financialMetrics.PotentialContractCompensation.Add(so.financialMetrics.PotentialStorageRevenue).Add(so.financialMetrics.PotentialDownloadBandwidthRevenue).Add(so.financialMetrics.PotentialUploadBandwidthRevenue))`

The same holds for line 574

nielscastien:
I believe this issue is linked with #1862

tbenz9:
Fixed in PR #2347 

Recommendation to devs: close

Issue #1862, opened 2017-06-01
Missed storage proof


mtlynch:
I'm seeing a lot of entries in my `host.log` saying "Missed storage proof. Revenue would have been...":

```
$ grep proof host.log
2017/04/23 10:23:20.374783 storageobligations.go:575: Missed storage proof. Revenue would have been 28041465401979693369917440.
2017/04/25 06:19:28.893826 storageobligations.go:575: Missed storage proof. Revenue would have been 42044787803384516685332480.
2017/04/29 03:47:09.905632 storageobligations.go:575: Missed storage proof. Revenue would have been 64044788828658827693916160.
2017/05/02 17:08:29.988152 storageobligations.go:575: Missed storage proof. Revenue would have been 93044789341295983198208000.
2017/05/03 19:04:56.117655 storageobligations.go:575: Missed storage proof. Revenue would have been 94044789341295983198208000.
2017/05/05 09:20:59.541486 storageobligations.go:575: Missed storage proof. Revenue would have been 108044789341295983198208000.
2017/05/05 09:20:59.563966 storageobligations.go:575: Missed storage proof. Revenue would have been 107044789341295983198208000.
2017/05/05 18:44:30.087300 storageobligations.go:575: Missed storage proof. Revenue would have been 107044789341295983198208000.
2017/05/07 00:08:43.077733 storageobligations.go:575: Missed storage proof. Revenue would have been 119044789341295983198208000.
2017/05/07 11:17:56.318427 storageobligations.go:575: Missed storage proof. Revenue would have been 120044789341295983198208000.
2017/05/08 01:48:35.603738 storageobligations.go:575: Missed storage proof. Revenue would have been 122044789341295983198208000.
2017/05/08 01:52:16.685029 storageobligations.go:575: Missed storage proof. Revenue would have been 121044789341295983198208000.
2017/05/08 01:52:16.799666 storageobligations.go:575: Missed storage proof. Revenue would have been 120044789341295983198208000.
2017/05/08 13:04:28.447039 storageobligations.go:575: Missed storage proof. Revenue would have been 120044789341295983198208000.
2017/05/08 13:04:28.658554 storageobligations.go:575: Missed storage proof. Revenue would have been 119044789341295983198208000.
2017/05/10 06:31:03.691053 storageobligations.go:575: Missed storage proof. Revenue would have been 125044789341295983198208000.
[...]

PR #2347, opened 2017-09-21
Fix log output

nielscastien:
This PR should fix the wrong output that shows the potential revenue of the host instead of the (missed) revenue of the storage obligation.

Fix for #2110 and #1862


lukechampine:
LGTM but @DavidVorick should weigh in on whether or not this behavior is intentional

DavidVorick:
oh. This explains a lot of confusing logs I've seen. No wonder they didn't make sense.

Thanks for this, the way it was previously was definitely wrong.

PR #2351, opened 2017-09-23
added auto-unlock functionality

eddiewang:
This PR adds auto-unlock capabilities on  start in you have a SIA_WALLET_PASSWORD env variable set.

Most of the wrappers are taken from siac. I slightly modified the apiPost func to automatically use    the APIPassword if it exists.


eddiewang:
What the terminal prints:
image

siad/daemon.go
```diff
@@ -295,6 +377,9 @@ func startDaemon(config Config) (err error) {
 	// connect the API to the server
 	srv.mux.Handle("/", a)
 
+	// Attempt to auto-unlock the wallet
+	walletAutoUnlock(config)
```
avahowell: It might make more sense to just pass `w` to `walletAutoUnlock` and use `(*wallet).Unlock` directly instead of duplicating all of this API call code from siac. Check out `walletUnlockHandler` in `api/wallet.go` for an example of how to extract a `crypto.TwofishKey` from a string.

eddiewang: @avahowell that actually makes alot more sense than the way I'm doing it. I'll get on that PR later tonight.

eddiewang:
@avahowell I changed it to use the wallet interface. code is much shorter now 👍

siad/daemon.go
lukechampine: I think using the API handler is better actually, since it more closely replicates the behavior of siac. But instead of duplicating the HTTP code, you can use `api.HttpPOST` and `api.HttpPOSTAuthenticated`.

lukechampine: I guess it's kind of a wash. Calling the HTTP endpoint means you don't have to do the awkward multiple-password handling. But it also means you're calling an HTTP endpoint, and it's undeniably a bit weird for a server to call its own endpoints.

eddiewang: @lukechampine with the HTTP endpoint, we pass english as the query param, but technically there could be another dictionary right? 

I don't think this commit covers the other languages, since english is hard coded as the param passed to dictionary.

eddiewang:
any updates on the direction for this PR? @lukechampine

ericflo:
This would be helpful for my use case as well.

siad/daemon.go
```diff
@@ -295,6 +325,9 @@ func startDaemon(config Config) (err error) {
 	// connect the API to the server
 	srv.mux.Handle("/", a)
 
+	// Attempt to auto-unlock the wallet
+	walletAutoUnlock(w)
```
lukechampine: probably better to pull some of the logic out of the function:
```go
// Attempt to auto-unlock the wallet
if password := os.Getenv("SIA_WALLET_PASSWORD"); password != "" {
    fmt.Println("Attempting to auto-unlock wallet with SIA_WALLET_PASSWORD environment variable")
    if err := unlockWallet(w, password); err != nil {
        fmt.Println("Auto-unlock failed")
    }
}
```

siad/daemon.go
```diff
@@ -109,6 +111,34 @@ func processConfig(config Config) (Config, error) {
...
+	validKeys = append(validKeys, crypto.TwofishKey(crypto.HashObject(password)))
+	for _, key := range validKeys {
+		err := w.Unlock(key)
+		if err == nil {
+			fmt.Println("Wallet unlocked")
+		}
+		if err != nil && err != modules.ErrBadEncryptionKey {
+			fmt.Println("Automatic unlock failed!")
+		}
+	}
```
lukechampine: won't this print `Automatic unlock failed!` for each key in `validKeys` that isn't correct? I think we want:
```go
for _, key := range validKeys {
    if err := w.Unlock(key); err == nil {
        return nil
    }
}
return modules.ErrBadEncryptionKey
```

eddiewang:
@lukechampine I've merged the current master from upstream and fixed the code based on your comments. Thank you for the guidance!

lukechampine:
FWIW you can approximate this behavior by having your script call `siac wallet unlock` after starting siad; but I suppose this could be tricky since your script needs to wait for siad to finish loading.

I guess my concern here is that people who want siac to auto-unlock may not necessarily want siad to auto-unlock on startup. I will merge this as-is, but we should discuss further whether more work is needed. For example, we could add a flag that disables auto-unlock. But it may also be the case that anyone using `SIA_WALLET_PASSWORD` doesn't care if siad auto-unlocks.

DavidVorick:
It seems weird to me that we would be unlocking the wallet automatically simply b/c we know the password, but I guess most users will probably prefer it  that way.

tbenz9:
Since I've been asked several times about where this feature is documented, it can be found with `siac wallet unlock --help`
Sample output:
```
tbenz9@spartan1:~/go/src/github.com/tbenz9/Sia/doc/api$ siac wallet unlock --help
Decrypt and load the wallet into memory.
Automatic unlocking is also supported via environment variable: if the
SIA_WALLET_PASSWORD environment variable is set, the unlock command will
use it instead of displaying the typical interactive prompt.

Usage:
  siac wallet unlock [flags]

Flags:
  -h, --help       help for unlock
  -p, --password   Display interactive password prompt even if SIA_WALLET_PASSWORD is set

Global Flags:
  -a, --addr string   which host/port to communicate with (i.e. the host/port siad is listening on) (default "localhost:9980")
```

davidlcassidy:
Yeah, setting this variable isn't working for me either on Windows 10.

PR #2352, opened 2017-09-23
split newRevision for clarity

petabytestorage:


modules/renter/proto/negotiate.go
```diff
@@ -172,26 +172,29 @@ func negotiateRevision(conn net.Conn, rev types.FileContractRevision, secretKey
...
-	rev.NewValidProofOutputs = make([]types.SiacoinOutput, 2)
-	rev.NewMissedProofOutputs = make([]types.SiacoinOutput, 3)
-	copy(rev.NewValidProofOutputs, current.NewValidProofOutputs)
-	copy(rev.NewMissedProofOutputs, current.NewMissedProofOutputs)
-
 	// move valid payout from renter to host
-	rev.NewValidProofOutputs[0].Value = current.NewValidProofOutputs[0].Value.Sub(cost)
-	rev.NewValidProofOutputs[1].Value = current.NewValidProofOutputs[1].Value.Add(cost)
-
+	rev.NewValidProofOutputs = newRevisionWithAdjustedPayout(current.NewValidProofOutputs, 2, 0, 1, cost)
```
DavidVorick: `newRevisionWithAdjustedPayout` is not a good name for the function, as all it really does is return adjusted payouts.

Second issue is with the inputs, they are all numeric. The function call is not readable unless you already know the function definition, as you have no idea what the numbers correspond to.

Finally, I would say that correctness is less obvious here. It needs to be obvious that money is moving from the renter to the host, and from the renter to the void. This code is actually a little confusing to me (the existing code, not the proposed code), because I thought that you were supposed to be moving money from the host to the void as well, but we can see here that we're not doing that.

--------

One definitely unfortunate thing here is that we didn't explain in the comments why we need to manually copy the slice memory over. I can't really remember the full details but I remember it was important. Maybe @lukechampine remembers? I think it's because slices are pointers, and we're using the underlying memory elsewhere in other parts of the execution.

modules/renter/proto/negotiate.go
```diff
@@ -172,26 +172,29 @@ func negotiateRevision(conn net.Conn, rev types.FileContractRevision, secretKey
...
 	rev.NewRevisionNumber++
 
 	return rev
 }
 
+// return a new revision that fixed-size copies an old SiacoinOutput and
+// makes one payout adjustment on the resulting copy.
+// TODO: figure out if we can use unsafeSize to eliminate "howBig" param
+func newRevisionWithAdjustedPayout(outputs []types.SiacoinOutput, howBig int,
+	losingIndex int, winningIndex int, cost types.Currency) []types.SiacoinOutput {
```
DavidVorick: we like to keep the full function definition on a single line of code, even if it's a long function definition. For function definitions with lots of parameters, or just opaque parameters in general (like these strings of ints), we prefer to use a struct as the input. For example:

```go
type payoutAdjustmentInput struct {
    outputs []types.SiacoinOutput
    numOuputs int
    payerIndex int
    payeeIndex int
    amountPaid types.Currency
}
```

This allows you to name the inputs to the function, making the code readable even if you've never actually seen the function definition.

modules/renter/proto/negotiate.go
DavidVorick: Reading this code at all does require understanding that in file contracts, we represent the renter money in output 0, the host money in output 1, and the void money in output 2

DavidVorick: That's probably worth documenting somewhere

modules/renter/proto/negotiate.go
```diff
@@ -12,6 +12,14 @@ import (
 	"github.com/NebulousLabs/Sia/types"
 )
 
+type payoutAdjustmentInput struct {
```
DavidVorick: all structs need a docstring

modules/renter/proto/negotiate.go
```diff
@@ -172,26 +180,26 @@ func negotiateRevision(conn net.Conn, rev types.FileContractRevision, secretKey
...
-
-	// move valid payout from renter to host
-	rev.NewValidProofOutputs[0].Value = current.NewValidProofOutputs[0].Value.Sub(cost)
-	rev.NewValidProofOutputs[1].Value = current.NewValidProofOutputs[1].Value.Add(cost)
-
-	// move missed payout from renter to void
-	rev.NewMissedProofOutputs[0].Value = current.NewMissedProofOutputs[0].Value.Sub(cost)
-	rev.NewMissedProofOutputs[2].Value = current.NewMissedProofOutputs[2].Value.Add(cost)
+	rev.NewValidProofOutputs = newPayoutAdjustment(payoutAdjustmentInput{outputs: current.NewValidProofOutputs, numOutputs: 2, payerIndex: types.FileContractRenterIndex, payeeIndex: types.FileContractHostIndex, amountPaid: cost})
+	rev.NewMissedProofOutputs = newPayoutAdjustment(payoutAdjustmentInput{outputs: current.NewMissedProofOutputs, numOutputs: 3, payerIndex: types.FileContractRenterIndex, payeeIndex: types.FileContractVoidIndex, amountPaid: cost})
```
DavidVorick: probably best to use this style for readability:

```newPayoutAdjustment(payoutAdjustmentInput{
    outputs: current.NewValidProofOutputs
    etc.
})
```

modules/renter/proto/negotiate.go
```diff
@@ -12,6 +12,14 @@ import (
 	"github.com/NebulousLabs/Sia/types"
 )
 
+type payoutAdjustmentInput struct {
+	outputs    []types.SiacoinOutput
+	numOutputs uint
```
DavidVorick: should be able to determine `numOutputs` by calling `len(outputs)`, if those aren't the same we've got other problems I think

types/constants.go
```diff
@@ -48,6 +48,10 @@ var (
 	OakMaxRise              *big.Rat
 	OakMaxDrop              *big.Rat
 	OakHardforkTxnSizeLimit = uint64(64e3) // 64 KB
+
+	FileContractRenterIndex = uint(0)
+	FileContractHostIndex   = uint(1)
+	FileContractVoidIndex   = uint(2)
```
lukechampine: makes more sense to define these in the `modules` package with the rest of the negotiation code

modules/renter/proto/negotiate.go
```diff
@@ -172,26 +184,35 @@ func negotiateRevision(conn net.Conn, rev types.FileContractRevision, secretKey
...
-	copy(rev.NewMissedProofOutputs, current.NewMissedProofOutputs)
-
-	// move valid payout from renter to host
-	rev.NewValidProofOutputs[0].Value = current.NewValidProofOutputs[0].Value.Sub(cost)
-	rev.NewValidProofOutputs[1].Value = current.NewValidProofOutputs[1].Value.Add(cost)
+	rev.NewValidProofOutputs = newPayoutAdjustment(payoutAdjustmentInput{
+		outputs:    current.NewValidProofOutputs,
+		payerIndex: types.FileContractRenterIndex,
+		payeeIndex: types.FileContractHostIndex,
+		amountPaid: cost})
```
lukechampine: rather than a new struct type, I think I'd prefer a helper function:
```go
func moveOutputValue(outputs []types.SiacoinOutput, amount types.Currency, from, to int) []types.SiacoinOutput {
    newOutputs := append([]types.SiacoinOutput(nil), outputs...)
    newOutputs[from].Value = newOutputs[from].Value.Sub(amount)
    newOutputs[to].Value = newOutputs[to].Value.Add(amount)
    return newOutputs
}

func newRevision(current types.FileContractRevision, cost types.Currency) types.FileContractRevision {
    rev := current
    // move valid payout from renter to host
    rev.NewValidProofOutputs = moveOutputValue(
        rev.NewValidProofOutputs,
        cost,
        modules.FileContractRenterIndex,
        modules.FileContractHostIndex,
    )
    // move missed payout from renter to void
    rev.NewMissedProofOutputs = moveOutputValue(
        rev.NewMissedProofOutputs,
        cost,
        modules.FileContractRenterIndex,
        modules.FileContractVoidIndex,
    )
    // increment revision number
    rev.NewRevisionNumber++
    return rev
}
```

petabytestorage: Feels like a lot of different variations for a basic cleanup task. I was partial to the one David led me to before and also like the one @avahowell helped me realize with https://github.com/NebulousLabs/Sia/pull/2353  Mostly I wanted to do this task to try to get "something" improved in terms of clarity in less than a few hours.

DavidVorick: I definitely prefer the style @lukechampine just presented, it's readable and doesn't introduce a whole new struct. Switching out the `from` and `two` ints to named consts really improves the understanding for what's happening here.

This pattern appears all over the `proto` package, if we're going to clean stuff up, it makes sense to do it right. When you put a lot of effort into getting the first few steps correct, the rest are a lot easier and you end up with a final product that you are much happier with.

lukechampine:
Closing because we weren't able to agree on implementation. We'll be overhauling the renter over the coming weeks, so perhaps we can revisit this down the line.

petabytestorage:
what about https://github.com/NebulousLabs/Sia/pull/2353 ? all bad?  better close all then and delete if so.

Issue #2007, opened 2017-06-25
Binaries siad and siac are dynamically linked but they don't have to


starius:
Binaries siad and siac from [latest release](https://github.com/NebulousLabs/Sia/releases/download/v1.2.2/Sia-v1.2.2-linux-amd64.zip):

```
$ file siad
siad: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked (uses shared libs), not stripped
```

This means that `siad` binary produced on a system with newer libc won't run a system with older libc. @cinerar reported on Slack that siad compiled on one machine failed to run on another machine and he believed it was related to the fact that the binary was dynamically linked.

According to [a go-nuts thread](https://groups.google.com/d/msg/golang-nuts/B8WPRKajvtI/do_NrAndlNkJ):

> [...] on Linux, networking code in the Go standard
library (the "net" package) by default links against libc to be able to
use its getaddrinfo().  The "os" package also links against libc to be
able to use its means for its "name service switch" interface to look
up users by their IDs and vice versa. 

Both `CGO_ENABLED=0` and `-tags netgo` result in static `siad` binary, which means in case of Sia it is only `net`, not `os`. The `net` package has pure Go replacement of `getaddrinfo`.

I think, `-tags netgo` can be added to Makefile.

lukechampine:
The full explanation give in the Go docs is:

> On Unix systems, the resolver has two options for resolving names. It can use a pure Go resolver that sends DNS requests directly to the servers listed in /etc/resolv.conf, or it can use a cgo-based resolver that calls C library routines such as getaddrinfo and getnameinfo.
>
>By default the pure Go resolver is used, because a blocked DNS request consumes only a goroutine, while a blocked C call consumes an operating system thread. When cgo is available, the cgo-based resolver is used instead under a variety of conditions: on systems that do not let programs make direct DNS requests (OS X), when the LOCALDOMAIN environment variable is present (even if empty), when the RES_OPTIONS or HOSTALIASES environment variable is non-empty, when the ASR_CONFIG environment variable is non-empty (OpenBSD only), when /etc/resolv.conf or /etc/nsswitch.conf specify the use of features that the Go resolver does not implement, and when the name being looked up ends in .local or is an mDNS name.

>The resolver decision can be overridden by setting the netdns value of the GODEBUG environment variable (see package runtime) to go or cgo.
>
>The decision can also be forced while building the Go source tree by setting the netgo or netcgo build tag.
>
[...]

PR #2357, opened 2017-09-24
build with netgo tag to make the executable static

starius:
This change didn't affect siad.exe built with `GOOS=windows make release-std` - it produces exactly the same file.

fix https://github.com/NebulousLabs/Sia/issues/2007


lukechampine:
I don't expect this cause any problems, but if it does, we'll be able to catch it during RC and easily revert this commit. 👍 for static binaries

Issue #1775, opened 2017-05-07
runtime: program exceeds 10000-thread limit


poiuty:
```
# siac version
Sia Client v1.2.1
Sia Daemon v1.2.1
```

```
runtime: program exceeds 10000-thread limit
fatal error: thread exhaustion

runtime stack:
runtime.throw(0xaa3788, 0x11)
	/usr/local/go/src/runtime/panic.go:596 +0x95
runtime.checkmcount()
	/usr/local/go/src/runtime/proc.go:519 +0xa4
runtime.mcommoninit(0xc457678800)
	/usr/local/go/src/runtime/proc.go:539 +0xa0
runtime.allocm(0xc42001a000, 0x0, 0xc400000000)
	/usr/local/go/src/runtime/proc.go:1341 +0x99
runtime.newm(0x0, 0xc42001a000)
	/usr/local/go/src/runtime/proc.go:1611 +0x39
runtime.startm(0xc42001a000, 0xe96700)
	/usr/local/go/src/runtime/proc.go:1698 +0x17e
runtime.handoffp(0xc42001a000)
	/usr/local/go/src/runtime/proc.go:1725 +0x55
runtime.retake(0x53f7ec8deb9d9, 0x0)
	/usr/local/go/src/runtime/proc.go:3909 +0x15d
runtime.sysmon()
	/usr/local/go/src/runtime/proc.go:3836 +0x1ea
runtime.mstart1()
	/usr/local/go/src/runtime/proc.go:1179 +0x11e
runtime.mstart()
	/usr/local/go/src/runtime/proc.go:1149 +0x64

goroutine 1 [chan receive, 8719 minutes]:
main.startDaemon(0x0, 0x0, 0xaa096a, 0xe, 0xa8fe07, 0x5, 0xa8fe0c, 0x5, 0x0, 0x7ffe3c9ebeea, ...)
	/home/luke/go/src/github.com/NebulousLabs/Sia/siad/daemon.go:293 +0xa23
main.startDaemonCmd(0xc420074fc0, 0xc4200b9880, 0x0, 0x2)
	/home/luke/go/src/github.com/NebulousLabs/Sia/siad/daemon.go:309 +0x6d
github.com/spf13/cobra.(*Command).execute(0xc420074fc0, 0xc420010100, 0x2, 0x2, 0xc420074fc0, 0xc420010100)
	/home/luke/go/src/github.com/spf13/cobra/command.go:648 +0x231
[...]

PR #2361, opened 2017-09-25
fix and optimize AddSectorBatch

starius:
AddSectorBatch is applied to the whole list of sectors of a contract being
renewed. The function started a goroutine per sector before. Each goroutine
made disk access to update the sector's reference counter. This resulted
is thousands of goroutines and potentially OS threads.
The tool crashed when a large contract was renewed:
https://github.com/NebulousLabs/Sia/issues/1775#issuecomment-330666550

This commit removes the goroutine-launching behaviour and instead
groups sector updates by storage folder and merges adjacent sectors
into groups writing them together in order to optimize the number
of write operations issued by the tool.

### DIFF
--- a/modules/host/contractmanager/sector.go
+++ b/modules/host/contractmanager/sector.go
@@ -3,6 +3,7 @@ package contractmanager
 import (
 	"encoding/binary"
 	"errors"
+	"sort"
 	"sync"
 	"sync/atomic"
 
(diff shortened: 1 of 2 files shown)


DavidVorick:
This PR needs some benchmarks. The reason, from what I remember, that I added all the goroutines was so that we could get better queue-depth on the disk. Typical usage pattern means that you aren't generally going to have many adjacent sectors on a contract renewal, because the sectors are placed on-disk completely randomly.

The use case I'm particularly concerned about is HDDs, because they suffer a lot from seek times. If we aren't doing the queue depth well, we're looking at as much as 7ms per renewed sector, which means 10 minutes for a 400GB contract. I'm actually not sure at all how the OS and how golang on the backend handles disk optimization, so it's something we have to proceed with guided by benchmarks.

The overall goal with AddSectorBatch is that we can commit to a 40+ TB contract (10 million sectors) renewal in under 500ms, and then apply it in less than 20 seconds. That's going to take some pretty aggressive cacheing and optimization, but I think it's an achieveable long term goal.

In the medium term, it'd be really great if we could hit those time targets for contracts that are 1 TB in size.

starius:
I'm closing this for now - the issue was mitigated in https://github.com/NebulousLabs/Sia/pull/2368 and https://github.com/NebulousLabs/Sia/pull/2373

Some proposals for future improvement of storage (not to forget them):
 * free unused storage space with fallocate: https://github.com/NebulousLabs/Sia/issues/2366
 * reuse unused storage sectors instead [choosing them randomly](https://github.com/NebulousLabs/Sia/blob/4e64b4176abbb83a05e49a174b8b2c33ced61d02/modules/host/contractmanager/storagefolder.go#L159)
 * store each sector in a separate file
 * provide a defragmentation tool (and probably run it on startup)

PR #2362, opened 2017-09-25
only commit host WAL if there are changes to commit

### LINKED ISSUE #2215: siad is constantly writing ~200 KiB to disk when 'host' module is enabled
I've setup siad and already host ~5GiB. I noticed that siad is constantly performing write operations.

```
Total DISK READ :     904.36 K/s | Total DISK WRITE :     239.43 K/s
Actual DISK READ:     904.36 K/s | Actual DISK WRITE:     344.83 K/s
  TID  PRIO  USER     DISK READ  DISK WRITE  SWAPIN     IO>    COMMAND                                  
  287 be/4 sia       208.20 K/s   39.04 K/s  0.00 %  9.93 % siad -d /var/lib/sia
 1272 be/4 sia        19.52 K/s   23.42 K/s  0.00 %  5.81 % siad -d /var/lib/sia
  274 be/4 sia        42.94 K/s   46.84 K/s  0.00 %  4.88 % siad -d /var/lib/sia
  383 be/4 sia         0.00 B/s   65.06 K/s  0.00 %  4.25 % siad -d /var/lib/sia
 1274 be/4 sia         0.00 B/s   52.05 K/s  0.00 %  4.02 % siad -d /var/lib/sia
  296 be/4 bitcoin   633.70 K/s    0.00 B/s  0.97 %  2.01 % bitcoind -daemon -con~.pid [bitcoin-msghand]
  343 be/4 sia         0.00 B/s   13.01 K/s  0.00 %  0.00 % siad -d /var/lib/sia
    1 be/4 root        0.00 B/s    0.00 B/s  0.00 %  0.00 % systemd --system --deserialize 20
    2 be/4 root        0.00 B/s    0.00 B/s  0.00 %  0.
```

The blockchain is fully synced. While it would be great, I doubt that there are constants file uploads going on, as my storage's folder used space does not grow.

Compare this to bitcoind running on the same system, which mostly performs reads and occasional writes (e.g. when a new block is found).

**Why does siad constantly write ~200 KiB/s to disk**?

<sup>[Original sia forum post](https://forum.sia.tech/topic/1537/siad-is-constantly-writing-to-disk)</sup>

lukechampine:
I don't understand the host well enough to say whether this change is safe, but we could definitely use this functionality. Currently the host creates, writes, syncs, and renames `contractmanager.wal_temp` and `contractmanager.json_temp` every 500ms, even if the host has been sitting idle. With this PR, the host performs no writes unless it's actually recording a change. This should greatly reduce the wear on idle host disks.

Fixes #2215


lukechampine:
btw, `inotifywait` is a great tool for monitoring disk actions. Start siad and run it like so:
```
sudo inotifywait -e modify -e move -e create -e delete -m -r [siad directory]
```

lukechampine:
looks like this is causing some failures in the host tests, although most of the smoke testing (upload+download) still appears to work.

pablos91:
Will this fix go to the next release ?

Best wishes

lukechampine:
I've updated this PR with changes that I believe accomplish the desired I/O reduction in a safe way.
There are two files that the host (actually the contractmanager) repeatedly writes, syncs, and renames: the settings file and the WAL. This happens even if the settings haven't changed and the WAL has no `uncommittedChanges`.
This PR prevents this, using a slightly different strategy for each file.

The WAL strategy is simple. If there are no `uncommittedChanges`, the WAL file is left untouched; otherwise, it is synced, closed, and renamed as usual.

The settings file is a little trickier, because there's no in-memory state (like `uncommittedChanges`) that tells us if the settings file needs to be committed. Instead, we now store the settings that were last written to disk in `wal.committedSettings`. Then, when the next save cycle rolls around, we compare the `committedSettings` to the current in-memory settings. If they match, we know we don't need to write a new file, so instead we set `wal.fileSettingsTmp = nil`. Otherwise, we proceed as normal.

Using the `inotifywait` command above, I confirmed that the host now performs no IO when idling.

tbenz9:
This also fixes #2590 (not sure if this comment will actually close the issue when merged though.

tbenz9:
@lukechampine We've had people asking about this on Discord again.  Would it be possible to get this PR merged in, then open a new PR for the "settings file"?

lukechampine:
David needs to give this at least a brief review before we can merge. I'll ping him about it.

PR #2368, opened 2017-09-27
fix AddSectorBatch on large contracts

starius:
AddSectorBatch is applied to the whole list of sectors of a contract being
renewed. The function started a goroutine per sector before. Each goroutine
made disk access to update the sector's reference counter. This resulted
is thousands of goroutines and potentially OS threads.
The tool crashed when a large contract was renewed:
https://github.com/NebulousLabs/Sia/issues/1775#issuecomment-330666550

This commit sets a cap for maximum parallel running goroutines in that group.

This PR is the mitigation to stop bleeding. Another PR which is more long term
and will take more time is https://github.com/NebulousLabs/Sia/pull/2361

### DIFF
--- a/modules/host/contractmanager/consts.go
+++ b/modules/host/contractmanager/consts.go
@@ -60,6 +60,10 @@ const (
 	// which is a high granluarity relative the to the TiBs of storage that
 	// hosts are expected to provide.
 	storageFolderGranularity = 64
+
+	// maxAddSectorBatchThreads is the maximum number of threads updating
+	// sector counters on disk in AddSectorBatch.
+	maxAddSectorBatchThreads = 100
 )
 
 var (
--- a/modules/host/contractmanager/sectorupdate.go
+++ b/modules/host/contractmanager/sectorupdate.go
@@ -411,10 +411,16 @@ func (cm *ContractManager) AddSectorBatch(sectorRoots []crypto.Hash) error {
 
 	// Add each sector in a separate goroutine.
 	var wg sync.WaitGroup
+	// Ensure only 'maxAddSectorBatchThreads' goroutines are running at a time.
+	semaphore := make(chan struct{}, maxAddSectorBatchThreads)
 	for _, root := range sectorRoots {
 		wg.Add(1)
+		semaphore <- struct{}{}
 		go func(root crypto.Hash) {
 			defer wg.Done()
+			defer func() {
+				<-semaphore
+			}()
 
 			// Hold a sector lock throughout the duration of the function, but release
 			// before syncing.


modules/host/contractmanager/sectorupdate.go
```diff
@@ -10,6 +10,12 @@ import (
 	"github.com/NebulousLabs/Sia/modules"
 )
 
+const (
+	// maxParallelThreads is the maximum number of threads updating
+	// sector counters on disk in AddSectorBatch.
+	maxParallelThreads = 100
+)
```
DavidVorick: belongs in consts.go, but will need a name like `maxAddSectorBatchThreads`

modules/host/contractmanager/sectorupdate.go
```diff
@@ -411,10 +417,15 @@ func (cm *ContractManager) AddSectorBatch(sectorRoots []crypto.Hash) error {
 
 	// Add each sector in a separate goroutine.
 	var wg sync.WaitGroup
+	semaphore := make(chan struct{}, maxParallelThreads)
 	for _, root := range sectorRoots {
 		wg.Add(1)
 		go func(root crypto.Hash) {
 			defer wg.Done()
+			semaphore <- struct{}{}
```
DavidVorick: Need to request the semaphore before you call `go func`, otherwise you're still going to end up with thousands of goroutines, they just won't be doing anything.

modules/host/contractmanager/sectorupdate.go
DavidVorick: I guess if we're being pedantic, it needs to say `maxAddSectorBatchGoroutines`

modules/host/contractmanager/sectorupdate.go
```diff
@@ -411,10 +417,15 @@ func (cm *ContractManager) AddSectorBatch(sectorRoots []crypto.Hash) error {
 
 	// Add each sector in a separate goroutine.
 	var wg sync.WaitGroup
+	semaphore := make(chan struct{}, maxParallelThreads)
```
DavidVorick: Can you add a quick comment here that says something like `// Ensure only 'maxParallelThreads' goroutines are running at a time`

DavidVorick:
Few structural comments, but over this PR is a good idea.

@starius If I remember correctly we use a similar pattern of spinning up a bajillion goroutines somewhere else in the contractor. Maybe in `RemoveSectorBatch`? Can be in another PR, but we probably should extend this pattern to that place as well.

modules/host/contractmanager/sectorupdate.go
lukechampine: we use `Threads` elsewhere in the codebase, so might as well stick with that

starius:
Updates for comments.

modules/host/contractmanager/consts.go
```diff
@@ -60,6 +60,10 @@ const (
 	// which is a high granluarity relative the to the TiBs of storage that
 	// hosts are expected to provide.
 	storageFolderGranularity = 64
+
+	// maxAddSectorBatchThreads is the maximum number of threads updating
+	// sector counters on disk in AddSectorBatch.
+	maxAddSectorBatchThreads = 100
```
DavidVorick: we typically sort consts alphabetically, prefer if you fix in another pr. I'll go ahead and merge this since it's a trivial fix, and I don't want to cause too much review friction.

Issue #2308, opened 2017-09-12
Align Contractor's Contracts into a Budget Cycle


DavidVorick:
Currently contracts will more or less arbitrarily align their end heights. This is no longer necessary for compatibility with the hosts (as of v1.2.1 iirc), and it does cause a lot of confusing transactions and a steady drain on the user's wallet instead of monthly deductions.

To more effectively match user expectations, we would like to align the end heights of the contracts with a billing cycle. The field `contractor.currentPeriod` indicates when the most recent contract period started. The end height of all contracts that get formed should become `contractor.currentPeriod + contractor.allowance.Period`, which means that all contracts should always end at the same height, unless the period is decreased in the middle.

The naming is not great here. 'Period' actually means 'contract duration', the user is going to be billed more frequently than the period, because a renew cycle happens every `allowance.Period - allowance.RenewWindow` blocks. So the billing cycle actually has a length/period of `allowance.Period - allowance.RenewWindow`.

When we hit the renew block height, we want to start a new billing period at height `contractor.currentPeriod + allowance.Period - allowance.RenewWindow`. Even if the current height is already a few weeks ahead of that (due to user machine being offline), we should still keep the alignment perfect and start the billing cycle at the time that it would have originally started. There is an edge case where the billing cycle is less than half of the period, which means we could potentially skip full billing cycles. So really, the equation for the new block height is `contractor.currentPeriod + ((allowance.Period - allowance.RenewWindow)*n)`, where `n` is the number of billing cycles that have elapsed. Usually `n` will equal 1, but if the user left the computer off for a long time or has funky allowance settings, it could be 2+.

tbenz9:
Was this fixed with #2369?

If yes, close.

MSevey:
I believe #3087  should resolve this as it is address #3077  which seems like a similar issue.

PR #2369, opened 2017-09-27
fix contractor currentPeriod tracking bug

avahowell:
This PR fixes a bug in the contractor, where the allowance period was not being accurately tracked.

### DIFF
--- a/modules/renter/contractor/allowance.go
+++ b/modules/renter/contractor/allowance.go
@@ -72,8 +72,12 @@ func (c *Contractor) SetAllowance(a modules.Allowance) error {
 
 	c.log.Println("INFO: setting allowance to", a)
 	c.mu.Lock()
+	// set the current period to the blockheight if the existing allowance is
+	// empty
+	if c.allowance.Funds.IsZero() && c.allowance.Hosts == 0 && c.allowance.Period == 0 && c.allowance.RenewWindow == 0 {
+		c.currentPeriod = c.blockHeight
+	}
 	c.allowance = a
-	c.currentPeriod = c.blockHeight
 	err = c.saveSync()
 	c.mu.Unlock()
 	if err != nil {
--- a/modules/renter/contractor/allowance.go
+++ b/modules/renter/contractor/allowance.go
@@ -2,6 +2,7 @@ package contractor
 
 import (
 	"errors"
+	"reflect"
 
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/types"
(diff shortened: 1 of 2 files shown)


aiorla:
I've also noticed that something wasn't right there today (I was trying to tackle #2308) and I said it on `#core-dev`. I was waiting for an answer to code it because I'm not sure about the expected behaviour.
I thought it was more like:
```go
if c.currentPeriod == 0 {
    c.currentPeriod = c.blockHeight
}
```
plus a few more lines to stop c.ProcessConsensusChange(...) to move it from 0.

lukechampine:
merge conflicts after I merged the other PR

avahowell:
fixed up the merge conflicts

modules/renter/contractor/allowance.go
```diff
@@ -73,6 +73,7 @@ func (c *Contractor) SetAllowance(a modules.Allowance) error {
 	c.log.Println("INFO: setting allowance to", a)
 	c.mu.Lock()
 	c.allowance = a
+	c.currentPeriod = c.blockHeight
```
DavidVorick: This is not quite correct behavior I think. You don't want to reset the current period every time that you change the allowance. For example, if all you are doing here is adjusting the spending amount, you would expect the current period to not change.

I'm not actually certain what you should do if you are adjusting the period. If you increase the period, should you change the currentPeriod? If you decrase it, should you change the currentPeriod? Not clear to me at the moment what the best choice is there.

DavidVorick: Generally I think we want to follow a pattern of sticking to a rhythmic billing cycle, for example one a month, always on the same day. But if you change from a month to 6 weeks... I'm not sure what you are supposed to do.

And, does this code properly clean up the contract lines? I think that if you yank the current period around like this you can end up with contract lines that actually span multiple periods.

DavidVorick: My gut right now says we should only change the current period if `c.currentPeriod+(a.Period-a.RenewWindow) < c.blockHeight`, meaning that we would want to renew instantly.

avahowell: Would we also want to start the `currentPeriod` at the current blockheight if an allowance is set and there are no previous allowances?

DavidVorick: In most cases, that would satisfy the above equation, but I guess there are some corner cases if your `a.Period-a.RenewWindow` is greater than the block height, you could have issues.

So yeah, seems fine to check if the allowance is blank, and if so then you set the currentPeriod to the block height.

DavidVorick:
@avahowell can you update this? Would like to see this merged into RC2.

avahowell:
@DavidVorick updated

PR #2371, opened 2017-09-28
Allow for noninteractive startup when using API authentication

ericflo:
I've been working on automating my setup into a kubernetes cluster, and I want it to be secure even though it's automated. Currently Sia asks for a password in an interactive way, which is a showstopper for this project.

This patch adds the environment variable ``SIA_API_PASSWORD``. On startup if it's detected to be non-empty, that will be used instead of prompting the user. Just as with ``SIA_WALLET_PASSWORD`` it prints a line indicating that it has picked up the password from the environment -- so nobody should be caught off-guard.


tbenz9:
Isn't this a duplicate of #2351?

avahowell:
Not quite: this PR adds an option to set the API password noninteractively, the linked PR adds automatic wallet unlocking.

lukechampine:
@ericflo Sorry for the delay; this LGTM. However, we'll need to fix the merge conflicts, which probably just means switching from `speakeasy` to `readPassword`.

ericflo:
This should be good to go now. It's running live on SiaCDN now too.

lukechampine:
Awesome, thanks!

PR #2372, opened 2017-09-28
first stab at limiter

lukechampine:
This should behave according to spec, but you (@DavidVorick) should confirm that the test code matches your expectations.

### DIFF
--- a/sync/limiter.go
+++ b/sync/limiter.go
@@ -76,6 +76,7 @@ func (l *Limiter) Release(n int) {
 	l.cond.L.Lock()
 	l.current -= n
 	if l.current < 0 {
+		l.cond.L.Unlock()
 		panic("units released exceeds units requested")
 	}
 	l.cond.L.Unlock()
--- a/sync/limiter.go
+++ b/sync/limiter.go
@@ -62,6 +62,9 @@ func (l *Limiter) Request(n int, cancel <-chan struct{}) bool {
 func (l *Limiter) Release(n int) {
 	l.cond.L.Lock()
 	l.current -= n
+	if l.current < 0 {
+		panic("units released exceeds units requested")
+	}
 	l.cond.L.Unlock()
 	l.cond.Signal()
 }
--- a/sync/limiter.go
+++ b/sync/limiter.go
@@ -19,10 +19,10 @@ import (
 // This design choice prevents any call to Request from blocking forever,
 // striking a balance between precise resource management and flexibility.
 type Limiter struct {
-	limit    int
-	current  int
-	requests chan struct{}
-	cond     *sync.Cond
+	limit   int
+	current int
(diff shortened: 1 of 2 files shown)


sync/limiter.go
```diff
@@ -0,0 +1,87 @@
...
+// Request returns true if the request was canceled, and false otherwise.
+func (l *Limiter) Request(n int, cancel <-chan struct{}) bool {
+	// wait until our request is "first in line"
+	select {
+	case <-cancel:
+		return true
+	case token := <-l.requests:
+		// return token when we're done
+		defer func() { l.requests <- token }()
+	}
```
lukechampine: this is basically just a mutex; the reason I implemented it with a channel is so that we can `select` to provide cancellation.
I also explored using the `sync.Cond` for this purpose (setting/checking a `locked` boolean var), but found it to be unwieldy.

Really hoping we get lock-channels in an upcoming Go release...

sync/limiter.go
```diff
@@ -0,0 +1,87 @@
...
+	return cancelled
+}
+
+// Release returns n units to l, making them available to future callers. It
+// is legal for n to be larger than l's limit, as long as n was previously
+// passed to Request.
+func (l *Limiter) Release(n int) {
+	l.cond.L.Lock()
+	l.current -= n
```
lukechampine: maybe `panic` here if `n > l.current`?

sync/limiter.go
```diff
@@ -0,0 +1,90 @@
+package sync
+
+import (
+	"sync"
+)
+
+// A Limiter restricts access to a resource.
```
avahowell: I think this docstring could be better. Reading this it's hard to get an idea of the high level goals of the `Limiter`, where a `Limiter` is useful, and what makes it different from other synchronization abstractions that restrict access to resources.

lukechampine:
@DavidVorick pls review and incorporate into upload code if this API is satisfactory

DavidVorick:
Correct me if I'm wrong, but it currently looks like small requests are able to starve large requests? We would want to make sure that if you make a large request, all small requests get denied until the large request is able to go through.

That means you'll need a queue I believe. It also means that technically only one request can be served at a time, which hopefully simplifies the concurrency structure. If you want to use channels, what you would do is have each thread add a channel to the queue along with the size of their resource request. When resources get returned, you check if there are enough resources to unblock the next request (or if there are 0 more resources in use), and then you close that resource's channel to signal that it can unblock.

You'll have to interact with the queue using a lock.

lukechampine:
I believe that the current design does not starve large requests. Requests are served on a first-come first-served basis. When a request is accepted, it obtains an exclusive lock. It holds the lock until enough resources are returned to fulfill the request. No other request receives resources while the lock is held.

PR #2373, opened 2017-09-28
cap on for maximum goroutines in RemoveSectorBatch

starius:
Also the constants were sorted alphabetically.

Follow-up for https://github.com/NebulousLabs/Sia/pull/2368

### DIFF
--- a/modules/host/contractmanager/consts.go
+++ b/modules/host/contractmanager/consts.go
@@ -46,6 +46,10 @@ const (
 	// a storageFolderGrow.
 	folderAllocationStepSize = 1 << 35
 
+	// maxSectorBatchThreads is the maximum number of threads updating
+	// sector counters on disk in AddSectorBatch and RemoveSectorBatch.
+	maxSectorBatchThreads = 100
+
 	// sectorMetadataDiskSize defines the number of bytes it takes to store the
 	// metadata of a single sector on disk.
 	sectorMetadataDiskSize = 14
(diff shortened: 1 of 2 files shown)


DavidVorick:
You mentioned that prior code was causing your machine to crash? Are you able to reproduce the situation and confirm that the crash issues are resolved?

starius:
The crash happened only once.
I built Sia 1.3.0 with these two pull requests cherry picked and restarted `siad`.
Now I need a contract of 20k sectors to be renewed to make sure it is working correctly.
How can I make and renew such a contract?

BTW can we also add a test reproducing this situation?

DavidVorick:
We probably cannot add an automated gotest to reproduce this, it'd have to be part of some other CI. We have talked about having a continuously running antfarm which would be able to reproduce this scale of test, but nothing we've built.

The core issue is that 20k sectors is a huge amount of work for the machine. Even in testing that requires generating and saving 80 MB of data, and then doing something on the order of 100,000 IOPS for the lifetime of the test. On an HDD, that's going to take a minimum of 10 minutes, simply because HDDs are slow.

To reproduce this in production (or on an antfarm), you'd want to create a host and then upload about 100 GB of data to that host over a single contract. Then wait for the contract to renew, and see if the host is able to handle it.

DavidVorick:
I'm in favor of starting to do more of these types of stress tests in some CI. We have a hefty machine with SSDs and HDDs both that is dedicated to running CI nonstop. But I wouldn't want to use the `go test` framework for our stress tests

And, even those might get harder if we start pushing the stress tests to multiple TBs of data

starius:
> you'd want to create a host and then upload about 100 GB of data to that host over a single contract.

Can the sector size be decreased in this test?

DavidVorick:
They could be, but I'm not sure exactly how well that would represent the bottlenecks. It's at least moderately useful though.

PR #2375, opened 2017-09-28
accurately calculate renter spending

avahowell:
This PR updates the renter to accurately reflect spending for the current billing period. The previous calculation did not take into account for all contracts in the billing line. This implementation provides a new contractor method, `PeriodSpending()`, which iterates through every contract _line_ in the current billing period to accurately calculate spending.

### DIFF
--- a/modules/renter.go
+++ b/modules/renter.go
@@ -256,6 +256,16 @@ type RenterContract struct {
 	PreviousContracts []RenterContract
 }
 
+// ContractorSpending contains the metrics about how much the Contractor has
+// spent during the current billing period.
+type ContractorSpending struct {
+	ContractSpending types.Currency `json:"contractspending"`
+	UploadSpending   types.Currency `json:"uploadspending"`
+	DownloadSpending types.Currency `json:"downloadspending"`
+	StorageSpending  types.Currency `json:"storagespending"`
+	Unspent          types.Currency `json:"unspent"`
+}
+
 // EndHeight returns the height at which the host is no longer obligated to
 // store contract data.
 func (rc *RenterContract) EndHeight() types.BlockHeight {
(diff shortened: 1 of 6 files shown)


modules/renter.go
```diff
@@ -256,6 +256,16 @@ type RenterContract struct {
...
 }
 
+// ContractorSpending contains the metrics about how much the Contractor has
+// spent during the current billing period.
+type ContractorSpending struct {
+	ContractSpending types.Currency `json:"contractspending"`
+	UploadSpending   types.Currency `json:"uploadspending"`
+	DownloadSpending types.Currency `json:"downloadspending"`
+	StorageSpending  types.Currency `json:"storagespending"`
+	Unspent          types.Currency `json:"unspent"`
```
DavidVorick: Needs to be sorted alphabetically

modules/renter/contractor/contractor.go
```diff
@@ -84,6 +84,31 @@ func (c *Contractor) Contract(hostAddr modules.NetAddress) (modules.RenterContra
...
+		spending.DownloadSpending = spending.DownloadSpending.Add(contract.DownloadSpending)
+		spending.UploadSpending = spending.UploadSpending.Add(contract.UploadSpending)
+		spending.StorageSpending = spending.StorageSpending.Add(contract.StorageSpending)
+		for _, pre := range contract.PreviousContracts {
+			spending.ContractSpending = spending.ContractSpending.Add(pre.TotalCost)
+			spending.DownloadSpending = spending.DownloadSpending.Add(pre.DownloadSpending)
+			spending.UploadSpending = spending.UploadSpending.Add(pre.UploadSpending)
+			spending.StorageSpending = spending.StorageSpending.Add(pre.StorageSpending)
+		}
+		allSpending = allSpending.Add(spending.ContractSpending).Add(spending.DownloadSpending).Add(spending.UploadSpending).Add(spending.StorageSpending)
```
DavidVorick: Does this end up with a heavy over-estimate for `allSpending`? I think you want to do this final addition outside of the contract loop, otherwise you are double-adding all previous spending every contract.

avahowell:
Updated with latest review comments

Issue #1713, opened 2017-04-16
Delete also *.sia_temp file


betaboot:
As a renter, when i remove an uploaded file (e.g. with 'siac renter delete [path]', the *.sia file is deleted but not the *.sia_temp file. Could they be removed as well?

DavidVorick:
Seems reasonable to me.

PR #2377, opened 2017-09-29
Delete .sia_temp files when .sia file is deleted. #1713

tbenz9:
* Delete .sia_temp files when .sia file is deleted.

This PR fixes issue #1713.

1: Delete the .sia_temp file when it's matching .sia file is deleted.
2: Import the Sia/persist module to get the "tempSuffix" variable.

### DIFF
--- a/persist/persist.go
+++ b/persist/persist.go
@@ -89,7 +89,7 @@ func (sf *safeFile) CommitSync() error {
 // NewSafeFile returns a file that can atomically be written to disk,
 // minimizing the risk of corruption.
 func NewSafeFile(filename string) (*safeFile, error) {
-	file, err := os.Create(filename + "_temp")
+	file, err := os.Create(filename + tempSuffix)
 	if err != nil {
 		return nil, err
 	}
@@ -103,4 +103,18 @@ func NewSafeFile(filename string) (*safeFile, error) {
 	}
 
 	return &safeFile{file, absFilename}, nil
+}
+
+// RemoveFile removes an atomic file from disk, along with any uncommitted
+// or temporary files.
+func RemoveFile(filename string) error {
+	err := os.RemoveAll(filename)
+	if err != nil {
+		return err
+	}
+	err = os.RemoveAll(filename + tempSuffix)
+	if err != nil {
+		return err
+	}
+	return nil
 }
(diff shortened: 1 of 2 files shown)


tbenz9:
Note: I'm a Golang beginner.  Please look closely at my PR as I'm not 100% confident in my skills yet.

lukechampine:
Thanks for contributing! This code looks solid for a beginner, no obvious mistakes.

Your second commit makes me think that perhaps a more general solution is called for here. Since the `persist` package only uses `_temp` files to achieve atomicity, I'd argue that that implementation detail shouldn't "leak." This seems clear from the fact that `tempSuffix` is unexported; the user of `persist` should never have to know about it. As a result, we should hold the `persist` package responsible for cleaning up the `_temp` file.

Pursuant to that, I suggest we add a `RemoveFile` function to the `persist` package, which simply calls `os.RemoveAll` on both the supplied filename and `filename+tempSuffix`. Then we can call that function inside `renter.DeleteFile`.

tbenz9:
@lukechampine Thanks for the suggestions.  I have implemented your ideas as I understand it.  Please take a look and let me know what else should be done.

Sorry about the slightly messy commit history, it's been a while since I've used Git and I'm a little out of practice.

persist/persist.go
```diff
@@ -104,3 +104,16 @@ func NewSafeFile(filename string) (*safeFile, error) {
 
 	return &safeFile{file, absFilename}, nil
 }
+
+// RemoveFile deletes a file from disk
+func RemoveFile(filename string) error {
```
lukechampine: docstring should be:
```
// RemoveFile removes an atomic file from disk, along with any uncommitted or temporary files.
```

lukechampine:
Easiest way to clean up your commits here is to squash them all together with `git reset`. I would do:
```
$ git reset --soft cb1d32b
$ git commit
```
Also, if you include `Fixes #1713` in your commit message, it'll automatically close the issue when I merge it.

tbenz9:
Thanks for the Git tips.  I squashed the commits, added "fixes #1713" keyword, and fixed the docstring.

lukechampine:
perfect, thanks!

PR #2378, opened 2017-09-30
sort and regroup constants

starius:
Another clean-up, following https://github.com/NebulousLabs/Sia/pull/2368#discussion_r141532311
Small tool to find out-of-order constants: https://gist.github.com/7a9a3f8fe1fff4f1e4e933baa0b34f69


lukechampine:
Nice script! Looks like it doesn't check `var` declarations, which is fine, but typically we sort those too when we use them in a `const`-like way (e.g. a lot of our "constants" are `types.Currency` variables)

starius:
I updated the gist to find all vars as well.
The output: https://gist.github.com/e0eafbd5a91556e108f17e2e6a16b48f
What do you think?

lukechampine:
that sure is a lot of unordered vars :sweat_smile:
not worth fixing the ones in tests, but we should probably alphabetize the rest

starius:
I have updated the script to ignore test files and sorted all detected variables.

modules/consensus/consensusdb.go
```diff
@@ -52,14 +49,17 @@ var (
...
-	// SiacoinOutputs is a database bucket that contains all of the unspent
-	// siacoin outputs.
-	SiacoinOutputs = []byte("SiacoinOutputs")
-
 	// FileContracts is a database bucket that contains all of the open file
 	// contracts.
 	FileContracts = []byte("FileContracts")
 
+	prefixDSCO = []byte("dsco_")
+	prefixFCEX = []byte("fcex_")
```
lukechampine: hmm. Might be nitpicking, but we probably want to keep exported and unexported variables separate (i.e. stick these back at the top of the `var` block)

modules/negotiate.go
```diff
@@ -21,44 +21,26 @@ const (
 	// termination, i.e. that the sender wishes to cease communication, but
 	// not due to an error.
 	StopResponse = "stop"
+)
 
+const (
```
lukechampine: why the extra `const` block here? Just separating the `Responses` from the other constants?

modules/wallet/scan.go
```diff
@@ -11,38 +11,38 @@ import (
...
 	switch build.Release {
 	case "dev":
-		return 1e6
+		return 10e3
 	case "standard":
-		return 100e6
+		return 1e6
 	case "testing":
-		return 100e3
+		return 1e3
```
lukechampine: this freaked me out a little but I think it's just a weird diff

modules/consensus/consensusdb.go
starius: I moved prefixDSCO and prefixFCEX back at the top and added the extra `var` block to separate them (otherwise `sorted_consts` complains).

modules/negotiate.go
starius: Yes, just to fix `sorted_consts`'s warning.

modules/wallet/scan.go
starius: I swapped large expression having similarities so diff is really weird.

DavidVorick:
unfortunately the recent changes to the upload code caused a conflict. Shouldn't be too bad to resolve though

starius:
Changes in `modules/renter/consts.go` are no longer needed.

I updated the PR.

Issue #2385, opened 2017-10-02
cleanCloseHandler causes data races


starius:
See https://travis-ci.org/NebulousLabs/Sia/builds/281807220#L678
The goroutine can run away and continue writing to the buffer at the same time [the calling code](https://github.com/golang/go/blob/d1bef43daf850912b52f6f743f8fce31dc7c7481/src/net/http/server.go#L1809) is closing the request (also accessing the buffer).

The pull request is coming.

petabytestorage:
see last step decision on https://github.com/NebulousLabs/Sia/pull/2402 also to close this up i guess

starius:
cleanCloseHandler is still in place in https://github.com/petabytestorage/Sia/commit/e274e4f940091d95ca4b10feb10cf64bda05e6f1

starius:
add "Bug" Label

PR #2386, opened 2017-10-02
Fix cleanCloseHandler

starius:
fix https://github.com/NebulousLabs/Sia/issues/2385 and two issues with the timer


api/api.go
```diff
@@ -119,35 +125,82 @@ func RequirePassword(h httprouter.Handle, password string) httprouter.Handle {
 	}
 }
 
+type cleanResponseWriter struct {
+	impl   http.ResponseWriter
+	m      sync.Mutex
```
DavidVorick: typical name for a mutex is `mu`

api/api.go
```diff
@@ -119,35 +125,82 @@ func RequirePassword(h httprouter.Handle, password string) httprouter.Handle {
 	}
 }
 
+type cleanResponseWriter struct {
+	impl   http.ResponseWriter
```
DavidVorick: not a huge fan of this name, but not sure what would be a good name.

api/api.go
```diff
@@ -119,35 +125,82 @@ func RequirePassword(h httprouter.Handle, password string) httprouter.Handle {
...
+func (c *cleanResponseWriter) Write(buffer []byte) (int, error) {
+	c.m.Lock()
+	defer c.m.Unlock()
+	if c.closed {
+		return 0, errCleanClosed
+	}
+	return c.impl.Write(buffer)
+}
+
+func (c *cleanResponseWriter) WriteHeader(code int) {
```
DavidVorick: All exported functions need docstrings

api/api.go
```diff
@@ -119,35 +125,82 @@ func RequirePassword(h httprouter.Handle, password string) httprouter.Handle {
...
+	impl   http.ResponseWriter
+	m      sync.Mutex
+	closed bool
+}
+
+func (c *cleanResponseWriter) Close() {
+	c.m.Lock()
+	defer c.m.Unlock()
+	c.closed = true
+}
```
DavidVorick: Shouldn't this also call close on `impl`?

api/api.go
```diff
@@ -12,6 +14,10 @@ import (
 	"github.com/julienschmidt/httprouter"
 )
 
+var (
+	errCleanClosed = errors.New("Client closed connection")
```
DavidVorick: I would probably prefer `errResponseWriterClosed`, and then also include one or two words in the error message to indicate that `Close` was called on the response writer prior to whatever call just failed. Object-specific error messages are useful, otherwise you can end up in situations where your error is something like `EOF` and you realize that there are probably 15 different places in your codepath that could have returned `EOF`.

api/api.go
```diff
@@ -119,35 +125,82 @@ func RequirePassword(h httprouter.Handle, password string) httprouter.Handle {
...
+	c.m.Lock()
+	defer c.m.Unlock()
+	c.closed = true
+}
+
+func (c *cleanResponseWriter) Header() http.Header {
+	c.m.Lock()
+	defer c.m.Unlock()
+	if c.closed {
+		return http.Header{}
```
DavidVorick: I know that the server doesn't have a logger right now, but would this be something that we'd want to log? It's probably worth knowing both for devs and for users if something is calling functions on the response writer after it has been closed. Same for all the other methods that check the closed status

api/api_test.go
```diff
@@ -0,0 +1,44 @@
...
+	"net/http/httptest"
+	"testing"
+	"time"
+)
+
+// TestCleanCloseHandler checks that if the handler keeps writing
+// after cleanCloseHandler cancels it, no race condition happens.
+// See https://github.com/NebulousLabs/Sia/issues/2385.
+func TestCleanCloseHandler(t *testing.T) {
+	t.Parallel()
```
DavidVorick: All tests that sleep, and all tests that touch disk in any way at all need to be skipped `if testing.Short()`. General rule of thumb for short tests is that the entire suite of short tests for a package should complete in under 0.5 seconds.

api/api_test.go
```diff
@@ -0,0 +1,44 @@
...
+	req = req.WithContext(ctx)
+	client := &http.Client{}
+	res, err := client.Do(req)
+	if err != nil {
+		t.Fatalf("client.Do: %v", err)
+	}
+	defer res.Body.Close()
+	if _, err := ioutil.ReadAll(res.Body); err == nil {
+		t.Fatalf("Expected to get timeout error")
+	}
```
DavidVorick: Code needs to be spaced out a bit more (needs like 2-3 newlines), and should probably have a few comments as well.

lukechampine:
Honestly I'm in favor of just calling `build.Critical` and not actually closing the connection. It's a developer error, so we should be eliminating the source of the problem instead of patching over it. `cleanCloseHandler` has been its own source of headaches, and at this point it doesn't feel worth it.

starius:
cleanCloseHandler is a workaround for the problem of canceling a running goroutine which has no general solution. Golang provides https://golang.org/pkg/context/ package; it requires adding `ctx` argument to all functions that can block. Currently Sia uses the approach with "stop chans". Ideally we should rewrite it all to context.

lukechampine:
If we're going to go that route, we should start by passing `context.Context` only where we need it most, i.e. in calls that are most likely to take a while to complete. It might be a good idea to test this out on renter endpoints as part of our redesign. Those routes seem like the most likely to take a while anyway.

Issue #2391, opened 2017-10-03
API package - split routes into their own file


DavidVorick:
Right now all of our api routes live alongside a bunch of helper functions, which can make them hard to find. Would like to see them get moved to a file like `api/routes.go`, where they would live alone as the set of routes for that file. Makes the routes easier to find, and easier to see as a whole.

tbenz9:
I think this was merged but not closed.

Recommendation to devs: close

PR #2396, opened 2017-10-04
split api into routes.go issue #2391

petabytestorage:
https://github.com/NebulousLabs/Sia/issues/2391


api/routes.go
```diff
@@ -0,0 +1,194 @@
...
+	"time"
+
+	"github.com/NebulousLabs/Sia/build"
+	"github.com/julienschmidt/httprouter"
+)
+
+// buildHttpRoutes sets up and returns an * httprouter.Router.
+// it connected the Router to the given api using the required
+// parameters: requiredUserAgent and requiredPassword
+func buildHttpRoutes(api *API, requiredUserAgent string, requiredPassword string) *httprouter.Router {
```
DavidVorick: make this a function of the API instead of passing in the api as an argument, otherwise it's not clear that the state of the api is being modified

api/routes.go
```diff
@@ -0,0 +1,194 @@
...
+			return
+		}
+		h(w, req, ps)
+	}
+}
+
+// cleanCloseHandler wraps the entire API, ensuring that underlying conns are
+// not leaked if the rmeote end closes the connection before the underlying
+// handler finishes.
+func cleanCloseHandler(next http.Handler) http.Handler {
```
DavidVorick: per convention, all lowercase functions (non-exported) are put in the file before all upper case functions

avahowell: Should this be defined in `routes.go`? Seem orthogonal to route definition, as do some of the other funcs (`RequireUserAgent` etc)

api/routes.go
avahowell: Alternatively, we could change this func to be stateless and return the router, and assign the API's `router` to the value returned by the func where the API is initialized.

api/routes.go
petabytestorage: done @DavidVorick

api/routes.go
petabytestorage: all good points and either one is a step up in terms of clarity but i would not know which is better.

api/routes.go
petabytestorage: i don't mind wherever it goes and it seems like it should stick with the `api.router = ` piece @avahowell

api/routes_test.go
```diff
@@ -0,0 +1,13 @@
+package api
+
+import (
+	"testing"
+)
+
+func TestBuildHttpRoutes(t *testing.T) {
```
DavidVorick: All functions, including test functions, need a docstring. This one seems a bit odd to me, because I could imagine adjusting the api object and buildRoutes method such that just calling buildHttpRoutes would cause a panic. The full set of routes get built every time you call `newServerTester`, so this might be a situation where we don't need this test.

petabytestorage: doc strings and tests should come after design decisions. i don't know where `api.router =` and cleanCloseHandler etc should live so I don't know how to document / improve this test. i agree it may likely be removed. so my suggestion is the lead core devs take this over and arbitrate.

avahowell: I think `api.router =` and `cleanCloseHandler` etc should live in `api.go`, because it's where the API is initialized. `routes.go` constructs the router, `api.go` constructs the API object.

api/routes.go
DavidVorick: Needs to be a method of the api I believe because when you build the routes you need to understand which modules have been initialized.

DavidVorick:
I am going to disagree with @avahowell about where to put the functions, I think they make sense where they have been placed in this PR because they have been put in the only files where they are used. They are middlewares that wrap the router, therefore that's a good place to have them.

The create router function should be a function of the api, since it is adding a router to the api. It probably should not return the router, since the api router gets set inside of the function, if you need it you can reference the api's router. So we can strip the return value I think.

I don't think we need the test. I'm not opposed to keeping it, but pretty much every other test we have also initializes the api router in some form, if there's a problem with initialization a test will definitely catch it.

This PR is almost ready to go.

fnvhash:
thanks @DavidVorick  all done i believe now

PR #2397, opened 2017-10-04
add ability to set number of hosts in siac

DavidVorick:


lukechampine:
can we make these optional, like they are in the API? I think most people don't want to worry about the renew window. On the other hand, siac is already more low-level than Sia-UI, so maybe it's okay

DavidVorick:
I would have but I don't know how to do optional params in siac

lukechampine:
aight, I will push a commit making them optional

siac/rentercmd.go
```diff
@@ -274,7 +275,15 @@ func rentersetallowancecmd(amount, period string) {
 	if err != nil {
 		die("Could not parse period")
 	}
-	err = post("/renter", fmt.Sprintf("funds=%s&period=%s", hastings, blocks))
+	_, err = strconv.Atoi(hosts)
+	if err != nil {
+		die("Count not parse host count")
+	}
+	_, err = parsePeriod(renewWindow)
```
petabytestorage: can you help me understand why the return value of parsePeriod is ignored here but used with blocks above @DavidVorick  ?  Seems like two different conventions going into the Sprintf but i don't understand the benefit.

PR #2402, opened 2017-10-05
[WIP] add siac optional parameters hosts and renewwindow

petabytestorage:
compiles but not runtime tested yet for https://github.com/NebulousLabs/Sia/pull/2397

n.b. the decision to use the return value of parsePeriod in both places for consistency. we can switch that as appropriate. @DavidVorick   @lukechampine

### DIFF
--- a/siac/rentercmd.go
+++ b/siac/rentercmd.go
@@ -65,7 +65,7 @@ blocks, and a week is 1008 blocks.
 Note that setting the allowance will cause siad to immediately begin forming
 contracts! You should only set the allowance once you are fully synced and you
 have a reasonable number (>30) of hosts in your hostdb.`,
-		Run: wrap(rentersetallowancecmd),
+		Run: rentersetallowancecmd,
 	}
 
 	renterContractsCmd = &cobra.Command{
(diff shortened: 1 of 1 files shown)


siac/rentercmd.go
```diff
@@ -266,24 +266,45 @@ func renterallowancecancelcmd() {
...
+	if len(args) > 3 {
+		renewWindow, err := parsePeriod(args[3])
+		if err != nil {
+			die("Could not parse renew window")
+		}
+		queryString += fmt.Sprintf("&renewwindow=%s", renewWindow)
+	}
+	if len(args) > 4 {
+		cmd.UsageFunc()(cmd)
+		os.Exit(exitCodeUsage)
```
DavidVorick: @lukechampine is this the preferred response when the wrong number of commands are used?

DavidVorick:
I believe that we need to update the helpstring for this command (`./siac renter setallowance -h`) so that it explains the optional commands.

petabytestorage:
thanks for helpstring reminder @DavidVorick  that is done

PR #2403, opened 2017-10-05
fix DustThreshold usage

aiorla:
I've now realized that `tpool.FeeEstimation()` returns hastings/byte and dustThreshold is/was used in a hastings/output context ([ex1](https://github.com/NebulousLabs/Sia/blob/master/modules/wallet/transactionbuilder.go#L97) and [ex2](https://github.com/NebulousLabs/Sia/blob/master/modules/wallet/money.go#L74)). I've used [this estimation](https://github.com/NebulousLabs/Sia/blob/master/modules/wallet/seed.go#L344) to get the consistent formula.

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -1077,7 +1077,7 @@ locked or unlocked.
   "siafundbalance":      "1",    // siafunds, big int
   "siacoinclaimbalance": "9001", // hastings, big int
 
-  "dustthreshold": "1234", // hastings, big int
+  "dustthreshold": "1234", // hastings / byte, big int
 }
 ```
 
--- a/doc/api/Wallet.md
+++ b/doc/api/Wallet.md
@@ -104,9 +104,9 @@ locked or unlocked.
   // increase before any claim transaction is confirmed.
   "siacoinclaimbalance": "9001", // hastings, big int
 
-  // Number of siacoins, in hastings, below which a transaction output cannot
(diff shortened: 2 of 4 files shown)


DavidVorick:
hmm.

I think that `DustThreshold` should be returned in bytes. Different outputs take different amounts of bytes to spend, for example a multisig output is going to require more signatures. So I think this PR is actually the incorrect one, and the existing code is okay.

DavidVorick:
We probably want to update the API documentation though to indicate that `DustThreshold` is a per-byte threshold.

aiorla:
Yes @DavidVorick, a per-byte threshold even makes more sense. :+1:
But I still think that the current code is not right, aside from the incomplete documentation. `dustThreshold()` is/was used in a per-output way (in [ex1](https://github.com/NebulousLabs/Sia/blob/master/modules/wallet/transactionbuilder.go#L97) and [ex2](https://github.com/NebulousLabs/Sia/blob/master/modules/wallet/money.go#L74) it is compared it with `output.Value` to determine if the `output` is dust) and in that PR it was directly changed to a per-byte formula.

aiorla:
Formula change reverted & documentation fixed. (last comment doubts about its current usage remain)

DavidVorick:
Thanks so much @aiorla, we really appreciate cleanup like this.

PR #2405, opened 2017-10-06
Add ContractSet type

lukechampine:
This is an initial design for the new `ContractSet` type, which will live in the `proto` package and provide safe concurrent access to contracts. Once we settle on the design, I can replace all contract manipulation in the `contractor` package with safe `ContractSet` methods.

Previously @DavidVorick and I ran into trouble when trying to design an access pattern that did not lead to deadlocks. As I recall, there were two classes of deadlocks:

1. After `Acquire`ing a contract, code may call out to a helper function that needs to modify the contract, thus calling `Acquire` again, causing deadlock. This was addressed by modifying the helper functions to take the contract as an argument, rather than independently acquiring it.
2. Concurrently trying to `Acquire` a contract while `Return`ing it in a separate goroutine caused a deadlock, because `Acquire` grabs the "outer" lock, which `Return` also needs in order to unlock the "inner" lock. This is addressed in this PR by altering the design so that `Acquire` doesn't hold the outer lock while waiting for the inner lock.

I've added tests to check that all the methods can be called concurrently without deadlocking, but I may have missed an edge case, so please check thoroughly.

Lastly: this code implements an optimization wherein a slice of contracts is cached so that it does not need to be rebuilt in each `Contracts` call. This may be overkill. I implemented it only because rebuilding the slice requires making a deep copy via the (slow) encoding package. Implementing the deep copy manually could make it fast enough that this is not needed. Additionally, `Contracts` should not be called too frequently, so performance isn't critical.

### DIFF
--- a/modules/renter/proto/contractset.go
+++ b/modules/renter/proto/contractset.go
@@ -1,0 +1,159 @@
+package proto
+
+import (
+	"sync"
+
+	"github.com/NebulousLabs/Sia/encoding"
+	"github.com/NebulousLabs/Sia/modules"
+	"github.com/NebulousLabs/Sia/types"
+)
+
+// A safeContract protects a RenterContract with a mutex.
+type safeContract struct {
+	modules.RenterContract
+	mu sync.Mutex
+}
+
+// A ContractSet provides safe concurrent access to a set of contracts. Its
+// purpose is to serialize modifications to individual contracts, as well as
+// to provide operations on the set as a whole.
+type ContractSet struct {
+	contracts map[types.FileContractID]*safeContract
+	mu        sync.Mutex
+
+	// caches (initialized by initCache)
+	flatIDs       []types.FileContractID
+	flatContracts []modules.RenterContract
+	once          sync.Once
+}
+
+// initCache populates the flatIDs and flatContracts fields so that repeated
+// work can be avoided when calling the IDs and Contracts methods. It must be
+// called under lock. initCache should be called by IDs and Contracts if a
+// call to Insert, Return, or Delete has invalidated the cache. This is
+// managed via a sync.Once.
+func (cs *ContractSet) initCache() {
+	cs.flatIDs = make([]types.FileContractID, 0, len(cs.contracts))
+	for id := range cs.contracts {
+		cs.flatIDs = append(cs.flatIDs, id)
+	}
+	cs.flatContracts = make([]modules.RenterContract, 0, len(cs.contracts))
+	for _, sc := range cs.contracts {
+		// construct deep copy, sans MerkleRoots
+		c := sc.RenterContract
+		c.MerkleRoots = nil
+		var contractCopy modules.RenterContract
+		encoding.Unmarshal(encoding.Marshal(c), &contractCopy)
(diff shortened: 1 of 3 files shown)


DavidVorick:
This blocking pattern has interesting implications. If one thread is performing a long running upload on a contract while the contract maintenance thread is trying to delete that contract, that could hang the contract maintenance thread.

I think we are okay right now because the maintenance thread actively invalidates the editor and downloader, but that also means we have to make sure that when they get invalidated, they return the contract locks.

modules/renter/proto/contractset.go
```diff
@@ -0,0 +1,159 @@
...
+// purpose is to serialize modifications to individual contracts, as well as
+// to provide operations on the set as a whole.
+type ContractSet struct {
+	contracts map[types.FileContractID]*safeContract
+	mu        sync.Mutex
+
+	// caches (initialized by initCache)
+	flatIDs       []types.FileContractID
+	flatContracts []modules.RenterContract
+	once          sync.Once
```
DavidVorick: I think the cache here is over-optimization. Definitely increases the complexity, and since you have to reset the cache frequently anyway, I don't think it really saves all that much. I think the deadlocks are resolved by releasing the maplock before you acquire the safeContract locks, which means you avoid nested lock calls.

My preference I think would be to remove the cache and just iterate through the map. You have to hold the map lock while you do the contract copying, but the deadlocks should be resolved.

modules/renter/proto/contractset_test.go
```diff
@@ -0,0 +1,86 @@
...
+		},
+	}
+	wg = sync.WaitGroup{}
+	for _, fn := range funcs {
+		wg.Add(1)
+		go func(fn func()) {
+			defer wg.Done()
+			for i := 0; i < 100; i++ {
+				time.Sleep(time.Duration(fastrand.Intn(100)))
+				fn()
```
avahowell: This is a really neat test 👍

modules/renter/proto/contractset.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+// initCache populates the flatIDs and flatContracts fields so that repeated
+// work can be avoided when calling the IDs and Contracts methods. It must be
+// called under lock. initCache should be called by IDs and Contracts if a
+// call to Insert, Return, or Delete has invalidated the cache. This is
+// managed via a sync.Once.
+func (cs *ContractSet) initCache() {
```
avahowell: I'm not sure that `init` is the clearest prefix for this func. Initialization usually refers to something that is done once, but this func is used to construct and reconstruct the cache in subsequent calls to IDs and Contracts.

lukechampine:
>This blocking pattern has interesting implications. If one thread is performing a long running upload on a contract while the contract maintenance thread is trying to delete that contract, that could hang the contract maintenance thread.

We could modify `Delete` so that it doesn't require the contract's lock. This would mean that `Return` needs to check whether the contract still exists in the set before returning it, and if not, silently drop the contract. I don't really like this -- the current model is explicit and easy to reason about, even if it causes more blocking than is theoretically required.

PR #2411, opened 2017-10-08
release.sh: build with -tags 'netgo'

starius:
Sync it with what `make release-std` does. See https://github.com/NebulousLabs/Sia/pull/2357

Ideally release.sh should invoke `make release-std` but the former uses `go build` and specifies the output file while the later uses `go install` (and installs to `$GOPATH/bin`).

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -36,7 +36,7 @@ for os in darwin linux windows; do
 		if [ "$os" == "windows" ]; then
 			bin=${pkg}.exe
 		fi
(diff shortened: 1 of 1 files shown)


starius:
Also maybe `release.sh` can be moved inside `Makefile`?

DavidVorick:
We try to keep the Makefile pretty clean, I think this would probably too much for it. But this PR LGTM

PR #2417, opened 2017-10-11
make AddressTransactions faster for large wallets

lukechampine:
Previously, `AddressTransactions` iterated through all the transactions tracked by the wallet in order to find those associated with a given address. After this PR, the wallet builds a separate index (`bucketAddrTransactions`) that maps each address seen to the transactions it was seen in. These transactions are then looked up in the existing `bucketProcessedTransactions` bucket. This dramatically improves performance when the address is only associated with a few transactions, which is typically the case. The downside is increased wallet.db size: if the wallet has 1M transactions associated with it, and we assume there are 5 unique addresses in each transaction, with each address associated with 10 transactions, then this new index consumes about 5 * 1M * (32 + 10 * 8) = 2 GB. In reality, the average transaction only contains 3 addresses, and the average address appears in 2 transactions; so a wallet with 1M transactions is more likely to consume 3/2 * 1M * (32 + 2 * 8) = 72MB. (That math needs double-checking, but I'm pretty sure I'm not off by more than one order of magnitude, and even 720MB seems acceptable for such a large wallet.)

This optimization is difficult to benchmark, mostly because the setup (generating thousands transactions) takes a long time. I could fake it by manually generating the transactions (instead of calling `SendSiacoins`), but that risks skewing the results. However, I did perform a manual benchmark of 10,000 transactions, all involving a specific address. When searching for that address, the performance is more or less identical, since you still have to iterate through every transaction. But when searching for an address that is not present in any of the 10k transactions, this PR improves performance by a factor of 134,000. Part of the reason is that, even if you could iterate through all the transactions really quickly, you still need to _decode_ them in order to determine whether they are relevant to an address, and decoding is slow. Using a separate index minimizes the amount of decoding required (you only decode transactions that you care about).

### DIFF
--- a/modules/wallet/database.go
+++ b/modules/wallet/database.go
@@ -19,6 +19,9 @@ var (
 	// chronological order. Only transactions relevant to the wallet are
 	// stored. The key of this bucket is an autoincrementing integer.
 	bucketProcessedTransactions = []byte("bucketProcessedTransactions")
+	// bucketAddrTransactions maps an UnlockHash to the
+	// ProcessedTransactions that it appears in.
+	bucketAddrTransactions = []byte("bucketAddrTransactions")
 	// bucketSiacoinOutputs maps a SiacoinOutputID to its SiacoinOutput. Only
 	// outputs that the wallet controls are stored. The wallet uses these
 	// outputs to fund transactions.
@@ -38,6 +41,7 @@ var (
 
 	dbBuckets = [][]byte{
 		bucketProcessedTransactions,
+		bucketAddrTransactions,
 		bucketSiacoinOutputs,
 		bucketSiafundOutputs,
 		bucketSpentOutputs,
@@ -204,6 +208,47 @@ func dbDeleteSpentOutput(tx *bolt.Tx, id types.OutputID) error {
 	return dbDelete(tx.Bucket(bucketSpentOutputs), id)
 }
 
+// bucketAddrTransactions is special: it's okay if the key doesn't exist, and
+// we need to avoid inserting duplicate slice elements.
+
+func dbPutAddrTransactions(tx *bolt.Tx, addr types.UnlockHash, txns []uint64) error {
+	return dbPut(tx.Bucket(bucketAddrTransactions), addr, txns)
+}
+func dbGetAddrTransactions(tx *bolt.Tx, addr types.UnlockHash) (txns []uint64, err error) {
+	err = dbGet(tx.Bucket(bucketAddrTransactions), addr, &txns)
+	if err == errNoKey {
+		err = nil // no key means no transactions associated yet
+	}
+	return
+}
+func dbAddAddrTransaction(tx *bolt.Tx, addr types.UnlockHash, txn uint64) error {
+	txns, err := dbGetAddrTransactions(tx, addr)
+	if err != nil {
+		return err
+	}
+	for _, i := range txns {
+		if i == txn {
+			return nil
+		}
+	}
+	return dbPutAddrTransactions(tx, addr, append(txns, txn))
+}
+func dbAddProcessedTransactionAddrs(tx *bolt.Tx, pt modules.ProcessedTransaction, txn uint64) error {
+	addrs := make(map[types.UnlockHash]struct{})
+	for _, input := range pt.Inputs {
+		addrs[input.RelatedAddress] = struct{}{}
+	}
+	for _, output := range pt.Outputs {
+		addrs[output.RelatedAddress] = struct{}{}
+	}
+	for addr := range addrs {
+		if err := dbAddAddrTransaction(tx, addr, txn); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
 // bucketProcessedTransactions works a little differently: the key is
 // meaningless, only used to order the transactions chronologically.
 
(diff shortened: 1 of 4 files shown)


modules/wallet/persist.go
```diff
@@ -71,6 +74,17 @@ func (w *Wallet) openDB(filename string) (err error) {
...
+		// build the bucketAddrTransactions bucket if necessary
+		if buildAddrTxns {
+			it := dbProcessedTransactionsIterator(tx)
+			for it.next() {
+				index, pt := it.key(), it.value()
+				if err := dbAddProcessedTransactionAddrs(tx, pt, index); err != nil {
+					return err
+				}
+			}
+		}
```
lukechampine: not sure how long this would take on a large wallet. It probably won't be noticeable to anyone with less than 10k transactions. We should include a warning that upgrading the wallet may be slow if you have lots of transactions.

lukechampine:
Added benchmarks:
```
BenchmarkAddressTransactions/indexed-4         	    1000	   2362481 ns/op
BenchmarkAddressTransactions/indexed-nosync-4  	  100000	     10750 ns/op
BenchmarkAddressTransactions/unindexed-4       	     100	  19454344 ns/op
```
So the index speeds up the call by a factor of 10 when there are 10k transactions. However, it's clear that the runtime is dominated by the `fsync` which happens at the beginning of the `AddressTransactions` method to ensure consistency. Without the `fsync`, we gain a factor of 2000.

We could eliminate the `fsync` calls by only calling `w.dbTx.Commit` if there were changes to commit. Sadly, there doesn't seem to be any way to tell from the `bolt.Tx` itself whether any writes have been performed. So we could either augment the wallet to set a boolean whenever it writes to the db, or we can augment our bolt fork to track this internally. I'm in favor of the latter because it would allow us to trivially extend this optimization to our other databases.

modules/wallet/database.go
```diff
@@ -204,6 +208,47 @@ func dbDeleteSpentOutput(tx *bolt.Tx, id types.OutputID) error {
 	return dbDelete(tx.Bucket(bucketSpentOutputs), id)
 }
 
+// bucketAddrTransactions is special: it's okay if the key doesn't exist, and
+// we need to avoid inserting duplicate slice elements.
+
+func dbPutAddrTransactions(tx *bolt.Tx, addr types.UnlockHash, txns []uint64) error {
+	return dbPut(tx.Bucket(bucketAddrTransactions), addr, txns)
+}
+func dbGetAddrTransactions(tx *bolt.Tx, addr types.UnlockHash) (txns []uint64, err error) {
```
DavidVorick: Need comments on each function indicating that they are helper functions

modules/wallet/database.go
```diff
@@ -204,6 +208,47 @@ func dbDeleteSpentOutput(tx *bolt.Tx, id types.OutputID) error {
...
+		return err
+	}
+	for _, i := range txns {
+		if i == txn {
+			return nil
+		}
+	}
+	return dbPutAddrTransactions(tx, addr, append(txns, txn))
+}
+func dbAddProcessedTransactionAddrs(tx *bolt.Tx, pt modules.ProcessedTransaction, txn uint64) error {
```
DavidVorick: these functions should have docstrings too, they are definitely nontrivial

modules/wallet/database.go
```diff
@@ -204,6 +208,47 @@ func dbDeleteSpentOutput(tx *bolt.Tx, id types.OutputID) error {
...
+		err = nil // no key means no transactions associated yet
+	}
+	return
+}
+func dbAddAddrTransaction(tx *bolt.Tx, addr types.UnlockHash, txn uint64) error {
+	txns, err := dbGetAddrTransactions(tx, addr)
+	if err != nil {
+		return err
+	}
+	for _, i := range txns {
```
DavidVorick: This could get bad if we end up with a lot of transactions coming to the same address. Not sure what bittrex's load looks like, and at least it's better than doing a linear operation using a disk iterator, but we could put a binary search here too.

We want to handle lopsided situations just as well as normal ones.

modules/wallet/database.go
```diff
@@ -237,29 +286,43 @@ func dbDeleteLastProcessedTransaction(tx *bolt.Tx) error {
...
+	indexBytes := make([]byte, 8)
+	binary.BigEndian.PutUint64(indexBytes, index)
+	val := tx.Bucket(bucketProcessedTransactions).Get(indexBytes)
+	err = encoding.Unmarshal(val, &pt)
+	if err != nil {
+		// COMPATv1.2.1: try decoding into old transaction type
+		var oldpt v121ProcessedTransaction
+		err = encoding.Unmarshal(val, &oldpt)
+		pt = convertProcessedTransaction(oldpt)
+	}
```
DavidVorick: We need compatibility code with v1.2.1? Wasn't that like 2 full releases ago?

modules/wallet/persist.go
DavidVorick: That's okay. People with that many transactions will be used to it being slow

modules/wallet/database.go
DavidVorick: oh I see. This is needed for people who upgrade directly from e.g. v1.0.4 -> v1.3.1

modules/wallet/database.go
lukechampine: It's linear, but decoding the transactions is also linear, and its constant factor is waaaay larger. If we need to optimize this call, the first thing to target is the decoding.

ChrisSchinnerl:
Since there already exists an integration test for `TestAddressTransaction` which checks if transactions are successfully added to the new index, I created `TestAddressTransactionRevertedBlock` to check if transactions are correctly removed from the new Index if the mined block is reverted.

PR #2418, opened 2017-10-12
extend explorer api with block.ID() and optional hexblock field

petabytestorage:
block.ID() fixes "missing block" explorer issue with blocks with no transaction. "hexblock" optional return field (with "hexblock"="true" optional form parameter) provides verbatim hex encoded binary block to enable external applications and developer scaffolding.


api/explorer.go
```diff
@@ -188,10 +193,17 @@ func (api *API) buildExplorerBlock(height types.BlockHeight, block types.Block)
 		panic("incorrect request to buildExplorerBlock - block does not exist")
 	}
 
+	hex_block := ""
```
DavidVorick: more idiomatic to do `var hex_block string`

Project consistently and strictly uses camel case throughout for variable names, need to do the same here.

petabytestorage:
hex_block_enabled -> hexBlockEnabled  and hex_block -> hexBlock  ty @DavidVorick

api/explorer.go
```diff
@@ -214,7 +226,8 @@ func (api *API) explorerBlocksHandler(w http.ResponseWriter, req *http.Request,
 		return
 	}
 	WriteJSON(w, ExplorerBlockGET{
-		Block: api.buildExplorerBlock(height, block),
+		Block: api.buildExplorerBlock(height, block,
+			req.FormValue("hexblock") == "true"),
```
DavidVorick: Sorry I didn't comment on this in the first pass. Wasn't sure what we wanted here, but after discussing with Luke, we agree that the call should return an error if `hexblock` is set to anything besides "true" or "false" - strict api reduces the  liklihood of dev mistakes, and coherent errors around incorrect input make it easier to find the source of the mistake.

api/explorer.go
```diff
@@ -271,7 +284,8 @@ func (api *API) explorerHashHandler(w http.ResponseWriter, req *http.Request, ps
 	if exists {
 		WriteJSON(w, ExplorerHashGET{
 			HashType: "blockid",
-			Block:    api.buildExplorerBlock(height, block),
+			Block: api.buildExplorerBlock(height, block,
+				req.FormValue("hexblock") == "true"),
```
DavidVorick: Same here.

DavidVorick:
LGTM except for the final comments. Looking for review from @lukechampine

api/explorer.go
```diff
@@ -188,10 +193,17 @@ func (api *API) buildExplorerBlock(height types.BlockHeight, block types.Block)
 		panic("incorrect request to buildExplorerBlock - block does not exist")
 	}
 
+	var hexBlock string
+	if hexBlockEnable {
+		hexBlock = fmt.Sprintf("%x", encoding.Marshal(block)[:])
```
lukechampine: `encoding.Marshal` returns a `[]byte` so the `[:]` slicing is unnecessary

api/explorer.go
```diff
@@ -188,10 +193,17 @@ func (api *API) buildExplorerBlock(height types.BlockHeight, block types.Block)
...
+	}
+
 	return ExplorerBlock{
 		MinerPayoutIDs: mpoids,
 		Transactions:   etxns,
 		RawBlock:       block,
+		HexBlock:       hexBlock,
+		BlockId:        fmt.Sprintf("%x", encoding.Marshal(block.ID())[:]),
```
lukechampine: ditto

api/explorer.go
lukechampine: also, I think we slightly prefer:
```go
import "encoding/hex"

hex.EncodeToString(encoding.Marshal(block))
```

api/explorer.go
petabytestorage: Makes sense to me and I will do it. I copied this pattern from `api/wallet.go` where it appears twice and does only look for "true". I also guessed that the `wallet.go` would be more strict/secure than an explorer module. So I'm going to suggest the wallet.go "force" field should be considered as well if you want to use this pattern for example to avoid misleading other devs into copying around or learning a bad pattern. @DavidVorick

petabytestorage: Can you give me an example of the new pattern you and luke want in the pre-existing source code please?  I'm looking through all the boolean flags in the system and not finding one "good" one according to your description.

api/explorer.go
```diff
@@ -20,6 +22,8 @@ type (
 		MinerPayoutIDs []types.SiacoinOutputID `json:"minerpayoutids"`
 		Transactions   []ExplorerTransaction   `json:"transactions"`
 		RawBlock       types.Block             `json:"rawblock"`
+		HexBlock       string                  `json:"hexblock"`
+		BlockId        string                  `json:"blockid"`
```
lukechampine: should be `BlockID`

api/explorer.go
```diff
@@ -188,10 +194,17 @@ func (api *API) buildExplorerBlock(height types.BlockHeight, block types.Block)
 		panic("incorrect request to buildExplorerBlock - block does not exist")
 	}
 
+	var hexBlock string
+	if hexBlockEnable {
+		hexBlock = hex.EncodeToString(encoding.Marshal(block))
+	}
```
lukechampine: for structs with optional fields, I generally prefer:
```go
b := ExplorerBlock{
    // required fields here...
}
if hexBlockEnable {
    b.HexBlock = hex.EncodeToString(encoding.Marshal(block))
}
return b
```

api/explorer.go
lukechampine: as per our discord discussion, we'll punt this to a follow-up PR

petabytestorage:
all done @DavidVorick  and @lukechampine  thanks for the idiomatic go

api/explorer.go
```diff
@@ -20,6 +22,8 @@ type (
 		MinerPayoutIDs []types.SiacoinOutputID `json:"minerpayoutids"`
 		Transactions   []ExplorerTransaction   `json:"transactions"`
 		RawBlock       types.Block             `json:"rawblock"`
+		HexBlock       string                  `json:"hexblock"`
+		BlockID        string                  `json:"blockid"`
 
 		modules.BlockFacts
```
lukechampine: hmm, changing to `BlockID` actually caused tests to start failing, because `modules.BlockFacts` has an embedded `BlockID` field.

Is it necessary to have a `BlockID string` field if we already have it as a `[32]byte`? If so, maybe call it `HexBlockID`?

petabytestorage: @lukechampine  when i remove the field i get:
```
➜  Sia git:(explorer-patch) ✗ make
go install -tags 'netgo' -ldflags='-s -w' ./api ./build ./compatibility ./crypto ./encoding ./modules ./modules/consensus ./modules/explorer ./modules/gateway ./modules/host ./modules/host/contractmanager ./modules/renter ./modules/renter/contractor ./modules/renter/hostdb ./modules/renter/hostdb/hosttree ./modules/renter/proto ./modules/miner ./modules/wallet ./modules/transactionpool ./persist ./siac ./siad ./sync ./types
# github.com/NebulousLabs/Sia/api
api/explorer.go:201: unknown field 'BlockID' in struct literal of type ExplorerBlock
Makefile:75: recipe for target 'release-std' failed
make: *** [release-std] Error 2
```
therefore i renamed it to HexBlockID however i left the json named the standard way because to me it seems cleaner. please feel free to adjust again if you think there is a cleaner way to handle this.

api/explorer.go
```diff
@@ -188,13 +194,18 @@ func (api *API) buildExplorerBlock(height types.BlockHeight, block types.Block)
 		panic("incorrect request to buildExplorerBlock - block does not exist")
```
ChrisSchinnerl: In my opinion we shouldn't crash if a user supplies an invalid height but return an error we write to the http response. What do you think @lukechampine @DavidVorick

api/explorer.go
```diff
@@ -20,6 +22,8 @@ type (
 		MinerPayoutIDs []types.SiacoinOutputID `json:"minerpayoutids"`
 		Transactions   []ExplorerTransaction   `json:"transactions"`
 		RawBlock       types.Block             `json:"rawblock"`
+		HexBlock       string                  `json:"hexblock"`
+		HexBlockID     string                  `json:"blockid"`
```
ChrisSchinnerl: `HexBlockID` can probably be `BlockID` and have `types.BlockID` as a type.
`types.BlockID` implements `String` and `MarshalJSON` which should implicitly convert it to a hex string in `WriteJSON`.

api/explorer.go
```diff
@@ -188,13 +194,18 @@ func (api *API) buildExplorerBlock(height types.BlockHeight, block types.Block)
 		panic("incorrect request to buildExplorerBlock - block does not exist")
 	}
 
-	return ExplorerBlock{
+	b := ExplorerBlock{
 		MinerPayoutIDs: mpoids,
 		Transactions:   etxns,
 		RawBlock:       block,
+		HexBlockID:     hex.EncodeToString(encoding.Marshal(block.ID())),
```
ChrisSchinnerl: As mentioned above this can probably just be a `type.BlockID` instead of having to encode yourself. That means it would be `BlockID: block.ID(),`

api/explorer.go
```diff
@@ -188,13 +194,18 @@ func (api *API) buildExplorerBlock(height types.BlockHeight, block types.Block)
...
 
 		BlockFacts: facts,
 	}
+	if hexBlockEnable {
+		b.HexBlock = hex.EncodeToString(encoding.Marshal(block))
```
ChrisSchinnerl: Could you implement something like the following in `types/block.go`?
```
// String returns a hex encoded marshalled Block.
func (b Block) String() string {                                                                                                                                                                                  
    return hex.EncodeToString(encoding.Marshal(b))                                                                                                                                                                                     
} 
```
That way you can simply do `b.HexBlock = block.String()` here.

Issue #2414, opened 2017-10-09
use golang.org/x/crypto/ssh/terminal instead of github.com/bgentry/speakeasy


starius:

avahowell:
Done in #2419

PR #2419, opened 2017-10-12
replace speakeasy with golang.org/x/crypto/ssh/terminal

avahowell:
See #2414

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ dependencies:
 	go get -u github.com/kardianos/osext
 	go get -u github.com/inconshreveable/mousetrap
 	# Frontend Dependencies
-	go get -u github.com/bgentry/speakeasy
+	go get -u golang.org/x/crypto/ssh/terminal
 	go get -u github.com/spf13/cobra/...
 	# Developer Dependencies
 	go install -race std
--- a/cmd/siac/main.go
+++ b/cmd/siac/main.go
@@ -9,7 +9,6 @@ import (
 	"os"
 	"reflect"
 
-	"github.com/bgentry/speakeasy"
 	"github.com/spf13/cobra"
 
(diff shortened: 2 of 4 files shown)


siac/walletcmd.go
```diff
@@ -507,10 +517,12 @@ func walletunlockcmd() {
...
 		}
 	}
-	password, err := speakeasy.Ask("Wallet password: ")
+	fmt.Print("Wallet password: ")
+	passwordBytes, err := terminal.ReadPassword(0)
 	if err != nil {
 		die("Reading password failed:", err)
 	}
+	password = string(passwordBytes)
 	qs := fmt.Sprintf("encryptionpassword=%s&dictonary=%s", password, "english")
```
lukechampine: I think the cast is actually unnecessary here; `%s` should format a `[]byte` as a `string`

lukechampine:
Perhaps we should define a helper function to match the speakeasy API:

```go
func passwordPrompt(prompt string) (string, error) {
    fmt.Print(prompt)
    pw, err := terminal.ReadPassword(0)
    return string(pw), err
}
```

avahowell:
Perhaps, but then we would likely have to duplicate that helper since this functionality is used in `siad` and `siac`. Probably not a big deal. Another thing to consider after doing some runtime testing is that `terminal` doesn't print a newline after reading the password, so to perfectly mirror the functionality the helper should be:

```go
func passwordPrompt(prompt string) (string, error) {
    fmt.Print(prompt)
    pw, err := terminal.ReadPassword(0)
    fmt.Println()
    return string(pw), err
}
```

avahowell:
Updated with the suggested helper function

lukechampine:
LGTM, but we need to update the `Makefile` now

avahowell:
Rebasedd and updated the makefile.

PR #2422, opened 2017-10-13
Minor documentation correction WRT Block ID in doc/Consensus.md

pedrorivera:
The timestamp bytes are not involved in the merkle root calculation as far as I can tell.


lukechampine:
good catch, thanks!

PR #2423, opened 2017-10-13
Modify the wallets Transactions function to use binary search

ChrisSchinnerl:


modules/wallet/transactions.go
```diff
@@ -86,6 +89,18 @@ func (w *Wallet) Transaction(txid types.TransactionID) (pt modules.ProcessedTran
...
+func decodeProcessedTransaction(ptBytes []byte, pt *modules.ProcessedTransaction) bool {
+	err := encoding.Unmarshal(ptBytes, pt)
+	if err != nil {
+		// COMPATv1.2.1: try decoding into old transaction type
+		var oldpt v121ProcessedTransaction
+		err = encoding.Unmarshal(ptBytes, &oldpt)
+		*pt = convertProcessedTransaction(oldpt)
+	}
+	return err == nil
+}
```
lukechampine: this should live in `database.go`, and should be called by the other `ProcessedTransaction` functions.

I'm also wondering if we should optimize the decoding a bit. We can get the `ConfirmationHeight` without decoding all the inputs and outputs, which is nice. But I'm not sure if it's going to be a bottleneck in practice. We'll see what the profiling suggests.

modules/wallet/transactions.go
```diff
@@ -101,18 +116,56 @@ func (w *Wallet) Transactions(startHeight, endHeight types.BlockHeight) (pts []m
...
+
+	// Get the largest key present in the processed transactions bucket
+	lastKeyBytes, _ := cursor.Last()
+	if lastKeyBytes == nil {
+		// bucket is empty
+		return
+	}
+
+	// Decode it to the the largest key
+	lastKey := binary.BigEndian.Uint64(lastKeyBytes)
```
lukechampine: I think you can do:
```go
lastKey := w.dbTx.Bucket(bucketProcessedTransactions).Sequence()
```

modules/wallet/transactions.go
```diff
@@ -101,18 +116,56 @@ func (w *Wallet) Transactions(startHeight, endHeight types.BlockHeight) (pts []m
...
+
+	// Conduct the binary search
+	sort.Search(int(lastKey), func(i int) bool {
+		// Create the key for the index
+		keyBytes := make([]byte, 8)
+		binary.BigEndian.PutUint64(keyBytes, uint64(i))
```
lukechampine: more efficient if we move the `keyBytes` allocation outside the loop

modules/wallet/transactions.go
```diff
@@ -101,18 +116,56 @@ func (w *Wallet) Transactions(startHeight, endHeight types.BlockHeight) (pts []m
...
+
+		// Retrieve the processed transaction
+		key, ptBytes := cursor.Seek(keyBytes)
+		if key == nil {
+			panic("Failed to retrieve processed Transaction by key")
```
lukechampine: all panics need to be wrapped in `build.DEBUG` checks

modules/wallet/transactions.go
```diff
@@ -101,18 +116,56 @@ func (w *Wallet) Transactions(startHeight, endHeight types.BlockHeight) (pts []m
...
+		}
+
+		// Decode the transaction
+		if success := decodeProcessedTransaction(ptBytes, &pt); !success {
+			panic("Failed to decode the processed transaction")
```
lukechampine: I suppose we could also wrap the `sort.Search` call in a function that would catch the panic. I'm not sure how else we could break out of it and return an error.

modules/wallet/transactions.go
```diff
@@ -101,18 +116,56 @@ func (w *Wallet) Transactions(startHeight, endHeight types.BlockHeight) (pts []m
...
+		}
+
+		return pt.ConfirmationHeight >= startHeight
+	})
+
+	// Gather all transactions until endHeight is reached
+	for pt.ConfirmationHeight <= endHeight {
+		if !(pt.ConfirmationHeight < startHeight) {
```
lukechampine: is this necessary? I guess it's ok to keep it as a sanity check. Maybe:
```go
if build.DEBUG && pt.ConfirmationHeight < startHeight {
    build.Critical("wallet processed transactions are not sorted")
}
pts = append(pts, pt)
```

lukechampine:
This looks pretty good, but it will probably conflict with #2417. Once we merge that, we can fixup this PR

modules/renter/contractor/contractor_test.go
```diff
@@ -394,6 +395,8 @@ func TestAllowanceSpending(t *testing.T) {
 	}
 	newReportedSpending := c.PeriodSpending()
 	if reflect.DeepEqual(newReportedSpending, reportedSpending) {
+		log.Print(reportedSpending)
+		log.Print(newReportedSpending)
```
lukechampine: I think this is failing because of the comparison of `types.Currency`, which contains a slice. If you print with `log.Printf("%+v", reportedSpending)` you may be able to see the precise difference. But at any rate, it seems like we need to write an `Allowance.Equals(Allowance)` method to manually compare allowances.

ChrisSchinnerl:
Seems like changing the locking order in this [PR](https://github.com/NebulousLabs/Sia/pull/2499) didn't fix the random `TestRemoteFileRepair` bug.

DavidVorick:
Is this guaranteed to return every transaction at a given height? Seems like the binary search could pick the middle transaction of a block if there are a bunch of blocks at a given height.

ChrisSchinnerl:
Should be fine. According to the documentation it finds the smallest index for which the condition is true.
https://golang.org/pkg/sort/#Search

Issue #1557, opened 2017-01-22
Renter shows >100% upload progress


vevarm:
sometimes sia is more than 100% done uploading files

vevarm:
Sia is 103% done, that is great.
...
58.70 MB  Yes       *103.12%*        4.12  Yes       backup-20161019-filelist-0007.tar
...

vevarm:
Seems this is very common. I have a script count completness ( for the purpose of upload status). 

x@nicke:~$ bash sia-count.sh 
     16     102.50%
     25     103.91%
     29     102.34%
     31     102.08%
     34     104.17%
     87     104.69%
     92     101.56%
    209     100.00%
    301     106.25%
    782     103.12%

DavidVorick:
We are looking into the issue, among a couple of other issues that we are investigating.

Though, it's not a super high priority because having an extra 3-5% uploaded does not seem like a huge deal to me.

vevarm:
agree 
not prio

vevarm:
https://m.youtube.com/watch?v=hW008FcKr3Q

i just had to link it

DavidVorick:
I have looked into this more, and what's happening is that Sia is replacing hosts that go offline. Basically, while uploading, some of the hosts have already disappeared, so you've had to replace them. The total uploaded counter does not decrease though. This I guess is a good way then to measure the total churn on your files.

106.25% = over the lifetime of the file 6.25% has disappeared, and you've had to repair it.

The redundancy is following something similar, though it's a less useful metric.

vevarm:
Ok cool I find this value intressting, however ... 

This is totally unintuative for most, this needs to be presented in a way all will understand. But I have no idea how.

tbenz9:
Recommendation to devs: label "Bug"

avahowell:
Fixed in #2429

PR #2429, opened 2017-10-16
modify uploadProgress to report at most 100%

avahowell:
See #1557.


lukechampine:
I don't think we decided whether this is a bug or a feature. Personally I'm in favor of merging this; `uploadProgress` shouldn't also be responsible for accurately reporting the file's redundancy. In my view, `uploadProgress` should move (perhaps monotonically) from 0% to 100%, and remain there.

Issue #2428, opened 2017-10-16
Provide local path of renter files in API


KaiRo-at:
In our usage of Sia in the Minebox backup procedure, we'd like to find out the local path for files uploaded with the Sia renter, esp. to make sure we don't delete local copies of files that aren't fully uploaded yet.
Unfortunately, while renter.json has those references to local files ("RepairPath"), the siad API does not expose them. It would be great if that field would be added to the info we can get for renter files via the API.

tbenz9:
Recommendation to devs: "Feature Request" "Programming Task"

tbenz9:
This can be closed.

PR #2431, opened 2017-10-16
return local filepath in renter api

avahowell:
See #2428. Adds a new `localpath` string to `modules.FileInfo`.

### DIFF
--- a/modules/renter/files.go
+++ b/modules/renter/files.go
@@ -229,8 +229,14 @@ func (r *Renter) FileList() []modules.FileInfo {
 	for _, f := range files {
 		f.mu.RLock()
 		renewing := true
+		var localPath string
+		tf, exists := r.tracking[f.name]
+		if exists {
+			localPath = tf.RepairPath
+		}
 		fileList = append(fileList, modules.FileInfo{
 			SiaPath:        f.name,
+			LocalPath:      localPath,
 			Filesize:       f.size,
 			Renewing:       renewing,
 			Available:      f.available(isOffline),
--- a/doc/API.md
+++ b/doc/API.md
@@ -858,6 +858,7 @@ lists the status of all files.
   "files": [
     {
       "siapath":        "foo/bar.txt",
+      "localpath":      "/home/foo/bar.txt",
       "filesize":       8192, // bytes
       "available":      true,
       "renewing":       true,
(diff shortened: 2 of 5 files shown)


modules/renter/files.go
```diff
@@ -229,8 +229,14 @@ func (r *Renter) FileList() []modules.FileInfo {
 	for _, f := range files {
 		f.mu.RLock()
 		renewing := true
+		var localPath string
```
petabytestorage: maybe this way https://github.com/NebulousLabs/Sia/pull/2418/files/0b81134071bb7d8d2de1a546e1829d01137de3ec#r144727139

petabytestorage: and fwiw i think the convention is probably better the way we use local variables because it's more maintainable. the elided local costs an explicit `b.` on each case so doesn't really save much and the `b.` itself is actually a bit too short a name to be useful in many stylistic conventions. the parallel structure enabled by always putting the explicit literally constructed struct at the end of functions increased readability and clarity a lot for me. the lexical scoping that luke suggests here smacks of tight linkage. so i would choose the other way if i were in charge. but in the name of consistency i just follow leaders on all matters of aesthetic that are not horribly offensive since consistency trumps everything in the land of bugs.

lukechampine: I admit that I'm unsure if one style is always better than the other. I agree that constructing the whole literal at the end of the loop/function is nice, but I also like avoiding temporary variables, and I don't like how the current style makes it seem like `LocalPath` is always specified, even though it may in fact be an empty string.

I guess this is one area where Go's lack of ternary expressions is painful. In that vein, you could do something like:
```go
fileList = append(fileList, modules.FileInfo{
    LocalPath: func() string {
        if tf, ok := r.tracking[f.name]; ok {
            return tf.RepairPath
        }
        return ""
    })(),
})
```
but that would obviously be a crime against readability. :P 

Actually, in this case, we can pull a trick that satisfies everyone:
```go
fileList = append(fileList, modules.FileInfo{
    LocalPath: r.tracking[f.name].RepairPath,
})
```
This works because `r.tracking` maps to `trackedFile` values (not pointers), and so if the lookup fails, you get a zero-valued `trackedFile` whose `RepairPath` is `""`. I like this approach, but we should still drop a comment indicating that `RepairPath` will be empty for files not in `r.tracking`.

petabytestorage: @lukechampine  this seems like it requires a lot more knowledge at once to understand. Sia-specific knowledge as well. Why make the learning curve steeper in the code?  What possible purpose to lower  clarity?

PR #2433, opened 2017-10-18
Add some details for the installation.

roipoussiere:


lukechampine:
Thanks for contributing. I think our README is fine as-is, but I bumped the version requirement to Go 1.9 and included a note about setting the GOPATH.

Issue #2364, opened 2017-09-26
Prices API reports wildly different numbers from call to call


mtlynch:
I've noticed that the `siac renter prices` output varies wildly from call to call. For example, I just ran the command:

```
$ $GOPATH/bin/siac renter prices
Renter Prices (estimated):
  Fees for Creating a Set of Contracts:   103.5 SC
  Download 1 TB:                          45.24 SC
  Store 1 TB for 1 Month:                 258.5 SC
  Upload 1 TB:                            57.13 SC
```

Then ran it again a second later:

```
$ $GOPATH/bin/siac renter prices
Renter Prices (estimated):
  Fees for Creating a Set of Contracts:   101 SC
  Download 1 TB:                          34.42 SC
  Store 1 TB for 1 Month:                 241.7 SC
  Upload 1 TB:                            20.65 SC
```

* **Expected behavior**: `siac renter prices` reports consistent prices when called a few seconds apart.
* **Actual behavior**: `siac renter prices` reports prices that are wildly different between calls, in some cases up to 276% higher.

I'm guessing this is due to the fact that there's randomness in picking host contracts. Would it be possible for this to be random, but repeatable so that given the same set of hosts with the same offerings, the client chooses the same random set of contracts?

tbenz9:
@mtlynch I'm not sure exactly what the purpose of this Issue is.  If you're asking a question I would recommend asking it on Discord.  If you're making a feature-request I would recommend coming up with some kind of suggested solution (maybe even submitting a Pull Request).

mtlynch:
@tbenz9, thanks for triaging these.

> I'm not sure exactly what the purpose of this Issue is. If you're asking a question I would recommend asking it on Discord. If you're making a feature-request I would recommend coming up with some kind of suggested solution (maybe even submitting a Pull Request).

This is not a question or a feature request. It is a bug report. The behavior I reported is unexpected and degrades user experience, so it's a bug.

>Just to clarify, I agree that the renter prices are a little strange. But i'm not sure what you hope to accomplish with this Issue, and since no one has commented in 11 days I'd guess they are confused as well.

I don't think that means much. There are 162 open bugs and until today, nobody's been triaging them, so lots of them have no comment from the dev team.

tbenz9:
[...]

PR #2434, opened 2017-10-18
make renter prices estimation less volatile

avahowell:
This PR addresses the UX concerns noted in #2364. Previously, hosts would be re-sampled on each call, causing confusing volatility. Now, the renter will cache renter price estimations, only resampling if more than a minute has passed since the last price estimation call.

### DIFF
--- a/modules/renter/consts.go
+++ b/modules/renter/consts.go
@@ -56,11 +56,4 @@ var (
 		Standard: time.Second * 61,
 		Testing:  time.Second,
 	}).(time.Duration)
-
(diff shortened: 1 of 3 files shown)


DavidVorick:
I'd probably reset the cache each time you get a block. That's a pretty natural feeling for a user, and it's also really easy to reset during testing and development.

Definitely in favor of the general idea of caching the prices stuff to reduce perceived volatility.

avahowell:
Updated per @DavidVorick's suggestion of having the price only re-estimate once per block.

PR #2444, opened 2017-10-24
requestNodes: only save if new nodes were added

lukechampine:
this is mostly relevant to antfarm setups, where you have a small number of nodes that repeatedly bounce their addresses back and forth to each other. After a few minutes, none of the nodes will ever receive another unique address, but they still save the node list over and over again, incurring needless IO.
This was discovered using the `inotifywait` command mentioned in #2362.


tbenz9:
@avahowell Are you waiting on something before you merge this?

lukechampine:
Typically only @DavidVorick and I have the final say as to whether a PR is merged (unless it's a trivial change like a typo).

PR #2446, opened 2017-10-24
better error message when api is loading

avahowell:
## before:
before
## after:
after

This should give users and API developers more helpful feedback when Sia is loading modules.

### DIFF
--- a/cmd/siad/server.go
+++ b/cmd/siad/server.go
@@ -14,7 +14,6 @@ import (
 	"os"
 	"path"
 	"path/filepath"
-	"reflect"
 	"runtime"
 	"sort"
 	"strings"
@@ -48,9 +47,16 @@ type (
 		httpServer    *http.Server
 		listener      net.Listener
 		config        Config
-		moduleClosers []io.Closer
+		moduleClosers []moduleCloser
 		api           http.Handler
 		mu            sync.Mutex
+	}
+
+	// moduleCloser defines a struct that closes modules, defined by a name and
+	// an underlying io.Closer.
+	moduleCloser struct {
+		name   string
+		closer io.Closer
 	}
 
 	// SiaConstants is a struct listing all of the constants in use.
@@ -143,6 +149,12 @@ func (rs byVersion) Less(i, j int) bool {
 	// we want the higher version number to reported as "less" so that it is
 	// placed first in the slice
 	return build.VersionCmp(rs[i].version(), rs[j].version()) >= 0
+}
+
+// Close implements the Close method for moduleCloser.
+func (mc *moduleCloser) Close() error {
+	fmt.Printf("Closing %v...\n", mc.name)
+	return mc.closer.Close()
 }
 
 // latestRelease returns the latest non-LTS release, given a set of arbitrary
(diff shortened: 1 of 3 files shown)


cmd/siad/server.go
```diff
@@ -366,6 +368,21 @@ func (srv *Server) daemonHandler(password string) http.Handler {
...
+	api.WriteError(w, api.Error{"siad is not ready. please wait for siad to finish loading."}, http.StatusServiceUnavailable)
+}
+
+// AttachAPI attaches an initialized Sia API to the server.
+func (srv *Server) AttachAPI(a *api.API) {
+	mux := http.NewServeMux()
+	mux.Handle("/daemon/", api.RequireUserAgent(srv.daemonHandler(srv.requiredPassword), srv.requiredUserAgent))
+	mux.Handle("/", a)
+	srv.mux = mux
+	srv.httpServer.Handler = mux
```
lukechampine: isn't this racey?

avahowell: It may be. I'll write a test for this behavior that uses the race detector

avahowell:
Latest commits fix the data race and restructure the code slightly so the daemon `Server` is responsible for loading and closing the modules.

cmd/siad/server.go
```diff
@@ -422,9 +578,19 @@ func (srv *Server) Serve() error {
...
 	// Close the listener, which will cause Server.Serve() to return.
 	if err := srv.listener.Close(); err != nil {
-		return err
+		errs = append(errs, err)
 	}
-	return nil
+	// Close all of the modules
+	for i := len(srv.moduleClosers) - 1; i >= 0; i-- {
+		m := srv.moduleClosers[i]
+		fmt.Printf("Closing %v...\n", reflect.TypeOf(m))
```
lukechampine: does this actually print the right thing?

avahowell: Example output:

```
Caught stop signal, quitting...
Closing *renter.Renter...
Closing *host.Host...
Closing *wallet.Wallet...
Closing *transactionpool.TransactionPool...
Closing *consensus.ConsensusSet...
Closing *gateway.Gateway...
Shutdown complete.
```

lukechampine: It's a cute trick but not good enough. Let's use a separate slice for the names, or define a type containing both the name and io.Closer.

avahowell: ACK, I added a new `moduleCloser` type which prints the name in its `Close` method.

cmd/siad/server.go
```diff
@@ -145,6 +151,12 @@ func (rs byVersion) Less(i, j int) bool {
 	return build.VersionCmp(rs[i].version(), rs[j].version()) >= 0
 }
 
+// Close implements the Close method for moduleCloser.
+func (mc *moduleCloser) Close() error {
+	fmt.Printf("Closing %v...\n", mc.name)
+	return mc.closer.Close()
+}
```
lukechampine: hmm. Having in the `Printf` inside the `Close` feels wrong. I'd keep it outside.

lukechampine: specifically, I'd embed the `io.Closer` in `moduleCloser`, so you don't need to define a new method at all

avahowell: Ah, gotcha. I'll push another commit

Issue #2274, opened 2017-08-30
Sia Renter Zeros out Everything


eddiewang:
Renter has been stable for the past few days... I think I triggered this bug by downloading a file.

Basically, after downloading the file, the renter details all show 0.
Here is a screenshot of the bug:
![image](https://user-images.githubusercontent.com/2057997/29856352-dce0f668-8d17-11e7-8145-4acdaa1b3aba.png)

Sia-UI also shows 0 SC spent, while the real amount is 160~ SC:
![image](https://user-images.githubusercontent.com/2057997/29856383-0aed944e-8d18-11e7-8406-262ff9817854.png)

Fornax96:
This happened to me too last night. But I didn't download any files. The renter log is completely empty, nothing logged for the last 10 days (it's been running for that long).

All the counters just restarted from 0, and are now going up again. Yesterday my storage spending was 930 SC, now it's 1.015 SC, same for all the other counters. Unspent funds says `30.63 SC` for some strange reason.

```
Renter info:
	Storage Spending:  1.015 SC
	Upload Spending:   3.623 mS
	Download Spending: 0 H
	Unspent Funds:     30.63 SC
	Total Allocated:   31.65 SC
```

```
Allowance:
	Amount: 2 KS
	Period: 12096 blocks
```

Fornax96:
Could this be a time-based thing? I noticed that on my desktop PC the stats have also returned to zero, and someone else on discord also reported their stats resetting!

That seems like too much of a coincidence.

tbenz9:
Recommendation to devs: "Bug"

MeijeSibbel:
I think we might have the same issue:  https://github.com/GooBox/goobox-community-gui/issues/80

PR #2450, opened 2017-10-26
fix host settings reset bug

avahowell:
This PR fixes a bug where if a user supplies an invalid host parameter to the `/host` API endpoint, all of the host's internal settings will be reset to zero-values.


tbenz9:
Is this related to #2274?  If yes, you may want to add the "fixes" keyword.

avahowell:
No, this is for the host settings, not the renter financial metrics.

PR #2451, opened 2017-10-27
Integrate contractset

DavidVorick:


modules/renter/contractor/update.go
```diff
@@ -5,6 +5,52 @@ import (
...
+	defer c.tg.Done()
+
+	// Determine the current block height.
+	c.mu.RLock()
+	currentHeight := c.blockHeight
+	c.mu.RUnlock()
+
+	// Loop through the current set of contracts and migrate any expired ones to
+	// the set of old contracts.
+	for _, contract := range c.contracts.ViewAll() {
```
rippaio: In other places, you lock c.contracts.  Should probably lock here as well.

modules/renter/contractor/update.go
```diff
@@ -20,30 +66,12 @@ func (c *Contractor) ProcessConsensusChange(cc modules.ConsensusChange) {
...
-		c.log.Println("INFO: archived expired contract", id)
-	}
-
 	// If we have entered the next period, update currentPeriod
 	// NOTE: "period" refers to the duration of contracts, whereas "cycle"
 	// refers to how frequently the period metrics are reset.
 	// TODO: How to make this more explicit.
 	cycleLen := c.allowance.Period - c.allowance.RenewWindow
-	if c.blockHeight > c.currentPeriod+cycleLen {
+	if c.blockHeight >= c.currentPeriod+cycleLen {
```
rippaio: This whole block lends itself to a method.  c.invalidateMetricsContract() or similar

modules/renter/contractor/update.go
```diff
@@ -58,11 +86,11 @@ func (c *Contractor) ProcessConsensusChange(cc modules.ConsensusChange) {
...
 
-	// Only attempt contract formation/renewal if we are synced
-	// (harmless if not synced, since hosts will reject our renewal attempts,
-	// but very slow).
+	// Perform contract maintenance if our blockchain is synced. Use a separate
+	// goroutine so that the rest of the contractor is not blocked during
+	// maintenance.
 	if cc.Synced {
-		// Perform the contract maintenance in a separate thread.
 		go c.threadedContractMaintenance()
```
rippaio: Does this really need to happen every ~10 minutes?

modules/renter/proto/contractset.go
```diff
@@ -22,11 +22,34 @@ type ContractSet struct {
...
+func (cs *ContractSet) Acquire(id types.FileContractID) (modules.RenterContract, bool) {
 	cs.mu.Lock()
-	defer cs.mu.Unlock()
-	return len(cs.contracts)
+	safeContract, ok := cs.contracts[id]
+	cs.mu.Unlock()
+	if !ok {
+		return modules.RenterContract{}, false
+	}
+	safeContract.mu.Lock()
```
rippaio: This seems like a deadlock waiting to happen.  Is there some way to ensure that this gets unlocked?

modules/renter/proto/contractset.go
```diff
@@ -22,11 +22,34 @@ type ContractSet struct {
...
+}
+
+// Delete removes a contract from the set. The contract must have been
+// previously acquired by Acquire. If the contract is not present in the set,
+// Delete is a no-op.
+func (cs *ContractSet) Delete(contract modules.RenterContract) {
+	cs.mu.Lock()
+	safeContract, ok := cs.contracts[contract.ID]
+	if !ok {
+		cs.mu.Unlock()
```
rippaio: your cs.mu.Unlock()s can occur in a defer here.

modules/renter/proto/contractset.go
```diff
@@ -68,56 +75,70 @@ func (cs *ContractSet) Insert(contract modules.RenterContract) {
...
 // Return returns a locked contract to the set and unlocks it. The contract
 // must have been previously acquired by Acquire. If the contract is not
 // present in the set, Return panics.
 func (cs *ContractSet) Return(contract modules.RenterContract) {
 	cs.mu.Lock()
-	defer cs.mu.Unlock()
-	sc, ok := cs.contracts[contract.ID]
+	safeContract, ok := cs.contracts[contract.ID]
 	if !ok {
+		cs.mu.Unlock()
```
rippaio: defer Unlock() and a missing return here.

modules/renter/proto/downloader.go
```diff
@@ -77,15 +83,15 @@ func (hd *Downloader) Sector(root crypto.Hash) (_ modules.RenterContract, _ []by
...
 		} else if err == nil {
-			hd.hdb.IncrementSuccessfulInteractions(hd.contract.HostPublicKey)
+			hd.hdb.IncrementSuccessfulInteractions(contract.HostPublicKey)
 		}
 	}()
 
 	// send the revision to the host for approval
-	extendDeadline(hd.conn, 2*time.Minute)
-	signedTxn, err := negotiateRevision(hd.conn, rev, hd.contract.SecretKey)
+	extendDeadline(hd.conn, 2*time.Minute) // TODO: Constant.
```
rippaio: Fix TODO

modules/renter/proto/downloader.go
```diff
@@ -153,16 +162,16 @@ func NewDownloader(host modules.HostDBEntry, contract modules.RenterContract, hd
...
 			hdb.IncrementSuccessfulInteractions(contract.HostPublicKey)
 		}
 	}()
 
 	// initiate download loop
 	conn, err := (&net.Dialer{
 		Cancel:  cancel,
-		Timeout: 15 * time.Second,
-	}).Dial("tcp", string(contract.NetAddress))
+		Timeout: 45 * time.Second, // TODO: Constant
```
rippaio: Fix TODO

modules/renter/proto/contractset.go
```diff
@@ -68,56 +75,70 @@ func (cs *ContractSet) Insert(contract modules.RenterContract) {
...
-	delete(cs.contracts, contract.ID)
-	sc.mu.Unlock()
+	c := safeContract.RenterContract
+	c.MerkleRoots = nil
+	return c, true
+}
+
+// ViewAll returns a copy of each contract in the set. The contracts are not
+// locked. Certain fields, including the MerkleRoots, are set to nil for
+// safety reasons.
```
starius: Can you elaborate on what are the reasons to reset MerkleRoots, please?

modules/renter/proto/contractset.go
lukechampine: `cs.mu` needs to be unlocked before `safeContract.mu`

modules/renter/proto/contractset.go
lukechampine: You should only be able to modify the merkle roots if you hold the contract's lock. `ViewAll` provides the `modules.RenterContract` but doesn't acquire the contract's lock. It's okay to return the `modules.RenterContract` because it's copied by value, so callers can't modify shared data. But `MerkleRoots` is a slice, and copying the slice pointer still gives you access to the underlying (shared) data. So we either have to copy the whole slice (not practical here, since the Merkle roots may be large), or set it to nil.

api/host_test.go
```diff
@@ -972,6 +972,7 @@ func TestRemoveStorageFolderForced(t *testing.T) {
 
 // TestDeleteSector tests the call to delete a storage sector from the host.
 func TestDeleteSector(t *testing.T) {
+	t.Skip("broken because Merkle roots are no longer exposed")
```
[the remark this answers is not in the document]
DavidVorick: Uncertain, we may have to move the test to inside the proto package.

modules/renter/contractor/update.go
DavidVorick: Potentially not, given that hosts tend to be in flux over a period of days. On the other hand, the process does not take very many computational resources to complete, I imagine just a few milliseconds. We don't do it while syncing, and we do it in a goroutine which means it won't affect other modules in the chain.

Overall, it didn't seem harmful to do it frequently, and doing it based on the consensus pulse means we don't need a background thread.

modules/renter/proto/contractset.go
DavidVorick: The reason we can't currently merge this PR is that the concurrency paradigms around the contract set are not sufficient. We can either choose between race conditions or deadlocks, or really slow performance.

The future work on this branch is moving the contracts fully into the `proto` package, which will allow us to avoid the deadlocks and race conditions without hurting performance.

PR #2456, opened 2017-11-02
Refactor Wallet.applyHistory

mtlynch:
At 189 LOC, the applyHistory() function is perilously long. This refactors some of the logic out of the function into helper functions to bring its LOC down to 176 (7% reduction).

### DIFF
--- a/modules/wallet/update.go
+++ b/modules/wallet/update.go
@@ -197,24 +197,37 @@ func (w *Wallet) revertHistory(tx *bolt.Tx, reverted []types.Block) error {
 	return nil
 }
 
+// getSiacoinSpentOutputs scans a slice of Siacoin output diffs for spent
+// outputs and collects them in a map of SiacoinOutputID -> SiacoinOutput.
+func getSiacoinSpentOutputs(diffs []modules.SiacoinOutputDiff) map[types.SiacoinOutputID]types.SiacoinOutput {
+	outputs := make(map[types.SiacoinOutputID]types.SiacoinOutput)
+	for _, diff := range diffs {
+		if diff.Direction == modules.DiffRevert {
+			// DiffRevert means spent.
+			outputs[diff.ID] = diff.SiacoinOutput
+		}
+	}
+	return outputs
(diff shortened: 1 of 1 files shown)


lukechampine:
thanks for this! I think the function names could be a little more descriptive (since they're living in the top-level namespace). Another good candidate for refactoring is the `ProcessedTransaction` part of `applyHistory`. Adding an `isTransactionRelevant` helper and a `buildProcessedTransaction` helper (names not final) would clean things up a lot.

mtlynch:
>I think the function names could be a little more descriptive (since they're living in the top-level namespace)

Sure, what name would be more helpful? My only thought is something like `getSiacoinSpentOutputsFromDiffs` but I feel like I worry I'm adding verbosity without adding value.

>Another good candidate for refactoring is the `ProcessedTransaction` part of `applyHistory`. Adding an `isTransactionRelevant` helper and a `buildProcessedTransaction` helper (names not final) would clean things up a lot.

Yeah, definitely. Do you mean within this PR? I was planning to do them in small, independent PRs to minimize the diff sizes and make it easier to spot possible mistakes in the refactoring.

DavidVorick:
This PR looks good to me, including the names. Let's move future refactoring to another PR so this one is not blocked.

Will let @lukechampine hit the merge button since he had the reservations about the names.

lukechampine:
how about `computeSpentOutputSets`, which would return both maps? I prefer `compute` to `get` because you aren't really retrieving anything, just restructuring it, and I like `Sets` because it tells you what sort of object you're getting back (i.e. "hmm, 'set' means it's probably a map").

If combining them feels messy, `computeSpentSia{coin,fund}OutputSet` also works.

mtlynch:
`compute` works for me.

The combined function feels a little unnatural, so I've left them as separate functions. That way it'll be easy for us to combine them [when Go adds generics](http://airlinepilotguy.com/wp-content/uploads/2014/12/pig-flying.png). I've updated the PR.

PR #2458, opened 2017-11-03
Explorer refactor - Enable a new generation of Sia applications

nitronick600:
1.) Block query ranges.  You can provide a to and from query parameter with a maximum of 10 blocks to retrieve
2.) Websockets for consensus and pending transaction updates.  This will be the most interesting change, since it will enable real-time sites.  Websockets enable server push mechanisms, so data will be available to clients in real time.
3.) HashType bucket.  To keep from having to lookup into every hash bucket, we keep track of the hash type as it's inserted.  This acts as an index and gives us a constant-time lookup of hashes without having to scan every hash bucket
4.) Consistency fixes.  I've removed some logic which causes the explorer to fail.  This makes the explorer more resilient and reduces the instance of "block not found" errors in the explorer.
5.) In memory pending block.  We keep a copy of the current, pending block in memory.  This allows us to add/remove transactions and update block facts without having to commit to any DB when pending txs are received


lukechampine:
I like a lot of these changes but we need to split this into multiple PRs -- approximately one PR for each of your bullet points. I'm thinking:

1. Consistency fixes. This is the most important to merge in the short term.
2. HashType bucket. Also, move all of the explorer code currently living in the `api` package into `explorer`.
3. Add a tpool to the explorer -- maybe. I don't totally understand why this is necessary; couldn't we just add more `/transactionpool` routes?
4. Block query ranges.
5. Websocket support.

PR #2459, opened 2017-11-03
Split applyHistory into applyHistoryFromBlock

mtlynch:
Following #2456, this further refactors the Wallet.applyHistory function by splitting the body of the giant loop it contains into a separate method.

This reduces the 176 LOC method to 170 LOC and reduces the control flow nesting depth by 1.

---
This change is [Reviewable](https://reviewable.io/reviews/nebulouslabs/sia/2459)

### DIFF
--- a/modules/wallet/update.go
+++ b/modules/wallet/update.go
@@ -10,6 +10,11 @@ import (
 	"github.com/NebulousLabs/bolt"
 )
 
+type (
+	spentSiacoinOutputSet map[types.SiacoinOutputID]types.SiacoinOutput
+	spentSiafundOutputSet map[types.SiafundOutputID]types.SiafundOutput
+)
+
 // threadedResetSubscriptions unsubscribes the wallet from the consensus set and transaction pool
 // and subscribes again.
 func (w *Wallet) threadedResetSubscriptions() error {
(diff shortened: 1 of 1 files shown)


mtlynch:
I'll rebase from `master` once the previous PR is merged in.

lukechampine:
Since I have a functional programming streak, what do you think about having the helper function just compute and return the new ProcessedTransactions, which would then be inserted into the DB?

mtlynch:
>Since I have a functional programming streak, what do you think about having the helper function just compute and return the new ProcessedTransactions, which would then be inserted into the DB?

That sounds reasonable. Are we talking about something like this?

```golang
func (w *Wallet) computeProcessedTransactionsFromBlock(tx *bolt.Tx, block types.Block, spentSiacoinOutputs spentSiacoinOutputSet, spentSiafundOutputs spentSiafundOutputSet) ([]modules.ProcessedTransaction, error)

...

   pts, err := computeProcessedTransactionsFromBlock(tx, block, spentSiacoinOutputs, spentSiafundOutputs)
   if err != nil {
      return err
   }
   for _, pt := range pts {
     err := dbAppendProcessedTransaction(tx, pt)
     if err != nil {
        return fmt.Errorf("could not put processed transaction: %v", err)
     }
  }
```

Note that example represents a behavior change in that the current implementation makes some `dbAppendProcessedTransaction` even if there are errors calculating later `ProcessedTransaction`s. If that detail is important, I can write it to preserve that behavior.

lukechampine:
yes, that's pretty much what I had in mind. In practice I would expect that `computeProcessedTransactions` would rarely return an error, since it should just be reading from the db. Perhaps it could just log any errors encountered and return the set of txns that it successfully computed (and no error).

mtlynch:
Updated based on your suggestion.

Issue #2447, opened 2017-10-24
Critical bug: all locked (not just risked) collateral lost at contracts expire


MaxMaximus:
Hello.

I have finished my full cycle of testing as a host on sia network (from July to mid of October) and results just awful. It not only have lot of minor to moderate bug and problems (i already reported some of them here) but test ended with such unacceptable critical situation:
- all contract already expired, but host did not get ANY of storage revenues (despite fact it successfully post some storage proofs to blockchain) and also had host all collateral locked in contracts. 
Not just risked collateral (it can be happening if host fail storage proof part), but all collateral sent to contract.

At first i thought it is a just wallet display bug as reported there: https://github.com/NebulousLabs/Sia/issues/2420
So host actually getting collateral back and this is just not reflected properly as incoming transaction in wallet.
But later i set up wallet balance monitoring - and found out that available  balance only goes down. There are no additions to balance after contract expire - only additional small deduction for fees each time a next contract expire.

I do not find anything interesting in logs what can lead to situation with full collateral lost. 
There are some messages about successful storage proofs, some errors with missed storage proofs as described in that issue: https://github.com/NebulousLabs/Sia/issues/1862 (data folder is intact, host online, wallet unlocked, but still somehow miss some of storage proofs). No other major errors.

There details:
windows host (window 7 64 bit Pro) running siad. ver. 1.22 initially, later host was updated to ver. 1.3.0 via built-in updater (siac update ) .
100 % hosting, never used for renting/uploading files

Here full wallet detail from test host:
[wallet-tx.txt](https://github.com/NebulousLabs/Sia/files/1412620/wallet-tx.txt)
As you can see there are only 2 incoming tx (50+350 SC). It was my own tx - loading this test wallet by external transaction (from my other wallet).
All other txs were made by host automatically. And there is no any incoming tx despite all contract already ended (last one about a week ago).
Final balance: 48 SC from initial 400 SC. So actual storage revenue is a negative value.

Here stats from host (siac host -v outputs i have taken manually from time to time)
[host-v.txt](https://github.com/NebulousLabs/Sia/files/1412656/host-v.txt)

Stat shows that all contracted already ended. Note additional minor bug: host stats reports some of Locked/risked Collateral and potential revenues still remaining despite fact that all contract already expired and amount of data stored is zero at last report.

[...]

PR #2469, opened 2017-11-13
check for address existence in host.checkUnlockHash

avahowell:
This should help with #2447. After this change, `checkUnlockHash` is only called on announce, should it be called once per block as well?

### DIFF
--- a/modules/host/announce_test.go
+++ b/modules/host/announce_test.go
@@ -136,3 +136,60 @@ func TestHostAnnounceAddress(t *testing.T) {
 		t.Error("announcement has wrong host key")
 	}
 }
+
+// TestHostAnnounceCheckUnlockHash verifies that the host's unlock hash is
+// checked when an announcement is performed.
+func TestHostAnnounceCheckUnlockHash(t *testing.T) {
+	if testing.Short() {
+		t.SkipNow()
+	}
+	t.Parallel()
+	ht, err := newHostTester(t.Name())
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer ht.Close()
+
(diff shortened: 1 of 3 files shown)


MaxMaximus:
Hello, You mention my host bug.

May be it is overkill and excessive load to check for it after each new block. May be once at each host start(program launch) would be enough? Or before each contract formation procedure call.
And at least client build-in functions like "wallet init" and "wallet init-seed" which replace primary wallet seed must also update host unlock hash after replacing seed. If user call one of those functions - client always know and 100% sure that wallet seed was already changed and must not use old unlock hash.

I used "siac wallet init --force --password" command to replace seed(and wallet encryption password - in ver. 1.2.x there was no option to change password on existing wallet, only recreate new one with custom pass - this was the reason why i decided to change seed and recreate wallet)
And I could not even think that the client would continue to use the keys or unlock hash from the old (destroyed because of --force option) wallet to form new contract after that. This is clearly an incorrect program behavior.

Also i suggest check for current implementation of such check - i have announced my host few times after wallet seed  was replaced but host has been still using unlock hash from old (deleted) seed even after announce. 
Or this check just did not exist in ver. 1.3.0 and was added later (e.g. in 1.3.1 RC)?

lukechampine:
Another complication of changing the wallet seed is that it will break any active file contracts.

I suggest that:

1) When changing the seed, we should provide the option to continue tracking the seed as an auxiliary seed. At best, this prevents loss of coins; at worst, it confuses the user into thinking their seed hasn't changed or that coins are only going to their new seed.
2) We should try to better inform users of the consequences of changing the seed.

PR #2475, opened 2017-11-18
Add a lock to guard concurrent map access in Renter.FileList

### LINKED ISSUE #2473: fatal error: concurrent map read and map write
Bug Report
==========

I'm testing out the Minio integration by uploading a bunch of files through Minio. It looks like it worked for ~7 hours and then crashed with `fatal error: concurrent map read and map write`.

I was uploading about 300 files at the time.

https://gist.github.com/mtlynch/2d7e2d693a6dfdef4ea8b934f80a6aed

* Sia version: `master` branch @ aacc30609e3bfc18baa2b1cf3cd89bd457a3715c
* OS: Ubuntu 16.04.3 LTS

mtlynch:
FileList reads the map, r.tracking, but does not acquire a read lock beforehand,
which leads to crashes. This adds a helper function that acquires the read lock
during the map read.

Fixes #2473

### DIFF
--- a/modules/renter/files.go
+++ b/modules/renter/files.go
@@ -207,8 +207,8 @@ func (r *Renter) DeleteFile(nickname string) error {
 	return nil
 }
 
-// getTrackedFile returns the metadata for a tracked file, if the file exists.
-func (r *Renter) getTrackedFile(nickname string) (tf trackedFile, exists bool) {
+// managedGetTrackedFile returns the metadata for a tracked file, if the file exists.
+func (r *Renter) managedGetTrackedFile(nickname string) (tf trackedFile, exists bool) {
 	lockId := r.mu.RLock()
 	defer r.mu.RUnlock(lockId)
 	tf, exists = r.tracking[nickname]
@@ -239,7 +239,7 @@ func (r *Renter) FileList() []modules.FileInfo {
 		f.mu.RLock()
 		renewing := true
 		var localPath string
-		tf, exists := r.getTrackedFile(f.name)
+		tf, exists := r.managedGetTrackedFile(f.name)
 		if exists {
 			localPath = tf.RepairPath
 		}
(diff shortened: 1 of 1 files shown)


lukechampine:
Our locking conventions mandate that unexported methods cannot acquire locks unless their name begins with `managed`. So one option would be to rename `getTrackedFile` to `managedGetTrackedFile`. In this case though, I think it would be simplest to just move the lock/unlock into the body of `FileList`.

mtlynch:
>Our locking conventions mandate that unexported methods cannot acquire locks unless their name begins with `managed`.

Do we have this documented? If not, I can file a bug to add it to [Developers.md](https://github.com/NebulousLabs/Sia/blob/master/doc/Developers.md).

>So one option would be to rename getTrackedFile to managedGetTrackedFile. In this case though, I think it would be simplest to just move the lock/unlock into the body of FileList.

That's what I tried originally, but then we end up with two different lock IDs in the same function, which makes me kind of antsy, even though they're pretty far from each other. I renamed to `managedGetTrackedFile`.

lukechampine:
Hmm, I thought it was documented, but I guess not. The `threaded` convention is documented; `managed` is basically the same except not called in a goroutine.
@DavidVorick and I discussed awhile back whether `threaded` and `managed` should be merged, but I forget if we concluded anything definitively. For now, let's document that `managed` is for helper methods that need to manage their own locking, usually because they make some blocking call (disk I/O, network I/O, waiting on a channel, etc.)

DavidVorick:
With managed, the implication is that you are going to call the function synchronously (you don't have to, but the function has to be designed so that synchronous calling is acceptable). With threaded, the requirement is that the function be called asynchronously. Typically, threaded functions contain an infinite loop and run some background process, which means they return only after some signal and definitely can't be called synchronously. There are a few other use cases for threaded functions as well.

We should definitely keep them separate.

PR #2476, opened 2017-11-18
Removing testing.T.Helper() call

mtlynch:
As far as I can see, this isn't a defined function and results in build errors:

```
modules/renter/proto/contractset_test.go:17: t.Helper undefined (type *testing.T has no field or method Helper)
```


lukechampine:
`t.Helper` was added in Go 1.9, which was released in August. Perhaps we should have a more general discussion regarding when it's appropriate for Sia to upgrade to a given version of Go. I prefer to switch to the newest version as soon as it's feasible, but I understand that there are competing concerns.

mtlynch:
Whoops, my mistake. I should have checked the latest Go docs.

PR #2477, opened 2017-11-18
Add backticks to Developers.md

mtlynch:
The developer documentation is inconsistent about backticking text that refers to code or shell commands. This adds backticks so non-prose text is formatted distinctly when rendered from Markdown.


lukechampine:
thanks!

Issue #2467, opened 2017-11-12
/renter/files API should expose upload progress in bytes


huetsch:
I recently wrote a download speed monitor for the Sia UI. I wanted to extend the code to display upload speed as well, but the "uploadprogress" field returned by the API tripped me up. It wasn't immediately obvious to me if a value of 100% included the parity bytes or not. Looking at the code, 100% does include parity. I realize I could figure out the file size inclusive of parity before I make my API call and then multiply that by the upload progress percentage to figure out how many bytes have been uploaded, but this makes using the API really painful.

How about a bytesUploaded() function in modules/renter/files.go? Then we could return a "bytesuploaded" field alongside the "uploadprogress" field in the /renter/files API.

I'll go ahead and write the code for a PR if you guys approve.

DavidVorick:
Basically you want to add a field to the API so that you do not have to do a calculation based on filesize + progress?

@lukechampine what do you think?

lukechampine:
I'm ok with `bytesUploaded` as long as we're clear about what it means. Namely, `bytesUploaded` won't progress to exactly `filesize` and then stop; it can grow pretty much unbounded, depending on the redundancy of the file and how much repair work is needed. As a consequence, you can't use `bytesUploaded` to infer when uploading will stop. It should only be used to estimate upload speed and to see how many total bytes have been uploaded over the lifetime of the file.

PR #2478, opened 2017-11-18
add uploadedBytes() to track how many bytes have been uploaded

huetsch:
This resolves https://github.com/NebulousLabs/Sia/issues/2467

Currently an API caller can't accurately measure upload throughput because the upload API fields don't have information about how a file was chunked or padded. This PR directly exposes how many bytes have been uploaded via a file's contracts.

### DIFF
--- a/api/renterhost_test.go
+++ b/api/renterhost_test.go
@@ -18,6 +18,7 @@ import (
 	"time"
 
 	"github.com/NebulousLabs/Sia/build"
+	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/types"
 )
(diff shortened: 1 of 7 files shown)


doc/api/Renter.md
```diff
@@ -212,6 +212,12 @@ lists the status of all files.
       // with 0 redundancy.
       "redundancy": 5,
 
+      // Total number of bytes successfully uploaded via current file contracts.
+      // This number includes padding and rendundancy, so a file with a size of
+      // 8192 bytes might be padded to 40 MiB and, with a redundancy of 5,
+      // striped to 200 MiB for upload.
```
DavidVorick: important nitpick - 'striped' is the wrong word here, need to use the word 'encoded'. The redundancy scheme that we use is far superior to striping, and I don't want anyone reading the docs to think we did our redundancy in a naive way.

DavidVorick:
This PR needs a test in the api package that proves the API is serving the field correctly. api/renterhost_test.go has some tests in it that are full-stack. You can probably insert code into one of those that checks that this new api field is giving sensible results.

If you do decide to inject the code into an existing test, please include a comment in the test docstring specifying that this function is responsible for testing certain api fields, so that we know to include that part of the code if we do some test consolidation later.

huetsch:
I implemented your suggestions. I went ahead and wrote an independent `TestUploadedBytesReporting` function.

api/renterhost_test.go
```diff
@@ -2171,6 +2172,133 @@ func TestRedundancyReporting(t *testing.T) {
...
+	// Monitor the file as it uploads. Ensure that the UploadProgress times
+	// the fully redundant file size always equals UploadedBytes reported
+	var rf RenterFiles
+	for i := 0; i < 60 && (len(rf.Files) != 1 || rf.Files[0].UploadProgress < 100); i++ {
+		st.getAPI("/renter/files", &rf)
+		if len(rf.Files) >= 1 {
+			uploadProgressBytes := uint64(float64(fullyRedundantSize) * rf.Files[0].UploadProgress / 100.0)
+			// Note: in Go 1.10 we will be able to write Math.Round(uploadProgressBytes) != rf.Files[0].UploadedBytes
+			if uploadProgressBytes != rf.Files[0].UploadedBytes && (uploadProgressBytes+1) != rf.Files[0].UploadedBytes {
+				t.Fatal(fmt.Sprintf("api reports having uploaded %v bytes when upload progress is %v%%, but the actual uploaded bytes count should be %v", rf.Files[0].UploadedBytes, rf.Files[0].UploadProgress, uploadProgressBytes))
```
ChrisSchinnerl: You can use `t.Fatalf("formatstring", args...)` to format the string instead of `t.Fatal(fmt.Sprintf("formatstring", args...)`. I would also add a newline after the format string and split that call over 2 lines.

api/renterhost_test.go
```diff
@@ -2171,6 +2172,133 @@ func TestRedundancyReporting(t *testing.T) {
...
+
+	// Upload progress should be 100% and redundancy should reach 2
+	if len(rf.Files) != 1 || rf.Files[0].UploadProgress < 100 || rf.Files[0].Redundancy != 2 {
+		t.Fatal("the uploading is not succeeding for some reason:", rf.Files[0])
+	}
+
+	// When the file is fully redundantly uploaded, UploadedBytes should
+	// equal the file's fully redundant size
+	if rf.Files[0].UploadedBytes != fullyRedundantSize {
+		t.Fatal(fmt.Sprintf("api reports having uploaded %v bytes when upload progress is 100%%, but the actual fully redundant file size is %v", rf.Files[0].UploadedBytes, fullyRedundantSize))
```
ChrisSchinnerl: same as above

huetsch:
Done.

DavidVorick:
Excellent, thank you very much for this PR.

PR #2496, opened 2017-11-27
fix random race condition in workerupload.go

ChrisSchinnerl:
I don't know if this causes all the race condition problems in the upload code but it seems like a good start.


DavidVorick:
Unfortunately the race problems in the upload code run a lot deeper, and can't be detected by a race detector because the threads themselves don't race on particular cells of memory, but rather they have whole copies of objects when it's not safe to have copies of the objects.

The contractset branch that's almost complete should fix most of the fundamental race conditions, Sia should be working a lot more smoothly after this.

PR #2504, opened 2017-12-03
Integrate contractset

DavidVorick:
This pull request takes the contracts of the contractor and moves them to a stateful object in the proto package. The job of the proto package (which may be renamed) is to manage individual contracts, and perform simple upload/download operations on those contracts.

The biggest advantage here is clean separation of the contracts from all the rest of the state of the contractor. The contracts also now have strict safety properties, which will eliminate the race conditions that we were running into. Finally, the contracts have been moved to a highly scalable 'one file per contract' model, which means we can perform all contract operations with ACID guarantees in a constant number of disk operations, without needing to amortize any expensive operations at all.

Once this is merged and tested, we should be ready to release RC2.

### DIFF
--- a/api/host_test.go
+++ b/api/host_test.go
@@ -1046,6 +1046,9 @@ func TestDeleteSector(t *testing.T) {
 	if len(contracts) != 1 {
 		t.Fatalf("expected exactly 1 contract to have been formed; got %v instead", len(contracts))
 	}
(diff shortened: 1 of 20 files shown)


lukechampine:
Did you mean to PR `contract-wal`? These aren't my latest commits. `contract-wal` extends `integrate-contractset`.

DavidVorick:
ah, did not realize.

PR #2506, opened 2017-12-04
Contract wal

DavidVorick:
This pull request takes the contracts of the contractor and moves them to a stateful object in the proto package. The job of the proto package (which may be renamed) is to manage individual contracts, and perform simple upload/download operations on those contracts.

The biggest advantage here is clean separation of the contracts from all the rest of the state of the contractor. The contracts also now have strict safety properties, which will eliminate the race conditions that we were running into. Finally, the contracts have been moved to a highly scalable 'one file per contract' model, which means we can perform all contract operations with ACID guarantees in a constant number of disk operations, without needing to amortize any expensive operations at all.

Once this is merged and tested, we should be ready to release RC2.


modules/renter/contractor/dependencies.go
```diff
@@ -72,46 +71,27 @@ type walletBridge struct {
...
-func (p *stdPersist) save(data contractorPersist) error {
-	if p.journal == nil {
-		var err error
-		p.journal, err = newJournal(p.filename, data)
-		return err
-	}
-	return p.journal.checkpoint(data)
+var persistMeta = persist.Metadata{
+	Header:  "Contractor Persistence",
+	Version: "1.3.2",
```
DavidVorick: v1.3.1

modules/renter/contractor/downloader.go
```diff
@@ -74,78 +92,53 @@ func (hd *hostDownloader) Sector(root crypto.Hash) ([]byte, error) {
...
 	}
-	// Update the contract to the most recent net address for the host.
-	contract.NetAddress = host.NetAddress
 
-	// acquire revising lock
+	// Acquire the revising lock for the contract, which excludes other threads
+	// from interacting with the contract.
+	//
+	// TODO: Because we have another layer of contract safety via the
+	// contractset, do we need the revising lock anymore?
```
DavidVorick: Need to double check everything that the revising lock is protecting, but I believe that the contractset now provides all the safety that you need, and that we can drop the revising lock.

modules/renter/contractor/editor.go
```diff
@@ -110,109 +104,55 @@ func (he *hostEditor) Upload(data []byte) (_ crypto.Hash, err error) {
...
 	id = c.ResolveID(id)
 	c.mu.RLock()
 	cachedEditor, haveEditor := c.editors[id]
 	height := c.blockHeight
-	contract, haveContract := c.contracts[id]
 	renewing := c.renewing[id]
 	c.mu.RUnlock()
 
 	if renewing {
+		// Cannot use the editor if the contract is being renewed.
```
DavidVorick: This is another check that might no longer be necessary if the contractset is now providing appropriate safety.

modules/renter/proto/contractset.go
```diff
@@ -1,32 +1,57 @@
...
+	safeContract, ok := cs.contracts[c.header.ID()]
+	if !ok {
+		cs.mu.Unlock()
+		return
+	}
+	delete(cs.contracts, c.header.ID())
+	cs.mu.Unlock()
+	safeContract.mu.Unlock()
+	// delete contract file
+	os.Remove(filepath.Join(cs.dir, c.header.ID().String()+".contract"))
```
ChrisSchinnerl: We use `.contract` a lot. I feel like we should put that in a const.

Issue #1715, opened 2017-04-16
siac host output should warn if wallet is locked


mtlynch:
Can the output of `siac host -v` warn the user if they're accepting contracts but their wallet is locked? Currently, it prints out this:

```
$ siac host -v
General Info:
        Estimated Competitive Price: 8.728 KS

Host Internal Settings:
        acceptingcontracts:   Yes
```

This is misleading because if the host's wallet is locked, they're not really accepting contracts.

DavidVorick:
Yeah. In general I think Sia doesn't have enough warnings if the user is shooting themselves in the foot.

If your wallet is locked the renter can't replace dud hosts.

If your wallet is locked the host can't form new contracts.

You host may be set to 'acceptingcontracts: No' but you don't want it that way (maybe some misconfiguration happened or something),

etc.

We should be able to flag the user's attention if they are doing something that seems undesirable.

PR #2509, opened 2017-12-06
 Implemented #1715 / Show wallet locked warning in host command

chrsch:


cmd/siac/hostcmd.go
```diff
@@ -283,6 +283,17 @@ RPC Stats:
 			currencyUnits(totalRevenue))
 	}
 
+	// if wallet is locked print warning
+	walletstatus := new(api.WalletGET)
+	walleterr := getAPI("/wallet", walletstatus)
+	if err != nil {
+		die("Could not get wallet status. A working wallet is needed in order to operate your host. Error:", walleterr)
```
lukechampine: I'm not sure if `die` is appropriate here; a printed warning (like below) is probably sufficient

cmd/siac/hostcmd.go
```diff
@@ -283,6 +283,17 @@ RPC Stats:
...
+	} else {
+		if !walletstatus.Unlocked {
+			fmt.Println("\nWARNING:\n	In order to operate your host you need to unlock your wallet using the command \"siac wallet unlock\"")
+		}
+	}
```
lukechampine: you can save some lines here by putting `else if` on the same line. Also, I don't think it's necessary to specify the exact command. How about `"\nWarning: Your wallet is locked. You must unlock your wallet for the host to function properly."`

lukechampine:
Thanks! We can merge this after a bit of tweaking.

chrsch:
Thanks for review & suggestions!

Issue #1478, opened 2016-10-26
Pagination in the wallet API


DavidVorick:
All wallet calls which return potentially large amounts of data (like the transactions or the addresses) need to support pagination in the API.

tbenz9:
I recommend adding the "Programming Task" label for this to get more visibility.  If you're planning on implementing it yourself assign it to Yourself.

hudaniel:
My attempt: https://github.com/NebulousLabs/Sia/pull/2511

PR #2511, opened 2017-12-07
Pagination for /wallet/addresses and /wallet/transactions

hudaniel:
New to golang. Please let me know what could be improved.

Changes:
Adding an optional `page` query to `/wallet/addresses` and optional queries `confirmPage` and `unconfirmedPage` to `/wallet/transactions`. Not specifying these queries will return the same response as before to better support the command line tools.

The response of these endpoints given the page queries will reply with a `total_pages` field which will tell the client how many pages are left, beginning with 0.


tbenz9:
I think it makes more sense to have `total_pages` start at 1 instead of 0, even if a response only has 1 entry, that should still count as 1 page.

Is it possible to specify how many results are included in each response instead of hard-coding it at 25?

mindjuice:
Most pagination APIs don't require you to send page numbers, but rather item numbers.

You provide `start` as the starting item number, and `limit` as the maximum number of items to return.  Then for the next request, you increment `start` by the number of items received.

This is more flexible and lets you have multiple ways of "paging":

- Move forward one item: `[ next ]`
- Move forward 5 items: `[ > ]`
- Move forward 10 items: `[ >> ]`

lukechampine:
It's also standard to include links to the next/previous pages in the header of the response. Here's an example from the GitHub API: https://developer.github.com/v3/guides/traversing-with-pagination. We don't have to add those now, but it's a good thing to keep in mind.

tbenz9:
@DavidVorick as the reporter of #1478 do you have a preference on having the pagination delineate by total pages or item numbers?  It seems there are pros/cons to both ideas and either would probably work fine.

hudaniel:
Cool, I've seen pagination done both ways. But yeah the one that specifies the start might be better.

hudaniel:
sample response for 
```
localhost:9980/wallet/addresses?start=1&limit=5
```

```
{
    "addresses": [
        "2f03d9fcf3de30b905ae5cb992aed0f2a82dac9bc3ddbce630c8fe3b64fa633e6782fde4549e",
        "371bd3af1f153d246403c442ee91c6b042f3d7d1e42b35b61b2f8a2e6f2ff7e9691a51612e68",
        "40315eec5bbb186c3ec02fd385db8bd7125dc42d8709d8cf4cec7f85535ecb5afca1812d7cb7",
        "4267a2f33659f98770b4d90db7a0ac0af4bfb241c27bfc5721b9f3a347129e32cf58118354f9",
        "75ca2454b01b78b1b1aadd69c78f619aac03faf6d045fe012c25b28dfdf1c56f46c91b1cc77e"
    ],
    "pagination": {
        "start": 1,
        "limit": 5,
        "total_pages": 2
    }
}
```

hudaniel:
Also, I'm not understanding the tests. I think the failure are unrelated to my changes?

DavidVorick:
Hey, sorry to let this go for two weeks without a response.

If you rebase to the current master and push, some of the test failures should disappear. The master branch has been in an unstable state for a while, but I think that we've finally cleaned it up.

With regards to the pagination, my preference would be to have a starting item and a maximum number of entries to return. Let's keep it simple, and just start with that for now, instead of also returning links for 'next' and 'prev'.

api/api.go
```diff
@@ -4,11 +4,19 @@ import (
...
+	"math"
+	"strconv"
 
 	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/modules"
 )
 
+const (
+	// size of each page for paginated API responses
+	DEFAULT_PAGINATION_SIZE = 25
```
DavidVorick: We can probably default to something more like 1000. Also, convention is to put constants like this in `consts.go`, and the naming scheme would be `DefaultPaginationSize` instead of using all caps and underscores.

api/api.go
```diff
@@ -29,6 +37,25 @@ type Error struct {
 	// be valid or invalid depending on the current state of a module.
 }
 
+type PaginationRequest struct {
+	HasPaginationQuery 			bool
+	PaginationQueryIsValid 		bool
+	Start 						int
+	Limit 						int
+}
```
DavidVorick: need to run `go fmt` here. If you don't run `go fmt` then the build will always fail.

Best thing to do though is to run `make test-long` before you push code, because that'll do all the formatting, linting, vetting, and of course testing for you. Takes about 10 minutes on a desktop w/SSD.

DavidVorick: I believe that typically we parse requests inside the function, without having a request struct to funnel code into.

api/api.go
```diff
@@ -29,6 +37,25 @@ type Error struct {
...
+	Start 		int `json:"start"`
+	Limit 		int `json:"limit"`
+	TotalPages 	int `json:"total_pages"`
+}
+
+type PaginationWrapper struct {
+	Start 		int
+	End 		int
+	TotalPages 	int
+}
```
DavidVorick: I'm not sure what the purpose of this wrapper is

DavidVorick:
have to run for a call, will continue review later. All functions need docstrings that explain what they do. This code overall is not very well commented.

tbenz9:
@hudaniel I would recommend becoming familiar with this document for naming conventions and Sia developer tips.  Thanks for your work so far, we're excited to see this PR get merged once it's complete.

https://github.com/NebulousLabs/Sia/blob/master/doc/Developers.md

hudaniel:
Thanks for the review, I will address these comments after the holidays!

tbenz9:
@hudaniel Any chance we can get you to revive this PR?  I know several people are still interested in the feature and you have got the best chance of pushing it through.

hudaniel:
Hey, sorry forgot about this haha. Yeah I’ll have some time this weekend to
get this out.

hudaniel:
updated @tbenz9 @DavidVorick

DavidVorick:
Thanks, we will be looking at this as soon as possible.

mtlynch:
Thanks for doing this work! In my [Sia load test](https://blog.spaceduck.io/sia-load-test-preview/), I'm seeing siad latency reaching up to 300+ seconds, so pagination is a welcome feature.

An issue with this implementation is that it doesn't protect the client against changes to the underlying data between paginated calls.

This doesn't feel like such a big deal with transactions because transactions can't be deleted and the API guarantees chronological order, but consider if we apply this to something like [`/renter/files`](https://github.com/NebulousLabs/Sia/blob/master/doc/API.md#renterfiles-get) where items can be removed and the server does not guarantee ordering.

Imagine this scenario:

1. The server has files `[X, Y, Z]`
1. Client calls `/renter/get?start=0&limit=1`
1. Server returns `{file=X, start=0, limit=1, total_pages=3}`
1. Server adds file `W`, so now it has files `[W, X, Y, Z]`
1. Client calls `/renter/get?start=1&limit=1`
1. Server returns `{file=X, start=1, limit=1, total_pages=4}` <- dupe result

What does the client do in that situation? Do they have to throw everything away and start the query over? If the underlying data is changing more frequently than the client can iterate through the total pages, the client would never finish a query.

APIs usually handle this with [page tokens](https://mixmax.com/blog/api-paging-built-the-right-way). BigQuery [does this](https://cloud.google.com/bigquery/docs/paging-results), for example.

Instead of the server reporting the state at the time of each page call, they give the client a token associated with a set of immutable results. Even if the underlying data changes, the server will let the client iterate through the results snapshotted at the time when the page token was created.

Relatedly, it would be nice to allow the client to do a "dry" query where they request no data except the page token. My reading of the current implementation is that the client can't get metadata about the total size of the list unless they request at least one element of data. This is awkward because what most clients want to do is something like:

```golang
token := getPageToken(query)
while token.hasMoreElements() {
  results = append(results, getNextPage(token))
}
```

lukechampine:
@mtlynch That's a nice property to have, but the implementation would be difficult. Could tokens be tacked on afterward without breaking compatibility?

I'm also wondering if pagination is really required for wallet transactions. You can already query them by block height, so shouldn't the client be able to handle the pagination? There also isn't an issue of reordering/invalidation, because new transactions should only be added to the end of the list.

tbenz9:
> I'm also wondering if pagination is really required for wallet transactions.

"All wallet calls which return potentially large amounts of data (like the transactions or the addresses) need to support pagination in the API." --David Vorick in #1478

mtlynch:
>@mtlynch That's a nice property to have, but the implementation would be difficult. Could tokens be tacked on afterward without breaking compatibility?

Would it be possible to make a page token, but just have it get invalidated any time an element is added or removed in the underlying data?

That way, the server doesn't have to maintain snapshots, but the client knows when they need to re-start a paged query because the data shifted. It should be backwards compatible from the client's end because they have to handle page token expiration anyway. If we later improve the server so that it keeps page tokens (and their associated snapshots) valid for 10 mins, then the clients don't have to change their code, but they'll encounter expired tokens much less frequently.

It still has the problem that the client can't achieve a single, consistent "view" of the underlying data, but that doesn't seem critical.

node/api/api.go
```diff
@@ -29,6 +32,25 @@ type Error struct {
 	// be valid or invalid depending on the current state of a module.
 }
 
+type PaginationRequest struct {
```
ChrisSchinnerl: I think we don't need that struct at all. We can just return multiple return values in the methods that use it.

node/api/api.go
```diff
@@ -162,3 +184,77 @@ func WriteJSON(w http.ResponseWriter, obj interface{}) {
...
+		End:        endingPageIndex,
+		TotalPages: totalPages,
+	}
+
+	paginationResponse := PaginationResponse{
+		Start:      startingPageIndex,
+		Limit:      paginationRequest.Limit,
+		TotalPages: totalPages,
+	}
+	return pagination, paginationResponse
```
ChrisSchinnerl: `pagination` and `paginationResponse` have exactly the same fields. Why not just do a 
`return startPageIndex, endingPageIndex, totalPages? That reduces the length of the method by 50%.
You might even want to use named return values instead. That way it is just a `return` at the end of the function.

node/api/api.go
```diff
@@ -162,3 +184,77 @@ func WriteJSON(w http.ResponseWriter, obj interface{}) {
...
+	}
+}
+
+// returns the starting index, end index, and total pages of a given slice for the page
+func GetPaginationIndicesAndResponse(sliceInterface interface{}, paginationRequest PaginationRequest) (PaginationWrapper, PaginationResponse) {
+	if paginationRequest.Limit <= 0 {
+		build.Critical("pagination request limit must be greater than 0")
+	}
+	slice := reflect.ValueOf(sliceInterface)
+	if slice.Kind() != reflect.Slice {
```
ChrisSchinnerl: This seems to be a bit complicated for just getting the length of arbitrary data. Why not just change `sliceInterface interface{}` to `dataLen int` ?

node/api/api.go
```diff
@@ -162,3 +184,77 @@ func WriteJSON(w http.ResponseWriter, obj interface{}) {
...
+		PaginationQueryIsValid: true,
+		Start: startIndex,
+		Limit: limit,
+	}
+}
+
+// returns the starting index, end index, and total pages of a given slice for the page
+func GetPaginationIndicesAndResponse(sliceInterface interface{}, paginationRequest PaginationRequest) (PaginationWrapper, PaginationResponse) {
+	if paginationRequest.Limit <= 0 {
+		build.Critical("pagination request limit must be greater than 0")
```
ChrisSchinnerl: `build.Critical` should be used as a last resort. In production this will only print to the terminal and will continue the execution of the program. In a debug build this panics.
You should rather return an `errors.New("pagination request limit must be greater than 0")` and write that error to the http response later.

node/api/api.go
```diff
@@ -162,3 +184,77 @@ func WriteJSON(w http.ResponseWriter, obj interface{}) {
...
+	}
+	slice := reflect.ValueOf(sliceInterface)
+	if slice.Kind() != reflect.Slice {
+		build.Critical("attempting to paginate on non-slice object type: ", slice.Kind())
+	}
+	startingPageIndex := int(math.Min(float64(paginationRequest.Start), float64(slice.Len())))
+	endingPageIndex := int(math.Min(float64(paginationRequest.Start+paginationRequest.Limit), float64(slice.Len())))
+	totalPages := slice.Len() / paginationRequest.Limit
+	if math.Mod(float64(slice.Len()), float64(paginationRequest.Limit)) != 0 {
```
ChrisSchinnerl: You probably don't need `math.Mod` with all the float casting. Instead you can do `if slice.Len() % paginationRequest.Limit != 0`

node/api/api.go
```diff
@@ -162,3 +184,77 @@ func WriteJSON(w http.ResponseWriter, obj interface{}) {
 func WriteSuccess(w http.ResponseWriter) {
 	w.WriteHeader(http.StatusNoContent)
 }
+
+func GetPaginationDefaultRequest(req *http.Request) PaginationRequest {
+	return GetPaginationRequest(req, "start", "limit")
+}
+
+func GetPaginationRequest(req *http.Request, startQueryParam string, limitQueryParam string) PaginationRequest {
```
ChrisSchinnerl: I would remove the `PaginationRequest` struct and simply return the  `startIndex`, `limit` and an error that indicates a parsing problem. That way the method gets significantly shorter since we don't always construct the `PaginationRequest` object and we can return `build.ExtendErr`-extended errors that indicate which parameter couldn't be parsed.

node/api/api.go
```diff
@@ -162,3 +184,77 @@ func WriteJSON(w http.ResponseWriter, obj interface{}) {
 func WriteSuccess(w http.ResponseWriter) {
 	w.WriteHeader(http.StatusNoContent)
 }
+
+func GetPaginationDefaultRequest(req *http.Request) PaginationRequest {
+	return GetPaginationRequest(req, "start", "limit")
```
ChrisSchinnerl: I'm not sure about the custom names for `start` and `limit`. If we do that, the names of the fields of `PaginationResponse` should probably also be custom to match the fields in the request.

node/api/api.go
ChrisSchinnerl: If you still want to be able to differentiate between a "no pagination params specified" error and a "parsing error" you can declare a custom error for which you can check in the calling method.

node/api/api.go
ChrisSchinnerl: Btw all exported structs and methods need a comment.

node/api/api.go
```diff
@@ -29,6 +32,25 @@ type Error struct {
...
+	Limit      int `json:"limit"`
+	TotalPages int `json:"total_pages"`
+}
+
+type PaginationWrapper struct {
+	Start      int
+	End        int
+	TotalPages int
+}
+
```
ChrisSchinnerl: There are probably enough new pagination types and methods to move the generic ones into a `node/api/pagination.go` file instead of having them in `api.go`.
The ones that are specific to a module can still be in the corresponding `[module].go` file.

node/api/wallet.go
```diff
@@ -179,9 +191,24 @@ func (api *API) walletAddressHandler(w http.ResponseWriter, req *http.Request, _
...
+			WriteError(w, Error{"start and limit queries must be nonnegative integers"}, http.StatusBadRequest)
+			return
+		}
+		pagination, paginationResponse := GetPaginationIndicesAndResponse(allAddresses, paginationRequest)
+		addresses := allAddresses[pagination.Start:pagination.End]
+		WriteJSON(w, WalletAddressesPaginatedGET{
+			Addresses:  addresses,
+			Pagination: paginationResponse,
+		})
+	}
```
ChrisSchinnerl: As mentioned above I would change the call to `GetPaginationRequest` so that you get something like the following:
```
start, limit, err := GetPaginationDefaultRequest(req)
...
if err == errNoPagination {
  // legacy behavior
  return
}
// new behavior

Comment (ChrisSchinnerl) [reply node/api/wallet.go]: You can create a custom error like `errNoPagination` by declaring it at the top of the file. You can find an example in `modules/transactionpool/transactionpool.go` where we declare `errNilCS` and `errNilGateway` at the top.
Of course that also means you have to do a `return errNoPagination` in `GetPaginationRequest` if no params were specified.

File: node/api/wallet.go [hunk 13]
```diff
@@ -521,10 +548,28 @@ func (api *API) walletTransactionsHandler(w http.ResponseWriter, req *http.Reque
...
-	})
+	confirmedPaginationRequest := GetPaginationRequest(req, "confirmedTransactionsStart", "confirmedTransactionsLimit")
+	unconfirmedPaginationRequest := GetPaginationRequest(req, "unconfirmedTransactionsStart", "unconfirmedTransactionsLimit")
+
+	if !confirmedPaginationRequest.HasPaginationQuery || !unconfirmedPaginationRequest.HasPaginationQuery || confirmedTxns == nil || unconfirmedTxns == nil {
+		WriteJSON(w, WalletTransactionsGET{
+			ConfirmedTransactions:   confirmedTxns,
+			UnconfirmedTransactions: unconfirmedTxns,
+		})
+	} else {
```
ChrisSchinnerl: We try to avoid nested `if` statements. You can get rid of the `else` by returning after the call to `WriteJSON`. In the API we usually don't do anything after calling `WriteJSON` or `WriteError` which means you can usually simplify your code by returning right afterwards.

types/constants.go
```diff
@@ -49,6 +49,9 @@ var (
 	SiafundCount     = NewCurrency64(10000)
 	SiafundPortion   = big.NewRat(39, 1000)
 	TargetWindow     BlockHeight
+
+	// size of each page for paginated API responses
+	DefaultPaginationSize = 50
```
ChrisSchinnerl: We usually try to keep the constants in a `consts.go` file within the module it is used in. Since it is only 1 declaration you can probably add `defaultPaginationSize` to a `node/api/pagination.go` file together with the code or move it into a `node/api/consts.go` file.

node/api/api.go
ChrisSchinnerl: May I also suggest a different name for the method? `ParsePaginationParams` might be better after getting rid of the `PaginationRequest` struct.

PR #2512, opened 2017-12-07
Move online check to gateway

ChrisSchinnerl:


modules/renter/repairscanner.go
```diff
@@ -306,7 +306,26 @@ func (r *Renter) threadedRepairScan() {
...
+		}
+		close(channel)
+	}
+
 	for {
+		// Check if we are online. If not, wait until we are
+		if !r.g.Online() {
+			wait := make(chan struct{})
+			go closeWhenOnline(wait)
+			<-wait
```
lukechampine: why a channel here? Wouldn't this work just as well?
```go
for {
    // Wait until we are online
    for !r.g.Online() {
        select {
        case <-r.tg.StopChan():
            return
        case <-time.After(time.Second):
        }
    }

    // Refresh the worker pool and get the set of hosts that are currently
    // useful for uploading.
    hosts := r.managedRefreshHostsAndWorkers()

    // etc...
```

lukechampine: oh, you still need to check `StopChan` in the main loop body though. Otherwise if `r.g.Online` returns true, you'll never check `StopChan`

modules/gateway/gateway.go
```diff
@@ -146,10 +146,11 @@ type Gateway struct {
...
 	// Utilities.
-	log        *persist.Logger
-	mu         sync.RWMutex
-	persistDir string
-	threads    siasync.ThreadGroup
+	log         *persist.Logger
+	mu          sync.RWMutex
+	persistDir  string
+	threads     siasync.ThreadGroup
+	remotePeers int64
```
lukechampine: why do we need `remotePeers`? Can't you calculate the value by looping through the peer list?

ChrisSchinnerl: I didn't want to loop through the whole peer map whenever we delete a key from the map

modules/renter/repairscanner.go
ChrisSchinnerl: Yes, it would. I'll change that. I just thought it looks nicer not having the double for loop :P

modules/gateway/gateway.go
lukechampine: you only loop when you call `Online` though, right? And that's not a call that we make frequently. Even if it were, locking a mutex and looping through <100 pointers isn't a slow operation by any means. Adding a new piece of mutable state introduces the possibility for `remotePeers` to desynchronize from the actual number of remote peers. It's an optimization that we would only make if calculating the value on-the-fly was very expensive.

modules/gateway/peers.go
```diff
@@ -586,7 +586,7 @@ func (g *Gateway) Disconnect(addr modules.NetAddress) error {
 	g.mu.Lock()
 	// Peer is removed from the peer list as well as the node list, to prevent
 	// the node from being re-connected while looking for a replacement peer.
-	delete(g.peers, addr)
+	delete(g.peers, p.NetAddress)
```
lukechampine: should be `addr`

modules/gateway/rpc.go
```diff
@@ -50,7 +50,7 @@ func (g *Gateway) managedRPC(addr modules.NetAddress, name string, fn modules.RP
 		g.log.Debugf("Could not initiate RPC with %v; disconnecting", addr)
 		peer.sess.Close()
 		g.mu.Lock()
-		delete(g.peers, addr)
+		delete(g.peers, peer.NetAddress)
```
lukechampine: ditto

modules/host/storageobligations_smoke_test.go
```diff
@@ -683,25 +686,31 @@ func TestAutoRevisionSubmission(t *testing.T) {
...
 	}
-	if !so.OriginConfirmed {
-		t.Fatal("origin transaction for storage obligation was not confirmed after blocks were mined")
-	}
-	if !so.RevisionConfirmed {
-		t.Fatal("revision transaction for storage obligation was not confirmed after blocks were mined")
-	}
-	if !so.ProofConfirmed {
-		t.Fatal("storage obligation is not saying that the storage proof was confirmed on the blockchain")
-	}
```
lukechampine: ?

ChrisSchinnerl: I wrapped it in a `retry` since it seemed to fail randomly.

modules/gateway/peers.go
```diff
@@ -604,3 +604,20 @@ func (g *Gateway) Peers() []modules.Peer {
...
+		return true
+	}
+
+	g.mu.RLock()
+	defer g.mu.RUnlock()
+	for _, p := range g.peers {
+		if !p.Local {
+			return true
+		}
+	}
```
DavidVorick: Eventually I'd like to see this switched to an atomic value that gets updated any time that we adjust the `g.peers` set, which means that we can call this function without having to do anything with locking.

modules/gateway/peers.go
```diff
@@ -604,3 +604,20 @@ func (g *Gateway) Peers() []modules.Peer {
 	}
 	return peers
 }
+
+// Online returns true if the node is connected to the internet. During testing
+// we always assume that the node is online
+func (g *Gateway) Online() bool {
+	if build.Release != "standard" {
```
DavidVorick: Should instead check whether it equals dev or testing. We're going to be adding a testnet release constant soon, and I don't want to miss this patch of code. It's much less noticeable if you are accidentally always saying online if you can actually be both online or offline vs. always saying offline when you shouldn't be.

DavidVorick: basically we're inverting this because if we mess up after adding future constants, it'll be more noticeable that we messed up.

modules/host/storageobligations_smoke_test.go
DavidVorick: separate concern I think that belonged in a separate PR, but that's okay, we'll leave it for now.

modules/renter/repairscanner.go
```diff
@@ -307,7 +307,16 @@ func (r *Renter) threadedRepairScan() {
 	defer r.tg.Done()
 
 	for {
-		// Return if the renter has shut down.
+		// Wait until we are online
+		for !r.g.Online() {
+			select {
+			case <-r.tg.StopChan():
+				return
+			case <-time.After(time.Second):
```
DavidVorick: Check every 20 seconds.

modules/gateway/peers.go
DavidVorick: no need to do that this PR, but we should probably open an issue for it labeled `Programming Task` on the issue tracker.

lukechampine: as I said in another comment, I don't think we need to worry about this code being highly performant. We aren't going to be calling `Online` very frequently, and the lock is held for a very short period. Adding more shared mutable state increases the likelihood of bugs and makes it harder for other devs to figure out how the value is calculated.

ChrisSchinnerl: An atomic variable would be overkill since we only update it when we update the peers for which we need a lock anyway.

PR #2514, opened 2017-12-08
Siatest pkg

DavidVorick:
Create a siatest package. This is more or less intended to replace all the integration testing that we do in the api package. There's a lot of repeat code throughout our test suite, and this package gives us a means to easily replace it, while also improving the amount of control we have over the tests we write.

Hopefully, using this package, we'll be able to spin up more advanced + continuous test networks that can probe some of the edge cases we've been seeing - like errors during simultaneous upload + download.

Eventually we'll be extending this package to include comprehensive dependency injection as well. That will allow us to test random disk failures, and random network failures.


DavidVorick:
The first intended use for this PR is to add some integration testing to the contract set, particularly with regards to testing the ACID properties, and verifying that host contracts can survive sudden interruptions at key moments of the contract negotiation.

lukechampine:
Can this PR include converting an existing test to use `siatest`, so that we can see what it looks like in practice?

DavidVorick:
Closing in favor of updated work

PR #2515, opened 2017-12-08
add comment about half-life being 144 data points

annalissac:


annalissac:
based on comments made in core-dev channel on 12/7/2017

PR #2519, opened 2017-12-10
Sending the detected public IP address in sessionHeader

dNetGuru:
See #2066.

Only side effect here is that for the first couple of seconds where the external IP is undetected all session headers will be invalid.


lukechampine:
Thanks for this. I don't think we'll reinstate the check, because it adds little value and risks disconnecting valid peers, but clearly it's better to send the external IP.

As you mentioned, my only concern here is the behavior when the external IP is unknown. Perhaps we could check for that, and send `conn.LocalAddr` if we don't know our external IP?

dNetGuru:
I'll look into it. I guess that might be a good idea since the newer nodes (1.3.1+) wouldn't care if the internal IP is sent or the external. However this would make for a very useless output from ```gateway list``` since it will be populated with a bunch of invalid IP addresses (could just output the ```remoteAddr``` instead of the claimed address on the sessionHeader)

 I'm running a node with this change implemented to see how it behaves. So far I'm seeing a very large number of valid connections that are coming in with an invalid IP which would have failed otherwise (even though I'm running 1.3.1 they would have rejected my header).

```
dnetguru@dHost:~$ siac gateway list | awk '{print $3}' | grep -P "^((172\.(1[6-9]|2[0-9|3[0-1])|192\.168|10\.\d{1,3})\.\d{1,3}\.\d{1,3})" | wc -l
82
dnetguru@dHost:~$ netstat -nt | awk '{print $5}' | grep 9981 | grep -P "((172\.(1[6-9]|2[0-9|3[0-1])|192\.168|^10\.\d{1,3})\.\d{1,3}\.\d{1,3})" | wc -l
0
```
As you can see there are some 82 hosts that have connected to my node with an invalid IP in their sessionHeader but none are actually from my local network. I guess this might be causing a lot of nodes not to communicate with a good number of other nodes.

About detecting the external IP, might I suggest using DNS instead of the HTTP service you are using now? OpenDNS always resolves ```myip.opendns.com``` to the caller's public IP address. 
```
dnetguru@dHost:~$ dig +short myip.opendns.com @208.67.222.222
291.239.58.162
```
As you know DNS protocol is much more lightweight so this will happen much faster (and is guaranteed to get through firewalls whereas I get "Warning: Unsafe Website Blocked! myexternalip.com" from COMODO's SecureDNS) and OpenDNS has a large number of geographically scattered servers so you are not relying on just the two servers ```myexternalip.com``` provides.

It seems like https://github.com/miekg/dns provides a facility to query a specific nameserver. I don't have much experience with Golang so you might know of a better way/library to implement that.

dNetGuru:
Okay @lukechampine, I made some minor changes to check for all zeros IP addresses being assigned to ```myAddr``` (when the socket is not bound to any specific interface since the initial value comes from ```g.listener.Addr().String()```) and am now sending the local IP before the external IP is detected. I also added a debug log message to see it in action.

Here's a run in which I immediately issued a ```siac gateway connect``` after running ```siad``` and another one a little later when it had the time to detect the external IP.

```
2017/12/11 04:03:36.220589 gateway.go:211: STARTUP: Logging has started. Siad Version 1.3.1
2017/12/11 04:03:36.220656 gateway.go:223: INFO: gateway created, started logging
2017/12/11 04:03:36.223555 peersmanager.go:78: [DEBUG] INFO: [PPM] Permanent peer manager has started
2017/12/11 04:03:36.223598 peersmanager.go:86: [DEBUG] [PPM] Node list is empty, sleeping
2017/12/11 04:03:42.198229 peers.go:446: [DEBUG] Sending sessionHeader with address 192.168.92.128:9981 true
2017/12/11 04:03:46.223942 peersmanager.go:86: [DEBUG] [PPM] Node list is empty, sleeping
2017/12/11 04:03:53.472996 upnp.go:99: WARN: could not automatically forward port 9981: no UPnP-enabled devices found: no UPnP-enabled gateway found
2017/12/11 04:03:56.223975 nodes.go:245: ERROR: could not pick a random node for uptime check: no peers
2017/12/11 04:03:56.224125 peersmanager.go:86: [DEBUG] [PPM] Node list is empty, sleeping
2017/12/11 04:04:06.224720 peersmanager.go:86: [DEBUG] [PPM] Node list is empty, sleeping
2017/12/11 04:04:16.224481 nodes.go:245: ERROR: could not pick a random node for uptime check: no peers
2017/12/11 04:04:16.225377 peersmanager.go:86: [DEBUG] [PPM] Node list is empty, sleeping
2017/12/11 04:04:23.475917 upnp.go:83: INFO: our address is 50.140.99.421:9981
2017/12/11 04:04:26.226410 peersmanager.go:86: [DEBUG] [PPM] Node list is empty, sleeping
2017/12/11 04:04:36.225105 nodes.go:245: ERROR: could not pick a random node for uptime check: no peers
2017/12/11 04:04:36.227112 peersmanager.go:86: [DEBUG] [PPM] Node list is empty, sleeping
2017/12/11 04:04:46.228180 peersmanager.go:86: [DEBUG] [PPM] Node list is empty, sleeping
2017/12/11 04:07:37.103978 peers.go:446: [DEBUG] Sending sessionHeader with address 50.140.99.421:9981 false
```

I will take a stab at using OpenDNS for external IP detection next weekend.

lukechampine:
Actually, can we do this? In `gateway/gateway.go` line 284, we can do the empty hostname check you added. Then we only need to worry about setting it once, and in one place:
```go
// Set the address and port of the gateway
host, port, _ := net.SplitHostPort(g.listener.Addr().String())
if host == "" {
    // if host is empty, set a dummy one for now. It will be overwritten
    // by threadedLearnHostname later on.
    host = "localhost"
}
g.port = port
g.myAddr = modules.NetAddress(net.JoinHostPort(host, port))
```

Setting `host = "localhost"` is technically not as accurate as setting it to `conn.LocalAddr`, but I don't think it matters since neither will be connectible. Thoughts?

lukechampine:
I'm onboard with switching to DNS lookup too. :+1:

dNetGuru:
@lukechampine ``host == ""``` will never evaluate to true!

If the socket is bound to a specific interface we'll get that interface's IP address (say it was bound to eth0 and eth0's address was 192.168.1.13) OR it is not bound to a specific interface (which I think will always be the case, although I haven't studied the whole source code yet) in that case ```host``` will either be ```0.0.0.0``` for IPv4 or ```::``` for IPv6. Which is why I'm testing for ```na.Host() == "0.0.0.0" || na.Host() == "::"```.
Neither those addresses are connectible so the host would become unconnectable before the external IP is detected. By sending the local IP we still keep the host accessible in case the external IP never ends up getting resolved (for any reason).

If you want minimal overhead, we could simply set it to 127.0.0.1:9981 while we are waiting for ```threadedLearnHostname``` to do its job, since as it is written right now ```g.listener.Addr()``` is always likely to return [::]:9981 or 0.0.0.0:9981. Plus as far as the remote client is concerned the actual local interface address (say 192.168.10.13) and the loopback address (127.0.0.1) are equally unroutable and useless outside the home router of the connecting client.

What do you think?

lukechampine:
Hmm. I don't want to always set the hostname at startup, because the user may want to bind to a specific hostname. (We don't currently support this, but there is some demand for it: see https://github.com/NebulousLabs/Sia/issues/1970.) So let's do the following:
```go
// Set the address and port of the gateway
host, port, _ := net.SplitHostPort(g.listener.Addr().String())
if ip := net.ParseIP(host); ip != nil && ip.IsUnspecified() {
    // if host is unspecified, set a dummy one for now. It will be overwritten
    // by threadedLearnHostname later on.
    host = "localhost"
}
g.port = port
g.myAddr = modules.NetAddress(net.JoinHostPort(host, port))
```
And then in peers.go, when we construct `ourHeader`:
```go
g.mu.RLock()
ourHeader := sessionHeader{
	GenesisID:  types.GenesisID,
	UniqueID:   g.id,
	NetAddress: g.myAddr,
}
g.mu.RUnlock()

**Comment by dNetGuru** [comment]
@lukechampine it seems to work just fine
I sent two connection requests here, one before the external IP was detected, and one after
```
2017/12/11 22:30:02.962799 gateway.go:211: STARTUP: Logging has started. Siad Version 1.3.1
2017/12/11 22:30:02.962852 gateway.go:223: INFO: gateway created, started logging
2017/12/11 22:30:02.963555 peersmanager.go:78: [DEBUG] INFO: [PPM] Permanent peer manager has started
2017/12/11 22:30:12.964284 peersmanager.go:86: [DEBUG] [PPM] Node list is empty, sleeping
2017/12/11 22:30:22.965370 peersmanager.go:86: [DEBUG] [PPM] Node list is empty, sleeping
2017/12/11 22:30:22.965461 nodes.go:245: ERROR: could not pick a random node for uptime check: no peers
2017/12/11 22:30:23.706193 peers.go:442: [DEBUG] Sending sessionHeader with address localhost:9981 true
2017/12/11 22:30:25.118711 upnp.go:83: INFO: our address is 50.140.99.421:9981
2017/12/11 22:30:32.965791 peersmanager.go:86: [DEBUG] [PPM] Node list is empty, sleeping
2017/12/11 22:30:40.728496 upnp.go:99: WARN: could not automatically forward port 9981: no UPnP-enabled devices found: no UPnP-enabled gateway found
2017/12/11 22:30:42.966482 peersmanager.go:86: [DEBUG] [PPM] Node list is empty, sleeping
2017/12/11 22:30:42.966499 nodes.go:245: ERROR: could not pick a random node for uptime check: no peers
2017/12/11 22:30:45.836373 peers.go:442: [DEBUG] Sending sessionHeader with address 50.140.99.421:9981 false
```

dNetGuru:
I've been running a node with this modification and currently I have 334 peers connected to that node as opposed to 8~10 that are connected to my vanilla 1.3.0 node. What's the usual number of incoming peers you've seen?

lukechampine:
Awesome, this should help users who are currently getting stuck.

PR #2520, opened 2017-12-11
fix the difficulty adjustment algorithm

DavidVorick:


modules/consensus/difficulty.go
```diff
@@ -68,10 +68,17 @@ var (
...
+	// Determine the delta of the current total time vs. the desired total time.
+	// The desired total time is the difference between the genesis block
+	// timestamp and the current block timestamp.
+	var expectedTime int64
+	if parentHeight < types.OakHardforkFixBlock {
+		expectedTime = int64(types.BlockFrequency * parentHeight)
+	} else {
+		expectedTime = int64(types.BlockFrequency*parentHeight) + int64(types.GenesisTimestamp)
+	}
+	delta := int64(expectedTime) - int64(parentTimestamp)
```
DavidVorick: This is the only change of substance in this pull request. Instead of comparing against an uncompressed timestamp, we compare against an absolute timestamp. We switch from one to the other at the block height specified by the constants.

modules/consensus/difficulty.go
```diff
@@ -68,10 +68,31 @@ var (
...
+		// This is the correct code. The expected time is an absolute time based
+		// on the genesis block, and the delta is an absolute time based on the
+		// timestamp of the parent block.
+		//
+		// Rules elsewhere in consensus ensure that the timestamp of the parent
+		// block has not been manipulated by more than a few hours, which is
+		// accurate enough for this logic to be safe.
+		expectedTime := int64(types.BlockFrequency*parentHeight) + int64(types.GenesisTimestamp)
+		delta = expectedTime - int64(parentTimestamp)
+	}
```
DavidVorick: This is the only change of substance in this PR. The first branch of logic in the `if` block should exactly match the original code. Please verify from the diff that this is the case.

The original block of code was incorrectly setting the delta by comparing an absolute time (`expectedTime`) to a time that gets compressed every block (`parentTotalTime`), which always resulted in a massive positive delta, causing the shifter to always target 30 minute blocks.

The new code correctly compares an absolute time (`expectedTime`) to an absoluteTime (`parentTimestamp`).

modules/consensus/difficulty.go
```diff
@@ -82,11 +103,11 @@ func (cs *ConsensusSet) childTargetOak(parentTotalTime int64, parentTotalTarget,
...
 	// Clamp the block time to 1/3 and 3x the target block time.
-	if targetBlockTime < int64(types.BlockFrequency)/3 {
-		targetBlockTime = int64(types.BlockFrequency) / 3
+	if targetBlockTime < int64(types.BlockFrequency)/types.OakMaxBlockShift {
+		targetBlockTime = int64(types.BlockFrequency) / types.OakMaxBlockShift
 	}
-	if targetBlockTime > int64(types.BlockFrequency)*3 {
-		targetBlockTime = int64(types.BlockFrequency) * 3
+	if targetBlockTime > int64(types.BlockFrequency)*types.OakMaxBlockShift {
+		targetBlockTime = int64(types.BlockFrequency) * types.OakMaxBlockShift
```
DavidVorick: Turned a magic number into a consensus constant.

modules/consensus/difficulty.go
```diff
@@ -95,19 +116,27 @@ func (cs *ConsensusSet) childTargetOak(parentTotalTime int64, parentTotalTarget,
...
 	visibleHashrate := parentTotalTarget.Difficulty().Div64(uint64(parentTotalTime)) // Hashes per second.
+	// Handle divide by zero risks.
 	if visibleHashrate.IsZero() {
 		visibleHashrate = visibleHashrate.Add(types.NewCurrency64(1))
 	}
+	if targetBlockTime == 0 {
+		// This code can only possibly be triggered if the block frequency is
+		// less than 3, but during testing the block frequency is 1.
+		targetBlockTime = 1
+	}
```
DavidVorick: Fixed a divide-by-zero which could only exist during testing, but nonetheless could exist.

modules/consensus/difficulty.go
DavidVorick: This change does require us to provide the parent timestamp in the function signature, which resulted in us changing `processedblock.go` to correctly use the new function signature.

modules/consensus/processedblock.go
```diff
@@ -144,7 +144,7 @@ func (cs *ConsensusSet) newChild(tx *bolt.Tx, pb *processedBlock, b types.Block)
 	if pb.Height < types.OakHardforkBlock {
 		cs.setChildTarget(blockMap, child)
 	} else {
-		child.ChildTarget = cs.childTargetOak(prevTotalTime, prevTotalTarget, pb.ChildTarget, pb.Height)
+		child.ChildTarget = cs.childTargetOak(prevTotalTime, prevTotalTarget, pb.ChildTarget, pb.Height, pb.Block.Timestamp)
```
DavidVorick: New function requires us to pass in the parent block timestamp.

modules/consensus/difficulty.go
lukechampine: the only thing that could potentially be clearer here is changing the last two lines to:
```go
expectedTime := int64(types.BlockFrequency*parentHeight)
delta = expectedTime - int64(parentTimestamp - types.GenesisTimestamp)
```
that way, the `expectedTime` is the same for both branches. In fact, you could do:
```go
var parentTime int64
if parentHeight < types.OakHardforkFixBlock {
    parentTime = parentTotalTime
} else {
    parentTime = int64(parentTimestamp - types.GenesisTimestamp)
}
delta := expectedTime - parentTime
```
but regardless, the logic seems correct.

PR #2521, opened 2017-12-11
Use remote address from opened socket instead of the claimed IP in peer list

dNetGuru:
Currently ```siac gateway list``` shows the claimed IP address sent in a sessionHeader. There are two main issues with this approach.
Since as it stands the local interface IP is sent in the sessionHeader (instead of the external IP) we end up with invalid IP addresses shown for inbound connections when we execute ```siac gateway list``` that don't correspond to actual open sockets in the system. This makes correlating outputs from network monitoring utilities like ```netstat``` and the likes with the output from ```siac``` impractical.
Example output from ```siac gateway list```
```
root@siad-dev:/home/siad-dev/Sia/gateway# siac gateway list | grep No | head -n 5
1.3.0    No        192.168.70.246:9981
1.3.0    No        192.168.0.20:9981
1.3.0    No        192.168.0.114:9981
1.3.0    No        192.168.199.225:9981
1.3.0    No        192.168.1.21:9981
```
Second, we are blindly trusting the remote party not to lie about their network address which will be shown in the peer list (since network check was removed in #2254).

### DIFF
--- a/modules/gateway/peers.go
+++ b/modules/gateway/peers.go
@@ -195,7 +195,9 @@ func (g *Gateway) managedAcceptConnv130Peer(conn net.Conn, remoteVersion string)
 			// NOTE: local may be true even if the supplied NetAddress is not
 			// actually reachable.
 			Local:      remoteAddr.IsLocal(),
-			NetAddress: remoteAddr,
+			// Ignoring claimed IP address (which should be == to the socket address)
+			// by the host but keeping note of the port number so we can call back
+			NetAddress: modules.NetAddress(net.JoinHostPort(remoteAddr.Host(), remoteHeader.NetAddress.Port())),
 			Version:    remoteVersion,
 		},
 		sess: newServerStream(conn, remoteVersion),
--- a/modules/gateway/peers.go
+++ b/modules/gateway/peers.go
@@ -185,14 +185,17 @@ func (g *Gateway) managedAcceptConnv130Peer(conn net.Conn, remoteVersion string)
 		return err
 	}
 
+	// Get the remote address from opened socket
+	remoteAddr := modules.NetAddress(conn.RemoteAddr().String())
+
 	// Accept the peer.
 	peer := &peer{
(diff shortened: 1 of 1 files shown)


lukechampine:
Perhaps we can extend `modules.Peer` to include the `ClaimedAddress`? Then there's no confusion about whether `NetAddress` is the "real" address or the "claimed" address, and `siac` can simply display the real address.

dNetGuru:
The only question I have is why?! I understand you put in the ```ClaimedAddress``` in sessionHeader so that it could later be checked against the socket address. However you scrapped that idea and opted not to run that check.
I might have missed it but is the ```ClaimedAddress`` being used anywhere?

With that said, if you think this might be somehow useful for whatever you have planned in your roadmap then sure. I'll add that.

DavidVorick:
I am always slow to remove things, but racking my brains, I can't really think of a good reason to keep the `ClaimedAddress` field around. We can't merge this before v1.3.1 because we can't risk any new bugs in our peer network right now, but if v1.3.1 has a stable peer network we can come back to this.

lukechampine:
@dNetGuru we need to send at least the port number, so that the recipient know how to dial back later. The hostname doesn't serve any purpose now, but it's possible that it could be useful later.
I'm ok not keeping the `ClaimAddress` around as long as we can demonstrate that everything works normally without it.

dNetGuru:
I see! If we need to call back, I need to edit my commit so the claimed port number is saved instead of the incoming one.

I don't think currently your callback feature works (except on a local network) since the claimedAddress will be a local address for incoming connections over the internet which would be unrourable.

dNetGuru:
I changed the code slightly so we take note of the ```claimedAddress```'s port number so we can call back but save the open socket's address instead of the claimed address in the header.

Later on, you might want to refactor the sessionHeader to just send a ```callbackPort``` field instead of a ```netaddress``` but that would break backward compatibility so I'm not touching that.

PR #2527, opened 2017-12-13
increase scoreLeeway

ChrisSchinnerl:


DavidVorick:
An RC2 adjustment to give hosts sightly more room for failure. Eventually we would like to see the leeway drop below 5, but it's clear that the network today is not stable enough for that to be a practical decision. As the network grows, especially with the introduction of proof of burn, this number can be brought back down.

PR #2532, opened 2017-12-13
don't upload unless there are enough good hosts

ChrisSchinnerl:
We don't want to upload a chunk unless we are sure there are potentially enough good contracts for that chunk.
This PR also moves the cooldown check into a separate function and lets the worker remember the last error that caused it to get penalized.


lukechampine:
Closing because (IIUC) this was addressed by David's upload overhaul

PR #2537, opened 2017-12-14
adjust pricing leeway for more stable uploads

DavidVorick:


ChrisSchinnerl:
I ran this on the live network and it seems to upload smoothly so far.

PR #2541, opened 2017-12-14
rename contractset.log to contractset.wal

ChrisSchinnerl:
This PR renames the contractset to have the 'wal' extension.


DavidVorick:
You should only perform this rename if `contractset.wal` doesn't already exist. If it already exists, you shouldn't even check if `contractset.log` exists.

PR #2550, opened 2017-12-16
Drop siacoin outputs if we failed to build transaction

ChrisSchinnerl:
It seems like we didn't drop outputs when we use the wallets sending functions. This might cause issues.

### DIFF
--- a/modules/wallet/money.go
+++ b/modules/wallet/money.go
@@ -183,7 +183,7 @@ func (w *Wallet) SendSiacoinsMulti(outputs []types.SiacoinOutput) (txns []types.
 		return nil, build.ExtendErr("unable to sign transaction", err)
 	}
 	if w.deps.disrupt("SendSiacoinsInterrupted") {
-		return nil, errors.New("Failed to accept transaction set. (SendSiacoinsInterrupted)")
+		return nil, errors.New("failed to accept transaction set (SendSiacoinsInterrupted)")
 	}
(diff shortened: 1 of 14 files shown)


DavidVorick:
This pull request needs a regression test. Need to write some test that fails because we aren't dropping the tx but succeeds if we start dropping the tx correctly.

Most natural way to do this is probably to set up the transaction pool to fail - an error on AcceptTransactionSet that causes the transaction to stumble and then not get dropped.

modules/wallet/money.go
```diff
@@ -112,6 +119,9 @@ func (w *Wallet) SendSiacoins(amount types.Currency, dest types.UnlockHash) ([]t
 		w.log.Println("Attempt to send coins has failed - failed to sign transaction:", err)
 		return nil, build.ExtendErr("unable to sign transaction", err)
 	}
+	if w.deps.disrupt("SendSiacoinsInterrupted") {
+		return nil, errors.New("Failed to accept transaction set. (SendSiacoinsInterrupted)")
```
DavidVorick: errors are not supposed to have punctuation, e.g. `failed to accept transaction set (SendSiacoins interrupted)`

modules/wallet/money.go
```diff
@@ -167,6 +182,9 @@ func (w *Wallet) SendSiacoinsMulti(outputs []types.SiacoinOutput) ([]types.Trans
...
```
DavidVorick: ditto

PR #2555, opened 2017-12-18
always send conn.LocalAddr in sessionHeader

lukechampine:
Since older nodes reject hostnames that don't match the connection, just send whatever hostname the connection reports. This should resolve the issue of IPv6 peers being rejected by IPv4 peers.

cc @dNetGuru

### DIFF
--- a/modules/gateway/peers.go
+++ b/modules/gateway/peers.go
@@ -166,13 +166,14 @@ func acceptableSessionHeader(ourHeader, remoteHeader sessionHeader, remoteAddr s
 // The requesting peer is added as a node and a peer. The peer is only added if
 // a nil error is returned.
 func (g *Gateway) managedAcceptConnv130Peer(conn net.Conn, remoteVersion string) error {
-	g.log.Debugln("Sending sessionHeader with address", g.myAddr, g.myAddr.IsLocal())
 	// Perform header handshake.
 	g.mu.RLock()
+	host, _, _ := net.SplitHostPort(conn.LocalAddr().String())
+	ourAddr := modules.NetAddress(net.JoinHostPort(host, g.myAddr.Port()))
 	ourHeader := sessionHeader{
 		GenesisID:  types.GenesisID,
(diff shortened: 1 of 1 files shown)


dNetGuru:
@lukechampine but this will break the IPv4 nodes behind NAT. What version does this conflict with? I'm going to run some tests.

lukechampine:
oh, true. NAT will change the hostname, so `conn.LocalAddr` won't match the receiver's `conn.RemoteAddr`.

Since this only affects connections to v1.3.0 nodes, I think it's best to keep things as they are. Once enough nodes have upgraded, it'll cease to be an issue.

dNetGuru:
This got me thinking and there is something to the point you brought up @lukechampine. It might be a good idea to detect if the host has both IPv6 and IPv4 or just IPv6 enabled and decide accordingly. Shouldn't be that hard to implement and since I'm working on the PR for using DNS to detect external IP and I'll incorporate this test in it too.

Reason being that in the event that we detect both IPv4 and IPv6 on the host then we should send both addresses so IPv4 only hosts still can connect back.

Which makes me think about what hosts that have both IPv4 and IPv6 should do when announcing their address on the blockchain! Should they publish their IPv6 (which would mean all the IPv4 hosts won't connect) or the other way around? Or perhaps they could announce both?

Something to think about.

lukechampine:
@dNetGuru one solution to the host announcement problem is to announce a domain name that can be resolved to either IPv4 or IPv6. Announcing a domain name is a good idea anyway if you don't have a static IP.
My guess is that everyone will speak IPv4 for at least another decade, so hosts will probably want to exclusively announce their IPv4 address.

dNetGuru:
@lukechampine so would you be okay with changing the external IP detection mechanism to detect the IPv4 address even when there is an IPv6 address available? (since now the address is only used to announce and the claimedAddress check is gone)
Right now myexternalip.com/raw always returns the IPv6 address if available

PR #2559, opened 2017-12-19
Sending interface address when connecting to local OR IPv6 nodes

dNetGuru:
Fixing the problem #2555 was trying to fix without breaking NAT.

cc: @lukechampine

### DIFF
--- a/modules/gateway/peers.go
+++ b/modules/gateway/peers.go
@@ -166,15 +166,24 @@ func acceptableSessionHeader(ourHeader, remoteHeader sessionHeader, remoteAddr s
 // The requesting peer is added as a node and a peer. The peer is only added if
 // a nil error is returned.
 func (g *Gateway) managedAcceptConnv130Peer(conn net.Conn, remoteVersion string) error {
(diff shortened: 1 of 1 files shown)


dNetGuru:
Fornax helped test this and the results are satisfactory. The only problem right now is that if ```siad``` is being run on a dual stack machine then ```http://myexternalip.com/raw``` always returns the IPv6 and that is set as the public IP and is sent indiscriminately to both IPv6 and IPv4 peers. This is only a problem for outgoing connections.

Detecting both IPv6 and IPv4 addresses of the host is certainly possible and I could add that later to the DNS resolver so the host can communicate on both stacks.

ChrisSchinnerl:
@dNetGuru your PR has a conflict. Probably because of [PR#2739](https://github.com/NebulousLabs/Sia/pull/2739). We changed the code to use the announced port but the IP from the connection object because of [#2735](https://github.com/NebulousLabs/Sia/issues/2735).

dNetGuru:
Cool!

PR #2568, opened 2017-12-21
Added 1.3.1 changelog notes

tbenz9:


lukechampine:
ah, I knew there was something I was forgetting...
we try to keep the release changelog to three items, so let's do:

- Add new efficient, reliable contract format
- Faster and smoother file repairs
- Fix difficulty adjustment hardfork

PR #2579, opened 2017-12-24
adding 'expected behavior' to bug report template

tbenz9:


lukechampine:
It seems like very few people are filling out the template properly. Wonder if there's something we can do about that.

tbenz9:
@lukechampine, We could add instructions to the beginning of the template, this has a chance of just adding more cruft to each issue, or helping people use it more effectively.  It's probably worth a try.

Here's a sample of a big project using templates for bug/feature requests.  They offer clear instructions.
https://github.com/kubernetes/kubernetes/issues/new

lukechampine:
Ah, looks like the kubernetes template puts the instructions inside `<!--` tags, which means they're hidden when rendered, even if the user forgets to delete them. We should do the same.

tbenz9:
Sounds good, PR coming in a few hours.

PR #2588, opened 2017-12-29
expose the api server from the api package

DavidVorick:
After some feedback from the community, I realized we should be exposing both server and client functions from the api package. Right now if you want to create your own Sia instance in go code, you have to create your own listener and server. Now we've got functions to do that.

You still have to create all of your own modules and dependency chains, so while this removes one of the annoying steps, it doesn't remove the biggest one.


lukechampine:
Looks good, just need to remove some errant printlns

PR #2595, opened 2017-12-31
use syscall.Stdin for stdin

### LINKED ISSUE #2587: password/seed input prompt in siac 1.3.1 broken on Windows
Bug Report
==========
It appears that when a password/seed input prompt is invoked in siac, an error is triggered giving the output below. There is no actual input accepted, the command exits immediately.

Issue is reproducible in both 'cmd' and 'powershell'. Tried executing "Get-Content seed.txt | .\siac wallet init-seed" in powershell, but doesn't help.

> .\siac wallet unlock
Wallet password:
Reading password failed: The handle is invalid.

> .\siac wallet init-seed
Seed:
Reading seed failed: The handle is invalid.

**Environment**
* Sia version: 1.3.1 binaries downloaded from https://sia.tech/apps/ on 27. Dec
* OS: Windows 10

DavidVorick:
fixes #2587

### DIFF
--- a/cmd/siac/walletcmd.go
+++ b/cmd/siac/walletcmd.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"math/big"
 	"os"
+	"syscall"
 
 	"github.com/spf13/cobra"
 	"golang.org/x/crypto/ssh/terminal"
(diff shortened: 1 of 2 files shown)


DavidVorick:
If anyone has any good ideas for how to write a regression test for this, we can probably run something on appveyor.

PR #2599, opened 2018-01-02
add bucket to allow for constant time lookups for /transaction endpoint

ChrisSchinnerl:


lukechampine:
This will need some compat code to generate `bucketProcessedTxnKey` for existing wallets.

I also think I slightly prefer `Index` to `Key`, if only because `Key` sounds like it could relate to crypto keys somehow.

DavidVorick:
Need test code that demonstrates the compat code functions correctly when upgrading an old wallet. Also we need some indication of how bad the performance hit is when performing the upgrade.

ChrisSchinnerl:
@DavidVorick I changed the compat test to create a wallet with ~2000 txns, delete the bucket, close the wallet, restart the wallet and check if the bucket is correctly repopulated afterwards.
The test runs in 12s but most of that time goes into mining 1000 blocks.

DavidVorick:
Looks good to me, but would prefer to hold off on merging it until we've got the vlong test merged as well.

PR #2605, opened 2018-01-03
add node creation framework

DavidVorick:
Now we've got an easy function to create a node, and templates that allow you to easily choose from some sane defaults. It's also pretty easy to start from a template by making a copy, and then make just a few tweaks to get what you need without having to fill out the whole struct yourself.

The goal here was to make it easy to create a Sia instance.

Next up will be the `server` package, and then the `client` package.

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -117,12 +117,12 @@ func (n *Node) Close() (err error) {
 	return err
 }
 
-// NewNode will create a new test node. The inputs to the function are the
-// respective 'New' calls for each module. We need to use this awkward method
-// of initialization because the siatest package cannot import any of the
-// modules directly (so that the modules may use the siatest package to test
+// New will create a new test node. The inputs to the function are the
+// respective 'New' calls for each module. We need to use this awkward method of
+// initialization because the siatest package cannot import any of the modules
+// directly (so that the modules may use the siatest package to test
 // themselves).
-func NewNode(params NodeParams) (*Node, error) {
+func New(params NodeParams) (*Node, error) {
 	dir := params.Dir
 
 	// Gateway.
--- a/node/node.go
+++ b/node/node.go
@@ -1,0 +1,289 @@
+// Package node provides tooling for creating a Sia node. Sia nodes consist of a
+// collection of modules. The node package gives you tools to easily assemble
+// various combinations of modules with varying dependencies and settings,
+// including templates for assembling sane no-hassle Sia nodes.
+package node
+
+// TODO: Add support for the explorer.
+
+// TODO: Add support for custom dependencies for all of the modules.
+
+import (
+	"path/filepath"
+
+	"github.com/NebulousLabs/Sia/modules"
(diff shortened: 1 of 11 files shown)


node/templates_test.go
```diff
@@ -0,0 +1,92 @@
...
+	if testing.Short() {
+		t.SkipNow()
+	}
+
+	// Test AllModulesTemplate.
+	dir, err := siatest.TestDir("node", t.Name()+"-AllModulesTemplate")
+	if err != nil {
+		t.Fatal(err)
+	}
+	n, err := NewNode(AllModules(dir))
```
DavidVorick: creating a node is pretty easy. If you already have the dir, just call `NewNode(AllModules(dir))`

node/templates_test.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Test WalletTemplate.
+	dir, err = siatest.TestDir("node", t.Name()+"-WalletTemplate")
+	if err != nil {
+		t.Fatal(err)
+	}
+	n, err = NewNode(Wallet(dir))
```
DavidVorick: Or maybe you only need a wallet. Then call `NewNode(Wallet(dir))`

Note that out of package, this would be `nodes.NewNode(nodes.Wallet(dir))`

DavidVorick:
going to switch it from `nodes.NewNode` to just `nodes.New`

node/templates_test.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Test WalletTemplate.
+	dir, err = siatest.TestDir("node", t.Name()+"-WalletTemplate")
+	if err != nil {
+		t.Fatal(err)
+	}
+	n, err = New(Wallet(dir))
```
DavidVorick: for an outside package, this would read `nodes.New(nodes.Wallet(dir))`

DavidVorick: If you wanted something custom, (say wallet + renter), you could do:

```
template := nodes.WalletTemplate
template.CreateRenter = true
n, err := nodes.New(template)
```

node/node.go
```diff
@@ -0,0 +1,289 @@
...
+			return nil, nil
+		}
+		/* Template for dealing with optional dependencies:
+		if params.GatewayDependencies == nil {
+			gateway.New(...
+		} else {
+			gateway.NewDeps(...
+		}
+		*/
+		return gateway.New("localhost:0", false, filepath.Join(dir, modules.GatewayDir))
```
lukechampine: we will need a way to specify this, as well as the host address and the `bootstrap` bool. Maybe just add them to `NodeParams`.

DavidVorick: Ah yeah. I think a good way to do this would just have a bunch of fields for each module. We can add them as we need them.

PR #2608, opened 2018-01-04
Add vlong test for wallet's transaction output

ChrisSchinnerl:
This PR adds a test that creates a node with 10,000 txns in its wallet and runs subtests on this node to verify the integrity of the transaction and transactions endpoint.

### DIFF
--- a/modules/wallet.go
+++ b/modules/wallet.go
@@ -409,7 +409,7 @@ type (
 
 	// WalletSettings control the behavior of the Wallet.
 	WalletSettings struct {
-		Defrag bool `json:"defrag"`
+		NoDefrag bool `json:"noDefrag"`
 	}
 )
 
--- a/modules/wallet.go
+++ b/modules/wallet.go
@@ -377,6 +377,12 @@ type (
 		// blockchain.
 		Rescanning() bool
 
+		// Settings returns the Wallet's current settings.
+		Settings() WalletSettings
+
+		// SetSettings sets the Wallet's settings.
+		SetSettings(WalletSettings)
+
 		// StartTransaction is a convenience method that calls
 		// RegisterTransaction(types.Transaction{}, nil)
 		StartTransaction() TransactionBuilder
@@ -399,6 +405,11 @@ type (
 		// DustThreshold returns the quantity per byte below which a Currency is
 		// considered to be Dust.
 		DustThreshold() types.Currency
+	}
+
+	// WalletSettings control the behavior of the Wallet.
+	WalletSettings struct {
+		Defrag bool `json:"defrag"`
 	}
 )
 
--- a/modules/wallet/defrag.go
+++ b/modules/wallet/defrag.go
@@ -128,6 +128,11 @@ func (w *Wallet) managedCreateDefragTransaction() ([]types.Transaction, error) {
 // operation is only performed if the wallet has greater than defragThreshold
 // outputs.
 func (w *Wallet) threadedDefragWallet() {
+	// Don't defrag if it was disabled
+	if w.defragDisabled {
+		return
+	}
+
 	err := w.tg.Add()
 	if err != nil {
 		return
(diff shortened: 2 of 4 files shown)


modules/wallet.go
```diff
@@ -400,6 +406,11 @@ type (
 		// considered to be Dust.
 		DustThreshold() types.Currency
 	}
+
+	// WalletSettings control the behavior of the Wallet.
+	WalletSettings struct {
+		Defrag bool `json:"defrag"`
```
lukechampine: let's make this `NoDefrag` to match the wallet's internal boolean

api/wallet_test.go
```diff
@@ -1474,3 +1474,158 @@ func TestWalletGETDust(t *testing.T) {
...
+
+// testWalletTransactionEndpoint is a subtest that queries the transaction endpoint of a node.
+func testWalletTransactionEndpoint(t *testing.T, st *serverTester) {
+	// Mining blocks should have created transactions for the wallet containing
+	// miner payouts. Get the list of transactions.
+	var wtg WalletTransactionsGET
+	err := st.getAPI("/wallet/transactions?startheight=0&endheight=-1", &wtg)
+	if err != nil {
+		t.Fatal(err)
+	}
```
DavidVorick: Should have a check here that the number of transactions matches the number of expected transactions.

DavidVorick: That may mean that you have to pass the number of expected transactions in as input.

api/wallet_test.go
```diff
@@ -1474,3 +1474,158 @@ func TestWalletGETDust(t *testing.T) {
...
+}
+
+// testWalletTransactionEndpoint is a subtest that queries the transactions endpoint of a node.
+func testWalletTransactionsEndpoint(t *testing.T, st *serverTester) {
+	var wtg WalletTransactionsGET
+	err := st.getAPI("/wallet/transactions?startheight=0&endheight=-1", &wtg)
+	if err != nil {
+		t.Fatal(err)
+	}
+	totalTxns := len(wtg.ConfirmedTransactions)
```
DavidVorick: Same here, should have an input passed in that tells you what `totalTxns` is supposed to be, to make sure that you got them all.

api/wallet_test.go
```diff
@@ -1474,3 +1474,158 @@ func TestWalletGETDust(t *testing.T) {
...
+	if len(wtg.UnconfirmedTransactions) != 0 {
+		t.Fatal("expecting 0 unconfirmed transactions")
+	}
+
+	// Execute tests
+	for _, subtest := range subtests {
+		t.Run(subtest.name, func(t *testing.T) {
+			subtest.f(t, st)
+		})
+	}
```
DavidVorick: I would like to see this test extended so that a giant, blanking reorg is introduced, and then the subtests re-run.

And then another pass where the original chain is mined enough to return to the original chain. Then run the subtests a second time. This will give us confidence that the wallet is correctly handling basic reorgs.

You can set that up by creating a separate node, a blank tester, and mining blocks on it until it has more blocks than this node. When you call `miner.AddBlock()`, the block is returned, which means you can keep the list of blocks without connecting the nodes together. Then you can add all the blocks to this first node via `cs.AcceptBlock()`.

Getting the reorg back to the original chain is a little trickier. You can use the test miner to mine a block without submitting it to the consensus set. You will have to do that, and then you will have to mine another block on top of the first block. Then you submit both blocks to the now-blanked node, and it should reorg back to the original chain, which means all of the transactions should return.

DavidVorick:
No need to add all of the extra testing before this gets merged, just wanted to make sure we are testing the reorg handling code correctly in addition to the other stuff we are writing.

Also would like to see (in another PR) a subtest added that creates a new consensus set subscriber, subscribes to get the full block history, and then uses the wallet's seed + addresses to manually scan for transactions in the block history that match the wallet's addresses. Then, it queries the wallet's list of transactions and makes sure that the lists are identical.

This check will complement the other subtests by making sure the core code for the wallet is actually picking up every transaction.

PR #2611, opened 2018-01-04
Make wallet rebroadcast transactions until they are confirmed

ChrisSchinnerl:
This is a work in progress but I would appreciate some code review before I continue this.


modules/wallet/consts.go
```diff
@@ -15,6 +15,15 @@ const (
...
 	defragThreshold = 50
+
+	// rebroadcastInterval is the number of blocks the wallet will wait until
+	// it rebroadcasts an unconfirmed transaction by adding it to the
+	// transaction pool again.
+	rebroadcastInterval = 10
+
+	// rebroadcastMaxTries is the maximum number of times a transaction set
+	// will be rebroadcasted before the wallet stops tracking it
+	rebroadcastMaxTries = 10
```
DavidVorick: There's a const called `RespendTimeout` in `wallet.go` (that should be moved to `consts.go`) which says the wallet will try to spend an output again after 40 blocks of it not being spent. Instead of having `rebroadcastMaxTries`, you should just stop trying once the respend timeout has expired.

Now that we are trying to rebroadcast transactions, we can probably jump the `RespendTimeout` to 72 blocks, which is 12 hours.

DavidVorick:
Part of the purpose of this PR is to make sure transactions get re-broadcast. Right now, I'm pretty sure that if you call `tpool.AcceptTransactionSet` with a transaction set that you've already added to the tpool, you will get an error, and it won't broadcast.

So we should modify the `AcceptTransactionSet` to go ahead and broadcast the transaction and reset the timer on the transactions (they get booted after 12 blocks) if the broadcast function is called again.

I guess it's a little complicated though, because the wallet is going to start respending the transactions at some point, and we need to make sure that by that point the previous transaction in the transaction pool has been properly booted. I guess we'll need to leverage some consts in the modules package.

modules/wallet/update.go
```diff
@@ -453,6 +456,87 @@ func (w *Wallet) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+// isRelevantTxn checks if a txn is relevant to the Wallet
+func (w *Wallet) isRelevantTxn(txn types.Transaction) (relevant bool) {
+	// determine whether transaction is relevant to the wallet
+	for _, sci := range txn.SiacoinInputs {
+		relevant = relevant || w.isWalletAddress(sci.UnlockConditions.UnlockHash())
+	}
+	for _, sco := range txn.SiacoinOutputs {
+		relevant = relevant || w.isWalletAddress(sco.UnlockHash)
+	}
+	return
```
DavidVorick: Just throwing it out there to put it on the radar, it's not important for this PR, but the wallet is going to need to start recognizing file contracts as well as siacoin + siafund transactions.

modules/wallet/update.go
```diff
@@ -453,6 +456,87 @@ func (w *Wallet) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+			for _, txn := range block.Transactions {
+				if _, exists := bts.confirmedTxn[txn.ID()]; exists {
+					bts.confirmedTxn[txn.ID()] = true
+				}
+			}
+		}
+	}
+
+	// Check if all transactions of the set are confirmed
+	for tSetID, bts := range w.broadcastedTSets {
```
DavidVorick: Are we careful to make sure the size of `w.broadcastedTSets` is relatively limited? This loop would get pretty slow if you are storing too many transactions. Practically speaking, we probably only need to keep the transactions from the last 6 blocks or so, if we have a reorg deeper than 6 blocks you probably don't want to be automatically rebroadcasting those transactions anyway - something may have changed.

modules/wallet/wallet.go
```diff
@@ -43,6 +43,32 @@ type spendableKey struct {
...
 }
 
+// broadcastedTSet is a helper struct to keep track of transaction sets and to
+// help rebroadcast them.
+type broadcastedTSet struct {
+	height       types.BlockHeight
+	tries        int
+	confirmedTxn map[types.TransactionID]bool
+	transactions []types.Transaction
+}
```
DavidVorick: should probably move this code into `broadcast.go` or something. I think there's enough logic here that it deserves its own file.

DavidVorick:
I think generally speaking this code is on the right track. The one thing that seems to be missing is persistence - if you create a transaction, you should persist it in the database, so that if you start the wallet again later, it can be re-broadcast again. Have to be careful about that though, because after `RespendTimeout` the wallet will start using those outputs to make other transactions again, and we want to make sure we aren't broadcasting transactions which the wallet will try to spend again.

4Dolio:
Great issue! Just throwing it out there, wr could use a traditional 'progressive' retry? Begin to retry in a short block height and get slower.  Not sure is a very short retry would induce problems or cause a feedback overload to the blockchain/network itself? So pick a reasonably fast yet safe start duration.  Say 5, then 10, 20, 40, 80, 160, untill acceptable max like 1 month? Week?

modules/wallet/consts.go
```diff
@@ -15,9 +17,28 @@ const (
...
 )
 
 var (
+	// rebroadcastTimeout is the amount of blocks after which we stop trying to
+	// rebroadcast transactions. The reason why we can't just use
+	// respendTimeout as the rebroadcastTimeout is, that the transaction pool
+	// will boot transactions after MaxTxnAge. We need to make sure that we
+	// leave at least MaxTxnAge blocks after the last broadcast to allow for
+	// the transasction to be pruned before the wallet tries to respend it.
+	rebroadcastTimeout = types.BlockHeight(respendTimeout - modules.MaxTxnAge)
```
DavidVorick: excellent comments

DavidVorick:
Well, it's bad for the network to broadcast frequently. And, if you haven't broadcasted in 24 blocks, your transactions will have been dropped from any nodes running standard code, and will be unlikely to reach miners. So at the very least, you'd want to try once every 24 blocks I'd think.

But, once every 6 blocks seems pretty sound / okay as well, at least for now. If transaction pools are full or the fees are too high, it'll likely change after 6 blocks.

jarcko:
Tranascations not found here
https://explore.sia.tech/hashes/50a860e7e4ce6976710d1d52b580bb83ca5e125dbbfd5e34722811a365676a05

screen shot 2018-01-11 at 7 54 55 pm

I see it every time so can't send coins at all, I reinstalled all from scratch and recovered the wallet, but still the same. Can you reproduce the issue?

DavidVorick:
Are you running a host? Did you just make that transaction? How long did you wait for the coins to send?

modules/consts.go
```diff
@@ -0,0 +1,10 @@
+package modules
+
+import "github.com/NebulousLabs/Sia/types"
+
+// Consts that are required by multiple modules
+const (
+	// maxTxnAge determines the maximum age of a transaction (in block height)
+	// allowed before the transaction is pruned from the transaction pool.
+	MaxTxnAge = types.BlockHeight(24)
```
DavidVorick: I believe that there are other consts throughout the modules package, we may want to collect them all here

modules/transactionpool/accept.go
```diff
@@ -335,10 +335,24 @@ func (tp *TransactionPool) AcceptTransactionSet(ts []types.Transaction) error {
...
+			// it's seen txn height when we rebroadcast it. If we don't do
+			// that, the transaction will be pruned from the tpool and they
+			// might no longer show up in the wallet while still beingt
+			// tracked.
+			if err != nil {
+				for _, txn := range ts {
+					tp.transactionHeights[txn.ID()] = tp.blockHeight
+				}
+			}
+			go tp.gateway.Broadcast("RelayTransactionSet", ts, tp.gateway.Peers())
```
DavidVorick: calling `tp.gateway.Peers()` while the tpool is under lock is a deadlock risk, need to restructure to avoid this (looks like the existing code had this deadlock risk in it as well)

DavidVorick: The general convention is that a module can't call exported methods of other modules while under lock

modules/wallet/broadcast.go
```diff
@@ -0,0 +1,74 @@
...
+type persistBTS struct {
+	FirstTry     types.BlockHeight   // first time the tSet was broadcasted
+	LastTry      types.BlockHeight   // last time the tSet was broadcasted
+	ConfirmedTxn []bool              // tracks confirmed txns of set
+	Transactions []types.Transaction // the tSet
+}
+
+// confirmed is a helper function that sets a certain transactions to confirmed
+// or unconfirmed. It also updates the state on disk.
+func (bts *broadcastedTSet) confirmed(txid types.TransactionID, confirmed bool) error {
```
DavidVorick: this function name suggests a query, as though you are asking for the confirmed status of a transaction. Should change the name to 'markConfirmation' or something similar.

modules/wallet/update.go
```diff
@@ -453,6 +456,98 @@ func (w *Wallet) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+	}
+	for _, sco := range txn.SiacoinOutputs {
+		relevant = relevant || w.isWalletAddress(sco.UnlockHash)
+	}
+	return
+}
+
+// rebroadcastOldTransaction rebroadcasts transactions that haven't been
+// confirmed within rebroadcastInterval blocks
+func (w *Wallet) rebroadcastOldTransactions(tx *bolt.Tx, cc modules.ConsensusChange) error {
```
DavidVorick: This function should probably be moved to `broadcast.go`

modules/wallet/update.go
```diff
@@ -453,6 +456,98 @@ func (w *Wallet) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+	// Get the current consensus height
+	consensusHeight, err := dbGetConsensusHeight(tx)
+	if err != nil {
+		return err
+	}
+
+	// Mark reverted transactions as not confirmed
+	for _, block := range cc.RevertedBlocks {
+		for _, bts := range w.broadcastedTSets {
```
DavidVorick: This is an n^2 algorithm. The thing that worries me is that for each broadcastedTSet, you are iterating through every transaction of every block. If a wallet has worked up something like 100 broadcastedTSets over the past few blocks (not unreasonable for an exchange with lots of inbound attention), and there are 2000 transactions per block, we could be seeing 200,000 operations happening to get through each block. Combine that with a large (say 5 blocks) reorg and you could be seeing 4 reversions + 5 additions = 1.8 million total operations.

Need to find a faster way to do this.

modules/wallet/update.go
```diff
@@ -453,6 +456,98 @@ func (w *Wallet) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+			for _, txn := range block.Transactions {
+				if _, exists := bts.confirmedTxn[txn.ID()]; exists {
+					err = bts.confirmed(txn.ID(), false)
+				}
+				if err != nil {
```
DavidVorick: error handling needs to happen in the same scope as error creation

modules/transactionpool/accept.go
```diff
@@ -331,18 +331,33 @@ func (tp *TransactionPool) AcceptTransactionSet(ts []types.Transaction) error {
 		return errors.New("consensus set does not support LockedTryTransactionSet method")
 	}
 
-	return cs.LockedTryTransactionSet(func(txnFn func(txns []types.Transaction) (modules.ConsensusChange, error)) error {
+	err := cs.LockedTryTransactionSet(func(txnFn func(txns []types.Transaction) (modules.ConsensusChange, error)) error {
 		tp.mu.Lock()
 		defer tp.mu.Unlock()
 		err := tp.acceptTransactionSet(ts, txnFn)
 		if err != nil {
```
DavidVorick: `if err == nil`, right now you are only updating subscribers if there is an error :P

modules/transactionpool/accept.go
```diff
@@ -331,18 +331,33 @@ func (tp *TransactionPool) AcceptTransactionSet(ts []types.Transaction) error {
...
+		// it's seen txn height when we rebroadcast it. If we don't do
+		// that, the transaction will be pruned from the tpool and they
+		// might no longer show up in the wallet while still beingt
+		// tracked.
+		if err != nil {
+			tp.mu.Lock()
+			for _, txn := range ts {
+				tp.transactionHeights[txn.ID()] = tp.blockHeight
+			}
+			tp.mu.Unlock()
```
DavidVorick: ah I missed this. This should happen under the same lock that you update the subscribers with.

modules/consts.go
ChrisSchinnerl: Should that be done in the same PR?

modules/wallet/update.go
```diff
@@ -453,6 +456,98 @@ func (w *Wallet) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+			go func(tSet []types.Transaction) {
+				if err := w.tpool.AcceptTransactionSet(tSet); err != nil {
+					w.log.Println("WARNING: Rebroadcast failed: ", err)
+				}
+			}(bts.transactions)
+			// Delete the transaction set once we have tried for RespendTimeout
+			// blocks
+			if consensusHeight >= bts.firstTry+RebroadcastTimeout {
+				if err := w.deleteBroadcastedTSet(tSetID); err != nil {
+					return err
```
DavidVorick: probably better to log the error than to return in this case.

PR #2614, opened 2018-01-05
add /tpool/confirmed/:id endpoint

lukechampine:
The tpool already had a `transactionConfirmed` method, so this is pretty simple stuff


DavidVorick:
Need to fill out the api documentation and write an integration test showing that the endpoint is functional.

PR #2623, opened 2018-01-07
Add api/server package - create a full Sia node with 1 function call

DavidVorick:


node/api/server/server.go
```diff
@@ -0,0 +1,88 @@
...
+		listener:          listener,
+		requiredUserAgent: requiredUserAgent,
+	}
+
+	// Spin up a channel that will block until the server has closed, and then
+	// send any error down the channel.
+	go func() {
+		err := srv.serve()
+		srv.errChan <- err
+		close(srv.errChan)
```
lukechampine: why not have `serve()` send down `srv.errChan` internally? Then the call is just `go srv.serve()`. Also, I'm not sure if closing the channel is necessary or desirable here.

DavidVorick: If you don't close the channel and a thread decides to listen on that channel, the thread will block forever.

Can send the error down the channel inside of `serve` though at least.

lukechampine: The channel only exists to send a single value asynchronously -- the `serve` error. If you read from the channel multiple times, you're doing it wrong. In particular, if you close the channel, then all reads (after the first) will return `nil`, even if `serve` returned a non-nil error.

If you want the error to be available to multiple callers, you could use a pattern like `rpc.Call`: https://golang.org/pkg/net/rpc/#Call
You listen on a `Done` channel, and once the read succeeds, you know that the `Error` field is safe to read. We could use an equivalent pattern like:

```go
type Server struct {
    // ...
    serveErr error
    done     chan struct{}
}

func (srv *Server) Close() error {
    <-srv.done
    err := srv.listener.Close()
    err = errors.Compose(err, srv.serveErr)
    // ...
}

func New() {
    // ...
    go func() {
        srv.serveErr = srv.serve()
        close(srv.done)
    }()
}
```

DavidVorick: That might be better.

PR #2624, opened 2018-01-07
add consensus change sanity check to tpool

DavidVorick:


DavidVorick:
For those watching the github, this issue relates to some verification about the wallet behavior. We've seen some errors across two modules which potentially suggest that the wallet is missing transactions due to faulty consensus change consistency. While all of these errors have other probable causes, it would be really bad if we were overlooking faulty consensus change consistency.

This code, while not in the wallet, cleanly verifies from an external module that the consensus changes are being applied correctly. If there is faulty consistency, we'll know after running this code in production for a bit.

I don't expect there to be errors related to this code, but we want to narrow down the set of things that are potentially going wrong.

PR #2640, opened 2018-01-10
Multi-block acceptance inconsistency fix

DavidVorick:
This is a code fix where inconsistency was being introduced into the update stream being broadcast to modules from the consensus package, often resulting in problems with modules such as the wallet, explorer, and miner.

This PR should fix that issue, though we will probably need to add some extra code to each of the modules before the next release so we can figure out if they are missing anything important.

### DIFF
--- a/modules/consensus/accept.go
+++ b/modules/consensus/accept.go
@@ -284,7 +284,7 @@ func (cs *ConsensusSet) managedAcceptBlocks(blocks []types.Block) (blockchainExt
 		os.Exit(1)
 	}
 	if setErr != nil {
-		if len(changes) < 1 {
+		if len(changes) == 0 {
 			fmt.Println("Received an invalid block set.")
 			cs.log.Println("Consensus received an invalid block:", setErr)
 		} else {
--- a/modules/consensus/accept.go
+++ b/modules/consensus/accept.go
@@ -241,8 +241,6 @@ func (cs *ConsensusSet) managedAcceptBlocks(blocks []types.Block) (blockchainExt
 	// invalid blocks (which includes the children of invalid blocks).
 	chainExtended := false
 	changes := make([]changeEntry, 0, len(blocks))
-	validBlocks := make([]types.Block, 0, len(blocks))
-	parents := make([]*processedBlock, 0, len(blocks))
 	setErr := cs.db.Update(func(tx *bolt.Tx) error {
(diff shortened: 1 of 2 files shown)


modules/consensus/accept.go
```diff
@@ -284,67 +285,26 @@ func (cs *ConsensusSet) managedAcceptBlocks(blocks []types.Block) (blockchainExt
...
-		if build.DEBUG && verifyExtended != chainExtended {
-			panic("chain extension logic does not match up between first and last attempt")
-		}
-		// Something has gone wrong. Maybe the filesystem is having errors for
-		// example. But under normal conditions, this code should not be
-		// reached. If it is, return early because both attempts to add blocks
-		// have failed.
-		if err != nil {
-			return false, err
+		if len(changes) < 1 {
```
lukechampine: `== 0` is clearer

modules/consensus/accept.go
```diff
@@ -284,67 +285,26 @@ func (cs *ConsensusSet) managedAcceptBlocks(blocks []types.Block) (blockchainExt
...
-	}
-
-	// If there were valid blocks and invalid blocks in the set that was
-	// provided, then the setErr is not going to be nil. Return the set error to
-	// the caller.
-	if setErr != nil {
-		return chainExtended, setErr
+	// Send any changes to subscribers.
+	for i := 0; i < len(changes); i++ {
+		cs.updateSubscribers(changes[i])
```
lukechampine: I don't think the `for _, change := range changes` form could have caused problems (since everything is being passed by value), but I don't blame you for being suspicious :P

modules/consensus/accept.go
```diff
@@ -284,68 +285,27 @@ func (cs *ConsensusSet) managedAcceptBlocks(blocks []types.Block) (blockchainExt
...
-		// example. But under normal conditions, this code should not be
-		// reached. If it is, return early because both attempts to add blocks
-		// have failed.
-		if err != nil {
-			return false, err
+		if len(changes) < 1 {
+			fmt.Println("Received an invalid block set.")
+			cs.log.Println("Consensus received an invalid block:", setErr)
+		} else {
+			fmt.Println("Received a partially valid block set.")
```
lukechampine: is this for debugging purposes? We don't want this in production, right?

PR #2641, opened 2018-01-10
[WIP] Improve unconfirmed txn handling in wallet

ChrisSchinnerl:
The wallet should be able to track transactions even if they haven't been accepted into the transaction set. For this to work, the wallet has to be able to add transactions to its set of unconfirmed transactions without relying on the transaction pool. This PR introduces a wrapper for `AcceptTransactionSet` to achieve this.

### DIFF
--- a/modules/wallet/defrag.go
+++ b/modules/wallet/defrag.go
@@ -175,7 +175,7 @@ func (w *Wallet) threadedDefragWallet() {
 		return
 	}
 	// Submit the defrag to the transaction pool.
-	err = w.tpool.AcceptTransactionSet(txnSet)
+	err = w.managedCommitTransactionSet(txnSet)
 	if err != nil {
 		w.log.Println("WARN: defrag transaction was rejected:", err)
 		return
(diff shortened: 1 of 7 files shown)


build/errors.go
```diff
@@ -24,6 +24,16 @@ func ComposeErrors(errs ...error) error {
 		return nil
 	}
 
+	// If there is only a single error return it directly. This still allows
+	// callers of functions to check for specific errors
+	if len(errStrings) == 1 {
```
lukechampine: this will have to be added to NebulousLabs/errors as well. We should make a PR that strips out build/errors.go.

ChrisSchinnerl: Maybe we should create an issue for that to make sure we don't forget about it.

ChrisSchinnerl:
The changes I made so far should allow the wallet to ignore certain errors returned by `tpool.AcceptTransactionSet` while still being able to keep track of the set in `wallet.unconfirmedSets`. The todo in `wallet.commitTransactionSet` still has to be replaced by the errors that should be ignored.

PR #2654, opened 2018-01-14
attempt to clean up smux concurrency

DavidVorick:


vendor/github.com/xtaci/smux/mux.go
```diff
@@ -24,15 +24,20 @@ type Config struct {
 	// MaxReceiveBuffer is used to control the maximum
 	// number of data in the buffer pool
 	MaxReceiveBuffer int
+
+	// WriteTimeout defines the default amount of time to wait before giving up
+	// on a write.
+	WriteTimeout time.Duration
```
DavidVorick: Added a WriteTimeout, so that all session operations will have some sane default timeout

vendor/github.com/xtaci/smux/session.go
```diff
@@ -3,36 +3,31 @@ package smux
 import (
 	"encoding/binary"
 	"io"
+	"net"
 	"sync"
 	"sync/atomic"
 	"time"
 
 	"github.com/pkg/errors"
+	siasync "github.com/NebulousLabs/Sia/sync" // TODO: Replace with github.com/NebulousLabs/trymutex
```
DavidVorick: We should move `TryMutex` to it's own top-level package, I think it's both useful and stable.

vendor/github.com/xtaci/smux/session.go
```diff
@@ -3,36 +3,31 @@ package smux
...
-
-type writeResult struct {
-	n   int
-	err error
-}
-
 // Session defines a multiplexed connection for streams
 type Session struct {
-	conn io.ReadWriteCloser
+	conn net.Conn
```
DavidVorick: for some reason the `conn` was reduced to an `io.ReadWriteCloser`, which meant I couldn't set deadlines.

Fixed now.

vendor/github.com/xtaci/smux/session.go
```diff
@@ -281,73 +276,76 @@ func (s *Session) recvLoop() {
...
 }
 
-func (s *Session) keepalive() {
-	tickerPing := time.NewTicker(s.config.KeepAliveInterval)
-	tickerTimeout := time.NewTicker(s.config.KeepAliveTimeout)
-	defer tickerPing.Stop()
-	defer tickerTimeout.Stop()
+// keepAliveSend will periodically send a keepalive meesage to the remote peer.
+func (s *Session) keepAliveSend() {
+	keepAliveTimeout := time.After(s.config.KeepAliveInterval)
```
DavidVorick: The ticker method means that keepalive messages can get missed. If it takes more than 1 tick to perform a keepalive send, then the timing could mean that you miss the next tick as well. Switching to an absolute timeout gives better control over when we do things.

vendor/github.com/xtaci/smux/session.go
```diff
@@ -281,73 +276,76 @@ func (s *Session) recvLoop() {
...
-			n, err := s.conn.Write(buf[:headerSize+len(request.frame.data)])
-
-			n -= headerSize
-			if n < 0 {
-				n = 0
-			}
-
-			result := writeResult{
-				n:   n,
-				err: err,
```
DavidVorick: This red diff is from the function `sendLoop` that I deleted, it's not from `keepAliveTimeout`

vendor/github.com/xtaci/smux/session.go
```diff
@@ -281,73 +276,76 @@ func (s *Session) recvLoop() {
...
 	}
 }
 
 // writeFrame writes the frame to the underlying connection
 // and returns the number of bytes written if successful
-func (s *Session) writeFrame(f Frame) (n int, err error) {
-	req := writeRequest{
-		frame:  f,
-		result: make(chan writeResult, 1),
+func (s *Session) writeFrame(frame Frame, timeout time.Time) (int, error) {
```
DavidVorick: writing a frame used to streamline through a thread that was listening on a channel for frames to send. Working in that setup required a lot of extra structs and channels, and made it easy to make mistakes. So easy, in fact, that there were deadlocks in the original code, and timeouts were not handled correctly at all.

Now, timeouts are handled correctly, things are still streamlined with the use of mutexes, and we've done away with all the extra code required to make this work with channels.

This is definitely a case where a mutex makes more sense. We aren't trying to pass information between actors, we are trying to make sure that only one actor is performing an action at a time.

vendor/github.com/xtaci/smux/stream.go
```diff
@@ -70,51 +68,27 @@ READ:
...
-				return sent, result.err
-			}
-		case <-s.die:
-			return sent, errors.New(errBrokenPipe)
-		case <-deadline:
-			return sent, errTimeout
+		n, err := s.sess.writeFrame(frames[k], deadline)
+		sent += n
+		if err != nil {
+			return sent, err
```
DavidVorick: just look at how much cleaner and more sane this code is now :P

vendor/github.com/xtaci/smux/session.go
lukechampine: probably to make it easier to mock

vendor/github.com/xtaci/smux/session.go
```diff
@@ -281,73 +277,88 @@ func (s *Session) recvLoop() {
...
-			s.notifyBucket() // force a signal to the recvLoop
-		case <-tickerTimeout.C:
-			if !atomic.CompareAndSwapInt32(&s.dataReady, 1, 0) {
-				s.Close()
-				return
-			}
 		case <-s.die:
 			return
+		case <-keepAliveTimeout:
+			keepAliveTimeout = time.After(s.config.KeepAliveInterval) // set before writing so we start sending the next one in time
```
lukechampine: if the timeout channel is reset immediately upon reading from the channel... what's the difference between this and a ticker?

I looked into the `time.Ticker` source code, and it appears that its channel has a buffer of 1. So if you're busy doing some other work and you "miss" a tick, the channel will still have a value in it when you come back, and thus the next read will succeed immediately. If you miss multiple ticks, `time.Ticker` skips the send (leaving the channel with one value in it). I think this prevents the "missed tick" behavior that your comment describes.

So if we did:
```go
sendTicker := time.NewTicker(s.config.KeepAliveInterval)
defer sendTicker.Stop()
for {
	select {
	case <-s.die:
		return
	case <-sendTicker.C:
		writeFrame()
	}
}
```
I think it's functionally equivalent. We can check by examining three scenarios:
1. `writeFrame` always takes less time than `s.config.KeepAliveInterval`
2. `writeFrame` always takes more time than `s.config.KeepAliveInterval`
3. `writeFrame` varies between taking more time and less time than `s.config.KeepAliveInterval`.

We'll use a `KeepAliveInterval` of 1 second for each case. In the first case, `writeFrame` will take 0.6 seconds. In the second, it will take 1.4 seconds. In the last case, it will alternate between 0.7 seconds and 1.5 seconds.

Case 1:

In case 1, both the PR code and the Ticker code call `writeFrame` at `t=1`, `t=2`, `t=3`, etc. `writeFrame` never affects the timing.

Case 2:

In case 2, both the PR code and the Ticker code call `writeFrame` at `t=1`, `t=2.4`, `t=3.8`, etc. Since the channel is filled during the `writeFrame` call, the next channel read always succeeds immediately.

Case 3:

In case 3, the PR code calls `writeFrame` at `t=1`, `t=2`, `t=3.5`, `t=4.5`, `t=6`, `t=7`... whereas the Ticker code calls `writeFrame` at `t=1`, `t=2`, `t=3.5`, `t=4.2`, `t=5.7`, `t=6.4`...

So these implementations are in fact different! Here's why. In the PR code, when `writeFrame` is called at `t=3.5`, it takes 0.7 seconds, finishing at `t=4.2`. Since `time.After` was called immediately prior to `writeFrame`, it sends down the channel at `t=4.5`, which kicks off the next `writeFrame`.

In the Ticker code, `writeFrame` is again called at `t=3.5` and finishes at `t=4.2`. The Ticker is dutifully ticking on each second, so during the `writeFrame` call, it sends down the channel at `t=4`. When `writeFrame` finishes at `t=4.2`, the channel is full, so the next `writeFrame` is called immediately.

[Playground link](https://play.golang.org/p/5pVAumYGzRe)

I'm not sure which version is preferable. Both seem to call `writeFrame` at reasonable intervals. Ideally we could devise a test cases that causes unacceptable behavior in one of them, but not the other.

vendor/github.com/xtaci/smux/session.go
lukechampine: streamline -> serialize :P 

I do agree that a channel is not optimal here. It would make sense if the frames needed to be ordered, but they don't. It was probably chosen to enable both serialization and timeouts. (`TryMutex` uses a channel internally for the same reason).
However, I'm also a little uneasy overhauling the concurrency of a package that we don't fully understand. We obviously need to test this code thoroughly on the real network before including it in a release. But I'm also curious whether we can fix the bug we're observing while making minimal changes to smux.

lukechampine: There's some discussion of deadlines here: https://github.com/xtaci/smux/issues/3

I suggest that we try calling `SetWriteDeadline` throughout the smux code and see if it breaks anything.

vendor/github.com/xtaci/smux/session.go
```diff
@@ -281,73 +281,91 @@ func (s *Session) recvLoop() {
...
-func (s *Session) keepalive() {
-	tickerPing := time.NewTicker(s.config.KeepAliveInterval)
-	tickerTimeout := time.NewTicker(s.config.KeepAliveTimeout)
-	defer tickerPing.Stop()
-	defer tickerTimeout.Stop()
+// keepAliveSend will periodically send a keepalive meesage to the remote peer.
+func (s *Session) keepAliveSend() {
+	// Use 'time.Ticker' to ensure that if the keepalive sending falls behind
+	// schedule, unnecessary sleeping does not occur.
+	ticker := time.NewTicker(s.config.KeepAliveTimeout)
```
lukechampine: shouldn't this be `KeepAliveInterval`?

lukechampine: also, need to `defer ticker.Stop`

vendor/github.com/xtaci/smux/session.go
```diff
@@ -281,73 +281,91 @@ func (s *Session) recvLoop() {
...
 }
 
-func (s *Session) sendLoop() {
-	buf := make([]byte, (1<<16)+headerSize)
+// keepAliveTimeout will periodically check that some sort of message has been
+// sent by the remote peer, closing the session if not.
+func (s *Session) keepAliveTimeout() {
+	// Use 'time.After' to ensure that at least 'n' amount of time passes
+	// between each check that a keepalive has been sent.
+	timeoutChan := time.After(s.config.KeepAliveTimeout)
```
lukechampine: the loop in this function isn't doing any blocking work (just a single atomic op) so a ticker should function identically here

vendor/github.com/xtaci/smux/session.go
```diff
@@ -281,73 +281,91 @@ func (s *Session) recvLoop() {
 	}
 }
 
-func (s *Session) keepalive() {
-	tickerPing := time.NewTicker(s.config.KeepAliveInterval)
-	tickerTimeout := time.NewTicker(s.config.KeepAliveTimeout)
-	defer tickerPing.Stop()
-	defer tickerTimeout.Stop()
+// keepAliveSend will periodically send a keepalive meesage to the remote peer.
```
ChrisSchinnerl: meesage

PR #2664, opened 2018-01-18
User-manual ‘disconnect’ adds a peer to a blacklist

ChrisSchinnerl:
This PR will cause a user initiated disconnect to blacklist a peer. The peer will only be blacklisted if it was actually connected to the node. For blacklisting nodes that are not connected we should probably add a separate endpoint (if we decide that we want that feature). 
The blacklist is persisted in the `blacklist.json` file.

### DIFF
--- a/modules/gateway/gateway_test.go
+++ b/modules/gateway/gateway_test.go
@@ -1,8 +1,6 @@
 package gateway
 
 import (
-	"errors"
-	"fmt"
 	"io/ioutil"
 	"net"
 	"os"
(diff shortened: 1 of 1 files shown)


modules/gateway/gateway.go
```diff
@@ -245,6 +249,10 @@ func New(addr string, bootstrap bool, persistDir string) (*Gateway, error) {
 	if loadErr := g.load(); loadErr != nil && !os.IsNotExist(loadErr) {
 		return nil, loadErr
 	}
+	// Do the same for the blacklist.
+	if loadErr := g.loadBlacklist(); loadErr != nil && !os.IsNotExist(loadErr) {
+		return nil, loadErr
+	}
```
lukechampine: all persist loading should happen in `load`

modules/gateway.go
```diff
@@ -162,9 +162,15 @@ type (
...
 		Connect(NetAddress) error
 
+		// ConnectManual is a Connect wrapper for a user-initiated Connect
+		ConnectManual(NetAddress) error
+
 		// Disconnect terminates a connection to a peer.
 		Disconnect(NetAddress) error
 
+		// DisconnectManual is a Disconnect wrapper for a user-initiated disconnect
+		DisconnectManual(NetAddress) error
```
lukechampine: why do we need these? Wouldn't a `Blacklist` method make more sense? It's not clear to me that a user-initiated disconnect always implies that the user wants to blacklist the peer.

As always, we may want to check how bitcoind solves these problems.

ChrisSchinnerl: I think bitcoind has a [`setban`](https://bitcoin.stackexchange.com/questions/43236/manually-banning-nodes-through-cli) function that allows a user to ban a node for a certain period of time.
If we don't like this behaviour we can also change this PR to introduce a new endpoint. I started it because it was one of the todos.

ChrisSchinnerl: @DavidVorick any thoughts on that?

PR #2685, opened 2018-01-24
[WIP] Download code overhaul

ChrisSchinnerl:
This PR is supposed to be a complete overhaul of the Renter's download code. The download code should be closer to the recent changes made to the upload code. That means that it shares a memory pool with the upload code and workers decide for themselves if they should help with the download of a chunk. User-initiated downloads also have a higher priority now.
This PR is a WIP and is still missing many features and changes. My first goal is to get it to pass all the tests we have so far. Unfortunately it still gets stuck while doing downloads for repairs.
The next step is to change it to use the same memory restrictions as the upload code. The last thing I'd like to add is more testing. We need more tests to verify correct behaviour in case of unreliable hosts.


ChrisSchinnerl:
Closing this since it is continued in [#2758](https://github.com/NebulousLabs/Sia/pull/2758).

PR #2688, opened 2018-01-25
create client package

lukechampine:
This is mostly copied from `node/api/client.go`, though I do clean up the code a little.

Next steps:
- Replace usage of `api.Client` with `client.Client`. This currently causes an import cycle: the tests in `api` should use a `client.Client`, but `client` imports `api`. We could use a separate `api_test` package here, or we could move the tests to another package.
- Expand `client` to provide more helper methods for the various endpoints.


DavidVorick:
`undefined: api.NewClient`

PR #2694, opened 2018-01-27
fixes #2693 32bit incompatibility

### LINKED ISSUE #2693: Overflow of time.Duration subtraction in modules/renter/hostdb/scan.go
# BUG REPORT

Complete crash:
```
(6/6) Loading renter...
panic: fastrand: argument to Intn is <= 0: -277250048
```

The problematic code is exactly: [modules/renter/hostdb/scan.go:354](https://github.com/NebulousLabs/Sia/blob/194d3995c8838f1a92070d3973ece006e5c7ef3d/modules/renter/hostdb/scan.go#L354)
```go
// Sleep for a random amount of time before doing another round of
// scanning. The minimums and maximums keep the scan time reasonable,
// while the randomness prevents the scanning from always happening at
// the same time of day or week.
sleepTime := defaultScanSleep
sleepRange := int(maxScanSleep - minScanSleep)
sleepTime = minScanSleep + time.Duration(fastrand.Intn(sleepRange))
```
So, what happens is that the `time.Duration` type subtraction gets you nanoseconds (2.76×10¹³ for standard release constants 08:00:00 and 01:20:00) and typecasting this to integer on my system causes overflow. I do concede that 32bit isn't officially supported (cold wallet usage here), but 32bit `int` might crop up elsewhere too. Overflow in 64bit will happen at 584y.

Also, it's easy to fix 32bit support. For me it seems to run fine after fixing this. I propose using the type's `.Seconds()` to get `minScanSleep`/`maxScanSleep` in more manageable units, overflowing 32bit at 136y.

**Expected Behavior**
No overflow and panic of `fastrand` on 32bit `int` systems

**How to reproduce it (as minimally and precisely as possible)**
Compile with `GOARCH=386`

**Environment**
* Sia version: 1.3.1 (Git Jan 22 2018 [194d3995](https://github.com/NebulousLabs/Sia/commit/194d3995c8838f1a92070d3973ece006e5c7ef3d))
* OS: Linux i686 (Xubuntu 17.10)

**Additional information**
Disclosure: I just discovered that Sia can't handle mapping a 9GB blockchain on a 32bit system (page indexing out of bounds), so this fix alone would likely not make 32bit siad possible.

tbenz9:


modules/renter/hostdb/scan.go
```diff
@@ -351,8 +351,8 @@ func (hdb *HostDB) threadedScan() {
 		// while the randomness prevents the scanning from always happening at
 		// the same time of day or week.
 		sleepTime := defaultScanSleep
-		sleepRange := int(maxScanSleep - minScanSleep)
-		sleepTime = minScanSleep + time.Duration(fastrand.Intn(sleepRange))
+		sleepRange := uint64(maxScanSleep - minScanSleep)
+		sleepTime = minScanSleep + time.Duration(fastrand.Uintn(sleepRange))
```
lukechampine: should be `Uint64n` :)

DavidVorick:
does this result in an explicit bug?

lukechampine:
yes, see https://github.com/NebulousLabs/Sia/issues/2693
although you can't run siad on 32-bit anyway

PR #2703, opened 2018-01-30
First siatesting test

ChrisSchinnerl:
This PR introduces a simple testcase and some changes to siatest.


node/api/server/server.go
```diff
@@ -88,3 +104,14 @@ func New(APIaddr string, requiredUserAgent string, requiredPassword string, node
 
 	return srv, nil
 }
+
+// MineBlock makes the underlying node mine a single block and broadcast it.
+func (srv *Server) MineBlock() error {
```
DavidVorick: This needs to be a function of the test node in the `siatest` package, the `Server` should be treated as a simplified object that we intend to use in production

node/api/server/server.go
```diff
@@ -76,9 +83,18 @@ func New(APIaddr string, requiredUserAgent string, requiredPassword string, node
 		},
 		done:              make(chan struct{}),
 		listener:          listener,
+		node:              node,
 		requiredUserAgent: requiredUserAgent,
 	}
 
+	// fund the node
+	for i := types.BlockHeight(0); i <= types.MaturityDelay; i++ {
```
DavidVorick: same here, funding a node before returning it is meant to be something you do when creating a `siatest` server/client pair, not something you do when you create a production server

node/node.go
```diff
@@ -205,7 +206,16 @@ func New(params NodeParams) (*Node, error) {
 		if !params.CreateWallet {
 			return nil, nil
 		}
-		return wallet.New(cs, tp, filepath.Join(dir, modules.WalletDir))
+		w, err := wallet.New(cs, tp, filepath.Join(dir, modules.WalletDir))
```
DavidVorick: And same with this stuff, creation of a key and initialization of the wallet should happen in the `siatest` package when you create a server/client pair, not in the node package.

siatest/consensus/consensus_test.go
```diff
@@ -0,0 +1,57 @@
...
+	if testing.Short() {
+		t.SkipNow()
+	}
+	testdir, err := siatest.TestDir(siatest.SiaTestingDir, t.Name())
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Create a new server
+	c, s, err := siatest.NewClientServerPair(node.AllModules(testdir))
```
DavidVorick: Instead of returning a `c` and an `s` separately, it should return a `TestNode` object which has a `c` and an `s` as public fields.

DavidVorick: oh hmm, or maybe they should just be embedded fields, that might be even easier.

node/api/client/get.go
```diff
@@ -0,0 +1,23 @@
+package client
+
+import (
+	"github.com/NebulousLabs/Sia/encoding"
+	"github.com/NebulousLabs/Sia/node/api"
+	"github.com/NebulousLabs/Sia/types"
+)
+
+// GetConsensus requests the /consensus api resource
+func (c *Client) GetConsensus() (cg api.ConsensusGET, err error) {
```
DavidVorick: Need to swap the verb here. `ConsensusGet`. Same for everything else. We do this so when it gets sorted alphabetically, you don't have to search through a bunch of `Get` prefixes to find what you want.

siatest/consensus/consensus_test.go
```diff
@@ -0,0 +1,57 @@
...
+	if cg.Height != height+1 {
+		t.Fatal("Height should have increased by 1 block")
+	}
+
+	// Close the server and check error
+	defer func() {
+		if err := testNode.Close(); err != nil {
+			t.Fatal(err)
+		}
+	}()
```
DavidVorick: The code here could be compacted a little more. For example, pulling the `height` out of the `cg` should probably happen in the same codeblock, since it follows the same thought of extracting the height. And then when you check the height, getting the height a second time and performing the check can probably also be the same codeblock.

The `defer close` block should happen right after the call to `NewNode`, because we want to make sure we free resources at the same exact time that we allocate them (to minimize mistakes).

Issue #2686, opened 2018-01-25
Feature Request: Send Transaction with Arbitraty Data Api Endpoint


geo-gs:
As titled; I'd like to be able to leverage the API to send transactions with arbitrary data in the following way:

```
// Number of hastings being sent. A hasting is the smallest unit in Sia. There
// are 10^24 hastings in a siacoin.
amount      // hastings

// Address that is receiving the coins.
destination // address

// data should be a Base64 representation of a byte[]. First 6 bytes must decode to 'NonSia'
// note: is optional
arbitrarydata // data

// JSON array of outputs. The structure of each output is:
// {"unlockhash": "<destination>", "value": "<amount>", "arbitrarydata":"<data>"}
outputs
```

lukechampine:
hmm. Can I ask what the intended use is? If you want to use arbitrarydata purely to get certain data on the blockchain, the `/wallet/siacoins` endpoint isn't a great fit, because you don't really need to be sending coins to anyone in order to get arbitrarydata into a transaction; you just need to pay the transaction fee.

geo-gs:
@lukechampine 
I'm writing a program called [sawwit](https://github.com/geo-gs/sawwit) which is basically a way for people to communicate with each other on top of a block-chain.

The use case is that I want people to use small amounts of sia coins when interacting with the program to:
* Bring in revenue
* Prevent spam and self-voting
* Allow users to profit-share

Specifically, I could really use this endpoint in my code here: https://github.com/geo-gs/sawwit/blob/master/src/main/java/com/sawwit/siad/WalletHelper.java#L50

Additionally, NghtRppr (Discord handle) already wrote a patch that looks functional: https://gist.github.com/rmcdaniel/acddb741377f408165806069e64aa7c4

DavidVorick:
I'm going to reject this idea. Communicating on top of a blockchain is very wasteful of the resources of everyone on the network. And while there's nothing in the official protocol that we can do to stop you, I would really ask that you don't do it. The Sia blockchain is already bloated enough, if storage transactions have to start competing in a fee market with chat apps, it might start to cost $5 or $10 just to get started on Sia.

DavidVorick:
*wrong button, will leave open for discussion

tbenz9:
I agree with DavidVorick, I think this is an interesting idea, but doesn't fit in with the goal and purpose of Sia itself.  Storing this type of data on the blockchain is going to cause issues long term and unneeded growth of the blockchain itself.

[...]

PR #2704, opened 2018-01-30
Add: Arbitrary Data to Wallet Api

geo-gs:
Pull request to fulfil git issue #2686

Making a pull request out of rmcdaniel's proposed patch:
  https://gist.github.com/rmcdaniel/acddb741377f408165806069e64aa7c4


DavidVorick:
This change makes me nervous. Though it's something you could always do with a modified client, if we put it right in the API it becomes a lot easier to spam the blockchain and force a fee market.

In an environment where cryptocurrencies are intentionally being spammed, presumably to take advantage of the relative price-insensitivity of fees, I'm not sure we are ready to enable this.

Issue #2702, opened 2018-01-30
Clean up lint errors


tbenz9:
If anyone is looking to contribute, some easy, beginner level tasks would involve adding more packages (one at a time please) to our $lintpkgs in our Makefile and cleaning up all of the linting errors
That's a good way to learn go if you aren't familiar, and it's something that would benefit the project.

Makefile: https://github.com/NebulousLabs/Sia/blob/master/Makefile

tbenz9:
Great work @mharkus & @mikkeljuhl!! 
Looks like these are all that are left:
- ./modules/host/contractmanager
- ./modules/miner
- ./node/api
- ./sync
- ./types

It'd be awesome to finish this up before 1.3.2 drops.

tbenz9:
@mikkeljuhl thanks!  v1.3.2 is arriving this week, think you could get the last 2 packages Lint'ed in the next day or two?

mharkus:
Sorry missed the message 4 days ago. Will try to to wrap it up tonight if @mikkeljuhl isn't working on it yet.

mikkeljuhl:
@mharkus good job, didn't see this until now!

mharkus:
I guess we can close this issue now that the last remaining lint-cleanup tasked has been merged. 
#2786

PR #2705, opened 2018-01-30
Adding ./modules/renter to lintpkgs and fixing all the lint errors. #…

mharkus:
This PR will add the ./modules/renter in lintpkgs in Makefile and also fixes the lint errors. This will be one of the many tasks for #2702 to clean up lint errors across the project.


lukechampine:
Thanks! I know this isn't the most glamorous work, but it helps enforce a high level of quality in the long term.

PR #2710, opened 2018-01-31
Added module ./modules/wallet to lint packages and fixed lint errors

mharkus:
This PR will add the module ./modules/wallet to lint packages and also fixing the lint errors. For Issue #2702

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,8 +50,8 @@ vet: release-std
 	go vet $(pkgs)
 
 # will always run on some packages for a while.
-lintpkgs = ./modules ./modules/gateway ./modules/host ./modules/renter ./modules/renter/hostdb ./modules/renter/contractor \
-           ./node ./node/api/server ./persist ./siatest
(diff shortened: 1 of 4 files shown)


DavidVorick:
Thank you very much! Reviewing now

Makefile
```diff
@@ -50,8 +50,8 @@ vet: release-std
 	go vet $(pkgs)
 
 # will always run on some packages for a while.
-lintpkgs = ./modules ./modules/gateway ./modules/host ./modules/renter ./modules/renter/hostdb ./modules/renter/contractor \
-           ./node ./node/api/server ./persist ./siatest
+lintpkgs = ./modules ./modules/wallet ./modules/gateway ./modules/host ./modules/renter ./modules/renter/hostdb \
```
DavidVorick: generally speaking we try to sort long lists alphabetically, which means `modules/wallet` would come after `modules/renter/contractor`. Sorting just makes it easier to scan really long lists of things.

It looks like other parts of this list aren't 100% sorted either, if you want to fix that up at the same time.

DavidVorick:
Except for the alphabetizing, this pull request looks good to go.

PR #2712, opened 2018-01-31
Add TestGroup to siatest package

ChrisSchinnerl:
This PR adds the TestGroup struct to the new siatest package. It allows users to spin up large groups of nodes ready for tests with minimal effort. 
This will allow us to easily create more reliable tests for the upcoming downloader overhaul and various other PRs that are currently queued for testing.


siatest/testgroup.go
```diff
@@ -0,0 +1,341 @@
...
+		miners  map[*TestNode]struct{}
+
+		dir string
+	}
+)
+
+var (
+	// defaultAllowance is the allowance used for the group's renters
+	defaultAllowance = modules.Allowance{
+		Funds:       types.SiacoinPrecision.Mul64(100000),
```
DavidVorick: 100e3

siatest/testgroup.go
```diff
@@ -0,0 +1,341 @@
...
+		Hosts:       10,
```
DavidVorick: 4

DavidVorick: Eh actually 5

siatest/testgroup.go
```diff
@@ -0,0 +1,341 @@
...
+		RenewWindow: 5,
+	}
+)
+
+// announceHosts adds storage to each host and announces them to the group
+func announceHosts(hosts map[*TestNode]struct{}) error {
+	for host := range hosts {
+		if err := host.HostStorageFoldersAddPost(host.Dir, 1048576); err != nil {
+			return err
+		}
```
DavidVorick: Adding a storage folder should be separate from announcing

siatest/testgroup.go
```diff
@@ -0,0 +1,341 @@
...
+	return
+}
+
+// fullyConnectNodes takes a list of nodes and connects all their gateways
+func fullyConnectNodes(nodes []*TestNode) error {
+	// Fully connect the nodes
+	for i, nodeA := range nodes {
+		for _, nodeB := range nodes[i+1:] {
+			if err := nodeA.GatewayConnectPost(nodeB.GatewayAddress()); err != nil {
+				return build.ExtendErr("failed to connect to peer", err)
```
DavidVorick: Is it possible that this returns 'already a peer' errors?

siatest/testgroup.go
```diff
@@ -0,0 +1,341 @@
...
+
+// waitForContracts waits until the renters have formed contracts with the
+// hosts in the group.
+func waitForContracts(miner *TestNode, renters map[*TestNode]struct{}, hosts map[*TestNode]struct{}) error {
+	expectedContracts := defaultAllowance.Hosts
+	if uint64(len(hosts)) < expectedContracts {
+		expectedContracts = uint64(len(hosts))
+	}
+	for renter := range renters {
+		err := build.Retry(100, time.Second, func() error {
```
DavidVorick: No need to wait a full second between tries, can wait 100ms

siatest/testgroup_test.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+}
+
+// TestCreateTestGroup tests NewGroup with no renter or host
+func TestNewGroupNoRenterHost(t *testing.T) {
+	if testing.Short() {
+		t.SkipNow()
+	}
+	// Create a group with nothing but a single miner
+	groupParams := GroupParams{
```
DavidVorick: "a single miner"

miners: 5

DavidVorick:
Overall this looks pretty good

siatest/testgroup.go
ChrisSchinnerl: Hm... that could probably happen if you have the nodes A, B and C and they connect in the following order: AB, AC, BC. 
B and C might already have connected due to being connect via A. This is probably not very likely for small groups but I will add a check.

siatest/testgroup.go
ChrisSchinnerl: I would recommend to leave it at 1 second and add a todo. Forming contracts is incredibly slow at the moment. If we change it to 100 milliseconds, 100 tries are not enough. That is not a bad thing usually but we might not want to mine >100 blocks while waiting for the renters to form contracts.

siatest/testgroup.go
```diff
@@ -0,0 +1,365 @@
...
+		}
+		if err := host.HostAnnouncePost(); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// Close closes the group and all its nodes
+func (tg *TestGroup) Close() (err error) {
```
DavidVorick: All methods of a type should be grouped together.

Beyond that, we try to keep everything alphabetically sorted.

siatest/testgroup.go
```diff
@@ -0,0 +1,365 @@
...
+		if err := renter.RenterPost(defaultAllowance); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// NewGroupFromTemplate will create hosts, renters and miners according to the
+// settings in groupParams.
+func NewGroupFromTemplate(groupParams GroupParams) (*TestGroup, error) {
```
DavidVorick: We try to cluster the exported functions (even that don't act on an object) together the same way that we try to cluster functions of an object together.

And by extension, we cluster exported functions of an object together, and the unexported functions of an object are clustered together in a separate cluster.

siatest/testgroup_test.go
```diff
@@ -0,0 +1,95 @@
...
+	// Create the group
+	_, err := NewGroupFromTemplate(groupParams)
+	if err == nil {
+		t.Fatal("Creating a group without miners should fail: ", err)
+	}
+}
+
+// TestCreateTestGroup tests NewGroup with no renter or host
+func TestNewGroupNoRenterHost(t *testing.T) {
+	if testing.Short() {
```
DavidVorick: This can probably be vlong

siatest/testgroup_test.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+		if len(wtg.ConfirmedTransactions) == 0 {
+			t.Errorf("Node has 0 confirmed funds")
+		}
+	}
+}
+
+// TestCreateTestGroup tests NewGroup without a miner
+func TestNewGroupNoMiner(t *testing.T) {
+	if testing.Short() {
```
DavidVorick: also vlong

Issue #2676, opened 2018-01-22
Add clear error message when siad cannot bind to a port.


tbenz9:
# Feature Request
Add a clear error message if the gateway fails to start because of `address already in use`.  This would require a small amount of error parsing in the gateway module, and an additional message to the user if that particular error is found.

**For example:**
I think `bind address already in use; are you running another instance of siad?` would be clear enough.

## Background information:
tbenz9: I've seen a lot of GitHub issues and support requests come through lately with the "bind: address already in use" error.  How hard would it be to add a check into Sia that fails gracefully with a clear error message if a siad process is already running?
Nemo: In theory that shouldn't be too difficult. Just need to check for that specific error and then print a more helpful error instead of just forwarding it to the user

Dynamikey:
Looks interesting!

lukechampine:
For contributors looking to implement this:

Don't do this:
```go
func isAddrInUseErr(err error) bool {
	return strings.Contains(err.Error(), "address already in use")
}
```
Instead do this:
```go
func isAddrInUseErr(err error) bool {
	if opErr, ok := err.(*net.OpError); ok {
		if syscallErr, ok := opErr.Err.(*os.SyscallError); ok {
			return syscallErr.Err == syscall.EADDRINUSE
		}
	}
	return false
}
```

PR #2715, opened 2018-02-02
Adds a clear error message when a 'bind: address already in use' error occured

mharkus:
This PR will add the feature request in #2676. siad will output the following when the address specified in config.Siad.APIaddr has already been in use by other applications:

Sia Daemon v1.3.1
Loading...
listen tcp 127.0.0.1:9980: bind: address already in use; are you running another instance of siad?


DavidVorick:
looks good to me @lukechampine

PR #2719, opened 2018-02-02
Added ./build to lintpkgs and fix lint errors

mharkus:
This PR adds ./build module to lint packages and also fixes the lint errors. Could someone verify the comment for VLONG? Not quite sure if it's correct. For issue #2702

### DIFF
--- a/build/vlong_on.go
+++ b/build/vlong_on.go
@@ -2,4 +2,5 @@
 
 package build
 
+// VLONG set to true will make the build time very long
 const VLONG = true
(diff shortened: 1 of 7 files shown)


build/vlong_on.go
```diff
@@ -2,4 +2,5 @@
 
 package build
 
+// VLONG set to true will make the build time very long
```
lukechampine: `// VLONG enables a set of test cases that take a very long time to run.`

PR #2723, opened 2018-02-03
Add ./cmd/siac to lint pkgs fixes #2705

mjmay08:
Add ./cmd/siac to list of packages to be linted. There are more packages that still need to be added to lint. This is my first small attempt at contributing to Sia.


lukechampine:
Thanks!

PR #2730, opened 2018-02-04
switch renter memory model to support multiple working threads

DavidVorick:
Changed the memory model from a model that could only support one thread (the upload thread) to a model that can now support multiple threads (upload thread and download thread).

### DIFF
--- a/modules/renter/renter.go
+++ b/modules/renter/renter.go
@@ -182,9 +182,14 @@ type Renter struct {
 	// renter can allocate before hitting the cap, and newMemory is a channel
 	// used to inform sleeping threads (the download loop and upload loop) that
 	// memory has become available.
+	//
+	// The newMemory channel is closed and replaced each time new memory becomes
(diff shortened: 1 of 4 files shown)


modules/renter/renter.go
```diff
@@ -182,9 +182,14 @@ type Renter struct {
...
 	// memory has become available.
+	//
+	// The newMemory channel is closed and replaced each time new memory becomes
+	// available, that way all threads can be notified together. The functions
+	// for working with the memory (managedMemoryGet, managedMemoryReturn)
+	// handle the creation and closing of the channels automatically.
 	baseMemory      uint64
 	memoryAvailable uint64
-	newMemory       chan struct{}
+	newMemory       *chan struct{}
```
ChrisSchinnerl: Why is `newMemory` a pointer now?

DavidVorick:
closing this, these commits, as well as commits that re-do this system, are in my new pr

PR #2731, opened 2018-02-04
make the chunk responsible for stanby workers

DavidVorick:
previously, workers would remember which chunks they were on standby for, and
periodically check whether they were needed to perform extra work. This had some
downtime issues, and complicated a lot of the worker functions.

Now, the chunk itself remembers which workers are on standby, and will notify
them if there's a failure that requires another worker to take action.

### DIFF
--- a/modules/renter/worker.go
+++ b/modules/renter/worker.go
@@ -22,27 +22,21 @@ type worker struct {
 	hostPubKey types.SiaPublicKey
 	renter     *Renter
 
-	// Channels that inform the worker of kill signals and of new work.
-	downloadChan         chan downloadWork // higher priority than all uploads
-	killChan             chan struct{}     // highest priority
-	priorityDownloadChan chan downloadWork // higher priority than downloads (used for user-initiated downloads)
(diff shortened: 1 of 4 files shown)


modules/renter/worker.go
```diff
@@ -22,27 +22,21 @@ type worker struct {
...
+	// Uploading variables.
+	unprocessedChunks         []*unfinishedChunk // Yet unprocessed work items.
+	uploadChan                chan struct{}      // Notifications of new work.
+	uploadConsecutiveFailures int                // How many times in a row uploading has failed.
+	uploadRecentFailure       time.Time          // How recent was the last failure?
+	uploadTerminated          bool               // Have we stopped uploading?
+
+	// Utilities.
+	killChan chan struct{} // Worker will shut down if a signal is sent down this channel.
+	mu       sync.Mutex
```
DavidVorick: Though I re-arranged a bunch of stuff, I also deleted the 'standbyChunks' field.

DavidVorick:
closing this, these commits (and more) are in my download overhaul pr

PR #2734, opened 2018-02-05
fix http server race

ChrisSchinnerl:
This PR fixes the random race condition when we call `WriteJSON`.
It seems like the std lib flushes the underlying buffer in `FinishRequest`, but before calling `FinishRequest` it cancels the context. This means, that `r.Context().Done()` will always trigger before `done` in our `select` and might cause a situation where `FinishRequest` is called at the same time as `ServeHTTP`. Both of those functions use the same buffer.

### DIFF
--- a/node/api/routes.go
+++ b/node/api/routes.go
@@ -145,7 +145,6 @@ func cleanCloseHandler(next http.Handler) http.Handler {
 		}(w, r)
 		select {
 		case <-done:
-		case <-r.Context().Done():
 		}
 
 		// Sanity check - thread should not take more than an hour to return. This


DavidVorick:
When you remove this line of code, is it still guaranteed to terminate at some point? If we're removing a return condition, is it possible that something will hang?

ChrisSchinnerl:
@DavidVorick I don't think that's possible. It seems like the only thing that happens after terminating the context is cleanup and recycling of buffers so that should be fine. Everything that might hang before terminating the context would also cause the current code to hang.

PR #2737, opened 2018-02-06
Add upload and download helpers to siatest package

ChrisSchinnerl:
This PR extends the siatest package to be able to easily create TestFiles and upload/download them to/from hosts.


ChrisSchinnerl:
I had to change the `Download` to return a channel instead of an error. Before that, we just waited for 100ms and hoped that the setup of the download would be done by then. That also introduced a constant 100ms latency for async download calls.
Now the `Download` method only returns when the setup is actually over and the caller knows instantly if there was an error during the setup. 
The reason for this change is, that there was no guarantee that the download actually made it into the download queue. This means a call to `/renter/downloads` might miss the previously started download due to the setup code not being finished yet.

Makefile
```diff
@@ -38,7 +38,8 @@ run = .
 pkgs = ./build ./cmd/siac ./cmd/siad ./compatibility ./crypto ./encoding ./modules ./modules/consensus ./modules/explorer \
        ./modules/gateway ./modules/host ./modules/host/contractmanager ./modules/renter ./modules/renter/contractor       \
        ./modules/renter/hostdb ./modules/renter/hostdb/hosttree ./modules/renter/proto ./modules/miner ./modules/wallet   \
-       ./modules/transactionpool ./node ./node/api ./persist ./siatest ./node/api/server ./sync ./types
+       ./modules/transactionpool ./node ./node/api ./persist ./siatest ./siatest/consensus ./siatest/renterhost           \
```
DavidVorick: let's just have it be `./siatest/renter` for now, when we write tests for the host I think they'll generally take a slightly different shape.

modules/renter.go
```diff
@@ -293,7 +293,7 @@ type Renter interface {
 
 	// Download performs a download according to the parameters passed, including
 	// downloads of `offset` and `length` type.
-	Download(params RenterDownloadParameters) error
+	Download(params RenterDownloadParameters) chan error
```
DavidVorick: I am hesitant to approve of a change like this, especially in a pull request that's dedicated mostly to testing.

Was there any particular reason you felt it was necessary?

node/api/renter.go
```diff
@@ -356,24 +356,21 @@ func (api *API) renterDownloadHandler(w http.ResponseWriter, req *http.Request,
...
 	}
 
-	if params.Async { // Create goroutine if `async` param set.
-		// check for errors for 5 seconds to catch validation errors (no file with
-		// that path, invalid parameters, insufficient hosts, etc)
-		errchan := make(chan error)
-		go func() {
-			errchan <- api.renter.Download(params)
-		}()
+	errChan := api.renter.Download(params)
```
DavidVorick: Again, not sure I consider this change to be an improvement over the original.

siatest/renterhost/renterhost_test.go
```diff
@@ -0,0 +1,100 @@
...
+import (
+	"bytes"
+	"testing"
+
+	"github.com/NebulousLabs/Sia/siatest"
+)
+
+// TestRenterHost executes a number of subtests using the same TestGroup to
+// save time on initialization
+func TestRenterHost(t *testing.T) {
```
DavidVorick: Overall test layout looks good, but I think fine to just change it to 'TestRenter'

siatest/renterhost/renterhost_test.go
```diff
@@ -0,0 +1,100 @@
...
+	tg, err := siatest.NewGroupFromTemplate(groupParams)
+	if err != nil {
+		t.Fatal("Failed to create group: ", err)
+	}
+	defer func() {
+		if err := tg.Close(); err != nil {
+			t.Fatal(err)
+		}
+	}()
+	// Specifiy subtests to run
```
DavidVorick: probably want to insert some spacing between creating the group and starting the subtests, as I think they are separate thoughts.

siatest/renterhost/renterhost_test.go
```diff
@@ -0,0 +1,100 @@
...
+		t.Fatal("Failed to start upload: ", err)
+	}
+	// Wait until upload reached 100% progress
+	if err := renter.WaitForUploadProgress(file, 1); err != nil {
+		t.Error(err)
+	}
+	// Wait until upload reaches len(tg.Hosts()) redundancy
+	if err := renter.WaitForUploadRedundancy(file, float64(len(tg.Hosts()))); err != nil {
+		t.Error(err)
+	}
```
DavidVorick: We should probably combine these 4 calls (NewFile, Uplaod, WaitForUploadProgress, WaitForUploadRedundancy) into a single call named `UploadFile` or something like that. These methods can still all exist individually, but we shouldn't need to call them all in a row just to perform a file upload.

Also, these functions that the corresponding helpers should live in the top level `siatest` package, so that other packages like the host tester and wallet can use them. We want the wallet  to be able to upload and download for example so that we can see that it correctly updates when it receives file contract revenue. (which I don't think that it actually does at the moment).

siatest/testnode.go
```diff
@@ -22,6 +30,165 @@ type TestNode struct {
 	primarySeed string
 }
 
+// DownloadToDisk downloads a previously uploaded file. The file will be downloaded
+// to a random location and returned as a TestFile object.
+func (tn *TestNode) DownloadToDisk(tf *TestFile, async bool) (*TestFile, error) {
```
DavidVorick: All of these helper functions should probably be split off into their respective files. So the download and upload stuff would go in renter.go, the mining stuff in miner.go, etc.

siatest/renterhost/renterhost_test.go
```diff
@@ -0,0 +1,100 @@
...
+	// Download the file synchronously directly into memory and compare the
+	// data to the original
+	downloadedData, err := renter.Download(file)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if bytes.Compare(downloadedData, file.Bytes()) != 0 {
```
DavidVorick: The compare piece can probably be a part of the `Download` call, makes sense to me that you would automatically check the integrity of the file after you download it every single time in a testing environment. Doesn't even need to be a separate function, as I can't imagine a situation where checking the integrity of a file after downloading it would be unwanted (unlike waiting for an upload to hit full redundancy, where it's easy to see testing situations where you don't want the file to hit full redundancy).

siatest/renterhost/renterhost_test.go
```diff
@@ -0,0 +1,100 @@
...
+	if err := renter.WaitForUploadProgress(file, 1); err != nil {
+		t.Error(err)
+	}
+	// Wait until upload reaches len(tg.Hosts()) redundancy
+	if err := renter.WaitForUploadRedundancy(file, float64(len(tg.Hosts()))); err != nil {
+		t.Error(err)
+	}
+	// Download the file synchronously directly into memory and compare the
+	// data to the original
+	downloadedData, err := renter.Download(file)
```
DavidVorick: Does this fetch the data by streaming it? I would probably call it `DownloadByStreaming` or something like that.

siatest/testfile.go
```diff
@@ -0,0 +1,60 @@
...
+// Compare is a convenience function that compares the contents of two
+// TestFiles on disk. Its behavior is similar to bytes.Compare.
+func (tf *TestFile) Compare(tf2 *TestFile) int {
+	tfData, err := ioutil.ReadFile(tf.path)
+	tf2Data, err2 := ioutil.ReadFile(tf2.path)
+	if err != nil || err2 != nil {
+		// Print error and return a mismatch instead of returning the error.
+		// This should be sufficient for a testing environment and makes for
+		// cleaner tests.
+		println(build.ComposeErrors(err, err2))
```
DavidVorick: You want to return an error so that the test can fail. We definitely don't read through the print statements of travis and appveyor every time that the build passes.

modules/renter.go
ChrisSchinnerl: Yes, without this change there is a race condition between the Download api call and the downloadqueue api call. Previously if you called Download, there was no guarantee that the download actually made it into the downloadqueue before the next call to downloadqueue. That's because we only waited for a fixed time of 100ms instead of returning a channel.

node/api/renter.go
ChrisSchinnerl: See the explanation above. It also gets rid of the hardcoded 100ms latency for async download calls.

siatest/renterhost/renterhost_test.go
ChrisSchinnerl: It downloads using the httpresponse flag

siatest/renterhost/renterhost_test.go
ChrisSchinnerl: I am not sure it can. `TestFile.Download(...)` doesn't require the file to be stored locally. In the future we might add calls like `TestFile.Delete(...)` to delete the local copy of a file. This means, that we might not have data to compare the download to.
We could store a hash in the `TestFile` object for that purpose but that still would allow us to check the contents when we use blocking download calls.

PR #2738, opened 2018-02-07
Added ./encoding to lint packages

mharkus:
This PR is for #2702


DavidVorick:
Need to run go fmt on this code.

PR #2745, opened 2018-02-08
Fix hostdb deadlock and update constants

ChrisSchinnerl:
This PR fixes a deadlock while scanning a large number of hosts and changes some constants to make scanning faster. On my machine the download bandwidth hardly reached 20 kib/s with 250 threads so I guess those settings are sane.

This should also fix this [issue](https://github.com/NebulousLabs/Sia/issues/2741).

### DIFF
--- a/modules/renter/hostdb/consts.go
+++ b/modules/renter/hostdb/consts.go
@@ -52,7 +52,7 @@ var (
 	// hostCheckupQuantity specifies the number of hosts that get scanned every
 	// time there is a regular scanning operation.
 	hostCheckupQuantity = build.Select(build.Var{
-		Standard: int(200),
+		Standard: int(2500),
 		Dev:      int(6),
 		Testing:  int(5),
 	}).(int)
@@ -60,7 +60,7 @@ var (
 	// scanningThreads is the number of threads that will be probing hosts for
 	// their settings and checking for reliability.
 	maxScanningThreads = build.Select(build.Var{
-		Standard: int(40),
+		Standard: int(80),
 		Dev:      int(4),
 		Testing:  int(3),
 	}).(int)
--- a/modules/renter/hostdb/scan.go
+++ b/modules/renter/hostdb/scan.go
@@ -39,9 +39,8 @@ func (hdb *HostDB) queueScan(entry modules.HostDBEntry) {
 	}
 
 	hdb.scanWait = true
-	scanPool := make(chan modules.HostDBEntry)
-
 	go func() {
+		scanPool := make(chan modules.HostDBEntry)
 		defer close(scanPool)
 
 		// Nobody is emptying the scan list, volunteer.
@@ -53,7 +52,9 @@ func (hdb *HostDB) queueScan(entry modules.HostDBEntry) {
 		}
 		defer hdb.tg.Done()
 
+		starterThread := false
 		for {
+			// If the scanList is empty, this thread can spin down.
 			hdb.mu.Lock()
 			if len(hdb.scanList) == 0 {
 				// Scan list is empty, can exit. Let the world know that nobody
@@ -62,6 +63,7 @@ func (hdb *HostDB) queueScan(entry modules.HostDBEntry) {
 				hdb.mu.Unlock()
 				return
 			}
+
 			// Get the next host, shrink the scan list.
 			entry := hdb.scanList[0]
 			hdb.scanList = hdb.scanList[1:]
@@ -75,7 +77,8 @@ func (hdb *HostDB) queueScan(entry modules.HostDBEntry) {
 			}
 
 			// Create new worker thread
-			if hdb.scanningThreads < maxScanningThreads {
+			if hdb.scanningThreads < maxScanningThreads || !starterThread {
+				starterThread = true
 				hdb.scanningThreads++
 				go func() {
 					hdb.threadedProbeHosts(scanPool)
--- a/modules/renter/hostdb/scan.go
+++ b/modules/renter/hostdb/scan.go
@@ -53,12 +53,13 @@ func (hdb *HostDB) queueScan(entry modules.HostDBEntry) {
 		defer hdb.tg.Done()
 
 		// Due to the patterns used to spin up scanning threads, it's possible
-		// that we get to thiis point while all scanning threads are currently
+		// that we get to this point while all scanning threads are currently
 		// used up, completing jobs that were sent out by the previous pool
-		// managing thread. This thread is at risk of deadlocking if there's not
-		// at least one scanning thread accepting work that it created itself,
-		// so we use a starterThread exception and spin up on-thread-too-many on
-		// the first iteration to ensure that we do not deadlock.
+		// managing thread. This thread is at risk of deadlocking if there's
+		// not at least one scanning thread accepting work that it created
+		// itself, so we use a starterThread exception and spin up
+		// one-thread-too-many on the first iteration to ensure that we do not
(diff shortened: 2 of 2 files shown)


S-anasol:
Tried to use maxScanningThreads ~100 for SiaHub, got crash every time.

Maybe coz of deadlock that you fixed...

S-anasol:
If i not mistaken - got wallet crash, tested it long time ago.
So I have kept default threads value and changed timings to make scan 'faster'

ChrisSchinnerl:
@S-anasol did siad crash right away? Maybe something else caused the problem and is fixed by now? I tested it with batches of 10,000 scans and 1000 threads and it still didn't crash on Ubuntu.

S-anasol:
> Maybe something else caused the problem and is fixed by now?
maybe, tested it in August 2017 on Debian.

Changed mine from 40 to 100 now, will check it later.
Looks like it work now.
Wallet sources from end of december.

ChrisSchinnerl:
@S-anasol Great! Let me know on Discord if you run into any errors.

modules/renter/hostdb/scan.go
```diff
@@ -74,14 +71,27 @@ func (hdb *HostDB) queueScan(entry modules.HostDBEntry) {
...
+				}
+				// Spin up scanning thread
 				hdb.scanningThreads++
 				go func() {
 					hdb.threadedProbeHosts(scanPool)
 					hdb.mu.Lock()
 					hdb.scanningThreads--
+					if !threadsDoneClosed && hdb.scanningThreads < maxScanningThreads {
+						close(threadsDone)
+						threadsDoneClosed = true
```
lukechampine: I think what you want here is a `sync.WaitGroup`, not a channel+boolean.

modules/renter/hostdb/scan.go
```diff
@@ -74,14 +71,27 @@ func (hdb *HostDB) queueScan(entry modules.HostDBEntry) {
 				entry = recentEntry
 			}
 
-			// Create new worker thread
+			// Create new worker thread. newPool allows us to ignore
+			// maxScanningThreads once as explained in the comment above.
```
lukechampine: `newPool`?

S-anasol:
@ChrisSchinnerl 
Seems like crash gone. Few scans passed since yesterday.

```
siad                             RUNNING    pid 9307, uptime 14:24:13
```

So probably some updates between august and december fixed it 🚀 
Then my warning about threads is out of date :)

modules/renter/hostdb/hostdb.go
```diff
@@ -44,11 +44,13 @@ type HostDB struct {
...
 	// handful of goroutines constantly waiting on the channel for hosts to
 	// scan. The scan map is used to prevent duplicates from entering the scan
-	// pool.
+	// pool. The scanWG is used to wait for the last generation of scanning
+	// threads to stop before creating new ones.
 	scanList        []modules.HostDBEntry
 	scanMap         map[string]struct{}
 	scanWait        bool
 	scanningThreads int
+	scanWG          *sync.WaitGroup
```
lukechampine: hmm. I would prefer to keep this out of the hostdb object if possible. The less shared state, the better. I was also under the impression that waitgroups couldn't be reused, but apparently that is not true. Still, it feels like `scanWait` and `scanWG` overlap in responsibility somewhat. If `scanWait` can tell you whether there's a scan running, why do we need to call `scanWG.Wait` at all?

ChrisSchinnerl: `scanWait` tells you if the "main thread" is running. The one that spawns scanning threads. Setting `scanWait` is therefore required to avoid creating a new "main thread" and requires the lock to be held. Unfortunately you can't hold the lock while waiting on `scanWG`. 
The issue is that you can't unlock the lock to wait on `scanWG` after checking if `scanList` is empty. If you do, there is a chance that new scan jobs are added to the list while we wait for the threads to finish. Those jobs are not executed and therefore cause some tests to randomly fail.

modules/renter/hostdb/consts.go
```diff
@@ -52,15 +52,15 @@ var (
...
+		Standard: int(10000),
 		Dev:      int(6),
 		Testing:  int(5),
 	}).(int)
 
 	// scanningThreads is the number of threads that will be probing hosts for
 	// their settings and checking for reliability.
 	maxScanningThreads = build.Select(build.Var{
-		Standard: int(40),
+		Standard: int(50),
```
DavidVorick: We need someone with a Mac to try this out when the maxScanningThreads value has been set to 200. I think we would run into the 'too many open files' error due to all of the open network connections, but I'm not sure. If it doesn't cause problems, going up to 200 threads would definitely help scanning go faster.

modules/renter/hostdb/consts.go
```diff
@@ -52,15 +52,15 @@ var (
 	// hostCheckupQuantity specifies the number of hosts that get scanned every
 	// time there is a regular scanning operation.
 	hostCheckupQuantity = build.Select(build.Var{
-		Standard: int(200),
+		Standard: int(10000),
```
DavidVorick: Should be 10e3.

How long does it take the hostdb to get through 10e3 threads? If each one takes 2 minutes and we have 50 threads, that's 6.6 hours, which I think is too long to be useful.

Should probably drop it to 2,500 unless we get confirmation from osx testing that we can jump to 200 threads. Then 10,000 is probably okay.

modules/renter/hostdb/scan.go
```diff
@@ -272,11 +287,6 @@ func (hdb *HostDB) managedScanHost(entry modules.HostDBEntry) {
 
 // threadedProbeHosts pulls hosts from the thread pool and runs a scan on them.
 func (hdb *HostDB) threadedProbeHosts(scanPool <-chan modules.HostDBEntry) {
-	err := hdb.tg.Add()
-	if err != nil {
-		return
-	}
-	defer hdb.tg.Done()
```
DavidVorick: Is there a reason that we are eliminating this call to `tg`? I don't think we want the hostdb returning successfully after a call to close unless `threadedProbeHosts` has shut down, which we achieve by adding it to the `tg`.

modules/renter/hostdb/scan.go
```diff
@@ -74,14 +70,33 @@ func (hdb *HostDB) queueScan(entry modules.HostDBEntry) {
...
+				}
+				// Spin up scanning thread
 				hdb.scanningThreads++
 				go func() {
 					hdb.threadedProbeHosts(scanPool)
 					hdb.mu.Lock()
 					hdb.scanningThreads--
+					// There is at least room for 1 new scanning thread now.
+					// Signal the world that a new thread can be started.
+					hdb.scanWait = false
```
DavidVorick: I think that this is incorrect. The purpose of the `scanWait` variable is to ensure that there's only one thread spinning up scanning threads. So the thread that should be setting `scanWait` to false is the thread that'sspinning up other threads, not the other threads being spun up. Here I think you end up settings `scanWait` to false a whole bunch of times, which can result in having many scanning pool threads.

modules/renter/hostdb/consts.go
DavidVorick: I guess for the RC we can actually try putting it at a high value like 200 and see if anyone has issues.

DavidVorick: on the other hand, that's a move that will potentially break master, something we probably shouldn't be doing.

PR #2747, opened 2018-02-08
Lint errors: renter/proto + hostdb/hosttree

mikkeljuhl:
Also added renter/hostdb/hosttree, but didn't have any lint errors. 

#2702


lukechampine:
thanks!

PR #2758, opened 2018-02-12
Download overhaul

DavidVorick:
I have replaced the old download code with some newer, fancier download code. It should be faster, less error prone, and easier to understand. It also has more flexibility for adding features in the future. And it uses the same memory pool as the renter.

I haven't done any live-testing with this code yet, but it's passing the full test suite from our previous code.

I also haven't  yet gone through the full PR with a fine-toothed comb, though I kept a long checklist of things to do while I was working, and I think I've gotten pretty much everything properly organized.

I think that I made a slight compromise to the repair code when I changed the memory model. The new memory model assumes that you can block waiting for memory and not need to listen on a channel for new stuff, but the repair code actually does stutter if it's stuck waiting for memory and new repair jobs are coming in. If this proves to be problematic, we'll have to switch the repair heap over to the same mutli-threaded-queue-ing system that the download heap has. This system is the better system, so it's a change we should make at some point anyway.


DavidVorick:
Worth noting that in addition to replacing the flow for processing download requests, I also updated the API to include more fields when asking for the download history. These new fields are not documented in the API docs yet, I still need to do that.

I strengthened just a few of our tests which were failing for the downloads. Now they fail in a way that makes it a little easier to pinpoint what is going wrong.

I completely overhauled the download buffer system.

The memory manager is now it's own object and really could be it's own package. Something we might want to consider doing at some point, though right now I don't think we have any other use case for it than the renter.

I added a new prefix to our conventions, the prefix 'static'. That prefix just indicates that you can access the field without any mutex because the field does not change at all. It's also a warning that the field should not be changed because it's going to be getting accessed without any mutexes. I think this is something we exploit a decent amount in our code, but up until now the only way to know when a field was static was to have it memorized, often it wasn't even documented in the struct comment, let alone the functions where the field was being used. Now it's pretty in-your-face, which hopefully makes it easier to leverage this type of function and variable.

I may have done a couple other things as well. This was definitely a coding adventure for me, got to put to practice a bunch of ideas I've been stewing on but haven't really had time to try out.

lukechampine:
I can do a surface-level review of this for now, but I can't give any feedback on the design until I have at least a rudimentary understanding of how the various functions and types fit together. A high-level overview would be very helpful.

modules/renter/renter.go
```diff
@@ -2,30 +2,19 @@
...
-// instead of occasionally rebuilding the whole file matrix it has just a
-// single matrix that it's constantly pulling chunks from. Have a separate loop
-// which goes through the files and adds them to the matrix. Have the loop
-// listen on the channel for new files, so that they can go directly into the
-// matrix.
+// NOTE: Some of the concurrency patterns in the renter are fairly complex. A
+// lot of this has been cleaned up, though some shadows of previous demons still
+// remain. Be careful when working with anything that involves concurrency.
 
 // TODO: Allow the 'baseMemory' to be set by the user.
```
ChrisSchinnerl: `programming task` ?

modules/renter/worker.go
```diff
@@ -16,103 +14,160 @@ import (
...
+	requiredCooldown := downloadFailureCooldown
+	for i := 0; i < w.downloadConsecutiveFailures && i < maxConsecutivePenalty; i++ {
+		requiredCooldown *= 2
+	}
+	return time.Now().Before(w.downloadRecentFailure.Add(requiredCooldown))
+}
+
+// onUploadCooldown returns true if the worker is on cooldown from failed
+// uploads.
+func (w *worker) onUploadCooldown() bool {
```
ChrisSchinnerl: Should we separate upload and download cooldowns? Many issues probably affect both.

modules/renter/worker.go
```diff
@@ -16,103 +14,160 @@ import (
...
-		var sleepDuration time.Duration
-		w.mu.Lock()
-		numStandby := len(w.standbyChunks)
+// managedNextDownloadChunk will pull the next potential chunk out of the work
+// queue for downloading.
+//
+// Concurrency in this function is a little funky because we do not want to be
+// holding the lock when we call 'processDownloadChunk'.
+func (w *worker) managedNextDownloadChunk() *unfinishedDownloadChunk {
+	w.mu.Lock()
```
ChrisSchinnerl: can we do `defer w.mu.Lock()` ?

modules/renter/workerdownload.go
```diff
@@ -1,55 +1,200 @@
...
-		pieceIndex    uint64
-		workerID      types.FileContractID
+// returnMemory will check on the status of all the workers and pieces, and
+// determine how much memory is safe to return to the renter. This should be
+// called each time a worker returns, and also after the chunk is recovered.
+func (udc *unfinishedDownloadChunk) returnMemory() {
+	// The maximum amount of memory is the pieces completed plus the number of
+	// workers remaining.
+	initialMemory := uint64(udc.staticOverdrive+udc.erasureCode.MinPieces()) * udc.staticPieceSize
+	maxMemory := uint64(udc.workersRemaining+udc.piecesCompleted) * udc.staticPieceSize
```
ChrisSchinnerl: Should we sanity check for `udc.workersRemaining+udc.piecesCompleted < udc.erasureCode.MinPieces()` ?

modules/renter/download.go
```diff
@@ -1,805 +1,484 @@
...
-		// Progress variables.
-		atomicDataReceived uint64
-		downloadComplete   bool
-		downloadErr        error
-		finishedChunks     map[uint64]bool
-		offset             uint64
-		length             uint64
+		// Data progress variables.
+		atomicDataReceived        uint64 // Incremented as data completes, will stop at 100% file progress.
+		atomicTotalDataTransfered uint64 // Incremented as data arrives, includes overdrive, contract negotitiaon, etc.
```
tbenz9: Typo: negotiation

modules/renter/workerdownload.go
```diff
@@ -1,55 +1,200 @@
...
+	// If the maxMemory exceeds the inital memory, set the max memory equal to
```
tbenz9: Minor nitpick: comments should explain why the code does what it does, not how

modules/renter/downloadbuffer.go
```diff
@@ -0,0 +1,172 @@
...
+
+	// Block until we are the correct offset for the stream. The blocking is
+	// coordinated by a new mutex which gets added to an array. When the earlier
+	// data is written, the mutex will be unlocked, allowing us to write.
+	myMu := new(sync.Mutex)
+	myMu.Lock()
+	ddw.blockingWriteCalls = append(ddw.blockingWriteCalls, offset)
+	ddw.blockingWriteSignals = append(ddw.blockingWriteSignals, myMu)
+	ddw.mu.Unlock()
+	myMu.Lock()
```
tbenz9: Where is myMu.Unlock()?

ChrisSchinnerl: It's unlocked in a different thread that clears the slices and writes the data.

modules/renter/downloadchunk.go
```diff
@@ -0,0 +1,149 @@
...
+// recover a piece of a chunk from a host. It is a value in a map where the key
+// is the file contract id.
+type downloadPieceInfo struct {
+	index uint64
+	root  crypto.Hash
+}
+
+// unfinishedDownloadChunk contains a chunk for a download that is in progress.
+//
+// TODO: Explore making workersStandby a heap sorted by latency or whatever
```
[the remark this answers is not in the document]
tbenz9: I'd be happy to tear out the `TODO` statements and turn them into GH Issues.  If these get merged in I can start to create issues from them.

modules/renter/downloadheap.go
```diff
@@ -0,0 +1,209 @@
...
+
+// managedBlockUntilOnline will block until the renter is online. The renter
+// will appropriately handle incoming download requests and stop signals while
+// waiting.
+func (r *Renter) managedBlockUntilOnline() bool {
+	for !r.g.Online() {
+		select {
+		case <-r.tg.StopChan():
+			return false
+		case <-time.After(offlineCheckFrequency):
```
lukechampine: careful -- might leak a timer, depending on how frequently `managedBlockUntilOnline` is called

modules/renter.go
```diff
@@ -55,19 +55,18 @@ type ContractUtility struct {
...
+	Length          uint64 `json:"length"`          // The length requested for the download.
+	Offset          uint64 `json:"offset"`          // The offset within the siafile requested for the download.
+	SiaPath         string `json:"siapath"`         // The siapath of the file used for the download.
+
+	Completed           bool      `json:"completed"`           // Whether or not the download has completed.
+	EndTime             time.Time `json:"endtime"`             // The time when the download fully completed.
+	Error               string    `json:"error"`               // Will be the empty string unless there was an error.
+	Received            uint64    `json:"received"`            // Amount of data confirmed and decoded.
+	StartTime           time.Time `json:"starttime"`           // The time when the download was started.
+	TotalDataTransfered uint64    `json:"totaldatatransfered"` // The total amount of data transfered, including negotiation, overdrive, etc.
```
lukechampine: what is "overdrive?" It doesn't appear to be documented in the download code either. If it's relevant to API consumers, it needs to be precisely defined in the docs

tbenz9: I agree, it needs to be documented.

From Discord: I added an 'overdrive' feature to file downloads. It's not exposed in the api yet, but basically instead of downloading with 10 hosts, it'll download with 10+overdrive hosts. So if you set overdrive to 2, it'll download with 12 hosts.

Then, up to 2 of your hosts can be slow or fail and it won't impact the latency of the download
We experimented with this earlier, but with the previous download code any changes were pretty disruptive and unstable
now it's just an inherent part of downloading

DavidVorick:
@lukechampine for a high level overview, look at `download.go`, there's a great description of the overall flow.

Recommended review path is to start with download.go, then downloadheap.go, then worker.go, then workerdownload.go. downloadbuffer.go is probably safe to review last. memory.go can be reviewed independently I think.

modules/renter/download.go
```diff
@@ -1,805 +1,484 @@
...
+// initially instead of have it grab a piece if the latency of the worker is
+// higher than the faster workers. This will prevent the slow workers from
+// bottlenecking a chunk that we are trying to download quickly, though it will
+// harm overall system throughput because it means that the slower workers will
+// idle some of the time.
+
+// TODO: Currently the number of overdrive workers is set to '2' for the first 2
+// chunks of any user-initiated download. But really, this should be a parameter
+// of downloading that gets set by the user through the API on a per-file basis
+// instead of set by default.
```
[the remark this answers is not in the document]
DavidVorick: probably we should :)

modules/renter/download.go
```diff
@@ -1,805 +1,484 @@
...
+// That's going to need to be changed to a partial sector. This is probably
+// going to result in downloading that's 64-byte aligned instead of perfectly
+// byte-aligned. Further, the encryption and erasure coding may also have
+// alignment requirements which interefere with how the call to Sector can work.
+// So you need to make sure that in 'managedDownload' you download at least
+// enough data to fit the alignment requirements of all 3 steps (download from
+// host, encryption, erasure coding). After the logical data has been recovered,
+// we slice it to whatever is meant to be written to the underlying
+// downloadWriter, that code is going to need to be adjusted as well to slice
+// things in the right way.
```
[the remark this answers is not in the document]
DavidVorick: yeah, that's an intended part of the partial downloads upgrade. Hosts would add a flat required fee for downloading on top of the per-byte fee. That flat fee would pay for the required IOPS, the computational power for signing and updating, and for the bandwidth associated with negotiation.

It's probably roughly equivalent to downloading 16kb for most HDD hosts, and probably equivalent to downloading 1-4kb for SSD hosts, though we'd want to analyze that more rigorously.

modules/renter/download.go
```diff
@@ -1,805 +1,484 @@
...
-		for i := range contract.Pieces {
-			// Only add pieceSet entries for chunks that are going to be downloaded.
-			m, exists := d.pieceSet[contract.Pieces[i].Chunk]
-			if exists {
-				m[id] = contract.Pieces[i]
-			}
-		}
+// staticComplete is a helper function to indicate whether or not the download
+// has completed.
+func (d *download) staticComplete() bool {
```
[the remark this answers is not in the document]
DavidVorick: static is a new convention that indicates you can access the field holding or not holding a lock.

if it were `complete`, it would be implied that a lock is required. If it were `managedComplete`, it would be implied that you must **not** be holding a lock. But static implies that it doesn't matter. Seems weird for a function, but it's true because the implementation of the function uses channels.

modules/renter/download.go
```diff
@@ -1,805 +1,484 @@
...
+	maxChunk := (params.offset + params.length - 1) / params.file.staticChunkSize()
+
+	// For each chunk, assemble a mapping from the contract id to the index of
+	// the piece within the chunk that the contract is responsible for.
+	chunkMaps := make([]map[types.FileContractID]downloadPieceInfo, maxChunk-minChunk+1)
+	for i := range chunkMaps {
+		chunkMaps[i] = make(map[types.FileContractID]downloadPieceInfo)
+	}
+	params.file.mu.Lock()
+	for id, contract := range params.file.contracts {
```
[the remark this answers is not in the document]
DavidVorick: yeah, definitely should add a check here

modules/renter/download.go
```diff
@@ -1,805 +1,484 @@
...
-		if downloadComplete {
-			// Download has already failed.
-			continue
+		// Set the writeOffset within the destination for where the data should
+		// be written.
+		udc.staticWriteOffset = writeOffset
+		writeOffset += int64(udc.staticFetchLength)
+
+		// TODO: Pick a smarter value for the overdrive setting.
+		if i < 2 {
```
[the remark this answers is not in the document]
DavidVorick: We had something similar at one point, but I don't think it ever got merged.

modules/renter/download.go
```diff
@@ -1,805 +1,484 @@
...
-			delete(dw.buffer, dw.offset)
-
-			// Increment offset to point to the beginning of the next chunk.
-			dw.offset += len(data)
-			totalDataSent += len(data)
+// DownloadHistory returns the list of downloads that have been performed. Will
+// include downloads that have not yet completed. Downloads will be roughly, but
+// not precisely, sorted according to start time.
+//
+// TODO: Currently only remembers downloads from the current session.
```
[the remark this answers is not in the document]
DavidVorick: I think that people will start looking at these TODOs at some point. Once you've gotten started with some issues, gotten familiar with the codebase, then it makes sense to dive into the code and start tackling all the TODOs lying around.

Also, no reason it can't both be a todo and a programming task.

Perhaps we can add a programming task to take all of the TODOs and turn them into programming tasks :)

modules/renter/downloadbuffer.go
```diff
@@ -0,0 +1,172 @@
...
+
+// downloadDestinationWriter is a downloadDestination that writes to an
+// underlying data stream. The data stream is expecting sequential data while
+// the download chunks will be written in an aribtrary order using calls to
+// WriteAt. We need to block the calls to WriteAt until all prior data has been
+// written.
+//
+// TODO: There is no timeout protection here. If there's some misalignment of
+// data, we'll never know it'll just hang forever.
+type downloadDestinationWriter struct {
```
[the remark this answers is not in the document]
DavidVorick: I know I explained it in a comment somewhere, will have to explain it here as well. The download code ensures that it allocates memory for all earlier chunks before it begins work on any later chunks. So even if a later chunk finishes first, we know that the earlier chunk at the very least has had memory allocated, so it will be finishing up soon.

But... now that I think about it I don't think we have the `Close()` function implemented correctly here. It needs to unblock all of the other threads and have them fail.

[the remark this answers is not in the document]
DavidVorick: so there is a potential deadlock here

modules/renter/downloadchunk.go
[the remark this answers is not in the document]
DavidVorick: I definitely use the TODO statements when I return to the code months later. I want to make sure that every idea that I've had which is potentially worthwhile is documented somewhere, and the TODOs I feel are a pretty good way to do that.

I'm also in full support of taking some of the TODOs and turning them into GH issues for people to tackle. I wouldn't want to overwhelm people, so maybe only 2-5 per package at a time, but as people tackle the ones we've turned into issues, we can add the remaining ones as issues as well.

modules/renter/downloadheap.go
```diff
@@ -0,0 +1,209 @@
...
+	// is enough memory available to send off the work. If the chunk does not
+	// need any memory to be allocated, it should be given to the workers
+	// directly and immediately. This is actually a requirement in our memory
+	// model. If a download chunk does not need memory, that means that the
+	// memory has already been allocated and will actually be blocking new
+	// memory from being allocated until the download is complete. If the job is
+	// put in the heap and ends up behind a job which get stuck allocating
+	// memory, you get a deadlock.
+	if !udc.staticNeedsMemory {
+		r.managedDistributeDownloadChunkToWorkers(udc)
```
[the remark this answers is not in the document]
DavidVorick: it adds the chunk to the work queue for each of the workers, which is completely different from the download heap.

Does the documentation do a poor job of distinguishing between the two?

modules/renter/worker.go
DavidVorick: many issues, such as `host has no remaining disk space` only affect uploads.

modules/renter/workerdownload.go
DavidVorick: no, because you might still have overdrive pieces. So even if the download has completed, there may still be memory that's tied up and can't be released.

modules/renter/downloadbuffer.go
DavidVorick: docstring for this function should probably explain how that works, @ChrisSchinnerl was very baffled the first time he saw it as well

modules/renter/downloadheap.go
DavidVorick: Wait what? I thought `time.After` was a pretty safe function.

modules/renter/downloadheap.go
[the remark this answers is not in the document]
ChrisSchinnerl: I think the difference between the heap and the queue is pretty clear but I am not sure if it is wise to call `managedDistributeDownloadChunkToWorkers` within `managedAddChunkToDownloadHeap`. I don't know if I would expect a heap function to add a chunk to the queue.

modules/renter/downloadchunk.go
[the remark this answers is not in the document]
ChrisSchinnerl: That would be great. It feels like there are dozens of TODOs and some of them are low-hanging fruits that could easily be tackled by the community. It's way too easy to lose track of them if there is no central organized list of TODOs because they are scattered over hundreds of files.
I wouldn't mind having the TODOs in the code as long as we also have an issue linking to that todo.

modules/renter/downloadheap.go
ChrisSchinnerl: How do you leak a timer? I also think this function is only called within a single thread.

modules/renter/downloadbuffer.go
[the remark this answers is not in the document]
ChrisSchinnerl: Ahh yeah I found the Note. Might be a bit tricky to find all the other chunks and make them fail or remove them from the heap after the first one failed though.

modules/renter/download.go
```diff
@@ -115,6 +115,14 @@ package renter
...
 // nearly as clean-cut as some of the other potential extensions that we can do.
 
+// TODO: Right now the whole download will build and send off chunks even if
+// there are not enough hosts to download the file, and even if there are not
+// enough hosts to download a particular chunk. For the downloads and chunks
+// which are doomed from the outset, we can skip some computation by checking
+// and failing earlier. Another optimization we can make is to not count a
+// worker for a chunk if the worker's contract does not appear in the chunk
+// heap.
+
```
ChrisSchinnerl: Should that be part of this PR? Feels like a step back to not have it from the start. Otherwise we waste money on bandwidth.

modules/renter/downloadheap.go
lukechampine: `time.After` creates a timer that only gets GC'd once the duration elapses. So it "leaks" if you're creating timers faster than they're expiring (on average).

In this case, where the only other `case` is a shutdown signal, I think it's okay. It's only a problem in code like:
```go
for {
    success := make(chan struct{})
    someAsyncFunction(success)
    select {
    case <-success:
        continue
    case <-time.After(longTimeoutDuration):
        return errors.New("timed out")
    }
}
```
because then there's a good chance that you'll be creating new `time.After`s faster than the old `time.After`s expire.

modules/renter.go
```diff
@@ -55,19 +55,18 @@ type ContractUtility struct {
...
-	WriteAt(b []byte, off int64) (int, error)
-	Destination() string
-	Close() error
+	Destination     string `json:"destination"`     // The destination of the download.
+	DestinationType string `json:"destinationtype"` // Can be "file", "memory buffer", or "http stream".
+	Length          uint64 `json:"length"`          // The length requested for the download.
+	Offset          uint64 `json:"offset"`          // The offset within the siafile requested for the download.
+	SiaPath         string `json:"siapath"`         // The siapath of the file used for the download.
+
+	Completed           bool      `json:"completed"`           // Whether or not the download has completed.
```
ChrisSchinnerl: Is this ever set? I can't find it. If it is, we should probably make sure it is set after we wrote the file to the writer to make sure it is actually on disk if the flag is set to true.

node/api/renter.go
```diff
@@ -138,12 +138,19 @@ type (
...
 	DownloadInfo struct {
-		SiaPath     string    `json:"siapath"`
-		Destination string    `json:"destination"`
-		Filesize    uint64    `json:"filesize"`
-		Received    uint64    `json:"received"`
-		StartTime   time.Time `json:"starttime"`
-		Error       string    `json:"error"`
+		Destination     string `json:"destination"`     // The destination of the download.
+		DestinationType string `json:"destinationtype"` // Can be "file", "memory buffer", or "http stream".
+		Filesize        uint64 `json:"filesize"`        // DEPRECATED. Same as 'Length'.
```
ChrisSchinnerl: Should we rename `Filesize` to `V131Filesize` as an indicator that this field only exists for compatibility reasons and is deprecated?

modules/renter/download.go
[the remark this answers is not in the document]
DavidVorick: I added a TODO because the change ended up being a little invasive, and I'd rather get this merged than have everything perfect.

modules/renter/downloadheap.go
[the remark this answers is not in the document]
DavidVorick: It's functionally the same as it being added to the heap, and then the heap immediately adding it to the worker queue. The heap adds it to the worker queue at some point anyway.

modules/renter.go
DavidVorick: We can document it when we expose it via the api.

modules/renter/download.go
DavidVorick: The workers check whether a chunk is viable before they start downloading. Also, the general operating assumption is that all files are able to be downloaded.

modules/renter.go
DavidVorick: download.go, line 475

We don't set it until after the file has been fully written to the writer.

DavidVorick: or rather, I should say that it won't be set to true until after the file has been fully written to the writer, and the writer closed.

node/api/renter.go
DavidVorick: No, because if we change the name, we break compatibility. This field is exposed in the API, to preserve compatibility with existing apps we need to make sure they read the expected/correct value when they read the 'Filesize' field.

modules/renter/workerdownload.go
```diff
@@ -31,7 +31,11 @@ func (udc *unfinishedDownloadChunk) cleanUp() {
 	desiredPiecesRegistered := udc.erasureCode.MinPieces() + udc.staticOverdrive - udc.piecesCompleted
 	if !chunkComplete && udc.piecesRegistered < desiredPiecesRegistered {
 		for i := 0; i < len(udc.workersStandby); i++ {
-			udc.workersStandby[i].managedQueueDownloadChunk(udc)
+			// Convention mandates that the udc be unlocked when passed in as an
+			// argument (and in this case, failing to do so causes a deadlock).
+			// Since we can't unlock the udc here, we'll have to use a goroutine
+			// to make the call to prevent a deadlock.
+			go udc.workersStandby[i].managedQueueDownloadChunk(udc)
```
ChrisSchinnerl: Couldn't we copy the `udc.workersStandby` while holding the lock of `udc` and then unlock the udc lock and iterate over the copied slice? That way we don't have to use a goroutine since we don't hold the udc lock anymore. `cleanUp` would probably have to be changed to `managedCleanUp` though.

node/api/renter.go
ChrisSchinnerl: I meant the field in the code, not the json tag.

modules/renter/downloadchunk.go
```diff
@@ -0,0 +1,149 @@
...
+	udc.mu.Lock()
+	udc.recoveryComplete = true
+	udc.cleanUp()
+	udc.mu.Unlock()
+
+	// Update the download and signal completion of this chunk.
+	udc.download.mu.Lock()
+	defer udc.download.mu.Unlock()
+	udc.download.chunksRemaining--
+	atomic.AddUint64(&udc.download.atomicDataReceived, udc.staticFetchLength)
```
ChrisSchinnerl: `atomicDataReceived` probably doesn't need to be atomic if we always hold the `download.mu` when we access it.

DavidVorick:
Documentation to explain the new API fields has been added. I haven't tested this on the live network yet, but I think it's ready to merge. I know that it's bottlenecking a few things that Chris was working on.

I also haven't written tests around the new API fields. I would like to do this using the `siatest` framework that Chris wrote for downloads, but that can't be merged until this is merged, due to some interface changes. I will write tests for the new API fields as soon as that is merged.

modules/renter/renter.go
```diff
@@ -155,36 +144,41 @@ type trackedFile struct {
...
-	// Work management.
-	//
-	// chunkQueue contains a list of incomplete work that the download loop acts
-	// upon. The chunkQueue is only ever modified by the main download loop
-	// thread, which means it can be accessed and updated without locks.
+	// Download management. The heap has a separate mutex because it is always
+	// accessed in isolation.
+	downloadHeapMu sync.Mutex         // Used to protect the downloadHeap.
+	downloadHeap   *downloadChunkHeap // A heap of priority-sorted chunks to download.
+	newDownloads   chan struct{}      // Used to notify download heap thread that new downlaods are available.
```
lukechampine: with a bit of work, `newDownloads` could be moved inside the `downloadChunkHeap` type, and you could give `downloadChunkHeap` a method for signaling new downloads. Then the implementation details around signal handling are hidden.

modules/renter/download.go
```diff
@@ -1,805 +1,505 @@
...
+// high and computationally efficient parallelism on downloads. When a download
+// is requested, it gets split into its respective chunks (which are downloaded
+// individually) and then put into the download heap. The primary purpose of the
+// download heap is to keep downloads on standby until there is enough memory
+// available to send the downloads off to the workers. The heap is sorted first
+// by priority, but then a few other criteria as well.
+//
+// Some downloads, in particualr downloads issued by the repair code, have
+// already had their memory allocated. These downloads get to skip the heap and
+// go straight for the workers.
```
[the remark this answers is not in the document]
DavidVorick: no, if they have to wait in the heap they can cause a deadlock, because their corresponding upload memory cannot be released, and they may get stuck waiting for another download to acquire memory.

This is effectively a priority lane. The purpose of the heap is exclusively to ratelimit the act of giving downloads to workers. By giving the pieces directly to the workers instead of putting them through the heap, this is functionally equivalent to putting them at the top of the heap and triggering the heap immediately.

modules/renter/download.go
```diff
@@ -1,805 +1,505 @@
...
+		destinationType   string              // "file", "buffer", "http stream", etc.
+		destinationString string              // The string to report to the user for the destination.
+		file              *file               // The file to download.
+
+		latencyTarget uint64 // Workers above this latency will be automatically put on standby initially.
+		length        uint64 // Length of download. Cannot be 0.
+		needsMemory   bool   // Whether new memory needs to be allocated to perform the download.
+		offset        uint64 // Offset within the file to start the download. Must be less than the total filesize.
+		overdrive     int    // How many extra pieces to download to prevent slow hosts from being a bottleneck.
+		priority      uint64 // Files with a higher priority will be downloaded first.
```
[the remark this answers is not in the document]
DavidVorick: discreet. they'll be whatever the user sets them to, has to be the same for the whole download. Downloads for repair have priority 0. We may want to change that so that the user can choose to make their personal downloads less important than repairing, not sure.

modules/renter/download.go
```diff
@@ -1,805 +1,505 @@
...
-
-// fail will mark the download as complete, but with the provided error.
-func (d *download) fail(err error) {
-	if d.downloadComplete {
-		// Either the download has already succeeded or failed, nothing to do.
+	// If the download is already complete, extend the error.
+	complete := d.staticComplete()
+	if complete && d.err != nil {
+		d.err = errors.Compose(d.err, err)
+		return
```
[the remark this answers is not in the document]
DavidVorick: I was thinking that it's better to have more information about which chunks failed. I guess it's not that useful if you know 14 chunks failed instead of 1.

modules/renter/download.go
```diff
@@ -1,805 +1,505 @@
...
-			case worker.priorityDownloadChan <- dw:
-			default:
-				r.log.Critical("Download work not immediately received by worker")
-			}
-			continue loop
-		}
+	// Queue the downloads for each chunk.
+	writeOffset := int64(0) // where to write a chunk within the download destination.
+	for i := minChunk; i <= maxChunk; i++ {
+		d.chunksRemaining++
```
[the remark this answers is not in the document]
DavidVorick: If we change the loop such that chunks can fail, or something to that effect, it may impact how this variable is used. I think better to increase the counter as you add chunks.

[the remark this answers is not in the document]
DavidVorick: oh actually, this could cause errors. If the download only has 2 chunks and those 2 chunks finish before it queues the third, it's going to think the download is complete. hmm. Guess it does need to be outside the loop.

modules/renter/download.go
```diff
@@ -1,805 +1,505 @@
...
-		if ds.activePieces+nextChunk.download.erasureCode.MinPieces() > maxActiveDownloadPieces {
-			// There is a limited amount of RAM available, and scheduling the
-			// next piece would consume too much RAM.
-			return
+		// Set the fetchLength - the number of bytes to fetch within the chunk
+		// that we start downloading from.
+		if i == maxChunk && (params.length+params.offset)%params.file.staticChunkSize() != 0 {
+			udc.staticFetchLength = ((params.length + params.offset) % params.file.staticChunkSize()) - udc.staticFetchOffset
+		} else {
+			udc.staticFetchLength = params.file.staticChunkSize() - udc.staticFetchOffset
```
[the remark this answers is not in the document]
DavidVorick: the offset will always be 0 for the middle chunks, so it doesn't have any impact on the logic. If for some reason its not zero, then you'd want to be sure to subtract it here. I think this is fewer lines of code total.

Overall the offset math is just a big headache.

modules/renter/download.go
```diff
@@ -1,805 +1,505 @@
...
-		nextChunk.download.mu.Lock()
-		downloadComplete := nextChunk.download.downloadComplete
-		nextChunk.download.mu.Unlock()
-		if downloadComplete {
-			// Download has already failed.
-			continue
+		// Set the writeOffset within the destination for where the data should
+		// be written.
+		udc.staticWriteOffset = writeOffset
+		writeOffset += int64(udc.staticFetchLength)
```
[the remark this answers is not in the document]
DavidVorick: I'm not sure how that is handled. The tests all pass though, and I believe that we do have tests which try partial downloads. I just copied the repair code from the old download code and it kept working.

modules/renter/downloadchunk.go
```diff
@@ -0,0 +1,247 @@
...
+
+	// Assemble a list of standby workers, release the udc lock, and then queue
+	// the chunk into the workers. The lock needs to be released early because
+	// holding the udc lock and the worker lock at the same time is a deadlock
+	// risk (they interact with eachother, call functions on eachother).
+	var standbyWorkers []*worker
+	for i := 0; i < len(udc.workersStandby); i++ {
+		standbyWorkers = append(standbyWorkers, udc.workersStandby[i])
+	}
+	udc.workersStandby = udc.workersStandby[:0] // Workers have been taken off of standby.
```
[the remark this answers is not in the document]
DavidVorick: it's still the same underlying memory. If you append to udc.workersStandby somewhere else it's going to overwrite your "copy" and probably cause a worker to get lost, which will over time reduce your available memory to 0 and cause a deadlock.

modules/renter/downloaddestination.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	// The stream has not yet progressed to 'offset'. We will block until the
+	// stream has made progress. We perform the block by creating a
+	// thread-specific mutex 'myMu' and adding it to the object's list of
+	// blocking threads. When other threads successfully call WriteAt, they will
+	// reference this list and unblock any which have enough progress. The
+	// result is a somewhat strange construction where we lock myMu twice in a
+	// row, but between those two calls to lock, we put myMu in a place where
+	// another thread can unlock myMu.
```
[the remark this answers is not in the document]
DavidVorick: if you use a sync.Cond you lose the first-in-first-out property

modules/renter/downloadheap.go
```diff
@@ -0,0 +1,213 @@
...
+// Download jobs are added to the heap via a function call.
+
+import (
+	"container/heap"
+	"errors"
+	"time"
+)
+
+const (
+	defaultFilePerm         = 0666
```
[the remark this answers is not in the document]
DavidVorick: it's a special constant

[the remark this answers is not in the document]
DavidVorick: you wouldn't understand :P

DavidVorick:
Fixed a deadlock that Chris and I found while working on the production version, in addition to addressing many of the comments made by @lukechampine

PR #2759, opened 2018-02-12
Fixing capitalization of storagespending

mtlynch:
storagespelling was spelled with capitalized letters, making the name
inconsistent with every other output of the Sia API.

This change will unfortunately break clients that depend on the incorrect
capitalization.

### DIFF
--- a/node/api/renter.go
+++ b/node/api/renter.go
@@ -98,7 +98,7 @@ type (
 		// Block height that the file contract began on.
 		StartHeight types.BlockHeight `json:"startheight"`
 		// Amount of contract funds that have been spent on storage.
-		StorageSpending types.Currency `json:"StorageSpending"`
+		StorageSpending types.Currency `json:"storagespending"`
 		// Total cost to the wallet of forming the file contract.
 		TotalCost types.Currency `json:"totalcost"`
 		// Amount of contract funds that have been spent on uploads.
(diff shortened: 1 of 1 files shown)


lukechampine:
this is actually an interesting case -- the implementation was wrong, but the API documentation always spelled it `"storagespending"`. Perhaps we can justify fixing the implementation by pointing out that the docs were always correct?

tbenz9:
My vote is we fix it and make a special note in the release notes of version 1.3.2 that this is a breaking API change.

mtlynch:
If we really want to be conservative about back-compat, we can add a new field like:

```
StorageSpending types.Currency `json:"storagespending"`
StorageSpendingLegacy types.Currency `json:"StorageSpending"`
```

And then just copy `StorageSpending` to `StorageSpendingLegacy` right before we return.

DavidVorick:
I am definitely in favor of the legacy / deprecated option. I don't know how many apps we have out there that depend on StorageSpending, but you can imagine that if it suddenly starts reporting '0' for an app that upgraded siad, there could be issues.

Big enough of a concern that I think compatibility is important here, even though the docs always gave the intended usage.

mtlynch:
>I am definitely in favor of the legacy / deprecated option. I don't know how many apps we have out there that depend on StorageSpending, but you can imagine that if it suddenly starts reporting '0' for an app that upgraded siad, there could be issues.

>Big enough of a concern that I think compatibility is important here, even though the docs always gave the intended usage.

Sure, updated.

PR #2761, opened 2018-02-13
Remove consensus.db code that could cause instability

lukechampine:
We don't know for certain that these changes will fix the corruption we've observed, but it seems probable. Removing this code only impacts performance, not safety/correctness, and we can easily restore it if it doesn't have any effect on corruption.


DavidVorick:
I think we determined that switching to bbolt is the best we can do?

PR #2762, opened 2018-02-13
debug contract deadlock

lukechampine:


modules/renter/proto/contractset.go
```diff
@@ -81,9 +90,11 @@ func (cs *ContractSet) Return(c *SafeContract) {
 	safeContract, ok := cs.contracts[c.header.ID()]
 	if !ok {
 		cs.mu.Unlock()
+		safeContract.mu.Unlock()
```
lukechampine: I think this is likely the culprit

PR #2763, opened 2018-02-13
Change contractset lock to safelock to make finding deadlocks easier

ChrisSchinnerl:


:q
```diff
@@ -0,0 +1,9 @@
+[33mcommit 1630a8e7039bbec39c8444dedb9aafef1901cd64[m[33m ([m[1;36mHEAD -> [m[1;32mmaster[m[33m, [m[1;31morigin/master[m[33m, [m[1;31morigin/HEAD[m[33m)[m
+Merge: 21b1a98b ffafd4ad
+Author: Luke Champine <luke.champine@gmail.com>
+Date:   Tue Feb 13 12:29:30 2018 -0500
+
+    Merge pull request #2746 from mikkeljuhl/lint-errors/transactionpool
+    
+    Lint errors: transaction pool
+
```
lukechampine: lol

DavidVorick:
It's not the failure we wanted, but looks like `siatest` failed: "failed to create group: insufficient balance".

ChrisSchinnerl:
Hm.. I thought I fixed the error by changing the funding code. The new siatest PR adds a lot more info to the returned errors though. That should make finding it if it happens again a lot easier. @DavidVorick

ChrisSchinnerl:
I've been restarting this test for 2 days now and it doesn't seem like we can reproduce the deadlock like this. It seems like the little overhead of the safelock prevents the deadlock from happening. 
https://github.com/NebulousLabs/Sia/pull/2767 should be able to fix the deadlock though and also a NDF that occurs every now and then during a testrun.
I am not in favor of merging this PR since I don't think the safelocks should make it into the production code. It's not easy to guess a good timeout duration for them which means they might cause more harm than good.

DavidVorick:
I'm going to close this for now, since you seem to have found a way to trigger a deadlock, and it seems reasonable to me that this is what we were seeing in the build. As you pointed out in that PR, the circumstances are pretty rare, which would explain why it's so hard to reproduce.

We can re-open it if the deadlock errors persist.

PR #2771, opened 2018-02-16
automatically adjust contract pricing

ChrisSchinnerl:
This PR is pretty much the successor of [PR#2070](https://github.com/NebulousLabs/Sia/pull/2070). It's easier and simply uses the transactionpool to figure out the correct fees for the contract price.
It also changes a few parts in the code that previously used the internal settings to use the external ones. I think the internal settings are meant to provide a minimum while the external ones reflect the current price of the renter. Therefore we should probably use the external ones if we do stuff like checking if the renter pays enough for bandwidth.

### DIFF
--- a/modules/host/consts.go
+++ b/modules/host/consts.go
@@ -78,10 +78,10 @@ var (
 	defaultCollateralBudget = types.SiacoinPrecision.Mul64(100e3)
 
 	// defaultContractPrice defines the default price of creating a contract
-	// with the host. The default is set to 0 siacoins. This should be fine
-	// since we use the FeeEstimation function for an estimation which already
-	// has a sane minimum > 0.
-	defaultContractPrice = types.ZeroCurrency // 0 siacoins
+	// with the host. The current default is 0.05. This was chosen since it is
+	// 1/2e6 times 10,000 bytes. 1/2e6 is the current transaction pool's fee
+	// estimation minimum.
+	defaultContractPrice = types.SiacoinPrecision.Div64(200) // 0.05 siacoins
 
 	// defaultDownloadBandwidthPrice defines the default price of upload
 	// bandwidth. The default is set to 10 siacoins per gigabyte, because
--- a/modules/host/consts.go
+++ b/modules/host/consts.go
@@ -78,9 +78,9 @@ var (
 	defaultCollateralBudget = types.SiacoinPrecision.Mul64(100e3)
 
 	// defaultContractPrice defines the default price of creating a contract
-	// with the host. The default is set to 30 siacoins, which the file
-	// contract revision can have 15 siacoins put towards it, and the storage
-	// proof can have 15 siacoins put towards it.
+	// with the host. The default is set to 0 siacoins. This should be fine
+	// since we use the FeeEstimation function for an estimation which already
+	// has a sane minimum > 0.
 	defaultContractPrice = types.ZeroCurrency // 0 siacoins
 
 	// defaultDownloadBandwidthPrice defines the default price of upload
--- a/modules/host/consts.go
+++ b/modules/host/consts.go
@@ -81,7 +81,7 @@ var (
 	// with the host. The default is set to 30 siacoins, which the file
 	// contract revision can have 15 siacoins put towards it, and the storage
 	// proof can have 15 siacoins put towards it.
-	defaultContractPrice = types.SiacoinPrecision.Mul64(3) // 3 siacoins
+	defaultContractPrice = types.ZeroCurrency // 0 siacoins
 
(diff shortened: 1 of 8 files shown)


modules/host/consts.go
```diff
@@ -78,10 +78,10 @@ var (
...
 
 	// defaultContractPrice defines the default price of creating a contract
-	// with the host. The default is set to 30 siacoins, which the file
-	// contract revision can have 15 siacoins put towards it, and the storage
-	// proof can have 15 siacoins put towards it.
-	defaultContractPrice = types.SiacoinPrecision.Mul64(3) // 3 siacoins
+	// with the host. The default is set to 0 siacoins. This should be fine
+	// since we use the FeeEstimation function for an estimation which already
+	// has a sane minimum > 0.
+	defaultContractPrice = types.ZeroCurrency // 0 siacoins
```
DavidVorick: default minimum of 0 will always make me nervous. Would feel better if it was 0.1 or 0.01

ChrisSchinnerl: Could we set it to 0.001? It is redundant due to the minimum in `FeeEstimation` but I understand that it makes you nervous :p 
While we're on it, could we decrease the sane minimum of the `FeeEstimation` by 100x? I feel like the usd price probably increased by 100x since that minimum was introduced. That would cause the actual minimum contract price to be 0.003 SC instead of 0.3 SC which is much more reasonable if you form 50-100 contracts.

modules/transactionpool/consts.go
```diff
@@ -68,7 +68,7 @@ var (
 var (
 	// minEstimation defines a sane minimum fee per byte for transactions.  This
 	// will typically be only suggested as a fee in the absence of congestion.
-	minEstimation = types.SiacoinPrecision.Div64(100).Div64(1e3)
+	minEstimation = types.SiacoinPrecision.Div64(2e6)
```
DavidVorick: At this minimum, it costs 1 siacoin to fill a block, or about 3 cents. Given that we've already had people propose adding things like messaging protocols on top of the Sia blockchain using the arbitrary data, I don't think we want to drop the default fee so low. The previous minimum, 60 cents to fill a block, is already really low.

That means a full set of 50 contracts would cost about 20 cents. That seems fine. Even if the price goes up 10x ($10 billion market cap), we're still only at $2 for a full set of contracts.

modules/host/consts.go
```diff
@@ -78,10 +78,10 @@ var (
...
 
 	// defaultContractPrice defines the default price of creating a contract
-	// with the host. The default is set to 30 siacoins, which the file
-	// contract revision can have 15 siacoins put towards it, and the storage
-	// proof can have 15 siacoins put towards it.
-	defaultContractPrice = types.SiacoinPrecision.Mul64(3) // 3 siacoins
+	// with the host. The current default is 0.05. This was chosen since it is
+	// 1/2e6 times 10,000 bytes. 1/2e6 is the current transaction pool's fee
+	// estimation minimum.
+	defaultContractPrice = types.SiacoinPrecision.Div64(200) // 0.05 siacoins
```
DavidVorick: Going to want to adjust this since the min fee for the tpool is not coming down.

DavidVorick: Though, it's probably fine to have the default contract price minimum be something like half of the lowest value that the tpool will ever spit out.

modules/host/negotiateformcontract.go
```diff
@@ -212,7 +212,8 @@ func (h *Host) managedVerifyNewContract(txnSet []types.Transaction, renterPK cry
 	blockHeight := h.blockHeight
 	lockedStorageCollateral := h.financialMetrics.LockedStorageCollateral
 	publicKey := h.publicKey
-	settings := h.settings
+	eSettings := h.externalSettings()
```
DavidVorick: I think here you want to be using the externalSettings that you provided at the beginning of the RPC, instead of asking for the external settings from the host again, as the value may have changed.

DavidVorick:
This looks alright to me. One thing I want you to double check in the code is the use of the `externalSettings()` function, I think there may be at least 1 misuse in this PR (maybe not though). Because the results returned by `externalSettings()` could change halfway through a negotiation, you want to make sure you use the same value returned from the first call to `externalSettings()` throughout the whole RPC. If another function needs the settings object, it needs to be passed in instead of asking the host for the external settings again.

Other than that, changes to the negotiate protocols always make me nervous. Other than the potentially incorrect call to `externalSettings`, I don't think I saw anything wrong. But mistakes in this code will lose people money. Can you double check that everything seems correct, and that it's identical to the previous code except for the change to how we determine the contract price?

PR #2780, opened 2018-02-19
Simplify download handler and remove hardcoded sleep

ChrisSchinnerl:


lukechampine:
Hmm. I don't like with splitting `Download` and `DownloadAsync`. It seems simpler for `Download` to return a `chan error`, or some other object that allows the caller to query the status of the download.

The main complication is that you want a way to distinguish between "setup errors" (e.g. `destination must be an absolute path`) and "in-progress errors" (e.g. losing internet halfway through the download). Sending both types of error down a `chan error` is one solution, but it makes async downloads ugly on the caller side; you have to do:
```go
errChan := Download()
// check for setup error
select {
case err := <-errChan:
    // handle setup error
default:
}
```
Alternatively, we could return the setup error and errChan separately, but that makes blocking downloads ugly:
```go
errChan, err := Download()
if err != nil {
    // handle setup error
} else if err := <-errChan; err != nil {
    // handle in-progress error
}
```

So perhaps splitting off a convenience function really is the best we can do. Supporting this is the design of Go's `net/rpc` package, which has a blocking form (`Call`) and an async form (`Go`):
```go
func (client *Client) Call(serviceMethod string, args interface{}, reply interface{}) error {
	call := <-client.Go(serviceMethod, args, reply, make(chan *Call, 1)).Done
	return call.Error
}
```

lukechampine:
I suppose there's one more possibility: we could eliminate the distinction between setup and in-progress errors. All errors would be reported asynchronously. I think there could be some merit to that approach.

DavidVorick:
The intention is not to distinguish between them at all to the caller. The reason that we distinguish between them on the backend is because we'll know if there was a setup error almost immediately. No reason to force the caller to poll the download status to figure out if there was a setup error.

modules/renter/download.go
```diff
@@ -233,6 +233,105 @@ func (d *download) Err() (err error) {
 	return err
 }
 
+// Download performs a file download using the passed parameters and blocks
+// until the download is finished.
+func (r *Renter) Download(p modules.RenterDownloadParameters) error {
+	return r.managedDownload(p, false)
```
DavidVorick:
```go
err := r.managedDownload(p)
if err != nil {
    return err
}

select {
    // Block until download is complete
}
return d.Err()
```

That way you don't need the true/false stuff in your function.

modules/renter/download.go
```diff
@@ -233,6 +233,105 @@ func (d *download) Err() (err error) {
...
+	r.downloadHistoryMu.Lock()
+	r.downloadHistory = append(r.downloadHistory, d)
+	r.downloadHistoryMu.Unlock()
+
+	// If async is true we can return right away.
+	if async {
+		return nil
+	}
+	// Otherwise we block until the download has completed.
+	select {
```
DavidVorick: The select can just be moved to the `Download` function

modules/renter/download.go
ChrisSchinnerl: How should we wait for the download without returning a channel?

modules/renter/download.go
```diff
@@ -233,6 +233,109 @@ func (d *download) Err() (err error) {
...
+		osFile, err := os.OpenFile(p.Destination, os.O_CREATE|os.O_WRONLY, os.FileMode(file.mode))
+		if err != nil {
+			return nil, err
+		}
+		dw = osFile
+		destinationType = "file"
+	}
+
+	// Create the download object.
+	d, err := r.newDownload(downloadParams{
```
DavidVorick: this needs to be changed to `r.managedNewDownload`.

PR #2784, opened 2018-02-20
remove all unused assignments

GlenDC:
the fixes made in this commit were detected with a tool called [ineffassign](https://github.com/gordonklaus/ineffassign),
it's not my tool, but I use for some other projects as well,
and it can be helpful these kind of situations,
as they usually indicate a bug or a potential false positive in a unit test

### DIFF
--- a/modules/consensus/difficulty_test.go
+++ b/modules/consensus/difficulty_test.go
@@ -283,12 +283,11 @@ func TestStoreBlockTotals(t *testing.T) {
 	// returned follow a decay. While storing repeatedly, check that the
 	// getBlockTotals values match the values that were stored.
 	err = cs.db.Update(func(tx *bolt.Tx) error {
-		totalTime := int64(0)
-		totalTarget := types.RootDepth
+		var totalTime int64
 		var id types.BlockID
-		parentTimestamp := types.Timestamp(0)
-		currentTimestamp := types.Timestamp(0)
+		var parentTimestamp, currentTimestamp types.Timestamp
 		currentTarget := types.RootTarget
+		totalTarget := types.RootDepth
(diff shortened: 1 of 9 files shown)


modules/consensus/difficulty_test.go
```diff
@@ -285,10 +285,13 @@ func TestStoreBlockTotals(t *testing.T) {
...
-		parentTimestamp := types.Timestamp(0)
-		currentTimestamp := types.Timestamp(0)
-		currentTarget := types.RootTarget
+		var (
+			id              types.BlockID
+			parentTimestamp types.Timestamp
+
+			currentTimestamp = types.Timestamp(0)
+			currentTarget    = types.RootTarget
+		)
```
lukechampine: this is an odd replacement. Looks like maybe it's complaining about `parentTimestamp` being assigned a value, but then that value is never used?

We don't use `var` blocks much, so we should replace this with:
```go
var id types.BlockID
var parentTimestamp, currentTimestamp types.Timestamp
currentTarget := types.RootTarget
```

lukechampine:
thanks! Maybe we'll added `ineffassign` to our CI.

modules/consensus/difficulty_test.go
GlenDC: yeah that could work as well, what's the reason for not using var blocks? To keep diffs as small as possible?

GlenDC:
> thanks! Maybe we'll added ineffassign to our CI.

Yeah you might want to do that, easy to miss these. The tool doesn't catch all of them though, but it does help catching at least some of them. Which is something.

GlenDC:
Force pushed my commit as a response to your code review @lukechampine.

PR #2786, opened 2018-02-21
Added ./sync and ./types to lint packages

mharkus:
1. I need some help in adding comments with some of the exported variables identified by {help}.
2. go fmt removes the tab spaces in variable declarations so right hand side assignments will not be vertically aligned anymore. Let me know if this is ok, otherwise, I will format them back to the original alignment.

for #2702


types/block.go
```diff
@@ -32,7 +32,7 @@ type (
 		Transactions []Transaction   `json:"transactions"`
 	}
 
-	// A BlockHeader, when encoded, is an 80-byte constant size field
+	// BlockHeader when encoded, is an 80-byte constant size field
```
lukechampine: Interesting. `golint` doesn't care about the `A`, but it does care about the `,`. I would rephrase this as:
```
A BlockHeader contains the data that, when hashed, produces the Block's ID.
```

types/block.go
```diff
@@ -42,9 +42,12 @@ type (
 		MerkleRoot crypto.Hash `json:"merkleroot"`
 	}
 
+	// BlockHeight refers to the height of the block
```
lukechampine: more specifically,
```
BlockHeight is the number of blocks between a given block and the Genesis block.
The BlockHeight of the Genesis block is 0.
```

types/block.go
```diff
@@ -42,9 +42,12 @@ type (
...
 	BlockHeight uint64
-	BlockID     crypto.Hash
-	BlockNonce  [8]byte
+	// BlockID is an identifier used to identify the block
+	BlockID crypto.Hash
+	// BlockNonce {help}
+	BlockNonce [8]byte
```
lukechampine:
```
The BlockNonce is a "scratch space" that miners can freely alter to produce
a BlockID that satisfies a given Target.
```

types/block.go
```diff
@@ -42,9 +42,12 @@ type (
 		MerkleRoot crypto.Hash `json:"merkleroot"`
 	}
 
+	// BlockHeight refers to the height of the block
 	BlockHeight uint64
-	BlockID     crypto.Hash
-	BlockNonce  [8]byte
+	// BlockID is an identifier used to identify the block
+	BlockID crypto.Hash
```
lukechampine:
```
A BlockID is the hash of a BlockHeader. A BlockID uniquely
identifies a Block, and indicates the amount of work performed
to mine that Block. The more leading zeros in the BlockID, the
more work was performed.
```

types/constants.go
```diff
@@ -13,42 +13,70 @@ import (
 )
 
 var (
-	BlockFrequency         BlockHeight
-	BlockSizeLimit         = uint64(2e6)
+	// BlockFrequency {help}
+	BlockFrequency BlockHeight
```
lukechampine:
```
BlockFrequency is the desired number of seconds that
should elapse, on average, between successive Blocks.
```

types/constants.go
```diff
@@ -13,42 +13,70 @@ import (
...
+	// BlockSizeLimit {help}
+	BlockSizeLimit = uint64(2e6)
```
lukechampine:
```
BlockSizeLimit is the maximum size of a binary-encoded Block
that is permitted by the consensus rules.
```

types/constants.go
```diff
@@ -13,42 +13,70 @@ import (
...
+	// ExtremeFutureThreshold {help}
 	ExtremeFutureThreshold Timestamp
```
lukechampine:
```
ExtremeFutureThreshold is a temporal limit beyond which Blocks are
discarded by the consensus rules. When incoming Blocks are processed, their
Timestamp is allowed to exceed the processor's current time by a small amount.
But if the Timestamp is further into the future than ExtremeFutureThreshold,
the Block is immediately discarded.
```

types/constants.go
```diff
@@ -13,42 +13,70 @@ import (
...
-	FutureThreshold        Timestamp
-	GenesisBlock           Block
+	// FutureThreshold {help}
+	FutureThreshold Timestamp
```
lukechampine:
```
ExtremeFutureThreshold is a temporal limit beyond which Blocks are
discarded by the consensus rules. When incoming Blocks are processed, their
Timestamp is allowed to exceed the processor's current time by no more than
FutureThreshold. If the excess duration is larger than FutureThreshold, but
smaller than ExtremeFutureThreshold, the Block may be held in memory until
the Block's Timestamp exceeds the current time by less than FutureThreshold.
```

types/constants.go
```diff
@@ -13,42 +13,70 @@ import (
...
+	// GenesisBlock is the first block of the block chain
+	GenesisBlock Block
 
-	// The GenesisID is used in many places. Calculating it once saves lots of
+	// GenesisID is used in many places. Calculating it once saves lots of
```
lukechampine: articles like `The` should be allowed by golint

types/constants.go
```diff
@@ -13,42 +13,70 @@ import (
...
 	// redundant computation.
 	GenesisID BlockID
 
+	// GenesisSiafundAllocation {help}
 	GenesisSiafundAllocation []SiafundOutput
```
lukechampine:
```
GenesisSiafundAllocation is the set of SiafundOutputs created in the Genesis block.
```

types/constants.go
```diff
@@ -13,42 +13,70 @@ import (
...
-	InitialCoinbase          = uint64(300e3)
-	MaturityDelay            BlockHeight
-	MaxAdjustmentDown        *big.Rat
-	MaxAdjustmentUp          *big.Rat
-	MedianTimestampWindow    = uint64(11)
-	MinimumCoinbase          uint64
+	// GenesisTimestamp is the timestamp when genesis block was mined
+	GenesisTimestamp Timestamp
+	// InitialCoinbase {help}
+	InitialCoinbase = uint64(300e3)
```
lukechampine:
```
InitialCoinbase is the coinbase reward of the Genesis block.
```

types/constants.go
```diff
@@ -13,42 +13,70 @@ import (
...
+	RootDepth = Target{255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255}
+	// RootTarget {help}
+	RootTarget Target
+	// SiacoinPrecision {help}
 	SiacoinPrecision = NewCurrency(new(big.Int).Exp(big.NewInt(10), big.NewInt(24), nil))
-	SiafundCount     = NewCurrency64(10000)
-	SiafundPortion   = big.NewRat(39, 1000)
-	TargetWindow     BlockHeight
+	// SiafundCount {help}
+	SiafundCount = NewCurrency64(10000)
```
lukechampine:
```
SiafundCount is the total number of Siafunds in existence.
```

types/constants.go
```diff
@@ -13,42 +13,70 @@ import (
...
+	// SiafundPortion {help}
+	SiafundPortion = big.NewRat(39, 1000)
```
lukechampine:
```
SiafundPortion is the percentage of siacoins that is taxed from FileContracts.
```

types/filecontracts.go
```diff
@@ -90,6 +92,7 @@ type (
 		HashSet  []crypto.Hash            `json:"hashset"`
 	}
 
+	// ProofStatus {help}
 	ProofStatus bool
```
lukechampine:
```
ProofStatus indicates whether a StorageProof was valid (true) or missed (false).
```

types/filecontracts.go
```diff
@@ -9,8 +9,10 @@ import (
 )
 
 var (
+	// ProofMissed {help}
 	ProofMissed ProofStatus = false
```
lukechampine:
```
ProofMissed indicates that a StorageProof was missed, which means that
no valid proof was submitted within the proof window.
```

types/filecontracts.go
```diff
@@ -9,8 +9,10 @@ import (
...
-	ProofValid  ProofStatus = true
+	// ProofValid {help}
+	ProofValid ProofStatus = true
```
lukechampine:
```
ProofValid indicates that a valid StorageProof was submitted within the proof window.
```

types/target.go
```diff
@@ -43,16 +45,16 @@ func (x Target) Cmp(y Target) int {
 }
 
 // Difficulty returns the difficulty associated with a given target.
-func (t Target) Difficulty() Currency {
-	if t == (Target{}) {
+func (x Target) Difficulty() Currency {
+	if x == (Target{}) {
```
lukechampine: these should use `t` throughout instead of `x`

types/currency.go
```diff
@@ -158,8 +158,8 @@ func (x Currency) RoundDown(y Currency) (c Currency) {
 }
 
 // IsZero returns true if the value is 0, false otherwise.
-func (c Currency) IsZero() bool {
-	return c.i.Sign() <= 0
+func (x Currency) IsZero() bool {
```
lukechampine: these should use `c` throughout instead of `x`

types/timestamp.go
```diff
@@ -8,7 +8,10 @@ import (
 )
 
 type (
-	Timestamp      uint64
+	// Timestamp is the digital record of the time of occurrence of a particular
+	// event
+	Timestamp uint64
```
lukechampine:
```
Timestamp is a Unix timestamp, i.e. the number of the seconds since Jan 1 1970.
```

types/transactions.go
```diff
@@ -14,6 +14,7 @@ import (
 )
 
 const (
+	// SpecifierLen {help}
 	SpecifierLen = 16
```
lukechampine:
```
SpecifierLen is the length in bytes of a Specifier.
```

types/transactions.go
```diff
@@ -59,11 +60,20 @@ type (
...
+	// SiafundOutputID refers to the identification used to identify a siafund
+	// output
 	SiafundOutputID crypto.Hash
-	FileContractID  crypto.Hash
-	OutputID        crypto.Hash
+	// FileContractID refers to the identification used to identify a file
+	// contract
+	FileContractID crypto.Hash
+	// OutputID refers to the identification used to identify an output
+	OutputID crypto.Hash
```
lukechampine: For these, instead of `ThingID refers to the identification used to identify a Thing`, I would say `A ThingID uniquely identifies a Thing`.

types/constants.go
```diff
@@ -13,42 +13,70 @@ import (
...
+	// MaturityDelay {help}
+	MaturityDelay BlockHeight
+	// MaxAdjustmentDown {help}
+	MaxAdjustmentDown *big.Rat
+	// MaxAdjustmentUp {help}
+	MaxAdjustmentUp *big.Rat
+	// MedianTimestampWindow {help}
+	MedianTimestampWindow = uint64(11)
+	// MinimumCoinbase {help}
+	MinimumCoinbase uint64
```
lukechampine:
```
MinimumCoinbase is the minimum coinbase reward for a block.
The coinbase decreases in each block after the Genesis block,
but it will not decrease past MinimumCoinbase.
```

types/target.go
mharkus: So for functions like this:

func (x Target) SubtractDifficulties(y Target) (t Target)

this needs to be changed to:

func (t Target) SubtractDifficulties(y Target) (x Target)

but does this also mean that callers of this function have to change as well now that the return variable changes? or that doesn't really matter?

It's complaining about duplicate argument if I leave them both as t.

types/constants.go
mharkus: golint follows a certain format for comments and so it fails with this message when linting:

types/constants.go:38:2: comment on exported var GenesisID should be of the form "GenesisID ..."

mharkus: apparently, constants are not allowed to have leading articles: https://github.com/golang/lint/issues/11. That's why it worked on some comments, for example: in block.go. BlockHeader was defined as type whereas in constants.go, GenesisBlock was defined as var.

mharkus:
@lukechampine I tried to do this:

        // Other consensus variables
	MaturityDelay         BlockHeight
	MaxAdjustmentDown     *big.Rat
	MaxAdjustmentUp       *big.Rat
	MedianTimestampWindow = uint64(11)

but still getting this when linting:

types/constants.go:49:2: comment on exported var MaturityDelay should be of the form "MaturityDelay ..."
types/constants.go:51:2: exported var MaxAdjustmentDown should have comment or be unexported

Should we just add TODO to those vars that cannot be commented on right now?

Something like:

// OakDecayDenom TODO: give description
OakDecayDenom int64

types/constants.go
lukechampine: huh. That's strange, but good to know.

types/target.go
lukechampine: golint just wants the name of the receiving variable to be consistent across methods. So it should be `func (t Target) Foo` for every method. Callers of the function shouldn't be affected.

lukechampine:
drat. Okay, I'll go through and come up with descriptions for the rest.

Makefile
```diff
@@ -52,9 +52,10 @@ vet: release-std
 
 # will always run on some packages for a while.
 lintpkgs = ./build ./cmd/siac ./cmd/siad ./compatibility ./crypto ./encoding ./modules ./modules/consensus ./modules/explorer \
-           ./modules/gateway ./modules/host ./modules/miner ./modules/host/contractmanager ./modules/renter ./modules/renter/contractor ./modules/renter/hostdb \
-           ./modules/renter/hostdb/hosttree ./modules/renter/proto ./modules/wallet ./modules/transactionpool ./node ./node/api ./node/api/server ./persist \
-           ./siatest ./siatest/consensus ./siatest/renter
+           ./modules/gateway ./modules/host ./modules/miner ./modules/host/contractmanager ./modules/renter 				          \
+		       ./modules/renter/contractor ./modules/renter/hostdb ./modules/renter/hostdb/hosttree ./modules/renter/proto        \
+		       ./modules/wallet ./modules/transactionpool ./node ./node/api ./node/api/server ./persist ./siatest                 \
```
DavidVorick: tabs vs. spaces issue here

types/constants.go
```diff
@@ -13,42 +13,70 @@ import (
...
-	MaxAdjustmentDown        *big.Rat
-	MaxAdjustmentUp          *big.Rat
-	MedianTimestampWindow    = uint64(11)
-	MinimumCoinbase          uint64
+	// GenesisTimestamp is the timestamp when genesis block was mined
+	GenesisTimestamp Timestamp
+	// InitialCoinbase {help}
+	InitialCoinbase = uint64(300e3)
+	// MaturityDelay {help}
+	MaturityDelay BlockHeight
```
DavidVorick: MaturityDelay specifies the number of blocks that a maturity-required output is required to be on hold before it can be spent on the blockchain. Outputs are maturity-required if they are highly likely to be altered or invalidated in the event of a small reorg. One example is the block reward, as a small reorg may invalidate the block reward. Another example is a siafund payout, as a tiny reorg may change the value of the payout, and thus invalidate any transactions spending the payout. File contract payouts also are subject to a maturity delay.

types/constants.go
```diff
@@ -13,42 +13,70 @@ import (
...
+	// MaxAdjustmentDown {help}
```
DavidVorick: This should probably be changed to `MaxTargetAdjustmentDown`, and the other one `MaxTargetAdjustmentUp`.

These values restrict how much the block difficulty is allowed to change in a single step, which is important to limit the effect of difficulty raising and lowering attacks.

types/constants.go
```diff
@@ -13,42 +13,70 @@ import (
...
+	MaxAdjustmentDown *big.Rat
+	// MaxAdjustmentUp {help}
+	MaxAdjustmentUp *big.Rat
+	// MedianTimestampWindow {help}
+	MedianTimestampWindow = uint64(11)
```
DavidVorick: `MedianTimestampWindow` tells us how many blocks to look back when calculating the median timestamp over the previous `n` blocks. The timestamp of a block is not allowed to be less than or equal to the median timestamp of the previous `n` blocks, where for Sia this number is typically 11.

types/constants.go
```diff
@@ -13,42 +13,70 @@ import (
...
 
 	// Oak hardfork constants. Oak is the name of the difficulty algorithm for
 	// Sia following a hardfork at block 135e3.
-	OakDecayDenom           int64
-	OakDecayNum             int64
-	OakHardforkBlock        BlockHeight
-	OakHardforkFixBlock     BlockHeight
+
+	// OakDecayDenom {help}
+	OakDecayDenom int64
```
DavidVorick: The denominator for how much the total timestamp is decayed each step.

types/constants.go
```diff
@@ -13,42 +13,70 @@ import (
...
+	// OakDecayNum {help}
+	OakDecayNum int64
```
DavidVorick: The numerator for how much the total timestamp is decayed each step.

types/constants.go
```diff
@@ -13,42 +13,70 @@ import (
...
+	// OakHardforkBlock {help}
+	OakHardforkBlock BlockHeight
```
DavidVorick: The height at which the hardfork to switch to the oak difficulty adjustment algorithm is triggered.

types/constants.go
```diff
@@ -13,42 +13,70 @@ import (
...
+	// OakHardforkFixBlock {help}
+	OakHardforkFixBlock BlockHeight
```
DavidVorick: The height at which the hardfork to switch from the broken oak difficulty adjustment algorithm to the fixed oak difficulty adjustment algorithm is triggered.

types/constants.go
```diff
@@ -13,42 +13,70 @@ import (
...
+	// OakHardforkTxnSizeLimit {help}
 	OakHardforkTxnSizeLimit = uint64(64e3) // 64 KB
```
DavidVorick: The maximum size allowed for a transaction, a change which I believe was implemented simultaneously with the oak hardfork.

types/constants.go
```diff
@@ -13,42 +13,70 @@ import (
...
+	// OakMaxRise {help}
+	OakMaxRise *big.Rat
 
-	RootDepth        = Target{255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255}
-	RootTarget       Target
+	// RootDepth {help}
+	RootDepth = Target{255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255}
+	// RootTarget {help}
+	RootTarget Target
+	// SiacoinPrecision {help}
```
DavidVorick: The number of base units in a siacoin. The Sia network has a very large number of base units. We call 10^24 of these a siacoin.

The base unit for Bitcoin is called a satoshi. We call 10^8 satoshis a bitcoin, even though the code itself only ever works with satoshis.

types/constants.go
```diff
@@ -13,42 +13,70 @@ import (
...
+	// OakMaxDrop {help}
+	OakMaxDrop *big.Rat
+	// OakMaxRise {help}
+	OakMaxRise *big.Rat
 
-	RootDepth        = Target{255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255}
-	RootTarget       Target
+	// RootDepth {help}
+	RootDepth = Target{255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255}
+	// RootTarget {help}
```
DavidVorick: The root target is the target for the genesis block - basically how much work needs to be done in order to mine the first block. The difficulty adjustment algorithm takes over from there.

types/constants.go
```diff
@@ -13,42 +13,70 @@ import (
...
+	OakMaxBlockShift int64
+	// OakMaxDrop {help}
+	OakMaxDrop *big.Rat
+	// OakMaxRise {help}
+	OakMaxRise *big.Rat
 
-	RootDepth        = Target{255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255}
-	RootTarget       Target
+	// RootDepth {help}
+	RootDepth = Target{255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255}
```
DavidVorick: The depth is the cumulative target of all blocks. The root depth is essentially the maximum possible target, there have been no blocks yet, so there is no cumulated difficulty yet.

types/constants.go
```diff
@@ -13,42 +13,70 @@ import (
...
+	// OakHardforkTxnSizeLimit {help}
 	OakHardforkTxnSizeLimit = uint64(64e3) // 64 KB
-	OakMaxBlockShift        int64
-	OakMaxDrop              *big.Rat
-	OakMaxRise              *big.Rat
+	// OakMaxBlockShift {help}
+	OakMaxBlockShift int64
+	// OakMaxDrop {help}
+	OakMaxDrop *big.Rat
+	// OakMaxRise {help}
```
DavidVorick: The shift is the maximum number of seconds that the oak algorithm will shift the difficulty.

The drop is the maximum amount that the difficulty will drop each block, and the rise is the maximum amount that the difficulty will rise each block.

DavidVorick:
@mharkus I think we've filled them all out at this point, let us know if we missed anything

DavidVorick:
*or if anything needs additional clarification.

Sorry for closing this PR, hit the wrong button by mistake.

mharkus:
Ok thanks! I will put all the comments above and update the PR.

tbenz9:
@mharkus bump.  :)

mharkus:
Hi @DavidVorick The TargetWindow BlockHeight is still missing a comment. Also, I will create a separate PR for the renaming of MaxAdjustmentDown and MaxAdjustmentUp.

types/constants.go
```diff
@@ -13,42 +13,113 @@ import (
...
 	SiacoinPrecision = NewCurrency(new(big.Int).Exp(big.NewInt(10), big.NewInt(24), nil))
-	SiafundCount     = NewCurrency64(10000)
-	SiafundPortion   = big.NewRat(39, 1000)
-	TargetWindow     BlockHeight
+	// SiafundCount is the total number of Siafunds in existence.
+	SiafundCount = NewCurrency64(10000)
+	// SiafundPortion is the percentage of siacoins that is taxed from FileContracts.
+	SiafundPortion = big.NewRat(39, 1000)
+	// TargetWindow {help}
+	TargetWindow BlockHeight
```
DavidVorick: TargetWindow is the number of blocks to look backwards when determining how much time has passed vs. how many blocks have been created. It's only used in the old, broken difficulty adjustment algorithm.

types/block.go
```diff
@@ -32,19 +32,25 @@ type (
...
+	// A BlockHeader contains the data that, when hashed, produces the Block's ID.
 	BlockHeader struct {
 		ParentID   BlockID     `json:"parentid"`
 		Nonce      BlockNonce  `json:"nonce"`
 		Timestamp  Timestamp   `json:"timestamp"`
 		MerkleRoot crypto.Hash `json:"merkleroot"`
 	}
 
+	// BlockHeight is the number of blocks between a given block and the Genesis block.
+	// The BlockHeight of the Genesis block is 0.
```
DavidVorick: BlockHeight is the number of blocks that exist after the genesis block.

DavidVorick:
Added a description for TargetWindow. Looks like BlockHeight already had a description, but I clarified it.

Thanks for taking the time to clean up this last, monster package.

PR #2791, opened 2018-02-23
Add a human-readable timestamp to siac wallet transactions output

mtlynch:
Add a column to the output of siac wallet transactions to show a human-readable timestamp. siac currently only outputs a block height for each transactions, but it's difficult for users to translate block heights to real world timestamps.

### DIFF
--- a/cmd/siac/walletcmd.go
+++ b/cmd/siac/walletcmd.go
@@ -6,6 +6,7 @@ import (
 	"math/big"
 	"os"
 	"syscall"
+	"time"
 
 	"github.com/spf13/cobra"
 	"golang.org/x/crypto/ssh/terminal"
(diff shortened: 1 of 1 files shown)


cmd/siac/walletcmd.go
```diff
@@ -505,6 +509,11 @@ func wallettransactionscmd() {
 		outgoingSiacoinsFloat, _ := new(big.Rat).SetFrac(outgoingSiacoins.Big(), types.SiacoinPrecision.Big()).Float64()
 
 		// Print the results.
+		if uint64(txn.ConfirmationTimestamp) != unconfirmedTransactionTimestamp {
+			fmt.Printf(time.Unix(int64(txn.ConfirmationTimestamp), 0).Format("2006-01-02 15:04:05-0700"))
```
lukechampine: `time.Stamp` may be a more human-friendly format. It doesn't include the year, but that might be ok

mtlynch: My use case is figuring out my transactions for tax purposes. Maybe not the common case, but also I don't think a terribly unusual use case.

For reconciling financial transactions, the year and time zone are pretty important. `time.Stamp`'s time zone is only an implicit one, but I'd like to avoid the ambiguity.

PR #2794, opened 2018-02-25
add SIP: store renter metadata remotely

starius:


doc/sips/remote-metadata/README.md
```diff
@@ -0,0 +1,310 @@
...
+  * parity
+  * free
+
+A free sector does not hold usable data. When it was not free (i.e.
+was used for something) it was encrypted and that data is still there.
+
+Sectors are encrypted with a stream cipher with random access (i.e.
+which can produce bytes for XOR in the arbitrary location of the
+sequence). One contract corresponds to one sequence (i.e. one
+encryption key). The cipher has large space for its pseudorandom
```
lukechampine: Encrypting all of the data in a contract with the same key causes problems for file-sharing. A contract can contain data from multiple files, so sharing the decryption key for one file would mean sharing the key for all other files.

doc/sips/remote-metadata/README.md
```diff
@@ -0,0 +1,310 @@
...
+in the sequence. The exact location is determined by hashing metadata
+of the sector. For a metadata sectors metadata is the number of that
+sector. For data and backup sectors normalized versions of their
+records can be used for that purpose.
+
+Metadata is written by appending records to a metadata sector.
+The unused end of a metadata sector cannot be encrypted with the same
+key (i.e. just have the pseudorandom sequence without any XORed data)
+because otherwise, the host can decrypt the subsequent records by
+XORing them with what was in the affected bytes before. To avoid this
```
lukechampine: You may want to look into XTS encryption. It uses a "tweakable block cipher" to enable an interesting property: you can overwrite the sector with new encrypted data without compromising security. The only downside is that it's slower.

tbenz9:
I think we were going to attempt to follow the Bitcoin/bips format.  In which case, this sip should be renamed to `sip-0001` and put in a folder with the same name.  Then a table will be created in the `doc/sips/README.md` file that maps `sip-0001` to `remote-metadata`.

Bitcoin/bips example: https://github.com/bitcoin/bips

starius:
@tbenz9 I renamed the directory to sip-0001-remote-metadata. I think short description is useful as a part of directory name, like human readable names of git branches.

doc/sips/README.md
```diff
@@ -3,3 +3,10 @@ People wishing to submit Sia Improvement Proposals (SIPs) should first propose t
 Having a SIP here does not make it a formally accepted standard until its status becomes Final or Active.
 
 Those proposing changes should consider that ultimately consent may rest with the consensus of the Sia users.
+
+| Number  | Layer   | Title                   | Owner     | Status  |
+| :-----: | :-----: | :---------------------: | :-------: | :-----: |
+| [1][1]  | renter  | Store renter metadata remotely | [@starius][starius] | draft |
```
DavidVorick: please adjust readme spacing to have aligned columns

starius: Done

doc/sips/remote-metadata/README.md
starius: I overlooked file sharing.

We should use a separate key for each file. Metadata will be encrypted with a per-host key. The only thing to take into account is that if the file is shared (with the key) and somebody knows the list of IDs of hosts involved in storing the contracts (e.g. hosts themselves), then they can decrypt the previous data from the file. It can be not OK for some users who want to publish the final version of a file but keep preliminary versions secretly.

More issues are added by parity sectors. If all data sectors in a group are from the same file, then parity sectors can be encrypted with the same key. But if data sectors of a group originate from multiple files, we have to make sure we don't leak other files when the user shares just one file involved in the group. One idea is to add parity sectors after encrypting the data, so people who obtain the shared file can only decrypt an encrypted version of data from other files. What do you think?

DavidVorick:
Took the time to review this in full. A lot of this looks good, though some of the proposals around the file format, and other suggested changes may conflict with other stuff that we are working on.

I think the biggest strength of this document is highlighting all of the requirements and blocking factors for doing seed based file and contract recovery. Each of the individual proposed solutions though should probably their own SIP, as I think we could go back and forth with each of them a bunch of times, and I wouldn't want those conversations to dilute this document.

PR #2796, opened 2018-02-26
Siatest additional download tests

ChrisSchinnerl:
Changes so far:
- DownloadInfo sanity checks fields of DownloadInfo object implicitly
- Parallel download test for large files

### DIFF
--- a/siatest/renter.go
+++ b/siatest/renter.go
@@ -1,17 +1,17 @@
 package siatest
 
 import (
-	"errors"
 	"fmt"
 	"math"
 	"path/filepath"
 	"strconv"
 	"time"
 
-	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/node/api"
+
+	"github.com/NebulousLabs/errors"
 	"github.com/NebulousLabs/fastrand"
 )
 
@@ -20,13 +20,13 @@ import (
 func (tn *TestNode) DownloadToDisk(rf *RemoteFile, async bool) (*LocalFile, error) {
 	fi, err := tn.FileInfo(rf)
 	if err != nil {
-		return nil, build.ExtendErr("failed to retrieve FileInfo", err)
+		return nil, errors.AddContext(err, "failed to retrieve FileInfo")
 	}
 	// Create a random destination for the download
 	fileName := strconv.Itoa(fastrand.Intn(math.MaxInt32))
 	dest := filepath.Join(SiaTestingDir, fileName)
 	if err := tn.RenterDownloadGet(rf.siaPath, dest, 0, fi.Filesize, async); err != nil {
-		return nil, build.ExtendErr("failed to download file", err)
+		return nil, errors.AddContext(err, "failed to download file")
 	}
 	// Create the TestFile
 	lf := &LocalFile{
@@ -39,7 +39,7 @@ func (tn *TestNode) DownloadToDisk(rf *RemoteFile, async bool) (*LocalFile, erro
 	}
 	// Verify checksum if we downloaded the file blocking
 	if err := lf.checkIntegrity(); err != nil {
-		return lf, build.ExtendErr("downloaded file's checksum doesn't match", err)
+		return lf, errors.AddContext(err, "downloaded file's checksum doesn't match")
 	}
 	return lf, nil
 }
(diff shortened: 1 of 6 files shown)


siatest/renter/renter_test.go
```diff
@@ -75,3 +79,49 @@ func testUploadDownload(t *testing.T, tg *siatest.TestGroup) {
...
+	uniqueRemoteFiles := 5
+	// Grab the first of the group's renters
+	renter := tg.Renters()[0]
+
+	// Upload files
+	dataPieces := uint64(len(tg.Hosts())) - 1
+	parityPieces := uint64(1)
+	remoteFiles := make([]*siatest.RemoteFile, 0, uniqueRemoteFiles)
+	for i := 0; i < uniqueRemoteFiles; i++ {
+		remoteFile, err := renter.UploadNewFileBlocking(fileSize, dataPieces, parityPieces)
```
DavidVorick: can we make fileSize a function of `i`? Perhaps `modules.SectorSize + i*modules.SectorSize`?

siatest/renter/renter_test.go
```diff
@@ -75,3 +79,49 @@ func testUploadDownload(t *testing.T, tg *siatest.TestGroup) {
 		t.Error(err)
 	}
 }
+
+// testDownloadMultipleLargeSectors downloads multiple large files (>5 Sectors)
+// in parallel and makes sure that the downloads are blocking each other.
+func testDownloadMultipleLargeSectors(t *testing.T, tg *siatest.TestGroup) {
```
DavidVorick: It looks like this test assumes a ratelimit. What it should probably do is fetch the current ratelimits, check that they are sufficiently low for this test to make sense, and then `t.Error` if not.

DavidVorick:
If you look at the travis build failure, it seems like you introduced a data race

ChrisSchinnerl:
@DavidVorick data race should be fixed in [#2831](https://github.com/NebulousLabs/Sia/pull/2831).

siatest/renter.go
```diff
@@ -59,17 +59,37 @@ func (tn *TestNode) DownloadByStream(rf *RemoteFile) (data []byte, err error) {
 
 // DownloadInfo returns the DownloadInfo struct of a file. If it returns nil,
 // the download has either finished, or was never started in the first place.
-func (tn *TestNode) DownloadInfo(lf *LocalFile, rf *RemoteFile) (*api.DownloadInfo, error) {
+// If the corresponding download info was found, DownloadInfo also performs a
+// few sanity checks on its fields.
+func (tn *TestNode) DownloadInfo(lf *LocalFile, rf *RemoteFile) (di *api.DownloadInfo, err error) {
```
DavidVorick: prefer not to have named variables for long functions, it's much easier for these to get lost or shadowed

siatest/renter.go
```diff
@@ -59,17 +59,37 @@ func (tn *TestNode) DownloadByStream(rf *RemoteFile) (data []byte, err error) {
...
 		if rf.siaPath == d.SiaPath && lf.path == d.Destination {
-			return &d, nil
+			di = &d
+			break
 		}
 	}
-	return nil, nil
+	if di == nil {
+		// No download info found.
+		return
```
DavidVorick: I believe that results in returning 'nil', and 'nil' - shouldn't this return an error (info not found or something)? If you return an error to a function that expects the object to exist, it's going to panic.

siatest/renter.go
```diff
@@ -59,17 +59,37 @@ func (tn *TestNode) DownloadByStream(rf *RemoteFile) (data []byte, err error) {
...
+	}
+	// Check if length and filesize were set correctly
+	if di.Length != di.Filesize {
+		err = errors.New("filesize != length")
```
DavidVorick: you can compose an error with nil, the result will be the standalone error. That would make this if block consistent with all the others, which means you are less likely to make mistakes if you end up shuffling code around later

siatest/renter.go
```diff
@@ -59,17 +59,37 @@ func (tn *TestNode) DownloadByStream(rf *RemoteFile) (data []byte, err error) {
...
+	}
+	// Received data can't be larger than transfered data
+	if di.Received > di.TotalDataTransferred {
+		err = build.ComposeErrors(err, errors.New("received > TotalDataTransfered"))
```
DavidVorick: the error handling functions from the `build` package are deprecated in favor of the `NebulousLabs/errors` package

siatest/renter/renter_test.go
```diff
@@ -75,3 +80,52 @@ func testUploadDownload(t *testing.T, tg *siatest.TestGroup) {
...
+	// parallelDownloads is the number of downloads that are run in parallel.
+	parallelDownloads := 10
+	// fileSize is the size of the downloaded file.
+	fileSize := int(10*modules.SectorSize) + siatest.Fuzz()
+	// set download limits and reset them after test.
+	ratelimit.SetLimits(int64(fileSize)*2, 0, modules.SectorSize)
+	defer ratelimit.SetLimits(0, 0, 0)
```
DavidVorick: If we do something like this we will be unable to run this test in parallel with other tests in any way.

I guess in general, tests on a TestGroup should not be run in parallel within that same TestGroup, that could cause a lot of issues. Maybe we should make the ratelimits specific to a TestGroup though?

DavidVorick: Thinking about it a little more...

should ratelimits be specific to each `server`? I think if you are a user running multiple `siad` servers, you may want to ratelimit them independently, even though they may all be running from the same go runtime.

ChrisSchinnerl: Makes sense but this also seems like it would belong into a different PR. That requires changing the ratelimit package itself.

Issue #1554, opened 2017-01-19
[feature] Implement bandwith limiter 


vevarm:
Now when sia got a speed boost I may be useful to have bw limiter. Not prio but sometime in future.

DavidVorick:
I don't think it will be too difficult to add, we may be able to get it into v1.1.1 or v1.1.2.

tbenz9:
Recommendation to devs: label "Feature Request"

PR #2798, opened 2018-02-26
Add bandwidth throttling

ChrisSchinnerl:
I don't know if we want to keep that in the Sia repo. It's not a lot of code so it might as well just be part of Sia. We should find a better place and name for it than `conn/conn.go`.


tbenz9:
This has been asked for many times.  It's being tracked in #1554 (one of our oldest open issues).

conn/conn.go
```diff
@@ -0,0 +1,131 @@
...
+
+// SetReadDeadline calls the underlying connection's SetReadDeadline method.
+func (rlc *RLConnection) SetReadDeadline(t time.Time) error {
+	return rlc.conn.SetReadDeadline(t)
+}
+
+// SetWriteDeadline calls the underlying connection's SetWriteDeadline method.
+func (rlc *RLConnection) SetWriteDeadline(t time.Time) error {
+	return rlc.conn.SetWriteDeadline(t)
+}
```
lukechampine: you can avoid all this boilerplate by embedding the `net.Conn` field. Then you only need to implement `Write`.

conn/conn.go
```diff
@@ -0,0 +1,131 @@
...
+	rlc := &RLConnection{
+		conn: conn,
+	}
+	return rlc, nil
+}
+
+// Read calls the underlying connection's Read method.
+func (rlc *RLConnection) Read(b []byte) (n int, err error) {
+	return rlc.conn.Read(b)
+}
```
lukechampine: no rate limiting on reads?

conn/conn.go
```diff
@@ -0,0 +1,131 @@
...
+		conn             net.Conn  // underlying net.Conn
+		packetSize       uint64    // size of single packet in bytes
+		packetsPerSecond uint64    // packets that are sent per second
+		writtenBytes     uint64    // number of bytes that have been written to the latest packet
+		lastWrite        time.Time // time when last packet was started
+	}
+)
+
+// Dial creates a RateLimitedDialer.
+func Dial(dialer *net.Dialer, network string, address string) (*RLConnection, error) {
```
lukechampine: Probably better to have a `NewRLConn(conn net.Conn) *RLConn` function. That way, you can add rate limiting to connections that already exist, and you can rate-limit anything that satisfies `net.Conn` (which makes mocking easier).

lukechampine: Might also be a good idea for the constructor to take `packetSize` and `packetsPerSecond` args; otherwise you have to make an extra call to `SetLimit` everywhere.

conn/conn.go
```diff
@@ -0,0 +1,131 @@
...
+
+	// Write data one packet at a time.
+	for packet := range packets {
+		written, err := rlc.conn.Write(packet)
+		n += written
+		if err != nil {
+			return 0, err
+		}
+	}
+	return
```
lukechampine: why does this require a goroutine + channels? Couldn't each `packets <- b` be replaced with a direct call to `Write`?

conn/conn.go
```diff
@@ -0,0 +1,131 @@
...
+	return rlc, nil
+}
+
+// Read calls the underlying connection's Read method.
+func (rlc *RLConnection) Read(b []byte) (n int, err error) {
+	return rlc.conn.Read(b)
+}
+
+// SetLimit sets the rate limit of the connection.
+func (rlc *RLConnection) SetLimit(packetSize uint64, packetsPerSecond uint64) {
```
lukechampine: are these the right units to use? What about just `bytesPerSecond`?

conn/conn.go
```diff
@@ -0,0 +1,131 @@
...
+			packets <- b[:remainingBytes]
+			b = b[remainingBytes:]
+		}
+		// Write the remaining data packet by packet.
+		timer := time.NewTimer(time.Until(rlc.lastWrite.Add(time.Second)))
+		for len(b) > 0 {
+			// Wait long enough
+			select {
+			case <-timer.C:
+			}
```
lukechampine: a single-case `select` is equivalent to inlining that case, i.e. `<-timer.C`. And since the timer can't be interrupted, this can actually be replaced with a `time.Sleep`.

Of course, that raises an important question: how do we interrupt these connections in order to facilitate clean shutdown? If we want to do it "right," that probably means overriding the `net.Conn`'s `Close` and `SetDeadline` methods so that they interrupt the sleep.

lukechampine:
I'm inclined to say that rate limiting is non-trivial, and perhaps we should consider using an existing package (most likely https://github.com/juju/ratelimit), or at least copying most of its code.

The only issue I see with juju's `ratelimit` is that it uses `time.Sleep` instead of an interruptable sleep. We'd have to create a new type to satisfy their `Clock` interface, so that the `Sleep` calls could be interrupted.

conn/conn.go
ChrisSchinnerl: Doesn't that overwrite the `Write` function? I'd probably have to manually implement the private `net.conn` struct if I did that.

conn/conn.go
ChrisSchinnerl: You can't just review the hell out of my WIP and expect everything to be implemented :P

ChrisSchinnerl:
@lukechampine I don't know if copying is save since juju uses LGPLv3 instead of MIT.

conn/conn.go
lukechampine: when you embed a type, your type gets all of its methods. If you define a method on your type with the same name, it overwrites the embedded type's method.

ChrisSchinnerl: I know and that's the issue. `net.Conn` is an interface while `net.conn` is the actual implementation. If I embed `net.Conn` and overwrite `Write` I have nothing to write to since the `c.fd` field that is used in the `net.conn` implementation is not present in `net.Conn` since it is only an interface.

ChrisSchinnerl:
@tbenz9 this PR might not fix that. It only adds bandwidth throttling on a connection level and not on a process level. That means you can throttle the bandwidth to/from each host but not of the whole application. That would require some kind of shared state between all connections.

DavidVorick:
@ChrisSchinnerl the original idea is that we would use a Singleton of some sort so that it was shared between all connections

ChrisSchinnerl:
@DavidVorick @lukechampine @tbenz9 I rewrote this PR pretty much 100% and it is now able to throttle connections for the whole process by introducing a global bandwidthManager. (I know the name is bad. I'm open for better suggestions)
The PR is not done completely (needs more tests + 1 Todo) but it should be ready for a first review.

ratelimit/ratelimit.go
```diff
@@ -0,0 +1,362 @@
...
+	BM = &bandwidthManager{
+		atomicWritePPS:   writePPS,
+		atomicReadPPS:    readPPS,
+		atomicPacketSize: packetSize,
+		writeSignal:      make(chan struct{}),
+		readSignal:       make(chan struct{}),
+	}
+	go BM.threadedWriteLoop(cancel)
+	go BM.threadedReadLoop(cancel)
+}
```
lukechampine: if you rename to `init` this function will be run automatically

tbenz9:
Don't forget a docstring for `package ratelimit`.  It'll automatically get pulled into the godoc at https://godoc.org/github.com/NebulousLabs/Sia

DavidVorick:
Too many top level packages. Can we move everything to lib?

lib/build
lib/compatibility
lib/crypto
lib/encoding
lib/persist
lib/sync
lib/types

means our top level becomes:

cmd/
doc/
lib/
modules/
node/
vendor/

even thinking that we could move modules to node/modules

but then we'd have a `/node/modules/renter/hostdb/hosttree` :>

DavidVorick:
I'm also fine with putting all this code in a new repo at `NebulousLabs/ratelimit`, like we've done with the `errors` stuff, the `wal`, and some of our `sync` stuff, this package is sufficiently generic.

ratelimit/ratelimit.go
DavidVorick: probably makes more sense to do it that way.

ChrisSchinnerl:
This will be continued in a separate [here](https://github.com/NebulousLabs/ratelimit).

PR #2803, opened 2018-02-28
export transaction helpers

lukechampine:
As of Go 1.10, `go test` also runs `go vet`. `go vet` complains when a test file in one package uses test-only identifiers in another package. This is unfortunate, but there may be good justifications for this behavior.

### DIFF
--- a/types/transaction_helpers.go
+++ b/types/transaction_helpers.go
@@ -1,5 +1,3 @@
-// +build testing
-
 package types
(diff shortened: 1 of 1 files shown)


DavidVorick:
One thing we could do is move the transaction graph stuff to the siatest package.

I guess that means we can't use them outside the siatest package, due to import loops.

PR #2808, opened 2018-03-02
Dependencies cleanup

ChrisSchinnerl:
This PR is supposed to clean up the dependencies interface of Sia. Currently we have a `dependencies.go` file for every (sub)module that uses the dependencies interface. That means we have the same interface and same production implementation in multiple files.
In this PR all of the `dependencies` and `productionDependencies` interfaces/implementations are merged into one single top layer `dependencies.go` file. That way we have a single point of entry for people trying to understand the dependencies system and can also provide a nice description in for GoDocs. 
Custom dependencies should be created by embedding the `modules.ProductionDependencies` and overwriting the required method.


modules/dependencies.go
```diff
@@ -0,0 +1,344 @@
...
+
+	// Probabilistically return a number greater than one.
+	var val uint64
+	for {
+		val++
+		coin := fastrand.Intn(2)
+		if coin == 0 {
+			break
+		}
+	}
```
lukechampine:
```go
val := uint64(1)
for fastrand.Intn(2) != 0 {
    val++
}
```

modules/dependencies.go
```diff
@@ -0,0 +1,338 @@
...
+type (
+	Dependencies interface {
+		// AtLeastOne will return a value that is at least one. In production,
+		// the value should always be one. This function is used to test the
+		// idempotency of actions, so during testing sometimes the value
+		// returned will be higher, causing an idempotent action to be
+		// committed multiple times. If the action is truly idempotent,
+		// committing it multiple times should not cause any problems or
+		// changes.
+		AtLeastOne() uint64
```
lukechampine: it looks like this function is only used by the host to simulate performing an operation more than once. I suggest that we refactor the code to instead use `Disrupt`. We should try to keep the number of methods as small as possible.

PR #2809, opened 2018-03-02
only update subscribers if there are any

tbenz9:
When syncing the consensus.db from genesis as much as 25% of the time it takes to sync is spent on `cs.updateSubscribers()`.  There is no need update subscribers if you don't have any.

If you start `siad` with only a subset of the modules you may not have any subscribers.  For example `siad -M cg -d /root/Sia` results in 0 subscribers.

https://github.com/NebulousLabs/Sia/blob/17432ba5e810652768b5909a725e5d599f0414b6/modules/consensus/accept.go#L311

### DIFF
--- a/modules/consensus/accept.go
+++ b/modules/consensus/accept.go
@@ -306,9 +306,12 @@ func (cs *ConsensusSet) managedAcceptBlocks(blocks []types.Block) (blockchainExt
 	if !chainExtended {
 		return false, modules.ErrNonExtendingBlock
 	}
-	// Send any changes to subscribers.
-	for i := 0; i < len(changes); i++ {
-		cs.updateSubscribers(changes[i])
+	// Check if there are any subscribers
+	if len(cs.subscribers) > 0 {
+		// Send any changes to subscribers.
+		for i := 0; i < len(changes); i++ {
+			cs.updateSubscribers(changes[i])
+		}
 	}
 	return chainExtended, nil
 }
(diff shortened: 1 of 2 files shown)


tbenz9:
I'll explain my use case a bit more, I like to always sync the blockchain from the network, and avoid bootstrapping when I can.  I've been a little frustrated with how slow it is, I discovered that if I start `siad` with only the consensus and gateway modules loaded I could sync the consensus much faster than usual.  My plan is to sync quickly, then restart `siad` with the rest of the modules.  This could be a relatively easy route for non-bootstrapping performance improvements.

modules/consensus/accept.go
```diff
@@ -306,9 +306,12 @@ func (cs *ConsensusSet) managedAcceptBlocks(blocks []types.Block) (blockchainExt
 	if !chainExtended {
 		return false, modules.ErrNonExtendingBlock
 	}
-	// Send any changes to subscribers.
-	for i := 0; i < len(changes); i++ {
-		cs.updateSubscribers(changes[i])
+	// Check if there are any subscribers
+	if len(cs.subscribers) > 0 {
```
ChrisSchinnerl: I feel like I would prefer this check inside of `updateSubscribers`. Something like the following at the beginning of the function. That way we don't need to add the check every time if we call `updateSubscribers` in multiple places.
```
if len(cs.subscribers) == 0 {
    return
}
```

PR #2810, opened 2018-03-03
Extend renter settings endpoint for ratelimiting bandwidth

ChrisSchinnerl:
This PR changes quite a few things about the renter's settings endpoint. All fields of the allowance must be optional to change because we don't want to change the allowance every time we want to change a different setting. That means many tests fail and also have to be adjusted.
Now `SetAllowance` will still sanity check the allowance before applying it, but every value that is not set will cause the setting to be unchanged. If that results in a field being not set in still not being set, `SetAllowance` should return an error. That means the user can't simply call the endpoint without providing parameters and still expect sane values to be set automatically.

### DIFF
--- a/modules/renter/contractor/allowance.go
+++ b/modules/renter/contractor/allowance.go
@@ -43,6 +43,9 @@ func (c *Contractor) SetAllowance(a modules.Allowance) error {
 	if reflect.DeepEqual(a, modules.Allowance{}) {
 		return c.managedCancelAllowance()
 	}
+	if reflect.DeepEqual(a, c.allowance) {
+		return nil
+	}
 
 	// sanity checks
 	if a.Hosts == 0 {
--- a/node/api/renter.go
+++ b/node/api/renter.go
@@ -1,8 +1,5 @@
 package api
 
-// TODO: When setting renter settings, leave empty values unchanged instead of
-// zeroing them out.
-
 import (
 	"fmt"
 	"net/http"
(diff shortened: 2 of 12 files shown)


modules/renter/contractor/allowance.go
```diff
@@ -43,6 +43,9 @@ func (c *Contractor) SetAllowance(a modules.Allowance) error {
 	if reflect.DeepEqual(a, modules.Allowance{}) {
 		return c.managedCancelAllowance()
 	}
+	if reflect.DeepEqual(a, c.allowance) {
```
lukechampine: hmm. At first I thought this was unsafe, because two `types.Currency` values could have different underlying bytes. But it looks like `big.Int` values are always normalized, so that should never occur.

node/api/renter.go
```diff
@@ -178,62 +175,81 @@ func (api *API) renterHandlerGET(w http.ResponseWriter, req *http.Request, _ htt
...
-	if !ok {
-		WriteError(w, Error{"unable to parse funds"}, http.StatusBadRequest)
-		return
-	}
+	// Get the existing settings
+	settings := api.renter.Settings()
 
+	// Scan the allowance amount. (optional parameter)
+	if f := req.FormValue("funds"); f != "" {
+		funds, ok := scanAmount(req.FormValue("funds"))
```
lukechampine: declaring a local variable (`f`) seems to hurt more than it helps. The lookup isn't costly, so the only reason to avoid multiple calls to `req.FormValue` is verbosity.

ChrisSchinnerl: Yep, that's the only reason why I did it. I don't like to have `"funds"` twice in 2 lines. If you want I can change that back. I missed replacing it with `f` in this line.

PR #2821, opened 2018-03-05
spellings

tbenz9:


node/api/renter.go
```diff
@@ -317,12 +317,12 @@ func (api *API) renterDownloadsHandler(w http.ResponseWriter, _ *http.Request, _
...
-			Error:               di.Error,
-			Received:            di.Received,
-			StartTime:           di.StartTime,
-			TotalDataTransfered: di.TotalDataTransfered,
+			Completed:            di.Completed,
+			EndTime:              di.EndTime,
+			Error:                di.Error,
+			Received:             di.Received,
+			StartTime:            di.StartTime,
+			TotalDataTransferred: di.TotalDataTransferred,
```
lukechampine: looks like the other field needs to be fixed too

tbenz9: Am I blind?  Which other field are you talking about?

lukechampine: The build failed with:

>node/api/renter.go:325:28: di.TotalDataTransferred undefined (type modules.DownloadInfo has no field or method TotalDataTransferred)

So `di` is a `modules.DownloadInfo` which still has the wrong spelling

PR #2822, opened 2018-03-05
increment revision number in hostSettings

ChrisSchinnerl:
Whenever `externalSettings` is called we calculate new contract fees according to the current transaction pool's fee estimation and in the future we will also dynamically adjust other metrics like storage and bandwidth price. That means two subsequent calls to `externalSettings` might not create the exact same settings.
That means we need to update the revision number every time we call `externalSettings` and can move the code that increments the revision number into `externalSettings` instead of incrementing it every time right before we call the method.


DavidVorick:
Can you explain this change a bit more in depth?

ChrisSchinnerl:
@DavidVorick I updated the description.

PR #2826, opened 2018-03-05
Improve redundancy reporting

ChrisSchinnerl:
This PR improves the redundancy report of the renter endpoint.
The renter doesn't count contracts that are not good for renewal towards redundancy while the renter endpoint does. That's why the renter might report redundancies > 3x.
Unfortunately we can't simply change the renter endpoint's behavior to match the renter's internal behavior since that might lead to redundancies < 1x which might lead to a poor UX even though the files can still be downloaded.
That's why this PR changes the endpoint to calculate both redundancies and if the redundancy without `goodForRenew` contracts drops below 1 it reports the higher redundancy.

### DIFF
--- a/modules/renter/files.go
+++ b/modules/renter/files.go
@@ -134,11 +134,12 @@ func (f *file) uploadProgress() float64 {
 // becomes available when this redundancy is >= 1. Assumes that every piece is
 // unique within a file contract. -1 is returned if the file has size 0. It
 // takes one argument, a map of offline contracts for this file.
-func (f *file) redundancy(isOffline func(types.FileContractID) bool) float64 {
+func (f *file) redundancy(isOffline func(types.FileContractID) bool, goodForRenew func(types.FileContractID) bool) float64 {
 	if f.size == 0 {
 		return -1
 	}
 	piecesPerChunk := make([]int, f.numChunks())
+	piecesPerChunkNoRenew := make([]int, f.numChunks())
 	// If the file has non-0 size then the number of chunks should also be
 	// non-0. Therefore the f.size == 0 conditional block above must appear
 	// before this check.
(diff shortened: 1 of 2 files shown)


modules/renter/files.go
```diff
@@ -239,6 +264,11 @@ func (r *Renter) FileList() []modules.FileInfo {
 		offline := r.hostContractor.IsOffline(id)
 		return offline
 	}
+	goodForRenew := func(id types.FileContractID) bool {
+		id = r.hostContractor.ResolveID(id)
+		cu, ok := r.hostContractor.ContractUtility(id)
```
DavidVorick: should we expand the `cu` object to also include the `IsOffline` information? That will limit the number of times that we are calling out of package

DavidVorick:
Overall looks good to me. My guess is a lot of this will change once we update the file structure, we should no longer need to loop through every chunk every time we want the redundancy of a file.

ChrisSchinnerl:
@DavidVorick I merged the `isOffline` and `goodForRenew` methods into a `contractStatus` method that returns both.

modules/renter/files.go
```diff
@@ -147,21 +148,46 @@ func (f *file) redundancy(isOffline func(types.FileContractID) bool) float64 {
 		return -1
 	}
 	for _, fc := range f.contracts {
+		offline, gfr := contractStatus(fc.ID)
```
DavidVorick: gfr is not readable to someone unfamiliar with the code

modules/renter/files.go
```diff
@@ -88,10 +88,10 @@ func (f *file) numChunks() uint64 {
 }
 
 // available indicates whether the file is ready to be downloaded.
-func (f *file) available(isOffline func(types.FileContractID) bool) bool {
+func (f *file) available(contractStatus func(types.FileContractID) (bool, bool)) bool {
```
DavidVorick: the return values here should probably be named, confusing to have a `(bool, bool)) bool` construction

modules/renter/files.go
ChrisSchinnerl: done

modules/renter/files.go
ChrisSchinnerl: done

PR #2829, opened 2018-03-06
fix make all

lukechampine:
#2827 broke `make all`, which used to run `release-std`. The comment was also inconsistent with the behavior. I changed the comment instead of the behavior, but I'm fine changing the behavior instead. In practice I only use `make` to quickly check for compile-time errors, so the exact build constants don't affect my workflow.


DavidVorick:
thanks for catching this

PR #2830, opened 2018-03-06
Sane defaults for renter settings

ChrisSchinnerl:
Seems like the previous change broke siac and Sia-UI since they expect the settings to choose sane defaults for them. This should fix that by setting sane defaults only if the value hasn't been set previously.

### DIFF
--- a/node/api/renter.go
+++ b/node/api/renter.go
@@ -198,6 +198,9 @@ func (api *API) renterHandlerPOST(w http.ResponseWriter, req *http.Request, _ ht
 		} else {
 			settings.Allowance.Hosts = hosts
 		}
+	} else if settings.Allowance.Hosts == 0 {
+		// Sane defaults if host haven't been set before.
+		settings.Allowance.Hosts = recommendedHosts
 	}
 	// Scan the period. (optional parameter)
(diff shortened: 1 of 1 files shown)


node/api/renter.go
```diff
@@ -198,6 +198,9 @@ func (api *API) renterHandlerPOST(w http.ResponseWriter, req *http.Request, _ ht
 		} else {
 			settings.Allowance.Hosts = hosts
 		}
+	} else if settings.Allowance.Hosts == 0 {
+		// Sane defaults if host haven't been set before.
+		settings.Allowance.Hosts = recommendedHosts
```
DavidVorick: hmm are we sure that we want this to be an api thing? I feel like it should be more the client's responsibility to set these defaults.

I guess I'm split on this

DavidVorick: The big reason being that different clients are going to have different defaults. If you are running a home app, 50 is pretty good. If you are running an SMB cold storage solution, you might want 100 or 200 hosts instead.

ChrisSchinnerl: I'm not a big fan of this either. Unfortunately that was the behavior in the past and I had to add it again to not break compatibility.

PR #2832, opened 2018-03-07
Port TestRenterLocalRepair and TestRenterRemoteRepair to siatest

ChrisSchinnerl:
This PR ports `TestRenterLocalRepair` and `TestRenterRemoteRepair` over to the siatest package since they seem to fail randomly in our legacy testing framework.


DavidVorick:
Looks like AppVeyor is failing due to an issue with `NewTestGroup` where the renter is unable to form contracts.

It looks like in the `NewGroup` function we announce the hosts, mine a block, wait for the hosts to be found in the hostdb, then set the allowance, then wait for contracts. When you set the allowance, it looks like in the code it immediately triggers a new cycle of contract forming, so I don't see what the issue might be.

Is is possible that the hostdb doesn't actually have enough hosts in it? Or maybe the hostdb updated, but somehow the contractor was able to run without seeing those updates (I don't think that's the case).

One of the weird things is that the `build.Retry` block is doing `1000, 100*time.Millisecond` as its timeout parameters. That should end up being 100 seconds total, but it looks like appveyor failed after only 36 seconds on one of the failures, and after only 19 seconds on the other failure. (as a side node, that should be changed to using the siatest Retry function) it looks like the `build.Retry` function is correctly implemented, so I'm not sure why appveyor is sleeping for only 1/5 the time it's supposed to be sleeping.

Not sure what's going on with that failure.

modules/renter/consts.go
```diff
@@ -111,6 +111,14 @@ var (
 		Testing:  3 * time.Second,
 	}).(time.Duration)
 
+	// remoteRepairDownloadThreshold defines the threshold in percent under
+	// which the renter starts repairing a file that is not available on disk.
+	remoteRepairDownloadThreshold = build.Select(build.Var{
+		Dev:      0.0,  // always download
+		Standard: 0.25, // download if at least 25% are missing
+		Testing:  0.0,  // always download
```
DavidVorick: if testing and dev are always `0`, you have no coverage of the case where where it's not supposed to download a file to do a repair. I think my preference is for `0.25`, because it limits the number of hosts that you need in order to have a file that's missing less than 25% of its redundancy but more than 0% of its redundancy.

ChrisSchinnerl: @DavidVorick the problem is that introduces non deterministic behavior when waiting for the redundancy to increment. If we want to repair 5 chunks by downloading them first and not all of them are added to the heap in one iteration of the repair loop, the chunk that ends up in the next iteration might not be downloaded because the redundancy is already too high.
The only other workaround would be to export this field and only check for 75% redundancy in the test.

siatest/renter.go
```diff
@@ -116,9 +117,9 @@ func (tn *TestNode) Upload(lf *LocalFile, dataPieces, parityPieces uint64) (*Rem
 }
 
 // UploadNewFile initiates the upload of a filesize bytes large file.
-func (tn *TestNode) UploadNewFile(filesize int, dataPieces uint64, parityPieces uint64) (rf *RemoteFile, err error) {
+func (tn *TestNode) UploadNewFile(filesize int, dataPieces uint64, parityPieces uint64) (lf *LocalFile, rf *RemoteFile, err error) {
```
DavidVorick: Same thing as the other PR, we really prefer to avoid named return values where we can, we've found it causes mistakes especially in longer functions. Sometimes it's just the easiest way to do things, but for longer functions it causes bugs. For example, a blank `return` halfway through the function can change meaning if you add another return value or change the order in which you do things in the function, and that change may not even appear in the diff for reviewers because the `return` value itself is buried somewhere between the lines that changed.

siatest/renter.go
```diff
@@ -134,8 +135,8 @@ func (tn *TestNode) UploadNewFile(filesize int, dataPieces uint64, parityPieces
 
 // UploadNewFileBlocking uploads a filesize bytes large file and waits for the
 // upload to reach 100% progress and redundancy.
-func (tn *TestNode) UploadNewFileBlocking(filesize int, dataPieces uint64, parityPieces uint64) (rf *RemoteFile, err error) {
-	rf, err = tn.UploadNewFile(filesize, dataPieces, parityPieces)
+func (tn *TestNode) UploadNewFileBlocking(filesize int, dataPieces uint64, parityPieces uint64) (lf *LocalFile, rf *RemoteFile, err error) {
+	lf, rf, err = tn.UploadNewFile(filesize, dataPieces, parityPieces)
 	if err != nil {
 		return
```
DavidVorick: same here

modules/renter/consts.go
DavidVorick: Seems reasonable to me to export the field and allow the test package to only check it goes above that redundancy.

Are you sure that is the problem though? If it gets started on a chunk it should take it to 100%. If it doesn't get started on a chunk, the chunk shouldn't have any redundancy missing on the next iteration.

To the best of my knowledge, it shouldn't drop a chunk that's halfway complete, once you pass pieces out to workers they don't get interrupted unless they are determined to be bad workers.

siatest/renter.go
```diff
@@ -213,3 +211,42 @@ func (tn *TestNode) WaitForUploadRedundancy(rf *RemoteFile, redundancy float64)
...
+		if file.Redundancy > redundancy {
+			return fmt.Errorf("redundancy should be %v but was %v", redundancy, file.Redundancy)
+		}
+		return nil
+	})
+}
+
+// knowsHost checks if tn has a certain host in its hostdb. This check is
+// performed using the host's public key.
+func (tn *TestNode) knowsHost(host *TestNode) error {
```
DavidVorick: Why is this unexported when most of the other methods are exported?

siatest/renter/renter_test.go
```diff
@@ -75,3 +79,106 @@ func testUploadDownload(t *testing.T, tg *siatest.TestGroup) {
 		t.Error(err)
 	}
 }
+
+// testRenterLocalRepair tests if a renter correctly repairs a file from disk
+// after a host goes offline.
+func testRenterLocalRepair(t *testing.T, tg *siatest.TestGroup) {
+	// Grab the first of the group's renters
+	renter := tg.Renters()[0]
```
DavidVorick: need a check both here and in the other test that there are at least 2 hosts, otherwise this is going to fail.

siatest/testgroup.go
```diff
@@ -244,25 +251,32 @@ func fundNodes(miner *TestNode, nodes map[*TestNode]struct{}) error {
...
+				if renter == host {
+					// We don't care if the renter is also a host.
+					return nil
 				}
-				return errors.New("renter doesn't have enough active hosts in hostdb")
+				// Check if the renter has the host in its db.
+				if err := renter.knowsHost(host); err != nil {
+					if numRetries%10 == 0 {
+						if err := miner.MineBlock(); err != nil {
+							return err
```
DavidVorick: This is too much indentation, can you re-structure this function to clean that up a bit?

modules/renter/consts.go
ChrisSchinnerl: Yes I'm quite sure. It usually happened in the following scenario:
The group has 4 offline hosts and 5 online hosts. The renter has seen all of those hosts and has contracts with all of them but the worker pool was not updated yet and contains only 8 hosts.
The first iteration of the repair starts but can not upload the last piece since one of the hosts is still missing from the pool.
Before the second iteration starts we update the pool and it now contains 9 workers. The chunks are queued again but this time they have too many completed pieces to be downloaded again.

PR #2833, opened 2018-03-07
Display build metadata on `siad` startup and via `siac version`

tbenz9:
When running your own build it can sometimes be difficult to identify exactly which commit you are running from.  This adds additional build information to the `siac version` command, and displays the git revision hash when `siad` starts up.  If the build was not completed via the Makefile `siad` will display a warning on start and `siac version` will not display any of the build information.

### DIFF
--- a/build/commit.go
+++ b/build/commit.go
@@ -2,5 +2,8 @@ package build
 
 // GitRevision and BuildTime get assigned via the Makefile when built.
 var (
-	GitRevision, BuildTime string
+	// GitRevision is the git commit hash used when built
+	GitRevision string
+	// BuildTime is the date and time the build was completed
+	BuildTime string
 )
--- a/build/commit.go
+++ b/build/commit.go
@@ -1,0 +1,7 @@
+package build
+
+// GitRevision, GitBranch, and BuildTime get assigned via the Makefile when
+// built.
+var (
+	GitRevision, GitBranch, BuildTime string
+)
--- a/build/commit.go
+++ b/build/commit.go
@@ -1,7 +1,6 @@
 package build
 
-// GitRevision, GitBranch, and BuildTime get assigned via the Makefile when
-// built.
+// GitRevision and BuildTime get assigned via the Makefile when built.
 var (
-	GitRevision, GitBranch, BuildTime string
+	GitRevision, BuildTime string
 )
--- a/cmd/siac/daemoncmd.go
+++ b/cmd/siac/daemoncmd.go
@@ -49,8 +49,7 @@ func versioncmd() {
 	fmt.Println("\tVersion " + build.Version)
 	if build.GitRevision != "" {
 		fmt.Println("\tGit Revision " + build.GitRevision)
-		fmt.Println("\tGit Branch " + build.GitBranch)
-		fmt.Println("\tBuild Time " + build.BuildTime)
+		fmt.Println("\tBuild Time   " + build.BuildTime)
 	}
 	var dvg api.DaemonVersionGet
 	err := getAPI("/daemon/version", &dvg)
@@ -62,8 +61,7 @@ func versioncmd() {
 	fmt.Println("\tVersion " + dvg.Version)
 	if build.GitRevision != "" {
 		fmt.Println("\tGit Revision " + dvg.GitRevision)
-		fmt.Println("\tGit Branch " + dvg.GitBranch)
-		fmt.Println("\tBuild Time " + dvg.BuildTime)
+		fmt.Println("\tBuild Time   " + dvg.BuildTime)
 	}
 }
 
--- a/cmd/siac/daemoncmd.go
+++ b/cmd/siac/daemoncmd.go
@@ -45,14 +45,26 @@ type updateInfo struct {
 
 // version prints the version of siac and siad.
 func versioncmd() {
-	fmt.Println("Sia Client v" + build.Version)
+	fmt.Println("Sia Client")
+	fmt.Println("\tVersion " + build.Version)
+	if build.GitRevision != "" {
+		fmt.Println("\tGit Revision " + build.GitRevision)
+		fmt.Println("\tGit Branch " + build.GitBranch)
+		fmt.Println("\tBuild Time " + build.BuildTime)
+	}
 	var dvg api.DaemonVersionGet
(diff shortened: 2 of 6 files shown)


tbenz9:
Sample output of `siac version`:
```tbenz9@spartan1:~$ siac version
Sia Client
        Version 1.3.2
        Git Revision 9a1aab6
        Git Branch build-version
        Build Time 2018-03-07T21:05:59-0800
Sia Daemon
        Version 1.3.2
        Git Revision 9a1aab6
        Git Branch build-version
        Build Time 2018-03-07T21:05:59-0800
```

Sample output of `siad` starting:
```tbenz9@spartan1:~$ ~/go/bin/siad -M cg -d /home/tbenz9/Documents/Sia/                   
Sia Daemon v1.3.2
Git Revision 9a1aab6
Loading...
(0/2) Loading siad...
(1/2) Loading gateway...
(2/2) Loading consensus...
Finished loading in 0.221536432 seconds
```

cmd/siac/daemoncmd.go
```diff
@@ -45,14 +45,26 @@ type updateInfo struct {
...
 		return
 	}
-	fmt.Println("Sia Daemon v" + dvg.Version)
+	fmt.Println("Sia Daemon")
+	fmt.Println("\tVersion " + dvg.Version)
+	if build.GitRevision != "" {
+		fmt.Println("\tGit Revision " + dvg.GitRevision)
+		fmt.Println("\tGit Branch " + dvg.GitBranch)
+		fmt.Println("\tBuild Time " + dvg.BuildTime)
+	}
```
ChrisSchinnerl: I think something like the following might improve the alignment of the left and right block and might make it easier to read.
```
fmt.Println("\tGit Revision: " + dvg.GitRevision)
fmt.Println("\tGit Branch:   " + dvg.GitBranch)
fmt.Println("\tBuild Time:   " + dvg.BuildTime)
```

ChrisSchinnerl:
I don't know if we really need the build time and branch. What do you think @lukechampine ?

lukechampine:
branch shouldn't be necessary; the commit hash should be sufficient. Build time isn't strictly necessary, but I could imagine it helping in some cases (e.g. "when did we build the last RC?"). The one thing I would change is using the default `date` command instead of the specially formatted one. `date` by default outputs a nice human-readable representation of the date.

tbenz9:
@lukechampine `date` is formatted that way because `make` is very weird about what it tries to interpret as arguments.  For example this doesn't work:
```
tbenz9@spartan1:~/go/src/github.com/NebulousLabs/Sia$ make release                                                                                           
go install -tags='netgo' -a -ldflags='-s -w -X github.com/NebulousLabs/Sia/build.GitRevision=✗-e9c221a -X github.com/NebulousLabs/Sia/build.BuildTime="Fri Mar  9 10:05:23 PST 2018"' ./build ./cmd/siac ./cmd/siad ...
```
But this does:
```
tbenz9@spartan1:~/go/src/github.com/NebulousLabs/Sia$ make release                                                                                           
go install -tags='netgo' -a -ldflags='-s -w -X github.com/NebulousLabs/Sia/build.GitRevision=✗-e9c221a -X github.com/NebulousLabs/Sia/build.BuildTime="FriMar910:05:23PST2018"' ./build ./cmd/siac ./cmd/siad ...
```
You would think the `""` around date would allow it to be passed in as a string, but that is not working for me.  We can either troubleshoot `make` or come up with a date string without spaces.

ChrisSchinnerl:
@tbenz9 I don't have lots of experience with that but can you escape the spaces with a `\` before the space?

tbenz9:
I tried that when I assign the `BUILD_TIME` variable but it doesn't seem to make any difference.
```
BUILD_TIME=$(shell date +%a\ %b\ %d\ %T\ %Z\ %Y)
```
Still results in:
```
tbenz9@spartan1:~/go/src/github.com/NebulousLabs/Sia$ make release                                                                                           
go install -tags='netgo' -a -ldflags='-s -w -X github.com/NebulousLabs/Sia/build.GitRevision=✗-e9c221a -X github.com/NebulousLabs/Sia/build.BuildTime="Fri Mar 09 10:45:09 PST 2018"' ./build ./cmd/siac ./cmd/siad ...
```

tbenz9:
@ChrisSchinnerl good idea but didn't work.
```
tbenz9@spartan1:~/go/src/github.com/NebulousLabs/Sia$ make release                                                                                                                                                                                                                                                            
go install -tags='netgo' -a -ldflags='-s -w -X github.com/NebulousLabs/Sia/build.GitRevision=✗-e9c221a -X github.com/NebulousLabs/Sia/build.BuildTime="Fri\ Mar\ 09\ 11:01:42PST2018"' ./build ./cmd/siac ./cmd/siad
``` 
still fails with a 'usage' error meaning the arguments aren't recognized.

ChrisSchinnerl:
In case anyone else is wondering, we found the issue. The following works:
`go install -tags='netgo' -a -ldflags="-s -w -X 'github.com/NebulousLabs/Sia/build.BuildTime=Fri Mar 9 10:05:23 PST 2018'" ./build ./cmd/siac`

lukechampine:
ah yeah, single quotes is the way to go.

You can actually also just enclose the commands in backticks (`\``); then, when you run the `go install` command, those commands will be run. e.g.

```
go install -ldflags="-X 'github.com/NebulousLabs/Sia/build.BuildTime=`date`'" 
```
but what we have now is fine

node/api/daemon.go
```diff
@@ -2,5 +2,7 @@ package api
 
 // DaemonVersionGet contains information about the running daemon's version.
 type DaemonVersionGet struct {
-	Version string
+	Version     string
+	GitRevision string
+	BuildTime   string
```
lukechampine: these will need to be documented, but we can handle that in a separate PR. I want to include this change in the rc

build/commit.go
```diff
@@ -0,0 +1,6 @@
+package build
+
+// GitRevision and BuildTime get assigned via the Makefile when built.
+var (
+	GitRevision, BuildTime string
```
lukechampine: looks like `golint` complains about these:
```
build/commit.go:5:2: exported var BuildTime should have its own declaration
```

PR #2839, opened 2018-03-08
Peer to peer ip discovery

ChrisSchinnerl:
So far the only centralized aspect of Sia has been the way a node discovers its public ip address if upnp isn't available. Our workaround was to use myexternalip.com as a fallback but recently they started banning certain ranges of ips and are therefore not reliable. This PR adds a second fallback (that will hopefully replace myexternalip.com completely one day) by using the peers the node is connected to. Once a certain number of peers is available, we query each one of them and if the majority of responses report the same ip, we accept it as our own public ip.

### DIFF
--- a/modules/gateway/ip.go
+++ b/modules/gateway/ip.go
@@ -36,10 +36,7 @@ func (g *Gateway) managedIPFromPeers() (string, error) {
 		g.mu.RUnlock()
 		// Check if there are enough peers. Otherwise wait.
 		if len(peers) < minPeersForIPDiscovery {
-			select {
-			case <-time.After(peerDiscoveryRetryInterval):
-			case <-g.peerTG.StopChan():
-			}
+			g.waitForPeerDiscoverySignal()
 			continue
 		}
 		// Ask all the peers about our ip in parallel
@@ -75,10 +72,7 @@ func (g *Gateway) managedIPFromPeers() (string, error) {
 		}
 		// If there haven't been enough successful responses we wait some time.
 		if successfulResponses < minPeersForIPDiscovery {
-			select {
-			case <-time.After(peerDiscoveryRetryInterval):
-			case <-g.peerTG.StopChan():
-			}
+			g.waitForPeerDiscoverySignal()
 			continue
 		}
 		// If an address was returned by more than half the peers we consider
@@ -89,9 +83,15 @@ func (g *Gateway) managedIPFromPeers() (string, error) {
 			}
 		}
 		// Otherwise we wait before trying again.
-		select {
-		case <-time.After(peerDiscoveryRetryInterval):
-		case <-g.peerTG.StopChan():
-		}
+		g.waitForPeerDiscoverySignal()
+	}
+}
+
+// waitForPeerDiscoverySignal blocks for the time specified in
+// peerDiscoveryRetryInterval.
+func (g *Gateway) waitForPeerDiscoverySignal() {
+	select {
+	case <-time.After(peerDiscoveryRetryInterval):
+	case <-g.peerTG.StopChan():
 	}
 }
--- a/modules/gateway/ip.go
+++ b/modules/gateway/ip.go
@@ -42,7 +42,7 @@ func (g *Gateway) managedIPFromPeers() (string, error) {
 		g.mu.RUnlock()
 		// Check if there are enough peers. Otherwise wait.
 		if len(peers) < minPeersForIPDiscovery {
-			g.waitForPeerDiscoverySignal(timeout)
+			g.managedSleep(peerDiscoveryRetryInterval)
 			continue
 		}
 		// Ask all the peers about our ip in parallel
@@ -78,7 +78,7 @@ func (g *Gateway) managedIPFromPeers() (string, error) {
 		}
 		// If there haven't been enough successful responses we wait some time.
 		if successfulResponses < minPeersForIPDiscovery {
-			g.waitForPeerDiscoverySignal(timeout)
+			g.managedSleep(peerDiscoveryRetryInterval)
 			continue
 		}
 		// If an address was returned by more than half the peers we consider
@@ -90,16 +90,6 @@ func (g *Gateway) managedIPFromPeers() (string, error) {
 			}
 		}
 		// Otherwise we wait before trying again.
-		g.waitForPeerDiscoverySignal(timeout)
-	}
-}
-
-// waitForPeerDiscoverySignal blocks for the time specified in
-// peerDiscoveryRetryInterval.
-func (g *Gateway) waitForPeerDiscoverySignal(timeout <-chan time.Time) {
-	select {
-	case <-time.After(peerDiscoveryRetryInterval):
-	case <-g.peerTG.StopChan():
-	case <-timeout:
+		g.managedSleep(peerDiscoveryRetryInterval)
 	}
 }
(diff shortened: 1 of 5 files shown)


ChrisSchinnerl:
The way it is implemented now `managedIPFromPeers` just keeps going forever or until the gateway is shut down. Maybe we want to add an argument that specifies a timeout if we ever call it from a place where we don't want to wait too long in a worst-case scenario.

modules/gateway/upnp.go
```diff
@@ -63,6 +63,9 @@ func (g *Gateway) threadedLearnHostname() {
 	if err != nil {
 		host, err = myExternalIP()
```
DavidVorick: we should have this check be last I think, and also have it only run if !build.Debug, that way debug builds of Sia get to see what happens if the centralized version doesn't exist as an alternative.

modules/gateway/ip.go
```diff
@@ -0,0 +1,97 @@
...
+func (g *Gateway) discoverPeerIP(conn modules.PeerConn) error {
+	conn.SetDeadline(time.Now().Add(connStdDeadline))
+	host, _, err := net.SplitHostPort(conn.RemoteAddr().String())
+	if err != nil {
+		return errors.AddContext(err, "failed to split host from port")
+	}
+	return encoding.WriteObject(conn, host)
+}
+
+// managedIPFromPeers asks the peers the node is connected to for the node's public ip address. If not enough peers are available
```
DavidVorick: comment not wrapped

modules/gateway/ip.go
```diff
@@ -0,0 +1,97 @@
...
+		// If there haven't been enough successful responses we wait some time.
+		if successfulResponses < minPeersForIPDiscovery {
+			g.waitForPeerDiscoverySignal()
+			continue
+		}
+		// If an address was returned by more than half the peers we consider
+		// it valid.
+		for addr, count := range addresses {
+			if count > successfulResponses/2 {
+				return addr, nil
```
DavidVorick: Let's add a log message here reporting our ip address as discovered from peer consensus.

DavidVorick:
Overall this looks good to me. Would want to see review from @lukechampine

modules/gateway/ip.go
```diff
@@ -0,0 +1,97 @@
...
+
+// managedIPFromPeers asks the peers the node is connected to for the node's public ip address. If not enough peers are available
+func (g *Gateway) managedIPFromPeers() (string, error) {
+	for {
+		// Check for shutdown signal.
+		select {
+		case <-g.peerTG.StopChan():
+			return "", errors.New("interrupted by shutdown")
+		default:
+		}
```
DavidVorick: Instead of being an infinite loop, this should be a loop that times out entirely after a few minutes. I'm thinking 5 minutes is probably a good amount of time.

modules/gateway/upnp.go
ChrisSchinnerl: I put the peer-to-peer discovery last because I think that we should give the network some time to upgrade to the new version. Otherwise the users who upgrade first will always experience slow ip discovery if they don't have a UPnP enabled router until the majority of the network supports the new RPC.
The next step would be to have `myExternalIP` as a fallback in a future update and the last one is to completely remove it once 99.99% of the network support the RPC.

modules/gateway/ip.go
ChrisSchinnerl: That makes be a bit nervous tbh. What if the gateway can't connect to peers for some reason and after 10 minutes it can? That means we stopped trying to discover our IP and the only way for the peer to get a valid ip is to restart the node.

DavidVorick: We have 2 issues then. The first is that means we may have a thread perpetually trying to collect an ip address, which for some reason isn't working. But the second is that we actually don't have any way right now of detecting if our ip address has changed. It should be uncommon, even for people with dynamic ip addresses (typically doesn't happen while you are online, but I believe it can if your ISP decides to do it). So we might want to have some loop that occasionally spins up another IP check.

Generally speaking, there is nothing in the function title `managedIPFromPeers` that indicates the function could potentially block for a long time (or even forever if something is wrong). Anything that has a super long return time should probably happen from a `threaded` function. For this use case, I think we can handle that pretty cleanly just by moving the outermost loop to the `threaded` function and leaving the rest of this logic here.

modules/gateway/upnp.go
```diff
@@ -54,33 +52,54 @@ func (g *Gateway) threadedLearnHostname() {
...
+		}
+		if err != nil {
+			g.log.Println("WARN: failed to discover external IP:", err)
+		}
+		// If we were unable to discover our IP we try again later.
+		if err != nil {
+			select {
+			case <-g.threads.StopChan():
+				return
+			case <-time.After(rediscoverIPIntervalFailure):
```
lukechampine: you can use `g.managedSleep` for calls like this

modules/gateway/upnp.go
```diff
@@ -54,33 +52,54 @@ func (g *Gateway) threadedLearnHostname() {
...
+		g.mu.Unlock()
+
+		g.log.Println("INFO: our address is", addr)
+
+		// Rediscover the IP later in case it changed.
+		select {
+		case <-g.threads.StopChan():
+			return
+		case <-time.After(rediscoverIPIntervalSuccess):
+		}
```
lukechampine: ditto

modules/gateway/ip.go
```diff
@@ -0,0 +1,105 @@
...
+			}
+		}
+		// Otherwise we wait before trying again.
+		g.waitForPeerDiscoverySignal(timeout)
+	}
+}
+
+// waitForPeerDiscoverySignal blocks for the time specified in
+// peerDiscoveryRetryInterval.
+func (g *Gateway) waitForPeerDiscoverySignal(timeout <-chan time.Time) {
```
lukechampine: `waitForSignal` sounds like a specific signal is being received, when in reality it's just a timeout.

Given that `peerDiscoveryRetryInterval` is much smaller than `timeoutIPDiscovery`, I don't think there's much harm in not checking `timeout` at all. Worst-case, you sleep a little longer than you would otherwise. So you could replace this function with `g.managedSleep(peerDiscoveryRetryInterval)`

modules/gateway/upnp.go
```diff
@@ -70,21 +70,22 @@ func (g *Gateway) threadedLearnHostname() {
...
-			select {
-			case <-g.threads.StopChan():
-				return
-			case <-time.After(rediscoverIPIntervalFailure):
-				continue
-			}
+		if err != nil && g.managedSleep(rediscoverIPIntervalFailure) {
+			continue
+		} else if err != nil {
+			return // shutdown interrupted sleep
```
lukechampine: this is a bit surprising. The boring, unsurprising form is:
```go
if err != nil {
    if !g.managedSleep(rediscoverIPIntervalFailure) {
        return // shutdown interrupted sleep
    }
    continue
}
```

modules/gateway/upnp.go
```diff
@@ -70,21 +70,22 @@ func (g *Gateway) threadedLearnHostname() {
...
 		addr := modules.NetAddress(net.JoinHostPort(host, g.port))
 		g.mu.RUnlock()
 		if err := addr.IsValid(); err != nil {
 			g.log.Printf("WARN: discovered hostname %q is invalid: %v", addr, err)
-			return
+			if err != nil && g.managedSleep(rediscoverIPIntervalFailure) {
+				continue
+			} else if err != nil {
+				return // shutdown interrupted sleep
+			}
```
lukechampine: ditto

PR #2842, opened 2018-03-09
Pass ratelimit object when wraping net.Conn instead of having global ratelimit.

ChrisSchinnerl:
This PR changes the ratelimit to be module specific and not global by having to pass a `RateLimit` object to the constructor that wraps a `net.Conn`

This should not be merged without [#3](https://github.com/NebulousLabs/ratelimit/pull/3).

### DIFF
--- a/modules/renter/renter.go
+++ b/modules/renter/renter.go
@@ -291,9 +291,9 @@ func (r *Renter) SetSettings(s modules.RenterSettings) error {
 	if s.MaxDownloadSpeed == 0 && s.MaxUploadSpeed == 0 {
 		r.hostContractor.SetRateLimits(0, 0, 0)
 	} else {
-		// TODO: In the future we might want the user to be able to configure the
-		// packetSize using the API. For now the sane default is modules.SectorSize
-		// if the user wants to limit the connection.
+		// TODO: In the future we might want the user to be able to configure
+		// the packetSize using the API. For now the sane default is 16kib if
+		// the user wants to limit the connection.
 		r.hostContractor.SetRateLimits(s.MaxDownloadSpeed, s.MaxUploadSpeed, 4*4096)
 	}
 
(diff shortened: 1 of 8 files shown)


node/api/client/renter.go
```diff
@@ -55,6 +55,17 @@ func (c *Client) RenterPost(allowance modules.Allowance) (err error) {
 	return
 }
 
+// RenterPostRateLimit uses the /renter endpoint to change the renter's bandwidth rate
+// limit.
+func (c *Client) RenterPostRateLimit(readBPS, writeBPS int64, packetSize uint64) (err error) {
+	values := url.Values{}
+	values.Set("downloadspeed", strconv.FormatInt(readBPS, 10))
+	values.Set("uploadspeed", strconv.FormatInt(writeBPS, 10))
+	values.Set("packetsize", strconv.FormatUint(packetSize, 10))
```
lukechampine: these need better names. `downloadspeed` doesn't imply that it's a maximum download speed. Also, I don't see these values being handled on the server-side?

ChrisSchinnerl: @lukechampine @DavidVorick let me know what you want them to be.
At the moment they are handled in `renterHandlerPost`.

modules/renter/renter.go
```diff
@@ -281,15 +284,17 @@ func (r *Renter) SetSettings(s modules.RenterSettings) error {
...
-	// Set ratelimit
 	if s.MaxDownloadSpeed == 0 && s.MaxUploadSpeed == 0 {
-		ratelimit.SetLimits(0, 0, 0)
+		r.hostContractor.SetRateLimits(0, 0, 0)
 	} else {
-		ratelimit.SetLimits(s.MaxDownloadSpeed, s.MaxUploadSpeed, 4*4096)
+		// TODO: In the future we might want the user to be able to configure the
+		// packetSize using the API. For now the sane default is modules.SectorSize
+		// if the user wants to limit the connection.
+		r.hostContractor.SetRateLimits(s.MaxDownloadSpeed, s.MaxUploadSpeed, 4*4096)
```
DavidVorick: sane default is not modules.SectorSize, it's just 16 KiB

ChrisSchinnerl: fixed

Issue #2440, opened 2017-10-23
Document and test legal siapath characters


mtlynch:
We don't document which characters are legal and illegal in a siapath. It seems like the only real restriction is leading path traversal characters. Are characters that are illegal in Win32 and UNIX filesystems, such as `*` and `?` legal siapath characters?

We should document in the API documentation and add tests to `TestRenterSiapathValidate`

https://github.com/NebulousLabs/Sia/blob/d04d0a40b93836b8fe70c31feb6e876905b7e8f6/modules/renter/upload_test.go#L12

The lack of documentation is creating confusion on the [minio integration](https://github.com/minio/minio/pull/4802#issuecomment-338734546).

tbenz9:
Recommendation to Devs: Tag as "Documentation"

tbenz9:
We probably want to move `validateSiapath` into a more accessible location and call it from other functions like `RenameFile`.  This came up in Discord today where a user was able to use the API rename endpoint and name the file with a path traversal character (`/`).  `validateSiapath` is probably an appropriate place to check other illegal siapath characters like mtlynch mentions.

https://github.com/NebulousLabs/Sia/blob/b3541d6cc4d0d0c031b07ede558bf0c4fc32ce1a/modules/renter/upload.go#L52

PR #2854, opened 2018-03-12
Use validateSiapath function with RenameFile

tbenz9:
This PR is in response to #2440 and another issue reported via Discord where a user was able to rename a file alias to a name with illegal characters.  Eventually we'll want to expand our validateSiapth function to prevent other illegal characters as they are discovered.

### DIFF
--- a/modules/renter/renter.go
+++ b/modules/renter/renter.go
@@ -362,12 +362,10 @@ func validateSiapath(siapath string) error {
 	if strings.HasPrefix(siapath, "./") {
 		return errors.New("siapath connot begin with ./")
 	}
-	// check Contains
-	if strings.Contains(siapath, "/../") {
-		return errors.New("siapath cannot contain /../")
-	}
-	if strings.Contains(siapath, "/./") {
-		return errors.New("siapath cannot contain /./")
+	for _, pathElem := range strings.Split(siapath, "/") {
+		if pathElem == "." || pathElem == ".." {
+			return errors.New("siapath cannot contain . or .. elements")
+		}
 	}
 	return nil
 }
--- a/modules/renter/renter.go
+++ b/modules/renter/renter.go
@@ -18,6 +18,7 @@ package renter
 import (
 	"errors"
 	"reflect"
+	"strings"
 	"sync"
 
 	"github.com/NebulousLabs/Sia/build"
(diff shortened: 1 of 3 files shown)


modules/renter/renter.go
```diff
@@ -338,6 +339,29 @@ func (r *Renter) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+func validateSiapath(siapath string) error {
+	if strings.HasPrefix(siapath, "/") {
+		return errors.New("siapath cannot begin with /")
+	}
+
+	if siapath == "" {
+		return ErrEmptyFilename
+	}
+
+	if strings.Contains(siapath, "../") {
```
DavidVorick: isn't this a superset of the next .Contains()? And, I think the rule is a bit more complicated (explained this in another PR) - what's not allowed is to have a full file that's named '..' or '.', but you can have a './' in the name if for example your folder is named 'foo.', then you might have a file 'foo./bar', which should be legal.

Instead we want to check for the prefix '../' and the prefix './', as well as the contains '/../' and '/./'.

modules/renter/renter.go
```diff
@@ -338,6 +339,29 @@ func (r *Renter) ProcessConsensusChange(cc modules.ConsensusChange) {
...
 }
 
+// validateSiapath checks that a Siapath is a legal filename.
+// ../ is disallowed to prevent directory traversal, and paths must not begin
+// with / or be empty.
+func validateSiapath(siapath string) error {
+	if strings.HasPrefix(siapath, "/") {
+		return errors.New("siapath cannot begin with /")
+	}
+
```
DavidVorick: no need for all of these newlines between each `if` block, can just condense the whole function.

tbenz9: I'll take it out, I didn't write of modify this function at all.  I simply moved it from `modules/renter/upload.go` to `modules/renter/renter.go`.  If you don't like the formatting, blame yourself or Luke.  :)

modules/renter/renter.go
```diff
@@ -338,6 +339,39 @@ func (r *Renter) ProcessConsensusChange(cc modules.ConsensusChange) {
...
+		return errors.New("siapath connot begin with ./")
+	}
+	// check Contains
+	if strings.Contains(siapath, "/../") {
+		return errors.New("siapath cannot contain /../")
+	}
+	if strings.Contains(siapath, "/./") {
+		return errors.New("siapath cannot contain /./")
+	}
+	return nil
```
lukechampine: I think these are all variants of:
```go
for _, pathElem := range strings.Split(siapath, "/") {
    if pathElem == "." || pathElem == ".." {
        return errors.New("siapath cannot contain . or .. elements")
    }
}
```

tbenz9: Would you prefer your method?  I find mine easier to understand albeit slightly longer.

modules/renter/renter.go
DavidVorick: no blame, only fix :)

modules/renter/renter.go
DavidVorick: hmm. I do prefer the loop actually. It's more clear to me that we're explicitly searching for standalone instances of '.' or '..', instead of doing an enumeration which risks missing a particular corner case.

PR #2855, opened 2018-03-12
Speedup contract formation

ChrisSchinnerl:
This PR adds the method `UnconfirmedParents` to the transaction builder which returns the unconfirmed parents of the transaction set that is being constructed by the builder.
That allows us to pass those unconfirmed parents to the host when we form a contract with it instead of having to wait for some time to make sure the unconfirmed parents are propagated through the network.
As a result we can rapidly form contracts without a timeout instead of having to wait 1 minute between contracts.

### DIFF
--- a/modules/renter/contractor/contractor.go
+++ b/modules/renter/contractor/contractor.go
@@ -8,6 +8,7 @@ package contractor
 // renter lock.
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -18,7 +19,6 @@ import (
 	"github.com/NebulousLabs/Sia/persist"
 	siasync "github.com/NebulousLabs/Sia/sync"
 	"github.com/NebulousLabs/Sia/types"
-	"github.com/NebulousLabs/errors"
 )
 
 var (
--- a/modules/renter/contractor/contractor.go
+++ b/modules/renter/contractor/contractor.go
@@ -65,8 +65,8 @@ type Contractor struct {
 	renewedIDs   map[types.FileContractID]types.FileContractID
 }
 
-// resolveID returns the ID of the most recent renewal of id.
-func (c *Contractor) resolveID(id types.FileContractID) types.FileContractID {
+// readlockResolveID returns the ID of the most recent renewal of id.
+func (c *Contractor) readlockResolveID(id types.FileContractID) types.FileContractID {
 	newID, exists := c.renewedIDs[id]
 	for exists {
 		id = newID
@@ -126,7 +126,7 @@ func (c *Contractor) PeriodSpending() modules.ContractorSpending {
 func (c *Contractor) ContractByID(id types.FileContractID) (modules.RenterContract, bool) {
 	c.mu.RLock()
 	defer c.mu.RUnlock()
-	return c.contracts.View(c.resolveID(id))
+	return c.contracts.View(c.readlockResolveID(id))
 }
 
 // Contracts returns the contracts formed by the contractor in the current
@@ -142,7 +142,7 @@ func (c *Contractor) Contracts() []modules.RenterContract {
 func (c *Contractor) ContractUtility(id types.FileContractID) (modules.ContractUtility, bool) {
 	c.mu.RLock()
 	defer c.mu.RUnlock()
-	return c.contractUtility(id)
+	return c.readlockContractUtility(id)
 }
 
 // CurrentPeriod returns the height at which the current allowance period
@@ -156,7 +156,7 @@ func (c *Contractor) CurrentPeriod() types.BlockHeight {
 // ResolveID returns the ID of the most recent renewal of id.
 func (c *Contractor) ResolveID(id types.FileContractID) types.FileContractID {
 	c.mu.RLock()
-	newID := c.resolveID(id)
+	newID := c.readlockResolveID(id)
 	c.mu.RUnlock()
 	return newID
 }
--- a/modules/renter/contractor/contractor.go
+++ b/modules/renter/contractor/contractor.go
@@ -249,6 +249,9 @@ func NewCustomContractor(cs consensusSet, w wallet, tp transactionPool, hdb host
 	}
 
 	// Mark contract utility.
+	// TODO we probably have to force this to succeed even if there is a scan
+	// going on. Otherwise all the contract utilities might end up
+	// uninitialized.
 	c.managedMarkContractsUtility()
 
 	// Subscribe to the consensus set.
--- a/modules/renter/contractor/contractor.go
+++ b/modules/renter/contractor/contractor.go
@@ -8,7 +8,6 @@ package contractor
 // renter lock.
 
 import (
-	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -19,6 +18,7 @@ import (
 	"github.com/NebulousLabs/Sia/persist"
 	siasync "github.com/NebulousLabs/Sia/sync"
 	"github.com/NebulousLabs/Sia/types"
+	"github.com/NebulousLabs/errors"
 )
 
 var (
@@ -247,12 +247,6 @@ func NewCustomContractor(cs consensusSet, w wallet, tp transactionPool, hdb host
 	if err != nil && !os.IsNotExist(err) {
 		return nil, err
 	}
-
-	// Mark contract utility.
-	// TODO we probably have to force this to succeed even if there is a scan
-	// going on. Otherwise all the contract utilities might end up
-	// uninitialized.
-	c.managedMarkContractsUtility()
 
 	// Subscribe to the consensus set.
 	err = cs.ConsensusSetSubscribe(c, c.lastChange, c.tg.StopChan())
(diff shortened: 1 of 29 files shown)


modules/renter/contractor/contracts.go
```diff
@@ -553,6 +552,15 @@ func (c *Contractor) threadedContractMaintenance() {
 		return
 	}
 
+	pendingScans := c.hdb.PendingScans()
+	if pendingScans > 0 {
```
DavidVorick: the way that I envisioned this working was having 'RandomHosts' return an error if the hostdb itself felt that it was not at a place where it could provide a safe random sampling of hosts. That way, instead of having the contractor trying to figure out when the hostdb is safe for use, the hostdb worries about when itself is safe for use, and we separate concerns better.

That also means you don't need to export a function to count the number pending scans, you can just let the hostdb handle all of that black magic itself (in an area of the code where it's no longer black magic)

DavidVorick:
Overall this looks pretty good. I would like to see a test that does daisy chaining the same way that the renter does daisy chaining when it reaches out to hosts.

And then the pending() stuff

modules/renter/contractor/contractor.go
```diff
@@ -239,6 +239,9 @@ func newContractor(cs consensusSet, w wallet, tp transactionPool, hdb hostDB, co
 	}
 
 	// Mark contract utility.
+	// TODO we probably have to force this to succeed even if there is a scan
+	// going on. Otherwise all the contract utilities might end up
+	// uninitialized.
```
ChrisSchinnerl: The tests passed but this still worries me.

modules/renter/contractor/contractor.go
```diff
@@ -140,7 +140,9 @@ func (c *Contractor) Contracts() []modules.RenterContract {
 
 // ContractUtility returns the utility fields for the given contract.
 func (c *Contractor) ContractUtility(id types.FileContractID) (modules.ContractUtility, bool) {
-	return c.staticContractUtility(id)
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+	return c.contractUtility(id)
```
DavidVorick: technically should be a writelock, unless we call it `readlockContractUtility`.

I wish these locking conventions were just builtins with compiler enforcement

modules/renter/contractor/contracts.go
```diff
@@ -74,7 +73,10 @@ func (c *Contractor) managedMarkContractsUtility() error {
 	c.mu.RLock()
 	hostCount := int(c.allowance.Hosts)
 	c.mu.RUnlock()
-	hosts := c.hdb.RandomHosts(hostCount+minScoreHostBuffer, nil)
+	hosts, err := c.hdb.RandomHosts(hostCount+minScoreHostBuffer, nil)
```
DavidVorick: that's a terrible name for that constant.

let's change it to `minHostBufferForScore` lol

DavidVorick: no.

how about `randomHostsBufferForScore`

modules/renter/contractor/contracts.go
```diff
@@ -481,7 +485,9 @@ func (c *Contractor) threadedContractMaintenance() {
 				return
 			}
 			// Return the contract if it's not useful for renewing.
-			oldUtility, ok := c.staticContractUtility(id)
+			c.mu.RLock()
+			oldUtility, ok := c.contractUtility(id)
```
DavidVorick: same situation

modules/renter/contractor/contracts.go
```diff
@@ -585,7 +591,11 @@ func (c *Contractor) threadedContractMaintenance() {
 	}
 	initialContractFunds := c.allowance.Funds.Div64(c.allowance.Hosts).Div64(3)
 	c.mu.RUnlock()
-	hosts := c.hdb.RandomHosts(neededContracts*2+10, exclude)
+	hosts, err := c.hdb.RandomHosts(neededContracts*2+10, exclude)
```
DavidVorick: this +10 should be the buffer constant

modules/renter/hostdb/scan.go
```diff
@@ -290,6 +290,23 @@ func (hdb *HostDB) managedScanHost(entry modules.HostDBEntry) {
...
+	for {
+		hdb.mu.Lock()
+		length := len(hdb.scanList)
+		hdb.mu.Unlock()
+		if length == 0 {
+			break
+		}
+		select {
+		case <-hdb.tg.StopChan():
+		case <-time.After(time.Second):
```
DavidVorick: should we turn this into a constant?

modules/renter/hostdb/scan.go
```diff
@@ -329,6 +346,37 @@ func (hdb *HostDB) threadedScan() {
...
+	// Wait until the consensus set is synced. Only then we can be sure that
+	// the initial scan covers the whole network.
+	for {
+		if hdb.cs.Synced() {
+			break
+		}
+		select {
+		case <-hdb.tg.StopChan():
+			return
+		case <-time.After(time.Second):
```
DavidVorick: should this also be a constant?

modules/renter/proto/contract.go
```diff
@@ -29,6 +29,11 @@ type updateSetHeader struct {
 	Header contractHeader
 }
 
+type v132UpdateSetHeader struct {
```
DavidVorick: we need a docstring, ideally indicating as well what went wrong with the old thingy and why we chose to upgrade

modules/renter/proto/contract.go
```diff
@@ -56,6 +61,26 @@ type contractHeader struct {
 	Utility          modules.ContractUtility
 }
 
+type v132ContractHeader struct {
```
DavidVorick: docstring here too

modules/transactionpool/transactionpool.go
```diff
@@ -256,6 +256,26 @@ func (tp *TransactionPool) Transaction(id types.TransactionID) (types.Transactio
...
 
+// TransactionSet returns the transaction set the provided object
+// appears in.
+func (tp *TransactionPool) TransactionSet(oid crypto.Hash) []types.Transaction {
+	tp.mu.RLock()
+	defer tp.mu.RUnlock()
+	var parents []types.Transaction
+	tSetID, exists := tp.knownObjects[ObjectID(oid)]
+	if !exists {
+		return parents
```
DavidVorick: more explicit to return `nil` here.

modules/transactionpool/transactionpool.go
```diff
@@ -256,6 +256,26 @@ func (tp *TransactionPool) Transaction(id types.TransactionID) (types.Transactio
...
+	}
+	tSet, exists := tp.transactionSets[tSetID]
+	if !exists {
+		return parents
```
DavidVorick: also here

modules/transactionpool/transactionpool.go
```diff
@@ -256,6 +256,26 @@ func (tp *TransactionPool) Transaction(id types.TransactionID) (types.Transactio
...
+	}
+	for _, txn := range tSet {
```
DavidVorick: `parents = append(parents, tSet...)`

modules/renter/uploadheap.go
```diff
@@ -167,6 +168,10 @@ func (r *Renter) buildUnfinishedChunks(f *file, hosts map[string]struct{}) []*un
 	for fcid, fileContract := range f.contracts {
 		recentContract, exists := r.hostContractor.ContractByID(fcid)
 		contractUtility, exists2 := r.hostContractor.ContractUtility(fcid)
+		if (exists && !exists2) || (!exists && exists) {
```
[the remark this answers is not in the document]
ChrisSchinnerl: lol that's embarassing. I'm dying. :laughing:

modules/wallet/transactionbuilder.go
```diff
@@ -382,6 +382,26 @@ func (tb *transactionBuilder) FundSiafunds(amount types.Currency) error {
 	return nil
 }
 
+// UnconfirmedParents returns any unconfirmed parents the transaction set that
+// is being built by the transaction builder could have.
```
DavidVorick: this sentence is grammatically confusing

modules/wallet/transactionbuilder.go
```diff
@@ -382,6 +382,26 @@ func (tb *transactionBuilder) FundSiafunds(amount types.Currency) error {
...
+	addedParents := make(map[types.TransactionID]struct{})
+	for _, p := range tb.parents {
+		for _, sci := range p.SiacoinInputs {
+			tSet := tb.wallet.tpool.TransactionSet(crypto.Hash(sci.ParentID))
+			for _, txn := range tSet {
+				txnID := txn.ID()
+				if _, exists := addedParents[txnID]; exists {
+					continue
+				}
+				addedParents[txnID] = struct{}{}
```
DavidVorick: technically this is over-generous I think? You are adding every txn in the set, which may include children txns as well.

I think that it's probably okay for what we're doing, especially if the txn isn't completely published yet, but if the txn is completely published and you are still using the builder then you could get extra txns.

not sure if we want to handle that

tbenz9:
@ChrisSchinnerl any benchmarks showing how much improvement this PR has?  Something like: forming contracts used to take 30min now it takes 10min.

DavidVorick:
@tbenz9 no official benchmarks yet (on the way), but it's on the order of 30 minutes -> 2 minutes.

ChrisSchinnerl:
@tbenz9 Before this PR we used a 60s sleep after each contract to make sure it propagates through the network. This means for 50 contracts we waited at least 49 minutes. 
Now that this sleep is gone the contractor should be able to form a new contract every couple of seconds depending on the response time of the hosts.

PR #2856, opened 2018-03-12
move TestRenterSiapathValidate to more appropriate location

tbenz9:
This test is for more than just the upload code now.  Moving it to a more appropriate place.


DavidVorick:
While you are at it, can you add some extra tests?

`.../test`
`test./test`
`test../test`
`test/test./test`
`test/test../test`

I believe that all of those examples should be valid, but before we merged an update they would have produced 'invalid' errors.

tbenz9:
Added, and tested locally:
```
tbenz9@spartan1:~/go/src/github.com/NebulousLabs/Sia$ go test -short ./modules/renter/ -run=.
ok      github.com/NebulousLabs/Sia/modules/renter      0.005s
```

Issue #2262, opened 2017-08-26
Locked Collateral exceeds collateralbudget


chargr:
A recently reduced my collateralbudget from 35ks to 32.5ks while my locked collat was at 32.4ks
Since doing this it has risen well above the budget.

```
Host Internal Settings:
	acceptingcontracts:   Yes
	maxduration:          25 Weeks
	maxdownloadbatchsize: 68.16 MB
	maxrevisebatchsize:   68.16 MB
	netaddress:           x:9982 (manually specified)
	windowsize:           24 Hours

	collateral:       165 SC / TB / Month
	collateralbudget: 32.5 KS
	maxcollateral:    250 SC Per Contract

	mincontractprice:          5 SC
	mindownloadbandwidthprice: 80 SC / TB
	minstorageprice:           165 SC / TB / Month
	minuploadbandwidthprice:   5 SC / TB

Host Financials:
	Contract Count:               213
	Transaction Fee Compensation: 3.045 KS
	Potential Fee Compensation:   13.36 KS
	Transaction Fee Expenses:     0 H

	Storage Revenue:           1.644 KS
	Potential Storage Revenue: 474.4 SC

	Locked Collateral: 32.79 KS
	Risked Collateral: 438.2 SC
	Lost Collateral:   0 H

	Download Revenue:           8.31 SC
	Potential Download Revenue: 6.107 SC
	Upload Revenue:             20.84 SC
	Potential Upload Revenue:   4.117 SC
```

I do see plenty of these in my host.log which is what i would expect.
```
incoming RPCRenewContract failed: verification of renewal failed: host has reached its collateral budget and cannot accept the file contract
```

Is this just a cosmetic error?

Cinerar:
https://github.com/NebulousLabs/Sia/issues/2144

tbenz9:
I believe this is a known but and will be improved on in the next release.  Either way this looks like a bug to me.

Recommendation to devs: "Bug"

Issue #2144, opened 2017-07-15
Host financials


S-anasol:
![image](https://user-images.githubusercontent.com/1709666/28243163-e0ec94d2-69c8-11e7-8aec-90f0f5188f1d.png)

Almost all host stats incorrect.

Contract count fixed in latest RC, but other numbers also incorrect.

Used direct obligations output to calculate my stats.
https://github.com/NebulousLabs/Sia/compare/master...S-anasol:obligationsapi

So wallet show incorrectly(?)
* Locked Collateral
* Risked Collateral
* Potential storage revenue
* Potential upload revenue
* Transaction fees

Correctly
* Potential download revenue

S-anasol:
maybe better to re-calculate it with each start like contracts?
https://github.com/NebulousLabs/Sia/blob/master/modules/host/persist.go#L177

PR #2858, opened 2018-03-13
Re-calculate locked collateral value at start up

nielscastien:
This is a fix for the locked collateral counter that is off for all the hosts. At least at start up it gives the right information in the same way as the contract counter.

### DIFF
--- a/modules/host/persist.go
+++ b/modules/host/persist.go
@@ -172,9 +172,12 @@ func (h *Host) load() error {
 		return err
 	}
 
-	// Get the contract count by observing all of the incomplete storage
-	// obligations in the database.
+	// Get the contract count and locked collateral by observing all of the incomplete
+	// storage obligations in the database.
+	// TODO: both contract count and locked collateral are not correctly updated during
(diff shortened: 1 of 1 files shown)


nielscastien:
Band-aid fix for #2262 and #2144

DavidVorick:
oof.

@nielscastien can you confirm that it causes the host to produce more sane values for users that are seeing discrepencies?

nielscastien:
@DavidVorick Yes I can confirm, tested it on my own host and my locked collateral was reduced by 100 KS after restart.

nielscastien:
@DavidVorick Right now I am running a modified siad instance where the contract counter seems to be fixed. I did not receive any contract renewals today so I'm not 100% sure. I can create a pull request for that one as well if you would like to review it.

DavidVorick:
Happy to take a look at that code, at the very least.

One of our longer term to-dos is to give the host contracting system a lot of attention. We know it's pretty fragile right now but are focused fully on improving the renter module.

PR #2859, opened 2018-03-13
Rename ratelimit api fields

ChrisSchinnerl:


modules/renter/renter.go
```diff
@@ -285,7 +285,12 @@ func (r *Renter) SetSettings(s modules.RenterSettings) error {
 	if s.DownloadSpeed < 0 || s.UploadSpeed < 0 {
 		return errors.New("download/upload rate limit can't be below 0")
 	}
-	ratelimit.SetLimits(s.DownloadSpeed, s.UploadSpeed, s.PacketSize)
+	// Set ratelimit
+	if s.DownloadSpeed == 0 && s.UploadSpeed == 0 {
+		ratelimit.SetLimits(s.DownloadSpeed, s.UploadSpeed, 0)
```
lukechampine: maybe pass `0`s explicitly here for all values

modules/renter.go
```diff
@@ -164,7 +164,6 @@ type RenterPriceEstimation struct {
 // RenterSettings control the behavior of the Renter.
 type RenterSettings struct {
 	Allowance     Allowance `json:"allowance"`
-	PacketSize    uint64    `json:"packetsize"`
 	UploadSpeed   int64     `json:"uploadspeed"`
 	DownloadSpeed int64     `json:"downloadspeed"`
```
lukechampine: these fields should be renamed to match the API

Issue #2479, opened 2017-11-19
Feature: add upload timestamp/integrity checksum 


MeijeSibbel:
Bug Report
==========
Check to see if a bug report already exists.  If you still want to create a new issue please include the information below.

**Stack Trace or error message**

**Environment**
* Sia version:
* OS:

**Any useful logs**

Feature Request
===============
Check to see if the feature is on the roadmap.  If it is not on the roadmap then clearly describe your feature request and why it would benefit Sia.

[Sia roadmap](https://trello.com/b/Io1dDyuI/sia-public-roadmap)

tbenz9:
Related to #2109 

Recommendation to devs: either update the body of this issue to be more descriptive or close it as a duplicate of #2109.

DavidVorick:
Want to go ahead and include in this issue all common file metadata such as creation time, etc.

MSevey:
@MeijeSibbel @tbenz9 @DavidVorick how are you defining creation time versus upload timestamp?  Is the intent that upload timestamp is when the upload was initiated and the creation time is when the file has reached it's full redundancy?  Or is that a duplication in the request?

tbenz9:
I was imagining Sia would capture the Linux atime, ctime, and mtime timestamp independent of the upload and last-download timestamps.

For example, I take a picture on 1/1/2018, rename that photo on 2/2/2018, view the file on 3/3/2018, upload it to Sia on 4/4/2018, download from Sia on 5/5/2018.  It would have a:
creation time of 1/1/2018
modify time of 2/2/2018
access time of 3/3/2018
upload time of 4/4/2018
last download time of 5/5/2018

My opinion is upload time should be when the file is actually submitted to Sia for upload, not necessarily when it reaches full redundancy. If we wanted to get fancy I guess we could have as many as 3 upload timestamps (submitted, 1x, 3x redundancy).

MSevey:
thanks @tbenz9 
and I saw your commit on adding the upload and mod times, were you planning on continuing to finish those edits and submit a PR?  If not I can take what you started with and continue on from there.  LMK

tbenz9:
I am not planning on finishing #2860.  Once I began to understand the scope of the PR I chickened out and decided I didn't have the skills to write the compatibility code for the on-disk .sia file changes.  Please feel free to pick up where I left off.

I'll add that if there is any other relevant metadata returned by `stat` for a file this would be an ideal time to add it.

tbenz9:
[...]

PR #2860, opened 2018-03-13
[WIP] Add uploadTime and modTime metadata to files

tbenz9:
uploadTime is a time.Now field for when the file was uploaded to the Sia network.  modTime is a os.Stat ModTime field for when the file was last modified as reported by the client filesystem.  Feature Request #2479

### DIFF
--- a/modules/renter/files.go
+++ b/modules/renter/files.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"path/filepath"
 	"sync"
+	"time"
 
 	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/crypto"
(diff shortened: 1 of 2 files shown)


tbenz9:
@lukechampine Can you tell me if I'm on the right path with this PR.  If you give me the go ahead I'll add these field to the API endpoint, `siac renter files -v` output, and update the documentation.

lukechampine:
Doing this properly would mean adding `uploadTime` and `modTime` to the on-disk `.sia` format, which would imply adding compatibility code to convert old `.sia` files to the new format. I'm open to it, but it's a fair amount of work. Might be better to wait for the larger overhaul of the renter (Coming Soon™)

tbenz9:
I think that's more than I want to tackle right now.  Hopefully renter overhaul comes soon!  I know several users are looking for this kind of metadata.

Issue #2863, opened 2018-03-14
panic: Critical error: negative currency not allowed


gls4:
Wallet: scanned to height 145517...
Done!
goroutine 215328 [running]:
runtime/debug.Stack(0x445d87, 0x0, 0xc42142fef8)
        /usr/local/go/src/runtime/debug/stack.go:24 +0xa7
runtime/debug.PrintStack()
        /usr/local/go/src/runtime/debug/stack.go:16 +0x22
github.com/NebulousLabs/Sia/build.Critical(0xc42142ffb0, 0x1, 0x1)
        /root/Sia/src/github.com/NebulousLabs/Sia/build/critical.go:16 +0x99
github.com/NebulousLabs/Sia/types.Currency.Sub(0x0, 0xc421b9c210, 0x2, 0x6, 0x0, 0xc42037a140, 0x2, 0x7, 0xc421b9cf00, 0xc421b9c210, ...)
        /root/Sia/src/github.com/NebulousLabs/Sia/types/currency.go:179 +0x12e
github.com/NebulousLabs/Sia/modules/host.renewContractCollateral(0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0xc422990d80, 0x2, ...)
        /root/Sia/src/github.com/NebulousLabs/Sia/modules/host/negotiaterenewcontract.go:44 +0xa6
github.com/NebulousLabs/Sia/modules/host.(*Host).managedVerifyRenewedContract(0xc42010e000, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0xc422990d80, ...)
        /root/Sia/src/github.com/NebulousLabs/Sia/modules/host/negotiaterenewcontract.go:266 +0x61b
github.com/NebulousLabs/Sia/modules/host.(*Host).managedRPCRenewContract(0xc42010e000, 0xb5e240, 0xc42000e178, 0x0, 0x0)
        /root/Sia/src/github.com/NebulousLabs/Sia/modules/host/negotiaterenewcontract.go:126 +0x41d
github.com/NebulousLabs/Sia/modules/host.(*Host).threadedHandleConn(0xc42010e000, 0xb5e240, 0xc42000e178)
        /root/Sia/src/github.com/NebulousLabs/Sia/modules/host/network.go:286 +0x7cb
created by github.com/NebulousLabs/Sia/modules/host.(*Host).threadedListen
[...]

PR #2865, opened 2018-03-14
Proposal: Sub underflow results in ZeroCurrency instead of noop

ChrisSchinnerl:
There have been multiple reported issues caused by a Currency underflow. I think we have fixed it twice by simply setting the currency to 0.
Issue [#2863](https://github.com/NebulousLabs/Sia/issues/2863) brought this to my attention once more and this time the consequence of not returning 0 in case of an underflow might be more severe. The host might post too much collateral and later fail its own verification when the renter uploads/changes data. (could be related to the issue of Fornax on discord)
By changing the behavior of `Sub` we should be notified when an underflow happens and can still expect somewhat sane behavior.

### DIFF
--- a/types/currency.go
+++ b/types/currency.go
@@ -175,7 +175,7 @@ func (x Currency) Sqrt() (c Currency) {
 // x < y.
 func (x Currency) Sub(y Currency) (c Currency) {
 	if x.Cmp(y) < 0 {
-		c = x
+		c = ZeroCurrency
 		build.Critical(ErrNegativeCurrency)
 	} else {
 		c.i.Sub(&x.i, &y.i)


DavidVorick:
ACK

lukechampine:
This makes me a little uneasy. But I guess if it's already spewing warnings, the actual subtraction behavior isn't too important.

We might want to add a `SubSat` method for when saturation behavior is desired.

DavidVorick:
The question here is whether we get more bugs in the long term by ignoring / no-op'ing subtractions that fail and result in bugs, or whether we get more bugs in the long term by zero-ing out subtractions that fail as a result of bugs.

My guess is that we're a lot more likely to run into problems if a failed subtraction returns the original value instead of zero.

It's not something that's ever supposed to happen anyway, but in practice it happens sometimes, particularly anywhere where we are doing long term accounting (such as the renter's allowance, and the host's historic revenue)

ChrisSchinnerl:
I agree. I think it is a lot more likely that we want to return zero instead of the original value. 
We should still find out why this happens sometimes in the host though.

lukechampine:
Fair enough.

PR #2871, opened 2018-03-15
Fix check on WindowEnd

nielscastien:


nielscastien:
This fix prevents a renter from forming a new contract with a too long duration. As a hoster I would like to have this fixed asap ;-) @lukechampine

lukechampine:
good catch!

DavidVorick:
I don't think this is correct

DavidVorick:
@lukechampine please be careful about merging anything related to verification or protocols, this stuff is security sensitive and I think this PR introduces a bug

PR #2872, opened 2018-03-15
Implement io.ReadSeeker interface for downloader

ChrisSchinnerl:
This PR adds the `/renter/stream` endpoint to the renter which is basically another download endpoint that uses the `http.ServeContent` method to serve a resource. This also allows for opening and streaming pretty much every browser supported file in the browser. (also videos).
This is a WIP, contains some nasty caching and disables the `Sia-Agent` check to make it work in a browser but I would appreciate some overall reviews.

### DIFF
--- a/modules/renter.go
+++ b/modules/renter.go
@@ -343,6 +343,10 @@ type Renter interface {
 	// ShareFilesAscii creates an ASCII-encoded '.sia' file.
 	ShareFilesASCII(paths []string) (asciiSia string, err error)
 
+	// Streamer creates a io.ReadSeeker that can be used to stream downloads
+	// from the Sia network.
+	Streamer(siaPath string) (string, io.ReadSeeker, error)
+
 	// Upload uploads a file using the input parameters.
 	Upload(FileUploadParams) error
 }
--- a/modules/renter.go
+++ b/modules/renter.go
@@ -344,7 +344,8 @@ type Renter interface {
 	ShareFilesASCII(paths []string) (asciiSia string, err error)
 
 	// Streamer creates a io.ReadSeeker that can be used to stream downloads
-	// from the Sia network.
+	// from the Sia network and also returns the fileName of the streamed
+	// resource.
 	Streamer(siaPath string) (string, io.ReadSeeker, error)
 
 	// Upload uploads a file using the input parameters.
--- a/modules/renter/downloadcache.go
+++ b/modules/renter/downloadcache.go
@@ -1,22 +1,34 @@
 package renter
 
-import "time"
+// TODO expose the downloadCacheSize as a variable and allow users to set it
+// via the API.
+
+import (
+	"time"
+
+	"github.com/NebulousLabs/errors"
+)
 
 // addChunkToCache adds the chunk to the cache if the download is a streaming
 // endpoint download.
+// TODO this won't be necessary anymore once we have partial downloads.
 func (udc *unfinishedDownloadChunk) addChunkToCache(data []byte) {
 	if udc.download.staticDestinationType == destinationTypeSeekStream {
-		udc.cacheMu.Lock()
-		// Prune cache if necessary.
-		for key := range udc.chunkCache {
-			if len(udc.chunkCache) < downloadCacheSize {
-				break
-			}
-			delete(udc.chunkCache, key)
-		}
-		udc.chunkCache[udc.staticCacheID] = data
-		udc.cacheMu.Unlock()
+		// We only cache streaming chunks since browsers and media players tend to only request a few kib at once when streaming data. That way we can prevent scheduling the same chunk for download over and over.
+		return
 	}
+	udc.cacheMu.Lock()
+	// Prune cache if necessary.
+	// TODO insteado of deleting a 'random' key, delete the
+	// least-recently-accessed element of the cache.
+	for key := range udc.chunkCache {
+		if len(udc.chunkCache) < downloadCacheSize {
+			break
+		}
+		delete(udc.chunkCache, key)
+	}
+	udc.chunkCache[udc.staticCacheID] = data
+	udc.cacheMu.Unlock()
 }
 
 // managedTryCache tries to retrieve the chunk from the renter's cache. If
@@ -38,7 +50,9 @@ func (r *Renter) managedTryCache(udc *unfinishedDownloadChunk) bool {
 	end := start + udc.staticFetchLength
 	_, err := udc.destination.WriteAt(data[start:end], udc.staticWriteOffset)
 	if err != nil {
-		r.log.Println("WARN: failed to write cached chunk to destination")
+		r.log.Println("WARN: failed to write cached chunk to destination:", err)
+		udc.fail(errors.AddContext(err, "failed to write cached chunk to destination"))
+		return true
 	}
 
 	// Check if the download is complete now.
--- a/modules/renter/downloadcache.go
+++ b/modules/renter/downloadcache.go
@@ -1,0 +1,53 @@
+package renter
+
+import "time"
+
+// addChunkToCache adds the chunk to the cache if the download is a streaming
+// endpoint download.
+func (udc *unfinishedDownloadChunk) addChunkToCache(data []byte) {
+	if udc.download.staticDestinationType == destinationTypeSeekStream {
+		udc.cacheMu.Lock()
+		// Prune cache if necessary.
+		for key := range udc.chunkCache {
+			if len(udc.chunkCache) < downloadCacheSize {
+				break
+			}
+			delete(udc.chunkCache, key)
+		}
+		udc.chunkCache[udc.staticCacheID] = data
+		udc.cacheMu.Unlock()
+	}
+}
+
+// managedTryCache tries to retrieve the chunk from the renter's cache. If
+// successful it will write the data to the destination and stop the download
+// if it was the last missing chunk. The function returns true if the chunk was
+// in the cache.
+// TODO in the future we might need cache invalidation. At the
+// moment this doesn't worry us since our files are static.
+func (r *Renter) managedTryCache(udc *unfinishedDownloadChunk) bool {
+	udc.mu.Lock()
+	defer udc.mu.Unlock()
+	r.cmu.Lock()
+	data, cached := r.chunkCache[udc.staticCacheID]
+	r.cmu.Unlock()
+	if !cached {
+		return false
+	}
+	start := udc.staticFetchOffset
+	end := start + udc.staticFetchLength
+	_, err := udc.destination.WriteAt(data[start:end], udc.staticWriteOffset)
+	if err != nil {
+		r.log.Println("WARN: failed to write cached chunk to destination")
+	}
+
+	// Check if the download is complete now.
+	udc.download.mu.Lock()
+	udc.download.chunksRemaining--
+	if udc.download.chunksRemaining == 0 {
+		udc.download.endTime = time.Now()
+		close(udc.download.completeChan)
+	}
+	udc.download.mu.Unlock()
+	return true
+}
--- a/modules/renter/downloadchunk.go
+++ b/modules/renter/downloadchunk.go
@@ -211,10 +211,24 @@ func (udc *unfinishedDownloadChunk) threadedRecoverLogicalData() error {
 		udc.physicalChunkData[i] = nil
 	}
 
+	// Add the chunk to the cache.
+	// TODO this should only happen for the streaming endpoint.
+	recoveredData := recoverWriter.Bytes()
+	cmu.Lock()
+	cacheID := fmt.Sprintf("%v:%v", udc.download.staticSiaPath, udc.staticChunkIndex)
+	for key := range cache {
+		if len(cache) < 5 {
+			break
+		}
+		delete(cache, key)
+	}
+	cache[cacheID] = recoveredData
+	cmu.Unlock()
+
 	// Write the bytes to the requested output.
 	start := udc.staticFetchOffset
 	end := udc.staticFetchOffset + udc.staticFetchLength
-	_, err = udc.destination.WriteAt(recoverWriter.Bytes()[start:end], udc.staticWriteOffset)
+	_, err = udc.destination.WriteAt(recoveredData[start:end], udc.staticWriteOffset)
 	if err != nil {
 		udc.mu.Lock()
 		udc.fail(err)
--- a/modules/renter/downloadchunk.go
+++ b/modules/renter/downloadchunk.go
@@ -43,6 +43,7 @@ type unfinishedDownloadChunk struct {
 
 	// Fetch + Write instructions - read only or otherwise thread safe.
 	staticChunkIndex  uint64                                     // Required for deriving the encryption keys for each piece.
+	staticCacheID     string                                     // Used to uniquely identify a chunk in the chunk cache.
 	staticChunkMap    map[types.FileContractID]downloadPieceInfo // Maps from file contract ids to the info for the piece associated with that contract
 	staticChunkSize   uint64
 	staticFetchLength uint64 // Length within the logical chunk to fetch.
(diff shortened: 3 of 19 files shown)


DavidVorick:
From a security perspective, disabling the user agent for this endpoint means that any javascript page is able to start streaming your files from Sia as long as they know the filenames. That is going to spend your money, and potentially violate your privacy.

I am not comfortable with that, but I don't have a great alternative either.

modules/renter/downloadchunk.go
```diff
@@ -72,6 +73,10 @@ type unfinishedDownloadChunk struct {
 	// The download object, mostly to update download progress.
 	download *download
 	mu       sync.Mutex
+
+	// Caching related fields
+	chunkCache map[string][]byte
+	cmu        *sync.Mutex
```
DavidVorick: need to name this cacheMu, cmu isn't sufficiently readable if you haven't memorized the struct.

modules/renter/downloadchunk.go
```diff
@@ -166,6 +171,24 @@ func (udc *unfinishedDownloadChunk) returnMemory() {
 	}
 }
 
+// addChunkToCache adds the chunk to the cache if the download is a streaming
+// endpoint download.
+func (udc *unfinishedDownloadChunk) addChunkToCache(data []byte) {
+	// TODO is it safe to read the type without a lock?
```
DavidVorick: naming convention is to call it 'static' if the field doesn't change. If you want to read it without holding a lock, you'll need to change it to static, then go through the code and check every place the field is used and make sure that it only ever changes at init.

lukechampine:
I haven't tested this, but perhaps we could use the basic auth form of URLs to secure streaming?
e.g.
`mpv http://user:password@localhost:9980/renter/stream/foo`

modules/renter/downloadchunk.go
```diff
@@ -166,6 +171,24 @@ func (udc *unfinishedDownloadChunk) returnMemory() {
 	}
 }
 
+// addChunkToCache adds the chunk to the cache if the download is a streaming
+// endpoint download.
+func (udc *unfinishedDownloadChunk) addChunkToCache(data []byte) {
```
DavidVorick: This should probably live in cache.go - we aren't sure how long term the cache is going to be, we should make it easy to track it down and remove it, or upgrade it.

modules/renter/downloadheap.go
```diff
@@ -153,6 +153,39 @@ func (r *Renter) managedNextDownloadChunk() *unfinishedDownloadChunk {
 	}
 }
 
+// managedTryCache tries to retrieve the chunk from the renter's cache. If
+// successful it will write the data to the destination and stop the download
+// if it was the last missing chunk. The function returns true if the chunk was
+// in the cache.
+// TODO in the future we might need cache invalidation. At the
+// moment this doesn't worry us since our files are static.
+func (r *Renter) managedTryCache(udc *unfinishedDownloadChunk) bool {
```
DavidVorick: this also to cache.go

DavidVorick: should probably call it downloadcache.go

ChrisSchinnerl:
@DavidVorick JavaScript shouldn't be an issue in a modern browser since it would be blocked due to the different origin. You could put a link on the website though and make the user start the stream itself. I think in that case the user would notice though.

siatest/renter.go
```diff
@@ -58,6 +58,15 @@ func (tn *TestNode) DownloadByStream(rf *RemoteFile) (data []byte, err error) {
...
 }
 
+// Stream uses the streaming endpoint to download a file.
+func (tn *TestNode) Stream(rf *RemoteFile) (data []byte, err error) {
+	data, err = tn.RenterStreamGet(rf.siaPath)
+	if err == nil && rf.checksum != crypto.HashAll(data) {
+		err = errors.New("downloaded bytes don't match requested data")
+	}
+	return
+}
```
DavidVorick: Stream function ideally jumps around a little bit as well, to mimic a user seeking. You'll need a checksum of each part of the file of course to do that.

node/api/routes.go
```diff
@@ -166,7 +167,7 @@ func cleanCloseHandler(next http.Handler) http.Handler {
 // UserAgent that contains the specified string.
 func RequireUserAgent(h http.Handler, ua string) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
-		if !strings.Contains(req.UserAgent(), ua) {
+		if !strings.Contains(req.UserAgent(), ua) && !whitelisted(req) {
```
DavidVorick: I think this is fine for now, though performance we should probably wrap each endpoint individually the same way that we do with `RequirePassword`, that way requests which don't require a user agent don't force us to do a strings.Contains every time another request is called.

With just one whitelisted endpoint, this shouldn't cause any performance issues, but if we start to adopt this as a pattern, it's a pattern that could cause problems if it gets out of hand.

modules/renter/streamer.go
```diff
@@ -0,0 +1,141 @@
+package renter
```
DavidVorick: filename should probably be downloadstreamer

modules/renter/streamer.go
```diff
@@ -0,0 +1,141 @@
...
+	min := uint64(math.MaxUint64)
+	for _, v := range values {
+		if v < min {
+			min = v
+		}
+	}
+	return min
+}
+
+// Streamer create a io.ReadSeeker that can be used to stream downloads from
```
DavidVorick: should be `Streamer creates a`

modules/renter/streamer.go
```diff
@@ -0,0 +1,141 @@
...
+// the sia network.
+func (r *Renter) Streamer(siaPath string) (string, io.ReadSeeker, error) {
+	// Lookup the file associated with the nickname.
+	lockID := r.mu.RLock()
+	file, exists := r.files[siaPath]
+	r.mu.RUnlock(lockID)
+	if !exists {
```
DavidVorick: `if !exists || file.deleted {`

modules/renter/streamer.go
```diff
@@ -0,0 +1,141 @@
...
+	case io.SeekCurrent:
+		newOffset = s.offset
+	case io.SeekEnd:
+		s.file.mu.RLock()
+		newOffset = int64(s.file.size)
+		s.file.mu.RUnlock()
+	}
+	newOffset += offset
+
+	if newOffset < 0 {
```
DavidVorick: should we be checking for overflow here as well?

DavidVorick: file.Seek I believe won't return an error, but that's because you can call `file.Write` at the end of a file even if you are past the last piece of already written data.

@lukechampine do you have thoughts here?

modules/renter/streamer.go
```diff
@@ -0,0 +1,141 @@
...
+	if s.offset >= fileSize {
+		return 0, io.EOF
+	}
+
+	// Calculate how much we can download. We never download more than a single chunk.
+	chunkSize := s.file.staticChunkSize()
+	remainingData := uint64(fileSize - s.offset)
+	requestedData := uint64(len(p))
+	remainingChunk := chunkSize - uint64(s.offset)%chunkSize
+	length := min(chunkSize, remainingData, requestedData, remainingChunk)
```
DavidVorick: chunkSize does not need to be included in the `min` function, because `remainingChunk` is a fundamental subset of that.

modules/renter/streamer.go
```diff
@@ -0,0 +1,141 @@
...
+	})
+	if err != nil {
+		return 0, errors.AddContext(err, "failed to create new download")
+	}
+
+	// Add the download object to the download queue.
+	// TODO: Maybe this is not necessary for streams?
+	s.r.downloadHistoryMu.Lock()
+	s.r.downloadHistory = append(s.r.downloadHistory, d)
+	s.r.downloadHistoryMu.Unlock()
```
DavidVorick: You are going to heavily clutter the download history if you append an item each time that `Read` is called. I would say that the best option is to add an item to `downloadHistory` when the `Streamer` function is called, and then otherwise we store the object in the `streamer` struct and just modify it to reflect that more data has been requested.

It is not typical for web browsers to add a streamed file to your download history (e.g. your download history doesn't extend if you watch a youtube video), which would be a solid argument towards not adding streaming items to the download history at all.

Since we're stuck with compatibility requirements if we add it, let's not add the streams to the download history for now, we can always add them in later if we decide that it's something that should be appearing in the download history.

modules/renter/streamer.go
```diff
@@ -0,0 +1,141 @@
...
+
+	// Download data
+	buffer := bytes.NewBuffer([]byte{})
+	d, err := s.r.newDownload(downloadParams{
+		destination:       newDownloadDestinationWriteCloserFromWriter(buffer),
+		destinationType:   destinationTypeSeekStream,
+		destinationString: "httpresponse",
+		file:              s.file,
+
+		latencyTarget: 25e3 * time.Millisecond, // TODO high default until full latency suport is added.
```
DavidVorick: Since the user is going to be blocking on this call until the download completes, a low latency target is going to be optimal. I'd place the default here at 50ms.

modules/renter/streamer.go
```diff
@@ -0,0 +1,141 @@
...
+		length:        length,
+		needsMemory:   true,
+		offset:        uint64(s.offset),
+		overdrive:     3,  // TODO: moderate default until full overdrive support is added.
```
DavidVorick: Since low latency is a top priority, we probably want to go heavier on the overdrive. I'm going to suggest a value of 5.

modules/renter/streamer.go
```diff
@@ -0,0 +1,141 @@
...
+		priority:      10, // TODO: high default until full priority support is added.
```
DavidVorick: I would put the priority at 1000, since it's a latency sensitive request directly from the user. That gives us lots of room for things that have a priority between "simple repair" and "streaming download"

DavidVorick: the priority I'm pretty sure is a uint64, so there's also plenty of room for things that have a higher priority.

modules/renter/downloadcache.go
```diff
@@ -0,0 +1,53 @@
+package renter
+
+import "time"
+
+// addChunkToCache adds the chunk to the cache if the download is a streaming
+// endpoint download.
+func (udc *unfinishedDownloadChunk) addChunkToCache(data []byte) {
+	if udc.download.staticDestinationType == destinationTypeSeekStream {
```
DavidVorick: anti-pattern - instead of having an `if` block that wraps the whole function, have an inverse that returns early.

```
if udc.download.staticDestinationType != destinationTypeSeekStream {
    // some comment explaining why we only cache destinationTypeSeekStream
    return
}
```

modules/renter/downloadcache.go
```diff
@@ -0,0 +1,53 @@
+package renter
+
+import "time"
+
+// addChunkToCache adds the chunk to the cache if the download is a streaming
+// endpoint download.
+func (udc *unfinishedDownloadChunk) addChunkToCache(data []byte) {
```
DavidVorick: Should add a TODO for this function that deletes the least-recently-accessed part of the cache instead of just deleting whatever map element is created first. No need to do that for this PR.

@tbenz9 this is probably a pretty beginner-friendly issue that we can let some newer contributor tackle.

modules/renter.go
```diff
@@ -343,6 +343,10 @@ type Renter interface {
 	// ShareFilesAscii creates an ASCII-encoded '.sia' file.
 	ShareFilesASCII(paths []string) (asciiSia string, err error)
 
+	// Streamer creates a io.ReadSeeker that can be used to stream downloads
+	// from the Sia network.
+	Streamer(siaPath string) (string, io.ReadSeeker, error)
```
lukechampine: What is the `string` return value? Should either be named or mentioned in the docstring.

modules/renter/streamer.go
lukechampine: and maybe `an`, if you pronounce the `io` when reading `io.ReadSeeker` :P

modules/renter/streamer.go
lukechampine: The `io.Seeker` docstring says:

>Seeking to an offset before the start of the file is an error. Seeking to any positive offset is legal, but the behavior of subsequent I/O operations on the underlying object is implementation-dependent. 

The standard library seems to be split on this. When seeking past the end of the stream, `strings.Reader` and `bytes.Reader` return `nil`, but future `Read` calls return `io.EOF`. On the other hand, `os.File` (and the more obscure `elf.readSeekerFromReader`) return an error immediately from `Seek`.

I'm slightly in favor of returning an error immediately, because seeking past the end does feel like an invalid operation that ought to return an error. The only downside that I can see is that people may be less likely to check `Seek` errors than `Read` errors. But any code that does responsible error checking should work fine with either approach.

node/api/routes.go
lukechampine: the advantage of a whitelist is that you get security by default, instead of needing to remember to add `RequirePassword` and `RequireUserAgent` manually every time you add a new endpoint.

Amazingly, I think this may be a place where `Context` comes in handy. It would allow us to whitelist routes by wrapping their handler in an `Unrestricted` function, like so:
```go
router.GET("/renter/stream/*siapath", Unrestricted(api.renterStreamHandler))

type unrestrictedContextKey struct {}

func Unrestricted(h http.Handler) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
		req = req.WithContext(context.WithValue(req.Context(), unrestrictedContextKey{}, 0))
		h.ServeHTTP(w, req)
	}
}

func isUnrestricted(req *http.Request) bool {
	return req.Context().Value(unrestrictedContextKey{}) != nil
}

func RequireUserAgent(h http.Handler, ua string) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
		if !strings.Contains(req.UserAgent(), ua) && !isUnrestricted(req) {
			WriteError(w, Error{"Browser access disabled due to security vulnerability. Use Sia-UI or siac."}, http.StatusBadRequest)
			return
		}
		h.ServeHTTP(w, req)
	})
}
```

siatest/renter.go
```diff
@@ -58,6 +58,15 @@ func (tn *TestNode) DownloadByStream(rf *RemoteFile) (data []byte, err error) {
 	return
 }
 
+// Stream uses the streaming endpoint to download a file.
+func (tn *TestNode) Stream(rf *RemoteFile) (data []byte, err error) {
+	data, err = tn.RenterStreamGet(rf.siaPath)
+	if err == nil && rf.checksum != crypto.HashAll(data) {
```
lukechampine: wait, why is `HashAll` being used here? For a `[]byte`, just use `crypto.HashBytes`.

modules/renter/streamer.go
ChrisSchinnerl: @lukechampine yeah that's what I did to decide on how to implement it. I first tried doing the same on an `os.File`.

modules/renter/downloadcache.go
```diff
@@ -0,0 +1,53 @@
+package renter
```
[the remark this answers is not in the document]
ChrisSchinnerl: Yeah that would be pretty cool. Especially in combination with the PR mentioned above. That would allow users to stream multiple videos in parallel efficiently if they have enough ram available.

siatest/renter/renter_test.go
```diff
@@ -75,13 +76,23 @@ func testUploadDownload(t *testing.T, tg *siatest.TestGroup) {
...
+	// Stream the file.
+	_, err = renter.Stream(remoteFile)
+	if err != nil {
+		t.Fatal(err)
+	}
+	// Stream the file partially.
+	_, err = renter.StreamPartial(remoteFile, localFile, 0, uint64(fileSize/2))
+	if err != nil {
+		t.Fatal(err)
+	}
```
DavidVorick: Would like to see a loop,  5 calls to StreamPartial, from a random starting point to a random ending point that is somewhere between the starting point and end of the file.

Otherwise this PR is looking pretty good.

PR #2877, opened 2018-03-16
Fix deleted renter file showing up after restart

ChrisSchinnerl:
When a file is deleted while being repaired, the upload of the scheduled chunks will continue. Since the upload code modifies the uploaded file's metadata, the deleted `.sia` file will reappear and after a restart of sia, it will also be tracked by the renter again.
This PR fixes this by introducing a `deleted` field that prevents deleted files from being saved to disk.

### DIFF
--- a/modules/renter/files.go
+++ b/modules/renter/files.go
@@ -36,6 +36,7 @@ type file struct {
 	erasureCode modules.ErasureCoder // Static - can be accessed without lock.
 	pieceSize   uint64               // Static - can be accessed without lock.
 	mode        uint32               // actually an os.FileMode
+	deleted     bool                 // indicates if the file has been deleted.
 
 	staticUID string // A UID assigned to the file when it gets created.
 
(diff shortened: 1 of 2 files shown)


DavidVorick:
This is too quick of a solution, and one I suspect would come back to bite us in some way.

We want to distinguish at the chunk level, including whether a chunk has been deleted or modified. We need some identifier for chunks which is only valid so long as the data remains unchanged. If the chunk status, chunk data, or file status changes, the chunk needs to be invalidated.

Though, I'm not super certain on the mechanics there. Because if you invalidate a chunk halfway through uploading, you end up with some garbage sectors lying around on the hosts, which makes me think we could be a bit more careful about how we orchestrate that.

ChrisSchinnerl:
@DavidVorick the change consists only of 7 lines of code and they don't directly touch the upload code. I think we won't change the upload code to support deleting single chunks anytime soon and when we do, removing those 7 lines shouldn't be a big effort.

PR #2879, opened 2018-03-16
Remove renter's merkle roots from ram

ChrisSchinnerl:
Instead of having all the merkle roots of every contract in memory, the renter should load them from disk if needed.

### DIFF
--- a/modules/renter/proto/consts.go
+++ b/modules/renter/proto/consts.go
@@ -11,6 +11,10 @@ import (
 const (
 	// contractExtension is the extension given to contract files.
 	contractExtension = ".contract"
+
+	// rootsDiskLoadBulkSize is the max number of roots we read from disk at
+	// once to avoid using up all the ram.
+	rootsDiskLoadBulkSize = 1024 * crypto.HashSize // 32 kib
 )
 
 var (
--- a/modules/renter/proto/consts.go
+++ b/modules/renter/proto/consts.go
@@ -15,6 +15,10 @@ const (
 	// rootsDiskLoadBulkSize is the max number of roots we read from disk at
 	// once to avoid using up all the ram.
 	rootsDiskLoadBulkSize = 1024 * crypto.HashSize // 32 kib
+
+	// remainingFile is a constant used to indicate that a fileSection can access
+	// the whole remaining file instead of being boung to a certain end offset.
+	remainingFile = -1
 )
 
 var (
--- a/modules/renter/proto/consts.go
+++ b/modules/renter/proto/consts.go
@@ -17,7 +17,7 @@ const (
 	rootsDiskLoadBulkSize = 1024 * crypto.HashSize // 32 kib
 
 	// remainingFile is a constant used to indicate that a fileSection can access
-	// the whole remaining file instead of being boung to a certain end offset.
+	// the whole remaining file instead of being bound to a certain end offset.
 	remainingFile = -1
 )
 
--- a/modules/renter/proto/contract.go
+++ b/modules/renter/proto/contract.go
@@ -133,6 +133,13 @@ type SafeContract struct {
 	// applied to the contract file.
 	unappliedTxns []*writeaheadlog.Transaction
 
+	// file is the file of the safe contract that contains the roots. This
+	// file is usually shared with the SafeContract which means multiple
+	// threads could potentially write to the same file. That's why the
+	// SafeContract should never modify the file beyond the
+	// contractHeaderSize and the merkleRoots should never modify data
+	// before that. Both should also use WriteAt and ReadAt instead of
+	// Write and Read.
 	f   *os.File // TODO: use a dependency for this
 	wal *writeaheadlog.WAL
 	mu  sync.Mutex
--- a/modules/renter/proto/contract.go
+++ b/modules/renter/proto/contract.go
@@ -2,12 +2,10 @@ package proto
 
 import (
 	"encoding/json"
-	"io"
 	"os"
 	"path/filepath"
 	"sync"
 
-	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/encoding"
 	"github.com/NebulousLabs/Sia/modules"
@@ -128,10 +126,8 @@ type SafeContract struct {
 	headerMu sync.Mutex
 	header   contractHeader
 
-	// merkleRoots are the Merkle roots of each sector stored on the host that
-	// relate to this contract.
-	//merkleRoots []crypto.Hash
-	numMerkleRoots int
+	// merkleRoots are the sector roots covered by this contract.
+	merkleRoots *merkleRoots
 
 	// unappliedTxns are the transactions that were written to the WAL but not
 	// applied to the contract file.
@@ -208,28 +204,6 @@ func (c *SafeContract) Utility() modules.ContractUtility {
 	return c.header.Utility
 }
 
-// merkleRoots returns the contracts merkle roots.
-func (c *SafeContract) merkleRoots() ([]crypto.Hash, error) {
-	merkleRoots := make([]crypto.Hash, 0, c.numMerkleRoots)
-	if _, err := c.f.Seek(contractHeaderSize, io.SeekStart); err != nil {
-		return merkleRoots, err
-	}
-	for {
-		var root crypto.Hash
-		if _, err := io.ReadFull(c.f, root[:]); err == io.EOF {
-			break
-		} else if err != nil {
-			return merkleRoots, errors.AddContext(err, "failed to read root from disk")
-		}
-		merkleRoots = append(merkleRoots, root)
-	}
-	// Sanity check: should have read exactly numMerkleRoots roots.
-	if len(merkleRoots) != c.numMerkleRoots {
-		build.Critical("Number of merkle roots on disk doesn't match numMerkleRoots")
-	}
-	return merkleRoots, nil
-}
-
 func (c *SafeContract) makeUpdateSetHeader(h contractHeader) writeaheadlog.Update {
 	c.headerMu.Lock()
 	id := c.header.ID()
@@ -274,8 +248,11 @@ func (c *SafeContract) applySetRoot(root crypto.Hash, index int) error {
 	if _, err := c.f.WriteAt(root[:], rootOffset); err != nil {
 		return err
 	}
-	if c.numMerkleRoots <= index {
-		c.numMerkleRoots++
+	numMerkleRoots := c.merkleRoots.len()
+	if index == numMerkleRoots {
+		c.merkleRoots.push(root)
+	} else {
+		return c.merkleRoots.insert(index, root)
 	}
 	return nil
 }
@@ -292,7 +269,7 @@ func (c *SafeContract) recordUploadIntent(rev types.FileContractRevision, root c
 
 	t, err := c.wal.NewTransaction([]writeaheadlog.Update{
 		c.makeUpdateSetHeader(newHeader),
-		c.makeUpdateSetRoot(root, c.numMerkleRoots),
+		c.makeUpdateSetRoot(root, c.merkleRoots.len()),
 	})
 	if err != nil {
 		return nil, err
@@ -316,7 +293,7 @@ func (c *SafeContract) commitUpload(t *writeaheadlog.Transaction, signedTxn type
 	if err := c.applySetHeader(newHeader); err != nil {
 		return err
 	}
-	if err := c.applySetRoot(root, c.numMerkleRoots); err != nil {
+	if err := c.applySetRoot(root, c.merkleRoots.len()); err != nil {
 		return err
 	}
 	if err := c.f.Sync(); err != nil {
@@ -441,8 +418,9 @@ func (cs *ContractSet) managedInsertContract(h contractHeader, roots []crypto.Ha
 		return modules.RenterContract{}, err
 	}
 	// write roots
-	for i, root := range roots {
-		if _, err := f.WriteAt(root[:], contractHeaderSize+crypto.HashSize*int64(i)); err != nil {
+	merkleRoots := newMerkleRoots(f)
+	for _, root := range roots {
+		if err := merkleRoots.push(root); err != nil {
 			return modules.RenterContract{}, err
 		}
 	}
@@ -450,10 +428,10 @@ func (cs *ContractSet) managedInsertContract(h contractHeader, roots []crypto.Ha
 		return modules.RenterContract{}, err
 	}
 	sc := &SafeContract{
-		header:         h,
-		numMerkleRoots: len(roots),
-		f:              f,
-		wal:            cs.wal,
+		header:      h,
+		merkleRoots: merkleRoots,
+		f:           f,
+		wal:         cs.wal,
 	}
 	cs.mu.Lock()
 	cs.contracts[h.ID()] = sc
@@ -476,18 +454,9 @@ func (cs *ContractSet) loadSafeContract(filename string, walTxns []*writeaheadlo
 		return err
 	}
 	// read merkleRoots
-	numMerkleRoots := 0
-	if _, err := f.Seek(contractHeaderSize, io.SeekStart); err != nil {
+	merkleRoots, err := loadExistingMerkleRoots(f)
+	if err != nil {
 		return err
-	}
-	for {
-		var root crypto.Hash
-		if _, err := io.ReadFull(f, root[:]); err == io.EOF {
-			break
-		} else if err != nil {
-			return err
-		}
-		numMerkleRoots++
 	}
 	// add relevant unapplied transactions
 	var unappliedTxns []*writeaheadlog.Transaction
@@ -518,11 +487,11 @@ func (cs *ContractSet) loadSafeContract(filename string, walTxns []*writeaheadlo
 	}
 	// add to set
 	cs.contracts[header.ID()] = &SafeContract{
-		header:         header,
-		numMerkleRoots: numMerkleRoots,
-		unappliedTxns:  unappliedTxns,
-		f:              f,
-		wal:            cs.wal,
+		header:        header,
+		merkleRoots:   merkleRoots,
+		unappliedTxns: unappliedTxns,
+		f:             f,
+		wal:           cs.wal,
 	}
 	return nil
 }
@@ -551,7 +520,7 @@ func (cs *ContractSet) ConvertV130Contract(c V130Contract, cr V130CachedRevision
 			return errors.New("contract set is missing contract that was just added")
 		}
 		defer cs.Return(sc)
-		if len(cr.MerkleRoots) == sc.numMerkleRoots+1 {
+		if len(cr.MerkleRoots) == sc.merkleRoots.len()+1 {
 			root := cr.MerkleRoots[len(cr.MerkleRoots)-1]
 			_, err = sc.recordUploadIntent(cr.Revision, root, types.ZeroCurrency, types.ZeroCurrency)
 		} else {
--- a/modules/renter/proto/contract.go
+++ b/modules/renter/proto/contract.go
@@ -401,7 +401,7 @@ func (cs *ContractSet) managedInsertContract(h contractHeader, roots []crypto.Ha
 	}
 	// create fileSections
 	headerSection := newFileSection(f, 0, contractHeaderSize)
-	rootsSection := newFileSection(f, contractHeaderSize+1, -1)
+	rootsSection := newFileSection(f, contractHeaderSize, -1)
 	// write header
 	if _, err := headerSection.WriteAt(encoding.Marshal(h), 0); err != nil {
 		return modules.RenterContract{}, err
@@ -436,7 +436,7 @@ func (cs *ContractSet) loadSafeContract(filename string, walTxns []*writeaheadlo
 		return err
 	}
 	headerSection := newFileSection(f, 0, contractHeaderSize)
-	rootsSection := newFileSection(f, contractHeaderSize+1, remainingFile)
+	rootsSection := newFileSection(f, contractHeaderSize, remainingFile)
 
 	// read header
 	var header contractHeader
--- a/modules/renter/proto/contract.go
+++ b/modules/renter/proto/contract.go
@@ -2,16 +2,17 @@ package proto
 
 import (
 	"encoding/json"
-	"errors"
 	"io"
 	"os"
 	"path/filepath"
 	"sync"
 
+	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/encoding"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/types"
+	"github.com/NebulousLabs/errors"
 	"github.com/NebulousLabs/writeaheadlog"
 )
 
@@ -129,7 +130,8 @@ type SafeContract struct {
 
 	// merkleRoots are the Merkle roots of each sector stored on the host that
 	// relate to this contract.
-	merkleRoots []crypto.Hash
+	//merkleRoots []crypto.Hash
+	numMerkleRoots int
 
 	// unappliedTxns are the transactions that were written to the WAL but not
 	// applied to the contract file.
@@ -206,6 +208,28 @@ func (c *SafeContract) Utility() modules.ContractUtility {
 	return c.header.Utility
 }
 
+// merkleRoots returns the contracts merkle roots.
+func (c *SafeContract) merkleRoots() ([]crypto.Hash, error) {
+	merkleRoots := make([]crypto.Hash, 0, c.numMerkleRoots)
+	if _, err := c.f.Seek(contractHeaderSize, io.SeekStart); err != nil {
+		return merkleRoots, err
+	}
+	for {
+		var root crypto.Hash
+		if _, err := io.ReadFull(c.f, root[:]); err == io.EOF {
+			break
+		} else if err != nil {
+			return merkleRoots, errors.AddContext(err, "failed to read root from disk")
+		}
+		merkleRoots = append(merkleRoots, root)
+	}
+	// Sanity check: should have read exactly numMerkleRoots roots.
+	if len(merkleRoots) != c.numMerkleRoots {
+		build.Critical("Number of merkle roots on disk doesn't match numMerkleRoots")
+	}
+	return merkleRoots, nil
+}
+
 func (c *SafeContract) makeUpdateSetHeader(h contractHeader) writeaheadlog.Update {
 	c.headerMu.Lock()
 	id := c.header.ID()
@@ -250,10 +274,9 @@ func (c *SafeContract) applySetRoot(root crypto.Hash, index int) error {
 	if _, err := c.f.WriteAt(root[:], rootOffset); err != nil {
 		return err
 	}
-	if len(c.merkleRoots) <= index {
-		c.merkleRoots = append(c.merkleRoots, make([]crypto.Hash, 1+index-len(c.merkleRoots))...)
+	if c.numMerkleRoots <= index {
+		c.numMerkleRoots++
 	}
-	c.merkleRoots[index] = root
 	return nil
 }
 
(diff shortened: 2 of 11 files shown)


ChrisSchinnerl:
This PR is dependent on [PR#19](https://github.com/NebulousLabs/merkletree/pull/19)

modules/renter/proto/contract.go
```diff
@@ -345,8 +337,9 @@ func (cs *ContractSet) managedInsertContract(h contractHeader, roots []crypto.Ha
 		return modules.RenterContract{}, err
 	}
 	// write roots
-	for i, root := range roots {
-		if _, err := f.WriteAt(root[:], contractHeaderSize+crypto.HashSize*int64(i)); err != nil {
+	merkleRoots := newMerkleRoots(f)
+	for _, root := range roots {
+		if err := merkleRoots.push(root); err != nil {
```
DavidVorick: need to add a TODO somewhere in this package that currently the startup routine involves reading all of the merkle roots into memory, then hashing all of the merkle roots in the the `merkleRoots` object. That's a slow process that we should fix at some point, but it's probably not necessary until after we fix up the file structures.

modules/renter/proto/merkleroots.go
```diff
@@ -0,0 +1,343 @@
...
+		// cached subTree in cachedSubTrees.
+		uncachedRoots []crypto.Hash
+
+		// file is the file of the safe contract that contains the root.
+		file *os.File
+		// numMerkleRoots is the number of merkle roots in file.
+		numMerkleRoots int
+	}
+
+	cachedSubTree struct {
```
DavidVorick: need docstring

modules/renter/proto/merkleroots.go
```diff
@@ -0,0 +1,343 @@
...
+// loadExistingMerkleRoots reads creates a merkleRoots object from existing
+// merkle roots.
+func loadExistingMerkleRoots(file *os.File) (mr *merkleRoots, err error) {
+	mr = &merkleRoots{
+		file: file,
+	}
+	// Get the number of roots stored in the file.
+	mr.numMerkleRoots, err = mr.lenFromFile()
+	if err != nil {
+		return
```
DavidVorick: convention would be to return `nil, err` anytime that there's an error, same follows for the rest of the function. For a function this long, it's we tend to strongly prefer unnamed return values, so that we have to explicitly state what is being returned.

modules/renter/proto/merkleroots.go
```diff
@@ -0,0 +1,343 @@
...
+	}
+	// Seek to the first root's offset.
+	if _, err = file.Seek(rootIndexToOffset(0), io.SeekStart); err != nil {
+		return
+	}
+	// Read the roots from the file without reading all of them at once.
+	for i := 0; i < mr.numMerkleRoots; i++ {
```
DavidVorick: I would add a TODO at the top of the file, reading the roots one at a time is pretty inefficient. The OS will handle a lot of that via disk caching, but I'm guessing we would see some i/o and performance improvements if we read say, 100,000 roots at a time instead of one at a time.

No need to do that in this PR, I'd just write a comment highlighting that it's something to do.

Actually, this is the type of optimization that would be really great to throw at a newer developer like Matt or one of the guys in the #core-dev channel.

DavidVorick: @tbenz9 maybe we should make an issue when this is merged and get someone in the open source community to look at it.

modules/renter/proto/merkleroots.go
```diff
@@ -0,0 +1,343 @@
...
+		var root crypto.Hash
+		if _, err = io.ReadFull(file, root[:]); err == io.EOF {
+			break
+		} else if err != nil {
+			return
+		}
+
+		// Append the root to the unachedRoots
```
DavidVorick: unachedRoots. I guess they used to ache, but now they have been unached :P

modules/renter/proto/merkleroots.go
```diff
@@ -0,0 +1,343 @@
...
+		mr.uncachedRoots = append(mr.uncachedRoots, root)
+
+		// If the uncachedRoots grew too large we add them to the cache.
+		if len(mr.uncachedRoots) == (1 << merkleRootsCacheHeight) {
```
DavidVorick: This code should be abstracted, there should be a function for adding a new root to the data structure. I know we have the insert() function, but that one writes to disk as well as adding to the in-memory structure. Here we only care about adding to the in-memory structure, but that should still be its own abstraction that gets called anywhere the in-memory structure is extended (including insert())

modules/renter/proto/merkleroots.go
```diff
@@ -0,0 +1,343 @@
...
+		}
+	}
+	return mr, nil
+}
+
+// newCachedSubTree creates a cachedSubTree from exactly
+// 2^merkleRootsCacheHeight roots.
+func newCachedSubTree(roots []crypto.Hash) *cachedSubTree {
+	// Sanity check the input length.
+	if len(roots) != (1 << merkleRootsCacheHeight) {
```
DavidVorick: So, height of 0 means it's the hash of a leaf, height of 1 means it's the root of 2 leaves, height of 2 means root of 4 leaves, etc?

I don't think that's clearly documented in the merkle tree cached functions, we should make sure the docstrings for all of the cached roots properly explain what counts as height 0 and 1 etc.

modules/renter/proto/merkleroots.go
```diff
@@ -0,0 +1,343 @@
...
+// Don't use this on a file that contains roots.
+func newMerkleRoots(file *os.File) *merkleRoots {
+	return &merkleRoots{
+		file: file,
+	}
+}
+
+// rootIndexToOffset calculates the offset of the merkle root at index i from
+// the beginning of the contract file.
+func rootIndexToOffset(i int) int64 {
```
DavidVorick: this function name took me a few reads to understand what it was saying. I understand it now, but maybe there's a different name which isn't as confusing on the first read.

modules/renter/proto/merkleroots.go
```diff
@@ -0,0 +1,343 @@
...
+		// persisted and are computed after startup.
+		cachedSubTrees []*cachedSubTree
+		// uncachedRoots contains the sector roots that are not part of a
+		// cached subTree. The uncachedRoots slice should never get longer than
+		// 2^merkleRootsCacheHeight since that would simply result in a new
+		// cached subTree in cachedSubTrees.
+		uncachedRoots []crypto.Hash
+
+		// file is the file of the safe contract that contains the root.
+		file *os.File
```
lukechampine: I don't like that both `merkleRoots` and `SafeContract` have a handle to the same file. If you're just looking at the `SafeContract` code, it's not obvious that calling a method on `sc.merkleRoots` modifies sc.f`. I'm not sure what the best way to resolve this is. Maybe it's a mistake to have a separate `merkleRoots` type, and it should just be merged with `SafeContract`.

modules/renter/proto/merkleroots.go
lukechampine: optimizing this is really simple thanks to `bufio`:
```go
r := bufio.NewReader(file)
```
You can get a little faster by using a custom (larger) buffer size, but the default buffer size will probably get you to 90% peak performance.

modules/renter/proto/merkleroots.go
```diff
@@ -0,0 +1,343 @@
...
+	return i / (1 << merkleRootsCacheHeight), true
+}
+
+// lenFromFile returns the number of merkle roots by computing it from the
+// filesize.
+func (mr *merkleRoots) lenFromFile() (int, error) {
+	offset, err := mr.file.Seek(0, io.SeekEnd)
+	if err != nil {
+		return 0, err
+	}
```
lukechampine: more idiomatic to get the size via `Stat`:
```go
stat, err := mr.file.Stat()
if err != nil {
    return 0, err
}
size := stat.Size()
```

modules/renter/proto/merkleroots.go
```diff
@@ -0,0 +1,343 @@
...
+	// If we haven't written a single root yet we just return 0.
+	rootStart := rootIndexToOffset(0)
+	if offset < rootStart {
+		return 0, nil
+	}
+
+	// Sanity check contract file length.
+	if (offset-rootStart)%crypto.HashSize != 0 {
+		build.Critical("contract file has unexpected length and might be corrupted.")
```
lukechampine: this is not deserving of a `build.Critical`. The contract could be truncated to the nearest hash, or it could simply return an error and treat the contract as unusable -- but it's not an unrecoverable error.

modules/renter/proto/merkleroots.go
```diff
@@ -0,0 +1,343 @@
...
+
+// lenFromFile returns the number of merkle roots by computing it from the
+// filesize.
+func (mr *merkleRoots) lenFromFile() (int, error) {
+	offset, err := mr.file.Seek(0, io.SeekEnd)
+	if err != nil {
+		return 0, err
+	}
+	// If we haven't written a single root yet we just return 0.
+	rootStart := rootIndexToOffset(0)
```
lukechampine: isn't this just `contractHeaderSize`? Not sure if one is more clear than the other.

lukechampine: also, is this branch actually necessary? Won't the remaining logic return 0 just fine?

modules/renter/proto/merkleroots.go
```diff
@@ -0,0 +1,343 @@
...
+// number as lenFromFile.
+func (mr *merkleRoots) len() int {
+	return mr.numMerkleRoots
+}
+
+// push appends a merkle root to the end of the contract. If the number of
+// uncached merkle roots grows too big we cache them in a new subTree.
+func (mr *merkleRoots) push(root crypto.Hash) error {
+	// Sanity check the number of uncached roots before adding a new one.
+	if len(mr.uncachedRoots) == (1 << merkleRootsCacheHeight) {
```
lukechampine: `(1 << merkleRootsCacheHeight)` is a common enough expression that we may want to define a `merkleRootsCacheSize` or `merkleRootsPerCache` const.

modules/renter/proto/merkleroots.go
```diff
@@ -0,0 +1,343 @@
...
+
+// merkleRootsFrom index readds all the merkle roots in range [from;to)
+func (mr *merkleRoots) merkleRootsFromIndex(from, to int) ([]crypto.Hash, error) {
+	merkleRoots := make([]crypto.Hash, 0, to-from)
+	if _, err := mr.file.Seek(rootIndexToOffset(from), io.SeekStart); err != nil {
+		return merkleRoots, err
+	}
+	for i := from; to-i > 0; i++ {
+		var root crypto.Hash
+		if _, err := io.ReadFull(mr.file, root[:]); err == io.EOF {
```
lukechampine: ditto here on `bufio` optimization

modules/renter/proto/merkleroots_test.go
```diff
@@ -0,0 +1,324 @@
...
+	roots, err = merkleRoots.merkleRoots()
+	if err != nil {
+		t.Fatal("failed to get roots from disk", err)
+	}
+	if roots[0] != newHash {
+		t.Fatal("root wasn't updated correctly on disk")
+	}
+	if merkleRoots.cachedSubTrees[0].sum != newCachedSubTree(roots[:1<<merkleRootsCacheHeight]).sum {
+		t.Fatal("cachedSubtree doesn't have expected sum")
+	}
```
DavidVorick: should extend this test to include restarting / reloading to make sure that it's still consistent on disk after use.

modules/renter/proto/merkleroots_test.go
```diff
@@ -0,0 +1,324 @@
...
+	// There should be 0 cached roots.
+	if len(merkleRoots.cachedSubTrees) != 0 {
+		t.Fatal("expected 0 cached roots but was", len(merkleRoots.cachedSubTrees))
+	}
+}
+
+// TestDeleteLastRoot tests the deleteLastRoot method by creating many roots
+// and deleting random indices until there are no more roots left.
+func TestDelete(t *testing.T) {
```
DavidVorick: Docstring mismatch

modules/renter/proto/merkleroots_test.go
```diff
@@ -0,0 +1,324 @@
...
+			subTreeLen := int(1 << merkleRootsCacheHeight)
+			from := cachedIndex * (1 << merkleRootsCacheHeight)
+			roots, err := merkleRoots.merkleRootsFromIndex(from, from+subTreeLen)
+			if err != nil {
+				t.Fatal("failed to read roots of subTree", err)
+			}
+			if merkleRoots.cachedSubTrees[cachedIndex].sum != newCachedSubTree(roots).sum {
+				t.Fatal("new cached root sum doesn't match expected sum")
+			}
+		}
```
DavidVorick: performance might take too much of a hit if you do it every iteration, but maybe 1 in 100 iterations or something (randomly) you re-load the file and check for consistency.

modules/renter/proto/merkleroots_test.go
```diff
@@ -0,0 +1,324 @@
...
+
+		// Insert
+		var hash crypto.Hash
+		copy(hash[:], fastrand.Bytes(len(hash)))
+		index := fastrand.Intn(merkleRoots.numMerkleRoots + 1)
+		if err := merkleRoots.insert(index, hash); err != nil {
+			t.Fatalf("failed to insert %v at %v: %v", hash, index, err)
+		}
+	}
```
DavidVorick: occasional re-inits here as well I think

DavidVorick:
Went through all the testing, mostly looks good. Want to make sure everything is being persisted correctly.

Are there any tests that would make sense to add to the `siatest` package? I want to make sure we have coverage for this, though I'm thinking right now that our existing `siatest` tests are probably good enough. The only other thing I can think that we'd need to check is the ACID guarantees, but we are working on that in a different set of pull requests.

modules/renter/proto/merkleroots_test.go
```diff
@@ -11,6 +12,51 @@ import (
...
+	for i := 0; i < len(roots1); i++ {
+		if roots1[i] != roots2[i] {
+			return errors.New("on-disk roots don't match")
+		}
+	}
+	for i := 0; i < len(m1.uncachedRoots); i++ {
+		if m1.uncachedRoots[i] != m2.uncachedRoots[i] {
+			return errors.New("uncached roots don't match")
+		}
+	}
```
lukechampine: `reflect.DeepEqual` makes these a bit cleaner

modules/renter/proto/merkleroots.go
```diff
@@ -0,0 +1,359 @@
+// TODO currently the cached trees are not persisted and we build them at
+// startup. For petabytes of data this might take a long time.
```
DavidVorick: File should take the form:

```go
// package asdf is an example package that doesn't exist.
pacakge asdf

// TODO: write the asdf package somewhere
```

modules/renter/proto/merkleroots.go
DavidVorick: going to agree with this, shouldn't have 2 objects with atomic assumptions pointing to the same file on disk, they might break eachother.

We should have this all happen under the SafeContract, using the wal of the SafeContract.

modules/renter/proto/merkleroots.go
```diff
@@ -0,0 +1,359 @@
...
+		file: file,
+	}
+	// Get the number of roots stored in the file.
+	var err error
+	mr.numMerkleRoots, err = mr.lenFromFile()
+	if err != nil {
+		return nil, err
+	}
+	// Seek to the first root's offset.
+	if _, err = file.Seek(fileOffsetFromRootIndex(0), io.SeekStart); err != nil {
```
DavidVorick: I'm thinking we want this function to be a method on the contract, instead of a global constant. I suspect that if we ever change the offset in the future, having it a method of the safecontract will clean things up when we transition from one thing to the next.

modules/renter/proto/merkleroots.go
```diff
@@ -0,0 +1,359 @@
...
+		return nil, err
+	}
+	// Read the roots from the file without reading all of them at once.
+	r := bufio.NewReader(file)
+	for i := 0; i < mr.numMerkleRoots; i++ {
```
DavidVorick: I'm thinking we should read more than 1 root at a time. I'm not sure what type of cacheing the OS does, but if we read 1000 or 10,000 roots at a time we'll probably see big speedups on systems without good OS level caching.

modules/renter/proto/merkleroots.go
DavidVorick: maybe it's not such a big deal

modules/renter/proto/merkleroots.go
```diff
@@ -0,0 +1,359 @@
...
+	}
+	// If it is not uncached we need to delete the last cached tree and load
+	// its elements into mr.uncachedRoots.
+	mr.cachedSubTrees = mr.cachedSubTrees[:len(mr.cachedSubTrees)-1]
+	rootIndex := len(mr.cachedSubTrees) * merkleRootsPerCache
+	roots, err := mr.merkleRootsFromIndex(rootIndex, mr.numMerkleRoots)
+	if err != nil {
+		return errors.AddContext(err, "failed to read cached tree's roots")
+	}
+	mr.uncachedRoots = append(mr.uncachedRoots, roots...)
```
DavidVorick: this logic of deleting the last cached and populating the uncached appears in a few places, probably worth combining them into a helper method.

modules/renter/proto/merkleroots.go
ChrisSchinnerl: Yeah that's what the `bufio.NewReader(file)` line should do. I'm not sure how much it actually buffers but it should do some good userspace level caching.  @lukechampine probably knows more about this.

modules/renter/proto/merkleroots.go
```diff
@@ -0,0 +1,368 @@
...
+// len returns the number of merkle roots. It should always return the same
+// number as lenFromFile.
+func (mr *merkleRoots) len() int {
+	return mr.numMerkleRoots
+}
+
+// appendRoot appends a root to the in-memory structure of the merkleRoots. If
+// the length of the uncachedRoots grows too large they will be compressed into
+// a cachedSubTree.
+func (mr *merkleRoots) appendRoot(root crypto.Hash) {
```
DavidVorick: may want to call this `appendRootMemory` or something, so that the function name clearly indicates that it only modifies the in-memory structure.

modules/renter/proto/merkleroots.go
```diff
@@ -0,0 +1,368 @@
...
+	mr.appendRoot(root)
+
+	// Increment the number of roots.
+	mr.numMerkleRoots++
+	return nil
+}
+
+// root returns the root of the merkle roots.
+func (mr *merkleRoots) root() crypto.Hash {
+	tree := crypto.NewCachedTree(sectorHeight)
```
DavidVorick: If we're working with pushing sub-roots, does it need to be a cached tree? Can it just be a normal tree?

I think that with the help of the `PushSubTree` function, we can stop using the `CachedTree` type.

modules/renter/proto/merkleroots.go
```diff
@@ -0,0 +1,368 @@
...
+	}
+	for _, root := range mr.uncachedRoots {
+		tree.Push(root)
+	}
+	return tree.Root()
+}
+
+// newRoot returns the root of the merkleTree after appending the newRoot
+// without actually appending it.
+func (mr *merkleRoots) newRoot(newRoot crypto.Hash) crypto.Hash {
```
DavidVorick: I would call this `checkNewRoot`

modules/renter/proto/contract.go
```diff
@@ -96,14 +95,20 @@ type SafeContract struct {
...
 	// applied to the contract file.
 	unappliedTxns []*writeaheadlog.Transaction
 
+	// file is the file of the safe contract that contains the roots. This
+	// file is usually shared with the SafeContract which means multiple
+	// threads could potentially write to the same file. That's why the
+	// SafeContract should never modify the file beyond the
+	// contractHeaderSize and the merkleRoots should never modify data
+	// before that. Both should also use WriteAt and ReadAt instead of
+	// Write and Read.
```
lukechampine: can make this safer like so:
```go
type fileSection struct {
	f     *os.File
	start int64
	end   int64
}

func (f *fileSection) WriteAt(p []byte, off int64) (int, error) {
	return f.f.WriteAt(b, f.start+off)
}

// also implement ReadAt, Sync, and Close
```
The tricky part is handling out-of-bounds writes. For the header part of the file, you want to disallow writes past the end. But for the merkle roots part, you need to allow them. Two approaches come to mind:

1) Set `end: -1` for the merkle roots `fileSection`, and check for `f.end == -1` in `WriteAt`.
2) Explicitly extend `f.end` before calling `WriteAt`.

Option 2 seems a bit safer, but I won't make the final call.

The calls to `WriteAt` and `ReadAt` in the merkle roots section also need to updated to use the correct offsets.

modules/renter/proto/merkleroots.go
```diff
@@ -0,0 +1,375 @@
...
+		tree.Push(root)
+	}
+	// Push the new root.
+	tree.Push(newRoot)
+	return tree.Root()
+}
+
+// merkleRoots reads all the merkle roots from disk and returns them. This is
+// not very fast and should only be used for testing purposes.
+func (mr *merkleRoots) merkleRoots() (roots []crypto.Hash, err error) {
```
DavidVorick: If this function is only intended for testing, I would move it to an _test.go file, and then probably rename it to `merkleRootsFromDisk()`

modules/renter/proto/merkleroots.go
```diff
@@ -0,0 +1,375 @@
...
+	}
+	// Sanity check: should have read exactly numMerkleRoots roots.
+	if len(roots) != mr.numMerkleRoots {
+		build.Critical(fmt.Sprintf("Number of merkle roots on disk (%v) doesn't match numMerkleRoots (%v)",
+			len(roots), mr.numMerkleRoots))
+	}
+	return
+}
+
+// merkleRootsFrom index readds all the merkle roots in range [from;to)
```
DavidVorick: typo - readds

modules/renter/proto/merkleroots.go
```diff
@@ -0,0 +1,375 @@
...
+func (mr *merkleRoots) merkleRootsFromIndex(from, to int) ([]crypto.Hash, error) {
```
DavidVorick: would probably also have it be `merkleRootsFromIndexFromDisk`. Though that's a bit of a mouthful.

modules/renter/proto/contract.go
```diff
@@ -411,17 +399,17 @@ func (cs *ContractSet) managedInsertContract(h contractHeader, roots []crypto.Ha
 	if err != nil {
 		return modules.RenterContract{}, err
 	}
-	// preallocate space for header + roots
-	if err := f.Truncate(contractHeaderSize + crypto.HashSize*int64(len(roots))); err != nil {
-		return modules.RenterContract{}, err
-	}
+	// create fileSections
+	headerSection := newFileSection(f, 0, contractHeaderSize)
+	rootsSection := newFileSection(f, contractHeaderSize+1, -1)
```
DavidVorick: my preference would be to change the `-1` sentinel to a constant. It's not super obvious, had to read the implementation to figure out what it did. But if it was a constant `remainingFile`, it might be more clear.

modules/renter/proto/filesection.go
```diff
@@ -0,0 +1,98 @@
...
+	return f.f.ReadAt(b, f.start+off)
+}
+
+// Sync calls Sync on the fileSection's underlying file.
+func (f *fileSection) Sync() error {
+	return f.f.Sync()
+}
+
+// Truncate calls Truncate on the fileSection's underlying file.
+func (f *fileSection) Truncate(size int64) error {
```
DavidVorick: should just return an error period if you call `Truncate` on a `fileSection` and `end != -1`

modules/renter/proto/merkleroots.go
```diff
@@ -0,0 +1,369 @@
...
+func (mr *merkleRoots) appendRootMemory(root crypto.Hash) {
+	mr.uncachedRoots = append(mr.uncachedRoots, root)
+	if len(mr.uncachedRoots) == merkleRootsPerCache {
+		mr.cachedSubTrees = append(mr.cachedSubTrees, newCachedSubTree(mr.uncachedRoots))
+		mr.uncachedRoots = mr.uncachedRoots[:0]
+	}
+}
+
+// delete deletes the sector root at a certain index.
+func (mr *merkleRoots) delete(i int) error {
```
DavidVorick: This operation isn't WAL-safe, it's not idempotent. If the WAL is going to be calling `delete` from a transaction, it needs to be able to call delete multiple times without causing harm.

the issue is that if we call `delete(5)` twice in a row, we delete two different roots, instead of deleting the same root. If this is to be called from the WAL, the instructions actually have to take 2 inputs:

1. some data to write at a specific index (for the swap) (might need a bool too indicating whether a swap is required)

2. some data indicating which index should be truncated to.

Those values combined will let you perform idempotent deletes.

modules/renter/proto/merkleroots.go
```diff
@@ -0,0 +1,369 @@
...
+	if err != nil {
+		return errors.AddContext(err, "failed to read cached tree's roots")
+	}
+	mr.uncachedRoots = append(mr.uncachedRoots, roots...)
+	return nil
+}
+
+// push appends a merkle root to the end of the contract. If the number of
+// uncached merkle roots grows too big we cache them in a new subTree.
+func (mr *merkleRoots) push(root crypto.Hash) error {
```
DavidVorick: `push` has the same problem as delete - it's not idempotent. WAL may end up calling an update multiple times if there are crashes at key places during disk management, so we need to restructure all of these functions to be idempotent.

This means we also need to write idempotency tests to verify that all of the operations are being updated idempotently. In the host we do this by adding a disrupt during WAL application that might write an update multiple times.

True idempotency though would be to apply some of the transactions in a WAL, perhaps even get partway through one of the transactions, and then apply them again a few times, each time stopping at a different update (or even finishing all the way), until finally on the last one you just go all the way.

If the WAL is built correctly, the result should be identical to just applying the transaction once.

modules/renter/proto/merkleroots.go
```diff
@@ -66,27 +81,26 @@ func loadExistingMerkleRoots(file *fileSection) (*merkleRoots, error) {
...
-	}
 	// Read the roots from the file without reading all of them at once.
-	r := bytes.NewBuffer(fileData)
-	for i := 0; i < mr.numMerkleRoots; i++ {
-		var root crypto.Hash
-		if _, err = io.ReadFull(r, root[:]); err == io.EOF {
+	readOff := int64(0)
+	rootsData := make([]byte, rootsDiskLoadBulkSize)
+	for {
+		n, err := file.ReadAt(rootsData, readOff)
```
lukechampine: are you not able to use `bufio` anymore because `file` only supports `ReadAt`, not `Read`?
That makes me sad.

modules/renter/proto/consts.go
```diff
@@ -11,6 +11,14 @@ import (
 const (
 	// contractExtension is the extension given to contract files.
 	contractExtension = ".contract"
+
+	// rootsDiskLoadBulkSize is the max number of roots we read from disk at
+	// once to avoid using up all the ram.
+	rootsDiskLoadBulkSize = 1024 * crypto.HashSize // 32 kib
+
+	// remainingFile is a constant used to indicate that a fileSection can access
+	// the whole remaining file instead of being boung to a certain end offset.
```
DavidVorick: typo - boung

modules/renter/proto/contract.go
```diff
@@ -447,27 +435,22 @@ func (cs *ContractSet) loadSafeContract(filename string, walTxns []*writeaheadlo
 	if err != nil {
 		return err
 	}
+	headerSection := newFileSection(f, 0, contractHeaderSize)
+	rootsSection := newFileSection(f, contractHeaderSize+1, -1)
```
DavidVorick: need to use the constant here

modules/renter/proto/filesection.go
```diff
@@ -0,0 +1,94 @@
...
+	if f.end != remainingFile {
+		panic("can't truncate a file that has a end != remainingFile")
+	}
+	if f.start+size < f.start {
+		panic("can't truncate file to be smaller than the section start")
+	}
+	if f.start+size > f.end && f.end != remainingFile {
```
DavidVorick: this block is unreachable, but also unneeded

modules/renter/proto/filesection.go
```diff
@@ -0,0 +1,94 @@
+package proto
```
DavidVorick: This file needs tests to prove correctness. Especially around all of the edge cases that it's supposed to protect against.

PR #2880, opened 2018-03-16
add dependencies field to renter

ChrisSchinnerl:
This PR adds a dependency to the renter and also removes an unused method.


modules/renter/renter.go
```diff
@@ -465,5 +467,5 @@ func New(g modules.Gateway, cs modules.ConsensusSet, wallet modules.Wallet, tpoo
 		return nil, err
 	}
 
-	return newRenter(g, cs, tpool, hdb, hc, persistDir)
+	return newRenter(g, cs, tpool, hdb, hc, persistDir, &modules.ProductionDependencies{})
```
DavidVorick: modules.ProdDependencies

PR #2882, opened 2018-03-16
Add dependencies to the contractSet

ChrisSchinnerl:
This PR adds a `deps` field to the contractSet.

### DIFF
--- a/modules/renter/contractor/contractor.go
+++ b/modules/renter/contractor/contractor.go
@@ -193,7 +193,7 @@ func New(cs consensusSet, wallet walletShim, tpool transactionPool, hdb hostDB, 
 	}
 
 	// Create the contract set.
-	contractSet, err := proto.NewContractSet(filepath.Join(persistDir, "contracts"), &modules.ProductionDependencies{})
+	contractSet, err := proto.NewContractSet(filepath.Join(persistDir, "contracts"), modules.ProdDependencies)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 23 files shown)


modules/renter/contractor/contractor.go
```diff
@@ -44,6 +44,7 @@ type Contractor struct {
 	tg      siasync.ThreadGroup
 	tpool   transactionPool
 	wallet  wallet
+	deps    modules.Dependencies
```
DavidVorick: alphabetize

modules/renter/contractor/contractor.go
```diff
@@ -210,6 +211,7 @@ func newContractor(cs consensusSet, w wallet, tp transactionPool, hdb hostDB, co
 		persist: p,
 		tpool:   tp,
 		wallet:  w,
+		deps:    deps,
```
DavidVorick: alphabetize

modules/renter/proto/contractset.go
```diff
@@ -19,6 +19,7 @@ import (
 type ContractSet struct {
 	contracts map[types.FileContractID]*SafeContract
 	wal       *writeaheadlog.WAL
+	deps      modules.Dependencies
```
DavidVorick: alphabetize

modules/renter/proto/contractset.go
```diff
@@ -134,7 +135,7 @@ func (cs *ContractSet) Close() error {
 
 // NewContractSet returns a ContractSet storing its contracts in the specified
 // dir.
-func NewContractSet(dir string) (*ContractSet, error) {
+func NewContractSet(dir string, deps modules.Dependencies) (*ContractSet, error) {
```
DavidVorick: alphabetize

DavidVorick:
in modules/dependencies.go, we can create a variable

```go
var (
    ProdDeps = new(ProductionDependencies)
)
```

Then we can use `modules.ProdDeps` as input instead of `&modules.ProductionDependencies{}` as input. I'm not 100% sure I got the syntax right in the declaration above, but I think it'll be something close to that.

PR #2884, opened 2018-03-16
WIP: Fix host counter

nielscastien:
@DavidVorick This is a first attempt to fix the host financial metrics that are off most of the time for all the hosts.

I noticed that whenever my host financial metrics increased a lot in a small period of time, there were at the same time consensus errors in the `host.log`. These errors always repeated themselves 6 times with 15s interval. See [consensus.log](https://gist.github.com/nielscastien/ca633c579e225dd3d5bcc5c287e9b2bb).

Therefore, I looked at `negotiate.go` and found out that there is a loop that calls
 `err = h.managedAddStorageObligation(so) `. 
Worst case this function is called 6 times if there are issues with the acceptance of the transaction set (leading to the logged errors). Every time this function is called on the same storage obligation, the host financial metrics are updated.

This PR tries to fix the double counting issue. But if I am correct, this will also lead to additional insertions into the host database. This might lead to several other bugs where a host can't provide a storage proof when a storage obligation ends.

I know the focus is on upgrading the renter. But this could potentially fix a lot of errors for the hosts. Any feedback is welcome...

cc @lukechampine

### DIFF
--- a/modules/host/storageobligations.go
+++ b/modules/host/storageobligations.go
@@ -369,16 +369,6 @@ func (h *Host) managedAddStorageObligation(so storageObligation) error {
 			return err
 		}
 
-		// Update the host financial metrics with regards to this storage
-		// obligation.
-		h.financialMetrics.ContractCount++
-		h.financialMetrics.PotentialContractCompensation = h.financialMetrics.PotentialContractCompensation.Add(so.ContractCost)
-		h.financialMetrics.LockedStorageCollateral = h.financialMetrics.LockedStorageCollateral.Add(so.LockedCollateral)
-		h.financialMetrics.PotentialStorageRevenue = h.financialMetrics.PotentialStorageRevenue.Add(so.PotentialStorageRevenue)
-		h.financialMetrics.PotentialDownloadBandwidthRevenue = h.financialMetrics.PotentialDownloadBandwidthRevenue.Add(so.PotentialDownloadRevenue)
-		h.financialMetrics.PotentialUploadBandwidthRevenue = h.financialMetrics.PotentialUploadBandwidthRevenue.Add(so.PotentialUploadRevenue)
-		h.financialMetrics.RiskedStorageCollateral = h.financialMetrics.RiskedStorageCollateral.Add(so.RiskedCollateral)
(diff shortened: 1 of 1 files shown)


modules/host/storageobligations.go
```diff
@@ -414,6 +404,22 @@ func (h *Host) managedAddStorageObligation(so storageObligation) error {
 		h.log.Println("Error with transaction set, redacting obligation, id", so.id())
 		return composeErrors(err, h.removeStorageObligation(so, obligationRejected))
 	}
+
+	// Increase contract count by one only if this is a new filecontract.
+	if len(so.SectorRoots) == 0 {
+		h.financialMetrics.ContractCount++
+	}
```
lukechampine: I'm not sure about this part. I guess it depends on the precise meaning of `ContractCount`. Even if you renew a contract, the host still needs to track the old contract until it ends, so it might make sense to increment `ContractCount` even for renewed contracts.

lukechampine:
This seems basically correct to me. Although, it makes me suspect that the host state isn't being properly rolled back in the event of an error. For example, there's a `db.Update` call earlier in the function; if `managedAddStorageObligation` ultimately fails, then we'd want to revert those changes, right? Perhaps that can be tackled in a subsequent PR.

modules/host/storageobligations.go
nielscastien: You are right! I reverted the commits. Was confused about the 'renewal' word, where it actually is a new storage obligation we need to track.

nielscastien:
@lukechampine I reverted the old and added some new commits. Please take a look if I am not missing something.

The only thing that still bothers me is the way a storage obligation with `obligatonRejected` status is removed within `removeStorageObligation`. In this PR I make use of it twice (to prevent duplication of code):
1. after `AcceptTransactionSet` return an error (this PR),
2. or if there is an error with a `queueActionItem` (original code)

Before the financial metrics are updated it checks:
`h.financialMetrics.TransactionFeeExpenses.Cmp(so.TransactionFeesAdded) >= 0`
but I don't know when this is true (or not). Maybe you have more insight if it works in this special case.

PR #2861, opened 2018-03-13
change accepttransactionset logging to debug

ChrisSchinnerl:
The logging output of `AcceptTransactionSet` fills up the log pretty quickly. This PR changes it to only log in a `DEBUG` build.

PR #2888, opened 2018-03-18
remove overly verbose logging in managedAcceptBlocks

lukechampine:
these logging statements cause `consensus.log` to grow to many megabytes.

### DIFF
--- a/modules/consensus/accept.go
+++ b/modules/consensus/accept.go
@@ -241,7 +241,6 @@ func (cs *ConsensusSet) managedAcceptBlocks(blocks []types.Block) (blockchainExt
 	chainExtended := false
 	changes := make([]changeEntry, 0, len(blocks))
 	setErr := cs.db.Update(func(tx *bolt.Tx) error {
-		cs.log.Printf("accept: starting block processing loop (%v blocks, height %v)", len(blocks), blockHeight(tx))
 		for i := 0; i < len(blocks); i++ {
 			// Start by checking the header of the block.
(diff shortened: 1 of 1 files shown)


ChrisSchinnerl:
@lukechampine I feel like I created a PR in the past that changes the logging level of those to `DEBUG` but I can't seem to find it. Removing them should also be fine though. They were only used to find a very specific bug.

nielscastien:
@ChrisSchinnerl That was PR #2861 for the `transactionpool.log`

tbenz9:
Is there any interest in having Sia support a logging verbosity flag?  I know `DEBUG` is much more verbose, but users may want to be able to adjust their logging verbosity locally to understand what it going on.  I'd imagine a `siad -v` or `siad --verbosity=2` could be popular.

lukechampine:
Different verbosity levels are probably overkill for now. We already have enough bikeshedding about whether a given log statement should be DEBUG or not.

DavidVorick:
This output was added to help us debug where users were seeing the transaction pool crash due to blocks being out of order when sent through the consensus change.

Given that we are still seeing users report that bug, I think we should keep these logs in here for a little while. Going to leave this PR open, because these logging messages are not meant to be left in long-term, but I think they still might be useful in the next release, because the bug they are targeting doesn't seem to be solved fully.

DavidVorick:
Fixed by https://github.com/NebulousLabs/Sia/pull/3030

We can get rid of the verbose logging now.

PR #2900, opened 2018-03-22
fix typos

tbenz9:


modules/consensus/synchronize.go
```diff
@@ -62,7 +62,7 @@ var (
 		Testing:  3 * time.Second,
 	}).(time.Duration)
 
-	// sendBlkTimeout is the timeout for the SendBlocks RPC.
+	// sendBlkTimeout is the timeout for the SendBlock RPC.
```
lukechampine: I think `SendBlocks` is the correct name

tbenz9: There are two function:
https://github.com/NebulousLabs/Sia/blob/ef9e811b3b6c3ebc521bf4126e4475906b1b2724/modules/consensus/synchronize.go#L468
and
https://github.com/NebulousLabs/Sia/blob/ef9e811b3b6c3ebc521bf4126e4475906b1b2724/modules/consensus/synchronize.go#L267


Which is why there are two variable blocks:
https://github.com/NebulousLabs/Sia/blob/ef9e811b3b6c3ebc521bf4126e4475906b1b2724/modules/consensus/synchronize.go#L66
and
https://github.com/NebulousLabs/Sia/blob/ef9e811b3b6c3ebc521bf4126e4475906b1b2724/modules/consensus/synchronize.go#L73

PR #2906, opened 2018-03-27
Add dependency injection to siatest package

ChrisSchinnerl:
This PR adds dependency injection to the siatest package by adding fields to the `NodeParams` for each module's dependencies. If the field is not set, a production dependency will be used instead.

e.g. a renter node could be created like this:
```
template := Renter("path/to/my/renter")
template.RenterDeps = myCustomDeps
renter, err := node.New(template)
```

### DIFF
--- a/modules/renter/contractor/dependencies.go
+++ b/modules/renter/contractor/dependencies.go
@@ -64,12 +64,12 @@ type (
 
 // Because wallet is not directly compatible with modules.Wallet (wrong
 // type signature for StartTransaction), we must provide a bridge type.
-type walletBridge struct {
-	w walletShim
+type WalletBridge struct {
+	W walletShim
 }
 
-func (ws *walletBridge) NextAddress() (types.UnlockConditions, error) { return ws.w.NextAddress() }
-func (ws *walletBridge) StartTransaction() transactionBuilder         { return ws.w.StartTransaction() }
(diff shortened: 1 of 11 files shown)


DavidVorick:
Looks good to me, @lukechampine can you review as well?

modules/renter/contractor/dependencies.go
```diff
@@ -62,14 +62,19 @@ type (
...
 }
 
-func (ws *walletBridge) NextAddress() (types.UnlockConditions, error) { return ws.w.NextAddress() }
-func (ws *walletBridge) StartTransaction() transactionBuilder         { return ws.w.StartTransaction() }
+// NextAddress computes and returns the next address of the wallet.
+func (ws *WalletBridge) NextAddress() (types.UnlockConditions, error) { return ws.W.NextAddress() }
+
+// StartTransaction creates a new transactionBuilder that can be used to create
+// and sign a transaction.
+func (ws *WalletBridge) StartTransaction() transactionBuilder { return ws.W.StartTransaction() }
```
lukechampine: revisiting this later, I think it's preferable for `StartTransaction` to just return a `modules.TransactionBuilder`. It doesn't make the mocking that much harder, and in doing so we can avoid the ugly `WalletBridge` code.

lukechampine:
LGTM

Issue #1927, opened 2017-06-14
Feature request: Offline transaction signing


mtlynch:
With the increase in the Siacoin's market value, there's a greater need to to manage Siacoins in a way that protects the user from malware infections or active attackers.

I realize that there's an [offline wallet generator](https://github.com/johnathanhowell/sia-coldstorage), but I'm reluctant to use it if I can't verify that I can move coins sent to an address it generates. Right now, the only way to do that is to bring my offline wallet online, potentially compromising the wallet if the online machine is compromised.

Offline transaction signing would allow users to manage their wallets without ever having to connect their wallet-holding system to the Internet.

lukechampine:
This sounds like a good project for a third-party dev, or a weekend project for a Sia dev.
What sort of API do you have in mind? Would this be part of `siad`/`siac` or a standalone binary?

mtlynch:
>Would this be part of siad/siac or a standalone binary?

Ideal would be if it were possible natively with `siad` / `siac`, but I'd be happy with a standalone tool.

tbenz9:
Recommendation to devs: "Feature Request"

rmcdaniel:
@lukechampine I will volunteer to work on this. I am building a web wallet and need this feature.

lukechampine:
@rmcdaniel excellent. DM me on Discord (nemo -- need to Friend Request me first) if you need any pointers.

Can you (and/or @mtlynch) elaborate on what the API for this would look like?

mtlynch:
I haven't thought it through to the API level, but here's the functionality I had in mind:

## Initiating the offline wallet

**Offline machine**: User generates an offline wallet and writes a watch-only copy to a USB key.

```
# Create offline wallet
$ ./siac wallet init

# Create watch-only wallet that does not contain private keys
$ ./siac wallet make-watchonly /media/usb0/watchonly-wallet.wal
Enter the seed to use for the watchonly wallet: <> # User enters seed
Wrote watchonly wallet to: /media/usb0/watchonly-wallet.wal
```

**Online machine**: User loads the watchonly wallet on the online machine
```
# Load the watch-only wallet.
$ ./siac wallet load watchonly /media/usb0/watchonly-wallet.wal

# Online machine can see offline wallet's addresses without access to its private keys.
$ ./siac wallet addresses
<addresses>

# Online machine can see offline wallet's balance without access to its private keys.
$ ./siac wallet balance
<balance information>
```

## Sending a transaction

[...]

PR #2907, opened 2018-03-28
Add support for offline transaction signing

lukechampine:
This mostly addresses #1927. I will make a follow-up PR shortly that adds support for watch-only wallets. In the meantime, you can simulate a watch-only wallet by creating a normal wallet and then locking it.

The basic process is:

1. GET `/wallet/unspent` to receive a list of spendable outputs
2. Construct an unsigned transaction using the outputs
3. Use `siac wallet sign` to sign the transaction
4. POST `/tpool/raw` to broadcast the transaction

wrt 3, `siac wallet sign` does not require `siad`; it's completely self-contained. The downside of this is that you need to regenerate a bunch of keys from the seed, and you pay that cost every time you sign a transaction. For wallets that have <10000 addresses, this cost should be barely noticeable, but for larger wallets, there is also a `/wallet/sign` endpoint that will perform the same signing operation using the keys already in memory.

Lastly, there's the question of how to encode the transactions. `/tpool/raw` expects base64-encoded binary, but that isn't a very convenient format for *constructing* transactions. I figured that JSON was a better choice for that, so `siac wallet sign` and `/wallet/sign` both expect the transaction to be JSON.

### DIFF
--- a/modules/wallet/transactionbuilder.go
+++ b/modules/wallet/transactionbuilder.go
@@ -634,3 +634,87 @@ func (w *Wallet) RegisterTransaction(t types.Transaction, parents []types.Transa
 func (w *Wallet) StartTransaction() modules.TransactionBuilder {
 	return w.RegisterTransaction(types.Transaction{}, nil)
 }
+
+// SpendableOutputs returns the outputs spendable by the wallet. For each
+// output, MaturityHeight is the height of the block containing the output.
+func (w *Wallet) SpendableOutputs() []modules.ProcessedOutput {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+	// ensure durability of reported outputs
+	w.syncDB()
+
+	var outputs []modules.ProcessedOutput
+	dbForEachSiacoinOutput(w.dbTx, func(scoid types.SiacoinOutputID, sco types.SiacoinOutput) {
+		outputs = append(outputs, modules.ProcessedOutput{
+			FundType:       types.SpecifierSiacoinOutput,
+			ID:             types.OutputID(scoid),
+			RelatedAddress: sco.UnlockHash,
+			Value:          sco.Value,
+			WalletAddress:  true,
+		})
+	})
+	dbForEachSiafundOutput(w.dbTx, func(sfoid types.SiafundOutputID, sfo types.SiafundOutput) {
+		outputs = append(outputs, modules.ProcessedOutput{
+			FundType:       types.SpecifierSiafundOutput,
+			ID:             types.OutputID(sfoid),
+			RelatedAddress: sfo.UnlockHash,
+			Value:          sfo.Value,
+			WalletAddress:  true,
+		})
+	})
+
+	// lookup the confirmation height of each output
+	// TODO: would be much better to store this alongside outputs
+	for i, o := range outputs {
+		txnIndices, _ := dbGetAddrTransactions(w.dbTx, o.RelatedAddress)
+		for _, j := range txnIndices {
+			pt, err := dbGetProcessedTransaction(w.dbTx, j)
+			if err != nil {
+				continue
+			}
+			for _, sco := range pt.Outputs {
+				if sco.ID == o.ID {
+					outputs[i].MaturityHeight = pt.ConfirmationHeight
+					break
+				}
+			}
+		}
+	}
+
+	return outputs
+}
+
+// SignTransaction signs txn using secret keys controlled by w, which must be
+// unlocked. For each SiacoinInput whose UnlockConditions are not set,
+// SignTransaction attempts to fill in the UnlockConditions and adds a
+// corresponding signature. It returns the indices of each signed input.
+func (w *Wallet) SignTransaction(txn *types.Transaction) []int {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+	if !w.unlocked {
+		return nil
+	}
+
+	var signed []int
+	for i, sci := range txn.SiacoinInputs {
+		// identify inputs with unset UnlockConditions
+		if sci.UnlockConditions.SignaturesRequired == 0 {
+			// locate output corresponding to this input
+			sco, err := dbGetSiacoinOutput(w.dbTx, sci.ParentID)
+			if err != nil {
+				continue
+			}
+			// lookup the signing key(s)
+			sk, ok := w.keys[sco.UnlockHash]
+			if !ok {
+				w.log.Critical("wallet is missing a signing key")
+				continue
+			}
+			txn.SiacoinInputs[i].UnlockConditions = sk.UnlockConditions
+			cf := types.CoveredFields{WholeTransaction: true}
+			addSignatures(txn, cf, sk.UnlockConditions, crypto.Hash(sci.ParentID), sk)
+			signed = append(signed, i)
+		}
+	}
+	return signed
+}
--- a/modules/wallet/transactionbuilder.go
+++ b/modules/wallet/transactionbuilder.go
@@ -701,11 +701,23 @@ func (w *Wallet) SignTransaction(txn *types.Transaction, toSign map[types.Output
 // the ParentID of each unsigned input to the UnlockHash of that input's
 // desired UnlockConditions. SignTransaction fills in the UnlockConditions for
 // each such input and adds a corresponding signature.
+//
+// SignTransaction must derive all of the keys from scratch, so it is
+// appreciably slower than calling the Wallet.SignTransaction method. Only the
+// first 1 million keys are derived.
 func SignTransaction(txn *types.Transaction, seed modules.Seed, toSign map[types.OutputID]types.UnlockHash) error {
-	// generate 1M keys
+	// generate keys in batches up to 1e6 before giving up
 	keys := make(map[types.UnlockHash]spendableKey, 1e6)
-	for _, sk := range generateKeys(seed, 0, 1e6) {
-		keys[sk.UnlockConditions.UnlockHash()] = sk
+	var keyIndex uint64
+	const keysPerBatch = 1000
+	for len(keys) < 1e6 {
+		for _, sk := range generateKeys(seed, keyIndex, keyIndex+keysPerBatch) {
+			keys[sk.UnlockConditions.UnlockHash()] = sk
+		}
+		keyIndex += keysPerBatch
+		if err := signTransaction(txn, keys, toSign); err == nil {
+			return nil
+		}
 	}
 	return signTransaction(txn, keys, toSign)
 }
--- a/modules/wallet/transactionbuilder.go
+++ b/modules/wallet/transactionbuilder.go
@@ -684,37 +684,52 @@ func (w *Wallet) SpendableOutputs() []modules.ProcessedOutput {
 	return outputs
 }
 
-// SignTransaction signs txn using secret keys controlled by w, which must be
-// unlocked. For each SiacoinInput whose UnlockConditions are not set,
-// SignTransaction attempts to fill in the UnlockConditions and adds a
-// corresponding signature. It returns the indices of each signed input.
-func (w *Wallet) SignTransaction(txn *types.Transaction) []int {
+// SignTransaction signs txn using secret keys known to the wallet. toSign
+// maps the ParentID of each unsigned input to the UnlockHash of that input's
+// desired UnlockConditions. SignTransaction fills in the UnlockConditions for
+// each such input and adds a corresponding signature.
+func (w *Wallet) SignTransaction(txn *types.Transaction, toSign map[types.OutputID]types.UnlockHash) error {
 	w.mu.Lock()
 	defer w.mu.Unlock()
 	if !w.unlocked {
-		return nil
+		return modules.ErrLockedWallet
 	}
+	return signTransaction(txn, w.keys, toSign)
+}
 
-	var signed []int
-	for i, sci := range txn.SiacoinInputs {
-		// identify inputs with unset UnlockConditions
-		if sci.UnlockConditions.SignaturesRequired == 0 {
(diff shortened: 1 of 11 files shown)


modules/wallet/transactionbuilder.go
```diff
@@ -634,3 +634,127 @@ func (w *Wallet) RegisterTransaction(t types.Transaction, parents []types.Transa
 func (w *Wallet) StartTransaction() modules.TransactionBuilder {
 	return w.RegisterTransaction(types.Transaction{}, nil)
 }
+
+// SpendableOutputs returns the outputs spendable by the wallet.
+func (w *Wallet) SpendableOutputs() []modules.SpendableOutput {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+	// ensure durability of reported outputs
+	w.syncDB()
```
ChrisSchinnerl: Do we need to sync here? Once you got the outputs and created a transaction, the transaction pool will decide if the transaction is valid. Therefore I think it is not really an issue if the output disappears from the wallet due to a crash.

modules/wallet/transactionbuilder.go
```diff
@@ -634,3 +634,127 @@ func (w *Wallet) RegisterTransaction(t types.Transaction, parents []types.Transa
...
+			keys[sk.UnlockConditions.UnlockHash()] = sk
+		}
+		keyIndex += keysPerBatch
+		if err := signTransaction(txn, keys, toSign); err == nil {
+			return nil
+		}
+	}
+	return signTransaction(txn, keys, toSign)
+}
+
```
ChrisSchinnerl: I know that it is not required for unexported functions but I like to have comments everywhere :P

modules/wallet/transactionbuilder.go
```diff
@@ -634,3 +634,127 @@ func (w *Wallet) RegisterTransaction(t types.Transaction, parents []types.Transa
...
+// SignTransaction signs txn using secret keys derived from seed. toSign maps
+// the ParentID of each unsigned input to the UnlockHash of that input's
+// desired UnlockConditions. SignTransaction fills in the UnlockConditions for
+// each such input and adds a corresponding signature.
+//
+// SignTransaction must derive all of the keys from scratch, so it is
+// appreciably slower than calling the Wallet.SignTransaction method. Only the
+// first 1 million keys are derived.
+func SignTransaction(txn *types.Transaction, seed modules.Seed, toSign map[types.OutputID]types.UnlockHash) error {
```
ChrisSchinnerl: Would it makes sense to create a standalone utility too? Something like `siasign` instead of `siac` that keeps the keys in memory once generated and can be used to rapidly sign transactions without having to have a full locked full node installed?

modules/wallet/transactionbuilder_test.go
```diff
@@ -450,3 +450,64 @@ func TestParallelBuilders(t *testing.T) {
 		t.Fatal("did not get the expected ending balance", expected, endingSCConfirmed, startingSCConfirmed)
 	}
 }
+
+func TestSignTransaction(t *testing.T) {
```
ChrisSchinnerl: missing comment

modules/wallet/wallet.go
```diff
@@ -111,6 +111,7 @@ type Wallet struct {
 func (w *Wallet) Height() types.BlockHeight {
 	w.mu.Lock()
 	defer w.mu.Unlock()
+	w.syncDB()
```
ChrisSchinnerl: Why do we need to sync here?

node/api/routes.go
```diff
@@ -124,6 +124,8 @@ func (api *API) buildHTTPRoutes(requiredUserAgent string, requiredPassword strin
 		router.GET("/wallet/verify/address/:addr", api.walletVerifyAddressHandler)
 		router.POST("/wallet/unlock", RequirePassword(api.walletUnlockHandler, requiredPassword))
 		router.POST("/wallet/changepassword", RequirePassword(api.walletChangePasswordHandler, requiredPassword))
+		router.GET("/wallet/unspent", RequirePassword(api.walletUnspentHandler, requiredPassword))
+		router.POST("/wallet/sign", RequirePassword(api.walletSignHandler, requiredPassword))
```
ChrisSchinnerl: Could you add those to the `client` package as well?
It would also be cool to extend the `siatest` package right away to support this and have a `siatest` test as well.

modules/wallet/transactionbuilder.go
lukechampine: if someone wants to, it certainly wouldn't be very difficult. In practice I think most people will just use `siac`.

modules/wallet/transactionbuilder.go
lukechampine: it's more about the principle. If the wallet reports that it has an output, and then it crashes, it should still have that output next time you ask it. The general rule is: "don't give the user any data that won't survive a crash."

modules/wallet/wallet.go
lukechampine: see above -- don't want to report information that won't survive a crash.

modules/wallet/transactionbuilder.go
ChrisSchinnerl: I see :+1:

siatest/wallet/wallet_test.go
```diff
@@ -0,0 +1,89 @@
...
+	defer func() {
+		if err := testNode.Close(); err != nil {
+			t.Fatal(err)
+		}
+	}()
+
+	// get an output to spend
+	unspentResp, err := testNode.WalletUnspentGet()
+	if err != nil {
+		t.Fatal("failed to get spendable outputs")
```
leevlad: would be good to report `err` here, too

modules/wallet/transactionbuilder.go
```diff
@@ -678,6 +697,21 @@ outer:
...
+	// add unconfirmed outputs
+	for _, pt := range w.unconfirmedProcessedTransactions {
+		for _, o := range pt.Outputs {
+			if o.WalletAddress {
+				outputs = append(outputs, modules.SpendableOutput{
+					FundType:           types.SpecifierSiacoinOutput,
+					ID:                 o.ID,
+					UnlockHash:         o.RelatedAddress,
+					Value:              o.Value,
+					ConfirmationHeight: types.BlockHeight(math.MaxUint64), // unconfirmed
```
lukechampine: not sure what the best approach is here. I guess this is an argument for making the field `Confirmations` instead of `ConfirmationHeight`; then we could set `Confirmations: 0`. If we instead set `ConfirmationHeight` to some special value, the API client has to do special processing to calculate the number of confirmations.

modules/wallet/transactionbuilder.go
```diff
@@ -660,6 +662,23 @@ func (w *Wallet) SpendableOutputs() []modules.SpendableOutput {
 		})
 	})
 
+	// don't include outputs marked as spent in pending transactions
```
lukechampine: @DavidVorick this seemed like a reasonable thing to do, but would like to hear your thoughts

modules/wallet/transactionbuilder.go
```diff
@@ -678,6 +697,21 @@ outer:
 		}
 	}
 
+	// add unconfirmed outputs
```
lukechampine: ditto here

lukechampine:
I've updated the `/wallet/unspent` API so that `modules.SpendableOutput` includes the `UnlockConditions` instead of the `UnlockHash`. Originally I thought that the `UnlockConditions` should not be revealed, but @DavidVorick assured me that leaking these is not a big concern (they are revealed anyway when you spend them). This makes the `/wallet/sign` API much nicer, because you can construct the full input beforehand (no need for `/wallet/sign` to fill in the `UnlockConditions`), and you only need to specify a list of `OutputID`s to sign, rather than a map from `OutputID` to `UnlockHash`.

modules/wallet/transactionbuilder.go
DavidVorick: at one point we had siag which was essentially a key generator. I extended it to support signing and multisig as well, but I don't think that we ever released that extension.

I think having `siac` do everything is fine.

lmckenzie:
Is this the PR that HitBTC wants before allowing deposits/withdraws to its Sia wallets? Just curious because I would like to do what it takes to get my coin back.

magmel48:
is there any update about this PR?

DavidVorick:
@magmel48 we're currently working with the ledger hardware and development kit to make sure that these API endpoints are compatible with hardware wallets. We take compatibility pretty seriously at Sia, and we don't want to merge anything that we aren't completely confident in. Since offline signing has the most obvious use case of hardware wallets, we want to make sure that our first API support is actually sufficient to create a smooth hardware wallet experience.

This is not likely to be ready until v1.3.4, but when we merge it we'll probably also have support for the ledger hardware wallet ready to be released as well.

lmckenzie:
@magmel48 so I should expect my Sia wallet on HitBTC to be unavailable for 2-3 more months?

lukechampine:
I'm gonna think out loud here for a bit, Mostly as a way to clarify my own thoughts, but maybe this will be of interest to others as well.

The wallet generates keys by hashing its seed with a monotonically-incremented counter, called the index. So private key 0 is `H(seed|0)`, private key 1 is `H(seed|1)`, etc. Hashing is one-way, so if you're just given a private key, you can't figure out which seed and index was used to generate it. However, if you already know the seed, you can try hashing successive indices until you get one that matches the key. This works for public keys too: you can hash successive indices to get private keys, compute the public key from the private key, and then compare it to what you're searching for.

This is what the PR currently does. You give `SignTransaction` a list of `UnlockHash`s to look up, and it checks each index to see if there's a match, giving up after some threshold. This is feasible on a desktop CPU, because you can check many thousands of indices per second. But on a less powerful processor, this search process can take much longer. On a hardware wallet, it could take up to a minute. Clearly we need a different approach.

We can remove the need for this search process if we have a lookup table from `UnlockHash` to index. Then we can just compute the private key directly, no searching required. To facilitate this, all we really need to do is add an `index` field to `spendableKey`. When signing a transaction with an unlocked `siad` wallet, you'll call `SignTransaction` with a `map[types.OutputID]types.UnlockHash` to specify which inputs you'd like to sign. The wallet will lookup the `spendableKey` for each `UnlockHash`, fill in the `UnlockConditions`, then use the index to generate the keypair and sign.

The workflow for a hardware wallet is a bit different. When the device is first set up, a seed is generated and stored in persistent memory. (The seed is device-specific, not a standard Sia wallet seed, and should never leave the device.) To get an address, `siad` will ask the device to derive a public key using a supplied index. `siad` will then create a set of `UnlockConditions` that use the public key, pair these with the index, and store the result in the wallet db. 
When signing, you'll provide the same `map[types.OutputID]types.UnlockHash` as before. `siad` will look up the corresponding `UnlockConditions` and index in the wallet db, and send the index and hash to the device. The device will derive the key from the index, sign the hash, and return the signature, which `siad` will use to construct the full, signed transaction. This approach seems to match how Bitcoin hardware wallets work. In Go terms, the hardware wallet is just:
```go
type HardwareWallet interface {
    DeriveKey(index uint64) (pubkey [32]byte)
    SignHash(hash [32]byte, index uint64) (signature [64]byte)
}
```
I like this approach because it requires minimal code and storage on the hardware wallet itself. Smaller code means less bug surface, and the two-function interface is quite flexible. For example, you could use it to sign an arbitrary message with the same private key that controls an address.

This will require modifying the wallet db slightly to add a bucket that stores `UnlockConditions`+index pairs. We may also want to create a new type, distinct from `spendableKey`, since that type has historically been used for `siag` keys (which aren't derived from an index). Not too difficult.

Final note: David and I also discussed storing the `UnlockHash` -> index lookup table on the device itself (instead of in the wallet db). This saves some code in `siad`, but it quickly becomes infeasible due to the highly constrained storage space of such devices. The Ledger Nano S, for example, has only 320kb of persistent storage, so it could only store a few thousand addresses. We could use the brute-force strategy to trade time for storage space, but this isn't great either; no one wants to wait 30 seconds for their signature. These approaches also fail if you want to use special `UnlockConditions`, like a multi-sig or timelocked output.

DavidVorick:
Looks pretty good to me. I'm wondering if we should do a little extra work here to make sure that we can support HD keys in the future. I'm not sure what that would entail entirely, but if the hardware wallet code and interface is able to do HD keys, then we won't have to upgrade a bunch of legacy hardware wallets to add it to siad later.

lukechampine:
I'm reversing what I said earlier about unlock conditions: it's ok to require them. At the end of the day, if you want to spend an output, you need to know what its unlock conditions are. If you're using a watch-only wallet, you may or may not know the unlock conditions of the addresses you're tracking; that's okay. You might be using it as part of an offline-signing scheme, or you could just be interested in tracking the balances of a set of addresses. If you're in the latter group, `/wallet/unspent` just won't contain the unlock conditions.

The really nice thing about this decision is that it greatly simplifies the standard signing process. Since the user will have to fill in the unlock conditions themselves, `SignTransaction` can now calculate the unlock hashes itself. That means that it can automatically sign any inputs that it has keys for. I think this is how `bitcoind` does it: you just provide the partially-completed transaction, and `SignTransaction` automagically signs it. Of course, if you only want to sign a specific set of inputs, that's possible too. The automagic version is just a convenience.

There's still two missing pieces of the offline-signing puzzle. First, we need a true watch-only wallet. At bare minimum this means methods/api calls for loading addresses to watch (and optionally the unlock conditions along with them). Second, `SpendableOutput` needs to include the information used to derive the associated signing key, whether that ends up being a BIP44 path or a simple index. This is required by the Ledger app and other hardware wallets. Once we have those two things, it will be feasible to conduct a proper offline-signing scheme with a hardware wallet.

PR #2908, opened 2018-03-28
Use BlockHeight to calculate uptime/downtime instead of timestamps.

mikkeljuhl:
Use BlockHeight to calculate uptime/downtime instead of timestamps. `timePassed` is calculated by checking how many blocks have passed * 10 minutes, instead of using the timestamps. 
This is more reliable, if people should change their system time or something like that. 

This is a move towards making a `GlobalAverageUptime`.


modules/renter.go
```diff
@@ -119,8 +119,9 @@ type HostDBEntry struct {
 
 // HostDBScan represents a single scan event.
 type HostDBScan struct {
-	Timestamp time.Time `json:"timestamp"`
-	Success   bool      `json:"success"`
+	Timestamp   time.Time         `json:"timestamp"`
```
ChrisSchinnerl: @DavidVorick is there still a need to keep the Timestamp?

modules/renter/hostdb/hostweight.go
```diff
@@ -311,6 +312,7 @@ func (hdb *HostDB) uptimeAdjustments(entry modules.HostDBEntry) float64 {
 	downtime := entry.HistoricDowntime
 	uptime := entry.HistoricUptime
 	recentTime := entry.ScanHistory[0].Timestamp
```
ChrisSchinnerl: If we decide to get rid of the `Timestamp` field we can probably also get rid of that line and check the order later using the blockheight instead. We should wait for @DavidVorick to give some input on this though.

modules/renter/hostdb/hostweight.go
```diff
@@ -322,12 +324,19 @@ func (hdb *HostDB) uptimeAdjustments(entry modules.HostDBEntry) float64 {
...
 		}
+
+		blockDiff := scan.BlockHeight - recentBlockHeight
+		timePassed := float64(time.Duration(blockDiff) * 10 * time.Minute)
+		halftime := float64(1 * time.Hour * 24 * 30)
+		decay := math.Pow(0.5, timePassed/halftime)
+
 		if recentSuccess {
-			uptime += scan.Timestamp.Sub(recentTime)
+			uptime += time.Duration(timePassed * decay)
```
ChrisSchinnerl: This code here should do pretty much the same as the one in `updateEntry`. At a first glance it seems like `uptime` and `downtime` themselves are not decayed.

ChrisSchinnerl: Now that I think about it we can probably also change `uptime` `downtime` `historicUptime` and `historicDowntime` to be of type `blockheight` instead of `Duration` if we decide to do everything using blocks.

When we calculate the ratio later we would just cast them to `float64`.

modules/renter/hostdb/hostweight.go
```diff
@@ -322,12 +324,19 @@ func (hdb *HostDB) uptimeAdjustments(entry modules.HostDBEntry) float64 {
 			// Ignore the unsorted scan entry.
 			continue
 		}
+
+		blockDiff := scan.BlockHeight - recentBlockHeight
+		timePassed := float64(time.Duration(blockDiff) * 10 * time.Minute)
+		halftime := float64(1 * time.Hour * 24 * 30)
```
ChrisSchinnerl: halftime should be in `modules/renter/hostdb/consts.go` instead of being declared inline in 2 files.

modules/renter/hostdb/hostweight.go
mikkeljuhl: We shouldn't turn `historicUptime` and `historicDowntime` into type `BlockHeight`. 
Since, according to David we should time (and not block height) should be used for things (such as hosts) that lives outside the protocol

mikkeljuhl:
@ChrisSchinnerl I've addressed your comments, please review 👍

modules/renter.go
DavidVorick: I'm going to go ahead and say yes, we should keep the timestamp in the scan, just in case we find some use for it later. We may (or may not) phase it out eventually, but I can think of a few possible reasons we may want it at some point.

modules/renter/hostdb/consts.go
```diff
@@ -34,6 +34,9 @@ const (
 	// scans start getting compressed.
 	minScans = 12
 
+	// halftimeUpDownTime is the halftime used to decay the host up and downtime
+	halftimeUpDownTime = 30 * 24 * time.Hour
```
DavidVorick: I think the word that you want here is halflife. I'd probably call this `uptimeHalflife`

modules/renter/hostdb/hostweight.go
```diff
@@ -366,6 +362,22 @@ func (hdb *HostDB) uptimeAdjustments(entry modules.HostDBEntry) float64 {
 	return math.Pow(uptimeRatio, exp)
 }
 
+// decayHostUpOrDowntime decays a host's historic uptime or historic downtime.
+// It also adds the new block height to the historic uptime or historic downtime.
+func decayUptimeOrDowntime(entry *modules.HostDBEntry, scan modules.HostDBScan, recentScan modules.HostDBScan) {
```
DavidVorick: functions that don't have any receiver should be moved to the top of the file, and then sorted alphabetically among eachother.

cmd/siac/hostdbcmd.go
```diff
@@ -2,6 +2,9 @@ package main
 
 import (
 	"fmt"
+	"github.com/NebulousLabs/Sia/modules"
+	"github.com/NebulousLabs/Sia/node/api"
+	"github.com/spf13/cobra"
```
DavidVorick: were these added by mistake? There are no other changes to this file.

cmd/siac/hostcmd.go
```diff
@@ -13,6 +13,7 @@ import (
 	"github.com/NebulousLabs/Sia/node/api/client"
 	"github.com/NebulousLabs/Sia/types"
 
+	"github.com/NebulousLabs/Sia/node/api"
```
DavidVorick: were these added by mistake? There are no other changes to this file.

modules/renter/hostdb/hostweight.go
```diff
@@ -322,13 +322,9 @@ func (hdb *HostDB) uptimeAdjustments(entry modules.HostDBEntry) float64 {
...
 		}
-		if recentSuccess {
-			uptime += scan.Timestamp.Sub(recentTime)
-		} else {
-			downtime += scan.Timestamp.Sub(recentTime)
-		}
-		recentTime = scan.Timestamp
-		recentSuccess = scan.Success
+
+		decayUptimeOrDowntime(&entry, scan, recentScan)
```
DavidVorick: This is the wrong place to apply decay - you are applying decay every time that the weight is being calculated. But we can calculate the weight for a lot of different reasons. We should only be applying decay after we scan the host. That means the decay algorithm should also be moved to `scan.go`.

modules/renter/hostdb/hostweight.go
DavidVorick: as pointed out above, this function needs to be moved to scan.go

modules/renter/hostdb/hostweight.go
```diff
@@ -366,6 +362,22 @@ func (hdb *HostDB) uptimeAdjustments(entry modules.HostDBEntry) float64 {
...
+	timePassed := time.Duration(blocksPassed) * 10 * time.Minute
+	decay := time.Duration(math.Pow(0.5, float64(timePassed)/float64(halftimeUpDownTime)))
+
+	if recentScan.Success {
+		entry.HistoricUptime *= decay
+		entry.HistoricUptime += timePassed * decay
+	} else {
+		entry.HistoricDowntime *= decay
+		entry.HistoricDowntime += timePassed * decay
+	}
```
DavidVorick: you need to decay both the historic uptime and the historic downtime any time that you apply a decay. The decayed time passed should be added only to the relevant field. The result should look something like this:

```go
entry.HistoricUptime *= decay
entry.HistoricDowntime *= decay
if recentScan.Success {
    entry.HistoricUptime += timePassed*decay
} else {
    entry.HistoricDowntime += timePassed*decay
}
```

modules/renter/hostdb/scan.go
```diff
@@ -206,12 +207,7 @@ func (hdb *HostDB) updateEntry(entry modules.HostDBEntry, netErr error) {
 
 	// Compress any old scans into the historic values.
 	for len(newEntry.ScanHistory) > minScans && time.Now().Sub(newEntry.ScanHistory[0].Timestamp) > maxHostDowntime {
-		timePassed := newEntry.ScanHistory[1].Timestamp.Sub(newEntry.ScanHistory[0].Timestamp)
-		if newEntry.ScanHistory[0].Success {
-			newEntry.HistoricUptime += timePassed
-		} else {
-			newEntry.HistoricDowntime += timePassed
-		}
+		decayUptimeOrDowntime(&newEntry, newEntry.ScanHistory[1], newEntry.ScanHistory[0])
```
DavidVorick: Upon further reflection (sorry for waffling around), the name should really be `updateUptime`, because you aren't just applying a decay you are also adding new information.

mikkeljuhl:
@DavidVorick thanks for reviewing, your comments have been addressed.

tbenz9:
@mikkeljuhl Looks like a few of the regression tests are failing.  Do the tests pass when run locally?

PR #2910, opened 2018-03-29
Rename MaxAdjustmentDown and MaxAdjustmentUp

mharkus:
Renaming MaxAdjustmentDown and MaxAdjustmentUp to MaxTargetAdjustmentDown and MaxTargetAdjustmentUp respectively as suggested by @DavidVorick  here:

https://github.com/NebulousLabs/Sia/pull/2786

### DIFF
--- a/cmd/siad/server.go
+++ b/cmd/siad/server.go
@@ -79,8 +79,8 @@ type (
 		RootTarget types.Target `json:"roottarget"`
 		RootDepth  types.Target `json:"rootdepth"`
 
-		MaxAdjustmentUp   *big.Rat `json:"maxadjustmentup"`
-		MaxAdjustmentDown *big.Rat `json:"maxadjustmentdown"`
+		MaxTargetAdjustmentUp   *big.Rat `json:"maxtargetadjustmentup"`
+		MaxTargetAdjustmentDown *big.Rat `json:"maxtargetadjustmentdown"`
 
 		SiacoinPrecision types.Currency `json:"siacoinprecision"`
 	}
@@ -349,8 +349,8 @@ func (srv *Server) daemonConstantsHandler(w http.ResponseWriter, _ *http.Request
 		RootTarget: types.RootTarget,
 		RootDepth:  types.RootDepth,
 
-		MaxAdjustmentUp:   types.MaxAdjustmentUp,
-		MaxAdjustmentDown: types.MaxAdjustmentDown,
+		MaxTargetAdjustmentUp:   types.MaxTargetAdjustmentUp,
+		MaxTargetAdjustmentDown: types.MaxTargetAdjustmentDown,
 
 		SiacoinPrecision: types.SiacoinPrecision,
 	}
--- a/doc/api/Daemon.md
+++ b/doc/api/Daemon.md
@@ -83,10 +83,10 @@ returns the set of constants in use.
   "rootdepth": [255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255],
 
   // Largest allowed ratio between the old difficulty and the new difficulty.
-  "maxadjustmentup": "5/2",
+  "maxtargetadjustmentup": "5/2",
 
   // Smallest allowed ratio between the old difficulty and the new difficulty.
-  "maxadjustmentdown": "2/5",
+  "maxtargetadjustmentdown": "2/5",
 
   // Number of Hastings in one siacoin.
   "siacoinprecision": "1000000000000000000000000" // hastings per siacoin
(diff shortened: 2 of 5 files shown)


cmd/siad/server.go
```diff
@@ -79,8 +79,8 @@ type (
 		RootTarget types.Target `json:"roottarget"`
 		RootDepth  types.Target `json:"rootdepth"`
 
-		MaxAdjustmentUp   *big.Rat `json:"maxadjustmentup"`
-		MaxAdjustmentDown *big.Rat `json:"maxadjustmentdown"`
+		MaxTargetAdjustmentUp   *big.Rat `json:"maxtargetadjustmentup"`
+		MaxTargetAdjustmentDown *big.Rat `json:"maxtargetadjustmentdown"`
```
DavidVorick: Can't remove these entirely, need to keep the deprecated values `MaxAdjustmentUp` and `MaxAdjustmentDown`  that are identical to the newly named values, so that old programs don't experience a break in compatibility.

mharkus: @DavidVorick you mean just keep the json properties the same as before? i.e MaxTargetAdjustmentUp will use `json:"maxadjustmentup"` and MaxTargetAdjustmentDown will use `json:"maxadjustmentdown"`

DavidVorick:
```
	// SiaConstants is a struct listing all of the constants in use.
	SiaConstants struct {
		BlockFrequency         types.BlockHeight `json:"blockfrequency"`
		BlockSizeLimit         uint64            `json:"blocksizelimit"`
		ExtremeFutureThreshold types.Timestamp   `json:"extremefuturethreshold"`
		FutureThreshold        types.Timestamp   `json:"futurethreshold"`
		GenesisTimestamp       types.Timestamp   `json:"genesistimestamp"`
		MaturityDelay          types.BlockHeight `json:"maturitydelay"`
		MedianTimestampWindow  uint64            `json:"mediantimestampwindow"`
		SiafundCount           types.Currency    `json:"siafundcount"`
		SiafundPortion         *big.Rat          `json:"siafundportion"`
		TargetWindow           types.BlockHeight `json:"targetwindow"`

		InitialCoinbase uint64 `json:"initialcoinbase"`
		MinimumCoinbase uint64 `json:"minimumcoinbase"`

		RootTarget types.Target `json:"roottarget"`
		RootDepth  types.Target `json:"rootdepth"`

		MaxAdjustmentUp   *big.Rat `json:"maxadjustmentup"`
		MaxAdjustmentDown *big.Rat `json:"maxadjustmentdown"`
		MaxTargetAdjustmentUp   *big.Rat `json:"maxtargetadjustmentup"`
		MaxTargetAdjustmentDown *big.Rat `json:"maxtargetadjustmentdown"`

		SiacoinPrecision types.Currency `json:"siacoinprecision"`
}
```

Change the struct to this, that way the new fields and old fields are both supported. We have to use duplication when we change a field name to preserve compatibility and also provide the fixed endpoint. We do this in a few other places in the API too.

cmd/siad/server.go
```diff
@@ -349,8 +349,8 @@ func (srv *Server) daemonConstantsHandler(w http.ResponseWriter, _ *http.Request
 		RootTarget: types.RootTarget,
 		RootDepth:  types.RootDepth,
 
-		MaxAdjustmentUp:   types.MaxAdjustmentUp,
-		MaxAdjustmentDown: types.MaxAdjustmentDown,
+		MaxTargetAdjustmentUp:   types.MaxTargetAdjustmentUp,
+		MaxTargetAdjustmentDown: types.MaxTargetAdjustmentDown,
```
DavidVorick:
```
		MaxAdjustmentUp:   types.MaxTargetAdjustmentUp,
		MaxAdjustmentDown: types.MaxTargetAdjustmentDown,
		MaxTargetAdjustmentUp:   types.MaxTargetAdjustmentUp,
		MaxTargetAdjustmentDown: types.MaxTargetAdjustmentDown,
```

Because we have both fields in the json version, we need to set both fields here.

doc/api/Daemon.md
```diff
@@ -83,10 +83,10 @@ returns the set of constants in use.
   "rootdepth": [255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255,255],
 
   // Largest allowed ratio between the old difficulty and the new difficulty.
-  "maxadjustmentup": "5/2",
+  "maxtargetadjustmentup": "5/2",
 
   // Smallest allowed ratio between the old difficulty and the new difficulty.
-  "maxadjustmentdown": "2/5",
+  "maxtargetadjustmentdown": "2/5",
```
DavidVorick: this is fine, we don't need to document the deprecated fields, so we don't need to have the duplication here even though we need the duplication in the actual code itself.

Issue #2465, opened 2017-11-08
Add option in siac to display host contracts.


DavidVorick:
I'm not sure if this is even supported in the API yet, we may need to extend the API. But we need to be able to see all of the contracts that a host is currently using, which will help with debugging and also just help users be more generally aware of the status of their hosts.

hudaniel:
Are there any dev environments we can use to test a wallet that accepted contracts?

lukechampine:
@hudaniel you can setup a local testnet with https://github.com/NebulousLabs/Sia-Ant-Farm, though it isn't super well-documented

hudaniel:
I'm getting this error when trying to `make dependencies`
```
$ go install -tags='dev' github.com/NebulousLabs/Sia/siad
can't load package: package github.com/NebulousLabs/Sia/siad: cannot find package "github.com/NebulousLabs/Sia/siad" in any of:
    /usr/local/Cellar/go/1.9.2/libexec/src/github.com/NebulousLabs/Sia/siad (from $GOROOT)
    /Users/Daniel/code/golang/src/github.com/NebulousLabs/Sia/siad (from $GOPATH)
``` 
any ideas?

lukechampine:
hmm, `go install -tags='dev' github.com/NebulousLabs/Sia/siad` is the wrong command. The `siad` package has been moved to the `cmd` subfolder. I've pushed the fix.

S-anasol:
In case someone want to imlement it:
https://github.com/S-anasol/Sia/blob/obligationsapi/api/host.go#L383-L387
Very simple api method to retrieve host contracts data.

UI for json exported with this api call:
https://github.com/S-anasol/hostdbviewer

UI Example/Tester/Viewer:
https://s-anasol.github.io/hostdbviewer/#/

tbenz9:
@nielscastien are you planning on adding your new API endpoints into a `siac` command?

nielscastien:
@tbenz9 No, but if one of the devs (or you?) can point me in the right direction on how to present and/or filter the data, I am willing to give it a try. There are quite some fields in the contract so I'm afraid that showing all will not work for the end user.

nielscastien:
Something like `siac host contracts -s` (`-s` flag for sorting on expiration height):

![image](https://user-images.githubusercontent.com/22688069/38930361-f6b87fd0-430f-11e8-9651-df6de1cff27e.png)

nielscastien:
or including locked and risked collateral:

[...]

PR #2923, opened 2018-04-06
[WIP] Host contract API

nielscastien:
This PR introduces an API call to get basic information about stored contracts on the host.

Usage: `curl -A "Sia-Agent" "localhost:9980/host/contracts?status=xxxx"`
where `xxxx` can any of the following:

1. `unresolved`
1. `succeeded`
1. `failed`
1. `rejected`
1. `all`

First step to implement #2465

### DIFF
--- a/modules/host.go
+++ b/modules/host.go
@@ -111,10 +111,29 @@ type (
 	// StorageObligation contains information about a storage obligation that
 	// the host has accepted.
 	StorageObligation struct {
-		NegotiationHeight types.BlockHeight `json:"negotiationheight"`
+		ObligationId             types.FileContractID `json:"obligationid"`
+		FileSize                 uint64               `json:"filesize"`
+		SectorRootsCount         uint64               `json:"sectorrootscount"`
+		ContractCost             types.Currency       `json:"contractcost"`
+		LockedCollateral         types.Currency       `json:"lockedcollateral"`
+		PotentialDownloadRevenue types.Currency       `json:"potentialdownloadrevenue"`
+		PotentialStorageRevenue  types.Currency       `json:"potentialstoragerevenue"`
+		PotentialUploadRevenue   types.Currency       `json:"potentialuploadrevenue"`
+		RiskedCollateral         types.Currency       `json:"riskedcollateral"`
+		TransactionFeesAdded     types.Currency       `json:"transactionfeesadded"`
 
+		// The negotiation height specifies the block height at which the file
+		// contract was negotiated. The expiration height and the proof deadline
+		// are equal to the window start and window end. Between the expiration height
+		// and the proof deadline, the host must submit the storage proof.
+		NegotiationHeight types.BlockHeight `json:"negotiationheight"`
+		ExpirationHeight  types.BlockHeight `json:"expirationheight"`
+		ProofDeadLine     types.BlockHeight `json:"proofdeadline"`
+
+		// Variables indicating whether the critical transactions in a storage
+		// obligation have been confirmed on the blockchain.
 		OriginConfirmed     bool   `json:"originconfirmed"`
-		RevisionConstructed bool   `json:"revisionconstructed"`
+		RevisionConstructed bool   `json:"revisioncontructed"`
 		RevisionConfirmed   bool   `json:"revisionconfirmed"`
 		ProofConstructed    bool   `json:"proofconstructed"`
 		ProofConfirmed      bool   `json:"proofconfirmed"`
--- a/modules/host/storageobligations.go
+++ b/modules/host/storageobligations.go
@@ -870,7 +870,20 @@ func (h *Host) StorageObligations() (sos []modules.StorageObligation) {
 				return build.ExtendErr("unable to unmarshal storage obligation:", err)
 			}
 			mso := modules.StorageObligation{
+				ObligationId:             so.id(),
+				FileSize:                 so.fileSize(),
+				SectorRootsCount:         uint64(len(so.SectorRoots)),
+				ContractCost:             so.ContractCost,
+				LockedCollateral:         so.LockedCollateral,
+				PotentialDownloadRevenue: so.PotentialDownloadRevenue,
+				PotentialStorageRevenue:  so.PotentialStorageRevenue,
+				PotentialUploadRevenue:   so.PotentialUploadRevenue,
+				RiskedCollateral:         so.RiskedCollateral,
+				TransactionFeesAdded:     so.TransactionFeesAdded,
+
 				NegotiationHeight: so.NegotiationHeight,
+				ExpirationHeight:  so.expiration(),
+				ProofDeadLine:     so.proofDeadline(),
 
 				OriginConfirmed:     so.OriginConfirmed,
 				RevisionConstructed: so.RevisionConstructed,
--- a/node/api/host.go
+++ b/node/api/host.go
@@ -24,6 +24,12 @@ var (
 )
 
 type (
+	// ContractInfoGET contains the information that is returned after a GET request
+	// to /host/contracts - information for the host about stored obligations.
+	ContractInfoGET struct {
+		Contracts []modules.StorageObligation `json:"obligations"`
+	}
+
 	// HostGET contains the information that is returned after a GET request to
 	// /host - a bunch of information about the status of the host.
 	HostGET struct {
--- a/node/api/host.go
+++ b/node/api/host.go
@@ -67,6 +67,44 @@ func folderIndex(folderPath string, storageFolders []modules.StorageFolderMetada
 	return -1, errStorageFolderNotFound
 }
 
+// hostContractInfoHandler handles the API call to get the contract information of the host.
+// Information is retrieved via the storage obligations from the host database.
+// TODO: filters are hard coded. Adding or removing storage obligation statuses will
+// currently break the API.
+func (api *API) hostContractInfoHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	statusStr := req.FormValue("status")
+	if statusStr == "" {
+		WriteError(w, Error{"Status must be provided to a /host/contracts call."}, http.StatusBadRequest)
+		return
+	}
+	var filter uint64
+	switch statusStr {
+	case "unresolved":
+		filter = 0
+	case "rejected":
+		filter = 1
+	case "succeeded":
+		filter = 2
+	case "failed":
+		filter = 3
+	case "all":
+		filter = 4
+	default:
+		WriteError(w, Error{"Unable to parse contract status."}, http.StatusBadRequest)
+		return
+	}
+	sos := []modules.StorageObligation{}
+	for _, so := range api.host.StorageObligations() {
+		if so.ObligationStatus == filter || filter == 4 {
+			sos = append(sos, so)
+		}
+	}
+	cg := ContractInfoGET{
+		Contracts: sos,
+	}
+	WriteJSON(w, cg)
+}
+
 // hostHandlerGET handles GET requests to the /host API endpoint, returning key
 // information about the host.
 func (api *API) hostHandlerGET(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
--- a/node/api/routes.go
+++ b/node/api/routes.go
@@ -45,6 +45,7 @@ func (api *API) buildHTTPRoutes(requiredUserAgent string, requiredPassword strin
 		router.GET("/host", api.hostHandlerGET)                                                   // Get the host status.
 		router.POST("/host", RequirePassword(api.hostHandlerPOST, requiredPassword))              // Change the settings of the host.
 		router.POST("/host/announce", RequirePassword(api.hostAnnounceHandler, requiredPassword)) // Announce the host to the network.
+		router.GET("/host/contracts", api.hostContractInfoHandler)                                // Get info about contracts.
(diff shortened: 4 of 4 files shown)


nielscastien:
@ChrisSchinnerl Please review.

node/api/host.go
```diff
@@ -24,6 +24,12 @@ var (
 )
 
 type (
+	// ContractInfoGET contains the information that is returned after a GET request
+	// to /host/contracts - information for the host about stored obligations.
+	ContractInfoGET struct {
+		Contracts []modules.StorageObligation `json:"obligations"`
```
lukechampine: generally these names should match -- either the Go field should be named `Obligations`, or the JSON field should be named `contracts`.

node/api/host.go
```diff
@@ -61,6 +67,44 @@ func folderIndex(folderPath string, storageFolders []modules.StorageFolderMetada
...
+	case "all":
+		filter = 4
+	default:
+		WriteError(w, Error{"Unable to parse contract status."}, http.StatusBadRequest)
+		return
+	}
+	sos := []modules.StorageObligation{}
+	for _, so := range api.host.StorageObligations() {
+		if so.ObligationStatus == filter || filter == 4 {
+			sos = append(sos, so)
```
lukechampine: I'm wondering if it wouldn't be simpler to just return all of the obligations, and let the caller filter them.
The tradeoff, of course, is that you may end up sending a ton of obligations when the caller only cares about a few. But as long as it's not a prohibitively large amount (i.e. thousands of large JSON objects), it's likely not going to be an issue. Also, if it does become an issue, we'd probably want to making the API paginated rather than filtering based on status. Consider that, even if we do filter based on status, we'd still end up needing pagination in the long run anyway, because a status like `all` is going to return too much data.

nielscastien: I agree it is simpler to just return all the obligations. However, since the ObligationStatus is an `int`, I'm not sure how the caller would know the "real" status of the obligations. That's why I chose the named filter where it is clear for the caller what he receives.

node/api/host.go
nielscastien: Thanks, I will update the JSON to `contracts`

node/api/host.go
lukechampine: It doesn't look like the `modules.StorageObligation` struct is being used anywhere except tests, so it's not bound by compatibility requirements. That means we can change the `ObligationStatus` field to a `string` (or maybe a `types.Specifier`?). Unless @DavidVorick has objections, I think this is the right approach to take. Something like:
```go
// modules/host.go

const (
    ObligationStatusSucceeded = "succeeded"    
    ObligationStatusRejected  = "rejected"
    // ... etc
)

type StorageObligation struct {
    // ... etc
    ObligationStatus string `json:"obligationstatus"`
}
```

nielscastien: @lukechampine implementing your review comments and I would like to use a switch statement in the `StorageObligations` function but I'm not sure how to throw an error when there is no match:
```go
// modules/host/storageobligations.go

switch so.StorageObligationStatus {
case obligationSucceeded:
    varString = modules.ObligationStatusSucceeded
case obligationRejected:
    varString = modules.ObligationStatusRejected
// ... etc
default:
    // how to error here, use return build.Err?
}
```
any thoughts on this?

lukechampine: if there's no match, the data should be treated as corrupted. I would call `h.log.Severe("WARN: invalid obligation status")` (with other helpful info included) and then skip the obligation entirely, moving on top the next one.

Also, a map may be more readable than a switch:
```go
varString := map[uint64]string{
    obligationSucceeded: modules.ObligationStatusSucceeded,
    obligationRejected:  modules.ObligationStatusRejected,
}[so.StorageObligationStatus]
if varString == "" {
    // handle invalid status
}
```

modules/host/storageobligations.go
```diff
@@ -869,15 +869,38 @@ func (h *Host) StorageObligations() (sos []modules.StorageObligation) {
...
 			}
+			statusString := map[storageObligationStatus]string{
+				obligationFailed:     modules.ObligationStatusFailed,
+				obligationRejected:   modules.ObligationStatusRejected,
+				obligationSucceeded:  modules.ObligationStatusSucceeded,
+				obligationUnresolved: modules.ObligationStatusUnresolved,
+			}[so.ObligationStatus]
+			if statusString == "" {
+				h.log.Severe("WARN: invalid obligation status encountered when reading database.")
+				return build.ExtendErr("unable to parse obligation status.", err)
```
lukechampine: `err` will be `nil` here. My preference would just be to `return nil`. That way, the obligation is skipped, but others are still added. This keeps everything mostly-functional, but the log message will still indicate that something is wrong.

lukechampine:
LGTM, although @DavidVorick should weigh in on the exact selection of fields exposed in the API.

modules/host.go
```diff
@@ -111,14 +119,33 @@ type (
 	// StorageObligation contains information about a storage obligation that
 	// the host has accepted.
 	StorageObligation struct {
+		ObligationId             types.FileContractID `json:"obligationid"`
+		FileSize                 uint64               `json:"filesize"`
```
DavidVorick: FileSize is the wrong name here, it's the data size. A contract houses many partial files, not a single file.

modules/host.go
```diff
@@ -111,14 +119,33 @@ type (
...
+		SectorRootsCount         uint64               `json:"sectorrootscount"`
+		ContractCost             types.Currency       `json:"contractcost"`
+		LockedCollateral         types.Currency       `json:"lockedcollateral"`
+		PotentialDownloadRevenue types.Currency       `json:"potentialdownloadrevenue"`
+		PotentialStorageRevenue  types.Currency       `json:"potentialstoragerevenue"`
+		PotentialUploadRevenue   types.Currency       `json:"potentialuploadrevenue"`
+		RiskedCollateral         types.Currency       `json:"riskedcollateral"`
+		TransactionFeesAdded     types.Currency       `json:"transactionfeesadded"`
```
DavidVorick: fields within a block should be sorted. A few of the fields in this block are not sorted

modules/host.go
```diff
@@ -111,14 +119,33 @@ type (
...
+
+		// The negotiation height specifies the block height at which the file
+		// contract was negotiated. The expiration height and the proof deadline
+		// are equal to the window start and window end. Between the expiration height
+		// and the proof deadline, the host must submit the storage proof.
 		NegotiationHeight types.BlockHeight `json:"negotiationheight"`
+		ExpirationHeight  types.BlockHeight `json:"expirationheight"`
+		ProofDeadLine     types.BlockHeight `json:"proofdeadline"`
```
DavidVorick: Same here, `ExpirationHeight` should appear above `NegotiationHeight` so that all the fields in the codeblock are sorted.

modules/host/storageobligations.go
```diff
@@ -869,15 +869,38 @@ func (h *Host) StorageObligations() (sos []modules.StorageObligation) {
 			if err != nil {
 				return build.ExtendErr("unable to unmarshal storage obligation:", err)
 			}
+			statusString := map[storageObligationStatus]string{
```
DavidVorick: Can we achieve this in a cleaner way by adding a `String()` method to the `storageObligationStatus` type?

DavidVorick:
Any pull request that extends the api needs to add documentation to doc/api.md and the relevant file in doc/api/

Also, any pull request that extends the api needs to add a test to the `node/api` package which calls the api endpoint and checks that every field is returning a sane value. In this case, I think you could piggyback off of one of the tests in renterhost.go, and just add some checking in the middle and at the end of one of the more full tests which inspects the api endpoint and validates that all the sane values are being returned.

Overall the general concept of this PR seems okay to me.

modules/host/storageobligations.go
lukechampine: I considered that too. I think it would be an improvement, although not a huge one, because `storageObligationStatus` isn't exported.

lukechampine: btw, this is what `stringer` was created to do: https://godoc.org/golang.org/x/tools/cmd/stringer

lukechampine:
@DavidVorick any concerns about exposes too many fields via the API?

nielscastien:
@lukechampine I think I implemented all the review comments. Not sure if the testing is complete so if there is anything more that I should do, please let me know.

modules/host/storageobligationstatus_string.go
```diff
@@ -0,0 +1,16 @@
+// Code generated by "stringer -type=storageObligationStatus"; DO NOT EDIT.
+
+package host
```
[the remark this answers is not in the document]
lukechampine: This is standard output for the `stringer` tool.

That said, I don't really want to add `stringer`-generated files if we can avoid it. Sia doesn't have any code-gen step right now, and I'd like to keep it that way unless there's a really compelling reason.

I'd prefer to write the `String` method by hand and place it in storageobligations.go.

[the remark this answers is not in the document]
DavidVorick: oh.

I would also prefer to avoid any autogen code.

doc/api/Host.md
```diff
@@ -432,6 +433,75 @@ netaddress string // Optional
...
+    "potentialuploadrevenue":	"1234",		// hastings
+
+    // Amount that the host might lose if the submission of the storage proof is not successful.
+    "riskedcollateral":		"1234",		// hastings
+
+    // Amount for transaction fees that the host added to the storage obligation.
+    "transactionfeesadded":	"1234",		// hastings
+
+    // Experation height is the height at which the storage obligation expires.
+    "experationheight":		123456,		// blocks
```
lukechampine: typo -- `expiration`

doc/api/Host.md
```diff
@@ -432,6 +433,75 @@ netaddress string // Optional
...
+
+    // Negotion height is the height at which the storage obligation was negotiated.
+    "negotionheight":		0,		// blocks
```
lukechampine: negotiation

doc/API.md
```diff
@@ -392,11 +393,45 @@ netaddress string // Optional
...
+      "obligationid":			"fff48010dcbbd6ba7ffd41bc4b25a3634ee58bbf688d2f06b7d5a0c837304e13",
+      "sectorrootscount":		2,
+      "potentialdownloadrevenue":	"1234",		// hastings
+      "potentialstoragerevenue":	"1234",		// hastings
+      "potentialuploadrevenue":		"1234",		// hastings
+      "riskedcollateral":		"1234",		// hastings
+      "transactionfeesadded":		"1234",		// hastings
+
+      "expirationheight":		123456,		// blocks
+      "negotionheight":			123456,		// blocks
```
lukechampine: typos

doc/API.md
```diff
@@ -392,11 +393,45 @@ netaddress string // Optional
...
+###### JSON Response [(with comments)](/doc/api/Host.md#json-response-1)
+```javascript
+{
+  "contracts": [
+    {
+      "contractcost":			"1234",		// hastings
+      "datasize":			500000,		// bytes
+      "lockedcollateral":		"1234",		// hastings
+      "obligationid":			"fff48010dcbbd6ba7ffd41bc4b25a3634ee58bbf688d2f06b7d5a0c837304e13",
+      "sectorrootscount":		2,
```
DavidVorick: S comes after p

DavidVorick: And r lol

modules/host.go
```diff
@@ -111,14 +111,33 @@ type (
 	// StorageObligation contains information about a storage obligation that
 	// the host has accepted.
 	StorageObligation struct {
+		ContractCost             types.Currency       `json:"contractcost"`
+		DataSize                 uint64               `json:"datasize"`
+		LockedCollateral         types.Currency       `json:"lockedcollateral"`
+		ObligationId             types.FileContractID `json:"obligationid"`
+		SectorRootsCount         uint64               `json:"sectorrootscount"`
```
DavidVorick: Unsorted

DavidVorick:
Just a few nitpicks left, but this looks more or less ready to go

node/api/renterhost_test.go
```diff
@@ -199,11 +216,34 @@ func TestHostAndRentVanilla(t *testing.T) {
 	}
 
 	// Check the host, who should now be reporting file contracts.
-	//
-	// TODO: Switch to using an API call.
-	obligations := st.host.StorageObligations()
-	if len(obligations) != 1 {
-		t.Error("Host has wrong number of obligations:", len(obligations))
+	var cts ContractInfoGET
+	err = st.getAPI("/host/contracts", &cts)
```
ChrisSchinnerl: Since we are adding a new endpoint, I'd like to see `node/api/client/host.go` extended to support the `/host/contracts` endpoint by adding a `HostContractsGet` method.

@lukechampine @DavidVorick do you think we should move this test to the `siatest` package while we're at it instead of extending an existing test? That would implicitly confirm that `HostContractsGet` is working.

tbenz9:
YAY!  Great work @nielscastien thanks for sticking with it!  This will be a huge help for hosts going forward!

nielscastien:
@lukechampine  @tbenz9 No problem, glad I could help!

DavidVorick:
Definitely appreciate that you stuck through all the back-and-forth, it's important that we get the API right. I think a lot of hosts will appreciate having this endpoint.

PR #2926, opened 2018-04-07
only add addrs relevant to the wallet

lukechampine:
Previously, all addresses in transactions relevant to the wallet would be added to the `bucketAddrTransactions` bucket, which maps addresses to the transactions they appear in.
This caused problems due to the "zero address" (`types.UnlockHash{}`) appearing in transactions. Since [miner fee outputs leave the `RelatedAddress` field empty](https://github.com/NebulousLabs/Sia/blob/53b53997d16ec8b9865ea0cb469c31be44e6b2ca/modules/wallet/update.go#L384-L389) (i.e. zero), the mapping for the zero address effectively maps to every transaction relevant to the wallet that has miner fees.
For really high-volume wallets, like exchanges, this caused the zero address to map to over 625,000 transactions. This subsequently led to a bug where the mapping could not be decoded, as 625k * (8 bytes per mapping) is the [maximum allowed size](https://github.com/NebulousLabs/Sia/blob/53b53997d16ec8b9865ea0cb469c31be44e6b2ca/encoding/marshal.go#L15-L23) of an encoded value.

This PR fixes the issue by only adding addresses to `bucketAddrTransactions` if they are relevant to the wallet. This means you will no longer be able to query the `AddressTransactions` method for addresses not relevant to the wallet, but that seems like a non-issue, given that the wallet only stores transactions relevant to it in the first place. If you need to search for non-owned addresses, you'll have to use an explorer.

### DIFF
--- a/modules/wallet/database.go
+++ b/modules/wallet/database.go
@@ -262,16 +262,17 @@ func dbAddAddrTransaction(tx *bolt.Tx, addr types.UnlockHash, txn uint64) error 
 func dbAddProcessedTransactionAddrs(tx *bolt.Tx, pt modules.ProcessedTransaction, txn uint64) error {
 	addrs := make(map[types.UnlockHash]struct{})
 	for _, input := range pt.Inputs {
-		if input.WalletAddress {
-			addrs[input.RelatedAddress] = struct{}{}
-		}
+		addrs[input.RelatedAddress] = struct{}{}
 	}
 	for _, output := range pt.Outputs {
-		if output.WalletAddress {
-			addrs[output.RelatedAddress] = struct{}{}
-		}
+		addrs[output.RelatedAddress] = struct{}{}
 	}
 	for addr := range addrs {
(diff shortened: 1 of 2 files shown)


lukechampine:
After some discussion, we decided that it's too restrictive to only map addresses owned by the wallet. This would mean that if you sent some coins to `addr`, you couldn't lookup that transaction by referencing `addr`; you'd need to know which address you sent *from*, which can be difficult to know.

Instead, only the zero address is ignored now. No other address should have the same amount of volume, so this should resolve the large wallet issue.

lukechampine:
After further discussion, we decided that filtering out the zero address is too restrictive, and breaks compatibility. Instead, we will simply skip miner fees. It was incorrect to treat miner fees like other inputs; they are really a different sort of value, just smushed into the `ProcessedOutput` box for convenience.

Hopefully no additional discussion is required. :P

PR #2927, opened 2018-04-08
update 'make cover'

DavidVorick:


Makefile
```diff
@@ -103,13 +103,18 @@ test-mem:
...
 	@mkdir -p cover/modules/host
-	@for package in $(pkgs); do                                                                                                 \
-		go test -tags='testing debug' -timeout=500s -covermode=atomic -coverprofile=cover/$$package.out ./$$package -run=$(run) \
-		&& go tool cover -html=cover/$$package.out -o=cover/$$package.html                                                      \
-		&& rm cover/$$package.out ;                                                                                             \
+	@mkdir -p cover/node
+	@mkdir -p cover/node/api
+	@mkdir -p cover/siatest
+	@for package in $(pkgs); do                                                                                                       \
+		go test -tags='testing debug netgo' -timeout=500s -covermode=atomic -coverprofile=cover/$$package.out ./$$package -run=$(run) \
```
lukechampine: might this work instead?
```
@for package in $(pkgs); do
    @mkdir -p cover/$$package
```

DavidVorick: unfortunately not. We only `mkdir` the packages that have subpackages. There might be some function we can apply to `pkgs` though that will return only the packages that have subpackages. Maybe trim everything after the final `/` and then filter out any duplicates?

lukechampine: looks like we could use `dirname` to get the parent directory, i.e.
```bash
@mkdir -p $(dirname cover/$$package)
```
for top-level packages, this just runs `mkdir -p cover`, which is a no-op. For e.g. `./renter/proto`, it runs `mkdir -p cover/renter`.

DavidVorick: that's not working, it's interpreting `$$package` as a string instead of a variable

DavidVorick: `@mkdir -p $(shell dirname cover/$$package)` doesn't work either

tbenz9: Pretty sure this is what you want:
```
mkdir -p `dirname cover/$$package`
```

DavidVorick:
updated

PR #2931, opened 2018-04-09
Persist contract utilities

ChrisSchinnerl:


modules/renter/contractor/contracts.go
```diff
@@ -28,6 +28,16 @@ func (c *Contractor) contractEndHeight() types.BlockHeight {
 	return c.currentPeriod + c.allowance.Period
 }
 
+// ContractUtility returns the ContractUtility for a contract with a given id.
+func (c *Contractor) contractUtility(id types.FileContractID) modules.ContractUtility {
+	sc, exists := c.contracts.Acquire(c.resolveID(id))
```
lukechampine: It shouldn't be necessary to `Acquire` the contract to view its utility; `View` should work fine.

modules/renter/contractor/contractor.go
```diff
@@ -135,10 +130,12 @@ func (c *Contractor) Contracts() []modules.RenterContract {
 
 // ContractUtility returns the utility fields for the given contract.
 func (c *Contractor) ContractUtility(id types.FileContractID) (modules.ContractUtility, bool) {
-	c.mu.RLock()
-	utility, exists := c.contractUtilities[c.resolveID(id)]
-	c.mu.RUnlock()
-	return utility, exists
+	sc, exists := c.contracts.Acquire(id)
```
lukechampine: should be `View` here as well

modules/renter/contractor/contracts.go
```diff
@@ -28,6 +28,15 @@ func (c *Contractor) contractEndHeight() types.BlockHeight {
 	return c.currentPeriod + c.allowance.Period
 }
 
+// ContractUtility returns the ContractUtility for a contract with a given id.
+func (c *Contractor) contractUtility(id types.FileContractID) modules.ContractUtility {
```
lukechampine: the return value here is inconsistent with `ContractUtility`. Either both should return a boolean, or neither. I prefer to return the boolean. Then `ContractUtility` can just call `contractUtility`

modules/renter/proto/contractset.go
```diff
@@ -49,6 +49,20 @@ func (cs *ContractSet) Acquire(id types.FileContractID) (*SafeContract, bool) {
...
+	cs.mu.Lock()
+	safeContract, ok := cs.contracts[id]
+	cs.mu.Unlock()
+	if !ok {
+		return errors.New("can't update non-existing contract")
+	}
+	safeContract.mu.Lock()
+	defer safeContract.mu.Unlock()
+	return safeContract.updateUtility(utility)
+}
```
lukechampine: maybe it makes more sense to make `UpdateUtility` a method on a `SafeContract` that you `Acquire`? Not sure, since the other methods that modify `SafeContract` aren't exported.

lukechampine: or perhaps have `cs.UpdateContractUtility` take a `SafeContract` as an argument, like `Delete` and `Renew`.

modules/renter/contractor/contractor.go
```diff
@@ -135,10 +130,7 @@ func (c *Contractor) Contracts() []modules.RenterContract {
 
 // ContractUtility returns the utility fields for the given contract.
 func (c *Contractor) ContractUtility(id types.FileContractID) (modules.ContractUtility, bool) {
-	c.mu.RLock()
-	utility, exists := c.contractUtilities[c.resolveID(id)]
-	c.mu.RUnlock()
-	return utility, exists
+	return c.contractUtility(id)
```
DavidVorick: convention violation - need to call `contractUtility` holding a lock. If there's no locking, it needs to be called `managedContractUtility`

modules/renter/contractor/contractor.go
```diff
@@ -248,7 +239,9 @@ func NewCustomContractor(cs consensusSet, w wallet, tp transactionPool, hdb host
 	}
 
 	// Mark contract utility.
-	c.managedMarkContractsUtility()
+	if err := c.managedMarkContractsUtility(); err != nil {
+		return nil, err
```
DavidVorick: I'm not sure that we want to return an error here, that means the contractor won't start up and that we won't be able to make one if there's an error setting the utility.

DavidVorick: Since we're now intending to persist the utility values, it's probably fine to just launch the threadedMaintenance function at startup and not worry about this at startup.

DavidVorick: On a related note, at startup we aren't calling tg.Stop() if the startup fails. I'm pretty sure that's a bug in the code, the contractor should be shut down fully if it encounters an error in the `New` function. Maybe we can make that a separate pull request.

modules/renter/contractor/contracts.go
```diff
@@ -28,6 +29,15 @@ func (c *Contractor) contractEndHeight() types.BlockHeight {
 	return c.currentPeriod + c.allowance.Period
 }
 
+// ContractUtility returns the ContractUtility for a contract with a given id.
+func (c *Contractor) contractUtility(id types.FileContractID) (modules.ContractUtility, bool) {
```
DavidVorick: docstring has wrong func name, and I think we actually want this to be `managedContractUtility`

modules/renter/contractor/contracts.go
```diff
@@ -609,3 +635,14 @@ func (c *Contractor) threadedContractMaintenance() {
 		}
 	}
 }
+
+// updateContractUtility is a helper function that acquires a contract, updates
+// its ContractUtility and returns the contract again.
+func (c *Contractor) updateContractUtility(id types.FileContractID, utility modules.ContractUtility) error {
+	sc, ok := c.contracts.Acquire(id)
```
DavidVorick: no context here for why `sc` is the variable name, need to call it safeContract

DavidVorick:
I think that we want to add update code. Since we're tweaking things to have persisted utility values, we don't need to block while we fetch the utility values at startup. But that means that when you upgrade, you'll default to all 'bad' utility values, which is probably not good.

Instead, if you are upgrading for the first time, you should probably block until the correct utility values are available. That means you'll have to keep calling 'managedMarkUtility` until it's not returning an 'hostdb has not completed initial scan` error. And then you'll be good to unblock and let the user access their system.

For most people, this won't take long because the hostdb will have already completed its initial scan.

modules/renter/contractor/contractor.go
ChrisSchinnerl: Can we make it `staticContractUtility` ? It doesn't really matter if you hold the contractor lock or not.

DavidVorick:
It seems that the persisted contract header is not json, which means that we'll need to upgrade the persist file. Unfortunately, the contract file is not versioned, which means that we'll have to get creative about how we do this upgrade.

The first item in the persisted contract happens to be a slice, which is preceeded by an 8 byte number indicating the length. To do the upgrade, we'll insert an 8 byte magic string 'contract', which will activate bits in every byte. If we load the contract from disk and the first 8 bytes do not match the string 'contract', we know that it's the old version, and that we need to run the compatibility code. This is safe, because the restrictions of the encoding we currently use will prevent the on disk value for the slice length from being high enough to be the numerical encoding of 'contract'. There will be no false negatives when checking if we need to upgrade.

We'll need a test file in the `compatibility` package that has an old iteration of the contract, and verifies that the upgrade happens smoothly.

Overall, not too bad, though it's not as simple as originally thought.

DavidVorick:
The other option that we have, which I think we discussed, is that we can just slip the utility values at the end of the file, and we can avoid re-arranging the ordering of the fields.

This option is more brittle, but it's not adding technical debt, nor is this the type of technical debt that stacks up over time. We'd just be avoiding the problem until a future date. And that may be sufficient.

PR #2937, opened 2018-04-13
update /consensus/blocks endpoint docs

ChrisSchinnerl:


DavidVorick:
did we add an API test for this endpoint?

ChrisSchinnerl:
@DavidVorick Yes, we do. I extended the Client package + added a simple test in the last PR.

Issue #2416, opened 2017-10-11
Stuck contracts


uvizhe:
I have a lot of contracts stuck. I have switched hosting off by the end of April so the contracts had to expire about a month ago but I still see them as well as expected SC.

```
host -v

General Info:
	Connectability Status: Host appears to be working.

Host Internal Settings:
	acceptingcontracts:   No
	maxduration:          20 Weeks
	maxdownloadbatchsize: 17.83 MB
	maxrevisebatchsize:   17.83 MB
	netaddress:           178.130.44.248:9982 (manually specified)
	windowsize:           24 Hours

	collateral:       100 SC / TB / Month
	collateralbudget: 1 MS
	maxcollateral:    5 KS Per Contract

	mincontractprice:          20 SC
	mindownloadbandwidthprice: 250 SC / TB
	minstorageprice:           500 SC / TB / Month
	minuploadbandwidthprice:   10 SC / TB

Host Financials:
	Contract Count:               85
	Transaction Fee Compensation: 520 SC
	Potential Fee Compensation:   10.3 KS
	Transaction Fee Expenses:     0 H

	Storage Revenue:           54.95 SC
	Potential Storage Revenue: 94.5 SC

	Locked Collateral: 1.868 KS
	Risked Collateral: 18.9 SC
	Lost Collateral:   0 H

	Download Revenue:           232.2 mS
	Potential Download Revenue: 146 mS
	Upload Revenue:             1.298 SC
	Potential Upload Revenue:   669.6 mS

RPC Stats:
	Error Calls:        0
	Unrecognized Calls: 1758
	Download Calls:     0
	Renew Calls:        0
	Revise Calls:       0
	Settings Calls:     10703
	FormContract Calls: 0

Storage Folders:
    Used        Capacity      % Used    Path
    96.47 MB    149.787 GB    0.06      /Volumes/WHITE/sia_storage
```

DavidVorick:
Can you post the result of `siac renter contracts`?

DavidVorick:
Hmm. Seems like we don't have a way for a user to display all of the contracts that the host has.

Let's add that as a feature request, then we can debug this once we've added the ability to see all of the host's contracts.

tbenz9:
Linking #2465

tbenz9:
Linking #2938

PR #2947, opened 2018-04-19
Add siac command to get data from /host/contracts endpoint

nielscastien:
This PR introduces a new `siac` command to display contract information on the host.
The command uses 2 formats to display the data:

1. `value`: shows information about the financial values of the obligation
2. `status`: shows information about the status of the obligation

A filter is used to specify what data should should be shown based on the obligation status.

Examples: 
`siac host contracts value rejected` shows financial information about all rejected obligations
`siac host contracts status all` shows status information about all obligations

Data is always sorted by expiration height.

Closes issues #2416 and #2465

### DIFF
--- a/cmd/siac/consensuscmd.go
+++ b/cmd/siac/consensuscmd.go
@@ -6,7 +6,6 @@ import (
 
 	"github.com/spf13/cobra"
 
-	"github.com/NebulousLabs/Sia/node/api"
 	"github.com/NebulousLabs/Sia/types"
 )
 
@@ -22,8 +21,7 @@ var (
 // consensuscmd is the handler for the command `siac consensus`.
 // Prints the current state of consensus.
 func consensuscmd() {
-	var cg api.ConsensusGET
-	err := getAPI("/consensus", &cg)
+	cg, err := httpClient.ConsensusGet()
 	if err != nil {
 		die("Could not get current consensus state:", err)
 	}
--- a/cmd/siac/daemoncmd.go
+++ b/cmd/siac/daemoncmd.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 
 	"github.com/NebulousLabs/Sia/build"
-	"github.com/NebulousLabs/Sia/node/api"
 	"github.com/spf13/cobra"
 )
 
@@ -38,11 +37,6 @@ var (
 	}
 )
 
-type updateInfo struct {
-	Available bool   `json:"available"`
-	Version   string `json:"version"`
-}
-
 // version prints the version of siac and siad.
 func versioncmd() {
 	fmt.Println("Sia Client")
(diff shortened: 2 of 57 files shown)


lukechampine:
This looks pretty good. My first take is that `format` and `filter` probably work better as flags instead of arguments. Also, can you connect with some example output?

nielscastien:
image
image

nielscastien:
@lukechampine I can implement the flags. What about `-t` for `--type` (iso format) and `-f` for `--filter`?

lukechampine:
On second thought, do we really need `filter`? After all, you can get the same result by piping to `grep`. I feel like this is a repeat of the discussion we had regarding the API endpoint, haha. In general my preference is "delay" filtering as much as possible, but I could be persuaded otherwise. One counterargument, for example, is that this approach caters to Unix platforms at the expense of Windows (which doesn't have `grep`).

Also, we should trim the `obligation` prefix from the Obligation Status column. And `ID` (as in `ObligationID`) should always be uppercase, not `Id`.

Lastly, all those boolean columns seem excessive to me. IIUC, the confirmations always proceed linearly; that is, you should always see zero or more `true`s followed by zero or more `false`s. So we can condense this a lot by simply printing the "rightmost" confirmation, i.e.:
```
Origin Confirmed    Revision Constructed   Revision Confirmed
true                true                   false
```
becomes:
```
Revision Constructed
```
Extra-lastly, "origin" isn't a great name. We should try to think of a better one, or at least document it's actual meaning somewhere.

nielscastien:
I indeed think `--filter` is handy for all hosts that are not on unix and for people that want to see the descriptive headers. With the default value of `all` you don't have to use it if you prefer `grep`.

Trimming the status will indeed lead to more readable output so I will change it.

The `true` and `false` fields _seem_ to be excessive, but as @DavidVorick said

> As a dev, I'm most curious about: the obligation status group, the whole group.

E.g. this is real data from my system (with `grep` this time):
image
All obligations were seen on the blockchain except for one. I would like to understand why? This sort of information will be missed of we only give the latest status.

IIUC "origin" has something to do with "File contract is seen in a block but it can still be reverted". Right now I don't have an answer for that one.

mtlynch:
Thanks for this work @nielscastien!

>On second thought, do we really need filter? After all, you can get the same result by piping to grep.

+1 to skipping `filter`. I think punting to `grep` is fine. Even on Windows systems, there's `Select-String` which achieves a similar effect.

While Sia is still largely constrained by dev bandwidth, any reduction in LOC is a win.

nielscastien:
@mtlynch thanks for weighing in on this. I removed the filter flag and rebased the code.
@lukechampine I had to rebase the code since there were some conflicts with `master` that had to be resolved. It turned out to be quite a mess and difficult for someone that is not using git on a daily basis. A lot of old commits showed up in this PR now. I hope that's ok.
Anyhow, I hope this PR is ready to be merged. If there is anything else that needs to be changed, please let me know.

nielscastien:
@mtlynch thanks for this walkthrough. Worked like a charm. You can find the result in #2966

lukechampine:
Glad you were able to clean it up! For future reference, you can just `git push --force origin siac-contract-cmd` to overwrite the existing PR branch. That way we don't need a separate PR.
(`git push --force` is generally frowned upon if multiple people are working on the branch, but it's totally fine if there's only one committer. I use it frequently to clean up my PRs.)

PR #2949, opened 2018-04-19
Update siac to use node.api.Client package and add flag to change UserAgent

ChrisSchinnerl:


cmd/siac/main.go
```diff
@@ -309,8 +134,15 @@ func main() {
...
-	root.PersistentFlags().StringVarP(&addr, "addr", "a", "localhost:9980", "which host/port to communicate with (i.e. the host/port siad is listening on)")
+	// Check if the api password environment variable is set.
+	apiPassword := os.Getenv("SIA_API_PASSWORD")
+	if apiPassword != "" {
+		fmt.Println("Using SIA_API_PASSWORD environment variable")
+	}
+	// initialize client
+	root.PersistentFlags().StringVarP(&httpClient.Address, "addr", "a", "localhost:9980", "which host/port to communicate with (i.e. the host/port siad is listening on)")
+	root.PersistentFlags().StringVarP(&httpClient.Password, "apipassword", "x", apiPassword, "the password for the API's http authentication")
+	root.PersistentFlags().StringVarP(&httpClient.UserAgent, "useragent", "u", "Sia-Agent", "the useragent used by siac to connect to the daemon's API")
```
lukechampine: `x` isn't a great shorthand. Also, `u` is used for BasicAuth in `curl`, so that couuld be confusing.

Shorthands are optional; maybe we shouldn't have shorthands for these flags.

siatest/testnode.go
```diff
@@ -58,14 +58,14 @@ func NewCleanNode(nodeParams node.NodeParams) (*TestNode, error) {
 	tn := &TestNode{*s, *c, ""}
 
 	// Init wallet
-	wip, err := tn.WalletInitPost("", false)
+	wip, err := tn.WalletInitPost("", "english", false)
```
lukechampine: I'm not sure if it's worth supporting alternate dictionaries. AFAIK no one uses anything but english. We can preserve compatibility in the actual API server, but we don't necessarily need to encourage it anywhere else.

node/api/client/host.go
```diff
@@ -1,24 +1,84 @@
...
 	"strconv"
 
+	"github.com/NebulousLabs/Sia/crypto"
+	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/node/api"
 )
 
+// HostParam is a parameter in the host's settings that can be changed via the
+// API. It is primarily used as a helper struct to ensure type safety.
+type HostParam string
```
lukechampine: this makes me a bit uneasy. Probably because the HTTP API doesn't match the Go API: the HTTP API takes a key+value pair, whereas the Go API takes a full `HostInternalSettings` struct.

I prefer the latter form because you can make multiple changes atomically. But the former form more closely matches how people will actually change their settings: one at a time.

I think my preference would be to update the `/host` route to accept a full `HostInternalSettings` object. But for now, a big list of `consts` is probably the best we can do. (btw, they can be `const`, not `var`.)

lukechampine: well, we could make the change in a backwards-compatible way, and have the `client` package only support the new way. Something to think about for later.

Issue #2952, opened 2018-04-21
Lots of failed storage obligations (lost revenues + Tx fees)


nielscastien:
There are a lot of failed storage obligations on my host. Looking at the data I can see that there are no `SectorRoots` on the failed obligations, `OriginConfirmed` and `RevisionConfirmed` are both `true` and all potential revenues are always 0.

The following code will fail the obligation when there are no `SectorRoots` leading to lost revenues for the host (`ContractCost` and spend `TransactionFeesAdded` for submitting the file contract revision):
https://github.com/NebulousLabs/Sia/blob/6f6ef34bc345963c8ca4142db32a3c46a8652a66/modules/host/storageobligations.go#L761-L777

If the renter never uploads any data, it is not correct to fail the storage proof. Since these (empty) contracts lock up a lot of collateral, it's fair to be paid at least the contract cost to cover the transaction fees.

Issue is also reported in #2138

ChrisSchinnerl:
@nielscastien take a look at `FormContract` in `formcontract.go`. The initial revision of the filecontract returns all the collateral to the host even if the storage proof was not provided because no data has been uploaded yet. The host simply can't provide a storage proof without any data to create it.

nielscastien:
@ChrisSchinnerl I am not saying the collateral not returned, it is. And because we are talking about empty contracts, there is no risked collateral. So far so good. I am saying that the contract cost is not awarded to the host to cover transaction fees. To be more specific, over 25% of my contracts are empty contracts where I have to lock a lot of collateral for long periods of time (25 weeks). This prevents me from accepting other (more valuable?) contracts because locked collateral has reached its upper limit. And in the end the host has to submit a revision and pays for the transaction fees. This just doesn't feel right...

If it is impossible to supply a storage proof with empty contracts or to return the contract cost to the host, we should prevent the host from submitting revisions to the block chain for empty contracts.

lukechampine:
@nielscastien an empty contract pays out both the collateral and the contract cost to the host:

https://github.com/NebulousLabs/Sia/blob/6f6ef34bc345963c8ca4142db32a3c46a8652a66/modules/renter/proto/formcontract.go#L57-L58

https://github.com/NebulousLabs/Sia/blob/6f6ef34bc345963c8ca4142db32a3c46a8652a66/modules/renter/proto/formcontract.go#L74-L87

It sounds like we should confirm this with a test, though.

[...]

PR #2959, opened 2018-04-24
Fix logging empty contracts as failed

nielscastien:
Fix for issue #2952

### DIFF
--- a/modules/host/storageobligations.go
+++ b/modules/host/storageobligations.go
@@ -578,7 +578,7 @@ func (h *Host) removeStorageObligation(so storageObligation, sos storageObligati
 		// storage obligation should equal the contract cost of the obligation
 		revenue := so.ContractCost.Add(so.PotentialStorageRevenue).Add(so.PotentialDownloadRevenue).Add(so.PotentialUploadRevenue)
 		if len(so.SectorRoots) == 0 {
-			h.log.Printf("Not submitted a storage proof for empty contract. Revenue is %v.\n", revenue)
+			h.log.Printf("No need to submit a storage proof for empty contract. Revenue is %v.\n", revenue)
 		} else {
 			h.log.Printf("Successfully submitted a storage proof. Revenue is %v.\n", revenue)
 		}
--- a/modules/host/storageobligations.go
+++ b/modules/host/storageobligations.go
@@ -771,18 +771,6 @@ func (h *Host) threadedHandleActionItem(soid types.FileContractID) {
 	if !so.ProofConfirmed && blockHeight >= so.expiration()+resubmissionTimeout {
 		h.log.Debugln("Host is attempting a storage proof for", so.id())
 
-		// If the window has closed, the host has failed and the obligation can
-		// be removed.
-		if so.proofDeadline() < blockHeight {
-			h.log.Debugln("storage proof not confirmed by deadline, id", so.id())
-			h.mu.Lock()
-			err := h.removeStorageObligation(so, obligationFailed)
-			h.mu.Unlock()
-			if err != nil {
-				h.log.Println("Error removing storage obligation:", err)
-			}
-			return
-		}
 		// If the obligation has no sector roots, we can remove the obligation and not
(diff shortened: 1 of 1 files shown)


modules/host/storageobligations.go
```diff
@@ -574,8 +574,15 @@ func (h *Host) removeStorageObligation(so storageObligation, sos storageObligati
 		}
 	}
 	if sos == obligationSucceeded {
+		// Empty obligations don't submit a storage proof. The revenue for an empty
+		// storage obligation should equal the contract cost of the obligation
+		if len(so.SectorRoots) == 0 {
+			h.log.Printf("Not submitted a storage proof for empty contract. Revenue is %v.\n", so.ContractCost.Add(so.PotentialStorageRevenue).Add(so.PotentialDownloadRevenue).Add(so.PotentialUploadRevenue))
```
ChrisSchinnerl: I'd probably print `No need to submit storage proof for empty contract` to make it clear that this is expected behavior.

modules/host/storageobligations.go
```diff
@@ -574,8 +574,15 @@ func (h *Host) removeStorageObligation(so storageObligation, sos storageObligati
...
+		} else {
+			h.log.Printf("Successfully submitted a storage proof. Revenue is %v.\n", so.ContractCost.Add(so.PotentialStorageRevenue).Add(so.PotentialDownloadRevenue).Add(so.PotentialUploadRevenue))
+		}
```
ChrisSchinnerl: Since the revenue is calculated the same way for both `Printf` statements we can simplify that to make it a bit more readable.
```
revenue := so.ContractCost.Add(so.PotentialStorageRevenue).Add(so.PotentialDownloadRevenue).Add(so.PotentialUploadRevenue)
if len(so.SectorRoots) == 0 { 
    h.log.Printf("No need to submit a storage proof for an empty contract. Revenue is %v.\n", revenue)
} else {
    h.log.Printf("Successfully submitted a storage proof. Revenue is %v.\n", revenue)
}
```

modules/host/storageobligations.go
```diff
@@ -775,6 +782,19 @@ func (h *Host) threadedHandleActionItem(soid types.FileContractID) {
...
+		if len(so.SectorRoots) == 0 {
+			h.log.Debugln("storage proof not submitted for empty contract, id", so.id())
+			h.mu.Lock()
+			err := h.removeStorageObligation(so, obligationSucceeded)
+			h.mu.Unlock()
+			if err != nil {
+				h.log.Println("Error removing storage obligation:", err)
+			}
+			return
+		}
```
ChrisSchinnerl: I think there is a chance that `len(so.SectorRots) == 0` and `so.proofDeadline() < blockHeight` are both `true`. In that case we probably also want to indicate a success for the storage obligation. That means the two `if` blocks should probably be reversed in order.

nielscastien:
@ChrisSchinnerl done!

modules/host/storageobligations.go
lukechampine: can clean this up a bit with a variable:
```go
revenue := so.ContractCost.Add(so.PotentialStorageRevenue).Add(so.PotentialDownloadRevenue).Add(so.PotentialUploadRevenue)
```

modules/host/storageobligations.go
```diff
@@ -775,6 +782,19 @@ func (h *Host) threadedHandleActionItem(soid types.FileContractID) {
 			}
 			return
 		}
+		// If the obligation has no sector roots, we can remove the obligation and not
+		// submit a storage proof. The host payout for a failed empty contract
+		// includes the contract cost and locked collateral.
+		if len(so.SectorRoots) == 0 {
```
lukechampine: this is clearer as an `else if` attached to the previous `if`, to make it clear that they're mutually exclusive

modules/host/storageobligations.go
lukechampine: ha, Chris got to it first

PR #2960, opened 2018-04-24
Improve tpool panic output

ChrisSchinnerl:
Since it seems like a few people are still experiencing this `panic` I think it wouldn't hurt the add some info to the output.

### DIFF
--- a/modules/transactionpool/update.go
+++ b/modules/transactionpool/update.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"sort"
 
-	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/Sia/modules"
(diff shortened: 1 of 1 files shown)


modules/transactionpool/update.go
```diff
@@ -185,15 +187,16 @@ func (tp *TransactionPool) ProcessConsensusChange(cc modules.ConsensusChange) {
 		tp.log.Println("NOTE: Upgrading tpool database to support consensus change verification.")
 		resetSanityCheck = true
 	} else if err != nil {
-		tp.log.Println("ERROR: Could not access recentID from tpool.")
+		tp.log.Println("ERROR: Could not access recentID from tpool:", err)
+		build.Critical("ERROR: Could not access recentID from tpool:", err)
```
lukechampine: does `tp.log.Critical` do the same thing?

ChrisSchinnerl: I bet it does :P

PR #2967, opened 2018-04-26
Speedup initial hostscan

ChrisSchinnerl:


modules/renter/hostdb/scan.go
```diff
@@ -275,19 +296,31 @@ func (hdb *HostDB) managedScanHost(entry modules.HostDBEntry) {
...
 	hdb.updateEntry(entry, err)
-	hdb.mu.Unlock()
+
+	// Update the initial scan timeout if the scan was successful and if the
+	// latency for this scan was greater than the current timeout. We want to
+	// find the worst case latency.
+	if success {
+		hdb.successfulScans++
+		if latency > hdb.initialScanTimeout {
+			hdb.initialScanTimeout = 2 * latency
```
DavidVorick: May want to add a comment explaining why the `if` compares them directly, but the update uses `latency * 2`

DavidVorick:
Have you run this on the live network, and if so were you able to find what the longest successful Dial was? I'm curious how many hosts are close to the total timeout. Maybe we should do a test run where we `log.Debugln`  the timeouts for all the hosts.

I'm tempted to say that we should extend this PR to support only caring about the worst 95%, instead of caring about the worst 100%. That makes the math more complicated though.

modules/renter/hostdb/scan.go
```diff
@@ -275,19 +302,32 @@ func (hdb *HostDB) managedScanHost(entry modules.HostDBEntry) {
 		copy(pubkey[:], pubKey.Key)
 		return crypto.ReadSignedObject(conn, &settings, maxSettingsLen, pubkey)
 	}()
+	success := false
```
DavidVorick: `success = err == nil`

PR #2974, opened 2018-04-30
Improve BucketProcessedTransactions usage

ChrisSchinnerl:
This PR fixes 3 potential issues in the wallet related to the processed transactions bucket.

1.) After deleting from the bucket we decrement the sequence number. This avoids gaps being created during a reorg which might lead to the binary search breaking.

2.) Whenever we remove a txn from `BucketProcessedTransactions` we also remove it from `BucketProcessedTxnIndex`. This way we can'tt run into issues with txnIDs pointing to nonexistent indices or even indices of the wrong transaction.

3.) Due to the changes in 1) we can now use the sequence number instead of `cursor.Last()` to find the last processed transaction. Apparently `cursor.Last()` is unrealiable when used before calling `txn.Commit()` on a bbolt transaction and might return `nil` instead of the last entry or even a wrong entry. This causes either no entry to be deleted or the wrong one during a reorg without anyone noticing.

### DIFF
--- a/siatest/wallet/wallet_test.go
+++ b/siatest/wallet/wallet_test.go
@@ -1,0 +1,118 @@
+package wallet
+
+import (
+	"errors"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/NebulousLabs/Sia/build"
+	"github.com/NebulousLabs/Sia/siatest"
+	"github.com/NebulousLabs/Sia/types"
+)
+
+func TestTransactionReorg(t *testing.T) {
+	if testing.Short() {
+		t.SkipNow()
+	}
+
+	// Create 2 miners
+	testdir, err := siatest.TestDir(t.Name())
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Create two miners
+	miner1, err := siatest.NewNode(siatest.Miner(filepath.Join(testdir, "miner1")))
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer func() {
+		if err := miner1.Close(); err != nil {
+			t.Fatal(err)
+		}
+	}()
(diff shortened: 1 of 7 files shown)


siatest/wallet/wallet_test.go
```diff
@@ -0,0 +1,118 @@
...
+			if t.TransactionID == txn {
+				return nil
+			}
+		}
+		return errors.New("txn isn't processed yet")
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	println("\ncreate node 2")
```
DavidVorick: leftover printlns in this file.

siatest/wallet/wallet_test.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/NebulousLabs/Sia/siatest"
+	"github.com/NebulousLabs/Sia/types"
+)
+
+func TestTransactionReorg(t *testing.T) {
+	if testing.Short() {
+		t.SkipNow()
+	}
+
+	// Create 2 miners
```
DavidVorick: comment in wrong place

siatest/wallet/wallet_test.go
```diff
@@ -0,0 +1,118 @@
...
+		for _, t := range wtg.ConfirmedTransactions {
+			if t.TransactionID == txn {
+				return errors.New("txn is still processed")
+			}
+		}
+		return nil
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
```
DavidVorick: Can you extend the test so that the transaction that miner1 made gets added to the chain when miner 2 sets up the reorg? That will require fetching the raw transaction I think from miner 1 and adding it to the tpool for miner 2.

You should be able to use the get and post raw txn endpoints.

PR #2975, opened 2018-05-01
Implement fmt.Scanner for UnlockHash

ChrisSchinnerl:
This PR implements the `fmt.Scanner` interface for the `UnlockHash` struct.

### DIFF
--- a/types/encoding_test.go
+++ b/types/encoding_test.go
@@ -723,13 +723,11 @@ func TestTransactionMarshalSiaSize(t *testing.T) {
 // works as expected.
 func TestUnlockHashScan(t *testing.T) {
 	// Create a random unlock hash.
-	uh := UnlockHash{}
-	copy(uh[:], fastrand.Bytes(crypto.HashSize))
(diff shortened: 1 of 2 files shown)


types/encoding_test.go
```diff
@@ -718,3 +718,20 @@ func TestTransactionMarshalSiaSize(t *testing.T) {
 		t.Errorf("sizes do not match: expected %v, got %v", len(encoding.Marshal(txn)), txn.MarshalSiaSize())
 	}
 }
+
+// TestUnlockHashScan checks if the fmt.Scanner implementation of UnlockHash
+// works as expected.
+func TestUnlockHashScan(t *testing.T) {
+	// Create a random unlock hash.
+	uh := UnlockHash{}
+	copy(uh[:], fastrand.Bytes(crypto.HashSize))
```
lukechampine: more idiomatic would be:
```go
var uh UnlockHash
fastrand.Read(uh[:])
```
:)

PR #2977, opened 2018-05-01
Download cache sort

MSevey:
previously the addChunkToCache function just deleted the first key it pulled from the chunkCache map.  I added a new struct cacheData to the chuckCache map so that a timestamp can be stored with the data byte slice.  Now the addChunkToCache function checks the timestamps and deletes the oldest chunk.

### DIFF
--- a/modules/renter/downloadcache.go
+++ b/modules/renter/downloadcache.go
@@ -39,7 +39,7 @@ func (udc *unfinishedDownloadChunk) addChunkToCache(data []byte) {
 		delete(udc.chunkCache, oldestKey)
 	}
 
-	udc.chunkCache[udc.staticCacheID] = cacheData{
+	udc.chunkCache[udc.staticCacheID] = &cacheData{
 		data:       data,
 		lastAccess: time.Now(),
 	}
@@ -56,12 +56,15 @@ func (r *Renter) managedTryCache(udc *unfinishedDownloadChunk) bool {
 	defer udc.mu.Unlock()
 	r.cmu.Lock()
 	cd, cached := r.chunkCache[udc.staticCacheID]
-	cd.lastAccess = time.Now()
-	r.chunkCache[udc.staticCacheID] = cd
 	r.cmu.Unlock()
 	if !cached {
 		return false
 	}
+
+	// chunk exists, updating lastAccess and reinserting into map
+	cd.lastAccess = time.Now()
+	r.chunkCache[udc.staticCacheID] = cd
+
 	start := udc.staticFetchOffset
 	end := start + udc.staticFetchLength
 	_, err := udc.destination.WriteAt(cd.data[start:end], udc.staticWriteOffset)
--- a/modules/renter/downloadcache.go
+++ b/modules/renter/downloadcache.go
@@ -9,8 +9,6 @@ import (
 	"github.com/NebulousLabs/errors"
 )
 
-var cd cacheData
-
 // addChunkToCache adds the chunk to the cache if the download is a streaming
 // endpoint download.
 // TODO this won't be necessary anymore once we have partial downloads.
@@ -38,8 +36,10 @@ func (udc *unfinishedDownloadChunk) addChunkToCache(data []byte) {
 		delete(udc.chunkCache, oldestKey)
 	}
 
-	cd.data = data
-	cd.timestamp = time.Now()
+	cd := cacheData{
+		data:      data,
+		timestamp: time.Now(),
+	}
 	udc.chunkCache[udc.staticCacheID] = cd
 }
 
--- a/modules/renter/downloadcache.go
+++ b/modules/renter/downloadcache.go
@@ -23,24 +23,25 @@ func (udc *unfinishedDownloadChunk) addChunkToCache(data []byte) {
 	defer udc.cacheMu.Unlock()
 
 	// Prune cache if necessary.
-	if len(udc.chunkCache) >= downloadCacheSize {
+	for len(udc.chunkCache) >= downloadCacheSize {
 		var oldestKey string
-		oldestTime := time.Now().Second()
+		oldestTime := time.Now()
 
-		for key := range udc.chunkCache {
-			if udc.chunkCache[key].timestamp.Second() < oldestTime {
-				oldestTime = udc.chunkCache[key].timestamp.Second()
-				oldestKey = key
+		// TODO: turn this from a structure where you loop over every element (O(n) per access) to a min heap (O(log n) per access).
+		// currently not a issue due to cache size remaining small (<20)
+		for id, chunk := range udc.chunkCache {
+			if chunk.lastAccess.Before(oldestTime) {
+				oldestTime = chunk.lastAccess
+				oldestKey = id
 			}
 		}
 		delete(udc.chunkCache, oldestKey)
 	}
 
-	cd := cacheData{
-		data:      data,
-		timestamp: time.Now(),
+	udc.chunkCache[udc.staticCacheID] = cacheData{
+		data:       data,
+		lastAccess: time.Now(),
 	}
-	udc.chunkCache[udc.staticCacheID] = cd
 }
 
 // managedTryCache tries to retrieve the chunk from the renter's cache. If
@@ -54,14 +55,14 @@ func (r *Renter) managedTryCache(udc *unfinishedDownloadChunk) bool {
 	defer udc.mu.Unlock()
 	r.cmu.Lock()
 	cd, cached := r.chunkCache[udc.staticCacheID]
-	data := cd.data
+	cd.lastAccess = time.Now()
 	r.cmu.Unlock()
 	if !cached {
 		return false
 	}
 	start := udc.staticFetchOffset
 	end := start + udc.staticFetchLength
-	_, err := udc.destination.WriteAt(data[start:end], udc.staticWriteOffset)
+	_, err := udc.destination.WriteAt(cd.data[start:end], udc.staticWriteOffset)
 	if err != nil {
 		r.log.Println("WARN: failed to write cached chunk to destination:", err)
 		udc.fail(errors.AddContext(err, "failed to write cached chunk to destination"))
--- a/modules/renter/downloadcache.go
+++ b/modules/renter/downloadcache.go
@@ -9,12 +9,16 @@ import (
 	"github.com/NebulousLabs/errors"
 )
 
+var cd cacheData
+
 // addChunkToCache adds the chunk to the cache if the download is a streaming
 // endpoint download.
 // TODO this won't be necessary anymore once we have partial downloads.
 func (udc *unfinishedDownloadChunk) addChunkToCache(data []byte) {
 	if udc.download.staticDestinationType != destinationTypeSeekStream {
-		// We only cache streaming chunks since browsers and media players tend to only request a few kib at once when streaming data. That way we can prevent scheduling the same chunk for download over and over.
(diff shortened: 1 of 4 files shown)


modules/renter/downloadcache.go
```diff
@@ -14,21 +14,33 @@ import (
...
+		delete(udc.chunkCache, oldestKey)
+	}
+
+	cd := cacheData{
+		data:      data,
+		timestamp: time.Now(),
 	}
-	udc.chunkCache[udc.staticCacheID] = data
-	udc.cacheMu.Unlock()
+	udc.chunkCache[udc.staticCacheID] = cd
```
ChrisSchinnerl: you could save a line here if you didn't save the object in a variable like this:
```
udc.chunkCache[udc.staticCacheID] = cacheData {
    data: data,
    timestamp: time.Now(),
}
```

modules/renter/downloadcache.go
```diff
@@ -41,7 +53,8 @@ func (r *Renter) managedTryCache(udc *unfinishedDownloadChunk) bool {
 	udc.mu.Lock()
 	defer udc.mu.Unlock()
 	r.cmu.Lock()
-	data, cached := r.chunkCache[udc.staticCacheID]
+	cd, cached := r.chunkCache[udc.staticCacheID]
+	data := cd.data
```
ChrisSchinnerl: Instead of assigning `cd.data` to `data` you could simply replace the following occurences of `data` with `cd.data`.

modules/renter/downloadchunk.go
```diff
@@ -22,6 +22,13 @@ type downloadPieceInfo struct {
 	root  crypto.Hash
 }
 
+// cacheData contatins the data and the timestamp for the unfinished
+// download chunks
+type cacheData struct {
```
ChrisSchinnerl: We usually try to sort declarations of structs, functions and fields. In this case `cacheData` should be declared above `downloadPieceInfo`

modules/renter/downloadchunk.go
```diff
@@ -75,7 +82,7 @@ type unfinishedDownloadChunk struct {
 	mu       sync.Mutex
 
 	// Caching related fields
-	chunkCache map[string][]byte
+	chunkCache map[string]cacheData
```
ChrisSchinnerl: my personal preference here would be to make this a `map[string]*cacheData`.

modules/renter/renter.go
```diff
@@ -363,10 +363,10 @@ func validateSiapath(siapath string) error {
 		return ErrEmptyFilename
 	}
 	if siapath == ".." {
-		return errors.New("siapath cannot be ..")
+		return errors.New("siapath cannot be .. ")
```
ChrisSchinnerl: Is there a reason for adding a space here?

modules/renter/renter.go
```diff
@@ -363,10 +363,10 @@ func validateSiapath(siapath string) error {
...
 	}
 	if siapath == "." {
-		return errors.New("siapath cannot be .")
+		return errors.New("siapath cannot be . ")
```
ChrisSchinnerl: and here?

modules/renter/downloadcache.go
```diff
@@ -14,21 +14,33 @@ import (
 // TODO this won't be necessary anymore once we have partial downloads.
 func (udc *unfinishedDownloadChunk) addChunkToCache(data []byte) {
 	if udc.download.staticDestinationType == destinationTypeSeekStream {
-		// We only cache streaming chunks since browsers and media players tend to only request a few kib at once when streaming data. That way we can prevent scheduling the same chunk for download over and over.
+		// We only cache streaming chunks since browsers and media players tend
```
tbenz9: Best to keep comments in the infinitive form.  The one exception is "TODO" comments which are typically less formal and could be in first person or the infinitive form.

modules/renter/downloadcache.go
```diff
@@ -14,21 +14,33 @@ import (
...
 	udc.cacheMu.Lock()
+	defer udc.cacheMu.Unlock()
+
 	// Prune cache if necessary.
-	// TODO insteado of deleting a 'random' key, delete the
-	// least-recently-accessed element of the cache.
-	for key := range udc.chunkCache {
-		if len(udc.chunkCache) < downloadCacheSize {
-			break
+	if len(udc.chunkCache) >= downloadCacheSize {
```
DavidVorick: should be a loop `for len(udc.chunkCache) >= downloadCacheSize {`

modules/renter/downloadcache.go
```diff
@@ -14,21 +14,33 @@ import (
...
+		var oldestKey string
+		oldestTime := time.Now().Second()
+
+		for key := range udc.chunkCache {
```
DavidVorick: should add a TODO here to turn this from a structure where you loop over every element (O(n) per access) to a min heap (O(log n) per access).

For now, we won't be supporting cache sizes large enough for it to matter. If `n` is less than 20 it's not an issue.

modules/renter/renter.go
MSevey: it gets ride of a warning. there was warning that errors should not end in .. or .

modules/renter/renter.go
MSevey: same as above

modules/renter/renter.go
lukechampine: let's enclose it in `'..'` instead

modules/renter/downloadcache.go
ChrisSchinnerl: I don't think it needs to be since we never add more than 1 element at once and we also check the cacheSize every time we add a single element.

MSevey:
Updated to include edits from @ChrisSchinnerl @DavidVorick and @lukechampine 

Will begin working on the test now.

modules/renter/downloadcache.go
```diff
@@ -14,21 +14,34 @@ import (
...
-	// TODO insteado of deleting a 'random' key, delete the
-	// least-recently-accessed element of the cache.
-	for key := range udc.chunkCache {
-		if len(udc.chunkCache) < downloadCacheSize {
-			break
+	for len(udc.chunkCache) >= downloadCacheSize {
+		var oldestKey string
+		oldestTime := time.Now()
+
+		// TODO: turn this from a structure where you loop over every element (O(n) per access) to a min heap (O(log n) per access).
```
DavidVorick: need to wrap this line

modules/renter/downloadchunk.go
DavidVorick: oh hmm, actually we do modify the timestamp in place. We modify the timestamp in place when a chunk is accessed in TryCache.

Maybe we should make it a pointer then.

MSevey:
Created unit test for addChunkToCache to verify oldest chunk is removed.

modules/renter/downloadcache_test.go
```diff
@@ -0,0 +1,39 @@
...
+	"sync"
+	"testing"
+	"time"
+)
+
+// TestAddChunkToCache tests that the oldest chunk is removed
+func TestAddChunkToCache(t *testing.T) {
+	udc := &unfinishedDownloadChunk{
+		download: &download{
+			staticDestinationType: "httpseekstream",
```
ChrisSchinnerl: This should use the `destinationTypeSeekStream` constant instead of a hardcoded value.

modules/renter/downloadcache_test.go
```diff
@@ -0,0 +1,39 @@
...
+		chunkCache: make(map[string]*cacheData),
+		cacheMu:    new(sync.Mutex),
+	}
+
+	data := []byte{1, 2, 3, 4}
+
+	// Fill Cache
+	for i := 0; i < downloadCacheSize; i++ {
+		udc.staticCacheID = strconv.Itoa(i)
+		udc.addChunkToCache(data)
```
ChrisSchinnerl: You don't really need to declare `data` since it also works with an empty data slice.
`udc.addChunkToCache([]byte{})` should work just fine.

modules/renter/downloadcache_test.go
```diff
@@ -0,0 +1,39 @@
...
+	}
+
+	// Add additional chunk to force deletion of a chunk
+	udc.staticCacheID = strconv.Itoa(downloadCacheSize)
+	udc.addChunkToCache(data)
+
+	// check if the chunk with staticCacheID = "0" was removed
+	// as that would have been the first to be added
+	if _, ok := udc.chunkCache["0"]; ok {
+		t.Errorf("Expected ok to be false instead it was %v", ok)
```
ChrisSchinnerl: The error message can probably be a simple `t.Error("The least recently accessed chunk wasn't pruned from the cache")` since that would give a better description of the problem if you see the error message without any context.
You also don't need a formatting string for a boolean since it can only have 2 different values anyway :P

modules/renter/downloadcache_test.go
```diff
@@ -0,0 +1,39 @@
...
+	}
+
```
ChrisSchinnerl: newline at the end of function

modules/renter/downloadcache.go
```diff
@@ -14,21 +14,35 @@ import (
...
+		// currently not a issue due to cache size remaining small (<20)
+		for id, chunk := range udc.chunkCache {
+			if chunk.lastAccess.Before(oldestTime) {
+				oldestTime = chunk.lastAccess
+				oldestKey = id
+			}
 		}
-		delete(udc.chunkCache, key)
+		delete(udc.chunkCache, oldestKey)
+	}
```
lukechampine: instead of a min heap, we could use a bit of extra memory to sort and then delete. The runtime is O(n log n), which is slightly worse than the minheap's O(m log n), but not terribly so. Something like:
```go
if len(udc.chunkCache) > downloadCacheSize {
	cacheIDs := make([]string, 0, len(udc.chunkCache))
	for id := range udc.chunkCache {
		cacheIDs = append(cacheIDs, id)
	}
	sort.Slice(cacheIDs, func(i, j int) bool {
		ci := udc.chunkCache[cacheIDs[i]]
		cj := udc.chunkCache[cacheIDs[j]]
		return ci.lastAccess.Before(cj.lastAccess)
	})
	for len(udc.chunkCache) > downloadCacheSize {
		delete(udc.chunkCache, cacheIDs[0])
		cacheIDs = cacheIDs[1:]
	}
}
```

although, come to think of it, don't we do this pruning every time we insert a chunk? That is, won't the `for len(udc.chunkCache) >= downloadCacheSize` loop only ever need to run once? If so, a min heap makes sense (because it will persist across calls), but sorting doesn't.

ChrisSchinnerl: The for loop doesn't really hurt but yeah you are right. It should only ever be executed once which makes the heap O(log n) for pruning vs O(n log n).

modules/renter/downloadcache_test.go
```diff
@@ -0,0 +1,36 @@
...
+	// Add additional chunk to force deletion of a chunk
+	udc.staticCacheID = strconv.Itoa(downloadCacheSize)
+	udc.addChunkToCache([]byte{})
+
+	// check if the chunk with staticCacheID = "0" was removed
+	// as that would have been the first to be added
+	if _, ok := udc.chunkCache["0"]; ok {
+		t.Error("The least recently accessed chunk wasn't pruned from the cache")
+	}
+}
```
DavidVorick: This test only checks that the first uploaded chunk gets removed, needs to be extended to check the behavior between the least recently accessed vs. the least recently added to the cache.

PR #2980, opened 2018-05-02
Add "height" and "id" fields to /consensus/blocks endpoint response + update docu

ChrisSchinnerl:


doc/API.md
```diff
@@ -196,6 +196,47 @@ height
...
+            ],
+            "filecontractrevisions": [],
+            "filecontracts": [],
+            "minerfees": [],
+            "siacoininputs": [],
+            "siacoinoutputs": [],
+            "siafundinputs": [],
+            "siafundoutputs": [],
+            "storageproofs": [],
+            "transactionsignatures": []
```
lukechampine: I think these will actually be `null`, not `[]`. Also, if we're going to provide a real example, might be best to use one with inputs+outputs instead of just arbitrary data.

doc/API.md
```diff
@@ -196,6 +196,47 @@ height
...
+    "nonce": [
+        253,
+        253,
+        0,
+        0,
+        0,
+        0,
+        0,
+        100
+    ],
```
lukechampine: this can be collapsed onto one line

doc/API.md
ChrisSchinnerl: This is actually a real example using `height=1` formatted using the python json tool. 
Not sure if having a bigger block is a good idea since even a small block can easily have 300 lines of json.

node/api/consensus.go
```diff
@@ -77,7 +84,11 @@ func (api *API) consensusBlocksHandler(w http.ResponseWriter, req *http.Request,
 		return
 	}
 	// Write response
-	WriteJSON(w, b)
+	WriteJSON(w, ConsensusBlocksGet{
+		b.ID(),
+		h,
+		b,
```
DavidVorick: should probably have names for these fields.

node/api/consensus.go
```diff
@@ -25,6 +25,13 @@ type ConsensusHeadersGET struct {
 	BlockID types.BlockID `json:"blockid"`
 }
 
+// ConsensusBlocksGet wraps a types.Block and adds an id field.
+type ConsensusBlocksGet struct {
+	BlockID     types.BlockID     `json:"id"`
+	BlockHeight types.BlockHeight `json:"height"`
```
DavidVorick: is the difference between the field name in the struct and the field name in the json intentional?

ChrisSchinnerl: Yes, because the embedded `Block` already has a `ID` method. That's why I changed it to `BlockID`. `BlockID` and `Height` looks weird so I changed the `Height` to `BlockHeight`. 
For the JSON response `id` and `height` make sense though.

hupengbuaa: Have you found a way to get `BlockHeight` from `ID`?   Besides, how long would it take for you to finish this? I need finish my job before next Monday. So I'm working in a new branch to get those data support.  I couldn't figure out except using the `current_height - 1`, but I  still need  find the `current_height` in my code.  If there is a better solution, plz let me know.

ChrisSchinnerl: @hupengbuaa this PR should be pretty much done. I'm just waiting for @DavidVorick or @lukechampine to approve and merge it.

hupengbuaa: Oh, thx. I see.

PR #2981, opened 2018-05-02
add id to /consensus/blocks endpoint

hupengbuaa:
Simply add  block_id info for `/consensus/blocks` endpoint


ChrisSchinnerl:
Thanks for the PR but we found another way to do it and already have an open [PR](https://github.com/NebulousLabs/Sia/pull/2980) for that.

hupengbuaa:
Great. I've subscribed the thread.

Issue #2978, opened 2018-05-02
Collateral ratio penalty could not be working properly, depleting renter's allowance


hakkane84:
**Collateral ratio penalty could not be working properly, depleting renter's allowance (potential Bug)**

# BUG REPORT

**Stack Trace or error message**
No error. This report comes from blockchain data analysis

**Expected Behavior**
The scoring system of hosts promote to increase their collateral, but after certain ratio of collateral vs storage cost the score should decrease to prevent renters from paying excessive Siafund fees (as they pay a 3.9% from the allowance + collateral). Long ago, I remember Taek mentioned the threshold to start decreasing the score should be around 3x-4x in this ratio collateral/storageprice. In another post, I remember Taek mentioned a renter usually ends up paying a 10% of their allowance in fees

**How to reproduce it (as minimally and precisely as possible)**
I am building a new Explorer for Sia. Among its features, I am curating and representing the metadata of file contracts included on the blockchain. One of the parameters I show is how much in SF fees a contract is paying and the percentage they represent, compared with the allowance paid by the renter. Take a look at the following examples from my Explorer prototype:

1 - https://siastats.info/navigator?search=d7683975103a0073b63aeb89f3592bb4c09ea3c4a8d6c856b469a6c4f1601d05 --> 51%

2- https://siastats.info/navigator?search=8e4e806122c3fff863fd760d12656b30fc205d76ebdfcc2edc31d345d2b1b6df --> 65%

3 - https://siastats.info/navigator?search=0b2afaf0675ac7a003f4c906ca1058626b0cf85acbe85ddb6638c1f8bcea0dad --> 40%

4- https://siastats.info/navigator?search=09b88f82881a98ee667b01bf35886b423c69f6d36fcb8d202d25702c9dec614d --> 28%

In the left side of the transaction schematic you can see the inputs, labelled as the renter allowance and the host's collateral. In the table bellow you can find the amount paid as SF fees and the % they represent of the renter's allowance.

While these cases are not the majority of contracts, there are multiple contracts where the renter is expending up to, and more than, 50% of their allowance in paying SF fees. This makes me believe that the part of the host scoring system that penalizes hosts with unproportioned collateral is not working as expected.

[...]

PR #2982, opened 2018-05-02
Reduce impact of host collateral on scoring the host

ChrisSchinnerl:
This PR changes the influence of the host collateral on the score from a linear function to `f(x)=x^(3/4)`


tbenz9:
Can you help me understand this?  It looks like the amount of collateral will now have less effect on a host score.  Is that correct?

ChrisSchinnerl:
@tbenz9 yes, that's correct. So instead of getting a 2x collateral bonus when you double your collateral you only get a ~1.68x bonus. A host that previously put up 10x the collateral of another host would now have to put up ~22x the collateral to get the same advantage as before. 
The reason for this was this [issue](https://github.com/NebulousLabs/Sia/issues/2978).

PR #2985, opened 2018-05-03
Fix archived contracts being revived on Windows after restart

ChrisSchinnerl:
This PR implements a `TODO` in the contract archival code, adds some logging in case of a failed `SafeContract` deletion and hopefully fixes an issue that caused the contracts of Windows users to be "revived" after being archived.

### DIFF
--- a/modules/renter/contractor/update.go
+++ b/modules/renter/contractor/update.go
@@ -7,17 +7,7 @@ import (
 
 // managedArchiveContracts will figure out which contracts are no longer needed
 // and move them to the historic set of contracts.
-//
-// TODO: This function should be performed by threadedContractMaintenance.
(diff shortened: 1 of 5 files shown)


modules/renter/contractor/update.go
```diff
@@ -90,6 +80,5 @@ func (c *Contractor) ProcessConsensusChange(cc modules.ConsensusChange) {
 	// maintenance.
 	if cc.Synced {
 		go c.threadedContractMaintenance()
-		go c.managedArchiveContracts()
```
DavidVorick: now that I see this... I wonder why we ever made it this way??? Should definitely have been doing the saving at the same point that we do the modifications.

PR #2992, opened 2018-05-06
Fix violation of locking conventions in FileList call

ChrisSchinnerl:


modules/renter/contractor/contractor.go
```diff
@@ -135,14 +135,12 @@ func (c *Contractor) ContractByID(id types.FileContractID) (modules.RenterContra
 func (c *Contractor) Contracts() []modules.RenterContract {
 	c.mu.RLock()
 	defer c.mu.RUnlock()
-	return c.contracts.ViewAll()
+	return c.staticContracts.ViewAll()
```
DavidVorick: If this is static, we should be good removing the readlock and unlock here.

modules/renter/contractor/contractor_test.go
```diff
@@ -459,7 +459,7 @@ func TestIntegrationSetAllowance(t *testing.T) {
 		t.Fatal(err)
 	}
 	c.mu.Lock()
-	clen := c.contracts.Len()
+	clen := c.staticContracts.Len()
 	c.mu.Unlock()
```
DavidVorick: it looks like we lock and then unlock just to get the length? Should be good to drop that lock and unlock, right?

modules/renter/files.go
```diff
@@ -257,21 +258,30 @@ func (r *Renter) DeleteFile(nickname string) error {
...
 	var files []*file
+	contractIDs := make(map[types.FileContractID]struct{})
 	lockID := r.mu.RLock()
 	for _, f := range r.files {
 		files = append(files, f)
+		for cid := range f.contracts {
+			contractIDs[cid] = struct{}{}
+		}
 	}
 	r.mu.RUnlock(lockID)
```
DavidVorick: This is a lock that's potentially going to be held for a long time, if there are a lot of files to iterate through.

DavidVorick: I guess there isn't any super easy way to deal with it.

PR #2996, opened 2018-05-07
add blockid,blockheight, transactionid,siacoinoutputid to /consensus/blocks endpoint

hupengbuaa:
I made several small changes to get the data I need to construct transactions manually. Wish you could help check if it's right. No need to merge to the codebase.

### DIFF
--- a/modules/consensus.go
+++ b/modules/consensus.go
@@ -192,7 +192,7 @@ type (
 
 		// BlocksByID returns a block found for a given ID, with a bool to
 		// indicate whether that block exists.
-		BlockByID(types.BlockID) (types.Block, bool)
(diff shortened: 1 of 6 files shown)


lukechampine:
Now that we've merged https://github.com/NebulousLabs/Sia/pull/2980, you can get the block ID and height. As for the siacoin output IDs, that's a bit trickier. Instead of modifying `types.SiacoinOutput`, you could have the response include a `map[types.UnlockHash]types.SiacoinOutputID`. You could also replace the `types.SiacoinOutput` with a `modules.ProcessedOutput`.

DavidVorick:
Looks like we've merged all the relevant changes, closing this for now. Feel free to re-open if the master branch is still missing things

PR #2997, opened 2018-05-07
Add tg.Add to exported wallet methods

ChrisSchinnerl:
Exported methods of top-level modules that are called by the API directly should always call `tg.Add`. Especially if those methods access closable resources. 
e.g. `wallet.ConfirmedBalance` can currently be called during a shutdown which results in a error since `ConfirmedBalance` calls `syncDB` on a closed database.

### DIFF
--- a/modules/consensus/accept_bench_test.go
+++ b/modules/consensus/accept_bench_test.go
@@ -129,7 +129,10 @@ func BenchmarkAcceptSmallBlocks(b *testing.B) {
 	for j := 0; j < b.N; j++ {
 		// Create a transaction with a miner fee, a normal siacoin output, and
 		// a funded file contract.
-		txnBuilder := cst.wallet.StartTransaction()
+		txnBuilder, err := cst.wallet.StartTransaction()
+		if err != nil {
+			b.Fatal(err)
+		}
 		err = txnBuilder.FundSiacoins(types.NewCurrency64(125e6))
 		if err != nil {
 			b.Fatal(err)
--- a/modules/consensus/accept_test.go
+++ b/modules/consensus/accept_test.go
@@ -504,7 +504,10 @@ func TestIntegrationDoSBlockHandling(t *testing.T) {
 
 	// Mine a block that is valid except for containing a buried invalid
 	// transaction. The transaction has more siacoin inputs than outputs.
-	txnBuilder := cst.wallet.StartTransaction()
+	txnBuilder, err := cst.wallet.StartTransaction()
+	if err != nil {
+		t.Fatal(err)
+	}
 	err = txnBuilder.FundSiacoins(types.NewCurrency64(50))
 	if err != nil {
 		t.Fatal(err)
@@ -794,7 +797,10 @@ func TestBuriedBadTransaction(t *testing.T) {
 
 	// Create a good transaction using the wallet.
 	txnValue := types.NewCurrency64(1200)
-	txnBuilder := cst.wallet.StartTransaction()
+	txnBuilder, err := cst.wallet.StartTransaction()
+	if err != nil {
+		t.Fatal(err)
+	}
 	err = txnBuilder.FundSiacoins(txnValue)
 	if err != nil {
 		t.Fatal(err)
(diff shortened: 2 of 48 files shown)


ChrisSchinnerl:
I'm not sure how to handle methods which don't return an `error` in this PR though. Right now I simply return the default datatypes but a user calling the API might expect a custom shutdown-error.

DavidVorick:
Well, we could always just add return values. If it's possible for the call to fail and for the user to recieve unexpected data, the function should definitely have an error return value.

So I would change the function signatures.

lukechampine:
I don't think it's worth it to overhaul everything and complicate every piece of code that touches the wallet just so that we can report one specific type of error that occurs under one specific circumstance. The same issue applies to other modules, so inevitably we'd have to add an `error` return value to every single method on every module.
You can avoid all of this by arranging things so that you never call a module's methods after you've called `Close`. That requires some careful code in the `node` package, but nowhere else. I think it's reasonable for a module to say "behavior of these methods is undefined after Close has been called."

lukechampine:
The methods of `os.File` all have to return errors anyway, because all I/O operations can fail. So there's no ergonomic penalty to adding a `file is closed` error. Notably, there are two methods of `os.File` that don't return an error: `Name` and `Fd`. [It looks like](https://play.golang.org/p/UrvYmr2mgo8) after calling `Close`, `Name` continues to return the file's name, and `Fd` returns garbage. Unfortunately there aren't many other examples in the standard library.

Calling methods after `Close` should be treated the same way as a "use after free" bug. Calling methods on a `nil` pointer causes a segfault. Sending to a closed channel results in a panic. These are developer errors.

>Closing a channel releases it as a resource. It makes no more sense to close a channel multiple times than it makes to close a file descriptor multiple times, or free a block of allocated memory multiple times. Such actions imply the code is broken, which is why closing a closed channel triggers a panic. (Rob Pike)

I do agree with your point about network-attached modules. But we can't blindly assume that every method needs to return an error because it *might* be implemented weirdly. There are lots of interfaces that don't expose any errors, like `fmt.Stringer` and `hash.Hash`, because it would be silly for them to do I/O. Similarly, `(*os.File).Name` doesn't return an error because the name is always small enough to keep around in memory; you should never have to do I/O to retrieve it. So I think we just need to use our best judgment when deciding which methods need errors. Specifically, I'm okay with adding errors to methods like `(modules.Wallet).SetSettings`, because it's reasonable to assume that an implementation of `SetSettings` would need to do I/O. I'm not okay with adding an error to `Rescanning` or `Unlocked`: if the wallet is closed, the logical return value of these methods is `false`. `Height` and `DustThreshold` are debatable, since they're small enough value to keep in memory, and it's much more ergonomic for them to not return an error. Lastly, methods like `Transaction` should probably just return an error instead of an error and a bool.

tl;dr: I'm okay adding errors to most of these methods, not because they should be safe to call after `Close`, but because they should be returning an error anyway.

One last thing that comes to mind is this commit: https://github.com/NebulousLabs/Sia/pull/1615/commits/5bef693b9144bf1e9eb8b9e9bbb0c31a7afad879. `crypto.SignHash` originally returned an error, even though it could never fail, because "switching libraries in the future may result in errors that can be returned." This turned out to be a case of premature generalization, and removing the error cleaned up a ton of code.

ChrisSchinnerl:
You are right, a `os.File` was probably not an appropriate comparison because in my opinion it's not necessarily a developer mistake to access the API after a shutdown. A webserver that is used by services like Nextcloud, Plex etc. should be able to receive calls during a graceful shutdown instead of being killed by a panic. e.g. I'd like to be able to take my Sia node offline, upgrade it and restart it while other services are using it without having to shut them down.

I don't think it's a problem for `Height` and `DustThreshold`to return an error. A `Height` of 0 to indicate an error while the http status code is still `200` feels like a violation of http conventions.
I also think it's reasonable for `DustThreshold` to return an error. `DustThreshold` calls `FeeEstimation` on the transactionpool which in turn calls methods like `requiredFeesToExtendPool` that could potentially use I/O in the future. I shouldn't have to worry as a dev about changing the signature of `DustThreshold` just because I changed the way the transactionpool does fee estimation.

DavidVorick:
I'm definitely in favor of having the function signatures all return errors. We had this discussion over a year ago, and we settled with not having errors. In this case, it has caused a handful of crashes for users. It also has forced us to change our API/intefaces for some functions which previously didn't touch disk, but now do touch disk because we switched from an in-memory wallet to a database wallet.

For complex, dynamic interfaces where we are continuing to upgrade the implementation and add new features, having error checking on all the functions is a defensive way to program that should limit bugs and allow us to be creative in the future without breaking compatibility.

PR #2998, opened 2018-05-07
turn flaky test in to a vlong test

DavidVorick:


ChrisSchinnerl:
Should we replace this test before merging this instead of just turning it into a `vlong` test?

Issue #1920, opened 2017-06-12
Add a /renter/file/__siapath__ endpoint


nitronick600:
It's inefficient to parse to the whole sia filesystem looking for a single file.  A singular endpoint would make sense.

tbenz9:
recommendation to devs: tag "feature request" and "Programming task"

MSevey:
@nitronick600 I just submitted a PR to add the endpoint, feel free to check it out #3003

tbenz9:
Closed via #3003

PR #3003, opened 2018-05-08
I1920 single file api

MSevey:
Addressing issue #1920 to add an API endpoint to request a single file.  Added test in the siatest package and updated documentation.

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -997,19 +997,17 @@ lists the status of specified file.
 ###### JSON Response [(with comments)](/doc/api/Renter.md#json-response-4)
 ```javascript
 {
-  "file": [
-    {
-      "siapath":        "foo/bar.txt",
-      "localpath":      "/home/foo/bar.txt",
-      "filesize":       8192, // bytes
-      "available":      true,
-      "renewing":       true,
-      "redundancy":     5,
-      "bytesuploaded":  209715200, // total bytes uploaded
-      "uploadprogress": 100, // percent
-      "expiration":     60000
-    }
-  ]
+  "file": {
+    "siapath":        "foo/bar.txt",
+    "localpath":      "/home/foo/bar.txt",
+    "filesize":       8192, // bytes
+    "available":      true,
+    "renewing":       true,
+    "redundancy":     5,
+    "bytesuploaded":  209715200, // total bytes uploaded
+    "uploadprogress": 100, // percent
+    "expiration":     60000
+  }
 }
 ```
 
--- a/doc/API.md
+++ b/doc/API.md
@@ -845,21 +845,21 @@ description of the byte encoding.
 Renter
 ------
 
-| Route                                                                                 | HTTP verb |
-| ------------------------------------------------------------------------------------- | --------- |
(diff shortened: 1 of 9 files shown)


node/api/renter.go
```diff
@@ -373,6 +379,15 @@ func (api *API) renterRenameHandler(w http.ResponseWriter, req *http.Request, ps
...
 }
 
+// renterFileHandler handles the API call to return specific file.
+func (api *API) renterFileHandler(w http.ResponseWriter, req *http.Request, ps httprouter.Params) {
+	file, err := api.renter.File(strings.TrimPrefix(ps.ByName("siapath"), "/"))
+	WriteJSON(w, RenterFile{
+		File:  file,
+		Error: err,
+	})
+}
```
lukechampine: instead of including the error in the JSON response, this endpoint should probably signal the error using a non-200 status code, e.g.
```go
file, err := api.renter.File(strings.TrimPrefix(ps.ByName("siapath"), "/"))
if err != nil {
    WriteError(w, Error{err.Error()}, http.StatusBadRequest)
    return
}
WriteJSON(w, file)
```

MSevey: @lukechampine will passing file into WriteJSON like like that negate the need for a RenterFile struct?

Also would that then make the resulting JSON be like so?
```
{
    {
      "siapath":        "foo/bar.txt",
      "localpath":      "/home/foo/bar.txt",
      "filesize":       8192, // bytes
      "available":      true,
      "renewing":       true,
      "redundancy":     5,
      "bytesuploaded":  209715200, // total bytes uploaded
      "uploadprogress": 100, // percent
      "expiration":     60000
    }
}
```

node/api/renter.go
```diff
@@ -123,6 +123,12 @@ type (
 		Downloads []DownloadInfo `json:"downloads"`
 	}
 
+	// RenterFile lists the file queried.
+	RenterFile struct {
+		File  modules.FileInfo `json:"file"`
+		Error error            `json:"error"`
```
ChrisSchinnerl: I don't think you need this `Error` field anymore.

doc/API.md
```diff
@@ -989,11 +990,34 @@ lists the status of all files.
...
 
+#### /renter/file/*__siapath__ [GET]
+
+lists the status of specified file.
+
+###### JSON Response [(with comments)](/doc/api/Renter.md#json-response-4)
+```javascript
+{
+  "file": [
+    {
```
ChrisSchinnerl: I don't think that this is the correct format since the `RenterFile` struct doesn't contain a slice of files but instead only a single file. You should double check that using `curl` but I think `[` and `]` are not necessary.

doc/api/Renter.md
```diff
@@ -309,6 +310,58 @@ lists the status of all files.
...
 
+#### /renter/file/*___siapath___ [GET]
+
+lists the status of specified file.
+
+###### JSON Response
+```javascript
+{
+  "file": [ 
+    {
```
ChrisSchinnerl: same here

PR #3004, opened 2018-05-08
Fix memory leak in repair/upload code

ChrisSchinnerl:
Previously the `logicalChunkData` of an `unfinishedUploadChunk` was a simple byte slice that was erasure coded into the shards/pieces that are eventually distributed across many hosts.
The problem is, that this slice is a single allocated block of memory. This means we can't free the memory for individual shards before all the pieces using this memory are uploaded.
This PR changes the `logicalChunkData` to a `[][]byte` slice and makes sure that every `[]byte` in it was allocated separately.


modules/renter/shardreader.go
```diff
@@ -0,0 +1,34 @@
...
+// NewShardReader creates a new shardReader from an object that implements the
+// ReaderAt interface.
+func NewShardReader(r io.ReaderAt) *shardReader {
+	return &shardReader{
+		r: r,
+	}
+}
+
+// ReadAt reads data into a slice of shards from a certain offset.
+func (sr *shardReader) ReadAt(d [][]byte, offset int64) (int, error) {
```
DavidVorick: isn't this the wrong type signature?

ChrisSchinnerl: Yeah but you can't really create a 100% correct signature if you want to read into a slice of slices.

lukechampine:
Instead of adding `shardReader`, I suggest adding one new method to `downloadDestinationBuffer`:

```go
func (buf downloadDestinationBuffer) ReadFrom(r io.Reader) (int64, error) {
    // read from r until shards are full
}
```
Then in `managedFetchLogicalChunkData`, do:
```go
buf := NewDownloadDestinationBuffer(chunk.length)
sr := io.NewSectionReader(osFile, chunk.offset, chunk.length*len(buf))
_, err := buf.ReadFrom(sr)
```

modules/renter/downloaddestination.go
```diff
@@ -37,20 +39,56 @@ type downloadDestination interface {
...
 // Close implements Close for the downloadDestination interface.
 func (dw downloadDestinationBuffer) Close() error {
 	return nil
 }
 
+// ReadFrom reads data from a io.Reader until the buffer is full.
+func (dw downloadDestinationBuffer) ReadFrom(r io.Reader) (int64, error) {
+	var n int64
+	for len(dw) > 0 {
+		read, err := r.Read(dw[0])
```
DavidVorick: should this be a ReadAll?

DavidVorick: can use the io.ReadFull helper

modules/renter/downloaddestination.go
```diff
@@ -37,20 +39,56 @@ type downloadDestination interface {
...
+		dw = dw[1:]
+		n += int64(read)
+	}
+	return n, nil
+}
+
 // WriteAt writes the provided data to the downloadDestinationBuffer.
 func (dw downloadDestinationBuffer) WriteAt(data []byte, offset int64) (int, error) {
-	if len(data)+int(offset) > len(dw) || offset < 0 {
+	if uint64(len(data)+int(offset)) > uint64(len(dw))*modules.SectorSize || offset < 0 {
```
DavidVorick: is there any reason that you have to wrap these in a `uint64`?

ChrisSchinnerl: Yeah the right side of the equation is a `uint64`, `len` is a `ìnt` and `offset` is a int64.

PR #3010, opened 2018-05-10
Heap cache sort

MSevey:
Implemented heap for the download cache for deleting least recently accessed chunk.  Added API endpoint to be able to manually set the download cache size.  Added Siatest to confirm API endpoint is working.  Updated API documentation

### DIFF
--- a/doc/api/Renter.md
+++ b/doc/api/Renter.md
@@ -70,9 +70,9 @@ returns the current settings along with metrics on the renter's spending.
     // manage bandwidth
     "maxdownloadspeed":   1234, // bytes per second
 
-    // The DownloadCacheSize is the number of data chunks that will be cached during
+    // The StreamCacheSize is the number of data chunks that will be cached during
     // streaming
-    "downloadcachesize":  4  
+    "streamcachesize":  4  
   },
 
   // Metrics about how much the Renter has spent on storage, uploads, and
@@ -138,9 +138,9 @@ maxdownloadspeed
 // setting is not persisted and will be reset by a shutdown
 maxuploadspeed
 
-// Download cache size specifies how many data chunks will be cached while 
+// Stream cache size specifies how many data chunks will be cached while 
 // streaming.  Must be greater than 0
-downloadcachesize
+streamcachesize
 ```
 
 ###### Response
--- a/doc/api/Renter.md
+++ b/doc/api/Renter.md
@@ -129,6 +129,16 @@ period // block height
 // fewer total transaction fees. Storage spending is not affected by the renew
 // window size.
 renewwindow // block height
+
+// Max download speed permitted
+maxdownloadspeed
+
+// Max upload speed permitted
+maxuploadspeed
+
+// Download cache size specifies how many data chunks will be cached while 
+// streaming.  Must be greater than 0
+downloadcachesize
 ```
 
 ###### Response
--- a/doc/api/Renter.md
+++ b/doc/api/Renter.md
@@ -130,10 +130,12 @@ period // block height
 // window size.
 renewwindow // block height
 
-// Max download speed permitted
+// Max download speed permitted, speed provide in bytes per second
+// setting is not persisted and will be reset by a shutdown
 maxdownloadspeed
 
-// Max upload speed permitted
+// Max upload speed permitted, speed provide in bytes per second
+// setting is not persisted and will be reset by a shutdown
 maxuploadspeed
 
 // Download cache size specifies how many data chunks will be cached while 
--- a/doc/api/Renter.md
+++ b/doc/api/Renter.md
@@ -139,7 +139,7 @@ maxdownloadspeed
 maxuploadspeed
 
 // Stream cache size specifies how many data chunks will be cached while 
-// streaming.  Must be greater than 0
+// streaming.  
 streamcachesize
 ```
 
--- a/modules/renter/download.go
+++ b/modules/renter/download.go
@@ -437,8 +437,8 @@ func (r *Renter) managedNewDownload(params downloadParams) (*download, error) {
 			physicalChunkData: make([][]byte, params.file.erasureCode.NumPieces()),
 			pieceUsage:        make([]bool, params.file.erasureCode.NumPieces()),
 
-			download:           d,
-			downloadChunkCache: r.downloadChunkCache,
+			download:    d,
+			streamCache: r.streamCache,
 		}
 
 		// Set the fetchOffset - the offset within the chunk that we start
--- a/modules/renter/download.go
+++ b/modules/renter/download.go
@@ -437,9 +437,10 @@ func (r *Renter) managedNewDownload(params downloadParams) (*download, error) {
 			physicalChunkData: make([][]byte, params.file.erasureCode.NumPieces()),
 			pieceUsage:        make([]bool, params.file.erasureCode.NumPieces()),
 
-			download:   d,
-			chunkCache: r.chunkCache,
-			cacheMu:    r.cmu,
+			download:       d,
+			chunkCacheMap:  r.chunkCacheMap,
+			chunkCacheHeap: r.chunkCacheHeap,
+			cacheMu:        r.cmu,
 		}
 
 		// Set the fetchOffset - the offset within the chunk that we start
--- a/modules/renter/download.go
+++ b/modules/renter/download.go
@@ -437,8 +437,8 @@ func (r *Renter) managedNewDownload(params downloadParams) (*download, error) {
 			physicalChunkData: make([][]byte, params.file.erasureCode.NumPieces()),
 			pieceUsage:        make([]bool, params.file.erasureCode.NumPieces()),
 
-			download:    d,
-			streamCache: r.streamCache,
+			download:          d,
+			staticStreamCache: r.staticStreamCache,
 		}
 
 		// Set the fetchOffset - the offset within the chunk that we start
--- a/modules/renter/download.go
+++ b/modules/renter/download.go
@@ -437,10 +437,9 @@ func (r *Renter) managedNewDownload(params downloadParams) (*download, error) {
 			physicalChunkData: make([][]byte, params.file.erasureCode.NumPieces()),
 			pieceUsage:        make([]bool, params.file.erasureCode.NumPieces()),
 
-			download:       d,
-			chunkCacheMap:  r.chunkCacheMap,
-			chunkCacheHeap: r.chunkCacheHeap,
-			cacheMu:        r.cmu,
+			download:           d,
+			downloadChunkCache: r.downloadChunkCache,
+			cacheMu:            r.cmu,
 		}
 
 		// Set the fetchOffset - the offset within the chunk that we start
--- a/modules/renter/download.go
+++ b/modules/renter/download.go
@@ -439,7 +439,6 @@ func (r *Renter) managedNewDownload(params downloadParams) (*download, error) {
 
 			download:           d,
 			downloadChunkCache: r.downloadChunkCache,
-			cacheMu:            r.cmu,
 		}
 
 		// Set the fetchOffset - the offset within the chunk that we start
--- a/modules/renter/downloadcache.go
+++ b/modules/renter/downloadcache.go
@@ -4,11 +4,59 @@ package renter
 // via the API.
 
 import (
+	"container/heap"
 	"time"
 
 	"github.com/NebulousLabs/Sia/build"
 	"github.com/NebulousLabs/errors"
 )
+
+// Required functions for use of heap for chunkCacheHeap
+func (cdpq chunkCacheHeap) Len() int { return len(cdpq) }
+
+// Less returns the lessor of two elements
+func (cdpq chunkCacheHeap) Less(i, j int) bool { return cdpq[i].lastAccess < cdpq[j].lastAccess }
+
+// Swap swaps two elements from the heap
+func (cdpq chunkCacheHeap) Swap(i, j int) {
+	cdpq[i], cdpq[j] = cdpq[j], cdpq[i]
+	cdpq[i].index = i
+	cdpq[j].index = j
+}
+
+// Push adds an element to the heap
+func (cdpq *chunkCacheHeap) Push(x interface{}) {
+	n := len(*cdpq)
+	cacheData := x.(*cacheData)
+	cacheData.index = n
+	*cdpq = append(*cdpq, cacheData)
+}
+
+// Pop removes element from the heap
+func (cdpq *chunkCacheHeap) Pop() interface{} {
+	old := *cdpq
+	n := len(old)
+	cacheData := old[n-1]
+	cacheData.index = -1 // for safety
+	*cdpq = old[0 : n-1]
+	return cacheData
+}
+
+// PopNoReturn will pop the element off the heap and is for when you do need the value
+func (cdpq *chunkCacheHeap) PopNoReturn() {
+	old := *cdpq
+	n := len(old)
+	cacheData := old[n-1]
+	cacheData.index = -1 // for safety
+	*cdpq = old[0 : n-1]
+}
+
+// update, updates the heap and reorders
+func (cdpq *chunkCacheHeap) update(cd *cacheData, data []byte, lastAccess int64) {
+	cd.data = data
+	cd.lastAccess = lastAccess
+	heap.Fix(cdpq, cd.index)
+}
 
 // addChunkToCache adds the chunk to the cache if the download is a streaming
 // endpoint download.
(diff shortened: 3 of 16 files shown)


modules/renter/downloadcache.go
```diff
@@ -4,72 +4,152 @@ package renter
...
-	udc.chunkCache[udc.staticCacheID] = &cacheData{
+	// Add chunk to Map and Heap
+	cd := &chunkData{
+		id:         cacheID,
 		data:       data,
 		lastAccess: time.Now(),
 	}
+	dcc.chunkCacheMap[cacheID] = cd
+	heap.Push(&dcc.chunkCacheHeap, cd)
+	dcc.chunkCacheHeap.update(cd, cd.id, cd.data, cd.lastAccess)
```
ChrisSchinnerl: I think this is redundant. `Push` and `Pop` should not require an extra call to `update`.

modules/renter/downloadcache.go
```diff
@@ -4,72 +4,152 @@ package renter
...
+}
+
+// init initializes the downloadChunkCache
+func (dcc *downloadChunkCache) init() {
```
ChrisSchinnerl: Just for the sake of completeness `init` should be `Init`, be threadsafe (acquire lock) and it should probably contain a sanity check (`build.Critical`) to make sure we can't call init twice on the same cache.
`heap.Init` is idempotent but our custom `Init` isn't.

modules/renter/downloadcache.go
```diff
@@ -4,72 +4,152 @@ package renter
...
 // if it was the last missing chunk. The function returns true if the chunk was
 // in the cache.
-// TODO in the future we might need cache invalidation. At the
+// Using the entire unfisihedDownloadChunk as the argument as there are seven different fields
+// used from unfinishedDownloadChunk and it allows using udc.fail()
+//
+// TODO: in the future we might need cache invalidation. At the
 // moment this doesn't worry us since our files are static.
-func (r *Renter) managedTryCache(udc *unfinishedDownloadChunk) bool {
+func (dcc *downloadChunkCache) retrieve(udc *unfinishedDownloadChunk) bool {
```
ChrisSchinnerl: This needs to be `Retrieve`. That way it is clear for a caller that no lock can be held while calling `Retrieve` and that the `chunkCache` will handle locking itself.

modules/renter/downloadcache.go
```diff
@@ -84,3 +164,8 @@ func (r *Renter) managedTryCache(udc *unfinishedDownloadChunk) bool {
 	udc.download.mu.Unlock()
 	return true
 }
+
+// s
```
ChrisSchinnerl: missing description

modules/renter/downloadcache_test.go
```diff
@@ -1,36 +1,83 @@
...
 	"testing"
 	"time"
 )
 
-// TestAddChunkToCache tests that the oldest chunk is removed
-func TestAddChunkToCache(t *testing.T) {
+// TestAdd tests that the oldest chunk is removed
+func TestAdd(t *testing.T) {
+	// Initializing minimum variables
 	udc := &unfinishedDownloadChunk{
```
ChrisSchinnerl: I think we don't need the whole `udc` in this test anymore after changing the `Add` method and moving the type check out of `Add`. We can just create and init the `downloadChunkCache` directly.

node/api/client/renter.go
```diff
@@ -93,6 +94,18 @@ func (c *Client) RenterPricesGet() (rpg api.RenterPricesGET, err error) {
 	return
 }
 
+// RenterPostDownloadCacheSize uses the /renter endpoint to change the renter's
+// downloadCacheSize for streaming
+func (c *Client) RenterPostDownloadCacheSize(cacheSize uint64) (err error) {
+	if cacheSize <= 0 {
+		return errors.New("Download Cache Size must be greater than 0")
```
ChrisSchinnerl: The client library doesn't need to do any checks. It should just serialize the inputs provided in safe types to strings and send the request to the API. The cacheSize check only needs to happen on the siad side. `SetSettings` is probably a good place for that.

node/api/client/renter.go
```diff
@@ -93,6 +94,18 @@ func (c *Client) RenterPricesGet() (rpg api.RenterPricesGET, err error) {
 	return
 }
 
+// RenterPostDownloadCacheSize uses the /renter endpoint to change the renter's
+// downloadCacheSize for streaming
+func (c *Client) RenterPostDownloadCacheSize(cacheSize uint64) (err error) {
```
ChrisSchinnerl: This should be `RenterSetStreamCacheSizePost`. I know that there some functions named `RenterPostXXX` but that was actually a mistake. It's not that easy in this case but usually a library method for an endpoint of the form `/end/point/path` should have the function name `EndPointPathPost` or `EndPointPathGet`.

siatest/renter/renter_test.go
```diff
@@ -278,6 +279,22 @@ func testRenterStreamingCache(t *testing.T, tg *siatest.TestGroup) {
 	// Grab the first of the group's renters
 	r := tg.Renters()[0]
 
+	// Testing setting downloadCacheSize for streaming
+	// Test setting it to larger than the defaultCacheSize
+	if err := r.RenterPostDownloadCacheSize(4); err != nil {
+		t.Fatal(err, "Could not set DownloadCacheSize to 4")
+	}
+
+	// Test resetting to the value of defaultDownloadCacheSize (2)
```
ChrisSchinnerl: This test needs to check if there are actually 4 cached chunks after setting it to 4 and after setting it back to 2 it needs to check if only 2 are cached.

modules/renter/downloadcache_test.go
```diff
@@ -1,36 +1,83 @@
...
 			staticDestinationType: destinationTypeSeekStream,
 		},
-		chunkCache: make(map[string]*cacheData),
-		cacheMu:    new(sync.Mutex),
+		downloadChunkCache: new(downloadChunkCache),
 	}
 
+	udc.downloadChunkCache.init()
+
+	// Testing Push to Heap
```
ChrisSchinnerl: I'd rename `TestAdd`to `TestStreamCache` and test the cache using its exported high-level methods which add chunks, change the cacheSize etc. It should also check if the least-recently-accessed element is removed and not merely the least-recently-added.
Then you could create a second low-level test `TestHeapImplementation` that checks if your custom heap works as expected (`Push`, `Pop`, `update`).

modules/renter/downloadcache.go
```diff
@@ -4,72 +4,152 @@ package renter
...
 	cd.lastAccess = time.Now()
-	r.chunkCache[udc.staticCacheID] = cd
+	dcc.chunkCacheMap[udc.staticCacheID] = cd
+	dcc.chunkCacheHeap.update(cd, cd.id, cd.data, cd.lastAccess)
 
 	start := udc.staticFetchOffset
 	end := start + udc.staticFetchLength
 	_, err := udc.destination.WriteAt(cd.data[start:end], udc.staticWriteOffset)
 	if err != nil {
-		r.log.Println("WARN: failed to write cached chunk to destination:", err)
```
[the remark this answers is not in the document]
MSevey: with the renter no longer as the receiver we can't call `r.log` 
we still capture the error with the `udc.fail`

modules/renter/download.go
```diff
@@ -434,9 +434,8 @@ func (r *Renter) managedNewDownload(params downloadParams) (*download, error) {
 			physicalChunkData: make([][]byte, params.file.erasureCode.NumPieces()),
 			pieceUsage:        make([]bool, params.file.erasureCode.NumPieces()),
 
-			download:   d,
-			chunkCache: r.chunkCache,
-			cacheMu:    r.cmu,
+			download:           d,
+			downloadChunkCache: r.downloadChunkCache,
```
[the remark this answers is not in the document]
MSevey: @ChrisSchinnerl Are we sure we wouldn't use some of this functionality for future downloads? Whether it is large file downloads, partial downloads, folder downloads, etc. I agree that in the current context streamChunkCache would be more accurate, but can you think of a benefit of keeping it generic for future use?

modules/renter/downloadcache.go
MSevey: @ChrisSchinnerl this is how the go documentation recommended it was done.  `Push` and `Pop` only add and remove elements.  After pushing an element onto the heap, the heap must be reordered with `update` to confirm it is still ordered properly

modules/renter/download.go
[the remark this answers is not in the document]
ChrisSchinnerl: I don't think we would want to keep that functionality. Caching should usually not be implemented by Sia but by the user's application instead. Once partial downloads are implemented we can get rid of the caching.

modules/renter/downloadcache.go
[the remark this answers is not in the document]
ChrisSchinnerl: Ah yeah, you are right.

siatest/renter/renter_test.go
```diff
@@ -311,6 +311,45 @@ func testRenterStreamingCache(t *testing.T, tg *siatest.TestGroup) {
 	// Grab the first of the group's renters
 	r := tg.Renters()[0]
 
+	// Testing setting downloadCacheSize for streaming
+	// Test setting it to larger than the defaultCacheSize
+	if err := r.RenterSetStreamCacheSizePost(4); err != nil {
```
ChrisSchinnerl: This test is pretty good but it also needs to check if increasing/decreasing the cache size actually increases/decreases the cache. So it would be good to set it to 2, upload a 2 chunk file and see if it is as fast as expected, then set it to 1 and check if the 2 chunk file takes longer to download now since setting the cache size to a smaller value sould prune the cache to that size right away. Finally set it to 3 and make sure you can download a 3 chunk file.

MSevey: As discussed, adding an additional unit test to verify cache is pruned when `cacheSize` is set to a smaller value, and making sure the cache is pruned from within the `SetStreamingCacheSize` method

modules/renter/streamcache.go
```diff
@@ -0,0 +1,181 @@
...
+		return
+	}
+
+	sc.streamMap = make(map[string]*chunkData)
+	sc.streamHeap = make(streamHeap, 0, defaultStreamCacheSize)
+	sc.cacheSize = defaultStreamCacheSize
+	heap.Init(&sc.streamHeap)
+}
+
+// pruneCache prunes the cache until it is the length of cacheSize
```
ChrisSchinnerl: The comment needs to be updated. `pruneCache` prunes until the cache has the length `size`.

modules/renter/streamcache.go
```diff
@@ -0,0 +1,181 @@
...
+func (sc *streamCache) pruneCache(size uint64) {
+	for len(sc.streamMap) > int(size) {
```
ChrisSchinnerl: Just to be safe we should add a sanity check that makes sure that `size` is not smaller than `0`. Otherwise we prune forever.

modules/renter/streamcache_test.go
```diff
@@ -0,0 +1,135 @@
...
+			lastAccess: time.Now(),
+		})
+	}
+
+	// Testing Heap update
+	// Confirming recently accessed elements get moved to the bottom of Heap
+	cd = sc.streamHeap[0]
+	sc.streamHeap.update(cd, cd.id, cd.data, time.Now())
+	if reflect.DeepEqual(cd, sc.streamHeap[0]) {
+		t.Error("Heap order was not updated. Recently accessed element not at bottom of heap")
```
ChrisSchinnerl: I don't think this check is necessary since it only proves that the new element doesn't end up as the root of the heap. If you still want to keep it you should update the error message though.

modules/renter/streamcache.go
ChrisSchinnerl: Nevermind. It's a `uint` so it should be fine.

doc/api/Renter.md
```diff
@@ -129,6 +129,18 @@ period // block height
...
+// setting is not persisted and will be reset by a shutdown
+maxdownloadspeed
+
+// Max upload speed permitted, speed provide in bytes per second
+// setting is not persisted and will be reset by a shutdown
+maxuploadspeed
+
+// Stream cache size specifies how many data chunks will be cached while 
+// streaming.  Must be greater than 0
+streamcachesize
```
lukechampine: I don't think specifying `Must be greater than 0` is necessary; the same is true for many other settings, and the wording makes it sounds like you *have* to set `streamcachesize` explicitly.

modules/renter/renter.go
```diff
@@ -193,18 +193,18 @@ type Renter struct {
...
-	deps           modules.Dependencies
-	g              modules.Gateway
-	hostContractor hostContractor
-	hostDB         hostDB
-	log            *persist.Logger
-	persistDir     string
-	mu             *siasync.RWMutex
-	tg             threadgroup.ThreadGroup
-	tpool          modules.TransactionPool
+	staticStreamCache *streamCache
```
lukechampine: does it make sense to call this `static`? I guess the renter doesn't modify it directly, but it's still modified by its methods, same as `tpool`/`hostDB`. I would just call it `streamCache`.

modules/renter/renter.go
```diff
@@ -433,17 +435,18 @@ func NewCustomRenter(g modules.Gateway, cs modules.ConsensusSet, tpool modules.T
...
+		cs:                cs,
+		deps:              deps,
+		g:                 g,
+		hostDB:            hdb,
+		hostContractor:    hc,
+		persistDir:        persistDir,
+		mu:                siasync.New(modules.SafeMutexDelay, 1),
+		tpool:             tpool,
 	}
+	r.staticStreamCache.Init()
```
lukechampine: if possible, replace the `Init` method with a `newStreamCache()` constructor.

modules/renter/streamcache.go
```diff
@@ -0,0 +1,181 @@
...
+	streamMap  map[string]*chunkData
+	streamHeap streamHeap
+	cacheSize  uint64
+	mu         sync.Mutex
+}
+
+// Required functions for use of heap for streamHeap
+func (sh streamHeap) Len() int { return len(sh) }
+
+// Less returns the lessor of two elements
```
lukechampine: lesser :wink: 
(A lessor is your landlord.)

modules/renter/streamcache.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+	return true
+}
+
+// SetStreamingCacheSize confirms that the cache size is being set
+// to a value greater than zero.  Otherwise it will remain the default
+// value set during the initialization of the streamCache.
+// It will also prune the cache to ensure the cache is always
+// less than or equal to whatever the cacheSize is set to
+func (sc *streamCache) SetStreamingCacheSize(cacheSize uint64) {
```
lukechampine: this behavior is a little unusual. I think I'd prefer to punt the `cacheSize > 0` check up to the caller of `SetStreamingCacheSize`.

modules/renter/renter.go
ChrisSchinnerl: I think it makes sense since `static` refers to the lock of the object that holds the `static` field. That means I only have read-only access to the pointer and can't replace it during the lifetime of the renter.
When I call a private method on a `static` object I still need to hold the object's lock, but I know that I don't have to hold the Renter's lock. On the other hand if I call a public method on the `static` object I can't hold any locks even though `static` usually means I can hold the lock of the owning object if I want to.

ChrisSchinnerl: That's btw also the reason why the `staticContracts` in the contractor are static now, even though the field has methods that modify its contents.

lukechampine: Doesn't the `streamCache` handle all of its own locking internally though? Speaking of which, do we still need `cmu`?

There isn't really a precedent for this so we'll have to use our best judgment. The `streamCache` looks to me like a dependency, not very different from the `tpool` or `hostDB`. If it could (in theory) be moved into its own package, I think that qualifies it as a dependency, and we currently don't apply the `static` convention to dependencies.

ChrisSchinnerl: We can probably get rid of the `cmu`.

Well, if we consider the `contractSet` a dependency, then we do apply the `static` convention already. The problem is, that there is no way to tell if you have to hold the lock of the parent object without making it `static`. In my opinion that introduces unnecessary ambiguity when mixing dependencies and simple helper structs.
I also don't think that calling a method that modifies the object's fields on an object violates the `static` rule. Static refers to the value of the pointer, which doesn't change since it is only a reference.

lukechampine: good point about the contract set. So there is a precedent, but we aren't very consistent about it; `hostContractor` stores the field as `staticContracts`, but `proto.Editor` and `proto.Downloader` just call it `contractSet`.

I think I see what you're getting at. When we call a method on a dependency, e.g. `r.tpool.AcceptTransactionSet`, we assume that `tpool` will handle the locking. But for a "helper" object, like `file`, it may be necessary to hold the parent lock. So how do you tell when you need to hold the lock?

Things get messy because sometimes you want to hold the parent lock throughout multiple helper operations, like calling multiple methods on a `file`. This is (IIRC) disallowed for dependencies because of deadlock risk. I think we previously encountered this when splitting out the `hostTree` from the `hostDB`.

Worse, I don't think we can introduce a convention that alters method names (like `managed`) because that could prevent a helper type from satisfying an interface. For example, you can't rename `streamHeap.Pop` to `streamHeap.managedPop`.

So I'm not really sure how to handle these helper objects. I don't think it's *wrong* to call them static, because it's true that the parent doesn't directly modify the pointer itself. But I also feel like it's potentially confusing, because the nature of the helper object isn't "static" at all; it has lots of internal state that changes frequently.

ChrisSchinnerl: Yes, exactly. It's probably not a problem for 'real' dependencies from different packages since you can only access them via exported methods anyway (and you are not allowed to use those while holding a lock). 
Otherwise I think calling it `static` is a good enough solution for now.

node/api/client/renter.go
```diff
@@ -118,6 +118,15 @@ func (c *Client) RenterRenamePost(siaPathOld, siaPathNew string) (err error) {
 	return
 }
 
+// RenterSetStreamCacheSizePost uses the /renter endpoint to change the renter's
+// streamCacheSize for streaming
+func (c *Client) RenterSetStreamCacheSizePost(cacheSize uint64) (err error) {
+	values := url.Values{}
+	values.Set("streamcachesize", strconv.FormatUint(cacheSize, 10))
```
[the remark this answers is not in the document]
lukechampine: the size is the number of chunks, not bytes; is there any reason it should be a power of 2?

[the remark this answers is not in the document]
lukechampine: `10` is the base of the number when formatted by `FormatUint`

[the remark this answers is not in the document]
DavidVorick: ahhh. Yeah I definitely misinterpreted that.

PR #3011, opened 2018-05-10
address contractor ndf; bump travis to go 1.10

DavidVorick:


.travis.yml
```diff
@@ -4,7 +4,7 @@ os:
   - linux
 
 go:
-  - 1.9
+  - 1.10
```
ChrisSchinnerl: I don't think you can do this. Iirc @lukechampine once said that this resolves to Go 1.1 instead of 1.10.

PR #3014, opened 2018-05-11
Fix download code memory leak

ChrisSchinnerl:


modules/miner/miner_test.go
```diff
@@ -326,6 +326,10 @@ func TestIntegrationStartupRescan(t *testing.T) {
 // TestMinerCloseDeadlock checks that the miner can cleanly close even if the
 // CPU miner is running.
 func TestMinerCloseDeadlock(t *testing.T) {
+	if testing.Short() {
+		// Can't run this as a short test since we sleep 5s.
```
DavidVorick: can't run anything that calls `createMinerTester` as a short test - nothing that touches disk I/O is allowed to be a short test, sometimes the CI will block on a disk read or write for longer than 5s

DavidVorick:
looks good to me

PR #3015, opened 2018-05-11
i3009 renter settings

### LINKED ISSUE #3009: Initialize missing fields in `renter.SetSettings`
In `modules/renter/renter.go` in the `Settings` method we currently don't initialize all the fields of the `RenterSettings` (ratelimit fields are missing):

https://github.com/NebulousLabs/Sia/blob/7a92bcca28e3c6e31a9ea2afb989dbf53e3e1735/modules/renter/renter.go#L344-L349

This causes a call to the `/renter` endpoint with a new set of settings (e.g. allowance change) to always reset the bandwidth.

MSevey:
resolves #3009 
upload and download speeds were being reset to defaults as they were not being returned when Settings() was called on the renter

### DIFF
--- a/doc/api/Renter.md
+++ b/doc/api/Renter.md
@@ -64,14 +64,15 @@ returns the current settings along with metrics on the renter's spending.
     }, 
     // MaxUploadSpeed by defaul is unlimited but can be set by the user to 
     // manage bandwidth
-    "maxuploadspeed":     1234, // BPS
+    "maxuploadspeed":     1234, // bytes per second
 
     // MaxDownloadSpeed by defaul is unlimited but can be set by the user to 
     // manage bandwidth
-    "maxdownloadspeed":   1234, // BPS
+    "maxdownloadspeed":   1234, // bytes per second
 
-    // The DownloadCacheSize is 
-    "downloadcachesize":  4   
+    // The DownloadCacheSize is the number of data chunks that will be cached during
+    // streaming
+    "downloadcachesize":  4  
   },
 
   // Metrics about how much the Renter has spent on storage, uploads, and
--- a/doc/api/Renter.md
+++ b/doc/api/Renter.md
@@ -61,7 +61,17 @@ returns the current settings along with metrics on the renter's spending.
       // contract is scheduled to end, the contract is renewed automatically.
       // Is always nonzero.
       "renewwindow": 3024 // blocks
-    }
+    }, 
+    // MaxUploadSpeed by defaul is unlimited but can be set by the user to 
+    // manage bandwidth
+    "maxuploadspeed":     1234, // BPS
+
+    // MaxDownloadSpeed by defaul is unlimited but can be set by the user to 
+    // manage bandwidth
+    "maxdownloadspeed":   1234, // BPS
+
+    // The DownloadCacheSize is 
+    "downloadcachesize":  4   
   },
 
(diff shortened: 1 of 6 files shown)


tbenz9:
Hey @MSevey, you may already know this but GitHub uses keywords to manage linked issues.  You might consider saying something like `resolves #3009` instead of `address issue #3009` so that the issue gets closed automatically when this PR gets merged in.  GitHub keywords work in titles and descriptions so it's best practice to title issues with `resolves #3009 renter settings` instead of `i3009 renter settings`.

Of course, you need to be careful if you don't want the PR to close the issue, in which case `addresses issue #3009` would be perfect since `addresses` is not a GitHub keyword.

Check out the link below for a full list of keywords: https://help.github.com/articles/closing-issues-using-keywords/

Keep up the good work, the community is loving the PRs you're working on!

MSevey:
@tbenz9 thanks for the info, I'll check out that link!

MSevey:
Build fail due to pending PR on NebulousLabs/ratelimit

doc/api/Renter.md
```diff
@@ -61,7 +61,17 @@ returns the current settings along with metrics on the renter's spending.
...
+    // MaxUploadSpeed by defaul is unlimited but can be set by the user to 
+    // manage bandwidth
+    "maxuploadspeed":     1234, // BPS
+
+    // MaxDownloadSpeed by defaul is unlimited but can be set by the user to 
+    // manage bandwidth
+    "maxdownloadspeed":   1234, // BPS
+
+    // The DownloadCacheSize is 
+    "downloadcachesize":  4
```
lukechampine: incomplete docstring

also I would use `bytes per second` instead of `BPS` for clarity

modules/renter/contractor/contractor.go
```diff
@@ -157,6 +157,13 @@ func (c *Contractor) ResolveID(id types.FileContractID) types.FileContractID {
 	return newID
 }
 
+// RateLimits sets the bandwidth limits for connections created by the
+// contractSet.
+func (c *Contractor) RateLimits() (int64, int64, uint64) {
+	download, upload, packetSize := c.staticContracts.RateLimits()
+	return download, upload, packetSize
+}
```
lukechampine: this can be shortened to `return c.staticContracts.RateLimits()`.

modules/renter/proto/contractset.go
```diff
@@ -104,6 +104,13 @@ func (cs *ContractSet) Return(c *SafeContract) {
 	safeContract.mu.Unlock()
 }
 
+// RateLimits sets the bandwidth limits for connections created by the
+// contractSet.
+func (cs *ContractSet) RateLimits() (int64, int64, uint64) {
+	download, upload, packetSize := cs.rl.Limits()
+	return download, upload, packetSize
```
lukechampine: ditto

modules/renter/renter.go
```diff
@@ -133,6 +133,10 @@ type hostContractor interface {
 	// ResolveID returns the most recent renewal of the specified ID.
 	ResolveID(types.FileContractID) types.FileContractID
 
+	// RateLimits Gets the bandwidth limits for connections created by the
+	// contractor and its submodules.
+	RateLimits() (int64, int64, uint64)
```
lukechampine: these return values should be named so that it's clear which limit is which

PR #3023, opened 2018-05-14
Resolve #3022 

### LINKED ISSUE #3022: Deadlock on single file API request with wrong file name
From discord user drexel.

i'm testing the new /renter/file endpoint and appears to be a deadlock on file not found. shouldn't there be an `r.mu.RUnlock(lockID)` before: https://github.com/MSevey/Sia/blob/9360c8bca82d7c143e54e23ac2f7aa824ade06e9/modules/renter/files.go#L324

MSevey:
Resolve issue #3022 

Updated unlock statements in File() in /modules/renter/files.go to avoid deadlocks when the wrong filename is used.


DavidVorick:
@MSevey can you be more descriptive in your pull request titles, perhaps:

"Resolve #3022 - fix deadlocks in File() method"

PR #3028, opened 2018-05-17
Add id fields to /consensus/blocks endpoint

ChrisSchinnerl:
This PR adds various `ID` fields to the JSON block returned by the `/consensus/blocks/...` endpoint.


lukechampine:
ok, LGTM. Need to update the API docs

DavidVorick:
Need to extend the /consensus/blocks test to check that the id fields are all being set correctly.

PR #3029, opened 2018-05-17
Replace contract if renew fails multiple times

ChrisSchinnerl:
Previously a contract that was `goodForRenew` would just continue being renewed until it is either successfully renewed or expired. This PR changes that behavior so that a contract will be marked as `!goodForRenew` as soon as it has failed `12` times and reached the second half of the renew window.

This means that a contract that has already failed `>=12` times in the first half of the renew window will be marked immediately after entering the second half of the window.
A contract that failed its renewal within the second half of the renew window will still have to fail 12 times before being marked.
A contract that doesn't have enough time left in the second half of the renew window to fail 12 times will simply expire without being marked.


modules/renter/contractor/contracts.go
```diff
@@ -502,9 +518,38 @@ func (c *Contractor) threadedContractMaintenance() {
...
+				c.mu.RUnlock()
+				secondHalfOfWindow := blockHeight+allowance.RenewWindow/2 >= md.EndHeight
+				replace := blockHeight-firstRenew >= consecutiveRenewalsBeforeReplacement
+				if failedBefore && secondHalfOfWindow && replace {
+					oldUtility.GoodForRenew = false
+					err := oldContract.UpdateUtility(oldUtility)
+					if err != nil {
+						c.log.Println("WARN: failed to mark contract as !goodForRenew:", err)
+					}
+					c.log.Printf("WARN: failed to renew %v, marked as bad: %v\n", id, err)
```
lukechampine: this reports `err` twice

modules/renter/contractor/consts.go
```diff
@@ -20,6 +20,14 @@ var (
 		Standard: 10,
 		Testing:  1,
 	}).(int)
+
+	// consecutiveRenewalsBeforeReplacement is the number of times a contract
+	// attempt to be renewed before it is marked as !goodForRenew.
+	consecutiveRenewalsBeforeReplacement = build.Select(build.Var{
+		Dev:      types.BlockHeight(12),
+		Standard: types.BlockHeight(12), // ~2h
+		Testing:  types.BlockHeight(12),
```
DavidVorick: Is 12 too high for the testing environments? A lot of our tests don't even have renew windows that long.

modules/renter/contractor/contracts.go
```diff
@@ -502,10 +518,39 @@ func (c *Contractor) threadedContractMaintenance() {
...
+			// to false.
+			newContract, errRenew := c.managedRenew(oldContract, amount, endHeight)
+			if errRenew != nil {
+				// Check if contract has to be replaced.
+				md := oldContract.Metadata()
+				c.mu.RLock()
+				firstRenew, failedBefore := c.firstFailedRenew[md.ID]
+				c.mu.RUnlock()
+				secondHalfOfWindow := blockHeight+allowance.RenewWindow/2 >= md.EndHeight
+				replace := blockHeight-firstRenew >= consecutiveRenewalsBeforeReplacement
```
DavidVorick: This doesn't mean that you've tried 12 times, it just means that 12 blocks have passed. That might have issues if you come online and start syncing, though I know we don't run this code if we aren't already sync'd.

I think it's probably less fragile though to make the check 12 discreet tries instead of 12 blocks passed since the first failed attempt.

DavidVorick: It feels less fragile to me because it eliminates the number of external factors that can interfere with the logic here.

modules/renter/contractor/consts.go
ChrisSchinnerl: I thought 12 is ok for testing since we probably don't want that to happen by accident since it might hide bugs in tests that don't explicitly want a failed renew to be replaced.

modules/host/negotiateformcontract.go
```diff
@@ -42,13 +42,9 @@ func (h *Host) managedAddCollateral(settings modules.HostExternalSettings, txnSe
 	if err != nil {
 		return
 	}
-	defer func() {
-		if err != nil {
-			builder.Drop()
-		}
-	}()
```
DavidVorick: I don't think that I agree with removing this defer. Even though logically the two things are the same, one is robust to the function being modified, and the other is not. The defers are nice because you know it'll be correct even if the function changes in the future.

modules/host/negotiaterenewcontract.go
```diff
@@ -55,13 +55,9 @@ func (h *Host) managedAddRenewCollateral(so storageObligation, settings modules.
 	if err != nil {
 		return
 	}
-	defer func() {
-		if err != nil {
-			builder.Drop()
-		}
-	}()
```
DavidVorick: Same here

siatest/renter/renter_test.go
```diff
@@ -354,6 +357,100 @@ func testRenterStreamingCache(t *testing.T, tg *siatest.TestGroup) {
...
+				notGoodForRenew++
+			}
+		}
+		if notGoodForRenew != 1 {
+			err := fmt.Errorf("there should be exactly 1 contract that is !goodForRenew but was %v",
+				notGoodForRenew)
+			return errors.Compose(miner.MineBlock(), err)
+		}
+		return nil
+	})
```
DavidVorick: would like to see this test extended.

I think it would be fine to start the group with 3 hosts, and have the renter settings established so that it only uses 2 of them. Then before the host gets realized as a bad host, there should only be 2 contracts and both goodForRenew. At some point one of them should be seen as !goodForRenew, and then at some point after that there should be 3 contracts, where 2 are goodForRenew and one is !goddForRenew

siatest/renter/renter_test.go
```diff
@@ -354,6 +357,100 @@ func testRenterStreamingCache(t *testing.T, tg *siatest.TestGroup) {
...
+	}
+	for _, c := range rcg.Contracts {
+		if !c.GoodForRenew {
+			t.Fatal("renter got a contract that is !goodForRenew")
+		}
+	}
+
+	// We should be within the renew window now. Somewhere between now and the
+	// contract's EndHeight one of the contracts should be marked as
+	// !goodForUpload.
```
DavidVorick: marked as !goodForRenew

Issue #3027, opened 2018-05-17
Wallet module is out of sync with consensus module


huetsch:
# BUG REPORT

**Expected Behavior**

Wallet blockheight should be in sync with consensus blockheight.

**How to reproduce it (as minimally and precisely as possible)**

Set up a new Sia environment. Sync consensus. Init a new wallet seed. Then run: `siac consensus` and `siac wallet` and compare the respective heights. Wallet is consistently behind consensus (usually by 10 blocks, but apparently not always). 

<img width="279" alt="consensus mismatch" src="https://user-images.githubusercontent.com/121910/40151416-cdf61e4c-59ba-11e8-9113-cdf1377339b8.png">

**Extra Comments**

Not sure why `keyConsensusHeight` in the wallet module is falling behind, my best guess is that there's some buffering going on or there's some issue with boltdb.

OKex is using the consensus blockheight in conjunction with the wallet API to grab info about incoming transactions, but their call to the `/wallet/transactions` API is throwing an error because, although these blocks are in consensus, the wallet API correctly won't accept a starting blockheight that it views as too high.

**Environment**
* Sia version: master
* OS: MacOS 10, Centos 7

DavidVorick:
We have reproduced this (inconsistently) locally, so at the very least we agree there's a problem.

My best guess right now is that we are somehow getting into a situation where we revert a consensus change due to an error, but then for some reason we don't revert any changes to the height. On the other hand, it doesn't seem like we are hitting any condition where the wallet shuts down or panics.

@huetsch can you confirm that you didn't experience any crashes, shutdowns, or wallet restarts (it won't start on it's own, you'd have to run a command to start it) while reproducing this bug? If not, that will narrow down some of the places that we have to look.

huetsch:
No crashes. I may have killed the Sia daemon (SIGINT) while syncing and then restarted it. But other than that, it was a fresh environment.

shadymeh:
Will it be post here or on discord if we are close to fixing the bug? Would be nice to see okex add sia.

nielscastien:
@DavidVorick I still have this issue on my instance (v1.3.3 updated from v1.3.2). No crashes or other problems but the wallet is still 8 blocks ahead:
![afbeelding](https://user-images.githubusercontent.com/22688069/41291325-b24153ca-6e4f-11e8-96e1-9e926c599bdf.png)

PR #3030, opened 2018-05-17
Fix wallet out-of-sync bug

ChrisSchinnerl:
This PR should fix [#3027](https://github.com/NebulousLabs/Sia/issues/3027).


modules/consensus/consensusset.go
```diff
@@ -12,7 +12,7 @@ import (
 	"github.com/NebulousLabs/Sia/encoding"
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/persist"
-	"github.com/NebulousLabs/Sia/sync"
+	siasync "github.com/NebulousLabs/Sia/sync"
```
DavidVorick: can probably switch this to the more refined NebulousLabs/threadgroup

DavidVorick: siasync.ThreadGroup should be considered deprecated

modules/consensus/subscribe.go
```diff
@@ -165,16 +168,23 @@ func (cs *ConsensusSet) managedInitializeSubscribe(subscriber modules.ConsensusS
 	})
 	cs.mu.RUnlock()
 	if err != nil {
-		return err
+		return modules.ConsensusChangeID{}, err
+	}
+
+	// Nothing to do if the changeEntry doesn't exist.
+	if !exists {
+		return start, nil
```
DavidVorick: is this code block reachable? Looks to me like it returns inside the loop if !exists?

modules/consensus/consensusset.go
ChrisSchinnerl: Unfortunately we use `tg.Flush` which is not available in the new package.

modules/consensus/subscribe.go
ChrisSchinnerl: I think so. It's not part of the loop. In line `165` we set `entry, exists = entry.NextEntry(tx)` to get the first entry that we have to send to the new subscriber. If that new entry doesn't exist we return early.

DavidVorick: ah I see

modules/consensus/subscribe.go
```diff
@@ -165,16 +168,23 @@ func (cs *ConsensusSet) managedInitializeSubscribe(subscriber modules.ConsensusS
...
 	}
 
 	// Send all remaining consensus changes to the subscriber.
+	var latestChangeID modules.ConsensusChangeID
 	for exists {
```
DavidVorick: at this point, we've already established that `exists` equals true, so we can set the latestChangeID to = `entry.ID()` without needing to worry about correctness. That eliminates any code where we haven't updated the latest change id yet, reducing surface area for mistakes.

modules/consensus/accept.go
```diff
@@ -278,10 +277,11 @@ func (cs *ConsensusSet) managedAcceptBlocks(blocks []types.Block) (blockchainExt
 			if err != nil {
 				return err
 			}
-			// Sanity check - If reverted blocks is zero, applied blocks should also
-			// be zero.
-			if build.DEBUG && len(changeEntry.AppliedBlocks) == 0 && len(changeEntry.RevertedBlocks) != 0 {
-				panic("after adding a change entry, there are no applied blocks but there are reverted blocks")
+			// Sanity check - we should never apply less blocks than we revert.
```
lukechampine: fewer

PR #3039, opened 2018-05-22
Make isOffline more tolerant to sudden connectivity issues

ChrisSchinnerl:


DavidVorick:
Is there some way that we can write a test to simulate brief connectivity issues? I guess we don't have the mocking in place, but this definitely seems like a corner where we'd want to have a test that can insert that sort of chaos.

We should have environments (maybe ant-farms) that can simulate these types of outages, and other common issues like constantly switching between VMs and stuff.

PR #3040, opened 2018-05-22
Calculate spending from contracts renewed during current period

### LINKED ISSUE #3037: BUG: siac renter shows incorrect wallet informaiton
The siac renter command shows available funds when the siac wallet command shows a zero balance

MSevey:
`siac renter` was displaying information about funds that was not matching the wallet.  There were comments in the code about `PreviousContracts` needing to be accounted for.  Fixed bug with using `oldContracts` as there were not enough comments to make it clear what the intention of the `PreviousContracts` were that was different that using the `oldContracts`.

@lukechampine do you remember why we would want to create `PreviousContracts` vs using `oldContracts`?

Resolves #3037

### DIFF
--- a/modules/renter/contractor/contractor.go
+++ b/modules/renter/contractor/contractor.go
@@ -107,7 +107,13 @@ func (c *Contractor) PeriodSpending() modules.ContractorSpending {
 	// Calculate spending from contracts that were renewed during the current period
 	for _, old := range c.oldContracts {
 		if old.EndHeight >= c.currentPeriod {
+			// Calculate ContractFees
+			spending.ContractFees = spending.ContractFees.Add(old.ContractFee)
+			spending.ContractFees = spending.ContractFees.Add(old.TxnFee)
+			spending.ContractFees = spending.ContractFees.Add(old.SiafundFee)
+			// Calculate TotalAllocated
 			spending.TotalAllocated = spending.TotalAllocated.Add(old.TotalCost)
+			// Calculate Spending
 			spending.DownloadSpending = spending.DownloadSpending.Add(old.DownloadSpending)
 			spending.UploadSpending = spending.UploadSpending.Add(old.UploadSpending)
 			spending.StorageSpending = spending.StorageSpending.Add(old.StorageSpending)
--- a/modules/renter/contractor/contractor.go
+++ b/modules/renter/contractor/contractor.go
@@ -102,13 +102,16 @@ func (c *Contractor) PeriodSpending() modules.ContractorSpending {
 		spending.DownloadSpending = spending.DownloadSpending.Add(contract.DownloadSpending)
 		spending.UploadSpending = spending.UploadSpending.Add(contract.UploadSpending)
 		spending.StorageSpending = spending.StorageSpending.Add(contract.StorageSpending)
-		// TODO: fix PreviousContracts
-		// for _, pre := range contract.PreviousContracts {
-		// 	spending.ContractSpending = spending.ContractSpending.Add(pre.TotalCost)
-		// 	spending.DownloadSpending = spending.DownloadSpending.Add(pre.DownloadSpending)
-		// 	spending.UploadSpending = spending.UploadSpending.Add(pre.UploadSpending)
-		// 	spending.StorageSpending = spending.StorageSpending.Add(pre.StorageSpending)
-		// }
+	}
+
+	// Calculate spending from contracts that were renewed during the current period
+	for _, old := range c.oldContracts {
+		if old.EndHeight >= c.currentPeriod {
+			spending.TotalAllocated = spending.TotalAllocated.Add(old.TotalCost)
+			spending.DownloadSpending = spending.DownloadSpending.Add(old.DownloadSpending)
+			spending.UploadSpending = spending.UploadSpending.Add(old.UploadSpending)
+			spending.StorageSpending = spending.StorageSpending.Add(old.StorageSpending)
+		}
 	}
 	// Calculate amount of spent money to get unspent money.
 	allSpending := spending.ContractFees
--- a/modules/renter/contractor/contractor.go
+++ b/modules/renter/contractor/contractor.go
@@ -106,7 +106,7 @@ func (c *Contractor) PeriodSpending() modules.ContractorSpending {
 
 	// Calculate spending from contracts that were renewed during the current period
 	for _, old := range c.oldContracts {
-		if old.EndHeight >= c.currentPeriod {
+		if old.StartHeight >= c.currentPeriod {
 			// Calculate ContractFees
 			spending.ContractFees = spending.ContractFees.Add(old.ContractFee)
 			spending.ContractFees = spending.ContractFees.Add(old.TxnFee)
--- a/siatest/renter/renter_test.go
+++ b/siatest/renter/renter_test.go
@@ -54,7 +54,6 @@ func TestRenter(t *testing.T) {
 		{"TestRenterDownloadAfterRenew", testRenterDownloadAfterRenew},
 		{"TestRenterLocalRepair", testRenterLocalRepair},
 		{"TestRenterRemoteRepair", testRenterRemoteRepair},
-		{"TestRenterSpendingReporting", testRenterSpendingReporting},
 	}
 	// Run subtests
 	for _, subtest := range subTests {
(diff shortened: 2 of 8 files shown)


modules/renter/contractor/contractor.go
```diff
@@ -101,13 +101,16 @@ func (c *Contractor) PeriodSpending() modules.ContractorSpending {
...
-		// }
+	}
+
+	// Calculate spending from contracts that were renewed during the current period
+	for _, old := range c.oldContracts {
+		if old.EndHeight >= c.currentPeriod {
+			spending.TotalAllocated = spending.TotalAllocated.Add(old.TotalCost)
+			spending.DownloadSpending = spending.DownloadSpending.Add(old.DownloadSpending)
+			spending.UploadSpending = spending.UploadSpending.Add(old.UploadSpending)
+			spending.StorageSpending = spending.StorageSpending.Add(old.StorageSpending)
```
ChrisSchinnerl: I think we are missing 3 lines here for the fees:
```
spending.ContractFees = spending.ContractFees.Add(old.ContractFee)
spending.ContractFees = spending.ContractFees.Add(old.TxnFee)
spending.ContractFees = spending.ContractFees.Add(old.SiafundFee)
```

lukechampine:
`PreviousContracts` was a way to link a renewed contact to its "ancestors." I don't recall exactly how this was supposed to play into metrics reporting. Seems like `oldContracts` is sufficient.

DavidVorick:
We now need to extend the testing in the `siatest` package to verify that the metrics reporting is correct and adds up to the expected values even as contracts get renewed due to the initial contract spending being exhausted.

siatest/renter/renter_test.go
```diff
@@ -569,3 +570,148 @@ func testRenterDownloadAfterRenew(t *testing.T, tg *siatest.TestGroup) {
...
+	}
+	mwg, err := m.WalletGet()
+	if err != nil {
+		t.Fatal(err)
+	}
+	rwag, err := r.WalletAddressGet()
+	if err != nil {
+		t.Fatal(err)
+	}
+	_, err = m.WalletSiacoinsPost(mwg.ConfirmedSiacoinBalance.Div64(2), rwag.Address)
```
DavidVorick: It's not a blocking thing, but the excessive use of short/truncated variable names in this test are impacting the readability. Generally we try to keep it to only a few truncated names per function. The real issue here isn't any particular use, it's just that altogether there are too many.

A good rule of thumb is that the further away from creation/instantiation that you use a variable, the longer / more complete the variable name should be. A similar rule of thumb is that you shouldn't have more than 1 or 2 truncated variables being used in the same scope. That's all just to improve readability.

siatest/renter/renter_test.go
```diff
@@ -569,3 +570,148 @@ func testRenterDownloadAfterRenew(t *testing.T, tg *siatest.TestGroup) {
...
+		t.Fatal("Failed to get RenterGet:", err)
+	}
+	wg, err = r.WalletGet()
+	if err != nil {
+		t.Fatal("Failed to get wallet:", err)
+	}
+
+	// Setting variables to easier reference
+	fm := rg.FinancialMetrics
+	// allowance := rg.Settings.Allowance
```
DavidVorick: this line is commented out, should it be brought back or deleted?

MSevey:
@DavidVorick @ChrisSchinnerl I had to move creating the testgroup to within `TestRenterSpendingReporting`.  When I used the test group from `TestRenter` it would pass when I ran the test by itself but it would fail when `test-long` ran.  All the spending amounts would no longer match, I couldn't think of what the edge case might be that would cause that.

siatest/renter/renter_test.go
```diff
@@ -744,3 +744,247 @@ func testRenterDownloadAfterRenew(t *testing.T, tg *siatest.TestGroup) {
...
+	r := tg.Renters()[0]
+	wg, err := r.WalletGet()
+	if err != nil {
+		t.Fatal("Failed to get wallet:", err)
+	}
+	initialBalance := wg.ConfirmedSiacoinBalance
+
+	// Set allowance
+	if err = tg.SetRenterAllowance(r, siatest.DefaultAllowance); err != nil {
+		t.Fatal("Failed to set renter:", err)
```
MSevey: Should say "Failed to set renter allowance" 
will update

siatest/renter/renter_test.go
```diff
@@ -744,3 +744,247 @@ func testRenterDownloadAfterRenew(t *testing.T, tg *siatest.TestGroup) {
...
+	fm := rg.FinancialMetrics
+
+	// Check balance after allowance is set
+	wg, err = r.WalletGet()
+	if err != nil {
+		t.Fatal("Failed to get wallet:", err)
+	}
+	balanceAfterSetAllowance := initialBalance.Sub(fm.TotalAllocated)
+	if balanceAfterSetAllowance.Cmp(wg.ConfirmedSiacoinBalance) != 0 {
+		t.Fatalf("Renter Reported Spending does not equal wallet confirmed balance, \n%v != \n%v", balanceAfterSetAllowance, wg.ConfirmedSiacoinBalance)
```
MSevey: Break into two lines for cleaner readability

siatest/renter/renter_test.go
```diff
@@ -744,3 +744,247 @@ func testRenterDownloadAfterRenew(t *testing.T, tg *siatest.TestGroup) {
...
+
+	// Check spending after contract renewal
+	wg, err = r.WalletGet()
+	if err != nil {
+		t.Fatal("Failed to get wallet:", err)
+	}
+	rg, err = r.RenterGet()
+	if err != nil {
+		t.Fatal("Failed to get RenterGet:", err)
+	}
```
MSevey: Remove, duplicated

siatest/renter/renter_test.go
```diff
@@ -744,3 +744,247 @@ func testRenterDownloadAfterRenew(t *testing.T, tg *siatest.TestGroup) {
...
+
+	wg, err = r.WalletGet()
+	if err != nil {
+		t.Fatal("Failed to get wallet:", err)
+	}
+
+	// Check that renter financial metrics add up to allowance
+	if total.Cmp(allowance.Funds) != 0 {
+		t.Fatalf("Combined Total of reported spending and unspent funds not equal to allowance, \n%v != \n%v", total, allowance.Funds)
+	}
```
MSevey: Break into two lines

siatest/renter/renter_test.go
```diff
@@ -744,3 +744,247 @@ func testRenterDownloadAfterRenew(t *testing.T, tg *siatest.TestGroup) {
...
+
+	// Check renter financial metrics against contracts contract spending
+	var spending modules.ContractorSpending
```
MSevey: contracts contract, remove one

siatest/renter/renter_test.go
```diff
@@ -744,3 +744,247 @@ func testRenterDownloadAfterRenew(t *testing.T, tg *siatest.TestGroup) {
...
+		// Calculate Spending
+		spending.DownloadSpending = spending.DownloadSpending.Add(contract.DownloadSpending)
+		spending.UploadSpending = spending.UploadSpending.Add(contract.UploadSpending)
+		spending.StorageSpending = spending.StorageSpending.Add(contract.StorageSpending)
+	}
+
+	// Compare contract fees
+	if fm.ContractFees.Cmp(spending.ContractFees) != 0 {
+		t.Fatalf("Financial Metrics Contract Fees not equal to Renter Contract Fees, \n%v != \n%v", fm.ContractFees, spending.ContractFees)
+	}
```
MSevey: break into two lines

siatest/renter/renter_test.go
```diff
@@ -744,3 +744,247 @@ func testRenterDownloadAfterRenew(t *testing.T, tg *siatest.TestGroup) {
...
+	// Compare Total Allocated
+	if fm.TotalAllocated.Cmp(spending.TotalAllocated) != 0 {
+		t.Fatalf("Financial Metrics Total Allocated not equal to Renter Total Allocated, \n%v != \n%v", fm.TotalAllocated, spending.TotalAllocated)
```
MSevey: break into two lines

siatest/renter/renter_test.go
```diff
@@ -744,3 +744,247 @@ func testRenterDownloadAfterRenew(t *testing.T, tg *siatest.TestGroup) {
...
+	}
+	// Compare Spending
+	allSpending := spending.ContractFees.Add(spending.UploadSpending).
+		Add(spending.DownloadSpending).Add(spending.StorageSpending)
+	if totalSpent.Cmp(allSpending) != 0 {
+		t.Fatalf("Financial Metrics Spending not equal to Renter Spending, \n%v != \n%v", totalSpent, allSpending)
```
MSevey: break into two lines

Issue #3013, opened 2018-05-11
1.3.3 RC2: "renter allowance cancel" destroys all renter contracts and data


mtlynch:
# BUG REPORT

**How to reproduce it (as minimally and precisely as possible)**

1. Create renter contracts with `siac renter setallowance`
1. Disable contract renewal with `siac renter allowance cancel`

**Expected Behavior**

Sia maintains existing contracts, but does not automatically renew them.

**Actual Behavior**

Sia destroys all contracts and loses all data.

**Environment**: 
* Sia version: Sia 1.3.3 RC2
* OS: Ubuntu 16 (Docker container)

**Logs**

* [Before allowance cancel](https://gist.github.com/mtlynch/b8cbb08ff6bc22d88bd528ffbc211dd2)
* [After allowance cancel](https://gist.github.com/mtlynch/d13988ef5a9189bb59aee72ab7a4d0f9)

tbenz9:
@lukechampine can you confirm that this is the expected behavior?

DavidVorick:
This is intended behavior, though there was some internal disagreement when it was implemented over whether it was the right way to design the api call.

I'm in favor of changing the call so that instead of destroying all of the contracts, it just collapses the allowance and stops renewing.

lukechampine:
that seems much more reasonable to me. Shouldn't be too difficult to implement.

PR #3049, opened 2018-05-25
add allowance cancel warning

tbenz9:
Add warning to allowance cancel.

This PR replaces #3021 since it appears that dev has lost interest.
This PR resolves #3017 and is a short-term fix for #3013 until a core dev can update the renter allowance API endpoint as discussed in #3013.

PR #3051, opened 2018-05-25
Cancel allowance shouldn't delete contracts

ChrisSchinnerl:
Fixes [#3013](https://github.com/NebulousLabs/Sia/issues/3013)

### DIFF
--- a/cmd/siac/rentercmd.go
+++ b/cmd/siac/rentercmd.go
@@ -274,8 +274,8 @@ func renterallowancecmd() {
 // renterallowancecancelcmd cancels the current allowance.
 func renterallowancecancelcmd() {
 	fmt.Println(`Canceling your allowance will disable uploading new files,
-	repairing existing files, and renewing existing files. All files will cease
-	to be accessible after a short period of time.`)
+repairing existing files, and renewing existing files. All files will cease
+to be accessible after a short period of time.`)
 again:
 	fmt.Print("Do you want to continue? [y/n] ")
 	var resp string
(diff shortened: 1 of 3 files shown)


tbenz9:
Wow, that was fast.  Seems a little silly but if this changes the behavior to keep existing contracts active you should probably remove the warning I just put in with #3049.

cmd/siac/rentercmd.go
```diff
@@ -273,7 +273,8 @@ func renterallowancecmd() {
 
 // renterallowancecancelcmd cancels the current allowance.
 func renterallowancecancelcmd() {
-	fmt.Println("Canceling your allowance will destroy all contracts and uploaded files immediately.")
+	fmt.Println(`Canceling your allowance will prevent the renewal of your contracts and you will no longer be able to upload new files.
+	Already uploaded files can still be downloaded until the contracts expire and/or the file's redundancy drops below 1x.`)
```
lukechampine: needs to be wrapped to 80 chars

DavidVorick:
We got lucky with the implementation here. If we had made a few different design choices earlier, this would have been more difficult to implement.

As it were, we do try to program such that changing things later is easy, but in this case I think we really did just get lucky.

cmd/siac/rentercmd.go
DavidVorick: "Canceling your allowance will disable uploading new files, repairing existing files, and renewing existing files. All files will cease to be accessible after a short period of time."

DavidVorick:
Other than the user-facing message, this looks good, including the test.

PR #3052, opened 2018-05-25
Persist renter cache settings, and renter bandwidth limiter settings

### LINKED ISSUE #3043: Persist Renter Settings
Currently the renter stream cache size and the renter memory bandwidth limits are not persisted, I believe that most users will prefer to have these values persisted.

Before we release v1.3.3 and get into compatibility issues, we should patch this up. Giving it to @MSevey

MSevey:
This resolves #3043

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -904,13 +904,13 @@ modify settings that control the renter's behavior.
 
 ###### Query String Parameters [(with comments)](/doc/api/Renter.md#query-string-parameters)
 ```
-funds // hastings
+funds             // hastings
 hosts
-period      // block height
-renewwindow // block height
+period            // block height
+renewwindow       // block height
 maxdownloadspeed  // bytes per second
-maxuploadspeed  // bytes per second
-streamcachesize // number of data chunks cached when streaming
+maxuploadspeed    // bytes per second
+streamcachesize   // number of data chunks cached when streaming
 ```
 
 ###### Response
--- a/doc/API.md
+++ b/doc/API.md
@@ -908,9 +908,9 @@ funds // hastings
 hosts
 period      // block height
 renewwindow // block height
-maxdownloadspeed  // bytes per second, not persisted and will be reset by a shutdown
-maxuploadspeed  // bytes per second, not persisted and will be reset by a shutdown
-streamcachesize // number of data chunks cached when streaming, not persisted and will be reset by a shutdown
+maxdownloadspeed  // bytes per second
+maxuploadspeed  // bytes per second
+streamcachesize // number of data chunks cached when streaming
 ```
 
 ###### Response
--- a/modules/renter/persist.go
+++ b/modules/renter/persist.go
@@ -207,11 +207,12 @@ func (r *Renter) saveSync() error {
 	return persist.SaveJSON(settingsMetadata, r.persist, filepath.Join(r.persistDir, PersistFilename))
 }
 
-// load fetches the saved renter data from disk.
-func (r *Renter) load() error {
+// loadSiaFiles walks through the directory searching for siafiles and loading
+// them into memory.
+func (r *Renter) loadSiaFiles() error {
 	// Recursively load all files found in renter directory. Errors
 	// encountered during loading are logged, but are not considered fatal.
-	err := filepath.Walk(r.persistDir, func(path string, info os.FileInfo, err error) error {
+	return filepath.Walk(r.persistDir, func(path string, info os.FileInfo, err error) error {
 		// This error is non-nil if filepath.Walk couldn't stat a file or
 		// folder.
 		if err != nil {
@@ -240,26 +241,40 @@ func (r *Renter) load() error {
 		}
 		return nil
 	})
+}
+
+// load fetches the saved renter data from disk.
+func (r *Renter) loadSettings() error {
+	r.persist = persistence{
+		Tracking: make(map[string]trackedFile),
+	}
+	err := persist.LoadJSON(settingsMetadata, &r.persist, filepath.Join(r.persistDir, PersistFilename))
+	if os.IsNotExist(err) {
+		// No persistence yet, set the defaults and continue.
+		r.persist.MaxDownloadSpeed = DefaultMaxDownloadSpeed
+		r.persist.MaxUploadSpeed = DefaultMaxUploadSpeed
+		r.persist.StreamCacheSize = DefaultStreamCacheSize
+	} else if err == persist.ErrBadVersion {
+		// Outdated version, try the 040 to 133 upgrade.
+		err = r.updatePersistVersionFrom040To133()
+		if err != nil {
+			// Nothing left to try.
+			return err
+		}
+		// Re-load the settings now that the file has been upgraded.
+		return r.loadSettings()
+	}
+
+	// Set the bandwidth limits on the contractor, which was already initialized
+	// without bandwidth limits.
+	err = r.setBandwidthLimits(r.persist.MaxDownloadSpeed, r.persist.MaxUploadSpeed)
 	if err != nil {
 		return err
 	}
 
-	// Load contracts and entropy.
-	r.persist = persistence{
-		Tracking: make(map[string]trackedFile),
-	}
-	err = persist.LoadJSON(settingsMetadata, &r.persist, filepath.Join(r.persistDir, PersistFilename))
-	if err != nil {
-		err = r.updatePersistVersionFrom040To133()
-		if err != nil {
-			return err
-		}
-		if err = r.load(); err != nil {
-			return err
-		}
-	}
-
-	return nil
+	// Need to set the settings for the values that are harvested from disk at
+	// startup.
+	return r.saveSync()
 }
 
 // shareFiles writes the specified files to w. First a header is written,
@@ -428,11 +443,13 @@ func (r *Renter) initPersist() error {
 	}
 
 	// Load the prior persistence structures.
-	err = r.load()
-	if err != nil && !os.IsNotExist(err) {
+	err = r.loadSettings()
+	if err != nil {
 		return err
 	}
-	return nil
+
+	// Load the siafiles into memory.
+	return r.loadSiaFiles()
 }
 
 // LoadSharedFiles loads a .sia file into the renter. It returns the nicknames
@@ -470,5 +487,8 @@ func (r *Renter) updatePersistVersionFrom040To133() error {
 		return err
 	}
 	metadata.Version = persistVersion133
+	r.persist.MaxDownloadSpeed = DefaultMaxDownloadSpeed
+	r.persist.MaxUploadSpeed = DefaultMaxUploadSpeed
+	r.persist.StreamCacheSize = DefaultStreamCacheSize
 	return persist.SaveJSON(metadata, r.persist, filepath.Join(r.persistDir, PersistFilename))
 }
--- a/modules/renter/persist.go
+++ b/modules/renter/persist.go
@@ -42,6 +42,10 @@ var (
 
 	shareHeader  = [15]byte{'S', 'i', 'a', ' ', 'S', 'h', 'a', 'r', 'e', 'd', ' ', 'F', 'i', 'l', 'e'}
 	shareVersion = "0.4"
+
+	// Persist Version Numbers
+	persistVersion040 = "0.4"
+	persistVersion133 = "1.3.3"
 )
 
 type (
@@ -240,13 +244,19 @@ func (r *Renter) load() error {
 		return err
 	}
 
-	// Load contracts, repair set, and entropy.
+	// Load contracts and entropy.
 	r.persist = persistence{
 		Tracking: make(map[string]trackedFile),
 	}
 	err = persist.LoadJSON(settingsMetadata, &r.persist, filepath.Join(r.persistDir, PersistFilename))
 	if err != nil {
-		return err
+		err = r.updatePersistVersionFrom040To133()
+		if err != nil {
+			return err
+		}
+		if err = r.load(); err != nil {
+			return err
+		}
 	}
 
 	return nil
@@ -447,4 +457,14 @@ func (r *Renter) LoadSharedFilesASCII(asciiSia string) ([]string, error) {
 
 	dec := base64.NewDecoder(base64.URLEncoding, bytes.NewBufferString(asciiSia))
 	return r.loadSharedFiles(dec)
+}
+
+func (r *Renter) updatePersistVersionFrom040To133() error {
+	settingsMetadata.Version = persistVersion040
+	err := persist.LoadJSON(settingsMetadata, &r.persist, filepath.Join(r.persistDir, PersistFilename))
+	if err != nil {
+		return err
+	}
+	settingsMetadata.Version = persistVersion133
+	return persist.SaveJSON(settingsMetadata, r.persist, filepath.Join(r.persistDir, PersistFilename))
 }
--- a/modules/renter/persist.go
+++ b/modules/renter/persist.go
@@ -37,7 +37,7 @@ var (
 
 	settingsMetadata = persist.Metadata{
 		Header:  "Renter Persistence",
-		Version: currentPersistVersion,
+		Version: persistVersion,
 	}
 
 	shareHeader  = [15]byte{'S', 'i', 'a', ' ', 'S', 'h', 'a', 'r', 'e', 'd', ' ', 'F', 'i', 'l', 'e'}
--- a/modules/renter/persist.go
+++ b/modules/renter/persist.go
@@ -460,11 +460,15 @@ func (r *Renter) LoadSharedFilesASCII(asciiSia string) ([]string, error) {
 }
 
 func (r *Renter) updatePersistVersionFrom040To133() error {
-	settingsMetadata.Version = persistVersion040
-	err := persist.LoadJSON(settingsMetadata, &r.persist, filepath.Join(r.persistDir, PersistFilename))
+	metadata := persist.Metadata{
+		Header:  settingsMetadata.Header,
+		Version: persistVersion040,
+	}
+
+	err := persist.LoadJSON(metadata, &r.persist, filepath.Join(r.persistDir, PersistFilename))
 	if err != nil {
 		return err
 	}
-	settingsMetadata.Version = persistVersion133
-	return persist.SaveJSON(settingsMetadata, r.persist, filepath.Join(r.persistDir, PersistFilename))
+	metadata.Version = persistVersion133
+	return persist.SaveJSON(metadata, r.persist, filepath.Join(r.persistDir, PersistFilename))
 }
--- a/modules/renter/persist.go
+++ b/modules/renter/persist.go
@@ -254,9 +254,13 @@ func (r *Renter) loadSettings() error {
 		r.persist.MaxDownloadSpeed = DefaultMaxDownloadSpeed
 		r.persist.MaxUploadSpeed = DefaultMaxUploadSpeed
 		r.persist.StreamCacheSize = DefaultStreamCacheSize
+		err = r.saveSync()
+		if err != nil {
+			return err
+		}
 	} else if err == persist.ErrBadVersion {
 		// Outdated version, try the 040 to 133 upgrade.
-		err = r.updatePersistVersionFrom040To133()
+		err = convertPersistVersionFrom040To133(filepath.Join(r.persistDir, PersistFilename))
 		if err != nil {
 			// Nothing left to try.
 			return err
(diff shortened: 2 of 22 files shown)


MSevey:
Need to replicate persistence struct from miner for renter.  replace data struct in saveSync

modules/renter/persist.go
```diff
@@ -235,17 +242,14 @@ func (r *Renter) load() error {
 	}
 
 	// Load contracts, repair set, and entropy.
-	data := struct {
-		Tracking  map[string]trackedFile
-		Repairing map[string]string // COMPATv0.4.8
```
DavidVorick: We shouldn't need this anymore, anyone upgrading from v0.4.8 is not going to have working files at this point.

modules/renter/persist.go
```diff
@@ -44,6 +44,17 @@ var (
 	shareVersion = "0.4"
 )
 
+type (
+	// persist contains all of the persistent renter data.
+	persistence struct {
+		MaxdownloadSpeed int64
+		MaxUploadSpeed   int64
+		StreamCacheSize  uint64
+		Repairing        map[string]string
```
DavidVorick: Don't need the repairing map, that's legacy code that's no longer necessary on master.

modules/renter/renter.go
```diff
@@ -455,6 +463,22 @@ func NewCustomRenter(g modules.Gateway, cs modules.ConsensusSet, tpool modules.T
...
 	}
 
+	// Set RenterSettings to persisted data
+	if r.persist.MaxdownloadSpeed == 0 {
+		r.persist.MaxdownloadSpeed = defaultMaxDownloadSpeed
+	}
+	if r.persist.MaxUploadSpeed == 0 {
+		r.persist.MaxUploadSpeed = defaultMaxUploadSpeed
+	}
+	if r.persist.StreamCacheSize == 0 {
```
DavidVorick: This prevents the user from setting the StreamCacheSize to 0. I realize that we currently don't allow users to set the stream cache size to zero, but if we enable that in the future users are going to run into an issue where they are unable to save a stream cache with size zero, it'll always load and reset to 2.

This is a similar issue for MaxDownloadSpeed and MaxUploadSpeed. It happens here that the defaults are also zero, and I don't see us ever changing those defaults (hard to imagine a sane default that would be some other value), but it's still clunky code that works due to unique circumstances rather than being inherently correct code.

The proper way to do this upgrade would be to increase the version of the renter persist. There's a metadata object that's currently got the version set to v0.4. What we would want to do is update that to v1.3.3. This would cause legacy renter persist to fail to load with a 'bad version' error, in which case you'd want to try again with the legacy version. If that succeeds, you then know to set these new values to their defaults. That means that instead of checking for the empty value, you know that you are only setting these fields for the user at startup if they are doing a genuine upgrade from an earlier version.

I think I'm in favor of switching to a version upgrade here. I know it's not strictly necessary, but taking a shortcut might lead to issues later. But this is also a really simple persistence upgrade to perform, and a good way to get your feet wet with our persistence system.

modules/renter/streamcache.go
```diff
@@ -172,9 +172,9 @@ func (sc *streamCache) SetStreamingCacheSize(cacheSize uint64) {
 func newStreamCache() *streamCache {
 	streamHeap := make(streamHeap, 0, defaultStreamCacheSize)
 	heap.Init(&streamHeap)
+	// cacheSize not set as it is persisted
 	return &streamCache{
 		streamMap:  make(map[string]*chunkData),
 		streamHeap: streamHeap,
-		cacheSize:  defaultStreamCacheSize,
```
DavidVorick: shouldn't this be set to `r.persist.StreamCacheSize`?

DavidVorick: oh, you will probably need to make `newStreamCache` a function of the renter. That should fix up all of the tests that broke as well.

siatest/renter/renter_test.go
```diff
@@ -716,6 +717,76 @@ func TestRenewFailing(t *testing.T) {
...
+	r := tg.Renters()[0]
+
+	// Check Settings, should be defaults
+	rg, err := r.RenterGet()
+	if err != nil {
+		t.Fatal(err, "Could not get Renter through RenterGet()")
+	}
+	// Current defaults are 2,0,0 for StreamCacheSize, MaxDownloadSpeed
+	// and MaxUploadSpeed
+	if rg.Settings.StreamCacheSize != 2 {
```
DavidVorick: you should be able to just compare to the default constants right, instead of having hardcoded values?

DavidVorick:
I could be mistaken, but it looks like (based on one of my comments) this PR introduced a bug where the stream cache was not being initialized correctly on startup, even though the user's stream cache size was being persisted correctly.

So we should probably extent the persistence test that you wrote to also check that the actual cache (instead of just the persisted value) is initialized correctly at startup.

modules/renter/streamcache.go
MSevey: @DavidVorick i think you are looking at an out commit, can you check my next commit and let me know if your comment still applies

doc/API.md
```diff
@@ -908,9 +908,9 @@ funds // hastings
 hosts
 period      // block height
 renewwindow // block height
-maxdownloadspeed  // bytes per second, not persisted and will be reset by a shutdown
-maxuploadspeed  // bytes per second, not persisted and will be reset by a shutdown
-streamcachesize // number of data chunks cached when streaming, not persisted and will be reset by a shutdown
+maxdownloadspeed  // bytes per second
+maxuploadspeed  // bytes per second
+streamcachesize // number of data chunks cached when streaming
```
DavidVorick: non-blocking, but we like to align the comments if they are all in a row, example:

```
abc     // comment 1
defghij // comment 2
lmop    // comment 3
```

modules/renter/persist.go
```diff
@@ -234,17 +244,19 @@ func (r *Renter) load() error {
...
-	if err != nil {
-		return err
+	// Load contracts and entropy.
+	r.persist = persistence{
+		Tracking: make(map[string]trackedFile),
 	}
-	if data.Tracking != nil {
-		r.tracking = data.Tracking
+	err = persist.LoadJSON(settingsMetadata, &r.persist, filepath.Join(r.persistDir, PersistFilename))
+	if err != nil {
```
DavidVorick: `if err == persist.ErrBadVersion` I believe is fine here. Then you'll have to have an `else if err !- nil` after the statement to handle the rest of the cases. That way corrupt files are handled correctly right away instead of trying a bunch of recovery steps.

modules/renter/persist.go
```diff
@@ -446,3 +458,17 @@ func (r *Renter) LoadSharedFilesASCII(asciiSia string) ([]string, error) {
...
+		Header:  settingsMetadata.Header,
+		Version: persistVersion040,
+	}
+
+	err := persist.LoadJSON(metadata, &r.persist, filepath.Join(r.persistDir, PersistFilename))
+	if err != nil {
+		return err
+	}
+	metadata.Version = persistVersion133
+	return persist.SaveJSON(metadata, r.persist, filepath.Join(r.persistDir, PersistFilename))
```
DavidVorick: Part of the upgrade is setting the new values that were added to their default settings.

modules/renter/persist_test.go
```diff
@@ -317,3 +317,28 @@ func TestSiafileCompatibility(t *testing.T) {
...
+
+	// Confirm loading of legacy persist file
+	r.persist = persistence{
+		Tracking: make(map[string]trackedFile),
+	}
+	settingsMetadata.Version = persistVersion
+	err := r.load()
+	if err != nil {
+		t.Fatal(err)
+	}
```
DavidVorick: After loading, need to check that the new values were properly set to the default values for those settings.

modules/renter/renter.go
```diff
@@ -451,9 +459,24 @@ func NewCustomRenter(g modules.Gateway, cs modules.ConsensusSet, tpool modules.T
...
 		return nil, err
 	}
+	r.mu.Unlock(id)
+
+	// Set RenterSettings to persisted data
+	if settingsMetadata.Version != persistVersion {
+		r.persist.MaxDownloadSpeed = DefaultMaxDownloadSpeed
+		r.persist.MaxUploadSpeed = DefaultMaxUploadSpeed
+		r.persist.StreamCacheSize = DefaultStreamCacheSize
+	}
```
DavidVorick: This block of code should be removed, now it gets handled in the upgrade code.

DavidVorick: oh wait, maybe I was wrong about that. I can't tell at the moment how the default settings are being established upon the first creation of the renter, but it looks like there's a test demonstrating that they actually do get set correctly.

MSevey: You were right the first time, this code can be removed now that it is in the upgrade code.  `initPersist` is called above which will set the persistence for legacy persistence, and then for current persistence files I just have to read the fields.

siatest/testnode.go
```diff
@@ -73,3 +100,18 @@ func NewCleanNode(nodeParams node.NodeParams) (*TestNode, error) {
 	// Return TestNode
 	return tn, nil
 }
+
+// LoadNode loads a TestNode when betwen restarted
```
ChrisSchinnerl: "when bewen restarted"

siatest/testnode.go
```diff
@@ -18,6 +18,33 @@ type TestNode struct {
...
+	err = tn.StartNode()
+	if err != nil {
+		return errors.AddContext(err, "Could not start node")
+	}
+	return nil
+}
+
+// StartNode starts a TestNode from an active group
+func (tn *TestNode) StartNode() error {
+	err := tn.LoadNode(tn.params)
```
ChrisSchinnerl: It seems like we only use `LoadNode` here. Why not just move the `LoadNode` code to `StartNode`?

ChrisSchinnerl: I think we also might need a boolean field to check if a node is running or not. Then we can sanity check `StartNode`, `StopNode` and `RestartNode`.

persist/testdata/renter.json
```diff
@@ -0,0 +1,9 @@
+"Renter Persistence"
```
ChrisSchinnerl: Was that file created by a test? Then it should be in `tmp/SiaTesting` instead.

persist/testdata/renter.json
```diff
@@ -0,0 +1,9 @@
...
+"1.3.3"
```
ChrisSchinnerl: Same for this one

persist/testdata/renter.json_temp
```diff
@@ -0,0 +1,9 @@
+"Renter Persistence"
```
ChrisSchinnerl: Also this one.

modules/renter/renter_test.go
```diff
@@ -27,6 +27,7 @@ type renterTester struct {
 	walletKey crypto.TwofishKey
 
 	renter *Renter
+	dir    string
```
ChrisSchinnerl: Do we need this now that we moved the test to the siatest package?

modules/renter/renter_test.go
```diff
@@ -85,6 +86,7 @@ func newRenterTester(name string) (*renterTester, error) {
 		wallet:  w,
 
 		renter: r,
+		dir:    testdir,
```
ChrisSchinnerl: Same for this

modules/renter/renter.go
```diff
@@ -286,31 +291,66 @@ func (r *Renter) PriceEstimation() modules.RenterPriceEstimation {
...
 // SetSettings will update the settings for the renter.
+//
+// NOTE: This function can't be atomic. Typically we try to have user requests
+// be atomic, so that either everything changes or nothing changes, but since
+// these changes happen progressively, it's possible for some of the settings
+// (like the allowance) to succeed, but then if the bandwidth limits for example
+// are bad, then the allowance will update but the bandwidth will not update.
 func (r *Renter) SetSettings(s modules.RenterSettings) error {
+	// Early input valudation.
+	if s.MaxDownloadSpeed < 0 || s.MaxUploadSpeed < 0 {
```
ChrisSchinnerl: Is that redundant on purpose? You also check for it in `setBandwidthLimits`.

modules/renter/renter.go
```diff
@@ -286,31 +291,66 @@ func (r *Renter) PriceEstimation() modules.RenterPriceEstimation {
...
+		return errors.New("bandwidth limits cannot be negative")
+	}
+	if s.StreamCacheSize <= 0 {
```
ChrisSchinnerl: The check above is redundant but the one only happens here and not in `SetStreamCacheSize`.

modules/renter/renter.go
```diff
@@ -165,8 +170,7 @@ type Renter struct {
 	//
 	// tracking contains a list of files that the user intends to maintain. By
 	// default, files loaded through sharing are not maintained by the user.
-	files    map[string]*file
-	tracking map[string]trackedFile // Map from nickname to metadata.
+	files map[string]*file
```
ChrisSchinnerl: Haha that is actually something I was wondering about when working on refactoring the file format. I'm glad that it's gone now.

modules/renter/streamcache.go
MSevey: The comment `cacheSize not set as it is persisted` can be removed now since the method was updated to take the cacheSize as an argument and set it

siatest/testnode.go
DavidVorick: ouch

modules/renter/renter.go
DavidVorick: It's not gone it just got moved

modules/renter/renter.go
DavidVorick: yes, want to catch the issues earlier rather than later

modules/renter/renter.go
DavidVorick: ok, will extend

modules/renter/persist.go
```diff
@@ -230,24 +241,40 @@ func (r *Renter) load() error {
 		}
 		return nil
 	})
-	if err != nil {
-		return err
+}
+
+// load fetches the saved renter data from disk.
```
MSevey: Update comment, load changed to loadSettings

modules/renter/renter_test.go
DavidVorick: all the other modules have it

modules/renter/renter_test.go
DavidVorick: same story

modules/renter/streamcache.go
DavidVorick: ok

modules/renter/persist.go
```diff
@@ -230,24 +241,40 @@ func (r *Renter) load() error {
...
+	} else if err == persist.ErrBadVersion {
+		// Outdated version, try the 040 to 133 upgrade.
+		err = r.updatePersistVersionFrom040To133()
+		if err != nil {
+			// Nothing left to try.
+			return err
+		}
+		// Re-load the settings now that the file has been upgraded.
+		return r.loadSettings()
 	}
```
MSevey: Shouldn't we still be checking for `err != nil` after those two error conditions?

modules/renter/persist_test.go
```diff
@@ -229,6 +252,17 @@ func TestRenterSaveLoad(t *testing.T) {
...
+	newSettings := rt.renter.Settings()
+	if newSettings.MaxDownloadSpeed != newDownSpeed {
+		t.Error("download settings not being persisted correctly")
+	}
+	if newSettings.MaxUploadSpeed != newUpSpeed {
+		t.Error("upload settings not being persisted correctly")
+	}
+	if newSettings.StreamCacheSize != newCacheSize {
+		t.Error("cache settings not being persisted correctly")
+	}
```
DavidVorick: we kept this test around

modules/renter/persist.go
```diff
@@ -446,3 +475,20 @@ func (r *Renter) LoadSharedFilesASCII(asciiSia string) ([]string, error) {
 	dec := base64.NewDecoder(base64.URLEncoding, bytes.NewBufferString(asciiSia))
 	return r.loadSharedFiles(dec)
 }
+
+func (r *Renter) updatePersistVersionFrom040To133() error {
```
ChrisSchinnerl: Missing comment

modules/renter/persist.go
```diff
@@ -446,3 +477,22 @@ func (r *Renter) LoadSharedFilesASCII(asciiSia string) ([]string, error) {
...
+	err := persist.LoadJSON(metadata, &r.persist, filepath.Join(r.persistDir, PersistFilename))
+	if err != nil {
+		return err
+	}
+	metadata.Version = persistVersion133
+	r.persist.MaxDownloadSpeed = DefaultMaxDownloadSpeed
+	r.persist.MaxUploadSpeed = DefaultMaxUploadSpeed
+	r.persist.StreamCacheSize = DefaultStreamCacheSize
+	return persist.SaveJSON(metadata, r.persist, filepath.Join(r.persistDir, PersistFilename))
+}
```
lukechampine: this function has some side effects that aren't obvious to the caller, like setting fields in `r.persist`. I suggest instead doing:
```go
func updatePersistVersionFrom040To133(persistPath string) error {
    var p persistence
    LoadJSON(&p, persistPath)
    p.MaxDownloadSpeed = DefaultMaxDownloadSpeed
    // etc.
    return SaveJSON(p, persistPath)
}
```
then it's a bit more clear that the old side effect is rewriting the file.

siatest/renter/renter_test.go
```diff
@@ -520,7 +522,7 @@ func testRenterStreamingCache(t *testing.T, tg *siatest.TestGroup) {
 	prev := rg.Settings.StreamCacheSize
 
 	// Test setting to 0
-	if err := r.RenterSetStreamCacheSizePost(0); err != nil {
+	if err := r.RenterSetStreamCacheSizePost(0); err == nil {
 		t.Fatal(err, "Error in calling RenterSetStreamCacheSizePost(0)")
```
lukechampine: should be `t.Fatal("expected error when settings stream cache size to zero")`

modules/renter/persist.go
```diff
@@ -230,24 +241,42 @@ func (r *Renter) load() error {
...
 	})
-	if err != nil {
+}
+
+// load fetches the saved renter data from disk.
+func (r *Renter) loadSettings() error {
+	r.persist = persistence{
+		Tracking: make(map[string]trackedFile),
+	}
+	err := persist.LoadJSON(settingsMetadata, &r.persist, filepath.Join(r.persistDir, PersistFilename))
```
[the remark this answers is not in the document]
DavidVorick: no we get panics if we remove that line

modules/renter/persist.go
DavidVorick: not sure I prefer it that way

DavidVorick: That requires initializing the map in the persistence, and it also requires the caller to identify the persist path, and it also means that the renter defaults aren't being set directly they must be initialized by reloading the file (which is what we do anyway, but it feels fragile to leave that out of this piece)

lukechampine: I don't see the issue with passing the persist path. It makes sense when you think of the function as rewriting a specific file. As for setting defaults, I can see why it might feel fragile to you. "Persist" can refer to both the file on disk and a field in the renter, so it makes sense that when you upgrade one, you should upgrade the other in tandem.

I think this might just be a problem with the name. In other modules, we've called this sort of function `convert`. In the contractor, there's a `convertPersist` function that converts old contractor.json to the new format. Similarly, in the wallet, there's a `convertPersistFrom112To120` function that converts the old wallet.json to a wallet.db. (Both of these functions take filenames as arguments.) "Convert" communicates the intent of the function clearly: it converts one thing into another thing. "Upgrade" is a more ambiguous name, especially when it's a method of the renter instead of a standalone function.

I suggest that we rename the function to `convertPersistFrom040To133` and pass it the filename as an argument. Then it will match the persist conversion functions in the contractor and wallet. Alternatively, we could think of a better name that makes it more obvious that the function also sets fields in the renter.

modules/renter/persist.go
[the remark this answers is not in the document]
lukechampine: oh, that's probably because there's no file to load (the `os.IsNotExist` branch), so the field stays `nil`. The old persist code handled this like so:
https://github.com/NebulousLabs/Sia/blob/f9c05eca7c88c3191be9bb0752ca9028626107ad/modules/renter/persist.go#L237-L248

neither solution is very pretty, unfortunately. Initializing the field inside the struct declaration seems a bit cleaner though.

modules/renter/persist.go
DavidVorick: alright

PR #3057, opened 2018-05-28
Replace TestRedundancyReporting with siatest

ChrisSchinnerl:
It seems like `TestRedundancyReporting` contains a NDF so I'd like to move it to the siatest package.
This PR depends on the siatest package changes from [PR#3052](https://github.com/NebulousLabs/Sia/pull/3052).

EDIT: This also fixes a few bugs in the siatest framework I noticed while writing porting this test over.


MSevey:
@ChrisSchinnerl what from #3052 are you waiting on?  My PR keeps failing because of `TestRedundancyReporting`.  This is a real chicken or the egg situation haha

siatest/testgroup.go
```diff
@@ -293,40 +292,46 @@ func setRenterAllowances(renters map[*TestNode]struct{}) error {
...
+func synchronizationCheck(nodes map[*TestNode]struct{}) error {
+	// Get node with longest chain.
+	var longestChainNode *TestNode
+	var longestChain types.BlockHeight
+	for n := range nodes {
+		ncg, err := n.ConsensusGet()
+		if err != nil {
+			return err
+		}
+		if ncg.Height > longestChain {
```
DavidVorick: What happens if two nodes have a chain of the same length but it's a different block?

siatest/testgroup.go
```diff
@@ -293,40 +292,46 @@ func setRenterAllowances(renters map[*TestNode]struct{}) error {
...
-					mcg.Height, ncg.Height, len(gwgNode.Peers), len(gwgMiner.Peers)), errMiner, errNode)
+			if lcg.Height != ncg.Height {
+				return errors.New("blockHeight doesn't match")
 			}
 			// If the miner's height is smaller than the node's we need a
 			// bit longer for them to sync.
-			if mcg.Height < ncg.Height {
-				return errors.New("the miner didn't catch up to the node's height")
+			if lcg.CurrentBlock != ncg.CurrentBlock {
+				return errors.New("ids don't match")
```
DavidVorick: yeah so this could be a perpetual error if you don't verify initially that the block at the highest height is the only block at that height.

DavidVorick:
looks good except for the check that needs to be added to `synchronizationCheck`.

I like that it's `+150 / -250` even though a lot of the +150 is even extending the siatest package. Means that the siatest package is a lot more powerful and doing a good job.

siatest/testgroup.go
ChrisSchinnerl: I've been thinking about that and I think we should keep that as intended behavior. 
`synchronizationCheck` should make sure that all nodes are at the same height but it is not expected to mine a block if that is not the case. 
That means if a test runs into issues because of that it should be handled somewhere else in my opinion.

siatest/testgroup.go
ChrisSchinnerl: In that case the check is going to fail.

siatest/testgroup.go
DavidVorick: The name `synchronizationCheck` implies that all of the nodes are on the same block - that they are sychronized. If the nodes are not on the same block, it should return an error that the nodes are not on the same block.

Catching this error here instead of earlier means that we have to wait for the full timeout of the `retry` when we could have returned the error immediately.

PR #3066, opened 2018-06-01
Switch from FileContractIDs to HostPubKeys for contractor interface

ChrisSchinnerl:
This PR is the first step towards the new sia file format. Since the new format no longer stores contract ids but instead the public key of the host, this PR changes the Contractor's interface to accept public keys.


modules/renter/contractor/contractor.go
```diff
@@ -147,7 +143,13 @@ func (c *Contractor) Contracts() []modules.RenterContract {
 }
 
 // ContractUtility returns the utility fields for the given contract.
-func (c *Contractor) ContractUtility(id types.FileContractID) (modules.ContractUtility, bool) {
+func (c *Contractor) ContractUtility(pk types.SiaPublicKey) (modules.ContractUtility, bool) {
+	c.mu.Lock()
```
DavidVorick: any reason you chose the lock instead of rlock?

modules/renter/contractor/contractor.go
```diff
@@ -237,13 +242,14 @@ func NewCustomContractor(cs consensusSet, w wallet, tp transactionPool, hdb host
...
-		oldContracts:    make(map[types.FileContractID]modules.RenterContract),
-		renewedIDs:      make(map[types.FileContractID]types.FileContractID),
-		renewing:        make(map[types.FileContractID]bool),
-		revising:        make(map[types.FileContractID]bool),
+		staticContracts:     contractSet,
+		downloaders:         make(map[types.FileContractID]*hostDownloader),
+		editors:             make(map[types.FileContractID]*hostEditor),
+		oldContracts:        make(map[types.FileContractID]modules.RenterContract),
+		contractIDToPubKey:  make(map[types.FileContractID]types.SiaPublicKey),
+		pubKeysToContractID: make(map[string]types.FileContractID),
```
DavidVorick: It looks like this map currently only stores the most recent contract associated with a host pubkey?

modules/renter/contractor/contracts.go
```diff
@@ -252,6 +246,12 @@ func (c *Contractor) managedRenew(sc *proto.SafeContract, contractFunding types.
 		return modules.RenterContract{}, err
 	}
 
+	// Add a mapping from the contract's id to the public key of the host.
```
DavidVorick: should probably note that this is going to obliterate any previous mapping.

modules/renter/contractor/contracts.go
```diff
@@ -194,6 +182,12 @@ func (c *Contractor) managedNewContract(host modules.HostDBEntry, contractFundin
...
+	c.mu.Lock()
+	c.contractIDToPubKey[contract.ID] = contract.HostPublicKey
+	c.pubKeysToContractID[string(contract.HostPublicKey.Key)] = contract.ID
+	c.mu.Unlock()
```
DavidVorick: should probably sanity check that you aren't obliterating a previous mapping, otherwise you are going to lose a contract and potentially not realize it.

modules/renter/contractor/contracts.go
```diff
@@ -318,7 +318,7 @@ func (c *Contractor) threadedContractMaintenance() {
 	var endHeight types.BlockHeight
 	var fundsAvailable types.Currency
 	var renewSet []renewal
-	refreshSet := make(map[types.FileContractID]struct{})
+	refreshSet := make(map[string]struct{})
```
DavidVorick: is there a reason that you changed this type from a `types.FileContractID` to a `string`?

modules/renter/contractor/contractor.go
ChrisSchinnerl: Yes, that's intentional. The contractor only needs this map to return the most recent contract with a host to the caller of the interface. Other modules don't really care about older revisions.

modules/renter/contractor/contractor.go
ChrisSchinnerl: Nope, I can change that.

modules/renter/contractor/contracts.go
ChrisSchinnerl: lol I just realized that we never used this set anyway. Seems like we can delete that.
The reason for the string was the change to a pubKey instead of a contract id. Go can't use byte slices as a mapkey.

modules/renter/contractor/contracts.go
```diff
@@ -185,9 +185,14 @@ func (c *Contractor) managedNewContract(host modules.HostDBEntry, contractFundin
 	// Add a mapping from the contract's id to the public key of the host.
 	c.mu.Lock()
 	c.contractIDToPubKey[contract.ID] = contract.HostPublicKey
+	_, exists := c.pubKeysToContractID[string(contract.HostPublicKey)]
 	c.pubKeysToContractID[string(contract.HostPublicKey.Key)] = contract.ID
 	c.mu.Unlock()
 
+	if exists {
+		build.Critical("We are forming a new contract but the host's pubKey is already mapped to a filecontract's id")
```
DavidVorick: I don't know if this needs to be a `build.Critical`, I feel like just returning and error and logging it is fine.

modules/renter/contractor/contractor.go
```diff
@@ -131,12 +122,17 @@ func (c *Contractor) PeriodSpending() modules.ContractorSpending {
...
-// contract will be resolved if possible to the most recent child contract.
-func (c *Contractor) ContractByID(id types.FileContractID) (modules.RenterContract, bool) {
-	c.mu.RLock()
-	defer c.mu.RUnlock()
-	return c.staticContracts.View(c.readlockResolveID(id))
+// ContractByPublicKey returns the contract with the key specified, if it
+// exists. The contract will be resolved if possible to the most recent child
+// contract.
+func (c *Contractor) ContractByPublicKey(pk types.SiaPublicKey) (modules.RenterContract, bool) {
+	c.mu.Lock()
```
DavidVorick: I believe this one can also be an RLock

modules/renter/contractor/contracts.go
```diff
@@ -197,6 +185,16 @@ func (c *Contractor) managedNewContract(host modules.HostDBEntry, contractFundin
 		return modules.RenterContract{}, err
 	}
 
+	// Add a mapping from the contract's id to the public key of the host.
+	c.mu.Lock()
+	c.contractIDToPubKey[contract.ID] = contract.HostPublicKey
+	_, exists := c.pubKeysToContractID[string(contract.HostPublicKey.Key)]
+	if exists {
+		c.log.Println("We are forming a new contract but the host's pubKey is already mapped to a filecontract's id")
```
DavidVorick: shouldn't we `continue` or `return` here?

DavidVorick: We should probably `return`

PR #3073, opened 2018-06-01
Fix bug where API password was being printed

tbenz9:


cmd/siac/main.go
```diff
@@ -143,7 +143,7 @@ func main() {
 	}
 	// initialize client
 	root.PersistentFlags().StringVarP(&httpClient.Address, "addr", "a", "localhost:9980", "which host/port to communicate with (i.e. the host/port siad is listening on)")
-	root.PersistentFlags().StringVarP(&httpClient.Password, "apipassword", "", apiPassword, "the password for the API's http authentication")
+	root.PersistentFlags().StringVarP(&httpClient.Password, "apipassword", "", "", "the password for the API's http authentication")
```
ChrisSchinnerl: Have you tested this? I think this breaks the Environment variable.

tbenz9: Damn, you're correct.  Sia reports that it's using the environment variable but the wallet is still locked.

ChrisSchinnerl: Cobra is usually Luke's expertise but maybe moving the environment check below the call to `StringVarP` and then setting `httpClient.Password` to `apiPassword` if it is not `""` works.

lukechampine: Looks like cobra doesn't handle env vars, but instead leaves that up to [viper](https://github.com/spf13/viper), which I don't really want to start using. Setting it manually seems fine to me.

lukechampine:
what's the precedence now? If you supply a flag and set the env var, which one is used?

ChrisSchinnerl:
@lukechampine Same as before. If you provide both, the flag will be used.

PR #3074, opened 2018-06-01
Add hostdb endpoint to indicate intial scan status and queued scans

ChrisSchinnerl:
This Pr adds a new endpoint that lets the user retrieve information about the hostdb.
At this point it contains the list of hosts that are being scanned at the moment and the a flag indicating if the initial scan of the hostdb is complete.
That is useful since we currently block uploads and contract formation/renewals until the initial scan is done.
Once this endpoint is reviewed and after I've received some feedback I'll update the documentation and add an automated test for it.

### DIFF
--- a/modules/renter/hostdb/hostdb.go
+++ b/modules/renter/hostdb/hostdb.go
@@ -15,8 +15,8 @@ import (
 	"github.com/NebulousLabs/Sia/modules"
 	"github.com/NebulousLabs/Sia/modules/renter/hostdb/hosttree"
 	"github.com/NebulousLabs/Sia/persist"
-	siasync "github.com/NebulousLabs/Sia/sync"
 	"github.com/NebulousLabs/Sia/types"
+	"github.com/NebulousLabs/threadgroup"
 )
 
 var (
@@ -38,7 +38,7 @@ type HostDB struct {
 	log        *persist.Logger
 	mu         sync.RWMutex
 	persistDir string
-	tg         siasync.ThreadGroup
+	tg         threadgroup.ThreadGroup
 
 	// The hostTree is the root node of the tree that organizes hosts by
 	// weight. The tree is necessary for selecting weighted hosts at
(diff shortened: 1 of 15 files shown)


modules/renter/hostdb/hostdb.go
```diff
@@ -242,3 +254,16 @@ func (hdb *HostDB) RandomHosts(n int, excludeKeys []types.SiaPublicKey) ([]modul
 	}
 	return hdb.hostTree.SelectRandom(n, excludeKeys), nil
 }
+
+// QueuedScans returns the currently queued scans of the hostdb.
+func (hdb *HostDB) QueuedScans() (hosts []modules.HostDBEntry, err error) {
```
DavidVorick: Should this return the list of hosts, or just the quantity? Returning the full list seems like overkill.

modules/renter/hostdb/hostdb.go
```diff
@@ -230,6 +230,18 @@ func (hdb *HostDB) Host(spk types.SiaPublicKey) (modules.HostDBEntry, bool) {
 	return host, exists
 }
 
+// InitialScanComplete returns a boolean indicating if the initial scan of the
+// hostdb is completed.
+func (hdb *HostDB) InitialScanComplete() (complete bool, err error) {
```
DavidVorick: Should we have this be a separate function entirely, or just an extra return value in `QueuedScans()`? Might be simpler to have just one function which does both things. I could go either way.

node/api/hostdb.go
```diff
@@ -35,8 +35,33 @@ type (
 		Entry          ExtendedHostDBEntry        `json:"entry"`
 		ScoreBreakdown modules.HostScoreBreakdown `json:"scorebreakdown"`
 	}
+
+	// HostDBGet holds information about the hostdb.
+	HostDBGet struct {
+		InitialScanComplete bool                  `json:"initialscancomplete"`
+		QueuedScans         []modules.HostDBEntry `json:"queuedscans"`
```
DavidVorick: Returning a large list of hostdb entries might be too much data to pass through the API, especially if it's early and there are thousands of scans remaining.

modules/renter/hostdb/hostdb.go
ChrisSchinnerl: Yep also something I thought about when writing it. I like to keep things separated but that's just my preference.

node/api/hostdb.go
ChrisSchinnerl: Yeah I wasn't sure about that. It introduces a lot of overhead but I'm wondering if it would be useful to have an endpoint for the UI that lets you check information about hosts that are currently being scanned.

DavidVorick:
Still need a test demonstrating that the endpoint is working. Or need to extend an existing hostdb test to include using this endpoint to verify that the scanning stuff is working correctly.

PR #3080, opened 2018-06-04
Only replace unrenewable contract if it's the host's fault

ChrisSchinnerl:


modules/errors.go
```diff
@@ -0,0 +1,12 @@
+package modules
+
+import "github.com/NebulousLabs/errors"
+
+// ErrHostFault is an error that is usually extended to indicate that an error
+// is the host's fault.
+var ErrHostFault = errors.New("")
```
DavidVorick: error string should be human readable. IIRC the `Extend` function will add some colons and brackets, so an empty error will look pretty confusing when printed out.

maybe say "host has returned an error:"

modules/errors.go
```diff
@@ -0,0 +1,12 @@
+package modules
```
DavidVorick: not sure creating a whole errors file is best, I think it's probably better to put these lines into the `renter.go` file since this error is really just relevant to the renter.

modules/renter/proto/formcontract.go
```diff
@@ -108,6 +109,7 @@ func (cs *ContractSet) FormContract(params ContractParams, txnBuilder transactio
 	defer func() {
 		if err != nil {
 			hdb.IncrementFailedInteractions(host.PublicKey)
+			err = errors.Extend(err, modules.ErrHostFault)
```
DavidVorick: hmm. not sure that I like this, because it assumes that any error which occurs later is automatically the host's fault.

It appears to be true here, but it might not be something that we think about correctly if we come back and modify this file later.

Though, it is a protocol file, which means it's not likely to ever be modified. We try not to touch those.

DavidVorick: no I take it back. Errors can be returned by `AcceptContractSet` and `managedInsertContract` and I'm not sure that either of those errors would indicate the host's fault.

modules/renter/proto/renew.go
```diff
@@ -110,6 +111,7 @@ func (cs *ContractSet) Renew(oldContract *SafeContract, params ContractParams, t
 		// A revision mismatch might not be the host's fault.
 		if err != nil && !IsRevisionMismatch(err) {
 			hdb.IncrementFailedInteractions(contract.HostPublicKey())
+			err = errors.Extend(err, modules.ErrHostFault)
```
DavidVorick: same here, a small number of the errors after this deferred statement appear like they don't imply a problem on the host-side. Mostly AcceptTransactionSet

modules/errors.go
ChrisSchinnerl: lol I told myself I would worry about a meaningful string later. Apparently I forgot about that.

modules/renter/proto/formcontract.go
ChrisSchinnerl: In that case we also need to adjust all calls to `IncrementFailedInteractions` since they already assume that it's the hosts fault.

DavidVorick: Agreed

Issue #3067, opened 2018-06-01
Feature Request: Renter should know and report (in API) if a file is apparently lost


KaiRo-at:
When a file has been deleted locally and at least some part of the file is not available on the network, the renter should be able to derive that this file appears to be lost, mark that internally and report that in the API. If there is a contract/host still covering the missing piece(s), repair may still go on so that we can recover the file if the right offline host(s) come back. If that chance doesn't exist, the renter may even give up and stop to repair it as it costs money and we'll never be able to retrieve the whole file again anyhow.
The UI or someone else interacting with the API may decide to delete all files marked as "lost" in that way or re-upload them if they can get to a local copy in some other way.

MSevey:
Task would be to recognize when redundancy is <1x and file can not be found on local disk

PR #3083, opened 2018-06-05
Detect when local file is lost before 1x redundancy

### LINKED ISSUE #2488: Detect when local file is lost before 1x redundancy
Bug Report
==========

Sia should detect if the file it's trying to upload is no longer present on the filesystem.

**To reproduce**:

```
$ head -c 300M </dev/urandom > /tmp/testfile.txt
$ ./siac renter upload /tmp/testfile.txt testfile.txt
Uploaded '/tmp/testfile.txt' as testfile.txt.
$ rm /tmp/testfile.txt
$ ./siac renter -v
...
File size  Available  Progress  Redundancy  Renewing  Sia path
314.57 MB  No           50.00%        0.00  Yes       testfile.txt
```

**Expected**

`./siac renter -v` output indicates that the source file is missing and the upload is stuck.
`renter.log` contains information about the error

**Actual**

`./siac renter -v` does not indicate any problems except that the user will notice no upload progress.
`renter.log` contains a vague error message that does not specify a particular file or say that it's missing:

```
2017/11/25 05:40:51.070595 download.go:501: Not enough workers to finish download: insufficient hosts to recover file
2017/11/25 05:40:51.078194 download.go:501: Not enough workers to finish download: insufficient hosts to recover file
2017/11/25 05:40:51.145722 download.go:501: Not enough workers to finish download: insufficient hosts to recover file
2017/11/25 05:40:51.150447 download.go:501: Not enough workers to finish download: insufficient hosts to recover file
```

This caused a blocking bug in the Sia Minio integration: https://github.com/minio/minio/pull/5233

**Environment**

* Sia version: 47514b49625fe841348d24773ba4c7c54da4fa17
* OS: Ubuntu 16.04 LTS x64

MSevey:
@DavidVorick @ChrisSchinnerl I want to check and see if I was going down the right path with this one.

So far I added a message in `siac renter -v` under the upload progress to notify the user that the local file was not found before 1x redundancy was reached.  I also added a log in `managedBuildChunkHeap()` noting that the local file was lost.

Couple questions:
1. Are there other locations, or better locations to capture this?
2. Are there additional steps that we should take when we identify this?  Or should we just notify the user and leave it up to them to decide what to do?

@mtlynch  This PR is to resolve #2488

### DIFF
--- a/cmd/siac/rentercmd.go
+++ b/cmd/siac/rentercmd.go
@@ -541,7 +541,7 @@ func renterfileslistcmd() {
 	fmt.Printf("Total uploaded: %9s\n", filesizeUnits(int64(totalStored)))
 	w := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
 	if renterListVerbose {
-		fmt.Fprintln(w, "File size\tAvailable\tUploaded\tProgress\tRedundancy\tRenewing\tSia path")
+		fmt.Fprintln(w, "File size\tAvailable\tUploaded\tProgress\tRedundancy\tRenewing\tOn Disk\tRecoverable\tSia path")
 	}
 	sort.Sort(bySiaPath(rf.Files))
 	for _, file := range rf.Files {
@@ -554,13 +554,13 @@ func renterfileslistcmd() {
 				redundancyStr = "-"
 			}
 			uploadProgressStr := fmt.Sprintf("%.2f%%", file.UploadProgress)
-			if _, err := os.Stat(file.LocalPath); os.IsNotExist(err) && file.Redundancy < 1 {
-				uploadProgressStr = fmt.Sprintf("Local File Lost")
-			}
+			_, err := os.Stat(file.LocalPath)
+			onDiskStr := yesNo(!os.IsNotExist(err))
+			recoverableStr := yesNo(!(file.Redundancy < 1))
 			if file.UploadProgress == -1 {
 				uploadProgressStr = "-"
 			}
-			fmt.Fprintf(w, "\t%s\t%9s\t%8s\t%10s\t%s", availableStr, filesizeUnits(int64(file.UploadedBytes)), uploadProgressStr, redundancyStr, renewingStr)
+			fmt.Fprintf(w, "\t%s\t%9s\t%8s\t%10s\t%s\t%s\t%s", availableStr, filesizeUnits(int64(file.UploadedBytes)), uploadProgressStr, redundancyStr, renewingStr, onDiskStr, recoverableStr)
 		}
 		fmt.Fprintf(w, "\t%s", file.SiaPath)
 		if !renterListVerbose && !file.Available {
--- a/cmd/siac/rentercmd.go
+++ b/cmd/siac/rentercmd.go
@@ -554,6 +554,9 @@ func renterfileslistcmd() {
 				redundancyStr = "-"
 			}
 			uploadProgressStr := fmt.Sprintf("%.2f%%", file.UploadProgress)
+			if _, err := os.Stat(file.LocalPath); os.IsNotExist(err) && file.Redundancy < 1 {
+				uploadProgressStr = fmt.Sprintf("Local File Lost")
+			}
 			if file.UploadProgress == -1 {
 				uploadProgressStr = "-"
 			}
--- a/modules/renter/uploadheap.go
+++ b/modules/renter/uploadheap.go
@@ -242,13 +242,7 @@ func (r *Renter) managedBuildChunkHeap(hosts map[string]struct{}) {
 	offline := make(map[types.FileContractID]bool)
 	for _, file := range r.files {
 		file.mu.RLock()
-		// Build maps that map contract id to its offline and goodForRenew
-		// status
-		contractIDs := make(map[types.FileContractID]struct{})
 		for cid := range file.contracts {
-			contractIDs[cid] = struct{}{}
-		}
-		for cid := range contractIDs {
 			resolvedID := r.hostContractor.ResolveIDToPubKey(cid)
 			cu, ok := r.hostContractor.ContractUtility(resolvedID)
 			goodForRenew[cid] = ok && cu.GoodForRenew
--- a/modules/renter/uploadheap.go
+++ b/modules/renter/uploadheap.go
@@ -237,38 +237,44 @@ func (r *Renter) buildUnfinishedChunks(f *file, hosts map[string]struct{}) []*un
 func (r *Renter) managedBuildChunkHeap(hosts map[string]struct{}) {
 	// Loop through the whole set of files and get a list of chunks to add to
 	// the heap.
-	id := r.mu.Lock()
+	id := r.mu.RLock()
+	goodForRenew := make(map[types.FileContractID]bool)
+	offline := make(map[types.FileContractID]bool)
 	for _, file := range r.files {
-		// check for local file
-		tf, exists := r.persist.Tracking[file.name]
-		if exists {
-			// Build maps that map contract id to its offline and goodForRenew
-			// status
-			contractIDs := make(map[types.FileContractID]struct{})
-			for cid := range file.contracts {
-				contractIDs[cid] = struct{}{}
-			}
-			goodForRenew := make(map[types.FileContractID]bool)
-			offline := make(map[types.FileContractID]bool)
-			for cid := range contractIDs {
-				resolvedID := r.hostContractor.ResolveID(cid)
-				cu, ok := r.hostContractor.ContractUtility(resolvedID)
-				goodForRenew[cid] = ok && cu.GoodForRenew
-				offline[cid] = r.hostContractor.IsOffline(resolvedID)
-			}
-			// Check if local file is missing and redundancy is less than 1
-			// log warning to renter log
-			if _, err := os.Stat(tf.RepairPath); os.IsNotExist(err) && file.redundancy(offline, goodForRenew) < 1 {
-				r.log.Println("File not found on disk:", tf.RepairPath)
-			}
+		file.mu.RLock()
+		// Build maps that map contract id to its offline and goodForRenew
+		// status
+		contractIDs := make(map[types.FileContractID]struct{})
+		for cid := range file.contracts {
+			contractIDs[cid] = struct{}{}
 		}
+		for cid := range contractIDs {
+			resolvedID := r.hostContractor.ResolveIDToPubKey(cid)
+			cu, ok := r.hostContractor.ContractUtility(resolvedID)
+			goodForRenew[cid] = ok && cu.GoodForRenew
+			offline[cid] = r.hostContractor.IsOffline(resolvedID)
+		}
+		file.mu.RUnlock()
 
 		unfinishedUploadChunks := r.buildUnfinishedChunks(file, hosts)
 		for i := 0; i < len(unfinishedUploadChunks); i++ {
 			r.uploadHeap.managedPush(unfinishedUploadChunks[i])
 		}
 	}
-	r.mu.Unlock(id)
+	for _, file := range r.files {
+		file.mu.RLock()
+		// check for local file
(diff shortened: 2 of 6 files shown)


modules/renter/uploadheap.go
```diff
@@ -236,6 +238,30 @@ func (r *Renter) managedBuildChunkHeap(hosts map[string]struct{}) {
...
+				cu, ok := r.hostContractor.ContractUtility(resolvedID)
+				goodForRenew[cid] = ok && cu.GoodForRenew
+				offline[cid] = r.hostContractor.IsOffline(resolvedID)
+			}
+			// Check if local file is missing and redundancy is less than 1
+			// log warning to renter log
+			if _, err := os.Stat(tf.RepairPath); os.IsNotExist(err) && file.redundancy(offline, goodForRenew) < 1 {
+				r.log.Println("File not found on disk:", tf.RepairPath)
+			}
+		}
```
lukechampine: is it possible to move the `offline` and `goodForRenew` map construction outside the loop? I think that's what `FileList` does. It's more efficient and it makes the check much more readable:
```go
// Check for "endangered" files. A file is endangered if its redundancy is
// less than 1 and no local copy is available.
if tf, exists := r.persist.Tracking[file.name]; exists {
    if _, err := os.Stat(tf.RepairPath); os.IsNotExist(err) && file.redundancy(offline, goodForRenew) < 1 {
        r.log.Println("File not found on disk:", tf.RepairPath)
    }
}

Comment (MSevey) [reply modules/renter/uploadheap.go]: `FileList` loops over the files to get a slice of `files` and `contractIDs` then loops over the `contractsIDs` to make the maps of `offline` and `goodForRenew`, and finally loops over the slice of `files` to generate the final splice of `filelist`.

So I could move the map contruction to its own for loop to help with the readability, but would it really make it that much more efficient?  In both cases you need to have the nested for loops of the getting the `contractIDs`, then its another for loop to generate the maps and one last one for check the file redundancy.  Isn't it O(nm) either way?

File: modules/renter/uploadheap.go [hunk 2]
```diff
@@ -236,6 +238,30 @@ func (r *Renter) managedBuildChunkHeap(hosts map[string]struct{}) {
 	// the heap.
 	id := r.mu.Lock()
 	for _, file := range r.files {
+		// check for local file
+		tf, exists := r.persist.Tracking[file.name]
+		if exists {
+			// Build maps that map contract id to its offline and goodForRenew
+			// status
+			contractIDs := make(map[types.FileContractID]struct{})
+			for cid := range file.contracts {
```
ChrisSchinnerl: When accessing fields of the file like `file.name` and `file.contracts` you need to lock `file.mu`. You can find an example in `FileList`.

modules/renter/uploadheap.go
```diff
@@ -236,6 +238,30 @@ func (r *Renter) managedBuildChunkHeap(hosts map[string]struct{}) {
...
+			for cid := range contractIDs {
+				resolvedID := r.hostContractor.ResolveID(cid)
+				cu, ok := r.hostContractor.ContractUtility(resolvedID)
+				goodForRenew[cid] = ok && cu.GoodForRenew
+				offline[cid] = r.hostContractor.IsOffline(resolvedID)
+			}
+			// Check if local file is missing and redundancy is less than 1
+			// log warning to renter log
+			if _, err := os.Stat(tf.RepairPath); os.IsNotExist(err) && file.redundancy(offline, goodForRenew) < 1 {
+				r.log.Println("File not found on disk:", tf.RepairPath)
```
ChrisSchinnerl: I'd probably change that message. It is not uncommon that a file is not found on disk but what we would like to warn the user about are files that are not found on disk and are possibly unrecoverable.

modules/renter/uploadheap.go
```diff
@@ -236,6 +238,30 @@ func (r *Renter) managedBuildChunkHeap(hosts map[string]struct{}) {
 	// the heap.
 	id := r.mu.Lock()
 	for _, file := range r.files {
+		// check for local file
+		tf, exists := r.persist.Tracking[file.name]
+		if exists {
```
ChrisSchinnerl: If the path doesn't exist for some reason, we still want to check if the redundancy is below 1x and log a warning if that's not the case. 
@lukechampine @DavidVorick is it even possible that a file doesn't have an entry in the tracking map? maybe we should also add a sanity check and logging for that case.

modules/renter/uploadheap.go
ChrisSchinnerl: @lukechampine I don't think we need to move it outside the loop but we can get rid of the `if exists` by doing something like the following:
```
// check for local file
tf, exists := r.persist.Tracking[file.name]
if !exists {
    r.log.Critical("Some message about why that shouldn't happen")
}

// Build maps and compute redundancy

// Check if local file is missing and redundancy is less than 1
// log warning to renter log
if _, err := os.Stat(tf.RepairPath); (!exists || os.IsNotExist(err)) && file.redundancy(offline, goodForRenew) < 1 {
    r.log.Println("File not found on disk:", tf.RepairPath)
```

cmd/siac/rentercmd.go
```diff
@@ -540,6 +540,9 @@ func renterfileslistcmd() {
 				redundancyStr = "-"
 			}
 			uploadProgressStr := fmt.Sprintf("%.2f%%", file.UploadProgress)
+			if _, err := os.Stat(file.LocalPath); os.IsNotExist(err) && file.Redundancy < 1 {
+				uploadProgressStr = fmt.Sprintf("Local File Lost")
```
ChrisSchinnerl: Maybe we should print something like "File probably unrecoverable". It is very unlikely that it is recoverable but not 100% certain.

modules/renter/uploadheap.go
```diff
@@ -62,7 +64,7 @@ func (uch *uploadChunkHeap) Pop() interface{} {
 	return x
 }
 
-// managedAddChunkToUploadHeap will add a chunk to the upload heap.
+// managedPush will add a chunk to the upload heap.
 func (uh *uploadHeap) managedPush(uuc *unfinishedUploadChunk) {
```
ChrisSchinnerl: @DavidVorick should be break out of this method once we realize that a file is not available offline and has less than 1x redundancy? At the very least I would add a `todo` for it because once we scan our used hosts every hour or so, we shouldn't end up in a state where we reach <1x redundancy due to the host being temporarily offline.

cmd/siac/rentercmd.go
MSevey: I was trying to be concise as possible so that it doesn't look to weird in the table otherwise I would include both statements, "Local File Lost and File Probably Unrecoverable"

In this case I think it makes more sense to tell the user that the Local File is Lost since that is the failure more we are trying to catch and we would want the user to know the issue with the local file.  If the user just sees that the File is probably unrecoverable that wouldn't direct them to look at their locally file immediately.

ChrisSchinnerl: Maybe we should just change the header of the table to `Unrecoverable` and the entries to `yes` and `no`. That should be pretty concise.

MSevey: I can do that.  I think I'd want to add another column too then for `Local File Present` and have a `true` or `false`.  

That way the <1 redundancy would trigger the `unrecoverable` and the checking for the file on disk would trigger the `Local File Present`

MSevey: Doing it this way should also resolve #3067

ChrisSchinnerl: 'Local File Present' is also a bit long. Maybe just 'On Disk' or something like that? We can probably play around with the name and see what looks best.

MSevey: :+1:

modules/renter/uploadheap.go
lukechampine: it used to be, and it may be again in the future: if you loaded a shared .sia file, you wouldn't have a local file to track.

modules/renter/uploadheap.go
lukechampine: my understanding was that there would be some overlap between files in the `offline` and `goodForRenew` maps, since files share contracts. So if you move the construction outside the loop, you only calculate the status of each contract once, whereas if it's inside the loop, you end up recalculating the status of each contract multiple times.

MSevey: So would this be what you were thinking @lukechampine 
```golang
func (r *Renter) managedBuildChunkHeap(hosts map[string]struct{}) {
	// Loop through the whole set of files and get a list of chunks to add to
	// the heap.
	id := r.mu.RLock()
	goodForRenew := make(map[types.FileContractID]bool)
	offline := make(map[types.FileContractID]bool)
	for _, file := range r.files {
		file.mu.RLock()
		// Build maps that map contract id to its offline and goodForRenew
		// status
		contractIDs := make(map[types.FileContractID]struct{})
		for cid := range file.contracts {
			contractIDs[cid] = struct{}{}
		}
		for cid := range contractIDs {
			resolvedID := r.hostContractor.ResolveIDToPubKey(cid)
			cu, ok := r.hostContractor.ContractUtility(resolvedID)
			goodForRenew[cid] = ok && cu.GoodForRenew
			offline[cid] = r.hostContractor.IsOffline(resolvedID)
		}

		unfinishedUploadChunks := r.buildUnfinishedChunks(file, hosts)
		for i := 0; i < len(unfinishedUploadChunks); i++ {
			r.uploadHeap.managedPush(unfinishedUploadChunks[i])
		}
		file.mu.RUnlock()
	}
	for _, file := range r.files {
		file.mu.RLock()
		// check for local file
		tf, exists := r.persist.Tracking[file.name]
		if exists {
			// Check if local file is missing and redundancy is less than 1
			// log warning to renter log
			if _, err := os.Stat(tf.RepairPath); os.IsNotExist(err) && file.redundancy(offline, goodForRenew) < 1 {
				r.log.Println("File not found on disk and possibly unrecoverable:", tf.RepairPath)
			}
		}
		file.mu.RUnlock()
	}
	r.mu.RUnlock(id)
}
```

modules/renter/uploadheap.go
MSevey: @DavidVorick  thoughts?

modules/renter/uploadheap.go
MSevey: So should I change anything or leave as is?

modules/renter/uploadheap.go
```diff
@@ -235,14 +237,44 @@ func (r *Renter) buildUnfinishedChunks(f *file, hosts map[string]struct{}) []*un
...
+	offline := make(map[types.FileContractID]bool)
 	for _, file := range r.files {
+		file.mu.RLock()
+		// Build maps that map contract id to its offline and goodForRenew
+		// status
+		contractIDs := make(map[types.FileContractID]struct{})
+		for cid := range file.contracts {
+			contractIDs[cid] = struct{}{}
+		}
+		for cid := range contractIDs {
```
lukechampine: Is the `contractIDs` map still necessary? Can't you just iterate through `file.contracts`?

modules/renter/uploadheap.go
```diff
@@ -235,14 +237,38 @@ func (r *Renter) buildUnfinishedChunks(f *file, hosts map[string]struct{}) []*un
...
+		file.mu.RLock()
+		for cid := range file.contracts {
+			resolvedID := r.hostContractor.ResolveIDToPubKey(cid)
+			cu, ok := r.hostContractor.ContractUtility(resolvedID)
+			goodForRenew[cid] = ok && cu.GoodForRenew
+			offline[cid] = r.hostContractor.IsOffline(resolvedID)
+		}
+		file.mu.RUnlock()
+
 		unfinishedUploadChunks := r.buildUnfinishedChunks(file, hosts)
```
ChrisSchinnerl: This actually introduces a locking violation and a potential race condition. We can't call `build.UnfinishedChunks` without holding a write lock.

PR #3084, opened 2018-06-05
Add clear download history methods and endpoints

### LINKED ISSUE #3068: Add a renter endpoint to clear the download history
Currently the download history is only cleared upon a restart of siad. For most of the long-running nodes, the history is just sitting there in memory and using up RAM.
I'd suggest adding one or two api endpoints to the renter that let you clear the whole download history or just single entries. Those endpoints might also prove to be useful for GUIs to avoid cluttering the interface.

MSevey:
Added methods to clear the entire download history for the renter and remove specific downloads from the download history.  Both methods are accessible through the API.

resolves #3068


modules/renter/download.go
```diff
@@ -517,6 +517,27 @@ func (r *Renter) DownloadHistory() []modules.DownloadInfo {
...
+}
+
+// RemoveFromDownloadHistory removes a provided download from
+// the download history
+func (r *Renter) RemoveFromDownloadHistory(staticSiaPath string) error {
+	r.downloadHistoryMu.Lock()
+	defer r.downloadHistoryMu.Unlock()
+
+	delete(r.downloadHistory, staticSiaPath)
+	if _, ok := r.downloadHistory[staticSiaPath]; ok {
```
ChrisSchinnerl: I don't think that is necessary. I don't want to live in a world in which I can't trust `delete`.

modules/renter/download.go
```diff
@@ -491,11 +491,11 @@ func (r *Renter) DownloadHistory() []modules.DownloadInfo {
 	defer r.downloadHistoryMu.Unlock()
 
 	downloads := make([]modules.DownloadInfo, len(r.downloadHistory))
-	for i := range r.downloadHistory {
+	i := 0
+	for _, d := range r.downloadHistory {
```
ChrisSchinnerl: That won't work I think. `d` starts from the first element while you use `i` to traverse the list from the back. That means you don`t hole the correct lock. You wouldn't notice if the length is 1 but for a longer list that might cause problems.

modules/renter.go
```diff
@@ -335,9 +335,16 @@ type Renter interface {
 	// blocking, including downloads of `offset` and `length` type.
 	DownloadAsync(params RenterDownloadParameters) error
 
+	// ClearDownloadHistory clears the download history of the renter.
+	ClearDownloadHistory()
```
ChrisSchinnerl: We try to make sure that new interface methods for the modules return an error and then at the beginning we do:
```
if err := tg.Add(); err != nil {
  return err
}
defer tg.Done()
```

modules/renter/renter.go
```diff
@@ -182,8 +182,8 @@ type Renter struct {
 	// accessed in isolation.
 	//
 	// TODO: Currently the download history doesn't include repair-initiated
-	// downloads, and instead only contains user-initiated downlods.
-	downloadHistory   []*download
```
ChrisSchinnerl: I'm not sure if we should change the history to a map. That means you can only have 1 entry for each file even if you download it multiple times.

modules/renter/download.go
```diff
@@ -517,6 +517,27 @@ func (r *Renter) DownloadHistory() []modules.DownloadInfo {
 		} else {
 			downloads[i].Error = ""
 		}
+		i++
 	}
 	return downloads
 }
+
+// ClearDownloadHistory clears the renter's download history
+func (r *Renter) ClearDownloadHistory() {
```
ChrisSchinnerl: What I mentioned above about returning an error and `tg.Add` and `tg.Done`

modules/renter/download.go
```diff
@@ -517,6 +517,27 @@ func (r *Renter) DownloadHistory() []modules.DownloadInfo {
...
+	r.downloadHistoryMu.Lock()
+	defer r.downloadHistoryMu.Unlock()
+	r.downloadHistory = make(map[string]*download)
+}
+
+// RemoveFromDownloadHistory removes a provided download from
+// the download history
+func (r *Renter) RemoveFromDownloadHistory(staticSiaPath string) error {
```
ChrisSchinnerl: Same here

siatest/renter/renter_test.go
```diff
@@ -1203,3 +1204,77 @@ func TestRedundancyReporting(t *testing.T) {
...
+func testClearDownloadHistory(t *testing.T, tg *siatest.TestGroup) {
+	// Grab the first of the group's renters
+	r := tg.Renters()[0]
+
+	rdg, err := r.RenterDownloadsGet()
+	if err != nil {
+		t.Fatal("Could not get download history:", err)
+	}
+	numDownloads := len(rdg.Downloads)
+	if numDownloads == 0 {
```
ChrisSchinnerl: I'd declare a variable here that indicates how many downloads you want to see in the history. Probably 5 is fine. If there are fewer than 5 you add enough downloads to fill the gap.
I like the idea of reusing existing downloads.

siatest/renter/renter_test.go
```diff
@@ -1203,3 +1204,77 @@ func TestRedundancyReporting(t *testing.T) {
...
+		// Upload and download files to show build download history
+		var remoteFiles []*siatest.RemoteFile
+		for i := 0; i < 10; i++ {
+			dataPieces := uint64(1)
+			parityPieces := uint64(1)
+			fileSize := 100 + siatest.Fuzz()
```
ChrisSchinnerl: Instead of 10 this should be the difference between the number you would like to have and `numDownloads`.
I`d also move dataPieces and parityPieces out of the loop since they don't change between iterations.

siatest/renter/renter_test.go
```diff
@@ -1203,3 +1204,77 @@ func TestRedundancyReporting(t *testing.T) {
...
+			_, rf, err := r.UploadNewFileBlocking(fileSize, dataPieces, parityPieces)
```
ChrisSchinnerl: If we decide to keep a list instead of a map we should probably don`t have to upload additional files but instead just download the same files multiple times. For that we might be able to reuse existing uploaded files from previous tests.

modules/renter/renter.go
MSevey: I had done that mostly for the method that removes a particular download so that I didn't have to iterate over the entire slice to remove the download I was looking for.

Do you think that iteration is fine in this case?  Or do you have a better option in mind?

MSevey: Change to a map of slices

modules/renter/download.go
MSevey: update to check for timestamp as well as additional identifier

MSevey:
add endpoint to clear before and clear after

MSevey:
change download history back to slice and use binary search

modules/renter.go
```diff
@@ -335,9 +335,16 @@ type Renter interface {
...
+
 	// DownloadHistory lists all the files that have been scheduled for download.
 	DownloadHistory() []DownloadInfo
 
+	// RemoveFromDownloadHistory removes a given download from the download
+	// history of the renter.
+	RemoveFromDownloadHistory(staticSiaPath string) error
```
DavidVorick: not good enough to remove every instance of downloading a file I think, I think we'd rather be able to remove a specific instance in the download history.

That means we probably need to give downloads in the history a unique identifier, probably a `fastrand` generated value

MSevey: @DavidVorick I was going to use the time stamp unless you'd prefer that i add a unique identifier

MSevey:
Need to update documentation

node/api/client/renter.go
```diff
@@ -43,6 +43,30 @@ func (c *Client) RenterDownloadFullGet(siaPath, destination string, async bool)
...
 }
 
+// RenterClearDownloadsPost requests the /renter/downloads/clear resource
+func (c *Client) RenterClearDownloadsPost() (err error) {
+	err = c.post("/renter/downloads/clear", "", nil)
+	return
+}
+
+// RenterClearDownloadsAfterPost requests the /renter/downloads/clear/after/:timestamp resource
+func (c *Client) RenterClearDownloadsAfterPost(after int64) (err error) {
```
ChrisSchinnerl: These should all take `time.Time` as arguments as well.

node/api/renter.go
```diff
@@ -162,7 +162,7 @@ type (
 		EndTime              time.Time `json:"endtime"`              // The time when the download fully completed.
 		Error                string    `json:"error"`                // Will be the empty string unless there was an error.
 		Received             uint64    `json:"received"`             // Amount of data confirmed and decoded.
-		StartTime            time.Time `json:"starttime"`            // The time when the download was started.
+		StartTime            int64     `json:"starttime"`            // The time when the download was started.
```
ChrisSchinnerl: Same as mentioned above

modules/renter/download.go
```diff
@@ -520,3 +521,66 @@ func (r *Renter) DownloadHistory() []modules.DownloadInfo {
...
+// ClearDownloadHistoryAfter clears the download history after
+// the given timestamp
+func (r *Renter) ClearDownloadHistoryAfter(timestamp int64) error {
+	if err := r.tg.Add(); err != nil {
+		return err
+	}
+	defer r.tg.Done()
+	r.downloadHistoryMu.Lock()
+	defer r.downloadHistoryMu.Unlock()
+	i := sort.Search(len(r.downloadHistory), func(i int) bool { return r.downloadHistory[i].staticStartTime.UnixNano() == timestamp })
```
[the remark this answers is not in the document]
MSevey: I had written this from the mindset that the user would submit a timestamp of a specific download, in which case why wouldn't `timeA == timeB` work? 

If you think it makes more sense to make it more generic where it is just a user submitted date, then yes I would need to change it to `timeA >= timeB`

Which do you think makes more sense?
A) User submits the time stamp from a specific download from the download history
B) User submits arbitrary date

modules/renter.go
```diff
@@ -80,7 +80,7 @@ type DownloadInfo struct {
 	EndTime              time.Time `json:"endtime"`              // The time when the download fully completed.
 	Error                string    `json:"error"`                // Will be the empty string unless there was an error.
 	Received             uint64    `json:"received"`             // Amount of data confirmed and decoded.
-	StartTime            time.Time `json:"starttime"`            // The time when the download was started.
+	StartTime            int64     `json:"starttime"`            // The time when the download was started.
```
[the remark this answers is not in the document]
lukechampine: why do we need to provide a timestamp in the API in the first place? Seems easier to me to let the client do the parsing.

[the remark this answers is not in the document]
MSevey: we need some sort of unique identifier for the user to submit through the API to know what download to delete, so the unix timestamp seemed like an easy option.

lukechampine:
If the intent is to use `StartTimeUnix` as a unique identifier, I would rather just add an explicit ID. Something like:
```go
func (d DownloadInfo) ID() string {
    hashStr := crypto.HashAll(
        d.Destination,
        d.Length,
        d.Offset,
        d.SiaPath,
        d.StartTime,
    ).String()
    return hashStr[:12]
}
```

MSevey:
@lukechampine i used the unix timestamp as a unique identifier and then also as the parameter for the binary search.  Can I do a binary search on a hash?

lukechampine:
oh, I missed that we were clearing all of the downloads before/after a given timestamp. A hash wouldn't work for that.
Although, is this functionality that we need to implement on the backend? It wouldn't be hard for the UI to loop over the downloads, get the ID of each download it wants to clear, and then make a separate request for each ID. (If separate requests are too expensive, we can accept a list of IDs.) I prefer to keep the API flexible so that you can do things like "delete all downloads containing the string 'foo'."

MSevey:
@lukechampine we could make that a future update, if we wanted the UI to handle the different deletion requests and send them to a single API endpoint and we make that endpoint flexible enough to handle the different requests.  

If we did that it would go back to my original plan where I created a map for the Download history and we could use the Hash ID as the key.  Then the UI could just send lists of the IDs to the single endpoint that just removes them from the map. 

Let me know if you'd like me to:
A) Submit an issue for a future feature request on the UI to handle the different clear history request and leave this PR as is
or
B) Set up the API now where we have 1 endpoint to clear the entire history, and 1 endpoint that takes a slice of IDs and removes them from the download histroy

Option B I would still have to submit a feature request for the UI that it adds the functionality send a history deletion request

MSevey:
As discussed, I will make the following changes:

1) Simplify down to 1 API endpoint that takes a range of times.
2) API will handle blanks input from the user, valid API calls are the following
```
/renter/downloads/clear
/renter/downloads/clear?start=today
/renter/downloads/clear?end=today
/renter/downloads/clear?start=yesterday&end=today
/renter/downloads/clear?start=yesterday&end=yesterday
```
These correspond to:
- Clear all
- Clear from start time to end of download history
- Clear from start of download history to end time
- Clear downloads from start time to end time
- Clear specific download

doc/API.md
```diff
@@ -991,6 +992,29 @@ lists all files in the download queue.
...
+Clears the download history of the renter for a range
+of unix time stamps.  Both parameters are optional, if
+no parameters are provided, the entire download history
+will be cleared.  To clear a single download, provide the
+timestamp for the download as both parameters.  Since
+download history is displayed from newest to oldest, providing
+only the start parameter will clear all downloads before, or
+older than the start timestamp.  Conversely, providing only
+the end parameter will clear all downloads after, or
+newer than the end timestamp.
```
lukechampine: This sounds backwards. If I specify `clear?start=10`, then it should clear everything *newer* than (after) 10. If I specify `clear?end=10`, it should clear everything *older* than (before) 10.

also, we should specify that the time stamp is in nanoseconds, and that the nanosecond value is reported in `/downloads`. I'm not sure how well other languages support nanosecond timestamps, so in some cases the easiest thing to do might be to simply copy the timestamps from the starting and ending downloads in the range you want to clear.

modules/renter/download.go
```diff
@@ -506,6 +509,7 @@ func (r *Renter) DownloadHistory() []modules.DownloadInfo {
 			EndTime:              d.endTime,
 			Received:             atomic.LoadUint64(&d.atomicDataReceived),
 			StartTime:            d.staticStartTime,
+			StartTimeUnix:        d.staticStartTimeUnix,
```
lukechampine: couldn't this be calculated in place, i.e. `StartTimeUnix: d.staticStartTime.UnixNano()`? I don't see the need for a separate variable.

modules/renter/download.go
```diff
@@ -520,3 +524,53 @@ func (r *Renter) DownloadHistory() []modules.DownloadInfo {
...
+func (r *Renter) ClearDownloadHistory(start, end time.Time) error {
+	if err := r.tg.Add(); err != nil {
+		return err
+	}
+	defer r.tg.Done()
+	r.downloadHistoryMu.Lock()
+	defer r.downloadHistoryMu.Unlock()
+
+	// Clear download history if both start and end times are zero values
+	if start.Equal(time.Unix(0, 0)) && end.Equal(time.Unix(0, 0)) {
```
lukechampine: you can use `start.IsZero()` for this

modules/renter/download.go
```diff
@@ -520,3 +524,53 @@ func (r *Renter) DownloadHistory() []modules.DownloadInfo {
...
+		}
+		r.downloadHistory = r.downloadHistory[i:]
+		return nil
+	}
+
+	// Clear range of downloads from history
+	i := sort.Search(len(r.downloadHistory), func(i int) bool { return r.downloadHistory[i].staticStartTime.After(end) })
+	if i > len(r.downloadHistory) {
+		// End time after all downloads will result in an error
+		return errors.New("end time after all downloads")
```
lukechampine: I think it's fine for this to just be a no-op instead of returning an error.

modules/renter/download.go
```diff
@@ -520,3 +524,53 @@ func (r *Renter) DownloadHistory() []modules.DownloadInfo {
...
+		r.downloadHistory = r.downloadHistory[:0]
+		return nil
+	}
+
+	// If End is zero value, search for start time to clear history
+	if end.Equal(time.Unix(0, 0)) {
+		i := sort.Search(len(r.downloadHistory), func(i int) bool { return r.downloadHistory[i].staticStartTime.After(start) })
+		if i > len(r.downloadHistory) {
+			// Start time is After all downloads, clear entire download history
+			r.downloadHistory = r.downloadHistory[:0]
```
lukechampine: if `start` is after all downloads, I would expect nothing to be cleared

modules/renter/download.go
```diff
@@ -520,3 +524,53 @@ func (r *Renter) DownloadHistory() []modules.DownloadInfo {
...
+	if i > len(r.downloadHistory) {
+		// End time after all downloads will result in an error
+		return errors.New("end time after all downloads")
+	}
+	// linear search for start
+	if !start.Equal(time.Unix(0, 0)) && !start.After(r.downloadHistory[len(r.downloadHistory)-1].staticStartTime) {
+		for j := i - 1; j < len(r.downloadHistory); j++ {
+			if r.downloadHistory[j].staticStartTime.After(start) {
+				//remove section from array
+				r.downloadHistory = append(r.downloadHistory[:i-1], r.downloadHistory[j:]...)
```
lukechampine: I would rearrange this like so:
```go
j := i
for j < len(r.downloadHistory) && r.downloadHistory[j].staticStartTime.Before(end) {
    j++
}
r.downloadHistory = append(r.downloadHistory[:i], r.downloadHistory[j:]...)
```

node/api/client/renter.go
```diff
@@ -43,6 +44,46 @@ func (c *Client) RenterDownloadFullGet(siaPath, destination string, async bool)
...
+
+// RenterClearDownloadsRangePost requests the /renter/downloads/clear resource
+// with both start and end times provided
+func (c *Client) RenterClearDownloadsRangePost(start, end time.Time) (err error) {
+	values := url.Values{}
+	values.Set("start", strconv.FormatInt(start.UnixNano(), 10))
+	values.Set("end", strconv.FormatInt(end.UnixNano(), 10))
+	err = c.post("/renter/downloads/clear", values.Encode(), nil)
+	return
+}
```
lukechampine: not sure if we need four separate functions. Calling `RenterClearDownloads(start, time.Time{})` doesn't seem so bad. Also, should it be `RenterDownloadsClear` to match the route?

node/api/renter.go
```diff
@@ -319,6 +320,44 @@ func (api *API) renterContractsHandler(w http.ResponseWriter, _ *http.Request, _
...
+		endInt, err = strconv.ParseInt(endStr, 10, 64)
+		if err != nil {
+			WriteError(w, Error{"parsing integer value for parameter `end` failed: " + err.Error()}, http.StatusBadRequest)
+			return
+		}
+	}
+	endTime := time.Unix(0, endInt)
+
+	if startTime.Before(endTime) && !startTime.Equal(time.Unix(0, 0)) {
+		WriteError(w, Error{"start time can not be before end time"}, http.StatusBadRequest)
```
lukechampine: I would move this check inside the renter function, since we'll want to catch it in Go programs as well

doc/API.md
MSevey: I can rephrase it a little bit, but the intention is that the start and end define the range and since we display to the user the downloads from newest to oldest it makes sense to think of it as where do I start clearing the download history and where to I end clearing the download history.  So clear from the start onward or up to the end. Maybe just removing before and after would make it clearer, so something like this:

`Since the download history is displayed from newest to oldest, providing
only the start parameter will clear all downloads from the start timestamp
onward, which would be all downloads older than the start timestamp.  
Conversely, providing only the end parameter will clear all downloads up to
the end timestamp, which would be all downloads newer than the end timestamp.
`

modules/renter/download.go
MSevey: yea I can change that

modules/renter/download.go
MSevey: :+1:

modules/renter/download.go
MSevey: yea that is a typo, it originally said "end is after" so it should be "start is before"

modules/renter/download.go
MSevey: :+1:

modules/renter/download.go
MSevey: :+1:

node/api/client/renter.go
MSevey: When only one value is submitted, how to you recognize if it is the start or end?

node/api/renter.go
MSevey: :+1:

node/api/client/renter.go
ChrisSchinnerl: I think I actually recommended using multiple functions since I thought it might improve usability for developers. Just feels a bit more intuitive to me. Unfortunately go doesn't have default parameters :/

doc/API.md
lukechampine: I think I understand, but IMO it's surprising that the "start" timestamp would come after the "end" timestamp. It's true that when presented to the user, downloads are usually ordered newest to oldest, but that's a presentation choice. The user isn't the one crafting the API calls, the app developer is, so we should prioritize the developer and try to meet their expectations.

One possible solution would be to rename `start`/`end`/ to `oldest`/`newest`. Then the meaning is unambiguous.

modules/renter/download.go
MSevey: turns out `time.Unix(0, 0).IsZero()` is false, so I'll leave it as `time.Unix(0, 0)`

node/api/client/renter.go
lukechampine: you can sort of fake default parameters:
```go
func RenterClearDownloads(args ...time.Time) {
    if len(args) == 1 {
        // assume args[0] is start time
    } else if len(args) == 2 {
        // assume args[0] is start time, args[1] is end time
    } else {
        // invalid
    }
}
```
but this is still suboptimal (e.g. when you only want to specify the end time) and it's easier to make a mistake.

Having four separate functions makes me sad, but I don't see a better alternative, so let's stick with it for now.

modules/renter/download.go
```diff
@@ -519,3 +521,57 @@ func (r *Renter) DownloadHistory() []modules.DownloadInfo {
...
+// provided newest and oldest timestamps
+func (r *Renter) ClearDownloadHistory(newest, oldest time.Time) error {
+	if err := r.tg.Add(); err != nil {
+		return err
+	}
+	defer r.tg.Done()
+	r.downloadHistoryMu.Lock()
+	defer r.downloadHistoryMu.Unlock()
+
+	if newest.Before(oldest) && !newest.Equal(time.Unix(0, 0)) {
```
ChrisSchinnerl: I think there might be a bug in this code when searching for a timestamp between existing timestamps.
In my opinion this code is also a bit more complicated than it has to be. I think it would be easier to do it like [this](https://play.golang.org/p/r_6bVd_MU81). I used an `[]int` in the example but apart from that it should be pretty much the same. 
The `main` method tests a bunch of edge cases. If any of those are missing in this PR we should make sure that we add them.

node/api/client/renter.go
```diff
@@ -43,6 +44,46 @@ func (c *Client) RenterDownloadFullGet(siaPath, destination string, async bool)
 	return
 }
 
+// RenterClearAllDownloadsPost requests the /renter/downloads/clear resource
+// with no parameters
+func (c *Client) RenterClearAllDownloadsPost() (err error) {
+	values := url.Values{}
+	values.Set("newest", "")
```
ChrisSchinnerl: I think if you want the value to be `""` you don't need to set them at all. Same for the other methods

node/api/renter.go
```diff
@@ -319,6 +320,39 @@ func (api *API) renterContractsHandler(w http.ResponseWriter, _ *http.Request, _
 	})
 }
 
+// renterClearDownloadsHandler handles the API call to request to clear the download queue.
+func (api *API) renterClearDownloadsHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	var newestInt, oldestInt int64
+	var err error
+
+	newestStr, oldestStr := req.FormValue("newest"), req.FormValue("oldest")
+
```
ChrisSchinnerl: I think we don't need a newline between `FormValue` and `if`

modules/renter/download.go
```diff
@@ -519,3 +521,57 @@ func (r *Renter) DownloadHistory() []modules.DownloadInfo {
...
+		return errors.New("newest timestamp can not be before oldest timestamp")
+	}
+
+	// Clear download history if both newest and oldest timestamps are zero values
+	if newest.Equal(time.Unix(0, 0)) && oldest.Equal(time.Unix(0, 0)) {
```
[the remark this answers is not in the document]
MSevey: Luke said the same thing and I tested it but `time.Unix(0,0).IsZero()` equates to false so `newest.IsZero()` is always false.

modules/renter.go
```diff
@@ -349,6 +350,10 @@ type Renter interface {
 	// blocking, including downloads of `offset` and `length` type.
 	DownloadAsync(params RenterDownloadParameters) error
 
+	// ClearDownloadHistory clears the download history of the renter
+	// inclusive for before and after times.
+	ClearDownloadHistory(before, after time.Time) error
```
lukechampine: the order of these parameters is confusing to me, because `before` is more recent than `after`, whereas normally you specify a range of time with the older time first (e.g. you say "From 2017 to 2018," not the other way around).

modules/renter/download.go
```diff
@@ -519,3 +521,74 @@ func (r *Renter) DownloadHistory() []modules.DownloadInfo {
...
+		// Checking for before not equal to zero for case of only after
+		// is submitted but it is before all the timestamps
+		afterIndex = 0
+	}
+	if beforeIndex == len(r.downloadHistory) {
+		beforeIndex = len(r.downloadHistory) - 1
+	}
+
+	r.downloadHistory = append(r.downloadHistory[:afterIndex], r.downloadHistory[beforeIndex+1:]...)
+	return nil
```
lukechampine: this function seems suspiciously complex. Given the number of edge cases, I'm strongly inclined to do away with the binary search optimization stuff and rewrite this in as clear a style as possible, which for me means an  O(n) loop:
```go
if len(r.downloadHistory) == 0 {
	return nil
}
if after.IsZero() {
	after = r.downloadHistory[0]
}
if before.IsZero() {
	before = r.downloadHistory[len(r.downloadHistory)-1]
}
withinTimespan := func(t time.Time) bool {
	return (t.After(after) || t.Equal(after)) && (t.Before(before) || t.Equal(before))
}
filtered := r.downloadHistory[:0]
for _, d := range r.downloadHistory {
	if !withinTimespan(d.staticStartTime) {
		filtered = append(filtered, d)
	}
}
r.downloadHistory = filtered
return nil
```
thoughts?

modules/renter.go
MSevey: That makes sense, I can switch them.

modules/renter/download.go
MSevey: I do agree that the function is a little complex for what it is trying to do.  If you think it's fine to go with an O(n) loop in order for the function to be more readable I can make those updates.

lukechampine: ok, let's do it. We can always come back and optimize later if needed.

doc/API.md
```diff
@@ -991,6 +992,27 @@ lists all files in the download queue.
...
+#### /renter/downloads/clear [POST]
+
+Clears the download history of the renter for a range
+of unix time stamps.  Both parameters are optional, if
+no parameters are provided, the entire download history
+will be cleared.  To clear a single download, provide the
+timestamp for the download as both parameters.  Providing
+only the before parameter will clear all downloads older
+than the timestamp.  Conversely, providing only the after
+parameter will clear all downloads newer than the timestamp.
```
DavidVorick: is this block formatted correctly? Looks like either the earlier lines cut off too early, or the later lines cut off too late. You can use `gq` in vim to format a block of text to a configured width (default 80 on most distros)

doc/api/Renter.md
```diff
@@ -274,10 +275,30 @@ lists all files in the download queue.
...
+#### /renter/downloads/clear [POST]
+
+Clears the download history of the renter for a range
+of unix time stamps.  Both parameters are optional, if
+no parameters are provided, the entire download history
+will be cleared.  To clear a single download, provide the
+timestamp for the download as both parameters.  Providing
+only the before parameter will clear all downloads older
+than the timestamp.  Conversely, providing only the after
+parameter will clear all downloads newer than the timestamp.
```
DavidVorick: same

modules/renter/download.go
```diff
@@ -519,3 +520,49 @@ func (r *Renter) DownloadHistory() []modules.DownloadInfo {
...
+		return nil
+	}
+
+	// Timestamp validation
+	if before.Before(after) && !before.IsZero() {
+		return errors.New("before timestamp can not be newer then after timestamp")
+	}
+
+	// Clear download history if both before and after timestamps are zero values
+	if before.IsZero() && after.IsZero() {
```
DavidVorick: while it makes more sense in the api, I'm not sure that this makes sense in a function call. There's a difference in the api between having a field that you set to a 0 value and not including that field at all. Here, it's not intuitive that `ClearDownloadHistory(0,0)` is the way that you wipe out everything.

modules/renter/download.go
```diff
@@ -519,3 +520,49 @@ func (r *Renter) DownloadHistory() []modules.DownloadInfo {
...
+		r.downloadHistory = r.downloadHistory[:0]
+		return nil
+	}
+
+	// Find and return downloads that are not within the given range
+	if after.IsZero() {
```
DavidVorick: this is intuitive, clear everything after time 0 = clear everything up to the `before` amount

modules/renter/download.go
```diff
@@ -519,3 +520,49 @@ func (r *Renter) DownloadHistory() []modules.DownloadInfo {
...
+		after = r.downloadHistory[0].staticStartTime
+	}
+	if before.IsZero() {
```
DavidVorick: This is not intuitive, clear everything before time 0 should indicate that you clear nothing at all.

modules/renter/download.go
```diff
@@ -519,3 +520,49 @@ func (r *Renter) DownloadHistory() []modules.DownloadInfo {
...
+		before = r.downloadHistory[len(r.downloadHistory)-1].staticStartTime
+	}
+	withinTimespan := func(t time.Time) bool {
+		return (t.After(after) || t.Equal(after)) && (t.Before(before) || t.Equal(before))
+	}
+	filtered := r.downloadHistory[:0]
+	for _, d := range r.downloadHistory {
```
DavidVorick: Might be worth dropping a TODO here to turn this in to a binary search. We don't need it right now because nobody has such a long download history that it matters, but at some point for completeness this should be performing better than O(n). Perhaps that's an easy point for a future coder to pick up.

modules/renter/download.go
MSevey: If `before` or `after` `IsZero()` that means a value wasn't submitted so this is checking for that condition and handling it accordingly.  I can add more comments to the function to clarify that.  From a user's POV they would just be submitting a value for `before` or `after` if they wanted to use that functionality, they don't need to submit a zero value for the other parameter.

modules/renter/download.go
MSevey: Originally it was a binary search, but in order to handle all the edge cases it becomes a little over complicated and harder to read through and understand what the function is doing.  Luke and I had decided to intentionally change it to an O(n) implementation so that it was cleaner to read the function.

I can drop a TODO in if you would like and call out that this can be changed to a binary search to improve look up time but to weigh it against readability of the code.

DavidVorick: If you use a helper function to do the binary search, write some nice comments, and then have good unit testing around the binary search it shouldn't be a problem.

I don't think we need to put all of that coding effort in now, but I definitely disagree with the idea of hurting performance because the code looks bad. If the code looks bad, we can clean it up by breaking it apart, not by switching to a worse-performing algorithm.

modules/renter/download.go
MSevey: what about creating a const of type time called `infinity` and set it equal to `time.Unix(math.MaxInt64,0)`  which is equal to `219250468-12-04 15:30:07 +0000 UTC`

MSevey: This should be sufficient for infinity as long as we aren't around in the year 219250468

modules/renter/download.go
MSevey: This is roughly what the binary search implementation looked like:
https://play.golang.org/p/s6nwzcTeK_r

modules/renter/download.go
MSevey: also `time.Unix(0,math.MaxInt64)` is the year 2262 if we think that is sufficient and then we wouldn't have to worry about overflow errors if someone used `.UnixNano()` on the const

MSevey: - [ ] Create modules.EndOfTime const

DavidVorick:
Looks good to me.

PR #3087, opened 2018-06-07
Fix renter spending reporting and current period

### LINKED ISSUE #3077: Change allowance mechanics to always extend by `period` time.
Currently, when your contracts renew, they renew so that there is `period` time left. This can create some confusing spending mechanics, where the first time you create contracts, you pay for 3 months of storage, but each time you renew after that you pay for 6 weeks of storage. It also means that up to your full allowance gets spent every 6 weeks, and users might be expecting that allowance to cover 3 months of storage, because indeed it does cover 3 months of storage for the first iteration only.

We can clear up a lot of this confusion just by making the allowance extend by the full period instead of extending up to the full period of time remaining.

MSevey:
Updated `PeriodSpending` for renter spending with siatest

updated renter's `CurrentPeriod` and added siatest

The only time that I can think of that the spending might still not be reported accurately is during the maturity delay window, when the unspent allocated funds have not be released to the renter.

This resolves #3077

### DIFF
--- a/cmd/siac/export.go
+++ b/cmd/siac/export.go
@@ -35,7 +35,8 @@ func renterexportcontracttxnscmd(destination string) {
 		die("Could not retrieve contracts:", err)
 	}
 	var contractTxns []types.Transaction
-	for _, c := range cs.Contracts {
+	contracts := append(cs.Contracts, cs.OldContracts...)
+	for _, c := range contracts {
 		contractTxns = append(contractTxns, c.LastTransaction)
 	}
 	destination = abs(destination)
--- a/cmd/siac/main.go
+++ b/cmd/siac/main.go
@@ -15,7 +15,7 @@ var (
 	// Flags.
 	hostContractOutputType string // output type for host contracts
 	hostVerbose            bool   // display additional host info
-	initForce              bool   // destroy and reencrypt the wallet on init if it already exists
+	initForce              bool   // destroy and re-encrypt the wallet on init if it already exists
 	initPassword           bool   // supply a custom password when creating a wallet
 	renterListVerbose      bool   // Show additional info about uploaded files.
 	renterShowHistory      bool   // Show download history in addition to download queue.
--- a/cmd/siac/main.go
+++ b/cmd/siac/main.go
@@ -17,6 +17,7 @@ var (
 	hostVerbose            bool   // display additional host info
 	initForce              bool   // destroy and re-encrypt the wallet on init if it already exists
 	initPassword           bool   // supply a custom password when creating a wallet
+	renterAllContracts     bool   // Show all active and expired contracts
 	renterListVerbose      bool   // Show additional info about uploaded files.
 	renterShowHistory      bool   // Show download history in addition to download queue.
 )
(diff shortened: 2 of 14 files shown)


MSevey:
This should also resolve #2308  @DavidVorick please confirm

modules/renter/contractor/contractor_test.go
```diff
@@ -116,69 +114,6 @@ func (stubHostDB) ScoreBreakdown(modules.HostDBEntry) modules.HostScoreBreakdown
 	return modules.HostScoreBreakdown{}
 }
 
-// TestAllowancePeriodTracking verifies that the contractor tracks its current
-// period correctly as renewals occur.
-func TestAllowancePeriodTracking(t *testing.T) {
```
MSevey: My siatests cover this now

modules/renter/contractor/contracts.go
```diff
@@ -416,16 +417,37 @@ func (c *Contractor) threadedContractMaintenance() {
...
-			renewAmount := contract.TotalCost.Sub(contract.ContractFee).Sub(contract.TxnFee).Sub(contract.SiafundFee).Sub(contract.RenterFunds)
+			//
+			// In order to accurately fund contracts based on variable spending,
+			// the cost per block is calculated based on the total spent
+			// over the length of time that the contract was active before
+			// renewal.
+			oldContractSpent := contract.TotalCost.Sub(contract.ContractFee).Sub(contract.TxnFee).Sub(contract.SiafundFee).Sub(contract.RenterFunds)
+			oldContractLength := blockHeight - contract.StartHeight
+			if oldContractLength <= 1 {
+				oldContractLength = types.BlockHeight(1)
```
lukechampine: since these are unsigned ints, `== 0` is probably clearer, and makes it more obvious that we're trying to avoid a divide-by-zero panic.

cmd/siac/main.go
```diff
@@ -124,6 +125,7 @@ func main() {
 	renterAllowanceCmd.AddCommand(renterAllowanceCancelCmd)
 
 	renterCmd.Flags().BoolVarP(&renterListVerbose, "verbose", "v", false, "Show additional file info such as redundancy")
+	renterContractsCmd.Flags().BoolVarP(&renterAllContracts, "all", "A", false, "Show all expired contracts in addition to active contracts")
```
DavidVorick: is there a reason that we use captial A and capital H as the flags instead of the lowercase alternatives?

cmd/siac/export.go
```diff
@@ -35,7 +35,8 @@ func renterexportcontracttxnscmd(destination string) {
 		die("Could not retrieve contracts:", err)
 	}
 	var contractTxns []types.Transaction
-	for _, c := range cs.Contracts {
+	contracts := append(cs.Contracts, cs.OldContracts...)
+	for _, c := range contracts {
```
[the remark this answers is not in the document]
MSevey: if you're worried about breaking compatibility I can remove that and we can discuss what to do with a separate PR

cmd/siac/main.go
MSevey: lower case a is a global flag i think so I did capital A to avoid any potential conflicts. If we prefer lower case letters I can do `old` and `o` instead.

modules/renter/contractor/contractor.go
```diff
@@ -108,8 +110,23 @@ func (c *Contractor) PeriodSpending() modules.ContractorSpending {
 			spending.DownloadSpending = spending.DownloadSpending.Add(old.DownloadSpending)
 			spending.UploadSpending = spending.UploadSpending.Add(old.UploadSpending)
 			spending.StorageSpending = spending.StorageSpending.Add(old.StorageSpending)
+		} else if exist && old.EndHeight+host.WindowSize+types.MaturityDelay > c.blockHeight {
+			// Calculate funds that are being withheld in contracts
+			spending.WithheldFunds = spending.WithheldFunds.Add(old.RenterFunds)
+			// Record the largest window size for worst case when reporting the spending
+			if host.WindowSize >= spending.ReleaseBlock {
```
DavidVorick: Should this be old.EndHeight + host.WindowSize + types.MaturityDelay?

PR #3098, opened 2018-06-11
Update stream cache to prevent duplicate chunks being added

### LINKED ISSUE #3094: Critical error: streamHeap and streamMap are not the same length
I didn't actually notice anything wrong but I saw this in my terminal. Feel free to remove this bug report if it's nothing.

# BUG REPORT

goroutine 2887318 [running]:
runtime/debug.Stack(0x445e67, 0x0, 0xc45824dd68)
	/usr/local/go/src/runtime/debug/stack.go:24 +0xa7
runtime/debug.PrintStack()
	/usr/local/go/src/runtime/debug/stack.go:16 +0x22
github.com/NebulousLabs/Sia/build.Critical(0xc45824de10, 0x1, 0x1)
	/home/luke/go/src/github.com/NebulousLabs/Sia/build/critical.go:16 +0x93
github.com/NebulousLabs/Sia/modules/renter.(*streamCache).pruneCache(0xc42032af90, 0x1)
	/home/luke/go/src/github.com/NebulousLabs/Sia/modules/renter/streamcache.go:108 +0x18b
github.com/NebulousLabs/Sia/modules/renter.(*streamCache).Add(0xc42032af90, 0xc44ef76120, 0x8, 0xc49f75a000, 0x27ffee8, 0x3bffe5c)
	/home/luke/go/src/github.com/NebulousLabs/Sia/modules/renter/streamcache.go:80 +0x92
github.com/NebulousLabs/Sia/modules/renter.(*unfinishedDownloadChunk).threadedRecoverLogicalData(0xc467940000, 0x0, 0x0)
	/home/luke/go/src/github.com/NebulousLabs/Sia/modules/renter/downloadchunk.go:226 +0x78b
created by github.com/NebulousLabs/Sia/modules/renter.(*worker).managedDownload
	/home/luke/go/src/github.com/NebulousLabs/Sia/modules/renter/workerdownload.go:60 +0x42a
Critical error: streamHeap and streamMap are not the same length

**Environment**
* Sia version: 1.3.3
* OS: Ubuntu 16.04

MSevey:
Current code allowed the same chunk to be added to the Heap but it would not be added to the Map.  Updated so chunks can only be added once and updated unit test.

Resolves #3094

### DIFF
--- a/modules/renter/streamcache_test.go
+++ b/modules/renter/streamcache_test.go
@@ -103,29 +103,13 @@ func TestPruneCache(t *testing.T) {
 	}
 
 	// Confirm the same chunk won't be added if already added
-	id := sc.streamHeap[0].id
+	sc.pruneCache(0)
+	id := "test"
 	for i := 0; i < 5; i++ {
 		sc.Add(id, []byte{})
 	}
-
-	// Manually remove ID and confirm it no longer exists in the Heap of Cache
-	cd := heap.Pop(&sc.streamHeap).(*chunkData)
-	if cd.id != id {
-		t.Fatal("Wrong chunk popped from Heap")
-	}
-	delete(sc.streamMap, cd.id)
-
-	// Check for any duplicate chunks
-	for len(sc.streamHeap) > 0 {
-		cd = heap.Pop(&sc.streamHeap).(*chunkData)
-		if cd.id == id {
-			t.Fatal("Duplicate ID found")
-		}
-	}
-	for k := range sc.streamMap {
-		if k == id {
-			t.Fatal("Duplicate ID found")
-		}
+	if len(sc.streamHeap) != 1 || len(sc.streamMap) != 1 {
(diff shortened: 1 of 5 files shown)


modules/renter/streamcache_test.go
```diff
@@ -102,6 +102,31 @@ func TestPruneCache(t *testing.T) {
 		t.Error("Cache size was changed by pruning to larger value")
 	}
 
+	// Confirm the same chunk won't be added if already added
+	id := sc.streamHeap[0].id
+	for i := 0; i < 5; i++ {
+		sc.Add(id, []byte{})
+	}
+
+	// Manually remove ID and confirm it no longer exists in the Heap of Cache
```
ChrisSchinnerl: I think it would be easier to just add the same id 5 times and check if the size of the heap and map increased by 1 each. That way you don't need to pop entries repeatedly and compare all the ids since you can be confident that it has only been added once anyway.

PR #3099, opened 2018-06-11
use one const for min acceptable peer version

DavidVorick:


DavidVorick:
This PR also fixes a bug where we allowed v1.3.0 peers to join the network in certain cases.

PR #3102, opened 2018-06-12
Change MaxUint64 to MaxInt64 in wallettransactionscmd

eddiewang:
Changes `math.MaxUint64` to `math.MaxInt64` so that the handler does not overflow when running parseInt on the end height.

This bug showed up due to merging PR #3097 - which is a necessary modification to maintain `-1` api compatibility with older versions of Sia-UI.

### DIFF
--- a/cmd/siac/walletcmd.go
+++ b/cmd/siac/walletcmd.go
@@ -467,7 +467,7 @@ func walletsweepcmd() {
 // wallettransactionscmd lists all of the transactions related to the wallet,
 // providing a net flow of siacoins and siafunds for each.
 func wallettransactionscmd() {
-	wtg, err := httpClient.WalletTransactionsGet(0, math.MaxUint64)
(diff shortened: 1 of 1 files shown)


lukechampine:
this makes me think that parsing as an int is probably the wrong thing to do. It wouldn't surprise me if a user assumed they could pass `^types.BlockHeight(0)`. Perhaps instead we should parse as a uint and check explicitly for `FormValue("endheight") == "-1"`.

DavidVorick:
Heh, do we want to make that change? I feel like it's the better choice as well.

Heights are always uint64, we should probably be checking the form value for a violation and then cleaning it up api-package side

ChrisSchinnerl:
@DavidVorick Nope, we don't. That's why we already changed it in a new [PR](https://github.com/NebulousLabs/Sia/pull/3107) :P

Issue #3106, opened 2018-06-13
Change wallet/transactions endpoint to explicitly check for -1


ChrisSchinnerl:
I think @lukechampine made a good point in [#3102](https://github.com/NebulousLabs/Sia/pull/3102) and we should probably implement that change. 
A user using the client library wouldn't expect that they can't pass any `BlockHeigh` for the `endHeight` in `WalletTransactionsGet`. So we should change `walletTransactionsHandler` in `node/api/wallet.go` to explicitly check for the `-1` string and parse the number using `strconv.ParUint(endheightStr, 10, 64)`

PR #3107, opened 2018-06-14
Explicitly check for -1 in WalletTransactionsHandler

ChrisSchinnerl:
Resolves [#3106](https://github.com/NebulousLabs/Sia/issues/3106).


node/api/wallet.go
```diff
@@ -555,15 +555,15 @@ func (api *API) walletTransactionsHandler(w http.ResponseWriter, req *http.Reque
 	// Parse endheightStr as an Int to account for negative numbers. If endInt is
 	// negative, explicitly set var end to MaxUint64.
```
lukechampine: comments needs updating

PR #3110, opened 2018-06-15
Add async flag to siac renter download

### LINKED ISSUE #2209: [feature] siac renter download allow to queue download async
Add a paramater to siac download to allow it to trigger the download to siad and then return without waiting for the file to be downloaded. There already is support to view status with siac renter downloads.

MSevey:
resolves #2209

### DIFF
--- a/cmd/siac/rentercmd.go
+++ b/cmd/siac/rentercmd.go
@@ -536,12 +536,11 @@ func renterfilesdownloadcmd(path, destination string) {
 	destination = abs(destination)
 	var err error
 	done := make(chan struct{})
-	if !renterDownloadAsync {
-		go downloadprogress(done, path)
-	}
-	err = httpClient.RenterDownloadFullGet(path, destination, renterDownloadAsync)
-	close(done)
-
+	go func() {
+		err = httpClient.RenterDownloadFullGet(path, destination, renterDownloadAsync)
+		close(done)
+	}()
+	downloadprogress(done, path)
 	if err != nil {
 		die("Could not download file:", err)
 	}
@@ -577,7 +576,6 @@ func downloadprogress(done chan struct{}, siapath string) {
 			fmt.Printf("\rDownloading... %5.1f%% of %v, %v elapsed, %.2f Mbps    ", pct, filesizeUnits(int64(d.Filesize)), elapsed, mbps)
 		}
 	}
-
 }
 
 // bySiaPath implements sort.Interface for [] modules.FileInfo based on the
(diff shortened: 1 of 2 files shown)


ChrisSchinnerl:
Does that still print the progress to stdout? If it is so fast that it is not noticeable we can probably leave it like that, but otherwise we can probably just avoid spawning the goroutine if `renterDownloadAsync` equals `true`.

DavidVorick:
Wouldn't it be annoying for the progress to be printed to stdout if you were trying to do async download?

DavidVorick:
Yeah I think sufficient to just skip the goroutine that prints out the progress

tbenz9:
I agree it shouldn't print progress, but I think it might be nice to print a message when the file is done downloading.

MSevey:
Ok so should it just be broken into an if statement then like this:
```
var err error
	if renterDownloadAsync {
		err = httpClient.RenterDownloadFullGet(path, destination, renterDownloadAsync)
	} else {
		done := make(chan struct{})
		go downloadprogress(done, path)

		err = httpClient.RenterDownloadFullGet(path, destination, renterDownloadAsync)
		close(done)
	}
```

ChrisSchinnerl:
@MSevey you can do:
```
done := make(chan struct{})
if !renterDownloadAsync { 
  go downloadprogress(done, path)
}
err = httpClient.RenterDownloadFullGet(path, destination, renterDownloadAsync)
close(done)
```
It doesn't really hurt to not use a channel.

MSevey:
Ok great

cmd/siac/rentercmd.go
```diff
@@ -470,11 +470,14 @@ func renterfilesdeletecmd(path string) {
...
 	done := make(chan struct{})
-	go downloadprogress(done, path)
-
-	err := httpClient.RenterDownloadFullGet(path, destination, false)
+	if !renterDownloadAsync {
+		go downloadprogress(done, path)
+	}
+	err = httpClient.RenterDownloadFullGet(path, destination, renterDownloadAsync)
 	close(done)
+
```
ChrisSchinnerl: Can you remove this newline? We want to make sure the line in which the `err` is assigned is close to the `if` that checks the error.

lukechampine:
I don't see much advantage to this over `siac renter download >/dev/null &`, but I guess that isn't as easy to do on Windows.

It might make sense to invert the goroutines here:
```go
go func() {
    err = httpClient.RenterDownloadFullGet(path, destination, renterDownloadAsync)
    close(done)
}()
downloadprogress(done, path)
```
since arguably it's the download that's happening asynchronously, not the printing.
In general, when structuring goroutines, I think it's best for the longest-lived goroutine to be in the main body. `downloadprogress` by definition outlives `RenterDownloadFullGet`, so putting it on the main code path ensures that all other goroutines have exited before the function returns. Whereas the way it's currently written, the `downloadprogress` goroutine could outlive `renterfilesdownloadcmd`. (Not a huge deal in this case, since the program is going to exit immediately afterward, but you get the idea.)

PR #3112, opened 2018-06-18
Change siac wallet send siacoins documentation to remove the legacy 10 SC fee

Fornax96:
The 10 SC transaction fee was replaced with a dynamic fee a long time ago, but the siac documentation still mentioned the old fee.


lukechampine:
thanks!

PR #3113, opened 2018-06-20
[WIP] Sia file format

ChrisSchinnerl:
This is still missing a few things but I think it's ready for a first round of reviews.


modules/renter/siafile/siafile.go
```diff
@@ -0,0 +1,180 @@
...
+		MerkleRoot: merkleRoot,
+	})
+	return nil
+}
+
+// ErasureCode returns the erasure coder used by the file.
+func (sf *SiaFile) ErasureCode() modules.ErasureCoder {
+	sf.mu.RLock()
+	sf.mu.RUnlock()
+	return sf.erasureCode
```
lukechampine: Should be a defer here (also missing on a few other functions)

modules/renter/siafile/siafile.go
```diff
@@ -0,0 +1,180 @@
...
+func New(siaPath string, erasureCode modules.ErasureCoder, pieceSize, fileSize uint64) *SiaFile {
+	file := &SiaFile{
+		metadata: Metadata{
+			fileSize:  int64(fileSize),
+			masterKey: crypto.GenerateTwofishKey(),
+			pieceSize: pieceSize,
+			siaPath:   siaPath,
+		},
+		erasureCode: erasureCode,
+		uid:         base32.StdEncoding.EncodeToString(fastrand.Bytes(20))[:20],
```
lukechampine: What is this used for? Why is it base 32? Why do you slice it to the first 20 bytes?

modules/renter/siafile/siafile.go
```diff
@@ -0,0 +1,180 @@
...
+// AddPiece adds an uploaded piece to the file. It also updates the host table
+// if the public key of the host is not aleady known.
+func (sf *SiaFile) AddPiece(pk types.SiaPublicKey, chunkIndex, pieceIndex uint64, merkleRoot crypto.Hash) error {
+	sf.mu.Lock()
+	defer sf.mu.Unlock()
+
+	// Get the index of the host in the public key table.
+	tableIndex := -1
+	for i, hpk := range sf.pubKeyTable {
+		if reflect.DeepEqual(hpk, pk) {
```
lukechampine: Reflection is Overkill and always makes me uneasy; I'd prefer `hpk.Algorithm == pk.Algorithm && bytes.Equal(hpk.Key, pk.Key)`.

modules/renter/siafile/siafile.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+// Pieces returns all the pieces for a chunk in a slice of slices that contains
+// all the pieces for a certain index.
+func (sf *SiaFile) Pieces(chunkIndex uint64) ([][]Piece, error) {
+	sf.mu.RLock()
+	defer sf.mu.RUnlock()
+	if chunkIndex >= uint64(len(sf.chunks)) {
+		return nil, fmt.Errorf("index %v out of bounds (%v)",
+			chunkIndex, len(sf.chunks))
```
lukechampine: it seems reasonable to panic here, since it's similar to a slice out of bounds

modules/renter/siafile/siafile.go
ChrisSchinnerl: I don't know. It's just the way we have done it before :P

modules/renter/siafile/metadata.go
```diff
@@ -0,0 +1,288 @@
...
+)
+
+type (
+	// Metadata is the metadata of a SiaFile and is JSON encoded.
+	Metadata struct {
+		version      [16]byte          // version of the sia file format used
+		fileSize     int64             // total size of the file
+		masterKey    crypto.TwofishKey // masterkey used to encrypt pieces
+		pieceSize    uint64            // size of a single piece of the file
+		trackingPath string            // file to the local copy of the file used for repairing
```
DavidVorick: probably better to use a name like `localPath`

modules/renter/siafile/metadata.go
```diff
@@ -0,0 +1,288 @@
...
+		siaPath      string
```
DavidVorick: I think it's worth having a comment here too, someone completely new to the codebase is not going to understand immediately the difference between the siapath and the localpath.

modules/renter/siafile/metadata.go
```diff
@@ -0,0 +1,288 @@
...
+
+		// The following fields are the usual unix timestamps of files.
+		modTime    time.Time // time of last content modification
+		changeTime time.Time // time of last metadata modification
+		accessTime time.Time // time of last access
```
DavidVorick: are we intending to have the accessTime and the downloadTime be the same?

My guess is not, does that mean that we should add a downloadTime as well?

modules/renter/siafile/metadata.go
```diff
@@ -0,0 +1,288 @@
...
+
+		// File ownership/permission fields.
+		mode os.FileMode // unix filemode of the sia file - uint32
+		uid  int         // id of the user who owns the file
+		gid  int         // id of the group that owns the file
+
+		// chunkHeaderSize is the size of each of the following chunk's metadata.
+		chunkHeaderSize uint64
+		// chunkBodySize is the size of each of the following chunk's bodies.
+		chunkBodySize uint64
```
DavidVorick: these two fields and comments aren't perfectly clear.

IF I understand correctly, the chunkHeaderSize is the amount of data in the siafile that is allocated to each chunk, which tells you how to perform random access within the siafile to read and modify chunk metadata.

And then chunkBodySize is the size of the logical data within the chunk so that you know what part of the logical file each chunk index maps to?

modules/renter/siafile/metadata.go
```diff
@@ -0,0 +1,288 @@
...
+
+		// The following fields are the offsets for data that is written to disk
+		// after the pubKeyTable. We reserve a generous amount of space for the
+		// table and extra fields, but we need to remember those offsets in case we
+		// need to resize later on.
+		//
+		// chunkOffset is the offset of the first chunk, forced to be a factor of
+		// 4096, default 16kib
```
DavidVorick: are you sure that a default of 4096 is not good enough? A default of 16kib means that there's going to be a lot of metadata overhead for tiny files.

So much actually that it makes sense to have really small files (<64kb) just to be stored entirely instead of the chunks, which I think is probably the best solution here, heh.

modules/renter/siafile/metadata.go
```diff
@@ -0,0 +1,288 @@
...
+		//
+		// pubKeyTableOffset is the office of the publicKeyTable within the
```
DavidVorick: can I run for `office of the publicKeyTable`? :p

modules/renter/siafile/metadata.go
```diff
@@ -0,0 +1,288 @@
...
+		// file.
+		//
+		chunkOffset       int64
+		pubKeyTableOffset int64
+	}
+)
+
+// Available indicates whether the file is ready to be downloaded.
+func (sf *SiaFile) Available(offline map[string]bool) bool {
```
DavidVorick: This function probably does not belong in the metadata file

modules/renter/siafile/siafile.go
```diff
@@ -0,0 +1,181 @@
...
+	// SiaFile is the disk format for files uploaded to the Sia network.  It
+	// contains all the necessary information to recover a file from its hosts and
+	// allows for easy constant-time updates of the file without having to read or
+	// write the whole file.
+	SiaFile struct {
+		// metadata is the mostly static metadata of a SiaFile. The reserved
+		// size of the metadata on disk should always be a multiple of 4kib.
+		// The metadata is also the only part of the file that is JSON encoded
+		// and can therefore be easily extended.
+		metadata Metadata
```
DavidVorick: @lukechampine @ChrisSchinnerl what do you think about embedding the `Metadata` field into the SiaFile struct, that way every method and field of the metadata can be accessed directly?

modules/renter/siafile/siafile.go
DavidVorick: since the uid is not persisted ever, we can just use fastrand.Bytes, there's no need to encode it.

I've been wondering whether we should persist the uid, but I think that could cause problems.

And, if we aren't going to persist the uid but instead keep everything in memory, another approach would be to use a uid counter and just count up each time that we add a file, because all we are really looking for is uniqueness, and it doesn't need any sort of collision resistance or security from an external party.

DavidVorick: the downside to a counter of course is that you need to keep a counter somewhere.

modules/renter/siafile/siafile.go
```diff
@@ -0,0 +1,181 @@
...
+		binary.LittleEndian.PutUint32(chunks[i].erasureCodeParams[4:8], uint32(erasureCode.NumPieces()-erasureCode.MinPieces()))
+		chunks[i].pieces = make([][]Piece, erasureCode.NumPieces())
+	}
+	file.chunks = chunks
+	return file
+}
+
+// AddPiece adds an uploaded piece to the file. It also updates the host table
+// if the public key of the host is not aleady known.
+func (sf *SiaFile) AddPiece(pk types.SiaPublicKey, chunkIndex, pieceIndex uint64, merkleRoot crypto.Hash) error {
```
DavidVorick: should add a TODO here so that we remember to add persistence changes when we move the persistence over.

modules/renter/siafile/siafile.go
```diff
@@ -0,0 +1,181 @@
...
+
+	"github.com/NebulousLabs/Sia/crypto"
+)
+
+type (
+	// SiaFile is the disk format for files uploaded to the Sia network.  It
+	// contains all the necessary information to recover a file from its hosts and
+	// allows for easy constant-time updates of the file without having to read or
+	// write the whole file.
+	SiaFile struct {
```
DavidVorick: we should be able to make a bunch of the fields static for both the SiaFile and the Metadata, right? then we can minimize lock contention when working with the files, something that for some situations may end up being fairly serious.

modules/renter/siafile/siafile.go
```diff
@@ -0,0 +1,181 @@
...
+		// pieces are the pieces of the file the chunk consists of.
+		pieces [][]Piece
+	}
+
+	// Piece represents a single piece of a chunk on disk
+	Piece struct {
+		KeyNonce   [4]byte            // nonce used for encrypting the piece
+		HostPubKey types.SiaPublicKey // public key of the host
+		MerkleRoot crypto.Hash        // merkle root of the piece
+	}
```
DavidVorick: so, this is something we aren't going to be able to do soon, but I do think that we are going to at some point need a mapping in the contractor of all of the pieces that are in each contract. it'll probably need to be on disk, but it's the only way I can think of to be sure that we actually have the data still in the contract, especially once we build in the rpc to allow hosts to tell us that they've lost some pieces (in the event of a corrupted disk, for example) but not all of them in the contract.

Once we have that mapping, we genuinely won't need 32 bytes of the merkle root per piece, we'll be able to safely trim it down to about 10 bytes and still have high confidence of correctness, because every piece in our contracts was of our own choosing, and if there's ever a conflict / collision we can just change the key nonce for the piece and try again.

DavidVorick: On the note of key nonces, a disadvantage of this strategy is that you can't modify only part of a chunk, if you are going to modify a file on Sia you have to modify the whole chunk to stay secure. I'm not sure if there are encryption strategies that allow you to perform safe in-place modification without needing a ton of nonce overhead though.

modules/renter/siafile/metadata.go
ChrisSchinnerl: I think just having the `accessTime` is fine since that is also the way Unix does it. `modTime` is bascially a subset of `accessTime` since it only includes uploads. 
If you just want to know the last time you manually downloaded a file, you can still take a look at the download history.

modules/renter/siafile/metadata.go
ChrisSchinnerl: Hm I'm actually not sure what does fields were about anymore :P I don't think we need both of them. We can always figure out the logical size of a chunk using the erasure coding and the pieceSize.

modules/renter/siafile/metadata.go
ChrisSchinnerl: I think that is just a possible value we talked. We can also do 4kib instead.

modules/renter/siafile/siafile.go
ChrisSchinnerl: The reason why I didn't do that is because the `Metadata` will be JSON encoded. That way we can just pass the `Metadata` field directly to the encoder. Otherwise we need a separate struct for that.

modules/renter/siafile/siafile.go
```diff
@@ -0,0 +1,181 @@
...
+		// Since multiple pieces from different chunks might be uploaded to the same host, this
+		// allows us to deduplicate the rather large public keys.
+		pubKeyTable []types.SiaPublicKey
+
+		// chunks are the chunks the file was split into.
+		chunks []Chunk
+
+		// utility fields. These are not persisted.
+		deleted     bool
+		erasureCode modules.ErasureCoder
```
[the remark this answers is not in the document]
ChrisSchinnerl: That's a good point. I'll change it so that it always uses `siafile.chunks[0].erasureCode` but we need to think about how we would like to report the redundancy on a file that has different settings for each chunk.

modules/renter/siafile/siafile.go
```diff
@@ -0,0 +1,181 @@
...
+		// erasureCodeType specifies the algorithm used for erasure coding
+		// chunks. Available types are:
+		//   0 - Invalid / Missing Code
+		//   1 - Reed Solomon Code
+		//
+		// erasureCodeParams specifies possible parameters for a certain
+		// erasureCodeType. Currently params will be parsed as follows:
+		//   Reed Solomon Code - 4 bytes dataPieces / 4 bytes parityPieces
+		//
+		erasureCodeType   [4]byte
```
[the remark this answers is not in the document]
ChrisSchinnerl: Probably not. It's more of a placeholder right now though so we can change that in the persistence PR together with other persistence related fields.

modules/renter/siafile/siafile.go
ChrisSchinnerl: Doesn't XTS let you do that without a nonce? I think it is used by Microsoft in Bitlocker.

modules/renter/siafile/metadata.go
DavidVorick: hmm.

Doesn't a download count as an access though? You did technically access the file if you performed a download.

And, are we treating these only as applies to the logical data? Because that means that repairing would not update any of these files at all (which I think is correct)

ChrisSchinnerl: Yeah, a download would be an access. Once we support updating the file we would also reset `modTime` when we do so.
I'm not sure about repairing since it also downloads the file. Would make sense for it not to change the timestamp though.

modules/renter/siafile/siafile.go
ChrisSchinnerl: Yes, I wanted to do that once everything else in this PR looks good to you and won't be changed anymore.

modules/renter/persist.go
```diff
@@ -174,26 +175,26 @@ func (f *file) UnmarshalSia(r io.Reader) error {
 }
 
 // saveFile saves a file to the renter directory.
-func (r *Renter) saveFile(f *file) error {
-	if f.deleted {
+func (r *Renter) saveFile(f *siafile.SiaFile) error {
+	if f.Deleted() { // TODO: violation of locking convention
```
DavidVorick: hmm. I'm not sure there's an easy path to fix this.

Typically you'd just turn `saveFile` into `managedSaveFile`, but I think that there are places where we need to keep holding the lock before and after we call `saveFile` to be safe.

modules/renter/siafile/siafile.go
```diff
@@ -0,0 +1,282 @@
...
+		// to indicate if a chunk is special.
+		extensionInfo [16]byte
+
+		// pieces are the pieces of the file the chunk consists of.
+		pieces [][]Piece
+	}
+
+	// Piece represents a single piece of a chunk on disk
+	Piece struct {
+		KeyNonce   [4]byte            // nonce used for encrypting the piece
```
DavidVorick: We don't need the KeyNonce anymore now that we have decided to switch to threefish

modules/renter/siafile/metadata.go
```diff
@@ -0,0 +1,199 @@
...
+	"github.com/NebulousLabs/Sia/types"
+)
+
+type (
+	// Metadata is the metadata of a SiaFile and is JSON encoded.
+	Metadata struct {
+		staticVersion   [16]byte          // version of the sia file format used
+		staticFileSize  int64             // total size of the file
+		staticMasterKey crypto.TwofishKey // masterkey used to encrypt pieces
+		staticPieceSize uint64            // size of a single piece of the file
```
DavidVorick: We need 2 more fields for encryption, the first field needs to indicate the type of encryption, and the second field needs to be a backup field

DavidVorick: *a backup key. Siafiles will have 2 keys, one that you give out when you share the file and one that you don't give out. The second key prevents hosts from deduplicating files that you've shared, because they can't decrypt the pieces you didn't give out the second key for.

Then I think at the chunk level we'll need a byte indicating what piece index starts using the backup key instead of the master key for that chunk.

modules/renter/siafile/siafile.go
```diff
@@ -0,0 +1,282 @@
...
+	}
+
+	// If the redundancy is smaller than 1x we return the redundancy that
+	// includes contracts that are not good for renewal. The reason for this is
+	// a better user experience. If the renter operates correctly, redundancy
+	// should never go above numPieces / minPieces and redundancyNoRenew should
+	// never go below 1.
+	if minRedundancy < 1 {
+		return minRedundancyNoRenew
+	}
```
DavidVorick:
```go
if minRedundancy < 1 && minRedundancyNoRenew >= 1 {
    return 1
}
```

modules/renter/upload.go
```diff
@@ -20,13 +20,31 @@ import (
...
 	"github.com/NebulousLabs/Sia/modules"
+	"github.com/NebulousLabs/Sia/modules/renter/siafile"
 )
 
 var (
 	// errUploadDirectory is returned if the user tries to upload a directory.
 	errUploadDirectory = errors.New("cannot upload directory")
 )
 
+// newFile is a helper to more easily create a new Siafile for testing.
```
DavidVorick: is this comment correct? Is this function in the right file? This isn't a testing file.

modules/renter/uploadheap.go
```diff
@@ -163,59 +167,54 @@ func (r *Renter) buildUnfinishedChunks(f *file, hosts map[string]struct{}) []*un
...
-	//
-	// TODO / NOTE: This process isn't going to make sense anymore once we
-	// switch to chunk-based saving.
-	if saveFile {
-		err := r.saveFile(f)
+
+	// Iterate through the pieces of the file and mark which hosts are already
+	// in use for the chunk. As you delete hosts from the 'unusedHosts' map,
+	// also increment the 'piecesCompleted' value.
+	for chunkIndex := uint64(0); chunkIndex < f.NumChunks(); chunkIndex++ {
```
DavidVorick: This comment is no longer correct, since the file format has changed.

modules/renter/uploadheap.go
```diff
@@ -163,59 +167,54 @@ func (r *Renter) buildUnfinishedChunks(f *file, hosts map[string]struct{}) []*un
...
+				redundantPiece := newUnfinishedChunks[chunkIndex].pieceUsage[pieceIndex]
+				if exists && !redundantPiece {
+					newUnfinishedChunks[chunkIndex].pieceUsage[pieceIndex] = true
+					newUnfinishedChunks[chunkIndex].piecesCompleted++
+					delete(newUnfinishedChunks[chunkIndex].unusedHosts, pk.String())
+				} else if exists {
+					// This host has a piece, but it is the same piece another host
+					// has. We should still remove the host from the unusedHosts
+					// since one host having multiple pieces of a chunk might lead
+					// to unexpected issues.
```
DavidVorick: Can you expand this comment to explain that if we leave the host in the set of unused hosts, it's possible that one host gets multiple pieces, then the other redundant hosts on those pieces go offline, and then you have one host providing false redundancy?

modules/renter/siafile/metadata.go
ChrisSchinnerl: Does it make sense to have partial file sharing?

modules/renter/siafile/siafile.go
```diff
@@ -0,0 +1,282 @@
...
+		extensionInfo [16]byte
+
+		// pieces are the pieces of the file the chunk consists of.
+		pieces [][]Piece
+	}
+
+	// Piece represents a single piece of a chunk on disk
+	Piece struct {
+		KeyNonce   [4]byte            // nonce used for encrypting the piece
+		HostPubKey types.SiaPublicKey // public key of the host
```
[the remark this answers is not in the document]
ChrisSchinnerl: Can we change that in the PR that changes the persistence? Right now it would only make the code in `compat.go` more complicated since the table doesn't have a fixed ordering.

PR #3115, opened 2018-06-22
Siatest persistence cleanup

ChrisSchinnerl:
This PR cleans up the madness in the `/tmp/SiaTesting` folder for the `siatest` framework.
`siatest`- Tests will create the following folder structure in the `/tmp/SiaTesting` directory.

`/tmp/SiaTesting/siatest/<packageName>/<testName>/<nodeName>`

The <nodeName> is the the folder name specified in the `Dir` field of the `NodeParams` struct.
If no `Dir` field is specified (e.g. when creating a whole group at once) the node name will be an encoded string of the enabled modules plus a random 4 byte, hex-encoded suffix. So for example a renter node could have the name `gctwr-deadbeef`.

One exception to this structure is the `/tmp/SiaTesting/siatest/data` folder. It contains all the randomly generated files of all `siatest` tests. The files follow the encoding `<sizeInBytes>bytes-<4byte-hex-suffix>`. e.g. `100bytes-deadbeef`

### DIFF
--- a/siatest/consensus/consts.go
+++ b/siatest/consensus/consts.go
@@ -2,7 +2,6 @@ package consensus
 
 import (
 	"os"
-	"testing"
 
 	"github.com/NebulousLabs/Sia/siatest"
 )
@@ -10,8 +9,8 @@ import (
 // consensusTestDir creates a temporary testing directory for a consensus. This
 // should only every be called once per test. Otherwise it will delete the
 // directory again.
-func consensusTestDir(t *testing.T) string {
-	path := siatest.TestDir("consensus", t.Name())
+func consensusTestDir(testName string) string {
+	path := siatest.TestDir("consensus", testName)
 	if err := os.MkdirAll(path, 0777); err != nil {
 		panic(err)
 	}
--- a/siatest/consensus/consts.go
+++ b/siatest/consensus/consts.go
@@ -1,0 +1,19 @@
+package consensus
+
+import (
+	"os"
+	"testing"
+
+	"github.com/NebulousLabs/Sia/siatest"
+)
+
+// consensusTestDir creates a temporary testing directory for a consensus. This
+// should only every be called once per test. Otherwise it will delete the
+// directory again.
+func consensusTestDir(t *testing.T) string {
+	path := siatest.TestDir("consensus", t.Name())
+	if err := os.MkdirAll(path, 0777); err != nil {
+		panic(err)
+	}
+	return path
+}
--- a/siatest/localfile.go
+++ b/siatest/localfile.go
@@ -23,9 +23,9 @@ type (
 
 // NewFile creates and returns a new LocalFile. It will write size random bytes
 // to the file and give the file a random name.
-func NewFile(size int) (*LocalFile, error) {
+func (tn *TestNode) NewFile(size int) (*LocalFile, error) {
 	fileName := fmt.Sprintf("%dbytes-%s", size, hex.EncodeToString(fastrand.Bytes(4)))
-	path := filepath.Join(DataDir(), fileName)
+	path := filepath.Join(tn.filesDir(), fileName)
 	bytes := fastrand.Bytes(size)
 	err := ioutil.WriteFile(path, bytes, 0600)
 	return &LocalFile{
--- a/siatest/localfile.go
+++ b/siatest/localfile.go
@@ -1,11 +1,11 @@
 package siatest
 
 import (
+	"encoding/hex"
+	"fmt"
 	"io/ioutil"
-	"math"
 	"os"
 	"path/filepath"
-	"strconv"
 
 	"github.com/NebulousLabs/Sia/crypto"
 	"github.com/NebulousLabs/errors"
@@ -24,8 +24,8 @@ type (
 // NewFile creates and returns a new LocalFile. It will write size random bytes
 // to the file and give the file a random name.
 func NewFile(size int) (*LocalFile, error) {
-	fileName := strconv.Itoa(fastrand.Intn(math.MaxInt32))
-	path := filepath.Join(SiaTestingDir, fileName)
+	fileName := fmt.Sprintf("%dbytes-%s", size, hex.EncodeToString(fastrand.Bytes(4)))
+	path := filepath.Join(DataDir(), fileName)
 	bytes := fastrand.Bytes(size)
 	err := ioutil.WriteFile(path, bytes, 0600)
 	return &LocalFile{
--- a/siatest/renter/consts.go
+++ b/siatest/renter/consts.go
@@ -2,7 +2,6 @@ package renter
 
 import (
 	"os"
-	"testing"
 
 	"github.com/NebulousLabs/Sia/siatest"
 )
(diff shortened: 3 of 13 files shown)


DavidVorick:
Different data is generated for each test, right? Is there any reason you can't add a data subfolder to the `tmp/SiaTesting/siatest/<packageName>/<testName>` folder?

siatest/consensus/consts.go
```diff
@@ -0,0 +1,19 @@
...
+	"os"
+	"testing"
+
+	"github.com/NebulousLabs/Sia/siatest"
+)
+
+// consensusTestDir creates a temporary testing directory for a consensus. This
+// should only every be called once per test. Otherwise it will delete the
+// directory again.
+func consensusTestDir(t *testing.T) string {
```
DavidVorick: We generally try not to pass around the `testing.T` object, because certain methods of that object only work if you are using it in the main goroutine, and we've run into issues with it before. Better would be to pass around the result of calling `t.Name()` since that's the only thing you need ti for.

siatest/consensus/consts.go
```diff
@@ -0,0 +1,19 @@
...
+	path := siatest.TestDir("consensus", t.Name())
+	if err := os.MkdirAll(path, 0777); err != nil {
+		panic(err)
```
DavidVorick: it's probably okay here but panics are also generally discouraged versus returning an error because a series of deffereed calls can sometimes cause deadlocks, especially if you panic after you called lock on something while using a function that you didn't realize could fail.

If a test hits a deadlock, it'll time out after 600 or 1200 seconds instead of reporting that the true error was a failed call to mkdir.

This might be okay because mkdir really shoudln't ever be failing unless something else is seriously wrong with the filesystem.

siatest/localfile.go
```diff
@@ -24,8 +24,8 @@ type (
 // NewFile creates and returns a new LocalFile. It will write size random bytes
 // to the file and give the file a random name.
 func NewFile(size int) (*LocalFile, error) {
```
DavidVorick: ah, looks like the `NewFile` functions aren't methods of any particular object. I wonder if that doesn't limit the ability to tell which object / test made which file?

siatest/renter/consts.go
```diff
@@ -0,0 +1,19 @@
...
+	"os"
+	"testing"
+
+	"github.com/NebulousLabs/Sia/siatest"
+)
+
+// renterTestDir creates a temporary testing directory for a renter test. This
+// should only every be called once per test. Otherwise it will delete the
+// directory again.
+func renterTestDir(t *testing.T) string {
```
DavidVorick: smae deal here with the panic and the passing around a testing.T object instead of a string that's intended to be the result of t.Name

siatest/testgroup.go
```diff
@@ -48,13 +50,15 @@ var (
 
 // NewGroup creates a group of TestNodes from node params. All the nodes will
 // be connected, synced and funded. Hosts nodes are also announced.
-func NewGroup(nodeParams ...node.NodeParams) (*TestGroup, error) {
+func NewGroup(groupDir string, nodeParams ...node.NodeParams) (*TestGroup, error) {
 	// Create and init group
```
DavidVorick: Now that groups have a dir, it might make more sense to make the new file function a method of the group.

siatest/testgroup.go
```diff
@@ -267,13 +274,36 @@ func mapToSlice(m map[*TestNode]struct{}) []*TestNode {
...
+	if nodeParams.Renter != nil || nodeParams.CreateRenter {
+		nodeDir += "r"
+	}
+	if nodeParams.Host != nil || nodeParams.CreateHost {
+		nodeDir += "h"
+	}
+	if nodeParams.Miner != nil || nodeParams.CreateMiner {
+		nodeDir += "m"
+	}
+	nodeDir += fmt.Sprintf("-%s", hex.EncodeToString(fastrand.Bytes(4)))
```
DavidVorick: any reason that you are doing this over using the `RandomSufffix()` function?

siatest/testgroup.go
```diff
@@ -267,13 +274,36 @@ func mapToSlice(m map[*TestNode]struct{}) []*TestNode {
...
-// randomDir is a helper functions that returns a random directory path
-func randomDir() string {
-	dir, err := TestDir(strconv.Itoa(fastrand.Intn(math.MaxInt32)))
-	if err != nil {
-		panic(errors.AddContext(err, "failed to create testing directory"))
+// randomNodeDir generates a random directory for the provided node params if
+// Dir wasn't set using the provided parentDir and a randomized suffix.
+func randomNodeDir(parentDir string, nodeParams *node.NodeParams) {
+	if nodeParams.Dir != "" {
+		return
```
DavidVorick: should this be returning an error?

siatest/wallet/consts.go
```diff
@@ -0,0 +1,19 @@
...
+	"os"
+	"testing"
+
+	"github.com/NebulousLabs/Sia/siatest"
+)
+
+// walletTestDir creates a temporary testing directory for a wallet test. This
+// should only every be called once per test. Otherwise it will delete the
+// directory again.
+func walletTestDir(t *testing.T) string {
```
DavidVorick: Same here

DavidVorick:
We can break some of this into another PR if you prefer, namely the NewFile stuff

siatest/consensus/consts.go
ChrisSchinnerl: Ok. My thought process was that there is no reason for anyone to ever be able to pass a custom name since it would violate the naming convention.

siatest/consensus/consts.go
ChrisSchinnerl: I know. I think in this case it's nice though since it really shouldn't ever happen during a test and it makes a lot of code cleaner.

siatest/localfile.go
ChrisSchinnerl: Yeah, as I mentioned above, they all land in the same folder since you can't really tell. Passing a path seems to defeat the simplicity this method tries to provide by abstracting everything except for the filesize. Maybe we can find a more satisfying solution than this.

siatest/testgroup.go
ChrisSchinnerl: Good point.

siatest/testgroup.go
ChrisSchinnerl: Uploading and Downloading are methods on the `TestNode` which doesn't know about its group. That means an upload function can't just implicitly create a file because it doesn't hold a reference to the group.

siatest/testgroup.go
ChrisSchinnerl: No, this just means that the user specified a custom directory. In that case we don't need to create one.

siatest/consensus/consts.go
lukechampine: IIRC the main annoyance we had with passing around `testing.T` is that it made failing tests harder to understand. If the helper function failed, you'd only see the name+line of the helper function, not the test that called it, so it could be difficult to track down the precise cause of the failure.
However, Go fixed this by adding the `t.Helper()` method, which causes the helper function to be omitted from the error trace. 

As for the main goroutine, the only issue I'm aware of is that `t.Fatal` only works from the main goroutine. But that's an important concern, because most of the appeal of passing around `testing.T` is that you can hide all the `t.Error` and `t.Fatal` calls in helper functions. So you'd have to be careful to avoid calling helper functions in goroutines.

My opinion on this has shifted over the years, but currently I think it's best to stick with plain old error return values. Go tests should strive to look like normal Go code, which means eschewing magic. So with regard to `t.Name()`, I agree with David that it should be passed as a string, not a `testing.T`. It makes it clear that the function only uses the name and won't cause the test to fail or something, and it also allows for custom names in the rare event that we need one (and most likely, we will).

ChrisSchinnerl:
@DavidVorick Yeah I think we should talk about how we would like to change the `NewFile` stuff and move that into another PR. It might touch more code than expected. Especially if we make it a method of the `TestGroup`.

PR #3124, opened 2018-06-27
Moving encoding helpers into encoding package

lukechampine:
These helpers are really nice for writing clean and efficient `MarshalSia`/`UnmarshalSia` methods, so there's no reason to hide them in the `types` package. Moving them into `encoding` also allows the `Marshal` and `Unmarshal` functions to make use of them.

This is a low-priority PR. It brings a nice performance benefit (approx. half as many allocations when encoding/decoding types without custom marshaler methods), and will be useful any time we want to write `MarshalSia`/`UnmarshalSia` methods down the line, but it certainly isn't blocking anything.

### DIFF
--- a/encoding/marshal.go
+++ b/encoding/marshal.go
@@ -5,6 +5,7 @@ package encoding
 
 import (
 	"bytes"
+	"encoding/binary"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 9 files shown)


encoding/marshal.go
```diff
@@ -55,12 +56,74 @@ type (
...
+	return e.err
+}
+
+// WriteInt writes an int value to the underlying io.Writer.
+func (e *Encoder) WriteInt(i int) error {
+	return e.WriteUint64(uint64(i))
+}
+
+// WritePrefix writes p to the underlying io.Writer, prefixed by its length.
+func (e *Encoder) WritePrefix(p []byte) error {
```
ChrisSchinnerl: Can we rename this to `WriteWithPrefix` ? Every time I come across this method I'm confused since I don't expect it to write the data but only the prefix.

DavidVorick: agree

DavidVorick:
I'm surprised these weren't already in the encoding package.

@lukechampine can you run an IBD on this branch once you are happy with it to make sure it can sync the current chain?

lukechampine:
I settled on `WritePrefixedBytes` so it would have symmetry with `ReadPrefixedBytes` (`ReadWithPrefix` sounds weird).

IBD completed in 2 hours with no issues.

PR #3131, opened 2018-06-29
Updates on Spend Reporting PR 3087

MSevey:
Removed old contracts from Export to avoid breaking backwards compatibility.

Updated releaseblock calculation.

@DavidVorick if there are other edits you find during your review please add TODO's to this PR so we can merge them in.

### DIFF
--- a/cmd/siac/export.go
+++ b/cmd/siac/export.go
@@ -30,7 +30,7 @@ var (
 // renterexportcontracttxnscmd is the handler for the command `siac renter export contract-txns`.
 // Exports the current contract set to JSON.
 func renterexportcontracttxnscmd(destination string) {
-	cs, err := httpClient.RenterActiveContractsGet()
+	cs, err := httpClient.RenterContractsGet()
 	if err != nil {
 		die("Could not retrieve contracts:", err)
 	}
--- a/cmd/siac/export.go
+++ b/cmd/siac/export.go
@@ -35,8 +35,7 @@ func renterexportcontracttxnscmd(destination string) {
 		die("Could not retrieve contracts:", err)
 	}
 	var contractTxns []types.Transaction
-	contracts := append(cs.Contracts, cs.OldContracts...)
-	for _, c := range contracts {
+	for _, c := range cs.Contracts {
 		contractTxns = append(contractTxns, c.LastTransaction)
 	}
 	destination = abs(destination)
--- a/cmd/siac/export.go
+++ b/cmd/siac/export.go
@@ -30,8 +30,7 @@ var (
 // renterexportcontracttxnscmd is the handler for the command `siac renter export contract-txns`.
 // Exports the current contract set to JSON.
 func renterexportcontracttxnscmd(destination string) {
-	getAllContracts := false
-	cs, err := httpClient.RenterContractsGet(getAllContracts)
+	cs, err := httpClient.RenterActiveContractsGet()
 	if err != nil {
 		die("Could not retrieve contracts:", err)
 	}
--- a/cmd/siac/export.go
+++ b/cmd/siac/export.go
@@ -30,7 +30,8 @@ var (
 // renterexportcontracttxnscmd is the handler for the command `siac renter export contract-txns`.
 // Exports the current contract set to JSON.
 func renterexportcontracttxnscmd(destination string) {
-	cs, err := httpClient.RenterContractsGet()
+	getAllContracts := false
+	cs, err := httpClient.RenterContractsGet(getAllContracts)
 	if err != nil {
 		die("Could not retrieve contracts:", err)
 	}
--- a/cmd/siac/rentercmd.go
+++ b/cmd/siac/rentercmd.go
@@ -424,12 +424,13 @@ func rentercontractscmd() {
 			activeTotalSpent = activeTotalSpent.Add(c.TotalCost.Sub(c.RenterFunds).Sub(c.Fees))
 			activeTotalFees = activeTotalFees.Add(c.Fees)
 		}
-		fmt.Printf(`Active Contract Summary:
-			Number of Contracts:  %v
-			Total stored:         %9s
-			Total Remaining:      %v
-			Total Spent:          %v
-			Total Fees:           %v
+		fmt.Printf(`
+Active Contract Summary:
+Number of Contracts:  %v
+Total stored:         %9s
+Total Remaining:      %v
+Total Spent:          %v
+Total Fees:           %v
 			
 			`, len(rc.ActiveContracts), filesizeUnits(int64(activeTotalStored)), currencyUnits(activeTotalRemaining), currencyUnits(activeTotalSpent), currencyUnits(activeTotalFees))
 		w := tabwriter.NewWriter(os.Stdout, 2, 0, 2, ' ', 0)
@@ -450,6 +451,7 @@ func rentercontractscmd() {
 				c.GoodForUpload,
 				c.GoodForRenew)
 		}
+		w.Flush()
 
 		// Display Inactive Contracts
 		sort.Sort(byValue(rc.InactiveContracts))
@@ -461,12 +463,13 @@ func rentercontractscmd() {
 			inactiveTotalSpent = inactiveTotalSpent.Add(c.TotalCost.Sub(c.RenterFunds).Sub(c.Fees))
 			inactiveTotalFees = inactiveTotalFees.Add(c.Fees)
 		}
-		fmt.Printf(`Inactive Contract Summary:
-			Number of Contracts:  %v
-			Total stored:         %9s
-			Total Remaining:      %v
-			Total Spent:          %v
-			Total Fees:           %v
+		fmt.Printf(`
+Inactive Contract Summary:
+Number of Contracts:  %v
+Total stored:         %9s
+Total Remaining:      %v
+Total Spent:          %v
+Total Fees:           %v
 			
 			`, len(rc.InactiveContracts), filesizeUnits(int64(inactiveTotalStored)), currencyUnits(inactiveTotalRemaining), currencyUnits(inactiveTotalSpent), currencyUnits(inactiveTotalFees))
 		w = tabwriter.NewWriter(os.Stdout, 2, 0, 2, ' ', 0)
@@ -513,12 +516,12 @@ func rentercontractscmd() {
 			expiredTotalFees = expiredTotalFees.Add(c.Fees)
 		}
 		fmt.Printf(`
-		Expired Contract Summary:
-		Number of Contracts:  %v
-		Total stored:         %9s
-		Total Remaining:      %v
-		Total Spent:          %v
-		Total Fees:           %v
+Expired Contract Summary:
+Number of Contracts:  %v
+Total stored:         %9s
+Total Remaining:      %v
+Total Spent:          %v
+Total Fees:           %v
 		
 		`, len(rce.ExpiredContracts), filesizeUnits(int64(expiredTotalStored)), currencyUnits(expiredTotalWithheld), currencyUnits(expiredTotalSpent), currencyUnits(expiredTotalFees))
 		w := tabwriter.NewWriter(os.Stdout, 2, 0, 2, ' ', 0)
--- a/cmd/siac/rentercmd.go
+++ b/cmd/siac/rentercmd.go
@@ -379,7 +379,7 @@ func (s byValue) Less(i, j int) bool {
 // rentercontractscmd is the handler for the comand `siac renter contracts`.
 // It lists the Renter's contracts.
 func rentercontractscmd() {
-	rc, err := httpClient.RenterContractsGet()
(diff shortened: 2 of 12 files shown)


DavidVorick:
And you need to update the api documentation for the /renter/contracts endpoint to reflect that old contracts are now returned as well.

And... we should consider having the old contracts get returned as a flag instead of always getting returned, because over time that will build up to be a bunch of data.

MSevey:
>And... we should consider having the old contracts get returned as a flag instead of always getting returned, because over time that will build up to be a bunch of data.

Are you saying to look for the flag in [`renterContractsHandler`](https://github.com/NebulousLabs/Sia/blob/4c73a7ddd05f8e80d345b341328ec6c03ecd2e7b/node/api/renter.go#L275-L363) so that only the active contracts are returned unless the flag is provided?  So adding like an `oldContracts` bool or something?

MSevey:
I also went ahead and renamed `OldContracts` to `ExpiredContracts` since active and expired make more sense than active and old.

DavidVorick:
Yeah, add a bool so that it only returns the expired contract  if a flag is passed in, that way long running renters don't have to worry about getting 50,000 returned objects when they make a query.

cmd/siac/rentercmd.go
```diff
@@ -475,11 +477,11 @@ func rentercontractscmd() {
 // rentercontractsviewcmd is the handler for the command `siac renter contracts <id>`.
 // It lists details of a specific contract.
 func rentercontractsviewcmd(cid string) {
-	rc, err := httpClient.RenterContractsGet()
+	rc, err := httpClient.RenterContractsGet(true)
```
DavidVorick: since it's not obvious at all what `true` means in the context of this function, convention is to use a variable:

```
getAllContracts := true
rc, err := httpCliend.RenterContractsGet(getAllContracts)
```

cmd/siac/export.go
```diff
@@ -30,13 +30,12 @@ var (
 // renterexportcontracttxnscmd is the handler for the command `siac renter export contract-txns`.
 // Exports the current contract set to JSON.
 func renterexportcontracttxnscmd(destination string) {
-	cs, err := httpClient.RenterContractsGet()
+	cs, err := httpClient.RenterContractsGet(false)
```
DavidVorick: since it's not obvious at all what `true` means in the context of this function, convention is to use a variable:

```
getAllContracts := true
rc, err := httpCliend.RenterContractsGet(getAllContracts)
```

cmd/siac/rentercmd.go
```diff
@@ -379,11 +379,11 @@ func (s byValue) Less(i, j int) bool {
 // rentercontractscmd is the handler for the comand `siac renter contracts`.
 // It lists the Renter's contracts.
 func rentercontractscmd() {
-	rc, err := httpClient.RenterContractsGet()
+	rc, err := httpClient.RenterContractsGet(renterAllContracts)
```
DavidVorick: oh wait, looks like this variable was created somewhere

DavidVorick: it should be used in the other calls as well

DavidVorick: oh nvm, this is a flag, not what I was talking about elsewhere

MSevey:
Now the thought is to leave just the one field `Contracts` and add flags to return the type of contracts requested.  Here is a proposal of the flags:

`active - contracts that are in the current period and are goodForRenew and goodForUpload (SH >= CP)
inactive - contracts in the current period that are !goodForRenew or !goodForUpload (SH >= CP)
expired - contracts not in the current period (ie EH <= CP)
all - all contracts`

DavidVorick:
That distinction looks good to me. `inactive` would include contracts that have been renewed due to depletion? (I think it would, as they would be !goodForRewew)

MSevey:
resolves #2308

MSevey:
@DavidVorick currently I'm running into the following edge cases that I need help deciding how to handle.

For reference, here is how I am currently defining the contracts.
From the renter's `Contracts`:
 - `active` contracts are when `goodForRenew && goodForUpload && c.EndHeight > currentPeriod`
 - `inactive` contracts are when `(!goodForRenew || !goodForUpload) && c.EndHeight > currentPeriod`

In both cases `c.EndHeight > currentPeriod` is probably redundant since they shouldn't be in `Contracts` if that were false but I'm leaving it in there for now.

From the renter's `OldContacts`:
 - `inactive` contracts are when `c.EndHeight > currentPeriod`
 - `expired` contracts are when `c.EndHeight <= currentPeriod || currentPeriod == 0`

As you might guess from looking at the `expired` vs `inactive` for the `OldContracts` when the `allowance` is cancelled the `currentPeriod` is equal to zero and both of those evaluate to true.  I can't think of a good way to distinguish between the two if the `allowance` is cancelled and the `currentPeriod` is 0.  Should we just make the definition that if there is no `allowance` all the `OldContracts` are seen as `expired`?

doc/API.md
```diff
@@ -935,7 +935,20 @@ standard success or error response. See
 
 #### /renter/contracts [GET]
 
-returns active contracts. Expired contracts are not included.
+returns all, active, inactive, or expired contracts.  Active contracts are
+contracts that the Renter is currently using to store, upload, and download.
+Inactive contracts are contracts where the end height of the contract is greater
+than the current block height but the contract is either !goodForUpload,
+!goodForRenew, or was renewed due to running out of funds.  Expired contracts
+are contracts with an end height less than the current block height.
```
DavidVorick: `!goodForUpload` and `!goodForRenew` are internal variables of siad, and not something that some person new to the api should be expected to understand. Need to spell out more explicitly that it's for contracts that are no longer being used to store data.

On that note, contracts that are `!goodForUpload && goodForRenew` should be considered active, the only inactive contracts are ones that are `!goodForRenew`.

Finally, this comment doesn't indicate what happens if none of the optional flags are passed in to fetch contracts, and so the comment needs to be expanded to explain the default behavior. I believe the default behavior should be to return active contracts, because that is what maintains compatibility with how it was before. (though maybe it should contain both active and inactive).

doc/api/Renter.md
```diff
@@ -148,7 +148,20 @@ standard success or error response. See
 
 #### /renter/contracts [GET]
 
-returns active contracts. Expired contracts are not included.
+returns all, active, inactive, or expired contracts.  Active contracts are
+contracts that the Renter is currently using to store, upload, and download.
+Inactive contracts are contracts where the end height of the contract is greater
+than the current block height but the contract is either !goodForUpload,
+!goodForRenew, or was renewed due to running out of funds.  Expired contracts
+are contracts with an end height less than the current block height.
```
DavidVorick: Need to update this comment block as well

siatest/renter/renter_test.go
```diff
@@ -2175,91 +2495,86 @@ func checkContractVsReportedSpending(r *siatest.TestNode, WindowSize types.Block
...
+			return fmt.Errorf("Download spending on renewed contract equal to %v, expected zero", c.DownloadSpending.HumanString())
 		}
 	}
 	return nil
 }
 
 // renewContractByRenewWindow mines blocks to force contract renewal
+//
+// TODO:
+// 1) remove excess tg.Sync() calls in the test since it is called in this
```
DavidVorick: which test?

siatest/renter/renter_test.go
```diff
@@ -2175,91 +2495,86 @@ func checkContractVsReportedSpending(r *siatest.TestNode, WindowSize types.Block
...
+// function
+// 2) make this more generic by referencing the contracts
+// 3) look for other places where we are mining blocks in tests and see if
```
DavidVorick: should probably turn this TODO into a `programming task`

MSevey: Yes I agree, I'll pull out my todo's into a programming task.  I was just documenting them as I went as to not forget.

cmd/siac/rentercmd.go
```diff
@@ -452,31 +489,41 @@ func rentercontractscmd() {
 		}
 		w.Flush()
 	}
+
 	if renterAllContracts {
-		if len(rc.OldContracts) == 0 {
+		rc, err := httpClient.RenterExpiredContractsGet()
```
MSevey: need to rename `rc` as this would overwrite the previous api call, and since `RenterExpiredContractsGet()` doesn't return `InactiveContracts` the following `if` statement could potentially say that there are no contracts when the renter has only `InactiveContracts`

cmd/siac/rentercmd.go
```diff
@@ -503,9 +550,8 @@ func rentercontractsviewcmd(cid string) {
 	if err != nil {
 		die("Could not get contract details: ", err)
 	}
-	contracts := append(rc.Contracts, rc.OldContracts...)
 
-	for _, rc := range contracts {
+	for _, rc := range rc.Contracts {
```
MSevey: need to update this to pull all active, inactive, and expired contracts to then search through

modules/renter/renter.go
```diff
@@ -113,7 +113,7 @@ type hostContractor interface {
 	// Contracts returns the active contracts formed by the contractor.
 	Contracts() []modules.RenterContract
 
-	// Contracts returns the old contracts formed by the contractor.
+	// OldContracts returns the Expired contracts formed by the contractor.
```
MSevey: update these comments to match those in `modules/renter.go`

node/api/renter.go
```diff
@@ -273,9 +275,36 @@ func (api *API) renterHandlerPOST(w http.ResponseWriter, req *http.Request, _ ht
...
-func (api *API) renterContractsHandler(w http.ResponseWriter, _ *http.Request, _ httprouter.Params) {
+// renterContractsHandler handles the API call to request the Renter's
+// contracts.
+//
+// Active contracts are contracts that the renter is actively using to store
+// data and can upload, download, and renew
+//
+// Inactive contracts are contracts that are not currently being used by the
+// renter because they are either !goodForUpload or !goodForRenew, but have
+// endheights that are in the future so could potentially become active again
```
MSevey: Update comment

siatest/renter/renter_test.go
```diff
@@ -1339,64 +1383,219 @@ func TestRenterOldContracts(t *testing.T) {
 	}
 	tg, err := siatest.NewGroupFromTemplate(groupParams)
 	if err != nil {
-		t.Fatal("Failed to create group: ", err)
+		t.Fatal("Failed to get files")
```
MSevey: revert, not sure how this got changed

siatest/renter/renter_test.go
```diff
@@ -1528,8 +1727,8 @@ func TestRenterPersistData(t *testing.T) {
 	}
 }
 
-// TestRenterResetAllowance tests that resetting the allowance after the
-// allowance was cancelled will trigger the correct contract formation.
+// TestRenterCancelAllowance tests that setting an empty allowance causes
+// uploads, downloads, and renewals to cease.
 func TestRenterResetAllowance(t *testing.T) {
```
MSevey: figure out why this comment doesn't match the test name

siatest/renter/renter_test.go
```diff
@@ -1668,6 +1870,7 @@ func TestRenterSpendingReporting(t *testing.T) {
 	}
 
 	// Get renter's initial siacoin balance
+	r := tg.Renters()[0]
```
MSevey: remove and go back to checking for `nodes[0]`

siatest/renter/renter_test.go
```diff
@@ -2016,8 +2251,6 @@ func TestRenterSpendingReporting(t *testing.T) {
 	}
 }
 
-// The following are helper functions for the renter tests
-
 // checkBalanceVsSpending checks the renters confirmed siacoin balance in their
```
MSevey: add comment back in, didn't mean to delete

MSevey:
`test-long` is passing locally

MSevey:
siac if no contracts have been formed
image

MSevey:
fixed column headers
image

node/api/renter.go
```diff
@@ -314,52 +342,69 @@ func (api *API) renterContractsHandler(w http.ResponseWriter, _ *http.Request, _
...
+				StartHeight:               c.StartHeight,
+				StorageSpending:           c.StorageSpending,
+				StorageSpendingDeprecated: c.StorageSpending,
+				TotalCost:                 c.TotalCost,
+				UploadSpending:            c.UploadSpending,
+			}
+			if expired && c.EndHeight < blockHeight {
+				expiredContracts = append(expiredContracts, contract)
+			} else if inactive && c.EndHeight >= blockHeight {
+				inactiveContracts = append(inactiveContracts, contract)
```
lukechampine: Shouldn't this be two separate `if`s, not an `else if`? What if you want both expired and inactive?

DavidVorick: no contract will be both inactive and expired, this is fine. We should add a comment though.

PR #3137, opened 2018-07-02
clean up siac renter download

DavidVorick:
Wasn't happy with the implementation.

I removed the unnecessary 1 second sleep, switched it so that there are no goroutines or channels needed, and I added some extra safety stuff like a timer for a potential infinite loop.

I also added a bunch of TODOs for things that aren't fully done correctly in the code yet.

### DIFF
--- a/cmd/siac/rentercmd.go
+++ b/cmd/siac/rentercmd.go
@@ -564,14 +564,14 @@ func renterfilesdownloadcmd(path, destination string) {
 
 	// If the download is async, report success.
 	if renterDownloadAsync {
-		fmt.Printf("\nQueued Download '%s' to %s.\n", path, abs(destination))
+		fmt.Printf("Queued Download '%s' to %s.\n", path, abs(destination))
 		return
 	}
 
 	// If the download is blocking, display progress as the file downloads.
 	err = downloadprogress(path)
 	if err != nil {
-		die("Download could not be completed:", err)
+		die("\nDownload could not be completed:", err)
 	}
 	fmt.Printf("\nDownloaded '%s' to %s.\n", path, abs(destination))
 }
--- a/cmd/siac/rentercmd.go
+++ b/cmd/siac/rentercmd.go
@@ -580,7 +580,7 @@ func renterfilesdownloadcmd(path, destination string) {
 // user, and return an error when the download is finished.
 func downloadprogress(siapath string) error {
 	// If the download does not appear after a long time, give up.
-	tries := 0
+	var tryTime time.Duration
 
 	for range time.Tick(OutputRefreshRate) {
 		// Get the list of downloads.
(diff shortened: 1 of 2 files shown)


DavidVorick:
I changed the number of tries to a try duration.

I kept it returning an error instead of panicking because in practice that error gets fed directly to a user, and I think it's more friendly to have an error like that than to panic on the user.

DavidVorick:
Ok, this code is now to my satisfaction.

cmd/siac/rentercmd.go
```diff
@@ -1,5 +1,23 @@
 package main
 
+// TODO: If you run siac from a non-existant directory, the abs() function does
```
ChrisSchinnerl: I think those are pretty good `programming tasks`. Would be good if you created issues for them too.

cmd/siac/rentercmd.go
```diff
@@ -534,48 +553,79 @@ func renterfilesdeletecmd(path string) {
...
-					break
-				}
+	for range time.Tick(OutputRefreshRate) {
+		// Get the list of downloads.
+		queue, err := httpClient.RenterDownloadsGet()
+		if err != nil {
+			continue // benign
+		}
+
+		// Search for the download in the list of downloads.
```
ChrisSchinnerl: This might also be a good programming task. A new endpoint that let's you fetch a single download.

cmd/siac/rentercmd.go
```diff
@@ -534,48 +553,79 @@ func renterfilesdeletecmd(path string) {
...
+				break
 			}
-			if d.Filesize == 0 {
-				continue // file hasn't appeared in queue yet
+		}
+		// If the download has not appeared in the queue yet, either continue or
+		// give up.
+		if !found {
+			tryTime += OutputRefreshRate
+			if tryTime > time.Minute {
```
ChrisSchinnerl: Should this also be a constant instead of just `time.Minute` ?

cmd/siac/rentercmd.go
```diff
@@ -534,48 +553,79 @@ func renterfilesdeletecmd(path string) {
...
+		queue, err := httpClient.RenterDownloadsGet()
+		if err != nil {
+			continue // benign
+		}
+
+		// Search for the download in the list of downloads.
+		var d api.DownloadInfo
+		found := false
+		for _, d = range queue.Downloads {
+			if d.SiaPath == siapath {
```
ChrisSchinnerl: I think this is actually not correct. The same `siapath` might be in the download history multiple times.

cmd/siac/rentercmd.go
lukechampine: probably clearer to write
```go
start := time.Now()
// ...
    if time.Since(start) > time.Minute {
```

cmd/siac/rentercmd.go
DavidVorick: It ... should ... grab the most recent one. Otherwise we have no way to distinguish.

DavidVorick: I guess if source and dest are both the same you are okay, we can at least add that

PR #3139, opened 2018-07-03
Clean up siatest testrenter and improve runtime

MSevey:
Summary of edits:
`renter_test.go` - reordered tests so that they are cleanly grouped and alphabetized so it is easier to work with.  combined interrupt tests into `TestRenterInterrupt` so that as a group they can be run in parallel to improve the runtime and not error out due to `too many files open`

`testgroup.go` - updated `AddNodes` to return the nodes added.  Needed this in order to combine the interrupt tests.

`testgroup_test.go` - added unit test for `AddNodes` to confirm the node added was returned.

### DIFF
--- a/siatest/testgroup_test.go
+++ b/siatest/testgroup_test.go
@@ -134,6 +134,9 @@ func TestAddNewNode(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	if len(nodes) != 1 {
+		t.Fatalf("More nodes returned than expected; expected 1 got %v", len(nodes))
+	}
 	renter := nodes[0]
 	for _, oldRenter := range oldRenters {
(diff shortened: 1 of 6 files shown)


siatest/testgroup_test.go
```diff
@@ -98,3 +99,45 @@ func TestNewGroupNoRenterHost(t *testing.T) {
...
+	// Record current nodes
+	oldRenters := tg.Renters()
+
+	// Test adding a node
+	testDir, err := TestDir(t.Name())
+	if err != nil {
+		t.Fatal(err)
+	}
+	renterTemplate := node.Renter(testDir + "/renter")
+	nodes, err := tg.AddNodes(renterTemplate)
```
ChrisSchinnerl: You should also check the length of `nodes` since it might have a different length than `1`.

PR #3143, opened 2018-07-04
Don't aggregate inputs via a parent tx when sending Siacoins

huetsch:
Currently the wallet module builds 2 transactions each time a user wants to send Siacoins. It does this via the `FundSiacoins` function. The first transaction aggregates all inputs into an output equalling the desired amount to be sent (and potentially a refund). The second transaction spends the desired amount output from the first transaction and creates an output to the actual desired recipient.

To the best of my knowledge, `FundSiacoins` is built this way to conveniently support transactions without outputs. In the frequent case where someone just wants to send Siacoins elsewhere, `SendSiacoins` and `SendSiacoinsMulti` in their current form seem to create twice as many transactions as necessary. This unnecessarily bloats the blockchain and can slow down the 2nd party's receipt of the Siacoins.

This PR creates a specialized function `FundSiacoinsForOutputs` to be used in the special case of a transaction that just sends Siacoins to an output or a set of outputs. `FundSiacoinsForOutputs` eliminates the double transaction issue.

Please let me know if I've missed an intentional design reason for the way the old code worked!

### DIFF
--- a/modules/wallet.go
+++ b/modules/wallet.go
@@ -115,6 +115,13 @@ type (
 	//
 	// Transaction builders are not thread safe.
 	TransactionBuilder interface {
+		// FundSiacoinForOutputs will aggregate enough inputs to cover the
+		// total value of the outputs and the miner fee if any. A refund
+		// output will be generated if necessary. All the outputs will be
+		// added to the transaction being built along with a miner fee if
+		// one is passed. The transaction will not be signed.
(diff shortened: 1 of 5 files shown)


DavidVorick:
We never built specialized functions for sending siacoins because we were putting our dev hours elsewhere. I haven't looked at this code yet, but there's nothing fundamentally wrong with the idea of using specialized functions to reduce the total overhead of sending money around the Sia platform - it is on our long term todo list.

PR #3155, opened 2018-07-11
Decrypt chunk pieces in parallel and in-place

ChrisSchinnerl:
Encrypting/Decrypting a whole chunk with a single thread is pretty slow. This PR moves that code to the worker threads to encrypt/decrypt on a per-piece basis.

### DIFF
--- a/crypto/encrypt.go
+++ b/crypto/encrypt.go
@@ -61,6 +61,35 @@ func (key TwofishKey) EncryptBytes(plaintext []byte) Ciphertext {
 	return aead.Seal(nonce, nonce, plaintext, nil)
 }
 
+// EncryptBytesInPlace encrypts a []byte using the key. EncryptBytesInPlace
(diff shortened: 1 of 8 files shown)


lukechampine:
are there places where we don't want to decrypt in-place?

ChrisSchinnerl:
@lukechampine It looks like quite a few tests in the wallet are failing if I change it to be in-place by default.

DavidVorick:
Need to add a new function to the encryption package that does encryption in-place. Then only have that function be called in the renter so that compatibility in the other places that we use encryption is not threatened.

ChrisSchinnerl:
I reverted the commits for the parallel encryption for now since it is just too much of a hack to do it in-place right now. Once we get rid of the nonce in the sector after moving to Threefish, we can implement a much easier in-place solution for this.

PR #3157, opened 2018-07-11
Update initial current period to be set in the past when allowance is created

### LINKED ISSUE #3117: Renter Allowance initial current period in the past
When the renter's allowance is first set, the current period should be set in the past

MSevey:
`TestRenterSpendingReporting` is failing in the siatest package, but I want #3131 to be merged so I can rebase before debugging.

This resolves #3117

### DIFF
--- a/modules/renter/contractor/allowance.go
+++ b/modules/renter/contractor/allowance.go
@@ -65,7 +65,7 @@ func (c *Contractor) SetAllowance(a modules.Allowance) error {
 	// set the current period to the blockheight if the existing allowance is
 	// empty
 	if reflect.DeepEqual(c.allowance, modules.Allowance{}) {
-		c.currentPeriod = c.blockHeight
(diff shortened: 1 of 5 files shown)


modules/renter/contractor/allowance.go
```diff
@@ -65,7 +65,7 @@ func (c *Contractor) SetAllowance(a modules.Allowance) error {
 	// set the current period to the blockheight if the existing allowance is
 	// empty
 	if reflect.DeepEqual(c.allowance, modules.Allowance{}) {
-		c.currentPeriod = c.blockHeight
+		c.currentPeriod = c.blockHeight - a.RenewWindow
```
MSevey: Currently the siatest is failing due to an underflow where the test package blockheight is less than the allowance renew window.  Should I bother checking for underflow here or can we assume that will never happen in real life because that would mean someone is setting their renew window to something like 3 years.

PR #3158, opened 2018-07-12
Contract maintenance

DavidVorick:
The `threadedContractMaintenance` thread was pretty opaque, and had some areas related to funding that were not very understandable, and probably not correct. I refactored the way that the loop checks for funds remaining, and I broke some of the more complex pieces of logic out into their own functions.

I think the tests are passing, though they actually keep timing out on my machine. The siatest renter tests seem to be taking more than 500 seconds, but not because one is deadlocked, rather by the time the later tests are going, the earlier tests have already pushed the timer over 400 seconds.

### DIFF
--- a/modules/renter/contractor/contracts.go
+++ b/modules/renter/contractor/contracts.go
@@ -58,7 +58,7 @@ func (c *Contractor) managedContractUtility(id types.FileContractID) (modules.Co
 // to be extended to support adding up all the parent spending too. These
 // spending estimates will apply to uploading and downloading, but not to
 // storage or fees or contract price.
-func (c *Contractor) managedEstimateRenewFundingRequirements(contract modules.RenterContract, allowancePeriod types.BlockHeight, blockHeight types.BlockHeight) (types.Currency, error) {
+func (c *Contractor) managedEstimateRenewFundingRequirements(contract modules.RenterContract, blockHeight types.BlockHeight, allowance modules.Allowance) (types.Currency, error) {
 	// Fetch the host pricing to use in the estimate.
 	host, exists := c.hdb.Host(contract.HostPublicKey)
 	if !exists {
@@ -68,7 +68,7 @@ func (c *Contractor) managedEstimateRenewFundingRequirements(contract modules.Re
 	// Estimate the amount of money that's going to be needed for existing
 	// storage.
 	dataStored := contract.Transaction.FileContractRevisions[0].NewFileSize
-	maintenanceCost := types.NewCurrency64(dataStored).Mul64(uint64(allowancePeriod)).Mul(host.StoragePrice)
+	maintenanceCost := types.NewCurrency64(dataStored).Mul64(uint64(allowance.Period)).Mul(host.StoragePrice)
 
 	// Estimate the amount of money that's going to be needed for new storage
 	// based on the amount of new storage added in the previous period. Account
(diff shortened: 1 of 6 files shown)


modules/renter/contractor/contracts.go
```diff
@@ -283,6 +359,132 @@ func (c *Contractor) managedRenew(sc *proto.SafeContract, contractFunding types.
...
+	oldUtility, ok := c.managedContractUtility(id)
+	if !ok || !oldUtility.GoodForRenew {
+		c.log.Printf("Contract %v slated for renew is marked not good for renew %v/%v",
+			id, ok, oldUtility.GoodForRenew)
+		c.staticContracts.Return(oldContract)
+		return types.ZeroCurrency, errors.New("contract is marked not good for renew")
+	}
+
+	// Calculate endHeight for renewed contracts
+	endHeight := currentPeriod + allowance.Period
```
MSevey: there is a function [`contractEndHeight()` ](https://github.com/NebulousLabs/Sia/blob/8da8220708ed647b7fa9e466a0b2ee5355666e47/modules/renter/contractor/contracts.go#L27) which can be called on the `Contractor`.  We should probably use that instead of manually calculating the endHeight in multiple places right?

modules/renter/contractor/contracts.go
```diff
@@ -331,185 +533,94 @@ func (c *Contractor) threadedContractMaintenance() {
 		return
 	}
 
-	// Figure out which contracts need to be renewed, and while we have the
-	// lock, figure out the end height for the new contracts and also the amount
-	// to spend on each contract.
+	// Create the renewSet and refreshSet, each a list of contracts that need to
```
MSevey: `each is a list` would read more easily

modules/renter/contractor/contracts.go
```diff
@@ -331,185 +533,94 @@ func (c *Contractor) threadedContractMaintenance() {
...
-		if fundsAvailable.Cmp(overspend) > 0 {
-			// We still have some funds available.
-			fundsAvailable = fundsAvailable.Sub(overspend)
-		} else {
-			// The overspend exceeds the available funds, set available
-			// funds to zero.
-			fundsAvailable = types.ZeroCurrency
-		}
-	}
+	endHeight := currentPeriod + allowance.Period
```
MSevey: same here about using `contractEndHeight()`

modules/renter/contractor/contracts.go
```diff
@@ -331,185 +533,94 @@ func (c *Contractor) threadedContractMaintenance() {
...
+	// allowance does not have enough money, the contractor will prefer to save
+	// data in the long term rather than renew a contract.
+	var renewSet []fileContractRenewal
+	var refreshSet []fileContractRenewal
+
+	// The rest of this function needs to know a few of the stateful variables
+	// from the contractor, build those up under a lock so that the rest of the
+	// function can execute without lock contention.
 	c.mu.RLock()
 	currentPeriod := c.currentPeriod
```
MSevey: shouldn't need to define `currentPeriod` if we use `contractEndHeight()`

modules/renter/contractor/contracts.go
DavidVorick: contractEndHeight needs a lock

modules/renter/contractor/contracts.go
DavidVorick: " without lock contention"

Issue #3130, opened 2018-06-29
Deduplicate host and gateway IP discovery code


ChrisSchinnerl:
It looks like we currently have almost the same IP discovery code in the host and gateway, except that the gateway now has a decentralized way to do it.
We might be able to simply put the gateway IP discovery code into a `publicIP` method and have an exported `PublicIP` method that gets called by the host instead of using it's own code.

After that is done a good followup PR might be to also deduplicate the UPnP code of the host.

PR #3161, opened 2018-07-12
Gateway discover

ChrisSchinnerl:
Resolves [#3130](https://github.com/NebulousLabs/Sia/issues/3130).


modules/gateway/upnp.go
```diff
@@ -42,6 +42,46 @@ func myExternalIP() (string, error) {
...
 }
 
+// managedLearnHostname tries to discover the external ip of the machine. If
+// discovering the address failed or if it is invalid, an error is returned.
+func (g *Gateway) managedLearnHostname(cancel <-chan struct{}) (modules.NetAddress, error) {
+	// create ctx to cancel upnp discovery during shutdown
+	ctx, ctxCancel := context.WithCancel(context.Background())
+	defer ctxCancel()
+	go func() {
+		timer := time.NewTimer(timeoutIPDiscovery)
```
lukechampine: the timer can be part of the context:
```go
ctx, ctxCancel := context.WithTimeout(context.Background(), timeoutIPDiscover)
defer ctxCancel()
go func() {
    select {
    case <-cancel:
        ctxCancel()
    case <-g.threads.StopChan():
        ctxCancel()
    case <-ctx.Done():
    }
}()
```

Issue #3148, opened 2018-07-09
Siac renter download rolling transfer speed


DavidVorick:
Currently the `siac renter download` command displays the transfer speed as an average over the whole life of the download, which is not the way that most programs display download speed. Instead, siac should display the current download speed as a function of the total amount of data downloaded over the past 'n' seconds, where 'n' is probably 60 (unless the download is less than n seconds old, in which case it should use the download's current duration as the denominator)

PR #3162, opened 2018-07-13
Siac renter download rolling transfer speed

ChrisSchinnerl:
Resolves [#3148](https://github.com/NebulousLabs/Sia/issues/3148)

[demo](https://asciinema.org/a/yfuDqI5UPUFH3rmBfHpOqDISP?t=0:15)

### DIFF
--- a/cmd/siac/consts.go
+++ b/cmd/siac/consts.go
@@ -7,7 +7,7 @@ import (
 const (
 	// OutputRefreshRate is the rate at which siac will update something like a
 	// progress meter when displaying a continuous action like a download.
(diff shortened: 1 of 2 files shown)


ChrisSchinnerl:
Theoretically a `Ybps` is `2^80` bps which is too large for the `uint64` of `bandwidthUnit` but you can never be future proof enough :P

ChrisSchinnerl:
Something that still bothers me is the way we update the progress of the download. We only update it whenever a chunk is complete instead of when a piece is complete. That means that downloading a 40MB file will be stuck at 0% until the chunk is downloaded an then go right to 100%.

PR #3164, opened 2018-07-13
Add API endpoint to cancel a Renter's contract

MSevey:


ChrisSchinnerl:
Would it make more sense to allow the user to set all the utility fields manually instead of just setting all the fields to `false`?

PR #3174, opened 2018-07-20
fix #3173

### LINKED ISSUE #3173: cannot read a persisted json object file with manual checksum
I modified some content in miner.json and changed checksum to manual

"Miner Settings"
"0.5.0"
"manual"
{
... content here ...
}

siad treat it as a wrong checksum and ignore the file content.

https://github.com/NebulousLabs/Sia/blob/master/persist/json.go#L70

```golang
	// Determine whether the leading bytes contain a checksum. A proper checksum
	// will be 67 bytes (quote, 64 byte checksum, quote, newline). A manual
	// checksum will be the characters "manual\n" (9 characters). If neither
	// decode correctly, it is assumed that there is no checksum at all.
	var checksum crypto.Hash
	err = json.Unmarshal(remainingBytes[:67], &checksum)
	if err == nil && checksum == crypto.HashBytes(remainingBytes[68:]) {
		// Checksum is proper, and matches the data. Update the data portion to
		// exclude the checksum.
		remainingBytes = remainingBytes[68:]
	} else {
		// Cryptographic checksum failed, try interpreting a manual checksum.
		var manualChecksum string
		err := json.Unmarshal(remainingBytes[:8], &manualChecksum)    // Line 70
		if err == nil && manualChecksum == "manual" {
			// Manual checksum is proper. Update the remaining data to exclude
			// the manual checksum.
			remainingBytes = remainingBytes[9:]    // Line 74
		}
	}
```

In line 70, it try to `json.Unmarshal` with `remainingBytes[:8]` which is only 7 bytes.
`json.Unmarshal([]byte("\"manual"), &manualChecksum)` is always return an error (`unexpected end of JSON input`), it should be changed to `remainingBytes[:9]`, and also, the `remainingBytes[9:]` in line 74 should be changed to `remainingBytes[10:]`.

No

**Expected Behavior**
treat "manual" as a correct checksum

**How to reproduce it (as minimally and precisely as possible)**
1. run siad with miner module (siad -M gctwm)
2. modify third line of <datadir>/miner/miner.json to "manual"
3. restart siad with miner module
4. it takes very long time to load miner module because latestChangeID from this file is ignored by siad.

**Environment**
* Sia version: v1.3.3
* OS: Debian 8.9 (it doesn't matter)

wincss:
fix the slice position

### DIFF
--- a/persist/json.go
+++ b/persist/json.go
@@ -67,11 +67,11 @@ func readJSON(meta Metadata, object interface{}, filename string) error {
 	} else {
 		// Cryptographic checksum failed, try interpreting a manual checksum.
 		var manualChecksum string
-		err := json.Unmarshal(remainingBytes[:8], &manualChecksum)
+		err := json.Unmarshal(remainingBytes[:9], &manualChecksum)
 		if err == nil && manualChecksum == "manual" {
 			// Manual checksum is proper. Update the remaining data to exclude
 			// the manual checksum.
-			remainingBytes = remainingBytes[9:]
+			remainingBytes = remainingBytes[10:]
 		}
 	}
 


DavidVorick:
What the hell? How did we miss this?

@wincss thanks a bunch, definitely appreciate it. Not sure what you went through to discover this bug but I'm amazed we haven't had issues with it before.

DavidVorick:
But also, we've migrated to gitlab: https://gitlab.com/NebulousLabs/Sia

PR #3189, opened 2018-08-01
revert to c222a06

lukechampine:
switching import paths broke old code that imported `github.com/NebulousLabs/Sia`. Instead of continuing to mirror the GitLab repo, I suggest that we revert the GitHub repo to its state immediately prior to the move, and then archive it with the README altered as in #3188. This will ensure that old code still works, while also marking this repo as deprecated. (I have already disabled mirroring from GitLab.)


tbenz9:
Can we disable GitHub issues as well?

lukechampine:
I believe archiving disables new issues from being opened. https://github.com/boltdb/bolt is an example.

